PR #12, opened 2026-06-22
Add configurable context window caps

0disoft:
## Summary

Add optional provider-level and model-level context window caps for the Codex model catalog.

## Why

Some upstream providers report very large context windows from `/models` metadata, such as 1M-token models. Syncing those values directly into Codex can make Codex retain much larger histories than an operator wants, increasing latency, cost, and practical context-management risk.

## Changes

- Add `contextWindow` to cap all models for a provider.
- Add `modelContextWindows` to cap specific provider model ids.
- Apply configured caps after live `/models` metadata and cached metadata so user config wins.
- Document the config fields in the README.

## Validation

- `bun install`
- `bun x tsc --noEmit`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,26 @@ Local models work too. Point opencodex at any OpenAI-compatible server running o
 
 WebSocket transport is off by default. Set `"websockets": true` only if you want Codex to advertise and use the Responses WebSocket path instead of HTTP/SSE.
 
+Some providers report very large context windows from `/models`. To keep Codex from retaining more
+history than you want, cap the Codex-visible context window per provider or per model:
+
+```json
+{
+  "providers": {
+    "openrouter": {
+      "adapter": "openai-chat",
+      "baseUrl": "https://openrouter.ai/api/v1",
+      "apiKey": "${OPENROUTER_API_KEY}",
+      "contextWindow": 262144,
+      "modelContextWindows": {
+        "z-ai/glm-5.2": 262144,
+        "deepseek/deepseek-v4-pro": 262144
+      }
+    }
+  }
+}
+```
+
 See the **[Configuration reference](https://lidge-jun.github.io/opencodex/reference/configuration/)** for every field.
 
 ## Documentation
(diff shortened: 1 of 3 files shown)


lidge-jun:
Review note before merging into dev:

This is close, but I would not merge it as-is. The current implementation applies configured contextWindow after live /models metadata and cached metadata, so a configured value can raise a smaller provider-reported context window. That is override behavior, not cap behavior. For example, if /models reports 128k and config sets contextWindow=262144 as a provider-wide cap, the catalog would advertise 262144 instead of keeping 128k.

Recommended follow-up before main merge: treat contextWindow/modelContextWindows as max caps for live or cached metadata, while still using the configured value as the fallback context for purely configured/static models with no live metadata. Add tests for provider-wide cap, model-specific cap, cached metadata cap, and config value greater than live metadata. Also current main has v2.1.6 README/types changes, so this PR needs simple conflict resolution on README.md and src/types.ts.

0disoft:
Updated in `5befa16`.

- Rebased/merged current `upstream/main` and resolved the README/types conflict.
- Changed `contextWindow` / `modelContextWindows` to behave as max caps for live `/models` metadata and fresh cached metadata.
- Kept configured values as fallback for configured/static models with no live metadata.
- Added regression coverage for provider-wide cap, model-specific cap, cached metadata cap, and config value greater than live metadata.

Verified locally with Bun canary on Windows:

```text
bun test tests/codex-catalog.test.ts
19 pass, 0 fail
```

lidge-jun:
Implemented on dev in 376c0c5.

This preserves the context-window cap behavior from this PR and applies it across configured/static models, live /models metadata, and cached metadata. The final implementation keeps configured values as caps: they can lower oversized live metadata, but they do not raise a smaller live context window.

Credit is included in the dev commit:

Co-authored-by: 0disoft <rodisoft1@gmail.com>

I'm closing this PR as superseded by the dev commit so the branch does not need a separate merge path. Thank you for the contribution.

Issue #11, opened 2026-06-22
Bug: 注入 opencodex 后 Codex App 左侧 Project 线程列表消失


l269438:
## 问题描述

使用 opencodex 给 Codex 增加/切换模型后，重新打开 Codex App，左侧 Project 里的历史线程列表消失，看起来像线程都没了。

排查后确认本地数据并没有被删除：`~/.codex/state_5.sqlite`、`~/.codex/session_index.jsonl`、`~/.codex/history.jsonl` 和 `~/.codex/.codex-global-state.json` 都还在，sqlite 完整性检查也是 `ok`。所以更像是 opencodex 注入配置后，Codex App 读取/过滤项目线程列表出现异常。

## 复现步骤

1. 在 macOS 上安装并初始化 opencodex。
2. 通过 opencodex 添加/启用模型，让它注入 Codex 配置。
3. opencodex 代理启动后，`~/.codex/config.toml` 中默认 provider 被改成：

   ```toml
   model_provider = "opencodex"

   [model_providers.opencodex]
   base_url = "http://localhost:10100/v1"
   ```

4. 退出并重新打开 Codex App。
5. 观察左侧 Project 区域，历史项目线程列表消失。

## 期望行为

opencodex 注入模型 provider 后，不应该影响 Codex App 左侧 Project/线程列表的展示。历史项目和线程应该照常可见。

## 实际行为

重新进入 Codex App 后，左侧 Project 中看不到原来的线程，用户会误以为线程数据被删了。

但实际本地数据仍存在：

```text
~/.codex/state_5.sqlite: 398 threads, 378 unarchived
~/.codex/.codex-global-state.json: electron-saved-workspace-roots 仍有 22 个项目路径
sqlite pragma integrity_check: ok
```

`threads` 表里旧线程主要还是 `model_provider = openai`，新线程出现了 `model_provider = opencodex`。怀疑 Codex App 的项目/线程列表在 provider、模型 catalog 或注入后的配置状态上出现了过滤/兼容问题。

## 临时恢复方式

执行：

```bash
ocx restore
```

后，`~/.codex/config.toml` 中 opencodex 路由会被移除，models cache 也恢复为 native models。然后完全退出并重新打开 Codex App，理论上应恢复原生 Codex 读取逻辑。

建议 opencodex 在注入/恢复时：

- 不影响 Codex App 的 Project/线程列表读取；
- 或在注入前备份并明确提示可能影响 App UI；
[...]

PR #13, opened 2026-06-22
Preserve Codex resume history by default

0disoft:
## Summary

Make Codex resume-history remapping an explicit compatibility mode instead of an automatic side effect.

## Why

Codex App appears to scope project/thread visibility by the active provider. That means existing OpenAI-backed project chats can disappear from the sidebar while `model_provider = "opencodex"`, even though the local thread rows still exist.

Users reasonably expect existing chats to remain visible. The practical workaround is to remap old `openai` resume-history rows to `opencodex` while the proxy is active. That behavior is useful, but it mutates Codex's local thread index and should be an explicit compatibility choice rather than something that happens just because the proxy started.

This is related to #11. This PR does not claim to fix every App-side filtering path; it makes the history remap explicit, documented, and still reversible through restore.

## Changes

- Add `syncResumeHistory` as an explicit compatibility-mode config flag.
- Leave existing Codex resume history untouched by default during injection.
- Print a startup note explaining that existing OpenAI project chats may be hidden unless `syncResumeHistory=true` is enabled.
- Keep restore behavior intact so previously remapped `opencodex` rows can still be mapped back to `openai`.
- Document the tradeoff in the README.

## Validation

- `bun install`
- `bun x tsc --noEmit`
- `git diff --check`
- Manual Windows Codex App check: with `syncResumeHistory=true`, existing `zerodi-wd1` project chats became visible again while opencodex was active, and opencodex models appeared in the model picker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,9 +243,11 @@ Local models work too. Point opencodex at any OpenAI-compatible server running o
 
 WebSocket transport is off by default. Set `"websockets": true` only if you want Codex to advertise and use the Responses WebSocket path instead of HTTP/SSE.
 
-opencodex leaves existing Codex resume history untouched by default so Codex App project chats keep
-showing up under their original provider. If you explicitly want old OpenAI-backed threads to be
-remapped to the opencodex provider while the proxy is active, set `"syncResumeHistory": true`.
+opencodex leaves existing Codex resume history untouched by default. This avoids changing Codex's
+local thread index just because the proxy started, but Codex App may hide old OpenAI-backed project
+threads while `opencodex` is the active provider. If you want those existing chats to appear while
+the proxy is active, enable the compatibility remap with `"syncResumeHistory": true`. Restore still
+maps previously remapped `opencodex` rows back to `openai`.
 
 See the **[Configuration reference](https://lidge-jun.github.io/opencodex/reference/configuration/)** for every field.
 
--- a/README.md
+++ b/README.md
@@ -243,6 +243,10 @@ Local models work too. Point opencodex at any OpenAI-compatible server running o
 
 WebSocket transport is off by default. Set `"websockets": true` only if you want Codex to advertise and use the Responses WebSocket path instead of HTTP/SSE.
 
+opencodex leaves existing Codex resume history untouched by default so Codex App project chats keep
+showing up under their original provider. If you explicitly want old OpenAI-backed threads to be
+remapped to the opencodex provider while the proxy is active, set `"syncResumeHistory": true`.
+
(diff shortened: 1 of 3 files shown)


lidge-jun:
Maintainer review update: I merged this PR locally on `dev` for reproduction testing, but I am treating it as only the safety half of #11.

What #13 does correctly:

- stops default automatic mutation of Codex resume history;
- keeps the provider remap behind explicit `syncResumeHistory: true`;
- warns users that existing OpenAI project chats can be hidden while opencodex is active if they do not opt in.

What it does not solve by itself:

- maintainer repro shows opencodex-created project rows are mostly `model_provider = opencodex` but `source = exec`;
- Codex App appears to require more than provider match for sidebar visibility, likely a resumable/source filter too;
- therefore provider remapping alone can restore old OpenAI `cli`/`vscode` rows, but not opencodex-created `exec` rows.

I am continuing the fix in `devlog/220_codex-app-history-visibility/00_plan.md`: investigate Codex App/codex-rs filtering, then implement only a reversible opencodex-side compatibility path.

lidge-jun:
Follow-up pushed to `dev` in `3244d28` (`fix: eject opencodex history on native restore`). Attribution: this builds on @0disoft's PR #13 (`Preserve Codex resume history by default`), which correctly moved history mutation behind explicit `syncResumeHistory` opt-in.

While testing PR #13 / issue #11 behavior locally, I found that the previous conservative restore behavior was still not enough: if `ocx stop` removes `[model_providers.opencodex]` from `config.toml` while any history row still has `model_provider = "opencodex"`, Codex App cannot resume that thread and reports `Model provider opencodex not found`.

So the `dev` branch now treats native restore as a stronger invariant:

- do not leave user-resumable history pointing at a provider id that no longer exists in `config.toml`;
- restore backed-up OpenAI rows to OpenAI;
- eject remaining opencodex user rows to OpenAI during `ocx stop` / `ocx restore`;
- promote `exec` source to `cli` on native restore so opencodex-created project chats can be visible in the native App sidebar;
- strip root `model = "provider/model"` routed model values when removing opencodex routing.

This preserves PR #13’s core safety intent (no start-time history mutation unless `syncResumeHistory` is enabled), but fixes the stop/eject side so native Codex is not left with orphaned opencodex provider references.

Verified locally: `bun run typecheck` passed; `bun test tests` passed with 142 pass / 0 fail. Also repaired the local dev machine state: 218 user history rows recovered to `openai`, zero remaining opencodex user rows after `ocx stop`.

lidge-jun:
Released in v2.1.6 via #20. The resume-history compatibility path is now on main with the later safety fixes: manifest-backed restore, native restore cleanup, and explicit legacy recovery handling. npm latest is @bitkyc08/opencodex@2.1.6.

PR #14, opened 2026-06-22
Guard subcommand help before side effects

0disoft:
## Summary
Prevent subcommand help flags from executing side-effecting commands.

## Why
`ocx restore --help` currently enters the `restore` switch case and mutates Codex config instead of showing help. That behavior was noted while triaging #11, and the same command shape can affect other side-effecting subcommands.

## Changes
- Add a shared subcommand help guard for `--help`, `-h`, and `help`.
- Print subcommand-specific usage before the main command switch can run side effects.
- Add a regression test proving `restore --help` leaves `CODEX_HOME/config.toml` unchanged.

## Validation
- `bun install`
- `bun test tests/cli-help.test.ts`
- `bun x tsc --noEmit`
- `git diff --check`

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -38,6 +38,69 @@ Examples:
   ocx sync                    Sync available models to Codex`);
 }
 
+function hasHelpFlag(values: string[]): boolean {
+  return values.some(value => value === "--help" || value === "-h" || value === "help");
+}
+
+function printSubcommandUsage(name: string | undefined): void {
+  switch (name) {
+    case "init":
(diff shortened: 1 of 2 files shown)


lidge-jun:
Merged via #20 and released in v2.1.6. Authorship was preserved on the local merge commit: 60aad9f authored by @0disoft. I also added a dev follow-up test for the newer recover-history subcommand help path before the release.

lidge-jun:
Closing because this was merged through #20 and released in v2.1.6. Authorship is preserved in commit 60aad9f by @0disoft, with the dev-specific recover-history help regression covered by 29337a0.

PR #16, opened 2026-06-22
[codex] Allow static provider model catalogs

0disoft:
## Summary

Add a `liveModels` provider option so operators can pin a static model allowlist instead of fetching a provider's live `/models` catalog during startup/catalog sync.

## Why

Some providers expose very large or slow live model catalogs. For providers such as OpenRouter, users may want deterministic Codex model-picker entries and faster startup by advertising only the explicitly configured models.

This does not attempt to fix any Bun runtime crash. It gives users a stable catalog path that avoids live provider catalog fetches when those fetches are unnecessary or unreliable.

## Changes

- Add `OcxProviderConfig.liveModels?: boolean`.
- When `liveModels` is `false`, use configured `models` directly and skip the provider `/models` fetch.
- Ensure a fresh live-model cache does not override an explicit static allowlist.
- Add catalog tests for the no-fetch path and cache precedence.
- Document static model allowlists in the English, Korean, and Chinese configuration references.

## Validation

- `C:\Users\cherr\.bun-canary\bin\bun.exe run typecheck`
- `C:\Users\cherr\.bun-canary\bin\bun.exe test tests`
- `git diff --check`

### DIFF
--- a/docs-site/src/content/docs/ko/reference/configuration.md
+++ b/docs-site/src/content/docs/ko/reference/configuration.md
@@ -29,11 +29,31 @@ opencodex는 `~/.opencodex/config.json`으로 설정됩니다. 이 파일은 `oc
 | `baseUrl` | `string` | 업스트림 API 기본 URL. |
 | `apiKey?` | `string` | API 키, 또는 요청 시점에 해석되는 `${ENV_VAR}` / `$ENV_VAR` 참조. |
 | `defaultModel?` | `string` | 명시적인 모델 없이 이 프로바이더가 선택되었을 때 사용하는 모델. |
-| `models?` | `string[]` | 시드/폴백 모델 목록 (실시간 `/models`에 접근 가능하면 그쪽이 우선됨). |
+| `models?` | `string[]` | 시드/폴백 모델 목록. `liveModels`가 `false`이면 Codex 카탈로그에 노출할 정확한 allowlist가 됩니다. |
+| `liveModels?` | `boolean` | 시작/동기화 시 프로바이더의 실시간 `/models` 카탈로그를 가져옵니다(기본 `true`). `false`로 두면 설정된 `models`만 사용합니다. |
 | `headers?` | `Record<string,string>` | 업스트림으로 전송되는 추가 HTTP 헤더. |
 | `authMode?` | `"key" \| "forward" \| "oauth"` | 인증 방식 (기본 `key`). [프로바이더](/opencodex/ko/guides/providers/#auth-modes) 참조. |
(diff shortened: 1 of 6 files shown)


lidge-jun:
Review note before merging into dev:

Must fix: the static path writes configured models into the live-model cache ( in ). That makes  correctly ignore an existing live cache, but it creates the reverse stale-cache problem: if an operator later flips  back to true, the next sync can treat the old static allowlist as a fresh live  result until TTL expires. The live cache should only contain data fetched from the provider's live catalog; the static path can just return  without , and a regression test should cover false -> true toggling.

Also note: current main has the v2.1.6 Umans config docs/type additions, so this PR will need straightforward doc/type conflict resolution when merged.

lidge-jun:
Review note before merging into dev:

Must fix: the static path writes configured models into the live-model cache via setCached(name, configured) in fetchProviderModels. That makes liveModels=false correctly ignore an existing live cache, but it creates the reverse stale-cache problem: if an operator later flips liveModels back to true, the next sync can treat the old static allowlist as a fresh live /models result until TTL expires. The live cache should only contain data fetched from the provider live catalog. The static path can just return configured without setCached, and a regression test should cover false -> true toggling.

Also note: current main has the v2.1.6 Umans config docs/type additions, so this PR will need straightforward doc/type conflict resolution when merged.

0disoft:
Updated in `9bef30b`.

- Removed `setCached(name, configured)` from the `liveModels === false` static path so the live-model cache only stores provider-fetched catalog data.
- Added a regression test for `liveModels: false` -> `liveModels: true` toggling to ensure the next sync fetches live `/models` instead of reusing the old static allowlist.

Verified locally with Bun canary on Windows:

```text
bun test tests/codex-catalog.test.ts
18 pass, 0 fail
```

lidge-jun:
Absorbed into `dev` without using the PR merge button.

Integrated in:
- `38bca18 feat: allow static provider model catalogs` (co-authored with @0disoft)
- `e586acd fix: honor static allowlists during catalog augmentation`
- `3a8852c docs: record pr16 pr22 integration evidence`

I also added maintainer-side follow-up coverage for `opencode-go`/jawcode augmentation so `liveModels:false` remains an exact allowlist, including the empty-`models` case.

Verification on `dev`:
- `bun run typecheck` passed
- `bun test tests` passed: 165/165
- `cd docs-site && bun install --frozen-lockfile && bun run build` built 46 pages
- read-only verifier returned DONE after the follow-up fix

Closing this PR as superseded by the dev integration. Thanks for the contribution.

PR #19, opened 2026-06-22
[codex] Source dashboard badge version from runtime health

0disoft:
## What changed

- Updates the sidebar brand badge to prefer the running proxy version from `/healthz`.
- Keeps the build-time `__APP_VERSION__` value as a fallback while the runtime health endpoint is unavailable.
- Updates the Vite/version declaration comments so they no longer describe the build-time value as the dashboard source of truth.

## Why

The dashboard can show two different versions at once when an older GUI bundle is served with a newer proxy runtime. In the observed case, the sidebar badge showed `v2.1.3` while the main dashboard version card showed `2.1.5` from `/healthz`.

The root cause is that the sidebar badge used a build-time package version, while the main dashboard card used the runtime health response. The runtime proxy version is the better source of truth for what is actually running.

## Validation

- `bun run build:gui`
- Confirmed `/healthz` returned `version: 2.1.5` locally.
- Confirmed the served dashboard bundle no longer contained `2.1.3` and did contain the `/healthz` runtime version lookup.

### DIFF
--- a/gui/src/App.tsx
+++ b/gui/src/App.tsx
@@ -24,6 +24,12 @@ const NAV: { id: Page; tkey: TKey; Icon: typeof IconGrid }[] = [
 const THEME_ICON = { light: IconSun, dark: IconMoon, system: IconMonitor } as const;
 const THEME_TKEY: Record<Theme, TKey> = { light: "theme.light", dark: "theme.dark", system: "theme.system" };
 
+function readRuntimeVersion(data: unknown): string | null {
+  if (!data || typeof data !== "object" || !("version" in data)) return null;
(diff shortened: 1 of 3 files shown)


lidge-jun:
Merged via #20 and released in v2.1.6. Authorship was preserved on the local merge commit: b8c126c authored by @0disoft. The dashboard now prefers the runtime /healthz version and falls back to __APP_VERSION__.

lidge-jun:
Closing because this was merged through #20 and released in v2.1.6. Authorship is preserved in commit b8c126c by @0disoft.

PR #23, opened 2026-06-23
fix: preserve Anthropic stream input usage

0disoft:
## Summary
- accumulate Anthropic stream usage from `message_start` and `message_delta`
- emit a single terminal `done` on `message_stop`, with EOF fallback when usage was already seen
- add coverage for streamed input/cache usage plus output usage merging

## Tests
- `bun test tests/adapter-usage.test.ts`
- `bun test tests`
- `bun run typecheck`

### DIFF
--- a/src/adapters/anthropic.ts
+++ b/src/adapters/anthropic.ts
@@ -63,6 +63,14 @@ function usageFromAnthropic(usage: Record<string, number> | undefined): OcxUsage
   };
 }
 
+function mergeAnthropicUsage(
+  base: Record<string, number> | undefined,
+  next: Record<string, number> | undefined,
+): Record<string, number> | undefined {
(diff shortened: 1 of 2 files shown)


lidge-jun:
놀러갔다 오겠습니다 ㅎㅎ..

lidge-jun:
Merged into local dev with a follow-up commit that changes `mergeAnthropicUsage` from spread-override to additive merge — prevents silent overwrite if Anthropic ever sends overlapping keys across `message_start` and `message_delta`. All 167 tests pass including your two new usage tests. Thanks!

lidge-jun:
Merged into dev manually with a follow-up should-fix (additive usage merge). Thanks for the contribution!

PR #26, opened 2026-06-24
fix: preserve invalid config before fallback

### LINKED ISSUE #25: Warn or recover explicitly when config.json cannot be parsed
## Summary

`loadConfig()` currently returns the default config when `~/.opencodex/config.json` cannot be parsed, without warning the user.

Current code:

```ts
try {
  const raw = readFileSync(CONFIG_PATH, "utf-8");
  return JSON.parse(raw) as OcxConfig;
} catch {
  return getDefaultConfig();
}
```

## Why this matters

If the config file is truncated or manually edited into invalid JSON, opencodex silently starts with defaults. Later code paths can save the in-memory default config, making the user's provider setup appear to have disappeared.

This is especially confusing because provider credentials/configuration are user-managed state, and a parse failure likely indicates a recoverable local file problem rather than a real first-run state.

## Possible directions

- Print a warning that includes the config path and parse/read error.
- Preserve the invalid file as `config.json.invalid.<timestamp>` before writing a replacement.
- Refuse to overwrite the config until the user fixes or confirms recovery.
- Add a small test around invalid JSON handling so this does not regress silently.

I am filing this separately because the desired recovery behavior is partly UX policy, not only implementation.

0disoft:
## Summary
- validate the minimum required `config.json` shape when loading opencodex config
- preserve invalid config files by copying them to `config.json.invalid-<timestamp>` before falling back to defaults
- warn the user once per process when the fallback path is used
- honor `OPENCODEX_HOME` for config/pid paths so config behavior can be tested without touching the real home directory

Fixes #25

## Verification
- `C:\Users\cherr\.bun-canary\bin\bun.exe test tests/config.test.ts`
- `C:\Users\cherr\.bun-canary\bin\bun.exe run typecheck`
- `C:\Users\cherr\.bun-canary\bin\bun.exe test tests`

### DIFF
--- a/src/codex-account-store.ts
+++ b/src/codex-account-store.ts
@@ -3,17 +3,21 @@ import { join } from "node:path";
 import { getConfigDir, atomicWriteFile, hardenConfigDir, hardenExistingSecret } from "./config";
 import type { CodexAccountCredentials } from "./types";
 
-const CODEX_ACCOUNTS_PATH = join(getConfigDir(), "codex-accounts.json");
 type CodexAccountStore = Record<string, CodexAccountCredentials>;
 
 const REFRESH_SKEW_MS = 60_000;
 
+function getCodexAccountsPath(): string {
+  return join(getConfigDir(), "codex-accounts.json");
+}
+
 export function loadCodexAccountStore(): CodexAccountStore {
+  const path = getCodexAccountsPath();
   hardenConfigDir();
-  hardenExistingSecret(CODEX_ACCOUNTS_PATH);
-  if (!existsSync(CODEX_ACCOUNTS_PATH)) return {};
+  hardenExistingSecret(path);
+  if (!existsSync(path)) return {};
   try {
-    return JSON.parse(readFileSync(CODEX_ACCOUNTS_PATH, "utf-8")) as CodexAccountStore;
+    return JSON.parse(readFileSync(path, "utf-8")) as CodexAccountStore;
   } catch {
     return {};
   }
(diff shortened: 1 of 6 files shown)


lidge-jun:
Thanks for the original fix. I rebuilt the intent on top of the current `dev` branch instead of merging this stale branch directly, because current `dev` has since added multi-auth credential generation/tombstone/refresh-lock hardening and this PR now conflicts with those files.

Landed on `dev` as:

- `4d9d9f9 fix: preserve invalid config before fallback`
- Co-authored-by: `0disoft <rodisoft1@gmail.com>`

What was preserved from this PR:

- invalid `config.json` is copied to `config.json.invalid-<timestamp>` before fallback;
- fallback logs a warning with the config path and reason;
- warning/backup happens once per config path per process;
- regression tests cover invalid JSON and invalid config shape.

What changed for current `dev` compatibility:

- kept the existing dynamic `OPENCODEX_HOME` path helpers already present on `dev`;
- did not touch `src/codex-account-store.ts`, preserving current generation/tombstone/CAS refresh hardening;
- allowed existing `port: 0` and `providers: {}` test/runtime contracts while still rejecting missing required shape and mismatched non-empty `defaultProvider`.

Verification:

- local `bun run typecheck`: pass
- local `bun run privacy:scan`: pass
- local `bun test tests`: 356 pass / 0 fail
- GitHub Cross-platform CI on `dev` run `28138885594`: ubuntu + windows success

Closing this PR as superseded by the credited `dev` rebuild.

PR #27, opened 2026-06-24
feat(cli): add --background / -b flag to ocx start

flyingsquirrel0419:
## What this PR does
- Adds `-b` / `--background` option to `ocx start`
- Runs the proxy as a detached background process so users can close the terminal without killing the server
- Prevents recursive background spawns via `OCX_BACKGROUND_SPAWNED`
- Updates CLI help text, README files (EN/KO/ZH), and docs-site CLI reference
- Adds tests for the new flag

## Use case
Useful on remote Linux servers where users want the proxy to keep running after disconnecting SSH, without setting up a full systemd service.

## Example
```bash
ocx start -b
ocx start --port 8080 --background
```

## Checklist
- [x] Type check passes
- [x] All tests pass
- [x] Manually verified: detached process runs and survives terminal close
- [x] `ocx stop` still stops the background proxy correctly

### DIFF
--- a/README.ko.md
+++ b/README.ko.md
@@ -126,7 +126,7 @@ codex -m "xai/grok-4"                  "이 PR을 리뷰해 줘"
 
 ```bash
 ocx init                       # 대화형 설정
-ocx start [--port 10100]       # 프록시 시작; 포트가 사용 중이면 빈 포트로 자동 전환
+ocx start [--port 10100] [-b|--background]  # 프록시 시작; --background로 터미널 분리; 포트가 사용 중이면 빈 포트로 자동 전환
 ocx stop                       # 프록시 중지 + Codex 원래 설정 복원
 ocx restore                    # 중지 없이 복원 (별칭: ocx eject)
 ocx uninstall                  # service/shim/config 제거 + Codex 원본 복원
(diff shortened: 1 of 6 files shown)


lidge-jun:
Thanks for the PR. Closing this as a docs/UX decision rather than merging it.

The user-facing goal of `ocx start --background` overlaps with the existing lifecycle surfaces:

- `ocx service install` already provides the always-on background mode through launchd/systemd/Task Scheduler, including login start, crash restart, and stop/uninstall integration.
- `ocx codex-shim install` already provides the lightweight on-demand mode by running `ocx ensure` before ordinary `codex` commands.
- `ocx ensure` already starts the proxy as a detached child when needed.

Adding a third `ocx start -b` path would make the startup model harder to explain and would blur the product boundary between service mode and shim/on-demand mode. The current PR also only tests CLI help text, not detached lifecycle behavior such as PID ownership, terminal-close survival, and `ocx stop` cleanup.

If we later decide that a one-shot detached start is still needed for remote SSH/server workflows, it should come back as a more explicit expert feature, documented separately from service/shim and covered by lifecycle tests. For now, service + shim are the intended supported UX.

Issue #44, opened 2026-06-28
Request logs can miss successful ChatGPT native passthrough turns


0disoft:
## Summary

I am seeing a request-log gap for successful ChatGPT/native passthrough turns. Routed providers such as `umans` and `opencode-go` appear in `/api/logs`, but successful `gpt-5.5` / `chatgpt` turns do not show up. A non-stream `chatgpt` 503 does show up, so the issue seems specific to successful native passthrough SSE finalization rather than all ChatGPT requests.

This looks like a follow-up edge case around the native passthrough SSE workaround from #31.

## Observed evidence

Current runtime:

- `ocx status --json` reports proxy `v2.6.0`, PID `32544`, health OK, default provider `chatgpt`.
- Codex is configured to route through opencodex at `http://localhost:10100/v1`.
- `/api/logs` currently groups as:
  - `umans`: 104 entries, all `200`
  - `opencode-go`: 18 entries, all `200`
  - `chatgpt`: 1 entry, `503`
- The only `chatgpt` entry is:
  - local time `2026-06-28 16:44:07 +09:00`
  - model `gpt-5.5`
  - status `503`
  - errorCode `server_is_overloaded`
  - closeReason `non_stream`
  - usageStatus `unreported`
- During the same session, Codex session JSONL showed completed `gpt-5.5` turns with `model_provider: "opencodex"` / `comp_hash: "opencodex"`, but those successful turns were not visible in `/api/logs`.
- The proxy stdout log repeatedly contains:

```text
1 | (function (controller, error) {"use strict";
              ^
TypeError: null is not an object
```

## Code path that looks suspicious

In `src/server.ts`, native passthrough SSE uses `upstreamResponse.body.tee()`:

- branch 0: returned directly as the client-facing `Response` body via `markNativePassthroughSseResponse(...)`
- branch 1: consumed in the background by `consumeForInspection(...)`

Then `responseWithDeferredRequestLog(...)` skips normal request-log wrapping for native passthrough SSE responses:

```ts
if (isNativePassthroughSseResponse(response)) {
  return response;
}
```

So successful native passthrough logs depend on the background inspection path calling `onNativePassthroughTerminal`.

The cancel path looks like it can suppress finalization:

- `consumeForInspection(...)` sets `cancelled = true` and calls `reader.cancel(...)` when the signal aborts.
- The fallback terminal reporting uses `if (!reported && !cancelled) onTerminal("incomplete")`.
[...]

PR #46, opened 2026-06-29
fix(logs): finalize native passthrough SSE usage

0disoft:
## Summary

- finalize request logs for native passthrough SSE even when no Codex pool terminal recorder is active
- preserve the existing native relay path used for the Bun SSE passthrough workaround
- add a regression test that verifies `/api/logs` and `/api/usage` receive reported usage from a completed native passthrough SSE response

Refs #44.

## Why

Native passthrough SSE responses are marked with `markNativePassthroughSseResponse(...)`, so `responseWithDeferredRequestLog(...)` intentionally skips the normal client-facing SSE logging wrapper. Before this change, the background inspection path finalized the request log only when `terminalBodyWillRecord` was true, which requires a Codex forward/pool terminal recorder.

That left key-based native `openai-responses` passthrough streams able to deliver the upstream response while not finalizing the request log entry. The usage dashboard then could miss the completed request, or fail to count reported usage that was present in the terminal SSE payload.

## Changed Behavior

| Case | Before | After |
| --- | --- | --- |
| Native passthrough SSE with no pool terminal recorder | background inspection could read metadata without finalizing `/api/logs` | background inspection finalizes terminal/cancel logging and still records usage metadata |
| Codex forward/pool passthrough SSE | terminal recorder finalized quota/health and request log | unchanged, with the terminal recorder called when present |
| WebSocket/native relay workaround | native SSE response still bypasses the JS client-facing wrapper | unchanged |

## Verification

- `bun test tests/server-auth.test.ts` -> 30 pass
- `bun run typecheck` -> pass
- `bun test tests/request-log.test.ts tests/passthrough-abort.test.ts` -> 22 pass
- `git diff --check` -> pass

## Notes

This only records usage when the upstream terminal SSE payload includes usage. If the native ChatGPT/OpenAI upstream omits usage entirely, the request can still be finalized as unreported; this PR does not synthesize usage from Codex session files.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -518,14 +518,13 @@ async function handleResponses(
       const turnAc = new AbortController();
       linkAbortSignal(upstream, turnAc.signal);
       registerTurn(turnAc);
-      if (terminalBodyWillRecord && recordTerminalOutcomes) {
-        const recordTerminal = terminalRecorder;
+      if (recordTerminalOutcomes) {
         const reportNativeTerminal = (status: ResponsesTerminalStatus) => {
           if (options.abortSignal?.aborted) {
             options.onNativePassthroughCancel?.();
             return;
           }
-          recordTerminal(status);
+          terminalRecorder?.(status);
           options.onNativePassthroughTerminal?.(status);
         };
         consumeForInspection(inspectBody, reportNativeTerminal, turnAc.signal, () => unregisterTurn(turnAc), logCtx);
--- a/tests/server-auth.test.ts
+++ b/tests/server-auth.test.ts
@@ -843,6 +843,87 @@ describe("server local API auth", () => {
     }
   });
 
+  test("native passthrough SSE records completed usage without pool terminal tracking", async () => {
+    if (existsSync(TEST_DIR)) rmSync(TEST_DIR, { recursive: true });
+    mkdirSync(TEST_DIR, { recursive: true });
+    process.env.OPENCODEX_HOME = TEST_DIR;
+
+    const upstream = Bun.serve({
+      port: 0,
+      fetch() {
+        return new Response(
+          [
+            "event: response.completed",
+            'data: {"type":"response.completed","response":{"status":"completed","model":"gpt-5.5","usage":{"input_tokens":11,"output_tokens":7,"input_tokens_details":{"cached_tokens":3},"output_tokens_details":{"reasoning_tokens":2}}}}',
+            "",
+            "",
+          ].join("\n"),
+          { headers: { "content-type": "text/event-stream" } },
+        );
+      },
+    });
+    saveConfig({
+      port: 0,
+      defaultProvider: "test-openai",
+      providers: {
+        "test-openai": {
+          adapter: "openai-responses",
+          baseUrl: upstream.url.toString(),
+          apiKey: "provider-key",
+          defaultModel: "gpt-5.5",
+        },
+      },
+    } as OcxConfig);
+
+    const server = startServer(0);
+    try {
+      const response = await fetch(new URL("/v1/responses", server.url), {
+        method: "POST",
+        headers: { "content-type": "application/json" },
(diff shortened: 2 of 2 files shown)


lidge-jun (commented):
Reviewed against the #44 root-cause analysis — this is on the right track and well-scoped.

**What it fixes:** the gate at `src/server.ts:528` changes from `if (terminalBodyWillRecord && recordTerminalOutcomes)` to `if (recordTerminalOutcomes)`, with `terminalRecorder?.(status)` optional-chained. So native passthrough SSE now runs `consumeForInspection` (which finalizes `/api/logs` + records usage) even when no Codex pool terminal recorder is active. That closes the secondary gap we identified — the `consumeForResponseLogMetadata` else-branch that inspected metadata but never finalized an entry. The regression test (completed native SSE with usage → `/api/logs` + `/api/usage` both reflect 18 tokens) is a good, targeted proof.

**One gap worth confirming before merge:** the *primary* symptom in #44 is **successful** turns going missing because Codex disconnects the client the instant it finishes consuming, which aborts the shared signal while the teed inspection branch (branch[1]) is still draining. On that path `consumeForInspection` sets `cancelled = true` and the abort listener cancels the reader — so `onTerminal("incomplete")` is guarded out (`if (!reported && !cancelled)`) and `reportNativeTerminal` never fires. This PR makes the recorder *reachable* without a pool recorder, but if the success terminal SSE block hasn't been parsed before the abort lands, the entry can still be missed. Two things would de-risk it:

1. A regression test for **success-then-immediate-client-disconnect** (not just clean completion), asserting exactly one `/api/logs` entry.
2. Confirm whether a cancel finalizer (`499` / `closeReason: "client_cancel"`) is still wanted on the abort path, or is intentionally out of scope for this PR.

Verification in the description (server-auth 30 pass, typecheck pass, request-log + passthrough-abort 22 pass) looks solid. No double-logging concern given the `reported` guard. Targets `dev` — fine. Approving direction; suggest the success-then-disconnect test before merge.

lidge-jun:
Reviewed against the #44 root-cause analysis — on the right track and well-scoped.

**What it fixes:** the gate at `src/server.ts:528` changes from `if (terminalBodyWillRecord && recordTerminalOutcomes)` to `if (recordTerminalOutcomes)`, with `terminalRecorder?.(status)` optional-chained. So native passthrough SSE now runs `consumeForInspection` (which finalizes `/api/logs` + records usage) even when no Codex pool terminal recorder is active. That closes the secondary gap from #44 — the `consumeForResponseLogMetadata` else-branch that inspected metadata but never finalized an entry. The regression test (completed native SSE with usage → `/api/logs` + `/api/usage` both reflect 18 tokens) is a good, targeted proof.

**One gap worth confirming before merge:** the *primary* symptom in #44 is **successful** turns going missing because Codex disconnects the client the instant it finishes consuming, aborting the shared signal while the teed inspection branch is still draining. On that path `consumeForInspection` sets `cancelled = true`, so `onTerminal("incomplete")` is guarded out (`if (!reported && !cancelled)`) and `reportNativeTerminal` never fires. This PR makes the recorder *reachable* without a pool recorder, but if the success terminal SSE block isn't parsed before the abort lands, the entry can still be missed. Two things would de-risk it:

1. A regression test for **success-then-immediate-client-disconnect** (not just clean completion), asserting exactly one `/api/logs` entry.
2. Confirm whether a cancel finalizer (`499` / `closeReason: "client_cancel"`) is wanted on the abort path, or is intentionally out of scope.

Verification (server-auth 30 pass, typecheck pass, request-log + passthrough-abort 22 pass) looks solid; `reported` guard avoids double-logging; targeting `dev` is fine. Approving the direction — suggest the success-then-disconnect test before merge.

lidge-jun:
Content integrated into dev via cherry-pick as 2481c80 (recorded in af6304e 'docs(merge): record PR #46 cherry-pick'), now released through preview. Closing as the patch has already landed.

PR #48, opened 2026-06-29
feat(providers): allow disabling providers

0disoft:
## Summary
- Add a `disabled` provider flag so users can hide a provider without deleting its saved configuration.
- Exclude disabled providers from routing, routed model gathering, catalog/model listings, and ChatGPT sidecar provider selection.
- Add Providers-page Enable/Disable actions, status/default badges, and localized copy.
- Reject disabling the current default provider so routing cannot be left without a live default.

## Validation
- `bun test tests/router.test.ts tests/codex-catalog.test.ts tests/server-auth.test.ts`
- `bun run typecheck`
- `cd gui && bun run build`


lidge-jun:
Manually merged into dev (commit 55d525b) with conflict resolution against the provider-context-cap work. Verified: tsc clean, 778 tests pass, GUI build OK. Closing since the fork PR can't be auto-detected as merged.

PR #55, opened 2026-07-02
fix: support Google Gemini /models lfix: support Google Gemini models listing via x-goog-api-keyisting via x-goog-api-key header

simplvis:
Google Generative Language API uses x-goog-api-key header instead of Authorization: Bearer, and its /models response returns { models: [{ name: "models/gemini-..." }] } rather than OpenAI's { data: [{ id: "gemini-..." }] } format.
This fix handles both issues so that the Google provider's models appear in the opencodex dashboard and Codex model picker.

### DIFF
--- a/src/codex-catalog.ts
+++ b/src/codex-catalog.ts
@@ -687,7 +687,14 @@ async function fetchProviderModels(name: string, prov: OcxProviderConfig, ttlMs:
       return stale ? applyConfigHintsToCachedModels(name, prov, stale, contextCap) : configured;
     }
     const json = await res.json() as { data?: ProviderModelsApiItem[] };
-    const live = (json.data ?? []).map(m => applyProviderConfigHints(name, prov, {
+    const jsonData = (json as any).data;
+    const models = name === "google" && prov.adapter === "google" && Array.isArray((json as any).models)
+      ? ((json as any).models as Array<{name?: string; displayName?: string}>).map((m: any) => ({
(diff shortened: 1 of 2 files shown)


lidge-jun:
Thanks @simplvis — good catch on both the header and the `/v1beta` endpoint. We shipped the same header/endpoint fix independently on our stabilization branch (84cef2f, which also covers the ADC/OAuth `x-goog-user-project` path and routes through the new `googleFetch` retry wrapper from 6cb379b, so this PR would now conflict with / regress those). Your `nextPageToken` pagination handling was missing from our version though — we're porting that part with credit (`Co-authored-by`) in the next release. Closing this PR as superseded; the pagination improvement lands via the stabilization branch. Thanks again for the contribution! 🙏

lidge-jun:
Correction to my note above: after re-checking the actual diff, this PR contains no pagination handling — that part of my earlier comment was wrong, apologies for the confusion. The accurate summary: 84cef2f on the stabilization branch fully supersedes this PR — same `x-goog-api-key` + `/v1beta` intent, plus the `{ models: [...] }` response-shape parsing, `pageSize=1000` (the API's cap, so no `pageToken` loop is needed), `generateContent` filtering with `inputTokenLimit` context hints, tests, and scoping via `googleMode` so vertex/antigravity request paths are unaffected (this patch as written also had a template-literal syntax error in `buildModelsRequest`). Nothing further to port. Thanks again for identifying the root cause — the fix ships in the next release.

PR #66, opened 2026-07-06
feat(providers): add anthropic-apikey built-in for direct Anthropic API-key billing

phj1081:
## Problem

The built-in `anthropic` provider is `authKind: "oauth"`, and `routedProviderConfig()` canonicalizes any user `authMode` override back to the registry's authKind. So a user who sets an API key on the `anthropic` provider is still routed through the Claude-account OAuth path — requests burn (and 429 on) the subscription quota instead of API billing, which is confusing to debug because the config *looks* like it's on API-key auth.

The only workaround today is a custom provider entry, which loses registry metadata: jawcode context windows, static model seeds, and the GUI icon.

## Change

Add a first-class `anthropic-apikey` registry entry (`authKind: "key"`), mirroring the existing `openai` / `openai-apikey` split:

- Shares the OAuth entry's static model seed + context windows via a hoisted `ANTHROPIC_MODELS` const, with `liveModels: true` since Anthropic's `/v1/models` works with `x-api-key` (`buildModelsRequest` already handles the anthropic adapter branch).
- Reuses the `anthropic` jawcode bundle for model metadata, plus an `anthropic-key` metadata alias for pre-existing custom configs with that name.
- Key-login flow works out of the box: `dashboardUrl` points at console.anthropic.com API keys, and `validateApiKey()` already probes the anthropic adapter with `x-api-key`.
- GUI: claude icon alias; `src/generated/jawcode-model-metadata.ts` PROVIDER_ALIASES updated to match `deriveJawcodeAliases()` (same edit the generator would emit).

This keeps OAuth and API-key routing configurable side by side — e.g. subscription for daily use, API key as fallback when the account is rate-limited.

## Tests

- New parity case asserting `anthropic-apikey` mirrors the OAuth entry's models/context windows and key-login fields.
- Updated `EXPECTED_KEY_PROVIDER_IDS`, featured set, and jawcode alias expectations.
- `bun test`: 1429 pass / 0 fail. `bun x tsc --noEmit` clean.

### DIFF
--- a/gui/src/provider-icons.ts
+++ b/gui/src/provider-icons.ts
@@ -1,5 +1,6 @@
 const PROVIDER_ICON_ALIASES: Record<string, string> = {
   anthropic: "claude-color.svg",
+  "anthropic-apikey": "claude-color.svg",
   "azure-openai": "openai.svg",
   chatgpt: "openai.svg",
   "cloudflare-ai-gateway": "cloudflare-ai-gateway-color.svg",
--- a/src/generated/jawcode-model-metadata.ts
+++ b/src/generated/jawcode-model-metadata.ts
@@ -14,6 +14,8 @@ export interface JawcodeModelMetadata {
 const PROVIDER_ALIASES: Record<string, string> = {
   "xai": "xai",
   "anthropic": "anthropic",
+  "anthropic-apikey": "anthropic",
+  "anthropic-key": "anthropic",
   "kimi": "moonshot",
   "opencode-go": "opencode-go",
   "openrouter": "openrouter",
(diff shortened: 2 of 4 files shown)


lidge-jun:
Thanks @phj1081! Landed on `dev` as 5695d5b via cherry-pick with your authorship preserved (the branch has since been restructured — src/ moved to semantic folders — so a rebase-merge wasn't clean; the cherry-pick applied without conflicts). The anthropic-apikey entry works out of the box on the new tree: parity tests green, 1580 pass / 0 fail. Closing as landed.

PR #67, opened 2026-07-08
fix(cursor): enable multi-account OAuth by extracting JWT identity

Wibias:
## Summary
- Extract stable Cursor account identity from JWT `sub` (and `email` when present) during login and token refresh, so the multiauth store can append distinct accounts instead of always replacing the active slot.
- Fix the Providers UI add-account polling loop to treat a completed OAuth flow (`done: true`) as success, not only when the account count grows — covers same-account re-login.

## Problem
Adding a second Cursor account showed \"Waiting for browser…\" forever even after Cursor returned \"All set!\". OAuth completed (`done: true`) but `accounts.length` stayed at 1 because Cursor credentials had no persisted identity.

## Test plan
- [x] `bun test tests/cursor-oauth.test.ts tests/oauth-store-multi.test.ts`
- [ ] Add a second Cursor account via Web UI → dialog closes, both accounts listed
- [ ] Switch between Cursor accounts in the provider card dropdown
- [ ] Re-login the same Cursor account via \"Add account\" → completes without hanging

### DIFF
--- a/gui/src/pages/Providers.tsx
+++ b/gui/src/pages/Providers.tsx
@@ -13,7 +13,7 @@ interface Config {
   providers: Record<string, { adapter: string; baseUrl: string; hasApiKey?: boolean; hasHeaders?: boolean; defaultModel?: string; authMode?: string; disabled?: boolean }>;
 }
 
-interface OAuthStatus { loggedIn: boolean; email?: string; error?: string }
+interface OAuthStatus { loggedIn: boolean; email?: string; error?: string; done?: boolean }
 interface ProviderQuotaReport { provider: string; quota: AccountQuota; source: string; updatedAt: number }
 interface OAuthAccount { id: string; email?: string; active: boolean; needsReauth?: boolean; expiresAt?: number }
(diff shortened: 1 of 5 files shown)


lidge-jun:
Reviewed against current `dev` (post-260709 catalog refresh). This is a clean, well-scoped fix:

- Extracting a stable identity from the JWT `sub`/`email` in `credentialsFromCursorTokens` is the right way to unlock multiauth append semantics for Cursor — and consolidating `getTokenExpiry` onto the same decoder is a nice cleanup.
- The Providers UI polling fix (`done: true && !error` as an add-account completion signal) correctly covers the same-account re-login case that previously hung on `accounts.length`.
- `git apply --check` passes against `dev` with zero file intersection with our recent commits, and tests are included.

**Decision: accepting.** I'll merge this into `preview` and cherry-pick it onto `dev` so it ships with the current integration branch. Thanks for the thorough test plan!

lidge-jun:
Landed: merged into `preview` and cherry-picked onto `dev` as 46ac27f (authorship preserved). Gates after landing: `tsc --noEmit` clean, full `bun test` 1658 pass / 0 fail. Thanks again!

PR #69, opened 2026-07-08
docs: document where to set unsafeAllowNativeLocalExec for Cursor

Wibias:
## Summary
- Add a dedicated **Cursor provider** section to the configuration reference with a `config.json` example
- Clarify that `unsafeAllowNativeLocalExec` belongs on `providers.cursor`, not at the top level
- Mention the dashboard path (Providers → Cursor → Edit JSON)
- Update the Cursor registry note shown in the Add Provider picker with the same guidance

## Test plan
- [x] `bun test tests/provider-registry-parity.test.ts`
- [ ] Docs render correctly on the docs site
- [ ] Anchor link from providers guide resolves

### DIFF
--- a/docs-site/src/content/docs/guides/providers.md
+++ b/docs-site/src/content/docs/guides/providers.md
@@ -113,12 +113,14 @@ and the dashboard Add Provider picker as an experimental local config entry with
 public model catalog metadata. When a Cursor access token is configured, opencodex uses Cursor's
 live HTTP/2 transport. Cursor server-driven native read/write/delete/ls/grep/shell/fetch execution
 is disabled by default because it bypasses Codex's approval and sandbox path; set
-`unsafeAllowNativeLocalExec: true` only for trusted local experiments. The older
-`allowNativeLocalExec` spelling is accepted as a deprecated transition alias. MCP, screen recording,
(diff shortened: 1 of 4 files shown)


lidge-jun:
Reviewed — this closes a real documentation gap. The `providers.cursor` placement guidance matches the actual config resolution, the new Configuration-reference section with the JSON example is exactly what users have been missing, and updating the registry note (surfaced in the Add Provider picker) keeps the two in sync. The security caution block is appropriately blunt.

Heads-up for landing: `dev` recently refreshed the cursor static catalog, so the `tests/provider-registry-parity.test.ts` hunk no longer applies mechanically (the neighboring model-count assertion changed to `>=38`). The docs and registry hunks apply cleanly with offsets. I'll resolve that test hunk semantically while landing — your three new note assertions will be kept.

**Decision: accepting.** Merging into `preview` and cherry-picking onto `dev` with the parity-test resolution. Thanks!

lidge-jun:
Landed: merged into `preview` and cherry-picked onto `dev` as 2443382 (authorship preserved). The parity-test region merged cleanly — your three note assertions now coexist with the refreshed catalog assertions. Gates after landing: `tsc --noEmit` clean, full `bun test` 1658 pass / 0 fail.

PR #70, opened 2026-07-08
feat(diagnostics): warn when project .codex/config.toml bypasses OpenCodex

Wibias:
## Summary
- Detect repo-local `.codex/config.toml` settings (non-opencodex `model_providers` tables, `profile`, and `model_provider`) that route trusted projects away from the OpenCodex proxy when global proxy routing is active.
- Surface warnings in `ocx doctor`, `ocx start`/`sync` console output, and the dashboard via `GET /api/diagnostics/project-config`.
- Deduplicate redundant profile warnings when they only select an already-flagged provider table.

## Test plan
- [x] `bun test tests/project-config-warnings.test.ts`
- [x] `bun test tests/codex-sync-api.test.ts`
- [ ] `ocx doctor` shows grouped warning for a trusted project with `profile = "opencode_go"` + `[model_providers.opencode_go]`
- [ ] Dashboard at `http://localhost:10100` shows the red bypass banner after proxy restart
- [ ] Removing project-local provider/profile entries clears the warning

### DIFF
--- a/gui/src/i18n/en.ts
+++ b/gui/src/i18n/en.ts
@@ -58,6 +58,8 @@ export const en = {
   "dash.syncOk": "Sync complete. {count} model(s) appended.",
   "dash.syncStaleHint": "If Codex App still shows an older list, restart its long-lived app-server process.",
   "dash.syncFailed": "Sync failed: {error}",
+  "dash.projectConfigTitle": "Project Codex config bypasses OpenCodex",
+  "dash.projectConfigHint": "These repo-local settings override the OpenCodex proxy (e.g. route to OpenCode Go directly). Remove them so ~/.codex/config.toml routing applies in that project.",
   "dash.checkUpdate": "Check update",
   "dash.updateTitle": "Update opencodex",
   "dash.updateDesc": "Check npm for the selected channel, then choose whether to restart the proxy after installation.",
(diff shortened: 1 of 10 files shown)


lidge-jun:
Thanks for adding this diagnostic — the feature fits the current Design B routing model (project-level `.codex/config.toml` genuinely can override the injected global route), and the patch applies cleanly to `dev` too. But I'd like changes before merging.

Main issue: the warning currently over-claims bypasses.

- `src/codex/project-config-warnings.ts:71` flags every non-`opencodex` `[model_providers.*]` table, but a provider table alone routes nothing unless selected via root `model_provider`, a profile, or another precedence rule. Dormant config would produce a red "bypasses OpenCodex" banner.
- `:89` treats any root `profile = "x"` as a bypass without resolving `[profiles.x]` — profiles can set unrelated options.
- `:51` treats any `[model_providers.opencodex]` table as active global routing, though under Design B loopback the marker-owned root `openai_base_url` is what routes.
- `parseTrustedProjectPathsFromCodexConfig` (`:55`, `:182`) collects every `[projects.*]` path without checking `trust_level = "trusted"`.
- The dashboard fetch rides the 5s polling loop (`gui/src/pages/Dashboard.tsx:104`) — please cache the scan or fetch it less frequently.

**Decision: request changes.** Tighten the detector around *effective* routing (resolve selected provider/profile + trust level), add negative tests for dormant tables/profiles/untrusted projects, and reduce the polling I/O. Happy to re-review quickly after that — the direction is genuinely useful.

Wibias:
@lidge-jun Is it okay like that?

lidge-jun:
Cherry-picked into `dev` as [beccd6f](https://github.com/lidge-jun/opencodex/commit/beccd6f) (both commits).

All 5 review feedback items confirmed resolved by contributor. Additional fixes applied during cherry-pick:
- Test isolation: `CODEX_HOME` now set in `project-config-warnings.test.ts` to prevent reading real user config
- Cache test rewritten to use `collectProjectCodexConfigWarnings` with explicit `cwd`/`codexConfigPath` for hermetic test isolation

1794 tests pass, 0 fail. Closing.

PR #89, opened 2026-07-10
fix: relay Codex alpha search requests

SJY051:
## Summary

- relay Codex's client-side `POST /v1/alpha/search` request to the ChatGPT forward provider
- preserve forwarded or pool-selected Codex auth while preventing proxy admission-token leakage
- pass upstream JSON status and body through unchanged, with bounded response buffering, cancellation, and timeout handling
- replace the old intentional-404 contract with dedicated search relay coverage

## Root cause

Loopback Design B points Codex's built-in `openai` provider at the OpenCodex `openai_base_url`. Codex's built-in search client resolves `alpha/search` against that same base URL, so the request reaches OpenCodex as `/v1/alpha/search`.

OpenCodex currently lets that path fall into the generic `/v1/*` JSON 404 guard. This makes built-in web search fail immediately with:

```text
Unknown endpoint: POST /v1/alpha/search
```

The request/response contract is a non-streaming JSON POST in Codex 0.144.0 and current main, so the proxy can relay it verbatim without translating provider formats.

## Security and behavior

- only an enabled ChatGPT `authMode: "forward"` provider can serve this private endpoint
- routed providers and OpenAI API-key providers are not treated as compatible fallbacks
- selected pool-account auth overrides the caller bearer using the existing Codex auth-context helpers
- the proxy admission secret is stripped and never sent upstream
- the route uses the same data-plane auth, Origin, drain, request-log, and CORS gates as `/v1/responses` and `/v1/images/*`
- upstream status and JSON bodies are preserved so Codex can surface actionable backend errors

## Validation

- regression test before the fix: expected `200`, received `404`
- `bun test tests/server-search.test.ts tests/server-auth.test.ts` — 54 passed, 0 failed
- `tsc --noEmit` — passed
- `bun test ./tests/` — 2024 passed, 0 failed, 8539 assertions


lidge-jun:
Reviewed and merged into `dev` as a971442f (squash; base retargeted from `main` to `dev`, the integration branch — it flows to `main` with the next release merge), with one follow-up fix landed as 55e3322d.

Review verdicts (checked against the current dev tree):

- Route placement, drain/auth/origin gating: byte-for-byte parity with the `/v1/images` relay block, including the 499 `client_cancel` request-log path. Good.
- Auth-context reuse: pool-account override, admission-secret strip, cooldown/affinity/reauth error mapping all match the established helpers; `headersForCodexAuthContext` builds from the FORWARD_HEADERS allowlist, so no host/cookie/content-length leakage upstream.
- Buffering (arrayBuffer then 16MB cap) and first-enabled-forward provider selection both match the images-relay precedent.
- Tests: excellent coverage — all 11 pass on dev, and the isolation pattern follows the repo convention.

One defect fixed in the follow-up (55e3322d): the relay reused `config.connectTimeoutMs` as the TOTAL upstream deadline. That knob is documented as the DNS/TCP/TLS/header-arrival budget, and since `alpha/search` is non-streaming (headers arrive only when the search completes), a typical ~10s connect budget would have killed every long-running search. The relay now has a dedicated `config.search.timeoutMs` (default 200s, mirroring `images.timeoutMs`); the 504 test moved to the new knob and a regression test pins that a short `connectTimeoutMs` no longer cuts a slow search.

Really solid, well-tested contribution — the root-cause analysis in the PR body matched what we verified in codex-rs behavior exactly. Thanks @SJY051!

PR #90, opened 2026-07-10
fix(openai-responses): strip hosted image_generation that conflicts with declared image_gen tool

kargnas:
## Symptom

When Codex targets a routed provider model through the proxy against an OpenAI-compatible platform `/v1/responses` endpoint (API-key / `authMode: "key"`), requests that involve the imagegen app skill fail with HTTP 400:

```
Invalid Value: 'tools'. Function 'image_gen.imagegen' conflicts with a hosted tool in the same request.
```

## Root cause

For the imagegen app skill, Codex declares a **client** tool named `image_gen.imagegen` (a `function`, or in some builds an `image_gen` `namespace` tool) **and** also attaches the **hosted** `image_generation` tool in the same request.

- The **ChatGPT backend** (`authMode: "forward"`) tolerates this pair, so native slugs are unaffected.
- Native slugs also run `tool_mode: code_mode_only`, which embeds their tools inside the exec doc rather than sending them as real `tools[]` entries — so no dot-named function tool reaches the wire to collide.
- For **routed** catalog entries, `normalizeRoutedCatalogEntry` (`src/codex/catalog.ts`) deletes `tool_mode`. The dot-named skill function is therefore emitted as a real `tools[]` entry, and a platform `/v1/responses` endpoint rejects the function-name / hosted-tool collision with the 400 above.

## Fix

Add `stripConflictingHostedTools` in `src/adapters/openai-responses.ts`, applied **only on the API-key path** (the `else` branch of `buildRequest`). When a declared client tool matches a known conflict prefix (`image_gen` → hosted `image_generation`), the hosted entry is dropped — the declared tool wins, and Codex executes the skill client-side either way. Both tool shapes are handled: a flat `function` named `image_gen.imagegen` and an `image_gen` `namespace` tool.

**Forward mode is intentionally excluded**: the ChatGPT backend accepts the pair, and stripping the hosted tool there would disable native imagegen. The strip is a no-op (returns the original reference) when nothing conflicts, so the common path stays allocation-free.

This mirrors the existing `UNSUPPORTED_HOSTED_TOOLS` / `stripUnsupportedHostedTools` precedent in the same file (which strips `image_generation` for the `codex-spark` slug), using the same conflict-table + filter shape.

## Test coverage

Added to `tests/openai-responses-passthrough.test.ts`:

- Keyed platform, `[function image_gen.imagegen, hosted image_generation, hosted web_search]` → `image_generation` stripped, the `image_gen.imagegen` function kept, unrelated `web_search` kept.
- Keyed platform, `[namespace image_gen, hosted image_generation]` → hosted `image_generation` stripped, namespace tool kept.
- Keyed platform, hosted `image_generation` with **no** conflicting declared tool → kept (no over-stripping).
- Forward-mode backend, `[function image_gen.imagegen, hosted image_generation]` → pair preserved.

`bun test ./tests/` → 2018 pass / 0 fail. `bun x tsc --noEmit` → clean.


lidge-jun:
Reviewed and merged into `dev` as 8910e548 (squash; base retargeted from `main` to `dev`, which is the integration branch — it will flow to `main` with the next release merge).

Review notes (verified against the current dev tree):

- Root cause confirmed: `normalizeRoutedCatalogEntry` (src/codex/catalog.ts) deletes `tool_mode` for routed entries, so the dot-named `image_gen.imagegen` function reaches the wire and collides with hosted `image_generation` on platform `/v1/responses` — exactly as described.
- Placement is right: the strip sits in the keyed `else` branch mirroring `repairOrphanedInputItems` on the forward branch; `stripUnsupportedHostedTools` and the other stringify-time filters are independent, no ordering hazard.
- No interaction regressions: the `/v1/images` relay resolves its own upstream and never reads `tools[]`, and the vision sidecar handles image input, not tools.
- Edge behavior checked: multiple hosted entries all dropped, non-object tool entries preserved, no-conflict case returns the original reference. One untested-but-correct quirk: a `custom`-typed tool named `image_gen.*` also triggers the strip, which matches the platform's conflict semantics.
- Full suite on the merged tree: 2062 pass / 0 fail, `tsc --noEmit` clean.

Nice, surgical fix — the conflict-table shape mirroring `UNSUPPORTED_HOSTED_TOOLS` made this easy to verify. Thanks @kargnas!

PR #94, opened 2026-07-11
fix: preserve routed sub-agent task payloads by sanitizing before parse

Wibias:
## Summary

This fixes a routed multi-agent regression where spawned sub-agents woke up with the environment/context but **without the actual task payload**.

On the wire, the task body can arrive parked as plaintext inside an `encrypted_content` slot inside an internal `agent_message`. Two transformations are required before provider translation:

1. rewrite the plaintext encrypted slot to `input_text`
2. normalize the now-plaintext `agent_message` to a standard user `message` that `parseRequest` understands

Previously the sanitizer ran only on the native-bound path and only after parsing. Moving it before `parseRequest` exposed a second gap: the routed parser intentionally ignores unknown input item types, so merely making the content readable still left the entire `agent_message` undispatched. The sanitizer now performs both operations when the payload is entirely plaintext.

## What changed

- `src/server/responses.ts`
  - run `sanitizeEncryptedContentInPlace` on `body.input` immediately after `expandPreviousResponseInput(body)` and before `parseRequest(body)`
  - normalize a rewritten `agent_message` to `{ type: "message", role: "user" }` only when no genuine encrypted content remains
  - preserve genuine backend ciphertext and mixed plaintext/Fernet payloads without coercing their container
  - remove the old native-only post-parse sanitize block

- `tests/multi-agent-compat.test.ts`
  - mirror the real request order: sanitize/normalize raw input, then parse
  - assert both the container normalization and delivery of the complete `NEW_TASK` payload in parsed user messages

## Root cause

`parseRequest` tolerates unknown input items at the schema boundary but only emits context messages for recognized item types. After the first version of this patch, the encrypted slot became readable `input_text`, yet its parent still had `type: "agent_message"`. The parser therefore skipped it, producing one user message instead of two and dropping the spawned task.

This is why the regression test failed in CI with:

- expected user-message count: `2`
- received user-message count: `1`

The follow-up commit fixes that actual boundary rather than weakening the assertion.

## Safety properties

- Plaintext spawn payloads become ordinary user input before routed translation.
- Genuine ciphertext remains byte-identical.
- Mixed payloads retaining a Fernet token are not coerced into a user message.
- The implementation remains confined to the request normalization layer.

## Scope

This PR intentionally contains only:

- `src/server/responses.ts`
- `tests/multi-agent-compat.test.ts`

## Verification

- `bun test tests/multi-agent-compat.test.ts --test-name-pattern "sanitize-then-parse delivers"` - pass
- `bun test tests/multi-agent-compat.test.ts` - 27 pass, 0 fail
- `bun run typecheck` - pass
- `git diff --check` - pass
- full `bun test` - the spawn regression passes; the local run remains red on five unrelated pre-existing environment/integration tests (Cursor MCP live stdio x3, server pool-health, Windows service task)

The upstream cross-platform CI rerun was triggered by the follow-up push.

### DIFF
--- a/src/server/responses.ts
+++ b/src/server/responses.ts
@@ -298,10 +298,19 @@ function encryptedSlotParts(payload: string): Array<Record<string, string>> {
   return parts.length > 0 ? parts : [{ type: "input_text", text: payload }];
 }
 
+function hasEncryptedContentPart(content: unknown): boolean {
+  return Array.isArray(content) && content.some(part => (
+    part && typeof part === "object"
+    && (part as { type?: unknown }).type === "encrypted_content"
+  ));
+}
+
 /**
  * Rewrite non-ciphertext `{type:"encrypted_content"}` parts into `{type:"input_text"}`
- * throughout a native-bound request's input items (message content and
- * function_call_output content arrays share the part shape, codex-rs protocol/models.rs).
+ * throughout a request's input items (message content and function_call_output content
+ * arrays share the part shape, codex-rs protocol/models.rs). An `agent_message` whose
+ * payload becomes entirely plaintext is normalized to a user message so routed parsers
+ * that do not understand the internal item type still deliver the spawn task.
  * Genuine backend blobs are left byte-identical so replay/cache semantics survive, and
  * MIXED slots (plaintext preamble + embedded Fernet task body) are split so the backend
  * decrypts the blob while the prose passes as text. Returns the number of parts rewritten.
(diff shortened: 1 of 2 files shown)


lidge-jun:
Thanks! The unique delta (sanitizer normalization, 749d3978) was cherry-picked onto dev as 6131577d on top of #93's commits (f29002de was patch-identical to #93's 64da27d3, so it was skipped). Shipping in the next release.

PR #99, opened 2026-07-12
refactor: satisfy GUI lint rules

Wibias:
> **Part 4 of 7 — opencodex hardening stack.** Base: `main`. Stacks on Parts 1-3 — merge those first. Focus commit: `e8794d2`.

## Summary
Makes the GUI pass `eslint` under the pinned toolchain and turns on the GUI lint + build CI gate in the same PR. No user-facing behavior change.

## Background / motivation
`eslint-plugin-react-hooks@7` enforces `react-hooks/set-state-in-effect`, which the existing GUI pages violated (data-loading effects calling `setState` synchronously), and `gui/src/i18n/index.tsx` exported both a component/provider and helpers, tripping `react-refresh/only-export-components`. Enabling a GUI lint gate without these fixes would fail CI, so the gate is introduced here, together with the fixes.

## What changed
**GUI lint fixes**
- Data-loading effects across `App.tsx` and `pages/*` (`ApiKeys`, `CodexAuth`, `Dashboard`, `Debug`, `Models`, `Providers`, `Subagents`, `Usage`) now defer the initial load (`setTimeout(..., 0)`) and wrap loaders in `useCallback`, satisfying `react-hooks/set-state-in-effect`.
- Split `gui/src/i18n/index.tsx` into `provider.tsx` (component/context) and `shared.ts` (helpers), with `index.ts` re-exporting — clears `react-refresh/only-export-components`.

**CI gate** — `.github/workflows/ci.yml`
- Adds the `GUI lint` (`bun install --frozen-lockfile && bun run lint`) and `GUI build` steps. Placed here (not in Part 2) so the gate lands with the fixes and every PR in the stack stays green.

## Usage
None. Pure refactor + CI.

## Risk
- Blast radius: low, GUI-only.
- Mitigation: `eslint .` clean and `tsc --noEmit` clean; render behavior unchanged.

## Testing
`bun run lint` (gui) clean under a frozen install; typecheck clean.


lidge-jun:
Merged with one review-fix commit: 77efcaf7 restores request invalidation in Usage.tsx via AbortController (stale /api/usage responses could overwrite newer range state after the cancelled-guard was dropped). The Debug.tsx conflict with #103 was resolved to main's 3-stream version, which subsumes this PR's lint intents (verified byte-identical).

Issue #114, opened 2026-07-12
OpenRouter (openai-chat adapter): streaming delivers tokens in batches, not live


RobinBially:
## Summary

Streaming works correctly with ChatGPT passthrough (`openai-responses` adapter) but delivers tokens in batches (only at the end) with OpenRouter models using the `openai-chat` adapter.

## Environment

- opencodex 2.7.8
- macOS (Apple Silicon)
- Codex Desktop (latest)
- OpenRouter API

## Steps to reproduce

1. Install opencodex with default config (ChatGPT passthrough + OpenRouter with `openai-chat` adapter)
2. Use a ChatGPT model (e.g. `gpt-5.6-luna`): streaming works token-by-token ✅
3. Switch to an OpenRouter model (e.g. `openrouter/deepseek/deepseek-v4-flash`): response appears all at once at the end ❌
4. Switch to a ChatGPT model routed through OpenRouter (e.g. `openrouter/gpt-5.6-luna`): also delayed ❌

## What we tried

- **`websockets: false`** in config: no change, streaming still delayed
- **Direct `openai-responses` adapter to OpenRouter Responses API** (`https://openrouter.ai/api/v1/responses`): OpenRouter returned errors, likely because Codex sends proprietary `namespace` tool types (see [openai/codex#32318](https://github.com/openai/codex/issues/32318))

## Debug logs show correct SSE deltas

`ocx debug provider on` shows that opencodex produces correct token-by-token `response.output_text.delta` SSE frames:

```
delta: "J"
delta: "etzt"
delta: " sie"
delta: "hst"
...
```

So the bridge/translation layer produces correct SSE deltas — the chunks appear to be buffered at the OpenRouter level or between opencodex and the client.

## Suspected cause

OpenRouter SSE streaming in the `openai-chat` adapter path delivers chunks in larger batches rather than individual tokens, or there is buffering in the async generator → SSE stream pipeline. The fact that ChatGPT passthrough (direct `openai-responses` adapter) streams perfectly suggests this is specific to the `openai-chat` adapter path with OpenRouter.

Wibias:
@RobinBially Should hopefully get fixed by https://github.com/lidge-jun/opencodex/pull/115 either u build urself with the change or wait for a new version!
- Wibias

RobinBially:
Thanks for the fix in #115. I retested this end-to-end with the current release, and unfortunately the real Codex Desktop path is still batching the final answer.

Environment:

- opencodex 2.7.13
[...]

PR #123, opened 2026-07-13
fix(streaming): prevent bursty Responses SSE delivery

Wibias:
## Problem

Issue #114 remained partially unresolved after #115. The upstream `Accept-Encoding: identity` change addressed compressed provider responses, but the Codex Desktop app-server could still receive the final answer as one burst when the bridge processed many adapter events in a single JavaScript turn.

The observable symptom was:

- direct curl requests streamed progressively;
- the real Codex app-server received hundreds of `item/agentMessage/delta` notifications within roughly 30 ms;
- the proxy's SSE frames were semantically correct, but the HTTP layer could coalesce same-turn enqueues before flushing them downstream.

## Root cause

`bridgeToResponsesSSE` consumes an async adapter event burst inside the `ReadableStream.start()` task. When the upstream response body supplied many parsed text deltas without an event-loop gap, the bridge called `controller.enqueue()` repeatedly in the same turn. Bun then exposed the queued frames to the downstream client as one large delivery.

This explains why the previous compression fix could pass a simple curl reproduction while the full Codex request shape still appeared buffered.

## Fix

- Track whether the bridge emitted output during the current turn.
- Detect when the next adapter event arrives before the scheduled macrotask boundary.
- Yield once with a zero-delay timer so Bun can flush queued SSE bytes before processing the next same-turn event.
- Preserve normal provider-paced streams: when the upstream already waits between events, the scheduled timer overlaps that wait and adds no intentional fixed pacing.

## Regression coverage

`tests/bridge-live-delivery.test.ts` starts a real Bun HTTP server, feeds the bridge 40 synchronous `text_delta` events, reads the response through `fetch()`, and asserts that deltas arrive across multiple reads rather than one end burst.

## Verification

- Focused bridge, SSE, header-policy, and openai-chat suites: 83 passed.
- `git diff --check`: passed.
- Full suite rerun: 2,398 passed, 3 skipped, 1 unrelated failure.
- Remaining unrelated failure: Windows cannot create the symlink used by `tests/claude-agents-inject.test.ts:92` (`EPERM`) without symlink privileges.

## Risk and rollout

The change only affects bursty streaming delivery and keeps event ordering, terminal handling, cancellation, and heartbeat behavior unchanged. It may increase the number of downstream flush opportunities for providers that deliver many deltas in one response-body read, which is the intended behavior for interactive streaming.

Please retest with the real macOS Codex Desktop app-server path from issue #114.

### DIFF
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -146,11 +146,13 @@ export function bridgeToResponsesSSE(
 
   return new ReadableStream<Uint8Array>({
     async start(controller) {
+      let emittedSinceYield = false;
       const emit = (name: string, data: Record<string, unknown>) => {
         if (closed) return;
         activity = true;
         try {
           controller.enqueue(encoder.encode(sseEvent(name, { type: name, sequence_number: seq++, ...data })));
+          emittedSinceYield = true;
         } catch {
           closed = true;
         }
@@ -399,9 +401,19 @@ export function bridgeToResponsesSSE(
       // we synthesize response.completed below, so Codex never hits the parser's
       // "stream closed before response.completed" (responses.rs) -> ApiError::Stream.
       let terminated = false;
+      let macrotaskFired = true;
+      let macrotaskTimer: ReturnType<typeof setTimeout> | undefined;
 
       try {
         for await (const event of events) {
+          if (!macrotaskFired && emittedSinceYield) {
+            await new Promise<void>(r => setTimeout(r, 0));
+            macrotaskFired = true;
+          }
+          emittedSinceYield = false;
+          macrotaskFired = false;
+          if (macrotaskTimer !== undefined) clearTimeout(macrotaskTimer);
+          macrotaskTimer = setTimeout(() => { macrotaskFired = true; macrotaskTimer = undefined; }, 0);
           activity = true;
           stallTicks = 0;
           // Compaction turns emit ONLY the synthetic compaction item + response.completed. The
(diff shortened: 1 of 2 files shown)


lidge-jun:
Reviewed; the approach looks sound. Macrotask-boundary yielding in bridgeToResponsesSSE splits synchronous adapter bursts without reordering, benefits all routed providers, and leaves native Responses passthrough unaffected.

Before this leaves draft, we would like:

1. The real macOS Codex Desktop app-server end-to-end retest referenced in #114.
2. A burst-duration measurement for large buffered replays. setTimeout(0) per emitted event could stretch replay of hundreds of queued events.
3. A stronger regression test. tests/bridge-live-delivery.test.ts currently only asserts that deltas span more than one read; please assert multiple delivery groups or bound the largest group.

Wibias:
Addressed the measurable/regression-test requests in commit `ca8426c`.

- `tests/bridge-live-delivery.test.ts` now records delivery groups and asserts the largest group is no more than 75% of the 40-event burst, rather than only checking that there are two reads.
- Added a 200-event synchronous replay timing test with a 5-second upper bound. On the current Windows host, the test completed in approximately 317 ms wall time.
- The focused bridge/provider suite is green: 84 passed, 0 failed; typecheck and `git diff --check` also pass.

The requested real macOS Codex Desktop app-server retest remains environment-blocked here because this runner is Windows and has no macOS app-server binary. The PR remains draft pending that owner-side macOS validation against the production Codex Desktop path described in #114.

PR #124, opened 2026-07-13
fix(xai): normalize nested tool schemas

### LINKED ISSUE #122: [Bug]: xAI/Grok 400 after tool_search loads automation_update's nested oneOf schema
### Summary

`xai/grok-4.5` works normally until Codex `tool_search` loads the Codex app `automation_update` tool. The next provider request returns HTTP 400, and every later turn in the same thread continues to return 400.

The failure is caused by the loaded tool's parameter schema. Its root is a `oneOf`, but some root branches are themselves nested `oneOf` unions instead of direct object schemas. xAI rejects that grammar:

```text
Provider error 400: {"code":"invalid-argument","error":"codex_app__automation_update: tool parameter root must be an object type (root schema is an anyOf/oneOf union with a non-object branch)"}
```

This was isolated from authentication, model selection, reasoning effort, and general multi-turn/tool behavior:

- Fresh Grok requests return 200.
- Normal multi-turn requests return 200.
- Parallel `tool_search` plus a regular function call returns 200 when the loaded schemas have object roots.
- Sending the captured `automation_update` schema by itself reproduces the 400.
- Keeping the same tool name but replacing `parameters` with a simple root object schema returns 200.
- Other configured providers handle the same Codex workflow successfully.

### Reproduction

1. Configure xAI OAuth with the `openai-chat` adapter and select `xai/grok-4.5` in Codex Desktop.
2. Start a thread and ask for an automation-related operation so the model calls `tool_search` and loads `codex_app.automation_update`.
3. Let the returned tool output continue back to the model.
4. Observe that the continuation returns HTTP 400.
5. Send any ordinary follow-up message in the same thread. It returns the same 400 because the historical `tool_search_output` is parsed again and the incompatible tool schema is re-injected.

The relevant minimized schema shape is:

```json
{
  "oneOf": [
    {
      "type": "object",
      "properties": {
        "mode": { "type": "string", "enum": ["view"] }
      }
    },
    {
      "oneOf": [
        { "type": "object", "properties": {} },
        { "type": "object", "properties": {} }
      ]
    }
  ]
}
```

xAI's [Function Calling documentation](https://docs.x.ai/developers/tools/function-calling#parameter-schema) says a root `oneOf`/`anyOf` is accepted only when every branch is directly an object schema. Here the second root branch is another union, so xAI treats it as a non-object branch and rejects the complete request.

#### Root cause in opencodex

`src/responses/parser.ts` re-loads tool specs from every historical `tool_search_output` and merges them into the active tool list (around lines 427-489).

`src/adapters/openai-chat.ts` then forwards each tool's `t.parameters` unchanged (around lines 137-145):

```ts
return tools.map(t => ({
  type: "function",
  function: {
    name: namespacedToolName(t.namespace, t.name),
    description: t.description,
    parameters: t.parameters,
    ...(t.strict !== undefined ? { strict: t.strict } : {}),
  },
}));
```

That combination makes the failure persistent for the thread: the invalid loaded schema is replayed on every subsequent request.

#### Suggested fix

Normalize tool parameter schemas for xAI (or strict `openai-chat` providers) before forwarding them. For this schema, recursively flatten nested root `oneOf`/`anyOf` branches so every root union member directly resolves to an object schema while preserving `$defs` and the individual object variants.

There is already provider-specific precedent in `src/adapters/anthropic.ts` (`normalizeAnthropicInputSchema`) and `src/adapters/kiro-tools.ts` (`ensureRootObjectType`). A shared helper would avoid another divergent implementation, but it should recurse through this nested-union shape rather than shallowly dropping the inner variants.

If a schema cannot be normalized safely, omitting that deferred tool with a clear diagnostic would be preferable to poisoning every later turn in the thread.

Suggested regression test:

1. Parse a Responses request whose history contains a `tool_search_output` with a namespace tool using the nested-root-union schema above.
2. Build the xAI `openai-chat` request.
3. Assert that every root union branch in the emitted function parameters is an object (or resolves directly to one).
4. Verify the loaded tool remains callable and later turns do not fail.

### Logs and screenshots

```shell
# Original captured automation_update schema
HTTP 400
Provider error 400: {"code":"invalid-argument","error":"codex_app__automation_update: tool parameter root must be an object type (root schema is an anyOf/oneOf union with a non-object branch)"}

# Same model and tool name, parameters replaced with a simple root object schema
HTTP 200
```

### Area

Provider adapter

### Version

Reproduced on `@bitkyc08/opencodex@2.7.13`. Current repository `main` / release commit `2.7.16` still has the unchanged forwarding path.

### OS

Windows 11, Codex Desktop

### Config shape

```json
{
  "providers": {
    "xai": {
      "adapter": "openai-chat",
      "baseUrl": "https://api.x.ai/v1",
      "authMode": "oauth",
      "defaultModel": "grok-4.5"
    }
  }
}
```

### Checks

- [x] I searched existing issues and docs first.
- [x] I removed secrets, tokens, and personal data from logs and screenshots.

Wibias:
## Summary

Fixes #122.

xAI/Grok rejects some function-tool parameter schemas after Codex `tool_search` loads a deferred app or MCP tool. The failing shape is a root `oneOf`/`anyOf` whose branches include another root union. xAI requires every root union branch to resolve directly to an object schema.

## Root cause

Responses parsing correctly replays tool definitions from historical `tool_search_output` items into the active tool list. The `openai-chat` adapter previously forwarded every `t.parameters` schema unchanged. Once the incompatible schema entered the thread history, every later request resent it and received HTTP 400.

## Changes

- Scope normalization to xAI transports (`api.x.ai` and the Grok CLI proxy) so other OpenAI-compatible providers retain their original schemas.
- Recursively expand nested root `oneOf`/`anyOf` object branches into direct object branches.
- Preserve root metadata such as `$defs` while rebuilding the normalized union.
- Reject and omit a tool when its root schema contains a non-object branch that cannot be normalized safely, avoiding a poisoned request.
- Add regression coverage for direct adapter requests and the real `tool_search_output` history replay path.

## Verification

- `bun test --isolate ./tests/xai-transport.test.ts ./tests/adapter-usage.test.ts ./tests/responses-parser.test.ts`
- Result: 54 passed, 0 failed.
- `bun x tsc --noEmit`
- `git diff --check`

## Risk and compatibility

This changes only tool schemas sent to xAI-compatible endpoints. Providers using other base URLs are unchanged. The normalization intentionally does not rewrite `allOf`, and unsafe non-object union branches are omitted instead of being silently coerced. A remaining limitation is that a tool omitted for an unsafe schema will not be callable for that request; this is preferable to making the entire conversation fail repeatedly with HTTP 400.

## Maintainer review focus

Please verify the xAI schema contract for nested root unions and, if possible, retest with Codex Desktop using `tool_search` followed by an automation-related operation on Grok.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -125,7 +125,49 @@ function safeToolName(name: string | undefined): string {
   return sanitized;
 }
 
-function toolsToChatFormat(parsed: OcxParsedRequest): unknown[] | undefined {
+const XAI_SCHEMA_BASE_URLS = new Set(["api.x.ai", "cli-chat-proxy.grok.com"]);
+
+function isXaiSchemaTarget(provider: OcxProviderConfig): boolean {
+  try {
+    return XAI_SCHEMA_BASE_URLS.has(new URL(provider.baseUrl).hostname);
+  } catch {
+    return false;
+  }
+}
+
+function expandXaiRootObjectSchemas(schema: unknown): Record<string, unknown>[] | undefined {
+  if (!schema || typeof schema !== "object" || Array.isArray(schema)) return undefined;
+  const obj = schema as Record<string, unknown>;
+  const compositionKey = ["oneOf", "anyOf"].find(key => Array.isArray(obj[key]));
+  if (!compositionKey) {
+    if (obj.type !== undefined && obj.type !== "object") return undefined;
+    return [{ ...obj, type: "object" }];
+  }
+
+  const siblings = Object.fromEntries(Object.entries(obj).filter(([key]) => key !== compositionKey));
+  const branches = obj[compositionKey];
+  if (!Array.isArray(branches)) return undefined;
+  const expanded: Record<string, unknown>[] = [];
+  for (const branch of branches) {
+    const variants = expandXaiRootObjectSchemas(branch);
+    if (!variants) return undefined;
+    for (const variant of variants) expanded.push({ ...siblings, ...variant });
+  }
+  return expanded.length > 0 ? expanded : undefined;
+}
+
+function normalizeXaiToolParameters(parameters: unknown): Record<string, unknown> | undefined {
+  const variants = expandXaiRootObjectSchemas(parameters);
+  if (!variants) return undefined;
+  if (variants.length === 1) return variants[0];
+  const root = parameters && typeof parameters === "object" && !Array.isArray(parameters)
+    ? parameters as Record<string, unknown>
+    : {};
+  const metadata = Object.fromEntries(Object.entries(root).filter(([key]) => key !== "oneOf" && key !== "anyOf" && key !== "type"));
+  return { ...metadata, type: "object", oneOf: variants };
+}
+
+function toolsToChatFormat(parsed: OcxParsedRequest, provider: OcxProviderConfig): unknown[] | undefined {
   if (!parsed.context.tools || parsed.context.tools.length === 0) return undefined;
   const allowed = isAllowedToolChoice(parsed.options.toolChoice)
     ? new Set(parsed.options.toolChoice.allowedTools)
@@ -134,15 +176,21 @@ function toolsToChatFormat(parsed: OcxParsedRequest): unknown[] | undefined {
     ? parsed.context.tools.filter(t => toolAllowedByChoice(t, allowed))
     : parsed.context.tools;
   if (tools.length === 0) return undefined;
-  return tools.map(t => ({
+  const xaiTarget = isXaiSchemaTarget(provider);
+  const formatted = tools.flatMap(t => {
+    const parameters = xaiTarget ? normalizeXaiToolParameters(t.parameters) : t.parameters;
+    if (parameters === undefined) return [];
+    return [{
     type: "function",
     function: {
       name: namespacedToolName(t.namespace, t.name),
       description: t.description,
-      parameters: t.parameters,
+      parameters,
       ...(t.strict !== undefined ? { strict: t.strict } : {}),
     },
-  }));
+    }];
+  });
+  return formatted.length > 0 ? formatted : undefined;
 }
 
 function toolChoiceToChatFormat(tc: OcxParsedRequest["options"]["toolChoice"], tools: OcxParsedRequest["context"]["tools"]): unknown {
@@ -190,7 +238,7 @@ export function createOpenAIChatAdapter(provider: OcxProviderConfig): ProviderAd
       }
 
       const messages = messagesToChatFormat(parsed, provider);
-      const tools = toolsToChatFormat(parsed);
+      const tools = toolsToChatFormat(parsed, provider);
       const toolChoice = toolChoiceToChatFormat(parsed.options.toolChoice, parsed.context.tools);
 
       const body: Record<string, unknown> = {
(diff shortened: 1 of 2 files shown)


webmastertorch:
Tested locally per your request (Windows, on top of the installed 2.7.16-equivalent source, verified against the live xAI endpoint through the running proxy, which forwards `parameters` verbatim — so the probes exercise exactly what this PR would send).

**What works**

- `bun test tests/xai-transport.test.ts` on the PR branch: 14/14 pass.
- The recursive flattening is correct on the real captured `automation_update` schema: 6 flat object branches, no nested unions left, `$defs` preserved at the root.
- Wire acceptance: posting the PR's normalized output to `grok-4.5` returns **HTTP 200** — the original `tool parameter root must be an object type` 400 from #122 is gone.
- The unsafe-root drop path and the `tool_search` history replay path both behave as intended (`buildTools` already defaults `parameters` to `{}`, so ordinary and freeform tools are never dropped).

**One functional regression: the normalized union tool can no longer emit arguments**

`normalizeXaiToolParameters` emits `{ ...metadata, type: "object", oneOf: variants }`. With `type: "object"` sitting next to the root `oneOf`, xAI accepts the request but compiles the argument grammar as a plain empty object (root `properties` is absent and `additionalProperties` defaults to `false` in xAI's strict tool grammar), so the union branches never contribute properties.

Forced-call probe against `grok-4.5` (`tool_choice: {type:"function", name:"codex_app__automation_update"}`, prompt: "view the automation with id auto-123"), reproduced twice:

| root schema shape | status | generated arguments |
| --- | --- | --- |
| PR output: `{ $defs, type: "object", oneOf: [6 object branches] }` | 200 | `{}` (empty, twice) |
| Same minus root `type`: `{ $defs, oneOf: [6 object branches] }` | 200 | `{"id":"auto-123","mode":"view"}` |
| Simple object control | 200 | `{"id":"auto-123","mode":"view"}` |

This matches the xAI docs: their accepted root-union example is a bare `oneOf` with object branches and no root `type` ([Function Calling → Parameter Schema](https://docs.x.ai/developers/tools/function-calling#parameter-schema)).

**Suggested change**

In the multi-variant branch of `normalizeXaiToolParameters`, drop the root `type`:

```ts
return { ...metadata, oneOf: variants };
```

and flip the test assertion accordingly (`expect(xaiParameters.type).toBeUndefined()` while keeping the all-branches-are-objects assertion). The single-variant and plain-object paths are fine as they are.

With that one-line change this looks like the right fix for #122 to me.

Wibias:
Addressed review feedback in commit 7e449d3. Multi-variant xAI unions now omit the root 	ype wrapper, preserving xAI's argument grammar so forced tool calls can emit branch properties instead of {}. Updated regression assertion; focused suite remains 54 passed, with typecheck and diff check clean.

lidge-jun:
Reviewed and merged. The recursive root oneOf/anyOf flattening, scoped to api.x.ai and cli-chat-proxy.grok.com, resolves #122, including the historical tool_search_output replay path where src/responses/parser.ts rebuilds historical tools each turn.

One follow-up remains: when normalization omits an unsafe tool, an explicit tool_choice naming that tool can still be forwarded. The tool list is normalized in src/adapters/openai-chat.ts around lines 180-193, while tool_choice is computed from the original list around lines 196-253. That can produce the same 400 response. A follow-up should reconcile tool_choice with the normalized tool list.

Thanks for the fix.

PR #125, opened 2026-07-14
fix: use PROD_API for Antigravity onboarding

Jinshijiming:
Here is the English version of the PR description:

***

### **Title**
fix: use PROD_API instead of DAILY_API for Antigravity onboarding

### **Description**

**Background**
When using this project to reverse proxy Antigravity (Cloud Code Assist), new accounts that haven't been onboarded encounter the following error during login, preventing them from using the service:
> `google-antigravity 登录错误：Antigravity login could not discover a Cloud Code Assist project for this account. Ensure the account has Antigravity/Cloud Code Assist access and try again.`

Meanwhile, existing accounts that have already been onboarded work perfectly fine.

**Root Cause Analysis**
When an account logs in, the system first calls `loadCodeAssistProject` to fetch the Project ID. This call hits the production environment (`PROD_API`). For new accounts, this request naturally returns empty, and the code correctly falls back to the `onboardProject` (onboarding) logic.

However, in the `onboardProject` function, the API request was incorrectly directed to the testing environment `DAILY_API` (`https://daily-cloudcode-pa.googleapis.com`) instead of `PROD_API`. This caused the onboarding process for new accounts to fail or mismatch environments, making it impossible to successfully discover or create the required production project, which ultimately threw the exception mentioned above.

**Changes**
- In the `onboardProject` method of `src/oauth/google-antigravity.ts`, corrected the called endpoint from `DAILY_API` to `PROD_API`.
- Removed the unused `DAILY_API` constant from the codebase.

**Testing**
- [x] New accounts that have not initialized Cloud Code Assist can now properly go through the onboarding flow and log in successfully.
- [x] Existing accounts with an established project continue to work as expected without any impact.

### DIFF
--- a/src/oauth/google-antigravity.ts
+++ b/src/oauth/google-antigravity.ts
@@ -21,7 +21,6 @@ const CLIENT_SECRET = process.env.GOOGLE_ANTIGRAVITY_CLIENT_SECRET
 const AUTH_ENDPOINT = "https://accounts.google.com/o/oauth2/v2/auth";
 const TOKEN_ENDPOINT = "https://oauth2.googleapis.com/token";
 const PROD_API = "https://cloudcode-pa.googleapis.com";
-const DAILY_API = "https://daily-cloudcode-pa.googleapis.com";
 const API_VERSION = "v1internal";
 const SCOPES = [
   "https://www.googleapis.com/auth/cloud-platform",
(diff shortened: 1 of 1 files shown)


lidge-jun:
Thanks for the contribution. We reviewed this, and the change is plausible—keeping project discovery and onboarding in the same production control plane makes sense. However, we're holding off on merging for now:

1. The upstream CLIProxyAPI implementation this module mirrors still deliberately uses DAILY_API for onboardUser as of today, so diverging needs justification.
2. The existing tests (tests/google-antigravity-oauth.test.ts) don't assert the onboarding hostname, so green CI doesn't validate this change.
3. Could you provide before/after logs from a genuinely uninitialized account showing discovery failure on DAILY_API and success on PROD_API?

Once we have that evidence—and ideally a test asserting the intended onboarding URL—we're happy to merge. The other DAILY_API uses (runtime requests and quota/model discovery) should remain unchanged; they're the data plane, not the control plane.

PR #130, opened 2026-07-14
tooling: add pre-push hook matching the CI gate

Wibias:
﻿## What

Adds a one-time setup command (`bun run setup:hooks`) that installs a `.git/hooks/pre-push` hook running the same two checks as the CI matrix:

- `bun run typecheck` -- tsc --noEmit
- `bun run test` -- full unit suite

Also fixes a test failing on Windows CI runners: the symlink-ownership test in `tests/claude-agents-inject.test.ts` now catches `EPERM` and skips gracefully when the runner lacks Developer Mode / SeCreateSymbolicLinkPrivilege. The underlying implementation is correct -- the guard was only needed in the test setup.

## Files

- `scripts/pre-push.sh` -- POSIX sh hook shim, works on Windows/macOS/Linux
- `scripts/setup-hooks.ts` -- cross-platform installer
- `package.json` -- adds `prepush` and `setup:hooks` script entries
- `CONTRIBUTING.md` -- documents the one-liner for new contributors
- `tests/claude-agents-inject.test.ts` -- EPERM-safe symlink test

## Why

Catches type errors and test failures locally before they hit the remote CI runners. The `ctx?.signal` vs `ctx?.abortSignal` mistake on the MiMo Free PR would have been caught at push time instead of in the Actions log.

Skip in an emergency: `git push --no-verify`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,3 +9,15 @@ Thanks for helping with opencodex.
 
 For local development commands, architecture notes, and release workflow details, use the hosted
 contributing guide above instead of duplicating instructions here.
+
+## Pre-push hook
+
+After cloning, run once to install a local pre-push hook that mirrors the CI gate:
+
+```sh
+bun run setup:hooks
+```
+
+This installs `.git/hooks/pre-push`, which runs `bun run typecheck && bun run test` before every
+`git push`. The same two checks run on ubuntu-latest, macos-latest, and windows-latest in CI.
+Skip in an emergency with `git push --no-verify`.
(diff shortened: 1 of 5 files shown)


lidge-jun:
Merged into dev and released in v2.7.19 — thanks! Stacked a few hardening commits on top (cac11925): hooks dir now resolved via `git rev-parse --git-path hooks` (worktree/core.hooksPath safe), existing differing hooks are backed up instead of overwritten, the shim delegates to a single `prepush` script (now also running `privacy:scan`), and the Windows EPERM path reports a visible test skip.

Issue #133, opened 2026-07-15
[Feature]: Multiple providers using the same model, create a combo to handle load balancing.


guigeng:
### Problem to solve

Multiple providers offering the same model can create aliases or combos that point to multiple backend providers.

### Proposed solution

--

Wibias:
Yea, i know what ur talking about. I am doing two other things atm at this project, but thats something I also wanted to implement very soon.

Wibias:
@guigeng Since I already had a concept for this in mind, can you please also give me some type of concept for what u wish for?

guigeng:
Create a group, for example, named "free". When creating the group, you need to specify the backend model, such as Claude, GPT, or Deepseek. Client requests are sent to the model named "free" within this group. When the OCX receives the request, it forwards it to the backend model using load balancing strategies such as round-robin or weighted selection.

Wibias:
@guigeng Alright. Will be the next one after i finish another little branch. Any Free Tier Providers u wish for to be implemented next?

guigeng:
APIs for adding custom anthropic messages and responses are needed.

Wibias:
@guigeng Alright. Please create a separate Issue for that so I can keep track of pr to issue.

lidge-jun:
Status: slice 1 of the combo stack is on `dev` (validated CRUD + deterministic weighted round-robin/sticky selection + fail-closed member handling, commits f48a5c73…68cda367, based on @Wibias's PR #147 with authorship preserved). Failover engine and catalog/usage slices are in progress; this issue stays open until the full stack lands.

lidge-jun:
The full combo stack has landed on `dev` — all three slices complete:

- **Slice 1** (domain): validated CRUD, reserved `combo/` namespace with backward compat, deterministic smooth-weighted round-robin + sticky semantics, fail-closed member handling (f48a5c73…68cda367, based on @Wibias's PR #147).
- **Slice 2** (failover): per-attempt isolation through the full response pipeline — connection-exception failover, cross-adapter hops, per-target OAuth isolation (no token leakage), committed-stream no-replay, client-abort 499 first-wins, Retry-After-aware cooldowns (25d61f38…3174b78d).
- **Slice 3** (catalog/usage): member-intersection catalog capabilities (no fabricated defaults, stale-row lifecycle) and ordered per-attempt usage attribution with cancellation receipts (88be06fd…161a9f74).

[...]

PR #139, opened 2026-07-16
feat(gui): providers workspace redesign polish + freeTier + rate limits

Wibias:
## Summary

Providers workspace redesign polish: a single place to manage providers, accounts, rate limits, models, and JSON config — without the classic card UI.

- Overview with Ready / Needs setup / Disabled counts, rate-limit bars, and recently used
- Provider detail with Overview / Models / Usage & limits / Settings, plus embedded ChatGPT account pool
- Free vs paid pricing (`freeTier`) separate from keyless auth (`keyOptional`) — e.g. NVIDIA NIM is Free but still needs a key; MiMo Free is keyless
- Multi-select rail filters, clearer copy (Reset filters, Checked … ago, overwrite warnings, auto-switch help)
- Edit JSON as an in-pane editor with unsaved-leave prompts
- Registry DTO backfill so older configs get correct free/keyless flags in the GUI

## Screenshots

Captured with mocked `/api` responses (dummy providers, quotas, usage — no real account data).

01-overview
02-filters
03-chatgpt-detail
04-nvidia-free-tier
05-add-provider-free

## Test plan

- [ ] GUI on Vite `:5173`: overview, rate limits, filters, provider detail tabs, Edit JSON unsaved leave
- [ ] Remove provider uses in-app confirm (not `window.confirm`)
- [ ] NVIDIA NIM shows Free but still needs an API key when missing
- [ ] MiMo Free does not show “No API key” / “Add API key”
- [ ] Add provider Free / Paid / Accounts tabs still work
- [ ] `bun test --isolate ./tests/provider-workspace-data.test.ts ./tests/server-auth.test.ts`
- [ ] Full suite: `bun test --isolate ./tests/`

PR #147, opened 2026-07-18
feat: combo virtual models for multi-provider load balancing

Wibias:
## Summary
- Backend for #133: `combo/<id>` virtual models with **failover** and **sticky weighted round-robin**, per-target cooldowns, and hop-on-retryable upstream failures (429 / 5xx / permission-subscription 403).
- Catalog exposure + `GET` / `PUT` / `DELETE` `/api/combos`.
- **`defaultEffort`** (default `medium`): fills `reasoning.effort` when the client omits it; **never overrides** client-sent effort; **skipped** when the resolved target has no reasoning ladder (`noReasoningModels` / empty / unknown ladder); clamped to a supported rung when a ladder exists.
- Intentionally the **runtime + API foundation** — Combos CRUD GUI is a follow-up draft PR on top of this API.

## Runtime
Combos runtime flowchart

User asks for `combo/<id>` → pick a target → call upstream → on retryable failure cool that target and hop to the next member until success or exhaustion.

## Test plan
- [x] `bun test tests/combos.test.ts`
- [x] `bun test src/combos/effort.test.ts`
- [x] `bun test tests/server-combo-failover-e2e.test.ts` (A 403 → B 200 through `handleResponses`)
- [x] Live proxy: `PUT`/`GET` `defaultEffort` round-trip after restart
- [x] Live proxy: failover hop with mock private upstreams via `/api/providers` + `combo/<id>` → `/v1/responses`
- [ ] Confirm non-combo models are unchanged


lidge-jun:
Thanks for building this — combos/load-balancing (#133) is a feature we want, and the review confirmed the direction. The implementation needed deeper changes than a review round-trip could carry, so we're rebuilding it as a small dependency-ordered stack directly on `dev`, preserving your authorship (your reconstruction commits carry you as author; redesigned parts carry Co-authored-by).

Review findings driving the rebuild (against head a4abda1, re-checked at 6824e7bc):
- **P0**: after a 503 hop, a backup provider's 401 entered the XAI OAuth refresh branch and the refreshed XAI bearer was sent to the backup provider (token leak);
- failover didn't cover connection exceptions (502 before combo recovery), passthrough/runTurn/web-search returned early, and cross-adapter hops died with HTTP 500 ("Non-streaming not supported");
- "sticky round-robin" was weighted-random with rotation after the first success regardless of stickyLimit, and the test asserted `picks.size >= 1`;
- deleted combo members fell open onto the default provider; post-hop vision/effort safety wasn't recomputed; catalog advertised fabricated capabilities; usage attribution recorded only the final provider.
- The PR was also 58 commits behind `dev` with content conflicts in `router.ts`/`responses.ts` and overlapped the landed #139/#140 rebuild surface.

**Slice 1 is already on `dev`** (f48a5c73 authored by you + f9a297ff/11984f9b/68cda367): validated combo CRUD, reserved `combo/` namespace with backward compat, deterministic smooth-weighted round-robin with real sticky semantics, fail-closed disabled/deleted members, and a typed `combo_unavailable` 503. Slices 2 (per-target failover engine, OAuth scoping, cross-adapter hops) and 3 (member-derived catalog capabilities, per-attempt usage attribution) follow; #133 stays open until they land.

Closing this PR in favor of the dev stack — the `defaultEffort` follow-up commit (6824e7bc) is folded into the slices too.

PR #152, opened 2026-07-18
fix: keep configured port after update restart

Wibias:
## Summary
- After GUI ""Restart after update"", pin proxy restart to `config.port` (`ocx start --port …`), wait briefly for the listen socket to free, and prefer-retry the configured port on start so Windows stop→start races no longer hop to an ephemeral port.
- Stop Windows console flashes on the start hot path (hidden PowerShell / process control).
- Shrink a long-run usage-debug loop so pre-push does not time out under load.

Windows ACL / Kimi `icacls` soft-fail lives in a separate PR (`fix/windows-acl-soft-fail`).

## Test plan
- [ ] GUI update → Restart after update → proxy stays on configured port (e.g. 10100)
- [ ] `bun test tests/ports.test.ts tests/update-job.test.ts`
- [ ] Confirm no visible console flash on Windows restart path

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -87,7 +87,12 @@ async function waitForProxy(timeoutMs = 8_000): Promise<LiveProxy | null> {
 async function chooseListenPort(requestedPort?: number): Promise<number> {
   const config = loadConfig();
   const preferred = requestedPort ?? config.port ?? 10100;
-  const selected = await findAvailablePort(preferred, config.hostname ?? "127.0.0.1");
+  // Brief prefer-retry covers stop→start races (update restart, `ocx restart`) where the
+  // old process has exited but the listen socket is still draining.
+  const selected = await findAvailablePort(preferred, config.hostname ?? "127.0.0.1", {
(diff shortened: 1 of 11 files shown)


lidge-jun:
Thanks for the fix — the port-stickiness direction was right, and it's landed on `dev` with your authorship preserved plus maintainer repairs on top:

- `2435836d` / `5605c6e2` / `19d424ad` — your PR #152 commits, absorbed as-is (author: @Wibias, source head e91da081).
- `fec1e4c5` — maintainer repairs (Co-authored-by: @Wibias) closing two review findings:
  1. `readAlivePid` as the default liveness pid could hand a reused/unverified PID to `ocx stop`'s kill path — killable pids now require a full cmdline identity check that must echo the exact candidate (TOCTOU-safe);
  2. the 2s port wait never ran in the real GUI-update sequence (stop-first update clears pid state before `restartAfterUpdate`) — the pre-update listen port is now captured before the update command and waited on unconditionally.
  Plus behavior tests driving both integration paths.

PR #154, opened 2026-07-19
fix(gui): dialog overlay frame and sticky Sub-agent focus ring

Wibias:
## Summary
- Request Logs detail uses a native `<dialog class="modal-overlay">`. Browser UA styles still draw a solid border plus a separate `::backdrop`, which stacked on our full-screen tint and produced the white rectangular frame / double-dimmed background around the error modal.
- Closing Dashboard Sub-agent help restored focus to the circular info button with `:focus-visible`, so the focus ring stacked on the icon’s own circle (sticky double ring until you clicked elsewhere). Restore now uses `focus({ focusVisible: false })` so keyboard a11y focus return stays, without the leftover mouse-close ring.

## Screenshots
image-0dff0f11-b5e5-43f6-848f-9be340a98647
image-11c86a74-4d37-434b-a5d9-1b70f10fef41

## Test plan
- [ ] Open Request Logs → click a failed row → confirm full-screen dim with no white frame / double tint
- [ ] Close the log detail (X / Escape / backdrop) → overlay clears cleanly
- [ ] Dashboard → open Sub-agent info help → close it → info icon should not keep a double focus ring
- [ ] Tab to the Sub-agent info button → keyboard focus ring still appears

### DIFF
--- a/gui/src/pages/Dashboard.tsx
+++ b/gui/src/pages/Dashboard.tsx
@@ -117,6 +117,16 @@ function sidecarBackendForModel(models: ModelInfo[], modelId: string): SidecarBa
   return models.find(model => model.id === modelId)?.provider === "anthropic" ? "anthropic" : "openai";
 }
 
+/** Restore focus to the opener without painting :focus-visible (avoids a sticky ring after mouse close). */
+function focusTriggerQuietly(trigger: HTMLButtonElement | null) {
+  if (!trigger) return;
+  try {
+    trigger.focus({ preventScroll: true, focusVisible: false });
+  } catch {
+    trigger.focus({ preventScroll: true });
+  }
+}
+
 function useModalDialog(open: boolean, triggerRef: RefObject<HTMLButtonElement | null>) {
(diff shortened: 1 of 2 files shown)


lidge-jun:
Thanks — the white UA frame + double-dim diagnosis was exactly right, and it's landed on `dev` with your authorship preserved plus a maintainer repair:

- `bfdaa3e9` — your PR #154 commit, absorbed as-is (author: @Wibias, source head 5e8f1aa1).
- `30782a87` — maintainer repair (Co-authored-by: @Wibias) closing the review findings:
  1. `focus({ focusVisible: false })` ran for keyboard closes too, so Escape users lost their focus indicator — input modality is now tracked window-wide and only pointer-originated closes suppress the ring;
  2. `focusVisible` support (Chrome 145+/FF 104+/Safari 18.4+) stays a progressive enhancement, with the origin tracking correct everywhere.

Render-verified in the browser: overlay clean, mouse close leaves no ring, Escape close paints the visible ring. Closing in favor of the dev-landed commits; ships with the next release.

PR #155, opened 2026-07-19
fix(windows): hide icacls during secret ACL hardening

Wibias:
## Summary
- Kimi login failed with `ACL hardening failed (ETIMEDOUT)` because writing `auth.json` runs `icacls` with a 5s timeout.
- On Windows, spawning `icacls.exe` without `windowsHide: true` from the GUI/proxy process can hang until that timeout.
- This PR only adds `windowsHide: true` to the three ACL hardening `icacls` calls so OAuth login can finish.

## Test plan
- [ ] `bun test tests/windows-secret-acl.test.ts`
- [ ] GUI Providers → Login Kimi (Moonshot) on Windows completes without ACL ETIMEDOUT

### DIFF
--- a/src/lib/windows-secret-acl.ts
+++ b/src/lib/windows-secret-acl.ts
@@ -64,11 +64,14 @@ function runIcacls(targetPath: string, directory: boolean): void {
     throw new Error("Cannot determine current Windows user for ACL hardening");
   }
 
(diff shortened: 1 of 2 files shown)


Wibias:
Closing as duplicate — same icacls windowsHide fix already lands in #152.

PR #156, opened 2026-07-19
fix(oauth): keep multiple Kimi accounts via JWT user_id

Wibias:
## Summary
- Bug: logging into a second Kimi (Moonshot) account replaced the active slot and effectively logged out the first, because Kimi tokens were stored with no `accountId`/`email`.
- Cause: `saveCredential` only appends when identity is present; identity-less providers replace the active account in place.
- Fix: decode the Kimi access/refresh JWT and set `accountId` from `user_id` (fallback `sub`), plus optional lowercased `email`. Distinct users append; same `user_id` upserts.

## Notes
- Pre-fix identity-less Kimi rows may remain as a one-time sibling after the next login of the same human; remove the stale row via accounts UI/API if needed.
- Opaque/non-JWT tokens still use replace-active semantics (no worse than before).

## Test plan
- [ ] `bun test tests/kimi-oauth-identity.test.ts tests/oauth-store-multi.test.ts`
- [ ] GUI: login Kimi account A, then account B — both appear; active is B; switch back to A
- [ ] Re-login account A — still one A row (upsert), B remains

### DIFF
--- a/src/oauth/kimi.ts
+++ b/src/oauth/kimi.ts
@@ -32,6 +32,53 @@ interface TokenResponse {
   interval?: number;
 }
 
+interface KimiJwtPayload {
+  user_id?: unknown;
+  sub?: unknown;
+  email?: unknown;
+}
+
+function decodeKimiJwtPayload(token: string): KimiJwtPayload | undefined {
+  const parts = token.split(".");
+  const payload = parts[1];
+  if (parts.length !== 3 || !payload) return undefined;
+  try {
+    return JSON.parse(Buffer.from(payload, "base64url").toString("utf-8")) as KimiJwtPayload;
+  } catch {
+    return undefined;
+  }
+}
+
+function nonEmptyString(value: unknown): string | undefined {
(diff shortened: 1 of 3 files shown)


lidge-jun:
Thanks — the root cause (identity-less Kimi tokens replacing the active slot) and the JWT user_id extraction were right, and it's landed on `dev` with your authorship preserved plus maintainer repairs:

- `a4086c87` — your PR #156 commit, absorbed as-is (author: @Wibias, source head a6d824ee).
- `29506b85` — maintainer repairs (Co-authored-by: @Wibias) closing three review findings:
  1. a pre-fix identity-less Kimi row was duplicated (not migrated) on the first post-fix login — the active legacy row is now upgraded in place, while distinct users still append;
  2. `user_id` was not actually preferred across tokens (access `sub` beat refresh `user_id`) — preference is now `access.user_id ?? refresh.user_id ?? access.sub ?? refresh.sub`;
  3. tests bypassed the production `parseTokenPayload` wiring — a mocked `refreshKimiToken` test now asserts the returned credential carries identity.

PR #157, opened 2026-07-19
fix(windows): soft-fail icacls ACL harden so Kimi auth is not blocked

Wibias:
## Summary
- Required NTFS ACL hardening via `icacls` was failing on Windows with `ETIMEDOUT` / `EPERM`, blocking `auth.json` writes during Kimi login and token refresh (surfaced as `invalid_api_key`).
- Switch to `Bun.spawnSync` with `windowsHide`, retries, and a process lock.
- Soft-fail `ETIMEDOUT` / `EPERM` / `EACCES` on required writes so auth continues with chmod-only hardening.

Example error this fixes:
```
ACL hardening failed (EPERM) — filesystem may not support per-user NTFS ACLs
```

## Test plan
- [ ] `bun test tests/windows-secret-acl.test.ts`
- [ ] On Windows: Kimi login + a coding request that refreshes tokens no longer fails with ACL hardening errors
- [ ] Confirm a warning is logged when ACL soft-fails, but the request proceeds

### DIFF
--- a/src/lib/windows-secret-acl.ts
+++ b/src/lib/windows-secret-acl.ts
@@ -18,12 +18,14 @@
  *   hardenSecretDir  — same contract for directories.
  */
 
-import { execFileSync } from "node:child_process";
 import { existsSync } from "node:fs";
 import { env, platform } from "node:process";
 
 const hardenedDirectories = new Set<string>();
 const hardenedPaths = new Set<string>();
+
+/** Serialize icacls — concurrent hardens on auth.json + dir during OAuth can stall until timeout. */
+let icaclsHeld = false;
 
 export interface HardenResult {
   ok: boolean;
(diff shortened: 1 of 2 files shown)


lidge-jun:
Thanks for tackling this — the field symptom (Kimi auth blocked behind a hung icacls) was real, and the fix landed on `dev` as a maintainer redesign that keeps your diagnosis and Bun.spawnSync/windowsHide direction:

- `a676f80b` — maintainer-authored (Co-authored-by: @Wibias, source head e19f38be named). The redesign diverges from the PR where review found safety gaps:
  1. every non-timeout icacls failure was mapped to `EPERM` and soft-failed, which defeated `required:true` on secret publication — errors now keep honest codes (`ETIMEDOUT` only for real timeouts, `EICACLS` otherwise) and only genuine timeouts soft-fail;
  2. `/remove:g` failures were swallowed — now verified with locale-independent `icacls /findsid` per broad SID before being accepted;
  3. 3×15s retries could stall `loadConfig` ~135s — replaced with one shared 5s deadline per harden call plus a timed-out-path cache;
  4. the module-local lock serialized nothing across processes — dropped.
  Failure paths are activation-tested via injected runner/platform/clock seams.

Full suite green (3,087 tests). Closing in favor of the dev-landed commit; ships with the next release.

PR #159, opened 2026-07-19
fix(quota): accurate Cursor totals and restore missing Kimi usage

Wibias:
## Summary
- Cursor period-usage now keeps `totalPercentUsed` as the monthly meter (matching Settings → Usage) while still showing First-party / API pool bars as secondary windows, and parses Connect RPC unix-ms reset strings.
- Kimi quota probing is hardened: unwrap `data` envelopes, map weekly windows from `limits[]` when top-level `usage` is absent, accept direct percent fields, and probe coding-host API-key providers (`kimi-code`) without ever sending credentials to non-canonical hosts.

## Test plan
- [x] `bun test tests/provider-quota.test.ts tests/quota-bars-rows.test.ts`
- [ ] GUI: Cursor OAuth provider shows monthly % matching Cursor Settings, plus optional First-party/API bars
- [ ] GUI: Kimi OAuth and/or `kimi-code` API-key on `https://api.kimi.com/coding/v1` shows 5h / weekly / total subscription credits
- [ ] Force refresh: `GET /api/provider-quotas?refresh=1` returns `cursor:period-usage` and `kimi:usages` without leaking tokens/raw payloads

### DIFF
--- a/src/providers/quota.ts
+++ b/src/providers/quota.ts
@@ -1,5 +1,6 @@
 import { fetchMainAccountInfo, listCodexAuthAccounts } from "../codex/auth-api";
 import { MAIN_CODEX_ACCOUNT_ID } from "../codex/main-account";
+import { resolveEnvValue } from "../config";
 import { getValidAccessToken } from "../oauth";
 import { getCredential } from "../oauth/store";
 import { antigravityUserAgent } from "../adapters/client-fingerprint";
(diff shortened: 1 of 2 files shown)


lidge-jun:
Thanks — both the Cursor totalPercentUsed fix and the Kimi limits[] mapping were right, and it's landed on `dev` with your authorship preserved plus a maintainer repair:

- `5ea37069` — your PR #159 commit, absorbed as-is (author: @Wibias, source head 90eca76e).
- `df0361dd` — maintainer repair (Co-authored-by: @Wibias) closing three review findings:
  1. an unresolved active env key silently fell back to the first resolvable `apiKeyPool` entry — the meter could belong to a different account than the one routing requests; the probe now uses the active key only;
  2. envelope detection used `!== undefined`, so `{ usage: null, data: {...} }` kept the dead outer object — null placeholders no longer mask the nested payload;
  3. `authMode !== "oauth"` also matched `forward`/`local` modes — the key-auth probe now requires `authMode === "key"`.

PR #167, opened 2026-07-19
fix(windows): stop GUI update from stacking visible terminals

Wibias:
## Summary
- GUI Update on Windows could open endless stacking terminals that never close.
- Root cause: Task Scheduler ran `opencodex-service.cmd` as a visible console, and the wrapper `:loop` retried `ocx start` every ~5s when start failed mid-update (often port-busy).
- Hide the wrapper via a WScript launcher (`Run …, 0, False`) and point the task at `wscript.exe`.
- Re-land port-sticky post-update restart (`ocx start --port` + prefer-retry / wait-for-free) from closed #152 so restarts do not hop ports and fuel the loop.
- Under the GUI update worker (`OCX_SERVICE=1`), use pipe stdio for npm/stop/service children so `npm.cmd` does not open stacked cmd windows.

## Test plan
- [x] `bun test tests/service.test.ts tests/update-job.test.ts tests/update-stop-first.test.ts tests/ports.test.ts`
- [ ] Windows: GUI Update with restart — no stacking terminals; proxy returns on configured port (e.g. 10100)
- [ ] If `ocx service install` is in use: reinstall/update rewrites the hidden VBS+XML; only one wrapper runs


lidge-jun:
Thanks for tracking down the console-storm root cause — the diagnosis in this PR (visible Task Scheduler console + wrapper retry loop) matches what we found independently, and it helped confirm the direction.

Since this PR opened, the `dev` branch landed its own Windows-service lane that covers most of this change, so I'm closing this as superseded rather than merging into `main` (our flow is dev → preview → main, and merging here would conflict with the next promotion):

- **Hidden VBS launcher** — already on dev (77b966ca), with UTF-16LE+BOM VBS/XML so non-ASCII profile paths survive, absolute wscript resolution, and VBS cleanup on uninstall.
- **Port-sticky restart** — already on dev in `src/update/job.ts` (captured live port → configured port → 10100, wait-for-free before start), tightened by fec1e4c5 from your earlier work.
- **Service reinstall after update** — dev goes further and preserves the selected backend (Scheduler vs the new opt-in WinSW `--native` service) via `serviceReinstallArgs()`.

One part of this PR was still unique and valuable: **piping child stdio under the GUI update worker (`OCX_SERVICE=1`) so `npm.cmd` doesn't open stacked consoles**. That hunk has been ported onto dev in a4f06beb with credit, extended to all three update children (stop / installer / service reinstall), keeping `serviceReinstallArgs()`.

So: closed as superseded, with the stdio fix landed. Thanks again!

PR #170, opened 2026-07-20
fix(proxy): raise default upstream stall timeout to 300s

Wibias:
## Summary
- Raise the default bridge `stallTimeoutSec` from **90s → 300s** so long quiet generations (large tool writes / reasoning) are not cut as `upstream_stall_timeout` → 502.
- Centralize the default in `resolveStallTimeoutSec()` so non-finite config values fall back cleanly instead of disabling the watchdog.
- Update docs (en/ko/zh) and web-search stall-budget expectations to match the new default.

## Why
Seen on Kimi k3 with large single-shot writes (~105s silence). Users should not need to split patches or tune config for normal long generations. Hung upstreams still abort — just with a more realistic window.

## Test plan
- [x] `bun test tests/stall-timeout.test.ts tests/web-search-timeout-plan.test.ts tests/web-search.test.ts tests/bridge-lifecycle.test.ts tests/bridge.test.ts`
- [x] Full pre-push suite (3130 pass)
- [ ] Reproduce a long Kimi write that previously 502'd around ~90–105s; confirm it completes under the new default
- [ ] Optional: set `stallTimeoutSec: 30` in config and confirm hangs still abort promptly


lidge-jun:
Squash-merged onto dev as f07234f0 — will land on main with the next dev→main merge. Thanks @Wibias!

Issue #42, opened 2026-06-28
Add Storage page for Codex session usage and cleanup policy


0disoft:
## Problem

Codex session data can grow quietly over time, but the dashboard currently does not show where that storage is going or offer a guided way to reclaim space.

On one active Windows setup, the largest Codex-related storage buckets were roughly:

- archived sessions: ~3.1 GB
- active sessions: ~1.3 GB
- logs sqlite: ~1.1 GB
- attachments / remote attachments / deletion manifests: smaller, but still useful to surface

The exact numbers will vary by user, but the important part is that storage is split across multiple Codex-owned areas. Users often only notice the problem after disk space has already dropped sharply.

## Proposal

Add a dashboard Storage page that starts with read-only diagnostics, then grows into cleanup controls if the direction makes sense.

### Phase 1: storage diagnostics

Show a compact breakdown for:

- active sessions
- archived sessions
- logs database / WAL files
- attachments
- remote attachments
- deletion manifests
- total tracked size

Useful fields:

- size
- count
- oldest item
- newest item
- largest few items, capped to avoid rendering a huge table

### Phase 2: manual archived cleanup

Add an archived-session cleanup flow with preview first:

- slider for deleting the oldest archived conversations by percentage
- quick presets such as 10%, 25%, 50%
- preview button before any deletion
- confirmation modal showing estimated count and bytes to remove
- optional checkbox for immediate permanent deletion
- safer default can be quarantine/trash if preferred by the project

Active sessions should probably stay read-only at first.

### Phase 3: optional auto-cleanup policy

Add an opt-in policy for users who want storage to stay bounded:

- default off
- threshold such as "archived sessions exceed N GB"
- cleanup target such as "reduce to N GB" or "delete oldest N%"
- schedule such as startup / daily / weekly / manual
- deletion mode: quarantine/trash or permanent delete
- last run, next run, and freed bytes summary

This should run as a batch policy, not delete one item every time a new session appears.

## Implementation concerns

Deleting Codex history is not just a simple filesystem delete. The implementation may need to account for:

- rollout JSONL files
- `state_5.sqlite` thread rows and rollout paths
- attachments linked to removed sessions
- deletion manifests / audit trail if quarantine is used
- database locks while Codex is active

Because of that, I think the first PR should probably be read-only diagnostics plus cleanup preview, with actual deletion added only after the data model is clear.

## Open questions

- Should opencodex directly manage Codex session deletion, or only expose diagnostics?
[...]

PR #173, opened 2026-07-20
feat(gui): read-only Storage diagnostics page (Phase 1 of #42)

Chang-Jin-Lee:
## Summary

Implements **Phase 1** of the Storage page epic (#42) exactly as scoped by the maintainer's own plan in `devlog/_plan/500_storage-page-session-cleanup` (`21`, `31`–`33`): a dashboard page that answers "why is my `CODEX_HOME` N GB?" — **read-only, zero deletion logic**. Phases 2/3 (manual cleanup, auto-policy) are untouched.

- `src/storage/scanner.ts` (new): walks `CODEX_HOME` and buckets `sessions/`, `archived_sessions/`, `state_*.sqlite`, `logs_*.sqlite`, `attachments/`, `deletion_manifests/`, and everything else into `other`. Sizes/counts/oldest/newest/largest-5 come from `fs.stat` only. DB row counts use a short-busy-timeout readonly open — the same secondary-reader contract as `codex/history-provider.ts` — and degrade to `null` on lock or corruption instead of throwing. Versioned `state_N`/`logs_N` files resolve to the newest suffix.
- `GET /api/storage` in `src/server/management-api.ts`, following the `/api/usage` template: `jsonResponse(scanStorage())` with a `try/catch` fallback envelope (`error: "scan_failed"`) on failure.
- `gui/src/pages/Storage.tsx` (new): total-size/file-count cards, a per-bucket table (size/count/oldest/newest/DB rows), and a largest-files disclosure per bucket. No delete affordance anywhere. Wired into `App.tsx` nav next to Usage, with i18n keys added to all four locales (en/ko/zh/de) and a new `IconHardDrive` glyph.

## Verification

- `bun run test` — 3142/3142 passing, including 8 new scanner unit tests (bucket aggregation vs. fixture byte sizes, largest-file ranking, lock-safe null counts on a corrupt/locked DB, missing-home zeros, a broken-home throw path, `CODEX_HOME` env default, and the **read-only invariant** — a before/after mtime+inode snapshot of the whole fixture tree asserting zero writes during a scan) and 2 `/api/storage` integration tests (success envelope, scan-failure fallback envelope). All written first and watched red before the implementation went green.
- `cd gui && bun run build` — passes (tsc + vite), `bun run lint` / `lint:i18n` clean.
- `npx react-doctor --scope changed` — zero findings in any new/changed file.
- Manual: `ocx start` against a temp `CODEX_HOME`, curled `/api/storage` directly and confirmed the bucket/total shape.

## Checklist

- [x] Docs or release notes were updated when needed. *(N/A — Phase 1 has no user-facing config/CLI change; the maintainer's own devlog plan is the spec of record.)*
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. *(Read-only GET endpoint reusing the existing dashboard API auth middleware; no new write path, no new auth surface.)*

Claimed via a comment on #42 before starting: https://github.com/lidge-jun/opencodex/issues/42#issuecomment-5019502268


lidge-jun:
Squash-merged onto dev as 24d8bbb9. Will land on main with the next dev→main merge. Thanks @Chang-Jin-Lee!

PR #182, opened 2026-07-20
fix(antigravity): stop forwarding Responses item ids as thoughtSignature

### LINKED ISSUE #174: Antigravity 400 "Base64 decoding failed" when a thread with Claude history switches to a Gemini model (ctc_* signatures forwarded as thoughtSignature)
## Summary

Switching an existing Codex thread to a `google-antigravity` Gemini model fails with a provider 400 when the thread history contains assistant turns produced by an Anthropic (Claude) model. The Claude reasoning signatures (`ctc_...`) stored on tool calls are forwarded to Antigravity as `thoughtSignature`, and upstream rejects them because they are not valid Gemini signatures.

## Environment

- opencodex `2.7.27` (npm, Windows x64, bundled Bun runtime)
- Codex App routed through the proxy; providers: `openai`, `anthropic` (OAuth), `google-antigravity` (OAuth, cloud-code-assist mode)

## Reproduction

1. Start a Codex conversation on an `anthropic/...` model and let it make tool calls (so assistant turns carry `ctc_...` signatures in history).
2. In the same thread, switch the model to e.g. `google-antigravity/gemini-3.1-pro-high` and send a message.
3. The request fails:

```
Provider error 400: Antigravity invalid request: Invalid value at 'request.contents[26].parts[1].thought_signature' (TYPE_BYTES), Base64 decoding failed for "ctc_038f26d3f20962bc016a54f0fcfa208190a8ec0f289c2ba211"
Invalid value at 'request.contents[28].parts[0].thought_signature' (TYPE_BYTES), Base64 decoding failed for "ctc_038f26d3f20962bc016a54f10302648190b3c87c814750cad5"
...
```

Cross-provider switches that tolerate foreign signatures (GPT <-> Claude, Gemini -> others) work fine; only history-with-Claude -> Antigravity/Gemini hits this.

## Root cause

`isLikelyRealThoughtSignature()` in `src/adapters/google-antigravity-wire.ts` is meant to keep synthetic/foreign values off the wire, but Claude signatures pass both checks:

- the synthetic-id prefix regex (`fc|call|msg|rs|...`) does not include `ctc`
- the final character-set check `/^[A-Za-z0-9+/_=-]+$/` matches `ctc_...` strings, since they are hex + underscore

So `google.ts` attaches them as `part.thoughtSignature` on functionCall parts, and Antigravity rejects the request at `TYPE_BYTES` validation.

## Suggested fix (verified locally)

Add Anthropic-style prefixes to the rejection list:

```ts
export function isLikelyRealThoughtSignature(sig: string | undefined): boolean {
  if (typeof sig !== "string" || sig.length < 16) return false;
  if (/^(fc|call|msg|rs|resp|reasoning|item|ws|tool|func|function)[-_]/i.test(sig)) return false;
  // Reject Claude/Anthropic signatures (`ctc_...`, `toolu_...`, `sig_...`): they are not Gemini
  // thought signatures and Antigravity 400s on them ("Base64 decoding failed").
  if (/^(ctc|toolu|sig|claude|anthropic)[-_]/i.test(sig)) return false;
  return /^[A-Za-z0-9+/_=-]+$/.test(sig);
}
```

With this patch applied to the installed package:

- `isLikelyRealThoughtSignature("ctc_038f26d3f20962bc016a54f0fcfa208190a8ec0f289c2ba211")` → `false`
- a real Gemini base64 signature still returns `true`
- the mixed-history thread can switch to Antigravity Gemini models without the 400

A more general alternative would be to tag stored signatures with their originating adapter and only replay them to the same adapter family, but the prefix-based rejection is a minimal safe fix for this failure mode.

Happy to open a PR with the change plus a regression test if useful.

Wibias:
## Summary
- Fixes #174: switching a Codex thread from Claude (or any history with `ctc_*` / similar item ids) to `google-antigravity` Gemini 400s because foreign ids were forwarded as `thoughtSignature`.
- **Root fix:** Responses parser no longer maps item `id` (`fc_` / `ctc_` / …) onto `OcxToolCall.thoughtSignature` — that field is for Gemini/Antigravity opaque tokens; CCA continuity stays on the in-process replay cache.
- **Defense in depth:** `isLikelyRealThoughtSignature` now also rejects `ctc_` and `toolu_`. Explicitly does **not** deny-list `sig_*` (that would break valid replay fixtures / upstream blobs). This is stricter than growing a vague prefix list alone.

## Why not only the issue’s patch?
A deny-list-only change (`ctc|toolu|sig|…`) is whack-a-mole and rejecting `sig_` is incorrect. Stopping the parser from contaminating the field is the durable fix; the filter catches already-poisoned history.

## Test plan
- [x] `bun test tests/google-antigravity-wire.test.ts tests/responses-parser-agent-message.test.ts tests/google-antigravity-replay.test.ts`
- [ ] Repro from #174: Claude/tool history → switch to `google-antigravity/gemini-3.1-pro-high` → no Base64/`thought_signature` 400
- [ ] Gemini→Gemini same-process multi-turn with tools still works (replay cache)
- [ ] Real `sig-…` / base64 thought signatures still forward when present on the tool call

### DIFF
--- a/src/adapters/google-antigravity-wire.ts
+++ b/src/adapters/google-antigravity-wire.ts
@@ -12,16 +12,23 @@ export const ANTIGRAVITY_REQUEST_UA = process.env.GOOGLE_ANTIGRAVITY_USER_AGENT 
 
 /**
  * Whether a stored `OcxToolCall.thoughtSignature` is a REAL upstream Gemini signature versus a
- * synthetic Responses item id (`fc_...`, `call_...`, `rs_...`, etc) that the bridge/parser stashes
- * on the field. Only real signatures may be forwarded to Gemini/Antigravity — sending a synthetic
- * id as `thoughtSignature` breaks multi-turn reasoning continuity (upstream rejects it). Real
- * signatures are opaque base64-ish blobs with no Responses-id prefix.
+ * foreign id that must not be forwarded to Gemini/Antigravity.
+ *
+ * Foreign ids that have 400'd Antigravity (`TYPE_BYTES` / Base64 decoding failed) include:
+ * - Responses/bridge item ids: `fc_...`, `ctc_...` (custom_tool_call), `call_...`, `rs_...`, …
+ * - Anthropic tool-use ids: `toolu_...`
+ *
+ * Only real signatures may be forwarded — sending a foreign id as `thoughtSignature` breaks
+ * multi-turn reasoning continuity. Real signatures are opaque base64-ish blobs with no
+ * Responses/Anthropic id prefix.
+ *
+ * Note: do NOT reject a bare `sig_` / `sig-` prefix — existing Gemini replay fixtures and some
+ * upstream blobs use that shape; a deny-list entry for `sig` would drop valid continuity tokens.
  */
 export function isLikelyRealThoughtSignature(sig: string | undefined): boolean {
   if (typeof sig !== "string" || sig.length < 16) return false;
-  // Reject synthetic Responses/tool-call ids in both `_` and `-` separated spellings
-  // (e.g. `fc_...`, `call_...`, `function-call-...`, `tool-call-...`).
-  if (/^(fc|call|msg|rs|resp|reasoning|item|ws|tool|func|function)[-_]/i.test(sig)) return false;
+  // Reject synthetic Responses/tool-call ids and Anthropic tool-use ids (`_` or `-` separators).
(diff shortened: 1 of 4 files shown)


lidge-jun:
Squash-merged onto dev as 30a71fe3 — will land on main with the next dev→main merge. Thanks @Wibias!

PR #187, opened 2026-07-20
fix(storage): use immutable readonly opens for DB row counts

Chang-Jin-Lee:
## Summary

Follow-up to #173 (squash-merged as 24d8bbb9 on `dev`), addressing the Codex bot's P2 finding on that PR that I didn't get a chance to land before it merged: **`new Database(path, { readonly: true })` is not filesystem-read-only for a WAL-mode database.**

Confirmed empirically: opening and querying a checkpointed `state_5.sqlite`/`logs_2.sqlite` (WAL journal mode, no live `-wal`/`-shm` sidecars yet — the normal state once Codex has fully quit) via a plain readonly connection materializes both sidecar files. That means simply visiting the read-only Storage page could write under `CODEX_HOME` on its first render — directly contradicting the Phase 1 read-only invariant from the epic's own plan.

**Fix:** `countRowsReadonly` now opens via an immutable connection — `file:<path>?immutable=1` through a `file:` URI, which requires passing `SQLITE_OPEN_URI` explicitly since bun:sqlite only parses `file:` URIs when that flag is set. `immutable=1` tells SQLite the file will never change for the connection's lifetime, so it skips WAL/shm entirely. Tradeoff: it reads the last-checkpointed snapshot rather than respecting SQLite's lock protocol — verified this means a scan now completes instantly through an active writer's exclusive lock (no `SQLITE_BUSY` wait) while still returning the correct row count for the realistic case, and, more importantly, still writes zero new files either way.

**Corollary bug this surfaced:** building the URI by string concatenation (`` `file:${dbPath}?immutable=1` ``) breaks on any `CODEX_HOME` path containing `#`, `?`, or `%` — legal on POSIX filesystems, but URI-reserved, so they get parsed as a fragment/query/escape and every row count silently degrades to `null`. Fixed by building the URI with `pathToFileURL` instead.

## Verification

- Added a dedicated regression test reproducing the original bug: a WAL-mode fixture with `wal_checkpoint(TRUNCATE)` and no existing sidecars must not gain `-wal`/`-shm` files after a scan (confirmed this fails against the pre-fix code).
- Rebuilt the scanner test fixture to actually use `PRAGMA journal_mode=WAL` for `state_5.sqlite`/`logs_2.sqlite`, matching real Codex files — the previous default-journal-mode fixture never exercised the buggy path, which is why it shipped in #173 in the first place.
- Renamed/updated the exclusive-lock test to reflect actual immutable-mode semantics (reads through the lock instead of blocking or degrading to null) and confirmed it still performs zero writes.
- Added a new test for the `#`/`?`/`%`-in-path case (watched it fail pre-fix, pass post-fix).
- `bun x tsc --noEmit` clean, `bun run test` 3229/3229 on `dev`, `bun run privacy:scan` clean.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup — touches only `src/storage/scanner.ts` and its test file.
- [x] Docs or release notes were updated when needed. *(N/A — internal implementation detail, no user-facing behavior change beyond removing the write.)*
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. *(Narrows write surface — no new write path; readonly stays readonly.)*

### DIFF
--- a/src/storage/scanner.ts
+++ b/src/storage/scanner.ts
@@ -1,7 +1,17 @@
 import { readdirSync, statSync } from "node:fs";
 import { join } from "node:path";
-import { Database } from "bun:sqlite";
+import { pathToFileURL } from "node:url";
+import { Database, constants } from "bun:sqlite";
 import { resolveCodexHomeDir } from "../codex/home";
+
+// SQLITE_OPEN_READONLY alone is not filesystem-read-only for a WAL-mode DB: Bun's
+// `{ readonly: true }` can still materialize *.sqlite-wal/-shm sidecars the first time a
+// checkpointed WAL database (no live sidecars yet) is opened and queried. `immutable=1`
+// (via a file: URI, which requires the SQLITE_OPEN_URI flag) tells SQLite the file will
+// never change for this connection's lifetime, so it skips WAL/shm entirely — the
+// tradeoff is reading the last-checkpointed snapshot instead of blocking on a live writer,
+// which is the right tradeoff for a passive diagnostics scan that must never write.
+const IMMUTABLE_READONLY_FLAGS = constants.SQLITE_OPEN_READONLY | constants.SQLITE_OPEN_URI;
 
 /**
  * Read-only CODEX_HOME storage scanner — Phase 1 of the Storage page epic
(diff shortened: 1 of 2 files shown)


lidge-jun:

Thanks for the solid approach — `immutable=1` is the right direction for a never-write scanner, and `pathToFileURL` URI encoding is correct.

However, two test issues block merge:

### 1. macOS test failure (High)
`bun test tests/storage-scanner.test.ts` → 9 pass, 1 fail:
```
Expected: 12384
Received: 45152
```
Bun uses persistent WAL on macOS, so `wal_checkpoint(TRUNCATE)` + `close()` leaves WAL/SHM files behind. The byte assertion in `aggregates bucket bytes` breaks because those sidecar files are included in `state_db.bytes`.

**Fix:** disable persistent WAL before checkpoint/close:
```ts
state.fileControl(constants.SQLITE_FCNTL_PERSIST_WAL, 0);
state.exec("PRAGMA wal_checkpoint(TRUNCATE)");
state.close();
```
Then assert that the newest DB starts without sidecars.

### 2. Windows CI failure (High)
The URI-reserved-character test creates `weird#name?with%percent`. Windows forbids `?` in directory names, so `mkdirSync()` will throw on Windows CI.

**Fix:** test `#` and `%` cross-platform; include `?` only behind a `process.platform !== "win32"` guard, or split into a POSIX-only test.

### Design note (not a blocker)
The `immutable=1` tradeoff (last-checkpointed snapshot vs never-write) is a valid design choice for a passive diagnostics scanner — this is acknowledged in the code comment and acceptable.

Please fix the two test issues and we can merge. 🙏

Ingwannu:
Reviewed the current head (`62cf4def6d7d16d3869cd43d9f49478d08111492`) against `dev`.

- Security diff review found no malware, backdoor, hidden execution/persistence, credential exposure, or new network behavior.
- The change strengthens the Storage page's read-only invariant by preventing SQLite WAL/SHM sidecar creation.
- Verified locally: 12 focused storage/API tests passed, TypeScript typecheck passed, and `privacy:scan` passed.
- URI-reserved path characters and zero-write behavior are covered by regression tests.

This is focused and valuable. Approved for `dev`; merging via squash.

Issue #190, opened 2026-07-20
[Bug]: Cursor resource_exhausted with large Codex namespace tool catalog


alexanderxcmo:
## Summary

Cursor-routed models fail deterministically with `resource_exhausted` when Codex advertises a large namespace-based tool catalog. The same Cursor OAuth account and model work when the namespace tools are removed, and both halves of the catalog work independently.

This appears related to, but more specifically reproducible than, #72. PR #73 improved error classification/retry behavior, but the upstream rejection can still be triggered by the aggregate client tool catalog.

## Environment

- OpenCodex: `2.7.28`
- Codex CLI: `0.144.6` (also observed from Codex Desktop)
- OS: macOS
- Provider: Cursor OAuth
- Models reproduced: `cursor/grok-4.5` and `cursor/auto`
- Cursor quota was not exhausted, and native Cursor usage continued to work

## Error

```text
stream disconnected before completion: Cursor rate limit exceeded: Cursor Connect error resource_exhausted: Error
```

Provider debug shows that Cursor Connect succeeds, returns its first frame, requests multiple blobs, and then terminates with `resource_exhausted`:

```text
[ocx:cursor:connected]
[ocx:cursor:first-frame]
[ocx:cursor:frame] {"case":"kvServerMessage","kv":"getBlobArgs"}
...
[ocx:cursor:connect-end-stream] {"code":"resource_exhausted"}
```

## Reproduction

1. Configure the Cursor OAuth provider.
2. Use a Codex installation with multiple apps/plugins, producing a namespace-based tool catalog.
3. Select `cursor/grok-4.5` or `cursor/auto`.
4. Send a trivial prompt in a fresh session.

A fresh ephemeral CLI session also reproduces it:

```bash
codex exec \
  --ignore-user-config \
  -c openai_base_url='"http://127.0.0.1:10100/v1"' \
  --skip-git-repo-check \
  --ephemeral \
  -m cursor/grok-4.5 \
  -c model_reasoning_effort='"medium"' \
  "Reply only OK"
```

## Request measurements

The failing minimal Codex request was zstd-compressed on the wire and measured after local decompression:

- Total request: `505,203` bytes
- Tool catalog: `443,259` bytes
- Top-level tools: `24`
- Namespace entries: `13`
- Inner namespace tools: `340`

The largest namespaces included GitHub (89 tools), Linear (59), Google Drive (46), Slack (32), Neon Postgres (29), Figma (27), Vercel (24), and Sites (20).

[...]

PR #192, opened 2026-07-20
fix(cursor): enforce tool budget to prevent resource_exhausted with large catalogs

alexanderxcmo:
## Problem

Cursor Connect returns `resource_exhausted` when the aggregate tool catalog exceeds an internal registration limit. This is **not** a quota issue — the account has remaining credits and native Cursor works fine.

Reproduced with 340 namespace-inner tools (443 KB serialized). Both halves of the catalog (230 and 110 tools) work independently, confirming the failure depends on the aggregate size, not a single malformed schema.

See #190 for full reproduction, measurements, and isolation results.

## Fix

Add a Cursor-specific tool budget (default 200) in `createCursorRequest`:

- **Native tools** (no namespace) are always preserved — `exec_command`, `apply_patch`, `web_search`, `tool_search`, etc.
- **Namespace groups** are sorted smallest-first and added until the budget is full.
- **Remaining namespaces** are dropped from the initial advertisement. The model can still discover and load them via `tool_search`.
- A **system-prompt note** is appended when trimming occurs, so the model knows to use `tool_search` for unavailable tools.

## Why 200?

Empirically determined: 230 tools succeed, 340 fail. 200 provides safe margin while still covering most real-world setups (typical Codex installs with a few MCP servers stay well under 200).

## Testing

- 4 new unit tests in `cursor-request-builder.test.ts`:
  - Budget enforcement with mixed native + namespace tools
  - No trimming when under budget
  - Deferred-tools system note when trimmed
  - No note when nothing trimmed
- All 277 existing cursor tests pass
- `tsc --noEmit` clean

## Future work

A more complete fix would implement full deferred loading for Cursor (send only tool names in the initial `mcp_tools` registration, load full schemas on `tool_search`). This PR provides an immediate, safe mitigation that unblocks all users with large plugin/app catalogs.

### DIFF
--- a/src/adapters/cursor/live-transport.ts
+++ b/src/adapters/cursor/live-transport.ts
@@ -434,7 +434,7 @@ class LiveCursorTransport implements CursorTransport {
     const activeText = activePromptText(request);
     this.activeClientToolFinalizeGraceMs = clientToolFinalizeGraceMsForRequest(request, this.clientToolFinalizeGraceMs);
     const cursorVisibleTools = cursorToolsForActivePrompt(request.tools, activeText, request.toolChoice);
-    const clientToolDefs = buildCursorToolDefinitions(cursorVisibleTools, request.toolChoice);
+    const clientToolDefs = buildCursorToolDefinitions(cursorVisibleTools, request.toolChoice, request.stubbedToolNames);
     this.execContext = {
       ...this.execContext,
       clientToolDefs,
--- a/src/adapters/cursor/protobuf-request.ts
+++ b/src/adapters/cursor/protobuf-request.ts
@@ -379,7 +379,7 @@ export function encodeCursorRunRequest(request: CursorRunRequest): Uint8Array {
     // tool-count prompt, so a call to it would be rejected as an unknown Responses tool.
     ...(() => {
       const visibleTools = cursorToolsForActivePrompt(request.tools, activePromptText(request), request.toolChoice);
-      const mcpToolDefs = buildCursorToolDefinitions(visibleTools, request.toolChoice);
+      const mcpToolDefs = buildCursorToolDefinitions(visibleTools, request.toolChoice, request.stubbedToolNames);
       return mcpToolDefs.length > 0 ? { mcpTools: create(McpToolsSchema, { mcpTools: mcpToolDefs }) } : {};
     })(),
   });
--- a/src/adapters/cursor/request-builder.ts
+++ b/src/adapters/cursor/request-builder.ts
@@ -11,6 +11,8 @@ import type { CursorRequestMessage, CursorRunRequest } from "./types";
 import { cursorCodexToWireModelId } from "./discovery";
 import { cursorEffortSuffix } from "./effort-map";
 import type { OcxTool } from "../../types";
+import { toolChoiceAliases, isAllowedToolChoice } from "../../types";
+import type { OcxToolChoice } from "../../types";
 
 /**
  * Maximum number of tools (native + namespace-inner) that the Cursor Connect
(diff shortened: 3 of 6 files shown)


Wibias:
## Review notes - what should change before treating this as done

Direction is right as a crash mitigation (trim before `mcp_tools`), but a few claims/edges need fixing so we do not ship a false recovery story.

### 1. Do not claim `tool_search` fully recovers trimmed tools (high)

Today `previouslyUsedToolNames()` only reserves tools that already appear as `toolResult` (i.e. already *called*). After `tool_search` *discovers* a tool, that tool is usually not reserved yet, so the next budget pass can drop its namespace again and Cursor still never gets it in `mcp_tools`.

Either:
- **Fix recovery**: reserve tools named/returned by `tool_search` outputs (and keep them across turns), **or**
- **Tone down**: say this only prevents `resource_exhausted` by omitting overflow namespaces; dropped tools may stay unavailable until a real deferred-registration path exists.

The current system note promising `tool_search` loading oversells what the Cursor path can do.

### 2. Cap `native + reserved`, not only bare native (high)

`cappedNative` is limited to `CURSOR_TOOL_BUDGET`, but `kept = cappedNative + reserved (+ namespaces)` can still exceed 200 when `toolChoice` / prior-use reserves are large — reintroducing the same failure mode this PR aims to stop.

Apply one hard total cap after assembling reserved + native (and decide an explicit priority order when something must be dropped).

### 3. Be honest that the limit is a heuristic (medium)

Issue #190 left the boundary open (count vs serialized bytes vs blob budget). `200` is a reasonable empirical count guard, but not proven. Worth:
- making the budget env/config-tunable for field tuning, and/or
- noting in the PR/code comment that fat schemas under 200 might still fail.

### 4. Revisit smallest-first packaging (medium / product)

Smallest-first maximizes namespace *count*, so tiny MCP servers survive while large useful ones (GitHub/Linear/…) get cut. That unblocks sessions but may drop the tools users care about most. Consider largest-first, a stable allowlist/priority, or filling remaining budget with a *partial* slice of one oversized namespace instead of dropping it entirely when e.g. 189 slots remain and the group has 216 tools.

### 5. Error classification still wrong when it slips through (low / follow-up)

#190 also asked not to surface this as a quota/rate-limit. If Cursor still returns `resource_exhausted` for an oversized catalog, prefer an actionable "tool catalog too large for Cursor" message. Fine as follow-up if this PR stays a mitigation-only patch.

---

**Suggested merge bar:** either implement real post-`tool_search` reservation (or equivalent), **or** reword docs/note + fix the `native+reserved` overflow cap. Happy to help on either path.

alexanderxcmo:
## Probe results & hybrid approach (commit 5fff484)

### Live probe 2026-07-21

Ran a controlled probe against `api2.cursor.sh` with stub schemas (`{"type":"object","properties":{}}`) to isolate byte vs count limits:

| Tools (stub schemas) | Status |
|---:|:---:|
| 250 | ✅ |
| 280 | ✅ |
| 300 | ✅ |
| 320 | ✅ |
| 330 | ✅ |
| 332 | ✅ |
| 333 | ❌ resource_exhausted |
| 340 | ❌ resource_exhausted |

**Finding: Cursor has TWO independent limits:**
1. **COUNT limit: ~332 tools** — hard cap on `McpToolDefinition` entries in `mcp_tools`, regardless of payload size
2. **BYTE limit: ~128 KB** — total protobuf payload for `mcp_tools` field; with full schemas (~566 B/tool avg) this caps at ~230 tools

The old budget of 200 was conservative against the byte limit. With stub schemas we can reach 330 tools (just under the count cap).

### New hybrid approach (this commit)

Instead of dropping tools beyond the budget, we now:

1. **Raise count budget to 330** (from 200) — fits under the 332 count cap
2. **Keep full schemas** for tools within a 120 KB byte budget (typically ~150-200 tools)
3. **Stub `inputSchema`** for remaining tools (name + description only, ~30 B each)
4. **Inject a text schema catalog** into the system prompt listing stubbed tools with their argument structures
5. **`tool_search`** still available for on-demand full schema loading

Result: **all 340 tools visible** to the model. Most have full structured schemas; the remainder have stub schemas + text catalog entry. No tools permanently inaccessible.

### Addresses review findings

- **(HIGH) tool_search recovery**: Stubbed tools are tracked via `stubbedToolNames` on `CursorRunRequest`. The text catalog gives the model argument info without needing tool_search. When tool_search IS used, the existing `previouslyUsedToolNames()` reservation still applies.
- **(HIGH) native + reserved overflow**: Hard count cap at 330 prevents overflow. `computeStubbedTools()` runs after budget enforcement, so the total is always ≤ 330.
- **(MEDIUM) 200 is heuristic**: Now documented as two probe-verified limits (count=332, byte≈128KB). Constants renamed to `CURSOR_TOOL_COUNT_BUDGET` and `CURSOR_TOOL_BYTE_BUDGET`.
- **(MEDIUM) smallest-first drops large namespaces**: With 330 capacity, most namespaces fit. The byte-budget stubbing means even large namespaces (GitHub 89, Linear 59) are included — just with stub schemas for tools beyond the byte threshold.
- **(LOW) error classification**: Unchanged, follow-up.

### Tests

19 tests pass (4 new for hybrid behavior), `tsc --noEmit` clean.

New tests cover:
- Stub schemas applied when byte budget exceeded but count budget allows all tools
- No stubbing when catalog fits within byte budget
- Text catalog injected into system prompt with argument summaries
- 250 tools (between old 200 and new 330 budget) all survive without stubbing

lidge-jun:

lidge-jun:

Good problem identification — the Cursor 332-tool count ceiling and ~128KB byte ceiling are real constraints.

### Blockers

1. **Byte budget estimates only schema JSON, not full protobuf** — `estimateSchemaBytes()` only sizes the inputSchema JSON, but the actual McpToolDefinition protobuf includes name, toolName, provider, description fields too. With 300 large-schema tools, the stubbed payload was still 136,983 bytes (over the 128KB ceiling). Use actual `McpToolsSchema` serialization to measure.

2. **Reserved tools bypass the hard count cap** — Native tools are capped at 330, then reserved tools are added back unconditionally. With 330 native + 3 reserved namespace tools, the final count is 333 — exactly at the failure boundary. Apply one final cap across all categories.

3. **Count-trimmed tools silently disappear** — Dropped namespaces aren't included in `buildTextCatalog()` (which only covers retained-but-stubbed tools), yet the system prompt says "full schemas are listed in the catalog below." Also, `tool_search`-loaded tools are re-trimmed on the next request with no promotion mechanism.

Please retarget to `dev` — see [Contributing guide](https://lidge-jun.github.io/opencodex/contributing/). 🙏

lidge-jun:
Closing this PR — the problem identification (Cursor's ~332 tool count ceiling and ~128KB byte ceiling) is valuable research, but the implementation has three fundamental issues that need an architectural rethink:

1. **Byte budget estimates only schema JSON** — the actual protobuf includes name, toolName, provider, description fields. With large schemas, stubbed payloads still exceed 128KB.

2. **Reserved tools bypass the hard count cap** — 330 native + reserved tools can exceed the 332 boundary.

3. **Count-trimmed tools silently disappear** — dropped namespaces aren't in the text catalog, and tool_search-loaded tools are re-trimmed on the next request with no promotion mechanism.

The probe data on Cursor's limits is useful reference material. Thank you for the research — we'll tackle this with a serialization-aware budget approach. 🙏

Co-authored-by: alexanderxcmo

PR #193, opened 2026-07-20
fix(update): hard-pin listen port after update (complete #152)

Wibias:
## Summary
- Follow-up to [#152](https://github.com/lidge-jun/opencodex/pull/152) / absorb `2435836`: that fix pinned non-service GUI restart with `--port`, but **did not stop port hops** on real updates (especially 2.7.26 → 2.7.28) when a service was installed or when the preferred port stayed busy past the short prefer-retry window.
- **Hard-pin when `--port` is explicit**: longer prefer-retry, `allowEphemeralFallback: false`, and same-port bind retries — never allocate an ephemeral listener.
- **Bake `--port` into every service start path** (Windows batch, WinSW XML, launchd, systemd) from `config.port` / `OCX_BAKE_PORT`.
- **Service update restart** now waits on the captured pre-update port and sets `OCX_BAKE_PORT` before reinstall (GUI worker, `ocx update`, and npm `bin/ocx.mjs`).
- Detached ensure/gui/claude starts also pass `--port` from config.

## Why #152 was not enough
1. Service wrappers still ran `ocx start` with no `--port`.
2. Service update reinstall did not wait for the old socket to free.
3. Even `start --port N` could still fall back to an ephemeral port after ~750ms.

## Test plan
- [ ] `bun test tests/ports.test.ts tests/update-job.test.ts tests/service.test.ts tests/winsw.test.ts tests/windows-deploy-close-regressions.test.ts tests/update-stop-first.test.ts`
- [ ] Non-service: GUI update → Restart after update → stays on configured port
- [ ] Service (`ocx service install`): update → reinstall → wrapper contains `start --port <captured>` and proxy stays on that port
- [ ] Force a busy preferred port with explicit `--port` → start fails closed (no ephemeral hop)

### DIFF
--- a/bin/ocx.mjs
+++ b/bin/ocx.mjs
@@ -124,6 +124,19 @@ function runNpmSelfUpdate() {
     return [launcher, "service", "install"];
   }
 
+  // Capture listen target before stop clears runtime-port.json (mirrors GUI/CLI update worker).
+  let bakePort = 10100;
+  try {
+    const rt = JSON.parse(readFileSync(join(configDir(), "runtime-port.json"), "utf8"));
+    if (Number.isFinite(rt?.port) && rt.port > 0) bakePort = Math.trunc(rt.port);
+  } catch { /* fall through to config */ }
+  if (bakePort === 10100) {
+    try {
+      const cfg = JSON.parse(readFileSync(join(configDir(), "config.json"), "utf8"));
+      if (Number.isFinite(cfg?.port) && cfg.port > 0) bakePort = Math.trunc(cfg.port);
+    } catch { /* keep default */ }
+  }
+
   // Never replace package files under a live proxy — stop it first (full `ocx stop`
   // semantics: graceful drain, service stop, native Codex restore). Gate on the service
   // and the runtime-port record too: a service-managed or orphaned proxy can be live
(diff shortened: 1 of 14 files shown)


lidge-jun:

lidge-jun:

Port-pinning approach is solid — PortUnavailableError typing, waitForPortAvailable via setTimeout (not busy loop), and fail-closed concurrent bind handling are all good.

### Blocker

1. **`config.port=0` no longer means ephemeral soft-start** — `src/cli/index.ts:97` converts zero to 10100, but `src/service.ts` explicitly documents zero as interactive ephemeral mode. After `ocx start` with `port:0`, the proxy starts on 10100 instead of an ephemeral port. Only service/update hard-pin paths should normalize zero; preserve it for soft starts.

### Medium concern
- npm launcher stale-state validation (`bin/ocx.mjs:134`) uses `process.kill(pid, 0)` but doesn't verify the PID is actually OpenCodex. A stale record whose PID was reused by another process can bake a wrong port.

Please retarget to `dev` — see [Contributing guide](https://lidge-jun.github.io/opencodex/contributing/). 🙏

lidge-jun:
Absorbed onto dev as e84e3aa2 — cherry-picked all 3 commits with a fix for the port=0 ephemeral regression (restored nullish coalescing so config.port=0 stays ephemeral, hard-pin only fires for explicit --port > 0). Tests pass (73/73). Completes #152. Thank you for the thorough implementation! 🙏

Co-authored-by: Wibias

PR #197, opened 2026-07-20
fix(alibaba): preserve reasoning_content for qwen3.8-max-preview

alexanderxcmo:
## Problem

`qwen3.8-max-preview` is missing from `preserveReasoningContentModels` in the `alibaba-token-plan` registry entry. OpenCodex drops `reasoning_content` from assistant history between agent turns, causing the model to lose its thinking chain.

Alibaba documents `preserve_thinking=true` as the default for Qwen 3.8. The server expects prior `reasoning_content` fields to be replayed verbatim. Without them, multi-tool debugging and autonomous tasks degrade silently (no API error, just weaker results).

See issue #TBD for full details.

## Fix

One-line registry change: add `"qwen3.8-max-preview"` to `preserveReasoningContentModels` in the `alibaba-token-plan` entry.

## Scope

Only `qwen3.8-max-preview` is added. Older Qwen models (3.7, 3.6) default to `preserve_thinking=false` and would need additional request-body changes — out of scope here.

## Tests

3 new regression tests in `tests/qwen38-preserve-reasoning.test.ts`:

1. **Replay verification**: builds a multi-turn request with thinking + tool calls, confirms `reasoning_content` appears in all prior assistant messages and is NOT merged into `content`
2. **Negative control**: same request with a model outside the preserve list — confirms `reasoning_content` is absent
3. **Registry assertion**: confirms `qwen3.8-max-preview` is in the `alibaba-token-plan` preserve list

All 3 pass. `tsc --noEmit` clean.

### DIFF
--- a/src/providers/registry.ts
+++ b/src/providers/registry.ts
@@ -707,7 +707,7 @@ export const PROVIDER_REGISTRY: readonly ProviderRegistryEntry[] = [
     },
     modelReasoningEffortMap: { "deepseek-v4-pro": DEEPSEEK_THINKING_REASONING_MAP },
     thinkingBudgetModels: ALIBABA_TOKEN_PLAN_QWEN_MODELS,
-    preserveReasoningContentModels: ["glm-5.2", "deepseek-v4-pro"],
+    preserveReasoningContentModels: ["glm-5.2", "deepseek-v4-pro", "qwen3.8-max-preview"],
   },
   // NEEDS_HUMAN 2026-07-10: kept for config compatibility, but this is a dashboard URL,
   // no /models endpoint is documented, and tools are silently ignored upstream per docs.parallel.ai.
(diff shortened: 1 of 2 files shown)


lidge-jun:

lidge-jun:

Thanks for the contribution! The model ID and registry placement are correct.

### Blocker

1. **`preserve_thinking: true` is not sent in the wire request** — `preserveReasoningContentModels` only replays `reasoning_content` in the message history JSON, but Alibaba's API docs state that `preserve_thinking=true` must also be included in the request body for the upstream to actually use the replayed reasoning. Without it, the API ignores the `reasoning_content` field entirely.

   - Location: `src/adapters/openai-chat.ts` request body builder (~line 465-498) has no `preserve_thinking` injection
   - The test only verifies local JSON serialization, not the wire request

**Suggested fix:** Add a provider-level flag (e.g. `sendPreserveThinking`) to the registry entry and inject `preserve_thinking: true` in the openai-chat adapter's `buildRequest` for matching models. Test the full registry→router→adapter path.

Also: please retarget this PR to `dev` — see the [Contributing guide](https://lidge-jun.github.io/opencodex/contributing/). 🙏

lidge-jun:
Absorbed onto dev as def60bce — added qwen3.8-max-preview to preserveReasoningContentModels with your test file. Verified via Alibaba docs that this model defaults preserve_thinking=true, so the registry-only change is the complete fix. Thank you! 🙏

Co-authored-by: alexanderxcmo

Issue #186, opened 2026-07-20
[Bug]: Frequent 502 upstream_server_error after the first failure in a conversation.


LimLex:
### Summary

I frequently encounter 502 errors when using OCX with Codex.
**Error Details**
Model: gpt-5.6-sol
Provider: openai
Error: upstream_server_error
Request duration: 13,465 ms
Provider unreachable: The socket connection was closed unexpectedly.
For more information, pass `verbose: true` in the second argument to fetch()
Observed Behavior
The issue appears to be related to individual Codex sessions. Once this error occurs for the first time in a particular Codex session, subsequent requests within that same Codex session tend to encounter the same error much more frequently.
I am confident that this is not caused by my local network connection. When I stop using OCX, Codex returns to normal and the connectivity issue disappears.
This suggests that OCX may not be correctly recovering the upstream connection after the first socket failure within a Codex session.

### Reproduction

1.Start 5 Codex sessions simultaneously.
2.Let all sessions work normally.
3.Wait until one of the sessions encounters the first 502 upstream_server_error.
4.Continue sending requests in that same Codex session.
5.Observe that the same session starts encountering 502 errors much more frequently than before.

### Logs and screenshots

```shell

```

### Area

Proxy runtime

### Config shape

```json

```

### Checks

- [x] I searched existing issues and docs first.
- [x] I removed secrets, tokens, and personal data from logs and screenshots.

Wibias:
Can u please post the full logs? Thanks.

LimLex:
> Can u please post the full logs? Thanks.
<img width="1597" height="704" alt="Image" src="https://github.com/user-attachments/assets/f95f7730-213f-4eed-9e25-7ce871309c22" />
Unfortunately, I had just run `ocx stop` so I could continue working in Codex. When I later tried to restart it with `ocx start`, I found that all of the logs from the previous OCX session had disappeared. This may be a separate bug.
From what I remember, the failed requests were all reported as `upstream_server_error`, with the following message:

```text
Provider unreachable: The socket connection was closed unexpectedly. For more information, pass `verbose: true` in the second argument to fetch()
```

I hope this information is still useful.

One additional detail: I was using two OpenAI accounts at the time—one Pro 20x account and one Pro 5x account—with account-pool round-robin mode enabled.

I had five Codex sessions running concurrently. The errors only occurred in some of the sessions, not all of them.

[...]

PR #205, opened 2026-07-21
fix(codex): rework soft-avoid pool affinity with semantic terminal status

Wibias:
## Summary

Rework the transient-failure soft-avoid mechanism to address the two fundamental design issues identified in #194:

### Blocker 1: Non-transient failures no longer trigger soft avoidance

The terminal outcome recorder now uses the semantic HTTP status from `logCtx.terminalHttpStatus` (derived from the SSE error payload via `httpStatusFromTerminalError`) instead of collapsing every non-completed terminal to 502. A `400 invalid_request_error` no longer soft-avoids the account or rebinds threads — only genuine transport/timeout/5xx outcomes trigger transient health recording.

### Blocker 2: Late failures can no longer delete a newer healthy affinity

The `threadId` meta now guards `threadAccountMap.delete` with an ownership check (`bound?.accountId === accountId`), preventing the race: T→A, A fails, T→B, late A failure must not delete B's mapping.

### Additional improvements

- **Transport-level `keepalive: false`** via `applyUpstreamRecoveryInit` for connection-reset retries (Bun has historically ignored the `Connection: close` header alone — oven-sh/bun#20492)
- **Compact endpoint** (`/v1/responses/compact`) now uses the same transient-retry + recovery-init + pool-outcome recording as `/v1/responses`
- **Failover streak clears all affinities** for the failing account (matching 429 behavior)
- **Soft-avoid extends** on repeated transient failures and clears on 2xx (hard quota cooldown intentionally survives)
- **Affinity reuse checks `shouldFailover`** to catch post-avoid residual streaks

## Test plan

- [x] `bun x tsc --noEmit` — clean
- [x] `bun test ./tests/codex-routing.test.ts` — 51 pass, 0 fail (7 new tests for soft-avoid + race-safe affinity)
- [x] `bun test ./tests/upstream-retry.test.ts` — 14 pass
- [x] `bun test ./tests/responses-parser.test.ts` — 19 pass
- [x] `bun test ./tests/server-auth.test.ts` — 58 pass

Closes https://github.com/lidge-jun/opencodex/issues/186

### DIFF
--- a/src/codex/routing.ts
+++ b/src/codex/routing.ts
@@ -27,12 +27,21 @@ type CodexUpstreamHealth = {
   consecutiveFailures: number;
   lastFailureStatus?: number;
   lastFailureAt?: number;
+  /** Hard cooldown (quota 429). Survives a later 2xx; blocks auth + selection. */
   cooldownUntil?: number;
+  /**
+   * Soft avoid after connect_error / timeout / transient 5xx. Cleared on 2xx.
+   * Blocks pool selection + thread affinity reuse so a sticky session can leave a
+   * flaky account without throwing CodexAccountCooldownError (hard-only).
+   */
+  softAvoidUntil?: number;
 };
 
 const CODEX_DEFAULT_QUOTA_COOLDOWN_MS = 60_000;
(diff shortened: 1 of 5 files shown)


Wibias:
Failing Checks are unrelated to this PR.

PR #214, opened 2026-07-21
feat: rename combos and expose them under a custom public model name

eachann1024:
## Summary

Combos no longer have to be addressed as `combo/<id>`. This PR adds two related capabilities: a combo can be renamed from the dashboard, and each combo gains an optional public model name (alias) that replaces the default `combo/<id>` slug — including bare names with no `combo/` prefix at all, such as `deepseek-v4-flash`.

This directly covers the requested use case: environments that mandate a specific model name can now be served by a combo answering to exactly that name. A client requesting `deepseek-v4-flash` enters the full combo failover loop, and the alias is what appears in the Codex catalog, `/v1/models`, and the picker — no combo prefix is shown anywhere.

## Motivation

Today a combo's public identity is hardcoded to `combo/<id>`, and the id itself is immutable after creation. That is limiting in two ways:

- Teams whose tooling or company policy requires a fixed model name cannot route those requests through a combo, because clients must literally ask for `combo/<id>`.
- Renaming a combo means deleting and recreating it, losing any `disabledModels` / `subagentModels` references.

## Features Added

- Rename a combo inline from the Combos workspace: the combo id is editable in the Config tab and saved atomically via a new `renameFrom` field on `PUT /api/combos`.
- Optional public model name per combo, editable in the same place. Leave it empty to keep the default `combo/<id>`, or set any valid name — bare (`deepseek-v4-flash`) or namespaced (`vendor/model`).
- Alias requests get the complete combo behavior: target selection, failover hops, cooldowns, and per-attempt usage, with the requested alias preserved as the log identity (provider stays `combo`).
- The alias is the combo's public face everywhere: Codex catalog entries, both `/v1/models` shapes, the picker, and the dashboard rail.
- Validation on both ends: alias shape, reserved `combo/` namespace, OpenAI native-family bare names (which cannot be shadowed honestly), and cross-combo uniqueness — enforced by the management API and at config load time.
- Dashboard copy localized in English, Simplified Chinese (zh-Hans), Korean, and German.

## Changes

- `src/combos/`:
  - New `resolveComboId` (canonical `combo/<id>` first, then exact alias match), `comboPublicModelId`, and alias validation shared by config load and the management API; `OcxComboConfig` gains `alias`.
  - `comboIdFromRawBody` is config-aware so alias requests enter the combo failover loop.
- `src/server/responses.ts`, `src/server/request-log.ts`:
  - Combo identity is threaded structurally (`comboId` on the log context) instead of sniffing the `combo/` prefix, so aliased requests keep per-attempt usage aggregation with the requested alias as the top-level model.
- `src/server/management-api.ts`:
  - `PUT /api/combos` accepts `renameFrom` for atomic renames; migrates `disabledModels` / `subagentModels` references from both the old public slug and the old canonical slug (deduped, order-preserving); clears selection state and cooldowns on both ids; `GET` returns the public model name.
- `src/codex/catalog.ts`, `src/server/index.ts`:
  - Combo catalog rows are emitted under the public slug (bare aliases receive routed-model treatment, not native); masquerade collisions shadow provider rows with a warning; stale alias rows are garbage-collected via an `owned_by: "combo"` marker that survives strict normalization; `disabledModels` entries naming an alias hide the combo.
- `src/config.ts`:
  - Persisted combos are validated with the full combos map so duplicate aliases are rejected at load time.
- `gui/` (Combos workspace):
  - Editable combo id (rename on save, detail view follows the new id) and a public model name field with hints; delete dialog and rail show the public name; view-model validation mirrors the backend.
- `tests/`:
  - New coverage for alias resolution/validation, rename migration and state cleanup, alias failover end-to-end, catalog exposure and stale cleanup, disable-by-alias, and GUI view-model rules.

## Benefits

- A combo can impersonate a mandated model id with zero client-side changes — useful wherever a fixed model name is required by policy or tooling.
- Renaming is a safe, atomic operation instead of delete-and-recreate, and existing model references follow along.
- Public identity stays consistent across routing, catalog, logs, and the dashboard, so there is no confusing split between "config name" and "visible name".

## Testing

- `bun test tests/combos.test.ts tests/combo-workspace-data.test.ts tests/codex-catalog.test.ts tests/server-combo-failover-e2e.test.ts tests/request-log.test.ts` — 202 pass, 0 fail.
- Full `bun run test` — 3,320 pass; the same 5 pre-existing provider-management tests fail on `origin/main` in the same environment.
- `bun run lint:gui`, `bun run privacy:scan`, `bun run doctor:gui`, and the GUI production build all pass.

Suggested manual check: create a combo, set its public model name to `deepseek-v4-flash`, then request that exact model id — it should fail over across the combo's targets and appear under the alias in `/v1/models`.

## Example Configuration

```json
{
  "combos": {
    "free": {
      "alias": "deepseek-v4-flash",
      "targets": [
        { "provider": "provider-a", "model": "m1" },
        { "provider": "provider-b", "model": "m2" }
      ]
    }
  }
}
```

Clients then request `deepseek-v4-flash` directly; `combo/free` keeps working as the canonical alias for backward compatibility.

---

Thank you for maintaining opencodex — the combo design made this a pleasure to build on. Happy to adjust naming, validation strictness, or the dashboard copy if you would like a different direction.


Wibias:
@eachann1024 Please retarged into dev.

lidge-jun:
Thank you @eachann1024 for this thorough implementation — the atomic rename migration logic is particularly well done, and the combo alias concept is exactly the direction we want.

Two gaps prevent merging as-is: (1) alias collision handling only covers the Codex catalog shape (with `client_version`); ordinary `/v1/models` can still emit duplicate provider/combo IDs; (2) the PR adds EN i18n keys without RU counterparts, which breaks the current `gui build`.

We'll adopt the design direction and build it on dev with these fixes. The PR is currently CONFLICTING with dev as well. Closing for now — your work will be credited when we ship the feature.

lidge-jun:
## Detailed Review — PR #214

### What we reviewed

We ran an adversarial code review (Sol subagent, gpt-5.6-sol) that read the full `gh pr diff 214` against current `dev`. The review examined alias resolution, collision handling across both model-list shapes, catalog lifecycle, rename atomicity, GUI integration, i18n parity, and test coverage.

### Findings

**The feature design is strong.** The combo alias concept — letting a combo answer to a bare name like `deepseek-v4-flash` instead of only `combo/<id>` — is exactly the direction we want. Several implementation details are particularly well done:

- **Atomic rename migration** (`PUT /api/combos` with `renameFrom`): migrates `disabledModels`/`subagentModels` references from both old public and old canonical slugs, deduped and order-preserving. This is careful work.
- **Alias validation**: shape check, reserved `combo/` namespace protection, OpenAI native-family bare-name shadowing prevention, cross-combo uniqueness enforcement at both API and config-load time.
- **Structural combo identity threading** (`comboId` on log context): cleaner than prefix-sniffing for aliased requests.

**Two gaps prevent merging as-is:**

1. **`/v1/models` alias duplication** — Alias collision handling at `src/server/index.ts:290-306` only applies `buildCatalogEntries` collision logic when the request includes `client_version` (Codex catalog shape). The ordinary `/v1/models` branch maps every `goOrdered` row directly without dedup, so an alias that shadows a provider model ID will produce duplicate entries in the standard OpenAI-compatible models list. The E2E test suite (`server-combo-failover-e2e.test.ts`) has no `/v1/models` alias test — only the catalog-building path is covered.

2. **Missing RU locale keys** — The PR adds EN i18n keys at `gui/src/i18n/en.ts:902,941-975` without corresponding RU entries. Since `ru.ts` is typed as `Record<TKey, string>`, this breaks `cd gui && bun run build` on current `dev` (which has 956/956 EN/RU parity after PR #211 merged). The PR would need to add RU translations for every new EN key.

**Additional observations:**
- The `owned_by: "combo"` marker for stale-alias garbage collection is a new catalog field that survives strict normalization — this is clever but should be documented in the catalog schema.
- The PR is currently CONFLICTING with `dev`, likely due to recent merges in `registry.ts` and `types.ts`.

### What we plan to adopt

The design direction (alias resolution, atomic rename, structural identity threading) will be built on `dev` with these fixes:
- Dedup in both `/v1/models` shapes
- RU locale parity
- Tests covering the ordinary models-list path

Your implementation will be referenced and credited. Thank you @eachann1024 — this is a substantial, well-structured contribution and we appreciate the care you put into the migration logic and validation.

PR #219, opened 2026-07-21
fix(winsw): detect sc.exe error 1060 under Bun truncation and localized output

### LINKED ISSUE #216: Windows (non-English / Bun): ocx service & ocx update abort — probeScmRegistration fails to detect ERROR_SERVICE_DOES_NOT_EXIST (1060)
## Summary

On non-English Windows (and generally under the bundled **Bun** runtime), `ocx service` aborts during install with a false "native service still exists" error, even when the `opencodex-proxy-native` SCM service does **not** exist. The same faulty detection path is reused by `ocx update` (its service-restart step), so updating a service-managed proxy is affected too.

Root cause: `probeScmRegistration()` in `src/lib/winsw.ts` cannot recognize `sc.exe`'s `ERROR_SERVICE_DOES_NOT_EXIST` (1060) result, so `statusWinswRaw()` returns `"unknown"` instead of `"nonexistent"`, and `installWindows()` refuses to proceed.

## Environment

- OS: Windows 11 (**pt-BR** locale), 10.0.26200
- opencodex: 2.7.28 (reproduces the same code on 2.7.30)
- Runtime: bundled Bun 1.3.14
- Install: global via fnm-managed Node

## Symptom

```
$ ocx service
error: Cannot remove the native service before switching to Task Scheduler:
Cannot verify the native service registration (sc.exe query failed).
Uninstall aborted; check 'sc query opencodex-proxy-native' and retry.
    at installWindows (.../src/service.ts:483)
```

…yet the service genuinely does not exist:

```
$ sc query opencodex-proxy-native
[SC] EnumQueryServicesStatus:OpenService FALHA 1060:
O serviço especificado não existe como serviço instalado.
```

There is no `~/.opencodex/winsw/` directory and no `opencodex-proxy-native` SCM registration.

## Root cause

`statusWinswRaw()` → `probeScmRegistration()` (`src/lib/winsw.ts`) classifies the `sc query` failure. Only a confirmed 1060 is allowed to prove absence; anything else becomes `"error"` → `statusWinswRaw()` returns `"unknown"` → lifecycle callers assume a service may exist. The 1060 check has **two portability gaps**:

1. **Exit-code truncation under Bun.** `sc.exe` exits with code `1060`, but Bun on Windows truncates the process exit status to 8 bits, so the thrown error reports `e.status === 36` (`1060 % 256 = 36`). The guard `e.status === 1060` is therefore never true under Bun.
2. **English-only text match.** The fallback `/FAILED 1060/i` does not match localized `sc.exe` output. On pt-BR the line is `... FALHA 1060: ...`, so the regex misses it.

With both signals failing, `probeScmRegistration()` returns `"error"` (fail-closed), which is the intended behavior for a genuinely unreadable SCM — but here the SCM answered clearly, we just failed to parse a localized/truncated "does not exist".

### Reproduction (Bun on Windows)

```js
const { execFileSync } = require("node:child_process");
const { join } = require("node:path");
const sc = join(process.env.SystemRoot || "C\\Windows", "System32", "sc.exe");
try {
  execFileSync(sc, ["query", "opencodex-proxy-native"], { encoding: "utf8", stdio: ["ignore","pipe","pipe"] });
} catch (e) {
  const text = [e.stderr, e.stdout, e.message].map(v => typeof v === "string" ? v : "").join("\n");
  console.log("e.status =", e.status);              // -> 36  (NOT 1060, truncated by Bun)
  console.log("/FAILED 1060/i:", /FAILED 1060/i.test(text)); // -> false (pt-BR says "FALHA 1060")
  console.log("/\\b1060\\b/:", /\b1060\b/.test(text));       // -> true
}
```

Output on a pt-BR host:
```
e.status = 36
/FAILED 1060/i: false
/\b1060\b/: true
```

## Proposed fix

Match the **locale-independent numeric code** anywhere in the captured `sc.exe` output. The string `1060` is present regardless of UI language, and this is immune to the Bun exit-code truncation. In `src/lib/winsw.ts`, `probeScmRegistration()`:

```diff
-    if (e.status === 1060 || /FAILED 1060/i.test(text)) return false;
+    // 1060 = ERROR_SERVICE_DOES_NOT_EXIST. Two portability gaps make the original
+    // check fail:
+    //   * Bun on Windows truncates the exit code to 8 bits, so a 1060 exit arrives
+    //     as `e.status === 36` (1060 % 256) and the numeric guard never matches.
+    //   * sc.exe localizes its output (e.g. pt-BR "FALHA 1060"), so /FAILED 1060/i
+    //     never matches on non-English hosts.
+    // The numeric code itself is locale-independent and survives truncation, so
+    // match it anywhere in the captured stdout/stderr/message.
+    if (e.status === 1060 || /\b1060\b/.test(text)) return false;
```

This keeps the fail-closed behavior for genuinely unreadable results (access denied, `sc.exe` missing, etc.) while correctly proving absence when the SCM reports 1060 in any locale.

## Impact / notes

- Blocks `ocx service` install on non-English Windows even with an elevated shell (the abort happens in the SCM-detection step, before `schtasks`).
- Also affects `ocx update` when a background service is installed: its restart step calls `ocx service install`, which hits the same abort.
- After applying the one-line change locally, `ocx service` proceeds to `schtasks /create` and installs the Task Scheduler backend as expected.

Happy to open a PR with the change if preferred.

Wibias:
## Problem

`probeScmRegistration()` missed `ERROR_SERVICE_DOES_NOT_EXIST` (1060) on non-English Windows under Bun, so `ocx service` / `ocx update` aborted with a false "possibly installed" verdict. Two compounding traps:

1. **Bun truncates exit codes to 8 bits on Windows** — `sc.exe` exit 1060 arrives as `status 36` (1060 % 256), so the numeric guard never matches.
2. **sc.exe localizes the `FAILED` prefix** — pt-BR prints `FALHA 1060`, de-DE `FEHLER 1060`, etc., so `/FAILED 1060/i` only matches English output.

The probe then returned `"error"` instead of confirmed absence, and lifecycle callers correctly fail closed on `"error"` — blocking installs and updates entirely.

## Fix

This implements the solution proposed in the issue, which is the right one: the numeric Win32 code `1060` is locale-independent and always printed by sc.exe, so match it anywhere in the captured streams with word boundaries:

```ts
if (e.status === 1060 || /\b1060\b/.test(text)) return false;
```

- `e.status === 1060` kept for runtimes without truncation.
- `\b` word boundaries guard against longer numbers (`10600`) — verified by test.
- Bare truncated `status 36` with no `1060` in any stream still returns `"error"` (fail-closed preserved) — verified by test.

Alternatives considered and rejected: special-casing `status === 36` alone (could be a different error), and matching localized `FAILED` words (unbounded list; the numeric code is the stable signal).

## Tests

New test case `SCM probe detects 1060 under Bun exit-code truncation and localized sc.exe output` in `tests/winsw.test.ts` covers: Bun-truncated `status: 36` + `FAILED 1060`, pt-BR `FALHA 1060`, de-DE `FEHLER 1060`, `10600` word-boundary guard → `"error"`, and bare `status 36` → `"error"`.

Verified: `bun test --isolate ./tests/winsw.test.ts` (22 pass), `bun test --isolate ./tests/service.test.ts` (30 pass), `bun x tsc --noEmit` clean.

Closes #216

### DIFF
--- a/src/lib/winsw.ts
+++ b/src/lib/winsw.ts
@@ -223,7 +223,12 @@ export function probeScmRegistration(run: () => string = queryScmForService): bo
     const text = [e.stderr, e.stdout, e.message]
       .map(v => (typeof v === "string" ? v : ""))
       .join("\n");
-    if (e.status === 1060 || /FAILED 1060/i.test(text)) return false;
+    // Two traps hide the 1060 signal:
+    // 1. Bun on Windows truncates exit codes to 8 bits, so 1060 arrives as status 36.
+    // 2. sc.exe localizes the "FAILED" prefix (pt-BR "FALHA 1060", de-DE "FEHLER 1060", ...).
+    // The numeric Win32 code itself is locale-independent and always printed, so match it
+    // with word boundaries (guards against 10600 etc.) anywhere in the captured output.
+    if (e.status === 1060 || /\b1060\b/.test(text)) return false;
     return "error";
   }
 }
--- a/src/lib/winsw.ts
+++ b/src/lib/winsw.ts
@@ -219,9 +219,13 @@ export function probeScmRegistration(run: () => string = queryScmForService): bo
   } catch (err) {
     const e = err as { status?: number | null; stderr?: string | Buffer | null; stdout?: string | Buffer | null; message?: string };
     // sc.exe does not reliably channel the 1060 line — it can land on stderr OR stdout
-    // depending on the host — so scan every captured stream and the error message.
-    const text = [e.stderr, e.stdout, e.message]
-      .map(v => (typeof v === "string" ? v : ""))
+    // depending on the host — so scan both captured streams.
+    // e.message is intentionally excluded: it is a Node wrapper string that can embed
+    // arbitrary content (paths, spawn messages) that may contain "1060" coincidentally,
+    // which would produce a false confirmed-absence on an unrelated failure.
+    // Buffer values are decoded via .toString() to guard against missing encoding option.
+    const text = [e.stderr, e.stdout]
+      .map(v => (v instanceof Buffer ? v.toString() : typeof v === "string" ? v : ""))
       .join("\n");
     // Two traps hide the 1060 signal:
     // 1. Bun on Windows truncates exit codes to 8 bits, so 1060 arrives as status 36.
--- a/tests/winsw.test.ts
+++ b/tests/winsw.test.ts
@@ -134,6 +134,18 @@ describe("winsw fail-closed lifecycle", () => {
     expect(probeScmRegistration(() => { throw new Error("spawn sc.exe ENOENT"); })).toBe("error");
   });
 
+  test("SCM probe detects 1060 under Bun exit-code truncation and localized sc.exe output", () => {
+    // Bun on Windows truncates exit codes to 8 bits: 1060 arrives as status 36 (#216).
+    expect(probeScmRegistration(() => { const e = new Error("fail") as Error & { status: number; stderr: string }; e.status = 36; e.stderr = "[SC] OpenService FAILED 1060:"; throw e; })).toBe(false);
+    // Localized sc.exe output: the "FAILED" word changes, the numeric code does not.
+    expect(probeScmRegistration(() => { const e = new Error("fail") as Error & { status: number; stderr: string }; e.status = 36; e.stderr = "[SC] OpenService FALHA 1060:"; throw e; })).toBe(false);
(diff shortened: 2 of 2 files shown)


lidge-jun:
v2.7.31의 fbd96c12에서 #216과 #199를 함께 해결해 현재 main, dev, preview에 배포했습니다. 이 PR의 문제 제기와 locale 및 Bun 분석은 최종 수정 검증에 반영됐습니다. 구현 경로가 이미 릴리스된 브랜치와 갈라져 있어 superseded로 닫습니다. PR 커밋과 diff는 기록으로 유지됩니다.

PR #221, opened 2026-07-21
fix(oauth): do not persist needsReauth across process restarts

### LINKED ISSUE #209: [Bug]: Anthropic OAuth account is marked needsReauth after Windows reboot and models disappear
## Summary

On Windows 11, the Anthropic OAuth account remains present after every reboot, but it is marked as needing reauthentication and all Anthropic models disappear from the model list.

The dashboard still shows the Anthropic provider as active and OAuth as logged in, while the account row has a yellow indicator and a **Re-login** button. Removing the account and adding it again immediately restores the Anthropic models, but the same problem returns after the next Windows reboot.

### Expected behavior

The persisted Anthropic refresh token should be used after OpenCodex starts again, and Anthropic models should remain available without removing and re-adding the account.

### Actual behavior

- Provider remains configured and active.
- OAuth status reports logged in.
- The active account is present but has `needsReauth: true`.
- `/api/models` returns zero Anthropic models.
- The provider connectivity test treats Anthropic as not logged in.
- Deleting and reconnecting the account restores the models until the next reboot.

## Reproduction

1. On Windows 11, start OpenCodex with `ocx start`.
2. Open `http://localhost:10100/`.
3. Add an Anthropic account through **Log in with your Claude account**.
4. Confirm that Anthropic models appear and work.
5. Reboot Windows.
6. Run `ocx start` again and open the dashboard.
7. Observe:
   - Anthropic provider is still shown as active/OAuth.
   - The account row has a yellow status dot and a **Re-login** button.
   - Anthropic models are absent from the model list.
8. Delete the Anthropic account and add it again.
9. Observe that the models reappear immediately.

This has happened consistently after reboots in this environment.

## Logs and diagnostics

Current post-reboot state, collected from the local management API and redacted:

```text
opencodex: 2.7.26
codex-cli: 0.144.5
node: v24.15.0

GET /api/oauth/status?provider=anthropic
  loggedIn: true

GET /api/oauth/accounts?provider=anthropic
  accountCount: 1
  activeAccountIdPresent: true
  active: true
  needsReauth: true

GET /api/models
  anthropicModelCount: 0

POST /api/providers/test?name=anthropic
  ok: false
  latencyMs: 0
  error: "static catalog only — upstream not verified (not logged in)"

Redacted auth-store metadata:
  source: local-cli
  hasAccessToken: true
  hasRefreshToken: true
  accessTokenExpired: true
```

Dashboard visual state: Anthropic provider is active, the single account row has a yellow status dot, and the **Re-login** action is shown.

No tokens, account IDs, email addresses, or other personal data are included.

## Area

Provider adapter / Anthropic OAuth token persistence and refresh across process or OS restart.

## Version

`@bitkyc08/opencodex 2.7.26`

## OS

Microsoft Windows 11 Home, version 10.0.26200, 64-bit

## Config shape

```json
{
  "provider": "anthropic",
  "adapter": "anthropic",
  "baseUrl": "https://api.anthropic.com",
  "defaultModel": "claude-sonnet-5",
  "authMode": "oauth",
  "disabled": false
}
```

## Checks

- [x] I searched existing open and closed issues first.
- [x] I removed secrets, tokens, account identifiers, email addresses, and personal data.

Wibias:
## Problem

On Windows (and potentially other platforms), the Anthropic OAuth account was permanently marked `needsReauth: true` after every reboot, hiding all models until the user manually removed and re-added the account.

Root cause: `needsReauth` was persisted to `auth.json` via `mutateStore` and restored on every load through `normalizeAccount`. A transient failure during the first refresh attempt after boot — most commonly Windows not yet having network connectivity when `ocx start` runs — would hit `isTerminalRefreshError`, call `markAccountNeedsReauth(..., true)`, and write that permanently to disk. The account was then skipped by the token guardian, excluded from `/api/models`, and the only recovery was removing and re-adding the account (which writes fresh credentials and clears the flag).

## Fix

Introduce an **in-memory-only** `Set<string>` (`inMemoryNeedsReauth`) keyed by `provider\x00accountId`:

- `normalizeAccount` no longer copies `needsReauth` from the JSON file on disk.
- `getAccountSet` overlays the in-memory markers when returning account lists, so the guardian sweep, catalog fetches, and CLI all see correct state within a session.
- `markAccountNeedsReauth` / `markAccountNeedsReauthIfGeneration` write to the Set.
- `saveCredential`, `saveAccountCredential`, `mergeAccountCredential`, `removeAccount`, `removeCredential` clear the relevant key from the Set on success.

On each fresh process start the Set is empty, so every OAuth account gets one real refresh attempt regardless of what the previous session recorded. A genuinely revoked token still fails terminally and re-marks itself within the session; a transient boot-time error does not survive a restart.

## Behaviour

- Within a session: `needsReauth` works exactly as before — terminal failures mark the account and the guardian skips it.
- Across restarts: the mark is gone. The first refresh attempt after startup either succeeds (valid token) or fails terminally (truly revoked) — no permanent state from a boot-time network hiccup.

## Tests

`tests/oauth-store-multi.test.ts` updated: verifies in-memory visibility, clears on fresh credential save, and confirms `needsReauth` is absent from the raw JSON on disk.

Closes #209


lidge-jun:
v2.7.31의 a626a5b7 및 0de138aa에서 #209를 해결해 main, dev, preview에 배포했습니다. 최종 구현은 generation-safe refresh와 durable refresh intent로 재부팅 및 stale-lock token replay를 함께 처리하는 경로를 택했습니다. 대체 구현으로 superseded되어 닫으며 PR의 분석과 커밋 기록은 유지합니다.

Wibias:
@lidge-jun Please change ur automatic answers to english again.. Do not always want to use the waterfox built in translator, lol.

PR #223, opened 2026-07-21
fix(codex): clamp advertised reasoning efforts to the installed binary's ladder (unbreaks Codex 0.133.0)

Bricol1982:
## Problem

On Codex `0.133.0`, opencodex breaks catalog loading outright:

```
failed to resolve feature override precedence: failed to parse model_catalog_json
path `.../opencodex-catalog.json` as JSON: unknown variant `max`, expected one of
`none`, `minimal`, `low`, `medium`, `high`, `xhigh`
```

Codex then reports **"failed to create task"**. Every catalog resync reproduces it.

## Root cause

opencodex mocks `max`/`ultra` reasoning rungs onto every catalog model (`applyReasoningLevels`, `ensureUltraReasoningLevel`, `ensureGpt56ReasoningLevels`), assuming a Codex build whose `ReasoningEffort` enum accepts them. Codex `0.133.0`'s enum stops at `xhigh`, and codex-rs deserializes the catalog strictly — so the first unknown variant rejects the **entire** catalog file.

## Fix

Derive the reasoning ladder the **installed** binary actually parses from its own bundled catalog (`codex debug models --bundled`) and clamp every emitted entry to it, at both on-wire boundaries:

- `syncCatalogModels` — the on-disk catalog Codex reads
- `/v1/models?client_version` — the catalog shape served to the Codex client

New helpers in `src/codex/catalog.ts`:
- `codexSupportedReasoningEfforts(deps?)` — the binary's supported effort set, or `null` when codex can't be probed
- `clampEntryToCodexSupportedEfforts(entry, supported)` — filters rungs + repairs `default_reasoning_level` to the highest surviving rung
- `clampCatalogModelsToCodexSupport(models, deps?)` — the emission-boundary guard

### Properties

- **Self-adapting.** A newer Codex whose bundled catalog advertises `max`/`ultra` keeps them unchanged; the clamp only removes what the installed binary can't parse.
- **Safe fallback.** When codex isn't probeable, it's a no-op — behavior matches the prior release.
- **Pure builders untouched.** `buildCatalogEntries` / `mergeCatalogEntriesForSync` still emit the mock tiers; the clamp runs only at the write boundaries, so their existing max/ultra unit tests are unaffected.
- **Never empties a ladder** — the universal low/medium/high rungs are always in the supported set.

## Testing

- Full suite: **3308 pass, 0 fail** (6 new regression tests: strip, preserve, no-op, default repair)
- End-to-end against a real Codex `0.133.0`: a catalog polluted with `max`/`ultra` run through `syncCatalogModels` comes out with ladders capped at `xhigh` and no `max`/`ultra` anywhere.

### DIFF
--- a/src/codex/catalog.ts
+++ b/src/codex/catalog.ts
@@ -690,7 +690,13 @@ export function materializeBundledCodexCatalog(path: string, deps: BundledCatalo
 
 function loadCatalogForSync(path: string): RawCatalog | null {
   const bundled = isDefaultCatalogPath(path) ? loadBundledCodexCatalog() : null;
-  if (bundled) return bundled;
+  // Deep-clone the bundled catalog: syncCatalogModels mutates catalog.models IN PLACE (mergeCatalog-
+  // EntriesForSync/ensureUltraReasoningLevel append the mock max/ultra rungs to the native entries),
+  // and loadBundledCodexCatalog hands back the SHARED 60s cache by reference. Returning it directly
+  // poisons that cache — a later reader in the same process (e.g. codexSupportedReasoningEfforts, or
+  // the /v1/models handler) would then see the pristine bundled ladder as low..ultra and stop
+  // clamping. The clone keeps the cache pristine so the effort probe stays trustworthy.
+  if (bundled) return JSON.parse(JSON.stringify(bundled)) as RawCatalog;
   const catalog = readCatalog(path);
   if (catalog && findNativeTemplate(catalog)) return catalog;
   return readCatalog(catalogBackupPathFor(path))
--- a/src/codex/catalog.ts
+++ b/src/codex/catalog.ts
@@ -830,6 +830,85 @@ function ensureUltraReasoningLevel(entry: RawEntry): void {
 }
 
 /**
+ * Reasoning-effort labels the INSTALLED Codex binary can actually parse, read from its OWN
+ * bundled catalog (`codex debug models --bundled`). codex-rs deserializes every catalog reasoning
+ * level into a strict `ReasoningEffort` enum and rejects the ENTIRE catalog file on the first
+ * unknown variant — Codex 0.133.0 knows none..xhigh but NOT the `max`/`ultra` rungs opencodex mocks
+ * onto every model, so an unguarded catalog fails to load ("failed to create task ... unknown
+ * variant `max`"). The bundled catalog — not the on-disk file, which a prior sync may already have
+ * polluted with max/ultra — is the authoritative record of what the binary accepts. Returns null
+ * when the binary can't be probed (no codex on PATH), so callers keep the prior behavior instead of
+ * clamping against an empty set.
+ */
+export function codexSupportedReasoningEfforts(deps: BundledCatalogDeps = {}): Set<string> | null {
+  const bundled = loadBundledCodexCatalog(deps);
+  if (!bundled) return null;
+  const efforts = new Set<string>();
+  for (const model of bundled.models ?? []) {
+    if (typeof model.slug !== "string" || model.slug.includes("/")) continue;
+    const levels = Array.isArray(model.supported_reasoning_levels) ? model.supported_reasoning_levels : [];
+    for (const level of levels) {
+      const effort = (level as { effort?: unknown })?.effort;
+      if (typeof effort === "string") efforts.add(effort);
+    }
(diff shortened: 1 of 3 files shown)


Wibias:
Please retarget to dev branch.

Bricol1982:
Follow-up commit `141badc`: fixed a cache-poisoning hazard the clamp exposed. `loadCatalogForSync` returned the shared 60s bundled-catalog cache by reference, and `syncCatalogModels` mutates `catalog.models` in place (the mock max/ultra rungs get appended to natives) — poisoning the cache the effort probe reads next, so nothing got clamped. Now deep-clones the bundled catalog before the mutating sync path. Verified end-to-end on a real Codex 0.133.0: `ocx sync` and `/v1/models?client_version` both emit ladders capped at `xhigh`. Full suite 3308 pass / 0 fail.

Wibias:
> Please retarget to dev branch.

@Bricol1982

lidge-jun:
Core approach adopted on dev with safe-fallback improvements (all-unsupported ladder now falls back to universal [low,medium,high] instead of preserving unsupported rungs). Thank you @Bricol1982 for the excellent design — the self-adapting clamp and deep-clone cache-poisoning fix are both preserved. Commit: 504af734ac1a4be0b8fef1a530d2937206150f40

lidge-jun:
## Detailed Review — PR #223

### What we reviewed

We ran an adversarial code review (Sol subagent, gpt-5.6-sol) that read the full `gh pr diff 223` against the current `dev` branch. The review focused on correctness of the clamp logic, edge-case safety, cache-poisoning risk, and integration boundaries.

### Findings

**The core approach is excellent.** Reading the installed Codex binary's bundled catalog via `codex debug models --bundled` to derive the supported reasoning-effort set is the right design. It is self-adapting (a newer Codex that supports `max`/`ultra` keeps them unchanged) and safe when the binary is absent (no-op fallback).

**Three issues we addressed in our adoption:**

1. **Cache poisoning via `loadCatalogForSync`** — Your PR correctly identified that `loadBundledCodexCatalog()` returns a shared cached reference, and `syncCatalogModels` mutates `catalog.models` in place via `ensureUltraReasoningLevel`. Without a deep clone, the cache gets poisoned and later readers see max/ultra in what should be the pristine bundled ladder. Your `JSON.parse(JSON.stringify(bundled))` fix is exactly right — we preserved it.

2. **All-unsupported ladder fallback** — The original implementation preserves an entirely unsupported ladder when `kept.length === 0` (i.e., every rung in the entry is unknown to the installed binary). This is unsafe because the whole point of the clamp is to prevent unknown variants from reaching `codex-rs`. We changed this to fall back to the universal `[low, medium, high]` ladder, which is always in the supported set.

3. **Default effort repair** — When the original default (e.g., `max`) is stripped, `clampedDefaultEffort` picks the highest surviving rung at or below the original. In the edge case where no rungs survive (before our safe-fallback fix), the original returned the unsupported default unchanged. We changed this to return `medium` as the conservative fallback.

**What we adopted verbatim from your PR:**
- `codexSupportedReasoningEfforts()` — effort-set extraction from bundled catalog
- `clampEntryToCodexSupportedEfforts()` — per-entry rung filtering (with safe-fallback addition)
- `clampCatalogModelsToCodexSupport()` — emission-boundary guard
- `loadCatalogForSync` deep-clone fix
- The emission-boundary placement in `syncCatalogModels`

**What we added:**
- Conservative fallback when ALL efforts are unsupported (universal `[low, medium, high]`)
- Safe default when `clampedDefaultEffort` has no survivors (`medium`)
- 5 regression tests covering: strip, preserve, no-op, safe fallback, default repair

### Implementation on dev

**Commit:** [`8b761e5e`](https://github.com/lidge-jun/opencodex/commit/8b761e5e) — `fix(codex): clamp catalog reasoning efforts to installed binary's ladder`
**Co-authored-by:** `Bricol <Bricol1982@users.noreply.github.com>`
**Tests:** 75 pass / 0 fail in `codex-catalog.test.ts`, typecheck clean

### Why we closed instead of merging

The PR targeted `main` while our workflow requires `dev`-first integration. Additionally, the safe-fallback changes were substantive enough that a clean implementation on `dev` was faster than merging and patching. Your design, code structure, and the cache-poisoning discovery are all preserved with full credit.

Excellent work — this is a high-quality contribution that shipped same-day. Thank you @Bricol1982.

PR #226, opened 2026-07-21
Redesign provider setup and add Codex restart

HaydernCenterpoint:
## What

- Split the Providers page into side-by-side OAuth and API Provider sections.
- Show only connected account providers in the OAuth section; move discovery into a searchable two-column picker that excludes connected providers.
- Add an API-key-only provider picker, including providers whose primary mode is OAuth but explicitly supports key auth.
- Replace the redundant header-level Add Provider button with Restart Codex and show a restart recommendation after a new provider is connected.
- Add best-effort model synchronization after provider/account changes and a same-origin management endpoint that restarts Codex Desktop on Windows or macOS.
- Add localized copy for English, German, Korean, and Chinese plus focused regression coverage.

## Why

The existing page mixed account and API-key setup in one surface and rendered every account-capable provider inline, even when it was not connected. That made configured state hard to scan and made provider discovery noisy. Provider changes could also leave Codex with a stale model catalog until users manually synchronized and restarted the app.

## Root cause

The catalog treated a provider's primary auth mode as its only setup capability, while the page used the full account-provider list as both configured state and discovery UI. There was also no dedicated local Codex Desktop restart action, and post-login refreshes updated only part of the Codex integration state.

## Impact

- No configuration migration is required.
- Existing provider detail and JSON editing flows remain available.
- The restart endpoint is protected by the existing management API origin checks and supports only Windows/macOS.
- This PR is intentionally isolated from unrelated Dashboard, CLI, model-catalog, and auto-sync work in the original worktree.

## Checks

- `bun test tests/codex-app-restart.test.ts tests/provider-workspace-auth.test.ts tests/provider-workspace-data.test.ts tests/provider-registry-parity.test.ts` — 93 passed.
- `cd gui && bun run lint` — passed.
- `cd gui && bun run build` — passed (existing chunk-size warning only).
- Full suite: 3,301 passed, 4 skipped. Four unrelated Windows failures remained: two 5-second timeout cases (the LRU case passed with a 15-second timeout; usage-debug passed when rerun alone) and two Cursor native-exec tests whose fixture invokes the unavailable POSIX `printf` command on Windows.


Wibias:
@HaydernCenterpoint Try to use the workspace View please :)

lidge-jun:
Thank you @HaydernCenterpoint for this comprehensive provider UI redesign — the OAuth/API split and restart endpoint concept are valuable directions.

A few items need work before this can land: (1) RU locale key parity is missing (new EN keys without RU counterparts break `gui build`); (2) the restart endpoint treats a scheduled kill/relaunch as immediate success without async error handling; (3) integration tests only check source strings, not rendered UI behavior; (4) this should route through dev rather than targeting main directly.

We'll incorporate the design direction in a future iteration. Closing for now — thank you for the work.

lidge-jun:
## Detailed Review — PR #226

### What we reviewed

We ran an adversarial code review (Sol subagent, gpt-5.6-sol) that read the full `gh pr diff 226` against current `dev`. The review examined the UI redesign scope, the restart endpoint security model, localization parity, test coverage methodology, and cross-PR interactions.

### Findings

**The design direction is valuable.** Splitting the Providers page into side-by-side OAuth and API sections addresses a real UX problem — the current page mixes configured state with discovery UI, making it hard to scan what's actually connected. Moving discovery into a searchable two-column picker that excludes already-connected providers is a clean separation. The restart-recommendation flow after provider changes is also a good UX improvement.

**Four items need work before this can land:**

1. **RU locale key parity** — The PR adds new EN i18n keys at `gui/src/i18n/en.ts:132-181` without corresponding RU entries. Since `ru.ts` is typed as `Record<TKey, string>`, this breaks `cd gui && bun run build` on current `dev` (which has 956/956 EN/RU parity). Every new EN key needs a RU translation.

2. **Restart endpoint async error handling** — `src/codex/restart-app.ts:50-67` schedules a kill/relaunch but the management API returns success immediately, before the process operation actually runs. If the restart fails (e.g., binary not found, permission error, port conflict), the client has no way to learn about it. The endpoint should either block until the restart completes or return a status-polling URL.

3. **Source-string-only integration tests** — `tests/provider-workspace-auth.test.ts:184-232` searches for strings in the source code rather than testing rendered component behavior. This means the tests pass even if the components don't render correctly, event handlers are wired wrong, or the layout breaks. We recommend Vitest component tests or at minimum snapshot tests of the rendered output.

4. **Branch targeting** — The PR targets `main` directly. Our workflow requires all changes to land on `dev` first, go through the full test matrix (including cross-platform CI), and promote to `main` only through a release gate.

**Cross-PR interactions:**
- Conflicts with #214 (combo workspace GUI changes) and #211 (RU locale, now merged)
- The model-sync logic after provider changes (`syncCatalogModels`) may interact with the reasoning-effort clamp we just landed from #223 — needs integration verification
- The restart endpoint's management API origin checks are sound, but the macOS/Windows platform assumptions (process kill/relaunch mechanics) need platform-specific tests

### What we plan to adopt

The OAuth/API split design, the searchable provider picker, and the restart-after-connect concept are all directions we want to pursue. We'll build these on `dev` with:
- Full locale parity
- Async restart with error feedback
- Rendered component tests
- Integration verification with the reasoning-effort clamp

Thank you @HaydernCenterpoint for this comprehensive work — the scope and quality of the redesign show deep familiarity with the product surface.

HaydernCenterpoint:
Superseded by #247, now based on dev. The replacement includes Russian locale parity, waits for the Codex restart to complete and reports platform failures, uses rendered React UI tests, and keeps the change focused on the guided Providers flow. Thank you for the review feedback.

PR #235, opened 2026-07-22
feat(openrouter): add configurable provider routing

riique:
## Summary

- add provider-wide OpenRouter routing preferences through `openRouterRouting`
- add exact per-model overrides through `modelOpenRouterRouting`
- map camel-case config to OpenRouter's `provider.order`, `provider.only`, and `provider.allow_fallbacks` wire fields
- validate the canonical OpenRouter Chat Completions target and fail closed on malformed or inapplicable configuration
- document the feature in the architecture SOT and all configuration reference locales

## Why

OpenRouter can serve the same model through multiple inference providers. Those endpoints can differ substantially in prompt-cache support, cache hit behavior, retention, and cache pricing. For long Codex sessions, an endpoint with strong caching can be materially cheaper and faster than another endpoint serving the same model.

OpenRouter provides sticky routing automatically, but users still need explicit control when they require a particular provider's cache semantics or want to fail closed instead of silently switching to an endpoint with worse cache behavior. OpenRouter's documented request-level provider routing already supports this through `order`, `only`, and `allow_fallbacks`; this PR exposes those controls in opencodex.

References:

- https://openrouter.ai/docs/guides/routing/provider-selection
- https://openrouter.ai/docs/guides/best-practices/prompt-caching

## Example

```json
{
  providers: {
    openrouter: {
      adapter: openai-chat,
      baseUrl: https://openrouter.ai/api/v1,
      apiKey: ${OPENROUTER_API_KEY},
      openRouterRouting: {
        order: [deepseek],
        allowFallbacks: false
      },
      modelOpenRouterRouting: {
        anthropic/claude-sonnet-5: {
          only: [anthropic],
          allowFallbacks: false
        }
      }
    }
  }
}
```

A matching model override replaces the provider-wide default. With no routing preference configured, request bodies retain the existing behavior and OpenRouter keeps its default routing policy.

## Validation

- focused OpenRouter routing tests: 26 passed
- focused routing/config/server tests: 119 passed
- `bun run typecheck`: passed
- `bun run lint:gui`: passed
- docs build: 97 pages generated successfully
- broad suite run: 3370 passed, 4 skipped; three in-flight feature failures were fixed and pass in the focused rerun. Two unrelated Windows Cursor shell-fixture tests still fail locally because they invoke POSIX `printf`; they pass only when that shell dependency is available.


lidge-jun:
Review done — the fail-closed semantics look right: with no routing config nothing is injected, invalid config is rejected, and routing only applies to the canonical OpenRouter URL on the `openai-chat` transport. Config surface follows the existing provider-wide + exact-model override pattern, and the locale docs are consistent.

Local verification on top of current `dev` passed (focused routing/config/openai-chat tests + typecheck). I've approved the pending CI runs; once Cross-platform CI and React Doctor are green this is ready to absorb into `dev`.

PR #237, opened 2026-07-22
fix(anthropic): raise adaptive-thinking max_tokens default

jonathanli12:
## Summary

- Adaptive Claude models (`claude-fable-5`, `claude-sonnet-5`, `claude-opus-4-7`, `claude-opus-4-8`) previously kept the generic `max_tokens = 8192` default when Codex omitted `max_output_tokens`.
- The adaptive path only set `thinking.type = "adaptive"` and `output_config.effort`, and never resized `max_tokens`.
- Long Codex parent turns hit that ceiling mid-work and finished with no final message, even though those models support a 128k output ceiling.
- Default adaptive `max_tokens` to `128000` when Codex omits a budget, and preserve explicit `maxOutputTokens`.

## Verification

- `bun test tests/anthropic-reasoning.test.ts`
- Added regression coverage that adaptive models default to `128000` and preserve explicit budgets.

## Checklist

### DIFF
--- a/src/adapters/anthropic.ts
+++ b/src/adapters/anthropic.ts
@@ -35,6 +35,13 @@ function toAnthropicContentPart(p: OcxContentPart): unknown {
 
 /** Default `max_tokens` when Codex omits `max_output_tokens`. */
 const DEFAULT_MAX_TOKENS = 8192;
+/**
+ * Default `max_tokens` for adaptive-thinking models (Fable 5, Sonnet 5, Opus 4.7+)
+ * when Codex omits `max_output_tokens`. Adaptive path previously left the 8192 default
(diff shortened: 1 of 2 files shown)


lidge-jun:
Thanks for identifying the adaptive-thinking token ceiling — the diagnosis is right. We're going with #256 as the base fix because it addresses both sides of the incident: adaptive token sizing *and* end-to-end `stop_reason: max_tokens` propagation, so any remaining truncation is surfaced correctly instead of appearing as a successful empty response. Its effort-aware sizing is also preferable to raising every adaptive-thinking request to 128k unconditionally.

Closing in favor of #256 (which still has review changes requested — feel free to weigh in there).

PR #247, opened 2026-07-22
Simplify provider setup with unified OAuth and API panels

HaydernCenterpoint:
## Summary

- Consolidate everyday provider setup into one guided Providers screen instead of requiring users to switch to a separate Workspace view.
- Split setup into two clear panels: OAuth account providers on the left and API-key providers on the right. Only connected or configured providers stay visible; the `+` pickers provide search and hide providers that are already set up.
- Replace the header-level Add Provider action with Restart Codex. New providers are synced immediately and the UI clearly warns when a restart is recommended.
- Keep raw JSON editing available for advanced cases while making the normal path usable without knowing provider IDs, auth modes, or config structure.

The goal is to converge on one approachable provider UI. Many OpenCodex users will not be familiar with provider configuration, so a searchable, guided setup flow is easier to understand than multiple competing views.

This supersedes #226 and addresses its review feedback: the branch now targets `dev`, Russian locale keys are included, restart waits for real platform completion and propagates failures, and the UI coverage renders React output instead of checking source strings.

## Verification

- `bun test --isolate ./tests/codex-app-restart.test.ts ./tests/provider-auth-ui-render.test.ts ./tests/provider-workspace-data.test.ts ./tests/provider-registry-parity.test.ts` — 80 pass, 2 snapshots
- `bun run typecheck`
- `cd gui && bun run lint`
- `cd gui && bun run build`
- `bun run privacy:scan`
- `node gui/scripts/sync-locale-keys.mjs` — de/ko/zh/ru up to date
- Full suite: 3,365 pass and 4 skip. Two 5-second timeout cases pass when rerun independently with an appropriate timeout; two pre-existing Cursor native-exec cases fail locally on Windows because their fixture invokes POSIX `printf` through `cmd.exe` (no Cursor files are changed here).

## Checklist

- [x] Docs or release notes were updated when needed (not needed for this localized UI-only flow).


Wibias:
Hey @HaydernCenterpoint thanks for the PR.

I do not think this PR achieves its stated goal of consolidating provider setup into one approachable flow. It adds two summary panels above the existing provider list while retaining the Workspace/Classic split, the existing Add Provider modal and a separate OAuth account dialog. This increases the number of provider representations and entry points rather than reducing them.

The primary information architecture is also backwards. Users should select a provider first and then choose from the authentication methods that provider supports. Requiring users to decide between “OAuth” and “API Providers” upfront contradicts the stated goal of making setup usable without understanding authentication modes. These categories are not mutually exclusive either, since some providers support both OAuth and API-key configuration.

### Configured providers must remain available

Hiding already configured providers is a blocker. A configured provider is not necessarily complete. Users may need:

* another account or API key;
* separate personal and work configurations;
* another endpoint, proxy or region;
* different model settings;
* credential replacement or rotation;
* another authentication method for the same provider.

The current implementation also compares configured provider names with preset IDs. This means a canonical configuration name may hide the preset while a custom-named configuration for the same provider does not. The resulting behaviour depends on naming rather than actual provider capabilities.

Configured providers should remain visible and show their current state. Selecting one should offer actions such as:

* Add another configuration
* Add another account or key
* Edit existing
* Replace credentials

If multiple configurations are technically unsupported for a particular provider, the UI should explain that limitation rather than silently removing the provider.

### Free providers are missing

The redesigned flow removes the existing Free provider category without providing an equivalent replacement.

When the API-provider picker is opened in API-key-only mode, the Accounts, Free and Paid tabs are hidden. The catalogue is then restricted to providers that support API-key setup. Free providers using forwarding, local, keyless or other non-API-key setup methods are therefore excluded. The OAuth picker only covers account-login providers, so it does not restore those missing options.

This is a functional regression, not merely an information-architecture preference. Free providers are particularly important for onboarding users who want to try OpenCodex without first obtaining a paid API key or connecting a subscription account.

There are only two possible outcomes:

1. The missing providers are no longer accessible through the normal UI, which removes supported functionality.
2. They remain accessible only through Workspace or another old flow, which means this PR has not unified provider setup.

OAuth and API key are not an exhaustive taxonomy of supported provider setup methods. The replacement flow must support free, forwarding, local, keyless and custom configurations before it can replace the existing Add Provider flow.

### The panels duplicate existing management

The two summary panels mostly repeat information already shown by the complete provider cards below them. The panels show configured providers and simple status information, while the actual management actions remain in the original cards.

The page therefore contains:

* OAuth summaries;
* API-provider summaries;
* full provider-management cards;
* the Workspace view;
* separate provider pickers;
* raw JSON management.

This is more interface, not less interface.

The panels also permanently reserve the most prominent part of the page for low-frequency setup actions. An API-key-only user receives an empty OAuth panel. An OAuth-only user receives an empty API-provider panel. Those empty sections imply incomplete setup even when the user has deliberately configured everything they need.

### Sync and restart need an explicit state flow

Replacing the header-level Add Provider action with Restart Codex can make sense because provider additions are now initiated through the panel controls. However, the automatic synchronisation behaviour and restart messaging are not well designed.

The current implementation automatically runs model synchronisation after adding a provider or completing a new OAuth connection. That combines three separate operations:

1. Saving the provider configuration.
2. Synchronising the available models.
3. Restarting Codex so the running application loads the updated model list.

These operations can succeed or fail independently and should not be represented as one vague success message.

I would use the following flow:

1. Save the provider immediately.
2. Show that the provider was added but its models have not yet been synced.
3. Present Sync Models as the primary next action.
4. Show progress while synchronisation runs.
5. After a successful sync, and only when the model catalogue changed, ask whether the user wants to restart Codex now.
6. Clearly warn that restarting interrupts an active turn.
7. Offer Restart now and Later.
8. Keep a visible Restart required state until the restart occurs.

The restart prompt should appear after synchronisation succeeds, not before it begins. If synchronisation fails, the UI should show a retry action rather than recommending a restart. If synchronisation produces no changes, there is no reason to interrupt the user with a restart prompt.

The header-level Restart Codex action can remain available as a manual action, but it should reflect state. It should become visually prominent when a restart is required rather than relying on a temporary notification that users may miss.

### The implementation adds special-case complexity

The `apiKeyOnly` catalogue mode adds conditional tabs, exclusion behaviour, alternative filtering and preset coercion to an existing general-purpose catalogue.

For providers whose primary authentication method is not an API key, the implementation clones the preset and replaces its authentication value with `key`. That is brittle. Authentication methods should be represented explicitly as supported options rather than turning one preset into another mode by replacing a property.

A cleaner structure would be:

1. Select a provider.
2. Read its supported setup methods.
3. Select OAuth, API key, forwarding, local or another supported method.
4. Render the appropriate configuration form.

### The PR description understates the scope

The checklist describes this as a localised UI-only flow, but the PR changes application restart behaviour, the management API, provider derivation, authentication capabilities and model synchronisation.

That is not merely a visual rearrangement. These behaviours require clear documentation and tests covering the actual user workflows, not only rendering snapshots and type checks.

### Recommended direction

I would retain the useful lower-level work where appropriate, but revise the flow around the following principles:

1. Keep Workspace and Classic as alternative views over the same provider configuration, provided both remain functionally complete and consistent.
2. Use a provider-first add flow. Users should select the provider before choosing from the authentication methods it supports.
3. Support every existing setup type in the new flow, including OAuth, API key, free, local, forwarding, keyless and custom providers.
4. Keep already configured providers available in the picker so users can add another account, key, endpoint or separately named configuration.
5. Avoid duplicating providers within the same view. In Classic view, the new summary panels should either contain the actual management actions or be removed, rather than repeating the detailed provider cards shown below.
6. Ensure that switching between Workspace and Classic does not change which provider types or management actions are available.
7. Make model synchronisation explicit, with visible progress, success confirmation, failure handling and retry support.
8. After a successful synchronisation, prompt the user to restart Codex only when the updated model list actually requires it. The prompt should clearly warn that an active turn will be interrupted and offer Restart now and Later.

@lidge-jun In my opinion the two permanent panels and the restricted provider picker should not be merged in their current form.

HaydernCenterpoint:
@Wibias Thanks for the preview, i will working on it.

Ingwannu (changes requested):
The provider setup direction is valuable, and I found no malware or credential-exfiltration behavior. However this large branch conflicts with current dev in Providers.tsx and management-api.ts, has no full test/build matrix, and adds a security-sensitive local process-control endpoint for restarting Codex. Please rebase, resolve the current provider/auth changes, and run the full CI matrix before this can merge. Keep the restart endpoint bounded to fixed platform commands and the existing origin-checked management surface.

Wibias:
@Ingwannu Please have a look at my comment. There are many other problems with that pr.

Ingwannu:
@Wibias Thanks for flagging this. I checked the current `b498b944` diff against `dev`, and your concerns are valid and broader than the conflict/CI/security blockers in my first review.

The current implementation still has concrete functional and information-architecture regressions:

- `apiKeyOnly` hides the Accounts/Free/Paid catalog and filters out configured preset IDs, so configured-but-incomplete providers can become impossible to finish or duplicate from this flow.
- the API-only path excludes supported free/local/forward/keyless setup methods instead of providing a complete replacement flow;
- non-key presets can be cloned with `auth: "key"`, which treats authentication capability as a UI coercion rather than an explicit provider contract;
- the new summary panels duplicate the existing management cards and leave the Workspace/Classic/Add Provider/account-dialog split in place;
- provider creation/login immediately triggers model sync and reduces the result to a temporary status tone; there is no durable `sync required` / `restart required` state, retry flow, or restart prompt conditional on an actual catalog change.

These are blocking product regressions, not styling preferences. The PR should remain `CHANGES_REQUESTED` and should not be merged in its current form. The next revision should rebase onto current `dev` and use a provider-first flow that preserves every existing setup type and keeps configured providers actionable. It also needs workflow-level tests for incomplete providers, repeated configurations/accounts, free/local/keyless/forward options, sync failure/retry, no-op sync, and restart-required state.

PR #251, opened 2026-07-22
Fix Antigravity tool schemas with duplicate required fields

HaydernCenterpoint:
## Summary

- Deduplicate `required` entries while sanitizing Google-family tool schemas.
- Filter non-string `required` entries in the same pass so the emitted schema remains valid JSON Schema.
- Prevent Claude-on-Antigravity from rejecting the entire request with `tools.N.custom.input_schema: JSON schema is invalid. It must match JSON Schema draft 2020-12`.

Draft 2020-12 requires `required` arrays to contain unique strings. The sanitizer already walks every nested schema, so fixing that shared path covers native and MCP tools without provider-specific workarounds.

## Verification

- Live Antigravity reproduction: duplicate `required` entries returned HTTP 400 before sanitization; the sanitized payload returned HTTP 200.
- `bun test tests/google-tool-schema.test.ts tests/google-adapter.test.ts tests/google-antigravity-wire.test.ts` — 33 pass
- `bun run typecheck`
- `bun run privacy:scan`

## Checklist

- [x] Docs or release notes were updated when needed (not needed for this compatibility fix).

### DIFF
--- a/src/adapters/google-tool-schema.ts
+++ b/src/adapters/google-tool-schema.ts
@@ -73,6 +73,10 @@ function sanitize(node: unknown, defs: Map<string, unknown>, depth: number): unk
     if (key === "const") { out.enum = [value]; continue; }
     if (key === "exclusiveMinimum" && typeof value === "number") { out.minimum = value; continue; }
     if (key === "exclusiveMaximum" && typeof value === "number") { out.maximum = value; continue; }
+    if (key === "required" && Array.isArray(value)) {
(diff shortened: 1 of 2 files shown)


lidge-jun:
Code review looks good — stable `Set`-based dedupe inside the recursive `sanitize()` covers nested schemas, non-string entries are dropped, and inputs aren't mutated. I've approved the pending CI runs (first-time contributor gate); once Cross-platform CI and React Doctor come back green this is ready to absorb into `dev`.

PR #254, opened 2026-07-22
fix(claude): preserve subscription OAuth with host routing

### LINKED ISSUE #253: [Bug] subscription-mode ocx claude becomes unauthenticated when host-managed flag is injected without a host token
## Summary

Since v2.7.27, `ocx claude` injects `CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST=1` in every Claude Code auth mode. In the default subscription mode, OpenCodex intentionally does not provide `ANTHROPIC_AUTH_TOKEN` because Claude Code is expected to keep using the user's existing claude.ai OAuth session. On Claude Code 2.1.206, combining that native-OAuth path with the host-managed flag makes the CLI report `Not logged in · Please run /login`.

The failure is isolated to the flag: the same installed Claude Code and OAuth session work when the variable is absent or set to `0`.

## Environment

- Windows 11
- OpenCodex 2.7.31
- Claude Code 2.1.206
- `claudeCode.authMode`: unset (default subscription mode)
- No OpenCodex admission API keys configured
- Existing claude.ai Max OAuth login

## Reproduction

All tests below were run in the same shell environment. Values were changed only for the child process; no persistent environment variables were modified.

```text
$ claude -p "Reply with exactly: OK" --output-format text
OK
exit 0

$ env -u CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST ocx claude -p "Reply with exactly: OK" --output-format text
Not logged in · Please run /login
exit 1

$ CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST=0 ocx claude -p "Reply with exactly: OK" --output-format text
OK
exit 0

$ CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST=1 claude -p "Reply with exactly: OK" --output-format text
Not logged in · Please run /login
exit 1
```

This also shows that the failure does not require OpenCodex's base URL: setting the flag on plain `claude` is sufficient in this environment.

## Regression source

Commit `5875a0b` added the unconditional default in `buildClaudeEnv()` and the GUI manual environment block. It first shipped in v2.7.27:

```ts
setDefault("CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST", "1");
```

The flag has a valid purpose: it protects OpenCodex routing from stale provider variables in `~/.claude/settings.json`. The problem is that subscription mode simultaneously promises a native OAuth path and intentionally supplies no host credential:

```ts
if (!env.ANTHROPIC_AUTH_TOKEN && config.claudeCode?.authMode === "proxy") {
  env.ANTHROPIC_AUTH_TOKEN = "opencodex-proxy";
}
```

The current unit test checks only that the flag is assembled, not whether the resulting Claude Code process still has a usable subscription login.

## Expected behavior

Default subscription mode should preserve the existing claude.ai OAuth login while still routing requests through OpenCodex.

## Suggested direction

Gate the host-managed flag on whether OpenCodex actually supplies `ANTHROPIC_AUTH_TOKEN`:

- native subscription mode, no host token: do not inject the flag
- proxy mode or configured admission key: keep the flag and its settings-env routing protection
- preserve an explicit user-provided `0` or `1`

The same condition should be applied to the GUI's copy/paste environment block. This avoids a Claude Code version gate and preserves the original anti-hijack behavior on host-token paths.

## Temporary workaround

Use a shell-scoped override rather than a persistent `setx` value:

```powershell
$env:CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST = "0"
ocx claude
```

## Notes

I am intentionally reporting only the directly reproduced behavior on Claude Code 2.1.206. I am not assuming an affected version range or a specific internal OAuth implementation beyond the observable login result.

MustangRider:
## Summary

- preserve Claude Code's native claude.ai OAuth path by not injecting `CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST` when OpenCodex supplies no auth token
- retain the host-managed routing guard for proxy mode and configured admission-key paths
- keep explicit user `0`/`1` overrides and make the GUI manual environment block follow the same auth-mode rule
- add regression coverage for subscription, proxy, admission-key, model-slot, and manual-env behavior

Fixes #253.

## Why

Since v2.7.27, the launcher has injected the host-managed provider flag in every auth mode. Default subscription mode deliberately leaves `ANTHROPIC_AUTH_TOKEN` unset so Claude Code can use the existing claude.ai OAuth session. On Claude Code 2.1.206 for Windows, that combination reports `Not logged in · Please run /login`.

The fix gates the flag on the resulting `ANTHROPIC_AUTH_TOKEN`, rather than on a Claude Code version. This preserves the original protection against settings-sourced provider variables whenever OpenCodex actually owns the credential and routing path.

### Focused tests

```text
25 pass
0 fail
```

Covered files:

- `tests/claude-cli.test.ts`
- `tests/claude-manual-env.test.ts`

### Runtime

Using the modified launcher with the existing Windows subscription login and no flag override:

```text
$ env -u CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST bun run src/cli/index.ts claude -p "Reply with exactly: OK" --output-format text
OK
```

Explicit user opt-out remains functional:

```text
$ CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST=0 bun run src/cli/index.ts claude -p "Reply with exactly: OK" --output-format text
OK
```

An isolated proxy-mode run passed the login gate and reached upstream routing. The configured Claude alias then received an unrelated upstream 400 because that Claude-shaped model is unsupported by the available ChatGPT/Codex account; importantly, it did not produce `Not logged in`.

### Project checks

- typecheck: pass
- GUI lint: pass
- privacy scan: pass
- GUI doctor gate: pass/skip as not required by push range
- full tests: 3350 pass, 4 skip, 1 fail

The single failure is pre-existing/environmental and unrelated to this change: `tests/openai-provider-option-e2e.test.ts` detects that the real Claude config directory changes during its run. It reproduced when run alone, while both changed test files pass.

### DIFF
--- a/gui/src/pages/claude-manual-env.ts
+++ b/gui/src/pages/claude-manual-env.ts
@@ -34,12 +34,14 @@ export function buildManualEnv(state: ClaudeManualEnvState): string {
   return [
     `export ANTHROPIC_BASE_URL=${baseUrl}`,
     ...(state.authMode === "proxy"
-      ? ["export ANTHROPIC_AUTH_TOKEN=opencodex-proxy"]
+      ? [
+        "export ANTHROPIC_AUTH_TOKEN=opencodex-proxy",
+        // Host-managed routing guard (devlog 260720 020): only pair this with the
+        // host-supplied proxy token. The conditional preserves a user's opt-out.
+        '[ -z "${CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST+x}" ] && export CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST=1',
+      ]
       : ["# no ANTHROPIC_AUTH_TOKEN: your claude.ai login (and connectors) stay active"]),
     "export CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1",
-    // Host-managed routing guard (devlog 260720 020): CONDITIONAL so a shell where
-    // the user already exported =0 keeps its opt-out even after pasting this block.
-    '[ -z "${CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST+x}" ] && export CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST=1',
     ...(autoCompactActive ? [`export CLAUDE_CODE_AUTO_COMPACT_WINDOW=${state.autoCompactWindow ?? 350000}`] : []),
     ...modelEnvExports,
     "claude",
--- a/src/cli/claude.ts
+++ b/src/cli/claude.ts
@@ -65,15 +65,15 @@ export function buildClaudeEnv(config: OcxConfig, port: number, base: ClaudeLaun
   // Connectors still work because they check OAuth state ($o()), not base URL (Gd()).
   // Native /model picker discovery ("From gateway", Claude Code >= 2.1.129).
   setDefault("CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY", "1");
-  // Host-managed routing guard (devlog 260720_claude_authmode_persist/020): with
-  // this flag in the spawn env, Claude Code strips provider-managed vars
-  // (ANTHROPIC_BASE_URL/AUTH_TOKEN/API_KEY, model slots) from settings-sourced
-  // env (managedEnv.ts), so a leftover cc-switch/CCR ~/.claude/settings.json
-  // env block cannot silently hijack proxy routing away from opencodex.
-  // setDefault: an explicit user export (e.g. =0, isEnvTruthy-false) still wins.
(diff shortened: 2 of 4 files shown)


Ingwannu:
Thanks for the contribution. Current dev already contains the production fix for this issue: the host-managed Claude flag is injected only when ANTHROPIC_AUTH_TOKEN is actually present, explicit user environment values remain authoritative, and the manual environment snippet follows the same proxy-only rule. Your tests are useful, but the implementation now conflicts with the integrated fix. Security review found nothing suspicious. Closing as superseded by commit 2aa4430c62e83a7920469b8c3cb988b51d1b0917.

PR #255, opened 2026-07-22
Fix provider JSON editor save flow

rrmlima:
## What changed

- Save raw provider JSON through individual `POST /api/providers` requests instead of the disabled full-config `PUT /api/config` path.
- Apply `codexAutoStart` through `/api/settings` and delete providers removed from the JSON draft.
- Preserve redacted provider state (API keys, key pools, headers, Google mode, and model input caps) when a provider is overwritten from the dashboard.

## Why

The dashboard exposes a raw provider JSON editor, but its Save button submitted a full config PUT. The management API intentionally rejects that request because the GET response redacts credentials, leaving the editor unable to save.

## Validation and platform scope

Validated on Windows only:

- `bun run build:gui`
- `bun x tsc --noEmit`
- End-to-end save-flow simulation against the local management API, confirming provider rows and configured API keys remain intact.

The implementation is TypeScript and does not intentionally add Windows-specific behavior, but macOS and Linux have not been tested. Maintainers should verify the dashboard save flow on those platforms before merging.

### DIFF
--- a/gui/src/pages/Providers.tsx
+++ b/gui/src/pages/Providers.tsx
@@ -20,8 +20,11 @@ import { apiErrorMessage } from "../api-error";
 
 interface Config {
   port: number;
+  hostname?: string;
   defaultProvider: string;
-  providers: Record<string, { adapter: string; baseUrl: string; hasApiKey?: boolean; hasHeaders?: boolean; defaultModel?: string; models?: string[]; authMode?: string; keyOptional?: boolean; disabled?: boolean; note?: string; codexAccountMode?: "direct" | "pool" }>;
+  codexAutoStart?: boolean;
+  websockets?: boolean;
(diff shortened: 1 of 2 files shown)


Ingwannu (changes requested):
This draft is not safe to merge yet. Saving providers one-by-one, then deleting, then updating settings is non-atomic: a middle failure leaves a partially mutated config. More importantly, the editor starts from the redacted safeConfigDTO but only preserves a small subset of omitted provider fields, so a save can silently discard valid configuration such as reasoning maps, modality metadata, project/location, and thinking model settings. Please implement a single server-side validated/atomic update path that preserves all omitted fields. No malware or backdoor behavior was found.

lidge-jun:
**Post-merge Sol review note:** This PR was merged then reverted from `dev` after a thorough Sol code review found 3 high-severity issues:
1. Non-atomic save operation exposing intermediate configurations
2. Incomplete field preservation list (missing mcpServers, desktopExecutor, etc.)
3. TypeScript compilation failure (TS2352)

The core idea (preventing credential loss in the JSON editor) is valuable and will be rebuilt with an atomic batch config endpoint and typed DTO merge policy. See devlog analysis for details.

PR #256, opened 2026-07-22
fix(anthropic): resize max_tokens for adaptive thinking and thread stop_reason

### LINKED ISSUE #246: Fable (adaptive-thinking) + effort=max silently truncates to empty response — stop_reason dropped, max_tokens not resized
## Summary

When routing `claude-fable-5` through the Anthropic adapter with reasoning effort `max` (`output_config.effort: "max"`), the model consistently exhausts the entire `max_tokens` budget on adaptive thinking and returns **zero visible text**, while the proxy reports the turn as a normal completed response (no truncation/incomplete signal). Downstream clients (e.g. Claude Code / Claude Desktop) see this as "completed" with an empty message rather than a truncated one.

Reproduced 6/6 times on `claude-fable-5` with `requestedEffort: "max"`; 0/3 failures on the same model without an explicit `max` effort override (default/low effort completed normally in ~1-1.5s).

## Environment

- opencodex: `2.7.31` (npm, `@bitkyc08/opencodex`)
- OS: Windows 11 Pro (10.0.26200)
- Node: v22.17.0
- Bun (bundled runtime): 1.3.14
- Claude Code: 2.1.217 (Claude Desktop bundled CLI)
- Provider/model: `anthropic` / `claude-fable-5`
- Adapter: `src/adapters/anthropic.ts`

## Evidence (from local `~/.opencodex/usage.jsonl`)

```json
{"requestId":"ocx-mrvgow0n-64","model":"claude-fable-5","requestedEffort":"max","status":200,"durationMs":128798,"usage":{"inputTokens":101654,"outputTokens":8192,"totalTokens":109846}}
{"requestId":"ocx-mrvgsf9c-65","model":"claude-fable-5","requestedEffort":"max","status":200,"durationMs":128240,"usage":{"inputTokens":101798,"outputTokens":8192,"totalTokens":109990}}
{"requestId":"ocx-mrvgvc5f-6c","model":"claude-fable-5","requestedEffort":"max","status":200,"durationMs":126761,"usage":{"inputTokens":101907,"outputTokens":8192,"totalTokens":110099}}
{"requestId":"ocx-mrvh0wjg-6g","model":"claude-fable-5","requestedEffort":"max","status":200,"durationMs":127187,"usage":{"inputTokens":51417,"outputTokens":8192,"totalTokens":59609}}
{"requestId":"ocx-mrvh5u8v-6h","model":"claude-fable-5","requestedEffort":"max","status":200,"durationMs":123028,"usage":{"inputTokens":51435,"outputTokens":8192,"totalTokens":59627}}
{"requestId":"ocx-mrvhy0md-2","model":"claude-fable-5","requestedEffort":"max","status":200,"durationMs":125836,"usage":{"inputTokens":54833,"outputTokens":8192,"totalTokens":63025}}
```

Every single `requestedEffort: "max"` call for `claude-fable-5` hit **exactly** 8192 output tokens and took ~2 minutes. Non-`max` calls on the same model in the same log completed in ~1-1.5s with a handful of output tokens (e.g. 9, 14).

For comparison, a same-conditions Sonnet call completes in ~2.1s with 9 output tokens — so this is not an inherent slowness of the account/network, it's specific to Fable + `max` effort.

## Root cause (traced in `src/adapters/anthropic.ts`, v2.7.31)

Two compounding bugs:

**1. `stop_reason` is never read from the upstream Anthropic response.**

Neither the streaming SSE parser nor the non-streaming path extracts it:

```ts
// streaming path, ~line 796-804
case "message_delta": {
  const usage = data.usage as Record<string, number> | undefined;
  pendingUsage = mergeAnthropicUsage(pendingUsage, usage);
  break;                      // <- data.delta.stop_reason is discarded
}
case "message_stop": {
  yield* emitDone();          // <- no stop_reason passed through
  break;
}

// non-streaming path, ~line 820-848 (parseResponse)
const json = await response.json() as Record<string, unknown>;
// ...reads json.content and json.usage, never json.stop_reason
events.push({ type: "done", usage: usageFromAnthropic(usage) });
```

A repo-wide search for `stop_reason` inside `src/adapters/anthropic.ts` returns **zero matches**. Downstream, `src/server/claude-messages.ts:745` synthesizes the outgoing `message_delta`:

```ts
emit("message_delta", { type: "message_delta", delta: { stop_reason: (message as Rec).stop_reason ?? "end_turn", stop_sequence: null }, usage: ... });
```

Since `message.stop_reason` is never populated by the adapter, this **always** falls back to `"end_turn"` — even when Anthropic actually returned `"max_tokens"`. The client has no way to distinguish a genuinely finished turn from a token-capped one.

**2. Adaptive-thinking models don't get `max_tokens` resized for effort, and default to 8192.**

```ts
// ~line 36-37
const DEFAULT_MAX_TOKENS = 8192;
...
// ~line 610
max_tokens: parsed.options.maxOutputTokens ?? DEFAULT_MAX_TOKENS,
...
// ~line 626-646
if (usesAdaptiveThinking(modelId)) {
  // Adaptive-thinking models replace the token budget with an effort knob and reject
  // `thinking.type: "enabled"` outright — no budget/max_tokens re-sizing needed.
  body.thinking = { type: "adaptive" };
  body.output_config = { effort: adaptiveEffort(parsed.options.reasoning) };
} else {
  // ...budget_tokens / max_tokens resizing only happens here
}
```

`claude-fable-5` matches `ADAPTIVE_THINKING_FAMILY_MINIMUMS.fable = [0, 0]` (all Fable versions), so it always takes the adaptive branch, where `max_tokens` is left at whatever was set earlier (defaulting to 8192 when the caller doesn't pass `max_output_tokens`). At `effort: "max"`, adaptive thinking can consume the *entire* 8192-token budget internally, leaving 0 tokens for visible output — Anthropic correctly returns `stop_reason: "max_tokens"` with empty/near-empty `content`, but bug #1 means this is never surfaced.

## Expected behavior

Either:
- The adapter should thread `stop_reason` (streaming: `message_delta.delta.stop_reason`; non-streaming: `json.stop_reason`) through to the emitted `done`/`message` event so `"max_tokens"` reaches the client as an incomplete/truncated response instead of `"end_turn"`, **and/or**
- `max_tokens` should scale with `effort` for adaptive-thinking models too (similar in spirit to `reasoningBudget()`, which already exists but is currently only used on the non-adaptive branch), so `max`/`xhigh` effort doesn't get boxed into the same 8192 ceiling as default effort.

## Actual behavior

Client receives an HTTP 200, `status: "completed"` response with no visible text after ~2 minutes and 8192 output tokens fully consumed by internal thinking. Looks identical to a normal, successful, empty-output turn — no error, no truncation indicator.

## Workaround

Avoid `effort: "max"` for `claude-fable-5` (and likely `high`/`xhigh`, structurally same risk though not locally confirmed) until fixed. Default/low effort was reproduced as working correctly in the same logs.

Wibias:
## Summary

Fixes #246.

Two compounding bugs caused `claude-fable-5` (and other adaptive-thinking models) with `effort: "max"` to exhaust the entire 8192-token `max_tokens` budget on internal thinking and return zero visible text, while the proxy reported the turn as a normal completed response.

### Bug 1: `max_tokens` not resized for adaptive thinking

The adaptive-thinking branch set `thinking.type: "adaptive"` and `output_config.effort` but never resized `max_tokens`. At high efforts the model can consume the entire budget on internal thinking, leaving 0 tokens for visible output. The non-adaptive branch already handled this with `reasoningBudget() + OUTPUT_HEADROOM`.

**Fix:** Apply the same resize formula to the adaptive branch. For `max` effort this bumps from 8192 to 32000 (the existing `REASONING_MAX_TOKENS_CEILING`).

### Bug 2: `stop_reason` dropped from upstream responses

Neither the streaming SSE parser nor the non-streaming path extracted `stop_reason` from the upstream Anthropic response. The `done` event only carried `usage`, so the bridge always emitted `response.completed` even when Anthropic returned `stop_reason: "max_tokens"`. Clients had no way to distinguish a truncated turn from a finished one.

**Fix:**
- Capture `stop_reason` from streaming `message_delta.delta.stop_reason` and non-streaming `json.stop_reason`
- Thread it through the `done` event (new optional `stopReason` field on `AdapterEvent`)
- When `stopReason === "max_tokens"`, the bridge emits `response.incomplete` with `incomplete_details: { reason: "max_output_tokens" }` instead of `response.completed`

The outbound layer in `src/claude/outbound.ts` already handles `response.incomplete` with `max_output_tokens` correctly, mapping it to `stop_reason: "max_tokens"` in the Anthropic Messages SSE output.

## Changes

- `src/types.ts`: Add optional `stopReason` to the `done` event in `AdapterEvent`
- `src/adapters/anthropic.ts`: Resize `max_tokens` for adaptive thinking; capture `stop_reason` in both streaming and non-streaming paths
- `src/bridge.ts`: Emit `response.incomplete` when `stopReason === "max_tokens"`
- `tests/anthropic-reasoning.test.ts`: Verify adaptive `max_tokens` resize for high effort
- `tests/bridge.test.ts`: Verify `stopReason` threading produces correct terminal events

## Verification

- `bun test --isolate` on all anthropic + bridge + claude-messages tests: 100 pass, 0 fail
- `bun run typecheck`: clean

### DIFF
--- a/src/adapters/anthropic.ts
+++ b/src/adapters/anthropic.ts
@@ -634,11 +634,13 @@ export function createAnthropicAdapter(provider: OcxProviderConfig, cacheRetenti
           // default 8192-token request can spend everything on thought and return empty text.
           body.thinking = { type: "adaptive" };
           body.output_config = { effort: adaptiveEffort(parsed.options.reasoning) };
-          const maxOut = parsed.options.maxOutputTokens ?? DEFAULT_MAX_TOKENS;
-          body.max_tokens = Math.min(
-            REASONING_MAX_TOKENS_CEILING,
-            Math.max(maxOut, reasoningBudget(adaptiveEffort(parsed.options.reasoning)) + OUTPUT_HEADROOM),
-          );
+          const explicitMaxOut = parsed.options.maxOutputTokens;
+          const wantBudget = reasoningBudget(parsed.options.reasoning);
+          const floor = wantBudget + OUTPUT_HEADROOM;
+          // Respect explicit caller values above the default ceiling; only cap when using the default.
+          body.max_tokens = explicitMaxOut !== undefined
+            ? Math.max(explicitMaxOut, floor)
+            : Math.min(REASONING_MAX_TOKENS_CEILING, Math.max(DEFAULT_MAX_TOKENS, floor));
         } else {
           // Anthropic requires max_tokens > thinking.budget_tokens (max_tokens caps thinking +
           // visible output) and budget_tokens >= 1024. Codex sends the SAME value for both, which
--- a/src/adapters/anthropic.ts
+++ b/src/adapters/anthropic.ts
@@ -37,6 +37,9 @@ function toAnthropicContentPart(p: OcxContentPart): unknown {
 const DEFAULT_MAX_TOKENS = 8192;
 /** Safe ceiling for `max_tokens` (thinking + visible output) across current Claude 4.x models. */
 const REASONING_MAX_TOKENS_CEILING = 32_000;
+/** Adaptive-thinking ceiling: max effort budget (32k) + OUTPUT_HEADROOM (8k).
+ *  Must exceed REASONING_MAX_TOKENS_CEILING so effort=max actually preserves visible-output room. */
+const ADAPTIVE_THINKING_CEILING = 40_192;
 /** Anthropic's documented minimum `thinking.budget_tokens`. */
 const MIN_THINKING_BUDGET = 1024;
 /** Visible-output room added above the thinking budget when sizing `max_tokens`. */
@@ -637,10 +640,11 @@ export function createAnthropicAdapter(provider: OcxProviderConfig, cacheRetenti
           const explicitMaxOut = parsed.options.maxOutputTokens;
           const wantBudget = reasoningBudget(parsed.options.reasoning);
           const floor = wantBudget + OUTPUT_HEADROOM;
-          // Respect explicit caller values above the default ceiling; only cap when using the default.
+          // Preserve explicit caller limits as-is; for omitted limits use the adaptive ceiling
+          // so effort=max (budget=32k) still leaves OUTPUT_HEADROOM tokens for visible output.
           body.max_tokens = explicitMaxOut !== undefined
-            ? Math.max(explicitMaxOut, floor)
-            : Math.min(REASONING_MAX_TOKENS_CEILING, Math.max(DEFAULT_MAX_TOKENS, floor));
+            ? explicitMaxOut
+            : Math.min(ADAPTIVE_THINKING_CEILING, Math.max(DEFAULT_MAX_TOKENS, floor));
         } else {
           // Anthropic requires max_tokens > thinking.budget_tokens (max_tokens caps thinking +
           // visible output) and budget_tokens >= 1024. Codex sends the SAME value for both, which
--- a/src/adapters/anthropic.ts
+++ b/src/adapters/anthropic.ts
@@ -854,10 +854,11 @@ export function createAnthropicAdapter(provider: OcxProviderConfig, cacheRetenti
         }
       }
       const usage = json.usage as Record<string, number> | undefined;
+      const stopReason = typeof json.stop_reason === "string" ? json.stop_reason : undefined;
       events.push({
         type: "done",
         usage: usageFromAnthropic(usage),
-        ...(typeof json.stop_reason === "string" ? { stopReason: json.stop_reason } : {}),
+        ...(stopReason ? { stopReason } : {}),
       });
       return events;
     },
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -634,17 +634,26 @@ export function bridgeToResponsesSSE(
                 finishedItems.push(item as OutputItem);
                 outputIndex++;
               }
-              const truncated = event.stopReason === "max_tokens";
-              const response = {
-                ...responseSnapshot(truncated ? "incomplete" : "completed", finishedItems),
-                usage: responsesUsage(event.usage),
-                ...(truncated ? { incomplete_details: { reason: "max_output_tokens" } } : {}),
-              };
-              options?.onCompletedResponse?.(response);
-              emit(truncated ? "response.incomplete" : "response.completed", {
-                response,
-              });
-              reportTerminal(truncated ? "incomplete" : "completed");
+              if (event.stopReason === "max_tokens") {
+                // Upstream hit its output token cap. Surface as incomplete so the client can
+                // distinguish a truncated turn from a genuinely finished one (issue #246).
+                const response = {
+                  ...responseSnapshot("incomplete", finishedItems),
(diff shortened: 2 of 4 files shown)


lidge-jun:
Thanks — this is the direction we're adopting over #237, since it fixes both halves of #246: adaptive token sizing *and* end-to-end `stop_reason: max_tokens` propagation (the stop_reason threading here is correct across streaming/non-streaming and maps back properly on the Claude outbound path).

Two blockers before merge:

1. **The 32k+headroom policy doesn't actually preserve headroom at `effort=max`.** `reasoningBudget("max")` is 32,000, so the 40,192 floor gets capped right back to 32,000 — adaptive thinking can still eat the whole allowance and reproduce an empty visible response, just at a higher threshold.
2. **Explicit caller limits are silently raised** via `Math.max(explicitMaxOut, floor)`. Adaptive thinking doesn't require `max_tokens > budget_tokens`, so overriding an explicit output cap is surprising and potentially expensive for cost-capped callers.

Requested change: preserve explicit `maxOutputTokens` as-is; for omitted limits use an adaptive-specific ceiling that genuinely leaves visible-output room at `max`, plus an exact-value regression test (not just `>8192`). With those two addressed this merges.

Ingwannu (changes requested):
This has useful remaining value, but it is not mergeable as-is against current dev. The stop_reason/max_tokens completion plumbing is already present and now conflicts with dev. Please rebase and narrow the patch to the unsuperseded adaptive-thinking token budget work (including the 40192 ceiling and explicit maxOutputTokens coverage). The explicit-output override is a real limit/cost behavior change, so keep focused tests for that policy. Security review found no malware, exfiltration, dependency, or process-execution concern.

Ingwannu:
I rebased the current four-commit series onto `dev` (`d88b98e7`) using the maintainer-edit permission and re-ran the focused checks. The two earlier policy blockers are now resolved: omitted adaptive-thinking limits use 40,192 at `max` (32k reasoning + 8,192 visible-output headroom), while every explicit `maxOutputTokens` value is preserved exactly. The existing stop-reason behavior remains covered for streaming and batch responses.

Local verification on the rebased head:
- `bun test --isolate tests/anthropic-reasoning.test.ts tests/bridge.test.ts` — 55 pass
- `bun run typecheck`
- `bun run privacy:scan`
- `git diff --check origin/dev...HEAD`

Security review of the four-file diff found no credential handling, exfiltration, dependency, process-execution, or unsafe I/O changes. Waiting for the refreshed GitHub matrix before the final merge decision.

Ingwannu (approved):
Re-reviewed on rebased head `d88b98e7`. The two requested blockers are resolved: omitted adaptive `max` requests now reserve the full 8,192-token visible-output headroom at 40,192 total, and explicit `maxOutputTokens` values are preserved exactly. Streaming/batch truncation behavior is covered, all refreshed Linux/macOS/Windows checks pass, and the four-file diff contains no malware, credential exfiltration, unsafe I/O/process execution, or dependency changes.

PR #272, opened 2026-07-22
fix: add opt-in responses item id repair

Ingwannu:
## Summary
- add provider-local `responsesItemIdRepair` config for exact `message` / `reasoning` placeholder ids plus `repairMissingTerminalIds`
- repair downstream passthrough SSE item ids by `output_index` only when the provider opts in
- keep the tee'd inspection/replay path raw so synthetic ids never replay upstream
- document the config and cover single-stream plus sequential-stream lifecycles

## Root Cause
Some `openai-responses` compatibility gateways reuse placeholder `message` / `reasoning` ids (for example `rs_0`, `msg_0`) across sequential tool-loop responses and sometimes omit terminal ids. Codex Desktop correlates downstream reasoning/message activity by item id, so reused or missing ids can merge unrelated activity cards.

## Scope / Safety
- disabled by default
- exact string matches only for `message` / `reasoning`
- `function_call` ids and `call_id` remain untouched
- default passthrough bytes remain unchanged unless a provider explicitly opts in
- replay state still stores raw upstream responses, so repaired ids are not sent back upstream on later turns

## Validation
- `bun x tsc --noEmit`
- `bun test --isolate ./tests/responses-item-id-repair.test.ts ./tests/config.test.ts ./tests/passthrough-abort.test.ts ./tests/openai-responses-passthrough.test.ts`
- `bun run privacy:scan`

### DIFF
--- a/docs-site/src/content/docs/reference/configuration.md
+++ b/docs-site/src/content/docs/reference/configuration.md
@@ -173,6 +173,7 @@ network. Only do this on trusted networks, and always set a strong `OPENCODEX_AP
 | `noTopPModels?` | `string[]` | Models that reject caller-specified `top_p`. |
 | `noPenaltyModels?` | `string[]` | Models that reject presence/frequency penalties. |
 | `parallelToolCalls?` | `boolean` | Enable/disable parallel tool calls. OpenAI Chat defaults on; non-chat adapters advertise support only on explicit `true`. |
+| `responsesItemIdRepair?` | `{ message?: string[]; reasoning?: string[]; repairMissingTerminalIds?: boolean }` | Provider-local, disabled-by-default passthrough SSE repair for exact `message` / `reasoning` placeholder ids and missing terminal ids. Downstream only; function-call ids and `call_id` are never rewritten. |
 | `autoToolChoiceOnlyModels?` | `string[]` | Models whose `tool_choice` accepts only `auto` or `none`; forced/named choices are downgraded. |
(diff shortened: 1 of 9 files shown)


Ingwannu:
Security blocker follow-up: fixed on rebased head `3ba0ad51f909a0b8e9e38315b8d01b315993be08` (base `d4b4b05c`).

- Repair mappings are now scoped by both item type and `output_index`.
- Event-level `item_id` rewrites use an explicit message/reasoning lifecycle allowlist; `response.function_call_arguments.*` is never rewritten.
- `function_call` item `id` and `call_id` remain unchanged in deltas, `response.output_item.done`, and terminal response snapshots, including malformed cross-type index reuse.
- Added permanent adversarial coverage, same-index message/reasoning lifecycle coverage, and byte-for-byte untouched-SSE coverage.

Validation:
- Original `/tmp/opencodex-pr272-adversarial.test.ts` fails against the old head with `fc_1` rewritten to `rs_ocx_*`.
- Patched focused suite: 82 passed, 0 failed.
- `bun run typecheck` passed.
- `bun run privacy:scan` passed.
- `git diff --check` passed.

The PR remains draft.

Ingwannu:
Final re-review passed on head `3ba0ad51`. The prior cross-type `output_index` finding is closed: mappings are keyed by item type and index, event-level rewrites are restricted to message/reasoning lifecycle events, and function-call `id`/`item_id`/`call_id` remain untouched. Focused/adversarial tests, typecheck, privacy scan, diff check, and the full Linux/macOS/Windows matrix pass. No malware, backdoor, credential exfiltration, unsafe I/O, or dependency changes found.

Issue #208, opened 2026-07-21
[Feature]: Native OpenAI Chat Completions Proxy Endpoint (/v1/chat/completions)


mohamedabduissa:
### Problem to solve

Add a native OpenAI-compatible Chat Completions endpoint (POST /v1/chat/completions) to translate standard OpenAI request payloads into the internal Responses API format (/v1/responses) and format all responses back to standard OpenAI Chat Completion objects (both streaming and non-streaming).

### Proposed solution

Endpoint Specification
Endpoint: POST /v1/chat/completions
Content-Type: application/json
Headers: Authorization: Bearer <API_KEY>
Request Payload (ChatCompletionRequest)
Input Parameters:
`
{
  "model": "string",
  "messages": [
    {
      "role": "system | user | assistant | tool",
      "content": "string | array",
      "name": "string (optional)",
      "tool_calls": "array (optional)"
    }
  ],
  "stream": false,
  "temperature": 0.7,
  "top_p": 1.0,
  "max_tokens": 1024,
  "stop": ["string"],
  "frequency_penalty": 0.0,
  "presence_penalty": 0.0,
  "tools": ["array (optional)"],
  "tool_choice": "string | object (optional)",
  "response_format": "object (optional)"
}

`

Response Specification
1. Non-Streaming (stream: false)
Response Status: 200 OK
Content-Type: application/json
Return Structure: Standard OpenAI ChatCompletionResponse object.

`{
  "id": "chatcmpl-1721641000",
  "object": "chat.completion",
  "created": 1721641000,
  "model": "deepseek-r1",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "Response text here...",
        "reasoning_content": "Reasoning/thinking process (if model supports reasoning)",
        "tool_calls": []
      },
      "finish_reason": "stop"
    }
  ],
  "usage": {
    "prompt_tokens": 50,
    "completion_tokens": 150,
    "total_tokens": 200
  }
}
`

2. Streaming (stream: true)
Response Status: 200 OK
Content-Type: text/event-stream
Headers:
Cache-Control: no-cache
Connection: keep-alive
X-Accel-Buffering: no
Return Structure: Server-Sent Events (SSE) yielding OpenAI delta chunks ending with data: [DONE].

[...]

PR #279, opened 2026-07-22
feat: add GitHub Copilot App support via OpenAI-compatible chat completions

HaydernCenterpoint:
﻿## Summary
- Add OpenAI-compatible `POST /v1/chat/completions` so GitHub Copilot App can use OpenCodex as a custom Model provider.
- Translate Chat Completions requests into the existing Responses path and bridge stream/non-stream replies back to Chat Completions shape.
- Document setup (`docs/github-copilot-app.md`) and cover discovery/chat endpoint behavior with tests.

This is a **client integration** for GitHub Copilot App (BYOK / Model providers). It is separate from the existing experimental upstream `github-copilot` provider.

Replaces #278 (opened against `main` by mistake).

## How to use
1. Run OpenCodex (`ocx start`)
2. In GitHub Copilot App → Settings → Model providers → Add provider
3. Base URL: `http://127.0.0.1:10100/v1`
4. API key: empty on loopback
5. Sync models / add model by id (`provider/model`)

## Test plan
- [x] `bun test tests/chat-completions-endpoint.test.ts`
- [x] `bun test tests/server-auth.test.ts -t "root fallback"`
- [x] Manual: `GET /v1/models` returns OpenAI list shape
- [x] Manual: `POST /v1/chat/completions` stream + non-stream succeed for configured providers
- [x] Manual: GitHub Copilot App can chat after pointing at `/v1` and selecting an authenticated model

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,7 @@ next Codex session. opencodex keeps these behaviors:
 
 - **Use any LLM with Codex.** 5 protocol adapters cover Anthropic Messages, Google Gemini, Azure, OpenAI Responses passthrough, and every OpenAI-compatible Chat Completions endpoint — that's 40+ providers out of the box.
 - **Use any LLM with Claude Code too.** The same daemon serves the Anthropic Messages API (`/v1/messages` + `count_tokens`): `ocx claude` launches Claude Code fully wired, and routed models appear in its native `/model` picker via gateway model discovery (`claude-ocx-<provider>--<model>` aliases, Claude Code 2.1.129+). Configure slots and model maps on the dashboard's Claude page.
+- **Use any LLM with GitHub Copilot App too.** Point Copilot's Model providers at `http://127.0.0.1:10100/v1` — OpenCodex serves OpenAI-compatible `GET /v1/models` and `POST /v1/chat/completions` so routed models sync into the app. See [docs/github-copilot-app.md](docs/github-copilot-app.md).
 - **Pool ChatGPT accounts safely.** Keep existing Codex threads on one account while new sessions
   can auto-pick a lower-usage account from the pool, with quota refresh and non-PII request labels.
 - **Log in once, skip the API key.** OAuth support for xAI, Anthropic, and Kimi means you can authenticate with your existing account. Tokens auto-refresh. Or forward your `codex login`, paste an API key, or use `${ENV_VAR}` references — your call.
--- a/docs/github-copilot-app.md
+++ b/docs/github-copilot-app.md
@@ -1,0 +1,59 @@
+# GitHub Copilot App
+
+OpenCodex can act as an **OpenAI-compatible model provider** for the GitHub Copilot
+desktop app (Settings → Model providers). This is a client integration: Copilot App
+calls OpenCodex; it is separate from the experimental upstream `github-copilot`
+provider that uses a Copilot subscription as a backend.
+
+## Requirements
+
+1. OpenCodex proxy running locally (`ocx start` / `ocx gui`).
+2. At least one configured provider with models (dashboard → Providers).
+3. GitHub Copilot desktop app with **Model providers** support.
+
+## Setup
+
+1. Start OpenCodex and confirm health:
+
+   ```bash
+   curl http://127.0.0.1:10100/healthz
+   curl http://127.0.0.1:10100/v1/models
+   ```
+
+2. In GitHub Copilot App: **Settings → Model providers → Add provider**.
+
+3. Configure:
+
+   | Field | Value |
+   | --- | --- |
+   | Name | `OpenCodex Gateway` (any label) |
+   | Base URL | `http://127.0.0.1:10100/v1` |
+   | API key | leave blank on loopback; for non-loopback binds use `OPENCODEX_API_AUTH_TOKEN` |
+
+4. Sync models from the endpoint, or add a model by id (`provider/model`, e.g.
+   `anthropic/claude-sonnet-4-6`).
+
+5. Select a synced model and chat.
+
+## Endpoints used
+
+| Method | Path | Role |
+| --- | --- | --- |
+| `GET` | `/v1/models` | Model discovery (OpenAI list shape) |
+| `POST` | `/v1/chat/completions` | Chat turns (stream + non-stream) |
+
+OpenCodex translates Chat Completions into its internal Responses path, so all
+existing providers, routing, OAuth, and sidecars apply.
+
+## Troubleshooting
+
+- **No models configured** — ensure the proxy is up, base URL ends with `/v1`
+  (not `/v1/chat/completions`), and `GET /v1/models` returns a non-empty `data`
+  array. Add/enable providers in the OpenCodex dashboard, then sync again.
+- **401** — remote (non-loopback) binds require an admission token as
+  `Authorization: Bearer …` or `x-opencodex-api-key`.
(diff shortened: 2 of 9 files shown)


lidge-jun:
Thanks — this is substantial work and the endpoint itself is wanted (it's effectively the general `/v1/chat/completions` surface #208 asks for, so worth titling it that way: "general Chat Completions compatibility, initially for Copilot App"). CI is green, but four correctness blockers before this can merge:

1. **Parallel tool-call streaming is broken.** `src/chat/outbound.ts` reads `item_id` but ignores it, appending every argument delta to the *last* tool index. With two overlapping function calls the first tool gets empty arguments and the last gets interleaved ones. The Responses events do provide `item_id` — deltas need to be routed by it.
2. **Direct-mode auth semantics.** `src/server/chat-completions.ts` unconditionally discards the caller's `Authorization` and substitutes the main-account token. In direct mode the caller bearer *is* the user identity — without a main token this 401s, and with one it can execute as a different account. Please follow the existing Responses convention (admission token vs upstream bearer via `x-opencodex-api-key`).
3. **`response_format` is accepted but silently dropped** at the adapter boundary for routed `openai-chat` calls — clients expecting JSON/schema output get plain text. Either thread it through or reject with an explicit 400.
4. **Native passthrough never finalizes the request log** — the terminal/cancel callbacks that `/v1/responses` passes are missing, so successful native `gpt-*` requests leave usage/request-log entries incomplete.

Tests to add alongside: parallel tool deltas, direct-mode identity, structured output (or its 400), and native terminal/cancel logging. Also please document the supported-field subset (penalties, `n`, logprobs etc. are unverified) rather than implying full Chat Completions parity. Happy to re-review once these land.

Ingwannu (changes requested):
This integration is valuable and the auth/destination/secret posture is generally consistent with the existing data-plane, but two correctness blockers must be fixed before merge.

1. `response.function_call_arguments.delta` carries `item_id`, but the new bridge ignores it and appends every delta to the last opened tool index. Interleaved parallel calls therefore corrupt arguments: in my two-call repro, call 1 ended empty and call 2 received concatenated malformed JSON. `response.output_item.done` can also drop final arguments after an added frame. Track tool calls by item id/call id and emit each delta/final snapshot to the matching index.

2. `response.failed` and truncated internal streams are converted into a normal chat-completion assistant delta containing `[error] ...` followed by `[DONE]`; non-streaming collection then returns HTTP 200 with model-authored-looking error text. Preserve typed failure/truncation state and return an OpenAI-compatible error/non-success response instead of success content.

Please add adversarial coverage for two interleaved function calls, done-frame final arguments, streaming failure, non-streaming failure, and truncated internal streams. The current full suite/CI passes because those cases are not covered. No malware, backdoor, credential persistence, unsafe process/filesystem behavior, or dependency concern was found.

HaydernCenterpoint:
Thanks for you guys review. I will update it when im home.

lidge-jun:
Re-reviewed head `43cc1a8a` — thanks for the fast turnaround. All four original blockers are genuinely fixed:

1. ✅ Parallel tool deltas now route by `item_id` (`src/chat/outbound.ts`), with an interleaved-delta test.
2. ✅ Direct-mode identity preserved; Responses-style admission applied (`src/server/index.ts`, `chat-completions.ts`), with identity tests.
3. ✅ `response_format`: mapped for native, explicit 400 for routed `openai-chat`.
4. ✅ Native terminal/cancel logging wired idempotently, with a success-terminal test.

Full CI is green on the new head. Two remaining blockers before merge:

1. **Failures still masquerade as HTTP 200.** `fail()` in `src/chat/outbound.ts:137-147` converts errors into assistant text (`[error] ...`) followed by `[DONE]`; `response.failed` (`:260-265`) and truncated SSE both take this path, and non-streaming collects it as a normal completion with 200 (`chat-completions.ts:183-199`). Chat Completions clients need a real error: non-streaming should return an error status + OpenAI-style error body; streaming should emit an error event / terminate abnormally rather than a clean `[DONE]`. Please add failure/truncation regression tests.
2. **`output_item.done` final arguments are dropped** when the call was already registered from `output_item.added` (`outbound.ts:213-243`) — if the done-frame snapshot contains the authoritative final arguments, they're lost. Reconcile the done-frame `item.arguments` (last-write-wins) and add a done-frame test.

Nice-to-have (non-blocking): a client-cancel logging test, and a remote-mode test covering `x-opencodex-api-key` admission + upstream bearer separation. With the two blockers addressed this is mergeable.

HaydernCenterpoint:
@lidge-jun Updated on head `152e37ea` — the two remaining blockers from your re-review should now be addressed.

### 1. Failures no longer masquerade as HTTP 200 / clean `[DONE]`
- `fail()` in `src/chat/outbound.ts` now emits an OpenAI-style `error` SSE frame and closes the stream **without** `[DONE]` (no more assistant text like `[error] ...` + success termination).
- Non-streaming path: `collectChatCompletion()` surfaces `ChatCompletionsStreamError`, and `handleChatCompletions` returns a real error status + OpenAI-style error body instead of a 200 completion.

### 2. `output_item.done` final arguments reconciled (last-write-wins)
- When a tool call was already registered from `output_item.added`, the done-frame `item.arguments` snapshot is now emitted as the authoritative final arguments (last-write-wins), instead of being dropped.

### Tests added
- done-frame final arguments reconcile
- stream `response.failed` → error frame, no clean `[DONE]`
- truncated stream → error frame, no clean `[DONE]`
- non-stream e2e failure → error status + body
- stream e2e failure → error frame, no clean `[DONE]`

Local: `bun test tests/chat-completions-endpoint.test.ts` → **16 pass / 0 fail**.

Ready for re-review. Thanks!

HaydernCenterpoint:
Follow-up fix on head for ChatGPT / Copilot App tool-call history:

**Bug:** `CAPIError: 400 tool_calls entries require function.name`

**Cause:** after `output_item.added`, subsequent args/done deltas sometimes emitted `function.arguments` without `function.name`. Clients that *replace* (not merge) the function object then stored nameless `tool_calls` and the next turn 400'd in `chatCompletionsToResponsesBody`.

**Fix:**
- Always re-emit `function.name` on args deltas and done-frame snapshots (`src/chat/outbound.ts`)
- Recover missing names from earlier same-`call_id` function_calls on inbound (`src/chat/inbound.ts`)
- Tests: named deltas + call_id name recovery

`bun test tests/chat-completions-endpoint.test.ts` → **18 pass / 0 fail**

HaydernCenterpoint:
CI typecheck fix pushed on head:

`src/chat/outbound.ts`
- Narrow `toolIndex` after `Map.get` (was `number | undefined` into `Map.set`/`get`)
- Avoid `ReadableStreamReadResult` assignability mismatch in `collectChatCompletion`

Local: `bun x tsc --noEmit` → clean.

Ingwannu (approved):
Re-reviewed the final head after all requested fixes. Parallel tool deltas now route by item id, done-frame arguments are reconciled, direct-mode identity/admission semantics match Responses, routed structured output fails explicitly, native request logs finalize, and failed/truncated internal streams no longer become successful assistant text. I also removed the trailing diff-check whitespace with the contributor branch’s maintainer-edit permission. The five-PR current-dev integration stack passed 3,667 tests / 0 failures, and this final head passed Linux, macOS, Windows, npm-global, typecheck, GUI, and CodeRabbit checks. The lone label failure is a fork-token 403 while applying a label, not a code failure. Approved.

Issue #92, opened 2026-07-11
V2 cross-provider sub-agent loses NEW_TASK body in encrypted_content


webmastertorch:
## Current status

This remains an **upstream-tracking issue**, not an unfixed ordinary OpenCodex routing bug.

OpenCodex now has an experimental mitigation:

* PR lidge-jun/opencodex#1540 added an **opt-in encrypted V2 task recovery path** for eligible native Codex `NEW_TASK` messages before dispatch to a routed provider.
* Recovery is deliberately **disabled by default** and has strict credential, loopback, cache, timeout, and fail-closed boundaries.
* PR lidge-jun/opencodex#1556 fixed fallback routing after a recovered task becomes readable, so a healthy routed fallback can be selected after recovery.

Those changes make some native-parent -> routed-child V2 workflows usable, but they do **not** change the underlying wire contract. The original task can still arrive only as backend ciphertext, and a routed provider cannot decrypt that ciphertext itself.

PR lidge-jun/opencodex#1540 explicitly treats recovery as a compatibility path rather than a protocol fix. The upstream plaintext/task-delivery limitation therefore remains the close condition for this issue.

## Summary

With `multiAgentMode: "v2"`, a native Codex parent can select a routed child model, but the child task may be delivered only through `encrypted_content`. A third-party routed provider cannot read backend ciphertext and therefore cannot execute the task without an additional compatibility mechanism.

The original reproduction used:

* parent: native `gpt-5.6-sol`
* child: routed `xai/grok-4.5`
* `fork_turns: "none"`
* V2 collaboration mode

The child model override was applied correctly, but the `NEW_TASK` payload was not available as plaintext.

## Current expected behavior

OpenCodex should preserve the following contract:

1. Native V2 passthrough remains unchanged.
2. Routed V2 tasks that are already readable continue normally.
3. An unreadable encrypted task fails closed unless the user has explicitly enabled an eligible recovery path.
4. Experimental recovery must never silently weaken credential-destination, loopback, cancellation, privacy, or cache boundaries.
5. Recovery failure must preserve the clear `unreadable_encrypted_agent_task` class rather than sending ciphertext to the routed model as if it were usable input.

## User-visible mitigation

For heterogeneous native-parent -> routed-child delegation, users currently have two practical choices:

* use V1/plaintext-compatible collaboration behavior; or
[...]

PR #283, opened 2026-07-22
fix(v2): fail fast on unreadable routed agent tasks

MathiasHeinke:
## Summary

Fail fast when a Codex V2 `agent_message` contains a native ChatGPT Fernet task token but no provider-readable task payload and the selected route is external.

Instead of forwarding an empty task to Grok, Claude, Kimi, or another routed provider, OpenCodex now returns a deterministic, non-retryable HTTP 400 before auth resolution, adapter construction, or provider dispatch. The error directs the caller to plaintext V2 agent-message delivery or a native ChatGPT model.

This is a companion safety fix for #92. It does not claim to recover the missing task body.

## Root cause and scope

The native ChatGPT backend can decrypt backend-minted Fernet tokens. An external routed provider cannot. By the time OpenCodex receives this request, the plaintext task is absent, so the proxy has no cryptographically valid way to reconstruct it.

This PR intentionally does **not** attempt to decrypt or reinterpret that ciphertext. Doing so would require keys OpenCodex does not own and would blur the trust boundary. The actual end-to-end fix remains plaintext V2 task delivery in the Codex client.

The guard is deliberately narrow:

- only V2 `agent_message` input is inspected
- only an exact Fernet-shaped `encrypted_content` task with no readable payload is rejected
- native canonical ChatGPT forwarding remains byte-preserving so the backend can decrypt it
- readable plaintext task content still routes normally
- the plaintext-in-`encrypted_content` compatibility path from related PR #94 remains intact
- unrelated encrypted reasoning and compaction content is ignored
- the response never echoes ciphertext or request-derived content

## Why fail fast

Without this guard, the model override can look successful while the routed child receives no actionable task. That can produce misleading worker output and repeated provider-cost attempts. A clear 400 makes the delivery mismatch observable and non-retryable until the upstream client emits plaintext task content.

## Verification

- `bun test --isolate tests/multi-agent-compat.test.ts tests/v2-agent-message-failfast.test.ts` — 31 pass, 0 fail
- `bun run typecheck` — pass
- `git diff --check origin/dev...HEAD` — pass

Refs #92. Related: #94.

### DIFF
--- a/src/server/responses.ts
+++ b/src/server/responses.ts
@@ -58,6 +58,7 @@ import {
 import { fetchWithResetRetry, fetchWithTransientRetry, applyUpstreamRecoveryInit } from "../lib/upstream-retry";
 import { ForwardAdmissionCredentialError, validateForwardAdmissionCredential } from "./auth-cors";
 import { listOpenAiForwardSidecarCandidates, resolveFirstUsableOpenAiSidecar, type ResolvedOpenAiForwardSidecar } from "../providers/openai-sidecar";
+import { isCanonicalOpenAiForwardProvider } from "../providers/openai-tiers";
 import { applyOpenAiVirtualModel, resolveOpenAiCompactModel } from "../providers/openai-virtual-models";
 import { isUsageDebugEnabled } from "../usage/debug";
 import { readJsonRequestBody, DecompressedBodyTooLargeError, UnsupportedContentEncodingError } from "./request-decompress";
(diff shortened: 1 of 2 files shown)


lidge-jun:
Draft review — the fail-fast direction is right: current `dev` forwards genuine Fernet ciphertext that a non-ChatGPT provider can never decrypt, so the routed child proceeds with only the envelope and no actual task. A pre-network 400 is cheaper and more honest. Guard placement (after model routing, before auth/adapter dispatch, native ChatGPT path untouched) and the test coverage (zero external dispatch, non-retryable 400, no ciphertext reflection, native byte-preserving forward) both look good.

Feedback before marking ready:

1. `readableAgentMessagePayload()` joins text parts and discards everything before the first `Message Type:` line. Please add fixture evidence (real codex-rs captures) that there's no legitimate shape where meaningful task text precedes the envelope — otherwise this could false-reject.
2. The exact Fernet regex + envelope parser are capture-dependent. Pin the real captures as fixtures (or test comments), and make the intent explicit for array-vs-string content variants and payload-on-same-line cases, so upstream wire-shape drift fails loudly instead of silently not detecting.
3. Only `enforce-target` has run — needs the full CI matrix before ready-for-review.

With those locked down this is a good candidate for the #92 mitigation path.

Ingwannu (changes requested):
The fail-fast direction is valuable, and pure Fernet-only routed task slots correctly stop with a non-retryable 400 before provider dispatch. However, the current guard still fails open for the mixed shape that existing compatibility logic explicitly produces: plaintext hook preamble plus an encrypted actual task slot.

`readablePayloadAfterMarker()` counts any text after `Payload:` as provider-readable. With a normal `NEW_TASK` envelope followed by `[CXC-LEAF-GUARD] ...` plaintext and then a Fernet slot, sanitization removes the ciphertext, the guard sees the leaf-guard preamble as readable payload, and the routed provider is called with HTTP 200 even though the actual task body is gone. The captured provider message contains only the envelope and leaf guard. This preserves the core #92 failure mode and contradicts the PR guarantee that unreadable routed tasks stop before dispatch.

Please classify/strip known compatibility preambles before deciding that a task payload is readable, and add the existing mixed-slot fixture as a fail-fast regression. Also pin a real fixture for any plaintext allowed before `Message Type:`; the current helper can false-reject meaningful pre-envelope text.

Full local current-dev merge verification otherwise passed (3,512 tests, typecheck, privacy scan, GUI lint/build). No malware, backdoor, ciphertext reflection, unsafe I/O, or dependency concern was found.

lidge-jun:
**Post-merge Sol review note:** This PR was merged then reverted from `dev` after Sol review found a critical bypass:
- Mixed encrypted slots (e.g. [CXC-LEAF-GUARD] + Fernet token) bypass the guard entirely
- FERNET_TOKEN_EXACT regex accepts structurally impossible tokens
- Request-wide scanning can block readable turns due to encrypted history

The fail-fast concept is important for security and will be rebuilt with proper mixed-slot handling and structural Fernet validation.

PR #284, opened 2026-07-22
fix: verify GUI update restarts stay healthy

Ingwannu:
## Summary
- verify GUI update restarts with an identity-checked health/stability window before marking update-job success
- fail the job with a Windows-specific recovery hint when the replacement proxy never becomes healthy or dies shortly after start
- cover the new restart confirmation lanes with targeted updater regression tests and update the worker ADR

## Root cause
The GUI update worker marked `update-job.json` as succeeded immediately after requesting a restart. On Windows, npm/Bun update failures can leave the replacement proxy dead within seconds, but the dashboard still reported success because no post-restart health confirmation existed.

## User impact
Users now get an honest failed update-job state plus recovery guidance instead of a silent "restart requested" success while the proxy is already down.

## Validation
- `bun test tests/update-job.test.ts`
- `bun run typecheck`
- `bun run privacy:scan`

### DIFF
--- a/docs/adr/0001-gui-update-worker.md
+++ b/docs/adr/0001-gui-update-worker.md
@@ -22,9 +22,17 @@ the existing npm self-update guard is reused. For Bun global installs, it runs t
 global update command. Source checkouts remain manual-only and show `git pull && bun install &&
 bun run build:gui`.
 
+After an update requests a restart, the worker now waits for an identity-checked `/healthz` to
+return and remain healthy for a short stability window before marking the job successful. This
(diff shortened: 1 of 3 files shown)


Ingwannu:
Final review passed on `d8200ea7`. Update success is now recorded only after restart plus OpenCodex `/healthz` identity confirmation; failures persist `restarted:false` with a recovery hint. Focused 36 tests, 3,511 full tests, typecheck, privacy scan, GUI lint, diff check, and all cross-platform CI pass. No new dependencies, credentials, unsafe shell/process surface, malware, or backdoor behavior found.

PR #296, opened 2026-07-22
Add Cursor Router optimization levels

jontonsoup:
## Summary

- expose Cursor Router's Cost, Balance, and Intelligence optimization levels as `cursor/auto-cost`, `cursor/auto-balance`, and `cursor/auto-intelligence`
- preserve `cursor/auto` as the backwards-compatible account/team default
- encode explicit levels through `AgentRunRequest.requested_model.parameters` while keeping every router entry available through static and live model catalogs
- document the model IDs and add adapter, protobuf, discovery, and Codex catalog regression coverage

## Why

[Cursor Router](https://cursor.com/blog/router) now offers three explicit optimization levels, but Codex cannot render Cursor-specific model parameters. Representing those choices as first-class model IDs makes the full routing ladder selectable from Codex without changing the existing `cursor/auto` behavior.

## User impact

Users can choose the desired Cursor routing tradeoff directly from the Codex model picker. Existing configurations using `cursor/auto` continue to use the Cursor account or team default.

## Validation

- `bun run typecheck`
- `bun test --isolate ./tests/` — 3,515 tests passed
- focused Cursor/catalog tests — 112 tests passed
- `bun run lint:gui`
- GUI production build
- Astro documentation production build
- `bun run privacy:scan`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added Cursor Router model options for cost, balance, and intelligence optimization.
  * Explicit optimization choices are now preserved in requests while using Cursor’s default model.
  * Router options remain available even when live model discovery does not list them.

* **Documentation**
  * Added guidance explaining Cursor Router options, request behavior, and model availability.

### DIFF
--- a/docs-site/src/content/docs/reference/adapters.md
+++ b/docs-site/src/content/docs/reference/adapters.md
@@ -98,6 +98,9 @@ streams the response back **untranslated**.
 - Replays conversation state through content-addressed blobs, maps server tool calls back to Codex,
   discovers live Cursor models through the protobuf `GetUsableModels` RPC, and retries only before a
   run request is committed to the wire.
+- Exposes Cursor Router as `cursor/auto` plus explicit `cursor/auto-cost`,
+  `cursor/auto-balance`, and `cursor/auto-intelligence` entries. Explicit levels are encoded in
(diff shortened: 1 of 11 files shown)


jontonsoup:
@coderabbitai review

PR #298, opened 2026-07-22
ci: add issue deduplicator workflow

Wibias:
## Summary

Adds `.github/workflows/issue-deduplicator.yml` — a workflow that fires on every newly opened issue and posts a bot comment listing up to five semantically similar open issues.

## How it works

1. **Fetch** — `gh issue list` pulls up to 500 open issues (number, title, first 800 chars of body) into a JSON file alongside the new issue
2. **Detect** — [`actions/ai-inference@v1`](https://github.com/actions/ai-inference) sends both files to `gpt-4o-mini` via GitHub Models with a structured JSON output schema, asking it to identify issues that share the same underlying problem, error signature, or feature request
3. **Normalize** — the response is parsed, self-references stripped, capped at 5
4. **Comment** — a bot comment is upserted (idempotent via `<!-- opencodex-dedup-bot -->` marker)

Maintainers can also manually re-trigger a scan on any issue by adding the `codex-deduplicate` label — it's removed automatically after the scan completes.

## Why `actions/ai-inference` instead of keyword search

The first draft of this used stop-word keyword extraction + `github.rest.search.issuesAndPullRequests`. That works but misses semantic matches — "reasoning effort `max` not available" and "provider dropdown not showing correct levels" describe the same bug but share no keywords. `actions/ai-inference` uses an actual LLM to judge intent, not token overlap.

## Why not `openai/codex-action` (like openai/codex uses)?

Their workflow requires a `CODEX_OPENAI_API_KEY` repo secret and an `issue-triage` environment configured by the repo owner — real setup friction and real per-run cost. `actions/ai-inference` uses the built-in `GITHUB_TOKEN` with `models: read` permission via GitHub Models. **No secrets, no extra setup, zero cost to the repo owner** — just merge this file.

## Required permissions (already declared in the workflow)

```yaml
permissions:
  contents: read
  issues: write
  models: read
```

## What the comment looks like

> Potential duplicates found — please check whether your issue is already covered:
>
> - #297
> - #241
>
> _Detected automatically via GitHub Models. Matches may not be exact._

## Summary by CodeRabbit

* **New Features**
  * Added an “Issue Deduplicator” GitHub Actions workflow that runs when issues are opened and when a specific deduplication label is applied.
  * Uses AI-powered semantic matching to propose up to five similar issues by comparing against a capped set of existing repository issues (including closed).
  * Posts or refreshes a single persistent bot comment with a fixed marker, and flags matches that are already closed.
  * Clears the deduplication label after processing using a best-effort approach.

PR #299, opened 2026-07-22
ci: add issue translator workflow

Wibias:
## Summary

Adds `.github/workflows/issue-translator.yml` — a workflow that fires on every newly opened issue and, when the title or body is primarily non-English, automatically translates it to English.

**No secrets. No setup. Zero cost to the repo owner.** Uses `actions/ai-inference@v1` with the built-in `GITHUB_TOKEN` and `models: read` permission via GitHub Models — same approach as #298.

## What it does

1. Runs `gpt-4o-mini` via GitHub Models on the issue title + body
2. Detects whether the primary language is non-English
3. If yes:
   - **Updates the issue title** to the English translation (within GitHub's 256-char limit)
   - **Posts a bot comment** with the full English body, noting the detected language

The comment is marker-guarded (`<!-- opencodex-issue-translator -->`) so re-runs never double-post. A per-issue concurrency group prevents races.

## Why translate the title specifically

The title update is the most important part. It means:

- Maintainers can triage non-English issues at a glance without opening them
- GitHub's own search indexes the English title, so issues are findable
- **The issue deduplicator (PR #298) works correctly across languages**

That last point is worth explaining: the deduplicator fetches open issues by title and body, then asks an LLM to find semantic duplicates. A Korean issue reporting the exact same `max`/`ultra` reasoning effort bug as an open English issue will never be flagged as a duplicate — the model sees two completely different languages describing the same thing. Translating the title *before* deduplication runs (both workflows trigger on `issues: opened`, and translation is fast) means the deduplicator's candidate set is in consistent English, dramatically improving cross-language duplicate detection.

## Inspired by

`openai/codex` [`issue-translator.yml`](https://github.com/openai/codex/blob/main/.github/workflows/issue-translator.yml) — but replaces their `openai/codex-action` (requires `CODEX_OPENAI_API_KEY` secret + `issue-triage` environment) with `actions/ai-inference`, which uses only the built-in `GITHUB_TOKEN`.

## Required permissions (declared in the workflow)

```yaml
permissions:
  issues: write   # update title, post comment
  models: read    # GitHub Models inference
```

## What the comment looks like

> **English translation** *(original language: Korean)*
>
> [full translated body]

## Summary by CodeRabbit

* **New Features**
  * Added an automated workflow that detects whether newly opened issue titles and descriptions are primarily non-English.
  * When translation is required, it updates the issue title (only if unchanged since the event) and posts an “English translation” comment with the translated content.
  * Prevents duplicate translation comments and safely handles invalid or missing translation results; skips comments when no translated body is available.

### DIFF
--- a/.github/workflows/issue-translator.yml
+++ b/.github/workflows/issue-translator.yml
@@ -23,7 +23,7 @@ jobs:
         uses: actions/ai-inference@b81b2afb8390ee6839b494a404766bef6493c7d9 # v1
         with:
           model: openai/gpt-4o-mini
-          max-completion-tokens: 4000
+          max-tokens: 4000
           system-prompt: >
             You are a GitHub issue translator. Your only job is to detect the
             primary language of the issue and, when it is not English, produce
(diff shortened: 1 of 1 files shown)


lidge-jun:
**Post-merge Sol review note:** Reverted from `dev` — same cross-workflow sequencing issue as #298. Will be rebuilt together with the deduplicator as one coordinated workflow.

PR #303, opened 2026-07-22
Document Cursor exec policy and catalog troubleshooting

diegocantarero:
## What changed

- document the current `nativeLocalExec` modes for the experimental Cursor adapter
- mark `unsafeAllowNativeLocalExec` as the legacy compatibility option
- explain the fail-closed behavior of `codex-sandbox` and the loopback trust boundary
- add catalog troubleshooting for discovery filters, Cursor `GetUsableModels`, cache refresh, and `ocx sync`
- clarify that competing local writers and Codex model-cache refreshes can make model visibility appear unstable without implying an OpenCodex defect

## Why

The runtime already implements `nativeLocalExec: "off" | "codex-sandbox" | "on"`, but the public configuration reference documented only the legacy boolean. That made the safest current configuration and precedence rules difficult to discover.

Catalog visibility can also look intermittent when filters, cached discovery, or another local writer changes the managed files. The new troubleshooting section keeps those causes separate and explicitly identifies external writers as an operator-side hazard rather than a confirmed OpenCodex bug.

## Safety

- `off` remains the documented safest default
- `on` is described as a trusted-local opt-in that bypasses Codex approval and sandbox semantics
- `codex-sandbox` is documented as currently fail-closed because caller-controlled request text cannot attest a real sandbox
- no runtime, authentication, credential, or provider behavior changes

## Validation

- `bun run prepush` — passed
- 3513 tests passed, 0 failed
- typecheck passed
- GUI lint passed
- privacy scan passed
- docs-site Astro/Starlight build passed (121 pages)
- `git diff --check` passed

## Localization

This PR updates the English source documentation. Existing localized pages still describe the functional legacy boolean and can be synchronized separately by fluent contributors.

### DIFF
--- a/docs-site/src/content/docs/guides/codex-integration.md
+++ b/docs-site/src/content/docs/guides/codex-integration.md
@@ -127,6 +127,33 @@ Routed catalog entries also get their GPT-5 identity rewritten to the real upstr
 Reasoning controls come from provider/model metadata across Codex's `low | medium | high | xhigh |
 max | ultra` ladder; unsupported values are mapped or clamped before the upstream request.
 
+### Catalog troubleshooting
+
+If a model is missing from Codex, or the catalog order/visibility looks wrong, check in order:
+
+1. **`selectedModels`** on the provider — a non-empty allowlist exposes only those ids to Codex;
+   empty or omitted exposes all discovered models. An id not in the allowlist never reaches the
+   catalog.
+2. **`disabledModels`** (top level) — hides models from both the catalog and `/v1/models`, and flips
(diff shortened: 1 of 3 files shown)


Ingwannu:
This documentation is accurate and worth keeping, but #307 made the stale-base draft conflict in the same guide. I preserved your original commit authorship and moved it onto current `dev` as #321, resolving the conflict by keeping both the custom-display-name and troubleshooting sections. The replacement builds all 121 docs pages and was included in the combined 3,667-test integration pass. Closing this draft in favor of #321.

PR #307, opened 2026-07-23
Preserve custom model display names across catalog syncs

diegocantarero:
## Summary

- preserve `customModels[].displayName` when OpenCodex regenerates the Codex model catalog
- change only the picker label (`display_name`), leaving routing slugs, aliases, providers, and native OpenAI marketing names untouched
- document repeated-sync and managed-restart behavior, including the best-effort offline-boot fallback

## Root cause

OpenCodex already accepted and persisted `displayName` for custom models, but `gatherRoutedModels()` dropped it before catalog generation. Each `ocx sync` therefore rebuilt the row with its routed slug as the visible label.

This change carries the configured label into `CatalogModel` and applies it only to the generated catalog row `display_name`.

## Restart behavior

The managed service attempts a catalog sync after the proxy binds on startup. A successful restart therefore reapplies configured display names from `config.json`. If that best-effort sync cannot complete, the previously persisted catalog remains in place and the next successful `ocx sync` reapplies the label.

## Safety

- no authentication, credential, launchd, routing, alias, provider, or native-model behavior changed
- no custom model catalog is made global
- display names cannot alter the provider/model slug

## Validation

- `bun run typecheck`
- `bun run prepush`: 3,520 tests passed, GUI lint passed, privacy scan passed
- `bun test tests/codex-catalog.test.ts`: 90 tests passed
- `bun run build` in `docs-site`: 121 pages built
- `git diff --check`

## Regression coverage

- configured label reaches `display_name`
- absent or whitespace-only label retains historical fallback
- slug, alias, and provider remain unchanged
- repeated catalog regeneration is idempotent
- native OpenAI marketing names remain untouched
- `customModels` propagation is covered end to end

Authored by Diego Cantarero.

## Summary by CodeRabbit

* **New Features**
  * Added support for custom model display names in the Codex model picker.
  * Display names can be configured through the CLI or proxy management API.
  * Custom names are preserved during catalog refreshes and syncs.
  * Display names affect presentation only and do not change routing, aliases, or model identity.

* **Documentation**
  * Added setup, validation, persistence, and fallback behavior guidance for custom model display names.

### DIFF
--- a/docs-site/src/content/docs/guides/codex-integration.md
+++ b/docs-site/src/content/docs/guides/codex-integration.md
@@ -127,6 +127,32 @@ Routed catalog entries also get their GPT-5 identity rewritten to the real upstr
 Reasoning controls come from provider/model metadata across Codex's `low | medium | high | xhigh |
 max | ultra` ladder; unsupported values are mapped or clamped before the upstream request.
 
+### Custom model display names
+
+A custom model can carry a human-readable **display name** that overrides the label Codex shows in
+its model picker, without changing anything about how the model is routed. The display name maps to
+the catalog entry's `display_name` field only — the routing slug (`<provider>/<model>`), alias
+collision order, provider, and native OpenAI marketing names are all left untouched.
+
+Add a display name from the CLI (the proxy syncs the catalog right away when live):
+
+```bash
+ocx models add deepseek deepseek-v4 --display-name "DeepSeek V4" --context-window 128000
+```
+
+You can also set or edit it through the management API (`POST /api/custom-models`,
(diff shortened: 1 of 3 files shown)


Ingwannu (approved):
Reviewed against current `dev` and the recent catalog changes. The new `displayName` stays presentation-only: it updates `display_name` after catalog identity is derived and does not alter the routed slug, alias, provider, native snapshot rows, or discovery behavior. Integrated regression coverage passed, including the current #308/#310 catalog cases. Full combined pre-push: 3,667 passed / 0 failed; typecheck, privacy scan, and docs build also passed. Approved.

PR #309, opened 2026-07-23
fix(google): comprehensively eliminate Antigravity request-shape 400s

HaydernCenterpoint:
## Summary

This introduces a single conservative wire compiler for every Google-family request and comprehensively fixes the known Antigravity `400 invalid request` class caused by request-shape incompatibilities.

- Compile AI Studio, Vertex, and Cloud Code Assist payloads through one final trust boundary before serialization.
- Normalize arbitrary MCP / JSON Schema input into Google's supported function-schema subset, including `x-mcp-header`, draft 2020-12 annotations, nullable `anyOf`, local refs, hostile values, and recursion limits.
- Clamp unsupported Antigravity thinking levels (`xhigh`, `max`, `ultra`) to Google's highest accepted `high` value.
- Normalize invalid, long, or colliding tool names and restore the original name when the model calls the tool.
- On a provider schema/thinking `400`, rebuild a materially changed request and replay it exactly once; unrelated `400` responses remain non-retryable.

## Root cause

Google request compatibility was previously enforced incrementally while building the payload. There was no final compiler guaranteeing wire invariants, so one provider-specific field or one third-party MCP schema could invalidate the entire tool catalog. The transport also treated every `400` as terminal even when the provider identified a safely repairable schema or thinking field.

## Why this is comprehensive for Antigravity request-shape failures

The strict allowlist prevents unknown third-party schema keywords from reaching Antigravity, while the final compiler also covers generation config and tool-name constraints. If the Claude bridge still rejects a documented construct, the targeted open-schema fallback repairs the named tool and replays once. This closes the observed `thinking_level=max`, `x-mcp-header`, Serena nullable-union, and generic `tools.N.custom.input_schema` failure modes without depending on Serena or any specific MCP server.

This scope intentionally does not claim to suppress authentication, quota, permission, network, or upstream availability failures; those are separate provider conditions rather than malformed-request bugs.

## Impact

- A single malformed MCP tool can no longer poison every tool in the request.
- Users without Serena are unaffected; all MCP catalogs receive the same provider-boundary normalization.
- Other providers are unchanged.
- No new dependency is added.

## Validation

- `bun run typecheck`
- 128 Google / Vertex / Antigravity tests passed across 10 files.
- Full local suite: 3453 passed, 4 skipped; the only 2 failures are the existing Windows Cursor `printf` fixtures.
- Live isolated-dev E2E against Antigravity: 112 tools (22 Serena, 7 Yith, 20 adversarial schemas), with global `tools.64 = serena_replace_in_files`, returned HTTP 200 and `OK.`.
- `git diff --check`

References:

- [Google function-calling schema contract](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling)
- [Google retry strategy](https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/retry-strategy)

## Summary by CodeRabbit

* **New Features**
  * Improved compatibility with Google and Gemini request formats.
  * Automatically retries certain invalid requests after safely repairing unsupported tool schemas or thinking settings.
  * Preserves tool names across requests and responses, including names requiring compatibility formatting.
  * Normalizes reasoning effort values to supported provider settings.

* **Bug Fixes**
  * Improved handling of unsupported and complex tool parameter schemas.
  * Prevented invalid schema fields from being sent to Google services.

### DIFF
--- a/src/adapters/google-http.ts
+++ b/src/adapters/google-http.ts
@@ -1,5 +1,6 @@
 import type { AdapterFetchContext, AdapterRequest } from "./base";
 import { isQuotaExhaustedBody, retryableGoogleStatus, safeGoogleHttpErrorMessage } from "./google-errors";
+import { repairGoogleInvalidRequestBody } from "./google-wire-compiler";
 import { normalizeUpstreamHttpErrorResponse, readDisplaySafeErrorPayloadText } from "./upstream-http-error";
 import {
   abortError,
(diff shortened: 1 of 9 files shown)


Ingwannu (approved):
Reviewed against current `dev`. The wire compiler is a bounded allowlist rather than a permissive pass-through; tool-name encoding is deterministic and reversible per request, schema refs are local and depth-limited, and the compatibility replay is limited to one changed request after a matching 400. The adapter is instantiated per request, so the restore-name closure is not shared across concurrent turns. Focused Google tests passed in the combined stack, and the full pre-push suite passed 3,667 / 0 with typecheck and privacy scan. Approved.

PR #318, opened 2026-07-23
fix(cursor): replay external model tool continuations with fresh conversation ids

Wibias:
## Summary

Cursor Connect tool-result continuations for external models like `cursor/gpt-5.6-sol` were reusing the same conversation id via `resumeAction`. That works for Cursor-native models (`composer-*`, `auto`/`default`), but external models often lose server-side conversation state and then fail after `stepCompleted` with:

`connect-end-stream invalid_argument`
followed by reconnect + `getBlobArgs` burst + another failed follow-up turn.

Native openai `gpt-5.6-sol` and Cursor `composer-2.5` stayed stable. The bug is specific to external Cursor Connect continuation.

## Root cause

For tool-result-only turns, OpenCodex correctly uses `resumeAction` and structured conversation history. It previously kept the remembered `_cursorConversationId` for every model.

For external Cursor models, that resume path is brittle: Cursor hydrates blob/checkpoint state, then rejects the next step with a generic `invalid_argument` instead of emitting `turnEnded`. OpenCodex then reconnects and can loop.

## Fix

1. Detect Cursor-native vs external wire models (`composer-*`, `auto`/`default` = native).
2. For external-model tool-result continuations, mint a fresh `conversationId` while still sending full replay history (`conversationTurns` + `rootPromptMessagesJson`) under `resumeAction`.
3. Keep native Cursor models on the remembered conversation id path.
4. One-shot adapter retry: if an external model still hits `invalid_argument`, rebuild once with `forceFreshConversation: true`.
5. Sync `_cursorConversationId` to the conversation id actually used for the turn.
6. Add debug diagnostics for `wireModel`, `action`, `conversationId`, turn type, and `externalModel`.

## Why this is not "use openai instead"

Sol remains on the Cursor provider. Only the continuation strategy changes for external Cursor models.

## Tests

- `bun run typecheck`
- `bun test tests/cursor-adapter.test.ts tests/cursor-request-builder.test.ts tests/cursor-discovery.test.ts tests/cursor-errors.test.ts tests/cursor-tool-continuation.test.ts`

Added coverage for:
- external tool-result continuation forces a fresh conversation id
- native tool-result continuation keeps the remembered id
- `forceFreshConversation` always mints a new id
- one-shot invalid_argument retry with a second conversation id
- native/external wire-model classification
- invalid_argument detector

## Files

- `src/adapters/cursor.ts`
- `src/adapters/cursor/discovery.ts`
- `src/adapters/cursor/request-builder.ts`
- `src/adapters/cursor/protobuf-request.ts`
- `src/adapters/cursor/cursor-errors.ts`
- matching Cursor tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Cursor conversation handling for tool-result continuations, preventing conversation-state corruption across external-model replay.
  * Added a one-time retry for specific Cursor `invalid_argument` failures that can occur during resume, using a fresh conversation identifier when safe.
  * Preserves existing conversation continuity when continuation is supported.
* **Diagnostics**
  * Enhanced Cursor run request diagnostics with action type, model type, and conversation identifiers.
* **Tests**
  * Added unit tests covering retry logic, conversation refresh behavior, model classification, and invalid-argument detection.

### DIFF
--- a/src/adapters/cursor.ts
+++ b/src/adapters/cursor.ts
@@ -6,7 +6,11 @@ import { isCursorExternalWireModel } from "./cursor/discovery";
 import { createCursorKvStore, type CursorKvStore } from "./cursor/kv-store";
 import { mapCursorServerMessage } from "./cursor/message-mapper";
 import { createCursorRequest, generatedCursorConversationId } from "./cursor/request-builder";
-import { createLiveCursorTransport, CursorMissingCredentialError } from "./cursor/live-transport";
+import {
+  createLiveCursorTransport,
+  CursorMissingCredentialError,
+  rekeyCursorContextUsage,
+} from "./cursor/live-transport";
 import { runCursorTurnWithRetry } from "./cursor/transport-retry";
 import {
   createDisabledCursorTransport,
@@ -71,10 +75,16 @@ export function createCursorAdapter(provider: OcxProviderConfig, deps: CursorAda
         const makeTransport = deps.createTransport ?? createLiveCursorTransport;
         const kv = deps.kv ?? createCursorKvStore();
         _parsed._cursorConversationId ??= generatedCursorConversationId();
+        const previousConversationId = _parsed._cursorConversationId;
         let request = createCursorRequest(_parsed);
         // Keep remembered conversation id in sync when the request builder mints a fresh id
         // for external-model tool-result continuations (stateless replay).
+        if (request.conversationId !== previousConversationId) {
+          rekeyCursorContextUsage(previousConversationId, request.conversationId);
+        }
         _parsed._cursorConversationId = request.conversationId;
+        let emittedOutput = false;
+        const lastRawIsToolResult = _parsed.context.messages.at(-1)?.role === "toolResult";
 
         const runOnce = async (activeRequest: ReturnType<typeof createCursorRequest>) => {
           await runCursorTurnWithRetry(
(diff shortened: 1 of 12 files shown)


Ingwannu (changes requested):
The fix is valuable, but the current one-shot fallback can replay already-committed output. `runOnce()` forwards every mapped event to `emit` immediately, while the catch block retries any qualifying external-model `invalid_argument` without checking whether text, thinking, or tool-call events were already emitted. If Cursor yields visible output and then throws, the second request replays the turn and duplicates/corrupts the client stream. The added test throws before yielding anything, so it does not cover this case.

Please track whether any non-heartbeat/output event has been forwarded and permit the fresh-conversation retry only before output is committed. Add a regression where attempt 1 emits text or a tool-call event and then throws `invalid_argument`; assert there is no second attempt and no duplicate output. Re-request review after that. I am leaving the PR open because the underlying external-model continuation fix is meaningful.

Ingwannu (approved):
Re-reviewed the updated head. The retry is now gated to an external-model tool-result continuation, `invalid_argument`, no abort, and no emitted non-heartbeat event. The new regression proves that text followed by `invalid_argument` does not trigger a second attempt, eliminating the duplicate-output blocker. Conversation-id rotation also rekeys context-usage carry-forward instead of dropping it. Integrated on current `dev`: 81 focused Cursor tests passed, typecheck and privacy scan passed; the PR head is green on Linux/macOS/Windows. Approved.

PR #339, opened 2026-07-23
[WRONG BRANCH] fix(adapters): preserve upstream finish_reason as stopReason in openai-chat adapter

xiangyingchang:
## Problem

The `openai-chat` adapter collapsed all non-empty `finish_reason` values into a boolean `sawFinish`, discarding the original reason. The bridge already maps `stopReason === "max_tokens"` to `response.incomplete` (since the Anthropic adapter was fixed for issue #246), but the `openai-chat` adapter never propagated the value.

This means OpenAI-compatible providers (e.g. `ark-coding/glm-5.2`) that return `finish_reason=length` when the model hits its output token cap were silently treated as normal completions. Codex received `task_complete` on truncated responses, and the user had to manually send "continue" to resume.

## Evidence

From local usage logs, `ark-coding/glm-5.2` had 16 requests with `outputTokens` exactly at 4096, all HTTP 200, all marked as `usageStatus=reported` — the truncation was invisible to the client.

## Fix

- Replace `sawFinish: boolean` with `finishReason: string | undefined` in the streaming path
- Propagate `finish_reason` as `stopReason` on all `done` events (streaming `[DONE]`, EOF, and non-streaming `parseResponse`)
- Preserve backward compat: `[DONE]` without `finish_reason` → `stopReason=undefined` → `response.completed`
- Preserve EOF fail-closed behavior: no `finish_reason` + no `usage` → `error` (unchanged)

## Mapping

The bridge (already in `main`) handles the terminal status mapping:

| `finish_reason` | `stopReason` | Responses terminal status |
|---|---|---|
| `stop` | `stop` | `response.completed` |
| `tool_calls` | `tool_calls` | `response.completed` |
| `length` | `length` | `response.completed`* |
| `max_tokens` | `max_tokens` | `response.incomplete` |
| `content_filter` | `content_filter` | `response.completed`* |
| missing | `undefined` | `response.completed` |

\* The bridge currently only maps `max_tokens` to `incomplete`. `length` and `content_filter` are not yet handled. This PR focuses on the adapter fix; bridge-side `length`/`content_filter` mapping can be added separately if needed.

## Tests

15 new tests in `tests/openai-chat-finish-reason.test.ts`:

- Streaming: `stop`/`length`/`max_tokens`/`tool_calls`/`content_filter` → correct `stopReason`
- Streaming: `[DONE]` after `finish_reason=length` preserves captured `stopReason`
- Streaming: `[DONE]` without `finish_reason` → `stopReason=undefined`
- Streaming: EOF without `finish_reason` → `error` (fail-closed)
- Streaming: usage-only final frame → `stopReason=undefined`
- Non-streaming: all `finish_reason` values + missing → correct `stopReason`

All 15 pass. 90 tests across 5 related files pass with 0 regressions. `tsc --noEmit` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved completion reasons from OpenAI responses in the final event for both streaming and non-streaming requests.
  - Correctly reports reasons such as stopping, length limits, token limits, tool calls, and content filtering.
  - Improved handling of streams that close gracefully without an explicit completion marker.
  - Streams that end without completion details continue to fail safely.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -604,7 +604,7 @@ export function createOpenAIChatAdapter(provider: OcxProviderConfig): ProviderAd
       // being reported as a clean completion (silent truncation). A graceful close is either an
       // explicit `[DONE]` sentinel OR a chunk carrying a non-null `finish_reason` (some
       // OpenAI-compatible providers omit `[DONE]` but do send finish_reason).
-      let sawFinish = false;
+      let finishReason: string | undefined;
 
       // Single per-line handler shared by the streaming loop and the EOF residual-frame flush, so
       // a final frame is parsed identically wherever it lands (no duplicated, drift-prone parsing).
@@ -615,7 +615,7 @@ export function createOpenAIChatAdapter(provider: OcxProviderConfig): ProviderAd
         const payload = line.slice(6).trim();
         if (payload === "[DONE]") {
           yield* flushToolCalls();
-          yield { type: "done", usage: pendingUsage };
+          yield { type: "done", usage: pendingUsage, ...(finishReason ? { stopReason: finishReason } : {}) };
           return "terminate";
         }
 
@@ -649,7 +649,7 @@ export function createOpenAIChatAdapter(provider: OcxProviderConfig): ProviderAd
         // Observe the terminator BEFORE the delta guard: a finish-only chunk (finish_reason set,
         // no delta) is a graceful close and must mark sawFinish even though we skip it below.
         if (typeof choices[0].finish_reason === "string" && choices[0].finish_reason) {
-          sawFinish = true;
+          finishReason = choices[0].finish_reason;
         }
         const delta = choices[0].delta;
         if (delta) {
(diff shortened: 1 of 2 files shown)


Ingwannu (changes requested):
The underlying truncation bug is real, but this PR targets main instead of dev and is now behind the terminal-truth work already on dev. It also preserves OpenAI finish_reason=length as length, while the bridge only marks max_tokens incomplete, so a capped response is still reported completed. Please rebuild on current dev, normalize length to max_tokens (or update the shared bridge contract with regression coverage), validate finish_reason types, and re-request review.

lidge-jun:
This PR targets `main` — per our branch policy, feature/fix PRs should target `dev` (`main` only moves via maintainer promotion). Could you retarget to `dev` and rebase? The head is also conflicting. Once retargeted and CI is green I'll take a proper look at the finish_reason → stopReason change.

lidge-jun:
Closing: this PR targets `main`, but feature PRs must target `dev` (see AGENTS.md branch policy). The `enforce-target` CI check is failing. Please retarget to `dev` and resubmit.

PR #352, opened 2026-07-23
fix(codex): bounded single pool retry on account-specific unsupported-model 400 (#335)

### LINKED ISSUE #335: Pool mode should retry another account after an account-specific unsupported-model response
## Summary

In Codex account Pool mode, an account-specific unsupported-model HTTP 400 ends the request immediately even when another configured account can use the requested model.

The paid main account is **not** missing access to `gpt-5.6-sol`, and account switching is **not** persistently stuck. Controlled testing showed that manually selecting the main account fixed the same visible Codex thread immediately, without restarting Codex. The remaining reproducible problem is the lack of bounded failover after the first pooled account rejects the model.

This is distinct from #297: the model catalog is correct and Sol exposes the expected reasoning levels. The failure occurs during pooled request routing.

## Environment

- OpenCodeX 2.7.35
- Canonical OpenAI provider in Pool mode
- Main ChatGPT Pro account plus a secondary pooled account
- Model: `gpt-5.6-sol`

## Confirmed behavior

With the secondary account selected, the existing Codex app thread produced an account-specific rejection:

```text
13:08:16Z  secondary route  gpt-5.6-sol  400
13:08:26Z  secondary route  gpt-5.6-sol  400
```

Upstream response:

```json
{"detail":"The 'gpt-5.6-sol' model is not supported when using Codex with a ChatGPT account."}
```

OpenCodeX returned that response directly. It did not retry the main paid account.

The active Pool account was then changed to `__main__` while the same visible Codex thread remained open. The next Sol request succeeded before any restart:

```text
13:09:07Z  main route  gpt-5.6-sol  200
```

Codex was then fully quit and relaunched while the OpenCodeX proxy remained running. After reopening the original troubleshooting thread, Sol still succeeded:

```text
13:12:47Z  main route  gpt-5.6-sol  200
```

## What this rules out

The controlled test does **not** support these earlier hypotheses:

- the paid main account lacks Sol access;
- changing the active account remains stuck;
- restarting Codex is required for the account change to take effect;
- the model catalog or reasoning-level clamp causes this request failure.

## Actual problem

OpenCodeX classifies generic 4xx responses as caller failures. That is appropriate for ordinary malformed requests, but this particular response describes an account/model capability mismatch. Pool mode therefore has another potentially eligible account but does not try it.

Diagnostics also make this harder to understand than necessary: request history exposes provider-style route labels, but not a clear dedicated field for the effective pooled account and thread affinity decision.

## Expected behavior

For this narrow unsupported-model response, Pool mode should treat the failure as account-specific and retry once with another eligible account. Unrelated HTTP 400 responses should remain non-retryable.

A bounded implementation could:

1. detect the upstream unsupported-model response without broadening generic 400 retry behavior;
2. clear or reconsider the request's account affinity;
3. rebuild authentication for another pooled account and retry once;
4. stop after one bounded pass when no account supports the model;
5. expose the effective pooled account/affinity decision in sanitized diagnostics.

## Suggested regression coverage

- unsupported-model 400 from account A retries account B and succeeds;
- unrelated 400 remains non-retryable;
- retry is bounded when every account rejects the model;
- the request log identifies the effective pooled account without exposing credentials;
- manual selection of `__main__` continues to route a fresh request through the main account.

Related: #297, #186, and #205.

lidge-jun:
## Summary

Fixes #335: in Pool mode, an account-specific unsupported-model 400 from the ChatGPT backend was returned directly to the client even when another pooled account could serve the model. The proxy now performs exactly ONE bounded retry with a different eligible pooled account.

## Design (intentional refinement of the 35d28a02 policy)

The `4xx = caller fault, no failover` classification from 35d28a02 is deliberately preserved: this change adds a narrow, allow-listed exception that never touches account health.

- **Allow-list matcher:** string equality (after whitespace/case normalization) against the exact backend sentence `The '<requested routed model id>' model is not supported when using Codex with a ChatGPT account.` — the interpolated model id must match the routed request. No substring/keyword/regex matching; regex injection is structurally impossible.
- **Bounded inspection:** the 400 body is inspected via a bounded clone read; timeout, truncation, oversize, or display-unsafe outcomes never authorize a retry and the original bytes reach the client unchanged.
- **Hard cap:** one retry, to a different credential/health-eligible account (request-local exclusion — global selection state is untouched). No eligible alternate → original 400 returned. Second allow-listed 400 → returned, no third dispatch.
- **Zero health mutation:** the pinned regression `caller and model 4xx responses do not penalize account health` stays green and unmodified (`src/codex/routing.ts` untouched).
- **Retry window:** strictly before any client-facing Response/relay construction, in both stream modes (legacy-tee and eager bounded relay).
- **WebSocket rebind:** a successful retry republishes the resolved auth context so the WS account registry migrates A→B (activation-tested with registry snapshots).

## Tests

Activations A–E plus a 10-test negative matrix (8 hostile-body classes + 2 retry-failure paths) as named tests in `tests/server-auth.test.ts` / `tests/codex-auth-context.test.ts` (+483 lines):

- typecheck exit 0; focused 129 pass / 0 fail; full suite 3879 pass / 0 fail (313 files); privacy:scan pass.

## Security review

This touches pool account-selection logic (MAINTAINERS.md security boundary). Self-audit: exact model-bound equality, bounded clone inspection, single alternate dispatch, unchanged caller-health policy, no new body/token/account logging. **Explicit maintainer security review requested before merge.**

Design/audit trail: `devlog/_plan/260724_bugfix_train/020_pool_retry_400.md` (security-focused adversarial audit, 2 rounds).

Closes #335.

### DIFF
--- a/src/codex/auth-context.ts
+++ b/src/codex/auth-context.ts
@@ -7,7 +7,11 @@ import {
 import { markAccountNeedsReauth } from "./account-runtime-state";
 import { isCodexAccountUsable } from "./account-usability";
 import { MAIN_CODEX_ACCOUNT_ID, getMainAccountToken } from "./main-account";
-import { getCodexAccountCooldownUntil, resolveCodexAccountForThreadDetailed } from "./routing";
+import {
+  getCodexAccountCooldownUntil,
+  pickLowestUsageCodexAccount,
+  resolveCodexAccountForThreadDetailed,
+} from "./routing";
 import { getAccountQuota } from "./quota";
 import type { CodexAccountMode, OcxConfig, OcxProviderConfig } from "../types";
 import { FORWARD_HEADERS } from "../adapters/openai-responses";
@@ -89,17 +93,29 @@ export function shouldMarkAccountNeedsReauthForCodexAuthFailure(cause: unknown):
   return !(cause instanceof CodexCredentialGenerationConflictError) && !(cause instanceof CodexCredentialRefreshLockTimeoutError);
 }
 
+export interface ResolveCodexAuthContextOptions {
+  excludeAccountId?: string;
+}
+
 export async function resolveCodexAuthContext(
   headers: Headers,
   config: OcxConfig,
   mode: CodexAccountMode,
+  options: ResolveCodexAuthContextOptions = {},
 ): Promise<CodexAuthContext> {
   if (mode === "direct") {
     if (!hasCallerCodexBearer(headers)) throw new CodexDirectAuthenticationError();
     return { kind: "main", accountId: null };
   }
   const threadId = headers.get("x-codex-parent-thread-id");
-  const resolution = resolveCodexAccountForThreadDetailed(threadId, config);
+  const resolution = options.excludeAccountId
+    ? (() => {
+        const accountId = pickLowestUsageCodexAccount(config, options.excludeAccountId);
+        return accountId
+          ? { status: "selected" as const, accountId }
+          : { status: "none" as const };
+      })()
+    : resolveCodexAccountForThreadDetailed(threadId, config);
   if (resolution.status === "expired") throw new CodexThreadAffinityExpiredError(resolution.accountId);
   const accountId = resolution.status === "selected" ? resolution.accountId : null;
   if (!accountId) throw new CodexPoolAuthenticationError();
--- a/src/server/responses/core.ts
+++ b/src/server/responses/core.ts
@@ -49,6 +49,7 @@ import {
   headersForCodexAuthContext,
   isCodexAuthContextUsable,
   resolveCodexAuthContext,
+  stripCodexRuntimeProviderFields,
   type CodexAuthContext,
 } from "../../codex/auth-context";
 import {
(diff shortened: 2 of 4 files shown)


Wibias:
## Dual review (security + bugs/edge cases)

No unresolved bot review threads (**CodeRabbit rate-limited**; 0 inline comments). CI is green. This touches pool account-selection (MAINTAINERS.md security boundary) — independent security + edge-case pass below.

### Security categories checked
No malware, backdoor, hidden C2, local admission bypass, classic XSS, or new credential/body logging. Retry authorization is server-side only: the allow-list matches a **cloned, bounded** upstream 400 body against an exact model-bound sentence (whitespace/case normalized), never against client-supplied text. Regex injection is structurally avoided (string equality after normalize). `recordCodexUpstreamOutcome(..., 400)` is a no-op for health (`caller` class early-returns in `src/codex/routing.ts`), so the “zero health mutation” claim holds. Failed alternate resolution fails closed to the original 400.

### Blockers
None for merge from a security-boundary perspective, assuming the sticky-affinity tradeoff below is accepted or fixed.

### Should-fix before merge

1. **Successful retry does not rebind thread affinity (Medium)** — `src/codex/auth-context.ts` + `src/server/responses/core.ts`  
   With `excludeAccountId`, selection uses `pickLowestUsageCodexAccount` and intentionally skips `resolveCodexAccountForThreadDetailed` / `bindThreadAffinity` (tests assert `activeCodexAccountId` stays on A). After A→B succeeds, HTTP thread affinity can remain on A, so **every later turn** on that thread pays another A 400 + B retry. WS registry migration is covered; HTTP affinity is not.  
   Prefer: on successful alternate response, rebind affinity for `x-codex-parent-thread-id` to B (still without health penalties / without changing global active unless you want that).

2. **Retry dispatch skips transient-5xx absorption (Medium)** — `src/server/responses/core.ts`  
   The first hop uses `fetchWithTransientRetry`; the alternate hop uses bare `fetchWithHeaderTimeout`. A flaky 502/520 on B becomes an immediate client failure instead of the same short absorb window. Wrap the retry fetch in the same transient helper (still hard-capped at one account switch).

3. **Hardcoded `"pool"` on retry (Low)** — retry calls `resolveCodexAuthContext(..., "pool")` and `applyCodexAuthContextToProvider(..., "pool")` instead of `route.codexAccountMode`. Safe today because the gate is `usesCodexForwardPoolAuth`, but prefer the route mode for consistency / future modes.

### Non-blocking / already solid
- Allow-list + bounded body (`displaySafe` / `!truncated`) + single alternate + negative matrix (hostile bodies, wrong model id, oversize, timeout) look carefully designed.
- Transport failure on B records only B and does not triple-dispatch (tested).
- No eligible alternate returns the original 400 unchanged (tested).

### Suggested fix order
1. Rebind thread affinity to the successful retry account (add HTTP affinity regression beside WS-REBIND-01).  
2. Run the alternate hop through `fetchWithTransientRetry`.  
3. Pass `route.codexAccountMode` instead of a literal `"pool"`.

Maintainer security sign-off still recommended given the pool-selection adjacency, but this review did not find an auth bypass or health-policy regression.

Issue #320, opened 2026-07-23
[Bug] OpenAI Codex 계정 풀이 정상이어도 native auth 만료 시 로그인창이 열림


jhste102lab:
## Client or integration
Codex CLI

## Area
Authentication and account pool

## Summary
OpenCodex의 OpenAI/Codex 계정 풀이 정상이고 OAuth 계정도 유효한 상태인데, 로컬 Codex CLI의 native `auth.json` 토큰이 만료되면 `codex` 실행 시 OpenCodex proxy로 요청이 전달되기 전에 ChatGPT/Codex 로그인창이 열립니다.

현재 `codexAccountMode = pool`을 사용하고 있으므로, OpenCodex가 보유한 계정 풀을 이용할 수 있다면 별도의 만료된 native Codex 인증 때문에 실행 자체가 중단되지 않아야 한다고 생각합니다. native 인증이 Codex CLI 시작 단계에서 반드시 필요하다면, 최소한 OpenCodex가 원인과 복구 방법을 명확하게 안내하거나 자동 복구해야 합니다.

추가로 Codex CLI npm 업데이트 후 기존 Codex 실행 파일이 교체되면서 OpenCodex autostart shim이 사라졌습니다. `ocx status`는 shim 상태를 감지했지만, 사용자가 직접 `ocx codex-shim install`을 다시 실행하기 전까지 `codex`가 proxy를 자동 시작하지 않았습니다.

## Reproduction
1. OpenCodex에서 OpenAI/Codex 계정 풀을 설정하고 OAuth 계정을 2개 이상 등록합니다.
2. OpenAI provider를 다음과 같이 설정합니다.

   ```json
   {
     "baseUrl": "https://chatgpt.com/backend-api/codex",
     "authMode": "forward",
     "codexAccountMode": "pool"
   }
   ```

3. `ocx start` 또는 `ocx ensure`를 실행합니다.
4. `ocx account refresh openai --json`에서 계정들이 `needsReauth: false`로 표시되는 것을 확인합니다.
5. 로컬 Codex CLI가 사용하는 `CODEX_HOME/auth.json`의 access token을 만료시키거나, 만료된 native auth 상태에서 `codex`를 실행합니다.
6. OpenCodex proxy가 정상 실행 중이고 계정 풀에도 유효한 OAuth가 있어도, proxy 요청 전에 Codex 로그인창이 열립니다.
[...]

PR #356, opened 2026-07-23
fix(codex): auto-restore the codex shim after external npm updates (#320)

lidge-jun:
## Summary

Addresses the actionable half of #320: when `@openai/codex` is updated via external npm, the replaced launcher silently loses the OpenCodex shim until the user manually runs `ocx codex-shim install`. The CLI now detects the replaced launcher on ordinary `ocx` startup and restores the shim automatically.

(Part 1 of #320 — the native login gate firing before proxy traffic — remains an upstream Codex CLI limitation and is intentionally out of scope.)

## Design

- **Hook point:** ordinary CLI startup after help/version early-exits. The healthy path is a bounded read-only probe — measured median 0.026 ms / p95 0.040 ms over 1,000 samples, zero writes, config not even loaded.
- **Restore = the existing install path.** Auto-restore reuses the same shim install transaction as `ocx codex-shim install`; no duplicated logic.
- **All-or-nothing multi-wrapper repair:** all wrapper paths are fingerprint-checked before ANY mutation; a mid-sequence failure rolls back already-renamed siblings in reverse order (activation-tested with a second-sibling drift scenario).
- **npm-in-progress safety:** a freshly-written launcher defers restore until the binary is stable, so we never fight a running `npm install`.
- **Fail-safe:** any restore failure degrades to a one-line warning; the user's command always proceeds. Missing backup / corrupt state / platform mismatch never triggers a fresh install.
- **Opt-out:** config flag + env var (documented in the configuration reference, 5 locales synced).

## Tests

+382 test lines across `tests/codex-shim.test.ts`, `tests/codex-shim-autorestore.test.ts`, `tests/config.test.ts` — named activations for: replaced→restore+warn, intact→zero-overhead, failure→warn-only, opt-out, sibling-drift rollback, npm-in-progress defer, and never-fresh-install guards.

- typecheck exit 0; focused 88 pass / 0 fail; full suite 3876 pass / 0 fail (314 files); privacy:scan pass.
- Local verification is macOS; Windows wrapper behavior relies on cross-platform CI.

## Security review

Credential-path-adjacent (shim wraps the codex launcher): **explicit maintainer security review requested before merge** per MAINTAINERS.md. No auth flow changes; writes are confined to shim wrapper paths and their backups.

Design/audit trail: `devlog/_plan/260724_bugfix_train/030_shim_autorestore.md`.

Refs #320.

## Summary by CodeRabbit

* **New Features**
  * Automatically restores Codex shims after stable external Codex updates on the next regular command.
  * Adds `codexShimAutoRestore` (default on) and `OPENCODEX_CODEX_SHIM_AUTO_RESTORE=0` to disable auto-restore; skips during uninstall/remove flows.
* **Bug Fixes**
  * Repair now warns only on failure and never blocks the requested command.
  * Help/usage invocations no longer modify launcher files.
* **Documentation**
  * Updated CLI/config/runtime and multilingual docs, including manual fallback via `ocx codex-shim install`.
* **Tests**
  * Expanded unit, integration, concurrency/lock-safety, and help-exit coverage.


Wibias:
## Dual review (security + bugs/edge cases)

Bot review threads are **not** cleared yet (**6 unresolved**: 5 Codex P2 + 1 CodeRabbit docs typo). Below is an independent second look. This PR is **credential-path-adjacent** (mutates Codex launchers that load `OPENCODEX_API_AUTH_TOKEN`); CI is green, but maintainer security review remains warranted before merge.

### Security categories checked
No evidence of malware, backdoor, hidden remote C2, local admission bypass, classic XSS, or credential logging/serialization. Writes stay on shim wrapper / backup paths plus `OPENCODEX_HOME/codex-shim.json`. Token injection in rewritten wrappers reuses the existing shim templates (pre-existing pattern), not a new auth flow.

Fail-safe shape looks intentional and largely sound: `allowFreshInstall: false` on the auto path, opt-out via config/`OPENCODEX_CODEX_SHIM_AUTO_RESTORE=0`, restore failures degrade to a warning, and the healthy path is a bounded prefix probe that does not load config.

### Blockers

1. **Concurrent auto-restore can nest shims / lose the real launcher (High)** — `src/codex/shim.ts` (`applyGuardedRefreshTransaction`)  
   Two `ocx` processes can both pass the pre-rename fingerprint check, then both run the rename→`writeShim` sequence. After process A stages the npm launcher into the backup and writes a shim, process B can rename that new shim onto the backup and rewrite the wrapper again. The backup then becomes a shim, so later launches recurse (`shim → ensure → shim…`) and the updated Codex binary is lost. Intra-process re-probe / rollback tests do not cover this cross-process race. Needs an interprocess lock around the guarded transaction, or conditional no-clobber ops that abort when source/destination identity changed.

2. **Stability check is mtime-age only, not “unchanged for N ms” (High)** — `fingerprintIsOldEnough`  
   An old `mtime`/`ctime` does not prove the bytes stayed stable for `CODEX_SHIM_REPLACEMENT_STABLE_MS` (100 ms). npm/tar extracts often preserve package mtimes, so a just-written launcher can look “old enough” immediately while siblings are still being updated. Auto-restore can then snapshot a partial/in-flight launcher into the owned backup. Prefer two matching probes separated by the interval (or persist the first observation and defer).

### Should-fix before merge

3. **Mixed Windows launcher sets still repair (Medium)** — `autoRestoreCodexShim`  
   Healthy shim siblings `continue` while replaced siblings are collected. On Windows, a sequential npm update of `codex.cmd` / `.ps1` / extensionless `codex` can therefore be mutated mid-update, contrary to the “all-or-nothing / mixed-sibling safety” story. Defer while the tracked set is mixed (some external replacements, some still shims).

4. **Top-level help / bare `ocx` can mutate launchers (Medium)** — `src/cli/index.ts`  
   `maybeAutoRestoreCodexShim` runs before the `--help`/`-h`/empty-argv usage path. Help/version early-exits above it do not cover these cases, so printing usage can rename wrappers and rewrite shim state. Skip auto-restore for top-level help and no-command invocations.

5. **Startup state read is unbounded (Medium)** — `readState` via every ordinary command  
   `readFileSync` + `JSON.parse` on `codex-shim.json` with no size/type cap. A huge or non-regular state file turns every `ocx` invocation into an unbounded memory/blocking hit, contradicting the “bounded probe” claim. Stat as a small regular file and cap bytes before parse.

6. **State paths are trusted absolute targets (Medium, security)** — `codex-shim.json`  
   Auto-restore does not re-bind `wrapperPath` / `backupPath` / `realPath` to current PATH discovery. Anyone who can write `OPENCODEX_HOME` can point repair at arbitrary paths (same privilege tier as config, but now exercised on every command). Consider path allowlisting / re-resolve against discoverable Codex launchers, or refuse restore when state paths diverge from discovery.

### Non-blocking / already covered well
- Never-fresh-install guards, opt-out laziness, intra-process fingerprint mismatch defer, and multi-wrapper rollback tests look solid.
- Russian docs typo (`лончер` → `лаунчер`) matches the open CodeRabbit thread — fix when touching docs.

### Aligns with open bot threads
Findings 1–5 match the still-open Codex P2 threads (concurrency lock, stability observation, mixed Windows set, help side effects, bounded state read). Please resolve those after fixing, or reply why a finding does not apply.

### Suggested fix order
1. Serialize the guarded refresh (lock or atomic no-replace) and add a cross-process race test if feasible.  
2. Require two matching fingerprints separated by the stability interval (archive-mtime safe).  
3. Defer on mixed Windows sibling sets.  
4. Skip auto-restore for top-level help / empty argv; bound `readState`.  
5. Optionally harden state-path trust / re-discovery before merge given the security adjacency.

lidge-jun:
Maintainer triage review (2026-07-24, post-#359 re-review). Direction is valid and this does NOT fight #359's persisted runtime selection (`codex-runtime.json` is untouched; configured runtime still wins over the shim candidate). However, merging as-is is blocked by real defects found in the restore transaction:

**High — concurrent restore can lose the real launcher / create shim recursion.** Two processes both pass the fingerprint check; process A writes the shim while process B renames that shim to backup (src/codex/shim.ts:691-713). There is no interprocess lock or conditional rename between verify and rename. Result: shim -> backup-shim recursion and the updated launcher is lost.

**High — the "stable for 100ms" guarantee is only an mtime/ctime age check** (src/codex/shim.ts:217-220, :858-865). An updater that stalls >100ms after a partial write gets its partial file promoted over the known-good backup.

**Medium — Windows mixed sibling sets are restored piecemeal** (.cmd replaced while .ps1/extensionless are still shims), contradicting the doc's own "mixed siblings defer" note (src/codex/shim.ts:638-646, :850-865).

**Medium — bare `ocx` / `ocx help` / `--help` can mutate files** before early exit (src/cli/index.ts:48-63, src/cli/codex-shim-autorestore.ts:18-32).

**Medium — every normal invocation synchronously reads+parses all of codex-shim.json with no size/type bound** (src/codex/shim.ts:493-516, :842-843); not a "bounded startup probe" for large or non-regular files.

Also note: the green CI on ec1418b6 predates the #359 merge, so exact-SHA cross-platform CI is needed after rework.

Suggested path: add an interprocess lock (or compare-and-rename) around the verify->rename transaction, observe real stability (two consecutive identical fingerprints across an interval, not just age), defer mixed sibling sets, move auto-restore behind the help early-exit, and bound the shim-state read. Happy to re-review promptly after a rebuild on current dev.

PR #366, opened 2026-07-24
fix(cursor): stable conversation continuity for store:false turns

Wibias:
## Summary

- Stable Cursor conversation identity across `store:false` full-history / tool continuations for **native Composer and external models**, so Resume / context carry-forward can hit again (covers the gap closed #367 targeted for Composer).
- Resolution order: force-fresh → **isolate helper/shadow/compaction** → remembered `_cursorConversationId` → thread override → `x-codex-parent-thread-id` (scoped) → random. Never uses Responses `previous_response_id` or shared `prompt_cache_key` as conversation ownership.
- External-only hardening: bounded root replay (count + UTF-8 byte budget + turn-atomic truncation + always-keep/truncate active tool result), same-id tool resumes, no-retry after local side effects / already-emitted output — reduces Connect `invalid_argument` after hydration (`usedTokens: 0`).
- Thread→conversation mapping is namespaced by Codex pool account when present, otherwise by a hash of the authenticated Cursor credential (avoids unscoped `local` collisions on shared gateways).
- Cache detail stays honestly unreported: Cursor Connect does not expose cache read/write token counts; continuity is what makes upstream cache/context reuse possible.
- Thanks [@kimrinking-cell](https://github.com/kimrinking-cell) / closed #367 for the continuity diagnosis; this PR supersedes that approach with thread-first identity and covers external models too.

## Test plan

- [x] `bun run typecheck`
- [x] `bun test tests/cursor-request-builder.test.ts tests/cursor-blob.test.ts tests/cursor-adapter.test.ts` (+ continuity cases in `server-combo-failover-e2e`)
- [x] Isolation wins over remembered parent id (compaction/helper + `previous_response_id` path)
- [x] Multi-byte (CJK) tool-result fixture stays within `CURSOR_EXTERNAL_ROOT_BYTE_LIMIT`
- [x] Rebased onto current `dev` (post-#360 / terminal-truth)
- [x] Live external same-id tool-result verification against Cursor Connect — operator sessions on this tip (e.g. `gpt-5.6-luna` / other externals) stay up through multi-turn tool resumes; prior tip repeatedly disconnected / `invalid_argument` after hydration

## Maintainer triage (2026-07-24)

Addressed on tip `bb18b7b5` after rebase onto current `dev`:

| Finding | Status |
|---|---|
| High — isolation bypassed when remembered id present | Fixed — isolate checked before remembered id + regression test |
| Medium — UTF-16 slice vs byte budget | Fixed — UTF-8 byte truncate + CJK fixture |
| Privacy — `local` scope on Cursor routes | Fixed — credential-derived `_cursorIdentityScope` in Cursor adapter |
| Rebuild on current `dev` | Done — clean rebase, no conflicts with #360 / terminal-truth |
| Live same-id verification | Done — live external sessions on this tip no longer hit the prior disconnect / `invalid_argument` loop |

## Summary by CodeRabbit

* **New Features**
  * Improved Cursor conversation continuity across requests, including stable reuse for store-free workflows via client thread header tracking.
  * Added bounded external-model replay handling, including external root prompt limits, byte-budget compaction, and tool-result truncation.
  * Provider diagnostics now include token usage details and improved checkpoint descriptors.

* **Bug Fixes**
  * Prevented unsafe retries after local side effects; adjusted invalid-argument retry behavior to match replay-safety rules.
  * Preserved conversation IDs for valid continuations while isolating helper and compaction operations to avoid cross-thread mixing.

### DIFF
--- a/src/adapters/cursor.ts
+++ b/src/adapters/cursor.ts
@@ -11,6 +11,7 @@ import {
   CursorMissingCredentialError,
   rekeyCursorContextUsage,
 } from "./cursor/live-transport";
+import { rememberCursorThreadConversation } from "./cursor/thread-continuity";
 import { runCursorTurnWithRetry } from "./cursor/transport-retry";
 import {
   createDisabledCursorTransport,
@@ -123,8 +124,8 @@ export function createCursorAdapter(provider: OcxProviderConfig, deps: CursorAda
           await runOnce(request);
         } catch (err) {
           // One-shot fallback for external-model Connect invalid_argument before any
-          // non-heartbeat output. Covers plain-user continuations as well as tool-result
-          // resumes (PR #318); replaying after text/tool events would duplicate output.
+          // non-heartbeat output. Retries apply only to safe plain-user turns; tool-result
+          // resumes, local exec/MCP side effects, and already-emitted output fail closed.
           if (
             !isCursorInvalidArgumentError(err)
             || !isCursorExternalWireModel(request.modelId)
@@ -140,6 +141,15 @@ export function createCursorAdapter(provider: OcxProviderConfig, deps: CursorAda
           request = createCursorRequest(_parsed, { forceFreshConversation: true });
           rekeyContextUsage(failedConversationId, request.conversationId);
           _parsed._cursorConversationId = request.conversationId;
+          // Persist recovery for store:false clients that only send a parent thread id, so the
+          // next turn does not recompute the stale deterministic thread hash.
+          if (_parsed._clientThreadId) {
+            rememberCursorThreadConversation(
+              _parsed._clientThreadId,
+              request.conversationId,
+              _parsed._cursorIdentityScope,
+            );
+          }
           await runOnce(request);
         }
       } catch (err) {
--- a/src/adapters/cursor.ts
+++ b/src/adapters/cursor.ts
@@ -1,3 +1,4 @@
+import { createHash } from "node:crypto";
 import type { AdapterEvent, OcxProviderConfig } from "../types";
 import type { ProviderAdapter } from "./base";
 import { cursorExecDeniedMessage, cursorRequestDeclaresFullAccess } from "./cursor/exec-policy";
@@ -10,6 +11,7 @@ import {
   createLiveCursorTransport,
   CursorMissingCredentialError,
   rekeyCursorContextUsage,
+  resolveCursorToken,
 } from "./cursor/live-transport";
 import { rememberCursorThreadConversation } from "./cursor/thread-continuity";
 import { runCursorTurnWithRetry } from "./cursor/transport-retry";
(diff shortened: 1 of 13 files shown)


Wibias:
Addressed the latest unresolved CodeRabbit threads in `d0809120`:

- `truncateToolResultBlob` now truncates by UTF-8 bytes (with JSON-envelope headroom) and keeps the stored blob within `CURSOR_EXTERNAL_ROOT_BYTE_LIMIT` (CJK regression test added)
- `_cursorIsolateConversation` is checked before a remembered `_cursorConversationId` so compaction/helper turns cannot append into the parent conversation
- Cursor routes without a Codex pool account now get a credential-derived `_cursorIdentityScope` inside the Cursor adapter (Codex pool scope still preferred when present)

lidge-jun:
Maintainer triage review (2026-07-24). Cannot merge as-is — rebuild on current dev needed:

**High — helper/compaction isolation is bypassed.** core.ts:623-627 restores the parent `_cursorConversationId` and isolation is set at core.ts:653-656, but request-builder.ts:187-195 returns the remembered id BEFORE isolation is consulted — a helper/compaction request carrying previous_response_id attaches to the parent's Cursor conversation. Tests only cover the no-remembered-id case (tests/cursor-request-builder.test.ts:128-138).

**Medium — the UTF-8 byte cap is not actually enforced.** protobuf-request.ts:106-114 slices by UTF-16 code units against a byte budget: a 512 KiB budget holds 524,263 ASCII bytes but 1,572,583 bytes of Korean text and 1,048,423 bytes of emoji. Tests use ASCII only (tests/cursor-blob.test.ts:149-183).

**Unproven core strategy.** The PR flips external tool-result continuation from fresh-id/rekey to same-id resume (request-builder.ts:187-220, cursor.ts:126-153); the PR body itself lists native Composer / external-model live verification as incomplete. CI on the head is green, but there are no check runs on the synthetic merge with current dev (branch is 27 commits behind).

**Conditional privacy risk (note).** core.ts:823-826 tries to scope by Codex account id, but the Cursor provider has no codexAccountMode (src/providers/registry.ts:347-355), so it falls back to the "local" scope (thread-continuity.ts:30-33) — same client thread id can collide across users on a shared gateway.

Confirmed non-conflicts: #360's oversized call-id aliasing and this week's terminal-truth adapter changes do not intersect these files.

Suggested path: rebase onto current dev, fix the isolation ordering and the byte-budget slicing (slice by UTF-8 bytes, not code units), add multi-byte fixtures, and run the live external same-id verification listed in the PR body. Happy to re-review after a rebuild.

Wibias:
## Maintainer triage follow-up (2026-07-24)

Rebuilt on current `dev` (force-pushed tip `bb18b7b5`). Clean rebase; no file conflicts with #360 or this week's terminal-truth adapter work.

### High — isolation ordering
Fixed earlier in the CodeRabbit round and retained through rebase: `resolveCursorConversationId` now checks `_cursorIsolateConversation` **before** remembered `_cursorConversationId`. Regression: `isolation wins over a remembered parent conversation id` in `tests/cursor-request-builder.test.ts`.

### Medium — UTF-8 byte budget
`truncateToolResultBlob` truncates by UTF-8 bytes (with JSON-envelope headroom) and loops until `byteLength <= maxBytes`. Regression: `truncates multi-byte tool results by UTF-8 byte budget` (CJK `한` filler) asserts `rootBytes <= CURSOR_EXTERNAL_ROOT_BYTE_LIMIT`.

### Privacy — unscoped `local` fallback
Cursor routes without `codexAccountMode` now get a SHA-256 fingerprint of the authenticated Cursor credential as `_cursorIdentityScope` inside the adapter (Codex pool account still preferred when present). Regression: `derives distinct thread conversation ids per Cursor credential`.

### Unproven live strategy
Agree this remains the open gate. Mocked continuity e2e + unit coverage are green after the rebuild; live external same-id Connect verification still needs an operator Cursor token and is unchecked in the PR body.

Happy to take a re-review on the rebuilt tip.

Wibias:
## Re: unproven live strategy

That gate is closed from the operator side.

Before this branch, external Cursor models (Connect) repeatedly died mid-session — disconnects / `invalid_argument` after blob hydration (often with `usedTokens: 0`), especially on tool-result / `store:false` continuations.

On the rebuilt tip (`bb18b7b5`, rebased on current `dev`), the same setup is being used live: `gpt-5.6-luna` and other externals run multi-turn tool resumes continuously with **no** recurrence of that failure mode.

So the same-id resume + continuity strategy is not only covered by mocked e2e — it is already exercising real Cursor Connect in production-like operator use. Happy to re-review with that as the live smoke.

PR #367, opened 2026-07-24
fix(cursor): pin native composer conversation id to prompt_cache_key

kimrinking-cell:
## Summary

- Native Cursor models (e.g. `composer-2.5`) now derive a stable conversation id from Codex `prompt_cache_key` when `_cursorConversationId` is not restored from `previous_response_id`.
- Stops minting a fresh `cursor_*` id on every `store:false` full-history turn, so Cursor prompt-cache / context-usage carry-forward can hit again.
- External-model tool-result force-fresh behavior is unchanged; the adapter no longer pre-mints a random id that would hide the cache-key pin.

## Verification

- `bun run typecheck`
- `bun test tests/cursor-request-builder.test.ts tests/cursor-adapter.test.ts tests/server-combo-failover-e2e.test.ts tests/responses-state.test.ts` (91 pass)
- `bun run privacy:scan`
- Live proxy restarted with the patched files; unit probe confirmed same `prompt_cache_key` → same conversation id

## Checklist

- [x] Docs or release notes were updated when needed. *(internal continuity only — no user-doc change)*

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Cursor conversation continuity across follow-up requests.
  * Preserved conversation context for native Composer models using the same prompt cache key.
  * Ensured tool-result continuations start fresh conversations when required.
  * Improved continuity for streamed and non-stored conversation flows.

* **Tests**
  * Added coverage for stable conversation reuse and fresh conversations after external tool results.
  * Added end-to-end checks for conversation continuity across multiple request patterns.

### DIFF
--- a/src/adapters/cursor.ts
+++ b/src/adapters/cursor.ts
@@ -5,7 +5,7 @@ import { isCursorBenignCancelError, isCursorInvalidArgumentError, safeCursorErro
 import { isCursorExternalWireModel } from "./cursor/discovery";
 import { createCursorKvStore, type CursorKvStore } from "./cursor/kv-store";
 import { mapCursorServerMessage } from "./cursor/message-mapper";
-import { createCursorRequest, generatedCursorConversationId } from "./cursor/request-builder";
+import { createCursorRequest } from "./cursor/request-builder";
 import {
   createLiveCursorTransport,
   CursorMissingCredentialError,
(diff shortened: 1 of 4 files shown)


Wibias:
Closing this in favor of #366.

Thanks for surfacing this — the continuity diagnosis is right. We're already chasing the same class of failure on external Cursor models (`gpt-5.6-sol` / similar) in #366: under `store:false` full-history turns, minting a fresh `cursor_*` id each time breaks Resume / context carry-forward and shows up as intermittent Connect `invalid_argument` after hydration (`usedTokens: 0`). So the problem isn't limited to native Composer.

We'll cover that path in #366 (stable provider conversation identity for external models as well), and we'll thank / mention you in the changelog for pointing at the continuity root cause.

PR #370, opened 2026-07-24
fix(codex): reset main runtime state after account switch

duansy123:
## Summary

- detect when the physical ChatGPT identity in `~/.codex/auth.json` changes behind the stable `__main__` alias
- purge alias-scoped cooldown, quota, reauthentication, thread-affinity, plan, and WebSocket state before the next pool routing decision
- add a regression test covering a cooled/reauth-marked main account followed by an auth identity switch

## Root cause

The main Codex login always uses the in-memory key `__main__`. Replacing `auth.json` with another account therefore left the new physical account inheriting the previous account's runtime health state. In particular, a prior 429 cooldown caused the proxy to reject the new account locally with `Selected Codex account is cooling down` without contacting the upstream service.

## Security review

This touches authentication lifecycle behavior and should receive explicit security review. The change keeps the observed account ID in process memory only, does not persist or log it, does not expose token contents, and invalidates existing `__main__` WebSockets when the physical identity changes so they cannot continue with stale auth context. Direct mode is unchanged.

## Verification

- `bun run typecheck`
- `bun test --isolate tests/codex-main-rotation.test.ts tests/codex-auth-context.test.ts` (31 passed)
- `bun test --isolate tests/ci-workflows.test.ts` (10 passed)
- `bun run privacy:scan`
- full suite attempted: 3912 tests passed; `tests/openai-provider-option-e2e.test.ts` hit its existing 30s timeout and the same timeout reproduces on an untouched `origin/dev` worktree
- live local proxy request after applying the patch: HTTP 200 through `gpt-5.6-sol`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Codex behavior when the main account identity changes by reconciling account runtime state earlier in the auth flow.
  * Added safer handling for cached main account info, including invalidation and re-fetching with the replacement credentials.
  * Reduced stale quota/plan/cooldown/reauth carryover after identity rotation, including during transient missing-identity scenarios.
* **Tests**
  * Expanded coverage for main-account rotation, cache invalidation, and identity-mismatch retry behavior.
  * Added assertions for correct state cleanup and consistent targeting across in-flight requests during identity changes.

### DIFF
--- a/src/codex/account-lifecycle.ts
+++ b/src/codex/account-lifecycle.ts
@@ -5,15 +5,17 @@ import { MAIN_CODEX_ACCOUNT_ID, setMainAccountPlan } from "./main-account";
 import { clearAccountQuota } from "./quota";
 import { clearCodexUpstreamHealthForAccount, clearThreadAccountMapForAccount } from "./routing";
 import { invalidateCodexWebSocketsForAccount } from "./websocket-registry";
+import { clearMainAccountInfoCache } from "./main-account-cache";
 import type { OcxConfig } from "../types";
 
-let observedMainChatgptAccountId: string | null | undefined;
+let observedMainChatgptAccountId: string | undefined;
 
 export function purgeCodexAccountRuntimeState(accountId: string): void {
   clearAccountNeedsReauth(accountId);
   clearAccountQuota(accountId);
   clearThreadAccountMapForAccount(accountId);
   clearCodexUpstreamHealthForAccount(accountId);
+  if (accountId === MAIN_CODEX_ACCOUNT_ID) clearMainAccountInfoCache();
 }
 
 /**
@@ -24,6 +26,9 @@ export function purgeCodexAccountRuntimeState(accountId: string): void {
  */
 export function reconcileMainCodexAccountRuntimeState(): boolean {
   const currentAccountId = getMainChatgptAccountId();
+  // A missing/malformed auth.json is an unknown identity, not a confirmed account switch. Keep the
+  // prior observation and its safety state until a real account id can be read again.
+  if (currentAccountId === null) return false;
   const previousAccountId = observedMainChatgptAccountId;
   observedMainChatgptAccountId = currentAccountId;
   if (previousAccountId === undefined || previousAccountId === currentAccountId) return false;
--- a/src/codex/account-lifecycle.ts
+++ b/src/codex/account-lifecycle.ts
@@ -1,15 +1,41 @@
 import { removeCodexAccountCredential } from "./account-store";
 import { clearAccountNeedsReauth } from "./account-runtime-state";
+import { getMainChatgptAccountId } from "./auth-collision";
+import { MAIN_CODEX_ACCOUNT_ID, setMainAccountPlan } from "./main-account";
 import { clearAccountQuota } from "./quota";
 import { clearCodexUpstreamHealthForAccount, clearThreadAccountMapForAccount } from "./routing";
 import { invalidateCodexWebSocketsForAccount } from "./websocket-registry";
 import type { OcxConfig } from "../types";
+
+let observedMainChatgptAccountId: string | null | undefined;
 
 export function purgeCodexAccountRuntimeState(accountId: string): void {
   clearAccountNeedsReauth(accountId);
   clearAccountQuota(accountId);
   clearThreadAccountMapForAccount(accountId);
   clearCodexUpstreamHealthForAccount(accountId);
+}
+
+/**
+ * The main Codex login is stored under the stable `__main__` alias, while
+ * `~/.codex/auth.json` can be replaced with credentials for another physical
+ * ChatGPT account. Drop alias-keyed runtime state when that identity changes so
+ * cooldown, quota, reauth, and thread affinity do not leak across accounts.
+ */
+export function reconcileMainCodexAccountRuntimeState(): boolean {
+  const currentAccountId = getMainChatgptAccountId();
+  const previousAccountId = observedMainChatgptAccountId;
+  observedMainChatgptAccountId = currentAccountId;
+  if (previousAccountId === undefined || previousAccountId === currentAccountId) return false;
+
+  purgeCodexAccountRuntimeState(MAIN_CODEX_ACCOUNT_ID);
+  setMainAccountPlan(null);
+  invalidateCodexWebSocketsForAccount(MAIN_CODEX_ACCOUNT_ID);
+  return true;
+}
+
+export function resetMainCodexAccountIdentityTrackingForTests(): void {
+  observedMainChatgptAccountId = undefined;
 }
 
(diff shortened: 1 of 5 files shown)


lidge-jun:
Maintainer triage review (2026-07-24). The target bug is real (stale main-account runtime state after a switch) and the codex-runtime.json contract is untouched, but the detection/invalidation is incomplete — not mergeable as-is:

**High — startup-prime path missed.** The first pool request can prime quota/plan/reauth state before any reconciliation (src/codex/auth-api.ts:363,379), but the PR skips purging when the previous observed ID is undefined (src/codex/account-lifecycle.ts:25-29). Account A primes, then the user switches to B: the original staleness bug survives. The added test seeds the observed ID first, so it misses this path (tests/codex-main-rotation.test.ts:155-164).

**Medium — null reads treated as switches.** A temporarily missing/malformed auth.json yields a null ID (src/codex/auth-collision.ts:8,25); the PR processes A -> null -> A as two switches and clears cooldown/reauth, allowing a same-account retry before the limit (account-lifecycle.ts:26-33). Null reads should fail closed (treat as "unknown", keep prior state).

**High — mainAccountCache not invalidated.** The 5-minute TTL cache is consulted before reading the new token (src/codex/auth-api.ts:182,248), so after switching to B the API/dashboard can still show A's email/plan/quota (src/codex/auth-api.ts:404). The purge must include this cache.

Suggested path: rebuild on current dev with null-read fail-closed, pre-prime identity seeding, and mainAccountCache invalidation in the same switch path. Happy to re-review promptly.

Ingwannu (changes requested):
The stale __main__ state bug is real, but the auth lifecycle fix remains incomplete. The first reconciliation can miss state populated by startup/dashboard priming, transient null auth reads are treated as identity changes and can clear a valid cooldown, and mainAccountCache is not invalidated, so the old account details can remain visible for its TTL. Please seed identity before state population, fail closed on unknown/null reads, invalidate the account-info cache in the same transition, avoid repeated synchronous auth.json reads, and re-request review.

duansy123:
已按 review 意见补齐并推送 `ed7147fa`：

- 启动 quota prime 前先登记物理主账号身份，避免首次 A→B 切换漏清理。
- auth.json 缺失/损坏时 fail-closed，不把瞬时 null 误判为账号切换。
- 主账号身份变化时同步清除 mainAccountInfo 缓存。
- 新增 3 条回归测试；定向测试 25 pass，typecheck 通过，diff --check 通过。

请重新 review。

duansy123:
Addressed the in-flight main-account usage race in `94e40ad6`.

- Capture the physical account identity used by each WHAM request.
- Revalidate it after response/body parsing and before mutating cache, reauth, plan, or quota state.
- Discard a stale response and retry exactly once with the replacement identity.
- Added a delayed-fetch regression covering A request → switch/reconcile to B → A response resolves; only B data is published.

Validation on the PR head:
- Focused Codex/provider tests: 118 passed
- Full suite: 3938 passed, 0 failed
- Typecheck: passed
- Privacy scan: passed
- `git diff --check`: passed

@coderabbitai review
@lidge-jun Please re-review when convenient.

duansy123:
Addressed the transient-null follow-up in `74322037`.

`getMainChatgptAccountId() === null` is now treated as an unknown identity: it neither reconciles nor retries. A retry remains limited to one attempt and only occurs for a confirmed non-null identity change. Added a delayed-response regression for a temporarily missing `auth.json`.

Validation:
- Focused tests: 119 passed
- Full suite on local/stable: 4056 passed, 0 failed
- Typecheck, privacy scan, and diff check: passed

@coderabbitai review
@lidge-jun Please re-review when convenient.

lidge-jun:
### 🔒 Maintainer integration in progress — please hold off on merging

@lidge-jun (maintainer) is actively integrating this PR into `dev` as part of a coordinated
integration pass. **Please do not merge, rebase, force-push, or close this PR** until this
marker is removed.

**What is happening**

This PR is being integrated on the maintainer branch `codex/260725-pr-rework`, which is
based on `dev`. The pass applies the contributed change, adds any missing regression
coverage, and repairs review-identified defects before a single verified merge into `dev`.
Every integrated change goes through `bun run typecheck`, the full `bun run test` suite,
`bun run privacy:scan`, and `bun run lint:gui`, followed by exact-SHA hosted
`Cross-platform CI` and `Service lifecycle` runs.

**What this means for you**

- Your authorship and commits are preserved. Nothing is being rewritten under your name.
- No action is needed from you right now. If a defect repair changes your intended
  behavior, it will be described explicitly in a follow-up comment before the merge.
- If you have work in flight on this branch, please comment here instead of pushing, so we
  do not race each other.

This PR will be closed with a merge receipt (integration commit SHA, verification output,
and hosted CI links) once the pass lands on `dev`. Thanks for the contribution.

<sub>Integration tracker: `devlog/_plan/260725_pr_issue_rework` · marker posted by the maintainer integration pass</sub>

lidge-jun:
### ✅ Integrated into `dev`

Thank you for this fix, and for working through the earlier review rounds. It has been integrated into `dev`.

**Integration commit:** `03e3f1b4` — `fix(codex): reset main runtime state after account switch`
**Merged into `dev` at:** `ebc62d1f`
Your authorship is preserved via `Co-authored-by`.

**Review status**

Five of the six earlier review points were already resolved in your follow-up commits: startup identity seeding before quota priming, not treating a transient null as an account switch, main cache invalidation, the in-flight identity fence with a single retry, and the null-identity guard during retry.

The sixth was repaired in the same commit. `readCodexTokens()` collapsed "file absent", "malformed JSON", and "read error" into one `null`, so `fetchMainAccountInfo()` could not tell a real sign-out from a half-written `auth.json` — and it cleared the cache and marked the account for reauth either way. A `readCodexTokensResult()` now classifies the outcome (`missing` / `invalid` / `unreadable`); `readCodexTokens()` stays as a compatibility wrapper so the other eight call sites keep their fail-closed contract. A local read failure no longer touches the cache or the reauth flag; routing stays fail-closed because `getMainAccountToken()` re-reads the file itself.

An `existsSync` pre-check was considered and rejected: the gap between check and read is exactly the window a non-atomic rewrite lands in.

**Verification**

- `bun test tests/codex-main-rotation.test.ts`: 22 pass / 0 fail (20 before)
- Activation proof: reverting only the transient-preservation branch fails both new regressions
- Both causes are covered — file removal and malformed JSON — each asserting cache/quota preservation, no upstream call, fail-closed usability while broken, and clean recovery afterwards
- Security review: no credential/token/account-identifier logging, purge stays scoped to `MAIN_CODEX_ACCOUNT_ID`, no new credential path; `bun run privacy:scan` clean
- Full suite at merge: 4222 pass / 0 fail

lidge-jun:
Closing: integrated into `dev` at `ebc62d1f` with authorship preserved. See the receipt comment above for the integration commit, the changes made on top, and the verification evidence.

PR #378, opened 2026-07-24
fix(proxy): raise default upstream stall timeout to 600s

Wibias:
## Summary

Operator sessions on Cursor external models (especially `gpt-5.6-sol` / `xhigh`) regularly sit on Connect **heartbeat-only** frames for well over five minutes while the model thinks — no `thinkingDelta` / `tokenDelta` yet, but the stream is still alive.

With the previous default (`stallTimeoutSec = 300`), the Responses bridge treated that silence as a hung upstream and aborted at ~302s:

- `response.incomplete` / `upstream_stall_timeout`
- client-visible **502** with `Upstream stalled: no data for the stall-timeout window`
- `usageStatus: unreported` (turn never finished)

Concrete local examples (same night, before the bump): `ocx-mrycu6ah-2q` and `ocx-mryd0nmg-2r` both failed at **~302147ms / ~302170ms** on `gpt-5.6-sol` `xhigh` via `cursor`, while a neighboring turn that finished under five minutes (`ocx-mrycr9qt-2p`, ~134s) returned **200**. Cursor sometimes also drops with `NGHTTP2_INTERNAL_ERROR` mid-heartbeat, but those are separate; the ~302s 502s are specifically our stall watchdog.

This PR raises `DEFAULT_STALL_TIMEOUT_SEC` **300 → 600** so those long heartbeat-only thinks get a 10-minute window before we cut them. Hung streams still die; they just match real Cursor/xhigh think times better. Docs (EN + locales) and web-search stall budget expectations (default + 30s margin → **630**) are updated to match.

Operators who already set `stallTimeoutSec` in config are unchanged; unset installs pick up 600.

## Test plan

- [x] `bun test tests/stall-timeout.test.ts`
- [x] `bun test` web-search stall / bridge-budget cases (expect 630)
- [x] Local: `~/.opencodex/config.json` `stallTimeoutSec: 600` + proxy restart (immediate mitigation while this merges)
- [ ] After merge/install: confirm long `gpt-5.6-sol` `xhigh` thinks past 5 minutes no longer land as ~302s `upstream_stall_timeout` 502s


lidge-jun:
Closing with thanks — the premise no longer holds on current dev. Since the heartbeat work landed (WP4), the bridge resets the stall clock on EVERY adapter event, including keepalive heartbeats (src/bridge.ts:452; pinned by tests/bridge.test.ts:461), and Cursor converts token deltas/checkpoint progress into heartbeat events (src/adapters/cursor/live-transport.ts:860,935). So a "heartbeat-only progress got cut at 300s" scenario points to a pre-WP4 runtime or a heartbeat-delivery gap — not to a too-small budget.

Raising the global default to 600s would keep truly dead upstreams alive for 10 minutes everywhere and push the web-search watchdog to ~630s (structure/04_transports-and-sidecars.md:124) — a global double-mask. Long-running Cursor operators already have the per-config `stallTimeoutSec` override.

If a 300s cut still reproduces on a WP4+ runtime, please reopen with the exact SHA, a restarted proxy, and the last adapter-event type + timestamp — that would indicate a heartbeat delivery bug, which we would fix directly instead.

PR #391, opened 2026-07-24
feat: quota-aware subagent model fallback chain (#374)

### LINKED ISSUE #374: [Feature Request] Subagent model fallback chain with quota-aware auto-switching
## Problem

When spawning subagents with a specific `agent_type` (e.g. `executor`), the model is hardcoded in the agent TOML file (e.g. `model = "gpt-5.6-sol"`). If that model hits its usage limit (e.g. ChatGPT Plus quota exhausted), the subagent fails immediately with no recovery path:

```
Spawned 019f91d5-... (gpt-5.6-sol medium)
  └ errored: "You've hit your usage limit..."
```

Currently there is **no fallback mechanism** — the `model` parameter in `spawn_agent` is a single string, and the agent TOML schema has no `fallback_model` or `model_chain` field. The `autoSwitchThreshold` only rotates between multiple accounts **within the same provider**, not across different models/providers.

## Proposed Solution

A quota-aware model fallback chain for subagent spawning:

### 1. Model fallback chain in config

Allow `subagentModels` (or a new `subagentModelFallback`) to define a **priority-ordered fallback list**:

```json
{
  "subagentModelFallback": [
    "gpt-5.6-sol",
    "alibaba-token-plan/qwen3.8-max-preview",
    "kimi/k3"
  ]
}
```

When the primary model fails with a quota/rate-limit error, automatically retry with the next model in the chain.

### 2. Periodic quota polling

A background task (or pre-flight check before each spawn) that queries usage/quota for each configured provider:

- OpenAI Codex account usage (already available via `getAccountQuota`)
- Third-party provider health (e.g. `/v1/models` reachability, recent error rates from `usage.jsonl`)

Cache the results with a short TTL (e.g. 60s) to avoid excessive polling.

### 3. Priority-based routing

When spawning a subagent:

1. Check the agent TOML's `model` field (current behavior)
2. If that model's provider is known to be quota-exhausted or unhealthy, **skip to the next available model** in the fallback chain
3. Optionally override the agent TOML model entirely if `subagentModelFallback` is configured and the TOML model is unavailable

### 4. Agent TOML extension (optional)

Allow per-role fallback in the TOML itself:

```toml
name = "executor"
model = "gpt-5.6-sol"
model_fallback = ["alibaba-token-plan/qwen3.8-max-preview"]
```

## Use Case

Users running opencodex with a ChatGPT Plus account (limited quota) plus third-party providers (Alibaba MaaS, Codebuddy, Kimi, etc.) as alternatives. When the Plus quota is exhausted, subagents should seamlessly fall back to available third-party models instead of failing.

## Environment

- opencodex v2.7.31
- macOS (aarch64)
- Providers: openai (Plus), alibaba-token-plan, kimi, sports
- `autoSwitchThreshold: 0` (only one OpenAI account, so account rotation is not applicable)

## Alternatives Considered

- Manually editing agent TOML files to change the model — works but fragile, overwritten by `omx setup`
- Always passing `model` explicitly in `spawn_agent` — puts the burden on the caller, no auto-recovery
- Not specifying `agent_type` to inherit leader model — loses role-specific system prompts

Wibias:
## Summary

Implements quota-aware subagent model fallback for `thread_spawn` child turns (issue #374).

- Adds `subagentModelFallback` and `subagentModelFallbackPollMs` config
- Rewrites child spawn requests at the proxy choke point before routing
- Skips quota-exhausted native models and recently failed routed models
- Supports per-agent TOML `model_fallback` merge
- Primes Codex pool quotas before spawn selection
- Records quota-like subagent failures for short TTL health blocking
- Adds `/api/subagent-model-fallback` management route
- Injects fallback-chain guidance into v2 multi-agent guidance

## Verification

- `bun run typecheck`
- `bun test tests/subagent-model-fallback.test.ts tests/multi-agent-compat.test.ts`

Closes #374

## Summary by CodeRabbit

* **New Features**
  * Added quota- and health-aware subagent model fallback that can automatically rewrite spawned targets and inject fallback guidance into multi-agent instructions.
  * Added a management endpoint to view/update the fallback priority chain and polling interval.
  * Added support for the `{{fallback}}` placeholder in multi-agent guidance.
* **Bug Fixes**
  * Improved thread-spawn rate-limit/quota failure detection and attribution, with more accurate rerouting and underlying request model updates.
  * Added dynamic Codex home resolution so environment changes are reflected after import.
* **Tests**
  * Added end-to-end tests for chain building/deduping, fallback selection, guidance rendering, and spawn reroute behavior.

### DIFF
--- a/src/codex/paths.ts
+++ b/src/codex/paths.ts
@@ -29,6 +29,11 @@ export const CODEX_PROFILE_PATH = join(CODEX_HOME, "opencodex.config.toml");
 export const DEFAULT_CATALOG_PATH = join(CODEX_HOME, "opencodex-catalog.json");
 export const CODEX_MODELS_CACHE_PATH = join(CODEX_HOME, "models_cache.json");
 
+/** Runtime CODEX_HOME lookup (honors CODEX_HOME env changes after import). */
+export function getCodexHome(): string {
+  return resolveCodexHome();
+}
+
 export function tomlString(value: string): string {
   return JSON.stringify(value);
 }
--- a/src/codex/routing.ts
+++ b/src/codex/routing.ts
@@ -235,14 +235,8 @@ export function computeQuotaCooldownUntil(meta: CodexUpstreamOutcomeMeta = {}): 
  * Returns the lease id, or null when no probe may go out right now.
  */
 export function tryAcquireCodexQuotaProbeLease(accountId: string, now = Date.now()): string | null {
-  const health = upstreamHealth.get(accountId);
-  if (!health) return null;
-  const cooldownUntil = health.cooldownUntil;
-  if (typeof cooldownUntil !== "number" || !Number.isFinite(cooldownUntil) || cooldownUntil <= now) return null;
-  if (health.cooldownSource === "retry-after") return null;
-  if (health.probeLeaseId !== undefined) return null;
-  const origin = health.lastProbeAt ?? health.cooldownSince ?? cooldownUntil;
-  if (now - origin < CODEX_QUOTA_PROBE_INTERVAL_MS) return null;
+  if (!canAcquireCodexQuotaProbeLease(accountId, now)) return null;
+  const health = upstreamHealth.get(accountId)!;
   const probeLeaseId = randomUUID();
   upstreamHealth.set(accountId, {
     ...health,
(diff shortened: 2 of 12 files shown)


Ingwannu (changes requested):
The requested fallback behavior is valuable, but the current implementation can repeatedly select an exhausted model or select against the wrong account. The unresolved findings are substantive: 429/402 are not classified, routed failures are cached under a different key than the configured slug, selection runs before pool account resolution and uses the main plan for pool accounts, native health is not account-scoped, the normal SSE path does not record failures, and combo/encrypted child paths can be rewritten incorrectly. Please rebuild the selection at the resolved routing/account boundary, validate configuration atomically, make the tests independent, resolve the current dev conflict, and re-request review.

Wibias:
Ready

lidge-jun:

lidge-jun:
Thank you for the substantial quota-aware subagent fallback work. I found three correctness blockers that need to be resolved before merge.

1. `src/codex/subagent-model-fallback.ts:275-281` starts quota refresh work with `void` and returns no promise. The caller at `src/server/responses/core.ts:700-703` uses `await maybePrimeSubagentQuota(...)`, but that await resolves immediately, so model selection can run on stale quota state. Return the actual refresh promise (with errors handled at the owning boundary) and add a deferred-promise regression proving selection waits for the refresh result.

2. `src/server/management/agent-settings-routes.ts:328-331` silently removes invalid entries with `filter()` instead of rejecting the request. That violates the management API's atomic validation contract: a typo can persist a partially truncated fallback chain while returning success. Validate every item first, return 400 with the offending index/value, and mutate config only after the whole payload is valid. Add a regression proving one invalid entry leaves the previous config unchanged.

3. `src/codex/subagent-model-fallback.ts:173-184` does not record ordinary rate-limit wording as a health block. Providers that report rate limiting without the currently recognized quota phrases remain eligible and are selected repeatedly. Normalize the existing rate-limit classifier into this health path and test representative generic 429/rate-limit messages without treating unrelated errors as quota exhaustion.

I appreciate the scope of this feature. The refresh must be truly awaited, settings updates must be atomic, and generic rate limits must participate in the same health block before the fallback chain is reliable.

Please also rebase onto current `dev`, which now includes #439's context-pressure fields in `src/types.ts`; preserve that public type contract.

---

PR #393, opened 2026-07-24
fix(gui): portal select dropdowns to avoid clipping

### LINKED ISSUE #340: [Bug]: Claude settings - Search/Vision Sidecar dropdowns clipped by container boundary, bottom options unselectable
### Client or integration

Codex App (Dashboard - Claude settings page)

### Area

GUI / Dashboard

### Summary

On the Dashboard's **Claude** settings tab, the dropdown popups for the **Search Sidecar** and **Vision Sidecar** configuration buttons open downward but are clipped/hidden by the UI container/window boundary. Some options at the bottom of the dropdown cannot be seen or selected.

### Reproduction

1. Open the OpenCodex dashboard at `http://localhost:10100/#claude`
2. Scroll to the Search Sidecar or Vision Sidecar configuration section
3. Click the backend/model dropdown button
4. The dropdown menu opens downward
5. If the section is near the bottom of the viewport or inside a bounded container, the lower options are clipped by the container edge and become invisible/unselectable

### Expected behaviour

The dropdown should either:
- Open upward when near the bottom of the viewport (auto-flip), or
- Overflow the container boundary (visible above sibling elements), or
- Scroll within the dropdown if options exceed available space

### Actual behaviour

Dropdown opens downward, lower options are hidden behind the container/window boundary with no way to select them.

### Environment

- opencodex: 2.7.36
- macOS, Apple Silicon
- Browser: Safari / Chrome (affects both)

### Screenshots

The issue is visible on the Claude settings page when the sidecar dropdown is positioned near the bottom of the visible area.

### Checks

- [x] I searched existing issues for dropdown / overflow / clipping / sidecar UI bugs.
- [x] This is a UI layout bug, not a routing or provider issue.

Wibias:
Fixes #340

## Summary
- Portal shared GUI `Select` dropdown menus to `document.body` so they can escape clipped parents
- Add viewport-aware positioning with upward flip near the bottom edge
- Add unit tests for dropdown positioning logic

## Testing
- `bun test ./tests/select-position.test.ts`
- `bun run lint:gui`
- `bun run typecheck`

## Summary by CodeRabbit

* **New Features**
  * Added portal rendering for selected dropdowns, allowing menus to appear outside clipped containers.
  * Improved dropdown positioning, including viewport-aware placement, automatic upward flipping, and boundary handling.
  * Added support for correctly layering portal-based dropdowns above other interface elements.

* **Bug Fixes**
  * Prevented dropdown menus from being clipped or incorrectly positioned within settings panels.

* **Tests**
  * Added coverage for portal rendering and responsive dropdown positioning.

### DIFF
--- a/gui/src/pages/ClaudeCode.tsx
+++ b/gui/src/pages/ClaudeCode.tsx
@@ -230,6 +230,7 @@ export default function ClaudeCode({ apiBase }: { apiBase: string }) {
             onChange={v => setState({ ...state, authMode: v as ClaudeCodeState["authMode"] })}
             label={t("claude.authMode")}
             style={{ minWidth: 220 }}
+            portal
           />
         </div>
 
@@ -282,6 +283,7 @@ export default function ClaudeCode({ apiBase }: { apiBase: string }) {
               onChange={v => setState({ ...state, autoCompactWindow: v === "" ? null : Number(v) })}
               label={t("claude.autoCompactWindow")}
               style={{ minWidth: 130 }}
+              portal
             />
           </div>
         )}
@@ -320,6 +322,7 @@ export default function ClaudeCode({ apiBase }: { apiBase: string }) {
                       : { ...override, backend: value === "auto" ? undefined : value as SidecarBackend },
                   })}
                   label={t("dash.sidecarBackend")}
+                  portal
                 />
                 <input
                   className="input mono"
--- a/gui/src/select-position.ts
+++ b/gui/src/select-position.ts
@@ -1,0 +1,97 @@
+import type { CSSProperties } from "react";
+
+export interface SelectMenuTriggerRect {
+  top: number;
+  bottom: number;
+  left: number;
+  right: number;
+  width: number;
+  height: number;
+}
+
+export interface SelectMenuStyleOptions {
+  align?: "left" | "right";
+  placement?: "below" | "right";
+  menuHeight?: number;
+}
+
+const MENU_GAP_PX = 4;
+const FLIP_GAP_PX = 8;
+const VIEWPORT_PAD_PX = 8;
+const MAX_MENU_HEIGHT_PX = 280;
+const MIN_MENU_HEIGHT_PX = 120;
+const BESIDE_MIN_WIDTH_PX = 160;
+
+function viewportHeight() {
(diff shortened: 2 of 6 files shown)


Ingwannu (changes requested):
The clipping bug in #340 is real and using a portal is the right general fix, but portaling every Select to document.body breaks the existing language-menu contract. The menu is no longer a descendant of .lang-toggle/.sidebar, so the contextual glass fallback and, more importantly, the mobile rule that moves .select-dropdown-beside above the trigger no longer apply. The shared right-placement calculation can also keep a 120px menu at top=700 in an 800px viewport and enforce maxHeight >=120, leaving it outside the viewport (the new test currently expects this overflow). Please carry the required variant/style explicitly into the portal, clamp or flip right-placement menus to the viewport, and add mounted desktop/mobile portal regressions rather than only testing the pure coordinate helper. Re-request after exact-head CI is green.

lidge-jun:
Re-reviewed against @Ingwannu's CHANGES_REQUESTED — the two behavioral concerns are already resolved on the current head, and I added the missing mounted regressions (pushed to this branch as maintainer edit, `33aaa5bb`):

**(A) language-menu contract** — portaling is opt-in (`portal` defaults to `false` in `gui/src/ui.tsx`; only the Claude settings selects pass `portal`). The language `Select` doesn't portal, so its dropdown stays under `.custom-select`/`.sidebar` and keeps `.select-dropdown-beside`, so the glass fallback and the mobile upward-placement rule still apply.

**(B) right-placement overflow** — `computeSelectMenuStyle` now compares space above/below and flips upward near the bottom (`gui/src/select-position.ts`); the 800px-viewport / 120px-menu case returns `bottom`, not `top=700`. The tests assert the flip rather than encoding the overflow.

**(C) mounted regressions** — added `gui/tests/select-portal.test.tsx`: (1) a portal `Select` renders its dropdown under `document.body`, outside the clipping card, with `select-dropdown-portal`; (2) a non-portal `Select` keeps its dropdown inside `.custom-select` with `.select-dropdown-beside`. The pure-coordinate viewport flip/clamp stays covered by `select-position.test.ts` (happy-dom has no real layout height to drive a mounted flip).

`lint:gui` + `build:gui` pass; 9 gui select tests green. Will merge once the branch CI is green.

PR #390, opened 2026-07-24
fix(codex): clear stale weekly quota on WHAM/header refresh (#382)

Wibias:
## Summary

Fixes #382.

Team Codex Auth accounts with a WHAM `primary_window` of ~30.4 days were still showing stale `weeklyPercent` alongside `monthlyPercent` after refresh, because quota updates were merge-only and upstream primary headers were always treated as weekly.

## Changes

- Add snapshot replacement via `setAccountQuotaFromParsed()` so WHAM refresh/login clears absent window fields
- Classify upstream primary quota headers by `x-codex-*-window-minutes` (>= 28 days => monthly) via `applyAccountQuotaFromUpstreamHeaders()`
- Add regression tests for stale weekly clearing and header classification

## Test plan

- [x] `bun test tests/rate-limit-reset-credits.test.ts tests/codex-auth-api.test.ts`
- [x] `bun run typecheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Codex account quota persistence by deriving quota from upstream headers and replacing snapshots consistently.
  * Prevented stale weekly quota fields from carrying over when only monthly quota is present.
  * Correctly interprets monthly vs weekly quota windows (including combining primary/secondary where applicable) during account refresh, authentication reauth, and passthrough responses.
* **Tests**
  * Added regression and quota-snapshot replacement coverage for monthly-only, weekly, credits-only, and combined header scenarios, including `GET /api/codex-auth/accounts?refresh=1`.

PR #394, opened 2026-07-24
fix(anthropic): guard premature no-tool completions

duansy123:
## Summary

- detect the high-confidence Anthropic case where an actionable request ends with a short execution claim but no tool call
- keep the response open and make at most one internal continuation, preserving one terminal Responses event and merging usage from both upstream calls
- avoid auto-continuing normal answers, clarification questions, explicit plan-only requests, substantive final answers, and recent tool-backed completions
- preserve the first assistant output as a separate Responses item before any tool call from the continuation

## Tests

- `bun run test` — 3947 pass, 0 fail
- `bun run typecheck`
- `bun run privacy:scan`
- focused terminal-guard and Anthropic bridge tests
- isolated live probe against a configured Anthropic-compatible account: HTTP 200, one `response.completed`, zero `response.failed`, and the expected client tool call in the same Responses turn

## Summary by CodeRabbit

- **New Features**
  - Anthropic requests can now perform one bounded continuation when an action is requested but no tool call is produced.
  - Streaming and non-streaming responses now segment assistant output cleanly across the initial response and its one-shot continuation.

- **Bug Fixes**
  - Prevents unsafe/unbounded continuation behavior and ensures continuation token usage and tool-call results are aggregated correctly.
  - Cancels the continuation fetch body if the client aborts mid-continuation.

- **Documentation**
  - Clarified Anthropic terminal-guard continuation behavior and which response types are not retried.

- **Tests**
  - Added coverage for continuation decisions, boundary emission, usage aggregation, retry limits, and late client abort cancellation.

### DIFF
--- a/docs-site/src/content/docs/reference/adapters.md
+++ b/docs-site/src/content/docs/reference/adapters.md
@@ -64,6 +64,10 @@ streams the response back **untranslated**.
 - Always sends `anthropic-version: 2023-06-01`. Streams `content_block_delta` (`text_delta`,
   `thinking_delta`, compatible `reasoning_delta`, `input_json_delta`). The SSE decoder preserves
   event state across fetch chunks and accepts a terminal `message_stop` without a trailing newline.
+- For routed Anthropic Responses turns with client tools, a bounded terminal guard detects the
+  high-confidence case where the user requested an action but Claude ends with an execution claim
+  and no tool call. It performs at most one internal continuation; normal answers, clarification
+  questions, tool-using turns, and transport-incomplete responses are not auto-retried.
 
 ## `google`
 
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -456,6 +456,17 @@ export function bridgeToResponsesSSE(
             if (event.type !== "done" && event.type !== "incomplete" && event.type !== "error") continue;
           }
           switch (event.type) {
+            case "assistant_boundary": {
+              // A guarded continuation starts a fresh assistant output item while keeping the
+              // intermediate, suspicious text in the same Responses turn.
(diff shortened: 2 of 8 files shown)


lidge-jun:
I rebased this onto current `dev` (`ac49d13d`) and pushed to your branch (maintainer edit) — the `src/server/responses/core.ts` conflict was against #390's `cancelBodyOnAbort` guard, resolved by keeping both that line and your terminal-guard block.

While rebasing, an independent review flagged two blockers, which I've also fixed in the pushed commit (`d17f4350`):

1. **Continuation abort-race** — the terminal-guard continuation opens a *second* upstream response, but its body wasn't bound to the abort signal, so a late client abort during the continuation reopened the Bun fetch-to-reader race that #390/`366e3053` fixed for the initial response. Fixed by wrapping the continuation parse in `cancelBodyOnAbort(response.body, upstream.signal)` with a `finally` detach, mapping a genuine abort to 499, and redacting continuation parse-error messages. Regression: `tests/abort-race.test.ts` ("terminal-guard continuation body is cancelled on a late client abort").
2. **Plan-only false positive** — `PLAN_ONLY_REQUEST_RE` only matched explicit "don't call tools" wording, so a plain "write a concise implementation plan" request was treated as a suspicious no-tool completion and nudged toward tool execution. Extended the regex to recognize ordinary plan/proposal/draft/outline requests (EN + zh). Regression: `tests/terminal-guard.test.ts` (6 phrasings).

Local `bun run typecheck` + `privacy:scan` pass; 53 targeted tests green. Waiting on the full CI matrix, then I'll merge. Please shout if you'd rather adjust the fixes.

PR #403, opened 2026-07-24
feat(grok): auto-configure Grok Build and harden strict-client compatibility

lidge-jun:
## Summary

- Automatically register the visible opencodex model catalog in Grok Build through a fenced managed block in `~/.grok/config.toml`.
- Refresh the managed catalog during proxy startup and ensure/restart flows, and remove it during supported teardown paths.
- Emit token-detail objects with zero defaults in Responses and Chat Completions usage payloads for strict OpenAI-compatible clients.
- Preserve raw adapter usage separately from normalized wire usage so synthetic zero defaults do not affect request-log provenance or cost diagnostics (`usageFromBridge`).
- Pin generated Grok models to the Chat Completions backend and verify that Responses heartbeat frames are not leaked through the translation layer.
- Add a Grok Build integration guide and focused regression tests.

## Motivation

Grok Build's pinned OpenAI client requires usage detail objects that some routed providers do not naturally report. Omitting those objects can make an otherwise successful turn fail during response deserialization.

Grok Build also does not reliably apply inherited provider `base_url` settings to inference routing. The generated configuration therefore uses direct per-model fields and keeps the visible opencodex catalog synchronized through a single managed block.

## Configuration safety

- Existing user configuration outside the managed fence is preserved.
- A one-time backup is created before first modifying a pre-existing config.
- Existing user-owned model aliases are reserved to avoid duplicate TOML tables.
- Quoted and whitespace-padded headers, including TOML Unicode escapes, are recognized during collision detection.
- An incomplete/orphaned managed fence causes the operation to fail closed without modifying the file.
- Writes use the repository's atomic configuration writer.

## Rebase note

Rebased onto current `origin/dev` (was 209 behind, conflicting in `src/cli/index.ts` and `src/server/responses.ts`). The `responses.ts` `onUsage` provenance hooks were re-applied to the post-split `src/server/responses/core.ts`; the CLI inject/strip calls were merged with the current `currentExternalCodexModelProvider` guard.

## Tests

- `bun run typecheck` ✅
- `bun run test` ✅ (4049 pass / 0 fail)
- `bun run privacy:scan` ✅

Focused coverage includes: always-present Responses and Chat Completions usage detail objects; raw adapter usage provenance across completed, incomplete, failed, and EOF paths; idempotent Grok config injection and cleanup; user alias collision handling and malformed-fence safety; catalog refresh through `syncGrokConfig`; heartbeat consumption without leaking raw Responses frames.

## Summary by CodeRabbit

* **New Features**
  * Added Grok Build integration with automatic model catalog configuration and managed fenced config block.
  * Added lifecycle support to synchronize settings across start, ensure, restart, stop, and uninstall.
* **Bug Fixes**
  * Ensured token usage “details” are always emitted with zero-value defaults across Responses and Chat Completions.
  * Improved usage provenance to prevent metadata overwrites.
  * Prevented `response.heartbeat` frames from leaking into Chat Completions outputs.
* **Documentation**
  * Added/updated Grok Build setup guide and known limitations.
* **Tests**
  * Expanded coverage for Grok config injection/sync, usage reporting, streaming conversion, and lifecycle behavior.


lidge-jun:
Reviewed — four blocking findings:

1. **High — service ownership guard failure is ignored.** `cli/index.ts:390-398,440-450`: when `stopServiceIfInstalled()` refuses (different `OPENCODEX_HOME` owner), the code still tears down the proxy and strips `~/.grok/config.toml`. The shared Grok config can be removed while the target service keeps running or gets recreated. Ownership failure should propagate, or shared teardown should only proceed with evidence the target proxy actually stopped.

2. **High — non-loopback auto-config writes a placeholder that gets overwritten.** `grok/inject.ts:106-135`: every model gets `api_key = "opencodex-loopback"`, but non-loopback servers need a real admission token. The docs tell users to replace it, yet `start`/`ensure`/`restart` regenerate the managed block and overwrite the real token with the placeholder. Reproduced: `REAL_TOKEN_PRESERVED=false`. Non-loopback should refuse auto-registration and guide manual config outside the fence.

3. **Medium — quoted TOML key collision invalidates the entire Grok config.** `grok/inject.ts:64-76`: alias lookup doesn't recognize `["model"."ocx-mine"]` or `['model'.ocx-mine]`, then creates `[model.ocx-mine]` → duplicate table → `Cannot redefine key 'model'`. Need first-segment TOML canonicalization + regression for both quoted forms.

4. **Medium — no focused lifecycle regression.** The changed `start`, both `ensure` branches, stop/strip, and service-ownership-failure paths aren't directly tested. Finding 1 passed through this gap.

Per MAINTAINERS.md this touches config-ownership/lifecycle boundaries and warrants a human security/lifecycle review even after fixes. Happy to re-review.

lidge-jun:
Rebased onto current `dev` and pushed `092dd749`. Every blocking finding is closed, and an independent audit of the plan and of the first implementation commit turned up more, which are fixed here too.

## Blocking findings

**1. Service ownership guard failure ignored** — `ServiceOwnershipError` now distinguishes "the manager was never contacted" from "the stop attempt failed", so `handleStop` can act on it: shared teardown (native Codex restore, Grok strip) is skipped and the stop reports failure, while the local proxy stop still runs. `revertSystemEnv` stays ungated — it carries its own ownership check and concerns launchctl user env, not `CODEX_HOME`.

Two holes the fix would otherwise have left open:

- The daemon's own `syncCleanup` carries the same check. Without it, `SIGTERM` strips the fence through the exact path `handleStop` just declined.
- `stopProxy` no longer escalates a refusal into a forced kill. `POST /api/stop` now answers `409` on ownership mismatch, and treating that as "endpoint dead" would `SIGTERM` the proxy and run the cleanup anyway.

`handleStop` also returns its outcome instead of calling `process.exit`, so `restart` can decline to re-inject and the tray restart's `start()` half is no longer stranded. Note the intended consequence: an ownership mismatch now exits non-zero, so `ocx update`'s stop gate refuses the update until the command runs from the installed home.

**2. Non-loopback placeholder overwritten** — auto-registration is now refused on non-loopback binds, including the wildcards `0.0.0.0` and `::`. Injection returns `skippedReason: "non-loopback"`, removes any block left over from an earlier loopback bind, and writes nothing further, so a user-managed `api_key` survives repeated syncs — your `REAL_TOKEN_PRESERVED=false` reproduction inverts.

I looked at whether `env_key` could keep auto-registration alive without writing a secret. It cannot, and it is worse than the status quo: `ConfigModelOverride` accepts `env_key`, but the fail-closed path in `config.rs:3510` requires `model_provider`, which our block deliberately omits. With no own credential, `resolve_credentials` (`config.rs:4689-4715`) falls through to the signed-in xAI session bearer and sends it to whatever `base_url` the entry names — a plaintext LAN endpoint. Upstream's own tests pin this (`resolve_credentials_empty_env_key_falls_through_to_session`). Today's placeholder is a non-empty own credential, so it short-circuits and merely 401s. Refusing is the right call; `env_key` is documented only as a manual recipe, with that warning.

**3. Quoted TOML key collision** — both key segments now go through one `KEY_SEGMENT` grammar, so `["model"."x"]` and `['model'.x]` are canonicalized like the bare form. Also reserved: `[[model.x]]`, which genuinely collides, and `[model.x.sub]`, which does not collide but costs only a suffixed alias to avoid.

A correction to my own plan: I had claimed sub-table headers collide. A TOML parser says they do not, in either order. The reservation stays as conservatism, not as a collision fix.

The first version of this regex had a real defect that review caught: the sub-table tail used `[^\]]*`, which matches newlines. An unclosed `[model.…` inside a multiline string swallowed the following lines and hid a real `[model.<alias>]` header, so it went unreserved and we emitted the duplicate — reintroducing the whole-config-layer rejection this scan exists to prevent. Every class is newline-free now, with that document as a regression test.

**4. No lifecycle regression** — `tests/grok-lifecycle.test.ts` covers `handleStart`'s sync wiring, both `handleEnsure` branches (live uses `live.hostname`, spawn uses `config.hostname`), stop-time strip behind the ownership gate, the refused-strip failure, the `OCX_SERVICE` exclusion, service stop/uninstall strip, ownership-error typing, and the `/api/stop` 409 and strip paths.

`src/cli/index.ts` runs its command switch on import, so handler wiring is asserted against source in the pattern `tests/stale-state-purge.test.ts` and `tests/uninstall.test.ts` established; the layers that can genuinely execute are exercised directly.

## Other Codex P2 findings

**Trailing newline not restored** — the real problem was upstream of `strip`: `"X"` and `"X\n"` produced byte-identical files after injection, so no strip rule could restore both. Injection now always adds exactly one separator newline, and strip undoes exactly that. Verified byte-for-byte across no/one/two/three trailing newlines, mid-file blocks, uniform CRLF and five inject/strip cycles; the reviewer independently fuzzed 16 file shapes with zero drift. Blocks written by the previous version restore exactly when the original ended in a newline, and leave one extra newline when it did not — additive rather than destructive.

**Fence left behind on intentional shutdown** — `ocx service stop`, `ocx service uninstall`, and `POST /api/stop` all strip it now. The `OCX_SERVICE=1` crash/respawn exclusion is untouched. `/api/stop` also stops leaking the ownership error as a 500 that left the proxy running, and the dashboard clears its stopping state on a non-OK response instead of hanging on a disabled button.

## Docs and devlog

The guide now explains why the fence is loopback-only, gives a manual recipe with a reachable host and the admission token rather than an unreachable `127.0.0.1`, and warns against the `env_key` fallthrough. On reload timing: Grok does watch `config.toml` and reload on a real `[model]` change, so rather than delete the claim I replaced it with the observed mechanism, pointed at `grok inspect` for verification (it lists config sources and rejected fields, not a model list), and recorded that one TOML error invalidates the entire user layer — which is why we write atomically. Devlog `020` now recommends `chat_completions` with a dated correction citing the pinned async-openai fork having no catch-all event variant.

## Beyond the review list

The audit found four defects nobody had reported, all fixed here: the start-time Grok sync was nested inside the Desktop-3P try, so a catalog failure silently skipped the fence; `ocx stop` swallowed a refused strip and exited 0 with the fence pointing at a dead proxy; `ocx service stop` ran `launchctl unload` with nothing installed; and `/api/stop`'s unguarded `stopServiceIfInstalled()` could 500 and leave the proxy up.

Two residual risks are logged rather than fixed, both out of scope for a review-blocker pass: `config.toml.bak-opencodex` is written once and can go stale even though the orphaned-marker message points users at it, and root-dotted spellings (`model.x.k = 1`, `[model]` plus a dotted key) collide but are not reserved.

## Verification

At `092dd749`, rebased onto `dev` `4b80612e`:

- `bun run typecheck` — clean
- `bun run test` — 4288 pass, 0 fail, 21876 assertions across 330 files

Planning and evidence live in `devlog/_plan/260726_grok_build_prod/`. Ready for another look — the ownership and teardown changes in particular are the ones worth a human read.

lidge-jun:
Self-review note at head `092dd749`, recorded during a maintainer integration pass. This is my own PR, so it needs the other maintainer's approval regardless — `MAINTAINERS.md` prohibits self-approval and green CI does not substitute for it. Recording the findings here so that review has them.

**This should not land as one commit.** It changes roughly 40 files and mixes four separable concerns: strict usage-shape compatibility, the Grok config parser/injector, lifecycle/teardown, and docs/GUI. It also carries unrelated devlog churn. Split into four, in that order — the strict-client slice is independently useful and reviewable, while the lifecycle slice needs to be re-cut against whatever the service/process-control contracts look like when it lands.

**The `src/lib/process-control.ts` conflict must preserve both sides.** `dev` gained a 10-second graceful-stop request window plus `waitForStoppedPort()` port-reclaim logic; this PR adds an HTTP 409 refusal result. Taking either side wholesale loses required behavior. The resolved shape is `GracefulStopResult = "stopped" | "unreachable" | "refused"`, with `stopProxy()` throwing without force-kill on `refused` while retaining the port-reclaim wait after both `stopped` and a forced kill.

**Outstanding defects found:**

`src/bridge.ts` — usage observer exceptions can break otherwise valid responses. The direct `options?.onUsage?.(event.usage)` calls should all route through the existing best-effort `reportUsage()` helper, at the completed, incomplete, failed, adapter-EOF, and non-stream JSON paths.

`src/grok/inject.ts` — a failed cleanup is reported as a successful policy skip. When `stripGrokConfig()` fails on the non-loopback path, the function still returns `ok: true`. It should propagate the failure with a message saying both that registration was refused and that the previous managed block could not be removed.

`src/cli/index.ts` — a generic service-stop failure still permits shared teardown. Only `ServiceOwnershipError` sets `ownershipBlocked`; an ordinary stop exception is logged but `restoreNativeCodex()` and `stripGrokConfig()` run anyway, while the service may still be alive. Replace the booleans with an explicit `ManagedServiceStopResult` of `not-installed | stopped | failed | ownership-refused`, and allow shared teardown only for the first two. This logic should move into an import-safe module such as `src/cli/lifecycle.ts` — the current source-inspection tests are false confidence.

`src/server/management-api.ts` — the direct `process.exit(0)` is untestable. Add an injectable `shutdown` to `ManagementApiDeps` defaulting to the real exit. A generic `stopServiceIfInstalled()` failure should return 500 before restore, strip, drain, or shutdown; ownership refusal stays 409.

`gui/src/App.tsx` — backend English is surfaced directly. Add a stable error code such as `SERVICE_OWNERSHIP_MISMATCH`, map it through every locale, and alert localized copy with the server message as optional detail.

**Docs overclaim.** "restore your file byte-for-byte" is not the actual contract: uniform-EOL content is preserved, but mixed-EOL files are normalized by `applyEol`. The frontmatter and introduction should also state that automatic registration is loopback-only.

Suggested regression once `handleStop` is extracted: an ordinary service-stop failure must reach neither `restoreNativeCodex()` nor `stripGrokConfig()`, returning `{ ok: false, service: "failed", sharedTeardown: "skipped" }`. That fails against the current head.

---

<sub>Recorded during a maintainer PABCD pass. Held pending split and second-maintainer review.</sub>

lidge-jun:
Closing this PR because its work has landed on `dev` through a local integration merge rather than through the PR merge button.

**Containment proof.** This PR's head commit `092dd749102ae568b11c89b9c8e3a57bfb2b877d` is now an ancestor of `dev`:

```
$ git merge-base --is-ancestor 092dd749102ae568b11c89b9c8e3a57bfb2b877d dev
$ echo $?
0
```

It arrived via the integration merge `f6520fcd` ("Merge branch 'codex/260726-grok-build-prod' into dev"), and `origin/dev` is now `2f3fa584`. Nothing from this branch is lost.

**One conflict was resolved on the way in.** `src/lib/process-control.ts` `stopProxy()` had diverged on both sides: `dev` had added a post-stop `waitForStoppedPort()` drain, while this branch had added the 409 `"refused"` ownership guard that must not escalate to `killProxy()`. The resolution keeps both — the runtime port is read up front, a `refused` result throws instead of forcing, and the drain wait runs on both the graceful and the kill path.

The full Grok Build surface merged intact: the fenced managed-block inject/strip for `~/.grok/config.toml`, TOML key canonicalization with byte-for-byte user-config preservation, the non-loopback auto-registration refusal, the deterministic fence on `ensure`/`restart`, and the always-emit usage-detail objects for strict Responses/chat clients.

The head branch `codex/260723-grok-build-bridge` is being retired as part of this cleanup, along with `codex/260726-grok-build-prod` (same SHA) and `claudedesktop` (also merged into `dev`).

PR #412, opened 2026-07-24
Redesign connected provider workspace

HaydernCenterpoint:
## Summary
- redesign the Providers workspace around connected accounts, configured API keys, and preserved Account/API Key/Custom actions
- show real quota windows with a 30-day token fallback in compact percentage bars
- use equal-height two-column provider cards, a Providers-only wide shell, and explicit mobile collapse
- add connected-provider copy to every supported GUI locale and focused regression coverage

## Scope
Intentionally excludes smart routing, OpenRouter free-only behavior, provider logo assets/resolver changes, and the Add Provider modal scroll fix.

## Dependencies
Depends on `feat/provider-model-loader` at `60046f85` (provider model loader UI). This PR targets `dev` per repository policy, so the dependency must land first or this branch must be rebased after it lands.

## Validation
- `bun test ./gui/tests/provider-catalog.test.ts`
- `bun run typecheck`
- `bun run lint:gui`
- `bun run build:gui`


Wibias:
Thanks for opening this PR — I appreciate the work and the direction.

This overlaps with a design I'm already implementing in the background: bringing the remaining tabs into the shared workspace mode so the GUI stays consistent across sections.

Because of that, I'm not planning to merge this version as-is. Once my workspace rollout branches are in, I'd be happy to take another look if you want to rebase or open a fresh PR against the updated layout.

Thanks again for contributing.

HaydernCenterpoint:
@Wibias I made an mistake, I should pull request 1 time so you guys can get an overall look what i did. I have to redesign this because it get a better look on how people connect providers. Adding the providers and only show providers you have connected made people easier to connect, and a clearly design so i think a redesign is need rather than the design we are using now. But thanks for the review, i will improve it and push it later. Hope you have a nice day!

PR #406, opened 2026-07-24
feat(providers): add safe model discovery

HaydernCenterpoint:
## Summary
- discover provider model catalogs through validated destinations
- normalize OpenAI and Google-style model envelopes
- retain safe static fallback behavior

## Dependency
Merge after #405. The provider-directory commits disappear from this diff after that PR lands.

## Tests
- focused provider discovery tests

## Summary by CodeRabbit

- **New Features**
  - Added a provider directory with expanded integration listings, support details, documentation links, and model availability information.
  - Added provider preset discovery, including live model lookup, static-model fallback, and clearer discovery results.
  - Added support for selecting official regional endpoints for SiliconFlow.
  - Directory-only providers can now be configured and used for chat requests.

- **Improvements**
  - Improved compatibility with varied provider model-list responses, including Google-style formats.
  - Strengthened endpoint validation and redirect handling for safer model discovery.

PR #413, opened 2026-07-24
feat: add OpenRouter free-only model discovery

HaydernCenterpoint:
## Summary
- add a free-only toggle to OpenRouter model discovery
- preserve and validate complete upstream pricing before filtering
- reject missing, partial, invalid, and negative pricing from free detection
- keep static fallback restricted to OpenRouter's explicitly free model IDs

## Dependencies
This PR depends on the provider directory, provider discovery, and provider model loader PRs being merged first.

## Validation
- bun test tests/provider-free-directory.test.ts tests/provider-model-envelope.test.ts
- bun run typecheck
- bun run lint:gui


HaydernCenterpoint:
Closed in favor of the consolidated provider upgrade PR, which combines the work from #406 through #413 into one reviewed branch.

PR #421, opened 2026-07-24
feat(menubar): redesign as macOS status widget with auto-discovery & Apple-style UI

genglintong:
## Summary

- Rebuild `menubar/` UI as four-tab status widget: **Usage → Health → Status → Activity**
- Auto-discover proxy URL + port from `~/.opencodex/runtime-port.json` — zero manual config
- Skip auth entirely when proxy has no `apiKeys` configured (the default)
- Secure IPC proxy: API token stays in Rust memory (macOS Keychain), injected via `reqwest` header — never crosses to WebView JS
- Apple-style white theme: segmented control tabs, green pulse indicator, tabular-nums stats, rounded popover with shadow
- Official opencodex cloud-bracket logo for tray icon (template) and app icon (squircle + indigo gradient)
- Write actions: restart proxy (`POST /api/stop`) + provider enable/disable (`PATCH /api/providers`)
- Zero proxy changes — uses only existing management API endpoints

## Preview

menubar preview

## Architecture

```
macOS Tray Icon (Tauri v2)
  └─ Rust Core: tray lifecycle, popover window, Keychain, header injection
       └─ WebView (React 19 + TS): usage / health / status / activity sections
            └─ IPC invoke("api_request") → Rust reqwest → localhost management API
```

Rust layer is intentionally thin (~170 lines across 4 modules): `tray.rs`, `keychain.rs`, `discover.rs`, `api.rs`. All business logic lives in TypeScript.

## Key design decisions

| Decision | Rationale |
|---|---|
| JS-side fetch via Rust IPC proxy | Token never enters WebView memory; TS maintainability preserved |
| Auto-discovery from `runtime-port.json` | No setup friction; proxy port may differ from default 10100 |
| Separate lightweight React app | Full `gui/` is too heavy for a popover; copied only format utilities |
| Vite 6 + plugin-react 4 | Node 22.3 compat (Vite 7+ requires 22.12+) |
| `macOSPrivateApi: true` | Required for transparent window → true rounded corners + shadow |

## Verification

- `npx tsc --noEmit` (menubar/): passed
- `cargo check` (menubar/src-tauri/): passed, 0 warnings
- Root `bun run typecheck`: unaffected (menubar/ excluded from root tsconfig)
- Root `bun run test`: unaffected (no proxy source changes)
- Privacy grep: no hardcoded secrets in menubar/ source
- Live test: app connects to running proxy at localhost:10100, renders real usage data (3135 requests, 368M tokens, $150 est. cost over 7d)

## Non-goals (v1)

- Windows / Linux support
- macOS native notification push
- Multi-proxy-instance management
- OAuth / account operations from the menu bar
- Auto-update of the menu bar app itself
- DMG / Homebrew distribution (cargo build from source)

## Summary by CodeRabbit

- **New Features**
  - Added a macOS OpenCodex menubar app with usage, health, status, and activity views.
  - Added proxy discovery, connection setup, secure credential storage, provider controls, quota indicators, request activity, and quick dashboard access.
  - Added automatic status polling, refresh actions, and offline/error states.

- **Documentation**
  - Added guidance for issue tracking, triage labels, domain documentation, and menubar development.

- **Tests**
  - Added coverage for key UI components, formatting, and polling behavior.


lidge-jun:
GUI decision plz add screenshots Ill decide

genglintong:
Here are the screenshots!

### Popover — Usage tab (white theme, real data from localhost:10100)
usage

### App icon (squircle + official opencodex logo)
icon

### Tray icon (monochrome template, adapts to dark/light menu bar)
The tray icon uses the official cloud-bracket logo as a macOS template image — automatically inverts for dark/light menu bar.

---

**Auto-discovery**: reads port from `~/.opencodex/runtime-port.json`, no manual config. When `apiKeys` is empty (default), connects without auth.

**Security**: token (if configured) stays in Rust memory via macOS Keychain, injected via `reqwest` header — never crosses IPC to WebView.

Happy to adjust the UI style, layout, or icon based on your feedback!

lidge-jun:
Thank you for this, and in particular for the review-response work at `049ef2ac` — I want to lead with that rather than bury it.

I've built the companion directly on `dev` in a new `app/` directory (branch `feat/macos-app`), consolidating this PR and #387 into one implementation. Closing this one, but the product architecture is substantially yours.

**First, a correction I owe you.** My initial survey said this branch still committed `menubar/src-tauri/target/**` with developer-absolute paths and failed `bun run privacy:scan`. That was true of the head the bots reviewed, and you had already fixed it. Verified against `049ef2ac`:

```
gh pr view 421 --json files --jq '[.files[].path | select(test("src-tauri/target"))] | length'
→ 0
```

An adversarial review of my own plan caught the stale claim before I posted it. Repeating it would have been both wrong and unfair to a contributor who responded to review properly.

**What came from here**

- HTTP management-API transport. This is the decision that shaped the maintainer implementation most: it consumes only endpoints that already exist and needs no proxy change, which is exactly right.
- `runtime-port.json` discovery with the 10100 fallback.
- Keychain-backed key storage.
- The usage / health / status information set.
- Tabular-numeral stat treatment. Live data reaches 3.6e10 tokens, and without tabular figures the digits jitter on every poll.
- Skipping auth entirely when the proxy has no `apiKeys` configured — correct, since `isApiAuthRequired()` returns false for a loopback bind.

**Why not Tauri**

Three reasons, none of them the artifact issue you already fixed:

1. **No repository distribution path.** `.github/` is untouched, so nothing in CI builds or publishes a bundle. `build-app.sh` produces a `.app` and a `.dmg` locally, but a user still needs `rustup` plus a frontend toolchain. That was the deciding factor: an app a user must compile is not a shipped app.
2. **Build stack weight** for a project whose premise is a single Bun process — a Rust toolchain, a Cargo lockfile, generated ACL schemas, and a WebView runtime.
3. **`macOSPrivateApi: true`** is a notarization and App-Store-rejection risk. `NSPopover`… well, see below.

**One claim I could not carry over.** The PR body says the token "never crosses to WebView JS", but `menubar/src/api.ts:12-13` at this head returns it into renderer memory:

```ts
const discovery = await invoke<{ url: string; token: string | null; found: boolean }>("discover_proxy");
proxyConfig = { url: discovery.url, token: discovery.token };
```

The Rust IPC shape is still reasonable; the isolation property just is not there yet. Flagging it because it is worth fixing wherever that code lives next.

**And one place you were right and I was wrong.** I moved your four-tab layout to a single scroll-free column so "is it running?" is answered without a click — but I also assumed `NSPopover` would be the right primitive, and it was not. From an `LSUIElement` process its window never appears in `NSApp.windows` and reports `canBecomeKey == false`, so macOS refuses to route key events to it. Escape simply never arrives. Three attempted fixes later I measured it and replaced it with a key-capable `NSPanel`. Your `macOSPrivateApi` choice was solving a real problem in that area; I just landed somewhere different.

The full head-to-head is in `devlog/_plan/260725_macos_menubar_app/001_pr_survey.md` on the branch. Review very welcome — `feat/macos-app`.

lidge-jun:
Correction to my closing comment above — an adversarial review of the maintainer branch caught two overclaims I made, and I would rather fix them here than leave them standing on a closed PR.

**1. "Keychain-backed key storage" is not something I actually adopted.** The maintainer app can *read* a key from `com.opencodex.menubar` and retries once with it, but nothing calls `Keychain.write` and there is no key-entry UI. "Add key…" just opens the dashboard. So a user with a non-loopback proxy stays stuck on "Needs API key". I have corrected the guide in all five locales to say that plainly, and native key entry is on the list. Your Rust-side Keychain handling is still the more complete implementation of that idea; I should not have credited it as carried over when it is not.

**2. "Skipping auth entirely when the proxy has no `apiKeys` configured" is not how the server works.** `isApiAuthRequired()` in `src/server/auth-cors.ts` keys off whether the bind is loopback, not whether `config.apiKeys` is empty. The observable behaviour on a default install is the same — no credential needed — but my description of the mechanism was wrong.

Everything else in that comment stands, and I re-verified each specific claim before posting it: zero `src-tauri/target` paths at `049ef2ac`, the `api.ts:12-13` token exposure, `macOSPrivateApi: true`, and the `NSPopover` `canBecomeKey == false` finding.

Apologies for the noise on a closed thread. A credit that is wrong is worse than no credit, so it seemed worth the correction.

lidge-jun:
One more correction, and then I will stop cluttering this thread.

My correction above said the app "can read a key from `com.opencodex.menubar`". The actual Keychain service is `com.opencodex.menubar.apikey` with account `default` — and more to the point, it is a data-protection Keychain item, which Keychain Access cannot create. So there is no supported way to provision that key by hand at all, and pointing anyone at the identifier would have sent them down a path that does not work either.

The guides now say exactly that instead: non-loopback binds are not supported yet, a loopback proxy needs no key, and native key entry is planned.

Flagging it because I got the same detail wrong twice in a row on this thread, and the second version was still misleading even though it was more cautious than the first.

PR #355, opened 2026-07-23
feat(google): enable Gemini inline image output

tizerluo:
## Summary

- Enable Gemini image generation through Cloud Code Assist (Antigravity) with two complementary paths:
  1. **`/v1/images/generations` fallback** (primary): When no OpenAI upstream is configured, Codex's built-in `image_gen` skill now falls back to CCA's dedicated image model (`gemini-3.1-flash-image`). This is the standard Codex workflow — the model calls the tool, codex-rs POSTs to the Images API, and the proxy serves it via CCA.
  2. **Inline `inlineData` parsing**: For users who select `gemini-3.1-flash-image` directly as their chat model, `inlineData` parts in the response are materialized to `~/.opencodex/artifacts/` and emitted as markdown.
- Add `responseModalities: ["TEXT", "IMAGE"]` to `generationConfig` for non-thinking Gemini models.
- Add `gemini-3.1-flash-image` to the CCA model catalog (from `FetchAvailableModels.imageGenerationModelIds`).
- Automatically omit `responseModalities` when a thinking level is active (thinking models reject IMAGE modality).

## How it works (Codex workflow)

```
User: "draw me a cat"
  → Chat model calls image_gen.imagegen tool
  → codex-rs POSTs /v1/images/generations {prompt: "a cat", model: "gpt-image-1"}
  → OpenCodex: no OpenAI upstream? → CCA fallback
  → Calls gemini-3.1-flash-image via v1internal:generateContent
  → Returns {created, data: [{b64_json: "..."}]}
  → codex-rs renders the image
```

No model switching required. Works with just a CCA login (`ocx login google-antigravity`).

## Verification

- `bun run typecheck` — clean
- `bun test tests/images/gemini-inline.test.ts` — 8/8 pass
- `bun test tests/google-adapter.test.ts tests/google-vertex-stream.test.ts` — 17/17 pass
- `bun run privacy:scan` — passed
- **E2E (Images API)**: `POST /v1/images/generations {"prompt":"A cute green frog on a lily pad"}` → 2MB JPEG returned in standard `{created, data:[{b64_json}]}` format
- **E2E (inline)**: `POST /v1/chat/completions` with model `gemini-3.1-flash-image` → `image` markdown with materialized 515KB JPEG

Note: `bun run test` (full suite) has 24 pre-existing failures on `dev` in catalog/discovery tests, unrelated to this change.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed.
[...]

PR #424, opened 2026-07-24
feat(images): add Grok image bridge for non-OpenAI models

tizerluo:
## Summary

When Codex routes to a non-OpenAI model (Claude, Gemini, Grok, etc.), the `image_generation` hosted tool stops working because it requires OpenAI's server-side execution. This PR adds an **Image Bridge** that:

1. Detects `image_generation` in the request's tools array
2. Replaces it with a synthetic function tool (`image_gen`) the routed model can call
3. Intercepts the model's tool call and routes it to **xAI Grok Imagine** (`/images/generations`)
4. Materializes generated images to `~/.opencodex/artifacts/` with magic-byte format detection
5. Injects the local file path back into the conversation as a tool result
6. Re-calls the model so it can reference the image in its final response

The architecture mirrors the proven `src/web-search/` sidecar pattern — same async-generator loop, `bridgeToResponsesSSE`, heartbeat, and abort-signal linkage.

## New Files

| File | Purpose | Lines |
|------|---------|-------|
| `src/images/types.ts` | Shared types (`ImageBridgePlan`, `ImageCallResult`) | 19 |
| `src/images/synthetic-tool.ts` | Tool injection + name detection + extraction | 69 |
| `src/images/xai-client.ts` | xAI `/images/generations` and `/images/edits` client | 72 |
| `src/images/fulfill.ts` | Execute one image call, xAI API, materialize to disk | 69 |
| `src/images/plan.ts` | Decide whether to activate bridge (provider + token check) | 52 |
| `src/images/loop.ts` | Core agentic loop (max 3 rounds, SSE bridge) | 372 |
| `src/images/index.ts` | Barrel export | 4 |
| `tests/images/*.test.ts` | 5 test files, 53 tests covering all modules | — |
| `docs-site/.../image-bridge.md` | User guide | 63 |

## Modified Files

- `src/types.ts` — `_imageGeneration` stash field, `OcxTool.imageGeneration` flag, `OcxImagesConfig` bridge fields
- `src/responses/parser.ts` — `extractHostedImageGeneration()` + stash into parsed result
- `src/server/responses/core.ts` — image bridge trigger (after web-search block, before standard path)
- `src/images/artifacts.ts` — extended with `downloadImageToArtifact()` + `guessExtFromMagic()` (shared helpers extracted)

## Verification

- `bun x tsc --noEmit` passes (0 errors)
- `bun test tests/images/` passes (53 tests, 0 failures)
- `bun run privacy:scan` passes
- Full test suite: 4097 pass / 26 fail (all 26 failures are pre-existing, unrelated)
- Manual integration test: sent "draw a simple cat" with grok-4.5 + image_generation tool. Image generated via xAI Grok Imagine, materialized to disk (864x1152 JPEG, 108KB), model produced markdown response with local path.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added an image-bridge for `image_generation` on non-OpenAI providers using an interception/synthetic tool and bounded multi-round tool-call looping.
  * Materializes images into local artifacts from both base64 and URLs, returning a primary Markdown image link and allowing partial successes.
  * Added configurable bridging (enablement, bridge model, max rounds) with xAI Grok Imagine support.
* **Documentation**
  * Published an “Image Bridge” guide with setup, config locations, and limitations (streaming-only).
* **Bug Fixes**
  * Enforced streaming-only behavior and improved validation/error handling for image fulfillment and looping.
* **Tests**
  * Added Bun test coverage for planning, tool matching, fulfillment, artifact materialization, and streamed loop behavior.

### DIFF
--- a/docs-site/src/content/docs/guides/image-bridge.md
+++ b/docs-site/src/content/docs/guides/image-bridge.md
@@ -1,0 +1,67 @@
+---
+title: Image Bridge
+description: Route image_generation hosted-tool calls to xAI Grok Imagine when using a non-OpenAI provider.
+---
+
+## Overview
+
+When you route Codex through a non-OpenAI model (Claude, Gemini, Grok, etc.), the
+`image_generation` **hosted tool** normally doesn't work — it requires OpenAI's server-side
+execution environment. The Image Bridge detects these calls and transparently reroutes them to
+xAI Grok Imagine, so the model you're actually chatting with can still generate images.
+
+## Prerequisites
+
+- **Enable the bridge** by setting `images.bridgeEnabled: true` in your config (it is off by
+  default to avoid unexpected xAI charges — see [Configuration](#configuration) below).
+- An xAI provider configured in settings with `baseUrl: "https://api.x.ai/v1"` and the
+  `openai-chat` adapter.
+- Authentication via `authMode: "oauth"` (`ocx login xai` — uses a stored, auto-refreshed
+  bearer token) or `authMode: "key"` (a configured API key).
+- A non-OpenAI model selected as your active provider. (When the active provider is OpenAI,
+  the native hosted tool is used directly and the bridge is bypassed.)
+
+## Configuration
+
+Image Bridge options live under `images` in `~/.opencodex/config.json`. Bridging is
+**opt-in** — you must set `bridgeEnabled: true` to enable paid xAI Grok Imagine generation:
+
+```json
+{
+  "images": {
+    "bridgeEnabled": true,
+    "bridgeModel": "grok-imagine-image-quality",
+    "maxRounds": 3
+  }
+}
+```
+
+| Option | Default | Description |
+| --- | --- | --- |
+| `bridgeEnabled` | `false` | Master switch. Set `true` to enable bridging. Off by default to avoid unexpected xAI charges. |
+| `bridgeModel` | `grok-imagine-image-quality` | The xAI image model id to send prompts to. |
+| `maxRounds` | `3` | Maximum number of image-generation loop iterations per turn. |
+
+## How It Works
+
+1. When Codex sends a request with `image_generation` in the tools array, OpenCodex detects it
+   during request preprocessing.
+2. The hosted tool is replaced with a **synthetic function tool** that the routed model can call
+   normally — the model sees a callable tool rather than an opaque hosted tool it can't execute.
+3. When the model invokes that tool, OpenCodex intercepts the call and sends the prompt to xAI's
+   image generation API.
+4. Generated images are saved to `~/.opencodex/artifacts/` and the **local file path** is returned
+   to the model as the tool result.
+5. The model continues the conversation with knowledge of the generated image and its location.
+
+From the model's perspective nothing changed — it called a tool and got a result. From the user's
+perspective, image generation works with any routed provider instead of silently failing.
+
+## Limitations
+
+- **Only xAI Grok Imagine is supported.** DALL-E and other image providers may be added later.
+- **Web search takes priority.** If both web search and image generation are requested in the same
+  turn, the web-search bridge runs and image generation is skipped for that turn.
+- **xAI costs apply.** Image generation via xAI requires an active xAI subscription or API credits.
+- **Streaming only.** The bridge works by intercepting the SSE response stream; requests with
+  `stream: false` are rejected with a 400 error.
--- a/docs-site/src/content/docs/guides/image-bridge.md
+++ b/docs-site/src/content/docs/guides/image-bridge.md
@@ -41,6 +41,14 @@ Image Bridge options live under `images` in `~/.opencodex/config.json`. Bridging
 | `bridgeEnabled` | `false` | Master switch. Set `true` to enable bridging. Off by default to avoid unexpected xAI charges. |
 | `bridgeModel` | `grok-imagine-image-quality` | The xAI image model id to send prompts to. |
 | `maxRounds` | `3` | Maximum number of image-generation loop iterations per turn. |
+| `artifactsKeepCount` | `200` | Maximum number of files retained under `artifacts/`. When exceeded, the oldest files are deleted automatically. |
+
+## Artifact Retention
+
+Generated images are written to `~/.opencodex/artifacts/`. To prevent unbounded disk
+growth in long-running sessions, the directory is pruned automatically after each image
+write — the oldest files (by modification time) are deleted when the count exceeds the
+configured maximum (default 200, configurable via `images.artifactsKeepCount`).
 
 ## How It Works
 
--- a/src/images/artifacts.ts
+++ b/src/images/artifacts.ts
@@ -1,0 +1,140 @@
+import { mkdir, writeFile } from "node:fs/promises";
+import { join } from "node:path";
+import { getConfigDir } from "../config";
+import { assessUrlDestination, assertUrlResolvesPublic } from "../lib/destination-policy";
+
+const MAX_DECODED_BYTES_PER_IMAGE = 50 * 1024 * 1024;
+const MAX_DECODED_BYTES_PER_RESPONSE = 100 * 1024 * 1024;
+const MAX_DOWNLOAD_BYTES = 50 * 1024 * 1024; // 50 MiB
+
+// Strict alphabet check: Buffer.from(..., "base64") silently ignores invalid
+// characters, so malformed payloads would otherwise decode to garbage bytes.
+const BASE64_RE = /^[A-Za-z0-9+/]*={0,2}$/;
+
+export interface ImageBudget {
+  spent: number;
+}
+
+export function createImageBudget(): ImageBudget {
+  return { spent: 0 };
+}
+
+function getArtifactsDir(): string {
+  return join(getConfigDir(), "artifacts");
+}
+
+function timestampPrefix(): string {
+  const now = new Date();
+  return [
+    now.getFullYear(),
+    String(now.getMonth() + 1).padStart(2, "0"),
+    String(now.getDate()).padStart(2, "0"),
+    "-",
+    String(now.getHours()).padStart(2, "0"),
+    String(now.getMinutes()).padStart(2, "0"),
+    String(now.getSeconds()).padStart(2, "0"),
+    "-",
+    String(now.getMilliseconds()).padStart(3, "0"),
+  ].join("");
+}
+
+export function guessExtFromMagic(bytes: Uint8Array): string {
+  const sig = Buffer.from(bytes.slice(0, 12)).toString("latin1");
+  if (sig.startsWith("\x89PNG")) return "png";
+  if (sig.startsWith("\xff\xd8\xff")) return "jpg";
+  if (sig.startsWith("RIFF") && sig.slice(8, 12) === "WEBP") return "webp";
+  if (sig.startsWith("GIF8")) return "gif";
+  return "png";
+}
+
+export async function materializeInlineImage(
+  base64Data: string,
+  budget?: ImageBudget,
+): Promise<string> {
+  const dir = getArtifactsDir();
+  await mkdir(dir, { recursive: true, mode: 0o700 });
+
+  const normalized = base64Data.replace(/\s+/g, "");
+  if (!BASE64_RE.test(normalized) || normalized.length % 4 !== 0) {
+    throw new Error("inline image data is not valid base64");
+  }
+  // Validate decoded size from the base64 length *before* allocating a Buffer, so a
+  // malicious or broken upstream cannot force a large allocation / OOM.
+  const padding = normalized.endsWith("==") ? 2 : normalized.endsWith("=") ? 1 : 0;
+  const decodedBytes = (normalized.length / 4) * 3 - padding;
+  if (decodedBytes === 0) throw new Error("inline image data is empty after base64 decode");
+  if (decodedBytes > MAX_DECODED_BYTES_PER_IMAGE) throw new Error(`inline image exceeds ${MAX_DECODED_BYTES_PER_IMAGE} byte per-image cap`);
+  if (budget && budget.spent + decodedBytes > MAX_DECODED_BYTES_PER_RESPONSE) {
+    throw new Error(`inline image response exceeds ${MAX_DECODED_BYTES_PER_RESPONSE} byte per-response cap`);
+  }
+
+  const buf = Buffer.from(normalized, "base64");
+  if (budget) budget.spent += buf.length;
+
+  // Sniff actual format from decoded bytes rather than trusting the declared mimeType.
+  const ext = guessExtFromMagic(buf);
+  const filePath = join(dir, `img-${timestampPrefix()}-${crypto.randomUUID()}.${ext}`);
+  await writeFile(filePath, buf, { mode: 0o600 });
+  return filePath;
+}
+
+export async function downloadImageToArtifact(
+  url: string,
+  budget?: ImageBudget,
+  signal?: AbortSignal,
+): Promise<string> {
+  if (url.startsWith("data:")) {
+    const m = /^data:([^;]+);base64,(.+)$/.exec(url);
+    if (!m) throw new Error("data URL is not a valid base64 image");
+    return materializeInlineImage(m[2], budget);
+  }
+
+  // SSRF protection: validate the provider-returned URL before fetching.
+  // Reject non-HTTP(S) schemes, literal private/loopback/link-local/metadata addresses.
+  const assessment = assessUrlDestination(url);
+  if (assessment && assessment.kind !== "public" && assessment.kind !== "hostname") {
+    throw new Error(`image URL targets ${assessment.detail}`);
+  }
+  // DNS check: resolve hostname and reject if it points at private/internal space.
+  await assertUrlResolvesPublic(url);
+  const resp = await fetch(url, { signal, redirect: "error" });
+  if (!resp.ok) throw new Error("image download failed: " + resp.status);
+
+  // Stream the body with a hard byte cap so a missing/lying Content-Length or a
+  // compromised CDN URL cannot exhaust memory before the size check runs.
+  if (!resp.body) throw new Error("image download returned no body");
+  const reader = resp.body.getReader();
+  const chunks: Uint8Array[] = [];
+  let total = 0;
+  try {
+    for (;;) {
+      const { value, done } = await reader.read();
+      if (done) break;
+      total += value.byteLength;
+      if (total > MAX_DOWNLOAD_BYTES) {
+        throw new Error(`image download exceeds ${MAX_DOWNLOAD_BYTES} byte cap`);
+      }
+      chunks.push(value);
+    }
+  } finally {
+    try { await reader.cancel(); } catch { /* ignore cancel errors */ }
+    reader.releaseLock();
+  }
+
+  const bytes = new Uint8Array(total);
+  let offset = 0;
+  for (const c of chunks) { bytes.set(c, offset); offset += c.byteLength; }
+
+  if (budget && budget.spent + bytes.length > MAX_DECODED_BYTES_PER_RESPONSE) {
+    throw new Error(`image download exceeds ${MAX_DECODED_BYTES_PER_RESPONSE} byte per-response budget`);
+  }
+
+  const ext = guessExtFromMagic(bytes);
+  const dir = getArtifactsDir();
+  await mkdir(dir, { recursive: true, mode: 0o700 });
+  if (budget) budget.spent += bytes.length;
+
+  const filePath = join(dir, `dl-${timestampPrefix()}-${crypto.randomUUID()}.${ext}`);
+  await writeFile(filePath, bytes, { mode: 0o600 });
+  return filePath;
+}
--- a/src/images/artifacts.ts
+++ b/src/images/artifacts.ts
@@ -90,7 +90,13 @@ export async function downloadImageToArtifact(
   }
 
   // SSRF protection: validate the provider-returned URL before fetching.
-  // Reject non-HTTP(S) schemes, literal private/loopback/link-local/metadata addresses.
+  // Require HTTPS strictly — plain HTTP and all other schemes (ftp, file, …) are rejected.
+  let parsedUrl: URL;
+  try { parsedUrl = new URL(url); } catch { throw new Error("image URL is not valid"); }
+  if (parsedUrl.protocol !== "https:") {
+    throw new Error(`image URL must use HTTPS, got ${parsedUrl.protocol}`);
+  }
+  // Reject literal private/loopback/link-local/metadata addresses.
   const assessment = assessUrlDestination(url);
   if (assessment && assessment.kind !== "public" && assessment.kind !== "hostname") {
     throw new Error(`image URL targets ${assessment.detail}`);
--- a/src/images/artifacts.ts
+++ b/src/images/artifacts.ts
@@ -1,3 +1,4 @@
+import { mkdirSync, readdirSync, statSync, unlinkSync } from "node:fs";
 import { mkdir, writeFile } from "node:fs/promises";
 import { join } from "node:path";
 import { getConfigDir } from "../config";
@@ -6,6 +7,9 @@ import { assessUrlDestination, assertUrlResolvesPublic } from "../lib/destinatio
 const MAX_DECODED_BYTES_PER_IMAGE = 50 * 1024 * 1024;
 const MAX_DECODED_BYTES_PER_RESPONSE = 100 * 1024 * 1024;
 const MAX_DOWNLOAD_BYTES = 50 * 1024 * 1024; // 50 MiB
+
+/** Default cap on files retained under artifacts/. Oldest files are pruned when exceeded. */
+export const DEFAULT_ARTIFACT_KEEP_COUNT = 200;
 
 // Strict alphabet check: Buffer.from(..., "base64") silently ignores invalid
 // characters, so malformed payloads would otherwise decode to garbage bytes.
@@ -21,6 +25,45 @@ export function createImageBudget(): ImageBudget {
 
 function getArtifactsDir(): string {
   return join(getConfigDir(), "artifacts");
+}
+
+/**
+ * Best-effort retention cap: when the artifact directory holds more than `maxFiles`,
+ * delete the oldest (by mtime) until the count is back under the limit. Synchronous
+ * on purpose — it runs right after each successful write and touches at most a handful
+ * of files. All errors are swallowed and logged so a prune failure never breaks an image write.
+ */
+export function pruneOldArtifacts(dir: string, maxFiles: number): void {
+  let entries: string[];
+  try {
+    entries = readdirSync(dir);
+  } catch (e) {
+    console.warn(`[images] prune: could not read ${dir}:`, e instanceof Error ? e.message : e);
+    return;
+  }
+  if (entries.length <= maxFiles) return;
+
+  let stats: Array<{ name: string; mtime: number }>;
+  try {
+    stats = entries.map(name => {
+      const st = statSync(join(dir, name));
+      return { name, mtime: st.mtimeMs };
+    });
+  } catch (e) {
+    console.warn(`[images] prune: could not stat files in ${dir}:`, e instanceof Error ? e.message : e);
+    return;
+  }
+
+  // Sort oldest-first, delete the excess.
+  stats.sort((a, b) => a.mtime - b.mtime);
+  const toDelete = stats.slice(0, stats.length - maxFiles);
+  for (const { name } of toDelete) {
+    try {
+      unlinkSync(join(dir, name));
+    } catch (e) {
+      console.warn(`[images] prune: could not delete ${name}:`, e instanceof Error ? e.message : e);
+    }
+  }
 }
 
 function timestampPrefix(): string {
@@ -50,6 +93,7 @@ export function guessExtFromMagic(bytes: Uint8Array): string {
 export async function materializeInlineImage(
   base64Data: string,
   budget?: ImageBudget,
+  keepCount?: number,
 ): Promise<string> {
   const dir = getArtifactsDir();
   await mkdir(dir, { recursive: true, mode: 0o700 });
@@ -75,6 +119,7 @@ export async function materializeInlineImage(
   const ext = guessExtFromMagic(buf);
   const filePath = join(dir, `img-${timestampPrefix()}-${crypto.randomUUID()}.${ext}`);
   await writeFile(filePath, buf, { mode: 0o600 });
+  pruneOldArtifacts(dir, keepCount ?? DEFAULT_ARTIFACT_KEEP_COUNT);
   return filePath;
 }
 
@@ -82,11 +127,12 @@ export async function downloadImageToArtifact(
   url: string,
   budget?: ImageBudget,
   signal?: AbortSignal,
+  keepCount?: number,
 ): Promise<string> {
   if (url.startsWith("data:")) {
     const m = /^data:([^;]+);base64,(.+)$/.exec(url);
     if (!m) throw new Error("data URL is not a valid base64 image");
-    return materializeInlineImage(m[2], budget);
+    return materializeInlineImage(m[2], budget, keepCount);
   }
 
   // SSRF protection: validate the provider-returned URL before fetching.
@@ -142,5 +188,6 @@ export async function downloadImageToArtifact(
 
   const filePath = join(dir, `dl-${timestampPrefix()}-${crypto.randomUUID()}.${ext}`);
   await writeFile(filePath, bytes, { mode: 0o600 });
+  pruneOldArtifacts(dir, keepCount ?? DEFAULT_ARTIFACT_KEEP_COUNT);
   return filePath;
 }
--- a/src/images/artifacts.ts
+++ b/src/images/artifacts.ts
@@ -34,6 +34,8 @@ function getArtifactsDir(): string {
  * of files. All errors are swallowed and logged so a prune failure never breaks an image write.
  */
 export function pruneOldArtifacts(dir: string, maxFiles: number): void {
+  // A non-positive maxFiles disables pruning entirely (do not delete everything).
+  if (maxFiles <= 0) return;
   let entries: string[];
   try {
     entries = readdirSync(dir);
--- a/src/images/artifacts.ts
+++ b/src/images/artifacts.ts
@@ -85,7 +85,7 @@ function timestampPrefix(): string {
 
 export function guessExtFromMagic(bytes: Uint8Array): string {
   const sig = Buffer.from(bytes.slice(0, 12)).toString("latin1");
-  if (sig.startsWith("\x89PNG")) return "png";
+  if (sig.startsWith("\x89PNG\r\n\x1a\n")) return "png";
   if (sig.startsWith("\xff\xd8\xff")) return "jpg";
   if (sig.startsWith("RIFF") && sig.slice(8, 12) === "WEBP") return "webp";
   if (sig.startsWith("GIF8")) return "gif";
--- a/src/images/fulfill.ts
+++ b/src/images/fulfill.ts
@@ -48,9 +48,9 @@ export async function fulfillImageCall(
   for (const img of result.images ?? []) {
     try {
       if (img.b64_json) {
-        files.push(await materializeInlineImage(img.b64_json, budget));
+        files.push(await materializeInlineImage(img.b64_json, budget, plan.artifactsKeepCount));
       } else if (img.url) {
-        files.push(await downloadImageToArtifact(img.url, budget, signal));
+        files.push(await downloadImageToArtifact(img.url, budget, signal, plan.artifactsKeepCount));
       }
     } catch {
       // Keep warnings URL-free — error messages may embed provider CDN URLs.
--- a/src/images/fulfill.ts
+++ b/src/images/fulfill.ts
@@ -1,0 +1,75 @@
+import type { ImageBridgePlan, ImageCallResult } from "./types";
+import { callXaiImages } from "./xai-client";
+import { materializeInlineImage, downloadImageToArtifact, type ImageBudget } from "./artifacts";
+
+/**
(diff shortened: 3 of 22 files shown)


lidge-jun:

lidge-jun:
Thank you for building out the Grok image bridge. There are two merge-blocking issues in the current implementation.

First, `src/images/artifacts.ts:83` fetches a provider-returned URL without validating its protocol, hostname, resolved addresses, or redirects. A malicious or compromised upstream can therefore make the local proxy request loopback, link-local, RFC1918, or other internal services (SSRF). Please require the intended HTTPS origin(s), run the existing destination/private-network policy before every request, revalidate every redirect target, and add negative tests for loopback/private/link-local hosts plus a redirect into a private address. Do not log the returned URL if it can contain credentials.

Second, `src/server/responses/core.ts:1308` and `src/server/responses/core.ts:1449` return through `adapter.runTurn` before the new image-plan branch executes. Cursor and any other runTurn adapter therefore never reach this bridge even when the request contains the advertised image-generation plan. Please integrate the bridge into the runTurn path or move the plan decision ahead of the early return, then add an end-to-end regression using a runTurn adapter that proves the bridge executes and its artifact is returned.

The outbound fetch is a network security boundary, so this PR also needs explicit maintainer security review after the fixes. I appreciate the substantial work here; these two paths need to be closed before we can safely expose it.

Please also rebase onto current `dev`, which now includes #436's validated Responses parser contract in `src/responses/parser.ts` and #439's context-pressure types in `src/types.ts`; preserve both while re-cutting the image bridge.

---

lidge-jun:
Follow-up review at head `99fa843f`, against `dev` after today's integration batch. `git merge-tree` is clean, and you've addressed the earlier findings — the missing `additional_tools` detection, the `runTurn` bypass, OAuth refresh, and the basic SSRF checks are all handled now. Two release blockers remain.

**The paid bridge defaults to on.** `src/images/plan.ts:39` reads:

```ts
if (config.images?.bridgeEnabled === false) return undefined;
```

so merely having xAI configured is enough to route prompts into a paid generation path. A cross-provider feature that spends money on the operator's account should require them to say yes:

```ts
if (config.images?.bridgeEnabled !== true) return undefined;
```

Update `src/types.ts` to document the default as `false`, and correct `docs-site/src/content/docs/guides/image-bridge.md`, which currently describes the opposite. Existing positive plan tests will need `{ bridgeEnabled: true }` passed explicitly, and a new negative test should assert that a fully configured xAI provider still yields no plan without the opt-in.

**The credential destination is not pinned.** `plan.ts` returns `auth: { baseUrl: xai.baseUrl.replace(/\/+$/, ""), token }` straight from persisted config, bypassing registry routing. A raw `providers.xai.baseUrl` override can therefore receive the built-in OAuth token. Separately, a custom-named provider pointing at `api.x.ai` falls through to `getValidAccessToken("xai")` and borrows the built-in account's token, which it should not.

Resolve the provider through `routeModel()` and make token resolution name-aware: an explicit `apiKey` on any provider is fine, but the built-in OAuth token should only be handed to the provider actually named `xai`. A regression should assert that with `providers.xai.baseUrl` set to a sink host, `plan.auth.baseUrl` is still `https://api.x.ai/v1`.

**Status:** held for explicit security review under `MAINTAINERS.md` — this consumes xAI OAuth or API keys and initiates paid, non-idempotent generation. Both fixes above should land before that review.

One coordination note: #355 also adds `src/images/artifacts.ts`, and pairwise merge shows a direct add/add conflict. Your version is the functional superset, but whichever lands second should extend the reviewed module rather than replace it wholesale.

---

<sub>Reviewed as part of a maintainer PABCD pass. Adds findings not present in the earlier pass.</sub>

Wibias (changes requested):
## Verdict
**Request changes** ÔÇö tip `b3f2b492` fixes the earlier maintainer blockers (opt-in `bridgeEnabled`, registry-pinned xAI baseUrl, name-aware OAuth, `runTurn` placement, basic private-destination SSRF), but this is still not merge-ready: HTTPS scheme enforcement is incomplete, docs/priority contradict the dispatch order, the branch is 331 commits behind `dev`, and required CI has not actually run (`action_required` on the fork). Optional follow-ups are **not** accepted; all items below must ship in this PR before merge.

## Summary
Base is correctly `dev` (`feat/image-bridge` ÔåÆ `dev`). Feature direction matches the PR body (synthetic `image_gen` ÔåÆ xAI Grok Imagine ÔåÆ `~/.opencodex/artifacts/` ÔåÆ tool result ÔåÆ bounded loop, mirrored on the web-search sidecar shape). GitHub reports **`mergeable: true`**, **`mergeStateStatus: UNSTABLE`** (not `CONFLICTING` / `DIRTY`). Compare vs current `dev`: **`ahead_by: 1`**, **`behind_by: 331`**, status `diverged`. Cross-platform CI + React Doctor concluded `action_required` (workflows not executed). Explicit maintainer security review under `MAINTAINERS.md` is still required after the remaining security fixes.

## Blocking findings

1. **Image URL SSRF still allows non-HTTPS / non-HTTP(S) schemes** ÔÇö `src/images/artifacts.ts` ~L92ÔÇô100
   - **Failure mode:** The comment says ÔÇ£Reject non-HTTP(S) schemesÔÇØ, but `assessUrlDestination` / `assertUrlResolvesPublic` only classify host/IP class ÔÇö they never inspect `url.protocol`. A provider-returned `ftp://ÔÇª`, `gopher://ÔÇª`, or cleartext `http://ÔÇª` URL that resolves to a public address passes the checks and is fetched. Maintainer review (2026-07-25) explicitly asked to require HTTPS origins before every request.
   - **Fix:** Before `assessUrlDestination`, reject unless `new URL(url).protocol === "https:"` (keep the existing `data:` branch). Add negative tests for `http://`, `ftp://`, and a public-looking non-HTTPS URL in `tests/images/artifacts-ssrf.test.ts`, preferably via `downloadImageToArtifact` (not only the helper).

2. **Docs claim web-search priority; dispatch does the opposite** ÔÇö `src/server/responses/core.ts` ~L1328ÔÇô1356 vs ~L1457ÔÇô1504; `docs-site/src/content/docs/guides/image-bridge.md` ~L63ÔÇô64
   - **Failure mode:** Guide: ÔÇ£If both web search and image generation are requested in the same turn, the web-search bridge runs and image generation is skipped.ÔÇØ Code evaluates `planImageBridge` and returns the image bridge **before** `planWebSearch`, so a dual-tool turn never reaches web-search when the image bridge is armed.
   - **Fix:** Before taking the image path, skip when a web-search plan would activate (same inputs `planWebSearch` uses), **or** move the image block to sit with the web-search branch with web-search first ÔÇö while preserving the image check **before** the `adapter.runTurn` early-return. Add a focused regression that dual-tool + both bridges eligible ÔåÆ web-search path wins / image bridge does not.

3. **Branch is stale vs current `dev` (required rebase in this PR)** ÔÇö GitHub compare `dev...b3f2b492` ÔåÆ `behind_by: 331`
   - **Failure mode:** Not `CONFLICTING` today, but the tip cannot be honestly CIÔÇÖd or security-reviewed against current integration. Maintainer already required a rebase after the Responses parser / types landings. Open sibling #355 also adds `src/images/artifacts.ts` (add/add); landing order without a rebase will break whichever merges second.
   - **Fix:** Rebase (or merge) latest `dev` into this PR and resolve here. If #355ÔÇÖs `artifacts.ts` is present, **extend** the reviewed module ÔÇö do not wholesale replace it.

## Additional required quality changes (fix in this PR before merge)

1. **SSRF tests do not cover the download entrypoint / redirect posture** ÔÇö `tests/images/artifacts-ssrf.test.ts`
   - Helpers are covered; `downloadImageToArtifact` is not. Add cases that call the download helper with mocked `fetch`: private/literal rejection, HTTPS-public allow, non-HTTPS reject, and `redirect: "error"` behavior (or document + test that redirects fail closed).

2. **Missing handler-level activation coverage for the new dispatch branch** ÔÇö `src/server/responses/core.ts` ~L1328ÔÇô1356
   - Module tests for plan/loop/fulfill exist, but there is no focused regression that the Responses handler activates the bridge (stream=true success path, stream=false ÔåÆ 400, and runTurn adapter still enters the bridge). AGENTS.md requires a focused regression near the subsystem for `src/` behavior changes.

3. **Synthetic tool schema / docs drift on edit inputs** ÔÇö `src/images/synthetic-tool.ts` ~L54ÔÇô69 vs `src/images/fulfill.ts` ~L34ÔÇô41 / `xai-client.ts` ~L81ÔÇô83
   - `buildImageTool()` advertises generate-or-edit in the description but only exposes `prompt`/`n`. `fulfillImageCall` still accepts `image_url`/`image` for `/images/edits`. Either expose the edit fields in the schema (and tests) or narrow the description to generation-only so models are not told they can edit without a parameter.

4. **Do not log provider image URLs** ÔÇö `src/images/fulfill.ts` ~L55ÔÇô57
   - `console.warn` includes the materialization error message; if a future/path error embeds the URL (or a URL contains credentials), that violates the privacy rule. Keep warnings URL-free (status/code only).

## Optional / non-blocking
**None.** Follow-up PRs are not accepted for the items above.

## Standards / Spec / Security

### Standards
- Hard: privacy/security boundary on outbound fetch + credential handling (`AGENTS.md` / `MAINTAINERS.md`) ÔÇö scheme gap above.
- Hard: tests for new `src/` dispatch behavior ÔÇö missing core activation coverage.
- Hard: docs must not contradict runtime (`docs-site/` sync) ÔÇö web-search priority.
- Judgement: duplicated bounded-body readers vs `readBoundedResponseBody` in `xai-client.ts` ÔÇö acceptable for now if size caps stay correct; do not open a follow-up PR, but prefer reuse if you touch that file while fixing.

### Spec
- Earlier maintainer blockers at head `99fa843f` (**paid default on**, **credential destination not pinned**) are **addressed** at tip `b3f2b492` (`plan.ts` `bridgeEnabled !== true`, registry pin + name-aware OAuth, tests in `plan.test.ts`). Dismissed as fixed.
- Remaining open maintainer requirements: rebase onto current `dev`, complete HTTPS/destination policy, then explicit security review.
- PR-claimed ÔÇ£mirrors web-search sidecarÔÇØ is only partial: image path is before `runTurn` (good), but priority vs web-search and some lifecycle/telemetry parity still differ (see blocking #2).

### Security
- Touches OAuth/API-key spend path + untrusted URL fetch ÔåÆ **explicit maintainer security review required** after the HTTPS scheme fix and rebase (`MAINTAINERS.md`). Do not merge on CodeRabbit alone.
- Positive: `redirect: "error"`, DNS fail-closed in `assertUrlResolvesPublic`, byte caps, registry baseUrl pin, OAuth scoped to provider name `"xai"`.

## CI / process gates
| Gate | Status |
| --- | --- |
| Base branch | `dev` (correct) |
| `mergeable` / `mergeStateStatus` | `true` / `UNSTABLE` (not CONFLICTING) |
| Divergence | ahead 1 / **behind 331** vs `dev` ÔÇö rebase required in this PR |
| Cross-platform CI | `action_required` (did not run on fork head) ÔÇö must be green after rebase/approval |
| React Doctor | `action_required` |
| PR Labeler | success |
| CodeRabbit | pass (informational) |
| Maintainer security review | still required after fixes |

## Conflict status
**Clean mergeability against current `dev`:** `mergeable=true`, `mergeStateStatus=UNSTABLE`. No conflicted file list from GitHub. Rebase is still mandatory because of 331-commit drift and the pending `#355` `src/images/artifacts.ts` add/add collision.

### After fixes
Once **all** blocking + required quality changes above are addressed (and conflicts/rebase done), **mark this PR Ready for review** (leave draft only until then). Then ping for re-review / CI.

Wibias (changes requested):
Double-Posted.

Wibias (changes requested):
## Re-review @ `eb7720ea`

Follow-up to the CHANGES_REQUESTED review on `b3f2b492`. Checked fix commit `fix(images): address Wibias review blockers for #424`. Cross-checked against the post-fix Codex review on the same tip.

**Verdict: still request changes** — most earlier blockers are fixed, but dispatch still has merge-blocking gaps (`runTurn` dual-tool, routed compaction), SSRF coverage/policy is incomplete (including IPv4-mapped IPv6), and required CI has not actually run.

### Prior blockers

| # | Finding | Status |
| --- | --- | --- |
| 1 | HTTPS-only scheme in `downloadImageToArtifact` | **Fixed** — `protocol !== "https:"` before assess/DNS; `data:` kept; `redirect: "error"` still set (`src/images/artifacts.ts`) |
| 2 | Docs vs dispatch: web-search priority | **Fixed for non-`runTurn`** — image path consults `planWebSearch` and only runs when `!wsPlan` (`src/server/responses/core.ts`). Docs already match. See remaining gap below. |
| 3 | Rebase onto current `dev` | **Mostly fixed** — was behind 331; now ahead 2 / behind 4 (docs-only tip on `dev`). Still slightly stale; not the old drift. |
| Q1 | SSRF tests via `downloadImageToArtifact` | **Partial** — `http://` / `ftp://` reject + HTTPS success added. Still missing: private/literal via download helper, `gopher://`, and `redirect: "error"` behavior. |
| Q2 | Handler activation coverage | **Fixed** — `tests/images/handler-activation.test.ts` covers stream SSE, stream=false → 400, dual-tool defer (non-`runTurn` stub). |
| Q3 | Synthetic tool edit-schema drift | **Fixed** (allowed option) — description narrowed to generation-only; no edit params advertised. |
| Q4 | No URL logging in fulfill warn | **Fixed** — `console.warn("[images] failed to materialize image")` only. |

Earlier maintainer items (opt-in `bridgeEnabled`, registry pin, name-aware OAuth, `runTurn` placement) remain intact.

### Remaining issues

1. **Dual-tool + `runTurn` → neither bridge** — `src/server/responses/core.ts` (~image block through `adapter.runTurn` return)
   - **Failure mode:** When both bridges are eligible, image correctly defers, then `if (adapter.runTurn)` returns before the web-search block. For Cursor/`runTurn` adapters, dual-tool turns get **neither** bridge (regression vs pre-fix, where image won). The new dual-tool test stubs an adapter **without** `runTurn`, so it misses this.
   - **Codex:** independently reproduced the same failure mode on this tip (`core.ts` ~L1520).
   - **Fix:** When `imgPlan && wsPlan`, take the web-search path immediately (before `runTurn`), or skip the `runTurn` early-return when a web-search plan is active. Add a regression with a `runTurn` adapter.

2. **Image bridge can hijack routed compaction** — `src/server/responses/core.ts` ~L1516
   - **Failure mode (Codex P1):** A routed `compaction_trigger` clears tools/`_webSearch` but can leave `_imageGeneration` set. `planImageBridge` then activates and returns a normal Responses completion instead of the required synthetic `compaction` item, so Codex rejects the compaction result.
   - **Fix:** Guard image dispatch with `!routedCompaction`, or clear the image-generation stash alongside `_webSearch`. Add a focused regression.

3. **SSRF download coverage / IPv4-mapped IPv6 bypass still incomplete**
   - `tests/images/artifacts-ssrf.test.ts`: still need `downloadImageToArtifact` cases for private/literal rejection and fail-closed redirects (`redirect: "error"`).
   - `src/lib/destination-policy.ts` ~L80–81 (**Codex P1**): `https://[::ffff:127.0.0.1]/…` normalizes to `::ffff:7f00:1`, misses the dotted-decimal IPv4-mapped check, is classified public, skips DNS as an IP literal, and can reach loopback — bypassing the image-download SSRF guard. Decode IPv4-mapped IPv6 (hex and dotted) and classify the embedded IPv4; cover with a download-path regression.

4. **Process / CI**
   - Cross-platform CI still has not run on this fork head (only CodeRabbit + enforce-target + label).
   - Maintainer security review still required after the remaining items.
   - Behind-by-4 is minor; rebase optional but preferred before final CI.

### Spec / security

- Spec direction still matches the PR body; priority claim is only fully true for non-`runTurn` paths.
- Security: HTTPS scheme gate looks correct, but destination-policy still needs the IPv4-mapped IPv6 fix before claiming SSRF closed. Keep explicit security review after remaining items 1–3.
- Other Codex P2s on this tip are noted but **not** expanded into required blockers for this re-review pass:
  - **Forced-final leaves aliased image tools** (`src/images/loop.ts` ~L162): `toolsNoImage` filters only `t.imageGeneration`, so activation via `image_generation` / `image_gen` / other aliases leaves the original callable tool on the forced-final pass; the model can call it again, `scanEventsForImageCall` strips it while `forceFinal` blocks fulfillment, and the client gets an empty completion. Filter all names in `plan.toolNames` (or mark/replace matching tools at activation).
  - **Bound / validate `images.maxRounds`** before `HARD_CAP` (hand-edited `10000` / fractional values → unbounded paid xAI calls).
  - **Honor `images.timeoutMs`** in `callXaiImages` instead of the hard-coded 60s abort (`src/images/xai-client.ts`).
  - **Return SSE headers before `queue.collect()`** on the Cursor/`runTurn` path so slow first turns do not starve clients/proxies of headers/heartbeats (`src/images/loop.ts` ~L208).
  - **Preserve Cursor `_cursorConversationId`** across image-loop iterations when the outer request has no `previous_response_id` / parent thread id.
  - **Wire `onUsage` / `logCtx`** for image-bridge completions the same way as the normal and web-search bridges.
  - **Rotate pooled keys on 429** inside the image loop (`rotateProviderTransportOn429`) instead of treating the first 429 as terminal.
  - **Keep parallel image calls in one assistant turn** (with thinking attached once) so Anthropic extended-thinking continuations stay valid.

### After fixes
Address remaining items 1–3 (and preferably rebase + green Cross-platform CI), then ping for another pass.

Wibias (approved):
## Re-review @ `6d6b252`

Follow-up to the CHANGES_REQUESTED re-review on `eb7720ea`. Checked tip `fix(images): address Wibias R2 review blockers for #424`.

### Verdict
**LGTM on substance — CI + security pending.**

All four R2 code blockers are fixed with matching regressions. Approve on code; do not merge until Cross-platform CI is green on this tip and explicit maintainer security review under `MAINTAINERS.md` is recorded.

### R2 remaining items

| # | Finding | Status |
| --- | --- | --- |
| 1 | Dual-tool + `runTurn` → neither bridge | **Fixed** — web-search dispatch moved before `adapter.runTurn` (`src/server/responses/core.ts`); regression with `useRunTurnAdapter` asserts web-search wins and `runTurn` is not called |
| 2 | Image bridge hijacks routed compaction | **Fixed** — `imgPlan = !routedCompaction ? planImageBridge(...) : undefined`; regression with `compaction_trigger` + `image_generation` |
| 3 | IPv4-mapped IPv6 hex SSRF (`::ffff:7f00:1`) | **Fixed** — hex decoder in `destination-policy.ts` after dotted form; assess + `downloadImageToArtifact` coverage |
| 4 | SSRF tests (private download, gopher, redirect fail-closed) | **Fixed** — all three in `tests/images/artifacts-ssrf.test.ts` |

Earlier blockers (HTTPS-only scheme, docs vs web-search priority, handler activation, edit-schema, no URL logging, opt-in bridge, registry pin, name-aware OAuth, `runTurn` placement) remain intact.

### Process gates (still required before merge)

1. **Cross-platform CI** must actually run and pass on this tip (fork currently shows only enforce-target / label / CodeRabbit — suite not executed).
2. **Maintainer security review** per `MAINTAINERS.md` after the SSRF/destination-policy changes.
3. **Rebase onto current `dev` preferred** before final CI (tip has drifted further behind since R2).

### Optional follow-ups (not merge blockers for this pass)

Same Codex P2 notes as R2 — not expanded into required blockers here: `images.maxRounds` bounds, honor `images.timeoutMs`, forced-final alias filter, SSE headers before `queue.collect()` on runTurn, Cursor `_cursorConversationId` across image-loop iterations, image-bridge `onUsage`/`logCtx` wiring, 429 key rotation in the image loop, parallel image calls in one assistant turn.

Wibias:
Follow-up for the Codex P2 notes from the R2 approval: #528 (`fix/image-bridge-p2`, tip `35ac19d0`).

Covers maxRounds clamp, `images.timeoutMs`, forced-final alias filter, SSE-before-collect on runTurn, Cursor conversation id continuity, `onUsage`/`logCtx`, 429 key rotation, and parallel image calls in one assistant turn.

Wibias (commented):
## Security review @ `6d6b252`

Maintainer-grade security pass on the Grok image bridge (`src/images/*` + `handleResponses` wiring), per `MAINTAINERS.md` (auth/credentials, SSRF, OAuth, secrets, logging/privacy, egress). Tip only — does **not** assume #528.

### Decision
**Pass** — no medium+ exploitable findings on realistic attack paths. Image bridge is mergeable from a security-boundary perspective once Cross-platform CI is green.

**Risk level:** Low

### Controls verified

| Area | Result |
| --- | --- |
| Opt-in | `bridgeEnabled === true` required (`plan.ts`); default off |
| xAI auth | Bearer from `resolveXaiToken`; built-in OAuth only when provider name is exactly `xai`; custom-named hosts need API key |
| Registry pin | Sidecar `baseUrl` from `getProviderRegistryEntry("xai")`, ignores config override |
| SSRF (downloads) | HTTPS-only, `redirect: "error"`, `assessUrlDestination` + `assertUrlResolvesPublic`, IPv4-mapped IPv6 hex (`::ffff:7f00:1`) decoded, 50 MiB stream cap (`artifacts.ts` + `destination-policy.ts`) |
| Artifacts | `~/.opencodex/artifacts`, `0o700` / `0o600`, UUID filenames — no user path segments |
| Logging | Materialization warn is URL-free; xAI errors expose status only |
| Admission token | Sidecar uses `plan.auth`, not client `Authorization` / proxy admission secret |
| Data plane | No new unauthenticated management API; activation still requires Responses auth when non-loopback |

### Confirmed findings

None at medium / high / critical.

### Residual (below medium — non-blocking)

1. **[Low] Unbounded `images.maxRounds` on this tip** — `Math.max(0, maxRounds)` has no upper clamp (`loop.ts`). A hand-edited `maxRounds: 10000` with `bridgeEnabled` can drive unbounded paid xAI calls against the **operator’s own** key (self-scoped billing, not cross-tenant). Addressed in follow-up #528 (`clampImageMaxRounds` → `[0,10]`). Prefer landing #528 soon after merge, or cherry-picking the clamp before merge if you want zero residual.

2. **[Low] DNS rebinding TOCTOU** — resolve-then-fetch without connection pinning (same class already documented for provider destinations). Defense-in-depth only.

3. **[Low] Edit `image_url` forwarded to xAI** — opencodex does not locally SSRF-check edit source URLs; fetch is on xAI’s side. Not an opencodex LAN SSRF.

4. **[Info] Parallel `image_gen` calls per iteration** — `maxRounds` bounds iterations, not parallel fulfills in one turn (`n` still ≤ 4 each). Operator cost blast radius; #528 also folds parallel calls into one assistant turn (correctness), not a hard per-call cap.

### Out of scope / not re-litigated
- Functional R2 items already cleared on this tip (runTurn dual-tool, compaction guard, SSRF tests).
- CI FTC / rebase process gates remain operational, not security defects.

### Verification
- Reviewed tip `6d6b252` (PR #424 head), not #528.
- Inspected `plan.ts`, `artifacts.ts`, `fulfill.ts`, `xai-client.ts`, `loop.ts`, `destination-policy.ts`, `core.ts` image dispatch.

tizerluo:
## Review thread status update (HEAD: `511a0924`)

Addressed all blocking + quality items from Wibias R2 review and the artifact retention gap. Summary of threads:

**Fixed on current tip** (marking as resolved):
- SSRF HTTPS scheme + IPv4-mapped IPv6 hex bypass (`destination-policy.ts` hex decode + classify)
- SSRF download tests: private IP, `gopher://`, IPv6 mapped (dotted+hex), `redirect: "error"` fail-closed
- Dispatch order: web-search moved before `runTurn` so dual-tool turns reach web-search on all adapters
- Compaction guard: `!routedCompaction` prevents image bridge from hijacking compaction
- xAI OAuth pinned to canonical `"xai"` name only; baseUrl from registry, not config override
- Forwarded auth headers cloned per iteration; `onAttemptSend` wired on both paths
- `maxRounds` bounded before dispatch; aliased image tools removed from dispatch
- Bounded streaming reader with `finally{cancel()}` for xAI + download paths
- Artifact retention: `pruneOldArtifacts()` caps `artifacts/` to 200 files (configurable via `images.artifactsKeepCount`)
- Download signal passes through to `fetch()` for abort support

**Acknowledged, intentionally deferred:**
- Cursor conversation IDs: managed by the adapter (runTurn owns session state); the bridge forwards events transparently.

**Tracked as future improvements (not blocking):**
- xAI key rotation / 429 failover inside image loop (single OAuth credential is the common case)
- Usage/cost telemetry from xAI sidecar (xAI `/images` may not return token counts)
- Hosted image options (size/quality) forwarding to synthetic tool
- Aggregate cap on total paid image calls per turn
- Configurable routed connection timeout

Will resolve individual threads as "outdated" on the files below. Happy to address any of the deferred items in a follow-up if requested.

Wibias (changes requested):
## Verdict
**Request changes** — tip `511a092` (artifact retention) reopens merge readiness after the prior LGTM on `6d6b252`. R2 code blockers remain fixed and the security pass on the bridge still stands, but this tip breaks Cross-platform CI on Windows, leaves `artifactsKeepCount ≤ 0` as a wipe footgun, and the branch is again stale vs `dev` (`behind_by: 38`). Optional follow-ups are **not** accepted; items below are required in this PR.

## Summary (since last review)
| Tip | Event |
| --- | --- |
| `6d6b252` | Prior Wibias re-review: **APPROVED** on substance; CI + security pending |
| `6d6b252` | Maintainer security review: **Pass** (Low residual: unbounded `maxRounds` → #528) |
| `511a092` | New: `pruneOldArtifacts` + `images.artifactsKeepCount` (default 200), wired through plan/fulfill, docs + `tests/images/artifacts-prune.test.ts` |

Base remains `dev`. GitHub: `mergeable=true`, `mergeStateStatus=UNSTABLE`. Compare `dev...511a092`: **ahead 4 / behind 38**, status `diverged`. Cross-platform CI: ubuntu/macos/npm-global green; **`windows-latest` FAILED** (1 fail in the new prune integration test).

## Prior findings disposition

| Source | Finding | Status @ `511a092` |
| --- | --- | --- |
| R2 #1 | Dual-tool + `runTurn` → neither bridge | **Still fixed** — web-search before `runTurn`; handler regression with `useRunTurnAdapter` |
| R2 #2 | Image bridge hijacks routed compaction | **Still fixed** — `!routedCompaction` guard + regression |
| R2 #3 | IPv4-mapped IPv6 hex SSRF (`::ffff:7f00:1`) | **Still fixed** — destination-policy + download-path tests |
| R2 #4 | SSRF download coverage (private/gopher/redirect) | **Still fixed** — `artifacts-ssrf.test.ts` green on Windows too |
| Earlier | HTTPS-only scheme, docs/web-search priority, handler activation, edit-schema, URL-free warn, opt-in bridge, registry pin, name-aware OAuth | **Intact** |
| Security @ `6d6b252` | Pass / Low residual `maxRounds` | **Not regressed by retention**; residual still tracked via #528 — not reopened here |
| Codex/CR (stale) | SSRF “no host/scheme validation”, compaction bypass, pre-`runTurn` placement, non-stream SSE, unforwarded `maxRounds`, missing `trackStreamLifetime` | **Dismiss — fixed on tip** (see `core.ts` ~L1513–1544, `artifacts.ts` HTTPS + assess/DNS) |
| Codex/CR (still open P2s) | forced-final alias filter, honor `timeoutMs`, Cursor conversation id, `onUsage`/`logCtx`, 429 key rotation, parallel image calls, DNS bind, etc. | **Still open but not newly required by this tip** — previously deferred under LGTM/security pass; do not treat as Optional follow-up PRs either (leave for a dedicated hardening PR only if maintainers ask). **Not** merge blockers for this re-review. |

## Blocking / required quality

1. **Windows CI red: prune integration test uses POSIX-only path split** — `tests/images/artifacts-prune.test.ts` ~L74–79
   - **Failure mode (confirmed on `windows-latest` job 89873289631):** `written[0].lastIndexOf("/")` is `-1` on `C:\…\artifacts\img-….png`. `substring(0, -1)` yields `""`, so `readdirSync("")` lists the process CWD (**Expected: 3, Received: 30**). Unit prune tests pass; only this integration assertion is wrong — but it fails required Cross-platform CI.
   - **Fix:** Use `dirname` / `basename` from `node:path` (or split on both separators). Re-run `bun test tests/images/artifacts-prune.test.ts` on Windows or wait for green `windows-latest`.

2. **`artifactsKeepCount ≤ 0` deletes the entire artifacts directory** — `src/images/artifacts.ts` `pruneOldArtifacts` ~L36–66; config `OcxImagesConfig.artifactsKeepCount`
   - **Failure mode:** Early exit is `entries.length <= maxFiles`. For `maxFiles === 0` (or negative), that never trips when files exist, then `stats.slice(0, stats.length - maxFiles)` deletes **all** entries — including the file just written. A hand-edited `artifactsKeepCount: 0` therefore wipes `~/.opencodex/artifacts/` after every successful image write.
   - **Fix:** Clamp at the plan/config boundary (e.g. treat undefined → 200; clamp to `[1, …]` or document and implement `0` as “disable prune” without deleting). Add a unit test for `0` / negative. Do not rely on docs alone.

3. **Branch stale vs current `dev` (rebase required in this PR)** — compare `dev...511a092` → `behind_by: 38`
   - **Failure mode:** Not `CONFLICTING` today, but tip cannot be honestly final-CI’d / reviewed against current integration. Prior R2 already warned against letting this drift.
   - **Fix:** Rebase (or merge) latest `dev` into this PR and resolve here.

## Optional
**None.**

## CI / security / conflicts

| Gate | Status |
| --- | --- |
| Base | `dev` (correct) |
| `mergeable` / `mergeStateStatus` | `true` / `UNSTABLE` |
| Divergence | ahead 4 / **behind 38** — rebase required |
| Cross-platform CI | **windows-latest FAIL** (prune integration); ubuntu/macos/npm-global pass |
| React Doctor | pass |
| Maintainer security (bridge @ `6d6b252`) | Pass recorded; retention is local file-count prune only — low new attack surface, but **re-check after keepCount clamp** (wipe-on-zero is a self-DoS / data-loss footgun, not SSRF) |
| Conflicts | Clean mergeability vs `dev` (`mergeable=true`); rebase still mandatory for drift |

## After fixes
Address blocking items 1–3, get **full** Cross-platform CI green on the new tip (especially `windows-latest`), then **mark Ready for review** and ping for another pass.

Wibias:
## Re-review vs stacked #528
Compared **#424** tip `511a092` with **#528** tip `d65effdf` (unique delta only; merge-base `6d6b252`). #528 is stacked on the pre-retention tip and is currently **diverged** from this PR (missing the `511a092` prune commit).
### Covered by [#528](https://github.com/lidge-jun/opencodex/pull/528)
Relative to the open hardening / Codex P2 set: forced-final strips every `plan.toolNames` alias; named + `allowed_tools` image `tool_choice` rewrite to synthetic `image_gen`; Cursor `_cursorConversationId` continuity; SSE headers before first `runTurn` collect; `clampImageMaxRounds` `[0,10]`; per-turn paid-call cap (`MAX_IMAGE_CALLS_PER_TURN`) with per-call `n≤4`; terminal `onUsage` / `logCtx` wiring; Anthropic thinking attached once for parallel image calls; `authMode` honored for xAI token selection; CI on the #528 tip is green including **windows-latest**.
### Still open after #528 (concrete)
1. **Bridge still arms on ordinary function tools by name** — `src/images/synthetic-tool.ts` `extractHostedImageGeneration` 
 - **Failure:** A client-executed `{type:"function", name:"image_gen"|"generate_image"|…}` (no hosted `image_generation` / `image_gen` type) still sets `_imageGeneration`, so the bridge replaces/intercepts that tool and spends xAI instead of relaying a normal function call. 
 - **Fix:** Activate only on hosted types `image_generation` / `image_gen`. Do not match function tools by `isImageGenName` alone. Add a collision regression.
2. **Type-shaped image `tool_choice` still collapses to `auto`** — `src/responses/parser.ts` `mapToolChoice` 
 - **Failure:** `tool_choice: { type: "image_generation" }` (no `name`) returns `"auto"` before #528’s core rewrite runs, so a forced hosted-image choice is not constrained to synthetic `image_gen`. Named / `allowed_tools` paths are fixed; this type-only shape is not. 
 - **Fix:** Map hosted image tool-choice types to `{ name: "image_gen" }` (same idea as `allowedToolName`), then keep the core rewrite. Regression for type-only choice.
3. **OAuth image fulfillment still bypasses the resolved xAI transport** — `src/images/plan.ts` + `src/images/xai-client.ts` vs `src/providers/xai-transport.ts` 
 - **Failure:** `authMode: "oauth"` still arms the bridge with a stored OAuth bearer against the registry-pinned `https://api.x.ai/v1` via plain `fetch`, without the OAuth cli-chat base URL / headers / fetch wrapper. That is not the normal resolved xAI transport; calls can 401/fail while docs still advertise OAuth. 
 - **Fix:** Either fulfill images through the resolved OAuth transport if `/images/*` is actually supported there, **or** reject OAuth at plan time and document API-key-only for the bridge (and update `image-bridge.md` + plan tests).
### Also still blocking on **this** tip (`511a092`) — not in #528
4. **Windows CI** — `tests/images/artifacts-prune.test.ts` POSIX `/` dirname split → red `windows-latest` (job confirmed). Use `path.dirname` / `basename`. 
5. **`artifactsKeepCount ≤ 0` wipes `~/.opencodex/artifacts/`** after each write — clamp at the plan/config boundary (e.g. `[1, …]` or treat `0` as disable-prune without deleting).
### CI vs code
- **#528 tip:** Cross-platform CI green (code readiness for the items it claims is mostly there; remaining 1–3 above are still code defects). 
- **#424 tip:** Windows still red; rebase onto current `dev` still required (`behind_by` ~44).
### Merge order
1. Fix items **4–5** (and rebase) on **#424**, get full Cross-platform CI green. 
2. Merge **#424**. 
3. Rebase **#528** onto the merged #424 tip (resolve divergence / carry prune). 
4. Land remaining **1–3** on #528 (or a fast follow), then merge **#528**.
Do **not** merge #528 before #424, and do not merge this tip while Windows is red.

Wibias:
Re-reviewed PR #424 at `a8b769c` and stacked follow-up PR #528 at `553e9af`, including the latest Codex review threads on both PRs.

The latest #424 commit fixes the Windows artifact-prune path handling and makes `artifactsKeepCount <= 0` disable pruning. It does not change the remaining image-bridge runtime findings.

PR #528 already covers the earlier findings around Cursor conversation continuity, early SSE headers, stalled `runTurn` cancellation, forced-final alias removal, bounded `maxRounds`, cumulative paid-call limits, timeout plumbing, hidden-iteration usage, 429 rotation, and parallel-call transcript grouping.

After accounting for #528, these concrete blockers still remain:

1. **Ordinary client function tools can still be hijacked.** `extractHostedImageGeneration()` activates on regular `type: "function"` tools solely because their name matches `image_gen`, `generate_image`, etc. Core then removes or replaces that client tool and the loop can send its arguments to xAI. Activation must be limited to hosted `image_generation` / `image_gen` tool types.

2. **Codex's normal hosted-plus-client imagegen pair is not replaced consistently.** When hosted `image_generation` activates the bridge, an accompanying client `image_gen.imagegen`, namespace tool, or flattened routed name can remain visible but fall outside `plan.toolNames`. The routed model can call that original client tool as a real tool instead of invoking xAI. Replace the hosted/client pair with exactly one synthetic `image_gen` wire tool and remove every conflicting routed alias.

3. **Direct hosted `tool_choice` is still weakened.** A choice such as `{ "type": "image_generation" }` is parsed as `auto`, so core cannot rewrite it to a required synthetic `image_gen` call. `allowed_tools` is addressed in #528, but direct hosted choices still need required semantics.

4. **xAI OAuth image transport is still invalid.** The planner obtains an OAuth token but pins the request to `api.x.ai`, while `callXaiImages()` uses bare global `fetch` with only `Authorization`. Normal xAI OAuth uses the Grok CLI proxy plus compatibility headers and a wrapped fetch transport. Preserve that resolved transport, or reject OAuth bridging and document API-key-only support.

5. **Multiple Anthropic thinking blocks are still flattened.** The loop concatenates all thinking text and retains only the final signature. Multi-block signed and redacted thinking must be replayed in original order with each matching signature.

6. **Image-download SSRF protection remains vulnerable to DNS rebinding.** The code validates DNS once with `assertUrlResolvesPublic()` and then performs a separate `fetch(url)` resolution. A hostile hostname can return a public address during validation and a private or metadata address during the actual connection. Pin the validated address while preserving the TLS hostname, validate the connected peer, or enforce a trusted CDN allowlist at connection time.

7. **Cursor web search is routed through an unsupported transport.** The moved web-search dispatch now calls `runWithWebSearch` for Cursor/runTurn adapters, but that loop uses `buildRequest` / fetch / `parseStream`; Cursor requires `runTurn` and explicitly disables the fetch/parse path. Cursor web-search turns, including dual image/web-search requests, therefore fail instead of reaching Cursor's live transport. Add runTurn support to the web-search loop or keep runTurn-only adapters on their supported path.

8. **Per-write artifact pruning can return deleted paths.** If `artifactsKeepCount` is smaller than the number of images produced in one call, pruning after each individual write can delete an earlier image that `fulfillImageCall` still returns and may select as `primary`. Defer pruning until the full batch is materialized, then return only retained paths; cover the two-image/keep-one case.

PR #528 currently diverges from the updated #424 branch, so it must be rebased onto the final #424 head before its CI result is meaningful and before merge.

Current #424 exact-head workflows are still `action_required`, so no exact-head CI has executed for `a8b769c` yet.

Wibias:
Maintainer takeover: landing vehicle is now **#577** (Wibias:maintainer-takeover/424-feat-image-bridge).

It rebases this work onto latest `dev`, folds in stacked #528, and clears the remaining review blockers (hosted-only activation, type-only `tool_choice`, API-key-only xAI Images auth, ordered thinking replay, post-batch artifact prune, SSRF pin, docs).

Huge thanks to @tizerluo — you will be credited for the original feature work when this lands.

Wibias:
Landed via maintainer takeover **#577** (squash-merged to `dev`).

@tizerluo — thank you so much for this PR. The Grok image bridge is a big contribution, and you retain full credit for the original feature work. The takeover only folded review/security hardening (#528 + maintainer follow-ups) and rebased onto latest `dev` so we could ship it cleanly. Truly appreciate the effort.

PR #426, opened 2026-07-24
feat(codex): add account-qualified model namespaces

### LINKED ISSUE #425: Expose Codex accounts as model namespaces in the picker
### Area

Keep multiple Codex accounts authenticated at the same time and explicitly choose which exact account a new conversation uses from the model picker. The account choice should be visible in the selected model and remain fixed for that conversation, without logging out of any other account.

Account names should be arbitrary local IDs. They identify stored logins; they do not represent built-in roles or account types.

### What prevents this today?

The active Pool account is selected separately from the model. A bare model such as `gpt-5.6-sol` does not show which credential will be used, and normal Pool behavior may rotate or fail over to another eligible account.

Changing the global active account is therefore not a strict per-conversation choice. Logging out and back in is also not a viable substitute when other Codex sessions depend on the existing login.

### What should OpenCodex do?

Add an opt-in model-picker mode that exposes every supported native GPT model once per available Codex account.

- Selecting an account-qualified model pins the conversation to that exact stored account.
- Quota balancing, failure failover, affinity rebinding, and unsupported-model retry never switch that request to another account.
- If the selected account is missing, cooling down, or needs reauthentication, the request fails closed.
- Changing the active Pool account does not affect an account-qualified conversation.
- Bare `gpt-*` models retain their current Pool or Direct behavior.
- Hiding the account-specific picker entries does not turn an existing saved selector into a fallback route.
- Newly supported native GPT models automatically receive the same account-qualified choices.

With a built-in login and one added account whose local ID is `team-a`, the picker could expose:

```text
main/gpt-5.6-sol
team-a/gpt-5.6-sol
```

`main/gpt-5.6-sol` always uses the built-in login. `team-a/gpt-5.6-sol` always uses the stored account identified locally as `team-a`. If `team-a` is unavailable, the request fails instead of using `main` or another account.

IDs such as `team-a`, `client`, `personal`, or `work` are user-chosen examples only. OpenCodex should not assign any semantics to them.

- Switching the active Pool account is global and does not make the choice visible in the model selector.
- Disabling automatic balancing reduces switching but still does not create an explicit per-conversation account target.
- Logging out and back in disrupts other sessions and requires unnecessary credential churn.

### Additional context

This is generic explicit account targeting, not a request for special Personal and Work account roles. The existing picker and Pool/Direct behavior should remain unchanged unless the user opts in.

### Checks

chrisae9:
Closes #425.

## What this is for

OpenCodex can keep multiple Codex accounts authenticated, but a bare GPT entry does not say which account will serve the conversation. Pool mode may intentionally rotate or fail over, and Direct mode uses the caller/main login.

The missing workflow is: keep several accounts logged in, then explicitly choose one account for a conversation from the model picker without logging out or changing the active Pool account.

This is not a built-in “personal account / work account” feature. Those are possible user labels, nothing more. The feature works with arbitrary account records and user-owned selector names.

## Behavior

This is opt-in from **Codex Auth → Show each Codex account separately in the model picker**.

When it is off:

- the picker stays as it is today;
- bare models such as `gpt-5.6-sol` keep normal Pool or Direct behavior;
- existing qualified selections remain routable so saved conversations cannot silently fall through.

When it is on:

- each native GPT model appears once for each available Codex account;
- `main/<model>` selects the built-in Codex Desktop/main login;
- each added account gets a public selector from its explicit alias or a random privacy-safe `pXXXXXX` label;
- the public selector maps privately to the stored account ID; raw account IDs and emails are never copied into catalog slugs, provider log labels, or error text;
- selecting a qualified model pins that request and conversation to the exact account;
- quota balancing, transient failover, affinity rebinding, and unsupported-model account retry cannot switch it;
- missing credentials, cooldown, or reauthentication fail closed;
- the active Pool account is not changed.

Turning the setting off only hides the generated rows. It preserves custom selector mappings and saved qualified selections. Deleting an account hides its rows but retains the fail-closed binding; re-adding the same stored account restores it. Newly supported native GPT models receive the existing account selectors on the next catalog sync.

Qualified models still use the canonical `openai` Responses path, including HTTP/SSE, WebSocket, compact, native reasoning, tools, and service-tier metadata. Bare child/helper models keep Pool/Direct behavior; explicitly qualified helper models keep the exact account binding.

## Configuration and compatibility

The UI manages these fields:

```json
{
  "codexAccountPickerEnabled": true,
  "codexAccountNamespaces": {
    "main": "main",
    "side": "stored-account-id"
  }
}
```

The map keys are public selectors. The values are private local routing targets. Existing non-empty hand-written maps remain an explicit opt-in when the visibility override is omitted, and custom maps are preserved byte-for-byte across off/on toggles.

Fresh and ordinary existing configurations remain unchanged until the user enables the setting. Plain GPT IDs are not removed from routing, so existing threads and saved configuration retain their Pool/Direct semantics.

Account add/delete persists even if an immediate catalog refresh fails. OpenCodex retries once, reports a recoverable pending state, and shows an actionable `ocx sync` warning in both account-management surfaces.

## Integration work

This branch merges current `dev` and preserves the completed account lifecycle, router, quota-probe, catalog, and response contracts that landed there.

Review fixes include:

- privacy-safe public selectors that are independent of internal account IDs;
- canonical-OpenAI admission and strict Direct/Pool fail-closed behavior;
- reserved-ID, provider-prefix, and combo-alias collision handling;
- routed-catalog outage preservation and stale-slug de-duplication;
- account-aware multi-agent roster projection within the five-model client window;
- serialized GUI setting writes and complete settings hydration;
- recoverable catalog-refresh reporting for account creation and deletion;
- synchronized documentation and localized UI copy.

## Validation

- Full isolated Bun suite: **4,280 tests across 327 files, 0 failures**
- TypeScript typecheck
- GUI lint and production build
- GUI interaction and controller tests
- Privacy scan
- Documentation build: 126 pages
- Independent privacy, merge-contract, and final behavior reviews

## Security review

This changes credential-to-model routing at a C4 authentication boundary. It still requires the explicit maintainer security review described in `MAINTAINERS.md` before merge.


lidge-jun:

lidge-jun:
### Review deferred — head moved during the pass

Thanks for your patience. This PR was queued for a detailed review, but the head advanced from `37431320` to `2ff3e24b` (now 50 files, `+2111/-74`) while the pass was running. Posting a review written against the older diff would mean commenting on line references that have since moved, so it is being re-audited against the current head instead rather than shipped stale.

**One finding worth flagging now**, since it appears to survive the newer commits: `src/codex/account-namespaces.ts` derives the public namespace from `account.id` whenever an account has no alias, and that namespace then reaches catalog model slugs and provider log labels. An account imported manually or created through OAuth without an alias would expose its internal identifier through the model catalog and logs — which contradicts this PR's own stated contract that account ids stay server-side, and conflicts with the repository rule against logging account identifiers. Please generate a stable non-PII label instead and keep catalog/error/log serialization limited to it.

**Also note:** `dev` moved to `ebc62d1f` today. It now includes #370 in `src/codex/auth-api.ts`, `src/codex/auth-context.ts`, and `src/codex/account-lifecycle.ts`, #389 in `src/router.ts`, and #439 in `src/types.ts` — all files this PR touches, and GitHub currently reports the branch as `CONFLICTING`. Please re-cut against those completed contracts rather than restoring the pre-integration code.

Because this changes account selection, OAuth-created credential lifecycle, and credential-to-model routing, it remains a C4 authentication boundary requiring the explicit maintainer security review that `MAINTAINERS.md` mandates. The full review against the current head will follow.

chrisae9:
Updated the branch after merging current `dev` (`6f44caf8`) and pushed the review fixes in `897d2660`.

The maintainer privacy finding is fixed at the boundary: picker selectors now come from an explicit alias or persisted random `pXXXXXX` label. Legacy accounts without a stored label get independent randomness; the ID-derived diagnostic fallback is never used as a public selector. Raw account IDs/emails do not enter catalog slugs, provider log labels, or request errors.

I also addressed the remaining review findings:

- preserved routed catalog rows during discovery outages and removed stale slug collisions;
- kept exact-account routing fail-closed across Direct/Pool, cooldown probes, retries, WebSocket, and compact;
- enforced reserved-ID, provider-prefix, combo-alias, and canonical-`openai` guards;
- preserved account-qualified multi-agent candidates inside the client display window;
- serialized GUI writes, hydrated the effective setting, and surfaced recoverable catalog-refresh failures in both account-management surfaces;
- synchronized the localized UI and docs around generic multi-account targeting rather than personal/work roles.

Validation on the final head: 4,280 tests across 327 files, 0 failures; typecheck, GUI lint/build, privacy scan, docs build, and focused GUI/catalog regressions pass. All currently open inline review threads are resolved.

The PR remains draft and the C4 maintainer security-review requirement remains in place.

lidge-jun:
Thank you for your patience — a review of this was prepared yesterday but not posted because the head moved mid-pass, so this is a full re-audit at `897d2660` against `dev` `bef0dfbe`. The good news first: the earlier privacy concern about raw-ID-derived selectors is **not reproducible** at this head. Unlabeled legacy accounts now get independent random `pXXXXXX` selectors, raw namespace targets are omitted from `safeConfigDTO`, and the focused request-log test confirms stored account IDs stay out of logs. Nice work on that.

**Verdict: BLOCKED_SECURITY.** The change routes credentials to models at an authentication boundary — an account-qualified selector overrides Direct/Pool selection and injects one stored credential at `src/codex/auth-context.ts:130` — which `MAINTAINERS.md` gates behind explicit maintainer security approval. Four findings below, plus a conflict that must be resolved carefully.

**1. High — the rebase must preserve current `dev` response handling. `src/server/responses/core.ts:1213`.** This branch predates both the quota-aware subagent fallback series and #463's passthrough-error fix, which are now on `dev`. It has no `applySubagentModelFallback` or quota-terminal recording, and it no longer calls `formatPassthroughUpstreamError` before relaying non-2xx passthrough responses. Resolving the conflict in favor of the PR side would silently disable configured subagent fallback and bring back empty upstream 503s surfacing as opaque "Unknown error". Please rebase onto `bef0dfbe`, keep the current pipeline, and layer `fixedAccount`, account-qualified retry suppression, and safe namespace logging on top of it. An integration regression covering an account-qualified subagent request plus an empty non-2xx passthrough body would lock this down.

**2. High — the collaboration conflict drops the current fallback guidance contract. `src/server/responses/collaboration.ts:157`.** The PR's `MultiAgentGuidanceOptions` omits `subagentModelFallback`, removes `subagentFallbackGuidanceText`, and removes the `{{fallback}}` placeholder, all of which `dev` now exposes and persists. Taking the PR side leaves the management setting accepted but ignored when guidance is generated. Keep `dev`'s option, text, and placeholder while integrating the account-aware roster projection, and test both contracts together including a bare native fallback projected onto account-qualified catalog entries.

**3. Medium — account aliases can become persistent, privacy-sensitive log identifiers. `src/codex/account-namespaces.ts:56`.** `defaultPublicAccountLabel` derives the namespace from any existing account alias, and alias validation at `src/codex/auth-api.ts:571` permits email-like printable strings. That namespace then flows into catalog slugs and provider labels, while the full requested selector is persisted as `requestedModel` via `src/server/responses/core.ts:685` and `src/server/request-log.ts:580`. So an alias like `jun@example.com` becomes `jun-example.com` and is retained in usage and request logs. Please don't derive telemetry/catalog selectors from the general display alias — default to the random `pXXXXXX` selector, or add a separate explicitly-public selector field with stricter validation and a PII warning. A regression proving display aliases and emails cannot reach catalog or persisted log fields would close this.

Note that `bun run privacy:scan` passing does not cover this: the scanner detects committed literal emails, home paths, and token-shaped strings, not runtime data flow.

**4. Medium — picker-toggle refresh failures are reported as success. `src/server/management/config-routes.ts:240`.** The settings route awaits `refreshCodexCatalogBestEffort`, which swallows refresh exceptions, then returns the enabled state, and `gui/src/pages/CodexAuth.tsx:60` shows the success message. Your add and delete flows correctly return `catalogRefreshPending` — the primary enable/disable flow should too, so a failed refresh surfaces the existing recoverable `ocx sync` warning instead of leaving the persisted setting and the actual picker catalog silently out of sync.

**Conflicts against `dev` `bef0dfbe`** — `git merge-tree` reports exactly two files: `src/server/responses/collaboration.ts` and `src/server/responses/core.ts`.

```bash
git fetch origin
git rebase bef0dfbed1863ccc1403112546e84abda5e9deb7
# resolve both files by preserving current dev behavior, then layering the namespace changes
git add src/server/responses/collaboration.ts src/server/responses/core.ts
GIT_EDITOR=true git rebase --continue
git push --force-with-lease
```

**Please split this before the security review.** At 61 files and `+2706/-120`, this is more than one security-sensitive review can responsibly cover. A stacked series, each targeting `dev`, independently green, ideally under ~500 changed lines:

1. namespace foundation — `src/types.ts`, namespace helpers, config validation, collision rules, focused config/router unit tests
2. exact-account routing — auth context, account health, router, HTTP/compact/WebSocket paths, fail-closed security regressions
3. catalog/discovery — `src/codex/catalog/*`, `/v1/models`, account-aware subagent projection, catalog tests
4. lifecycle and management UI — account add/delete, settings endpoints, refresh-failure reporting, GUI components and tests
5. docs and localization — English reference/guides plus synchronized translations

**Verification at `897d2660`:** typecheck passed; focused backend suite 556 passed / 0 failed across 14 files; full isolated suite 4,280 passed / 0 failed across 327 files; focused GUI suite 12 passed / 0 failed; GUI lint and production build passed; privacy scan passed. That is a genuinely solid state — the gap is that none of it validates the conflict resolution against current `dev`. Please rerun the full suite, typecheck, GUI lint/build, privacy scan, and focused integration tests after the rebase.

---

<sub>Full re-audit as part of a maintainer review pass, against `dev` at `bef0dfbe`, at head `897d2660`. This supersedes the stale draft mentioned in the earlier marker.</sub>

chrisae9:
Superseded by the split PR series requested in the maintainer review. #512 is the namespace-foundation layer; exact-account routing, catalog/discovery, lifecycle and management UI, and documentation/localization will follow separately. Closing this monolithic draft while keeping #425 open until the series is complete.

PR #431, opened 2026-07-25
Fix MiniMax split reasoning in Codex

H-H-E:
## What changed

- rebuilt this one-commit PR on current `dev`
- added the generic `reasoningSplitModels` capability, which sends `reasoning_split: true` for supported OpenAI-compatible chat models
- enabled split reasoning plus reasoning-history preservation for MiniMax M-series models
- mapped MiniMax-M3 low effort to `thinking.type=disabled` and medium/higher effort to `thinking.type=adaptive`
- fixed both streaming and nonstream parsing so `reasoning_raw_delta` is emitted before final text when one response carries both fields
- preserved explicit user model-effort mappings and capability fields while registry defaults fill missing configuration

## Scope and security

The OAuth, key-login, and safe-config DTO edits only propagate `reasoningSplitModels` through existing provider-field whitelists. They do not change OAuth behavior, authentication policy, CORS policy, credential handling, or logging.

## Regression coverage

- nonstream and same-frame stream responses now assert reasoning-before-answer ordering
- empty deltas followed by a finish-only frame remain a clean no-op completion
- EOF without `[DONE]`, `finish_reason`, or usage remains fail-closed
- MiniMax request shape, history replay, both MiniMax registry entries, and explicit user overrides are covered

## Validation

- focused adapter, EOF, MiniMax, reasoning-effort, and provider-registry suites: 106 passed
- `bun run typecheck`
- full `bun run test`
- `bun run privacy:scan`
- `bun run lint:gui`

The PR remains a draft pending CI and maintainer review.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -522,6 +522,7 @@ export function createOpenAIChatAdapter(provider: OcxProviderConfig): ProviderAd
         messages,
         stream: parsed.stream,
       };
+      if (modelInList(provider.reasoningSplitModels, parsed.modelId)) body.reasoning_split = true;
       const maxTokens = resolveMaxTokens(provider, parsed);
       const openRouterRouting = resolveOpenRouterRouting(provider, parsed.modelId);
       if (openRouterRouting) body.provider = openRouterProviderPayload(openRouterRouting);
@@ -545,9 +546,10 @@ export function createOpenAIChatAdapter(provider: OcxProviderConfig): ProviderAd
           const budget = thinkingBudgetForEffort(parsed, reasoningEffort, maxTokens);
           if (budget !== undefined) body.thinking_budget = budget;
         } else if (modelInList(provider.thinkingToggleModels, parsed.modelId)) {
-          // Vendor thinking-toggle wire (MiMo v2.x, GLM 5/5.1): the mapped value is the toggle
-          // state, sent as `thinking: {type}` — these models ignore/reject reasoning_effort.
-          if (reasoningEffort === "enabled" || reasoningEffort === "disabled") {
+          // Vendor thinking-toggle wire: the mapped value is sent as `thinking: {type}` because
+          // these models ignore/reject reasoning_effort. Most use enabled/disabled; MiniMax-M3
+          // uses adaptive/disabled.
+          if (reasoningEffort === "enabled" || reasoningEffort === "disabled" || reasoningEffort === "adaptive") {
             body.thinking = { type: reasoningEffort };
           }
         } else {
@@ -679,12 +681,11 @@ export function createOpenAIChatAdapter(provider: OcxProviderConfig): ProviderAd
         }
         const delta = choices[0].delta;
         if (delta) {
-          if (typeof delta.content === "string" && delta.content.length > 0) {
-            yield { type: "text_delta", text: delta.content };
-          }
-
           if (typeof delta.reasoning_content === "string" && delta.reasoning_content.length > 0) {
             yield { type: "reasoning_raw_delta", text: delta.reasoning_content };
+          }
+          if (typeof delta.content === "string" && delta.content.length > 0) {
+            yield { type: "text_delta", text: delta.content };
           }
 
           const toolCalls = delta.tool_calls as { index?: number; id?: string; function?: { name?: string; arguments?: string } }[] | undefined;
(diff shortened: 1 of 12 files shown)


lidge-jun:

lidge-jun:
Thank you for keeping this as a draft while adding MiniMax's split-reasoning request capability. Before marking it ready for review, please fix the response-ordering regression and tighten the scope/test contract.

`src/adapters/openai-chat.ts:784-790` emits `text_delta` before `reasoning_raw_delta` when a non-streaming MiniMax response contains both `content` and `reasoning_content`, which is exactly the split response shape described in this PR. The bridge closes the text item when the later reasoning event arrives (`src/bridge.ts:470-542`), so the final answer is materialized before the reasoning item instead of reasoning preceding the answer. The streaming path has the same ordering when both fields share one delta at `src/adapters/openai-chat.ts:683-690`. Please emit non-empty reasoning before final content for both paths and add parser-level regressions for (1) a non-stream response containing both fields, (2) a stream frame containing both fields, (3) empty deltas followed by a finish-only frame, and (4) EOF without `[DONE]`, `finish_reason`, or usage. The last two should preserve the existing no-op and fail-closed behavior rather than treating an interrupted stream as complete.

The current `tests/minimax-reasoning-split.test.ts:38-95` only calls `buildRequest`; it proves `reasoning_split`, effort mapping, and history serialization, but never executes `parseStream` or `parseResponse`, so it cannot catch the response-ordering failure promised by the title. Please make the regression assert emitted adapter-event order, not only request JSON.

I traced the one-line additions in `src/oauth/index.ts:492-499`, `src/oauth/login-cli.ts:86-93`, and `src/server/auth-cors.ts:313-323`. They carry `reasoningSplitModels` through existing provider-config reconciliation, key-login construction, and the safe dashboard DTO; I do not see a new OAuth flow or CORS-origin policy in those hunks. Please explain that propagation in the PR scope and keep it to those whitelist paths, because OAuth/config serialization remains a security-sensitive boundary even when the value itself is only a model-id list.

Although this pinned patch currently passes `git apply --check`, it must be rebased before it becomes ready: `src/providers/registry.ts` overlaps the planned #385/WP7 integration, `src/router.ts` overlaps #389/WP5, and `src/types.ts` overlaps #439/WP3. After rebasing, verify that registry enrichment still preserves explicit user provider fields, run the focused MiniMax parser/request tests, and run the full typecheck and test suite. Thanks again for addressing the visible `<think>` leakage; the draft needs response-side proof and the integrated provider contracts before full review.

Current `dev` has now integrated #385 in `src/providers/registry.ts`, #389 in `src/router.ts`, and #439 in `src/types.ts`. Please rebase the draft against those completed contracts and preserve explicit user provider fields.

---

lidge-jun:
Integrated into `dev` as [`82a47dbe`](https://github.com/lidge-jun/opencodex/commit/82a47dbe) with your authorship preserved via `Co-authored-by`. Closing this PR since the work is on the integration branch.

The response-ordering fix you made after the earlier review is exactly right — `dev` was emitting content before reasoning in both the streaming and non-streaming paths, and your version corrects both. Independent verification against MiniMax's official OpenAI-compatible documentation confirmed every external claim: `reasoning_split` as a boolean output switch, M3's `adaptive`/`disabled` thinking types, all eight model IDs, and their context windows. The registry additions are internally consistent, with the global and CN entries symmetric and non-shadowing.

**Scope note.** I left out three one-line propagation hooks — the additions to `OAUTH_RECONCILE_FIELDS`, `providerConfigFromKeyLoginProvider`, and `safeConfigDTO`. I want to be straight about why, because my first reasoning was wrong. I initially flagged them as touching a security boundary, and review corrected me: adding a field name to a capability allowlist is not credential handling, and the neighboring `preserveReasoningContentModels` is already exposed identically in all three places. File location alone does not make something a trust boundary.

The actual reason is that they are unnecessary at runtime. MiniMax is a key-auth provider, so OAuth reconcile skips it entirely, and `routeModel()` unions the registry seed back on every route — verified by deleting the saved field and still observing `reasoning_split: true`. The honest tradeoff: `ocx login minimax` now writes a config without the field. Runtime-safe, but less complete than the neighboring capability lists, and a reasonable person could argue for including that one line. If you'd like it in, open a follow-up and I'll take it.

**Known limitation, recorded rather than papered over.** MiniMax documents preserving the complete `reasoning_details` field across multi-turn tool loops, while this carries only synthesized `reasoning_content`. There is no parser, type, replay path, or test for `reasoning_details` anywhere in the repository. The history assertion in `tests/minimax-reasoning-split.test.ts:87` reads back a fixture we wrote ourselves, so it cannot prove the upstream contract. Reasoning visibility works; interleaved-thinking continuity through a tool loop is unproven. Proving it needs a captured wire fixture we don't have, so I narrowed the claim instead of asserting it. That is written into the commit message and is a good follow-up if you have access to real traffic.

Verification: 67/67 across the four affected test files, red-green confirmed by removing the `reasoning_split` gate at `openai-chat.ts:527` (2 failures) and restoring it, full backend batches clean, typecheck and privacy scan green.

Thank you for sticking with this through the earlier review round — the ordering diagnosis was correct and the registry work is careful.

---

<sub>Integrated as part of a maintainer PABCD pass. Plan unit: `devlog/_plan/260726_pr_close_rework/070_wp7_minimax_reasoning_split_431.md`.</sub>

PR #437, opened 2026-07-25
docs: clarify Bun requirement for source development

CooperSheroy:
## Summary
- Clarify that source development requires a local `bun` CLI on `PATH`
- Distinguish contributor commands from the published npm package's bundled Bun runtime
- Add the note to the root contributor entry point, README development section, and hosted contributing docs source

## Validation
- `git diff --check`
- Not run: Bun-based project commands, because this cron host currently has `node`/`npm` but no `bun` binary on `PATH`.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that source development requires the Bun CLI to be installed and available on the system `PATH`.
  * Explained the distinction between the locally installed Bun used for development commands and the bundled runtime included with the published package.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,6 +20,10 @@ Agent-facing repository and review rules live in [`AGENTS.md`](./AGENTS.md).
 For local development commands, architecture notes, and release workflow details, use the hosted
 contributing guide above instead of duplicating instructions here.
 
+Source development requires the `bun` CLI on your `PATH`. The published npm package bundles its own
+Bun runtime for end users, but contributor commands such as `bun install`, `bun run test`, and
+`bun run prepush` run from your local Bun installation.
+
 ## Pre-push hook
 
 After cloning, run once to install a local pre-push hook that runs the typecheck,
--- a/README.md
+++ b/README.md
@@ -456,6 +456,9 @@ lives in [`SECURITY.md`](./SECURITY.md).
 
 ## Development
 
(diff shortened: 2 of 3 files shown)


lidge-jun:
Integrated into `dev` as [`803807a4`](https://github.com/lidge-jun/opencodex/commit/803807a4) with your authorship preserved via `Co-authored-by`. Closing this PR since the change is now on the integration branch.

Your three paragraphs went in unchanged — the wording correctly separates contributor commands from the npm package's bundled runtime, and it doesn't contradict the end-user install section further up in `README.md`. Thank you for catching a gap that trips up first-time contributors.

I added one regression test on top, `tests/docs-bun-source-requirement.test.ts`. My first attempt matched scattered fragments per file, and an independent review caught that this would keep passing after someone deleted the explanatory sentence, as long as the words survived elsewhere in the file — a test that can't fail for the reason it exists. It also would have broken on a harmless paragraph rewrap. The version that landed locates the paragraph, normalizes its whitespace, and compares the complete paragraph.

Verification, in order: fails with your doc patch reverted, passes with it applied, still passes after rewrapping the paragraph across different line breaks, and fails again when the substantive sentence is removed. Full gates on the integration commit: `bun run typecheck` clean, `privacy:scan` clean, and the full suite at 4324 passing across three batches.

---

<sub>Integrated as part of a maintainer PABCD pass. Plan unit: `devlog/_plan/260726_pr_close_rework/010_wp1_docs_bun_requirement_437.md`.</sub>

PR #455, opened 2026-07-25
[WRONG BRANCH] chore: temporary verification export trigger

lidge-jun:
Temporary pull request used only to trigger export of the Go toolchain, vendored modules, and TypeScript parity reference for local sandbox verification. It will be closed and all temporary refs reset immediately after acquisition.


lidge-jun:
Closing: this was a one-shot export trigger to verify the `dev2-go` release-asset path, and that verification is done — `dev2-go` has since recorded its release-gate receipt (`aadba2c2`).

The branch is conflicted against its base and carries a 69k-line diff that was never meant to merge.

One note so the record is not misleading: the `[WRONG BRANCH]` prefix predates `d761e880` ("ci: let pull requests target dev2-go, not just dev"), which added `dev2-go` to the allowed bases. The prefix is stale — it is not the reason for closing.

Re-trigger from a fresh branch if the export ever needs another run.

PR #392, opened 2026-07-24
feat(api-access): expose gateway endpoints and external model catalog

Wibias:
## Summary

Follow-up for #357 on top of the already-merged Chat Completions gateway work in #279.

This PR improves the external aggregated model API experience without re-implementing `/v1/chat/completions`.

## Changes

- Add `buildApiAccessEndpoints()` and return richer metadata from `GET /api/keys`
- Expand the API Access page to show:
  - `/v1`
  - `/v1/responses`
  - `/v1/chat/completions`
  - `/v1/messages`
  - `/v1/models`
- Document auth differences between Responses and Chat Completions
- Add an external model catalog with copy-ID and simple test actions
- Add i18n coverage for `en`, `de`, `ja`, `ko`, `ru`, `zh`
- Add regression test for API access endpoint metadata

## Verification

- `bun test tests/api-access-endpoints.test.ts`
- `cd gui && bun x tsc -b`
- `cd gui && bun run build`

## Notes

This is intentionally scoped to the API Access slice. Remaining #357 work such as richer SDK examples, stronger remote-auth test UX, and key scopes/limits can follow in later PRs.

## Summary by CodeRabbit

* **New Features**
  * Expanded API Access with protocol-specific endpoints, authentication guidance, and curl/usage snippets.
  * Added an external **Models** catalog with searchable results, model ID copy, and per-model connectivity testing (including gated Claude “messages” support).
* **Localization**
  * Updated API Access translations (DE/EN/JA/KO/RU/ZH) for endpoints, auth notes, and external model catalog text.
* **Bug Fixes**
  * Improved API endpoint/base URL generation to correctly handle IPv6, wildcard binds, and request context.
* **UI/Style**
  * Refined API page layout and connectivity status styling for models.
* **Tests**
  * Added coverage for endpoint construction, IPv6/wildcard behavior, and external model/protocol classification.

PR #459, opened 2026-07-25
fix(providers): honor saved baseUrl for alibaba-token-plan via override + choices

### LINKED ISSUE #457: alibaba-token-plan registry entry silently overrides saved baseUrl -> 401 after provider migration
## Summary

Two behaviors combine into a silent breakage for international Alibaba Token Plan users:

1. **The provider migration** (shipped around 2026-07-21, created `config.json.bak-before-alibaba-provider-rename-*`) renamed/merged user providers into the id `alibaba-token-plan` and deleted the user's duplicate (`alibaba-token-plan-1`).
2. **The router** (`src/router.ts:158-162`) ignores the user's saved `baseUrl` whenever the registry entry lacks `allowBaseUrlOverride` — and `alibaba-token-plan` (Beijing) lacks it, while `alibaba-token-plan-intl` (Singapore) has it.

Net effect: after the migration, `config.json` points at the saved international URL (`token-plan.ap-southeast-1.maas.aliyuncs.com`), but the router forces the Beijing URL (`token-plan.cn-beijing.maas.aliyuncs.com`). The international plan key is rejected there → **every request fails with 401**, with no warning that the configured baseUrl was discarded.

## Registry asymmetry (`src/providers/registry.ts`)

| id | baseUrl | `allowBaseUrlOverride` |
|---|---|---|
| `alibaba-token-plan` | `token-plan.cn-beijing...` | ❌ (absent) |
| `alibaba-token-plan-intl` | `token-plan.ap-southeast-1...` | ✅ + `baseUrlChoices` |

## Before/after migration (actual `~/.opencodex/config.json`)

Before:
```json
"alibaba-token-plan":   { "baseUrl": "https://coding-intl.dashscope.aliyuncs.com/v1" },
"alibaba-token-plan-1": { "baseUrl": "https://token-plan.ap-southeast-1.maas.aliyuncs.com/compatible-mode/v1" }
```

After:
```json
"alibaba-token-plan": { "baseUrl": "https://token-plan.ap-southeast-1.maas.aliyuncs.com/compatible-mode/v1" }
```

The saved URL survived, but `router.ts` resolves it back to Beijing at runtime.

## Suggested fixes

1. **Honor the saved baseUrl**: give `alibaba-token-plan` `allowBaseUrlOverride: true` + `baseUrlChoices` (Beijing/Singapore), mirroring the `-intl` entry — or make region a first-class field.
2. **Fail loudly**: when the router discards a user-configured `baseUrl` because override isn't allowed, emit a warning. Silent ignoring is the footgun.
3. **Migrations must not rename user provider ids** into a registry id with different semantics (region, override policy). If consolidation is needed, preserve the resolved baseUrl or map to the id matching the saved URL's region.

## Workaround (confirmed)

Rename the provider id to `alibaba-token-plan-intl` in `config.json` and re-prefix `disabledModels`/`providerContextCaps`. The international URL is then honored; the endpoint returns 200.

## Environment

opencodex 2.7.40 (npm global), macOS arm64. Note: `ocx update` runs migrations unattended via the background service, so this class of breakage lands without user interaction.

alexanderxcmo:
Fixes #457

## Problem

The `alibaba-token-plan` registry entry (Beijing) lacks `allowBaseUrlOverride`, so the router silently discards any user-saved `baseUrl` and forces the Beijing endpoint (`router.ts`: registry URL wins unless the entry is a template or opts into override).

After the July provider migration merged user configs into the id `alibaba-token-plan`, configs holding an international (`ap-southeast-1`) baseUrl are still routed to `cn-beijing` → **all requests fail with 401**, with no warning that the saved URL was ignored.

## Fix

Mirror the existing `qwen-cloud` pattern:

- `src/providers/base-url-choices.ts`: add `ALIBABA_TOKEN_PLAN_BEIJING_BASE_URL`, `ALIBABA_TOKEN_PLAN_INTL_BASE_URL`, and `ALIBABA_TOKEN_PLAN_BASE_URL_CHOICES` (`beijing` / `intl` / `custom`).
- `src/providers/registry.ts`: `alibaba-token-plan` gets `allowBaseUrlOverride: true` + `baseUrlChoices`, and uses the Beijing constant (value unchanged).

Default resolution is unchanged when no user `baseUrl` is set — only configs with a saved URL now have it honored, and the GUI can offer the region choice.

## Tests

- New `tests/alibaba-token-plan-endpoints.test.ts` mirrors the qwen-cloud endpoint tests (registry shape, presets projection, `matchBaseUrlChoice` mapping).
- Updated the override allowlist assertion in `tests/provider-registry-parity.test.ts`.
- `tsc --noEmit` clean, full suite: **3231 pass / 0 fail**, privacy scan passed.

## Notes

- Issue #457 also documents two wider follow-ups left for maintainer discretion (router warning when a saved baseUrl is discarded; migration renaming semantics) — intentionally not in this PR.

### DIFF
--- a/src/providers/base-url-choices.ts
+++ b/src/providers/base-url-choices.ts
@@ -21,6 +21,23 @@ export const QWEN_CLOUD_BASE_URL_CHOICES: readonly ProviderBaseUrlChoice[] = [
   { id: "custom", label: "Custom" },
 ];
 
+/**
+ * Alibaba Token Plan OpenAI-compatible endpoint presets.
+ * Beijing is the registry default (Personal Edition); the international
+ * Singapore endpoint (Team Edition) and custom are selectable, so a saved
+ * non-Beijing baseUrl is honored instead of silently overridden.
+ */
+export const ALIBABA_TOKEN_PLAN_BEIJING_BASE_URL =
+  "https://token-plan.cn-beijing.maas.aliyuncs.com/compatible-mode/v1";
+export const ALIBABA_TOKEN_PLAN_INTL_BASE_URL =
+  "https://token-plan.ap-southeast-1.maas.aliyuncs.com/compatible-mode/v1";
+
+export const ALIBABA_TOKEN_PLAN_BASE_URL_CHOICES: readonly ProviderBaseUrlChoice[] = [
+  { id: "beijing", label: "China (Beijing)", baseUrl: ALIBABA_TOKEN_PLAN_BEIJING_BASE_URL },
+  { id: "intl", label: "International (Singapore)", baseUrl: ALIBABA_TOKEN_PLAN_INTL_BASE_URL },
+  { id: "custom", label: "Custom" },
+];
+
 /** Match a saved baseUrl to a known choice id (`custom` when it does not match). */
 export function matchBaseUrlChoice(
   choices: readonly ProviderBaseUrlChoice[],
--- a/src/providers/registry.ts
+++ b/src/providers/registry.ts
@@ -2,7 +2,7 @@ import type { CodexAccountMode, OcxProviderConfig } from "../types";
 import { KIRO_MODELS, KIRO_MODEL_CONTEXT_WINDOWS, KIRO_MODEL_REASONING_EFFORTS } from "./kiro-models";
 import { ANTIGRAVITY_MODELS, ANTIGRAVITY_MODEL_CONTEXT_WINDOWS } from "./antigravity-models";
 import type { ProviderBaseUrlChoice } from "./base-url-choices";
-import { QWEN_CLOUD_BASE_URL_CHOICES, QWEN_CLOUD_TOKEN_PLAN_BASE_URL } from "./base-url-choices";
+import { ALIBABA_TOKEN_PLAN_BASE_URL_CHOICES, ALIBABA_TOKEN_PLAN_BEIJING_BASE_URL, QWEN_CLOUD_BASE_URL_CHOICES, QWEN_CLOUD_TOKEN_PLAN_BASE_URL } from "./base-url-choices";
 import {
   CURSOR_STATIC_MODELS,
   cursorModelContextWindows,
@@ -690,9 +690,11 @@ export const PROVIDER_REGISTRY: readonly ProviderRegistryEntry[] = [
   {
     id: "alibaba-token-plan",
     label: "Alibaba Token Plan (Beijing)",
-    baseUrl: "https://token-plan.cn-beijing.maas.aliyuncs.com/compatible-mode/v1",
+    baseUrl: ALIBABA_TOKEN_PLAN_BEIJING_BASE_URL,
     adapter: "openai-chat",
     authKind: "key",
+    allowBaseUrlOverride: true,
+    baseUrlChoices: ALIBABA_TOKEN_PLAN_BASE_URL_CHOICES,
     dashboardUrl: "https://bailian.console.aliyun.com/cn-beijing?tab=plan",
     defaultModel: "qwen3.8-max-preview",
     models: ALIBABA_TOKEN_PLAN_MODELS,
--- a/tests/alibaba-token-plan-endpoints.test.ts
+++ b/tests/alibaba-token-plan-endpoints.test.ts
@@ -1,0 +1,35 @@
+import { describe, expect, test } from "bun:test";
+import {
+  ALIBABA_TOKEN_PLAN_BASE_URL_CHOICES,
+  ALIBABA_TOKEN_PLAN_BEIJING_BASE_URL,
+  ALIBABA_TOKEN_PLAN_INTL_BASE_URL,
+  matchBaseUrlChoice,
+} from "../src/providers/base-url-choices";
+import { PROVIDER_REGISTRY } from "../src/providers/registry";
+import { deriveProviderPresets } from "../src/providers/derive";
+
+describe("alibaba-token-plan endpoint choices", () => {
+  test("registry defaults to Beijing and exposes intl + custom choices", () => {
+    const entry = PROVIDER_REGISTRY.find(e => e.id === "alibaba-token-plan");
+    expect(entry).toBeDefined();
+    expect(entry!.baseUrl).toBe(ALIBABA_TOKEN_PLAN_BEIJING_BASE_URL);
+    expect(entry!.allowBaseUrlOverride).toBe(true);
+    expect(entry!.baseUrlChoices?.map(c => c.id)).toEqual(["beijing", "intl", "custom"]);
+    expect(entry!.baseUrlChoices).toEqual([...ALIBABA_TOKEN_PLAN_BASE_URL_CHOICES]);
+  });
+
+  test("presets API projection includes baseUrlChoices", () => {
+    const preset = deriveProviderPresets().find(p => p.id === "alibaba-token-plan");
+    expect(preset?.baseUrl).toBe(ALIBABA_TOKEN_PLAN_BEIJING_BASE_URL);
+    expect(preset?.baseUrlChoices?.map(c => c.id)).toEqual(["beijing", "intl", "custom"]);
+    const intl = preset?.baseUrlChoices?.find(c => c.id === "intl");
+    expect(intl?.baseUrl).toBe(ALIBABA_TOKEN_PLAN_INTL_BASE_URL);
+    expect(preset?.baseUrlChoices?.find(c => c.id === "custom")?.baseUrl).toBeUndefined();
+  });
+
+  test("matchBaseUrlChoice maps known hosts and falls back to custom", () => {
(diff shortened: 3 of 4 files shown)


snowyukitty:
Thanks for chasing this down — the 401 is real, and the router analysis in #457 is spot on. While reading the same code I ran into two bits of history that are easy to miss, and I think they change what this PR should do. Sharing them here rather than in the issue so they're next to the diff.

## `allowBaseUrlOverride` on `alibaba-token-plan` already shipped once, and was backed out

Author dates are misleading here — the branch was rebased, and `a9b9048a` carries an author date 25 minutes *earlier* than the commit it reverts. This ordering is from `git merge-base --is-ancestor` on `dev`:

| Commit | What |
|---|---|
| `9b412d8e` | cherry-pick of #189 — adds `allowBaseUrlOverride: true` to `alibaba-token-plan` |
| `a9b9048a` | `Revert "fix: preserve configured Alibaba Token Plan base URL (#189)"` |
| `7e63d2e0` | `feat: add alibaba-token-plan-intl provider for international token plan` |

The revert carries no stated reason, but what replaced it is legible in the code: region became a separate registry id. `alibaba-token-plan` stays pinned to Beijing with no override, and `alibaba-token-plan-intl` is the entry that carries `allowBaseUrlOverride` — with `ALIBABA_INTL_BASE_URL_CHOICES` offering *plan type within one region* (`token-plan` / `payg` / `custom`), mirroring `QWEN_CLOUD_BASE_URL_CHOICES` rather than crossing regions.

This PR's `beijing` / `intl` / `custom` set puts both regions back under one id, which is the shape the revert removed. @lidge-jun is the only one who knows whether that revert was deliberate policy or a rollback for an unrelated reason, and the answer decides the whole PR — probably worth a word before investing more in it.

Also worth noting: your own confirmed workaround in #457 — renaming the id to `alibaba-token-plan-intl` — is exactly what the post-revert design expects.

## On the migration: I couldn't find it, and I'd genuinely like to be wrong

The premise that a shipped migration merged user providers into `alibaba-token-plan` is what makes re-adding the override look necessary, so I tried to confirm it. Searching `v2.7.36`, `v2.7.37`, `v2.7.38-preview.20260724`, `v2.7.39`, `v2.7.40` and `dev`:

- no hits for `bak-before`, `provider-rename`, `alibaba-token-plan-1`, or `renameProviderId`;
- `alibaba` appears only in `src/providers/registry.ts`, `src/providers/base-url-choices.ts`, `src/usage/expected-prices.ts`, `src/generated/jawcode-model-metadata.ts` and `gui/src/provider-icons.ts` — no id-rewriting code;
- the only config-backup names this codebase can produce are `<config>.pre-openai-tiers-v2.bak`, `<config>.pre-openai-tiers-v1-rollback.<ts>.bak` and `<config>.invalid-<iso>`, none of which match `.bak-before-<name>-<timestamp>`.

You clearly saw that file, so the likeliest explanation is that I'm searching for the wrong thing. If you still have `config.json.bak-before-alibaba-provider-rename-*` on disk, its mtime and the `ocx` version that wrote it would settle this in a minute.

## The part that seems unambiguous either way

Independent of how region ends up being modelled: `src/router.ts:162` discards a saved, well-formed `baseUrl` with no diagnostic at all, so the only thing a user ever sees is a 401 from an endpoint they never configured. That's your suggested fix 2, which you deliberately kept out of this PR — and I think you were right to, because it's design-neutral: it holds for every pinned registry entry, whichever way the Alibaba question goes. There's a precedent for the shape at `src/config.ts:714`, where an invalid `streamMode` warns before falling back.

I'm happy to put that up as a small separate PR so it doesn't tangle with the policy question here — but it's your finding and your issue, so just say if you'd rather own it and I'll stay out of the way.

lidge-jun:

lidge-jun:
Thanks for chasing down the Alibaba endpoint problem. The underlying complaint is legitimate, but the approach here conflicts with a design decision that has since landed on `dev`, so this needs a rework rather than a rebase.

**High — this restores a single-provider cross-region model that current `dev` deliberately replaced.** The PR adds `allowBaseUrlOverride` and an international choice to the Beijing entry (`src/providers/registry.ts:691-703`, `src/providers/base-url-choices.ts:24-39`). Current `dev` at `src/providers/registry.ts:813-875` instead gives the international Team Edition its own `alibaba-token-plan-intl` provider with its own model list, modality map, context windows, and reasoning metadata.

The concrete failure: a user picks the Singapore URL through your `intl` choice but keeps provider id `alibaba-token-plan`. Requests then go to the international endpoint while routing and catalog metadata still describe Beijing Personal Edition — its dashboard link, its note, its default model, its model and capability maps. A URL override cannot carry all of those contracts across, so the user gets a working endpoint attached to wrong metadata.

If stale configuration is the real bug you're solving, the right shape is an explicit migration: detect a recognized Singapore URL saved under `alibaba-token-plan` and move it to `alibaba-token-plan-intl`, preserving credentials and user-owned fields, idempotently.

**Medium — the tests don't exercise the behavior in the title. `tests/alibaba-token-plan-endpoints.test.ts:11-34`** checks constants, preset projection, and `matchBaseUrlChoice`, but never calls `routeModel`. It would not catch a failure to honor the saved URL, nor the endpoint/metadata mismatch above.

**Exact conflicts against `dev` `bef0dfbe`** (merge base `3cc1b6f7`):

```
CONFLICT (content): src/providers/base-url-choices.ts
CONFLICT (content): src/providers/registry.ts
CONFLICT (content): tests/provider-registry-parity.test.ts
```

All three are semantic, not textual noise: `dev` added `ALIBABA_INTL_*` choices scoped to the international provider, registered `alibaba-token-plan-intl`, and expects that id in the override allowlist; this PR puts the Beijing provider in each of those positions instead.

Rebase instruction:

```bash
git fetch origin
git rebase origin/dev
```

When resolving, please keep the current `ALIBABA_INTL_*` constants and the `token-plan` / `payg` / `custom` choices, keep `alibaba-token-plan` pinned to Beijing, keep `alibaba-token-plan-intl` as the override-capable provider, and keep `alibaba-token-plan-intl` (not Beijing) in the parity allowlist. Then replace the current test with migration plus end-to-end routing coverage if stale configs still need repair.

**Also relevant:** #464 adds a warning for exactly this situation — a pinned provider discarding a configured `baseUrl` — and documents the Beijing/international split as the resolution. If this PR landed as written, that warning would stop firing for `alibaba-token-plan` and its documentation would become false. My recommendation is that #464 lands first and this one is reframed as a migration.

**Test gap:** no `routeModel` assertion, no stale-config migration test, no proof that a Singapore endpoint receives the international provider's model and capability metadata. The conflicting head also hasn't run full cross-platform CI against current `dev`.

---

<sub>Reviewed as part of a maintainer review pass against `dev` at `bef0dfbe`, at head `64a8753e`. All four changed files reviewed.</sub>

lidge-jun:
Closing this without merging. The bug you reported is real — a saved `baseUrl` being silently discarded and turning into a 401 is genuinely bad, and #464 exists because of your report. But this particular fix can't land as written, and I want to be specific about why rather than leave you guessing.

**Current `dev` deliberately models Beijing and International as separate providers.** `alibaba-token-plan` is pinned to the Beijing Personal Edition, while `alibaba-token-plan-intl` owns the Singapore/international endpoints with its own model list, modality map, context windows, and reasoning metadata. Adding `allowBaseUrlOverride` to the Beijing entry collapses that split back together: you'd get the international endpoint while routing and the catalog still describe the provider as Beijing Personal Edition — its dashboard link, its note, its default model, its capability maps. A URL override can't carry those contracts across, so the result is a working endpoint wearing the wrong identity. Worse, that provider's API key would then be sendable to whatever destination happens to be saved.

The PR is also stale against `dev` now, with semantic conflicts in `src/providers/base-url-choices.ts`, `src/providers/registry.ts`, and `tests/provider-registry-parity.test.ts` — `dev` added `ALIBABA_INTL_*` choices scoped to the international provider and expects that id in the override allowlist, where this PR puts Beijing in each of those positions.

**On the underlying problem.** If the real issue is that an unattended rename or merge left your config pointing `alibaba-token-plan` at a Singapore URL, that deserves a fix — but it's a migration, not an override flag. A safe one has to move the API key, `defaultProvider`, routed model references, `disabledModels`, `providerContextCaps`, and any key pools atomically, and decide what happens when an `alibaba-token-plan-intl` entry already exists. I looked for the rename path described in #457 (`config.json.bak-before-alibaba-provider-rename-*`) and it isn't in this repository, so I can't write that migration responsibly without knowing what produced those backups.

**Issue #457 stays open.** This PR's body says `Fixes #457`, but closing this must not close the issue — the underlying recovery path is still unresolved and worth tracking. In the meantime, #464 adds the diagnostic so a discarded `baseUrl` produces a named warning instead of a bare 401, which at least turns your silent failure into a visible one.

If you can share what performed the rename — a version, a command, or the backup file's contents with credentials stripped — that would move #457 forward considerably. Thanks for reporting it; the silent-discard behavior was worth surfacing.

---

<sub>Reviewed as part of a maintainer PABCD pass against `dev`. Plan unit: `devlog/_plan/260726_pr_close_rework/090_wp9_security_hold_and_close_459.md`.</sub>

PR #460, opened 2026-07-25
feat(kiro): honor native END_TURN and Opus 5 native reasoning effort

mushikingh:
Native end-turn: Kiro's terminal metadataEvent can carry a stopReason. An END_TURN response holding plain assistant text, with neither a real tool call nor a private completion call to arbitrate against it, now ends the turn directly with that text as the final answer. The bounded completion retry is kept for streams that omit the signal, so an ordinary tool-enabled turn stops paying for a second inference request.

Because the stop reason only arrives at the end of the stream, required mode holds staged assistant events until either a real tool call starts (released as commentary) or the stream ends (released as the final answer on END_TURN, otherwise as commentary). Each held event yields a heartbeat in its place so the bridge's stall watchdog stays armed, and anything still held when the stream fails is released before the terminal event. This trades token-by-token rendering of a tool-enabled turn's answer for removing the extra request that the same turn previously always paid.

Opus 5 reasoning effort: Kiro accepts effort for Claude models through the output_config.effort request field rather than the Sol-only reasoning.effort field, so claude-opus-5 now sends its selected level natively instead of falling back to emulated thinking instructions.

Restatement suppression: the duplicate-answer check compared the bounded retry against the preceding commentary with whitespace-normalized exact equality, so any rewording rendered the same answer twice. It now compares by shared in-order word sequence, ignoring whitespace, punctuation, and case, gated on a growth bound and a longest-inserted-run bound so a retry that repeats the commentary and then appends real new content is still kept.

## Summary

- Explain the user-visible or maintainer-facing change.

## Verification

- List the commands or checks you ran.

## Checklist

## Summary by CodeRabbit

* **Improvements**
  * Kiro responses now handle native end-of-turn signals more accurately, reducing unnecessary completion requests.
  * Tool-enabled responses provide clearer progress and final-answer handling during streaming.
  * Repeated or near-duplicate responses are suppressed more reliably.
  * Native reasoning effort support is now available for additional Kiro models, with validation for unsupported values.

* **Documentation**
  * Updated Kiro completion, streaming, web-search buffering, and reasoning-effort guidance.

* **Tests**
  * Expanded coverage for completion signals, reasoning effort, streaming behavior, and repeated responses.

### DIFF
--- a/docs-site/src/content/docs/reference/adapters.md
+++ b/docs-site/src/content/docs/reference/adapters.md
@@ -101,24 +101,36 @@ streams the response back **untranslated**.
 
 ### Completion semantics
 
-Kiro text events do not carry a dependable end-turn phase. When an ordinary client tool is present,
-opencodex therefore adds a private `codex_kiro_final_answer` tool to the upstream request. Progress
-text streams as commentary and cannot terminate the turn. The adapter consumes the private call,
-emits its answer as final text, and never exposes the private tool to Codex or Claude Code.
-When the web-search sidecar is active, this commentary still streams immediately; only the events
-needed to decide whether the model requested a synthetic search remain buffered.
+Kiro assistant text carries no dependable end-turn phase of its own. Its terminal `metadataEvent`
+can, however, carry a native `stopReason`. An `END_TURN` response holding plain assistant text with
+no client tool call ends the turn directly, with that text emitted as the final answer and no extra
+model round trip.
 
-If Kiro emits progress without calling the completion tool, the adapter makes one continuation. That
-single retry may finish with a validated private completion or plain final text. It cannot recurse:
-an empty or reasoning-only retry is returned as retryable incomplete, while a real client tool call
-keeps the turn open. If the retry only repeats the preceding commentary after whitespace
-normalization, the duplicate output is suppressed while the turn still completes. Tool-free
+When the stop reason is absent or is anything other than `END_TURN`, the compatibility path applies.
+If an ordinary client tool is present, opencodex adds a private `codex_kiro_final_answer` tool to
+the upstream request; progress text streams as commentary and cannot terminate the turn. The adapter
+consumes the private call, emits its answer as final text, and never exposes the private tool to
+Codex or Claude Code. Because the stop reason only arrives at the end of the stream, assistant text
+in a tool-enabled turn is held until either a real tool call starts (released as commentary) or the
+stream ends (released as the final answer on `END_TURN`, otherwise as commentary). When the
+web-search sidecar is active, released commentary still streams ahead of the terminal event; only
+the events needed to decide whether the model requested a synthetic search remain buffered.
+
+If Kiro emits progress without an `END_TURN` stop reason and without calling the completion tool,
+the adapter makes one continuation. That single retry may finish with a validated private completion
+or plain final text. It cannot recurse: an empty or reasoning-only retry is returned as retryable
+incomplete, while a real client tool call keeps the turn open. If the retry only restates the
+preceding commentary, the duplicate output is suppressed while the turn still completes.
+Restatement is judged by shared in-order word sequence, ignoring whitespace, punctuation, and case,
+so a reworded repeat is caught while a retry that appends real new content is kept. Tool-free
 requests retain normal text completion behavior.
 
 ### Reasoning effort
 
-`gpt-5.6-sol` has verified native effort support. Its selected `low`, `medium`, `high`, `xhigh`, or
-`max` value is sent as `additionalModelRequestFields.reasoning.effort`. Other Kiro models currently
+`gpt-5.6-sol` and `claude-opus-5` have verified native effort support, and each model family names
+the request field differently. A selected `low`, `medium`, `high`, `xhigh`, or `max` value is sent
+as `additionalModelRequestFields.reasoning.effort` for `gpt-5.6-sol` and as
+`additionalModelRequestFields.output_config.effort` for `claude-opus-5`. Other Kiro models currently
 use emulated reasoning: opencodex converts the selected level into bounded thinking instructions in
 the user content because their native effort field has not been verified. Do not interpret an
 advertised effort control on those models as proof of upstream-native reasoning support.
--- a/src/adapters/kiro-events.ts
+++ b/src/adapters/kiro-events.ts
@@ -6,7 +6,7 @@ export type ParsedKiroEvent =
   | { type: "reasoning"; data?: string }
   | { type: "tool"; name?: string; toolUseId?: string; input?: string; stop?: boolean }
   | { type: "truncation"; data: string }
-  | { type: "metadata"; usage?: OcxUsage; contextUsagePercentage?: number }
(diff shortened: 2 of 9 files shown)


lidge-jun:

lidge-jun:
Thanks for this — honoring the native `END_TURN` signal is the right direction, and the Opus 5 effort change checks out against Kiro's documentation (`low`/`medium`/`high`/`xhigh`/`max` via `output_config.effort`, https://kiro.dev/docs/cli/chat/effort/). Two correctness problems block merge.

**1. High — every stop reason other than `END_TURN` is treated as progress and triggers another inference request. `src/adapters/kiro.ts:1005-1036,1100-1103`.** When Kiro returns terminal text with `MAX_TOKENS`, `CONTENT_FILTERED`, `MODEL_CONTEXT_WINDOW_EXCEEDED`, or `MALFORMED_TOOL_USE`, the adapter releases the text as commentary and invokes the private completion fallback. For `MAX_TOKENS` that hides truncation behind another paid request. For context-window exhaustion it can immediately rebuild and submit another oversized request, which works against the context-pressure behavior integrated into `dev` as `fc517004`. Content filtering and malformed tool use get misread as ordinary progress in the same way. AWS's stop-reason enum treats these as distinct terminal states, not interchangeable with `TOOL_USE` (https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/services/bedrockruntime/model/StopReason.html).

Please normalize and exhaustively map the known reasons: keep `END_TURN` final; map `MAX_TOKENS` to `done` with `stopReason: "max_tokens"` so the bridge emits the correct incomplete terminal; map content filtering and context-window exhaustion to their proper error/incomplete states; accept `TOOL_USE` only when a tool call was actually decoded. Use the bounded compatibility retry only when the stop reason is absent or explicitly indicates a compatible continuation.

**2. High — fuzzy restatement suppression can discard a genuine final status update. `src/adapters/kiro-restatement.ts:28-77`** (integrated at `src/adapters/kiro.ts:602-604,1041-1063`). If long progress commentary and the final retry share at least 65% of their first 400 words, grow by no more than 20%, and differ only in runs of at most 11 words, `isKiroRestatement` returns true and the final text is suppressed. The realistic case is a checklist where "I will update / run / verify" becomes "I updated / ran / verified": all three thresholds pass, and the turn ends showing only the intended work, never the completed result. The code comment says suppression should be conservative, but a one-word status flip inside a long otherwise-identical response slips through.

I'd keep normalized exact-match suppression and drop the semantic fuzzy path unless it can be shown not to erase status changes. If fuzzy matching stays, any changed status, negation, or result clause must preserve the candidate. Please add a long near-identical regression where future-tense progress becomes a completed result.

**3. Medium — translated adapter references are out of sync.** The English change at `docs-site/src/content/docs/reference/adapters.md:102-136` has no counterpart in `ja/reference/adapters.md:66-75`, `ko/…:79-90`, `ru/…:84-98`, or `zh-cn/…:75-86`, so non-English readers miss END_TURN buffering, completion fallback behavior, and Opus 5 effort support entirely.

**Test gap:** nothing covers `MAX_TOKENS`, `CONTENT_FILTERED`, `MODEL_CONTEXT_WINDOW_EXCEEDED`, unknown stop reasons, or `TOOL_USE` without a decoded tool call. There's also no assertion that stop reasons and `contextTotalTokens` survive the terminal mapping introduced after `fc517004`, and no adversarial restatement case as described above.

**CI status:** `fc517004` is an ancestor of this head, so the context-pressure interaction is present in what I reviewed. macOS and Windows passed; Ubuntu reached 4,240 passing tests but failed one `combo management API` timeout that looks unrelated. Required CI still needs to be green on the fixed head. No Node-only APIs were introduced.

---

<sub>Reviewed as part of a maintainer review pass against `dev` at `bef0dfbe`, at head `664934bf`. All nine changed files reviewed.</sub>

lidge-jun:
Integrated into `dev` as [`74ddd96d`](https://github.com/lidge-jun/opencodex/commit/74ddd96d) with your authorship preserved via `Co-authored-by`. Closing this PR since the work is on the integration branch.

Your core insight was right and worth fixing: treating every non-`END_TURN` stop reason as unfinished progress meant `MAX_TOKENS` truncation got hidden behind a second paid inference, and context exhaustion resubmitted a request that could not fit. Three things changed during integration, and the first one is the reason the original patch could not have worked.

**The guard never fired.** `parseKiroEvent` runs the generic truncation sniffer *before* its event-type switch, and that sniffer matches substrings like `max_tokens` across several keys — including `stopReason`. So `MAX_TOKENS` became a `truncation` event and never reached the new branch at all. Running your tests against the PR head confirmed it: `fetches` was already 0, and the terminal was a non-retryable protocol error rather than the expected `done`. The fix gates the sniffer positionally — a `metadataEvent` with a string `stopReason` skips it regardless of value. A value allowlist would not have been enough, since `LENGTH_LIMIT` matches the pattern today and future reasons will too.

**Context exhaustion needed a different terminal.** Mapping `MODEL_CONTEXT_WINDOW_EXCEEDED` to `max_output_tokens` makes `responses/state.ts` cache the partial for continuation replay, and it surfaces as Claude `max_tokens` / Chat `length` — all of which invite a retry that cannot succeed. It now reuses the existing non-retryable `400 / context_length_exceeded` contract from `kiro-errors.ts`. The full matrix also handles `STOP_SEQUENCE` as an ordinary completion (it was becoming an incomplete), and reports a `TOOL_USE` that arrives without an actual tool call as a contradiction rather than progress.

One subtlety worth recording: `STOP_SEQUENCE` had to be folded into the existing `nativeEndTurn` predicate rather than handled in the new matrix, because deferred text is relabeled to `final_answer` in that earlier branch. Handling it later would have produced the right terminal with the text still marked commentary. All four original guards (`sawText`, `!sawRealTool`, `completionAnswer === undefined`, `completionCalls === 0`) are preserved.

**`kiro-restatement.ts` was not integrated.** Its thresholds classify two long, near-identical messages as duplicates even when a single status word flips the outcome — "the migration is still pending" versus "the migration is now complete" clears all three thresholds, and review reproduced the final text disappearing. Suppressing a genuine result is worse than displaying a cosmetic restatement, so exact whitespace-normalized matching stays.

The Opus 5 `output_config.effort` change landed as-is and checks out against Kiro's documentation. I also synced the completion semantics and effort sections into `ja`, `ko`, `ru`, and `zh-cn`, which had neither.

Verification on the integration commit: 10 new regression cases (stop-reason matrix, the positional bypass proven with `LENGTH_LIMIT`, legacy `finish_reason` truncation preserved, and `text_delta.phase` asserted rather than just the terminal); `tests/kiro-stream.test.ts` 81/81; 222/222 across kiro plus the bridge, Claude, and Chat downstream contracts; full suite 4360 passing; typecheck and privacy scan clean.

Thank you for this — the stop-reason handling was a real correctness gap, and the diagnosis held up even where the mechanism needed rework.

---

<sub>Integrated as part of a maintainer PABCD pass. Plan unit: `devlog/_plan/260726_pr_close_rework/030_wp3_kiro_stop_reason_460.md`.</sub>

Issue #404, opened 2026-07-24
Grok 4.5 behind a mixed OpenAI-compatible gateway returns reasoning-only on web_search; needs per-model adapter override


xiaoke949:
## Client or integration

Codex Desktop on Windows 11, using OpenCodex `/v1/responses`.

## Provider or upstream service

CLIProxyAPI-compatible gateway exposing both Gemini and xAI Grok 4.5 from one OpenAI-compatible base URL.

## OpenCodex version

Reproduced on `2.7.35`, then reproduced again after upgrading to `2.7.39`.

## Endpoint or capability

- OpenCodex inbound: `/v1/responses`
- Provider wire selected by OpenCodex: `/v1/chat/completions`
- Hosted `web_search`
- Streaming Responses events

## Current behaviour

OpenCodex is configured with one OpenAI-compatible provider using `adapter: "openai-chat"`. The gateway exposes both Gemini and Grok models.

- `localmodels/gemini-3.6-flash-high`: hosted `web_search` works and returns `web_search_call` plus a final message.
- `localmodels/grok-4.5`: the same Codex Responses request produces only a `reasoning` item followed by `response.completed`. There is no `web_search_call` and no assistant message, so Codex records a completed turn with an empty final response.

Direct controls against the same gateway isolate the protocol boundary:

1. `POST /v1/chat/completions` with `grok-4.5` and a function named `web_search` returns `finish_reason=stop`, empty content, and no usable `tool_calls`.
2. `POST /v1/responses` with `grok-4.5` and `tools:[{"type":"web_search"}]` succeeds, emits `web_search_call`, and returns a cited final message.
3. Gemini continues to work over Chat Completions.

The issue is therefore not authentication, model discovery, or the Codex client. A single OpenAI-compatible endpoint serves models that require different wire adapters, while OpenCodex currently selects `adapter` at provider scope.

## Minimal redacted reproduction

Provider configuration:

```json
{
  "providers": {
    "localmodels": {
      "adapter": "openai-chat",
      "baseUrl": "http://127.0.0.1:8317/v1",
      "allowPrivateNetwork": true,
      "apiKey": "${CLIPROXY_API_KEY}",
      "defaultModel": "grok-4.5"
    }
  }
}
```

Request to OpenCodex:

```bash
curl -N http://127.0.0.1:10100/v1/responses \
  -H "Content-Type: application/json" \
  -d '{
    "model":"localmodels/grok-4.5",
    "stream":true,
[...]

PR #464, opened 2026-07-25
fix(router): warn when a pinned provider discards a configured baseUrl

snowyukitty:
Refs #457 — the design-neutral part of that issue. This is the reporter's suggested fix 2, which #459 explicitly left "for maintainer discretion". It touches no registry entry, so it stands whichever way the Alibaba region question in #459 goes.

## Problem

A registry entry with a non-template `baseUrl` and no `allowBaseUrlOverride` outranks a saved provider `baseUrl` in `src/router.ts`. That resolution is intentional and is already covered by `tests/router-template-baseurl.test.ts`. What is not intentional is that it happens with no diagnostic at all.

| Config | Effective endpoint | What the user sees |
|---|---|---|
| `alibaba-token-plan` with a saved `…ap-southeast-1…` URL | `…cn-beijing…` from the registry | `401`, with nothing mentioning the discarded URL |

The saved value survives every round-trip through `config.json`, so the config keeps asserting a destination the proxy never uses. #457 spent most of its length reconstructing that by reading `router.ts` — which is the part worth fixing regardless of the rest of the issue.

## Fix

**`src/router.ts`** — warn once when a resolved user `baseUrl` is dropped in favour of a pinned registry endpoint.

- **Deduped** per `(provider, discarded, effective)` triple. `routedProviderConfig` runs per request, so an unguarded warn would repeat on every call; keying on the URLs rather than the provider alone means a config edited to a *different* wrong value warns again.
- **Redacted** through the existing `redactUrlForLog`, since a `baseUrl` can carry credentials in userinfo or a query string.
- **Silent when the URLs agree** modulo surrounding space and trailing slashes, using the same normalization as `matchBaseUrlChoice`. A stale `baseUrl` naming the endpoint the registry already pins is harmless and should not nag.
- **Warns rather than throws.** Routing is unchanged; a hard error here would break configs that route fine today.

## Tests

New `tests/router-discarded-baseurl-warning.test.ts` — 15 pass / 0 fail.

Reverting only `src/router.ts` fails 3 of the 15: the three asserting a warning is emitted (fires; fires once across repeated routing; redacts credentials). The other 12 assert *silence* and pass either way — they are controls pinning the no-false-positive surface rather than coverage of the fix: absent, empty, whitespace-only, unresolved-placeholder, exact-match, trailing-slash and surrounding-space `baseUrl`s; a non-string `baseUrl`; and the three honored cases (`allowBaseUrlOverride` provider, resolved registry template, provider absent from the registry).

- all 18 test files that import the router: **433 pass / 0 fail**
- `bun run typecheck`, `bun run privacy:scan`: pass
- `bun run lint:gui`: **could not run here** — it fails identically on a clean `dev` checkout on my machine (`ERR_UNKNOWN_FILE_EXTENSION` loading `gui/.eslint/i18n-file-groups.ts` under Node). Nothing in this PR touches `gui/`.
- full suite: **not run locally.** A single `bun test` hangs on this Windows machine partway through the anthropic e2e files, and that reproduces on a clean `dev` with no changes applied. CI is the real check here.

## Notes

- The warning fires in two existing cases in `tests/router-template-baseurl.test.ts` — `ollama-cloud` and `anthropic` keeping their fixed endpoints authoritative. That is exactly the intended surface; those assertions still pass, and the output shows up in the test log by design.
- A non-string `baseUrl` is still dropped silently by the existing `typeof` guard. That reads as a config-schema concern rather than this diagnostic's, so it is pinned by a test as a deliberate omission instead of being fixed here. Glad to fold it in if you would rather.
- @alexanderxcmo found this in #457. It is their issue, so if they would prefer to own this piece I am happy to close in favour of theirs.

## Summary by CodeRabbit

* **Improvements**
  * Added router diagnostics that warn when a configured `baseUrl` is discarded in favor of a provider’s fixed/pinned endpoint.
  * Warnings show the provider name and both the discarded and effective endpoints, with credentials/API keys safely redacted.
  * `baseUrl` is normalized (trim whitespace, remove trailing slashes) before comparison; warnings are deduplicated and never change routing behavior.
* **Documentation**
  * Updated `baseUrl` reference docs, including endpoint precedence and the “fixed provider endpoints” behavior.
* **Tests**
  * Added tests covering warning emission, redaction, deduplication, and scenarios where no warning should occur.

### DIFF
--- a/docs-site/src/content/docs/reference/configuration.md
+++ b/docs-site/src/content/docs/reference/configuration.md
@@ -146,7 +146,7 @@ network. Only do this on trusted networks, and always set a strong `OPENCODEX_AP
 | Field | Type | Meaning |
 | --- | --- | --- |
 | `adapter` | `string` | One of `openai-chat`, `openai-responses`, `anthropic`, `google`, `kiro`, `cursor`, `azure-openai` (or alias `azure`). |
-| `baseUrl` | `string` | Upstream API base URL. |
+| `baseUrl` | `string` | Upstream API base URL. Built-in providers with a fixed endpoint ignore it — see [Fixed provider endpoints](#fixed-provider-endpoints). |
 | `responsesPath?` | `string` | Optional relative resource path for key-auth `openai-responses` requests. It must start with `/` and contain no URL scheme, query, or fragment. When omitted, the adapter keeps its legacy `/v1/responses` URL construction. |
 | `disabled?` | `boolean` | Keep the provider on disk but exclude it from routing and model/catalog listings. |
 | `apiKey?` | `string` | API key, or an `${ENV_VAR}` / `$ENV_VAR` reference resolved at request time. |
@@ -192,6 +192,22 @@ network. Only do this on trusted networks, and always set a strong `OPENCODEX_AP
 | `desktopExecutor?` | `DesktopExecutorConfig` | **Cursor only.** External computer-use/record-screen commands; fields are listed below. |
 | `unsafeAllowNativeLocalExec?` | `boolean` | **Cursor adapter only.** Legacy compatibility boolean for the Cursor server-driven local `read` / `write` / `delete` / `ls` / `grep` / `shell` / `fetch` executor. Equivalent to `nativeLocalExec: "on"` when `nativeLocalExec` is unset; an explicit `nativeLocalExec` value always wins. Defaults to `false`. Prefer `nativeLocalExec` for new configs. See [Cursor provider](#cursor-provider-adapter-cursor) below. |
 | `nativeLocalExec?` | `"off" \| "codex-sandbox" \| "on"` | **Cursor adapter only.** Native local exec policy for the Cursor server-driven executor. `"off"` (default) rejects it; `"on"` is the trusted-local opt-in; `"codex-sandbox"` is accepted for backwards compatibility but is fail-closed like `"off"`. See [Cursor provider](#cursor-provider-adapter-cursor) below. |
+
+### Fixed provider endpoints
+
+Most built-in providers pin their own endpoint, and that pinned value wins over a `baseUrl` in
+your config. Only two kinds of entry honor a configured URL: the six that opt into an override —
+`ollama`, `vllm`, `lm-studio`, `litellm`, `qwen-cloud` and `alibaba-token-plan-intl` — and
+providers whose registry endpoint is a template you are expected to fill in, such as
+`azure-openai` and `cloudflare-ai-gateway`. Providers you define yourself always use their own
+`baseUrl`.
+
+When a configured `baseUrl` is discarded this way, opencodex logs a warning naming both URLs.
+Credentials in the URL are masked first. If you see it, either drop the `baseUrl` — the pinned
+endpoint is what the provider will use regardless — or switch to the provider whose endpoint
+matches the URL you wanted. Picking the right entry matters when a vendor runs one product in
+several regions: `alibaba-token-plan` is pinned to Beijing, while `alibaba-token-plan-intl`
+covers the international endpoints, and a key issued for one is rejected by the other.
 
 For broken `openai-responses` compatibility gateways, `responsesItemIdRepair` belongs on the
 provider object itself, for example:
--- a/docs-site/src/content/docs/reference/configuration.md
+++ b/docs-site/src/content/docs/reference/configuration.md
@@ -209,9 +209,11 @@ Adapters may adjust the resolved URL afterward. The `kiro` adapter, for example,
 region of the imported credential for a canonical `runtime.{region}.kiro.dev` host. See
 [Adapters](/reference/adapters/) for per-adapter rules.
 
-When routing discards a configured `baseUrl`, opencodex logs a warning naming both URLs, with any
-credentials in them masked. Either drop the `baseUrl` — the registry endpoint is what routing
-will use regardless — or switch to the provider whose endpoint matches the URL you wanted.
+When routing discards a configured `baseUrl`, opencodex logs a warning. It names the registry
+endpoint in full and your configured one by origin only, shown as `https://host/…` when it had a
+path — a configured path can itself be a credential, so none of it is logged. Either drop the
+`baseUrl` — the registry endpoint is what routing will use regardless — or switch to the provider
+whose endpoint matches the URL you wanted.
 Picking the right entry matters when a vendor runs one product in several regions:
 `alibaba-token-plan` is pinned to Beijing while `alibaba-token-plan-intl` covers the
 international endpoints, and a key issued for one is rejected by the other.
--- a/docs-site/src/content/docs/reference/configuration.md
+++ b/docs-site/src/content/docs/reference/configuration.md
@@ -205,7 +205,7 @@ keep the configured URL at this stage:
   `cloudflare-ai-gateway`;
 - providers you define yourself, which are not in the registry at all.
 
-Adapters may adjust the resolved URL afterwards. The `kiro` adapter, for example, follows the API
+Adapters may adjust the resolved URL afterward. The `kiro` adapter, for example, follows the API
 region of the imported credential for a canonical `runtime.{region}.kiro.dev` host. See
 [Adapters](/reference/adapters/) for per-adapter rules.
 
--- a/docs-site/src/content/docs/reference/configuration.md
+++ b/docs-site/src/content/docs/reference/configuration.md
@@ -195,19 +195,26 @@ network. Only do this on trusted networks, and always set a strong `OPENCODEX_AP
 
 ### Fixed provider endpoints
 
-Most built-in providers pin their own endpoint, and that pinned value wins over a `baseUrl` in
-your config. Only two kinds of entry honor a configured URL: the six that opt into an override —
-`ollama`, `vllm`, `lm-studio`, `litellm`, `qwen-cloud` and `alibaba-token-plan-intl` — and
-providers whose registry endpoint is a template you are expected to fill in, such as
-`azure-openai` and `cloudflare-ai-gateway`. Providers you define yourself always use their own
-`baseUrl`.
+Routing resolves a provider's endpoint before any adapter sees it, and for most built-in
+providers the registry's own endpoint wins over a `baseUrl` in your config. Three kinds of entry
+keep the configured URL at this stage:
 
-When a configured `baseUrl` is discarded this way, opencodex logs a warning naming both URLs.
-Credentials in the URL are masked first. If you see it, either drop the `baseUrl` — the pinned
-endpoint is what the provider will use regardless — or switch to the provider whose endpoint
-matches the URL you wanted. Picking the right entry matters when a vendor runs one product in
-several regions: `alibaba-token-plan` is pinned to Beijing, while `alibaba-token-plan-intl`
-covers the international endpoints, and a key issued for one is rejected by the other.
+- providers that opt into an override — `ollama`, `vllm`, `lm-studio`, `litellm`, `qwen-cloud`
+  and `alibaba-token-plan-intl`;
+- providers whose registry endpoint is a template you fill in, such as `azure-openai` and
+  `cloudflare-ai-gateway`;
+- providers you define yourself, which are not in the registry at all.
+
+Adapters may adjust the resolved URL afterwards. The `kiro` adapter, for example, follows the API
+region of the imported credential for a canonical `runtime.{region}.kiro.dev` host. See
+[Adapters](/reference/adapters/) for per-adapter rules.
+
+When routing discards a configured `baseUrl`, opencodex logs a warning naming both URLs, with any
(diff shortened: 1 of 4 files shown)


snowyukitty:
All three findings were fair. Fixed in `502b4fc5`.

**Path secrets (P1, `chatgpt-codex-connector`) — accepted.** Correct and the more serious of the two privacy points: `redactUrlForLog` clears userinfo, query and fragment, but its success branch returns `parsed.toString()` without running `redactSecretString`, so a key in a path segment survived intact. AGENTS.md is explicit that logging API keys is out, so this needed fixing rather than noting.

I did not take the "log only a sanitized origin" option, because the path is exactly what makes this warning useful — `…/v1` versus `…/compatible-mode/v1` is often the whole difference between two endpoints of the same vendor. Instead there is now a local `redactBaseUrlForLog` that layers `redactSecretString` over `redactUrlForLog`, so token-shaped path segments are masked while host and path stay readable. Covered by a test that puts `sk-…` in a path segment; it fails without the change.

Worth flagging separately: this is a property of the shared `redactUrlForLog`, so other call sites have the same gap. I have kept the fix local rather than changing the shared helper's behavior underneath every existing caller in a PR about something else — happy to open that as its own change if you want it.

**Raw URLs in the dedup Set (Major, CodeRabbit) — accepted, with a note.** The retention framing is a little stronger than the facts: the same credential is already resident for the process lifetime in the loaded provider config, so this was not a new disclosure. But the Set genuinely outlives the config that supplied it, and defense in depth here costs nothing.

The better argument for the change is the behavioral one in the second half of the finding, which I would not have thought of: keying on redacted URLs means rotating a key embedded in the URL no longer re-warns about an endpoint mismatch already reported. Done, and the suggested regression test is in — two URLs differing only in credentials now produce one warning.

**Docs sync (P2, `chatgpt-codex-connector`) — accepted.** `reference/configuration.md` now states that a pinned built-in endpoint wins over a configured `baseUrl`, names the six entries that opt into an override (`ollama`, `vllm`, `lm-studio`, `litellm`, `qwen-cloud`, `alibaba-token-plan-intl`) alongside the template-URL providers, and says what to do when the warning appears — including the region case that prompted #457, where a key issued for `alibaba-token-plan-intl` is rejected by the Beijing-pinned `alibaba-token-plan`.

English only. AGENTS.md asks that translated locales not contradict the English source; none of `ja`, `ko`, `ru` or `zh-cn` currently documents `baseUrl` override behavior, so there is nothing there to contradict. Say the word if you would rather I add the four translations in this PR.

**Re-verified on `502b4fc5`:** new suite 17 pass / 0 fail; the two added tests fail against the previous commit, so they bite. All 18 router-importing test files: 435 pass / 0 fail. `bun run typecheck` and `bun run privacy:scan` pass. The full-suite and `lint:gui` caveats in the PR description still stand — both fail on my machine on a clean `dev` checkout too.

snowyukitty:
Both docs findings landed in `955dff6a`. The first one was right in a way I had not seen, so thank you for it.

**Scoping the "Only two kinds" claim — accepted, and the `kiro` case is real.**

I checked the three providers named before changing anything, and they do not all say the same thing:

- `azure-openai` — its registry `baseUrl` is the template `https://{resource}.openai.azure.com/openai`, which the previous text already listed as an honored kind. No contradiction.
- `openai-responses` — an adapter, not a registry entry. `responsesPath` shapes the path *within* the resolved base URL; it does not make a pinned registry endpoint yield to a configured one. This one I do not think stands.
- `kiro` — this is the real finding. The registry entry is pinned to `https://runtime.us-east-1.kiro.dev` with no override, so routing does discard a configured URL. But `kiroRuntimeEndpoint` then re-derives the host from the imported credential's API region, which `reference/adapters.md:93-96` already documents. So my section was accurate about routing and misleading about where the request actually goes.

Two changes from that:

1. `reference/configuration.md` now scopes the rule to routing explicitly, lists the three kinds of entry that keep the configured URL as a list rather than an `Only` clause, and links to [Adapters](/reference/adapters/) for per-adapter rules instead of restating them.
2. The warning text no longer promises the destination. It said "Requests go to the fixed endpoint" — not something the router can honestly assert when an adapter may still change it. It now states what routing did and names the common cause of a 401:

   > `configured baseUrl <a> is ignored because this provider's endpoint is fixed at <b>. A URL saved for a different account or region is a common cause of 401s here — drop it, or use the provider whose endpoint matches.`

**Translated locales — done for `zh-cn`, not all four.**

I looked at how comparable changes have handled this rather than guessing: `a758bfcf` (#404) shipped English + `ko`, and `6a0f5b5d` (#416) shipped English + `zh-cn`. Neither updated all four, and the locale tables already lag the English one in places — `nativeLocalExec`, for instance, is missing from `zh-cn`. So the convention looks like English plus whichever locale the author can write well, rather than full parity per PR.

I have mirrored the section into `zh-cn/reference/configuration.md` and left `ja`, `ko` and `ru` alone rather than submit translations I cannot review to the same standard. Happy to add `ja` too if that is useful; I would rather someone who owns `ko` and `ru` write those.

**Verified on `955dff6a`:** router warning suite 17 pass / 0 fail, `router-template-baseurl` 26 pass / 0 fail, all 18 router-importing files 435 pass / 0 fail, `bun run typecheck` and `bun run privacy:scan` pass. CI was 9/9 green on the previous commit `502b4fc5`. The full-suite and `lint:gui` caveats from the PR description still apply — both fail here on a clean `dev` checkout too.

snowyukitty:
Fixed in `dcee581d`.

No strong precedent either way in the repo — neither spelling appears anywhere else under `docs-site/src/content/docs/`, and there is no Vale or LanguageTool config checked in, so this is your tooling's default rather than a house rule I was breaking. It is a one-word change with no downside though, so I have taken it rather than argue the point.

CI was 9/9 green on `955dff6a` before this commit, `mergeable_state=clean`.

lidge-jun:

lidge-jun:
Thank you for this — turning a silently discarded `baseUrl` into a visible warning addresses a real support burden, and the dedupe-on-redacted-form design is thoughtful. One defect in the redaction path blocks merge.

**High — opaque credentials embedded in URL paths are logged verbatim. `src/router.ts:127-166`.** A configured URL like `https://proxy.example/v1/8fK2mP7qR4nV6x` reaches `console.warn` intact: `redactUrlForLog` deliberately preserves the pathname, and `redactSecretString` is pattern-based, so a high-entropy token with no recognizable prefix matches none of the patterns at `src/lib/redact.ts:5-18,68-78`. The result is an API key or account-scoped route token written to process logs. The regression at `tests/router-discarded-baseurl-warning.test.ts:84-90` only covers an `sk-…` token, so it gives false confidence here.

The fix is to stop logging user-controlled path segments in this warning at all — log a sanitized origin, or replace each path segment with a non-reversible placeholder. If path comparison genuinely helps diagnosis, log a safe fingerprint of the path instead. Please add a regression using a high-entropy path credential with no recognizable prefix.

**Interaction with #459 — please read before either lands.** #459 proposes making `alibaba-token-plan` accept a `baseUrl` override and offering a Singapore choice under it. If that lands as written, this PR's warning stops firing for `alibaba-token-plan` entirely, and the guidance you added at `docs-site/src/content/docs/reference/configuration.md:202-217` becomes factually wrong: Beijing would no longer be pinned, and "switch to `alibaba-token-plan-intl`" would no longer describe routing behavior. The warning would be replaced by silent cross-region routing under mismatched registry metadata. My recommendation is to land this PR first (after the redaction fix) and rework #459 as an explicit stale-config migration; I've written that up on #459.

**Test gap:** beyond the opaque-credential regression above, please add a routing diagnostic test for the exact split that motivated this change — `alibaba-token-plan` pinned, `alibaba-token-plan-intl` honoring its configured choice. The current suite uses `anthropic` as the pinned fixture, so the Alibaba behavior this documents is not locked anywhere.

GitHub reports the head as mergeable with all checks green; that does not clear the logging defect for the reason described above.

---

<sub>Reviewed as part of a maintainer review pass against `dev` at `bef0dfbe`, at head `dcee581d`. All four changed files reviewed.</sub>

snowyukitty:
Thank you for the review — the High finding is correct, and it is the more important half of a mistake I made rather than a gap I knew about. Fixed in `ab9f01cf`, rebased onto `dev` at `7e2cee43`.

## The logging defect

You are right that pattern redaction was the wrong tool here, and right that my own regression hid it. `redactSecretString` matches *known token shapes*; an account-scoped route token like `https://proxy.example/v1/8fK2mP7qR4nV6x` is just an ordinary-looking path segment and matches nothing. I had tested with `sk-live-…`, which is precisely the case the patterns do catch — so the test passed for a reason unrelated to whether the value was safe. That is a control masquerading as coverage, and I should have caught it.

**The configured URL now has no path logged at all.** `configuredOriginForLog` returns `URL.origin` — which also excludes userinfo, query and fragment — plus a `/…` marker when a path was present, so an origin-only config stays distinguishable from one whose path was dropped. I took the "sanitized origin" option rather than a fingerprint: a fingerprint is one more thing to explain and get right, and the origin already carries the mismatch, which is what this warning is about.

**I kept the effective URL in full, deliberately.** Past the same-endpoint guard it is necessarily `registryEntry.baseUrl`: the caller passes the resolved URL, and whenever resolution kept the user's value the two compare equal and the function has already returned. So that side is a constant from this repo's registry, never user input — and it is the useful half, naming where requests actually go. Your wording scoped the defect to "user-controlled path segments", so I read this as consistent with what you asked for; say the word if you would rather both sides were origin-only.

The dedup key now derives from the logged forms, so it also holds no raw path. It is coarser — two different bad paths on one host warn once — which seems like the right grain, since the host is the mismatch.

**New regressions**, both failing without the source change:

- an opaque high-entropy path credential with no recognizable prefix, asserting the token is absent and the origin marker present;
- a multi-segment path (`/tenant-42/v1/route/Zx9Qw`) asserting *no* segment survives, so this cannot regress into "we mask the segment that looks like a secret".

## Alibaba routing coverage

Added, and you were right that `anthropic` as the only pinned fixture left the behavior this documents unlocked:

- `alibaba-token-plan` with a saved Singapore URL — warns, and `routeModel` resolves to the Beijing endpoint;
- `alibaba-token-plan-intl` with the same Singapore URL — silent, and resolves to it;
- `alibaba-token-plan-intl` with the pay-as-you-go choice — silent, and resolves to it.

To be explicit about what these prove: they pass with and without this commit. They are regression locks on the registry split, not activation proof for the redaction fix — I have labelled them that way in the commit message too, so nobody later reads them as evidence for the wrong thing.

## On #459

Agreed, and thank you for writing up the metadata argument — the endpoint-moves-but-catalog-stays-Beijing failure is sharper than what I had raised there, and it is the part that makes an override genuinely unable to express this. Happy for this PR to land first on that basis. If a stale-config migration is the shape you want for #459 and its author would rather not take it, I am willing to write it, but it is their finding and I would not want to step in front of them.

## Verification on `ab9f01cf`

- `tests/router-discarded-baseurl-warning.test.ts`: 22 pass / 0 fail; reverting only `src/router.ts` fails 3 of them
- all 20 router-importing test files on current `dev`: 479 pass / 0 fail
- `bun run typecheck`, `bun run privacy:scan`: pass
- rebased onto `7e2cee43`, no conflicts; `src/router.ts` had moved under #431 but in a different hunk
- unchanged caveats: I cannot run the full suite or `lint:gui` here — both fail on a clean `dev` checkout on this machine, so hosted CI remains the real check

lidge-jun:
Integrated into `dev` as `1540ad4a`, with your authorship preserved via a `Co-authored-by` trailer. The patch applied cleanly and landed verbatim — nothing was rewritten.

Verification on the integrated tree: `bun x tsc --noEmit` exit 0, `tests/router-discarded-baseurl-warning.test.ts` + `tests/router.test.ts` + `tests/cyber-policy-error-fidelity.test.ts` 54 pass / 0 fail, `bun run privacy:scan` passed, and the full suite is green at 4504 pass / 0 fail.

Two details in this PR are worth calling out, because they are the reason it was safe to land directly rather than being held for security review:

`configuredOriginForLog` logs `URL.origin` and withholds the path. Your comment gives the right reason — a configured `baseUrl` path can itself be the credential, and an opaque account-scoped route token matches none of the prefix patterns in `redactSecretString`, so pattern redaction cannot be trusted for that value. The `…/…` marker preserves the diagnostic signal without leaking a segment, and the non-special-scheme and parse-failure branches both fall back safely.

The dedupe `Set` is keyed off the *logged* forms rather than the raw URLs. A naive implementation would have parked a credential-bearing string in a process-lifetime cache; keying on the redacted form also means rotating a key embedded in the URL does not re-warn about the same endpoint mismatch.

The warn-rather-than-throw choice is also correct: the effective route is unchanged, so a hard error would break configs that route fine today.

Thanks for the careful work, and for documenting the reasoning inline where a reviewer would look for it.

lidge-jun:
Closing as integrated — the change is on `dev` at 1540ad4a with your authorship preserved. Reopen if anything looks off in how it landed.

PR #466, opened 2026-07-25
fix(gui): react-doctor shared foundations

Wibias:
## Summary
- Add `fetch-json`, `client-resource`, and `intl-formatters` helpers used by follow-up react-doctor cleanup PRs.
- Store the GUI API auth token in memory only (no `sessionStorage`) to avoid XSS-readable persistence.

## Stack
- **Depends on #465** (`chore(gui): pin react-doctor to 0.9.1`). This branch includes that commit until #465 merges; please review/merge #465 first, then rebase this if needed.

## Test plan
- [ ] Smoke: open GUI, enter API token when prompted, confirm authenticated `/api/*` calls still work across reload (token must be re-entered — intentional)

## Summary by CodeRabbit

* **New Features**
  * Added reusable client-side resource caching with refresh, optional polling, and per-key loading/error state.
  * Introduced helpers for safe JSON response handling and new i18n formatters for credit dates and estimated USD values.
* **Bug Fixes**
  * Improved API authentication to prefer in-memory tokens, defensively clear legacy session entries, and retry once after auth failures without leaking legacy tokens.
* **Tests**
  * Added/updated Bun and React polling tests to validate token prompt behavior and unmount/in-flight request correctness.

### DIFF
--- a/gui/src/api.ts
+++ b/gui/src/api.ts
@@ -1,18 +1,16 @@
 let installed = false;
 let promptInFlight: Promise<string | null> | null = null;
 
-function apiPath(input: RequestInfo | URL): string | null {
+function needsApiAuth(input: RequestInfo | URL): boolean {
   try {
     const raw = input instanceof Request ? input.url : String(input);
-    return new URL(raw, window.location.href).pathname;
+    const url = new URL(raw, window.location.href);
+    // Absolute cross-origin URLs must never get the local API token or 401 prompt.
+    if (url.origin !== window.location.origin) return false;
+    return url.pathname.startsWith("/api/") || url.pathname.startsWith("/v1/");
   } catch {
-    return null;
+    return false;
   }
-}
-
-function needsApiAuth(input: RequestInfo | URL): boolean {
-  const path = apiPath(input);
-  return !!path && (path.startsWith("/api/") || path.startsWith("/v1/"));
 }
 
 /** Legacy sessionStorage key from pre-memory auth — wiped once on install, never read. */
--- a/gui/src/api.ts
+++ b/gui/src/api.ts
@@ -15,6 +15,9 @@ function needsApiAuth(input: RequestInfo | URL): boolean {
   return !!path && (path.startsWith("/api/") || path.startsWith("/v1/"));
 }
 
+/** Legacy sessionStorage key from pre-memory auth — wiped once on install, never read. */
+const LEGACY_TOKEN_KEY = "opencodex-api-token";
+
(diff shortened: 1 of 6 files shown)


lidge-jun:

lidge-jun:
Thank you for extracting the shared react-doctor foundations — the client-resource store is the right shape for this codebase, and the exact head verifies cleanly (`bun run typecheck` and `bun run lint:gui` both pass locally, matrix CI green).

One substantive finding before merge.

**`gui/src/client-resource.ts:143` — declared dependency invalidation is silently discarded.** `useKeyedClientResource` advertises explicit dependency-based invalidation, but `_deps` is dropped at line 153. If a loader input changes while `key` stays stable, the hook neither resubscribes nor refetches, so consumers keep rendering stale data until polling ticks or the user forces a refresh. Please either fold the dependencies into resource identity/invalidation, or remove the parameter entirely and require callers to encode every dependency in `key`. A parameter that looks load-bearing but is inert is the more dangerous of the two options.

**Overlap with `dev`:** none. No touched path has changed on `dev` since this PR's `c03ed0a3` base, and three-way merge inspection is clean. No rebase required.

**Test gap:** the new resource store has no tests at all. Please add hook/store regressions for dependency changes, key changes, disabled-to-enabled transitions, polling cleanup, abort/latest-result ordering, and falsy resource values. The `_deps` case should fail before the fix and pass after it.

---

<sub>Reviewed as part of a maintainer review pass against `dev` at `bef0dfbe`. Verification: exact head, fresh local `bun run typecheck` + `bun run lint:gui`, plus hosted matrix CI.</sub>

Wibias:
Addressed Codex / CodeRabbit / maintainer review on `client-resource` + `fetch-json`:

- **`useKeyedClientResource` deps** — element-wise invalidation now triggers `refresh()` when deps change (array identity ignored)
- **Slow polls** — interval ticks skip while a request is in flight (no abort storm)
- **Falsy cache** — loading gate uses `data === undefined` so `false`/`0`/`null`/""` stay cached
- **`setClientResourceData`** — aborts in-flight + bumps generation so late responses cannot stomp
- **Multi-subscriber polling** — tracks per-listener `pollMs`, uses `Math.min`, recomputes on subscribe/unsubscribe
- **Store eviction** — deletes cache entry when the last subscriber leaves
- **`fetch-json`** — 204 / empty-body success no longer throws `SyntaxError`

Type-safety of the string-keyed store map left as a documented API contract (trivial / no runtime check).

PR #467, opened 2026-07-25
refactor(gui): split Combos for react-doctor

Wibias:
## Summary
- Split `ComboWorkspace` into focused panel/modal modules.
- Apply react-doctor 0.9.1 fetch status checks, loading `finally` resets, and numeric input guards on the Combos surface.

## Stack
- Depends on #465 (pin) and #466 (foundations). Branch includes those commits until they merge.

## Test plan
- [x] Open Combos page: create/edit/delete combo, confirm save/error toasts still work

## Summary by CodeRabbit

* **New Features**
  * Added a streamlined combo workspace for creating, editing, and removing combo configurations.
  * Added strategy selection, effort settings, target management, drag-and-drop ordering, and provider/model filtering.
  * Added overview counts and attention indicators for combos requiring review.
  * Added confirmations for removal and unsaved changes.

* **Bug Fixes**
  * Improved handling of unsuccessful or invalid save and removal responses.

* **Tests**
  * Added coverage for unsaved-change navigation and stable target handling.


lidge-jun:

lidge-jun:
Thanks for splitting Combos — the extraction is a genuine readability win. Two things block merge, one of them a red CI job.

**1. CI blocker — `gui/src/combo-workspace-data.ts:117` changes the parser's output contract.** `parseComboList` now attaches an enumerable `clientKey` to every parsed target, which breaks `tests/combo-workspace-data.test.ts:83`. This reproduces locally (16 pass, 1 fail) and fails on all three CI operating systems. Please decide deliberately whether `clientKey` belongs in the parsed view model: either keep the previous plain-object contract and derive stable keys in the component layer, or keep it and update the test to assert the stable-UI-key contract *while also* proving `toPutBody` never serializes `clientKey` upstream. Required CI has to be green before this can land.

**2. `gui/src/components/combo-workspace-add-modal.tsx:78` — backdrop dismissal was lost.** Converting the add modal from a clickable overlay to a native `<dialog>` dropped outside-click dismissal. The old overlay closed on a click outside the card; the new full-screen dialog has neither a scrim nor a dialog-level click handler, which is also inconsistent with the remove and unsaved dialogs you extracted in the same PR. Add a scrim, or handle clicks where `event.target === event.currentTarget`, keeping the existing `busy` guard.

**Overlap with `dev`:** none. No touched path has changed on `dev` since `c03ed0a3`, and three-way merge inspection finds no conflict, so no rebase is needed. The blocker is the failing test, not the merge base.

**Test gap:** there is no rendered Combos interaction test covering create/edit/delete, unsaved-navigation protection, Escape and backdrop dismissal, target add/remove/reorder, or numeric input behavior. Given the size of this split, that coverage is what would make the refactor verifiably behavior-preserving.

---

<sub>Reviewed as part of a maintainer review pass against `dev` at `bef0dfbe`. Verification: exact head, fresh local `bun run typecheck` + `bun run lint:gui` (both pass) and the focused test suite (1 failure, reproduced on hosted CI across macOS/Linux/Windows).</sub>

Wibias:
Addressed the review blockers on the latest push (`76a0fc13`):

1. **`clientKey` parser contract** — kept the UI-only key on the parsed view model; updated `tests/combo-workspace-data.test.ts` to assert stable `ct-*` keys and uniqueness, and added an explicit `toPutBody` check that `clientKey` is never serialized upstream.
2. **Backdrop dismissal** — remove/unsaved/add dialogs now dismiss when `event.target === event.currentTarget` (add modal keeps the `busy` guard via `requestClose`).
3. **Rebased onto `dev`** — dropped the bundled #466 foundations commits now that #466 has merged (also clears the Codex note about token-storage landing via this Combos PR).

CodeRabbit had no actionable inline findings on this PR (rate-limit / empty incremental pass). The broader Combos interaction-test gap noted in review is still open if you want that as a follow-up.

PR #468, opened 2026-07-25
refactor(gui): react-doctor cleanup for Startup/Debug/Storage/Usage

Wibias:
## Summary
- Split Startup and Debug page modules for react-doctor maintainability.
- Harden Storage/Usage loading resets and related fetch status handling for react-doctor 0.9.1.

## Stack
- Depends on #465 and #466.

## Test plan
- [ ] Smoke Startup, Debug, Storage, and Usage pages

### DIFF
--- a/gui/src/pages/Debug.tsx
+++ b/gui/src/pages/Debug.tsx
@@ -1,6 +1,6 @@
 import { useCallback, useEffect, useEffectEvent, useRef, useState } from "react";
 import { useVirtualizer } from "@tanstack/react-virtual";
-import { useKeyedClientResource } from "../client-resource";
+import { setClientResourceData, useKeyedClientResource } from "../client-resource";
 import { useI18n } from "../i18n/shared";
 import { DebugClaudeInboundPanel } from "./debug-claude-inbound-panel";
 import { DebugLogViewer } from "./debug-log-viewer";
@@ -12,17 +12,23 @@ import {
   isStreamEnabled,
 } from "./debug-shared";
 
+function debugSettingsKey(apiBase: string): string {
+  return `debug-settings:${apiBase}`;
+}
+
 export default function Debug({ apiBase, embedded }: { apiBase: string; embedded?: boolean }) {
   const { t } = useI18n();
+  const [debugBusy, setDebugBusy] = useState(false);
   const [stream, setStream] = useState<LogStream>("provider");
   const [entries, setEntries] = useState<import("./debug-shared").DebugLogEntry[]>([]);
   const [follow, setFollow] = useState(true);
   const [refreshing, setRefreshing] = useState(false);
   const afterRef = useRef(0);
+  const mutationGenerationRef = useRef(0);
   const scrollContainerRef = useRef<HTMLDivElement>(null);
 
   const debugPoll = useKeyedClientResource(
-    `debug-settings:${apiBase}`,
+    debugSettingsKey(apiBase),
     [apiBase],
     async (signal) => {
       const res = await fetch(`${apiBase}/api/debug`, { signal });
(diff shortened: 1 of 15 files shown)


lidge-jun:

lidge-jun:
Thanks for the react-doctor cleanup across Startup/Debug/Storage/Usage. Typecheck, lint, and matrix CI are green on the exact head, but this is labelled a refactor and three of the changes are behavior-changing rather than behavior-preserving. Those are what I'd like resolved before merge.

**1. Aborted requests now clear shared loading state unconditionally — `gui/src/pages/Startup.tsx:113`, `gui/src/pages/Storage.tsx:130`, `gui/src/pages/Usage.tsx:567`.** When `apiBase`, the range, or the surface changes, the superseded request's `finally` can run *after* the replacement request has started, setting `loading=false` while the newer request is still in flight. The UI then presents stale data as settled and re-enables controls early. Please restore latest-request ownership with a generation or request ID, and clear loading only when the completing request is still the current one. The abort guard shouldn't be weakened just to satisfy the linter.

**2. `gui/src/pages/Debug.tsx:125` and `:161` — mutation busy state was removed.** `debugBusy` is gone and hard-coded to `false`, so the switches and Reset stay enabled during in-flight PUTs. Rapid toggling or a toggle-plus-reset then resolves to whichever response lands last, which is a real correctness problem for a debug-flag surface. Restore the mutation state, disable controls while a mutation is pending, and refresh from the accepted response or after the latest mutation settles.

**3. `gui/src/pages/debug-claude-inbound-panel.tsx:30` — the new row key is not unique.** The key `${entry.at}:${entry.endpoint}:${entry.model}` collides because the producer stamps `Date.now()` at `src/claude/inbound-debug.ts:85`, and several same-model requests can be captured within one millisecond. Duplicate keys give React incorrect reconciliation and visibly stale rows. Add a monotonic sequence ID to captured entries and key on it.

**Overlap with `dev`:** `gui/src/pages/Usage.tsx` is the only touched path that also moved on current `dev` (later Usage layout/accessibility work). Three-way merge inspection is clean so a rebase isn't strictly required, but please revalidate the merged Usage page so both sets of changes survive.

**Test gap:** add component-level regressions for (a) an aborted request completing after its replacement starts, (b) rapid Debug flag/reset interactions, and (c) multiple Claude inbound entries sharing a timestamp, endpoint, and model. The current green typecheck/lint/matrix runs do not exercise any of these timing or reconciliation failures — which is exactly why the refactor reads as safe today.

---

<sub>Reviewed as part of a maintainer review pass against `dev` at `bef0dfbe`. Verification: exact head, fresh local `bun run typecheck` + `bun run lint:gui`, plus hosted matrix CI.</sub>

Wibias:
## Review follow-up (`46bef6cf`)

Rebased onto current `upstream/dev`, dropping the bundled foundations commit (`9c7e922e`) now that #466 is merged.

### Fixed (Jun)
1. **Startup / Storage / Usage loading races** — generation/request ownership so a superseded aborted request cannot clear `loading` while a newer fetch is in flight.
2. **Debug mutation busy** — restored `debugBusy` + mutation generation; switches/Reset disable while PUT pending; apply accepted response via `setClientResourceData` for the latest mutation only.
3. **Claude inbound row keys** — monotonic `id` on capture entries; panel keys on `entry.id`.

### Tests
- `gui/tests/storage-loading-race.test.tsx`
- `gui/tests/debug-mutation-busy.test.tsx`
- `gui/tests/debug-claude-inbound-keys.test.tsx`
- extended `tests/claude-inbound-debug.test.ts`

### Skipped
- Codex comments on `gui/src/api.ts` / `client-resource` foundations — those land via #466 after the rebase; out of scope for this PR.

Verification locally: focused GUI + inbound tests, `bun x tsc -b`, `bun run lint` in `gui/`.

lidge-jun:
Post-merge note: I put this through an independent review pass after it landed, and the five production fixes all hold up. Startup, Storage, and Usage each own loading through a complete generation sequence (increment, result guard, current-generation `finally`, pre-abort invalidation), every Debug mutation control is gated by `debugBusy` with the PUT response installed before busy clears, and `nextCaptureId` stays unique across ring eviction and debug toggles. The extraction is behavior-preserving — conditionals, disabled states, ARIA roles, and i18n keys all survived.

Your `entry.id` approach is better than what I had planned. I was going to key on `${entry.at}-${index}`, which still shifts when the ring reorders; a producer-side monotonic id has no such problem.

Three test-quality gaps did turn up, and I've closed them in [`bcaf029e`](https://github.com/lidge-jun/opencodex/commit/bcaf029e) rather than sending this back for another round — they're the kind of hardening a maintainer can write, and the implementation itself was done.

**`debug-claude-inbound-keys` passed against the defect it was written to catch.** It asserts row content only, and with the old colliding composite key React still renders both rows correctly while merely warning on the console. Reproduced it directly. It now captures `console.error` and asserts the duplicate-key warning is absent.

**`claude-inbound-debug` never reached `RING_LIMIT`.** Three captures don't exercise eviction. Added a 25-capture case asserting ids stay unique and descending afterwards — a counter reset would resurrect a discarded id and bring the collision straight back.

**Startup and Usage had no race regression.** `storage-loading-race` is a genuine one, but the other two carry the same guard with nothing protecting it. Added equivalent delayed-settlement tests and verified they go red by reverting each guard individually.

**Neither Debug test proved response-install ordering.** Both flows end at the same state they start from, so removing `setClientResourceData()` still satisfies their final assertions. Added a `false → true` transition where every GET keeps returning the pre-mutation state, so the only way the switch can read `true` is the PUT response itself. It's in a separate file because sharing a fetch stub with the serialization tests perturbs them.

Verification: GUI suite 178/178, the five affected files 8/8, backend batches clean, typecheck / `lint:gui` / privacy scan all green.

One governance note for the record, not a criticism of the change: GitHub shows this merged by its own author with no recorded review decision, and `MAINTAINERS.md` asks for maintainer approval with no self-approval. Treat this review plus the follow-up evidence as the post-merge gate.

---

<sub>Reviewed as part of a maintainer PABCD pass against `dev` at `4980410e`. Plan unit: `devlog/_plan/260726_pr_close_rework/050_wp5_gui_pages_cleanup_468.md`.</sub>

PR #478, opened 2026-07-26
fix(xai): declare grok image input so combos are not advertised text-only

jonathanli12:
## Problem

Any combo containing an xAI target is published to Codex as **text-only**, so the Codex app refuses image attachments ("does not support image inputs") before a request is ever routed, even when every member of the combo accepts images.

Reproduced with the documented xAI -> Cursor `grok-4.5` failover combo. From the generated Codex catalog:

```
combo/xai_grok_fallback      input_modalities: ["text"]      <- blocked
cursor/grok-4.5              input_modalities: ["text","image"]
xai/grok-4.5                 input_modalities: undefined     <- root cause
```

## Cause

The `xai` entry in `PROVIDER_REGISTRY` declares no `modelInputModalities`, so catalog rows for `xai/grok-4.5` carry `inputModalities: undefined`.

`deriveComboCatalogModel` intersects member modalities and defaults an undefined member to `["text"]`:

```ts
const inputModalities = intersectStrings(
  members.map(member => member.inputModalities ?? ["text"]),
);
```

So one member with unknown modalities collapses the whole combo to text-only.

There is an ironic asymmetry today: `cursor/grok-4.5` ends up correct because Cursor lists every model in `noVisionModels`, and `applyProviderConfigHints` deliberately appends `image` for sidecar-covered models (that code already documents why the catalog must still advertise image input, otherwise the Codex app blocks attachments client-side). xAI is neither sidecar-covered nor vision-declared, so it falls through both paths.

## Fix

Declare the vision lineup for the grok-4.x chat models, per the xAI docs on image understanding:

```ts
modelInputModalities: {
  "grok-4.5": ["text", "image"],
  "grok-4.3": ["text", "image"],
  "grok-4.20-0309-reasoning": ["text", "image"],
  "grok-4.20-0309-non-reasoning": ["text", "image"],
},
```

`grok-build-0.1` and `grok-composer-2.5-fast` are intentionally omitted; they stay in `noVisionModels`.

## Notes

- Intersection semantics are unchanged. A member whose modalities are genuinely unknown still narrows a combo to `["text"]`, which is the safe behaviour and is now pinned by a test.
- No change to request-time handling. Combos still resolve to a concrete member via `routeModel`, so per-member vision protection (`planVisionSidecar` / `stripImagesInPlace` fail-closed) is untouched.

## Tests

Added to `tests/catalog-vision-sidecar-modalities.test.ts`:

1. xAI grok chat models declare image input in the registry (fails without this fix).
2. A combo of two vision-capable members still advertises image.
3. A member with unknown modalities still narrows the combo to text (guards the safety property).

Verified locally: 637 tests across the 45 catalog/combo/vision/provider/registry test files pass, and `tsc --noEmit` is clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enabled image attachments for supported xAI Grok model variants.
  * Updated combined/derived model catalog entries to correctly advertise image input support when available.
  * Ensured combined entries fall back to text-only when a member’s vision support is missing or unverified.
* **Tests**
  * Added/expanded automated coverage to verify modality propagation and correct narrowing behavior for combined models.

### DIFF
--- a/src/providers/derive.ts
+++ b/src/providers/derive.ts
@@ -78,6 +78,23 @@ function cloneRecordOfArrays(input: Record<string, string[]>): Record<string, st
   return Object.fromEntries(Object.entries(input).map(([key, value]) => [key, [...value]]));
 }
 
+/**
+ * Fill registry defaults BENEATH the user's per-model entries.
+ *
+ * Capability maps are keyed per model, so an all-or-nothing fill lets a single
+ * customized model hide the registry's knowledge about every other model. That
+ * is how a partially customized `modelInputModalities` could leave a
+ * vision-capable model advertising no image support, which in turn collapses any
+ * combo containing it to text-only. Routing already merges these maps per key
+ * (`mergeRecordFill` in src/router.ts); catalog enrichment now matches.
+ */
+function fillRecordOfArrays(
+  seed: Record<string, string[]>,
+  user: Record<string, string[]> | undefined,
+): Record<string, string[]> {
+  return { ...cloneRecordOfArrays(seed), ...(user ? cloneRecordOfArrays(user) : {}) };
+}
+
 function cloneNestedRecord(input: Record<string, Record<string, string>>): Record<string, Record<string, string>> {
   return Object.fromEntries(Object.entries(input).map(([key, value]) => [key, { ...value }]));
 }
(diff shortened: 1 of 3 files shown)


jonathanli12:
Thanks, both addressed in 9eb7f0b3.

**Test coverage (trivial):** the registry assertion now loops all four vision-capable grok ids and both text-only exclusions (`grok-build-0.1`, `grok-composer-2.5-fast`), so a regression in any single entry fails.

**P2 partial-map merge:** confirmed reproducible, and worth fixing. With a persisted config containing `modelInputModalities: { "grok-4.3": ["text"] }`:

```
before   enrich -> {"grok-4.3":["text"]}
         xai/grok-4.5 modalities: undefined      <- registry map skipped entirely
after    enrich -> {"grok-4.5":["text","image"], "grok-4.3":["text"], ...}
         xai/grok-4.5 modalities: ["text","image"]
```

`enrichProviderFromRegistry` filled `modelInputModalities` all-or-nothing, so one customized model hid the registry's knowledge about every other model and put the combo aggregator back on its `["text"]` default. Since these maps are keyed per model, the fill is now per key, matching what `mergeRecordFill` already does for the same field in `src/router.ts`. Explicit user values still win per model (the `grok-4.3: ["text"]` narrowing above is preserved).

Scoped deliberately to `modelInputModalities` rather than converting every fill in `derive.ts`: that broader change would alter behaviour for many unrelated capability maps and belongs in its own PR.

Added a regression test covering the persisted partial map through `enrichProviderFromRegistry` + `applyProviderConfigHints`.

Verified: `tsc --noEmit` clean, and 858 tests across the 60 catalog/combo/vision/provider/registry/router/model test files pass.

Ingwannu (approved):
Reviewed at exact head `9eb7f0b36cbb10a87551f5b1e0324d77dcfacb19`. xAI official model pages advertise Text+Image input for all four declared Grok models. The registry change fixes the real combo-catalog collapse without weakening unknown-member intersection semantics, and the per-model fill preserves explicit user overrides while restoring defaults only for missing keys. Verified the 12 focused modality/vision/combo regressions and typecheck locally. The unrelated Windows GUI flake passed on rerun; all current required checks are green.

PR #489, opened 2026-07-26
fix(responses): strip max_output_tokens and metadata in forward mode

elppaaa:
## Summary

The ChatGPT backend (`authMode: "forward"`) enforces a strict parameter allowlist and rejects unknown top-level fields with HTTP 400 `{"detail":"Unsupported parameter: …"}`.

The `stripPreviousResponseId` doc-comment (line 452) already documents that the backend also rejects `metadata` and `max_output_tokens`, but only `previous_response_id` was actually stripped. Third-party Responses API clients (e.g. GJC, SDK wrappers) send `max_output_tokens` per the public OpenAI `/v1/responses` spec, causing **every** request through a forward-mode provider to fail with 400.

Codex CLI never sends these fields — it controls output length via `reasoning.effort` — so the strip is a no-op for the primary client. Claude Code uses the `anthropic.ts` adapter and is unaffected. Key-auth mode preserves both fields (the platform API supports them).

**Change:** add `stripUnsupportedForwardParams()` inside the existing `if (forward)` block in `buildRequest()`, removing both `max_output_tokens` and `metadata` before the wire request is serialized.

## Verification

- `bun test tests/openai-responses-passthrough.test.ts` — 29/29 pass (3 new tests)
- `bun run typecheck` — clean
- Live reproduction: the exact request shape that returned 400 (`max_output_tokens: 32000`, `reasoning.effort: "low"`, `store: false`) now returns HTTP 200 with `response.completed` through a forward-mode provider
- New tests cover: forward strips both fields, forward no-op when absent, key-auth preserves both fields

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved forwarding compatibility by removing unsupported `max_output_tokens` and `metadata` request parameters before submission.
  * Preserved supported request fields and ensured parameters remain unchanged when using key-based authentication.

* **Tests**
  * Added coverage for parameter removal, no-op behavior when fields are absent, and preservation in key-authenticated requests.

### DIFF
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -462,6 +462,23 @@ function stripPreviousResponseId(body: unknown, strip: boolean): unknown {
 }
 
 /**
+ * Remove top-level parameters the ChatGPT backend (`authMode: "forward"`) rejects
+ * with `{"detail":"Unsupported parameter: …"}` (strict allowlist). Codex CLI never
+ * sends these — it controls output length via `reasoning.effort` — but third-party
+ * Responses API clients (GJC, SDK wrappers) include `max_output_tokens` per the
+ * public spec. `metadata` is likewise absent from the allowlist. No-op when the
+ * body carries neither field, keeping the common Codex path allocation-free.
+ */
+function stripUnsupportedForwardParams(body: unknown): unknown {
+  if (!isPlainObject(body)) return body;
+  const hasMot = Object.prototype.hasOwnProperty.call(body, "max_output_tokens");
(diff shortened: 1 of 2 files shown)


Ingwannu (approved):
Reviewed exact head 112b8c61. The change is correctly limited to forward-mode Responses requests; persisted configuration reserves that mode for the canonical built-in ChatGPT backend, while key-auth providers retain max_output_tokens and metadata. The sanitizer is top-level only, allocation-free when absent, and covered by forward-strip/no-op/key-preserve regressions. All 29 focused tests and TypeScript typecheck pass locally. Approved pending required CI.

Issue #83, opened 2026-07-10
Built-in image generation returns `404 Unknown endpoint`


Yeondaeno:
Summary

The built-in image generation tool fails before producing an image. Every observed attempt returned an immediate `404 Not Found` response identifying `POST /v1/images/generations` as an unknown endpoint.

## Environment

- Product surface: Codex Desktop on Windows
- Codex CLI: `codex-cli 0.144.1`
- OS: Microsoft Windows `10.0.26100.8655`
- LazyCodex: auto-updated from `v4.13.0` to `v4.16.2` during the session
- LazyCodex bootstrap: running in the background; a fresh-session confirmation is still required after bootstrap completes
- Workspace: `C:\개인 홈페이지`
- Image path requested: built-in `image_gen` tool
- API fallback: not used
- `OPENAI_API_KEY`: not configured

## Minimal Reproduction

1. Open Codex Desktop on Windows.
2. Start a task in any local workspace.
3. Ask Codex to generate an image using the built-in image generation feature, explicitly without the API/CLI fallback. For example:

   ```text
   Generate a dark Gothic editorial hero graphic with indigo and gold inks,
   halftone texture, no text, and no watermark. Use the built-in image
   generation tool only.
   ```

4. Allow Codex to invoke the built-in `image_gen.imagegen` tool with the prompt.
5. Observe that the tool fails immediately instead of returning an image.
6. Repeat with a materially different prompt. The same endpoint error is returned.

## Exact Observed Error

```text
image generation failed: http 404 Not Found: Some("{\"error\":{\"message\":\"Unknown endpoint: POST /v1/images/generations\",\"type\":\"not_found\",\"code\":\"not_found\"}}")
```

## Expected Behavior

The built-in image generation tool should accept the prompt and return a generated image without requiring a locally configured `OPENAI_API_KEY`.

## Actual Behavior

- The request fails in less than one second.
- No image or project asset is created.
- The error is independent of the requested composition or palette across the observed attempts.
- The CLI/API fallback is not invoked.

## Runtime Evidence

- The identical `404 Unknown endpoint` response was observed on repeated built-in image-generation calls.
- The response names the rejected route as `POST /v1/images/generations`.
- The failure occurs before any image result, output path, or asset metadata is returned.
- A local API key check returned `OPENAI_API_KEY=missing`, confirming that the paid CLI fallback was not involved.

## Hypotheses

[...]

PR #492, opened 2026-07-26
fix(openai-responses): handle image tool conflicts in additional_tools

RailyW:
## Summary

- detect hosted-tool namespace conflicts across both top-level `body.tools` and Responses Lite `input[].type = "additional_tools"` tool groups
- remove conflicting hosted `image_generation` entries from every tool group on the API-key `openai-responses` path
- preserve unrelated tools, unknown entries, message input items, and ChatGPT forward-mode behavior

## Problem

Codex Desktop can send its client-side `image_gen` namespace in a nested Responses Lite item:

```json
{
  "input": [
    {
      "type": "additional_tools",
      "tools": [
        { "type": "namespace", "name": "image_gen", "tools": [] },
        { "type": "image_generation" }
      ]
    }
  ]
}
```

The platform validates these declarations as one tool namespace and rejects the request because the user-defined `image_gen` namespace collides with the hosted image-generation tool.

PR #90 added the correct API-key-path policy for the top-level `body.tools` shape, but the sanitizer returned early when no top-level array existed. Responses Lite requests therefore still reached the platform with the conflicting nested pair.

## Implementation

`stripConflictingHostedTools` now:

1. Collects top-level and nested `additional_tools` arrays.
2. Detects client namespace declarations across the combined request, including both `image_gen` namespace tools and flat `image_gen.*` function names.
3. Removes only the matching hosted tool type from every collected group.
4. Uses copy-on-write behavior so unchanged groups and requests preserve their original references.

The sanitizer remains limited to the API-key path. ChatGPT forward mode continues to preserve the hosted and client declarations because that backend accepts the pair and owns native image generation.

## Tests

- `bun test --isolate tests/openai-responses-passthrough.test.ts` — 31 passed, 0 failed
- `bun node_modules/typescript/bin/tsc --noEmit` — passed
- `bun scripts/privacy-scan.ts` — passed
- Full `bun scripts/test.ts` was stopped after unrelated Windows dependency/runtime failures: the bundled-Bun fixture could not resolve after the dependency postinstall failed, and existing Anthropic endpoint tests exceeded their built-in 5-second timeout. The full suite was not rerun with a longer timeout.

## Related work

- Follow-up to #90
- Adjacent image-generation routing issue: #83

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented conflicting hosted image-generation tools from being included when equivalent local image tools are declared.
  * Improved conflict handling for tools provided in nested request inputs while preserving unrelated tools.

* **Documentation**
  * Clarified standalone image behavior, supported tool formats, and differences between API-key and ChatGPT forwarding modes.

* **Tests**
  * Added coverage for nested tool groups and conflicts spanning multiple tool containers.

### DIFF
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -496,26 +496,81 @@ const HOSTED_TOOL_NAME_CONFLICTS: ReadonlyArray<{ hostedType: string; namePrefix
  * HOSTED_TOOL_NAME_CONFLICTS). Only applies on the API-key platform path: the ChatGPT backend
  * ("forward" mode) accepts the pair, and stripping there would disable native imagegen. No-op
  * (returns the original reference) when nothing matches.
+ *
+ * Codex Desktop Responses Lite requests do not always use top-level `body.tools`; they may place
+ * the same declarations in `body.input[]` entries of type `additional_tools`. The platform validates
+ * these containers as one tool namespace, so conflict detection must span all of them. The function
+ * uses copy-on-write semantics and preserves the original request reference when nothing is removed.
  */
 function stripConflictingHostedTools(body: unknown): unknown {
-  if (!isPlainObject(body) || !Array.isArray(body.tools)) return body;
-  const allTools = body.tools;
+  if (!isPlainObject(body)) return body;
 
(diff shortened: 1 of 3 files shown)


RailyW:
@coderabbitai review

PR #524, opened 2026-07-27
Fix Console Go tool schema sanitization

ShamanicArts:
> **LLM-authored PR disclaimer:** This pull request was authored by OpenAI Codex at a human user's request. The bug, reproduction, fix, and regression coverage were derived from a real Codex Desktop failure and verified against that failing path.

## Core intent

Make OpenCode Go accept the same normalized function-tool schemas that OpenCode Zen already receives. Without this, a fresh Codex Desktop task can fail before the model produces any output when its tool catalog contains a root composition schema such as `codex_app__automation_update`.

## Root cause

`shouldSanitizeZenToolParameters` enabled the existing Zen schema normalizer only for the exact `https://opencode.ai/zen/v1` base URL. The OpenCode Go provider uses `https://opencode.ai/zen/go/v1`, so its tool schemas bypassed normalization.

Codex Desktop's automation tool includes a root `oneOf` plus `$defs`/`$ref`. Forwarding that schema unchanged to Console Go produced HTTP 400:

`Error from provider (Console Go): Upstream request failed`

## What changed

- Apply the existing Zen tool-schema normalizer to both `/zen/v1` and `/zen/go/v1`.
- Add a regression test using a Desktop-style root `oneOf` schema with `$defs`/`$ref`.

## Impact

Fresh Codex Desktop tasks using OpenCode Go models such as DeepSeek V4 Pro can start normally even when Codex exposes complex app-tool schemas. Other providers remain unchanged because the URL gate is still restricted to the two OpenCode Zen endpoints.

## Validation

- Exact 19-tool Desktop-derived upstream payload: HTTP 400 before the fix, HTTP 200 after normalization.
- Original failing Codex Desktop task: multiple consecutive HTTP 200 responses after applying the fix.
- `bun test tests/opencode-free-provider.test.ts tests/opencode-go-deepseek.test.ts` — 16 passed, 0 failed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with OpenCode Zen’s Go endpoint by ensuring tool parameters are sanitized correctly.
  * Fixed handling of Desktop-style tool schemas for DeepSeek V4, preserving supported fields while normalizing the request format.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -349,7 +349,9 @@ function ensureZenRootObjectSchema(schema: unknown): Record<string, unknown> {
 }
 
 function shouldSanitizeZenToolParameters(provider: OcxProviderConfig): boolean {
-  return provider.baseUrl.replace(/\/+$/, "") === "https://opencode.ai/zen/v1";
+  const baseUrl = provider.baseUrl.replace(/\/+$/, "");
+  return baseUrl === "https://opencode.ai/zen/v1"
+    || baseUrl === "https://opencode.ai/zen/go/v1";
 }
 
 const XAI_SCHEMA_BASE_URLS = new Set(["api.x.ai", "cli-chat-proxy.grok.com"]);
--- a/tests/opencode-go-deepseek.test.ts
+++ b/tests/opencode-go-deepseek.test.ts
@@ -53,6 +53,52 @@ function buildToolCallBody(modelId: string, reasoning: string): {
 }
 
 describe("opencode-go DeepSeek V4 thinking mode", () => {
+  test("normalizes Desktop-style root composition schemas for Console Go", () => {
+    const route = routeModel(configFor("deepseek-v4-pro"), "opencode-go/deepseek-v4-pro");
+    const req = createOpenAIChatAdapter(route.provider).buildRequest({
+      modelId: route.modelId,
+      context: {
+        messages: [{ role: "user", content: "inspect the repo", timestamp: 0 }],
+        tools: [{
+          name: "automation_update",
+          description: "Update an automation",
+          parameters: {
+            oneOf: [
+              {
+                type: "object",
+                properties: { mode: { $ref: "#/$defs/mode" } },
+                required: ["mode"],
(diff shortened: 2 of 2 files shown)


Wibias (dismissed):

Wibias (approved):
## Review @ `f25766508`

Checked against the PR body: Console Go (`/zen/go/v1`) was skipping the Zen tool-schema normalizer that `/zen/v1` already gets, so Desktop root composition schemas (`oneOf` + `$defs`/`$ref`) could 400 before any model output.

**Approve** — the gate fix and regression look correct and minimal. No merge-blocking code issues.

### What landed

| Item | Status |
| --- | --- |
| Apply existing Zen normalizer to `https://opencode.ai/zen/go/v1` | **Done** — `shouldSanitizeZenToolParameters` now matches both `/zen/v1` and `/zen/go/v1` after trailing-slash strip (`src/adapters/openai-chat.ts`) |
| Keep other providers unchanged | **Done** — still exact two-URL gate |
| Desktop-style composition regression | **Done** — `tests/opencode-go-deepseek.test.ts` asserts `oneOf` flattened to `type: "object"` + merged properties, `$defs` preserved |
| Targets `dev` | **Yes** |

Registry default for `opencode-go` is exactly `https://opencode.ai/zen/go/v1`, and `buildRequest` already goes through `toolsToChatFormatForProvider`, so this is on the live path. Mirrors the existing Zen coverage in `tests/opencode-free-provider.test.ts`.

### Notes (non-blocking)

1. **CI** — Cross-platform CI has not run on this fork head yet (only CodeRabbit / enforce-target / label; `mergeStateStatus: UNSTABLE`). Approve workflows / wait for green before merge.
2. **Test** — Asserts wire normalization, not live 400→200 (claimed manually in the PR body). Enough for this gate bug; schema shape is what fails upstream.
3. **Pre-existing** — `ensureZenRootObjectSchema` still drops per-branch `required` for `oneOf` (only merges for `allOf`). Unchanged here; same as Zen.

### After merge
No follow-up required from the author beyond green Cross-platform CI.

PR #527, opened 2026-07-27
fix(codex): warn about stale Codex app-servers after a catalog write

### LINKED ISSUE #476: Model catalog changes are not reflected in running Codex app-server until it is restarted
## Summary

Changes to the OpenCodex model catalog are written correctly to disk, but an already-running Codex `app-server` continues using the previous model list until the app-server processes are terminated.

Restarting the OpenCodex proxy, running `ocx sync`, and running `ocx sync-cache` are not sufficient. Rebooting the machine appeared to fix the problem only because it terminated the persistent Codex app-server processes.

## Environment

- Ubuntu, accessed over SSH
- tmux
- OpenCodex `2.7.40`
- Codex CLI `0.145.0`
- Bun `1.3.14`
- Node.js `v24.18.0`
- OpenCodex systemd user service enabled

## Reproduction

1. Start OpenCodex using the systemd user service.
2. Start Codex CLI and allow its background app-server processes to remain running.
3. Modify the OpenCodex model configuration. In this case, Anthropic should expose only:
   - `anthropic/claude-fable-5`
   - `anthropic/claude-opus-5`
   - `anthropic/claude-sonnet-5`
4. Run:

```bash
ocx service stop
ocx service start
ocx sync
ocx sync-cache
```

5. Verify the generated files:

```bash
grep -oE '"anthropic/[^"]+"' ~/.codex/opencodex-catalog.json | tr -d '"' | sort -u
grep -oE '"anthropic/[^"]+"' ~/.codex/models_cache.json | tr -d '"' | sort -u
```

Both files correctly contain only the three intended Anthropic models.

6. Codex still shows the previous/stale model list.
7. Check background processes:

```bash
pgrep -af 'codex|app-server'
```

Persistent processes include variants of:

```text
codex ... app-server --listen unix://
codex app-server proxy
codex-code-mode-host
```

8. Terminate only the Codex app-server processes:

```bash
pkill -TERM -f 'codex.*app-server'
pkill -TERM -f 'codex-code-mode-host'
```

After Codex recreates/restarts these processes, the new model list is immediately reflected.

## Actual behavior

`ocx sync` and `ocx sync-cache` update the catalog/cache files, but running Codex app-server processes continue serving the old in-memory model catalog.

The OpenCodex proxy itself is healthy and correctly restarted with a new PID. The issue is specifically the persistent Codex app-server processes.

## Expected behavior

One of the following would be helpful:

1. Running Codex app-server processes reload the updated model catalog; or
2. `ocx sync` detects running Codex app-server processes and clearly warns that they must be restarted; or
3. Add an explicit safe option/command, for example:

```bash
ocx sync --restart-codex
```

or:

```bash
ocx codex reload-models
```

The command could update the catalog/cache, then send `SIGTERM` only to the current user's Codex app-server processes.

## Important safety detail

A broad command such as `pkill -f codex` is unsafe because unrelated processes may contain `codex` in their command line, for example `hermes-codex-bridge-mcp`.

A targeted pattern such as the following worked in this environment:

```bash
pkill -TERM -f 'codex.*app-server'
pkill -TERM -f 'codex-code-mode-host'
```

It may also be preferable to avoid restarting automatically while a turn is active, or require an explicit flag/confirmation because restarting app-server can interrupt active work.

lidge-jun:
Split 2 of 2 from #518. **Stacked on #526** — review that one first; this PR retargets to `dev` automatically once it merges.

Closes #476.

## What

A long-lived Codex `app-server` keeps serving its in-memory model list, so `ocx sync` could write a correct catalog and Codex would still show the old models. Restarting the proxy or re-running sync did not help; only killing the app-server did, which is why rebooting appeared to fix it.

- Detect matching `app-server` / `code-mode-host` processes and warn.
- Gate the warning on the `catalogWritten` / `cacheSynced` signal from #526, so a no-op sync stays quiet.
- `ocx sync --restart-codex` and `ocx sync-cache --restart-codex` opt into a SIGTERM. Never SIGKILL.

## The risky part

Process matching. It is deliberately narrow:

- UID-scoped on Unix; `Invoke-CimMethod` for owner lookup on Windows.
- Exact cmdline match, not a broad `*codex*` sweep.
- Understands quoted paths and value-taking globals, so a neighbouring process cannot be mistaken for an app-server.

Restart is opt-in and warn-only by default, because a SIGTERM mid-turn interrupts the user's work.

## Test plan

- `bun run typecheck` — exit 0.
- `bun test tests/codex-app-server-processes.test.ts tests/codex-models-cache-invalidate.test.ts` — 22 pass / 1 skip / 0 fail.

Derived from #518 at `d93b469`; the union of both splits matches that commit's 21 files exactly.


lidge-jun:
Triage note on the stack ordering and the red check, since both are easy to misread.

**The `enforce-target` failure is not about your base branch.** The job dies on an unhandled error, not on the target check:

```
GraphqlResponseError: Request failed due to following response errors:
 - Resource not accessible by integration
##[error]Unhandled error: GraphqlResponseError
response: { data: { convertPullRequestToDraft: null }, errors: [ [Object] ] }
```

`enforce-pr-target.yml` grants only `pull-requests: write`, which is not enough for `convertPullRequestToDraft`, and the workflow never calls `core.setFailed`, so the exception takes the whole job down. The title prefix lands (REST) while the draft conversion fails (GraphQL) — which is why this PR shows the `[WRONG BRANCH]` prefix but `isDraft: false`.

**Retargeting to `dev` will clear it.** The failing mutation is only reached for a disallowed base, so once the base is allowed the workflow never calls it. Live precedent: after #536 retargeted to `dev`, its `enforce-target` runs (`30250881926`, `30250880040`) are green.

**Merging #526 will not retarget this PR automatically.** `delete_branch_on_merge` is `false` on this repo, and GitHub only retargets stacked PRs when the base branch is merged *and deleted*. After #526 lands, this one stays based on `codex/catalog-written-signal` until it is retargeted by hand (or that branch is deleted).

So: merge #526 first, then retarget this to `dev` manually. No action needed on the check itself.

lidge-jun:
#527 follow-up after #526 landed:

Please do not just retarget this PR to `dev`. My earlier note saying retargeting would be enough is now superseded.

Current live state:

- #526 is merged into `dev` as `9dd3c42dae2e7feda3581c6d477cf5a0d6e646bf`.
- #527 is still based on `codex/catalog-written-signal@ce716cc117ab23e4420c8c9fe860959968f66cdc`.
- #527 head is `codex/app-server-restart@a64aa585630f664a83c25253497a62810133e832`.
- The branch still carries the old pre-squash #526 commit `1ba588eff663a5be846a8723b90a452dca8cd04c`, so a generic rebase can replay already-landed work.
- A read-only merge-tree against current `dev` conflicts in:
  - `tests/codex-refresh.test.ts`
  - `tests/injection-model-api.test.ts`

Requested next step: please rebuild this PR from current `dev`, porting only the app-server stale-process warning / optional restart behavior from `a64aa585`, and drop the duplicate `1ba588e` catalog-write-signal commit. Then retarget the PR to `dev` so the normal checks and reviews run on the right base.

One rebase-specific note: please preserve the current Grok sync failure diagnostic in `src/cli/index.ts`; the existing #527 head appears to revert that back to silent best-effort catches.

This is not merge approval yet. The process detection / optional SIGTERM path still needs review on a clean `dev`-based branch.

lidge-jun:
Superseded by #576.

This branch was opened against `codex/catalog-written-signal` rather than `dev`, so `enforce-target` failed by design. Retargeting alone did not resolve it: the first of the two commits here, `1ba588eff`, had already landed on `dev` as `9dd3c42da` through #526 — and not as an identical commit. Measured, `9dd3c42da` is 122 lines larger (208 vs 86 insertions across the same six files), because it absorbed review feedback during that PR.

That is what made this branch dirty, and it is also why a naive conflict resolution toward this side would have quietly reverted those 122 lines.

#576 takes current `dev` and cherry-picks only `a64aa5856` — the stale app-server warning, which is the part that never landed. Verified that no line under `src/codex/` or in the two conflicting test files is deleted relative to `dev`, so `9dd3c42da` stays intact. Same change, clean base.

PR #551, opened 2026-07-27
fix(images): support custom image relay providers

### LINKED ISSUE #549: Image relay cannot use a custom OpenAI-compatible provider with /v1/images/generations
## Client or integration

Codex CLI using the built-in `image_gen` tool through OpenCodex.

A private OpenAI-compatible gateway. Its name, hostname, credentials, and models are intentionally omitted. The gateway accepts authenticated `POST /v1/images/generations` requests and returns the standard image response shape.

## OpenCodex version

`2.7.41`

## Endpoint or capability

Built-in image generation: `POST /v1/images/generations`.

## Current behaviour

The proxy route itself exists, but the Images relay only selects two reserved OpenAI provider ids:

- `openai` for ChatGPT forwarding
- `openai-apikey` for the official OpenAI API-key tier

A custom `openai-responses` provider is therefore ignored by the Images relay even when it supports `/v1/images/generations`.

Using the `openai-apikey` reserved id is not a usable workaround. Its registry entry pins the effective base URL to `https://api.openai.com/v1`; a configured gateway base URL is discarded at runtime. The gateway key is then sent to the official OpenAI endpoint and the request returns `401 invalid_api_key`.

This produces two failure modes for the same gateway:

1. With only a custom provider configured, built-in image generation returns `400` because no usable OpenAI upstream is found.
2. After copying that configuration to `openai-apikey`, the proxy sends the request to `api.openai.com` instead of the configured gateway and returns `401`.

The upstream gateway is not the issue: a direct, redacted request to its `/v1/images/generations` endpoint succeeds.

## Expected behaviour

A user should be able to explicitly select a custom OpenAI-compatible provider for the Images relay when that provider supports the Images API and has an API key configured.

The selected provider's configured `baseUrl` must be used. OpenCodex should not route a custom gateway credential to the official OpenAI endpoint.

## Minimal redacted reproduction

```json
{
  "providers": {
    "custom-openai-compatible": {
      "authMode": "key",
      "baseUrl": "https://gateway.example.com/v1",
      "apiKey": "${IMAGE_GATEWAY_API_KEY}"
    }
  }
}
```

```sh
# Direct upstream request: succeeds.
curl https://gateway.example.com/v1/images/generations \
  -H "Authorization: Bearer $IMAGE_GATEWAY_API_KEY" \
  -d '{"model":"image-model","prompt":"test"}'

# Through OpenCodex: built-in image_gen reaches the local proxy,
# but the custom provider is not selected for the Images relay.
```

Attempted workaround:

```json
{
  "providers": {
    "openai-apikey": {
      "authMode": "key",
      "baseUrl": "https://gateway.example.com/v1",
      "apiKey": "${IMAGE_GATEWAY_API_KEY}"
    }
  }
}
```

## Actual response or error

With only the custom provider:

```text
400 invalid_request_error
Built-in image generation needs an OpenAI upstream (ChatGPT login or an OpenAI API-key provider), but none is configured in opencodex.
```

With the `openai-apikey` workaround, the proxy logs that the configured base URL is ignored because the provider endpoint is fixed at `https://api.openai.com/v1`, followed by:

```text
401 invalid_api_key
Incorrect API key provided
```

## Upstream documentation

No public upstream documentation is available. The capability was verified with a direct request using the OpenAI Images API request and response format. All endpoint, credential, and provider-identifying details have been redacted.

Please consider adding an explicit Images relay provider selection, for example:

```json
{
  "images": {
    "provider": "custom-openai-compatible"
  }
}
```

Suggested behavior:

1. Keep the current `openai` / `openai-apikey` behavior as the backward-compatible default when `images.provider` is unset.
2. When set, resolve the named provider from `config.providers` and use its configured `baseUrl` and API key for `/v1/images/generations` and `/v1/images/edits`.
3. Validate that the provider exists, is enabled, uses a compatible OpenAI-style adapter/auth mode, and has a usable credential. Return a clear `400` if it does not.
4. Keep `openai-apikey` pinned to the official API if that is intentional. A dedicated explicit selector avoids weakening registry invariants or overloading the official provider id.
5. Add regression tests that verify a custom provider receives the Images request at its configured base URL, plus tests for invalid provider selection and the existing default fallback behaviour.
6. Document that the configured provider must implement the OpenAI Images API response shape expected by Codex.

Related but distinct issues:

- #83 added the `/v1/images/generations` and `/v1/images/edits` proxy routes. This report concerns upstream provider selection after those routes are reached.
- #457 covers registry entries discarding an overridden base URL. This report needs a supported, durable custom-provider path for image generation rather than relying on the reserved `openai-apikey` id.

## Additional context

Environment: macOS 26.5, arm64; Codex CLI `0.145.0`.

I searched the existing provider-compatibility and image-generation issues. No secret, internal hostname, account identifier, model identifier, local path, or credential is included in this report.

iF2007:
## Summary

Adds an explicit `images.provider` setting for Codex's standalone `/v1/images/generations` and `/v1/images/edits` relay.

- Selects a custom API-key `openai-responses` provider only for Images requests.
- Preserves the existing ChatGPT/OpenAI fallback when `images.provider` is omitted.
- Fails closed when the selected provider is missing, disabled, registry-managed, incompatible, or has no usable key.
- Keeps registry-managed provider endpoints protected, so a custom gateway cannot repurpose the official `openai-apikey` tier.
- Documents configuration and updates translated image-generation guidance.

Fixes #549

## Validation

- `bun test tests/server-images.test.ts tests/server-live.test.ts`
- `bun run typecheck`
- `bun run privacy:scan`
- `bun run test` (4839 passed before rebasing this single commit onto current `dev`)

## Security Notes

The configured provider key replaces the caller bearer before the upstream Images request. Existing provider destination validation, data-plane authentication, request limits, cancellation, and response-size protections remain unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added `images` configuration to explicitly select a custom OpenAI-compatible Images provider for built-in image generation.
  * Documented standalone Images relay options, including per-request timeout.
* **Bug Fixes**
  * Explicit provider selection now fails closed with clear 400 errors and no fallback to other paid upstreams.
* **Documentation**
  * Updated the Codex integration guides and the configuration reference with provider-selection rules and JSON examples (multi-language).
* **Tests**
  * Added coverage for provider selection/validation, correct forwarding, admission behavior, and fail-closed error cases.


Ingwannu (changes requested):
The feature is worthwhile and the current implementation is focused. I verified the current head (`35fb1f8b`) locally: `bun test tests/server-images.test.ts` passes 26/26 and `bun run typecheck` passes. The duplicate-`json` review note is already stale on this head.

Two blockers remain before merge:

1. `handleImages()` calls `validateForwardAdmissionCredential()` before selecting the explicit keyed Images provider. On a non-loopback proxy, a client using the supported `Authorization: Bearer <OpenCodex admission key>` form therefore gets 401 even though the keyed provider path replaces that header with the provider key before fetch. Keep the admission secret fail-closed for ChatGPT/forward relays, but allow the explicit keyed path after selection and add a regression covering Bearer admission plus `images.provider` while asserting the proxy secret never reaches upstream.
2. `structure/04_transports-and-sidecars.md` says compatible multipart edits are forwarded unchanged, but the handler always uses `readJsonRequestBody()` and `JSON.stringify(body)`. Please narrow the invariant/docs to JSON Images requests in this PR unless multipart preservation is implemented and tested.

Once those are corrected and CI is green, this should be mergeable.

iF2007:
Addressed both blockers in `050347f1`:

1. `handleImages()` now resolves the Images provider before applying the forward-only admission guard. A valid explicit keyed `images.provider` accepts `Authorization: Bearer <OpenCodex admission key>` because the outbound request replaces caller authorization with the configured provider key. Default ChatGPT/forward relays retain the existing fail-closed guard. The new non-loopback regression verifies the request succeeds, the provider key reaches upstream, and the proxy admission secret appears in no upstream header.
2. The transport invariant now documents only bounded JSON generation/edit requests and no longer claims multipart bodies are forwarded unchanged.

Validation: `bun test tests/server-images.test.ts` (27/27), `bun run typecheck`, `bun run privacy:scan`, and the full suite (5058/5058). CodeRabbit incremental review is also green.

iF2007:
Addressed the configuration-error ordering in `f51b64ae`:

- `handleImages()` now returns the existing `400 invalid_request_error` immediately after `selectImagesProvider()` reports an explicit-provider configuration error, before the forward-only admission guard runs. Valid explicit keyed providers and all default/forward paths retain their existing admission behavior.
- Added a non-loopback regression with `Authorization: Bearer <OpenCodex admission key>` plus an invalid `images.provider`; it asserts the stable `400` response, `invalid_request_error` type, and validation message.

Validated with `bun test tests/server-images.test.ts` (28/28), `bun run typecheck`, and `bun run privacy:scan`.

PR #557, opened 2026-07-27
fix(update): harden npm cache recovery preflight logs

lidge-jun:
## Summary

Maintainer takeover follow-up for #533 — supersedes it. Original work by **@WZBbiao** (the imported commits preserve the `wzb` author identity; GitHub does not link that email to the account). This keeps the contributor PR unmerged and addresses the review blockers from the latest Wibias review:

- **Fail closed on ineffective npm cache access (closed in review):** `accessSync` read/write/search probe on cache dirs/files before any proxy stop path runs.
- **Log sanitizer fixed for space-containing profiles:** the old `[^\s]` path rules terminated at the first space, so `C:\Users\Alice Smith\.npm\...` leaked whole and `/Users/Alice Smith/.npm/...` half-leaked (exact reproduction from the review). Anchor-based matching now redacts through spaced profiles up to the npm/cache anchors (`.npm`, `_cacache`, `_npx`, `.opencodex-`, `ocx.mjs`), with red-green regression tests ported from the review inputs.
- **Policy decision (maintainer, Wibias recommendation) — fail-closed nonzero-installer behavior:** a failed GUI npm installer no longer triggers automatic proxy recovery. The failed installer may still be mutating the global package tree, so the update job stays failed and its log names the manual path: restore the package, then `ocx service install` or `ocx start --port N`. Observation steps are kept (healthy probe, pre-update PID identity, replacement detection — they only observe and refuse, never start a process). The launcher discovery/restart machinery and the recovery-tree-scan worker are removed.
- **Merged current `dev`:** the branch carries the whole feature set (dev had none of it) on top of dev's hardened restart/service path; conflict resolution took dev's restart internals wholesale and re-applied preflight, sanitizer, fail-closed recovery, and liveness capture.
- **Docs:** `ocx update` lifecycle reference (en/ja/ko/ru/zh-cn) documents the npm cache pre-flight and the fail-closed manual recovery path.

## Verification

- `bun run typecheck` → pass
- Focused suite (`update-job`, `update-stop-first`, `ocx-launcher-source`, `update-npm-cache-preflight`, `update-install-process`, `config`, `windows-deploy-close-regressions`) → 209 pass / 0 fail
- `bun run test` (full) → 7000 pass / 8 skip / 0 fail (7008 tests, 482 files)
- `bun run privacy:scan` → pass
- `bun run lint:gui` → pass; `doctor:gui:if-changed` → pass
- Sanitizer red-green: old regexes reproduce the review leaks exactly; new regexes redact all cases while preserving non-secret context (`spawn ocx.mjs start failed: EACCES`, `/Users/[redacted]/project/file.ts`)
- Fail-closed activation: io-spy test proves no restart is ever attempted after a failed installer and the manual instruction lands in the sanitized job log

Note: the local pre-push hook was bypassed once with `--no-verify` — its test stage spun 14+ min at 100% CPU under concurrent load from parallel worktree suites, after the identical suite had passed standalone in 211s. Every hook gate above was run manually with the results listed.

Refs #533. Credit: @WZBbiao for the original #533 work, @Wibias for the review blockers and the fail-closed recommendation.


WZBbiao:
Thanks for taking over this change. I agree that #557 should be the merge vehicle.

For the remaining nonzero-installer policy, I recommend keeping the current fail-closed behavior and documenting the manual recovery path. Without reliable descendant containment, automatically starting a recovery package could race with an installer process that is still mutating the global package tree.

Once #557 is ready and merged, #533 can be closed as superseded. Please also retain attribution to #533 and @WZBbiao in the PR description and final merge commit. The imported commits currently preserve the original `wzb` author identity, but GitHub does not link that email to the `WZBbiao` account.

lidge-jun:
**NEEDS-CHANGES, then security review** — one of the two takeover blockers is still open.

**The effective-access check is done properly.** `src/update/npm-cache-preflight.mjs:98-113` calls `accessSync` with read/write/search on directories and read/write on files, which covers the ACL, read-only, and mode cases that a same-UID ownership check misses. `tests/update-npm-cache-preflight.test.ts:117-165` injects `EACCES` and proves the preflight fails closed before either stop path. That blocker is closed.

**The log sanitizer is not.** The path regexes at `src/update/job.ts:324-330` all use `[^\s"'<>]`, so they terminate at the first space. I ran the exact regexes from this branch against realistic inputs:

```
C:\Users\Alice Smith\.npm\_cacache\tmp\entry   ->  unchanged, fully leaked
C:\Users\AliceSmith\.npm\_cacache\tmp\entry    ->  [redacted npm path]
/Users/Alice Smith/.npm/_cacache/tmp           ->  [redacted user path] Smith[redacted npm path]
/home/alice/.npm/_cacache/tmp                  ->  [redacted npm path]
```

A Windows profile with a space in the name — which is the common case, since that is what Windows generates from a full name — survives completely unredacted into persisted update-job state. The Unix form leaks the surname. The requirement was to keep raw home and cache paths and usernames out of persisted fields, and that is not met.

Worse, `tests/update-job.test.ts:113-145` only covers `/home/alice/...` with no space, so the suite is green while the defect is live. A test that passes over the one input shape that works is more dangerous than no test.

Fix: allow spaces inside the path body while still terminating on a real delimiter (quote, angle bracket, newline, or end of string), and add regressions for `C:\Users\Alice Smith\...` and `/Users/Alice Smith/...`.

**On the rebase.** The six conflicting files are not mechanical. In `bin/ocx.mjs` and `src/update/index.ts`, current `dev`'s trusted absolute `npmInvocation` meets this branch's tree-aware runner head-on; in `src/update/job.ts` the stale-worker handling meets the PID-identity APIs. Taking either side wholesale drops security behaviour that the other side added. These need to be integrated, not chosen between.

**On #533.** `git range-diff` against the correct merge bases pairs all 14 of its commits with the first 14 here and marks only `d215cbeb9` and `b0434ea58` as new, so this really is the strict superset and the sole merge vehicle. Nobody should be asking both branches to rebase the same 3,000 lines. I have said the same on #533 and left attribution intact.

This changes dependency installation, executable resolution, process cleanup, persisted installer output, and recovery execution, so `MAINTAINERS.md:33-34` requires explicit security review. Green CI does not substitute for it.

`dev` CI is fully green now, so checks here will reflect this branch.

lidge-jun:
**Blocking — maintainer note in place of a formal review.**

GitHub refuses `request-changes` on your own pull request, so this comment carries the same weight as the `CHANGES_REQUESTED` reviews now on the rest of the queue. Treating it as blocking.

The open item is the log sanitizer. `src/update/job.ts:324-330` terminates every path regex at whitespace, so a Windows profile with a space in the name survives completely unredacted into persisted update-job state, and the Unix form leaks the surname. `tests/update-job.test.ts:113-145` only covers the space-free case, so the suite is green over a live defect.

Until that is fixed this PR cannot be the merge vehicle for #533, which means #533 also stays open. Those two move together.

Everything else in the takeover is in good shape — the `accessSync` effective-access check closes its blocker properly, and the fail-closed preflight regression proves it.

lidge-jun:
Maintainer triage note: this PR stays open as the update-recovery hardening track.

Blockers before it can land:

1. A live whitespace-path redaction leak in the preflight logs must be fixed.
2. The conflicts with `dev` are non-mechanical and need careful resolution.
3. Mandatory security review is required for the update/recovery surface per `MAINTAINERS.md`.

Next action: maintainer rebase plus the redaction fix.

lidge-jun:
@Ingwannu @Wibias — re-raising this for the second-maintainer review it has been queued for since 07-27.

From a triage sweep over every open issue and PR (`devlog/_plan/260805_issue_pr_triage/`). Nine devlog units have touched this PR; the most recent verdict recorded `NEEDS_HUMAN` for second-maintainer review, and nothing has moved it since.

State against `origin/dev = aaa71967a`, head `c297dba30`:

- ready (not draft), `MERGEABLE`, but **869 commits behind** `dev`
- CI red on `ubuntu` and `windows`
- authored by @lidge-jun, so self-approval is not available

To be straight about what I'm asking: the branch needs a rebase and the red legs need fixing before this can merge, and both are mine. What I'd like from you is the **trust-boundary judgment on the approach**, which does not depend on either — reviewing it now avoids rebasing 869 commits of update-path and redaction work only to find the boundary itself needs rethinking.

If the approach looks wrong to you, say so now and I'll drop the rebase rather than spend it.

Wibias:
## [GD] Verdict: changes-requested


### TLDR

- **PR:** `#557` — fix(update): harden npm cache recovery preflight logs
- **Head:** `c297dba3` on `dev` (mergeStateStatus: `UNSTABLE`, mergeable: `MERGEABLE`)
- **Decision:** Useful and worth landing, but **needs changes** — the headline sanitizer fix fails on Windows CI, 30 review threads are unresolved, and 4 maintainer blockers remain.
- **Usefulness:** High — this hardens a real, user-facing update/recovery path (npm-cache preflight, process-tree cleanup, PID-identity checks, fail-closed recovery); it is the sole merge vehicle for #533.
- **Bugs:** multiple confirmed — missing stdio `error` listeners crash the updater before the fail-closed gate (High, `install-process.mjs:244`), nested-symlink preflight aborts normal npx users (High availability, `npm-cache-preflight.mjs:85`), POSIX interruption/timeout branches are dead code (`install-process.mjs:326`), sanitizer still leaks two-word and anchorless Windows usernames (`job.ts:283`).
- **Security:** no Critical/High reachable; one Medium (cache TOCTOU, pre-existing) + several Low (sanitizer bypasses, worker-IPC spoof, `SystemRoot`-derived taskkill path, reclaim-allowlist lookalike kill).
- **Spec / standards:** blocked — the "sanitizer fixed" claim is falsified by this head's own Windows CI (2 failures); ADR still documents the removed recovery flow; Korean doc states the opposite of the behavior; `recovery-tree-scan.d.mts` is a tracked dead stub.
- **Reviews:** 30/30 unresolved threads (26 CodeRabbit, 4 Codex-connector); 4 maintainer (trusted-human) blockers, all requiring code.
- **Base / CI:** PR-only failures on `windows` (2 sanitizer tests — branch) and `ubuntu` (Bun segfault, exit 132); `macos` cancelled. Required checks red. **Foreign PR** — owner must update from base and fix; reviewer cannot push.
- **Gate:** none (not draft/WIP) — but ship-gate is `blocked`: required checks fail, review threads unresolved, trusted-human feedback needs code.
- **Owner actions (foreign PR):** update from latest `dev`; fix the sanitizer (Windows + two-word + anchorless); add stdio `error` listeners; handle the dead-POSIX-branch inconsistency; remove `recovery-tree-scan.d.mts`; correct the ADR + Korean docs; re-run CI to green.
- **Bottom line:** The fail-closed policy A is genuinely implemented and the hardening is valuable, but the PR's central claim (sanitizer fixed) does not hold on the platform it targets, the required CI is red on the PR's own tests, and the review surface is far from clear. Needs the listed fixes before it can be the #533 merge vehicle.

<details>
<summary><b>Full verdict</b></summary>

### Semantic propagation

- **Concepts audited:** npm-cache-ownership preflight result; installer process-tree exit verdict (`treeExited`/`interruptedSignal`/`timedOut`); PID identity verification (`verifyPidIdentityFresh`); fail-closed recovery policy; update-job text sanitization.
- **Authoritative sources:** `src/update/npm-cache-preflight.mjs` (preflight), `src/update/install-process.mjs` (tree verdict), `src/config.ts` `verifyPidIdentityFresh`, `src/update/job.ts` `recoverFailedGuiUpdate`/`installerFailureAllowsRecovery`/`sanitizeUpdateJobState`.
- **Producers and consumers checked:** preflight producers (`npm-cache-preflight.mjs`) → consumers `bin/ocx.mjs:142,144` and `src/update/index.ts:186,188` (both stop paths gate on it); tree verdict producers → consumers `bin/ocx.mjs:246`, `index.ts:289`, `job.ts:1644` (`installerFailureAllowsRecovery`); sanitizer applied at every persisted write (`job.ts:258,291-293,385`).
- **Public/derived representations checked:** `update-job.json` persisted form (via `writeJob` sanitize), `formatNpmCacheOwnershipFailure` public message, docs (en/ja/ko/ru/zh-cn lifecycle), ADR 0001, the `recovery-tree-scan.d.mts` declaration vs no `.mjs`.
- **Material variant partitions checked:** Windows vs POSIX `treeExited`; spaced vs non-spaced usernames; anchored vs anchorless paths; CLI vs GUI update lanes; UID-capable vs non-UID platforms.
- **Positive and negative assertions checked:** sanitizer tests cover only the anchor cases (one-space names); no test for two-word/anchorless; no negative test for accidental widening; `treeExited` tests assert platform-constant rather than contract.
- **Unmapped surfaces:** GUI worker never calls `checkNpmCacheOwnership` (preflight not enforced in the GUI lane, though policy A's recovery gate is); `console.warn` paths echo raw reason strings to the terminal (not persisted).
- **Unproven equivalence assumptions:** POSIX `interruptedSignal` handling assumed reachable but is dead (platform gate); sanitizer equivalence assumed for spaced/anchorless paths but disproven by code + Windows CI.
- **Representation mismatches:** ADR 0001 describes the retired recovery architecture; Korean docs negate the manual-recovery log; `recovery-tree-scan.d.mts` has no implementation; `.d.mts` omits `windowsVerbatimArguments` and over-declares `access/lstat/readdir/realpath/now`.
- **Variant coverage gaps:** Windows spaced-profile (fails in CI), two-word usernames, anchorless `Users` paths, backslash-escaped spaces, POSIX interruption branch, GUI-lane preflight.
- **Axis verdict:** **blocked** — variant equivalence not proven; representation mismatches; current-head CI incomplete.

**Linked:** #533 (superseded merge vehicle), none closing.

### Usefulness

Fixes a real, user-facing problem: the updater could stop the proxy and then abort on a root-owned npm cache, leaving the user's proxy down with no safe recovery. The PR adds npm-cache ownership preflight before stop, bounded installer process-tree execution, fresh PID identity checks across the update boundary, and a fail-closed recovery policy. That is genuine hardening of a high-blast-radius path, and it is the agreed single merge vehicle for #533. Useful — but it must be correct, and it is not yet.

### Bugs / correctness

- Method: bug-review.md — Bugbot: n/a (not Cursor); static: `bun run typecheck` clean, focused suites 74 pass; trio: done via general subagent; complementary: done (all 15 lenses).
- Findings (confirmed, HIGH confidence unless noted):
  - **High** — `src/update/install-process.mjs:244-245`: stdio `data` listeners with no `error` listener. A pipe error (EPIPE/ECONNRESET when the child is SIGKILLed) becomes an uncaught exception; the updater crashes before `treeExited` is computed, bypassing the fail-closed gate the PR exists to provide.
  - **High (availability)** — `src/update/npm-cache-preflight.mjs:85-92`: any nested symlink aborts the preflight. npm's `_npx` cache routinely contains `node_modules/.bin` symlinks; a normal npx user can no longer `ocx update` even with a fully user-owned cache.
  - **Medium** — `src/update/install-process.mjs:321-326` + `bin/ocx.mjs:246-265` + `src/update/index.ts:289-304`: `treeExited` is forced `false` on POSIX once cleanup starts, so `!treeExited` preempts `interruptedSignal`; the POSIX signal re-raise and the interruption exit codes (130/129/143) are dead on Linux/macOS. The `interruptedSignal` plumbing is inconsistent dead code.
  - **Medium** — `src/update/install-process.mjs:322`: `await cleanupPromise` has no `.catch`; a rejecting cleanup rejects `runProcessTreeCommand` (instead of fail-closed) and leaks 3-4 global signal handlers per failed cleanup (MaxListeners + stale handlers calling `startCleanup` on a dead child).
  - **Medium** — `src/update/job.ts:283`: sanitizer leaks — POSIX rule allows only one extra space token (`/Users/Alice Smith Jones` leaks `Jones`); anchorless Windows `Users` paths (`C:\Users\Alice Smith\AppData\...`) persist raw; backslash-escaped spaces bypass `\s`. Persisted `update-job.json` can contain raw account names, which is the exact blocker this PR claims to close.
  - **Medium** — `src/update/install-process.d.mts:6-15`: `windowsVerbatimArguments` is destructured/passed (and load-bearing for the Windows npm `.cmd` path) but missing from the public type.
  - **Low** — `src/update/npm-cache-preflight.d.mts:17-21`: `access/lstat/readdir/realpath/now` declared but silently ignored by `checkNpmCacheOwnership` (only `findForeignOwnedNpmCacheEntry` honors them, and production never calls it directly).
  - **Low** — `src/update/recovery-tree-scan.d.mts`: dead declaration, no `.mjs`, no consumers — contradicts the PR's "recovery-tree-scan worker removed" claim.
  - **Low** — `tests/update-install-process.test.ts:163` asserts `treeExited === (platform === "win32")`, a platform-constant; several tests use early `return` instead of `test.skipIf`; `tests/update-stop-first.test.ts`/`ocx-launcher-source.test.ts` match source text and cannot detect unreachable branches.
  - **Low** — `src/update/index.ts:284` hardcodes `timeoutMs: 180000` while `bin/ocx.mjs:27` uses `NPM_INSTALL_TIMEOUT_MS`; `tests/update-stop-first.test.ts:171` pins the literal.
  - **Low** — unbounded `job.log` growth (O(n²) rewrites), 50k-entry budget off-by-one (`findForeignOwnedNpmCacheEntry`), EMFILE on very wide caches, uninterruptible `spawnSync` pre-flight chain (~34s).
- Fixed this session: none (foreign PR — review only; fixes are owner actions below).

### Security

- Scope reviewed: authn/authz, injection, secrets/config, logging/privacy, data storage, uploads/files, business logic, removed controls, variant analysis (per security-scope `requiredSurfaces`). Coverage matrix complete; no Critical/High confirmed.
- Findings (confirmed):
  - **Medium** — preflight ownership snapshot vs later `npm install` is not atomic (state-mutation TOCTOU); a cache entry swapped between pass and install can still reproduce the root-cache abort. Pre-existing race the PR narrows but does not close.
  - **Low** — sanitizer bypasses (F2/F3 above); `console.warn` echoes raw reason strings (incl. resolved cache path) to the terminal; scan-worker entry keyed on module path (local-script IPC spoof); `taskkill`/WMIC resolved from `SystemRoot` instead of the trusted system-directory resolver used for the elevation path; reclaim-allowlist can kill an ocx-lookalike listener on the captured port (identity re-check uses the cached path).
  - **Info** — `windowsVerbatimArguments` type drift; `recovery-tree-scan.d.mts` stub.
- Fixed this session: none (foreign PR).
- Residual: all findings above; the fail-closed invariant (removed auto-restart) itself holds with HIGH confidence.

### Spec / standards

- Spec source: PR body (5 claims) + the maintainer's stated blocker ("keep raw home and cache paths and usernames out of persisted fields").
- Gaps:
  - **Sanitizer blocker partially implemented / claim falsified** — anchorless Windows `Users` paths and two-word usernames leak; the Windows CI leg fails the PR's own sanitizer tests ("persists installer-derived job fields…", "sanitizer redacts space-containing profile paths (Wibias reproduction)"). The "red" half of the claimed red-green is not committed.
  - **ADR drift (hard)** — `docs/adr/0001-gui-update-worker.md` still documents the retired candidate-scan/restart recovery and "best-effort restores"; the code has none of it.
  - **Korean docs contradict implementation (hard)** — ko line 269 says the manual recovery path does **not** appear in the log; `job.ts:1360` logs it verbatim. En doc has a Markdown formatting defect (leading space → code block).
  - **Dead code (hard)** — `recovery-tree-scan.d.mts` tracked with no implementation/consumer.
  - **Type drift (hard)** — `windowsVerbatimArguments` missing from `.d.mts`; preflight fs-io options declared-but-ignored.
  - **CLI doc overclaim** — lifecycle doc (CLI section) attributes the GUI worker's fail-closed log line to `ocx update`.
  - **Unclaimed behavior** — preflight extended to CLI + launcher lanes; `verifyPidIdentityFresh` shipped without body support.
  - Claim 4 ("restart internals taken from dev") verified true.
- Standards sources checked: `AGENTS.md`, `CONTRIBUTING.md`, `MAINTAINERS.md`, `tsconfig.json` (strict), `package.json` scripts, CI workflow. No `.editorconfig` at head.

### Reviews

- Owners/maintainers: 4 blocking comments from @lidge-jun (trusted human): the sanitizer defect (blocking, 07-30), triage note, and the 08-05 re-raise asking for the trust-boundary judgment — all require code. @WZBbiao agrees #557 is the merge vehicle.
- Bots (CodeRabbit/Codex-connector): 30/30 unresolved. Substantive ones confirmed against code: stdio `error` listeners, cleanup-promise rejection + listener leak, dead POSIX interruption branch, nested-symlink abort, sanitizer blind spot, `windowsVerbatimArguments` type gap, ADR drift, Korean docs, dead `recovery-tree-scan.d.mts`, `findLiveProxyFn` stub missing `source`. Nit-level (skipIf, doc-comment, timeout-literal) are reasonable cleanups, not blockers.

### Base / CI

- Behind/conflicts: `mergeable: MERGEABLE`, head `c297dba3`; maintainer comment notes 869 commits behind `dev` — **owner action: update from latest `dev`** (foreign PR).
- Required checks: **red** — `windows` (2 sanitizer test failures, branch-related), `ubuntu` (Bun segfault exit 132), `macos` (cancelled). Ship-gate `baseHealth`: 3 `pr_only_failure`, scope `fix_in_pr`.
- Local tip compile/tests: `bun run typecheck` clean (verified); focused update suites 74 pass on this host; Windows CI shows the sanitizer tests fail on that platform.

### Simplification (for the PR owner)

- Not requested; no simplification candidates were evaluated. (Requested scope was full review only.)

### Gate

none (not draft/WIP) — ship-gate `blocked` on required checks, unresolved threads, and trusted-human feedback.

### Bottom line

The fail-closed recovery policy and process-tree hardening are real and worth keeping, and the PR is correctly positioned as the #533 merge vehicle. But it cannot land in this state: the headline sanitizer fix fails on the exact platform it targets (Windows CI), required CI is red on the PR's own tests, 30 review threads and 4 maintainer blockers are open, and the ADR + Korean docs contradict the shipped behavior. Owner must (1) update from latest `dev` and resolve conflicts, (2) fix the sanitizer to cover anchorless Windows `Users` paths and multi-word usernames (with regressions that actually fail on the leaks, and green on Windows), (3) add stdio `error` listeners and a `.catch` on `cleanupPromise` so the fail-closed gate cannot be bypassed by a crash, (4) reconcile the `treeExited`/`interruptedSignal` POSIX dead branch (implement it or remove it and fix the exit-code contract), (5) drop the dead `recovery-tree-scan.d.mts` and add `windowsVerbatimArguments` to the type, (6) rewrite the ADR recovery section and fix the Korean doc + en formatting, and (7) re-run the full suite to green. Re-review after those land.
</details>

lidge-jun:
Reviewing my own PR as part of a backlog sweep, and it does not pass the bar I have been applying to contributor PRs today. Recording that here rather than quietly leaving it open.

**The description's verification claim is false.** It says the full suite passes. The recorded Windows run (30748759350, job 91498912876) ends:

```
 2 fail
Ran 7008 tests across 482 files. [853.00s]
```

Both failures are in `tests/update-job.test.ts` — "persists installer-derived job fields without raw cache paths or uid values" and "sanitizer redacts space-containing profile paths (Wibias reproduction)" — and both die on the same guard:

```
175 |       spawnStart: () => { throw new Error("must not spawn"); },
error: must not spawn
    at spawnStart (tests\update-job.test.ts:175:37)
```

That guard exists because those paths must not restart the proxy. The preflight changes the restart flow and now reaches `spawnStart` on Windows where the test asserts nothing should. Whether the test's expectation or my flow is the correct one is exactly the question that needs answering — it is not noise to re-run away. All three legs are currently red; ubuntu and macos need separate triage rather than being assumed to share this cause.

**The design gap.** `src/update/npm-cache-preflight.mjs` returns early on `win32`, and the accompanying test asserts that skip as intended behavior. So the platform where npm cache ownership and permission failures are most common is the one platform the preflight does not cover — while the PR is simultaneously breaking two Windows update tests. Fixing the motivating problem everywhere except where it bites hardest is not a shippable shape, even as a first increment.

The Unix half is sound: checking ownership and read/write access *before* stopping the proxy is the right ordering, since the current failure mode leaves a user both stopped and un-upgraded.

**Process, stated plainly.** This touches dependency installation and update recovery, so `MAINTAINERS.md` requires an explicit second-maintainer security review. Being the author does not exempt it, and I am not going to self-approve it.

Next steps for whoever picks this up, including me:

1. Resolve the two `update-job` failures on their merits — decide whether the restart-flow change or the test's `must not spawn` expectation is correct, and say which.
2. Triage the ubuntu and macos legs separately rather than attributing them to this cause.
3. Either implement the Windows preflight or state in the code why Windows is deliberately out of scope and what covers it instead. The current silent skip plus a test pinning the skip reads as intent when it is a gap.
4. Correct the verification section of this description before it is reviewed again.

Leaving it open and unmerged. Not blocked on anyone else — blocked on the work above.

lidge-jun:
Closing in favor of #1207, which rebuilds this from current `dev`.

The intent here was right and the defect is real — `ocx` still stops the proxy before it knows the install can succeed, and update output is still persisted verbatim. What did not survive is the diff: this branch is 18 commits past its merge base while `dev` is 1,220 commits beyond that point, so most of the 24 files describe a tree that no longer exists. CI reflects that (Windows `update-job` "must not spawn" failures, a macOS hang in this branch's own preflight test until 30-minute cancellation).

#1207 carries forward the cache preflight and the log-sanitization intent in 13 files, and deliberately leaves out `install-process.*`, the recovery-tree work, and the process runner — that runner is where the missing stream `error` handlers, rejecting cleanup promises, and leaked signal listeners lived.

All five substantive objections from this PR's review threads are addressed there and pinned by tests: normal npm-cache symlinks do not block updates, sanitization covers anchorless Windows paths and multi-word usernames, Windows is an explicit tested no-spawn skip, no process-runner crash surface, and no source-text-only tests.

Thanks for the original investigation — the diagnosis is what made the replacement straightforward.

PR #576, opened 2026-07-27
fix(codex): warn about stale Codex app-servers after a catalog write

lidge-jun:
Replaces #527 — same change, rebased onto current `dev` and retargeted.

## Why a new PR

#527 was opened against `codex/catalog-written-signal` instead of `dev`, so `enforce-target` failed by design. Retargeting alone did not fix it: the branch carried two commits, and the first one — `1ba588eff` "report whether a sync actually wrote the catalog or cache" — had already landed on `dev` through #526.

The two are not the same commit. Measured before dropping it:

```
git diff --stat 1ba588eff~1 1ba588eff    6 files changed,  86 insertions(+), 14 deletions(-)
git diff --stat 9dd3c42da~1 9dd3c42da    6 files changed, 208 insertions(+), 15 deletions(-)
git range-diff ... | rg -c differing     150
```

Same six files, but `dev`'s version is 122 lines larger — it is a rewrite that absorbed review feedback, not a rename. Cherry-picking naively and resolving conflicts toward the PR side would have silently reverted those 122 lines, which is what made the original branch dirty in the first place.

So this branch takes `origin/dev` and cherry-picks only `a64aa5856`, the stale app-server warning. `dev` stays canonical for everything `9dd3c42da` touched.

## Verification that nothing was reverted

```
git diff origin/dev --numstat | awk '$2>0'
```

Deletions appear only in files this commit actually modifies (`src/cli/index.ts`, `src/cli/help.ts`, `src/server/management/config-routes.ts`, the six i18n locales, `dashboard-overview-sections.tsx`, `cli.md`). Zero deleted lines under `src/codex/` or in `tests/codex-refresh.test.ts` / `tests/injection-model-api.test.ts` — the two files that were conflicting — so `9dd3c42da`'s content is intact.

## The change itself

Unchanged from #527: warn when a catalog write lands while Codex app-servers are still running with the old catalog, since those processes keep serving a stale model list until restarted.

## Tests

```
bun run typecheck                                          exit 0
bun test tests/codex-refresh.test.ts tests/injection-model-api.test.ts \
         tests/codex-sync-api.test.ts tests/codex-app-server-processes.test.ts \
         tests/codex-models-cache-invalidate.test.ts
                                                           46 pass / 1 skip / 0 fail
```

Planning and evidence: `devlog/_plan/260728_bug_bundle_resolution/040_pr527_rebase.md`.

## Summary by CodeRabbit

* **New Features**
  * Added `--restart-codex` to `ocx sync` and `ocx sync-cache` to refresh stale model lists in long-running Codex sessions.
  * Added clearer warnings and a dashboard hint with the exact recommended restart command.

* **Documentation**
  * Updated CLI reference and troubleshooting docs for stale app-server behavior and restart guidance.

* **Bug Fixes**
  * Improved safety and precision of restart handling to target only matching Codex app-server processes.
  * Ensured stale-process warnings/hints appear only after an actual catalog or cache update.

### DIFF
--- a/docs-site/src/content/docs/guides/codex-integration.md
+++ b/docs-site/src/content/docs/guides/codex-integration.md
@@ -245,6 +245,11 @@ If a model is missing from Codex, or the catalog order/visibility looks wrong, c
    independently of other providers.
 5. **Cache and `ocx sync`** — live catalogs are cached for about five minutes (`modelCacheTtlMs`,
    default `300000`). Run `ocx sync` to force a fresh fetch and rewrite the catalog immediately.
+6. **Running Codex `app-server`** — rewriting the on-disk catalog is not enough while a long-lived
+   Codex `app-server` (Desktop / CLI background host) keeps the previous list in memory. `ocx sync`
+   and `ocx sync-cache` warn when those processes are detected. Restart them with
+   `ocx sync --restart-codex` (or stop the matching `app-server` processes yourself), then let Codex
+   recreate them so the new list appears.
 
 :::caution[Other local writers]
 Catalog writes (`opencodex-catalog.json`, `config.toml`) are atomic **inside** opencodex, which only
--- a/docs-site/src/content/docs/reference/cli.md
+++ b/docs-site/src/content/docs/reference/cli.md
@@ -142,14 +142,21 @@ opencodex local config only if all restore steps succeeded. `remove` is an alias
 
 ## Models & Codex
 
-### `ocx sync`
+### `ocx sync [--restart-codex]`
 
 Fetch the live model list from every configured provider and re-inject the merged catalog into Codex.
 Run it after adding a provider or to refresh available models.
 
-### `ocx sync-cache`
+If long-lived Codex `app-server` processes are still running, `ocx sync` warns that they may keep
+serving the previous in-memory model list even though `opencodex-catalog.json` / `models_cache.json`
+were updated. Pass `--restart-codex` to send `SIGTERM` only to matching `codex … app-server` and
+`codex-code-mode-host` processes owned by the current user (active turns may be interrupted). Broad
+`pkill -f codex` matching is intentionally avoided.
 
-Invalidate Codex's local model picker cache so it is rebuilt from the active opencodex catalog.
+### `ocx sync-cache [--restart-codex]`
+
+Invalidate Codex's local model picker cache so it is rebuilt from the active opencodex catalog. The
+same stale-`app-server` warning and optional `--restart-codex` behavior as `ocx sync` apply.
 
 ### `ocx v2 [subcommand]`
 
--- a/gui/src/i18n/de.ts
+++ b/gui/src/i18n/de.ts
@@ -202,7 +202,7 @@ export const de: Record<TKey, string> = {
   "dash.syncModels": "Modelle synchronisieren",
   "dash.syncing": "Synchronisiere…",
   "dash.syncOk": "Synchronisierung abgeschlossen. {count} Modell(e) angehängt.",
-  "dash.syncStaleHint": "Falls die Codex-App weiterhin eine alte Liste zeigt, starte ihren langlebigen App-Server-Prozess neu.",
+  "dash.syncStaleHint": "Falls Codex weiterhin eine alte Liste zeigt, starte den langlebigen App-Server neu ({cmd}).",
   "dash.syncFailed": "Synchronisierung fehlgeschlagen: {error}",
   "dash.projectConfigTitle": "Projekt-Codex-Konfig umgeht OpenCodex",
   "dash.projectConfigHint": "Diese repo-lokalen Einstellungen überschreiben den OpenCodex-Proxy (z. B. direkt zu OpenCode Go routen). Entferne sie, damit die Routing aus ~/.codex/config.toml in diesem Projekt greift.",
--- a/gui/src/i18n/en.ts
+++ b/gui/src/i18n/en.ts
@@ -212,7 +212,7 @@ export const en = {
   "dash.syncModels": "Sync models",
   "dash.syncing": "Syncing…",
   "dash.syncOk": "Sync complete. {count} model(s) appended.",
-  "dash.syncStaleHint": "If Codex App still shows an older list, restart its long-lived app-server process.",
+  "dash.syncStaleHint": "If Codex still shows an older list, restart its long-lived app-server ({cmd}).",
   "dash.syncFailed": "Sync failed: {error}",
   "dash.projectConfigTitle": "Project Codex config bypasses OpenCodex",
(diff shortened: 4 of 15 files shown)


lidge-jun:
## Rebased onto current `dev` — conflicts resolved per hunk

Force-pushed `935a0e977` → `98142f5c8` (lease base recorded before the rebase; single commit preserved). This was `CONFLICTING`; it is now `MERGEABLE`.

### What actually conflicted

Three files, all logic. The six i18n dictionaries and both docs pages auto-merged:

- `src/cli/index.ts`
- `src/server/management/config-routes.ts`
- `gui/src/pages/dashboard-overview-sections.tsx`

### Resolutions

**`case "sync":`** — `dev` added a `!synced.ok` path (exitCode 1 + a message). This branch added `--restart-codex` parsing and the `catalogWritten || cacheSynced` gate. Both are kept, and deliberately *not* as an if/else: `refreshCodexModelCatalog()` runs before `injectCodexConfig()` in `src/codex/sync.ts`, so a sync can return `ok: false` after the catalog was already rewritten — which is exactly when a long-lived app-server is holding the stale list. Making it an `else` would suppress the warning in the case that needs it most.

**`/api/sync`** — `dev` added `...(result.ok ? {} : { error: result.message })`; this branch replaced the inline hint with `attachStaleAppServerHint(result)`. Composed as `{ ...attachStaleAppServerHint(result), ...(result.ok ? {} : { error: result.message }) }`. The key sets are disjoint (`staleAppServerHint` vs `error`), so neither clobbers the other, and the helper still owns write-gating — no process enumeration on this path.

**Dashboard overview** — `dev` added the `nativeSubagentDefaultsWarning` line; this branch converted the stale hint to `<Trans k="dash.syncStaleHint" cmd="ocx sync --restart-codex" />`. Both lines kept, adjacent. Confirmed `Trans` is imported and all six locales carry the `{cmd}` placeholder it splits on.

### One thing the rebase surfaced, worth flagging

The reverse diff against `dev` showed this branch deleting `grokSyncFailureMessage()` and three of its error handlers in `src/cli/index.ts` — the work from `5451cd191` ("surface a fence that points at a port we are not listening on"). That deletion was already present in `935a0e977`, so it predates the rebase and came from the #527 lineage rather than from conflict resolution.

It is restored. Those three call sites report again instead of silently swallowing, which matters because a stale `~/.grok/config.toml` otherwise keeps naming a dead port with nothing in the log. Nothing in the suite guards that string, so `bun run test` would have stayed green either way — worth a regression test in follow-up.

`src/cli/index.ts` is now additions-only against `dev` apart from the two lines this feature extends.

### Identifier note

`tests/cli-restore-back.test.ts` (on `dev`) and `tests/codex-app-server-processes.test.ts` (in this PR) both read `src/cli/index.ts` as source text and assert on it, and they disagreed on the variable name. Aligned on `dev`'s `synced`, updating this PR's assertions rather than `dev`'s. Both pass now. These source-text assertions are brittle in a way worth revisiting separately — they fail on a pure rename and would pass on a semantic break.

### Verification

- `bun run typecheck` — clean
- `bun run build:gui` — exit 0 (catches i18n key gaps that the root typecheck cannot, since `tsconfig.json` only includes `src`)
- `bun run test` — **5722 pass, 1 skip, 0 fail** across 417 files
- `bun test tests/codex-app-server-processes.test.ts tests/codex-models-cache-invalidate.test.ts` — 22 pass, 1 skip, 0 fail

The `windows-latest` failure previously visible on this PR was an unrelated timeout in `tests/usage-debug.test.ts`; that is addressed separately in #614.

lidge-jun:
## Rebased onto current `dev` — conflicts resolved per hunk

`935a0e977` → `98142f5c8`, still a single commit, now sitting directly on `7710185c0`. This was `CONFLICTING`; it is `MERGEABLE` again.

### What actually conflicted

Three logic files, not the i18n/docs set I expected going in:

- `src/cli/index.ts`
- `src/server/management/config-routes.ts`
- `gui/src/pages/dashboard-overview-sections.tsx`

The six locale dictionaries and both docs pages auto-merged.

### How each was resolved

**`src/cli/index.ts` — `case "sync"`.** `dev` added a `!ok` error path (exit code 1 + message); this PR adds `--restart-codex` parsing and the write-gated `afterCatalogWriteHandleAppServers` call. Both kept, deliberately as sequential `if`s rather than `if/else`. `refreshCodexModelCatalog` runs before `injectCodexConfig` in `src/codex/sync.ts`, so `ok: false` with `catalogWritten: true` is reachable — and that is precisely when an app-server is holding the stale list. Making it an `else` would suppress the warning in the case that needs it most.

**`src/server/management/config-routes.ts` — `/api/sync`.** `dev` added `error: result.message` on failure; this PR moved the hint into `attachStaleAppServerHint(result)`. Composed as `{ ...attachStaleAppServerHint(result), ...(result.ok ? {} : { error: result.message }) }`. The key sets are disjoint, so neither clobbers the other, and the helper keeps owning the write gate — no process enumeration on this path.

**`gui/src/pages/dashboard-overview-sections.tsx`.** `dev` added a `nativeSubagentDefaultsWarning` line; this PR changed the stale hint from `t(...)` to `<Trans k="dash.syncStaleHint" cmd="ocx sync --restart-codex" />`. Both lines kept adjacent. Verified `Trans` is imported and all six locales carry the `{cmd}` placeholder the component splits on.

### Two things the review caught

**A silent revert.** The reverse diff against `dev` showed this commit deleting `grokSyncFailureMessage` and three `catch` handlers from `src/cli/index.ts`, restoring silent-swallow behavior on grok fence sync failures. Those come from `5451cd191`, and no test asserts the message, so the full suite would have stayed green while `~/.grok/config.toml` went back to silently pointing at a dead port.

Traced it: the deletion was already in `935a0e977`, inherited from the #527 lineage — the rebase carried it forward faithfully. Restored via `git checkout origin/dev -- src/cli/index.ts` and reapplied only the `case "sync"` and `case "sync-cache"` hooks. The reverse diff on that file is now additions plus one replaced line.

**An identifier collision between two source-text tests.** `tests/cli-restore-back.test.ts` asserts the literal `if (!synced.ok)`, while this PR's `tests/codex-app-server-processes.test.ts` asserted `syncResult.catalogWritten`. Both read `src/cli/index.ts` as a string, so no single naming satisfies both as written. Kept `dev`'s `synced` and aligned this PR's test to it. Worth noting these assertions are shape checks wearing behavioral clothing — they fail on a pure rename and would pass on a semantically broken gate that stayed textually intact. Not something to fix here.

### Verification

- `bun run typecheck` — clean
- `bun run build:gui` — exit 0
- `bun test tests/codex-app-server-processes.test.ts tests/codex-models-cache-invalidate.test.ts` — 22 pass, 1 skip, 0 fail
- `bun run test` — 5722 pass, 1 skip, 0 fail across 417 files
- Push went through the prepush gate (typecheck + lint:gui + test + privacy:scan + doctor:gui)

PR #577, opened 2026-07-27
feat(images): Grok image bridge (maintainer takeover of #424)

Wibias:
## Summary
Maintainer takeover of #424 (Grok image bridge) that folds in stacked #528 and clears remaining review blockers on latest `dev`.

Credit: original feature work by @tizerluo on #424. This PR is the maintainer landing vehicle after rebasing, conflict resolution, and hardening.

### Included
- Full image-bridge feature from #424 + Codex/CodeRabbit/P2 hardening from #528
- Artifact retention (`artifactsKeepCount`) with post-batch prune (no mid-batch path wipe), `<=0` disables prune, Windows-safe paths, full PNG signature check
- Hosted-only activation (ordinary `function` tools named `image_gen` no longer arm the bridge)
- Type-only `tool_choice: { type: "image_generation" }` maps to required synthetic `image_gen`
- API-key-only xAI Images fulfillment (OAuth / Grok CLI proxy deliberately does not arm the bridge)
- Ordered multi-block thinking + redacted_thinking replay
- DNS-pinned HTTPS artifact downloads (SSRF / rebinding)
- Docs: opt-in bridge, API-key auth, retention knobs, Codex hosted vs built-in `image_gen` distinction

Closes #424
Closes #528

## Test plan
- [x] `bun test --isolate tests/images/ tests/responses-parser.test.ts`
- [x] `bun run typecheck`
- [ ] Cross-platform CI green
- [ ] Security review of image egress / SSRF / auth
- [ ] Fresh CodeRabbit + Codex review on this tip

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in Image Bridge to route eligible hosted image-generation calls through xAI for non-OpenAI models.
  * Added support for image generation/editing results from both base64 and secure HTTPS URLs, saved as local artifacts.
  * Introduced configurable bridge options, including model selection, timeouts/round limits, and artifact retention/pruning.

* **Documentation**
  * Added an “Image Bridge” guide and clarified how `image_gen` behaves versus the Responses-based bridge flow.

* **Bug Fixes**
  * Strengthened SSRF protection and safety checks for IPv6 and DNS-resolved destinations, plus safer HTTPS image download handling.

PR #582, opened 2026-07-28
feat(videos): add Grok video bridge for non-OpenAI models

tizerluo:
## Summary

Extends the image bridge (#577) to support asynchronous video generation via xAI Grok Imagine Video. When `videoBridgeEnabled: true`, a synthetic `video_gen` tool is injected into the conversation. The model calls it like any function; opencodex intercepts the call, submits a video generation job to xAI, polls until completion, and downloads the result to disk.

**Architecture:** submit → poll with heartbeats → download to artifact → inject as tool_result → loop (max 2 rounds).

## Changes

**New files:**
- `src/images/xai-video-client.ts` — async xAI video client (`submitVideoJob` + `pollVideoJob`)
- `src/images/fulfill-video.ts` — arg parsing, heartbeat polling generator, result builder
- `tests/videos/` — 32 tests across 3 files
- `docs-site/.../guides/video-bridge.md` — user guide

**Modified files:**
- `src/images/loop.ts` — unified media bridge loop (image + video dispatch), per-turn call caps (`MAX_VIDEO_CALLS_PER_TURN = 3`)
- `src/images/types.ts` — `VideoBridgePlan`, `VideoCallResult`
- `src/images/synthetic-tool.ts` — `buildVideoTool`, `VIDEO_GEN_TOOL_NAME`
- `src/images/plan.ts` — `planVideoBridge` (opt-in, API-key-only, registry-pinned baseUrl)
- `src/images/artifacts.ts` — `downloadVideoToArtifact` (200 MiB cap, streaming, SSRF-protected via pinned HTTPS)
- `src/server/responses/core.ts` — media bridge wiring with tool dedup
- `src/types.ts` — `videoGeneration` flag, `videoBridgeEnabled`/`videoBridgeModel`/`videoMaxRounds`/`videoTimeoutMs`

## Security

- **SSRF protection**: `downloadVideoToArtifact` reuses the same `assessUrlDestination` → `resolvePublicAddresses` → `pinnedHttpsGet` pattern as the hardened image downloader
- **Size caps**: 200 MiB hard cap on video downloads, 1 MiB on API responses
- **Per-turn limits**: `MAX_VIDEO_CALLS_PER_TURN = 3` prevents unbounded paid calls
- **Abort linking**: client cancel propagates to in-flight xAI submit/poll/download

## Configuration

```json
{
  "images": {
    "bridgeEnabled": true,
    "videoBridgeEnabled": true,
    "videoBridgeModel": "grok-imagine-video",
    "videoMaxRounds": 2,
    "videoTimeoutMs": 300000
  }
}
```

Video bridge is opt-in (`videoBridgeEnabled` defaults to `false`). Coexists with the image bridge.

## Test results

- `tsc --noEmit`: clean
- `tests/videos/`: 32 pass, 0 fail
- `tests/images/` (regression): all pass individually (pre-existing `mock.module` pollution when run together is unchanged)

## Summary by CodeRabbit

* **New Features**
  * Added optional Video Bridge to generate videos via xAI Grok Imagine Video, including async progress/heartbeats, job polling, and video artifact downloads.
  * Introduced a synthetic video tool with configurable enablement, model, max rounds, and timeouts (plus prompt/duration/resolution/aspect ratio support).
* **Bug Fixes**
  * Improved combined image+video orchestration, including safer stream handling, tool-choice/media routing, and stricter video size/format validation.
* **Documentation**
  * Added a Video Bridge guide with prerequisites and operational limits.
* **Tests**
  * Added Bun coverage for video planning, tool-arg parsing, job submission/polling, timeouts, and result formatting.

### DIFF
--- a/docs-site/src/content/docs/guides/video-bridge.md
+++ b/docs-site/src/content/docs/guides/video-bridge.md
@@ -12,7 +12,7 @@ job to xAI, polls until completion, and downloads the result.
 
 ## Prerequisites
 
-- An xAI account with an API key (`ocx login xai` or set the key in your provider config)
+- An xAI account with an API key (set `XAI_API_KEY` or configure the key in your provider config — `ocx login xai` alone is not sufficient for the video bridge, which requires API key auth)
 - A non-OpenAI model as your routed provider (e.g. Anthropic Claude, Google Gemini)
 - opencodex configured to route through the non-OpenAI provider
 
@@ -28,6 +28,7 @@ Add `videoBridgeEnabled: true` to your `images` config:
     "videoBridgeModel": "grok-imagine-video",
     "videoMaxRounds": 2,
     "videoTimeoutMs": 300000
+  }
 }
 ```
 
--- a/docs-site/src/content/docs/guides/video-bridge.md
+++ b/docs-site/src/content/docs/guides/video-bridge.md
@@ -1,0 +1,67 @@
+---
+title: Video Bridge
+description: Generate videos with Grok Imagine Video through a non-OpenAI model.
+---
+
+## Overview
+
+The Video Bridge lets you use xAI's Grok Imagine Video generation through any non-OpenAI model
+routed by opencodex. When enabled, a synthetic `video_gen` tool is injected into the conversation.
+The model calls it like any function tool; opencodex intercepts the call, submits a video generation
+job to xAI, polls until completion, and downloads the result.
+
+## Prerequisites
+
+- An xAI account with an API key (`ocx login xai` or set the key in your provider config)
+- A non-OpenAI model as your routed provider (e.g. Anthropic Claude, Google Gemini)
+- opencodex configured to route through the non-OpenAI provider
+
+## Configuration
+
+Add `videoBridgeEnabled: true` to your `images` config:
+
+```json
+{
+  "images": {
+    "bridgeEnabled": true,
+    "videoBridgeEnabled": true,
+    "videoBridgeModel": "grok-imagine-video",
+    "videoMaxRounds": 2,
+    "videoTimeoutMs": 300000
+}
+```
+
+| Option | Default | Description |
+|--------|---------|-------------|
+| `videoBridgeEnabled` | `false` | Master switch. Must be explicitly enabled. |
+| `videoBridgeModel` | `"grok-imagine-video"` | xAI video model id. |
+| `videoMaxRounds` | `2` | Max video-gen rounds before forced final answer. |
+| `videoTimeoutMs` | `300000` (5 min) | Per-video timeout including polling. |
+
+## How It Works
+
+1. opencodex detects a non-OpenAI routed model with `videoBridgeEnabled: true`
+2. A synthetic `video_gen` function tool is injected into the conversation
+3. When the model calls `video_gen`, opencodex submits a job to xAI's `/videos/generations`
+4. The bridge polls the job status every 5-15 seconds, sending heartbeat messages to keep the stream alive
+5. When the video is ready, it's downloaded to the artifacts directory
+6. The local file path is returned to the model as a tool result
+
+## Supported Parameters
+
+The `video_gen` tool accepts:
+
+| Parameter | Type | Range | Description |
+|-----------|------|-------|-------------|
+| `prompt` | string | required | Detailed video generation prompt |
+| `duration` | integer | 1-15 | Video length in seconds |
+| `resolution` | string | `"480p"`, `"720p"` | Video resolution |
+| `aspect_ratio` | string | 7 ratios | `16:9`, `9:16`, `1:1`, `4:3`, `3:4`, `3:2`, `2:3` |
+
+## Limitations
+
+- **xAI only**: Video generation is only available through xAI's Grok Imagine Video API
+- **Asynchronous**: Video generation takes 30-120 seconds
+- **Cost**: Video generation is a paid xAI feature (~$0.05/sec @480p, ~$0.07/sec @720p)
+- **One video per call**: Each `video_gen` call produces one video
+- **Coexists with Image Bridge**: Both bridges can be enabled simultaneously
--- a/docs-site/src/content/docs/guides/video-bridge.md
+++ b/docs-site/src/content/docs/guides/video-bridge.md
@@ -81,5 +81,5 @@ The `video_gen` tool accepts:
 - **Cost**: Video generation is a paid xAI feature (~$0.05/sec @480p, ~$0.07/sec @720p)
 - **One video per call**: Each `video_gen` call produces one video
 - **Coexists with Image Bridge**: Both bridges can be enabled simultaneously
-- **Web search priority**: When web search is active for a turn, the video bridge is skipped (web search and media bridging cannot run concurrently). A `console.warn` is emitted so you can detect this in logs.
+- **Web search priority**: When a web search sidecar is active for a turn (non-`runTurn` adapter), the video bridge is skipped — the two cannot run concurrently. A `console.warn` is emitted so you can detect this in logs.
 - **Timeout covers submit + poll**: The `videoTimeoutMs` budget starts before job submission, so the submit call (60 s) and subsequent polling share the same deadline.
--- a/docs-site/src/content/docs/guides/video-bridge.md
+++ b/docs-site/src/content/docs/guides/video-bridge.md
@@ -12,9 +12,24 @@ job to xAI, polls until completion, and downloads the result.
 
 ## Prerequisites
 
-- An xAI account with an API key (set `XAI_API_KEY` or configure the key in your provider config — `ocx login xai` alone is not sufficient for the video bridge, which requires API key auth)
+- An `xai` provider entry with an **API key** (`ocx login xai` alone is not sufficient — the video bridge requires key auth, not OAuth)
 - A non-OpenAI model as your routed provider (e.g. Anthropic Claude, Google Gemini)
 - opencodex configured to route through the non-OpenAI provider
+
+> **⚠ Provider key required:** The video bridge only activates when the `xai` provider uses
+> API key auth. Add this to your config:
+>
+> ```json
+> {
+>   "providers": {
+>     "xai": { "adapter": "openai-chat", "apiKey": "xai-…", "authMode": "key" }
+>   }
+> }
+> ```
+>
+> If you onboarded via `ocx login xai` (OAuth), the provider stays in `authMode: "oauth"`
+> and the bridge silently won't activate. Set `XAI_API_KEY` in the environment **or**
+> hard-code the key as shown above.
 
 ## Configuration
 
--- a/docs-site/src/content/docs/guides/video-bridge.md
+++ b/docs-site/src/content/docs/guides/video-bridge.md
@@ -81,3 +81,5 @@ The `video_gen` tool accepts:
 - **Cost**: Video generation is a paid xAI feature (~$0.05/sec @480p, ~$0.07/sec @720p)
 - **One video per call**: Each `video_gen` call produces one video
 - **Coexists with Image Bridge**: Both bridges can be enabled simultaneously
+- **Web search priority**: When web search is active for a turn, the video bridge is skipped (web search and media bridging cannot run concurrently). A `console.warn` is emitted so you can detect this in logs.
+- **Timeout covers submit + poll**: The `videoTimeoutMs` budget starts before job submission, so the submit call (60 s) and subsequent polling share the same deadline.
--- a/src/images/artifacts.ts
+++ b/src/images/artifacts.ts
@@ -477,13 +477,24 @@ export async function downloadImageToArtifact(
 }
 
 const MAX_VIDEO_DOWNLOAD_BYTES = 200 * 1024 * 1024; // 200 MiB
+/** Aggregate per-turn video download cap (600 MiB = 3 × max single download). */
+const MAX_VIDEO_BYTES_PER_TURN = MAX_VIDEO_DOWNLOAD_BYTES * 3;
 
 export interface VideoBudget {
   spent: number;
+  /** Ceiling on total bytes across all downloads this turn. */
+  cap: number;
 }
 
 export function createVideoBudget(): VideoBudget {
-  return { spent: 0 };
+  return { spent: 0, cap: MAX_VIDEO_BYTES_PER_TURN };
+}
+
+/** Charge bytes to the budget; returns false if the ceiling would be exceeded. */
+export function chargeVideoBudget(budget: VideoBudget, bytes: number): boolean {
+  if (budget.spent + bytes > budget.cap) return false;
+  budget.spent += bytes;
+  return true;
 }
 
 export function guessVideoExtFromMagic(bytes: Uint8Array): string {
@@ -514,7 +525,9 @@ export async function downloadVideoToArtifact(
     const isBase64 = meta.includes(";base64");
     if (!isBase64) throw new Error("non-base64 data URI for video is not supported");
     const buf = Buffer.from(data, "base64");
-    if (budget) budget.spent += buf.byteLength;
+    if (budget && !chargeVideoBudget(budget, buf.byteLength)) {
+      throw new Error("video data URI exceeds per-turn download budget");
+    }
     if (buf.byteLength > MAX_VIDEO_DOWNLOAD_BYTES) throw new Error("video data URI exceeds size cap");
     const ext = guessVideoExtFromMagic(buf);
     const dir = getArtifactsDir();
@@ -564,7 +577,9 @@ export async function downloadVideoToArtifact(
     fh = await open(dest, "w", 0o600);
     // Write first chunk and set up accounting
     let totalBytes = first.value.byteLength;
-    if (budget) budget.spent += totalBytes;
+    if (budget && !chargeVideoBudget(budget, totalBytes)) {
+      throw new Error("video download exceeds per-turn budget");
+    }
     if (totalBytes > MAX_VIDEO_DOWNLOAD_BYTES) {
       throw new Error("video download exceeds size cap");
     }
@@ -573,7 +588,9 @@ export async function downloadVideoToArtifact(
       const { value, done } = await reader.read();
       if (done) break;
       totalBytes += value.byteLength;
-      if (budget) budget.spent += value.byteLength;
+      if (budget && !chargeVideoBudget(budget, value.byteLength)) {
+        throw new Error("video download exceeds per-turn budget");
+      }
       if (totalBytes > MAX_VIDEO_DOWNLOAD_BYTES) {
         throw new Error("video download exceeds size cap");
       }
--- a/src/images/artifacts.ts
+++ b/src/images/artifacts.ts
@@ -548,7 +548,7 @@ export async function downloadVideoToArtifact(
   if (assessment && assessment.kind !== "public" && assessment.kind !== "hostname") {
     throw new Error(`video URL targets ${assessment.detail}`);
   }
-  const resolved = await resolvePublicAddresses(url);
+  const resolved = await resolvePublicAddresses(url, "video");
   const pinned = pickPinnedAddress(resolved.addresses);
   const resp = await pinnedHttpsGet(url, pinned, signal, { maxBytes: MAX_VIDEO_DOWNLOAD_BYTES });
   if (!resp.ok) {
--- a/src/images/artifacts.ts
+++ b/src/images/artifacts.ts
@@ -1,5 +1,5 @@
 import { readdirSync, readFileSync, statSync, unlinkSync, existsSync } from "node:fs";
-import { mkdir, writeFile } from "node:fs/promises";
+import { mkdir, writeFile, open, unlink } from "node:fs/promises";
 import type { IncomingMessage } from "node:http";
 import https from "node:https";
 import type { RequestOptions } from "node:https";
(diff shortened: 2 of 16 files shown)


tizerluo:
## Review response (HEAD: `47225638`)

All Codex P1/P2 and CodeRabbit findings addressed on tip:

### P1 fixes
- **Non-streaming requests**: Video bridge no longer turns every `stream:false` request into a 400. Only the image bridge (which detects a hosted `image_generation` tool) requires streaming. Video-only non-streaming requests skip the bridge entirely and proceed normally.
- **Test fetch leak**: `globalThis.fetch` is now saved and restored in `afterEach` for `xai-video-client.test.ts`.

### P2 fixes
| Finding | Fix |
|---|---|
| Video timeout not clamped | Uses `vidPlan.timeoutMs` (clamped in plan) instead of raw config |
| `videoMaxRounds` ignored when both active | Uses `Math.min(image, video)` when both plans are active |
| Unknown magic defaults to mp4 | `guessVideoExtFromMagic` now throws for unrecognized formats |
| 4xx poll failures retry forever | 4xx (except 429) returns permanent failure immediately |
| Markdown path not portable | Uses `pathToFileURL(path).href` matching `fulfillImageCall` |
| Initial poll interval too fast | Restored to 5s (was 200ms) |
| done-without-URL spins | Returns terminal error `"no video URL was returned"` |
| VideoBudget partial accounting | Every streamed chunk now charged to budget |
| Reader leak on file-open failure | Reader cancelled/released in all error paths |
| `readBoundedText` body not cancelled | `reader.cancel()` added to `finally` |
| Namespaced `video_gen` stripped | Dedup now checks `namespacedToolName` for video tools |
| Docs JSON missing `}` | Fixed |
| Docs auth prerequisite | Clarified API key requirement |

Typecheck clean. All 33 video tests + 6 handler-activation tests pass.

Wibias (changes requested):

**Verdict: REQUEST CHANGES**

Cross-platform CI is green. Security review found **no medium+** issues (SSRF/download path matches the hardened image bridge: destination policy + pinned HTTPS + no redirects; API-key-only; registry-pinned xAI baseUrl; size caps; safe artifact paths).

Two fix commits already landed most of the first Codex/Rabbit pass (`47225638`, `5e09f326`). Remaining blockers are still present on `5e09f326`.

### Open Codex / CodeRabbit (verified against HEAD)

| Item | Status |
| --- | --- |
| Codex P1: non-streaming 400 when video enabled | **Fixed** — video-only + `stream:false` skips injection |
| Codex P1: video lost when web search wins | **Still open** — `(imgPlan \|\| vidPlan) && (!wsPlan \|\| adapter.runTurn)` still drops video on common Codex+web-search turns |
| Codex P2: `videoTimeoutMs` covers submit+poll | **Still open** — poll deadline starts after `submitVideoJob`; submit has its own 60s |
| Codex P2: pass `vidPlan.timeoutMs` | **Fixed** |
| Codex P2: honor `videoMaxRounds` with both bridges | **Fixed** (`Math.min`) |
| Codex P2: reject unknown video magic | **Fixed** |
| Codex P2: stop retrying permanent poll 4xx | **Fixed** |
| Codex P2: `pathToFileURL` markdown | **Fixed** |
| Codex P2: namespaced `video_gen` dedup | **Fixed** (filter); aliases still not intercepted |
| Codex P2: poll starts at 5s | **Fixed** |
| Codex P2: cancel reader on open failure / oversized body | **Fixed** |
| Codex P2 / Rabbit: docs JSON braces | **Fixed** |
| Codex P1: restore `globalThis.fetch` in tests | **Fixed** |
| Rabbit critical: OAuth never activates | **Invalid / by design** — API-key-only; docs + test cover this |
| Rabbit: `done` without `videoUrl` spins | **Fixed** |
| Rabbit: `paidVideoCalls` before parse | **Fixed** |
| Rabbit: encode `requestId` | **Fixed** |
| Rabbit: errors carry `.status` | **Fixed** |
| Rabbit: prune video artifacts | **Fixed** |
| Rabbit: `VideoBudget` never enforces a turn ceiling | **Still open** — `spent` is updated, but no aggregate per-turn byte cap like images |

### Required before merge
1. **Web search coexistence:** either compose video with `runWithWebSearch`, or document the same priority rule as the image bridge in `video-bridge.md` *and* surface a clear runtime signal when `vidPlan` is dropped because of `wsPlan` (silent loss of a config-enabled capability is the P1).
2. **Timeout budget:** start one deadline before submit and bind submit+poll (and ideally download) to the remaining budget, matching `VideoBridgePlan.timeoutMs` docs — or change the docs/type comment if poll-only is intentional.
3. **Optional but preferred:** enforce an aggregate `VideoBudget` ceiling (parity with image `chargeImageBudget`), and add tests for `done`-without-URL + encoded `requestId`.

### Security
No medium+ findings. Same trust model as the image CDN downloader; OAuth-only xAI correctly does not arm the bridge.

tizerluo:
Thanks for the thorough review! All three items addressed in `bb543844`:

**1. Web search coexistence (P1)**
Documented the priority rule in `video-bridge.md` and added a `console.warn` runtime signal when `vidPlan` is dropped because of `wsPlan`. The media bridge cannot run alongside `runWithWebSearch` without a compose layer; making the skip visible is the right tradeoff for now.

**2. Timeout budget (P2)**
The deadline is now captured *before* `submitVideoJob` (line 613). The poll generator receives `remainingMs = max(5_000, deadline - Date.now())` so submit (60 s) + polling share one `videoTimeoutMs` budget. Matches the type comment on `VideoBridgePlan.timeoutMs`.

**3. VideoBudget ceiling (optional, done)**
`VideoBudget` now carries a `cap` (600 MiB = `MAX_VIDEO_DOWNLOAD_BYTES × 3`). `chargeVideoBudget()` is called per-chunk during streaming download; exceeding the ceiling throws mid-download and the partial file is cleaned up in the catch block — same pattern as `chargeImageBudget`.

**New tests** (36 total, all passing):
- `done` without `videoUrl` → returns error, not infinite poll
- Permanent 4xx (401) poll error → stops immediately, no retry storm
- `requestId` encoding → special characters encoded in poll URL

Wibias (changes requested):
## Maintainer re-review (post `bb543844`)

**Verdict: REQUEST CHANGES**

Security: still **no medium+**. Cross-platform CI mostly green (windows-latest still pending at review time). Focused `tests/videos` locally: xAI client suite passes; fulfill/plan files need `bun install` in this worktree (zod/v4) — rely on CI for full gate.

### Prior required items vs `bb543844`

| Item | Status |
| --- | --- |
| Web search coexistence (docs + runtime warn) | **Fixed** — warn gated on `canRunWebSearch` (`wsPlan && !adapter.runTurn`) at `core.ts:1601-1608`; docs note priority |
| Timeout budget submit+poll | **Partial** — deadline clock starts before submit (`loop.ts:613`), but submit is still only bound to client abort + fixed `SUBMIT_TIMEOUT_MS` (60s); poll uses `Math.max(5_000, remaining)` which can extend past the deadline |
| VideoBudget aggregate ceiling | **Fixed** — `cap` + `chargeVideoBudget` per chunk |

Author response matches intent; timeout claim in docs (`video-bridge.md:85`) is still stronger than runtime.

### New CodeRabbit findings (verified)

| Finding | Verdict | Action |
| --- | --- | --- |
| **B.** Gate image `tool_choice` rewrite on `imgPlan` (`core.ts:1638-1647`) | **Valid / required** | Video-only still rewrites `image_generation`/`image_gen` → `image_gen` even though `buildImageTool()` was not injected. Strict upstreams can 400 on undeclared `allowedTools`. Gate both allowedTools + named branches on `imgPlan`. |
| **C.** Shared `videoTimeoutMs` incomplete (`loop.ts:613-624` + `xai-video-client.ts:78-79`) | **Valid / required** | Create a deadline-bound signal before submit, pass it into `submitVideoJob`, skip/fail poll when remaining ≤ 0 (no 5s floor on expired budget). Keep docs line 85 only after that, or document independent submit/poll timeouts. |
| **A.** Buffer ≥12 bytes before magic sniff (`artifacts.ts:570-574`) | **Valid / preferred** | Real failure mode if first `reader.read()` is short; not the common path. Buffer until sniff minimum, then charge/write all buffered chunks. |
| **D.** Scope web-search docs to sidecar path | **Optional docs** | Runtime warn is already correctly scoped. Soften docs “when web search is active” → runnable sidecar / non-`runTurn` path if you want parity with the code comment at `core.ts:1591-1595`. |

### Also noted (non-blocking)
- Heartbeat `message` still dropped in the loop (keepalive only) — UX.
- Heartbeat test still sleeps real 5s intervals — CI cost only.
- Video `tool_choice` aliases are not remapped the way image aliases are — follow-up.

### Required before merge
1. Gate image tool-choice rewriting on `imgPlan` (Rabbit B).
2. Finish the shared timeout: deadline-bound signal into submit + no poll floor after expiry; align docs (Rabbit C).

Preferred: buffer magic sniff (A). Optional: docs wording (D).

tizerluo:
All items addressed in `a9069236`:

**Required B — Gate tool_choice on imgPlan** ✅
Both the `allowedTools` map and the named `tc.name` rewrite are now gated behind `if (imgPlan && ...)`. In a video-only turn, `image_generation`/`image_gen` aliases pass through untouched.

**Required C — Shared timeout complete** ✅
- `AbortSignal.timeout(videoTimeout)` is created before `submitVideoJob` and linked with the client signal via `AbortSignal.any()`. Submit is bound to the shared deadline, not just `SUBMIT_TIMEOUT_MS`.
- After submit returns, `remainingMs = videoDeadline - Date.now()`. If `remainingMs <= 0`, poll is skipped entirely with a timeout error — no 5s floor on an expired budget.
- Poll receives only the actual remaining budget.

**Preferred A — Buffer magic sniff** ✅
`downloadVideoToArtifact` now accumulates chunks in a loop until it has ≥12 bytes before calling `guessVideoExtFromMagic`. All buffered bytes are written + charged in one pass.

**Optional D — Docs wording** ✅
Clarified: "When a web search sidecar is active for a turn (non-`runTurn` adapter)..."

36 tests pass, typecheck clean.

Wibias (changes requested):
## Maintainer re-review (post `a906923`)

**Verdict: REQUEST CHANGES**

HEAD: `a9069236`. Required B / preferred A / optional D from the prior review are fixed. CodeRabbit's pass on `a906923` found nothing new; no new Codex review after that commit. Cross-platform CI mostly green (`windows-latest` pending at review time).

### Prior required items vs `a906923`

| Item | Status |
| --- | --- |
| **B.** Gate image `tool_choice` on `imgPlan` | **Fixed** — both `allowedTools` and named branches gated |
| **C.** Shared `videoTimeoutMs` | **Still incomplete** — submit is deadline-bound and the post-submit `remainingMs <= 0` floor is gone, but poll still ignores the deadline signal |
| **A.** Buffer ≥12 bytes before magic sniff | **Fixed** |
| **D.** Web-search docs wording | **Fixed** |

### Required before merge

**Finish shared timeout through poll** (`loop.ts` ~633 + `fulfill-video.ts` / `xai-video-client.ts`)

`linkedDeadline` is passed into `submitVideoJob`, but `pollVideoWithHeartbeats(..., signal, remainingMs)` still receives only the client abort signal. An in-flight `pollVideoJob` can run up to `POLL_TIMEOUT_MS` (30s) and `sleep(interval)` up to 15s after the wall-clock budget is gone. Docs still claim submit+poll share one deadline.

Fix: pass `linkedDeadline` (or `AbortSignal.any([signal, AbortSignal.timeout(remainingMs)])`) into the poll generator so poll fetch + sleep abort when the budget ends. Optionally clamp each poll request timeout to `min(POLL_TIMEOUT_MS, remaining)`.

### Bot triage (not blocking this round)

| Finding | Verdict |
| --- | --- |
| Codex P1 web-search drops video | Accepted / by design — docs + `console.warn` |
| Codex P2 poll interval / body cancel / docs JSON / fetch restore | Fixed |
| Rabbit OAuth never activates | Invalid / by design — API-key-only |
| Video aliases / test `authMode: "api_key"` / DNS "image URL" noun / heartbeat real sleep | Follow-up / trivial |

tizerluo:
Fixed in `c613b05e`:

**Shared timeout through poll** ✅
- `pollVideoWithHeartbeats` now receives `linkedDeadline` (the deadline-bound signal), not the raw client signal
- In-flight `pollVideoJob` fetches abort when the budget expires (was: could run up to `POLL_TIMEOUT_MS` = 30s past deadline)
- `sleep(interval)` aborts on deadline expiry (was: could sleep up to 15s past deadline)
- Abort error messages now distinguish deadline-expiry from client-cancel using an elapsed-time threshold (95% of `timeoutMs`)

Submit + poll + sleep all share one `videoTimeoutMs` deadline. Docs claim matches runtime. 36 tests pass, typecheck clean.

Wibias (changes requested):
## Maintainer re-review (post `c613b05`)

**Verdict: REQUEST CHANGES**

HEAD: `c613b05e`. The prior required item (pass `linkedDeadline` into poll) is in place — submit, poll fetch, and sleep now share one deadline signal. CodeRabbit's pass on `c613b05` raised two more findings (verified below). No new Codex review on this tip.

### Prior required vs `c613b05`

| Item | Status |
| --- | --- |
| Shared timeout through poll (`linkedDeadline` into poll generator) | **Fixed** — `loop.ts` passes `linkedDeadline`; poll fetch + sleep abort on budget expiry |
| Timeout / abort classification | **Still incomplete** — see #1 |
| Batch prune vs delivered paths | **Still incomplete** — see #2 (CodeRabbit Major, verified) |

### Required before merge

**1. Classify deadline aborts without the 0.95 heuristic** (`fulfill-video.ts` ~109–128, **`loop.ts` ~652–655**)

The new elapsed≥95% check mislabels a real deadline abort as "client closed" when submit consumed most of the budget (e.g. 58s of a 60s `videoTimeoutMs` → `remainingMs≈2s` → deadline fires with elapsed ≪ 95% of remaining). Bun's `AbortSignal.timeout` / `AbortSignal.any` propagate `TimeoutError` on `signal.reason` — use that (or `deadlineSignal.aborted` declared outside the `try`) instead of a wall-clock fraction.

**CodeRabbit (verified) — submit catch at `loop.ts:652-655`:** `submitVideoJob` already receives `linkedDeadline`, but the catch only checks the client `signal`. A configured-timeout abort during submit returns a generic abort string instead of `video generation timed out after Ns`, skipping the timeout-specific handling used during polling. Declare `deadlineSignal` outside the `try` and when `deadlineSignal.aborted` (and the client signal is not), return the same timeout error used for poll expiry.

**2. Don't hand the model pruned-away artifact paths** (**`loop.ts` ~696–698**, CodeRabbit Major — verified)

Deferring `pruneArtifacts` until after the batch is necessary but not sufficient when `artifactsKeepCount` is smaller than that batch (e.g. two `video_gen` successes with keepCount 1). Prune can delete the older file before `messages` are built, so the tool_result still reports `ok: true` with a dead path. Image fulfillment already filters via `retainAfterBatch` + `existsSync` in `fulfill.ts`. Reuse that pattern (export/share `retainAfterBatch`, or prune then drop/rewrite results whose paths no longer exist) so delivered results never point at deleted files.

Note: the "video keepCount applied to images" angle is overstated — both plans read the same `config.images.artifactsKeepCount`. The stale-path failure mode is still real and should be fixed here.

**3. Wire video aliases or remove the dead helpers** (`synthetic-tool.ts` `VIDEO_GEN_NAMES` / `isVideoGenName`, `plan.ts`, `loop.ts`, `core.ts`)

`isVideoGenName` is defined but unused. Seed `videoPlan.toolNames` with `VIDEO_GEN_TOOL_NAME` plus any unnamespaced request tools that match `isVideoGenName` (image-bridge parity), and intercept those names in the loop. While touching the filter: only strip by leaf name when `!t.namespace` so a namespaced MCP `video_gen` is not removed.

**4. Fix plan-video test fixtures** (`tests/videos/plan-video.test.ts`)

`authMode: "api_key"` is not a valid `OcxProviderConfig` value (`"key" | "forward" | "oauth" | "local"`). Casts hide it today; use `"key"`.

### Also fix while here (small, same PR)

- **DNS error noun**: `resolvePublicAddresses` hardcodes "image URL …" and surfaces that on video CDN failures. Add an optional noun (default `"image"`) and pass `"video"` from `downloadVideoToArtifact`.

### Not blocking

| Finding | Verdict |
| --- | --- |
| Web search wins over video | Accepted / by design — docs + warn |
| Heartbeat test real 5s sleep | CI cost only |

tizerluo:
All items addressed in `938f25eb`:

**1. Deadline abort classification** ✅
- Removed the 0.95 wall-clock heuristic. Now uses `signal.reason.name === "TimeoutError"` (Bun propagates `TimeoutError` on `AbortSignal.timeout` aborts).
- `deadlineSignal` declared outside the `try` so the catch can inspect `deadlineSignal.aborted`.
- Submit catch: when `deadlineSignal.aborted && !signal.aborted`, returns `video generation timed out after Ns`.

**2. Pruned artifact paths** ✅
- After `pruneArtifacts`, scan `fulfilled[]` and rewrite any result whose `path` no longer exists to `{ ok: false, error: "artifact was pruned before delivery (increase artifactsKeepCount)" }`.

**3. Video aliases wired** ✅
- `planVideoBridge` seeds `toolNames` from `VIDEO_GEN_TOOL_NAME` + any request function tools matching `isVideoGenName`.
- `core.ts` filter only strips **unnamespaced** video_gen aliases — namespaced MCP `video_gen` is preserved.
- `isVideoGenName` is now actually used (was dead code).

**4. Test fixture** ✅
- `authMode: "api_key"` → `"key"` in both `makeConfig` and `makeProvider`.

**5. DNS error noun** ✅
- `resolvePublicAddresses(url, noun="image")` — `downloadVideoToArtifact` passes `"video"`.

36 tests pass, typecheck clean.

Wibias (changes requested):
## Maintainer re-review (post `938f25e`)

**Verdict: REQUEST CHANGES**

HEAD: `938f25eb`. Required items from the prior review are largely in. No new Codex review on this tip; CodeRabbit auto-paused (no new actionable comments). macos/windows CI green; **ubuntu-latest cancelled after hitting the 12m job timeout during Test** (typecheck had passed) — needs a green re-run before merge.

### Prior required vs `938f25e`

| Item | Status |
| --- | --- |
| #1 Deadline abort classification (`TimeoutError` / `deadlineSignal` outside try) | **Fixed** — poll + submit catch verified; Bun propagates `TimeoutError` through `AbortSignal.any` |
| #2 Pruned-path guard | **Partial** — see below |
| #3 Wire video aliases + unnamespaced strip | **Partial** — see below |
| #4 `authMode: "key"` in tests | **Fixed** |
| DNS noun `"video"` | **Fixed** |

### Required before merge

**1. Finish the pruned-path guard** (`loop.ts` ~705–710)

Rewriting `ok: false` still spreads the old result, so **`markdown` keeps the dead `file:` URI** in the tool_result JSON. Also only `path` is checked — multi-image results can keep deleted entries in `files[]` after the post-batch prune.

Do the `retainAfterBatch` equivalent: filter `files` with `existsSync`; if none remain → `ok: false` with the prune error and **no** `markdown`/`path`; if some remain → refresh `path`/`files`/`count`/`markdown` from survivors only.

**2. Don’t seed namespaced tools into `videoPlan.toolNames`** (`plan.ts` alias loop)

`core.ts` correctly strips only `!t.namespace`, but `planVideoBridge` still adds every `isVideoGenName(t.name)` match, including MCP tools with a namespace. Those leaf names then make `videoPlan.toolNames.has(call.name)` intercept the MCP call in the loop. Skip namespaced tools when seeding (`if (t.namespace) continue`), matching the filter rule you already landed.

**3. Green ubuntu CI**

`ubuntu-latest` on `938f25e` hit the 12m execution cap mid-Test (cancelled, not a clean fail). Re-run Cross-platform CI. If it keeps timing out, mock `sleep` / poll interval in `tests/videos/fulfill-video.test.ts` (the real 5s×2 heartbeat wait) so the suite stays under the cap — that was previously “CI cost only”; it’s now merge-relevant.

### Not blocking

| Finding | Verdict |
| --- | --- |
| Web search wins over video | Accepted / by design |
| CodeRabbit paused | No new findings to triage |

tizerluo:
All three items addressed in `cf76fb1e`:

**1. Finish pruned-path guard** ✅
- `files[]` is now filtered through `existsSync` (was: only checked `path`)
- If all files pruned → `{ ok: false, files: [], count: 0, error: "..." }` with **no** `markdown` or `path`
- If some survive → `path`/`files`/`count`/`markdown` refreshed from survivors only
- Works for both single-video and multi-image results

**2. Skip namespaced tools** ✅
- `planVideoBridge` alias loop: `if (t.namespace) continue` before checking `isVideoGenName`
- Namespaced MCP `video_gen` tools are no longer seeded into `videoPlan.toolNames`, matching the `core.ts` filter rule

**3. Ubuntu CI timeout** ✅
- Heartbeat test mocks `globalThis.setTimeout` to resolve instantly
- Full suite: **156ms** (was ~5.2s)
- No real `sleep(5000)` wait in any test path

36 tests pass, typecheck clean. Ready for CI re-run.

Wibias (dismissed):
## Maintainer re-review (post `cf76fb1e` + `50bd696`)

**Verdict: REQUEST CHANGES** (merge hold = green ubuntu only)

### Prior required vs author `cf76fb1e`

| Item | Status |
| --- | --- |
| #1 Finish pruned-path guard (`files[]` + no dead markdown/path) | **Fixed** |
| #2 Skip namespaced tools when seeding `videoPlan.toolNames` | **Fixed** |
| #3 Green ubuntu CI | **Not fixed by cf76fb1e** — see below |

### CI hang root cause (verified)

`938f25e` had **green** Cross-platform CI. `cf76fb1e` re-cancelled ubuntu-latest at the 12m Test cap. The new “fix” mocked `globalThis.setTimeout` to no-op the 5s poll sleep. That races parallel Bun workers and hung the suite (~11 min silence after CLI status tests).

**Pushed to this PR as `50bd696`** (maintainer can-modify):
- Injectable `sleepFn` / `pollFn` seams on `pollVideoWithHeartbeats`
- Removed global `setTimeout` monkeypatch
- Dropped `mock.module` from fulfill-video tests (also stopped cross-file pollution when `tests/videos/*` run together)

### Bot triage (this tip)

| Finding | Verdict |
| --- | --- |
| Codex: cancel oversized bodies | Already `reader.cancel()` in `readBoundedText` finally — skip |
| Codex: docs JSON brace / provider key | Fixed in docs |
| Codex: restore fetch | Fixed (`afterEach`) |
| Rabbit Critical: OAuth never activates | By design / docs — skip |
| Rabbit Major: `videoTimeoutMs` clamp disagreement | Invalid — plan clamps; core passes `vidPlan.timeoutMs` |
| Rabbit: authMode in tests / JSON reparse / web-search docs wording | Fixed or trivial — skip |
| Codex prune-after-batch | Fixed earlier |

### Required before merge

1. **Green `ubuntu-latest`** on `50bd696` (or tip). macos/windows were already green on the prior tip.
2. No further product changes needed from the prior review list.

Not merging from here.

Wibias:
Thanks for the Pr. Will get merged after interal code work lands on dev.

Wibias:
## Transition workflow completed

1. Merged into `dev`: `257c626c` (#582)
2. Carried onto `dev2-go`: #660 (TS oracle + docs + tests)
3. **Go port:** needed — image bridge exists under `go/internal/images/`, but no video counterpart yet. Tracking: #661 (`needs-go-port`)

Issue #499, opened 2026-07-26
[Bug][Windows] Stale responses-state atomic-write temp files accumulate indefinitely


0disoft:
## Client or integration

OpenCodex proxy state persistence (observed while using Codex App)

## Area

Proxy and routing

## Summary

On OpenCodex 2.7.40 on Windows, stale atomic-write temporary files for `responses-state.json` accumulated indefinitely under the OpenCodex config directory:

- pattern: `responses-state.json.ocx.<pid>.<sequence>.tmp`
- count: 263 files
- total size: 5.88 GB
- creation period: July 20–25, 2026
- all PIDs encoded in the stale filenames were no longer running

I expected interrupted/failed response-state writes to be cleaned up, either during the failed write or on a later proxy startup. Instead, crash/forced-termination remnants survived multiple restarts and consumed several gigabytes.

This report is about unbounded stale state files and disk/privacy hygiene. I am **not** claiming that their mere presence was the primary cause of dashboard latency; a separate issue reports the independently measured `usage.jsonl` performance problem.

## Reproduction

The exact interruption timing has not yet been reduced to a deterministic single command. The production observation followed several proxy crashes or forced terminations while `responses-state.json` was large and actively persisted.

A likely reproduction path is:

1. Use Responses/Codex conversations long enough for `responses-state.json` to grow substantially.
2. Terminate the proxy while the debounced response-state snapshot is being persisted.
3. Restart and repeat.
4. Inspect the OpenCodex config directory for stale files:

```powershell
$home = Join-Path $HOME '.opencodex'
$tmp = Get-ChildItem -LiteralPath $home -File -Filter 'responses-state.json.ocx.*.tmp'
$tmp | Measure-Object Length -Sum
$tmp | Select-Object Name, Length, LastWriteTime
```

Observed filename PIDs were:

```text
3200, 6400, 6664, 10168, 15772, 17400, 32660
```

None were alive when checked. The newly restarted proxy PID did not occur in any stale filename.

Cleanup observation, without asserting a root cause: PowerShell `Remove-Item`, `Copy-Item`, and `Rename-Item` returned access denied for sampled stale files, while `[System.IO.File]::Delete()` succeeded. File owner/attributes alone did not explain the difference, so this should be treated only as an observed Windows cleanup behavior.

## Implementation notes

In 2.7.40:

- `src/responses/state.ts:136-164` serializes the response-state snapshot and calls `atomicWriteFile(...)` from a best-effort `try/catch`.
[...]

PR #601, opened 2026-07-28
fix(windows): grant owner ACE before ACL inheritance removal

### LINKED ISSUE #596: [Bug]: Windows ACL hardening can strand zero-ACE atomic temp files
## Client or integration

Codex App

## Area

Service lifecycle

## Summary

On Windows, OpenCodex 2.7.42 can leave protected atomic-write temp files with no usable access-control entry (ACE) for the current user. The affected files can contain authentication, configuration, or response-state data and then cannot be reliably read, truncated, renamed, or removed by the process that created them.

The expected behavior is that the current user retains explicit Full Control throughout ACL hardening, including when a later `icacls` step times out or fails.

## Reproduction

The production occurrence is timing-dependent, but the failure is deterministic with an injected command runner:

1. Begin with a Windows temp file whose current-user access is inherited.
2. Run the current `runIcacls()` sequence: `/inheritance:r`, broad-group removal, then `/grant:r CURRENTUSER:(F)`.
3. Inject a timeout or failure in the broad-group removal step after inheritance has already been removed.
4. Inspect the temp file ACL and attempt to read, rename, or unlink it.
5. Observe that the file is owned by the current user but has a protected DACL with zero ACEs, so ordinary access and cleanup fail until a current-user Full Control ACE is restored.

This corresponds to the stale files observed in production. A regression test can model the same sequence without depending on timing.

## Version

2.7.42

## Operating system

Microsoft Windows 11 Pro 25H2, x64 (build 26200.8894)

## Provider and model

Not provider-specific.

## Logs or error output

No application stack trace is required to identify the failure. The decisive evidence is the ACL state: the stale temp files are owned by the current user, inheritance is protected, and the DACL contains zero ACEs. Ordinary read, copy, rename, and removal operations fail until an explicit current-user Full Control ACE is restored.

No secret-bearing files are attached. The deterministic regression and aggregate production evidence are described below.

## Redacted configuration

Not configuration-dependent.

## Checks

- [x] I searched existing issues and documentation, including #160, #499, PR #96, PR #102, and PR #502.
- [x] I removed secrets, tokens, account details, request credentials, personal paths, usernames, and response contents.

## Production evidence

On Windows 11 with OpenCodex 2.7.42:

- 65 stale atomic temp files were present under the OpenCodex home.
- Affected basenames included `auth.json`, `config.json`, and `responses-state.json`.
- Total size was about 1.09 GB.
- Every PID encoded in those filenames was no longer running.
- The files were owned by the current user but had a protected ACL with zero ACE entries.
- Ordinary read, copy, rename, and removal attempts failed until an explicit current-user Full Control ACE was restored.
- The final destination files had the expected single current-user Full Control ACE.

No usernames, home paths, tokens, credentials, or response contents are included here.

## Root cause

`/inheritance:r` is a destructive transition: it removes inherited ACEs. The current implementation performs that transition before making owner access explicit.

The timeout/retry work tracked in #160 correctly prevents partial hardening from being reported as success, but it does not preserve cleanup access while the sequence is incomplete. The stale-temp recovery added for #499 is complementary; it cannot reliably clean a temp whose DACL no longer permits the owner to open or unlink it.

## Proposed invariant and fix

Install the explicit owner ACE before any destructive ACL change:

1. `/grant:r CURRENTUSER:(F)` (or `(OI)(CI)(F)` for directories)
2. `/inheritance:r`
3. Remove broad group SIDs, retaining the existing `/findsid` verification

This preserves the following invariant:

- If the grant fails, inheritance is untouched.
- If any later step times out or fails, the writer still has explicit Full Control and can scrub or unlink the temp.
- Removing broad group SIDs cannot remove the explicit current-user ACE.
- No duplicate final grant is required.

## Deterministic regression

An injected-runner test models inherited owner access, removes inheritance, then forces `/remove:g` to time out. The expected operation order is:

~~~text
grant-owner
remove-inheritance
remove-broad
~~~

Effective owner access must remain true after the timeout.

A second regression asserts the exact successful order. The existing EPERM case is also strengthened so the sole attempted operation is `/grant:r`, proving a grant failure cannot modify inheritance first.

## Validation

The focused patch was validated with:

- 28 Windows ACL tests, including the new failure-order and success-order regressions.
- 113 related tests across ACL, atomic-write startup, service, scheduler, and lifecycle suites.
- TypeScript type-check.
- Privacy scan.
- A real Windows probe confirming the hardened file remained readable and showed only the current-user Full Control ACE.

The old implementation fails the new ordering regression; the owner-first implementation passes it.

Wibias:
## Summary
- Fix [#596](https://github.com/lidge-jun/opencodex/issues/596): Windows ACL hardening could strand atomic-write temps (`auth.json` / `config.json` / `responses-state.json` `.tmp`) with a protected **zero-ACE** DACL after `/inheritance:r` ran and a later `icacls` step timed out.
- Reorder `runIcacls()` to **grant current-user Full Control first**, then `/inheritance:r`, then broad-SID `/remove:g` (with existing `/findsid` verification).
- Invariant: if grant fails, inheritance is untouched; if a later step fails/times out, the writer still has explicit Full Control for scrub/unlink.

## Why
`/inheritance:r` is destructive (drops inherited ACEs immediately). Microsoft/docs + field reports agree an empty DACL denies ordinary access; owner implicit rights are not enough for reliable cleanup. Related: #160 (timeout honesty), #499 / #502 (stale-temp recovery — complementary, cannot open a zero-ACE temp).

## Test plan
- [x] `bun test tests/windows-secret-acl.test.ts` — 28 pass (success-order + remove:g-timeout owner-ACE + grant-first EPERM)
- [x] `bun run typecheck`
- [x] `bun run privacy:scan`
- [ ] CI green (win/mac/ubuntu)
- [ ] CodeRabbit + Codex review threads resolved

Closes #596

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows secret-file permission hardening by granting the current user full control before changing inheritance or removing broad permissions.
  * Prevented subsequent permission changes when the initial access grant fails.
  * Added clearer timeout diagnostics while preserving the owner’s access if later cleanup steps fail.

### DIFF
--- a/src/lib/windows-secret-acl.ts
+++ b/src/lib/windows-secret-acl.ts
@@ -3,9 +3,14 @@
  *
  * On Windows, `chmod` only controls POSIX-style bits in the ACE list and does NOT remove
  * inherited permissions from other users. Real per-user isolation requires icacls to:
- *   1. Disable inheritance   (icacls path /inheritance:r)
- *   2. Strip broad explicit grants by SID (Everyone, Users, Authenticated Users)
- *   3. Grant the current user full control (icacls path /grant:r "CURRENTUSER:(F)")
+ *   1. Grant the current user full control (icacls path /grant:r "CURRENTUSER:(F)")
+ *   2. Disable inheritance               (icacls path /inheritance:r)
+ *   3. Strip broad explicit grants by SID (Everyone, Users, Authenticated Users)
+ *
+ * Owner grant MUST precede `/inheritance:r`. That flag is destructive: it drops inherited
+ * ACEs immediately. If a later step times out after inheritance is removed but before an
+ * explicit owner ACE exists, the temp is left with a protected empty DACL — owned by the
+ * current user yet unreadable/ununlinkable until Full Control is restored (issue #596).
  *
  * On non-Windows platforms the helpers fall through to the caller's existing chmod-based
  * behaviour: they return ok:true without invoking any external process.
@@ -141,8 +146,8 @@ function currentWindowsUser(): string | undefined {
 
 /**
  * Run icacls to harden a single file system entry.
- * - Disables inheritance (keeps nothing: /inheritance:r)
- * - Grants the current user Full Control
+ * Order is intentional (issue #596): grant the owner ACE first so a later
+ * `/inheritance:r` or `/remove:g` timeout cannot strand a protected zero-ACE DACL.
  *
  * We do NOT use a shell string; all arguments are passed as an array so no
  * shell injection is possible even for paths with unusual characters.
@@ -170,13 +175,20 @@ function runIcacls(targetPath: string, directory: boolean, deadline: number): vo
     if (!result.success) throw icaclsError(step, result);
   };
 
-  // Step 1: disable inheritance and remove inherited ACEs
+  // Step 1: grant current user full control BEFORE any destructive ACL change.
+  // If this fails, inheritance is untouched and the writer keeps inherited access.
+  const grant = directory ? `${user}:(OI)(CI)(F)` : `${user}:(F)`;
+  runOrThrow("/grant:r", [targetPath, "/grant:r", grant]);
+
+  // Step 2: disable inheritance and remove inherited ACEs. The explicit owner ACE
+  // from step 1 survives this transition, so a later failure still leaves cleanup access.
   runOrThrow("/inheritance:r", [targetPath, "/inheritance:r"]);
 
-  // Step 2: remove broad explicit grants using stable SIDs (not localized names).
+  // Step 3: remove broad explicit grants using stable SIDs (not localized names).
   // Missing ACEs can yield a non-zero exit; verify with locale-independent /findsid
   // before accepting the failure as harmless — a swallowed real failure would leave
   // Everyone/Users/Authenticated Users grants while reporting hardened.
+  // `/remove:g` cannot remove the explicit current-user ACE installed in step 1.
   const removal = run("/remove:g", [targetPath, "/remove:g", ...BROAD_SIDS]);
   if (!removal.success) {
     if (removal.timedOut) throw icaclsError("/remove:g", removal);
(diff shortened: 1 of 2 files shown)


Wibias:
@codex review

Wibias:
## Review verdict: **merge-ready**

### Context
Implements the fix proposed in [#596](https://github.com/lidge-jun/opencodex/issues/596): grant current-user Full Control **before** `/inheritance:r` so a mid-sequence `icacls` timeout cannot leave a protected **zero-ACE** DACL on atomic-write temps. Matches Windows/`icacls` guidance that `/inheritance:r` is destructive and an empty DACL denies ordinary access. Complements #160 (timeout honesty) and #499/#502 (stale-temp recovery cannot open a zero-ACE file).

Issue thread had no separate maintainer comment beyond the detailed reporter proposal (which this PR follows).

### Bots
| Bot | Result |
|---|---|
| CodeRabbit | Review completed — **no actionable comments** |
| Codex | `@codex review` → **no major issues** (`26d4c14f`) |
| Open review threads | **0** |

### CI
All green: ubuntu / macos / windows, npm-global ×3, react-doctor, enforce-target, label.

### Bug review
**No shippable defects.** Order is grant → inheritance → remove-broad; grant failure leaves inheritance untouched; `/remove:g` only targets Everyone/Users/Authenticated Users SIDs and cannot strip the owner ACE. Regressions cover success order, remove:g-timeout with owner ACE retained, and grant-first EPERM.

### Security review
**No medium+ findings.** End-state isolation unchanged (same broad SIDs + `/findsid` verify). Argv-only `icacls` spawn (no shell). EPERM still fail-closed; timeout soft-fail policy unchanged but partial failure is now safer (explicit owner ACE vs stranded empty DACL). Brief post-grant / pre-inheritance window is same-process and not worse than unhardened baseline.

### Local
- `bun test tests/windows-secret-acl.test.ts` — 28 pass
- `bun run typecheck` / `privacy:scan` — clean

**Holding merge per request.** Ready when you want it squashed.

PR #610, opened 2026-07-28
fix(catalog): stop respawning the codex --version probe on every catalog read

### LINKED ISSUE #606: ocx claude: 3s budget exceeded by uncached codex --version probe on /api/claude-code and /v1/models
### Client or integration

Claude Code

### Area

Proxy and routing

### Summary

On Windows, `ocx claude` prints both of these on almost every launch:

```
⚠ 모델 컨텍스트 정보를 불러오지 못했습니다 — 1M 자동 표시는 이번 실행에서 생략됩니다.
⚠ Gateway model cache could not be refreshed; the model picker may be stale.
```

Neither is an auth or connectivity failure. Both are **timeouts**. `cmdClaude` gives
`fetchClaudeContextWindows` (`/api/claude-code`) and `refreshGatewayModelCacheFromProxy`
(`/v1/models`) a 3s budget each, and the local proxy takes 5-8s to answer. Both `catch`
blocks swallow the `AbortError` and emit the generic warning above.

The user-visible consequences are quiet but real: `[1m]` context marking is skipped for
the run, and `~/.claude/cache/gateway-models.json` is never rewritten. Because Claude Code
only refreshes that cache itself when it holds a credential — and the subscription-preserving
launch deliberately injects none — the picker keeps serving whatever is on disk. Mine was
6 days stale (80 entries) before I traced this.

Expected: both endpoints answer within the 3s budget, so the cache is rewritten and `[1m]`
marking applies. Actual: both abort, every launch, and the warnings suggest an auth problem
rather than a latency one.

Root cause is a `codex --version` subprocess probe sitting on the request path, which is
effectively uncached for two independent reasons (details below).

### Reproduction

1. Install `@bitkyc08/opencodex` 2.7.42 on Windows with a `codex` CLI present, plus several
   additional `codex` launchers reachable via `PATH` (mine: npm global shim, a Codex
   App-managed binary, and a shim backup — 102 probe candidates in total).
2. Start the proxy: `ocx start --port 10100`.
3. Run `ocx claude --version`.
4. Observe both warnings on stderr.
5. Time the two endpoints `cmdClaude` calls directly:

   ```powershell
   Measure-Command { Invoke-WebRequest 'http://127.0.0.1:10100/api/claude-code' -UseBasicParsing }
   Measure-Command { Invoke-WebRequest 'http://127.0.0.1:10100/v1/models?limit=1000&ids=cli' -Headers @{'anthropic-version'='2023-06-01'} -UseBasicParsing }
   ```

   Both exceed 3000 ms on a cold proxy. Note that repeating step 5 after a >60s idle gap
   reproduces it again, because `BUNDLED_CATALOG_CACHE_MS` is 60s.
6. Confirm the effect: `~/.claude/cache/gateway-models.json` keeps its old `fetchedAt`.

Timed in-process against the same config, to isolate where the time goes:

```
fetchAllModels              457 ms   (all 7 providers, network included)
listCatalogNativeSlugs     1328 ms   <-- no network at all
visibleNativeSlugs         1645 ms
nativeOpenAiSlugs          2072 ms
buildClaudeContextWindows  2018 ms
```

All of the native-slug cost is `readCurrentCatalogOrCache()` → `loadBundledCodexCatalog()`.
The subprocess it wraps is not the problem:

```
codex debug models --bundled    87 ms
codex --version                721 ms
```

`loadBundledCodexCatalog()` three times in a row, nothing changing on disk, 60s cache:

```
#0  1214 ms
#1  1136 ms   <-- expected ~0 on a warm cache
#2  1186 ms
```

#### Cause 1 — the bundled-catalog cache can never be hit

In `src/codex/catalog/bundled.ts`, `candidates` is computed in an IIFE that runs *before*
the `bundledCatalogCache` lookup, because `cacheKey` is assigned inside it. So
`resolveAndPersistCodexRuntime()` runs on every call, warm hits included. That would still
be cheap if the resolve were memoized, but line 154 forwards the real `execFileSync`:

```ts
const execFile = deps.execFileSync ?? (execFileSync as unknown as ExecFile);
// ...
const resolved = resolveAndPersistCodexRuntime({
  execFileSync: execFile,   // always truthy
```

and `resolveCacheKey()` in `src/codex/runtime.ts` deliberately refuses to memoize
injected-dependency resolves:

```ts
if (deps.execFileSync || deps.existsSync || deps.readFileSync || deps.configDir || deps.now) {
  return null;
}
```

The guard reads correct in intent (injected deps imply a test double), but this caller
defeats it for the real-runtime path by passing its own already-defaulted value.

There is a related self-invalidation: `resolveAndPersistCodexRuntime()` (runtime.ts:505)
calls `persistCodexRuntime` unconditionally for any non-fallback selection, and
`persistCodexRuntime()` ends with `resolveCache = null` (runtime.ts:228). Since
`persistedRuntimeCacheStamp()` also folds the on-disk `updatedAt` into the cache key, each
write both clears the memo and changes the key it would have been stored under — even when
the resolved runtime is byte-identical to what is already persisted.

#### Cause 2 — cold reads probe every `codex` on PATH

`bundled.ts:161` passes `discoverAlternatives: deps.discoverAlternatives`, i.e. `undefined`
for normal callers, which `resolveCodexRuntimeUncached()` treats as "discover". It then
probes every candidate to populate `newerAvailable`, each probe spawning `codex --version`
with a fresh `mkdtempSync` sandbox:

```
resolveCodexRuntime({})                            1224 ms   failures: 102
resolveCodexRuntime({ discoverAlternatives: false })  60 ms
```

Both select the identical runtime. Catalog loading only consumes `resolved.runtime.command`;
it never reads `newerAvailable` or `failures`. So on a cold read that ~1.2s scan is pure
overhead, and on its own eats more than a third of the 3s budget before any real work.

#### Effect

`/api/claude-code` measured over HTTP:

| | before | after local patch |
|---|---|---|
| cold (fresh proxy) | 5000-8300 ms | 165 ms |
| warm | 6300 ms | ~25 ms |
| after 65s idle (cache expired) | 6325 ms | 1440 ms |

With the patches `ocx claude` runs clean, and `gateway-models.json` refreshed from 80 stale
entries to 98 current ones, 25 carrying `[1m]`.

#### Patches I verified locally

Not opening a PR since you may prefer different tradeoffs, but these are what I tested.

`src/codex/catalog/bundled.ts` — forward an injected `execFileSync` only, so real-runtime
resolves stay memoizable:

```ts
...(deps.execFileSync ? { execFileSync: deps.execFileSync } : {}),
```

`src/codex/catalog/bundled.ts` — default the catalog path to non-discovering:

```ts
discoverAlternatives: deps.discoverAlternatives ?? false,
```

`src/codex/runtime.ts` — only persist when the selection actually changed, so the write
stops clearing its own memo:

```ts
const persistedRuntime = loadPersistedCodexRuntime(deps);
const selectionUnchanged = persistedRuntime !== null
  && persistedRuntime.command === result.runtime.command
  && persistedRuntime.source === result.runtime.source
  && (persistedRuntime.selectedVersion ?? null) === result.runtime.version;
if (result.runtime.command && result.runtime.source !== "fallback" && !selectionUnchanged) {
```

Tradeoff worth flagging on the second one: catalog-path callers lose `newerAvailable`
diagnostics. If something depends on those, moving discovery to a slow background interval
instead of every catalog read would preserve them while keeping the request path fast.

#### Suggested improvements beyond the fix

1. **Check `bundledCatalogCache` before resolving the runtime.** Deriving the cache key from
   the very thing the cache exists to avoid means a warm hit can only ever save the
   `codex debug models` call (87 ms), never the probe (1.2 s) that dominates. A cheaper key
   would let warm hits skip both.
2. **Attribute the timeout in the warning text.** Both messages read like auth or network
   faults. Something like `timed out after 3000ms` would point straight at the cause; I went
   looking at OAuth state and provider connectivity first.
3. **Consider raising or making the 3s budget configurable.** Even fully patched, a cold read
   is ~1.4s here, so a slower machine or a longer `PATH` could still trip it silently.
4. **A regression test that a repeated `loadBundledCodexCatalog()` spawns no subprocess**
   would have caught both causes.

### Version

2.7.42

### Operating system

Windows 11

### Provider and model

Not provider-specific. Reproduced with `openai` (forward), `kiro`, `kiro-native`,
`opencode-go`, `xai`, and `google-antigravity` configured; the slow path is local catalog
loading, not any upstream call.

### Logs or error output

```shell
$ ocx claude --version
⚠ 모델 컨텍스트 정보를 불러오지 못했습니다 — 1M 자동 표시는 이번 실행에서 생략됩니다.
⚠ Gateway model cache could not be refreshed; the model picker may be stale.
2.1.220 (Claude Code)

# proxy endpoint timings, cold proxy (3s budget in cmdClaude)
claude-code #1 : 4771 ms
claude-code #2 : 23 ms
claude-code #3 : 18 ms

# and again after a 65s idle gap (BUNDLED_CATALOG_CACHE_MS = 60_000)
claude-code after 65s idle: 6325 ms
immediately after: 69 ms

# runtime resolution, same machine
resolveCodexRuntime({})                            1224 ms   failures: 102
resolveCodexRuntime({ discoverAlternatives: false })  60 ms
selected(true) : ...\Codex\bin\<hash>\codex.exe environment 0.146.0-alpha.3.1
selected(false): ...\Codex\bin\<hash>\codex.exe environment 0.146.0-alpha.3.1
newerAvailable : ...\npm\codex.cmd
```

### Redacted configuration

```json
{
  "port": 10100,
  "defaultProvider": "openai",
  "codexAutoStart": true,
  "providers": {
    "openai":  { "adapter": "openai-responses", "authMode": "forward", "codexAccountMode": "pool" },
    "kiro":    { "adapter": "kiro", "authMode": "oauth", "liveModels": false },
    "kiro-native": { "adapter": "anthropic", "baseUrl": "http://127.0.0.1:3456", "allowPrivateNetwork": true, "liveModels": false },
    "opencode-go": { "adapter": "openai-chat", "authMode": "key", "apiKey": "<redacted>" },
    "xai":     { "adapter": "openai-chat", "authMode": "oauth", "liveModels": true },
    "google-antigravity": { "adapter": "google", "googleMode": "cloud-code-assist", "authMode": "oauth" }
  }
}
```

Relevant environment: `CODEX_CLI_PATH` is set, so the runtime resolves with
`source: "environment"`; `codex-cli` is `0.146.0-alpha.14`; the proxy runs under the bundled
Bun 1.3.14. No combos configured, `disabledModels` empty.

### Checks

---

Unrelated aside, happy to split out if useful: model discovery for the `adapter: "google"` /
`googleMode: "cloud-code-assist"` provider fails with HTTP 404 and degrades to the static
catalog (`Provider model discovery for "google-antigravity" failed with HTTP 404
[urlClass=provider-models, fallback=configured]`).

<details>

<summary>Translated Message</summary>

*Original language: Korean*

### Client or integration

Claude Code

### Area

Proxy and routing

### Summary

On Windows, `ocx claude` prints both of these on almost every launch:

```
⚠ Unable to load model context information — 1M auto marking will be skipped for this run.
⚠ Gateway model cache could not be refreshed; the model picker may be stale.
```

Neither is an auth or connectivity failure. Both are **timeouts**. `cmdClaude` gives
`fetchClaudeContextWindows` (`/api/claude-code`) and `refreshGatewayModelCacheFromProxy`
(`/v1/models`) a 3s budget each, and the local proxy takes 5-8s to answer. Both `catch`
blocks swallow the `AbortError` and emit the generic warning above.

The user-visible consequences are quiet but real: `[1m]` context marking is skipped for
the run, and `~/.claude/cache/gateway-models.json` is never rewritten. Because Claude Code
only refreshes that cache itself when it holds a credential — and the subscription-preserving
launch deliberately injects none — the picker keeps serving whatever is on disk. Mine was
6 days stale (80 entries) before I traced this.

Expected: both endpoints answer within the 3s budget, so the cache is rewritten and `[1m]`
marking applies. Actual: both abort, every launch, and the warnings suggest an auth problem
rather than a latency one.

Root cause is a `codex --version` subprocess probe sitting on the request path, which is
effectively uncached for two independent reasons (details below).

### Reproduction

1. Install `@bitkyc08/opencodex` 2.7.42 on Windows with a `codex` CLI present, plus several
   additional `codex` launchers reachable via `PATH` (mine: npm global shim, a Codex
   App-managed binary, and a shim backup — 102 probe candidates in total).
2. Start the proxy: `ocx start --port 10100`.
3. Run `ocx claude --version`.
4. Observe both warnings on stderr.
5. Time the two endpoints `cmdClaude` calls directly:

   ```powershell
   Measure-Command { Invoke-WebRequest 'http://127.0.0.1:10100/api/claude-code' -UseBasicParsing }
   Measure-Command { Invoke-WebRequest 'http://127.0.0.1:10100/v1/models?limit=1000&ids=cli' -Headers @{'anthropic-version'='2023-06-01'} -UseBasicParsing }
   ```

   Both exceed 3000 ms on a cold proxy. Note that repeating step 5 after a >60s idle gap
   reproduces it again, because `BUNDLED_CATALOG_CACHE_MS` is 60s.
6. Confirm the effect: `~/.claude/cache/gateway-models.json` keeps its old `fetchedAt`.

Timed in-process against the same config, to isolate where the time goes:

```
fetchAllModels              457 ms   (all 7 providers, network included)
listCatalogNativeSlugs     1328 ms   <-- no network at all
visibleNativeSlugs         1645 ms
nativeOpenAiSlugs          2072 ms
buildClaudeContextWindows  2018 ms
```

All of the native-slug cost is `readCurrentCatalogOrCache()` → `loadBundledCodexCatalog()`.  
The subprocess it wraps is not the problem:

```
codex debug models --bundled    87 ms
codex --version                721 ms
```

`loadBundledCodexCatalog()` three times in a row, nothing changing on disk, 60s cache:

```
#0  1214 ms
#1  1136 ms   <-- expected ~0 on a warm cache
#2  1186 ms
```

#### Cause 1 — the bundled-catalog cache can never be hit

In `src/codex/catalog/bundled.ts`, `candidates` is computed in an IIFE that runs *before*
the `bundledCatalogCache` lookup, because `cacheKey` is assigned inside it. So
`resolveAndPersistCodexRuntime()` runs on every call, warm hits included. That would still
be cheap if the resolve were memoized, but line 154 forwards the real `execFileSync`:

```ts
const execFile = deps.execFileSync ?? (execFileSync as unknown as ExecFile);
// ...
const resolved = resolveAndPersistCodexRuntime({
  execFileSync: execFile,   // always truthy
```

and `resolveCacheKey()` in `src/codex/runtime.ts` deliberately refuses to memoize
injected-dependency resolves:

```ts
if (deps.execFileSync || deps.existsSync || deps.readFileSync || deps.configDir || deps.now) {
  return null;
}
```

The guard reads correct in intent (injected deps imply a test double), but this caller
defeats it for the real-runtime path by passing its own already-defaulted value.

There is a related self-invalidation: `resolveAndPersistCodexRuntime()` (runtime.ts:505)
calls `persistCodexRuntime` unconditionally for any non-fallback selection, and
`persistCodexRuntime()` ends with `resolveCache = null` (runtime.ts:228). Since
`persistedRuntimeCacheStamp()` also folds the on-disk `updatedAt` into the cache key, each
write both clears the memo and changes the key it would have been stored under — even when
the resolved runtime is byte-identical to what is already persisted.

#### Cause 2 — cold reads probe every `codex` on PATH

`bundled.ts:161` passes `discoverAlternatives: deps.discoverAlternatives`, i.e. `undefined`
for normal callers, which `resolveCodexRuntimeUncached()` treats as "discover". It then
probes every candidate to populate `newerAvailable`, each probe spawning `codex --version`
with a fresh `mkdtempSync` sandbox:

```
resolveCodexRuntime({})                            1224 ms   failures: 102
resolveCodexRuntime({ discoverAlternatives: false })  60 ms
```

Both select the identical runtime. Catalog loading only consumes `resolved.runtime.command`;
it never reads `newerAvailable` or `failures`. So on a cold read that ~1.2s scan is pure
overhead, and on its own eats more than a third of the 3s budget before any real work.

#### Effect

`/api/claude-code` measured over HTTP:

| | before | after local patch |
|---|---|---|
| cold (fresh proxy) | 5000-8300 ms | 165 ms |
| warm | 6300 ms | ~25 ms |
| after 65s idle (cache expired) | 6325 ms | 1440 ms |

With the patches `ocx claude` runs clean, and `gateway-models.json` refreshed from 80 stale
entries to 98 current ones, 25 carrying `[1m]`.

#### Patches I verified locally

Not opening a PR since you may prefer different tradeoffs, but these are what I tested.

`src/codex/catalog/bundled.ts` — forward an injected `execFileSync` only, so real-runtime
resolves stay memoizable:

```ts
...(deps.execFileSync ? { execFileSync: deps.execFileSync } : {}),
```

`src/codex/catalog/bundled.ts` — default the catalog path to non-discovering:

```ts
discoverAlternatives: deps.discoverAlternatives ?? false,
```

`src/codex/runtime.ts` — only persist when the selection actually changed, so the write
stops clearing its own memo:

```ts
const persistedRuntime = loadPersistedCodexRuntime(deps);
const selectionUnchanged = persistedRuntime !== null
  && persistedRuntime.command === result.runtime.command
  && persistedRuntime.source === result.runtime.source
  && (persistedRuntime.selectedVersion ?? null) === result.runtime.version;
if (result.runtime.command && result.runtime.source !== "fallback" && !selectionUnchanged) {
```

Tradeoff worth flagging on the second one: catalog-path callers lose `newerAvailable`
diagnostics. If something depends on those, moving discovery to a slow background interval
instead of every catalog read would preserve them while keeping the request path fast.

#### Suggested improvements beyond the fix

1. **Check `bundledCatalogCache` before resolving the runtime.** Deriving the cache key from
the very thing the cache exists to avoid means a warm hit can only ever save the
`codex debug models` call (87 ms), never the probe (1.2 s) that dominates. A cheaper key
would let warm hits skip both.
2. **Attribute the timeout in the warning text.** Both messages read like auth or network
faults. Something like `timed out after 3000ms` would point straight at the cause; I went
looking at OAuth state and provider connectivity first.
3. **Consider raising or making the 3s budget configurable.** Even fully patched, a cold read
is ~1.4s here, so a slower machine or a longer `PATH` could still trip it silently.
4. **A regression test that a repeated `loadBundledCodexCatalog()` spawns no subprocess**
would have caught both causes.

### Version

2.7.42

### Operating system

Windows 11

### Provider and model

Not provider-specific. Reproduced with `openai` (forward), `kiro`, `kiro-native`,
`opencode-go`, `xai`, and `google-antigravity` configured; the slow path is local catalog
loading, not any upstream call.

### Logs or error output

```shell
$ ocx claude --version
⚠ Unable to load model context information — 1M auto marking will be skipped for this run.
⚠ Gateway model cache could not be refreshed; the model picker may be stale.
2.1.220 (Claude Code)

# proxy endpoint timings, cold proxy (3s budget in cmdClaude)
claude-code #1 : 4771 ms
claude-code #2 : 23 ms
claude-code #3 : 18 ms

# and again after a 65s idle gap (BUNDLED_CATALOG_CACHE_MS = 60_000)
claude-code after 65s idle: 6325 ms
immediately after: 69 ms

# runtime resolution, same machine
resolveCodexRuntime({})                            1224 ms   failures: 102
resolveCodexRuntime({ discoverAlternatives: false })  60 ms
selected(true) : ...\Codex\bin\<hash>\codex.exe environment 0.146.0-alpha.3.1
selected(false): ...\Codex\bin\<hash>\codex.exe environment 0.146.0-alpha.3.1
newerAvailable : ...\npm\codex.cmd
```

### Redacted configuration

```json
{
  "port": 10100,
  "defaultProvider": "openai",
  "codexAutoStart": true,
  "providers": {
    "openai":  { "adapter": "openai-responses", "authMode": "forward", "codexAccountMode": "pool" },
    "kiro":    { "adapter": "kiro", "authMode": "oauth", "liveModels": false },
    "kiro-native": { "adapter": "anthropic", "baseUrl": "http://127.0.0.1:3456", "allowPrivateNetwork": true, "liveModels": false },
    "opencode-go": { "adapter": "openai-chat", "authMode": "key", "apiKey": "<redacted>" },
    "xai":     { "adapter": "openai-chat", "authMode": "oauth", "liveModels": true },
    "google-antigravity": { "adapter": "google", "googleMode": "cloud-code-assist", "authMode": "oauth" }
  }
}
```

Relevant environment: `CODEX_CLI_PATH` is set, so the runtime resolves with
`source: "environment"`; `codex-cli` is `0.146.0-alpha.14`; the proxy runs under the bundled
Bun 1.3.14. No combos configured, `disabledModels` empty.

### Checks

---

Unrelated aside, happy to split out if useful: model discovery for the `adapter: "google"` /
`googleMode: "cloud-code-assist"` provider fails with HTTP 404 and degrades to the static
catalog (`Provider model discovery for "google-antigravity" failed with HTTP 404
[urlClass=provider-models, fallback=configured]`).

</details>

mihneaptu:
Fixes the root cause behind #606.

## Problem

`loadBundledCodexCatalog()` spawns `codex --version` on every call, including warm
60s cache hits. On Windows that pushes `/api/claude-code` and `/v1/models` to 5-8s,
past the 3s budget `cmdClaude` allows each, so `ocx claude` aborts both and prints:

```
⚠ 모델 컨텍스트 정보를 불러오지 못했습니다 — 1M 자동 표시는 이번 실행에서 생략됩니다.
⚠ Gateway model cache could not be refreshed; the model picker may be stale.
```

The consequences are quiet: `[1m]` marking is skipped for the run, and
`~/.claude/cache/gateway-models.json` is never rewritten. Since Claude Code only
refreshes that cache itself when it holds a credential — and the
subscription-preserving launch deliberately injects none — the picker keeps serving
whatever is on disk. Mine was 6 days stale (80 entries vs 98 current).

## Root causes

**1. The bundled-catalog cache can never be hit.** `candidates` is computed in an
IIFE that runs *before* the `bundledCatalogCache` lookup, because `cacheKey` is
assigned inside it — so the runtime resolve happens on every call. That would still
be cheap if it were memoized, but the call forwarded its own already-defaulted
`execFileSync`:

```ts
const execFile = deps.execFileSync ?? (execFileSync as unknown as ExecFile);
// ...
resolveAndPersistCodexRuntime({ execFileSync: execFile, /* always truthy */ });
```

and `resolveCacheKey()` deliberately refuses to memoize injected-dependency
resolves. The guard is right in intent (injected deps imply a test double), but this
caller defeated it for the real-runtime path. Now only a genuinely injected
`execFileSync` is forwarded.

**2. The persist write cleared its own memo.** `resolveAndPersistCodexRuntime()`
wrote on any non-fallback selection. `persistCodexRuntime()` ends with
`resolveCache = null`, and `persistedRuntimeCacheStamp()` folds the on-disk
`updatedAt` into the cache key — so each write both cleared the memo and changed the
key it would have been stored under, even when the resolved runtime was
byte-identical to what was already persisted. Now it only writes when the selection
actually changed.

**3. Cold reads probed every `codex` on PATH.** The catalog path passed
`discoverAlternatives: deps.discoverAlternatives` (i.e. `undefined`), which
`resolveCodexRuntimeUncached()` treats as "discover". It then probed every candidate
to populate `newerAvailable`, each probe spawning `codex --version` with a fresh
`mkdtempSync` sandbox:

```
resolveCodexRuntime({})                              1224 ms   failures: 102
resolveCodexRuntime({ discoverAlternatives: false })    60 ms
```

Both select the identical runtime, and catalog loading never reads `newerAvailable`
or `failures`. Defaulted to `false` here; callers that want discovery diagnostics
still opt in explicitly.

## Measurements

`/api/claude-code` over HTTP, Windows 11:

| | before | after |
|---|---|---|
| cold (fresh proxy) | 5000-8300 ms | 165 ms |
| warm | 6300 ms | ~25 ms |
| after 65s idle (cache expired) | 6325 ms | 1440 ms |

`loadBundledCodexCatalog()` three times in a row, nothing changing on disk:

```
before:  1214 ms / 1136 ms / 1186 ms
after:   1483 ms /    1 ms /    0 ms
```

After the fix `ocx claude` runs clean and `gateway-models.json` refreshed from 80
stale entries to 98 current ones, 25 carrying `[1m]`.

## Tests

Three regression tests in `tests/codex-runtime.test.ts`, all failing on `dev` and
passing here:

- **`repeated catalog reads reuse the runtime probe instead of respawning it`** —
  uses a real launcher script that appends to a log on each `--version`, so it counts
  actual subprocess spawns rather than mocking them. Asserts the count stops growing
  across repeated reads. Note it deliberately allows the *second* read to probe once
  more: the first read has no persisted selection yet, so it legitimately writes one
  and that write clears the memo. Observed progression is `[1,2,2,2,2,2]` — the
  invariant is convergence, not a single probe. The POSIX launcher uses shell built-ins
  only, so the test remains isolated with `PATH=""` on Linux and macOS.
- **`repeat resolveAndPersistCodexRuntime keeps an unchanged selection's persisted
  stamp`** — injects a monotonic clock and asserts an identical selection does not
  rewrite `updatedAt`.
- **`treats missing persisted and resolved versions as the same selection`** —
  covers JavaScript callers that surface `undefined` despite the TypeScript `null`
  contract, preventing the same rewrite/memo invalidation loop for unversioned runtimes.

## Verification

- `bun run typecheck` — clean
- `bun test tests/codex-runtime.test.ts tests/codex-catalog*.test.ts` — **145 pass,
  0 fail** across 5 files (655 assertions)
- Confirmed all three regression tests fail against the relevant prior implementation
- POSIX launcher smoke-tested through Git Bash with `PATH=""`

One gap I should be upfront about: the full `bun run test` sweep did not complete on
my machine. Bun 1.3.14 on Windows panics with `Internal assertion failure` at
~2.9GB peak RSS partway through the 415-file suite — the same runtime memory issue
`ocx doctor` warns about on this platform. It is not triggered by this change, so
please treat CI's cross-platform run as authoritative for the full suite.

## Suggestions beyond this fix (not implemented here)

1. **Check `bundledCatalogCache` before resolving the runtime.** Deriving the cache
   key from the very probe the cache exists to avoid means a warm hit can only ever
   save the `codex debug models` call (~87ms), never the probe (~1.2s) that
   dominates. A cheaper key would let warm hits skip both. I left the ordering alone
   to keep this change minimal.
2. **Attribute the timeout in the warning text.** Both messages read like auth or
   network faults; something like `timed out after 3000ms` would point straight at
   the cause. I went looking at OAuth state and provider connectivity first.
3. **Consider making the 3s budget configurable.** Even fully patched, a cold read is
   ~1.4s here, so a slower machine or longer `PATH` could still trip it silently.

## Tradeoff worth reviewing

Defaulting `discoverAlternatives` to `false` on the catalog path means catalog-path
callers no longer receive `newerAvailable` diagnostics. Nothing in that path consumes
them today, but if you would rather keep discovery observable, moving it to a slow
background interval instead of every catalog read would preserve the diagnostics
while keeping the request path fast. Happy to rework it that way.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Codex runtime caching so repeated bundled catalog reads reuse the existing version probe.
  * Avoids rewriting the persisted Codex runtime when the resolved selection is unchanged, preserving timestamps.
  * Improved bundled runtime discovery behavior when optional injected execution settings are present, including safer defaults for alternative discovery.
* **Tests**
  * Added/expanded coverage for cache warm-start behavior and persisted-selection stability.

### DIFF
--- a/src/codex/catalog/bundled.ts
+++ b/src/codex/catalog/bundled.ts
@@ -151,14 +151,21 @@ export function loadBundledCodexCatalog(deps: BundledCatalogDeps = {}): RawCatal
   let cacheKey: string | null = null;
   const candidates = deps.commandCandidates?.() ?? (() => {
     const resolved = resolveAndPersistCodexRuntime({
-      execFileSync: execFile,
+      // Forward an INJECTED execFileSync only. Passing the real one unconditionally made
+      // resolveCacheKey() bail out (it refuses to memoize injected-dep resolves), so every
+      // catalog read re-ran the ~1s `codex --version` probe even on a warm cache hit.
+      ...(deps.execFileSync ? { execFileSync: deps.execFileSync } : {}),
       configDir: deps.configDir,
       env: deps.env,
       platform: deps.platform,
       existsSync: deps.existsSync,
       readFileSync: deps.readFileSync,
       now: deps.now,
-      discoverAlternatives: deps.discoverAlternatives,
+      // Catalog loading only consumes `resolved.runtime.command`, never `newerAvailable`.
+      // Full PATH discovery probes every candidate launcher (100+ on a dev machine, ~1.2s),
+      // which alone can exceed the 3s budget `ocx claude` allows /api/claude-code. Priority
+      // selection is identical either way; callers wanting discovery diagnostics opt in.
+      discoverAlternatives: deps.discoverAlternatives ?? false,
     });
     if (useCache) {
       cacheKey = [
--- a/src/codex/runtime.ts
+++ b/src/codex/runtime.ts
@@ -510,7 +510,7 @@ export function resolveAndPersistCodexRuntime(
   const selectionUnchanged = persistedRuntime !== null
     && persistedRuntime.command === result.runtime.command
     && persistedRuntime.source === result.runtime.source
-    && (persistedRuntime.selectedVersion ?? null) === result.runtime.version;
+    && (persistedRuntime.selectedVersion ?? null) === (result.runtime.version ?? null);
   if (result.runtime.command && result.runtime.source !== "fallback" && !selectionUnchanged) {
     try {
       persistCodexRuntime(result.runtime, deps);
--- a/src/codex/runtime.ts
+++ b/src/codex/runtime.ts
@@ -502,7 +502,16 @@ export function resolveAndPersistCodexRuntime(
   deps: ResolveCodexRuntimeDeps = {},
 ): ResolveCodexRuntimeResult {
   const result = resolveCodexRuntime(deps);
-  if (result.runtime.command && result.runtime.source !== "fallback") {
+  // Only WRITE when the selection actually changed. persistCodexRuntime() clears the
+  // in-process resolve memo and persistedRuntimeCacheStamp() folds `updatedAt` into the
+  // cache key, so an unconditional rewrite made every caller re-run the ~1s
+  // `codex --version` probe even when the resolved runtime was byte-identical.
+  const persistedRuntime = loadPersistedCodexRuntime(deps);
+  const selectionUnchanged = persistedRuntime !== null
+    && persistedRuntime.command === result.runtime.command
+    && persistedRuntime.source === result.runtime.source
+    && (persistedRuntime.selectedVersion ?? null) === result.runtime.version;
+  if (result.runtime.command && result.runtime.source !== "fallback" && !selectionUnchanged) {
     try {
       persistCodexRuntime(result.runtime, deps);
     } catch (error) {
--- a/tests/codex-runtime.test.ts
+++ b/tests/codex-runtime.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, test } from "bun:test";
-import { existsSync, mkdtempSync, writeFileSync } from "node:fs";
+import { existsSync, mkdtempSync, readFileSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join, dirname } from "node:path";
 import {
@@ -285,6 +285,124 @@ describe("resolveCodexRuntime", () => {
     expect(failed.runtime.command).toBe("C:\\keep\\codex.exe");
     expect(typeof failed.persistError).toBe("string");
     expect(failed.persistError!.length).toBeGreaterThan(0);
+  });
+
+  test("repeated catalog reads reuse the runtime probe instead of respawning it", async () => {
+    const { chmodSync, mkdirSync } = await import("node:fs");
+    const {
+      loadBundledCodexCatalog,
+      resetBundledCatalogCacheForTests,
+    } = await import("../src/codex/catalog/bundled");
+
+    const home = tempConfigDir();
+    const binDir = join(home, "bin");
+    mkdirSync(binDir, { recursive: true });
+    const bin = process.platform === "win32" ? join(binDir, "codex.cmd") : join(binDir, "codex");
+
+    // Count --version probes by having the launcher append a line per invocation.
+    const probeLog = join(binDir, "probes.log");
+    const catalog = JSON.stringify({
+      models: [{
+        slug: "gpt-5.5",
+        base_instructions: "x",
+        supported_reasoning_levels: [{ effort: "medium", description: "medium" }],
+        default_reasoning_level: "medium",
+      }],
+    });
+    writeFileSync(join(binDir, "catalog.json"), `${catalog}\n`, "utf8");
+    if (process.platform === "win32") {
+      writeFileSync(bin, [
+        "@echo off",
+        `if "%~1"=="--version" (`,
+        `  echo probe>>"%~dp0probes.log"`,
+        "  echo codex-cli 0.145.0-alpha.30",
+        "  exit /b 0",
+        ")",
+        `type "%~dp0catalog.json"`,
+        "",
+      ].join("\r\n"), "utf8");
+    } else {
+      writeFileSync(bin, [
+        "#!/bin/sh",
+        `d=$(dirname "$0")`,
+        `if [ "$1" = "--version" ]; then`,
+        `  echo probe >> "$d/probes.log"`,
+        `  echo "codex-cli 0.145.0-alpha.30"`,
+        "  exit 0",
+        "fi",
+        `cat "$d/catalog.json"`,
+        "",
+      ].join("\n"), "utf8");
+      chmodSync(bin, 0o755);
+    }
+
+    const countProbes = (): number => {
+      if (!existsSync(probeLog)) return 0;
+      return readFileSync(probeLog, "utf8").split("\n").filter(line => line.trim().length > 0).length;
+    };
+
+    const previousHome = process.env.OPENCODEX_HOME;
+    const previousCli = process.env.CODEX_CLI_PATH;
+    const previousPath = process.env.PATH;
+    process.env.OPENCODEX_HOME = home;
+    process.env.CODEX_CLI_PATH = bin;
+    process.env.PATH = "";
+    resetCodexRuntimeResolveCacheForTests();
+    resetBundledCatalogCacheForTests();
+
+    try {
+      expect(loadBundledCodexCatalog()?.models?.[0]?.slug).toBe("gpt-5.5");
+      expect(countProbes()).toBeGreaterThan(0);
+
+      // The first read has no persisted selection yet, so it legitimately writes one and the
+      // write clears the resolve memo — the second read re-probes once and then persists
+      // nothing. From there the count must STOP GROWING.
+      expect(loadBundledCodexCatalog()?.models?.[0]?.slug).toBe("gpt-5.5");
+      const warm = countProbes();
+
+      // Warm reads must not respawn the probe. Two regressions broke this and made the count
+      // grow once per read: (1) loadBundledCodexCatalog forwarding its own already-defaulted
+      // execFileSync, which opts resolveCacheKey() out of memoizing, and (2) an unconditional
+      // persist whose `updatedAt` both clears the memo and rekeys it. Either one made every
+      // catalog read spawn `codex --version` (~1s), pushing /api/claude-code past the 3s
+      // budget ocx claude allows and silently skipping the gateway-model cache refresh.
+      for (let i = 0; i < 4; i++) {
+        expect(loadBundledCodexCatalog()?.models?.[0]?.slug).toBe("gpt-5.5");
+      }
+      expect(countProbes()).toBe(warm);
+    } finally {
+      if (previousHome === undefined) delete process.env.OPENCODEX_HOME;
+      else process.env.OPENCODEX_HOME = previousHome;
+      if (previousCli === undefined) delete process.env.CODEX_CLI_PATH;
+      else process.env.CODEX_CLI_PATH = previousCli;
+      if (previousPath === undefined) delete process.env.PATH;
+      else process.env.PATH = previousPath;
+      resetCodexRuntimeResolveCacheForTests();
+      resetBundledCatalogCacheForTests();
+    }
+  });
+
+  test("repeat resolveAndPersistCodexRuntime keeps an unchanged selection's persisted stamp", () => {
+    const configDir = tempConfigDir();
+    const deps = {
+      configDir,
+      env: { CODEX_CLI_PATH: "C:\\keep\\codex.exe", PATH: "" },
+      platform: "win32" as const,
+      existsSync: () => true,
+      execFileSync: (() => "codex-cli 0.145.0-alpha.30") as RuntimeExecFile,
+    };
+
+    const first = resolveAndPersistCodexRuntime(deps);
+    expect(first.persistError).toBeUndefined();
+    const firstStamp = loadPersistedCodexRuntime({ configDir })?.updatedAt;
+    expect(typeof firstStamp).toBe("string");
+
+    // An identical selection must not rewrite the file: the write clears the resolve memo
+    // and its `updatedAt` feeds the memo cache key, so rewriting defeats caching entirely.
+    const second = resolveAndPersistCodexRuntime(deps);
+    expect(second.runtime.command).toBe(first.runtime.command);
+    expect(second.runtime.version).toBe(first.runtime.version);
+    expect(loadPersistedCodexRuntime({ configDir })?.updatedAt).toBe(firstStamp);
   });
 
   test("catalog clamp clears diagnostics inside deps.configDir when probe fails", async () => {
--- a/tests/codex-runtime.test.ts
+++ b/tests/codex-runtime.test.ts
@@ -324,13 +324,15 @@ describe("resolveCodexRuntime", () => {
     } else {
       writeFileSync(bin, [
         "#!/bin/sh",
-        `d=$(dirname "$0")`,
+        "d=${0%/*}",
         `if [ "$1" = "--version" ]; then`,
-        `  echo probe >> "$d/probes.log"`,
-        `  echo "codex-cli 0.145.0-alpha.30"`,
+        `  printf "%s\\n" probe >> "$d/probes.log"`,
+        `  printf "%s\\n" "codex-cli 0.145.0-alpha.30"`,
         "  exit 0",
         "fi",
-        `cat "$d/catalog.json"`,
+        `while IFS= read -r line || [ -n "$line" ]; do`,
+        `  printf "%s\\n" "$line"`,
+        `done < "$d/catalog.json"`,
         "",
       ].join("\n"), "utf8");
       chmodSync(bin, 0o755);
@@ -384,12 +386,14 @@ describe("resolveCodexRuntime", () => {
 
   test("repeat resolveAndPersistCodexRuntime keeps an unchanged selection's persisted stamp", () => {
     const configDir = tempConfigDir();
+    let now = 0;
     const deps = {
       configDir,
       env: { CODEX_CLI_PATH: "C:\\keep\\codex.exe", PATH: "" },
       platform: "win32" as const,
       existsSync: () => true,
       execFileSync: (() => "codex-cli 0.145.0-alpha.30") as RuntimeExecFile,
+      now: () => ++now,
     };
 
     const first = resolveAndPersistCodexRuntime(deps);
@@ -403,6 +407,38 @@ describe("resolveCodexRuntime", () => {
     expect(second.runtime.command).toBe(first.runtime.command);
     expect(second.runtime.version).toBe(first.runtime.version);
     expect(loadPersistedCodexRuntime({ configDir })?.updatedAt).toBe(firstStamp);
+  });
+
+  test("treats missing persisted and resolved versions as the same selection", () => {
+    const configDir = tempConfigDir();
+    const statePath = join(configDir, "codex-runtime.json");
+    writeFileSync(statePath, JSON.stringify({
+      version: 1,
+      command: "codex",
+      source: "environment",
+      updatedAt: "2026-01-01T00:00:00.000Z",
+    }));
+
+    const previousHome = process.env.OPENCODEX_HOME;
+    process.env.OPENCODEX_HOME = configDir;
+    resetCodexRuntimeResolveCacheForTests();
+    const deps = { env: { PATH: "" }, discoverAlternatives: false };
+
+    try {
+      const cached = resolveCodexRuntime(deps);
+      expect(cached.runtime.source).toBe("fallback");
+      cached.runtime.source = "environment";
+      (cached.runtime as { version?: string | null }).version = undefined;
+      const before = readFileSync(statePath, "utf8");
+
+      resolveAndPersistCodexRuntime(deps);
+
+      expect(readFileSync(statePath, "utf8")).toBe(before);
+    } finally {
+      if (previousHome === undefined) delete process.env.OPENCODEX_HOME;
+      else process.env.OPENCODEX_HOME = previousHome;
+      resetCodexRuntimeResolveCacheForTests();
+    }
   });
 
   test("catalog clamp clears diagnostics inside deps.configDir when probe fails", async () => {


lidge-jun:
## The P1 on the POSIX launcher does not reproduce — measured

Reviewing this from the maintainer side. You already addressed the bot's P1 in `056aa2d6e` with the builtin launcher (`d=${0%/*}` plus a `while IFS= read -r line || [ -n "$line" ]` loop), and that form is correct and slightly better than what I would have written — the `|| [ -n "$line" ]` guard keeps a final line with no trailing newline.

Worth recording, though: the failure the P1 predicts does not actually occur, so you were not blocked by a real defect. I measured it three ways against `dev`.

**1. The catalog does load.** I temporarily instrumented `loadBundledCodexCatalog()` inside the `persisted runtime stamp busts resolve memo` test on `dev` (the version still using `cat "$(dirname "$0")"`):

```
[PROBE] oldCatalog = loaded, models=1
[PROBE] newCatalog = loaded, models=1
(pass) resolveCodexRuntime > persisted runtime stamp busts resolve memo [140.50ms]
```

Not `null` — the launcher produced its catalog.

**2. A `null` return would have failed the test, not passed it.** The natural worry is that optional chaining makes a `null` catalog satisfy `.toBe(false)`. It does not:

```
expect(undefined).toBe(false)
  Expected: false
  Received: undefined
(fail)
```

So the assertion pair is a real oracle for catalog loading; it is not silently green.

**3. `PATH=""` never reaches the child.** `runCodexDebugModels()` at `src/codex/catalog/bundled.ts:137-143` calls `execFile` without an `env` option, so the child inherits the environment captured at process start — a later `process.env.PATH = ""` does not propagate. Same launcher, three conditions:

```
A: process.env.PATH='' , no env opt      OK    {"models":[{"slug":"x"}]}
B: explicit env {...process.env}         THREW 127
C: explicit env {PATH:''}                THREW 127
```

A is the path this code actually takes. B and C are what you get when `env` is passed explicitly, which is presumably how the reviewer modelled it.

**One real observation this did surface**, unrelated to your PR: because `PATH=""` does not propagate, the isolation that line intends is not in effect. The test is still safe today — it pins the launcher through `CODEX_CLI_PATH` and `persistCodexRuntime()` with absolute paths — but the guard is decorative rather than load-bearing. Changing how `runCodexDebugModels` passes `env` would touch probe behavior and overlaps your change, so I am leaving it for a maintainer call rather than patching around you.

Nothing needed from you on this. Your builtin launcher is a fine improvement regardless.

lidge-jun:
## Correction to my previous comment — two things I got wrong

A follow-up review caught two errors in what I wrote above, and one of them was bad advice about this codebase. Correcting both, with measurements.

**1. The env behavior is Bun-specific, not general.** I wrote that "the child inherits the environment captured at process start" as though that were ordinary `execFile` behavior. It is not. Same script, both runtimes:

```
process.env.PATH = "";  execFileSync(bin, [], { encoding: "utf8" })

BUN   -> CHILD_PATH=[/Users/.../bin:/opt/homebrew/bin:...]   ← original PATH
NODE  -> CHILD_PATH=[]                                        ← mutation propagates
```

Bun snapshots the environment at startup for child spawns and does not observe later `process.env` mutations; Node does. So **the P1's reasoning is correct under Node semantics** — it simply does not fire on our Bun runtime. The bot was not wrong about the shell; it was reasoning about a runtime we do not use.

That also means the test's green state rests on a Bun implementation detail. If Bun ever aligns with Node here, `tests/codex-runtime.test.ts:360` starts failing with no code change. Worth knowing.

There is a related asymmetry I should have spotted: `src/codex/runtime.ts:261` *does* pass `env` explicitly (`{ ...(deps.env ?? process.env), CODEX_HOME: probeHome }`), so within one test run the `--version` probe sees `PATH=""` while the catalog probe sees the original PATH. `--version` survives only because `echo` is a shell builtin needing no lookup. That asymmetry is the actual reason the suite is green.

**2. `PATH=""` is load-bearing — I was wrong to call it decorative.** This is the part worth retracting properly, since it was advice about your test's surroundings.

`PATH` is not only consumed by child processes. `pathCandidates()` at `src/codex/runtime.ts:308-312` reads `deps.env ?? process.env` and splits `PATH` to build the candidate binary list in-process, where the mutation is visible immediately. Measured with a fake `codex` planted on PATH:

```
PATH=<fakedir> -> source: path      version: 9.9.9
PATH=''        -> source: fallback  version: null
```

So that line does exactly what it was written to do: it keeps a machine-installed `codex` out of the candidate list. The `CODEX_CLI_PATH` pinning is a second layer, not a substitute. Anyone acting on my earlier "decorative" framing and deleting it would have made the test machine-dependent.

My NOOP conclusion stands — nothing here needs fixing, and nothing is needed from you. But the reasoning I published was wrong in both places, so it seemed worth correcting rather than leaving on the record.

mihneaptu:
Ping on CI approval — the two `pull_request` workflows (Cross-platform CI, React Doctor) are sitting in `action_required`, so the full suite has not run on this branch yet.

Since I flagged in the description that my local `bun run test` sweep could not complete (Bun 1.3.14 panics with `Internal assertion failure` at ~2.9GB RSS on Windows partway through the 415-file suite), CI is the authoritative signal here, and I would rather not push for review before it is green.

Current state, for convenience:

- All three review threads are resolved — the bot P1 on the POSIX launcher was addressed in `056aa2d6e` with a builtins-only launcher (`d=${0%/*}` plus a `while IFS= read -r line || [ -n "$line" ]` loop).
- Still `mergeable` against `dev`. The two commits it is behind (`c48a6eee3`, `ca7b1049f`) touch unrelated files, so there is no overlap with the three files here.
- Head is `056aa2d6e`; I have not pushed since, so an approved run would cover the reviewed commit.

Whenever one of you has a moment to approve the run, thanks.

lidge-jun:
**LAND-AFTER** — merges the fix, but does not close #606. Worth being precise about the difference.

**What it actually fixes.** The repeated `codex --version` spawning is genuinely gone on warm reads, through two changes that are easy to under-sell. Forwarding only an *injected* `execFileSync` (`src/codex/catalog/bundled.ts:153-158`) is the important one — passing the real default unconditionally made `resolveCacheKey()` refuse to memoize, so every catalog read re-ran the ~1s probe even on a cache hit. Defaulting `discoverAlternatives` to `false` removes the PATH-wide sweep, which on a developer machine with 100+ candidate launchers is ~1.2s on its own and can blow the 3s budget without any other help. And no longer rewriting an unchanged persisted runtime selection (`src/codex/runtime.ts:500-524`) stops the write that was clearing and rekeying the resolver memo.

The comment in the diff explaining the first one is the kind of note that saves the next person an hour.

**What it does not do.** #606 asked for the catalog cache to be consulted before any runtime resolution. The structure is unchanged: the `candidates` builder calls `resolveAndPersistCodexRuntime()` first, and `bundledCatalogCache` is only checked afterwards — necessarily, since the cache key is derived from the resolved runtime. So a cold read, or one where the resolver memo or catalog entry has expired, still performs runtime resolution before anything cheap is consulted.

That is not a criticism of the PR. Memoizing the resolve is a smaller, safer change than restructuring the key derivation, and it removes the pathological repeat case that made #606 visible. It just means merging this does not entitle us to close the issue.

**The test is the best part.** `tests/codex-runtime.test.ts:290-385` uses a real launcher script that appends to a file on every `--version` invocation, then asserts the probe count converges and stops growing. That observes the behaviour instead of asserting on source, which is exactly right for a "stop doing this repeatedly" fix. The persisted-timestamp and `undefined`-versus-`null` version cases at `:387-441` are good supporting coverage.

What happens next: this lands after ordinary maintainer review — it merges cleanly against current `dev` and `dev` CI is fully green. #606 stays open pending a re-measurement on Windows of cold, warm and >60s endpoint timings against the 3s budget. I will note that on the issue so nobody closes it on this PR's back.

lidge-jun:
Closing as superseded. `716f39cb6 fix(codex): key catalog cache by runtime and address CodeRabbit follow-ups` landed this on `dev`: `src/codex/catalog/bundled.ts:153-168` forwards only an injected `execFileSync` (so the cache key no longer bails out and re-probes) and defaults `discoverAlternatives` to `false` on the catalog path. `git cherry` reports both commits here as patch-equivalent.

#606 is closed with that evidence. Thanks — an earlier triage round recorded this head as not-yet-on-dev, and it landed since.

PR #611, opened 2026-07-28
feat(providers): add Volcengine Ark provider presets

yrooogerg:
## Summary

- add first-class presets for Volcengine Ark pay-as-you-go, Coding Plan, and Agent Plan
- keep the three billing routes explicit so Plan users do not accidentally send traffic to the pay-as-you-go endpoint
- add registry-derived `responsesPath` support for the Agent Plan native Responses endpoint
- provide curated static model catalogs and set `doubao-seed-2-1-pro-260628` as the pay-as-you-go default
- include current DeepSeek and GLM text models in the pay-as-you-go catalog
- document the providers in English, Simplified Chinese, Japanese, Korean, and Russian
- add focused registry, routing, adapter, GUI metadata, and reasoning-toggle regression coverage

## Why

Volcengine Ark is a major model and cloud provider, particularly for developers using Doubao and China-region model services. A built-in preset removes substantial manual configuration and gives the three products clear billing boundaries:

- `volcengine`: `https://ark.cn-beijing.volces.com/api/v3`
- `volcengine-coding-plan`: `https://ark.cn-beijing.volces.com/api/coding/v3`
- `volcengine-agent-plan`: `https://ark.cn-beijing.volces.com/api/plan/v3/responses`

The ordinary `/api/v3` route may incur pay-as-you-go charges even when the user has a Plan subscription, so keeping these as separate provider IDs is important for billing safety.

## Model discovery findings

Authenticated, read-only `/models` probes showed:

- pay-as-you-go returns an OpenAI-shaped catalog containing 126 resources
- Coding Plan returns the same broad 126-resource catalog and omits the plan routing alias `ark-code-latest`
- Agent Plan returns 404 for `/models`

The 126-resource catalog mixes chat models with embeddings, image generation, video generation, 3D generation, and historical model versions. Publishing it directly would expose resources that the Codex chat adapter cannot call and models that a Plan subscription may not authorize. The three presets therefore use curated static catalogs with `liveModels: false`.

A minimal pay-as-you-go smoke request against `doubao-seed-2-1-pro-260628` succeeded. The available test key did not include Coding Plan or Agent Plan entitlements: the Coding request returned `InvalidSubscription`, and the Agent request returned `UnsupportedModel`. No credential is stored in the repository, test fixtures, or PR content.

## Validation

- `bun run typecheck`
- `bun run test` — 5,705 passed, 0 failed
- focused Volcengine and registry tests — 36 passed, 0 failed
- `bun run privacy:scan`
- docs build — 141 pages
- `git diff --check`

## Follow-up opportunities

Volcengine also has strong multimodal generation products. Follow-up work can evaluate dedicated integrations for:

- Seedream image generation
- Seedance video generation

Those APIs have different request/response shapes and media lifecycle requirements, so they are intentionally outside this provider-preset PR and should be designed as explicit image/video capabilities.

## Summary by CodeRabbit

* **New Features**
  * Added Volcengine Ark, Coding Plan, and Agent Plan providers with plan-specific model catalogs and defaults, including native Responses routing.
  * Increased built-in provider preset catalog to 63.
* **UI / Localization**
  * Added localized display names for Volcengine provider variants and made provider labels i18n-aware across provider UI surfaces.
* **Documentation**
  * Updated Providers guides with expanded preset counts and a larger API-key provider catalog, plus clarified Volcengine billing routes and model-catalog behavior.
* **Tests**
  * Expanded Volcengine provider, routing, model, and reasoning/thinking coverage.


Wibias (dismissed):

**Verdict: REQUEST CHANGES**

### Security
No medium+ findings on the new Volcengine presets (static catalog + key auth). No new secret handling.

### Bug review / open bot threads
| Item | Status |
| --- | --- |
| Rabbit: localize Volcengine labels | **Fixed** (i18n keys + locale files) |
| Codex P1: preserve DeepSeek reasoning on Coding Plan tool turns | **Still open** — `volcengine-coding-plan` has no `preserveReasoningContentModels` for `deepseek-v4-pro` / `deepseek-v4-flash` |
| Codex P1: DeepSeek V4 effort mappings on Coding Plan | **Still open** — no `modelReasoningEfforts` / `modelReasoningEffortMap` on that preset (will advertise/send raw Codex ladder) |
| Codex P2: image inputs for multimodal Coding Plan models | **Still open** — `kimi-k2.6` / `minimax-m3` lack `modelInputModalities` while `liveModels: false` |
| Codex P2: carry `responsesPath` on custom rename | **Partially fixed** — Agent Plan declares `responsesPath: "/responses"` and derive seeds it by registry id; rename-away-from-id still needs confirmation |

Please mirror the metadata already present on `volcengine` / Alibaba DeepSeek presets onto `volcengine-coding-plan` (and Agent Plan where applicable), with focused registry/parity tests.

### Notes
- Fork PR: CI surface incomplete on head SHA; blockers above are from code review of `upstream/dev...HEAD`.

yrooogerg:
@Wibias check this

lidge-jun:
**NEEDS-CHANGES** — one of the items is a real credential-safety gap, not paperwork.

**The conflict is smaller than it looks.** GitHub shows 29 changed files, but `git merge-tree` against the current `dev` head reports exactly one conflicted file and one hunk: `gui/src/pages/Providers.tsx`, where two adjacent declarations collide. Current `dev` has `buildAccountLoginStatus(config, oauthStatusWithCodex)`; this branch changes the preceding line to `buildAddModalAccountRows(config, oauthProviders, t)`. Keep both:

```ts
const addModalAccountRows = buildAddModalAccountRows(config, oauthProviders, t);
const accountLoginStatus = buildAccountLoginStatus(config, oauthStatusWithCodex);
```

Everything else auto-merges.

**The gap worth fixing regardless of the evidence question.** None of the three presets sets `preserveCustomDestination`. I counted: this branch adds three new key-provider IDs with zero occurrences, while the comparable Chutes/Hyperbolic submission sets it four times and `dev` has two. Without it, a user who already saved a custom provider named `volcengine` with a matching adapter and auth shape can have their destination canonicalized onto Ark. That is someone's API key going somewhere they did not choose. The tests do not cover that case either.

Add `preserveCustomDestination` to all three and a regression that saves a same-name custom provider with a different base URL and asserts the host survives derivation.

**The evidence gate.** `MAINTAINERS.md:33-42` treats a preset as a credential-destination change:

- documented endpoints — the body states URLs and probe results, but a probe is not vendor documentation
- authenticated `GET /v1/models` — not applicable, all three set `liveModels: false`
- ToS and operating legal entity — missing
- routing authorization — **missing and applicable**: the Coding and Agent plan catalogs bundle DeepSeek, GLM, Kimi and MiniMax models, not just first-party Doubao
- named maintenance owner — missing
- citable verification date — missing; the probe descriptions carry no date

If that evidence proves hard to assemble, the same clause points at an inert `src/providers/free-directory.ts` row as the preferred landing spot. Directory rows never become canonical login or routing presets, so they carry no credential-destination risk.

The three-way split (`volcengine`, `volcengine-coding-plan`, `volcengine-agent-plan`, the last on `openai-responses` with `responsesPath: "/responses"`) is well constructed, and the tests bind real endpoint construction, thinking mapping, and `responsesPath` preservation under a custom name. The structure is not the problem.

What happens next: resolve the single GUI hunk, add `preserveCustomDestination` plus its regression, assemble the evidence package including routing authorization, then request explicit security review. `dev` CI is fully green now, so your checks will reflect this branch.

lidge-jun (dismissed):

Two blocking items beyond the evidence gate:

1. None of the three presets sets `preserveCustomDestination`. A user who already saved a custom provider named `volcengine` with a matching adapter and auth shape can have their destination canonicalized onto Ark — that is their API key going somewhere they did not choose. Add it to all three plus a regression.
2. The merge conflict is one hunk in `gui/src/pages/Providers.tsx`; keep both the `buildAddModalAccountRows` and `buildAccountLoginStatus` declarations.

Evidence gate items still missing: primary-source endpoint documentation, ToS and operating legal entity, routing authorization (the Coding and Agent catalogs bundle DeepSeek, GLM, Kimi and MiniMax), a named maintenance owner, and a citable verification date.

Full reasoning in my earlier comment.

Ingwannu (dismissed):
Security review completed against `bdfc05c230fde700e531ea2bae339a96ffb3df64`.

Requesting changes because the canonical credential-destination evidence gate in `MAINTAINERS.md:35-42` is still incomplete.

The earlier code-level credential-retargeting issue is fixed: all three presets now set `preserveCustomDestination`, and the focused regression covers a same-name custom destination. The requested DeepSeek reasoning metadata, multimodal declarations, localized labels, and Agent Plan `responsesPath` propagation are also present on the current head.

The remaining blocker is evidence, not implementation shape. The PR body still provides endpoint strings and private probe results, but no citable primary-source package for:

- documented Ark pay-as-you-go, Coding Plan, and Agent Plan endpoints;
- Volcengine's applicable terms and operating legal entity;
- authorization for the Coding/Agent Plan routes to expose the bundled DeepSeek, GLM, Kimi, and MiniMax catalogs;
- a named maintenance owner; and
- a citable verification date.

Please add those sources to the PR description. If public routing authorization cannot be established, keep the entry inert/non-canonical as required by the repository policy. Do not merge the current head as a canonical preset.

yrooogerg:
Evidence package for the remaining security-review gate, verified on 2026-07-31 by @yrooogerg.

- Provider / operating entity / terms:
  - Volcengine Terms: https://www.volcengine.com/docs/6256/64903
  - Product and Service Agreement: https://www.volcengine.com/docs/6256/68938
  - Ark-specific terms: https://www.volcengine.com/docs/82379/1104498
  - The Volcengine Terms identify the website/operator as 北京火山引擎科技有限公司.

- Pay-as-you-go Ark endpoint:
  - Base URL / auth rules: https://www.volcengine.com/docs/82379/1298459
  - Chat API: https://www.volcengine.com/docs/82379/1494384
  - Effective OpenAI-compatible base URL used by this PR: https://ark.cn-beijing.volces.com/api/v3

- Coding Plan endpoint and model scope:
  - Coding Plan overview: https://www.volcengine.com/docs/82379/1925114
  - Quick start: https://www.volcengine.com/docs/82379/1928261
  - Third-party tools / OpenAI-compatible access: https://www.volcengine.com/docs/82379/2160841
  - Effective base URL used by this PR: https://ark.cn-beijing.volces.com/api/coding/v3
  - Official overview states the plan supports GLM, MiniMax, Kimi, DeepSeek, and Doubao-family models.

- Agent Plan endpoint and model scope:
  - Agent Plan overview: https://www.volcengine.com/docs/82379/2366394
  - Agent Plan third-party tools / OpenAI-compatible access: https://www.volcengine.com/docs/82379/2373746
  - Agent Plan supported-model API: https://www.volcengine.com/docs/82379/2546385
  - Agent Plan model launch notes: https://www.volcengine.com/docs/82379/2578669
  - Effective base URL used by this PR: https://ark.cn-beijing.volces.com/api/plan/v3

- Maintenance owner:
  - @yrooogerg owns this preset initially and will refresh or remove stale model metadata when Volcengine changes the public plan catalog.

One boundary to call out: I found public Volcengine documentation that these Plan catalogs expose the listed third-party model families, but I did not find a separate public legal authorization document for Volcengine's third-party model routing beyond the product docs and Ark terms above.

lidge-jun (approved):
Approving after rebasing this onto current `dev` and re-running the gates myself.

## What I changed on your branch

Rebased `feat/volcengine-providers` from its 234-commit-old base onto current `dev` (now `8c8832137`), and resolved the conflicts:

- **Preset counts across five locales.** Both sides had moved: `dev` was at 62/51 and this branch claimed 63/52 against an older base. Neither was right after the rebase, so I counted the registry directly (`PROVIDER_REGISTRY.length` = 65: 54 key, 7 OAuth, 3 local, 1 forward) and set en/ja/ko/ru/zh-cn to 65/54.
- **`gui/src/pages/Providers.tsx`.** `dev` had switched `buildAccountLoginStatus` to `oauthStatusWithCodex` while this branch added the `t` argument to `buildAddModalAccountRows`. Kept both.

I also pushed one commit of my own:

- **`glm-5.1` → `glm-5.2`** in `VOLCENGINE_CODING_PLAN_MODELS` and `VOLCENGINE_AGENT_PLAN_MODELS`, plus the matching assertions in `tests/volcengine-providers.test.ts`. Volcengine's current published Plan model list carries `glm-5.2`; `glm-5.1` is not in it. Because both presets are `liveModels: false`, discovery cannot repair a stale id — it reaches the user as an upstream model error at request time. This is exactly the maintenance cost a static catalog takes on, so please keep an eye on that list going forward.

## Verification on the rebased head

- `bun run typecheck`
- `bun run test` — 6415 pass, 4 skip, 0 fail across 463 files
- `gui`: `bun test tests` — 463 pass, 0 fail
- `bun run lint:gui`, `bun run privacy:scan`, `bun run build:gui`
- Focused: `volcengine-providers`, `provider-registry-parity`, `tencent-siliconflow-providers`, `provider-workspace-data` — 92 pass

## On the credential-destination policy

`MAINTAINERS.md` asks for an evidence ledger on any new preset. Your PR body documents the endpoints, the three-way billing split, and your authenticated `/models` probe results, which covers the substance. What it does not state explicitly is the operating legal entity, a citable verification date, and a named maintenance owner. Please add those to the description — for the record, the entity behind the Ark terms is 北京火山引擎科技有限公司.

The three presets keep discovery registry-owned, set `liveModels: false` against a mixed 126-resource catalog rather than publishing embeddings and video models into the picker, and use `preserveCustomDestination` so an existing same-name custom provider is not silently retargeted. That is the right shape for this integration.

One note on the Plan endpoints: Volcengine documents them as subscription routes for AI coding tools. That is precisely what opencodex is, so the preset is inside the documented use, and it matches how `tencent-coding-plan` already ships. The `note` field is the surface that tells users which route they are on — keep it accurate if Volcengine changes those terms.

lidge-jun:
## Credential-destination evidence ledger

Posting this as owner because @Ingwannu's security review blocked on the `MAINTAINERS.md` evidence gate rather than on the code. I ran the research myself; sources below were opened, not read from search snippets. Verification date: **2026-08-01**.

| Gate requirement | Result | Primary source |
| --- | --- | --- |
| Documented OpenAI-compatible endpoints | satisfied | `/api/v3`, `/api/coding/v3` ([82379/1528783](https://www.volcengine.com/docs/82379/1528783), updated 2026-07-07), `/api/plan/v3` native Responses ([82379/2165245](https://www.volcengine.com/docs/82379/2165245)) |
| Authenticated `GET /v1/models` | **not applicable** | all three entries are `liveModels: false`; the clause is conditional on declaring live discovery |
| Terms of service + operating legal entity | satisfied | 北京火山引擎科技有限公司 — [ToS 6256/64903](https://www.volcengine.com/docs/6256/64903), published 2024-06-14, effective 2024-06-21; ICP 京ICP备20018813号-3 |
| Resale / routing authorization | satisfied, see below | [Codex CLI integration guide 82379/2556056](https://www.volcengine.com/docs/82379/2556056); [supported-client list 82379/2188957](https://www.volcengine.com/docs/82379/2188957) |
| Named maintenance owner | **still open** | not in the PR description — @yrooogerg please add it |
| Citable verification date | satisfied | 2026-08-01, this comment |

### On routing authorization

The gate asks for "resale or routing authorization for aggregators". `contributing.md` defines that as a gateway selling access to third-party models. This is not that:

1. **Nothing is resold or pooled.** Each user supplies their own Ark plan key and it serves only that user's session. The credential never leaves their machine.
2. **Volcengine documents the exact clients opencodex exists to serve.** There is a first-party Codex CLI integration guide instructing users to put a plan key in `~/.codex/config.toml`, and Claude Code, OpenCode, and OpenClaw appear in the supported-client list. The traffic opencodex forwards is a supported client's traffic.

I ran a deliberate counter-search for evidence against this — enforcement actions, key revocations, proxy crackdowns, and third-party model-owner objections across V2EX, 掘金, CSDN, Zhihu, GitHub, Reddit, and HN. No incident was found. Absence of a reported ban is not authorization on its own, which is why the disclosure below is part of this decision rather than optional.

### The defect this review found

Volcengine states plainly that plan quota is valid only inside supported AI coding tools, and that other API use of the key **may suspend the subscription or ban the account** ([article/37156](https://www.volcengine.com/article/37156), 2026-04-09; `curl`, Postman, and Dify are named exclusions in [article/37935](https://www.volcengine.com/article/37935)).

The presets did not say so. `tencent-coding-plan` already ships exactly this disclosure, so the two Volcengine Plan entries were inconsistent with an existing precedent, and a user could have lost their subscription without ever being warned. Fixed in `983b7b605`:

- both Plan `note` strings now carry the restriction, worded per product so Agent Plan does not quote Coding-Plan-specific policy as its own
- **the note is now resolved by destination, not by provider name.** The dashboard lets a preset be saved under a custom name; a renamed row kept the credential destination but silently lost the warning about it. Verified by ablation — reverting the resolver makes the new regression fail, restoring it makes it pass
- text-only Plan models are declared so the vision sidecar cannot advertise image input for models that cannot accept it, matching the Tencent precedent
- the restriction is documented in the provider guide

### Also landed on this branch

- rebased from a 234-commit-old base onto current `dev`, resolving preset-count drift across five locales (counted from the registry: 65 presets — 54 key, 7 OAuth, 3 local, 1 forward) and a `Providers.tsx` collision between `oauthStatusWithCodex` and the new `t` argument
- `glm-5.1` → `glm-5.2` in both Plan catalogs; the old id is absent from Volcengine's current published list, and `liveModels: false` means discovery cannot repair it
- resolved a `responsesPath` duplicate-identifier collision introduced by the rebase

### Verification

`bun run typecheck` clean · `bun run test` 6487 pass / 0 fail across 470 files · `gui` 463 pass / 0 fail · `lint:gui`, `privacy:scan`, `build:gui` green · CI 8/8 green on `688fd7715`.

One note on CI history: an earlier head was cancelled on `windows-latest` at the Test step. That was a pre-existing `dev` defect — the `dev` tip reproduced the identical cancellation with no Volcengine code present — and it was fixed on `dev` by #813. Rebasing onto that cleared it. A later macOS Bun segfault in `sidecar-tracker.test.ts` and a Windows Kiro OAuth timeout were flakes in files this PR does not touch; both passed on rerun.

Merging. @yrooogerg — please add the maintenance owner line to the description; that is the one gate item still outstanding, and it is yours to own going forward since these catalogs are static and will need updating when Volcengine changes its plan model list.

lidge-jun:
Owner decision on the credential-destination evidence gate, with the primary-source ledger it asks for. I also pushed one more commit to this branch; details below.

## Evidence ledger (verification date 2026-08-01, sources opened)

| `MAINTAINERS.md` requirement | Result | Primary source |
|---|---|---|
| Documented OpenAI-compatible endpoints | satisfied | `/api/v3`, `/api/coding/v3` ([82379/1528783](https://docs.volcengine.com/docs/82379/1528783), updated 2026-07-07), `/api/plan/v3` native Responses ([82379/2165245](https://docs.volcengine.com/docs/82379/2165245), 2026-05-28) |
| Authenticated `GET /v1/models` | **not applicable** — all three entries are `liveModels: false` |
| Terms of service + operating legal entity | satisfied | 北京火山引擎科技有限公司, [6256/64903](https://www.volcengine.com/docs/6256/64903), published 2024-06-14, effective 2024-06-21; ICP 京ICP备20018813号-3 |
| Routing authorization | satisfied for the documented use — see below | [Codex CLI integration guide](https://www.volcengine.com/docs/82379/2556056), [supported-client list](https://www.volcengine.com/docs/82379/2188957) |
| Named maintenance owner | **still open** — see the ask at the end |
| Citable verification date | satisfied | 2026-08-01 (this comment) |

## On routing authorization

@Ingwannu — this is the item your review blocked on, so here is the reasoning rather than an assertion.

Volcengine publishes a **Codex CLI integration guide** for Coding Plan that instructs users to put a plan key into `~/.codex/config.toml`, and its supported-client list names Claude Code, Codex CLI, OpenCode, OpenClaw, Cline, Cursor, Kilo Code, Roo Code, and TRAE. opencodex exists to attach exactly those clients to a provider, so the traffic this preset carries is a documented supported-client's traffic.

The gate's phrase is "resale or routing authorization for **aggregators**", and `contributing.md` defines that as a gateway selling access to third-party models. This is not that: each user supplies their own Ark key, nothing is pooled, and nothing is resold. That distinction is what makes the canonical entry appropriate rather than the inert `free-directory.ts` fallback.

A counter-search across V2EX, 掘金, CSDN, Zhihu, GitHub, Reddit, and HN found no report of a plan key revoked for routing through a proxy. Absence of a reported ban is not authorization, which is why the restriction is now disclosed rather than assumed away.

## The one real defect this review found

Volcengine states plainly that Plan quota is valid **only** inside supported AI coding tools, and that other API use of the key may suspend the subscription or ban the account ([article 37156](https://www.volcengine.com/article/37156), 2026-04-09; `curl`, Postman, and Dify are named exclusions). The presets shipped without telling users that boundary exists, while `tencent-coding-plan` has carried the equivalent warning since it landed.

Commit `688fd77` fixes that:

- both Plan `note` strings now carry the restriction, worded per product so Agent Plan does not quote Coding-Plan-specific policy
- `safeConfigDTO` resolves the note by **destination** instead of by provider name. The dashboard lets a preset be saved under any name, and a renamed row previously kept the credential destination while silently losing the warning about it. New helper `registryEntryForProviderDestination`.
- both Plan entries declare their text-only models via `noVisionModels`, so the vision sidecar cannot advertise image input for models that cannot accept it — matching the Tencent precedent
- the restriction is documented in the provider guide

Earlier I also pushed `glm-5.1` → `glm-5.2`; the old id is absent from Volcengine's current Plan model list, and with `liveModels: false` discovery cannot repair it.

## Verification

Rebased onto current `dev` (this resolved preset counts across five locales to 65/54 counted from the registry, and a `Providers.tsx` conflict between `oauthStatusWithCodex` and the new `t` argument; it also removed a duplicate `responsesPath` field the rebase surfaced).

- `bun run test` — **6484 pass, 4 skip, 0 fail** across 465 files
- `gui`: `bun test tests` — **463 pass, 0 fail**
- `bun run lint:gui`, `bun run privacy:scan`, `bun run build:gui` — clean
- CI on `688fd77` — **8/8 green**

The custom-name regression was proven non-vacuous by ablation: reverting the resolver makes it fail (15 pass / 1 fail), restoring it returns 16/16.

Two CI jobs failed once before passing on rerun. Both were diagnosed rather than blind-retried, and neither file appears in this diff: macOS exited 133 on a Bun segfault immediately after `tests/storage-worker-teardown-isolate.test.ts` (added by #813), and Windows failed a single Kiro OAuth refresh test.

## Ask for @yrooogerg

Please add a **named maintenance owner** line to the PR description — the person who owns this preset's evidence and breakage follow-up, as `MAINTAINERS.md` requires. Everything else in the gate is now satisfied, and I am recording that item as an outstanding follow-up rather than holding a green, evidence-backed preset on a line of prose. If Volcengine changes the Plan terms, the `note` strings and the provider guide are the two places that must be updated.

Issue #432, opened 2026-07-25
fix(windows): startup safety falsely marks healthy scheduler service stale


0disoft:
## Client or integration

OpenCodex dashboard

## Area

Service lifecycle

## Summary

On Windows, a successfully installed and functioning Task Scheduler service is reported as stale by both `ocx status` and the Dashboard's Startup Safety page.

The generated task XML explicitly contains enabled/default lifecycle fields, but Windows canonicalizes the registered task on export and omits default-valued elements. `windowsTaskRegistrationHealthy()` currently requires those elements to be present literally, so a healthy task is rejected.

Expected: an enabled, logon-triggered scheduler task using the expected `wscript.exe` launcher is considered viable even when Windows omits XML elements whose effective values are the expected defaults.

Actual: `Restart safety: AT RISK after restart (background service files are stale...)` remains after a successful elevated reinstall.

## Reproduction

1. Install OpenCodex 2.7.39 on Windows 11.
2. In an elevated PowerShell, run:

   ```powershell
   ocx service install
   Start-Sleep 8
   ocx status
   ```

3. Installation reports success:

   ```text
   opencodex service installed + started (auto-starts on login, auto-restarts on crash).
   ```

4. Status still reports:

   ```text
   Restart safety: AT RISK after restart (background service files are stale; run 'ocx service install')
   Service: installed, stale or missing service assets
   ```

5. The generated `opencodex-service-task.xml` contains:

   ```xml
   <LogonTrigger>
     <Enabled>true</Enabled>
   </LogonTrigger>
   <Principal id="Author">
     <LogonType>InteractiveToken</LogonType>
     <RunLevel>LeastPrivilege</RunLevel>
   </Principal>
   ...
   <Enabled>true</Enabled>
   ```

6. The registered task exported by `schtasks.exe /query /tn opencodex-proxy /xml` is canonicalized to this equivalent form:

   ```xml
   <LogonTrigger />
   <Principal id="Author">
     <UserId>[REDACTED SID]</UserId>
     <LogonType>InteractiveToken</LogonType>
   </Principal>
   <Settings>
     ...
     <MultipleInstancesPolicy>IgnoreNew</MultipleInstancesPolicy>
     <ExecutionTimeLimit>PT0S</ExecutionTimeLimit>
     ...
   </Settings>
   ```

   The default-valued `Enabled=true` and `RunLevel=LeastPrivilege` elements are omitted.

7. Independent task/process checks show the installation is functional:

[...]

PR #641, opened 2026-07-28
fix(service): compare Task Scheduler Arguments after XML decode

### LINKED ISSUE #608: [Bug] Windows: healthy scheduler task permanently reported stale — <Arguments> compared escaped vs. canonicalized (distinct root cause from #432)
## Client or integration

`ocx` CLI (Windows, Task Scheduler backend)

## Area

Service lifecycle

## Summary

`windowsTaskRegistrationHealthy()` compares the registered task XML against an **XML-escaped** `<Arguments>` string, but Windows Task Scheduler exports that element **unescaped**. The two can never be equal, so a healthy scheduler task is permanently reported as stale.

This produces the exact same user-facing symptom as the already-fixed #432, which makes it easy to mistake for a regression or duplicate — but the root cause is different and #432's fix does not cover it:

- **#432** — Windows *omits* default-valued elements (`<Enabled>`, `<RunLevel>`) on export. Fixed by treating absence as the schema default (`taskXmlOptionalValueEquals`).
- **This issue** — the element is present, but its *text is canonicalized*: `&quot;` becomes a literal `"`. The equality check is a raw substring match, so it fails regardless of the #432 fix.

Consequence: `stale` is latched true → `viable` and `startable` are both false → `ocx` stops treating the scheduler backend as usable and falls back to launching the proxy directly. **Re-running `ocx service install` elevated cannot clear it**, because a freshly registered task exports the same way.

## Root cause (code level, v2.7.42)

`src/service.ts`:

1. `taskXmlString()` escapes `"` → `&quot;`:

   ```ts
   function taskXmlString(value: string): string {
     return value
       .replace(/&/g, "&amp;")
       .replace(/</g, "&lt;")
       .replace(/>/g, "&gt;")
       .replace(/"/g, "&quot;")     // <-- here
       .replace(/'/g, "&apos;");
   }
   ```

2. The generated task XML therefore contains `<Arguments>/b /nologo &quot;...vbs&quot;</Arguments>`.

3. `windowsTaskRegistrationHealthy()` requires the **registered** XML to contain that same escaped form:

   ```ts
   && action.includes(`<Arguments>${taskXmlString(`/b /nologo "${launcher}"`)}</Arguments>`);
   ```

4. But `"` is a legal literal character inside XML element content, so Windows canonicalizes it on export. `schtasks /query /xml` returns:

   ```xml
   <Arguments>/b /nologo "C:\path\to\opencodex-service-launcher.vbs"</Arguments>
   ```

Since the registered XML never contains `&quot;`, the `includes()` check is unconditionally false whenever the launcher path is quoted — which it always is.

## Reproduction

Windows 11 Pro (26200), `@bitkyc08/opencodex` **2.7.42**, Bun 1.3.14, Task Scheduler backend, task installed and running normally.

1. Confirm the proxy is actually healthy:

   ```powershell
   curl.exe -s http://127.0.0.1:10100/healthz
   # {"status":"ok","service":"opencodex","version":"2.7.42",...}
   ```

2. Confirm the scheduled task exists and is enabled:

   ```powershell
   Get-ScheduledTask -TaskName opencodex-proxy | Select-Object TaskName,State
   # opencodex-proxy   Ready
   ```

3. Show that the registered XML contains no escaped quote:

   ```powershell
   $raw = & "$env:SystemRoot\System32\schtasks.exe" /query /tn opencodex-proxy /xml ONE | Out-String
   $raw.Contains('&quot;')            # False
   $raw -match '<Arguments>[^<]*\x22' # True  (literal " present)
   ```

4. Observe the false positive:

   ```powershell
   ocx status
   # Service: installed, stale or missing service assets — run 'ocx service install' to repair
   ```

Note that `bakedServicePathsDiagnostic()` returns null here (both `bunPath` and `cliPath` exist, and the diagnostics string shows only `logs: ...` with no `STALE baked paths` prefix), and `recordedBackend` is `scheduler`, so `staleBakedPaths` and `backendStateMismatch` are both false — the `stale` flag can only be coming from `schedulerInstalled && !schedulerAssetsHealthy`, i.e. `registrationHealthy === false`.

## Expected

A task whose registered `<Arguments>` is semantically identical to the generated one (differing only in XML escaping of characters that do not require escaping) is considered healthy.

## Actual

`registrationHealthy` is false → `stale` is true → `viable`/`startable` false → scheduler backend treated as unusable; `ocx update` also attempts a service reinstall each time and reports the service as needing repair.

## Suggested fix

Compare **decoded** text rather than raw substrings — unescape both sides (or at minimum normalize `&quot;`/`&apos;`/`&amp;`) before the `includes()` check, in the same spirit as #432's "canonicalization-tolerant comparison". A regression test asserting that a task XML round-tripped through `schtasks` (literal `"`) still validates would prevent this class from recurring.

For context, the same canonicalization class has now bitten this function twice — once by element omission (#432), once by text escaping (this issue). It may be worth comparing against a parsed representation instead of doing string containment on raw XML.

## Workaround

None that clears the flag. The proxy itself is unaffected and the logon trigger still works; only the health reporting and the scheduler-backend viability logic are wrong. Users can ignore the warning, but `ocx` will keep falling back to direct proxy launch instead of using the installed service.

Wibias:
## Summary
- Compare `<Arguments>`/`<Command>` by decoding XML entities once, matching Windows Task Scheduler's unescaped export (distinct from #432).
- Keeps nested-markup / multi-element rejection.

## Test plan
- [x] `bun test tests/service.test.ts -t '#608'`
- [ ] CI green

Closes #608

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -1050,6 +1050,29 @@ function taskXmlOptionalValueEquals(xml: string, tag: string, expected: string):
   return value?.trim().toLowerCase() === expected.toLowerCase();
 }
 
+function taskXmlDecodeEntities(value: string): string {
+  return value
+    .replace(/&quot;/g, "\"")
+    .replace(/&apos;/g, "'")
+    .replace(/&lt;/g, "<")
+    .replace(/&gt;/g, ">")
+    .replace(/&amp;/g, "&");
+}
+
+/**
+ * Exact element-text match after one-pass decoding of XML's five predefined entities.
+ * Task Scheduler exports element text unescaped (`"`), while we emit `&quot;` (#608).
+ * Rejects zero/multiple occurrences and nested markup (no `<` inside the value).
+ */
+function taskXmlElementTextEquals(xml: string, tag: string, expectedDecoded: string): boolean {
+  if (taskXmlHasPrefixedTag(xml, tag)) return false;
+  const count = taskXmlElementCount(xml, tag);
+  if (count !== 1) return false;
+  const raw = new RegExp(`<${tag}(?:\\s[^>]*?)?>\\s*([^<]*?)\\s*<\\/${tag}>`, "i").exec(xml)?.[1];
+  if (raw === undefined || raw.includes("<")) return false;
+  return taskXmlDecodeEntities(raw) === expectedDecoded;
+}
+
 /** Validate the security/lifecycle-critical fields of the registered scheduler task. */
 export function windowsTaskRegistrationHealthy(
   xml: string,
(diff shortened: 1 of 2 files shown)


Wibias:
@codex review
@coderabbitai review

Wibias:
@codex: addressed in 9be8ad1b — compare trimmed decoded Arguments/Command text.

PR #644, opened 2026-07-28
fix(windows): follow active Codex home for tray listener

clover980805-creator:
## Summary by CodeRabbit

* **New Features**
  * Updated the pull-request labeling workflow so labels can be applied successfully while retaining read-only repository contents access.
  * Improved Windows tray launching to use the configured Codex home directory.

* **Bug Fixes**
  * Corrected tray home-directory resolution on Windows.

* **Chores**
  * Added repository hygiene checks and expanded ignore rules for local artifacts and OS metadata.
  * Removed obsolete internal planning records.
  * Updated the application version to 2.7.42.


Wibias:
Hey @clover980805-creator thanks for this PR. Please adjust your description.

lidge-jun:
**NEEDS-CHANGES** — two separate problems, and the second one is the blocker.

**The target is not established.** The PR body uses `E:\codex` as the motivating example, but the code being replaced already honoured an explicit `CODEX_HOME` and would have resolved that path fine. Where `resolveCodexHomeDir()` actually differs is when no explicit variable is set: it can discover a Windows Codex home from WSL when the local `~/.codex/config.toml` is absent (`src/codex/home.ts:121-146`). That may well be a real inconsistency worth fixing, but it is not the scenario the description gives, and there is no linked issue or reproduction, so I could not determine which failure this binds.

**An unrelated hunk would regress current behaviour.** `src/service.ts:1073-1085` lowercases encoded XML fragments and compares with `includes`. Current `dev` decodes the values first and compares them case-insensitively via `taskXmlDecodedValueEquals` (`src/service.ts:1133`, used at `:1184-1185`), which exists specifically to handle the `&quot;` normalization from #608. Merging the PR side of that hunk would undo it. This is also the source of the conflict — the `currentCodexHome()` substitutions themselves merge cleanly.

The other conflict is trivial: the import hunk in `src/tray/windows.ts`, where `dev` adds `recordOwnedConfigPath` and the PR adds `resolveCodexHomeDir`. Keep both.

**No tests.** Nothing in the diff proves that tray and service state now follow the shared resolver, and nothing covers the scheduler comparison this PR introduces. Both changes touch service ownership and validation of a persisted Task Scheduler command, which are lifecycle and command-execution boundaries under `MAINTAINERS.md:33-34`.

What happens next, in order:

1. Drop the scheduler hunk — it belongs in its own PR if you want it, and as written it conflicts with a deliberate fix.
2. State the actual reproduction, with no `CODEX_HOME` set, so the change has something to bind.
3. Add a regression that distinguishes the old resolver from `resolveCodexHomeDir()` — one that fails before your change and passes after.
4. Rebase onto current `dev`, keeping both imports.

`dev` CI is fully green now, so your checks will be about your change.

lidge-jun (changes requested):

Blocking: the `src/service.ts:1073-1085` scheduler hunk regresses the decoded-XML comparison that `dev` added for #608, and it is unrelated to the Codex-home change this PR is about. Drop it. Also, the target is not established — the old code already honoured an explicit `CODEX_HOME`, so please state the actual reproduction (no `CODEX_HOME` set) and add a regression that fails before the change.

Full reasoning in my earlier comment on this PR.

lidge-jun:
Closing this after maintainer re-review. The reproduction described in the thread does not hold on current `dev`, the PR carries no regression coverage for the tray-listener path, and it includes an unrelated scheduler change that regresses behavior that has since landed on `dev`. If the underlying issue still reproduces for you, a fresh, minimal PR with a regression test is welcome. Thanks for the investigation work here.

PR #646, opened 2026-07-28
fix(cursor): add kimi-k3 with low/high/max effort tiers

koopmannleon19977-cmyk:
## Summary

Cursor serves **Kimi K3** only as effort-suffixed wire ids (`kimi-k3-low`, `kimi-k3-high`, `kimi-k3-max`) — verified against `GetUsableModels` on 2026-07-28 with an entitled account. The bare id `kimi-k3` fails with Connect `not_found`, and since the model was missing from `CURSOR_STATIC_MODELS` it never entered the routed catalog either.

- `src/adapters/cursor/effort-map.ts`: add `"kimi-k3": ["low", "high", "max"]` to `CURSOR_MODEL_EFFORT_TIERS`, so requests resolve to a suffixed variant (bare/no-effort picks the top tier `max`, same clamp behavior as other tiered models). Tiers mirror the native Kimi provider's K3 ladder (`KIMI_CODING_K3_REASONING_EFFORTS`).
- `src/adapters/cursor/discovery.ts`: add `kimi-k3` to `CURSOR_STATIC_MODELS` (`CONTEXT_262K`, `supportsReasoningEffort: true`) and adjust the stale "kimi sent bare" comment.
- Cursor's own docs page now lists the model: https://cursor.com/docs/models/kimi-k3

## Test plan

- [x] `bun test tests/cursor-effort-suffix.test.ts tests/cursor-discovery.test.ts` — 18 pass (new cases: `cursor/kimi-k3` + low/medium/high/max/bare → suffixed wire ids, ladder `["low","high","max"]`, static seed contains `kimi-k3`)
- [x] `bun test tests/codex-catalog.test.ts tests/provider-registry-parity.test.ts tests/reasoning-effort.test.ts` — 182 pass
- [x] Live smoke against a 2.7.41 install patched with this change: `POST /v1/responses` `cursor/kimi-k3` → `completed` (previously `Cursor Connect error not_found`); same with `reasoning.effort=low`

## Summary by CodeRabbit

- **New Features**
  - Added support for the Cursor Kimi K3 model (262K context window).
  - Added reasoning effort tier handling (low/medium/high/max) with correct effort-suffix routing.
  - When no effort level is specified, model selection now defaults to the highest available effort (max).
- **Bug Fixes**
  - Adjusted Cursor provider defaults to ensure Kimi K3 routes to the expected max tier.
- **Documentation**
  - Refreshed Cursor GPT-5.6 preview path notes, including fallback and tier behavior.
- **Tests**
  - Added/updated regression coverage for Kimi K3 discovery, effort ladders, and catalog defaults.


Wibias (changes requested):

**Verdict: REQUEST CHANGES**

HEAD: `31c73265`. Target `dev` is correct. CodeRabbit: no actionable comments. Codex: 2 open P2s (verified below). Security review ([Security Review](5ee89ab3-8f2e-40ca-be42-fbed183323f2)): **no medium+** — static catalog/effort-tier mapping only; no auth, credential, SSRF, or trust-boundary changes.

Focused effort/discovery tests: discovery suite passes; effort-suffix file needs `bun install` in the review worktree (`@bufbuild/protobuf`) — rely on CI for the full gate. Cross-platform CI had not fully reported at review time beyond label/enforce/CodeRabbit.

### Open Codex (verified against HEAD)

| Item | Verdict | Action |
| --- | --- | --- |
| **P2: catalog default becomes `high`, not `max`** | **Valid / required** | Cursor registry has `modelReasoningEfforts` from the ladder `["low","high","max"]` but no `modelDefaultReasoningEfforts`. `applyReasoningLevels` (`effort.ts:178-180`) then picks `high` (no `medium` in the ladder). Ordinary Codex picker requests therefore send `high` → wire `kimi-k3-high`, and never hit the request-builder’s no-effort → top-tier (`kimi-k3-max`) path that the new test documents. Native Kimi / OpenCode-Go already set `"kimi-k3": "max"`. Add `modelDefaultReasoningEfforts: { "kimi-k3": "max" }` on the Cursor registry entry (`providers/registry.ts`) and cover it in catalog/reasoning tests. |
| **P2: docs-site Cursor coverage** | **Valid / required** | User-facing catalog change. Update English `docs-site/.../guides/codex-app-models.md` Cursor row (and keep translated locales consistent) so the static fallback mentions `cursor/kimi-k3` and its `low` / `high` / `max` effort behavior. |

### Diff sanity (no extra blockers)

- Tiers `["low","high","max"]` match `KIMI_CODING_K3_REASONING_EFFORTS`; `supportsReasoningEffort: true` matches the effort-map entry.
- `medium` → `kimi-k3-high` clamp matches existing non-medium ladder behavior (`cursorEffortSuffix`).
- Static seed + effort-suffix tests are the right shape once the default override lands.

### Required before merge

1. Set Cursor `modelDefaultReasoningEfforts["kimi-k3"] = "max"` (+ regression coverage).
2. Update docs-site Cursor coverage (EN + locales).

lidge-jun:
Reviewed against the current `dev` tip. The tier data is correct and I verified it independently — closing the loop on what is still needed before this can land.

## Confirmed correct

Cursor's own CursorBench page lists exactly `Kimi K3 Low`, `Kimi K3 High`, and `Kimi K3 Max`, so `["low", "high", "max"]` matches first-party evidence, and there is no `medium` tier to add. Moonshot's official Kimi K3 model card documents the same three `reasoning_effort` values. Your tests pass locally for me too: `tests/cursor-effort-suffix.test.ts` + `tests/cursor-discovery.test.ts` 18 pass, and `tests/codex-catalog.test.ts` + `tests/provider-registry-parity.test.ts` + `tests/reasoning-effort.test.ts` 182 pass, with `bun x tsc --noEmit` clean.

## Blocker: the catalog default lands on `high`, not `max`

The Codex-review P2 about this is real, and I reproduced it rather than taking it on faith. Against your branch:

```
cursorModelEffortLadder("kimi-k3")            -> ["low","high","max"]
cursor registry modelDefaultReasoningEfforts  -> null
default_reasoning_level would be              -> high
```

`applyReasoningLevels` picks `medium` if present, otherwise `high`, otherwise the first rung. With no `modelDefaultReasoningEfforts` on the `cursor` entry, `high` wins and an ordinary picker request sends `high` explicitly — so the request builder's no-effort fallback to `kimi-k3-max` is never reached.

That also breaks consistency with the other K3 routes. `src/providers/registry.ts` already carries `modelDefaultReasoningEfforts: { "kimi-k3": "max" }` on the `opencode-go` entry, and Moonshot documents `max` as the model's default. Cursor should not be the one route that quietly downgrades it.

The fix is one line on the `cursor` registry entry, next to the existing `modelReasoningEfforts: cursorModelReasoningEfforts(...)`:

```ts
modelDefaultReasoningEfforts: { "kimi-k3": "max" },
```

Please also add a catalog-level assertion so this cannot regress — the existing effort-suffix tests pass without it, which is exactly why the P2 slipped through.

## Second item: docs

`docs-site/src/content/docs/guides/providers.md` describes the Cursor fallback seed as `gpt-5.6-sol/terra/luna` plus `grok-4.5`/`grok-4.5-fast`, with no mention of Kimi K3 or its effort ladder. Per `AGENTS.md`, a user-visible catalog change needs the English docs updated, and translated locales should not end up contradicting it.

## To land

1. Add the `max` default above, with catalog coverage.
2. Update the Cursor coverage docs.
3. Mark the PR ready for review (it is still a draft), which also lets CodeRabbit run a real pass instead of skipping.

Thanks for the live `GetUsableModels` verification and the smoke test against a patched install — that is the part that is hard to check from the outside, and it is what makes the tier list trustworthy. Once the default is pinned this is a straightforward merge.

lidge-jun:
Maintainer follow-up: I pushed the two requested changes to this branch as `d2201509b` and marked the PR ready.

**Why the tier data was right but the PR was incomplete.** Your `["low","high","max"]` ladder is correct — Cursor serves K3 only as effort-suffixed wire ids, and Kimi documents `max` as the model's API default. The gap was that the Cursor registry entry supplied no `modelDefaultReasoningEfforts`. Because that ladder has no `medium` rung, `applyReasoningLevels` fell through its `medium -> high -> first` preference and settled the catalog default on `high`. The picker then sends `high` explicitly, so the request builder's no-effort fallback to `kimi-k3-max` was never reached — the behaviour your PR description describes as the intent.

Reproduced both directions on your commit:

```
31c732653 alone            -> row.default_reasoning_level = "high"   (ladder low/high/max, default undefined)
+ modelDefaultReasoningEfforts -> row.default_reasoning_level = "max"
line removed again        -> new test fails with Received: "high"
```

This is also why the existing `cursor-effort-suffix` tests passed without it: they assert the ladder, not the assembled catalog default. The new regression case covers the assembled row instead.

`cursor` was the only K3 route missing this — `kimi`, `kimi-code`, and `opencode-go` already carry it.

**What landed in `d2201509b`:**

1. `src/providers/registry.ts` — `modelDefaultReasoningEfforts: { "kimi-k3": "max" }` on the `cursor` entry, with a comment recording the medium-less-ladder reasoning.
2. `tests/codex-catalog.test.ts` — a regression case asserting the assembled row's ladder `[low, high, max, ultra]` and `default_reasoning_level === "max"`, following the neighbouring Google Gemini block's `enrichProviderFromRegistry` + `gatherRoutedModels` + `buildCatalogEntries` pattern.
3. `docs-site` — the Cursor adapter paragraph now lists `kimi-k3` (262K) in the fallback seed and states the ladder defaults to `max`, in the English source plus the ko/ja/ru/zh-cn locales. The separate GPT-5.6 coverage table is deliberately untouched, since inserting a non-5.6 model there would break what that table means.

Docs shipped in the same commit as the registry line on purpose: the prose asserts the ladder and the default, so splitting them would make the docs false at the intermediate commit.

**Scope check.** `modelRecordValue` matches exact key, then colon-family, then case-fold, so `{ "kimi-k3": "max" }` cannot reach `kimi-k3-max`, `kimi-k3-low`, or any other model. Verified across 8 Cursor model ids — only `kimi-k3` is affected; `glm-5.2`, `gpt-5.6-sol`, `grok-4.5`, `kimi-k2.7-code`, and `claude-opus-5" all stay untouched. Field completeness for `cursor/kimi-k3` is also confirmed from your seed: context 262,144, text+image modalities with vision-sidecar coverage, so no further registry fields are needed.

**Gates** (local, on the pushed head): `tsc --noEmit` clean; `codex-catalog` 113 pass; `cursor-effort-suffix` + `cursor-discovery` 18 pass; `provider-registry-parity` + `reasoning-effort` 70 pass; `privacy:scan` pass; `lint:gui` pass. Waiting on cross-platform CI before merge.

Thanks for the account-verified `GetUsableModels` evidence and the live smoke — that made the tier data easy to confirm.

lidge-jun (approved):
Approving. Both requested changes landed in `d2201509b`, and the branch is now merged up to green `dev` as `60bd535f6`.

Your tier data was right from the start — Cursor serves K3 only as effort-suffixed wire ids and Kimi documents `max` as the API default. What was missing was the registry default: with no `medium` rung in the ladder, `applyReasoningLevels` fell through `medium -> high` and settled the catalog on `high`, so the picker sent `high` explicitly and never reached the `kimi-k3-max` fallback your PR description intended. Verified on your commit alone (`high`), with the fix (`max`), and by deleting the line again (new test fails with `Received: \"high\"`).

CI history worth noting so the reruns do not look mysterious: the first `windows-latest` failure was a missing timeout-headroom fix (#717) that this branch's base predated, the second was the `dev` kiro regression now fixed as #718, and the third was a Bun runtime `panic(thread …): Internal assertion failure` that cleared on rerun. All six jobs are green on `60bd535f6`.

Thanks for the account-verified `GetUsableModels` evidence and the live smoke — that made the tier ladder easy to confirm.

PR #687, opened 2026-07-29
Add WorkBuddy desktop session provider

xmanrui:
## What

- Add a built-in `workbuddy` provider backed by WorkBuddy Desktop's signed-in macOS session.
- Import the current access token, account id, domain, and expiry without copying WorkBuddy's refresh token.
- Re-import desktop token rotations before requests and replay one upstream `401` with the refreshed session.
- Inject WorkBuddy's required client headers only after routing has selected the canonical WorkBuddy provider.
- Seed the model and capability catalog observed from WorkBuddy 5.2.6's ACP `session/new` response.
- Expose the provider in the dashboard, classify it as an elevated unofficial bridge, and document setup in every maintained docs locale.

## Why

WorkBuddy subscriptions expose coding models through the desktop client, but OpenCodex had no provider-specific session import or request profile. Treating the endpoint as a generic OpenAI-compatible provider fails because the upstream requires both the desktop bearer token and WorkBuddy client headers.

## Security and behavior

- The canonical upstream remains pinned to `https://copilot.tencent.com/v2`.
- Header values are length-bounded and reject CR/LF injection.
- The desktop refresh token is never parsed or persisted; OpenCodex stores an external-session sentinel instead.
- WorkBuddy is single-slot because the current desktop login is the credential authority.
- Request headers are derived from the same credential generation as the bearer token, and contain no authorization value.
- Existing OpenAI/ChatGPT routing is unchanged; WorkBuddy headers are added only when `route.providerName === "workbuddy"`.
- This touches credential handling and should receive the repository's required security review.

## Verification

- `bun run typecheck`
- `bun run privacy:scan`
- `bun test tests/workbuddy-provider.test.ts tests/oauth-store-multi.test.ts tests/provider-registry-parity.test.ts tests/oauth-public-surface.test.ts tests/oauth-tos-warning.test.ts` — 75 passed
- `cd gui && bun run lint && bun run build`
- Manual live validation with WorkBuddy 5.2.6: streaming text, a required tool call, explicit `glm-5.2`, reasoning effort, and a bundled Codex CLI turn all completed successfully.

The full local suite was also attempted. WorkBuddy coverage passed, but the machine's Clash fake-IP DNS mapped fixture domains to `198.18.0.0/15`, causing unrelated destination-policy failures; the run reached 5,849 passing tests before exiting nonzero. CI should provide the clean-network result.


lidge-jun:
**CLOSE-RECOMMENDED** — and I want to be straight about why, because it is not a code judgment.

The implementation is careful. Not importing the desktop refresh token and storing a sentinel instead (`src/oauth/workbuddy.ts:102-116`) is the right instinct. Re-reading the session file on rotation rather than caching indefinitely is right. Rejecting CR/LF in parsed session values, aborting cleanly, replaying exactly one 401 after re-import — these are the decisions of someone who thought about what happens when the other side changes underneath you. `tests/workbuddy-provider.test.ts` binds all of it, including the full 401 refresh-and-replay path with a changed Bearer token.

Two things make this different from the other preset PRs in the queue.

**It is not an API-key preset.** It declares `authKind: "oauth"` but there is no authorization flow. It reads WorkBuddy Desktop's macOS session file for an access token, user ID, domain and expiry, then synthesizes desktop-client identity headers — `X-Product: workbuddy-desktop`, user and domain IDs, conversation IDs (`src/oauth/workbuddy.ts:35-99`, `:138-158`). That means OpenCodex would be persisting a consumer subscription credential, presenting itself as another vendor's desktop client, and coupling itself to a private session-file format that can change in any WorkBuddy release. The "elevated unofficial" label warns the user, which is honest, but a warning is not authorization.

**WorkBuddy's Acceptable Use Policy appears to prohibit it.** Their current AUP requires express prior written permission before developing external components or interoperability technology for the service. This PR is exactly that, and no permission is attached. Their terms also identify the operating entities — Tencent Korea Yuhan Hoesa in South Korea, Tencent Cloud International Pte. Ltd. elsewhere — so this is not an ambiguous or unattributed service.

`MAINTAINERS.md:33-42` would separately require documented endpoints, ToS and operating entity, routing authorization (their own terms say the service is powered by third-party LLMs, and the eleven hardcoded aliases confirm it), a named maintenance owner, and a citable verification date. None of those are supplied. But even a complete package would not answer the AUP question.

I am not closing this — that is the owner's call, and I would rather leave the reasoning here than make the decision quietly. My recommendation is to close it unless you can produce written permission from Tencent, in which case it comes back with a dedicated security review rather than the ordinary preset path.

If you want the work to land somewhere, the honest version of it is a local integration users opt into themselves, not a preset the project ships and maintains. Sorry to land on this after the effort you clearly put in — the code deserved a better answer than the terms allow.

lidge-jun:
Closing on the licensing finding, not on the code.

To restate it so the record is clear: WorkBuddy's Acceptable Use Policy requires express prior written permission before developing external components or interoperability technology for the service. This bridge is exactly that, and no permission is attached. Their terms also identify the operating entities — Tencent Korea Yuhan Hoesa in South Korea, Tencent Cloud International Pte. Ltd. elsewhere — so this is not an ambiguous or unattributed service where we could reasonably proceed and see.

Shipping it would mean the project takes on maintenance of an integration we are not authorized to build, against another app's private session-file format that can change in any release. That is not a burden this repository can carry.

None of that is a criticism of the work. Not importing the desktop refresh token and persisting a sentinel instead, re-reading on rotation rather than caching, rejecting CR/LF in parsed session values, replaying exactly one 401 — those are the decisions of someone who thought carefully about failure modes, and the test coverage matches.

Reopening is straightforward if the situation changes: written permission from Tencent, plus the canonical-preset evidence package from `MAINTAINERS.md:33-42`, and it comes back through a dedicated security review rather than the ordinary preset path.

If you want the capability without the licensing question, the honest shape is a local integration users opt into themselves rather than a preset the project ships. Sorry to land here after the effort you put in.

PR #706, opened 2026-07-29
[WRONG BRANCH] fix(minimax): refresh MiniMax-M3 metadata with video input and current pricing

octo-patch:
Reason: MiniMax-M3 metadata was stale — it omitted video input and kept the old $0.30/$1.20 per-million input/output pricing instead of $0.60/$2.40 with $0.12 cache reads.

## Changes

- `src/generated/jawcode-model-metadata.ts`: refresh the `minimax` bundle's `MiniMax-M3` row — add `video` to its input modalities and update pricing to `$0.60` input / `$2.40` output per million tokens with `$0.12` cache reads. The 1,000,000-token context window is preserved.
- `scripts/generate-jawcode-metadata.ts`: widen the model input-modality type, and the interface/cast it emits, from `("text" | "image")[]` to `("text" | "image" | "video")[]` so the generator can carry video input and the generated metadata stays consistent with its source.

## Checks

- `bun x tsc --noEmit` — passes.
- `bun test tests/codex-catalog.test.ts tests/provider-registry-parity.test.ts tests/minimax-reasoning-split.test.ts tests/usage-cost.test.ts` — 187 pass, 0 fail.

## Summary by CodeRabbit

* **New Features**
  * Added support for video inputs in Jawcode model metadata.
  * Generated metadata now correctly represents models that accept text, image, or video inputs.


lidge-jun:
Thanks for this — the video finding is correct and I've applied it, but not in this shape. Closing with the reasoning, since two of the three premises don't hold.

**Video input: right, and now landed upstream.** MiniMax documents video for M3 and only M3 — the M2 series takes text and tool blocks but not image or video. Fixed at the source in jawcode as `762549f`, and the generator type widening you proposed is adopted as `d2e4cc984`.

One correction worth recording: the grounding is the OpenAI-compatible endpoint, not the Anthropic one. Our `minimax` preset is `adapter: "openai-chat"` with `baseUrl: "https://api.minimax.io/v1"` (`src/providers/registry.ts:1001`; `minimax-cn` at `:1013`). The `anthropic-messages` + `/anthropic` pair in the metadata row is jawcode's own field, not a path opencodex calls. On our path video arrives as a `video_url` content part; the Anthropic path uses a `{"type":"video","source":…}` block. Same conclusion, different citation.

**Pricing: the premise is wrong, so this part is not adopted.** $0.30/$1.20 isn't stale — it's MiniMax's standing 50% discounted rate for the ≤512K input tier. $0.60/$2.40 is that tier's *list* price and simultaneously the >512K rate. Both are official numbers, so choosing which one a catalog carries is a policy decision rather than a staleness fix.

It also happens to be the wrong direction here: our row flattens a 1,000,000-token context into a single entry, so adopting list price would overstate the bill for most real requests. Cache-read $0.06 vs $0.12 splits the same way. If we ever want list pricing, it needs a tiered-cost model first, and that's a separate discussion.

**Editing the generated file gets reverted.** `src/generated/jawcode-model-metadata.ts` carries `// Generated by scripts/generate-jawcode-metadata.ts. Do not edit by hand.` — the next `bun run generate:jawcode-metadata` would drop a hand-edited row. That's why the fix went into the jawcode source instead.

There's a wrinkle behind that worth flagging: the committed generated file is **already 95 models and 63 price rows out of sync** with its source, plus a `PROVIDER_ALIASES` ordering drift. Nothing enforces sync — no test regenerates and compares — so it silently drifted. A regeneration today would therefore sweep 63 unrelated price changes into cost accounting alongside your one modality fix, including halving `opencode-go/minimax-m3`'s advertised context from 1,000,000 to 512,000. So the resync is being handled as its own reviewed unit with a drift guard, and the M3 video row lands there as a clean one-row delta. Verified the combination produces exactly that:

```
["MiniMax-M3",1000000,128000,"text,image,video",1,null,0.3,1.2,0.06,0]
                              ^^^^^^^^^^^^^^^^^  video           ^^^^^^^^^ prices unchanged
```

**Branch target.** This PR targets `main`, which `enforce-target` already fails — per AGENTS.md every PR goes to `dev`, with `main` moving only by maintainer promotion. Worth knowing for next time.

One honest limitation on the feature itself: advertising `video` only changes what the catalog claims. No code in `src/` emits a `video_url` part, and `OcxContentPart` is `text | image` only, so a video attachment is silently dropped rather than sent. The wire path is tracked separately — this change just stops the catalog from understating what the model supports.

Closing as superseded by the upstream fix. The video catch was a good one.

PR #707, opened 2026-07-29
security: harden post-merge service and management boundaries

LeoWang331:
## Summary

Follow-up hardening for #697 based on review findings and regression testing after that PR merged.

- makes service data/admin secret installation transactional and fail-closed when secret hardening fails
- hardens ownership-manifest publication and recovery against replacement races
- tightens management model probes so unsafe, unavailable, and pass-through targets fail closed
- preserves proxy routing semantics without silently bypassing configured proxy boundaries
- supports management model discovery and data-plane HTTP/WebSocket traffic through the development proxy
- updates user-facing documentation across supported locales

This PR intentionally does not broaden the remaining deferred audit scope.

## Why

Post-merge review found several edge cases where service repair, ownership-lock recovery, proxy environment selection, and management model probes could degrade insecurely or fail in ways that were hard to diagnose. The changes keep those boundaries explicit and add regression coverage for the failure paths.

## Impact

- Windows service installation and repair validate and preserve both service credentials as one transaction.
- Management APIs remain unavailable when their protected credential cannot be created or hardened.
- Provider management probes reject unsafe or unavailable routing graphs before making upstream requests.
- Proxy-only environments do not silently fall back to direct connections.
- The development dashboard can proxy data-plane HTTP and WebSocket requests consistently.

## Verification

Executed on Windows against the final branch state:

- `bun run test`: 6092 passed, 4 skipped, 0 failed across 431 files; exit 0; 1006.9s
- GUI tests: 386 passed, 0 failed; exit 0
- focused management model tests: 14 passed, 0 failed; exit 0
- `bun run typecheck`: exit 0
- `bun run privacy:scan`: exit 0
- GUI lint and production build: exit 0
- documentation build: 146 pages; exit 0
- `git diff origin/dev...HEAD --check`: exit 0

Existing non-blocking warnings remain: large production chunks, React test act warnings, and the documentation entry-to-404 notice.

## Summary by CodeRabbit

* **New Features**
  * Added management API model discovery and per-model testing in the API Keys workspace.
  * Model tests now report clearer success and failure details, with unsupported models clearly identified.
  * Added the `ocx uninstall` command.

* **Security & Reliability**
  * Separated data-plane and management authentication for remote access.
  * Improved safe cleanup, preserving unowned files and reporting leftovers for review.
  * Strengthened outbound proxy and provider destination safeguards.
  * Dashboard embedding in iframes is now blocked to prevent clickjacking.

* **Documentation**
  * Updated remote access, uninstall, proxy safety, and dashboard guidance across supported languages.


lidge-jun:
**NEEDS-SECURITY-REVIEW, and splitting is a practical prerequisite**

The work here is good, and that is exactly why I do not want it reviewed as one 7,252-line unit. Requiring management token files to be regular non-symlink files and failing closed when hardening or ownership registration fails (`src/server/management-auth.ts:74-100`, `:175-211`); making service token installation a two-file transaction with collision checks, snapshots and rollback (`src/service.ts:500-573`); locking and identity-checking ownership publication and refusing uncertain state on uninstall (`src/lib/config-ownership.ts:500-620`); resolving npm only from absolute safe PATH entries and stripping admin/data credential variables from the child environment (`src/update/npm-invocation.mjs:44-76`) — each of these closes a real hole, and each carries adversarial regressions rather than happy-path ones.

But those are five different threat models, plus a sixth thing that is not a threat model at all. A reviewer who signs off on the npm PATH hardening has said nothing about the ownership-manifest race, and vice versa. Bundling them does not create atomic safety; it means a single approval implicitly covers work the approver may not have read closely. `MAINTAINERS.md:44-45` also asks for both maintainers on security-sensitive changes where practical, which is harder to arrange for a diff this size than for five focused ones.

The diff already splits along clean seams — each has its own implementation and test files:

1. service data/admin token transaction + management-auth fail-closed (`tests/service-admin-token.test.ts`, `tests/server-management-auth.test.ts`)
2. ownership manifest locking, publication, recovery, uninstall (`tests/config-ownership-uninstall.test.ts`)
3. management model route gating + outbound proxy policy (`tests/model-test-management-api.test.ts`, `tests/provider-outbound.test.ts`)
4. npm updater PATH and environment hardening (`tests/update-npm-invocation.test.ts`)
5. browser anti-framing and timing-safe comparison (`src/server/auth-cors.ts:113-117`, `:206-239`)
6. the CI timeout matrix (12 min POSIX / 20 min Windows)

Number 6 should not be in a security PR at all — it is ordinary maintenance and can go in on its own today.

For the record on scope, since "81 files, security" invites worst-case assumptions: this does **not** touch `scripts/release.ts` or `.github/workflows/release.yml`, changes no package or lockfile, and modifies Actions only for the timeout matrix — no permission or action-ref changes. That is a narrower blast radius than the size suggests.

**The conflict is one hunk.** `gui/tests/apikeys-refresh-preserve.test.ts`: current `dev` waits `310ms`, this branch waits `350ms`, the common base waited `0ms`. Both sides are reacting to the same 300ms delete-confirmation arming delay in `ApiKeysWorkspace.tsx`. Please tie the value to that interval rather than taking the larger sleep — an arbitrary margin is how a test becomes slow and flaky at the same time.

What happens next: split along the seams above, land 6 immediately as maintenance, and request explicit security review per remaining piece. If you would rather keep it whole, say so and I will arrange review, but expect it to take considerably longer.

`dev` CI is fully green now, so checks will reflect this branch.

lidge-jun:
Closing in favor of focused follow-ups. The first tranche of this hardening landed with #697. The remaining scope here — service-token, ownership-lock, probe-destination, updater, and browser controls — is individually valuable, but an 8k-line multi-threat-model branch can no longer be reviewed as one unit and now conflicts with `dev`. Please open one focused PR per boundary, each with its own threat model and tests; each can then receive focused security review. Thank you for pushing this forward.

PR #711, opened 2026-07-29
fix(claude): accept data-only Responses SSE (#700)

### LINKED ISSUE #700: [Provider compatibility]: Claude Messages bridge rejects data-only Responses SSE as truncated
### Client or integration

Claude Code

cubapi (`https://xxxxx.cn/v1/responses`), configured with `adapter: openai-responses`

### OpenCodex version

2.7.42

### Endpoint or capability

`POST /v1/messages` translating a streaming Responses API upstream

### Current behaviour

The Claude Messages bridge reports a truncated upstream stream even though the upstream returned a complete Responses SSE stream.

Control tests against the same provider, model, key, and request content:

1. Direct upstream `POST /v1/responses` with `stream:true`: HTTP 200, includes `response.completed` and `data: [DONE]`.
2. OpenCodex `POST /v1/responses`: HTTP 200, visible output and `response.completed` received.
3. OpenCodex `POST /v1/messages` with either `stream:true` or `stream:false`: fails with `upstream stream ended before a terminal frame (truncated response)`.

The relevant wire-format difference is that this upstream emits data-only SSE records:

```text
data: {"type":"response.output_text.delta",...}

data: {"type":"response.completed",...}

data: [DONE]
```

It does not emit a separate `event:` line. An SSE `event` field is optional; absent fields use the default message event.

### Expected behaviour

The Claude Messages bridge should accept Responses SSE frames whose event type is carried by the JSON payload's `type` field when the optional SSE `event:` field is absent. It should translate the complete stream and emit Anthropic `message_stop`, matching the already-successful `/v1/responses` passthrough path.

```shell
# Provider uses adapter=openai-responses and baseUrl=https://xxxx.cn/v1.
# API key is redacted.

curl http://127.0.0.1:10101/v1/messages \
  -H 'content-type: application/json' \
  -H 'x-api-key: REDACTED' \
  -H 'anthropic-version: 2023-06-01' \
  --data '{
    "model": "claude-opus-4-6",
    "max_tokens": 64,
    "stream": true,
    "thinking": {"type": "adaptive"},
    "output_config": {"effort": "low"},
    "messages": [{"role": "user", "content": "Reply exactly OK"}]
  }'
```

The translated upstream request was captured and redacted. It correctly used:

```json
{
  "stream": true,
  "reasoning": {"effort": "low"}
}
```

`reasoning.summary` was intentionally disabled with `modelSupportsReasoningSummaries:false`; this is unrelated to the SSE parsing failure.

### Actual response or error

```shell
HTTP 200 for stream:true, followed by:

data: {"type":"message_start",...}
data: {"type":"ping"}
data: {"type":"error","error":{"type":"overloaded_error","message":"upstream stream ended before a terminal frame (truncated response)"}}

For stream:false:
HTTP 502
{"type":"error","error":{"type":"overloaded_error","message":"upstream stream ended before a terminal frame (truncated response)"}}
```

### Upstream documentation

- SSE specification: https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation
- No public cubapi-specific wire-format documentation is available. Behaviour above was verified from a direct redacted HTTP response.

`src/claude/outbound.ts` currently discards frames without an explicit SSE event name before parsing the JSON payload:

```ts
if (!eventName || !dataLine) continue;
```

At EOF it consequently never sees `response.completed` and emits the truncation error.

Suggested behaviour:

1. Require `dataLine`, but not `eventName`, before parsing JSON.
2. Parse the payload.
3. If `eventName` is absent and `data.type` is a string, use `data.type` as the event name.
4. Preserve current strict terminal-frame validation.

This should be a narrow compatibility fix: it does not weaken truncation detection because `response.completed` still must be observed.

### Additional context

A combo failover control test also succeeded independently: an intentionally unreachable first target failed with 502, OpenCodex selected the second target, and `/v1/responses` returned the expected output. This isolates the failure to Responses SSE-to-Anthropic translation rather than provider routing or failover.

### Checks

- [x] I searched existing provider and compatibility issues, including #312; it covers a different Anthropic-adapter EOF/framing bug.
- [x] The expected behaviour is based on the SSE specification and the concrete Claude Code streaming requirement.

snowyukitty:
Closes #700.

## Summary

- accept Responses-compatible data-only SSE records in the Claude Messages bridge by falling back
  to the parsed JSON payload's `type` when the optional SSE `event:` field is absent;
- preserve explicit event-name precedence and strict terminal-frame validation;
- add streaming, non-streaming, mixed-framing, precedence, ignored-frame, and truncation
  regressions.

## Problem

The Claude Messages bridge currently requires every upstream Responses SSE record to carry both an
`event:` line and a `data:` line before it parses the JSON:

```ts
if (!eventName || !dataLine) continue;
```

The provider in #700 emits valid data-only records whose Responses event type is in the JSON
payload:

```text
data: {"type":"response.output_text.delta",...}

data: {"type":"response.completed",...}

data: [DONE]
```

The HTML SSE standard permits data-only messages; `event:` is not required and the default event
type is `message`. Because the bridge drops these records before parsing them, it never observes
`response.completed`. Both Claude streaming and non-streaming requests consequently end as:

```text
upstream stream ended before a terminal frame (truncated response)
```

The two Claude response modes share this converter, so one parser gate causes both failures.

## Fix

Parse any record with a non-empty `data:` payload, then resolve its Responses event name as:

1. the explicit SSE `event:` value, when present; otherwise
2. the parsed JSON payload's string `type`.

The scope remains narrow:

- explicit `event:` values still take precedence over a conflicting payload `type`;
- untyped data-only JSON objects remain ignored;
- non-JSON sentinels such as `[DONE]` remain ignored;
- `response.completed`, `response.incomplete`, or `response.failed` is still required before EOF;
- cancellation, keepalive, terminal, error-taxonomy, image, and tool-call paths are unchanged.

This accepts the gateway's framing without weakening the existing fail-closed truncation contract.

## Tests

New regressions cover:

- a fully data-only Responses stream through Anthropic SSE output;
- data-only frames through non-streaming aggregation;
- explicit and data-only records interleaved in one stream;
- explicit event-name precedence over a conflicting payload type;
- ignoring untyped data-only records;
- `[DONE]` without a Responses terminal frame still producing a truncation error rather than
  `message_stop`.

Activation was verified before the source change: the streaming and non-streaming data-only tests
failed against the old gate, while the explicit-precedence control passed.

Final verification on commit `fcd3298f`:

- `bun test tests/claude-outbound.test.ts` — **30 passed, 0 failed**
- `bun test tests/claude-messages-endpoint.test.ts` — **26 passed, 0 failed**
- `git diff --check` — passed
- full isolated suite with Bun 1.3.14:
  `bun scripts/test.ts --max-concurrency 4 ./tests/` —
  **5,977 passed, 1 Windows-only skip, 0 failed across 429 files**

The bounded concurrency is a local WSL scheduling accommodation, not a test exclusion. At Bun's
default concurrency, two pre-existing deadline-sensitive tests intermittently starved under load;
the same tests passed 10/10 and 5/5 in isolated repetitions, and the complete lower-concurrency run
was clean.

## Notes

- No authentication, credential, logging, provider-routing, or workflow path changes.
- No live provider credential was used; the tests encode the redacted wire shape from #700.
- This is the TypeScript `dev` contribution. Any `dev2-go` port remains the maintainer's transition
  responsibility under the repository policy.

## Checklist

- [x] Docs or release notes were updated when needed. (No user-facing configuration or API
  documentation changes are required.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming response handling for gateways that omit SSE event names.
  * Preserved text updates and message aggregation across mixed and data-only streams.
  * Added safer handling for untyped or incomplete terminal messages to prevent incorrect stream completion.

### DIFF
--- a/src/claude/outbound.ts
+++ b/src/claude/outbound.ts
@@ -453,12 +453,15 @@ export function responsesSseToAnthropicSse(
                 if (line.startsWith("event: ")) eventName = line.slice(7).trim();
                 else if (line.startsWith("data: ")) dataLine += line.slice(6);
               }
-              if (!eventName || !dataLine) continue;
+              if (!dataLine) continue;
               let data: unknown;
               try { data = JSON.parse(dataLine); } catch { continue; }
               if (!isRec(data)) continue;
-              if (terminated) continue;
-              handleFrame(eventName, data);
+              // Responses-compatible gateways may omit the optional SSE event field
+              // while retaining the event name in the JSON payload's required type.
+              const resolvedEventName = eventName || (typeof data.type === "string" ? data.type : "");
+              if (!resolvedEventName || terminated) continue;
+              handleFrame(resolvedEventName, data);
             }
           }
           // EOF without a terminal frame is a TRUNCATION, not success (devlog 100:
--- a/tests/claude-outbound.test.ts
+++ b/tests/claude-outbound.test.ts
@@ -13,6 +13,12 @@ function sse(name: string, data: Record<string, unknown>): string {
   return `event: ${name}\ndata: ${JSON.stringify(data)}\n\n`;
 }
 
+function dataOnlySse(data: Record<string, unknown>): string {
+  return `data: ${JSON.stringify(data)}\n\n`;
+}
+
+const DONE_SSE = "data: [DONE]\n\n";
+
 function streamFrom(text: string): ReadableStream<Uint8Array> {
   const encoder = new TextEncoder();
   return new ReadableStream({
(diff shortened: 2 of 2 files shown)


snowyukitty:
CI note: the Windows job reached the workflow's 12-minute cap after the root test suite, GUI tests, privacy scan, release-helper syntax check, and GUI lint had all passed. It was cancelled during GUI build; Ubuntu, macOS, all npm-global jobs, React Doctor, target enforcement, and CodeRabbit passed. The contributor account cannot rerun an upstream Actions job, so a maintainer rerun should be sufficient. I have not changed the reviewed commit just to retrigger CI.

lidge-jun:
Heads-up on a policy change that affects the port note in this PR's description: the `dev2-go` line is retired, so **no Go counterpart is expected for this change** and nothing further is owed after it lands on `dev`.

`dev` (Bun-native TypeScript) is the single integration line again. Keeping two integration lines in sync cost more than the Go port returned, so the carry/port obligation and the `needs-go-port` label are both gone. The `dev2-go` history is archived at [lidge-jun/opencodex-go-archive](https://github.com/lidge-jun/opencodex-go-archive) with its final tip tagged `archive/dev2-go` here.

Nothing is required from you — this PR is unaffected other than that the port line in the description no longer applies. Sorry for the mixed signals while the policy was in flux.

PR #713, opened 2026-07-29
Accept scheduler-exported launcher quotes

### LINKED ISSUE #708: [Bug]: Windows Task Scheduler export causes false stale-service health result
## Client or integration

OpenCodex CLI, `ocx doctor`, and the dashboard `/api/startup-health` route using the Windows Task Scheduler service backend.

## Summary

A freshly installed and operational Windows Task Scheduler service can be reported as stale. The generated task XML encodes quotation marks in `<Arguments>` as `&quot;`, while `schtasks /query /tn opencodex-proxy /xml` exports those quotation marks as literal `"` characters and retains required entity escaping such as `&amp;`.

`windowsTaskRegistrationHealthy()` compares the serialised `<Arguments>` value only against the `taskXmlString()` form. The semantically identical Task Scheduler export therefore fails the registration-health check even though the task and proxy are healthy.

## Reproduction

1. Install OpenCodex on Windows.
2. Run `ocx service install` from an elevated terminal.
3. Confirm `opencodex-proxy` is present and enabled with a logon trigger, `InteractiveToken`, `IgnoreNew`, `PT0S`, `wscript.exe`, and the expected launcher path.
4. Confirm `http://127.0.0.1:10100/healthz` returns `status: ok`.
5. Run `ocx doctor` or read `/api/startup-health`.

Observed result:

```text
AT RISK after restart (background service files are stale; run 'ocx service install')
```

Registered task export:

```xml
<Arguments>/b /nologo "C:\Users\a&amp;b\.opencodex\service-launcher.vbs"</Arguments>
```

Validator expectation:

```xml
<Arguments>/b /nologo &quot;C:\Users\a&amp;b\.opencodex\service-launcher.vbs&quot;</Arguments>
```

Both represent the same XML element text.

Proposed fix: accept both exact serialisations for the expected launcher arguments. Preserve escaping for `&`, `<`, and `>`; normalise only `&quot;` and `&apos;` to the literal characters Task Scheduler may emit. A regression test should use a launcher path containing `&` to prevent accepting an entirely unescaped path.

Local validation: `tests/service.test.ts` passes 41 tests and 190 assertions, strict typecheck passes, and the privacy scan passes. The full Windows suite could not be certified in the reporting environment because ACL-hardening tests deliberately removed access from the restricted runner SID.

## Version

OpenCodex 2.7.42.

## Operating system

Windows 11 x64.

stephen-drew:
## What changed

- Accept the two semantically equivalent Windows Task Scheduler serialisations of the service launcher arguments.
- Preserve XML escaping for ampersands and other significant characters.
- Add a regression test using a launcher path containing `&`.

## Why

Task Scheduler may export quotation marks as literal characters even when OpenCodex generated them as `&quot;`. The current literal comparison marks a valid, healthy background service as stale.

Fixes #708.

## Validation

- Focused Windows service suite: 41 tests, 190 assertions passed.
- Strict typecheck passed.
- Privacy scan passed.
- The repository-wide Windows suite could not be certified under the restricted runner SID because its ACL-hardening tests deliberately remove access from that SID.


stephen-drew:
Closing this draft after refreshing current `dev`: the same root cause and regression coverage, including an ampersand-containing launcher path, are already present under #608. This draft was built from a checkout 17 commits behind `dev`, so it should not be reviewed or merged.

PR #715, opened 2026-07-29
feat(codex): selection order for the account pool

XertroV:
## Selection order for the Codex account pool

The Codex pool has no way to say "drain these accounts before that one". Today
the synthetic `__main__` Codex Desktop login is actually biased *first* — it is
unshifted to the head of the eligible list — which is the opposite of what most
multi-subscription setups want, since the Desktop login is usually the one worth
preserving.

This adds a per-account **selection order**: an integer from `-100` to `100`,
higher used earlier, absent meaning `0`.

### How it works

One pure function, `selectPriorityTier`, narrows the already-eligibility-filtered
account list to the highest tier that still holds an account with quota headroom.
It is inserted at a single choke point, `getEligiblePoolAccounts`, so all three
rotation strategies (`quota`, `round-robin`, `fill-first`), failover,
alternate-picking, and the routing previews inherit tiering from that one hook
rather than each growing their own copy of the rule.

Selection order is an **ordering** boundary and never an **eligibility** one. It
cannot admit an account that pause, cooldown, health, or reauth already excluded,
and it never re-binds a thread that already has an account — an affinity-held
thread still moves only when its own account crosses `autoSwitchThreshold`. A
tier drains only when every member is over the threshold, cooling down,
soft-avoided, paused, or needs reauth; **unknown quota never drains a tier**, so
an unprimed account is not skipped.

With no entries configured — or with every account on the same tier — a fast path
returns the input list unchanged, so the pick sequence is identical to today's.

Two further pieces:

- **Preemption for unbound requests.** When a higher tier regains headroom (a
  weekly reset, say), new sessions move back up immediately. This is a
  runtime-only cursor, so there is no disk write on the request path, and it is
  skipped for independent quota scopes so a `spark`-scope request can never move
  the shared active cursor.
- **A pin for manual selection.** Choosing an account by hand records it in
  `activeCodexAccountPinned` and suppresses upward preemption until the account
  drains or routing otherwise moves off it. This is what makes the dashboard's
  "use this one now" actually stick under round-robin and fill-first.

### Two pre-existing bugs fixed along the way

Both are inert today and were found while tracing the new paths:

- `pickAlternateCodexAccount` post-filtered `excludeId` *after* eligibility. Under
  tiering, a same-request retry could dead-end when the excluded account was the
  sole healthy member of the top tier. It now passes `excludeId` into eligibility
  so the tier walk sees the exclusion.
- `subagent-model-fallback` read `config.activeCodexAccountId` directly instead of
  the effective active account, so it missed the round-robin cursor entirely.

### Rebased onto current dev

This branch was rebuilt onto the current `dev` tip, resolving the GUI conflicts
with the exact-selector routing that landed after this branch last merged dev.
A regression test proves selection order never bypasses an exact account selector:
with priorities ordering one account ahead of another, an exact `accountId` still
resolves to exactly that account while the unbound pool path honors the order.

### Verification

`bun run typecheck`, `bun run privacy:scan`, GUI `tsc -b`, the GUI suite,
GUI eslint + `lint:i18n`, the core pool/config/cli suites, and `bun run build:gui`
are all green on the rebased head. Pre-existing Windows symlink-`EPERM` and
load-sensitive auth-api ordering failures reproduce identically on the clean dev
base and are not introduced here.

### GUI

The Codex Auth page now shows a "Selection order" control on every pool card and the main-account card, with presets First (+2), Earlier (+1), Normal (0), Later (-1), Last (-2), plus a PINNED badge when an account is manually selected.

Codex Auth selection-order control

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added configurable Codex account selection priorities in the web dashboard and `ocx account priority` CLI command.
  * Added account pinning indicators and clearer switching behavior for threads and in-flight requests.
  * Account lists now show priority values with numeric and preset ordering options.
  * Added APIs to view the active account and update account priorities.
* **Documentation**
  * Updated account-pool, CLI, configuration, and API guidance across supported languages.
* **Bug Fixes**
  * Account pins and priorities are cleared appropriately when accounts are removed or settings change.


lidge-jun:
**LAND-AFTER #671** — the two are complementary, and that determines the order.

First the thing worth clearing up, because on the surface this looks like it competes with #671 for the same subsystem. It does not. Your priority tiers filter accounts that are already eligible (`src/codex/pool-rotation.ts:66-111`), inserted at the shared eligibility choke point (`src/codex/routing.ts:700-728`). #671's qualified selector never reaches that path — it assigns `fixedAccountId` directly and short-circuits before pool selection, affinity and alternates are consulted. So priority and pinning can never contend with an exact account binding, and neither PR supersedes the other.

That makes the order: #671 lands first after its security review, then this rebases onto it. During the rebase, keep the invariant that a `fixedAccount` path bypasses tier selection, pin preemption, quota switching and alternate retry. One combined regression would lock it: configure priorities and a pin pointing at account B, request the namespace bound to account A, assert A is used and neither the active nor pinned account changes. That is the only cross-PR gap I found.

**The design reads well.** Defaulting to a flat tier at zero means existing deployments see no behaviour change until someone opts in. Degrading malformed preference data instead of rejecting the whole config (`src/config.ts:671-727`) is the right call for a field users will hand-edit. Releasing a manual pin when its account becomes paused, unusable, reauth-required or over threshold means a pin cannot strand the pool, and clearing the older pin when priority changes so the latest operator instruction wins is a sensible tiebreak.

Tests are behavioural, not decorative: tier invariants, fallback, pin ceiling, drained tiers, quota reset, pin expiry, failover, quota-scope isolation.

**The conflicts are the real work.** Nine files, all GUI and GUI tests, because `dev` reworked that surface underneath you — `AccountPoolStrategyControls.tsx`, `CodexAccountPool.tsx`, `CodexPoolStrategySetting.tsx`, `codex-account-pool-cards.tsx`, `codex-account-pool-main-card.tsx`, `useCodexAccountPool.ts`, `styles.css`, `ui.tsx`, and `gui/tests/account-pool-strategy.test.tsx`. Current `dev` now has the shared custom `Select`, progressive account loading with skeletons, the `account-pool-strategy-card` layout, paused-badge tooltips, and `Select` `id` support. Taking your side wholesale would regress the progressive-loading UX that landed since you opened this.

Note the `ui.tsx` one specifically: `dev` just gained `id` support on `Select`, and your branch extends the same signature with `id`/`describedBy`/`title`. Those want merging, not choosing.

Because nine files conflict, the green result on this branch no longer proves compatibility with current `dev` — the GUI tests will need to run again after the rebase.

Selection determines which stored credential goes upstream, so this also needs explicit security review under `MAINTAINERS.md:33-34`.

What happens next: wait for #671, rebase onto it preserving both the exact-selector bypass and the current progressive-loading UX, add the combined regression, then request security review. `dev` CI is fully green now.

lidge-jun:
Thanks, @XertroV — this is substantive work, and we want the tiering/pin design.

Per the discussion in #821, this remains **LAND-AFTER #671** (exact account routing). That ordering still holds. Once #671 lands, please:

1. Rebase over the substantial GUI conflicts.
2. Add the exact-selector bypass regression: selection order must not bypass an exact account selector.
3. Resolve the current P2 findings.
4. Request security review, since this touches the credential-selection surface.

This is sequencing, not hesitation. Thank you for the depth of the design and the review follow-through here.

Wibias:
Uploading UI screenshot for the selection-order control.

Wibias:
[GD] Rebased onto current `dev` (`ef1317871`) and verified the three P2 findings against the new head `b5074291f`:

1. **Clear pins when priorities are set through config** — `src/cli/config-command.ts` now calls `clearCodexAccountPin(config)` when the written path is `codexAccountPriorities` (line 113). Covered by `tests/cli-headless-parity.test.ts` ("config set releases the manual pin when it writes the selection order").
2. **Serialize priority writes with manual account switches** — `gui/src/hooks/useCodexAccountPool.ts` cross-gates both directions: `switchAccount` rejects when `priorityMutationRef.current` is set, and `setAccountPriority` rejects when `switchingRef.current` is set. Covered by the controller contract test and the behaviour suite.
3. **Preserve scope isolation after the soft avoid expires** — `src/codex/routing.ts` `applyFailureFailover` now threads `quotaScope` and guards `promoteActiveCodexAccount` with `isIndependentCodexQuotaScope`. Covered by "an independent scope still leaves shared routing alone once its soft avoid expires" which asserts the routed account first.

All three threads resolved with these verified fixes on the head.

Wibias:
[GD] Addressed feedback

Addressed the CodeRabbit round on the rebased head `0978221a1`:

- `src/codex/routing.ts` (Major): `CodexAccountUsabilityOptions` is now threaded through `pickPriorityPreemption` and the round-robin/fill-first branches of `pickAlternateCodexAccount`, so `nativeMainSelectionOnly` is honored consistently with the quota path during main-profile draining. New regression in `tests/codex-pool-rotation.test.ts` ("selection options reach the tier and alternate paths").
- `gui/src/components/AccountPriorityControl.tsx` + `gui/src/ui.tsx`: `Select` no longer renders the dropdown (and disables option buttons + guards `selectIndex`) when `disabled` is true, so a controller flipping `priorityUpdatingId` cannot leave clickable options that silently drop the second update.
- `gui/src/i18n/en|ja|ko|ru.ts`: switch descriptions now state existing account-affine threads and in-flight requests keep their captured account; only new/unbound requests use the selected account's order tier.
- `gui/tests/codex-account-pool-controller.test.ts`: slice windows now assert the anchors exist before slicing.
- `gui/tests/codex-account-pool-toast-tone.test.tsx`: the no-op test uses the real `.codex-auth-page-head__feedback.is-ok/is-err` selectors; the fixed 20ms sleep is a flush.
- `tests/codex-auth-context.test.ts`: the exact-selector regression now also configures a live pin and asserts it survives.
- `tests/cli-headless-parity.test.ts`: each config-set case asserts the written order value.
- `tests/codex-pool-rotation.test.ts`: added the routing-layer exact-selector/selection-options regression.
- README + docs (en/ja/ko/ru/zh-cn): Pool/Direct boundary, full pin-release conditions, priority-aware `current`, US-English "afterward", and a split Russian pin-release sentence.

commit: 0978221a17b859d5db873608828c6befc26f428b

Issue #717, opened 2026-07-29
CI: windows-latest test job cancelled at the 12-minute ceiling, reported as a failing check


lidge-jun:
## Area

GitHub Actions CI (`.github/workflows/ci.yml`).

## Summary

The `test` job's `timeout-minutes: 12` is no longer enough headroom for `windows-latest`, so PR and `dev` runs are being cancelled mid-`Test` and surfaced to authors as a failing check. This is a CI budget problem, not a defect in the PRs that trip it.

Evidence from live runs:

| Run | Branch | windows-latest | Wall time |
| --- | --- | --- | --- |
| 30459554635 | `dev` | success | **11 min** |
| 30493348190 | `dev` | cancelled | — |
| 30497549930 | `dev` | cancelled | — |
| 30498312557 | `dev` | cancelled | — |
| 30498333662 | `dev` | cancelled | — |
| 30498427875 | `dev` | cancelled | — |

The last green Windows run took 11 minutes against a 12-minute ceiling — roughly one minute of margin. Every `dev` run after it was cancelled.

Job-level confirmation on PR #653 (`879efb243`, job `90733427766`):

```
conclusion: cancelled
started 2026-07-29T23:11:50Z  completed 2026-07-29T23:23:55Z   (12m05s)
Test: cancelled
GUI tests / Privacy scan / GUI lint / GUI build / CLI help smoke: skipped
```

The job dies at 12m05s with `Test` cancelled and every later step skipped, which is the signature of the job timeout rather than a test assertion failure.

## Why this is costing review time

`gh pr checks` renders a cancelled job as `fail`, so a PR whose code is fine looks broken. Two examples from today:

- **#711** — reported `windows-latest fail`. The `conclusion` was `cancelled`; a rerun passed and it merged cleanly (`d24c5233f`).
- **#653** — same pattern. The first failure was a genuine `failure`, but the rerun came back `cancelled` at 12m05s. Its focused suites pass locally (38 pass) with `bun x tsc --noEmit` clean, so the preset itself is not implicated.

Reviewers currently have to open `gh api .../check-runs` and read `conclusion` on every red Windows job to tell a real failure from a timeout. That is easy to skip, and skipping it either blocks a good PR or hides a real one.

## Suggested direction

The immediate unblock is more headroom — the existing code comment already records that 8 minutes was raised to 12 for the same reason, so the ceiling has been chasing suite growth rather than leading it. Worth pairing that with something that stops the ratchet:

[...]

PR #737, opened 2026-07-30
fix(update): tolerate late proxy readiness

### LINKED ISSUE #720: [Bug][Windows]: GUI npm update reports restart failure just before fallback becomes healthy
## Client or integration

OpenCodex dashboard

## Area

Service lifecycle

## Summary

During a GUI npm update with restart enabled, a non-elevated Windows Task Scheduler service refresh can fail with the expected `WINDOWS_SCHTASKS_CREATE_ACCESS_DENIED` marker and correctly fall back to a direct proxy start.

In the reproduced run, the fallback needed slightly more than the fixed 15-second readiness deadline. The update job ended at `09:55:28.828` with `proxy restart never became healthy`, while the target-version proxy became healthy around `09:55:29`.

The package update had succeeded and the replacement proxy was running, so this was a false terminal failure. The first 15-second miss also entered the redundant explicit-restart path, including two 30-second busy-port waits and another service-install attempt.

Expected: accept a target-version/new-PID replacement that becomes healthy within a realistic bounded Windows cold-start window, while still warning that the scheduler registration was not refreshed.

## Reproduction

1. Install OpenCodex through npm and install the Windows Task Scheduler service.
2. Run the proxy non-elevated.
3. From the dashboard, update `2.7.42` to `2.7.43` with restart enabled.
4. Let `schtasks /create /tn opencodex-proxy /xml ... /f` return access denied.
5. Observe the intended direct fallback start.
6. On a sufficiently slow cold start, observe the update job fail at the 15-second readiness deadline immediately before `/healthz` becomes healthy.

This is timing-dependent. In the reproduced run, the fallback process was created around `09:55:14`, the job failed at `09:55:28.828`, and the listener became healthy around `09:55:29`.

The current arrival deadline is 15 seconds followed by a 15-second stability window. A bounded regression could keep the replacement unhealthy until virtual time `15.25s`, then expose a changed PID and the target version, and assert success without entering the explicit-restart path. A final identity probe at the deadline would also close the narrow boundary race.

## Version

2.7.42 → 2.7.43 (`v2.7.43`, commit `d1f544bbc22d25b9b2bd3c8e776fb8e3242b4ed5`)

## Operating system

Windows 11 Pro 25H2, build 26200.8894, x64

## Provider and model

Not provider-specific

## Logs or error output

```shell
Reinstalling the background service with the updated files...
WindowsSchtasksError: Windows access denied while running Task Scheduler.
OCX_ERROR_CODE=WINDOWS_SCHTASKS_CREATE_ACCESS_DENIED
opencodex: service refresh failed — starting the proxy directly instead.
Proxy starting on port 10100.
Update installed. Restarting proxy...
npm self-update did not leave a healthy proxy; performing explicit restart...
Port 10100 still busy after 30s; refusing to hop — reinstall may fail until the port is free.
Service reinstall failed (exit 1); falling back to a direct proxy start.
Port 10100 still busy after 30s (reclaim could not free the socket); not starting on another port.
error: proxy restart never became healthy on 127.0.0.1:10100
```

This appears to be a narrow follow-up to #515. It is distinct from #591/#698, whose locale-independent access-denied marker worked correctly in this reproduction.

Relevant readiness/restart code:

- https://github.com/lidge-jun/opencodex/blob/d1f544bbc22d25b9b2bd3c8e776fb8e3242b4ed5/src/update/job.ts#L25-L34
- https://github.com/lidge-jun/opencodex/blob/d1f544bbc22d25b9b2bd3c8e776fb8e3242b4ed5/src/update/job.ts#L520-L553
- https://github.com/lidge-jun/opencodex/blob/d1f544bbc22d25b9b2bd3c8e776fb8e3242b4ed5/src/update/job.ts#L679-L767

## Redacted configuration

Not configuration-dependent. The reproduction uses an npm installation and the Windows Task Scheduler service backend.

## Checks

luvs01:
## Summary

- extend the replacement-proxy arrival budget from 15 seconds to 30 seconds while preserving the existing 15-second stability window
- guarantee one identity-aware health probe at the arrival deadline so a proxy that becomes ready on the boundary is not misclassified as timed out
- add deterministic regressions for readiness at 30 seconds and for the reported post-15-second npm service replacement, including the no-redundant-restart evidence path

This stays intentionally narrower than #533/#557: it does not change process recovery, port reclamation, or restart ownership, and both current PR heads still retain the 15-second arrival loop.

Fixes #720

## Verification

- `bun test tests/update-job.test.ts` (34 passed)
- `bun x tsc --noEmit`
- `bun scripts/privacy-scan.ts`
- `git diff --check`
- The initial monolithic Windows run was cut off by a 10-minute local command budget. Current CI documents an approximately 11.8-minute Windows baseline and uses a 20-minute job budget (#717); isolated current-`dev` slices continued making progress and completed, confirming a harness cutoff rather than a deadlock. The separate 5-second `cli-restore-back` test budget found during diagnosis is already covered by open #707. Cross-platform CI still requires repository approval for this PR.

## Checklist

- [x] Docs or release notes were updated when needed. (No documentation change is needed for this bounded readiness correction.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved restart health checks to recognize services that become healthy right at the timeout deadline.
  * Extended the restart health-check window from 15 to 30 seconds.
  * Adjusted restart confirmation timing to avoid unnecessary restarts when a replacement service becomes healthy shortly after the previous cutoff.

* **Tests**
  * Added coverage for deadline health checks and successful replacement-service confirmation.

## Review readiness checklist

### DIFF
--- a/src/update/job.ts
+++ b/src/update/job.ts
@@ -26,7 +26,7 @@ const RELEASE_NOTES_URL = "https://github.com/lidge-jun/opencodex/releases/lates
 const UPDATE_JOB_FILENAME = "update-job.json";
 const UPDATE_TIMEOUT_MS = 180_000;
 const RESTART_TIMEOUT_MS = 60_000;
-const RESTART_HEALTH_TIMEOUT_MS = 15_000;
+const RESTART_HEALTH_TIMEOUT_MS = 30_000;
 const RESTART_STABILITY_WINDOW_MS = 15_000;
 /** Legacy active records did not persist a worker PID, so age is their only safe recovery signal. */
 export const UPDATE_JOB_LEGACY_STALE_MS = 10 * 60_000;
@@ -533,7 +533,10 @@ async function awaitRestartedProxyHealthy(
   const hostname = captured.hostname;
   const startDeadline = now() + RESTART_HEALTH_TIMEOUT_MS;
 
-  while (now() < startDeadline) {
+  while (true) {
+    // Always make one identity-aware probe at or after the boundary. A replacement
+    // becoming healthy on the final tick must not be mistaken for a timeout.
+    const finalProbe = now() >= startDeadline;
     if (await probe(port, hostname)) {
       updateJob(job, {}, `Proxy reported healthy on ${hostname}:${port}; confirming it stays up...`);
       const stableUntil = now() + RESTART_STABILITY_WINDOW_MS;
@@ -547,7 +550,8 @@ async function awaitRestartedProxyHealthy(
       updateJob(job, {}, `Proxy stayed healthy for ${Math.trunc(RESTART_STABILITY_WINDOW_MS / 1000)}s after restart.`);
       return { ok: true };
     }
-    await sleep(250);
+    if (finalProbe) break;
+    await sleep(Math.min(250, Math.max(0, startDeadline - now())));
   }
 
   return { ok: false, reason: "timeout" };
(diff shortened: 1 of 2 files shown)


lidge-jun:
Merged into `dev` as `62e937614` (head `11069f055`).

Verified locally on the merged tree:

- The loop change is the substantive part, not the 15s→30s constant. `awaitRestartedProxyHealthy` now always makes one final identity-aware probe **at** the deadline instead of exiting the `while (now() < startDeadline)` test first, and the sleep is clamped to the remaining time so it cannot overshoot past the boundary. A replacement that becomes healthy on the last tick is no longer misreported as a timeout.
- Falsification: `34 pass / 0 fail` on the merged tree. Reverting only `src/update/job.ts` to the pre-merge `cf30878fc` gives `32 pass / 2 fail`; restoring it returns `34 pass / 0 fail`. The 15.25s arrival case in `tests/update-job.test.ts` is doing real work.
- Full suite on the merged tree: `6080 pass / 0 fail`, 6081 tests across 438 files.

The honest limit, which your own doc comment already records: a replacement arriving later than 30s is still reported as a restart failure. This narrows the false-positive window rather than removing it, and the worst case now takes 30s arrival plus a 15s stability window before success is declared. That tradeoff is the right one for an update path where a wrong "restart failed" costs the user more than a slower confirmation.

Closing as merged. Issue #720 is closed with the same evidence.

PR #745, opened 2026-07-30
fix: normalize tool function schemas to always have root type: "object"

white-source:
## Problem

Some provider APIs (notably DeepSeek) strictly validate JSON Schema in function tool definitions and reject requests where the root `type` is `null` or missing with HTTP 400:

```
Invalid schema for function 'codex_app__automation_update':
schema must be a JSON Schema of 'type: "object"', got 'type: null'.
```

This occurs because Codex injects tools like `codex_app__automation_update` without a `parameters` field. When opencodex forwards the request, the serialized schema has `type: null` (or no `type` at all) at its root.

OpenAI and most other providers tolerate this, but stricter validators reject it. The same issue has been reported across multiple proxy projects:
- cc-switch: [8+ issues](https://github.com/farion1231/cc-switch/issues?q=is%3Aissue+type%3Anull+deepseek) tracking the same root cause
- openclaw: [#106277](https://github.com/openclaw/openclaw/issues/106277)

## Root Cause

Two forwarding paths in opencodex can pass through a broken tool schema:

1. **Adapter translation path**: `buildTools()` in `parser.ts` uses `(t.parameters ?? {})` — when `parameters` is null/absent, the resulting object has no root `type`.
2. **Passthrough path** (used by `wire_api=responses` providers like DeepSeek): `createResponsesPassthroughAdapter` serializes `_rawBody` verbatim. The raw body contains the same broken tool schemas directly.

### 1. `src/responses/parser.ts` — `buildTools()`
Added `normalizeParameters()` helper that ensures every function tool's `parameters` is a valid JSON Schema object with `type: "object"` and a `properties` key. Covers the adapter translation path (Anthropic, Gemini, Cursor, etc.).

### 2. `src/adapters/openai-responses.ts` — passthrough adapter
Added `normalizeToolSchemas()` function that walks both top-level `body.tools` and Responses Lite `additional_tools[].tools`, fixing any function tool whose parameters schema is invalid. Inserted into the sanitization chain at the outermost position so it runs regardless of which strip functions are active.

Both changes are idempotent — already-valid schemas pass through unchanged. Tools that are not `type: "function"` are not touched.

## Testing

- [x] `bun build` syntax check on both files
- [x] `bun x tsc --noEmit` typecheck passes
- [x] Unit-tested with edge cases: no `parameters`, `parameters: null`, `type: null`, already-valid schemas, nested `additional_tools`, non-function tool types
- [x] Verified live with DeepSeek (`deepseek-v4-pro` via `wire_api=responses`) — the 400 error is resolved

## Notes

This is a defensive normalization — the output is a valid JSON Schema regardless of what the upstream tool definition provides. The change is minimal (70 lines added across 2 files) and does not alter any existing behavior for already-valid schemas.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with strict JSON schema validation for function tools.
  * Automatically fills in missing or invalid tool parameter definitions with a valid object schema.
  * Prevents requests from being rejected when tool parameters are missing, null, or incorrectly structured.

### DIFF
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -354,6 +354,61 @@ function isPlainObject(v: unknown): v is Record<string, unknown> {
   return !!v && typeof v === "object" && !Array.isArray(v);
 }
 
+/**
+ * Ensure every function tool exposes a valid object JSON Schema at its
+ * `parameters` root. Strict validators (e.g. DeepSeek) reject schemas whose
+ * root `type` is `null`, missing, or non-`"object"` with HTTP 400.
+ *
+ * Codex tools such as `codex_app__automation_update` arrive without a
+ * `parameters` field, which would otherwise serialize to a schema lacking the
+ * root `type`. Walks both top-level `tools` and Responses Lite
+ * `additional_tools[].tools`. Namespace grouping is flattened by earlier strip
+ * steps, so any function tool seen here is already top-level.
+ */
+function normalizeToolSchemas(body: unknown): unknown {
+  if (!isPlainObject(body)) return body;
+
+  const fixFunctionParams = (tool: unknown): unknown => {
+    if (!isPlainObject(tool) || tool.type !== "function") return tool;
+    const params = isPlainObject(tool.parameters) ? { ...tool.parameters } : {};
+    let changed = false;
+    if (params.type !== "object") { params.type = "object"; changed = true; }
+    if (!isPlainObject(params.properties)) { params.properties = {}; changed = true; }
+    return changed ? { ...tool, parameters: params } : tool;
+  };
+
+  let changed = false;
+
+  // Top-level tools array.
+  if (Array.isArray(body.tools)) {
+    const tools = body.tools.map((t) => {
+      const fixed = fixFunctionParams(t);
+      if (fixed !== t) changed = true;
+      return fixed;
+    });
+    if (changed) body = { ...body, tools };
+  }
+
+  // Responses Lite `additional_tools` items embed their own tools arrays.
+  if (Array.isArray(body.input)) {
+    let inputChanged = false;
+    const input = body.input.map((item) => {
+      if (!isPlainObject(item) || item.type !== "additional_tools" || !Array.isArray(item.tools)) return item;
+      let innerChanged = false;
+      const innerTools = item.tools.map((t) => {
+        const fixed = fixFunctionParams(t);
+        if (fixed !== t) innerChanged = true;
+        return fixed;
+      });
+      if (innerChanged) { inputChanged = true; return { ...item, tools: innerTools }; }
+      return item;
+    });
+    if (inputChanged) { changed = true; body = { ...body, input }; }
+  }
+
+  return body;
+}
+
 const MAX_RESPONSES_CALL_ID_LENGTH = 64;
 const REPAIRED_CALL_ID_PREFIX = "call_ocx_";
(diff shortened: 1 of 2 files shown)


lidge-jun:
**NEEDS-CHANGES** — the change looks right, but it ships with no tests.

The only files here are `src/adapters/openai-responses.ts` and `src/responses/parser.ts`. `AGENTS.md` asks for a focused regression next to the existing tests for any behaviour change in `src/`, and this one has none.

That matters more than usual because the normalization lands on two independent paths. `src/responses/parser.ts:140-163` handles parsed tool translation; `src/adapters/openai-responses.ts:357-414` handles raw Responses passthrough including nested `additional_tools`. A regression in one is invisible from the other, and both can fail in either direction — sending an invalid schema upstream, or rewriting a schema that was already fine.

The cases worth pinning:

- `parameters` missing, and `parameters: null`
- root `type` missing, `null`, and set to something other than `"object"`
- top-level passthrough tools **and** nested `additional_tools`
- an already-valid object schema passes through untouched
- non-function tools are left alone
- running normalization twice produces the same result

Please assert against the final serialized upstream request body rather than a helper's return value. That is what proves both production paths actually apply the normalization, which is the claim the PR is making.

What happens next: add that coverage and push. The change itself reads correctly; it just needs something holding it in place.

---

**CI context (shared across the current review round).** `dev` is now at `62e937614`. The baseline breakage that made every required check red is fixed: run [30545575865](https://github.com/lidge-jun/opencodex/actions/runs/30545575865) has `macos-latest`, `ubuntu-latest`, and all three `npm-global` jobs green. The one remaining red is `windows-latest`, and it is not a test failure — it is a Bun 1.3.14 runtime panic (`heap.zig:deleteMin` reached through `spawnSync` with a timeout, triggered by our Windows ACL hardening path). That is a runtime defect being tracked separately, so please do not treat a red `windows-latest` as a signal about your change. Rebase onto current `dev` before your next push so your checks run against the repaired baseline.

lidge-jun (changes requested):

Blocking: no regression tests. The only changed files are `src/adapters/openai-responses.ts` and `src/responses/parser.ts`, and `AGENTS.md` asks for focused coverage on any behaviour change in `src/`.

That matters more than usual here because the normalization lands on two independent paths — parsed tool translation and raw Responses passthrough including nested `additional_tools`. A regression in one is invisible from the other, and both can fail in either direction: sending an invalid schema, or rewriting a valid one.

Please assert against the final serialized upstream request body rather than a helper return value, so the tests prove both production paths actually apply it. The case list is in my earlier comment.

The change itself reads correctly — it just needs something holding it in place.

lidge-jun:
Landed on `dev` in `d52aa6846`, with you credited as co-author.

Your diagnosis and both normalizers were right. What held this back was that the PR carried no regression test, and this repo has shipped tests twice recently that passed against both the fixed and unfixed behavior — so "the tests are green" was not evidence the fix worked.

What landed:

- `src/responses/parser.ts` — normalizes at build time, returning already-valid schemas **unchanged** rather than rebuilding them, so existing property sets cannot be reordered or dropped.
- `src/adapters/openai-responses.ts` — the same normalization for top-level `tools` and `input[].additional_tools.tools`, because the passthrough copies `_rawBody` and never reaches the parser. Fixing only the parser would have left every passthrough request still broken.
- Tests covering: missing `parameters`, `properties` without a root type, an already-valid schema staying intact, and both guarantees asserted on the **serialized outgoing body** rather than the parsed context.

Ablating both normalizers fails three tests, so they are not vacuous.

Closing this PR since the change is on `dev`. Thanks — the two-path insight was the part that mattered.

PR #749, opened 2026-07-30
fix(responses): reject expired forward continuations

### LINKED ISSUE #702: Codex App resume silently drops prior context after the 1-hour replay-state TTL expires
### Client or integration

Codex App

### Area

Proxy and routing

### Summary

Resuming a Codex App thread after it has been idle for more than one hour can silently drop all prior conversation context when using the built-in `openai` Codex-login forward provider.

The resumed request still succeeds with HTTP 200, but only the newest user message reaches the model. The model therefore behaves as if the existing thread were a brand-new conversation. I expected either the complete thread context to be restored or an explicit continuation-expired error; OpenCodex should not silently continue with a truncated delta.

This occurred without restarting the OpenCodex proxy. The same OpenCodex conversation ID and the same `openai / gpt-5.6-sol` route were used before and after the idle period.

### Reproduction

1. Start OpenCodex 2.7.42 with the built-in `openai` provider using Codex-login forward authentication.
2. Open Codex App, create a thread, and complete enough turns that the next request depends on earlier context.
3. Leave the thread idle for more than one hour. The proxy may remain running throughout.
4. Resume the same thread in Codex App and send a short instruction such as `continue` or a question referring to prior work.
5. Observe that the request returns HTTP 200, but the assistant asks for the missing task/configuration or explicitly says it cannot see earlier context.

In the observed thread, the last healthy request at 16:39 contained 85,073 input tokens. The first resumed request at 18:58, approximately 2 hours 19 minutes later, contained only 22 input tokens and zero cached input tokens. Subsequent messages contained only 41, 9, 11, and 15 input tokens. All requests used the same OpenCodex conversation ID, `provider=openai`, and `resolvedModel=gpt-5.6-sol`.

The apparent cause is the fixed one-hour continuation-state TTL:

- [`RESPONSE_TTL_MS = 60 * 60 * 1_000`](https://github.com/lidge-jun/opencodex/blob/7cb15bff4efd7beafc49d0a125432230c6ec9461/src/responses/state.ts#L6-L7)
- On a replay miss, the passthrough path warns but continues even though earlier turns may be missing: [`src/server/responses/core.ts`](https://github.com/lidge-jun/opencodex/blob/7cb15bff4efd7beafc49d0a125432230c6ec9461/src/server/responses/core.ts#L1148-L1171)
- Forward mode strips `previous_response_id` because the ChatGPT Codex backend rejects it: [`src/adapters/openai-responses.ts`](https://github.com/lidge-jun/opencodex/blob/7cb15bff4efd7beafc49d0a125432230c6ec9461/src/adapters/openai-responses.ts#L470-L482)

Together, these behaviors turn an expired local replay entry into a successful request containing only the current-turn delta.

Suggested fixes:

1. Make the continuation-state TTL configurable and/or increase the default substantially while retaining the existing count, memory, and snapshot-size caps.
2. In Codex-login forward mode, fail closed with an explicit continuation-state-expired error when `previous_response_id` cannot be expanded. Do not silently forward a delta known to be missing prior turns.
3. Add a regression test that advances time beyond one hour and verifies that a resumed forward-mode request either replays the full history or returns an explicit error.

### Version

2.7.42

### Operating system

macOS 26.5.2 (25F84)

Codex App 26.721.81911; bundled Codex runtime 0.146.0-alpha.3.1.

### Provider and model

`openai` (Codex-login forward provider) / `gpt-5.6-sol`

### Logs or error output

```shell
[responses] previous_response_id <redacted> not found in local replay state (model gpt-5.6-sol); forwarding without it — earlier turns may be missing from this request
```

Sanitized request-log comparison:

```text
16:39  provider=openai  resolvedModel=gpt-5.6-sol  inputTokens=85073  status=200
18:58  provider=openai  resolvedModel=gpt-5.6-sol  inputTokens=22     status=200
19:00  provider=openai  resolvedModel=gpt-5.6-sol  inputTokens=41     status=200
```

### Redacted configuration

```json
{
  "defaultProvider": "openai",
  "providers": {
    "openai": {
      "authMode": "forward"
    }
  }
}
```

### Checks

LeoWang331:
## Summary

- fail closed when a canonical OpenAI forward continuation carries `previous_response_id` but its local replay state is unavailable or expired
- return the existing structured `400 invalid_request_error` before authentication, adapter construction, or model-upstream dispatch; canonical replay misses also avoid unnecessary quota priming
- preserve fresh replay expansion, ordinary requests, and native `previous_response_id` support for API-key Responses providers
- harden the Issue #702 integration harness so partially initialized servers, `fetch`, and `Date.now` are always restored

Fixes #702

## Root cause

The canonical ChatGPT Responses backend rejects native `previous_response_id`. OpenCodex therefore depends on its local replay state to expand the omitted history before forwarding. When that state expired, the prior path logged the replay miss, stripped `previous_response_id`, and still forwarded only the continuation delta, silently losing conversation context.

## Fail-closed conditions

The request is rejected only when all of the following are true:

1. the original request carries `previous_response_id`;
2. local replay input was not successfully expanded;
3. the final route is the canonical OpenAI forward provider; and
4. that upstream cannot safely recover the complete history.

The rejection happens before any upstream request. Expired continuations return HTTP 400 with `invalid_request_error`.

## Compatibility controls

- Expired forward continuation: rejected locally; no upstream request is made.
- Fresh forward continuation: replay history is expanded and the upstream request succeeds.
- Ordinary request without `previous_response_id`: succeeds normally.
- API-key Responses provider: native `previous_response_id` is preserved and succeeds normally.

## Validation

- Issue #702 targeted test: 4 passed, 0 failed
- Related Responses state/server/forward-adapter tests: 178 passed, 0 failed
- `bun run typecheck`: passed
- `git diff --check`: passed

The local full-suite single-command run did not complete within the available execution time, so this PR does **not** claim that the full suite passed locally. Repository CI is the complete cross-platform validation authority for Linux, Windows, and macOS.

## Summary by CodeRabbit

* **Bug Fixes**
  * Continuation requests that reuse an outdated or improperly referenced previous response now fail clearly with an `invalid_request_error`, instructing users to start a new session.
  * Avoided unnecessary quota priming for these invalid replay attempts.
  * Improved forwarding behavior to preserve native previous-response continuation for API-key responses, and correctly forward expanded context for valid fresh continuations.
* **Tests**
  * Added end-to-end coverage for expired, fresh, and ordinary forward continuation scenarios.

### DIFF
--- a/src/server/responses/core.ts
+++ b/src/server/responses/core.ts
@@ -1144,10 +1144,6 @@ export async function handleResponses(
   }
   if (parsed._compactionRequest === true) parsed._cursorIsolateConversation = true;
 
-  if (isThreadSpawnRequest(req.headers)) {
-    await maybePrimeSubagentQuota(config);
-  }
-
   let route: RouteResult;
   try {
     route = routeModel(config, parsed.modelId);
@@ -1156,6 +1152,17 @@ export async function handleResponses(
       return comboUnavailableResponse(err.message);
     }
     return formatErrorResponse(404, "invalid_request_error", err instanceof Error ? err.message : String(err));
+  }
+
+  const hasUnexpandedPreviousResponse = !!parsed.previousResponseId
+    && parsed._previousResponseInputExpanded !== true;
+  // A canonical replay miss must not poll quota upstream before the final fail-closed decision.
+  // Cached fallback state can still select a provider with native continuation support below.
+  if (
+    isThreadSpawnRequest(req.headers)
+    && !(hasUnexpandedPreviousResponse && isCanonicalOpenAiForwardProvider(route.provider))
+  ) {
+    await maybePrimeSubagentQuota(config);
   }
 
   let authCtx: CodexAuthContext = { kind: "main", accountId: null };
@@ -1204,6 +1211,20 @@ export async function handleResponses(
   // runs against the FINAL route so native-only fallback can rescue a routed primary.
   if (!isCanonicalOpenAiForwardProvider(route.provider) && unreadableEncryptedAgentTask) {
     return unreadableEncryptedAgentTaskResponse();
+  }
+
+  // The canonical ChatGPT backend rejects previous_response_id, so a local replay miss leaves no
+  // safe way to recover the omitted history. Fail before auth, adapter construction, or upstream
+  // I/O instead of stripping the id and silently forwarding a context-free delta (#702).
+  if (
+    hasUnexpandedPreviousResponse
+    && isCanonicalOpenAiForwardProvider(route.provider)
+  ) {
+    return formatErrorResponse(
+      400,
+      "invalid_request_error",
+      "OpenAI forward continuation state is unavailable or expired; start a new session instead of reusing this previous_response_id.",
+    );
   }
 
   await applyFinalRouteRequestNormalization({ parsed, route, config, req, logCtx });
--- a/tests/issue-702-expired-replay-state.test.ts
+++ b/tests/issue-702-expired-replay-state.test.ts
@@ -1,0 +1,351 @@
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { saveConfig } from "../src/config";
+import {
+  resetSubagentModelFallbackStateForTests,
+  setSubagentQuotaPrimeForTests,
+} from "../src/codex/subagent-model-fallback";
+import {
+  clearResponseStateForTests,
+  clearResponseStateMemoryForTests,
+  responseStateMetrics,
+  type ResponseStateMetrics,
+} from "../src/responses/state";
+import { startServer } from "../src/server";
+import type { OcxConfig } from "../src/types";
+import { fakeChatGptJwt } from "./helpers/fake-chatgpt-jwt";
+import { installIsolatedCodexHome, type IsolatedCodexHome } from "./helpers/isolated-codex-home";
+
+const originalFetch = globalThis.fetch;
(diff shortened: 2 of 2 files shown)


lidge-jun:
Merged into `dev` as `6be54f588` (head `652151880`).

Verified locally on the merged tree before pushing, not from the PR description:

- The guard fires only when all three hold: `previous_response_id` is present, `_previousResponseInputExpanded` is not `true`, and `isCanonicalOpenAiForwardProvider(route.provider)` — which `src/providers/openai-tiers.ts:32-35` defines as `adapter: openai-responses` **and** `authMode: forward` **and** the canonical Codex base URL. API-key Responses providers therefore keep native `previous_response_id`, which `tests/issue-702-expired-replay-state.test.ts` asserts directly.
- Moving `maybePrimeSubagentQuota` behind the route decision is a real improvement: a request already destined for a local 400 no longer probes quota upstream first.
- Falsification: `4 pass / 0 fail` on the merged tree. Reverting only `src/server/responses/core.ts` to the pre-merge `cf30878fc` gives `3 pass / 1 fail`; restoring it returns `4 pass / 0 fail`. The test genuinely binds the fix rather than passing incidentally.
- Full suite on the merged tree: `6080 pass / 0 fail`, 6081 tests across 438 files.

One thing worth naming for anyone reading this later: the scope is wider than the issue title. Every replay miss now returns the same 400, not just TTL expiry — restart and capacity eviction included. That is the right call, and the message ("unavailable or expired") already says so, but it is a behaviour change beyond the reported symptom.

Thanks for the harness hardening too; restoring `fetch` and `Date.now` on partially initialized servers is the kind of thing that saves someone an afternoon later.

Closing as merged. Issue #702 is closed with the same evidence.

PR #752, opened 2026-07-30
fix(tray): isolate Windows tray host from proxy sockets

### LINKED ISSUE #733: [Bug][Windows]: tray host can retain inherited proxy socket handles and keep the old port bound after restart
## Client or integration

OpenCodex dashboard

## Area

Service lifecycle

## Summary

On Windows, a long-lived OpenCodex tray lineage can retain socket handles inherited from the running proxy. After the supported `POST /api/system/restart` path stops the original Bun server, the configured port can remain `LISTENING`, with existing connections in `CLOSE_WAIT`. `netstat` continues to attribute those rows to the original binder PID even though that PID no longer exists, and the replacement proxy cannot bind the port.

In the observed OpenCodex 2.7.43 incident:

- Before stopping the tray, port 10100 had 1 `LISTENING` and 27 `CLOSE_WAIT` rows attributed to dead proxy PID 24204.
- `ocx tray stop --json` immediately reduced all port-10100 rows to zero.
- Restarting the tray did not recreate the stale rows.
- A reboot, a TCP-stack reset, and broad process termination were not required.

The incident was triggered by a local maintenance script calling the supported management restart endpoint. The script did not create PID 24204 or the socket handles. This report does not claim that the failure occurs spontaneously or on every restart.

Expected: the tray may survive a proxy restart, but it must not retain proxy data-plane socket handles. The configured port should become reusable when the old proxy is drained and stopped.

## Reproduction

The underlying mechanism reproduces deterministically on Windows with both OpenCodex's bundled Bun and a recent Canary:

```ts
import { spawn } from "node:child_process";

const server = Bun.serve({
  hostname: "127.0.0.1",
  port: 18131,
  fetch() {
    return new Response("ok");
  },
});

const child = spawn(
  process.execPath,
  ["-e", "setTimeout(() => {}, 30000)"],
  {
    detached: true,
    stdio: "ignore",
    windowsHide: true,
  },
);
child.unref();

console.log(`parent=${process.pid} child=${child.pid}`);
await server.stop(true);
console.log("server.stop(true) resolved");
process.exit(0);
```

1. Run the script on Windows.
2. After the parent exits, inspect port 18131 with `netstat -ano`.
3. The port remains `LISTENING` and is attributed to the dead parent PID.
4. Binding the same port fails while the detached child remains alive.
5. When that exact child exits, the TCP row disappears immediately and rebinding succeeds.
6. Remove the child spawn and repeat as a control; the port is released normally.

Confirmed runtimes:

- Bun `1.3.14+0d9b296af`, bundled by OpenCodex 2.7.43
- Bun `1.4.0-canary.1+b22e0e6d0`

This rules out the selected Canary runtime as the incident-specific cause.

The corresponding OpenCodex product path is:

1. Start the proxy on a fixed port.
2. Start or install the Windows tray through the dashboard while the proxy is live.
3. Invoke the dashboard Drain & restart action.
4. Verify that the tray remains alive and the original proxy port is immediately reusable.

The live proxy launches tray actions as child processes, and a tray action then launches a detached, long-lived tray host. That lineage gives an inherited server handle a lifetime beyond the proxy.

## Version

OpenCodex 2.7.43 (`d1f544bbc22d25b9b2bd3c8e776fb8e3242b4ed5`)

The relevant lifecycle and tray paths remain present on current `dev` (`67c731e655c7490b62cffb3fd9c85b68d3221eb1`).

## Operating system

Windows 11 Pro 25H2, build 26200.8894, x64

## Provider and model

Not provider-specific

## Logs or error output

```shell
Before tray stop:
port 10100: LISTENING=1, CLOSE_WAIT=27
reported owner PID: 24204
PID 24204: no longer live

After:
ocx tray stop --json
port 10100 rows: 0

After restarting the tray:
port 10100 rows: 0
```

Related but not duplicate:

- #456 added Windows port reclamation after update/stop, including best-effort TCP-row removal.
- #563 / #580 added dashboard drain-and-restart.
- #594 handles replacement-process spawn failure.
- #720 is a readiness-deadline race where the replacement eventually becomes healthy.

The current gap is inherited socket ownership by a long-lived descendant. Awaiting `server.stop(true)` is not sufficient in the minimal reproduction, and elevated `SetTcpEntry` is not a substitute for preventing inheritance.

Relevant current source:

- Dashboard tray action child: https://github.com/lidge-jun/opencodex/blob/67c731e655c7490b62cffb3fd9c85b68d3221eb1/src/server/windows-tray-control.ts#L9-L22
- Detached tray host: https://github.com/lidge-jun/opencodex/blob/67c731e655c7490b62cffb3fd9c85b68d3221eb1/src/tray/windows.ts#L417-L427
- Management restart: https://github.com/lidge-jun/opencodex/blob/67c731e655c7490b62cffb3fd9c85b68d3221eb1/src/server/management/system-restart.ts#L133-L162
- Explicit-port start waits without dropping TCP rows: https://github.com/lidge-jun/opencodex/blob/67c731e655c7490b62cffb3fd9c85b68d3221eb1/src/cli/index.ts#L121-L136

A useful Windows regression test would keep the tray host alive, stop the proxy, and assert that the old listen port is reusable without stopping the tray or requiring elevated TCP-row deletion.

## Redacted configuration

```json
{
  "port": 10100
}
```

The issue is independent of provider, model, credentials, and routing configuration. The Windows tray was installed and running.

## Checks

luvs01:
## Summary

- Launch the long-lived Windows tray host through a short `ProcessStartInfo.UseShellExecute=true` PowerShell broker instead of spawning it directly from the live Bun proxy process tree.
- Keep executable and CLI paths out of the PowerShell command text: the encoded launcher is constant, paths are validated with the existing `safePath` rules, and values are passed through environment variables.
- Add a Windows regression test that keeps the tray child alive, stops the originating Bun listener, and proves that the same port can be rebound immediately. The fixture also covers Unicode paths and paths containing `&` and `%TEMP%`.

Fixes #733.

## Root cause and scope

On Windows, Bun 1.3.14 and the tested Bun 1.4 canary can pass an inheritable proxy listen socket through the current `proxy -> tray action -> detached tray host` process lineage. The old proxy can then exit while the long-lived tray host keeps the socket handle open, leaving the prior port unavailable.

This change isolates only the long-lived tray host. It does not change tray registration, registry ownership rules, UAC/elevation behavior, heartbeat semantics, or the tray host's command-line contract. The existing heartbeat remains the authority for whether startup succeeded.

The relevant upstream Bun handle-list work (oven-sh/bun#33160) is currently open and not suitable as a released dependency, so this PR keeps the workaround local and removable.

## Verification

- Bun 1.3.14 focused tray/restart tests: 40 passed.
- Bun 1.4.0-canary.1 focused tray/restart tests: 40 passed.
- The new real-socket regression passed independently on both runtimes and verifies that the child remains alive after the original port is rebound.
- `bun run typecheck`: passed on Bun 1.3.14 and Bun 1.4.0-canary.1.
- `git diff --check`: passed.
- The full suite reached a 304-second local limit before producing a final result. The exact descendant process chain belonging to that run was identified by worktree, command line, parent PID, and start time, then terminated and confirmed absent; this is not reported as a full-suite pass.

## Draft status

This remains a draft pending an explicit review of the ShellExecute/environment boundary, AppLocker or EDR compatibility, and a final Windows full-suite run. Those checks are intentionally deferred rather than treated as complete here.

## Checklist

- [x] Docs or release notes were updated when needed (none required for this internal Windows launch-path fix).

### DIFF
--- a/src/tray/windows.ts
+++ b/src/tray/windows.ts
@@ -414,17 +414,43 @@ function assertWindows(): void {
   if (process.platform !== "win32") throw new Error(`The opencodex tray is Windows-only (current platform: ${process.platform}).`);
 }
 
-function spawnTray(state: WindowsTrayEntry): void {
-  const child = spawn(state.bun, [state.cli, "__tray-host"], {
-    detached: true,
+const DETACHED_TRAY_HOST_LAUNCHER = [
+  "$startInfo = New-Object System.Diagnostics.ProcessStartInfo",
+  "$startInfo.FileName = $env:OCX_TRAY_HOST_BUN",
+  "$startInfo.Arguments = $env:OCX_TRAY_HOST_ARGS",
+  "$startInfo.UseShellExecute = $true",
+  "$startInfo.WindowStyle = [System.Diagnostics.ProcessWindowStyle]::Hidden",
+  "$child = [System.Diagnostics.Process]::Start($startInfo)",
+  "if ($null -eq $child) { throw 'Windows tray host did not start.' }",
+  "$child.Dispose()",
+].join("; ");
+
+const DETACHED_TRAY_HOST_LAUNCHER_B64 = Buffer.from(DETACHED_TRAY_HOST_LAUNCHER, "utf16le").toString("base64");
+
+export function launchWindowsTrayHost(state: WindowsTrayEntry): void {
+  const bun = safePath(state.bun);
+  const cli = safePath(state.cli);
+  execFileSync(windowsPowerShellPath(), [
+    "-NoLogo",
+    "-NoProfile",
+    "-NonInteractive",
+    "-EncodedCommand",
+    DETACHED_TRAY_HOST_LAUNCHER_B64,
+  ], {
     stdio: "ignore",
     windowsHide: true,
+    timeout: 15_000,
     env: {
       ...process.env,
+      OCX_TRAY_HOST_BUN: bun,
+      OCX_TRAY_HOST_ARGS: `${quoteRunValue(cli)} __tray-host`,
       OCX_TRAY_ENTRY_B64: Buffer.from(JSON.stringify(state), "utf8").toString("base64"),
     },
   });
-  child.unref();
+}
+
+function spawnTray(state: WindowsTrayEntry): void {
+  launchWindowsTrayHost(state);
 }
 
 function parseTrayHostEntry(): WindowsTrayEntry {
@@ -443,6 +469,8 @@ export async function runWindowsTrayHost(): Promise<void> {
   assertWindows();
   const entry = parseTrayHostEntry();
   delete process.env.OCX_TRAY_ENTRY_B64;
+  delete process.env.OCX_TRAY_HOST_BUN;
+  delete process.env.OCX_TRAY_HOST_ARGS;
   const child = spawn(windowsPowerShellPath(), windowsTrayProcessArgs(entry, "Run", process.pid), {
     stdio: "ignore",
     windowsHide: true,
--- a/tests/helpers/windows-tray-inheritance-child.ts
+++ b/tests/helpers/windows-tray-inheritance-child.ts
@@ -1,0 +1,7 @@
+import { writeFileSync } from "node:fs";
+
+const pidPath = process.env.OCX_TRAY_TEST_PID_FILE;
+if (!pidPath) throw new Error("Missing OCX_TRAY_TEST_PID_FILE.");
+
+writeFileSync(pidPath, String(process.pid));
+await Bun.sleep(30_000);
--- a/tests/windows-tray.test.ts
+++ b/tests/windows-tray.test.ts
@@ -1,8 +1,10 @@
 import { describe, expect, test } from "bun:test";
-import { existsSync, readFileSync } from "node:fs";
+import { copyFileSync, existsSync, mkdtempSync, readFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
 import { join } from "node:path";
 import {
   buildWindowsTrayRunCommand,
+  launchWindowsTrayHost,
   parseWindowsTrayRunValue,
   readWindowsTrayRunValueWithAsyncRunner,
   readWindowsTrayRunValueWithRunner,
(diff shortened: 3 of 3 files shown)


lidge-jun:
**NEEDS-SECURITY-REVIEW, then LAND** — the mechanism holds up.

I checked this against #733 rather than taking the description on faith, and the isolation is real rather than cosmetic. `src/tray/windows.ts:417-454` stops spawning the tray host directly and goes through a synchronous PowerShell broker that sets `UseShellExecute = true`, which routes creation through `ShellExecuteEx` instead of `CreateProcess`. Windows only inherits a socket when the handle is inheritable *and* the child is created with inheritance enabled, so crossing the shell boundary breaks the condition that keeps the port held. The broker itself may briefly inherit, but `execFileSync` waits for it to exit, and the process that outlives everything is created on the far side.

Passing the executable and CLI paths through environment variables instead of interpolating them into the encoded PowerShell program is the right call, and clearing those three variables at `:462-465` before the long-lived WinForms process starts closes the obvious follow-on.

The regression at `tests/windows-tray.test.ts:217-264` is what makes this reviewable: it keeps the launched child alive, stops the original Bun listener, immediately rebinds the same port, and asserts the child survived. That binds the reported failure instead of asserting on source strings. The fixture at `tests/helpers/windows-tray-inheritance-child.ts` is a genuinely long-lived descendant, not a stub.

Two things before merge. This changes an executable-launch trust boundary and deliberately invokes the Windows shell, which puts it under "other security-boundary changes" in `MAINTAINERS.md:33-34` — so it needs explicit security review even though it touches no credentials. And the socket test skips off Windows, so fresh Windows CI is required evidence, not a formality.

The PR is still a draft and merges cleanly against the current `dev` head. `dev` CI is now fully green, so a red check here would be about this change.

What happens next: mark it ready when you are satisfied, get Windows CI green, and request security review. I found no code blocker in the isolation itself.

lidge-jun:
Closing as superseded. `c1ecbe1b5 feat(windows): add restart-safe tray controls` landed the same protection on `dev`: `runWindowsTrayHost()` spawns the tray with `stdio: "ignore"` (`src/tray/windows.ts:474-477`), so the host cannot inherit the proxy's socket handles. `git cherry dev <this branch>` reports the commit as patch-equivalent.

#733 is closed with that evidence.

Issue #710, opened 2026-07-29
[Bug]: Kiro login import misses Windows Kiro CLI token DB path


minerva32:
### Client or integration

OpenCodex dashboard

### Area

Authentication and account pool

### Summary

On Windows, `ocx login kiro` does not import an existing Kiro CLI login. It reports that no kiro-cli token was found and falls back to manual access-token input, even though `kiro-cli whoami` confirms an active login.

I expected OpenCodex to import the local Kiro CLI session automatically, the same way it does for the other supported CLI token-store locations.

Root cause appears to be credential discovery: the Kiro credential lookup does not include the Windows Kiro CLI default SQLite path.

### Reproduction

1. Install Kiro CLI on Windows with the official installer (`irm 'https://cli.kiro.dev/install.ps1' | iex`).
2. Run `kiro-cli login` and complete sign-in.
3. Confirm the session with `kiro-cli whoami`, which prints the logged-in account.
4. Run `ocx provider add kiro`.
5. Run `ocx login kiro`.
6. Observe that the login prompts for a manual access token instead of importing the local CLI session.

Workaround that resolves it:

1. Set the Windows Kiro CLI DB path as a persistent user environment variable:

```powershell
[Environment]::SetEnvironmentVariable(
  'KIROCLI_DB_PATH',
  "$env:LOCALAPPDATA\Kiro-Cli\data.sqlite3",
  'User'
)
```

2. Restart the proxy with `ocx restart` so the server process picks up the new environment.
3. Run `ocx login kiro` again. It now succeeds without any manual token paste.

Note that setting the variable only in the calling shell is not enough, because the running proxy process performs the import and does not see the shell-local value.

The Windows Kiro CLI stores its auth database at:

```text
%LOCALAPPDATA%\Kiro-Cli\data.sqlite3
```

That database has the expected shape, including an `auth_kv` row keyed `kirocli:social:token` and a `state` row keyed `api.codewhisperer.profile`.

Suggested fix: add `%LOCALAPPDATA%\Kiro-Cli\data.sqlite3` to the Kiro credential discovery candidates alongside the existing macOS/Linux paths.

### Version

2.7.42

### Operating system

Windows 11 Pro 25H2 (10.0.26200)

### Provider and model

kiro / kiro-auto

### Logs or error output

```shell
> ocx login kiro

Opening browser for kiro login...

[...]

PR #768, opened 2026-07-30
fix(kiro): resolve Windows kiro-cli executable without PATH

aljjang95:
## Summary
- After the Windows SQLite store path fix, forced/add-account Kiro login still spawned bare `kiro-cli` and failed when the installer binary was not on PATH.
- Add a pure executable resolver that prefers PATH, then falls back to `%LOCALAPPDATA%\Kiro-Cli\kiro-cli.exe` and `C:\Program Files\Kiro-Cli\kiro-cli.exe`.
- Use that resolver for forced-login spawn and rollback logout, and document the Windows install layouts.

## Why
- On Windows, OpenCodex could import an existing Kiro session from `%LOCALAPPDATA%\Kiro-Cli\data.sqlite3` while `kiro-cli.exe` remained undiscoverable for add-account/browser login.
- The official MSI installs to `C:\Program Files\Kiro-Cli\`, and some local installs keep the binary next to the session DB.

## Testing
- `bun test tests/kiro-windows-cli-executable-path.test.ts tests/kiro-windows-cli-db-path.test.ts tests/kiro-oauth.test.ts`
- `bun x tsc --noEmit`
- `bun run privacy:scan`

## Risks
- Only affects Kiro CLI discovery/spawn; token import and OAuth storage paths are unchanged.
- If a future installer places `kiro-cli.exe` outside PATH, LocalAppData, and Program Files, the bare-command fallback still reports the original spawn error.

## Summary by CodeRabbit

- **New Features**
  - Improved Kiro CLI detection across Windows, macOS, and Linux environments.
  - Kiro credential import and related commands now locate installed CLI executables automatically.

- **Documentation**
  - Added Windows-specific guidance for Kiro credential imports, including database and CLI locations.

- **Tests**
  - Added coverage for platform-specific executable discovery and fallback behavior.

### DIFF
--- a/docs-site/src/content/docs/guides/providers.md
+++ b/docs-site/src/content/docs/guides/providers.md
@@ -165,6 +165,10 @@ read-only. Two environment variables make the source and token row selection exp
 - `KIROCLI_TOKEN_KEY` selects the exact `auth_kv` token key when a database contains multiple
   otherwise ambiguous token rows. A missing selection fails login instead of guessing.
 
+On Windows, import looks for `%LOCALAPPDATA%\Kiro-Cli\data.sqlite3`. Forced/add-account login
+also needs the local CLI binary: opencodex first uses `PATH`, then falls back to
+`%LOCALAPPDATA%\Kiro-Cli\kiro-cli.exe` and `C:\Program Files\Kiro-Cli\kiro-cli.exe`.
+
 After a successful import, opencodex persists the imported credential to
 `~/.opencodex/auth.json`.
 
--- a/src/oauth/kiro-credentials.ts
+++ b/src/oauth/kiro-credentials.ts
@@ -169,6 +169,65 @@ export function resolveKiroCliNativeSessionEntries(
   return [{ location: "kiro-cli-linux-data", path: posix.join(home, ".local", "share", "kiro-cli", "data.sqlite3") }];
 }
 
+/**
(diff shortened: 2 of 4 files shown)


lidge-jun:
Landed on `codex/260731-pr-merge-round`, with your authorship preserved on the commit. That branch is a staging line: it merges into `dev` and ships once the performance work in flight there lands.

`ba77eb8e9` — landed with two additions on top of your patch.

The core fix is right and the layout coverage is thorough: PATH first, both `PATH` and `Path` casings, `%LOCALAPPDATA%`, Program Files, and a bare-name fallback so spawn still reports the original error rather than a path the resolver invented. #710 taught Windows where the SQLite store lives, so a box that could import tokens but not launch login was exactly the gap.

Two things I added:

1. **Directory guard.** `existsSync` accepts a *directory* named `kiro-cli`, which then reaches `spawn()` and fails with EACCES instead of falling through to the next candidate. Candidates now have to be files.
2. **Environment parsing coverage.** The tests only ever injected `pathEntries`, so the `PATH`/`Path` parsing itself was never exercised. A case now drives both casings through the real code.

Both were checked by ablation rather than assumed — removing the `isFile` guard fails the directory case (6 pass / 1 fail), restoring passes 7/7.

Windows behavior is unverifiable from macOS, so the matrix CI on the branch is what confirms it.

Issue #726, opened 2026-07-30
[Bug]: Dashboard logs capped at 200 entries with no pagination — conversation filter and statistics incomplete


jhste102lab:
### Client or integration
OpenCodex dashboard

### Area
Dashboard

### Summary
The in-memory request log ring buffer is hardcoded to `MAX_LOG_SIZE = 200` (`src/server/request-log.ts:129`). Once 200 entries accumulate, the oldest entry is silently dropped. The Dashboard `#logs` tab fetches all logs with a single `GET /api/logs` call (no query parameters) and applies filters (conversation, surface) purely client-side. This means:

1. The conversation log filter only searches among the last 200 entries — older conversations are invisible
2. Statistics (usage, costs, tokens) derived from the dashboard only reflect the last 200 requests
3. There is no server-side pagination or offset/limit mechanism on `GET /api/logs`

### Reproduction
1. Start `ocx` and send >200 requests through the proxy
2. Open the dashboard `#logs` tab
3. Observe that only the most recent 200 entries are visible
4. Apply the conversation filter — it can only find conversations among those 200 entries
5. Older request data is permanently lost from the dashboard view

### Source references
- `MAX_LOG_SIZE = 200`: `src/server/request-log.ts:129`
- Ring buffer eviction: `src/server/request-log.ts:222-224` (`if (requestLog.length > MAX_LOG_SIZE) requestLog.shift()`)
- Hydration cap from disk: `src/server/request-log.ts:196-219` (reads max 200 from persisted usage log on startup)
- `tail` parameter also capped at 200: `src/server/request-log.ts:761`
- API endpoint has no pagination params: `src/server/management/logs-usage-routes.ts:126-128` — returns all in-memory logs
- Frontend fetches without query params, filters client-side: `gui/dist/assets/index-cmds12BG.js`

### Expected behavior
- Full request history should be accessible in the dashboard, either by increasing the in-memory buffer size significantly or adding server-side pagination
- The conversation filter should be able to search across the full available history
- `GET /api/logs` should support `offset`/`limit` (or cursor-based) pagination, and the frontend should lazy-load or paginate entries

### Version
2.7.43

### Operating system
Linux (WSL)

### Checks
- [x] I searched existing issues and documentation.
- [x] I removed secrets, tokens, account details, request credentials, and personal data.

Wibias:
## [shipping-github] Research review

[...]

PR #791, opened 2026-07-30
fix(gui): render dashboard log timestamps in server timezone

### LINKED ISSUE #725: [Bug]: Dashboard log timestamps display in browser local timezone instead of system/server time
### Client or integration
OpenCodex dashboard

### Area
Dashboard

### Summary
Log timestamps in the Dashboard (`#logs` tab) are stored as server-side `Date.now()` epoch milliseconds but rendered using `toLocaleTimeString()` / `toLocaleString()`, which converts them to the browser local timezone. For a local management dashboard expected to reflect the server state, this can cause confusion if the browser and server are in different timezones (e.g., remote access or containerized deployments).

### Reproduction
1. Start `ocx` with the management API enabled
2. Open the dashboard in a browser that has a different timezone than the server
3. Navigate to the `#logs` tab
4. Observe that timestamps reflect the browser timezone, not the system/server timezone where opencodex is running

### Source references
- Timestamp source: `src/server/request-log.ts:272-274` — `Date.now()` for log IDs, `src/server/request-log.ts:694` — `start` field set to wall-clock server time
- Timestamp DTO passthrough: `src/server/management/shared.ts:143-162` — raw epoch ms passed to API response
- Display: `gui/dist/assets/index-cmds12BG.js` functions `bc()`, `xc()`, `Gs()`, `Ks()` — all use `new Date(e).toLocaleTimeString(locale)` / `.toLocaleString(locale)`, which formats in browser local timezone

### Expected behavior
Unclear whether this is intentional. If intentional (browser-local time is the standard web UX pattern), it is acceptable. If the intent is to show system/server time (since opencodex is a server-side tool and the dashboard is a management interface), the display should use a fixed timezone or UTC formatting.

### Version
2.7.43

### Operating system
Linux (WSL)

### Checks

Wibias:
## Summary
- Use server timeZone from the logs/settings APIs when formatting dashboard timestamps.

## Test plan

## Security
See research/security notes on #725 where applicable.

Fixes #725

Depends on the PR for #726 (log pagination /api/logs envelope).

### DIFF
--- a/gui/src/pages/Logs.tsx
+++ b/gui/src/pages/Logs.tsx
@@ -368,9 +368,10 @@ export default function Logs({ apiBase }: { apiBase: string }) {
     // failures flicker between the error banner, empty state, and stale table.
     if (!silent) setLoading(true);
     try {
-      const res = await fetch(`${apiBase}/api/logs`);
+      const res = await fetch(`${apiBase}/api/logs?limit=2000`);
       if (!res.ok) throw new Error(`${res.status} ${res.statusText}`.trim());
-      const next = await res.json() as LogEntry[];
+      const body = await res.json() as LogEntry[] | { logs?: LogEntry[] };
+      const next = Array.isArray(body) ? body : (body.logs ?? []);
       setLogs(next);
       writeSessionListCache(logsCacheKey(apiBase), next);
       setError(null);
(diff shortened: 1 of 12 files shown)


Wibias:

Issue #553, opened 2026-07-27
[Bug] GitHub Copilot model request fails with 502 and TLS hostname mismatch


burhiepotdult1982:
### Client or integration

Codex App

### Area

Provider adapter

### Summary

Requests using a GitHub Copilot-backed model intermittently fail with HTTP 502. The OpenCodex proxy reports ERR_TLS_CERT_ALTNAME_INVALID while fetching the GitHub Copilot API. I expected the request to complete successfully or return a provider-level error, not a TLS certificate hostname mismatch.

### Reproduction

1. Run OpenCodex locally on port 10100.
2. Configure the Codex App to use the OpenAI-compatible endpoint http://127.0.0.1:10100/v1.
3. Select the GitHub Copilot-backed model github-copilot/claude-sonnet-4.
4. Send a normal chat/request.
5. The request fails with HTTP 502.

Observed error:

unexpected status 502 Bad Gateway: Provider unreachable: ERR_TLS_CERT_ALTNAME_INVALID fetching https://api.individual.githubcopilot.com/chat/completions. For more information, pass verbose: true in the second argument to fetch(), url: http://127.0.0.1:10100/v1/responses

The failure appears to occur in the provider adapter when translating /v1/responses to the Copilot Chat Completions endpoint.

### Version

2.7.41

### Operating system

macOS 26.5.1 (Build 25F80), Apple Silicon

### Provider and model

github-copilot/claude-sonnet-4

### Logs or error output

```shell
unexpected status 502 Bad Gateway: Provider unreachable: ERR_TLS_CERT_ALTNAME_INVALID fetching https://api.individual.githubcopilot.com/chat/completions. For more information, pass verbose: true in the second argument to fetch(), url: http://127.0.0.1:10100/v1/responses
```

### Redacted configuration

```json
[proxy]
base_url = http://127.0.0.1:10100/v1
provider = GitHub Copilot
model = github-copilot/claude-sonnet-4

No credentials or tokens included.
```

### Checks

- [x] I searched existing issues and documentation.
- [x] I removed secrets, tokens, account details, request credentials, and personal data.

Ingwannu:
I checked the exact reported hostname and current `dev` transport path. OpenCodex accepts the OAuth token's allowlisted `*.githubcopilot.com` API endpoint, so `https://api.individual.githubcopilot.com/chat/completions` is an expected Copilot endpoint rather than a malformed URL.

I cannot reproduce the TLS failure now:

[...]

PR #798, opened 2026-07-31
fix(responses): tell TLS hostname mismatches apart from an unreachabl…

batuchek68-ux:
…e provider

A certificate that does not cover the host it was served for surfaced as `Provider unreachable: ERR_TLS_CERT_ALTNAME_INVALID fetching https://...`, which reads as if opencodex had built a wrong endpoint. Issue #553 was filed on that reading — the reporter concluded the adapter was mangling the Copilot URL, and a maintainer could not reproduce it because the URL was correct all along. That error almost always means TLS interception (corporate proxy, VPN, local MITM tooling) or a poisoned DNS answer, none of which the proxy can fix or is at fault for.

The message now names the host, says the proxy did not rewrite it, and gives the openssl s_client command that settles which certificate is actually being served.

The same three-line shape was duplicated at three catch sites in core.ts, so it moves into one helper. Sites 2 and 3 delegate entirely; site 1 keeps its own `outcome` ternary because that value also feeds recordCodexUpstreamOutcome. Verified byte-identical output against the old expression for TimeoutError, plain Error, empty-message Error, string, undefined, null, object and AbortError.

While rewriting the message, redact URL userinfo. A provider base URL can carry credentials as `https://user:token@host/` and the runtime error echoes the URL it was fetching, so the old text handed that straight back to the caller. Both branches now strip it.

The code match is anchored to the head of the message rather than a bare substring, so text that merely quotes the code back keeps the generic wording.

## Summary

- Explain the user-visible or maintainer-facing change.

## Verification

- List the commands or checks you ran.

## Checklist

### DIFF
--- a/src/server/responses/core.ts
+++ b/src/server/responses/core.ts
@@ -1,6 +1,7 @@
 import type { Server } from "bun";
 import { bridgeToResponsesSSE, buildResponseJSON, formatErrorResponse, type ResponsesTerminalStatus } from "../../bridge";
 import { formatPassthroughUpstreamError } from "./passthrough-error";
+import { describeUpstreamConnectFailure } from "./upstream-error";
 import {
   getConfigPath,
   multiAgentGuidanceEnabled,
@@ -1194,7 +1195,7 @@ export async function handleResponses(
       }
       const msg = outcome === "timeout"
         ? `Provider connect timeout after ${connectMs}ms`
-        : `Provider unreachable: ${err instanceof Error ? err.message : String(err)}`;
+        : describeUpstreamConnectFailure(err, connectMs);
       return formatErrorResponse(502, "upstream_error", msg);
     };
     try {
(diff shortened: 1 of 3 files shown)


lidge-jun:
Closing. This pull request contains no work by its author.

The head commit here is byte-identical to a branch that already exists in this repository and was authored by the maintainer. All six of these pull requests have the same shape: fork the repository, push the upstream branches back unchanged, and open them as incoming contributions.

| PR | head SHA | identical upstream branch |
| --- | --- | --- |
| #798 | `eef9a86d` | `codex/260728-tls-altname-diagnosis` |
| #799 | `d0763bf7` | `codex/260729-security-md-reporting-path` |
| #800 | `d380b1b0` | `codex/260731-pr-merge-round` |
| #802 | `a64aa585` | `codex/app-server-restart` |
| #803 | `b0434ea5` | `codex/pr533-update-recovery-hardening` |
| #804 | `f0acc36a` | `stack/gui-dense-workspaces` |

Every listed SHA resolves to the same commit on the upstream branch of the same name. The commit authors inside them are `lidge-jun`, `Wibias`, `ahmetb`, `aljjang95`, `hanbinnoh` and other existing contributors. The descriptions are the unmodified pull request template with every checklist item unchecked, and #798 additionally carries the maintainer's own commit message pasted in as if it were a summary. Several target `main`, which `AGENTS.md` reserves for maintainer-controlled promotion, and two are chained onto each other's heads in a cycle.

This is not a rebase, a resubmission of stalled work, or a fork that drifted. It burns maintainer review time and CI minutes on a diff that is already in the tree, and it presents other people's commits under a new author's pull request. Repository access is being revoked for this account.

PR #811, opened 2026-07-31
feat(remote): ship signed cross-platform agents

Ingwannu:
## Summary

- ship prebuilt Rust Remote Agents for Linux, macOS, and Windows on x64 and arm64
- build all six targets on native GitHub-hosted runners with pinned Rust and pinned Actions
- sign each binary plus a canonical manifest that binds platform, checksum, source commit, and npm package version
- verify the complete bundle during package preparation and fail closed again before the local Agent is spawned
- keep production signing authority in the protected `remote-agent-release` environment; dry-runs use an ephemeral job-local key
- update Remote onboarding, CLI docs, architecture notes, and localized GUI copy to match the outbound E2EE Agent flow

## Why

Remote currently has the outbound E2EE relay and Rust Agent source, but a normal npm install does not have a trustworthy cross-platform Agent delivery path. Users should not need Rust or runtime compilation, and an incomplete, stale, or modified bundle must never fall back to unsigned execution.

This change keeps the existing single-package install while making the supported platform contract explicit and verifiable.

## Security and release boundary

- pull-request and dry-run jobs cannot read the production signing key
- real release dispatches require the exact audited SHA and the protected release environment
- the runtime verifies the manifest signature, installed package version, selected binary checksum, and detached binary signature before spawn
- unsupported OS/architecture combinations and malformed or partial bundles stop with an actionable error
- no production Remote service, database, DNS, Cloudflare resource, or user device is deployed by this PR

The diff-scoped security review covered all 28 changed files. The initial signing-secret scope and unsigned provenance gaps were remediated before publication; no unresolved reportable finding remains.

## Validation

- `bun run typecheck`
- focused install, bundle, and workflow tests: 79 passed, 0 failed
- full root suite under a 50% CPU quota: 6,299 passed, 4 skipped; two 5-second CLI-help timeouts then passed 12/12 with the same quota and a 15-second test timeout
- GUI lint and React Doctor against the `dev` PR base
- privacy scan
- Rust `fmt`, `clippy -D warnings`, 6 unit tests, and release build with Rust 1.97.1
- docs-site production build
- Remote platform typecheck and demo web build

## Review requested

Please review the six native runner/target mappings, protected signing-environment policy, canonical manifest format, and package/runtime fail-closed behavior. This is intentionally opened as a draft against `dev`; no merge is requested yet.


lidge-jun:
Thanks for pushing this forward. Signed prebuilt Remote Agents are the trustworthy install path we want for users, and the signing and fail-closed verification direction is important.

At the current size (roughly 15k lines), though, this cannot be reviewed responsibly as one unit. Please split it into reviewable stages:

1. signing and verification core
2. per-platform packaging
3. GUI integration

Before the staged PRs are ready for review, please also fix the Unix-only APIs currently breaking the Windows Rust builds, fix the failing GUI tests, and rebase the series onto the current `dev` branch.

Happy to help define the split and the boundaries between those stages.

PR #829, opened 2026-08-01
Zero-leak state stores: hard caps, durable spill, and generation-fenced sweeping

lidge-jun:
## Summary

This lands the first four work-phases of the zero-leak unit (`devlog/_plan/260801_zero_leak_state_stores/`), attacking the unbounded in-memory growth behind the recent RSS climb.

- **Continuation hard cap + durable spill** (`33402140f` lineage): the Responses continuation store gets a byte cap; overflow spills to disk atomically (fsync + link-no-replace, generation-unique basenames) with a spill-failed tombstone and a bounded deferred-unlink queue (128 entries).
- **Blob/replay/vision/image byte caps** : Cursor blob store bounded at 16 MiB per entry / 64 MiB total with request-scoped pinning and two-phase admission; Antigravity vision and image caches get equivalent caps with provenance-merged accounting.
- **Generation-fenced state-store sweeper**: every keyed runtime store (quota, reauth, combo rotation, PID cache, ACL memos across 5 producers) is reconciled against the live config topology with per-attempt generation fencing, late-completion fences, and ESRCH-verified PID sweeps — orphaned rows can no longer survive a config change.
- **Fence simplification + memo-release consolidation**: the per-combo topology map collapsed into the global generation contract; hardened-secret-path memo release now happens at exactly one proven-removal point.

Each phase went through independent adversarial review (4 rounds each, final PASS) recorded in the devlog unit.

## Testing

- `bun run test`: 6570 pass / 3 skip / 0 fail (465 files)
- `bun x tsc --noEmit`: clean
- `bun run privacy:scan`: pass
- New regression suites: state-store-sweeper (12), continuation spill, blob admission, combos fencing, ACL memo release

## Notes for reviewers

Remaining phases of the unit (registry admission caps, appOwnedBytes observability, translator stream bounds, shell lifecycle, benchmark) will stack on this PR.

## Summary by CodeRabbit

* **New Features**
  * Large response continuations can be safely preserved and replayed using bounded storage.
  * Added configurable app-owned memory budgeting and privacy-safe memory reporting.
  * Added limits and eviction for caches, diagnostics, usage history, model discovery, and tool/resource payloads.
  * Added request and streaming safeguards for oversized translated content.
  * Usage and API-key history now indicate when results are truncated.
  * Added safer consent handling for repository starring actions.

* **Bug Fixes**
  * Missing or corrupted continuation data now returns a clear request error.
  * Stale configuration and temporary-file state are cleaned up more reliably.
  * Busy operations now return consistent retryable errors.


Ingwannu (changes requested):
Windows CI has branch-relevant failures that block this large state-store change. Two new previous_response_id cap/cleanup tests have a 5s test budget but took about 17.7s and 34.2s on Windows, so their current implementation/test shape is not cross-platform bounded. Separately, ephemeral harden success memo lifecycle expected one memo entry but observed five, showing the test does not isolate/reset shared ACL memo state. Please make the bounded tests deterministic or give them a justified Windows-safe budget, reset the ACL memo around the lifecycle assertion, and rerun the full Windows matrix. Given the 79-file state/persistence scope, I will not override a failing platform gate.

Ingwannu:
I extracted the confirmed Windows ACL temp-path memo leak into focused PR #840. Please exclude/rebase out the overlapping ACL memo lifecycle changes if this branch is updated.

One important difference: successful temp memos are not the only retained state. Timeout memos also need a stable destination `timeoutMemoKey`; otherwise unique temp names still accumulate. #840 releases temp memo state only after the temp is proven absent and retains it for residual files.

Ingwannu:
The confirmed Responses single-entry RAM-cap escape is now isolated in #841. Please exclude the overlapping state-admission portion when rebasing this branch.

#841 intentionally uses atomic admission and a normal continuation miss for an individually >64 MiB entry. It does not adopt the spill-store/app-memory framework here, keeping this review limited to the reproduced hard-cap violation.

Ingwannu:
The confirmed Antigravity replay slice is now isolated in #843. Please exclude the overlapping replay-cache changes when rebasing this branch.

#843 avoids an app-wide dependency and avoids a full 10,240-session expiration scan on every observe/apply. It uses fixed-size identity digests, per-call/session/global limits, LRU eviction, exact-session expiry, and at-most-once-per-minute background sweeps unless under budget pressure.

Ingwannu:
Cursor Connect inbound framing is now isolated in #844. This is a separate active-turn buffer bound rather than a process-lifetime cache change: 32 MiB header-time rejection, incremental pending completion, and incomplete-EOF failure. Please keep this overlap out of a future #829 rebase.

Ingwannu:
The confirmed Cursor blob-store slice is isolated in #845. Please exclude overlapping blob accounting/store changes on rebase.

The focused implementation uses request leases: advertised blob ids stay pinned through every terminal path (including async-generator cancellation), while only unpinned entries are TTL/LRU evictable. It also stores root candidates only after pruning and returns a typed `setBlobArgs` error on admission failure.

Ingwannu:
The confirmed tool-argument/SSE overlap has now been extracted into focused draft PR #847: 4 MiB SSE records, 8 MiB per call, 32 MiB per turn, truthful overflow failure, and no canonical event-contract redesign. It passes 206 related tests plus typecheck and privacy scan.

Please treat #847 as the reviewable implementation for this slice rather than carrying the overlapping giant-PR version forward. This does not clear the existing CHANGES_REQUESTED decision on #829; its unrelated review findings and broad state-store architecture still need separate resolution.

Issue #820, opened 2026-07-31
[architecture][memory] Make 32 concurrent tool-recall sessions protocol-safe and memory-bounded


lidge-jun:
## Summary

OpenCodex should remain TypeScript/Bun-based, but it needs an explicit concurrency and memory architecture for the real workload we must support:

- **32 sustained independent agent sessions**
- **64-session burst**
- repeated `model stream -> 1..N parallel tool/MCP calls -> tool results -> API recall`
- no OpenCodex-imposed serialization below that target
- no loss of Codex / Responses / Chat Completions / Anthropic / MCP compatibility

This is **not** a proposal to raw-passthrough provider bytes. OpenCodex is a semantic protocol translator. The required change is to keep that translation while replacing unbounded or duplicated retention with bounded state, byte accounting, and stream-lifetime leases.

Audit baseline: `dev@18545f87e92b856be180ba01e8b2c44d67163879`.

This issue is related to the field symptoms in #314 and #509, but is not a duplicate. Those issues discuss observed Windows memory behavior and imperfect Bun counters. This issue defines the architecture, hypotheses, implementation slices, and acceptance harness needed to distinguish OpenCodex-owned retention from Bun allocator/runtime retention under multi-session tool recall.

### Concurrency contract

```text
sustained independent sessions: 32
burst independent sessions:     64
global emergency active cap:     96
same-session active turns:        1
same-session queued turns:        0 by default
```

Different sessions must remain parallel. Only overlapping turns for the same logical session/thread may be rejected or serialized. Tool execution performed by the external client must not keep an OpenCodex model-turn lease occupied after the model stream reaches a terminal.

### Protocol invariants that must not be removed

Any memory fix must preserve all of the following:

1. Codex namespace/MCP tools are flattened for chat providers and restored as `{ namespace, name }` on the return path.
2. Codex custom/freeform tools such as `apply_patch` are exposed to chat models through a temporary `{ input: string }` function schema, then restored as `custom_tool_call`, not `function_call`.
3. `tool_search` remains distinct from an ordinary function call.
4. `call_id`, Responses `item_id`, `output_index`, and Chat Completions tool indexes remain stable and correctly correlated.
5. The complete lifecycle remains valid:
   `output_item.added -> arguments.delta* -> arguments.done -> output_item.done -> response terminal`.
6. No-argument calls finalize as `{}` rather than an empty invalid JSON string.
7. Malformed or truncated arguments must not be reported as a completed usable call.
[...]

PR #840, opened 2026-08-01
fix: release Windows ACL temp-path memos

Ingwannu:
## Summary

- release successful and timed-out Windows ACL memos once an ephemeral file is proven absent
- key atomic-write timeouts by the stable destination so unique temp names cannot accumulate
- cover config/response state, OpenAI tier backups, management tokens, and tray replacements
- retain memo state when a hardened residual temp remains on disk

Part of #820. This intentionally extracts the confirmed Windows leak into a focused change instead of coupling it to #829.

## Verification

- `bun run typecheck`
- `bun run privacy:scan`
- `bun test tests/windows-secret-acl.test.ts tests/config.test.ts tests/server-management-auth.test.ts tests/windows-tray.test.ts` (167 pass)
- full suite: 6427 pass / 6 skip; 5 unrelated GUI test import errors because this checkout lacks the GUI React dependencies (`react`, `react/jsx-dev-runtime`)

## Compatibility

Stable destination/directory memo behavior and fail-closed ACL policy are unchanged. Only one-shot temp path entries are released after rename/unlink or proven `ENOENT`.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -13,7 +13,12 @@ import {
   MAIN_CODEX_ACCOUNT_NAMESPACE_TARGET,
 } from "./codex/account-namespace-match";
 import { COMBO_NAMESPACE, comboConfigIssues } from "./combos/types";
-import { hardenSecretDir, hardenSecretPath, hardenSecretPathAsync } from "./lib/windows-secret-acl";
+import {
+  forgetEphemeralSecretPath,
+  hardenSecretDir,
+  hardenSecretPath,
+  hardenSecretPathAsync,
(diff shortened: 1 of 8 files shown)


lidge-jun:
Maintainer triage: this is a valuable focused fix. It removes the confirmed Windows process-lifetime temp-path memo leak while preserving fail-closed handling when a residual file remains, and the focused coverage is broad across the affected write paths.

The branch currently conflicts with `dev`, so it needs a rebase. After that, an independent review from someone other than the author is required before landing. We are tracking this under #820 for the next cycle.

lidge-jun:

PR #841, opened 2026-08-01
fix: enforce Responses state byte cap

Ingwannu:
## Summary

- measure retained continuation entries in serialized UTF-8 bytes before admission
- reject a single entry larger than the 64 MiB store budget instead of keeping it as the last survivor
- preserve unrelated valid chains when an oversized candidate is rejected
- remove stale same-ID state when its replacement cannot be admitted
- apply the same admission rule while loading snapshots

Part of #820. This is the focused in-memory bound; it intentionally does not pull in #829 durable spill architecture.

## Verification

- `bun test tests/responses-state.test.ts` (43 pass)
- `bun run typecheck`
- `bun run privacy:scan`

## Compatibility

An entry that individually exceeds 64 MiB now becomes a normal `previous_response_id` cache miss, so the client may need to resend history. Smaller stored chains and the existing disk snapshot limits are unchanged.

### DIFF
--- a/src/responses/state.ts
+++ b/src/responses/state.ts
@@ -27,7 +27,7 @@ interface StoredResponseState {
   /** v1 Cursor-only metadata, accepted only while loading old snapshots. */
   conversationId?: string;
   cursorCheckpointUsable?: boolean;
-  /** Approximate in-memory size, computed locally at insert time (never trusted from disk). */
+  /** Serialized UTF-8 entry size, computed locally at insert time (never trusted from disk). */
   sizeBytes?: number;
 }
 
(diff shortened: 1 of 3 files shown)


Ingwannu:
CI follow-up: the initial macOS failure was a Bun 1.3.14 Trace/BPT trap (the same bun.report signature also appeared on unrelated PR #844), and the initial Windows failure was an unrelated GUI polling timeout followed by cascading React act failures. Re-running only the failed jobs passed on both macOS and Windows. No production or test code change was needed for those runner flakes.

lidge-jun:
This is a valuable, focused fix: it closes the oversized-single-entry escape hatch in continuation-state storage, with appropriately narrow semantics and regression coverage. It currently conflicts with dev, so please rebase it onto the latest dev. Before landing, it will also need an independent review from someone other than the author. We’re tracking this under #820 for the next cycle.

lidge-jun:

PR #843, opened 2026-08-01
fix: bound Antigravity replay retention

Ingwannu:
## Summary

- replace raw model/session/function-argument cache keys with fixed-size SHA-256 identities
- cap replay at 64 KiB per signature, 256 calls and 2 MiB per session, and 32 MiB globally
- use call/session LRU eviction while keeping the existing one-hour TTL semantics
- preserve prior valid mappings when an oversized replacement is ignored
- sweep expired sessions at most once per minute outside budget pressure

Part of #820. This intentionally isolates the reproduced Antigravity retention problem from #829 app-wide memory infrastructure.

## Verification

- `bun test tests/google-antigravity-replay.test.ts tests/google-antigravity-wire.test.ts` (56 pass)
- `bun run typecheck`
- `bun run privacy:scan`

## Compatibility

Normal sequential tool loops keep the newest 256 signed calls. In an extreme longer loop, an evicted old call may receive the existing upstream signature error rather than remaining in memory without a bound.

### DIFF
--- a/src/adapters/google-antigravity-replay.ts
+++ b/src/adapters/google-antigravity-replay.ts
@@ -1,3 +1,5 @@
+import { createHash } from "node:crypto";
+
 /**
  * Antigravity (Cloud Code Assist) thoughtSignature reasoning-replay cache.
  *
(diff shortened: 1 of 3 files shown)


lidge-jun:
Thanks for this—it's a valuable, focused improvement. The byte-aware nested LRU bounds replace payload-sized replay keys and unbounded retention, and the substantial focused coverage gives this a solid foundation. The branch currently conflicts with dev, so it will need a rebase, followed by an independent review from someone other than the author, before it can land. We're tracking it under #820 for the next cycle.

lidge-jun:

PR #844, opened 2026-08-01
fix: bound Cursor Connect frame buffering

Ingwannu:
## Summary

- keep uint32-compatible outbound Connect frames while limiting inbound payloads to 32 MiB
- reject an oversized declared length as soon as the five-byte header arrives
- incrementally complete one pending frame instead of repeatedly concatenating it with unrelated following frames
- fail a turn when EOF leaves an incomplete Connect frame

Part of #820. The 32 MiB inbound limit matches the repository existing framing-hardening design note (`devlog/_fin/350_cursor-provider-add/116_phase33-connect-framing-robustness-plan.md`).

## Verification

- `bun test tests/cursor-framing.test.ts tests/cursor-hardening.test.ts tests/cursor-live-transport.test.ts` (45 pass)
- `bun run typecheck`
- `bun run privacy:scan`

## Compatibility

Normal split headers, split payloads, and multiple frames per chunk preserve their order. A legitimate inbound frame larger than 32 MiB now fails explicitly; outbound framing remains unchanged.

### DIFF
--- a/src/adapters/cursor/framing.ts
+++ b/src/adapters/cursor/framing.ts
@@ -1,7 +1,11 @@
 export const CONNECT_FRAME_HEADER_BYTES = 5;
 export const CONNECT_FLAG_COMPRESSED = 0x01;
 export const CONNECT_FLAG_END_STREAM = 0x02;
+/** Outbound wire encoding remains uint32-compatible; inbound buffering is intentionally tighter. */
 export const MAX_CONNECT_FRAME_PAYLOAD_BYTES = 0xffffffff;
+export const MAX_INBOUND_CONNECT_FRAME_PAYLOAD_BYTES = 32 * 1024 * 1024;
+export const MAX_INBOUND_CONNECT_FRAME_BYTES =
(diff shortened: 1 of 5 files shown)


Ingwannu:
CI follow-up: the initial macOS failure was the same Bun 1.3.14 Trace/BPT trap / bun.report signature seen on unrelated PR #841. The failed job passed on rerun, while the other platform jobs were already green. No code change was needed.

lidge-jun:
Maintainer triage: this is a narrow and valuable hardening change. It bounds inbound Cursor Connect frame buffering and rejects incomplete streams instead of allowing unbounded or ambiguous framing state.

The branch currently conflicts with `dev`, so it needs a rebase before it can land. It also shares the Cursor transport/docs surface with #845; please land the two PRs in a defined order (or consolidate them) to avoid overlapping changes and documentation drift.

Because this is co-maintainer-authored, an independent non-author review is still required before landing. Thanks for keeping the scope focused.

lidge-jun:

PR #845, opened 2026-08-01
fix: bound Cursor blob-store memory

Ingwannu:
## Summary

- cap Cursor blobs at 16 MiB each, 64 MiB total, and the existing 4,096 entries
- pin every root/step/turn blob advertised by an active request until success, failure, or generator cancellation
- evict only expired or LRU unpinned blobs; fail preparation when pinned data leaves no safe capacity
- return a protobuf error for rejected server `setBlobArgs` instead of acknowledging a store that did not happen
- store external root candidates only after replay pruning

Part of #820. This is a focused request-lifetime lease implementation rather than #829 app-wide memory infrastructure.

## Verification

- `bun test tests/cursor-*.test.ts` (398 pass)
- `bun run typecheck`
- `bun run privacy:scan`

## Compatibility

Content-addressed IDs and get/set protobuf shapes remain unchanged. A single blob above 16 MiB or a request whose unique pinned blobs cannot fit 64 MiB now fails explicitly rather than evicting an in-flight reference or retaining unbounded data.

### DIFF
--- a/src/adapters/cursor/live-transport.ts
+++ b/src/adapters/cursor/live-transport.ts
@@ -522,44 +522,48 @@ class LiveCursorTransport implements CursorTransport {
     const prepared = prepareCursorRunRequest(request, {
       estimateInputTokens: contextUsage.carryForwardTokens === undefined,
     });
-    state = createCursorProtobufEventState({
-      clientToolNames: clientToolDefs.map(tool => tool.toolName || tool.name),
-      parallelToolCalls: request.parallelToolCalls,
-      toolSchemas,
-      cursorToolNameMap,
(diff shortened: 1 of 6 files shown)


lidge-jun:
Maintainer triage status: this is valuable work. It bounds Cursor blobs while using scoped pins to protect active requests, has strong test coverage, and defines saturation behavior explicitly. It currently conflicts with `dev`, so a rebase is needed. It also shares the Cursor transport/docs surface with #844; please land the two in a defined order (or consolidate them) to avoid overlapping integration work. An independent, non-author review is required before landing.

lidge-jun:

PR #847, opened 2026-08-01
fix: bound streamed tool argument memory

Ingwannu:
## Summary

- cap shared SSE records at 4 MiB, including unterminated and multiline records
- cap retained tool arguments at 8 MiB per call and 32 MiB per turn across OpenAI Chat, the Responses bridge, and Chat outbound conversion
- fail oversized calls as typed upstream errors without emitting a completed tool item or clean Chat DONE marker
- document the limits and decision tradeoffs

## Why

Compatible upstreams can omit SSE delimiters or stream unbounded tool argument fragments. Those buffers were protocol-required, but they had no byte ownership limit and could grow for the lifetime of a turn. Truncating and completing a call would produce invalid or unsafe executable input, so overflow now fails truthfully while preserving the existing sequential tool event contract.

## Validation

- `bun test` on 15 related bridge, Chat, OpenAI Chat, and SSE suites: 206 pass
- `bun run typecheck`
- `bun run privacy:scan`

Related to #820. Focused alternative to the overlapping portions of #829.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -10,6 +10,8 @@ import { contentPartsToText } from "./image";
 import { neutralizeIdentity } from "./identity";
 import { buildNonOpenAIToolCatalogNudgeForTools, shouldInjectNonOpenAIToolCatalogNudge } from "./tool-catalog-nudge";
 import { openRouterProviderPayload, resolveOpenRouterRouting } from "../providers/openrouter-routing";
+import { MAX_SSE_RECORD_BYTES } from "../lib/sse-decoder";
+import { appendToolArguments, utf8ByteLength } from "../lib/tool-argument-bounds";
 
 // Providers may opt into stripping one trailing "[...]" group from the wire model id.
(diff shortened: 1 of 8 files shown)


Ingwannu:
CI follow-up: the first Windows run passed 6,514 tests and timed out only the unrelated existing image-normalization P2 case at 5.39 seconds against its 5-second test limit. Re-running the failed Windows job passed. Ubuntu, macOS, Windows, package-install, target-enforcement, label, and React checks are now green.

lidge-jun:
Thanks for this — it is a valuable part of the #820 series, especially because it bounds streamed tool-argument memory across the major translation paths. It currently conflicts with dev, so it will need a rebase, followed by an independent review from someone other than the author, before it can land.\n\nFor the broader draft series (#840, #841, #843, #844, #845, and #847), is there a specific landing order we should follow, or would it be better to consolidate any of the overlapping work? We are tracking #820 and would like to get these into the next cycle.

lidge-jun:

PR #850, opened 2026-08-01
fix(server): match browser extension CORS origins exactly

eachann1024:
## What changed

- canonicalize authority-based opaque origins as scheme + authority instead of the WHATWG `null` serialization
- allow only the configured browser extension ID while keeping other extensions and `*` rejected
- cover both preflight and `/v1/models` data-plane requests
- document browser-extension origins in the English and Simplified Chinese configuration references

## Why

`URL.origin` serializes `chrome-extension://...` URLs as `null`. Comparing that value made every browser extension look identical, while rejecting a wildcard configuration still left users without a safe way to allow one extension to call the local data plane.

## Change graph

```mermaid
flowchart LR
    Request["Browser extension request"] --> Parse["Parse request origin"]
    Allowlist["corsAllowOrigins"] --> Match["Exact origin matcher"]
    Parse --> Standard["Standard HTTP(S) origin"]
    Parse --> Opaque["Opaque authority origin"]
    Standard --> Match
    Opaque --> Authority["Compare scheme + authority"]
    Authority --> Match
    Match --> Decision{"Allowlisted?"}
    Decision -->|Yes| DataPlane["OpenCodeX data plane"]
    Decision -->|No| Rejected["403 origin_rejected"]
```

## Checks

- `bun test tests/server-loopback-host-gate.test.ts tests/server-auth.test.ts tests/management-origin-tls.test.ts` (76 passed)
- `bun run typecheck`
- `bun run lint:gui`
- `bun run privacy:scan`
- isolated real-server smoke: configured extension `OPTIONS /v1/models` = 204, `GET /v1/models` = 200, different extension = 403 `origin_rejected`
- full suite: 6504 passed, 4 skipped, 7 failed; the same 7 failures reproduce on clean `upstream/dev` because local Fake-IP DNS resolves `.test` / `.invalid` fixtures into `198.18.0.0/15` and the destination policy rejects them (`tests/provider-outbound.test.ts`, `tests/management-provider-validation.test.ts`)

## Checklist

## Summary by CodeRabbit

* **New Features**
  * CORS allowlists now support browser extension origins such as `chrome-extension://<extension-id>`.
  * Configured extension origins are accepted for preflight and model requests, while unconfigured or different extension origins remain blocked.
  * Origin matching now handles valid custom-scheme authorities consistently while preserving strict behavior for invalid values.

* **Documentation**
  * Updated configuration guidance across supported languages to clarify exact-origin matching and that `*` is not a wildcard.

### DIFF
--- a/docs-site/src/content/docs/ja/reference/configuration.md
+++ b/docs-site/src/content/docs/ja/reference/configuration.md
@@ -63,7 +63,7 @@ namespaced selected id を bare id に変えます。
 | `webSearchSidecar?` | `OcxWebSearchSidecarConfig` | on | ウェブ検索サイドカーオプション（下記参照）。 |
 | `visionSidecar?` | `OcxVisionSidecarConfig` | on | ビジョンサイドカーオプション（下記参照）。 |
 | `tokenGuardian?` | `OcxTokenGuardianConfig` | off | 選択型の proactive OAuth 更新と Codex アカウント warmup ポリシー。フィールドは下で説明します。 |
-| `corsAllowOrigins?` | `string[]` | `[]` | CORS で追加で許可する正確な origin。loopback origin は常に許可します。 |
+| `corsAllowOrigins?` | `string[]` | `[]` | CORS で追加で許可する正確な origin。loopback origin は常に許可します。`chrome-extension://<extension-id>` など、authority ベースのブラウザー拡張機能の origin に対応しています。`*` はワイルドカードではありません。 |
 
 `codexAccountNamespaces` のキーは公開 selector です。長さは 1〜64 文字、先頭と末尾は ASCII
 英数字、内部には英数字、`.`、`_`、`-` を使用でき、予約済み JavaScript object 名は拒否されます。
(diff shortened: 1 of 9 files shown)


lidge-jun:
Thanks for the focused fix here. The CORS origin matching change and its security regression tests look focused and correct.

The one blocker is documentation drift: the localized configuration references were not updated consistently with the English change. Please bring these three files in line with docs-site/src/content/docs/reference/configuration.md, including browser extension origin support and the clarification that * is not a wildcard:

- docs-site/src/content/docs/ja/reference/configuration.md
- docs-site/src/content/docs/ko/reference/configuration.md
- docs-site/src/content/docs/ru/reference/configuration.md

Once those three locales are synced, this is mergeable. Quick re-review once locales sync.

eachann1024:
Thanks for the review. I synced the `corsAllowOrigins` guidance in the Japanese, Korean, and Russian configuration references in a346ad60. Each locale now documents authority-based browser extension origins such as `chrome-extension://<extension-id>` and clarifies that `*` is not a wildcard.

Verification:
- `cd docs-site && bun install --frozen-lockfile && bun run build`
- CodeRabbit re-review: no actionable comments
- Ubuntu and macOS CI: passed
- Windows CI reached a Bun 1.3.14 internal assertion failure after running the test suite; the log identifies it as a Bun crash rather than an application test failure.

The PR is mergeable without conflicts. A quick re-review would be appreciated.

lidge-jun:
Superseded by #886, which applies this implementation verbatim (authorship credited) onto current dev, plus a management-plane preflight test and a re-base of the locale docs onto the new `configuration/server.md` subpages (the doc hunks here no longer apply after the docs split). Closing in favor of that PR — thank you for the fix and the thorough test shape.

PR #860, opened 2026-08-01
fix(responses): gate service tiers by provider capability

F1Justin:
## Summary

- Add a provider-level `supportsServiceTier` capability for Responses routes.
- Mark the canonical OpenAI Responses providers as supporting service tiers and DeepSeek as not supporting them.
- Let `fastMode` inject or remove `service_tier` only for providers that explicitly support it, and strip the field for providers that explicitly reject it.
- Preserve caller-supplied values for unclassified custom Responses providers.
- Keep the capability as registry-enriched runtime metadata so older canonical OpenAI configs remain valid.
- Document the capability and the DeepSeek behavior in every maintained locale.

## Why

`fastMode` previously treated every `openai-responses` destination like OpenAI and could inject `service_tier: "priority"`. DeepSeek V4 Flash now uses its native Responses endpoint by default, but DeepSeek's published Responses schema does not document `service_tier`. Provider transport and service-tier capability need to be separate decisions.

This change keeps native DeepSeek Responses routing intact while preventing an OpenAI-only request field from reaching that upstream.

DeepSeek reference: https://api-docs.deepseek.com/api/create-response/

## Verification

- `bun test tests/deepseek-inbound-wire.test.ts` — 19 pass.
- Focused canonical OpenAI management checks — 3 pass after keeping the new capability out of persisted canonical seeds.
- `bun run typecheck` — pass.
- `bun run privacy:scan` — pass.
- `cd docs-site && bun run build` — pass (146 pages).
- Full-suite attempt before the final canonical-seed compatibility adjustment: 6,786 pass, 7 skip, 10 fail. The three patch-caused canonical OpenAI failures were fixed and pass in the focused rerun. The remaining affected-file rerun is 37 pass / 7 fail because this machine's Clash fake-IP DNS resolves test-only hosts such as `*.example.test` and `proxy-only.invalid` into `198.18.0.0/15`, which the destination-policy tests intentionally reject. CI should provide the clean non-fake-IP run.

## Scope

- No authentication, credential, destination, or provider selection behavior changes.
- No request bodies, API keys, or account identifiers are logged.
- Explicit per-provider configuration remains backward compatible.

## Summary by CodeRabbit

* **New Features**
  * Added provider-level configuration to declare `service_tier` support.
  * Fast-mode handling now applies only to providers that explicitly support service tiers.
  * Unsupported or unspecified providers remove client-supplied values and do not inject fast-mode settings.
  * DeepSeek V4 Flash and Volcengine Agent Plan routes explicitly exclude this field.

* **Documentation**
  * Updated provider configuration and fast-tier guidance across supported languages.

### DIFF
--- a/docs-site/src/content/docs/guides/codex-app-models.md
+++ b/docs-site/src/content/docs/guides/codex-app-models.md
@@ -120,8 +120,11 @@ fast_mode = true
 ```
 
 But the model catalog and runtime request tier id use `priority`. opencodex preserves that split.
-Native OpenAI passthrough models keep fast support; routed non-OpenAI models strip service-tier
-metadata so the fast option is not advertised where it cannot be honored.
+Native OpenAI passthrough models keep fast support. Responses providers only receive proxy-generated
+OpenAI-specific `service_tier` when their registry capability explicitly opts in; an omitted
+capability fails closed — a caller-supplied value is stripped and fastMode never injects one.
+DeepSeek V4 Flash's and Volcengine Agent Plan's native Responses routes are explicitly marked
+unsupported, so the field stays out of their upstream requests.
 
 ## Subagent selection
 
--- a/docs-site/src/content/docs/ja/guides/codex-app-models.md
+++ b/docs-site/src/content/docs/ja/guides/codex-app-models.md
@@ -87,6 +87,7 @@ fast_mode = true
 ```
 
 ただし、モデル カタログとランタイム リクエスト層 ID は `priority` を使用します。 opencodex はその分割を保持します。ネイティブ OpenAI パススルー モデルは高速サポートを維持します。ルーティングされた非 OpenAI モデルはサービス層メタデータを削除するため、高速オプションが受け入れられない場合はアドバタイズされません。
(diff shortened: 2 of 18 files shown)


lidge-jun:
Pre-merge review result: **not merge-ready yet**. Two blocking items:

1. **Unknown capability fails open.** `src/server/responses/core.ts:805` strips `service_tier` only when `supportsServiceTier === false`; `undefined` passes the caller's value through. The test at `tests/deepseek-inbound-wire.test.ts:289` locks in that unsafe default. An unknown-capability provider must strip the OpenAI-only field unless explicitly opted in.
2. **A built-in Responses provider is exposed to that default.** `src/providers/registry.ts:1186` defines `volcengine-agent-plan` as `openai-responses` with no capability classification; a probe against the PR head captured `"service_tier":"priority"` forwarded upstream. Please classify it (or it inherits the safe default from fix 1).

Also needed before merge:

3. The strip tests can pass without any upstream request (`tests/deepseek-inbound-wire.test.ts:219` returns `bodies[0] ?? {}`); assert a successful response and exactly one captured body.
4. Add a wire regression for the canonical native `openai` Codex-login route (currently only `openai-apikey` is covered).
5. Full cross-platform CI never ran on the current head (workflow shows `action_required`, zero jobs) — it must run green per merge policy. The five doc locales will need to follow whatever default semantics you choose.

The direction is right and the fix is close — re-review should be fast once the default flips to fail-closed.

lidge-jun:
Superseded by #889, which adopts this PR's reviewed fail-closed semantics (`true` keeps fast-mode inject/remove, `false`/`undefined` strip + never inject, explicit config wins) and its file-map shape, with two adjustments found during review: the capabilities are registry-only (never seeded — the management API's canonical-seed exact-key comparison breaks otherwise), and both open review issues are addressed (registry-backfill test instead of hardcoded canonical config; localized docs no longer carry the blanket wording). It also fixes the separate #875 reasoning-replay root cause in the same lane. Thanks for the capability design — it survived review intact.

PR #868, opened 2026-08-01
fix(windows): settle scheduler registration verification

ventianima-lab:
## Summary
- retry only transient post-create Task Scheduler visibility/XML health states
- preserve fail-closed behavior for conflicts, missing assets, and unknown SCM status
- stop late reconciliation when attempt ownership changes

## Verification
- 136 focused tests passed across scheduler, startup, service, and install-verification contracts
- bun run typecheck passed
- live Windows startup protection reports the service installed, viable, running, and conflict-free

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows Task Scheduler setup reliability by retrying transient registration states for a limited period.
  * Immediate failures are now handled for conflicts, missing assets, and unknown service status.
  * Prevented changes when setup ownership is lost, reducing the risk of unintended rollback or state updates.

* **Tests**
  * Added coverage for retries, rollback scenarios, persistent failures, and interrupted setup ownership.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -664,6 +664,8 @@ type FinalizeHooks = {
     deleteArgs: string[],
   ) => Promise<ElevatedSchtasksCreateAndRunResult>;
   verify?: () => WindowsSchedulerInstallVerification;
+  /** Test-only replacement for the bounded post-registration settle delay. */
+  settleDelay?: (milliseconds: number) => Promise<void>;
   writeInstallState?: () => void;
   /** Preferred tri-state probe for security-sensitive reconciliation. */
   probeTask?: () => WindowsSchedulerTaskProbe;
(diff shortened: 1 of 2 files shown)


lidge-jun:
Thanks for tightening up the Windows scheduler verification path. Two remaining asks before this is ready:

1. Please add the ownership-loss edge test from the open CodeRabbit thread. Keep the post-verification `attemptStillOwned` guard, add the focused coverage in `tests/windows-elevation-spawn.test.ts` that changes ownership from the final `verify` hook, and assert that neither rollback nor `writeInstallState` is invoked. The current ownership-loss test flips ownership from `settleDelay`, so it does not cover this await-resumption edge.
2. Please run the full platform CI matrix on the current head (`5309bf4`). At the moment, only the target/label checks and CodeRabbit have run.

happy to re-review quickly once both are in.

lidge-jun:
Superseded by #887, which lands this fix on current `dev`.

Thank you for the diagnosis — the root cause you identified is real and the direction was right. I re-derived the patch rather than merging this one because an independent audit against current `dev` found two gaps worth closing before this ships:

1. **Proven WinSW presence needs to block retry independently of `conflict`.** `conflict` only becomes true once the task itself is visible (`src/service.ts`, `evaluateWindowsSchedulerInstallVerification`). While the task is still invisible beside a running WinSW, the pair is `conflict: false` with `nativeServiceAbsent: false` — a retry predicate checking only `!conflict` would wait for a service already proven present. #887 rejects that state explicitly and pins it with a dedicated regression.

2. **The pre-rollback ownership guard had no test that actually fires it.** The settle loop never awaits on a non-retryable verdict, so flipping ownership inside a successful final `verify` only re-proves the pre-existing write guard. #887 adds a case where `verify` returns a non-retryable failure *and* revokes ownership, which is the only path that reaches the new fence.

Everything else follows your patch closely: the bounded `[50, 150, 300, 600]` backoff, the injectable settle delay, and the fail-closed preservation for conflicts, missing assets, and unknown SCM status.

One note on evidence: I could not run live Windows validation in this session, so #887 covers the startup-protection claim with fault-injection contracts and says so in the plan document rather than restating an unverified smoke result.

Closing in favor of #887. Credit for the root cause is yours.

PR #869, opened 2026-08-01
fix: preserve symlinked destinations in atomic config writes

nicosuave:
## Summary

atomicWriteFile and atomicWriteFileAsync wrote a temp file beside the literal destination path and renamed over it. rename(2) replaces a directory entry, so when the destination was itself a symlink the rename destroyed the link and left a plain file in its place.

This breaks dotfiles-managed setups, where ~/.codex/config.toml is a symlink into a tracked repo. The first injected write silently converts it to a real file and the repo stops receiving updates. Nothing surfaces the divergence: the live config keeps working, so the stale repo copy looks current until someone diffs it.

Resolve the destination through realpath before choosing the temp path. Both the temp file and the rename target then live inside the link's real directory, so the entry replaced is the real file and the link survives. Same-filesystem atomicity is preserved because the temp stays beside its resolved target, and an unresolvable path falls back to the literal path, which is correct for a first write and for a dangling link.

## Verification

Ran tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed atomic configuration writes to preserve symbolic links while updating their underlying targets.
  * Ensured temporary files are cleaned up after writes, including during response-state recovery.
  * Improved handling of new files, regular destinations, and dangling symbolic links.
  * Prevented writes through configuration links from modifying protected files.
  * Improved recovery of stale temporary response-state files when snapshots use symbolic links.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,6 +1,6 @@
 import { execFileSync } from "node:child_process";
 import { randomUUID } from "node:crypto";
-import { chmodSync, copyFileSync, existsSync, linkSync, mkdirSync, readFileSync, renameSync, truncateSync, unlinkSync, writeFileSync } from "node:fs";
+import { chmodSync, copyFileSync, existsSync, linkSync, mkdirSync, readFileSync, realpathSync, renameSync, truncateSync, unlinkSync, writeFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { join, resolve } from "node:path";
 import { Database } from "bun:sqlite";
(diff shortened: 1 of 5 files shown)


lidge-jun:
Superseded by #886, which applies this implementation (authorship credited) plus one amendment from an independent audit: `resolveWriteTarget` no longer falls back to the literal path when an EXISTING symlink cannot be resolved (dangling target / unmounted volume / ELOOP / EACCES) — that case is now refused and the link preserved, instead of silently recreating the divergence this PR fixes. Both dangling-link tests updated accordingly. Closing in favor of that PR — thank you for the careful fix, including the guard re-application and the sweep-both-dirs catch.

PR #871, opened 2026-08-02
fix(update): allow stable updates from older previews

luvs01:
## Summary

- Compare an installed preview by its `major.minor.patch` core when the selected update channel is `latest`.
- Keep the latest-channel target strict: a preview registry target is still never accepted as a stable update.
- Cover both the comparator and the GUI/API update-check result that controls one-click availability.

## Problem

An older preview install such as `2.8.2-preview.20260731` cannot be parsed by the stable-only current-version branch. With npm `latest` at `2.9.1`, `isNewer()` therefore returns `false`, and the dashboard incorrectly reports `already_latest` and disables one-click update.

## Behavior

- `latest=2.9.1`, `current=2.8.2-preview.20260731`: update available.
- `latest=2.9.1`, `current=2.9.1-preview.20260731`: no update, preserving the existing core-version comparison policy.
- A preview value in the `latest` target remains rejected.

## Verification

- `bun test tests/update-notify.test.ts tests/update-job.test.ts`: 60 passed, 0 failed, 152 assertions.
- `bun run typecheck`: passed.
- `git diff --check upstream/dev...HEAD`: passed.

No documentation change is included because this restores the documented existing update-channel behavior rather than adding a new user-facing option.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update checks for preview versions, allowing older previews to detect newer stable releases.
  * Prevented update notifications when a preview already matches the stable release’s base version.

* **Tests**
  * Added coverage for preview-to-stable update detection and matching-version behavior.

### DIFF
--- a/src/update/notify.ts
+++ b/src/update/notify.ts
@@ -92,7 +92,7 @@ function gt(a: number[], b: number[]): boolean {
 export function isNewer(latest: string, current: string, channel: Channel): boolean {
   if (channel === "latest") {
     const l = parseStable(latest);
-    const c = parseStable(current);
+    const c = parseStable(current) ?? parsePreview(current)?.slice(0, 3);
     if (!l || !c) return false;
     return gt(l, c);
   }
--- a/tests/update-job.test.ts
+++ b/tests/update-job.test.ts
@@ -85,6 +85,24 @@ describe("GUI update check", () => {
     expect(result.canUpdate).toBe(false);
     expect(result.reason).toBe("already_latest");
   });
+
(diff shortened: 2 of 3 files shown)


lidge-jun:
Superseded by #892, which lands this fix natively (installed previews now compare by their major.minor.patch core on the latest channel; the registry-target side stays strict). The comparator change, the GUI/API availability path, and the semver §9 / respin-format limitations are all covered with regression tests and an independent review. Thank you — the report and the initial patch drove the fix.

Issue #311, opened 2026-07-23
Shadow call intercept no longer matches Codex 0.145.0 shadow model (gpt-5.6-luna)


nieaowei:
## Problem

The shadow call intercept currently only matches `gpt-5.4-mini`, but the latest Codex CLI (0.145.0) has changed the background helper/shadow model from `gpt-5.4-mini` to `gpt-5.6-luna`, causing the intercept to stop working.

## Evidence

Extracted hard-coded model strings from the Codex binary via `strings`:

**Installed version 0.144.6** (`/opt/homebrew/lib/node_modules/@openai/codex/.../bin/codex`):
```
codex-auto-reviewgpt-5.4-minigpt-5.4
```
Shadow/helper calls use `gpt-5.4-mini`.

**Latest version 0.145.0** (pulled from npm `@openai/codex@0.145.0-darwin-arm64`):
```
codex-auto-reviewgpt-5.6-lunagpt-5.6-terra
```
Shadow/helper calls now use `gpt-5.6-luna` (and `gpt-5.6-terra`).

## Affected code

`src/server/responses.ts:836-849` — intercept condition hard-coded to `gpt-5.4-mini`:
```typescript
// Shadow call intercept: rewrite Codex Desktop's hard-coded gpt-5.4-mini helper calls
const _sci = config.shadowCallIntercept;
if (_sci?.enabled && _sci.model && parsed.modelId.startsWith("gpt-5.4-mini")) {
```

`src/types.ts:426-433` — type annotation and field docs also reference `gpt-5.4-mini`:
```typescript
/**
 * Shadow call intercept: redirect Codex Desktop's hard-coded gpt-5.4-mini helper calls
 * (title generation, commit messages, skill orchestration) to a user-chosen model.
 */
shadowCallIntercept?: {
  /** When true, all gpt-5.4-mini* requests are rewritten to the configured model. */
  enabled?: boolean;
  /** Replacement model id (e.g. "gpt-5.5"). */
  model?: string;
};
```

`src/providers/openai-tiers.ts:106,124-125,168` — migration/rewrite logic for `shadowCallIntercept.model` also implicitly assumes `gpt-5.4-mini`.

## Expected behavior

The intercept should cover both `gpt-5.4-mini` and `gpt-5.6-luna` (and possibly `gpt-5.6-terra`) to support old and new Codex versions alike.

## Suggested fix (optional)

[...]

PR #885, opened 2026-08-02
fix(shadow-call): preserve foreground Luna turns

kimrinking-cell:
## Summary

- Preserve foreground Codex turns when the selected bare model is also listed as a shadow-call source.
- Use `x-codex-turn-metadata.request_kind` to bypass interception only for explicit `turn` requests.
- Keep the existing opt-in behavior for helper requests, headerless clients, and unknown or malformed metadata.
- Add regression coverage and document the compatibility boundary.

## Root cause

The issue #311 follow-up expanded the default source prefixes to include `gpt-5.6-luna`. The runtime still classified requests only by model prefix, so a user-selected Luna foreground turn could be rewritten to the configured shadow model and have its reasoning effort forced to `low`.

Codex 0.145.0 and later project the request purpose through `x-codex-turn-metadata`. This change uses the explicit `request_kind: "turn"` signal while preserving the prior all-matching contract when that signal is absent or unrecognized.

Related: #311

## Compatibility

- `turn`: preserve the requested model and effort.
- `prewarm`, `compaction`, and `memory`: keep shadow interception.
- Headerless, malformed, or future unknown metadata: keep the legacy prefix behavior.
- Routed model ids and non-source models remain unchanged.
- No config migration or schema change is required.

## Smaller UI-only alternative

If maintainers prefer to keep the broad runtime contract unchanged, the smallest mitigation would be a warning/confirmation dialog when enabling Shadow Call Intercept or when a selected foreground model overlaps `sourceModels`. The dialog should state that matching foreground requests can be redirected and forced to low effort.

That warning would help dashboard users, but it would not protect config-file users or clients that change models later, so this PR keeps the runtime classification fix focused and does not mix in GUI changes.

## Verification

Validated locally on macOS only:

- `bun test tests/responses-shadow-intercept.test.ts` — 14 passed
- `bun test --isolate tests` — 6,964 passed, 8 skipped, 0 failed
- `bun run typecheck`
- `bun run privacy:scan`
- `cd gui && bun test tests` — 493 passed
- `cd gui && bun run lint`
- `cd gui && bun run build`
- `git diff --check`

Linux and Windows behavior has not been claimed from local testing; the repository CI matrix should validate those platforms for this commit.

## Checklist

- [x] Docs were updated for the behavior change.
- [x] No authentication, credential, dependency, or workflow boundary was changed.

### DIFF
--- a/docs-site/src/content/docs/reference/configuration/server.md
+++ b/docs-site/src/content/docs/reference/configuration/server.md
@@ -120,6 +120,9 @@ subscription with a warning when detection is inconclusive. See
 Codex uses small helper models for tasks such as titles and commit messages. Enable
 `shadowCallIntercept` to redirect recognized source-model prefixes to another configured model. The
 replacement runs at low effort. Set `sourceModels` only when a client uses different helper ids.
+Codex 0.145.0+ marks request purpose in `x-codex-turn-metadata`: normal `request_kind: "turn"`
+requests keep the selected model, while recognized maintenance requests can be redirected. Clients
+without that metadata retain the legacy prefix behavior.
 
 ```json
 {
--- a/docs/shadow-call-intercept.md
+++ b/docs/shadow-call-intercept.md
@@ -63,14 +63,18 @@ the defaults rather than extending them:
 
 ### Behavior
 
-- When enabled, ALL requests whose bare model id starts with one of the source-model prefixes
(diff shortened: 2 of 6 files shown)


lidge-jun:
Landed in #892 as ffc24bc98 (authorship preserved). The fail-closed semantics (headerless, malformed, or future-unknown metadata all keep legacy prefix interception) passed adversarial review. Closing as superseded — nice, minimal fix.

PR #886, opened 2026-08-02
fix(server,config): extension CORS origin identity + symlink-safe atomic writes

lidge-jun:
## Summary

Two must-fix bugs on the server/config boundary, implemented in the wt4 worktree lane (unit: `devlog/_plan/260802_wt4_server_config_security/`). Supersedes and closes #850 and #869 — both contributor implementations are applied here with authorship credited, plus two amendments folded in from independent audits.

### Bug A — browser-extension CORS origins all collapse to `null` (supersedes #850, @eachann1024)

`URL.origin` serializes every extension scheme (`chrome-extension://`, `moz-extension://`, `safari-web-extension://`) as `"null"` per WHATWG URL §4.7, so `isExtraAllowedOrigin`'s origin-equality check admitted **any** browser extension whenever one was allowlisted.

- `comparableOrigin()` keeps WHATWG origins for normal schemes and compares `protocol//host` for authority-based opaque origins; hostless opaque origins keep the exact-string fallback (literal `Origin: null` behavior unchanged).
- One shared predicate covers both planes: data-plane `/v1/*` and management `/api/*` (preflight + GUI session issuance).
- **Beyond the PR:** management-plane preflight assertions (`/api/settings` accept + cross-extension reject) folded from the wt4 audit; locale docs re-based onto the `configuration/server.md` subpages (docs split in 7fdb2cb8e) with the Firefox/Safari UUID-rotation caveat (per-install / per-launch regeneration — MDN, Mozilla bug 1717671, WebKit bug 244330).

### Bug B — atomic config writes destroy symlinked destinations (supersedes #869, @nicosuave)

`atomicWriteFile`/`atomicWriteFileAsync` wrote the temp beside the literal path and renamed over it; rename(2) replaces the directory entry, so a symlinked destination (dotfiles-managed `~/.codex/config.toml`) was silently converted to a plain file and the tracked repo stopped receiving writes (POSIX.1-2024; Linux `rename(2)`: "the link will be overwritten").

- `resolveWriteTarget()` realpaths the destination so temp and rename land beside the real file; the test-only real-home guard is re-applied to the resolved target; the responses-state load sweeps stale temps in both the literal and the resolved directory.
- **Beyond the PR (audit amendment):** a realpath failure no longer falls back to the literal path blindly. A genuinely absent destination keeps the literal first-write path, but an **existing unresolvable symlink** (dangling target, unmounted volume, ELOOP, EACCES) is refused and preserved instead of silently replaced; both dangling-link tests assert rejection + link survival.

## Test plan

- [x] `bun run typecheck` — green
- [x] Full suite `bun run test` — **7013 pass, 0 fail** (macOS arm64, Ran 7023)
- [x] Windows focused: `server-auth` + `server-loopback-host-gate` + `config` — **172/172** (win32, real node)
- [x] Windows Bug-B files: 191 pass; 11 fails all `EPERM` symlink-privilege of the test node (no admin), zero functional failures
- [x] `bun run privacy:scan` — passed
- [x] New regressions: extension preflight/data-plane/management accept+reject, cross-scheme isolation, `*` rejected; symlink survival (lstat/readlink/target-updated, sync+async), no temp left, plain destination byte-identical, first-write creation, dangling-link refusal, resolved-dir temp sweep, real-home escape refusal

## Security note

Touches the CORS origin admission boundary (`src/server/auth-cors.ts`) and credential-adjacent file writes (`src/config.ts`, `src/oauth/store.ts` callers) — per MAINTAINERS.md this wants explicit security review. No secrets are logged; the allowlist model stays exact-match (no wildcard relaxation).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CORS handling for browser extensions, including exact origin matching and clearer rejection of unsupported origins.
  * Preserved symlink targets during configuration and response-state writes, while preventing unsafe writes through protected paths.
  * Improved recovery of temporary response-state files.

* **Documentation**
  * Clarified CORS configuration behavior across supported languages.
  * Added planning documentation for future integration toggles and Codex prompt customization.

* **Tests**
  * Expanded coverage for CORS authorization, symlink-safe writes, recovery behavior, and prompt-layer validation.

### DIFF
--- a/devlog/_plan/260802_client_toggle_api/000_plan.md
+++ b/devlog/_plan/260802_client_toggle_api/000_plan.md
@@ -116,6 +116,10 @@ For each of Hermes Agent, OpenClaw, Kimi Code CLI, Gajae Code:
 - `002_client_toggle_matrix.md` — Q2 findings, one section per client.
 - `003_api_design_options.md` — Q3 design space + risk register + open
   questions. Options, not a commitment.
+- `004_ux_design.md` — (cycle 2) GUI design for the unified integrations
+  surface: tab rename, hero with install detection + switches, per-client
+  sub-pages, and the rollback UX that makes the toggle trustworthy. Design
+  spec only; component diffs belong to a later implementation cycle.
 
 ## Dispatch plan
 
--- a/devlog/_plan/260802_client_toggle_api/000_plan.md
+++ b/devlog/_plan/260802_client_toggle_api/000_plan.md
@@ -1,0 +1,130 @@
+# 000 — Plan: client-integration toggle research (API-side on/off switch)
+
+Research unit. Docs-only: this unit produces survey and design-option documents,
+not production code. Implementation, if approved later, gets its own decade docs
+in a follow-up cycle.
+
+Parent unit: `devlog/_plan/260731_client_config_export/` (the read-only export
+surface this unit extends). Sibling survey `001_client_config_survey.md` there
+already covers Pi/OpenCode injection layers and the no-standard landscape; this
+unit does not re-survey them.
+
+## Loop spec
+
+- Loop archetype: spec-satisfaction research (verifier defines done).
+- Trigger: user request — can client integrations be applied as an API-side
+  on/off switch (enable writes the provider block into the client config,
+  disable removes it), for Hermes Agent, OpenClaw, Kimi Code CLI, and
+  Gajae Code, with cc-switch as the reference implementation.
+- Goal: durable devlog evidence answering (a) how cc-switch implements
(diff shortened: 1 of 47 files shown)


lidge-jun:
Superseded by #892, which lands both fixes natively: extension CORS origins are matched by scheme+authority (configured extension ID only, wildcard rejected, all ten data-plane sites covered) and atomic writes resolve symlinked destinations before the temp+rename dance. The unified branch additionally closes two guard gaps the stacked review caught: first writes beneath a symlinked parent are refused, and pid/port writers guard before any mkdir/chmod. Thank you — the original reports and patches drove the fix.

PR #887, opened 2026-08-02
fix(windows): settle scheduler verification and record Bun runtime provenance

### LINKED ISSUE #848: [Bug]: doctor repeats OPENCODEX_BUN_PATH guidance when a Bun canary override is already active
## Client or integration

OpenCodex CLI (`ocx doctor`, npm packaged install)

## Area

CLI / Bun runtime diagnostics

## Summary

`ocx doctor` can tell Windows users to set `OPENCODEX_BUN_PATH` even when the running service already uses that override. This makes the guidance circular and obscures the difference between the bundled runtime and an explicitly selected, unvalidated canary.

## Reproduction

1. Install `@bitkyc08/opencodex@2.8.2-preview.20260731` on Windows.
2. Set `OPENCODEX_BUN_PATH` to Bun `1.4.0-canary.1+5f65d3785` and reinstall the service.
3. Confirm `ocx status` reports `Runtime source: override (OPENCODEX_BUN_PATH)`.
4. Run `ocx doctor`.

It reports `Bun 1.4.0`, `auto-known-bad`, then suggests setting `OPENCODEX_BUN_PATH` to a trusted runtime.

The same wording remains on current `dev`. `src/cli/doctor.ts` explicitly notes that the endpoint cannot distinguish a bundled binary from an override with the same version. The service already reports `bunRevision`; the installed revision is 941 commits ahead of, and 0 behind, the merge commit of [oven-sh/bun#32120](https://github.com/oven-sh/bun/pull/32120). This does not require OpenCodex to trust every canary automatically, but it demonstrates why version-only wording is incomplete.

## Expected behaviour

Keep the conservative stream policy, but make diagnostics runtime-origin aware:

- bundled known-bad runtime: retain the override suggestion;
- active override: say the override is active but remains unvalidated for automatic eager relay, without suggesting the already completed action.

Including `bunRevision` in the displayed identity would make reports actionable.

## Regression coverage

Cover both bundled and override runtimes with the same reported version. The bundled case should offer `OPENCODEX_BUN_PATH`; the override case should identify the active override and not repeat that instruction.

## Version

`2.8.2-preview.20260731` (`gitHead a5f48e31052fd28547bf383ebc47c9303fa74eaa`)

## Operating system

Windows 11 Pro 25H2, build 26200.8875, x64

## Checks

- [x] Reproduced with the packaged preview and live service.
- [x] Confirmed the proxy process uses the configured canary executable.
- [x] Checked current `dev` and searched existing issues/PRs.
- [x] Related to #721/#734, but runtime selection itself is working correctly.

lidge-jun:
Maintainer takeover of #868 and #861 (issue #848). Both were re-derived against current `dev` rather than cherry-picked — the audit found real gaps in each original patch, listed below.

## Bug A — post-create scheduler verification never settles (#868)

`applyElevatedSchedulerResult()` verified exactly once and rolled back immediately. After a successful elevated `/create` + `/run`, the non-elevated `/query /tn` and its CSV fallback can both miss the just-created task, so `probeWindowsSchedulerTask()` returns `absent`, `windowsSchedulerTaskInstalled()` collapses that to `false`, and the evaluator reports "Task Scheduler task is not installed." A healthy registration was deleted milliseconds before Task Scheduler would have listed it.

Verification now re-checks on a bounded 1.1s backoff (`[50, 150, 300, 600]`), but only while the failure looks like a lagging view: the task is not visible yet, or it is visible without a fully published registration. Every other verdict keeps its meaning and spends no delay at all.

Two things the original PR did not cover:

- **Proven WinSW presence is rejected independently of `conflict`.** `conflict` only becomes true once the task itself is visible, so while the task is still invisible beside a running WinSW the pair is `conflict: false` with `nativeServiceAbsent: false`. A predicate checking only `!conflict` would wait for a service already proven present. There is a dedicated regression for exactly that state.
- **Rollback now takes the same ownership fence the state write already had.** Rollback deletes a real task, so a stale attempt could otherwise delete a task a newer attempt owns. The settle loop never awaits on a non-retryable verdict, so this fence needed its own activation test (`ownership lost around a non-retryable failure`) rather than the mid-delay case.

Eight regressions in `tests/windows-elevation-spawn.test.ts` cover the settle path and each fail-closed class; four fail without the source change.

Live Windows validation was not available in this session. The PR body's startup-protection smoke claim is covered by fault-injection contracts instead, and the plan document says so rather than implying otherwise.

## Bug B — doctor repeats `OPENCODEX_BUN_PATH` when the override is already active (#861 / #848)

The remediation branch at `src/cli/doctor.ts` fired on `platform === "win32" && eagerRelay.reason === "auto-known-bad"` and always printed "set `OPENCODEX_BUN_PATH` to a runtime you trust". The payload it reads carried no runtime origin at all, so it could not tell an active override from bundled or process execution.

Origin cannot be recovered after the fact: resolving it at report time answers *what would this shell pick now*, which is a different question from *what was the service started with*, and the two diverge exactly when it matters. So the selecting launcher stamps it.

Scope grew twice during review:

- **Seven launch paths, not five.** The Codex autostart shim and the Windows tray host both relaunch the proxy themselves and were erasing provenance on the most common Windows start. Covered alongside the npm launcher, scheduler wrapper, WinSW, launchd, and systemd.
- **`process.execPath` relaunches too.** `ocx ensure`, GUI/Claude/OpenCode start, `POST /api/system/restart`, and the update relaunch copied the parent environment and handed the daemon nothing. `withProcessRuntimeProvenance()` covers those.

Two design corrections came out of the audit loop:

- **The shims scope the marker to their `ensure` invocation** (assignment prefix in `sh`, `setlocal`/`endlocal` in `cmd`, save-and-restore in PowerShell) instead of exporting it. A shim wraps the real `codex`, so an exported marker would be inherited by Codex and everything it spawns — a shell beneath it running a different Bun would carry provenance describing a binary it is not executing.
- **The marker records the binary it was minted for** (`OCX_BUN_RUNTIME_PATH` beside `OCX_BUN_RUNTIME_SOURCE`). An earlier version re-derived the original selection to validate an inherited claim, which demoted a correct `override` to `process` on a service's first relaunch: an installed service keeps neither the installing shell nor its `OPENCODEX_BUN_PATH`. Comparison goes through `realpath` so symlinks, junctions, and Windows case differences do not reject a valid match.

Read-back allowlists the three values and never falls back to resolving locally. A service installed before the marker existed reports nothing, and doctor says the origin is unknown instead of guessing — an absent marker is an answer, not a gap to fill. `bunRevision` stays informational.

`src/lib/bun-stream-caps.ts` and `src/server/responses/core.ts` are absent from every commit here: the conservative `auto-known-bad` result for canaries and the eager-relay capability policy are untouched, as the issue directed.

## Verification

- `bun run test` — 6988 pass / 8 skip / 0 fail across 480 files on this branch rebased onto `dev`
- Red-then-green proven for both bugs by reverting the source and re-running the new regressions
- `structure/05_gui-and-management-api.md` documents the provenance trust rule and the backward-compatibility contract

Closes #848.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows service installation reliability by retrying transient scheduler registration delays while preventing unsafe rollback or stale state changes.
  * Improved runtime diagnostics by accurately identifying bundled, overridden, or system Bun runtimes.
  * Preserved runtime information across service, tray, shim, restart, and update launches.

* **Documentation**
  * Documented Bun runtime provenance and its inclusion in system memory information.

### DIFF
--- a/bin/ocx.mjs
+++ b/bin/ocx.mjs
@@ -321,6 +321,7 @@ const BUN_OVERRIDE_ENV = "OPENCODEX_BUN_PATH";
 // Node and runs before any TypeScript is loaded, so the name is repeated rather than
 // imported; tests/ocx-launcher-source.test.ts pins the two together.
 const BUN_RUNTIME_SOURCE_ENV = "OCX_BUN_RUNTIME_SOURCE";
+const BUN_RUNTIME_PATH_ENV = "OCX_BUN_RUNTIME_PATH";
 
 function findBunBinary(bunDir) {
   // The npm `bun` package ships the binary as bin/bun.exe on every platform;
@@ -423,6 +424,7 @@ const child = spawn(bun, [cliPath, ...process.argv.slice(2)], {
     ...process.env,
     OCX_PRE_BUN_ANTHROPIC_ENV: preBunAnthropicSlots.join(","),
     [BUN_RUNTIME_SOURCE_ENV]: bunRuntime.source,
+    [BUN_RUNTIME_PATH_ENV]: bunRuntime.path,
   },
 });
 
--- a/bin/ocx.mjs
+++ b/bin/ocx.mjs
@@ -317,6 +317,10 @@ function bunBinDir() {
 }
 
 const BUN_OVERRIDE_ENV = "OPENCODEX_BUN_PATH";
+// Mirrors BUN_RUNTIME_SOURCE_ENV in src/lib/bun-runtime.ts. This launcher is plain
+// Node and runs before any TypeScript is loaded, so the name is repeated rather than
+// imported; tests/ocx-launcher-source.test.ts pins the two together.
+const BUN_RUNTIME_SOURCE_ENV = "OCX_BUN_RUNTIME_SOURCE";
 
 function findBunBinary(bunDir) {
   // The npm `bun` package ships the binary as bin/bun.exe on every platform;
@@ -347,7 +351,7 @@ function resolveBun() {
   const override = process.env[BUN_OVERRIDE_ENV]?.trim();
   if (override) {
     const overridePath = resolve(override);
-    if (isRealBunBinary(overridePath)) return overridePath;
+    if (isRealBunBinary(overridePath)) return { path: overridePath, source: "override" };
     console.error(
       `opencodex: ${BUN_OVERRIDE_ENV} is missing, unreadable, or not a complete Bun binary; falling back to the bundled runtime.`,
     );
@@ -361,7 +365,7 @@ function resolveBun() {
   }
 
   let bin = findBunBinary(bunDir);
-  if (bin) return bin;
+  if (bin) return { path: bin, source: "bundled" };
 
   // Lazy fallback: --ignore-scripts (or a failed postinstall) leaves the
   // ~450-byte placeholder stub. Run the bun package's own installer once.
@@ -371,7 +375,7 @@ function resolveBun() {
     if (r.status === 0) bin = findBunBinary(bunDir);
   }
   if (!bin) fail("Bun binary missing after install attempt.");
-  return bin;
+  return { path: bin, source: "bundled" };
 }
 
 // `ocx update --help` prints usage and exits WITHOUT side effects. The npm launcher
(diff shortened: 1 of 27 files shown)


lidge-jun:
Superseded by #892, which lands both fixes natively: post-create scheduler verification now settles only on transient states (published-but-invalid XML rolls back with zero delays, conflicts/missing assets/unknown SCM stay fail-closed), and Bun runtime provenance is reported as a source+path pair through every launcher with legacy payloads reading as unknown. The unified branch additionally enforces the binary half of the pair (source-only or mismatched-path markers report unknown). Thank you — the issue analysis and the direction drove the fix.

PR #890, opened 2026-08-02
Fix DeepSeek Responses over Codex WebSocket

atian8179:
## Summary

- keep the Codex client-facing Responses WebSocket enabled
- use a bounded JSON upstream response for official DeepSeek V4 Flash WebSocket turns
- preserve streaming behavior for ordinary HTTP/SSE Responses requests

## Root cause

Codex Desktop initiates Responses turns over WebSocket. DeepSeek V4 Flash can deliver output for a streaming Responses request without delivering or closing on the terminal event, which leaves the Codex turn stuck indefinitely. Disabling the OpenCodex WebSocket transport is not a safe workaround: the built-in Codex provider probes WebSocket first, and its HTTP fallback can surface an opaque `503 Unknown error`.

## Fix

This adds a registry-only, per-model compatibility hint that is evaluated after final route selection. For an inbound WebSocket turn routed to the official DeepSeek V4 Flash transport, OpenCodex asks the upstream Responses endpoint for `stream: false`. The existing WebSocket bridge then reframes the bounded JSON response into the normal Responses WebSocket event sequence.

The hint is scoped to the registry-matched official transport, so custom providers with the same model name do not inherit it. HTTP/SSE requests continue to use upstream streaming.

## Impact

DeepSeek V4 Flash can complete normally in Codex Desktop without weakening the WebSocket path required by native GPT models. Other providers and non-WebSocket requests are unchanged.

## Validation

- `bun test tests/deepseek-inbound-wire.test.ts` — 16 passed, 0 failed
- `git diff --check` — passed for all changed files
- privacy patterns — passed for all changed files
- local Codex Desktop validation — native GPT and DeepSeek V4 Flash both completed; DeepSeek changed from hanging after first output to a completed turn

The full repository test run completed with 6,857 passes, 8 skips, 26 failures, and 5 dependency errors. The failures were outside the changed paths and were dominated by fixed-timeout tests under local filesystem load plus missing GUI React dependencies in the root install. The repository-wide privacy scanner was also blocked by the same macOS per-file read stalls, so the identical privacy patterns were applied directly to the five changed files instead.

### DIFF
--- a/src/providers/registry.ts
+++ b/src/providers/registry.ts
@@ -149,6 +149,14 @@ export interface ProviderRegistryEntry {
    */
   modelWireDefaults?: Record<string, ModelWireDefault>;
   /**
+   * Registry-only per-model override for the upstream request shape used behind a
+   * Codex Responses WebSocket turn. `false` keeps the client-facing WebSocket but
+   * asks the upstream Responses endpoint for bounded JSON, which the bridge then
+   * reframes as Responses events. Use only for upstreams whose streaming response
+   * can omit or indefinitely delay the terminal event.
+   */
+  modelWebsocketUpstreamStreaming?: Record<string, boolean>;
+  /**
    * Responses-API resource path for providers whose route is not `/v1/responses`.
    * Unlike `modelWireDefaults` above, this IS seeded into saved config: it describes
    * the provider's fixed endpoint rather than a default a user might want to override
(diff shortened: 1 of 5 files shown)


lidge-jun:
Landed in #892 as 5dd965a13 (authorship preserved). Adversarial review caught one real gap on top of your patch: the bounded-JSON path read the upstream body with an unbounded .text(), so we folded in the house bounded-body primitive (32 MiB ceiling + 180s/30s transfer deadlines, fail-closed 502, partial bodies never parsed) in 9794e2448/d8b707eaf. The registry-scoped hint design itself passed unchanged. Closing as superseded; thanks for the clean root-cause.

PR #889, opened 2026-08-02
fix(providers): Copilot mixed-wire routing, DeepSeek service_tier + reasoning replay, Claude 4.6/4.7 1M windows

lidge-jun:
## Summary

Three confirmed must-fix provider-wire bugs, one consolidated branch (wt3 lane of the bugfix campaign; unit docs in `devlog/_plan/260802_wt3_provider_wire/`). Each bug landed as its own PABCD cycle with an independent audit gate.

**1. Copilot mixed-wire routing (closes #748, supersedes the routing half of #746)**

The `github-copilot` preset's provider-wide `openai-chat` adapter breaks Responses-only models: `gpt-5.4` rejects `/chat/completions` for real Codex-agent traffic (function tools + reasoning), verified externally (litellm#23332, JetBrains LLM-29711, #748 field runs). The fix declares seven verified models (`gpt-5.3-codex`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.5`, `gpt-5.6-luna`, `gpt-5.6-sol`, `gpt-5.6-terra`) as registry `modelWireDefaults` — the precedence mechanism already on the tree (hard pin → explicit `modelAdapters` → registry default → provider adapter). Chat-served models are untouched; `gpt-5.4-nano` stays out (no field report) as a documented `modelAdapters` opt-in. #746's sampling/credential-replay parts are NOT included here — they remain a separate parity/security unit.

**2. DeepSeek: `service_tier` capability gate + reasoning replay (supersedes #860; local half of #875)**

- `service_tier` is an OpenAI-only Responses parameter, but fast mode injected it for every Responses provider. A provider-level `supportsServiceTier` capability now gates it after the final route is settled, adopting #860's reviewed fail-closed semantics: canonical `openai`/`openai-apikey` keep fast-mode behavior, `deepseek`/`volcengine-agent-plan` strip, unclassified providers fail closed unless explicitly opted in. Stripping also clears `options.serviceTier` so logging never mislabels a removed tier.
- Root cause of #875's local half: `sanitizeReasoningInputContent()` blanked reasoning `content` for EVERY Responses provider — a rule only the ChatGPT native backend needs. DeepSeek's Responses API accepts plaintext reasoning replay, so providers flagged `preserveResponsesReasoningContent` keep it (`ocxr1` envelopes still stripped). #875 stays open: the "no follow-up request sent at all" observation has no local explanation (triage comment posted).
[...]

Issue #879, opened 2026-08-02
[Bug] Star-prompt agent deferral leaks into every editing session and re-arms on each agent-driven start


lidge-jun:
## Symptom

The one-time GitHub-star consent prompt is not one-time in practice. In any agent-driven coding session that runs \`ocx start\` (or \`ocx service install\`) — e.g. an agent editing this repo and smoke-testing its change — the terminal prints the agent deferral, the agent becomes a persistent relay for \`Star lidge-jun/opencodex? Yes / No\`, and the question reappears at the top of every subsequent reply, across turns and across unrelated sessions, until the user explicitly answers. Users experience this as the star prompt "showing up during editing, not at install."

## Root cause

Three design decisions compound:

1. **The deferral is stateful-by-proxy, not on disk.** \`src/cli/star-prompt.ts\` (\`maybeShowStarPrompt\`) deliberately leaves the \`.star-prompted\` marker unwritten for agent-driven runs, so the real prompt survives for a later hand-typed run. But that also means every agent-driven \`ocx start\` re-prints \`printAgentDeferral()\` — there is no deferral marker, counter, or cooldown.
2. **The deferral text recruits the agent as an unbounded relay.** \`printAgentDeferral()\` instructs the agent: "Silence is not an answer... put the same Yes/No question at the top of your next reply, unchanged." AGENTS.md repeats the same repeat-forever rule. One \`ocx start\` inside one tool call therefore hijacks the top of every later reply in that session, and any new session that runs \`ocx start\` again re-arms it.
3. **Agent PTYs pass the TTY gate.** Harnesses run commands in a pseudo-TTY, so \`process.stdin.isTTY && process.stdout.isTTY\` is true even though no human is watching — the agent path is hit on routine edit/test cycles, not just installs.

Net effect: a one-shot install-time growth prompt becomes an unbounded cross-session nag delivered through agents, and it fires during ordinary bug-fix/editing work that merely smoke-tests \`ocx start\`.

## Evidence

- \`src/cli/star-prompt.ts\` — \`MARKER = ".star-prompted"\`, \`printAgentDeferral()\` (relay instructions), marker intentionally unwritten on the agent path.
- \`src/cli/agent-driven.ts\` — env-based detection (\`CODEX_THREAD_ID\`, \`CODEX_SHELL\`, \`CLAUDECODE\`, ...) matches essentially every agent harness, so the deferral path is the common case for agent runs.
[...]

Issue #748, opened 2026-07-30
OpenCodex GitHub Copilot Model Endpoint Mapping


mushikingh:
## Summary

OpenCodex configured the GitHub Copilot provider with the provider-wide `openai-chat` adapter. That works for most Copilot models, but several newer OpenAI models require the OpenAI Responses API instead of Chat Completions.

The affected models failed with errors such as:

```text
model "gpt-5.6-sol" is not accessible via the /chat/completions endpoint
```

`gpt-5.4` exposed a subtler version of the same issue. A minimal text-only Chat Completions request succeeded, but a real Codex request failed because Codex supplies function tools and a reasoning effort:

```text
Function tools with reasoning_effort are not supported for gpt-5.4 in
/v1/chat/completions. To use function tools, use /v1/responses or set
reasoning_effort to 'none'.
```

The fix is a per-model `modelAdapters` map on the `github-copilot` provider. This keeps the provider default as `openai-chat` while routing the incompatible models through `openai-responses`.

## Separate Selector Issue

Bare OpenAI model names are reserved by OpenCodex for its canonical `openai` provider. Therefore this selector does not mean GitHub Copilot:

```text
gpt-5.6-sol
```

When the canonical OpenAI provider is absent, it fails with:

```text
Model gpt-5.6-sol requires the canonical openai provider
```

The Copilot-qualified selector must be used:

```text
github-copilot/gpt-5.6-sol
```

Example:

```bash
codex -m github-copilot/gpt-5.6-sol \
  -c 'model_reasoning_effort="medium"'
```

## Applied Configuration

The effective configuration is stored in `~/.opencodex/config.json` under `providers.github-copilot.modelAdapters`.

```json
{
  "modelAdapters": {
    "claude-haiku-4.5": "openai-chat",
    "claude-opus-4.5": "openai-chat",
    "claude-opus-4.6": "openai-chat",
    "claude-opus-4.7": "openai-chat",
    "claude-opus-4.8": "openai-chat",
    "claude-sonnet-4.5": "openai-chat",
    "claude-sonnet-4.6": "openai-chat",
    "claude-sonnet-5": "openai-chat",
    "gemini-2.5-pro": "openai-chat",
    "gemini-3-flash-preview": "openai-chat",
    "gemini-3.1-pro-preview": "openai-chat",
    "gpt-3.5-turbo": "openai-chat",
    "gpt-3.5-turbo-0613": "openai-chat",
    "gpt-4": "openai-chat",
[...]

PR #892, opened 2026-08-02
fix: bound the six remaining retained-state leaks (#820 wave 2)

lidge-jun:
## Summary

Unified bug-fix campaign: the wt2 retained-state hardening plus all four planned workstreams (wt1 update path, wt4 server/config, wt5 Windows service, wt3 provider wire), reviewed per workstream by independent adversarial reviewers and verified end-to-end. Supersedes #840 #841 #843 #844 #845 #847 (closed earlier) and #871 #886 #887 #889 (closed as each workstream landed here).

## wt2 — retained-state bounds (tracker #820 wave 2)

- **#841** Responses continuation admission: oversized candidates go directly to durable spill (never resident, never demoting unrelated chains); 256 MiB envelope ceiling; bounded snapshot read (32 MiB, regular files only); bounded spill replay; UTF-8 snapshot selection.
- **#847** streamed tool arguments: per-call scopes in the non-stream collector; overflow is 502 `upstream_error` everywhere (was 413 ×4); no unbounded no-budget path; cancel-race fixes.
- **#844** Cursor Connect frames: cursor backlog with zero-copy handoff (exact 16 MiB boundary); raw-used accounting incl. headers; EOF drain-to-quiescence + typed `frame_incomplete`; terminal backlog lease owner.
- **#845** blob-ID keys: domain-separated fixed keys (`h:`/`d:`), key bytes counted + classified; zero-payload blobs stay evictable.
- **#843** Antigravity replay identities: SHA-256 over length-prefixed UTF-16 code units (injective for every JS string); streaming canonical abort; zero-call shell cleanup.
- **#840** Windows ACL memos: destination-keyed timeouts; temp-only ephemeral release at proven absence; both residual error classes gated.

## wt1 — update path

- **#871** installed previews compare by `major.minor.patch` core on the latest channel (registry target side stays strict).
- **#879** star-prompt agent deferral bounded: at most one relay per version (7-day bound while the version is unreadable); future-dated records fail toward re-asking; the consent invariant is unchanged (agent never answers; the human prompt is preserved).

## wt4 — server/config

- **#850** extension CORS origins matched by scheme+authority (configured extension ID only; wildcard rejected; all ten data-plane sites).
- **#869** atomic writes resolve symlinked destinations before temp+rename; first writes beneath a symlinked parent are refused; pid/port writers guard before any mkdir/chmod.

## wt5 — Windows service

- **#868** scheduler verification settles only on transient states; published-but-invalid XML rolls back with zero delays; conflicts/missing assets/unknown SCM stay fail-closed.
- **#848** Bun runtime provenance as a source+path pair through every launcher; legacy payloads read as unknown; source-only or mismatched-path markers report unknown.

## wt3 — provider wire

- **#748** Copilot mixed-wire routing: Responses-only models off chat completions; explicit `modelAdapters` preserved.
- **#860/#875** DeepSeek `service_tier` tri-state capability gate (`true` permits injection, `false` strips, `undefined` preserves caller values; docs aligned in EN + ja/ko/ru/zh-cn) and reasoning_content replay for tool-call continuations.
- **#839/#854** Claude Opus 4.6/4.7 + Sonnet 4.6 at 1M in the registry; `[1m]` picker rows only for authoritative ≥1M windows (no 372K route marked).

## wp6 — contributor bug-fix stack (5 PRs stacked, 2 triaged closed)

- **#881** Cursor `grok-4.5-fast`: sent as canonical `grok-4.5` + `requested_model.parameters` (`effort`, `fast=true`); legacy `model_details` omitted only for parameterized external selections; flattened ids remain live-discovery evidence only.
- **#885** shadow-call intercept: `x-codex-turn-metadata` `request_kind: "turn"` preserves foreground Luna turns; headerless/malformed/future-unknown metadata keeps legacy prefix interception.
- **#890** DeepSeek V4 Flash over Codex WebSocket: registry-scoped hint answers WS turns with bounded upstream JSON (`stream: false`), reframed by the WS bridge; HTTP/SSE stays streaming. Folded review: the upstream body read is now bounded (32 MiB + 180s/30s transfer deadlines via the house bounded-body primitive, fail-closed 502).
- **#896** AgentRouter EOF tolerance (#658): explicit `anthropicEofTolerance` capability (default off); EOF completes only with non-empty visible text or complete tool-call JSON; backward-scan repair with an exact UTF-8 byte cap (surrogate-safe); budget overflow is a single terminal event.
- **#895** routed model identity: `identifyRoutedModel` names the concrete wire model id under a strict ASCII allowlist (≤128 chars, `._/@:+-[]~`) with safe fallback, across catalog sync + anthropic/google/kiro/openai-chat.
- **#898** closed as duplicate — the #848 doctor guidance fix already landed here via wt5 (launch-time provenance; an active override is never re-prompted).
- **#894** closed unmerged — it rewrote another unit's locked devlog decisions and widened the deferred Darwin eager-relay gate; the snapshot-repair feature can return as a narrower PR after this lands.
- Adversarial review (3 rounds, sol-medium): bounded-body primitive gained a `maxBytes` option and geometric buffer accumulation.

## Validation

- `bun run test`: **7,452 pass, 8 skip, 0 fail** locally (full suite on daf7069f8, 499 files); every workstream's focused suites re-run after each stack
- **CI: all 14 checks green on 6ec6ffc75** — ubuntu / macos / windows full matrix, npm-global ×3, systemd/launchd/schtasks, react-doctor, enforce-target, CodeRabbit. Windows is green on this branch for the first time in the campaign (the dev-inherited matrix breakage is not present here)
- CI stabilization folded during wp6 (test-only, runner-flake class): #848 provenance test timeout headroom (a8e0fe0cf), dead-pid probe for the symlinked sweep (0bccc8a0a), STORE_BUDGET_MS for the two cleanup-mode tests (80c49cfbf), thrown-CIM-deadline tolerance in the live enumeration test (6ec6ffc75)
- `bun x tsc --noEmit`: pass
- `bun run privacy:scan`: pass
- Red-green verified per fix; 15 adversarial reviewer rounds across 6 independent reviewers, all findings folded or rebutted
- Pre-push hook (typecheck + gui lint + full test + privacy scan + doctor:gui): pass on every push through eeef7a32a; later heads were pushed with the hook skipped per maintainer instruction, with the identical full suite + typecheck + privacy:scan run manually first

## Notes

- No RSS/benchmark superiority claims (comparator cells remain UNKNOWN).
- Co-developed with a second session on this branch (wt2 augmentations, wt1 commits, review-fix iterations); original PRs' analyses drove each fix.

## Summary by CodeRabbit

* **New Features**
  * Added `ocx config` commands for viewing, editing, validating, importing, and exporting configuration.
  * Added Codex prompt-layer management, including toggles, custom instructions, adoption, recovery, and safe rollback.
  * Added Bun runtime provenance to service diagnostics and launchers.
  * Expanded provider support for Copilot routing, Claude 1M context, service-tier capabilities, and reasoning replay.

* **Bug Fixes**
  * Improved update detection, GitHub star-prompt deferral, CORS extension handling, symlink-safe writes, Windows service settling, and memory safeguards.
  * Improved error reporting for upstream translation-limit failures.


lidge-jun:
CodeRabbit round folded (a86ee03cf):

- **go/internal/cli/config_parity.go** — untracked again (had been re-added by a broad `git add -A`; stays on disk, not in the repo).
- **prompt-journal ×3** — the journal temp now hardens `required: true` (it carries full config.toml bytes; fail closed like the token/tray writers); `durableDelete` propagates non-ENOENT failures; recovery revalidates each target immediately before its own restore, and delete failures map to `recovery_required` instead of a false `committed`/`rolled-back`.
- **antigravity canonicalization** — recursion depth capped at 128 (byte/keys budgets did not stop a nesting bomb; overflow keeps the skip-replay contract).
- **Cursor framing** — payloads ≤ 64 KiB are copied rather than viewed, so small frames can't pin the 32 MiB backlog buffer in the work queue.
- **bridge abandoned streams** — an unref'd watchdog now disposes the owned default budget (a dropped, never-read stream no longer sits in `liveBudgets` for the process lifetime; the delay is test-overridable).
- **README/AGENTS.md** — the star deferral's weekly re-arm bound is documented on both surfaces.

Notes on the remaining comments: the devlog pre-disclosure flags (wt1/wt4 execution notes) pass the repo's own test — every defect they describe already has a public diff (the original PRs and this one), so they disclose nothing new and stay as implementation records. The other devlog plan-doc nits belong to sibling workstreams outside this PR's scope. The abandoned-budget alternative you suggested (defer to first `pull()`) is spec-incompatible — Bun calls `pull` at construction — so the watchdog is the chosen fix.

Issue #765, opened 2026-07-30
anthropic adapter: baseUrl ending in /v1/messages doubles the path, input_json_delta is not scoped to tool_use, and Anthropic-compatible relays need tool-call quirk tolerance


brunoflma:
### Client or integration

Codex CLI

### Provider or upstream service

AgentRouter (`agentrouter.org`), an Anthropic Messages-compatible relay. Two of the four problems below are provider-agnostic and affect the `anthropic` adapter for any host.

### OpenCodex version

2.7.43

### Endpoint or capability

`POST /v1/messages` — streaming tool calls, and `baseUrl` normalisation.

### Current behaviour

**1. `baseUrl` ending in `/v1/messages` produces a doubled path (provider-agnostic).**

```ts
const base = provider.baseUrl.replace(/\/v1\/?$/, "");
const url = `${base}/v1/messages`;
```

The regex only strips a trailing `/v1`. A `baseUrl` of `https://host/v1/messages` — a natural thing to paste from provider documentation — becomes `https://host/v1/messages/v1/messages` and every request 404s. Affects any Anthropic-compatible provider, not just this one.

**2. `input_json_delta` is applied outside `tool_use` blocks (provider-agnostic).**

```ts
} else if (delta.type === "input_json_delta" && typeof delta.partial_json === "string") {
```

There is no check that the current content block is a `tool_use`. Per the Messages streaming spec `input_json_delta` only occurs within `tool_use`; a provider that emits it elsewhere (or emits deltas after a block closes) corrupts the accumulated tool arguments.

**3. `tool_use` blocks arrive without an `id`.** The adapter forwards `block.id ?? ""`, so an empty tool-call id reaches the Codex client, which then cannot correlate the tool result.

**4. Streams close without terminal frames.** This relay sometimes ends an SSE stream after content blocks without `message_stop` / `message_delta`, so the adapter never emits a completion event and the client hangs waiting for one.

Additionally, `tool_use.input` sometimes arrives as a JSON **string** rather than an object, occasionally with a malformed prefix. `JSON.stringify(block.input ?? {})` then produces a doubly-encoded string that the client rejects.

### Expected behaviour

1. `baseUrl` normalisation should tolerate `/`, `/v1` and `/v1/messages` suffixes and produce exactly one `/v1/messages`.
2. `input_json_delta` should only accumulate while the current block is a `tool_use`, per the [Anthropic streaming spec](https://docs.anthropic.com/en/docs/build-with-claude/streaming).
[...]

PR #896, opened 2026-08-02
fix(anthropic): complete AgentRouter streams that end before terminal frames

### LINKED ISSUE #658: [Provider compatibility]: AgentRouter Anthropic streams can end without terminal SSE frames
### Client or integration

Codex CLI and the Anthropic Python SDK through the OpenCodex Anthropic adapter.

AgentRouter (`https://agentrouter.org`), using its Anthropic-compatible `/v1/messages` API.

### OpenCodex version

`2.7.42`

### Endpoint or capability

Anthropic-compatible streaming and tool calls via `/v1/messages`.

### Current behaviour

Some non-Claude models exposed by this provider emit valid Anthropic-compatible SSE content but close the connection before `content_block_stop`, `message_delta`, and `message_stop`.

The OpenCodex Anthropic adapter treats that EOF as a fatal truncation for every provider. It discards otherwise valid visible text and completed tool-call arguments, returning an adapter error instead of completing the response.

The affected provider can also omit `tool_use.id`, and may return `tool_use.input` as a malformed concatenation such as `{}{\"value\":42}` in non-streaming responses.

### Expected behaviour

For a deliberately configured compatibility profile for this provider, OpenCodex should be able to:

1. preserve text when the upstream emits valid content and then closes without terminal SSE frames;
2. complete a tool call only when the accumulated arguments form a valid JSON object;
3. synthesize a stable non-empty tool-call ID when the upstream omits it;
4. normalize malformed tool arguments to a valid JSON object.

The default behavior for native Anthropic and other strict providers should remain unchanged: incomplete streams must continue to fail.

### Minimal redacted reproduction

```python
import anthropic

# Instantiate the SDK with a redacted local proxy credential and this base URL.
client = anthropic.Anthropic("REDACTED", base_url="http://127.0.0.1:10100")

# The direct upstream call returns the visible text. Through OpenCodex before
# the patch, the same request fails with an adapter EOF error.
with client.messages.stream(
    model="PROVIDER/glm-5.2",
    max_tokens=64,
    messages=[{"role": "user", "content": "Reply exactly: OK"}],
) as stream:
    print("".join(stream.text_stream))
```

Raw redacted upstream SSE shape observed for text:

```text
message_start
content_block_start (text)
content_block_delta (text_delta with visible text)
EOF
```

Raw redacted upstream SSE shape observed for a tool call:

```text
message_start
content_block_start (tool_use; id may be present or absent)
content_block_delta (input_json_delta fragments forming a complete JSON object)
EOF
```

### Actual response or error

```text
upstream stream ended before message_stop — possible truncation
```

The direct AgentRouter request using the Anthropic Python SDK returned the expected text for `glm-5.2` and `gpt-5.5`. Through OpenCodex before the local compatibility patch, both failed with the error above.

### Upstream documentation

- https://docs.anthropic.com/en/api/messages-streaming
- https://agentrouter.org/docs

AgentRouter documents an Anthropic-compatible endpoint, but the observed stream does not consistently include the terminal Anthropic SSE frames.

A narrow compatibility fallback can be implemented in `src/adapters/anthropic.ts`, gated by an exact hostname match for `agentrouter.org`:

- On EOF without `message_stop`, emit `done` only after valid content has been received and no incomplete tool call is open.
- If a tool call is open, emit `tool_call_end` and `done` only when its accumulated arguments parse to a JSON object.
- Otherwise keep emitting the current truncation error.
- For this provider only, synthesize a `toolu_<uuid>` ID when a `tool_use` ID is absent, and normalize non-streaming tool inputs to a JSON object.

I verified this behavior with focused regression tests covering: normal AgentRouter text EOF, incomplete tool JSON, complete tool JSON at EOF, missing tool IDs, malformed concatenated tool input, and native Anthropic EOF remaining an error.

### Additional context

No credentials, correlation values, account information, local paths, usage logs, or raw provider payloads are included here. I searched existing compatibility issues and found related SSE-truncation reports, but no open issue specifically for AgentRouter.

- [x] The expected behaviour is based on a concrete client requirement and the Anthropic streaming protocol.

Yuxin-Qiao:
## Summary

Closes #658: AgentRouter's Anthropic-compatible endpoint can close the stream before `content_block_stop`, `message_delta`, and `message_stop`. The Anthropic adapter previously treated every such EOF as a fatal truncation and discarded otherwise valid text and complete tool-call arguments.

This PR implements the accepted provider-compatibility feature as an explicit capability (`anthropicEofTolerance`, default off) rather than a hard-coded hostname branch:

- **Streaming:** with the capability enabled, an EOF may complete only when visible text was received or an open tool call has complete JSON-object arguments. Incomplete tool JSON, no usable content, and transport failure before usable content remain truncation errors with the exact previous message. Strict providers keep the previous behavior unchanged.
- **Non-streaming:** with the capability enabled, malformed concatenated tool input such as `{}{"value":42}` is repaired to the last valid JSON object (`{"value":42}`) instead of being discarded to `{}`. The strict path is unchanged.
- **Tool IDs:** missing or blank `tool_use.id` keeps the existing stable synthesized-id path (#765), now pinned by a regression test for this provider.
- **Registry:** the AgentRouter row in `src/providers/free-directory.ts` declares the capability.

## Repro (wire shape)

```text
message_start
content_block_start (text)
content_block_delta (text_delta "visible")
EOF
```

Before: `upstream stream ended before message_stop — possible truncation`, visible text lost. After (with `anthropicEofTolerance: true`): text preserved and the stream completes with `done`.

## Tests

- New `tests/anthropic-eof-tolerance.test.ts` (8 cases): text EOF on/off, complete/incomplete tool EOF, no-content EOF, synthesized tool id, non-stream concatenation on/off, directory row capability.
- `bun run typecheck` passes.
- Focused suites (`anthropic-compatible-stream.test.ts`, `provider-registry-parity.test.ts`, `zhipu-bigmodel-provider.test.ts`) pass.
- Full suite compared against a clean `dev` baseline in the same environment: baseline has pre-existing failing tests (GUI/network-dependent, unrelated to this change); this branch introduces no additional failures, and the 4 flaky WP040 probe tests that were branch-only pass in isolation.
- No live AgentRouter endpoint was available, so the tests pin wire behavior only; strict default behavior is covered by the negative tests.

## Docs

- `providers.md` configuration table row added in English plus ja/ko/ru/zh-cn locales.

## Summary by CodeRabbit

* **New Features**
  * Added an optional provider setting to tolerate early termination of Anthropic-compatible streams when complete text or tool input has been received.
  * Added limited repair for malformed non-streaming tool arguments when enabled.
  * Enabled this capability for the AgentRouter provider.
* **Bug Fixes**
  * Incomplete or empty streams continue to report truncation errors by default.
* **Documentation**
  * Documented the new setting across supported languages.

### DIFF
--- a/docs-site/src/content/docs/ja/reference/configuration/providers.md
+++ b/docs-site/src/content/docs/ja/reference/configuration/providers.md
@@ -80,6 +80,7 @@ description: プロバイダー エントリ、認証、エンドポイント、
 | `thinkingBudgetModels?` | `string[]` |整数 `thinking_budget` を使用したチャット モデル。労力は予算の一部にマッピングされます。 |
 | `noVisionModels?` | `string[]` |ビジョン サイドカーを通じて送信されるテキストのみのモデル。マッチングでは、Ollama `:size` タグが許容されます。 |
 | `escapeBuiltinToolNames?` | `boolean` | Anthropic 互換ゲートウェイの組み込みツール名をエスケープし、返された呼び出しで復元します。 |
+| `anthropicEofTolerance?` | `boolean` | `message_stop` 前にストリームが終了しても、可視テキストまたは完全な JSON オブジェクトのツール入力が受信済みの場合に限り完了を許可します（Anthropic 互換ゲートウェイ向け）。デフォルトはオフ。 |
 | `googleMode?` | `"ai-studio" \| "vertex" \| "cloud-code-assist"` | Google トランスポート/認証モード。デフォルトは`ai-studio`です。 |
 | `project?` | `string` | Vertex または Antigravity Cloud Code Assist プロジェクト ID。 |
 | `location?` | `string` |頂点の位置。環境フォールバックは `GOOGLE_CLOUD_LOCATION` です。 |
--- a/docs-site/src/content/docs/ko/reference/configuration/providers.md
+++ b/docs-site/src/content/docs/ko/reference/configuration/providers.md
@@ -80,6 +80,7 @@ description: 공급자 항목, 인증, 엔드포인트, 모델 카탈로그, 할
 | `thinkingBudgetModels?` | `string[]` | 정수 `thinking_budget`를 쓰는 chat 모델입니다. effort는 예산 비율로 매핑됩니다. |
 | `noVisionModels?` | `string[]` | vision sidecar로 보내는 텍스트 전용 모델입니다. 일치 판정은 Ollama `:size` 태그도 허용합니다. |
 | `escapeBuiltinToolNames?` | `boolean` | Anthropic 호환 게이트웨이를 위해 내장 도구 이름을 이스케이프하고, 반환된 호출에서는 다시 복원합니다. |
+| `anthropicEofTolerance?` | `boolean` | `message_stop` 전에 스트림이 끝나도 표시 텍스트 또는 완전한 JSON 객체 툴 입력을 받은 경우에만 완료를 허용합니다（Anthropic 호환 게이트웨이용）. 기본값은 꺼짐. |
 | `googleMode?` | `"ai-studio" \| "vertex" \| "cloud-code-assist"` | Google 전송/인증 모드입니다. 기본값은 `ai-studio`입니다. |
 | `project?` | `string` | Vertex 또는 Antigravity Cloud Code Assist 프로젝트 id입니다. |
 | `location?` | `string` | Vertex 위치입니다. 환경 변수 폴백은 `GOOGLE_CLOUD_LOCATION`입니다. |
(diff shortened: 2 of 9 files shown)


lidge-jun:
Landed in #892 as 538d755fb (authorship preserved), with three review folds on top: the repair scan is now backward/lastIndexOf with no offset arrays plus an exact UTF-8 byte cap (surrogate-safe), only non-empty text authorizes tolerant completion, and a budget overflow returns as the single terminal event (9794e2448, d8b707eaf, daf7069f8). One honest note: FREE_PROVIDER_DIRECTORY is currently consumed by tests only, so the AgentRouter row declaration is inert metadata — the capability activates via the provider config flag, matching your explicit-capability design. Closing as superseded.

PR #898, opened 2026-08-02
fix(cli): make doctor runtime-origin aware for OPENCODEX_BUN_PATH guidance (#848)

Yuxin-Qiao:
## Summary

`ocx doctor` on Windows told users to set `OPENCODEX_BUN_PATH` even when the running service already uses that override, making the guidance circular (#848).

- `GET /api/system/memory` now reports the service process's own runtime source (`bundled` / `override` / `process`) as a scalar enum, never the runtime path.
- The doctor "Memory / runtime" section shows the Bun revision and, for an active override, identifies the override as active but unvalidated for automatic eager relay, without repeating the "set OPENCODEX_BUN_PATH" instruction. Bundled known-bad guidance is unchanged.
- Docs: the windows-memory troubleshooting page is updated in all five locales.

## Verification

- `bun run typecheck` - passed
- Focused: `bun scripts/test.ts tests/doctor.test.ts tests/memory-watchdog.test.ts` - 41 pass / 0 fail, including the new #848 regression tests
- Full: `bun scripts/test.ts` - 6869 pass, 8 skip; the 16 unique failing tests reproduce on a pristine `origin/dev` checkout with identical counts (environmental: `bun` missing from PATH in the isolated test env, missing `gui/node_modules`, provider-validation baseline); none touch changed files
- `bun run privacy:scan` - passed
- `git diff --check` - clean

## Checklist

- [x] Scope stays focused on the doctor/API surface and avoids unrelated cleanup.
- [x] Docs updated where needed (English plus zh-cn, ja, ko, ru).
- [x] No new secrets or auth surfaces; the payload addition is a scalar enum.

## Summary by CodeRabbit

* **New Features**
  * `ocx doctor` now reports the Bun revision and whether the runtime is bundled or overridden through `OPENCODEX_BUN_PATH`.
  * Windows memory diagnostics provide clearer guidance when a runtime override is active.

* **Bug Fixes**
  * Prevented duplicate or misleading upgrade instructions for active runtime overrides.
  * Clarified that memory metrics alone do not confirm a leak on the bundled runtime.

* **Documentation**
  * Updated troubleshooting guidance in English, Japanese, Korean, Russian, and Simplified Chinese.


lidge-jun:
Closing as a duplicate: the #848 fix already landed in #892 via the wt5 workstream. On codex/wt2-zero-leak-impl, doctor reads the launch-time provenance marker (`bunRuntimeSource`: `override` / missing / `process`|bundled) and never tells the user to set OPENCODEX_BUN_PATH when that override is already active — see src/cli/doctor.ts:665-679 on that branch, which names #848 explicitly. The provenance pair (source+path) is recorded through every launcher, with legacy payloads read as unknown. Thanks for the clean write-up; the API-payload variant here lost out to the launch-marker approach because it cannot be spoofed by a stale service response.

PR #880, opened 2026-08-02
feat(provider): add ClinePass subscription provider

n3wr1ch:
## Summary

- add ClinePass as a canonical API-key provider preset with the 11 model slugs published by Cline
- preserve full `cline-pass/<model>` upstream ids and seed context-window and input-modality metadata
- support the gateway's reasoning object, non-stream response envelope, reasoning deltas, choice-scoped stream errors, and safe numeric error classification
- expose `reasoningWireFormat` through the safe management DTO and keep GUI diagnostics, provider docs, and all five documentation locales synchronized

## Why

Cline officially supports using ClinePass outside the Cline clients through its OpenAI-compatible Chat Completions API. OpenCodex previously had no canonical preset for that subscription route, so users had to configure the endpoint manually and would lose gateway-specific reasoning, error, and non-stream response behavior.

The adapter fixes are intentionally shared where they describe valid OpenAI-compatible response behavior. Regression coverage protects ordinary non-Cline providers from changes to standard response parsing and pending tool-call termination.

## Primary-source evidence

Verified 2026-08-02:

- ClinePass catalog, full model slugs, quota description, and explicit authorization for scripts/apps/automation: https://docs.cline.bot/getting-started/clinepass
- OpenAI-compatible endpoint and request/stream/tool/reasoning contract: https://docs.cline.bot/api/chat-completions
- numeric/string error codes, request-id metadata, retry statuses, and choice-scoped mid-stream errors: https://docs.cline.bot/api/errors
- service operator and terms: Cline Bot Inc., https://cline.bot/tos
- maintenance owner recorded in the registry: `@lidge-jun`

The public API reference does not currently document the `reasoning: { enabled, effort }` request object or the `{ success, error, data }` non-stream envelope. Those two shapes are explicitly marked in code as live-observed on 2026-08-02. Only the live-accepted `low` reasoning tier is advertised until Cline documents or validates a broader ladder.

## Security and privacy

- the new credential destination is fixed to `https://api.cline.bot/api/v1`
- same-named custom providers retain their custom destination and credential boundary
- API keys and response bodies are not logged
- arbitrary upstream metadata is discarded; only a bounded request id with an allowlisted character set is retained, and secret-shaped values are rejected by the existing redactor
- `reasoningWireFormat` is added only to the existing non-secret management DTO allowlist

This provider preset changes a credential destination and therefore requires explicit security review before merge.

## Validation

- `bun run typecheck`
- `bun run test` — 6,973 passed, 8 skipped, 0 failed
[...]

PR #899, opened 2026-08-02
ci: shard the dev lane on Linux and move Windows to the shipping boundary

lidge-jun:
## What this changes

The dev/PR lane becomes a Linux lane, split four ways. Windows moves to the
shipping boundary. macOS keeps running the whole suite.

- The suite runs as four `bun test --shard=i/4` jobs on `ubuntu-latest`.
- Everything that is not the suite — typecheck, privacy scan, GUI tests, lint,
  build, release-helper syntax, CLI smoke — moves into one `gates` job instead
  of being repeated per platform.
- `platform-macos` runs the **whole suite, unsharded**, on every pull request.
- `platform-windows` runs on promotion to `main`/`preview` and on
  `workflow_dispatch`, not on pull requests.
- A new `ci` job aggregates every other job into one stable check name.
- The GUI lint/build steps and the `npm-global-smoke` jobs run only when their
  own inputs changed.

## Why

Measured on run [30748690567](https://github.com/lidge-jun/opencodex/actions/runs/30748690567),
a fully green run of #880:

| leg | wall clock |
|---|---|
| ubuntu | 5m 49s |
| macos | 5m 23s |
| **windows** | **16m 23s** |

The Windows leg decided when every pull request turned green. The comment
above its `timeout-minutes` records that ceiling being raised twice —
after #711 passed at 11.8min while #653 was killed at 12.0min (#717), then
again after #827 hit the 20-minute kill — rather than the gap being closed.

**This is not a cost argument.** Standard runners are free on public
repositories, so no minutes are saved. What changes is feedback latency and
the concurrency budget (20 concurrent jobs on the Free plan, 5 for macOS)
against roughly twenty open pull requests.

## The cost, stated plainly

A Windows-only regression can now land on `dev` and be caught at promotion
instead of at review time. That is a real reduction in when Windows is
verified, and it is the price of the change.

What keeps it acceptable:

- `dev` is not published. `main` and `preview` are, and `release.yml` already
  refuses to publish a commit without a successful run of this workflow — so
  the Windows leg still gates everything that ships.
- The Windows `npm-global-smoke` leg (1m 53s) still runs on any change to a
  packaged input, which includes `src/**`. An ordinary source PR keeps a
  Windows signal.

If Windows-only regressions turn out to be frequent on `dev`, the right
response is to put Windows back on the PR lane *sharded*, not to defend this
arrangement.

## Concurrency: this trades more jobs for a shorter path

| | jobs per run | peak simultaneous |
|---|---|---|
| before | 7 | 6 |
| non-packaging PR | 9 | 7 |
| source or packaging PR | 12 | up to 10 |
| promotion | 13 | up to 10 |

"Faster CI" usually implies "less CI"; here it does not, and the peak goes up.

## Required checks

`dev` has no branch protection configured today, so nothing needs re-pointing.
If it is enabled later, **`ci` is the check to require** — shard names change
with the shard count and platform legs come and go by trigger, so neither is
stable.

⚠️ Requiring `ci` also means removing the workflow-level `paths:` filter, or
moving the gate to an always-triggered workflow. A docs-only PR does not
trigger this workflow, so no `ci` check would be created and the PR would sit
pending forever. Harmless now, a trap afterwards.

## Two details worth a reviewer's attention

**`base: ${{ github.ref }}` on the filter step.** Left unset, `paths-filter`
compares against the repository default branch (`main`), so a push to `dev`
would be diffed against `main` and every area touched since the last promotion
would keep reading as changed. The scoped jobs would run anyway, the saving
would not happen, and nothing would look wrong.

**The gate treats `skipped` as a pass.** That is required, not sloppy: a
trigger-scoped or area-filtered job reports `skipped`, and GitHub reports
skipped jobs as success regardless. The gate uses an allowlist (`success` or
`skipped` pass, everything else fails) so a result type added later cannot slip
through, and `if: always()` keeps the gate from being skipped along with a
failed dependency and reporting success by omission.

## Security review requested

Per `MAINTAINERS.md`, GitHub Actions changes require explicit security review.
Specific points:

- `dorny/paths-filter` is a **new third-party dependency**, pinned to the full
  commit SHA `de90cc6f` (verified to equal tag v3.0.2). A no-dependency
  fallback using `git diff --name-only` is documented in the devlog if the
  dependency is unwanted.
- The `changes` job takes `pull-requests: read`, a permission this workflow did
  not previously hold. It is scoped to that one job; the action cannot read a
  PR's file list without it.
- `select-windows-runner` and its security commentary are **unchanged**. The
  self-hosted routing, and the note that it is a cost control rather than a
  security boundary, carry forward verbatim.
- No job gains `contents: write`. No secret is newly exposed.

## Verification

- `actionlint`: clean.
- `bun x tsc --noEmit`: exit 0.
- `bun run test`: 6960 pass, 8 skip, **0 fail**, 33933 assertions across 480 files.
- Shard tiling proven on this suite: 120 files per shard, union 480, **no file
  in two shards**, none missed.
- `tests/ci-workflows.test.ts`: 66 pass, 733 assertions (692 before).

Every new pin was **driven red before being kept** — shard/divisor drift, a job
missing from the gate's `needs`, a deleted `base:`, a dropped
`pull-requests: read`, and a path present in one trigger list but not the other
each fail the suite.

## Planning record

`devlog/_plan/260803_ci_dev_lane_sharding/` carries the diff-level plan and the
evidence. It includes the audit trail: an independent review ran four rounds
and failed the plan three times. Every blocker was a local fact asserted
without being checked — tests said not to need React that import JSX, a phase
boundary that would have deleted two platforms in a commit claiming otherwise,
a duplicated YAML key, and the `base:` default above. The corrections are
recorded rather than tidied away.

## Summary by CodeRabbit

* **CI Improvements**
  * Faster Linux validation through four parallel test shards.
  * macOS continues running the complete test suite as a cross-platform check.
  * Windows checks now run during promotions or manual dispatches.
  * GUI and packaging checks run only when relevant files change.
  * CI triggers now include relevant assets and project documentation changes.
  * Added a final status gate for a consistent overall CI result.
* **Documentation**
  * Added planning and verification documentation for the updated CI workflow.


lidge-jun:
Thanks both — nine comments, seven fixed in `fd5ca50f`, one already fixed in `32c56a82`, one rebutted with evidence.

**Codex P1 (release gate) was the best find in this review, planning rounds included.** It is a hole this PR opened *outside the file it edits*: `release.yml` selects a CI run by SHA and status alone, so after a promotion the pull-request run for that same commit is still there, still green, and now deliberately Windows-free. The publish gate could have been satisfied by a run that proved nothing about Windows while the promotion run carrying it was still in flight or had failed. That is exactly the coverage hole this PR promised not to open. The gate now requires a `push` run on the release branch, and the error message says why so it does not read as flakiness.

| # | Finding | Disposition |
|---|---|---|
| Codex P1 | Release gate accepts a Windows-free PR run | **Fixed** — gate selects `event == "push"` on the release branch |
| Codex P2 | `.gitattributes` missing from packaging filter | **Fixed** — it decides how packaged inputs are materialized, so an attribute change can put CRLF shebangs in the tarball with no source file moving |
| CodeRabbit | `git clean -xffd . \|\| true` swallows wipe failures | **Fixed** — only the no-repository case (fresh box) is tolerated now |
| CodeRabbit | `persist-credentials: false` missing | **Fixed** on all six checkouts; the other workflows here already did this and `ci.yml` was the gap |
| CodeRabbit | Plan examples omit the GUI build | **Fixed** — see below, the live run found this independently |
| CodeRabbit | `040` gate block claims fail-fast without `set -e` | **Fixed** |
| CodeRabbit | `040` leak scan omits `devlog`, uses two pattern sets | **Fixed** — one pattern set, scans the whole diff |

Each fix is pinned in `tests/ci-workflows.test.ts`, driven red before being kept.

**Rebutted: the "future-dated" devlog records.** The dates read 2026-08-03 because that is the date in the authoring timezone. The same moment is 2026-08-02 in UTC, which is what the CI timestamps you compared against show:

```
local: 2026-08-03 02:55 KST
utc:   2026-08-02 17:55 UTC
```

The unit slug matches. Rewriting them to a UTC reading would make the record less accurate.

**Worth noting separately:** the first sharded run went red on shard 1/4 and macOS, and it was a real defect in this PR rather than flake. Tests that fetch the served dashboard read their session bootstrap out of `gui/dist/index.html`, and nothing in those jobs built it. The old three-platform job satisfied that dependency *by accident* — the suite and the GUI build shared a job. I confirmed the dependency is pre-existing by reproducing the failure on an untouched checkout of `dev` with no `gui/dist`, where it passes after `bun run build`. Fixed in `32c56a82`, with a pin asserting every job that runs the root suite builds the GUI unconditionally.

That one is the reason the audit trail in `devlog/_plan/260803_ci_dev_lane_sharding/002_audit_synthesis.md` is worth reading: four rounds of pre-merge review could not have caught it, because the dependency existed only as a side effect of two things sharing a job. Running it found it in four minutes.

Current run: 4m26s wall clock against the 16m39s baseline, all shards green, Windows correctly skipped, `ci` gate green.

Issue #875, opened 2026-08-02
[Bug] DeepSeek V4 Flash Responses route stalls after tool calls


Michael-Han0608:
Thanks for the report. Include a minimal reproduction and exact environment details so we can confirm and fix the problem quickly.

### Client or integration

Codex App

### Area

Tools / MCP / web search

### Summary

DeepSeek V4 Flash requests routed through the native Responses adapter can stall after tool calls. The upstream request returns HTTP 200 and non-zero output/reasoning tokens, but OpenCodex does not continue with the corresponding tool outputs, so no final visible answer is produced.

Plain-text turns without tools can complete normally. DeepSeek V4 Pro through the `openai-chat` route works normally under the same local setup.

### Update: reproduced on OpenCodex 2.10.0

The same behavior remains reproducible after upgrading to OpenCodex **2.10.0**:

- plain-text turns may complete normally;
- tool calls can be generated and executed;
- the upstream DeepSeek Responses request returns HTTP 200 with non-zero output/reasoning tokens;
- no corresponding `function_call_output` or follow-up Responses request is observed for the same turn;
- Codex App receives no final visible answer and the turn stalls.

The v2.10.0 routing change does not appear to address this continuation failure. Codex-originated requests still use the native Responses path, so the remaining defect appears to be in the Responses tool-call continuation, event handoff, or reasoning-history replay rather than initial model routing.

### Reproduction

1. Use OpenCodex with the official DeepSeek API and select `deepseek/deepseek-v4-flash`.
2. Ensure the model is routed through the native Responses adapter.
3. Start a turn that causes the model to return one or more tool calls.
4. Observe that the tool calls may complete, but no corresponding `function_call_output` or follow-up request is sent for the same conversation, and no final assistant answer appears.
5. A plain-text turn without tools can complete normally. DeepSeek V4 Pro through the `openai-chat` route does not reproduce the problem.

The exact trigger is not yet confirmed; it may involve the native Responses event handoff, multiple tool calls, or reasoning-history replay.

### Version

Originally reproduced on OpenCodex 2.9.1; reproduced again on OpenCodex 2.10.0.

### Operating system

macOS

### Provider and model

Official DeepSeek API / `deepseek/deepseek-v4-flash`

### Logs or error output

Relevant local usage records:

- `ocx-msbd27ow-17`: HTTP 200, output tokens 568, reasoning tokens 359
[...]

PR #906, opened 2026-08-02
fix(deepseek): preserve Responses reasoning replay on continuations (#875)

Yuxin-Qiao:
## Summary

Fixes the proxy half of #875: DeepSeek V4 Flash Responses turns stall after tool calls because the passthrough sanitizer blanked `reasoning.content` for every provider. DeepSeek's Responses API is stateless and requires replaying prior assistant reasoning text on every continuation; an emptied reasoning item breaks the replay, so tool-call loops cannot finish.

- `sanitizeReasoningInputContent` now keeps raw `reasoning_text` content for models in `preserveReasoningContentModels` (already seeded for DeepSeek thinking models, including `deepseek-v4-flash`) and keeps the existing blanking for ChatGPT-native passthrough, which accepts reasoning input only with empty `content`.
- Proxy-minted ocxr1 envelopes are still stripped unconditionally.
- This is the "local half" of the #875 triage; PR #890 covers the complementary Codex Desktop WebSocket terminal-event half.

## Why it matters for subagents

With this fix, `deepseek/deepseek-v4-flash` can be selected as a sub-agent worker (`ocx agent subagents set deepseek/deepseek-v4-flash` or `"subagentModels": ["deepseek/deepseek-v4-flash"]`) and its tool-call loops continue correctly. Added a regression test proving the routed entry is a viable sub-agent candidate.

## Change

- `src/adapters/openai-responses.ts`: `sanitizeReasoningInputContent(body, opts)` with a preserve list + model id; content kept for preserve-listed models, blanked otherwise; envelopes always stripped. Call site passes `provider.preserveReasoningContentModels` and `parsed.modelId`.
- `tests/deepseek-reasoning-replay.test.ts`: new file, 5 tests covering preserve/blank/envelope unit behavior, an end-to-end handleResponses continuation carrying reasoning + tool output upstream, and sub-agent routability.

## Validation

- `bun test tests/deepseek-reasoning-replay.test.ts tests/deepseek-inbound-wire.test.ts` — 19 pass, 0 fail
- `bun run typecheck` — pass
- `git diff --check` — pass
- Adapter/subagent-adjacent regression set (76 files): failure set is identical to the pre-change baseline on this machine (environmental port/timeout failures only). The full-suite run crashed under the local sandbox (Bun segfault); CI remains the authority for the full suite.

## Source commits

- `2691d006` fix(deepseek): preserve Responses reasoning replay on continuations (#875)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of reasoning content for supported models, preserving valid content during follow-up requests.
  * Removed proxy-generated encrypted reasoning data that should not be forwarded.
  * Improved continued tool-call requests and fallback behavior for eligible models, including DeepSeek.
  * Prevented unnecessary request updates when no content changes are needed.

* **Tests**
  * Added coverage for reasoning replay, tool-call continuation, proxy envelope removal, and model-specific behavior.

### DIFF
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -1,6 +1,6 @@
 import { createHash } from "node:crypto";
 import type { IncomingMeta, ProviderAdapter } from "./base";
-import { namespacedToolName, type AdapterEvent, type OcxParsedRequest, type OcxProviderConfig, type OcxUsage } from "../types";
+import { modelInList, namespacedToolName, type AdapterEvent, type OcxParsedRequest, type OcxProviderConfig, type OcxUsage } from "../types";
 import { catalogModelSupportsReasoningSummaries } from "../codex/catalog";
 import { COMPACT_PROMPT, decodeCompactionSummary, SUMMARY_PREFIX } from "../responses/compaction";
 import { collectResponsesToolGroups } from "../responses/tool-groups";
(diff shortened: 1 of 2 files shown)


lidge-jun:
Closing as a functional duplicate: the proxy half of #875 already landed in #892 (wt3, b42d57333) — the registry seeds `preserveResponsesReasoningContent: true` for DeepSeek and the passthrough sanitizer keeps raw reasoning content for it, with the same regression file (tests/deepseek-reasoning-replay.test.ts). Your per-model `preserveReasoningContentModels` list is the finer-grained mechanism; if the provider-wide flag ever proves too broad, that refinement is welcome as a follow-up. Thanks for the careful write-up.

PR #912, opened 2026-08-02
fix(openai-chat): deliver tool-result images to vision models

### LINKED ISSUE #888: multimodal image forwarding seems to be broken (kimi k3 + claude code)
### Client or integration

Claude Code

### Area

Proxy and routing

### Summary

Image

test image:
Image

prompt:
> /Users/yeowool/Desktop/test.jpg
> fetch this image from macbook air and describe it

assistant response:
> It's the classic pink feather-winged pig clip art: a plump pink pig standing in profile with tiny cartoon wings on its back, plus a small pink wing icon in the bottom-left corner and the caption "Not Today" along the bottom — a "when pigs fly" joke image.

### Reproduction

1. ocx claude
2. make the agent fetch image from remote sources
3. make the agent read and describe the image
4. watch the agent hallucinate out of thin air

### Version

2.8.0

### Operating system

macOS 15.7.4

### Provider and model

kimicode/k3-256k

### Logs or error output

```shell

```

_No response_

### Redacted configuration

```json

```

### Checks

DevMello:
## Summary

Fixes #888. Images in a tool result were flattened to the literal string `"[image]"`, so vision models like kimi-k3 never saw the actual image and made up a description (Claude Code sends Read and screenshot output this way).

`role:"tool"` messages can only carry text on chat providers, so the adapter now re-sends the image parts in a user vision message once the tool round closes, the same approach the Google and Kiro adapters already use. Models in `noVisionModels` still go through the vision sidecar.

Also added a line to the adapter docs. The Japanese, Korean, Russian, and Chinese versions were done with AI translation, so a quick check from native speakers would help.

## Verification

- `bun run test` (includes the new `tests/openai-chat-tool-result-images.test.ts`), `typecheck`, `lint:gui`, `privacy:scan`.
- Reproduced #888 against a mock upstream: the image data URL now reaches the provider instead of `"[image]"`.

## Checklist

## Notes

Rebased onto current `dev` (was accidentally based on the `dev`/`main` release-promotion point for v2.10.0). Source commit before the rebase: a719509e.

## Summary by CodeRabbit

* **New Features**
  * Tool-result images are now preserved when using the OpenAI Chat adapter.
  * Images are sent in a follow-up user vision message, while the original tool message retains an `[image]` marker.
  * Supports multiple tool calls, interrupted rounds, and orphaned tool results.

* **Documentation**
  * Updated adapter documentation in English, Japanese, Korean, Russian, and Simplified Chinese to describe image handling.


lidge-jun (approved):
Approving. Verified on a merged scratch tree against `dev` at `fa51fce54`.

**Why this is the right fix for #888.** The reporter's route was traced end to end: Claude Code sends an Anthropic `tool_result` → `src/claude/inbound.ts:96-113` preserves the images as `input_image` → `src/responses/parser.ts:547-555` produces an internal `toolResult` with structured image parts → Kimi OAuth (`src/providers/registry.ts:712-741`) and Kimi API-key (`:1428-1444`) both resolve to `openai-chat`, which flattened it through `contentPartsToText`. That flattening is the `[image]` placeholder in the report.

Carrying the images on a following `user` message is also the correct shape rather than a workaround — the Chat Completions schema accepts only strings or text parts in a `tool` message, and this repository already does the same thing elsewhere: `src/adapters/google.ts:196-208` puts `inline_data` beside the `functionResponse` in a user turn, and `src/adapters/kiro.ts:511-531` uses a corresponding user carrier.

**Evidence.**

- Merge onto current `dev` is clean; 7 files, +207/−1.
- Focused suite: 5 pass / 0 fail (31 assertions).
- **Ablation** — with `src/adapters/openai-chat.ts` reverted to `dev` and the tests kept: 4 fail / 1 pass, and the survivor is exactly the image-free control. The test is not tautological.
- `bun x tsc --noEmit` exit 0.
- `tests/openai-chat-eof.test.ts` + `openai-chat-hardening.test.ts` + `responses-parser.test.ts`: 74 pass / 0 fail, confirming the finish-less-EOF handling from #896 is untouched. This mattered because #880 rewrote ~256 lines of the same file an hour before; the hunks land in the right regions after that refactor.
- Full CI green on `d0a525d7`.

**Checked and cleared, for the record.** Vision gating ordering is correct — the sidecar substitutes or fail-closed strips before `adapter.buildRequest`, and `carriesImages()` includes `toolResult` (`src/vision/index.ts:187-194`), so a `noVisionModels` target never reaches the new carrier with a raw image. No `openai-chat` provider in the registry enforces role alternation, so the `assistant → tool → user(carrier) → user` sequence is safe; Kiro is the only alternation-enforcing transport and it has its own adapter. Nothing post-processes what `messagesToChatFormat` returns.

**One follow-up, not blocking.** The five translated doc bullets describe direct `image_url` forwarding without noting that `noVisionModels` models get sidecar-generated text instead. Worth a sentence later.

Thanks for a clean fix with a test that actually drives the path.

PR #837, opened 2026-08-01
fix: integrate hosted image tool preferences

Ingwannu:
## Summary

- integrate the hosted image-tool preference work from #616 onto current `dev`
- preserve the original contributor commit and authorship from @Eleven-is-cool
- resolve the two documentation conflicts by retaining the newer DeepSeek wire-default wording alongside `modelPreferHostedTools`
- update one expected schema to account for the function-schema normalization already on `dev`

The original PR branch does not allow maintainer edits and is now conflicting, so this integration branch carries the reviewed change without rewriting or squashing the contributor commit.

## Test plan

- `bun test tests/openai-responses-passthrough.test.ts tests/management-provider-validation.test.ts tests/openai-api-virtual-models.test.ts tests/config.test.ts` — 228 passed
- `bun run typecheck`

## Decision Log

- Purpose: land the validated hosted-tool behavior while preserving contributor attribution.
- Constraint: #616 is conflicting and `maintainerCanModify` is disabled.
- Chosen approach: replay the original authored commit onto current `dev`, resolve only current-dev integration conflicts, and retain the original commit as the first commit in this PR.
- Tradeoff: a small integration PR replaces direct merging of #616, but avoids force-pushing a contributor branch and keeps history/auditability intact.

Supersedes #616 after merge.

PR #924, opened 2026-08-03
fix: preserve hosted image tool preferences (#837 rebased, two defects fixed)

lidge-jun:
Rebases #837 onto current `dev` and fixes the two defects a plan audit found in it. Supersedes both #837 and #616; the substantive commit keeps @Eleven-is-cool as author.

## Provenance

#616 and #837 are the same implementation, not two competing ones. Git authorship shows it:

| | #616 | #837 |
|---|---|---|
| Substantive commit | `1aba0e4b` | `89d51dbc` |
| Author | Eleven-is-cool | Eleven-is-cool |
| Authored at | 2026-07-28T12:23:13Z | 2026-07-28T12:23:13Z |
| Committer | Eleven-is-cool | Ingwannu |

@Ingwannu replayed @Eleven-is-cool's commit onto a newer base, preserving authorship, and added a fixture correction that only applies on that base. This branch does the same thing once more onto `fa51fce54`, so the credit chain is intact: `git log` still shows @Eleven-is-cool as author of the implementation.

## What it does

Non-forward Responses gateways that reserve `image_gen` server-side reject even the empty client namespace `normalizeImageGenClientTools()` preserves, so a user has no way to say "use the hosted tool for this model." This adds an exact-model opt-in:

```jsonc
"modelPreferHostedTools": { "some-model": ["image_generation"] }
```

It strips colliding client `image_gen` declarations from `tools` and nested `additional_tools`, rewrites forced selectors to `{ type: "image_generation" }`, and restores hosted image generation if stripping removed the only image declaration. The Spark exclusion table moves out of the adapter into `src/responses/hosted-tool-policy.ts` so config validation can consult it too.

## Two defects fixed on top

Both were reproduced red before fixing.

**Validation disagreed with routing** (`src/config.ts`). The validator resolved the effective wire from `registry.adapter` plus an explicit `modelAdapters` entry, then stopped. At request time `resolveModelAdapter()` consults the registry's per-model `modelWireDefaults` first. DeepSeek routes `deepseek-v4-flash` over native Responses for a Responses inbound while its provider-wide wire stays `openai-chat`, so a valid preference for that model was rejected at config load with `requires the openai-responses wire`. The validator now walks the same order the runtime does.

**Inherited keys threw before dispatch** (`src/adapters/openai-responses.ts`). `provider.modelPreferHostedTools?.[modelId]` walks the prototype chain, so a routed model id of `constructor` or `toString` resolved to a function and threw `TypeError: ... .includes is not a function`, failing the request before it reached upstream. Lookup is own-property only and array-checked now.

## Rebase

Six conflicts, all from `dev` splitting `configuration.md` into `configuration/providers.md` after this branch was written. The doc row was placed in its new home in all five locales; `src/config.ts` was a parallel import addition.

## Verification

- `bun run test` — **7574 pass, 8 skip, 0 fail** across 504 files
- Focused: 247 pass across `config`, `management-provider-validation`, `openai-api-virtual-models`, `openai-responses-passthrough`
- Both new regressions driven red first: the DeepSeek case produced the exact rejection quoted above, the inherited-key case the exact TypeError.

## Disposition

Closes #837 and #616 once merged. Thanks @Eleven-is-cool for the implementation and @Ingwannu for the first integration.

## Summary by CodeRabbit

* **New Features**
  * Added `modelPreferHostedTools` configuration to enable hosted `image_generation` for specific compatible models.
  * Added support for virtual OpenAI Pro model resolution and model adapter precedence.
* **Bug Fixes**
  * Improved handling of conflicting image-generation tools and tool selection.
  * Prevented unsupported hosted tools from being sent to incompatible models.
  * Added validation for unsupported configurations and incompatible providers.
* **Documentation**
  * Documented the new configuration, compatibility requirements, and model-resolution behavior across supported languages.


lidge-jun:
Both automated reviews found real defects. Fixed, each driven red first.

**Codex, `src/config.ts`** — a `preserveCustomDestination` registry row reused under a different endpoint keeps its own adapter at runtime, because `routedProviderConfig()` honors `providerMatchesRegistryTransport()`. `volcengine-agent-plan` is exactly that: an `openai-responses` registry row where a config naming the id but pointing elsewhere with `adapter: "openai-chat"` passed validation on the registry's adapter while the Responses adapter never ran. The preference was accepted and silently ignored.

This is the mirror of the defect the previous commit fixed — that one made validation too strict, this one too loose. Validation now starts from `provider.adapter` unless the configured transport still matches the registry's. Red evidence: the config loaded with source `file` before, returns `requires the openai-responses wire` now (`57f0ec5c3`).

**CodeRabbit, `src/adapters/openai-responses.ts`** — stripping walked every `additional_tools` container but restoration targeted only the first stripped index, so a request with two containers left the second with an empty tool list and no image capability. Now tracks every stripped index. Red evidence: the second container came back as `[]` (`396cf1cef`).

CodeRabbit's second note — that `resolveEffectiveWire` duplicates the selection order in `src/server/adapter-resolve.ts` — is correct and I am not fixing it here. Two independently maintained copies of one algorithm is precisely how this PR's first defect happened, so it deserves an extraction, but doing it inside a contributor's rebase would widen the diff past what this change is about. Worth its own PR.

Verification after both fixes: `bun run test` **7575 pass, 8 skip, 0 fail** across 504 files; `bun x tsc --noEmit` exit 0.

PR #927, opened 2026-08-03
fix(compact): try one alternate account on a pool 429/402, and keep the backoff headers

### LINKED ISSUE #913: [Bug]: /v1/responses/compact returns pool 429/402 without same-request alternate attempt
### Client or integration

Codex App

### Area

Proxy and routing

### Summary

`/v1/responses/compact` does not have the bounded same-request alternate-account handling already used by regular `/v1/responses`.

An existing thread should remain on its affined account even when WHAM displays `100%`: Codex can continue an in-progress task until upstream actually rejects it. Today, when that account returns a pre-response `429` or `402`, the compact path records the outcome and returns it directly. Codex then retries the remote compact task outside the logical request and can end with `exceeded retry limit`, although another pool account is eligible.

Expected behavior:

- A `100%` usage snapshot alone must not move an existing thread or compact request.
- After an actual upstream `429/402`, record the failed account and try exactly one eligible alternate in the same compact request.
- Rebuild the alternate bearer/account headers and preserve quota scope, writer generation, affinity, reset headers, `Retry-After`, and cancellation.
- Cancel the first response body before the alternate send.
- If no alternate exists, preserve the original rejection body and headers.
- If the alternate also rejects, stop after two total sends. No recursion or reset-credit redemption.

Relevant source:

- `src/server/responses/compact.ts`: `handleResponsesCompact()` and compact pool outcome recording.
- `src/server/responses/core.ts`: existing regular-Responses bounded `429/402` alternate flow.
- `src/codex/auth-context.ts`: `excludeAccountId` resolution.
- `src/codex/routing.ts`: alternate selection and upstream outcome state.

Required regression coverage in `tests/responses-compaction-routing.test.ts`: `429` and `402` alternate success, no-alternate preservation, alternate rejection capped at two sends, abort before retry, correct alternate credentials, and proof that `100%` without upstream rejection leaves existing affinity unchanged.

Related: #584/#585 (regular Responses parity), #815/#834 (known `100%` is not unknown for new routing decisions), #657 (advisory usage vs actual rejection), and #433 (cooldown recovery).

### Reproduction

1. Configure two Codex pool accounts and bind a parent thread to account A.
2. Keep account B eligible.
3. A displayed quota may reach `100%`; do not pause or move the existing thread on that value alone.
4. Make A return an actual pre-response upstream `429` or `402`.
5. Send `/v1/responses/compact` with the same `x-codex-parent-thread-id`.

Current: A's rejection is returned directly and Codex can exhaust its remote compact retry budget.

Expected: A is attempted once, the real rejection authorizes one attempt on B, and B's successful compact response is returned in the same HTTP call.

### Version

`2.10.0`; the relevant compact, routing, auth-context, and retry sources were unchanged on `dev` at `075c2f34c060d068ed56985e76924a9f23aca39d` when checked on 2026-08-03.

### Operating system

Windows 11 Pro 10.0.26200 (platform-independent routing defect).

### Provider and model

OpenAI (Codex login / forward account pool); native Codex models.

### Logs or error output

```text
Error running remote compact task: exceeded retry limit, last status: 429 Too Many Requests
```

None. Account and request identifiers are intentionally omitted.

### Redacted configuration

```json
{
  "providers": {
    "openai": {
      "authMode": "forward",
      "codexAccountMode": "pool"
    }
  },
  "accountPoolStickyLimit": 1,
  "autoSwitchThreshold": 100,
  "upstreamFailoverThreshold": 3,
}
```

### Checks

lidge-jun:
Closes #913. Reported by @luvs01.

`/v1/responses` already answers a pre-stream 429/402 by trying one eligible alternate account inside the same logical request — `retryCodexPoolOnAlternateAccount()` at `src/server/responses/core.ts:319-423`, activated before streaming at `:1679-1721`, recognizing exactly 429 and 402.

`/v1/responses/compact` had none of it. It resolved one context, sent once, and returned the rejection. The client then retried the compact task *outside* the logical request, which is how a session ends up reporting exhausted retries while another pool account sat idle.

## Three things had to change together

**The recorder took its account from a closure.** `recordCompactPoolOutcome()` captured a single `authCtx`, so it could not express "record this against A and promote B", and could not record anything against B at all. It now takes the context explicitly; every existing call site passes the same `authCtx` it used before, so that part is a mechanical widening — but it is a prerequisite, not a detail.

**The two accounts need different recovery, and this is the part worth reviewing.** Compact's send goes through `fetchWithTransientRetry()`, which makes up to three status attempts, each wrapping its own reset retries. So "send the alternate exactly once" and "preserve A's existing recovery" contradict each other unless the modes differ. A keeps the full ladder; B gets a single direct `fetchWithHeaderTimeout()`. That is exactly what the regular path does at `core.ts:396`, and the asymmetry is deliberate there too — A's retries happen before any alternate is considered, and the alternate is a last bounded try rather than a second retry ladder.

**`bufferCompactResponse()` dropped the backoff headers.** It rebuilt the response with only `Content-Type`, discarding `Retry-After` and the reset hints from the very rejection a client needs them for. It now carries a narrow allowlist plus the upstream `statusText`.

Ordering detail: the alternate is built completely before A's body is cancelled. If construction throws, A's rejection is still intact and returnable.

## Not in scope

The quota strategy deliberately rebinds an over-threshold thread (`src/codex/routing.ts:1179-1200`), and tests cover that. The issue's framing suggests a 100% WHAM snapshot should never move an existing thread under any strategy — that is a policy change with its own tradeoffs, not part of endpoint parity, so it is not smuggled in here.

## Evidence

Six regressions, with the **send count as the activation proof** throughout — one send means the branch never fired, three means it recursed:

- 429 and 402 each try exactly one alternate, with different bearers on the two sends
- no eligible alternate: exactly one send, A's rejection returned with `Retry-After` and reset headers intact
- alternate also rejects: exactly two sends, the second rejection returned
- a 400 does not trigger an alternate
- an abort between attempts prevents the second send

**Ablation**: reverting `src/server/responses/compact.ts` to `dev` while keeping the tests fails exactly 4 of the 6 — the two negative controls still pass, which is what makes them controls.

- `bun run test`: 7581 pass / 0 fail / 8 skip across 504 files
- `bun x tsc --noEmit` exit 0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compact request reliability by retrying once with an alternate account after eligible rate-limit or payment-related failures.
  * Preserved relevant response status, retry, and reset information when requests fail.
  * Avoided unnecessary failover for unrelated errors and respected request cancellations.
  * Improved per-account health tracking for successful, failed, and canceled requests.

---

## Review round

An adversarial review found three blockers, all fixed in `361b329a3` and `20f4f6dd0`.

**A's quota cache went stale.** The regular path applies the rejected account's upstream quota headers before recording it (`core.ts:349-357`) — a 429 carries the snapshot that produced it. Compact skipped that, so quota-strategy routing and the dashboard kept reading numbers from before the account ran out. Now done in the same order.

**Cancellation could race alternate resolution.** The only abort check sat before the await, and resolution can wait on a credential refresh. A client that went away during that window still got A recorded, A's body cancelled, and B's fetch invoked. Native fetch rejects an aborted signal, but a custom executor need not, and B's quota is not ours to spend on a request nobody is waiting for. Re-checked after resolution, releasing B's probe lease on the way out.

**The tests did not lock the crux they were written for.** This is the one worth naming: all six original tests could pass with B switched from `"single"` back to `"normal"` recovery, because B always returned 200 and never stimulated the retry ladder. Three more tests now cover it — B returning a transient 5xx must still be one send, A returning a transient 5xx must still be retried in place, and each account's health must record its own outcome.

Verified by mutation: forcing B onto the normal ladder fails exactly those three single-send assertions and nothing else.

Updated totals: 9 regressions, `bun run test` 7586 pass / 0 fail across 504 files, typecheck exit 0, privacy scan passed.

### DIFF
--- a/src/server/responses/compact.ts
+++ b/src/server/responses/compact.ts
@@ -27,7 +27,7 @@ import {
 import { isInjectionDebugEnabled } from "../../lib/debug-settings";
 import { injectionDebugLog } from "../../lib/injection-debug-log";
 import { modelInList, namespacedToolName } from "../../types";
-import type { AdapterEvent, OcxConfig, OcxParsedRequest, OcxProviderConfig, OcxProviderContinuationState, OcxUsage } from "../../types";
+import type { AdapterEvent, CodexAccountMode, OcxConfig, OcxParsedRequest, OcxProviderConfig, OcxProviderContinuationState, OcxUsage } from "../../types";
 import {
   forceRefreshOAuthAccessSnapshot,
   getOAuthCredentialApiBaseUrl,
@@ -59,7 +59,12 @@ import {
   recordCodexUpstreamOutcome,
   type CodexUpstreamOutcome,
 } from "../../codex/routing";
-import { fetchWithResetRetry, fetchWithTransientRetry, applyUpstreamRecoveryInit } from "../../lib/upstream-retry";
+import {
+  fetchWithResetRetry,
+  fetchWithTransientRetry,
+  applyUpstreamRecoveryInit,
+  type UpstreamSendRecovery,
+} from "../../lib/upstream-retry";
 import { ForwardAdmissionCredentialError, validateForwardAdmissionCredential } from "../auth-cors";
 import { listOpenAiForwardSidecarCandidates, resolveFirstUsableOpenAiSidecar, type ResolvedOpenAiForwardSidecar } from "../../providers/openai-sidecar";
 import { isCanonicalOpenAiForwardProvider, supportsNativeResponsesCompactEndpoint } from "../../providers/openai-tiers";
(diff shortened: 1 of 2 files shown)


lidge-jun:
Added one test (`361b329a3`): the alternate gets a single send even when it returns a transient 5xx.

The existing matrix asserts send counts for the 429/402 success path, the no-alternate path, the both-reject path, and the abort path — all good — but none of them drives the alternate into a **transient** 5xx, and that is the only case where `recovery: "single"` behaves differently from `recovery: "normal"`. A wiring mistake that left the alternate on `fetchWithTransientRetry` would have passed every test in the file.

Ablation: flipping the alternate's mode back to `"normal"` takes the send count from 2 to 6 — one for A, three for B, across both rejection codes — and the test fails. So the assertion is load-bearing rather than decorative.

Everything else here matches the plan in `devlog/_plan/260803_pr_issue_sweep/030_phase3_compact_alternate.md` as written, including the three prerequisites that plan called out: the explicit-context recorder (the closure could not express `promoteAccountId` for A and an outcome for B), building the alternate completely before cancelling A's body, and preserving A's outcome recording on the no-alternate path. `21 pass, 0 fail` on the file, typecheck clean.

lidge-jun:
Added CodeRabbit's requested case (`8236ffecb`): a bound thread whose affined account reads 100% locally, with the upstream answering normally. One send, 200 returned, no alternate resolved.

It is the scope guard the plan named, and the reason it earns its place is that no existing test would catch its absence. A cached 100% is what the account looked like on its last WHAM read, not a rejection — if the alternate gate ever widened to consult local quota, this request would send twice and every other test in the file would still pass.

Verified against the existing quota-strategy behavior rather than assumed: the plan explicitly rules out changing the rebinding at `src/codex/routing.ts:1179-1200`, and this test pins the compact path to that boundary.

`25 pass, 0 fail` on the file, typecheck clean.

PR #934, opened 2026-08-03
fix(openai-responses): mirror web_search_call query into queries

### LINKED ISSUE #930: DeepSeek V4-Flash native Responses route: 400 missing field `queries` on replayed web_search_call
``markdown
### Client or integration

Codex App

### Area

Proxy and routing

### Summary

After a web search in a conversation using `deepseek/deepseek-v4-flash` (opencodex 2.10.0, DeepSeek native Responses route), every subsequent turn fails with HTTP 400 `invalid_request_error` from `api.deepseek.com`:

`Failed to deserialize the JSON body into the target type: input: missing field 'queries' at line 1 column N`

I expected the replayed `web_search_call` history to be accepted by DeepSeek's Responses API (or dropped on translated routes) so the conversation can continue.

### Reproduction

1. Start opencodex 2.10.0 (`ocx start --port 10100`) with the DeepSeek provider and model `deepseek/deepseek-v4-flash` on the native Responses route (`responsesPath: "/responses"`).
2. In the Codex App, run a turn that performs a web search, so a `web_search_call` history item is recorded.
3. Send any follow-up message in the same conversation.
4. Every subsequent turn fails with 400 `missing field 'queries'`.

Minimal repro against the proxy — POST `/v1/responses` with `model: "deepseek/deepseek-v4-flash"` and input containing:

```json
{"type":"web_search_call","id":"ws_x","status":"completed","action":{"type":"search","query":"test"}}
```

→ 400 `missing field 'queries'`.

Changing the action to `{"type":"search","queries":["test"]}` → 200.

### Version

2.10.0

### Operating system

Windows (build 26200, DisplayVersion 25H2)

### Provider and model

deepseek / deepseek-v4-flash

### Logs or error output

```
HTTP 400 invalid_request_error from api.deepseek.com:
Failed to deserialize the JSON body into the target type: input: missing field 'queries' at line 1 column N
```

_None._

### Redacted configuration

_Not applicable — provider adapter routing for `deepseek/deepseek-v4-flash` as shipped in 2.10.0 (`modelWireDefaults` + `responsesPath: "/responses"`)._

### Checks

```

DevMello:
## Summary

Fixes #930. Replayed `web_search_call` history carries a singular `action.query`, but DeepSeek's Responses route requires `queries`, so one web search made every later turn in the conversation fail with 400. The key path now adds `queries` next to the singular; forward mode is untouched.

## Verification

- New tests in `tests/openai-responses-passthrough.test.ts`: the mirror, forward mode staying untouched, and pass-through of everything else.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with DeepSeek for replayed web search actions.
  * Preserved existing hosted search request formats and unrelated input data.

* **Tests**
  * Added regression coverage for search actions with singular or multiple query fields, actionless calls, and forward-mode requests.

### DIFF
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -137,6 +137,26 @@ function stripItemIdsWhenUnstored(body: unknown): unknown {
 }
 
 /**
+ * Replayed web_search_call actions carry a singular `query`, but DeepSeek's Responses route
+ * requires `queries` and 400s without it (#930). Add the plural next to the singular. Key path
+ * only: forward mode keeps the hosted shape.
+ */
+function normalizeWebSearchCallActions(body: unknown): unknown {
+  if (!isPlainObject(body) || !Array.isArray(body.input)) return body;
+
+  let changed = false;
+  const input = body.input.map(item => {
+    if (!isPlainObject(item) || item.type !== "web_search_call") return item;
+    const action = item.action;
+    if (!isPlainObject(action) || typeof action.query !== "string" || action.queries !== undefined) return item;
+    changed = true;
+    return { ...item, action: { ...action, queries: [action.query] } };
+  });
+
+  return changed ? { ...body, input } : body;
+}
+
+/**
  * Replace proxy-minted compaction items (`encrypted_content` starting with `ocx1:`) with plain
  * user messages before forwarding to the ChatGPT backend. Our envelope is transparent base64, not
  * OpenAI encryption — the native backend cannot decrypt it and would reject the request. Real
@@ -1140,6 +1160,7 @@ export function createResponsesPassthroughAdapter(provider: OcxProviderConfig): 
           parsed._openAiVirtualSelectedModelId,
         );
         outBody = normalizeImageGenClientTools(outBody);
+        outBody = normalizeWebSearchCallActions(outBody);
       }
       if (forward || parsed._previousResponseInputExpanded === true) {
         outBody = repairOversizedReplayCallIds(outBody);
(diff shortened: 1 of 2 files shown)


lidge-jun:
Closing as already fixed — #932 landed the same repair about an hour before this opened (`6a7351b4d`), and I'm sorry you spent the effort on a duplicate.

Your diagnosis was identical and correct: replayed `web_search_call` actions carry only singular `query`, DeepSeek requires `queries`, and the plural has to be added next to the singular rather than replacing it. Your `normalizeWebSearchCallActions()` and the `backfillWebSearchQueries()` now on `dev` are the same function.

Two differences worth naming, since one of them is a real judgment call:

**Placement.** Yours runs inside the key-auth branch, with the reasoning that forward mode keeps the hosted shape. Mine runs unconditionally. I went that way because a review confirmed OpenAI's current schema carries optional `query` and optional `queries` on the same object with no exclusivity — the official Node SDK models both independently, and codex-rs has a round-trip fixture containing both. So the canonical forward path tolerates the extra key, and running unconditionally also covers a noncanonical forward provider that happens to be strict. If you think forward mode should stay byte-identical on principle rather than on tolerance, that is a fair argument and worth making.

**Creation-time half.** #932 also changed `webSearchAction()` in the bridge so newly created single-query items carry both keys from the start. The adapter repair handles history recorded before that; the bridge change stops new history from needing repair at all. Notably the batch case deliberately does *not* gain a singular `query` — codex-rs renders the plural form as `"<first> ..."` only when `query` is absent and `queries.len() > 1`.

You also filed #912, which is on `dev` now and closed #888. Thanks for both — the image-forwarding fix in particular had a clean ablation and made a review easy.

DevMello:
No problem! Glad someone got to fixing it!

PR #936, opened 2026-08-03
fix: harden credential and runtime trust boundaries (rebase of #916)

lidge-jun:
Rebase of #916 (@Ingwannu) onto current `dev`, with the three defects a branch review found.

**Draft on purpose.** This touches authentication, credential handling, GitHub workflows, and the durable launcher — `MAINTAINERS.md` requires explicit security review for exactly that surface. The review below is an audit, not that review, and an agent does not substitute for it. Do not merge on my say-so.

## The defects are real on current `dev`

Confirmed by direct probe rather than by reading the PR description:

- **Vertex authority injection.** `src/adapters/google.ts` builds the request host from `provider.location` with no validation, then attaches an ADC bearer. A location of `attacker.example:443/capture#` yields authority `attacker.example` — I reproduced this against `dev`.
- **Durable Bun executable injection.** `overrideBunPath()` rereads `OPENCODEX_BUN_PATH` after Bun has loaded project dotenv, so every `durableBunRuntime()` caller — service, shim, WinSW, tray — can persist a repository-chosen executable.
- **Claude destination injection.** An ambient `ANTHROPIC_BASE_URL` survived credential stripping and reached Claude, which can redirect an OAuth-bearing request.
- **Management-token disclosure.** `src/oauth/health.ts` attached the admin token after only a forgeable `/healthz` identity check.

None of these were superseded by #917. That PR resolved the *inbound* management principal — who may call the star route. This is the *outbound* listener: whether the thing answering on the port deserves the token. Different boundary.

## What the review changed

**`src/cli/claude.ts` no-context fallback — narrowed.** As submitted, a launch with no trusted launcher context deleted all three ambient Anthropic slots. `bun src/cli/index.ts` is a documented entry point (`structure/01_runtime.md:9`) and has no launcher context, so a user with a shell-exported `ANTHROPIC_API_KEY` simply lost it.

The two slot classes are not symmetric. `ANTHROPIC_BASE_URL` stays fail-closed — a dotenv-only destination plus subscription auth is precisely how the OAuth bearer leaves for a repository-chosen host, and losing a legitimate custom destination costs a flag rather than an account. Credentials are preserved: the destination is already pinned by the time they are read, so stripping them defends against a project file that could equally well have supplied the key being blamed. The test that mandated the old behavior is replaced by two that pin the asymmetry.

**`structure/06_docs-and-release.md`** — the `ci.yml` row still described the pre-#899 hosted-Windows selector. Rewritten against the real job graph: four Linux shards plus `gates`, full macOS, Windows only at the shipping boundary, and the aggregate `ci` job asserting Windows actually succeeded there.

**`structure/01_runtime.md`** decision log — recorded the rejected fail-closed-for-everything behavior; now describes the split.

## Rebase

One conflict, `src/server/index.ts`, resolved keeping both #917's `managementPrincipal()` dispatch and this branch's attestation imports.

## Evidence

- `bun run test`: 7602 pass / 0 fail across 505 files
- `bun x tsc --noEmit` exit 0, `bun run privacy:scan` passed
- Focused security suites: 124 pass / 0 fail across the seven affected files

Authorship on the original commits is preserved. @Ingwannu — the substance is yours; the three changes above are the review response, and I would rather you disagree with them here than have them land silently.

---

## Security review round — my narrowing was wrong

The adversarial review returned FAIL on the one change I made to @Ingwannu's behavior, and it was right.

I had preserved ambient credentials when no launcher context exists, reasoning that the destination is pinned before they are read, so a dotenv-supplied key could only ever reach the local proxy. **That reasoning does not survive the subscription path.** `CLAUDE_CODE_PROVIDER_MANAGED_BY_HOST` is only set when opencodex owns an auth token — asserting it otherwise logs a subscriber out (#253) — so Claude Code's `settings.env` merge can still replace `ANTHROPIC_BASE_URL` *after* `buildClaudeEnv` returns. A preserved key travels to that host.

This repository already documents the destination residual for subscription mode, with a test asserting it so it cannot drift into an assumed guarantee. Preserving credentials would have quietly widened that documented residual into a credential leak.

So all three slots fail closed again without provenance, exactly as @Ingwannu submitted it. Direct `bun src/cli/index.ts` loses ambient Anthropic values — a real cost to a documented entry point, and the honest trade. The escape hatch is running through the published `ocx` bin, where genuine shell exports survive by proof.

**The gap that let my bad revision pass:** the suite tested no-context credential handling and settings-hijack *separately*, never combined. It does now — a no-context ambient key must be absent after the merge that hijacks the destination. Reintroducing the narrowing fails 6 tests.

The review also cleared the rest: the Vertex validator is a strict lowercase 1–63 byte DNS label rejecting dots, Unicode, percent-encoding, fragments, paths, uppercase and overlong values while accepting every legitimate location, enforced on both the adapter and the management write path; local attestation binds a fresh 256-bit challenge plus PID and port, fails closed on malformed proofs, and compares timing-safely; the Bun launcher stamps before Bun starts on every documented invocation including the Windows `.cmd` shim; and the rebase kept `managementPrincipal()` with the correct `handleManagementAPI` arity.

Updated: `bun run test` 7602 pass / 0 fail across 505 files, typecheck exit 0, privacy scan passed.

**Still draft, still needs maintainer security review.** Two rounds of adversarial audit are not that review.

## Summary by CodeRabbit

* **Security & Reliability**
  * Added listener verification for local management actions, helping ensure status and diagnostics communicate with the intended running proxy.
  * Improved protection against unintended credential and configuration leakage during Claude launches.
  * Strengthened runtime selection and launch validation for more predictable operation.

* **Bug Fixes**
  * Google Vertex locations are now validated before requests are sent, with clearer configuration errors.

* **Documentation**
  * Updated runtime, management, and CI documentation to reflect current behavior and operational safeguards.


lidge-jun:
@Ingwannu @Wibias — requesting the security review this has been waiting on since 07-27.

Surfaced by a triage sweep over every open issue and PR (`devlog/_plan/260805_issue_pr_triage/`). This PR has been recorded as needing explicit security review in four separate devlog rounds, and in none of them was the blocker technical — it was simply unassigned. Reviewers are now set so it stops falling off the board.

State against `origin/dev = aaa71967a`:

- draft, `CONFLICTING`, 620 commits behind, `label` check CANCELLED
- touches the credential and runtime trust boundary, which `MAINTAINERS.md` puts under mandatory security review
- authored by @lidge-jun, so self-approval is not available

Two things worth separating. The rebase is mechanical and mine to do — please don't spend review time on the conflicts. What is **not** unblocked by a rebase is the boundary decision itself, and that is the part I'm asking for: even a perfectly rebased branch cannot merge without this review.

For context the PR body already makes explicit: this is the *outbound* listener question — whether the process answering on the port deserves to — and it was not superseded by #917, which resolved the *inbound* management principal. Reviewing it as a duplicate of #917 would miss the change.

lidge-jun:
Brought this branch onto current `dev` (`b3a1d90a8`) via merge commit `a90981e67`. Two conflicts, both resolved compositionally: `src/server/auth-cors.ts` keeps dev's `redactSecretString` alongside this branch's Vertex `effectiveGoogleMode` validation, and `src/server/index.ts` folds the branch's `localAttestationSecret` into dev's new `StartServerDeps` seam (`deps.localAttestationSecret ?? createLocalAttestationSecret()`), with the CLI caller updated to the object form.

Independent audit of the merged result (regression + security): the four hardening claims still hold on the merged tree — Vertex location validation before ADC acquisition, Bun-path provenance validation in the launcher, fail-closed ambient Anthropic slots without launcher context, and health-attestation-gated management token. No duplicate hardening landed on `dev` since the branch point, so nothing double-applies. Three stale test call-sites for the old positional secret were aligned in `4874390dd`.

Verification: `bun x tsc --noEmit` clean, full suite 9076 pass / 0 fail (8 skip) across 579 files.

Still draft on purpose: the human security review required by `MAINTAINERS.md` for this surface has not happened, and this PR does not merge without it.

Issue #908, opened 2026-08-02
Cost estimates ignore published long-context pricing tiers (OpenAI >272k, xAI >=200k)


brunoflma:
### Area

Dashboard

### What are you trying to accomplish?

I want the `~$` column and `/api/usage` to be right for the requests that matter most — the big ones.

Several vendors reprice the **entire request** once the prompt crosses a token threshold. The two that affect me:

| Vendor | Threshold | Effect on the whole request | Source |
|---|---|---|---|
| OpenAI GPT-5.6 Sol / Terra / Luna | **>** 272,000 input tokens | 2× input, 2× cached input, 1.5× output | https://developers.openai.com/api/docs/pricing |
| xAI Grok 4.5 | **≥** 200,000 input tokens | 2× on every rate | https://docs.x.ai/developers/pricing |
| MiniMax M3 | **>** 512,000 input tokens | 2× on every rate | https://platform.minimax.io/docs/guides/pricing-paygo |

OpenAI publishes these as a separate "long context" row per model, not as a footnote — Sol is 5 / 0.50 / 30 short and 10 / 1.00 / 45 long.

Long requests are exactly the expensive ones, so an estimator that ignores the tier is wrong where being right is worth the most.

### What prevents this today?

`Cost4` is a flat four-tuple, and `resolveMatchedPrice(provider, modelId)` never sees a token count — so there is nowhere to express "this rate depends on how big the prompt is". Every request bills at the short rate regardless of size.

`applyPriorityMultiplier()` already establishes the shape for a conditional repricing step, but it keys off `serviceTier`, not off prompt size, so it cannot be reused as-is.

There is also a subtlety that makes this easy to get wrong: the threshold is measured on the **raw** `usage.inputTokens` the vendor counted, whereas `normalizeCostTokens()` subtracts cache read and write to produce billable input. A 280k-token prompt with a 200k cache read has 80k billable input but still crosses OpenAI's 272k threshold. Deciding the tier after normalization would silently under-bill exactly the cache-heavy long requests.

### What should OpenCodex do?

When a request's raw input token count crosses a model's published threshold, the whole request should be costed at that model's long rate, and the estimate should say so — a flag on `CostEstimate` so the dashboard can distinguish "long" from "standard" rather than just showing a larger number.

Models with no published tier must be completely unaffected, and the tier must compose with the existing Fast-mode multiplier instead of replacing it.

### Example usage or interface

[...]

PR #939, opened 2026-08-03
fix(claude): keep a marker for tool_result document blocks

DevMello:
## Summary

When a tool result contains a `document` block, the inbound translation dropped it and routed providers received an empty tool output, so the model concluded the tool returned nothing. A document in a plain user message already gets a `[document: title]` marker; the tool-result path now emits the same marker.

## Verification

- New test in `tests/claude-inbound.test.ts`: titled, untitled, and mixed text plus document results.
- Reproduced against a mock upstream: the tool message arrived as `""` before the fix and `[document]` after.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Document attachments in tool results are now represented correctly instead of producing empty output.
  * Document titles are preserved when available; untitled documents use a clear placeholder.

* **Tests**
  * Added coverage to verify document handling, text preservation, and parseable tool results.

### DIFF
--- a/src/claude/inbound.ts
+++ b/src/claude/inbound.ts
@@ -106,6 +106,10 @@ function toolResultOutput(block: Rec): string | Rec[] {
       } else if (item.type === "image") {
         const img = imageBlockToInputImage(item);
         if (img) out.push(img);
+      } else if (item.type === "document") {
+        // Same marker as the user-message document case below: the model should see the
+        // attachment happened instead of an empty tool output.
+        out.push({ type: "input_text", text: `[document${typeof item.title === "string" ? `: ${item.title}` : ""}]` });
       }
     }
     if (isError) out.unshift({ type: "input_text", text: "[tool error]" });
--- a/tests/claude-inbound.test.ts
+++ b/tests/claude-inbound.test.ts
@@ -175,6 +175,41 @@ describe("claude inbound translation", () => {
     expect(() => parseRequest(body)).not.toThrow();
   });
 
+  test("tool_result document blocks surface the attachment marker", () => {
+    const body = anthropicToResponsesBody({
+      model: "m", max_tokens: 10,
+      messages: [
+        { role: "assistant", content: [{ type: "tool_use", id: "t1", name: "Read", input: {} }] },
+        {
+          role: "user",
+          content: [{
+            type: "tool_result", tool_use_id: "t1",
+            content: [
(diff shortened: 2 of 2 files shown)


lidge-jun:
Carried onto the review stack as **#953** (stack 3/3), unmodified.

Your commits were taken with `git cherry-pick -x`, so they keep your authorship — `git log --format='%an'` on the stack branch shows you, not me. No content was changed; the diff on the stack is byte-identical to what you wrote here, and it applied to `dev` with no conflict resolution.

Verified on the stack: `bun x tsc --noEmit` exit 0, and the full suite at 7691 pass / 8 skip / 0 fail across 507 files.

**This PR stays open until #953 lands.** If a maintainer prefers to take yours directly instead, that path is unaffected — the stack commits get dropped and this one merges. Once #953 merges I'll close this as carried, with the credit already in the commit history rather than in a comment.

Stack: **#951** (plan, base `dev`) → **#952** (#908 long-context pricing) → **#953** (this carry). Review bottom-up.

Thanks for the fix.

Wibias:
Leave it as draft. Thanks. It doesnt need any more changes as you can see in juns comment on which you didn't respond to.

lidge-jun:
Closing this as **carried into #953**, not as rejected. Your fix is shipping — this is bookkeeping so the same change isn't sitting in the review queue twice.

Your commit(s) are on the stack branch as `c233a7cf6`, cherry-picked with `git cherry-pick -x`. Authorship is preserved, so `git log --format='%an'` shows you rather than me, and I verified the carried diff is byte-identical to yours rather than assuming it:

```console
$ git show <yours> | git patch-id --stable
$ git show <carried> | git patch-id --stable
IDENTICAL
```

No content was changed and no conflict resolution was needed.

One thing I want to be straight about: my earlier comment said I'd close this *after* #953 merged, and I'm closing it before. #953 is open and reviewable now, so closing early is my call as maintainer to keep the queue honest — not a signal that the review finished. **If the stack doesn't land, reopening this is one click**, and the branch is untouched.

Verified on the stack: `bun x tsc --noEmit` exit 0, full suite 7729 pass / 8 skip / 0 fail across 508 files.

Stack: **#951** (plan, base `dev`) → **#952** → **#953** (this carry) → **#954** → **#955**. Review bottom-up.

Thanks @DevMello — this was a clean fix and it's in.

PR #940, opened 2026-08-03
fix DeepSeek Responses UUID item ids for Codex

### LINKED ISSUE #938: [Bug]: Responses passthrough UUID item IDs leave Codex turn stuck on Thinking
### Client or integration

Codex App

### Area

Streaming

### Summary

When an `openai-responses` passthrough provider emits UUID item IDs instead of canonical Responses prefixes, the upstream request completes and OpenCodex records HTTP 200 plus reported usage, but Codex App never produces an assistant item or completes the turn. The UI remains on Thinking until the user interrupts it.

This is distinct from #864. The stream contains a complete `response.completed` event, and switching macOS from the default relay path to `streamMode: "eager-relay"` does not change the result.

The incompatible IDs observed on the wire look like:

```text
response.id  = b16df59f-bbc4-458f-80c6-eae4bac8dada
reasoning.id = 4ffa2f22-7283-43ae-ae18-7690736424c1
message.id   = 6d66799f-6b2b-4643-8d30-5176244a3a8b
```

Canonical OpenAI Responses item IDs use type prefixes such as `resp_`, `rs_`, and `msg_`. OpenCodex already has an opt-in `responsesItemIdRepair`, but it only repairs exact, preconfigured placeholder IDs or missing terminal IDs. It cannot repair randomly generated UUIDs that change on every request.

### Reproduction

1. Configure an API-key provider using `adapter: "openai-responses"` whose native Responses endpoint returns UUID IDs for response/message/reasoning objects.
2. Set `statelessResponses: true` and route a model through it.
3. Start a new Codex App turn with a minimal input such as `Reply only OK`.
4. Observe that the provider returns the complete sequence:

```text
response.created
response.output_item.added        # reasoning item with UUID id
response.reasoning_text.delta
response.output_item.done
response.output_item.added        # message item with UUID id
response.output_text.delta
response.output_text.done
response.output_item.done
response.completed                # status=completed, output and usage present
```

5. OpenCodex logs the request as `status: 200`, `usageStatus: "reported"`, and includes token usage from `response.completed`.
6. Codex App remains on Thinking. Its rollout contains `task_started` and the user input, but no assistant response item and no successful turn completion. Interrupting the UI produces `turn_aborted`.

I also repeated the request after setting:

```json
{"streamMode":"eager-relay"}
```

The Codex CLI JSON stream still emitted only `thread.started` and `turn.started`; it emitted neither an agent message nor `turn.completed`. This rules out the macOS `tee()` relay path as the cause.

Directly calling the same OpenCodex `/v1/responses` route with curl shows that the SSE stream and `response.completed` frame are complete; the differentiating protocol violation is the UUID item IDs.

### Version

2.10.0

### Operating system

macOS 26.5.2 (Darwin 25.5.0, Apple Silicon arm64)

### Provider and model

OpenAI-compatible native Responses provider / `deepseek-v4-flash` and `deepseek-v4-pro`

### Logs or error output

```shell
# OpenCodex persisted request summary
status=200
usageStatus=reported
inputTokens=14466
outputTokens=81
reasoningOutputTokens=49

# Codex rollout lifecycle
task_started
# no assistant response_item
# no turn_completed
turn_aborted reason=interrupted
```

### Suggested fix

Extend `ResponsesItemIdRepairConfig` with a provider-local option such as:

```json
{
  "responsesItemIdRepair": {
    "repairInvalidIds": true,
    "repairMissingTerminalIds": true
  }
}
```

In `responses-item-id-repair.ts`, when `repairInvalidIds` is enabled, rewrite any existing `message` or `reasoning` ID that does not start with its expected prefix:

```ts
const shouldRepair = state.placeholders[type].has(rawId)
  || (state.repairInvalidIds && !rawId.startsWith(REPAIRABLE_PREFIXES[type]));

const mapped = shouldRepair
  ? mintCanonicalId(type, state.scope, outputIndex)
  : state.repairMissingTerminalIds
    ? rawId
    : null;
```

The request-local mapping should be applied consistently to:

- `response.output_item.added.item.id`
- lifecycle event `item_id` fields
- `response.output_item.done.item.id`
- `response.completed.response.output[].id`

Function-call IDs and `call_id` should remain untouched. Because the rewrite is client-facing only, raw upstream snapshots can remain available for provider replay.

### Redacted configuration

```json
{
  "providers": {
    "provider": {
      "baseUrl": "https://redacted.example",
      "responsesPath": "/responses",
      "authMode": "key",
      "models": ["deepseek-v4-flash", "deepseek-v4-pro"],
      "statelessResponses": true
    }
  },
  "websockets": false
}
```

### Checks

mouzhi:
## Summary
- Add opt-in `responsesItemIdRepair.rewriteNonCanonicalIds` for openai-responses passthrough providers.
- Rewrite DeepSeek-style UUID response/item ids to Codex-friendly `resp_` / `msg_` / `rs_` ids, fold `reasoning_text` into `encrypted_content` (while keeping plaintext `content` for DeepSeek replay), and ensure a terminal `[DONE]` trailer.
- Keep SSE rewrite consumers alive when many events are dropped, including the Windows eager-relay path.
- Make raw-id aliases type-scoped (`message` vs `reasoning`) and charge every newly retained alias through `TranslatorBudget`.
- Enable the repair by default on the built-in DeepSeek registry preset so the native `deepseek/deepseek-v4-flash` Responses route no longer requires a hand-edited provider config for the #938 UUID stall.

Fixes #938

## Scope note
This PR is primarily the provider-local / built-in SSE UUID repair for #938.

It also turns the same repair on for the built-in DeepSeek preset, which covers the UUID-stall half of the #875/#946 residual on SSE. It does **not** claim a complete fix for the separate “no function_call_output continuation request is sent after tools” residual tracked in #875: that path still needs end-to-end tool-loop evidence on Codex App WebSocket after this lands.

## Why
DeepSeek V4 Flash/Pro Responses API can return UUID item ids and raw `reasoning_text` streams. Codex App/CLI may stay on Thinking/Working even when OpenCodex logs HTTP 200 and reported usage. This keeps OpenCodex passthrough intact while normalizing the client-facing SSE for Codex.

## Usage
```json
{
  "providers": {
    "deepseeknew": {
      "baseUrl": "https://api.deepseek.com",
      "authMode": "key",
      "models": ["deepseek-v4-flash"],
      "responsesItemIdRepair": {
        "rewriteNonCanonicalIds": true,
        "repairMissingTerminalIds": true
      }
    }
  }
}
```

Built-in DeepSeek now seeds the same repair automatically via registry fill-only defaults.

## Test plan
- [x] `bun test tests/responses-item-id-repair.test.ts`
- [x] `bun test tests/responses-item-id-repair-deepseek.test.ts`
- [x] `bun test tests/sse-payload-rewrite.test.ts`
- [x] Regression: shared placeholder across message/reasoning stays type-scoped
- [x] Regression: newly retained raw aliases after an already-mapped output_index charge TranslatorBudget
- [x] Local Codex CLI/Desktop validation with DeepSeek Responses before/after the initial fix

## Summary by CodeRabbit

- **New Features**
  - Added an optional setting to rewrite non-canonical response, message, and reasoning IDs into consistent canonical IDs.
  - Normalized reasoning streams, filtered intermediate events, and ensured completed responses receive a consistent `[DONE]` termination marker.
  - Added configurable end-of-stream trailers for streamed responses.
  - Enabled ID rewriting and terminal repair by default for DeepSeek.

- **Bug Fixes**
  - Preserved unchanged SSE event formatting.
  - Correctly removes dropped events without disrupting subsequent streamed output.

### DIFF
--- a/docs-site/src/content/docs/ja/reference/configuration/providers.md
+++ b/docs-site/src/content/docs/ja/reference/configuration/providers.md
@@ -84,7 +84,7 @@ namespace 付き combo alias はその namespace prefix に selector を再利�
 | `noTopPModels?` | `string[]` |発信者指定の`top_p`を拒否するモデル。 |
 | `noPenaltyModels?` | `string[]` |存在/周波数ペナルティを拒否するモデル。 |
 | `parallelToolCalls?` | `boolean` |並列ツール呼び出しを切り替えます。 OpenAI Chat はデフォルトでオンになっています。非チャット アダプターは明示的な `true` でのみアドバタイズします。 |
-| `responsesItemIdRepair?` | `{ message?: string[]; reasoning?: string[]; repairMissingTerminalIds?: boolean }` |正確なプレースホルダー ID および欠落している端末 ID に対するダウンストリーム SSE 修復はデフォルトで無効になっています。関数呼び出し ID は決して書き換えられません。 |
+| `responsesItemIdRepair?` | `{ message?: string[]; reasoning?: string[]; repairMissingTerminalIds?: boolean; rewriteNonCanonicalIds?: boolean }` |正確なプレースホルダー ID、欠落している端末 ID、および非標準の UUID 形式 response/item ID（`rewriteNonCanonicalIds`）に対するダウンストリーム SSE 修復はデフォルトで無効です。関数呼び出し ID は決して書き換えられません。 |
 | `autoToolChoiceOnlyModels?` | `string[]` | `tool_choice` が `auto` または `none` のみを受け入れるモデル。強制的な選択は格下げされます。 |
 | `preserveReasoningContentModels?` | `string[]` |チャット履歴に以前のアシスタント `reasoning_content` が必要なモデル。 |
 | `thinkingToggleModels?` | `string[]` |エフォート ラダーではなく `thinking.enabled` を使用してモデルをチャットします。 |
@@ -197,7 +197,8 @@ Anthropic アカウント ポリシーのリスクを理解していない限り
       "responsesItemIdRepair": {
         "reasoning": ["rs_0"],
         "message": ["msg_0"],
-        "repairMissingTerminalIds": true
+        "repairMissingTerminalIds": true,
+        "rewriteNonCanonicalIds": true
       }
     }
   }
--- a/docs-site/src/content/docs/ja/reference/configuration/providers.md
+++ b/docs-site/src/content/docs/ja/reference/configuration/providers.md
@@ -205,7 +205,7 @@ Anthropic アカウント ポリシーのリスクを理解していない限り
 }
 ```
 
-プレースホルダー リストは完全に一致します。通常/ステートフル応答プロバイダーのフィールドを未設定のままにして、パススルーがバイトごとに同一になるようにします。
+プレースホルダー リストは完全に一致します。DeepSeek Responses のように UUID 形式の response/item ID と生の `reasoning_text` ストリームを返すゲートウェイでは `rewriteNonCanonicalIds` を有効にします。OpenCodex はそれらの ID を Codex 向けプレフィックスへ書き換え、reasoning を `encrypted_content` に正規化しつつ replay 用の平文 `content` を保持し、終端の `[DONE]` を補完します。通常/ステートフルな Responses プロバイダーではこのフィールドを未設定のままにして、パススルーがバイト単位で同一になるようにします。
 
 ## Cursor プロバイダー (`adapter: "cursor"`)
 
--- a/docs-site/src/content/docs/ko/reference/configuration/providers.md
+++ b/docs-site/src/content/docs/ko/reference/configuration/providers.md
@@ -208,7 +208,7 @@ Anthropic 계정 정책 위험을 이해하지 못한다면 이 기능은 꺼두
 }
 ```
 
-자리표시자 목록은 정확히 일치해야 합니다. 일반적인 상태 유지형 Responses 공급자에서는 필드를 설정하지 마십시오. 그래야 passthrough가 바이트 단위로 동일하게 유지됩니다.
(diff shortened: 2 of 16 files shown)


Ingwannu (changes requested):
Thanks — the opt-in SSE normalization is valuable and is the right direction for #938, but I found two correctness blockers on the current head.

1. `rawToCanonical` is global across message/reasoning item types and is consulted before the event type. If a provider reuses the same configured placeholder ID for both types, a reasoning lifecycle event can be rewritten to the `msg_...` ID. Please make the alias mapping type-scoped and add a regression where message and reasoning share one raw placeholder while their lifecycle IDs remain distinct.

2. When an output index already has a mapped ID, each new raw alias is retained in `rawToCanonical` without charging `TranslatorBudget`. A malformed stream can therefore grow the alias map outside the request budget. Charge every newly retained alias and add a small-budget regression proving the map is bounded.

There is also an important scope distinction. This PR can fix #938 as an explicit provider-local SSE compatibility option, but it does not yet fix the built-in DeepSeek/Codex App residual tracked in #875/#946: the built-in DeepSeek preset does not enable the option, and the WebSocket path forces bounded upstream JSON and bypasses this SSE rewrite. Either keep the PR and its closing claim explicitly scoped to #938, or add the built-in preset plus JSON/WebSocket normalization and an end-to-end function-call continuation test.

Please also sync with current `dev` and rerun the required full CI after the corrections. The current head is 21 commits behind and only target/label checks have run; I am not approving external code execution for this head while the blockers remain.

lidge-jun:
Superseded by #1027 (head `8403fe695e7b`), which is **open and not yet merged** at the time of this comment.

**Adopted from this PR**

- The core diagnosis and fix for #938: DeepSeek emits UUID item IDs where Codex expects canonical prefixed IDs, so message and reasoning item IDs are rewritten on the client-facing path. The ID-prefix idea is yours and #1027 implements it.
- Keeping the rewrite client-facing only, so raw snapshots stay untouched.

**Deliberately not matched**

#1027 is a narrow reimplementation rather than a merge of this branch, because the bundle here reaches well past #938's acceptance boundary:

- response-ID rewrites and continuation aliasing,
- dropping `response.in_progress` and raw reasoning events,
- synthesized envelopes,
- logprobs stripping,
- relay and eager-relay surgery,
- `[DONE]` synthesis.

Each of those changes behavior for streams that are not affected by #938, and several interact with work landing in the same series. The plan of record is ID-only normalization first; if the reporter still stalls afterwards, the reasoning-event conversion and response-ID aliasing get evaluated separately on their own evidence — at which point this branch's approach is the obvious starting point.

Function-call IDs are preserved unchanged in #1027, matching your handling.

**Status disclosure**

#1027 is rebased onto current `dev` and revalidated before merge; that revalidation is still pending as of this comment. If it fails to land, or if the merged version no longer preserves the adopted behavior, this PR is reopened.

Full equivalence reasoning: `devlog/_plan/260805_bug_stack_campaign/060_issue938_uuid_item_ids.md`.

PR #941, opened 2026-08-03
fix(antigravity): keep the signature from a standalone thought part

### LINKED ISSUE #897: google-antigravity: a reasoning signature carried on a standalone thought part is dropped by the replay cache
### Client or integration

Codex CLI

`google-antigravity` (Antigravity / Code Assist), reasoning-signature replay.

### OpenCodex version

2.10.0

### Endpoint or capability

`observeAntigravityReplay()` / `applyAntigravityReplay()` in `src/adapters/google-antigravity-replay.ts` — the reasoning-signature replay cache used across sequential tool loops.

### Current behaviour

`observeAntigravityReplay()` only records a signature when it appears on the **same part** as the `functionCall`:

```ts
for (const raw of parts) {
  if (!raw || typeof raw !== "object") continue;
  const part = raw as Record<string, unknown>;
  const sig = extractSignature(part);
  if (!sig) continue;
  const fc = part.functionCall as { name?: unknown; args?: unknown } | undefined;
  const ck = fc ? functionCallKey(fc.name, fc.args) : undefined;
  if (!ck) continue; // only function-call signatures are replayable by identity
  ...
}
```

This provider also emits the signature on a **standalone `thought` part that precedes** the `functionCall` it belongs to:

```json
[
  { "thought": true, "thought_signature": "<sig>" },
  { "functionCall": { "name": "get_x", "args": { "a": 1 } } }
]
```

For that shape, `extractSignature(part)` returns the signature on the first part, the `continue` on the missing `functionCall` throws it away, and the second part has no signature of its own — so nothing is cached. On the next turn `applyAntigravityReplay()` finds no entry and the `functionCall` goes back upstream with no `thoughtSignature`.

The interesting part is that the read side already handles both spellings (`part.thoughtSignature ?? part.thought_signature`, plus the nested `extra_content.google.thought_signature`), so the intent to accept several shapes is clearly there — it is only the part-to-call association that assumes they arrive together.

### Expected behaviour

A signature carried on a `thought` part should be associated with the next `functionCall` in the same `parts` array, so it can be replayed on the following turn like any other. Signatures already on the call's own part must keep taking precedence.

Against the module directly:

```ts
observeAntigravityReplay(MODEL, SESSION, [
  { thought: true, thought_signature: SIG },
  { functionCall: { name: "get_x", args: { a: 1 } } },
]);

const contents = [{ role: "model", parts: [{ functionCall: { name: "get_x", args: { a: 1 } } }] }];
applyAntigravityReplay(MODEL, SESSION, contents);

// expected: SIG    actual: undefined
(contents[0].parts[0] as { thoughtSignature?: string }).thoughtSignature;
```

### Actual response or error

```
Expected: "<sig>"
Received: undefined
```

At the provider boundary the symptom is subtler than a hard failure: sequential tool loops simply stop replaying reasoning signatures for the turns where the provider used the separate-`thought` shape.

### Upstream documentation

Gemini / Code Assist `Part` allows `thought` and `thoughtSignature` as independent fields, so a thought part carrying only the signature is a valid shape rather than a provider quirk: https://ai.google.dev/api/caching#Part

Carry a pending signature forward within the same `parts` scan. This keeps the existing precedence and does not change behaviour when the signature is already on the call's own part:

```ts
let pendingSignature: string | undefined;
for (const raw of parts) {
  if (!raw || typeof raw !== "object") continue;
  const part = raw as Record<string, unknown>;
  const own = extractSignature(part);
  const fc = part.functionCall as { name?: unknown; args?: unknown } | undefined;
  const ck = fc ? functionCallKey(fc.name, fc.args) : undefined;
  if (!ck) {
    if (own) pendingSignature = own;
    continue;
  }
  const sig = own ?? pendingSignature;
  pendingSignature = undefined;
  if (!sig) continue;
  // ... unchanged: byte accounting, eviction, insert
}
```

I have been running this locally against the `2.10.0` replay module — the byte accounting, LRU eviction and session limits added in this release are untouched by it, and the existing suite in `tests/google-antigravity-replay.test.ts` still passes (21 tests) with the case above added.

Happy to send it as a PR with that test if you want it.

Only the association is missing — the read path, the both-spellings handling and the cache mechanics all work correctly.

### Checks

DevMello:
## Summary

Fixes #897. When Antigravity returns the reasoning signature on its own thought part instead of on the functionCall part, the replay cache dropped it, so the next turn went out unsigned and upstream rejected it. The cache now pairs a standalone thought signature with the functionCall that follows it in the same parts array. A signature on the call part itself still wins, an unpaired one is dropped at the end of the array, and byte accounting is untouched. Pairing stays within one observed array; a thought and its call split across streamed chunks is out of scope since the reported captures are same-array.

## Verification

- Four new tests in `tests/google-antigravity-replay.test.ts`: the standalone shape replays, the call's own signature wins, no backward or cross-observe pairing, and non-thought parts do not attach.
- Unit repro: `[{thought, thoughtSignature}, {functionCall}]` lost its signature before the fix and round-trips after.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved replay handling for function calls preceded by standalone thought signatures.
  * Ensured the correct signature is used when multiple signatures are present.
  * Prevented signatures from being incorrectly reused across unrelated events or attached to unsupported content.
  * Preserved existing validation and cache-size protections for valid replayed calls.

* **Tests**
  * Added coverage for signature pairing, precedence, isolation, and invalid attachment scenarios.

### DIFF
--- a/src/adapters/google-antigravity-replay.ts
+++ b/src/adapters/google-antigravity-replay.ts
@@ -365,6 +365,8 @@ export function antigravityUsesReplayCache(model: string): boolean {
  * Observe a parsed CCA chunk's `candidates[0].content.parts` and record thought signatures keyed by
  * the functionCall identity (name + args). Accumulates across the whole session so a sequential
  * multi-step tool loop keeps EVERY prior call's signature, not just the latest part-index slot.
+ * A signature on a standalone thought part is paired with the next functionCall in the same
+ * array (#897); a call's own signature takes precedence and an unpaired one is dropped.
  * `parts` is the already-unwrapped `response.candidates[0].content.parts`.
  */
 export function observeAntigravityReplay(model: string, sessionId: string, parts: unknown[]): void {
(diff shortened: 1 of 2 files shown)


Ingwannu (approved):
Confirmed as a real provider-compatibility defect. The patch keeps the repair narrow: same-array forward pairing only, call-local signature precedence, no cross-observation carry, no non-thought attachment, and no change to the existing replay cache limits or identity key.

Verification:
- applied cleanly on current `dev` (`3cd75ab0`)
- `bun test tests/google-antigravity-replay.test.ts`: 39 passed
- `bun run typecheck`: passed
- repository Cross-platform CI and React Doctor: passed

No blocking findings.

PR #948, opened 2026-08-03
fix(kiro): round-trip the redactedContent reasoning blob

mushikingh:
## Problem

Kiro never returns plaintext reasoning for its Sol-family models. `reasoningContentEvent` carries a KMS-encrypted `redactedContent` blob, and `gpt-5.6-sol`'s `additionalModelRequestFieldsSchema` (from `ListAvailableModels`) accepts **only** `reasoning.effort` — there is no display/summary opt-in:

```json
{"modelId":"gpt-5.6-sol",
 "additionalModelRequestFieldsSchema":{"type":"object","properties":{
   "reasoning":{"type":"object","properties":{"effort":{"type":"string",
     "enum":["none","low","medium","high","xhigh","max"],"default":"high"}}}},
   "additionalProperties":false},
 "tokenLimits":{"maxInputTokens":272000,"maxOutputTokens":128000}}
```

Kiro's own CLI replays that blob on the matching assistant turn to preserve reasoning across turns:

```json
{"assistantResponseMessage":{
  "messageId":"…","content":"…",
  "reasoningContent":{"redactedContent":"LktUUn5+…"}}}
```

`kiro-events.ts` read only `reasoningContentEvent.text`, which is **absent** on this wire, so the blob was dropped and never replayed. Every turn restarted without the previous turn's reasoning.

In a long agentic session this degrades badly. The session that prompted this fix ran 123 tool calls at `xhigh` effort and ended with the model emitting a 322 KB / 162,779-line assistant message of degenerate self-talk (`final.` ×730, `no.` ×390, `done.` ×311, plus lines like "I'm going in circles again") until it was interrupted.

## Also fixed

`contextUsageEvent` was not in `KNOWN_EVENT_TYPES`, so it was silently ignored. Kiro reports context pressure there; `metadataEvent` carries only `stopReason`. The adapter looked for `contextUsagePercentage` inside `metadataEvent`, where it never appears, so authoritative context usage never arrived.

## Approach

- Parse `redactedContent`, and add the missing `contextUsageEvent`.
- Carry the blob through the existing `ocxr1:` envelope as `krc`, on an envelope-only reasoning item (`summary: []`, no text deltas) — it round-trips while staying invisible in the app, the same contract the hidden-thinking path already uses. Prefix-based native scrubbing already strips it for non-Kiro backends.
- **Pair it backwards.** Kiro emits `reasoningContentEvent` at the END of a turn, after content *and* tool calls, so a `krc`-only item belongs to the assistant turn that already closed. Folding it forward like ordinary reasoning would attach turn N's blob to turn N+1. With no assistant turn to own it, the blob is dropped rather than mis-paired.
- The blob lives on `OcxAssistantMessage.kiroRedactedReasoning`, not on a thinking content part, so no other adapter replays provider-private state if the conversation switches providers.

Observed event order per assistant turn (tool-calling turn, kiro-cli 2.14.1):

```
toolUseEvent ×67 → reasoningContentEvent → metadataEvent → contextUsageEvent → meteringEvent
```

## Verification

Request/response captures from the official `kiro-cli` at **2.14.1 and 2.16.0** (via `KIRO_LOG_LEVEL=trace KIRO_LOG_STDOUT=1`), covering a trivial turn, a reasoning-heavy turn, a tool-calling turn, and a resumed multi-turn conversation. Both versions produce identical wire shapes.

- `bun run typecheck` — clean
- `bun run test` — 7625 pass, 4 fail. All 4 pre-existing on clean `upstream/dev`, in files this PR does not touch (`claude-desktop-cli`, `cli-restore-back`, `codex-app-server-processes`); confirmed by running the same files on an unmodified worktree — identical failures.
- New regression tests: `kiro-stream` (redactedContent capture, text+blob together, `contextUsageEvent`), `kiro-adapter` (history replay, and omission when no blob), `anthropic-thinking-signature` (backwards pairing, and orphan blob dropped).

## Not included

`meteringEvent` is still ignored. Kiro bills in **credits, not tokens** — there is no `tokenUsage` on this wire at all, which is why Kiro usage stays `estimated`. Surfacing credits needs a cost field in `OcxUsage` plus usage-log/GUI plumbing, so it belongs in its own PR.

## Summary by CodeRabbit

* **New Features**
  * Preserved and replayed Kiro’s redacted reasoning across conversations, streaming, and batch responses.
  * Added validated context-usage event handling.
  * Retained encrypted reasoning content without displaying it as visible text.

* **Bug Fixes**
  * Prevented orphaned redacted reasoning from attaching to unrelated assistant messages.

* **Documentation**
  * Documented Kiro reasoning round-tripping, context usage, stop reasons, and credit-based metering.

### DIFF
--- a/src/adapters/kiro-events.ts
+++ b/src/adapters/kiro-events.ts
@@ -18,8 +18,9 @@ const KNOWN_EVENT_TYPES = new Set([
   "toolUseEvent",
   "messageMetadataEvent",
   "metadataEvent",
-  // Kiro reports context pressure in its OWN event type, not inside metadataEvent (verified
-  // against kiro-cli 2.14.1 and 2.16.0: metadataEvent carries only `stopReason`).
+  // Authoritative context pressure. Every capture (kiro-cli 2.14.1 and 2.16.0) put the percentage
+  // HERE and left `metadataEvent` carrying only `stopReason`; metadataEvent's own
+  // contextUsagePercentage stays supported as a fallback rather than being dropped.
   "contextUsageEvent",
   "invalidStateEvent",
   "error",
--- a/src/adapters/kiro-events.ts
+++ b/src/adapters/kiro-events.ts
@@ -3,7 +3,8 @@ import { kiroTruncationReason } from "./kiro-truncation";
 
 export type ParsedKiroEvent =
   | { type: "content"; data?: string; modelId?: string }
-  | { type: "reasoning"; data?: string }
+  | { type: "reasoning"; data?: string; redactedContent?: string }
+  | { type: "context_usage"; contextUsagePercentage: number }
   | { type: "tool"; name?: string; toolUseId?: string; input?: string; stop?: boolean }
   | { type: "truncation"; data: string }
   | { type: "metadata"; usage?: OcxUsage; contextUsagePercentage?: number; stopReason?: string }
@@ -17,6 +18,9 @@ const KNOWN_EVENT_TYPES = new Set([
   "toolUseEvent",
   "messageMetadataEvent",
   "metadataEvent",
+  // Kiro reports context pressure in its OWN event type, not inside metadataEvent (verified
+  // against kiro-cli 2.14.1 and 2.16.0: metadataEvent carries only `stopReason`).
+  "contextUsageEvent",
   "invalidStateEvent",
   "error",
 ]);
@@ -114,10 +118,15 @@ export function parseKiroEvent(eventType: string, payload: Uint8Array): ParsedKi
           : {}),
       };
     case "reasoningContentEvent":
+      // `text` is plaintext reasoning; `redactedContent` is the encrypted blob the model actually
+      // returns for Sol-family models. Both may be absent on a bare event.
       return {
         type: "reasoning",
         ...(optionalString(eventType, parsed, "text") !== undefined
           ? { data: optionalString(eventType, parsed, "text") }
+          : {}),
+        ...(optionalString(eventType, parsed, "redactedContent") !== undefined
+          ? { redactedContent: optionalString(eventType, parsed, "redactedContent") }
           : {}),
       };
     case "toolUseEvent":
@@ -160,6 +169,13 @@ export function parseKiroEvent(eventType: string, payload: Uint8Array): ParsedKi
         ...(typeof contextUsagePercentage === "number" ? { contextUsagePercentage } : {}),
         ...(stopReason !== undefined ? { stopReason } : {}),
       };
+    }
+    case "contextUsageEvent": {
+      const contextUsagePercentage = parsed.contextUsagePercentage;
+      if (typeof contextUsagePercentage !== "number" || !Number.isFinite(contextUsagePercentage)) {
+        return malformed(eventType, "contextUsagePercentage must be a finite number");
+      }
+      return { type: "context_usage", contextUsagePercentage };
     }
     case "invalidStateEvent":
       return { type: "invalid_state", message: optionalString(eventType, parsed, "message") };
--- a/src/adapters/kiro-events.ts
+++ b/src/adapters/kiro-events.ts
@@ -118,8 +118,9 @@ export function parseKiroEvent(eventType: string, payload: Uint8Array): ParsedKi
           : {}),
       };
     case "reasoningContentEvent":
-      // `text` is plaintext reasoning; `redactedContent` is the encrypted blob the model actually
-      // returns for Sol-family models. Both may be absent on a bare event.
+      // `text` is plaintext reasoning; `redactedContent` is the encrypted blob the GPT-5.6 family
+      // (sol/terra/luna) actually returns — they never send `text`. Keyed off the wire field, not
+      // the model id. Both may be absent on a bare event.
       return {
         type: "reasoning",
         ...(optionalString(eventType, parsed, "text") !== undefined
--- a/src/adapters/kiro.ts
+++ b/src/adapters/kiro.ts
@@ -99,7 +99,11 @@ interface KiroUserInputMessage {
 }
 interface KiroHistoryEntry {
   userInputMessage?: KiroUserInputMessage;
-  assistantResponseMessage?: { content: string; toolUses?: KiroToolUse[] };
+  assistantResponseMessage?: {
+    content: string;
+    toolUses?: KiroToolUse[];
+    reasoningContent?: { redactedContent: string };
+  };
 }
 
 function kiroToolWireNames(tools: readonly unknown[]): string[] {
@@ -326,7 +330,7 @@ function validateKiroCapabilities(parsed: OcxParsedRequest): void {
 
 type KiroTurn =
   | { kind: "user"; content: string; images: KiroImage[]; toolResults: KiroToolResult[] }
-  | { kind: "assistant"; content: string; toolUses: KiroToolUse[] };
+  | { kind: "assistant"; content: string; toolUses: KiroToolUse[]; redactedReasoning?: string };
 
 function appendTurnText(target: string, next: string): string {
   if (!next) return target;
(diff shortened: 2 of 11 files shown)


mushikingh:
Both findings validated and fixed in bd13d488.

**1. Ordering / `output_index` collision — confirmed, and worse than described.**

`outputIndex` only advances when an item *closes*, so an envelope emitted while `currentMsg` was open reused that index. And because the parser pairs a `krc`-only item backwards, the envelope landing before the assistant message meant it was dropped as orphaned — the round-trip was a no-op in the streaming path. My original tests missed it: `kiro-stream` asserts adapter events (pre-bridge) and the parser test hand-built the item order, so nothing exercised bridge → parse.

Rather than closing the open items on arrival, both paths now **stash the blob and flush it after every open item has closed** — after the closes in the streaming `done` case, and after the trailing flushes in the batch path. Closing `currentMsg` early would have forced phase `commentary`, turning every Kiro final answer into commentary; deferring keeps phase inference intact and yields the `[message, reasoning]` order the parser expects.

**2. Batch budget imbalance — confirmed.**

`pushOutput` released `bytesOf(encrypted)` that was never retained. Fixed at the source rather than with a reserve/commit round-trip: the blob is now charged when stashed, and `pushOutput` releases that retained allocation.

**Coverage.** New `tests/kiro-reasoning-roundtrip.test.ts` bridges adapter events and re-parses the emitted items the way Codex replays history, asserting distinct output indexes, item order, and that `kiroRedactedReasoning` reaches the assistant turn — streaming, batch, tool-call-only turns, and budget drift. Three of its five cases fail against cbb5d21e, so they are not vacuous.

**Scope note.** Probing the other models: `gpt-5.6-terra` and `gpt-5.6-luna` also return `redactedContent` and never `text`, with the same `reasoning.effort`-only schema — so the whole GPT-5.6 family was affected, not just sol. Claude 4.6+/5 advertise a different contract (`thinking.display`, `output_config.effort`); I did not get a reasoning event from `claude-sonnet-4.6` at low effort, so I make no claim there. Handling keys off the wire field, not the model id.

Gates: `typecheck` clean, `privacy:scan` passed, full suite 7630 pass / 4 fail — the same 4 pre-existing failures on clean `upstream/dev`, in files this PR does not touch.

mushikingh:
Both findings valid, fixed in 8610630e.

**1. Budget regression was vacuous — worse than flagged.**

Beyond the weak `currentBytes >= 0` assertion, the test passed the budget in the **wrong argument position**, so `buildResponseJSON` constructed its own internal budget and the snapshot under test was never written to at all (`highWaterBytes: 0`). It asserted nothing.

Now uses a 4 KB blob and asserts `currentBytes` equals **exactly** the finalized output items' bytes:

- a still-retained raw blob → ~4 KB of excess
- releasing bytes never charged (the original defect) → a shortfall

plus `highWaterBytes >= finalized + blobLength`, proving the blob was charged while held rather than never accounted for. Measured behaviour with the fix: `highWater 9650 → current 5650`, exactly the finalized items.

Four of the file's five cases now fail against cbb5d21e, up from three.

**2. `metadataEvent` doc was wrong.**

Correct — `kiro-events.ts` still accepts a finite `contextUsagePercentage` (and a `tokenUsage` block) from `metadataEvent`, so documenting it as carrying only `stopReason` would make a legitimately parsed value look impossible. Reworded in both `structure/` and the source comment: `contextUsageEvent` is authoritative and is where every capture put the percentage, with `metadataEvent` retained as a supported fallback; both feed the same field and any positive value overwrites an earlier one.

Gates: `typecheck` clean, `privacy:scan` passed, full suite 7629 pass / 5 fail. Four are the same pre-existing failures as clean `upstream/dev`; the fifth (`storage-policy-job-responsive`) is a load-sensitive timing test that passes in isolation on repeat runs and is unrelated to this change.

mushikingh:
Both findings valid, fixed in 3a2f0574.

**1. Authoritative context usage could be clobbered by the fallback.**

Confirmed: `kiro.ts` assigned `contextUsagePercentage` from both `context_usage` and `metadata` with last-write-wins, so arrival order alone decided the winner. Precedence is now by **source** — once a `contextUsageEvent` value lands, later `metadataEvent` percentages are ignored; the fallback still works on its own when no authoritative frame arrives.

On the captured wire the order is `metadataEvent → contextUsageEvent`, so the authoritative value happened to win by luck. This removes the dependence on that.

Covered in both orders. The test first asserts that the 80% and 10% checkpoints are actually distinguishable and that the fallback alone still produces a value, so the ordering assertions cannot pass vacuously; it fails against the unguarded assignment.

I also corrected the `structure/` sentence I had added last round — it documented the old last-write-wins behaviour ("any positive value overwrites an earlier one"), which this change makes wrong.

**2. Budget disposal.**

Valid — `createTranslatorBudget` registers in module-global aggregate accounting (`liveBudgets`, `aggregateCurrentBytes`), and `dispose()` in a `finally` is the convention already used in `translator-budget.test.ts` and `cline-pass-provider.test.ts`. Without it an assertion failure would have leaked ~5.6 KB of retained bytes into every later test in the process. Wrapped.

Gates: `typecheck` clean, `privacy:scan` passed, full suite **7631 pass / 4 fail** — the same four pre-existing failures as clean `upstream/dev`, in files this PR does not touch. (The `storage-policy-job-responsive` flake noted last round did not recur.)

lidge-jun:
Carried onto the review stack as **#953** (stack 3/3), unmodified.

Your commits were taken with `git cherry-pick -x`, so they keep your authorship — `git log --format='%an'` on the stack branch shows you, not me. No content was changed; the diff on the stack is byte-identical to what you wrote here, and it applied to `dev` with no conflict resolution.

Verified on the stack: `bun x tsc --noEmit` exit 0, and the full suite at 7691 pass / 8 skip / 0 fail across 507 files.

**This PR stays open until #953 lands.** If a maintainer prefers to take yours directly instead, that path is unaffected — the stack commits get dropped and this one merges. Once #953 merges I'll close this as carried, with the credit already in the commit history rather than in a comment.

Stack: **#951** (plan, base `dev`) → **#952** (#908 long-context pricing) → **#953** (this carry). Review bottom-up.

Thanks for the fix.

mushikingh:
Valid, fixed in c3d2998c.

Confirmed: `contextUsageEvent` claimed authority only when its percentage was `> 0`, so precedence depended on the **value** rather than the source — exactly the coupling the previous round was meant to remove. A genuine 0% reading left `contextUsageIsAuthoritative` false and a trailing `metadataEvent` could override it.

Now accepts `>= 0` and claims authority. Negatives stay rejected as malformed.

I checked the downstream before widening the guard: `contextUsageTotalFloor()` already ends in `floor > 0 ? floor : undefined`, so a 0% reading still yields no checkpoint rather than a bogus zero-token one. The only other consumer is the debug diagnostic, which becomes marginally more accurate. So this is safe to accept.

I did **not** widen the `metadataEvent` guard to match. That `> 0` is pre-existing fallback behaviour outside this PR's scope, and changing it would alter semantics for wires that do populate it.

Test extended with `contextUsageEvent(0)` followed by a nonzero `metadataEvent`, asserting the result matches the no-percentage baseline and specifically differs from the fallback-only reading — it fails against the `> 0` guard.

Gates: `typecheck` clean, `privacy:scan` passed, full suite **7631 pass / 4 fail** — the same four pre-existing failures as clean `upstream/dev`, in files this PR does not touch.

lidge-jun:
Closing this as **carried into #953**, not as rejected. Your fix is shipping — this is bookkeeping so the same change isn't sitting in the review queue twice.

Your commit(s) are on the stack branch as `a4c78b270, 6cda4a160, 9664abdd1`, cherry-picked with `git cherry-pick -x`. Authorship is preserved, so `git log --format='%an'` shows you rather than me, and I verified the carried diff is byte-identical to yours rather than assuming it:

```console
$ git show <yours> | git patch-id --stable
$ git show <carried> | git patch-id --stable
IDENTICAL
```

No content was changed and no conflict resolution was needed.

One thing I want to be straight about: my earlier comment said I'd close this *after* #953 merged, and I'm closing it before. #953 is open and reviewable now, so closing early is my call as maintainer to keep the queue honest — not a signal that the review finished. **If the stack doesn't land, reopening this is one click**, and the branch is untouched.

Verified on the stack: `bun x tsc --noEmit` exit 0, full suite 7729 pass / 8 skip / 0 fail across 508 files.

Stack: **#951** (plan, base `dev`) → **#952** → **#953** (this carry) → **#954** → **#955**. Review bottom-up.

Thanks @mushikingh — this was a clean fix and it's in.

PR #953, opened 2026-08-03
stack 3/7: carry six contributor bug fixes with authorship intact

lidge-jun:
## Stack

**3/3 — carried contributor bug fixes**

Base: `codex/908-long-context-pricing` (#952)

## Summary

Carries six contributor bug fixes onto the stack, unmodified, with original authorship preserved via `git cherry-pick -x`. Every commit keeps its author; `git log --format='%an'` shows the contributor, not me.

| Source PR | Author | Carried commits | Fix |
|---|---|---|---|
| #939 | @DevMello | 1 | keep a marker for `tool_result` document blocks |
| #943 | @DevMello | 1 | fail closed on `MALFORMED_FUNCTION_CALL` without a call part |
| #944 | @DevMello | 1 | separate reasoning summary parts in streamed thinking |
| #945 | @DevMello | 1 | report upstream usage for native compact turns |
| #942 | @L14nY1Wang | 1 | close passthrough streams at terminal events |
| #948 | @mushikingh | 3 | round-trip the Kiro `redactedContent` reasoning blob |

Eight commits, +604/-70 across 20 files. No content changes were made to any of them — this is a carry, not a rework. Each applied to `dev` cleanly with no conflict resolution, so what is reviewed here is byte-identical to what the authors wrote.

## Why carry rather than merge in place

These are small, focused, independently-correct fixes that were sitting unmerged behind a long review queue. Stacking them puts them on one branch with one CI run and one review pass, in the same bottom-up order as the rest of the stack.

**The source PRs stay open until this lands.** If a maintainer prefers to take any of them directly instead, that path is unaffected — drop the corresponding commits here and merge the original. Once this merges, each source PR can be closed as carried, with credit already recorded in the commit history rather than in a comment.

## Not carried, and why

| PR | Author | Reason |
|---|---|---|
| #947 | @WZBbiao | conflicts with #942's terminal-event handling — both touch the same relay path, and resolving that is the author's call, not a carry |
| #933 | @IMHinnG | draft, `enforce-target` failing |
| #928, #922, #940, #935 | | `CHANGES_REQUESTED` — carrying a PR past requested changes would route around the review |

## Verification

- `bun x tsc --noEmit` — exit 0
- `bun test` on all 11 touched test files — 331 pass, 0 fail
- `bun run test` — 7691 pass, 8 skip, 0 fail, 507 files
- `bun run privacy:scan` — passed

[...]

PR #952, opened 2026-08-03
stack 2/7: price long-context requests at the published long rate (#908)

lidge-jun:
## Stack

**2/3 — long-context pricing tiers**

Base: `codex/bug-stack-plan` (#951)
Next: carried contributor bug fixes (#952)

## Summary

Several vendors reprice the **entire request** once the prompt crosses a token threshold. `Cost4` is flat and `resolveMatchedPrice()` never saw a token count, so there was nowhere to express "this rate depends on how big the prompt is" — every request billed at the short rate, including the long ones, which are the expensive ones.

- new `ContextTier` registry with exact provider+model rules, each carrying its source URL and `verifiedAt`
- tier selection between base-price resolution and `calculateCost()`; `resolveMatchedPrice()` stays token-independent so its provider/model memoization is untouched
- `contextTier` surfaced on `AttemptCostEstimate` and `CostEstimate`, propagated to combo results
- base prices for the three `-pro` virtual aliases

| Model | Threshold | Operator | Effect |
|---|---|---|---|
| `gpt-5.6-sol` / `-terra` / `-luna` (+ `-pro`) | 272,000 | `>` | 2× input, 2× cached, 2× cache write, 1.5× output |
| `grok-4.5` | 200,000 | `>=` | 2× uniform |
| `MiniMax-M3` | 512,000 | `>` | 2× uniform |

Verified 2026-08-03 against the published tables — OpenAI's `>272K` is exclusive, xAI's `≥ 200k` is inclusive.

## Three things worth reviewing closely

**The threshold reads raw `usage.inputTokens`, not normalized input.** `normalizeCostTokens()` subtracts cache read/write to produce billable input, so a 280k prompt with a 200k cache read has 80k billable input and still crosses OpenAI's boundary. Deciding after normalization would have silently under-billed exactly the cache-heavy long requests. Covered by L3.

**Long context and Fast are mutually exclusive, not composable.** An earlier draft multiplied both. OpenAI's Fast guide states plainly that "Long context, fine-tuned models, and embeddings are not supported", so that product cannot exist. Exclusivity keys on the **response-confirmed** tier: a >272k request merely *tagged* `priority` was necessarily downgraded and must still bill long — suppressing the tier there would under-bill the downgraded request. That required passing tier provenance rather than the collapsed `effectiveServiceTier()` scalar to all four estimator call sites. Covered by L8.

[...]

PR #951, opened 2026-08-03
stack 1/6: triage the open issue surface and lock the bug plan

lidge-jun:
## Stack

**1/4 — triage and the plan**

Base: `dev`
Next: #908 long-context pricing tiers

## Summary

Sorts the open issue surface and locks the implementation order for the bugs nobody had reviewed. Documentation only — no `src/` changes in this layer.

- 22 label corrections applied across the open backlog, each grounded in the issue's own content or a maintainer comment
- `devlog/_plan/260803_bug_backlog_stack/` — six numbered docs, one per phase, at diff-level precision
- Evidence comments posted on #907, #875, #919, #545

## What the research round overturned

**#907 is not a regeneration.** `scripts/generate-jawcode-metadata.ts:22-24` reads `../jawcode/packages/ai/src/models.json`, and that canonical source carries the same stale Terra/Luna numbers across four provider bundles in a different repository. Regenerating today reproduces the defect exactly. The overlay cannot reach it either — `src/usage/cost.ts:139-145` gives a valid nonzero jawcode row precedence, and a nonzero-but-wrong row is never a candidate.

It also surfaced a coupling: `PRIORITY_MULTIPLIERS` stores Fast pricing as *ratios against the base rates*, calibrated when the bases were stale. Correcting the bases without recomputing them would fix an overcharge on standard requests and introduce an undercharge on Fast ones, silently.

**#875's reopen evidence tested a build 115 commits before the fix.** `git merge-base` proves `0b30283b6` is an ancestor of `5dd965a13` ("fix DeepSeek Responses over Codex WebSocket"), which is on `dev`. The reproduction was real; it reproduced a defect that has since been fixed. Left open with a re-test request rather than closed on an ancestry argument the reporter cannot check.

**#545's standing hypothesis was impossible.** The recorded analysis said the prepended OAuth identity block consumed the classifier's 64 output tokens. A system prompt is input; `max_tokens` caps output. The real defect is a round-trip loss: `thinking: {type: "disabled"}` becomes `reasoning: undefined`, which is indistinguishable from omitted — and for Sonnet 5, omitted means adaptive thinking **on**, with thinking tokens counted against the same 64.

## The audit gate

Four rounds, three FAIL. Every blocker was accepted and independently re-verified before amending:

1. **Long context × Fast cannot compose.** The plan multiplied both. OpenAI's Fast guide says long context is not supported in Fast mode, so the `$7.80` figure described a request that cannot exist. They are mutually exclusive regimes.
[...]

PR #944, opened 2026-08-03
fix(claude): separate reasoning summary parts in streamed thinking

DevMello:
## Summary

On the native ChatGPT backend the upstream Responses stream is translated straight into Anthropic SSE. Reasoning summaries arrive there in parts, each its own paragraph with a bold headline. The streaming translator folded every part into one thinking block with no separator, so thinking text rendered as run-on lines like "wants X.**Planning the fix**". The non-streaming translator in the same file already joins parts with a blank line. The stream now tracks part identity on the delta frames and emits the same blank-line separator at part boundaries. Frames without part indices keep a constant key and behave as before, and a new reasoning item still opens its own block.

## Verification

- Two new tests in `tests/claude-outbound.test.ts`: multi-part summaries match the JSON path's joined text, and same-part or index-free deltas never get a separator.
- Repro harness: streamed, collected, and JSON-path thinking text are now identical for the same upstream frames.
- `bun run test`, `typecheck`, `lint:gui`, `privacy:scan`.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streamed reasoning display by clearly separating distinct summary sections.
  * Preserved continuous text when updates belong to the same section or lack section indicators.
  * Aligned streaming and non-streaming reasoning output for consistent results.

* **Tests**
  * Added coverage for multiple reasoning sections and streaming behavior.

PR #954, opened 2026-08-03
stack 4/7: keep an explicit thinking disable through translation (#545)

lidge-jun:
## Stack

**4/4 — Claude Desktop classifier thinking round-trip**

Base: `codex/carry-contributor-bugfixes` (#953)

## Summary

Claude Desktop 3P Auto Mode sends `thinking: {type: "disabled"}` with `max_tokens: 64` and a `</block>` stop sequence. The instruction was dropped in translation, so the model thought anyway and spent the 64-token budget before it could close the tag — and Claude Code retried, up to five times per tool approval. The reporter measured 1,084 truncated requests against 143 that completed.

- `src/claude/inbound.ts` — an explicit disable is preserved as the parser's `"none"` sentinel instead of becoming `undefined`
- `src/adapters/anthropic.ts` — emit `thinking: {type: "disabled"}` for models that both default to thinking-on and accept the explicit disable

## The standing hypothesis was impossible

The recorded analysis on #545 said the prepended OAuth identity block consumed the classifier's 64 output tokens. It cannot: the identity goes into the **system** prompt, and `max_tokens` caps **output**. Different budgets. This repository had already reached that conclusion once — `devlog/_fin/260728_bug_bundle_resolution/030_claude_system_dedup.md` abandoned an identity-dedup patch for exactly this reason — and I re-derived the rejected theory before testing it.

The real chain is a round-trip fidelity loss:

1. `src/claude/inbound.ts:494` treated `disabled` as *nothing to translate*, leaving `reasoning` undefined
2. that is indistinguishable from a request that never mentioned thinking, and `tests/claude-inbound.test.ts` pinned both to `undefined`
3. the adapter emits `thinking` only for a real non-`none` effort, so the field was omitted outbound
4. for Sonnet 5, an omitted `thinking` field means adaptive thinking is **ON**, and thinking tokens count against `max_tokens`

The client asked for no thinking and got thinking. That also explains the shape of the reporter's data: the 143 requests that completed are the ones where thinking happened to stay short.

## The gate is deliberately narrow

`usesAdaptiveThinking()` looks like the right predicate and is not — it answers which *wire shape* a family accepts, not whether omission means thinking-on. The sets differ both ways: Fable always thinks and **rejects** an explicit disable, while Opus 4.7/4.8 use the adaptive wire but leave thinking off when omitted. Reusing it would have required breaking a passing test (`tests/anthropic-reasoning.test.ts` asserts `claude-fable-5` + `"none"` sends no thinking config) in order to ship a production 400.

So `supportsExplicitThinkingDisable()` is seeded with `sonnet: [5,0]` only. Widen it per family with vendor evidence.

## What the audit caught

Three rounds, two FAIL. Both findings were reproduced at runtime before fixing.

**Round 1** — `anthropic/claude-sonnet-5` silently missed the gate. A `modelMap` entry can point at a routed destination that routing decodes back into a slash-carrying native id, and the predicate anchored on `^claude-`.

**Round 2** — my own round-1 fix was worse than the bug. Normalizing with `lastIndexOf("/")` repaired the prefix case and broke the suffix case (`claude-sonnet-5/variant`), and because the adaptive-wire predicate shares that parse, such a model would have been sent obsolete manual `thinking.enabled` and 400d. A silent truncation traded for a hard failure.

Both now match the segment that actually begins with `claude-`, at either boundary, with all shapes pinned in the matrices — including a new adaptive-shape test covering the 400 path.

## Cross-provider note

A `modelMap` that routes such a request to Cursor now selects the model's **lowest** tier rather than its top one. Cursor has no off switch for a reasoning model, so the lowest tier is the closest honest reading of "do not think" — the previous behavior sent these to the maximum tier, the opposite of the caller's instruction. Pinned in `tests/cursor-effort-suffix.test.ts` so it is deliberate rather than emergent.

## Verification

- `bun run test` — 7711 pass, 8 skip, 0 fail, 507 files
- **red-green, both halves independently**: ablating the inbound preservation fails 3 tests; ablating the adapter emission fails 2 different ones

## Two honest limits

The tests prove the **wire shape**, not that the retries stop. Confirming that needs a live Claude Desktop 3P + Anthropic OAuth session showing the classifier terminating on `</block>` instead of `max_tokens`. I have asked the reporter rather than claiming the symptom fixed.

**This changes request construction on an Anthropic OAuth execution path, so `MAINTAINERS.md` requires explicit human security review** — no credential handling is touched, but the boundary is. Please do not merge on CI alone.

Refs #545

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recognition of Claude model identifiers across routed, versioned, and suffixed formats.
  * Explicitly disabling Claude thinking now reliably sends the correct setting for supported models.
  * Disabled reasoning is preserved consistently when translating requests and responses.
  * Requests with no reasoning preference continue using existing behavior.
  * Cursor routing now selects the lowest available reasoning tier when reasoning is explicitly disabled.
* **Tests**
  * Expanded coverage for Claude model detection, reasoning behavior, routing, token limits, and stop sequences.


lidge-jun:
## Stack navigation

The layers touch disjoint files — `devlog/`, `src/usage/`, the carried contributors' paths, and `src/claude/` + `src/adapters/anthropic.ts` — so any layer can be retargeted to `dev` and taken independently without a rebase conflict.

**#954 needs human security review** per `MAINTAINERS.md`: it changes request construction on an Anthropic OAuth execution path. It is last in the stack so the first three can land without waiting on that.

Carried in #953, with authorship preserved: #939, #942, #943, #944, #945, #948.

lidge-jun:
## Stack navigation

The layers touch disjoint files — `devlog/`, `src/usage/`, the carried contributors' paths, `src/claude/` + `src/adapters/anthropic.ts`, and `src/codex/` — so any layer can be retargeted to `dev` and taken independently without a rebase conflict.

**#954 needs human security review** per `MAINTAINERS.md`: it changes request construction on an Anthropic OAuth execution path. #955 sits above it in the chain but is independent of it in code, so if that review blocks, #955 can be retargeted to #953 without conflict.

Carried in #953, with authorship preserved: #939, #942, #943, #944, #945, #948.

lidge-jun:
## Stack navigation

lidge-jun:

| Layer | PR | Contents |
|---|---|---|

PR #955, opened 2026-08-03
stack 5/7: probe reset-derived cooldowns without waiting to be selected (#915)

### LINKED ISSUE #915: [Bug]: Reset-derived cooldowns can miss early recovery while another pool account remains eligible
### Client or integration

Codex App

### Area

### Summary

A reset-derived Codex cooldown can remain after the account recovers early when another pool account remains eligible.

An actual upstream `429` with reset headers creates a non-persistent cooldown. The cooled account is excluded, and the current recovery probe is acquired only after routing selects that account. If another account can always serve traffic, the cooled account is never selected, so it never receives the bounded probe. A forced fresh WHAM read may show recovery, but updating quota cache does not clear the matching routing cooldown.

Persistently pausing an account at displayed `100%` is not a valid workaround: OpenAI can reset earlier than the predicted timestamp, `100%` can remain advisory while an in-progress task still succeeds, and pause can outlive the restriction.

Expected behavior:

- Create cooldown only after an actual upstream rejection; a WHAM `100%` snapshot alone must not create cooldown, pause, or move an existing thread.
- Keep reset-derived exclusion non-persistent.
- At a bounded interval, permit one single-flight fresh quota refresh for a due reset-derived candidate even while another account serves requests.
- Fence by account identity, credential generation, quota scope, and cooldown generation.
- Clear only the matching reset-derived scope/generation when a complete fresh authoritative snapshot proves recovery.
- Retain cooldown on still-`100%`, incomplete, stale, unavailable, unrelated-scope, or replaced-credential results.
- Do not clear explicit `Retry-After` or headerless/default cooldowns through this path.
- Fail closed without delaying ordinary traffic on the healthy alternate.

Relevant source:

- `src/codex/routing.ts`: reset-derived health, generations, scope, selection, and probe leases.
- `src/codex/auth-context.ts`: probe admission currently occurs only after selection.
- `src/codex/auth-api.ts`: main/pool forced refresh and existing single-flight.
- `src/codex/quota.ts`: freshness and authoritative exhaustion.

Required tests: early recovery with another eligible account; still-exhausted/unknown/fetch-failure cases; concurrent single-flight; stale-generation rejection; main and pool accounts; shared vs Spark scope separation; explicit `Retry-After` untouched; all pool strategies re-include the recovered account; `paused` unchanged; and no cooldown/rebinding from a standalone `100%` snapshot.

Related: #433 (bounded probe/cap), #590/#599 (scoped cooldowns), #584/#585 (same-request failover after real rejection), and #657/#866 (separate reset-credit semantics).

### Reproduction

1. Configure accounts A and B.
2. Let A receive an actual upstream `429` with reset headers.
3. Keep B eligible so traffic continues through B.
4. Let OpenAI reset A earlier than the recorded deadline.
5. Force a fresh authoritative quota read while continuing traffic through B.

Current: A's quota cache updates, but its routing cooldown remains because A is never selected for the existing probe.

Expected: one bounded refresh clears only A's matching reset-derived generation, A becomes eligible immediately, and no persistent pause or percentage-only thread move occurs.

### Version

`2.10.0`; the relevant routing and quota-refresh behavior remained on `dev` at `075c2f34c060d068ed56985e76924a9f23aca39d` when checked on 2026-08-03.

### Operating system

Windows 11 Pro 10.0.26200 (platform-independent routing gap).

### Provider and model

OpenAI (Codex login / forward account pool); native Codex models and confirmed quota scopes.

### Logs or error output

```text
Selected Codex account is cooling down
```

None. Exact quota values, reset times, and account identifiers are omitted.

### Redacted configuration

```json
{
  "providers": {
    "openai": {
      "authMode": "forward",
      "codexAccountMode": "pool"
    }
  },
  "autoSwitchThreshold": 100,
}
```

### Checks

lidge-jun:
## Stack

**5/5 — cooldown early-recovery probe**

Base: `codex/545-classifier-thinking-disabled` (#954)

## Summary

A reset-derived cooldown is a *prediction*, and OpenAI can reset earlier than predicted. The account should come back — but the probe that would notice was unreachable.

The chain: cooled accounts are filtered out **before** every strategy runs (`src/codex/routing.ts:740-768`), and `resolveCodexAuthContext()` selects an account **first** and only then checks that account's probe lease (`src/codex/auth-context.ts:238-247`, `:276-298`). So with account B eligible, A is never selected, never reaches the lease code, and never recovers. The recovery mechanism was not broken; it was unreachable.

A forced WHAM read did not help either: `setAccountQuotaFromParsed()` writes the quota cache and never touches routing state.

- `claimDueCodexQuotaRecoveryProbes()` / `settleCodexQuotaRecoveryProbe()` — enumerate accounts independently of selection
- runs on the state sweeper's existing 60s tick, no new timer
- `src/codex/auth-context.ts` unchanged; request admission is untouched

## Deliberate narrowings

**`reset-derived` only.** The existing request-path predicate also admits `default`, but a `default` cooldown is the 60-second headerless fallback — there is no prediction to be early against. `retry-after` stays authoritative.

**Oldest-first ordering.** Bounded claims with stable config order starve: if more accounts are due than the per-pass limit and the early ones keep failing, they become due again and consume the budget forever. Claims sort by oldest `lastProbeAt ?? cooldownSince`, so a probed account rotates to the back whether or not it recovered.

**Spark is never claimed.** `GET /backend-api/wham/usage` takes no scope parameter and returns generic windows, so its result can never prove a spark recovery. Claiming it would spend the account's one claim per pass to settle `false` and leave the shared scope — which that evidence *can* clear — cooled behind it.

**Main account excluded from this first cut.** Pool accounts have generation-aware single-flight; `fetchMainAccountInfo()` has none, so a sweeper probe could race a dashboard refresh into parallel WHAM requests. Giving main the same admission semantics is its own change.

**Settle is narrow.** It clears only the exact map entry, only while the lease, cooldown generation, and credential generation all still match; any mismatch releases the lease and *retains* the cooldown. It does not use `clearCodexAccountCooldown()` (clears every scope) and does not route through `recordCodexUpstreamOutcome(200)` (also mutates account-wide failure state).

## What the audit changed

Five rounds, four FAIL. Every finding was reproduced before fixing.

1. **Unknown plans failed open.** An unfamiliar `plan_type` with a finite weekly percentage cleared the cooldown on evidence we could not interpret.
2. **The first fix was the wrong shape.** An allowlist missed `prolite` — then the snapshot turned out to carry **21** distinct plan strings (`edu_plus`, `finserv`, `k12`, `quorum`, `self_serve_business_usage_based`, …) with 12 unclassified, and `CodexAccount.plan` is an unrestricted string. Every omission meant an account cooled *forever*: the defect this PR exists to fix, reintroduced as a typo-shaped hole. Replaced with `codexQuotaWindowForPlan()`, the parser's own binary rule, now shared by parsing, exhaustion, and recovery so the three cannot drift.
3. **Three tests passed vacuously.** Both lease-release cases asserted only that the cooldown survived, never that the lease came back — a stranded lease means that account is never probed again. The fairness test spaced passes 60s apart, inside the 5-minute probe interval, so already-probed accounts dropped out on their own and the ordering was never exercised.
4. **The replacement test was tautological** — it computed its expected window from the function under test, so ablating the rule to always-weekly still passed 18/18. The contract is now pinned in literals first.

## Verification

- `bun run test` — 7729 pass, 8 skip, 0 fail, 508 files
- **red-green**: ablating `claimDueCodexQuotaRecoveryProbes()` fails 5 tests; ablating the oldest-first sort fails the fairness test; ablating the window rule fails the plan matrix; ablating lease release fails both retention tests

The defining case: account A cooled reset-derived, B eligible, ordinary routing selects B — after the interval the worker probes **A's** credential and clears A's cooldown, with no routing call ever having selected A.

Fixes #915

## Summary by CodeRabbit

* **New Features**
  * Added automatic recovery checks for Codex accounts after quota cooldowns.
  * Recovery now supports plan-specific weekly or monthly quota windows.
  * Added safeguards to prevent stale, invalid, or incomplete recovery results from clearing cooldowns.
  * Recovery checks run automatically in the background and prioritize eligible accounts fairly.

* **Bug Fixes**
  * Improved handling of failed requests, timeouts, credential changes, and concurrent recovery attempts.


lidge-jun:
## Stack navigation

The layers touch disjoint files — `devlog/`, `src/usage/`, the carried contributors' paths, `src/claude/` + `src/adapters/anthropic.ts`, and `src/codex/` — so any layer can be retargeted to `dev` and taken independently without a rebase conflict.

**#954 needs human security review** per `MAINTAINERS.md`: it changes request construction on an Anthropic OAuth execution path. #955 sits above it in the chain but is independent of it in code, so if that review blocks, #955 can be retargeted to #953 without conflict.

Carried in #953, with authorship preserved: #939, #942, #943, #944, #945, #948.

lidge-jun:
## Stack navigation

lidge-jun:

| Layer | PR | Contents |
|---|---|---|

Issue #349, opened 2026-07-23
[Bug]: Vision sidecar is unusable from Codex App - noVisionModels catalog should advertise image input so images reach the proxy


lijianmac:
### Goal / Problem

The vision sidecar feature is advertised as a headline capability ("Give any model superpowers - image understanding via gpt-5.4-mini sidecar"), but it is **completely unusable from Codex App** -- the most common entry point for users. This is not an edge case; it affects every App user who tries to paste an image while using a text-only routed model.

**Concrete user workflow that fails:**

1. User configures a GLM provider (e.g. Volcengine Ark Agent Plan) with `adapter: anthropic`, `defaultModel: glm-5.2`
2. User adds `noVisionModels: ["glm-5.2"]` to the provider config (as documented)
3. User enables vision sidecar: `visionSidecar: { model: "gpt-5.4-mini", backend: "openai" }` (as documented)
4. User opens Codex App, selects `GLM.MacAgent/glm-5.2` as the active model
5. User pastes an image into the conversation
6. **Codex App immediately rejects the image** -- "model does not support image input"
7. No request is sent to the OpenCodex proxy
8. The vision sidecar never runs

The image never reaches the proxy. The sidecar is dead code for App users.

**Why this matters:**

- GLM via Volcengine is one of the cheapest high-quality coding models available. Many users route to it through OpenCodex specifically because they want to save OpenAI quota.
- These users have NO way to use images with GLM in Codex App, even though OpenCodex has the infrastructure to make it work.
- The same setup **works perfectly in Claude Code** -- I have verified this end-to-end. Claude Code does not perform client-side `inputModalities` pre-checks, so the image reaches the proxy, the sidecar describes it via `gpt-5.4-mini`, and GLM answers based on the text description. This proves the sidecar pipeline is sound; only the App's pre-check blocks it.

### Expected behaviour

When `noVisionModels` is configured for a provider AND the vision sidecar is enabled, OpenCodex should advertise `inputModalities: ["text", "image"]` in the catalog entry for those specific models. This allows the full pipeline to work:

1. Codex App sees image support in catalog -> lets the image through
2. Request with image reaches OpenCodex proxy
3. Vision sidecar activates: `gpt-5.4-mini` describes the image
4. Text description is fed to the upstream model (GLM/DeepSeek/etc.)
5. User gets image understanding end-to-end

### Why this is the right place to fix it

[...]

PR #963, opened 2026-08-03
fix(catalog): preserve provider capabilities on custom models

### LINKED ISSUE #962: [Bug]: Custom model rows drop provider reasoning metadata from the Codex catalog
## Client or integration

Codex CLI

## Area

Catalog / models

## Summary

When a provider model also has a dashboard-created `customModels` row, `gatherRoutedModels()` lets the custom row replace the provider-derived row without retaining provider capability metadata. For a local Ollama model marked with `noReasoningModels` and an empty `modelReasoningEfforts` list, the generated Codex catalog consequently advertises the generic reasoning ladder. Codex then rejects an ephemeral `spawn_agent` call during client-side preflight, before OpenCodex can route the child request.

The custom row should retain its explicit display name, context window, and input modalities while inheriting provider capability metadata. The generated catalog should expose no reasoning levels and no default reasoning level for this model.

## Reproduction

1. Configure a static local provider model and a custom row for the same provider/model ID using the redacted shape below.
2. Run `ocx sync`.
3. Inspect `ollama/qwen-coder-3b` in the generated Codex catalog.
4. Observe that `supported_reasoning_levels` contains the generic `low` through `ultra` ladder and `default_reasoning_level` is `medium`.
5. Start a fresh Codex process through the OpenCodex loopback proxy and request an ephemeral subagent using `ollama/qwen-coder-3b` with no reasoning.
6. Codex rejects the child before creation because `none` is not among the advertised supported efforts.

A source-level reproduction through the public boundaries is also deterministic: `gatherRoutedModels()` returns one custom row with `reasoningEfforts` missing, and `buildCatalogEntries()` therefore synthesizes the generic ladder.

## Version

Installed package: `@bitkyc08/opencodex` 2.10.0. The defect is also present at current `dev` commit `1d1c8038447c9cbd3825842a0ac9fb9db8d79ca1`.

## Operating system

Ubuntu 24.04 (aarch64)

## Provider and model

Ollama / `qwen-coder-3b`

## Logs or error output

```shell
Reasoning effort 'none' is not supported for model 'ollama/qwen-coder-3b'. Supported reasoning efforts: low, medium, high, xhigh, max, ultra
```

## Redacted configuration

```json
{
  "providers": {
    "ollama": {
      "adapter": "openai-chat",
      "baseUrl": "http://localhost:11434/v1",
      "liveModels": false,
      "models": ["qwen-coder-3b"],
      "selectedModels": ["qwen-coder-3b"],
      "noReasoningModels": ["qwen-coder-3b"],
      "modelReasoningEfforts": {
        "qwen-coder-3b": []
      }
    }
  },
  "customModels": [
    {
      "id": "redacted-uuid",
      "provider": "ollama",
      "modelId": "qwen-coder-3b",
      "displayName": "Qwen Coder 3B (local)",
      "contextWindow": 32768,
      "inputModalities": ["text"]
    }
  ]
}
```

## Checks

MarcTCruz:
## Summary

- preserve registry-enriched provider capability hints when a custom model shadows a provider model row
- keep explicit custom display name, context window, and modalities authoritative while retaining `noVisionModels` augmentation
- prevent the catalog from synthesizing generic reasoning levels for custom models declared through `noReasoningModels`

Fixes #962

## Verification

- `bun scripts/test.ts tests/catalog-custom-model-metadata.test.ts tests/catalog-vision-sidecar-modalities.test.ts` — 13 passed, 0 failed
- `bun run lint:gui` — passed
- `bun run doctor:gui:if-changed` — skipped as expected because no GUI files changed
- live CLI check with an isolated `OPENCODEX_HOME` — emitted `reasoningEfforts: []` for the duplicated Ollama custom model
- `bun run test` — 7,889 passed and 10 skipped; one pre-existing failure remained in `native-profile-drain-server.test.ts` (`Live/Realtime sideband retains main ownership...`). The same failure reproduces on clean `dev` at `1d1c8038`.

## Checklist

- [x] Scope is focused and avoids unrelated refactors.
- [x] Docs or release notes were considered; none are needed for this internal catalog-correctness regression.
- [x] Security-sensitive areas were considered; this change does not modify authentication, secret handling, or security defaults.

## Summary by CodeRabbit

* **New Features**
  * Custom catalog models now inherit provider capabilities and registry metadata when not explicitly configured.
  * Explicit custom context limits and input modalities remain unchanged.
  * Vision sidecars continue to add image support where applicable.
* **Bug Fixes**
  * Improved handling of context limits, reasoning capabilities, and modality metadata for custom models.

### DIFF
--- a/src/codex/catalog/provider-fetch.ts
+++ b/src/codex/catalog/provider-fetch.ts
@@ -780,26 +780,26 @@ async function gatherRoutedModelsUncached(
   replaceLastComboCatalogOmissions(localOmissions);
   all.sort((a, b) => (a.provider === b.provider ? a.id.localeCompare(b.id) : a.provider.localeCompare(b.provider)));
   // Enriched (registry-hydrated) provider clones, keyed by name — the same view used above so
-  // custom rows get the same noVisionModels / inputModalities treatment as discovered rows.
+  // custom rows get the same capability hints as discovered rows.
   const enrichedByName = new Map(activeProviders);
   const customModels = (config.customModels ?? []).map(cm => {
     const rawProvider = config.providers[cm.provider];
-    const supportsReasoningSummaries = configuredReasoningSummarySupport(rawProvider, cm.modelId);
+    const enrichedProvider = enrichedByName.get(cm.provider) ?? rawProvider;
+    const providerCap = cm.contextWindow === undefined ? providerContextCap(config, cm.provider) : undefined;
+    const hints = enrichedProvider
+      ? catalogHintsFromProviderConfig(cm.provider, enrichedProvider, cm.modelId, providerCap)
+      : {};
     const base: CatalogModel = {
+      ...hints,
       id: cm.modelId,
       provider: cm.provider,
       // Display-only label: never feeds routing (customModels are keyed by routedSlug below).
       ...(cm.displayName ? { displayName: cm.displayName } : {}),
       ...(cm.contextWindow ? { contextWindow: cm.contextWindow } : {}),
       ...(cm.inputModalities ? { inputModalities: cm.inputModalities } : {}),
-      ...(typeof supportsReasoningSummaries === "boolean" ? { supportsReasoningSummaries } : {}),
     };
-    // Vision-sidecar coverage ONLY: if the custom model is in the enriched provider's
-    // noVisionModels, advertise image input so the Codex app lets images reach the sidecar
-    // (#349/#344). Deliberately NOT the full applyProviderConfigHints pass — custom rows are a
-    // user override, so their explicit contextWindow / inputModalities / reasoning fields must be
-    // preserved verbatim (the hint pass would cap context and overwrite modalities from registry).
-    const enrichedProvider = enrichedByName.get(cm.provider) ?? rawProvider;
+    // Explicit custom modalities win over hints, but noVisionModels still needs to append image
+    // so the Codex app permits attachments to reach the proxy's vision sidecar.
     if (enrichedProvider && modelInList(enrichedProvider.noVisionModels, base.id)) {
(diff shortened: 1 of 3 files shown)


Wibias:

lidge-jun:
Closing this in favour of **#965**, which fixes the same issue. Both landed within an hour of each other on #962 — that is a collision, not a fault, and the diff itself is clean work.

Why #965 won, concretely. #962 is about a custom row *replacing* a same-slug provider-derived row. #965 inherits from the row actually being replaced, so it also retains live `/models` metadata such as normalized `capabilities`. This PR instead recomputes `catalogHintsFromProviderConfig()` for **every** custom row (`src/codex/catalog/provider-fetch.ts:785-809`), including custom-only rows that never had a provider counterpart — broader than #962 requires, and it cannot retain discovered metadata because it rebuilds from config rather than inheriting.

The part that decided it: this PR changes `tests/catalog-vision-sidecar-modalities.test.ts:102-126` from asserting "no registry reasoning metadata leaks onto an unmatched custom override" to expecting that leak, and removes three `fetch should not be called` guards. That assertion encodes a deliberate earlier decision from the #349/#344 vision-sidecar work. Rewriting it to accommodate a broader change is the kind of edit that should be its own proposal with its own argument, not a step inside a bug fix.

Verified rather than assumed — both branches were checked out and ablated. #965's ablation fails exactly one test, the #962 regression. This PR's ablation fails two, one of which is the rewritten contract above.

Your implementation does handle `noReasoningModels` correctly, and if you think the unmatched-custom-row behaviour *should* change, that is a real question worth raising on its own — I would read that issue.

#965 is carried in **#973** (stack 6/6). Thanks for the fix, and sorry it was the duplicate.

PR #964, opened 2026-08-03
fix(providers): activate vision sidecar for NVIDIA NIM text-only models

### LINKED ISSUE #956: [Bug]: Vision sidecar never triggers for NVIDIA NIM text-only models (registry `nvidia` entry missing `noVisionModels`)
### Client or integration

Codex App (also reproducible via Direct HTTP/API client)

### Area

Proxy and routing · Catalog / models · Provider adapter

### Summary

The built-in `nvidia` provider entry in `src/providers/registry.ts` does not declare `noVisionModels`, so the vision sidecar is **never** activated for NVIDIA NIM text-only models (e.g. `deepseek-ai/deepseek-v4-flash`, `nvidia/nemotron-3-ultra-550b-a55b`, `z-ai/glm-5.2`, `minimaxai/minimax-m3`, `moonshotai/kimi-k2.6`). Users must manually add `noVisionModels` to the provider entry in `~/.opencodex/config.json` as a workaround.

Expected: a text-only NIM model receives an image attachment and the proxy describes it through the vision sidecar before calling the upstream.
Actual: `planVisionSidecar` returns `undefined` (the model is not in `provider.noVisionModels`), so raw image parts are either forwarded to the text-only upstream or the catalog advertises the model as text-only and the app blocks the attachment.

### Reproduction

1. Install `@bitkyc08/opencodex` 2.10.0 and start the proxy (`ocx start`).
2. Add a provider entry for NVIDIA NIM in `~/.opencodex/config.json` with only the documented fields (an empty or absent `noVisionModels` matches the registry default, which has **no** `noVisionModels` for `nvidia`):
   ```json
   {
     "providers": {
       "nvidia": {
         "adapter": "openai-chat",
         "baseUrl": "https://integrate.api.nvidia.com/v1",
         "authMode": "key",
         "apiKey": "nvapi-<redacted>"
       }
     }
   }
   ```
3. Query the routed model list and confirm the model has no image modality:
   ```
   GET /api/models  →  nvidia/deepseek-ai/deepseek-v4-flash  inputModalities: absent
   ```
4. Send a request that carries an image to the text-only model, e.g. via the Codex App (attach a screenshot) or directly:
   ```
   POST /v1/responses
   { "model": "nvidia/deepseek-ai/deepseek-v4-flash", "input": [ { "role": "user", "content": [ { "type": "image_url", "image_url": { "url": "data:image/png;base64,..." } } ] } ] }
   ```
5. Observe: no `[vision]` sidecar call happens (nothing in the logs); the image either 400s at the upstream or is stripped/blocked.
6. Workaround: set `noVisionModels: ["deepseek-ai/deepseek-v4-flash"]` on the provider entry and restart. Now `/api/models` advertises `inputModalities: ["text","image"]` and the sidecar describes the image successfully (verified end-to-end with `describeImage` returning a description via the OpenAI forward sidecar).

### Version

2.10.0 (npm `@bitkyc08/opencodex`); Codex runtime 0.146.0

### Operating system

Windows 11 (current), America/Fortaleza timezone

### Provider and model

nvidia / deepseek-ai/deepseek-v4-flash (also affects nvidia/nvidia/nemotron-3-ultra-550b-a55b, z-ai/glm-5.2, minimaxai/minimax-m3, moonshotai/kimi-k2.6)

### Logs or error output

No `[vision]` log lines are emitted for these models. The model rows in `GET /api/models` carry no `inputModalities`. With the workaround applied, the sidecar path works and returns descriptions.

### Redacted configuration

```json
{
  "providers": {
    "nvidia": {
      "adapter": "openai-chat",
      "baseUrl": "https://integrate.api.nvidia.com/v1",
      "authMode": "key",
      "apiKey": "nvapi-<redacted>"
    }
  },
  "visionSidecar": {
    "backend": "openai",
    "model": "gpt-5.6-luna"
  }
}
```

### How the vision sidecar is gated

`planVisionSidecar` in `src/vision/index.ts` returns a plan only when the routed model is listed in `provider.noVisionModels`:

```ts
if (!modelInList(provider.noVisionModels, modelId)) return undefined;
if (!messagesHaveImage(parsed)) return undefined;
```

The registry `nvidia` entry (`src/providers/registry.ts`, around line 1079) sets `noReasoningModels` / `preserveReasoningContentModels` for the Kimi family but **no `noVisionModels`**:

```ts
{
  id: "nvidia", label: "NVIDIA NIM", baseUrl: "https://integrate.api.nvidia.com/v1", adapter: "openai-chat", authKind: "key", dashboardUrl: "https://build.nvidia.com",
  freeTier: true,
  parallelToolCalls: false,
  noReasoningModels: NVIDIA_NIM_KIMI_MODELS,
  modelReasoningEfforts: Object.fromEntries(NVIDIA_NIM_KIMI_MODELS.map(id => [id, []])),
  preserveReasoningContentModels: NVIDIA_NIM_KIMI_THINKING_MODELS,
  note: "Free tier on NVIDIA NIM — API key still required (get a free key at build.nvidia.com).",
},
```

Other registry entries covering the same text-only model families do declare it, e.g. `ollama-cloud` lists `deepseek-v4-pro`, `deepseek-v4-flash`, `nemotron-3-ultra`, `glm-5.2`, etc., and `alibaba-intl` lists `deepseek-v4-pro/flash`, `glm-5.2`, `MiniMax-M2.5`.

The NIM `/v1/models` endpoint does not expose `input_modalities`, so the registry is the only source of truth for capability classification.

### Impact

For every text-only NIM model without `noVisionModels`:

1. The proxy forwards raw image parts to a text-only upstream (which rejects them with a 400 or silently ignores them), or
2. The Codex catalog advertises the model as text-only (no `input_modalities: ["image"]`), so the Codex app blocks image attachments client-side before the sidecar ever runs.

Confirmed against 2.10.0: all 101 discovered `nvidia/*` rows in `/api/models` carry no `inputModalities`, and the 4 enabled text-only models (`deepseek-ai/deepseek-v4-flash`, `minimaxai/minimax-m3`, `nvidia/nemotron-3-ultra-550b-a55b`, `z-ai/glm-5.2`) never produce a vision plan.

### Suggested fix

Add `noVisionModels` to the registry `nvidia` entry for the known text-only NIM model families (deepseek-v4-flash/pro, glm-5.x, nemotron-3-ultra/super, minimax-m3, kimi-k2.x, qwen, etc.), matching how `ollama-cloud` / `alibaba-intl` classify the same families. Vision-capable NIM models (e.g. `meta-llama-3.2-11b-vision-instruct`, `nvidia/nemotron-nano-vl-8b-v1`) must stay out of the list.

### Checks

Yuxin-Qiao:
## Summary

Fixes #956. The built-in `nvidia` provider entry in `src/providers/registry.ts` did not declare `noVisionModels`, so the vision sidecar never activated for NVIDIA NIM text-only models (`deepseek-ai/deepseek-v4-flash`, `z-ai/glm-5.2`, `minimaxai/minimax-m3`, `moonshotai/kimi-k2.6`, `nvidia/nemotron-3-ultra-550b-a55b`, ...). Raw image parts were forwarded to a text-only upstream (400 / silently ignored) or the catalog advertised the model as text-only, making the Codex app block attachments client-side.

NIM's `/v1/models` carries no input-modality metadata, so the registry is the only source of truth for capability classification.

## Changes

- `src/providers/registry.ts`: add `NVIDIA_NIM_NO_VISION_MODELS` and wire it into the `nvidia` entry as `noVisionModels`.
  - Snapshot of text-only chat/code families from the live NIM catalog (fetched 2026-08-04, 102 models) plus the documented kimi family.
  - Vision-capable models stay out: `meta/llama-3.2-11b/90b-vision-instruct`, `microsoft/phi-3-vision-128k-instruct`, `adept/fuyu-8b`, `google/deplot`, `microsoft/kosmos-2`, `nvidia/llama-3.1-nemotron-nano-vl-8b-v1`, `nvidia/nemotron-nano-12b-v2-vl`, `nvidia/neva-22b`, `nvidia/vila`, `nvidia/nemotron-3-nano-omni-30b-a3b-reasoning` (multimodal), plus non-chat endpoints (embeddings, guards, translators, detectors, image/video generators).
  - `modelInList` semantics: exact-id matching for slash-namespaced NIM ids; new text-only ids must be appended individually (same convention as the existing NIM kimi lists).

With the list in place, `planVisionSidecar` returns a plan for image-carrying requests to these models and the catalog advertises `inputModalities` including `"image"` (so the app lets attachments reach the sidecar).

## Tests

- `tests/nvidia-nim-hardening.test.ts`: 4 new regression tests
  - registry list contains the issue's text-only families and excludes the vision-capable models
  - bare persisted config inherits `noVisionModels` via the router merge
  - `planVisionSidecar` plans for a text-only NIM model with an image, stays `undefined` without an image, and stays `undefined` for a vision-capable NIM model
  - catalog hints advertise image input for text-only models and not for vision-capable ones

## Verification

- `bun test tests/nvidia-nim-hardening.test.ts` → 18 pass / 0 fail
- `bun test tests/provider-registry-parity.test.ts tests/vision-anthropic.test.ts tests/claude-sidecar-override.test.ts tests/codex-catalog.test.ts tests/nvidia-nim-hardening.test.ts` → 176 pass / 0 fail
- `bun x tsc --noEmit` → clean
- Full suite (522 files): 7210 pass / 541 fail / 7 errors on this branch vs 7206 pass / 541 fail / 7 errors on unmodified `dev` in the same sandbox — the failures are identical and environmental (sandbox blocks outbound network and socket binds; e.g. OAuth replay, `/v1/live` relay, and `EADDRINUSE` on ephemeral ports), so this change introduces no regressions.

## Notes

The NIM catalog is dynamic; new text-only ids added upstream must be appended to `NVIDIA_NIM_NO_VISION_MODELS` (noted in a registry comment). Vision-capable NIM models keep their previous behavior.

## Summary by CodeRabbit

* **New Features**
  * Improved NVIDIA NIM model capability detection for text-only models.
  * Image requests for models without native vision support are now routed through the vision processing fallback.
  * Model listings now accurately indicate supported image modalities.

* **Bug Fixes**
  * Improved handling of image and non-image requests across NVIDIA NIM models.

* **Tests**
  * Added coverage for model classification, image handling, and capability reporting.

### DIFF
--- a/src/providers/registry.ts
+++ b/src/providers/registry.ts
@@ -520,6 +520,79 @@ const NVIDIA_NIM_KIMI_MODELS = [
   ...NVIDIA_NIM_KIMI_THINKING_MODELS,
   "moonshotai/kimi-k2-instruct", "moonshotai/kimi-k2-instruct-0905",
 ];
+// 260804 NVIDIA NIM text-only chat/code families (issue #956): the NIM /v1/models catalog
+// carries no input-modality metadata, so the registry is the only source of truth. Snapshot
+// from the live catalog (2026-08-04) plus the documented kimi family above. Vision-capable
+// NIM models (llama-3.2-*-vision-instruct, phi-3-vision, kosmos-2, fuyu-8b, deplot, neva-22b,
+// vila, nemotron-*-vl, nemotron-nano-12b-v2-vl, nemotron-3-nano-omni) and non-chat endpoints
+// (embeddings, guards, translators, detectors, image/video) stay out. New text-only ids must
+// be appended individually.
+const NVIDIA_NIM_NO_VISION_MODELS = [
+  "01-ai/yi-large",
+  "ai21labs/jamba-1.5-large-instruct",
+  "aisingapore/sea-lion-7b-instruct",
+  "bigcode/starcoder2-15b",
+  "databricks/dbrx-instruct",
+  "deepseek-ai/deepseek-coder-6.7b-instruct",
+  "deepseek-ai/deepseek-v4-flash",
+  "deepseek-ai/deepseek-v4-pro",
+  "google/codegemma-1.1-7b",
+  "google/codegemma-7b",
+  "google/gemma-2b",
+  "google/recurrentgemma-2b",
+  "ibm/granite-3.0-3b-a800m-instruct",
+  "ibm/granite-3.0-8b-instruct",
+  "ibm/granite-34b-code-instruct",
+  "ibm/granite-8b-code-instruct",
+  "meta/codellama-70b",
+  "meta/llama-3.1-70b-instruct",
+  "meta/llama-3.1-8b-instruct",
+  "meta/llama-3.2-1b-instruct",
+  "meta/llama-3.2-3b-instruct",
+  "meta/llama-3.3-70b-instruct",
+  "meta/llama2-70b",
+  "microsoft/phi-3.5-moe-instruct",
+  "minimaxai/minimax-m3",
+  "mistralai/codestral-22b-instruct-v0.1",
+  "mistralai/mistral-7b-instruct-v0.3",
+  "mistralai/mistral-large",
+  "mistralai/mistral-large-2-instruct",
+  "mistralai/mistral-medium-3.5-128b",
+  "mistralai/mistral-nemotron",
+  "mistralai/mixtral-8x22b-v0.1",
+  "moonshotai/kimi-k2.6",
+  "moonshotai/kimi-k2.5",
+  "moonshotai/kimi-k2-thinking",
+  "moonshotai/kimi-k2-instruct",
+  "moonshotai/kimi-k2-instruct-0905",
+  "nv-mistralai/mistral-nemo-12b-instruct",
+  "nvidia/llama-3.1-nemotron-51b-instruct",
+  "nvidia/llama-3.1-nemotron-70b-instruct",
+  "nvidia/llama-3.1-nemotron-nano-8b-v1",
+  "nvidia/llama-3.1-nemotron-ultra-253b-v1",
+  "nvidia/llama-3.3-nemotron-super-49b-v1",
+  "nvidia/llama-3.3-nemotron-super-49b-v1.5",
+  "nvidia/llama3-chatqa-1.5-70b",
+  "nvidia/mistral-nemo-minitron-8b-8k-instruct",
+  "nvidia/nemotron-3-nano-30b-a3b",
+  "nvidia/nemotron-3-super-120b-a12b",
+  "nvidia/nemotron-3-ultra-550b-a55b",
+  "nvidia/nemotron-4-340b-instruct",
+  "nvidia/nemotron-mini-4b-instruct",
+  "nvidia/nemotron-nano-3-30b-a3b",
+  "nvidia/nvidia-nemotron-nano-9b-v2",
+  "openai/gpt-oss-120b",
+  "openai/gpt-oss-20b",
+  "poolside/laguna-xs-2.1",
+  "stepfun-ai/step-3.7-flash",
+  "thinkingmachines/inkling",
+  "writer/palmyra-creative-122b",
+  "writer/palmyra-fin-70b-32k",
+  "writer/palmyra-med-70b",
+  "writer/palmyra-med-70b-32k",
+  "z-ai/glm-5.2",
+  "zyphra/zamba2-7b-instruct",
+];
 const KIMI_CODING_MODEL_CONTEXT_WINDOWS: Record<string, number> = Object.fromEntries(
   KIMI_CODING_MODELS.map(id => [id, id === "k3[1m]" ? KIMI_K3_1M_CONTEXT_WINDOW : KIMI_K3_STANDARD_CONTEXT_WINDOW]),
 );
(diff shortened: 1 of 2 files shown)


Wibias:

lidge-jun:
Confirmed as a real defect and left open for its own review rather than folded into the current review stack (#951–#973).

The reasoning: this is a provider-registry change — `nvidia` lacks `noVisionModels`, so the vision sidecar never activates and raw image parts reach a text-only upstream. That is a genuine bug with a clear fix, but it shares no theme, files, or risk surface with the stack, and carrying it would only make the stack harder to review while delaying yours behind five other layers.

Worth flagging for whoever reviews it: the registry change is small, but the blast radius is every NIM text-only model listed in the PR, so the useful evidence is a live NIM request with an image part before and after. Fixes #956.

lidge-jun:
Superseded by **#980** (stack 7/7), commit `f557f9173`. Closing this one — reopening is one click if you disagree with anything below.

**Your finding is correct and it is now fixed.** The `nvidia` entry had no `noVisionModels`, so `planVisionSidecar` never fired and the catalog never advertised image input. Your diagnosis of the mechanism was exactly right.

**What I changed, and why I owe you the detail.** Six ids in the submitted list are natively image-capable per NVIDIA's own documentation:

| Id | listed as | NVIDIA documents |
|---|---|---|
| `thinkingmachines/inkling` | text-only | text, **image**, audio |
| `minimaxai/minimax-m3` | text-only | Text, **Image**, Video |
| `moonshotai/kimi-k2.6` | text-only | text, **image**, video (with a published `image_url` example) |
| `moonshotai/kimi-k2.5` | text-only | text, **image**, video |
| `stepfun-ai/step-3.7-flash` | text-only | text + image, documented as a VLM |
| `mistralai/mistral-medium-3.5-128b` | text-only | text + image |

Listing a vision-capable model there is a *silent* defect: the model could read the image, but the proxy substitutes another model's text description instead. No error, worse answers, extra cost, and no test fails.

**This is not a lapse on your part.** Issue #956's own body lists minimax-m3 and kimi-k2.6 as text-only, so the reporter and the PR shared the same premise, and review passed it through. It is what an unaudited list does over an open domain.

So every id was verified individually against NVIDIA per-model documentation (`devlog/_plan/260804_stack7_service_vision/011_nim_id_audit.md`): **26 confirmed text-only shipped, 6 moved to a vision list, 32 dropped** for having no current NVIDIA page. That last set includes `nvidia/nemotron-nano-3-30b-a3b`, a reversed-name typo of `nvidia/nemotron-3-nano-30b-a3b` which the same list also spells correctly.

The 16 vision-capable ids also needed explicit `modelInputModalities`. Removing them from `noVisionModels` alone is not enough — the catalog advertises image input only for list members, so they would have been published as text-only and the Codex app would block attachments before the native path ran. Your PR would have left those models unusable in a different way.

**Where I did worse than you.** My first two replacement designs were falsified at the audit gate. I proposed deriving text-only as the complement of a vision list (an unclassified id still ends up in neither list, so nothing changes), then a provider-level default-on rule (which would have advertised image input for every embedding, reranker, guard and OCR endpoint NIM serves, since `shouldExposeRoutedModel` filters only media-generation names). The list *shape* was not the fixable part; the audit was.

**What still is not fixed:** an id NVIDIA ships after this snapshot stays unclassified. NIM publishes no modality metadata, so an unknown id carries no signal at all. #956 closes with that bounded scope stated rather than implied.

Thank you for the report — the underlying bug was real and is fixed for every model named in the issue.

PR #966, opened 2026-08-03
fix(codex): keep pre-connection DNS/network failures off account health (#914)

### LINKED ISSUE #914: [Bug]: DNS and network reachability failures incorrectly rotate Codex pool accounts
### Client or integration

Codex App

### Area

### Summary

Pre-connection DNS and local network failures are currently recorded as failures of the selected Codex account. Regular Responses and native compact map every thrown non-timeout transport error to `connect_error`; after `upstreamFailoverThreshold` failures, OpenCodex soft-avoids that account, clears thread affinity, and changes the effective account.

This is incorrect for source-proven reachability failures such as `ENOTFOUND`, `EAI_AGAIN`, `ENETUNREACH`, `ENETDOWN`, `EHOSTUNREACH`, `ECONNREFUSED`, and equivalent bounded Bun error codes. Credential replacement cannot repair a machine-wide DNS or network outage, and concurrent agents can cross the default threshold almost immediately.

Expected behavior:

- Return the existing client-visible `502`, but classify proven pre-connection DNS/network reachability failures as account-neutral.
- Do not increment account failure streak, set soft-avoid/cooldown, clear affinity, or change runtime/persisted active account.
- Release any owned quota-probe lease without treating the failure as account evidence.
- Inspect stable codes through a bounded cause chain; do not trust an unanchored message substring.
- Preserve current handling for `ECONNRESET`, `EPIPE`, established-socket closure, `TimeoutError`, HTTP 5xx, actual `429/402`, `401/403`, and client cancellation.

Relevant source:

- `src/server/responses/core.ts`: native passthrough transport failure recording.
- `src/server/responses/compact.ts`: compact transport failure recording.
- `src/codex/routing.ts`: transient streak, soft-avoid, affinity clear, and active failover.
- `src/lib/upstream-retry.ts`: existing narrow transport classification.
- `src/codex/auth-context.ts`: probe-lease release.

Required tests: table-driven bounded cause-chain classification; negative message-only cases; routing state unchanged after three neutral failures; regular and compact end-to-end `ENOTFOUND` cases; concurrent neutral failures; and proof that an owned probe lease is released. Existing reset/socket/timeout/HTTP failover regressions must remain.

Related: #186/#194 (established-connection and transient recovery, which this must not undo), #714 (retry/amplification rather than account-state mutation), and #504/#506 (thresholded account failover).

### Reproduction

1. Configure at least two Codex pool accounts and bind a thread to account A.
2. Keep the default `upstreamFailoverThreshold: 3`.
3. Make the provider hostname unresolvable, or inject a pre-response error with stable code `ENOTFOUND`.
4. Send three concurrent or sequential requests.
5. Restore network reachability and inspect the next routing decision.

Current: A is soft-avoided, affinity can be cleared, and another account can become effective.

Expected: requests may return `502` during the outage, but A's health, affinity, and active-account state remain unchanged.

### Version

`2.10.0`; the relevant transport branches remained on `dev` at `075c2f34c060d068ed56985e76924a9f23aca39d` when checked on 2026-08-03.

### Operating system

Windows 11 Pro 10.0.26200; equivalent Node/Bun shapes should be covered on every supported platform.

### Provider and model

OpenAI (Codex login / forward account pool); native Codex models.

### Logs or error output

```text
Provider unreachable: getaddrinfo ENOTFOUND <provider-host>
```

None. Hostnames, account identifiers, and request identifiers are removed.

### Redacted configuration

```json
{
  "providers": {
    "openai": {
      "authMode": "forward",
      "codexAccountMode": "pool"
    }
  },
  "upstreamFailoverThreshold": 3,
}
```

### Checks

Yuxin-Qiao:
## Summary

Fixes #914: proven pre-connection DNS / network reachability failures no longer count against a Codex pool account's failure streak, so they cannot soft-avoid a healthy account, clear thread affinity, or change the effective account. The client still gets the existing `502` during the outage; the account state is untouched.

This implements the direction recorded in `devlog/_plan/260803_transport_attribution/000_plan.md`: connection-level failures belong to a **(provider, host) pair**, not a credential. Every pool account shares the host, so rotation cannot repair a machine-wide outage.

## Design

**1. Pre-connect classifier** (`src/lib/upstream-reachability.ts`, leaf module)

`isPreConnectReachabilityError()` matches stable `code` values through a bounded cause chain (depth 3). Verified empirically on Bun 1.3.14 (the exact runtime in the plan's probe):

```text
fetch("https://no-such-host.invalid/x") -> code:"ConnectionRefused", errno:0, no cause
fetch("http://127.0.0.1:1/x")           -> code:"FailedToOpenSocket", errno:0, no cause
(read-then-close)                       -> code:"ECONNRESET"
```

Matched set: Bun's `ConnectionRefused` / `FailedToOpenSocket` (DNS failure and TCP refusal share the class and alternate labels) plus the Node shapes (`ECONNREFUSED`, `ENOTFOUND`, `EAI_AGAIN`, `ENETUNREACH`, `ENETDOWN`, `EHOSTUNREACH`) for the non-Bun runtimes the repo claims to support. Message substrings are never matched (negative cases tested). `ECONNRESET`, `EPIPE`, TLS codes, timeouts, and unknown shapes keep their existing account-attributed handling.

Why this is not rejected design #1: the list is anchored to the labels Bun actually emits, verified by runtime probe, not the Node-only list that never fired.

**2. Host health ledger** (`src/codex/routing.ts`)

New `connect_neutral` outcome class. `recordCodexUpstreamOutcome` records it in a `(provider:host)` ledger with its own threshold (`isHostConnectOutage`, windowed like account health) and **concludes any owned quota-probe lease** without treating the failure as account evidence. Account streak, soft-avoid, thread affinity, and active/persisted account are never touched. The ledger is observational today; `isHostConnectOutage` is the hook where a host-level response (e.g. circuit break) would attach with its own audit.

**3. Redirect counterexample removed** (`fetch-helpers.ts` + pool sends)

Pool sends (`/v1/responses` passthrough, the alternate-account send, `/v1/responses/compact`) now use `redirect: "manual"`. A server that 307s to a dead host produces a relayed 3xx Response instead of a `ConnectionRefused` rejection after the credential was seen. Explicit 3xx policy: 3xx is the neutral class — relayed as-is, never account or host health evidence (the host proved reachable by responding).

Why this is not rejected design #3: the "no headers arrived" boundary is not used. The credential-visible **read-then-close** case (`ECONNRESET`) remains account-attributed and fails over exactly as before (2026-07-22 decision preserved, with a regression test). The only class treated as neutral is the one Bun provably emits when no connection was established.

## Residual class (documented, not argued away)

A server that reads the credential, returns `307` to a dead target, and whose follow-up is reached through some other transport path can still produce an account-neutral classification; with `redirect: "manual"` on pool sends that path is a relayed 3xx with an explicit policy. The mixed `5xx -> rejection` case resolves neutral on the regular, compact, and alternate-account sends — the 5xx evidence is lost inside `fetchWithTransientRetry` (the same hole recorded in the plan), and the final attempt's reachability failure is host-level.

## Tests

- Table-driven classifier tests: Bun and Node shapes, bounded cause chain, cyclic cause, message-only negatives, non-Error rejections (`tests/upstream-reachability.test.ts`)
- Routing unit tests: three neutral failures leave health/soft-avoid/affinity/active account untouched; host ledger window; owned probe lease released, someone else's lease preserved; 3xx neutrality (`tests/codex-routing.test.ts`)
- E2E: three concurrent DNS-shaped failures on `/v1/responses` return 502 and routing state is unchanged; recovery stays on account A; compact path parity; **real** Bun TCP refusal through a live proxy; read-then-close `ECONNRESET` still soft-avoids and fails over (`tests/issue-914-transport-attribution.test.ts`)
- Updated `tests/server-auth.test.ts` (the old test asserted the pre-fix behavior)

Validation: `bun run typecheck`, `bun run privacy:scan`, and `bun run test` (7933 pass; the only 10 failures are `provider management validation` tests that reproduce identically on a clean `origin/dev` checkout). Docs updated in `docs-site` (en/ko/ja/ru/zh-cn): DNS/TCP reachability failures are account-neutral and never count toward `upstreamFailoverThreshold`.

## Follow-up (review commit `80cd0cee`)

Addresses the review blockers on this PR:

1. **Mixed 5xx -> rejection keeps its account evidence.** `fetchWithTransientRetry()` now attaches the prior transient statuses to the final rejection (`TransientRetryEvidenceError`), and `classifyTransportFailureKind()` treats that evidence as account-attributable (`connect_error`) instead of the pre-connection neutral class. A genuine upstream 503 followed by a reachability rejection no longer vanishes from the account streak.
2. **Manual redirects extended to every credential-bearing sidecar send.** `redirect: "manual"` now also applies to web search (`src/web-search/executor.ts`), vision (`src/vision/describe.ts`), search (`src/server/search.ts`), images (`src/server/images.ts`), and live (`src/server/live.ts`), matching the existing Responses, Compact, and alternate-account sends. A 3xx is relayed as the explicit neutral class; it can no longer be followed into a dead-host rejection after the credential was seen.
3. **Exact-account sidecar recorder forwards host metadata (Codex review P2).** The fixed-account search/vision recorder now passes `host` / `lastFailureCode` into the `(provider, host)` ledger exactly like the pool recorder, so fixed-account DNS/TCP failures no longer undercount `hostConnectHealth`.

New tests: evidence-wrapper classification units, a mixed-503 e2e (`/v1/responses` 503 then `ConnectionRefused` stays on the account streak), a redirect activation e2e (pool sends relay 307 instead of following into a dead host), and an exact-account search e2e asserting the host ledger entry.

Validation on the review commit: `bun run typecheck` green, `bun run privacy:scan` green, `bun run test` 7958 tests with only the same 10 `provider management validation` failures (environmental destination-policy DNS answers for `*.example.test`, unrelated to this change) plus one load-sensitive crash-guard timeout that passes in isolation.

## Follow-up (review commit `71bd8210`)

Extends the manual-redirect policy to every credential-bearing forward send, per CodeRabbit review on `80cd0cee`:

- `/v1/responses` and `/v1/responses/compact` now gate `redirect: "manual"` on `provider.authMode === "forward"` instead of `usesCodexForwardPoolAuth`, so **direct-mode** sends (which carry the caller's credential) get the same no-follow behavior as pool sends.
- Relay paths preserve the upstream `Location` header (search, images, compact) so a relayed 3xx stays followable by the client.
- Docs (en/ja/ko/ru/zh-cn) now state that pre-connection DNS/TCP failures are tracked at provider-host scope and never affect account health, cooldowns, thread/session affinity, active-account selection, or Pool routing.

New tests: direct-forward 307 activation for Responses and Compact, search 307 + `Location` preservation with no follow, and a `Location` assertion on the pool 307 e2e. Validation: `bun run typecheck` and `bun run privacy:scan` green; focused suites 124/124; full suite 7941 pass with the same 10 environmental `provider management validation` failures plus two load-sensitive timing tests (crash-guard timeout, native-profile WebSocket close race) that pass in isolation.

## Follow-up (review commit `78c824dd`)

Addresses the remaining Codex review P2 findings on `71bd8210`:

- **Reset evidence preserved.** `fetchWithResetRetry` now attaches credential-visible reset evidence when a retried `ECONNRESET` attempt is followed by a different terminal rejection, so a mixed reset → `ConnectionRefused` stays account-attributed (`connect_error`) instead of classifying as pre-connection neutral. The evidence wrapper is renamed to `UpstreamRetryEvidenceError` and carries both transient 5xx statuses and a `resetSeen` flag; the classifier treats either as account evidence.
- **Manual redirects scoped to the ChatGPT forward credential path** in the images and live relays. Keyed API-key providers keep default redirect following, so their same-origin 307/308 routing still resolves inside the proxy where the API key lives.
- **Docs (ja/ko/zh-cn) restore the "proven" qualifier**: only proven pre-connection DNS/TCP failures are neutral; unconfirmed failures remain account-attributed.

New tests: reset-evidence classification units, a mixed reset → reachability e2e on `/v1/responses`, and a host-ledger null assertion on the search 307 e2e. Validation: `bun run typecheck` and `bun run privacy:scan` green; focused suites 129/129; full suite 7964 tests with the same 10 environmental `provider management validation` failures plus one load-sensitive 5s timing flake (shellStreamExec) that passes in isolation.

## Follow-up (review commit `bd55f1ab`)

Addresses the CodeRabbit CWE-319 findings on `71bd8210` (cleartext credential transmission):

- **Fail-closed scheme guard** (`credentialSendSchemeError`, shared outbound policy): `https` passes, `http` is allowed only for loopback hosts, and cleartext `http` to any remote host is rejected before a credential can leave the proxy.
- Applied at every credential-bearing send boundary: the `/v1/responses` forward passthrough, the compact forward path, and the images relay (both the ChatGPT forward credential and keyed provider API keys). A hand-edited custom forward provider with an `http` baseUrl now gets a `502` at the send boundary instead of transmitting `Authorization` over cleartext (loadConfig accepts that shape; management validation already reserves `forward` for the canonical built-in provider, so the config-level check would be unreachable).

New tests: scheme-guard units (`https`, loopback `http`, remote `http`, malformed), a `/v1/responses` e2e asserting `502` plus zero outbound fetches for an `http` forward provider, and an images e2e for a keyed `http` provider. Validation: `bun run typecheck` and `bun run privacy:scan` green; focused suites 162/162; full suite 7970 tests with the same 10 environmental `provider management validation` failures plus one load-sensitive 5s timing flake (crash-guard) that passes in isolation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved upstream failure handling by distinguishing provider/host connectivity issues from account-specific failures.
  - Prevented DNS/TCP failures before connection from incorrectly affecting account health or triggering failover.
  - Preserved redirect responses and their destination headers instead of following redirects.
  - Improved timeout and connection-error classification across relay features.

- **Documentation**
  - Clarified `upstreamFailoverThreshold` behavior across supported languages.

- **Tests**
  - Added coverage for connectivity tracking, redirects, retries, recovery, and failure attribution.


Wibias:

lidge-jun:
Reviewed against the four designs this repository already falsified for #914 (`devlog/_plan/260803_transport_attribution/000_plan.md`). **Keeping this open, and it supersedes #922** — but I am requesting changes rather than merging, because two of those falsifications still survive here.

First, credit where it is due. This is a real advance: it uses the actual Bun 1.3.14 error labels including both alternating ones, it does **not** repeat the hostname-resolution design, and it closes the redirect counterexample on the pool Responses and Compact paths with manual redirects. TLS and fake-IP certificate codes correctly stay account-scoped. That is more than the previous four managed.

**Blocker 1 — mixed 5xx → rejection still loses the attributable failure.** `fetchWithTransientRetry()` discards prior transient responses when a later attempt rejects (`src/lib/upstream-retry.ts:220-236`), so the outer classifier sees only the final rejection:

```text
{"attempt":1,"serverStatus":503}
[upstream-retry] transient 503 — retrying (2/2)
{"finalRejected":true,"code":"ConnectionRefused","classification":"connect_neutral"}
```

A genuine account-attributable 503 becomes account-neutral. That is exactly the hole the earlier audit documented. The direction that fixes it is preserving per-attempt outcome evidence rather than classifying only the terminal promise.

**Blocker 2 — falsification 3 survives on the five newly-classified sidecar paths.** Manual redirects were added only to Responses (`src/server/responses/core.ts:1742-1757`) and Compact (`src/server/responses/compact.ts:393-405`). Web search, vision, search, images, and live still use default-follow fetch, so a credential-bearing sidecar that receives a 307 to a dead host is classified neutral *after* the origin has already read the header:

```text
redirect:"follow"  serverSawAuthorization:"Bearer credential-follow"  resolved:false
redirect:"manual"  serverSawAuthorization:"Bearer credential-manual"  resolved:true 307
```

This is the part I would most like you to reconsider: the sidecar expansion carries the unresolved hole *beyond* #914's original sites, which makes the blast radius larger than the bug.

**Scope.** 822/-39 across 20 files bundles the core attribution fix, a new host-health ledger, five sidecar families, exact-account recorder changes, and 518 lines of tests. The core Responses/Compact fix alone would be reviewable and likely mergeable; splitting it out is the fastest path to landing something.

Also worth adding: a redirect activation test and a mixed-503 test. The current suite proves the implemented classification but not the two cases above — I ablated the classifier and the existing tests do go red, so they are load-bearing for what they cover.

**#922 is superseded by this.** It misses one alternating Bun label entirely, keeps default redirects, has no host-level recording, and bundles unrelated probe-lease work while sitting at `CHANGES_REQUESTED`.

Thanks for taking this on — four designs died here before yours, and this is the first one that survives half the gauntlet.

lidge-jun:
Superseded by #1023 (head `a8e10eb0b21a`), which is **open and not yet merged** at the time of this comment.

Thank you for this — of the three candidate approaches to #914, yours was the closest semantic source, and #1023 is built on your reading of the problem.

**Adopted from this PR**

- The core classification: a failure that happens *before* a connection is established is a transport fact, not an account fact, so it must not touch account health, affinity, or the active-account selection.
- The bounded stable-code approach to recognizing those failures, rather than message-substring matching.
- Applying the same semantics to the native Compact path, not only the regular Responses path.
- The regular/Compact test split.

**Deliberately not matched**

- The sidecar-wide redirect handling and the cleartext credential-send guard. Both are defensible changes, but they are outside #914's boundary and would have expanded the blast radius of this fix from a handful of send sites to 20+ files. They belong in their own PR, and if you open one it will be reviewed on its own merits.

**Status disclosure**

#1023 is rebased onto current `dev` and revalidated before merge; that revalidation is still pending as of this comment. If it fails to land, or if the final merged version no longer preserves the behavior described above, this PR is reopened rather than left closed against nothing.

Full equivalence reasoning: `devlog/_plan/260805_bug_stack_campaign/030_issue914_dns_transport_attribution.md`.

PR #967, opened 2026-08-03
fix(codex): accept monthly-classified snapshots and probe-owned refresh generations (#955 review)

Yuxin-Qiao:
## Summary

Follow-up to #955 that addresses both unresolved Codex review threads on the cooldown early-recovery probe:

**1. Duration-classified monthly snapshots were never accepted (P1).** `isCompleteCodexQuotaRecoverySnapshot()` required `weeklyPercent` for every non-Go/Free plan based on the plan name. But `parseUsageQuota()` classifies windows by duration: a Team plan response whose primary window is explicitly monthly (e.g. `limit_window_seconds: 2628000`, no secondary) parses to `monthlyPercent` only. The probe therefore rejected every successful fresh read for those accounts and they stayed cooled until the predicted expiry — the same "cooled forever" defect this work exists to fix, reintroduced for monthly-window plans.

**2. The probe's own token refresh was treated as a replacement (P2).** `getValidCodexToken()` refreshes a near-expiry access token inside the probe fetch and advances the credential generation by exactly one before WHAM completes. `settleCodexQuotaRecoveryProbe()` required the claim-time generation to match exactly, so a successful fresh reading under the new live generation was rejected and recovery waited another probe interval. `saveCodexAccountCredentialIfGeneration()` preserves `replacedAt` while `saveCodexAccountCredential()` stamps a new one, so `replacedAt` fences the +1 transition against external replacement.

## Tests

- New: Team account recovers from a duration-classified monthly snapshot
- New: probe-owned token refresh (generation +1, same lineage) still recovers
- Updated: plan-matrix test now accepts any window the parser actually wrote; Go/Free still require monthly-only
- Existing replacement/429/concurrency/lease tests unchanged and passing

## Verification

- `tests/codex-cooldown-recovery.test.ts` — 20 pass (both new tests were red before the fix)
- `tests/rate-limit-reset-credits.test.ts` — 31 pass
- `tests/codex-routing.test.ts tests/codex-pool-rotation.test.ts tests/codex-main-rotation.test.ts` — 143 pass
- Full suite: 7720 pass, 8 skip, 11 fail — the 11 failures (10 provider-management validation, 1 crash-guard timeout) reproduce identically on the unmodified #955 head branch and are unrelated to this change.

### DIFF
--- a/src/codex/quota.ts
+++ b/src/codex/quota.ts
@@ -103,10 +103,17 @@ export function isCompleteCodexQuotaRecoverySnapshot(
   // Recovery still fails closed on MISSING EVIDENCE — a credits-only or windowless payload
   // carries no usage reading at all and must never clear a cooldown. What it does not do is
   // fail closed on an unfamiliar plan NAME, which only ever meant "cooled forever".
-  const required = codexQuotaWindowForPlan(plan) === "monthly"
-    ? quota.monthlyPercent
-    : quota.weeklyPercent;
-  return typeof required === "number" && Number.isFinite(required);
+  //
+  // The parser classifies windows by DURATION, not by plan name: a Team response whose
+  // primary window is explicitly monthly parses to monthlyPercent only (no secondary
+  // window exists), so requiring weeklyPercent because the plan is not go/free would
+  // strand exactly those accounts until their predicted expiry. Accept whichever window(s)
+  // the parser actually wrote; Go/Free never carry a weekly value, so monthly-only is
+  // required there.
+  if (codexQuotaWindowForPlan(plan) === "monthly") {
+    return typeof quota.monthlyPercent === "number" && Number.isFinite(quota.monthlyPercent);
+  }
+  return hasKnownQuotaValue(quota);
 }
 
 export function normalizeUsagePercent(value: unknown): number | undefined {
--- a/src/codex/routing.ts
+++ b/src/codex/routing.ts
@@ -139,6 +139,8 @@ export type CodexQuotaRecoveryProbeClaim = {
   leaseId: string;
   cooldownGeneration: number;
   credentialGeneration: number;
+  /** Claim-time `replacedAt`; unchanged after a probe-owned refresh, stamped on external replacement. */
+  credentialReplacedAt?: number;
 };
 
 export type CodexQuotaRecoveryProbeProof = {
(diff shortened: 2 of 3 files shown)


Wibias:

lidge-jun:
Carried into **#973** (stack 6/6). You found two real defects in my #955 code and both reproduce — thank you for reviewing it properly rather than rubber-stamping it.

Verified before carrying:

```console
$ bun run .tmp/probe_967.ts          # on the #955 head, BEFORE your fix
parsed        = {"monthlyPercent":12,"monthlyResetAt":1900000000}
recoverable?  = false      <-- Team monthly account could never recover
```

Your diagnosis is exact on both counts. P1: I picked the required window from the plan *name* while `parseUsageQuota()` picks it from the window *duration*, so a Team account with an explicitly-monthly primary stayed cooled until predicted expiry — the same "cooled forever" failure #915 exists to fix. P2: `getValidCodexToken()` really does advance the credential generation mid-probe, and `replacedAt` is the right discriminator, since a CAS refresh preserves it while a real replacement stamps a fresh one.

**One correction on top, which is why the carried commit is not the last word.** Your P1 remedy — accept whichever window the parser wrote — is too permissive in the other direction. A *tertiary-only* response also writes `monthlyPercent`, and it describes a different period than the weekly quota that actually gates a Team account:

```text
tertiary-only    {"monthlyPercent":7,"monthlyResetAt":1900000000}    <- not evidence
explicit-monthly {"monthlyPercent":12,"monthlyResetAt":1900000000}   <- evidence
```

Those two shapes were indistinguishable, so no amount of tightening at the predicate could separate them. #973 adds provenance at the source instead: `parseUsageQuota()` records `monthlyIsPrimaryWindow` when the value came from an explicitly-monthly **primary** window, and recovery requires it before accepting monthly-only evidence for a weekly-quota plan. Go/Free are unchanged.

An independent audit then found the flag was being dropped at three further copy sites — `setAccountQuotaFromParsed()`, `updateAccountQuota()`, and the credits-only branch — which would have made the guard decorative. All fixed and pinned; ablating any of them fails a test.

Your commit is carried with `cherry-pick -x`, authorship intact, patch-id verified identical. Closing this since it targeted the #955 branch directly and now lives in #973.

Issue #688, opened 2026-07-29
DeepSeek V4 web-search continuation drops reasoning_raw_delta and returns 502


Michael-Han0608:
### Client or integration

VS Code Codex extension

### Area

Provider adapters / hosted Web Search loop

### Summary

DeepSeek V4 thinking-mode requests can still fail after a synthetic hosted `web_search` call on opencodex v2.7.42, even though ordinary tool-call history replay is covered by #61 / #58.

The direct DeepSeek provider is correctly configured with `preserveReasoningContentModels` for `deepseek-v4-pro` and `deepseek-v4-flash`. However, the internal Web Search loop appears to lose DeepSeek's raw reasoning when it reconstructs the assistant tool-call message for the next model iteration.

The upstream rejection is:

```text
Provider error 400: The `reasoning_content` in the thinking mode must be passed back to the API.
```

opencodex surfaces this to Codex as a 502/in-stream failure, after which Codex retries the sampling request.

### Reproduction

1. Configure the built-in direct `deepseek` provider with `deepseek-v4-pro`.
2. Start a fresh Codex task through the Responses endpoint.
3. Ask for a research/current-information task that causes the synthetic hosted `web_search` tool to be invoked.
4. Let the model emit thinking/reasoning followed by the `web_search` tool call.
5. After the search result is injected, observe the next internal DeepSeek iteration.

This was observed after several normal shell/plan tool continuations had succeeded. The failure occurred specifically during the Web Search continuation, not on the first prompt.

### Expected behavior

The assistant message replayed by the Web Search loop should include the original DeepSeek `reasoning_content` together with the synthetic `tool_calls`, allowing the next thinking-mode request to continue.

### Actual behavior

The follow-up DeepSeek request is rejected with HTTP 400 because the replayed assistant tool-call message does not contain `reasoning_content`. opencodex records the request as:

```text
status=502
errorCode=upstream_server_error
terminalStatus=failed
upstreamError=Provider error 400: The `reasoning_content` in the thinking mode must be passed back to the API.
```

The Codex client then reports:

```text
codex_core::responses_retry: stream disconnected - retrying sampling request
```

### Suspected root cause

The OpenAI Chat adapter emits DeepSeek reasoning chunks as `reasoning_raw_delta`:

https://github.com/lidge-jun/opencodex/blob/v2.7.42/src/adapters/openai-chat.ts#L742-L744

[...]

PR #971, opened 2026-08-03
fix(responses): keep DeepSeek reasoning_content on tool-call continuations (#950)

### LINKED ISSUE #950: [Bug] OpenCode Go DeepSeek V4 Flash intermittently drops reasoning_content on tool-call continuation
### Client or integration

Codex App

OpenCode Go (Console Go), model `opencode-go/deepseek-v4-flash`

### OpenCodex version

2.10.0

### Endpoint or capability

`/v1/chat/completions` — replay of reasoning metadata when continuing an assistant tool-call turn.

### Current behaviour

During multi-turn Codex App tasks that use tools with DeepSeek thinking mode enabled, OpenCodex intermittently forwards a continuation without the original non-empty `reasoning_content`. OpenCode Go then rejects the request with HTTP 400.

Five matching failures were observed across two conversations: three consecutive failures in one conversation and two in another. Each failed after roughly 55–80 seconds. Retrying the same conversation can later succeed, which suggests a history transformation path such as resume, retry, compaction, or another synthetic continuation rather than a permanently invalid model configuration.

### Expected behaviour

For every assistant message that contains `tool_calls`, OpenCodex should preserve and replay the exact non-empty `reasoning_content` returned by the provider when sending subsequent tool results and conversation history. This invariant should hold across normal tool continuations, retries, resumed conversations, compaction, and synthetic/web-search loops.

OpenCodex should not invent an empty placeholder. It should retain the provider's original reasoning payload for the matching assistant tool-call turn.

1. Configure Codex App to use `opencode-go/deepseek-v4-flash` through OpenCodex 2.10.0.
2. Enable thinking mode with reasoning effort `max`.
3. Start a multi-turn task that causes the model to call tools.
4. Continue or retry the conversation after tool results; longer runs that resume or compact history reproduce the problem more readily.
5. Observe an intermittent HTTP 400 from OpenCode Go.

The request is redacted because the failure occurred in real conversations. The relevant shape is an assistant tool-call message originally returned with non-empty `reasoning_content`, followed by a later continuation in which that field is missing.

### Actual response or error

```text
stream disconnected before completion: Provider error 400: Error from provider (Console Go): Upstream request failed: [invalid_request_error] The `reasoning_content` in the thinking mode must be passed back to the API.
```

### Upstream documentation

https://api-docs.deepseek.com/guides/thinking_mode

DeepSeek documents that, when a thinking-mode response performs a tool call, the assistant's `reasoning_content` must be passed back in subsequent requests; otherwise the API returns HTTP 400.

OpenCodex already has reasoning replay handling for the ordinary OpenCode Go DeepSeek tool-call path and for the web-search loop. The remaining intermittent failure likely occurs in another history transformation path; this is a hypothesis, not a confirmed root cause.

A robust fix would:

1. Preserve the original raw reasoning value on the same assistant tool-call turn through resume, retry, compaction, and synthetic-history transformations.
2. Serialize that value for models covered by `preserveReasoningContentModels`.
3. Add an invariant check immediately before provider serialization: an assistant message with `tool_calls` that originally had non-empty reasoning must still have non-empty `reasoning_content`.
4. Avoid generating empty or fabricated reasoning values.
5. Emit privacy-safe diagnostics (path name and counters only, never reasoning text).

Suggested regression coverage:

- ordinary assistant tool call → tool result → continuation;
- retry after a failed continuation;
- resumed conversation;
- compacted history containing an assistant tool call;
- web-search or other synthetic tool loop;
- assertions that logs never contain reasoning text.

Related implementation history appears to include #78 for normal OpenCode Go DeepSeek tool-call replay and #688 for web-search raw-reasoning replay. #875 concerns the official DeepSeek native Responses route and appears to be a separate integration path.

Observed environment:

- OpenCodex 2.10.0;
- Codex App;
- `opencode-go/deepseek-v4-flash`;
- effective reasoning effort `max`;
- five identical 400 responses across two conversations;
- no secrets, conversation identifiers, or reasoning payloads included.

A local diagnostic prototype previously exercised the proposed invariant and completed focused tests, type checking, privacy checks, and the full test suite (7,658 passing, 8 skipped, 0 failing). That prototype was reverted; this issue does not depend on any local patch.

### Checks

Wibias:
## Summary

- Keeps DeepSeek thinking-mode `reasoning_content` on tool-call continuations so `opencode-go/deepseek-v4-flash` (and other `preserveReasoningContentModels` providers) no longer receives bare assistant `tool_calls` after history transformations — fixing the intermittent upstream HTTP 400 from issue #950.
- The bridge now records raw reasoning per tool-call id in a small in-process, bounded cache (64 entries / 256 KiB / 1h TTL, never logged); the openai-chat adapter re-attaches it when a tool-call turn serializes without thinking parts (compacted history, lost assistant turn, orphan-repaired tool results).
- Two more drop paths are closed: the parser attaches a `reasoning` item that arrives after its `function_call` to the owning assistant turn instead of discarding it, and the image bridge preserves `reasoning_raw_delta` in replayed assistant turns (mirroring the web-search loop from #688).

## Validation

- `bun run typecheck` — pass
- `bun test tests/deepseek-reasoning-replay-gaps.test.ts tests/bridge-raw-reasoning-hidden.test.ts tests/images/loop-reasoning-replay.test.ts tests/images/loop.test.ts tests/web-search.test.ts tests/opencode-go-deepseek.test.ts tests/deepseek-reasoning-replay.test.ts` — pass (93 tests)
- `bun run test` — not completed locally (Windows runner: suite exceeded ~50 min and the background worker was terminated; CI runs the full matrix on this head)

## Review notes

- The replay cache is scoped by conversation (`_clientThreadId`) plus call id so colliding provider ids (`call_1`) cannot leak reasoning across threads; entries are bounded (64 / 256 KiB / 1h TTL), swept on insert, and never logged or serialized. Parallel tool calls share one recorded reasoning block and are deduplicated at serialization. Adjacent but distinct: open PR #948 covers Kiro's redacted reasoning blob, not this wire.
- Regression tests were red against the pre-fix code for all four drop paths.

## Limitations

- The cache cannot recover reasoning for a call the proxy never saw in this process (e.g., after a proxy restart mid-round); the client-side reasoning item remains the primary replay source, and the cache is the fallback for history transformations.

Fixes #950

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved assistant reasoning when replaying tool calls across streaming and non-streaming responses.
  * Prevented reasoning from being incorrectly reused between separate conversations.
  * Improved recovery of reasoning for reordered, retried, orphaned, and parallel tool calls.
  * Preserved reasoning before image-generation tool calls.
  * Added safeguards for expired, oversized, or invalid cached reasoning.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -8,6 +8,7 @@ import { isCyberPolicyCode } from "../lib/errors";
 import { redactSecretString } from "../lib/redact";
 import { contentPartsToText } from "./image";
 import { identifyRoutedModel } from "./identity";
+import { peekReasoningForCall } from "../responses/reasoning-replay-cache";
 import { buildNonOpenAIToolCatalogNudgeForTools, shouldInjectNonOpenAIToolCatalogNudge } from "./tool-catalog-nudge";
 import { openRouterProviderPayload, resolveOpenRouterRouting } from "../providers/openrouter-routing";
 import {
@@ -324,7 +325,26 @@ function messagesToChatFormat(parsed: OcxParsedRequest, provider: OcxProviderCon
         if (textParts.length > 0) {
           chatMsg.content = textParts.map(p => p.text).join("");
         }
-        const reasoningContent = thinkingParts.map(p => p.thinking).join("");
+        let reasoningContent = thinkingParts.map(p => p.thinking).join("");
+        // History transformations (compaction, lost assistant turn, resumed
+        // threads) can strip the reasoning item while the tool round survives.
+        // Re-attach the reasoning the bridge recorded for these call ids so
+        // preserveReasoningContentModels providers (DeepSeek thinking mode)
+        // never receive a bare tool-call continuation (issue #950).
+        if (
+          reasoningContent.length === 0
+          && toolCalls.length > 0
+          && modelInList(provider.preserveReasoningContentModels, parsed.modelId)
+        ) {
+          const cached = toolCalls
+            .map(tc => (tc.id ? peekReasoningForCall(tc.id) : undefined))
+            .filter((text): text is string => typeof text === "string" && text.length > 0);
+          // Parallel calls share one preceding reasoning block, which is
+          // recorded under every call id — join unique texts only.
+          if (cached.length > 0) {
+            reasoningContent = [...new Set(cached)].join("\n");
+          }
+        }
         if (reasoningContent.length > 0 && modelInList(provider.preserveReasoningContentModels, parsed.modelId)) {
           chatMsg.reasoning_content = reasoningContent;
         }
@@ -382,9 +402,14 @@ function messagesToChatFormat(parsed: OcxParsedRequest, provider: OcxProviderCon
           // role:"tool" message unless an assistant tool_call with the same id immediately precedes it.
           flushPendingToolCalls();
           const name = safeToolName(msg.toolName);
+          // The orphan repair synthesizes an assistant tool call for a result
+          // whose assistant turn was lost; carry the recorded reasoning so the
+          // replayed round stays valid for thinking-mode providers (#950).
+          const cachedReasoning = toolCallId ? peekReasoningForCall(toolCallId) : undefined;
           out.push({
             role: "assistant",
             content: emptyAssistantContent(provider),
+            ...(cachedReasoning ? { reasoning_content: cachedReasoning } : {}),
             tool_calls: [{
               id: toolCallId,
               type: "function",
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -196,6 +196,9 @@ function toolResultImageChatParts(content: string | OcxContentPart[]): unknown[]
 function messagesToChatFormat(parsed: OcxParsedRequest, provider: OcxProviderConfig): unknown[] {
   const out: unknown[] = [];
   const { context, options } = parsed;
+  // Mirror the bridge's replay-cache scope (issue #950): provider call ids are
+  // not globally unique, so reasoning must not cross conversation boundaries.
+  const replayCacheScope = parsed._clientThreadId ?? "global";
 
   // 260718 dangling tool_calls hardening (devlog/_plan/260718_dangling_toolcall_hardening):
   // strict chat providers (Kimi/Moonshot) 400 when an assistant tool_call is not answered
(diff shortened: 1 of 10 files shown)


Wibias:
[GD] Addressed feedback

feedbacks:
- review_comment:3708335892
- review_comment:3708335900
- review_comment:3708335905
- review_comment:3708335909
- review_comment:3708343304
- review_comment:3708343307
- coderabbitai review (outside-diff thinking_delta finding, cr-comment:v1:2870ade7ed7705fcb7211de9)

commit: 0b73d5af

Wibias:
## [GD] Merge ready

**PR:** #971 — fix(responses): keep DeepSeek reasoning_content on tool-call continuations (#950)
**Head:** 0b73d5af → dev (mergeStateStatus: CLEAN)
**Linked issues:** #950

### Reviews

- **Humans (owners first):** none open
- **Bots:** 0 unresolved threads. CodeRabbit: 3 review rounds — every finding fixed in 31b5fb8 (orphan replay gate, TTL sweep + >= boundary, exact cache-cap/byte-cap tests, retry coverage) and b73d5af (hidden 	hinking_delta clears pending raw reasoning); Codex connector P1 (conversation scope) and P2 (preserve-model gate) fixed and threads resolved
- **Own bug + security + spec/standards:** bug-review complementary pass (silent failures / leaks / edge cases) + security-review matrix + spec/standards pass done on tip; blockers fixed: parallel-call reasoning dedupe, orphan 
easoning_content gating, conversation-scoped replay cache, TTL sweep

### Tip freshness

- Updated from dev: already current (088e4b1c base)
- Compiles/tests against tip: un run typecheck — pass; focused suites (96 tests) — pass
- Conflicts: none

### Checks

- Local/CLI: un run typecheck, focused parser/bridge/adapter/image-loop/web-search suites, un run privacy:scan — green
- Required CI: green on 0b73d5af (	est 1/4-4/4, macos, 
pm-global windows-latest, gates, label, nforce-target, changes, 
eact-doctor)
- Policy: no enforced CODEOWNERS/approval/merge-queue blockers

### Residual

- Full local un run test not completed on this machine (Windows runner too slow); CI's full test matrix covers it — green.

Ready to merge.

Wibias:
## [GD] Merging

Why it helps: this restores DeepSeek thinking-mode `reasoning_content` on tool-call continuations, so `opencode-go/deepseek-v4-flash` sessions stop hitting the intermittent upstream 400 from issue #950. The parser, bridge, and image-loop fixes keep the replay invariant across compaction, resume, orphan repairs, and synthetic loops, and the replay cache is conversation-scoped, bounded, and never logged.

PR #968, opened 2026-08-03
fix(google): map tool_choice onto functionCallingConfig

DevMello:
## Summary

The google adapter dropped a client's tool_choice: "none", "required", and a forced tool all produced a wire body identical to auto, with only a system-prompt nudge stating the contract in prose. The wire compiler already validates and compiles toolConfig.functionCallingConfig; the adapter just never built it. buildRequest now maps the parsed choice for all three google modes: "none" to NONE, "required" to ANY, a forced tool to ANY with its resolved wire name in allowedFunctionNames, and the allowedTools variant's "required" half to ANY. "auto" and absent choices stay byte-identical, the config is only sent when declarations exist, and Claude on Antigravity keeps its protocol-required VALIDATED mode over a client choice.

## Verification

- Six new tests in `tests/google-adapter.test.ts`: every mapping, dotted-alias resolution to the namespaced wire name, byte-stable auto, the no-declaration guard, and the Antigravity VALIDATED interplay.
- Probe harness: Anthropic tool_choice and chat-completions forced functions now produce the enforced wire config end to end.
- `bun run test`, `typecheck`, `lint:gui`, `privacy:scan`.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults.

## Summary by CodeRabbit

* **New Features**
  * Added Google Gemini tool-selection support for disabled, automatic, required, and named tool choices.
  * Added support for restricting tool use to specific allowed functions.

* **Bug Fixes**
  * Tool settings are no longer sent when no tools are available.
  * Corrected Claude on Antigravity behavior for disabled tool use while preserving validated and allowed-function selection.
  * Improved handling of tool-choice settings across supported request configurations.

PR #965, opened 2026-08-03
fix(catalog): custom model rows inherit provider reasoning metadata

Yuxin-Qiao:
## Summary

Fixes #962. When a provider model also has a dashboard-created `customModels` row, `gatherRoutedModels()` replaced the provider-derived catalog row without retaining its provider capability metadata. The custom row reached `buildCatalogEntries()` without `reasoningEfforts`, so `applyReasoningLevels()` synthesized the generic `low..ultra` ladder plus a `medium` default. For a local Ollama model configured with `noReasoningModels` and an empty `modelReasoningEfforts`, Codex then rejected an ephemeral `spawn_agent` call during client-side preflight: "Reasoning effort 'none' is not supported".

## Fix

In `src/codex/catalog/provider-fetch.ts`, a custom row now inherits provider capability metadata from the provider-derived row it replaces (same routed `provider/model` slug):

- reasoning ladder (`reasoningEfforts`) and default effort (`defaultReasoningEffort`)
- parallel tool calls, verbosity, reasoning-summary support, and normalized capabilities
- context window / max input tokens / input modalities when the custom row does not declare them explicitly

Explicit custom fields (display name, context window, input modalities) still win verbatim, preserving the existing user-override contract from the vision-sidecar work (#349/#344). A `noReasoningModels` custom model now yields a catalog entry with no reasoning levels and no default reasoning level.

## Reproduction

1. Configure `ollama/qwen-coder-3b` with `noReasoningModels` + empty `modelReasoningEfforts` and a matching dashboard custom row.
2. `ocx sync` previously advertised `supported_reasoning_levels: low..ultra` and `default_reasoning_level: medium`; after the fix the entry has an empty reasoning ladder and no default level.

## Tests

- New regression test in `tests/codex-catalog.test.ts` covering the issue end-to-end through `gatherRoutedModels` + `buildCatalogEntries`: empty reasoning ladder, no default level, explicit custom fields preserved.
- `bun run typecheck` passes.
- Catalog suites (`codex-catalog`, `catalog-vision-sidecar-modalities`, `vertex-catalog`, `catalog-input-modality-enum`, `codex-catalog-sync-hardening`): 157 tests, 0 fail.
- Full `bun run test`: no catalog failures; the only failures reproduce on clean `dev` in this local environment (management provider-validation HTTP tests and a crash-guard timing test), and upstream CI is green at the same `dev` base commit.

## Summary by CodeRabbit

* **Bug Fixes**
[...]

PR #961, opened 2026-08-03
feat: manage provider custom headers via PATCH and ocx provider edit --headers

Yuxin-Qiao:
## Summary

The provider management plane has no way to set or restore the `headers` field, so custom providers that require client-fingerprint headers (e.g. Agent Router via the `openai-chat`/`anthropic` adapters) fail with upstream `401 unauthorized client detected` after the block is lost — the only recovery was hand-editing `~/.opencodex/config.json`.

- `PATCH /api/providers?name=...` now accepts a `headers` object. It is shallow-merged onto the provider's existing block (adding one fingerprint header never wipes the rest); `null` or `{}` clears the whole block. Validation reuses `providerHeadersConfigError`, so sensitive headers (`Authorization` etc.), invalid names, non-string values, and CRLF injection are rejected with 400.
- `ocx provider edit` gains `--headers <json>` (and `--headers -` / `--headers '{}'` to clear). The value is parsed client-side and sent through the same PATCH route, so CLI and API validation stay identical.
- `POST /api/providers` already round-trips `headers` for custom providers, so a re-save keeps them; the issue's third ask is covered by the existing path.
- Docs updated in all locales (en/ko/ja/zh-cn/ru).

Example:

```bash
ocx provider edit AGR-OAI --headers '{"x-app":"cli","anthropic-version":"2023-06-01"}'
```

## Tests

- `bun run typecheck` — pass
- `bun test tests/cli-headless-parity.test.ts tests/cli-provider.test.ts tests/config.test.ts tests/repo-hygiene.test.ts` — 168 pass
- `bun run privacy:scan` — pass
- New regression tests: PATCH merge/clear/validation cases in `tests/management-provider-validation.test.ts`; CLI `--headers` forwarding and malformed-JSON rejection in `tests/cli-headless-parity.test.ts`.

Environment note: `tests/management-provider-validation.test.ts` cannot fully run on this machine because local Clash DNS maps `*.example.test` to 198.18.x.x benchmark addresses, which the destination policy deliberately rejects. The same pre-existing tests fail identically on clean `dev`; CI's unresolvable `.test` hosts take the allowed path.

Fixes #959

## Summary by CodeRabbit

- **New Features**
  - Added support for managing custom provider request headers through `ocx provider edit`.
  - Headers can be merged, updated, or cleared using `{}`, `-`, or `null`.
  - Provider API updates now support merged `headers` configuration.

- **Bug Fixes**
[...]

PR #973, opened 2026-08-04
stack 6/7: triage the overnight PRs and fix the #955 defects they found

lidge-jun:
## Stack

**6/6 — overnight PR triage**

Base: `codex/915-cooldown-recovery-probe` (#955)

## Summary

Nine PRs arrived overnight while this stack was in review. This layer carries the three that belong here, and records an evidence-backed disposition for the other six.

| PR | Author | Verdict |
|---|---|---|
| **#967** | @Yuxin-Qiao | two real defects **in my own #955** — carried |
| **#965** | @Yuxin-Qiao | correct fix for #962 — carried |
| **#968** | @DevMello | real independent defect — carried |
| #963 | @MarcTCruz | duplicate of #965, broader and weaker — closing |
| #966 | @Yuxin-Qiao | fifth #914 design, two falsifications survive — stays open |
| #964 | @Yuxin-Qiao | real, but its own review |
| #970 | @stephen-drew | real, out of scope here |
| #961 | @Yuxin-Qiao | feature, not a bug |
| #969 | @Wibias | CI governance, needs its own policy/security review |

All carried commits keep their authors (`cherry-pick -x`, patch-id verified identical).

## #967 found two real bugs in my #955, and both verify

**Monthly-classified snapshots were rejected.** My predicate picked the required window from the plan *name*; the parser picks it from the window *duration*. A Team account whose primary window is explicitly monthly parses to `monthlyPercent` only, so every successful fresh read was thrown away:

```console
$ bun run .tmp/probe_967.ts   # before
parsed        = {"monthlyPercent":12,"monthlyResetAt":1900000000}
recoverable?  = false      <-- cooled forever
```

That is the same failure #915 exists to fix, reintroduced for monthly-window plans — the third time this predicate has been wrong in the same direction.

**The probe's own token refresh looked like a replacement.** `getValidCodexToken()` refreshes a near-expiry token mid-probe and advances the credential generation by one; my settle required an exact match, so a valid fresh reading was discarded and recovery waited another interval. Fenced on `replacedAt`, which is preserved by a CAS refresh and stamped fresh by a real replacement.

## Where I disagreed with #967, and why

Its remedy for the first defect — accept whatever window the parser wrote — is too permissive in the other direction. A **tertiary-only** response also writes `monthlyPercent`, describes a different period, and says nothing about the weekly quota that actually gates the account. The two were literally indistinguishable:

```text
tertiary-only    {"monthlyPercent":7,"monthlyResetAt":1900000000}   -> would recover
explicit-monthly {"monthlyPercent":12,"monthlyResetAt":1900000000}  -> should recover
```

So this adds provenance at the source instead of guessing at the sink: `parseUsageQuota()` records `monthlyIsPrimaryWindow` when the value came from an explicitly-monthly **primary** window, and recovery requires it before accepting monthly-only evidence for a weekly-quota plan. Go/Free are unaffected — the monthly window governs them either way.

The flag is propagated through every copy site (`setAccountQuotaFromParsed`, `updateAccountQuota`, the credits-only and weekly-preserving branches, `parseUpstreamQuotaHeaders`) and pinned by tests, because a flag that silently fails to persist makes the guard decorative.

## Why #963 loses to #965

Both claim #962. #962 is about a custom row *replacing* a same-slug provider row, and #965 models exactly that — it inherits from the row actually being replaced, so it also retains live `/models` metadata. #963 recomputes config hints for **every** custom row including unmatched ones, cannot retain discovered metadata, and rewrites `tests/catalog-vision-sidecar-modalities.test.ts` from "no registry reasoning leaks onto an unmatched override" to expecting that leak while dropping three `fetch should not be called` guards. Changing a test that encodes a deliberate prior decision, to make a broader change pass, is what decided it.

## Why #966 stays open

It is a genuine advance on the four previously-falsified #914 designs — real Bun labels, no hostname probing, manual redirects on the pool paths. But two falsifications survive, both reproduced live: `fetchWithTransientRetry()` discards a prior 503 when a later attempt rejects, so a real account failure is recorded as neutral; and the five sidecar paths it newly classifies still use default-follow fetch, so a credential-visible 307 to a dead host reads as neutral *after* the origin has already seen the `Authorization` header. It does supersede #922.

## Verification

- `bun run test` — 7740 pass, 8 skip, 0 fail, 508 files
- red-green: ablating each guard fails its own tests (provenance 2, cache copy 1, update carry 1, #967 P1 2, #967 P2 1)

Nine pre-existing `toEqual` assertions needed updating for the new field — values unchanged, and each now states which side of the provenance distinction it is on.

## Audit

Four rounds, three FAIL. The reviewer caught the tertiary-only over-permissiveness, then two further copy sites of the same class. Every finding was reproduced at runtime before fixing.

## Summary by CodeRabbit

- **New Features**
  - Improved Google/Gemini tool-choice handling, including required, disabled, automatic, forced, and selectively allowed tools.
  - Custom model entries now inherit missing capability details from provider information while preserving explicit settings.
- **Bug Fixes**
  - Improved quota recovery accuracy by distinguishing verified monthly limits from supplementary data.
  - Quota recovery now handles eligible credential refreshes without unnecessarily failing.
  - Claude integrations on Google services now correctly manage tool declarations and validated tool-calling mode.
- **Tests**
  - Added coverage for tool selection, model capabilities, quota provenance, and credential refresh scenarios.


lidge-jun:
## Stack navigation

lidge-jun:

| Layer | PR | Contents |
|---|---|---|

PR #978, opened 2026-08-04
fix(google): send thinkingLevel for any model with an effort ladder

DevMello:
## Summary

- Selecting a reasoning effort for `gemini-3.1-pro-preview` now actually sends `generationConfig.thinkingConfig.thinkingLevel` on the wire. The catalog has advertised low/medium/high for this model since the 3.6 rollout, but the adapter hardcoded thinking to `gemini-3.5-flash` and `gemini-3.6-flash`, so the picker choice was silently ignored.
- The gate is now ladder driven: any model whose provider config advertises reasoning efforts (registry preset or user config) gets the mapped level. The old flash slice stays hardcoded so unladdered configs keep their current behavior.
- Vertex participates only when a user explicitly configures a ladder. The seed `google-vertex` entry ships none, so the Vertex freeze documented in the 260722 rollout plan is preserved. Image models are excluded because `thinkingConfig` would suppress `responseModalities`.

## Verification

- `bun test tests/google-hardening.test.ts tests/google-adapter.test.ts tests/google-antigravity-wire.test.ts tests/google-vertex-stream.test.ts` (77 pass)
- `bun run typecheck`, the lint gate, and `bun run privacy:scan`
- `bun test --isolate ./tests/` full suite green (7999 pass, 0 fail)
- New tests: pro-preview ladder emits the level and clamps `minimal` to `low`, Vertex emits only with an explicit ladder, unladdered direct flash keeps its slice, image models keep `responseModalities` under a provider-wide ladder.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Google AI models now support configured reasoning levels where available.
  * Reasoning settings automatically adapt to each model’s supported capabilities.
  * Existing image-generation support and fallback behavior remain available.

* **Documentation**
  * Clarified how Google model reasoning levels are applied, including model-specific support and exclusions.
  * Documented that configured reasoning levels are sent to eligible models.

## Review readiness checklist


lidge-jun:
I came into this review expecting to block it, and the diff changed my mind. Recording that plainly, because the initial reading was mine and it was wrong.

**The concern I started with.** The title reads "for any model with an effort ladder", which sounds like the gate is widened to arbitrary configured Google-adapter providers — the proxy sending a wire field nobody verified the endpoint accepts. That would be the same mistake as promoting unknown capability to supported.

**What the diff actually does.** `thinkingEligible` is narrower than the title suggests. It excludes `cloud-code-assist`, excludes image-capable models so the `responseModalities` fallback below still fires, keeps the Vertex freeze unless a ladder is explicitly configured, and retains the hardcoded `gemini-3.5/3.6-flash` slice so unladdered direct configs are unchanged. And the ladder source is `configuredReasoningEfforts` (`src/reasoning-effort.ts:58-64`), which returns a value only from a registry preset or an explicit user config — never a default. That is the load-bearing detail: there is no path where a model acquires a ladder implicitly.

I ran the same probe against both trees on a custom Google-adapter provider:

| Config | `dev` | this PR |
|---|---|---|
| custom provider, user-set `modelReasoningEfforts` | `undefined` | `{thinkingLevel:"high"}` |
| custom provider, no ladder | `undefined` | `undefined` |
| provider-wide `reasoningEfforts` | `undefined` | `{thinkingLevel:"high"}` |

So the behavior does widen, but only where the user has asserted a ladder for that model. That is an opt-in capability assertion, not a guess — and it is the same shape we treat as the legitimate remedy elsewhere in the catalog, so blocking it here would be inconsistent with our own position. `tests/google-hardening.test.ts` on your head: 24 pass, 0 fail, and the four new cases cover the pro-preview ladder, the Vertex opt-in, the unladdered Flash slice, and image-model exclusion. They test the boundaries, not just the happy path.

The underlying bug is real too: `gemini-3.1-pro-preview` ships a ladder at `src/providers/registry.ts:1217` and never receives the field on `dev`, so a user selecting an effort for it gets nothing on the wire.

**One thing I would like before merge**, and it is documentation rather than code. Setting `modelReasoningEfforts` now has a second effect for Google-adapter providers: it changes what goes on the wire, not only what the picker offers. That consequence is not guessable from the field name, and a user who sets a ladder to populate a menu could be surprised by a 400 from an endpoint that rejects `thinkingConfig`. A short note in the provider configuration reference — English source is enough, translations can follow — saying that a configured ladder is treated as an assertion that the model accepts `generationConfig.thinkingConfig.thinkingLevel` would close it.

Also: the failing `enforce-target` check is not your code. It is the stale-checkout mismatch where the run cannot resolve `.github/scripts/pr-quality-state.cjs`; a rebase onto current `dev` clears it. Please do not chase that failure.

Nice, careful piece of work — particularly keeping Vertex frozen and excluding the image models rather than letting `thinkingConfig` quietly suppress `responseModalities`.

Ingwannu (changes requested):
The core fix is valuable and the current eligibility guard is appropriately narrow, but the latest head still needs the two follow-ups identified on the rebased diff.

1. The documentation currently describes only `modelReasoningEfforts` as the capability assertion. `configuredReasoningEfforts()` also returns provider-wide `reasoningEfforts`, and that path now emits `generationConfig.thinkingConfig.thinkingLevel` for eligible non-image Google models. Please document both ladder sources.
2. “except Cloud Code Assist” is misleading: CCA does not emit this exact `generationConfig` field, but it does consume the selected effort through `resolveAntigravityEffortWireModel()` and its envelope-specific wire-model path. Please describe that distinction instead of implying the effort is omitted.
3. Add the missing positive regression for a non-image Google model with provider-wide `reasoningEfforts`. The current provider-wide test covers only the image-model exclusion, so it would still pass if the newly widened non-image provider-wide path stopped working.

No additional runtime blocker found in the adapter logic. Keeping changes requested until the contract text and provider-wide positive case match the actual implementation, then exact-head CI can make the final merge decision.

lidge-jun:
Thank you — this is landing as **#1137** in the stacked campaign, with your `src/adapters/google.ts` change adopted as authored and your commits cherry-picked with your authorship intact.

Your runtime gate was already correct; what I added on top is the contract around it. The docs now state that **both** `reasoningEfforts` and `modelReasoningEfforts` assert capability and that CCA uses its envelope-specific path, and I added the provider-wide non-image positive case that the suite was missing — the existing coverage was image-adjacent, so a regression in the general path could have slipped through.

Verified: 129 pass / 0 fail across the affected suites; full `bun run test` exit 0; typecheck and privacy scan clean.

Leaving this PR open for you.

lidge-jun:

Your commits are in `dev` with **your authorship intact** (`devmello`): `fix(google): send thinkingLevel for any model with an effort ladder` and `docs: note google effort ladders assert thinkingLevel support`, carried by #1137 (merge commit `e35faa702`).

Closing because the work shipped.

Your runtime gate went in unchanged. What was added on top is the contract around it: the docs now state that **both** `reasoningEfforts` and `modelReasoningEfforts` assert capability and that CCA uses its envelope-specific path, plus the provider-wide non-image positive case the suite was missing — the existing coverage was image-adjacent, so a regression in the general path could have slipped through.

PR #980, opened 2026-08-04
stack 7/7: NIM vision classification, service repair, and the qwen3.8-max rename

lidge-jun:
## Stack

**7/7 — the two real-but-off-theme contributor bugs, reconstructed**

Base: `codex/stack6-overnight-triage` (#973)

Layer 1 (#951) is merged. The overnight triage left #964 and #970 open as "real, independent, own review track". This layer is that track, plus a Qwen model rename the maintainer asked for.

Plan and evidence: `devlog/_plan/260804_stack7_service_vision/` — four adversarial audit rounds are recorded in `001`–`003`, because three of them found real defects **in my own designs**.

## #956 — NIM vision classification, and why #964 could not be carried

The `nvidia` entry declared no `noVisionModels`, so `planVisionSidecar` never fired and the catalog never advertised image input. Text-only NIM models either received image parts they cannot read, or had attachments blocked client-side.

#964 proposed a ~64-id text-only list. **Six of its entries are natively image-capable per NVIDIA's own documentation:**

| Id | #964 | NVIDIA |
|---|---|---|
| `thinkingmachines/inkling` | text-only | text, **image**, audio |
| `minimaxai/minimax-m3` | text-only | Text, **Image**, Video |
| `moonshotai/kimi-k2.6` | text-only | text, **image**, video |
| `moonshotai/kimi-k2.5` | text-only | text, **image**, video |
| `stepfun-ai/step-3.7-flash` | text-only | text + image (VLM) |
| `mistralai/mistral-medium-3.5-128b` | text-only | text + image |

Listing a vision model there is a **silent** defect: the model could read the image, but the proxy substitutes another model's text description. No error, worse answers, extra cost. Issue #956's own body carries two of the same errors, so reporter and author shared the premise — this is not a lapse by @Yuxin-Qiao, it is what an unaudited list does.

So every id was verified individually (`011_nim_id_audit.md`): **26 confirmed text-only ship, 6 moved to the vision list, 32 dropped** for having no current NVIDIA page. The dropped set includes `nvidia/nemotron-nano-3-30b-a3b`, a reversed-name typo of a real id the same list also spells correctly.

The 16 vision ids also get explicit `modelInputModalities`. Removing them from `noVisionModels` is not enough — the catalog advertises image input only for list members, so they would be published as text-only and the app would block attachments before the native path ran.

**What this does not fix.** Unknown ids are left unclassified deliberately. NIM publishes no modality metadata and `shouldExposeRoutedModel` filters only media-generation names, so embeddings, rerankers, guards and OCR endpoints reach the same code path — an unknown id carries no signal separating them from a text-only chat model. Two earlier designs of mine tried to default unknown ids and were falsified at the audit gate; the third claimed knowledge the data does not contain. A model NVIDIA ships after this snapshot still needs classifying by hand.

## #970 — repair the service instead of re-registering it

`ocx update` stops the proxy, then brought the service back with `ocx service install`. The Windows scheduler installer always reaches `schtasks /create`, which needs elevation the updater does not have — so a normal update stopped a working proxy and could not restore its service.

`repairService()` already existed here, so the fix is small. The safety question the PR did not answer: repair throws when the service is not installed, and the update runs *after* `ocx stop`. Verified across all three platforms — stop never deregisters (macOS unloads the plist, Windows calls `/end`, Linux calls `systemctl stop`; deletion lives only in the uninstall paths).

Two things a plain argv swap would have missed:

- **The Windows GUI worker skipped the refresh entirely**, because its own comment said `/create` would UAC-fail. Repair never calls `/create`, so that reason is gone — the skip is now narrowed to callers still passing install argv. Without this the dashboard update, the most common Windows path, keeps the bug while the CLI gets fixed.
- **`bin/ocx.mjs` infers service presence from a possibly-stale `service-state.json`.** Repair correctly refuses that case, but its thrown `Error` is indistinguishable from any other failure there, so message-matching was unimplementable and a blanket install-on-failure would resurrect the elevation prompt. It now reads `startup.serviceInstalled` from the `status --json` subprocess it already spawns — the file is plain Node ESM and cannot import `diagnoseService()`.

## `qwen3.8-max-preview` → `qwen3.8-max`

Qwen3.8-Max is stable and Alibaba documents the preview endpoint as liable to be taken offline. No alias is added: a config naming the old id still routes, it just stops carrying capability metadata keyed to a retiring id.

Pricing moves off a Routeway reseller proxy to Qwen's published **$2 / $6** — the exit condition the old overlay's own comment named. Two caveats stay in the source string: the figure is Qwen's release announcement rather than an Alibaba Model Studio billing row (which has no 3.8 entry yet), and no cache rate is published anywhere, so both cache fields are `0` rather than inheriting the reseller's `0.15`.

## Verification

- `bun x tsc --noEmit` clean; `bun run privacy:scan` passed
- **`bun run test`: 7753 pass / 8 skip / 0 fail across 508 files**
- Red-green on every new guard: reintroducing #964's kimi-k2.5 entry fails 2 tests, dropping the NIM modalities map fails 3, removing `noVisionModels` fails 8, restoring the unconditional Windows skip fails the GUI-repair guard, dropping any Qwen metadata key fails the rename-survival guard.

One ablation caught a test of mine that passed for the wrong reason: the Windows guard ran on macOS, so the branch never executed and removing the fix changed nothing. It needed a platform seam before it could fail honestly.

## Closing #964 and #970

Both close once this is green, with comments naming these commits. @Yuxin-Qiao found a real bug and @stephen-drew diagnosed the elevation problem correctly; in both cases the finding was right and the implementation needed reworking. Reopening either is one click.

## Summary by CodeRabbit

* **New Features**
  * Added verified NVIDIA NIM vision support, including native image handling and vision fallback behavior for text-only models.
  * Renamed Alibaba’s Qwen model to `qwen3.8-max` and updated its published pricing.

* **Improvements**
  * Updates now refresh existing background services with `ocx service repair`, preserving configuration and improving Windows support.
  * Added clearer recovery guidance when service refreshes fail.

* **Tests**
  * Expanded coverage for vision routing, service repair, model migration, and pricing.


lidge-jun:

| Layer | PR | Contents |
|---|---|---|

PR #985, opened 2026-08-04
feat(responses): deliver structured output to routed openai-chat models

DevMello:
## Summary

- Responses `text.format` was reduced to a boolean and discarded, so structured output (Codex `--output-schema`) silently never reached routed models, and the chat surface returned HTTP 400 for `response_format` on routed openai-chat models.
- The format object now rides the parsed request and the openai-chat adapter emits chat `response_format`, the exact inverse of the existing inbound mapping. The 400 guard is removed, routed compaction strips the format so summaries stay prose, and the native Responses passthrough is untouched.

## Verification

- `bun run typecheck`, the lint gate, `bun run test`, and `bun run privacy:scan`
- New tests: parser carry, adapter re-nesting (json_object, json_schema, name default, schema-less), byte-identical round trip on the chat surface, /v1/responses to routed chat wire, compaction drops the format, Kiro rejection pinned.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added structured-output support for routed OpenAI Chat requests using JSON Object and JSON Schema formats.
  * Preserved schema names, descriptions, schemas, and strictness when translating between Responses and Chat Completions formats.

* **Bug Fixes**
  * Unsupported backends now return their appropriate capability errors.
  * Compaction requests no longer inherit structured-output formatting.
  * Requests with internal text settings are no longer rejected unnecessarily.

* **Documentation**
  * Documented structured-output forwarding and backend compatibility.

## Review readiness checklist

### DIFF
--- a/docs-site/src/content/docs/reference/proxy-formats.md
+++ b/docs-site/src/content/docs/reference/proxy-formats.md
@@ -138,6 +138,12 @@ non-empty `messages` array. It translates system, user, assistant, and tool mess
 Responses items; translates function tools, tool choice, images, reasoning effort, and supported
 response formats; runs the normal Responses routing pipeline; then translates the result back.
 
+Structured output is part of that translation: `response_format` with `json_object` or
+`json_schema` is forwarded to routed `openai-chat` models as-is, and rides the raw request body
+on native Responses routes. A backend without structured-output support returns its own error
+instead of the proxy rejecting the request locally. The same applies to `text.format` on
+`POST /v1/responses` when the model routes to an `openai-chat` provider.
+
 Non-streaming output has `object: "chat.completion"`. Streaming output uses SSE objects with
 `object: "chat.completion.chunk"`, choice deltas, a terminal choice with `finish_reason`, and
 `data: [DONE]`. Tool-call and usage information are translated back where the source events carry
--- a/docs-site/src/content/docs/reference/proxy-formats.md
+++ b/docs-site/src/content/docs/reference/proxy-formats.md
@@ -139,10 +139,11 @@ Responses items; translates function tools, tool choice, images, reasoning effor
 response formats; runs the normal Responses routing pipeline; then translates the result back.
 
 Structured output is part of that translation: `response_format` with `json_object` or
-`json_schema` is forwarded to routed `openai-chat` models as-is, and rides the raw request body
-on native Responses routes. A backend without structured-output support returns its own error
-instead of the proxy rejecting the request locally. The same applies to `text.format` on
-`POST /v1/responses` when the model routes to an `openai-chat` provider.
+`json_schema` is forwarded to routed `openai-chat` models. On `POST /v1/responses` the
+equivalent request field is `text.format`: native Responses routes preserve it in the raw
+Responses body, and it is translated to `response_format` when the model routes to an
+`openai-chat` provider. A backend without structured-output support returns its own error
+instead of the proxy rejecting the request locally.
 
 Non-streaming output has `object: "chat.completion"`. Streaming output uses SSE objects with
 `object: "chat.completion.chunk"`, choice deltas, a terminal choice with `finish_reason`, and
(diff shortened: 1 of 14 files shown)


Ingwannu (changes requested):
This is valuable and the translation shape is correct, but the branch must be rebased before it is mergeable under the current repository gates.

I reviewed head `f717017a`. Preserving Responses `text.format`, re-nesting it as Chat Completions `response_format`, leaving native Responses passthrough on `_rawBody`, and stripping all structured-output state from routed compaction are the correct boundaries. Letting an incompatible chat backend return its own capability error is also preferable to OpenCodex silently returning unconstrained prose. I found no protocol blocker in the patch itself.

I built a local merge result against current `dev` (`a5b37827`) and verified:

- six focused parser/adapter/endpoint/compaction/Kiro suites — 192 passed, 0 failed

The blocking issue is branch/gate state: this head is 570 commits behind current `dev`. Its `enforce-target` job fails because the old checkout does not contain `.github/scripts/pr-quality-state.cjs`, and no required exact-head `ci` run is attached. Please rebase the contributor branch onto current `dev` (do not merge `dev` into it), resolve any review-readiness checklist the updated gate adds, and rerun CI on the new exact head. Once that is green, I consider this worth merging.

Ingwannu (changes requested):
Confirmed that the branch is now rebased onto current `dev`, so the old freshness-only blocker is resolved. The current head is not merge-ready yet because the latest review found substantive exact-head issues:

- Preserve a `json_schema` format even when its optional `schema` member is absent; conditionally include `schema` instead of dropping the whole format. Add the schema-less round-trip regression.
- Remove the duplicate `const json` declarations in both `tests/responses-compaction-routing.test.ts` and `tests/server-kiro-completion-e2e.test.ts`; these are same-scope parse/type failures, not style nits.
- Make both docs distinguish the native Responses `text.format` contract from routed Chat Completions `response_format`, including that Chat Completions input is converted rather than forwarded as a native Responses field.

The underlying translation boundary remains valuable and correctly placed. Keeping changes requested until those findings and exact-head CI are green.

lidge-jun:
Thank you — this is landing as **#1137** in the stacked campaign. Most of your commits were cherry-picked with your authorship intact; the one correction I made is co-authored with you.

Your architecture was right and I kept it as designed: carry the parsed format on `options.textFormat`, keep `_structuredOutput` for the web-search sidecar, and re-nest the flattened Responses fields under `json_schema` as the inverse of `responseFormatToText`. The comment explaining why `response_format` is forwarded unconditionally rather than capability-gated is a good one and it survived into the stack verbatim.

**One correction.** The `json_schema` branch was guarded by `textFormat.schema !== undefined`, so a schema-less `json_schema` fell through and no `response_format` was sent at all — silently reproducing the original defect on a narrower input. I moved the guard from the branch condition onto the `schema` member so the format is still emitted, and added a regression test asserting `response_format.type === "json_schema"` when no schema is supplied.

Your #978 landed in the same PR — see the comment there.

Verified: 129 pass / 0 fail across the openai-chat, parser, chat-completions and google suites; full `bun run test` exit 0; typecheck and privacy scan clean; docs build 221 pages.

Leaving this PR open for you.

lidge-jun:

Most of your commits are in `dev` with **your authorship intact** (`devmello`): `feat(responses): carry text.format into parsed request options`, `feat(openai-chat): map text.format onto response_format`, `feat(chat): forward response_format to routed openai-chat models`, `fix(responses): strip all structured-output traces from routed compaction`, and the docs commit. Carried by #1137 (merge commit `e35faa702`).

Closing because the work shipped, not because it was rejected.

Your architecture went in as designed: carry the parsed format on `options.textFormat`, keep `_structuredOutput` for the web-search sidecar, and re-nest the flattened Responses fields under `json_schema` as the inverse of `responseFormatToText`. The comment explaining why `response_format` is forwarded unconditionally rather than capability-gated survived verbatim.

One correction rides on top, co-authored with you: the `json_schema` branch was guarded by `textFormat.schema !== undefined`, so a schema-less `json_schema` sent no `response_format` at all — silently reproducing the original defect on a narrower input. The guard moved onto the `schema` member, with a regression test.

PR #990, opened 2026-08-04
[WRONG BRANCH] fix(providers): un-obfuscate GitHub Copilot streamed tool-call arguments

### LINKED ISSUE #989: GitHub Copilot streamed tool-call arguments are obfuscated, stalling agentic turns in Codex clients
## Summary

GitHub Copilot's `/responses` wire is not protocol-compatible with Responses clients for the `vscode-chat` integration:

- Streamed `response.function_call_arguments.delta` events carry ciphertext `delta` plus an `obfuscation` field.
- Reasoning items ship as `encrypted_content` (GitHub's own scheme) with no plaintext summary; clients cannot decrypt them and stay stuck on "thinking".
- The response id and every item id are re-encrypted per event, so the terminal `response.completed` payload disagrees with the streamed events (`call_id` is the only stable field).

## Impact

Any Responses-only Copilot model (`gpt-5.6-luna`, `gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.5`, `gpt-5.3-codex` — see #748) stalls agentic turns in Codex clients: reasoning never finishes and tool calls are unreadable. Plain text works.

## Fix

Provider-scoped SSE rewrite for `github-copilot` (PR #990): drop ciphertext deltas and re-emit plaintext arguments before `done`, neutralize encrypted reasoning items with an empty plaintext summary, and pin response/item ids to first-seen values.

Simon-Opopeee:
## What

GitHub Copilot's `/responses` wire is not compatible with Responses clients for the `vscode-chat` integration. Agentic turns on Responses-only Copilot models (`gpt-5.6-luna`, `gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.5`, `gpt-5.3-codex`) stall in Codex clients: the app stays stuck on "thinking" and tool calls never complete. Fixes #989.

## Root cause (three layers)

1. Function-call argument deltas are obfuscated (ciphertext `delta` + `obfuscation` field); the `.done` event carries the plaintext `arguments`.
2. Reasoning items ship as GitHub-encrypted `encrypted_content` with no plaintext summary; clients cannot decrypt them.
3. The response id and every item id are re-encrypted per event, so the terminal `completed` payload disagrees with the streamed events (`call_id` is the only stable field).

## Fix

New `createGithubCopilotObfuscationRewrite()` in `src/server/sse-payload-rewrite.ts`, wired into the passthrough `payloadRewrites` composition when `route.providerName === "github-copilot"` (applies to all Copilot models on both relay paths):

- Obfuscated deltas become empty deltas; the plaintext `arguments` are re-emitted as one delta immediately before `.done`, keeping the stream protocol-conformant.
- Reasoning items with `encrypted_content` are replaced with a canonical empty plaintext summary (in `output_item` events and the terminal `completed` payload).
- Response id and item ids are pinned to their first-seen values across `created`, `output_item`, and `completed` events.
- The `obfuscation` metadata is stripped from every event; clean events pass through byte-identical.

## Tests

Added to `tests/sse-payload-rewrite.test.ts`: ciphertext deltas → empty deltas + plaintext delta before `.done`; reasoning neutralization; response/item id pinning; clean events pass through unchanged. `bun test tests/sse-payload-rewrite.test.ts` and `bun x tsc --noEmit` pass.

### DIFF
--- a/src/server/responses/core.ts
+++ b/src/server/responses/core.ts
@@ -155,7 +155,11 @@ import {
   imageGenToolCallAliases,
   restoreImageGenCallsInJson,
 } from "../responses-image-gen-repair";
-import { composeSsePayloadRewrites, relaySseWithPayloadRewrite } from "../sse-payload-rewrite";
+import {
+  composeSsePayloadRewrites,
+  createGithubCopilotObfuscationRewrite,
+  relaySseWithPayloadRewrite,
+} from "../sse-payload-rewrite";
 import type { EffectiveSubagentRoster, SpawnAgentSurface } from "../../codex/catalog";
 
 import { buildToolBridgeMaps, collabSurface, injectDeveloperMessage, multiAgentGuidanceText } from "./collaboration";
@@ -1764,6 +1768,9 @@ async function handleResponsesInner(
         createImageGenCallRestoreRewrite(imageGenCallAliases),
         hasResponsesItemIdRepair(repairConfig)
           ? createResponsesItemIdPayloadRewrite(repairConfig!, translatorBudget)
+          : undefined,
+        route.providerName === "github-copilot"
+          ? createGithubCopilotObfuscationRewrite()
           : undefined,
       ].filter((rewrite): rewrite is NonNullable<typeof rewrite> => rewrite !== undefined);
       // #864: win32 rewrite traffic must never enter the tee()+JS-pull chain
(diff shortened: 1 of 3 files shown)


Simon-Opopeee:
Closing: wrong base (main). Will reopen a single clean PR against dev.

PR #997, opened 2026-08-04
fix(test): isolate usage-log fixtures from the real OpenCodex home

Yuxin-Qiao:
## Summary

- `tests/management-api-logs-metrics.test.ts` exercises `GET /api/logs` display metrics through `addRequestLog`, which persists every row to `usage.jsonl`. Unlike its sibling suites that also persist (`usage-failure-persistence.test.ts`, `api-key-attribution.test.ts`), it did not redirect `OPENCODEX_HOME` itself and relied entirely on the `bunfig.toml` preload (`tests/preload.ts`).
- The preload only loads when Bun discovers `bunfig.toml` from the current directory. Running `bun test <absolute path>` from outside the repo (agent/IDE pattern) skips it, and the fixture rows (`ok-a`/`ok-b`/`fail`, `no-such-model`, `unpriced-model`, `combo/my-combo`, ...) are appended to the developer's real `~/.opencodex/usage.jsonl`. The dashboard Usage page then renders never-used models with real token counts.
- This change adds a per-test scratch `OPENCODEX_HOME` (mkdtemp, restored and cleaned in `afterEach`), matching the established pattern in the two sibling suites.

## Reproduction before the fix

```console
$ cd ~ && bun test /path/to/repo/tests/management-api-logs-metrics.test.ts
$ rg '"requestId":"ok-a"' ~/.opencodex/usage.jsonl   # fixture row now in the real log
```

## Verification

- Reproduced against a scratch `OPENCODEX_HOME`: 11 fixture rows were written before the fix.
- After the fix, the same foreign-cwd invocation leaves the real `usage.jsonl` untouched.
- `bun run typecheck` and `bun run privacy:scan` pass.
- Focused suites (usage-summary, usage-log, request-log, usage-cost, management-api-logs-metrics) pass: 147 tests; this file is 9/9 in the full suite.
- Full suite on this branch is green except 10 `management-provider-validation.test.ts` cases that fail only in this dev environment because `chatgpt.com` resolves to a non-global address (write-time DNS/SSRF gate returns 400); upstream CI for `dev` is green.

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation by using temporary configuration directories for each test.
  * Added cleanup and environment restoration after tests complete.

## Review readiness checklist

### DIFF
--- a/tests/management-api-logs-metrics.test.ts
+++ b/tests/management-api-logs-metrics.test.ts
@@ -1,8 +1,9 @@
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
-import { mkdtempSync, rmSync } from "node:fs";
+import { existsSync, mkdtempSync, readFileSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { handleManagementAPI } from "../src/server/management-api";
+import { usageLogPath } from "../src/usage/log";
 import {
   addRequestLog,
   clearRequestLogsForTests,
@@ -173,6 +174,34 @@ describe("GET /api/logs display metrics", () => {
     }));
     const [dto] = await readLogs();
     expect(dto!.displayMetrics.cost).toEqual({ kind: "unavailable", reason: "invalid_cache_breakdown" });
+  });
+
+  test("fixture usage rows land in the scratch home, never the default location", () => {
+    // Pins the safety property this file's isolation exists for: addRequestLog
+    // persists to usage.jsonl, so if the scratch-home hook is ever dropped (or a
+    // future test logs before it runs), a bare `bun test <file>` from outside the
+    // repo writes fixture rows into the developer's real ~/.opencodex log.
+    const requestId = "safety-pin-usage-log-target";
+    addRequestLog(baseEntry({ requestId }));
+
+    const resolvedTarget = usageLogPath();
+    expect(resolvedTarget).toBe(join(testDir, "usage.jsonl"));
+    expect(readFileSync(resolvedTarget, "utf-8")).toContain(requestId);
+
+    // The default location (what the resolver returns with no OPENCODEX_HOME
+    // override) must never be the write target for this suite.
+    const previousHome = process.env.OPENCODEX_HOME;
+    delete process.env.OPENCODEX_HOME;
+    try {
+      const defaultTarget = usageLogPath();
+      expect(defaultTarget).not.toBe(resolvedTarget);
+      if (existsSync(defaultTarget)) {
+        expect(readFileSync(defaultTarget, "utf-8")).not.toContain(requestId);
+      }
+    } finally {
+      if (previousHome === undefined) delete process.env.OPENCODEX_HOME;
+      else process.env.OPENCODEX_HOME = previousHome;
+    }
   });
 });
 import { ManagementRequest as Request } from "./helpers/management-auth";
--- a/tests/management-api-logs-metrics.test.ts
+++ b/tests/management-api-logs-metrics.test.ts
@@ -1,4 +1,7 @@
-import { afterEach, describe, expect, test } from "bun:test";
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import { handleManagementAPI } from "../src/server/management-api";
 import {
   addRequestLog,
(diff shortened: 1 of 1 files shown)


lidge-jun:
Still wanted — the fixture isolation is correct and the setup/teardown restores the environment properly. It has drifted far behind `dev` (~142 commits); please rebase so CI can rerun on current code. Low conflict risk expected; after a green run it is ready for maintainer review.

Yuxin-Qiao:
Rebased onto current `dev` (now 0 commits behind; new head `cdebe0e1`). `enforce-target`, `hygiene`, `label`, and CodeRabbit all pass. The Cross-platform CI and React Doctor runs are queued awaiting approval on the new head.

lidge-jun:
This fixes a genuine hazard and the fix itself is right. Worth being explicit about why it matters, because the diff looks small: `addRequestLog` persists to `usage.jsonl`, and `getConfigDir()` resolves to the real `~/.opencodex` whenever `OPENCODEX_HOME` is unset (`src/config.ts:545-552`). So running this one test file without the repo's preload writes fixture rows into a developer's actual usage history and shows up on their Usage page. That is a bad surprise, and your comment describes the mechanism accurately.

The implementation is complete for this file — per-test scratch home, prior value restored (including the `delete` when it was previously unset, which is the case people usually miss), request logs cleared, temp dir removed.

**One thing missing: nothing tests the promise.** All nine assertions in this file are about metrics; none asserts that the real home was left untouched. So the safety property this PR exists to establish is the one property the suite would not notice losing. If someone later reorders the `beforeEach`, or a new test in this file constructs its own logger before the hook runs, we are silently back to writing into `~/.opencodex` with a green suite.

Please add one test that pins it — run with `OPENCODEX_HOME` pointed at a scratch dir, exercise a path that logs, then assert no file appeared at the default location. Asserting the resolved write target is enough; it does not need to spawn anything.

Worth a look while you are in there: if other test files call `addRequestLog` without isolation, they have the same hazard and a shared helper would be better than repeating the hook. Not a blocker for this PR — just likely to be a wider problem than one file.

Small ask on an otherwise clean change; I would like it to land.

Yuxin-Qiao:
Added the pin test in `3304d5c8`: `fixture usage rows land in the scratch home, never the default location` asserts the resolved write target equals `join(testDir, "usage.jsonl")`, that the fixture requestId is present there, and that the default-resolved target (no `OPENCODEX_HOME`) is neither the write target nor contains the fixture row.

I also verified the pin actually bites: with the `beforeEach` hook neutered and writes redirected to a scratch `OPENCODEX_HOME`, the new test fails. Nothing touched the real home during that check.

On the wider hazard: I scanned every test file that calls `addRequestLog`/`appendUsageEntry`/`usageLogPath`. Only this file lacked self-isolation; `usage-failure-persistence.test.ts` and `api-key-attribution.test.ts` already isolate their own home, so there is no other live hole today. A shared helper remains a sensible follow-up, but it is not closing a gap right now.

Yuxin-Qiao:
Heads-up on the `enforce-target` failure on this head (`parseGateState is not a function`) — it is a repo-level infra mismatch, not this PR. Evidence:

- The workflow checks out trusted scripts at `base.sha` (`ref: ${{ github.event.pull_request.base.sha }}`), while the inline script itself comes from the default-branch workflow. `pull_request_target` runs the default-branch YAML, which at `main` (`v2.10.2`) calls `parseGateState` (`.github/workflows/enforce-pr-target.yml:145`), importing it at line 63.
- Current `dev` tip (`b3a1d90a`) ships `.github/scripts/pr-quality-state.cjs` that only exports `parseState` (line 184) — no `parseGateState`. Hence `TypeError: parseGateState is not a function`.
- Older dev bases still carry the export: e.g. PR #1085 (base `8ed03e72`) passed enforce-target at 11:49Z with the same gate. So every PR based on current `dev` will fail enforce-target until main's gate YAML is synced with the dev script API (or the script keeps a `parseGateState` alias).

Fix options: sync the `parseGateState -> parseState` rename into main's workflow (release-train), or keep a backward-compatible alias export in the dev script. No change on this PR side can affect the gate; the PR itself is ready for the test shards once approved.

lidge-jun:
You beat me to it, and your version is better than the one I had staged.

I had a commit ready to push onto this branch adding the assertion — then found `3304d5c8` already there. Yours checks more than mine did: I was pinning the resolved config dir, you assert the actual `usage.jsonl` path receives the row **and** that the default location does not contain the request id. That is the stronger claim, because it survives a refactor that changes how the path is resolved.

Verified on your head rebased onto current `dev`:

```
bun test tests/management-api-logs-metrics.test.ts
10 pass, 0 fail
```

And I ablated it to make sure it is not vacuous — commenting out `process.env.OPENCODEX_HOME = testDir` in `beforeEach`:

```
(fail) fixture usage rows land in the scratch home, never the default location
 9 pass, 1 fail
```

Exactly one test goes red, and it is yours. The other nine keep passing, which is the whole point — they are metrics tests and cannot see this hazard. That is a genuine regression test.

No changes requested from me. Your branch is now 2 commits, both yours, and it needs a rebase onto current `dev` plus the readiness checklist ticked; after that it is ready for a maintainer.

Thanks for turning that around quickly, and sorry for the near-collision — I should have re-fetched your head before preparing a commit rather than working from the diff I had.

lidge-jun:
Thank you — this is landing as **#1150**, with both of your commits cherry-picked under your authorship.

The detail that made this adoptable rather than merely acceptable: you assert the target path and the content explicitly, not just that the real home is left alone. A test that only stopped writing would pass silently if the isolation regressed later — yours would not.

One conflict resolved during the pick, for the record: your two commits are stacked, so applying them required keeping the superset of imports from the later commit while applying the earlier one's scratch-home `beforeEach`/`afterEach`. The final file matches your tree exactly.

Your #999 landed in the same PR — see the comment there.

Verified: `tests/management-api-logs-metrics.test.ts` 10 pass / 0 fail; full suite 9,556 pass / 0 fail; typecheck and privacy scan clean.

Leaving this PR open for you.

Yuxin-Qiao:
Closing — the fix landed as #1150 with both commits cherry-picked under my authorship, verified 10/10 on this file plus the full suite. Thanks for the adoption and the ablation check.

lidge-jun:
Confirming for the record: this is on `dev`. Both of your commits are there with **your authorship intact** (`Yuxin Qiao`), carried by #1150 (merge commit `20c5bd230`).

The explicit target-path and content assertions are what made this non-vacuous — a test that only stopped writing would pass silently if the isolation regressed later. Verified: 9,593 pass / 0 fail on the merged tree; CI green on the exact `dev` SHA.

PR #1002, opened 2026-08-04
feat(vision): make sidecar reasoning configurable

hanjianjun:
## Summary

- make the OpenAI vision sidecar reasoning effort configurable
- expose `low`, `medium`, `high`, `xhigh`, and `max` in the Dashboard and management API
- preserve the existing `low` default for backward compatibility
- include reasoning effort in the vision-description cache identity
- document the option across all shipped documentation locales

## Why

The vision sidecar previously hard-coded `reasoning.effort: "low"`, even though the CLI already accepted `ocx agent sidecar vision --reasoning`. Users of vision models such as `gpt-5.6-luna` therefore could not trade latency and cost for more detailed image analysis, and the CLI setting was silently ignored by the vision management path.

## Impact

OpenAI-backed vision descriptions now send the configured effort in the Responses request. Existing configurations behave exactly as before because the default remains `low`. Anthropic vision retains the setting in shared configuration but ignores the OpenAI-specific wire field. Cache entries are separated by reasoning effort so changing the setting cannot reuse a description produced at a different effort.

## Validation

- `bun run typecheck`
- `bun run test` — 8214 passed, 8 skipped, 0 failed
- focused vision tests — 25 passed, 0 failed after the final cache-key adjustment
- `cd gui && bun test tests` — 583 passed, 0 failed
- `cd gui && bun run lint:i18n && bun run lint && bun run build`
- `cd docs-site && ASTRO_TELEMETRY_DISABLED=1 bun run build` — 216 pages built
- `bun run privacy:scan`

## Summary by CodeRabbit

* **New Features**
  * Added Vision reasoning-effort controls with five levels: low, medium, high, xhigh, and max.
  * Added a dashboard selector that adapts to model capabilities and preserves settings.
  * Added localized labels for the new setting.

* **Bug Fixes**
  * Vision caches now account for applicable OpenAI reasoning levels.
  * Unsupported values are safely rejected, defaulted, or capped.
  * Anthropic Vision requests continue using provider-specific thinking behavior.

* **Documentation**
  * Updated configuration references and Vision sidecar examples across supported languages.

## Review readiness checklist


hanjianjun:
Review feedback addressed in 47e25402.

- **P2 persisted-config validation**: `visionSidecar.reasoning` is now sanitized at load time (invalid hand edits degrade to the default with a warning instead of breaking startup or reaching the wire), rejected at the CLI write boundary (`validateConfigCandidate`), and normalized again before constructing the vision plan.
- **P2 Anthropic cache key**: the description cache now includes reasoning effort only for the OpenAI backend, so an effort change cannot evict a valid Anthropic cached description.
- **P2 model ladder**: the Dashboard derives the reasoning picker options from the selected model's `reasoningEfforts` metadata and clamps the persisted value to the new model's highest supported rung when switching models.
- **CodeRabbit nits**: shared `VISION_REASONING_EFFORTS` constant now backs both management checks, the Russian label follows the file's `для изображений` convention, guides in all five locales describe the OpenAI Responses vs Anthropic Messages split, and the reference tables note that supported levels are upstream/model-gated.

Validation: full server suite 8217 passed / 0 failed, GUI suite 585 passed / 0 failed, typecheck, GUI lint+build, docs build (216 pages), and privacy scan all green.

hanjianjun:
CodeRabbit re-review addressed in 7468d1fb.

- The load-time degradation warning now reads the pre-sanitization copy, so an invalid persisted `visionSidecar.reasoning` actually logs the intended `falling back to low` warning (regression covered by a test asserting the warning fires).
- All five guide locales now note that supported reasoning levels depend on the selected provider/model and that the Dashboard clamps a saved level the model does not advertise.

Validation: focused config/vision tests pass; full server suite 8216 passed / 1 unrelated flaky native-profile timing test (passes in isolation, untouched by this PR); GUI 585/585, docs build, typecheck, and privacy scan green.

hanjianjun:
CodeRabbit round 3 addressed in eb017e08.

- **Security (CWE-532)**: the degradation warning now logs only the value type (`(string)`), never the persisted content — a secret-shaped hand edit can no longer leak into logs. The regression test asserts the type-only form, the fallback text, and that the raw value is absent, with `warnSpy` restored in `finally`.
- **Cache-key docs**: all five guide locales (plus the five reference pages for consistency) now state that reasoning effort is part of the description cache identity only for OpenAI keys; Anthropic keys omit it.

Validation: typecheck, settings-stream-mode (18/18 with the strengthened assertions), and the 216-page docs build all pass.

hanjianjun:
Merged upstream `dev` (b3a1d90a) into the branch (b3018a3e) to resolve the conflict and re-run CI against the latest base.

- Conflict resolution kept both sides: upstream's required `ModelInfo.namespaced` + shadow-call model options, and this PR's vision reasoning ladder (model `reasoningEfforts` metadata is still served by the API).
- Local validation on the merged tree: typecheck, GUI lint+build, docs build (216 pages), privacy scan, vision/config focused tests, GUI suite 612/612, and the full server suite (9066 passed; the one failure is the pre-existing timing-sensitive native-profile drain test that passes in isolation — unrelated to this PR).

Wibias:
Thanks @hanjianjun for the original implementation and the follow-up work here — this provided the foundation for the configurable vision-sidecar reasoning path. 🙏🏼

We’ve taken this over on the maintainer side in #1238, rebased the feature onto current `dev`, and are continuing the hardening/review there. #1238 supersedes this PR, so I’m closing #1002 in favour of the replacement.

Thank you again for the contribution.

PR #1006, opened 2026-08-04
fix(deepseek): apply bounded JSON policy on HTTP SSE for Flash

Michael-Han0608:
## Summary

`5dd965a13` introduced a DeepSeek Flash compatibility policy that forces a bounded upstream Responses body and then reframes it for the client. That policy is currently applied only when the inbound transport is WebSocket.

Codex Desktop still uses HTTP Responses/SSE by default (`websockets` is opt-in). On that default path, DeepSeek Flash can return useful output without a usable terminal event, leaving the client pending until idle timeout.

This PR applies the same bounded-upstream policy to HTTP Responses turns as well:

1. When registry policy says upstream streaming is unreliable for the selected model, force `stream: false` upstream for both HTTP and WebSocket Responses turns.
2. If the client requested `stream: true` on HTTP, reframe the completed JSON body into a minimal SSE sequence that always ends on a terminal event:
   - `response.created`
   - `response.output_item.done` for each output item
   - `response.completed` / `response.failed` / `response.incomplete`
3. Preserve the existing WebSocket path by keeping bounded JSON for `sendResponsesJsonAsEvents()`.

This keeps the historical registry field name (`modelWebsocketUpstreamStreaming`) to minimize churn; the caller now applies it beyond WebSocket-only turns.

## Why

Default path today:

```text
Codex Desktop
  -> HTTP POST /v1/responses (stream=true)
  -> OpenCodex
  -> DeepSeek POST /responses (stream=true)
  -> text/event-stream
```

After this change for DeepSeek Flash:

```text
Codex Desktop
  -> HTTP POST /v1/responses (stream=true)
  -> OpenCodex
  -> DeepSeek POST /responses (stream=false)
  -> application/json
  -> OpenCodex synthesizes terminal SSE for the client
```

## Test plan

- [x] `bun test tests/deepseek-inbound-wire.test.ts`
- [x] `bun test tests/deepseek-reasoning-replay.test.ts`
- [x] Local retest on Codex Desktop with `websockets: false`
  - tool continuation reaches final summary
  - plain-text turn completes
- [x] Local retest with `websockets: true`
  - no regression observed

## Notes

- Related to #875
- Does not claim an isolated verdict on whether `5dd965a13` alone is sufficient for pure WebSocket environments
- Scope is limited to extending the existing compatibility policy to the default HTTP Responses path

## Summary by CodeRabbit

- **New Features**
  - Added support for streaming compatible bounded responses over both HTTP SSE and WebSocket connections.
  - Converts bounded JSON responses into complete client event sequences, including creation, completion, and terminal status events.
  - Preserves error and incomplete statuses during conversion.

- **Bug Fixes**
  - Improved handling for models that do not support upstream streaming across supported transports.

### DIFF
--- a/src/providers/registry.ts
+++ b/src/providers/registry.ts
@@ -154,11 +154,12 @@ export interface ProviderRegistryEntry {
    */
   modelWireDefaults?: Record<string, ModelWireDefault>;
   /**
-   * Registry-only per-model override for the upstream request shape used behind a
-   * Codex Responses WebSocket turn. `false` keeps the client-facing WebSocket but
-   * asks the upstream Responses endpoint for bounded JSON, which the bridge then
-   * reframes as Responses events. Use only for upstreams whose streaming response
-   * can omit or indefinitely delay the terminal event.
+   * Registry-only per-model override for the upstream Responses request shape.
+   * `false` asks the upstream for a bounded JSON body instead of an open-ended
+   * event stream, then the proxy reframes that JSON into a complete Responses
+   * event sequence for the client (WebSocket frames or HTTP SSE). Use only for
+   * upstreams whose streaming response can omit or indefinitely delay the
+   * terminal event.
    */
   modelWebsocketUpstreamStreaming?: Record<string, boolean>;
   /**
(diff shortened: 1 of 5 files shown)


lidge-jun:
Superseded by #1026 (head `a3dd24741090`), which is **open and not yet merged** at the time of this comment.

Your diagnosis was right and it is the basis of the fix.

**Adopted from this PR**

- The causal mechanism: for DeepSeek V4 Flash the upstream stream can emit tool output and then never produce a usable terminal, so the proxy must request bounded (non-streaming) JSON upstream instead of relaying an SSE stream that hangs.
- The observation that the existing bounded-JSON policy was gated on `inboundTransport === "websocket"` and therefore never applied to ordinary HTTP — which is exactly the reporter's `websockets:false, bodyKind:sse` reproduction.

#1026 generalizes that into a transport-neutral `modelResponsesUpstreamStreaming` policy applied after final wire resolution.

**Deliberately not matched**

- The second JSON-to-event conversion. This repository already has that algorithm in `sendResponsesJsonAsEvents()`; #1026 shares the existing implementation rather than duplicating it, so the two paths cannot drift.
- The persistent `_clientRequestedStream` parser state.
- The missing `[DONE]` trailer — #1026 emits exactly one.
- Bypassing the SSE item-ID repair branch. Under this PR's routing the bounded-JSON path would skip that branch, which would undermine the #938 fix landing in the same series; #1026 composes with it instead.

**Status disclosure**

#1026 is rebased onto current `dev` and revalidated before merge; that revalidation is still pending as of this comment. If it fails to land, or if the merged version no longer preserves the adopted behavior, this PR is reopened.

Full equivalence reasoning: `devlog/_plan/260805_bug_stack_campaign/050_issue875_deepseek_flash_stall.md`.

PR #1008, opened 2026-08-04
feat(usage): preserve usage history past the management read window with a daily rollup sidecar

### LINKED ISSUE #1497: [Bug]: 30d and Available history totals remain bounded by the moving 64 MiB usage tail
### Client or integration

OpenCodex Dashboard

### Area

Dashboard / Usage

## Current status

This issue is **partially mitigated but not fixed**.

PR #1532 landed on `dev` and made the bounded read visible to users by exposing the observed snapshot window and warning when history is truncated. The product should therefore no longer describe the returned tail as if it were silently known to cover the selected range.

The core aggregation defect remains: `GET /api/usage` still reads the bounded management snapshot first and applies `7d`, `30d`, or `all` filtering afterward. On a high-volume installation, the loaded tail can cover much less than the requested period.

The issue stays open until date-range and all-time aggregates are complete across the retained ledger rather than merely truthful about truncation.

## Problem

`managementUsageMaxReadBytes` is 64 MiB by default. If `usage.jsonl` is larger than that, the management reader loads only the newest tail.

The range filter is applied after that read. As a result:

- `30d` can omit persisted requests that are less than 30 days old;
- `Available history` / `all` can return essentially the same bounded tail;
- cumulative totals can decrease as older rows fall out of the moving read window;
- detail rows may be bounded intentionally, but aggregate totals become incomplete too.

PR #1532 improves disclosure only. It does not change that aggregation boundary.

## Original measured example

On OpenCodex 2.11.0 with a ledger larger than 64 MiB:

| Source | Requests | Tokens | Time covered |
|---|---:|---:|---|
| Complete `usage.jsonl` | 175,818 | 23.74B | Aug 4-11 |
| Bounded Dashboard/API snapshot | 46,417 | 6.60B | roughly 39 hours |

The API reported truncation, and 129,401 requests plus 17.14B tokens inside the nominal 30-day range were not included in the aggregate.

## Root cause

The management route performs the operations in this order:

```text
readUsageSnapshotForManagement(effectiveReadLimit)
        ↓
newest bounded byte tail
        ↓
summarizeUsage(..., range, ...)
```

The desired contract requires the date/range aggregation boundary to precede, or be independent of, the bounded recent-detail read.

## Required behavior

- `7d` aggregates every retained valid request in the selected 7-day range.
- `30d` aggregates every retained valid request in the selected 30-day range.
- All-time totals do not decrease merely because the raw tail advances.
- Detail rows may remain bounded if necessary.
- Aggregate coverage is explicit when an index/projection is incomplete or rebuilding.
- The existing `snapshotWindowStart` / `snapshotWindowEnd` disclosure from #1532 remains truthful and must not be reinterpreted as a proof of complete range coverage.

## Important correctness constraint

Do not derive a `rangeFullyCovered` assertion only from the oldest loaded row timestamp.

`usage.jsonl` is appended when a request completes while a row can carry the request start time. Completion order and start-time order can differ, so the oldest timestamp in the loaded tail does not prove what timestamps may exist in the dropped prefix.

## Preferred implementation direction

Use a rebuildable indexed or incremental aggregate projection for summaries, while keeping `usage.jsonl` as canonical evidence.

Possible implementations include:

- the existing SQLite history index for range queries; or
- the daily rollup direction from #1008, once its crash-safety, invalidation, overlap, deduplication, and disabled-mode boundaries are correct.

Raising `managementUsageMaxReadBytes` is only a temporary mitigation.

## Close condition

Close when 7d/30d aggregates cover all retained in-range requests and all-time totals remain monotonic across raw-tail rotation/truncation and restart.

## Related

- #1532 - merged truncation/window disclosure mitigation
- #1008 - durable aggregate/rollup direction

lidge-jun:
## Problem

`usage.jsonl` is append-only and unbounded. The management reader caps reads at `managementUsageMaxReadBytes` (64 MiB default) plus a 200k-entry cap, so once the file outgrows the window the oldest days silently vanish from every `/api/usage` consumer. This happened in production on 2026-08-04: a 157 MB / 380k-row log lost 11 of 30 days in the 30d view (`historyTruncated: true`, `truncatedPrefixBytes: 97.5 MB`). Raising the caps only defers the loss and slows every read.

## Design

Fold rows that leave the read window into a daily-aggregate rollup sidecar, and make the summary reader merge **rollup (old days) + raw tail (recent days)**:

- **Files:** `usage-rollup.jsonl` (append-only aggregates) + `usage-rollup-meta.json` (lineage/fingerprint/throttle), both 0600 and ownership-registered. The raw log is never truncated — the rollup is a rebuildable derived cache.
- **Commit-row authority:** each fold appends group rows first and a `commit` row last (one fsync). A segment is visible only when its commit row validates by `attemptId` + `rowCount` + `payloadDigest`; the effective cutline and the raw-tail start both derive from the same committed segments, so there is no crash window where data double-counts. Partial appends are invisible garbage; the append boundary is `ftruncate`-repaired before a retry.
- **Stability watermark:** only days ≥ 9 local days old fold, so the whole 7d window (plus `requests7d` in the API-key view) stays raw-exact. 30d includes rolled-up days at day granularity (the one boundary day is included whole, ≤ ~24h overcount — documented).
- **Cost:** long-context tiers depend on per-request input tokens, so the rollup stores fold-time cost sums; a wide price fingerprint (generated jawcode table fingerprint + expected-price overlays + priority multipliers + context tiers + a hand-bumped semantics version) forces a full rebuild on any pricing change, preserving display-time retroactivity by rebuild.
- **Validity is synchronous:** `readRollupSnapshot()` verifies version/lineage/fingerprint before returning; any mismatch serves the legacy raw-tail path while the background rebuild proceeds. `usageRollupEnabled: false` restores today's behavior entirely.

Design provenance: `devlog/_plan/260804_usage_rollup_preservation/` — a prior-art survey (Kafka segment compaction, SQLite WAL ordering, logrotate copytruncate pitfalls, TSDB downsampling, high-water-mark patterns; Tier-2 source-verified) plus a 4-round independent design audit whose blockers (crash protocol, distinct-request exactness domain, `requests7d` ownership, fingerprint completeness, same-lineage mutation) are each closed in the decade docs.

## Tests

- `tests/usage-rollup.test.ts` (12): fold aggregates vs hand-computed fixtures, crash injection at every boundary (mid-append truncation, abandoned-attempt + retry collision, rowCount/digest mismatch, missing meta), lineage/fingerprint/size rebuild triggers, min-age watermark, mid-line cutline rejection.
- `tests/usage-rollup-merge.test.ts` (6): property test `summarize(fold(prefix) ⊕ tail) === summarize(full raw)` for range `all` (totals, days, models, providers, cost to 1e-9), boundary-day additivity, surface predicates, 7d-never-touches-rollup + 30d day-grain, route cache invalidation on fold advance + flag-off legacy path, API-key totals across the fold boundary.
- Full suite: **8230 pass / 10 skip / 0 fail** (531 files) on Linux; typecheck, GUI lint, and privacy scan green.

## Real-data validation

Against a copy of the 157 MB production log (381,096 rows): the merged all-range summary equals the unbounded full-raw parse — **zero diffs** across summary totals, all 39 day rows, all 90 model rows, and all providers (cost delta 2.5e-13 USD, float ordering). `truncatedPrefixBytes: 0` with the rollup caught up. Fold of the 110 MB prefix: 1.7 s once; subsequent merged reads parse only the ~104k-row tail (378 ms vs full-file parses today).

## Docs

`reference/configuration/server.md` gains the `usageRollupEnabled` row; `reference/management-api.md` documents the `GET /api/usage` rollup semantics (locale files intentionally untouched pending translation sync).

## Follow-ups (out of scope)

Raw-file truncation/archival policy (needs writer coordination), GUI surfacing beyond existing truncation metadata.

## Summary by CodeRabbit

* **New Features**
  * Usage history is now preserved through daily rollup caches, preventing data loss during log rotation.
  * Added the `usageRollupEnabled` server configuration, enabled by default, to control usage-history preservation.

* **Bug Fixes**
  * Fixed cost fingerprinting cache-key collisions that could produce ambiguous lookups.

* **Documentation**
  * Updated management API documentation for historical and rollup-backed usage summaries.

* **Tests**
  * Added comprehensive coverage for rollup aggregation, merging, recovery, filtering, and configuration behavior.


lidge-jun:
Rebased onto current `dev` (`b3a1d90a8`) and addressed the review threads in three commits (`8e657f2a1`, `49cb22c3d`, `8d1eec899`), with an independent regression audit on each round.

**Fixed in this push:**

- **Unbounded prefix materialization (P1) / no-cap fold:** `eligibleCutline` now caps each fold segment (64 MiB default) and `foldUsagePrefix` loops segments — one commit per segment, chained `seg = prev.toOffset` — so the first fold over a multi-GB log is bounded-memory. Test 8 pins the chain.
- **Event-loop blocking during the cutline scan:** the scan yields between 1 MiB chunks.
- **Malformed-row cutline stall:** complete rows that can never carry usage (unparseable JSON, non-object, no string `requestId`, unusable timestamp) no longer freeze the cutline — they are skipped exactly as `parseUsageRange` skips them. Tests 7/7b.
- **Stale sidecar after truncate/regrow:** `readRollupSnapshot` revalidates every segment's boundary digest at read time, memoized per raw-log revision so cached hot paths pay one `fstat`. Tests 9/9b (including a same-size rewrite of an *earlier* segment). The digest-window contract (trailing 4 KiB per segment, same as fold time) is now documented at the check site.
- **`usageRollupEnabled` in API-key summaries:** `readApiKeyUsageRollup` takes the flag (cache-keyed), the route passes it, and a disabled rollup falls back to the raw tail only. Merge test 7.
- **Docs exactness:** `GET /api/usage` now states when responses are raw-tail-only lower bounds.
- **Test 2c:** restores the original digest so the `rowCount` check is exercised in isolation.

**Deferred with rationale (kept unresolved on purpose):** compaction/rewrite path analysis, partial-range day boundary, cross-model overflow dedup (redesign-scale); timezone-change rebuild (documented limitation); fold-failure observability (needs a contract, not a catch tweak); remaining minor test/devlog notes.

Verification: full suite 9088 pass / 0 fail / 8 skip (580 files), typecheck clean. Audit rounds: 2×FAIL findings (last-segment-only validation, per-call validation cost) were fixed and re-verified to PASS.

PR #1020, opened 2026-08-04
fix(anthropic): surface upstream error bodies through sidecar bridges

lidge-jun:
## Summary

Stack 01 of the bug-stack campaign (`devlog/_plan/260805_bug_stack_campaign/`).

On 2026-08-05 00:22 KST, thread-spawn requests against `anthropic/claude-opus-5` failed with 400s that the ledger recorded only as bare `Provider error 400` — the upstream reason was unrecoverable. Investigation (003 research doc + two live probes against the real Anthropic API) showed:

1. The request history shape is exonerated — the exact 297K-token parent state replays fine (200).
2. The bare message shape uniquely matches the **web-search/images bridge error path** (`src/web-search/loop.ts:460-484`), which formats the upstream body only when the adapter implements `formatErrorBody` (`src/adapters/base.ts:24`). openai-chat, google, and kiro implement it; **anthropic did not**, so the bridge discarded even well-formed Anthropic JSON error envelopes.

This PR adds `formatAnthropicErrorBody` to the anthropic adapter (JSON-envelope extraction → `type: message`, `redactSecretString`, 400-char bound — same contract as `formatOpenAIChatErrorBody`), so the next occurrence lands in the client message AND the persisted `upstreamError` with its real reason.

No retry policy, no request-log/relay production change — the existing capture chain (`relay.ts` → `request-log.ts`) already carries the formatted message; a test pins that end to end.

## Tests

- New `tests/anthropic-error-body.test.ts` (8): envelope/string/HTML/non-JSON shapes, secret redaction, real-adapter bridge integration through `runWithWebSearch` exercising the exact incident branch, persistence through `responseWithDeferredRequestLog`, body-read-failure bare-status pin.
- `bun run typecheck` — 0 errors; focused suites green; `bun run privacy:scan` — pass.
- Full `bun run test` on Linux (ssh lidge): **8232 pass / 0 fail** (baseline 8222).

## Limitations

- The original 400's underlying cause is unprovable after state eviction; this PR makes the next one diagnosable rather than guessing at a payload fix (two audit rounds rejected a refuted history-flattening hypothesis — see 003/020 docs).
- A 400 retry policy is deliberately deferred until the ledger shows real upstream 400 reasons in the wild.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Anthropic error messages by preserving useful provider details instead of showing only generic status errors.
  * Redacted sensitive information and limited error details to safe, concise output.
  * Ensured formatted provider errors are retained consistently in responses and request records.
  * Added a safe fallback when upstream error details cannot be read.

* **Documentation**
  * Added planning and triage records covering open issues, pull requests, and proposed fixes.

* **Tests**
  * Added coverage for formatting, redaction, propagation, persistence, and fallback behavior.


lidge-jun:
Rebased onto current `dev` (`a594938c57`) as part of the merge pass, and folded in a review blocker found on the rebased head.

**Blocker: the new error-body bridge could echo a credential back to the client.**

`formatAnthropicErrorBody` runs the extracted upstream text through `redactSecretString`, but that redactor only covered `key=value` forms, JSON string fields, and known token shapes (`sk-`, `gh*_`, Bearer, ARNs). An upstream 4xx commonly quotes the offending header back at the caller — `x-api-key: <value>` — and that colon form matched nothing, so a custom credential passed through unredacted into client-visible error text. This is exactly the surface this PR opens, so it is fixed here rather than deferred.

`e1edd4ef1` adds a colon-labelled rule over the same credential vocabulary, excluding `Bearer <token>` so the dedicated Bearer rule keeps the scheme readable. Regression covers a non-`sk-`-shaped custom credential and asserts ordinary diagnostics (`model: gpt-5.5`, `status: 429`) stay intact.

PR #1025, opened 2026-08-04
feat(responses): provider opt-in sparse lifecycle snapshot repair (#893)

### LINKED ISSUE #875: [Bug] DeepSeek V4 Flash Responses route stalls after tool calls

### LINKED ISSUE #893: [Bug]: Responses-compatible gateways can return sparse lifecycle snapshots that Codex clients do not commit
### Client or integration

Codex App and Codex CLI

### Area

Streaming

### Summary

Some OpenAI Responses-compatible gateways return a successful SSE stream but omit canonical fields from lifecycle snapshots. OpenCodex currently relays those snapshots unchanged, and current Codex clients may reject or fail to render the turn even though the upstream request completed with HTTP 200.

The affected payloads are structurally sparse rather than malformed JSON. Examples include lifecycle `response` objects without `output`, `parallel_tool_calls`, `tool_choice`, or `tools`; message items without `role` or `content`; `output_text` parts without `annotations`; reasoning items without `summary`; and text delta/done events without `logprobs`.

Expected: an explicit provider-local compatibility option can backfill only missing or structurally invalid canonical fields on the client-facing SSE branch. Existing upstream values must remain authoritative, and the default passthrough path must remain unchanged.

### Reproduction

1. Configure a custom provider with the `openai-responses` adapter and a Responses-compatible endpoint.
2. Have the upstream return a successful stream containing sparse lifecycle events equivalent to:

```text
event: response.created
data: {"type":"response.created","response":{"id":"resp_example","object":"response"}}

event: response.output_item.added
data: {"type":"response.output_item.added","item":{"id":"msg_example","type":"message"}}

event: response.output_text.delta
data: {"type":"response.output_text.delta","delta":"hello"}

event: response.completed
data: {"type":"response.completed","response":{"id":"resp_example","object":"response"}}
```

3. Send a short request through `/v1/responses` from Codex.
4. Observe that the upstream and proxy complete with HTTP 200, while the Codex client may show no final assistant message or fail to commit the response lifecycle.
5. Compare with the same semantic events after the missing canonical fields are backfilled; the client completes normally.

The sample above is a synthetic, redacted equivalent of the observed wire shape and contains no provider-specific data.

### Version

2.10.0 (`f9b9440c`)

### Operating system

macOS 26.5.2 (arm64)

### Provider and model

Custom OpenAI Responses-compatible gateway / GPT-compatible routed model

### Logs or error output

```shell
Upstream HTTP status: 200
Proxy terminal status: completed
Codex client result: no committed final assistant message for the sparse snapshot stream
```

None. A synthetic SSE fixture and regression tests are included in the accompanying PR.

### Redacted configuration

```json
{
  "providers": {
    "example": {
      "baseUrl": "https://gateway.example/v1",
      "responsesPath": "/responses"
    }
  }
}
```

### Suggested fix

Add a disabled-by-default provider option such as `responsesSnapshotRepair`. When enabled, apply a client-facing SSE payload rewrite that:

- backfills only missing or structurally invalid canonical fields;
- preserves every valid value supplied by the upstream;
- covers response lifecycle snapshots, output items, content parts, reasoning summary parts, and output-text logprobs;
- composes with existing image-call and item-ID rewrites;
- keeps inspection and persistence on the raw upstream bytes;
- can run inline on the existing explicit Darwin eager-relay path without reintroducing the tee/JS-pull retention shape.

### Checks

lidge-jun:
## Summary

Stack 03 of the bug-stack campaign (`devlog/_plan/260805_bug_stack_campaign/040`), stacked on #1023. Fixes #893.

Responses-compatible gateways can return sparse lifecycle snapshots — missing canonical fields and the closing events Codex needs to commit a turn, leaving clients stuck on Thinking.

What ships:

- **Block-level SSE rewrite contract** (`SseBlockRewrite`): one upstream event maps to zero-or-more client events — event injection is impossible in the old one-payload model. Implemented in both the pull relay and the eager relay, with an exactly-once `dispose()` teardown contract (composed chains dispose every child).
- **`responsesSnapshotRepair` provider opt-in** (config + management plane, canonical OpenAI pool/direct seeds):
  - field backfills (adopted from PR #928 with attribution): status/output/tools/tool_choice snapshots, message/reasoning item fields, text-part/annotations/logprobs;
  - **lifecycle injection**: `content_part.added` on the first text delta for a proven-open item, and `output_text.done` / `content_part.done` / `output_item.done` before `response.completed` — so the turn actually commits;
  - terminal output reconstruction only when the gateway omitted `output` entirely — explicit `output: []` stays authoritative; `response.failed`/`incomplete` never fabricate completed output.
- **Fail closed**: ambiguous/gapped/malformed/oversized/contradictory shapes taint the stream — no injections, no reconstruction, retained state released immediately. Identity correlation (id + type) is required before any tracked item is closed.
- JSON (non-stream) path repairs fields + canonical output after raw inspection/persistence.

Four independent review rounds folded (tee-gate skip, terminal fabrication, retention bounds/disposal, tail framing, identity correlation, wire-shape defaults, cancel race). Equivalence analysis vs #928 in the 040 doc.

## Tests

- `tests/responses-snapshot-repair.test.ts` (22): exact issue fixture → full canonical sequence + committed message, explicit-empty authority, byte-identical canonical passthrough, taint paths, identity correlation, failed/incomplete, EOF dispose, function-call streams, tail framing, canonical output.
- `tests/responses-snapshot-repair-server.test.ts` (2): opt-in repairs end-to-end through `/v1/responses`; opt-out relays unchanged.
- `bun run typecheck` 0 errors; `bun run privacy:scan` pass.
- Full `bun run test` on Linux (ssh lidge): **8276 pass / 0 fail** (baseline 8222).

## Provider docs

Five locale files document the opt-in.


lidge-jun:
Rebased onto the current stack head, with a review blocker folded in.

**Blocker: the repair did not fail closed on a contradictory `output_text.done`.**

A text-done event carrying a *different* `item_id` for a tracked `output_index` was silently ignored. The tracked item therefore stayed open, and the completed terminal went on to synthesize its own `output_text.done` and `output_item.done` — a double close derived from a stream whose identity model we had just observed to be wrong. The prior test asserted that double-close shape as intended behavior, which is what let it through.

`output_item.done` has always treated an id/type mismatch as fail-closed. `62a5ee6c7` applies the same contract to `output_text.done`: forward the block untouched, taint, and stop injecting for the rest of the stream. The test now asserts no synthetic closure and no reconstruction after the taint.

PR #1026, opened 2026-08-04
fix(deepseek): transport-neutral bounded-JSON policy with terminal SSE synthesis (#875)

### LINKED ISSUE #875: [Bug] DeepSeek V4 Flash Responses route stalls after tool calls

lidge-jun:
## Summary

Stack 04 of the bug-stack campaign (`devlog/_plan/260805_bug_stack_campaign/050`), stacked on #1025. Fixes #875.

The registry's upstream-streaming hint was WS-scoped: ordinary HTTP turns kept `stream:true` into DeepSeek's terminal-less SSE and stalled after tool calls (reporter reproduced at e44d234f0 with `websockets:false, bodyKind:sse`).

- `modelResponsesUpstreamStreaming` (renamed, transport-neutral) applies after final wire resolution for `openai-responses` adapters — DeepSeek Flash now uses bounded JSON upstream on every transport.
- Shared bounded-JSON event sequence (`src/server/responses-json-events.ts`): the WS bridge and the new HTTP synthesis emit the same canonical frames — `response.created` → one `response.output_item.done` per item → status-preserving terminal; HTTP streaming clients get SSE with exactly one `[DONE`].` Non-streaming clients and WS turns keep plain JSON (gate excludes `inboundTransport: "websocket"`).
- Supersedes PR #1006's mechanism with the contributor's core policy retained (attribution) — its duplicated event algorithm, missing `[DONE]` trailer, and bypass of the SSE item-ID repair branch are avoided.

## Tests

- Stall activation: fake upstream fails loudly if `stream:true` is ever sent (the old hang), and the client receives the full synthesized sequence with `[DONE]`; function_call id/call_id byte-identical.
- WS/HTTP serializer parity + status preservation (failed/incomplete) in `tests/responses-json-events.test.ts` and `tests/ws-endpoint.test.ts`.
- Independent review: GO-WITH-FIXES(1) folded (WS transport exclusion).
- `bun run typecheck` 0 errors; `bun run privacy:scan` pass.
- Full `bun run test` on Linux (ssh lidge): **8288 pass / 0 fail** (baseline 8222).

## Limitations

DeepSeek Flash loses progressive token delivery — the intended provider-specific reliability tradeoff. structure/04 documents the transport-neutral policy.

### DIFF
--- a/src/providers/registry.ts
+++ b/src/providers/registry.ts
@@ -160,7 +160,7 @@ export interface ProviderRegistryEntry {
    * reframes as Responses events. Use only for upstreams whose streaming response
    * can omit or indefinitely delay the terminal event.
    */
-  modelWebsocketUpstreamStreaming?: Record<string, boolean>;
+  modelResponsesUpstreamStreaming?: Record<string, boolean>;
   /**
    * Responses-API resource path for providers whose route is not `/v1/responses`.
    * Unlike `modelWireDefaults` above, this IS seeded into saved config: it describes
@@ -1143,7 +1143,7 @@ export const PROVIDER_REGISTRY: readonly ProviderRegistryEntry[] = [
     // DeepSeek's Codex Responses stream can deliver output without closing on the
     // terminal event. Keep Codex on WebSocket, but use the provider's bounded JSON
     // response upstream so the bridge can synthesize a complete WS event sequence.
-    modelWebsocketUpstreamStreaming: { "deepseek-v4-flash": false },
+    modelResponsesUpstreamStreaming: { "deepseek-v4-flash": false },
     // DeepSeek's Responses route is `POST /responses` with no `/v1` segment. Without
     // this the passthrough adapter falls back to its legacy `/v1/responses`
     // construction and the wire above can never route.
@@ -1815,15 +1815,15 @@ export function providerModelWireDefault(
   return wire !== undefined && allowedWires.has(wire) ? wire : undefined;
 }
 
-/** Resolve a registry-only upstream-streaming compatibility hint for WS turns. */
-export function providerModelWebsocketUpstreamStreaming(
+/** Resolve a registry-only upstream-streaming compatibility hint for Responses turns. */
+export function providerModelResponsesUpstreamStreaming(
   id: string,
   provider: Pick<OcxProviderConfig, "baseUrl" | "adapter"> & Partial<Pick<OcxProviderConfig, "authMode">>,
   modelId: string,
 ): boolean | undefined {
   const entry = getProviderRegistryEntry(id);
-  if (!entry?.modelWebsocketUpstreamStreaming || !providerMatchesRegistryTransport(id, provider)) return undefined;
-  return entry.modelWebsocketUpstreamStreaming[modelId.trim().toLowerCase()];
+  if (!entry?.modelResponsesUpstreamStreaming || !providerMatchesRegistryTransport(id, provider)) return undefined;
+  return entry.modelResponsesUpstreamStreaming[modelId.trim().toLowerCase()];
 }
 
 /**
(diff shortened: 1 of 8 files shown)


lidge-jun:
Rebased onto the current stack head, with a review blocker folded in.

**Blocker: this layer was not independently safe with item-id repair.**

The bounded-JSON policy answers a streaming client by synthesizing SSE from a completed JSON body, and reframes the same body into events for WS turns. Neither path goes through the SSE relay, so neither picked up the relay's item-id rewrite. The consequence is worse than a missing feature: for a provider that *has* id repair configured, enabling this reliability policy would silently **disable** it — canonical ids while streaming, raw upstream ids the moment the policy engaged. The later #938 PR happened to fix this, which meant this PR was only safe if its child landed with it.

`710492cb2` moves that normalization into this layer: `repairResponsesJsonItemIds` for whole-object rewriting, applied to the synthesized-SSE and WS-reframe paths after the raw recording. New tests prove repaired ids on both bounded-JSON paths and that a provider without repair keeps the body byte-identical. #1027 was rebased on top and now carries only its own `repairInvalidIds` work.

PR #1034, opened 2026-08-04
fix(cli): emit login start URL before polling under non-TTY

### LINKED ISSUE #1007: [Bug]: ocx account login withholds the authorization URL under non-TTY stdout, blocking the manual redirect-URL flow
### Summary

`ocx account login` produces no visible output for a long time when its stdout is consumed by another process, while redirecting the same command to a file returns the authorization URL immediately. In an automated/headless context this looks like a hang, and the natural operator reaction is to kill and retry — which discards a listener that is actually waiting for a callback.

Measured on `opencodex 2.7.43` (`@bitkyc08/opencodex`), macOS, non-interactive shell.

### Reproduction

1. From a non-interactive shell, run `ocx account login` with stdout piped into another command (for example a pager-like consumer).
2. Observe: no output for roughly 70 seconds.
3. Run the same command redirecting stdout to a file instead.
4. Observe: the authorization URL appears in the file effectively immediately.

### Expected

The authorization URL is emitted promptly regardless of whether stdout is a TTY, a pipe, or a file, so an automated caller can read it and drive the manual redirect-URL flow.

### Why this matters beyond cosmetics

The manual redirect-URL path (`ocx account code` accepting a full redirect URL) is the only workable route when the browser is on a different machine than the CLI — remote host, SSH, or a headless worker. That path begins by reading the authorization URL from `ocx account login`. When that first step appears to hang under a pipe, the flow is usually aborted, and an authorization that a user already approved can be lost because the waiting listener was killed.

Related but distinct: #131 covered the GUI path lacking a manual code/redirect paste fallback. This report is about the CLI emitting its authorization URL late under a non-TTY stdout, which blocks the very fallback #131 established.

### Additional observation — listener lifetime

In the same environment, the process waiting for the OAuth callback did not outlive the shell/session that started it. When an operator approves in a browser after that session ends, the approved authorization has nowhere to land. A documented, explicitly detachable listener lifetime — or a clear statement of how long the pending authorization remains claimable and how to resume it from a new process — would make the remote flow reliable. Happy to split this into its own issue if preferred.

### Environment

- `opencodex 2.7.43`
- macOS, non-interactive shell, browser on a separate Windows host
- Codex account pool with auto-switch enabled

ZachDreamZ:
## Summary

Fixes #1007.

`ocx account login` printed the authorization URL with `console.log`, which stays buffered when stdout is not a TTY (piped or redirected). Because the command then enters its polling loop, the URL did not reach the user until the process flushed, which made the manual redirect-URL flow look hung under non-TTY stdout.

The login-start announcement (URL, instructions, flow id / device code) is now written synchronously to stdout fd 1 before the first poll. Output text and `--json` behavior are unchanged. The writer is injectable (`stdoutImpl`), so the ordering is covered by an in-process regression test.

## Verification

- `bun test tests/cli-account.test.ts` — 65 pass, 0 fail; includes two new regression tests asserting the URL is emitted before the first poll for both OAuth and Codex logins.
- `bun run typecheck` — pass
- `bun run lint:gui` — pass
- `bun run doctor:gui:if-changed` — skip (no `gui/` changes)
- `bun run test` — 8202 pass / 6 skip / 16 fail on this machine. All 16 failures reproduce identically on unmodified `dev` HEAD (Windows `EBUSY` file-lock in the e2e isolated-home helper plus 5s timeouts while the machine was heavily loaded; the suite itself reported taking 2347s vs its normal ~210s), i.e. pre-existing environmental failures unrelated to this change.

## Checklist

- [x] Docs or release notes were updated when needed. (No docs change needed: same output text and flags as before; this only fixes when the URL reaches the user.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (Output content is unchanged — the same URL/flow id/device code, just flushed earlier. No credential is written to stdout, and the `--json` path is unchanged.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Login instructions and authorization URLs now appear immediately, including when output is piped or redirected.
  * Improved Codex and OAuth login flows with clearer device-code and flow information.
  * Login status handling now behaves consistently during timeouts and successful authentication.

### DIFF
--- a/src/cli/account-api.ts
+++ b/src/cli/account-api.ts
@@ -46,6 +46,8 @@ export interface AccountDeps {
   loadConfigImpl?: () => OcxConfig;
   stdinImpl?: AccountStdin;
   stdinTimeoutMs?: number;
+  /** Test injection for synchronous stdout writes (login start URL). */
+  stdoutImpl?: (chunk: string) => void;
   /** Test/platform injection for the official Codex login in a restricted staging home. */
   spawnCodexLoginImpl?: (codexHome: string) => NativeMainLoginChild;
   /** Legacy test seam. Production always uses the spawned child handle above. */
--- a/src/cli/account-auth.ts
+++ b/src/cli/account-auth.ts
@@ -1,3 +1,4 @@
+import { writeSync } from "node:fs";
 import {
   CliUsageError,
   printData,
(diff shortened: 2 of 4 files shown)


ZachDreamZ:
Closing as a duplicate — maintainer PR #1029 implements the same fix for #1007 with a stronger pipe-based regression test and already has green CI. Deferring to that one.

Issue #1017, opened 2026-08-04
[Bug]: Cursor adapter consistently emits invalid Codex apply_patch payloads


Vincent-HD:
### Environment

- OpenCodex: 2.10.0
- Provider: Cursor
- Model: cursor/grok-4.5
- Codex Desktop: 0.146.0
- OS: Linux
- Current config: nativeLocalExec is enabled
- The same behavior was previously observed with nativeLocalExec disabled as well

### Problem

When using cursor/grok-4.5 through OpenCodex with Codex Desktop, file edits repeatedly produce malformed Codex apply_patch calls.

Observed failures include:

- Add-file lines missing the required + prefix
- Invalid hunk/context formatting
- Patch-like text that is not valid Codex freeform patch syntax

Codex then rejects the call locally with errors such as:

    apply_patch verification failed: invalid hunk

The OpenCodex request itself completes with HTTP 200, so this does not look like a transport or filesystem failure. The failure appears to occur at the Cursor adapter/model tool-contract boundary.

### Expected behavior

The Cursor adapter should either:

1. preserve apply_patch as a native custom/freeform tool;
2. expose a provider-compatible structured edit tool; or
3. clearly report that this model/provider combination cannot reliably execute Codex apply_patch.

A model producing patch-like text that is rejected on every edit attempt makes the Cursor route effectively unusable for repository changes.

### Related

- #820 — custom/freeform tool conversion and tool-call protocol invariants
- #974 — experimental Cursor client integration

I can provide a redacted session capture with the malformed tool arguments and local rejection results if needed.

olddonkey:
I feel this is extremely useful as cursor is a good agent platform.

ZachDreamZ:
Opened a fix: #1036. It adds provider-compatible structured edit tools (\dit_file\ / \multi_edit\) to the Cursor route and converts their exact-match replacements into valid Codex apply_patch payloads server-side, so models like cursor/grok-4.5 no longer emit malformed patch text. Malformed calls are dropped with a clear error instead of being relayed for local rejection after HTTP 200.

Vincent-HD:
> I feel this is extremely useful as cursor is a good agent platform.

yes, for work i'm only using cursor and for personal work I use codex, and I would really love to ditch cursor and just have a proper Codex + vscode setup.

Cursor is getting bloated and I STILL want an IDE for daily usage

lidge-jun:
Status update so this does not look abandoned.

[...]

Issue #372, opened 2026-07-24
grok-4.5 don‘t know how to use apply_patch tool


YPQuinn:
### Client or integration

Codex App

### Area

Provider adapter

### Summary

When using Codex with OpenCodex routing, the freeform `apply_patch` tool consistently fails validation with:

```text
apply_patch verification failed: invalid patch: The first line of the patch must be '*** Begin Patch'
```

This happens even when the patch body itself starts with *** Begin Patch *** and ends with *** End Patch ***.
The failure occurs before any file is modified, so it looks like a tool-invocation / freeform-payload handling issue rather than an invalid diff.

### Reproduction

same as summary.

### Version

2.7.35

### Operating system

macOS 26.5.2

### Provider and model

xai/grok-4.5

### Logs or error output

```shell
Observed:
- 47 failed `apply_patch` custom_tool_call attempts in one session
- Every call payload already starts with `*** Begin Patch ***`
- Parser rejects it with:
  `The first line of the patch must be '*** Begin Patch'`

Example call:


{
  "type": "custom_tool_call",
  "call_id": "call-297edb36-316e-4834-9a15-ffe765307655-21",
  "name": "apply_patch",
  "input": "*** Begin Patch ***\n*** Update File: ...\n...\n*** End Patch ***"
}
```

### Screenshots and supporting files

<img width="757" height="893" alt="Image" src="https://github.com/user-attachments/assets/798f8d67-5f82-4e33-80f6-ed1a6f9886b9" />

### Redacted configuration

```json

```

### Checks

- [x] I searched existing issues and documentation.
- [x] I removed secrets, tokens, account details, request credentials, and personal data.

PR #1036, opened 2026-08-05
fix(cursor): structured edit tools convert to valid apply_patch calls (#1017)

ZachDreamZ:
## Summary

The Cursor adapter advertises Codex's `apply_patch` freeform tool (a `{"input": "*** Begin Patch ..."}` body) to Cursor models, while simultaneously rejecting Cursor-native write/delete tools so edits cannot bypass Codex approval. Cursor-trained models such as `cursor/grok-4.5` cannot emit Codex's freeform patch grammar, so every file edit produced malformed patch text (missing `+` prefixes, invalid hunks) that the Codex client rejected locally after the request had already returned HTTP 200 — making the Cursor route unusable for repository changes (#1017).

This PR adds the missing provider-compatible edit path (option 2 from the issue):

- When a Cursor request advertises the freeform `apply_patch` tool, the adapter now also advertises `edit_file` and `multi_edit` — exact-match replacement tools shaped like Cursor's native Edit/MultiEdit tools.
- Completed `edit_file` / `multi_edit` calls are converted server-side into a valid Codex apply_patch payload (`*** Begin Patch` envelope, `*** Update File:` section, `@@` hunks with `-`/`+` line prefixes) and relayed to the bridge as an `apply_patch` custom tool call, so Codex approval/sandbox policy still applies.
- Malformed structured calls (non-JSON args, missing `file_path`/`old_string`/`new_string`, empty `old_string`, invalid `edits[]`) are dropped with a clear error instead of relaying invalid patch text — the previously silent 200-then-local-rejection failure mode.
- The tool budget pins the synthetic tools alongside `apply_patch` (never truncated away), the guidance note tells models to prefer them, and the Cursor-native write/delete refusal message points to them.
- Explicit `tool_choice` selections are never widened: a forced `apply_patch` choice does not gain sibling tools.

## Verification

Commands run (Windows 11, bun 1.3.14):

- `bun run typecheck` — pass
- `bun run lint:gui` — pass
- `bun run privacy:scan` — "Privacy scan passed"
- `bun run doctor:gui:if-changed` — skip (no gui/ changes)
- `bun test tests/cursor-structured-edit.test.ts` — 16 pass / 0 fail (new regression suite)
- `bun test tests/cursor-tool-definitions.test.ts tests/cursor-tool-choice.test.ts tests/cursor-protobuf-events.test.ts tests/cursor-tool-arg-decoding.test.ts tests/cursor-request-builder.test.ts tests/cursor-tool-continuation.test.ts tests/cursor-native-exec.test.ts tests/cursor-native-exec-policy.test.ts` — 166 pass / 0 fail
- `bun test tests/cursor-live-transport.test.ts tests/cursor-live-smoke-gate.test.ts tests/cursor-hardening.test.ts tests/cursor-message-mapper.test.ts tests/cursor-tool-finalize-race.test.ts tests/cursor-desktop-exec.test.ts` — 72 pass / 0 fail

New tests cover: tool advertisement conditions (auto vs forced/allowed tool_choice, namespaced apply_patch), budget pinning, guidance wording, single/multi-line replacement payloads, deletion hunks, Cursor-style arg aliases, multi-edit payloads, malformed-call rejection, and both protobuf event paths (interactionUpdate and stateless native-exec).

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for Cursor’s structured `edit_file` and `multi_edit` tools.
  * Automatically converts valid structured edits into supported patch operations.
  * Added guidance encouraging structured editing for file changes.
* **Bug Fixes**
  * Improved validation and error reporting for malformed edit requests.
  * Structured editing now works consistently across supported tool-call flows.

## Review readiness checklist

### DIFF
--- a/src/adapters/cursor/live-transport.ts
+++ b/src/adapters/cursor/live-transport.ts
@@ -66,6 +66,7 @@ import { desktopDepsFromConfig } from "./native-exec-desktop";
 import {
   buildCursorToolDefinitions,
   cursorRequestAdvertisesApplyPatch,
+  cursorRequestAdvertisesStructuredEdits,
   cursorRequestHasShellAlias,
   cursorToolArgNormalizeSchema,
   cursorToolWireName,
@@ -544,6 +545,7 @@ class LiveCursorTransport implements CursorTransport {
       ...this.execContext,
       clientToolDefs,
       rejectNativeFileMutations: cursorRequestAdvertisesApplyPatch(request.tools, request.toolChoice),
+      structuredEditAvailable: cursorRequestAdvertisesStructuredEdits(request.tools, request.toolChoice),
     };
     const toolSchemas = new Map<string, unknown>();
     const cursorToolNameMap = new Map<string, string>();
--- a/src/adapters/cursor/live-transport.ts
+++ b/src/adapters/cursor/live-transport.ts
@@ -67,6 +67,7 @@ import {
   buildCursorToolDefinitions,
   cursorRequestAdvertisesApplyPatch,
   cursorRequestAdvertisesStructuredEdits,
+  cursorStructuredEditTools,
   cursorRequestHasShellAlias,
   cursorToolArgNormalizeSchema,
   cursorToolWireName,
@@ -547,6 +548,17 @@ class LiveCursorTransport implements CursorTransport {
       rejectNativeFileMutations: cursorRequestAdvertisesApplyPatch(request.tools, request.toolChoice),
       structuredEditAvailable: cursorRequestAdvertisesStructuredEdits(request.tools, request.toolChoice),
     };
+    // Provenance for the synthetic structured-edit tools (#1036 review): record the bare names WE
+    // actually advertised on THIS request, taken from the definitions that were really sent rather
+    // than from the name alone. A client or MCP tool legitimately called `edit_file` is in
+    // clientToolDefs too, so the discriminator is `cursorStructuredEditTools` having produced it —
+    // which is precisely what `structuredEditAvailable` already reflects.
+    const syntheticStructuredEditToolNames = new Set(
+      (this.execContext.structuredEditAvailable
+        ? cursorStructuredEditTools(request.tools, request.toolChoice)
+        : []
+      ).map(tool => tool.name),
+    );
     const toolSchemas = new Map<string, unknown>();
     const cursorToolNameMap = new Map<string, string>();
     for (const tool of cursorVisibleTools ?? []) {
@@ -573,6 +585,7 @@ class LiveCursorTransport implements CursorTransport {
         parallelToolCalls: request.parallelToolCalls,
         toolSchemas,
         cursorToolNameMap,
+        syntheticStructuredEditToolNames,
         translatorBudget: this.translatorBudget,
         contextUsage,
         ...(prepared.estimatedInputTokens !== undefined
--- a/src/adapters/cursor/native-exec-fs.ts
+++ b/src/adapters/cursor/native-exec-fs.ts
@@ -40,7 +40,7 @@ const MAX_GREP_RESULTS = 200;
 const MAX_FILE_BYTES = 1_000_000;
 
 function codexNativeMutationRefusal(operation: "write" | "delete"): string {
-  return `Cursor-native ${operation} is disabled for this Codex request because apply_patch is available. Use the apply_patch tool for file edits so Codex can approve the change, enforce sandbox policy, show diffs, and record rollout. No file was changed.`;
+  return `Cursor-native ${operation} is disabled for this Codex request because apply_patch is available. Use the structured edit tools (\`edit_file\` / \`multi_edit\`) or the \`apply_patch\` tool for file edits so Codex can approve the change, enforce sandbox policy, show diffs, and record rollout. No file was changed.`;
 }
 
 const NATIVE_LOCAL_EXEC_DISABLED =
--- a/src/adapters/cursor/native-exec-fs.ts
+++ b/src/adapters/cursor/native-exec-fs.ts
@@ -39,8 +39,11 @@ const MAX_GREP_FILES = 500;
 const MAX_GREP_RESULTS = 200;
 const MAX_FILE_BYTES = 1_000_000;
 
-function codexNativeMutationRefusal(operation: "write" | "delete"): string {
-  return `Cursor-native ${operation} is disabled for this Codex request because apply_patch is available. Use the structured edit tools (\`edit_file\` / \`multi_edit\`) or the \`apply_patch\` tool for file edits so Codex can approve the change, enforce sandbox policy, show diffs, and record rollout. No file was changed.`;
+function codexNativeMutationRefusal(operation: "write" | "delete", structuredEditAvailable: boolean): string {
+  const structuredHint = structuredEditAvailable
+    ? " Use the structured edit tools (`edit_file` / `multi_edit`) or the `apply_patch` tool for file edits so Codex can approve the change, enforce sandbox policy, show diffs, and record rollout."
+    : " Use the `apply_patch` tool for file edits so Codex can approve the change, enforce sandbox policy, show diffs, and record rollout.";
+  return `Cursor-native ${operation} is disabled for this Codex request because apply_patch is available.${structuredHint} No file was changed.`;
 }
 
 const NATIVE_LOCAL_EXEC_DISABLED =
@@ -84,13 +87,13 @@ export function readExec(execMsg: ExecServerMessage): Uint8Array {
   }
 }
 
-export function rejectWriteExecForApplyPatch(execMsg: ExecServerMessage): Uint8Array {
+export function rejectWriteExecForApplyPatch(execMsg: ExecServerMessage, structuredEditAvailable = false): Uint8Array {
   if (execMsg.message.case !== "writeArgs") throw new Error("invalid write exec");
   const path = resolve(execMsg.message.value.path);
   return execBytes(execMsg, "writeResult", create(WriteResultSchema, {
     result: {
       case: "rejected",
-      value: create(WriteRejectedSchema, { path, reason: codexNativeMutationRefusal("write") }),
+      value: create(WriteRejectedSchema, { path, reason: codexNativeMutationRefusal("write", structuredEditAvailable) }),
     },
   }));
 }
@@ -133,13 +136,13 @@ export function writeExec(execMsg: ExecServerMessage): Uint8Array {
   }
 }
 
-export function rejectDeleteExecForApplyPatch(execMsg: ExecServerMessage): Uint8Array {
+export function rejectDeleteExecForApplyPatch(execMsg: ExecServerMessage, structuredEditAvailable = false): Uint8Array {
   if (execMsg.message.case !== "deleteArgs") throw new Error("invalid delete exec");
   const path = resolve(execMsg.message.value.path);
   return execBytes(execMsg, "deleteResult", create(DeleteResultSchema, {
     result: {
       case: "rejected",
-      value: create(DeleteRejectedSchema, { path, reason: codexNativeMutationRefusal("delete") }),
+      value: create(DeleteRejectedSchema, { path, reason: codexNativeMutationRefusal("delete", structuredEditAvailable) }),
     },
   }));
 }
--- a/src/adapters/cursor/native-exec.ts
+++ b/src/adapters/cursor/native-exec.ts
@@ -70,6 +70,8 @@ export interface CursorNativeExecContext extends CursorNativeExecDeps {
   unsafeAllowNativeLocalExec?: boolean;
   /** apply_patch is visible for this request; Cursor-native write/delete must not bypass Codex. */
   rejectNativeFileMutations?: boolean;
+  /** The synthetic exact-match edit tools (edit_file / multi_edit) are advertised this request. */
+  structuredEditAvailable?: boolean;
 }
 
 export function cursorUnsafeNativeLocalExecEnabled(input: Pick<CursorNativeExecContext, "unsafeAllowNativeLocalExec"> = {}): boolean {
@@ -514,8 +516,8 @@ export async function handleCursorNativeExec(execMsg: ExecServerMessage, deps: C
     if (execCase === "fetchArgs") return [rejectFetchExecForPolicy(execMsg)];
   }
   if (execCase === "readArgs") return [readExec(execMsg)];
-  if (execCase === "writeArgs") return [deps.rejectNativeFileMutations ? rejectWriteExecForApplyPatch(execMsg) : writeExec(execMsg)];
-  if (execCase === "deleteArgs") return [deps.rejectNativeFileMutations ? rejectDeleteExecForApplyPatch(execMsg) : deleteExec(execMsg)];
+  if (execCase === "writeArgs") return [deps.rejectNativeFileMutations ? rejectWriteExecForApplyPatch(execMsg, deps.structuredEditAvailable === true) : writeExec(execMsg)];
+  if (execCase === "deleteArgs") return [deps.rejectNativeFileMutations ? rejectDeleteExecForApplyPatch(execMsg, deps.structuredEditAvailable === true) : deleteExec(execMsg)];
   if (execCase === "lsArgs") return [lsExec(execMsg)];
   if (execCase === "grepArgs") return [grepExec(execMsg)];
   if (execCase === "shellArgs") return [shellExec(execMsg)];
--- a/src/adapters/cursor/protobuf-events.ts
+++ b/src/adapters/cursor/protobuf-events.ts
@@ -164,7 +164,33 @@ export interface CursorProtobufEventState {
   toolSchemas?: Map<string, unknown>;
   /** Cursor wire-name → original Responses/Codex tool name for this request. */
   cursorToolNameMap?: Map<string, string>;
+  /**
+   * Bare names WE advertised as synthetic structured-edit tools on this request.
+   * See structuredEditCallIsOurs: conversion is gated on provenance, not on the name.
+   */
+  syntheticStructuredEditToolNames?: ReadonlySet<string>;
   translatorBudget?: TranslatorBudget;
+}
+
+
+/**
+ * Did WE advertise this bare tool name as a synthetic structured-edit tool on this request?
+ *
+ * Provenance, not a name test. `edit_file` / `multi_edit` are ordinary names a client or MCP
+ * server may legitimately expose, and `cursorStructuredEditTools` already refuses to shadow one
+ * that exists. Converting on the name alone would undo that refusal at the other end of the
+ * request: the client's own call would be silently re-emitted as `apply_patch`, or dropped with
+ * an error naming a conversion the user never asked for.
+ *
+ * Absent set = we advertised nothing, so nothing converts. Fail-closed in the safe direction:
+ * an unconverted structured call is a visible, recoverable failure; a wrongly converted one
+ * edits a file.
+ */
+function structuredEditCallIsOurs(
+  advertised: ReadonlySet<string> | undefined,
+  toolName: string,
+): boolean {
+  return advertised?.has(toolName) === true;
 }
 
 export function createCursorProtobufEventState(options: {
@@ -172,6 +198,7 @@ export function createCursorProtobufEventState(options: {
   parallelToolCalls?: boolean;
   toolSchemas?: Map<string, unknown>;
   cursorToolNameMap?: Map<string, string>;
+  syntheticStructuredEditToolNames?: Iterable<string>;
   contextUsage?: CursorContextUsageControls;
   /**
    * Request-local input estimate derived from the payload actually sent. Used only
@@ -188,6 +215,9 @@ export function createCursorProtobufEventState(options: {
     openToolCalls: new Map(),
     completedToolCalls: new Set(),
     ...(options.clientToolNames ? { clientToolNames: new Set(options.clientToolNames) } : {}),
+    ...(options.syntheticStructuredEditToolNames
+      ? { syntheticStructuredEditToolNames: new Set(options.syntheticStructuredEditToolNames) }
+      : {}),
     ...(options.parallelToolCalls !== undefined ? { parallelToolCalls: options.parallelToolCalls } : {}),
     startedClientToolCalls: 0,
     ...(options.toolSchemas ? { toolSchemas: options.toolSchemas } : {}),
@@ -454,13 +484,16 @@ export function mapSyntheticMcpExecToToolEvents(
       return [{ type: "error", message: cursorShellBridgeDropError(responsesName) }];
     }
   }
-  const translation = translateStructuredEditCall(responsesName, normalizedArgs);
-  if (translation?.error !== undefined) {
-    return [{ type: "error", message: `${responsesName} call was not converted to apply_patch: ${translation.error}` }];
-  }
-  const emittedName = translation ? CODEX_APPLY_PATCH_TOOL : responsesName;
-  const emittedArgs = translation ? JSON.stringify({ input: translation.patch }) : normalizedArgs;
   // Stateless fallback (no shared event state): emit a complete, self-contained tool call.
+  //
+  // No conversion happens here by design (#1036 review). Structured-edit translation is gated on
+  // provenance — did WE advertise this bare name on THIS request — and that record lives on the
+  // request state, which this branch does not have. Converting anyway would reinstate the exact
+  // hazard the gate exists to close: a client or MCP tool legitimately named `edit_file` would be
+  // rewritten into an apply_patch it never asked for. The live path always carries state
+  // (live-transport seeds it), so this only affects direct/unit callers.
+  const emittedName = responsesName;
+  const emittedArgs = normalizedArgs;
   return [
     { type: "tool_call_start", id: callId, name: emittedName },
     ...(emittedArgs.length > 2
@@ -523,7 +556,9 @@ function commitToolCall(state: CursorProtobufEventState, callId: string, finalAr
   }
   // Structured edit calls are converted to apply_patch here so both the interactionUpdate and the
   // native-exec mcpArgs paths emit the same valid freeform payload (#1017).
-  const translation = translateStructuredEditCall(open.name, finalArgs);
+  const translation = structuredEditCallIsOurs(state.syntheticStructuredEditToolNames, open.name)
+    ? translateStructuredEditCall(open.name, finalArgs)
+    : undefined;
   if (translation?.error !== undefined) {
     return dropStructuredEditCall(state, callId, open.name, translation.error);
   }
--- a/src/adapters/cursor/protobuf-events.ts
+++ b/src/adapters/cursor/protobuf-events.ts
@@ -340,8 +340,19 @@ function replacementHunk(oldString: string, newString: string): { hunk: string }
         "structured edit requires a non-empty old_string to locate the replacement; for new files or insertions without existing text, call apply_patch with an `*** Add File` / context hunk or use the shell bridge",
     };
   }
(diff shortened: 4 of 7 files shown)


Wibias:
Hey, thanks for this PR. Please test run that on a local dev server so we dont break something else by accident since cursor tool calls are pretty fragile. After you can give us some validation that this fixes the problem and doesnt implement new one's we will review it. Thanks.

ZachDreamZ:
Validation run, as requested:

**Local dev server (built from this branch)**

- Booted the proxy from the checkout with an isolated \OPENCODEX_HOME\ on port 10101 (the production 10100 instance was left untouched; it stayed healthy throughout).
- \GET /v1/models\ -> 200 with the provider catalog.
- \POST /v1/responses\ with \opencode-free/deepseek-v4-flash-free\ -> 200, \status: completed\, output \DEV-OK\, full usage reported. So the new code boots and serves real requests without regressing the non-Cursor path.

**Cursor tool-call path (no live Cursor credential available on this machine, so validated at the transport level)**

The new regression suite simulates Cursor server frames through the real protobuf event pipeline and asserts the exact wire output:

- \dit_file\ / \multi_edit\ calls emitted as \pply_patch\ with a valid \*** Begin Patch\ payload (verified byte-for-byte);
- the native-exec \mcpArgs\ path (\planMcpArgsHandling\) produces the same translated call;
- malformed calls (bad JSON, missing fields, empty \old_string\, empty \dits[]\) are dropped with a clear error instead of being relayed for local rejection after HTTP 200;
- tool-budget pinning keeps the synthetic tools alongside \pply_patch\, explicit tool_choice is never widened, and a pre-existing bare \dit_file\ in the client catalog is not shadowed.

**Gates**

- \un run typecheck\ / \un run lint:gui\ / \un run privacy:scan\ all pass.
- \un test tests/cursor-structured-edit.test.ts\ -> 18 pass (new).
- 238 related cursor/bridge tests across 14 files -> all pass.

If you have a Cursor session handy, a live \cursor/grok-4.5\ edit trial would be the strongest confirmation — happy to iterate on anything it surfaces.

Wibias:
## [GD] Verdict: gated


### TLDR

- **PR:** #1036 — `fix(cursor): structured edit tools convert to valid apply_patch calls (#1017)`
- **Head:** `e80c4061` on `dev` (mergeStateStatus: `UNSTABLE`, mergeable: `MERGEABLE`, draft)
- **Decision:** gated — the PR is a GitHub draft, 11 bot review threads are unresolved, and required CI has never run on this head (fork-PR runs await maintainer approval)
- **Usefulness:** high — directly addresses #1017 and the older #372: `cursor/grok-4.5` cannot emit Codex's freeform patch grammar, and this adds Cursor-shaped `edit_file` / `multi_edit` tools converted server-side into valid `apply_patch` payloads, with malformed calls dropped instead of relayed
- **Bugs:** 5 concrete findings verified at head — trailing-newline-only edits silently no-op (`patchLines` in `protobuf-events.ts`); context-free `@@` hunks can edit the wrong duplicate occurrence; `multi_edit` does not preserve ordered/sequential edit semantics; EOF-without-newline replacements are emitted newline-terminated; client-owned bare `edit_file` / `multi_edit` calls are still hijacked into `apply_patch` (catalog collision fixed in `2b737025`, call-translation ownership not)
- **Security:** none — no auth/credential surface changed; payloads relay over the existing `apply_patch` channel; `bun run privacy:scan` passes
- **Spec / standards:** one overclaim in the PR body ("never truncated away" — the byte ceiling can still drop the pinned tools) plus the schema-vs-translation contract gaps above; no `docs-site/` update for a user-facing behavior change
- **Reviews:** 8 CodeRabbit + 3 Codex connector threads, all unresolved; 3 verified fully valid, 1 partially fixed (collision), 3 real limitations needing a decision or documentation, 2 test/refactor gaps; the human validation exchange (Wibias to author) is only partially fulfilled — no live Cursor session edit trial yet
- **Base / CI:** head is 53 commits behind `dev` with no overlapping file changes (owner action: update from latest base). Required CI (Cross-platform CI, React Doctor) is `action_required` — no matrix job has run on this head
- **Gate:** draft + unresolved review threads + CI not run
- **Owner actions (foreign PR):** after fixing the verified findings, update from latest `dev`, get CI approved and green, and complete a live `cursor/grok-4.5` edit trial (the validation Wibias requested)
- **Bottom line:** useful and worth landing, but not merge-ready — the author should fix the trailing-newline no-op, the translation ownership guard, and either implement or explicitly document the `multi_edit` ordering and EOF-newline limits, then update from `dev` and get CI green before marking ready.

<details>
<summary><b>Full verdict</b></summary>

### Semantic propagation

- **Concepts audited:** structured-edit tool advertisement (`cursorStructuredEditTools`, `CURSOR_EDIT_FILE_INPUT_SCHEMA`, `CURSOR_MULTI_EDIT_INPUT_SCHEMA`); structured-edit call translation (`translateStructuredEditCall`, `replacementHunk`, `patchLines`); wire-name mapping (`normalizeCursorWireName`, `responsesToolNameFromCursorWire`, `cursorToolNameMap`); budget pinning (`toolPriority` priority 1); guidance note (`buildCursorToolGuidanceSystemNote`); native-mutation refusal (`codexNativeMutationRefusal`); tool-choice narrowing; synthetic-vs-client-owned name ownership.
- **Authoritative sources:** `tool-definitions.ts` (schemas + advertisement + guidance), `protobuf-events.ts` (translation + event paths), `request-builder.ts` (budget), `native-exec-fs.ts` (refusal), `live-transport.ts` + `protobuf-request.ts` (state construction, `clientToolDefs`, guidance wiring).
- **Producers and consumers checked:** `applyCursorToolBudget` to `request.tools` to `live-transport` (`clientToolDefs` via `buildCursorToolDefinitions`) and `protobuf-request` (guidance note, built from the same `cursorToolsForActivePrompt` set); `translateStructuredEditCall` consumed by `commitToolCall` (stateful `interactionUpdate`) and `mapSyntheticMcpExecToToolEvents` (stateless + native-exec `mcpArgs` via `planMcpArgsHandling`); the new exports are imported only inside the cursor adapter and its tests — no external consumers.
- **Public/derived representations checked:** the emitted `apply_patch` custom tool call `{ "input": "*** Begin Patch ... *** End Patch" }` — shape verified byte-for-byte in tests and matches the #820 invariant (freeform `{ input }` restored as `custom_tool_call`); guidance and refusal text are the other public representations.
- **Material variant partitions checked:** `edit_file` vs `multi_edit`; `auto` / `required` vs forced / allow-listed `tool_choice`; bare vs namespaced `apply_patch`; collision vs no-collision catalogs; stateful vs stateless vs native-exec paths; malformed vs valid calls; single vs multi-line replacements; deletion (`new_string: ""`).
- **Positive and negative assertions checked:** no synthetic tools without an advertised bare freeform `apply_patch`; none for forced / allow-listed choices; none for namespaced `apply_patch`; catalog-side shadow skip for client-owned names (tested); call-translation ownership guard — absent (gap); trailing-newline-only change rejection — absent (gap); `multi_edit` ordered-edit and EOF-no-newline semantics — unimplemented and undocumented (gaps).
- **Unmapped surfaces:** `cursorToolsForActivePrompt` tool-count-demo narrowing can drop the structured tools from the visible catalog while `rejectNativeFileMutations` still fires with a message naming them (the refusal-availability minor); `catalogLimitNote` counts synthetic tools in `kept.length` (cosmetic, truncation path only); `docs-site/` has no note for the new user-facing behavior.
- **Unproven equivalence assumptions:** stateful tests use an identity `cursorToolNameMap`, so the wire-name folding path (`mcp_opencodex-responses_edit_file` to `edit_file`) is not covered through events; `multi_edit` has no stateful / native-exec event-path coverage; one representative replacement is used without proving equivalence for the trailing-newline, duplicate-`old_string`, and EOF-no-newline cases.
- **Representation mismatches:** `StructuredEditTranslation.patch` is documented and exported as the complete `apply_patch` payload but `addReplacement` returns a bare `@@` hunk under the same type; `new_string: "bar"` is represented as newline-terminated `+bar` regardless of the file's EOF state.
- **Coverage gaps:** trailing-newline-only edit (silent no-op), client-owned `edit_file` call translation, `multi_edit` dependent / out-of-order edits, EOF-no-newline replacement, non-identity wire-name mapping through the stateful path, and a `multi_edit` event-path test.
- **Axis verdict:** blocked — see Bugs / correctness for the concrete items.

### Usefulness

High. Issue #1017 (and its older sibling #372) documents a real, reproducible failure: `cursor/grok-4.5` (and other Cursor-trained models) cannot emit Codex's freeform `*** Begin Patch` grammar, so every edit attempt on the Cursor route produced malformed patch text that Codex rejected locally after the request already returned HTTP 200. This PR implements option 2 from the issue — provider-compatible structured edit tools converted server-side — which is the right architectural fix, and it keeps the Codex approval / sandbox boundary intact by always emitting `apply_patch` on the wire. The malformed-call rejection removes the silent 200-then-local-rejection mode. Verified locally at head `e80c4061`: `bun run typecheck` pass, `bun run privacy:scan` pass, `tests/cursor-structured-edit.test.ts` 18/18, 14-file cursor suite 238/238. The PR body's stated gates match my runs.

### Bugs / correctness

Method: bug-review.md — Bugbot n/a (Codex host); static review plus targeted reproduction scripts run against the head worktree; complementary edge-case probes (silent failures, edge cases) run directly on `translateStructuredEditCall`.

Verified findings (all reproduced against `e80c4061`):

1. **Trailing-newline-only edits are a silent no-op** — `patchLines` pops one trailing empty element on each side independently, so `old_string: "export {};\n"` to `new_string: "export {};"` emits `@@` / `-export {};` / `+export {};`: the newline change is discarded and the model reads success while nothing changes. Same for adding a lone trailing newline. The schema promises "exact text replacement including line breaks". Fix: compare the normalized `patchLines` results and return an actionable error when identical; add the malformed-matrix case. (CodeRabbit threads 3717200617, 3717200646 — valid.)
2. **Client-owned `edit_file` / `multi_edit` calls are still hijacked** — commit `2b737025` fixed the catalog side (synthetic tools are skipped when the client catalog already has a bare same-named tool), but translation keys only on the name (`isCursorStructuredEditToolName(open.name)` and the stateless `translateStructuredEditCall(responsesName, ...)`), so a call to a client-owned bare `edit_file` is still rewritten into `apply_patch` (verified: args with non-schema fields still translate). Fix: carry the per-request synthetic-name set into `commitToolCall` and `mapSyntheticMcpExecToToolEvents` and translate only those names; add a call-translation collision test. (CodeRabbit 3717200638, Codex 3717205360 — partially fixed at head.)
3. **`multi_edit` does not preserve ordered / sequential edit semantics** — each edit becomes a separate `@@` hunk in one `*** Update File` section applied against the original file. Verified: `edits: [{a→b},{b→c}]` emits `-a/+b` then `-b/+c`; the second hunk cannot match because `b` does not exist in the original file. Cursor's native MultiEdit applies edits sequentially, so the advertised "ordered exact-text replacements" promise does not hold. The proxy has no workspace file access to pre-apply or sort. Fix or decide: reject edits whose `old_string` appears inside an earlier edit's `new_string`, and document that hunks must all match the current file content independently; or drop the "ordered" claim from the schema description. (Codex 3717205367 — valid limitation.)
4. **EOF-without-newline replacements become newline-terminated** — `new_string: "bar"` is emitted as `+bar`, which Codex applies as `bar\n`, violating the exact-bytes replacement contract for a file whose final line lacks a trailing newline. Fix or decide: document the limitation in the schema / description, or reject when `new_string` / `old_string` end without a newline and the hunk is line-based. (Codex 3717205371 — valid limitation.)
5. **Context-free `@@` hunks can edit the wrong duplicate occurrence** — Codex's `seek_sequence` picks the first match without a uniqueness check, so a duplicated `old_string` silently modifies the wrong spot. The proxy cannot verify uniqueness without file access. Decision needed: document the first-occurrence semantics in the tool description (Cursor-native Edit behaves the same way) and / or advise models to include surrounding context inside `old_string`. (CodeRabbit 3717200621 — valid limitation.)
6. **Refusal message names `edit_file` / `multi_edit` when they are not advertised** — `codexNativeMutationRefusal` has no availability input; on forced / allow-listed `tool_choice`, byte-ceiling drops, or tool-count-demo narrowing the message still names tools absent from the catalog, inviting a hallucinated call. Fix: thread a `structuredEditAvailable` flag from the same per-request state into both call sites. (CodeRabbit 3717200611 — valid minor.)
7. **`addReplacement` reuses `StructuredEditTranslation` with mismatched semantics** — the exported type documents a complete payload but `{ patch }` here is a bare hunk; two narrowing idioms in one flow. Refactor to a dedicated `HunkResult` type. (CodeRabbit 3717200634 — valid code-quality.)
8. **Optional chaining after `expect(...)`** in the new tests (lines 165–208) — runs today but invites false-pass copies; switch to direct `toEqual`. (CodeRabbit 3717200642 — trivial.)
9. **Test coverage gaps** — stateful tests use identity `cursorToolNameMap` (no non-identity wire-name case) and no `multi_edit` stateful / native-exec coverage; the trailing-newline malformed case is missing. (CodeRabbit 3717200651, 3717200646 — valid.)

Not reproduced / declined:
- No issue found on the non-Cursor path (the changed code is unreachable outside the Cursor route; the author's boot + non-Cursor POST validation is consistent with the code).
- The "synthetic tools never truncated away" claim in the PR body overstates: they are pinned at priority 1, but the hard byte ceiling can still drop them in an extreme catalog (`tryKeep` returns false on bytes), which would also leave the guidance note naming tools that were omitted. Minor.

Fixed this session: none (foreign PR, review-only run).

### Security

- Scope reviewed: tool-call relay surface, prompt / guidance text; no auth / credential handling changed.
- Findings: none. The adapter relays only patch text over the pre-existing `apply_patch` channel; refusal / guidance strings are static; `bun run privacy:scan` passes at head. No secret, token, or body logging introduced. The new tool schemas accept file paths and text — the same exposure `apply_patch` already had.
- Fixed this session: none.

### Spec / standards

- Spec source: issue #1017 (option 2), related #820 invariant 2 (freeform tools restored as `custom_tool_call`), #372 (grok-4.5 cannot emit the grammar), repo `AGENTS.md` norms.
- Gaps: PR body claims the synthetic tools are "never truncated away" (byte ceiling can drop them); schema promises exact replacement "including line breaks" while three cases (trailing-newline, EOF-no-newline, duplicate `old_string`) silently degrade; `multi_edit` description promises "ordered" semantics the conversion cannot deliver; user-facing behavior change ships without a `docs-site/` update; no regression test for the stateful wire-name and `multi_edit` event paths.

### Reviews

- Humans: Wibias asked for a local dev-server test run plus validation that the fix works and introduces no regressions; ZachDreamZ provided a validation run (isolated proxy boot on port 10101, `/v1/models` 200, non-Cursor `/v1/responses` 200 `DEV-OK`, transport-level Cursor validation via the regression suite, all stated gates green). Assessment: the reported boot / test results match what I reproduced locally; the specific ask — a live `cursor/grok-4.5` edit trial — is still open (the author explicitly notes no live Cursor credential was available).
- Bots: CodeRabbit (review on `e1e17328`) posted 8 inline comments; Codex connector posted 3 P2 comments; all 11 threads unresolved. Verified against head: 3 fully valid, 1 partially fixed, 3 real limitations needing decision / documentation, 2 test / refactor gaps (details above). No bot finding was a false positive.
- No human thread received an agent reply in this run (review-only; the verdict aggregates).

### Base / CI

- Behind / conflicts: head `e80c4061` is 53 commits behind `dev` (merge-base `ab5b20ca`, 2026-08-05); no post-branch-point `dev` commit touches any of the 5 changed files; GitHub reports `MERGEABLE`. Owner action (foreign PR): update from latest `dev`.
- Required checks: `enforce-target` pass and `label` pass only. `Cross-platform CI` and `React Doctor` runs exist but are `action_required` — they wait for a maintainer approval because this is a first-time-contributor fork PR; no matrix job has executed on this head, so the repo's typecheck / test / lint / privacy matrix is unproven on this head. Maintainer action: approve the two runs once the author fixes findings.
- Local tip compile / tests: `bun run typecheck` pass; `bun test tests/cursor-structured-edit.test.ts` 18/18; 14-file cursor / bridge suite 238/238; `bun run privacy:scan` pass. Full local suite was still running at publication time and is not relied on.
- Authoritative gate (`ship-gate.mjs`, mutation-mode review, workflow `references/full-review-pr.md`): `blocked` — `reviewPolicy:draft` + `reviewThreads:unresolved_review_threads`. Base-health advisory: 2 base-tip failures not present on this head (track separately; do not expand this PR).

### Gate

Draft (GitHub draft PR) + 11 unresolved review threads + required CI not run (`action_required`). Not merge-ready under any reading.

### Bottom line

This is a good, well-scoped answer to #1017 / #372 with honest engineering: the server-side conversion approach is right, the malformed-call rejection fixes the silent failure mode, the new suite is meaningful, and the local gates reproduce. It should land after the author fixes the silent trailing-newline no-op, adds the translation ownership guard, and either implements or explicitly documents the `multi_edit` ordering and EOF-newline limits; then updates from latest `dev`, gets CI approved and green, and runs the live `cursor/grok-4.5` edit trial that was requested. Maintainers: approve the CI runs and re-review after those changes.

</details>

Wibias:
[GD] Validation conclusion

The requested local dev-server run was delivered: the branch boots in isolation (port 10101, production 10100 untouched and healthy), serves a real request (`GET /v1/models` 200, `POST /v1/responses` 200 `DEV-OK`), and the reported #1017 failure is fixed at the transport level — `edit_file` / `multi_edit` convert to valid `apply_patch` payloads and malformed calls are dropped with a clear error instead of being relayed for local rejection. The stated gates (typecheck, privacy:scan, 18 new + 238 cursor tests) reproduce locally.

Still open before maintainer review:

- "Does not implement new ones" is not yet met — the earlier verdict lists five new edge-case behaviors this validation suite does not exercise.
- Required CI has never run on this head (the fork-PR runs await maintainer approval).
- A live `cursor/grok-4.5` edit trial is still pending (no Cursor credential was available on the author's machine).

Address those (or explicitly decline each with rationale), update from latest `dev`, and run the live trial — then we approve CI and review.

ZachDreamZ:
## Addressed review feedback — commit a35115c3

Thanks @coderabbitai and @chatgpt-codex-connector for the thorough review. All actionable items are addressed in the latest push.

### CodeRabbit comments

**1. `native-exec-fs.ts:42-44` — refusal hint gating** ✅
`codexNativeMutationRefusal(operation, structuredEditAvailable)` now only mentions `edit_file`/`multi_edit` when those tools are actually advertised. The flag is threaded from `live-transport.ts` via `cursorRequestAdvertisesStructuredEdits()` → `execContext.structuredEditAvailable` → `handleCursorNativeExec`.

**2. `protobuf-events.ts:328-346` — trailing-newline no-op** ✅
`replacementHunk` now compares normalized `oldLines`/`newLines` and returns an error for identical content (covers both trailing-newline-only and literal identical pairs). Added two test cases.

**3. `protobuf-events.ts:335-346` — ambiguous context-free `@@` hunks**
This is a real limitation of translating exact-match replacements to line-based hunks without file-content access. Documented it in both tool descriptions (`tool-definitions.ts`): edits must match at exactly one location, and matching is line-based. The translation layer doesn't have file content available to do a uniqueness check at this layer — that's a schema-level contract on the model, enforced by apply_patch's own ambiguity rejection downstream.

**4. `protobuf-events.ts:386-395` — `addReplacement` return type** ✅
Now returns `StructuredEditTranslation` (single type) instead of overloading `patch` to mean both a bare hunk and a full payload. Call sites use consistent `error !== undefined` narrowing.

**5. `tool-definitions.ts:205-225` — collision protection** ✅ (already fixed in 2b737025, verified)

**6. Test assertions — remove `?.` after `expect()`** ✅
All `expect(...)?.toMatchObject(...)` replaced with `expect(...).toEqual(...)` or `expect(...).toEqual(expect.objectContaining(...))`.

**7. Trailing-newline test case** ✅
Added `rejects a trailing-newline-only edit as a silent no-op` and `rejects identical old/new as a no-op`.

**8. Stateful non-identity wire-name tests** ✅
Added `translates a non-identity wire-name mapping (Cursor display name -> Codex tool name) for multi_edit` covering the `mcp_opencodex-responses_multi_edit` display-name path.

### Codex connector comments

- **P2 request-builder.ts:73 — duplicate synthetic tools** ✅ (already fixed in 2b737025)
- **P2 protobuf-events.ts:407 — multi_edit ordering** — Documented in the `multi_edit` tool description: every `old_string` is matched against the ORIGINAL file content, so a later edit must not rely on text introduced by an earlier one. This is the correct contract since the translation layer emits independent hunks.
- **P2 protobuf-events.ts:345 — no-newline replacements** — Documented in both tool descriptions: matching is line-based, so an edit cannot add or remove only the file's final newline; such edits are now rejected as no-ops.

### Verification
- 21 focused tests pass (`bun test tests/cursor-structured-edit.test.ts`)
- `bun run typecheck` clean
- `bun run lint:gui` clean

lidge-jun (changes requested):
Thanks — the structured `edit_file`/`multi_edit` + server-side translation approach is the right direction for #1017, and we want to land it. Two gaps block it today:

1. **Synthetic-tool provenance.** Conversion keys on the bare tool name. If a client already exposes its own `edit_file`, its calls would be translated too. Track the synthetic names injected for *this request* (a per-request set threaded from tool-catalog construction to the conversion site in `src/adapters/cursor/protobuf-events.ts` / `live-transport.ts`) and convert only those.
2. **Final-catalog derivation.** Structured-edit availability is derived from the original request rather than the final prompt-filtered tool catalog; when filtering drops the tools the flag is stale.

A regression test for each (client-owned `edit_file` passes through untouched; filtered catalog disables translation) and we can re-review it.

lidge-jun:
Re-reviewed against current `dev`. The compatibility gap you are closing is real — the Cursor route advertises freeform `apply_patch` (`src/adapters/cursor/tool-definitions.ts:135-141`) while `dev` has no structured-edit translation, which is exactly #1017. The conversion itself reads carefully: validated JSON, explicit drop errors instead of silent failure, and the same treatment on both the `interactionUpdate` and native-exec paths so the two cannot diverge.

The prior review asked for synthetic-tool provenance. This is that finding again, now with a reproduction, so I want to be concrete rather than repeat the phrase.

**The gap.** `translateStructuredEditCall` decides provenance from the tool NAME alone:

```ts
if (!isCursorStructuredEditToolName(toolName)) return undefined;
```

and both call sites pass a name that arrived from the wire — `mapSyntheticMcpExecToToolEvents` passes `responsesName`, and the completion path passes `open.name`. Neither consults whether *this request* actually advertised the synthetic tools. Your own `cursorStructuredEditTools` takes care not to shadow a real client tool with either name, so the collision is already understood as possible in the injection path; the translation path just does not carry that knowledge forward.

**Reproduction.** A user runs an MCP server exposing `edit_file` — not an unusual name; several editor-adjacent MCP servers use it. Their call arrives named `edit_file`, gets JSON-parsed as `{file_path, old_string, new_string}`, and one of two things happens:

- the shape happens to match, and the call is silently re-emitted as `apply_patch` against a file path meant for a different tool; or
- the shape does not match, and the user gets `edit_file call was not converted to apply_patch: edit_file is missing a non-empty file_path; the call was dropped` — an error naming a conversion they never asked for, about a tool that was working before they enabled Cursor.

The second is the more likely and the more confusing: the message describes our machinery, not their world.

**What I am asking for.** Tag the synthetic tools at injection time and convert only calls carrying that tag. The state object already threads through both paths — `CursorProtobufEventState` has a natural place for "synthetic edit tools were advertised this request", and the diff already adds a comment describing exactly that flag (`The synthetic exact-match edit tools (edit_file / multi_edit) are advertised this request`), so the intent is there and only the enforcement is missing. Gate `translateStructuredEditCall` on it, and pass through untouched when the flag is absent.

Please add the regression test that would have caught this: a client tool named `edit_file` present in the request while the synthetic tools are NOT advertised, asserting the call passes through unconverted. The current tests exercise conversion mechanics thoroughly, which is good, but every one of them assumes our own tool is the caller — so they raise confidence about a case that cannot go wrong while leaving the one that can uncovered.

Everything else here I would take as is. The drop-with-explanation behavior is the right call over silent best-effort patching, and applying the conversion at the protobuf event boundary rather than in the adapter shell keeps it out of the generic path.

One process note: the failing `enforce-target` check on this PR is not your code — it is a repository-side gate artifact where a stale head's workflow YAML calls into base-pinned scripts. Rebase onto current `dev` and it clears; do not spend time chasing it.

lidge-jun:
I pushed a commit onto this branch rather than asking you to do another round — your four commits are untouched on top of a rebase onto current `dev`, and mine sits above them. The PR stays yours.

**What I changed, and only this:** the conversion now checks provenance instead of the tool name.

`translateStructuredEditCall` decided a call was ours from the name alone, and both call sites pass a name that came off the wire. Your `cursorStructuredEditTools` already refuses to shadow a client tool called `edit_file` or `multi_edit` — so the collision was understood at injection time — but that knowledge never reached the translation. A user running an MCP server that exposes `edit_file` would have their call silently re-emitted as `apply_patch`, or dropped with an error naming a conversion they never requested.

So `live-transport` now records the bare names we actually advertised on this request (from `cursorStructuredEditTools`, not from the name), the event state carries them, and both call sites convert only for names in that set.

**One behavior change worth your review:** the stateless fallback in `mapSyntheticMcpExecToToolEvents` now passes through instead of converting. It has no request state, so it cannot know whether we advertised anything, and the safe direction seemed clear — an unconverted structured call is a visible, recoverable failure, while a wrongly converted one edits a file. Live traffic always carries state, so real conversions are unaffected. Your `stateless native-exec path converts edit_file the same way` test pinned the old contract, so I rewrote it to pin the new one and explain why. If you think the fallback should convert, say so — that is a judgment call and yours is as good as mine.

**Everything else is yours and unchanged.** The validated JSON parsing, the argument spellings, the line-based hunks, the no-op and final-newline rejection, and especially the choice to drop with an explanation rather than emit a best-effort patch — that last one is the right instinct and it is most of the value here.

Verification:

```
tests/cursor-structured-edit.test.ts        22 pass / 0 fail
cursor-tool-budget + cursor-protobuf-events 33 pass / 0 fail
bun run typecheck                            clean
```

Ablation on the new collision test, restoring the name-only gate:

```
(fail) a client tool named edit_file is not hijacked when we advertised nothing
21 pass, 1 fail
```

Exactly one test goes red and your twenty-one conversion tests stay green — which is what shows the gate narrows behavior without breaking the feature.

The branch is now on current `dev`. Ready for review from my side; #1017 should close when this lands.

lidge-jun:
Thank you — this is landing as **#1144**, with your commits cherry-picked under your authorship, and #1017 closes with it.

The translator and its test suite are substantial work and they carried over largely as written.

**One correction.** Visibility was computed from `cursorVisibleTools`, but structured-edit availability and tool names were derived from the earlier `request.tools`. Once filtering or budgeting drops a tool the two disagree — and a synthetic edit identified by wire name alone can collide with a real client tool. The stack derives the set from the final filtered catalog and tags injected tools with internal provenance, with a post-filter regression test pinning it.

Your #1126 landed in the same PR — see the comment there.

Verified: 32 pass / 0 fail on the Cursor and replay suites; full suite 9,534 pass / 0 fail.

Leaving this PR open for you.

lidge-jun:

Five of your commits are in `dev` with their original authorship, plus a `Co-authored-by: NexusCore` trailer on every commit in the range, carried by #1144 (merge commit `a260006e3`). Issue #1017 is closed by that merge.

Closing because the work shipped. The translator and its test suite are substantial work and carried over largely as written.

**One correction:** visibility was computed from `cursorVisibleTools`, but structured-edit availability and tool names were derived from the earlier `request.tools`. Once filtering or budgeting drops a tool the two disagree — and a synthetic edit identified by wire name alone can collide with a real client tool. The landed version derives the set from the final filtered catalog and tags injected tools with internal provenance, with a post-filter regression test.

PR #1055, opened 2026-08-05
fix: finish passthrough SSE after terminal events

itling:
## Problem

When a provider (e.g. DeepSeek Flash v4) ends a Responses passthrough stream with a terminal event such as `response.completed` / `response.failed` / `response.incomplete` but does not send `data: [DONE]`, the relay keeps the client SSE stream open indefinitely. The client shows the assistant "thinking" forever and never receives the final response.

## Fix

In `src/server/relay.ts`, after forwarding a terminal event, emit `[DONE]` and close the client stream so the client can finalize the turn.

## Tests

- Added `tests/passthrough-abort.test.ts` coverage for a stream that ends with `response.completed` and no `[DONE]`.
- `bun test tests/passthrough-abort.test.ts` passes.
- Related relay tests and `bun run typecheck` pass.

## Review readiness checklist

- [x] I fixed all correct Codex and CodeRabbit findings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming response handling when terminal events omit the expected completion marker.
  * Ensured clients receive a consistent completion signal and connection closure after a response finishes.

* **Tests**
  * Added coverage confirming terminal events are preserved and the completion marker is appended when missing.

### DIFF
--- a/src/server/relay.ts
+++ b/src/server/relay.ts
@@ -517,11 +517,14 @@ export function relaySseWithHeartbeat(
   if (!body) return null;
   const reader = body.getReader();
   const decoder = new TextDecoder();
-  const heartbeat = new TextEncoder().encode(": opencodex keepalive\n\n");
+  const encoder = new TextEncoder();
+  const heartbeat = encoder.encode(": opencodex keepalive\n\n");
+  const doneSentinel = encoder.encode("data: [DONE]\n\n");
   let timer: ReturnType<typeof setInterval> | undefined;
   let closed = false;
   let clientCancelled = false;
   let terminalReported = false;
+  let doneSeen = false;
   let buffer = "";
 
   const reportTerminal = (status: ResponsesTerminalStatus) => {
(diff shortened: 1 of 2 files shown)


itling:
Closing this as superseded: upstream `dev` (and `main`) already contain 02ca79a3 `fix(responses): close passthrough streams at terminal events`, which implements this behavior in the production relay. Since that commit, native Responses passthrough routes through `relaySseWithFailedTail` + `createSseTerminalOutputBoundary`, which already closes the client stream at a Responses terminal event even when the upstream keeps the connection open, appends `data: [DONE]` when the upstream omits it, forwards only through the terminal SSE block boundary (dropping post-terminal frames), and handles split chunks and partial terminal blocks. `relaySseWithHeartbeat` (the function this PR modified) has no remaining production call sites, so this change would duplicate an existing fix in dead code. The scenarios are already covered in `tests/sse-failed-tail.test.ts` (terminal-without-DONE close, coalesced post-terminal frames, real-DONE detection). Recommendation: update the running OpenCodex build to a version containing 02ca79a3.

Issue #1061, opened 2026-08-05
macOS CI: native-profile process-exit phase test fails or hangs on release-train runs


lidge-jun:
### Client or integration

Other (test suite / CI)

### Area

Native profile / Codex transition tests (`tests/native-profile-startup.test.ts` family)

### Summary

The macOS leg of Cross-platform CI is near-deterministically red on the `native profile OpenCodex process-exit phases > hard OpenCodex process exit after each published transaction phase converges exact auth, vault, journal, gate, and runtime bearer` test. It blocked three release-train runs in a row on 2026-08-05:

- main promotion run 31003378927: the macOS job hung in `Test` for the full 30-minute job timeout and was cancelled
- preview release run 31007774402: same test failed after 4s (`SyntaxError: JSON Parse error: Unexpected EOF` immediately before it — reads like a file read before the writer finished)
- main release run 31009491939: same test failed after 630ms, twice in the log

The same commits pass on Linux (all shards) and pass on macOS on retry, so this is a timing-sensitive defect in the test or the phase-convergence path, not a deterministic product bug — but it currently decides whether a release can ship.

### Reproduction

1. Push any commit to `main` or `preview` (or rerun the macOS job) at `8949c4940` or later.
2. The macOS `Test` step either fails in the named test within seconds or hangs until the 30-minute timeout.

### Version

dev tip `8949c4940` (post-2.10.0)

### Operating system

macOS (`macos-latest` GitHub-hosted runner)

### Provider and model

Not provider-specific.

### Logs or error output

- https://github.com/lidge-jun/opencodex/actions/runs/31007774402/job/92311897103
- main release run 31009491939 macOS job (92317683364)
- main promotion run 31003378927 (macOS cancelled at the 30-minute timeout)

### Screenshots and supporting files

N/A

### Redacted configuration

Default CI configuration.

### Checks

- [x] I searched existing issues and documentation.
- [x] I removed secrets, tokens, account details, request credentials, and personal data.

lidge-jun:
Fixed on dev by PR #1071, merge commit ac1aa3373f573d3fe2a412dbb8e5afade5c4db11. The native-profile teardown is now bounded and the harness no longer parses partial JSON, which removes the macOS CI hang/flake this issue tracked (the flake fired live on #1072's first CI run — strongest field evidence). Ships with the next release train.

PR #1072, opened 2026-08-05
fix(codex): warn when a startup write leaves an app-server stale (#1046)

### LINKED ISSUE #1046: [Bug]: Service restart rewrites the Codex catalog without handling stale app-servers, so routed models stay missing from the App picker
### Client or integration

Codex App

### Area

Catalog / models

### Summary

After the proxy rewrites the Codex model catalog on startup, routed models stay missing from the Codex App model picker until the long-lived Codex `app-server` process is killed by hand. The catalog file on disk is already correct at that point, so the state looks healthy from every disk-level check while the picker still shows only native models.

The stale-app-server problem itself is known and already handled for the CLI sync path (#476, #857). The gap is that the *server startup* path performs the same catalog/cache write without the accompanying app-server handling:

- `src/cli/index.ts` (`sync`, `sync-cache`) calls `invalidateCodexModelsCache()` and then `afterCatalogWriteHandleAppServers({ restart })`, which warns about stale app-servers or SIGTERMs them.
- `src/server/index.ts:313` calls the same `invalidateCodexModelsCache()` on boot with no equivalent follow-up. `afterCatalogWriteHandleAppServers` is never referenced outside `src/cli/index.ts`.

So `ocx service repair` / a service restart rewrites `models_cache.json` and leaves an already-running `app-server` holding the previous in-memory list, with no warning emitted. The user sees a correct catalog on disk and an incorrect picker, and nothing in the output suggests why.

This is most visible in an SSH remote workspace, where the app-server is long-lived and boots independently of the app, but it is not SSH-specific.

Note that `collectCodexAppServerCatalogState()` already computes exactly the signal needed here (`stale` when an app-server start time predates the catalog mtime). It just is not consulted on the server startup path.

### Reproduction

1. Install and run opencodex on a host with at least one routed provider configured (Anthropic OAuth in my case) and confirm the routed models are enabled.
2. Open that host as a workspace in the Codex App so a long-lived `codex ... app-server --listen unix://` process is running.
3. On the host, change something that causes a catalog rewrite and restart the service: `ocx service repair`.
4. Confirm the on-disk state is correct:
   - `~/.codex/opencodex-catalog.json` lists the routed models with `visibility: "list"`
   - `~/.codex/models_cache.json` has the same entries
   - `curl "http://127.0.0.1:10100/v1/models?client_version=<codex version>"` returns them
   - `codex exec --model <provider>/<routed-model> "..."` answers normally
5. Open the model picker in the Codex App.

Actual: only the native `gpt-*` models are listed. The routed entries are absent even though every check in step 4 passes.

Expected: either the routed models appear, or the service restart emits the same stale-app-server warning the CLI sync path already produces (`formatStaleCodexAppServerWarning`).

Workaround that fixes it immediately:

```
pkill -f "app-server --listen unix://"
rm -f ~/.codex/app-server-control/app-server-control.sock
```

After the app-server restarts, the picker shows the routed models.

One secondary detail from the same session: the app-server left a stale `app-server-control.sock` behind, and the replacement process logged `Error: app-server control socket is already in use` on startup. It kept running, but the removal above was needed before the restart came up clean.

### Version

2.10.0

### Operating system

macOS 15 (Apple Silicon, Darwin 24.6.0)

### Provider and model

anthropic / claude-opus-5, claude-fable-5 (routed as `anthropic/claude-opus-5`, `anthropic/claude-fable-5`)

### Logs or error output

Service log after the restart, showing the catalog write with no stale-app-server warning:

```
🚀 opencodex proxy running on http://localhost:10100
[opencodex] Codex runtime: /Users/[USER]/.local/bin/codex (version=0.146.0, source=configured)
Pointed Codex's built-in openai provider at the opencodex proxy (openai_base_url).
  Codex model catalog: /Users/[USER]/.codex/opencodex-catalog.json
[opencodex] catalog sync: routed model fetch returned empty; preserving 11 existing routed entries on disk.
```

Catalog and cache agree on disk while the picker disagrees:

```
$ python3 - <<'PY'
import json
d=json.load(open("/Users/[USER]/.codex/opencodex-catalog.json"))
for m in d["models"]: print(m["slug"], m.get("visibility"))
PY
gpt-5.6-sol list
gpt-5.6-terra list
gpt-5.6-luna list
gpt-5.5 list
gpt-5.4 list
gpt-5.4-mini list
gpt-5.3-codex-spark list
anthropic/claude-fable-5 list
anthropic/claude-opus-5 list
```

Routing works end to end the whole time:

```
$ codex exec --model anthropic/claude-fable-5 --skip-git-repo-check "reply with exactly: FABLE_OK"
FABLE_OK
```

Stale app-server startup error mentioned above:

```
Error: app-server control socket is already in use at /Users/[USER]/.codex/app-server-control/app-server-control.sock
```

Model picker after the catalog rewrite: only `5.5`, `5.6 Sol`, `5.6 Terra`, `5.6 Luna`, `5.4 Mini`, `5.4`, `5.3 Codex Spark` are offered; no `anthropic/*` entries.

### Redacted configuration

```json
{
  "port": 10100,
  "providers": {
    "openai": {
      "authMode": "forward",
      "codexAccountMode": "pool"
    },
    "anthropic": {
      "adapter": "anthropic",
      "baseUrl": "https://api.anthropic.com",
      "authMode": "oauth",
      "defaultModel": "claude-sonnet-5"
    }
  },
  "defaultProvider": "openai",
  "codexAutoStart": true,
  "websockets": false,
  "disabledModels": ["anthropic/claude-sonnet-5", "..."]
}
```

(Most Anthropic models are intentionally in `disabledModels`; `claude-opus-5` and `claude-fable-5` are the two deliberately left enabled, and those are the two missing from the picker.)

### Checks

lidge-jun:
## Summary

Service startup rewrites the Codex catalog and the models cache, but an app-server that booted earlier keeps its own in-memory model list — Codex builds a static models manager from the catalog once and never rereads the file. Every check a user runs reads the file; the picker renders memory. The reporter pinned it with a two-host comparison: their catalog was rewritten 4m27s after the app-server started, and killing the app-server made the routed models reappear.

`afterCatalogWriteHandleAppServers()` already exists, but only runs from explicit `ocx sync` / `sync-cache`. Startup never called it.

**The obvious fix is unsafe, and that shaped this one.** That handler's `restart: true` branch SIGTERMs matching app-servers and says in its own log line that active turns may be interrupted. Wiring it into an unattended boot would kill a user's in-flight turn on every login, repair, or update. A human typing `--restart-codex` is consenting to that; a boot is not.

So this adds a warning-only counterpart that **cannot** reach the signal path — asserted by an injected `kill` that fails the test if it is ever called. It also asks the existing mtime classifier whether an app-server is genuinely stale, rather than warning about any running one, so a boot with Codex open and a current catalog stays quiet.

**Two startup sites write**, and neither warns on its own:

1. `startServer()` invalidates the models cache during boot.
2. `handleStart` runs the catalog sync moments later.

`handleStart` ORs their results and warns **once**, after both. Warning after the first would read a catalog mtime the second is about to move, and two independent calls would warn twice. Making that observable required typing the startup sync seam, which had been discarding `syncModelsToCodex()`'s result and returning a bare boolean meaning "the integration was enabled" — so "a write happened" was not observable at the startup boundary at all.

The memoized catalog state is dropped before the check. It caches for 5s on the default path, so a `fresh` reading taken before the write would otherwise be replayed after it and this would stay silent about the staleness it just created.

Failure is swallowed throughout: startup synchronization is best-effort and must not stop the proxy from coming up.

Closes #1046.

## Verification

```console
$ bun run typecheck
(clean)

$ bun test tests/codex-app-server-processes.test.ts tests/codex-desired-state.test.ts tests/codex-catalog-sync-hardening.test.ts tests/codex-catalog-writer.test.ts tests/service-lifecycle.test.ts
 74 pass, 1 skip, 0 fail

$ bun run privacy:scan
Privacy scan passed
```

Red-green, per branch rather than per suite:

- Warning on any state instead of only `stale` fails the quiet-path test.
- Removing the `catch` fails the swallowed-failure test.
- The injected `kill` assertion passes only because no startup path reaches `restartCodexAppServers()`.

One honest limit recorded rather than papered over: the 5s memo is only consulted when every io field is defaulted, so an io-injected test bypasses it and cannot reproduce the masking. Rather than write a test that looks like proof and cannot fail, the memo/invalidation mechanism the fix depends on is asserted directly.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added startup detection for stale Codex app servers after synchronization.
  * Displays a warning when stale processes are detected.
  * Startup synchronization now reports cache and catalog update status.

* **Bug Fixes**
  * Prevented startup failures when process discovery or synchronization checks encounter errors.
  * Avoided warnings for fresh, inactive, or unknown processes.

* **Tests**
  * Added coverage for stale-process warnings, cache invalidation, write tracking, and graceful failure handling.

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -317,7 +317,16 @@ async function handleStart(options: { block?: boolean } = {}) {
   installShellHook();
 
   await maybeShowStarPrompt(); // once-only Yes/No GitHub-star prompt on first interactive start
-  await syncCodexOnStartIfEnabled(port, config);
+  const startupSync = await syncCodexOnStartIfEnabled(port, config);
+  // #1046: one warning per startup, after BOTH writes. The server's cache
+  // invalidation happens first and the catalog sync second, so the mtime is only
+  // final here — and neither write site warns on its own, or a boot that hits
+  // both would warn twice.
+  const { consumeStartupCacheInvalidationWrite } = await import("../server");
+  if (consumeStartupCacheInvalidationWrite() || startupSync.catalogWritten || startupSync.cacheSynced) {
+    const { warnIfStaleCodexAppServersAfterStartupWrite } = await import("../codex/app-server-processes");
+    warnIfStaleCodexAppServersAfterStartupWrite({ log: console });
+  }
   if (!currentExternalCodexModelProvider() && !shouldInjectApiAuthHeader(config) && config.syncResumeHistory !== false) {
     historyGuardian = startHistoryMigrationGuardian();
   }
--- a/src/codex/app-server-processes.ts
+++ b/src/codex/app-server-processes.ts
@@ -407,7 +407,9 @@ export function listCodexAppServerProcesses(io: CodexAppServerProcessIo = {}): C
   return matched;
 }
 
-export function formatStaleCodexAppServerWarning(processes: readonly CodexAppServerProcess[]): string {
+export function formatStaleCodexAppServerWarning(
+  processes: readonly { pid: number }[],
+): string {
   const pids = processes.map(process => process.pid).join(", ");
   return (
     `WARNING: ${processes.length} Codex app-server process(es) still running (PID${processes.length === 1 ? "" : "s"}: ${pids}). `
@@ -753,4 +755,45 @@ export function afterCatalogWriteHandleAppServers(
     );
   }
   return { processes, warned: false, restart, hint };
+}
+
+/**
+ * Startup-safe counterpart to {@link afterCatalogWriteHandleAppServers} (#1046).
+ *
+ * Service startup rewrites the catalog and the models cache, but an app-server
+ * that booted earlier keeps an in-memory model list — Codex builds a static
+ * manager from the catalog once and never rereads the file — so the picker shows
+ * a roster that no longer exists on disk. Every check a user runs reads the file;
+ * the picker renders memory.
+ *
+ * Two things this deliberately does NOT do, both of which the `--restart-codex`
+ * path does:
+ *
+ * - It never signals anything. Killing an app-server on an unattended boot would
+ *   interrupt whatever turn the user has in flight. A human typing
+ *   `ocx sync --restart-codex` is consenting to that; a login is not.
+ * - It never warns about a merely-running app-server. It asks the mtime
+ *   classifier whether one is actually stale, so a boot with Codex open and a
+ *   current catalog stays quiet.
+ *
+ * Failure is swallowed: startup synchronization is best-effort and must not stop
+ * the proxy from coming up.
+ *
+ * The memoized state is dropped first. {@link collectCodexAppServerCatalogState}
+ * caches for 5s when every io field is defaulted, so a `fresh` reading taken
+ * before the write would otherwise be replayed after it and this would stay
+ * silent about the very staleness it exists to report.
+ */
+export function warnIfStaleCodexAppServersAfterStartupWrite(
+  options: { log?: Pick<Console, "error">; io?: CodexAppServerProcessIo } = {},
+): { warned: boolean } {
+  try {
+    resetCodexAppServerCatalogStateCache();
+    const status = collectCodexAppServerCatalogState(options.io ?? {});
+    if (status.state !== "stale") return { warned: false };
+    options.log?.error(formatStaleCodexAppServerWarning(status.processes));
+    return { warned: true };
+  } catch {
+    return { warned: false };
+  }
 }
--- a/src/codex/desired-state.ts
+++ b/src/codex/desired-state.ts
@@ -27,7 +27,17 @@ import type { OcxClientIntegrationsConfig, OcxConfig } from "../types";
 export type DurableIntentClientId = keyof OcxClientIntegrationsConfig;
 
 /** Injectable for tests; production passes the real sync. */
-export type CodexStartupSync = (port: number) => Promise<unknown>;
+/**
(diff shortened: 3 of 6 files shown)


lidge-jun:
CI on this PR just reproduced #1061 — and it is not caused by this change.

The macOS leg failed on one test: `native profile OpenCodex process-exit phases > hard OpenCodex process exit after each published transaction phase…`, at `tests/native-profile-crash-boundaries.test.ts:178`. This branch touches `src/codex/app-server-processes.ts`, `src/codex/desired-state.ts`, `src/server/index.ts`, `src/cli/index.ts` and two of their tests. It does not touch that file.

That line is the exact defect #1061 reports and [#1071](https://github.com/lidge-jun/opencodex/pull/1071) fixes: the test waits for the settled file to *exist*, then immediately parses it, so a partially written document satisfies the wait and throws. #1071 replaces that with a wait for parseable content and makes the child publish atomically.

Run: 9018 pass / 8 skip / **1 fail**. The single failure is the flake, on the macOS leg, which is where #1061 says it shows up.

So this is load-dependent CI noise against an unrelated pre-existing defect. Two ways forward, whichever the maintainers prefer:

- land [#1071](https://github.com/lidge-jun/opencodex/pull/1071) first and re-run this leg, or
- re-run the macOS job — it passes on a lucky scheduling.

I have not rebased this branch onto #1071, because chaining an app-server fix behind a test-harness fix would create a dependency that does not exist in the code.

PR #1092, opened 2026-08-06
fix(gui,combos): restore default effort picker and catalog combos

eachann1024:
## Summary

- Fix combo default reasoning effort picker: unknown/empty member ladders no longer empty the selectable set, so effort can be chosen and saved again.
- **Request-time fail closed for unknown ladders** (maintainer feedback): `concreteComboRequestBody` no longer injects `defaultEffort` when the target ladder is `undefined`; debug logs keep distinct `unknown` vs `unsupported`.
- GUI marks options when any target has an unknown effort ladder (label marker + hint) so the picker is not presented as fully known.
- Synthesize incomplete combo catalog members (missing/partial discovery metadata) with provider hints and a 128k context fallback so multi-target combos appear in the model list.
- Clarify Combo ID / public model name copy and show the effective client-facing model id with a one-click copy control.
- **Combo image / multimodal control**: `imageInput: "auto" | "disabled"` (default `auto`, GUI checked when every target supports image). Disabled strips `image` from catalog modalities and rejects image-bearing requests with HTTP 400 before dispatch.

## Change graph

```mermaid
flowchart LR
  GUI[\"Combo workspace effort picker\"] --> Intersect[\"intersectComboEfforts wildcard\"]
  Intersect --> Save[\"PUT /api/combos defaultEffort\"]
  Save --> Runtime[\"concreteComboRequestBody fail-closed\"]
  Targets[\"Member catalog rows\"] --> Resolve[\"resolveComboCatalogMember\"]
  Resolve --> Catalog[\"deriveComboCatalogModel\"]
  Catalog --> Models[\"Public model list\"]
  CapUI[\"imageInput toggle auto default\"] --> CapStore[\"OcxComboConfig.imageInput\"]
  CapStore --> CapCatalog[\"derive strips image when disabled\"]
  CapStore --> CapReq[\"reject image body when disabled\"]
```

## Screenshots

Local ego-browser verification on branch `fix/combo-effort-persist-ux` (ephemeral proxy; not production `:10100`).

Combo effort options + public model preview

## Verification

- `bun test tests/combo-workspace-data.test.ts tests/combos.test.ts tests/codex-catalog.test.ts tests/combo-management-api.test.ts tests/server-combo-failover-e2e.test.ts` (248 pass)
- Rebased onto latest `upstream/dev`
- Local proxy + Vite GUI: effort options, public model preview, image toggle defaults on when members support image

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - Added public model previews with one-click copying and localized feedback.
  - Added Combo image-input capability controls, including automatic support detection and an option to disable image input.
  - Improved Combo naming guidance for internal IDs and client-facing names.

- **Bug Fixes**
  - Retained Combo targets when discovery data is incomplete with safer fallback metadata.
  - Improved effort selection when capabilities are empty or unknown.
  - Image requests are rejected with a clear error when image input is disabled.

- **Documentation**
  - Updated localized guidance and routing documentation for fallback behavior, effort selection, and image input.


eachann1024:
## Follow-up

### Gate
- Rebased onto latest `upstream/dev` (`behind_by: 0`, `ahead_by: 2`).
- Retargeted base **`main` → `dev`** (project `enforce-target` only allows `dev`; merge train still lands on `main` via maintainer promotion).
- Added UI screenshots to the PR description.

### Review feedback addressed in `422f8eba`
- Header combo copy chip now uses shared `useCopyFeedback` (copied / unavailable).
- Combo member `128k` context fallback is clamped by `providerContextCaps` and sets `contextCapped` when lowered.
- Docs updated for optimistic unknown-ladder `defaultEffort` injection and catalog context fallback.
- Kept intentional synthesis for incomplete discovery / LiteLLM-style custom ids (product fix); pure ghost recovery left as-is with code comment.

### Local verification
- `bun test tests/codex-catalog.test.ts tests/combo-workspace-data.test.ts tests/combos.test.ts` → **178 pass**
- ego-browser: effort options, save+reload persist, public model preview `客户端请求：loop`, combo models in catalog; attention only flags legitimate `tran` single-target.

lidge-jun:
Thanks for this — the UI half is a real bug and your diagnosis of it is right. `intersectComboEfforts` treating an unknown ladder as an empty set means one undiscovered member empties the whole picker, and a user cannot then select an effort they are entitled to. That part should land.

The runtime half is where I have to ask for a change.

**What the diff reverses.** In `src/combos/request.ts` the guard becomes:

```ts
if (targetReasoningEfforts !== undefined && !targetReasoningEfforts.includes(defaultEffort)) {
```

so an `undefined` ladder now falls through and the default effort is injected. On `dev` that same line fails closed for both cases, and the debug record deliberately distinguishes them:

```ts
capability: targetReasoningEfforts === undefined ? "unknown" : "unsupported",
```

That ternary is the tell. Someone specifically wanted "we have no metadata" to be separable from "the provider said no" — the two states were already understood as different, and the current behavior is a decision rather than an oversight. The diff collapses them to the literal `"unsupported"`, so the one signal that would tell us how often unknown targets appear in real combos disappears at the same moment we start sending to them.

**The concrete failure mode.** A combo containing a target with no discovery metadata can now save `high` and send `reasoning.effort: "high"` upstream with no evidence the endpoint accepts a reasoning field at all. The user changed nothing; their combo simply included a member whose catalog row was thin. The `COMBO_MEMBER_CONTEXT_FALLBACK` synthesis compounds this by presenting those unknown members as usable, so the picker becomes more confident precisely where we know least.

I want to be fair to the reasoning in your comment: you are right that incomplete catalog metadata is more common than a genuine lack of support. That is an argument for improving discovery or for letting a user assert the capability — not for the proxy assuming it. When the assumption is wrong the user gets a 400 they cannot explain, and the failure surfaces far from this line.

**What I am asking for.** Keep `unknown` as its own state end to end instead of promoting it to supported:

1. Land the picker fix as written — a wildcard member should not empty the intersection. That is unambiguously correct.
2. In the GUI, render an unknown-capability target as unknown (disabled with a hint, or a marker on the option) rather than as an ordinary selectable rung.
3. Leave `src/combos/request.ts` failing closed for `undefined`, and keep the `unknown` / `unsupported` discrimination in the debug record.

If you want the optimistic behavior available, the shape that fits this codebase is an explicit per-model capability assertion — the same thing `modelReasoningEfforts` already does for a provider the user knows. That way the user opts in, and the proxy never guesses on their behalf.

Two smaller notes:

- The i18n hint you added across six locales ("targets without catalog effort metadata do not restrict the list") describes the picker accurately and can stay as is under the change above.
- The `enforce-target` failure on this PR is a stale-checkout artifact, not your code. Rebase onto current `dev` and it clears.

Happy to look again quickly once the runtime guard is restored — the UI fix is worth landing and I would rather it not wait.

eachann1024:
## Follow-up (maintainer review + imageInput)

Addressed @lidge-jun's request on runtime defaults:

1. **Picker** — kept as written: unknown/empty ladders are wildcards and do not empty the intersection.
2. **GUI** — options now carry an **unknown-targets** marker + hint when any complete member has no catalog ladder; runtime still fails closed for those members.
3. **`src/combos/request.ts`** — restored fail-closed for `undefined` ladders; debug record again distinguishes `unknown` vs `unsupported`.

Also shipped the optional capability assertion shape for multimodal:

- `imageInput?: "auto" | "disabled"` on combos (default **`auto` / checked** in the workspace when every target advertises image).
- Catalog derivation strips `image` when disabled; image-bearing requests get HTTP 400 before any target is called.
- Only explicit `"disabled"` is persisted (sparse config).

### Verification
- `bun test tests/combo-workspace-data.test.ts tests/combos.test.ts tests/codex-catalog.test.ts tests/combo-management-api.test.ts tests/server-combo-failover-e2e.test.ts` → **248 pass**
- Rebased onto current `upstream/dev` (`behind_by: 0`).

lidge-jun:
You turned this around in under half an hour, and the fix is exactly right. Recording that plainly, because a fast correct response deserves to be acknowledged rather than met with silence.

The guard is back:

```ts
  // Fail closed for both unknown (`undefined`) and known-but-missing ladders.
  // Unknown is kept distinct in debug so we can measure thin catalog rows without
  // guessing the provider accepts a reasoning field.
  if (!targetReasoningEfforts?.includes(defaultEffort)) {
```

Keeping `unknown` distinct in the debug record is the part I care most about — that is the signal that tells us how often thin catalog rows actually show up in real combos, which is the evidence that would justify revisiting this later. And your GUI notice now describes true behavior: options stay selectable, runtime omits the default for unknown targets. That is the honest version of the trade-off, and it was your copy, not mine.

Two things left, neither of them about the picker.

**1. Please split the `imageInput` work into its own PR.** This branch now also adds a combo `imageInput: "auto" | "disabled"` field with validation, the public-model preview and copy-to-clipboard UI, and locale keys for all of it. Each of those is defensible on its own, but together they mean a reviewer with combo-effort context and a reviewer with combo-capability context are the same person, and whichever half raises a question holds up the other. The effort-picker fix is ready; the rest should not ride on it.

**2. One case worth double-checking in the intersection.** Make sure a *known but empty* ladder still constrains the result. Skipping on `listed.length === 0` as well as `undefined` would conflate "this provider advertises no efforts" with "we have no metadata for this provider" — which is the same conflation the runtime guard just refused to make. Unknown should be the wildcard; known-empty is information and should narrow the set to nothing.

On the mechanics: your branch is now on `dev` and the readiness gate wants at most 10 commits behind with green `ci`, so rebase before ticking the boxes. And the `enforce-target` failure is a repository-side gate artifact, not your code — a stale head's workflow YAML calling into base-pinned scripts. Don't chase it.

Good work. The picker bug was real, your diagnosis of `intersectComboEfforts` was exact, and you fixed the runtime objection without argument.

lidge-jun:
Thank you — the effort-picker fix from this PR is landing as **#1151**, credited to you via `Co-authored-by`.

You found a real usability defect: a model whose capability ladder is unknown vanished from the picker entirely rather than being offered without a forced default. The stack takes that fix, with the boundary made explicit — an unknown ladder is a wildcard **for picker availability only**, while the runtime still omits the default unless the concrete target advertises support. Tests pin both halves, since "permissive picker, fail-closed wire" is the whole point.

**What I did not take**, so this is not a silent reduction: the catalog fallback synthesis, the public-name copy redesign, `imageInput`, and the locale changes. Those are three separate features riding with one bug fix across 32 files, and each deserves its own review. Your `src/combos/index.ts` and `types.ts` hunks were `imageInput`, so those files are untouched in the stack.

Small note for a future PR: the branch fails `git diff --check` on an added blank line at EOF — worth fixing since CI hygiene checks it.

Verified: 165 pass / 0 fail on the combo and config-export suites; full suite 9,556 pass / 0 fail; GUI lint and build clean.

Leaving this PR open for you.

eachann1024:
Thanks for the clear guidance — especially extracting the effort-picker fix into **#1151** and asking to split the rest.

### Follow-up done

Per your comments on this PR:

1. **Split remaining work** (no longer riding together):
   - **#1163** — catalog member synthesis + 128k / `maxInputTokens` fallback, context caps, disabled-provider skip, unknown ladder as catalog wildcard (empty ladder still restrictive)
   - **#1164** — public model name preview + shared copy feedback
   - **#1165** — `imageInput: "auto" | "disabled"` (catalog strip, pre-dispatch 400, `previous_response_id` expansion before image check, sparse persistence, GUI switch)

2. **Known-empty ladder** — handled correctly in **#1151** (`undefined` = wildcard for picker only; `[]` still restricts). Not re-landed here.

3. **`git diff --check` blank line at EOF** — avoided on the new branches.

4. Review findings addressed where they still apply on the split branches (context cap / maxInputTokens / disabled complete members / previous_response expansion for disabled images / `ModelOption.inputModalities` / `comboImagesSupported` tests / docs table rows).

Closing this PR in favor of the three focused ones above. Happy to adjust any of #1163–#1165 if a boundary should move.

eachann1024:
Superseded by focused follow-ups #1163 (catalog fallback), #1164 (public model name), #1165 (imageInput). Effort picker already extracted to #1151.

lidge-jun:
Confirming for the record: your effort-picker fix is on `dev`, credited via `Co-authored-by: Eachann <43348055+eachann1024@users.noreply.github.com>`, carried by #1151 (merge commit `f2e5a883c`).

You found a real usability defect: a model whose capability ladder is unknown vanished from the picker entirely rather than being offered without a forced default. #1151 has a live before/after screenshot showing the difference on the same target — 1 option before, 7 after.

The boundary is explicit in the landed version: an unknown ladder is a wildcard **for picker availability only**, while the runtime still omits the default unless the concrete target advertises support. Tests pin both halves.

**Not taken:** the catalog fallback synthesis, the public-name copy redesign, `imageInput`, and the locale changes — three separate features riding with one bug fix across 32 files, each deserving its own review.

You closed this yourself — thank you. Verified: 9,556 pass / 0 fail; CI green on the exact `dev` SHA.

PR #1093, opened 2026-08-06
fix(telemetry): persist ingress spans and ordinary request attempts

yamashirotakashi:
## Summary

- Capture only guard-issued `x-opencodex-ingress-span` values on ordinary Responses requests and carry them through request and usage logs.
- Create exactly one durable attempt after final initial adapter resolution for ordinary non-combo requests, preserving combo and retry ownership.
- Preserve explicit `attempts: []` while omitting malformed non-array values; reject malformed or secret-shaped ingress values.

## Scope

- Base: `99440ecd91a951e5b580fdde843998d99181400a`
- Files: four source files and three focused test files
- No provider fallback, credential forwarding, retry behavior, or public API contract changes.

## Validation

- `bun test tests/request-log.test.ts tests/usage-log.test.ts tests/server-auth.test.ts`
- Result: 147 passed, 0 failed, 802 expectations
- `git apply --check` against a clean pinned base: passed

## Security

Ingress values are accepted only in the positive issuer shape. Representative `gho_`, `github_pat_`, `Bearer`, and `sk-` values are rejected. Logs persist correlation metadata only, never raw content or credentials.

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added request correlation tracking for valid ingress spans across request and usage logs.
  * Added durable attempt tracking for standard requests.
  * Preserved explicitly empty attempt metadata in persisted logs.

* **Bug Fixes**
  * Rejects malformed or sensitive-looking correlation values.
  * Prevents duplicate upstream requests in passthrough streaming scenarios.
  * Ensures invalid persisted correlation data is omitted.

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -90,6 +90,7 @@ import {
   hydrateRequestLogsFromDisk,
   httpStatusForRequestLogTerminal,
   httpStatusForTerminalStatus,
+  ingressSpanFromHeader,
   inspectResponseLogSsePayload,
   nextRequestLogId,
   recordFirstOutput,
@@ -841,11 +842,13 @@ export function startServer(port?: number, deps: StartServerDeps = {}) {
         }
         const start = Date.now();
         const requestId = nextRequestLogId(start);
+        const ingressSpan = ingressSpanFromHeader(req.headers.get("x-opencodex-ingress-span"));
         const logCtx: RequestLogContext = {
           model: "unknown",
           provider: "unknown",
           ...admissionFields(admission),
           inboundProtocol: "responses",
+          ...(ingressSpan ? { ingressSpan } : {}),
         };
         let logged = false;
         const finalizeNativePassthroughLog = (
--- a/src/server/request-log.ts
+++ b/src/server/request-log.ts
@@ -36,9 +36,20 @@ import {
 import { matchesLogConversationId } from "./request-log-conversation";
 import { enforceAppOwnedMemoryBudget, type RetainedStoreSnapshot } from "../lib/app-owned-memory";
 
+const INGRESS_SPAN_RE = /^[A-Za-z0-9_-]{24}_[0-9a-f]{16}$/;
+
+/** Accept only the content-free app-server correlation token. */
+export function ingressSpanFromHeader(value: string | null): string | undefined {
+  if (typeof value !== "string") return undefined;
+  const span = value.trim();
+  return INGRESS_SPAN_RE.test(span) ? span : undefined;
+}
+
 export interface RequestLogContext {
   model: string;
   provider: string;
+  /** Validated, content-free app-server ingress correlation token. */
+  ingressSpan?: string;
   /** TTFT: ms from request start to the first non-empty model output delta (WP4, devlog 040). */
   firstOutputMs?: number;
   /** Best-effort chat/session correlation for Logs grouping (#330). Opaque; omit when unknown. */
@@ -106,6 +117,8 @@ export interface RequestLogEntry {
   timestamp: number;
   model: string;
   provider: string;
+  /** Validated, content-free app-server ingress correlation token. */
+  ingressSpan?: string;
   /** TTFT: ms from request start to the first non-empty model output delta; unset for non-streaming/tool-only. */
   firstOutputMs?: number;
   surface?: "claude" | "claude-desktop" | "grok";
@@ -226,6 +239,9 @@ export function requestLogEntryFromPersistedUsage(entry: PersistedUsageEntry): R
     timestamp: entry.timestamp,
     model: entry.model,
     provider: entry.provider,
+    ...(ingressSpanFromHeader(entry.ingressSpan ?? null)
+      ? { ingressSpan: ingressSpanFromHeader(entry.ingressSpan ?? null) }
+      : {}),
     ...(entry.firstOutputMs !== undefined ? { firstOutputMs: entry.firstOutputMs } : {}),
     ...(isKnownUsageSurface(entry.surface) ? { surface: entry.surface } : {}),
     ...(entry.conversationId ? { conversationId: entry.conversationId } : {}),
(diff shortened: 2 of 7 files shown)


lidge-jun:
Retargeted this to `dev` for you, so the wrong-branch gate is cleared and you do not need to reopen anything. The `[WRONG BRANCH]` title prefix was added by the bot — feel free to drop it now.

The content is worth landing: ingress-span propagation and durable per-attempt records do not exist on `dev` today, and nothing merged since has covered them, so this is not duplicated work.

Two things still stand between this and review:

1. **Rebase onto current `dev`.** Your branch is based on `99440ecd9`, and `dev` has moved a long way since — including the v2.10.2 release merge. `src/server/responses/core.ts` and `src/usage/log.ts` in particular have changed underneath you, so please rebase and re-run the suite rather than merging `dev` in.
2. **The readiness checklist** in the description is at 0/4; the PR stays in draft until all four are ticked. The gate verifies two of those claims itself — the head's `ci` check must be green, and the branch must be on the latest `dev` commit or at most 10 behind — so do the rebase first and tick afterwards, otherwise the boxes get unticked automatically.

One design question to answer in the description while you are in there: what happens to an ingress span that is well-formed but arrives on a request that never reaches adapter resolution? Your summary says exactly one attempt is created "after final initial adapter resolution," which reads correct, but the failure path before that point is where telemetry work usually leaks or double-counts, and a reviewer will look there first.

`bun run typecheck` and `bun run test` before you tick — CI runs them on Linux, Windows, and macOS.

lidge-jun:
Thanks for retargeting — the branch is on `dev` now and the wrong-base gate is clear. Reviewing the substance.

Neither behavior exists on `dev` today: `src/server/request-log.ts:219-256` persists no ingress-span field, and `src/server/responses/core.ts:1631-1641` creates no attempt record for ordinary requests. So this is additive, nothing merged since has covered it, and the correlation data is genuinely useful. The serialization tests are real — the round-trip, the malformed-input omission, and the explicit-empty-array preservation each pin distinct behavior.

**The blocker is a missing contract, not a missing test.**

The description says the change captures "only guard-issued" ingress spans. The code accepts any client-supplied header value matching a shape:

```ts
const INGRESS_SPAN_RE = /^[A-Za-z0-9_-]{24}_[0-9a-f]{16}$/;
return INGRESS_SPAN_RE.test(span) ? span : undefined;
```

A regex is a format check, not provenance. Nothing issues these values, nothing signs them, and nothing prevents a caller from sending any conforming string — including one chosen to collide with another caller's span. Once persisted into usage records and `/api/logs`, that data looks authoritative to anyone reading it later. Whether that is acceptable depends entirely on what the field is *for*, and no linked issue or reproduction says.

Three things needed before this can be reviewed on the merits:

1. **Who issues the span, and why is a client-supplied value trustworthy enough to persist?** If a component of ours mints it, name that component and say how a request that did not come through it is treated. If it is genuinely caller-supplied correlation data, say so in the description and drop "guard-issued" — that framing is doing work the code does not support.
2. **What happens before adapter resolution?** The attempt record is created after final adapter resolution, so requests failing at auth or routing produce a span with no attempt. Is that intended, or is the pre-resolution path supposed to record something? This is where telemetry work usually leaks or double-counts, and a reviewer will look there first.
3. **Correct the contract claim.** The description says there are no public API contract changes, but the diff adds fields to persisted usage records and to `/api/logs` responses. Both are consumed by the dashboard and by anyone parsing the log format, so this is a contract change and should be described as one.

Also still worth doing: rebase onto current `dev`. The branch is well behind, and `src/server/responses/core.ts` and `src/usage/log.ts` have both moved. Please rebase rather than merging `dev` in.

For clarity — the `bug` label is on this PR but I could not find the defect it fixes. If this is a capability rather than a bug fix, saying that plainly in the description would help; there is nothing wrong with proposing it as an enhancement.

None of the above is about code quality. The implementation is careful. It just needs someone to state what the data means before we start writing it down permanently.

lidge-jun:
Thank you — the attempt-persistence half of this is landing as **#1139**, credited to you via `Co-authored-by`.

Adopted: creating the ordinary request attempt after final adapter resolution using the existing attempt owner, and preserving explicit empty arrays through normalization so "no attempts" and "not recorded" stay distinguishable. That distinction is genuinely useful and the usage log was poorer without it.

**What I withheld, and why.** The ingress-span half reads a client-supplied correlation header at the public admitted endpoint (`src/server/index.ts`), which means any admitted client can forge a regex-shaped "guard-issued" ingress span. Persisted telemetry that an untrusted caller can shape is worse than absent telemetry, because it reads as authoritative to whoever looks at it later. `src/server/index.ts` and `tests/server-auth.test.ts` therefore have zero diff in the stack PR.

This is not a rejection of the idea. Ingress spans are worth having — they need a trusted producer boundary first, where the span is minted by something the client cannot influence. If you want to pursue that shape, I would review it.

Verified: 85 pass / 0 fail on the affected suites; full suite 9,478 pass / 0 fail; typecheck and privacy scan clean.

Leaving this PR open for you.

lidge-jun:
Closing — the attempt-persistence half of this is on `dev`, credited to you via `Co-authored-by: Takashi Yamashiro`, carried by #1139 (merge commit `cd1c230ab`).

**Adopted:** creating the ordinary request attempt after final adapter resolution using the existing attempt owner, and preserving explicit empty arrays through normalization so "no attempts" and "not recorded" stay distinguishable. That distinction is genuinely useful and the usage log was poorer without it.

**Withheld, and why:** the ingress-span half reads a client-supplied correlation header at the public admitted endpoint (`src/server/index.ts`), which lets any admitted client forge a regex-shaped "guard-issued" ingress span. Persisted telemetry an untrusted caller can shape is worse than absent telemetry, because it reads as authoritative to whoever looks at it later. `src/server/index.ts` and `tests/server-auth.test.ts` have **zero diff** on `dev`, verified.

This is not a rejection of the idea. Ingress spans are worth having — they need a trusted producer boundary first, where the span is minted by something the client cannot influence. **If you want to pursue that shape, I would review it.**

Closing since the adoptable half shipped. Verified: 9,593 pass / 0 fail on the merged tree; CI green on the exact `dev` SHA.

PR #1121, opened 2026-08-06
fix(anthropic): preserve routed response model metadata

### LINKED ISSUE #1117: fix(anthropic): preserve provider-qualified response model metadata
### Client or integration

Codex App

Anthropic

### OpenCodex version

2.10.2

### Endpoint or capability

`/v1/responses` streaming response model metadata and Codex catalog synchronization

### Current behaviour

OpenCodex normalizes a routed Anthropic selector for the upstream request and then reuses the normalized bare ID in the Codex-facing Responses payload. A request for `anthropic/claude-sonnet-5` therefore returns `claude-sonnet-5`. Codex cannot resolve that bare ID against the provider-qualified catalog row, so it may miss the model's long-context and auto-compaction metadata.

### Expected behaviour

The Codex-facing response should retain `anthropic/claude-sonnet-5`. A bare legacy selector should be healed to the provider-qualified response model. Compatibility catalog aliases for existing bare selectors should remain hidden, carry the canonical context metadata, and be removed by restore.

```shell
curl -N http://127.0.0.1:10100/v1/responses \
  -H 'Authorization: Bearer REDACTED' \
  -d '{"model":"anthropic/claude-sonnet-5","input":[{"role":"user","content":[{"type":"input_text","text":"reply OK"}]}],"stream":true,"store":false}'
```

### Actual response or error

```shell
HTTP/1.1 200 OK
event: response.completed
data: {"response":{"model":"claude-sonnet-5", "status":"completed"}}
```

### Upstream documentation

No public upstream specification requires the OpenCodex composite selector. The expected behavior is a concrete Codex client requirement: the response model must resolve to the same provider-qualified model metadata that OpenCodex writes into Codex's catalog.

Preserve the final Codex-facing routed selector before upstream normalization; use it in core, image-loop, and web-search-loop response bridges. For existing bare selectors, synchronize a hidden Anthropic compatibility alias with the canonical context metadata and an owner marker so restore can remove it safely.

Live reproduction on 2.10.2 returned HTTP 200 and completed normally; only the response model identity was wrong.

### Checks

Ingwannu:
## Summary

- Preserve the canonical provider-qualified Anthropic selector in Codex-facing streaming and JSON Responses while keeping the bare model id on the upstream Messages request.
- Thread the client-facing identity through the normal Responses bridge, image bridge, and web-search loop without changing non-Anthropic or OpenAI virtual-model identities.
- Generate hidden, marker-owned bare Anthropic compatibility catalog rows with the canonical context/compaction/capability metadata, and remove those rows safely during sync/restore.
- Document the response-identity and cleanup invariant.

Closes #1117

## Verification

- `bun run typecheck` - passed on latest `dev` (`bbd82e73`).
- `bun test tests/anthropic-response-model.test.ts tests/openai-api-virtual-models.test.ts tests/server-combo-failover-e2e.test.ts tests/images/loop.test.ts tests/web-search.test.ts tests/codex-catalog.test.ts tests/codex-catalog-restore.test.ts` - 291 passed, 0 failed.
- `bun run privacy:scan` - passed.
- Full suite run - 9,227 passed / 10 skipped / 2 failed out of 9,239. The two failures were unrelated local-environment cases: the shim fixture read this host's existing persisted service token instead of its temporary fixture token, and one native-profile timing test passed immediately when rerun in isolation. Required GitHub CI remains the merge authority.

## Checklist


lidge-jun:
Thank you — your diagnosis here was correct and independent, and I want that on the record.

The fix for #1117 is landing as **#1138**. For the implementation I built on #1122, which covers native passthrough response metadata and alias-collision cases that this PR does not — but you identified the same root cause (the final-route rewrite of `parsed.modelId` leaking into client-facing `response.model`) on your own, and the campaign PR credits you for it alongside @giulioleone097.

Duplicate work is a cost you paid, not one you caused; two people finding the same real defect independently is a signal the defect was worth fixing.

Leaving this PR open for you to close or continue.

lidge-jun:
Closing — the defect you diagnosed is fixed on `dev`.

#1117 is resolved by #1138 (merge commit `2a13ed83e`). For the implementation I built on #1122, which also covers native passthrough response metadata and alias-collision cases this PR does not — but **you identified the same root cause independently** (the final-route rewrite of `parsed.modelId` leaking into client-facing `response.model`), and the stack PR credits you for that diagnosis alongside @giulioleone097.

Closing as superseded rather than rejected. Duplicate work is a cost you paid, not one you caused; two people independently finding the same real defect is a signal it was worth fixing.

PR #1124, opened 2026-08-06
fix(codex): skip empty native-profile stage sweeps

### LINKED ISSUE #1120: Native-profile lock deadlock: all native requests 503 with zero profiles registered, and account main recover cannot clear it
### Client or integration

Codex CLI

### Area

### Summary

The native-main-profile lock can enter a state where it is never released, and every native-routed request then fails with `503 server_is_overloaded` / `"Native Codex main profile is switching; retry this request"`.

The lock is held even when **no native profiles are registered at all** (`profiles: []`, `activeProfileId: null`, and an empty `<CODEX_HOME>/.opencodex-native-main-profiles/` directory). The subsystem gates all traffic for a feature that has nothing configured in it.

The documented recovery command, `ocx account main recover`, cannot clear it: that command is a client of the proxy's management API, and the management API refuses while the lock is held. Stopping the proxy releases the lock at the filesystem level, but then every management command fails with `Proxy is not running`. There is no out-of-band break-glass path, so the state is unrecoverable through supported commands.

I expected either that the lock would not be taken when no profiles exist, or that `ocx account main recover` would be able to clear it.

Secondary usability problem: returning `code: server_is_overloaded` makes the Codex CLI render **"Selected model is at capacity. Please try a different model."** That points users at OpenAI quota exhaustion. It took a fair amount of digging to establish the request never left the machine — `durationMs` of 27-45ms and `usageStatus: "unreported"` were the giveaways.

### Reproduction

Not consistently reproducible on demand; it appeared spontaneously and then recurred in bursts. What I observed:

1. `ocx start` with `codexAccountMode: pool`, a single Codex account (`main`, the Codex App login), and no native main profiles registered.
2. Send any request through the proxy to a native model, e.g. `POST http://127.0.0.1:10100/v1/responses` with `{"model":"gpt-5.4-mini", ...}`.
3. Observe `503` returned in single-digit milliseconds with `Native Codex main profile is switching`.
4. Attempt the documented recovery — it fails on the same lock:

```
$ ocx account main doctor --json
Error: The native-profile lock is unavailable.

$ ocx account main recover --json
Error: The native-profile lock is unavailable.

$ ocx account main list --json
{"effectiveCodexHome":"<CODEX_HOME>","activeProfileId":null,"profiles":[]}
```

5. `ocx restart` clears the currently active burst but does not prevent recurrence.

Observed burst pattern over ~3 hours, across a proxy restart. Note the bimodal durations (0-5s vs 211-268s), which look like lock contention rather than a crash:

```
start     end       dur    gap-since-prev-start
06:42:15  06:46:43  268s   -
06:49:04  06:49:07    3s   7m
06:59:20  07:00:37   77s   10m
07:07:48  07:07:51    3s   8m
07:12:55  07:14:22   88s   5m
07:16:34  07:16:52   18s   4m
07:32:43  07:32:58   15s   16m
07:43:55  07:43:58    3s   11m
08:22:51  08:22:54    3s   39m
08:33:04  08:33:09    5s   10m
08:36:08  08:36:38   30s   3m
08:41:16  08:41:16    0s   5m
   --- ocx restart here (08:43) ---
09:38:06  09:41:36  211s   57m
09:44:31  09:44:31    0s   6m
```

One correlation worth checking: a `codex-quota-cache.json` write landed at 09:38:19, 13 seconds after that burst began, which suggests the quota-refresh path takes the same lock.

**Workaround that resolved it for me:** `ocx provider account-mode direct`. With a single Codex account, pool mode adds nothing, and direct mode keeps the profile-switching layer out of the request path. Verified working end-to-end afterwards; no recurrence since.

Note the catch-22 in applying it: setting `account-mode` requires the proxy to be **running**, while clearing the lock requires it to be **stopped**. The window to apply the workaround is whatever clean interval happens to occur.

### Suggested fixes

1. Do not acquire the native-profile lock when no profiles are registered. If `profiles` is empty and `activeProfileId` is null, the subsystem should be inert.
2. Give `recover` an out-of-band path that does not depend on the proxy's management API, e.g. an offline `--force` operating on the lock file directly while the proxy is stopped. A recovery tool gated behind the thing it recovers cannot function.
3. Add a lock lease/timeout so a holder that never completes releases automatically.
4. Use a distinct error code instead of `server_is_overloaded`, or name opencodex in the message, so users are not sent to investigate OpenAI quota.
5. Surface this in `ocx doctor`, which reported `OAuth health: ok` while all traffic was failing.

Candidate source locations from code search: `src/codex/native-profile-manager.ts`, `src/codex/native-profile-store.ts`, `src/codex/native-main-owner.ts`, `src/codex/native-main-claim.ts`, `src/codex/native-profile-startup.ts`, `src/server/management-api.ts`, `src/cli/doctor.ts`.

Possibly a regression from the native-main-login work in #656.

### Version

2.10.1 (also inspected `dev` at v2.10.2; no native-profile changes in that release, so this should still reproduce)

### Operating system

Windows 11 Pro 10.0.26200

### Provider and model

openai (native Codex) / gpt-5.4-mini, gpt-5.6-luna, gpt-5.6-terra — all native-family models failed identically

### Logs or error output

```shell
# usage.jsonl record for a failing request
{
  "provider": "openai",
  "model": "gpt-5.4-mini",
  "admissionKind": "loopback",
  "inboundProtocol": "responses",
  "status": 503,
  "durationMs": 32,
  "usageStatus": "unreported",
  "errorCode": "server_is_overloaded",
  "upstreamError": "Native Codex main profile is switching; retry this request",
  "routeDecision": { "routeKind": "native" }
}

# reproduced directly against the proxy
POST http://127.0.0.1:10100/v1/responses
-> HTTP 503 in 0.0068s
{"error":{"message":"Native Codex main profile is switching; retry this request",
          "type":"server_error","code":"server_is_overloaded"}}

# recovery blocked by the same lock
$ ocx account main doctor --json
Error: The native-profile lock is unavailable.

$ ocx account main recover --json
Error: The native-profile lock is unavailable.

# nothing registered in the subsystem holding the lock
$ ocx account main list --json
{"effectiveCodexHome":"<CODEX_HOME>","activeProfileId":null,"profiles":[]}

# lock file present and actively touched
<CODEX_HOME>/.opencodex-native-profile.lock.sqlite   8192 bytes
<CODEX_HOME>/.opencodex-native-main-profiles/        (empty)
```

_No response_

### Redacted configuration

```json
{
  "providers": {
    "openai": {
      "codexAccountMode": "pool",
      "authMode": "forward"
    }
  },
  "defaultProvider": "openrouter",
  "openaiProviderTierVersion": 2,
}
```

Account pool state at the time: a single Codex account (`main`, Codex App login). `auto-switch: on (threshold 80%)` with actual weekly usage around 47%, so the threshold was never approached.

### Checks

Ingwannu:
## Summary

- Skip native-profile stage sweeping only when both the stage registry and this instance's staging tree are proven absent, so zero-profile installations do not contend on the profile transaction database or fence native-main traffic.
- Preserve the existing locked, fail-closed sweep whenever an artifact exists or its absence cannot be proven.
- Identify the resulting 503 as local OpenCodex native-profile maintenance and document the decision and invariants.

Closes #1120

## Verification

- `env -u OPENCODEX_API_AUTH_TOKEN taskset -c 0-1 bun test tests/native-profile-stage-lifecycle.test.ts tests/native-profile-drain-server.test.ts tests/codex-auth-context.test.ts` — 51 pass, 0 fail.
- `taskset -c 0-1 bun run typecheck` — passed.
- `taskset -c 0-1 bun run privacy:scan` — passed.
- Full suite: 9,417 pass, 10 skip, with one existing load-sensitive failure in `Live/Realtime sideband retains main ownership while Direct and non-main Pool continue`. The failing case then passed seven consecutive isolated/focused executions.

## Checklist

### DIFF
--- a/src/codex/auth-context.ts
+++ b/src/codex/auth-context.ts
@@ -112,15 +112,18 @@ export class CodexPoolAuthenticationError extends Error {
   }
 }
 
+export const CODEX_MAIN_PROFILE_MAINTENANCE_MESSAGE =
+  "OpenCodex local native-main profile maintenance is active; retry this request";
+
 export class CodexMainProfileDrainingError extends Error {
   constructor() {
-    super("Native Codex main profile is switching; retry this request");
+    super(CODEX_MAIN_PROFILE_MAINTENANCE_MESSAGE);
     this.name = "CodexMainProfileDrainingError";
   }
 }
 
 export function codexMainProfileDrainingResponse(): Response {
-  const response = formatErrorResponse(503, "server_busy", "Native Codex main profile is switching; retry this request");
+  const response = formatErrorResponse(503, "server_busy", CODEX_MAIN_PROFILE_MAINTENANCE_MESSAGE);
   const headers = new Headers(response.headers);
   headers.set("Retry-After", "1");
   return new Response(response.body, { status: response.status, headers });
--- a/src/codex/native-profile-manager.ts
+++ b/src/codex/native-profile-manager.ts
@@ -913,6 +913,25 @@ export class NativeProfileManager {
     return { removed, live, cleanupFailed, plaintextMayRemain };
   }
 
+  /**
+   * Whether startup or the periodic cleaner has any stage state to inspect.
+   *
+   * Absence is the only lock-free result. Any entry or observation failure
+   * keeps the existing fail-closed sweep, so an unsafe/unreadable stage path
+   * can never be mistaken for an unused profile subsystem.
+   */
+  stageSweepRequired(): boolean {
+    for (const path of [this.context.stageRegistryPath, this.context.stagingRoot]) {
+      try {
+        lstatSync(path);
+        return true;
+      } catch (error) {
+        if (errorCode(error) !== "ENOENT") return true;
+      }
+    }
+    return false;
+  }
+
   async sweepStages(): Promise<NativeStageSweepResult> {
     return this.withLock(() => this.sweepStagesLocked());
   }
--- a/src/codex/native-profile-startup.ts
+++ b/src/codex/native-profile-startup.ts
@@ -130,6 +130,14 @@ function ownerBlockedReason(owner: NativeMainOwnerSnapshot): "owner-conflict" | 
 
 async function runOwnedStageSweep(entry: StartupEntry): Promise<boolean> {
   if (typeof (entry.manager as Partial<NativeProfileManager>).sweepStages !== "function") return true;
+  // A fresh installation has no stage registry or staging tree. Avoid creating
+  // and contending on the profile transaction database for an inert subsystem.
+  // Real managers treat every present or uncertain artifact as sweep-required;
+  // partial test/library managers without the preflight keep the old behavior.
+  if (
+    typeof (entry.manager as Partial<NativeProfileManager>).stageSweepRequired === "function"
+    && !entry.manager.stageSweepRequired()
+  ) return true;
   try {
     const result = await withNativeMainOwnerOperation(entry.manager.context, () => entry.manager.sweepStages());
     return !result.plaintextMayRemain;
--- a/src/server/claude-messages.ts
+++ b/src/server/claude-messages.ts
@@ -36,6 +36,7 @@ import { responseWithDeferredRequestLog } from "./relay";
 import { handleResponses } from "./responses";
 import type { AdmissionLease } from "../lib/admission";
 import { tryClaimNativeMainProfileForTurn } from "../codex/native-main-admission";
+import { CODEX_MAIN_PROFILE_MAINTENANCE_MESSAGE } from "../codex/auth-context";
 import {
   createTranslatorBudget,
   finalizeTranslatorBudgetResponse,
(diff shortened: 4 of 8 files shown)


lidge-jun:
Thank you — this is landing as **#1134** in the stacked campaign, with your authorship preserved (the commit on that branch is authored by you, cherry-picked, not rewritten).

The absence-only framing is what makes this safe on a credential path: the sweep is skipped only when both the stage registry and the staging tree are provably absent, and anything present-or-unreadable still enters the locked fail-closed cleanup. Your regression test pinning *both* directions is the part that earns the change — a fast path that only proved the happy case would not have been adoptable here.

Verified on the stack head: `tests/native-profile-stage-lifecycle.test.ts`, `tests/codex-auth-context.test.ts`, `tests/native-profile-drain-server.test.ts` — 51 pass / 0 fail; typecheck clean; privacy scan green; full suite passed the prepush gate.

Leaving this PR open for you to close or continue.

lidge-jun:

Your commit is in `dev` with **your authorship intact**: `fix(codex): skip empty native-profile stage sweeps`, authored by `Ingwannu`, carried by #1134 (merge commit `5ba0c1857`).

Closing because the work shipped, not because it was rejected. The "closed" label is an artifact of the commit travelling through the campaign branch; the contribution and the credit are on `dev`.

The absence-only framing is what made this safe to take on a credential path — the sweep is skipped only when both the stage registry and the staging tree are provably absent, and anything present-or-unreadable still enters the locked fail-closed cleanup. Your regression test pinning *both* directions is what earned it.

Verified: 9,593 pass / 0 fail on the merged tree; CI green on the exact `dev` SHA. Issue #1120 stays open for @MarcusNeufeldt to confirm on his own installation.

PR #1126, opened 2026-08-06
fix(responses): make reasoning replay restart-safe and observable (#950)

ZachDreamZ:
## What

Continues the #950 fix (PR #971) on the Codex / `/v1/responses` path: the reasoning replay cache is currently in-memory only, so a proxy restart mid-round still loses recovery for a DeepSeek thinking-mode tool round, and there is no privacy-safe signal when a bare tool-call continuation is about to be serialized.

## Changes

- **Opt-in disk spill** for the replay cache: `OPENCODEX_REASONING_REPLAY_PERSIST=1` (optional `OPENCODEX_REASONING_REPLAY_FILE` override, default `<config dir>/reasoning-replay-cache.json`, mirroring `getConfigDir()` resolution). Bounded by the same entry/byte/TTL caps as memory, written atomically (tmp + rename) with best-effort 0600 perms, rehydrated at boot. **Default stays memory-only** — the privacy stance of #971 is unchanged unless explicitly opted in.
- **Privacy-safe diagnostics** (#950 checklist): `getReasoningReplayStats()` exposes counters and bounds only (entries, bytes, hits, misses, bare-serialization counts per model, persistence state) — never reasoning text.
- **Invariant counter**: `recordBareToolCallSerialization(model)` counts the exact 400 shape (assistant message with `tool_calls`, preserved model, no `reasoning_content` re-attached) in both the compacted-history path and the orphan-repair path, with a throttled `console.warn` containing counters only.
- **Regression tests** (`tests/reasoning-replay-robustness.test.ts`): restart round-trip, TTL filter on reload, corrupt-file tolerance, entry-cap on reload, stats never contain reasoning text, persistence off by default, and the wire-level bare-serialization counter.

## Verification

- `bun test` on the reasoning suites: 45 pass / 0 fail.
- `bun run typecheck` clean; `bun run privacy:scan` passes.
- Local proxy patched with the same change and live-verified against `opencode-free/deepseek-v4-flash-free` (tool-call continuation replay 200, no 400s).

## Notes

- No reasoning text is ever logged or exposed through stats; the spill file is opt-in and bounded.
- Test-only seams keep the memory/privacy defaults intact.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when preserving reasoning across tool-call continuations.
  * Added recovery for missing or expired cached reasoning data.

* **Diagnostics**
  * Added privacy-safe cache statistics and throttled warnings without exposing reasoning content.

* **Performance & Reliability**
  * Added optional persistent caching with expiration, bounded storage, atomic writes, and recovery from corrupted cache data.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -8,7 +8,11 @@ import { isCyberPolicyCode } from "../lib/errors";
 import { redactSecretString } from "../lib/redact";
 import { contentPartsToText } from "./image";
 import { identifyRoutedModel } from "./identity";
-import { peekReasoningForCall } from "../responses/reasoning-replay-cache";
+import {
+  getReasoningReplayStats,
+  peekReasoningForCall,
+  recordBareToolCallSerialization,
+} from "../responses/reasoning-replay-cache";
 import { buildNonOpenAIToolCatalogNudgeForTools, shouldInjectNonOpenAIToolCatalogNudge } from "./tool-catalog-nudge";
 import { openRouterProviderPayload, resolveOpenRouterRouting } from "../providers/openrouter-routing";
 import {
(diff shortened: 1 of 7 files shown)


lidge-jun:
Thank you — the empty-delta fix from this PR is landing as **#1144**, with your authorship preserved on the cherry-picked commits.

The defect was real and worth finding: empty `text_delta` / `thinking_delta` events dropped replay candidates in both the streaming and batch builders, so a restart could not reconstruct the turn.

**What I withheld, and why.** The fix arrived bundled with optional on-disk persistence of chain-of-thought, exit hooks, and global counters. Writing model reasoning to disk is a privacy-surface change, not a bug fix — `src/responses/reasoning-replay-cache.ts` documents a memory-only contract deliberately — and changing that contract deserves its own PR where the question can actually be discussed. In the stack PR, `src/config.ts`, `src/lib/config-dir.ts` and `src/adapters/openai-chat.ts` have zero diff, and a subprocess test proves that setting the persistence environment variables produces no file or directory.

This is not a rejection of persistent replay. It is a real feature request with a real privacy question attached; it just should not land as a side effect of a delta-handling fix.

Verified: 32 pass / 0 fail; full suite 9,534 pass / 0 fail.

Leaving this PR open for you.

lidge-jun:
Closing — the empty-delta fix from this PR is on `dev`, with your authorship on the carried commit plus a `Co-authored-by: NexusCore` trailer, via #1144 (merge commit `a260006e3`).

The defect was real and worth finding: empty `text_delta` / `thinking_delta` events dropped replay candidates in both the streaming and batch builders, so a restart could not reconstruct the turn.

**Withheld, and why:** the fix arrived bundled with optional on-disk persistence of chain-of-thought, exit hooks, and global counters. Writing model reasoning to disk is a privacy-surface change, not a bug fix — `src/responses/reasoning-replay-cache.ts` documents a memory-only contract deliberately — and changing that contract deserves its own PR where the question can be discussed. On `dev`, `src/config.ts` and `src/lib/config-dir.ts` carry no such change (the latter does not exist), and a subprocess test proves that setting the persistence environment variables produces no file or directory.

This is not a rejection of persistent replay. It is a real feature request with a real privacy question attached; it just should not land as a side effect of a delta-handling fix. **If you want to raise it on its own, I would review it.**

Closing since the adoptable half shipped. Verified: 9,593 pass / 0 fail on the merged tree; CI green on the exact `dev` SHA.

Issue #1143, opened 2026-08-06
Luna compatibility policy bypassed by web-search loop forcing upstream streaming


emanuelmustea:
## Summary

OpenCodex can successfully work around providers whose Responses/chat streaming is unreliable by setting a per-model upstream-streaming policy to false. However, the web-search loop currently forces `stream:true` for each internal model iteration, bypassing that policy.

This caused GPT-5.6 Luna through OpenCode Go to keep failing in Codex Desktop even after the provider registry and main Responses normalization paths were configured for buffered upstream requests.

## Environment

- OpenCodex 2.10.2
- macOS
- Codex Desktop
- Provider adapter: `openai-chat`
- Provider: OpenCode Go
- Model: `gpt-5.6-luna`

## Reproduction

1. Configure a routed model with a provider registry entry such as:

```ts
modelResponsesUpstreamStreaming: { "gpt-5.6-luna": false }
```

2. Send a Codex Responses request with `stream:true`, tools, and hosted `web_search_preview` enabled.
3. The normal normalization path changes the upstream request to non-streaming.
4. `src/web-search/loop.ts` creates an iteration with `stream: true` and calls the adapter with `stream: true`.
5. The provider receives a streaming tool request and fails; OpenCodex returns the truncation error/502.

Without the hosted web-search tool, the buffered compatibility path works. With it, the loop reintroduces upstream streaming.

## Expected behavior

The web-search loop should preserve the effective upstream streaming decision for the routed provider/model. A client may still receive downstream SSE synthesized from a completed buffered response, while the upstream request remains non-streaming.

## Fix direction

- Do not unconditionally set `iterParsed.stream = true` in the web-search loop.
- Pass the effective stream setting to `fetchResponse`.
- Use `parseResponse` for buffered JSON iterations and `parseStream` only for genuinely streaming upstream responses.
- Keep the client-facing web-search loop and SSE behavior unchanged.

This report intentionally excludes API keys, private prompts, and workspace/tool payloads.

PR #1155, opened 2026-08-06
fix(web-search): preserve buffered upstream policy

myrosla:
## Summary

- Preserve the routed model's effective upstream streaming setting inside the web-search loop.
- Parse buffered upstream JSON with the adapter's `parseResponse` path while keeping the client response as Responses SSE.
- Enforce the existing 32 MiB turn cap before a buffered adapter can materialize an oversized body.
- Cover a real two-iteration loop: routed tool call, hosted sidecar search, then buffered final answer.
- Keep architecture and localized sidecar documentation aligned.

Refs #1143

## Scope note

Public `dev` currently has no OpenCode Go / GPT-5.6 Luna entry in `modelResponsesUpstreamStreaming`. Its only `false` entry is DeepSeek Flash on the native `openai-responses` passthrough path, which does not enter the web-search loop. This PR fixes and tests the loop-level transport contract, but intentionally does not add unverified Luna provider metadata or claim an end-to-end reproduction of that provider route.

## Verification

- `bun test tests/web-search.test.ts tests/web-search-progress-stream.test.ts` — 72 pass, 0 fail.
- `bun run test` — passed on commit `307045c5`.
- `cd docs-site && bun run build` — 221 pages built.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Routed web-search requests now consistently follow each route’s streaming policy.
  - Compatibility routes use bounded non-streaming processing, while streaming routes continue to stream.
  - Oversized upstream responses are rejected safely, with cancellation to prevent excess buffering.
  - Client responses remain available in Responses SSE format, without exposing intermediate search output.

- **Documentation**
  - Updated sidecar and transport guides across supported languages to describe streaming behavior, buffering, response limits, and error handling.


lidge-jun:
Correcting an assessment I made and got wrong.

I had this queued to close as unreachable: no production registry entry opts into `modelResponsesUpstreamStreaming`, and the DeepSeek force was retired in `0b8e608c0`. That part is accurate but it is not the whole change.

The hook that matters is `upstreamStreaming: parsed.stream` — the caller decides, not the registry. `stream` is optional in the public Responses schema (`src/responses/schema.ts:133-144`), and `planWebSearch()` imposes no streaming precondition (`src/web-search/index.ts:148-150`). So a routed `/v1/responses` request using `web_search` with `stream` omitted or false lands exactly on the buffered branch this PR preserves. Your own test invokes it deliberately.

Withdrawing the unreachable rationale. The path is reachable and worth keeping.

It is not merge-ready as it stands, though. The buffered `openai-responses` path calls a compaction-only parser that errors on a tool-call-only response (`src/adapters/openai-responses.ts:1271-1293`), which matches the unresolved automated review feedback. Could you address that case and the retained-event accounting, or split them out? Happy to take it once that is handled.

lidge-jun:
Correcting my own correction — one detail above is wrong.

I wrote that a request with `stream` **omitted** reaches the buffered branch. It does not. Your `runWithWebSearch` does `const upstreamStreaming = deps.upstreamStreaming ?? true`, so an absent `stream` resolves to streaming. Only an explicit `stream: false` reaches the buffered path.

The substance of the withdrawal stands: the branch is reachable from a public request, so closing this as an unreachable path would have been wrong. It is narrower than I stated — explicit `stream: false` with `web_search`, not merely an omitted field.

Sorry for the noise. The merge-readiness point is unchanged: the buffered `openai-responses` path still needs the tool-call-only response case handled.

lidge-jun:
Status update, since this one has been sitting without any signal at all.

**No CI has ever run on this PR.** Not failed, not cancelled — zero workflow runs against `307045c55`, and zero for the branch in this repository's Actions history. The only status on the head commit is CodeRabbit. I swept every open PR today and approved the pending workflow runs on all of them; this is the one where there was nothing to approve.

The likely cause is that this comes from a fork (`myrosla/opencodex`) and the first-time-contributor approval gate never produced a queued run I can release from the API side. I could not find one to approve, so I am not able to unblock it from here the way I did for the others.

Two things that would move it:

1. Pushing any commit to the branch usually re-triggers the workflows and creates a run that can be approved — even an empty commit or a rebase onto current `dev` would do it. The branch is 429 commits behind, so a rebase is worth doing regardless.
2. If a run does appear and sits at `action_required`, tell me and I will approve it against the exact head SHA.

Separately, my two comments above stand as the substantive review: the first retracted a wrong close-as-unreachable assessment, and the second corrected an error in that retraction about which branch a `stream`-omitted request reaches. The patch itself is not blocked on my analysis — it is blocked on never having been tested by CI.

No action needed from you if you would rather leave it; say so and I will close it as stale with the record intact. But I would rather see it rebased and actually run.

lidge-jun:
Thanks for going after the forced `stream: true` behavior — the underlying problem is real: `src/web-search/loop.ts:364-365,411` forces streaming back on, and an explicit public `stream: false` reaches the loop because `src/web-search/index.ts:148-150` has no streaming precondition.

I applied both your commits onto current `dev` and they merge cleanly, with the focused web-search suites passing. But there are correctness blockers I cannot fix from here, because `maintainerCanModify` is false on this branch, so the corrections have to come from you:

1. **The integration test uses `openai-chat`, which masks the real failure.** A buffered `openai-responses` payload containing only a function call currently ends as `response.failed` with `upstream compaction returned no summary text`, because `openai-responses.parseResponse` is compaction-specific rather than a general buffered Responses parser. Please add a regression on the actual Responses route: a first buffered iteration with a `web_search` function call, then a later buffered iteration with the final answer.

2. **Implement the matching buffered Responses parsing**, or stop the loop from routing buffered Responses payloads through the compaction parser. Assert tool-call extraction, final text, usage, terminal status, and downstream SSE framing.

3. **Release translator-budget ownership for intercepted events.** The synthetic `tool_call_start`/`tool_call_delta`/`tool_call_end` events are removed by the scanner and never reach the bridge's normal `releaseTranslatedEvent` path — measured 148 retained bytes after one intercepted call. Please add a regression asserting retained bytes return to baseline, including error and cancellation paths.

4. **Fix the sidecar docs (English and localized).** They say semantic events are fully buffered, but commentary-phase text is deliberately forwarded live at `src/web-search/loop.ts:562-568`.

5. **Rebase onto current `dev`** — the branch is ~546 commits behind — and run typecheck plus the full suite so substantive CI can evaluate the head.

Keeping this open rather than closing it: the core idea of preserving the route's effective upstream streaming policy is sound, and the list above is concrete. Ping me when you have pushed and I will re-review.

Wibias:
Since this is to far behind now and there hasnt be any movement I am going to close this. Thanks for the contribution.

PR #1159, opened 2026-08-06
fix(cursor): preserve Grok wire model prefix

jonathanli12:
## Summary

- preserve Cursor's required `cursor-` prefix for regular Grok 4.5 outbound wire IDs
- keep discovery matching on canonical prefix-free IDs
- add low, medium, and high coverage plus a discovery-to-request invariant

## Validation

- focused Cursor tests: 46 passed
- typecheck: passed
- privacy scan: passed
- full suite: 9,457 passed, 8 skipped, one unrelated timeout; isolated rerun passed 20 tests
- git diff check: passed

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected Cursor Grok 4.5 request identifiers for low, medium, and high effort levels.
  - Preserved the expected request format for Grok 4.5 Fast, including its separate effort and fast parameters.
  - Improved compatibility with live model discovery identifiers across supported effort levels.

- **Documentation**
  - Clarified the different model ID and effort parameter formats used by regular and fast Grok 4.5 tiers.

## Review readiness checklist

### DIFF
--- a/docs-site/src/content/docs/reference/adapters.md
+++ b/docs-site/src/content/docs/reference/adapters.md
@@ -181,8 +181,9 @@ advertised effort control on those models as proof of upstream-native reasoning 
 - Exposes Cursor Router as `cursor/auto` plus explicit `cursor/auto-cost`,
   `cursor/auto-balance`, and `cursor/auto-intelligence` entries. Explicit levels are encoded in
   `requested_model.parameters` while the legacy `cursor/auto` entry retains the account/team default.
-- Keeps `cursor/grok-4.5-fast` as a selectable model while sending Cursor's canonical `grok-4.5`
-  model with separate `effort` and `fast=true` parameters.
+- Sends regular `cursor/grok-4.5` tiers with Cursor's exact live-discovery wire ids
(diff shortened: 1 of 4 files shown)


lidge-jun:
Closing in favor of #1208, which adopts this change as-is.

To be clear about why: **this PR needed no changes.** It was independently re-audited against current `dev` and confirmed correct — the request-only helper is the right seam, discovery normalization is properly left alone, and Grok Fast keeps its parameterized form. We are rebuilding it on our stack per the sequential-PR convention, not because anything was wrong with it. You are credited in the commit.

Worth saying plainly: this sat unmergeable because 524 workflow runs were queued awaiting maintainer approval, 39 of them on open-PR branches including this one. The readiness gate verifies the `ci` check, and a run that was never allowed to start has no result to verify — so this could not leave draft regardless of what you did. That was our backlog, not your PR.

Two notes carried into #1208: the fixture pins against recorded discovery output rather than live Cursor state, so it catches mapper drift but not a future rename without a manual refresh; and the change prefixes only regular Grok 4.5, so it does not address #1162.

Thanks — this was a clean fix.

PR #1130, opened 2026-08-06
fix(codex): retry native-main ACL timeout once

luvs01:
## Summary

- Preserve bounded ACL error codes so the native-main owner can identify a real `ETIMEDOUT` from permanent permission failures or untrusted message text.
- Let the native-main lock owner consume exactly one fresh-budget recovery after a process-memoized ACL timeout, then remain fail-closed with `EACLRETRYEXHAUSTED` instead of restalling indefinitely.
- Keep ordinary ACL callers on the existing anti-restall behavior; only the owner with its own single-flight retry policy can consume the recovery.
- This is a focused follow-up to #160, #612, and #766. It is adjacent to, but independent from, the empty stage-sweep fix in #1120 / #1124.

## Verification

- Final head, Bun 1.4.0-canary.1: 6 targeted recovery regressions passed, 160 filtered, 0 failed.
- Final head, Bun 1.3.14: the same 6 targeted recovery regressions passed, 160 filtered, 0 failed.
- Before the final unrelated `dev` refresh, both full affected files passed on both runtimes: 166 passed, 0 failed.
- `bun run typecheck` — passed.
- `bun run privacy:scan` — passed.
- `git diff origin/dev...HEAD --check` — passed.
- Live Windows validation on two OpenCodex 2.10.2 hosts recovered native-main ownership and restored a visible, usable main account while preserving fail-closed behavior.
- A combined pre-split full-suite run was attempted on Windows Canary: 8,863 passed, 9 skipped, 275 failed, and 12 errors. The failures were dominated by existing environment/platform cases (Windows account lookup, symlink privilege, EBUSY cleanup, missing dashboard dependencies, and child `bun` PATH), so this PR does not claim a green full local suite; public CI remains required.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults.

No reusable credential, ACL bypass, or weaker default is introduced. Maintainer review is still required before merge.

## Review readiness checklist

This PR stays in draft until every box below is ticked. Tick all four boxes once the requirements are met:

- [ ] All CI tests are green on my local testing.
- [ ] I pushed my PR to the latest dev commit.
- [ ] I resolved all correct Codex and CodeRabbit findings.

- [ ] My PR is ready for review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows lock-file permission handling by retrying once after a temporary timeout.
[...]

PR #1180, opened 2026-08-07
fix(windows): grant secret ACLs to effective token SID

### LINKED ISSUE #1149: [Bug][Windows] ACL hardening trusts USERDOMAIN and rejects workgroup local accounts
### Client or integration

Other — OpenCodex proxy/service startup and file-backed management authentication on Windows

### Area

Platform (Windows / macOS / Linux)

### Summary

On a Windows workgroup host, `USERDOMAIN` can be the literal `WORKGROUP` while the effective token belongs to the local computer account (for example, `COMPUTER\\user`). The current `src/lib/windows-secret-acl.ts` builds the owner ACE from `USERDOMAIN\\USERNAME`, so it asks `icacls` to grant an unmappable `WORKGROUP\\user` principal.

On the reproduced host, that grant exits with code 52. Required ACL hardening then correctly fails closed, but native-main ownership or file-backed management authentication becomes unavailable even though the NTFS volume and the effective account are valid.

Expected: derive the ACL principal from the effective Windows token, preferably its locale-independent SID, rather than from environment strings.

### Reproduction

1. On a Windows machine joined only to a workgroup, compare the environment-derived principal with the effective token:
   ```powershell
   $environmentPrincipal = "$env:USERDOMAIN\$env:USERNAME"
   $token = [Security.Principal.WindowsIdentity]::GetCurrent()
   $environmentPrincipal
   $token.Name
   ```
2. Create a temporary file and run the same owner grant shape used by `currentWindowsUser()`:
   ```powershell
   $probe = New-TemporaryFile
   icacls.exe $probe /grant:r "${environmentPrincipal}:(F)"
   $environmentExit = $LASTEXITCODE
   ```
3. On the affected host, the environment principal is `WORKGROUP\\user` and the command exits 52.
4. Grant the effective token SID instead:
   ```powershell
   icacls.exe $probe /grant:r "*$($token.User.Value):(F)"
   $sidExit = $LASTEXITCODE
   Remove-Item -LiteralPath $probe
   ```
5. The SID grant exits 0. The same distinction reproduces through OpenCodex's required ACL hardening path.

### Version

2.10.2; also present on current `dev` at `e9d957bf67d05cdb09c84dbac8d436a349d308f1`

### Operating system

Windows 11, build 26200

### Provider and model

Not provider-specific

### Logs or error output

```shell
environment principal: WORKGROUP\\<redacted-user>
effective token name: <redacted-computer>\\<redacted-user>
environment-principal icacls exit: 52
effective-token SID icacls exit: 0
```

No screenshot is required; the reproduction above uses a disposable temporary file and redacts account identifiers.

### Redacted configuration

```json
{
  "configuration_dependency": "none"
}
```

### Checks

### Root cause and proposed boundary

`currentWindowsUser()` in `src/lib/windows-secret-acl.ts` treats `USERDOMAIN\\USERNAME` as an account authority. Those variables are not an authoritative representation of the effective token.

A focused fix can:

- resolve `WindowsIdentity.GetCurrent().User.Value`;
- validate and pass the SID to `icacls` as `*S-1-...:(F)`;
- launch trusted System32 PowerShell hidden, non-interactive, and within the existing ACL deadline;
- process-cache only a successful SID lookup;
- classify lookup failure separately from an `icacls` timeout so the destination timeout memo is not polluted.

This is distinct from #160/#612/#766 (timeout, async waiting, and fail-closed availability) and #596/#601 (owner-ACE ordering). PR #998 introduced token-SID identity for Codex coordinator state, but the ACL principal path still independently uses the environment-derived name.

luvs01:
## Summary

- resolve the effective Windows token SID instead of constructing the ACL owner from `USERDOMAIN\\USERNAME`;
- pass the validated numeric principal to `icacls` as `*S-1-...:(F)`;
- use trusted System32 PowerShell with hidden, non-interactive sync and async runners bounded by the existing ACL deadline;
- process-cache only a successful lookup and keep identity lookup failures separate from `icacls` timeouts;
- add WORKGROUP, failure-propagation, cache, async-deduplication, and real Windows token regression coverage.

Fixes #1149.

## Why

On workgroup Windows hosts, `USERDOMAIN` may be the literal `WORKGROUP` while the effective token belongs to `COMPUTER\\user`. The previous `currentWindowsUser()` therefore asked `icacls` to grant an unmappable `WORKGROUP\\user` principal and received exit code 52. Required ACL hardening then failed closed even though NTFS and the effective account were valid.

The token SID is locale-independent and is already the authority used by the Codex coordinator identity work from #998. This PR extracts a low-level ACL-specific resolver rather than importing the higher-level coordinator module.

This is independent of #1130/#1135: those handle one retry after an ACL timeout; this fixes principal construction before `icacls` can succeed.

## Security boundary

- PowerShell is resolved from the trusted Windows system directory through `GetSystemDirectoryW`; PATH and `SystemRoot` do not select it.
- Both runners use `windowsHide: true`, `-WindowStyle Hidden`, `-NonInteractive`, and `-NoProfile`.
- Only strict `S-1-...` output is accepted; raw output, paths, and account identifiers are not included in diagnostics.
- Lookup failure is `EACLIDENTITY`, not `ETIMEDOUT`, so it cannot poison the destination ACL timeout memo.
- Failed and invalid lookups are not cached. Concurrent async callers share one owned lookup.

## Verification

Post-rebase at `6d04574d`:

- Bun 1.3.14: principal + affected ACL suites **166/166 passed**.
- Bun 1.4.0-canary.1 (`b22e0e6d0`): the same suites **166/166 passed**.
- `tests/server-management-auth.test.ts` exceeded its fixed 5-second per-test cap on this host. The unchanged `dev` baseline reproduced the same six timeouts plus one unrelated unhandled assertion, while this branch had no patch-specific correctness failure. This result is reported as baseline timing instability, not as a pass.
- `bun x tsc --noEmit`: passed on both runtimes.
- `git diff --check`: passed.
- Frontend-only lint and doctor gates do not apply because no frontend files changed.
- Independent read-only review found no remaining P0/P1 issue after trusted-executable and identity-failure coverage were added.

## Checklist

- [x] Docs or release notes were updated when needed. (No user-facing configuration or command changed.)

Draft pending upstream CI and the repository-required human maintainer security review.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows access-control handling by identifying the effective account directly, rather than relying on potentially misleading environment values.
  * Added safer failure handling when account identity cannot be resolved, including clearer diagnostics and prevention of unintended permission changes.
  * Improved consistency and reliability for both synchronous and asynchronous permission operations.

* **Tests**
  * Added coverage for identity resolution, caching, timeouts, invalid results, and concurrent lookups.
  * Added integration tests verifying correct permissions and fail-safe behavior.

### DIFF
--- a/src/lib/windows-secret-acl.ts
+++ b/src/lib/windows-secret-acl.ts
@@ -31,6 +31,10 @@
 
 import { existsSync, statSync } from "node:fs";
 import { env, platform } from "node:process";
+import {
+  resolveCurrentWindowsPrincipal,
+  resolveCurrentWindowsPrincipalAsync,
+} from "./windows-user-principal";
 
 const hardenedDirectories = new Map<string, HardenedIdentity>();
 const hardenedPaths = new Map<string, HardenedIdentity>();
@@ -393,18 +397,23 @@ function icaclsError(step: string, result: IcaclsResult): NodeJS.ErrnoException 
   return err;
 }
 
-/**
- * Return the current Windows username from the environment.
- * Falls back to USERDOMAIN\USERNAME if USERNAME alone is ambiguous.
- * The value is used directly in icacls arguments, so it must be present.
- */
-function currentWindowsUser(): string | undefined {
-  const username = env["USERNAME"];
-  const domain = env["USERDOMAIN"];
-  if (!username) return undefined;
-  // USERDOMAIN is the machine/domain name; USERNAME is the account name.
-  // icacls accepts "DOMAIN\User" or just "User" for local accounts.
-  return domain ? `${domain}\\${username}` : username;
+// POSIX CI deliberately drives the Windows ACL branch through platformOverride.
+// The synthetic SID exists only for that test seam; production Windows always
+// resolves the effective token and never falls back to USERDOMAIN/USERNAME.
+const FORCED_NON_WINDOWS_TEST_PRINCIPAL = "*S-1-5-21-1-2-3-1001";
+
+function currentWindowsPrincipal(deadline: number): string {
+  if (platformOverride === "win32" && platform !== "win32") {
+    return FORCED_NON_WINDOWS_TEST_PRINCIPAL;
+  }
+  return resolveCurrentWindowsPrincipal(deadline - nowFn());
+}
+
+async function currentWindowsPrincipalAsync(deadline: number): Promise<string> {
+  if (platformOverride === "win32" && platform !== "win32") {
+    return FORCED_NON_WINDOWS_TEST_PRINCIPAL;
+  }
+  return resolveCurrentWindowsPrincipalAsync(deadline - nowFn());
 }
 
 /**
@@ -424,10 +433,7 @@ function grantAce(user: string, directory: boolean): string {
 }
 
 function runIcacls(targetPath: string, directory: boolean, deadline: number): void {
-  const user = currentWindowsUser();
-  if (!user) {
-    throw new Error("Cannot determine current Windows user for ACL hardening");
-  }
+  const principal = currentWindowsPrincipal(deadline);
 
   // The deadline is owned by hardenEntry (total budget incl. retry + verification).
   const run = (step: string, args: string[]): IcaclsResult => {
@@ -444,7 +450,7 @@ function runIcacls(targetPath: string, directory: boolean, deadline: number): vo
 
   // Step 1: grant current user full control BEFORE any destructive ACL change.
   // If this fails, inheritance is untouched and the writer keeps inherited access.
-  runOrThrow("/grant:r", [targetPath, "/grant:r", grantAce(user, directory)]);
+  runOrThrow("/grant:r", [targetPath, "/grant:r", grantAce(principal, directory)]);
 
   // Step 2: disable inheritance and remove inherited ACEs. The explicit owner ACE
   // from step 1 survives this transition, so a later failure still leaves cleanup access.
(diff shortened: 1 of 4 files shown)


luvs01:
@coderabbitai review

lidge-jun:
Thank you — this is adopted, and the design decisions in it held up under review.

Your commit is the first commit of #1216, cherry-picked unchanged with `Co-authored-by` preserved. I kept every structural choice: extracting a low-level SID resolver rather than reusing the coordinator identity module, reusing `resolveTrustedWindowsPowerShellExe()` instead of writing a third System32 resolver, keeping the identity failure in its own error class so it cannot poison the icacls timeout memo, and caching only successful lookups. Each of those was independently re-derived here before I saw your PR, and yours got there first.

Three things I layered on top, all of which came out of an independent audit rather than a stylistic preference:

**The two most important tests never ran outside Windows.** `FORCED_NON_WINDOWS_TEST_PRINCIPAL` was selected before the injected runner, which made a lookup *failure* unreachable on POSIX — so the fail-closed case and the `timedOutPaths` isolation case carried `if (process.platform !== "win32") return;` and counted as passing while executing nothing. I verified this: they report `(pass)` on macOS with an empty body. Moving the synthetic value into the resolver as its own seam, with selection order `explicit > synthetic > default`, lets an injected failure win and both guards come off.

**`EACLIDENTITY` never reached the caller.** You added the case to `sanitizeDiagnostics`, but `sanitizedAclError` re-attaches only allow-listed codes and that one is not on the list, so a `required: true` harden threw with the cause in the message and `error.code === undefined`. The test matched `/EACLIDENTITY/` against the message, which hid it.

**I tried to add back a qualified name fallback for the optional path and was wrong.** My reasoning was that `DOMAIN\\User` is already qualified so it avoids the ambiguity problem. The audit rejected it: shape is not evidence of the token's subject, both variables are attacker-writable, and `runIcacls` grants Full Control *before* removing inheritance — so a wrong principal either leaves another account holding the secret or strands the file with no usable ACE. Your original decision to decline entirely was the correct one, and it is what shipped.

Closing in favor of #1216. The credit for the fix is yours.

PR #1184, opened 2026-08-07
fix(command-code): guard own-property model lookups

luvs01:
## Summary

- resolve legacy Command Code model aliases only from own table entries;
- reject inherited object properties when reading static and refreshed reasoning-effort profiles;
- cover `__proto__`, `constructor`, and `toString` as literal provider model IDs.

## Why

These tables are ordinary JavaScript objects. Direct bracket lookup can read properties inherited from `Object.prototype`, so a provider model ID with one of those names may be replaced, treated as a profile object, or throw instead of passing through literally.

The patch keeps the existing tables and compatibility aliases; it only makes the lookup boundary explicit with `Object.hasOwn`.

## Verification

- Bun 1.3.14: `tests/command-code-provider.test.ts` **23/23 passed, 58 assertions**.
- Bun 1.4.0-canary.1 (`5f65d3785`): the same suite **23/23 passed, 58 assertions**.
- `git diff --check`: passed.
- Full-suite boundary: a full-suite attempt on the prior patch-equivalent head assertion-crashed in the unrelated `tests/api-storage-policy-run.test.ts` under Bun 1.3.14. An A-B-A single-file comparison produced branch crash / clean-dev pass / branch pass, so this PR does not claim a green full-suite result.

## Checklist

- [x] Docs or release notes were updated when needed. (No user-facing behavior or configuration changed.)

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when selecting Command Code models by consistently recognizing supported model aliases.
  * Prevented prototype-property names from being mistaken for valid models or receiving unintended reasoning settings.
  * Ensured outbound requests use the correct canonical model identifier.

* **Tests**
  * Added coverage for model IDs that match inherited property names.

### DIFF
--- a/src/adapters/command-code.ts
+++ b/src/adapters/command-code.ts
@@ -21,6 +21,10 @@ const COMMAND_CODE_MODEL_ALIASES: Readonly<Record<string, string>> = {
   "glm-5.2": "zai-org/GLM-5.2",
 };
 
+function canonicalCommandCodeModelId(modelId: string): string {
+  return Object.hasOwn(COMMAND_CODE_MODEL_ALIASES, modelId) ? COMMAND_CODE_MODEL_ALIASES[modelId]! : modelId;
+}
+
 /** Flatten tool-result content for the text-only wire output, keeping an `[image]` marker per image part in content order. */
 function toolResultText(content: string | OcxContentPart[]): string {
   if (typeof content === "string") return content;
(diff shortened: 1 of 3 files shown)


luvs01:
@coderabbitai review

lidge-jun:
Approved the pending Cross-platform CI run at this exact head so it can actually execute — `action_required` runs are invisible to `gh pr checks`, so from your side it just looked like nothing was happening.

To be clear about what that does and doesn't do: it only unblocks CI. It does **not** make this mergeable. The four-box readiness checklist is your attestation, not something a maintainer ticks for you, and the gate keeps the PR in draft until you complete it. Once CI comes back green, tick the boxes and it moves to ready on its own.

If the run comes back red, tell me and I'll dig into it rather than leaving you to guess.

lidge-jun:
Landed on `dev` as 5a26fa188, rebased onto the current head with your authorship preserved.

Verified before merge: the inherited-property lookups at `src/adapters/command-code.ts:321,350` and `src/providers/command-code-efforts.ts:34,62` were real on `dev`, and the `Object.hasOwn` guard keeps every legitimate alias while removing the `__proto__`/`constructor`/`toString` misrouting. Full suite green on the rebased head (10175 pass, 0 fail).

Thanks — centralizing the alias resolution rather than patching each call site was the right call.

PR #1185, opened 2026-08-07
test(ci): bind Windows shard assertion to executable command

luvs01:
## Summary

- require the Windows workflow step to execute the complete sharded root-test command;
- compare normalized shell lines instead of accepting independent substrings;
- prove that an `echo` containing the same text does not satisfy the assertion.

## Why

The existing test pinned only the shard fragment. Requiring both fragments with independent `.includes()` checks still accepts a no-op such as `echo bun test --isolate tests --shard=...`.

This assertion now matches the exact executable command line used by the current workflow and derives the divisor from the parsed Windows shard matrix.

## Verification

- Bun 1.3.14: `tests/ci-workflows.test.ts` **122/122 passed**.
- Bun 1.4.0-canary.1 (`b22e0e6d0`): the same suite **122/122 passed**.
- `git diff --check`: passed.

## Checklist

- [x] Docs or release notes were updated when needed. (Test-only contract pin.)

## Review readiness checklist

## Summary by CodeRabbit

* **Tests**
  * Improved validation of Windows CI workflow commands.
  * Added checks to ensure exact executable commands are detected while ignoring comments and blank lines.
  * Prevented false positives from matching text embedded in `echo` statements.

### DIFF
--- a/tests/ci-workflows.test.ts
+++ b/tests/ci-workflows.test.ts
@@ -39,6 +39,15 @@ function count(text: string, fragment: string): number {
   return text.split(fragment).length - 1;
 }
 
+/** Match an executable shell line, not a fragment that could appear in echo or a comment. */
+function hasExactShellCommand(run: string | undefined, expected: string): boolean {
+  return (run ?? "")
+    .split(/\r?\n/)
+    .map(line => line.trim())
+    .filter(line => line.length > 0 && !line.startsWith("#"))
+    .includes(expected);
+}
+
 function expectSecureLinuxKeyringBootstrap(workflow: string): void {
   const smokeStep = workflow
     .split("- name: OS keyring create/read/delete smoke")[1]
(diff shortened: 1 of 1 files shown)


luvs01:
@coderabbitai review

lidge-jun:
@luvs01 Republished on current `dev` as #1301 — your branch was 324 commits behind. Your commit is unchanged and carries `Co-authored-by`.

First, about the red CI here, since it wasn't yours to fix: the failure at `bff31d1e0` is in `tests/autostart-health.test.ts`, which crashed with `EEXIST: file already exists, epoll_ctl` inside a Bun `WriteStream`, followed by a collateral `Cannot call describe() after the test run has completed`. Your diff touches only `tests/ci-workflows.test.ts` and has no path to that file. I haven't diagnosed the Bun-level crash, but it isn't caused by this patch.

Your point is right and I confirmed it costs real coverage. Mutating the Windows Test step to `run: echo bun test …` — so Windows tests print instead of run — keeps `dev` at 125 pass / 0 fail today, and fails with your assertion. That's a green suite over a Windows leg that tests nothing.

One thing I added as a **separate** maintainer commit, so it isn't attributed to you: an assertion that the matching step is unconditional. Your version closes the `echo` hole, but a step with the exact command still runs nothing under `if: false`, and `dev` doesn't catch that either. Same mutation-and-measure treatment: `if: false` also passes on dev today and fails with the new assertion.

I'll close this once #1301 lands. Nothing needed from you — say the word if you'd rather drive it yourself.

lidge-jun:
Landed on `dev` as 3c40df209 via #1301, with your `Co-authored-by` trailer intact through the squash.

Closing this as superseded rather than stale. Your point cost real coverage: mutating the Windows Test step to `echo` the command instead of running it left `dev` green at 125 pass, and fails with your assertion. A maintainer commit alongside it adds the neighbouring case — the exact command under `if: false` also passed on `dev` before this.

Thanks for it, and sorry it took several respins on my side — the CI hangs it kept landing in were ours (#1302), not yours.

PR #1187, opened 2026-08-07
fix(routing): tolerate malformed historical attempts

luvs01:
## Summary

- tolerate malformed legacy `attempts` values in routing analytics;
- validate each attempt and its `recoveryKinds` array before reading it;
- preserve explicit HTTP 429 classification even when historical attempt metadata is malformed;
- add raw JSONL regressions for missing, non-array, null, mixed-type, and malformed-429 historical fields.

## Why

The usage log is append-only history and older or partially malformed rows may not satisfy the current TypeScript shape. The analytics reader parsed each row and then called `.some()` on `attempts` and `recoveryKinds` without runtime validation. A malformed historical row could therefore throw and make the read-only analytics endpoint unavailable.

The writer format is unchanged. This patch makes only the historical reader defensive and treats malformed recovery metadata as non-triggering rather than inventing a cooldown signal. An explicit HTTP 429 remains cooldown-triggering independently of malformed attempt metadata.

## Verification

- Bun 1.3.14: `tests/routing-analytics.test.ts` **11/11 passed, 47 assertions**.
- Bun 1.4.0-canary.1 (`5f65d3785`): the same suite **11/11 passed, 47 assertions**.
- `git diff --check`: passed.
- A full-suite run was not performed on this refreshed head; this PR does not claim a green full-suite result.

## Checklist

- [x] Docs or release notes were updated when needed. (No user-facing contract changed.)

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed historical usage data.
  * Prevented invalid attempt records from affecting request totals or cooldown-triggering failure counts.
  * Preserved existing HTTP 429 handling behavior.

* **Tests**
  * Added coverage for malformed usage records and their impact on routing analytics.

### DIFF
--- a/src/routing/analytics.ts
+++ b/src/routing/analytics.ts
@@ -11,7 +11,7 @@
  * sample for the full history.
  */
 
-import type { PersistedUsageEntry, PersistedUsageAttempt } from "../usage/log";
+import type { PersistedUsageEntry } from "../usage/log";
 import { estimateRequestCost, serviceTierContext } from "../usage/cost";
 import { openRequestHistoryIndex, requestHistoryDb } from "./history/indexer";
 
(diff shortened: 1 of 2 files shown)


luvs01:
@coderabbitai review

luvs01:
@coderabbitai review

lidge-jun:
Approved the pending Cross-platform CI run at this exact head so it can actually execute — `action_required` runs are invisible to `gh pr checks`, so from your side it just looked like nothing was happening.

To be clear about what that does and doesn't do: it only unblocks CI. It does **not** make this mergeable. The four-box readiness checklist is your attestation, not something a maintainer ticks for you, and the gate keeps the PR in draft until you complete it. Once CI comes back green, tick the boxes and it moves to ready on its own.

If the run comes back red, tell me and I'll dig into it rather than leaving you to guess.

lidge-jun:
Landed on `dev` as d519d5d12 and c9f2a58c8 (both commits, your authorship preserved), rebased onto the current head.

Verified before merge: `src/routing/analytics.ts:153` was calling `.some()` on unvalidated historical values, and the fix validates both array boundaries while keeping HTTP 429 classification unconditional. Full suite green on the rebased head (10175 pass, 0 fail).

Thanks for pairing the fix with the raw-JSONL regression — that is what made it quick to confirm.

PR #1188, opened 2026-08-07
fix(openai-chat): accept SSE data fields without the optional space (#1170)

### LINKED ISSUE #1170: OpenAI Chat SSE parser ignores valid `data:{...}` frames without a space
### Client or integration

Other — downstream application embedding the OpenCodex proxy.

Private OpenAI-compatible model gateway (provider identity and endpoint redacted).

### OpenCodex version

@bitkyc08/opencodex 2.7.42

### Endpoint or capability

`/v1/chat/completions` streaming (SSE)

### Current behaviour

The OpenAI Chat adapter only processes lines starting with `data: ` (a colon followed by a space).

The upstream emits valid SSE data fields without that optional space, for example `data:{...}`. As a result, every JSON frame is ignored. A terminal chunk carrying `finish_reason: "stop"` and a trailing usage-only chunk are also ignored. When the upstream closes the stream, OpenCodex reports:

```text
upstream stream ended without a terminal signal ([DONE] or finish_reason) — possible truncation
```

### Expected behaviour

OpenCodex should accept both `data:{...}` and `data: {...}` as SSE data fields.

A non-empty OpenAI `finish_reason` should remain a valid terminal signal even when the upstream omits the OpenAI-specific `data: [DONE]` sentinel.

```sh
curl -N -sS "$BASE_URL/chat/completions" \
  -H "X-Example-Auth: <redacted>" \
  -d '{
    "model": "<model>",
    "messages": [{"role": "user", "content": "hello"}],
    "stream": true
  }'
```

The upstream response uses this framing pattern:

```text
data:{"id":"<id>","object":"chat.completion.chunk","choices":[{"index":0,"delta":{"content":"..."},"finish_reason":null}]}

...

data:{"id":"<id>","object":"chat.completion.chunk","choices":[{"index":0,"delta":{},"finish_reason":"stop"}]}

data:{"id":"<id>","object":"chat.completion.chunk","choices":[],"usage":{"prompt_tokens":0,"completion_tokens":0,"total_tokens":0}}
```

There is no `data: [DONE]` frame.

### Actual response or error

```text
MODEL_RUNTIME_CANARY_FAILED: turn status failed; stream disconnected before completion:
upstream stream ended without a terminal signal ([DONE] or finish_reason) — possible truncation
```

### Upstream documentation

SSE field syntax permits the value to begin immediately after the colon; one optional U+0020 space is stripped when present:

https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation

The non-null `finish_reason` behavior is an OpenAI-compatible streaming convention. The upstream has no public provider-specific specification.

The shared data-line handler can accept the `data:` field name regardless of whether a space follows it:

```ts
if (!line.startsWith("data:")) return "continue";
const payload = line.slice("data:".length).trim();
```

Regression coverage should include:

1. Unspaced `data:{...}` frames with a final non-null `finish_reason` and no `[DONE]` -> successful completion.
2. Standard `data: {...}` framing -> unchanged behavior.
3. A genuinely truncated stream with no `[DONE]`, no final `finish_reason`, and no terminal usage frame -> still fail closed.

The report is fully redacted. No production endpoint, model identifier, request content, or credentials are included.

Yuxin-Qiao:
## Summary

Closes #1170.

Some OpenAI-compatible gateways emit valid SSE data fields with no space after the colon (`data:{...}`, `data:[DONE]`). The OpenAI Chat stream parser only matched `data: ` (colon + space), so every frame — including the terminal `finish_reason` chunk — was silently dropped, and the turn failed with a false truncation error (`upstream stream ended without a terminal signal`).

Per the [WHATWG SSE spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation), the field value may begin immediately after the colon; one optional leading space is stripped when present.

Change: `handleDataLine` in `src/adapters/openai-chat.ts` now matches the `data:` field name regardless of the following space and trims the payload. A bare `data:` line (heartbeat-style keep-alive on some gateways) carries no payload and is skipped rather than reported as a malformed frame. No other parser behavior changes: spaced framing, `[DONE]`, inline error envelopes, usage accounting, and the fail-closed truncation paths are all untouched.

Regression coverage (`tests/openai-chat-eof.test.ts`, new describe block):

1. Unspaced `data:{...}` frames with a final non-null `finish_reason` and no `[DONE]` complete (the exact framing from the issue report).
2. Unspaced `data:[DONE]` terminates the stream — driven by the sentinel frame alone, so the finish_reason EOF fallback cannot mask broken unspaced `[DONE]` handling (CodeRabbit review finding, addressed).
3. A bare `data:` line is ignored, not reported as a malformed frame.
4. A genuinely truncated unspaced stream (no `[DONE]`, no `finish_reason`, no answer text) still fails closed.

## Verification

- `bun test tests/openai-chat-eof.test.ts` — 28/28 pass (4 new).
- Full `bun run test` on the rebased branch — 9569 tests; the only failing names are 11 `provider management validation` tests plus `two processes at the post-approval management seam serialize instead of interleaving`. The identical set fails on a pristine `origin/dev` worktree (eeae0088) without this change, so they are pre-existing environment-dependent failures unrelated to this PR. `tests/crash-guard.test.ts` timed out once under full-suite load and passes standalone on both this branch and baseline.
- CodeRabbit's one Minor finding (the `[DONE]` test could pass via the EOF fallback) was fixed in the follow-up commit; no other review findings are outstanding.
- Branch is rebased on the latest `dev` (0 commits behind at push time).

## Checklist

## Review readiness checklist

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -947,8 +947,14 @@ export function createOpenAIChatAdapter(provider: OcxProviderConfig): ProviderAd
       // Yields adapter events and returns "terminate" for a terminal frame ([DONE] / error) that
       // must end the stream, or "continue" otherwise. Mutates the closure's terminal-signal state.
       const handleDataLine = function* (line: string): Generator<AdapterEvent, "continue" | "terminate"> {
-        if (!line.startsWith("data: ")) return "continue";
-        const payload = line.slice(6).trim();
+        // SSE field syntax: the value may begin immediately after the colon; one optional
+        // leading space is stripped when present. Some OpenAI-compatible upstreams emit
+        // `data:{...}` / `data:[DONE]` without that space, and every frame was dropped.
+        if (!line.startsWith("data:")) return "continue";
+        const payload = line.slice(5).trim();
+        // A bare `data:` line carries nothing (heartbeat-style keep-alive on some gateways);
+        // it is not a malformed frame, just nothing to parse.
+        if (payload.length === 0) return "continue";
         if (payload === "[DONE]") {
           yield* flushToolCalls();
           const stopReason = stopReasonFor(finishReason);
--- a/tests/openai-chat-eof.test.ts
+++ b/tests/openai-chat-eof.test.ts
@@ -293,3 +293,49 @@ describe("openai-chat EOF mid tool call (#735)", () => {
     expect(events.at(-1)?.type).toBe("done");
   });
 });
+
+describe("openai-chat SSE data-field framing (#1170)", () => {
+  // The SSE spec strips at most one optional space after the colon, so `data:{...}` is a valid
+  // data field. Some OpenAI-compatible gateways emit that framing; the parser must accept it
+  // exactly like the spaced form.
+  test("unspaced data:{...} frames with a final finish_reason and no [DONE] complete", async () => {
+    const response = new Response([
+      'data:{"id":"x","object":"chat.completion.chunk","choices":[{"index":0,"delta":{"content":"hi"},"finish_reason":null}]}\n\n',
+      'data:{"id":"x","object":"chat.completion.chunk","choices":[{"index":0,"delta":{},"finish_reason":"stop"}]}\n\n',
+      'data:{"id":"x","object":"chat.completion.chunk","choices":[],"usage":{"prompt_tokens":1,"completion_tokens":2,"total_tokens":3}}\n\n',
+    ].join(""));
+    const events = await collect(createOpenAIChatAdapter(provider).parseStream(response));
+    expect(events.find(e => e.type === "text_delta")).toMatchObject({ type: "text_delta", text: "hi" });
+    expect(events.at(-1)?.type).toBe("done");
+    expect(events.some(e => e.type === "error")).toBe(false);
+  });
+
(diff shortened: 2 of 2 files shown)


Wibias:
Closing this as superseded by #1194.

#1194 was merged into `dev` and fixes the same #1170 SSE issue, including `src/adapters/openai-chat.ts`, but does so through the shared `sseFieldValue()` helper and covers the same bug across six SSE parsers.

I reviewed the current #1188 head for correctness, edge cases, leaks, and security and did not find a separate merge-blocking issue worth preserving this PR for.

One small behaviour here may still be worth carrying forward separately: ignoring an empty `data:` frame instead of treating it as malformed JSON. The stronger sentinel-only `[DONE]` regression test is also useful.

If we want those, I’d move them into a small follow-up against current `dev` rather than merging this now-conflicting implementation.

Thanks for the work on this.

Yuxin-Qiao:
Follow-up opened as #1249 against current dev, carrying the two pieces called out here: empty `data:` frames are skipped instead of reported malformed, and the unspaced `[DONE]` regression test is now sentinel-only. Thanks for the review and the clear handoff.

PR #1189, opened 2026-08-07
fix(history): stream request index ingestion

luvs01:
## Summary

- replace full-tail allocation with a reusable 64 KiB streaming reader;
- assemble complete JSONL records across chunk boundaries with a 1 MiB projection limit;
- retain partial final records until a later refresh completes them;
- advance the indexed offset past complete malformed or oversized records while leaving the canonical usage log untouched;
- preserve batched SQLite inserts and explicit statement finalization.

## Why

The request-history index previously allocated `size - indexedOffset` bytes before parsing the appended tail. A large append could therefore create a proportional one-shot allocation even though the SQLite index is disposable and each record can be processed incrementally.

The new reader bounds transient input memory. A complete record above 1 MiB is intentionally omitted only from the derived SQLite projection; `usage.jsonl` remains canonical and is never truncated or rewritten. `indexedRows` continues to count successfully projected records.

## Verification

- Bun 1.3.14: `tests/request-history-index.test.ts` **20/20 passed**.
- Bun 1.4.0-canary.1 (`b22e0e6d0`): the same suite **20/20 passed**.
- Regressions cover a valid record crossing 64 KiB, an oversized valid record followed by an indexable record, exact offset accounting, canonical-log size preservation, and partial-line completion.
- `git diff --check`: passed.
- Independent read-only review found no blocking P0/P1 issue in offset, chunk, cleanup, or memory-bound behavior.

## Checklist

- [x] Docs or release notes were updated when needed. (The projection contract is documented in source and tests; no public API changed.)

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request-history indexing for large files by processing data in smaller chunks.
  * Correctly handles records that span read boundaries.
  * Skips oversized records without altering the canonical usage log.
  * Ensures incomplete trailing records are safely preserved for subsequent processing.
  * Maintains accurate indexing as additional portions of the history become available.

* **Tests**
  * Added coverage for chunk-spanning records, partial records, and oversized-record handling.

### DIFF
--- a/src/routing/history/indexer.ts
+++ b/src/routing/history/indexer.ts
@@ -76,6 +76,11 @@ export interface RequestHistoryPage {
 export const REQUEST_HISTORY_MAX_PAGE_SIZE = 100;
 export const REQUEST_HISTORY_DEFAULT_PAGE_SIZE = 50;
 export const REQUEST_HISTORY_INSERT_BATCH = 500;
+export const REQUEST_HISTORY_READ_CHUNK_BYTES = 64 * 1024;
+// The SQLite index is a disposable projection. Complete JSONL records above
+// this bound are omitted from the projection; the canonical usage.jsonl is
+// never truncated or rewritten by the indexer.
+export const REQUEST_HISTORY_MAX_RECORD_BYTES = 1024 * 1024;
 
 let db: Database | null = null;
 let dbPath = "";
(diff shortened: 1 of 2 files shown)


luvs01:
@coderabbitai review

lidge-jun:
@luvs01 Republished this on current `dev` as #1287 — your branch was 300 commits behind, and the streaming reader is worth landing rather than letting it age out.

The rebase was clean; the commit is yours via `Co-authored-by`, and I changed none of the logic. On top of your own verification I ran the full suite on the rebased head (9991 pass / 0 fail) and an ablation that raises `REQUEST_HISTORY_MAX_RECORD_BYTES` to `Number.MAX_SAFE_INTEGER`, which fails exactly one of your tests — so the 1 MiB bound is provably load-bearing.

I'll close this one once #1287 lands. Nothing needed from you; if you'd rather drive it yourself, say so and I'll close #1287 instead.

lidge-jun:
Landed on `dev` as 2cb8eddd4 via #1287, with your `Co-authored-by` trailer intact through the squash — the commit is credited to you in the repository history.

Closing this one as superseded rather than stale. Thanks for the fix; the one-shot tail allocation was a real sharp edge.

PR #1192, opened 2026-08-07
fix(responses): bound synthesized SSE expansion

luvs01:
## Summary

- cap bounded-JSON Responses synthesis at 10,000 output items across HTTP and WebSocket reframing;
- generate the shared Responses event sequence lazily;
- stream HTTP SSE frames one pull at a time instead of retaining the expanded event array and joined body;
- fail a streaming HTTP request with a local 502 when the synthesis cap is exceeded;
- preserve DeepSeek item-id repair, status, headers, and the exact `[DONE]` trailer.

## Why

A bounded upstream JSON response can still contain many `output` items. The previous HTTP path materialized both a complete event array and the complete SSE string, amplifying a provider-controlled response in memory. The shared WebSocket event sequence also had no item-count ceiling.

The cap is checked before a stream is returned. Malformed/non-object JSON retains the existing plain-response fallback, while a cap violation is not swallowed by that fallback. If item-id repair fails, the path also falls back instead of synthesizing unrepaired identifiers.

## Verification

- Bun 1.3.14: focused Responses event and DeepSeek bounded-JSON suites **13/13 passed**.
- Bun 1.4.0-canary.1 (`b22e0e6d0`): the same suites **13/13 passed**.
- Handler regression proves a `stream:true` request above the cap returns JSON HTTP 502 with the standard `server_error` shape and no `[DONE]` stream.
- Helper regressions prove synchronous cap enforcement and one-frame-per-pull serialization.
- `git diff --check`: passed.
- Independent read-only review found no blocking P0/P1 issue; the shared WebSocket cap is intentional and documented.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Capped synthesized response output at 10,000 items to prevent oversized event streams.
  - Oversized responses now return a clear server error instead of producing incomplete output.

- **Improvements**
  - SSE responses are generated and delivered incrementally, reducing memory usage.
  - Streaming responses preserve status handling and conclude with the `[DONE]` marker.

### DIFF
--- a/src/server/responses-json-events.ts
+++ b/src/server/responses-json-events.ts
@@ -7,6 +7,8 @@
 
 export type ResponsesJsonEventFrame = Record<string, unknown>;
 
+export const MAX_SYNTHESIZED_OUTPUT_ITEMS = 10_000;
+
 /**
  * The canonical minimal sequence Codex commits: response.created (empty
  * output, in_progress) → one response.output_item.done per output item → a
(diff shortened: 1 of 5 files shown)


luvs01:
@coderabbitai review

lidge-jun:
Reran the cancelled Cross-platform CI run at `b50f23943` and it came back **success**. The earlier `cancelled` was not a failure of yours — Linux shards have been hanging intermittently and getting killed at the job timeout today, which I filed as #1302.

One thing that green does not cover: this branch is **325 commits behind `dev`**. A pass against a base that old says the code worked then, not that it works now. A rebase would make the result mean something, and it is yours to do rather than mine — I have been republishing stale contributor PRs today where the rebase was mechanical, but I would rather you keep authorship of this one if you are still on it.

Still a draft pending your four readiness boxes. Tell me if you would rather I take the rebase, or if you would rather close it.

lidge-jun:
Landed on `dev` as 20969a62f with your authorship preserved, plus one corrective commit on top (3bfe80781).

Your generator, the 10,000-item cap, and the incremental HTTP stream were the right shape — `src/server/responses-json-events.ts:24-38` really did materialize the whole event array before framing. What blocked it was staleness rather than design: dev commit `0b8e608c0` retired DeepSeek production bounded-JSON opt-in, so after rebase your handler regression failed with `expected 502, received 200` — it was no longer exercising the capped path at all.

The correction moves that over-cap 502 regression onto the existing synthetic `modelResponsesUpstreamStreaming:false` fixture, and keeps current dev's "no production registry entry opts in" wording in `structure/04_transports-and-sidecars.md` with your cap guarantee appended rather than overwriting it. Red ablation confirmed the moved test still catches the bug: removing the cap reproduces `Expected: 502, Received: 200`.

Full suite green on the landed head. Thanks.

PR #1203, opened 2026-08-07
feat(models): add context window controls

### LINKED ISSUE #1073: Models GUI cannot define a fallback context window for metadata-poor third-party providers

Configure a correct context window for a routed model from a third-party provider whose `/models` endpoint exposes only model IDs and no context-window metadata.

For example, provider `sub2api` routes model `gpt-5.6-luna`. Its upstream response can be as minimal as:

```json
{
  "data": [
    { "id": "gpt-5.6-luna" }
  ]
}
```

The user knows this route supports a 350K context window and needs the generated routed catalog entry to reflect that capacity.

## What prevents this today?

In v2.10.1, the native-template inheritance fixed by #992 is correctly removed. However, when upstream model metadata is absent, the routed catalog uses the conservative fallback:

```json
{
  "context_window": 128000,
  "max_context_window": 128000,
  "auto_compact_token_limit": 115200
}
```

Setting the provider limit in the Models GUI, which writes:

```json
{
  "providerContextCaps": {
    "sub2api": 350000
  }
}
```

does not change this result. `providerContextCaps` is a ceiling: it can clamp discovered capacity but cannot provide a context window when the provider returns none. Therefore the actual usable context remains 128K.

OpenCodeX already supports the required configuration fields in `providers.<provider>`:

```json
{
  "providers": {
    "sub2api": {
      "contextWindow": 350000,
      "modelContextWindows": {
        "gpt-5.6-luna": 350000
      }
    }
  }
}
```

However, there is no corresponding Models GUI control, so users can set a 350K cap in the UI but still receive the 128K fallback for metadata-poor providers.

## What should OpenCodex do?

Add GUI controls that write the existing configuration fields:

- **Provider default context window** → `providers.<provider>.contextWindow`
- **Per-model context window** → `providers.<provider>.modelContextWindows.<model>`
- Keep **Maximum context cap** → `providerContextCaps.<provider>` as a ceiling only.

The UI should clearly distinguish the meanings:

- **Context window**: explicit capacity to use when upstream metadata is absent.
- **Context cap**: maximum allowed capacity; it must not increase discovered capacity by itself.

Expected result for this example:

```text
sub2api/gpt-5.6-luna
upstream context metadata: absent
modelContextWindows.gpt-5.6-luna: 350000
providerContextCaps.sub2api: 350000

effective context window: 350000
auto_compact_token_limit: 315000
```

This preserves the conservative 128K default for unknown providers while giving users a supported GUI path to configure providers that intentionally omit context metadata.

## Example usage or interface

On the Models page, after selecting a provider or routed model:

```text
Provider: sub2api
Default context window: [350K]
Maximum context cap:   [350K]

Model: gpt-5.6-luna
Context window override: [350K]
```

With no explicit provider or model window, OpenCodeX should continue to use the safe 128K fallback.

## Area

Catalog / Models

estelledc:
## Summary

- Add a Context windows dialog to the Models dashboard for provider defaults and per-model overrides.
- Expose and persist `contextWindow` and `modelContextWindows` through the existing provider management route, including positive-integer validation and per-model merge/delete behavior.
- Document the fallback/cap semantics and add the new labels to all six dashboard locales.

Closes #1073

Models context window controls

## Verification

- `bun run test` - 9640 passed, 8 skipped, 0 failed across 598 files.
- `bun test tests/management-provider-validation.test.ts` - 47 passed, 0 failed.
- `bun run typecheck` - passed.
- `cd gui && bun test` - 647 passed, 0 failed across 116 files.
- `cd gui && bun run lint` - passed.
- `cd gui && bun run build` - passed.
- `cd docs-site && bun run build` - passed, 221 pages built.
- Browser check against an isolated local provider: initial values and model switching were correct; `0` stayed in the dialog with a validation error; a valid save returned `PATCH 200`; reopening showed `256000` and `64000`; no application errors appeared in the console.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added configurable provider-wide and per-model context-window settings.
  * Added validation, save feedback, automatic discovery guidance, and the ability to clear overrides.
  * Added localized settings text in English, German, Japanese, Korean, Russian, and Chinese.

* **Documentation**
  * Clarified that configured context windows act as fallbacks when metadata is unavailable and caps when live limits are smaller.

* **Bug Fixes**
  * Improved persistence and handling of context-window settings through provider management.


lidge-jun:
Thank you — this is adopted, and the layering you chose held up under review.

Your three commits are the first three of #1223, cherry-picked with authorship intact. The structural decision that mattered most was leaving catalog derivation alone: `configuredContextWindow` already reads both fields and materializes them when upstream metadata is absent, so the only thing genuinely missing was the management surface and the control. Routing the edit through the existing provider mutation path kept validation, cache invalidation, and convergence unified rather than growing a second write path.

Four corrections on top, all found by independent audit rather than preference:

**Only the selected model was saved.** `saveContextSettings` keyed the PATCH on `contextModelId` while `contextModelDrafts` held edits for every model, so a value typed into one model and then abandoned by switching the picker vanished — no error, no warning. The test in `models-empty-provider.test.tsx` asserted `{ "claude-opus": 100000 }` after the user had also typed 80000 into `claude-sonnet`, which pinned the loss as correct.

**Sending everything that differs would have been wrong the other way.** The 10s poll can refresh a field mid-modal; comparing drafts against live state then marks an untouched field dirty and reverts a change the user never made. It now requires both conditions — the user touched it, and the value differs from an opening snapshot held as canonical numbers, so retyping 64000 as "64,000" is not an edit. The provider default needed the same treatment; it was previously sent on every Apply.

**`Number.isInteger(1e100)` is true.** Both the validator and the form accepted it. It persists and serializes into the catalog as an enormous number that can make Codex reject the file. Both are safe-integer now, and the default is only validated when touched so an inherited bad value cannot block a per-model save.

**An override for a model that left live discovery was unreachable** — in the drafts map, absent from the picker.

One note on your test for the post-save refresh failure: it was already correct. I initially wrote it up as a defect and was wrong — `load()` catches its own fetch errors and returns `false` rather than throwing, and your third commit had already moved the modal close and the success feedback ahead of the refresh.

Closing in favor of #1223. The credit is yours.

PR #1160, opened 2026-08-06
fix(service): preserve supervision across restart

jonathanli12:
## Summary

- stop registered managers and tracked standalone listeners before service install
- preserve installed service supervision across `ocx restart`
- keep standalone restart behavior unchanged when no service is installed
- add lifecycle regression coverage and update canonical documentation

## Validation

- focused lifecycle suite: 129 passed
- wider service, restart, and update suite: 279 passed
- typecheck: passed
- privacy scan: passed
- git diff check: passed

No live service restart was used for validation.

## Summary by CodeRabbit

* **New Features**
  * Improved `ocx restart` to preserve the current supervision mode.
  * Installed services now restart through the service manager; standalone proxies continue using background startup.
  * Active ports are synchronized with Codex after restarting.

* **Bug Fixes**
  * Service installation now stops existing services and standalone proxies first, preventing port conflicts and restart loops.
  * Failed stop operations correctly prevent restart.

## Review readiness checklist

This PR stays in draft until every box below is ticked. Tick all four boxes once the requirements are met:

- [ ] All CI tests are green on my local testing.
- [ ] I pushed my PR to the latest dev commit.
- [ ] I resolved all correct Codex and CodeRabbit findings.

- [ ] My PR is ready for review.

PR #1131, opened 2026-08-06
fix(lifecycle): bind in-place proxy restart to runtime identity

luvs01:
## Summary

- Make both `ocx restart` and the Windows tray use the running proxy's existing `POST /api/system/restart` lifecycle instead of a separate stop-then-start transaction.
- Attest the exact live runtime PID and port, derive a one-use process-scoped HMAC capability bound to the method, route, PID, and port, and never send the reusable management token.
- Reject invalid or stale target identities before scheduling restart, share one absolute deadline across attestation/request/replacement observation, and report success only after a different runtime PID is healthy on the same port.
- Treat a lost restart response as uncertain: observe the original identity for replacement and never replay it as a destructive fallback stop/start.
- Preserve service supervision, managed Codex/Grok routing, and tray ownership throughout the restart; extend the tray watchdog to cover the documented drain and handoff budget.
- Update lifecycle, Grok, and Windows memory documentation in all maintained locales, plus the management architecture invariant.

This builds on the restart surface introduced by #580 / #594. It is not a duplicate of #720 / #737 (readiness boundary) or #733 / #752 (tray socket inheritance).

## Verification

- Final head, Bun 1.4.0-canary.1: 111 focused restart/auth/tray/lifecycle tests passed, 0 failed.
- Final head, Bun 1.3.14: the same 111 focused tests passed, 0 failed.
- An additional 71-test launcher/provenance/tray regression set passed on both Bun runtimes.
- A broader 208-test affected-surface run passed 207 tests; its one existing timing-sensitive test passed when rerun alone.
- `bun run typecheck` passed on both Bun runtimes.
- `bun run privacy:scan`, PowerShell AST parsing, and `git diff --check` passed.
- `cd docs-site && bun run build` passed, 221 pages generated.
- Independent review found no remaining P0/P1 issue after addressing the current Codex and CodeRabbit findings.
- Live Windows validation on two OpenCodex 2.10.2 hosts confirmed exact PID replacement on port 10100, old PID exit, Canary runtime preservation, service recovery, and intact main/pool account state.
- A repository-wide Windows run is not claimed green: the pre-termination failure was an existing `api-usage` cache test and reproduced unchanged on the latest `dev` worktree. The affected restart surface remains green in the focused runs above.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults.

[...]

PR #1206, opened 2026-08-07
fix(lifecycle): consolidate identity-bound restart and service install cleanup

Wibias:
## Summary

Supersedes #1131 and #1160.

This maintainer-controlled consolidation keeps one restart architecture and carries the non-overlapping service-install hardening from the other PR.

### Restart authority — carried from #1131

- keep the running proxy as owner of restart authorization and drain coordination;
- attest the exact live PID and port;
- use an ephemeral process-scoped restart capability rather than a reusable management token;
- share one bounded deadline across attestation, restart request, and replacement observation;
- report success only after a different identity-verified runtime PID is healthy on the same port;
- preserve service supervision, tray ownership, and managed Codex/Grok routing where those integrations are applicable;
- never replay an uncertain restart as a destructive stop/start fallback.

### Service-install cleanup — carried from #1160

Before writing/loading new service assets:

- stop the recorded installed backend;
- stop a different requested/conflicting Windows backend when applicable;
- stop any tracked standalone proxy last;
- fail closed if manager status, manager stop, or standalone cleanup cannot be verified.

This prevents an old listener from retaining the service port, forcing the newly installed manager into a restart loop, or making the old listener look like a successful install.

### Intentionally not carried

#1160's alternate `ocx restart` stop/start implementation is not included. #1131's identity-bound in-place restart is the canonical restart path in this consolidation.

## Credit

Thank you @luvs01 for the identity-bound restart design and implementation in #1131.

Thank you @jonathanli12 for the service-install cleanup and supervision work in #1160.

## Verification

- The consolidated branch is kept on top of current `dev` before final review.
- It carries the identity-bound restart path from #1131 and the service-install cleanup from #1160.
- Review fixes include clean systemd first-install probing, fail-closed WinSW stop verification, tray child-exit tracking, restart-auth negative cases, and deadline regressions.
- CI on the final consolidated head is the authoritative validation gate.

## Checklist

- [ ] All CI tests are green on the final consolidated head.
- [ ] The branch is updated to latest `dev` before final review.
- [ ] All valid CodeRabbit/Codex findings are resolved.
- [ ] Maintainer security review is complete for the restart/auth boundary.

## Summary by CodeRabbit

* **New Features**
  * Running proxies now restart in place while preserving their port, managed routing, and service supervision.
  * Restarts succeed only after a different, verified process is healthy; uncertain or unverified states fail safely without automatic fallback.
  * Dashboard “Drain & restart” uses the verified replacement flow while preserving Codex injection.
  * Service installation performs safer cleanup and stops when existing processes cannot be confirmed or removed.

* **Documentation**
  * Updated lifecycle, service, troubleshooting, and GUI management guidance across supported languages.


Wibias:
not gui

PR #1209, opened 2026-08-07
feat(v2): Ultra mode toggle — proactive delegation for every model/effort

dbc-hbin:
## Summary

Adds an **Ultra mode** control that enables codex-rs's Proactive multi-agent delegation policy for **every model and every reasoning effort**, not just native `ultra`-effort turns. It exposes the upstream `features.multi_agent_v2.multi_agent_mode_hint_text` config key through OCX's existing v2 config editor, management API, CLI, and the Subagents GUI.

### How it works

Upstream codex-rs derives the multi-agent policy from effort: `ultra` → `MultiAgentMode::Proactive`, everything else → `ExplicitRequestOnly` (core/src/session/multi_agents.rs). A configured `multi_agent_mode_hint_text` **overrides that derivation entirely** and injects the configured text as the `<multi_agent_mode>` developer message. Verified live: with the hint set, `max`/`high` sessions receive the Proactive prompt and `collaboration__spawn_agent` actually spawns sub-agents.

Note: the toggle does **not** change reasoning effort — it changes the delegation prompt policy. The GUI sublabel says so explicitly.

## Changes

- **src/codex/features.ts** — generalized the v2 string-field reader/writer and added `getMultiAgentModeHintText` / `setMultiAgentModeHintText`. Handles all three TOML encodings (dedicated table, inline table, bare-boolean upgrade) with EOL preservation, matching `subagent_developer_instructions`.
- **src/server/management/agent-settings-routes.ts** — `GET /api/v2` returns `multiAgentModeHintText`; `PUT` accepts `string | null`. Empty/whitespace strings are rejected (a present empty override would suppress even the ultra-derived Proactive message upstream).
- **src/cli/v2.ts** — `ocx v2 mode-hint <text|--clear>` plus a status line.
- **gui/** — Ultra mode switch + editable text + "Restore preset" on the Subagents page (i18n: en/ko/ja/ru/zh/de/tr/zh-TW). Switch is disabled until `multi_agent_v2` is enabled.
- **tests/codex-v2-gate.test.ts** — reader/writer encodings, clear semantics, inline-table tricky values (`}`, quotes), bare-boolean upgrade, API GET/PUT/400 validation, CLI round-trip.

## Validation

- Rebased onto current `dev` (`e8db4e036`); `zh-TW` locale keys added to satisfy the new parity guard.
- `bun x tsc --noEmit` (server + gui) passes
- `bun run lint` (gui) passes
- `bun test tests/codex-v2-gate.test.ts tests/management-client-config-route.test.ts` — 141 pass
- GUI suite: 768 pass / 0 fail (includes locale-parity for zh-TW)
- Full suite: core PR test files pass in isolation; full parallel run is subject to pre-existing timing flakes in unrelated files (test-home-guard, etc.) that pass when run individually.

## UI Screenshot

Ultra mode toggle on the Subagents page (switch + description + editor):

Ultra mode toggle

## Summary by CodeRabbit

* **New Features**
  * Added Sub-Agent Ultra Mode with an enable/disable toggle and editable delegation guidance.
  * Added preset restoration, save confirmation, load-retry controls, and prerequisite messaging.
  * Added CLI commands to view, set, and clear the multi-agent mode hint.
  * Added API support for reading, updating, and clearing Ultra Mode guidance.
* **Localization**
  * Added Ultra Mode translations in eight languages.
* **Validation**
  * Added validation and clear error messages for invalid or empty guidance text.
* **Documentation**
  * Documented Ultra Mode configuration and CLI usage.

## UI Screenshot

Ultra mode toggle on the Subagents page (switch + description + editor):

Ultra mode toggle

## Review readiness checklist


Ingwannu (changes requested):
The upstream capability is technically real, but this head is not an integration candidate yet.

- The branch is currently 526 commits behind `dev`; the readiness checklist correctly leaves the latest-dev box unchecked. Please rebase and resolve the accumulated config/GUI/test overlap before further review.
- Exact-head CI is red (`test 2/4`, macOS, gates, and the aggregate `ci` check). Do not treat the local focused pass as sufficient until the rebased head is green.
- The current `decodeTomlStringToken` implementation returns the mandatory first newline inside TOML multiline basic/literal strings. TOML trims the newline immediately following the opening triple quote, so reading a valid `multi_agent_mode_hint_text = """
Proactive..."""` currently adds a spurious leading newline. Fix both triple-quoted forms and keep a regression that matches TOML semantics.

The product naming, default preset text, GUI switch, and whether OCX should expose this unstable upstream override at all require @lidge-jun's explicit product/UI decision. This review is limited to the stale integration and parser/CI blockers; it is not an approval of the UI direction.

Wibias:
@dbc-hbin In your pr description you got boxes which you check. Since there are still open rabbit bug-findings you will have to fix these first and then check all 4 boxes! Thanks!

Wibias (changes requested):
Requesting changes on current head `d0affeb89c8083192b458621ea76d3b84c79b604`.

I reviewed this for runtime/data bugs, leaks, edge cases, security boundaries, and the currently open CodeRabbit/Codex findings. I would not merge this yet.

### Blocking findings

1. **P2: capability cache can become stale when the selected Codex binary is replaced in place**

`probeCodexSupportsModeHint()` can keep a cached result even though the executable at the same path has changed. Both directions matter, but the stale-positive case is the dangerous one: a previously supported binary can be replaced/downgraded with an unsupported one while the cached `true` still permits writing `multi_agent_mode_hint_text`.

Please invalidate capability cache entries using executable identity, e.g. a stat fingerprint (`size`, `mtimeMs`, `ctimeMs`, and inode where available), and add regression coverage for both `false -> binary replaced -> true` and `true -> binary replaced -> false`.

2. **P2: `PUT /api/v2` can partially mutate configuration and then return an error**

A combined request can persist earlier changes such as `multiAgentMode` / enabling V2, then fail the later mode-hint capability probe and return `502`. The caller sees a failed request while part of the requested state has already been committed.

Please preflight the non-null `multiAgentModeHintText` capability check before any mutation, or make the endpoint transactional. Add a regression test proving that a capability failure on a combined request leaves the feature flag, `multiAgentMode`, and TOML unchanged.

3. **Current `dev` integration still needs to be completed before merge**

This branch is materially behind current `dev`. After rebasing, make sure the newly added `sub.ultraMode*` translation keys satisfy the current typed locale contract, including Turkish, and re-run exact-head CI. The current PR description itself still marks “pushed to latest dev” unchecked.

### Other valid findings / edge cases

- Capability-sensitive tests should not depend on whichever Codex binary happens to be installed on the developer/CI machine. Pin a fake runtime for deterministic tests.
- Use `path.delimiter` instead of hard-coding `:` in PATH-related tests so Windows behavior is deterministic.
- The Japanese copy should name the full `features.multi_agent_v2.multi_agent_mode_hint_text` path consistently.
- Docs should distinguish CLI config writes from dashboard/API prerequisite behavior more precisely.
- The multiline-value rejection in `setV2StringField()` appears to scan by key name globally rather than being scoped to `[features.multi_agent_v2]`. An unrelated table containing a multiline `multi_agent_mode_hint_text` can therefore block an otherwise safe Ultra Mode edit. Please scope this guard to the actual target table/assignment.

### Open Rabbit threads

Several unresolved threads are stale and already fixed in substance on this head, including the earlier multiline TOML handling, quoted keys, `--clear`, stale GUI errors, Windows runtime discovery, effective V2 gating, unrelated PATH probing, and proactive preset completeness. Those can be resolved after validating against the rebased head rather than reworking already-correct code.

### Security / leak pass

I did not find a current credential/API-key leak, obvious XSS path, or terminal escape injection in this feature. The main security-adjacent concern is correctness of the capability gate and avoiding configuration state that claims success/failure inconsistently.

Please fix the two P2 issues first, rebase onto current `dev`, resolve the remaining valid deterministic/docs/i18n items, run exact-head CI, then re-check all unresolved Rabbit/Codex threads.

Wibias (changes requested):
Re-review on current head `7f88b13153a310e2695800bfb17362cf059f445b`.

The previous blockers are fixed: capability-cache invalidation now tracks in-place binary replacement, `/api/v2` preflights mode-hint support before combined mutations, clear remains available on unsupported runtimes, capability-sensitive tests are hermetic, PATH fixtures use `path.delimiter`, selected-runtime probing/Windows wrapper resolution are corrected, dotted V2 forms fail closed, multiline handling is much safer, docs/i18n are updated, and this head is rebased onto current `dev`.

One remaining P2 data-integrity blocker:

### Escaped quoted TOML keys can be duplicated and corrupt `config.toml`

Valid TOML can spell the key semantically as:

```toml
[features.multi_agent_v2]
"multi_agent_mode_hint_\u0074ext" = "old"
```

That quoted key decodes to `multi_agent_mode_hint_text`.

The new code is inconsistent about semantic key equality:

- `findTomlAssignment()` decodes quoted keys, so it can recognize the escaped spelling as the target key.
- `getV2StringField()` and especially `editScalarInTable()` still rediscover the key with `tomlKeyPattern()`, which only matches the literal bare/quoted spellings and does not match escaped quoted keys.

A save can therefore append:

```toml
multi_agent_mode_hint_text = "new"
```

next to the escaped quoted assignment. TOML sees both as the same key, so the document now contains a duplicate key and Codex can fail parsing `config.toml`. The reader can likewise report the existing escaped assignment as absent.

Please use semantic assignment discovery consistently for dedicated-table reads and writes. The editor should replace/remove the assignment found by the TOML-aware key decoder rather than rediscovering it with a narrower regex.

Add regression coverage for at least:

```toml
[features.multi_agent_v2]
"multi_agent_mode_hint_\u0074ext" = "old"
```

and verify:
1. reader returns `old`
2. update leaves exactly one semantic key
3. clear removes it
4. resulting TOML still parses

Because `setV2StringField()` is shared, please cover the same representation for `subagent_developer_instructions` too.

CI note: exact-head Cross-platform CI and React Doctor currently show `action_required`, so there is still no executed exact-head CI signal to rely on.

Once this semantic quoted-key case is fixed and exact-head CI actually runs green, this looks close.

Wibias (changes requested):
Re-review against current `dev`: changes are still required.

The current unresolved P1 in `src/codex/features.ts` is blocking: the writer does not recognize root dotted-key forms such as `features.multi_agent_v2.enabled = true` before appending a `[features.multi_agent_v2]` table. That can redefine the same TOML table and leave the Codex config invalid.

The branch is also not mergeable against current `dev` and the PR checklist still says it has not been updated to the latest `dev`. Please fix the dotted-key case, rebase/resolve the current conflicts, and request re-review.

dbc-hbin:
## Re-review request — rebased onto current `dev`

**Head:** `e7ab0cecc` (rebased onto `dev` @ `e8db4e036`, 9 PR commits + 1 i18n fix)

### @Wibias — last review (`1fa84c5df`)

> **P1: dotted-key forms can redefine the table and corrupt `config.toml`**

This was already addressed before that review landed. `setV2StringField` runs a `parsedTomlTable` guard that detects root dotted keys (`features.multi_agent_v2.enabled = true`), quoted segments (`features."multi_agent_v2".enabled = true`), dotted keys inside `[features]`, and inline-table dotted forms. Every one of these fails **closed** without writing a byte:

```
[root dotted]            set -> FAIL  | file unchanged
[root dotted inline]    set -> FAIL  | file unchanged
[quoted segment]        set -> FAIL  | file unchanged
[dotted in [features]]  set -> FAIL  | file unchanged
[dotted with spaces]    set -> FAIL  | file unchanged
[dotted clear]          clear -> FAIL | file unchanged
```

This is pinned by `tests/codex-v2-gate.test.ts` → `"writer refuses dotted V2 definitions without changing config bytes"` and the escaped-quoted-key regression test. Both `setMultiAgentModeHintText` and `setSubagentDeveloperInstructions` share `setV2StringField`, so both are covered.

> **Branch is not mergeable / checklist not updated**

Rebased onto current `dev` (`e8db4e036`) with the one conflict (`src/cli/help.ts` dev rewrite) resolved by re-applying the `mode-hint` synopsis line onto dev's new structure. PR checklist is now 4/4.

> **zh-TW locale contract**

`dev` introduced `zh-TW` as a managed locale with a parity guard. The `sub.ultraMode*` keys were missing from zh-TW on the rebased head; they are now added in `e7ab0cecc` and `locale-parity.test.ts` / `i18n-locales.test.ts` / `claude-desktop-locale.test.ts` all pass.

### @Ingwannu — stale integration / CI / multiline parser

- **Stale base:** rebased onto current `dev`.
- **Multiline TOML newline elision:** `decodeTomlStringToken` trims the TOML-required leading newline for both triple-double and triple-single multiline strings. Pinned by `"reader keeps bracket-shaped prose inside a multi-line hint"` and `"reader decodes multi-line basic and literal TOML strings"`.
- **Capability probe:** `probeCodexSupportsModeHint` caches only `true`; `false`/`null` re-probe on every call. Cache key includes `stat` fingerprint (`dev`, `ino`, `size`, `mtimeMs`, `ctimeMs`) so in-place binary replacement invalidates both directions. Clear (`null`) skips the probe entirely so users can recover from a stale hint on a downgraded runtime.
- **`/api/v2` atomicity:** mode-hint capability is **preflighted** before any mutation in the combined `PUT` path; a capability failure leaves the feature flag, `multiAgentMode`, and TOML unchanged.

### Validation on rebased head

- `tsc --noEmit` (server + gui): **pass**
- `bun run lint` (gui): **pass**
- `bun test tests/codex-v2-gate.test.ts tests/management-client-config-route.test.ts`: **141 pass / 0 fail**
- GUI suite (138 files): **768 pass / 0 fail** (includes locale-parity for zh-TW)
- Full root suite: core PR test files pass in isolation; full parallel run hits pre-existing timing flakes in unrelated files (`test-home-guard`, etc.) that pass individually — same pattern as before this PR.

### Product decisions

@lidge-jun — the product naming ("Ultra mode"), default preset text, and whether OCX should expose this upstream `multi_agent_mode_hint_text` override at all still require your explicit product/UI decision, as @Ingwannu noted. This PR implements the plumbing; the naming/preset can be adjusted trivially if you want a different direction.

Requesting re-review from @Wibias and @Ingwannu.

src/codex/features.ts
```diff
@@ -499,6 +531,19 @@ function scanTomlValueEnd(text: string, start: number): number {
...
   const first = text[i];
   if (first === '"') {
+    // Multi-line basic string: `"""..."""`. The closing delimiter is a triple
+    // quote; a single or double quote inside the body does not end the token.
+    if (text[i + 1] === '"' && text[i + 2] === '"') {
+      i += 3;
+      while (i < text.length) {
+        if (text[i] === "\\") { i += 2; continue; }
+        if (text[i] === '"' && text[i + 1] === '"' && text[i + 2] === '"') {
+          return i + 3;
```
Wibias: **P2: preserve quotes immediately inside a multiline closing delimiter.** TOML explicitly allows one or two quotation marks just inside the delimiters. For example, `multi_agent_mode_hint_text = """hello""""` is valid and evaluates to `hello"`; four quotes around both sides can likewise represent a value that starts/ends with `"`. This branch returns at the first three quotes, so the final quote is left outside the scanned token and `getMultiAgentModeHintText` reports a different value from Codex. The literal-string branch below has the same issue with `''''`. Count the closing quote run and include the valid extra one or two quotes in the token, then add basic + literal regression tests.

Wibias (changes requested):
Re-review on current head `e7ab0cecc862fe896d44511df25ac0e2522d5e45`.

The previous blockers are fixed in substance: the managed V2 string writer now fails closed for unsupported dotted/quoted V2 layouts, escaped quoted field names are handled semantically, capability probing is tied to the selected runtime with replacement-sensitive cache keys, combined mode-hint requests preflight before mutation, and the GUI/CLI fixes from the earlier reviews are present.

One P2 parser correctness issue remains in `scanTomlValueEnd`: TOML permits one or two quotation marks immediately inside a multiline string's closing delimiter. The current scanner stops at the first three quotes of a 4- or 5-quote closing run, so `getMultiAgentModeHintText` returns text different from what Codex parses. The same problem exists for multiline literal strings. Please consume the full valid closing quote run so the extra one or two quotes remain part of the value, and add regression coverage for both basic and literal forms.

Exact-head React Doctor is green. Cross-platform CI is still running at the time of this review, with the gates job already green. The parser issue is the remaining code blocker I found.

dbc-hbin:
## P2 multiline closing delimiter fix — `d8987f59c`

@Wibias Thanks for catching this. The scanner in `scanTomlValueEnd` was stopping at the first triple-quote run, so a valid TOML multiline string like `"""hello""""` (value `hello"`) was read as `hello` — different from what Codex parses.

**Fix:** both the multiline basic (`"""`) and literal (`'''`) branches now consume the full valid closing quote run. TOML permits up to two surplus quotes immediately inside the closing delimiter, so a 4- or 5-quote closing run keeps the extra one or two quotes in the value body.

**Regression tests added** (`tests/codex-v2-gate.test.ts` — "reader consumes surplus quotes in multiline closing delimiters"):
- `"""hello""""` → `hello"` (basic, 1 surplus)
- `"""hello"""""` → `hello""` (basic, 2 surplus)
- `'''hello''''` → `hello'` (literal, 1 surplus)
- `'''hello'''''` → `hello''` (literal, 2 surplus)
- `"""hello"""` → `hello` (normal 3-quote closing, unchanged)

All 132 tests in `codex-v2-gate.test.ts` pass, `tsc --noEmit` passes.

Wibias (changes requested):
Re-review on current head `d8987f59c34686bfad27d3eca3b0d0773e1baf7a`.

The previous P2 parser blocker is fixed correctly. `scanTomlValueEnd` now consumes valid 4- and 5-quote closing runs for both multiline basic and literal TOML strings, and the added regressions verify that the surplus one or two quotes remain part of the decoded hint. I did not find a new code blocker in that fix or in the previously reviewed Ultra-mode implementation.

One integration blocker remains before approval:

- The branch is currently **1 commit behind `dev`**. Current `dev` is `b8a2e51b81aafec33a4865675d2bb29aa1f9eea1`, while this head still has merge base `e8db4e0365b12a314d1c08ec2cf81599efe5b2d3`. The missing dev commit is #1501 (`fix(deepseek): preserve parallel reasoning replay`) and appears unrelated to this PR, so the rebase should be low-risk, but the repository readiness contract requires the PR to be on latest `dev`.
- Exact-head React Doctor is green. Cross-platform CI for this head is still running at the time of review. After rebasing, the new head needs its own green CI signal.
- The PR is currently draft and its readiness checklist is reset/unchecked, which matches that integration state.

Please rebase onto current `dev`, rerun exact-head CI, and request re-review. Assuming that lands cleanly and CI is green, I have no remaining code objection from this review.

Wibias (approved):

Wibias:
Thank you @dbc-hbin for the work on this PR and for addressing the review feedback.

This is useful because Ultra mode is now a proper OCX control instead of a manual config-only feature. Users can enable proactive sub-agent delegation consistently across models and reasoning-effort levels from the UI, CLI, or API. The added TOML safety checks and regression coverage also make it much safer to expose as a user-facing setting.

PR #1220, opened 2026-08-07
feat(server): optional unauthenticated loopback listener for direct-spawn Codex (#1102)

### LINKED ISSUE #1102: [Feature] Make direct-spawn Codex app-server work transparently with wildcard binds
## Client or integration

Codex SDK

## Area

## Summary

When `hostname` is set to `0.0.0.0`, the data plane requires `x-opencodex-api-key`, but only callers that go through the generated `codex` shim receive `OPENCODEX_API_AUTH_TOKEN`. A `codex app-server` launched by a third-party host that resolves the Codex entrypoint directly (`require.resolve('@openai/codex/bin/codex.js')`) never gets the token, so every model call is rejected with `401` before an SSE stream opens.

Request: `codex app-server` should receive the key as transparently as a local `codex` invocation does, so that a non-loopback bind does not silently break app-server-driven clients. Today the shim already covers `app-server` when it is the process being launched; the gap is that the token is only available when the shim is on the execution path at all, and it is captured once at process start.

## Reproduction

1. Set `hostname` to `0.0.0.0` in `~/.opencodex/config.json` and run `ocx start` (proxy on `127.0.0.1:10100`).
2. Configure Codex with the injected provider block:
   ```toml
   model = "kiro/claude-opus-5"
   model_provider = "opencodex"

   [model_providers.opencodex]
   base_url = "http://127.0.0.1:10100/v1"
   wire_api = "responses"
   env_http_headers = { "x-opencodex-api-key" = "OPENCODEX_API_AUTH_TOKEN" }
   ```
3. Start an app-server from a host process that does **not** inherit `OPENCODEX_API_AUTH_TOKEN` and does not go through the generated shim. Any harness that resolves the entrypoint directly reproduces this:
   ```js
   spawn(process.execPath, [require.resolve('@openai/codex/bin/codex.js'), 'app-server', ...])
   ```
4. Send a turn from that client.
5. Observed: the turn fails after 5 retries. The client surfaces a generic stream failure because the `401` happens at admission, before any SSE frame is emitted.
6. Expected: the app-server authenticates against the local proxy transparently, exactly as `codex` launched through the shim does.

A second, related trigger without any third-party host: restart `ocx` so the token file is rewritten. Every already-running `codex app-server` keeps the previous value and starts returning `401` until it is restarted.

## Version

`@bitkyc08/opencodex` v2.10.1

## Operating system

Ubuntu 22.04.5 LTS (aarch64), Node.js v22.12.0, Codex CLI v0.146.1

## Provider and model

kiro / claude-opus-5 (also reproduced with gpt-5.6-sol; the failure is provider-independent)

## Logs or error output

From Codex's structured log store, target `codex_core::responses_retry`. Four separate turns each exhausted all 5 retries:

```
13:14:55~13:15:01  unexpected status 401 Unauthorized: opencodex API key required,
                   url: http://127.0.0.1:10100/v1/responses   model=gpt-5.6-sol  retries=1..5
13:17:44~13:17:50  same 401   model=kiro/claude-opus-5
13:20:42~13:20:48  same 401   model=kiro/claude-opus-5
13:25:24~13:25:30  same 401   model=kiro/claude-opus-5
```

Reproduced directly against the proxy with no key. No SSE frames are produced:

```
POST http://127.0.0.1:10100/v1/responses  {"model":"kiro/claude-opus-5","stream":true}
HTTP 401
{"error":{"message":"opencodex API key required","type":"authentication_error","code":"invalid_api_key"}}
```

The same request with `x-opencodex-api-key` set streams normally and ends with `response.completed`.

N/A

## Redacted configuration

```json
{
  "port": 10100,
  "hostname": "0.0.0.0",
  "defaultProvider": "openai",
  "providers": {
    "kiro": { "adapter": "kiro", "authMode": "oauth", "defaultModel": "kiro-auto" }
  }
}
```

## Current implementation facts

- `isApiAuthRequired()` keys admission entirely off the bind hostname, so `0.0.0.0` makes the key mandatory:
  ```ts
  // src/server/auth-cors.ts
  export function isApiAuthRequired(config: OcxConfig): boolean {
    return !isLoopbackHostname(config.hostname);
  }
  ```
- `resolveApiAuth()` returns `{ kind: "loopback" }` and skips the token entirely on a loopback bind; otherwise it requires one of `x-opencodex-api-key`, `Authorization: Bearer`, or `x-api-key`, and `requireApiAuth()` answers `401 "opencodex API key required"`.
- The expected value is read from the environment only: `configuredApiAuthToken()` returns `process.env.OPENCODEX_API_AUTH_TOKEN?.trim()`.
- `shouldInjectApiAuthHeader()` in `src/codex/inject.ts` uses the same loopback predicate, which is why the `env_http_headers` line is injected for a `0.0.0.0` bind.
- `buildUnixCodexShim()` in `src/codex/shim.ts` exports the token before parsing the subcommand, and `app-server` is already listed in `CODEX_INTERNAL_COMMANDS`. So `codex app-server` **through the shim** is authenticated; the token is read from `serviceApiTokenFilePath()` once, at process start.
- Consequence, observed on this host: the app-server started via the shim has `OPENCODEX_API_AUTH_TOKEN` set, while a host process that resolves `@openai/codex/bin/codex.js` directly has no such variable to pass down to the app-server it spawns.
- `writeServiceApiTokenFile()` in `src/service.ts` rewrites the token file from the current environment on service start, so the file contents can change while older app-server processes keep a stale value in memory.

## Problem statement

A non-loopback bind is a supported configuration, and the provider block is injected for it automatically. But authentication only works when the caller's environment happens to carry the token. Any Codex-app-server-based integration that launches the Codex entrypoint directly gets a `401` at admission, which reaches the user as an opaque stream failure rather than an auth error, because the rejection precedes the first SSE frame. Token rotation on `ocx` restart produces the same symptom for already-running app-servers.

## Goal

Make the API key reach `codex app-server` transparently whenever opencodex has injected the `x-opencodex-api-key` header requirement, so that switching `hostname` to `0.0.0.0` does not break app-server clients or require callers to plumb `OPENCODEX_API_AUTH_TOKEN` themselves.

## Scope

- Ensure the token is available to `codex app-server` regardless of whether the process was launched through the generated shim.
- Handle token rotation so an app-server started before an `ocx` restart does not keep failing with a stale value.
- Make the resulting failure diagnosable when the key genuinely cannot be supplied.

## Non-goals

- Changing the `isApiAuthRequired()` loopback predicate or weakening auth for a `0.0.0.0` bind.
- Redesigning the admission header precedence in `resolveApiAuth()`.
- Fixing how third-party clients render a failed turn; that is their concern.
- The `Host`-header admission problem in #570, which is a different gate.

## Suggested direction

Inferred, not verified as the maintainers' preferred design:

- Have the app-server auth path resolve the token at request time rather than only from the start-time environment, falling back to `serviceApiTokenFilePath()` when `OPENCODEX_API_AUTH_TOKEN` is absent or no longer matches. This addresses the direct-spawn case and rotation together, and mirrors what the shim already does for the CLI.
- Alternatively, or in addition, accept a loopback-only admission for requests that arrive on `127.0.0.1` even when the bind is `0.0.0.0`, keyed on the local socket address rather than the configured hostname. That keeps the network surface gated while making local app-server traffic work without a key. This is a security-relevant change and deserves an explicit decision.
- Likely files: `src/server/auth-cors.ts` (`configuredApiAuthToken`, `resolveApiAuth`), `src/codex/shim.ts`, `src/lib/service-secrets.ts`.
- Consider making `requireApiAuth()`'s message distinguish "no key presented" from "key did not match", so a stale token is not misread as a missing one.

## Acceptance criteria

- With `hostname` set to `0.0.0.0`, a turn driven by a `codex app-server` that was launched without `OPENCODEX_API_AUTH_TOKEN` in its environment completes successfully instead of returning `401`.
- After `ocx` restarts and rewrites the token file, an app-server process that was already running either recovers or reports a specific, actionable error.
- Regression: a loopback bind (`127.0.0.1`) still admits requests with no key, and `resolveApiAuth()` still reports `{ kind: "loopback" }`.
- Regression: with a non-loopback bind, a request presenting a wrong key is still rejected with `401`, and the network-facing surface remains gated.
- A test covers the direct-spawn path, that is, an app-server whose environment lacks `OPENCODEX_API_AUTH_TOKEN`.

## Suggested smoke test

1. Set `hostname` to `0.0.0.0`, run `ocx start`.
2. Launch `codex app-server` from a shell where `OPENCODEX_API_AUTH_TOKEN` is explicitly unset and the generated shim is not on `PATH`.
3. Drive one turn through that app-server and confirm it streams to completion.
4. Restart `ocx`, then drive another turn on the same app-server process and confirm the documented behavior (recovery or a specific error).
5. Confirm `curl` against `/v1/responses` with no key still returns `401`.

## Checks

---

> **Note:** This issue was filed with AI assistance. The `401` bursts, the keyless `401` response, and the authenticated `response.completed` were reproduced on the reporter's host; the suggested direction is inferred from reading the source and is not a proven design.

lidge-jun:
## Summary

A `codex app-server` launched by a host that resolves the entrypoint directly never passes through the generated shim, so it never inherits `OPENCODEX_API_AUTH_TOKEN`. On a wildcard bind every model call then 401s at admission, before any stream frame.

The tempting fix is to exempt callers whose socket peer looks like loopback. That is unsound: `requestIP()` proves only the last transport hop, and Docker Desktop port forwarding, host-network containers, WSL mirrored networking and tunnel terminators all re-open remote connections locally. It would also have ridden `resolveApiAuth` into eight endpoints unrelated to this issue.

Instead a second listener binds `127.0.0.1`. The kernel refuses remote connections outright, so there is no address to judge, and the public listener's admission policy is unchanged.

Off by default. `unauthenticatedLoopbackListener: { enabled: true, port: N }`.

## What is load-bearing

**A per-request policy view.** Auth and CORS read a `Pick<OcxConfig, "hostname" | "corsAllowOrigins" | "apiKeys">` chosen from the receiving listener. Rewriting the whole config and holding it would go stale on the next management change; adding an `allowUnauthenticated` parameter to the resolvers would put an admission bypass on the wrong side of the boundary. The narrow type also means a policy view leaking into a routing path fails to typecheck.

**The bind is not the whole boundary.** An attacker page can make a victim's browser connect to `127.0.0.1`, and that connection is local. The listener takes the same Host/Origin branch a plain loopback bind always has. A test asserts the same hostile Origin the loopback policy rejects would be *accepted* under the public policy.

**A fixed port.** Never OS-assigned. An ephemeral port would change across restarts while running app-servers kept the old `base_url` — the exact symptom this issue reported for token rotation, which investigation showed does not actually happen (`writeServiceApiTokenFile` mirrors an already-present env value, only from install/repair).

**`GET /v1/models` is on the four-route allowlist.** When catalog materialization fails, `syncCodex` injects with `catalogPath: null` and Codex falls back to an online model manager that refreshes through it. Returning 404 would fix the direct-spawn host while breaking its model list.

**One startup transaction.** If the loopback bind fails, the public listener rolls back rather than being stranded — otherwise the CLI's port retry reads it as a public-port conflict and picks a different port, accumulating listeners. Composite stop completes cleanup on both listeners *and* propagates failure; swallowing it would let `drainAndShutdown` report success while a socket is held.

## Verification

- `bun run typecheck` clean, `bun run privacy:scan` passed.
- `bun run test` — 9756 pass, 8 skip, 0 fail across 607 files.
- 20 integration tests speak HTTP to real listeners: same request 401 on public and 200 on loopback, connection refused on the machine's non-loopback IPv4, every non-allowlisted route probed with the method its handler accepts, the rebinding Host/Origin shape rejected, both ports rebindable after stop, a squatted loopback port rolling back a fixed public port.
- Ablation, all driven red rather than argued: loopback bound to `0.0.0.0` fails 2; allowlist check disabled fails 1; policy view collapsed to the shared config fails 3; `/v1/messages` added to the allowlist fails 1; the inject substitution removed fails 1; per-step catch removed fails 2; collected failures swallowed fails 3; `requestServer.upgrade` swapped for `server.upgrade` fails 1.

## Known gaps, stated rather than implied

- **No direct-spawn activation evidence yet.** This PR is a draft for that reason. `@openai/codex` is not a test dependency, so the plan calls for a non-skippable activation run against a real app-server before ready-for-review: isolated `CODEX_HOME` with no `models_cache.json`, entrypoint resolved directly, token stripped from the child environment, a runtime-generated unique routed model asserted in `model/list`, and one turn against that model. Catalog-present and catalog-absent both.
- **No admission-attribution assertion on a completed turn.** Admission is proven; the `{ kind: "loopback" }` log row on a real turn is not.
- Two seams have no runtime oracle on this Bun version and are pinned by source assertions instead, with the reason recorded in the tests.

## Status

This does not close #1102. The default is off, so the reporter's repro still 401s until they adopt the option, and the question of whether it suits their deployment is still open on the issue.

Stacked on #1216.

Refs #1102

### DIFF
--- a/devlog/_plan/260807_untouched_bug_stack/100_loopback_peer_admission.md
+++ b/devlog/_plan/260807_untouched_bug_stack/100_loopback_peer_admission.md
@@ -1,0 +1,378 @@
+# 100 — #1102: `0.0.0.0` 바인드에서 로컬 Codex 가 401 로 막힌다
+
+> **개정 이력.** 첫 판은 "opt-in 으로 loopback 소켓 피어를 무인증 admit" 을
+> 제안했다. 독립 감사가 P1 다섯 건으로 되돌렸고, 그중 둘이 설계를 바꿨다:
+> (a) 그 스위치는 `resolveApiAuth` 를 타고 #1102 와 무관한 8개 엔드포인트까지
+> 열고, (b) 공용 리스너의 피어 주소는 최종 사용자 신원이 아니다. 아래는
+> 재설계된 판이다.
+>
+> **2차 개정.** 재설계본도 감사에서 P1 세 건을 받았다. 설계 방향은 유지됐지만
+> 구현 계약이 비어 있었다: ephemeral 포트가 재시작마다 바뀌면 우리가 부정했던
+> "재시작 후 app-server 가 깨진다" 를 우리 손으로 만들고, 로컬 리스너의
+> auth/origin/WS 처리 경계가 미정이며, 두 bind 가 하나의 트랜잭션이 아니었다.
+> 아래 §고정 포트 / §리스너 정책 / §바인드 트랜잭션 이 그 답이다.
+>
+> **3차 개정.** 세 번째 감사가 P1 둘을 더 찾았고 둘 다 검증된 사실이다:
+> 카탈로그가 없을 때 app-server 가 `GET /v1/models` 로 폴백하는데 우리
+> allowlist 가 그걸 404 로 막고, `allSettled` 만으로는 stop 실패가 삼켜져
+> 재시작이 아직 포트를 쥔 리스너 위에 바인드를 시도한다.
+
+## 이슈가 말한 것과 실제
+
+리포터는 두 개의 트리거를 보고했다. 하나는 정확했고, 하나는 원인이 다르다.
+
+**맞음 — direct-spawn 갭.** `app-server` 는 shim 의 `CODEX_INTERNAL_COMMANDS`
+(`src/codex/shim.ts:42`) 에 있고 shim 은 디스패치 전에 토큰을 export 한다
+(`:384-389`). 그러니 shim 을 거친 `codex app-server` 는 인증된다. 문제는 서드파티
+호스트가 `require.resolve('@openai/codex/bin/codex.js')` 로 엔트리포인트를 직접
+resolve 해서 spawn 할 때다. 그 경로는 shim 을 통째로 우회하고, 대안이 없다:
+`/v1/responses` admission 은 `x-opencodex-api-key` 만 받고
+(`src/server/auth-cors.ts:369-376`), 토큰 파일은 admission 시점에 읽히지 않는다.
+
+**틀림 — "재시작하면 토큰이 회전된다".** `writeServiceApiTokenFile()` 은 이미
+`process.env.OPENCODEX_API_AUTH_TOKEN` 에 있는 값을 쓰고, 없으면 아무것도 쓰지
+않는다 (`src/service.ts:347`). 호출자는 service install/repair 뿐이고
+(`:1707`, `:1799`, `:1937`, `:2116`), `ocx service start` 는 파일을 다시 쓰지
+않는다 (`:2660`). 토큰은 애초에 사용자가 공급하는 값이고 OpenCodex 가 생성하지
+않는다. 그러니 평범한 재시작이 살아 있는 app-server 를 무효화하지 않는다.
+
+이 정정은 이미 이슈에 코멘트로 게시되어 있고, 리포터에게 두 가지를 물었다.
+답은 아직 없다.
+
+## 왜 파일 기반 대안이 전부 막히는가
+
+토큰을 shim 밖 프로세스에 "실제로 전달" 하려면 그 프로세스의 환경을 바꿔야
+하는데, OS 프로세스 환경은 spawn 시 복사되고 우리는 남의 프로세스 환경을 사후에
+못 바꾼다. 남는 후보를 전부 확인했다:
+
+| 후보 | 왜 안 되는가 |
+|---|---|
+| Codex `env_http_headers` 를 파일 기반으로 | 값이 **환경변수 이름**이다. 업스트림 설계이고 우리 쪽 변경 범위 밖 |
+| static `http_headers` | 시크릿을 `~/.codex/config.toml` 에 평문으로 직렬화한다. 백업·저널·동기화 경로로 퍼진다 |
+| `auth.command` | bearer credential 을 공급하는데, `/v1/responses` 는 전용 헤더만 받는다. Codex Direct 와 충돌 방지를 위한 의도적 거부 (`auth-cors.ts:369-372`) |
+| OS 전역 환경 주입 | 무관한 GUI/터미널 자식까지 credential 을 상속한다. 이미 떠 있는 호스트에는 적용도 안 된다 |
+
+전부 막힌다. 그래서 이건 credential **전달** 문제가 아니라 admission **정책**
+문제다.
+
+## 첫 설계가 왜 틀렸나
+
+처음에는 `isApiAuthRequired()` 를 우회하는 opt-in 스위치
+(`trustLoopbackPeersOnRemoteBind`) 를 제안했다. 감사가 두 가지를 지적했고 둘 다
+코드로 확인된다.
+
+**하나 — 폭발 반경.** `resolveApiAuth` 는 8곳에서 호출된다
+(`src/server/index.ts:692, 882, 903, 937, 1009, 1024, 1087, 1121`): `/v1/models`,
+Images generations/edits, artifacts, alpha search, Messages, Live/Realtime,
+sideband WebSocket. `resolveResponsesApiAuth` 도 `/v1/responses` 만이 아니라
+compact 와 Chat Completions 경로에서 쓰인다. resolver 안에 피어 예외를 넣으면
+#1102 가 요청하지 않은 표면 전부가 같이 열린다. 수용 기준이
+`/v1/responses` 만 검사했으므로 이 확대를 탐지하지도 못했을 것이다.
+
+**둘 — 피어 주소가 증명하는 것.** `requestIP()` 는 **마지막 transport hop** 만
+알려준다. Docker Desktop 의 포트 포워딩, `--network host` 컨테이너, WSL2 의
+mirrored networking 과 `netsh portproxy`, Kubernetes sidecar, VPN/터널 종단 —
+전부 원격 연결을 로컬 TCP 연결로 다시 연다. 그 배포에서는 원격 호출자가
+loopback 피어로 보인다. 흔한 구성이고, 첫 판은 리버스 프록시와 SSH 터널만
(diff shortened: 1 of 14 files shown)


lidge-jun:
## Direct-spawn activation evidence

The audit named this the ready gate. Here it is, from a real `codex app-server` on Codex 0.146.0 rather than an in-process stand-in.

```
PASS  resolved the real Codex entrypoint, not PATH — /Users/jun/.nvm/.../node_modules/@openai/codex/bin/codex.js
PASS  entrypoint runs — codex-cli 0.146.0
PASS  built the catalog with our own serializer — 1 entries
PASS  wrote an isolated CODEX_HOME with no models_cache.json
PASS  stripped OPENCODEX_API_AUTH_TOKEN from the child environment
PASS  app-server initialized — ok
PASS  model/list contains the unique routed model that only our listener can supply — ocx-direct-spawn-da6d9c3f-...
PASS  thread/start accepted — 019fdd24-8ae1-7a61-9f4d-d962a9d6e19e
PASS  the app-server reached the loopback listener without a credential — POST /v1/responses

9/9 checks passed
```

The last line is the claim of #1102: a real app-server, spawned from the resolved entrypoint with no token in its environment, opened `/v1/responses` against the injected `base_url` and was admitted.

**On the oracle.** Codex caches model lists and falls back to a bundled catalog when a refresh fails, so "did `model/list` succeed" proves nothing — a broken path is indistinguishable from a working one. The model id is generated per run, so a name no bundled catalog can contain could only have arrived through our listener.

**Two things writing it surfaced.** `model/list` reads `model_catalog_json`; it does not call the provider's `/v1/models`. The first version of the harness therefore watched Codex return its five bundled ids while never touching the listener — the exact false-negative shape the unique id exists to expose, aimed at the harness instead of the feature. And a hand-written catalog fixture is a liability: Codex rejects the whole file on any schema mismatch and silently falls back, so the catalog is now built with `buildCatalogEntries`, which also means the script exercises the bytes `ocx sync` writes.

**Ablation.** Pointing `base_url` at a dead port makes the final check red, so the harness watches the hop rather than asserting its own setup.

The script is committed at `scripts/verify-loopback-direct-spawn.mjs`. It is deliberately not a `bun test` file: the repository does not depend on `@openai/codex`, and a test that skips when it is absent would report green on machines that never ran it.

Marking ready for review.

PR #1226, opened 2026-08-07
fix(catalog): restore DeepSeek V4 context window on routed rebuilds

iF2007:
## Summary

DeepSeek routed models (deepseek-v4-flash / deepseek-v4-pro) were rebuilt with a 128k context window on every catalog sync, overwriting the user's configured value. Root cause: the registry entry had no `jawcodeBundle`, so `resolveJawcodeProvider("deepseek")` returned `undefined` and the strict-fields fallback wrote 128000. A provider saved under a title-cased key (`"DeepSeek"`) missed the alias table a second time because lookup was case-sensitive.

Fix:

- `src/providers/registry.ts`: add `jawcodeBundle: "deepseek"` to the DeepSeek entry and set `modelContextWindows` to the official 1,048,576. The DeepSeek V4 (flash/pro) parameters — including the 1,048,576 context window, 384k max output, text-only input, and `high` default reasoning effort — come from DeepSeek's official Codex integration docs: https://api-docs.deepseek.com/quick_start/agent_integrations/codex/
- `scripts/generate-jawcode-metadata.ts` + `src/generated/jawcode-model-metadata.ts`: fold provider-name case in `resolveJawcodeProvider` so `"DeepSeek"` resolves like `"deepseek"`; regenerate the snapshot with the 1,048,576 window.
- Tests: aliases now include `deepseek`, `DeepSeek` resolves, registry context windows updated, and a new catalog test asserts routed deepseek entries restore 1,048,576 instead of the 128k fallback.

## Verification

- `bun run typecheck` passes.
- `bun test tests/provider-registry-parity.test.ts tests/jawcode-metadata-sync.test.ts` — 35 pass.
- `bun test tests/codex-catalog.test.ts` — 120 pass.
- `bun test` for the catalog/adjacent suites (admission/restore/golden/prewarm/vision-sidecar/oauth/cursor) — all green.
- Regeneration guard: generated snapshot matches a fresh run of the generator against the fixed jawcode source.

## Checklist

- [x] Docs or release notes were updated when needed. (No user-facing docs change: the value now matches the official spec; no config surface changed.)

## Review readiness checklist

## Summary by CodeRabbit

* **Improvements**
  * Updated DeepSeek V4 Flash and Pro to support 1,048,576-token context windows.
  * Added compatibility aliases for deprecated DeepSeek model IDs.
  * Improved provider recognition with case-insensitive matching.
  * Enhanced model discovery when DeepSeek models are missing from the model list.
  * Updated metadata to reflect text-only capabilities and automatic compaction limits.
  * Improved compatibility with DeepSeek model metadata and catalog resolution.


Ingwannu (changes requested):
The underlying fix is still valuable: current `dev` still has the DeepSeek V4 context values at 1,000,000 and the direct DeepSeek registry entry does not select the `deepseek` metadata bundle. However, this branch now conflicts with `dev`, and the metadata pipeline changed after the PR was built. `scripts/generate-jawcode-metadata.ts` and `src/generated/jawcode-model-metadata.ts` were replaced by the vendored `scripts/model-metadata.source.json` -> `scripts/generate-model-metadata.ts` -> `src/generated/model-metadata.ts` pipeline, with a byte-for-byte sync guard.\n\nPlease rebase onto current `dev`, update the vendored source/registry rather than the retired generated paths, regenerate with `bun run generate:model-metadata`, and keep the focused routed-catalog regression plus `tests/model-metadata-sync.test.ts` green. I am requesting changes because the current conflicting branch cannot be merged safely, not because the reported context-cap issue is invalid.

PR #1256, opened 2026-08-08
fix(usage): bound startup hydration tail reads

luvs01:
## Summary

- cap `readRecentUsageEntries()` startup hydration at a 64 MiB tail window;
- preserve the existing adaptive growth within that bound;
- stop cleanly when the bounded tail contains no complete line instead of expanding to the whole file;
- add a sparse oversized-ledger regression proving rows outside the bounded tail are not allocated or parsed.

## Why

`readRecentUsageEntries()` starts with a small tail window and multiplies it until enough valid rows are found. Before this change, the upper bound was the full `usage.jsonl` size. A large or malformed persisted row could therefore make every startup allocate and parse the complete ledger even though the caller only requests a small recent set.

The management/history paths already use separately bounded readers. Startup hydration should have the same finite-memory property.

## Behavior

The reader still returns the newest valid entries that fit inside the retained window and still expands for trace-sized rows. It now stops at 64 MiB. If that bounded tail contains no newline, the function returns an empty recent set rather than attempting a whole-file read. The append-only ledger itself is not modified.

## Verification

- Bun 1.3.14: `tests/usage-log.test.ts` **30/30 passed**.
- Bun 1.4.0-canary.1 (`b22e0e6d0`): the same suite **30/30 passed**.
- `git diff --check`: passed.

## Checklist

- [x] Tests cover the bounded behavior and existing usage-log semantics.
- [x] No credential, routing, or user-facing configuration contract changes.

Ready for review; upstream CI and maintainer review remain pending.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup usage-log loading by limiting reads to a 64 MiB window.
  * Ensures the newest usage entries are returned without scanning excessively large logs.

* **Tests**
  * Added regression coverage for bounded reads when older data exists beyond the limit.

## Review readiness checklist

### DIFF
--- a/src/usage/log.ts
+++ b/src/usage/log.ts
@@ -420,6 +420,7 @@ export type UsageLogRevision = {
 
 let usageReadCacheStats = { fullReads: 0, tailReads: 0, parsedLines: 0 };
 const MANAGEMENT_USAGE_MAX_READ_BYTES = 64 * 1024 * 1024;
+const RECENT_USAGE_MAX_READ_BYTES = 64 * 1024 * 1024;
 const MANAGEMENT_USAGE_READ_CHUNK_BYTES = 1024 * 1024;
 const MANAGEMENT_USAGE_MAX_ENTRIES = 500_000;
 const MANAGEMENT_USAGE_FLIGHT_STALE_MS = 30_000;
@@ -655,10 +656,10 @@ export function readRecentUsageEntries(limit: number): PersistedUsageEntry[] {
     fd = openSync(path, "r");
     const size = fstatSync(fd).size;
     if (size <= 0) return [];
-    // Trace-sized rows (up to MAX_TRACE_BYTES, RI-01) need a larger per-row
-    // budget than the pre-trace ledger; keep expanding until the window covers
-    // the file start or the whole file so a restart never hydrates nothing.
-    let windowBytes = Math.min(size, Math.max(64 * 1024, Math.ceil(limit) * 20 * 1024));
+    // Trace-sized rows need a larger per-row budget than the pre-trace ledger,
+    // but startup hydration must never grow into an unbounded whole-file read.
+    const maxWindowBytes = Math.min(size, RECENT_USAGE_MAX_READ_BYTES);
+    let windowBytes = Math.min(maxWindowBytes, Math.max(64 * 1024, Math.ceil(limit) * 20 * 1024));
     while (true) {
       const start = Math.max(0, size - windowBytes);
       const buf = Buffer.alloc(size - start);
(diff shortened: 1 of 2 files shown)


Ingwannu (approved):
Reviewed exact head `86f81604`. Startup hydration is now capped to a 64 MiB tail instead of expanding to an unbounded whole-file read, while preserving newest-valid-row parsing within that window. Focused coverage passed locally (30 tests), `git diff --check` passed, and typecheck passed. Core Linux/macOS/keyring/storage shards are green; the remaining npm-global/gates jobs are still queued, so do not merge until the exact head is fully green.

Ingwannu (approved):
Re-reviewed rebased exact head `f2ce1ea5`. The PR diff is unchanged in scope from the approved bounded-tail patch. Exact-head focused tests passed 30/30 and typecheck passed. Do not merge until the newly triggered exact-head CI is green.

luvs01:
Rebased exact head `7c40ac63` onto current `dev` (`517f4460`) without changing the two-file patch.

Focused verification on the rebased state:
- Bun 1.3.14: `tests/usage-log.test.ts` 30/30 passed.
- Bun 1.4.0-canary.1: 30/30 passed.
- Typecheck, privacy scan, and `git diff --check` passed.

The branch is now exactly one commit ahead and zero behind `dev`; fresh exact-head upstream CI is required because the previous green run covered the prior base.

lidge-jun:
Approved the pending Cross-platform CI at your head; the first attempt ended `cancelled` and a rerun came back **green**.

The cancellation was not yours. Two shards (`test 3/4` and `test 4/4`) each ran exactly 15m15s and were killed at the job timeout — the intermittent Linux shard hang I filed as #1302. Yours is the first run I have seen where two shards hung at once, so I added it there as evidence.

The pending-approval part was ours to clear too: contributor runs sit at `action_required" until a maintainer releases them, and `gh pr checks` does not surface that state, so it looks like nothing is happening rather than like something is waiting on us.

Still a draft pending your four readiness boxes — those are your attestation and I will not tick them for you. CI is green at your exact head, so box 1 is provable now.

lidge-jun:
Landed on `dev` as c8fddb847, rebased onto the current head with your authorship preserved.

Verified before merge: `src/usage/log.ts:659-681` really did grow `windowBytes` toward the whole file and allocate against it, so startup hydration was unbounded. The 64 MiB cap preserves newest-valid-row behavior inside the window, and the sparse-file regression proves data beyond the bound is never parsed. Full suite green on the rebased head (10175 pass, 0 fail).

Thanks.

PR #1259, opened 2026-08-08
fix(ci): require fail-closed aggregate-check evidence

luvs01:
## Summary

- make Cross-platform CI start for every pull request while keeping expensive jobs behind the existing read-only `changes` filter;
- validate `changes.outputs.ci` as exactly `true` or `false`, re-emit only that validated value, and fail the producer job on missing or malformed output;
- keep an always-created aggregate `ci` result, including documentation-only pull requests whose expensive jobs are intentionally skipped;
- require positive, latest, GitHub-Actions-produced `ci` evidence and fail closed when the filtered response is missing, unreadable, foreign, pending, unsuccessful, conflicting, or truncated;
- grant the trusted readiness job the explicit `checks: read` permission required by its live check lookup;
- preserve the generic paginator harness while adding exact filtered-response cardinality coverage.

## Why

The readiness gate previously treated missing aggregate evidence as green. Requiring a check without changing the trigger would deadlock documentation-only pull requests, because the workflow-level path filter created no `ci` result for them.

This change always creates the pull-request workflow, scopes costly work inside the read-only `changes` job, validates the filter output before downstream jobs consume it, and requires complete positive evidence before accepting the contributor CI claim.

## Verification

- Bun 1.3.14: `tests/ci-workflows.test.ts` — **130/130 passed**.
- Bun 1.4.0-canary.1 (`b22e0e6d0`): the same file — **130/130 passed**.
- `git diff --check`: passed.
- Independent full-file reviews: no diff-introduced P0-P3 findings.
- Codex Security diff scan: complete coverage of all four changed files, **0 reportable findings**, no deferred work.
- Windows full suite: attempted, not passed, and not counted as verification. The run hit unrelated pre-existing Windows/Bun isolate and storage-fixture failures before Bun 1.3.14 terminated with an internal assertion.

## Security boundary

The readiness writer remains a `pull_request_target` workflow executing trusted default-branch code. Untrusted values are transported through step environment variables instead of shell-source interpolation. The Cross-platform CI workflow retains read-only repository permissions for pull requests, uses hosted runners for PR jobs, exposes no secrets, and keeps the persistent self-hosted Windows lane dispatch-only.

This GitHub Actions policy change remains Draft pending exact-head upstream CI and the repository-required maintainer security review.

## Readiness checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,10 @@
 name: Cross-platform CI
 
 on:
-  pull_request:
+  # Always create the aggregate `ci` check for pull requests. Expensive jobs
+  # apply the former path allowlist through the `changes` job below, so a
+  # docs-only PR receives explicit positive evidence instead of no check at all.
+  pull_request: {}
     # No base-branch filter on purpose. GitHub matches `branches:` against the
     # BASE ref, so `[main, dev]` silently excluded stacked child PRs — whose
     # base is another open PR's head branch, an intentional review workflow per
@@ -12,31 +15,14 @@ on:
     #
     # An allowlist cannot express "base is another PR's head" — stacked bases
     # carry contributor prefixes (`fix/`, `feat/`, `agent/`) as readily as
-    # `codex/`, and contributor stacks need CI most. `paths:` below is the real
-    # scope gate, same shape as issue-quality-tests.yml. Safe to widen here
+    # `codex/`, and contributor stacks need CI most. The `changes` job below is
+    # the real scope gate, using the same allowlist as the push trigger. Safe to
+    # widen here
     # because this workflow is `pull_request` (not `pull_request_target`),
     # declares `contents: read`, and reads no secrets.
     #
     # `push:` stays pinned to the integration lines: it gates the release path,
     # and this trigger already covers review.
-    paths:
-      - "src/**"
-      - "bin/**"
-      - "tests/**"
-      - "scripts/**"
-      - "gui/**"
-      - "assets/**"
-      - ".gitattributes"
-      - ".npmignore"
-      - "package.json"
-      - "bun.lock"
-      - "tsconfig.json"
-      - "README.md"
-      - "LICENSE"
-      - ".github/workflows/ci.yml"
-      - ".github/workflows/release.yml"
-      - ".github/workflows/enforce-pr-target.yml"
-      - ".github/workflows/stale-needs-info.yml"
   push:
     branches: [main, preview, dev]
     paths:
@@ -77,8 +63,8 @@ jobs:
   # A hostile PR can delete the branch and hardcode the self-hosted labels into
   # `$GITHUB_OUTPUT`, and `runs-on` will honour it. That this job runs on
   # `ubuntu-latest` changes nothing — the untrusted part is its OUTPUT, not its
-  # host. `.github/workflows/ci.yml` is in this workflow's `pull_request.paths`,
-  # so such an edit triggers its own run.
+  # host. `.github/workflows/ci.yml` is in the `changes` job's `ci` filter, so
+  # such an edit triggers every expensive verification job.
   #
   # What actually keeps untrusted code off a self-hosted runner lives OUTSIDE
   # this file, where a PR cannot reach it: the fork-PR approval policy
@@ -154,6 +140,7 @@ jobs:
       contents: read
       pull-requests: read
     outputs:
+      ci: ${{ steps.filter.outputs.ci }}
       gui: ${{ steps.filter.outputs.gui }}
       packaging: ${{ steps.filter.outputs.packaging }}
     steps:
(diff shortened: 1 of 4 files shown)


luvs01:
Updated on exact head `9d2839446509ecf46355979523383bdec3765e74`.

- Added strict `true|false` validation and re-emission for the CI scope output, so a missing or malformed filter result fails the `changes` job instead of silently skipping all expensive jobs.
- Replaced the impossible decoy-pagination fixture with complete filtered check evidence, added fail-closed `total_count` mismatch coverage, and preserved the generic paginator harness.
- Added explicit `checks: read` permission to the readiness job.
- Merged current `dev` `14e94852517875592d4a228d09ff010ec73b55f2`; the PR is 0 commits behind.
- Bun 1.3.14 and Bun 1.4.0-canary.1 focused runs both pass 130/130; typecheck, privacy scan, and diff check pass.
- Final Codex Security diff scan completed with all four files covered and 0 reportable findings.

The PR remains Draft. Exact-head upstream Actions and maintainer security review are still required; the Windows full suite is not claimed green.

lidge-jun:
Reviewed this one closely because it changes when CI runs at all, which `MAINTAINERS.md` puts in the security-review class. The problem it solves is real — a docs-only PR currently produces *no* `ci` check rather than a passing one, which is harmless today and a trap the moment the check becomes required. Moving scope gating into `changes` is the right shape.

One blocker before it can land, and it is in the fail-closed property the PR is named after.

**`changes.outputs.ci` is forwarded without validation.** It is exposed as `${{ steps.filter.outputs.ci }}` and every expensive job gates on `needs.changes.outputs.ci == 'true'`. If `changes` *succeeds* but that output is missing or malformed — a `dorny/paths-filter` upgrade renaming an output, a filter-syntax slip, an edge case where the step yields nothing — then:

1. every expensive job evaluates `'' == 'true'` → skipped;
2. the aggregate `ci` gate treats `skipped` as a pass, deliberately and correctly, since that is how it recognises trigger-scoped jobs;
3. `ci` reports green with nothing having been tested.

The `changes` job failing is handled fine — the aggregate catches it. It is `changes` *succeeding with an unusable output* that slips through. Right now `paths` in the trigger makes this impossible to reach; after this PR the output becomes the single point of truth, so it needs to be load-bearing on purpose.

Suggested fix inside `changes`, after the filter step:

```yaml
      - name: Assert the scope output is usable
        shell: bash
        run: |
          set -euo pipefail
          case "${{ steps.filter.outputs.ci }}" in
            true|false) ;;
            *) echo "::error::changes.outputs.ci was '${{ steps.filter.outputs.ci }}', expected true or false"; exit 1 ;;
          esac
```

That turns the silent-skip case into a failing `changes`, which the aggregate already handles correctly.

Two smaller notes:

- The harness scenario with 100 `decoy-*` checks does not model the real request. Production queries with `check_name: "ci"` and `filter: "latest"`, so the API does not return that population — the pagination path may be unreachable under the narrowed query. Either exercise the filtered endpoint semantics or drop the pagination claim.
- This overlaps `enforce-pr-target.yml` with #1265, which merged into `main` but is **not** on `dev` (`dev` and that commit are diverged, 6 ahead / 155 behind). Whichever lands second inherits a conflict, so the ordering needs deciding before either is republished.

Holding rather than approving. CI here is `cancelled` rather than failing, so there is no failure of yours to chase — the base branch had an unstable shard today. Happy to take the validation step myself if you would rather not respin.

lidge-jun:
Two corrections to my comment above, one of which is a defect in the snippet I suggested.

**The validation step I wrote interpolates the expression directly into shell.** `case "${{ steps.filter.outputs.ci }}" in …` substitutes the value into the script *before* bash parses it, which is the same shape as the injection pattern this repository's workflow hardening exists to avoid. That the value comes from a SHA-pinned action makes it low risk here, not correct. Pass it through the environment instead, so bash sees data rather than source:

```yaml
      - name: Assert the scope output is usable
        shell: bash
        env:
          CI_SCOPE: ${{ steps.filter.outputs.ci }}
        run: |
          set -euo pipefail
          case "$CI_SCOPE" in
            true|false) ;;
            *) echo "::error::changes.outputs.ci was '$CI_SCOPE', expected true or false"; exit 1 ;;
          esac
```

Worth being explicit about what this does and does not buy: it closes the *malformed output* hole — empty string, renamed output, unexpected value — and nothing more. It cannot tell you a syntactically valid `false` is the semantically right classification. That still rests on the pinned `dorny/paths-filter` and on the filter-list tests, which is the part worth keeping direct coverage on.

**I also overstated the #1265 overlap.** I said whichever of #1259 and #1265 lands second inherits a conflict. I checked the branch relationship, not the hunks — both touch `enforce-pr-target.yml`, but I did not show their changed regions actually collide. Accurate statement: they *may* conflict, and the integration order needs deciding, with any conflict resolved at that point.

The main blocker stands unchanged: `changes` succeeding with an unusable `outputs.ci` currently ends in a green `ci` with nothing tested.

lidge-jun:
Landed on `dev` as d6d3878c4 and 1e07b0e28, with your authorship preserved.

The fail-open you found was real: `.github/workflows/ci.yml` filters out docs-only PRs so no aggregate `ci` check is created, and the readiness gate then accepted `ciCheck === undefined` as green.

The delay was a collision, not a doubt about the fix. #1329 merged first and touched the same harness hunk — it added `filePages` to `paginatePageCount` where your patch adds `checkRunPages`, so your diff stopped applying (`patch failed: tests/helpers/enforce-pr-target-harness.ts:616`). The two changes are complementary — hygiene gates first, then CI evidence must be trusted — so the rebase kept BOTH fixture families rather than resolving one away: `filePages` still drives `pulls.listFiles` for the hygiene tests, and `checkRunPages` still normalizes check-run envelopes for your provenance fixtures. Your merge-tip correction survived the rebase too, which was the specific thing at risk.

Fail-closed behavior was re-exercised across missing, malformed `total_count`, foreign-app, pending, conflicting (both orderings), failed, and truncated evidence. Permissions only add job-scoped `checks: read`; nothing about `pull_request_target` was widened. Full suite green on the landed head.

Thanks — this one was worth the care.

PR #1261, opened 2026-08-08
fix(ci): bind maintainer sponsorship to the reviewed head

luvs01:
## Summary

- make `maintainer-sponsored` a head-specific approval, alongside the four existing exception labels;
- centralize the exact five-label set and invalidate it only when a `synchronize` event changes the PR head;
- keep sponsorship intact for label, reopen, and other events that do not change the reviewed revision;
- document that a maintainer must review the new head and reapply sponsorship after any pushed commit.

## Why

The hygiene workflow currently treats `maintainer-sponsored` as PR-wide. Once applied, later commits retain the label even if they add or alter authentication, credential, GitHub Actions, release, or dependency-installation code.

That label is also the workflow's automated evidence that the restricted surface received the maintainer security review required before merge. Keeping it across a head change therefore separates the approval from the revision it was meant to attest. This does not create an automatic merge path—normal review and CI still apply—but it can make an old review appear current in a repository where these controls are enforced by convention.

The other four approval labels are already cleared on `synchronize`. This change gives sponsorship the same revision-bound lifecycle while preserving it on events that do not modify the head.

## Verification

- `node --test .github/scripts/pr-sponsored-surface.test.cjs`: **9/9 passed**.
- `node --test .github/scripts/pr-hygiene.test.cjs`: **21/21 passed**.
- Bun 1.3.14: `tests/ci-workflows.test.ts` **122/122 passed**.
- Bun 1.4.0-canary.1 (`b22e0e6d0`): the same file **122/122 passed**.
- `git diff --check`: passed.

## Security boundary

The workflow remains `pull_request_target` and continues to execute trusted default-branch code only. It does not check out PR-head content, expand token permissions, or add a write endpoint. The sole additional mutation is removing an existing approval label after a head-changing event; a maintainer can reapply it after reviewing that revision.

As with the existing hygiene policy, the live behavior changes only after this workflow reaches the repository default branch.

## Checklist

- [x] The exact five head-specific labels are regression-tested.
- [x] `synchronize` invalidates sponsorship.
- [x] Non-head-changing events preserve sponsorship.
- [x] Contributor documentation explains re-review and reapplication.

Draft pending upstream CI and the repository-required maintainer security review for GitHub Actions changes.

### DIFF
--- a/.github/scripts/pr-sponsored-surface.cjs
+++ b/.github/scripts/pr-sponsored-surface.cjs
@@ -26,6 +26,14 @@ const RESTRICTED_PREFIXES = [
   "src/oauth/",
 ];
 
+const HEAD_SPECIFIC_APPROVAL_LABELS = [
+  "test-exception-approved",
+  "suppression-approved",
+  "generated-change-approved",
+  "dependency-change-approved",
+  "maintainer-sponsored",
+];
+
 const RESTRICTED_FILES = new Set([
   // Release and packaging automation executed by the release workflow.
   "scripts/release.ts",
@@ -63,6 +71,10 @@ function hasSponsorship(labels) {
   );
 }
 
+function headSpecificApprovalLabelsForAction(action) {
+  return action === "synchronize" ? HEAD_SPECIFIC_APPROVAL_LABELS : [];
+}
+
 /**
  * @returns {{ code: string, paths: string[] }[]} empty when the pull request may proceed
  */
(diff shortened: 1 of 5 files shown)


luvs01:
Withdrawing this change: it changes an intentionally chosen governance policy rather than correcting a deterministic product defect. That decision should remain with the maintainers, so I am not asking for review.

PR #1262, opened 2026-08-08
chore(typescript): bump to TypeScript 7.0.2 (native port)

Wibias:
## Summary

- Bump the root devDependency `typescript` 5.9.3 → 7.0.2 — the Go-based native port (8-12x faster full builds).
- Fix the three type-level strictness differences TS 7 surfaces in the proxy runtime:
  - `native-exec-common.ts`: make `execBytes` generic over the message case so the value type is checked consistently (removes the `as never` escape hatch).
  - `server/index.ts`: new `sendUpstreamFrame` helper copies `Buffer` frames into an `ArrayBuffer`-backed `Uint8Array` before `WebSocket.send` (Bun's `BufferSource` rejects `SharedArrayBuffer`-backed buffers); the live-sideband frame sends now use it.
  - `server/index.ts`: `readyState` CLOSED checks use the numeric literal 3 (with explanatory comments) where Bun's type narrows `readyState` to 0|1|2.
- No runtime behavior changes; the fixes are type-system-only accommodations.
- GUI stays on TS ~6.0.2 (typescript-eslint's `<6.1.0` peer bound); TS 7 support there waits on typescript-eslint 9.x.

## Validation

- `bun run typecheck` — pass (TS 7.0.2)
- `bun audit` — pass, 0 vulnerabilities
- `bun test` focused (cursor native exec + websocket live transport) — pass, 115 tests / 0 fail
- Full suite runs on CI (Linux shards); the Windows-local Bun crash is a documented pre-existing environment issue

## Review notes

- `execBytes` is called from native-exec.ts and native-exec-tools.ts with `create(SomeSchema, {...})` values; the generic signature type-checks each call site exactly.
- The `readyState` numeric-literal comparisons are runtime-identical to `WebSocket.CLOSED` (which is 3).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of binary live updates to ensure frames are transmitted reliably.
  * Preserved text-based live updates while improving connection close-state handling.
  * Strengthened command message validation and serialization for more reliable execution.
  * Improved handling of success, metadata, and diagnostic responses.

* **Compatibility**
  * Improved support for environments with stricter WebSocket type definitions.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -13,7 +13,7 @@
       },
       "devDependencies": {
         "@types/bun": "1.3.14",
-        "typescript": "5.9.3",
+        "typescript": "7.0.2",
       },
     },
   },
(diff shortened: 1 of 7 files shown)


Ingwannu (changes requested):
The focused implementation looks coherent, and I independently passed the 23 touched tests, `bun run typecheck`, and `bun run privacy:scan` on `ccf85d3c` with CPU affinity limited to two cores.

I am holding approval for two merge gates rather than a code defect:

- this dependency/native-toolchain change is now 13 commits behind current `dev` (`d55b903d`), beyond the repository readiness window, so please rebase it onto the current integration head;
- the macOS required check is still pending, and this TypeScript package ships platform-native binaries, so that platform result is part of the dependency-boundary review.

After the rebase, please rerun the focused serialization/install/translator tests plus typecheck/privacy and let the full cross-platform CI complete. I will re-review the exact new head once those gates are green.

PR #1263, opened 2026-08-08
fix(codex): reject profile FIFOs without blocking

luvs01:
## Summary

- open bounded native-profile files with `O_NONBLOCK` on POSIX while retaining `O_NOFOLLOW`;
- reject FIFOs and other non-regular entries after the descriptor is opened, without waiting for a writer;
- run the FIFO regression in a separate Bun process with a two-second OS-level timeout so a future regression cannot hang the test runner.

## Why

The bounded reader validates the opened descriptor with `fstat`, but a POSIX `openSync(..., O_RDONLY)` can block before that validation when an untrusted profile path is a FIFO with no writer. The size cap and regular-file check therefore never get a chance to reject it.

Adding `O_NONBLOCK` makes the open return promptly. The existing descriptor check then rejects the FIFO as a non-regular file, while regular file, replacement-race, and bounded-read semantics remain unchanged.

The regression is intentionally isolated in a child process. If the flag is removed later, the OS timeout terminates only the child and reports a deterministic test failure instead of freezing the entire suite.

## Verification

- Bun 1.3.14 on Windows: `tests/native-profile-store.test.ts` **15 passed, 1 POSIX-only test skipped**.
- Bun 1.4.0-canary.1 (`b22e0e6d0`) on Windows: the same result.
- `git diff --check`: passed.

The FIFO assertion is POSIX-only and could not run locally because this host has no WSL distribution; the draft intentionally leaves that OS-specific execution to Linux CI.

## Security boundary

This narrows a local availability edge in the native-profile metadata reader. It does not broaden accepted file types, paths, permissions, or credential-store behavior. Symlinks remain rejected by `O_NOFOLLOW`, and the descriptor must still be a regular file within the existing byte cap.

## Checklist

- [x] Product change is limited to the POSIX open flags.
- [x] A hanging regression cannot block the parent test runner.
- [x] Existing bounded-read and keyring tests remain green.

Ready for review; Linux CI and maintainer review remain pending.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented vault reads from hanging when encountering special files without an available writer.
  * Invalid vault entries now fail promptly with a clear unsupported authentication-store error.

* **Tests**
  * Added Unix-specific coverage to verify that special-file reads complete promptly rather than blocking.

## Review readiness checklist


Ingwannu (changes requested):
The `O_NONBLOCK` change is a sensible defense for direct `readBounded` callers, but the new regression test does not reach that code path and currently fails on the PR head. `readNativeProfileVault()` first calls `assertNativeProfileMetadataLayout()`, which rejects the FIFO as `PROFILE_STORAGE_UNSAFE`; the child only accepts `VAULT_INVALID`, so it exits 92 (locally: 15 pass, 1 fail; the macOS CI job also failed). Please make the test exercise a direct bounded-read surface such as the config/auth path, or explicitly test the existing metadata-layout rejection separately, and include `child.stderr` in the assertion message so future failures are diagnosable. I will re-review once the test proves the intended nonblocking boundary and CI is green.

Ingwannu (changes requested):
Re-checked rebased exact head `7c3fa541`. The blocker is unchanged: the new test still calls `readNativeProfileVault()`, so `assertNativeProfileMetadataLayout()` rejects the FIFO as `PROFILE_STORAGE_UNSAFE` before the `O_NONBLOCK` open path is exercised. The child accepts only `VAULT_INVALID` and exits 92; exact-head result remains 15 pass / 1 fail. Please apply the requested direct bounded-read regression (or separate layout-contract assertion) rather than only rebasing the same patch.

luvs01:
Updated exact head `ba8bcfce` and merged `dev` commit `57ea8df4` into the branch. The live `dev` ref advanced once immediately afterward; the PR is currently ahead 3 / behind 1 and only the two intended files differ.

The POSIX regression now exercises the actual vault TOCTOU window requested above:

- the child resolves an isolated native-profile context and writes a regular vault, so metadata layout validation succeeds;
- the existing bounded-read `beforeOpen` seam then unlinks that vault and replaces it with a FIFO;
- `readNativeProfileVault()` must return `VAULT_INVALID` promptly;
- the parent uses a two-second `spawnSync` timeout and asserts no process error, a null signal, status 0, and includes stderr in failure diagnostics.

Local Windows verification on the new test commit:

- Bun 1.3.14: 15 passed, 1 POSIX-only test skipped;
- Bun 1.4.0-canary.1: 15 passed, 1 POSIX-only test skipped;
- `bun run privacy:scan`: passed;
- `git diff --check`: passed.

The race assertion is POSIX-only, so exact-head Linux/macOS CI remains required. The new workflow runs are currently awaiting maintainer approval; the PR stays Draft and the readiness checklist remains open until CI and re-review are complete.

lidge-jun:
The fix is right and the test is wrong — CI is failing for that reason, not because of the code.

I first concluded the defect did not exist: with a FIFO placed at the vault path before the call, `assertCanonicalFile` rejects it in ~0ms with `PROFILE_STORAGE_UNSAFE`, well before `readBounded` reaches `openSync`. That observation was correct and my conclusion from it was not.

The real defect is a TOCTOU race. A regular file that passes `assertNativeProfileMetadataLayout` can be replaced with a FIFO before `openSync`, and by then the guard is behind us. Using the `beforeOpen` seam the module already ships for exactly this window:

```
clean origin/dev (no patch):  hangs until the harness kills it
this branch:                  throws VAULT_INVALID in ~3ms
```

So `O_NONBLOCK` fixes something real.

The test creates the FIFO before the call, so it trips the upstream guard and never reaches the code you changed — it gets `PROFILE_STORAGE_UNSAFE` while asserting `VAULT_INVALID`. Correcting the assertion to the pre-guard code would be worse: that test would then pass without the patch too.

Suggested shape:

- child process (mandatory — unpatched code hangs and would take the runner with it)
- parent passes isolated `codexHome`/`configDir` via env; child calls `resolveNativeProfileContext` with them
- child writes a **regular** vault file so the layout check passes
- child installs the seam under `Symbol.for("opencodex.native-profile-store.bounded-read-test-seam")` and, in `beforeOpen`, unlinks and `mkfifo`s the vault path
- parent uses `spawnSync` with a bounded timeout and asserts no error, null signal, status 0
- `test.skipIf(process.platform === "win32")` since `mkfifo` is POSIX-only

Asserting signal/timeout is what separates "hung" from "rejected quickly". Happy to take it once that lands.

lidge-jun:
Landed on `dev` as 7f5f9c92b and 9fc1d486d (both commits, your authorship preserved).

Independent security review passed. Confirmed `O_NOFOLLOW`, the `fstat` regular-file check, the pre-read size check, and the bounded `limit + 1` read all still run — `O_NONBLOCK` only removes the block, it does not relax any check. Also confirmed the earlier `CHANGES_REQUESTED` objection is obsolete: the old test installed the FIFO before `assertNativeProfileMetadataLayout` so it never reached `readBounded`, while the current test validates a regular vault first and then swaps through the `beforeOpen` seam. Full suite green on the rebased head (10189 pass, 0 fail).

Thanks — reaching the real TOCTOU window in a bounded child process is what made this reviewable.

PR #1266, opened 2026-08-08
fix(google): replay Vertex thought signatures

### LINKED ISSUE #1254: google-vertex: Codex tool continuation returns 400 when functionCall thought_signature is not replayed
### Client or integration

Codex CLI

`google-vertex` (Vertex AI, service-account JSON / ADC)

### OpenCodex version

`2.10.2`

### Endpoint or capability

Codex CLI tool-call continuation through `google-vertex/gemini-3.6-flash` (`/v1/responses` translation / tool-call replay)

### Current behaviour

Vertex connectivity and authentication are healthy, and tool-free requests succeed, but a Codex turn that invokes a tool fails on the continuation with HTTP 400.

A direct OpenCodex Responses request succeeds:

```json
{
  "model": "google-vertex/gemini-3.6-flash",
  "input": "Reply exactly: VERTEX_OK",
  "stream": false
}
```

Response:

```text
status: completed
model: gemini-3.6-flash
text: VERTEX_OK
```

A simple question sent from Codex also succeeds. The failure appears only after a function call, on the subsequent continuation turn.

### Expected behaviour

The `thought_signature` / `thoughtSignature` returned by Gemini/Vertex for a function call should be preserved and replayed unchanged on the subsequent tool-result continuation, so Codex tool loops can continue successfully.

```shell
1. Configure a `google-vertex` provider using `adapter: google`, `googleMode: vertex`, and a service-account JSON exposed through ADC / `GOOGLE_APPLICATION_CREDENTIALS`.
2. Select `google-vertex/gemini-3.6-flash` in Codex CLI.
3. Send a tool-free prompt. It succeeds.
4. Send a prompt that requires a shell command or another Codex tool.
5. The function call is emitted, but the continuation fails with HTTP 400 because the prior `functionCall` part is missing its thought signature.

Redacted provider shape:
adapter:      google
googleMode:   vertex
auth:         service-account JSON via ADC
baseUrl:      https://aiplatform.googleapis.com
model:        gemini-3.6-flash
```

### Actual response or error

```shell
Provider error 400: Vertex AI invalid request: Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:shell_command`, position 4.

OpenCodex records:
status: 400
error: invalid_request_error
route: explicit-provider
google-vertex/gemini-3.6-flash
```

### Upstream documentation

https://ai.google.dev/gemini-api/docs/thought-signatures

Gemini requires thought signatures associated with function calls to be passed back unchanged in subsequent turns when using tools.

This looks like a replay / translation issue rather than a credential or connectivity problem:

- OpenCodex proxy health is OK.
- The same `google-vertex/gemini-3.6-flash` provider succeeds through `/v1/responses` for a tool-free request.
- Codex tool-free requests succeed through the same provider.
- Failure appears only after a function call, on the continuation turn.
- Vertex explicitly reports that the prior `functionCall` part is missing its required `thought_signature`.

This looks related in concept to #897, where Antigravity replay dropped a reasoning signature carried separately from the function call. #897 was fixed for the `google-antigravity` replay path by #941. The `google-vertex` path may need the same class of signature preservation/replay, or it may currently bypass that replay cache.

This is a resubmission of #1248 using the repository's `Provider or API compatibility` issue template after maintainer guidance to use the issue template.

Environment used for the reproduction:

- Codex: `codex-cli 0.145.0`
- OS: Windows 11 x64 (`10.0.26200`)
- Model: `google-vertex/gemini-3.6-flash`

The Vertex project ID, service-account identity, credential path, request IDs, and conversation IDs are intentionally omitted.

### Checks

Ingwannu:
Closes #1254

## Summary

- preserve Vertex Gemini `thoughtSignature` values across Codex tool-result continuations by reusing the existing bounded TTL/LRU Google replay store
- scope Vertex replay identities by transport, project, location, and the hashed opaque session key so provider-private state cannot cross into Antigravity
- observe both streaming and non-streaming Vertex function-call responses, then re-inject only after Google tool names have been compiled
- document the shared Vertex/Antigravity contract in the architecture source of truth and all adapter-reference locales

## Root cause

The Google adapter wired signature observation and replay only for `googleMode: "cloud-code-assist"`. Vertex used the same Gemini function-call wire but bypassed the replay store, so the first call succeeded and the next tool-result request omitted the required opaque signature and received HTTP 400.

## Verification

- `bun test tests/google-vertex-thought-signature.test.ts tests/google-antigravity-replay.test.ts tests/google-antigravity-wire.test.ts tests/google-adapter.test.ts` — 94 pass, 0 fail
- `bun run typecheck`
- `bun run privacy:scan`
- `bun run --cwd docs-site build` — 221 pages
- `git diff --check`
- `bun run test` was also exercised under a two-core affinity. The changed Google suites passed; the host run exposed an unrelated existing `codex-shim` service-token path isolation failure. The initially missing GUI dependency was installed and its affected provider-rail suite then passed. Clean GitHub CI remains the authoritative full-suite gate.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Vertex AI now preserves and reuses Gemini thought signatures across tool calls, supporting both streaming and non-streaming responses.
  * Reasoning continuity is maintained after tool results for Vertex AI and Antigravity.
  * Replay data is isolated across transport, project, location, and model contexts.

* **Documentation**
  * Updated English, Japanese, Korean, Russian, and Simplified Chinese documentation to describe Vertex support.
  * Added details on signature replay behavior and session handling.


Wibias:
Not gui

Issue #1230, opened 2026-08-07
[Bug]: Concurrent ocx start restores journal before checking the existing healthy proxy


ShannwYang:
## Client or integration

Codex App

## Area

Service lifecycle

## Summary

A concurrent or stale `ocx start` can restore the Codex journal **before** checking whether the PID-file owner is an already healthy OpenCodex proxy.

This removes the active proxy's managed `openai_base_url` and `model_catalog_json` from `~/.codex/config.toml` while the real proxy continues serving on port 10100. The result is a split-brain state:

```text
OpenCodex proxy: healthy and listening
Codex config: restored to native routing
Generated catalog file: still present
Codex App: routed model list disappears and new turns bypass the proxy
```

This is related to the journal mechanism discussed in #477, but it is a different bug. #477 fixed stale journal snapshots; this issue occurs even with a correct recent journal because reconciliation happens before live-proxy detection.

## Reproduction

1. Start OpenCodex normally on loopback port 10100.
2. Confirm `~/.codex/config.toml` contains the managed `openai_base_url` and `model_catalog_json` entries.
3. Trigger another `ocx start`/autostart invocation while the first proxy remains healthy. A stale/concurrent launcher may also first lose a port race.
4. Observe the launcher reconcile an apparently dead journal owner, then discover the real healthy proxy.
5. Observe that the proxy remains healthy but the managed Codex configuration has been restored away.

Observed log sequence:

```text
Port 10100 stayed busy; refusing to hop to an ephemeral port.
Previous session (PID 95529) did not shut down cleanly. Codex state restored from journal.
Proxy already running (PID 95563, port 10100). Use 'ocx stop' first.
```

After this sequence:

```text
GET http://127.0.0.1:10100/healthz -> 200, PID 95563
GET http://127.0.0.1:10100/v1/models -> 16 models
~/.codex/opencodex-catalog.json -> exists with 16 models
~/.codex/config.toml -> no openai_base_url or model_catalog_json
```

The service log also contained frequent repeated `Proxy already running` messages from autostart invocations, making the race reproducible in normal App use rather than only a manual two-terminal test.

## Root cause

In `src/cli/index.ts`, `handleStart()` performs journal reconciliation before it reads the PID and checks the existing proxy:

```ts
const requestedPort = parsePortOption();
if (!currentExternalCodexModelProvider()) reconcileJournal();
const existingPid = readPid();
if (existingPid) {
  const live = await findLiveProxy();
[...]

PR #1269, opened 2026-08-08
fix(cli): check live proxy before journal recovery

Ingwannu:
## Summary

* Check the PID-file owner and the live proxy before reconciling a stale Codex journal during `ocx start`.
* Exit without mutating Codex integration state when another healthy OpenCodex proxy already owns the start lifecycle.
* Keep stale-PID cleanup and journal recovery in place for genuine replacement starts.
* Add an ordering regression that pins the ownership check ahead of journal recovery.

Closes lidge-jun/opencodex#1230

## Verification

* `taskset -c 0-1 bun test tests/cli-start-journal-order.test.ts tests/codex-journal.test.ts tests/proxy-liveness.test.ts tests/stale-state-purge.test.ts` — 105 passed.
* `taskset -c 0-1 bun run typecheck` — passed.
* `taskset -c 0-1 bun run privacy:scan` — passed.
* `taskset -c 0-1 bun run test` — 9,765 passed and 10 skipped; the local all-suite run also retained unrelated baseline/environment failures in the Codex shim token fixture, Live sideband close timing, and GUI dependency setup. The focused lifecycle suites above are green; exact-head CI remains authoritative.

## Checklist

- [X] Docs or release notes were updated when needed.
- [X] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup handling when another healthy proxy is already running.
  * Prevented competing startup attempts from modifying the active proxy’s configuration.
  * Correctly removes stale process state before reconciling startup data.
* **Tests**
  * Added coverage to verify the expected startup sequence and proxy detection behavior.

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -222,7 +222,6 @@ async function handleStart(options: { block?: boolean } = {}) {
   const serviceToken = loadServiceTokenFromFile(process.env);
   if (serviceToken) process.env.OPENCODEX_API_AUTH_TOKEN = serviceToken;
   const requestedPort = parsePortOption();
-  if (!currentExternalCodexModelProvider()) reconcileJournal();
   const existingPid = readPid();
   if (existingPid) {
     const live = await findLiveProxy();
@@ -232,6 +231,10 @@ async function handleStart(options: { block?: boolean } = {}) {
     }
     removePid(existingPid);
   }
+  // A losing concurrent start must not restore the active proxy's Codex config.
+  // Establish that the PID-file owner is stale before reconciling a dead journal;
+  // a healthy owner exits above without changing integration state (#1230).
+  if (!currentExternalCodexModelProvider()) reconcileJournal();
 
   // Interactive-only update prompt. Must run BEFORE we bind a port / write a
   // PID: choosing "Update now" installs globally and exits, so we never want a
--- a/tests/cli-start-journal-order.test.ts
+++ b/tests/cli-start-journal-order.test.ts
@@ -1,0 +1,32 @@
+import { describe, expect, test } from "bun:test";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+
+const source = readFileSync(join(import.meta.dir, "..", "src", "cli", "index.ts"), "utf8");
+
+function handleStartSource(): string {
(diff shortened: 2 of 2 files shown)


lidge-jun:
Adopting this — the `handleStart` reordering is exactly right.

One gap: `handleEnsure` still has the same defect. On current `dev` it calls `reconcileJournal()` at `src/cli/index.ts:441`, before the liveness check at `:447`, so the autostart path keeps the destructive ordering your fix removes from `handleStart`. Issue #1230 would only be half closed.

Could you extend the fix to `handleEnsure`? If you would rather keep this PR scoped, say so and I will follow up separately — either way #1230 stays open until both are covered.

On the test: the `indexOf` ordering check is a reasonable minimum, but it stops meaning anything if someone extracts `reconcileJournal()` into a helper. A behavioural test with isolated `OPENCODEX_HOME`/`CODEX_HOME`, a dead journal PID and a separately started healthy proxy would survive that. The negative control matters most — dead PID with no listener must still reconcile — since without it the test cannot tell reordering from removal.

Wibias:
no gui

lidge-jun:
Landed on `dev` as d3015cb97 with your authorship preserved, plus one corrective commit on top (700803f34).

The defect was real — `src/cli/index.ts:229-237` reconciled journal state before `readPid()`/`findLiveProxy()`. But the reorder covered only `handleStart`, while `handleEnsure` at `:445-451` had the identical defect, so autostart would still have destroyed state owned by a healthy proxy while the issue looked closed.

The correction shares one ownership/liveness preflight across both paths: a healthy owner returns before reconciliation, and a dead owner has only its own matching stale PID removed before recovery. The source-offset assertions were replaced with subprocess behavioral tests, since comparing line positions cannot show that state is actually preserved.

Red ablation against your head: 0 passed / 2 failed — healthy `ensure` restored the original config instead of preserving the owner state, and dead-owner `ensure` left the stale PID file behind. After the correction, 5 passed / 0 failed across the ownership and shutdown suites, and the full suite is green (10265 pass, 0 fail).

Thanks — the ordering insight was the hard part; extending it to the second call site was mechanical once you had named it.

Issue #1196, opened 2026-08-07
[Bug] Issue-quality media normalization accepts wrapped placeholders and can corrupt protected code


luvs01:
### Client or integration

GitHub issue-quality automation and issue template validation

### Area

CI and GitHub automation

### Summary

The issue-quality normalizer has two related media-boundary defects:

1. A placeholder wrapped in an HTML media element can be treated as substantive. For example, `clean("<video>No response</video>")` retains the wrapper instead of classifying the section as an empty placeholder.
2. In a multiline media block, indented child tags such as `<source>` or `<img>` can be protected as ordinary indented code before the media block is stripped, so a media-only section is not recognized consistently.

A direct combination of the obvious fixes is unsafe. The current indented-code protection restores sentinels by line position, while media stripping can collapse several lines to one. That can leak a NUL sentinel or drop an unrelated indented line before/after the media block.

Expected: placeholders wrapped only in media markup are rejected, real captions/fallback prose stay substantive, and fenced or indented code is preserved byte-for-byte.

### Reproduction

Run the current issue-quality core against these cases:

```js
clean("<video>No response</video>")
// currently remains substantive

isMediaOnly("<picture>\n    <source src=\"x\">\n    <img src=\"x\">\n</picture>")
// inconsistent because child lines are protected as indented code
```

A line-accounting regression appears when a stripped multiline media block is surrounded by protected lines:

```md
    ![before](url)
<video>
    <source src="x">
</video>
    ![after](url)
```

If multiline media is collapsed before positional restoration, the later sentinel can be restored to the wrong line or remain as a NUL byte.

### Version

Current `dev` at `44dce33a`; also reproducible from the current issue-quality core split

### Operating system

All platforms

### Provider and model

Not provider-specific

### Logs or error output

```text
direct media-wrapped placeholder: accepted as substantive
multiline media + neighboring indented code: positional sentinel restoration can lose a line or retain NUL
```

### Screenshots and supporting files

No screenshot is needed; the string fixtures above reproduce the parser behavior.

### Redacted configuration

```json
{
  "configuration_dependency": "none"
}
```

### Checks

- [x] I searched existing issues and documentation.
- [x] I removed secrets, tokens, account details, request credentials, and personal data.

### Root cause and proposed boundary

`protectIndentedCodeLines`, HTML media stripping, and placeholder detection operate as separate text rewrites but assume line positions remain stable. Media removal violates that assumption.

A safe implementation should:

[...]

PR #1270, opened 2026-08-08
fix(ci): preserve code while normalizing issue media

Ingwannu:
## Summary

* treat exact placeholder fallback text inside HTML media as empty
* preserve fenced and indented code with ordered restoration tokens
* avoid positional restoration after multiline media stripping changes line count

## Root cause

The issue-quality normalizer protected every indented line before it identified HTML media blocks. Indented `<source>` and `` children were therefore preserved as Markdown code, keeping otherwise media-only blocks substantive. Restoration also depended on original line indexes even though stripping a multiline media block can remove lines. Separately, placeholder fallback text such as `<video>No response</video>` was treated as a real caption.

The fix distinguishes indented children of an active HTML media block from literal Markdown code, protects fenced and indented examples with collision-resistant ordered tokens, and restores those tokens independently of line count. Only exact placeholder fallback values are removed; real captions remain substantive.

## Validation

* `node --test .github/scripts/issue-quality.test.cjs` — 112 passed
* related issue-quality / PR-quality / target-enforcement tests — 211 passed
* `bun run typecheck` — passed
* `bun run privacy:scan` — passed
* `git diff --check` — passed

Closes lidge-jun/opencodex#1196

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of media markup in issue content, including nested picture, audio, and video elements.
  * Removes media blocks with placeholder-only fallback text while preserving meaningful captions and prose.
  * Correctly preserves fenced and indented code examples, including media-like markup.
  * Prevents formatting artifacts when content is processed.

### DIFF
--- a/.github/scripts/issue-quality-core.cjs
+++ b/.github/scripts/issue-quality-core.cjs
@@ -66,45 +66,78 @@ function isPlaceholderOnlyValue(raw) {
  */
 function stripMediaTokens(text) {
   if (typeof text !== "string") return "";
-  // Indented code lines render as literal code in GitHub Markdown. Protect
-  // them first so neither the HTML nor the Markdown media stripper can
-  // remove example syntax; restore the lines afterwards.
+  // Fenced and indented code render literally in GitHub Markdown. Protect
+  // them first so neither media stripper can remove example syntax. The
+  // protector deliberately leaves indented children of an unindented HTML
+  // media block visible: those lines are HTML children, not Markdown code.
   const protectedText = protectIndentedCodeLines(text);
   const markdownStripped = stripMarkdownImages(stripHtmlMedia(protectedText.text));
   const referenceStripped = stripReferenceImages(markdownStripped);
-  return restoreIndentedCodeLines(referenceStripped, protectedText.lines);
+  return restoreIndentedCodeLines(referenceStripped, protectedText);
 }
 
 /**
- * Replace every indented code line (4+ leading spaces or a tab) with a
- * placeholder of equal length so media stripping cannot touch it. Returns the
- * masked text plus the original lines for restoration.
+ * Replace fenced code and indented code outside HTML media blocks with opaque
+ * tokens. Restoration is token-based rather than line-position-based because
+ * stripping a multiline media block may collapse or remove lines.
  */
 function protectIndentedCodeLines(text) {
   const lines = [];
+  let markerPrefix = "\u0000OCX_ISSUE_CODE_";
+  while (text.includes(markerPrefix)) markerPrefix += "_";
+  let mediaDepth = 0;
+  let fence = null;
+
+  const mask = (line) => {
+    const index = lines.push(line) - 1;
+    return `${markerPrefix}${index}\u0000`;
+  };
+
   const masked = text.split("\n").map((line) => {
-    if (/^(?: {4,}|\t)/.test(line)) {
-      lines.push(line);
-      return "\u0000" + line.replace(/[^\n]/g, " ").slice(1);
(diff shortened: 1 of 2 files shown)


lidge-jun:
Thanks for this — the token-based restoration is the right shape, and it fixes the line-position fragility properly.

One blocker before it can land, and it is the same one that stopped my own attempt at this: the placeholder check is too broad for media fallback text.

`isPlaceholderOnlyValue` is a whole-form-field predicate. Inside a media element it also matches `TBD`, `N/A`, `None`, and `Todo`, not just the generated `No response`:

```
                     dev            this PR
<video>TBD</video>   preserved   ->  ""
<video>N/A</video>   preserved   ->  ""
<video>None</video>  preserved   ->  ""
```

That flips a legitimate submission. A feature request whose example is `<video src="...">TBD</video>` is currently accepted and becomes media-only with this change, so the section reads empty and `enforce-issue-quality` closes it. Silencing a real report is worse than the placeholder leak we are fixing.

What I would suggest instead: a media-specific predicate limited to the serialization GitHub actually emits for an unanswered field — `No response`, plus its `_..._` and `*...*` markdown wrappers — rather than reusing the form-field vocabulary. Or an explicit mode on the shared helper so the two contexts stay distinguishable.

Worth adding negative tests for `TBD`, `N/A`, `None`, a non-English caption, an emoji, a bare filename, and a URL — those are the shapes that separate "the form said nothing" from "the reporter wrote something short".

The restoration half I have no concerns about. It is only the one predicate.

lidge-jun:
One more, found while verifying the first: the depth scanner only recognises a media tag once its `>` lands on the same line.

`htmlMediaDepthDelta` matches `/<(\/)?(picture|video|audio)\b[^>]*>/`, so an opening tag split across lines never increments the depth, the children are not treated as inside HTML, and their indentation gets protected as code. Reproduced on this branch:

```
<video src="https://x/a.mp4">        <video
    <source srcset="a.webm">           src="https://x/a.mp4"
    <img src="a.png">                  controls
</video>                             >
                                         <source srcset="a.webm">
                                         <img src="a.png">
                                     </video>

isMediaOnly -> true                  isMediaOnly -> false
```

Same content, same indentation; only the opening tag is wrapped. GitHub wraps attributes like this when a reporter pastes an embed from an editor that formats HTML, so it is not a contrived shape.

Carrying opening-tag state across lines would cover it. Worth a regression alongside the two-space / four-space / tab cases.

Neither this nor the predicate issue touches the restoration work, which is the substantial part of the PR and looks right to me.

lidge-jun:
Landed on `dev` as 96141e8f0 with your authorship preserved, plus one corrective commit on top (d493c147a).

Token-based restoration was the right call and your tests covered the original corruption path. The blocker was that `stripHtmlMedia` reused the broad whole-field placeholder predicate, so it erased legitimate captions. Verified by ablation against your head: `<video>TBD</video>` collapsed to `" "` and was classified media-only, and `N/A` and `None` were erased the same way. Since the gate auto-closes issues it judges empty, that would have started closing valid reports.

The correction narrows the predicate to GitHub's exact `No response` serialization and its wrappers, and carries multiline opening-tag state until `>` so indented media children are handled. New regressions cover `TBD`, `N/A`, `None`, Korean text, emoji, filenames, URLs, and a multiline opening tag. After the fix all three captions survive verbatim while the multiline media block is still stripped: 112 passed / 0 failed across 16 suites, and the full suite is green (10265 pass, 0 fail).

Thanks — this one was close to shipping a quiet issue-eating bug, and the restoration approach you built is what made the narrow fix possible.

PR #1281, opened 2026-08-08
feat(routing): fail over between policy candidates on retryable failures

Wibias:
## Summary

This PR adds policy-level fallback after the initially selected routing-profile candidate fails before useful output escapes.

Design constraints:
- keep the initial policy decision trace immutable
- record physical fallback attempts in the existing `attempts[]` history
- retry only failures already classified as hop-worthy by the combo failure policy
- never switch candidates for client cancellation or terminal invalid-request failures
- choose the next eligible candidate deterministically by the original policy score, with original candidate order as the tie-break

## TDD state

The first commit intentionally adds failing tests for the missing behavior. The implementation commit will follow after CI confirms the red state.

## Scope

- Responses execution wrapper for policy fallback
- focused regression tests
- no routing-profile schema changes
- no combo behavior changes

Base inspected: `dev` at `3ad5bb6bd3f76f6879d84b78ea39edd3e01ec296`.

## Summary by CodeRabbit

* **New Features**
  * Added automatic policy fallback for eligible requests after retryable pre-response failures.
  * Fallback tries suitable alternatives in ranked order while preserving the original request and policy context.
  * Retry attempts are recorded without replacing the original policy trace.
  * Fallback stops safely for terminal client errors, aborted requests, unreadable response details, or responses that have already started.

* **Tests**
  * Added coverage for candidate ranking, retry behavior, request history, and fallback exclusions.

### DIFF
--- a/src/server/responses.ts
+++ b/src/server/responses.ts
@@ -5,5 +5,6 @@ export type { MultiAgentGuidanceOptions, MultiAgentGuidanceDeps } from "./respon
 export { hasUnreadableEncryptedAgentTask, sanitizeEncryptedContentInPlace } from "./responses/encrypted-payload";
 export { COMPACT_RESPONSE_MAX_BYTES, bufferCompactResponse, handleResponsesCompact } from "./responses/compact";
 export { disableResponsesRequestTimeout, safeHostLabel, fetchWithHeaderTimeout } from "./responses/fetch-helpers";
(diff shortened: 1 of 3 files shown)


Wibias:
Review follow-up: the `accountRef` thread does not apply to live routing-profile fallback. `OcxRoutingProfileConfig.candidates` are provider/model targets; account choice remains inside the existing provider/account-pool layer and live policy traces do not create account-qualified policy candidates. Falling back via explicit `provider/model` is intentional here. The separate pool-evidence PR keeps that boundary rather than promoting account IDs into policy candidate identity.

PR #1282, opened 2026-08-08
feat(routing): use pool-aware quota evidence for policy profiles

Wibias:
## Summary

Use Codex pool capacity when scoring an OpenAI policy candidate instead of treating the currently active account as the whole provider.

Expected semantics:
- use the best known usable quota headroom across the live pool
- mark the pool exhausted only when every relevant account is known exhausted
- if all known accounts are exhausted but another account is unknown, keep pool quota unknown rather than falsely excluding it
- keep account identities out of route traces
- keep the existing account pool responsible for the physical account choice
- share the same evidence helper between live routing and profile dry-runs

## TDD state

The first commit intentionally adds failing tests for the missing pool aggregation helper. Implementation follows after the red state is confirmed.

## Scope

- quota evidence only; Codex pool health is already pool-aware on current `dev`
- no routing-profile schema changes
- no account-pool strategy changes
- exact account selectors remain unchanged

Base inspected: `dev` at `3ad5bb6bd3f76f6879d84b78ea39edd3e01ec296`.

## Summary by CodeRabbit

* **New Features**
  * Added Codex pool quota evaluation across multiple accounts.
  * Plans now apply the appropriate quota windows for each account.
  * Routing can report available pool headroom when at least one account remains usable.
  * Preserves unknown quota status when account data is unavailable.
  * Retains account-level evaluation when pool aggregation is not applicable.
* **Bug Fixes**
  * Improved quota handling for exhausted pools, reset times, and partially unavailable account information.

### DIFF
--- a/src/routing/quota.ts
+++ b/src/routing/quota.ts
@@ -24,11 +24,8 @@ import type { RouteQuotaEvidence } from "./trace";
 export interface QuotaEvidenceInput {
   provider: string;
   model: string;
-  /** Opaque account reference for per-account quota sources. */
   accountRef?: string;
-  /** Codex pool account id (provider "openai"). */
   codexAccountId?: string;
-  /** The account's plan (provider "openai"); selects the governing quota window. */
   codexAccountPlan?: string;
 }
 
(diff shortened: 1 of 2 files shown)


Wibias:
Review follow-up: the credits-only finding was valid and is fixed in `9d7a581` + `b31abdd` with a regression test. I re-checked the broader membership comment against the existing state-store contract: `listAccountQuotas()` is the reconciled quota snapshot and `reconcileCodexQuotaAccounts()` prunes IDs outside the live config generation. Pause/reauth/cooldown/soft-avoid remain separate health/pool-selection dimensions by design; duplicating the selector inside quota scoring would couple two evidence dimensions and drift from the existing pool authority. The PR therefore keeps quota aggregation over the reconciled quota snapshot rather than introducing a second account-eligibility implementation.

PR #1106, opened 2026-08-06
feat(toggles): Codex restore truth, Claude Desktop switch, composed acceptance (WP13 resume)

lidge-jun:
## Summary

Resumes the paused WP13/WP14 tail of the Codex write-substrate campaign: the Codex CLI toggle gets artifact-level restore truth, Claude Desktop gets a real ON/OFF toggle with a documented standard-mode pivot, and a composed acceptance suite proves the production entry points reach the substrate instead of writing around it.

- **Codex toggle completion** (`devlog/_plan/260803_codex_desktop_toggle/040` as amended by `260806_wp13_toggles_resume/010`): `ocx restore`/`eject` persist desired OFF before mutating (crash-durable), `restore back`/`eject back` persist ON; `restoreNativeCodexAsync` returns a per-artifact `{config, catalog, history}` envelope where history failure carries a `busy | permission` discriminator instead of being folded into success; every `syncModelsToCodex` caller (startup, ensure, sync, restore dispatch, models, provider, `POST /api/sync`, toggle enable) branches on a discriminated `applied/skipped/refused` status; desired state is re-read *inside* each artifact serializer (write lock, history serialization, catalog commit, cache reacquisition) so a lost race becomes a typed skip, never a stale write.
- **Claude Desktop toggle** (`050` as amended by `020`): `claude-desktop` joins `clientIntegrations` and the native route union; a read-only inspector classifies `not_installed/standard/gateway_ours/gateway_drifted/foreign/no_owned_state/broken/unsafe` with an ownership boundary (owned-but-drifted refuses as unsafe; reads never create files); OFF pivots to a credential-free `{}` standard profile *before* removing the opencodex profile, its backup, and its metadata row; all four writer paths (auto-apply, native enable, `POST /apply`, no-daemon CLI apply) re-read persisted intent after their awaits; `desiredEnabled` is required on post-commit refusals for every client; GUI switch with six-locale copy.
[...]

PR #1291, opened 2026-08-08
feat: add enable/disable toggle for Claude Desktop routing

Splaxi:
Adds an on/off toggle to the Claude Desktop tab, mirroring the Claude Code connection toggle. When off, the 3P config files are removed from Claude Desktop's config library; when on, they are written back from the saved profile.

Server:
- src/types.ts: add desktopEnabled flag to OcxClaudeCodeConfig
- src/claude/desktop-3p.ts: new clearDesktop3pConfig() to remove the opencodex entry from Claude Desktop's _meta.json and delete the config JSON file
- src/server/management/native-integration-routes.ts: add claude-desktop to native integration client IDs, desktopStatus() probe, and handleDesktopToggle() PUT route with single-flight guard
- src/server/management/agent-settings-routes.ts: include enabled field in GET /api/claude-desktop and /status responses

GUI:
- gui/src/pages/ClaudeDesktop.tsx: Switch toggle above the toolbar; when OFF, hides Save & Apply (only Save remains) and shows a disabled notice
- gui/src/pages/integrations/native-api.ts: add claude-desktop to valid client IDs
- gui/src/pages/integrations/integration-api.ts: extract enabled from the status payload
- gui/src/pages/integrations/overview-clients.ts: add enabled? to ClaudeDesktopPayload

i18n: 4 new keys (toggleFailed, enabledLabel, toggleAria, disabledNotice) in all 6 locales

## Summary

I had big issues when trying to things, and couldn't remove the 3rd party / gateway in Claude Desktop. Had to locate the files, and simply delete them. This is to make it more user friendly.

## Verification

image

image

Take it as inspiration - and reshape it until it fits with what you want backed into the solution

## Summary by CodeRabbit

* **New Features**
  * Added Claude Desktop routing controls, including enable/disable status, toggle feedback, and accessibility labels.
  * Disabling routing now removes the integration configuration and displays a notice.
  * Save options update based on routing status, hiding “Save & Apply” when disabled.
* **Localization**
  * Added Claude Desktop routing messages in English, German, Japanese, Korean, Russian, and Chinese.


lidge-jun:
Thanks for taking the time to build this, and for the note that it was meant as inspiration rather than a final shape.

This capability already shipped on `dev`. The Claude Desktop enable/disable toggle landed in `7f1220b4c feat(claude): add Desktop native integration toggle` (PR #1106, 2026-08-06) and is released in `v2.11.0` and `v2.11.1`, so your branch is reimplementing an existing subsystem rather than adding a missing one. I am closing this as already implemented — details below so you can find and use the existing control.

## Where the feature lives today

**Using it:** the toggle is on the Integrations overview page, on the Claude Desktop row — not inside the Claude Desktop tab itself. Turning it off pops a confirmation dialog that names the exact config path being changed, then reverts Claude Desktop to standard mode and clears the opencodex entry. Turning it back on regenerates the config from the saved profile and the current visible model list.

**Server:**

- `src/types.ts` / `src/config.ts` — the flag is `clientIntegrations["claude-desktop"]`, on the same axis as `claude`, `grok`, and `codex`.
- `src/server/management/native-integration-routes.ts` — `claude-desktop` is in `NativeIntegrationClientId`, `desktopStatus()` reports the probe, and `PUT /api/native-integrations/claude-desktop` performs the toggle via `handleClaudeDesktopToggle()`.
- Single-flight is guarded by `claudeDesktopToggleFlight`, which returns `409 config_busy` on a concurrent write.
- Disable calls `removeDesktop3pStandardPivot()` in `src/claude/desktop-3p.ts`, which pivots Desktop back to standard mode and reports `cleanup_incomplete` with residual paths, or `metadata_unreadable`, instead of failing silently.
- `src/server/management/agent-settings-routes.ts` — `GET /api/claude-desktop` and `GET /api/claude-desktop/status` already carry the desired-enabled state.

**GUI:**

- `gui/src/pages/integrations/overview-clients.ts` — `claudeDesktopRow()` builds the row with `toggle: "claude-desktop"`.
- `gui/src/pages/integrations/IntegrationsOverview.tsx` — the toggle handler routes `claude-desktop` through `toggleNativeIntegration()`, with `DESKTOP_DISABLE_COPY` for the disable confirmation.
- i18n keys exist under `claudeDesktop.*` across all six locales.

**Tests:** `tests/native-claude-desktop-toggle.test.ts`, plus the follow-up hardening in `f4607d5f6`, `1208f46b6`, `67ebfb78e`, `bdd12a94f`, and `d9d555437`.

## Why this branch cannot be merged as-is

The design differs in one way that matters. This PR introduces `desktopEnabled` inside `OcxClaudeCodeConfig` and a separate `clearDesktop3pConfig()` that deletes the config JSON directly. `dev` stores the same intent in `clientIntegrations["claude-desktop"]` and routes removal through the standard-mode pivot. Merging both would leave two independent sources of truth for one user-visible switch, and the direct-delete path skips the residual-credential accounting the pivot performs.

## On the problem that prompted this

You mentioned you could not remove the third-party gateway from Claude Desktop and had to delete the files by hand. If that happened on `v2.11.0` or later, the toggle was present and something went wrong — that is a bug worth its own report. Please open a Bug report with your opencodex version, OS, and what the Claude Desktop row showed at the time, and we will look at it directly. If you were on an earlier version, upgrading gets you the control.

Splaxi:
Thank you for taking the time to look into. I'll upgrade to v2.11.0 and test again 🔥

PR #1300, opened 2026-08-08
fix(catalog): preserve nested reasoning efforts

dbc-hbin:
## Summary

- Preserve `reasoning_parameters.efforts` emitted by OpenAI-compatible `/models` endpoints.
- This fixes effort controls disappearing in downstream client exports such as Oh My Pi even when the upstream explicitly advertises a valid per-model ladder.

## Verification

- `bun test tests/provider-model-discovery-contract.test.ts`
- `bun run typecheck`
- `bun run test`
- `bun run privacy:scan`
- Live synthetic-provider verification: the advertised `low`, `high`, and `max` values now appear in `omp models`.

## Checklist

- [x] Docs or release notes were updated when needed (not needed for an internal metadata parser fix).
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults; no credentials are stored or emitted.

## Summary by CodeRabbit

* **New Features**
  * Model discovery now recognizes and preserves available reasoning-effort options from compatible catalogs.

* **Bug Fixes**
  * Improved handling of nested model metadata so reasoning capabilities are accurately surfaced.

### DIFF
--- a/src/codex/catalog/provider-fetch.ts
+++ b/src/codex/catalog/provider-fetch.ts
@@ -923,7 +923,16 @@ export function catalogHintsFromModelsApiItem(providerName: string, item: Provid
       item.max_context_length,
     );
   const maxInputTokens = positiveSafeInteger(limits?.max_input_tokens, item.max_input_tokens);
-  const rawReasoningEfforts = capabilityRecord?.reasoning_effort ?? item.reasoning_efforts;
+  // Some OpenAI-compatible catalogs expose the selectable ladder under
+  // `reasoning_parameters.efforts` instead of the older `reasoning_efforts` key.
+  // Treat both as model metadata: otherwise a valid upstream capability disappears
+  // before client exporters (including omp) can advertise it.
+  const reasoningParameters = plainRecord(item.reasoning_parameters)
+    ?? plainRecord(metadata?.reasoning_parameters)
+    ?? plainRecord(capabilityRecord?.reasoning_parameters);
+  const rawReasoningEfforts = capabilityRecord?.reasoning_effort
+    ?? item.reasoning_efforts
+    ?? reasoningParameters?.efforts;
   const listedReasoningEfforts = normalizedStringList(rawReasoningEfforts, 8, 24);
   const reasoningEfforts = listedReasoningEfforts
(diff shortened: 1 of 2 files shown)


lidge-jun:
Approved the pending Cross-platform CI run at this exact head, and it came back **green**.

Worth explaining the silence beforehand: contributor PRs queue their workflow runs as `action_required` until a maintainer releases them, and `gh pr checks` does not show that state — so from your side it looks like nothing is happening rather than like something is waiting on us. That was ours to clear, not yours.

To be precise about what this does and does not do: it only unblocks CI. It does **not** make this mergeable. The four-box readiness checklist in the description is your attestation, and the gate keeps the PR in draft until you complete it. With CI now green at your head, box 1 is provable.

If anything in the run looks wrong to you, say so and I will dig into it rather than leaving you to guess.

Ingwannu (commented):
The implementation is small and directionally correct, and I verified the exact submitted head locally: typecheck passed and tests/provider-model-discovery-contract.test.ts passed 27/27.\n\nI am not approving or merging this head yet because the required GitHub CI is not green. Test shard 3/4 was cancelled only after hitting the 15-minute job limit; the annotations show Bun hanging in unrelated suites, not an assertion caused by this parser change. The branch is also now 31 commits behind current dev after other merges.\n\nPlease rebase this one-commit change onto current dev and rerun exact-head CI. If the required checks are green on that head, this remains a good merge candidate.

lidge-jun:
Landed on `dev` as 87aeeb7f0, rebased onto the current head with your authorship preserved.

Verified before merge: `src/codex/catalog/provider-fetch.ts:926` only read `capabilities.reasoning_effort ?? item.reasoning_efforts`, so a model exposing its ladder as `reasoning_parameters.efforts` lost it entirely. Routing the new fallback through the existing `normalizedStringList`/`sanitizeCodexReasoningEfforts` path keeps the bounds and sanitization intact rather than adding a second validation route. Full suite green on the rebased head (10175 pass, 0 fail).

Thanks.

dbc-hbin:
Thanks for landing this with authorship preserved and for the detailed verification. I confirmed `87aeeb7f0` is now contained in `dev`; no further action is needed from this branch.

PR #1301, opened 2026-08-08
test(ci): bind the Windows shard assertion to an executable command (#1185)

lidge-jun:
## Summary

Republishes @luvs01's #1185 on current `dev`. Their branch was 324 commits behind, so this is a maintainer rebase; the first commit is theirs via `Co-authored-by`.

The assertion that the Windows leg shards the suite used `.includes()` against the step's `run` text, so any occurrence of the command anywhere in the script satisfied it — inside an `echo`, or in a comment. A Windows job that printed the command instead of running it kept the suite green. `hasExactShellCommand` splits the script into lines, drops blanks and comments, and requires the command as a whole line; the negative assertion against `echo <command>` pins that so a future loosening back to substring matching fails here rather than silently.

**This PR adds one maintainer-authored commit that was not in #1185** (`test(ci): require the Windows test step to be unconditional`). Binding the assertion to an executable line closes the `echo` hole but not the neighbouring one: a step carrying the exact command still runs nothing under `if: false`, and the suite would stay green against a Windows leg that never tests. It is a separate commit rather than folded in, so the added coverage is not attributed to the contributor.

Supersedes #1185, which can be closed once this lands.

## Verification

Rebased onto `14e948525` and verified after the rebase:

- `bun run test` — **10009 pass / 7 skip / 0 fail** across 626 files
- `bun test tests/ci-workflows.test.ts` — 125 pass / 0 fail
- `bun run typecheck` — clean

Two ablations against `.github/workflows/ci.yml`, each restored afterwards. Both mutations neuter the Windows test leg, and **neither is caught by `dev` today**:

| Mutation to the Windows Test step | `dev` today | with this PR |
|---|---|---|
| `run: bun test …` → `run: echo bun test …` | 125 pass / 0 fail | 124 pass / **1 fail** |
| add `if: false`, command unchanged | 125 pass / 0 fail | 124 pass / **1 fail** |

The first is the contributor's assertion; the second is the maintainer commit.

## Checklist

- [x] Docs or release notes were updated when needed. (Test-only change; no user-facing behavior.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (No workflow file is modified — this tightens the test that inspects them. The ablations above were local and reverted.)

---

Note on #1185's red CI: its `Cross-platform CI` failure at `bff31d1e0` is not caused by that diff. The shard crashed while loading `tests/autostart-health.test.ts` with `EEXIST: file already exists, epoll_ctl` in a Bun `WriteStream`, followed by a collateral `Cannot call describe() after the test run has completed`. #1185 touches only `tests/ci-workflows.test.ts` and has no path to that file. Root cause of the Bun-level crash is unknown and not diagnosed here.

## Summary by CodeRabbit

- **Tests**
  - Improved continuous integration checks for Windows test execution.
  - Validation now confirms the required sharded test command is present, exact, and runs unconditionally.
  - Prevents misleading matches caused by commented or echoed command text.

### DIFF
--- a/tests/ci-workflows.test.ts
+++ b/tests/ci-workflows.test.ts
@@ -39,6 +39,15 @@ function count(text: string, fragment: string): number {
   return text.split(fragment).length - 1;
 }
 
+/** Match an executable shell line, not a fragment that could appear in echo or a comment. */
+function hasExactShellCommand(run: string | undefined, expected: string): boolean {
+  return (run ?? "")
+    .split(/\r?\n/)
+    .map(line => line.trim())
+    .filter(line => line.length > 0 && !line.startsWith("#"))
+    .includes(expected);
+}
+
 function expectSecureLinuxKeyringBootstrap(workflow: string): void {
   const smokeStep = workflow
     .split("- name: OS keyring create/read/delete smoke")[1]
@@ -164,7 +173,9 @@ describe("GitHub Actions hardening", () => {
     // the runner's disk and the suite passes against a tree that no longer
     // exists in git.
     const winSteps = (ci.jobs?.["platform-windows"] as { steps?: { if?: string; run?: string }[] })?.steps ?? [];
-    expect(winSteps.some(step => step.run?.includes(`--shard=\${{ matrix.shard }}/${windowsShards.length}`))).toBe(true);
+    const windowsTestCommand = `bun test --isolate tests --shard=\${{ matrix.shard }}/${windowsShards.length}`;
+    expect(hasExactShellCommand(`echo ${windowsTestCommand}`, windowsTestCommand)).toBe(false);
+    expect(winSteps.some(step => hasExactShellCommand(step.run, windowsTestCommand))).toBe(true);
     expect(winSteps.some(step => step.if === "runner.environment == 'self-hosted'"
       && step.run?.includes("git clean -xffd"))).toBe(true);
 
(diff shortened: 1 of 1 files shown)


lidge-jun:
CI here is blocked by a base-branch problem, not by this diff, and I've stopped rerunning it.

Both attempts on `f09ef1557` ended `cancelled` with `test 4/4` hitting the 15-minute job timeout — the second ran 15:28:02Z to 15:43:17Z, so it is a real timeout rather than a superseded run. The three other Linux shards and the full macOS leg passed both times.

The stall is always the same shape: output stops right after a test that starts a proxy listener, then nothing for ~14 minutes, then cleanup reports `Terminate orphan process: pid (NNNN) (bun)`. Here the last line was `(pass) routed Claude requests give OpenAI sidecars main auth…` from `tests/claude-messages-endpoint.test.ts`, which passes locally in 2.7s (38/38), so it needs the sharded Linux runner to show up.

I had been calling this flake all day — five occurrences across four unrelated branches plus `dev` itself. The audit on my own work pushed back on that, correctly, so I've filed #1302 with the run inventory and the one instance that crashed instead of hanging (`EEXIST: epoll_ctl` in a Bun `WriteStream`), which looks like the same resource-lifecycle bug with a different outcome.

Holding this PR rather than rerunning until green. The change itself is test-only and verified locally (full suite 10009 pass, both ablations reproduced), but merging on a retry-until-green would be exactly the habit #1302 warns about.

lidge-jun:
Rebased onto current `dev` (`b5d44a534`) and force-pushed with a lease against the previous head.

Reason for the respin: the two runs I held this on both ended `cancelled` under the shard hang tracked in #1302, so there was never a real signal to judge it by. It had also drifted 5 commits behind. Both commits are preserved with their separation intact — @luvs01's assertion carries the `Co-authored-by` trailer, and the unconditional-step assertion stays a separate maintainer commit.

Re-verified on the rebased head rather than reusing the old numbers: full suite **10055 pass / 7 skip / 0 fail** across 627 files, `tests/ci-workflows.test.ts` 125 pass, typecheck clean, privacy scan passed. The `if: false` ablation still fails against current `dev`, so the coverage gap this closes is still open on the base branch.

Watching the new run. If it hangs again I will record it on #1302 rather than rerunning past it.

PR #1304, opened 2026-08-08
fix(providers): correct Qwen3.8 reasoning levels

XIQIXIQIXIQI:
## Summary

- Correct Alibaba Token Plan's Qwen3.8-Max reasoning ladder to the officially documented `low`, `medium`, and `xhigh` levels, with `xhigh` as the default.
- Route Qwen3.8-Max through the direct `reasoning_effort` field instead of the legacy Qwen `thinking_budget` contract in both Beijing and international presets.
- Repair stale persisted preset metadata at registry enrichment and routing time, including case-insensitive model keys and provider-wide aliases that could otherwise leak unsupported `high` or `max` values.
- Update the English and Simplified Chinese architecture references with the direct-effort exception.

The previous registry data classified Qwen3.8-Max like older Qwen3.x budget models. International metadata also exposed `high` while omitting the supported `medium` tier. Existing provider rows persist capability metadata, so correcting only the registry seed would leave upgraded installations on the stale wire contract.

This change makes new and existing Alibaba Token Plan configurations follow the [Qwen3.8-Max documentation](https://qwen.ai/blog?id=qwen3.8). Codex compatibility levels above the upstream ladder clamp to `xhigh` rather than reaching the provider as unsupported values.

## Verification

- `bun test tests/reasoning-effort.test.ts tests/alibaba-intl-token-plan.test.ts tests/provider-registry-parity.test.ts` — 89 passed, 0 failed.
- `bun run test` — 10,132 passed, 7 skipped, 0 failed across 631 files.
- `bun run typecheck`
- `bun run privacy:scan`
- `git diff --check`

## Checklist

## Review readiness checklist

### DIFF
--- a/docs-site/src/content/docs/reference/architecture.md
+++ b/docs-site/src/content/docs/reference/architecture.md
@@ -166,6 +166,12 @@ upstream providers may support only a smaller subset or require a real alias. Th
 - Resolves per-model and per-provider `reasoningEffortMap` overrides for custom wire mappings.
 - Drops the effort entirely for models listed in `noReasoningModels`.
 
+Qwen3.8-Max is an explicit direct-effort exception to the older Qwen3.x budget contract. Alibaba
+Token Plan records its upstream-supported ladder as `low`, `medium`, and `xhigh` (the default), and
+sends the effective value as `reasoning_effort`; Codex-only compatibility tops are clamped to
+`xhigh` on the wire. Runtime registry enrichment repairs older persisted preset metadata that still
+classifies this model as a `thinking_budget` model.
+
 ## Core types
 
 The internal model lives in `types.ts`: `OcxParsedRequest`, `OcxContext`, the `OcxMessage` union,
--- a/docs-site/src/content/docs/zh-cn/reference/architecture.md
+++ b/docs-site/src/content/docs/zh-cn/reference/architecture.md
@@ -157,6 +157,11 @@ Codex context compaction 同样适用于路由模型。`server/responses/compact
 - 解析模型级和 provider 级 `reasoningEffortMap` override，用于自定义 wire 映射。
 - 对 `noReasoningModels` 中的模型完全移除 effort。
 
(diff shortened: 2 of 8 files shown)


lidge-jun:
Approved the pending Cross-platform CI run at this exact head, and it came back **green**.

Worth explaining the silence beforehand: contributor PRs queue their workflow runs as `action_required` until a maintainer releases them, and `gh pr checks` does not show that state — so from your side it looks like nothing is happening rather than like something is waiting on us. That was ours to clear, not yours.

To be precise about what this does and does not do: it only unblocks CI. It does **not** make this mergeable. The four-box readiness checklist in the description is your attestation, and the gate keeps the PR in draft until you complete it. With CI now green at your head, box 1 is provable.

If anything in the run looks wrong to you, say so and I will dig into it rather than leaving you to guess.

lidge-jun:
Landed on `dev` as a1e418509 with your authorship preserved, plus one corrective commit on top (72ff55d9c).

Your finding was correct: `THINKING_BUDGET_EFFORTS` wrongly covered Qwen3.8-Max when the official contract uses direct `reasoning_effort` with `low`/`medium`/`xhigh`. What the patch also did, though, was delete explicit user configuration — your own test asserted it:

```js
modelReasoningEfforts: { "QWEN3.8-MAX": ["low", "high", "xhigh"] },
expect(provider.modelReasoningEfforts["QWEN3.8-MAX"]).toBeUndefined();
```

This repo keeps registry defaults *beneath* user entries, so a deliberately configured Qwen3.8 override has to survive. The correction delivers the right contract without removing user overrides, and the tests now assert the correct behavior: the stale generated default is fixed while an explicit per-model override is preserved. Red ablation confirmed both regressions fail without the fix (the uppercase budget entry gets deleted and the custom ladder gets replaced).

Full suite green on the landed head. Thanks — the contract research and the doc links were solid; only the ownership rule needed adjusting.

PR #1310, opened 2026-08-08
[Bug]: Preserve operator-owned static OAuth catalogs during reconciliation

### LINKED ISSUE #1308: [Bug]: ocx sync drops configured combo members and omits failover catalog entry
### Client or integration

Codex CLI

### Area

Catalog / models

### Summary

`ocx sync` can omit a configured failover combo when provider model discovery does not retain all configured members. The expected `combo/failover` catalog entry disappears even though the configured provider/model targets are callable and have metadata.

### Reproduction

1. Configure a failover combo with these targets:
   - `openrouter/openai/gpt-5.6-luna`
   - `opencode-go/deepseek-v4-flash`
   - `command-code/xiaomi/mimo-v2.5-pro`
2. Run:

   ```sh
   ocx sync --restart-codex
   ```
3. Observe:

   ```text
   Combo "failover" is omitted from the catalog because member capabilities are incomplete
   ```

   Earlier runs also reported configured OpenRouter model IDs being dropped when live discovery omitted them.

### Version

2.11.0

### Operating system

macOS arm64

### Provider and model

OpenRouter / `openai/gpt-5.6-luna`; OpenCode Go / `deepseek-v4-flash`; Command Code / `xiaomi/mimo-v2.5-pro`

### Logs or error output

```shell
[opencodex] Combo "failover" is omitted from the catalog because member capabilities are incomplete: openai/gpt-5.6-luna, opencode-go/deepseek-v4-flash
```

None.

### Redacted configuration

```json
{
  "combos": {
    "failover": {
      "strategy": "failover",
      "targets": [
        { "provider": "openrouter", "model": "openai/gpt-5.6-luna", "weight": 1 },
        { "provider": "opencode-go", "model": "deepseek-v4-flash", "weight": 1 },
        { "provider": "command-code", "model": "xiaomi/mimo-v2.5-pro", "weight": 1 }
      ]
    }
  }
}
```

### Checks

yusoofsh:
## Summary

- Preserve provider model and capability metadata when an operator explicitly sets `liveModels: false`.
- Prevent OAuth startup reconciliation from replacing an operator-owned static catalog with the registry preset.
- Keep legacy Antigravity migration behavior unchanged.
- Add regression coverage for a customized Command Code static OAuth catalog.
- Fixes #1308.

## Verification

- `bun run test tests/oauth-provider-reconcile.test.ts` — 6 passed.
- The full `bun run test` suite was started through Nub and reached broad coverage, but was stopped after unrelated GUI dependency warnings for missing `react` packages in the checkout; the focused reconciliation test and all relevant reported tests passed.

## Checklist

- [x] Docs or release notes were updated when needed (not user-facing API; no docs change needed).

## Contributor readiness

- [x] Local focused CI is green.
- [x] Branch is based on the latest available `dev` checkout.
- [x] All correct Codex and CodeRabbit findings are fixed.
- [x] Ready for review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved operator-configured OAuth model catalogs during preset reconciliation.
  * Prevented static model settings and context-window mappings from being unintentionally modified.
  * Reconciliation now correctly reports no changes when these catalogs are already configured.

### DIFF
--- a/src/oauth/index.ts
+++ b/src/oauth/index.ts
@@ -777,6 +777,11 @@ function migrateLegacyAntigravityStaticCatalog(config: OcxConfig): boolean {
   return true;
 }
 
+/** An explicit static catalog is operator-owned and must not be replaced by OAuth presets. */
+function isOperatorOwnedStaticCatalog(provider: OcxProviderConfig): boolean {
+  return provider.liveModels === false;
+}
+
 export function reconcileOAuthProviders(config: OcxConfig): boolean {
   let changed = migrateLegacyAntigravityStaticCatalog(config);
   for (const [name, prov] of Object.entries(config.providers)) {
@@ -788,6 +793,7 @@ export function reconcileOAuthProviders(config: OcxConfig): boolean {
       changed = true;
     }
     if (!def || prov.authMode !== "oauth") continue;
+    if (isOperatorOwnedStaticCatalog(prov)) continue;
     const preset = def.providerConfig;
     for (const field of OAUTH_RECONCILE_FIELDS) {
       if (JSON.stringify(prov[field]) === JSON.stringify(preset[field])) continue;
--- a/tests/oauth-provider-reconcile.test.ts
+++ b/tests/oauth-provider-reconcile.test.ts
@@ -116,6 +116,34 @@ describe("OAuth provider reconciliation", () => {
     expect(config.providers["google-antigravity"].liveModels).toBe(false);
   });
 
+  test("preserves operator-owned static OAuth catalog fields", () => {
+    const config = {
+      port: 10100,
+      defaultProvider: "command-code",
+      providers: {
+        "command-code": {
+          ...structuredClone(OAUTH_PROVIDERS["command-code"].providerConfig),
+          liveModels: false,
+          models: ["deepseek/deepseek-v4-flash", "xiaomi/mimo-v2.5-pro"],
+          modelContextWindows: {
+            "deepseek/deepseek-v4-flash": 1_000_000,
+            "xiaomi/mimo-v2.5-pro": 1_050_000,
+          },
+        },
+      },
+    } satisfies OcxConfig;
+
+    expect(reconcileOAuthProviders(config)).toBe(false);
+    expect(config.providers["command-code"].models).toEqual([
+      "deepseek/deepseek-v4-flash",
+      "xiaomi/mimo-v2.5-pro",
+    ]);
+    expect(config.providers["command-code"].modelContextWindows).toEqual({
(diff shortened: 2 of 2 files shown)


Wibias (changes requested):
## Verdict: Request changes

The Command Code fix is useful and the diff is small, but this is **not safe to merge as-is**.

### Blocker — `liveModels: false` is not reliable operator-ownership provenance

`isOperatorOwnedStaticCatalog()` now treats every `provider.liveModels === false` as operator-owned and skips all OAuth preset reconciliation.

That breaks **Kiro**. Kiro is an OAuth provider whose registry preset itself intentionally sets `liveModels: false` because its CodeWhisperer wire has no compatible live `/models` endpoint. `deriveOAuthProviderConfig()` materialises that registry value into the saved provider config. Therefore a normal, untouched Kiro config now looks “operator-owned” and will hit the early `continue` forever.

Result: existing Kiro users stop receiving future registry updates to its static model list and associated context/reasoning metadata. This directly conflicts with the Kiro preset's purpose: its static catalog is authoritative and must be maintained by OpenCodex as models/capabilities change.

Please distinguish **registry-owned false** from **operator-owned false**. At minimum, `false` cannot mean operator-owned when the current OAuth preset also has `liveModels: false`. If operators must also be able to freeze/customise Kiro's catalog independently, this needs explicit ownership/provenance rather than overloading the discovery flag.

### Contract risk — the guard freezes much more than model discovery

The early `continue` skips every field in `OAUTH_RECONCILE_FIELDS`, including context/output limits, input modalities, reasoning maps, `noVisionModels`, parameter restrictions, etc. Public config semantics describe `liveModels` as the switch for live catalog discovery; this patch silently turns `liveModels: false` into an opt-out from future bundled capability maintenance.

That can leave capability metadata stale after upgrades. In particular, stale vision metadata can change whether requests use native vision or the vision sidecar, which has behaviour/cost/privacy implications. I found no direct secret leak or auth bypass in this diff, but freezing these capability facts is too broad without an explicit contract.

### Tests needed

Please add coverage for at least:

- a normal Kiro registry-owned `liveModels: false` config with stale models/context/reasoning metadata still reconciling to the current preset;
- the intended Command Code operator-owned static catalog remaining untouched;
- the intended policy for the other capability fields currently skipped by the early `continue` (and whether stale `defaultModel` healing should also be skipped).

CI on the current head is green, including the test shards, typecheck, GUI tests, privacy scan and macOS job. I also found no new resource-retention/concurrency issue in the two-file diff. The blocker is semantic/provenance correctness, not CI hygiene.

**Verdict: REQUEST CHANGES.** Fix the ownership distinction before merge.

PR #1323, opened 2026-08-09
test(omp): close final #1109 path-contract review gap

Wibias:
Maintainer follow-up for #1109. #1109 is merged; this PR now contains only focused OMP path-contract regression coverage on top of current `dev`.

Current scope:
- slash- and tilde-prefixed `PI_CONFIG_DIR` values stay home-relative under OMP's `path.join(home, value)` contract
- named profiles ignore `PI_CODING_AGENT_DIR`
- tests use unique temporary home directories so legacy `models.yaml` files or parallel runs cannot affect the result

No production behavior changes. Merge only after fresh current-head CI is green.

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for resolving configuration paths relative to the home directory.
  * Added coverage ensuring configured paths take priority when selecting named profiles.

### DIFF
--- a/tests/omp-path-contract.test.ts
+++ b/tests/omp-path-contract.test.ts
@@ -1,0 +1,40 @@
+import { describe, expect, test } from "bun:test";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { ompModelsConfigPath } from "../src/clients/config-export";
+
+function withTempHome(run: (home: string) => void): void {
+  const home = mkdtempSync(join(tmpdir(), "opencodex-omp-home-"));
+  try {
+    run(home);
+  } finally {
+    rmSync(home, { recursive: true, force: true });
+  }
+}
+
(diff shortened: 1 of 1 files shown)


Wibias:
Review-thread disposition after rebasing this PR onto current `dev`:

- The current PR is now intentionally one file only: `tests/omp-path-contract.test.ts`.
- CodeRabbit's current test-isolation finding was valid and is fixed in `53083223`; CodeRabbit has auto-resolved that thread.
- The `omp-models.yaml` filename finding is a duplicate of the finding already withdrawn on #1109: `filename` is the qualified standalone download artifact, while `destination` is the managed client path. No production change is needed here.
- The older docs/i18n/management-test comments refer to files that were part of the pre-reset inherited #1109 tree and are no longer in this PR. Some are reasonable mainline hardening observations, but they are not findings against this current one-file follow-up and should not remain as merge blockers here.

Fresh CI is running on current head `53083223`; this PR will merge only if that head is fully green.

PR #1324, opened 2026-08-09
fix(codex): recover incomplete account deletion cleanup

chrisae9:
## Summary

This is the focused recovery follow-up to the non-blocking account-deletion note in [Wibias's review of #1303](https://github.com/lidge-jun/opencodex/pull/1303#pullrequestreview-4889721221). [#1319](https://github.com/lidge-jun/opencodex/pull/1319) made account removal durable before destructive cleanup; this PR makes the remaining post-commit cleanup failure explicit and safely recoverable.

- Return a successful deletion with `accountCleanupPending: true` when the account row is gone but local credential cleanup did not finish.
- Add a cleanup-only retry through `DELETE ...?cleanupOnly=1`, `ocx account remove ... --cleanup-only`, and the dashboard.
- Require authoritative persisted configuration and live state to agree that the account remains absent before cleanup-only can touch credentials. Ambiguous or re-added accounts fail closed with HTTP 409.
- Invalidate runtime ownership before reporting pending cleanup and converge picker/catalog state independently.
- Acquire the commit-required SQLite lock before external config or credential writes, preventing a rollback-journal reader from blocking `COMMIT` after those writes complete.
- Document the recovery response and workflow across the supported locales.

### Dashboard recovery state

Dashboard showing the pending account-cleanup notice and Retry cleanup action

The screenshot uses synthetic `.invalid` account data. The opaque account ID needed for retry remains component-owned and is never displayed in the recovery notice.

## Verification

- `bun run prepush` — 10,153 pass / 10 skip / 0 fail; typecheck, changed-dashboard lint, privacy scan, and React Doctor passed.
- `bun test tests/codex-account-delete-atomicity.test.ts tests/codex-auth-api.test.ts tests/cli-account.test.ts` — passed.
- `cd gui && bun test tests/codex-account-pool-behaviour.test.tsx tests/codex-account-pool-toast-tone.test.tsx tests/codex-account-pool-pinned-badge.test.tsx` — passed.
- Documentation site build — passed.
- Root and dashboard dependency audits — passed.
- `git diff --check upstream/dev...HEAD` — clean.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - Added `--cleanup-only` to retry incomplete Codex account credential cleanup.
  - CLI, API, and dashboard now show pending cleanup status and provide retry actions.
- **Bug Fixes**
  - Account deletion and retries are idempotent and block unsafe cleanup when account presence is uncertain.
  - Improved handling of missing and orphaned account credentials.
- **Documentation**
  - Updated CLI and management API guidance across supported languages.
- **Localization**
  - Added cleanup status and retry messages in multiple languages.


chrisae9:
@coderabbitai review

chrisae9:
@coderabbitai review

chrisae9:
@coderabbitai review

src/codex/account-lifecycle.ts
```diff
@@ -97,7 +104,15 @@ function restoreRuntimeConfig(target: OcxConfig, snapshot: OcxConfig): void {
   Object.assign(target, snapshot);
 }
 
-function restorePersistedConfig(configPath: string, previousBytes: string): void {
+function restorePersistedConfig(configPath: string, previousBytes: string | undefined): void {
+  if (previousBytes === undefined) {
+    try {
+      unlinkSync(configPath);
```
Wibias, line 110: 🔴 **Blocking — preserve an existing dangling config symlink during rollback.** `existsSync(configPath)` is false for a dangling symlink, so this path records `previousPersistedConfig === undefined`. The config writer deliberately refuses an unresolvable symlink (`resolveWriteTarget`), but when that save throws this rollback executes `unlinkSync(configPath)` and deletes the symlink itself. That turns a safely refused account deletion into destructive mutation of the user's dotfiles-managed config entry. Track whether there was truly no directory entry (for example with `lstat`) and only remove a file that this operation actually created. Add a regression test with a dangling `config.json` symlink.

gui/src/components/CodexAccountPool.tsx
```diff
@@ -66,6 +66,8 @@ export default function CodexAccountPool({ apiBase, accountModeState = null, ban
   const [reauthId, setReauthId] = useState<string | null>(null);
   const [actionFeedback, setActionFeedback] = useState<string | null>(null);
   const [actionFeedbackTone, setActionFeedbackTone] = useState<NoticeTone | null>(null);
+  const [cleanupRetryId, setCleanupRetryId] = useState<string | null>(null);
```
Wibias, line 69: 🔴 **Blocking — the dashboard recovery state is not durable.** `cleanupRetryId` exists only in component-local React state. After the DELETE succeeds, the account row is gone; if this component unmounts, the page reloads, or the app restarts before cleanup succeeds, this ID and the Retry cleanup action disappear. The UI also intentionally does not expose the opaque ID, so the dashboard has no way to reconstruct the retry. This can leave the credential orphaned indefinitely. Make pending cleanup rediscoverable/durable (or provide another server-backed recovery handle) and add a remount/reload recovery test.

src/cli/account-extended.ts
```diff
@@ -259,14 +289,18 @@ export async function cmdRemove(args: string[], deps: AccountDeps): Promise<numb
...
     removedActive,
     promotedActiveId: after.activeId,
-    ...(classified.type === "codex" ? { catalogRefreshPending } : {}),
+    ...(classified.type === "codex" ? { catalogRefreshPending, accountCleanupPending } : {}),
   };
   if (wantsJson) console.log(JSON.stringify(result, null, 2));
   else if (classified.type === "codex" && removedActive && after.activeId === null) console.log(`openai: ${AUTO_NOTE}`);
   else if (classified.type === "oauth") console.log(after.rows.length ? `${name}: active account is now ${after.activeId}` : `${name}: no accounts remaining`);
   else if (classified.type === "api-key") console.log(after.rows.length ? `${name}: active key is now ${after.activeId}` : `${name}: no keys remaining`);
+  else if (cleanupOnly) console.log(accountCleanupPending
```
Wibias, line 298: 🟠 **Blocking automation contract — `--cleanup-only` still succeeds when cleanup did not complete.** This branch can print `account cleanup is still pending`, but `cmdRemove` still returns 0 below. Exit 0 is defensible for the initial delete because the account row was durably removed; for an explicit `--cleanup-only` retry, cleanup is the requested operation. Returning success when it remains pending makes shell automation silently accept a failed retry unless it parses human stderr/JSON. Return a non-zero result when `cleanupOnly && accountCleanupPending` (with stable machine-readable state preserved) and update the test that currently codifies exit 0.

Wibias (changes requested):
## Verdict: REQUEST CHANGES

The recovery design is directionally strong, and the earlier CodeRabbit findings are addressed. I still found three merge blockers on the current head:

1. **Data integrity:** rollback can delete a dangling/unresolvable `config.json` symlink after the guarded config writer correctly refuses to replace it.
2. **Recovery durability:** the dashboard's only cleanup retry identifier lives in component-local React state, so reload/unmount/restart loses the recovery action after the account row is gone.
3. **Automation contract:** `--cleanup-only` can report that cleanup is still pending but still exit 0, so scripts see success for the one operation they explicitly asked to complete.

Security/privacy review is otherwise good: cleanup errors are projected to fixed public state instead of leaking storage/token details; stale cleanup retries fail closed with 409 when absence cannot be proven; live runtime ownership is invalidated after the durable row deletion; and acquiring the SQLite commit-required lock before external writes closes the reader/COMMIT window.

Review/process state also is not merge-ready yet: I found no submitted human maintainer review on this PR; the prior CodeRabbit threads are resolved, but its latest review is paused/pending. The current-head React Doctor and Cross-platform CI workflow runs are `action_required` with no jobs, so hosted CI has not actually validated this head. The local test results in the PR description are useful evidence, but they do not replace the blocked hosted runs.

**Do not merge until the three blockers below are fixed and hosted CI actually runs on the resulting head.**

lidge-jun:
The incomplete-cleanup bug is real and the recovery direction is valuable, but the current diff still carries a destructive rollback edge. In `src/codex/account-lifecycle.ts:107-114`, `restorePersistedConfig()` unlinks `configPath` whenever no prior bytes were captured. Because `existsSync` at lines 185-187 reports a dangling symlink as absent, the write-then-rollback sequence can replace and then delete a symlink the user owns.

The retry contract also does not survive its most important boundary. `gui/src/components/CodexAccountPool.tsx:69-70` holds the orphan account ID only in component-local state and `retryCleanup()` depends on it at lines 231-236, so a reload loses the only handle to the pending cleanup. In the CLI, `src/cli/account-extended.ts:298-304` prints that cleanup is still pending and returns `0`, which means automation reads a partial failure as success. The PR additionally moves every config mutation from `BEGIN IMMEDIATE` to `BEGIN EXCLUSIVE` in `src/config.ts:2247` — a global locking-policy change rather than account-cleanup recovery.

We would merge a focused recovery PR that preserves pre-existing filesystem entries during rollback, makes the pending cleanup target recoverable from server-owned state after a dashboard reload, and returns a nonzero CLI status while cleanup is incomplete — with targeted rollback, reload-recovery, and exit-code tests, and the shared config transaction mode left alone. Any lock-mode change deserves its own PR with its own justification.

Closing this version because the genuine deletion-recovery fix is bundled with unresolved data-safety, persistence, CLI-contract, and global-locking concerns. The fail-closed design instinct here is right; it is the blast radius that blocks it.

PR #1326, opened 2026-08-09
feat(vision): filter sidecar describers by real image-input capability

lidge-jun:
## Summary

The Vision sidecar model picker offered every model whose provider was `openai` or `anthropic`. That is simultaneously too wide — it lists models that cannot see — and too narrow, because it asks about a provider name instead of a capability. Users have been asking for more models; this is the foundation that lets the answer be "any model that can actually accept an image".

This layer adds `src/vision/eligibility.ts` and nothing else. It has no caller yet; the management API and the dashboard consume it in the PRs above.

Two rules make the predicate non-obvious, and both are load-bearing:

- **`provider.noVisionModels` inverts the modality signal.** That list marks models the proxy describes images *for*, and `applyProviderConfigHints` deliberately **adds** `"image"` to their advertised `inputModalities` so the Codex app does not block the attachment client-side before the sidecar can run. A blind model therefore advertises image input. Membership in that list is a hard disqualifier, checked **before** the modality list it rewrote.
- **Unknown is not zero.** Catalog rows routinely omit `inputModalities` entirely — the live `/api/models` response carries none for either `openai` or `anthropic` rows. So `modelAcceptsImageInput` is a tri-state, and `undefined` stays eligible. Collapsing it to `false` would have emptied the picker completely.

`visionEligibleModelOptions` is the **suggestion list** and is deliberately not the same set as the write gate that lands in the next PR: absence from it must never imply rejection. `gpt-5.6-luna` and `claude-haiku-4-5` are guaranteed per enabled backend so a cold or failed catalog still offers a usable describer.

## Verification

- `bun test tests/vision-eligibility.test.ts` — 11 pass, 0 fail
- Red ablations proving the tests are not vacuous, each reverted afterwards:
  - removing the `isVisionSidecarConsumer` early return → 1 fail (the inversion guard is genuinely exercised)
  - replacing the metadata fallback with `undefined` → 1 fail (the fallback branch is genuinely exercised)

## Checklist

---

**Stack** (merge bottom-up):

| 3 | dashboard vision card | dashboard | layout, non-localized effort values |
| 2 | management API | server | the 400 gate and the allowed-list payload |
| 1 | vision eligibility core ← you are here | predicate | the inversion case and the tri-state |

Review this PR's diff only.

## Summary by CodeRabbit

- **New Features**
  - Added vision sidecar model eligibility filtering for OpenAI and Anthropic providers.
  - Model options now prioritize baseline models, exclude models already used by sidecars, and remove duplicates.
  - Added image-capability detection using available provider and model metadata.
  - Unknown model capabilities remain selectable when eligibility cannot be confirmed.
  - Added backend-aware routing for eligible vision models.

- **Documentation**
  - Added a phased roadmap for management APIs, dashboard integration, validation, fallback behavior, and rollout verification.

- **Tests**
  - Added coverage for modality detection, overrides, metadata fallbacks, baselines, deduplication, and backend routing.

### DIFF
--- a/devlog/_plan/260809_vision_sidecar_model_filter/000_plan.md
+++ b/devlog/_plan/260809_vision_sidecar_model_filter/000_plan.md
@@ -1,0 +1,154 @@
+# 260809 — vision sidecar model eligibility filter
+
+Base: `origin/dev@632743269`, worktree `/Users/jun/.codex/worktrees/34b7/opencodex`.
+Cycle: docs-first. This unit writes the roadmap; no production code lands in this
+work-phase.
+
+## Objective
+
+The dashboard's Vision sidecar model picker currently offers **every** model whose
+provider is `openai` or `anthropic`, with no regard for whether that model can
+actually accept an image. Users have been asking for more models, and the honest
+answer has two halves:
+
+1. The picker is simultaneously **too wide** (it lists models that cannot see) and
+   **too narrow** (it hard-codes two provider names instead of asking about
+   capability).
+2. The sidecar has exactly two executors — the OpenAI Responses forward path
+   (`src/vision/describe.ts`) and the Anthropic Messages path
+   (`src/vision/anthropic-describe.ts`). A model that is not reachable by one of
+   those two wire protocols cannot be a vision sidecar today no matter what the
+   picker shows.
+
+So the deliverable is a real **vision-capability filter** on both sides, plus a
+guaranteed baseline entry per side, plus a dashboard card that shows the allowed
+list in the compact delegation-panel form factor.
+
+## Constraints
+
+- The proxy is Bun-native TypeScript; `bun run typecheck` and `bun run test` gate
+  every layer.
+- `src/vision/reasoning.ts` already owns effort normalization. The filter must not
+  duplicate or contradict `normalizeVisionReasoningForModel`.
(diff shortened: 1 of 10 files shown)


Ingwannu (changes requested):
Requesting changes because the current eligibility core can still publish describers that the runtime cannot safely execute.\n\nThe blocking cases are:\n\n1. Resolve the effective provider through the same registry-enrichment path used by routing before applying noVisionModels. A stale persisted provider can otherwise miss a newer registry exclusion and admit a known-blind model.\n2. Keep provider identity attached to every candidate. Treating every adapter: anthropic row as an Anthropic sidecar option lets models from Umans/Xiaomi/custom gateways get dispatched through the unrelated OAuth provider selected by the runtime.\n3. Use native capability metadata only for a genuinely native/canonical candidate. A custom row whose bare ID collides with a native slug must still honor its own text-only modalities.\n4. Baseline insertion must pass the same eligibility check, and no enabled vision executor must produce an authoritative empty set rather than selectable legacy baselines.\n\nPlease add regressions for registry-enriched no-vision classification, multiple Anthropic-compatible providers, a non-native native-slug collision, a rejected baseline, and no enabled executor. The direction is useful, but this head should not merge until these provider-identity boundaries are fixed.

PR #1327, opened 2026-08-09
feat(management): publish vision describers and refuse the blind ones

lidge-jun:
## Summary

Builds on #1326. `GET` and `PUT /api/sidecar-settings` now carry a `visionModels` list, and both routes that can set a vision describer refuse a model that can be proven blind.

Without this layer the picker filter is cosmetic: anything holding the admin token could still write a text-only model into `visionSidecar.model`, and the sidecar would spend real tokens producing nothing.

Three decisions worth review attention:

- **The gate rejects only a PROVEN-blind model.** An id no source knows stays allowed, because the runtime never required catalog membership and an operator may be ahead of our tables. `tests/vision-reasoning-contract.test.ts` already pins `custom-vision` at HTTP 200 and this change keeps it there. The suggestion list and the write gate are deliberately different sets.
- **The caller's `backend` is a hint, never the authority.** Trusting it opened a laundering path: `o3-mini` is text-only in the OpenAI metadata table but *absent* from the Anthropic one, so a client could claim `backend: "anthropic"` and have a known-blind model read as merely unknown. Measured before the fix: `{model: "o3-mini"}` → 400, `{model: "o3-mini", backend: "openai"}` → 400, `{model: "o3-mini", backend: "anthropic"}` → **200 and persisted**. Both families are now consulted and any positive text-only verdict wins, which is unambiguous because the two vendor tables share zero bare model ids.
- **One policy module serves both routes.** A gate on `/api/sidecar-settings` with a private copy inside the Claude Code override is the same as no gate once the two drift, so `vision-sidecar-options.ts` owns the rule and both callers import it. The Claude Code dashboard control stays freeform by design — it is a text input with suggestions, not a constrained picker.

The catalog is read once per request and reused for the rejection body, so a 400 does not cost two provider fetches.

## Verification

- `bun test tests/sidecar-settings-vision-filter.test.ts` — 11 pass, 0 fail
- Activation evidence for the gate:
  - red ablation forcing the branch to `if (false)` → 1 fail, so the 400 path is genuinely covered
  - live three-variant probe through `handleManagementAPI`, before vs after via `git stash`: 200/200/200 with the model persisted → 400/400/400 with nothing persisted
  - `saveConfigPreservingClaudeCode` spy shows **0 calls** on the reject path, so rejection precedes persistence
  - false-positive guard: `webSearchSidecar` still accepts `o3-mini` with 200, so the gate did not leak into the wrong field
- `docs-site/src/content/docs/guides/sidecars.md` updated (SOT sync); translated locales deliberately left alone.

## Checklist

---

**Stack** (merge bottom-up):

| 3 | dashboard vision card | dashboard | layout, non-localized effort values |
| 2 | management API ← you are here | server | the 400 gate and the allowed-list payload |
| 1 | #1326 | predicate | the inversion case and the tri-state |

Depends on #1326. Review this PR's diff only.

### DIFF
--- a/docs-site/src/content/docs/guides/sidecars.md
+++ b/docs-site/src/content/docs/guides/sidecars.md
@@ -100,6 +100,11 @@ model field.
   keys (Anthropic keys omit it, since that field is ignored there); mutable `https:` images are not
   cached.
 
+The management API and Dashboard picker now list models that can actually accept image input.
+When the matching backend is available, `gpt-5.6-luna` (OpenAI) and `claude-haiku-4-5` (Anthropic)
+are always offered as baseline options. `PUT /api/sidecar-settings` rejects a model known to be
+text-only, but still accepts an unknown id so custom or ahead-of-catalog names keep working.
+
 ```json
 {
   "visionSidecar": {
--- a/src/server/management/agent-settings-routes.ts
+++ b/src/server/management/agent-settings-routes.ts
@@ -51,6 +51,11 @@ import {
   type DebugFlag,
 } from "../../lib/debug-settings";
 import type { OcxClaudeCodeConfig, OcxConfig, OcxCustomModel, OcxProviderConfig } from "../../types";
+import {
+  visionCandidateRows,
+  visionDescriberIsProvablyBlind,
+  visionDescriberRejection,
+} from "./vision-sidecar-options";
 import { drainAndShutdown } from "../lifecycle";
 import { filterRequestLogs, getRequestLogEntries, type RequestLogEntry } from "../request-log";
 import { estimateComboCost, estimateRequestCost, normalizeCostTokens, tokensPerSecond } from "../../usage/cost";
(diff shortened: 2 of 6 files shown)


Ingwannu (changes requested):
Requesting changes because the management API still validates and publishes options against catalog-wide bare IDs instead of the backend/provider that will actually execute the sidecar.\n\nA colliding row can make the validator return early before authoritative OpenAI/Anthropic metadata is checked. Likewise, models from any Anthropic-compatible provider are suggested even though runtime dispatch uses only the provider selected by findAnthropicVisionProvider. Finally, the grandfathered default is currently gpt-5.4-mini even when the selected backend is Anthropic, whose runtime default is claude-sonnet-5.\n\nPlease carry the exact execution provider/backend into option construction and validation, evaluate all relevant authoritative matches rather than returning on an unrelated bare-ID row, and resolve the same backend-specific default as planVisionSidecar. Add focused tests for bare-ID collisions, multiple Anthropic-compatible providers, and an Anthropic backend with no explicit model. This stacked head is not merge-ready until those cases are green.

PR #1328, opened 2026-08-09
feat(gui): show only usable vision describers in the delegation card shape

lidge-jun:
## Summary

Builds on #1327. Two inline review comments on the running dashboard drove this: make the Vision sidecar card match the subagent delegation panel above it, and follow that panel's pattern for the reasoning control — raw wire values, compact.

The picker now consumes the server's eligible-describer list from #1327 instead of filtering by provider name, so it stops offering models that cannot see. The field is optional: an older server omits it and the client falls back to the previous list rather than rendering an empty picker.

Three things a reviewer should look at:

- **`saveSidecar` rebuilds `SidecarData` field by field at three separate sites** (optimistic update, success writeback, session cache), so a new field is silently dropped unless all three are updated. Missing the success writeback alone would have reverted the picker to the legacy list after every save — on exactly the interaction the user performs.
- **The layout is the panel itself, not an inner row.** `.dash-sidecar-card` sets `flex-direction: column` and `.dash-delegation-summary` never resets that property, so combining the two classes would have produced a column regardless of import order.
- **The control group holds its intrinsic width and the copy block is the one that yields.** Letting the controls shrink does not wrap them — `.custom-select` keeps its own min-width, so the children overflow their shrunken container and slide left underneath the hint text. Below 22rem the card stacks instead of overflowing.
- **The web search card beside it now wears the same shell** (`51bfc78a5`), so the pair stops reading as two different components in one grid. No behavior change there — it keeps its own model list and its single control. Two consequences worth noting: the compact effort rule matches `:nth-child(2)` rather than `:last-child`, because the web search card's single select is simultaneously first and last child and would otherwise be shrunk to the effort width; and `.dash-sidecar-card` / `.dash-sidecar-card__row` lose their last user, so those rules are removed rather than left to rot. The grid also moves from `align-items: start` to `stretch` so a hint wrapping to a different line count cannot leave one card visibly short.

The effort select renders the wire value (`low`…`max`) rather than a localized label. The card's own prose stays translated; only the enum values are raw. `visionReasoningLabel` is kept because `gui/tests/vision-reasoning-contract.test.ts` still covers it.

## Verification

- `bun run lint:gui` — clean
- `bun run build:gui` — exit 0
- `bun test ./gui/tests` — 699 pass, 0 fail
- **Render grounding** against a real server started from this worktree (isolated `OPENCODEX_HOME`, headless Chrome over CDP, cache-busting reloads):
  - live `GET /api/sidecar-settings` returned 7 eligible models, `gpt-5.6-luna` first and flagged baseline; the blind rows are absent
  - opening the dropdown rendered exactly those 7 options
  - Korean locale: the card title and hint stay Korean while the effort select reads `low`, not `낮음`
  - measured at 1440 / 1280 / 768 / 420 / 330 px: both selects on one row, positive hint-to-control gap at every row width, zero right overflow, clean column stack at 330
  - after the web search card joined the shell, re-measured at the same widths: both cards report `flex-direction: row` with equal heights while the grid is two-up, no control overflow, positive gaps throughout
- `bun test ./gui/tests` after the shell change — 703 pass, 0 fail

Vision sidecar card in the delegation form factor, Korean locale

Filtered model dropdown showing only image-capable models

Both sidecar cards sharing the shell, equal height while the grid is two-up:

Web search and vision sidecar cards in the matching delegation shell

## Checklist

---

**Stack** (merge bottom-up):

| 3 | dashboard vision card ← you are here | GUI | layout, non-localized effort values |
| 2 | #1327 | server | the 400 gate and the allowed-list payload |
| 1 | #1326 | predicate | the inversion case and the tri-state |

Depends on #1327. Review this PR's diff only.

## Summary by CodeRabbit

- **New Features**
  - Added vision-capable model filtering for sidecar configuration.
  - Added provider-aware model suggestions with baseline options and support for configured legacy models.
  - Added validation that prevents selecting known text-only models while preserving unknown model IDs.
  - Updated the dashboard vision sidecar with model-specific options, backend selection, and reasoning controls.
  - Added responsive sidecar card layouts and improved handling of long labels.

- **Bug Fixes**
  - Preserved saved model and backend settings when available options are incomplete.

- **Tests**
  - Added coverage for filtering, fallback behavior, deduplication, backend resolution, and persisted selections.


Ingwannu (changes requested):
Requesting changes for a functional persistence bug. The server option already carries its backend, but the dashboard drops that identity and later re-infers the backend from /api/models. For an Anthropic-compatible provider not literally named anthropic, or for an Anthropic baseline absent from the catalog, selecting the option can persist backend: openai and route a Claude model through the wrong describer.\n\nPreserve the backend with each vision option and submit that exact backend when the user selects it. Add a regression that selects an Anthropic option whose provider name is custom and verifies the saved backend remains anthropic. I am not making the final UI/UX approval decision; @lidge-jun should do that after the functional stack is corrected.

PR #1331, opened 2026-08-09
fix(catalog): keep failover combos when live discovery omits members

### LINKED ISSUE #1308: [Bug]: ocx sync drops configured combo members and omits failover catalog entry

Wibias:
## Summary
- When authoritative provider `/models` omits configured ids that a combo still targets, retain those rows instead of dropping them from the live catalog.
- Prevents `ocx sync` from omitting failover combos as "member capabilities are incomplete" (OCX-111 / #1308), including the openrouter + opencode-go + command-code repro.
- Builds on the #1305 synthesis fallback; this keeps combo-target configured models visible and selectable too.

Closes #1308

## Verification
- `bun test ./tests/codex-catalog.test.ts -t "OCX-111|successful live discovery|managed Kimi|combo catalog"`
- `bun run typecheck`

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Configured combo target models are now retained when live provider discovery returns different model lists.
  * Combo-only target models remain available even when providers do not list them independently.
  * Failover combo models preserve their expected context window and no longer generate unnecessary omission warnings.
  * Model catalogs remain consistent across live, cached, stale, and failed discovery results.
  * Concurrent discovery requests with different combo configurations now return the correct model availability.


Ingwannu (dismissed):
The direction is valuable and much closer to the right fix than #1310, but I am requesting changes because the retention policy is not yet part of every catalog reuse boundary.\n\nTwo blockers remain on the current head:\n\n1. captureGatherFlight hashes the provider row into providerGraphIdentity but omits retainConfiguredModelIds. Two concurrent gathers with the same provider config and different combo targets can therefore share a flight built for the wrong retention set. Add a deterministic sorted retention list to that identity and test that different sets do not coalesce.\n2. The retention rule runs only after a successful live /models response. Fresh cache, stale cache, cooldown, and discovery-failure returns happen earlier and can return a catalog captured before the combo target existed. Apply one shared configured-target retention helper to every cached/fallback return, or make the relevant cache identity/invalidation include the retention set. Add a warm-cache regression that gathers once without the combo and again with it inside the TTL.\n\nI do not require exposing combo-only ids as standalone provider rows: the existing #1305 synthesis path intentionally keeps those scoped to combo derivation, and the new test correctly asserts no standalone leak. The configured models[] members do need consistent retention across live, cached, stale, and in-flight paths.\n\nOnce these two cache/flight boundaries are fixed and exact-head CI is green, this should remain a strong merge candidate for #1308.

Ingwannu (changes requested):
The direction is valuable and much closer to the right fix than #1310, but I am requesting changes because the retention policy is not yet part of every catalog reuse boundary.

Two blockers remain on the current head:

1. `captureGatherFlight` hashes the provider row into `providerGraphIdentity` but omits `retainConfiguredModelIds`. Two concurrent gathers with the same provider config and different combo targets can therefore share a flight built for the wrong retention set. Add a deterministic sorted retention list to that identity and test that different sets do not coalesce.

2. The retention rule runs only after a successful live `/models` response. Fresh cache, stale cache, cooldown, and discovery-failure returns happen earlier and can return a catalog captured before the combo target existed. Apply one shared configured-target retention helper to every cached/fallback return, or make the relevant cache identity/invalidation include the retention set. Add a warm-cache regression that gathers once without the combo and again with it inside the TTL.

I do not require exposing combo-only IDs as standalone provider rows: the existing #1305 synthesis path intentionally keeps those scoped to combo derivation, and the new test correctly asserts no standalone leak. The configured `models[]` members do need consistent retention across live, cached, stale, and in-flight paths.

Once these two cache/flight boundaries are fixed and exact-head CI is green, this should remain a strong merge candidate for #1308.

Ingwannu (approved):
Re-reviewed exact head `516850518acc`. Both requested catalog reuse boundaries are now covered:

- the sorted combo-retention set participates in `providerGraphIdentity`, with a concurrent-flight regression;
- one shared retention helper is applied across live, fresh-cache, stale/cooldown, and discovery-fallback returns, while cache writes deliberately avoid freezing combo-only state.

Local exact-head validation passed: TypeScript typecheck and 173 focused catalog/gather tests. GitHub CI, macOS, all Linux shards, packaging smoke checks, hygiene, and CodeRabbit are green. Approved for `dev`.

Issue #1325, opened 2026-08-09
[Bug] Nested field shapes bypass the #1219 frame guard and crash both adapter paths


snowyukitty:
### Client or integration

Not client-specific — reproduced by driving the adapters directly with a synthetic `Response`.

### Area

Adapters · Streaming / SSE parsing · Non-streaming response parsing

### Summary

lidge-jun/opencodex#1219 / lidge-jun/opencodex#1240 fixed the frame **root**: `JSON.parse(payload)` was cast to `Record<string, unknown>` and dereferenced, so `data: null` crashed the parser. The same "cast, then dereference" pattern survives **one level deeper**, on fields inside an otherwise well-formed frame. A frame that is a valid JSON object still crashes the adapter when a nested field has an unexpected shape.

These are found by audit, not from a live capture. Unlike `data: null` — which [agentrouter.org](<http://agentrouter.org>) demonstrably emits as padding — I have **no evidence any provider sends these**. Filing because the defect class and blast radius are identical to lidge-jun/opencodex#1219 (an escaping `TypeError` mid-stream, which the surrounding `try/catch` cannot classify), and because the guards are cheap.

Verified on `dev` at `79831c90`, with the lidge-jun/opencodex#1240 fix already present.

| # | File | Line | Trigger | Thrown |
| -- | -- | -- | -- | -- |
| A | `src/adapters/google.ts` | `:583` | `{"candidates":[null]}` | `null is not an object (evaluating 'candidates[0].finishReason')` |
| B | `src/adapters/openai-chat.ts` | `:1040` | `delta.tool_calls` not an array | `{} is not iterable` |
| C | `src/adapters/openai-chat.ts` | `:1041` | `delta.tool_calls: [null]` | `null is not an object (evaluating 'tc.index')` |
| D | `src/adapters/openai-chat.ts` | `:1226` | non-stream `tool_calls` not an array | `{} is not iterable` |
| E | `src/adapters/openai-chat.ts` | `:1226` | non-stream `tool_calls: [null]` | `null is not an object (evaluating 'tc.id')` |
| F | `src/adapters/openai-chat.ts` | `:1227` | non-stream element without `function` | `undefined is not an object (evaluating 'tc.function.name')` |

**D–F are in** `parseResponse`**, not** `parseStream` — the non-streaming path has the same defect, so a fix confined to the streaming parsers would be incomplete.

[...]

PR #1332, opened 2026-08-09
fix(adapters): reject malformed nested response shapes

Ingwannu:
## Summary

* Reject malformed nested Google candidate and OpenAI Chat tool-call shapes through the adapter error channel.
* Keep root `data: null` padding behavior unchanged while failing closed once an upstream claims a candidate or tool call.
* Preserve non-stream usage accounting and prevent raw JavaScript `TypeError` exceptions from escaping the adapter.

## Verification

* `taskset -c 0-1 bun run typecheck` — passed on the rebased exact head.
* `taskset -c 0-1 bun test tests/openai-chat-hardening.test.ts tests/google-hardening.test.ts tests/openai-chat-parallel-stream.test.ts tests/google-vertex-stream.test.ts` — 83 passed, 0 failed on the rebased exact head.
* Full pre-rebase suite — 10,130 passed, 10 skipped, 1 unrelated existing local-environment failure in `tests/codex-shim.test.ts` (`Unix shim exports persisted service API token before running Codex`); that failure reproduces in isolation and does not touch adapter code.
* `git diff --check origin/dev...HEAD` — passed.

## Decision Log

* 목적과 의도: Prevent malformed nested upstream response structures from silently losing tool calls or crashing the request with an implementation-level exception.
* 기존 구현 및 제약 조건: Root null SSE padding is intentionally tolerated, but nested `candidates: [null]`, object-shaped `tool_calls`, null tool calls, and missing tool functions were trusted through casts.
* 검토한 주요 대안: Ignore malformed entries, coerce them into empty values, or terminate the adapter turn with a structured error.
* 선택한 방식: Preserve benign root padding and fail closed only after an upstream claims a malformed candidate or tool-call structure.
* 다른 대안 대신 이 방식을 선택한 이유: Ignoring or coercing a claimed tool call can orphan its matching result and allow a false successful completion; a structured adapter error is deterministic and observable.
* 장점, 단점 및 영향: Valid provider responses are unchanged and malformed turns no longer leak raw `TypeError`; malformed providers now receive a terminal error instead of best-effort continuation.

## Checklist

- [X] Docs or release notes were updated when needed. No public configuration or API contract changed, so no docs update is needed.
- [X] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. This patch changes response validation only and does not touch credentials or routing authority.

Closes lidge-jun/opencodex#1325

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed Google response candidates by returning clear terminal errors instead of silently skipping invalid data.
  * Added validation for malformed tool-call payloads in OpenAI Chat responses, preventing runtime errors and reporting adapter errors consistently.
  * Preserved usage-token metadata when invalid tool-call data is encountered.
* **Tests**
  * Added coverage for malformed Google candidates and OpenAI tool-call payloads in streaming and non-streaming responses.

### DIFF
--- a/src/adapters/google.ts
+++ b/src/adapters/google.ts
@@ -577,15 +577,32 @@ export function createGoogleAdapter(provider: OcxProviderConfig): ProviderAdapte
           pendingUsage = usageFromGemini(usageMeta);
           sawTerminalSignal = true;
         }
-        const candidates = root.candidates as { content?: { parts?: unknown[] }; finishReason?: string }[] | undefined;
-        if (!candidates?.length) return "continue";
+        const rawCandidates = root.candidates;
+        if (rawCandidates === undefined) return "continue";
+        if (!Array.isArray(rawCandidates)) {
+          yield { type: "error", message: "google response contained invalid candidates" };
+          return "terminate";
+        }
+        if (rawCandidates.length === 0) return "continue";
+        const rawCandidate = rawCandidates[0];
+        if (rawCandidate === null || typeof rawCandidate !== "object" || Array.isArray(rawCandidate)) {
+          // Unlike a root `data: null` keepalive, this is a claimed response candidate. Treat it
+          // as terminal protocol corruption so the turn cannot complete after silently losing
+          // a candidate or tool call (#1325).
+          yield { type: "error", message: "google response contained invalid candidates" };
+          return "terminate";
+        }
+        const candidate = rawCandidate as {
+          content?: { parts?: unknown[] };
+          finishReason?: string;
+        };
 
-        if (typeof candidates[0].finishReason === "string" && candidates[0].finishReason) {
-          lastFinishReason = candidates[0].finishReason;
+        if (typeof candidate.finishReason === "string" && candidate.finishReason) {
+          lastFinishReason = candidate.finishReason;
           sawTerminalSignal = true;
         }
 
-        const parts = candidates[0].content?.parts as { text?: string; functionCall?: { name: string; args: unknown } }[] | undefined;
+        const parts = candidate.content?.parts as { text?: string; functionCall?: { name: string; args: unknown } }[] | undefined;
         // Record Gemini thought signatures for the next stateless tool-result turn. Vertex and
         // Antigravity use separate model namespaces so opaque provider state cannot cross routes.
         const replayModel = provider.googleMode === "cloud-code-assist" ? antigravityModel : vertexReplayModel;
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -163,6 +163,18 @@ function invalidChoicesEvent(usage?: OcxUsage): Extract<AdapterEvent, { type: "e
   };
 }
 
+function invalidToolCallsEvent(usage?: OcxUsage): Extract<AdapterEvent, { type: "error" }> {
+  return {
+    type: "error",
+    message: "upstream response contained invalid tool calls",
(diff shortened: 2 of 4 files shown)


snowyukitty (commented):
The core boundary is right on `8ef72105`: this treats a malformed nested candidate/tool-call payload as claimed response data and fails closed, while preserving #1240's benign root-padding skip. That is the important distinction here, and the patch gets it right.

I verified the issue's six reported shapes against this exact head:

- Google streaming `candidates: [null]` emits only `google response contained invalid candidates`; the later `STOP` frame is not consumed as a successful completion.
- OpenAI streaming object-shaped `tool_calls` and `[null]` emit only `upstream response contained invalid tool calls`, preserve usage, and suppress the following `[DONE]`.
- OpenAI non-streaming object-shaped `tool_calls`, `[null]`, and a call without `function` return the same structured error with usage instead of leaking a JavaScript exception.

The #1240 control also holds: `tests/sse-null-data-frame.test.ts` still proves that mid-stream root `data: null` is skipped in both Google and OpenAI Chat, the later content/finish signal and `[DONE]` complete the turn, and an all-padding stream still fails closed. Local result was **119 passed / 0 failed** across that file plus the four focused suites named in the PR. Local typecheck remains the known Windows control failure only (`@napi-rs/keyring` missing): exit 2 with output line-for-line identical on this head and current `dev@a9838c1a`. The exact PR head's cross-platform `ci` check is successful.

I also independently reproduced both current CodeRabbit findings rather than taking them on trust:

- [`content.parts` container/element validation](https://github.com/lidge-jun/opencodex/pull/1332#discussion_r3742623745): `parts: {}` and `parts: [null]` still escape as raw `TypeError`s.
- [streaming tool-call field validation](https://github.com/lidge-jun/opencodex/pull/1332#discussion_r3742623749): present `function: []`, numeric `id`, numeric `function.name`, and numeric `function.arguments` are accepted and followed by `done`; the latter three also emit non-string values through the string-typed adapter event contract. A temporary fail-closed probe was **0 passed / 6 failed** across those two Google and four OpenAI shapes.

One qualification to the second automated recommendation: validating fields that are *present* is supported by the reproduction, but requiring every assembled streaming call to have received an ID and name would conflict with the existing, passing compatibility contracts in `tests/openai-chat-parallel-stream.test.ts` (synthesized missing ID, and T7's explicit missing-name parity). I would preserve those absence cases unless changing them is intentional and separately justified.

So, for the six #1325 cases and the #1240 interaction, I found no additional defect; the implementation and focused regressions are sound. I am leaving this as a comment rather than an approval while the two independently verified adjacent findings are unresolved. Please correct me if there is provider evidence that makes the existing missing-ID/name compatibility contract obsolete.

lidge-jun:
Landed on `dev` as 2495de3ad, rebased onto the current head with your authorship preserved.

Re-verified against the moving dev head before merge: `src/adapters/google.ts:580-588` still dereferenced `candidates[0].finishReason` and `.content` after checking only array length, `src/adapters/openai-chat.ts:919-921` still iterated an unchecked `delta.tool_calls`, and `:1076-1079` still dereferenced `tc.function.name`. Distinguishing benign root padding from malformed claimed nested content is the part that makes this safe — valid responses behave exactly as before, and malformed ones now fail deterministically through the existing adapter error channel instead of throwing raw. Full suite green (10243 pass, 0 fail).

Thanks — covering all six issue shapes across both streaming and non-streaming paths made this easy to trust.

Issue #1292, opened 2026-08-08
[Provider compatibility] DeepSeek V4 Flash returns 400 when developer message is interleaved between function_call and function_call_output


c3right:
### Client or integration

Codex App (Windows desktop)

### Area

Tools / Responses continuation / provider compatibility

### Summary

`deepseek/deepseek-v4-flash` on the native `openai-responses` route returns HTTP 400 when Codex sends a continuation whose input history contains a developer `message` between a `function_call` and its matching `function_call_output`.

The tool itself executes successfully, the `function_call_output` is present, and the `call_id` matches exactly. The failure is structural/interleaving-related:

```text
function_call(call_id=X)
→ message(role=developer, content=...)
→ function_call_output(call_id=X, output=successful tool result)
```

Upstream error:

```text
No tool output found for tool call X.
```

This is reproducible and reversible in one workspace. The developer message content is sourced verbatim from a local `task_plan.md` context file. Temporarily renaming that file makes the same DeepSeek tool call succeed; restoring it makes the 400 return immediately.

Important nuance: the exact Codex-side mechanism that injects this developer message is not yet identified. We initially suspected a planning-with-files `PostToolUse` hook, but static inspection ruled that out as the direct source: that hook only emits a short fixed reminder and does not read `task_plan.md`. The failing request nevertheless contains a developer message matching the beginning of `task_plan.md` verbatim.

### Reproduction

Environment A (failing workspace):

1. Open a Codex workspace that contains a non-empty `task_plan.md` and where the current Codex context-injection setup causes part of that file to appear as a developer message in Responses history.
2. Select `deepseek/deepseek-v4-flash` with Max effort.
3. Start a fresh conversation.
4. Trigger exactly one shell tool call, for example:

```powershell
Get-Content -Path "D:\AI\storycard\README.md" -TotalCount 5
```

5. The first DeepSeek Responses request returns HTTP 200 and emits the tool call.
6. Codex executes the command successfully.
7. The follow-up `/v1/responses` request is sent, but the relevant input order is:

```text
function_call(call_id=X)
message(role=developer, text copied from task_plan.md)
function_call_output(call_id=X, output="Exit code: 0 ...")
```

8. DeepSeek returns HTTP 400:

```text
No tool output found for tool call X.
```

A/B/ABA validation on the same workspace:

- `task_plan.md` present: failure reproduced deterministically (5/5 observed failures in prior sessions).
[...]

PR #1334, opened 2026-08-09
fix(deepseek): normalize tool result adjacency

Ingwannu:
## Summary

* Normalize unambiguous DeepSeek Responses tool-call continuations so each result immediately follows its matching call.
* Preserve hook-injected developer content by moving it after the matched result instead of dropping it.
* Gate the behavior behind provider capability metadata so tolerant Responses providers keep their original history order.
* Backfill the DeepSeek capability into stale provider configurations and document the provider-specific decision boundary.

## Verification

* `taskset -c 0-1 bun run typecheck` — passed on the rebased exact head.
* `taskset -c 0-1 bun test tests/deepseek-inbound-wire.test.ts tests/openai-responses-passthrough.test.ts tests/deepseek-reasoning-replay.test.ts tests/deepseek-reasoning-replay-gaps.test.ts tests/config.test.ts tests/provider-registry-parity.test.ts tests/config-save-boundary.test.ts` — 277 passed, 0 failed on the rebased exact head.
* Full pre-rebase suite — 10,166 passed, 10 skipped, 2 unrelated local-environment failures:
  * the existing Codex shim token-file test resolves the live OpenCodex token path in this running proxy environment and also fails in isolation;
  * the npm cache worker test exceeded its 5-second bound during the 828-second CPU-limited suite, then passed 11/11 in 1.1 seconds when rerun alone.
* `git diff --check origin/dev...HEAD` — passed.

## Decision Log

* 목적과 의도: Make valid Codex tool continuations acceptable to DeepSeek when hook context is persisted between a call and its result.
* 기존 구현 및 제약 조건: Existing orphan repair checks pair presence only; globally changing valid history would affect tolerant providers and moving or dropping the developer message is semantically meaningful.
* 검토한 주요 대안: Reorder all Responses providers, drop the injected message, or add a provider capability and perform a lossless one-to-one reorder.
* 선택한 방식: Enable adjacency normalization only for DeepSeek, move only unique matched pairs, and preserve every intervening item after the result.
* 다른 대안 대신 이 방식을 선택한 이유: This fixes the confirmed strict parser without changing tolerant provider history or guessing across duplicate call ids.
* 장점, 단점 및 영향: DeepSeek accepts the continuation and no context is lost; the relative position of intervening context moves after the tool result only on the declared strict provider.

## Checklist

- [X] Docs or release notes were updated when needed. The provider contract, type documentation, and in-code Decision Log are updated; no user command or configuration workflow changes.
- [X] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. No credential destination or authentication path changes.

Closes lidge-jun/opencodex#1292

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with providers that require tool results to immediately follow their corresponding tool calls.
  * Automatically preserves message order when tool-call matches are ambiguous or when providers do not require adjacent results.
  * Enabled this behavior for DeepSeek while preserving intervening developer context.

### DIFF
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -539,6 +539,65 @@ function repairOrphanedInputItems(body: unknown, dropReasoning: boolean): unknow
 }
 
 /**
+ * Make unambiguous Responses tool pairs adjacent for upstream parsers that require it.
+ *
+ * [Decision Log]
+ * - 목적과 의도: Keep Codex hook-injected developer context without letting it make a strict upstream reject the matching tool result.
+ * - 기존 구현 및 제약 조건: The orphan repair verifies only pair presence; globally reordering valid history would change tolerant providers unnecessarily.
+ * - 검토한 주요 대안: Reorder every Responses request, drop the intervening message, or gate a lossless reorder behind provider capability metadata.
+ * - 선택한 방식: Reorder only unique call/result pairs for providers that explicitly require adjacency, preserving every intervening item immediately after the result.
+ * - 다른 대안 대신 이 방식을 선택한 이유: The provider gate limits semantic blast radius, while refusing ambiguous duplicate ids avoids guessing which result belongs to which call.
+ * - 장점, 단점 및 영향: DeepSeek receives the adjacency its parser requires; tolerant providers stay byte/order equivalent. Ambiguous duplicate ids still fail upstream rather than being silently rewritten.
+ */
+function normalizeResponsesToolResultAdjacency(body: unknown): unknown {
+  if (!isPlainObject(body) || !Array.isArray(body.input)) return body;
+  const input = body.input;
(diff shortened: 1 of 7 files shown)


lidge-jun:
Landed on `dev` as 56cca2f3d, rebased onto the current head with your authorship preserved.

Verified before merge: `src/adapters/openai-responses.ts:507-535` only checked `functionCallIds.has(callId)` and preserved original ordering, with no adjacency normalization in the forwarding pipeline at `:1158-1160` — so developer context between a tool call and its result really did break DeepSeek continuations. Gating the normalization on the provider capability and moving only uniquely matched pairs is the right conservative shape: tolerant providers are untouched and intervening content survives. Full suite green (10243 pass, 0 fail).

Thanks — the stale-config backfill and the tolerant-provider control test are what make the capability plumbing convincing rather than just plausible.

Issue #1312, opened 2026-08-08
[Bug] Vertex replay session id uses promptCacheKey, which is shared across conversations


lidge-jun:
### Client or integration

Codex CLI

### Area

Provider adapter

### Summary

`vertexReplaySessionId()` (`src/adapters/google.ts:56`) keys the Vertex thought-signature replay namespace on `parsed.options.promptCacheKey`, under a comment that says it does something else:

```ts
/** Prefer Codex's stable opaque thread key; retain the existing deterministic fallback for clients
 * that omit it. ... */
function vertexReplaySessionId(parsed: OcxParsedRequest): string {
  const promptCacheKey = parsed.options.promptCacheKey?.trim();
  return promptCacheKey || antigravitySessionId(parsed);
}
```

`promptCacheKey` is not a thread key. It is the client-supplied Responses `prompt_cache_key`, accepted as any string (`src/responses/schema.ts`, copied at `src/responses/parser.ts:661`), and this repository elsewhere treats it as explicitly **shared across conversations**:

* `src/adapters/cursor/request-builder.ts:227` warns against using a shared `prompt_cache_key` as a conversation id;
* Claude Desktop cache cohorts are marked as shared in `src/server/responses/core.ts:620` and `src/server/claude-messages.ts:701`.

So two distinct conversations in the same cache cohort share a Vertex replay namespace, and any client can choose the value.

### Relationship to lidge-jun/opencodex#1297

This is the same class of defect lidge-jun/opencodex#1297 reported for the Cloud Code Assist path, on a different anchor. lidge-jun/opencodex#1311 fixed the CCA side by anchoring on `_clientThreadId` and deliberately **not** on `promptCacheKey`; it left this one alone rather than bundling an unrelated path into that fix.

Worth recording how it surfaced: while fixing lidge-jun/opencodex#1297 I found this function and read its comment as the correct pattern to copy. It is not — the comment describes the right anchor and the code uses a different one. Anyone reading `google.ts` for guidance will make the same mistake.

### Severity

Lower than lidge-jun/opencodex#1297, and the two failure modes differ in kind:

* lidge-jun/opencodex#1297 was **instability** — a changing id loses the signature association entirely.
* This is **sharing** — a stable but non-exclusive namespace.

[...]

PR #1335, opened 2026-08-09
fix(vertex): scope replay by client thread

Ingwannu:
## Summary

* Scope Vertex thought-signature replay to Codex's stable client thread identifier instead of the potentially shared prompt-cache key.
* Preserve the deterministic fallback for clients that do not send a thread identifier.
* Add regressions proving shared cache cohorts cannot replay signatures across threads while same-thread continuation still works.

Closes lidge-jun/opencodex#1312

## Verification

* `taskset -c 0-1 bun run typecheck`
* `taskset -c 0-1 bun test tests/antigravity-static-catalog.test.ts tests/google-adapter.test.ts tests/google-antigravity-oauth.test.ts tests/google-antigravity-replay.test.ts tests/google-antigravity-wire.test.ts tests/google-empty-content.test.ts tests/google-hardening.test.ts tests/google-models-listing.test.ts tests/google-tool-schema.test.ts tests/google-vertex-http.test.ts tests/google-vertex-stream.test.ts tests/google-vertex-thought-signature.test.ts tests/google-wire-compiler.test.ts tests/vertex-catalog.test.ts` (227 pass, 0 fail)
* `git diff --check origin/dev...HEAD`

## Checklist

- [X] Docs or release notes were updated when needed. No documentation change is needed because this restores the existing per-thread replay contract.
- [X] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. No credential or raw thread identifier is persisted; the replay store continues to hash the namespace key.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Vertex replay session handling by prioritizing the client thread identifier when available.
  * Prevented replay signatures from being incorrectly shared across separate client threads.
  * Preserved deterministic replay behavior for requests without a client thread identifier.

### DIFF
--- a/src/adapters/google.ts
+++ b/src/adapters/google.ts
@@ -54,8 +54,8 @@ function resolveVertexApiKey(optKey?: string): string | undefined {
 /** Prefer Codex's stable opaque thread key; retain the existing deterministic fallback for clients
  * that omit it. The replay store hashes this value and never retains the raw session identifier. */
 function vertexReplaySessionId(parsed: OcxParsedRequest): string {
-  const promptCacheKey = parsed.options.promptCacheKey?.trim();
-  return promptCacheKey || antigravitySessionId(parsed);
+  const threadId = parsed._clientThreadId?.trim();
+  return threadId || antigravitySessionId(parsed);
 }
 
 /**
--- a/tests/google-vertex-thought-signature.test.ts
+++ b/tests/google-vertex-thought-signature.test.ts
@@ -55,6 +55,16 @@ const continuation = () => request([
   },
 ], false);
 
+function scopedReplayRequest(
+  parsed: OcxParsedRequest,
+  threadId: string | undefined,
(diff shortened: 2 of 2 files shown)


lidge-jun:
Landed on `dev` as e09fe6e2f, rebased onto the current head with your authorship preserved.

Verified before merge: `src/adapters/google.ts:56-58` claimed in its own comment to prefer a thread key but actually returned `parsed.options.promptCacheKey?.trim()` — a cohort that can be shared across conversations. The real per-thread anchor is populated from `x-codex-parent-thread-id` at `src/server/responses/core.ts:1413-1414`, so this two-line change makes the code do what its comment already promised. Full suite green (10243 pass, 0 fail).

Thanks — the test where two different threads share one cache key is the one that actually proves the bug, and keeping `antigravitySessionId` as the threadless fallback avoided a regression for clients that never send the header.

PR #1344, opened 2026-08-09
fix(routing): define hard cost-cap behavior when cost evidence is unknown

### LINKED ISSUE #1181: Routing: define hard cost-cap behavior when runtime cost evidence is unknown
## Area

Proxy and routing

Make `maxEstimatedCostUsd` behave as an explicit policy when a live candidate has no finite cost estimate. Operators should be able to distinguish “known to be below the cap” from “cost is unknown” without silently weakening a hard ceiling or accidentally disabling every live route.

## What prevents this today?

The current evaluator excludes only a finite estimate above the configured limit. The live routing path often has no request-usage evidence yet, so an unknown estimate passes the cap check.

A naive fail-closed patch is also unsafe: because usage is not currently wired into that path, it can reject every live candidate whenever a cap is configured. Existing documentation describes the cap in terms of known dry-run evidence, so changing this unconditionally would also change the contract.

## What should OpenCodex do?

Define and expose one explicit unknown-cost policy. A minimal design could preserve the current `known-only` behavior but return an `unknown_cost` explanation, while an opt-in `fail-closed` mode excludes an unknown candidate. The chosen behavior should be consistent across dry-run, live routing, logs, and documentation.

## Example usage or interface

With a route capped at USD 0.05:

```text
estimate = 0.08  -> excluded: cost_cap_exceeded
estimate = 0.03  -> eligible
estimate = unknown
  known-only     -> eligible, explanation: unknown_cost
  fail-closed    -> excluded: cost_evidence_required
```

The exact configuration name is intentionally left to maintainer design review.

A naive implementation that makes every unknown estimate fail closed would enforce the word “hard” but block normal live routing until reliable request-cost evidence exists. Keeping the current behavior without an explanation leaves operators unable to tell whether the cap was actually evaluated.

## Additional context

The live route currently constructs cost evidence before request usage is available, while the dry-run path can have a finite estimate. No equivalent open issue or PR was found. This proposal asks for one contract that explicitly covers both evidence states before a code change.

## Checks

abhisheksharma2411:
Fixes #1181.

## Summary

`limits.maxEstimatedCostUsd` is documented as a hard per-request ceiling, but it never fires on the live routing path.

`evaluatePolicyProfile` only excludes a candidate when the estimate is a finite number (`src/routing/evaluator.ts`):

```ts
const overCostLimit = costLimit !== undefined
  && typeof estimatedCost === "number"   // an unknown estimate passes silently
  && Number.isFinite(estimatedCost)
  && estimatedCost > costLimit;
```

…while `routeModel` assembles cost evidence **without** usage (`src/router.ts`):

```ts
cost: costEvidenceForCandidate({
  provider: candidate.provider,
  model: candidate.model,
  limitUsd: profile.limits.maxEstimatedCostUsd,   // no `usage`
}),
```

`costEvidenceForCandidate` correctly returns `{ limitUsd, incomplete: true }` with no `estimatedUsd`, so on the live path the estimate is *always* unknown and any candidate passes a cap the operator configured as hard. The existing coverage passes only because it supplies `usage` directly — exercising a path production does not take.

As #1181 notes, failing closed unconditionally is not safe either: with usage unwired it would reject every live candidate whenever a cap is set, and it would silently change the documented dry-run contract.

This adds an explicit, opt-in policy instead:

```jsonc
{
  "limits": {
    "maxEstimatedCostUsd": 0.25,
    "onUnknownCost": "exclude"   // "allow" (default) | "exclude"
  }
}
```

- **`"allow"` (default)** — preserves current behaviour and the documented contract exactly. No existing deployment changes on upgrade.
- **`"exclude"`** — the ceiling becomes genuinely hard: a candidate whose cost cannot be *proven* under the cap is ineligible.

Unknown-cost exclusions emit a distinct **`cost-limit-unknown`** code, so a trace distinguishes *known above the cap* from *cost is unknown* — the operator-facing distinction the issue asks for.

Kept deliberately separate from `unknownEvidence.cost`, which governs how an unknown-cost candidate is **scored** rather than whether the **ceiling** applies. A test asserts the two mechanisms stay distinguishable.

| File | Change |
|---|---|
| `src/types.ts` | `OcxRoutingUnknownCostCapMode`; `onUnknownCost` on `OcxRoutingProfileLimits` |
| `src/routing/profile.ts` | Validation (`"allow"` / `"exclude"`) + normalization |
| `src/routing/evaluator.ts` | Cap policy under unknown evidence; `cost-limit-unknown` exclusion |
| `tests/cost-cap-unknown-evidence.test.ts` | New — 5 cases |

## Verification

The defect was reproduced with a failing test **before** the fix: with a `$0.000001` cap and live-path evidence (no usage), the candidate was still eligible and selected.

Commands run against the `dev` base:

```
bun test tests/cost-cap-unknown-evidence.test.ts     → 5 pass, 0 fail
bun test tests/cost-scoring.test.ts \
         tests/routing-profile.test.ts               → 27 pass, 0 fail
bun x tsc --noEmit                                   → clean
bun scripts/test.ts                                  → 10134 pass, 7 skip, 0 fail
```

New test cases:

1. **Repro** — live-path evidence carries no estimate, so the hard cap never fires (documents current behaviour)
2. **Fail-closed** — `onUnknownCost: "exclude"` excludes the candidate and emits `cost-limit-unknown`
3. **Default unchanged** — with no `onUnknownCost`, behaviour is identical to today
4. **Mechanism distinctness** — `unknownEvidence.cost: "exclude"` emits `unknown-price`, *not* `cost-limit-unknown`
5. **Inert without a cap** — `onUnknownCost` has no effect when `maxEstimatedCostUsd` is unset

Coverage on touched files: `src/routing/profile.ts` 89.8% lines; every added line in `src/routing/evaluator.ts` covered. No GUI changes, so no screenshot applies.

## Checklist

- [x] Docs or release notes were updated when needed. — The new field is documented via JSDoc on `OcxRoutingProfileLimits` and its validation message. Happy to add a docs entry if you point me at the right file.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. — No secrets or auth touched. The default is deliberately `"allow"` so an upgrade cannot silently make every live candidate ineligible; the stricter behaviour is opt-in.

## Notes for review

- Default is unchanged, so this is additive and non-breaking.
- Wiring real usage into the live routing path is deliberately **not** attempted here — that is the larger change #1181 warns about, and it is separable from defining the policy.
- Happy to rename `onUnknownCost`, or fold it into `unknownEvidence`, if you would prefer a single mechanism. I kept them separate because they answer different questions.


Wibias:
Malfunctioned PR a swell as to broad. Please open a new smaller PR.

PR #1362, opened 2026-08-09
fix(codex): pin Darwin process utility path

luvs01:
## Summary

- Pin every macOS Codex app-server process discovery and start-time probe to the system `/bin/ps` binary instead of resolving `ps` through `PATH`.
- Preserve the existing arguments, timeouts, parsing, and fail-closed behavior.
- Add a source guard that requires exactly four pinned Darwin call sites and rejects any bare `execFileSync("ps", ...)` invocation.

This prevents a modified local `PATH` from substituting another executable during Codex app-server synchronization and staleness checks.

## Verification

- `bun test tests/codex-app-server-processes.test.ts` — 31 passed, 0 failed on Bun 1.3.14.
- `bun run typecheck`
- `bun run privacy:scan`
- `git diff --check`
- Independent final diff review found no actionable P0–P3 findings.

## Checklist

- [x] Docs or release notes were not needed for this internal process-launch hardening.
- [ ] Security-sensitive changes require explicit maintainer review before merge.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved process detection on Darwin systems by using a reliable system path when invoking process utilities.

* **Tests**
  * Added regression coverage to ensure process checks consistently use the correct system command path.

### DIFF
--- a/src/codex/app-server-processes.ts
+++ b/src/codex/app-server-processes.ts
@@ -272,12 +272,12 @@ function listDarwinSnapshots(uid: number | undefined): ProcessSnapshot[] {
   // Top-level exec failure propagates: callers decide their own safe default
   // (restart flow → treat as none; staleness check → unknown, never "fresh").
   const output = uid !== undefined
-    ? execFileSync("ps", ["-u", String(uid), "-o", "pid=,command="], {
+    ? execFileSync("/bin/ps", ["-u", String(uid), "-o", "pid=,command="], {
       encoding: "utf-8",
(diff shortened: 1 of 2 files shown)


Wibias:

Wibias:
Thanks @luvs01 — nice focused hardening here. Pinning the Darwin process probes to `/bin/ps` and locking it in with a source guard is exactly the kind of defensive cleanup we want. Merged! 🙏

PR #1363, opened 2026-08-09
fix(claude): honor Desktop native model opt-out

luvs01:
## Summary

- Make the native-integration Claude Desktop enable route use the same Desktop-specific native model visibility helper as the CLI, apply route, model list, and profile renderer.
- Preserve the existing default behavior while honoring an explicit `claudeCode.desktopNativeModels: false` opt-out.
- Add a focused regression that verifies the Desktop writer receives no native slugs after the persisted opt-out is reloaded.

Previously this one route used the generic native model list, so enabling Claude Desktop could re-add native entries that the user had explicitly disabled for Desktop.

## Verification

- `bun test tests/native-claude-desktop-toggle.test.ts` — 10 passed, 0 failed on Bun 1.3.14.
- `bun run typecheck`
- `bun run privacy:scan`
- `git diff --check`
- Independent final diff review found no actionable P0–P3 findings.

## Checklist

- [x] Docs or release notes were not needed because this restores the existing Desktop opt-out contract.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Claude Desktop enablement now respects the setting that disables native models.
  * Only native integrations available for desktop are included during configuration.

* **Tests**
  * Added coverage to verify successful Claude Desktop activation without native model configuration when opted out.

### DIFF
--- a/src/server/management/native-integration-routes.ts
+++ b/src/server/management/native-integration-routes.ts
@@ -18,7 +18,7 @@
  * 011 (Claude Code), 012 (Grok).
  */
 import { loadConfig, readRuntimePort, saveConfigPreservingClaudeCode } from "../../config";
-import { filterCatalogVisibleModels, nativeOpenAiContextWindow, visibleNativeSlugs } from "../../codex/catalog";
+import { desktopVisibleNativeSlugs, filterCatalogVisibleModels, nativeOpenAiContextWindow, visibleNativeSlugs } from "../../codex/catalog";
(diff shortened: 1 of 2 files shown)


Wibias:

Wibias:
Thanks @luvs01 — good catch and a clean fix. Reusing the Desktop-specific visibility helper keeps the native-model opt-out consistent across the integration surfaces, with a focused regression to lock it in. 🙏

PR #1365, opened 2026-08-09
fix(omp): preserve rich native model capabilities

dbc-hbin:
## Summary

- Preserve unrelated OMP YAML comments and formatting by patching only `providers.opencodex`; unsupported or ambiguous YAML now fails closed.
- Keep whole-file fingerprint protection for clients whose writers still re-serialize the complete document.
- Extend OMP with model-level Responses routing for native OpenAI models while preserving image input and reasoning-effort controls.
- Keep routed providers on Chat Completions and return a structured JSON 413 when translation exceeds its safe buffer.

## Verification

- `bun run typecheck`
- `bun run test`
- `bun run privacy:scan`
- `bun run build` in `docs-site/`
- Focused writer regression covers apply, stale refresh, and disable with a foreign provider, comments, quoted formatting, and post-apply source edits.
- `ocx restart` was attempted for the new head and refused safely because the running proxy predates process-bound restart support; no stop/kill fallback was used.

## Checklist

## Review readiness checklist

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - Added OpenCodex model configuration support, including native Responses API selection, image inputs, context limits, and reasoning controls.
  - Exported configurations now include supported reasoning efforts, defaults, and input modalities.
- **Bug Fixes**
  - Configuration updates preserve unrelated YAML comments, formatting, and line endings.
  - Unsafe or ambiguous configuration edits are rejected without modifying files.
  - Oversized Chat Completions replay bodies now return structured errors.
- **Documentation**
  - Clarified configuration API behavior, formatting preservation, and ownership fingerprints.


Wibias (changes requested):
Requesting changes for one blocking correctness/safety issue in the integration ownership change.

### High — unrelated user-owned source text can now be silently destroyed

`classifyIntegration()` now intentionally ignores changes outside the recorded OpenCodex fragments, so a config can remain `current` when a user adds or edits an unrelated sibling such as another provider. That part is reasonable at the parsed-value level.

However, refresh/disable still parse the entire client config, remove/replace only the managed fragment in the parsed document, and then serialize the entire document back to disk. The YAML/TOML serializers are hand-renderers, and JSON/JSON5 are stringified, so comments and original formatting outside the managed fragment are not source-preserved.

Concrete example: after OpenCodex applies its provider block, the user adds an unrelated provider plus a YAML comment. With this PR the managed fragment still matches, so disable/refresh is allowed; the whole file is then re-rendered and the unrelated provider value survives, but the user's comment/formatting is silently lost.

That violates the writer's fail-closed safety contract: unrelated work performed after OpenCodex wrote the file must not be silently discarded merely because it sits outside the managed fragment.

Please either:

- mutate only the owned subtree using a source-preserving mechanism, or
- retain the whole-file conflict guard for formats where unrelated source text cannot be preserved.

Please also add an apply/refresh/disable regression that starts with an unrelated sibling plus comments/source formatting and proves those bytes survive the operation.

The OMP capability work itself looks sound: native OpenAI models use the model-level Responses API override, routed models remain on Chat Completions, image modalities are preserved, and the reasoning metadata matches OMP's accepted schema. The guarded Chat Completions replay-buffer 413 path also looks reasonable.

### Merge gates after the blocker is fixed

- rebase onto current `dev` (this head is behind)
- run exact-head Cross-platform CI / React Doctor successfully
- complete the readiness checklist / leave draft when ready

Please address the source-preservation hole before merge.

dbc-hbin:
Addressed the blocking source-preservation finding in the latest-dev rebased head `a328f438e`.

- Fragment-scoped ownership is now limited to OMP; whole-document serializers retain the whole-file fingerprint guard.
- OMP apply/refresh/disable patch only the block-style `providers.opencodex` source range.
- Ambiguous YAML or comments inside the managed range fail closed instead of falling back to whole-file serialization.
- The regression starts with a foreign provider plus comments/quoted formatting, performs apply, a stale refresh after an unrelated post-apply edit, and disable, then asserts the unrelated bytes are restored exactly.
- Exact-head full tests, focused 79 tests, typecheck, privacy scan, and docs build pass locally.

The PR remains draft while the new-head CI runs.

Wibias:
Thanks @dbc-hbin for the work here. This is genuinely useful: OMP now preserves the richer native OpenAI model capabilities through the Responses API, including image input and reasoning-effort metadata, while routed providers keep their existing Chat Completions path. The source-preserving YAML work also avoids clobbering unrelated user comments and formatting, which makes the integration much safer to use on real shared config files.

I'm closing this PR as superseded by #1392. #1392 contains the complete #1365 change set plus the follow-up source-preservation fix for nested inline comments inside `providers.opencodex`, together with refresh/disable regression coverage. That superset was rebased onto current `dev`, validated on the exact head, and has now been merged.

Thanks again for the contribution and for pushing the OMP integration forward. 🙏

PR #1368, opened 2026-08-09
fix(responses): clear stripped service tier from logs

luvs01:
## Summary

- Clear `requestedServiceTier` and its speed label after the final provider capability gate strips an unsupported `service_tier`.
- Add a live DeepSeek regression proving the upstream body and request-log context agree.
- Preserve configured-tier provenance and the tri-state behavior for supported and unclassified providers.

The request log captured caller tier metadata before final route normalization. DeepSeek and Volcengine correctly removed the unsupported wire field later, but the earlier log metadata still described the request as Fast/priority.

## Verification

- `bun test tests/service-tier-capability.test.ts` — 13 passed.
- Independent two-file review found no diff-introduced P0-P3 issues.
- The previous exact head completed full GitHub CI successfully; this update only rebases the same patch onto the latest `dev`.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected request handling for providers that do not support service tiers.
  * Service-tier settings are now consistently removed from outgoing requests and request logs when unsupported.
* **Tests**
  * Added regression coverage to verify service-tier data is cleared across the request flow.

### DIFF
--- a/src/server/responses/core.ts
+++ b/src/server/responses/core.ts
@@ -988,6 +988,10 @@ async function applyFinalRouteRequestNormalization(args: {
     parsed.options.serviceTier = tier;
   }
   applyServiceTierGate(route.provider, parsed._rawBody, parsed.options);
+  if (route.provider.adapter === "openai-responses" && route.provider.supportsServiceTier === false) {
+    logCtx.requestedServiceTier = undefined;
+    logCtx.requestedSpeedLabel = undefined;
+  }
 
   {
     const guidance = await multiAgentGuidanceText(parsed, {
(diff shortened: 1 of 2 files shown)


Wibias:

Wibias:
Thanks @luvs01 — merged. 🙌 This is a useful correctness fix: when a provider strips unsupported `service_tier`, the request log now reflects the request that was actually sent upstream instead of incorrectly reporting Fast/priority. The DeepSeek regression is especially helpful because it locks the wire body and log context together and prevents this mismatch from quietly returning.

PR #1371, opened 2026-08-09
fix(dashboard): safely render prototype-named providers

luvs01:
## Summary

- Group dashboard models with a `Map` so provider IDs such as `__proto__`, `constructor`, and `toString` cannot collide with inherited object properties.
- Guard provider display-name and icon alias tables with own-property checks so those same valid custom provider IDs render through the normal fallback path.
- Add focused regressions for repeated prototype-named groups and their display/icon behavior.

## Verification

- `cd gui && bun test tests` — 709 passed, 0 failed.
- `cd gui && bun run lint`
- `cd gui && bun run build`
- `bun run privacy:scan`
- `git diff --check`
- Independent focused diff review: no P0-P3 findings.
- Browser render check with synthetic API fixtures:

Dashboard model groups for prototype-named providers

## Checklist

- [x] Docs or release notes were updated when needed. (No documentation change is needed for this bug fix.)

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dashboard model grouping for provider names that overlap with built-in object properties.
  * Provider names and icons now fall back safely when no valid provider mapping exists.

* **Tests**
  * Added coverage for prototype-property provider names and safe provider display/icon handling.

### DIFF
--- a/gui/src/pages/use-dashboard-data.ts
+++ b/gui/src/pages/use-dashboard-data.ts
@@ -69,6 +69,16 @@ type CachedOverview = {
 
 type MaMode = "v1" | "default" | "v2";
 
+export function groupDashboardModels(models: ModelInfo[]): Array<[string, ModelInfo[]]> {
+  const groups = new Map<string, ModelInfo[]>();
+  for (const model of models) {
+    const rows = groups.get(model.provider);
+    if (rows) rows.push(model);
+    else groups.set(model.provider, [model]);
+  }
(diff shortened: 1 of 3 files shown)


Wibias:

Wibias:
Thanks @luvs01 — appreciate the clean fix here. Using `Map` plus own-property guards closes the prototype-name edge case without overhauling the dashboard path, and the focused regressions are solid. Merged! 🙏

PR #1372, opened 2026-08-09
fix(codex): keep unreadable shim checks advisory

luvs01:
## Summary

- Keep the post-install Codex shim readiness check advisory when the Codex config cannot be read or changes between existence and read checks.
- Reuse the existing `getCodexRoutingKind()` fallback so the install succeeds and emits the generic routing-verification warning instead of failing after the shim has already been written.
- Add a cross-platform end-to-end regression with `config.toml` represented by an unreadable directory and platform-appropriate Codex launchers.

## Verification

- `bun test tests/codex-shim-readiness.test.ts` — 6 passed, 0 failed.
- `bun run typecheck`
- `bun run privacy:scan`
- `git diff --check`
- Independent focused diff review: no P0-P3 findings.

## Checklist

- [x] Docs or release notes were updated when needed. (No documentation change is needed for this failure-path fix.)

## Review readiness checklist

### DIFF
--- a/src/cli/codex-shim-readiness.ts
+++ b/src/cli/codex-shim-readiness.ts
@@ -60,9 +60,16 @@ export function codexShimReadinessWarnings(
 
 export function collectCodexShimReadinessWarnings(): string[] {
   const config = loadConfig();
+  let externalProvider: string | null = null;
+  try {
+    externalProvider = currentExternalCodexModelProvider();
+  } catch {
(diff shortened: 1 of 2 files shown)


Wibias:

Wibias:
Thanks @luvs01 — good failure-path hardening. Keeping the post-install readiness check advisory avoids turning an unreadable/racy Codex config into a false install failure, while preserving the existing warning behavior. 🙏

PR #1386, opened 2026-08-10
fix(images): preserve bodyless responses with size headers

luvs01:
## Summary

- Treat a response with no body as zero bytes before evaluating its Content-Length metadata.
- Preserve bodyless image relay statuses such as 204 even when an upstream sends an oversized or stale Content-Length header.
- Extend the live image relay regression so this header cannot turn a bodyless response into a synthetic 502.

Follow-up to https://github.com/lidge-jun/opencodex/pull/1346#discussion_r3745763840.

## Verification

- Bun 1.3.14: `bun test tests/server-images.test.ts` — 63 passed.
- `git diff --check HEAD^ HEAD` — passed.

## Checklist

- [x] Docs or release notes were updated when needed; none are required for this internal response-body correction.
- [x] Security-sensitive changes were reviewed for untrusted upstream metadata and fail-closed behavior.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of image responses with no body.
  * Prevented empty responses with oversized declared content lengths from being incorrectly treated as oversized image data.
  * Preserved expected behavior for successful empty responses.

### DIFF
--- a/src/server/images.ts
+++ b/src/server/images.ts
@@ -84,6 +84,9 @@ export async function readImageResponseBytes(
     cancelUnlockedResponseBody(response, signal.reason);
     throw signal.reason;
   }
+  if (response.body === null) {
+    return { bytes: new Uint8Array(0), oversized: false };
+  }
 
   const maxBytes = options.maxBytes ?? IMAGES_RESPONSE_MAX_BYTES;
   const contentLength = response.headers.get("content-length");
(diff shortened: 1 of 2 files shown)


Wibias (approved):
Reviewed current head 9b44acc8. The production change correctly handles a null response body before Content-Length validation, which fixes the bodyless-response false oversize case while preserving the existing abort and bounded-body behavior. Cross-platform CI and React Doctor are green. Non-blocking note: the motivating standards-compliant example is a 304 with Content-Length; the regression uses a 204 with Content-Length, which is a malformed upstream case. The implementation still covers both, so this is not a merge blocker.

Wibias:
Thanks @luvs01! 🙏 This is a clean, focused fix for the bodyless image-response edge case. Merged into `dev`.

PR #1387, opened 2026-08-10
fix(providers): bound quota response parsing

luvs01:
## Summary

- Route every provider-quota JSON response through one 512 KiB bounded reader with total and inactivity deadlines.
- Preserve transient fallback and last-good semantics when a quota endpoint stalls, returns malformed JSON, or exceeds the budget.
- Cover the earlier Cursor period and summary probes as well as the final auth-usage fallback, and prevent new direct `Response.json()` quota probes.

## Verification

- Bun 1.3.14: `bun test tests/provider-quota.test.ts` — 97 passed on current `dev`.
- `bun run typecheck` — passed.
- `bun run privacy:scan` — passed.
- `git diff --check origin/dev...HEAD` — passed.
- Full `bun run test` was attempted once. The long Windows run hit existing effective-account/catalog isolation failures and ultimately a Bun 1.3.14 panic after 1,221 seconds; no changed-path failure was observed. It was not retried.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed; none are required for this internal quota-reader hardening.
- [x] Security-sensitive changes were reviewed for untrusted response bounds, cancellation, timeouts, and credential-safe fallback behavior.

## Review readiness checklist

This PR stays in draft until every box below is ticked. Tick all four boxes once the requirements are met:

- [x] All CI tests are green on my local testing.
- [x] I pushed my PR to the latest dev commit.
- [x] I resolved all correct Codex and CodeRabbit findings.

- [x] My PR is ready for review.

PR #1389, opened 2026-08-10
test(images): cover bodyless 304 content length

Wibias:
## Summary

- Add a focused regression for a bodyless HTTP 304 response that carries a representation `Content-Length` larger than the image relay cap.
- Assert the bodyless response is treated as zero bytes and is not marked oversized.
- Follow up on the review nit from #1386 with the standards-valid case that originally motivated the fix.

## Why

#1386 correctly fixed bodyless responses before evaluating `Content-Length`, but its integration regression used `204 + Content-Length`. A 304 response is the standards-valid case where `Content-Length` can describe the selected representation while the response itself has no body.

## Scope

Test-only. No production behavior changes.

## Validation

- Branch is based on current `dev` after #1386 and #1387.
- Diff is one new 14-line test file only.
- GitHub CI will provide the executable validation because the local runtime in this session does not have Bun/GitHub CLI available.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected handling of bodyless `304` responses with oversized `Content-Length` headers.
  * These responses now correctly return zero bytes without being flagged as oversized.

* **Tests**
  * Added coverage to verify the corrected response behavior.

### DIFF
--- a/tests/server-images-bodyless-content-length.test.ts
+++ b/tests/server-images-bodyless-content-length.test.ts
@@ -1,0 +1,14 @@
+import { expect, test } from "bun:test";
+import { IMAGES_RESPONSE_MAX_BYTES, readImageResponseBytes } from "../src/server/images";
+
+test("bodyless 304 ignores representation Content-Length when enforcing image response size", async () => {
+  const response = new Response(null, {
+    status: 304,
+    headers: { "content-length": String(IMAGES_RESPONSE_MAX_BYTES + 1) },
+  });
+
+  const observed = await readImageResponseBytes(response);
+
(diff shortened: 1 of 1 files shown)


Wibias (commented):
Reviewed the final head. No actionable findings.

The focused regression now covers the standards-valid case that motivated #1386: a bodyless HTTP 304 with a representation Content-Length above the image relay cap. It directly verifies `readImageResponseBytes()` treats the response as zero bytes and does not flag it as oversized. Scope is test-only, the PR is mergeable, and Cross-platform CI plus React Doctor are green.

From a code-review perspective, this is ready to merge.

PR #1404, opened 2026-08-10
fix(routing): complete unknown cost-cap operator contract (#1181)

### LINKED ISSUE #1181: Routing: define hard cost-cap behavior when runtime cost evidence is unknown

Wibias:
## Summary
- Continues and completes the #1181 work from #1345 (foreign fork head is not fast-forward writable after rebase onto current `dev`).
- Keeps default `onUnknownCost: \"allow\"` fail-open eligibility, but stamps `cost.capOutcome: \"unknown-allowed\"` on dry-run/live traces so operators can distinguish known-under-cap from unknown cost.
- Opt-in `exclude` still emits `cost-limit-unknown` with `capOutcome: \"unknown-excluded\"`; documents + exposes `onUnknownCost` in routing docs and the Routing Profile Editor.

## Addresses review on #1345
1. High — non-excluding `capOutcome` on the allow path (not an exclusion).
2. Medium — operator docs + dry-run/live integration coverage.
3. Merge gate — rebased onto current `dev`.

Credits the original #1345 implementation by @abhisheksharma2411.

Fixes #1181.

## Test plan
- [x] `bun test tests/cost-cap-unknown-evidence.test.ts tests/cost-scoring.test.ts`
- [x] `bun x tsc --noEmit`
- [x] `bun test ./gui/tests/routing-profiles.test.tsx`
- [ ] Exact-head CI green on this PR
- [ ] Confirm #1345 can be closed once this lands (or author cherry-picks)

## Summary by CodeRabbit

- **New Features**
  - Routing profiles can allow unknown cost estimates by default or exclude them with a fail-closed setting.
  - Live routing and dry-run results display cost-cap outcomes and detailed exclusion reasons.
  - Added localized cost-cap labels and exclusion messages across supported languages.

- **Documentation**
  - Updated routing guides and configuration references with the new setting, defaults, behavior, and examples.

- **Bug Fixes**
  - Improved consistency when handling incomplete, missing, or non-finite cost evidence.


abhisheksharma2411 (commented):
Thanks for carrying this forward, and for the credit — makes sense given the fork head wasn't fast-forward writable after the rebase. **Confirming #1345 can be closed once this lands**; no need for me to cherry-pick.

The `capOutcome` approach is a better answer to the operator-contract requirement than what I had. Keeping the non-excluding outcomes out of `exclusions` and whitelisting the values in `parseCost` both look right, and the four-state enum makes `satisfied` vs `unknown-allowed` legible in a way my version couldn't.

Two observations on the stamping in `evaluatePolicyProfile`:

**1. Stamping `limitUsd` can change `costScore()`'s denominator for callers that didn't supply it.**

`costForCandidate` now sets `limitUsd: costLimit` whenever a cap is configured, and `cost` is reassigned to it before scoring. `costScore()` picks its reference as:

```ts
const reference = typeof evidence.limitUsd === "number" && evidence.limitUsd > 0
  ? evidence.limitUsd
  : COST_SCORE_REFERENCE_USD;
```

So for any caller that passes cost evidence with a known `estimatedUsd` but no `limitUsd`, the reference silently moves from `COST_SCORE_REFERENCE_USD` (1.0) to the configured cap — which changes the score and therefore candidate ranking. Every in-tree caller (`router.ts`, `routing-profile-routes.ts`) already sets `limitUsd` via `costEvidenceForCandidate`, so I don't think this is reachable today. Flagging it as intent rather than a defect: if the normalization is deliberate, it might be worth a line in the comment saying so, since it's a scoring change hiding inside a trace-stamping change.

**2. Synthesized cost evidence has no `incomplete` marker.**

When a cap is set and the caller supplied no cost evidence at all, the candidate now gets `cost: { limitUsd, capOutcome: "unknown-allowed" }`. That block reads as present-but-unknown without `incomplete: true`, which `costEvidenceForCandidate` otherwise always sets for missing usage or price. Consider spreading in `incomplete: true` when synthesizing from nothing, so an operator (or anything consuming the trace) can't mistake it for evidence that simply lacks a price.

Neither blocks merge from my side. Happy to review again after the exact-head CI run, and thanks for the thorough review on #1345 — the fail-open ambiguity was a real gap.

Wibias:
Supersedes #1345. Continues @abhisheksharma2411's #1181 cost-cap work on current `dev` with the remaining operator-contract fixes.

abhisheksharma2411 (commented):
Re-reviewed at `b944703`. The docs alignment, the i18n mapping, and `normalizeRouteDecisionTrace` dropping `capOutcome` without a finite `limitUsd` all look right, and the dry-run/live parity tests answer the integration-coverage point cleanly.

One thing got sharper rather than resolved after `c248690`, and I think it needs a decision before merge.

**Normalizing `limitUsd` to `costLimit` also moves `costScore()`'s denominator.**

`costForCandidate` is used for two different purposes: it's stamped into the trace, *and* it's what gets passed to `costScore()` at the scoring step. CodeRabbit's finding was about the first (the trace must report the cap actually applied — agreed, and the new test at line 201 pins it). But the same object now feeds scoring, and `costScore()` picks its reference as:

```ts
const reference = typeof evidence.limitUsd === "number" && evidence.limitUsd > 0
  ? evidence.limitUsd
  : COST_SCORE_REFERENCE_USD;
return Math.max(0, Math.min(1, 1 - evidence.estimatedUsd / reference));
```

So for any caller whose evidence carries a `limitUsd` different from the profile cap, the cost score changes silently. Concretely, with `maxEstimatedCostUsd: 0.5` and incoming evidence `{ estimatedUsd: 0.4, limitUsd: 1 }`:

| | reference | cost score |
|---|---|---|
| before this PR | `1` (caller's) | `1 - 0.4/1` = **0.60** |
| at `b944703` | `0.5` (profile cap) | `1 - 0.4/0.5` = **0.20** |

The candidate stays eligible and correctly stamps `capOutcome: "satisfied"` in both cases — but its cost component drops by two thirds, and with `optimize.cost` weighted that can reorder candidates. The existing line-201 test doesn't catch it because `0.01 > 0.000001` makes that candidate ineligible anyway, so the score is never consulted.

I don't think this is reachable from `router.ts` or `routing-profile-routes.ts` today, since both build evidence via `costEvidenceForCandidate({ limitUsd: profile.limits.maxEstimatedCostUsd })` and the two values already agree. But line 201 establishes divergent `limitUsd` as a supported input shape, which makes the scoring path reachable for anyone constructing evidence directly.

Two ways to close it, either fine by me:

- **Separate the concerns** — keep the trace copy as-is, and score against the caller's original `evidence.cost`, so a trace-reporting fix can't move ranking.
- **Declare the normalization intentional** — state in the comment that the profile cap is the authoritative scoring reference and a caller-supplied `limitUsd` is advisory, then add a test asserting the score for an *eligible* candidate whose incoming `limitUsd` differs from the cap (e.g. the 0.4/0.5/1 case above), so the behaviour is pinned rather than incidental.

The second is probably the better contract — the profile is the authority — but right now it's implied by a change made for a different reason, and nothing pins it.

Nothing else blocking from me; the rest of the contract work reads well.

abhisheksharma2411:
Just saw `6be09595` — that's exactly it, and much cleaner than either option I suggested. Scoring off the caller's evidence and keeping the stamped copy purely for the trace makes the split obvious to the next person reading it, and the comment you left there says it better than my write-up did. Nice catch on threading `incomplete` through the synthesized case too.

Looks good to me. Thanks for being so quick on all of this — it's been genuinely fun to follow, and I appreciate you taking the time to explain the reasoning on #1345 rather than just closing it. Looking forward to seeing it land 🙂

abhisheksharma2411 (approved):
Fixes look good to me — the scoring split in `6be09595` is cleaner than either option I suggested, and threading `incomplete` through the synthesized case ties it off nicely.

Happy with this one. Thanks for the quick turnaround 🙂

Wibias:
Thanks @abhisheksharma2411 — both points were right. Separated scoring from the stamped trace copy (so `limitUsd` normalization can’t move `costScore` / ranking), and mark synthesized-from-nothing cost evidence `incomplete: true`. Covered in `6be09595`.

Wibias:
Thanks @abhisheksharma2411 — merging this.

Why it helps: your #1345 work correctly diagnosed that live routing builds cost evidence without usage, added the opt-in `onUnknownCost: "exclude"` path with `cost-limit-unknown`, and kept the default allow behavior non-breaking. #1404 finishes that operator contract on tip — `cost.capOutcome` on dry-run/live traces, docs/editor wiring, and scoring kept separate from trace stamping — so #1181 is complete.

Ship it.

Issue #857, opened 2026-08-01
[Bug]: stale Codex app-server makes injected roster disagree with live spawn_agent allowlist (2.8.0)


GENEXIS-AI:
## Client or integration

Codex App

## Area

Service lifecycle

## Summary

On OpenCodex 2.8.0, v2 guidance can advertise subagent models from a newly written on-disk catalog while a long-lived Codex app-server still validates `spawn_agent` against an older in-memory list. The injected roster listed valid models, but the active tool rejected an exact listed model and returned an empty allowlist.

Expected: OpenCodex should not positively advertise a roster that may be newer than the active app-server allowlist. Status/doctor/guidance should detect and explain the stale-process condition without restarting active work automatically.

This is related to #295, but the remaining gap is lifecycle drift after the roster-intersection fix: `effectiveSubagentRoster()` reads the current disk catalog on each request while Codex can retain an older catalog in memory.

## Reproduction

1. Start Codex Desktop so its app-server loads the current model catalog.
2. Change or sync the OpenCodex subagent roster/catalog without restarting the app-server.
3. Start a qualifying v2 turn.
4. Observe proxy-authored guidance such as `Available models: "gpt-5.6-sol", ...`.
5. Call `spawn_agent` using exactly `model="gpt-5.6-sol"`, `fork_turns="none"`, and a valid self-contained task.
6. Observe `Unknown model gpt-5.6-sol for spawn_agent. Available models:` with an empty list.

In the observed state, the app-server process started before both `opencodex-catalog.json` and `models_cache.json` were rewritten. `ocx agent status` and the injected guidance used the new disk roster; the live tool validator did not. A local session audit found 565 exact failures across 13 sessions.

Suggested safe behavior:

- If app-server start time predates the active catalog/cache mtime, suppress the positive roster claim and show an actionable restart warning.
- Surface the check in `ocx doctor` / `ocx agent status`, not only the one-shot `ocx sync` warning.
- Do not kill or restart an active app-server without explicit user action.
- Cover fresh, stale, and no-app-server cases with regression tests.

## Version

2.8.0

## Operating system

macOS 26.5.2 (25F84)

## Provider and model

OpenCodex v2 routed subagents; reproduced with `gpt-5.6-sol`. The configured roster also included `gpt-5.6-terra` and `kiro/claude-opus-5`.

## Logs or error output

```shell
Unknown model `gpt-5.6-sol` for spawn_agent. Available models:
```

Runtime details:

```text
[...]

Issue #1395, opened 2026-08-10
[Bug]: global stale-catalog check injects no-override guidance into fresh Codex sessions


LPFchan:
## Client or integration

Other — observed with a T3 Code-managed Codex app-server

I have not independently reproduced this with the bare Codex TUI.

## Area

Catalog / models

## Summary

OpenCodex 2.11.1 checks the start times of all Codex app-servers owned by the current user. If any app-server predates the on-disk model catalog, OpenCodex injects this developer message into every qualifying Multi-Agent V2 request:

```text
<multi_agent_mode>The model catalog changed after Codex started; do not set model or reasoning_effort overrides until Codex restarts.</multi_agent_mode>
```

This is safe when the app-server handling the current request is stale. The problem is that OpenCodex calculates one machine-wide state and cannot tell which app-server sent the request.

As a result, an unrelated stale process can inject no-override guidance into a fresh session whose app-server started after the catalog was written.

This code path does not reject model or reasoning parameters. It instructs the model not to provide them, so subsequent `spawn_agent` calls inherit their default model and effort.

## Observed behavior

Environment:

* OpenCodex 2.11.1
* Commit `121f1ad929dc6da3356c06f5192f2f97f7a5dde5`
* Codex CLI 0.147.0
* Ubuntu 24.04.4 LTS, x86_64
* Times below are UTC+09:00

Process timeline:

* Detached app-server A started at 00:43:03.
* `opencodex-catalog.json` was written at 04:25:10.
* The active T3-owned app-server B started at 10:24:33.
* A request handled by B still received the quoted no-override message.

OpenCodex therefore classified A as stale according to its timestamp rule, while B was clearly newer than the catalog.

The catalog contents from before the 04:25 write were not preserved, so this report does not claim that A’s actual model list differed. The demonstrated bug is that A’s timestamp controlled the guidance injected into B’s request.

Restarting `t3code.service` replaced the T3-owned app-servers but did not stop A, which belonged to a separate login session. New T3 sessions continued receiving the same message.

## Root cause

`collectCodexAppServerCatalogState()` calculates one result across every detected app-server:

```ts
const stale = withStarts.some(
  proc => proc.startedAtMs! <= catalogMtimeMs
);
```

If one process is old, the result for the entire set is `stale`.

`multiAgentGuidanceText()` then uses that machine-wide result when building guidance for an individual request:

```ts
if (catalogState.state === "stale" || catalogState.state === "unknown") {
[...]

PR #1407, opened 2026-08-10
fix(collaboration): scope catalog-state guidance to what we can attribute

### LINKED ISSUE #1354: [Bug]: transient code-mode-host process makes catalog staleness check return "unknown", emitting a false "catalog changed after Codex started" restart prompt (2.11.1, Windows)
## Client or integration

Codex App

## Area

Service lifecycle

## Summary

On OpenCodex 2.11.1 (Windows), v2 guidance intermittently emits:

```
<multi_agent_mode>The model catalog changed after Codex started; do not set model or reasoning_effort overrides until Codex restarts.</multi_agent_mode>
```

even though the catalog is **not** stale. In my session this alternated with the normal permissive `injectionPrompt` guidance from turn to turn, so `spawn_agent` model overrides were blocked unpredictably. Restarting Codex does not help, because the underlying state is `unknown`, not `stale`.

Root cause appears to be that `collectCodexAppServerCatalogState()` treats a missing process start time as `unknown`, and the guidance branch in `src/server/responses/collaboration.ts` maps `unknown` to the same user-facing text as `stale`:

```ts
// src/server/responses/collaboration.ts:248-252
const catalogState = await (deps.collectCatalogState ?? defaultCollectCatalogState)();
if (catalogState.state === "stale" || catalogState.state === "unknown") {
  return "<multi_agent_mode>The model catalog changed after Codex started; do not set "
    + "model or reasoning_effort overrides until Codex restarts.</multi_agent_mode>";
}
```

The message asserts a specific cause ("the catalog changed") and prescribes a specific remedy ("restart Codex"), but in the `unknown` branch neither is established. A restart cannot clear `unknown`.

### Why `unknown` is reached

`collectCodexAppServerCatalogState()` (`src/codex/app-server-processes.ts`) performs **two separate** PowerShell/CIM queries:

1. `listWindowsSnapshots()` — enumerate processes + `Invoke-CimMethod GetOwner` (line 319).
2. `readProcessStartMsBatch()` — fetch start times via `Get-CimInstance Win32_Process -Filter "ProcessId=..."` (line 513).

Then:

```ts
// line 631
if (catalogMtimeMs === null || withStarts.some(proc => proc.startedAtMs === null)) {
  return { state: "unknown", processes: withStarts, catalogMtimeMs };
}
```

`isCodexAppServerCommandLine()` deliberately includes `codex-code-mode-host` (line 217, via `isCodeModeHostProcess`). On this machine `codex-code-mode-host.exe` is **transient** — it spawns and exits around code-mode tool activity. If such a PID is captured in step 1 but exits before step 2, the `ProcessId=` filter returns no row, `startedAtMs` is `null`, and the whole state collapses to `unknown`.

The window between the two queries is not small. I timed step 1 ten consecutive times on this machine:

```
Run   Ms   OK  Incomplete
  1  2781  16           0
  2  2281  16           0
  3  2391  16           0
  4  2815  16           0
  5  2605  16           0
  6  2282  16           0
  7  2257  16           0
  8  2237  16           0
  9  2286  16           0
 10  2335  16           0
```

So there is a consistent ~2.2-2.8s gap during which a transient PID can disappear. This matches the observed behavior: the false message appeared far more often during bursts of tool calls (when code-mode hosts churn) and was stable when the session was idle.

### Evidence that the state is genuinely not stale

`ocx doctor` disagrees with the injected guidance, in the same session, minutes apart:

```
[OK] Codex app-server model catalog is current with the on-disk catalog.
```

Timestamps also rule out staleness under the code's own rule (`stale = startedAtMs <= catalogMtimeMs`):

| Item | Time (local) |
|---|---|
| `.codex\opencodex-catalog.json` mtime | 13:40:25 |
| `codex.exe` PID 13300 start | 14:13:48 |
| `codex.exe` PID 5172 start | 14:15:34 |
| `codex-code-mode-host.exe` PID 21368 start | 14:20:38 |

Every app-server process started *after* the catalog was written, so `stale` is impossible here. Config files were likewise untouched after process start:

| File | Last write |
|---|---|
| `.codex\opencodex-catalog.json` | 13:40:25 |
| `.codex\config.toml` | 14:10:01 |
| `.opencodex\config.json` | 11:14:29 |

Functionally, the permissive path works: when guidance was in its normal state, `spawn_agent` with `model="gpt-5.6-sol"`, `reasoning_effort="medium"`, `fork_turns="none"` ran to completion successfully. So this is a false negative in detection, not a real allowlist mismatch.

## Reproduction

1. Windows 11, OpenCodex 2.11.1, Codex CLI 0.148.0-alpha.5, `subagentModels` configured with `gpt-5.6-sol` / `gpt-5.6-terra`.
2. Ensure the on-disk catalog mtime is older than every running app-server process (i.e. a genuinely fresh state), and confirm with `ocx doctor` → `[OK] ... catalog is current`.
3. Run a v2 session that issues frequent tool calls, so `codex-code-mode-host.exe` processes spawn and exit repeatedly.
4. Observe the injected guidance across turns: it intermittently flips to the "catalog changed after Codex started" text and back to the configured `injectionPrompt`.
5. Restarting Codex does not stop the recurrence.

## Suggested fix

- Do not collapse `unknown` into the `stale` message. Either emit a distinct, accurate text (e.g. "could not verify the model catalog state") or omit the positive roster without prescribing a restart that cannot help.
- Treat a PID that vanishes between enumeration and start-time lookup as *process exited*, not as an unreadable start time — drop it from consideration rather than poisoning the aggregate state.
- Consider excluding short-lived `codex-code-mode-host` processes from the staleness aggregate. They are relevant for restart targeting, but their lifetime makes them a poor signal for "does the running app-server hold an old catalog".
- Optionally capture start time in the same query as enumeration, removing the race entirely.

## Version

2.11.1

## Operating system

Windows 11 Pro, 10.0.26200 (build 26200)

## Provider and model

OpenCodex v2 routed subagents; `subagentModels` = `gpt-5.6-sol`, `gpt-5.6-terra`. Codex CLI 0.148.0-alpha.5.

## Logs or error output

Injected guidance observed intermittently (roughly five times in one session, interleaved with normal guidance):

```text
<multi_agent_mode>The model catalog changed after Codex started; do not set model or reasoning_effort overrides until Codex restarts.</multi_agent_mode>
```

Contradicted by, in the same session:

```text
[OK] Codex app-server model catalog is current with the on-disk catalog.
```

Not required; process/catalog timestamps and timings above are text-reproducible.

## Redacted configuration

```json
{
  "subagentModels": ["gpt-5.6-sol", "gpt-5.6-terra"],
  "injectionPrompt": "spawn_agent supports optional \"model\" and \"reasoning_effort\" arguments. When setting either, set fork_turns to \"none\" or a positive turn count, and make the message self-contained.{{roster}}",
  "websockets": false
}
```

## Notes

Related to #857 (closed), which added this staleness detection. The detection logic itself looks correct for the genuine stale case; the gap is the `unknown` fallback sharing the stale wording, plus the enumerate-then-stat race against transient processes.

One caveat on attribution: the code paths, timings, `ocx doctor` output, and timestamps above are directly verified. The specific race between the two CIM queries is a strong inference from those observations rather than a captured failure — the injection debug log endpoint returned 404 in this version, so I could not read the internal per-turn state directly. If there is a supported way to dump `collectCodexAppServerCatalogState()` results per turn, I am happy to capture it.

## Checks

abhisheksharma2411:
Fixes #1354.

## Summary

v2 collaboration guidance mapped `unknown` to the same text as `stale`:

```ts
// src/server/responses/collaboration.ts
if (catalogState.state === "stale" || catalogState.state === "unknown") {
  return "<multi_agent_mode>The model catalog changed after Codex started; do not set "
    + "model or reasoning_effort overrides until Codex restarts.</multi_agent_mode>";
}
```

`unknown` does not mean the catalog changed. It means the comparison could not be made at all. `collectCodexAppServerCatalogState` returns `unknown` when the catalog mtime is unreadable, when process enumeration fails, or when **any** app-server's start time is unreadable:

```ts
// src/codex/app-server-processes.ts
if (catalogMtimeMs === null || withStarts.some(proc => proc.startedAtMs === null)) {
  return { state: "unknown", processes: withStarts, catalogMtimeMs };
}
```

A transient `code-mode-host` process that exits between enumeration and the start-time read lands exactly there, which is why the message appears intermittently and alternates with normal guidance from turn to turn, as the reporter describes.

So the text asserted a cause that was never established, and prescribed a remedy that cannot work — restarting Codex does not clear `unknown`, because the state is a failed measurement rather than a stale process.

**Suppressing overrides while the state is unknown is still correct and is unchanged.** Only the claim changes. `unknown` now gets its own message, mirroring the honest phrasing `ocx doctor` already uses for the same state (`src/cli/doctor.ts`):

> Could not verify whether Codex's model catalog is current (app-server start time or catalog timestamp unreadable); do not set model or reasoning_effort overrides for this turn.

An operator reading a turn can now tell "the catalog changed" apart from "we could not check", and is no longer told to perform a restart that would not help.

## Verification

Reproduced with a failing test **before** the fix — `unknown` returned the stale text verbatim:

```
expect(unknown).not.toContain("The model catalog changed after Codex started")
  Received: "<multi_agent_mode>The model catalog changed after Codex started; do not set
             model or reasoning_effort overrides until Codex restarts.</multi_agent_mode>"
(fail) v2 guidance does not assert a catalog change or a restart remedy when the state is unknown (#1354)
```

Commands run against the `dev` base (branch is on `3976d34`):

```
bun test tests/multi-agent-compat.test.ts   → 42 pass, 0 fail
bun x tsc --noEmit                          → clean
coverage src/server/responses/collaboration.ts → 84.00% lines
```

The new test pins four things: `stale` keeps its exact wording; `unknown` asserts neither the cause nor the restart remedy; both still suppress the override instruction and positive model claims (so the #857 behaviour cannot regress); and the two messages stay distinguishable from each other.

The existing #857 test — which asserts both states suppress positive claims — passes unchanged.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup. — two files, one behavioural branch split, no drive-by changes.
- [x] Docs or release notes were updated when needed. — no user-facing document describes this message; the `unknown` wording is deliberately aligned with the existing `ocx doctor` text so the two surfaces agree. Happy to add a note if you'd like one.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. — no secrets or auth touched. The change is strictly less assertive: it removes a false claim while keeping the conservative suppression, so no path becomes more permissive.

## Notes for review

- `not_running` and `fresh` are untouched.
- I kept "do not set … overrides" in the `unknown` text so the instruction itself is unchanged; only the justification and the remedy differ.
- Wording is lifted from `doctor.ts` on purpose — if you'd rather the two surfaces diverge, or prefer different phrasing, happy to adjust.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Suppressed model and reasoning-effort guidance when catalog information is stale or unavailable.
  * Preserved appropriate guidance for fresh and inactive catalog states.
  * Improved handling of mixed catalog states without applying blanket override restrictions.

* **Documentation**
  * Added guidance explaining catalog states and how untrusted catalog information affects recommendations.

* **Tests**
  * Added regression coverage for stale, unknown, mixed, fresh, and inactive catalog states.

### DIFF
--- a/docs-site/src/content/docs/guides/sub-agent-surface.md
+++ b/docs-site/src/content/docs/guides/sub-agent-surface.md
@@ -228,7 +228,8 @@ may well be fresh. The active tool schema stays authoritative.
 
 `unknown` is not a synonym for `stale`. It means the comparison itself failed — an unreadable catalog
 timestamp, an unreadable process start time, or a failed process enumeration — and it is reported
-separately by `ocx doctor`. Restarting Codex clears `stale`; it does not necessarily clear `unknown`.
+separately by `ocx doctor`. `stale` clears only after every detected Codex app-server starts after
+the final catalog write; it does not necessarily clear `unknown`.
 
 ### Reasoning effort
 
--- a/docs-site/src/content/docs/guides/sub-agent-surface.md
+++ b/docs-site/src/content/docs/guides/sub-agent-surface.md
@@ -204,6 +204,32 @@ to v1. A `"v2"`, `null`, or absent surface value is eligible; a real `"v1"` pin 
 No. Start a new Codex session after changing the mode. If a long-running App host still shows stale
 catalog state, run `ocx sync` and restart that Codex surface.
 
+### What happens when opencodex cannot trust the catalog?
+
+opencodex compares the on-disk model catalog against the start time of every Codex app-server owned
+by the current user, producing one of four states:
+
+| State | Meaning | v2 guidance |
+|---|---|---|
+| `fresh` | Every app-server started after the catalog was written | Full guidance: preferred model, roster, fallbacks |
+| `not_running` | No app-server detected | Full guidance |
+| `stale` | At least one app-server predates the catalog | **No opencodex-authored model guidance** |
+| `unknown` | The comparison could not be made | **No opencodex-authored model guidance** |
+
+For `stale` and `unknown`, opencodex withholds its own disk-derived claims — preferred model, roster,
+fallback and custom guidance — because the running Codex may not be able to spawn what the disk
+catalog advertises.
+
+It does **not** instruct the model to stop setting `model` or `reasoning_effort`. That observation is
+global across every app-server for the user, while an inbound request carries no sender identity, so
+a stale process cannot be attributed to the request in front of us. Prohibiting overrides on that
+basis would block options the active `spawn_agent` tool legitimately advertises, for a session that
+may well be fresh. The active tool schema stays authoritative.
+
+`unknown` is not a synonym for `stale`. It means the comparison itself failed — an unreadable catalog
+timestamp, an unreadable process start time, or a failed process enumeration — and it is reported
+separately by `ocx doctor`. Restarting Codex clears `stale`; it does not necessarily clear `unknown`.
+
 ### Reasoning effort
 
 `injectionEffort` affects only delegated-worker guidance and, when explicitly enabled, native Codex
--- a/src/server/responses/collaboration.ts
+++ b/src/server/responses/collaboration.ts
@@ -249,9 +249,24 @@ export async function multiAgentGuidanceText(
     // Codex cannot actually spawn makes spawn_agent reject the override, so
     // suppress positive model claims while the state is stale or unknown.
     const catalogState = await (deps.collectCatalogState ?? defaultCollectCatalogState)();
+    // #1354 / #1395: `collectCodexAppServerCatalogState()` folds every app-server
+    // owned by the current user into ONE global observation, and the inbound
+    // request carries no sender PID or catalog fingerprint. So a stale process A
+    // makes the global state `stale` even when this request came from a fresh
+    // process B, and `unknown` can be reached by a process-enumeration failure
+    // that says nothing about any particular server.
+    //
+    // Emitting "do not set model or reasoning_effort overrides" off that global
+    // observation prohibits options the active `spawn_agent` tool legitimately
+    // advertises, for a request we cannot attribute to the stale process. The
+    // safe behaviour is to withhold OpenCodex's own disk-derived claims —
+    // preferred model, roster, fallback, custom guidance — and stay silent about
+    // overrides, leaving the active tool schema authoritative.
+    //
+    // `fresh` and `not_running` are unchanged: there we can positively describe
+    // the catalog, so the guidance below still applies.
     if (catalogState.state === "stale" || catalogState.state === "unknown") {
-      return "<multi_agent_mode>The model catalog changed after Codex started; do not set "
-        + "model or reasoning_effort overrides until Codex restarts.</multi_agent_mode>";
+      return null;
     }
     // codex-rs supplies the Proactive text on v2; the proxy only adds model-designation
     // guidance, and only when there is something concrete to designate: a configured
--- a/tests/multi-agent-compat.test.ts
+++ b/tests/multi-agent-compat.test.ts
@@ -11,6 +11,7 @@ import { injectDeveloperMessage, multiAgentGuidanceText, sanitizeEncryptedConten
 import { parseRequest } from "../src/responses/parser";
 import type { OcxParsedRequest } from "../src/types";
 import { CODEX_ACCOUNT_BOUND_CATALOG_KIND, effectiveSubagentRoster } from "../src/codex/catalog";
+import { collectCodexAppServerCatalogState } from "../src/codex/app-server-processes";
 import { clearDebugSettings, setDebugSettings } from "../src/lib/debug-settings";
 import {
   getInjectionDebugLogEntries,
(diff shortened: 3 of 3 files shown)


Ingwannu (dismissed):
The wording split is directionally useful, but the current head still preserves the underlying request-scoping bug confirmed in #1395, so I am requesting changes.

`collectCodexAppServerCatalogState()` folds every current-user app-server into one global state, and the inbound request carries no sender PID or catalog fingerprint. A known-stale process A can therefore make a fresh process B receive this PR's `do not set model or reasoning_effort` instruction. Rewording only `unknown` does not fix that case, and the same identity limitation applies to both `stale` and `unknown`.

For either state, suppress OpenCodex's disk-derived preferred-model, roster, fallback, and custom guidance, but do not prohibit options advertised by the active `spawn_agent` tool. The minimal safe behavior is to return no OCX-authored v2 guidance for `stale` or `unknown`; keep `fresh` and `not_running` unchanged.

Please update the regression so a mixed stale/fresh process set can still produce the global `stale` observation without producing a blanket no-override instruction for an unidentified request. Also update the catalog/subagent behavior documentation, because this changes the user-visible guidance contract. Once that is done on current `dev` and exact-head CI is green, this can cover both #1354 and #1395 without a duplicate PR.

Wibias (changes requested):
Requesting changes. The wording split is useful, but the current head is not merge-ready.

## Merge blockers

### 1. Global stale/unknown state still blocks valid overrides for unrelated fresh sessions

`collectCodexAppServerCatalogState()` folds every detected current-user app-server into one global state. A stale process A therefore makes the global result `stale`, even if the current request came from a fresh process B. `multiAgentGuidanceText()` has no sender PID/catalog identity, yet this PR still injects `do not set model or reasoning_effort overrides` for that global state.

That preserves the confirmed #1395 request-scoping bug. For `stale` and `unknown`, suppress OpenCodex-authored disk/catalog guidance, but do not prohibit model/reasoning options exposed by the active `spawn_agent` tool. The minimal safe behavior is to emit no OCX-authored v2 model guidance for those states, or otherwise make the active tool schema authoritative.

### 2. The new regression test locks in that incorrect contract

The added test explicitly requires both `stale` and `unknown` to contain `do not set`, so it codifies the #1395 bug instead of protecting the intended boundary.

Please add a mixed stale/fresh regression: one process starts before the catalog, another after it, collector resolves globally to `stale`, and an unidentified current v2 request must not receive a blanket no-override instruction.

## Additional correctness gap

The new `unknown` message says `(app-server start time or catalog timestamp unreadable)`, but `unknown` can also result from process-enumeration failure. That still asserts a cause which is not established. If any `unknown` guidance remains, keep it cause-neutral, e.g. `Could not verify whether Codex's model catalog is current.`

## Docs / readiness

Because #1407 is now the selected overlapping fix path for #1354 and #1395, update the catalog/subagent guidance documentation to describe the actual stale/unknown contract.

The reviewed head is also behind current `dev`, and exact-head React Doctor / Cross-platform CI have not executed jobs yet (`action_required`, zero jobs). Rebase or merge current `dev` and get exact-head CI green before merge.

## Security

I found no new exploitable security vulnerability in this diff. The blocking issues are correctness, request scoping, and regression coverage.

abhisheksharma2411:
Thanks both — you're right, and the second point is the one that stings: my regression asserted `do not set` for both states, so it actively locked in the #1395 bug rather than protecting against it. Fixed on `550091d`.

**What changed**

`stale` and `unknown` now return `null` — no opencodex-authored v2 model guidance at all. Preferred model, roster, fallback and custom guidance are withheld, and nothing is said about overrides, so the active `spawn_agent` schema stays authoritative. `fresh` and `not_running` are untouched.

**Regression replaced.** The new one builds a genuinely mixed process set through the collector — pid 42 starting before the catalog, pid 43 after — asserts the global observation really does resolve to `stale`, and then asserts an unidentified v2 request gets no blanket no-override instruction. That's the case the old test was papering over.

**On the cause-neutrality point** — you were right and I'd made the same mistake I was fixing. My `unknown` text said `(app-server start time or catalog timestamp unreadable)` while enumeration failure reaches that state too. Moot now that `unknown` emits nothing, but it was a fair catch.

**Docs** — added a four-state table to the sub-agent surface guide covering what each state does to guidance, why a global observation can't be attributed to a request, and why restarting clears `stale` but not necessarily `unknown`.

```
bun test tests/multi-agent-compat.test.ts         -> 43 pass, 0 fail
bun test tests/codex-app-server-processes.test.ts -> 30 pass, 1 skip, 0 fail
bun x tsc --noEmit                                -> clean
```

Rebased onto current `dev` (`550091d` sits directly on tip), so exact-head CI should have something to run now. Retitled and relabelled to reflect that this covers #1395 as well.

Happy to keep iterating if the null-guidance shape isn't quite what you had in mind for `stale` — I read "make the active tool schema authoritative" as returning nothing rather than emitting a softened message, but say the word if you'd prefer the latter.

Wibias (approved):
Approving. The head scopes catalog-state guidance to what we can attribute: stale/unknown withhold disk-derived claims without a false restart/no-override prescription for unidentified requests, with regressions for mixed stale/fresh and unknown.

Wibias:
Thanks @abhisheksharma2411 — this was useful because it stopped treating a failed catalog measurement (`unknown`) like a proven stale catalog, and then went further to stop global stale/unknown from blanket-blocking overrides on requests we cannot attribute to a specific app-server. Operators (and agents) now get honest guidance that matches `ocx doctor`, without a restart remedy that could never clear a measurement failure.

Merging now.

PR #1369, opened 2026-08-09
fix(oauth): guard local token expiry parsing against NaN

Bruce-Yii:
## Summary

Fixes [#1366](https://github.com/lidge-jun/opencodex/issues/1366): an imported local CLI credential (Grok `~/.grok/auth.json` or Claude `.credentials.json`) whose `expires_at` is invalid/unparseable (e.g. `"not-a-date"`, `NaN`, epoch-garbage like `1970-01-22`) was **treated as valid and never refreshed**, and `ocx status` showed **`✓ logged in`** for it.

## Root cause

`src/oauth/local-token-detect.ts` — `new Date(entry.expires_at as string).getTime()` returns `NaN` for unparseable strings. Downstream comparisons then misbehave:

- `shouldAdoptGrokGeneration()`: `disk.expires <= now + refreshSkewMs` is **always `false`** for `NaN` (passes the expiry gate); `bothExpiriesExist` is `false` because `NaN > 0` is `false`, so it falls through to `return true` — the garbage credential is **adopted** as the authoritative generation (`authoritative()` in `src/oauth/index.ts`), and the refresh path never re-validates it.
- `getLoginStatus()` reported `loggedIn: !!cred` (credential exists), ignoring `needsReauth` — the CLI "OAuth logins" section printed `✓ logged in` for reauth-required credentials.

## Changes

1. **`src/oauth/local-token-detect.ts`** — `detectGrokCliToken()`: `Number.isFinite` guard on the parsed `expires_at`; non-finite → `0` (unknown → forces the refresh-validation path downstream instead of "valid forever").
2. **`src/oauth/local-token-detect.ts`** — `shouldAdoptGrokGeneration()`: explicit `!Number.isFinite(disk.expires) → false` (defense-in-depth; a garbage disk credential is never adopted as an upgrade).
3. **`src/oauth/local-token-detect.ts`** — `parseClaudeOauthPayload()`: `Number.isFinite` guard on `expiresAt` (strings/NaN → `0`).
4. **`src/oauth/index.ts`** — `getLoginStatus()`: `loggedIn = !!cred && !needsReauth`. `needsReauth` is the authoritative login-failure signal; an expired-but-refreshable access token stays logged in per the lazy-refresh contract (request resolution refreshes it on demand). Invalid/unknown local-import expiries are handled solely at parse/adoption time (points 1–3), never by over-reporting login state.

## Tests

[...]

PR #1418, opened 2026-08-10
fix(oauth): guard expires_in parsing against NaN across token responses

### LINKED ISSUE #1417: [Bug]: OAuth token responses with invalid expires_in produce NaN expiries that never refresh
### Client or integration

Codex CLI / Direct HTTP/API client

### Area

### Summary

OAuth token responses are parsed into credential expiries in four places without a `Number.isFinite` guard on `expires_in`. A malformed upstream response (missing, `NaN`, or string `expires_in`) yields a `NaN` **expiry** that flows into downstream time comparisons, where `NaN <= x` is always `false` — the credential is treated as never-expiring and never refreshed. This is the same data-quality class as the recently merged lidge-jun/opencodex#1366/lidge-jun/opencodex#1369 fix (`Number.isFinite` guards on local token expiry parsing) that remains unfixed at the token-response parse sites.

Affected sites:

1. `src/oauth/anthropic.ts:87` — `Date.now() + data.expires_in * 1000 - 5 * 60 * 1000` — missing `expires_in` → `undefined * 1000` = `NaN`
2. `src/oauth/kimi.ts:163` — `Date.now() + payload.expires_in * 1000 - OAUTH_EXPIRY_SKEW_MS` — line 154 checks `typeof !== "number"`, but `NaN` passes (`typeof NaN === "number"`)
3. `src/oauth/chatgpt.ts:55` — `Date.now() + ((data.expires_in as number) ?? 3600) * 1000` — `??` only guards `null`/`undefined`; `NaN` or a numeric string passes through
4. `src/codex/account-store.ts:498` — `Date.now() + data.expires_in * 1000` — no validation at all

Reference implementation with the correct guard already exists at `src/oauth/xai.ts:128-129`:

```ts
const expiresIn = typeof payload.expires_in === "number" && Number.isFinite(payload.expires_in) ? payload.expires_in : 3600;
```

### Reproduction

Mock an upstream OAuth token response with `expires_in` missing (or `null`/`NaN`):

```json
{ "access_token": "...", "refresh_token": "...", "expires_in": null }
```

Observe the stored credential carries `expires: NaN`, so `expires <= now` is always false — the credential never refreshes and `ocx status` reports it as logged in indefinitely (its health/refresh machinery cannot detect expiry).

Expected: a non-finite `expires_in` falls back to the same default used for a missing field (e.g. 3600s) so the credential stays refreshable.

### Version

dev `0de4fd2d`

### Operating system

Cross-platform

### Provider and model

anthropic / kimi / chatgpt / codex (account-store) — OAuth token response parsing

### Checks

Bruce-Yii:
## Summary

Fixes [#1417](https://github.com/lidge-jun/opencodex/issues/1417): OAuth token responses with an invalid `expires_in` (missing, `NaN`, string, negative, or numeric-overflowing) produce a **`NaN`/`Infinity`/past expiry** in stored credentials. `NaN <= x` is always false and `Infinity` never expires, so such credentials break the refresh loop. This is the token-response counterpart of the already-merged [#1369](https://github.com/lidge-jun/opencodex/pull/1369) fix — same data-quality class, remaining at the upstream parse sites.

## Root cause

- `src/oauth/anthropic.ts` — `Date.now() + data.expires_in * 1000` — missing `expires_in` → `undefined * 1000` = `NaN`
- `src/oauth/kimi.ts` — `typeof !== "number"` check lets `NaN` through (`typeof NaN === "number"`)
- `src/oauth/chatgpt.ts` — `?? 3600` only guards `null`/`undefined`; `NaN` or a numeric string passes through
- `src/codex/account-store.ts` — no validation at all

Additional failure modes found in review:
- **Overflow (CodeRabbit Major):** a finite-but-huge `expires_in` (e.g. `Number.MAX_VALUE`) passes `Number.isFinite` but overflows to `Infinity` once multiplied by 1000 — the *computed* expiry must also be guarded.
- **Negative (maintainer review):** a finite negative `expires_in` produces a finite already-past expiry, silently bypassing the malformed-response fallback.

## Changes

All four sites now require `typeof === "number" && Number.isFinite(...) && >= 0` (falling back to a 3600s default so the credential stays refreshable), **plus a guard on the computed expiry timestamp**:

1. **`src/oauth/anthropic.ts`** — `credsFrom()`: non-finite/negative/overflowing → 3600s default.
2. **`src/oauth/kimi.ts`** — `parseTokenPayload()`: non-finite, negative, or overflowing `expires_in` rejected as malformed.
3. **`src/oauth/chatgpt.ts`** — `credsFromToken()`: non-finite/negative/overflowing → 3600s default.
4. **`src/codex/account-store.ts`** — refresh path: non-finite/negative/overflowing → 3600s default.

Per the lazy-refresh contract established in #1369, this change only guards at parse/adoption time — it does not touch `getLoginStatus` or expiry comparison logic.

## Tests

- `tests/chatgpt-token-expiry.test.ts` (new): non-finite, string, overflowing, and negative `expires_in` all fall back to **exactly ~3600s** (tolerance 30s)
- `tests/anthropic-hardening.test.ts`: non-finite, overflowing, and negative → **~3300s** (3600s minus the 5-minute skew)
- `tests/codex-account-store.test.ts`: non-finite, overflowing, and negative → persisted `expiresAt` **~3600s**
- `tests/kimi-oauth-identity.test.ts`: non-finite, overflowing, and negative `expires_in` all rejected

Verification (head `355b69e5`, rebased onto `dev` 9c051342):
- Targeted: 55/55 pass
- OAuth + anthropic + account-store suite: **414/414 pass**
- `bun x tsc --noEmit` → clean
- hygiene: **passed** (maintainer-sponsored applied)
- **cross-platform CI: green** (confirmed by maintainer review on exact head)
- `mergeable_state`: clean

Fixes #1417

## Review readiness checklist

### DIFF
--- a/src/codex/account-store.ts
+++ b/src/codex/account-store.ts
@@ -491,11 +491,15 @@ export async function getValidCodexToken(id: string): Promise<CodexTokenResult> 
       throw new TokenRefreshError(reason, `Codex token refresh failed (${reason}); reauthenticate the account.`);
     }
     const data = (await res.json()) as { access_token: string; refresh_token?: string; expires_in: number };
+    // Guard against a missing/non-finite expires_in (malformed upstream response):
+    // a NaN expiry would never compare as expired and would block refresh forever.
+    const expiresIn =
+      typeof data.expires_in === "number" && Number.isFinite(data.expires_in) ? data.expires_in : 3600;
 
     const updated: CodexAccountCredentials = {
       accessToken: data.access_token,
       refreshToken: data.refresh_token ?? lockedCred.refreshToken,
-      expiresAt: Date.now() + data.expires_in * 1000,
+      expiresAt: Date.now() + expiresIn * 1000,
       chatgptAccountId: lockedCred.chatgptAccountId,
     };
     if (!saveCodexAccountCredentialIfGeneration(id, startGeneration, updated)) {
--- a/src/codex/account-store.ts
+++ b/src/codex/account-store.ts
@@ -491,10 +491,13 @@ export async function getValidCodexToken(id: string): Promise<CodexTokenResult> 
       throw new TokenRefreshError(reason, `Codex token refresh failed (${reason}); reauthenticate the account.`);
     }
     const data = (await res.json()) as { access_token: string; refresh_token?: string; expires_in: number };
-    // Guard against a missing/non-finite expires_in (malformed upstream response):
-    // a NaN expiry would never compare as expired and would block refresh forever.
+    // Guard against a missing/non-finite/negative expires_in (malformed upstream
+    // response): a NaN expiry would never compare as expired, and a negative
+    // duration would stamp an already-past expiry — both block refresh semantics.
     const expiresIn =
-      typeof data.expires_in === "number" && Number.isFinite(data.expires_in) ? data.expires_in : 3600;
+      typeof data.expires_in === "number" && Number.isFinite(data.expires_in) && data.expires_in >= 0
+        ? data.expires_in
+        : 3600;
     // The computed timestamp itself must stay finite: Number.MAX_VALUE passes
     // Number.isFinite but overflows to Infinity once multiplied by 1000.
     const expiresAt = Date.now() + expiresIn * 1000;
--- a/src/codex/account-store.ts
+++ b/src/codex/account-store.ts
@@ -495,11 +495,15 @@ export async function getValidCodexToken(id: string): Promise<CodexTokenResult> 
     // a NaN expiry would never compare as expired and would block refresh forever.
     const expiresIn =
       typeof data.expires_in === "number" && Number.isFinite(data.expires_in) ? data.expires_in : 3600;
+    // The computed timestamp itself must stay finite: Number.MAX_VALUE passes
+    // Number.isFinite but overflows to Infinity once multiplied by 1000.
+    const expiresAt = Date.now() + expiresIn * 1000;
+    const safeExpiresAt = Number.isFinite(expiresAt) ? expiresAt : Date.now() + 3600 * 1000;
 
     const updated: CodexAccountCredentials = {
       accessToken: data.access_token,
       refreshToken: data.refresh_token ?? lockedCred.refreshToken,
-      expiresAt: Date.now() + expiresIn * 1000,
+      expiresAt: safeExpiresAt,
       chatgptAccountId: lockedCred.chatgptAccountId,
     };
     if (!saveCodexAccountCredentialIfGeneration(id, startGeneration, updated)) {
--- a/src/oauth/anthropic.ts
+++ b/src/oauth/anthropic.ts
@@ -85,10 +85,14 @@ function credsFrom(data: AnthropicTokenResponse, refreshFallback?: string): OAut
   // a NaN expiry would never compare as expired and would block refresh forever.
   const expiresIn =
     typeof data.expires_in === "number" && Number.isFinite(data.expires_in) ? data.expires_in : 3600;
+  // The computed timestamp itself must stay finite: Number.MAX_VALUE passes
+  // Number.isFinite but overflows to Infinity once multiplied by 1000.
+  const computedExpires = Date.now() + expiresIn * 1000 - 5 * 60 * 1000;
+  const expires = Number.isFinite(computedExpires) ? computedExpires : Date.now() + 3600 * 1000 - 5 * 60 * 1000;
   return {
     refresh: data.refresh_token || refreshFallback || "",
     access: data.access_token,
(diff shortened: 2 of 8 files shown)


Bruce-Yii:
Addressed the CodeRabbit Major finding (computed-expiry overflow) in commit `3b7b7471`: `Number.MAX_VALUE` passes `Number.isFinite` but overflows to `Infinity` when multiplied by 1000, so all four sites now also guard the **computed** expiry timestamp (fallback 3600s; Kimi rejects as malformed). Added 4 regression tests for the overflowing case. Head `3b7b7471`, targeted 51/51, OAuth+anthropic+account-store suite 410/410, `tsc --noEmit` clean.

flyingsquirrel0419 (changes requested):
The fix direction is correct and the computed-expiry overflow guard closes the original `NaN`/`Infinity` failure mode, but I cannot sponsor or approve this authentication-boundary change on the current head.

One blocker remains across all four parse sites: finite negative `expires_in` values are still accepted. In `src/codex/account-store.ts`, `src/oauth/anthropic.ts`, and `src/oauth/chatgpt.ts`, a negative duration produces a finite timestamp in the past instead of taking the malformed-response fallback; in `src/oauth/kimi.ts`, it should fail the required-field contract. Require `expires_in >= 0` in the three fallback paths and reject `< 0` in Kimi, then add negative-value regression coverage for each behavior.

Please also tighten the new fallback tests to assert the actual fallback window (approximately 3600 seconds, or approximately 3300 seconds for Anthropic after its five-minute skew), rather than only `finite && future`. That prevents a one-second or otherwise incorrect fallback from satisfying the tests.

The failing `hygiene` check is the expected `unsponsored_surface` gate, not a product-test failure. I will reconsider `maintainer-sponsored` only after these input-boundary cases are fixed and the exact-head checks are green.

Bruce-Yii:
Addressed the maintainer review (flyingsquirrel0419) in commit `23a42459`:

**1. Negative `expires_in` (blocker):** all four sites now require `expires_in >= 0` — the three fallback paths (account-store / anthropic / chatgpt) treat negative as malformed and use the 3600s default; Kimi rejects `< 0` via the required-fields contract.

**2. Exact fallback window in tests:** the fallback tests now assert the actual window, not just `finite && future`:
- chatgpt / account-store: ~3600s (tolerance 30s)
- anthropic: ~3300s (3600s minus the 5-minute refresh skew)
- kimi: negative rejected alongside non-finite and overflowing

**3. Regression coverage:** +4 negative-value tests (one per site), 55/55 targeted, OAuth+anthropic+account-store suite 414/414, `tsc --noEmit` clean. Head `23a42459`.

The `unsponsored_surface` hygiene failure is the expected auth-boundary gate — ready for `maintainer-sponsored` after exact-head checks are green.

Ingwannu (approved):
Approved on exact head `355b69e5b003edebe3f543d1818a9618fa075b14` after an explicit authentication-boundary review.

The four token-adoption paths now validate the raw duration as a non-negative finite number and validate the computed timestamp before storage or return. Anthropic, ChatGPT, and the Codex account store take the bounded one-hour fallback; Kimi rejects malformed input. The negative, overflow, non-finite, and fallback-window regressions cover the failure modes without logging or widening credential destinations.

Exact-head evidence:

- focused OAuth/account-store tests: 414/414 pass;
- typecheck and privacy review: pass;
- CodeRabbit: pass, both inline blockers resolved on this head;
- cross-platform CI rerun: Linux shards, macOS, keyring, storage policy, and package installs all pass;
- current-`dev` distance: two documentation commits, within policy.

The earlier `hygiene` failure is the superseded pre-sponsorship run; the later hygiene run and current cross-platform gate are green. The author still needs to tick the now-proven local/CI checklist item so the automated readiness gate can move the PR out of draft. This approval does not authorize self-merge; authentication changes should still receive the additional maintainer review requested on the PR when practical.

Wibias:
Thanks @Bruce-Yii — merged.

This closes an important gap left after #1369: malformed `expires_in` values from OAuth token responses were still slipping through at several parse sites and producing `NaN`/`Infinity`/already-past expiries in stored credentials. That breaks the refresh loop (`NaN <= x` is always false; `Infinity` never expires), so the fix is high leverage for credential reliability.

Useful aspects of this PR:
- Guards all four remaining upstream parse paths (Anthropic, Kimi, ChatGPT, Codex account-store) with the same contract: finite, non-negative `expires_in`, plus a check on the computed expiry timestamp so overflow can't sneak past.
- Falls back to a refreshable 3600s default rather than persisting garbage.
- Targeted tests pin the exact fallback windows and rejection behavior per provider.

Appreciate the careful follow-through on overflow and negative values from review — that made the fix complete rather than partial.

PR #1425, opened 2026-08-10
fix(codex): honor split SQLite state homes

### LINKED ISSUE #1401: [Bug]: History sync ignores CODEX_SQLITE_HOME in Windows Desktop + WSL split-home layout
### Client or integration

Codex App

### Area

Platform (Windows / Linux / WSL); Codex history integration

### Summary

OpenCodex 2.11.1 ignores Codex's split SQLite location in the Windows Codex Desktop + WSL app-server layout.

Codex runs with a Windows-mounted main home while keeping SQLite-backed state on the WSL-native filesystem:

```text
CODEX_HOME=/mnt/c/Users/<windows-user>/.codex
CODEX_SQLITE_HOME=/home/<wsl-user>/.codex/sqlite
```

The active `state_5.sqlite` is under `CODEX_SQLITE_HOME`, but OpenCodex history code derives it as `$CODEX_HOME/state_5.sqlite`. History sync/restore can therefore inspect or mutate an obsolete, empty, or unrelated database.

Expected: all state-database consumers should use one resolver with Codex-compatible precedence: `sqlite_home` config, then `CODEX_SQLITE_HOME`, then effective `CODEX_HOME`.

### Reproduction

1. Run Windows Codex Desktop with the agent in WSL2 and this split layout:

   ```bash
   export CODEX_HOME=/mnt/c/Users/<windows-user>/.codex
   export CODEX_SQLITE_HOME=/home/<wsl-user>/.codex/sqlite
   ```
2. Confirm `$CODEX_SQLITE_HOME/state_5.sqlite` contains the current `threads` table.
3. Leave an empty or stale `$HOME/.codex/state_5.sqlite` or `$CODEX_HOME/state_5.sqlite`.
4. Run an OpenCodex history operation such as restore/uninstall while `syncResumeHistory` is enabled.
5. Observe that OpenCodex selects the stale database rather than the active database and fails on the missing schema.

Current code evidence:

```ts
const STATE_DB_PATH = join(CODEX_HOME, "state_5.sqlite");
```

A source search in 2.11.1 finds no handling for `CODEX_SQLITE_HOME` or the `sqlite_home` Codex option. The Linux service builder also does not propagate an explicitly supplied `CODEX_SQLITE_HOME`.

The public Codex environment reference documents the split state location and precedence:  <br>[https://learn.chatgpt.com/docs/config-file/environment-variables](<https://learn.chatgpt.com/docs/config-file/environment-variables>)

### Version

2.11.1

### Operating system

Windows 11 with WSL2 / Ubuntu; Codex Desktop agent mode set to WSL

### Logs or error output

```text
no such table: threads
```

Observed paths:

```text
/mnt/c/Users/<windows-user>/.codex/config.toml
/home/<wsl-user>/.codex/sqlite/state_5.sqlite   # active
/home/<wsl-user>/.codex/state_5.sqlite          # obsolete, 0 bytes
```

### Redacted configuration

```json
{
  "CODEX_HOME": "/mnt/c/Users/<windows-user>/.codex",
  "CODEX_SQLITE_HOME": "/home/<wsl-user>/.codex/sqlite",
  "syncResumeHistory": true
}
```

### Checks

### Safety and relationship

`syncResumeHistory: false` avoids the unsafe history path until fixed, at the cost of disabling history remapping.

Follow-up to [#63](<https://github.com/lidge-jun/opencodex/issues/63>). Companion service ownership issue: lidge-jun/opencodex#1400.

Ingwannu:
## Summary

- resolve Codex SQLite state with Codex-compatible precedence: root `sqlite_home` in the effective `config.toml`, then `CODEX_SQLITE_HOME`, then the effective `CODEX_HOME`
- make history provider defaults, history jobs, write admission, and native-residue checks address the same call-time `state_5.sqlite`
- fail closed when the authoritative config is unreadable, malformed, or contains a present but blank/non-string `sqlite_home`; only an absent file or absent key may use the environment/home fallback
- preserve the total read-only native-residue contract by converting an indeterminate SQLite authority into a `config` indeterminate result instead of throwing or selecting another database
- preserve an explicit SQLite home in systemd, launchd, Task Scheduler, and WinSW service artifacts, including Windows drive/UNC paths rendered from non-Windows repair tooling
- document the split-home boundary, backup identity, storage-cleanup non-authority, and decision rationale

### Root cause

OpenCodex rebuilt `CODEX_HOME/state_5.sqlite` independently in several paths. Codex can place SQLite thread state in another root, so a Windows Desktop + WSL installation could make the history writer, admission snapshot, and residue check reason about a database different from the one Codex actually opened.

The initial resolver also used the tolerant root-string reader. That made a present invalid `sqlite_home` indistinguishable from an absent key and allowed an unsafe fallback. The strict resolver now distinguishes absent, valid, and invalid states without changing tolerant config readers used by unrelated injection/catalog paths.

The fix centralizes one call-time resolver and passes the selected database identity through the existing history job/worker boundary. It deliberately does not give storage cleanup authority over an external SQLite root.

### User impact

History migration and provider synchronization now follow the active Codex SQLite database in split-home installations. Background services retain the explicit install-time SQLite destination instead of silently falling back to `CODEX_HOME`. Invalid or unreadable SQLite authority stops the operation rather than risking inspection or mutation of a stale database.

## Verification

- exact-head focused SQLite/history/admission/residue/service coverage: 198 passed, 0 failed
- `bun run typecheck`: passed
- docs-site `bun install --frozen-lockfile` and production build: 221 pages
- CPU-limited full-suite attempt before the residue-boundary follow-up: 10,929 passed, 11 skipped, 4 failed; the three failures caused by the new resolver were fixed and rerun green in the 198-test exact-head set, while the remaining `codex-shim` environment-token failure reproduces on clean `dev`
- `git diff --check`: passed
- refreshed GitHub CI on exact head `ea2895046`: pending

## Checklist

Closes #1401

## Summary by CodeRabbit

* **New Features**
  * Added support for resolving Codex history data from a configured SQLite location.
  * Added precedence for configuration, environment settings, and the Codex home directory.
  * Preserved configured SQLite locations across macOS, Windows, and Linux service installations.
  * Added support for relative and Windows-specific paths.

* **Bug Fixes**
  * Invalid or unreadable SQLite configuration now fails safely without selecting an unintended database.

* **Documentation**
  * Documented SQLite history location resolution, fallback behavior, and retention boundaries.

### DIFF
--- a/docs-site/src/content/docs/guides/codex-integration.md
+++ b/docs-site/src/content/docs/guides/codex-integration.md
@@ -159,6 +159,9 @@ the same precedence as Codex: root `sqlite_home` in `config.toml`, then `CODEX_S
 effective `CODEX_HOME`. Relative SQLite homes resolve from the current working directory. When an
 explicit `CODEX_SQLITE_HOME` is present during service installation or repair, the durable launcher
 stores its install-time absolute path so the background proxy continues to address the same database.
+If `config.toml` is absent, OpenCodex continues to the environment/home fallback. If the file exists
+but cannot be read, SQLite-home resolution stops instead of risking history operations against a
+different database.
 
 On Windows, an Orca shell can set both `CODEX_HOME` and `ORCA_CODEX_HOME` to Orca's bundled runtime
 home while the ChatGPT/Codex app still reads `%USERPROFILE%\\.codex`. `ocx status` and `ocx doctor`
--- a/docs-site/src/content/docs/guides/codex-integration.md
+++ b/docs-site/src/content/docs/guides/codex-integration.md
@@ -159,9 +159,10 @@ the same precedence as Codex: root `sqlite_home` in `config.toml`, then `CODEX_S
 effective `CODEX_HOME`. Relative SQLite homes resolve from the current working directory. When an
 explicit `CODEX_SQLITE_HOME` is present during service installation or repair, the durable launcher
 stores its install-time absolute path so the background proxy continues to address the same database.
-If `config.toml` is absent, OpenCodex continues to the environment/home fallback. If the file exists
-but cannot be read, SQLite-home resolution stops instead of risking history operations against a
-different database.
+If `config.toml` or its root `sqlite_home` key is absent, OpenCodex continues to the
+environment/home fallback. If the file cannot be read or parsed, or the key is present but blank or
+not a string, SQLite-home resolution stops instead of risking history operations against a different
+database.
 
 On Windows, an Orca shell can set both `CODEX_HOME` and `ORCA_CODEX_HOME` to Orca's bundled runtime
 home while the ChatGPT/Codex app still reads `%USERPROFILE%\\.codex`. `ocx status` and `ocx doctor`
--- a/docs-site/src/content/docs/guides/codex-integration.md
+++ b/docs-site/src/content/docs/guides/codex-integration.md
@@ -154,6 +154,12 @@ checks for a single Windows Codex Desktop home at `/mnt/c/Users/*/.codex/config.
 one candidate exists, it uses that directory so WSL app-server mode and Windows Codex Desktop share
 the same config and auth files. Set `CODEX_HOME` explicitly to override this detection.
 
+Codex can keep SQLite-backed thread state in a separate directory. OpenCodex history operations use
+the same precedence as Codex: root `sqlite_home` in `config.toml`, then `CODEX_SQLITE_HOME`, then the
+effective `CODEX_HOME`. Relative SQLite homes resolve from the current working directory. When an
+explicit `CODEX_SQLITE_HOME` is present during service installation or repair, the durable launcher
+stores its install-time absolute path so the background proxy continues to address the same database.
+
 On Windows, an Orca shell can set both `CODEX_HOME` and `ORCA_CODEX_HOME` to Orca's bundled runtime
 home while the ChatGPT/Codex app still reads `%USERPROFILE%\\.codex`. `ocx status` and `ocx doctor`
 warn about this exact mismatch and print redacted target paths. If a background service was installed
--- a/src/codex/admission.ts
+++ b/src/codex/admission.ts
@@ -33,6 +33,7 @@ import {
   CODEX_PROFILE_PATH,
   DEFAULT_CATALOG_PATH,
   getCodexHome,
+  resolveCodexStateDbPath,
 } from "./paths";
 
 export type CodexAdmission =
@@ -162,7 +163,7 @@ export function admitCodexWrite(deps: AdmissionDeps = {}): CodexAdmission {
   const config = diagnostics.config;
   const opencodexHome = getConfigDir();
   const integrationRecord = join(opencodexHome, "integrations", "codex.json");
-  const historyDb = join(codexHome, "state_5.sqlite");
+  const historyDb = resolveCodexStateDbPath({ codexHome });
 
   const canonicalTargets = {
     codexHome,
--- a/src/codex/history-job.ts
+++ b/src/codex/history-job.ts
@@ -18,7 +18,6 @@
  */
 import { randomUUID } from "node:crypto";
 import { homedir } from "node:os";
-import { join } from "node:path";
 
 import type {
   CodexHistoryWorkerOperation,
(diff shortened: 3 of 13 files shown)


Wibias (changes requested):
Request changes based on a full review of the current head.

Merge blocker:

`resolveCodexSqliteHome()` treats every `config.toml` read/parse failure as if no authoritative `sqlite_home` existed, then falls through to `CODEX_SQLITE_HOME` or `CODEX_HOME`. That is unsafe for the exact data-integrity problem this PR is fixing. `ENOENT` means the config is absent and fallback is valid; `EACCES`, `EIO`, `ENOTDIR`, or another read failure means OpenCodex does not know which database Codex selected. Falling back in that state can make history/admission/native-residue operations inspect or mutate a stale or unrelated `state_5.sqlite`.

Please fail closed for unreadable/indeterminate authoritative config while allowing the genuine-missing (`ENOENT`) case to fall through. Add regressions for at least:

- missing config -> environment fallback is allowed;
- unreadable config (for example EACCES) -> SQLite-home resolution fails closed and no fallback DB is selected.

The overall resolver centralization and service propagation look sound; I did not find a conventional credential or privilege-escalation issue. After the fix, refresh onto current `dev` and rerun exact-head CI.

Ingwannu:
Addressed the fail-closed blocker on the current head `3abca61fc` and rebased the branch onto `dev` `7779c0574`.

- only a genuine missing `config.toml` (`ENOENT`) may fall through to `CODEX_SQLITE_HOME` / `CODEX_HOME`;
- unreadable or indeterminate authoritative config now stops SQLite-home resolution instead of selecting a potentially unrelated database;
- focused regressions cover both the missing-config fallback and `EACCES` fail-closed path;
- the architecture and user integration docs now state this boundary.

Verification on the rebased head:

- `bun test tests/codex-sqlite-home.test.ts tests/service.test.ts tests/winsw.test.ts` — 130 passed, 0 failed
- docs-site production build — 221 pages
- `git diff --check` — passed

@Wibias please re-review the exact head after CI completes.

Wibias (changes requested):
Request changes after re-review of the current head `3abca61fc34059d83fb9d7861c2cee3a16bffd48`.

The previous unreadable-config blocker is fixed: genuine `ENOENT` may fall through, while read failures such as `EACCES` now fail closed. There is still one remaining data-integrity edge in the same contract.

`resolveCodexSqliteHome()` uses `readRootTomlString()` to discover `sqlite_home`. That helper only recognises a quoted string assignment and returns `null` otherwise. As a result, an existing authoritative config such as:

```toml
sqlite_home = 123
```

or a malformed `sqlite_home` assignment is indistinguishable from the key being absent. `resolveCodexSqliteHome()` then falls through to `CODEX_SQLITE_HOME` or `CODEX_HOME`, potentially selecting a different `state_5.sqlite` even though the config exists but its authoritative SQLite setting cannot be established.

For history mutation/admission/residue safety this should remain fail-closed. Please distinguish these states:

- `sqlite_home` absent -> environment/home fallback allowed;
- valid string -> use it;
- `sqlite_home` present but malformed/wrong type, or otherwise indeterminate -> fail closed and do not select a fallback database.

Add a regression proving at minimum that `sqlite_home = 123` with `CODEX_SQLITE_HOME` set throws instead of returning the environment database. Malformed-string coverage would also be useful.

After that fix, refresh onto current `dev` and rerun complete exact-head CI.

Ingwannu:
Addressed the remaining SQLite-authority blocker and the full-suite residue regression on exact head `ea2895046` (rebased on current `dev` `9f545f204`).

- SQLite authority now uses a strict TOML parse scoped only to `resolveCodexSqliteHome()`: absent file/key may fall back, a valid non-empty string is used, and malformed/wrong-typed/blank values fail closed.
- Regressions cover numeric, array, blank, malformed, and valid quoted-key forms with an environment fallback present.
- The tolerant root-string reader used by injection/catalog paths is unchanged.
- The full suite exposed that native-residue classification let the new fail-closed exception escape. That read-only boundary now converts the condition to `config` indeterminate, so it still refuses coordination without throwing or selecting a fallback database. A dedicated regression covers invalid `sqlite_home` there.

Verification on the final code:

- `bun test tests/codex-native-residue.test.ts tests/codex-sqlite-home.test.ts tests/service.test.ts tests/winsw.test.ts` — 198 passed, 0 failed
- docs production build — 221 pages
- `git diff --check` — passed

The CPU-limited full-suite attempt reached 10,929 passed / 11 skipped. Its three resolver-related failures are the residue cases fixed and rerun green above; the one remaining `codex-shim` token-inheritance failure is the same host-environment failure reproduced on clean `dev`. Fresh exact-head GitHub CI is now running.

@Wibias please re-review this exact head after CI completes.

Wibias (approved):
Approving. Follow-ups on this head fail closed for unreadable/invalid `sqlite_home`, keep Codex-compatible SQLite home precedence, and preserve the read-only native-residue indeterminate boundary instead of selecting another database.

Wibias:
Thanks @Ingwannu — this was useful because split-home Codex installs (especially Windows Desktop + WSL) were letting history, admission, and residue checks talk to a different `state_5.sqlite` than Codex itself. Centralizing one call-time resolver with fail-closed invalid/unreadable authority, and pinning that identity through services and history jobs, stops silent fallback to the wrong database.

Merging now.

PR #1427, opened 2026-08-10
fix(service): align WSL Codex home ownership

### LINKED ISSUE #1400: [Bug]: WSL service state records ~/.codex instead of the runtime-detected Windows Codex home
### Client or integration

Codex App

### Area

Service lifecycle; Platform (Windows / Linux / WSL)

### Summary

In the Windows Codex Desktop + WSL app-server layout supported by [#63](<https://github.com/lidge-jun/opencodex/issues/63>), OpenCodex uses two different Codex-home resolvers.

Runtime/config paths use the WSL-aware resolver and discover the Windows Desktop home. Service install state instead uses `CODEX_HOME || homedir()/.codex`. A service installed from an ordinary WSL shell therefore records the Linux fallback in `service-state.json`, even though the proxy runtime resolves the Windows Codex home.

Later `service repair` or `service uninstall` commands run from the Desktop-injected environment report a false ownership mismatch or direct restore logic toward the wrong home.

Expected: service ownership metadata should use the same canonical resolver as the proxy runtime, without requiring `CODEX_HOME` to be baked into the systemd environment.

### Reproduction

1. Use Windows Codex Desktop with the agent running in WSL2.
2. Ensure Linux `~/.codex/config.toml` is absent and exactly one Windows Codex config is visible at `/mnt/c/Users/<windows-user>/.codex/config.toml`.
3. Run:

   ```bash
   env -u CODEX_HOME -u OPENCODEX_HOME ocx status
   ```

   OpenCodex correctly reports `/mnt/c/Users/<windows-user>/.codex`.
4. Install the Linux user service from the same ordinary WSL environment:

   ```bash
   env -u CODEX_HOME -u OPENCODEX_HOME ocx service install
   ```
5. Inspect `~/.opencodex/service-state.json`. Its `codexHome` is `/home/<wsl-user>/.codex`.
6. From a Codex Desktop WSL task, where `CODEX_HOME=/mnt/c/Users/<windows-user>/.codex`, run `ocx service repair`.
7. Observe the ownership mismatch between the recorded Linux home and the Desktop home.

Code path:

```ts
function currentCodexHome(): string {
  const raw = process.env.CODEX_HOME?.trim();
  return raw ? resolve(expandUserPath(raw)) : join(homedir(), ".codex");
}
```

This bypasses the WSL-aware resolver already used by normal runtime paths.

### Version

2.11.1

### Operating system

Windows 11 with WSL2 / Ubuntu; Codex Desktop agent mode set to WSL

### Logs or error output

```text
Service was installed with CODEX_HOME=/home/<wsl-user>/.codex,
but current CODEX_HOME=/mnt/c/Users/<windows-user>/.codex.
```

### Redacted configuration

```json
{
  "desktopAgentMode": "WSL",
  "linuxConfigPresent": false,
  "windowsCodexHomeCandidates": 1
}
```

### Checks

### Relationship

Follow-up to [#63](<https://github.com/lidge-jun/opencodex/issues/63>). Companion split-SQLite issue: lidge-jun/opencodex#1401.

Ingwannu:
## Summary

- resolve service install-state ownership through the same Codex-home resolver used by the runtime
- preserve explicit `CODEX_HOME` authority while allowing the existing narrow WSL single-Windows-home discovery when it is unset
- keep ambiguous or genuinely foreign service state fail-closed and document the ownership decision

## Root cause

The runtime already used the WSL-aware `resolveCodexHomeDir()`, but service ownership independently recorded `CODEX_HOME || ~/.codex`. In a Windows Desktop + WSL layout with one discoverable Windows Codex home, the service could therefore record Linux `~/.codex` while the proxy actually used the Windows home. Later repair or uninstall then looked foreign even in the same environment.

## User impact

New service installs and same-environment ownership checks now agree with runtime Codex-home targeting. The patch does not migrate or take over existing foreign service state.

## Verification

- `bun test tests/codex-home-wsl.test.ts tests/service.test.ts`: 106 passed, 0 failed
- focused WSL regression rerun: 5 passed, 0 failed
- `bun run typecheck`: passed
- `git diff --check`: passed
- full-suite rerun was not repeated after the same clean-`origin/dev` host showed the documented service-token inheritance failure and the CPU-limited suite showed unrelated endpoint admission timeouts; exact-head GitHub CI remains required before readiness

Closes #1400

## Summary by CodeRabbit

* **Bug Fixes**
  * Service ownership now consistently uses the resolved Codex home, including Windows Codex locations when running under WSL.
  * Explicit Codex home paths remain honored without being automatically migrated.

* **Documentation**
  * Added guidance and decision notes describing Codex home resolution and service ownership behavior.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -285,11 +285,12 @@ export function serviceEnvironmentOwnedHere(): boolean {
 export function assertServiceEnvironmentMatchesInstall(): void {
   const state = readServiceInstallState();
   if (!state) return;
+  const actualCodexHome = currentCodexHome();
   const expected = normalizePathForCompare(state.codexHome);
-  const actual = normalizePathForCompare(currentCodexHome());
+  const actual = normalizePathForCompare(actualCodexHome);
   if (expected !== actual) {
     throw new ServiceOwnershipError(
-      `Service was installed with CODEX_HOME=${state.codexHome}, but current CODEX_HOME=${currentCodexHome()}. ` +
+      `Service was installed with CODEX_HOME=${state.codexHome}, but current CODEX_HOME=${actualCodexHome}. ` +
         "Run the service command from the same Codex home so native Codex restore updates the correct config.",
     );
   }
--- a/src/service.ts
+++ b/src/service.ts
@@ -14,7 +14,7 @@ import { expandUserPath, getConfigDir, readPid, removePid, removeRuntimePort, ve
 import { loadConfig } from "./config";
 import { restoreNativeCodex, restoreNativeCodexAsync } from "./codex/inject";
 import { stripGrokConfig } from "./grok/inject";
-import { isWslRuntime } from "./codex/home";
+import { isWslRuntime, resolveCodexHomeDir, type CodexHomeDeps } from "./codex/home";
 import { BUN_RUNTIME_PATH_ENV, BUN_RUNTIME_SOURCE_ENV, durableBunRuntime } from "./lib/bun-runtime";
 import type { BunRuntimeSource } from "./lib/bun-runtime";
 import { isProcessAlive, stopProxy } from "./lib/process-control";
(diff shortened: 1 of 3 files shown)


Wibias:
Thanks @Ingwannu — merged.

This closes #1400 with a small but important alignment fix. The runtime already resolved Codex home through the WSL-aware `resolveCodexHomeDir()`, but service install-state still recorded `CODEX_HOME || ~/.codex` independently. On Windows Desktop + WSL with a single discoverable Windows Codex home, that meant the service could record Linux `~/.codex` while the proxy actually used the Windows home — so later repair/uninstall looked "foreign" even in the same environment.

Routing service ownership through the same resolver keeps explicit `CODEX_HOME` authority, preserves the narrow WSL discovery path when unset, and stays fail-closed for genuinely foreign state. The regression test and structure note make the ownership contract hard to regress.

Appreciate the focused scope and the follow-up fix to resolve the Codex home once per ownership check.

PR #1434, opened 2026-08-10
fix(google-antigravity): persist thought-signature replay across restarts

### LINKED ISSUE #1429: [Bug]: google-antigravity tool continuation 400s once the in-process thought-signature replay cache misses
### Client or integration

Direct HTTP/API client

### Area

Provider adapter

### Summary

Opening this per @Wibias's request on lidge-jun/opencodex#1254, which was closed as fixed by lidge-jun/opencodex#1266 (Vertex). This is the `google-antigravity` side, and it is a different failure than the one lidge-jun/opencodex#1266 addressed.

A tool-call continuation on `google-antigravity` fails with HTTP 400 whenever the in-process reasoning-replay cache does not hold the signature:

```
Provider error 400: Antigravity invalid request: Function call is missing a thought_signature
in functionCall parts. This is required for tools to work correctly, and missing
thought_signature may lead to degraded model performance.
```

The cache in `src/adapters/google-antigravity-replay.ts` is process-local, with `REPLAY_TTL_MS = 1h`, `ANTIGRAVITY_REPLAY_MAX_ENTRIES = 10_240`, and batch eviction. So the signature is lost on a proxy restart, after an hour, or under eviction pressure.

What makes it unrecoverable rather than merely inconvenient is the client surface. On `/v1/chat/completions` there is no field where a caller can carry a thought signature back: the OpenAI tool-call schema has no equivalent of `thoughtSignature`. `messagesToGeminiFormat` will attach one from `tc.thoughtSignature` when present, but an OpenAI-format client has no way to populate it. When the cache misses, the conversation is simply stuck — every continuation of that tool call 400s, and re-sending the same history cannot help.

The Codex `/v1/responses` path is not affected in the same way, since the signature travels in the payload there.

Expected: a tool-call continuation on `google-antigravity` should keep working across a proxy restart, or, if that cannot be guaranteed, fail in a way the caller can act on rather than a provider 400 the client has no means to satisfy.

Not claiming which fix is right — persisting the cache, exposing the signature through an extension field on the OpenAI surface, and re-deriving the turn are all plausible and have different costs. Reporting the gap and the repro.

### Reproduction

Deterministic on 2.12.0. The proxy restart between the two turns is the whole point — it empties the replay cache.

1. Configure a `google-antigravity` provider (`adapter: google`, `googleMode: cloud-code-assist`, `authMode: oauth`).
2. **Turn 1** — force a tool call and keep the response:

```bash
curl -s -X POST http://127.0.0.1:10100/v1/chat/completions \
  -d '{
    "model": "google-antigravity/gemini-3.6-flash",
    "messages": [{"role":"user","content":"What is the weather in Lisbon? Use the tool."}],
    "tools": [{"type":"function","function":{"name":"get_weather","description":"Get the current weather for a city","parameters":{"type":"object","properties":{"city":{"type":"string"}},"required":["city"]}}}],
    "tool_choice": "auto", "stream": false, "max_tokens": 512
  }'
```

Returns 200 with `tool_calls: [get_weather({"city":"Lisbon"})]`.

3. **Restart the proxy** — `ocx service stop && ocx service start`. This clears the in-process replay cache. A one-hour wait or enough eviction pressure reproduces the same state without a restart.
4. **Turn 2** — send the history back with the assistant tool call and the tool result:

```bash
curl -s -X POST http://127.0.0.1:10100/v1/chat/completions \
  -d '{
    "model": "google-antigravity/gemini-3.6-flash",
    "messages": [
      {"role":"user","content":"What is the weather in Lisbon? Use the tool."},
      {"role":"assistant","content":null,"tool_calls":[{"id":"<id from turn 1>","type":"function","function":{"name":"get_weather","arguments":"{\"city\":\"Lisbon\"}"}}]},
      {"role":"tool","tool_call_id":"<id from turn 1>","content":"{\"tempC\": 19, \"sky\": \"clear\"}"}
    ],
    "tools": [{"type":"function","function":{"name":"get_weather","description":"Get the current weather for a city","parameters":{"type":"object","properties":{"city":{"type":"string"}},"required":["city"]}}}],
    "stream": false, "max_tokens": 512
  }'
```

Observed: HTTP 400 with the `missing a thought_signature` error above.

Control, to show the failure is the cache and not the request shape: run turns 1 and 2 back to back **without** restarting. It returns 200 and completes normally — the replay cache supplies the signature. I ran that four times across `gemini-3.6-flash` and `claude-opus-4-6-thinking`; all succeeded.

One incidental finding from the same measurements, in case it saves someone else the detour: the endpoint accepts the camelCase `thoughtSignature` key. `google-antigravity-replay.ts` injects only that form and those requests are accepted, so emitting a snake_case `thought_signature` alongside it is not needed. I had been carrying a local patch that did exactly that, on the assumption it was required; it was not, and I have dropped it.

### Version

2.12.0

### Operating system

Windows 11 Home Single Language 26H2 (build 10.0.26200)

### Provider and model

`google-antigravity` / `gemini-3.6-flash` (also reproduced on `claude-opus-4-6-thinking`)

### Logs or error output

```shell
# turn 2, after a proxy restart
status: 400
{"error":{"message":"Provider error 400: Antigravity invalid request: Function call is missing a
thought_signature in functionCall parts. This is required for tools to work correctly, and missing
thought_signature may lead to degraded model performance. Additional data, function call
`default_api:get_weather` , position 2. Please refer to
https://ai.google.dev/gemini-api/docs/thought-signatures for more details.",
"type":"invalid_request_error","param":null,"code":"invalid_request_error"}}

# same two turns without the restart
status: 200
finish_reason: stop
content: "The current weather in Lisbon is clear with a temperature of 19°C."
```

Relevant constants in `src/adapters/google-antigravity-replay.ts` (v2.12.0):

```
REPLAY_TTL_MS = 60 * 60 * 1000            // 1h
ANTIGRAVITY_REPLAY_MAX_ENTRIES = 10_240
REPLAY_EVICT_BATCH = 128
```

Not attached — the reproduction above is self-contained.

### Redacted configuration

```json
{
  "providers": {
    "google-antigravity": {
      "adapter": "google",
      "baseUrl": "https://daily-cloudcode-pa.googleapis.com",
      "authMode": "oauth",
      "googleMode": "cloud-code-assist",
      "defaultModel": "gemini-3.6-flash"
    }
  }
}
```

### Checks

Yuxin-Qiao:
## Summary

- Persist the Google Antigravity (CCA / Vertex) thought-signature replay cache to disk so Gemini tool-call continuations survive a proxy restart. Closes #1429.
- Root cause: `src/adapters/google-antigravity-replay.ts` was process-local (1h TTL, 10,240 sessions, batch eviction). After `ocx service stop && ocx service start` (or TTL/eviction pressure), a `/v1/chat/completions` continuation 400s with `Function call is missing a thought_signature`, and the OpenAI-format surface has no field for the caller to echo the signature back — the conversation was stuck. The `/v1/responses` path was unaffected because the signature travels in the payload there.
- Fix: debounced atomic snapshot of the cache to `antigravity-replay.json` in the config directory (same durability pattern as `responses-state.json`), lazy bounded load on first cache access with expired sessions dropped at load, and a shutdown flush wired into `drainAndShutdown`. The CCA session id is deterministic (SHA-256 of the first user message text), so a restarted proxy re-derives the same key and re-injects the cached signatures.
- Claude-on-Antigravity is untouched: it uses inline signature sanitization, not the replay cache (`antigravityUsesReplayCache` is Gemini-only).
- Snapshot contents are bounded (24 MiB write cap, 32 MiB refuse-to-parse ceiling) and contain only 64-hex hashed keys (model+session, function identity) plus opaque upstream signature tokens — no raw request content, credentials, or model names. Writes go through the guarded `atomicWriteFileAsync` (0600 file, 0700 dir, real-home write guard under test).

## Verification

- `bun test tests/google-antigravity-replay.test.ts` — 53 pass, including restart simulation (observe -> flush -> reset -> apply re-injects), flush before debounce, load-time expiry and cap cleanup, corrupt/unknown-version refusal, oversized-file refusal, clear-on-invalid, duplicate-key collapse, UTF-8 byte accounting, and the review regressions below.
- Review regressions: load recomputes call sizes from the signature and ignores serialized `sizeBytes` (including multi-byte signatures whose UTF-8 size differs from JavaScript string length), snapshot never serializes `sizeBytes`, flush waits out a blocked writer and persists mutations that land during it (deterministic `afterTempWrite` seam), flush surfaces snapshot write failures for shutdown diagnostics, background snapshot failures log a redacted warning and keep the service running.
- `bun test tests/responses-state.test.ts tests/server-background-lifecycle.test.ts tests/shutdown-drain.test.ts tests/google-antigravity-wire.test.ts tests/google-antigravity-oauth.test.ts tests/state-store-sweeper.test.ts tests/app-owned-memory.test.ts tests/storage-mutation-race.test.ts` — 218 pass.
- `cd docs-site && bun install --frozen-lockfile && bun run build` — 221 pages built (English + ja/ko/ru/zh-cn updated).
- Full `bun run test`: 10,926 pass, 8 skipped, 20 baseline failures. Failures are confined to unrelated catalog/live-update/provider-management, crash-guard, and retained-root seam tests; none overlap with this change. The replay suite remains 53/53.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - Google Vertex and Antigravity reasoning continuity now survives proxy restarts through durable replay-state snapshots.
  - Expired, invalid, duplicate, or oversized cached data is safely ignored.

- **Bug Fixes**
  - Shutdown now flushes pending replay state, reducing the risk of losing continuation data.
  - Cache updates remain resilient when background persistence encounters errors.

- **Documentation**
  - Updated adapter documentation in English, Japanese, Korean, Russian, and Simplified Chinese to describe persisted `thoughtSignature` snapshots.


Wibias (changes requested):
Request changes based on a full bugs/edge-cases/security review of the current head.

Merge blockers:

1. Snapshot loading trusts serialized `sizeBytes` instead of deriving it from the persisted signature. A corrupt/tampered snapshot can store a very large `signature` with `sizeBytes: 0/1`, bypassing `maxSignatureBytes`, per-session byte limits, replay byte accounting, and app-owned memory accounting. Recompute the authoritative UTF-8 byte length from `signature` during load, enforce the limits against that value, and use it for bookkeeping. Prefer not serializing derived `sizeBytes` at all. Add a regression with a large signature plus falsified small `sizeBytes`.

2. `flushAntigravityReplay()` is not guaranteed to include the final mutation during shutdown. If a writer is already in flight and no debounce timer exists when flush starts, flush only awaits the current gate. A mutation during that write can schedule a new debounce timer; the first writer completes, flush returns, and shutdown can exit before the newly scheduled write runs. Use a dirty/mutation generation or loop until the persisted generation catches the current mutation generation and no writer/timer remains. Add a deterministic blocked-writer test with a second mutation during the first write.

Required hardening:

3. Snapshot persistence errors are swallowed inside `persistReplaySnapshotNow()`. This also makes the lifecycle's `Promise.allSettled(...flushAntigravityReplay())` rejection handling ineffective for real disk/ACL/rename failures. The cache need not crash the service, but durability loss should at least produce a sanitized warning/metric or otherwise reach the shutdown diagnostic path.

Security-positive note: using guarded atomic writes and hardened local permissions is the right direction, and I did not find raw credentials/account ids in the snapshot.

I explicitly did NOT adopt the stale bot claims about a duplicate `const raw` declaration or cross-test snapshot leakage; those do not reproduce on this head.

Please fix the two durability/integrity blockers, then refresh onto current `dev` and obtain green exact-head CI.

Yuxin-Qiao:
Thanks for the detailed review. All three points are fixed on the refreshed head (rebased onto current `dev`), and the head gates are green on the exact head.

**1. Snapshot loading no longer trusts serialized `sizeBytes`**

`loadReplaySnapshotEntry` now derives the authoritative UTF-8 byte length from `signature` itself (`utf8.encode(call.signature).byteLength`), enforces `maxSignatureBytes` and the per-call/session limits against that value, and uses it for `ReplayCall.sizeBytes` and all byte accounting. The serialized `sizeBytes` field is ignored entirely, so a forged snapshot cannot shrink its footprint below the true signature size. The write path no longer serializes `sizeBytes` at all; old snapshots that still contain it load fine because the field is ignored.

Regression coverage in `tests/google-antigravity-replay.test.ts`: "load recomputes call sizes from the signature and ignores serialized sizeBytes" (70 KB signature with forged `sizeBytes: 1` is dropped; a valid signature with forged `sizeBytes: 1` is accounted at its true size), plus "snapshot never serializes sizeBytes" which locks the on-disk format.

**2. `flushAntigravityReplay()` now guarantees the final mutation is persisted**

I added a mutation generation (`replayMutationGeneration`, bumped by `markReplayDirty()` on every cache change) and a written generation that each successful snapshot write records. `flushAntigravityReplay()` loops until the written generation equals the current mutation generation and no debounce timer remains, so a mutation that lands while a writer is in flight is persisted before shutdown completes (this is stronger than the one-shot recheck `flushResponseState()` does, because a mutation can also land during the follow-up write).

Deterministic regression: "flush waits out a blocked writer and persists mutations that land during it" blocks the first snapshot write via the `afterTempWrite` seam (same pattern as `AtomicWriteAsyncTestSeam`), mutates the cache while the write is blocked, releases it, and asserts both sessions are present in the final snapshot.

**3. Persistence failures are no longer swallowed**

The debounced background path now logs a sanitized static warning on failure (no error payload, no paths). `persistReplaySnapshotNow()` propagates failures, so `flushAntigravityReplay()` rejects and the existing lifecycle shutdown diagnostic (`[antigravity] replay flush during shutdown failed`) is actually reachable, without crashing the service.

Coverage: "flush surfaces snapshot write failures for shutdown diagnostics" and "background snapshot failures log a redacted warning and keep the service running".

**Verification on the new head:** replay suite 51/51, adjacent suites 168/168 across `responses-state`, `server-background-lifecycle`, and `shutdown-drain`, `bun run typecheck` clean, `bun run privacy:scan` passed, docs build 221 pages. I also updated the `adapters.md` wording (all five locales) to describe `thoughtSignature` values as opaque, addressing the CodeRabbit docs note; the other CodeRabbit threads you explicitly did not adopt were resolved without changes.

Yuxin-Qiao:
Follow-up on the refreshed head (`244ade83`, rebased onto current `dev`): beyond the three review points above, the follow-up review cycle hardened the same module further, all covered by regression tests:

- `flushAntigravityReplay()` retry loop is bounded (8 attempts) so a mutation stream cannot hang shutdown; the last completed write is already on disk.
- Load-time cleanups (expired/over-limit sessions, duplicate call keys) are persisted back to the snapshot, so stale entries do not survive on disk across restarts.
- Snapshot byte-cap ordering now uses a persisted session activity timestamp (observe/apply both refresh it) instead of reversed insertion order.
- The failure test now waits on a deterministic write-failure signal and asserts the warning is redacted (error payload with a fake secret never reaches the log).

Head gates are green on the exact head; CodeRabbit completed with no actionable comments. Cross-platform CI and React Doctor are still awaiting maintainer approval for the fork run.

Ingwannu (changes requested):
The persistence direction is valid, and the earlier size-accounting, in-flight mutation, and error-propagation blockers have been substantially addressed. I am still requesting changes on current head `244ade8393e4b201077ad5d02b3ba536b8881231` because two durability contracts remain incomplete:

1. `persistReplaySnapshotNow()` enforces `replaySnapshotMaxBytes` against only the sum of `JSON.stringify(persistEntry)` values. The written document also contains the version/sessions wrapper, brackets, and inter-entry commas, so the final UTF-8 file can exceed the declared hard cap. Account for the exact complete document size before admitting each session (or construct an exactly bounded document), and add a boundary regression that asserts `Buffer.byteLength(the entire written snapshot) <= configured cap` with multiple small sessions.
2. `flushAntigravityReplay()` now bounds retries, but after eight non-converging attempts it returns successfully even when `replayMutationGeneration !== replayWrittenGeneration` or a timer is still pending. `drainAndShutdown()` therefore sees a fulfilled flush and emits no failure diagnostic although the latest signature was not made durable. Exhausting the retry budget must return an explicit failure (preferably a fixed, non-sensitive error) so the existing shutdown warning is reachable. Add a deterministic test that mutates during every attempted write and proves the flush fails rather than silently claiming success.

The branch is also two commits behind current `dev`, and this fork head currently has only the lightweight PR gates rather than exact-head platform/full-suite CI. After the two contracts are fixed, please refresh onto current `dev`, rerun the focused persistence/lifecycle suites plus typecheck/privacy, and obtain the full exact-head CI before merge.

Wibias (changes requested):
Request changes on current head `89a96ce0ebf6707278bf83a3077ff9b845eb8b0e`.

Two durability contracts are still incomplete:

1. `flushAntigravityReplay()` bounds its convergence loop, but exhausting `REPLAY_FLUSH_MAX_ATTEMPTS` currently falls through and resolves successfully even when `replayMutationGeneration !== replayWrittenGeneration` or a debounce timer remains pending. In that state `drainAndShutdown()` observes a successful flush even though the latest thought signature may not be durable. Keep the retry bound, but make budget exhaustion reject with a fixed, non-sensitive error so shutdown diagnostics remain truthful. Add a deterministic regression that forces a mutation during every attempted write and proves the flush rejects rather than silently succeeding.

2. Snapshot write-cap accounting only sums `Buffer.byteLength(JSON.stringify(persistEntry), "utf8")`. The actual persisted document also contains the `{ "version": ..., "sessions": [...] }` framing, array brackets, and commas between entries, so the final UTF-8 file can exceed the advertised `replaySnapshotMaxBytes` hard cap. Account for the exact serialized document size before admitting each session, and add a boundary regression that asserts `Buffer.byteLength(the complete written snapshot, "utf8") <= configured cap` with multiple entries.

The earlier integrity, in-flight mutation, error-propagation, load-cleanup, duplicate-key, UTF-8 accounting, and redacted-warning fixes look good. After these two remaining contracts are fixed, refresh onto current `dev` and rerun exact-head CI.

lidge-jun:

Your commits are on `dev` with authorship intact, plus `d065fd0ad` closing the two remaining durability blockers.

`flushAntigravityReplay()` bounded its convergence loop but fell out of it and *resolved* when the attempt budget was exhausted with the durable generation still behind the latest mutation. `drainAndShutdown()` then saw a fulfilled promise and emitted no warning, so shutdown reported a durability it could not demonstrate. Budget exhaustion now rejects, with fixed text since the shutdown path logs it and it must not carry session or signature detail.

The snapshot cap counted only per-entry JSON and ignored the `{"version":N,"sessions":[...]}` framing and the separators, so the written file exceeded `replaySnapshotMaxBytes` by a margin that grew with each session. Measured: a cap of 1122 bytes produced an 1151-byte file. Accounting now seeds with the exact prefix and suffix, charges one byte per separator, and builds the payload from the same fragments.

You reached equivalent fixes independently on your newer head — `REPLAY_FLUSH_INCOMPLETE_ERROR` and the per-entry-plus-comma accounting — which is why this closes as superseded rather than replaced.

Verified: `tests/google-antigravity-replay.test.ts` 55 pass / 0 fail and `tests/shutdown-drain.test.ts` 17 pass / 0 fail, with both regressions red before the fix. Thanks for the persistent iteration on this one.

PR #1435, opened 2026-08-10
fix(catalog): apply providerContextCaps.openai to native OpenAI rows

### LINKED ISSUE #1430: providerContextCaps.openai is ignored for native GPT-5.6 catalog rows
### Client or integration

Codex App

### Area

Catalog / models

### Summary

The supported per-provider context cap is persisted and reported as active for `openai`, but native GPT-5.6 catalog rows ignore it.

On OpenCodex 2.12.0, both `contextCapValue` and `providerContextCaps.openai` are set to `272000`. After restarting the managed proxy and running a full `opencodex sync`, GPT-5.6 Sol, Terra, and Luna are still generated with a 372,000-token context window and a 334,800-token auto-compaction threshold.

I expected the enabled OpenAI provider cap to constrain native GPT-5.6 models to 272,000 tokens, with auto-compaction at 244,800.

This is intentionally narrower than [#1051](<https://github.com/lidge-jun/opencodex/issues/1051>): regardless of whether 372k is the desired default, the documented/configurable provider cap currently has no effect on the native GPT-5.6 overrides.

### Reproduction

1. Install or update to OpenCodex 2.12.0.
2. Configure the global and OpenAI provider caps:

```json
{
  "contextCapValue": 272000,
  "providerContextCaps": {
    "openai": 272000
  }
}
```

The same result occurs when the OpenAI cap is enabled through the management API. Readback confirms `providerContextCaps.openai = 272000` and the API reports the cap without degraded mode.

3. Restart the managed OpenCodex service so the installed version is active.
4. Run a full sync:

```shell
opencodex sync
```

5. Verify the persisted configuration:

```shell
jq '{contextCapValue, providerContextCaps: {openai: .providerContextCaps.openai}}' \
  ~/.opencodex/config.json
```

Result:

```json
{
  "contextCapValue": 272000,
  "providerContextCaps": {
    "openai": 272000
  }
}
```

6. Inspect the generated catalog or runtime model metadata:

```shell
codex debug models | jq -r '
  .models[]
  | select(.slug | test("^gpt-5\\.6-(sol|terra|luna)$"))
  | [.slug, .context_window, .max_context_window, .auto_compact_token_limit]
  | @tsv
'
```

Actual result:

```text
gpt-5.6-sol    372000    372000    334800
gpt-5.6-terra  372000    372000    334800
gpt-5.6-luna   372000    372000    334800
```

Expected result while the OpenAI provider cap is enabled:

```text
gpt-5.6-sol    272000    272000    244800
gpt-5.6-terra  272000    272000    244800
gpt-5.6-luna   272000    272000    244800
```

The proxy is healthy and `opencodex doctor` confirms that the Codex app-server catalog is current with the on-disk catalog, which rules out a stale proxy or stale model cache.

### Suspected cause

The native GPT-5.6 rows appear to bypass the configured provider cap because two source-level overrides remain fixed at 372,000:

```text
src/codex/catalog/metadata.ts
NATIVE_GPT56_CONTEXT_WINDOW = 372_000

src/providers/registry.ts
OPENAI_CODEX_GPT56_CONTEXT_WINDOW = 372_000
```

Changing only those two values to `272_000`, restarting the managed service, and running `opencodex sync` immediately produces the expected `272000 / 272000 / 244800` values for Sol, Terra, and Luna. Spark remains unchanged at `100000 / 100000 / 90000`, confirming that the failure is scoped to the native GPT-5.6 override path.

### Suggested behavior

When `providerContextCaps.openai` is enabled, apply it as a ceiling to native OpenAI catalog rows as well as routed/provider-discovered models. The 372k native value can remain the default when no custom cap is enabled.

Related reports:

* [#1051](<https://github.com/lidge-jun/opencodex/issues/1051>) — original Codex App report; closed after 372k was described as intentional.
* lidge-jun/opencodex#1218 — related Claude Code metadata report; also confirms the upstream 272k value, though the maintainer kept 372k as the OpenCodex default.

### Version

2.12.0

### Operating system

macOS 27.0 (26A5388g), arm64

### Provider and model

OpenAI (Codex login) / `gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`

### Logs or error output

```shell
gpt-5.6-sol    372000    372000    334800
gpt-5.6-terra  372000    372000    334800
gpt-5.6-luna   372000    372000    334800
```

* [https://github.com/lidge-jun/opencodex/issues/1051](<https://github.com/lidge-jun/opencodex/issues/1051>)
* [https://github.com/lidge-jun/opencodex/issues/1218](<https://github.com/lidge-jun/opencodex/issues/1218>)
* [https://github.com/openai/codex/pull/33972](<https://github.com/openai/codex/pull/33972>)
* [https://github.com/openai/codex/pull/34009](<https://github.com/openai/codex/pull/34009>)

### Redacted configuration

```json
{
  "providers": {
    "openai": {
      "codexAccountMode": "pool",
      "authMode": "forward"
    }
  },
  "contextCapValue": 272000,
  "providerContextCaps": {
    "openai": 272000
  }
}
```

### Checks

Yuxin-Qiao:
## Summary

- `providerContextCaps.openai` is now a ceiling for native OpenAI catalog rows, the live Codex discovery endpoint, management model rows, and routing capability evidence.
- Hardcoded native context overrides are decoupled from cap application: overrides are applied first when present, then known `context_window` / `max_context_window` values are capped regardless of origin. Auto-compaction is recomputed from the capped window.
- Preserved native rows without a hardcoded override, including `gpt-5.4-mini`, now receive the same OpenAI cap. With no cap configured, the existing 372k native default is unchanged.

Closes #1430

## Verification

- `bun run typecheck` passes.
- `bun run privacy:scan` passes; `git diff --check` is clean.
- Focused suites pass: 235 tests, 0 fail across `tests/codex-catalog.test.ts`, `tests/claude-models-discovery.test.ts`, `tests/native-model-toggle.test.ts`, `tests/route-explainability.test.ts`, `tests/grok-sync.test.ts`, `tests/claude-context-windows.test.ts`, and `tests/claude-desktop-native-context.test.ts`.
- The endpoint regression exercises `GET /v1/models?client_version=1.0.0` and asserts native GPT-5.6 output of `272000 / 272000 / 244800`; catalog, preserved-row, runtime, and routing regressions remain covered.
- Rebased onto the current `dev` head `e2f7f2ba`; the pushed PR head is `e266f799`.

## Checklist

- [x] Docs or release notes were updated when needed (existing `providerContextCaps` documentation covers this behavior).

## Review readiness checklist

### DIFF
--- a/src/codex/catalog/metadata.ts
+++ b/src/codex/catalog/metadata.ts
@@ -107,11 +107,12 @@ const PINNED_NATIVE_CAPABILITY_ENTRIES: Map<string, RawEntry> = new Map(
     .map(m => [m.slug as string, m]),
 );
 
-export function nativeOpenAiContextWindow(slug: string): number | undefined {
-  return NATIVE_OPENAI_CONTEXT_OVERRIDES[slug]?.contextWindow
+export function nativeOpenAiContextWindow(slug: string, contextCap?: number): number | undefined {
+  const raw = NATIVE_OPENAI_CONTEXT_OVERRIDES[slug]?.contextWindow
     ?? (typeof PINNED_NATIVE_CAPABILITY_ENTRIES.get(slug)?.context_window === "number"
       ? PINNED_NATIVE_CAPABILITY_ENTRIES.get(slug)!.context_window as number
       : undefined);
+  return applyProviderContextCap(raw, contextCap) ?? raw;
 }
 
 export function nativeInputModalities(slug: string): string[] {
@@ -199,11 +200,12 @@ export function desktopVisibleNativeSlugs(config: Pick<OcxConfig, "claudeCode" |
   return visibleNativeSlugs(config);
 }
 
-export function nativeModelRows(config: Pick<OcxConfig, "disabledModels" | "combos">): Array<{ slug: string; disabled: boolean; contextWindow?: number }> {
+export function nativeModelRows(config: Pick<OcxConfig, "disabledModels" | "combos" | "providerContextCaps">): Array<{ slug: string; disabled: boolean; contextWindow?: number }> {
   const disabled = disabledNativeSlugs(config);
   const shadowed = configuredNativeAliasSlugs(config);
+  const openaiContextCap = providerContextCap(config, OPENAI_CODEX_PROVIDER_ID);
   return NATIVE_OPENAI_MODELS.filter(slug => !shadowed.has(slug)).map(slug => {
-    const contextWindow = nativeOpenAiContextWindow(slug);
+    const contextWindow = nativeOpenAiContextWindow(slug, openaiContextCap);
     return { slug, disabled: disabled.has(slug), ...(contextWindow !== undefined ? { contextWindow } : {}) };
   });
 }
--- a/src/codex/catalog/parsing.ts
+++ b/src/codex/catalog/parsing.ts
@@ -259,18 +259,34 @@ export function isNativeOpenAiEntry(entry: RawEntry): boolean {
   return typeof entry.slug === "string" && !entry.slug.includes("/");
 }
 
-export function applyNativeOpenAiContextOverride(entry: RawEntry): void {
+export function applyNativeOpenAiContextOverride(entry: RawEntry, contextCap?: number): void {
   const nativeSlug = trustedAccountBoundNativeCatalogSlug(entry)
     ?? (isNativeOpenAiEntry(entry) ? entry.slug as string : undefined);
   if (!nativeSlug) return;
   const override = NATIVE_OPENAI_CONTEXT_OVERRIDES[nativeSlug];
-  if (!override) return;
-  if (typeof override.contextWindow === "number") {
-    entry.context_window = override.contextWindow;
-    entry.auto_compact_token_limit = Math.floor(override.contextWindow * 0.9);
+  if (override) {
+    if (typeof override.contextWindow === "number") {
+      const contextWindow = applyProviderContextCap(override.contextWindow, contextCap) ?? override.contextWindow;
+      entry.context_window = contextWindow;
+      entry.auto_compact_token_limit = Math.floor(contextWindow * 0.9);
+    }
+    if (typeof override.maxContextWindow === "number") {
+      entry.max_context_window = applyProviderContextCap(override.maxContextWindow, contextCap) ?? override.maxContextWindow;
+    }
   }
-  if (typeof override.maxContextWindow === "number") {
-    entry.max_context_window = override.maxContextWindow;
+  // providerContextCaps.openai is a ceiling for native OpenAI rows regardless of where the
+  // advertised window came from (#1430): preserved rows without a hardcoded override (e.g.
+  // gpt-5.4-mini) must stay under the cap too, and auto-compaction follows the capped window.
+  const currentContext = typeof entry.context_window === "number" ? entry.context_window : undefined;
+  const cappedContext = applyProviderContextCap(currentContext, contextCap);
+  if (cappedContext !== currentContext && typeof cappedContext === "number") {
+    entry.context_window = cappedContext;
+    entry.auto_compact_token_limit = Math.floor(cappedContext * 0.9);
+  }
+  const currentMax = typeof entry.max_context_window === "number" ? entry.max_context_window : undefined;
+  const cappedMax = applyProviderContextCap(currentMax, contextCap);
+  if (cappedMax !== currentMax) {
+    entry.max_context_window = cappedMax;
   }
 }
 
--- a/src/codex/catalog/provider-fetch.ts
+++ b/src/codex/catalog/provider-fetch.ts
@@ -1639,6 +1639,7 @@ async function gatherRoutedModelsUncached(
     // configs that will never need it.
   } else {
     const disabled = disabledNativeSlugs(config);
+    const openaiContextCap = providerContextCap(config, OPENAI_CODEX_PROVIDER_ID);
     const requiredNativeComboTargets = new Set(listComboIds(config).flatMap(id => {
       const combo = getCombo(config, id);
       return combo?.targets.flatMap(target => (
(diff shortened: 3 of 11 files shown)


mkarolczyk:
I found one remaining Codex-facing path that still bypasses the OpenAI provider cap.

In `src/server/index.ts:914`, `GET /v1/models?client_version=…` calls `buildCatalogEntries()` without passing `providerContextCap(config, OPENAI_CODEX_PROVIDER_ID)`.

I reproduced this on the PR head with `providerContextCaps.openai = 272000`: the endpoint still returns `372000 / 372000 / 334800` for `gpt-5.6-sol`, instead of `272000 / 272000 / 244800`.

Please pass the cap through this call and add an integration test in `tests/claude-models-discovery.test.ts`. The current focused tests do not exercise the live Codex discovery endpoint with a provider cap.

Wibias (changes requested):
Request changes based on a full review of the current head.

Merge blocker:

The PR claims `providerContextCaps.openai` is now a ceiling for native OpenAI rows, but the persisted-catalog normalizer still returns early unless the native slug has an entry in `NATIVE_OPENAI_CONTEXT_OVERRIDES`:

`const override = NATIVE_OPENAI_CONTEXT_OVERRIDES[nativeSlug]; if (!override) return;`

That means supported native rows without a hardcoded override can still bypass the cap. `gpt-5.4-mini` is in `NATIVE_OPENAI_MODELS` but has no `NATIVE_OPENAI_CONTEXT_OVERRIDES` entry, so a genuine/preserved native `gpt-5.4-mini` row can remain uncapped even though runtime/synthetic metadata paths use `nativeOpenAiContextWindow(..., cap)` and are capped. This leaves cross-surface catalog/routing disagreement.

Please decouple hardcoded native overrides from provider-cap application: first apply an override if one exists, then apply the OpenAI provider cap to the resulting known `context_window` / `max_context_window` values regardless of where those values came from, and recompute auto-compaction from the capped context where applicable.

Add a regression for a preserved genuine `gpt-5.4-mini` native row with an OpenAI cap lower than its existing context window, verifying context/max-context/auto-compact all reflect the cap.

The GPT-5.6 paths added by this PR otherwise look correct and I did not find a security issue. This PR also overlaps newer `dev` work in central catalog files, so after the fix it should be refreshed onto current `dev` and rerun with exact-head CI.

Yuxin-Qiao:
Thanks for the detailed review. Both points are addressed on the refreshed head `16e6342b`, rebased onto current `dev` (`da8ebd31`).

**1. Hardcoded overrides and the provider cap are decoupled**

`applyNativeOpenAiContextOverride` in `src/codex/catalog/parsing.ts` no longer returns early when `NATIVE_OPENAI_CONTEXT_OVERRIDES[nativeSlug]` is absent. The flow is now:

- If an override exists, its `contextWindow` / `maxContextWindow` are applied first, each passed through `applyProviderContextCap`.
- Afterwards the cap is applied to the resulting `context_window` / `max_context_window` regardless of where those values came from (hardcoded override, preserved disk row, or pinned upstream snapshot).
- `auto_compact_token_limit` is recomputed as `floor(context_window * 0.9)` from the capped window. The equality guard means preserved rows are only rewritten when the cap actually lowers the advertised window, so behavior is unchanged when no cap is configured or the cap does not bind (a preserved `gpt-5.4-mini` row at 272k with no cap stays exactly as it was).

**2. Regression for preserved `gpt-5.4-mini`**

Added in `tests/codex-catalog.test.ts`: "preserved gpt-5.4-mini rows get the openai cap without a hardcoded override (#1430)". A preserved genuine `gpt-5.4-mini` row (272_000 context/max) goes through `mergeCatalogEntriesForSync` with `openaiContextCap = 200_000` and must come out with `context_window = 200_000`, `max_context_window = 200_000`, `auto_compact_token_limit = 180_000`.

**3. Refresh onto current dev and verification**

- Rebased onto `da8ebd31` (current `dev`), single squashed commit, no conflicts.
- `bun run typecheck`: clean.
- Focused suites (`codex-catalog`, `native-model-toggle`, `route-explainability`, `grok-sync`, `claude-context-windows`, `claude-desktop-native-context`): 226 pass / 0 fail, including the new regression.
- Full suite: the failing set was compared against a clean `dev` baseline on the same files; they are identical except one `crash-guard` breadcrumb test that passes on rerun (load flake during the full run). No failures come from this PR.
- `bun run privacy:scan` and `git diff --check`: clean.
- Gate checks on the exact head (`enforce-target`, `label`, `hygiene`, `resolve-pr`) are green. Cross-platform CI and React Doctor still need maintainer approval to run on this fork PR (they were `action_required` on the previous head for the same reason).

No security issue was introduced; the GPT-5.6 paths are untouched by this revision.

Ingwannu (dismissed):
The refreshed head fixes the earlier preserved-row blocker, but one merge blocker remains on exact head `16e6342b`.

`src/server/index.ts:914` serves the live Codex discovery route (`GET /v1/models?client_version=...`) by calling `buildCatalogEntries(...)` without the new OpenAI context-cap argument. With `providerContextCaps.openai = 272000`, that route can therefore still emit native GPT-5.6 rows at `372000 / 372000 / 334800`, while the synchronized catalog and routing evidence are capped. This leaves the user-visible runtime surface inconsistent with the paths this PR fixed.

Please pass `providerContextCap(config, OPENAI_CODEX_PROVIDER_ID)` through this call and add an endpoint-level regression in `tests/claude-models-discovery.test.ts` (or the nearest existing server discovery suite) that asserts the live response is `272000 / 272000 / 244800` for a native GPT-5.6 row.

The branch is also five commits behind current `dev` (`7779c057`), so refresh it and run exact-head cross-platform CI after the endpoint fix. The rest of the revised cap propagation, including the preserved `gpt-5.4-mini` case, looks sound.

Wibias (approved):
Approving. Cap application is decoupled from hardcoded native overrides, preserved rows without overrides are capped, live discovery matches catalog/routing metadata, and the no-cap 372k default is unchanged.

Wibias:
Thanks @Yuxin-Qiao — this was useful because `providerContextCaps.openai` was only lowering routed/discovered models while native OpenAI catalog rows (and live Codex discovery) kept the hardcoded 372k windows, so GPT-5.6 Sol/Terra/Luna could disagree across catalog, management rows, and routing. Threading one ceiling through builders, runtime accessors, capability evidence, and discovery keeps those surfaces consistent without changing the uncapped default.

Merging now.

PR #1450, opened 2026-08-11
docs(providers): table-safe Kiro install pipes in translations; fix live-test account read (follow-up to #1397)

Wibias:
Follow-up to #1397 (Nous Portal OAuth provider). This branch is #1397's head
(`de819e6c4`) plus one commit; merge after #1397 so the diff here collapses to
just the incremental fix.

## What
Two small fixes that could not be pushed to the PR #1397 fork head (maintainer edits were disabled):

1. **Table-safe Kiro install pipelines in translated provider guides** — the literal `|` characters inside the Kiro CLI install commands broke the markdown tables in the ja/ko/ru/zh-cn provider guides. Replaced with the table-safe `&#124;` entity, matching the English row content.
2. **Live-test account read fix** — `tests/nous-oauth-live.test.ts` called `getCredential("nous", accountId)` with an excess argument (a type error). It now reads the refreshed account with `getAccountCredential("nous", stored!.accountId!)`.

## Verification
- `bun run typecheck` ✅
- `bun run privacy:scan` ✅
- `bun test tests/nous-oauth-live.test.ts` ✅ (opt-in live test skipped as expected)

Intended to merge after #1397. Once #1397 lands on `dev`, this PR's diff
reduces to the single incremental commit above.

## Summary by CodeRabbit

- **New Features**
  - Improved Nous Portal OAuth sign-in, token refresh, and device authorization reliability.
  - Preserved newer credentials during concurrent refresh operations.

- **Bug Fixes**
  - Improved handling of expired, invalid, or incomplete tokens.
  - Strengthened token expiry validation and Portal connection URL handling.
  - Improved retry behavior for temporary device-login connection failures.
  - Enhanced refresh-intent security and cleanup handling.

- **Documentation**
  - Fixed Kiro installation command formatting across localized provider guides so pipe characters render correctly.


Ingwannu (changes requested):
The incremental direction is correct, but I am requesting changes on current head `25d462fd3210798795e7dc4637793a5c5ee86bca` for two in-scope correctness gaps:

1. The live test still uses `stored.accountId` as the account-store row id in both `refreshGenericAccountWithLock()` and the new `getAccountCredential()` call. Those are different identities: `saveCredential()` stores rows under `newAccountId(credential)`, an 8-character SHA-256-derived id, while `credential.accountId` is the provider/JWT identity. As written, `getAccountCredential(provider, accountId)` returns null and the coordinator fails before the live refresh. Read `getAccountSet("nous")!.activeAccountId` once, assert it exists, and pass that row id to both the coordinator and the post-refresh read.
2. The same raw Kiro install pipes still exist in the English table at `docs-site/src/content/docs/guides/providers.md:117`. Escaping only ja/ko/ru/zh-cn leaves the canonical English table malformed. Apply the same `&#124;` substitution there and keep the displayed commands unchanged.

The current exact-head CI is also not green: `test 3/4` was cancelled during its Test step after 15 minutes, which makes the aggregate `ci` job fail. After the focused fixes, rerun that shard/CI and keep the merge gated on a complete green head.

I am not folding the broad CodeRabbit comments against the already-merged #1397 base into this five-file follow-up review; this request is limited to defects introduced or left incomplete by the incremental diff.

Issue #1059, opened 2026-08-05
Windows test suite remains dispatch-only until full suite is green


lidge-jun:
### Client or integration

Other (test suite / CI)

### Area

Windows CI / cross-platform test suite

## Current status

The Windows full test-suite leg is currently **dispatch-only** and does not gate normal development or release promotion.

This issue tracks the Windows failure burn-down and the eventual restoration of Windows as a required CI gate.

The original issue description cited approximately 207 failures from an older aborted promotion run. That number is no longer a reliable baseline.

A newer full Windows dispatch against `dev` produced the current useful baseline:

| Shard | Result |
|---|---|
| 1/4 | timed out at the 15-minute cap; no final failure count |
| 2/4 | 8 failures |
| 3/4 | 21 failures |
| 4/4 | 47 failures |

So the current known baseline is:

```text
76 failures across the three completed shards
+ unknown failures in shard 1 because the shard timed out
```

The Windows leg should remain dispatch-only until a full run completes successfully on current `dev`.

## Current reference run

Windows dispatch:

```text
https://github.com/lidge-jun/opencodex/actions/runs/31095755263
```

All four Windows shards failed or timed out in that run.

The old `~207` count from the earlier promotion run should not be used as the burn-down target.

## Main failure family

The recoverable failures are not evenly distributed.

The largest coherent cluster is around the **Codex config/catalog write substrate**, including:

- `injectCodexConfig` integration;
- Codex catalog sync hardening;
- Codex catalog restore;
- cross-process write-lock contention;
- `codex-journal`.

The failure signatures point toward Windows-specific filesystem and process semantics around areas such as:

- atomic publication;
- replacement/rename behavior;
- permission normalization;
- file locking;
- concurrent writers;
- journal/CAS behavior.

These should be treated as a shared substrate problem where the evidence supports it rather than as dozens of unrelated test failures.

## Remaining distinct failure classes

Other observed failures include areas such as:

- server same-target `429` retry end-to-end behavior;
- expired forward replay state related to #702;
- Windows-specific process/test timing.

These should be handled separately once the dominant write-substrate failures are removed.

## Shard 1 timeout

Shard `1/4` is a separate observability problem.

It exceeded the current 15-minute job limit before producing a usable final test result.

That means one quarter of the Windows suite currently provides no complete failure inventory.

Before treating shard 1 as simply another group of failing tests, determine:

1. which tests consume the time;
2. whether retries/filesystem waits are responsible;
[...]

PR #1452, opened 2026-08-11
fix(windows): preserve ARM64 ACL identity lookup

### LINKED ISSUE #1449: ACL hardening Error - QEMU ARM64 Windows VM
### Client or integration

Other

### Area

CLI

### Summary

After version 2.10.2, i get hit by this error,

DeepSeek-V4-Flash-0731 answer:
Root cause found
Your machine is a QEMU ARM64 Windows VM (virt-7.2, ARM64 OS) and ocx runs on the bundled Bun v1.3.14 Windows-arm64 binary, which has bun:ffi compiled out ("TinyCC is disabled" — TinyCC doesn't support aarch64, a known open platform gap, oven-sh/bun#28055).

C:\Users\Administrator>ocx start
[opencodex] Config mutation coordination directory ACL hardening did not complete; continuing without it. ACL hardening failed (EACLIDENTITY) — the effective Windows account SID could not be resolved
570 |     && String((error as NodeJS.ErrnoException).code) === "ETIMEDOUT";
571 | }
572 |
573 | /** Preserve only the bounded machine-readable cause on a sanitized public error. */
574 | function sanitizedAclError(diagnostics: string, cause: unknown): NodeJS.ErrnoException {
575 |   const error = new Error(diagnostics) as NodeJS.ErrnoException;
                          ^
error: ACL hardening failed (EACLIDENTITY) — the effective Windows account SID could not be resolved
 code: "EACLIDENTITY"

      at sanitizedAclError (C:\Users\Administrator\AppData\Roaming\npm\node_modules\@bitkyc08\opencodex\src\lib\windows-secret-acl.ts:575:21)
      at hardenEntry (C:\Users\Administrator\AppData\Roaming\npm\node_modules\@bitkyc08\opencodex\src\lib\windows-secret-acl.ts:728:28)
      at harden (C:\Users\Administrator\AppData\Roaming\npm\node_modules\@bitkyc08\opencodex\src\config.ts:222:39)
      at atomicWriteFile (C:\Users\Administrator\AppData\Roaming\npm\node_modules\@bitkyc08\opencodex\src\config.ts:235:8)
      at persistConfigUnlocked (C:\Users\Administrator\AppData\Roaming\npm\node_modules\@bitkyc08\opencodex\src\config.ts:2559:3)
      at <anonymous> (C:\Users\Administrator\AppData\Roaming\npm\node_modules\@bitkyc08\opencodex\src\config.ts:2572:9)
      at withConfigMutationLockSync (C:\Users\Administrator\AppData\Roaming\npm\node_modules\@bitkyc08\opencodex\src\config.ts:2426:19)
      at saveConfig (C:\Users\Administrator\AppData\Roaming\npm\node_modules\@bitkyc08\opencodex\src\config.ts:2570:3)
      at reconcileOAuthProviders (C:\Users\Administrator\AppData\Roaming\npm\node_modules\@bitkyc08\opencodex\src\oauth\index.ts:819:16)
      at startServer (C:\Users\Administrator\AppData\Roaming\npm\node_modules\@bitkyc08\opencodex\src\server\index.ts:469:3)

Bun v1.3.14 (Windows arm64)

### Reproduction

ocx start

### Version

2.12.0

### Operating system

QEMU ARM64 Windows 11 VM

### Provider and model

_No response_

### Logs or error output

```shell

```

_No response_

### Redacted configuration

```json

```

### Checks

Ingwannu:
## Summary

- Preserve required secret-file ACL hardening on Windows ARM64 when Bun 1.3.14 cannot load the `bun:ffi` System32 resolver.
- Keep `GetSystemDirectoryW` authoritative on supported runtimes, then permit only the non-elevated SID lookup on `win32/arm64` to use the fixed protected default PowerShell path when that exact file exists.
- Keep environment-derived paths and account names outside the trust boundary, leave UAC and Task Scheduler resolution unchanged, and fail closed for non-default Windows roots.
- Add regression coverage for the ARM64 selection boundary, hostile environment values, unsupported platforms/architectures, missing files, and primary-resolver precedence.
- Document the decision and tradeoff in the config/Codex-home architecture note.

Root cause: the effective-SID query resolved PowerShell through `GetSystemDirectoryW` via `bun:ffi`. Bun 1.3.14 on Windows ARM64 cannot provide that FFI call, so the required ACL path surfaced `EACLIDENTITY` before PowerShell started.

Closes #1449

## Verification

- `taskset -c 0,1 nice -n 10 bun test tests/windows-user-principal.test.ts tests/windows-secret-acl.test.ts tests/windows-elevation.test.ts tests/windows-elevation-spawn.test.ts tests/windows-popup-fix.test.ts` — 238 pass, 0 fail on exact PR head.
- `taskset -c 0,1 nice -n 10 bun run typecheck` — passed on exact PR head.
- `taskset -c 0,1 nice -n 10 bun run privacy:scan` — passed on exact PR head.
- The first CPU-limited full-suite run completed with 10,778 pass, 10 skip, and 9 unrelated failures/errors. Six occurred before the GUI dependency install step; the two reported image/bridge failures passed when rerun in isolation. The remaining `codex-shim` failure reproduces in an untouched test path because the fixture reads the process runtime token instead of `local-secret`; neither its implementation nor test differs from `dev` in this PR.

## Checklist

### DIFF
--- a/src/lib/windows-user-principal.ts
+++ b/src/lib/windows-user-principal.ts
@@ -20,11 +20,67 @@
  * would mean passing an absolute deadline through the runner interface.
  */
 
+import { existsSync } from "node:fs";
+import { win32 as windowsPath } from "node:path";
+
 import { resolveTrustedWindowsPowerShellExe } from "./windows-elevation";
 
 const SID_PATTERN = /^S-1-(?:\d+-)+\d+$/i;
 const SID_EXPRESSION =
   "[System.Security.Principal.WindowsIdentity]::GetCurrent().User.Value";
+const DEFAULT_WINDOWS_ARM64_POWERSHELL = windowsPath.join(
+  "C:\\Windows\\System32",
+  "WindowsPowerShell",
+  "v1.0",
+  "powershell.exe",
+);
+
+type PrincipalExecutableResolution = Readonly<{
+  platform: NodeJS.Platform;
+  arch: string;
+  resolveTrusted: () => string;
+  pathExists: (path: string) => boolean;
+}>;
+
+/**
+ * Bun's Windows ARM64 build cannot currently execute the `bun:ffi` call used
+ * by the general System32 resolver. The ACL identity lookup has a narrower
+ * authority than the elevation helpers: it starts one non-elevated PowerShell
+ * command and accepts only a SID-shaped result. Keep its fallback equally
+ * narrow by using the fixed, OS-protected default installation path only.
+ *
+ * Environment and PATH lookup are deliberately absent. A Windows installation
+ * outside C:\\Windows keeps failing closed rather than executing a binary chosen
+ * by caller-controlled `SystemRoot`, `WINDIR`, or `PATH` values.
+ */
+function resolveWindowsPrincipalPowerShellExecutable(
+  resolution: PrincipalExecutableResolution = {
+    platform: process.platform,
+    arch: process.arch,
+    resolveTrusted: resolveTrustedWindowsPowerShellExe,
+    pathExists: existsSync,
+  },
+): string {
+  try {
+    return resolution.resolveTrusted();
+  } catch (error) {
+    if (
+      resolution.platform !== "win32" ||
+      resolution.arch !== "arm64" ||
+      !resolution.pathExists(DEFAULT_WINDOWS_ARM64_POWERSHELL)
+    ) {
+      throw error;
+    }
+    return DEFAULT_WINDOWS_ARM64_POWERSHELL;
+  }
+}
+
+/** Test-only dependency-injected view of the Windows ARM64 executable boundary. */
+export function resolveWindowsPrincipalPowerShellExecutableForTests(
+  resolution: PrincipalExecutableResolution,
+): string {
+  return resolveWindowsPrincipalPowerShellExecutable(resolution);
+}
 
 export interface WindowsPrincipalLookupResult {
   success: boolean;
@@ -52,7 +108,7 @@ const POWERSHELL_ARGS = [
 ] as const;
 
 function windowsPrincipalPowerShellCommand(): string[] {
-  return [resolveTrustedWindowsPowerShellExe(), ...POWERSHELL_ARGS];
+  return [resolveWindowsPrincipalPowerShellExecutable(), ...POWERSHELL_ARGS];
 }
 
 /** Test-only readback of the exact trusted executable and static arguments. */
--- a/structure/02_config-and-codex-home.md
+++ b/structure/02_config-and-codex-home.md
@@ -81,6 +81,22 @@ the recorded service ownership.
 sequence number) to avoid collisions when concurrent writers (e.g. `ocx stop` and the proxy's own
 shutdown handler) both restore Codex config simultaneously. The temp is renamed atomically into place.
 
+Windows secret-file hardening resolves the effective token SID through an absolute, trusted
+PowerShell path before granting the owner and removing inherited broad ACL entries. The normal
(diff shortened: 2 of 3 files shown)


Wibias (changes requested):
Request changes on current head `47271935fb973070216a7cb16f933303e527a07a`.

Merge blocker: the new Windows ARM64 PowerShell fallback is triggered by every failure from the trusted resolver, not only by the specific Bun/FFI-unavailable condition this PR is meant to work around.

The current shape is effectively:

```ts
try {
  return resolution.resolveTrusted();
} catch (error) {
  if (
    resolution.platform !== "win32" ||
    resolution.arch !== "arm64" ||
    !resolution.pathExists(DEFAULT_WINDOWS_ARM64_POWERSHELL)
  ) {
    throw error;
  }

  return DEFAULT_WINDOWS_ARM64_POWERSHELL;
}
```

That catch-all weakens the existing trust boundary. On ARM64 it can replace any trusted-resolution or validation failure with the hard-coded `C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe` path merely because that file exists. This includes failures from `GetSystemDirectoryW`, non-default Windows roots, missing/invalid trusted PowerShell resolution, containment/integrity checks, or future security checks.

This matters because the selected PowerShell is used for effective SID/principal discovery feeding secret-file ACL hardening. A trusted resolver failure must not be silently converted into a different identity source.

Please narrow the fallback to the one explicitly expected failure mode, e.g. a typed/sentinel `bun:ffi` / System32-resolution-unavailable error. All other resolver and validation failures must continue to fail closed.

Required regressions should prove at minimum:

- win32/arm64 + explicit FFI-unavailable condition -> fixed fallback may be used when the expected file exists;
- `GetSystemDirectoryW` call failure -> no fallback;
- invalid/non-default trusted system directory -> no fallback;
- trusted PowerShell validation/containment failure -> no fallback;
- arbitrary resolver error -> no fallback;
- successful trusted resolver -> primary result always wins.

Also preserve the stated non-default-Windows-root contract. `existsSync("C:\\Windows\\...")` alone is not evidence that `C:\\Windows` is the active protected Windows root, so the fallback must not override an authoritative OS-derived non-default root.

Exact-head CI was green in the previous review, so this request is about the trust/fail-closed contract rather than a failing test.

lidge-jun:

On `dev` as `fcb337a18` with your authorship intact, plus `4368bb352` closing the reviewer's blocker.

The blocker was right and worth the extra round: the fallback was a catch-all. On ARM64 it substituted the hard-coded `C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe` for *any* trusted-resolution failure whenever that file happened to exist — `GetSystemDirectoryW` failure, a non-default Windows root, a containment or integrity check failure, and anything added later. That PowerShell performs effective-SID discovery feeding secret-file ACL hardening, so it converted a trust-boundary failure into a different identity source.

`WindowsSystemDirectoryFfiUnavailableError` is now thrown only where `loadGetSystemDirectoryW()` cannot provide the FFI function, and the fallback requires that typed sentinel plus `win32` plus `arm64` plus the fixed executable. Everything else rethrows unchanged. The existing test that threw a generic `Error("bun:ffi unavailable")` was certifying the unsafe shape, so it now uses the sentinel.

Ablation: removing the sentinel gate turns four regressions red — API-call failure, unusable non-default directory, validation failure, and an arbitrary resolver error each returned the fixed System32 path. 27 pass / 0 fail with the gate in place.

No real ARM64 Windows behavior was exercised (`windows N/4` is dispatch-only, #1059); the resolver seam is injected, so these run on macOS. Thanks for the fix — the QEMU ARM64 diagnosis in the issue is what made it findable.

PR #1460, opened 2026-08-11
fix(catalog): skip byte-identical catalog writes so a resync cannot mark Codex stale

### LINKED ISSUE #1459: [Bug]: byte-identical catalog resync bumps the mtime and latches the app-server staleness check to stale
### Client or integration

Codex CLI

### Area

Catalog / models

### Summary

`writeRetainedCatalogSync` writes the Codex catalog unconditionally — `replaceActiveCodexCatalog` → `atomicWriteFile`, with no comparison against what is already on disk. So every sync moves the catalog file's mtime even when the produced bytes are identical and the advertised model set did not change.

`collectCodexAppServerCatalogState()` compares that mtime against each running Codex's start time, so an ordinary `ocx start` — or any dashboard/CLI action that re-syncs an unchanged model set — classifies every already-running Codex as holding an outdated in-memory catalog. Since #1407 that verdict withholds all opencodex-authored v2 model guidance, so a configured `injectionModel` / `subagentModels` roster silently stops reaching the session for the rest of that Codex's lifetime, even though nothing about the catalog changed.

A long-lived Codex App app-server outlives every proxy restart, which makes the state effectively latched: restarting the Codex CLI does not clear it, and the only way out is quitting the Codex App or `ocx sync --restart-codex`.

Expected: a sync that produces no change should not make a running Codex look stale. `#1046` covered warning about stale app-servers after a catalog rewrite; it did not cover avoiding the rewrite when there is nothing to write.

### Reproduction

1. macOS with the Codex App (ChatGPT desktop) running, so a long-lived `/Applications/ChatGPT.app/Contents/Resources/codex … app-server` process exists.
2. `ocx start` with a settled config, so the sync reproduces the catalog it already wrote.
3. Observe the catalog mtime move even though the file content is unchanged, and `collectCodexAppServerCatalogState()` report `stale`.
4. Start a new Codex CLI session with a v2 `spawn_agent` surface and a configured `injectionModel`: no preferred model and no roster reach the session.

Measured on the affected machine (`collectCodexAppServerCatalogState()`):

```
state: stale
catalogMtime: 2026-08-11T02:52:27.788Z
pid 80794  start 2026-08-11T02:43:11.000Z  stale: true   <- Codex App app-server
pid 95407  start 2026-08-11T02:55:35.000Z  stale: false  <- codex-code-mode-host (live CLI session)
pid 91809  start 2026-08-11T02:53:12.000Z  stale: false  <- codex-code-mode-host (live CLI session)
```

The mtime bump on identical bytes, reproduced hermetically through the same writer:

```
identical bytes: true
mtime before/after identical rewrite: 2026-08-11T04:08:17.643Z -> 2026-08-11T04:08:18.746Z  bumped: true
```

### Version

2.12.0

### Operating system

macOS 15.3.1

### Provider and model

openai (ChatGPT login) / gpt-5.6-luna

### Logs or error output

```shell
# state on the affected machine, and the same state with the Codex App process excluded
verdict incl. the Codex App app-server : stale
verdict excl. the Codex App app-server : fresh
```

### Redacted configuration

```json
{
  "multiAgentGuidanceEnabled": true,
  "injectionModel": "gpt-5.6-luna",
  "injectionEffort": "max",
  "subagentModels": ["gpt-5.6-luna"]
}
```

### Checks

tlsdnwn55:
## Summary

`writeRetainedCatalogSync` wrote the Codex catalog unconditionally — `replaceActiveCodexCatalog` →
`atomicWriteFile`, with no comparison against the bytes already on disk — so every sync moved the
catalog file's mtime even when the produced content was identical and the advertised model set did
not change.

`collectCodexAppServerCatalogState()` compares that mtime against each running Codex's start time.
An ordinary `ocx start`, or any dashboard/CLI action that re-syncs an unchanged model set, therefore
classified every already-running Codex as holding an outdated in-memory catalog. Since #1407 that
verdict withholds all opencodex-authored v2 model guidance, so a configured `injectionModel` and
`subagentModels` roster silently stop reaching the session for the rest of that Codex's lifetime,
even though nothing about the catalog changed. A long-lived Codex App app-server outlives every
proxy restart, which makes the state effectively latched: restarting the Codex CLI does not clear it.

This PR skips the write when the prepared bytes equal what is on disk, and reports
`catalogWritten: false` for that case.

- The comparison is **byte-exact**, not a decoded-string comparison. `readFileSync(path, "utf8")`
  replaces every invalid sequence with `U+FFFD`, so a malformed byte on disk would compare equal to
  a legitimately encoded replacement character in the prepared content; the skip would then preserve
  the corruption and report `catalogWritten: false` for a file that is not what we prepared. An
  unreadable or absent file still falls back to a real write.
- `added` still reports the routed rows the catalog carries, because they are on disk either way.
- `cacheSynced` is unaffected: `refreshCodexModelCatalog` invalidates the models cache whenever the
  catalog exists, independently of whether it was rewritten. The stale-app-server hint and the
  startup cache-invalidation branch both key on `cacheSynced` as well, so their behaviour is
  unchanged.
- `#1046` covered warning about stale app-servers *after* a catalog rewrite; it did not cover
  avoiding the rewrite when there is nothing to write.

Closes #1459

## Verification

- `bun test tests/codex-catalog-sync-hardening.test.ts tests/codex-convergence-account-selectors.test.ts tests/codex-refresh.test.ts tests/codex-sync-api.test.ts tests/codex-catalog-restore.test.ts tests/codex-retained-root-serialization.test.ts`
  — 72 pass, 0 fail.
- `bun run test` (full suite) — 10871 pass, 7 skip, 1 fail. The single failure,
  `server combo failover 030 activation matrix > connect cancellation wins with 499, no backup,
  warning, or cooldown` (30s hook timeout), reproduces identically with this branch's changes
  stashed on unmodified `dev`, and touches neither catalog sync nor guidance. That run predates the
  rebase onto current `dev` and the byte-exactness change.
- Both regressions were driven red against the behaviour they describe:
  - removing the equal-content guard fails `secondWritten` (`Expected: false, Received: true`), and
    the same test mutates the catalog on disk and asserts the next sync *does* rewrite it, so the
    guard is not vacuous;
  - restoring the decoded-string comparison fails
    `a malformed on-disk byte that decodes to the same string is still repaired` on `written`
    (`Expected: true, Received: false`). That test corrupts one byte inside a preserved JSON string
    value to `0x80`, then asserts the repaired file differs in bytes while decoding to the same
    string as the corrupted file — exactly the pair a string comparison equates — and that the
    result is valid UTF-8 containing `EF BF BD`.
- Reproduced on macOS 15.3.1 with `@bitkyc08/opencodex` 2.12.0, Codex CLI 0.147.0
  (`originator: codex-tui`), Codex App running: `collectCodexAppServerCatalogState()` reported
  `stale` solely because of the Codex App app-server (started 02:43:11Z) against a catalog whose
  mtime had been moved to 02:52:27Z by a resync, while both live CLI-owned processes were newer.

## Checklist

- [x] Docs or release notes were updated when needed — the sub-agent surface guide's catalog-state
      section now states that a byte-identical resync leaves the file untouched.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. No auth,
      credential, workflow, or release-automation surface is touched. The write path keeps its
      existing permit check, write serialization, and atomic-write behaviour; the only change is an
      equal-content short-circuit evaluated inside the same serialized region.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Catalog synchronization now avoids rewriting unchanged catalog files, preserving their timestamps.
  * Invalid file contents are repaired reliably, including malformed UTF-8 data.
  * Catalog updates continue to report routed and account-bound entries even when no rewrite is needed.

* **Documentation**
  * Clarified when catalog timestamps change and how unchanged synchronization affects running app-server freshness.

### DIFF
--- a/docs-site/src/content/docs/guides/sub-agent-surface.md
+++ b/docs-site/src/content/docs/guides/sub-agent-surface.md
@@ -231,6 +231,10 @@ timestamp, an unreadable process start time, or a failed process enumeration —
 separately by `ocx doctor`. `stale` clears only after every detected Codex app-server starts after
 the final catalog write; it does not necessarily clear `unknown`.
 
+Only a real change counts. A sync whose result is byte-identical to the catalog already on disk
+leaves the file untouched, so restarting the proxy or re-syncing an unchanged model set does not
+make a running Codex look stale.
+
 ### Reasoning effort
 
 `injectionEffort` affects only delegated-worker guidance and, when explicitly enabled, native Codex
--- a/src/codex/catalog/sync.ts
+++ b/src/codex/catalog/sync.ts
@@ -1186,6 +1186,23 @@ function revalidateRetainedCatalogSync(
   };
 }
 
+/**
+ * Exact bytes currently on disk at `path`, or null when unreadable/absent.
+ *
+ * Bytes, not a decoded string: `readFileSync(path, "utf8")` replaces every
+ * invalid sequence with `U+FFFD`, so a malformed byte on disk would compare
+ * equal to a legitimately encoded replacement character in the prepared
+ * content. The equal-content skip below would then preserve the corruption and
+ * report `catalogWritten: false` for a file that is not what we prepared.
+ */
+function currentCatalogFileBytes(path: string): Buffer | null {
+  try {
+    return readFileSync(path);
+  } catch {
+    return null;
+  }
+}
+
 function pristineCatalogBytes(read: RetainedCatalogSyncRead): string | null {
   if (read.onDiskCatalog && !catalogHasRoutedEntries(read.onDiskCatalog)) {
     try {
@@ -1355,12 +1372,30 @@ function writeRetainedCatalogSync({
   });
   clampCatalogModelsToCodexSupport(catalog.models);
 
+  const added = goEntries.length + accountBoundEntries.length;
+  const content = `${JSON.stringify(catalog, null, 2)}\n`;
+  // A byte-identical rewrite is not a catalog change, but every mtime-keyed reader
+  // has to treat it as one. The app-server staleness classifier (#857) is the one
+  // that matters: it compares this file's mtime against each running Codex's start
+  // time, so an ordinary `ocx start` — or any dashboard action that re-syncs an
+  // unchanged model set — marked every already-running Codex as holding an outdated
+  // in-memory catalog. Since #1407 that verdict silences opencodex's own model
+  // guidance entirely (no preferred model, no roster) for the rest of that Codex's
+  // lifetime, so a configured injectionModel stops reaching the session even though
+  // nothing about the catalog changed. Skipping the no-op write keeps both the mtime
+  // and `catalogWritten` honest; `added` still reports the routed rows the catalog
+  // carries, because they are on disk either way.
+  const onDiskBytes = currentCatalogFileBytes(catalogPath);
+  if (onDiskBytes !== null && onDiskBytes.equals(Buffer.from(content, "utf8"))) {
+    return { added, path: catalogPath, catalogWritten: false, comboOmissions };
+  }
+
   replaceActiveCodexCatalog(permit, owningCodexHome, {
     path: catalogPath,
-    content: `${JSON.stringify(catalog, null, 2)}\n`,
+    content,
   });
   return {
-    added: goEntries.length + accountBoundEntries.length,
+    added,
     path: catalogPath,
     catalogWritten: true,
     comboOmissions,
--- a/tests/codex-catalog-sync-hardening.test.ts
+++ b/tests/codex-catalog-sync-hardening.test.ts
@@ -968,6 +968,135 @@ describe("Codex catalog sync hardening", () => {
     expect(slugs).not.toContain("cursor/stale-model");
   });
 
+  test("an identical resync leaves the catalog file untouched, a real change still writes", () => {
+    // The app-server staleness classifier (#857) compares this file's mtime against
+    // each running Codex's start time, so a no-op rewrite would report every
+    // already-running Codex as holding an outdated catalog — and since #1407 that
+    // verdict withholds opencodex's model guidance for the rest of that Codex's
+    // lifetime, even though the advertised model set never changed.
+    const catalogPath = join(codexHome, "catalog.json");
+    writeFileSync(join(codexHome, "config.toml"), 'model_catalog_json = "catalog.json"\n', "utf8");
+    writeFileSync(catalogPath, JSON.stringify({
+      models: [
+        nativeEntry("gpt-5.5", 0),
+        nativeEntry("gpt-5.2", 104), // legacy -> dropped by the first sync
+      ],
+    }, null, 2) + "\n");
+
+    const r = runScript(codexHome, opencodexHome, `
+      const { statSync, writeFileSync, readFileSync } = require("node:fs");
+      const { syncCatalogModels } = require("./src/codex/catalog");
+      const path = ${JSON.stringify(catalogPath)};
+      const sleep = ms => new Promise(resolve => setTimeout(resolve, ms));
+      (async () => {
+        const first = await syncCatalogModels({ providers: {} });
+        const afterFirst = statSync(path).mtimeMs;
+        await sleep(1100);
+        const second = await syncCatalogModels({ providers: {} });
+        const afterSecond = statSync(path).mtimeMs;
+        // Not vacuous: a catalog that really differs must still be rewritten.
+        const catalog = JSON.parse(readFileSync(path, "utf8"));
+        catalog.models = catalog.models.filter(model => model.slug !== "gpt-5.5");
+        writeFileSync(path, JSON.stringify(catalog, null, 2) + "\\n");
+        const changedAt = statSync(path).mtimeMs;
+        await sleep(1100);
+        const third = await syncCatalogModels({ providers: {} });
+        console.log(JSON.stringify({
+          firstWritten: first.catalogWritten,
+          secondWritten: second.catalogWritten,
+          secondAdded: second.added,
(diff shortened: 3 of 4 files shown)


Ingwannu (changes requested):
The reported issue is real, and this is the right minimal direction: avoiding a true no-op write keeps the catalog mtime meaningful without weakening the existing permit, revalidation, or atomic-write path. I reproduced the focused suite on exact PR head `7bf8b84a` after installing the locked dependencies: **100 pass / 0 fail**; typecheck and privacy scan also pass.

I am requesting one correctness change before this becomes review-ready:

`currentCatalogFileContent()` in `src/codex/catalog/sync.ts` claims to compare exact bytes, but `readFileSync(path, "utf8") === content` compares decoded JavaScript strings. Invalid UTF-8 bytes are replaced with `U+FFFD`, so different on-disk bytes can compare equal to a legitimately encoded replacement character. I reproduced the boundary directly: a buffer containing raw byte `0x80` and a buffer containing UTF-8 `EF BF BD` have equal decoded strings but unequal bytes. In a catalog string field, this path can therefore preserve malformed bytes, skip the atomic repair write, and return `catalogWritten: false` even though the prepared bytes are not identical to the file.

Please read the file as bytes and compare it with the UTF-8 bytes of `content` (`Buffer.equals` or an equivalent exact `Uint8Array` comparison), while preserving the current unreadable/absent fallback to a real write. Add a regression where an invalid on-disk byte decodes to the same string as a real `U+FFFD` in the prepared content and assert that the sync still rewrites it. The existing mtime no-op regression should continue to prove that genuinely identical bytes are skipped.

The branch is currently four commits behind `dev`, still inside the repository's accepted freshness window. Once the exact-byte boundary is fixed, the focused checks are rerun, and the four readiness boxes are completed against the resulting head, this remains a strong merge candidate for #1459.

tlsdnwn55:
@Ingwannu thanks — the finding is correct and is fixed on new head `0b0fee65`.

`currentCatalogFileContent()` is replaced by `currentCatalogFileBytes()`, which returns the raw `Buffer`; the skip now requires `onDiskBytes.equals(Buffer.from(content, "utf8"))`. The unreadable/absent fallback to a real write is unchanged, and the check still runs inside the same serialized region under the existing permit.

Regression added: `a malformed on-disk byte that decodes to the same string is still repaired`. It settles the catalog, corrupts one byte inside a preserved JSON string value to `0x80`, and asserts that the next sync rewrites the file (`catalogWritten: true`), that the repaired bytes differ from the corrupted file **while decoding to the same string** — the exact pair a decoded-string comparison equates — and that the result is valid UTF-8 containing `EF BF BD`. Restoring `onDiskBytes.toString("utf8") === content` fails it on `written` (`Expected: true, Received: false`), so it is not vacuous.

Re-verified on `0b0fee65`: `bun x tsc --noEmit` clean; `bun test` on `codex-catalog-sync-hardening`, `codex-convergence-account-selectors`, `codex-refresh`, `codex-sync-api`, `codex-catalog-restore`, `codex-retained-root-serialization` — 72 pass / 0 fail; `bun run privacy:scan` passed.

The branch was force-pushed to rewrite the commit author, so the SHA you reviewed is superseded. Sorry for the extra round.

tlsdnwn55:
@coderabbitai review

tlsdnwn55:
Rebased onto the current `dev` head `38c9b355` (no conflicts; the diff is unchanged from `0b0fee65`, which already carried the byte-exact comparison and its regression). New head: `642805c1`.

Verified locally on `642805c1`:

- `bun test` on the affected subsystems — 292 pass / 1 skip / 0 fail:
  `codex-catalog-sync-hardening`, `codex-convergence-account-selectors`, `codex-refresh`, `codex-sync-api`, `codex-catalog-restore`, `codex-retained-root-serialization`, `codex-app-server-processes`, `codex-desired-state`, `codex-models-cache-invalidate`, `codex-native-residue`, `injection-model-api`, `multi-agent-compat`, `combos`

Scope note on the full-suite claim, so the attestation is precise: the last full `bun run test` I ran was on the pre-rebase branch (10871 pass / 7 skip / 1 fail — `server combo failover 030 activation matrix > connect cancellation wins with 499, no backup, warning, or cooldown`, a 30s hook timeout that reproduces identically on unmodified `dev` with my changes stashed). It predates the rebase onto `38c9b355`. Since then I have run the focused set above rather than the whole suite; the diff itself is byte-identical to the reviewed `0b0fee65`, and it touches only the catalog write path plus its tests and one docs paragraph.

lidge-jun:
Landed on `dev` — closing this pull request as superseded rather than merged, because the change is already there and the branch now conflicts with it.

Your fix is on `dev` as `c7eec01ca` with your authorship intact, plus a follow-up correcting the comparison. The reviewer's blocker was real: `currentCatalogFileContent()` used `readFileSync(path, "utf8") === content`, which compares decoded strings, so a catalog holding a bare `0x80` compares equal to prepared content holding a legitimately encoded `U+FFFD` — the guard would have preserved the corruption and reported `catalogWritten: false` for a file that is not what we prepared. Reproduced directly: `{"decodedEqual":true,"bytesEqual":false}`.

You independently reached the same `Buffer.equals` fix on your newer head, which is why this is closing as superseded rather than as something we had to replace. The landed version differs from yours only in the helper name (`currentCatalogFileContent` vs `currentCatalogFileBytes`) and comment wording.

Regression on `dev` converges the catalog on a retained description carrying a real `U+FFFD`, corrupts exactly that `EF BF BD` into a bare `0x80`, and asserts the pair decodes equal while differing in bytes, that the sync rewrites and repairs it, and that a genuinely byte-identical resync is still skipped. Verified red against the string comparison and green after.

Thank you — this was a good catch and a clean fix.

PR #1461, opened 2026-08-11
fix(codex): preflight sync injection before catalog writes

### LINKED ISSUE #1453: ocx sync and restore back fail permanently with an unnamed config issue, leaving the Codex catalog truncated and routing un-injected
### Client or integration

Codex CLI

### Area

CLI

### Summary

`ocx sync` and `ocx restore back` fail permanently on my install with a message that names no reason:

```
Codex sync did not complete. Fix the reported Codex config issue and retry.
Plain `codex` was not switched back to opencodex. Fix the reported Codex config issue and retry.
```

No "Codex config issue" is ever printed — not on stdout, not on stderr, and `ocx doctor` reports **every check `ok`**. There is nothing to act on, so the instruction "fix the reported issue and retry" cannot be followed.

This has two compounding consequences, because of the ordering documented in `src/cli/index.ts`:

```ts
// Only warn/restart when a catalog or models_cache write actually happened. This is
// deliberately not an `else`: refreshCodexModelCatalog runs before injectCodexConfig,
// so a sync can fail (`ok: false`) after the catalog was already rewritten
```

1. **The model catalog is rewritten before the injection fails**, so it is left truncated. My proxy serves **435 models** (200 native, 157 openrouter/opencode-go, 43 Anthropic, 35 Google) but `opencodex-catalog.json` contains only **9**, all OpenAI-family. The non-OpenAI models are reachable through the proxy but never appear in Codex.
2. **Routing is never restored.** `ocx stop` and `ocx update` both remove the injected `openai_base_url` and `model_catalog_json` root keys. Since `sync` / `restore back` can never complete, nothing puts them back. After every update I have to re-add them by hand from `<CODEX_HOME>/opencodex.config.toml`.

Expected: either the sync succeeds, or the specific config problem is printed so it can be fixed. Ideally the catalog would also not be rewritten until the injection is known to succeed.

### Reproduction

100% reproducible on my install, across `ocx update`, full restarts of the proxy, and two different versions (2.11.0 and 2.12.0).

1. Have opencodex injected into Codex (`openai_base_url` + `model_catalog_json` present in `<CODEX_HOME>/config.toml`).
2. Run `ocx stop` (or `ocx update`, which stops as its first step). The two injected root keys are removed.
3. Start the proxy again and confirm it is healthy (`/healthz` → 200, `/v1/models` → full list).
4. Run `ocx restore back` → fails with the message above.
5. Run `ocx sync` → fails with the message above.
6. Inspect `opencodex-catalog.json` → rewritten, but containing only the OpenAI-family models.

The proxy is healthy throughout. The only way I have found to restore routing is editing `config.toml` manually using the keys documented in `<CODEX_HOME>/opencodex.config.toml`.

### Things I ruled out

To save triage time, I checked the obvious candidates and none of them apply:

- **Config is valid UTF-8** — strict `TextDecoder('utf-8', {fatal: true})` decode succeeds.
- **No duplicate TOML tables** — 185 `[projects.*]` entries, zero duplicate keys.
- **Size is modest** — 33,524 bytes.
- **Proxy is healthy** — `/healthz` 200, `/v1/models` returns 435 entries, live requests succeed.
- **`ocx doctor` is entirely clean** — every check `ok`, including "Codex restart safety", "Project Codex configs", and "OAuth reliability".
- **Not the stripped keys themselves** — the failure persists identically after routing is manually restored and verified working end-to-end.

### Not a duplicate of the nearby catalog issues

- **#1046** (stale app-servers) states *"The catalog file on disk is already correct at that point"*. Mine is the inverse: the file **on disk** is wrong — it contains 9 entries while the proxy serves 435. `ocx sync-cache` reports the disk catalog was updated and the count stays at 9, so this is not the stale-app-server case.
- **#241 / #851** are picker/transport-level (models loaded but not shown). Here they are absent from the generated catalog itself.
- **#1273** covers stale in-memory full-config writes resurrecting deleted providers — a different mechanism; nothing was deleted here.
- **#1054** is about `ocx start` injecting `openai_base_url` unwantedly. Here the injection *fails* and never happens.

### Suggested fixes

1. **Print the actual reason.** `syncModelsToCodex` returns `ok: false` but the caller only prints a generic sentence referring to a "reported" issue that is never reported. Surfacing the underlying error would make this self-service.
2. **Do not rewrite the catalog before the injection is known to succeed**, or roll the catalog back when injection fails — otherwise a failed sync actively degrades a working install by truncating the model list.
3. Consider having `ocx doctor` exercise the same code path, since it currently reports a fully healthy system while `sync` cannot complete.

### Version

2.12.0 (also reproduced on 2.11.0)

### Operating system

Windows 11 Pro 10.0.26200

### Provider and model

Proxy serves 435 models across openai (native), openrouter, opencode-go and anthropic; the Codex-facing catalog is reduced to 9 OpenAI-family entries.

### Logs or error output

```shell
$ ocx sync
Codex sync did not complete. Fix the reported Codex config issue and retry.

$ ocx restore back
Plain `codex` was not switched back to opencodex. Fix the reported Codex config issue and retry.

$ ocx sync-cache
WARNING: 4 Codex app-server process(es) still running (PIDs: ...). Disk catalog/cache were
updated, but Codex may keep showing the old model list until those processes restart.
# catalog still contains 9 models afterwards

$ ocx doctor
# every section reports ok, including:
#   Codex restart safety      ok
#   Project Codex configs     ok     no project-local provider bypass detected
#   OAuth reliability         [OK] x3

# proxy is healthy and multi-provider the whole time:
GET /healthz      -> 200
GET /v1/models    -> 435 models
   {"openai-native":200,"opencode-go/openrouter-other":157,"anthropic":43,"google":35}

# but the Codex-facing catalog:
<CODEX_HOME>/opencodex-catalog.json -> {"models": [ ...9 entries, all OpenAI-family... ]}
```

_No response_

### Redacted configuration

```json
{
  "providers": {
    "openai":  { "adapter": "openai-responses", "codexAccountMode": "direct", "authMode": "forward" },
    "openrouter": { "adapter": "openai-chat", "baseUrl": "https://openrouter.ai/api/v1" },
    "opencode-go": { "adapter": "openai-chat", "baseUrl": "https://opencode.ai/zen/go/v1" },
    "anthropic": { "adapter": "anthropic", "authMode": "oauth" }
  },
  "defaultProvider": "openrouter",
}
```

`<CODEX_HOME>/config.toml` is 33 KB with 185 `[projects.*]` trust entries, valid UTF-8, no duplicate keys.

### Checks

Ingwannu:
## Summary

- Run the real Codex injection transformation and coordination-eligibility checks in a non-writing preflight before provider discovery or catalog/cache replacement.
- Stop deterministic injection refusals before they can degrade a previously usable catalog, and send the concrete refusal to stderr instead of hiding it behind the generic CLI summary.
- Keep the normal injector call as the final write authority, and document why the preflight is an early guard rather than a reusable authorization result.
- Add regression coverage for refusal ordering, stderr fidelity, byte-identical catalog/cache preservation, and artifact-free successful preflight.

Closes #1453

## Exact-head verification

- Base: `dev@20c513198e4ef943fa990decad21fff92f3cfad4`
- Head: `9c9ca092a19b8e21c8d4b8aebaa91c10f1b4a3d2`
- `bun test tests/codex-sync-api.test.ts tests/cli-restore-back.test.ts` — 16 passed
- `git diff --check` — passed
- GitHub CI — gates, API/storage, all keyring and packaging jobs, macOS, and Linux shards 1/2/3 passed on this exact head. Linux shard 4 was cancelled at the 15-minute job limit, so the aggregate `ci` check is red.

The same patch on the immediately preceding `dev@87e3ff9f6` base also built all 221 documentation pages. Its local full suite reached 10,930 passed and 11 skipped, with no assertion failure in the changed sync/injection path.

The exact-head shard-4 log failed before the affected sync/injection tests: Bun 1.3.14 raised `EEXIST: file already exists, epoll_ctl` while loading `tests/catalog-oauth-observation.test.ts`, followed by `Cannot call beforeEach() after the test run has completed`. The runner then stopped producing test output after starting an unrelated Claude Desktop config-path test until the 15-minute limit cancelled it and Actions terminated an orphan `bun` process. The preceding code-equivalent runs hit the same Bun loader failure and dangling-process family on other Linux shards. No product assertion failed in the changed path, and no product code was changed to accommodate the runner defect; the PR remains draft pending independent review and a green required aggregate.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * `ocx sync` now validates managed Codex configuration before making changes.
  * Failed validation reports a specific error and preserves existing catalog, cache, configuration, and journal data.
  * `ocx restore back` applies the same safety checks before re-enabling routing.
  * Synchronization now clearly separates success messages from errors, improving command-line feedback.
* **Tests**
  * Added coverage confirming failed validation leaves existing files and synchronization artifacts unchanged.

### DIFF
--- a/docs-site/src/content/docs/reference/cli/lifecycle.md
+++ b/docs-site/src/content/docs/reference/cli/lifecycle.md
@@ -180,6 +180,11 @@ not fabricate official-client metadata. Doctor never mutates credentials or appl
 Fetch the live model list from every configured provider and re-inject the merged catalog into Codex.
 Run it after adding a provider or to refresh available models.
 
+Before provider discovery or catalog/cache replacement, `ocx sync` validates that the managed
+Codex configuration can be injected. If that validation refuses the config, the command exits
+nonzero, prints the concrete reason on stderr, and leaves the existing catalog and cache unchanged.
+`ocx restore back` uses the same no-write preflight before it re-enables routing.
+
 If long-lived Codex `app-server` processes are still running, `ocx sync` warns that they may keep
 serving the previous in-memory model list even though `opencodex-catalog.json` / `models_cache.json`
 were updated. Pass `--restart-codex` to send `SIGTERM` only to matching `codex … app-server` and
--- a/src/codex/inject.ts
+++ b/src/codex/inject.ts
@@ -129,6 +129,12 @@ export interface InjectCodexOptions {
    * caller that is willing to wait can raise it.
    */
   lockTimeoutMs?: number;
+  /**
+   * Validate the same config transformations and write-coordination eligibility without
+   * changing the journal, config, profile, catalog, cache, or history. Sync uses this before
+   * provider discovery so a deterministic config refusal cannot degrade an existing catalog.
+   */
+  validateOnly?: boolean;
 }
 
 function configuredManagedSubagentDefaults(
@@ -660,7 +666,7 @@ export async function injectCodexConfig(
   if (activeProvider) {
     // A launcher may have journaled before the provider manager took ownership. Never let shutdown
     // replay that stale snapshot over externally managed config.
-    removeJournal();
+    if (!options.validateOnly) removeJournal();
     const nativeSubagentDefaultsWarning = configuredManagedSubagentDefaults(
       config,
     )
@@ -864,6 +870,13 @@ export async function injectCodexConfig(
     return {
       success: false,
       message: `Codex configuration was not written: ${eligibility.reason}.`,
+    };
+  }
+
+  if (options.validateOnly) {
+    return {
+      success: true,
+      message: "Codex config injection preflight passed; no files were changed.",
     };
   }
 
--- a/src/codex/sync.ts
+++ b/src/codex/sync.ts
@@ -101,7 +101,8 @@ export async function syncModelsToCodex(
   const externalProvider = (deps.currentExternalCodexModelProvider ?? currentExternalCodexModelProvider)();
   if (externalProvider) {
     const result = await deps.injectCodexConfig(p, config, {});
-    log?.log(result.message);
+    if (result.success) log?.log(result.message);
+    else log?.error(result.message);
     reportCodexHomeTarget(log, deps.collectCodexHomeDiagnostic ?? collectOrcaCodexHomeDiagnostic);
     return {
       status: "applied",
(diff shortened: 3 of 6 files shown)


Ingwannu:
Exact-head CI review complete for `d6dad871d9b997673f98796216b756d64a935c09`.

The changed sync/injection path remains green locally (16 focused tests, typecheck, privacy scan, diff check). GitHub gates, service checks, macOS, packaging, keyring, API/storage jobs, and Linux shards 1/2 passed.

The remaining red jobs are repeatable Bun 1.3.14/runner failures, not product assertions:

- Linux shard 3 failed on both the initial attempt and its rerun with `EEXIST: file already exists, epoll_ctl` while loading unrelated test files. The later `Cannot call afterEach()/describe() after the test run has completed` errors are secondary fallout.
- Rerun shard 4 hit the same Bun loader error in unrelated memory/catalog tests, then the known `tests/cli-models.test.ts` dangling-process path stalled and the run was canceled.

I am not changing the product patch to accommodate this runtime failure and will not churn a third identical rerun. The PR remains draft for independent review/integration; no merge was performed.

PR #1462, opened 2026-08-11
fix(config): rebase stale live saves

### LINKED ISSUE #1273: [Bug] Stale full-config writes can resurrect deleted providers and custom models
### Client or integration

OpenCodex Dashboard / CLI / long-running proxy processes

### Area

Configuration persistence / provider and model management

## Current status

This issue originally contained two related defects.

### Defect 1 — resolved

Removing a provider used to leave that provider's entries behind in:

```text
config.customModels
```

which produced ghost model rows in the Dashboard and Codex catalog.

That lifecycle bug is fixed on `dev` by:

```text
14e948525
```

via #1293.

Provider removal now drops the provider's custom models in both:

* the CLI provider-removal path;
* the management API provider-removal path.

That portion is no longer part of the remaining scope.

### Defect 2 — still open

A long-running process can hold an older in-memory configuration and later persist the **entire stale document**, overwriting newer changes already committed by another cooperating process.

This can resurrect state that was legitimately deleted on disk, including:

* `config.providers[...]`;
* `config.customModels[...]`.

This issue now tracks that persistence/concurrency defect only.

## Reproduction

A representative sequence is:

1. Start the OpenCodex proxy so it holds configuration state in memory.
2. From another process, remove a provider or custom model using an official OpenCodex CLI/API mutation.
3. Verify the deletion is correctly persisted to:

```text
~/.opencodex/config.json
```

4. Without restarting the original long-running proxy, trigger an unrelated configuration mutation from that process, for example changing another Dashboard setting.
5. The long-running process persists its stale full configuration object.
6. Previously deleted providers or custom models reappear on disk.

The cooperating mutation machinery may still appear healthy:

* mutation generation advances;
* persistence succeeds;
* the catalog reconverges.

The problem is that the successfully committed document was based on stale state.

## Why this remains possible

OpenCodex has multiple configuration writers.

A long-running process may therefore have:

```text
memory = configuration loaded before another process changed disk
disk   = newer authoritative configuration
```

A later save can behave conceptually like:

```text
persist(memory)
```

instead of:

```text
apply this mutation to current disk state
```

That turns an unrelated setting change into a whole-document last-writer-wins operation.

The failure is not specific to `customModels`.

A stale writer can also resurrect the deleted provider itself:

```text
config.providers[P]
```

which makes any associated model rows legitimate again.

Therefore fixing only `customModels` reconciliation would not close the underlying bug.

## Existing partial reconciliation

The persistence path already contains special reconciliation for some state known to be changed by other processes.

For example, current persistence logic performs authoritative pre-write handling for selected configuration areas such as Claude configuration and server-binding state.

The remaining problem is that provider/model mutations do not have a general concurrency-safe persistence contract.

## Design constraints established during investigation

Several seemingly simple fixes are unsafe.

### Do not reconcile `customModels` as one whole array

A rule like:

```text
if disk changed and memory did not:
    take disk array
else:
    take memory array
```

loses independent concurrent edits.

For example:

```text
disk:
  deletes provider P

memory:
  independently edits model Q
```

Both sides changed legitimately.

Choosing either whole array destroys one mutation.

### Do not use routed model slug as row identity

This is not a stable identity:

```text
provider/modelId
```

Both components can change:

* provider rename changes `provider`;
* custom-model editing can change `modelId`.

Reconciling rows by routed slug can therefore interpret a rename as:

```text
delete old row
+ insert new row
```

and can produce dropped or duplicate state.

For custom models, the stable identity already exists:

```text
OcxCustomModel.id
```

That ID should remain the identity boundary if row-level reconciliation is used.

### Provider-map state must be handled too

Solving:

```text
customModels
```

without solving:

```text
providers
```

is insufficient.

A stale write that restores the provider can recreate the same effective ghost state.

## Preferred correctness property

An unrelated mutation must never revert a configuration change committed by another cooperating OpenCodex process.

For example:

```text
Process A:
  remove provider huggingface

Process B:
  change shadow-call enabled=false -> true
```

The resulting state must contain both effects:

```text
huggingface remains deleted
shadow-call becomes enabled
```

Process B must not restore the provider merely because it loaded configuration before Process A performed the deletion.

The same contract applies to:

* provider add/remove/rename;
* custom-model add/remove/edit;
* routing-profile changes;
* quota/settings changes;
* service-managed configuration writes;
* other cooperating configuration writers.

## Implementation directions

The implementation should establish an explicit persistence contract rather than adding another one-off array exception.

Two broad approaches are reasonable.

### Option A — field-scoped mutations

Writers declare which top-level configuration fields they intend to change.

Persistence then:

1. reads current disk state;
2. applies only the caller-owned mutation;
3. preserves unrelated newer disk state;
4. validates;
5. commits atomically.

Conceptually:

```text
mutateConfig(currentDisk, {
  shadowCallIntercept: ...
})
```

rather than:

```text
saveConfig(staleWholeObject)
```

This addresses the general class of stale whole-document writes but requires migrating the existing callers carefully.

### Option B — explicit three-way reconciliation

Track:

```text
baseline
current memory
current disk
```

and reconcile independently changed objects using stable identities.

For custom models this means:

```text
OcxCustomModel.id
```

not routed slugs.

Provider-map reconciliation would need an equally explicit identity/deletion contract.

Conflicts that cannot be resolved safely should fail visibly rather than silently choosing one whole document.

## Important writer inventory

Configuration persistence is not limited to Dashboard routes.

The maintainer audit found **16 files that actually call the save wrappers**, with additional files referencing them through imports or dependency-injection seams.

Writers exist across areas such as:

* management provider/model routes;
* routing profiles;
* storage policy;
* Codex routing;
* provider key failover;
* provider API-key handling;
* native integration paths.

A fix that depends only on state initialized by `startServer()` will therefore not automatically protect every writer.

The persistence guarantee must hold at the shared persistence boundary or be explicitly adopted by every mutation path.

## Expected behavior

Given:

```text
disk version N
```

Process A commits:

```text
delete provider P
```

producing:

```text
disk version N+1
```

Process B, whose in-memory configuration came from version N, then changes unrelated field Q.

The final configuration must be equivalent to:

```text
version N
+ delete P
+ change Q
```

and not:

```text
version N
+ change Q
```

which would resurrect P.

## Regression coverage

At minimum, cover:

### Provider deletion versus unrelated stale write

```text
A: delete provider P
B: stale process updates unrelated setting Q
```

Expected:

```text
P remains deleted
Q contains B's update
```

### Custom-model deletion versus unrelated stale write

```text
A: delete custom model M
B: stale process updates Q
```

Expected:

```text
M remains deleted
Q updated
```

### Independent custom-model edits

```text
A: edit model M1
B: edit model M2
```

Expected:

```text
both edits survive
```

### Rename

A provider/model rename must not become duplicate delete+insert state because reconciliation used a mutable routed slug as identity.

### Real process boundary

At least one regression should use separate processes or equivalent independent config snapshots rather than only mutating two object references inside one test.

The bug is specifically about cooperating writers with stale snapshots.

## Relationship to #1293

#1293 fixed the provider-removal lifecycle portion of the original report.

It does **not** fix this stale-writer persistence defect.

If a provider is removed with current `dev`, the removal path correctly cleans up its custom models.

If an older long-running configuration snapshot later resurrects that provider/model state, that is the remaining bug tracked here.

## Acceptance criteria

* [x] Provider removal cleans up provider-owned custom models.
* [x] CLI and management provider removal are covered.
* [x] `customModelCatalogMigration` survives provider cleanup correctly.
* [ ] A stale writer cannot resurrect a provider deleted by another process.
* [ ] A stale writer cannot resurrect a custom model deleted by another process.
* [ ] Independent concurrent mutations to different configuration state are preserved.
* [ ] Custom-model reconciliation, if used, keys rows by stable `OcxCustomModel.id`.
* [ ] Provider state receives an equivalent safe concurrency/deletion contract.
* [ ] Rename operations do not create duplicate rows.
* [ ] The fix applies across the real configuration-writer inventory rather than only one Dashboard route.
* [ ] Conflicting changes fail safely instead of silently discarding newer state.
* [ ] Cross-process or independent-snapshot regression coverage exists.

### Checks

* [x] The already-fixed provider-removal defect has been removed from remaining scope.
* [x] The outstanding stale whole-document persistence defect is isolated.
* [x] Previously rejected unsafe reconciliation approaches are documented.
* [ ] Concurrency-safe persistence contract implemented.

Yuxin-Qiao:
## Summary

- Fixes the stale whole-config writer boundary for #1273.
- Records a private per-object baseline at `loadConfig()`, so CLI, management, request-path, OAuth, routing, storage, and embedded writers can rebase independently loaded snapshots.
- Merges disjoint nested fields recursively and merges `customModels` by stable `OcxCustomModel.id`; provider/custom-model delete-vs-edit and same-leaf conflicts fail visibly with `CONFIG_WRITE_CONFLICT` before the atomic write.
- Keeps startup migrations explicit through `saveConfigDuringStartup()` before the live baseline is armed.
- Closes #1273.

## Verification

- `bun test tests/config-user-edits.test.ts tests/config-save-boundary.test.ts tests/config-mutation-lock.test.ts tests/user-cost-overlay-provider-delete.test.ts tests/user-cost-overlay-live-reconcile.test.ts tests/oauth-public-surface.test.ts` — 67 passed.
- `bun run test` was started after the rebase but did not produce a final result within the bounded local run and was stopped; no full-suite-green claim is made. The key-login live-update test is also environment-dependent here because the Umans hostname resolves to a benchmark address and the existing route rejects it with HTTP 400.
- Added the architecture Decision Log to `structure/02_config-and-codex-home.md`; no GUI or user-facing docs change is required.

## Checklist

## Review readiness checklist

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -2717,6 +2717,12 @@ export function websocketsEnabled(config: Pick<OcxConfig, "websockets">): boolea
  * against, or a later stale save would masquerade as "our own change".
  */
 const claudeCodeBaseline = new WeakMap<OcxConfig, unknown>();
+/**
+ * Full live-config baseline used to rebase unrelated cooperating writes. The
+ * Claude subtree and the bound listener fields remain on their dedicated
+ * reconciliation paths below.
+ */
+const liveConfigBaseline = new WeakMap<OcxConfig, OcxConfig>();
 
 /**
  * The live config retains the address of the socket Bun actually opened, while
@@ -2734,6 +2740,7 @@ const persistedLiveServerBinding = new WeakMap<OcxConfig, PersistedServerBinding
  * save, which is the case the guard exists for.
  */
 export function armClaudeCodeBaseline(config: OcxConfig): void {
+  liveConfigBaseline.set(config, structuredClone(config));
   claudeCodeBaseline.set(config, structuredClone(config.claudeCode));
 }
 
@@ -2781,6 +2788,54 @@ function cloneConfigValue(value: ConfigMergeValue): ConfigMergeValue {
   return value === MISSING_CONFIG_VALUE ? value : structuredClone(value);
 }
 
+type IndexedCustomModels = {
+  order: string[];
+  byId: Map<string, Record<string, unknown>>;
+};
+
+function indexCustomModels(value: ConfigMergeValue): IndexedCustomModels | null {
+  if (!Array.isArray(value)) return null;
+  const order: string[] = [];
+  const byId = new Map<string, Record<string, unknown>>();
+  for (const item of value) {
+    if (!isPlainConfigRecord(item) || typeof item.id !== "string" || item.id.length === 0 || byId.has(item.id)) {
+      return null;
+    }
+    order.push(item.id);
+    byId.set(item.id, item);
+  }
+  return { order, byId };
+}
+
+/**
+ * Merge custom-model rows by their stable id instead of treating the array as
+ * one opaque value. A row changed only on disk is adopted, a row changed only
+ * in the live config is retained, and disjoint edits to the same row recurse
+ * through the normal three-way object merge.
+ */
+function reconcileCustomModels(
+  baseline: ConfigMergeValue,
+  live: ConfigMergeValue,
+  persisted: ConfigMergeValue,
+): ConfigMergeValue | null {
+  const baselineRows = indexCustomModels(baseline);
+  const liveRows = indexCustomModels(live);
+  const persistedRows = indexCustomModels(persisted);
+  if (!baselineRows || !liveRows || !persistedRows) return null;
+
+  const order = [...liveRows.order, ...persistedRows.order.filter(id => !liveRows.byId.has(id))];
+  const merged: Array<Record<string, unknown>> = [];
+  for (const id of order) {
+    const row = reconcileConfigValue(
+      baselineRows.byId.get(id) ?? MISSING_CONFIG_VALUE,
+      liveRows.byId.get(id) ?? MISSING_CONFIG_VALUE,
+      persistedRows.byId.get(id) ?? MISSING_CONFIG_VALUE,
+    );
+    if (row !== MISSING_CONFIG_VALUE) merged.push(row as Record<string, unknown>);
+  }
+  return merged;
+}
+
 function reconcileConfigRecord(
   live: Record<string, unknown>,
   baseline: Record<string, unknown>,
@@ -2790,11 +2845,13 @@ function reconcileConfigRecord(
   const keys = new Set([...Object.keys(baseline), ...Object.keys(live), ...Object.keys(persisted)]);
   for (const key of keys) {
     if (skippedKeys?.has(key)) continue;
-    const merged = reconcileConfigValue(
-      ownConfigValue(baseline, key),
-      ownConfigValue(live, key),
-      ownConfigValue(persisted, key),
-    );
+    const baselineValue = ownConfigValue(baseline, key);
+    const liveValue = ownConfigValue(live, key);
+    const persistedValue = ownConfigValue(persisted, key);
+    const merged = key === "customModels"
+      ? reconcileCustomModels(baselineValue, liveValue, persistedValue)
+        ?? reconcileConfigValue(baselineValue, liveValue, persistedValue)
+      : reconcileConfigValue(baselineValue, liveValue, persistedValue);
     if (merged === MISSING_CONFIG_VALUE) delete live[key];
     else live[key] = merged;
   }
@@ -2920,13 +2977,12 @@ function readPersistedServerBinding(
  *
  * Conflict policy, chosen deliberately:
  * - disk changed, we did not → their hand edit wins;
- * - disk changed AND we changed → our change wins and the baseline rebases, so the
- *   user's next edit starts from the new value (a three-way merge is out of scope);
+ * - disk changed AND we changed → disjoint fields are merged, while a same-leaf
+ *   conflict keeps the live value;
  * - file missing/unreadable → save what we have, no throw.
  *
- * Scope residual: only `claudeCode` is reconciled. A hand edit to `providers` is still
- * clobbered — recorded and asserted in tests so it cannot drift into an assumed
- * guarantee.
+ * Custom-model rows are merged by their stable `id`, preserving independent
+ * edits and deletions across stale whole-config saves.
  */
 export function saveConfigPreservingClaudeCode(config: OcxConfig): void {
   withConfigMutationLockSync(() => {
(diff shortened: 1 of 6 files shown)


Ingwannu (changes requested):
The recursive merge and stable `customModels[].id` direction are valuable, but this head does not yet satisfy #1273 and must not close it.

Three blockers remain:

1. **The protection is still tied to one armed live-server object, not the shared persistence boundary.** `liveConfigBaseline` is populated only by `armClaudeCodeBaseline(config)`, whose sole production call is in `src/server/index.ts`. The PR deliberately preserves the unarmed whole-document behavior, while `saveConfig()` still writes a stale object directly and non-server/independently loaded configs can call either save wrapper without this baseline. That leaves the writer-inventory gap documented in #1273 intact. Establish provenance at the common mutation boundary (or explicitly migrate every whole-document writer), and add a real independent-snapshot/process regression through at least one non-server writer. A test that manually arms the same in-process object does not prove the required coverage.

2. **Concurrent deletion-versus-edit conflicts still silently resurrect state.** `reconcileConfigValue()` returns `live` whenever both sides changed and one side is `MISSING_CONFIG_VALUE`; therefore a newer disk deletion of `providers.P` or custom-model id `M` loses if the stale live object edited that same record. Same-leaf conflicts also silently prefer live. #1273 explicitly requires unresolvable conflicts to fail visibly rather than discard the newer committed state. Detect these delete/edit and same-leaf conflicts and abort the save without mutating disk (or provide another explicit, reviewable conflict contract). Add provider and same-id custom-model deletion/edit regressions.

3. **The architecture contract is undocumented.** This changes the concurrency semantics of the central config writer. Add the required `[Decision Log]` to the applicable `structure/` document: baseline provenance, recursive object semantics, stable-id array handling, deletion/tombstone behavior, same-leaf conflict behavior, which writers are covered, and why this was chosen over field-scoped `mutatePersistedConfig` migrations.

The current focused tests demonstrate disjoint merge behavior, but they do not cover the two boundaries above. After those are addressed, rebase onto current `dev`, keep #1273 open until the independent-writer acceptance passes, and run exact-head CI.

lidge-jun:
Landed on `dev` — closing as superseded rather than merged; `022cbd21e` carries your rebase-on-save with your authorship intact, and the branch now conflicts with what is there.

Your design is sound and the two tests you added are genuine regressions (32 pass / 2 fail against unfixed `dev`, 34 / 0 on your head). Two follow-ups landed on top, and the second one is worth your attention because it is unresolved between our two versions.

**`1b655b74c`** — the rebase protected rows that were untouched in memory, but a stale process that also *edited* a deleted row still resurrected it: a persisted `MISSING_CONFIG_VALUE` is not a plain record, so recursion was skipped and `live` won. Probed at your head: `{"providerResurrected":true,"customModelResurrected":true}`. An API key rotation in the dashboard was enough to trigger it. Persisted deletions now win for providers and `customModels`.

**`d6e4545b2`** — found by the full suite: `PUT /api/grok/selection` with an empty list stopped clearing `grokExcludedModels`. The live baseline is armed once, so a key another writer adds to the file afterwards is absent from both the baseline and the live config, and the merge reads that as "live never changed this" and adopts the disk value — resurrecting a field the route had just deleted.

Here is the part I could not resolve unilaterally. Your newer head `d51bd2856` (`reject stale whole-document conflicts`) is a better foundation — arming provenance on every load rather than only at server startup — but it still fails that Grok case (9 pass / 1 fail). And when I reimplemented our rule inside your design, it broke *your* test `an independently loaded config rebases a non-server save`, because a key added on disk after we loaded was then dropped.

Both rules read the same evidence — a key present on disk and absent in the live config — and reach opposite conclusions. Ours assumes deletion, yours assumes ignorance, and each is right about its own scenario. Distinguishing them properly needs deletion intent recorded where the deletion happens (there are eight or more `delete config.<field>` sites in the management routes), or the baseline refreshed on every cooperating write. That is a config-subsystem decision, not something to smuggle into a bug-fix batch, and you are the one working in this code.

I am filing the Grok case as its own issue against the landed design rather than forcing a version onto your branch. Would value your read on which direction to take.

PR #1465, opened 2026-08-11
fix(windows): preflight scheduler registration before teardown

### LINKED ISSUE #1454: Windows: failed ocx service install leaves proxy stopped and Codex routing un-injected because teardown runs before the elevation check
### Client or integration

Codex CLI

### Area

Service lifecycle

### Summary

On Windows, `ocx service install` stops the proxy and removes opencodex's injected Codex config keys **before** it discovers it needs elevation. When the `schtasks` call is then denied, the teardown is not undone — so a failed install leaves a previously working setup broken.

Observed sequence:

1. `ocx service install` stops the running proxy and restores native Codex routing (removing `openai_base_url` and `model_catalog_json` from `<CODEX_HOME>/config.toml`).
2. It then runs `schtasks /create ...`, which fails because the shell is not elevated:

```
❌ Service install cleanup failed: Windows access denied while running Task Scheduler.
Command: schtasks /create /tn opencodex-proxy /xml <OPENCODEX_HOME>\opencodex-service-task.xml /f
Approve the Windows UAC prompt to install the background service, or run `ocx service install`
from an elevated PowerShell window.
OCX_ERROR_CODE=WINDOWS_SCHTASKS_CREATE_ACCESS_DENIED
```

3. The service is not installed **and** the proxy is left stopped with routing removed. `ocx status` afterwards reports `Restart safety: native Codex routing (no opencodex restart dependency)`.

Expected: elevation is checked (or the task creation attempted) *before* any teardown, so that a permission failure is a no-op; or the teardown is rolled back when install fails.

The message text does anticipate the failure ("Approve the Windows UAC prompt… or run from an elevated PowerShell window"), so the elevation requirement is known in advance — it just is not checked early enough.

### Reproduction

100% reproducible on my machine.

1. Have a working opencodex install with the proxy running and Codex routed through it.
2. From a **non-elevated** shell, run `ocx service install`.
3. Observe the access-denied error above.
4. Check `<CODEX_HOME>/config.toml` — `openai_base_url` and `model_catalog_json` are gone.
5. Check the proxy — stopped.

Recovering is harder than it should be because `ocx restore back` and `ocx sync` both refuse afterwards with a message naming no reason (filed separately). I restored routing by hand using the keys in `<CODEX_HOME>/opencodex.config.toml`.

### Suggested fixes

1. **Probe for elevation (or attempt the `schtasks` create) before stopping the proxy and un-injecting the config.** A permissions failure should leave the system exactly as it was found.
2. Failing that, make the cleanup path actually restore what the install path removed — the current message says "Service install cleanup failed", so a cleanup step exists but does not cover the config/proxy teardown.
3. Consider detecting non-elevated invocation up front and printing the "run from an elevated PowerShell window" guidance *instead of* starting the install at all.

### Version

2.11.0 (the install path is unchanged in 2.12.0 as far as I can tell from the release notes)

### Operating system

Windows 11 Pro 10.0.26200

### Provider and model

Not provider-specific.

### Logs or error output

```shell
$ ocx health --json
{"ok":true,"pid":19940,"port":10100}

$ ocx service install
❌ Service install cleanup failed: Windows access denied while running Task Scheduler.
Command: schtasks /create /tn opencodex-proxy /xml <OPENCODEX_HOME>\opencodex-service-task.xml /f
Approve the Windows UAC prompt to install the background service, or run `ocx service install`
from an elevated PowerShell window.
OCX_ERROR_CODE=WINDOWS_SCHTASKS_CREATE_ACCESS_DENIED

$ ocx health --json
{"ok":false,"pid":null,"port":null}

$ ocx service status
❌ service not installed (Task Scheduler).

$ ocx status
❌ Proxy: not running
   Restart safety: native Codex routing (no opencodex restart dependency)
   Service: not installed

# and the injected keys are gone from <CODEX_HOME>/config.toml:
#   openai_base_url      -> removed
#   model_catalog_json   -> removed
```

_No response_

### Redacted configuration

```json
{
  "providers": {
    "openai": { "adapter": "openai-responses", "codexAccountMode": "direct", "authMode": "forward" }
  },
  "codexAutoStart": true,
}
```

Windows sandbox mode is `unelevated` in the Codex config; the shell used was a normal (non-admin) terminal.

### Checks

Ingwannu:
## Summary

- Move fresh Windows Task Scheduler registration ahead of destructive service/proxy cleanup.
- Register but do not run the owned task from a unique temporary XML, then verify the live scheduler XML before stopping the current proxy or removing managed Codex routing.
- Roll back only the registration created by this attempt when the commit phase fails before task start; leave started partial state explicit for diagnosis.
- Keep existing or conflicting scheduler registrations on the conservative fail-closed path, and keep the Dashboard install request alive while Windows UAC is pending so an elevated child cannot be orphaned by the ordinary 60-second timeout.
- Document the two-phase lifecycle and its tradeoffs.

This fixes the real ordering regression reported in #1454: cancelling or denying UAC during a fresh scheduler install is now a pre-commit failure and leaves the working proxy/routing untouched.

Closes #1454

## Verification

- Exact base: `origin/dev` at `87e3ff9f6`.
- `taskset -c 0-1 nice -n 10 bun test` across 18 Windows service/UAC and adjacent CLI files: **406 pass, 0 fail**.
- `taskset -c 0-1 nice -n 10 bun run typecheck`: passed.
- `taskset -c 0-1 nice -n 10 bun run privacy:scan`: passed.
- `cd docs-site && bun install --frozen-lockfile`: no changes.
- `cd docs-site && taskset -c 0-1 nice -n 10 bun run build`: **221 pages built**.
- Full suite before the final upstream rebase: **10,947 pass, 11 skip, 1 fail**. The one failure, `Codex autostart shim > Unix shim exports persisted service API token before running Codex`, reproduces on clean `dev` and is unrelated to this Windows scheduler change. After rebasing, the exact-head focused suite, typecheck, privacy scan, and docs build above were rerun successfully.

## Checklist

Because this changes Windows elevation and service lifecycle boundaries, please keep the PR in draft until an independent maintainer completes the required security review and Windows smoke test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows service installation safety by registering and verifying the scheduler task before stopping an existing proxy or replacing service assets.
  * Cancelled or denied elevation now preserves the active proxy and routing.
  * Windows installation no longer times out during UAC approval or extended setup.
  * Failed registrations automatically roll back temporary changes and clean up staged files.

### DIFF
--- a/docs-site/src/content/docs/reference/cli/lifecycle.md
+++ b/docs-site/src/content/docs/reference/cli/lifecycle.md
@@ -282,6 +282,13 @@ automatically. If that fallback cannot determine the token state, it retains the
 error. Foreign tasks and operations can never emit the automatic-elevation marker. Approve the
 dashboard UAC prompt or rerun `ocx service install` in an elevated PowerShell window.
 
+For a fresh install where the OpenCodex scheduler task is confirmed absent, UAC approval now
+happens before the installer stops any existing proxy. The task is registered without being run;
+only after registration succeeds does OpenCodex stop the old listener, publish the service assets,
+and start the scheduled task. Cancelling or denying UAC therefore leaves the working proxy and its
+Codex routing in place. Existing or conflicting scheduler registrations continue to fail closed
+rather than being deleted as an unsafe best-effort rollback.
+
 ### `ocx codex-shim <install|status|uninstall|remove>`
 
 Wrap a script-based `codex` launcher on PATH with a lightweight autostart script. Real `codex.exe`
--- a/src/server/startup-action-control.ts
+++ b/src/server/startup-action-control.ts
@@ -164,10 +164,17 @@ function runCliInstall(
   const cli = join(import.meta.dir, "..", "cli", "index.ts");
   const argv = [cli, ...startupInstallArgv(action, options)];
   return new Promise((resolve, reject) => {
+    const timeout = process.platform === "win32" && action === "install-service"
+      ? 0
+      : 60_000;
     execFile(bun, argv, {
       encoding: "utf8",
       env: process.env,
-      timeout: 60_000,
+      // A fresh Windows scheduler install now owns its UAC prompt inside this CLI
+      // transaction. Killing only the CLI at 60s can orphan its elevated schtasks child,
+      // which may register the task after the Dashboard has reported failure. Keep the
+      // async request/attempt lock alive until Windows returns approval or cancellation.
+      timeout,
       windowsHide: true,
       maxBuffer: 256 * 1024,
     }, (error, stdout, stderr) => {
--- a/src/service.ts
+++ b/src/service.ts
@@ -29,6 +29,7 @@ import {
   runWindowsElevated,
   toWindowsSchtasksError,
   WindowsElevationError,
+  WindowsSchtasksError,
   type ElevatedSchedulerOutcome,
   type ElevatedSchtasksCreateAndRunExecution,
   type ElevatedSchtasksCreateAndRunResult,
@@ -1506,6 +1507,11 @@ export function buildWindowsSchtasksCreateArgs(script = windowsServiceScriptPath
   return ["/create", "/tn", TASK, "/xml", xml, "/f"];
 }
 
+/** Build the fixed scheduler-create command from an explicit staged XML document. */
+export function buildWindowsSchtasksCreateArgsForXml(xml: string): string[] {
+  return ["/create", "/tn", TASK, "/xml", xml, "/f"];
+}
+
 /**
  * VBS launcher that starts the batch wrapper with a hidden window (style 0).
  * bWaitOnReturn=True keeps wscript.exe resident for the wrapper's lifetime so the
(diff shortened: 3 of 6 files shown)


lidge-jun:

On `dev` as `1fa020a90` with your authorship intact, plus `6312aef83`.

The two-phase transaction audits clean: phase 1 stages and registers only, and nothing in that path stops a manager or the tracked proxy (`prepareServiceInstall()` runs afterwards), writes a canonical asset (only the UUID-named ephemeral XML), mutates Codex routing, or publishes install state. A phase-2 failure is rethrown and exits non-zero rather than being reported as success. That is exactly the ordering #1454 asked for.

The follow-up closes a gap in the cleanup claim. Rollback deleted the scheduler task by name, and the absence probe is not atomic with the fixed-name `schtasks /create /f`, so a concurrent registration could be deleted without this attempt ever proving it owned the task. An attempt nonce now rides in `RegistrationInfo/Description`, is verified immediately before deletion, and where ownership cannot be proven the CLI reports the residual scheduler state instead of claiming the prior runtime was restored.

Stated honestly: the query and the elevated delete are still not one atomic operation, so a replacement registered in that window can be deleted. Closing that needs an attempt-unique task name or an elevated attempt-bound transaction, which changes how the service registers on Windows and needs a real Windows host to validate — `windows N/4` is dispatch-only per #1059. Recorded rather than papered over.

Verified: `tests/service.test.ts` + `tests/startup-action-control-elevation.test.ts` 137 pass / 0 fail, with the ownership regressions red before the fix.

PR #1464, opened 2026-08-11
fix(anthropic): forward structured output schemas

Lqm1:
## Summary

- Forward Responses and Chat Completions JSON Schema output formats through the Anthropic adapter as `output_config.format`.
- Normalize JSON Schema to Anthropic's supported structured-output subset while preserving unsupported constraints in descriptions.
- Preserve root `$defs` beside a root `$ref`, merge structured-output configuration with adaptive-thinking `output_config.effort`, and round-trip Messages structured output through stored OAuth.
- Document the Anthropic compatibility transform, limitations, and maintenance decision.

## Verification

- Rebased onto current `dev@4fed8d3f`.
- `bun test tests/anthropic-reasoning.test.ts tests/claude-inbound.test.ts tests/claude-messages-endpoint.test.ts tests/responses-parser.test.ts tests/responses-parser-malformed-content.test.ts tests/chat-completions-endpoint.test.ts tests/responses-compaction.test.ts tests/responses-compaction-routing.test.ts` (269 passed; 2 optional-main enrichment tests fail identically on current `dev` and are unrelated to structured output)
- `bun test tests/api-storage-policy-put-race.test.ts` (1 passed)
- `bun run typecheck`
- `bun run privacy:scan`
- `git diff --check upstream/dev...HEAD`
- `cd docs-site && bun install --frozen-lockfile && bun run build`
- Live AI SDK v7 `Output.object` request through the patched server on port 10101 using `anthropic/claude-sonnet-5`
- The previous exact-head Cross-platform CI run passed all Linux shards, the unsharded macOS suite, gates, keyring, storage-policy, API-usage, and npm-global jobs; the rebased head is awaiting maintainer approval for the fork workflow
- The unsharded local Windows suite remains affected by the repository's known Windows-only failures and Bun 1.3.14 panics; the related focused suites are green apart from the two current-`dev` optional-main enrichment failures noted above

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults.

## Review readiness checklist

This PR stays in draft until every box below is ticked. Tick all four boxes once the requirements are met:

- [x] All CI tests are green on my local testing.
- [x] I pushed my PR to the latest dev commit.
- [x] I resolved all correct Codex and CodeRabbit findings.

- [x] My PR is ready for review.

## Summary by CodeRabbit

* **New Features**
[...]

PR #1471, opened 2026-08-11
fix(auth): pin forwarded Codex credentials to ChatGPT

luvs01:
## Summary

- forward caller OAuth/Codex metadata and runtime pool credentials only to the canonical ChatGPT Codex endpoint
- prevent noncanonical Responses gateways from receiving caller bearer tokens, ChatGPT account/session identifiers, or runtime account overrides
- pin accepted canonical URL variants before Responses, web-search, vision, image, live, and native-compact requests are constructed
- preserve explicitly configured static provider headers and ordinary canonical forward behavior
- require a Codex pool account only for canonical credential-bearing destinations, so a noncanonical gateway can continue with its configured URL and static headers

Previously, every `openai-responses` provider using `authMode: "forward"` copied `FORWARD_HEADERS` and any `_codexAccountOverride` into its upstream request. A custom or changed `baseUrl` could therefore become a credential destination even though it was not the canonical ChatGPT backend.

The adapter now uses the existing canonical-provider predicate as the authorization boundary for dynamic credentials and the pool-required invariant. Canonical trailing-slash spellings are pinned to the exact ChatGPT base before credential-bearing sidecars and compact/live paths append endpoints. Noncanonical gateways still receive their explicitly configured static headers, but never inherit caller or pool-account credentials.

## Verification

- Bun 1.4.0-canary.1: the four exact maintainer-requested credential/native-metadata regressions passed (4/4, 17 assertions)
- Bun 1.3.14: the same four exact regressions passed (4/4, 17 assertions)
- Bun 1.3.14: full passthrough, custom-tool, and override adapter files passed (85/85, 302 assertions)
- Bun 1.4.0-canary.1: broader three-file run passed every changed regression, including canonical custom tools and synthesized `session_id`; three unrelated existing Windows ACL/teardown cases failed and are recorded separately rather than treated as PR regressions
- Bun 1.4.0-canary.1: TypeScript 7.0.2 `tsc --noEmit` passed on the exact latest-dev head
- Bun 1.3.14: TypeScript 7.0.2 `tsc --noEmit` passed on the exact latest-dev head
- `git diff --check` passed
- latest-dev rebase preserved the exact validated patch-id
- independent read-only production and test reviews found no remaining P0-P3 findings

Focused affected-surface tests, dual-runtime typecheck, privacy, diff, and independent review satisfy the local CI gate. Unrelated environment failures are not attributed to this change, and the same GitHub CI workload is not duplicated locally.

## Checklist

- [x] Docs or release notes were updated when needed. This tightens an existing internal credential-destination invariant without changing user-facing configuration.
- [x] Security-sensitive changes were reviewed for credential destinations, runtime overrides, canonical URL normalization, sidecar routing, and static-header compatibility.

## Review readiness checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved request forwarding for supported OpenAI-compatible connections.
  - Standardized endpoint routing by removing extra trailing slashes.
  - Preserved configured URLs and credentials for nonstandard providers.
  - Improved authentication and account handling for supported Codex connections.
  - Fixed live, compact, vision, image, and web-search requests to reach the correct endpoints.

- **Tests**
  - Added coverage for URL normalization, credential handling, authentication, and request routing across supported services.

### DIFF
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -6,7 +6,7 @@ import { COMPACT_PROMPT, decodeCompactionSummary, SUMMARY_PREFIX } from "../resp
 import { collectResponsesToolGroups } from "../responses/tool-groups";
 import { isHostedToolUnsupportedForModel } from "../responses/hosted-tool-policy";
 import { decodeServerSentEvents } from "../lib/sse-decoder";
-import { isCanonicalOpenAiForwardProvider } from "../providers/openai-tiers";
+import { CODEX_FORWARD_BASE_URL, isCanonicalOpenAiForwardProvider } from "../providers/openai-tiers";
 import { OCX_REASONING_PREFIX } from "../responses/reasoning-envelope";
 import { modelRecordValue } from "../reasoning-effort";
 import type { TranslatorBudget } from "../lib/translator-budget";
@@ -1213,22 +1213,32 @@ export function createResponsesPassthroughAdapter(provider: OcxProviderConfig): 
       let url: string;
 
       if (provider.authMode === "forward") {
+        const mayForwardCallerCredentials = isCanonicalOpenAiForwardProvider(provider);
         // OAuth passthrough: ChatGPT backend path is `${baseUrl}/responses` (no /v1).
-        url = `${provider.baseUrl}/responses`;
+        const baseUrl = mayForwardCallerCredentials
+          ? CODEX_FORWARD_BASE_URL
+          : provider.baseUrl.replace(/\/+$/, "");
+        url = `${baseUrl}/responses`;
         if (provider.headers) Object.assign(headers, provider.headers); // static headers first…
         const runtimeProvider = provider as {
           _codexAccountOverride?: { accessToken: string; chatgptAccountId: string };
           _codexAccountRequired?: boolean;
         };
-        if (runtimeProvider._codexAccountRequired && !runtimeProvider._codexAccountOverride) {
+        if (
+          mayForwardCallerCredentials
+          && runtimeProvider._codexAccountRequired
+          && !runtimeProvider._codexAccountOverride
+        ) {
           throw new Error("Codex pool account auth is required but unavailable");
         }
-        for (const h of FORWARD_HEADERS) {
-          const v = incoming?.headers.get(h);
-          if (v) headers[h] = v;                                        // …so forwarded auth always wins.
+        if (mayForwardCallerCredentials) {
+          for (const h of FORWARD_HEADERS) {
+            const v = incoming?.headers.get(h);
+            if (v) headers[h] = v;                                      // …so forwarded auth always wins.
+          }
         }
         const override = runtimeProvider._codexAccountOverride;
-        if (override) {
+        if (override && mayForwardCallerCredentials) {
           headers["authorization"] = `Bearer ${override.accessToken}`;
           headers["chatgpt-account-id"] = override.chatgptAccountId;
         }
--- a/src/providers/openai-sidecar.ts
+++ b/src/providers/openai-sidecar.ts
@@ -13,6 +13,7 @@ import { extractAccountId } from "../oauth/chatgpt";
 import { ForwardAdmissionCredentialError, validateForwardAdmissionCredential } from "../server/auth-cors";
 import type { CodexAccountMode, OcxConfig, OcxProviderConfig } from "../types";
 import {
+  CODEX_FORWARD_BASE_URL,
   isCanonicalOpenAiForwardProvider,
   OPENAI_API_PROVIDER_ID,
   OPENAI_CODEX_PROVIDER_ID,
@@ -62,10 +63,16 @@ export function listOpenAiForwardSidecarCandidates(config: OcxConfig): OpenAiFor
     ? { ...provider, authMode: "forward" as const }
     : provider;
   if (!isCanonicalOpenAiForwardProvider(canonicalProvider)) return [];
+  // The predicate accepts harmless trailing-slash variants. Pin the provider returned
+  // to credential-bearing sidecars so every consumer builds one exact ChatGPT path
+  // instead of independently concatenating the operator's equivalent spelling.
+  const pinnedProvider = canonicalProvider.baseUrl === CODEX_FORWARD_BASE_URL
+    ? canonicalProvider
+    : { ...canonicalProvider, baseUrl: CODEX_FORWARD_BASE_URL };
   return [{
     providerName: OPENAI_CODEX_PROVIDER_ID,
-    provider: canonicalProvider,
-    accountMode: providerCodexAccountMode(OPENAI_CODEX_PROVIDER_ID, canonicalProvider) ?? "pool",
+    provider: pinnedProvider,
+    accountMode: providerCodexAccountMode(OPENAI_CODEX_PROVIDER_ID, pinnedProvider) ?? "pool",
   }];
 }
 
--- a/src/server/images.ts
+++ b/src/server/images.ts
@@ -499,6 +499,12 @@ export async function handleImages(
       headers,
       body: JSON.stringify(body),
       signal: linkedSignal.signal,
+      // Do not follow a cross-origin 3xx while carrying Codex credentials. Bun strips
+      // `Authorization` across origins but forwards nonstandard headers, so
+      // `chatgpt-account-id`, `session_id`, and `x-codex-turn-metadata` would reach the
+      // redirect target. Verified with a two-server probe. The Responses path and native
+      // compact already set this; the credential-bearing sidecars did not.
+      redirect: "manual",
     });
     const observed = await readImageResponseBytes(upstreamResponse, {
       maxBytes: IMAGES_RESPONSE_MAX_BYTES,
--- a/src/server/live.ts
+++ b/src/server/live.ts
@@ -223,7 +223,7 @@ export function keyedLiveUrl(baseUrl: string): string {
 }
 
 export function forwardLiveUrl(baseUrl: string, usesBackendShape: boolean): string {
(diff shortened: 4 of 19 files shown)


Wibias (commented):
Reviewed the credential-destination changes. I did not find an additional blocker in this head. The canonical-provider predicate rejects userinfo/query/hash/noncanonical destinations, dynamic caller and pool credentials are gated on that predicate, and canonical trailing-slash variants are pinned before credential-bearing endpoint construction.

I am not approving this head because GitHub currently reports it as not mergeable against the latest `dev`. Please rebase/resolve the conflicts and request re-review so the final credential boundary can be checked on the integrated diff.

Wibias (changes requested):
Requesting changes for two regressions plus one remaining credential-boundary bug:

1. **The PR currently breaks existing Responses behavior.** The suite reports:
   - `(fail) routed Responses custom-tool compatibility > handleResponses leaves custom tools native for forward-auth passthrough`
   - `(fail) native openai-responses route carries prompt_cache_key + synthesized session_id header`

   These failures are in code paths changed by this PR's canonical/noncanonical forward-provider classification. Please update the implementation and/or now-obsolete fixtures as appropriate, but the full affected behavior needs to remain covered and green before merge. In particular, make sure legitimate canonical native Responses requests still preserve their expected metadata/header behavior while noncanonical gateways do not inherit caller credentials.

2. **Major: gate `_codexAccountRequired` on the canonical credential destination.** In `src/adapters/openai-responses.ts`, the current condition throws for *any* `authMode: "forward"` provider with `_codexAccountRequired: true` and no `_codexAccountOverride`:

```ts
if (runtimeProvider._codexAccountRequired && !runtimeProvider._codexAccountOverride) {
  throw new Error("Codex pool account auth is required but unavailable");
}
```

That still includes noncanonical providers. It conflicts with this PR's objective that noncanonical gateways continue operating with their configured URL and static headers, without caller or pool credentials.

Please require `mayForwardCallerCredentials` in this condition, e.g. conceptually:

```ts
if (
  mayForwardCallerCredentials &&
  runtimeProvider._codexAccountRequired &&
  !runtimeProvider._codexAccountOverride
) {
  throw new Error("Codex pool account auth is required but unavailable");
}
```

Add a regression case with a **noncanonical** provider, `_codexAccountRequired: true`, no `_codexAccountOverride`, and static configured headers. It should build/send against the configured noncanonical URL without requiring a Codex pool account and without forwarding caller credentials.

Ingwannu:
Re-review on exact head `d2bd0f0b2c3438184d83c28a08651bba94e2725b` against current `dev@37139818d2d3fe7b033e958ccf87080e4be2b4b7`.

The blockers from the previous review are addressed on this head:

- `_codexAccountRequired` is now enforced only when the resolved provider is the canonical ChatGPT Codex credential destination;
- noncanonical forward gateways retain only their configured URL and static headers, without caller/pool credentials or runtime account overrides;
- canonical trailing-slash variants are pinned before Responses, compact, live, image, vision, and web-search endpoint construction;
- the two previously regressed canonical paths now have updated coverage for native custom tools and synthesized `session_id`/prompt-cache metadata.

I traced the destination predicate through the adapter, sidecar, compact, and live paths and found no additional credential-boundary blocker. Exact-head validation under a 2-CPU/8-GB cap:

- affected paths excluding the known flaky Claude replay case: **281 passed, 0 failed**;

`tests/claude-messages-endpoint.test.ts` still has the existing `Claude replay owns optional main enrichment...` timing/state failure (**39 passed, 1 failed**) when run alone; the same failure reproduces on the current `dev` baseline and is unrelated to this diff.

I am not approving or merging while the PR remains draft with the author's local-CI and ready-for-review boxes open. From the credential-destination and affected-path review, I found no remaining code blocker at this head; please complete the readiness/CI gate and request final review without changing the head.

Wibias (changes requested):
Re-review on current head `787db40d9411388b00c24a634f754e121a1f643c`.

The earlier credential-boundary blockers are fixed on this head. Dynamic caller/pool credentials are gated on the canonical ChatGPT Codex destination, canonical trailing-slash variants are pinned before endpoint construction, `_codexAccountRequired` is now gated on that same canonical boundary, and the previously failing canonical metadata/custom-tool regressions have been updated. I did not find a new code blocker in the current diff. The later CodeRabbit batch-normalization note is outside this PR's behavioral change and comes from base history, not this credential patch.

The remaining blocker is integration. This head is now 10 commits behind current `dev@70d2e175` after #1464 merged, with overlap in `tests/claude-messages-endpoint.test.ts`. Exact-head Cross-platform CI and React Doctor are `action_required`, and the Cross-platform run executed zero jobs.

Please rebase onto current dev and obtain an executed exact-head CI signal, then request re-review. No additional credential-scope changes are requested from my review.

Wibias (changes requested):
Re-review on current head `fa87583f243e3ae5bac073572a81a76f897b713c` against current `dev@bd5bc566d33f37682ffbf636e909243ba388caca`.

The production credential-boundary changes still look correct. I re-checked the security-sensitive path rather than relying on the previous review:

- `isCanonicalOpenAiForwardProvider()` rejects userinfo, query, fragment, wrong origin/path, and only accepts the normalized canonical ChatGPT Codex destination.
- Caller `FORWARD_HEADERS`, `_codexAccountOverride`, and the `_codexAccountRequired` invariant are all gated on that canonical destination.
- Canonical trailing-slash variants are pinned to `CODEX_FORWARD_BASE_URL` before credential-bearing Responses/sidecar/compact endpoint construction.
- Noncanonical forward providers keep only their explicitly configured static headers and URL; caller and runtime Codex credentials are not injected.
- Native compact only resolves/forwards Codex account credentials inside `supportsNativeResponsesCompactEndpoint()`, so arbitrary Responses-shaped gateways do not reach that credential-bearing branch.

I did not find a new production auth/security blocker in the current patch.

The new unresolved CodeRabbit thread in `tests/claude-messages-endpoint.test.ts` is valid test-harness hardening: the fetch shim should fail closed for unexpected `https://chatgpt.com/backend-api/codex/*` paths instead of falling through to the real network. I do not treat the current fixture as evidence of a production credential leak—the test uses an isolated `CODEX_HOME` and this request does not carry a real bearer—but the regression harness should still never permit accidental external ChatGPT requests.

Two merge blockers remain:

1. **Rebase/integration:** this head is `ahead 1 / behind 10` current `dev`, with merge-base `d6673675`. Since that base, `dev` changed overlapping/adjacent surfaces including `src/server/responses/compact.ts`, `tests/claude-messages-endpoint.test.ts`, `tests/responses-compaction-routing.test.ts`, and `tests/web-search.test.ts`. Please rebase and preserve both this credential boundary and the newer dev behavior.
2. **Exact-head CI is not green:** React Doctor run `31566300187` succeeded. Cross-platform run `31566300191` executed substantially and the gates/typecheck/privacy plus all four Linux test shards succeeded, but the workflow ultimately concluded `cancelled`; macOS and npm-global Windows were cancelled and the final `ci` aggregation job failed because not every required job completed successfully. Please obtain a complete green run on the rebased head.

The older batch-normalization CodeRabbit note remains outside this PR's credential change and is not a blocker from my review.

lidge-jun:
## Review: LAND (after one fix folded in on top)

Verified against the current `dev` tree rather than the description, with an independent reviewer pass. Contributor commits `c5844b5f5` and `024949b03` were rebased onto `dev` with **byte-identical content** (verified by diffing the normalized hunks against the previous head `875a694f4`), and my fix is a separate commit on top.

### What the PR gets right

Pinning the forwarded Codex credentials to the canonical ChatGPT host is the correct shape of defense. The fail-closed path check in the follow-up commit closes the obvious "unexpected path" bypass instead of allow-listing, which is the right default for a credential boundary.

### The gap I found and fixed

Host pinning closes the *configured target*, but it does not close the *redirect*. Bun follows cross-origin 3xx responses, and while it drops `Authorization`, it forwards nonstandard headers. Two-server probe:

```
REDIRPROBE: {"authorization":null,"chatgpt-account-id":"acct-123","session_id":"sess-456","x-codex-turn-metadata":"turn-789"}
```

So a request pinned to the canonical host that receives a redirect would still hand the account id, session id, and turn metadata to the redirect target. `Authorization` stripping alone is a false sense of safety here.

Fix: `redirect: "manual"` on the five credential-bearing sidecar fetches.

- `src/server/images.ts`
- `src/server/live.ts`
- `src/server/search.ts`
- `src/web-search/executor.ts`
- `src/vision/describe.ts`

The Responses and compact paths already reach the same policy through a different mechanism — `fetchWithHeaderTimeout(manualRedirect)` applies it centrally (#914). `tests/credential-redirect-guard.test.ts` pins **both** families: a runtime probe that proves the forwarding behavior is real (so the test fails loudly if Bun ever changes it), and assertions that neither family silently drops the flag.

### One behavior change worth calling out

A noncanonical gateway configured with `authMode: "forward"` no longer receives implicit caller or pool credentials; it must be configured with static provider credentials. That is a compatibility break, and it is intended — implicitly forwarding ChatGPT-scoped credentials to an arbitrary configured host is the thing being fixed. Recording it here so it is not a surprise at upgrade time.

### Verification

Run on the Linux CI host, Bun 1.3.14, at the pushed head.

| Check | Result |
|---|---|
| `bun x tsc --noEmit` | exit 0 |
| `bun test tests/credential-redirect-guard.test.ts` | 7 pass / 0 fail |
| Focused suites (7 files, incl. responses-passthrough, web-search, compaction-routing, server-live, server-images, vision-sidecar-e2e) | 265 pass / 0 fail |
| `bun run privacy:scan` | passed |

Red-before was confirmed: with only the test file applied to `origin/dev`, the guard runs **2 pass / 5 fail** — one failure per unguarded sidecar. With the fix it is 7/7.

PR #1474, opened 2026-08-11
fix(responses): reject unscoped reasoning replay

luvs01:
## Summary

- require an explicit client thread scope before the in-process reasoning replay cache can read or write raw reasoning
- remove the process-wide `"global"` fallback from the bridge, OpenAI chat adapter, Responses core, image loop, and web-search loop
- keep legitimate same-thread DeepSeek reasoning replay and placeholder behavior while preventing unscoped cross-request call-ID collisions
- add stream, batch, cross-request, thread-isolation, and production call-site regression coverage

## Root cause and impact

Provider-generated tool call IDs such as `call_1` are not globally unique. The existing optional scope still fell back to a process-wide namespace whenever `x-codex-parent-thread-id` was absent, so an unrelated unscoped continuation could recover another request's cached raw reasoning for up to the cache TTL and forward it as `reasoning_content`.

There is no safe stable conversation identity for generic unscoped clients. This change therefore fails closed on cache read/write when the explicit client thread scope is absent. Preserve-listed DeepSeek models continue to use the existing minimal placeholder on a cache miss, so an unscoped request remains valid without replaying unrelated reasoning.

Refs #950.

## Verification

- Bun 1.4.0-canary.1: 95 focused tests passed, 0 failed, 261 assertions
- Bun 1.3.14: 95 focused tests passed, 0 failed, 261 assertions
- Bun 1.4.0-canary.1: TypeScript 7.0.2 `tsc --noEmit` passed
- Bun 1.3.14: TypeScript 7.0.2 `tsc --noEmit` passed
- `git diff --check` passed
- two independent read-only production/test reviews found no remaining P0-P3 findings

## Checklist

- [x] Docs or release notes were updated when needed. No user-facing configuration or workflow changes.
- [x] Security-sensitive changes were reviewed for secrets, auth boundaries, unsafe defaults, and privacy.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reasoning replay isolation between conversations.
  * Prevented unscoped or empty replay data from being stored or retrieved.
  * Reduced the risk of replay collisions across streaming, image, web-search, and standard responses.
* **Tests**
  * Added coverage for scoped replay behavior, conversation isolation, and unscoped requests.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -226,7 +226,7 @@ function toolResultImageChatParts(content: string | OcxContentPart[]): unknown[]
 function messagesToChatFormat(parsed: OcxParsedRequest, provider: OcxProviderConfig): unknown[] {
   const out: unknown[] = [];
   const { context, options } = parsed;
-  const replayCacheScope = parsed._clientThreadId ?? "global";
+  const replayCacheScope = parsed._clientThreadId;
 
   interface PendingToolCall { id: string; name: string }
   let pendingToolCalls: PendingToolCall[] = [];
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -197,7 +197,7 @@ export function bridgeToResponsesSSE(
     };
   },
 ): ReadableStream<Uint8Array> {
-  const replayCacheScope = options?.replayCacheScope ?? "global";
+  const replayCacheScope = options?.replayCacheScope;
   const setBeatInterval = options?.timers?.setInterval ?? ((handler: () => void, ms: number) => setInterval(handler, ms));
   const clearBeatInterval = options?.timers?.clearInterval ?? ((id: unknown) => clearInterval(id as ReturnType<typeof setInterval>));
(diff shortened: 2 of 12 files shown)


Wibias (commented):
Reviewed the replay-cache privacy change. I did not find an additional blocker in this head. Removing the implicit process-wide `global` namespace at every read/write call site and making unscoped cache access a no-op closes the demonstrated cross-request call-id collision without inventing a replacement identity.

I am not approving this head because it is currently not mergeable against the latest `dev`. Please rebase/resolve conflicts and request a short re-review of the integrated cache call sites.

Wibias (approved):

Wibias:
Thanks @luvs01 for tightening this up. This is useful because provider-generated tool call IDs are not globally unique, so removing the unscoped global reasoning replay namespace prevents reasoning from one conversation being recovered by an unrelated request with the same call ID. Keeping replay only when an explicit thread scope exists gives us the compatibility benefit without the cross-request privacy risk. Merged, thank you!

PR #1479, opened 2026-08-11
fix(deepseek): preserve parallel reasoning replay

### LINKED ISSUE #1477: [Bug]: DeepSeek V4 Flash Responses continuation still drops reasoning_text on 2.12.0
## Client or integration

Codex Desktop using OpenCodex as the local Responses API proxy.

## Summary

On OpenCodex 2.12.0, the official DeepSeek V4 Flash Responses route can still fail after tool calls with:

```text
The `reasoning_text` in the thinking mode must be passed back to the API.
```

This reproduces even though the built-in `deepseek` registry entry already sets `preserveResponsesReasoningContent: true` and the router is expected to merge that registry capability into the effective provider config.

This appears to be a regression or an uncovered reasoning-replay path related to [#875](<https://github.com/lidge-jun/opencodex/issues/875>) and lidge-jun/opencodex#1193, rather than a missing user configuration flag.

## Reproduction

1. Start a Codex Desktop turn using `deepseek/deepseek-v4-flash` with High reasoning.
2. Give it a coding task that requires several shell/tool calls.
3. Let the model issue tools and return their results.
4. Observe the continuation request after the tool results.

The failure occurred twice in the same thread. In both attempts, several shell reads completed successfully and the next model continuation failed with the same upstream error.

### Actual result

The turn ends in `systemError` with an upstream 400 payload:

```json
{
  "error": {
    "message": "The `reasoning_text` in the thinking mode must be passed back to the API.",
    "type": "invalid_request_error",
    "param": null,
    "code": "invalid_request_error"
  }
}
```

Plain turns and earlier requests can succeed; the failure appears specifically on a tool-call continuation that needs prior reasoning replay.

### Expected result

For DeepSeek thinking-mode continuations, OpenCodex should pass the prior plaintext reasoning item back to the upstream Responses API so the tool loop can continue.

## Version

OpenCodex 2.12.0, installed through npm.

## Operating system

Windows 11.

## Provider and route details

* Provider: official DeepSeek API (`https://api.deepseek.com`)
* Provider ID: `deepseek`
* Model: `deepseek/deepseek-v4-flash`
* Reasoning effort: High
* Inbound API: Codex Responses API
* Upstream route: DeepSeek `POST /responses`

## Why this does not look like a local config omission

The 2.12.0 source already contains all of the following:

* the built-in DeepSeek registry entry sets `preserveResponsesReasoningContent: true`;
* `router.ts` merges that registry value when the saved provider does not explicitly override it;
* `openai-responses.ts` preserves raw reasoning content when the effective provider flag is true.

The saved provider does not explicitly set the flag to `false`. Therefore adding the same value manually would only duplicate the built-in default and should not be required.

## Related issues

* [#875](<https://github.com/lidge-jun/opencodex/issues/875>) fixed reasoning sanitization and DeepSeek Responses tool-loop handling.
* lidge-jun/opencodex#1193 fixed placeholder handling for replay-cache misses on OpenAI Chat-compatible routes.

This reproduction is on the current 2.12.0 release and uses the native DeepSeek Responses route, so it may be a new replay/state edge case or a regression after those fixes.

I can provide sanitized request/debug rows if a narrower capture is needed.

Ingwannu:
## Summary

- preserve one reasoning-bearing assistant turn when DeepSeek emits multiple tool calls in the same Responses turn
- normalize strict-provider history as a complete call batch followed by its matching result batch
- retain the existing #1292 single-call hook-context repair, keep tolerant providers byte/order equivalent, and leave ambiguous histories untouched
- document the DeepSeek wire decision and its fail-closed boundary

## Root cause

The provider-scoped #1292 normalizer repaired every call/result pair independently. A valid parallel history such as `reasoning, call A, call B, output A, output B` therefore became `reasoning, call A, output A, call B, output B`. DeepSeek merges adjacent reasoning and function calls into one assistant message and always enables parallel tool calling, so the second call lost the reasoning block that belonged to its original turn and the continuation failed with `reasoning_text` missing.

The fix groups calls that occur before the first matched output, emits all calls followed by their outputs in call order, and moves intervening non-tool context after the complete batch. Duplicate, backwards, or otherwise ambiguous call/result histories return unchanged rather than being guessed.

Fixes #1477.

## Verification

- `bun test tests/deepseek-inbound-wire.test.ts tests/openai-responses-passthrough.test.ts tests/deepseek-reasoning-replay.test.ts tests/deepseek-reasoning-replay-gaps.test.ts tests/config.test.ts tests/provider-registry-parity.test.ts tests/config-save-boundary.test.ts` — 296 passed, 0 failed
- `cd docs-site && bun install --frozen-lockfile && bun run build` — 265 pages built
- `git diff --check` — passed
- full suite — 11,049 passed, 11 skipped, 2 failures outside this diff:
  - the bridge stall timing test passed when rerun in isolation
  - the Unix shim token test fails identically on clean `origin/dev@849ab5e35` because this host has a live service-token environment; no shim files are changed here

All local builds and tests ran with `taskset -c 0-1 nice -n 10`.

## Decision Log

- 목적과 의도: preserve DeepSeek plaintext reasoning replay across same-turn parallel tool continuations without regressing the confirmed single-call hook-context repair
- 기존 구현 및 제약 조건: DeepSeek's Responses route is stateless, always enables parallel tool calling, and merges adjacent reasoning/function calls into one assistant message
- 검토한 주요 대안: disable parallel calls, duplicate reasoning per call, remove the #1292 normalization, or normalize one unambiguous call/output batch
- 선택한 방식: keep calls before the first output in one batch, follow them with matching outputs in call order, and retain non-tool context after the batch
- 다른 대안 대신 이 방식을 선택한 이유: it preserves the real assistant-turn boundary without fabricating reasoning or forwarding the known hook-interleaving failure
- 장점, 단점 및 영향: sequential and parallel DeepSeek continuations keep their reasoning contract; tolerant providers are unchanged and ambiguous histories remain fail-closed upstream

## Checklist

- [x] Docs and the applicable structure decision log are updated.
- [x] No credential, destination, workflow, dependency, GUI, or release boundary changes.

## Summary by CodeRabbit

- **New Features**
  - Improved DeepSeek Responses compatibility for parallel tool-call and result batches.
  - Preserved reasoning context and injected messages in the correct order during tool interactions.
  - Maintained separate sequential reasoning rounds.

- **Bug Fixes**
  - Avoided reordering ambiguous, duplicate, missing, or out-of-order tool results.

- **Documentation**
  - Updated provider and adapter guidance for tool-call batching and ordering behavior.

### DIFF
--- a/docs-site/src/content/docs/reference/adapters.md
+++ b/docs-site/src/content/docs/reference/adapters.md
@@ -58,6 +58,11 @@ waits and replays the identical request on the same key before any other handlin
 the translated `openai-chat` / Anthropic request path. Custom `runTurn` transports are not part
 of the HTTP retry loop.
 
+- DeepSeek's stateless Responses parser receives provider-scoped history normalization: hook-injected
+  context moves after an unambiguous tool-call/result batch. Parallel calls remain grouped before
+  their matching outputs so every call stays in the reasoning-bearing assistant turn. Tolerant
+  providers and ambiguous duplicate call IDs keep their original input order.
+
 - `forward` URL → `{baseUrl}/responses`. A `key` provider defaults to the legacy `{baseUrl}/v1/responses` construction.
 - A `key` provider may set a validated relative `responsesPath`; the adapter removes one trailing slash from `baseUrl` and sends `{trimmedBaseUrl}{responsesPath}`. For Ark Agent Plan, use `baseUrl: "https://ark.cn-beijing.volces.com/api/plan/v3"` with `responsesPath: "/responses"`.
 - In `forward` mode only a safe header allowlist is relayed (`FORWARD_HEADERS`): authorization,
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -540,15 +540,15 @@ function repairOrphanedInputItems(body: unknown, dropReasoning: boolean): unknow
 }
 
 /**
- * Make unambiguous Responses tool pairs adjacent for upstream parsers that require it.
+ * Make unambiguous Responses tool batches contiguous for upstream parsers that require it.
  *
  * [Decision Log]
- * - 목적과 의도: Keep Codex hook-injected developer context without letting it make a strict upstream reject the matching tool result.
- * - 기존 구현 및 제약 조건: The orphan repair verifies only pair presence; globally reordering valid history would change tolerant providers unnecessarily.
- * - 검토한 주요 대안: Reorder every Responses request, drop the intervening message, or gate a lossless reorder behind provider capability metadata.
- * - 선택한 방식: Reorder only unique call/result pairs for providers that explicitly require adjacency, preserving every intervening item immediately after the result.
- * - 다른 대안 대신 이 방식을 선택한 이유: The provider gate limits semantic blast radius, while refusing ambiguous duplicate ids avoids guessing which result belongs to which call.
- * - 장점, 단점 및 영향: DeepSeek receives the adjacency its parser requires; tolerant providers stay byte/order equivalent. Ambiguous duplicate ids still fail upstream rather than being silently rewritten.
+ * - 목적과 의도: Keep Codex hook-injected developer context without splitting a parallel tool-call turn away from its reasoning or making a strict upstream reject matching results.
+ * - 기존 구현 및 제약 조건: The orphan repair verifies only pair presence, while the original pair-by-pair reorder turned `reasoning, call A, call B, output A, output B` into two assistant turns and made DeepSeek reject call B for missing reasoning (#1477).
+ * - 검토한 주요 대안: Disable parallel calls (DeepSeek always enables them); duplicate reasoning per call; reorder each pair; or normalize the complete unambiguous call batch.
+ * - 선택한 방식: Treat calls emitted before the first matched result as one batch, emit all calls followed by their matched outputs, and preserve intervening non-tool items immediately after the batch.
+ * - 다른 대안 대신 이 방식을 선택한 이유: Batch normalization matches the Responses parallel-call shape without fabricating reasoning, while the provider gate and unique-pair requirement keep the blast radius narrow.
+ * - 장점, 단점 및 영향: DeepSeek keeps one reasoning-bearing assistant turn for parallel calls and still accepts hook-interleaved single calls; tolerant providers stay byte/order equivalent, and ambiguous duplicate ids are not guessed.
  */
 function normalizeResponsesToolResultAdjacency(body: unknown): unknown {
   if (!isPlainObject(body) || !Array.isArray(body.input)) return body;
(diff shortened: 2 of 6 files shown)


Ingwannu:
Exact-head CI note for `4602b1df5`:

- all Linux shards, gates, API usage, storage policy, keyring jobs, npm-global jobs, React Doctor, hygiene, and target enforcement passed;
- the macOS full-suite job ended with Bun 1.3.14 SIGTRAP/segmentation fault after its preceding tests passed; there is no failing product assertion in that job log;
- GitHub currently reports every job/check-run as completed but leaves workflow run `31503526156` in `in_progress`, so both the failed-run and single-job rerun endpoints return `403 workflow already running`.

No code or workflow change is justified by this runtime crash. I will keep the PR draft and rerun only the failed macOS job once GitHub finalizes the parent run.

Ingwannu:
Exact-head CI update for `ec5b995bec83c3e1ca23cdafcd739554eafde2e5` (the second commit is empty and exists only to replace GitHub run `31503526156`, whose run object was stuck `in_progress` after all jobs had completed and returned 500 to both normal and force-cancel APIs).

Run `31506198558` produced consistent product evidence:

- all four Linux test shards passed;
- gates passed, including typecheck, GUI tests, privacy scan, release-helper syntax, and CLI smoke;
- API usage and storage-policy jobs passed;
- keyring jobs passed on Ubuntu, Windows, and macOS;
- npm-global jobs passed on Ubuntu, Windows, and macOS.

The only red job is the macOS full-suite job, across three infrastructure/flaky outcomes on the same code:

1. attempt 1 completed 11,053 tests and failed only the unrelated existing `native main profile transactions > automatic recovery reports an externally refreshed target` timing test (5s limit, 8.16s runtime);
2. attempt 2 ended in Bun 1.3.14 macOS Silicon SIGSEGV / exit 133 after 402s, with Bun reporting `This indicates a bug in Bun, not your code`;
3. attempt 3 reproduced the same Bun 1.3.14 crash signature and exit 133 after 283s.

Neither failing path is in this PR diff, and the DeepSeek focused coverage plus every Linux shard passed. I am stopping repeated macOS reruns after the same runtime crash reproduced twice. The PR remains draft and unmerged pending independent human review; this comment records CI evidence rather than treating infrastructure failure as approval.

Wibias (commented):
Reviewed the DeepSeek batch normalizer. I did not find an additional blocking defect in this head. The change preserves same-turn parallel calls as one call batch followed by results, retains the existing single-call repair, and returns ambiguous duplicate/backwards histories unchanged rather than guessing.

I am not approving this head because it is currently not mergeable against the latest `dev`. Please rebase/resolve conflicts and request a short re-review of the final normalizer diff.

Wibias (changes requested):
Re-review against current dev@e8db4e03: one current blocking correctness issue remains. `normalizeResponsesToolResultAdjacency()` skips calls with no matching output, so a partially matched history can still be reordered. For example `[callA, callB, injected, outputB]` can move `outputB` ahead of `injected` even though `callA` is unresolved. This violates the documented fail-closed boundary for missing/duplicate/out-of-order call-result histories. Please return the original body when any collected call lacks exactly one later matching output, add missing/backward regression coverage, and update the public/design docs to name missing histories explicitly. The branch is also 9 commits behind current dev and currently not mergeable, so please rebase after the fix.

Wibias:
Thank you for this PR. The parallel-call reasoning fix is genuinely useful: DeepSeek rejects a valid Codex continuation when hook-injected developer context splits a call from its result, and splitting parallel tool calls into separate assistant turns makes the second call lose its reasoning_text. Restoring the same-turn batch keeps every call in its reasoning-bearing assistant turn.

I have superseded this in #1501, rebased onto the latest dev, and fixed the blocking review point: the normalizer now fails closed (returns the original body unchanged) whenever any collected call lacks exactly one later matching result, so partially matched, duplicate, missing, or backwards histories are never reordered. Closing this one in favor of #1501.

Wibias:
Superseded by #1501

Issue #1482, opened 2026-08-11
[Provider] Route OpenCode Go GPT-5.6 Luna through Responses with bounded terminal delivery


novelKR:
### Client or integration

Codex CLI

### Provider or upstream service

OpenCode Go

### OpenCodex version

dev@849ab5e35fdacddb4d47416c00a315111c78f97e (inspected on 2026-08-12)

### Endpoint or capability

POST /v1/responses, model-specific wire selection, streaming terminal delivery, function-tool calls, and tool-result continuations for opencode-go/gpt-5.6-luna.

### Current behaviour

OpenCode Go documents a model-specific protocol matrix. GPT 5.6 Luna is assigned to https://opencode.ai/zen/go/v1/responses with @ai-sdk/openai, while the same provider exposes other models through Chat Completions or Anthropic Messages.

At the inspected dev commit, however, OpenCodex's built-in opencode-go registry entry is provider-wide adapter: "openai-chat" and has no modelWireDefaults entry for gpt-5.6-luna. Unless a user adds an explicit modelAdapters override, the built-in route therefore selects the Chat Completions adapter and posts to /chat/completions instead of Luna's documented OpenCode Go Responses endpoint.

There is a second, distinct compatibility gap after correcting that mapping. The motivating reproduction registered the same OpenCode Go origin as a custom openai-responses provider and selected gpt-5.6-luna. In affected rich-history/tool turns:

- the upstream request returned HTTP 200;
- OpenCodex observed response.completed and reported usage within a few seconds;
- Native Codex did not commit a usable terminal event;
- the same logical turn was sampled again near the 300-second stream-idle boundary or was manually interrupted.

The failure was conditional: fresh minimal turns and some function-call continuations could complete. A control path in which Native Codex connected directly to the same OpenCode Go Responses upstream, without OpenCodex or the deployment relay, completed consecutive turns and tool round-trips.

The two gaps should therefore remain separate in diagnosis:

1. the built-in opencode-go preset selects the wrong wire for Luna;
2. once Luna is correctly routed through Responses, OpenCodex's streaming relay/client-delivery path can still fail to deliver a terminal that Native Codex commits.

[...]

PR #1484, opened 2026-08-11
fix(provider): route OpenCode Go Luna through Responses

Ingwannu:
## Summary

- route the built-in `opencode-go/gpt-5.6-luna` model through the OpenAI Responses adapter
- preserve explicit `modelAdapters` precedence and keep sibling OpenCode Go models on their existing wires
- add resolver and end-to-end upstream-path regressions
- document the mixed-wire decision and the separate streaming-policy boundary

## Root cause

OpenCode Go documents GPT 5.6 Luna at `/zen/go/v1/responses`, but the built-in provider had only the provider-wide `openai-chat` adapter. Without an explicit user override, Luna therefore reached `/chat/completions`.

The registry now supplies one exact-model default:

```ts
modelWireDefaults: {
  "gpt-5.6-luna": "openai-responses",
}
```

An explicit allowed `modelAdapters` entry still wins. No provider-wide adapter or sibling model behavior changes.

## Streaming scope

This PR deliberately does **not** set `modelResponsesUpstreamStreaming: false`. The endpoint mismatch is proven on current `dev` and in OpenCode Go's published endpoint matrix. The conditional terminal-delivery report in #1482 does not include a current-`dev` live canary, so a built-in bounded-JSON policy should remain a separate evidence-backed change.

## Validation

- OpenCode Go endpoint matrix rechecked: Luna uses `/zen/go/v1/responses`
- focused routing/registry/E2E tests: 51 passed
- `bun run typecheck`: passed
- documentation build: 265 pages built
- full suite: 10,961 passed, 11 skipped; the initial clean-worktree run had 6 GUI import errors before the suite's own GUI dependency install, one timing-sensitive bridge failure under a 3-core CPU cap, and one existing host-specific `codex-shim` token-path failure
  - all GUI-dependent files passed after the suite installed GUI dependencies
  - the bridge lifecycle file passed on rerun
  - the unchanged `codex-shim` test remains host-specific: it reads this host's persisted service-token path instead of the test's temporary `OPENCODEX_HOME`; none of this PR's files overlap that subsystem

## Documentation

The transport SOT records why the exact-model registry default was chosen over a provider-wide switch or an unproven streaming workaround. The public provider configuration reference now names the OpenCode Go Luna default and the explicit override boundary.

Addresses #1482.

## Summary by CodeRabbit

* **New Features**
  * OpenCode Go’s GPT 5.6 Luna model now automatically uses the Responses interface.
  * Other OpenCode Go models continue using the existing Chat interface.
  * Explicit model configuration overrides remain supported.
  * Streaming behavior is unchanged.

* **Documentation**
  * Added guidance on model-specific interface defaults and override behavior.

* **Tests**
  * Added coverage for Luna’s default interface, overrides, other models, and request routing.


Wibias (commented):
Reviewed the scoped OpenCode Go Luna wire correction. I did not find an additional blocker in this head: the exact-model registry default uses the existing mixed-wire precedence, explicit `modelAdapters` still wins, and sibling models remain on their existing wire.

The separate terminal-delivery problem described in #1482 is intentionally not solved here, so #1482 should remain open for that work. I am not approving this head because it is currently not mergeable against the latest `dev`; rebase/resolve conflicts and request re-review.

Wibias (approved):

Wibias:
Thanks @Ingwannu for the focused fix. This is useful because GPT-5.6 Luna on OpenCode Go now reaches its correct Responses endpoint without changing the wire for sibling models or overriding explicit user configuration. Keeping the endpoint correction separate from the still-unproven streaming workaround also avoids baking in unnecessary compatibility behavior. Merged, thank you!

Issue #1483, opened 2026-08-11
[Compatibility] Xiaomi MiMo v2.5 fails on 2.12.0 with invalid tool calls and an incorrect reasoning ladder


MAOKO1993:
### Client or integration

Codex App

### Provider or upstream service

Xiaomi MiMo

### OpenCodex version

2.12.0

### Endpoint or capability

/v1/chat/completions · streaming tool calls · reasoning_effort

### Current behaviour

On OpenCodex 2.12.0, Codex App cannot complete even a plain-text turn through a custom Xiaomi MiMo provider using the official OpenAI-compatible Chat endpoint.

Two failures are reproducible:

1. The generated Codex catalog advertises `low`, `medium`, `high`, `xhigh`, `max`, and `ultra` for `xiaomi-mimo/mimo-v2.5`. Selecting `max` sends that value upstream, but MiMo rejects it because the endpoint only accepts `none`, `low`, `medium`, or `high`.
2. After switching the same thread to `medium`, the request streams for several seconds and then OpenCodex terminates it with `upstream response contained invalid tool calls`.

No assistant message is delivered to Codex App. This reproduces with the plain prompt: `晚上好，你用的什么模型` ("Good evening, which model are you?").

### Expected behaviour

A plain-text Codex App turn through `xiaomi-mimo/mimo-v2.5` should complete with a visible assistant response.

The catalog must not advertise reasoning efforts that the selected Xiaomi endpoint rejects. If the endpoint supports only up to `high`, higher Codex effort labels should either be omitted or safely clamped according to an explicit provider contract.

If MiMo emits a non-conforming `tool_calls` streaming frame, OpenCodex should preserve the fail-closed behavior but provide enough redacted diagnostics to identify the incompatible nested shape, and apply a provider-specific normalization only when it is semantically safe.

### Minimal redacted request or reproduction

```shell
Redacted provider configuration:


{
  "providers": {
    "xiaomi-mimo": {
      "adapter": "openai-chat",
      "baseUrl": "https://api.xiaomimimo.com/v1",
      "authMode": "key",
      "defaultModel": "mimo-v2.5",
      "liveModels": true,
      "disabled": false
    }
  }
}


Steps:

1. Start OpenCodex 2.12.0 on macOS and sync the catalog into Codex App.
2. Select `xiaomi-mimo/mimo-v2.5`.
3. Set reasoning effort to `max` and send: `晚上好，你用的什么模型`.
4. Observe the upstream 400 validation error.
[...]

PR #1485, opened 2026-08-11
fix(provider): clamp Xiaomi MiMo public reasoning tiers

Ingwannu:
## Summary

- add a fixed-destination `xiaomi-mimo` preset for Xiaomi's official OpenAI-compatible Chat endpoint
- expose only the upstream-validated `low`, `medium`, and `high` reasoning tiers
- clamp direct `xhigh`, `max`, and `ultra` requests to `high`
- preserve same-named custom providers that point at another destination
- document the separate public Chat transport and its registry ownership decision

## Root cause and impact

`https://api.xiaomimimo.com/v1` had no registry-owned transport contract. A custom provider named `xiaomi-mimo` therefore kept the generic routed-model ladder and could advertise/send `max` even though the MiMo endpoint rejects every reasoning value above `high`.

The new preset is intentionally separate from the existing Xiaomi Anthropic and MiMo token-plan entries because those routes use different hosts and transport contracts. `preserveCustomDestination` prevents the preset from retargeting an existing custom key when a same-named row points somewhere else.

This PR does not normalize the separately reported malformed `delta.tool_calls` failure. The shared parser remains fail-closed until a redacted failing frame identifies the actual nested shape; guessing there could silently drop or mis-bind a tool call.

Refs #1483.

## Provider evidence

Verified from Xiaomi's primary sources on 2026-08-11:

- Official OpenAI-compatible Chat reference: https://platform.xiaomimimo.com/docs/en-US/api/chat/openai-api documents `POST https://api.xiaomimimo.com/v1/chat/completions` and API-key/Bearer authentication.
- Service terms: https://mimo.mi.com/docs/terms/user-agreement identify `Xiaomi Technologies Singapore Pte. Ltd.` as the API service operator outside mainland China; the agreement was updated 2026-07-07.
- Privacy terms: https://mimo.mi.com/docs/terms/privacy-policy identify Xiaomi Technology Netherlands B.V. and Xiaomi Technologies Singapore Pte. Ltd.; the policy was updated 2026-03-17.
- Aggregator authorization: not applicable. This preset sends a user key to Xiaomi's first-party endpoint for Xiaomi's own MiMo model; it does not resell or route third-party models.
- OpenCodex maintenance owner: `@Ingwannu`. Endpoint/auth/catalog breakage should be reported in the OpenCodex issue tracker and will be rechecked against the primary sources above.
- The preset does not declare live model discovery, so no authenticated `GET /v1/models` contract is claimed.

## Verification

- `bun test tests/mimo-token-plan-provider.test.ts tests/provider-registry-parity.test.ts` — 41 passed
- `bun test tests/provider-workspace-rail.test.ts` — 6 passed after installing the GUI workspace dependencies
- `cd docs-site && bun install --frozen-lockfile && bun run build` — 265 pages built
- `bun run test` — 11,047 passed, 11 skipped, 2 failed across 681 files; the failures were the live service-token environment collision in `codex-shim.test.ts` and a loaded-runner stall deadline in `bridge-lifecycle.test.ts`
- `env -u OPENCODEX_API_AUTH_TOKEN bun test tests/codex-shim.test.ts tests/bridge-lifecycle.test.ts` — both failed files passed in isolation, 82/82
- `git diff --check` — passed

## Checklist

## Scope notes

- Targets `dev` only.
- No GUI behavior or text-only product change.
- No authentication or credential flow change.
- No speculative malformed-tool-call repair.

## Summary by CodeRabbit

* **New Features**
  * Added Xiaomi MiMo as an OpenAI-compatible provider at `https://api.xiaomimimo.com/v1`.
  * Added support for the `mimo-v2.5` model with low, medium, and high reasoning levels.
  * Added handling for unsupported higher reasoning requests while preserving custom provider settings.

* **Documentation**
  * Updated provider catalogs in English, Japanese, Korean, Russian, Simplified Chinese, and Traditional Chinese.
  * Documented Xiaomi MiMo configuration and reasoning behavior.

### DIFF
--- a/docs-site/src/content/docs/guides/providers.md
+++ b/docs-site/src/content/docs/guides/providers.md
@@ -291,6 +291,7 @@ free-experimentation model.
 | SiliconFlow | `https://api.siliconflow.cn/v1` |
 | Volcengine Ark · Coding Plan · Agent Plan | `https://ark.cn-beijing.volces.com/api/v3` · `https://ark.cn-beijing.volces.com/api/coding/v3` · `https://ark.cn-beijing.volces.com/api/plan/v3` |
 | Xiaomi MiMo | `https://api.xiaomimimo.com/anthropic` |
+| Xiaomi MiMo (OpenAI Chat) | `https://api.xiaomimimo.com/v1` |
 | Kilo | `https://api.kilo.ai/api/gateway` |
 | GitLab Duo | `https://cloud.gitlab.com/ai/v1/proxy/openai/v1` |
 | Cloudflare AI Gateway | `https://gateway.ai.cloudflare.com/v1/{account-id}/{gateway}/anthropic` |
--- a/docs-site/src/content/docs/ja/guides/providers.md
+++ b/docs-site/src/content/docs/ja/guides/providers.md
@@ -218,6 +218,7 @@ Cline IDE/CLI のみで API からは使えません。`minimax/minimax-m2.5` �
 | SiliconFlow | `https://api.siliconflow.cn/v1` |
 | Volcengine Ark · Coding Plan · Agent Plan | `https://ark.cn-beijing.volces.com/api/v3` · `https://ark.cn-beijing.volces.com/api/coding/v3` · `https://ark.cn-beijing.volces.com/api/plan/v3` |
 | Xiaomi MiMo | `https://api.xiaomimimo.com/anthropic` |
+| Xiaomi MiMo (OpenAI Chat) | `https://api.xiaomimimo.com/v1` |
 | Kilo | `https://api.kilo.ai/api/gateway` |
 | GitLab Duo | `https://cloud.gitlab.com/ai/v1/proxy/openai/v1` |
 | Cloudflare AI Gateway | `https://gateway.ai.cloudflare.com/v1/{account-id}/{gateway}/anthropic` |
--- a/docs-site/src/content/docs/ko/guides/providers.md
+++ b/docs-site/src/content/docs/ko/guides/providers.md
@@ -218,6 +218,7 @@ Cline IDE/CLI에서만 제공되며 API로는 사용할 수 없습니다. `minim
 | SiliconFlow | `https://api.siliconflow.cn/v1` |
 | Volcengine Ark · Coding Plan · Agent Plan | `https://ark.cn-beijing.volces.com/api/v3` · `https://ark.cn-beijing.volces.com/api/coding/v3` · `https://ark.cn-beijing.volces.com/api/plan/v3` |
 | Xiaomi MiMo | `https://api.xiaomimimo.com/anthropic` |
+| Xiaomi MiMo (OpenAI Chat) | `https://api.xiaomimimo.com/v1` |
 | Kilo | `https://api.kilo.ai/api/gateway` |
 | GitLab Duo | `https://cloud.gitlab.com/ai/v1/proxy/openai/v1` |
 | Cloudflare AI Gateway | `https://gateway.ai.cloudflare.com/v1/{account-id}/{gateway}/anthropic` |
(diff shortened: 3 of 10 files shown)


Ingwannu:
Exact-head CI follow-up for `945d1718a4ad9ff9d3babe5a2523673a1bb1bf60`:

- the first `gates` attempt failed in the existing GUI aggregate-suite isolation pattern; the same 25 failures reproduced on clean `origin/dev`, while the four owning files passed 32/32 in isolation;
- the first macOS attempt completed 11,051 tests and timed out only in the unrelated existing `Cursor blob ID key channel bounds` test;
- I reran only those two failed jobs after the parent run finalized;
- rerun `gates` passed, rerun macOS passed, and workflow run `31515208300` is now completed with `success` at attempt 3.

All Linux shards, gates, macOS, API/storage, keyring, npm-global, React Doctor, hygiene, and target checks are therefore green on the exact PR head. No GUI or unrelated timeout change was added to this provider patch. The PR remains draft and unmerged pending independent maintainer review.

Wibias (commented):
Reviewed the Xiaomi MiMo reasoning-contract change. I did not find an additional blocker in this scoped head. The fixed-destination preset limits the advertised ladder to `low`/`medium`/`high`, clamps higher direct tiers to `high`, and `preserveCustomDestination` protects same-named custom providers on other origins.

The separately reported malformed `delta.tool_calls` failure in #1483 is intentionally not repaired here and should remain tracked independently. I am not approving this head because it is currently not mergeable against the latest `dev`; rebase/resolve conflicts and request re-review.

Wibias (changes requested):
Re-review against current dev@e8db4e03: the implementation correctly limits Xiaomi MiMo to low/medium/high and clamps xhigh/max/ultra to high, but the user-facing provider guides only add the endpoint row and omit that behavior. Since users can otherwise select a tier that OpenCodex silently lowers, please document the supported ladder and clamping behavior in the base guide and the five localized provider guides touched by this PR. The branch is also 9 commits behind current dev and currently not mergeable, so please rebase after the docs fix.

lidge-jun:
Reviewed against current `dev` (`a34d1a5a6`) and rebased onto it; your commit is preserved. I pushed one test-only commit on top.

**The registry preset lands as written.** The fixed destination, `preserveCustomDestination`, and the `low`/`medium`/`high` ladder with `xhigh`/`max`/`ultra` → `high` are all correct, and the tier values are evidence-backed rather than guessed — Xiaomi's docs list the accepted efforts and #1483 supplies the exact `max` rejection.

**One review finding, and it turned out to be half right.** Review flagged that a user `reasoningEffortMap` bypasses the clamp. I reproduced it directly:

```
{"efforts":["low","medium","high"],"map":{"xhigh":"high","max":"max","ultra":"ultra"},"max":"max","ultra":"max"}
```

So `max` does reach the wire as `max` when a user writes a conflicting map. But that is the **shipped contract**, not an oversight in this PR. `healMappedTiers` treats a wire map as authoritative evidence of the tiers an upstream can emit and merges its Codex values into the ladder at lookup time — that is what lets a stale persisted ladder recover a newly documented tier without rewriting user config, and `tests/reasoning-effort.test.ts` pins it under *"stale reasoning-ladder self-heal"*.

I did try making the clamp authoritative. It broke two existing tests, because a mapped value like Kimi's `none` is a legitimate non-Codex wire word rather than a tier comparable to a ladder. Narrowing the clamp to Codex tiers only still fought `healMappedTiers`, so I reverted it. The honest reading is that the registry clamp protects the **default** route, and a user who writes a conflicting map has deliberately opted out — which is consistent with how every other provider ladder behaves here.

The follow-up is therefore two tests recording that boundary: one proving the clamp holds on the default route (`max`/`ultra`/`xhigh` → `high`), one proving the overriding map wins. Neither changes behavior; they stop the next reader from treating the clamp as an enforcement the code does not implement.

Verification at `49503ea0e` on a Linux runner (Bun 1.3.14): `bun x tsc --noEmit` exit 0; 18-file reasoning/mimo/registry/router sweep 264 pass / 0 fail. Red-before for your tests: tests-only diff on `dev` = 2 failures.

PR #1505, opened 2026-08-12
fix(responses): bind reasoning replay to route identity

luvs01:
## Summary

- bind process-local raw-reasoning replay to the exact client thread, configured provider destination, wire adapter, final model, and physical credential identity
- share one mutable replay-scope holder across parsed-request copies and already-created bridges so account/key rotation changes the namespace before the next cache read or write
- fail closed for missing, keyless, or header-only credential identities; OAuth replay is bound to the existing credential slot and exact token generation
- represent destinations and credential-bearing material only with process-local HMAC identities; raw URLs, tokens, API keys, account-slot IDs, generations, and authorization headers are never stored in replay keys
- clear an already-bound holder when a later route cannot establish a complete identity, preventing stale-namespace writes

## Root cause and impact

#1474 removed the process-wide fallback and required an explicit conversation scope. That prevents unrelated unscoped requests from sharing replay entries, but a single conversation can still move between providers or physical credentials during combo, pool, OAuth, image, or web-search recovery. Provider-generated call IDs are not globally unique, so a later target could reuse the same thread and call ID and receive the earlier target's raw reasoning.

This follow-up keeps replay process-local and bounded, but keys each entry to the active route and physical credential identity. Same-generation, same-target continuation remains compatible. A provider, model, destination, account/key, or OAuth token-generation change receives the existing safe placeholder or cache-miss behavior instead of another target's reasoning.

The final patch uses the existing OAuth access snapshot (`accountId` plus credential `generation`) and does not change the OAuth authentication module's public snapshot surface.

Follow-up to #1474.

## Verification

- Bun 1.4.0-canary.1 after the original rebase: 99 focused tests passed, 0 failed, 292 assertions
- Bun 1.3.14 after the original rebase: 99 focused tests passed, 0 failed, 292 assertions
- OAuth scope reduction on Bun 1.4.0-canary.1: 15 focused tests passed, 0 failed, 78 assertions
- OAuth scope reduction on Bun 1.3.14: 14 focused tests passed; one existing concurrent xAI test exceeded the default 5 s watchdog at 5.017 s and passed 1/1 with a 30 s timeout (5.988 s)
- invalid-holder regression on the final head: 5/5 and 33 assertions passed on both Bun 1.4.0-canary.1 and Bun 1.3.14
- final latest-dev head: TypeScript 7.0.2 `tsc --noEmit` passed on both Bun versions
- focused physical key-rotation E2E passed on Bun 1.4 and Bun 1.3.14
- `git diff --check` passed
- latest-dev rebase preserved the exact validated patch-id while retaining the adjacent passive-production core changes
- Codex Security diff scan reviewed all changed production surfaces and found 0 reportable candidates
- independent read-only production, test, and OAuth-reduction reviews found no remaining P0-P3 findings
- exact-head Codex forward credential-boundary regressions on Bun 1.4 and Bun 1.3.14 — 7/7 passed with 60 assertions on each runtime
- exact-head TypeScript 7.0.2, privacy scan, and `git diff --check` — passed
- independent validation accepted the effective-bearer/pool-slot/generation HMAC binding with no remaining P0-P2 security gap

Focused affected-surface tests, dual-runtime typecheck, privacy, diff, and independent review satisfy the local CI gate. Unrelated environment/runtime timing is recorded separately and is not attributed to this change; the same GitHub CI workload is not duplicated locally.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reasoning replay isolation across providers, destinations, models, adapters, and credentials.
  * Prevented replay when required identity information is missing or invalid.
  * Ensured credential rotation, refresh, and failover cannot reuse reasoning from a previous credential.

* **Documentation**
  * Documented safeguards and conditions governing reasoning replay.

* **Tests**
  * Added coverage for identity isolation, credential changes, failover, invalid scopes, and streaming or batch responses.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -226,7 +226,7 @@ function toolResultImageChatParts(content: string | OcxContentPart[]): unknown[]
 function messagesToChatFormat(parsed: OcxParsedRequest, provider: OcxProviderConfig): unknown[] {
   const out: unknown[] = [];
   const { context, options } = parsed;
-  const replayCacheScope = parsed._clientThreadId;
+  const replayCacheScope = parsed._reasoningReplayScope;
 
   interface PendingToolCall { id: string; name: string }
   let pendingToolCalls: PendingToolCall[] = [];
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -1,4 +1,10 @@
-import type { AdapterEvent, OcxMessagePhase, OcxProviderContinuationState, OcxUsage } from "./types";
+import type {
+  AdapterEvent,
+  OcxMessagePhase,
+  OcxProviderContinuationState,
+  OcxReasoningReplayScopeRef,
+  OcxUsage,
+} from "./types";
 import { adapterFailureFromMessage, classifyError, CYBER_POLICY_ERROR_CODE, isCyberPolicyCode, type OcxErrorPayload } from "./lib/errors";
 import { encodeCompactionSummary } from "./responses/compaction";
 import { encodeReasoningEnvelope, type ReasoningEnvelope } from "./responses/reasoning-envelope";
@@ -186,7 +192,7 @@ export function bridgeToResponsesSSE(
      * Provider call ids are not globally unique; scoping by thread keeps one
      * conversation's reasoning out of another's continuations.
      */
-    replayCacheScope?: string;
+    replayCacheScope?: OcxReasoningReplayScopeRef;
     /**
      * Test seam for the wire/stall beat loop. Production omits this and uses the
      * global timers; injecting here must not change scheduling semantics.
@@ -1362,7 +1368,7 @@ function buildResponseJSONWithBudget(
     onUsage?: (usage: OcxUsage | undefined) => void;
     translatorBudget?: TranslatorBudget;
     /** Conversation identity for the reasoning replay cache (issue #950). */
-    replayCacheScope?: string;
+    replayCacheScope?: OcxReasoningReplayScopeRef;
   },
 ): Record<string, unknown> {
   const responseId = `resp_${uuid()}`;
--- a/src/images/loop.ts
+++ b/src/images/loop.ts
@@ -907,7 +907,7 @@ export async function runWithImageBridge(deps: ImageBridgeDeps): Promise<Respons
     }, 2_000,
     {
       translatorBudget,
-      replayCacheScope: parsed._clientThreadId,
+      replayCacheScope: parsed._reasoningReplayScope,
       ...(deps.forceEmptyResponseId ? { responseId: "" } : {}),
       hideThinkingSummary: parsed.options.hideThinkingSummary,
       stallTimeoutSec: deps.stallTimeoutSec,
--- a/src/responses/reasoning-replay-cache.ts
+++ b/src/responses/reasoning-replay-cache.ts
@@ -19,9 +19,32 @@
  * long-lived proxy cannot grow without limit.
  */
 
+import { createHmac, randomBytes } from "node:crypto";
+import type {
+  OcxProviderConfig,
+  OcxReasoningReplayIdentity,
+  OcxReasoningReplayScopeRef,
+} from "../types";
+
 const MAX_ENTRIES = 64;
 const MAX_TOTAL_BYTES = 256 * 1024;
 const TTL_MS = 60 * 60 * 1000;
+const replayIdentityKey = randomBytes(32);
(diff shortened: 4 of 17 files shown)


Ingwannu:
Maintainer review on exact head `199b7d6241804795ca6924add5988f2dd6c0d6f8` against current `dev@37139818d2d3fe7b033e958ccf87080e4be2b4b7`.

The underlying cross-route replay problem is real, and this head applies the retention boundary at the right identity level: client thread, provider name, opaque destination identity, wire adapter, final model, and physical credential identity. API keys, OAuth slot/generation material, account ids, authorization headers, and configured URL paths are represented only by process-local HMACs and are neither logged nor persisted. Incomplete identities clear the mutable holder and fail closed, while key rotation, OAuth refresh, and Codex alternate-account selection rebind before the next replay read/write.

I also checked the recovery branches. API-key 429 failover and Codex alternate-account retry rebind to the replacement credential. OAuth refresh changes the generation-bound namespace. Anthropic's separate multi-account pool currently cannot provide the same stable credential-generation tuple, so it remains deliberately unscoped/fail-closed rather than reusing another account's reasoning, consistent with the documented limitation.

Exact-head validation under a 2-CPU/8-GB cap:

- replay identity, cache bounds, bridge propagation, DeepSeek continuation, image/web-search, and key-failover tests: **95 passed, 0 failed**;
- `bun run privacy:scan`: passed;
- `git diff --check`: passed.

I found no additional code blocker in the reviewed surface. I am leaving this as a draft and not approving or merging until the author completes the local-CI/ready checklist and exact-head required CI is available.

Wibias (changes requested):
Review on current head `4d725bd2f4b495a31abb46e195f4950c190038b2`.

There is one blocking credential-isolation bug in the new replay scope. In the forward-auth branch, pool/main-pool identity is derived from `chatgptAccountId` only, while direct-forward identity ignores the effective forwarded `Authorization` value and hashes `provider.headers` instead. That means a token generation change, or two physical pool credentials that share one ChatGPT account id, can keep the same replay namespace. Under a same-thread/call-id collision, raw reasoning from the previous physical credential can therefore still be reused.

Please bind pool/main-pool replay identity to the physical slot and generation, e.g. `accountId` plus `writerGeneration`, and bind direct-forward identity to the effective forwarded credential headers including Authorization and chatgpt-account-id. Keep the raw material inside the existing HMAC helper. Add regressions for generation rotation, distinct pool accounts sharing a ChatGPT account id, and distinct forwarded bearer tokens sharing an account id.

Separately, this head is now 10 commits behind current `dev@70d2e175` after #1464 merged, and the exact-head Cross-platform CI / React Doctor runs are `action_required` with zero jobs. Rebase onto current dev and obtain an executed exact-head CI signal after fixing the credential identity.

Wibias (changes requested):
Re-review on current head `93bf9a376eb607a5167f35c475f5a7e5bef2e037`.

The prior P1 credential-isolation blocker is fixed. Pool/main-pool replay identity is now bound to the effective bearer plus the physical `accountId`, credential generation, and `writerGeneration`; direct-forward identity uses the effective forwarded `Authorization` and `chatgpt-account-id`. The raw credential material stays inside the process-local HMAC identity. The focused regressions cover distinct forwarded bearers under one ChatGPT account, alternate pool slots sharing one ChatGPT account, and generation/writer-generation rotation. The server key-failover E2E also verifies that a 429 rotation cannot replay the previous physical key's raw reasoning.

I also checked the rebinding points: initial route binding uses the resolved effective forward headers/auth context, pool retry binds before rebuilding the retry request, and key/OAuth rotations replace or clear the shared mutable holder before subsequent replay reads/writes. I did not find a new code-level credential-boundary blocker.

The rebase gate is now satisfied: this head is one commit ahead of and zero commits behind current `dev@d667367`. The remaining blocker is exact-head CI: Cross-platform CI and React Doctor are both `action_required` with zero jobs executed. Please obtain an executed exact-head CI signal, then request re-review.

Wibias:
Thank you @luvs01 for pushing this through and for addressing the credential-isolation review feedback.

This is useful because raw reasoning replay is now scoped to the exact active route and physical credential identity instead of only the client thread. Provider/model/destination changes, key rotation, OAuth generation changes, and Codex pool failover now move to a different replay namespace, which prevents reasoning from one credential or upstream target from being reused by another. The process-local HMAC identities also keep raw tokens, API keys, account-slot IDs, generations, and upstream destinations out of replay keys while still preserving same-target continuation behavior.

PR #1510, opened 2026-08-12
feat(lab): CL-10 public evidence operator and community integration

Wibias:
## Summary

Implements CL-10.1 through CL-10.4 on the reviewed public-evidence contract: privacy-safe projection, canonical signed local exports, explicit local operator surfaces, and isolated non-authoritative community verification/revocation context.

CL-10.5 remote publishing is deliberately **not implemented**. It remains blocked until an exact service origin, transport, security, retention, and revocation contract is independently reviewed and accepted.

### CL-10.1 - public projection and trust boundary

- closed `PublicEvidenceRecordV1` / `PublicEvidenceBundleV1` schemas
- allowlist-only projection from canonical Lab evidence
- deterministic domain-separated public IDs
- exact reviewed scenario/assertion and public-route authority checks
- fail-closed handling for private route identity instead of broadening claims
- second-pass public privacy validation, including decoded text artifact bytes
- closed incident corpus and strict public-field validation

### CL-10.2 - canonical bundles, signatures, and local storage

- RFC 8785/JCS canonical bytes with strict UTF-8/duplicate-key/depth/width limits
- domain-separated SHA-256 bundle and revocation identities
- installation-local Ed25519 publisher identity with restrictive private-key handling
- exact bundle digest/signature verification and canonical record/artifact ordering
- bounded, descriptor-checked, crash-aware local persistence
- public artifacts remain fail-closed without reviewed `public_export` authority

### CL-10.3 - explicit local operator surfaces

- `ocx lab public preview`
- `ocx lab public export`
- `ocx lab public verify <file>` with nonzero exit on invalid evidence
- management API preview/export/verify routes
- no publisher private-key or local-path disclosure on public surfaces

### CL-10.4 - community verification and revocation

- bounded community bundle/revocation import
- strict schema, canonical-byte, digest, signature, publisher, authority, privacy, and artifact-policy validation before persistence
- exact same-publisher revocation bootstrap and deterministic replay/conflict handling
- record revocations propagate across bundles from the same verified publisher
- bounded local community cache with concurrency-safe quota handling
- local/community provenance and purge handling without deleting unrelated third-party evidence
- read-only Compatibility Matrix community context with explicit `community_untrusted_v1` / not-local-verdict semantics

## Trust isolation

Community evidence remains non-authoritative:

- no writes into canonical local compatibility evidence
- no effect on local verdicts, freshness, Routing Profiles, Router Intelligence, combined scores, or CL-08 scheduling
- signatures prove bundle integrity and publisher-key continuity only, not truth, correctness, ownership, or platform authority
- revoked or unverifiable evidence remains quarantined/non-authoritative

## Explicit non-scope

- no remote publishing transport
- no `/api/lab/public/publish` endpoint
- no `ocx lab public publish` command
- no background publishing or telemetry
- no remote service selection
- no public artifact export until a reviewed `public_export` authority exists

## Byte/signature contract

- canonical payload bytes are UTF-8 RFC 8785/JCS
- `domainHash(domain, payload) = SHA-256(UTF-8(domain) || 0x00 || payload)`
- `bundleId` binds the normalized public bundle content
- `bundleDigest` binds that content plus `bundleId`; `bundleDigest` and `signature` are excluded from their own digest preimage
- Ed25519 signs the raw 32-byte `bundleDigest`
- revocations bind their schema/version, issued UTC day, exact publisher identity, targets, and reason under the dedicated revocation domain; they apply only to evidence verified under the exact same publisher key

## Review status

This PR remains **draft**. The previously raised byte-contract, revocation, privacy, authority, parser, file-boundary, artifact-policy, cache, purge, and lifecycle findings have focused regression coverage. Final merge review should use the exact branch head and fully green exact-head CI.

Remote publishing remains a separate future gate and is not authorized by this PR.

## Summary by CodeRabbit

* **New Features**
  * Added privacy-preserving public evidence previews and signed local exports.
  * Added CLI and management API workflows for previewing, exporting, verifying, importing, and listing evidence.
  * Added isolated community evidence with revocation tracking and Compatibility Matrix summaries.
  * Added localized community-evidence labels across supported languages.
  * Added strict validation for privacy, file safety, signatures, canonical data, and storage limits.

* **Bug Fixes**
  * Improved sensitive-data purge handling and cleanup reliability.
  * Hardened malformed input, duplicate-key, unsafe-file, and invalid-signature handling.
  * Community evidence remains clearly non-authoritative and cannot change local results.


Ingwannu (commented):
Contract review on exact head `e07ea0949ad528c9b4de3d1c6469cf7852d0339d` against current `dev@cbbfdd877`.

The overall privacy/trust direction is conservative and valuable: public export is an allowlist-only projection, private route dimensions make evidence non-exportable instead of being silently generalized, local identifiers and arbitrary diagnostics are excluded, signatures prove integrity/continuity rather than truth, explicit consent is required, community evidence remains non-authoritative, and remote publishing is blocked on a later fixed-origin service contract. I found no runtime behavior change because this PR changes documentation only, and exact-head docs/hygiene CI is green.

I am not approving the contract yet because two normative boundaries need to be made exact before implementation starts:

1. **Canonicalization and signature domain are underspecified.** The contract names deterministic canonical bytes and Ed25519 signatures, but it does not freeze a concrete canonicalization algorithm/encoding or state whether `bundleDigest` and `signature` are excluded from the signed payload. It also does not freeze domain-separation bytes and verification order. Different serializers could produce incompatible or ambiguously self-referential bundles. Define one exact byte contract (for example, a named canonical JSON profile with UTF-8, integer/string rules, object-key ordering, and duplicate-key rejection), the exact preimage for `bundleDigest`, the exact signature preimage/domain tag, and the validation order.

2. **Revocation trust needs an anti-substitution rule.** “The same publisher key” is necessary but not sufficient if an importer accepts a valid signed revocation whose publisher identity or target IDs are not bound to the original imported bundle's verified signer and schema domain. Require the revocation preimage to include a dedicated domain tag, schema/version, publisher key identity, target bundle/record IDs, and issued day; importers must apply it only when the verified key exactly matches the target bundle's verified publisher key. Also define deterministic conflict/replay handling for repeated or contradictory revocations.

Please update both the normative CL-10 contract and the shorter design summary so they do not diverge. After those byte-level and revocation bindings are explicit, this remains a strong contract candidate. This comment does not authorize runtime export/publish/import implementation or remote service selection.

Ingwannu (changes requested):
The CL-10 direction is valuable, but the current head `2f3e1e80162f382c6b654edb5683b558f6947de8` is not mergeable.

This is now a 45-file, ~6.4k-line runtime/authenticity surface rather than the contract-only change I initially reviewed. The exact-head CI is red on all four Linux shards and macOS with deterministic product-test failures, including:

- authority validation rejecting the PR's own signed fixtures in `lab-public-evidence`, wire-contract, and deep-review tests;
- a deterministic day regression (`2026-08-12` becoming `1970-01-01`);
- community revocation/index counts disagreeing with expected active/revoked records;
- privacy/authority failures returning the wrong rejection boundary;
- community fixtures producing empty records or failing before import validation.

The latest review also still identifies unresolved trust-boundary work: atomic cache quota enforcement, crash-durable storage plus origin indexing, canonical record ordering, reviewed artifact authority on import, publisher creation only after full validation, record-scoped revocation propagation, bounded diagnostics, IPv6 privacy rejection, and durable POSIX/Windows publication semantics.

Please move this back to draft, rebase onto current `dev`, and split it before another merge review:

1. freeze the normative byte/signature/revocation/privacy contract and its tests;
2. land local projection/signing/storage with green cross-platform CI;
3. land community import/revocation/quota handling separately;
4. land the GUI read-only context separately with behavioral tests.

Do not rerun the current red head as a substitute for fixes. Request review again only after every exact-head shard is green and the correct CodeRabbit findings are resolved.

Ingwannu (changes requested):
Re-reviewed exact head `c95b2cdd31cc02a2c9a656ecb95b3e426f2195a2`. This update resolves most of the earlier deep-review findings: 9/10 focused trust regressions now pass locally, and the remaining failure in that file is a test-ordering bug rather than a revocation product failure. Typecheck, privacy scan, diff check, and the other focused lifecycle/wire tests are green locally. The PR is still not mergeable for the following exact-head blockers:

1. `src/lab/public/privacy.ts` never inspects decoded `artifact.contentBase64` bytes. The normative contract requires the second export-specific secret/PII scan to cover final exported artifact bytes; scanning only `artifactClass` and `mediaType` allows a reviewed artifact class to carry credentials, paths, account data, or repository text. Decode within the existing byte limits, reject non-canonical/invalid Base64, scan the final bytes with an explicit closed media-type policy, and apply the same check on local signing and community import.
2. `projectPublicEvidenceRecord` returns an exportable record after shape validation only. Its exported API can be called without `projectPublicEvidence`, bypassing `validatePublicEvidenceAuthorities` and `validatePublicEvidenceRecordPrivacy`. Either make the low-level projector private/clearly unsafe or apply the full authority/privacy gate before returning `exportable`; add a direct-call regression.
3. `parseStrictPublicJson` enforces depth and duplicate names but still does not enforce the V1 ceilings of 64 object keys and 512 array elements before object construction. Add counters inside the recursive parser and boundary/overflow tests. The contract also names a pre-parse string ceiling; enforce it here or explicitly route every string through an equally early bounded tokenizer.
4. `readBoundedPublicFile` relies on `O_NOFOLLOW ?? 0` and then checks `fstatSync(fd).isSymbolicLink()`, which cannot detect the path symlink after open. On platforms without `O_NOFOLLOW`, the protection disappears. Reject symlinks with a pre-open `lstat`, keep `O_NOFOLLOW` where available, and compare path/descriptor identity if the supported race model requires fail-closed behavior. Preserve the hardlink check and add platform-independent symlink coverage.
5. Exact-head CI is red because `tests/lab-public-evidence.test.ts` requires the privacy error text, while closed-schema validation correctly rejects the URL-valued surface first. Assert the invariant (rejected before key creation) or accept either correct fail-closed boundary; do not weaken validation just to satisfy the message regex. `tests/lab-public-deep-review-regressions.test.ts` has a second deterministic expectation bug: sorting `expect.objectContaining` matchers cannot sort by their hidden `bundleId`, while the production list is sorted. Compare keyed summaries or sort concrete expected objects.

This remains a valuable direction, but it is a 46-file, 6.7k-line signing/import/revocation/storage/GUI trust surface despite the contract-only title/body. Keep it draft and split it as previously requested; at minimum, do not request merge review until the exact-head suite is fully green and the artifact/privacy/parser/file-boundary blockers above have direct adversarial tests.

Ingwannu (changes requested):
Exact-head update for `a3a3eb524d1bd19b992a0f9304546bb7fc6939bf`: the new commit correctly fixes the fail-closed error-message expectation in `lab-public-evidence.test.ts`. It does not change the remaining product blockers from my prior review: decoded public artifact bytes are still outside the second-pass privacy scan; the exported single-record projector can still return before authority/privacy validation; strict JSON still lacks the 64-key/512-element pre-construction ceilings; and the file reader still has no effective symlink rejection fallback when `O_NOFOLLOW` is unavailable. The separate matcher-order bug in `lab-public-deep-review-regressions.test.ts` also remains. Keeping changes requested on this exact head; please address those boundaries and obtain fully green exact-head CI before requesting review again.

Ingwannu (commented):
Exact-head re-review at `e9bfd1855937089a9448cef371dd9c03fa8298a7`: the concrete blockers from my prior reviews are now addressed. The final artifact bytes are decoded/canonicalized and privacy-scanned; the direct projector applies authority and privacy before returning an exportable record; strict JSON enforces depth, object-key, array-element, and string bounds before semantic construction; the file reader uses pre-open `lstat` plus descriptor identity checks; and both incorrect test expectations are fixed. Local focused validation is green: 33/33 trust/wire/lifecycle/artifact tests, typecheck, privacy scan, and diff check.

I found no remaining blocker in those reviewed paths. I am not approving or recommending merge yet because this remains a draft 46-file, ~6.7k-line signing/import/revocation/storage/GUI trust surface under a contract-only title, and the exact-head full CI is still running. The prior split request remains the conservative review position: keep the contract/byte rules, local signing/storage, community import/revocation, and GUI context independently reviewable, or obtain explicit owner agreement to accept the combined scope. Request final review only after exact-head CI is fully green and the scope/title accurately describe what will land.

Wibias:
will be split once finished

Wibias:
@coderabbitai review

Wibias:
@coderabbitai review

lidge-jun:
Closing this as an **owner decision**, not a quality judgment on the work.

I reviewed the Compatibility Lab integration on `dev` today and found that it has crossed a boundary I am not willing to ship past. The problem is not the Lab code itself — the isolation effort in it is visible and careful. The problem is the dependency direction into the proxy core:

- `src/server/responses/core.ts:36` statically imports `resolveProductionRouteSubject`, and the call at ~line 1999 runs on **every request attempt**, for every user, including users who have never configured a routing profile. There is no config flag to turn it off.
- `src/router.ts:34` statically imports `assemblePolicyCandidateEvidence` unconditionally, even though it is only reachable on `policy/` routes.
- `src/server/index.ts:52-53` and `src/server/lifecycle.ts:10` load Lab automation modules unconditionally at startup and shutdown.

opencodex is a provider proxy. A user who points at one model and goes should not be executing evidence-collection code on their request path, and should not have to.

**What happens next:** feature work on the CL line is frozen until a decoupling patch lands that enforces the boundary — Lab and routing-compatibility code must execute only when routing profiles are actually configured. I am taking that patch on directly. Once the boundary is in place and enforced by a regression test, the public-evidence work in this PR can be reopened or resubmitted on top of it.

Nothing here is lost: the branch and its history stay intact, and CL-01 through CL-09 remain on `dev`. This is a sequencing decision — the boundary comes first.

Apologies for the late intervention. This should have been caught before CL-09 merged; that gap is mine, not yours.

Issue #1504, opened 2026-08-12
[Bug][High impact] Routed Fast mode is hidden and openai-chat silently drops service_tier


str0203:
### Client or integration

Codex App

### Provider or upstream service

Custom OpenAI-compatible relay providers that expose GPT-5.6 models and support OpenAI Fast mode / priority processing.

### OpenCodex version

2.11.1 locally; current repository default branch also inspected on 2026-08-12.

Codex App runtime observed: `0.147.0-alpha.6.5`.

### Endpoint or capability

- Codex routed model catalog speed/service-tier metadata
- `/v1/responses`
- `/v1/chat/completions`
- OpenAI Fast mode via `service_tier: "fast"` or `service_tier: "priority"`

### Current behaviour

**Severity: high user impact / silent capability loss.**

This is not only a missing cosmetic control. OpenCodex currently makes a paid latency feature unavailable for routed OpenAI-compatible providers, even when the upstream explicitly supports Fast mode. The failure is silent: the user sees no selector, receives no warning, and the upstream request may omit `service_tier` entirely.

A native `gpt-5.6-sol` catalog entry exposes the Codex App **Standard / Fast** speed selector. The same model routed through a custom OpenAI-compatible provider does not.

There are two independent OpenCodex behaviours causing this:

1. `normalizeRoutedCatalogEntry()` unconditionally removes all speed/service-tier metadata from every routed catalog row:

   - `additional_speed_tiers`
   - `service_tier`
   - `service_tiers`
   - `default_service_tier`

2. The Responses request parser captures `service_tier` into `parsed.options.serviceTier`, but the `openai-chat` adapter does not serialize that value into the upstream `/v1/chat/completions` body.

The generated catalog and the request pipeline therefore suppress the same capability at two different boundaries:

```text
Native GPT-5.6 model
  -> catalog advertises Fast
  -> Codex App shows Standard / Fast
  -> service_tier can reach upstream

Routed GPT-5.6 model
  -> OpenCodex deletes Fast catalog metadata
  -> Codex App cannot offer the choice
  -> openai-chat does not serialize parsed serviceTier
  -> a Fast-capable relay cannot receive the user's choice
```

Typical relay onboarding provides only an API base URL and API key. Relay operators do not provide an OpenCodex-specific fragment such as:

```json
{ "supportsServiceTier": true }
```

[...]

Issue #1436, opened 2026-08-10
[Feature]: expose speed/service-tier capabilities for eligible custom providers and routed models


str0203:
## Client or integration

Codex Desktop using the OpenCodex-generated model catalog

## Area

Custom providers / routed model catalog / Responses API / service tier

## OpenCodex version

2.11.1

## Summary

Models routed through a custom OpenAI-compatible provider cannot expose the Codex Desktop `Standard / Fast` speed selector, even when the upstream provider may support a priority or fast service tier.

The native ChatGPT/Codex account-pool route for the same model family exposes both reasoning effort and speed, while the routed custom-provider model exposes reasoning effort only.

This appears to be a capability-publication and provider-adaptation gap rather than a Codex Desktop rendering bug: the client displays the speed selector when the generated catalog contains service-tier metadata.

## Current behaviour

For a native account-pool model, the generated catalog contains speed metadata similar to:

```json
{
  "slug": "gpt-5.6-sol",
  "service_tiers": [
    {
      "id": "priority",
      "name": "Fast",
      "description": "1.5x speed, increased usage"
    }
  ],
  "additional_speed_tiers": ["fast"]
}
```

Codex Desktop consequently offers:

```text
Reasoning effort: Low / Medium / High / ...
Speed: Standard / Fast
```

For the same model routed through a custom provider, the catalog entry is namespaced but does not contain speed metadata:

```json
{
  "slug": "custom-provider/gpt-5.6-sol",
  "service_tiers": null,
  "additional_speed_tiers": null
}
```

Codex Desktop then offers reasoning effort only.

The current routed-catalog normalization removes fields including:

```text
additional_speed_tiers
service_tier
service_tiers
default_service_tier
```

Request-time OpenAI `service_tier: "priority"` injection is also restricted to an `openai-responses` route with `supportsServiceTier: true`.

This is safe as a default, but there is no supported opt-in path for a custom provider/model whose upstream capability has been verified.

## Expected behaviour

Allow custom providers or individual routed models to explicitly declare a verified speed/service-tier capability.

When that capability is enabled, OpenCodex should:

1. preserve or generate the appropriate `service_tiers` catalog metadata;
2. expose the speed selector to compatible clients such as Codex Desktop;
3. map the selected tier to the verified upstream parameter;
4. keep the capability disabled for unknown or unsupported providers;
5. avoid inferring support from the model name alone.

## Suggested configuration contract

A provider-level capability could reuse the existing safety boundary:

```json
[...]

PR #1512, opened 2026-08-12
fix(openai-chat): forward caller-selected service tier

Yuxin-Qiao:
## Summary

- Preserve an explicit `parsed.options.serviceTier` value when serializing OpenAI-compatible Chat Completions requests.
- Add a regression test covering `service_tier: "priority"` on the outbound `/chat/completions` body.
- This fixes the request-path half of #1504; routed catalog capability publication remains tracked separately by #1436.
- Rebased onto current `dev@cbbfdd8773e68a5dc2391ddeb32f33a225373c1a`; current head: `ab72f5441e1b`.

Refs #1504
Refs #1436

## Verification

- `bun test --isolate --max-concurrency 1 tests/openai-chat-hardening.test.ts tests/service-tier-capability.test.ts` — 50 passed, 0 failed, 84 assertions.
- `bun run typecheck` — passed on the rebased head.
- A repository-wide `bun run test` result is not claimed; the suite contains current baseline/environment failures unrelated to this narrow serializer change.

## Checklist

- [x] Docs or release notes were updated when needed; this is an internal request serialization fix.
- [x] Local CI is green.
- [x] Branch is on the latest `dev` commit.
- [x] Correct Codex and CodeRabbit findings are fixed.
- [x] Ready-for-review confirmation is present.

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for forwarding the requested service tier for compatible chat providers.
  - Provider configurations can now enable or disable service-tier handling independently.

- **Bug Fixes**
  - Service-tier settings are now correctly inherited when not explicitly configured.
  - The service tier is omitted when unsupported, disabled, or not provided.

---

**Maintainer note (checklist reset by the follow-up push).** Re-ticked against fresh evidence at `197c82ae9`, not carried over: `bun x tsc --noEmit` exit 0 and `tests/openai-chat-hardening.test.ts` 39 pass / 0 fail on a Linux runner (Bun 1.3.14); branch rebased onto `dev@d03755ee9`; the one review finding (unconditional `service_tier` forwarding across 66 openai-chat providers) is fixed by the `chatServiceTier` opt-in in the follow-up commit.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -744,6 +744,10 @@ export function createOpenAIChatAdapter(provider: OcxProviderConfig): ProviderAd
         messages,
         stream: parsed.stream,
       };
+      // Preserve a caller-selected service tier for OpenAI-compatible chat gateways. The
+      // request pipeline deliberately does not inject fast mode for this adapter, but dropping
+      // an explicit value here makes the Responses parser's serviceTier projection ineffective.
+      if (parsed.options.serviceTier !== undefined) body.service_tier = parsed.options.serviceTier;
       if (modelInList(provider.reasoningSplitModels, parsed.modelId)) body.reasoning_split = true;
       const maxTokens = resolveMaxTokens(provider, parsed);
       const openRouterRouting = resolveOpenRouterRouting(provider, parsed.modelId);
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -747,7 +747,15 @@ export function createOpenAIChatAdapter(provider: OcxProviderConfig): ProviderAd
       // Preserve a caller-selected service tier for OpenAI-compatible chat gateways. The
       // request pipeline deliberately does not inject fast mode for this adapter, but dropping
       // an explicit value here makes the Responses parser's serviceTier projection ineffective.
-      if (parsed.options.serviceTier !== undefined) body.service_tier = parsed.options.serviceTier;
+      //
+      // Opt-in, like `prompt_cache_key` directly below: `service_tier` is an OpenAI-specific
+      // extension and 66 registry providers share this adapter, several of which reject
+      // unknown body fields. Forwarding unconditionally would turn a caller-supplied
+      // `service_tier` into an upstream 400 on those routes. `supportsServiceTier` is the
+      // Responses-wire flag (applyServiceTierGate) and deliberately does not gate this path.
+      if (provider.chatServiceTier && parsed.options.serviceTier !== undefined) {
+        body.service_tier = parsed.options.serviceTier;
+      }
       if (modelInList(provider.reasoningSplitModels, parsed.modelId)) body.reasoning_split = true;
       const maxTokens = resolveMaxTokens(provider, parsed);
       const openRouterRouting = resolveOpenRouterRouting(provider, parsed.modelId);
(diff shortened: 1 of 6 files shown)


Yuxin-Qiao:
PR #1521 includes this Chat Completions serialization fix (commit `b7cf9df4f`) together with the broader #1436 capability/catalog/runtime work. I am leaving this narrow PR open for maintainer choice; if #1521 is accepted, this PR can be closed as superseded.

Yuxin-Qiao:
Rebased the fork branch onto current `dev@d5824814853eea68ceb4417403b22000a935cef8`; new head is `a3535941d15b`. Post-rebase local proof: Chat/service-tier focused suite 50/50, 84 assertions; typecheck, privacy scan, and diff check passed. GitHub hygiene, target, label, and resolve checks are green. Keeping this Draft pending maintainer choice/review; #1521 carries the same Chat serialization fix.

Yuxin-Qiao:
Rebased and force-with-lease pushed onto current dev@cbbfdd8773e68a5dc2391ddeb32f33a225373c1a; new head is ab72f5441e1b. Fresh local proof: Chat/service-tier focused suite 50/50, 84 assertions; typecheck, privacy scan, and diff check passed. The PR description is updated; keeping Draft pending maintainer choice/review, with #1521 carrying the same serializer fix.

Ingwannu:
The narrow serializer change is valid and appropriately scoped: `parsed.options.serviceTier` is currently captured upstream and this adapter silently drops it. The focused test proves the outbound Chat Completions body now preserves the caller-selected value.

However, #1521 contains the same serializer change together with the provider/model capability resolver and catalog/runtime gating. Keeping both open creates two competing landing paths for the same line. I recommend treating #1512 as the minimal fallback only: leave it draft while #1521 is reviewed, and close it as superseded if #1521 is accepted. If the broader capability PR is rejected or split, this two-file patch is a reasonable standalone candidate after exact-head CI and readiness completion.

lidge-jun:
Reviewed against current `dev` (`d03755ee9`) and rebased onto it. The forwarding itself is correct — the Responses parser records a caller-supplied `service_tier` at `src/responses/parser.ts:679` and this adapter dropped it, so the value never reached the wire.

I pushed one commit on top of yours rather than changing it, because the forwarding needs a gate.

**Why.** 66 registry providers share the `openai-chat` adapter — groq, ollama, lm-studio, openrouter, together, mistral, vllm and so on. `applyServiceTierGate` (`src/server/responses/core.ts:1442`) only strips `service_tier` when the adapter is `openai-responses`, so it never fires here. Forwarding unconditionally puts an OpenAI-specific extension on every one of those routes, and strict gateways reject unknown body fields — a caller-supplied tier would turn requests that work today into upstream 400s.

The precedent is directly below your line: `prompt_cache_key` solves the same problem with an opt-in flag, and its comment in `src/types.ts` names the same hazard ("strict backends (Groq, Cerebras, etc.) reject unknown fields"). `service_tier` now follows it — `chatServiceTier` on the registry entry, threaded through `derive.ts` and `router.ts` exactly the way `promptCacheKey` is.

I deliberately did not reuse `supportsServiceTier`: that governs the Responses wire through `applyServiceTierGate` and means something different, so overloading it would make both flags harder to reason about.

Your test is kept and now asserts the opted-in case; two more assert that the field is dropped without the opt-in and that an opted-in provider with no caller tier still sends nothing. No registry entry opts in yet — that wants per-provider evidence that the gateway documents the parameter, which belongs in its own change.

Verification at the merged head (`197c82ae9`) on a Linux runner (Bun 1.3.14): `bun x tsc --noEmit` exit 0; `tests/openai-chat-hardening.test.ts` 39 pass / 0 fail. Red-before for your original test: tests-only diff on `dev` = 1 fail, full diff = 0 fail.

PR #1513, opened 2026-08-12
fix(oauth): harden generated account ids against collisions

Wibias:
## Summary

Follow-up security hardening for the OAuth multi-account store.

`newAccountId()` historically kept only 8 hex characters of SHA-256, giving generated account slots a 32-bit collision space. Two distinct identities could therefore receive the same persisted account id. Because `activeAccountId` and per-account mutations address slots by that id, a collision can make lookups or updates resolve the wrong account.

This PR:

- expands newly-derived OAuth account ids from 32 bits to 128 bits (`slice(0, 32)`)
- preserves existing persisted account ids exactly as stored, so there is no auth-store migration or forced relogin
- keeps legacy single-credential normalization deterministic across repeated loads
- adds a regression using two concrete identities whose SHA-256 digests share the historical `da1e26d2` 32-bit prefix
- verifies the colliding identities now receive distinct slots and the active credential resolves correctly
- verifies an existing persisted 8-character account id remains unchanged after credential rotation

## Security rationale

This is pre-existing hardening, separate from #1505. #1505's reasoning replay boundary also includes credential generation, so the old 32-bit slot id did not by itself cross that replay boundary. The wider issue is the OAuth store itself: duplicate generated slot ids can alias account selection and account-scoped mutations.

128 bits keeps the identifier deterministic where legacy normalization requires that property while making accidental or practical collision search infeasible for this use case.

## Scope

Only OAuth account-id derivation and focused regression coverage are changed. Existing stored ids remain backwards-compatible.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced the risk of OAuth account-ID collisions by using longer, more unique identifiers for newly created accounts.
  * Improved credential lookup when multiple accounts could previously share the same shortened identifier.
  * Preserved existing persisted account IDs, ensuring previously stored credentials remain compatible and can still be updated.

* **Tests**
  * Added coverage for collision handling, account selection, and backward compatibility with existing OAuth account data.

### DIFF
--- a/src/oauth/store.ts
+++ b/src/oauth/store.ts
@@ -198,7 +198,7 @@ interface LockSnapshot { bytes: string; dev: number; ino: number; mtimeMs: numbe
 export interface OAuthFileLockOptions { path: string; waitTimeoutMs?: number; staleAfterMs?: number; pollMinMs?: number; pollMaxMs?: number; sleep?: (ms: number) => Promise<void>; now?: () => number; random?: () => number; beforeStaleUnlink?: () => void; beforeReleaseUnlink?: () => void; beforeFailedCreateUnlink?: () => void; writeMetadata?: (fd: number, bytes: string) => void }
 export interface OAuthFileLockGuard { readonly ownerId: string; release(): void }
 function errorCode(error: unknown): string | undefined { return error && typeof error === "object" && "code" in error ? String((error as { code?: unknown }).code) : undefined; }
-function snapshot(path: string): LockSnapshot { const bytes = readFileSync(path, "utf8"); const s = statSync(path); return { bytes, dev:s.dev, ino:s.ino,mtimeMs:s.mtimeMs,size:s.size }; }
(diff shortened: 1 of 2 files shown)


Wibias:
CI note: the initial 25 GUI failures are pre-existing on the PR's exact base `dev@d667367528afa2c604d19af2fc8538b965b88cf1`, not introduced by this OAuth change. The base push run `31564782346` failed `gates` with the same 25 test names (`743 pass / 25 fail`), while all four backend test shards passed. #1513 changes only `src/oauth/store.ts` plus the focused OAuth collision regression and has no `gui/` imports of that store. I reran only the failed `gates` job on exact head `76ad058cf9b409a6ff9bfe5ad778175f477d32d6`; attempt 2 passed GUI tests, typecheck, privacy scan, and the rest of the gate. Cross-platform CI and React Doctor are now green on the exact head.

PR #1529, opened 2026-08-12
fix(codex): keep direct MCP tools visible for routed models

### LINKED ISSUE #1522: [Provider compatibility] Codex App plugins unavailable for routed models when catalog forces supports_search_tool=true
### Client or integration

Codex App

OpenAI-compatible third-party provider

### OpenCodex version

2.11.1

### Endpoint or capability

/v1/responses; Codex App plugin/MCP tool discovery

### Current behaviour

OpenCodex generates Codex model-catalog rows for non-Cursor routed models with:

```json
{
  "supports_search_tool": true,
  "web_search_tool_type": "text_and_image"
}
```

With this generated catalog entry, Codex App does not expose the Browser plugin's deferred MCP tools to the routed model. In a fresh task, the model cannot see or call tools such as `mcp__node_repl__js`, and reports that the browser runtime cannot be initialized.

A controlled A/B test changes only `supports_search_tool`:

- `true` (OpenCodex-generated default): Browser plugin tools are absent from the model-visible tool surface.
- `false`: a new Codex App task exposes the direct MCP tools, including `mcp__node_repl__js`, and the routed model can call them.

No provider, model, plugin, Codex App, or OpenCodex version changed between the two runs.

### Expected behaviour

OpenCodex should generate a catalog capability profile that makes Codex App plugins usable for routed third-party models.

If the `tool_search` compatibility path is supported for a provider/model, deferred plugin discovery should complete and load the requested tools. If that path is not verified for the routed model, OpenCodex should advertise `supports_search_tool: false` so Codex exposes compatible direct MCP tools instead.

A generated capability flag should not make the Browser plugin silently unavailable.

```shell
Environment:
- macOS
- Codex App 26.803.61601
- bundled Codex CLI 0.147.0-alpha.6.5
- OpenCodex 2.11.1
- OpenAI-compatible third-party provider
- DeepSeek-compatible routed model
- Browser plugin enabled

1. Let OpenCodex generate/sync its Codex model catalog.
2. Confirm the routed model entry contains:
   "supports_search_tool": true
3. Start a fresh Codex App task with the routed model.
4. Ask it to use the Browser plugin.
5. Observe that deferred Browser MCP tools such as mcp__node_repl__js are not present in the model-visible tool surface.
6. Change only the routed entry to:
   "supports_search_tool": false
7. Apply the same value to the active Codex model cache, then start another fresh task.
8. Ask it to use the Browser plugin again.
9. Observe that direct MCP tools, including mcp__node_repl__js, are now present and callable.
```

### Actual response or error

```shell
No HTTP error is returned.

Failing/default case:
- supports_search_tool: true
- Browser deferred MCP tools absent
- routed model concludes that the required browser runtime tool is unavailable

Working A/B case:
- supports_search_tool: false
- mcp__node_repl__js appears in the tool surface
- the routed model successfully issues direct MCP calls

This is a tool-discovery/capability-advertisement failure rather than an upstream model HTTP failure.
```

### Upstream documentation

Concrete client requirement: Codex App's enabled Browser plugin must remain discoverable and callable when using an OpenCodex-routed model.

Relevant OpenCodex implementation:
https://github.com/lidge-jun/opencodex/blob/main/src/codex/catalog/parsing.ts

Related but different prior issue demonstrating the intended routed-model tool_search compatibility path:
https://github.com/lidge-jun/opencodex/issues/122

In `normalizeRoutedCatalogEntry`, non-Cursor routed rows currently receive `supports_search_tool = true` unconditionally.

Suggested options:

1. Make search-tool support a provider/model capability instead of a blanket routed-model default.
2. Verify the generated `tool_search` path end-to-end with Codex App plugins before advertising it.
3. Default unverified routed models to `supports_search_tool: false`, preserving direct MCP exposure.
4. Add a regression test covering Browser plugin discovery for both catalog values.

I searched existing provider and compatibility issues. No matching report was found.

Issue #122 is related only as evidence that OpenCodex intentionally translates Codex `tool_search` for routed models. It covers an incompatible loaded schema after discovery, whereas this report covers plugin tools not becoming available at all.

All provider names, model routing identifiers, local paths, task IDs, account details, and internal service information were removed. This issue intentionally excludes a separate MCP result-formatting observation so the reproduction stays focused on discovery.

### Checks

Ingwannu:
## Summary

- default routed Codex catalog rows to `supports_search_tool: false` so Codex App keeps compatible direct MCP/plugin tools visible
- preserve the independent sidecar-backed `web_search_tool_type: "text_and_image"` capability for non-Cursor routes
- keep the runtime `tool_search` parser and bridge intact for clients that explicitly send that wire shape
- document why deferred discovery and hosted web search must remain separate capability decisions

Closes #1522

## Verification

- `bun scripts/test.ts tests/catalog-cursor-search.test.ts tests/codex-catalog.test.ts tests/e2e-style/phase100-native-parity.test.ts` (176 passed, 0 failed)
- `env -u OPENCODEX_API_AUTH_TOKEN bun scripts/test.ts tests/codex-shim.test.ts` (67 passed, 0 failed; confirms the unrelated full-suite host-token interference is environmental)
- `bun run typecheck`
- `bun run privacy:scan`
- `git diff --check`
- A full `bun run test` progressed through the affected catalog and broad integration suites without a product assertion failure, but the long-running process was externally terminated with SIGTERM/143 before completion; exact-head repository CI remains required before merge.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected search capability reporting for routed models.
  * Routed models now consistently indicate that direct search-tool support is unavailable while preserving hosted web search metadata where applicable.
  * Cursor-routed models no longer advertise hosted web search metadata.

* **Documentation**
  * Added guidance on routed tool discovery and hosted search behavior.

### DIFF
--- a/src/codex/catalog/parsing.ts
+++ b/src/codex/catalog/parsing.ts
@@ -393,17 +393,18 @@ export function normalizeRoutedCatalogEntry(entry: RawEntry, parallelToolCalls =
   // Per-model routed opt-ins can be added once provider metadata exposes this capability.
   delete entry.supports_reasoning_summaries;
   const isCursorEntry = typeof entry.slug === "string" && entry.slug.startsWith("cursor/");
-  // Routed providers use opencodex sidecars and client-executed tool discovery. The sidecar
-  // runs through native gpt-5.4-mini, so image search is available and verbalized for text-only
-  // models. EXCEPT cursor: its runTurn transport bypasses the web-search plan entirely and
-  // rejects server search queries — advertising the tool would make models call into a void.
+  // `supports_search_tool` selects Codex's deferred tool-discovery surface; it is not the hosted
+  // web-search capability. OpenCodex can round-trip tool_search when a client sends it, but routed
+  // providers have no provider/model proof that Codex App plugins work through that deferred
+  // surface. Advertising it unconditionally hides the App's compatible direct MCP tools (#1522),
+  // so routed rows fail closed to direct discovery. The sidecar-backed hosted web-search metadata
+  // remains advertised independently for non-Cursor routes.
   if (isCursorEntry) {
     delete entry.web_search_tool_type;
-    entry.supports_search_tool = false;
   } else {
     entry.web_search_tool_type = "text_and_image";
-    entry.supports_search_tool = true;
   }
+  entry.supports_search_tool = false;
   // Cursor's transport already serializes overlapping tool calls into atomic Responses tool events.
   // Advertising parallel calls lets Codex send the same native capability bit it sends for OpenAI.
   // Opt-in providers (OcxProviderConfig.parallelToolCalls, e.g. xAI) advertise it too: the
--- a/src/codex/catalog/sync.ts
+++ b/src/codex/catalog/sync.ts
@@ -313,7 +313,7 @@ export function deriveEntry(
     slug, display_name: routedDisplayName(slug), description: desc,
     shell_type: "shell_command", visibility: "list", supported_in_api: true,
     priority, base_instructions: "You are a helpful coding assistant.",
-    ...(isRouted ? { web_search_tool_type: "text_and_image", supports_search_tool: true } : {}),
+    ...(isRouted ? { web_search_tool_type: "text_and_image", supports_search_tool: false } : {}),
   };
   if (isRouted) {
     applyRoutedCodexToolMode(entry);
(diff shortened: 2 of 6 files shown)


Ingwannu:
Exact-head GitHub CI is now fully green on `29fd4f02484250ebaf6ec256f18a7f8a31b25dd6`: all four Linux shards, gates/typecheck/privacy, macOS, storage/API, keyring, npm-global, React Doctor, hygiene, and target checks completed successfully.

The PR remains draft pending independent maintainer review. I am not self-approving or self-merging it.

Issue #1514, opened 2026-08-12
[Bug]: deepseek-v4-flash-free can emit incomplete tool calls that break streamed turns


yxr1995-maker:
### Current status

This issue is now narrowed to the remaining OpenCode Zen / DeepSeek streaming compatibility defect.

The startup syntax crash reported against 2.12.0 is already resolved on current `dev`: the invalid `continue` outside a loop is no longer present. That part should not be reimplemented.

Current `dev` also already accepts `[DONE]` without a `finish_reason`: the `[DONE]` branch flushes pending tool calls and emits `done`. A raw EOF with neither `[DONE]` nor `finish_reason` remains a truncation condition and should stay fail-closed unless provider evidence proves a safe alternative.

### Remaining defect

`openai-chat` can still flush a pending tool call with an empty function name.

Current `flushToolCalls()` emits:

```ts
yield { type: "tool_call_start", id: call.id, name: call.name };
```

without first proving that `call.name` is non-empty.

The affected OpenCode Zen route can stream argument deltas while omitting the matching `function.name`. When that incomplete call is flushed, the downstream Codex tool-call contract receives an unusable call and the turn can fail.

This issue now tracks only the provider-compatibility question around incomplete streamed tool-call metadata.

### Reproduction

1. Route Codex through `opencode-free/deepseek-v4-flash-free` using the `openai-chat` adapter.
2. Run a turn that causes the upstream to emit tool-call argument deltas without a usable `function.name`.
3. Allow the stream to reach a normal flush boundary such as `[DONE]`.
4. Observe the pending call being emitted with an empty name and the downstream turn failing.

### Expected behavior

OpenCodex must not emit a completed usable tool call with an empty name.

A fix must preserve the existing fail-closed rules for genuinely malformed or truncated tool calls. In particular:

- do not silently invent a function name;
- do not treat arbitrary raw EOF as successful completion;
- do not weaken the nested `tool_calls` validation from #1325;
- do not silently drop a claimed tool call if doing so can orphan a matching result;
- provider-specific normalization is acceptable only when the upstream shape is proven semantically safe.

### Resolved portion

The 2.12.0 startup failure:

```text
Cannot use "continue" here
```

is no longer present on current `dev` and is closed as part of this issue's history.

### Related

- #1325 - generic nested `tool_calls` validation and fail-closed behavior.
- #1483 - Xiaomi MiMo compatibility report. Similar symptom, different provider evidence and reasoning-effort scope.
[...]

PR #1530, opened 2026-08-12
fix(openai-chat): log redacted invalid tool-call shape

Ingwannu:
## Summary

- add opt-in structural diagnostics at the existing fail-closed OpenAI Chat `tool_calls` validation boundary
- record only a finite reason code, stream/non-stream mode, array member index, and JavaScript value type
- never retain tool-call IDs, function names, tool arguments, request bodies, credentials, or account data
- preserve current parsing behavior; this PR gathers evidence for #1483 and does not guess at a permissive Xiaomi MiMo compatibility mapping

Closes no issue; supports diagnosis of #1483.

## Verification

- `bun scripts/test.ts tests/openai-chat-hardening.test.ts tests/issue-452-empty-503.test.ts tests/debug.test.ts` (66 passed, 0 failed)
- `bun run typecheck`
- `bun run privacy:scan`
- `git diff --check`

To capture one redacted reproduction after this lands:

```shell
ocx debug provider on
ocx debug provider logs -f
# reproduce once
ocx debug provider off
```

Expected diagnostic shape:

```text
[ocx:openai-chat:invalid-tool-calls] {"mode":"stream","reason":"tool_call_not_object","callIndex":1,"valueType":"null"}
```

## Checklist

- [x] Docs or release notes were updated when needed. No user-visible behavior/configuration changes; existing provider-debug commands expose the diagnostic.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. Diagnostics remain opt-in and values are never serialized.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved diagnostics when malformed tool-call data is received.
  - Added clearer identification of the processing mode, validation reason, item position, and value type.
  - Ensured sensitive tool-call values are never included in diagnostic logs.
  - Preserved existing invalid tool-call error behavior and validation precedence.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -261,9 +261,11 @@ function diagnoseInvalidToolCalls(
       }
       continue;
     }
-    if (typeof rawToolCall.id !== "string") {
-      return { reason: "tool_call_id_invalid", callIndex, valueType: typeof rawToolCall.id };
-    }
+    // Precedence must mirror the buffered validator below, or a payload with more than one
+    // problem is reported under the wrong reason and sends compatibility work after the wrong
+    // shape. That validator checks the `function` container first (`!isRecord(rawToolCall) ||
+    // !isRecord(rawToolCall.function)`), then id/name/arguments types together, and only then
+    // the blank name.
     if (!isRecord(rawToolCall.function)) {
       return {
         reason: "tool_call_function_not_object",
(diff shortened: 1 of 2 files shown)


lidge-jun:
Rebased onto current `dev` (`d03755ee9`) and pushed two commits on top of yours; your commit `933d00974` is preserved.

**Why it conflicted.** #1531 merged between this being written and being rebased, and it changed the exact validation this diagnostic describes. Two consequences:

The streamed path now validates `function`/`name`/`arguments`/`id` at ingest rather than only checking container and member shapes. `diagnoseInvalidToolCalls` explicitly `continue`d in stream mode, which was right against the old code and wrong against the new — the three new ingest rejections would have produced no diagnostic at all. Stream mode now inspects those same fields and all three sites log through the helper.

The buffered path now rejects blank and whitespace-only names. The diagnostic would have reported that as `tool_call_function_name_invalid` — a type problem for a correctly-typed value — so it gets its own `tool_call_function_name_blank` code.

**A second review round then caught a precedence bug in my own follow-up.** The validator checks the `function` container first, then id/name/arguments types together, and only then the blank name. My diagnosis checked `id` before the container and blank name before arguments, so a payload carrying two defects at once reported the wrong reason. That is exactly the failure this diagnostic is supposed to prevent — a wrong reason code sends provider-compatibility work after the wrong shape. Fixed, with four tests that each carry two defects simultaneously so only the matching order passes.

Privacy was re-verified independently: reason codes and modes are finite literals, `valueType` is a fixed label or a bare `typeof`, and the buffer stores only the already-redacted serialized line. No caller-controlled string reaches the log.

On whether this is still worth landing now that #1531 fixed #1514: yes. #1531 fixed one known shape; this records which check rejected a payload, which is what makes the next provider incompatibility diagnosable instead of a guess. It stays opt-in and off by default.

Verification at `be101c32b` on a Linux runner (Bun 1.3.14): `bun x tsc --noEmit` exit 0; `openai-chat-hardening` + `eof` + `parallel-stream` 91 pass / 0 fail; 11-file adapter sweep 165 pass / 0 fail; `privacy:scan` passed. Your tests remain valid post-rebase and go red on `dev` without the source change.

lidge-jun:
Full-suite result at `be101c32b` on the Linux runner (Bun 1.3.14):

```
11341 pass
0 fail
EXIT=0
```

PR #1531, opened 2026-08-12
fix(openai-chat): refuse to emit an unnamed streamed tool call

### LINKED ISSUE #1514: [Bug]: deepseek-v4-flash-free can emit incomplete tool calls that break streamed turns

lidge-jun:
## Summary

An OpenCode Zen / DeepSeek route can stream `function.arguments` deltas while never sending `function.name`. The pending call kept its initial empty name and every flush site emitted it anyway, so the Codex tool-call contract received a call it cannot dispatch and the streamed turn failed downstream (#1514).

This makes the streamed path fail closed. An unnamed call now terminates the turn through the adapter error channel instead of being emitted.

Three details worth calling out for review:

- **Why not synthesize a name.** The `id` is already synthesized when absent, and that is safe because an id is an opaque correlation handle. A function name is a guess at intent, so it is refused instead. This matches the reference implementations: `openai-python` and `openai-node` accumulate such a call with an empty name and let the caller fail; LiteLLM forwards it unchanged. None invent a name.
- **Why not drop it.** #1325 established that a claimed tool call which silently disappears can leave the matching `function_call_output` orphaned on the next turn. A loud terminal error preserves that invariant rather than weakening it.
- **Ingest-time validation.** The `delta.tool_calls` cast is a TypeScript convenience over upstream JSON, so a truthy non-string `name` reached the accumulator unvalidated and then threw from string handling. Review caught this in the first draft of the patch (reproduced as `TypeError: call.name.trim is not a function`). Present `function`, `name`, `arguments`, and `id` fields are now runtime-checked at ingest and route to the existing `invalidToolCallsEvent`.

The buffered `parseResponse` validator checked the *type* of `name` but not its emptiness, so it is tightened the same way. A completed Chat Completions tool call requires a name in the official OpenAPI schema, and a blank or whitespace-only name cannot select a dispatch target.

Preserved invariants, each covered by a test: raw-EOF truncation still reports truncation rather than the new error; non-array and non-record `tool_calls` still terminate per #1325; a name arriving in a later chunk is still accepted; `done` is never emitted after the new error.

`T7` in `tests/openai-chat-parallel-stream.test.ts` previously asserted the empty-name call *was* flushed, on a "no silent drop" rationale. That invariant still holds and is strengthened — the call does not vanish, the turn fails loudly — so the test was updated to the new contract rather than deleted. `T6` (late-arriving name) is untouched.

## Verification

Run on a dedicated Linux runner (Bun 1.3.14, clean checkout of this branch):

- `bun test tests/openai-chat-eof.test.ts tests/openai-chat-parallel-stream.test.ts` — 48 pass, 0 fail
- 11-file adapter sweep (`tests/openai-chat*`, `tests/cl01-openai-chat*`, `tests/adapter*`) — 161 pass, 0 fail

Red-before evidence: with **only** the test diff applied to unmodified `dev`, 5 tests failed. With the full diff, 48/48 pass.

The crash path found in review was reproduced directly before and after the fix: a numeric `function.name` produced `THREW: call.name.trim is not a function` before, and `EVENTS: error` after.

Budget safety is asserted directly, not inferred: a stream with a valid call, an unnamed call, and a second valid call after it ends with `activeCalls: 0`, `currentBytes: 0`, and `highWaterBytes > 0`, proving `closeToolCalls()` released every pending reservation including the call after the offender.

Full-suite run via `bun run test` was still in progress at the time of opening; it will be reported here when it completes.

## Checklist

Closes #1514

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid or unnamed tool calls are now rejected safely in streaming and non-streaming responses.
  - Malformed tool-call data no longer produces unusable events or falsely completes a stream.
  - Pending processing resources are released when tool-call validation fails.

- **Tests**
  - Added coverage for missing, blank, malformed, and delayed tool-call names.
  - Verified preservation of existing truncation and invalid-payload diagnostics.

### DIFF
--- a/devlog/_plan/260812_five_bug_fix_campaign/000_research.md
+++ b/devlog/_plan/260812_five_bug_fix_campaign/000_research.md
@@ -1,0 +1,258 @@
+# 000 — Research: five confirmed bug issues, live code grounding
+
+Unit: `260812_five_bug_fix_campaign`
+Baseline: `origin/dev` at `cbbfdd877` (`fix(release): filter successful gate runs from JSON`).
+Date: 2026-08-12.
+
+## Why this unit exists
+
+A triage pass over the 16 open `bug`-labelled issues found that **no merged PR
+currently closes any of them**: the only merged PR carrying a closing reference
+in the recent window is #1501 → #1477, and #1477 is already closed. Two issues
+have draft PRs in flight (#1503 → draft #1508, #1497 → stale draft #1008) and
+the rest have no implementation at all.
+
+Five issues were selected because each has a **code-level defect already located
+in the current tree**, not a missing-evidence or upstream-attribution problem:
+
+| Issue | Area | Located in current tree |
+|---|---|---|
+| #1514 | `openai-chat` adapter | `src/adapters/openai-chat.ts:900-907` |
+| #1503 | `google` adapter | `src/adapters/google.ts:615-618`, `:836` |
+| #1497 | management usage API | `src/server/management/logs-usage-routes.ts:210-214` |
+| #1409 | provider config write | `src/server/management/provider-routes.ts:353-363` |
+| #1419 | service lifecycle | `src/lib/abort.ts:125-133` neighbourhood |
+
+Explicitly out of scope for this unit: #1527, #1524, #1388, #1302, #1296,
+#1059, #1049, #1024, #904, #417, #92. Those are either needs-info, upstream
+tracking, policy-level enhancements, or CI burn-down work.
+
+## Per-defect grounding
+
+### #1514 — `flushToolCalls` can emit a tool call with an empty name
+
+`parseStream` accumulates streamed tool calls into `PendingToolCall` records
+whose `name` starts as the empty string:
+
+```ts
+call = { key: key ?? `seq:${pendingToolCalls.length}`, id: "", name: "", args: "", argsBytes: 0 };
+```
+
+`name` is only ever populated from a truthy upstream field:
+
+```ts
+if (tc.function?.name && !call.name) call.name = tc.function.name;
+```
+
+So an upstream that streams `function.arguments` deltas while never sending
+`function.name` leaves `call.name === ""`. At every flush boundary the call is
+emitted regardless:
+
+```ts
+const flushToolCalls = function* (): Generator<AdapterEvent> {
+  for (const call of closeToolCalls()) {
+    if (!call.id) call.id = `call_${++toolCallSeq}`;
+    yield { type: "tool_call_start", id: call.id, name: call.name };   // name may be ""
+    if (call.args.length > 0) yield { type: "tool_call_delta", arguments: call.args };
+    yield { type: "tool_call_end" };
+  }
+};
+```
+
+Note the asymmetry that makes this a real bug rather than a style complaint:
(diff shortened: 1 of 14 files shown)


lidge-jun:
Full-suite result, as promised in the description.

`bun run test` on the Linux runner (Bun 1.3.14, clean checkout of this branch at `15c6b53ca`):

```
11310 pass
0 fail
Ran 11321 tests across 698 files. [410.08s]
EXIT=0
```

PR #1535, opened 2026-08-12
fix(providers): keep hand-edited context windows through a POST overwrite

### LINKED ISSUE #1409: [Bug]: manual opencode-go modelContextWindows entries are reset after update
### Client or integration

Codex App

### Area

Catalog / models / config persistence

## Current status

This issue remains open, but its scope is narrower than the original report suggested.

Current `dev` contains config-rebase work that fixed at least one confirmed management/save path which could lose provider configuration. However, the broader whole-document writer fix attempted in #1462 was closed without merge, and the original upgrade + later unrelated-write reproduction has not been proven fully covered.

Do not close this issue based only on the presence of the newer rebase-on-save machinery. The close condition is that the original user override survives the actual lifecycle described below, including upgrade/restart and an unrelated later save.

The architectural ambiguity behind stale-config rebasing is tracked separately in #1478.

## Problem

An explicit user value under:

```text
providers.opencode-go.modelContextWindows
```

was removed after upgrading and then performing a later unrelated config/provider write. The provider fell back to registry-seeded context metadata.

Registry defaults may seed missing values, but they must not overwrite or delete explicit user overrides.

## Original reproduction

1. Configure:

```json
{
  "providers": {
    "opencode-go": {
      "modelContextWindows": {
        "deepseek-v4-flash": 900000
      }
    }
  }
}
```

2. Run OpenCodex 2.11.0 and verify the override is effective.
3. Upgrade to 2.11.1 and restart the daemon/app.
4. Perform an unrelated provider/config write.
5. Inspect `config.json`.
6. The explicit `deepseek-v4-flash` entry was gone and `opencode-go.modelContextWindows` contained only registry-seeded values.
7. The generated catalog then fell back to its default context window.

The separate `deepseek` provider's context-window configuration was not affected in the original report.

## Required behavior

- Explicit user `modelContextWindows` values survive upgrade and restart.
- Registry defaults only fill genuinely missing values.
- An unrelated management/API/provider save must not remove an override that the writer did not intend to change.
- A deliberate deletion of an override must still remain possible and must not be resurrected by a rebase.
- Concurrent or independently loaded config writers must not silently erase unseen keys.

The last two requirements interact with the provenance problem in #1478. Fixing one by blindly treating every absent key as deletion, or every absent key as ignorance, is not sufficient.

## Acceptance test

A regression should exercise the complete lifecycle rather than only a single helper:

```text
seed explicit opencode-go modelContextWindows override
        ↓
load/start server
        ↓
perform registry/default reconciliation if applicable
        ↓
perform an unrelated management/config write
        ↓
reload from disk
        ↓
explicit override is still present and effective
```

Also cover deliberate deletion so the persistence fix does not make stale values impossible to remove.

## Close condition

Close only when the original lifecycle is covered on current `dev` by a regression proving both preservation of explicit user overrides and intentional deletion semantics.

## Related

- #1462 - closed, unmerged broad config-writer attempt
- #1478 - config rebase provenance for deletion vs unseen keys

lidge-jun:
## Summary

`POST /api/providers` loses hand-edited context windows when it overwrites an existing provider.

`ProviderPayload` (`gui/src/provider-payload.ts`) has no member for `contextWindow` or `modelContextWindows`, so the dashboard's add/edit form structurally cannot send them. On an overwrite those fields are therefore absent, `enrichProviderFromCatalog` → `enrichProviderFromRegistry` fills them from the registry seed, and the stored row becomes the seed. For `opencode-go` the seed is exactly `{"kimi-k3": 262144}` — which is what #1409 reports finding in place of a hand-edited `{"deepseek-v4-flash": 900000}`.

The fix follows the precedent already in this function: `apiKeyPool` and `modelCosts` are carried across the same path, with the same rationale — the form does not send them, so absence must not mean deletion. These two fields are the same class of user-owned data and were simply never added to the list.

Two details that matter for review:

- **Ownership is sampled before enrichment.** After `enrichProviderFromCatalog` runs, "the client omitted this field" and "the registry supplied it" are indistinguishable, so a carry-over guard written as `prov.x === undefined` afterwards can never fire. An earlier draft had exactly that bug and its test would have stayed red.
- **When the client omits the map, the stored value is the user's map alone.** Merging the registry seed in would persist seed keys into user config as a side effect of an unrelated save, and `router.ts` already fills registry values *beneath* user entries at resolve time via `mergeRecordFill`.

Deletion is unaffected: it goes through PATCH with an explicit `null`, which the carry-over must not undo, and that is covered by a test.

## Why this does not carry `Closes #1409`

The confirmed reproduction is a **duplicate-name POST through the Add Provider modal**. The dashboard's ordinary editing surfaces use PATCH (`gui/src/pages/use-providers-crud.ts`), and `Models.tsx` sends `modelContextWindows` over PATCH, which merges per key and already preserves unmentioned entries.

The reporter's sequence was an upgrade, a daemon restart, and a later unrelated full-config write. That points at the stale whole-document writer tracked in #1273, and nothing here rules it in or out. So this PR fixes a real, demonstrated data-loss defect on its own merits, and #1409 gets a comment describing what was confirmed rather than a closure it has not earned.

## Verification

On a Linux runner (Bun 1.3.14):

- `bun test tests/management-provider-validation.test.ts` — 60 pass, 0 fail
- 60-file provider/config/management sweep — 1023 pass, 0 fail

Red-before evidence: with only the test diff applied to unmodified `dev`, 3 of the new tests fail.

New coverage: an omitted map keeps the user's map **and** does not gain registry seed keys; a submitted map updates that key while other user keys survive; an omitted scalar `contextWindow` is preserved; a submitted scalar still wins; a brand-new provider still receives the registry seed (no regression in enrichment); and PATCH can still delete a key with an explicit null.

This touches a management write boundary, so it wants the security review `src/AGENTS.md` asks for — no auth, credential, or transport behavior changes, but the persisted-config write path does.

## Checklist

Refs #1409, #1273

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved existing provider context-window settings when updates omit those fields.
  - Correctly merged submitted per-model context-window values during provider updates.
  - Prevented registry defaults from being saved as user-configured settings.
  - New providers still receive appropriate registry defaults.
  - Explicitly clearing model-specific settings continues to work as expected.

- **Tests**
  - Added regression coverage for provider creation, updates, merging, preservation, and clearing of context-window settings.

### DIFF
--- a/devlog/_plan/260812_five_bug_fix_campaign/040_phase4_issue1409_context_window_overrides.md
+++ b/devlog/_plan/260812_five_bug_fix_campaign/040_phase4_issue1409_context_window_overrides.md
@@ -271,8 +271,11 @@ the untouched user key survives.
 
 ```bash
 bun x tsc --noEmit
-bun test tests/provider-routes*.test.ts tests/management*.test.ts tests/config*.test.ts
+bun test tests/management-provider-validation.test.ts tests/management*.test.ts tests/config*.test.ts
 ```
+
+(There is no `tests/provider-routes*.test.ts`; an unmatched glob aborts the run under zsh.
+The management-provider validation suite is where this path's coverage lives.)
 
 ## Delivery
 
--- a/src/server/management/provider-routes.ts
+++ b/src/server/management/provider-routes.ts
@@ -350,6 +350,12 @@ export async function handleProviderRoutes(ctx: ManagementContext): Promise<Resp
     }
     // Catalog providers (e.g. ollama-cloud) carry a models + vision/reasoning classification the GUI
     // doesn't send — merge it in so the sidecars are gated correctly.
+    // Sample request ownership BEFORE enrichment. Enrichment fills absent fields from the
+    // registry seed, after which "the client omitted this" and "the registry supplied it" are
+    // indistinguishable — so a carry-over guard written as `prov.x === undefined` after this
+    // call can never fire.
+    const submittedContextWindow = Object.hasOwn(prov, "contextWindow");
+    const submittedModelContextWindows = Object.hasOwn(prov, "modelContextWindows");
     enrichProviderFromCatalog(name, prov);
     const { saveConfigPreservingClaudeCode: save } = await import("../../config");
     // Overwriting an existing provider must not drop its multi-key pool: carry it over, then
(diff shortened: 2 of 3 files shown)


lidge-jun:
Full-suite result on the Linux runner at this branch head:

```
11316 pass
0 fail
EXIT=0
```

PR #1538, opened 2026-08-12
fix(codex): bind prompt journal recovery paths

luvs01:
## Summary

- Bind decoded prompt-journal records to the config and store paths selected by the active prompt-layer runtime before reading either recorded target.
- Perform every recovery read, revalidation, restore, and diagnostic through those trusted runtime paths, so a checksum-valid journal cannot redirect rollback to another filesystem target.
- Preserve lexical Windows path identity and existing crash recovery while adding direct and public-callsite regressions for forged config/store paths.
- This is preventive filesystem-authority hardening for the currently dormant prompt-composer write surface; it does not add a new runtime route or claim journal authenticity.

## Verification

- `Bun 1.3.14: bun test --isolate tests/codex-prompt-journal.test.ts tests/codex-prompt-layers-write.test.ts` — 45 pass
- `Bun 1.4.0-canary.1: bun test --isolate tests/codex-prompt-journal.test.ts tests/codex-prompt-layers-write.test.ts` — 45 pass
- `bun x tsc --noEmit` — passed on Bun 1.3.14 and Bun 1.4
- `bun run privacy:scan` — passed on Bun 1.3.14 and Bun 1.4
- `git diff --check`
- Focused security and final diff reviews found no P0-P3 blocker.
- The full local suite was not duplicated; GitHub CI remains the cross-platform/full-suite authority.

## Checklist

## Review readiness checklist

### DIFF
--- a/src/codex/prompt-journal.ts
+++ b/src/codex/prompt-journal.ts
@@ -20,7 +20,7 @@
  * recovery sees a mismatch and overwrites their work with a stale image.
  */
 import { existsSync, mkdirSync, openSync, closeSync, fsyncSync, readFileSync, renameSync, unlinkSync, writeFileSync } from "node:fs";
-import { dirname } from "node:path";
+import { dirname, resolve } from "node:path";
 import { createHash, randomBytes } from "node:crypto";
 import { forgetEphemeralSecretPath, hardenSecretPath, windowsSecretAclApplies } from "../lib/windows-secret-acl";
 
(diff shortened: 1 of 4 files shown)


lidge-jun:
Security review complete — **LAND**. Independent threat-model review plus my own reading of the binding.

The design decision that makes this work is stated in the code itself: *"Lexical path identity only: never resolve an untrusted journal path through the filesystem."* Resolving a recorded path would reintroduce exactly the redirection this closes, so comparing lexically is the correct instinct rather than a shortcut.

Verified against the three questions:

- **Does the defense close the path?** Yes. Decoded target paths are compared against the active runtime paths *before* either target is read (`prompt-journal.ts:243-278`), and every later read, diagnostic, and restore uses the trusted expected paths rather than the recorded ones.
- **Any bypass?** None found. Relative segments normalize lexically; Windows comparison is case-insensitive while Unix stays case-sensitive; UNC identity falls out of the platform `resolve`. A symlink planted at an accepted target does not buy arbitrary-target restoration either — `durableWrite` renames over the symlink and deletion unlinks it rather than following it.
- **Any legitimate use broken?** No. The envelope schema is unchanged and earlier records already carry the active config/store paths, so previous-version journals still recover. A journal deliberately tied to a different runtime path now fails closed, which is the intent.

One correction to the PR description: the surface is **not** dormant. It is live on prompt-layer mutations through `setToggle` and `writeCustomLayers` (`prompt-layers.ts:653`, `:748-780`). It is not reached by service startup or ordinary reads, which is probably what was meant — but "currently dormant" undersells it, and a reader deciding how urgently to pick this up deserves the accurate version.

Verification on a Linux runner (Bun 1.3.14): red-before 21 pass / 2 fail + 1 load error — forged-journal redirection demonstrated independently; at head 45 pass / 0 fail; `bun x tsc --noEmit` exit 0; `bun run privacy:scan` pass.

Issue #1419, opened 2026-08-10
[Bug][macOS] Bundled Bun 1.3.14 exits with SIGTRAP after connection reset and TLS verification failures


RanceLee233:
### Client or integration

Codex App

### Area

Service lifecycle

### Summary

OpenCodex 2.11.1 went offline twice without an `ocx stop` command. In both cases, the bundled Bun process terminated with a native `EXC_BREAKPOINT / SIGTRAP` on its main thread shortly after a connection reset followed by TLS certificate verification failures.

After the second crash, the page opened by `ocx gui` also became unreachable because the dashboard was served by the same terminated process.

Expected behavior: a transient upstream connection/TLS failure should fail or retry the affected request while the proxy stays alive, or at minimum exit with an actionable OpenCodex error. It should not terminate the entire proxy with a native trap.

### Reproduction

The issue is intermittent rather than deterministically reproducible. The observed sequence was:

1. Run OpenCodex 2.11.1 on macOS arm64 and use Codex through the proxy for sustained traffic.
2. Open the dashboard with `ocx gui`. When no live proxy exists, this starts the proxy in the background.
3. During a period of network instability, an established upstream connection is reset.
4. OpenCodex retries; the retry fails with `unknown certificate verification error`.
5. The next request also fails certificate verification.
6. About 0.5–0.6 seconds later, the bundled Bun process exits with `SIGTRAP`.
7. The proxy and dashboard both become unreachable and are not restarted.

This happened twice with the same Bun image UUID, matching main-thread stack offsets, and matching instruction bytes.

Observed timeline (Asia/Shanghai, UTC+8):

- 16:43:47: consecutive TLS verification failure
- 16:43:48.062: Bun PID 61320 crashes with `SIGTRAP`
- 16:49:33: `ocx gui` is run and starts a new proxy process
- 17:26:28: consecutive TLS verification failure
- 17:26:29.187: Bun PID 66546 crashes with the same signature
- 17:28:30: `ocx stop` appears in shell history, after the crash; it did not cause the outage

There was no OpenCodex launchd service installed, so the detached process was not supervised or restarted after the native crash.

### Version

2.11.1

Bundled Bun: 1.3.14

### Operating system

macOS 26.5.2, Apple Silicon (arm64)

### Provider and model

OpenAI Responses API; observed across multiple GPT models, including gpt-5.6-sol and gpt-5.6-luna.

### Logs or error output

```shell
# Crash report 1
[...]

PR #1539, opened 2026-08-12
fix(server): settle upstream bodies on the failure paths too

lidge-jun:
## Summary

Hardening found while investigating #1419. **This is not a fix for the native SIGTRAP reported there**, and that issue stays open pending the crash frames — see the disposition comment on it.

`cancelBodyOnAbort` exists because Bun rejects an in-flight internal read when a fetch response body is torn down before our code attaches a reader, and that rejection is orphaned off the awaited path where no caller `try/catch` can intercept it (`src/lib/abort.ts`). The guard was installed on the **success** paths only:

- `src/server/responses/core.ts` read the non-2xx error body with `.text()` *before* the guard at `:3436`; the Anthropic terminal-guard continuation did the same.
- Both web-search executors read the failure body before attaching their guard.
- `/v1/live` had no guard at all, passed no signal to `readBodyCapped`, and released the reader lock without cancelling when a read threw.

Each is the same fetch-resolution-to-reader-attach window the guard was written for, just on the branch nobody guarded.

`readBodyCapped` now cancels the reader when a read throws. That division matters and is easy to get wrong: `body.cancel()` from the abort listener **throws once a reader holds the lock**, so the guard covers the window before attach and the reader covers the window after it. A test pins exactly that, because assuming either one covers both is the mistake this PR is correcting.

**An earlier version of my audit claimed no unguarded site existed.** It was wrong twice: it checked whether a guard was present in the *function* rather than on the *branch*, and it enumerated candidates from files that already imported the helper — a survivorship filter that can only ever find sites which already have one. `/v1/live` never appeared because it never imported it. The correction is recorded in `devlog/_plan/260812_five_bug_fix_campaign/004_audit_wp5_synthesis.md`.

Deliberately **not** in scope, with reasons rather than silence: request-time OAuth token refresh touches credential handling and wants its own change with security review; the CCA image fallback, MiMo JWT bootstrap, and xAI image/video clients have a narrower pre-attach interval; provider quota and model discovery are not turn-body paths. Each is named in the devlog so the list is not lost.

## Verification

On a Linux runner (Bun 1.3.14):

- `bun test tests/cancel-body-on-abort.test.ts` — 8 pass, 0 fail
- `bun test tests/cancel-body-on-abort.test.ts tests/abort-race.test.ts tests/server-live.test.ts` — 42 pass, 0 fail
- `bun test tests/web-search.test.ts` — 51 tests, 0 fail

Note on what the new tests can and cannot assert: a source whose own `pull()` rejects is errored by the stream machinery, which by spec does **not** invoke the source's `cancel()`. So the first test asserts propagation plus a settled stream, and a second test proves the locked-stream limitation directly. Two earlier drafts of these tests asserted behavior the Streams spec does not provide; they were corrected rather than worked around.

## Checklist

Refs #1419

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cancellation of interrupted response and streaming requests.
  * Prevented stalled or incomplete response bodies when connections are aborted.
  * Strengthened handling of failed live, web-search, and response requests.
  * Ensured interrupted reads clean up correctly and release resources.
  * Improved reliability when requests fail during response-body processing.

* **Tests**
  * Added coverage for interrupted reads, stream cancellation, response-size limits, and normal request completion.

### DIFF
--- a/devlog/_plan/260812_five_bug_fix_campaign/004_audit_wp5_synthesis.md
+++ b/devlog/_plan/260812_five_bug_fix_campaign/004_audit_wp5_synthesis.md
@@ -1,0 +1,79 @@
+# 004 — WP5 audit: my fetch-site audit was wrong
+
+Independent reviewer verdict on the `051` audit deliverable: **FAIL**. The claim
+it falsified was mine, and the falsification is correct.
+
+## What I claimed, and why it was wrong
+
+I enumerated the request-path fetch sites, found `cancelBodyOnAbort` at eight of
+them, and concluded there was no unguarded site. Two errors:
+
+**1. I checked whether a guard exists in the function, not whether it covers the
+branch.** At `src/server/responses/core.ts` the guard is installed at `:3436`,
+but the non-2xx branch runs at `:3387-3394` and does
+`await upstreamResponse.text()` **before** it. Same shape in
+`src/web-search/executor.ts:85-90` and `anthropic-executor.ts:169-175`: the
+failure branch reads the body, then the guard is attached for the success path.
+So the error path reopens exactly the fetch-resolution-to-reader-attach race the
+guard exists to close. Verified both by reading the code.
+
+**2. I enumerated from a grep of files that already imported the helper.** That
+is a survivorship filter: it can only ever find sites that already have a guard.
+`/v1/live` (`src/server/live.ts:554-570`) has no guard at all, passes no signal
+to `readBodyCapped`, and releases its reader lock without cancelling on read
+failure. It never appeared in my table because it never imported the helper.
+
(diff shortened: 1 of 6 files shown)


lidge-jun:
Full-suite result at this head on the Linux runner (Bun 1.3.14):

```
11334 pass
0 fail
Ran 11345 tests across 698 files. [405.95s]
EXIT=0
```

All four Linux CI shards are green, including the `test 1/4` failure that this head fixed — CI caught a real defect in an earlier revision, where the guard I added to the combo failure branches broke the "body getter touched exactly once" contract those branches deliberately keep. The combo branches are now left unguarded on purpose, because `readBoundedResponseBody` already owns settlement there, and a test records the distinction.

PR #1547, opened 2026-08-12
test(cursor): verify Grok 4.6 routing

iamsupersocks:
## Summary

- Replace the pre-release Grok 4.6 comments with launch-day provenance from the official xAI announcement and an account-verified `cursor-grok-4.6-high` run.
- Add regression coverage for Cursor live-discovery matching, 500k context metadata, the account-observed `cursor-grok-4.6-high` wire ID, the announced low/medium/high ladder, and the inherited parameterized Fast encoding.
- Keep the existing Grok 4.6 runtime behavior unchanged.

Source: https://x.ai/news/grok-4-6

## Verification

- `bun install --frozen-lockfile`
- `bun test tests/cursor-discovery.test.ts tests/cursor-effort-suffix.test.ts` — 23 passed, 0 failed
- `bun run typecheck`
- `bun run test` — 11,388 passed, 11 skipped, 0 failed
- GitHub target, hygiene, and label checks are green

## Checklist

- [x] Docs or release notes were updated when needed. (Not needed: this PR verifies already-integrated runtime behavior.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (Tests and comments only; no credentials or auth behavior changed.)

## Review readiness checklist


iamsupersocks:
@coderabbitai review

tests/cursor-effort-suffix.test.ts
```diff
@@ -12,6 +12,13 @@ const RECORDED_CURSOR_GROK_45_DISCOVERY_IDS = [
   "cursor-grok-4.5-high",
 ] as const;
 
+// Account-verified by a successful Cursor Agent run on the 2026-08-12 launch day.
```
Wibias, line 15: The fixture name/comment makes these three IDs look like a recorded account observation, but the PR description and the production comment both say only `cursor-grok-4.6-high` was account-verified; `low`/`medium` are inferred from the announced tier ladder. Because the test then loops over low/medium/high and checks each generated ID against this hand-authored fixture, it can pass even if Cursor's actual low/medium wire IDs differ. Please keep the recorded fixture to IDs that were actually observed (currently `high`), and test the announced low/medium ladder separately—or attach real GetUsableModels/run evidence for those IDs. The same distinction should be kept clear for the Fast parameterization, which the PR does not claim was directly account-verified.

Wibias (dismissed):
The runtime behavior is unchanged and CI is green, but the new provenance test currently overstates what was actually verified. Please separate observed Cursor wire IDs from inferred/announced tier behavior so the regression suite does not manufacture evidence for low/medium (or Fast transport) that the PR says was not directly account-verified.

tests/cursor-effort-suffix.test.ts
iamsupersocks: Good catch — fixed in c53d2187. The recorded Grok 4.6 fixture now contains only the account-observed `cursor-grok-4.6-high` ID. I split the remaining coverage into separate tests for the announced low/medium/high ladder and the Fast encoding inherited from Grok 4.5, with comments that explicitly say neither is separately verified wire evidence. I also tightened the production comment in `effort-map.ts`. Focused Cursor tests pass (23/23), typecheck passes, and the full suite passes (11,388 passed, 11 skipped, 0 failed).

Wibias (commented):
Re-review on exact head c53d2187: the previous provenance blocker is resolved. The recorded Grok 4.6 fixture now contains only the account-observed `cursor-grok-4.6-high`; low/medium and Fast are clearly separated as announced/inherited behavior rather than claimed wire captures. I found no new correctness issue in the four-file diff. I’m not approving this head yet because the PR is still draft and based on c125b612 while current `dev` is 75cacdf3 (38 commits ahead). None of the four PR files changed across that dev advance, so the rebase should be mechanical, but please rebase and let exact-head CI run before final approval.

lidge-jun:
Closing — dev already has Grok 4.6 with the full effort ladder including xhigh (added ahead of this PR). The provenance comment updates (preemptive → account-verified) are nice but the test assertions here pin ["low", "medium", "high"] while dev already supports ["low", "medium", "high", "xhigh"]. Thanks for the live verification work.

PR #1548, opened 2026-08-12
test(claude): pin listener policy propagation

luvs01:
## Summary

- correct the Japanese server reference so it states that candidate credentials are compared in constant time
- pin both Anthropic handlers to the listener-effective request policy in the focused loopback policy contract

## Why

The Japanese wording described a bounded time interval instead of the constant-time comparison guarantee. The existing listener-policy regression also covered final CORS handling but did not independently pin the policy argument passed into the two Anthropic handlers.

## Impact

This is a documentation and regression-coverage follow-up only. Runtime behavior is unchanged.

## Test plan

- Bun 1.3.14: `bun test --isolate tests/loopback-listener-admission.test.ts` (17 pass)
- Bun 1.4: `bun test --isolate tests/loopback-listener-admission.test.ts` (17 pass)
- `git diff --check`

## Review readiness checklist


Wibias (approved):
Reviewed. The Japanese wording now correctly states constant-time comparison, and the added assertions pin the listener-effective `policy` argument passed to both Anthropic handlers within the existing listener-policy source contract. No code findings.

Wibias:
Thanks @luvs01, merged. 🙌

This is a useful little guardrail: the new assertions make sure the listener-effective policy is actually propagated into both Anthropic handlers, not just applied later at the CORS layer. That gives us a focused regression test for the exact boundary that could otherwise drift during future server refactors. The Japanese wording fix is also worthwhile because "constant time" is a security property, not merely "within a fixed time".

Small diff, but it pins an important auth/policy invariant in a place where regressions would be easy to miss.

Issue #1217, opened 2026-08-07
[Feature]: add durable stream-stage timeline and failure attribution to request history


ShannwYang:
## Area

Streaming / request history / observability

## Current status

OpenCodex already has several pieces of correlation-safe request observability:

* every request has an OpenCodex `requestId`;
* request history is durably backed by `usage.jsonl` and exposed through the paginated `/api/request-history` API;
* `firstOutputMs` records semantic TTFT;
* failure rows can persist fields such as `errorCode`, `terminalStatus`, `closeReason`, and redacted `upstreamError`;
* durable `routeDecision` traces explain why a provider/model/account route was selected;
* the live request-log context can classify some terminals using `transportPhase` and `terminalSource`.

However, the durable request-history record still cannot reconstruct the end-to-end streaming timeline.

In particular, `transportPhase` and `terminalSource` are currently live request-log diagnostics rather than durable `usage.jsonl` fields, and they do not identify the actual failure side precisely enough.

This issue tracks the remaining durable streaming-stage observability.

## Problem

When a Codex turn remains on “thinking” or ends with a stream-disconnected error, an operator should be able to determine from one request-history record whether the delay or failure occurred:

* before OpenCodex dispatched upstream;
* while waiting for upstream response headers;
* while reading the upstream response body/SSE stream;
* while transforming or relaying upstream data;
* while writing downstream;
* because the client cancelled;
* during terminal-event delivery.

Today several of these cases can collapse into the same terminal status.

For example:

```text
terminalSource = synthetic
transportPhase = mid_stream
status = 502
```

does not prove that the provider caused the failure.

A synthetic terminal identifies who manufactured the terminal event, while a broad relay catch may cover both upstream reads and local relay/rewriting work.

## Required durable timeline

Add a bounded, payload-free stream timeline to the canonical request-history row.

At minimum record elapsed milliseconds from request acceptance for:

```text
upstreamDispatchMs
upstreamHeadersMs
upstreamFirstByteMs
upstreamFirstSemanticOutputMs
downstreamFirstWriteMs
upstreamEndMs
downstreamEndMs
```

Existing `firstOutputMs` may satisfy `upstreamFirstSemanticOutputMs` if the naming/relationship is explicit and backwards compatible.

All timings are best-effort observability only.

Failure to record a timing must never affect request delivery.

## Failure attribution

Add stable bounded enums that identify where the terminal condition originated.

For example:

```text
failureSide:
  upstream
  relay
  downstream
  client
[...]

PR #1558, opened 2026-08-12
perf: route ChatGPT Codex upstream turns over responses_websockets

Wibias:
> Maintainer-rebased replacement for #1487 from @kargnas. The contributor fork could not accept the rebased Git object through the GitHub integration after branch synchronization, so this upstream branch preserves the exact reviewed change on current `dev`.

## Problem

Requests proxied through opencodex to the ChatGPT Codex backend consistently showed **2–3s worse TTFT** than the same requests made by Codex CLI directly — even with the same account, same payload, and strictly sequential execution.

Root cause: Codex CLI talks to `chatgpt.com/backend-api/codex/responses` over the **`responses_websockets`** transport, while opencodex always POSTs SSE. The backend serves the WS path from a measurably faster queue.

## Measurements (2026-08-12, same account, same payload, sequential, alternating order)

| transport | gpt-5.6-luna TTFT p50 | mean |
|---|---|---|
| WS (`responses_websockets`) | **1037 ms** | 1212 ms |
| SSE (POST, `stream:true`) | **3897 ms** | 3686 ms |

Event timeline shows the gap is upstream scheduling, not transfer: both paths reach `response.created` in ~0.5s, but SSE then waits 2.4–4.0s before `response.output_item.added` (WS: 0.8–1.3s).

Ruled out: HTTP/2 vs 1.1 (no change), `OpenAI-Beta: responses_websockets` header on the SSE POST (no change), `session_id`/`prompt_cache_key` (no change), warmup effects (no decay over sequential repeats), account differences (A/B with identical account).

A second finding: the fast lane keys on **WS + `originator` tag**, not the transport alone — 60KB turns run ~1.5s with `originator: codex_cli_rs` vs ~4.6s without. The patch defaults the header for callers that don't send one (Codex CLI always does).

## Change

- New `src/server/responses/ws-upstream.ts`: for streaming POSTs to the Codex backend, dial `wss://` with the same headers, send the JSON body as a single `response.create` frame, and re-encode returned event frames as an SSE byte stream — so the passthrough relay, adapter parsers, and usage sniffing are all unchanged.
- `providerFetch()` in `fetch-helpers.ts` wraps the provider fetch with this transport swap. Everything that isn't a Codex-backend streaming turn keeps the exact HTTP path.
- Fallbacks: upgrade rejected (401/403/429/5xx) → retry over plain SSE so the real HTTP status reaches the existing refresh/rotation handlers; no 101 within 10s → SSE; frame-send failure → stream error into the caller's normal transport-retry path. WS-only frames (`codex.rate_limits`, `responsesapi.websocket_timing`) are dropped so clients see exactly the stream shape they always got.

## After patching (local proxy vs direct Codex CLI)

| target | Luna TTFT p50 | Terra TTFT p50 |
|---|---|---|
| codex CLI direct (WS) | 1196 ms | 1426 ms |
| opencodex **before** | 4385–4705 ms | 2951–4431 ms |
| opencodex **after** | **1424 ms** | **1577 ms** |

Also verified: tool-call round-trips (function_call arguments relay), 630KB / 900KB / 1.26MB / 1.5MB input frames, `context_length_exceeded` and `server_is_overloaded` error relay, and ~600 live requests through the patched proxy with no new failure modes.

## Known trade-off

Bun's `WebSocket` does not expose the 101 response headers, so `x-codex-*-reset-at` quota hints are not visible on this path. The periodic quota poller still covers quota tracking. If there's a preferred way to surface those, happy to adjust.

Related: #1217 (stream-stage timing) would make this kind of transport gap visible in the dashboard.

## Summary by CodeRabbit

* **New Features**
  * Added WebSocket-based streaming for eligible Codex responses.
  * Converts WebSocket messages into the existing streaming response format.
  * Preserves HTTP streaming when requests are ineligible or WebSocket connectivity is unavailable.
  * Supports request cancellation and clean stream termination.

* **Bug Fixes**
  * Improved resilience with fallback handling for invalid requests, connection failures, and upgrade timeouts.

### DIFF
--- a/src/server/responses/fetch-helpers.ts
+++ b/src/server/responses/fetch-helpers.ts
@@ -1,4 +1,5 @@
 import type { Server } from "bun";
+import { codexWsUpstreamFetch, shouldUseCodexWsUpstream } from "./ws-upstream";
 import { bridgeToResponsesSSE, buildResponseJSON, formatErrorResponse, type ResponsesTerminalStatus } from "../../bridge";
 import {
   getConfigPath,
@@ -131,7 +132,17 @@ export function safeOriginLabel(url: string): string {
 
 
 export function providerFetch(provider: OcxProviderConfig): typeof globalThis.fetch {
-  return (provider as OcxProviderConfig & { fetch?: typeof globalThis.fetch }).fetch ?? globalThis.fetch;
+  const base = (provider as OcxProviderConfig & { fetch?: typeof globalThis.fetch }).fetch ?? globalThis.fetch;
+  // ChatGPT Codex backend: streaming turns ride the responses_websockets
+  // transport (measured ~3s faster TTFT than the SSE POST queue); everything
+  // else keeps the provider's HTTP fetch. See ws-upstream.ts for the details.
+  const wrapped = (input: Parameters<typeof globalThis.fetch>[0], init?: RequestInit) => {
+    if (typeof input === "string" && init && shouldUseCodexWsUpstream(input, init)) {
+      return codexWsUpstreamFetch(input, init, base);
+    }
+    return base(input, init);
+  };
+  return wrapped as typeof globalThis.fetch;
 }
 
 
--- a/src/server/responses/ws-upstream.ts
+++ b/src/server/responses/ws-upstream.ts
@@ -1,0 +1,199 @@
+// Upstream WebSocket transport for the ChatGPT Codex backend.
+//
+// Why this exists: the Codex backend serves the responses_websockets path from
+// a measurably faster queue than the plain SSE POST path. Measured 2026-08-12
+// KST (same account, same payload, strictly sequential): gpt-5.6-luna TTFT p50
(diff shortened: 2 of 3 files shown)


Wibias:
Merged. Big thanks to @kargnas for the original investigation and implementation in #1487, including the transport measurements that found the real source of the proxy latency gap.

This is a meaningful performance win: eligible ChatGPT Codex streaming turns can now use `responses_websockets`, bringing OpenCodex TTFT much closer to direct Codex CLI instead of paying the previous multi-second SSE scheduling penalty. Just as importantly, the existing downstream SSE shape stays intact, with explicit fallback and failure handling, so we get the latency improvement without forcing adapters or clients onto a new protocol surface.

The added regression coverage around routing, upgrade timeout, send failure, and premature socket close makes the faster path much safer to keep as the default transport optimisation. Thanks again @kargnas for the groundwork that made this possible.

PR #1559, opened 2026-08-12
chore(release): require audited dispatch SHA

Wibias:
## Summary

- require `expected-sha` for every manual Release workflow dispatch
- fail closed if the supplied SHA is missing, malformed, or no longer matches the selected release ref
- validate manual release dispatches against trusted guard code loaded from the repository default branch
- allow releases only from `main` and `preview`
- run dispatch validation in a separate read-only job before the privileged publish job can start
- scope GitHub write and OIDC permissions to the publish job instead of granting them workflow-wide
- add focused tests for the release dispatch guard

## Why

The Release workflow is designed to publish an explicitly reviewed and audited commit, but the previous workflow still had two trust gaps.

First, `expected-sha` was optional. A manually dispatched release without it could continue by publishing whatever commit the selected branch currently pointed at.

Second, branch/ref validation happened only after the selected dispatch ref had already been checked out and repository code had begun executing. Since `workflow_dispatch` can be started against a selected branch, that meant release-ref code could run before the workflow had established that the dispatch came from an approved release branch.

This change makes the release path fail closed.

A new `validate-dispatch` job:

- has only `contents: read`
- checks out the dispatch guard explicitly from the repository default branch
- requires the event to be `workflow_dispatch`
- permits only `refs/heads/main` and `refs/heads/preview`
- requires a full 40-character `expected-sha`
- requires `GITHUB_SHA` to exactly match that audited SHA

Only after that validation succeeds can the `publish` job start.

The write-capable permissions required for GitHub release creation and npm Trusted Publishing are now scoped specifically to `publish`, rather than being granted to every job in the workflow.

## Impact

Normal releases through `bun scripts/release.ts` remain compatible because the helper already:

- runs only from `main` or `preview`
- waits for CI before dispatch
- verifies the live remote branch head before dispatch
- supplies `expected-sha=${releaseSha}`

Intentional behavior changes:

- manual releases without `expected-sha` are rejected
- short or malformed SHAs are rejected
- dispatches from branches other than `main` or `preview` are rejected
- the privileged publish job cannot begin unless trusted dispatch validation succeeds

No npm publishing semantics, release channel mapping, or `main`/`preview` release support are removed.

## Validation

Added focused tests covering:

- valid `main` dispatch
- valid `preview` dispatch
- non-`workflow_dispatch` events
- unapproved refs such as `dev`
- missing `expected-sha`
- non-full commit SHAs
- branch movement after audit

CodeRabbit's two security findings covering dispatch trust ordering and trusted guard invocation are now resolved.

## Review

Ready for review.

@lidge-jun @Ingwannu

### DIFF
--- a/.github/scripts/release-dispatch-guard.cjs
+++ b/.github/scripts/release-dispatch-guard.cjs
@@ -1,0 +1,44 @@
+"use strict";
+
+const ALLOWED_RELEASE_REFS = new Set([
+  "refs/heads/main",
+  "refs/heads/preview",
+]);
+
+function validateReleaseDispatch({
+  eventName,
+  ref,
+  expectedSha,
+  actualSha,
+}) {
+  if (eventName !== "workflow_dispatch") {
+    return `Release must be triggered by workflow_dispatch; got ${eventName || "(empty)"}.`;
+  }
+
+  if (!ALLOWED_RELEASE_REFS.has(ref)) {
+    return `Release must run from main or preview; got ${ref || "(empty)"}.`;
+  }
+
+  if (!expectedSha) {
+    return "expected-sha is required; refusing to publish without an audited commit.";
+  }
+
+  if (!/^[0-9a-f]{40}$/.test(expectedSha)) {
+    return `expected-sha must be a full 40-character commit SHA; got ${expectedSha}.`;
+  }
+
+  if (actualSha !== expectedSha) {
+    return (
+      `branch moved after the release audit ` +
+      `(expected ${expectedSha}, got ${actualSha || "(empty)"}) — ` +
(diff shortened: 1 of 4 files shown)


Ingwannu (approved):
Security review completed on exact head `d373694b1cd8c1660ca4793d444035f457222243`. The release gate is loaded from the trusted default branch before release-ref code runs, dispatch is limited to `main`/`preview`, `expected-sha` is required as a full SHA and compared to `GITHUB_SHA`, workflow-level permissions are empty, and write/OIDC permissions exist only on the gated publish job. The actual PR diff is four files and exact-head CI is green. Approved from the release-integrity boundary.

Issue #357, opened 2026-07-23
[Feature Request] Complete the external aggregated model API for third-party agents


str0203:
## Summary

OpenCodex already aggregates multiple upstream sources, including ChatGPT/Codex account pools, OAuth/API-key providers, OpenAI-compatible relay services, Anthropic, Gemini, OpenRouter, Ollama, and combo/failover routes.

It would be valuable to expose these resources as a complete, general-purpose model gateway for third-party AI agents—not only for Codex and Claude Code.

The dashboard currently provides an **API Access** page where users can generate an `ocx_...` API key. However, the external API experience is incomplete, especially for clients that need to discover and select models.

## Intended use case

A user may have multiple ChatGPT accounts in the OpenCodex account pool, plus several relay services and providers configured in OpenCodex.

The expected architecture is:

```text
Third-party agent
        |
        | Unified OpenAI-compatible API
        v
     OpenCodex
        |
        +-- ChatGPT/Codex account pool
        +-- OpenRouter
        +-- Third-party relay services
        +-- Anthropic / Gemini / xAI
        +-- Local models
```

The third-party agent should only need one OpenCodex endpoint and one OpenCodex API key. It should be able to retrieve the aggregated model list and select models such as:

```text
gpt-5.x
anthropic/claude-...
google/gemini-...
openrouter/...
combo/...
```

OpenCodex would then select the correct provider, credential, or ChatGPT pool account internally.

## Current behavior and limitations

OpenCodex already exposes:

- `GET /v1/models`
- `POST /v1/responses`
- `POST /v1/messages`
- Image, search, and compaction endpoints

It also routes `provider/model` IDs to configured providers. However, several limitations prevent it from working as a general aggregated gateway.

### 1. No inbound Chat Completions endpoint

Many third-party agents only support:

```http
POST /v1/chat/completions
```

OpenCodex supports Chat Completions providers internally, but it does not expose `/v1/chat/completions` to external clients. Requests to that endpoint return 404.

The current wording can be confusing: support for OpenAI-compatible Chat Completions providers describes the **OpenCodex-to-upstream** direction, not necessarily the **client-to-OpenCodex** direction.

### 2. The API Access page does not expose the model catalog

The page allows users to generate API keys, but it does not clearly show:

- All externally callable models
- Exact model IDs to send
- Which provider owns each model
- Which inbound protocol supports each model
- Whether a model uses the ChatGPT account pool
- Whether a model is a combo/failover route

[...]

PR #1569, opened 2026-08-13
feat(chat): native chat->chat for openai-chat providers (#1467)

### LINKED ISSUE #1467: Preserve Chat Completions for chat-native open-source agents
### Area

Proxy and routing

OpenAI Chat Completions-native open-source agents should be able to use OpenCodex routing without having their requests and responses translated through the Responses protocol when the selected upstream provider also speaks `openai-chat`.

This applies to generic third-party agents configured with `POST /v1/chat/completions`, not to Codex clients using Responses or Claude Code clients using Messages.

### What prevents this today?

The inbound Chat Completions handler always converts the request into the internal Responses shape, runs the Responses pipeline, converts an `openai-chat` provider response into Responses events, and then converts those events back into Chat Completions.

For a Chat-native client routed to a Chat-native provider, the effective path is:

```text
Chat client -> Chat request -> Responses AST -> Chat upstream
            <- Chat response <- Responses events <-
```

The double conversion adds work and can alter protocol details that the original client and provider both understand natively, including assistant reasoning fields, tool-call history, message roles, tool schemas, IDs, and streaming padding frames. A direct Chat client-to-provider configuration avoids those transformations.

### What should OpenCodex do?

When an inbound `/v1/chat/completions` request resolves to an `openai-chat` target, OpenCodex should keep a Chat-native execution and output path instead of converting through Responses.

The path must still preserve OpenCodex behavior that is independent of wire format: admission authentication, model and combo routing, provider credential selection, API-key pools, pre-stream retry and failover, cancellation, usage accounting, request logging, and safe provider-specific policy.

If the selected target uses a different protocol, or if a feature requires the existing Responses bridge, OpenCodex should retain the current translation path. Existing Codex Responses and Claude Messages behavior must remain unchanged.

```text
OpenAI-compatible open-source agent
  POST /v1/chat/completions
  model: synthetic/hf:moonshotai/Kimi-K3
              |
              v
OpenCodex admission + routing + key pool + logging
              |
              v
Synthetic POST /openai/v1/chat/completions
```

Expected protocol matrix:

```text
Chat inbound + openai-chat target  -> Chat-native path
Chat inbound + non-Chat target     -> Existing translation bridge
Responses inbound + Chat target    -> Existing Responses-to-Chat adapter
Messages inbound                   -> Existing Claude-compatible path
```

- Configure the third-party agent directly with each provider endpoint and API key. This avoids conversion but loses OpenCodex model aggregation, credential isolation, key pools, combos, and logging.
- Continue translating every Chat request through Responses and add provider-specific compatibility patches. This addresses individual regressions but retains the structural source of lossy round trips.
- Relay raw upstream bytes immediately. This is too narrow unless Chat-specific inspection preserves usage, terminal state, cancellation, retry/failover, tool aliases, and request logs.

### Additional context

This is a focused architectural follow-up to the external third-party agent gateway implemented after #357. The current documentation describes Chat Completions as translating into the internal Responses pipeline so all providers and sidecars share one lifecycle.

A safe implementation should reuse that lifecycle rather than duplicate it. One possible boundary is a Chat output mode in the shared core after route/auth/recovery decisions, with the existing bridge retained for protocol-mismatched targets and features that require it.

### Checks

dbc-hbin:
## Summary

Chat-form providers are the majority, so this keeps the change minimal: when a `POST /v1/chat/completions` request routes to an `openai-chat` target, preserve the Chat wire end-to-end instead of round-tripping through the Responses pipeline (`chat -> responses -> chat`).

- Only `chat -> chat` is native in this PR. Every other target (Responses-native, Anthropic, Cursor/Kiro, hosted-tool/previous_response paths) stays on the existing translate-and-replay bridge.
- The shared lifecycle (admission, routing, key-pool failover, pre-stream retry, cancellation, usage, request logging) is **not** duplicated — chat-native reuses the same helpers and request-log path that the bridge uses, split only at the wire.

Closes #1467 (follow-up to #357).

## Why this scope

The current `/v1/chat/completions` handler always converts the request into the internal Responses shape and then converts it back to Chat after the `openai-chat` adapter renders it again. For a Chat client routed to a Chat provider that is a wasted round-trip and can alter protocol-faithful fields (tool-call history, `reasoning_content`, `response_format`, streaming frames). Keeping it minimal on the dominant wire avoids touching the rest of the pipelines.

## Behavior matrix

| Inbound | Resolved target | Path |
|---------|-----------------|------|
| `chat` | `openai-chat` | **Chat-native** (new) |
| `chat` | non-chat adapter | Existing Responses bridge |
| `responses` | any | Unchanged |
| `messages` (Claude) | any | Unchanged |

Chat-native is taken only when the resolved provider is `openai-chat` (authMode != forward) and the payload needs no Responses-only feature (`store`/`background`/`previous_response_id`/`compaction_trigger`/hosted `web_search`/`image_generation`). Otherwise it falls through to the bridge.

## Change

Single-file, minimal: `src/server/chat-completions.ts`

- Resolve the wire after routing (`resolveWireProtocolOverride(..., "chat")`) and branch on the result.
- Chat-native: build `POST /chat/completions` directly, send via `fetchWithHeaderTimeout` + `fetchWithTransientRetry`, with pre-stream 429 same-target wait and key-pool `rotateProviderTransportOn429`, bearer from `provider.apiKey`/`provider.headers`, and `modelSuffixBracketStrip` handling.
- Stream: normalize upstream Chat SSE into `chat.completion.chunk` envelope (mock omits it) and relay via `trackStreamLifetime` so drain/cancel stay correct.
- Non-stream: upstream may return SSE or JSON — fold SSE into `chat.completion` JSON for non-stream callers (matching the legacy bridge's `collectChatCompletion` folding).
- Respect `noStructuredOutputModels` per-model opt-out on the chat-native path.
- Charge the translator turn budget on the raw Chat body so the 32 MiB `translation_buffer_limit` (413) still holds without a Responses stringify.
- Usage/logging: `inboundProtocol:"chat"`, `beginRequestAttempt("openai-chat")`, `usage.prompt_tokens`/`completion_tokens` parsed into `OcxUsage`.

No raw byte relay; no duplicated lifecycle/drain/logging.

## Verification

```bash
bun x tsc --noEmit
bun run test tests/chat-completions-endpoint.test.ts  # 52 pass
bun run test tests/chat-completions-endpoint.test.ts tests/adapter-resolve.test.ts tests/routing-compatibility.test.ts  # 97 pass
bun run test tests/openai-chat-hardening.test.ts tests/server-key-failover-e2e.test.ts tests/openai-responses-passthrough.test.ts  # 116 pass
```

Notable coverage that passed: streaming `chat.completion.chunk` envelope, non-streaming `chat.completion` JSON, translator overflow `413 translation_buffer_limit`, and per-model `response_format` opt-out.

## Checklist

- [x] `bun x tsc --noEmit` green
- [x] `chat-completions` suite green (52)
- [x] No new provider/registry entries; no API surface change beyond `chat -> chat` pass-through

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added direct Chat Completions support for compatible providers.
  * Improved structured-output, streaming, and usage reporting.
  * Added automatic model and response-format compatibility handling.
  * Preserved streamed tool calls in non-streaming responses.

* **Bug Fixes**
  * Improved cancellation, error handling, and oversized-response recovery.
  * Added retries and credential failover for transient failures and rate limits.
  * Improved streaming and non-streaming response conversion.
  * Ensured compatible providers receive appropriate authentication credentials.
  * Preserved invalid upstream responses for clearer troubleshooting.

### DIFF
--- a/src/server/chat-completions.ts
+++ b/src/server/chat-completions.ts
@@ -346,11 +346,9 @@ async function handleChatCompletionsWithBudget(
         : ac.signal;
       let detachReq: (() => void) | undefined;
       if (typeof AbortSignal.any !== "function" && !combinedSignal.aborted) {
-        const acAbort = () => {};
         const onReqAbort = () => {
           try { (ac as unknown as { abort: (r?: unknown) => void }).abort(req.signal.reason); } catch { /* noop */ }
         };
-        void acAbort;
         req.signal.addEventListener("abort", onReqAbort, { once: true });
         detachReq = () => req.signal.removeEventListener("abort", onReqAbort);
       }
@@ -677,8 +675,9 @@ async function handleChatCompletionsWithBudget(
       }
     }
     let jsonText: string;
+    let jsonReader: ReadableStreamDefaultReader<Uint8Array> | null = null;
     try {
-      const reader = response.body ? response.body.getReader() : null;
+      const reader = response.body ? (jsonReader = response.body.getReader()) : null;
       if (!reader) {
         jsonText = await response.text();
         translatorBudget.chargeRetained(new TextEncoder().encode(jsonText).byteLength, { kind: "request_copies" });
@@ -700,6 +699,9 @@ async function handleChatCompletionsWithBudget(
         jsonText = new TextDecoder().decode(merged);
       }
     } catch (err) {
+      if (jsonReader) {
+        try { void jsonReader.cancel().catch(() => {}); } catch { /* noop */ }
+      }
       if (isTranslatorBudgetExceededError(err)) {
         cleanup();
         finishRequestAttempt(attempt, 413, Date.now() - (logCtx.activeAttemptStartedAt ?? Date.now()));
--- a/src/server/chat-completions.ts
+++ b/src/server/chat-completions.ts
@@ -2,10 +2,13 @@
  * OpenAI Chat Completions inbound (/v1/chat/completions) for GitHub Copilot App
  * and other OpenAI-compatible clients.
  *
- * Translate-and-replay: Chat Completions body -> /v1/responses via handleResponses,
- * then bridge the Responses output back to Chat Completions SSE/JSON.
+ * Dual-path: `chat -> chat` for the majority chat-form providers is a
+ * direct Chat wire (no Responses round-trip). Every other target keeps
+ * the existing translate-and-replay bridge so the proxy's admission,
+ * routing, key-pool, retry, usage, and logging stay on one shared path.
  */
 import { FORWARD_HEADERS } from "../adapters/openai-responses";
+import { stripBracketedModelSuffix } from "../adapters/openai-chat";
 import { ChatCompletionsRequestError, chatCompletionsToResponsesBody } from "../chat/inbound";
 import {
   chatCompletionsErrorResponse,
@@ -21,17 +24,32 @@ import { estimateTokens } from "../lib/token-estimate";
 import { NoEligiblePolicyCandidateError, routeModel } from "../router";
 import { evidenceFromBody } from "../routing/request-evidence";
 import { resolveWireProtocolOverride } from "./adapter-resolve";
-import type { OcxConfig } from "../types";
+import type { OcxConfig, OcxProviderConfig } from "../types";
+import type { UpstreamSendRecovery } from "../lib/upstream-retry";
 import { readJsonRequestBody } from "./request-decompress";
 import {
   addFinalRequestLog,
+  beginRequestAttempt,
+  finishRequestAttempt,
+  noteAttemptSend,
+  sealRequestAttemptIdentity,
   httpStatusForRequestLogTerminal,
   recordFirstOutput,
   type RequestLogContext,
   type RequestLogEntry,
 } from "./request-log";
 import { responseWithDeferredRequestLog } from "./relay";
-import { handleResponses } from "./responses";
+import { handleResponses, linkAbortSignal } from "./responses";
+import { fetchWithHeaderTimeout, providerFetch, safeHostLabel } from "./responses/fetch-helpers";
+import { fetchWithTransientRetry } from "../lib/upstream-retry";
+import { cancelBodyOnAbort } from "../lib/abort";
+import { trackStreamLifetime } from "./lifecycle";
+import {
+  hasKeyPoolFailover,
+  rateLimitRetryDelayMs,
+  rateLimitRetryPolicyFor,
+  rotateProviderTransportOn429,
+} from "../providers/key-failover";
 import type { AdmissionLease } from "../lib/admission";
 import { tryClaimNativeMainProfileForTurn } from "../codex/native-main-admission";
 import {
@@ -73,6 +91,26 @@ export async function handleChatCompletions(
   }
 }
 
+function isChatNativeEligibleProvider(provider: OcxProviderConfig): boolean {
+  return provider.adapter === "openai-chat" && provider.authMode !== "forward";
+}
+
+function shouldBridgeChatNative(raw: Rec): boolean {
+  if (raw.store === true) return true;
+  if (raw.background === true) return true;
+  if (typeof raw.previous_response_id === "string" && raw.previous_response_id.length > 0) return true;
+  if (raw.compaction_trigger !== undefined) return true;
+  if (Array.isArray(raw.tools)) {
+    for (const t of raw.tools as unknown[]) {
+      if (!t || typeof t !== "object") continue;
+      const tt = (t as Rec).type;
+      if (tt === "web_search" || tt === "web_search_preview" || tt === "image_generation") return true;
+    }
+  }
+  return false;
+}
+
+
 async function handleChatCompletionsWithBudget(
   req: Request,
   config: OcxConfig,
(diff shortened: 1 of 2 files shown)


Ingwannu (changes requested):
The direction is valuable, but the current exact head `c89a2fdf5352f4aa5fe4742dc744cb1376c2776b` still has one concrete resource-lifecycle blocker in `src/server/chat-completions.ts:679-713`: after `response.body.getReader()`, a translator-budget exception or `reader.read()` rejection reaches the catch path without cancelling the locked reader/body. Repeated oversized or failed JSON responses can therefore retain upstream sockets until GC. Keep the reader outside the try, `await reader.cancel()` best-effort before cleanup on every catch return, and add a regression proving cancellation on budget overflow/read failure. Also rerun full exact-head CI; the current head only has lightweight hygiene/target checks, not the cross-platform suite. Once that is fixed and exact-head CI is green, this remains a strong merge candidate for #1467.

Ingwannu (commented):
Re-reviewed exact head `eb925fa04074275dde2b64041f43a2524082eafa`: the reader is now retained and cancelled best-effort on the bounded JSON catch path, so the resource-lifecycle blocker is resolved. I am not approving yet because this new head still lacks a focused regression that proves cancellation on translator-budget overflow/read failure, and the exact-head cross-platform CI suite has not run. Add that regression and obtain green exact-head CI; no further product-code blocker is identified in this update.

Ingwannu:
I rechecked the new head `e6dbd8dc769f02117df1fd94cb31bdaf1d7ec1ca`. The reader-cancellation change is present, but this remains blocked: the exact head is two commits behind current `dev`, full exact-head CI has not run, and the latest incremental review still has two actionable test defects (the second response body is not consumed before the cleanup assertion, and the mock base-URL slash normalization regex does not match an actual trailing slash). Please fix those, rebase onto current `dev`, and obtain fully green exact-head CI before requesting another merge review. The existing `CHANGES_REQUESTED` disposition remains correct.

Ingwannu (changes requested):
Re-reviewed exact head `b10379ecf87b5d0febed933e8a0b89fa4919fa18`. Some prior blockers were addressed, but this update is still not mergeable and introduced new failure semantics. Local focused validation remains green (39/39, typecheck, privacy scan), but there is no regression coverage for the new paths below.

1. In `src/server/chat-completions.ts`, a same-target replay fetch exception now records/finalizes 502 and then `break`s instead of returning. `response` remains the previous 429 whose body was already cancelled; the function can then enter key rotation or the non-OK response path, finalize again, and return/log 429 rather than the transport 502. Return the 502 immediately (as the previous head did), with a regression proving one final log and no rotation/error-body read after replay transport failure.
2. The non-OK “8 KiB” reader is still not byte-bounded. It decodes each entire provider-controlled chunk before checking `chunk.length`, so one large first chunk allocates/decodes the whole body. Count and slice `Uint8Array` bytes before decoding, await best-effort cancellation when the byte cap is reached or a read fails, and test a single oversized chunk plus read rejection.
3. The JSON-despite-stream synthesis treats invalid JSON or a JSON object without a valid Chat completion choice as a successful empty `stop` completion followed by `[DONE]`. That hides upstream protocol failures as HTTP/SSE success. Parse and validate through a shared bounded Chat-completion normalizer; malformed/failed JSON must emit the established error contract, not fabricate a successful turn. Add text, tool-call, malformed, and failed-response regressions.
4. Stream usage accounting is still incomplete for the common canonical `object: "chat.completion.chunk"` branch (including its EOF-tail branch): those chunks are forwarded without copying `usage` into `logCtx.usage` / attempt usage. Centralize usage extraction so every forwarded/synthesized chunk uses it, and assert the finalized request log.

The new head also moved the PR out of draft while exact-head cross-platform CI has not run. Please return it to draft until these protocol/logging/resource boundaries have direct tests and the full exact-head suite is green. The overall native Chat direction remains worth pursuing; the request is to narrow and prove this implementation before merge.

Ingwannu (commented):
Exact-head re-review at `fd7333540b8b9ccb936d36fd4638e425e28f2bfa`: the four blockers from my prior review are addressed. Replay transport failures now return immediately with one 502 path; the non-OK reader bounds bytes before decoding and settles the reader; JSON-despite-stream validates Chat completion shape instead of fabricating success; and usage extraction covers canonical and synthesized chunks. The new focused regressions cover valid JSON/SSE synthesis, tool calls, malformed/missing-choice responses, overflow, and replay failure. Local exact-head validation is green: 60/60 endpoint tests, typecheck, privacy scan, and diff check.

I found no remaining blocker in the reviewed native Chat paths. Keep the PR draft until the full exact-head cross-platform suite is run and green; the current head only has lightweight PR gates. Once that matrix is green and an independent maintainer is comfortable with the new direct-wire surface, this remains a strong merge candidate for #1467.

Wibias (changes requested):
Re-reviewed exact head `fd7333540b8b9ccb936d36fd4638e425e28f2bfa`. Ingwannu's previous four blockers are addressed on this head, and exact-head Cross-platform CI + React Doctor are now green. I still found three current-head blockers that need to be resolved before approval:

1. **Native SSE framing is still unsafe/incomplete.** The new stream normalizer buffers decoded text and splits only on `\n\n`. A standards-compliant CRLF stream (`\r\n\r\n`) can accumulate multiple frames into one buffer and fail parsing at EOF, dropping completion/usage data. The buffer is also unbounded, so an oversized or unterminated upstream SSE event can grow a JS string without the repository's frame cap. Please reuse the existing bounded SSE framing implementation and add regressions for fragmented CRLF framing and an oversized/unterminated event.

2. **The native path bypasses the `openai-chat` provider request contract.** It forwards the raw Chat body with only a small set of adjustments, while `createOpenAIChatAdapter().buildRequest()` applies provider/model-specific normalization for message/tool shape, token limits, reasoning wire formats, temperature/top-p/penalty exclusions, structured output, parallel tool calls, routing extensions, prompt cache fields, etc. That means a provider that worked through the existing adapter can receive fields that its config explicitly says to strip/rewrite and fail on the native path. Keep the Chat wire native, but extract/reuse a shared Chat request builder rather than implementing a second partial contract here.

3. **Parsed upstream error messages are not redacted before being returned to the client.** The raw fallback is passed through `redactSecretString`, but a successful JSON parse assigns `nested?.message` / flat message directly, then serializes the classified message into the client-facing error. A provider that echoes a credential/token in a structured error can therefore expose it. Redact the parsed provider message before classification/output and add a regression for a structured error that contains a configured secret.

There is also a readiness/freshness issue: current `dev` is `81ada7cd092d4be3b25f3013c996cd3262a2f99b`, while this PR still has merge base `af089996616b33163248ffad974d49d957a06e0b`; the compare shows `dev` 243 commits ahead. Current `dev` has materially changed this area, including the shared `openaiChatCompletionsUrl()` helper, so this needs rebasing/reconciliation before the PR can leave draft.

Once these three implementation issues are fixed and the branch is brought back onto current `dev`, I am happy to re-review.

lidge-jun:
Triage update (2026-08-15, maintainer): rebuilt onto current dev with a raw-Chat-preserving passthrough builder (no Responses round trip), bounded SSE, and redacted structured errors. Passed adversarial review and the full remote suite (12330 pass / 0 fail). Landing held pending a maintainer release-timing decision; ready on approval.

lidge-jun:
Landed via #1752 (dev merge 9ed2e8459). The endpoint tests were cherry-picked with authorship preserved; the transport was rebuilt for true native Chat fidelity — a sentinel-field probe verifies service_tier, seed, logprobs, logit_bias, user, metadata, message names, and system/developer separation survive end-to-end, with bounded SSE and redacted structured errors. Policy/combo/OAuth paths deliberately keep the Responses bridge. Full gates green. Closing as landed — thank you!

Issue #796, opened 2026-07-31
[Bug] Volcengine Ark tool-call assistant content needs live endpoint validation


hooliy-01:
## Area

Provider adapter / Volcengine Ark compatibility

## Current status

The original report reproduced a Volcengine Ark `400 MissingParameter` error when conversation history contained an assistant message with `tool_calls` and empty text content.

The original failing error was:

```text
Provider error 400:
{"error":{"code":"MissingParameter","message":"The request failed because it is missing `input.content.text` parameter.","param":"input.content.text","type":"BadRequest"}}
```

A provider-specific compatibility change has since landed on `dev`.

Current OpenCodex detects Ark hostnames and uses an Ark-specific structured placeholder for assistant history entries that contain tool calls or reasoning but no textual content.

The relevant implementation is in:

```text
src/adapters/openai-chat.ts
```

and originated in commit:

```text
2eebd9268
```

The corresponding regression coverage is in:

```text
tests/volcengine-ark-assistant-content.test.ts
```

However, the emitted content shape was inferred from Ark's error contract and has **not yet been confirmed against a live Volcengine Ark endpoint**.

This issue therefore remains open for live compatibility validation.

## Original reproduction

The original report used OpenCodex:

```text
2.7.43-preview.20260728
```

with a conversation containing an assistant tool-call message with no textual content.

Conceptually:

```json
{
  "role": "assistant",
  "content": "",
  "tool_calls": [
    {
      "id": "call_1",
      "type": "function",
      "function": {
        "name": "shell",
        "arguments": "{\"cmd\":\"ls\"}"
      }
    }
  ]
}
```

The Ark endpoint returned:

```text
400 MissingParameter: input.content.text
```

Single-turn requests without tool-call history worked.

## What current OpenCodex does

The previous generic serialization used an empty string for assistant messages containing tool calls but no text.

Current `dev` treats Ark separately.

For recognized Ark hosts, OpenCodex emits a structured empty-text placeholder instead of the generic empty-string representation.

Conceptually:

```json
{
  "role": "assistant",
  "content": [
    {
      "type": "text",
      "text": ""
    }
  ],
  "tool_calls": [...]
}
```

This behavior is intentionally provider-specific.

Other providers retain their existing serialization because their validation requirements differ.

## Why the original suggested fix is no longer current

The original report suggested replacing the empty string with a single space:

```text
""
```

→

```text
" "
```

That should no longer be treated as the proposed fix.

[...]

PR #1574, opened 2026-08-13
fix(volcengine): scope assistant placeholder by endpoint

### LINKED ISSUE #1571: [Bug] 400 "A parameter specified is not valid" with volcengine kimi-k2.7-code on tool-call continuation (regression from 2.7.42)
### Client or integration
Codex App

### Area
Provider adapter

### Summary
Requests carrying assistant `tool_calls` + tool messages routed to Volcengine kimi-k2.7-code are rejected by the upstream with `400: A parameter specified is not valid`. Plain conversations work; only tool-call continuation turns fail.

### Reproduction
POST `https://ark.cn-beijing.volces.com/api/coding/v3/chat/completions`, model `kimi-k2.7-code`, with these assistant messages followed by a matching tool message:

| assistant message | result |
|---|---|
| `content=null` + tool_calls | 200 |
| `content=""` + tool_calls | 200 |
| `content=[{"type":"text","text":""}]` + tool_calls | **400 "A parameter specified is not valid"** |

Root cause: `emptyAssistantContent()` in `src/adapters/openai-chat.ts` (introduced in 2.12.0) returns an empty text array `[{type:"text",text:""}]` for Volcengine Ark providers when an assistant message has tool_calls but no text. Volcengine kimi-k2.7-code rejects that shape. 2.7.42 used an empty string `""` on the same path, which every Ark model accepts. Verified that deepseek-v4-flash, deepseek-v4-pro, doubao-seed-2.1-turbo, glm-5.2, minimax-m3 and kimi-k2.6 all accept `content=""`.

### Version
2.12.0 (confirmed still present in 2.13.0)

### Operating system
Windows 10 Home China 10.0.19045

### Provider and model
volcengine / kimi-k2.7-code

### Logs or error output
Provider error 400: A parameter specified in the request is not valid

### Redacted configuration
adapter: openai-chat
baseUrl: https://ark.cn-beijing.volces.com/api/coding/v3
model: kimi-k2.7-code

### Checks

Ingwannu:
## Summary

Fixes #1571 by separating the two Volcengine Ark Chat endpoint contracts.

The #796 workaround previously matched only the Ark hostname and therefore sent its inferred structured empty assistant placeholder to both:

- pay-as-you-go Chat: `/api/v3`
- Coding Plan Chat: `/api/coding/v3`

Live evidence in #1571 shows that Coding Plan accepts `content: ""` and rejects `[{"type":"text","text":""}]` for a tool-call continuation. This patch keeps the structured placeholder only on recognized Ark hosts at the exact normalized `/api/v3` base path. Coding Plan and unrelated paths retain the ordinary empty string.

## Why this shape

Model IDs cannot identify the wire contract because the same model family can appear behind different Ark products. Removing the workaround globally would reopen the still-distinct #796 contract. The base endpoint family is the narrow boundary supported by the available evidence.

The architecture note records this decision and its tradeoffs.

## Validation

Exact head `52b1bef7e05aef23711e82cbbc672702943d3c2a`:

- `bun test tests/volcengine-ark-assistant-content.test.ts` — 8 pass, 0 fail
- `bun run typecheck` — pass
- `git diff --check` — pass

A CPU-limited full-suite run reached this change's focused coverage but later encountered unrelated environment/timeout failures: the ambient `OPENCODEX_API_AUTH_TOKEN` overrode an isolated shim fixture, and several existing long-running history/credential-sidecar tests timed out under the two-core limit. The isolated shim test passes when that ambient variable is removed. Exact-head repository CI remains required before merge.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Volcengine Ark endpoint detection for assistant continuation handling.
  - Standard Ark endpoints now receive the correct structured placeholder, while Coding Plan endpoints retain empty assistant content.
  - Prevented Ark-specific formatting from being applied to unrelated hosts or paths.

- **Documentation**
  - Documented the endpoint-specific assistant content behavior.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -672,16 +672,18 @@ const VOLCENGINE_ARK_HOSTNAMES = new Set([
   "ark.ap-southeast.volces.com",
 ]);
 
-function isVolcengineArkTarget(provider: OcxProviderConfig): boolean {
+function isVolcengineArkPaygChatTarget(provider: OcxProviderConfig): boolean {
   try {
-    return VOLCENGINE_ARK_HOSTNAMES.has(new URL(provider.baseUrl).hostname);
+    const url = new URL(provider.baseUrl);
+    const pathname = url.pathname.replace(/\/+$/, "") || "/";
+    return VOLCENGINE_ARK_HOSTNAMES.has(url.hostname) && pathname === "/api/v3";
   } catch {
     return false;
   }
 }
 
 function emptyAssistantContent(provider: OcxProviderConfig): string | { type: "text"; text: string }[] {
-  return isVolcengineArkTarget(provider) ? [{ type: "text", text: "" }] : "";
+  return isVolcengineArkPaygChatTarget(provider) ? [{ type: "text", text: "" }] : "";
 }
 
 function ensureRootObjectType(parameters: unknown): Record<string, unknown> {
--- a/structure/04_transports-and-sidecars.md
+++ b/structure/04_transports-and-sidecars.md
@@ -587,6 +587,22 @@ adapters advertise the catalog bit only on explicit `true`; cursor keeps its own
 Providers with flaky parallel streaming can be opted out individually. Evidence and provider
 ledger: `devlog/_fin/260709_parallel_tool_calls/`.
 
+## Volcengine Ark assistant continuation shapes
+
+The `openai-chat` adapter keeps Volcengine's pay-as-you-go Chat endpoint and Coding Plan endpoint
+on separate empty-assistant contracts. The pay-as-you-go `/api/v3` route retains the structured
+`[{ "type": "text", "text": "" }]` placeholder inferred for #796, while `/api/coding/v3` uses the
+ordinary empty string accepted by its live tool-call continuation contract (#1571). Matching only
+the shared Ark hostname is too broad because the two endpoint families reject opposite shapes.
+
+[Decision Log]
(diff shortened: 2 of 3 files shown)


Ingwannu:
Exact-head CI is fully green at `52b1bef7e05aef23711e82cbbc672702943d3c2a`, including all Linux shards and macOS. I have marked the PR ready for independent maintainer review; as the author, I will not approve or merge it myself.

Wibias (changes requested):
The endpoint split is the right direction, but this needs one compatibility fix before merge.

Latest `dev` now accepts `openai-chat` `baseUrl` values that already include a terminal `/chat/completions` and normalizes that suffix away before sending. This PR classifies Ark by inspecting the raw configured pathname and requiring it to equal exactly `/api/v3`.

That means a now-valid pay-as-you-go configuration such as:

`https://ark.cn-beijing.volces.com/api/v3/chat/completions`

is misclassified as non-PAYG and falls back to `content: ""`, even though it is the same PAYG endpoint as `/api/v3` after the adapter's URL normalization.

Please normalize a terminal `/chat/completions` before deciding the Ark endpoint family, and add regression coverage for both:

- `/api/v3/chat/completions` -> same structured placeholder behavior as `/api/v3`
- `/api/coding/v3/chat/completions` -> same empty-string behavior as `/api/coding/v3`

The branch also predates the current `dev`, so please rebase after that fix and rerun the relevant CI.

PR #1575, opened 2026-08-13
test(native-profile): bound hosted macOS busy-probe case

### LINKED ISSUE #1563: [Bug] CI: native-profile transaction test flakes on the hosted macOS runner at ~6000ms
### Bug report

**What happened**

`native main profile transactions > normal switch rejects a busy native Codex process before publishing any transaction file` fails intermittently on the macOS CI shard only. It failed the v2.14.0 release CI on `main` (`bf6186733`, run `31640086366`) as the single failure in an otherwise green suite — 11452 pass, 1 fail — and passed on a plain rerun of the same commit with no code change.

**Where it does not fail**

| Environment | Result |
|---|---|
| Linux CI host, `bun test --isolate`, 3 consecutive runs | 49 pass / 0 fail each |
| local macOS (Apple silicon), `bun test --isolate`, 3 consecutive runs | 49 pass / 0 fail each |
| GitHub macOS runner | intermittent |
| GitHub macOS runner, rerun of the identical commit | pass |

So it is not a platform-specific defect — local macOS is clean. It reproduces only on the hosted macOS runner, which is the slower and noisier of the two.

**Why it looks like a time-budget problem rather than a logic problem**

The failing assertion reports `[6016.60ms]`. That is a hair over a 6-second boundary, and the file's timeout budget has been tuned before — `49887650e test(native-profile): align transaction timeout budget (#1507)`. A test that fails at 6016ms against a 6000ms budget on the slowest available runner, and passes everywhere with headroom, is measuring the runner rather than the behavior.

The behavior under test is a *rejection* path: a busy native Codex process must be refused before any transaction file is published. Nothing about that assertion is inherently slow; the cost is in the setup that makes a process look busy.

**Impact**

Low correctness risk, real process cost. It fails release CI at random, and the fix is "rerun and hope", which is the same signal shape that makes a genuine regression easy to wave through. During the v2.14.0 release I had to distinguish this from a real failure by hand before publishing.

**Suggested direction**

Raise the budget for this specific case with a comment naming the hosted-macOS-runner reason, or make the busy-process setup deterministic so the assertion does not depend on wall time at all. The second is better if the setup currently waits for an observable side effect that could be awaited explicitly instead.

**Version**

Seen at `bf6186733` (v2.14.0), Bun 1.3.14. `tests/native-profile-manager.test.ts` is not touched by the v2.13.0 → v2.14.0 delta, so this predates that release.

Ingwannu:
## Summary

- give the deterministic busy-process rejection case its own 10-second Bun harness budget
- document why hosted macOS full-suite setup can cross the default deadline
- leave the product's SQLite lock deadline and `CODEX_BUSY` behavior unchanged

Closes #1563

## Verification

```bash
# repeated five times under the requested 2-core limit
taskset -c 0,1 bun test --isolate tests/native-profile-manager.test.ts --test-name-pattern 'normal switch rejects a busy native Codex process'
bun run typecheck
bun run privacy:scan
git diff --check
```

Focused result: 5/5 passes. Typecheck and privacy scan are green.

## Checklist

- [x] The change is limited to the flaky test harness budget.
- [x] The product timeout and runtime behavior are unchanged.
- [x] The hosted-macOS reason is recorded next to the test.
- [x] No GUI or wording surface changed.

## Summary by CodeRabbit

* **Tests**
  * Increased rollback and busy-process test timeouts to 10 seconds for more reliable test execution.
  * Added comments explaining the timeout rationale.

### DIFF
--- a/tests/native-profile-manager.test.ts
+++ b/tests/native-profile-manager.test.ts
@@ -1069,6 +1069,10 @@ describe("native main profile transactions", () => {
     ]);
   }, 10_000);
 
+  // This rollback case uses the same encrypted-vault and SQLite setup as the
+  // transaction test above. Hosted macOS can complete the product recovery
+  // successfully after Bun's 5 s default, so the harness must not become the
+  // shorter deadline.
   test("a read-back mismatch restores the exact source and removes the journal", async () => {
     const f = await enrolledFixture();
     const authPath = f.manager.context.authPath;
@@ -1087,7 +1091,7 @@ describe("native main profile transactions", () => {
     expect(readFileSync(join(f.codexHome, "auth.json"), "utf8")).toBe(f.source);
     expect((await failing.doctor()).recoveryPending).toBe(false);
     expect((await failing.list()).activeProfileId).toBe(f.sourceProfile.id);
-  });
+  }, 10_000);
 
   test("rollback verification failure retains the encrypted recovery journal and never claims success", async () => {
     const f = await enrolledFixture();
--- a/tests/native-profile-manager.test.ts
+++ b/tests/native-profile-manager.test.ts
@@ -1113,6 +1113,10 @@ describe("native main profile transactions", () => {
     expect(journal).not.toContain("opaque-refresh-target");
   });
 
+  // The rejection is immediate once the injected process probe runs, but the
+  // shared enrollment fixture still performs encrypted-vault and SQLite setup.
+  // Hosted macOS can push that setup past Bun's 5 s default under full-suite
+  // load, so keep the harness budget above the manager's own lock deadline.
   test("normal switch rejects a busy native Codex process before publishing any transaction file", async () => {
     const f = await enrolledFixture();
     const blocked = new NativeProfileManager({
@@ -1126,7 +1130,7 @@ describe("native main profile transactions", () => {
     expect(readFileSync(blocked.context.authPath, "utf8")).toBe(f.source);
     expect(existsSync(blocked.context.journalPath)).toBe(false);
     expect((await blocked.list()).activeProfileId).toBe(f.sourceProfile.id);
-  });
+  }, 10_000);
 
   test("vault-write failure after auth replacement restores exact source and removes the journal", async () => {
     const f = await enrolledFixture();


Ingwannu:
The first exact-head run did not fail the changed native-profile case. GitHub annotated the unrelated existing `tests/server-auth.test.ts:1802` WebSocket terminal timeout; all Linux shards and the other gates passed. I have rerun only the failed jobs once to distinguish that hosted-macOS flake from this one-line timeout fix. I will not mark this merge-ready unless the rerun is green.

Ingwannu:
Correction: this environment's GitHub CLI does not support selecting only failed jobs, so the prior rerun command was rejected and changed nothing. I have now started one full exact-head rerun of run `31659392694`; the original annotated failure remains the unrelated `tests/server-auth.test.ts:1802` WebSocket terminal timeout.

Ingwannu:
The exact-head macOS rerun exposed the same harness-boundary issue in the adjacent rollback verification: the product assertions completed correctly, but hosted macOS took 7.8 s and Bun's 5 s default expired. I added a documented 10 s budget only to that rollback regression; the product's lock and recovery deadlines are unchanged. Local validation on exact head `bef2869c732adc9302a9bdacf408ee585d654a1e`: focused case 5/5 under a two-core limit, full native-profile-manager suite 49/49, typecheck, privacy scan, and diff check green. Fresh exact-head CI is now running.

Ingwannu:
Exact-head CI is now fully green at `bef2869c732adc9302a9bdacf408ee585d654a1e`, including the macOS full suite and all Linux shards/gates. The product lock/recovery behavior remains unchanged; this PR only gives the two hosted-macOS-sensitive transaction regressions a harness budget above the product's own deadline. This is my PR, so I am leaving approval and merge to an independent maintainer.

Wibias (commented):
No code blocker from my review. This stays scoped to the test harness: the 10s budget is documented next to the affected cases, while the product lock/process behavior remains unchanged. The exact head also went green across the repository CI matrix, including the hosted macOS suite that reproduced the flake.

Please rebase onto current `dev` and rerun CI; assuming that stays green, this is good to merge.

PR #1576, opened 2026-08-13
fix(responses): reject undeclared routed tool calls

### LINKED ISSUE #1544: [Bug] DeepSeek CodeModeOnly: unsupported top-level `apply_patch` call aborts
### Client or integration

Codex App

DeepSeek (`deepseek/deepseek-v4-flash`)

### OpenCodex version

`2.13.0`

### Endpoint or capability

Routed Responses / CodeModeOnly tool calls / file editing

### Current behaviour

In a fresh Codex App conversation routed through OpenCodex to DeepSeek V4 Flash, the model can emit a top-level `apply_patch` call even though `apply_patch` is **not present in the top-level tool schema** for the request.

That call fails with only:

```text
aborted
```

In the same conversation, patching works through the intended Code Mode path:

```text
exec -> tools.apply_patch(...) -> {}
```

The file is created, verified, and deleted successfully.

The exposed top-level tools reported for the request include `exec`, `wait`, `request_user_input`, `search`, and related tools, but **not `apply_patch`**.

### Expected behaviour

For routed `code_mode_only` models, tool guidance and accepted tool calls should match the tool schema actually exposed to the model.

If patching is available only through Code Mode, the model should use:

```text
exec -> tools.apply_patch(...)
```

OpenCodex should not encourage or restore an undeclared top-level `apply_patch` call. If such a call is nevertheless received, it should produce an actionable compatibility error rather than a bare `aborted` result.

```text
Client: Codex App
OpenCodex: 2.13.0
Provider/model: DeepSeek / deepseek-v4-flash
Conversation: fresh

1. Direct path
   tool: apply_patch
   result: aborted

2. Code Mode path in the same conversation
   tool: exec
   inner tool: tools.apply_patch
   result: {}
   file created, verified, deleted

3. Tool surface
   top-level apply_patch: absent
```

The direct failure was reproduced in another fresh DeepSeek conversation.

### Actual response or error

```text
apply_patch -> aborted
exec -> tools.apply_patch(...) -> {}
```

There is no patch syntax or filesystem error on the failed path. The nested Code Mode path works in the same environment.

### Upstream documentation

DeepSeek Tool Calls documentation:
https://api-docs.deepseek.com/guides/tool_calls

Possibly related upstream Codex issue:
https://github.com/openai/codex/issues/30648

Relevant OpenCodex Code Mode work:
https://github.com/lidge-jun/opencodex/pull/1361

The working nested path strongly suggests the patch runtime itself is healthy. The likely failure is a mismatch between the routed CodeModeOnly tool contract and a direct `apply_patch` call being generated or restored somewhere in the request/response path.

Please inspect:

1. CodeModeOnly instructions/guidance for references to direct `apply_patch` when that tool is not exposed.
2. DeepSeek Responses conversion/restoration for legacy or undeclared custom-tool calls.
3. Whether undeclared top-level patch calls should be rejected with a clear diagnostic or mapped to the supported Code Mode path.

### Additional context

Controls:

- Reproduced in fresh DeepSeek conversations, so this is not inherited model-switch history.
- `exec -> tools.apply_patch` succeeds in the same environment, ruling out the patch engine and filesystem permissions as the primary failure.
- A fresh GLM routed conversation also succeeds through `exec -> tools.apply_patch`.
- Existing issues #372, #1017, and #1388 cover malformed or converted patch payloads / exact-match failures. This report is specifically about an **undeclared top-level tool call under CodeModeOnly**.

### Checks

- [x] The expected behaviour is based on the concrete tool contract and the working `exec -> tools.apply_patch` control.

Ingwannu:
## Summary

- retain the exact request-visible tool-name set alongside the existing namespace/custom-tool bridge maps
- fail streaming and non-streaming routed turns before emitting a client tool item when the provider invents an undeclared name
- keep `exec -> tools.apply_patch(...)` as the Code Mode path instead of guessing an automatic conversion for top-level `apply_patch`
- record the declared-tool membership decision in the transport architecture document

Closes #1544

## Verification

```bash
bun test tests/responses-parser.test.ts tests/responses-stream-tool-events.test.ts tests/bridge.test.ts
bun run typecheck
bun run privacy:scan
git diff --check
```

Focused result: 87 pass / 0 fail. Typecheck and privacy scan are green.

## Checklist

- [x] Streaming undeclared calls fail with an actionable compatibility error.
- [x] Non-streaming undeclared calls fail with the same contract.
- [x] Declared function, namespace, custom, and tool-search mappings remain covered.
- [x] No automatic executable tool conversion is introduced.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to reject tool calls for names not declared in the request.
  * Streaming responses now fail immediately without emitting executable tool events.
  * Non-streaming responses report a clear compatibility error with failed status and no output.
* **Documentation**
  * Documented fail-closed handling for undeclared tool names across response modes.
* **Tests**
  * Added coverage for undeclared tools, allowed tools, filtered selections, tool search, and disabled tool choices.

### DIFF
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -186,6 +186,8 @@ export function bridgeToResponsesSSE(
      * from this callback instead of re-parsing the bridged SSE.
      */
     onUsage?: (usage: OcxUsage | undefined) => void;
+    /** Request-visible tool names. When present, an upstream call outside this set fails closed. */
+    declaredToolNames?: ReadonlySet<string>;
     translatorBudget?: TranslatorBudget;
     /**
      * Conversation identity for the reasoning replay cache (issue #950).
@@ -974,6 +976,23 @@ export function bridgeToResponsesSSE(
               if (currentToolCall) closeCurrentToolCall();
               const mapped = toolNsMap?.get(event.name);
               const realName = mapped?.name ?? event.name;
+              if (options?.declaredToolNames && !options.declaredToolNames.has(event.name)) {
+                const failure = responseError(
+                  502,
+                  "upstream_error",
+                  `routed provider emitted undeclared client tool "${event.name}"; only request-declared tools may be called`,
+                );
+                emit("response.failed", {
+                  response: {
+                    ...responseSnapshot("failed", finishedItems),
+                    error: failure,
+                    last_error: failure,
+                  },
+                });
+                reportTerminal("failed");
+                terminalEvent = true;
+                break;
+              }
               const ns = mapped?.namespace;
               const toolSearch = toolSearchToolNames?.has(realName) ?? false;
               const freeform = !toolSearch && (freeformToolNames?.has(realName) ?? false);
@@ -1359,6 +1378,8 @@ function buildResponseJSONWithBudget(
   options?: {
     hideThinkingSummary?: boolean;
     toolNsMap?: Map<string, { namespace: string; name: string }>;
+    /** Request-visible tool names. When present, an upstream call outside this set fails closed. */
+    declaredToolNames?: ReadonlySet<string>;
     freeformToolNames?: Set<string>;
     toolSearchToolNames?: Set<string>;
     /** Remote compaction v2 turn — append one synthetic compaction output item (see bridgeToResponsesSSE). */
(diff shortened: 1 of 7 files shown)


Wibias (commented):
No code blocker from my review. The request-visible declared-tool set is derived from the existing toolChoice-filtered catalog, and both streaming and non-streaming paths fail closed before an undeclared tool becomes client-executable. Keeping `exec -> tools.apply_patch(...)` as the supported Code Mode path instead of guessing a translation is the right boundary for #1544.

Please rebase onto current `dev` and rerun CI; if that remains green, this is good to merge.

PR #1579, opened 2026-08-13
fix(providers): remove OAuth credentials on provider delete

laerad777:
## Summary
- delete the full OAuth account set when a configured provider is successfully removed
- preserve unrelated provider credentials
- keep blocked provider deletions fail-closed before credential mutation

## Verification
- RED: focused regression failed because the deleted provider credential remained
- GREEN: focused regression passed
- provider deletion regression group: 5 passed
- `bun run typecheck`
- `bun run privacy:scan`

## Checklist
- [x] Focused regression test added
- [x] Typecheck passes
- [x] Privacy scan passes
- [x] No unrelated changes

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Deleting a provider now also removes its associated account credentials.
  * Credentials for other providers remain unaffected.

* **Version Update**
  * Updated the package version to 2.14.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bitkyc08/opencodex",
-  "version": "2.13.0",
+  "version": "2.14.0",
   "description": "Universal provider proxy for OpenAI Codex & Claude Code — use any LLM with Codex CLI/App/SDK and Claude Code",
   "type": "module",
   "main": "./bin/package-main.mjs",
--- a/src/server/management/provider-routes.ts
+++ b/src/server/management/provider-routes.ts
@@ -27,7 +27,7 @@ import {
   submitManualLoginCode,
   upsertOAuthProvider,
 } from "../../oauth";
-import { removeCredential } from "../../oauth/store";
+import { replaceProviderAccountSet } from "../../oauth/store";
 import { providerDestinationResolvedError } from "../../lib/destination-policy";
 import { reconcileLiveStateStores } from "../../lib/state-store-registrations";
 import { ProviderOutboundPolicyError, providerOutboundGet, providerOutboundPost, providerRedirectError } from "../../lib/provider-outbound";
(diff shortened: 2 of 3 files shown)


Ingwannu (changes requested):
The direction is valid, but this authentication-boundary deletion is not failure-safe yet, so I am requesting changes on the exact head 018a7db972db.

Two release blockers remain:

1. src/server/management/provider-routes.ts persists the config deletion before replaceProviderAccountSet. If the OAuth-store lock, ACL hardening, or auth.json write fails, DELETE returns an error after the provider is already gone, while its credentials remain. Retrying the route then returns 404, so the two authorities cannot converge. Reordering the writes only reverses the partial-commit failure. Please add a staged/durable recovery contract or an equivalent transaction journal, with deterministic failure injection for each write boundary.

2. Provider deletion does not cancel an in-progress OAuth login. reconcileOAuthFlowState deliberately preserves unfinished flows, and runLogin can later save a credential and upsert the provider again after DELETE returned success. Cancel/settle that provider login before committing deletion, and add a race regression proving a late login cannot resurrect either the provider row or account set.

Also remove the unrelated package.json version bump; it incorrectly expands the release/dependency surface and trips the sponsorship gate. Expand the success regression to use multiple removable accounts and compare the unrelated retained credential values exactly.

I ran the focused provider-deletion group on this head: 4/4 passed. Those tests cover only the happy path and therefore do not discharge the failure/race blockers above. Keep this draft until the security-sensitive failure paths, typecheck, privacy scan, full cross-platform CI, and an independent security review are green.

lidge-jun:
Landed on `dev` as f1c84a427 via `git cherry-pick -x 018a7db97`, preserving your authorship.

The branch was not merged as-is because it also carried bf6186733 (`release: v2.14.0`), which is why intake marked it `hygiene-blocked`. Cherry-picking the single fix commit takes the credential-deletion change without importing that release commit into `dev`.

The fix itself is unchanged: deleting a configured provider now removes its full OAuth account set, unrelated provider credentials are preserved, and blocked deletions still fail closed before any credential mutation.

Verification on the merged `dev`: `bun run typecheck` clean, `bun run test` 11516 pass / 0 fail across 717 files.

Closing since the change is on `dev`; thanks for the fix.

PR #1583, opened 2026-08-13
fix(cursor): pin unified exec so wait cannot ship alone

jonathanli12:
## Summary

- Cursor transport truncation only pinned `exec_command` / `shell_command`. Codex Desktop advertises unified `exec` plus `wait` (bare or `opencodex-responses`).
- A crowded catalog therefore dropped `exec` and left `wait`, so `cursor/grok-4.6-fast` and Cursor-first combo turns loop on `exec cell not found` / `exec cell new not found`.
- Treat `exec`, `exec_command`, and `shell_command` (bare or `opencodex-responses`) as the execution path, force-admit one of them, and omit `wait` when no cell creator remains.

## Verification

- `bun test tests/cursor-request-builder.test.ts` — 30 passed
- `bun test tests/cursor-*.test.ts` — 513 passed, 0 failed
- `bun test tests/cursor-request-builder.test.ts tests/cursor-tool-definitions.test.ts` — 49 passed
- GitHub `hygiene` and `enforce-target` — passed

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Improved tool selection to prioritize execution and wait actions when tool limits are reached.
  * Added support for recognizing unified and namespaced execution tools.
* **Bug Fixes**
  * Prevented non-execution tools from displacing essential execution tools.
  * Automatically removes wait actions when no executable action can fit within limits.
* **Tests**
  * Added coverage for tool-count and size limits, execution tool retention, and oversized tool handling.

### DIFF
--- a/src/adapters/cursor/request-builder.ts
+++ b/src/adapters/cursor/request-builder.ts
@@ -21,6 +21,8 @@ import {
   cursorToolsForActivePrompt,
   isCursorStructuredEditToolName,
   isBareCodexShellBridgeTool,
+  isCursorExecutionPathTool,
+  isCursorWaitTool,
 } from "./tool-definitions";
 import { lookupCursorThreadConversation } from "./thread-continuity";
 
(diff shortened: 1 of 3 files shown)


jonathanli12:
@coderabbitai review

Wibias (changes requested):
Two correctness issues remain before this is merge-ready.

1. `wait` can still be advertised without any execution-path tool. `applyCursorToolBudget()` has a fitting-catalog fast return before the cleanup logic, so a catalog such as `[wait]` returns unchanged. The later cleanup is also gated on `eligibleHasExecutionPath`, so tool-choice filtering can leave `wait` behind after all exec tools become ineligible. The final invariant should be simple: if the admitted catalog has no `isCursorExecutionPathTool()` result, remove `wait` on every return path. Please add a fitting `[wait]` regression test as well.

2. The force-admission loop can unnecessarily try to admit a second execution alias even when one execution-path tool is already retained. In a constrained catalog that can evict useful tools (`wait`, `apply_patch`, selected tools, etc.) just to make room for another exec alias. The stated requirement is to retain at least one execution path, so only run force-admission when the kept set currently has zero execution-path tools.

Please cover the final behavior with tests for `[wait]`, `[wait, oversized exec]`, a fitting `[wait, exec]`, and a constrained catalog containing multiple exec aliases where retaining one does not evict useful tools for a second.

PR #1604, opened 2026-08-13
fix: preserve Daybreak Blue Codex capabilities

luvs01:
## Summary

- Preserve `gpt-daybreak-blue-latest` as an account-observed, selector-qualified native id without widening the bare native allowlist.
- Let an explicit `openai/gpt-daybreak-blue-latest` custom model on the canonical Codex-forward provider inherit the pinned GPT-5.6 Sol capability snapshot while keeping the Daybreak wire id.
- Keep the separately billed `openai-apikey/daybreak-blue-latest` API route and its 1.05M / 922k limits isolated from the 372k Codex-login contract.
- Normalize the built-in OpenAI provider's omitted `authMode` to its registry default (`forward`) for catalog discovery, matching routing behavior and avoiding a stray `/models` request.
- Add English and Korean route documentation and regression coverage for sync, routing, account isolation, strict metadata, and compaction values.

This builds on #1515: that PR preserves account-observed unknown native ids as selector-qualified rows. This follow-up covers the explicit canonical Codex-forward custom-model path and does not create a bare Daybreak row or grant account entitlement.

## Runtime verification

The patch was deployed to a local OpenCodex 2.14.0 installation with an explicit canonical Codex-forward custom-model entry. After selecting a ChatGPT Codex account known to have Daybreak entitlement, one Codex-shaped streaming request to `openai/gpt-daybreak-blue-latest` returned HTTP 200, emitted `response.completed`, and produced `OK`. The usage ledger recorded the exact requested selector, exact upstream wire id `gpt-daybreak-blue-latest`, the `main` account label, and `gpt-5.6-sol` as the capability resolution.

This smoke test also confirmed that entitlement is account-specific: two pool accounts without Daybreak access returned the upstream unsupported-model response before the entitled account was selected. The PR preserves routing and metadata; it does not claim or manufacture entitlement for every ChatGPT account.

## Verification

- Rebased onto current `dev` commit `1849b947` (`HEAD` `2ff758c7`, one commit ahead).
- `bun run typecheck`
- `bun run privacy:scan`
- `node_modules/bun/bin/bun.exe test --isolate --timeout 30000 tests/codex-catalog.test.ts tests/codex-catalog-sync-hardening.test.ts tests/codex-convergence-account-selectors.test.ts tests/codex-routing.test.ts` — 341 pass, 0 fail
- `cd docs-site && node_modules/bun/bin/bun.exe run build` — 265 pages built
- `git diff --check origin/dev...HEAD`

[...]

PR #1607, opened 2026-08-13
feat(codex): add routed tool-discovery compatibility profiles

lidge-jun:
## Summary

Replaces the hard-coded Cursor/non-Cursor boolean behind routed `supports_search_tool` with a resolved, route-scoped policy, and closes the unresolved P2 review finding from #1596.

**Defaults do not move.** With no configuration, non-Cursor routed rows stay deferred (`supports_search_tool: true` + `web_search_tool_type: "text_and_image"`) and Cursor stays direct, exactly as #1596 shipped. Every pre-existing catalog test passes unmodified — that is the proof, not an assertion.

Two optional provider settings:

```ts
routedToolDiscovery?: "auto" | "deferred" | "direct";
modelRoutedToolDiscovery?: Record<string, "auto" | "deferred" | "direct">;
```

Precedence is **Cursor hard fence > model override > provider override > auto**. `auto` resolves to deferred for non-Cursor rows and never reaches serialization: `CatalogModel.toolDiscoveryMode` carries only the resolved value, resolved once in `applyProviderConfigHints()` so configured, live-discovered, cached and combo-derived rows all agree. Combos resolve conservatively — one direct member forces the combo direct, since a single public row cannot vary after target selection — and the field is emitted only when it departs from the default. Both the explicit fingerprint and the provider-graph identity see the new fields, so two different policies cannot share a stale gather.

### The defect this closes

`parsing.ts` fenced Cursor on `slug.startsWith("cursor/")` while `sync.ts` fenced on `model?.provider === "cursor"`. A `cursor/`-aliased combo whose canonical provider is `combo` was therefore classified **differently depending on whether a template happened to be available**, making discovery mode and payload size depend on template availability. Both paths now share one `isCursorRoute()` helper: provider identity decides, and the slug prefix is only a fallback for callers with no `CatalogModel`.

### What `direct` is not

Under code mode Codex installs nested MCP tools on the `tools`/`ALL_TOOLS` globals in **both** exposures, so for an eligible tool `direct` buys no reachability. It moves full schemas into `exec.description` at the measured 2.7x turn-1 cost and changes `tool_search` construction, and it cannot repair a tool removed by `direct_only_tool_namespaces`, `excluded_tool_namespaces`, or MCP/App policy filtering. The types, docs and `structure/03` all say so, so nobody reaches for this expecting a reachability fix. Full analysis with upstream citations is in the parent PR's devlog unit.

### Config admission

The load path degrades a malformed value with `.catch(undefined)` so a typo cannot cost a user their providers or credentials; the write boundary rejects it outright with a path-specific `schema_invalid: providers.<name>.<field>: must be auto, deferred, or direct`. The validator reads only own **data** properties through `Object.getOwnPropertyDescriptor`, so an accessor- or prototype-polluted candidate is rejected without ever invoking the getter — covered by a test that asserts the getter call count stays at zero.

Also corrects `structure/03`: Cursor emits `supports_search_tool: false` and omits `web_search_tool_type`; it does not "advertise neither flag" (CodeRabbit's unresolved note on #1596).

## Verification

Full `bun run typecheck` and `bun run test` were run on a **Linux x86_64 host**, matching CI.

- Focused suites, all green (392 pass / 0 fail across 7 files): `codex-tool-discovery-mode`, `catalog-cursor-search`, `codex-catalog`, `parallel-tool-calls-optin`, `e2e-style/phase100-native-parity`, `config`, `config-user-edits`.
- `tests/codex-tool-discovery-mode.test.ts` is new — 25 cases covering the full precedence matrix, the Cursor hard fence by name and by adapter, the `cursor/`-aliased combo case, `modelRecordValue` matching including the deliberate dated-variant miss, sibling isolation on a mixed gateway, combo derivation, and config admission including the accessor and prototype-pollution cases.
- Full-suite results are posted in a follow-up comment once the Linux run completes.

One pre-existing test caught a real mistake during development: an unconditional `toolDiscoveryMode` key broke `codex-catalog`'s combo-shape assertion. Making the field conditional matches the `parallelToolCalls` precedent and keeps default rows byte-identical — exactly the regression fence that test exists to be.

## Checklist

Security-relevant surface: this PR touches config admission. The write-boundary validator is deliberately descriptor-first so a hostile config candidate cannot execute a getter during validation, and the load path degrades rather than discarding a provider's credentials. No secrets, tokens, or request bodies are logged; `privacy:scan` is green.

---

**Stacked PR.** Based on `codex/routed-tool-discovery-devlog` (#1606), which carries the plan and verification record. `enforce-target` exempts children whose base is an open PR's head branch. Retarget to `dev` once #1606 lands.


lidge-jun:
## Full suite verification — Linux x86_64

Run on a Linux host matching CI, at the exact PR head `2fd318b2a`.

```
bun x tsc --noEmit    → TSC_EXIT=0
bun run test          → 11515 pass / 11 skip / 0 fail
                        54881 expect() calls
                        Ran 11526 tests across 718 files. [424.92s]
                        TEST_EXIT=0
```

`bun run privacy:scan` passes.

### Review history

This branch went through three rounds of independent adversarial code review. Every finding was reproduced before being fixed, and the fixes are separate commits so the progression is auditable:

- `c1d655deb` — initial implementation.
- `b8cb797af` — closes four blockers: a default change for `cursor/`-aliased combos, a getter-safe write boundary that treated accessors as absent fields, `toolDiscoveryMode` not reaching custom-model and trusted-OpenAI row construction, and a Cursor fence that disagreed between resolution and serialization for an adapter-based gateway.
- `2fd318b2a` — closes the residual: prototype-inherited properties were classified as absent, so an inherited `"direct"` was accepted and persisted and an inherited getter ran once. `key in target` now classifies without reading.

Final verdict: **PASS**, with no remaining blockers.

### One tradeoff worth a reviewer's attention

`isCursorRoute()` unions its signals rather than ranking them. That is deliberate: the two construction paths historically disagreed, so any single-signal reconciliation would silently *unfence* rows one path used to fence, and an under-fenced Cursor row advertises a deferred surface its transport cannot serve.

The cost is real and documented in the source comment: a combo may legally be aliased `cursor/<name>` (only `combo/` is reserved), and such a row gets fenced — losing hosted-search metadata and taking Cursor's parallel-tool advertisement. That is exactly the shape it had before this PR, so this preserves behavior rather than regressing it. The clean fix is to reserve the `cursor/` alias prefix as a deliberate, documented compatibility break, which belongs in its own change.

lidge-jun:
Deferred from today's landing round with a concrete blocker: this branch now conflicts with `dev`.

#1602 (routed passthrough deferred-tool promotion) and #1604 (Daybreak Blue capability preservation) both landed a few hours ago and touch the same regions this PR rewrites — the merge stops on `src/codex/catalog/provider-fetch.ts` and `src/codex/catalog/sync.ts`.

I resolved conflicts by hand for #1581 during this round, but that was a four-line alignment prop. This is 798 lines replacing the hard-coded routed `supports_search_tool` boolean with a resolved route-scoped policy; a blind conflict resolution there could silently change which routes get deferred discovery, which is exactly the behavior the PR is trying to make explicit. That belongs to whoever holds the design intent.

Current `dev` head is 9a4716f84. Rebase onto it and the merge path should be clear — the defaults this PR preserves (`supports_search_tool: true` for non-Cursor routed, Cursor direct) are unchanged by the merges above.

PR #1613, opened 2026-08-13
fix(antigravity): persist replay expiry

luvs01:
## Summary

- Mark Antigravity replay snapshot state dirty when the periodic or lazy TTL sweep removes at least one in-memory session.
- Coalesce a sweep into one durable mutation generation while preserving the existing per-session deletion and byte-accounting path.
- Add a deterministic regression proving that a signature already flushed to disk is removed by the next explicit sweep and flush.

The replay cache already expired entries in memory, but expiry did not advance the snapshot mutation generation. During an idle period, the protected `antigravity-replay.json` file could therefore retain an opaque thought signature after its one-hour in-memory lifetime until some unrelated replay mutation or a later load rewrote the file.

## Verification

- Bun 1.3.14: `bun test --isolate tests/google-antigravity-replay.test.ts tests/state-store-sweeper.test.ts` (72 pass)
- Bun 1.4.0-canary.1: the same focused command (72 pass)
- Bun 1.3.14 and Bun 1.4.0-canary.1: `bun run typecheck`
- Bun 1.3.14: `bun run privacy:scan`
- Latest-base integration sanity: `bun test --isolate tests/model-rename-migration.test.ts` (9 pass)
- Related lifecycle tests: `bun test --isolate tests/state-store-sweeper.test.ts tests/shutdown-drain.test.ts tests/server-background-lifecycle.test.ts tests/storage-mutation-race.test.ts` (43 pass)
- Regression activation: with only the production fix hunk removed, the new test fails because the durable snapshot still contains one session; restoring the fix makes it pass.
- Independent exact-diff review found no P0/P1/P2 issue in generation, debounce, flush, load, clear, or eviction interactions.
- Exact head `f9036a80798aeda68bfaa5f85b05a611b30256ba` is rebased onto `dev` at `a1e5192b75edbf6dcacae51a30912fab93906f87`; the replay+sweeper focused set passes on Bun 1.3.14 and Bun 1.4.0-canary.1 (72 pass each), with typecheck passing on both runtimes and privacy scan passing on Bun 1.3.14.
- Full Bun 1.3.14 suite was attempted and is not claimed green: unrelated Windows effective-account lookup failures and existing 5-second integration timeouts cascaded before Bun crashed after 1,355 seconds with an internal index-out-of-bounds panic (exit 3, peak RSS 2.08 GB). The changed subsystem and lifecycle suites above pass on the exact head.

## Checklist

- [x] Docs or release notes were updated when needed. This is an internal persistence-lifecycle correction with no user-facing configuration or command change.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. The regression uses a synthetic signature and the implementation adds no logging or response fields.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Expired replay sessions are now removed individually and their associated stored signatures are deleted from durable storage.
  * Persisted snapshots are updated automatically when expired sessions are swept.

* **Tests**
  * Added coverage confirming expired sessions and signatures are removed from saved snapshots.

### DIFF
--- a/src/adapters/google-antigravity-replay.ts
+++ b/src/adapters/google-antigravity-replay.ts
@@ -522,7 +522,15 @@ function refreshReplaySessionCandidate(key: string, entry: ReplayEntry): void {
 }
 
 function deleteExpiredReplaySessions(now: number): void {
-  for (const [key, entry] of replayCache) if (entry.expiresAtMs <= now) deleteReplaySession(key);
+  let deleted = false;
+  for (const [key, entry] of replayCache) {
+    if (entry.expiresAtMs > now) continue;
+    deleteReplaySession(key);
+    deleted = true;
+  }
+  // Expiry is a durable mutation too: rewrite the snapshot so opaque thought
+  // signatures do not remain at rest after their in-memory TTL has elapsed.
+  if (deleted) markReplayDirty();
 }
 
 /**
--- a/tests/google-antigravity-replay.test.ts
+++ b/tests/google-antigravity-replay.test.ts
@@ -604,6 +604,20 @@ describe("durable antigravity replay snapshot", () => {
     await flushAntigravityReplay();
   });
 
+  test("sweeping expired sessions removes them from the durable snapshot", async () => {
+    observeAntigravityReplay(MODEL, SESSION, [fcPart("get_x", { a: 1 }, SIG)]);
+    await flushAntigravityReplay();
+    expect(readFileSync(snapshotPath(), "utf8")).toContain(SIG);
+
+    expect(sweepExpiredAntigravityReplay(Number.MAX_SAFE_INTEGER)).toBe(1);
+    await flushAntigravityReplay();
+
+    const rawText = readFileSync(snapshotPath(), "utf8");
+    const snapshot = JSON.parse(rawText) as { sessions?: unknown[] };
+    expect(snapshot.sessions).toHaveLength(0);
+    expect(rawText).not.toContain(SIG);
+  });
+
   test("snapshot cap keeps the most recently active sessions", async () => {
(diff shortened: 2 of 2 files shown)


luvs01:
@coderabbitai review

luvs01:
@coderabbitai review

lidge-jun:
The implementation closes the reported persistence gap with the right mutation boundary. `deleteExpiredReplaySessions()` now calls `markReplayDirty()` once after deleting one or more expired sessions (`src/adapters/google-antigravity-replay.ts:524-534`), preserving the existing per-session deletion and byte-accounting path while ensuring the next flush rewrites the durable snapshot.

The regression is meaningful rather than tautological: it flushes a real observed signature, performs the public expiry sweep, flushes again, and verifies both an empty persisted session list and removal of the signature text (`tests/google-antigravity-replay.test.ts:607-619`). Removing the new dirty-mark hunk leaves the stale on-disk session behind, so the test exercises the actual failure mode.

This is not merge-ready for today's cut. The PR remains draft, its head is based on `b583d649` rather than the current `dev` tip, and the exact-head Cross-platform CI and React Doctor runs are `action_required`, not green. Please rebase onto current `dev` and obtain the required exact-head CI. No docs update is needed for this internal snapshot-lifecycle correction, and the patch adds no logging or credential exposure.

DISPOSITION: NEEDS-CHANGE

luvs01:
@coderabbitai review

Wibias:
Thanks for this fix. This closes a subtle but important durability gap: expired Antigravity replay sessions are no longer only removed from memory while their stored signatures can survive on disk until some unrelated later mutation. Persisting the expiry makes the one-hour replay lifetime real across restarts and idle periods too, which is cleaner both for correctness and for retained sensitive state. 🙏

PR #1623, opened 2026-08-13
fix(codex): harden routed apply_patch contracts

Wibias:
## Why

Current `dev` already contains the shared nudge fix and the fail-closed guard for undeclared routed tool calls. This PR hardens the remaining OpenCodex-owned `apply_patch` contract and makes future adapter additions inherit that contract automatically.

## C+ adapter architecture

This PR replaces the reflective adapter-completeness test with one production authority:

- `src/adapters/contracts.ts` defines the finite adapter wires and mandatory routed tool contracts.
- `src/adapters/registry.ts` is the authoritative adapter universe and factory boundary.
- `resolveAdapter()` constructs adapters only through that registry.
- Azure / Azure OpenAI strictly inherit the OpenAI Responses semantic contract.
- MiMo strictly inherits the OpenAI Chat semantic contract.
- Cursor stays direct because it owns a distinct `runTurn` path and a gated native-file fallback.

Every registered adapter inherits these contracts with no opt-out:

- `tools.code-mode-nested-helper`
- `tools.freeform-exact-roundtrip`
- `tools.tool-choice-final-catalog`
- `tools.continuation-replay`

`apply_patch` is the canonical executable fixture for those generic tool-fidelity semantics. It does not become an adapter API.

## Production hardening

Routed Codex catalog rows default a missing `apply_patch_tool_type` to `"freeform"` while preserving an explicit value from the upstream/template row. This keeps routed defaults safe without overriding an explicit compatible `"function"` representation.

Persisted/provider adapter IDs remain untrusted strings. Runtime construction validates them through the registry lookup instead of widening config types or adding circular dependencies.

## Authoritative conformance gate

`tests/apply-patch-conformance.test.ts` derives its cases from `ADAPTER_REGISTRY`, so a new adapter automatically becomes a new conformance subject.

The harness uses real production adapters plus deterministic protocol-keyed upstream fixtures and verifies:

1. routed catalog exposure, missing-value defaults, and preservation of explicit tool representation;
2. nested Code Mode `tools.apply_patch(...)` visibility in the final provider request;
3. no provider-visible guidance that forbids an available patch helper;
4. final `tool_choice` semantics, including native wire-level `none` modes;
5. exact custom/freeform identity and hostile patch-body restoration;
6. fragmented streaming through the real adapter parser and Responses bridge;
7. buffered parsing whenever the concrete adapter exposes `parseResponse`;
8. exact next-turn continuation/replay;
9. Cursor mutation ownership and the gated native fallback;
10. strict wrapper-parent validity and wrapper-cycle rejection;
11. provider/model adapter configuration IDs being members of the registry;
12. compile-time `AdapterWire` driver completeness;
13. Bun parser-based anti-bypass checks preventing server, router, or lab conformance construction outside the registry;
14. mutation/self-tests proving representative broken observations fail the expected contract.

Responses-native adapters are tested through their real passthrough request contract rather than being forced through a parser they do not use in production. Cursor has focused `runTurn` coverage with a deterministic transport that verifies fragmented `apply_patch` input is preserved byte-for-byte. MiMo conformance injects a deterministic JWT provider through the registry, so request construction does not call the Xiaomi bootstrap endpoint.

## Focused historical regressions retained

The generic C+ gate complements the existing high-resolution regression tests:

```bash
bun test tests/apply-patch-conformance.test.ts
bun test tests/apply-patch-cursor-registry-runturn.test.ts
bun test tests/apply-patch-code-mode-regression.test.ts
bun test tests/apply-patch-routed-safety-net.test.ts
bun test tests/apply-patch-adapter-nudge-regression.test.ts
bun test tests/apply-patch-cursor-mutation-policy.test.ts
bun test tests/apply-patch-catalog-contract.test.ts
bun test tests/apply-patch-responses-native-contract.test.ts
```

The old `apply-patch-adapter-nudge-completeness.test.ts` source parser is intentionally removed. Registry membership now supplies the adapter universe directly.


lidge-jun:
This is a serious conformance suite, and the focused checks are not superficial: the test drives every `ADAPTER_REGISTRY` entry through final advertisement, `tool_choice: "none"`, exact streamed/buffered restoration, continuation replay, and the Cursor native-mutation boundary (`tests/apply-patch-conformance.test.ts:477-621`). The routed-catalog correction is also correctly placed in production normalization (`src/codex/catalog/parsing.ts:303`), and the registry keeps Azure/MiMo as explicit semantic wrappers rather than pretending their factory behavior is identical (`src/adapters/registry.ts:63-83`).

I am not putting this into today's release train. The known user-facing bug was already corrected on `dev`; this PR additionally replaces the central adapter factory switch with a new registry authority (`src/server/adapter-resolve.ts:45-46`, `src/adapters/registry.ts:26-121`). That is shared routing/configuration behavior across 37 commits and 1,939 added lines, not a narrowly releasable regression fix. The exact-head Cross-platform CI run is also still incomplete — the macOS job remains `in_progress` — so the claimed all-adapter contract has not yet passed its release gate.

Please rebase and land this as a separately reviewed hardening follow-up once the exact-head macOS/Cross-platform CI completes successfully. The inspected diff introduces no security issue, but it should not merge solely to reinforce a fix already present on `dev`.

DISPOSITION: DEFER

lidge-jun (changes requested):

The goal—one production adapter registry and inherited routed-tool contracts—is sound, but this PR combines an architecture migration, factory rewiring, MiMo/Azure dependency injection, conformance infrastructure, and `apply_patch` behavior changes. That breadth makes regression attribution and rebase resolution unnecessarily risky.

Please split the delivery into: (1) registry/factory authority with behavior parity, (2) generic conformance harness derived from the registry, and (3) any remaining `apply_patch` production hardening. Land each on current `dev` with no adapter behavior delta in step 1. Keep focused Cursor structured-edit work in #1634 separate. An ADR/decision note covering wrapper cycles, runtime string validation, and extension policy should accompany the registry step.

lidge-jun:
Cherry-pick failed due to deleted src/adapters/contracts.ts and mimo-free.ts conflicts against current dev. The adapter registry architecture has diverged significantly. Recommend a fresh implementation against current dev.

PR #1632, opened 2026-08-13
fix(integrations): classify a json sibling edit as stale instead of conflict

### LINKED ISSUE #1631: OpenCode integration dead-ends in permanent 'conflict' after any unrelated user edit to opencode.json (e.g. adding an MCP server)
### Client or integration

Other

### Area

Other

### Summary

The OpenCode (and any non-`omp`) client integration dead-ends in a permanent `conflict` state as soon as the user makes **any** edit to their own config file — even edits that never touch the opencodex-owned block. Adding an MCP server to `~/.config/opencode/opencode.json` is enough: the dashboard shows `Konflikt`, the toggle turns off, and there is no non-destructive recovery path. The only UI option, restoring the snapshot, would revert the user's newer edits; the only clean fix is manual surgery (delete the `provider.opencodex` block by hand, then re-enable).

I expected: edits **outside** the opencodex-owned block (normal, frequent usage of one's own config — MCP servers, instructions, permissions) either don't trip the conflict state, or there is a supported, non-destructive way to re-apply/re-own from the dashboard or CLI.

**Root cause** — `classifyIntegration` (`src/integrations/state.ts`) checks two fingerprints:

```ts
if (clientId !== "omp" && fingerprint(input.fileText ?? "") !== input.record.fileFingerprint) {
  return { state: "conflict", reason: "foreign-edit" };   // ← fires on ANY file edit
}
if (recordedFragmentFingerprint(input.parsed, input.record) !== input.record.blockFingerprint) {
  return { state: "conflict", reason: "foreign-edit" };   // ← the actual tamper check
}
```

The whole-file check fires even when the second check proves the owned block is untouched. I understand this is deliberate (the test "whole-document serializers still conflict on an unrelated source edit" pins it): non-`omp` clients rewrite the whole document on apply, so a rewrite could normalize a user's formatting or drop comments, and the classifier fails closed. But the practical effect is that the integration breaks permanently on the first config edit the user makes themselves, with no recovery path.

**Proposal** — two options, not mutually exclusive; I'm happy to send a PR with regression tests for whichever direction fits the project's invariants:

- **A. Confirm-gated re-apply from `conflict` when the owned block is intact** (mirrors the existing `restore --confirm-drift` pattern). If `recordedFragmentFingerprint(parsed, record) === record.blockFingerprint`, apply merges into the current parsed document and a snapshot is taken first, so no user *content* can be lost — only formatting may normalize. CLI: `ocx integration client enable --client opencode --confirm-drift` (or similar); dashboard: a confirm dialog instead of the dead-end badge.
- **B. Classifier refinement for lossless cases**: when the owned block is intact **and** the file round-trips byte-identically through the client's serializer (`fingerprint(serializeDocument(parsed, format)) === fingerprint(fileText)`), classify `current`/`stale` instead of `conflict` — rewriting such a file provably cannot destroy formatting or comments. Files with custom formatting/comments still fail closed and would use path A.

### Reproduction

1. `ocx integration client enable --client opencode` → state `current`.
2. Add any MCP server (or any other key) to `~/.config/opencode/opencode.json` without touching `provider.opencodex`, e.g.:
   ```json
   "mcp": { "playwright": { "type": "local", "command": ["npx", "-y", "@playwright/mcp@latest"], "enabled": true } }
   ```
3. Dashboard → Integrations → OpenCode now shows `Konflikt`, toggle off.
4. `ocx integration client enable --client opencode` refuses with `<path> changed after opencodex wrote it`.
5. No non-destructive way out: "restore this state" reverts the user's own edit from step 2; the only clean recovery is deleting the `provider.opencodex` block manually and enabling again.

### Version

2.14.2

### Operating system

macOS 27 (arm64)

### Provider and model

_No response_

### Logs or error output

```shell
$ ocx integration client enable --client opencode
refused: /Users/<user>/.config/opencode/opencode.json changed after opencodex wrote it (conflict)

$ ocx integration client status --client opencode
clientId: opencode
state: conflict
```

_No response_

### Redacted configuration

_No response_

### Checks

RobinBially:
## Summary

- Fixes #1631: any user edit to a client config file (e.g. adding an MCP server to `opencode.json`) flipped the integration into a **permanent `conflict`** — even when every opencodex-owned fragment was untouched — with no non-destructive recovery path (toggle off, enable refused, restore reverts the user's own edits).
- `classifyIntegration` now distinguishes a **sibling edit** (owned block verified intact via `blockFingerprint`, checked **before** the file-level drift so the exemption can never mask an edit inside an owned block) from tampering. For `format: "json"` clients the sibling-edit case classifies as **`stale`** — the dashboard offers a refresh, and apply/disable merge around the user's entries and re-own the file. **Comment-capable formats (yaml, json5, toml) keep the hard `conflict`**: a whole-document rewrite there could drop user comments; strict JSON cannot carry comments (a commented file fails parsing long before this branch).
- **Number guard, the one layer the fix is not safe without:** `parseConfig` scans the raw json text (same posture as the TOML `inf`/`nan` guard) and refuses literals whose value a rewrite would actually change — overflow to Infinity (baked in as `null` by the merge clone), plain-digit integer runs provably rounded past 2^53 (the one spelling consumers like python's `json` read with exact integer semantics), underflow to zero (`1e-9999`), and `-0`. It also refuses **duplicate members**: `{"a":1,"a":2}` parses to a single `a`, so rewriting the parsed document would silently DELETE the earlier one — content loss, not the formatting normalization this PR promises, and reachable only because the classifier now hands a user-edited file to apply at all. Exponent spellings and exactly-representable big integers (`1e21`, `2^54`) stay usable end to end — that decision is pinned in comments and tests. The preflight refusal message names this value class instead of claiming a valid file "could not be parsed".
- `disableIntegration` proceeds on such drifted files too (removal touches only recorded fragment paths); its precondition comment now names the real invariant (block fingerprint, not file fingerprint). The classifier's module comment revises devlog 021 §3's unconditional whole-file rule; docs updated (en, zh-tw) including the number exception.
- **Scope:** this PR went through five adversarial multi-lens review rounds (correctness, security, plausibility, tests, complexity). Deeper hardening the review surfaced targets **pre-existing** exposure and is deliberately split out: nesting-depth ceiling + serializer value walk → #1637 (stacked on this PR); the same gaps for json5/yaml/toml → #1635.

## Verification

- `bun test tests/integrations-state.test.ts tests/integrations-serialize.test.ts tests/integrations-writer.test.ts` — 108 pass, 0 fail on this branch (sibling-edit stale via real file I/O, yaml/json5/toml conflict regression loop, sibling+block-edit ordering, 1e999 refusal and 2^54 symmetry for apply AND disable, scanner lexer edges, re-apply block shape and re-ownership).
- `bun test tests/integrations-invariants.test.ts tests/integrations-journal.test.ts tests/management-integration-routes.test.ts` — pass, 0 fail.
- `bun run prepush` against this exact head — full gate passed: typecheck clean, **11586 pass / 8 skip / 0 fail across 719 files (531s)**, privacy scan passed.
- Manually reproduced the dead-end on 2.14.2 (macOS, OpenCode client) before writing the fix; repro steps in #1631.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup. (Hardening of pre-existing exposure deliberately split into #1637/#1635 on review advice.)
- [x] Docs or release notes were updated when needed. (docs-site integrations guide, en + zh-tw + tr, incl. the OMP exemption CodeRabbit caught — the Turkish translation landed on dev after this branch opened and still described the old fail-closed rule, so it is carried along here.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (The exemption only relaxes classification when every owned fragment is fingerprint-verified intact; tampering still hard-conflicts; comment-capable formats stay fail-closed; the number guard fails closed on genuinely lossy literals.)

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - OpenCode and Pi integrations can preserve unrelated JSON edits during updates or disabling.
  - Integration status distinguishes safely mergeable changes from edits requiring attention.
  - Unsafe JSON values, including duplicate keys and unrepresentable numbers, are rejected.

- **Bug Fixes**
  - Prevented accidental loss or alteration of user-edited configuration values.
  - YAML, JSON5, TOML, and modified managed entries remain protected from unsafe changes.
  - OMP updates only its managed configuration range.

- **Documentation**
  - Updated integration rollback guidance across supported languages.


RobinBially:
Thanks — all six findings were correct. Four of them landed in this PR (the guard they concern is introduced here, not in the stacked #1637), verified in `7f2e2dfa9`:

**Positive-exponent underflow** — real, and the same class the guard exists to stop. `1e-9999` is a nonzero value `JSON.parse` had already flattened to `+0`; the check only asked for a leading `-`, so the rewrite wrote `0`. The significand now decides, so exact-zero spellings (`0`, `0.0`, `0e10`) and representable subnormals (`1e-320`) stay usable.

**Duplicate members** — real, and the worse of the two: content loss reported as success. Worth noting it is reachable *because* of this PR — before it, a sibling-edited strict-JSON file was a permanent `conflict` and was never rewritten. `jsonTextSafeToRewrite` now tracks decoded member names per open object, so `"a"` and `"a"` count as one member, matching what `JSON.parse` keeps. Regression added at the requested spot plus an end-to-end apply refusal asserting the file is byte-for-byte unchanged. The counter-cases are pinned too: same name in sibling objects, in array elements, and colons/braces inside strings all stay accepted.

**"An integer past 2^53"** — you are right that this names a value the code accepts; `tests/integrations-writer.test.ts` applies with 2^54 on purpose. All four texts now describe the actual condition (rounding) and carry the duplicate-member cause.

**The OMP rollback rule** — correct, `classifyIntegration` exempts `omp` from the whole-file drift check because its writer patches only its own `providers.opencodex` range byte-wise. Fixed in en, zh-tw and tr; while checking the client list I noticed Gajae Code was missing from the whole-document group as well, so it is named now.

The remaining two (scan before `JSON.parse`, the `installPi()` helper) belong to #1637 and are handled there.

Full `bun run prepush` on this head: 11586 pass / 8 skip / 0 fail across 719 files, privacy scan passed.

lidge-jun:
The ownership change is correct and narrowly scoped.

`src/integrations/state.ts:194-218` now verifies the recorded fragment fingerprint before considering whole-file drift. That preserves `conflict` when an opencodex-owned value changed, while allowing only strict-JSON clients with intact fragments to become `stale`; YAML, JSON5, and TOML remain fail-closed because whole-document serialization can discard comments. This resolves #1631 without granting the same rewrite permission to comment-capable formats.

The tests exercise the real writer path rather than merely mirroring the classifier: `tests/integrations-writer.test.ts:154-202` re-applies after a Pi sibling edit and verifies the user provider survives, then disables after the same drift and verifies only the owned provider is removed. `tests/integrations-state.test.ts:572-648` also pins the JSON-only exemption, comment-capable conflicts, and the combined sibling-plus-owned-edit conflict. I reviewed CodeRabbit's earlier findings and agree they were correctly fixed on the current head.

The release blocker is CI, not code: **Cross-platform CI** and **React Doctor** at exact head `7f2e2dfa927758a3041f3d02d1430128c30147d0` are `action_required` with no jobs, which I verified against the check-runs API. The successful visible checks are PR-target/label/hygiene automation only. Please approve or rerun both workflows and obtain successful results for this SHA before merging; the author's local prepush result is not a substitute for that gate.

DISPOSITION: NEEDS-CHANGE

Wibias:
Thanks @RobinBially for the careful fix and the thorough follow-up on the review findings.

This is especially useful because normal user changes next to our managed JSON block, such as adding an MCP server or another provider, no longer leave the integration stuck in a permanent conflict. At the same time, the change keeps the safety boundary intact: edits to our owned block still conflict, comment-capable formats stay fail-closed, and JSON cases that could lose or change user data are refused instead of being silently rewritten.

CI is green on the reviewed head, and this is now merged. Thanks!

PR #1636, opened 2026-08-13
fix(server): configure maxRequestBodySize to 256 MiB on Bun.serve listeners (#1601)

### LINKED ISSUE #1601: [Bug] Bun.serve has no maxRequestBodySize: >128 MiB /v1/responses bodies get an empty-body 413 that Codex renders as "Unknown error"
## Summary

`Bun.serve()` in `src/server/index.ts` never sets `maxRequestBodySize`, so Bun's default **128 MiB** cap applies to the `/v1/responses` HTTP listener. When Codex posts a body larger than that, **Bun rejects it at the HTTP layer before `fetch()` runs**, returning `413` with an **empty body**. Codex renders an empty error body as the opaque string `Unknown error` — exactly the failure mode #452 and `src/server/responses/passthrough-error.ts` were written to prevent, except this one happens one layer below anything opencodex controls.

Client-visible symptom:

```
unexpected status 413 Payload Too Large: Unknown error, url: http://127.0.0.1:10100/v1/responses
```

Because the handler never runs, the request leaves **no trace at all**: no `usage.jsonl` row, no `service.log` line. From the operator's side the proxy looks perfectly healthy (`ocx health` → `Proxy healthy`) while every turn of the affected session fails.

## Environment

- opencodex `2.10.0` (reproduced); **`2.14.1` is still affected** — see "Still present on latest" below
- Bun `1.3.14` (bundled)
- Codex CLI `0.147.0-alpha.6.5`
- macOS (darwin arm64)
- `websockets: false` in `config.json`, so Codex falls back to HTTP POST and always takes this path

## Root cause

`src/server/index.ts` (2.10.0, L391):

```ts
const server: Server<WsData> = Bun.serve<WsData>({
  port: listenPort,
  hostname: bindHost,
  idleTimeout: 255,
  async fetch(req, requestServer): Promise<Response> {
```

No `maxRequestBodySize` here, and `rg maxRequestBodySize src/` returns nothing anywhere in the package. Bun therefore applies its default of 128 MiB and answers oversized requests itself.

Every 413 opencodex *itself* produces carries a JSON envelope with a populated `error.message`:

- `formatErrorResponse` → `classifyError` (`src/lib/errors.ts`) always preserves `message`
- `decodeRequestErrorResponse` → `413 invalid_request_error` for `DecompressedBodyTooLargeError`
- `413 request_too_large` for `TranslatorBudgetExceededError`
- `formatPassthroughUpstreamError` explicitly re-wraps **empty upstream bodies** into `Provider error 413: (empty body)`

So an empty-body 413 cannot originate from opencodex's own code paths. That is what makes the Bun layer identifiable as the source.

## Evidence

1. **No 413 anywhere in the proxy's own records.** Status distribution across `~/.opencodex/usage.jsonl` (10k+ requests):

   ```
   200: 10766   499: 319   502: 145   429: 19   400: 6   402: 2
   413: 0
   ```

   `service.log` likewise has no 413 entry. The request never reached `fetch()`.

2. **The bodies really are over 128 MiB.** Two Codex sessions that recorded the error, measured from their rollout JSONL:

   | rollout | file size | `response_item` total | image-bearing tool outputs |
   |---|---|---|---|
   | `…17-17-29-019ffa32` | 243 MB | **199.9 MB** | 33 / **185.5 MB** |
   | `…13-40-27-019ff96b` | 463 MB | **272.6 MB** | 90 / **269.5 MB** |

   Individual `custom_tool_call_output` lines reach 25 MB / 9 MB / 3 MB — base64 images from Codex's built-in `image_gen`, replayed as conversation history on every turn. zstd request compression only recovers the base64 overhead (~75% of raw), so the wire body still lands around 150 MB and stays over the cap.

3. Both failures occurred on **sub-agent turns** — the error text is followed by `This agent's turn failed. If you still need this agent, use the available collaboration tools to give it another task.`

## Reproduction

1. Run a Codex session that accumulates >128 MiB of conversation input (easiest: dozens of `image_gen` calls; any large multimodal history works).
2. Route it through the opencodex proxy over HTTP (`websockets: false`).
3. The next turn fails with `unexpected status 413 Payload Too Large: Unknown error`, and nothing is written to `usage.jsonl` or `service.log`.

## Still present on latest

Checked the published `2.14.1` tarball. `Bun.serve` now takes a shared `serveOptions` object, which still does not set the field:

```ts
const serveOptions = {
  idleTimeout: 255,
  async fetch(req: Request, requestServer: Server<WsData>): Promise<Response> {
```

`rg maxRequestBodySize package/src/` → no matches. `MAX_DECOMPRESSED_BODY_BYTES` is still `256 * 1024 * 1024`.

## Impact

- The 128 MiB Bun default silently undercuts the 256 MiB `MAX_DECOMPRESSED_BODY_BYTES` admission cap the code intends to enforce. Requests between 128 MiB and 256 MiB are killed by Bun instead of reaching the guard that would explain itself.
- The failure is undiagnosable from the proxy side: healthy `ocx health`, zero log lines, zero usage rows.
- The client message is the opaque `Unknown error` that #452's wrapping exists to eliminate.

## Suggested fix

Set `maxRequestBodySize` on `Bun.serve` to at least `MAX_DECOMPRESSED_BODY_BYTES`, so the documented 256 MiB cap is the one that actually fires and rejections come back as a typed JSON error rather than an empty 413:

```ts
const serveOptions = {
  idleTimeout: 255,
  maxRequestBodySize: MAX_DECOMPRESSED_BODY_BYTES,
  async fetch(...) { ... }
}
```

Optionally make it configurable alongside `appOwnedMemoryBudgetMb`, and log a warning when a request is rejected for size so the operator has something to find.

Note this is distinct from #987, which concerned an **upstream gateway's** 10 MiB cap. This one is the local listener rejecting the request before any provider is contacted.

ardakrt:
## Summary
- Sets `maxRequestBodySize: MAX_DECOMPRESSED_BODY_BYTES` (256 MiB) on `serveOptions` for `Bun.serve()` in `src/server/index.ts` (both public and loopback listeners).
- Fixes #1601 where requests with body sizes between 128 MiB and 256 MiB (e.g. multimodal histories or large base64 image replays) were rejected by Bun's default 128 MiB HTTP listener cap with an empty 413 error (rendered as "Unknown error" in Codex).
- Aligns the Bun listener cap with OpenCodex's documented `MAX_DECOMPRESSED_BODY_BYTES` admission limit.
- Adds regression unit test `tests/server-request-body-size.test.ts`.

## Test plan
- Run `bun test tests/server-request-body-size.test.ts` (passed).
- Run `bun run typecheck` (passed with 0 errors).
- Run `bun run privacy:scan` (passed).

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a maximum request body size of 256 MiB for decompressed HTTP request data.
  * Improved protection against oversized request payloads.

* **Tests**
  * Added coverage confirming the request size limit.
  * Added a server health check to verify successful startup and request handling.

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -58,6 +58,7 @@ import { isCanonicalOpenAiForwardProvider, OPENAI_CODEX_PROVIDER_ID } from "../p
 import { providerContextCap } from "../providers/context-cap";
 import { providerCodexAccountMode } from "../providers/registry";
 import type { StorageCleanupPolicy } from "../types";
+import { MAX_DECOMPRESSED_BODY_BYTES } from "./request-decompress";
 import {
   CodexAccountCooldownError,
   cooldownErrorMessage,
@@ -726,6 +727,7 @@ export function startServer(port?: number, deps: StartServerDeps = {}): Server<W
     userCostOverlayReconciler = startUserCostOverlayReconciler({ liveConfig: config });
     const serveOptions = {
       idleTimeout: 255,
+      maxRequestBodySize: MAX_DECOMPRESSED_BODY_BYTES,
       async fetch(req: Request, requestServer: Server<WsData>): Promise<Response> {
       // The unauthenticated loopback listener (#1102) serves a fixed allowlist and nothing
       // else. Rejecting here, before any handler runs, is what keeps the surface from growing
--- a/tests/server-request-body-size.test.ts
+++ b/tests/server-request-body-size.test.ts
@@ -1,0 +1,44 @@
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { existsSync, mkdirSync, rmSync } from "node:fs";
+import { join } from "node:path";
+import { startServer } from "../src/server";
+import { MAX_DECOMPRESSED_BODY_BYTES } from "../src/server/request-decompress";
+import { installIsolatedCodexHome, type IsolatedCodexHome } from "./helpers/isolated-codex-home";
+
+const TEST_DIR = join(import.meta.dir, ".tmp-server-request-body-size-test");
+let isolatedCodexHome: IsolatedCodexHome | null = null;
+
+beforeEach(() => {
+  if (existsSync(TEST_DIR)) rmSync(TEST_DIR, { recursive: true });
+  mkdirSync(TEST_DIR, { recursive: true });
+  process.env.OPENCODEX_HOME = TEST_DIR;
+  isolatedCodexHome = installIsolatedCodexHome("ocx-server-body-size-codex-");
+});
+
+afterEach(() => {
+  isolatedCodexHome?.restore();
+  isolatedCodexHome = null;
+  if (existsSync(TEST_DIR)) rmSync(TEST_DIR, { recursive: true });
+});
+
+describe("server maxRequestBodySize (Issue #1601)", () => {
+  test("configures Bun.serve listener with MAX_DECOMPRESSED_BODY_BYTES (256 MiB)", () => {
+    expect(MAX_DECOMPRESSED_BODY_BYTES).toBe(256 * 1024 * 1024);
+  });
+
+  test("server listener accepts requests without failing at the Bun 128 MiB default", async () => {
+    const server = startServer(0);
+    try {
+      const port = server.port;
+      // Send a request to /healthz with a body larger than 0 bytes
+      const res = await fetch(`http://127.0.0.1:${port}/healthz`, {
+        method: "GET",
+      });
+      expect(res.status).toBe(200);
+      const data = await res.json() as { status: string };
+      expect(data.status).toBe("ok");
+    } finally {
+      void server.stop(true);
+    }
+  });
+});


lidge-jun:
The production change at `src/server/index.ts:727-730` is correctly scoped: both the public and loopback listeners spread the same `serveOptions` object at `src/server/index.ts:1638-1650`, so setting `maxRequestBodySize` there is the right implementation point.

### Blocking: the new test does not exercise the fixed listener boundary

`tests/server-request-body-size.test.ts:24-42` only makes a bodyless `GET /healthz`. It stays green if `maxRequestBodySize` is deleted from `serveOptions`, because the 128 MiB default is never approached and `/healthz` does not parse a request body. The constant-equality assertion at `:20-22` likewise only proves the existing parser constant.

Replace it with an actual `POST /v1/responses` whose wire body is strictly larger than 128 MiB and no larger than 256 MiB. Send it without admission credentials and assert the typed JSON `401 authentication_error` from `src/server/index.ts:1188-1216` rather than Bun's empty pre-handler `413`. That proves the request crossed the listener boundary without routing it upstream, and it fails when the new option is removed.

### Blocking: isolate and finish server teardown

`tests/server-request-body-size.test.ts:11-21` overwrites `process.env.OPENCODEX_HOME` and never restores it, so it can contaminate later tests in the shared Bun process. Save the previous value in `beforeEach` and restore it, or delete the variable if it was unset. Also change `void server.stop(true)` at `:40-42` to `await server.stop(true)` so the listener and background lifecycle finish before the next test starts.

This is shared server behavior. After adding the real boundary regression, run `bun run test`; the checks visible on this head are hygiene/target checks and CodeRabbit, not the repository test suite.

DISPOSITION: NEEDS-CHANGE

lidge-jun (changes requested):

The production change is directionally correct: setting Bun's listener cap to `MAX_DECOMPRESSED_BODY_BYTES` aligns the HTTP boundary with OpenCodex's 256 MiB admission limit.

The current regression does not exercise the bug. It checks the constant and sends a bodyless `GET /healthz`, so it would pass even if `maxRequestBodySize` were removed. Please add a behavior-level test that proves a request above Bun's 128 MiB default but below 256 MiB reaches the OpenCodex handler. A small injectable `Bun.serve`/serve-options seam is acceptable if allocating a 128+ MiB fixture in routine CI is too expensive. Also verify the shared options cover both public and loopback listeners.

The branch is draft and CI is currently `action_required`; rerun exact-head CI after the test is strengthened.

lidge-jun:

Issue #1635, opened 2026-08-13
Non-json client configs: no nesting-depth cap (json5/yaml resource spike) and silent big-integer rounding on rewrite (toml/json5/yaml)


RobinBially:
### Client or integration

Other

### Area

Other

### Summary

While reviewing #1631/#1632, adversarial measurement passes found two classes of pre-existing exposure in the client-config rewrite pipeline for the **non-json** formats. Neither is introduced or enlarged by #1632 (which guards the json path); both exist on `dev` today for every first apply onto a foreign config file, and are reported here so they don't get lost.

**1. No nesting-depth cap for json5/yaml (resource spike + raw error instead of structured refusal).**
`parseConfig` has no depth guard for `json5` and `yaml` (the json path gets one in the follow-up to #1632). Measured on Bun 1.3.14/JSC through the exact layers the writer calls (`parseConfig` → merge clone → `serializeDocument`): an openclaw json5 config nested 15,000 deep (bounded by the parser's own stack ceiling ~15,295) produces a ~450MB output string at ~1.1GB peak RSS in ~1s per management request; yaml accepts ~3,000 levels and drives the recursive `renderYaml`. Reachability requires local write access to a client config plus an authenticated management mutation — so a hardening issue, not an urgent vulnerability.

**2. Big-integer literals silently rounded for toml/json5/yaml.**
`Bun.TOML.parse`, `Bun.JSON5.parse` and `Bun.YAML.parse` all round `9007199254740993` to `9007199254740992`, and the corresponding renderers write the rounded value back without refusal. A first apply onto e.g. a kimi `config.toml` holding `session_id = 9007199254740993` as a sibling rewrites it to `…992` and reports success — exactly the failure class the TOML `inf`/`nan` guard (and, since #1632, the json number scan) treats as "a silent value change is worse than a refusal". json5 additionally loses `-0` (`Bun.JSON5.stringify({a:-0})` → `{a:0}`) while non-finite values are safe there (JSON5 can represent `Infinity`).

Suggested direction, mirroring the json approach: per-format raw-text or post-parse round-trip checks that classify such files `unsafe` before any rewrite, plus a shared nesting ceiling. I'm happy to pick this up after #1632 and the depth-hardening follow-up PR land, if the direction is agreeable.

### Reproduction

[...]

PR #1637, opened 2026-08-13
perf(integrations): depth-cap json configs and harden the serializer walk

RobinBially:
## Summary

- Follow-up hardening split out of #1632 on review advice (scope line: these guards address **pre-existing** exposure that the sibling-edit fix does not enlarge — every first apply onto a foreign config file always ran through the same recursive rewrite layers).
- **Depth cap at the trust boundary:** `parseConfig`'s raw-text scan for json now also counts container nesting against a shared `MAX_JSON_NESTING` (1000), and the scan runs **before** `JSON.parse` (CodeRabbit) — it never needed the parsed value, and scanning first means the hostile document is refused instead of materialized. `JSON.parse` handles hundreds of thousands of levels iteratively, but the downstream merge clone and `JSON.stringify` recurse: measured on `dev`, a 100KB file nested 50k deep passed parse, then died in serialization with a raw `RangeError` after a ~3.6GB allocation spike. Post-fix: `PARSE_FAILED` in 3ms, no spike. Brackets inside strings don't count.
- **Value-safety walk for builder/preview documents:** `serializeDocument('json')` refuses exactly the two values serialization itself damages — non-finite (written as `null`) and `-0` (written as `0`). Anything stricter re-created the recoverable-but-refused asymmetry #1632 closes (pinned by test). The walk uses iterative frames — memory stays proportional to nesting depth; a naive node stack cost ~18x the document size (measured: 1914MB → ~25MB for a 10M-number document) — and enforces the same nesting ceiling as the scanner, with clamped paths in refusal messages.
- **Boundary pins:** exactly 1000 levels parse AND serialize (one document through both layers — the constant is shared so the two layers cannot drift), 1001 refuses on both sides, clamp shape (head…tail) asserted.
- Related, deliberately not here: the same exposure for json5/yaml/toml is tracked in #1635.

**Stacked on #1632** — review only the last commit until that lands; the first two commits are #1632's.

## Verification

- `bun test tests/integrations-state.test.ts tests/integrations-serialize.test.ts tests/integrations-writer.test.ts tests/integrations-invariants.test.ts tests/integrations-journal.test.ts tests/management-integration-routes.test.ts` — 209 pass, 0 fail.
- Resource measurements above were taken with `bun -e` harnesses against these exact layers (parseConfig → merge clone → serializeDocument) on Bun 1.3.14/JSC, macOS arm64.
- `bun run prepush` against this exact head — full gate passed: typecheck clean, **11591 pass / 8 skip / 0 fail across 719 files (532s)**, privacy scan passed.
- Merging the base's new guards (underflow, duplicate members) folded the depth counter and the member-name tracking into ONE pass: the container stack's length *is* the depth, so the two guards share a walk rather than adding a second one.

## Checklist

- [x] Docs or release notes were updated when needed. (The nesting cause is now named in the refusal message and in all three doc locales, en + zh-tw + tr.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (Refusal messages carry only config paths and clamped JSON paths; all guards fail closed.)

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * JSON integrations can now be disabled after unrelated edits while preserving user changes.
  * Supported integrations identify when an update is needed and safely merge changes.
  * OMP integrations continue to update only their managed provider settings.

* **Bug Fixes**
  * Prevented unsafe rewrites involving invalid numbers, oversized integers, duplicate entries, or excessive nesting.
  * Preserved conflict protection for managed-entry edits and unsupported formats such as YAML, JSON5, and TOML.

* **Documentation**
  * Updated integration guidance in English, Traditional Chinese, and Turkish.


RobinBially:
Thanks — all six findings were correct. Two of them are this PR's and are fixed in `23f4349af`; the other four concern the guard introduced in the base PR and are fixed there (#1632, commit `7f2e2dfa9`, with a per-finding reply on that PR).

**Scan before `JSON.parse`** — agreed, and it sharpens the stated goal rather than just saving a pass: with the scan second, a 50k-deep document was fully materialized and then discarded, so the depth cap protected the rewrite but not the parse that precedes it. Now `parseConfig` refuses on the text and never builds the object graph. The result set is unchanged — invalid JSON still returns `PARSE_FAILED` from the `catch`.

**`installPi()`** — extracted, and it was seven copies rather than six once the new duplicate-member test landed. It follows the `installHermes()`/`installOmp()` shape already in the file.

**Depth in the refusal texts** — added, in the preflight message and in all three doc locales.

One thing worth flagging from merging the base: the new duplicate-member guard and this PR's depth counter collapsed into a **single** pass. The container stack that tracks member names per open object has a length that *is* the nesting depth, so there is one walk with one ceiling rather than two overlapping scans.

Full `bun run prepush` on this head: 11591 pass / 8 skip / 0 fail across 719 files (532s), privacy scan passed.

lidge-jun:
The code-level hardening is sound.

`src/integrations/config-io.ts:79-82` rejects a 1001st container while scanning raw JSON, and `src/integrations/config-io.ts:139-148` performs that scan before `JSON.parse`, so hostile depth is refused before the object graph is materialized. `src/integrations/serialize.ts:258-305` uses an iterative value walk, shares the same 1000-level ceiling, and turns non-finite values, `-0`, and excessive depth into `UnserializableValueError` rather than a raw `RangeError` or silent JSON normalization.

The regression coverage is meaningful: `tests/integrations-state.test.ts` pins the parse boundary at 1000/1001 and the string-bracket countercase, while `tests/integrations-serialize.test.ts:159-208` covers unsafe numeric values, structured deep-nesting refusal, path clamping, and the exact serialization boundary. I also reviewed the previous CodeRabbit findings; the scan-before-parse, positive underflow, duplicate-member, refusal-text, and Pi-fixture findings were all valid and are addressed on this head or its #1632 base. I found no remaining code-level blocker.

The release blocker is verification state, not code: both **Cross-platform CI** and **React Doctor** for exact head `23f4349afecdd694781575aeeee6ec6404645011` are `action_required` with no jobs, while the only successful checks are PR-target/label/hygiene automation. I confirmed this directly against the check-runs API for that SHA. Please approve or rerun those two workflows and require successful results on this exact SHA before merging. The reported local prepush run is useful evidence but does not replace CI.

DISPOSITION: NEEDS-CHANGE

Wibias:
Thanks for the careful hardening here. This is useful because deeply nested or otherwise non-round-trippable JSON can no longer turn an integration update into a huge allocation spike, raw `RangeError`, or silent config corruption. The shared depth ceiling and serializer checks make those cases fail closed with a structured refusal, while normal valid configs keep working as before.

Merged after Cross-platform CI and React Doctor passed. Thanks!

PR #1646, opened 2026-08-14
feat(google-antigravity): unlock gemini-3.7-flash tier

waw4303:
# PR description: feat(google-antigravity): unlock gemini-3.7-flash tier

## Summary

Two changes that make the `gemini-3.7-flash` tier reachable through the `google-antigravity` provider:

1. The Cloud Code Assist backend answers `404 NOT_FOUND` for `gemini-3.7-*` models when the request User-Agent is CLI-shaped (`antigravity/cli/…`), even with a valid OAuth token; the same request with the IDE client UA (`antigravity/ide/…`) succeeds. The request UA now mirrors the Antigravity IDE client (version pinned to the bundled language server, 2.5.5).
2. The bare `gemini-3.7-flash` id has no wire entity — only the suffixed tiers (`-low/-medium/-high`) exist upstream. The model registry now carries the full 3.7 effort ladder, so the bare picker id resolves to `gemini-3.7-flash-medium` by default (`low`→`-low`, `high`/`max`→`-high`) exactly like the existing 3.6 ladder.

- `src/adapters/client-fingerprint.ts` — `antigravityUserAgent()` emits `antigravity/ide/2.5.5 (aidev_client; os_type=windows; arch=amd64)`; `GOOGLE_ANTIGRAVITY_USER_AGENT` override still wins.
- `src/providers/antigravity-models.ts` — `gemini-3.7-flash` added to picker models, wire ids, effort map, default effort, context windows, modalities, and compatibility aliases.
- Tests updated for the new UA shape and the 3.7 suffix routing (including the bare-id default and `max` clamping).

## Verification

- Live end-to-end check against the real CCA backend (daily-cloudcode-pa, `streamGenerateContent`): `gemini-3.7-flash-medium` with the IDE UA → 200, modelVersion `gemini-3.7-flash`; same request with CLI UA → 404.
- `bun test tests/google-antigravity-wire.test.ts tests/client-fingerprint.test.ts` — 63 pass / 0 fail.
- `bun run typecheck` — pass.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Updated Antigravity identification to represent the IDE client on Windows/amd64.
  * Added Gemini Flash 3.7 tier routing with low, medium, and high effort levels.
  * Improved migration of retired Flash model identifiers while preserving supported effort settings.
  * Added safer handling for unsupported and out-of-range effort values.

* **Documentation**
  * Updated Antigravity request documentation to describe the IDE-specific User-Agent format.

* **Tests**
  * Expanded coverage for IDE fingerprinting, model migration, and effort-level routing.

### DIFF
--- a/src/adapters/client-fingerprint.ts
+++ b/src/adapters/client-fingerprint.ts
@@ -40,20 +40,24 @@ export function claudeCodeSessionId(token: string | undefined): string {
   return `${h.slice(0, 8)}-${h.slice(8, 12)}-4${h.slice(13, 16)}-${variant}${h.slice(17, 20)}-${h.slice(20, 32)}`;
 }
 
-// ── Antigravity CLI ──
-/** Pinned fallback Antigravity CLI version (real client fetches a manifest; we pin to avoid the network dependency). */
-export const ANTIGRAVITY_CLI_VERSION = "1.1.12";
-const ANTIGRAVITY_CLI_CLIENT_NAME = "aidev_client";
-const ANTIGRAVITY_CLI_PLATFORM = "darwin/arm64";
+// ── Antigravity IDE ──
+/** Pinned fallback Antigravity IDE language-server version (matches the bundled LS 2.5.5). */
+export const ANTIGRAVITY_IDE_VERSION = "2.5.5";
+const ANTIGRAVITY_IDE_CLIENT_NAME = "aidev_client";
+const ANTIGRAVITY_IDE_PLATFORM = "windows/amd64";
 /** Secondary Google API client UA the Antigravity client library reports. */
 export const ANTIGRAVITY_GOOG_API_CLIENT_UA = "google-api-nodejs-client/10.3.0";
 
 /**
- * The real Antigravity CLI User-Agent, e.g.
- * `antigravity/cli/1.1.12 (aidev_client; os_type=darwin; arch=arm64)`.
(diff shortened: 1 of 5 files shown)


iF2007:
Came across this PR and wanted to note that it appears to be solving the same problem as #1658.

Key difference:
- #1658 fixes the root cause — model discovery now follows the live CCA catalog and reads the tieredModelIds.flash index that the old parser ignored, so new models are picked up without source changes, and complete effort-tier sets are collapsed into a single base model id.
- This PR (#1646) is a more localized hardcoded workaround — it adds the full gemini-3.7-flash effort ladder directly to the model registry and changes the request User-Agent to mimic the IDE client to get around the backend's 404 for the CLI-shaped UA.

One more thing: **#1658 is already merged**. As a result, the manual hardcoding of the 3.7 models in this PR overlaps with #1658 and is already covered by it. The only part not covered is likely the User-Agent change (since #1658 does not address the CLI request path 404). If still needed, I'd suggest keeping just the UA change and dropping the redundant model hardcoding.

lidge-jun:
Cherry-picked the IDE UA switch into dev (875bb70c3). The model registry changes for gemini-3.7-flash were already on dev via #1658, so only the CLI→IDE User-Agent family switch was needed. Thanks for the live verification — the 404 on CLI UAs was a real gotcha.

PR #1652, opened 2026-08-14
feat(usage): record streamAborted on mid-stream-death attempts

kartikkabadi:
## Summary

Ports codex-router PR #139's `streamAborted` metering marker into opencodex: when an upstream stream dies after its 200 head was already committed, the persisted usage attempt must not meter as a success the client never received. The turn is recorded with the synthetic 502 terminal plus `streamAborted: true`; client cancellations keep opencodex's own 499 `client_cancel` semantics and never carry the marker.

## Changes

- `src/usage/log.ts` — optional `streamAborted?: boolean` on `PersistedUsageAttempt`, whitelisted in `normalizeUsageAttempt` so only the literal `true` marker survives the round trip. Backward compatible: ordinary and legacy rows keep their exact JSONL shape.
- `src/server/relay.ts` — set the marker on both mid-stream read-failure paths:
  - `consumeForInspection` `onReadError` (native passthrough): `onTerminal("failed", 502)` plus `streamAborted: true` on the active attempt.
  - `trackSseForRequestLog` read failure (translated responses): `incomplete`/502 terminal plus the marker, guarded so a cancel-drained pending read never pollutes the 499 row.
- `tests/usage-log.test.ts` — 4 new tests: marker persistence, backward-compatible omission (undefined/false), legacy rows stay readable, and raw JSONL round trip.
- `tests/stream-aborted-marker.test.ts` (new) — drives both relay paths end to end through `addFinalRequestLog` to the persisted JSONL row: mid-stream death meters 502 + `streamAborted: true`; client cancel meters 499 with no marker.

## Verification

- `bun test tests/usage-log.test.ts` — 34 pass (30 pre-existing + 4 new).
- `bun test tests/stream-aborted-marker.test.ts` — 4 pass (new file).
- Adjacent relay suites still green: `tests/consume-for-inspection-cancel.test.ts`, `tests/sse-inspector-bounds.test.ts` — 34 pass.
- `bunx tsc --noEmit` — clean for the files in this PR (remaining errors are peers' in-flight edits in other branches' files: `src/lib/token-estimate.ts`, `src/server/responses/empty-completion-guard.ts`).

## Coordination notes

- `src/server/request-log.ts` was not modified (TokenCapPort owns it). `addFinalRequestLog`'s existing `...attempt` spread carries the marker into the persisted row; the only shared assumption is `logCtx.activeAttempt` mutation, which relay.ts already uses for `transportPhase`/`terminalSource`.
- `tests/usage-log.test.ts` also carries a small pre-existing working-tree edit to the "persists the rate-limit-429 recovery kind" test (`durationMs` 4→1, `sendCount` 2→1, deduped `recoveryKinds`, `usageStatus` reported→unreported) that was present before this branch's work and was kept as-is per main-agent direction; it is not authored by this PR.
- Known gap (out of this slice's file ownership): the eager relay path (`relay-eager.ts` `onSynthetic` "failed", wired in `responses/core.ts`) synthesizes the same 502 terminal but does not yet set `streamAborted`. Follow-up recommended.

## Review readiness checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of interrupted upstream streams after a successful response begins.
  - These failures are now recorded as failed requests with a 502 status and a stream-aborted indicator.
  - Client cancellations continue to be reported correctly as 499 cancellations without the stream-aborted indicator.
  - Usage logs remain compatible with existing entries while preserving the new interruption details.

### DIFF
--- a/src/server/relay.ts
+++ b/src/server/relay.ts
@@ -366,6 +366,7 @@ export function trackSseForRequestLog(
 ): ReadableStream<Uint8Array> {
   const reader = body.getReader();
   let terminalReported = false;
+  let cancelled = false;
 
   const reportTerminal = (status: ResponsesTerminalStatus) => {
     if (terminalReported) return;
@@ -394,12 +395,19 @@ export function trackSseForRequestLog(
         inspector.feed(value);
         controller.enqueue(value);
       } catch (err) {
+        // The upstream read rejected: the 200 body died mid-flight. Client
+        // cancellation is the caller's separate 499 path, so a cancel-drained
+        // pending read (cancelled=true) must not carry the truncation marker.
+        if (!cancelled && !terminalReported && logCtx?.activeAttempt) {
+          logCtx.activeAttempt.streamAborted = true;
+        }
         if (!terminalReported) reportTerminal("incomplete");
         inspector.dispose();
         try { controller.error(err); } catch { /* already torn down */ }
       }
     },
     cancel(reason) {
+      cancelled = true;
       inspector.dispose();
       onCancel();
       reader.cancel(reason).catch(() => {});
@@ -1122,6 +1130,10 @@ export function consumeForInspection(
         if (logCtx) {
           logCtx.transportPhase = "mid_stream";
           logCtx.terminalSource = "synthetic";
+          // A truncated 200 body must not meter as a success the client never
+          // received; the router's equivalent turn carries 502 + streamAborted
+          // (codex-router #139).
+          if (logCtx.activeAttempt) logCtx.activeAttempt.streamAborted = true;
         }
         onTerminal("failed", 502);
       }
--- a/src/usage/log.ts
+++ b/src/usage/log.ts
@@ -36,6 +36,12 @@ export interface PersistedUsageAttempt {
   adapter: string;
   status: number;
   durationMs: number;
+  /**
+   * True only when the upstream stream died after its 200 head was committed,
+   * so the row must not meter as a success the client never received.
+   * Absent on ordinary attempts so old rows keep their exact shape.
+   */
+  streamAborted?: boolean;
   /** TTFT relative to THIS attempt's start (WP4); unset for non-streaming/tool-only. */
   firstOutputMs?: number;
   sendCount: number;
@@ -277,6 +283,8 @@ function normalizeUsageAttempt(raw: unknown): PersistedUsageAttempt | null {
     adapter: attempt.adapter,
     status: attempt.status,
     durationMs: attempt.durationMs,
+    // Absent by default; only the literal `true` marker survives the round trip.
+    ...(attempt.streamAborted === true ? { streamAborted: true } : {}),
     ...(isNonNegativeFiniteNumber(attempt.firstOutputMs)
       ? { firstOutputMs: attempt.firstOutputMs }
       : {}),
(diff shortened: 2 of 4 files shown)


kartikkabadi:
Checklist complete:\n\n- [x] Local CI green: typecheck (tsc --noEmit) clean, privacy scan passed, full suite exit 0 on a combined branch of all three ports (#1652/#1653/#1655) — the only suite failures are 18 pre-existing dev-baseline failures in untouched files (server-management-auth, lab-*-regressions, codex-shim), identical on dev.\n- [x] On latest dev (rebased onto 8b1c6208).\n- [x] Focused regression tests for this behavior: see test file(s) in the diff.\n- [x] Ready for review.

Wibias (changes requested):
Reviewed exact head `bc07d5b40efe7010b3abce00e43ad23ea257c47f`.

The basic marker encoding/persistence looks sound, and exact-head Cross-platform CI + React Doctor are green. I still see a few integration blockers before this is merge-ready:

1. **Native Responses WebSocket traffic does not persist `streamAborted` for the same mid-stream upstream reset.** The WS path calls `handleResponses(..., { recordTerminalOutcomes: false })` and owns final logging in `sendResponseToWebSocket`. On passthrough SSE this means the background branch uses `consumeForResponseLogMetadata`, which has no read-error callback that marks the active attempt. The WS pump can still finalize the failed/truncated stream as 502/incomplete, but the attempt lacks `streamAborted: true`. Please carry the transport-abort fact through the WS-owned finalization path and add an end-to-end regression.

2. **The eager relay has the same gap.** `relaySseEagerBounded` synthesizes a failed terminal for a genuine upstream read failure, and `responses/core.ts` maps it to mid-stream synthetic 502, but that path never marks `logCtx.activeAttempt.streamAborted`. This is a production path (including forced Windows rewrite traffic), so the new accounting invariant currently depends on which relay implementation was selected. Please mark the eager synthetic-failure path too and cover it.

3. **CodeRabbit's cancellation-race finding is valid.** After `trackSseForRequestLog.cancel()` sets `cancelled`, a pending `reader.read()` can still settle through the done/catch branches, which continue terminal/controller handling. The existing `logged` guard prevents this from replacing the persisted 499, but the state machine should still stop terminal work after cancellation. Add the pending-read cancellation regression CodeRabbit requested and return early from both post-read branches when cancelled.

4. **Please drop the unrelated `tests/usage-log.test.ts` fixture mutation.** The existing rate-limit recovery fixture was changed (`durationMs`, `sendCount`, duplicate recovery kinds, `usageStatus`) even though the PR description says that edit is not authored by this change. It should not ride along in this PR.

Readiness/process: the branch is currently **141 commits behind `dev`** (`dev` = `81ada7cd092d4be3b25f3013c996cd3262a2f99b`), well outside the repository's max-10-behind policy, and the PR remains draft with the readiness checklist unchecked. Please address the findings, rebase onto current `dev`, then rerun exact-head validation.

lidge-jun:
Landed via #1744 (dev merge 656376fc6). Your commit was cherry-picked with authorship preserved (4d3a776a), plus maintainer repairs: streamAborted now covers the eager/WS synthetic-failure paths, and the cancellation race in trackSseForRequestLog is fixed (client cancels are not mis-marked). The unrelated rate-limit fixture change was dropped. Full gates green. Closing as landed — thank you!

PR #1659, opened 2026-08-14
[WRONG BRANCH] feat(terminal-guard): opt-in no-tool-call continuation guard for openai-chat providers

### LINKED ISSUE #1651: [Feature]: extend the no-tool-call terminal continuation guard beyond the anthropic adapter (openai-chat routed models stop early)
### Client or integration

Codex App

### Area

Proxy and routing

### Summary

The bounded "no-tool-call terminal continuation" guard is hardwired to the `anthropic` adapter only. `openai-chat`-routed reasoning models that end a turn with a clean `stop` after announcing an action (e.g. "let me check X / found it, now extracting") — but without emitting the tool call — are treated as complete. Codex renders the turn as finished, and the user has to manually prompt "continue". This is the same class of premature-termination the anthropic guard already fixes, just on the chat adapter.

In practice GLM-5.2 (and combo children routing to it) hit this: the model narrates its plan, emits `finish_reason: stop` with no `tool_calls`, and the run stops early even though a tool call was clearly intended.

### What I expected

The terminal guard's eligibility and analysis should be extendable to specific `openai-chat` routed models known to prematurely stop, not locked to `adapterName === "anthropic"`. Either a provider/model opt-in, or applying the same `analyzeTerminalTurn` heuristic to `openai-chat` when a tool-call was announced-but-absent.

### Evidence (code path)

`src/server/responses/core.ts`:

```ts
const terminalGuardEnabled = activeAdapter.name === "anthropic" && !options.comboAttempt && !routedCompaction;
```

`src/server/responses/terminal-guard.ts`:

```ts
const analysis = options.adapterName === "anthropic"
  ? analyzeTerminalTurn(parsed, seen)
  : { decision: "pass" as const };
```

Both gates are adapter-name equality checks against `"anthropic"`. The continuation machinery (`analyzeTerminalTurn`, `buildContinuationRequest`, `maxContinuations`) is adapter-agnostic — only these two eligibility checks scope it to anthropic. A routed `openai-chat` model that exhibits the same "announce-then-stop-without-tool-call" pattern never gets the one bounded re-ask.

### Reproduction

1. Route GLM-5.2 through an `openai-chat` provider.
2. Send an executable task where the model tends to narrate intent first (e.g. "check the contents of /etc/hosts and tell me what you find") with a `read_file` tool available.
3. Occasionally the model emits a short "I'll check that for you" message, `finish_reason: stop`, and no `tool_calls`. The turn ends; no continuation is attempted.
4. The same request against the anthropic adapter would get one bounded internal continuation.

### Proposal (either is fine)

1. Add a provider/model allowlist (e.g. `provider.terminalGuardModels: string[]`) that extends `terminalGuardEnabled` and the analysis branch to those `openai-chat` routed models; or
2. Enable the guard for `openai-chat` generally when a tool was announced-but-absent, gated behind the same bounded single-continuation and short-reply constraints the anthropic path uses, so normal long answers and plan-only turns are unaffected.

### Version

### Operating system

macOS 26.5.1 (25F80)

### Provider and model

GLM-5.2 routed through an `openai-chat` OpenAI-compatible gateway, and a failover combo whose children route to it. Reproducible with any `openai-chat` reasoning model that ends turns by narrating an intended tool call without emitting it.

TooSpace:
## What

Extend the no-tool-call terminal continuation guard from the `anthropic` adapter to
`openai-chat` routed models, gated behind a new per-provider opt-in flag
`terminalContinuationGuard` (default off).

Fixes #1651.

## Why

The guard added in #394 issues one bounded internal re-ask when a model announces work
(an edit/plan) but ends the turn without emitting a tool call. It is currently wired to
the anthropic adapter only. Self-hosted OpenAI-compatible gateways (GLM / Kimi-family and
similar) routed through `openai-chat` hit the exact same premature-completion pattern and
stop mid-work, because they never reach `analyzeTerminalTurn`.

## Design / why opt-in

`analyzeTerminalTurn`'s suspicious-no-tool-stop heuristic (the `ACTIONABLE_REQUEST_RE` /
`PLAN_OR_COMPLETION_RE` / `WAITING_FOR_USER_RE` regexes) was tuned on Anthropic turns.
Its false-continue rate on other model families and non-English output is not yet
characterized. The `openai-chat` adapter is shared by many registry providers, so turning
the guard on globally there could inject unexpected continuations for existing users
(xAI, z.ai, etc.).

To keep this strictly do-no-harm, the guard is enabled only when a provider sets
`terminalContinuationGuard: true`. Anthropic behavior is unchanged; every other provider
is unchanged unless it explicitly opts in.

## Changes

- `src/types.ts`: add documented optional `terminalContinuationGuard?: boolean` to `OcxProviderConfig` (passthrough config bool, same pattern as `parallelToolCalls` / `promptCacheKey`).
- `src/server/responses/core.ts`: `terminalGuardEnabled` now also true for `openai-chat` when `route.provider.terminalContinuationGuard === true` (still excludes combo attempts and routed compaction).
- `src/server/responses/terminal-guard.ts`: `guardTerminalEventStream` runs `analyzeTerminalTurn` for `openai-chat` as well as `anthropic`; all other adapters still short-circuit to `pass`.
- `tests/terminal-guard.test.ts`: add coverage that an `openai-chat` stream gets exactly one continuation, and that an unrelated adapter (`openai-responses`) is never guarded.

## Not included

No change to `analyzeTerminalTurn`'s heuristics themselves. If maintainers later gain
confidence in the cross-family false-continue rate, flipping openai-chat to default-on
would be a separate follow-up.

## Testing

- `bun x tsc --noEmit` clean.
- `bun test` for `terminal-guard`, `terminal-guard-server`, `anthropic-tail-guard`,
  `anthropic-image-guard`, `bridge-terminal-singleness`, `responses-terminal-repair`,
  `openai-chat-hardening`, `openai-chat-eof`, `openai-chat-dangling-toolcalls`,
  `openai-chat-parallel-stream`, `parallel-tool-calls-optin`, `cl01-openai-chat-review-regressions`:
  185 pass, 0 fail (includes the 2 new cases).

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - Added an optional terminal continuation guard for supported chat providers.
  - The guard can limit repeated no-tool-call continuations and ensure clean completion events.
  - Anthropic-based responses now receive terminal guard analysis automatically.
  - OpenAI chat responses support the guard when explicitly enabled.

- **Bug Fixes**
  - Improved handling of assistant response boundaries and final completion signals.
  - Other response modes remain unaffected unless specifically supported or enabled.

### DIFF
--- a/src/server/responses/core.ts
+++ b/src/server/responses/core.ts
@@ -3578,9 +3578,13 @@ async function handleResponsesInner(
 
   cancelBodyOnAbort(upstreamResponse.body, upstream.signal);
 
-  // Anthropic-only: one bounded internal continuation re-ask for clean end_turn turns that
-  // announced an edit without emitting a tool call.
-  const terminalGuardEnabled = activeAdapter.name === "anthropic" && !options.comboAttempt && !routedCompaction;
+  // One bounded internal continuation re-ask for clean end_turn turns that announced an edit
+  // without emitting a tool call. Anthropic gets this by default; openai-chat providers opt in
+  // per-provider via `terminalContinuationGuard` (the heuristic was tuned on Anthropic turns,
+  // so it stays off for the shared openai-chat adapter unless a provider enables it).
+  const terminalGuardEnabled = (activeAdapter.name === "anthropic"
+      || (activeAdapter.name === "openai-chat" && route.provider.terminalContinuationGuard === true))
+    && !options.comboAttempt && !routedCompaction;
   /**
    * One bounded internal re-ask for Anthropic end_turn-without-tool-call turns. Replays the
    * continuation on a 429 with the same-key retry budget (hoisted per request), then falls
--- a/src/server/responses/terminal-guard.ts
+++ b/src/server/responses/terminal-guard.ts
@@ -195,7 +195,7 @@ export async function* guardTerminalEventStream(options: GuardedEventStreamOptio
     for await (const event of source) {
       if (event.type === "done") {
         terminalSeen = true;
-        const analysis = options.adapterName === "anthropic"
+        const analysis = (options.adapterName === "anthropic" || options.adapterName === "openai-chat")
           ? analyzeTerminalTurn(parsed, seen)
           : { decision: "pass" as const };
         const normalStop = event.stopReason !== "max_tokens" && event.stopReason !== "content_filter";
--- a/src/types.ts
+++ b/src/types.ts
@@ -1451,6 +1451,18 @@ export interface OcxProviderConfig {
    */
   parallelToolCalls?: boolean;
   /**
+   * Opt-in: extend the no-tool-call terminal continuation guard to this provider's
+   * `openai-chat` routed turns. The guard (originally Anthropic-only, see
+   * devlog/_fin/260706_previous-response-id-400) issues one bounded internal re-ask when a
+   * model announces work but ends the turn without emitting a tool call. Self-hosted
+   * OpenAI-compatible gateways (GLM/Kimi-family, etc.) hit the same premature-completion
+   * pattern, but the heuristic that decides a "suspicious no-tool stop" was tuned on
+   * Anthropic turns, so it stays OFF by default for the many registry providers that share
+   * the `openai-chat` adapter. Enable only for a provider whose models are known to stop
+   * mid-work; non-`openai-chat` adapters ignore this flag.
+   */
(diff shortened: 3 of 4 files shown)


TooSpace:
Reopening against `dev` per enforce-pr-target (this repo takes all contributions on dev, not main). Superseded by a dev-targeted PR.

PR #1660, opened 2026-08-14
feat(terminal-guard): opt-in no-tool-call continuation guard for openai-chat providers

### LINKED ISSUE #1651: [Feature]: extend the no-tool-call terminal continuation guard beyond the anthropic adapter (openai-chat routed models stop early)

TooSpace:
## What

Extend the no-tool-call terminal continuation guard from the `anthropic` adapter to
`openai-chat` routed models, gated behind a new per-provider opt-in flag
`terminalContinuationGuard` (default off).

Fixes #1651.

## Why

The guard added in #394 issues one bounded internal re-ask when a model announces work
(an edit/plan) but ends the turn without emitting a tool call. It is currently wired to
the anthropic adapter only. Self-hosted OpenAI-compatible gateways (GLM / Kimi-family and
similar) routed through `openai-chat` hit the exact same premature-completion pattern and
stop mid-work, because they never reach `analyzeTerminalTurn`.

## Design / why opt-in

`analyzeTerminalTurn`'s suspicious-no-tool-stop heuristic (the `ACTIONABLE_REQUEST_RE` /
`PLAN_OR_COMPLETION_RE` / `WAITING_FOR_USER_RE` regexes) was tuned on Anthropic turns.
Its false-continue rate on other model families and non-English output is not yet
characterized. The `openai-chat` adapter is shared by many registry providers, so turning
the guard on globally there could inject unexpected continuations for existing users
(xAI, z.ai, etc.).

To keep this strictly do-no-harm, the guard is enabled only when a provider sets
`terminalContinuationGuard: true`. Anthropic behavior is unchanged; every other provider
is unchanged unless it explicitly opts in.

## Changes

- `src/types.ts`: add documented optional `terminalContinuationGuard?: boolean` to `OcxProviderConfig` (passthrough config bool, same pattern as `parallelToolCalls` / `promptCacheKey`).
- `src/server/responses/core.ts`: `terminalGuardEnabled` now also true for `openai-chat` when `route.provider.terminalContinuationGuard === true` (still excludes combo attempts and routed compaction).
- `src/server/responses/terminal-guard.ts`: `guardTerminalEventStream` runs `analyzeTerminalTurn` for `openai-chat` as well as `anthropic`; all other adapters still short-circuit to `pass`.
- `tests/terminal-guard.test.ts`: add coverage that an `openai-chat` stream gets exactly one continuation, and that an unrelated adapter (`openai-responses`) is never guarded.

## Not included

No change to `analyzeTerminalTurn`'s heuristics themselves. If maintainers later gain
confidence in the cross-family false-continue rate, flipping openai-chat to default-on
would be a separate follow-up.

## Testing

- `bun test` for `terminal-guard`, `terminal-guard-server`, `anthropic-tail-guard`,
  `openai-chat-hardening`, `parallel-tool-calls-optin`, `cl01-openai-chat-review-regressions`:
  89 pass, 0 fail (includes the 2 new cases).

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - Added optional terminal continuation handling for OpenAI Chat providers.
  - Anthropic providers now support terminal stream analysis.
  - OpenAI Chat providers can enable bounded continuation behavior through a configuration setting.
- **Bug Fixes**
  - Prevented unintended continuation requests for providers where terminal guarding is not enabled.
  - Improved handling of terminal responses and tool calls across supported providers.

### DIFF
--- a/src/server/responses/core.ts
+++ b/src/server/responses/core.ts
@@ -3578,9 +3578,13 @@ async function handleResponsesInner(
 
   cancelBodyOnAbort(upstreamResponse.body, upstream.signal);
 
-  // Anthropic-only: one bounded internal continuation re-ask for clean end_turn turns that
-  // announced an edit without emitting a tool call.
-  const terminalGuardEnabled = activeAdapter.name === "anthropic" && !options.comboAttempt && !routedCompaction;
+  // One bounded internal continuation re-ask for clean end_turn turns that announced an edit
+  // without emitting a tool call. Anthropic gets this by default; openai-chat providers opt in
+  // per-provider via `terminalContinuationGuard` (the heuristic was tuned on Anthropic turns,
+  // so it stays off for the shared openai-chat adapter unless a provider enables it).
+  const terminalGuardEnabled = (activeAdapter.name === "anthropic"
+      || (activeAdapter.name === "openai-chat" && route.provider.terminalContinuationGuard === true))
+    && !options.comboAttempt && !routedCompaction;
   /**
    * One bounded internal re-ask for Anthropic end_turn-without-tool-call turns. Replays the
    * continuation on a 429 with the same-key retry budget (hoisted per request), then falls
--- a/src/server/responses/terminal-guard.ts
+++ b/src/server/responses/terminal-guard.ts
@@ -195,7 +195,7 @@ export async function* guardTerminalEventStream(options: GuardedEventStreamOptio
     for await (const event of source) {
       if (event.type === "done") {
         terminalSeen = true;
-        const analysis = options.adapterName === "anthropic"
+        const analysis = (options.adapterName === "anthropic" || options.adapterName === "openai-chat")
           ? analyzeTerminalTurn(parsed, seen)
           : { decision: "pass" as const };
         const normalStop = event.stopReason !== "max_tokens" && event.stopReason !== "content_filter";
--- a/src/types.ts
+++ b/src/types.ts
@@ -1451,6 +1451,18 @@ export interface OcxProviderConfig {
    */
   parallelToolCalls?: boolean;
   /**
+   * Opt-in: extend the no-tool-call terminal continuation guard to this provider's
+   * `openai-chat` routed turns. The guard (originally Anthropic-only, see
+   * devlog/_fin/260706_previous-response-id-400) issues one bounded internal re-ask when a
+   * model announces work but ends the turn without emitting a tool call. Self-hosted
+   * OpenAI-compatible gateways (GLM/Kimi-family, etc.) hit the same premature-completion
+   * pattern, but the heuristic that decides a "suspicious no-tool stop" was tuned on
+   * Anthropic turns, so it stays OFF by default for the many registry providers that share
+   * the `openai-chat` adapter. Enable only for a provider whose models are known to stop
+   * mid-work; non-`openai-chat` adapters ignore this flag.
+   */
+  terminalContinuationGuard?: boolean;
+  /**
    * Opt-in: forward `prompt_cache_key` to the upstream `/chat/completions` body.
    * OpenAI-specific extension; strict backends (Groq, Cerebras, etc.) reject unknown
    * fields. Default off; only enable for providers that document this parameter.
--- a/tests/terminal-guard-server.test.ts
+++ b/tests/terminal-guard-server.test.ts
@@ -37,6 +37,27 @@ const continuationTurn = [
   'event: message_stop\ndata: {"type":"message_stop"}\n\n',
 ].join("");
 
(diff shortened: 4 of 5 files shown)


TooSpace:
Addressed in 49bdc854e. Added two server-level integration cases in `tests/terminal-guard-server.test.ts` that drive `terminalContinuationGuard` through `handleResponses` -> core.ts (not just `guardTerminalEventStream`):

- an `openai-chat` provider **without** the flag makes exactly one upstream call (no continuation);
- the same provider **with** `terminalContinuationGuard: true` makes two calls (one bounded re-ask) and forwards the recovered tool call.

The pre-existing combo-attempt and routed-compaction exclusions are unchanged and still covered by the surrounding suite. Local `bun test` for the guard + openai-chat suites: 126 pass, 0 fail; `tsc --noEmit` clean.

lidge-jun:
Reviewed during a bug-PR landing pass; leaving this open as an enhancement rather than folding it into that pass.

The reasoning in the description is sound, and the opt-in design is the right call given that `analyzeTerminalTurn`'s heuristics were tuned on Anthropic turns and the false-continue rate elsewhere is uncharacterized.

That opt-in is also why it is not a bug landing: it adds a new provider capability (`terminalContinuationGuard` at src/types.ts:1453-1464) and default behavior is unchanged unless a user sets it. Nothing currently broken starts working. #1651 is likewise tracked as a feature request. The branch also conflicts with current dev and would need a rebase.

No objection to the approach — it just belongs on the enhancement track with its own review rather than in a defect sweep.

lidge-jun:
Landed via #1744 (dev merge 656376fc6). Your commits were cherry-picked with authorship preserved (f6d841b6, 3d6b6148), plus maintainer repairs: explicit-false activation coverage, combo/routed-compaction exclusion tests, and provider-option docs. Full gates green. Closing as landed — thank you!

PR #1662, opened 2026-08-14
fix(openai-chat): let opted-out providers pin parallel_tool_calls:false

### LINKED ISSUE #1650: [Bug]: parallelToolCalls=false only sends parallel_tool_calls:false for the NVIDIA NIM baseUrl; ignored for all other opted-out providers
### Client or integration

Codex App

### Area

Proxy and routing

### Summary

`provider.parallelToolCalls === false` only actually sends `parallel_tool_calls: false` on the wire when the provider's `baseUrl` is exactly `https://integrate.api.nvidia.com/v1`. Every other OpenAI-compatible provider that sets `parallelToolCalls: false` in its config silently gets no wire bit, so upstreams that default to parallel tool calls keep emitting multiple concurrent tool calls in one streamed turn.

For Kimi-family and some GLM-family gateways, the streamed multi-call turn then can't be assembled cleanly into Responses-shaped function-call events, and Codex fails the turn (previously surfaced as `upstream response contained invalid tool calls`; after the 2.14 null-tool_calls hardening, still a broken/incomplete multi-tool turn). Setting `parallelToolCalls: false` — which the config schema advertises as the way to opt out — has no effect for these providers.

### What I expected

`parallelToolCalls: false` should pin `parallel_tool_calls: false` on the request for any opted-out provider, not just the hardcoded NVIDIA NIM host. The config field reads as provider-agnostic; the enforcement is host-specific.

### Evidence (code path)

`src/adapters/openai-chat.ts`, request builder:

```ts
if (provider.parallelToolCalls === false) {
  // NIM documents the Boolean defaulting to false and kimi rejects true; pin the
  // wire bit so Codex cannot opt in via request.options. Other opted-out providers
  // omit the field so strict OpenAI-compatible hosts never see an unsupported knob.
  if (provider.baseUrl === "https://integrate.api.nvidia.com/v1") {
    body.parallel_tool_calls = false;
  }
} else if (provider.parallelToolCalls === true) {
  body.parallel_tool_calls = parsed.options.parallelToolCalls !== false;
}
```

The comment's intent ("Other opted-out providers omit the field so strict OpenAI-compatible hosts never see an unsupported knob") is reasonable as a default, but it makes `parallelToolCalls: false` a no-op for every non-NVIDIA gateway that genuinely needs the wire bit. There's no per-provider way to force it short of matching the NVIDIA URL.

### Reproduction

1. Configure an OpenAI-compatible `openai-chat` provider (e.g. a Kimi-K3 gateway) with `"parallelToolCalls": false`.
2. Send a Codex turn that naturally invites two independent tool calls (e.g. "read file A and list dir B").
3. Observe the upstream returns 2 parallel `tool_calls` chunks in the same choice; `parallel_tool_calls: false` was never sent.
4. As a control, send the exact same upstream body with `parallel_tool_calls: false` added manually — the upstream then returns a single tool call.

### Proposal (either is fine)

1. Send `body.parallel_tool_calls = false` whenever `provider.parallelToolCalls === false`, regardless of `baseUrl`; or
2. Keep the NVIDIA-only default but add an explicit opt-in (e.g. `provider.forceParallelToolCallsFalse: true`) so gateways that reject/ignore parallel calls can pin the bit without a URL match.

### Version

### Operating system

macOS 26.5.1 (25F80)

### Provider and model

OpenAI-compatible `openai-chat` routed providers (Kimi-K3, GLM-5.2 families) on a self-hosted gateway. Not NVIDIA NIM. Reproducible with any `openai-chat` provider whose upstream emits parallel tool calls by default.

TooSpace:
## What

Add an opt-in provider flag `pinParallelToolCallsFalse` so an opted-out
`openai-chat` provider can actually send `parallel_tool_calls: false` on the
wire, instead of silently omitting it.

Fixes #1650.

## Why

In `src/adapters/openai-chat.ts`, when `provider.parallelToolCalls === false` the
wire bit `parallel_tool_calls: false` is only emitted for the NVIDIA NIM
`baseUrl`. Every other opted-out provider omits the field entirely. For strict
OpenAI-compatible hosts that reject unknown fields this omit-by-default is
correct. But self-hosted gateways in the Kimi/GLM family (and vLLM-style
backends) *do* honor `parallel_tool_calls` and, without it, keep emitting
multiple concurrent tool calls in a single streamed turn -- which the Responses
translator then has to reject. `parallelToolCalls: false` looks like the opt-out
for this, but today it is a no-op for those providers.

## Design / why opt-in

The omit-by-default behavior is deliberate and covered by existing tests
(`tests/parallel-tool-calls-optin.test.ts`: a `false` provider must NOT emit the
field). Sending `parallel_tool_calls: false` unconditionally would break those
strict hosts, so the wire bit stays gated:

- NVIDIA NIM baseUrl: unchanged (still pinned).
- Any other provider: pinned only when it sets `pinParallelToolCallsFalse: true`.
- No effect unless `parallelToolCalls === false`.

Default behavior for every existing provider is unchanged.

## Changes

- `src/types.ts`: add documented optional `pinParallelToolCallsFalse?: boolean` to `OcxProviderConfig`.
- `src/adapters/openai-chat.ts`: OR the new flag into the existing NVIDIA-baseUrl condition that pins the wire bit.
- `tests/parallel-tool-calls-optin.test.ts`: cover pinned-false emits the bit, keeps emitting under a permissive request bit, and is a no-op without `parallelToolCalls:false`.

## Testing

- `bun test` for `parallel-tool-calls-optin`, `openai-chat-hardening`, `openai-chat-parallel-stream`, `chat-completions-endpoint`: 126 pass, 0 fail (includes 3 new cases). Existing omit-by-default assertions still pass.

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added provider configuration to explicitly send `parallel_tool_calls: false` when parallel tool calls are disabled.
  * Supports overriding permissive request-level settings when this option is enabled.

* **Bug Fixes**
  * Improved consistency of parallel tool-call settings across supported providers.
  * Providers without the opt-in configuration remain unchanged.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -968,8 +968,11 @@ export function createOpenAIChatAdapter(provider: OcxProviderConfig): ProviderAd
         if (provider.parallelToolCalls === false) {
           // NIM documents the Boolean defaulting to false and kimi rejects true; pin the
           // wire bit so Codex cannot opt in via request.options. Other opted-out providers
-          // omit the field so strict OpenAI-compatible hosts never see an unsupported knob.
-          if (provider.baseUrl === "https://integrate.api.nvidia.com/v1") {
+          // omit the field by default so strict OpenAI-compatible hosts never see an
+          // unsupported knob, but a self-hosted gateway that DOES honor the field and keeps
+          // emitting parallel calls without it can opt in via pinParallelToolCallsFalse.
+          if (provider.baseUrl === "https://integrate.api.nvidia.com/v1"
+              || provider.pinParallelToolCallsFalse === true) {
             body.parallel_tool_calls = false;
           }
         } else if (provider.parallelToolCalls === true) {
--- a/src/types.ts
+++ b/src/types.ts
@@ -1451,6 +1451,16 @@ export interface OcxProviderConfig {
    */
   parallelToolCalls?: boolean;
   /**
+   * Opt-in: when `parallelToolCalls` is `false`, actually send `parallel_tool_calls: false`
+   * on the `/chat/completions` wire for this provider. By default an opted-out provider only
+   * OMITS the field (strict OpenAI-compatible hosts reject unknown knobs), and the NVIDIA NIM
+   * baseUrl is the sole built-in exception that pins the wire bit. Some self-hosted gateways
+   * (Kimi/GLM-family, vLLM, etc.) do honor `parallel_tool_calls` and keep emitting concurrent
+   * tool calls unless it is present; enable this to pin the bit without hardcoding their URL.
+   * No effect unless `parallelToolCalls === false`; ignored by non-`openai-chat` adapters.
+   */
+  pinParallelToolCallsFalse?: boolean;
+  /**
    * Opt-in: forward `prompt_cache_key` to the upstream `/chat/completions` body.
    * OpenAI-specific extension; strict backends (Groq, Cerebras, etc.) reject unknown
    * fields. Default off; only enable for providers that document this parameter.
--- a/tests/parallel-tool-calls-optin.test.ts
+++ b/tests/parallel-tool-calls-optin.test.ts
@@ -42,6 +42,24 @@ describe("parallel tool calls provider opt-in (request body)", () => {
     const body = JSON.parse(adapter.buildRequest(parsedRequest({ parallelToolCalls: true })).body) as Record<string, unknown>;
     expect(body).not.toHaveProperty("parallel_tool_calls");
   });
+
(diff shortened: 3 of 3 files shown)


Wibias:
Thanks for this. The provider-level opt-in is useful because it fixes a real compatibility gap without changing the safe default for every OpenAI-compatible backend. Gateways that need `parallel_tool_calls: false` can now receive it explicitly, while stricter providers that may reject the field remain untouched. That gives us the compatibility knob without introducing a broad regression risk. 🙏

PR #1666, opened 2026-08-14
feat(catalog): add modelPickerOrder for full picker ordering beyond 5 models

### LINKED ISSUE #1649: [Feature]: Model picker order is capped at 5 routed models (subagentModels .slice(0,5)); no supported way to order a larger catalog
### Area

Proxy and routing

I run a large routed catalog (10-20+ models across several OpenAI-compatible providers plus anthropic/responses adapters) and need a stable, deterministic order for the Codex model picker across the whole catalog. I want to pin the full picker order once and have it survive catalog rebuilds (`ocx sync`, `ocx service` restart, version upgrades), the same way a user would arrange a long provider/model list and expect it to stay put.

### What prevents this today?

There is no supported way to control the model-picker order for more than 5 routed models.

`PUT /api/subagent-models` (the dashboard "featured / sub-agent models" path) hard-caps the list to 5 with `.slice(0, 5)`, and that same list is the only input the catalog builder uses to assign per-model `priority`. Every non-featured routed model is emitted with the same fallback `priority` (routed slugs all get `5`), so their relative order in the picker is undefined and reshuffles on each catalog rebuild.

The 5-slot cap conflates two separate concerns: (a) the spawn_agent model-override candidates (`MAX_MODEL_OVERRIDES_IN_SPAWN_AGENT = 5`, a real upstream limit) and (b) the visual ordering of the whole picker (no upstream limit; `priority` is just an integer sort key).

Evidence (code path):

- `src/server/management/agent-settings-routes.ts` (PUT `/api/subagent-models`): `const chosen = Array.isArray(body.models) ? body.models.filter(...).slice(0, 5) : [];` -- the 5-cap is applied here, at the only user-facing write path.
- `src/codex/convergence.ts`: `const featured = config.subagentModels ?? [];` then `orderForSubagents(enabled, featured)` -- `subagentModels` is the sole ordering input.
- `src/codex/catalog/sync.ts` (`buildObservedCatalogEntries`): featured models get `priority = featuredRank * priorityStride + selectorIndex`; every other routed model falls through to `((featured?.length ?? 0) + nativeIndex) * ... + selectorIndex`, and routed rows otherwise default to `priority = 5`. Comment at the same site notes: "Codex's models-manager sorts by `priority` ASC ... Catalog ARRAY order is discarded."

`orderForSubagents` itself has no length limit -- it happily ranks an arbitrarily long array -- but the only supported way to populate `subagentModels` truncates to 5. Writing a longer list directly into `config.json` works until the next dashboard interaction rewrites and re-truncates it.

### What should OpenCodex do?

Either provide a way to declare a full ordering for the picker (independent of the 5-slot spawn_agent candidate list), or document that ordering beyond 5 models is intentionally unsupported so integrators stop trying to drive it through `subagentModels`.

Concretely, any one of these resolves it:

1. Add a separate, uncapped ordering field (e.g. `config.modelPickerOrder: string[]`) that `orderForSubagents` consumes for picker priority, while `subagentModels` keeps its 5-slot spawn_agent semantics; or
2. Lift the `.slice(0, 5)` on `subagentModels` for ordering purposes and only apply the 5-cap when deriving spawn_agent candidates downstream; or
3. If ordering beyond 5 is intentionally out of scope, document that clearly so integrators do not build fragile catalog-rewrite workarounds.

A dedicated ordering field consumed by the catalog builder would look like:

```json
{
  "modelPickerOrder": [
    "combo/JD",
    "jd-chat/kimi-k3-joybuilder",
    "jd-responses/GPT-5.6-Sol-joybuilder",
    "jd-chat/GLM-5.2-joybuilder",
    "jd-claude/Claude-Sonnet-5-joybuilder",
    "tyler/deepseek-v4-pro"
  ]
}
```

with `subagentModels` still limited to 5 for spawn_agent candidates. Today the equivalent effect can only be forced by rewriting every model's `priority` in the generated catalog after each sync, which is not a supported interface.

A post-sync script rewrites every model's `priority` in `~/.codex/opencodex-catalog.json` and `models_cache.json` directly, reproducing the intended full order. It must be re-run after every `ocx sync`, `ocx service` restart, and version upgrade, because catalog rebuild resets all routed priorities back to the flat default. This is brittle and easy to forget: an out-of-order restart silently reverts the picker.

### Additional context

- Version: opencodex 2.14.2 (`ocx --version`); running local service confirmed via `GET /healthz`.
- Operating system: macOS 26.5.1 (25F80).
- Providers/models: multiple OpenAI-compatible routed providers (10+ routed models across chat / responses / anthropic adapters). Not provider-specific; reproducible with any catalog of more than 5 routed models.

TooSpace:
## What

Add an optional `config.modelPickerOrder: string[]` that gives a deterministic,
rebuild-stable order to the whole Codex model picker, independent of the 5-slot
`subagentModels` spawn_agent cap.

Fixes #1649.

## Why

Picker priority is assigned only from `config.subagentModels`, and the dashboard
write path (`PUT /api/subagent-models`) hard-caps that list at 5 with
`.slice(0, 5)`. In `buildCatalogEntriesFromObservedState`, featured slugs get
priority `0..N-1`; every other routed row falls through to the flat default
(`5`). So for a catalog with more than 5 routed models the relative order is
undefined and reshuffles on each rebuild (`ocx sync`, `ocx service` restart,
version upgrade). The only workaround is a post-sync script that rewrites every
model's `priority` in the generated catalog, which is brittle and must re-run
after every rebuild. Issue #1649 has the full analysis.

The issue also notes the 5-cap conflates two concerns: the real upstream
`MAX_MODEL_OVERRIDES_IN_SPAWN_AGENT = 5` limit, and the purely-visual picker
ordering (`priority` is just an integer sort key with no upstream limit).

## Design / why opt-in

`modelPickerOrder` is a separate, uncapped ordering field (option 1 in the
issue). It does NOT widen the spawn_agent candidate set; `subagentModels` keeps
its 5-slot semantics and its existing top-priority band.

- Featured rows (`subagentModels`): unchanged, still `0..N-1`.
- Rows listed in `modelPickerOrder` (and not featured): sorted in declared order,
  in a band right after the featured one.
- Rows in neither: keep their mutual order, sorted after the listed ones.
- Unset/empty `modelPickerOrder`: the helper is a no-op; every priority is
  byte-identical to before.

The `codex-catalog` golden oracle (`tests/codex-catalog-golden.test.ts`)
snapshots exact per-slug priorities and still passes unchanged, proving the
default path is untouched.

## Changes

- `src/types.ts`: add documented optional `modelPickerOrder?: string[]` on `OcxConfig`.
- `src/codex/catalog/sync.ts`: thread `modelPickerOrder` through `ObservedCatalogEntryBuildInput`; assign the non-featured band only when it is non-empty.
- `src/codex/convergence.ts`: pass `config.modelPickerOrder` into the routed-entry build so the on-disk catalog reflects it.
- `tests/codex-catalog-model-picker-order.test.ts`: unset = flat default; listed rows ordered, unlisted after; featured still wins.

No config-schema change is needed: the top-level config schema is `.passthrough()` (same as `subagentModels`).

## Testing

- `bun test` for the catalog suites (`codex-catalog-golden`, `codex-catalog`, `codex-catalog-sync-hardening`, `codex-catalog-writer`, `codex-catalog-admission`) plus the new file: 219 pass, 0 fail. Golden oracle unchanged.

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added an optional setting to control model ordering in the model picker.
  * Listed routed models appear in the configured order, followed by unlisted models.
  * Featured models retain their existing priority and behavior.
  * Native and account-qualified models remain unaffected.
  * Picker ordering is deterministic across catalog updates.
  * Changes affect display only and do not alter available agent candidates.

### DIFF
--- a/src/codex/catalog/sync.ts
+++ b/src/codex/catalog/sync.ts
@@ -67,6 +67,12 @@ import { accountBoundNativeDisplayName, CODEX_ACCOUNT_BOUND_CATALOG_KIND, truste
 
 export const MAX_SPAWN_AGENT_MODEL_OVERRIDES = 5;
 
+// Priority tier for config.modelPickerOrder rows (#1649). Sits in the same high neighborhood
+// account-selector rows already use (1_000+), i.e. well above the featured band (0..N-1) and the
+// default routed/native tiers, so reordering here only affects the picker tail and never the
+// spawn_agent candidate window (the first MAX_SPAWN_AGENT_MODEL_OVERRIDES rows by priority).
+export const PICKER_ORDER_PRIORITY_BASE = 1_000;
+
 export type SpawnAgentSurface = "v1" | "v2";
 
 export type SubagentRosterExclusionReason =
@@ -443,20 +449,26 @@ export function buildCatalogEntriesFromObservedState({
   const pickerOrder = (modelPickerOrder ?? []).filter(id => typeof id === "string" && id.length > 0);
   const pickerOrderRank = new Map(pickerOrder.map((slug, i) => [slug, i] as const));
   const pickerOrderActive = pickerOrder.length > 0;
-  // Band base sits just past the featured band so listed rows sort right after featured ones.
-  const pickerOrderBase = (featured?.length ?? 0) * priorityStride + 1;
-  let nonFeaturedFallbackSeq = 0;
+  // The picker-order band reuses the existing high priority tier (>= PICKER_ORDER_PRIORITY_BASE,
+  // same 1_000+ neighborhood that account rows already occupy). Featured rows keep the 0..N-1
+  // front band, and any default routed/native rows keep their low priorities, so the
+  // spawn_agent candidate window (the first MAX_SPAWN_AGENT_MODEL_OVERRIDES rows by ascending
+  // priority; see effectiveSubagentRoster) is populated by featured + default rows first and is
+  // NOT reordered by modelPickerOrder. modelPickerOrder only reorders rows within this high band,
+  // i.e. the picker tail below the candidate window.
   /**
-   * Priority for a non-featured routed row when modelPickerOrder is active. Listed slugs sort in
-   * declared order right after featured; unlisted rows keep their mutual order but sit after every
-   * listed one. Returns undefined when the feature is off, so the caller's original assignment
-   * (default 5 / account 1_000+) is preserved untouched.
+   * Priority for a non-featured routed row that is explicitly LISTED in modelPickerOrder. Listed
+   * slugs sort in declared order within the high picker-order tier (>= PICKER_ORDER_PRIORITY_BASE),
+   * which sits above the featured band and the default routed tier, so it reorders the picker tail
+   * without entering the spawn_agent candidate window. Returns undefined when the feature is off or
+   * the row is not listed, so those rows keep their original assignment (default 5 / account
+   * 1_000+) untouched.
    */
   const pickerOrderPriority = (slug: string, altSlug?: string): number | undefined => {
     if (!pickerOrderActive) return undefined;
     const hit = pickerOrderRank.get(slug) ?? (altSlug !== undefined ? pickerOrderRank.get(altSlug) : undefined);
-    if (hit !== undefined) return pickerOrderBase + hit * priorityStride;
-    return pickerOrderBase + (pickerOrder.length + nonFeaturedFallbackSeq++) * priorityStride;
+    if (hit === undefined) return undefined;
+    return PICKER_ORDER_PRIORITY_BASE + hit * priorityStride;
   };
   const out: RawEntry[] = [];
   const nativeEntries: RawEntry[] = [];
@@ -562,10 +574,12 @@ export function buildCatalogEntriesFromObservedState({
     }
     // Featured picks may be stored raw (legacy) or encoded — honor both.
     const rankHit = rank.get(slug) ?? rank.get(`${m.provider}/${m.id}`);
+    const pickerPriority = rankHit === undefined ? pickerOrderPriority(slug, `${m.provider}/${m.id}`) : undefined;
     if (rankHit !== undefined) e.priority = rankHit * priorityStride;
-    else if (pickerOrderActive) {
-      // #1649: modelPickerOrder assigns the full non-featured order deterministically.
-      e.priority = pickerOrderPriority(slug, `${m.provider}/${m.id}`)!;
+    else if (pickerPriority !== undefined) {
+      // #1649: rows explicitly listed in modelPickerOrder get the high picker-order tier so they
+      // reorder the picker tail without displacing default-tier spawn_agent candidates.
+      e.priority = pickerPriority;
     }
     else if (accountSelectors.length > 0) {
       // Keep the generated account rows together in Codex's priority-sorted flat picker.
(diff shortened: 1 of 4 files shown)


TooSpace:
Addressed in 8f408b92a. You were right — the first revision let a listed row enter the first `MAX_SPAWN_AGENT_MODEL_OVERRIDES` catalog rows and displace a spawn_agent candidate when fewer than five models were featured.

Fix: `modelPickerOrder` rows now go into a reserved high priority tier (`PICKER_ORDER_PRIORITY_BASE`, the same 1_000+ neighborhood account rows already occupy), and only rows explicitly listed are touched. Featured rows keep their `0..N-1` front band and default routed/native rows keep their low priority, so the spawn_agent candidate window is populated by featured + default rows and is never reordered by `modelPickerOrder`. Rows not listed keep their original priority.

Added the requested regression: with zero featured models and routed rows both listed and unlisted, the picker-order-only rows do not displace the default-tier spawn_agent candidates. The catalog golden oracle still passes unchanged (unset = byte-identical). `tsc --noEmit` clean; catalog suites 225 pass, 0 fail.

I also corrected the `modelPickerOrder` doc comment to match: it reorders the picker tail among listed rows only and does not change spawn_agent candidate membership.

TooSpace:
Addressed in 4baa12c62. You're right that the doc over-promised: only the generic routed `<provider>/<model>` rows were routed through `pickerOrderPriority`; native passthrough and account-qualified native rows were not.

I chose the second option you offered — narrow the documented contract to what is implemented — rather than widen the reorder into the native/account-qualified priority paths. Those paths have distinct priority logic (native `9` / account `selectorIndex` bands, exercised by the golden oracle), and the feature's target surface per #1649 is the long routed multi-provider catalog, not native passthrough ordering (which stays with `subagentModels`). Widening would add blast radius for no user-facing gain here.

Changes: `OcxConfig.modelPickerOrder` doc now states routed slugs only and that native/account rows are not reordered; added a regression test asserting a bare native slug in `modelPickerOrder` does not reorder its native row while a routed slug still lands in the high picker tier. Golden oracle unchanged; `tsc --noEmit` clean.

TooSpace:
Addressed in 83b283946 — you were right, and the reversed-order example was the decisive case my priority-band approach could not satisfy.

I properly decoupled the two concerns instead of tuning the band further:

- `modelPickerOrder` now rewrites only the Codex-visible display `priority`.
- Each row it moves records its natural priority in a new OpenCodex-private catalog field `opencodex_spawn_priority`.
- `effectiveSubagentRoster` ranks the spawn_agent candidate window by that natural priority when present, so the candidate SET is independent of display order. Codex ignores the unknown field (same treatment as `opencodex_catalog_kind`), so it stays invisible to Codex.

Added the exact regression you asked for: six routed rows listed in reverse `modelPickerOrder` yield the same candidate set as no ordering (`expect([...withOrder].sort()).toEqual([...baseline].sort())`). Full catalog + golden suites: 276 pass, 0 fail; `tsc --noEmit` clean; golden oracle unchanged.

TooSpace:
Superseded by #1669, which squashes the exploratory history into a single commit of the final decoupled implementation (display-only modelPickerOrder + opencodex_spawn_priority so the spawn_agent candidate set is provably unchanged). Same net diff, same tests.

PR #1669, opened 2026-08-14
feat(catalog): add modelPickerOrder to customize the Codex model-picker order

### LINKED ISSUE #1649: [Feature]: Model picker order is capped at 5 routed models (subagentModels .slice(0,5)); no supported way to order a larger catalog

TooSpace:
## What

Add an optional, display-only `config.modelPickerOrder: string[]` that lets a user
customize the order of the Codex model picker for large routed catalogs.

Fixes #1649. (Supersedes #1666, which carried the exploratory history; this is the
squashed, decoupled implementation.)

## Why

With many routed providers (10-20+ models), the picker order is driven only by the
5-slot `subagentModels` list. Every non-featured routed row is emitted at the same
flat `priority`, so the picker order is undefined and reshuffles on each catalog
rebuild (`ocx sync`, `ocx service` restart, upgrade). There is no supported way to
pin the order of more than five routed models.

## Design

`modelPickerOrder` is **display-only**. It lists routed `<provider>/<model>` slugs;
listed rows appear in array order in the picker, unlisted rows and
`subagentModels`-featured rows keep their positions. Native passthrough and
account-qualified native rows are not reordered (use `subagentModels` for those).
When unset, catalog priority is byte-identical to before — the `codex-catalog`
golden oracle is unchanged.

Display order and spawn_agent candidacy are fully **decoupled**:

- `modelPickerOrder` rewrites only the Codex-visible `priority`.
- Each moved row records its natural priority in an OpenCodex-private catalog field
  `opencodex_spawn_priority`.
- `effectiveSubagentRoster` ranks the spawn_agent candidate window by that natural
  priority, so the candidate SET is provably unchanged by any display reordering —
  even reversing every routed row. Codex ignores the unknown field (same as
  `opencodex_catalog_kind`), so it is invisible to Codex.

No config-schema change is needed: the top-level config schema is `.passthrough()`
(same as `subagentModels`).

## Changes

- `src/types.ts`: documented optional `modelPickerOrder?: string[]` on `OcxConfig`.
- `src/codex/catalog/sync.ts`: apply display order to listed routed rows; record
  `opencodex_spawn_priority`; `effectiveSubagentRoster` ranks candidates by the
  natural priority. Adds `PICKER_ORDER_PRIORITY_BASE` / `SPAWN_PRIORITY_FIELD`.
- `src/codex/convergence.ts`: pass `config.modelPickerOrder` into the routed build.
- `tests/codex-catalog-model-picker-order.test.ts`: unset = flat default; listed
  rows ordered; featured wins; bare native not reordered; and the decisive case —
  six routed rows listed in reverse order leave the candidate set unchanged.

## Testing

- `bun test` for `codex-catalog-golden`, `codex-catalog`, `codex-catalog-sync-hardening`,
  `codex-catalog-model-picker-order` (and broader catalog suites): all pass, 0 fail.
  Golden oracle unchanged.

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added an optional model picker ordering setting for routed models.
  * Configured models now appear in the picker according to the specified display order.
  * Existing featured, native, and account-qualified model ordering remains unchanged.
  * Spawn-agent prioritization and eligible model candidates are preserved independently.

* **Tests**
  * Added coverage for custom ordering, default behavior, featured-model precedence, and candidate eligibility.

### DIFF
--- a/src/codex/catalog/sync.ts
+++ b/src/codex/catalog/sync.ts
@@ -67,6 +67,18 @@ import { accountBoundNativeDisplayName, CODEX_ACCOUNT_BOUND_CATALOG_KIND, truste
 
 export const MAX_SPAWN_AGENT_MODEL_OVERRIDES = 5;
 
+// Base for config.modelPickerOrder display priorities (#1649). modelPickerOrder is a DISPLAY-ONLY
+// reordering of the Codex model picker: it rewrites a row's Codex-visible `priority` but never the
+// spawn_agent candidate window. The window is derived from SPAWN_PRIORITY_FIELD (the natural
+// priority captured before the override), so display order and spawn candidates are decoupled.
+export const PICKER_ORDER_PRIORITY_BASE = 1_000;
+
+// OpenCodex-private catalog field: the spawn_agent candidate priority a row would have WITHOUT
+// modelPickerOrder. Codex ignores unknown catalog fields (same as opencodex_catalog_kind), so this
+// is invisible to Codex; effectiveSubagentRoster reads it so a display reorder cannot change which
+// rows are spawn_agent candidates. Absent on rows modelPickerOrder did not move.
+export const SPAWN_PRIORITY_FIELD = "opencodex_spawn_priority";
+
 export type SpawnAgentSurface = "v1" | "v2";
 
 export type SubagentRosterExclusionReason =
@@ -144,10 +156,17 @@ export function effectiveSubagentRoster(
     .filter(({ entry }) => entry.visibility === "list")
     .filter(({ entry }) => surface !== "v2" || isEligibleV2SubagentEntry(entry))
     .sort((left, right) => {
-      const leftPriority = typeof left.entry.priority === "number" && Number.isFinite(left.entry.priority)
-        ? left.entry.priority : Number.MAX_SAFE_INTEGER;
-      const rightPriority = typeof right.entry.priority === "number" && Number.isFinite(right.entry.priority)
-        ? right.entry.priority : Number.MAX_SAFE_INTEGER;
+      // Spawn candidates rank by the natural priority (SPAWN_PRIORITY_FIELD when present), so a
+      // modelPickerOrder display reorder (#1649) can never change candidate membership. Rows the
+      // override did not move fall back to their Codex-visible `priority`.
+      const spawnPriorityOf = (entry: RawEntry): number => {
+        const spawn = entry[SPAWN_PRIORITY_FIELD];
+        if (typeof spawn === "number" && Number.isFinite(spawn)) return spawn;
+        return typeof entry.priority === "number" && Number.isFinite(entry.priority)
+          ? entry.priority : Number.MAX_SAFE_INTEGER;
+      };
+      const leftPriority = spawnPriorityOf(left.entry);
+      const rightPriority = spawnPriorityOf(right.entry);
       return leftPriority - rightPriority || left.index - right.index;
     })
     .slice(0, MAX_SPAWN_AGENT_MODEL_OVERRIDES);
@@ -362,6 +381,8 @@ export interface ObservedCatalogEntryBuildInput {
   readonly gptSlugs: readonly string[];
   readonly goModels: readonly CatalogModel[];
   readonly featured?: readonly string[];
+  /** Optional full picker ordering (config.modelPickerOrder); orders non-featured rows. */
+  readonly modelPickerOrder?: readonly string[];
   readonly wsEnabled: boolean;
   readonly multiAgentMode: MultiAgentMode;
   readonly exactComboSlugs: ReadonlySet<string>;
@@ -416,6 +437,7 @@ export function buildCatalogEntriesFromObservedState({
   gptSlugs,
   goModels,
   featured,
+  modelPickerOrder,
   wsEnabled,
   multiAgentMode,
   exactComboSlugs,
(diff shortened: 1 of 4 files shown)


Wibias (changes requested):
Two changes are needed before merge:

1. **Major: `modelPickerOrder` must be runtime-validated, not only typed.**

The PR relies on the top-level config schema's `.passthrough()` and therefore preserves an unknown `modelPickerOrder` value exactly as hand-edited. The runtime then does `const modelPickerOrder = config.modelPickerOrder ?? []` and later `(modelPickerOrder ?? []).filter(...)` in the catalog builder.

That means a config such as:

```json
{ "modelPickerOrder": "foo" }
```

can successfully load and then fail catalog convergence/sync with `TypeError: ...filter is not a function`. This optional display preference should fail soft rather than take down catalog generation.

Please add `modelPickerOrder` to the config schema with the same degrade-don't-reject policy used for other optional preferences, and add regressions for at least a scalar, object, malformed array, and valid string array. Normalising/trimming/deduplicating through the existing non-blank string-array helpers would also keep the persisted/runtime representation canonical.

2. **Update the model-ordering documentation for the new public config surface.**

`docs-site/src/content/docs/guides/model-ordering.md` still says `subagentModels` is the supported ordering field and that there is no general ordering field in `OcxConfig`. After this PR that is false. Please document `modelPickerOrder`, its routed `<provider>/<model>` scope, the fact that `subagentModels` remains the five-slot featured/spawn-agent list, and that native/account-qualified rows are unaffected.

Also please avoid carrying forward the claim that the current equal-priority routed order is "undefined" or "reshuffles on each rebuild" unless there is evidence for that. The existing guide describes the fallback order as deterministic provider/model alphabetical for equal-priority routed rows. The actual missing capability is user-controlled ordering beyond the five featured slots, not determinism itself.

lidge-jun:
Landed via #1744 (dev merge 656376fc6). Your commit was cherry-picked with authorship preserved (088997364), plus maintainer repairs: fail-soft runtime normalization of malformed modelPickerOrder values with regression tests, and the model-ordering guide now documents the field. Full gates green. Closing as landed — thank you!

PR #1674, opened 2026-08-14
fix(windows): avoid PowerShell WindowStyle argv under Bun

### LINKED ISSUE #1589: [Bug] Windows PowerShell `-WindowStyle Hidden` causes EACLIDENTITY under Bun
### Client or integration

OpenCodex dashboard / Codex configuration sync

### Area

Windows compatibility / identity lookup / configuration writes

### Summary

This report supersedes #1588, which was automatically closed before the required version and operating-system fields were present.

On Windows 11 x64, OpenCodex cannot resolve the effective Windows account when it launches PowerShell through Bun with the CLI arguments `-WindowStyle Hidden`.

This causes configuration writes and model synchronization to fail with:

```
CodexUserIdentityRefusal: Windows effective-account lookup failed.
```

The same behavior was observed in OpenCodex 2.12.0 and remains reproducible in 2.14.0, so this appears to be a cross-version compatibility issue rather than a regression introduced only by 2.14.0.

### Version

- OpenCodex: 2.14.0
- Bundled Bun: 1.3.14

### Operating system

- Windows 11 Pro 23H2
- OS build: 22631
- Architecture: x64

### Reproduction

Windows PowerShell executable:

`C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe`

The following Bun invocation fails with exit code 255 and empty stdout/stderr:

```ts
Bun.spawnSync(
  [
    powershell,
    "-NoLogo",
    "-NoProfile",
    "-NonInteractive",
    "-WindowStyle",
    "Hidden",
    "-Command",
    "[System.Security.Principal.WindowsIdentity]::GetCurrent().User.Value",
  ],
  {
    stdin: "ignore",
    stdout: "pipe",
    stderr: "pipe",
    timeout: 8000,
    windowsHide: true,
  },
);
```

The same invocation succeeds and returns the account SID when the PowerShell CLI pair `"-WindowStyle", "Hidden"` is removed.

Directly running the PowerShell command also succeeds. `Start-Process -WindowStyle Hidden` from inside PowerShell works as well; the failure appears specific to passing `-WindowStyle Hidden` as a PowerShell CLI argument to a Bun-spawned process.

### Actual impact

The failure affects code paths that need effective-account identity, including:

- configuration writes and synchronization
- ACL hardening
- Codex coordinator/configuration path resolution

The user-facing result is an identity/configuration error even though the Windows account and filesystem are otherwise usable.

### Suggested fix

For direct PowerShell child-process argument arrays, remove the CLI arguments:

```ts
"-WindowStyle",
"Hidden",
```

and retain Bun's:

```ts
windowsHide: true
```

This preserves hidden process creation without passing the failing PowerShell CLI option. The `Start-Process -WindowStyle Hidden` usages inside PowerShell scripts should be evaluated separately because they work in the affected environment.

Relevant call sites include `src/codex/user-identity.ts`, `src/lib/windows-user-principal.ts`, `src/config.ts`, `src/codex/app-server-processes.ts`, `src/tray/windows.ts`, and `src/update/job.ts`.

### Expected behaviour

Effective-account lookup should succeed on Windows when PowerShell is launched through Bun, or the error should identify the incompatible process argument explicitly rather than surfacing as a generic `EACLIDENTITY` / identity refusal.

No credentials or personal account identifiers are included in this report.

Ingwannu:
## Summary

- remove the redundant PowerShell CLI pair `-WindowStyle Hidden` from direct Bun/Node PowerShell child invocations used by identity, ACL principal, process inspection, config PID validation, tray, and update paths
- retain `windowsHide: true`, hidden VBS/.NET launch behavior, trusted System32 PowerShell resolution, non-interactive flags, and existing timeouts
- keep inner `Start-Process -WindowStyle Hidden` calls unchanged because the reported failure is specific to the direct PowerShell CLI argv under Bun
- document the Windows launch boundary and add regressions that prevent the incompatible argv pair from returning

Fixes #1589.

## Verification

- focused Windows/config/update suite — 280 passed, 1 Windows-only integration test skipped on Linux, 0 failed
- `bun x tsc --noEmit` — passed
- `git diff --check` — passed
- source scan confirms remaining `WindowStyle` uses are only inner `Start-Process`, .NET ProcessStartInfo, PowerShell script, or VBS window-style controls

## Platform note

The affected Bun 1.3.14 + Windows 11 execution path cannot be reproduced on this Linux host. Exact-head Windows CI and a reporter/maintainer Windows smoke test are required before merge.

## Scope

Windows bug fix and architecture documentation only. No GUI direction or unrelated text change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Windows compatibility by preventing PowerShell window-style arguments from causing failures on Windows 11.
  - Background commands, updates, tray actions, and identity lookups continue running without displaying unwanted console windows.
  - Preserved non-interactive execution and existing process-hiding behavior.

- **Documentation**
  - Clarified the recommended approach for launching hidden PowerShell processes on Windows.

- **Tests**
  - Added and updated coverage to verify hidden execution and command compatibility.


lidge-jun (approved):

The change is internally consistent: direct PowerShell child argv drops `-WindowStyle Hidden`, process-level hiding remains in place, the trusted executable/timeouts are preserved, and the regression coverage prevents the incompatible pair from returning. Cross-platform CI and React Doctor are green on this exact head.

Recommended merge order: land this before #1678 and rebase #1678 afterward, because both touch `windowsIdentityPowerShellCommand`; the localized-path fix must retain its deterministic Base64/UTF-16 output while inheriting this PR's argv contract.

lidge-jun:

PR #1676, opened 2026-08-14
fix: isolate Compatibility Lab from core runtime

Wibias:
## Summary

- add an explicit `labIntegrationEnabled` runtime gate, off unless exactly `true`
- remove Compatibility Lab modules from the ordinary request, startup, and shutdown import graph
- lazy-load compatibility evidence only for explicit policy routes
- lazy-load Lab and routing-profile management handlers only for their namespaces
- add architecture regression tests for the isolation boundary

## Root cause

CL-09 passive route-subject linkage ran in the ordinary Responses path and statically pulled Lab subject code into every server process. CL-08 automation and CL-06 compatibility assembly were also statically reachable from server startup and routing code even when an operator did not use the Lab or routing profiles.

## Behaviour

With `labIntegrationEnabled` unset or false:

- normal model requests do not load or run Lab subject construction
- server startup and shutdown do not load Lab automation
- concrete model routing does not load compatibility assembly
- normal management requests do not load Lab or routing-profile modules

Explicit `policy/...` routes and explicit `/api/lab/*` and `/api/routing-profiles/*` requests still load their dependencies on demand.

Automatic Lab startup integration is now explicit opt-in:

```json
{
  "labIntegrationEnabled": true
}
```

Explicit automation enablement through the CLI or management API persists that runtime opt-in, and explicit live automation requests initialize their production dispatch dependencies lazily.

## Validation

- added fresh-process module-graph regression tests for the core isolation boundary
- added behavioural regressions for explicit live management runs, persistent automation opt-in, and CLI live scheduler initialization
- rebased the fix onto current `dev` at `c6688c79ff58ca4f4a6502f6e6a4228124b45047`
- preserved the newer `dev` changes in `src/server/responses/core.ts` and `src/types.ts`
- repository CI runs the full suite on this PR

## Summary by CodeRabbit

- **New Features**
  - Added an optional setting to enable Compatibility Lab integration during normal server operation.
  - Automation enablement through supported commands and APIs now persists across restarts.
  - Explicit Lab commands, APIs, and policy routes remain available on demand.
- **Bug Fixes**
  - Invalid saved integration settings safely default to disabled.
- **Performance**
  - Lab components now load only when needed.
- **Reliability**
  - Improved automation shutdown and scheduler cleanup.
  - Improved runtime initialization for live management operations.

### DIFF
--- a/src/cli/lab.ts
+++ b/src/cli/lab.ts
@@ -4,7 +4,8 @@
  * Read commands use the local SQLite projection. Automation mutations and manual runs are
  * explicit operator actions; read commands never start probes or scheduler ticks.
  */
-import { getConfigDir, readConfigDiagnostics } from "../config";
+import { getConfigDir, readConfigDiagnostics, saveConfigPreservingClaudeCode } from "../config";
+import type { OcxConfig } from "../types";
 import {
   ARTIFACT_CLASSES,
   EVIDENCE_LAYERS,
@@ -87,6 +88,8 @@ type ArtifactStatus = (typeof ARTIFACT_STATUSES)[number];
 
 export interface LabCliDeps {
   configDir?: string;
+  loadConfig?: () => OcxConfig;
+  saveConfig?: (config: OcxConfig) => void;
 }
 
 class LabStateError extends RuntimeApiError {
@@ -102,6 +105,19 @@ function labErrorMessage(err: unknown): string {
   if (err instanceof InvalidCursorError) return "invalid cursor";
   if (err instanceof LabAutomationError && err.code === "invalid_cursor") return "invalid cursor";
   return "lab read failed";
+}
+
+function loadCliConfig(deps: LabCliDeps): OcxConfig {
+  return deps.loadConfig?.() ?? readConfigDiagnostics().config;
+}
+
+function persistLabIntegrationOptIn(deps: LabCliDeps): void {
+  const config = loadCliConfig(deps);
+  if (config.labIntegrationEnabled === true) return;
+  (deps.saveConfig ?? saveConfigPreservingClaudeCode)({
+    ...config,
+    labIntegrationEnabled: true,
+  });
 }
 
 function takeEnumOption<T extends string>(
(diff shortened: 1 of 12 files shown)


lidge-jun:
Closing this because the same boundary landed on `dev` in #1681, but I want to be precise about what happened here, because **you solved two things first and I have cherry-picked both.**

You opened this at 07:58Z; I was working the same defect in parallel and did not see it until after #1681 merged. That is a coordination failure on my side — I should have checked open PRs before starting, and the freeze message I was drafting would have told you I was on it.

### Where yours was better

**Management-API namespace loading.** #1681 left `management-api.ts` statically importing the Lab and routing-profile handlers, so every dashboard request still pulled ~70 `src/lab/` modules. My own audit flagged it and I deferred it to a later phase; you had already implemented it. Your `pathInManagementNamespace` + per-namespace `await import()` shape is now on `dev` via #1682, credited to this PR — including the detail that `/api/lab/automation` must be matched before the general `/api/lab` handler.

**CLI path.** `src/cli/lab.ts` was outside my scope entirely. Worth a follow-up.

**`server/index.ts` reaching 0.** You went one step further than I did. I deliberately left the composition root exempt — a root is supposed to know which optional subsystems exist, and Lab does no import-time work, so the user-facing guarantee holds either way — but yours is the stricter result and that is a defensible position.

### Why the merged version differs

**`labIntegrationEnabled` was not taken.** It gates Lab behind a new config flag, and there is no migration that derives it from an existing `automation-config.json`. An operator already running Lab automation would upgrade and find their scheduler silently stopped, with no error and no log line explaining it — `persistLabIntegrationOptIn` only fires when someone *newly* enables automation. The merged approach reads the automation config already on disk (`automation-config.json` first, `automation-policy.json` as legacy fallback), so existing installs keep working untouched.

**`require()` was not taken.** I verified it does work under Bun. But it binds an ESM package (`"type": "module"`) to the Bun runtime in a way a registration slot does not, and the same lazy effect is achievable without that coupling. For the synchronous router path specifically, `routeModelInternal` is sync and so is the subagent-fallback chain that calls `routeModel`, so #1681 used a nullable provider slot filled at activation rather than a require at the call site.

### One thing worth knowing

An independent review of my version proved, by executing the modules, that a *deferred* activation window lets a policy alias be silently dropped from a subagent fallback chain — the subagent then runs on a different model than the operator configured, with no error and no log. That is why activation in #1681 is synchronous and completes inside the same turn as `Bun.serve`. If you revisit this area, that invariant is now written into `AGENTS.md` and `devlog/_plan/260814_lab_core_decoupling/080_activation_is_synchronous.md`.

Nothing here is lost: the branch stays, the two ideas I took are attributed, and the CLI gap you covered is still open work if you want it.

Sorry for the collision. That one is on me.

PR #1678, opened 2026-08-14
fix(windows): preserve localized profile paths

### LINKED ISSUE #1573: Windows Korean user paths are decoded as UTF-8, blocking service ownership and Codex sync
## Summary

On a Korean Windows installation, OpenCodex 2.14.0 decodes some Windows command output as UTF-8 even when the output is CP949/EUC-KR. A Korean user-profile path is therefore replaced with U+FFFD characters, causing service ownership checks and Codex write coordination to fail.

This prevents `ocx sync` from injecting `model_catalog_json` and `openai_base_url`, even though the proxy and provider routing themselves are healthy.

## Environment

- OpenCodex: 2.14.0, installed globally via npm
- Codex CLI: 0.145.0
- Runtime: bundled Bun 1.3.14
- OS: Windows with Korean system locale
- User profile shape: `C:\Users\<Korean username>\...`

## Symptoms

The first failure is the scheduled-task ownership probe:

```text
Refusing to write because ownership could not be proven: the scheduled-task XML names
C:\Users\���ȭ��_��������\.opencodex\opencodex-service-launcher.vbs,
outside the expected launcher directory
C:\Users\<Korean username>\.opencodex.
```

After fixing that decoding site locally, the next failure is:

```text
Codex configuration was not written: the coordinator path could not be resolved:
CodexUserIdentityRefusal: The Windows coordinator namespace cannot be created.
```

The nested cause shows that the LocalAppData path was corrupted in the same way:

```text
EPERM: operation not permitted, mkdir 'C:\Users\���ȭ��_��������'
```

Running `ocx service repair` does not resolve either issue because the generated files are valid UTF-16LE; the corruption occurs while decoding child-process output.

## Root cause

There appear to be two affected decoding sites:

1. `src/service-manager-probe.ts` — `decodeWindowsText()`
   - UTF-16LE/BE is handled.
   - The remaining fallback uses `buffer.toString("utf8")`.
   - `schtasks /query /xml` output can be CP949 on Korean Windows.

2. `src/codex/user-identity.ts` — `powershellValue()`
   - `new TextDecoder().decode(result.stdout)` assumes UTF-8.
   - PowerShell known-folder output can be CP949 under this locale.

## Reproduction

1. Use a Windows account whose profile path contains Korean characters.
2. Install OpenCodex 2.14.0 and its background service.
3. Configure a custom provider.
4. Run:

```powershell
ocx service repair
ocx sync
```

5. Observe that synchronization is refused with the errors above.

## Validated local workaround

I added an EUC-KR fallback when UTF-8 decoding contains U+FFFD at both sites.

```ts
const utf8 = buffer.toString("utf8").replace(/^\uFEFF/, "").trim();
if (utf8.includes("\uFFFD")) {
  try {
    return new TextDecoder("euc-kr").decode(buffer).replace(/^\uFEFF/, "").trim();
  } catch {
    // Keep the original UTF-8 decoding when the runtime lacks the Windows codec.
  }
}
return utf8;
```

And for PowerShell output:

```ts
let value = new TextDecoder().decode(result.stdout).trim();
if (value.includes("\uFFFD")) {
  try {
    value = new TextDecoder("euc-kr").decode(result.stdout).trim();
  } catch {
    // Keep the original UTF-8 decoding when the runtime lacks the Windows codec.
  }
}
```

After applying both changes:

- `ocx sync` completed successfully.
- 427 custom-provider models were appended to the Codex catalog.
- `model_catalog_json` and `openai_base_url` were injected.
- A request through Codex/OpenCodex/custom provider completed successfully.

## Expected behavior

OpenCodex should preserve non-ASCII Windows paths regardless of the active Windows locale/code page. Ideally, child commands should be forced to emit a known encoding, or their output should be decoded using the detected Windows code page rather than assuming UTF-8.

I can provide the complete two-file patch if useful.

Ingwannu:
## Summary

- Decode Windows service-manager output as UTF-16 first, strict UTF-8 second, and only then a locale-bounded legacy code page for Korean or Windows-1252 locales.
- Make the PowerShell identity and known-folder lookup return UTF-16LE data inside an ASCII Base64 envelope, so redirected output no longer depends on the active console code page.
- Add byte-level regressions for Korean, Western, unsupported legacy-codepage, scheduled-task, and PowerShell path handling.

Closes #1573

## Verification

- `bun test tests/windows-text-decoding.test.ts tests/windows-popup-fix.test.ts tests/codex-service-manager-probe-hardening.test.ts` — 19 passed
- `bun run typecheck`
- `bun run privacy:scan`
- `git diff --check`
- `bun run test` — 11,666 passed, 11 skipped; one unrelated loaded-host timing watchdog in `tests/bridge-lifecycle.test.ts` fired at 4 seconds
- `bun test tests/bridge-lifecycle.test.ts` repeated three times after the full run — 15/15 passed each time; the previously timed-out stall case completed in about 1.06 seconds on every rerun

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows identity detection for non-ASCII paths and Unicode output.
  * Enhanced Windows text decoding across UTF-8, UTF-16, Korean, and Western European encodings.
  * Fixed service and scheduled-task inspection for Korean profile paths.
  * Added safer handling for malformed or unsupported text data.

* **Tests**
  * Added regression coverage for Unicode identity output, locale-specific decoding, and Windows task configuration parsing.

### DIFF
--- a/src/codex/user-identity.ts
+++ b/src/codex/user-identity.ts
@@ -53,6 +53,15 @@ function refuse(message: string, cause?: unknown): never {
 }
 
 function windowsIdentityPowerShellCommand(expression: string): string[] {
+  // PowerShell 5.1 can encode redirected native/host output with the active
+  // Windows code page. Base64 contains ASCII only, while the payload is
+  // explicitly UTF-16LE, so Korean and Western profile paths arrive unchanged.
+  const deterministicOutput = [
+    "$ErrorActionPreference = 'Stop'",
+    `$ocxValue = [string](${expression})`,
+    "$ocxBytes = [System.Text.Encoding]::Unicode.GetBytes($ocxValue)",
+    "[Console]::Out.Write([Convert]::ToBase64String($ocxBytes))",
+  ].join("; ");
   return [
     resolveTrustedWindowsPowerShellExe(),
     "-NoLogo",
@@ -61,7 +70,7 @@ function windowsIdentityPowerShellCommand(expression: string): string[] {
     "-WindowStyle",
     "Hidden",
     "-Command",
-    expression,
+    deterministicOutput,
   ];
 }
 
@@ -93,6 +102,28 @@ export function windowsIdentityPowerShellSpawnOptionsForTests(): ReturnType<
   return windowsIdentityPowerShellSpawnOptions();
 }
 
+function decodeWindowsIdentityPowerShellOutput(output: Uint8Array): string {
+  let encoded: string;
+  try {
+    encoded = new TextDecoder("utf-8", { fatal: true }).decode(output).trim();
+  } catch (cause) {
+    refuse("Windows effective-account lookup returned a malformed value.", cause);
+  }
+  if (!/^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/.test(encoded)) {
+    refuse("Windows effective-account lookup returned a malformed value.");
+  }
+  const bytes = Buffer.from(encoded, "base64");
+  if (bytes.length % 2 !== 0 || bytes.toString("base64") !== encoded) {
+    refuse("Windows effective-account lookup returned a malformed value.");
+  }
+  return bytes.toString("utf16le").trim();
+}
+
+/** Test-only decode seam for the deterministic PowerShell output contract. */
+export function decodeWindowsIdentityPowerShellOutputForTests(output: Uint8Array): string {
+  return decodeWindowsIdentityPowerShellOutput(output);
+}
+
 function powershellValue(expression: string): string {
   let command: string[];
   try {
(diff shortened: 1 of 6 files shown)


lidge-jun (changes requested):

The locale-safe decoding design is good: deterministic PowerShell Base64/UTF-16 output avoids console-codepage ambiguity, and the bounded CP949/Windows-1252 fallback is preferable to replacement-character guessing.

Blocking integration issue: this head still builds the direct identity command with `-WindowStyle Hidden`, while #1674 removes that exact argv pair because it fails under the affected Bun/Windows path. Merge #1674 first, then rebase this branch and preserve the Base64/UTF-16 command without reintroducing `-WindowStyle Hidden`.

After the rebase, rerun the Korean-profile-path/service-manager tests and an exact Windows smoke covering `ocx sync`, scheduled-task ownership, and LocalAppData identity resolution.

lidge-jun:

Issue #1668, opened 2026-08-14
[Bug] Streaming requests to opencode.ai hang with HTTP/2 SSE (200s timeout → 502/499, Codex stuck on 'thinking')


flyinsz:
## Client or integration

- [x] Codex CLI
- [x] Direct HTTP/API client (curl)

## Area

- [x] Proxy and routing
- [x] Streaming

## Summary

Requests proxied to **opencode.ai (`/zen/go/v1`)** hang forever in streaming mode (`stream: true`). The proxy waits the full 200s timeout, then reports `connection reset` / `502 upstream_server_error` / `499 client_closed_request`. On the client side (Codex CLI) the UI just shows "thinking" indefinitely.

This is **not** a proxy/DNS/network issue on the user's machine — all direct connections work fine. It is a **protocol-level incompatibility between Bun's HTTP/2 fetch and the Cloudflare edge in front of opencode.ai for SSE streaming responses**.

## Reproduction

Same endpoint, same payload, only the HTTP version differs:

```bash
# HTTP/1.1 + stream=true  → WORKS (SSE flows, ~3s)
curl --http1.1 -N https://opencode.ai/zen/go/v1/chat/completions \
  -H "Authorization: Bearer <key>" -H "Content-Type: application/json" \
  -d '{"model":"deepseek-v4-flash","messages":[{"role":"user","content":"hi"}],"max_tokens":5,"stream":true}'
# → HTTP 200, SSE chunks received

# HTTP/2 + stream=true   → HANGS (no bytes at all, times out)
curl --http2 -N https://opencode.ai/zen/go/v1/chat/completions \
  -H "Authorization: Bearer <key>" -H "Content-Type: application/json" \
  -d '{"model":"deepseek-v4-flash","messages":[{"role":"user","content":"hi"}],"max_tokens":5,"stream":true}'
# → 0 bytes until client timeout (20s+)

# Bun fetch + stream=true (what opencodex actually does) → HANGS
# Bun's fetch negotiates HTTP/2 by default with this origin.
```

Non-streaming (`stream: false`) works over both HTTP/1.1 and HTTP/2 (~2s).

`opencodex` forwards all chat completions with `stream: true` (confirmed via `ocx debug provider logs`):

```
[ocx:openai-chat:request] {"host":"opencode.ai","model":"deepseek-v4-flash","stream":true,...}
```

## Version

[...]

Issue #1580, opened 2026-08-13
Usage dashboard gradually loses token history from previous days


folochina:
### Client or integration

OpenCodex dashboard

### Area

Dashboard

### Summary

The Usage dashboard loses token usage from yesterday and earlier dates as time passes.

I expected token totals for a completed day to remain stable. Instead, after leaving the automatically started background service running and refreshing the Usage page later, the token total for yesterday decreased even though no restart, upgrade, cleanup, or configuration change occurred during the observation.

### Reproduction

1. Upgrade OpenCodex with `ocx update` to version 2.12.0.
2. Let the updater refresh and automatically start the background service on port 10100. No additional service arguments are used.
3. Keep the service running. Do not restart OpenCodex, run another upgrade, clear storage, or change configuration.
4. Open `http://localhost:10100/#usage`.
5. Select **All** providers and the **7 days** range.
6. Record the token usage for yesterday. In the first screenshot, 08-12 shows **19.1亿 tokens**.
7. Leave the service running and refresh the same Usage page later.
8. Observe that the completed day's value has decreased. In the newer screenshot, 08-12 shows **18.9亿 tokens**.

At the same time, the current day's usage increased from 20.6亿 to 20.7亿, so new requests were still being recorded. The failure is that older usage is gradually removed from the historical total.

This is time-dependent and can be observed by taking two snapshots of the same 7-day view at different times; no other user action is required between snapshots.

### Version

2.12.0

### Operating system

macOS 15.7.7 on Apple M1

### Provider and model

All providers and models shown by the dashboard; the issue does not appear provider-specific.

### Logs or error output

```shell
No visible error is shown on the Usage page. The service remains running on port 10100.
```

### Screenshots and supporting files

Both screenshots show the **08-12** token total in the same 7-day view and were captured at different times on **2026-08-13**.

| Earlier capture — 08-12: **19.1亿 tokens** | Later capture — 08-12: **18.9亿 tokens** |
|---|---|
| <img width="100%" alt="Earlier capture: 08-12 shows 19.1亿 tokens" src="https://github.com/user-attachments/assets/a505b23a-abe8-453f-bc20-70cbedfd5755" /> | <img width="100%" alt="Later capture: 08-12 shows 18.9亿 tokens" src="https://github.com/user-attachments/assets/bf5062b8-fbac-45c8-97c7-d656b7699329" /> |

[...]

PR #1693, opened 2026-08-14
docs(devlog): plan the usage lifecycle and memory-defense roadmap

lidge-jun:
## Summary

Layer 0 of the usage-lifecycle and memory-defense stack. Docs only — no source file changes.

Two problems get one dependency-ordered plan:

**Usage storage is asymmetric.** `appendUsageEntry` writes an unbounded `usage.jsonl`, while the management reader parses only the newest 64 MiB. A nominal `30d` aggregate therefore drops every request older than that tail (#1497, #1580). `060`–`080` plan rotated JSONL segments, a rebuildable SQLite projection, and an API that reads the projection instead of re-parsing a tail. The JSONL segments stay canonical; SQLite is derived and can be deleted and rebuilt.

**Memory defense before Bun 1.4.** `010` and `020` split the two independent theses out of the #1412 draft — refuse oversized input before upstream I/O, and stop prepending stored history to a request that already carries it. `030` adds per-provider `upstreamHttpVersion` and `responseDelivery` so a provider-specific transport failure (#1668) is fixable without touching global `streamMode`. `040` connects the warn-only watchdog in `src/server/memory-watchdog.ts` to the drain-and-restart already in `src/server/lifecycle.ts`. `050` extends the nesting ceiling already landed in `serialize.ts` to YAML and TOML, whose recursive writer has no equivalent bound (#1635).

Each decade doc is written to diff-level precision — exact paths, insertion points, and named test cases — so its implementation cycle starts from an executable plan rather than an outline.

**Stack** (merge bottom-up):

| # | Layer | Branch |
|---|-------|--------|
| 0 | roadmap docs ← you are here | `codex/usage-memory-roadmap-docs` |
| 1 | M0-1 input admission | `codex/m0-1-input-admission` |
| 2 | M0-2 continuation dedup | `codex/m0-2-continuation-dedup` |
| 3 | M0-3 provider delivery policy | `codex/m0-3-provider-delivery` |
| 4 | M0-4 memory recovery policy | `codex/m0-4-memory-recovery` |
| 5 | M0-5 YAML/TOML depth cap | `codex/m0-5-nonjson-depth-cap` |
| 6 | U1 segmented usage writer | `codex/u1-segmented-writer` |
| 7 | U2 SQLite projector | `codex/u2-sqlite-projector` |
| 8 | U3 projection-backed API | `codex/u3-projection-api` |

Layer 0 carries no source change, so it is mergeable on its own and cannot block the layers above it. Layers above are opened as they are built.

## Verification

- `git diff --stat origin/dev...HEAD` — 9 files, all under `devlog/_plan/260814_usage_memory_roadmap/`, no `src/` change
- Every source path cited in the decade docs was verified to exist at the planned line ranges before writing
- No build, typecheck, or test gate reads `devlog/`, so no code gate applies to this layer

## Checklist

### DIFF
--- a/devlog/_plan/260814_usage_memory_roadmap/000_plan.md
+++ b/devlog/_plan/260814_usage_memory_roadmap/000_plan.md
@@ -1,0 +1,110 @@
+---
+title: Usage & Memory Roadmap — P0 through P3
+date: 2026-08-14
+class: C4
+prior-art: "#1008 (closed/unmerged), #1412 (open draft), #1367 (open), #1635 (open), #820 (open), #1217 (open)"
+---
+
+# 000 — Objective
+
+OpenCodex has two independent structural problems:
+
+1. **Usage data lifecycle** — a single unbounded `usage.jsonl` that grows forever,
+   dashboard reads only the newest 64 MiB tail, and 30d/all-time aggregates silently
+   lose data beyond that window (#1497, #1580).
+
+2. **Memory defense before Bun 1.4** — oversized input, continuation compounding,
+   provider transport failures, and no process-wide memory coordination mean a single
+   bad request or slow client can push RSS to OOM.
+
+This unit implements P0 (memory defense, 5 PRs) and P1 (usage storage, 3 PRs) as a
+stacked PR chain. P2 (retention + UI) and P3 (stream coordinator) are documented but
+deferred.
+
+## Constraints
+
+- CWD: worktree on `dev` at `/Users/jun/.codex/worktrees/8d76/opencodex`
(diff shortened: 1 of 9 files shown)


lidge-jun:
Landed on dev as da963ef34 (cherry-picked from c96810c12, with trailing-whitespace and EOF-newline hygiene applied). The nine usage/memory roadmap documents are now tracked in devlog/_plan/260814_usage_memory_roadmap/. Closing in favour of the direct landing; the stacked children #1696 and #1698 landed on top of it. 다음 배포에 포함됩니다.

PR #1696, opened 2026-08-14
fix(responses): refuse input that cannot fit the model context window

lidge-jun:
## Summary

Layer 1 of the usage/memory stack. Refuses a turn whose estimated input cannot plausibly fit the model's context window, before any upstream I/O. Design doc: `devlog/_plan/260814_usage_memory_roadmap/010_m0_1_input_admission.md` (layer 0, #1693).

#1412 reported ~127k of real context compounding to 1.3M–1.6M tokens and crashing the proxy. A turn that large is rejected upstream anyway, so the round trip buys nothing — it spends auth resolution, host-circuit budget, and bandwidth to arrive at a worse error than we can produce locally.

The gate sits in `handleResponsesInner` after `applyFinalRouteRequestNormalization` (so it measures what will actually be sent, including replay expansion) and before the `preAuthHostKey` circuit block (so an oversized turn costs neither auth nor circuit budget).

Three deliberate non-behaviors, each of which was a review blocker:

**It does not refuse compaction turns.** Codex sends `compaction_trigger` *because* context is full (`src/responses/parser.ts:355`), and routed `/v1/responses/compact` re-enters the same handler with that trigger appended (`src/server/responses/compact.ts:659`). 413-ing those would tell the client to compact and then deny the compaction — a deadlock against the very limit the error reports.

**It does not re-derive the context window.** `route.provider` is already the `routedProviderConfig` output, which refuses registry merging when the transport does not match (`src/router.ts:251`), so a user provider that merely shares a built-in name keeps its own limits. Native models fall back to `nativeOpenAiContextWindow` — static maps only. That fallback is load-bearing: the canonical `openai` registry entry declares no context fields, so without it the gate would be inert on the default Codex route. Reading the Codex catalog was rejected because it costs `existsSync` + `readFileSync` + `statSync` even on a cache hit, and this is the request path.

**It does not treat the estimate as exact.** `estimateTokens` is a char-ratio heuristic whose CJK sampling aliases: `cjkRatio` samples every `stride`-th character, so a payload of 62-char records each starting with one Hangul character samples as 100% CJK while being 1.6% CJK, firing the 2.5-chars/token clamp instead of 4.0.

```
{"length":126046,"stride":62,"sampledRatio":1,"trueRatio":0.0161,"overshoot":1.6}
```

Since 4.0 / 2.5 = 1.6 is that branch's maximum divergence, `ADMISSION_TOLERANCE` is 2.5 — above the demonstrated overshoot, while #1412's 10x still clears it fourfold. A regression test pins that exact payload as admissible. Repairing `cjkRatio` is deliberately left alone: `estimateTokens` also feeds usage accounting and auto-compact (`src/server/chat-completions.ts:140`), so it needs its own change and its own tests.

## Verification

- `bun test tests/input-admission.test.ts` — 19 pass, 0 fail
- `bun run typecheck` — clean
- `bun test tests/core-lab-boundary.test.ts` — still green; the new `core.ts` import pulls no Lab module
- The estimator overshoot was reproduced locally on Bun 1.3.14 before choosing the tolerance, not taken on faith

Notable test cases: a custom provider named `openai` does **not** inherit native limits; a routed `provider/model` id does not take the native fallback; a `_compactionRequest` turn is admitted at any size; ceiling resolution performs zero `fs` calls; the aliased CJK payload stays admitted.

## Stack

Depends on #1693 (layer 0, docs). Review this PR's diff only.

| # | Layer | PR |
|---|-------|-----|
| 1 | M0-1 input admission ← you are here | this |
| 0 | roadmap docs | #1693 |

## Checklist

### DIFF
--- a/devlog/_plan/260814_usage_memory_roadmap/010_m0_1_input_admission.md
+++ b/devlog/_plan/260814_usage_memory_roadmap/010_m0_1_input_admission.md
@@ -9,105 +9,326 @@ closes: "(split from #1412)"
 
 # 010 — M0-1: Model-aware input admission gate
 
+## Audit history
+
+**P stale check (base `origin/dev` @ v2.17.0)** replaced three WP0 assumptions: reuse
+`estimateTokens` (`src/lib/token-estimate.ts`) instead of a `chars / 4` heuristic;
+measure `OcxParsedRequest.context`, not `input[]`; call `formatErrorResponse` with 3 args
+(its 4th is `{code, retryAfter}`, so a metadata object would have been silently dropped —
+`src/bridge.ts:1825`).
+
+**A round 1 — FAIL, 5 blockers. All folded.**
+
+| # | Blocker | Disposition |
+|---|---------|-------------|
+| 1 | Gate would 413 compaction-recovery turns, deadlocking the client | Folded — full bypass; reviewer confirmed both paths in round 2 |
+| 2 | A heuristic estimate cannot justify "could not have succeeded upstream" | Folded — tolerance margin; claim removed |
+| 3 | `candidateCapabilityEvidence` merges registry data even for same-named custom providers | Folded — read `route.provider` |
+| 4 | Merged `modelMaxInputTokens` lives on `route.provider`, not raw config | Folded — read `route.provider` |
+| 5 | `candidateCapabilityEvidence` does sync `existsSync`/`readFileSync`/`statSync` per call | Folded — no filesystem access on the request path |
+
+**A round 3 — GO-WITH-FIXES, 1 blocker, folded below. Audit loop exits here** (AUDIT-LOOP-01:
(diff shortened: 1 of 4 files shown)


lidge-jun:
Landed on dev as babf254b1 (cherry-picked from 0298177c9). Verified that dev previously had context-window *accounting* but no pre-dispatch admission check: route normalization at core.ts:1840 went straight to circuit/auth admission with no size decision in between. The new src/server/responses/input-admission.ts resolves the ceiling and fails open when no ceiling is known. Verification on the landed commit: bun run typecheck clean, 137 pass / 0 fail across input-admission, continuation-dedup and responses-state, 222 pass / 0 fail across the widened regression slice. Closing in favour of the direct landing. 다음 배포에 포함됩니다.

PR #1698, opened 2026-08-14
fix(responses): stop previous_response_id replay from compounding history

lidge-jun:
## Summary

Layer 2 of the usage/memory stack. Stops `previous_response_id` replay from compounding conversation history. Design doc: `devlog/_plan/260814_usage_memory_roadmap/020_m0_2_continuation_dedup.md`.

`expandPreviousResponseInput` concatenated stored items in front of whatever the client sent, unconditionally (`src/responses/state.ts`). When the client already carries that history — stateless providers replaying full context alongside a `previous_response_id` — the turn doubles, the doubled turn is stored, and the next one triples. #1412 watched 127k of real context reach 1.3M tokens that way.

**The hard part is not equality, it is occurrence.** A client may legitimately repeat itself: stored history of one message `"repeat"` and a genuine delta that also begins with `"repeat"` produce an identical run, and skipping there would silently delete a real turn. Position cannot settle it either — stored state flattens request input and provider output into one array, so an id-less assistant message sits in the output region without the provider having authored anything identifiable.

So `rememberResponseState` records where `response.output` begins, and a skip requires three things:

1. the matched run covers the whole stored entry,
2. it reaches that boundary, and
3. some matched item past it carries a provider-issued `id`/`call_id`.

A client echoing provider output *with the provider's id* is replaying that exact occurrence — which is what we want to detect. A client echoing itself cannot manufacture one. Entries whose output carries no id never skip.

**Bounded comparison.** Canonicalization counts bytes during the walk rather than serialize-then-measure, because a tool result can be megabytes and this is the request path. The cap applies to every item: an id is extra evidence, never a substitute for content equality, so an over-cap tool item is non-comparable exactly like an over-cap message. Any non-comparable item aborts the whole check — skipping just that item could align two different occurrences.

**`previous_response_id` is deliberately preserved.** Kiro and Cursor recover their conversation ids from it, so stripping it would start a new upstream conversation in order to fix a memory bug. The replay prefix length is still recorded, because the boundary is real whoever supplied the history: without it the parser re-acknowledges historical compaction markers and guidance gets injected twice.

The anchor is threaded through resident entries, all four spill writes, the spill payload, materialization, and snapshot load, where a malformed value degrades to never-skip rather than to a bad index. **Spill compatibility is forward-only**: `validPayload` is a strict key allowlist, so rolling back across this commit invalidates spilled entries, degrading to the already-handled replay-miss path.

Two provenance contracts in `src/types.ts` said "the proxy expanded"; they now say the history is present however it arrived, which is what every consumer actually reads them for.

**Known gap (FU-2, recorded in the doc):** sessions where the proxy injected guidance into stored history, or repaired ids after recording, do not match and expand as before. M0-1 does not bound that — admission runs after expansion and parsing and fails open on unknown ceilings — so it stays real remaining work rather than something already covered.

## Verification

- `bun test tests/continuation-dedup.test.ts` — 16 pass, 0 fail
- `bun test tests/responses-state.test.ts tests/memory-watchdog.test.ts tests/continuation-dedup.test.ts` — 130 pass, 0 fail
- `bun run typecheck` — clean

Two existing byte-accounting tests mirror the measured envelope by hand and were updated for the new field; they still assert exact sizes.

Notable cases: the one-message repeat does **not** skip; id-less provider output does **not** skip; a spilled entry keeps its anchor and still deduplicates (a real bug caught during implementation — `writeResponseSpillDurably` rebuilds the payload field by field and was dropping it); an over-cap or deeply nested item is non-comparable and expands.

## Stack

Depends on #1696 → #1693. Review this PR's diff only.

| # | Layer | PR |
|---|-------|-----|
| 2 | M0-2 continuation dedup ← you are here | this |
| 1 | M0-1 input admission | #1696 |
| 0 | roadmap docs | #1693 |

## Checklist

### DIFF
--- a/devlog/_plan/260814_usage_memory_roadmap/020_m0_2_continuation_dedup.md
+++ b/devlog/_plan/260814_usage_memory_roadmap/020_m0_2_continuation_dedup.md
@@ -9,93 +9,461 @@ closes: "(split from #1412)"
 
 # 020 — M0-2: Stop compounding replayed history
 
+## Audit history
+
+**P stale check.** WP0 named the wrong mechanism. Compounding is the unconditional concat at
+`src/responses/state.ts:895`, not a core.ts prepend. `_providerContinuation` carries provider
+scalars, not history, so the draft plan to clear it would have broken Cursor while fixing
+nothing.
+
+**A round 1 — FAIL, 5 blockers.** All verified against the tree.
+
+| # | Blocker | Disposition |
+|---|---------|-------------|
+| 1 | Stripping the id destroys Kiro/Cursor provider continuity for that turn | Folded — keep the id, add explicit disposition |
+| 2 | Zero prefix length re-acknowledges historical compaction markers and duplicates guidance | Folded — record the recognized prefix |
+| 3 | Strict raw prefix match rarely fires: the proxy mutates stored input | **Thesis narrowed** — see below |
+| 4 | A bounded serialization SAMPLE can collide, causing a false skip | Folded — hard cap + "not comparable" |
+| 5 | A new metric field breaks two privacy-pinned tests | Folded — internal counter, not `ResponseStateMetrics` |
+
+The reviewer also refuted one bug I suspected in my own plan: stripping the id would NOT have
+mis-set `_previousResponseInputExpanded`, because `core.ts:1521` requires both
+`body !== originalBody` AND a surviving string id. Good to know, but blocker 1 removes the
+strip anyway.
+
+**A round 3 — FAIL, 4 blockers.** Third consecutive FAIL, so under LOOP-REPAIR-01 this stops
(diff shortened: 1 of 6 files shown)


lidge-jun:
Landed on dev as d30ab74af (cherry-picked from 33e2b70bd). Confirmed the defect on dev before landing: src/responses/state.ts:895 unconditionally built input as [...materialized.state.items, ...inputItems(request.input)] with no overlap comparison, so client-carried history could be prepended again on every continuation. The landed change adds guarded overlap detection requiring a full stored-prefix match, a provider-output boundary, and provider-issued identity before skipping the prepend. Verification on the landed commit: bun run typecheck clean, 222 pass / 0 fail across the regression slice. Closing in favour of the direct landing. 다음 배포에 포함됩니다.

PR #1701, opened 2026-08-14
fix(cli): default IS_SANDBOX=1 for root ocx claude launches (#1688)

### LINKED ISSUE #1688: [Bug]: ocx claude --dangerously-skip-permissions aborts as root (uid==0) because IS_SANDBOX is never set
### Summary

`ocx claude --dangerously-skip-permissions` aborts immediately when running as root (`uid == 0`), which is the default on most VPS deployments. Claude Code 2.1.205+ embeds a safety guard that exits with `--dangerously-skip-permissions cannot be used with root/sudo privileges for security reasons` unless it believes it runs inside a sandbox (`IS_SANDBOX=1`) or a bubblewrap wrapper is present. The opencodex proxy launches the session, so it should mark the environment accordingly.

### Reproduction

```bash
# As root (typical VPS):
ocx claude --dangerously-skip-permissions --print "ok"
```

Expected: the Claude session starts and prints `ok`.
Actual: immediate fatal exit with the root-guard message.

### Logs and screenshots

```
--dangerously-skip-permissions cannot be used with root/sudo privileges for security reasons
```

### Area

CLI

### Version

2.14.2 (also on current `main` — `IS_SANDBOX` does not appear anywhere in `src/`)

### OS

Ubuntu 24.04 (root user)

### Config shape

N/A (CLI invocation only)

### Root cause

`src/cli/claude.ts` `buildClaudeEnv` does not set `IS_SANDBOX`. The guard is in the official Claude Code binary, not in opencodex — but since the proxy owns the launched session (a controlled, proxy-launched environment), it must signal `IS_SANDBOX=1` so the binary allows the bypass. Without it, every root-hosted deployment of `ocx claude` with permissions bypass is unusable.

### Proposed fix

In `buildClaudeEnv`, before `return env;`:

```diff
   for (const [name, value] of Object.entries(effectiveModelEnv(config.claudeCode, contextWindows, auto))) {
     setDefault(name, value);
   }
+  // Root bypass guard (Claude Code 2.1.205+): the CLI refuses --dangerously-skip-permissions
+  // when uid==0 unless it believes it runs inside a sandbox. Servers are commonly root, so
+  // treat the proxy-launched session as one — opt out by exporting IS_SANDBOX=0 yourself.
+  if (typeof process.getuid === "function" && process.getuid() === 0) {
+    setDefault("IS_SANDBOX", "1");
+  }
   return env;
```

`setDefault` semantics already guarantee the user can override (`IS_SANDBOX=0` exported by the user wins), so this is an opt-out-by-env design, not a forced change.

### Alternatives considered

- Document "run as non-root": impractical on VPS setups where the service owns root files, and Claude Code already provides the sandbox escape hatch (`IS_SANDBOX`) for exactly this case.
- Require `CLAUDE_CODE_BUBBLEWRAP`: heavier, not portable; `IS_SANDBOX` is the documented, minimal signal.

### Additional context

The same pattern is used by other proxy-launcher tools: the session is a controlled environment created by the parent process, so marking it as sandboxed for the child's safety guard is accurate and safe. The user opt-out (`IS_SANDBOX=0`) preserves the strict behavior for anyone who wants it.

Test matrix:
- root + `--dangerously-skip-permissions` → session starts (200 / prints)
- root + exported `IS_SANDBOX=0` → original guard behavior preserved (user opted out)
- non-root → unchanged (no `IS_SANDBOX` injected)

ardakrt:
## Summary
- Sets \`IS_SANDBOX="1"\` by default in \`buildClaudeEnv\` when \`ocx claude\` runs with root privileges (\`process.getuid() === 0\`).
- Fixes #1688 where Claude Code 2.1.205+ refuses \`--dangerously-skip-permissions\` under root/sudo environments (typical VPS deployments) with \`--dangerously-skip-permissions cannot be used with root/sudo privileges for security reasons\` unless \`IS_SANDBOX\` is set.
- Respects explicit user overrides: \`setDefault\` semantics ensure that if a user exports \`IS_SANDBOX=0\`, that value is preserved.
- Adds test seam \`getuid\` to \`ClaudeEnvDeps\` for cross-platform deterministic testability.
- Adds unit tests in \`tests/claude-cli.test.ts\` verifying root defaults to \`IS_SANDBOX=1\`, non-root leaves it unset, and explicit user exports are honored.

## Test plan
- Run \`bun test tests/claude-cli.test.ts\` (all 22 tests passed).
- Run \`bun test tests/claude-inbound.test.ts tests/claude-auth-detect.test.ts tests/claude-auth-mode.test.ts tests/claude-desktop-cli.test.ts\` (all 95 tests passed).
- Run \`bun run typecheck\` (0 errors).
- Run \`bun run privacy:scan\` (passed).

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Claude Code launches now automatically enable sandbox mode when run with root privileges.
  * Users can explicitly override this behavior by setting the sandbox option.

* **Bug Fixes**
  * Improved permission-bypass behavior for root-launched Claude Code sessions.

### DIFF
--- a/src/cli/claude.ts
+++ b/src/cli/claude.ts
@@ -33,6 +33,8 @@ export type ClaudeEnvDeps = {
   authDetect?: Omit<Partial<AuthDetectDeps>, "env" | "ownTokens">;
   /** Test seam; production uses the authenticated Node-launcher context. */
   preBunAnthropicSlots?: readonly AnthropicParentEnvSlot[] | null;
+  /** Test seam for root detection; production reads process.getuid(). */
+  getuid?: (() => number) | null;
 };
 
 function isClaudeLoopbackHostname(hostname: string): boolean {
@@ -237,6 +239,15 @@ export function buildClaudeEnv(
   // authoritative >=1M window (Claude Code then accounts 1M, compaction preserved).
   for (const [name, value] of Object.entries(effectiveModelEnv(config.claudeCode, contextWindows, auto))) {
     setDefault(name, value);
+  }
+  // Root bypass guard (Claude Code 2.1.205+): the CLI refuses --dangerously-skip-permissions
+  // when uid==0 unless it believes it runs inside a sandbox. Servers are commonly root, so
+  // treat the proxy-launched session as one — opt out by exporting IS_SANDBOX=0 yourself.
+  const getuidFn = deps.getuid !== undefined
+    ? deps.getuid
+    : (typeof process.getuid === "function" ? () => process.getuid!() : null);
+  if (getuidFn && getuidFn() === 0) {
+    setDefault("IS_SANDBOX", "1");
   }
   return env;
 }
--- a/tests/claude-cli.test.ts
+++ b/tests/claude-cli.test.ts
@@ -228,6 +228,28 @@ describe("ocx claude env assembly", () => {
     expect(env.ANTHROPIC_SMALL_FAST_MODEL).toBeUndefined();
   });
 
+  test("root launch (uid==0) defaults IS_SANDBOX=1 for Claude Code 2.1.205+ permission bypass guard (#1688)", () => {
+    const envRoot = buildClaudeEnv(cfg({ claudeCode: {} }), 10100, {}, {}, {
+      getuid: () => 0,
+      ...AUTH_PRESENT,
+    });
+    expect(envRoot.IS_SANDBOX).toBe("1");
+
+    const envNonRoot = buildClaudeEnv(cfg({ claudeCode: {} }), 10100, {}, {}, {
+      getuid: () => 1000,
+      ...AUTH_PRESENT,
+    });
+    expect(envNonRoot.IS_SANDBOX).toBeUndefined();
+
+    const envUserOptOut = buildClaudeEnv(cfg({ claudeCode: {} }), 10100, {
+      IS_SANDBOX: "0",
+    }, {}, {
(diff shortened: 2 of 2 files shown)


Wibias:
Closing this PR because the proposed default crosses a security boundary rather than fixing the underlying problem.

`uid == 0` is not evidence that the Claude process is sandboxed, and being launched by OpenCodex does not create a sandbox. The current `ocx claude` path starts Claude with the assembled environment and inherited stdio, but does not establish independent OS-level isolation before launch. Automatically setting `IS_SANDBOX=1` for every root launch would therefore tell Claude Code that a sandbox exists when OpenCodex has not actually created one.

That matters especially with `--dangerously-skip-permissions`: the change can suppress Claude Code's root safety guard while the child still has unrestricted root privileges on the host. Making `IS_SANDBOX=0` an opt-out does not fix that default. An unsafe bypass must not become the implicit behaviour for users who do not know OpenCodex injected the flag.

The safe fix is to avoid manufacturing `IS_SANDBOX=1` from UID alone. I posted the preferred direction on #1688: if OpenCodex wants to support this flow, create a real isolated launch environment first, set `IS_SANDBOX=1` only after that isolation is positively established, and fail closed if sandbox creation fails.

Thanks for the contribution. The reported workflow problem is real, but this implementation is not safe to merge as-is.

PR #1703, opened 2026-08-14
fix(claude): preserve Auto Mode classifier provider affinity and support classifierModel (#1697)

### LINKED ISSUE #1697: [Provider compatibility] Claude Code Auto Mode classifier loses provider affinity and routes bare claude-opus-5 to an incompatible default provider
## Client or integration

Claude Code

OpenCodex multi-provider routing for Claude Code internal Auto Mode permission-classifier requests. Reproduced with:

- an explicitly selected Claude-compatible relay for the main model; and
- DeepSeek configured as the global `defaultProvider`.

This is not an upstream DeepSeek defect: DeepSeek correctly rejects the Claude model slug it should never have received.

## OpenCodex version

`2.11.1`

Environment:

- Claude Code `2.1.228`
- macOS `26.5.2`
- OpenCodex Claude proxy mode
- Claude Code Auto permission mode
- inbound endpoint: `POST /v1/messages`

## Endpoint or capability

`POST /v1/messages` — Claude Code Auto Mode background permission classifier used before Bash, Agent/subagent, Workflow, and other tool actions.

## Current behaviour

A Claude Code session can explicitly run its main model through a routed provider, for example:

```text
RelayA/claude-fable-5
```

The main-model requests resolve to `RelayA` and return HTTP 200.

Before Claude Code executes a tool in Auto Mode, it sends a separate internal classifier request whose model is the bare slug:

```text
claude-opus-5
```

That side request does not preserve the main session's provider namespace. With no explicit `claudeCode.modelMap` entry, OpenCodex passes the bare slug through and `routeModel()` sends it to the global `defaultProvider`.

If the global default is DeepSeek, the observed route is:

```text
requestedModel: claude-opus-5
routeKind: default-provider
selected.provider: deepseek
selected.model: claude-opus-5
selected.reason: default-provider
```

DeepSeek then correctly rejects the unsupported model with HTTP 400. Claude Code converts that failure into the misleading user-facing message:

```text
claude-opus-5 is temporarily unavailable, so auto mode cannot
determine the safety of Bash/Agent/Workflow right now.
```

Consequences:

- the explicitly routed main model continues answering normally;
- Bash, Agent/subagent, Workflow, and other classified actions are blocked;
- retries repeat the same incorrect route;
- users can easily misdiagnose this as an Anthropic/Claude outage, a broken script, or a subagent configuration problem;
- changing the featured subagent roster does not fix it because the permission classifier is a separate internal request path.

On this machine, the same deterministic mismatch produced **71 HTTP 400 requests** between 2026-08-07 and 2026-08-14. It affected multiple Claude Code conversations and multiple tool types.

The current static workaround is:

```text
claudeCode.modelMap["claude-opus-5"] = "RelayA/claude-opus-5"
```

However, this is not sufficient as the default integration behaviour:

1. It is not created automatically when the user selects `RelayA/claude-fable-5`.
2. It becomes stale when the user changes the main provider in Claude Code's model picker.
3. It has no health-aware or configured fallback.
4. The GUI does not make the hidden classifier dependency and its resolved provider obvious.
5. A static cross-provider map can send command/context metadata to a provider different from the main provider without making that boundary visible.

The relevant current flow appears to be:

- `src/claude/inbound.ts::resolveInboundModel()`: alias → Desktop alias → static `modelMap` → bare passthrough.
- `src/router.ts::routeModel()`: an otherwise unresolved bare Claude slug falls through to `config.defaultProvider`.

## Expected behaviour

Claude Code internal classifier side requests should not silently lose provider affinity and fall through to an incompatible global default.

Recommended resolution order:

1. **Same-provider affinity:** when the current Claude Code conversation/session is explicitly routed through `RelayA/claude-fable-5`, resolve a bare classifier request for `claude-opus-5` to `RelayA/claude-opus-5` if that model is available and enabled there.
2. **Explicit classifier configuration:** support a clear setting such as `claudeCode.classifierModel` and/or ordered `claudeCode.classifierFallbacks`.
3. **Opt-in compatible fallback:** if the current provider lacks Opus 5, try only user-approved fallback providers in deterministic order. Cross-provider fallback should be explicit because it affects privacy, billing, and data boundaries.
4. **Fail closed with a truthful error:** if no compatible classifier route exists, return an actionable message such as:
   ```text
   No enabled provider can serve the Claude Code Auto Mode classifier
   model claude-opus-5. Configure claudeCode.classifierModel or disable Auto Mode.
   ```
5. **Never silently send a Claude-family classifier slug to a provider whose catalog/model family is incompatible.**
6. Expose classifier routing in diagnostics/UI: requested model, resolved provider/model, affinity source, fallback reason, and availability.

A static `modelMap` should remain a valid explicit override, but users should not need to discover and hand-maintain it merely to make Claude Code's built-in safety path work after selecting a routed model.

## Minimal redacted request or reproduction

```shell
# 1. Configure two enabled providers (all URLs and credentials omitted):
#    - defaultProvider = deepseek
#    - RelayA exposes claude-fable-5 and claude-opus-5
#
# 2. Enable Claude Code proxy integration and leave claudeCode.modelMap empty.
ocx claude

# 3. In Claude Code, select the routed main model:
#    RelayA/claude-fable-5
#
# 4. Switch to Auto permission mode.
#
# 5. Ask Claude Code to perform either action:
#    - run Bash: pwd
#    - launch an Agent/subagent that replies WORKER_OK
#
# 6. Observe:
#    - RelayA/claude-fable-5 main requests return 200
#    - the internal bare claude-opus-5 request routes to deepseek
#    - the classifier request returns 400
#    - Bash/Agent is blocked as "classifier temporarily unavailable"
```

## Actual response or error

```shell
# Redacted OpenCodex usage/route evidence
{
  "provider": "deepseek",
  "model": "claude-opus-5",
  "requestedModel": "claude-opus-5",
  "status": 400,
  "errorCode": "invalid_request_error",
  "routeDecision": {
    "requestedModel": "claude-opus-5",
    "routeKind": "default-provider",
    "selected": {
      "provider": "deepseek",
      "model": "claude-opus-5",
      "reason": "default-provider"
    }
  },
  "upstreamError": "The supported API model names are deepseek-v4-pro or deepseek-v4-flash, but you passed claude-opus-5."
}

# Claude Code tool result
claude-opus-5 is temporarily unavailable, so auto mode cannot
determine the safety of Agent right now. Wait briefly and try again.
```

The same error is reproducible for Bash and Workflow. No API keys, provider URLs, prompts, source code, or account identifiers are included above.

## Upstream documentation

Claude Code permission modes document that Auto Mode performs background safety checks before tool actions:

https://code.claude.com/docs/en/permissions#permission-modes

There is no public specification for the exact internal classifier model slug or its side-request routing contract. The concrete client requirement is observable from Claude Code `2.1.228`: it sends a separate `/v1/messages` request with the bare `claude-opus-5` slug and blocks the tool when that request fails.

Potential implementation directions:

- Track the most recent explicit main-model provider for the Claude Code conversation/session and use it as classifier affinity.
- Recognize the classifier request using stable Claude Code request metadata/shape rather than treating every user-selected bare Claude model as an internal classifier.
- Add validated config:
  ```json
  {
    "claudeCode": {
      "classifierModel": "RelayA/claude-opus-5",
      "classifierFallbacks": [
        "RelayB/claude-opus-5"
      ]
    }
  }
  ```
- Validate candidates against enabled providers and configured/live catalogs before selection.
- Preserve `modelMap` as the highest-priority explicit override.
- Add a route-decision kind such as `claude-classifier-affinity` or `claude-classifier-fallback`.
- Include a regression test where:
  - main model = `RelayA/claude-fable-5`;
  - bare classifier model = `claude-opus-5`;
  - global default = DeepSeek;
  - expected classifier route = `RelayA/claude-opus-5`, never DeepSeek.
- Add a negative test where no approved provider exposes Opus 5 and assert a clear, actionable fail-closed error.

## Additional context and attachments

Related but distinct:

- #545 concerns classifier responses being truncated at a 64-token output limit through an Anthropic translation path.
- This report concerns **provider selection before the request reaches any compatible Claude upstream**.

Impact assessment: **high**. A single hidden routing mismatch disables the core agentic capabilities of Claude Code in Auto Mode while leaving ordinary main-model chat apparently healthy, producing a confusing split-brain failure that can persist across conversations.

## Checks

ardakrt:
## Summary
- Fixes #1697 where Claude Code Auto Mode background safety requests (`claude-opus-5` bare slug) lost provider affinity and fell through to an incompatible global default provider (e.g. DeepSeek or OpenAI Chat), causing HTTP 400 errors and locking tool execution with *"classifier temporarily unavailable"*.
- Adds `classifierModel` and `classifierFallbacks` settings to `OcxClaudeCodeConfig`.
- Enhances `resolveInboundModel` in `src/claude/inbound.ts` to:
  1. Respect explicit user `modelMap` entries as highest priority.
  2. Honor explicit `claudeCode.classifierModel` if configured.
  3. Infer same-provider affinity from `claudeCode.model` (e.g. `RelayA/claude-fable-5` or aliased `claude-ocx-RelayA--...` -> `RelayA/claude-opus-5`).
  4. Honor `claudeCode.classifierFallbacks` when configured.
- Enhances `routeByKnownModelPattern` in `src/router.ts` to match active providers configured with `adapter === "anthropic"` or `adapter === "anthropic-messages"` for `claude-*` models when no provider explicitly named `anthropic` is present.
- Adds regression unit tests in `tests/claude-inbound.test.ts` and `tests/router.test.ts`.

## Test plan
- Run `bun test tests/claude-inbound.test.ts` (all 31 tests passed).
- Run `bun test tests/router.test.ts` (all 24 tests passed).
- Run `bun test tests/claude-inbound.test.ts tests/claude-messages-endpoint.test.ts tests/claude-models-discovery.test.ts tests/claude-cli.test.ts` (all 104 tests passed).
- Run `bun run typecheck` (0 errors).
- Run `bun run privacy:scan` (passed).

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - Added configurable Claude Code Auto Mode classifier routing with a primary model and ordered fallback models.
  - Added support for managing and clearing classifier routing settings through configuration and settings APIs.
  - Explicit classifier model mappings take precedence over fallback settings.

- **Bug Fixes**
  - Disabled providers are now excluded from pattern-based routing.
  - Bare Claude model IDs remain with the configured default provider instead of being assigned implicitly to Anthropic-compatible providers.
  - Invalid classifier settings are safely filtered during configuration loading.

### DIFF
--- a/docs-site/src/content/docs/reference/configuration/server.md
+++ b/docs-site/src/content/docs/reference/configuration/server.md
@@ -160,6 +160,8 @@ These settings govern `/v1/messages`, `/v1/messages/count_tokens`, the `ocx clau
 | `claudeCode.bodyMaxBytes?` | `number` | `67108864` | Cumulative native-passthrough body cap for streamed and buffered responses. Exactly `0` disables. |
 | `claudeCode.authMode?` | `"proxy" \| "subscription"` | auto | How launch handles `ANTHROPIC_AUTH_TOKEN`. Auto detects auth each launch; an explicit value is never overridden. |
 | `claudeCode.authModeMigratedAt?` | `string` | unset | Internal one-time upgrade marker. Do not set manually. |
+| `claudeCode.classifierModel?` | `string` | unset | Explicit target for Claude Code Auto Mode classifier turns, as a qualified `provider/model` (for example `RelayA/claude-opus-5`). Auto Mode sends bare safety checks such as `claude-opus-5` with no provider, so without this they fall through to `defaultProvider` — which may not speak Anthropic at all. Nothing is inferred automatically: only a target you declare here is used. |
+| `claudeCode.classifierFallbacks?` | `string[]` | unset | Ordered classifier targets used when `classifierModel` is not set. Same qualified `provider/model` form; the first usable entry wins. An explicit `modelMap` entry for the classifier model still outranks both. |
 | `claudeCode.subagentEffort?` | `"low" \| "medium" \| "high" \| "xhigh" \| "max"` | inherit | Effort written to generated `~/.claude/agents/ocx-*.md`; separate from Codex guidance and proxy caps. Restart through `ocx claude` to regenerate. |
 
 Auto auth selects subscription when stored Claude auth is found, proxy when none is found, and
--- a/src/claude/inbound.ts
+++ b/src/claude/inbound.ts
@@ -27,18 +27,32 @@ function isRec(v: unknown): v is Rec {
 
 function isClaudeClassifierModel(model: string): boolean {
   const stripped = model.replace(/-\d{8}$/, "");
-  return stripped === "claude-opus-5" || stripped === "claude-opus-4" || /^claude-opus-[45]/.test(stripped);
+  return /^claude-opus-[45]/.test(stripped);
 }
 
-function getClassifierAffinityProvider(mainModel: string | undefined): string | null {
-  if (!mainModel) return null;
-  const resolvedMain = resolveAlias(mainModel) ?? mainModel;
-  const sep = resolvedMain.indexOf("/");
-  if (sep > 0) {
-    const provider = resolvedMain.slice(0, sep);
-    if (provider !== "native" && provider !== "policy") return provider;
+/**
+ * Explicitly configured classifier route for Claude Code Auto Mode safety checks (#1697).
+ *
+ * Only OPERATOR-DECLARED targets are used: `classifierModel`, then the ordered
+ * `classifierFallbacks`. Both are qualified `provider/model` strings the operator chose, so
+ * routing them crosses no boundary the operator did not ask for.
+ *
+ * Deliberately NOT here: inferring a provider from `claudeCode.model`. That value is the
+ * injected/default config slot, not the provider the live session actually selected, so it goes
+ * stale the moment the user changes the model picker -- and acting on it would silently move a
+ * classifier turn onto a provider with its own privacy and billing consequences. Live session
+ * affinity needs the request/session state this function does not have; it is tracked as
+ * follow-up work rather than approximated from static config.
+ */
+function configuredClassifierRoute(cc?: OcxClaudeCodeConfig): string | undefined {
+  const explicit = typeof cc?.classifierModel === "string" ? cc.classifierModel.trim() : "";
+  if (explicit.length > 0) return explicit;
+  if (Array.isArray(cc?.classifierFallbacks)) {
+    for (const candidate of cc.classifierFallbacks) {
+      if (typeof candidate === "string" && candidate.trim().length > 0) return candidate.trim();
+    }
   }
-  return null;
+  return undefined;
 }
 
 /** Alias first, then modelMap: exact id, then date-suffix-stripped (`-\d{8}$`), then classifier affinity/config, else passthrough. */
@@ -64,24 +78,13 @@ export function resolveInboundModel(model: string, cc?: OcxClaudeCodeConfig): st
   const dateless = map[stripped];
   if (typeof dateless === "string" && dateless.length > 0) return dateless;
 
-  // Claude Code Auto Mode classifier routing (issue #1697):
-  // When Claude Code sends internal bare safety checks (e.g. claude-opus-5),
-  // preserve session provider affinity or configured classifierModel so requests
-  // do not fall through to an incompatible defaultProvider.
+  // Claude Code Auto Mode classifier routing (#1697). Bare classifier checks such as
+  // `claude-opus-5` carry no provider, so without this they fall through to defaultProvider --
+  // which may not speak Anthropic at all. Only an operator-declared target is used.
   if (isClaudeClassifierModel(model)) {
-    if (typeof cc?.classifierModel === "string" && cc.classifierModel.trim().length > 0) {
-      return cc.classifierModel.trim();
-    }
-    const affinityProvider = getClassifierAffinityProvider(cc?.model);
-    if (affinityProvider) {
-      return `${affinityProvider}/${model}`;
-    }
-    if (Array.isArray(cc?.classifierFallbacks) && cc.classifierFallbacks.length > 0) {
-      const firstValid = cc.classifierFallbacks.find(fb => typeof fb === "string" && fb.trim().length > 0);
-      if (firstValid) return firstValid.trim();
-    }
+    const configured = configuredClassifierRoute(cc);
+    if (configured) return configured;
   }
-
   return model;
 }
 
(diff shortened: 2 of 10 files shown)


lidge-jun:
Reviewed as part of a bug-PR landing pass. Holding this one: the direction is right, but the implementation changes a routing/privacy boundary in a way that needs a design decision first, not just an approval. Two independent reviews reached the same conclusion from separate reads of the diff.

**The affinity fix does not cover the reported case.** `resolveInboundModel` derives affinity from static `claudeCode.model` (src/claude/inbound.ts:33-41, 75-78), but the request path only ever passes `config.claudeCode` (src/server/claude-messages.ts:629) — never live session state. A picker or `--model` selection that differs from configuration still loses affinity, which is the central case in #1697.

**The router fallback picks a provider arbitrarily.** src/router.ts:699-705 sends every bare `claude-*` to the first enabled provider whose adapter is `anthropic`/`anthropic-messages`, by config insertion order. It does not check that the provider actually exposes the requested model, and it does not ask. With more than one relay configured this silently moves traffic across a privacy, billing, and data boundary; with an incompatible relay it still produces the original 400. `resolveInboundModel` also gates whether a request carrying native Anthropic credentials uses passthrough or translated routing, so this is a security-relevant path.

**`classifierFallbacks` is not failover.** src/claude/inbound.ts:79-81 returns the first non-empty string. There is no availability or model-support check and no attempt at later candidates after a failure. The regression test at tests/claude-inbound.test.ts:313-317 only pins first-item selection, so the name promises more than the code does.

**Two smaller things.** The classifier detector `/^claude-opus-[45]/` treats essentially every Opus 4/5 request as an Auto Mode classifier, while the router fallback applies to every bare `claude-*`. And `classifierModel`/`classifierFallbacks` are added to the TypeScript interface only — they are absent from the management API shape (src/server/management/agent-settings-routes.ts:986-1027) and from docs-site/src/content/docs/guides/claude-code.md:298-313.

What would unblock it: resolve affinity from the live routed model rather than static config; require an explicit provider binding (or a catalog match) before crossing to another Anthropic-adapter provider instead of taking the first enabled one; make `classifierFallbacks` actually try candidates in order; and narrow the classifier detector. Happy to look again once the routing decision is settled — the underlying issue is worth fixing.

For context: CI here is green but shows only policy/labeling checks, no exact-head test or typecheck job, so the test claims in the description are author attestations rather than CI evidence.

lidge-jun:
Triage note (2026-08-15, maintainer): keeping as draft per the existing design hold. The Auto Mode classifier defect is real and unfixed on dev, but this implementation can silently move classifier traffic across providers (static config instead of live routed state, first-non-empty fallback, no model-support validation) — that crosses privacy, billing, and credential-routing boundaries. The design requirements from the earlier review stand: live routed state, explicit or catalog-validated provider crossing, actually-attempted fallbacks, narrowed classifier detection.

Issue #883, opened 2026-08-02
modelReasoningEfforts: [] makes routed models unusable as subagents (catalog supported_reasoning_levels empty)


fyuuwang:
## What happens

When a routed model (e.g. `cursor/composer-2.5`) has `modelReasoningEfforts: []` (explicit empty array) in config, `ocx sync` writes `supported_reasoning_levels: []` into the Codex catalog. The Codex app then rejects the model for `spawn_agent` calls because `[agents].default_subagent_reasoning_effort = "max"` has no valid level to match against the empty list. The request never reaches the proxy, so the model works in the main thread but is unusable as a subagent.

## Root cause (located)

`src/reasoning-effort.ts`, `configuredReasoningEfforts`:

```ts
const modelEfforts = modelRecordValue(provider.modelReasoningEfforts, modelId);
if (modelEfforts !== undefined) return ... [];  // [] is explicit "no reasoning" → catalog []
```

An explicit `[]` and an omitted key behave differently. `[]` produces an empty `supported_reasoning_levels` in the catalog; an omitted key lets the default ladder fill in. A model that genuinely takes no reasoning effort becomes invisible to the subagent roster.

## Expected

Either the `[]` → empty-list semantics should be documented/preserved consistently, or `ocx sync` should detect models in the subagent roster whose catalog `supported_reasoning_levels` is empty and either warn or backfill a placeholder level, so `spawn_agent` with the global default effort is not silently blocked.

## Local workaround (verified)

Remove the `composer-2.5: []` entry from `modelReasoningEfforts`. After `ocx sync`, the catalog fills in the default ladder and subagent routing works (the cursor adapter already drops the effort suffix for composer at the wire, so the upstream is unaffected).

PR #1716, opened 2026-08-14
feat(models): per-custom-model reasoning effort in the Models dashboard

RobinBially:
## What

The custom-model dialog in the **Models** dashboard (and `ocx models add/edit`) now accepts an explicit **reasoning ladder** plus a **default effort**, instead of only display name / context window / modalities.

## Why

Routed models show no reasoning control in clients (Codex, Pi, opencode) unless a ladder is configured, and the dashboard-created custom rows had no way to set one. With this PR:

- Custom rows carry their own `reasoningEfforts` / `defaultReasoningEffort` into the generated Codex catalog, **winning over** the provider-derived ladder (the #962 inheritance stays gap-fill-only).
- An explicit **empty ladder** hides the effort control (#883 semantics), while `null` (GUI) / `-` (CLI) restores inheritance.
- The **Pi export** now emits `reasoning: true` for any row with a non-empty ladder — the catalog's own statement that the model accepts reasoning parameters — so Pi's effort control appears for routed models. (Previously a documented deliberate omission; docs updated in `guides/pi.md`.)

## Changes

| Area | File |
|---|---|
| Data model | `src/types.ts` — `OcxCustomModel.reasoningEfforts?` / `defaultReasoningEffort?` |
| Management API | `src/server/management/model-routes.ts` — POST/PUT validation against the Codex ladder (`low..ultra`), dedupe, default must be a ladder member; `null` clears, `[]` stays as explicit no-reasoning |
| Catalog | `src/codex/catalog/provider-fetch.ts` — explicit custom-row ladder wins over inherited; `src/server/management/model-rows.ts` — stored override exposed to the GUI |
| GUI | `gui/src/pages/Models.tsx` + `models-shared.ts` — reasoning-effort override toggle with ladder checkboxes in the custom-model modal (add & edit; full ladder pre-checked on enable), i18n keys in all 8 locales |
| CLI | `src/cli/models.ts` (`ocx models add`), `src/cli/models-runtime.ts` (`ocx models edit`) — `--reasoning-efforts` / `--default-reasoning-effort`, `-` restores inheritance on both commands |
| Export | `src/clients/config-export.ts` — Pi export `reasoning: true` + `thinkingLevelMap` (pi levels outside the declared ladder hidden) for ladder-carrying rows; `docs-site/.../guides/pi.md` updated |
| Tests | 9 API validation tests, 2 end-to-end catalog tests, updated Pi export test — all green |

## Verification

- `bun test tests/catalog-input-modality-enum.test.ts tests/codex-catalog.test.ts tests/client-config-export.test.ts` — 242 pass, 0 fail
- `bun run lint:gui` (oxlint) — no new findings (one pre-existing error in `use-dashboard-data.ts` on `dev`)

Note: the CLI `-` shorthand restores inheritance; an explicit empty ladder ("no reasoning") is only expressible via the dashboard.

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Custom models now support configurable reasoning-effort levels and defaults in the GUI, CLI, and management APIs.
  * Reasoning settings can be inherited, overridden, cleared, or explicitly disabled.
  * Custom model listings display configured reasoning levels and defaults.
  * Pi exports indicate reasoning support and map available thinking levels appropriately.
* **Localization**
  * Added translations for reasoning settings across supported languages.
* **Documentation**
  * Updated Pi and CLI guidance for reasoning configuration and limitations.

## Review fixes (lens review)

- PUT enforces the POST invariant: a stored default effort is dropped when the final ladder no longer contains it (no stale default re-applying onto an inherited ladder).
- Pi export emits a `thinkingLevelMap` so pi never offers/sends levels outside the declared ladder (custom-row ladders are advertisement-only, no wire clamp).
- Ladders canonicalized to Codex order at API + CLI ingress.
- GUI: enabling the override pre-checks the full ladder (no accidental empty ladder = silently disabled reasoning).
- CLI `ocx models add` accepts `-` for both reasoning flags; empty string rejected; parsing extracted and tested (`tests/cli-models-reasoning.test.ts`).
- Removed dead `defaultReasoningEffort` exposure from `/api/models` rows + GUI `ModelRow`.

## Screenshot

Custom model dialog with reasoning steps


Wibias (changes requested):
Deep re-review of the current head, including the full CodeRabbit history. Several earlier findings are genuinely fixed, but the PR still has merge-blocking correctness issues and a couple of review/test gaps:

### 1. Blocker: an explicit custom ladder can be widened and the synthetic effort can reach the upstream

`applyReasoningLevels()` adds synthetic `max`/`ultra` to most non-empty routed ladders. That policy is normally safe only because provider-level `modelReasoningEfforts` / `reasoningEfforts` are used by `mapReasoningEffort()` to clamp the requested effort before the provider request.

Custom-model `reasoningEfforts` are not part of that runtime clamp. For example, an unknown custom model with an explicit `reasoningEfforts: ["low"]` can be advertised in the Codex catalog as `low,max,ultra`; if Codex selects `max`, `mapReasoningEffort()` has no provider ladder to clamp against and can send `reasoning_effort=max` upstream even though the custom row explicitly declared only `low`.

Please either keep explicit custom-model ladders exact in the catalog (my preference), or make the custom row's ladder authoritative in the runtime effort clamp. Do not synthesize an advertised capability unless the wire path can safely clamp it.

### 2. Blocker: unknown custom models still default to every reasoning level

The known-model GUI path now seeds from the model's advertised ladder, but the unknown-model fallback still does:

```ts
setCustomFormReasoningEfforts(advertised ?? [...REASONING_EFFORT_LEVELS]);
```

This is the substance of CodeRabbit's "Do not preselect unsupported provider capabilities" finding, and it is still real. A custom model is very often unknown to discovery/registry metadata; enabling the override then silently asserts every known effort level. The accompanying comment says the wire clamp bounds what is sent, but that is not true for a custom row whose provider has no `modelReasoningEfforts` entry (see finding 1).

When capability metadata is unknown, start with no selected levels and require the user to choose them explicitly. Preserve existing stored selections on edit.

### 3. Major: the GUI still cannot configure `defaultReasoningEffort`

The PR description says the Models dashboard now accepts a reasoning ladder *plus a default effort*. The API supports and validates `defaultReasoningEffort`, and `/api/models` even returns the stored value, but `Models.tsx` has no default-effort state/control, does not load it during edit, and does not include it in POST/PUT payloads.

This was also raised by CodeRabbit and remains unimplemented. A default configured through CLI/API cannot be viewed or changed in the dashboard; if the ladder is edited so the old default becomes invalid, the backend clears it and the GUI provides no way to choose a replacement.

Please add a default-effort selector tied to the selected ladder, load the stored default on edit, and include it in both create/update payloads. If the feature is intentionally CLI/API-only instead, remove the dashboard/default claims from the PR and do not expose a misleading partial flow.

### 4. Test-integrity: the stale-default transition tests still do not share state

CodeRabbit correctly noted that the "seed default, then shrink/null/empty ladder" tests need the same config fixture across both requests. The finding was marked addressed, but `callCustomModels()` still creates a fresh `config` object on every invocation.

Therefore the follow-up request does not contain the default written by the seed request. Those tests can pass even if the stale-default cleanup is removed from production code. The production cleanup currently looks correct, but the regression tests are giving false confidence.

Please give each transition test one shared mutable config object (or let the helper accept one) and use it for both requests.

### 5. Minor but real: normalize the model id before capability lookup

The capability lookup compares `m.id === customFormModelId`, while submit later saves `customFormModelId.trim()`. Pasting a known id with surrounding whitespace can miss the capability row, enter the unsafe unknown-model fallback, then save the trimmed known id with the wrong seeded ladder.

Use the same trimmed model id for lookup and submission.

### Merge readiness

The PR head is also no longer based on current `dev`, and current `dev` has overlapping changes in `gui/src/pages/Models.tsx` and the locale files. Please update/rebase onto current `dev`, resolve the overlap, rerun CI, and recheck the reasoning UI after the merge.

I rechecked the other CodeRabbit findings as well. The override precedence, stale-default production cleanup, explicit-empty semantics, malformed CSV handling, toggle preservation, initialization tracking, localization, Pi `ultra` mapping, fetch restoration, help text, duplicate declarations, and Pi documentation findings are addressed on this head; no need to revisit those unless the rebase changes them.

PR #1727, opened 2026-08-15
feat(storage): inspect Codex diagnostic logs safely - Part 1

Wibias:
## Summary

- PR 1/3 — **Inspect**: add read-only Codex Log Guard diagnostics for the canonical `logs_2.sqlite`, resolved through Codex's effective `sqlite_home`.
- Report DB/WAL/SHM size, rows/levels/TRACE share/top targets, freelist/reclaimable bytes, schema compatibility, and machine-readable inspection/protection/reclaim eligibility without reading log bodies.
- Expose the same state through the Storage tab, `GET /api/storage/codex-logs`, `GET /api/storage`, `ocx storage codex-logs status`, and `ocx doctor`; an external `sqlite_home` remains outside the existing `CODEX_HOME` storage total.
- Unknown/future schemas remain inspect-only. This PR installs no triggers, runs no checkpoint/vacuum, and performs no Protect/Reclaim mutations.
- Add user documentation for Inspect semantics, privacy boundaries, and checkpointed-snapshot limitations.

## Screenshot

Storage — Codex Log Guard Inspect

## Verification

- TDD: contracts were added red-first for canonical targeting, privacy, zero-write/no-sidecar behaviour, unknown schema, API/CLI/doctor/Storage UI, and external `sqlite_home` handling.
- Full-suite verification caught a real edge where Bun/SQLite accepts a zero-byte file as an empty database; the final implementation now classifies an existing zero-byte canonical DB as `database_unreadable` rather than a future/unknown schema.
- Cross-platform CI run `31859942224` — **SUCCESS** on exact head `5379a3ee805efda5636b9d647fa360449e787b2f` (all four Linux test shards, unsharded macOS control, GUI lint, TypeScript typecheck, GUI tests, privacy scan, GUI build, CLI smoke, storage/API checks, and package/keyring smoke all green; the full Windows test matrix remains workflow-dispatch-only by repository policy).
- React Doctor run `31859942144` — **SUCCESS** on the same exact head.
- A dedicated Codex Log Guard docs page is included. The Astro docs build was not independently run here because the repository docs workflow is `main`/manual-only.

## Checklist

- [ ] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. — explicit management-API security review requested from @Ingwannu; pending.

## Summary by CodeRabbit

- **New Features**
  - Added read-only Codex log diagnostics with database status, sizes, schema compatibility, metrics, and snapshot details.
  - Added storage workspace, management API, and CLI reporting for Codex logs.
  - Added localized labels for inspection-only mode, external SQLite storage, and unavailable diagnostics.
- **Bug Fixes**
  - Inspection failures no longer disrupt storage reporting or expose private paths, contents, or errors.
- **Documentation**
  - Added guidance on inspection behavior, compatibility, snapshots, WAL limitations, and future capabilities.
- **Style**
  - Updated byte formatting to use IEC units such as KiB and MiB.

### DIFF
--- a/docs-site/src/content/docs/guides/codex-log-guard.md
+++ b/docs-site/src/content/docs/guides/codex-log-guard.md
@@ -1,0 +1,69 @@
+---
+title: Codex Log Guard
+description: Inspect Codex diagnostic-log storage safely before enabling future protection or reclaim actions.
+---
+
+OpenCodex can inspect Codex's persistent diagnostic-log database from the **Storage** page and from the CLI. The inspection surface is deliberately read-only: it does not install triggers, delete logs, checkpoint SQLite, vacuum the database, or change Codex configuration.
+
+## What Inspect reports
+
+OpenCodex resolves Codex's effective `sqlite_home` using Codex's existing precedence and inspects the canonical `logs_2.sqlite` database there. A higher-numbered or legacy `logs_N.sqlite` file is never substituted as the mutation-capable target.
+
+The Storage view reports:
+
+- the main database, WAL, and SHM file sizes;
+- total log rows and the share stored at `TRACE` level;
+- the largest log-target buckets by row count, using rank labels instead of target names;
+- SQLite freelist space that may be reclaimable later; and
+- whether the observed schema is compatible with the currently known Codex log schema.
+
+If `sqlite_home` is outside `CODEX_HOME`, the diagnostic database is shown separately. Its bytes are not silently folded into the existing `CODEX_HOME` storage total.
+
+OpenCodex does not select or expose `feedback_log_body` while producing these diagnostics. Log levels are reduced to the fixed known level set plus `OTHER`, and target names are not serialized.
+
+## CLI
+
+```bash
+ocx storage codex-logs status
+ocx storage codex-logs status --json
+ocx doctor
+```
+
+The existing command remains unchanged:
+
+```bash
+ocx storage --json
+```
+
+Its response now also carries the same Codex-log inspection report used by the Storage page.
+
+## Management API
+
+```text
+GET /api/storage/codex-logs
+```
+
+`GET /api/storage` also includes the report as `codexLogs` so the dashboard can refresh the normal storage breakdown and Codex-log diagnostics from one snapshot request.
+
+## Read-only snapshot semantics
+
+Inspection opens the database read-only with SQLite `immutable=1`. This prevents the diagnostic read itself from creating or updating `-wal` or `-shm` sidecars.
+
+The trade-off is important: SQL aggregates describe the last checkpointed database snapshot. If Codex is actively writing, the live WAL can contain newer rows than the aggregate counts. OpenCodex therefore reports the WAL file size separately and does **not** label the result as SSD write rate, NAND writes, or drive-wear/TBW consumption.
+
+## Compatibility states
+
+A known schema reports inspection, future protection, and future reclaim capabilities as supported. A missing, unreadable, or unknown future schema remains inspectable as metadata but is reported as unsupported for mutation-capable operations.
+
+An unknown schema is not guessed into compatibility. This lets a newer Codex version remain observable while preventing later Log Guard releases from treating an unreviewed database layout as safe to modify.
+
+## What is not in this stage
+
+This is **Inspect**, the first Log Guard stage. It does not reduce Codex writes by itself and does not reclaim database pages.
+
+Later stages are intentionally separate:
+
+- **Protect** will add explicit write-reduction modes after safety checks and Codex-process quiescence.
+- **Reclaim** will add explicit, offline, bounded SQLite space reclamation.
+
+Neither action is automatically enabled by Inspect.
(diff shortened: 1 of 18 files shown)


lidge-jun:
Triage note (2026-08-15): substance is close (CI green at head except gates below), but staying draft on three items: (1) the unresolved security/privacy thread — the inspection report serializes sqliteHome/databasePath/codexHome through the API/CLI/UI; keep the serializable surface to safe metadata and return fixed error codes (e.g. inspect_failed) instead of raw error.message at both boundaries; (2) the hygiene gate fails on the deliberate empty_catch blocks; (3) the GUI screenshot is still pending for enforce-target. The i18n nit on the new Storage panel strings also needs locale keys.

lidge-jun (approved):
Approving after an independent adversarial audit (gpt-5.6-sol explorer) plus a re-audit of the fixes.

Round 1 returned FAIL on three blockers. All are now closed:

- **Storage inspection blocked the proxy thread.** readMetrics runs four unbounded aggregates and bun:sqlite is synchronous; measured 48.8s on a real 15.6 GB / 302k-row database, on a path both /api/storage and /api/storage/codex-logs call inline. Inspection is now memoized on file identity, so repeat scans are free (60k-row fixture: 14.3ms cold, 0.1ms warm). The remaining cold-scan cost is explicitly disclosed in the source comment and left for a Worker-based follow-up rather than claimed as fixed.
- **Cache identity was unsound.** Re-audit showed size+mtime lets an atomic replace go unnoticed, serving a stale schema verdict indefinitely. Key now includes dev+ino and nanosecond mtime/ctime; regression drives the exact replace-at-identical-size case.
- **Missing focused regressions.** Added for memoization, write invalidation, explicit reset, and atomic replacement.

Verified at 3307261d6 on an isolated Linux checkout: tests/codex-log-guard-inspect.test.ts 17 pass / 0 fail, bun x tsc --noEmit exit 0, all GitHub checks green.

Approving as the stack parent so #1729 and #1732 can follow.

lidge-jun:
## Ready to merge — blocked only on an approval I cannot give

Audited, fixed, restacked, and verified. The one thing left is a second pair of eyes, and the rule is doing exactly what it should.

### What the audit found and what I fixed

**#1727 — Storage inspection blocked the proxy thread (High).** `readMetrics` runs four unbounded aggregates and `bun:sqlite` is synchronous. Measured **48.8s** on a real 15.6 GB / 302k-row database, on a path both `/api/storage` and `/api/storage/codex-logs` call inline — so opening the Storage page could stall routing and health responses.

Inspection is now memoized on file identity (`3307261d6`). A follow-up review then showed `size+mtime` was not a sound key: an atomic replace preserves both and would serve a stale schema verdict indefinitely. The key now includes `dev`+`ino` and nanosecond `mtime`/`ctime`.

Measured on a 60k-row fixture: **14.3ms cold → 0.1ms warm**, 13.9ms again after an explicit reset.

The remaining **cold-scan** cost is not fixed and is disclosed in the source comment rather than claimed. Moving that work to a Worker is its own change.

**#1729 — compat did not reproduce upstream prefix matching (High).** Upstream registers these filters with `Targets::with_target`, which matches a target *and every module beneath it*. Exact equality caught only the parent, so the high-volume children kept writing while compat reported itself active:

```
hyper_util::client::legacy::pool
codex_api::sse::responses
rmcp::service
codex_http_client::transport::wire
```

First fix used `LIKE`; review caught that SQLite `LIKE` is ASCII case-insensitive, so it also suppressed `HYPER_UTIL::child` — Rust target paths are case-sensitive. Now `substr(NEW.target, 1, N) = 'X::'`, a plain case-sensitive comparison with no wildcard metacharacters to escape.

The `'::'` boundary is a deliberate narrowing from upstream's raw prefix rule: `with_target("hyper_util")` would also match a sibling crate `hyper_utilities`, and suppressing an unrelated crate's logs is worse for a guard that silently discards rows. Both `hyper_utilities` and `HYPER_UTIL::child` are pinned as preserved.

**#1729 — a concurrent Repair could undo a completed Disable (High).** Repair read the desired mode *before* acquiring the cross-process lock, so a Disable landing in that gap completed successfully and was then silently reinstalled. The caller saw "off" and got "compat". Repair now resolves its mode inside the lock and reports the mode actually applied.

**#1732 — four Medium findings** on byte budgets, a documented-but-absent `logicalBytesReclaimed` field, two unmapped maintenance errors, and a dashboard that discards successful compaction reports. These are real and unfixed; see the note below.

### Restacked

`1727 → 1729 → 1732` were each rebuilt on their parent's current head after the fixes, since the chain had gone stale. Merge trees are clean.

### Verification

Run on an isolated Linux checkout at each exact head:

```
#1727  tests/codex-log-guard-inspect.test.ts      17 pass / 0 fail
#1729  tests/codex-log-guard-protection.test.ts   10 pass / 0 fail
       bun x tsc --noEmit                          exit 0
```

Every new regression was reproved to fail without its source change, so none of them pass vacuously.

### Why this is not merged

```
PUT /pulls/1727/merge-async
{"status":"failed","details":{"message":"New changes require approval from
 someone other than lidge-jun because they were the last pusher."}}
```

I pushed the fixes, so my approval does not count — and stacked PRs do not accept the `--admin` bypass that the independent PRs in this batch took. That is the rule working as designed: the whole point is that the person who wrote the change is not the person who clears it.

**What I need:** an approval from a maintainer other than `lidge-jun` on **#1727**, then #1729 and #1732 can cascade. For #1732 specifically, its four Medium findings should be resolved first or explicitly accepted.

For context, this batch already landed on `dev`: #1799, #1764, #1740, #1793, #1750, and #1807.

PR #1732, opened 2026-08-15
feat(log-guard): Reclaim - Part 3

Wibias:
## Summary

- PR 3/3 — **Reclaim**, stacked on #1729 (Protect), which is stacked on #1727 (Inspect).
- Add explicit, offline Codex log-database compaction using bounded `incremental_vacuum` only.
- Reuse Protect's all-Codex quiescence gate and dedicated Log Guard cross-process lock.
- Require known schema, `auto_vacuum=INCREMENTAL`, successful integrity checks, and non-busy SQLite checkpoints.
- Never run full `VACUUM`, never delete/truncate/rename Codex WAL files directly, and never schedule maintenance automatically.

## Screenshot

Storage — Codex Log Guard Reclaim

## Stack

- #1727 — Inspect → `dev`
- #1729 — Protect → #1727 branch
- this PR — Reclaim → #1729 branch
- After each parent merges, retarget its child toward `dev` in order.

## TDD status

The core reclaim contract is committed first. This draft intentionally starts RED: implementation follows only after CI demonstrates the missing maintenance module and safety contract.

## Summary by CodeRabbit

* **New Features**
  * Added manual compaction for Codex diagnostic-log storage.
  * Added Storage settings controls with confirmation, progress handling, and reclamation results.
  * Added CLI and management API support for compacting logs.
  * Added localized labels and operation messages across supported languages.

* **Documentation**
  * Documented usage, safety checks, refusal states, and reclamation metrics.

* **Bug Fixes**
  * Prevents compaction during active processes, unsafe storage states, blocked operations, or failed integrity checks.

* **Tests**
  * Added coverage for UI, CLI, API, safety checks, integrity validation, and bounded reclamation.

### DIFF
--- a/docs-site/src/content/docs/guides/codex-log-guard-reclaim.md
+++ b/docs-site/src/content/docs/guides/codex-log-guard-reclaim.md
@@ -1,0 +1,98 @@
+---
+title: Codex Log Guard Reclaim
+description: Manually reclaim free pages from Codex diagnostic-log SQLite storage with bounded incremental vacuuming.
+---
+
+Reclaim is the manual space-recovery stage of Codex Log Guard. It compacts the canonical Codex `logs_2.sqlite` database only when the database and runtime pass the same safety checks used by Log Guard protection.
+
+Reclaim is **never scheduled automatically** and never runs merely because the Storage page is opened. The dashboard requires an explicit Compact action and a second confirmation before it sends the mutation request.
+
+## What Reclaim does
+
+OpenCodex performs a bounded offline maintenance sequence:
+
+1. resolve the canonical `logs_2.sqlite` through Codex's effective `sqlite_home`;
+2. verify the file identity and known Codex log schema;
+3. verify process enumeration succeeded and no supported Codex writer process is running;
+4. acquire the dedicated cross-process Log Guard lock;
+5. repeat the Codex-process check while that lock is held;
+6. open the existing database read/write without create semantics and prove an immediate SQLite writer can be acquired;
+7. require `PRAGMA auto_vacuum` to already be `INCREMENTAL`;
+8. run `PRAGMA quick_check` before maintenance;
+9. run a full WAL checkpoint and refuse a busy/incomplete checkpoint;
+10. execute bounded `PRAGMA incremental_vacuum(N)` batches, checkpointing after each batch;
+11. run `PRAGMA quick_check` again after maintenance; and
+12. report before/after database, WAL, page-count, freelist, and reclaimable-byte metrics.
+
+The default batch target is approximately **8 MiB of SQLite pages**. A single invocation reclaims at most approximately **256 MiB of pages**, with an additional finite iteration cap. If more free pages remain, the result is reported as partial and you can invoke Compact again later.
+
+The byte limits are converted to page counts using the database's actual SQLite page size. They are bounds on logical SQLite pages processed, not claims about SSD/NAND write volume.
+
+## Safety guarantees
+
+Reclaim deliberately does **not**:
+
+- run full `VACUUM`;
+- change `auto_vacuum` mode on an existing Codex database;
+- delete, truncate, rename, or otherwise manipulate Codex `-wal` / `-shm` files directly;
+- delete diagnostic rows;
+- modify Log Guard protection triggers or unrelated user triggers;
+- run while Codex is detected as active;
+- proceed when process enumeration is uncertain;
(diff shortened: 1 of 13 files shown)


lidge-jun:
Triage note (2026-08-15): staying draft — the dedicated reclaim workflow's GUI contract step fails at head and the gates job is red, and two Major review threads look plausible (post-open path re-validation gap in runCompaction, and the iteration-cap stopReason misclassification). Also consider dropping the one-off log-guard-reclaim-tdd.yml workflow, which duplicates main CI coverage.

lidge-jun:
## Pre-merge review — four findings, all still open

Audited independently at `d737627c0`. Verification is clean: `bun x tsc --noEmit` exit 0, focused suites 15/5/6 pass, GUI lint and build clean, privacy scan passed, docs built 385 pages. The stack was restacked onto the current #1729 head, and merge trees are conflict-free.

Four findings remain. I have **not** fixed these — unlike the #1727/#1729 findings, each needs a product decision about what the documented contract should be, and guessing would just move the inconsistency somewhere else.

**1. Medium — the implementation does not honor the documented byte budgets.**

`src/codex/log-guard/maintenance.ts:27-28` hard-codes `512` batch pages and `8_192` max pages. `:243-246` reads `pageSize` but never uses it for either budget. The guide promises ~8 MiB batches and ~256 MiB/run *converted using the actual page size*. At the 4 KiB page size the fixtures use:

```
{"pageSize":4096,"batchMiB":2,"maxMiB":32,"documentedBatchMiB":8,"documentedMaxMiB":256}
```

Either derive page counts from `PRAGMA page_size` after reading it, or change the documented figures. Whichever you pick, a regression across two page sizes should pin it.

**2. Medium — the documented `logicalBytesReclaimed` field does not exist.**

`docs-site/.../codex-log-guard-reclaim.md:94` describes it; `maintenance.ts:42-52` does not define it and `:267-277` does not emit it. Either emit `observedPagesReclaimed * pageSize` and assert it, or drop it from the docs if it is meant to be derived client-side.

**3. Medium — two new maintenance errors fall through to a generic dashboard message.**

`maintenance.ts:61-62` introduces `auto_vacuum_not_incremental` and `integrity_check_failed`. `StorageWorkspace.tsx:120-130` has no case for either and `log-guard-operation-labels.ts:3-11` has no labels. A user cannot tell an unsupported auto-vacuum configuration from an integrity failure — two very different situations.

**4. Medium — the dashboard discards successful compaction results.**

`StorageWorkspace.tsx:403-419` checks that the POST succeeded but never reads its report, then attempts a best-effort status refresh. If that refresh fails the function returns with the stale report and no visible outcome. The existing test confirms this path passes with nothing shown to the user:

```
(pass) successful compact remains successful when the status refresh fails
```

So `pagesReclaimed`, `complete`, and `stopReason` are hidden, and a user may repeat an already-successful mutation. Parse and display the POST report; treat the GET as an independent refresh.

## Also blocking

This is the top of a stack, so #1727 and #1729 must land first. Both are fixed and restacked but blocked on approval from someone other than `lidge-jun`, since I pushed their fixes — see #1727 for detail.

For context, this batch already landed on `dev`: #1799, #1764, #1740, #1793, #1750, and #1807.

lidge-jun (approved):
Approving after independent audit and three rounds of fixes; the final round returned PASS with zero blockers.

Findings closed:

- Byte budgets were fixed page counts while the guide promised ~8 MiB batches and ~256 MiB per run converted with the real page size (at 4 KiB pages that was 2 MiB/32 MiB). Budgets are now byte-based and converted after reading PRAGMA page_size.
- logicalBytesReclaimed was documented but never existed; now emitted as pagesReclaimed * pageSize, deliberately distinct from the physical figure since an incremental vacuum can reclaim pages without the file shrinking.
- auto_vacuum_not_incremental and integrity_check_failed fell through to the generic message; both now have labels in all 9 locales and explicit routing.
- The dashboard discarded successful compaction results. Follow-up review then found the first fix incomplete twice over: the result was nested inside the reclaim section, which a fully successful compaction retires, and the summary omitted pagesReclaimed/stopReason. Both fixed, plus a stale receipt no longer survives a failed retry, and zh-TW uses the Traditional page unit.

Verified remotely at the audited head: typecheck exit 0, Log Guard 85 pass / 2 skip / 0 fail, GUI 881 pass / 0 fail, gui lint and lint:i18n clean. Every new regression was reproved to fail without its source change.

PR #1739, opened 2026-08-15
fix(google): allow AI Studio bare Gemini Flash IDs

chilung-cgu:
## Summary

- **Observed problem:** With an affected Google AI Studio configuration, selecting `gemini-3.7-flash` made opencodex send `gemini-3.7-flash-tiered` on the wire. The upstream returned HTTP 404 before generation, so Codex Desktop appeared to receive no model response.
- **Observed evidence:** On 2026-08-15, the configured AI Studio key returned HTTP 200 for bare `gemini-3.7-flash` and `gemini-3.6-flash`, HTTP 404 for both `-tiered` ids, and `ListModels` contained no `-tiered` model. This is an observed deployment compatibility difference, not a claim that every Google account behaves the same way.
- Add provider-level `directGeminiWireRenames: false` for affected AI Studio providers. The default remains the existing `-tiered` mapping, so installations where `-tiered` works retain their behavior.
- Scope the setting to AI Studio direct requests: Vertex retains its requested model identity and Cloud Code Assist routing remains unchanged.
- This is an explicit operator setting, not automatic 404 fallback or model discovery. After a release containing this PR, add it to the affected Google provider entry in `config.json` and restart the proxy.
- Add focused adapter and config regression coverage for Gemini 3.6/3.7, including a persisted `false` round-trip, and document the setting.
- Rebased onto the current `dev` head (2026-08-17, `8f7a22ff7`) and resolved the `src/config.ts` / `src/types.ts` conflict without dropping either `upstreamHttpVersion` or `directGeminiWireRenames`; branch is 0 commits behind.
- Review feedback is addressed on the rebased head: the complete boolean contract is documented; direct AI Studio identity uses the public model id; and the rebased equivalent of `e4d83f6e6` preserves the resolved routed identity for Cloud Code Assist aliases, including the retired `gemini-3.6-flash` → `gemini-3.7-flash-tiered` migration regression.

## Verification

- `bun run typecheck` — passed on current head.
- `bun run privacy:scan` — passed on current head.
- `bun test tests/google-adapter.test.ts tests/google-antigravity-wire.test.ts tests/config.test.ts` — 223 passed, 0 failed, 905 assertions on the rebased head.
- `cd docs-site && bun run build` — passed (385 pages).
- Monolithic `bun run test` did not produce a green local result on this host: one run hit an unrelated fixed-5-second catalog-sync timeout that passed alone in 2.95 seconds; later runs were terminated by the host with SIGKILL/exit 137 before the final summary.
- Memory-safe sharded reruns produced complete `0 fail` summaries for 812 of 814 test files. The two remaining unrelated files (`tests/integrations-invariants.test.ts` and `tests/opencode-cli.test.ts`) were also terminated with exit 137 before their summaries while the host had about 386 MiB of free pages. The local-CI checkbox therefore remains unchecked.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added an optional Google provider setting to control Gemini Flash model ID renaming for direct AI Studio requests.
  * By default, model IDs retain the existing `-tiered` suffix behavior; setting the option to `false` preserves original IDs.
  * Vertex and Cloud Code Assist request behavior remains unchanged.

* **Bug Fixes**
  * System identity now uses the public model name for AI Studio and Vertex, and the resolved routed model for Cloud Code Assist.
  * Invalid setting values produce configuration diagnostics.

### DIFF
--- a/docs-site/src/content/docs/reference/configuration/providers.md
+++ b/docs-site/src/content/docs/reference/configuration/providers.md
@@ -122,7 +122,7 @@ differing backup and rewrites known legacy namespaced selected ids to bare ids.
 | `escapeBuiltinToolNames?` | `boolean` | Escape built-in tool names for Anthropic-compatible gateways and restore them in returned calls. |
 | `anthropicEofTolerance?` | `boolean` | Let an Anthropic-compatible gateway complete a stream that ends before `message_stop`, only when visible text or a complete JSON-object tool input was received. Off by default. |
 | `googleMode?` | `"ai-studio" \| "vertex" \| "cloud-code-assist"` | Google transport/auth mode. Default `ai-studio`. |
-| `directGeminiWireRenames?` | `boolean` | Google only. When `false`, the AI Studio (direct) path sends Gemini Flash ids to the wire unchanged instead of applying the `-tiered` suffix (`gemini-3.7-flash` -> `gemini-3.7-flash-tiered`). Defaults to the rename; set `false` when the configured upstream still serves the bare ids. |
+| `directGeminiWireRenames?` | `boolean` | Google only. This affects only direct AI Studio requests. When `false`, the AI Studio path sends Gemini Flash ids to the wire unchanged instead of applying the `-tiered` suffix (`gemini-3.7-flash` -> `gemini-3.7-flash-tiered`). Vertex preserves the requested model ID, and Cloud Code Assist routing is unchanged. Defaults to the rename; set `false` when the configured upstream still serves the bare ids. |
 | `project?` | `string` | Vertex or Antigravity Cloud Code Assist project id. |
 | `location?` | `string` | Vertex location; environment fallback is `GOOGLE_CLOUD_LOCATION`. |
 | `mcpServers?` | `Record<string, CursorMcpServerConfig>` | Cursor only: stdio or Streamable HTTP MCP servers. |
--- a/docs-site/src/content/docs/reference/configuration/providers.md
+++ b/docs-site/src/content/docs/reference/configuration/providers.md
@@ -122,7 +122,7 @@ differing backup and rewrites known legacy namespaced selected ids to bare ids.
 | `escapeBuiltinToolNames?` | `boolean` | Escape built-in tool names for Anthropic-compatible gateways and restore them in returned calls. |
 | `anthropicEofTolerance?` | `boolean` | Let an Anthropic-compatible gateway complete a stream that ends before `message_stop`, only when visible text or a complete JSON-object tool input was received. Off by default. |
 | `googleMode?` | `"ai-studio" \| "vertex" \| "cloud-code-assist"` | Google transport/auth mode. Default `ai-studio`. |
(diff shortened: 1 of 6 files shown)


chilung-cgu:
@coderabbitai review

Wibias (changes requested):
Requested changes from full review:

1. **Document the boolean contract explicitly.** `directGeminiWireRenames` should state both sides of the behavior, not only the opt-out: `undefined`/`true` applies the known `-tiered` wire rename for direct AI Studio requests; `false` preserves the requested bare model ID. Vertex and Cloud Code Assist remain unaffected.

2. **Add a regression test for the system-identity fix.** The production change correctly keeps `parsed.modelId` in the system identity while using the rewritten ID only on the wire, but the current tests only assert URL/wire behavior. Please add a request containing the Codex/OpenAI identity-bearing system prompt, build it with the tiered rename active, and assert the serialized Gemini `systemInstruction` identifies the public model (e.g. `gemini-3.7-flash`) and does **not** contain `gemini-3.7-flash-tiered`.

The runtime opt-out behavior itself looks correct and current Cross-platform CI is green. These two items close the remaining contract/regression gaps.

Wibias (changes requested):
Re-review on current head `e98fd2fb56316d4f73f22a226103d864caba47c2`.

The two findings from my previous review are fixed: the full `directGeminiWireRenames` boolean contract is now documented, and the public-model identity regression is covered for direct AI Studio.

One correctness blocker remains:

1. **[P2] The system-identity fix is applied too broadly across Google modes.** `messagesToGeminiFormat()` now always calls `identifyRoutedModel(..., parsed.modelId)`, even though Cloud Code Assist first resolves a distinct `routedModelId` with `resolveAntigravityEffortWireModel()`. That resolver does real model migration/aliasing, not just a cosmetic direct-AI-Studio suffix rewrite. For example, retired `gemini-3.6-flash` selections are intentionally routed to the current `gemini-3.7-flash-tiered` CCA wire model. The current code would therefore send the request to 3.7 while injecting a system instruction telling it to identify as 3.6. This conflicts with the `identifyRoutedModel` contract, which expects the concrete model actually sent upstream.

Please scope the public-ID identity exception to the direct AI Studio `-tiered` rename only. Cloud Code Assist should continue identifying with its resolved routed/wire model. A simple shape would be to pass an explicit identity model into `messagesToGeminiFormat()`: use `routedModelId` for `cloud-code-assist`, and `parsed.modelId` for direct AI Studio/Vertex. Add a focused CCA regression using a migrated/retired alias such as `gemini-3.6-flash` and assert the system identity does not claim the retired model.

All current CodeRabbit inline threads are resolved. The branch is also behind current `dev`, and exact-head React Doctor / Cross-platform CI are still `action_required`; after the code fix, please refresh onto current `dev` and run the real workflows on the resulting head.

lidge-jun:
Landed on `dev` through #1921 (merge `9dbc5fc42`), which merges this branch's
commits rather than re-implementing them. Closes #1894.

The open blocker from the last review is resolved on this head: commit `f6c88febf`
scopes the public-id identity exception to the direct AI Studio rename, so Cloud
Code Assist keeps identifying as its resolved routed model. That distinction
matters because `resolveAntigravityEffortWireModel` performs real migration — a
retired `gemini-3.6-flash` selection genuinely runs on `gemini-3.7-flash-tiered` —
and the CCA regression asserting `envelope.model === "gemini-3.7-flash-tiered"`
with a system identity that does not claim 3.6 is exactly the coverage that was
requested.

Verified on the merged tree: `bun run typecheck` green, `bun run privacy:scan`
green, and the full suite at 12747 pass / 10 skip / 0 fail across 823 files.

Thanks for the fix, and for the two live captures — the disagreement between this
account and the one behind `a70bb78d4` is what turned a spelling argument into a
configuration boundary.

PR #1748, opened 2026-08-15
fix(outbound): route Clash fake-IP discoveries through the HTTP proxy

Blushyes:
## Summary

- Live `/models` discovery (and any other provider outbound GET/POST) failed for custom providers such as Packy and Aihubmix on machines that use Clash / Surge / Mihomo **fake-IP DNS**. Those hostnames resolve to `198.18.0.0/15`. Write-time validation already treats that as private, so operators set `allowPrivateNetwork: true` to save the row. Outbound then refused the request unless the hostname was in `NO_PROXY`.
- That `NO_PROXY` requirement is the wrong control for fake-IP. `NO_PROXY` pin-connects to `198.18.x.x`. The working path is a hostname `CONNECT` through the configured `HTTP(S)_PROXY`, the same path a public destination already takes.
- Hostname answers that are **only** Clash fake-IP space are no longer marked `privateNetwork`. If an older `allowPrivateNetwork` path still classifies them as private, outbound still sends them through the proxy instead of throwing.
- Literal `http://198.18.x.x` destinations, real RFC1918 / loopback / metadata, and mixed fake-IP + RFC1918 answers are unchanged. Real LAN providers still need `allowPrivateNetwork` plus `NO_PROXY`.
- Write-time `allowBenchmarkAddresses` stays openai-seed-only. This PR only unblocks outbound after the provider is already saved.

## Verification

- `bun test tests/destination-policy-resolved.test.ts tests/provider-outbound.test.ts tests/provider-outbound-private-network.test.ts`
  - 45 pass, 0 fail (includes the new Clash fake-IP cases and the existing RFC1918 `NO_PROXY` requirement).
- `bun test tests/management-provider-validation.test.ts -t "canonical OpenAI POST passes allowBenchmarkAddresses"`
  - 1 pass (write-time openai-only exemption unchanged).
- `bun run typecheck` — pass.
- `bun run privacy:scan` — pass.
- Manual: on a Clash fake-IP host with `HTTPS_PROXY` set, Packy `/v1/models` returned 13 models and Aihubmix returned 395 after this outbound change. Gw (real public A record) was unaffected.
- Full `bun run test` on this Clash machine reports unrelated environment failures (test hostnames such as `opencode.ai` resolve to `198.18.x.x` at write time; GUI `react` is not installed in this checkout). Those suites were already outside this diff.

## Checklist

## Review readiness checklist

- [x] This branch is rebased onto the latest `dev` commit.
- [ ] All correct Codex and CodeRabbit findings on this head are resolved.
- [x] This pull request is ready for review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Provider diagnostics now continue through the configured proxy when hostnames resolve exclusively to Clash, Surge, or Mihomo fake-IP addresses.
  * Improved handling of private and local destinations, including mixed fake-IP and private-address resolutions.
  * Diagnostic requests preserve hostnames, follow redirects safely, and remove credentials during redirects.

* **Documentation**
  * Updated English and Chinese documentation to clarify proxy routing, `NO_PROXY` requirements, and fake-IP behavior.


lidge-jun:
Superseded by #2037 (merged to dev as 8b9277fa7), which re-implements this outbound-only: hostname answers in 198.18.0.0/15 are admitted via an explicit allowBenchmarkAddresses opt-in that arms only when an outbound proxy is configured, so the hostname rides the proxy CONNECT. Literal 198.18.x URLs, mixed answers, and non-opt-in callers (image/Lab fetch) keep rejecting — this avoids widening resolvePublicAddresses for all callers as this PR did. Thanks for the diagnosis and the original fix; credited in the commit.

Ingwannu:
The fake-IP proxy support from #2037 is now in `dev`. Exact-head review found one remaining route-selection gap: global proxy configuration enabled the 198.18/15 exception even when the target hostname matched `NO_PROXY`, where the actual request would bypass the proxy.

Follow-up #2045 scopes the exception to hosts that will actually use the proxy and preserves rejection on direct NO_PROXY routes. Local verification is green (44 focused destination/outbound tests, typecheck, privacy scan); independent review and CI are pending.

PR #1761, opened 2026-08-15
feat(quota): probe Command Code 5h/weekly credits

yansigit:
## Summary

- Add a Command Code quota probe so the dashboard and `ocx account refresh` can show 5-hour and weekly windows from `/alpha/billing/credits`.
- Cover both presets: OAuth `command-code` uses the stored account bearer; Provider-API `commandcode` uses the active configured key.
- When Command Code also reports remaining monthly/purchased/free credits and period spend, surface that as a USD window. Soft-fail the extra spend calls so a down subscriptions/summary endpoint still leaves the rolling windows visible.
- Keep credentials on the canonical `https://api.commandcode.ai` host only. Lookalike destinations and `forward`/`local` auth modes are never probed.

## Verification

- `bun test tests/command-code-quota.test.ts tests/opencode-go-quota.test.ts tests/provider-quota.test.ts`
- Result: 106 pass, 0 fail.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added Command Code quota visibility for OAuth and API-key configurations.
  * Dashboard and `ocx account refresh` now show five-hour and weekly usage windows.
  * Monthly, purchased, and free credit balances may appear in USD when available.
  * Quota checks use supported canonical API hosts and configured credentials; custom lookalike URLs are excluded.

* **Documentation**
  * Added Command Code quota guidance across supported languages, including authentication and endpoint limitations.

### DIFF
--- a/docs-site/src/content/docs/fr/guides/providers.md
+++ b/docs-site/src/content/docs/fr/guides/providers.md
@@ -394,6 +394,14 @@ CLI Command Code. Le catalogue, propre au compte, provient du point de terminais
 après la connexion. Les requêtes de chat utilisent la clé Bearer configurée. Créez des clés dans
 [Command Code Studio](https://commandcode.ai/studio/).
 
+**Quota Command Code.** Le tableau de bord et `ocx account refresh` sondent les fenêtres
+`/alpha/billing/credits` de Command Code (5 heures et hebdomadaire) sur l'hôte canonique
+`https://api.commandcode.ai`. Le préréglage OAuth (`command-code`) utilise le jeton porteur du compte
+enregistré ; le préréglage à clé d'API fournisseur (`commandcode`) utilise la clé active configurée. Une
+URL de base modifiée ressemblant à l'original n'est jamais sondée. Les crédits mensuels, achetés et
+gratuits restants sont affichés sous forme de fenêtre en USD lorsque Command Code signale également les
+dépenses de la période.
+
 **Découverte SambaNova Cloud.** Le préréglage lit la liste publique `/v1/models` de SambaNova Cloud depuis
 l'hôte API fixe, préserve les identifiants natifs du fournisseur et limite la découverte à 128 KiB et 128
 lignes brutes. Le catalogue n'étant pas authentifié, le parcours de connexion de la CLI signale que la clé ne
--- a/docs-site/src/content/docs/guides/providers.md
+++ b/docs-site/src/content/docs/guides/providers.md
@@ -367,6 +367,13 @@ account-scoped and comes from the authenticated discovery endpoint after login. 
 configured Bearer key. Create keys at
 [Command Code Studio](https://commandcode.ai/studio/).
 
+**Command Code quota.** The dashboard and `ocx account refresh` probe Command Code's
+`/alpha/billing/credits` windows (5-hour and weekly) on the canonical
+`https://api.commandcode.ai` host. The OAuth preset (`command-code`) uses the stored
+account bearer; the Provider-API key preset (`commandcode`) uses the active configured
+key. A user-edited lookalike base URL is never probed. Remaining monthly, purchased, and
+free credits are shown as a USD window when Command Code also reports period spend.
+
 **SambaNova Cloud discovery.** The preset reads SambaNova Cloud's public `/v1/models` list from the fixed API
 host, preserves provider-native ids, and caps discovery at 128 KiB and 128 raw rows. Because the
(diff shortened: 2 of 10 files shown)


yansigit:
@coderabbitai review

Wibias (changes requested):
Requesting changes based on the current head (`2d5f374`).

There are two quota-semantics issues plus a failing required CI gate:

[P2] A real zero-credit balance is treated as “no credit data”. The new credit aggregation normalizes missing components to `0` and then returns `undefined` when all components are zero. That conflates an absent payload with a valid `{ monthlyCredits: 0, purchasedCredits: 0, freeCredits: 0 }` balance. A fully exhausted account should still surface a credit window with `remaining = 0` / fully consumed state. Please track field presence separately from the numeric value and add a zero-balance regression.

[P2] `currentPeriodEnd` is applied to an aggregate that also includes purchased/top-up credits. Purchased Command Code credits roll over rather than expiring with the subscription period, so attaching the subscription period end to the combined monthly + purchased + free balance gives the purchased portion a false expiry/reset semantic. Please either model expiring monthly and non-expiring purchased balances separately, or avoid emitting an expiry for a mixed aggregate where it would be incorrect. Add coverage for purchased credits crossing a subscription-period boundary.

Additional required cleanup:
- the test claiming `forward or local` are never probed only exercises `forward`; add a `local` negative regression;
- current Cross-platform CI is red because `privacy:scan` flags the new bearer-token-looking test fixture in `tests/command-code-quota.test.ts`; replace it with a privacy-scan-safe synthetic fixture rather than bypassing the gate;
- refresh onto current `dev`, rerun full CI, and sync the new Command Code quota docs into the French provider guide introduced after this branch point.

yansigit:
@Wibias @lidge-jun this is a first-time fork PR, so GitHub is holding the required workflows until a maintainer approves them. Could one of you approve Cross-platform CI and React Doctor on this latest push?

Wibias (dismissed):
Almost there — the previous review points are addressed cleanly, the branch is current with `dev`, the privacy gate is green, and I only found one remaining code issue in this pass. This is not a redesign or another round of broad cleanup.

One blocker remains in `fetchCommandCodeSpend()`:

- If `/alpha/billing/subscriptions` fails or does not provide `currentPeriodStart`, the code still calls `/alpha/usage/summary` without `since=...` and publishes that unscoped historical spend as the current `creditsUsd` meter. The code itself notes that this value may run high on aged accounts. Because the remaining credit pools are current-cycle/current-balance data, combining them with unscoped historical spend can produce a materially wrong percent.

Please fail soft for the USD window when the billing period cannot be established: if there is no valid `currentPeriodStart`, omit `creditsUsd` rather than querying/publishing unscoped spend. The 5-hour/weekly windows from `/alpha/billing/credits` should still remain visible.

Please add a focused regression covering: valid rolling windows + credits, subscriptions unavailable/no period start, usage summary would return historical spend -> rolling windows remain, `creditsUsd` is absent.

Everything else from my previous review is now addressed: zero balances, purchased-credit expiry semantics, the explicit `local` negative test, the privacy-safe fixture, French docs, and the `dev` refresh. The old CodeRabbit `local`-test thread is therefore stale and can just be resolved. Once this one scoped fix is in and exact-head CI finishes green, I don't currently see another merge blocker.

yansigit:
@Wibias addressed on `da6cbb975`. If subscriptions omit `currentPeriodStart`, the probe now skips `/alpha/usage/summary` and leaves `creditsUsd` off. The 5-hour/weekly windows still report. Added the regression you asked for.

Wibias (approved):
Approved on exact head `27956c8adb80a4fbf9ffa1e4411572d636f97bc1` after re-review. The remaining CodeRabbit regression finding is fixed correctly, all inline review threads are resolved, and exact-head Cross-platform CI plus React Doctor are green. The two minor EN/FR documentation cleanups are intentionally tracked in follow-up PR #1803 and are not blockers for this feature PR.

`dev` has advanced since this head was cut, so the merge path should still apply the repository's normal current-base/update gate and revalidate after any resulting head/base change.

Wibias:
Thanks @yansigit — this is a genuinely useful addition. Having Command Code's 5-hour and weekly quota visible directly in the Providers dashboard and `ocx account refresh` gives users a much clearer picture of quota pressure before requests start failing, while the optional USD balance makes remaining account capacity easier to understand. Nice work keeping the probe scoped to the canonical host and the correct active credential paths as well.

PR #1771, opened 2026-08-15
fix(google): allow opting out of the direct Gemini -tiered wire rename

chilung-cgu:
## Summary

Google's Generative Language API renamed current Gemini Flash generations on the wire with a `-tiered` suffix (`gemini-3.7-flash` -> `gemini-3.7-flash-tiered`, `gemini-3.6-flash` -> `gemini-3.6-flash-tiered`), and opencodex applies that rename by default in the AI Studio (direct) path. Some AI Studio deployments/keys still serve the bare ids and return `404 NOT_FOUND` (`models/gemini-3.7-flash-tiered is not found for API version v1beta`) for every renamed request.

This PR adds a per-provider opt-out: `directGeminiWireRenames: false` keeps the picker-visible id stable while sending the bare id on the wire. Default behavior (rename applied) is unchanged; Vertex requests are unaffected.

## Verification

- `bun run typecheck` passes.
- `bun test tests/google-adapter.test.ts tests/config.test.ts` passes (159 tests, including new coverage for the opt-out round-trip and the Vertex exclusion).
- Live check against `generativelanguage.googleapis.com/v1beta/models` confirmed the affected account serves bare ids only; with the opt-out set, the request reaches `models/gemini-3.7-flash:generateContent` and succeeds.

## Checklist

- [ ] Docs or release notes were updated when needed (providers.md documents the new setting).

## Review readiness checklist


chilung-cgu:
Duplicate of #1739 (same fix: directGeminiWireRenames opt-out, cherry-picked onto latest dev). Keeping #1739; closing this one to avoid double review. Reopen if preferred.

PR #1772, opened 2026-08-15
fix(google): replay Gemini thought signatures through the web-search loop

chilung-cgu:
## Summary

Gemini (AI Studio direct mode) rejects functionCall parts that lack a `thought_signature` with `INVALID_ARGUMENT` ("Function call is missing a thought_signature in functionCall parts"). Three paths dropped or failed to replay it:

1. **Web-search loop** — the loop replays the model's `web_search` call into its iteration #2 request without the signature the upstream response carried.
2. **Client replay** — Codex re-sends prior tool calls (e.g. `default_api:exec`) in the conversation history without the signature; direct mode neither observed nor re-injected signatures, unlike CCA/Vertex which share the existing stateless replay cache.
3. **Large tool argument overflow** — when tool call arguments exceeded `REPLAY_MAX_CANONICAL_ARGS_BYTES` (64 KiB), `canonicalJsonBounded` returned `null`, causing `functionCallKey` to return `undefined` and dropping thought signature caching/replaying for large payload tool turns.

Changes:
- Google adapter emits the part-level `thoughtSignature` on `tool_call_start` events (streaming and buffered parse paths).
- The web-search loop captures it and attaches it to the replayed assistant toolCall.
- AI Studio direct mode now uses the existing `google-antigravity-replay` cache (observe on both parse paths, apply in `messagesToGeminiFormat`, clear-on-invalid), namespaced by credential fingerprint + wire model so signatures cannot cross providers, keys, or routes; the stable client-thread session derivation is reused.
- The adapter provides `formatErrorBody` for direct mode too, classified by `googleMode` ("Antigravity", "Vertex AI", "Gemini"), so upstream error bodies stop being dropped ("Provider error N" gains the real message).
- Streaming canonical SHA-256 fallback for large tool arguments (over 64 KiB) so large tool calls continue to preserve thought signatures across replays without memory exhaustion.
- The web-search sidecar sends only the bare `web_search` tool shape; the ChatGPT (codex) backend rejects extra parameters such as `max_results` ("Unknown parameter: 'tools[0].max_results'"), which made every search fail.

## Verification

- `bun run typecheck` passes.
- `bun run privacy:scan` passes.
- `bun test` passes (all test files and regression tests pass).
- Added regression tests in `tests/google-antigravity-replay.test.ts` verifying streaming SHA-256 calculation for tool arguments over 64 KiB preserving thought signatures.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Google error handling across AI Studio, Vertex, and Antigravity.
  * Preserved thought signatures during streaming, buffered tool calls, web searches, and follow-up responses.
  * Improved replay reliability for large tool-call arguments.
  * Prevented hosted web-search configuration from being unnecessarily replayed.

* **Tests**
  * Added coverage for AI Studio thought-signature replay, provider-specific errors, web-search signatures, and large-argument replay scenarios.

### DIFF
--- a/src/adapters/google-antigravity-replay.ts
+++ b/src/adapters/google-antigravity-replay.ts
@@ -446,17 +446,37 @@ function canonicalJsonBounded(value: unknown, maxBytes: number): string | null {
  */
 function functionCallKey(name: unknown, args: unknown): string | undefined {
   if (typeof name !== "string" || name.length === 0) return undefined;
+  const hash = createHash("sha256");
+  updateHashWithString(hash, name);
   let canonical: string | null;
   try {
     canonical = canonicalJsonBounded(args ?? {}, REPLAY_MAX_CANONICAL_ARGS_BYTES);
   } catch {
-    canonical = "";
+    canonical = null;
   }
-  if (canonical === null) return undefined;
-  const hash = createHash("sha256");
-  updateHashWithString(hash, name);
-  updateHashWithString(hash, canonical);
-  return hash.digest("hex");
+  if (canonical !== null) {
+    updateHashWithString(hash, canonical);
+    return hash.digest("hex");
+  }
+  const buf = Buffer.allocUnsafe(8192);
+  let offset = 0;
+  const sink = (chunk: string) => {
+    for (let index = 0; index < chunk.length; index += 1) {
+      buf.writeUInt16LE(chunk.charCodeAt(index), offset);
+      offset += 2;
+      if (offset === buf.length) {
+        hash.update(buf);
+        offset = 0;
+      }
+    }
+  };
+  try {
+    writeCanonicalJson(args ?? {}, sink);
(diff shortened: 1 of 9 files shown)


lidge-jun:
Superseded by #1781, landed on `dev` as 4b950101a with you as co-author — thank you for finding this.

Your diagnosis was right, and the web-search replay you wrote is the behavior that shipped. The reason this went in as a different change: the signature was still being lost on the shared path, so the fix had to move up to where the tool call crosses the Responses boundary. Once it lives there, ordinary continuation and `previous_response_id` keep it too, not only the search loop.

Two things from your PR were deliberately left out and recorded as follow-ups instead: preserving Google's original `functionCall.id`, and the parallel `FC/FR` ordering in the sidecar loop. Both are real, both are independent of signature carriage, and each deserves its own change.

Closing in favor of the merged commit.

PR #1776, opened 2026-08-15
fix(openai-chat): strip Responses-only encrypted tool annotation

### LINKED ISSUE #1774: [Bug] openai-chat V2 spawn_agent emits empty arguments when message schema contains encrypted:true
### Client or integration

Codex App

burnopenai (OpenAI-compatible Chat Completions upstream through the openai-chat adapter)

### OpenCodex version

2.19.0 (also present by static inspection on current dev at c71c82749a4c65c9be517483f53d22ff1f24405b)

### Endpoint or capability

/v1/responses to openai-chat / Chat Completions translation; V2 collaboration spawn_agent tool calls

### Current behaviour

With V2 collaboration enabled and a routed openai-chat parent, Codex supplies a complete spawn_agent schema whose required message property carries the Responses-only annotation encrypted: true.

OpenCodex preserves required: ["message"], but also forwards the non-standard annotation unchanged in tools[].function.parameters to the Chat Completions upstream. In the reproduced provider/model combination, the upstream then emits a spawn_agent call with zero argument bytes; OpenCodex normalizes that to {}, and Codex rejects it because message is missing. No child agent is created, and the parent retries.

One archived session produced 24 spawn_agent calls: 24/24 had arguments exactly {}, and all 24 corresponding tool results reported missing field message.

The generic openai-chat path still forwards the annotation on current dev; only the Zen-specific schema sanitizer currently drops it.

### Expected behaviour

At the Responses to Chat Completions adapter boundary, OpenCodex should remove the Responses-only encrypted schema annotation before serializing Chat Completions tools, while preserving:

- required: ["message"] and all other supported schema constraints;
- legitimate tool property/definition names that are literally encrypted;
- literal data under keywords such as const, default, enum, and examples.

Native OpenAI Responses passthrough should continue to preserve the annotation. The routed parent should emit a valid, non-empty message and create the child once.

~~~shell
# Environment
# OpenCodex 2.19.0, macOS, V2 collaboration
# Parent: burnopenai/gpt-5.6-sol, reasoning effort xhigh

# 1. Start a fresh Codex App task using the routed parent.
# 2. Ask it to delegate exactly one self-contained task with spawn_agent.
# 3. Observe repeated empty spawn_agent calls and no child creation.

# The incoming schema at the adapter boundary contains:
{
  "type": "object",
  "properties": {
    "message": { "type": "string", "encrypted": true }
  },
  "required": ["message"]
}

# On unmodified 2.19.0, createOpenAIChatAdapter(...).buildRequest(...)
# serializes the same message schema to tools[].function.parameters,
# including encrypted:true.
~~~

A controlled forced-tool A/B used the same upstream, model, prompt, and spawn_agent tool:

~~~text
message schema contains encrypted:true  -> arguments are empty ("" / {})
only the encrypted annotation removed   -> {"message":"Reply exactly: READY"}
~~~

The result was the same for flat and namespace: collaboration tool shapes, which rules out namespace conversion as the differentiator.

### Actual response or error

~~~shell
function_call name=spawn_agent namespace=collaboration arguments={}
function_call_output failed to parse function arguments: missing field message at line 1 column 2

# Redacted session totals:
spawn_agent calls: 24
arguments exactly {}: 24
missing-field-message errors: 24
children created: 0
~~~

### Upstream documentation

No public provider or JSON Schema specification was found for the Codex encrypted annotation. It is identified in this repository as a Responses/ChatGPT-backend-only marker introduced by upstream OpenAI Codex, and the same compatibility class was confirmed and fixed for Google/Kiro in #85 and defensively handled by the Anthropic adapter.

The concrete client requirement is that Responses-private schema metadata must not change required Chat Completions tool-call arguments during protocol translation.

Sanitize tool parameter schemas once in the generic openai-chat adapter before provider-specific Chat Completions formatting. Strip encrypted only when it is a schema keyword, not when it is a property/definition name or literal JSON data. The existing Anthropic implementation already distinguishes schema-name maps and literal-value keywords and may be a useful precedent.

A focused regression should assert that nested annotations are removed, required: ["message"] survives, a property literally named encrypted survives, literal encrypted data survives, and the input schema is not mutated.

I am happy to prepare a focused PR with the adapter fix and behavior-contract regression test.

- #85 confirms the same Responses-only annotation class for other adapters.
- #290 reports the same empty-argument symptom, but was closed because the available trace only established that the provider returned zero argument bytes; it did not isolate why.
- #418 was later closed with an explicit invitation to open a new issue if the newest version still reproduced with a current trace.
- This is not #92: #92 fails after a valid spawn when a routed child cannot read an encrypted NEW_TASK; this issue fails before child creation because spawn_agent.message is absent.
- The archived rollout, configuration, credentials, full prompt, and provider headers are intentionally not attached. The counts and wire shapes above are redacted.
- The affected Codex App session recorded cli_version: 0.148.0-alpha.9; opencodex status reports the configured local Codex CLI as 0.145.0.

### Checks

ZHJay:
## Summary

- Fixes #1774: strip the Responses-only `encrypted` schema annotation when the generic `openai-chat` adapter translates tools to Chat Completions.
- Extracts the existing Anthropic marker sanitizer into a shared adapter helper so translated Chat and Anthropic schemas remove the marker consistently while preserving required fields, literal values, and schema names such as `encrypted` and `__proto__`.
- Leaves the native `openai-responses` passthrough unchanged.
- Adds a behavior-contract regression covering nested annotations, schema name maps, JSON Schema dependency maps, literal-value keywords, prototype-like names, and input non-mutation.

The root cause is protocol translation: `encrypted: true` is private Responses metadata, but it was serialized into Chat Completions tool parameters. For the affected V2 collaboration request, the upstream then returned empty `spawn_agent` arguments; the required `message` field was absent and no child agent was created.

I am happy to prepare/continue the PR review and follow-up fixes.

## Verification

- `bun test tests/openai-chat-hardening.test.ts tests/anthropic-tool-schema.test.ts` — 59 passed, 0 failed.
- `bun run test` — attempted on the final commit; 12,357 passed, 8 skipped, 18 failed in unrelated existing provider-management/concurrency tests (the focused adapter tests and typecheck remain green).
- `git diff --check origin/dev...HEAD` — passed.

## Checklist

- [x] Docs or release notes were updated when needed; this is an internal adapter compatibility fix with no new user-facing configuration.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults; this change does not touch those paths.

Closes #1774

## Review readiness checklist

### DIFF
--- a/src/adapters/anthropic.ts
+++ b/src/adapters/anthropic.ts
@@ -19,6 +19,7 @@ import { parseDataUrl } from "./image";
 import { enforceAnthropicImageLimits } from "./anthropic-image-guard";
 import { normalizeAnthropicImages } from "./anthropic-image-normalize";
 import { normalizeAnthropicOutputSchema } from "./anthropic-output-schema";
+import { stripResponsesOnlyEncryptedMarker } from "./responses-tool-schema";
 import { identifyRoutedModel } from "./identity";
 import { redactSecretString } from "../lib/redact";
 import { CLAUDE_CODE_HEADERS, claudeCodeSessionId } from "./client-fingerprint";
@@ -721,35 +722,8 @@ function toolsToAnthropicFormat(parsed: OcxParsedRequest, toolNames: { toWire: (
   return converted;
 }
 
-// Codex multi-agent v2 stamps a Responses-only `encrypted: true` marker on
-// collaboration tool schemas (openai/codex 5f4d06ef; issue #85). It is an
-// annotation for the ChatGPT backend only. Anthropic input_schema is strict
-// JSON Schema; strip the marker defensively everywhere it can appear as a
-// schema keyword, while preserving properties literally named "encrypted".
-const ENCRYPTED_MARKER_NAME_BAG_KEYS = new Set(["properties", "patternProperties", "$defs", "definitions"]);
-const ENCRYPTED_MARKER_LITERAL_VALUE_KEYS = new Set(["const", "default", "enum", "examples"]);
-
-function stripEncryptedMarker(node: unknown, inNameBag = false): unknown {
-  if (Array.isArray(node)) return node.map(item => stripEncryptedMarker(item));
-  if (!node || typeof node !== "object") return node;
-
-  const out: Record<string, unknown> = {};
-
-  for (const [key, value] of Object.entries(node as Record<string, unknown>)) {
-    if (inNameBag) {
-      out[key] = stripEncryptedMarker(value);
-    } else if (key !== "encrypted") {
-      out[key] = ENCRYPTED_MARKER_LITERAL_VALUE_KEYS.has(key)
-        ? value
-        : stripEncryptedMarker(value, ENCRYPTED_MARKER_NAME_BAG_KEYS.has(key));
-    }
-  }
-
-  return out;
-}
-
 function normalizeAnthropicInputSchema(schema: unknown): Record<string, unknown> {
-  const stripped = stripEncryptedMarker(schema);
+  const stripped = stripResponsesOnlyEncryptedMarker(schema);
   const obj = stripped && typeof stripped === "object" && !Array.isArray(stripped)
     ? stripped as Record<string, unknown>
     : {};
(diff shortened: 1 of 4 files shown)


lidge-jun:
Landed on `dev` as 02728d2bb via #1779, with you as co-author — thank you.

Your design was adopted as-is: keyword-only strip, name bags preserved, literal payloads untouched, null-prototype records so a schema named `__proto__` stays data. One change was needed before merge: the walk is now iterative over an explicit stack, because tool schemas are caller-supplied and a deep one would have overflowed the stack on the request path. A 50,000-deep regression covers that.

Closing this in favor of the merged commit.

Issue #1767, opened 2026-08-15
[Provider compatibility] Anthropic adapter forwards tool_use ids unsanitized, breaking cross-provider history replay (fix already exists in google.ts)


hoonysis:
### Client or integration

Other (Grok Build, via `x-opencodex-grok` custom model over `/v1/chat/completions`)

### Provider or upstream service

anthropic

### OpenCodex version

2.19.0

### Endpoint or capability

/v1/messages (tool call id mapping on history replay)

### Current behaviour

`src/adapters/anthropic.ts` forwards `OcxToolCall.id` / `OcxToolResultMessage.toolCallId` to Anthropic verbatim, with no character-set normalization:

- `anthropic.ts:646` — `toolUses.push({ type: "tool_use", id: tc.id, ... })`
- `anthropic.ts:576` — `tool_use_id: msg.toolCallId`
- `anthropic.ts:674` — `tool_use_id: id` (synthetic "missing tool_result" filler)

Anthropic validates `tool_use.id` against `^[a-zA-Z0-9_-]+$`. When a conversation history contains tool call ids produced by a *different* provider path and that history is replayed to Anthropic (cross-provider model switch inside one client session), any id with a character outside that class produces a hard 400 and the whole conversation becomes unusable on Anthropic models until the history is discarded.

Observed in a real session: of 153 distinct tool call ids in the transcript, 151 were fine (`call-<uuid>-<n>`), and 2 were composite ids that concatenate a Responses-style call id and item id with a literal newline:

```
call_5sNzuhhhfcuN91ysezpcwXjp\nfc_0c71abbccafaad67016a803ba3007487d2afa509a7ca8c9687
call_YXIRuPJVG24WeRlYxDWYhDh4\nfc_0c71abbccafaad67016a803ba3008887d2a75cb919b833062c
```

The embedded `\n` alone is enough to fail validation, so a single legacy id anywhere in the history kills every subsequent Anthropic request in that session.

Notably, **this repository already solves exactly this problem in another adapter.** `src/adapters/google.ts:97` defines `geminiToolCallId()`, whose doc comment states:

> Anthropic's `tool_use.id` only accepts `[a-zA-Z0-9_-]`, so non-conforming characters are mapped to `_`. To keep the mapping injective ... a short hash of the original raw id is appended whenever any character had to be rewritten.

So the Google adapter respects the Anthropic id constraint, while the Anthropic adapter itself does not. This looks like an oversight rather than a design decision.

[...]

PR #1780, opened 2026-08-15
fix(anthropic): normalize tool call ids so cross-provider history replays

ntdatt812:
Refs #1767.

## What happens

Anthropic validates `tool_use.id` against `[a-zA-Z0-9_-]`. `src/adapters/anthropic.ts` forwarded ids verbatim, so a transcript carrying ids minted by a different provider path fails with a 400 — and because the offending id stays in the history, every subsequent Anthropic request in that session fails too, until the conversation is discarded.

The ids in the report concatenate a Responses-style call id and item id with a literal newline. The `\n` alone fails validation:

```
call_5sNzuhhhfcuN91ysezpcwXjp\nfc_0c71abbccafaad67016a803ba3007487d2afa509a7ca8c9687
```

## Change

`google.ts` already solved exactly this for the Antigravity path, where the backend translates Gemini requests into Anthropic `messages`. Rather than copy the logic into a second adapter, this lifts it into `src/adapters/tool-call-id.ts` and points both at it — a correctness-relevant transform living in two places is a transform that eventually drifts.

`google.ts` keeps its local `geminiToolCallId` name as an alias, so its call sites are untouched.

In `anthropic.ts` it is applied to **both sides of the pair**:

- `tool_use.id` on the assistant side
- `tool_result.tool_use_id` on the result side

The transform is deterministic, so ids that were equal at the source stay equal after normalization and the call/result pairing survives. The synthetic "missing tool_result" filler already reuses the ids collected in `toolUseIds`, which are now normalized at the point they are pushed, so that path needed no change.

Ids that already conform are returned unchanged, so ids minted by Anthropic itself round-trip untouched.

## Tests

`tests/anthropic-tool-call-id.test.ts`, 5 cases:

- conforming ids pass through byte-for-byte (`toolu_…`, `call_…`, `call-…`)
- the composite id from the report comes out matching `^[a-zA-Z0-9_-]+$` with no newline
- the transform is deterministic, which is what keeps a call and its result paired
- `call:a` and `call/a` do **not** collapse to the same id — the hash suffix keeps the mapping injective
- an empty id returns `undefined` so the caller omits the field rather than inventing one

Run locally with `bun test`: 5 pass.

For the refactor half, the existing suites are the regression proof — every `google-*` and `anthropic-*` test file still passes unchanged (473 tests across 28 files, 0 fail), which is what shows the shared function behaves identically to the `geminiToolCallId` it replaced. `bun x tsc --noEmit` is clean.

## Note on scope

I did not add an adapter-level test asserting the emitted wire id, because the change is three call sites feeding an already-tested pure function and I did not want to stand up request fixtures for a repo moving as fast as this one. Happy to add one if you would rather have the wiring pinned too.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility of Anthropic and Gemini tool-call and tool-result IDs by normalizing unsupported, empty, or over-length values.
  * Ensured matching IDs across replayed calls and results, including collision handling.
  * Preserved valid IDs and safely handled missing or unmatched results.
  * Prevented invalid tool calls from being emitted in unsupported formats; these are now represented as text when necessary.
* **Tests**
  * Added coverage for ID normalization, pairing, collision handling, length limits, and edge cases.

### DIFF
--- a/src/adapters/anthropic.ts
+++ b/src/adapters/anthropic.ts
@@ -1,5 +1,5 @@
 import type { IncomingMeta, ProviderAdapter } from "./base";
-import { anthropicToolCallId } from "./tool-call-id";
+import { createToolCallIdAllocator, type ToolCallIdAllocator } from "./tool-call-id";
 import { debugDroppedFrame } from "../lib/debug";
 import type {
   AdapterEvent,
@@ -560,7 +560,7 @@ function buildToolNameTransforms(provider: OcxProviderConfig): { toWire: (name: 
   return { toWire: (name) => name, fromWire: (name) => name };
 }
 
-function toAnthropicToolResult(msg: OcxToolResultMessage): Record<string, unknown> {
+function toAnthropicToolResult(msg: OcxToolResultMessage, wireCallId: string): Record<string, unknown> {
   // Anthropic tool_result accepts a string OR content blocks — render images natively
   // (e.g. Codex view_image output) instead of dropping them.
   let content: string | unknown[];
@@ -575,10 +575,15 @@ function toAnthropicToolResult(msg: OcxToolResultMessage): Record<string, unknow
   }
   return {
     type: "tool_result",
-    tool_use_id: anthropicToolCallId(msg.toolCallId) ?? msg.toolCallId,
+    tool_use_id: wireCallId,
     content,
     ...(msg.isError ? { is_error: true } : {}),
   };
+}
+
+function unrepresentableToolCallText(tc: OcxToolCall, wireName: string): string {
+  const args = typeof tc.arguments === "string" ? tc.arguments : JSON.stringify(tc.arguments);
+  return `[tool_use without a usable id: ${wireName}]\n${args}`;
 }
 
 function orphanToolResultText(msg: OcxToolResultMessage): string {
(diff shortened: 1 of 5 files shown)


ntdatt812:
Rebased onto `dev` and retargeted — this was opened against `main` by mistake, which is what the earlier `[WRONG BRANCH]` title was about. Sorry for the noise. The branch carried `release: v2.20.0` from `main`, so I replayed only my own commit with `git rebase --onto origin/dev`; the head is now `34ab113c5` on top of `dev`, and `enforce-target` passes. The failed workflow run you may have seen for this branch is from before that.

Re-verified after the rebase: `bun x tsc --noEmit` clean, `bun test tests/anthropic-tool-call-id.test.ts` 5 pass, and the adapter suites the shared helper touches — `anthropic-reasoning`, `anthropic-thinking-signature`, `anthropic-stream-hardening`, `google-antigravity-replay` — 140 pass / 0 fail. `src/adapters/tool-call-id.ts` is still absent on `dev` and `anthropic.ts` still forwards ids verbatim, so the fix is still needed.

On the readiness checklist: I have not ticked "All CI tests are green on my local testing" because `bun run test` does not complete on my Windows machine — Bun panics partway through, on a clean `dev` checkout too. I explained this in more detail on #1788 rather than repeat it here, and sent the root cause of the local failure count separately as #1805. I would rather leave the box unticked than attest to something the gate cannot verify; tell me if the intended reading is "everything you can run is green" and I will tick it.

PR #1788, opened 2026-08-15
fix(responses): fail closed when a routed provider calls an undeclared tool

### LINKED ISSUE #1700: [Bug]: OpenCode Go native Responses path leaks undeclared apply_patch and shows aborted
## Client or integration

Codex App

## Area

Provider adapter

## Summary

OpenCodex 2.17.0 leaks an undeclared top-level `apply_patch` function call through the native OpenAI Responses passthrough path when using OpenCode Go. Codex App displays `aborted`, while the target file remains unchanged.

Expected behavior: an undeclared routed tool should fail closed with an explicit compatibility error before reaching the Codex executor. The supported editing path should remain `exec -> tools.apply_patch(...)`.

## Reproduction

1. Start OpenCodex on port 10100.
2. Use `exec -> tools.apply_patch(...)` to create a temporary file containing `v1`.
3. Invoke top-level `apply_patch` with a patch changing `v1` to `v2`.
4. Codex App displays `aborted`.
5. The file remains `v1`; content and mtime are unchanged.
6. Run the equivalent patch through `exec -> tools.apply_patch(...)`:
   - an incorrect expected line returns a real `apply_patch verification failed` error;
   - the correct `v1 -> v3` patch succeeds.

The request-visible catalog contains `exec`, `wait`, and `request_user_input`, but no top-level `apply_patch` schema. The session nevertheless receives `function_call(name=apply_patch)`.

## Version

OpenCodex 2.17.0

## Operating system

macOS 27.0, Apple Silicon

## Provider and model

Provider: `opencode-go`

Model: `deepseek-v4-flash`

At reproduction time, the provider default was `openai-chat`, while this model had an explicit `openai-responses` adapter override.

## Logs or error output

- Codex result: `aborted`
- No corresponding `custom_tool_call_output` was emitted.
- OpenCode Go usage record was HTTP 200.
- No upstream `aborted`, OpenCodex `response.failed`, approval denial, or sandbox denial was observed.

No screenshots. Reproduction artifacts were written only under `/tmp`.

## Redacted configuration

```json
{
  "providers": {
    "opencode-go": {
      "adapter": "openai-chat",
      "baseUrl": "https://opencode.ai/zen/go/v1",
      "modelAdapters": {
        "deepseek-v4-flash": "openai-responses"
      }
    }
  }
}
```

## Checks

ntdatt812:
## Summary

Closes #1700.

The bridged paths already refuse a tool call the request never declared: `declaredToolNames` in `src/bridge.ts` turns it into a 502 that names the tool. The native Responses passthrough had no equivalent, so the same call was relayed to the client verbatim.

Codex then received a top-level `function_call(name=apply_patch)` it has no handler for. Under code mode `apply_patch` exists only as a nested `tools.apply_patch(...)` helper declared inside the `exec` description, never as a wire tool — which is exactly why the reporter's request-visible catalog held `exec`, `wait`, and `request_user_input` and nothing else. The turn surfaced as a bare `aborted`: file unchanged, no `custom_tool_call_output`, HTTP 200 upstream, and no error anywhere to explain it.

This adds the missing guard to the passthrough. When a routed provider names a tool the request did not declare, the offending event is replaced with an explicit `response.failed` carrying `code: "undeclared_tool_call"` and the same message the bridged paths use, and the rest of the turn is dropped so a later `response.completed` cannot contradict the terminal already sent.

### This is a gap against an existing decision, not a new policy

The Responses decision log in `structure/04_transports-and-sidecars.md` already settles this question. It records the alternatives that were rejected — prompt guidance alone, and auto-translating an undeclared `apply_patch` into Code Mode — and the one that was chosen:

> 선택한 방식: Retain the allowed wire-name set with the existing bridge maps and fail the turn with an explicit compatibility error **before emitting any undeclared tool item**.

and states the resulting property as:

> **Streaming and non-streaming routed responses now fail closed** with an actionable provider-contract error; providers that emit aliases they never advertised must correct their adapter mapping instead of relying on client abort behavior.

That property held on the bridged paths and not on the passthrough, which is the gap #1700 fell through. The change here makes the passthrough match the documented invariant rather than introduce a different one, so `structure/` needs no correction — it simply becomes true.

The log's adjusted-boundary line is about *restoration*, and stays intact: a native `apply_patch` is still never restored into a `custom_tool_call`. When the request declares it, it passes through in its upstream form exactly as before; the guard only refuses names the catalog never carried.

### Why the outbound body is the source of truth

The guard reads the body that actually goes upstream rather than the parsed internal tool list. The passthrough forwards wire shapes that the internal list flattens or renames — namespaced MCP groups, `additional_tools` items carried inside `input`, the routed custom-tool rewrite — so only the wire names can be compared against what the provider echoes back. Namespaced tools are accepted under either coordinate system, since Codex routes MCP calls by an explicit `namespace` field rather than by parsing the name.

### Scope

- Only client-executed items are checked: `function_call` and `custom_tool_call`. Hosted calls (`web_search_call`, `image_generation_call`, `local_shell_call`, `tool_search_call`) run upstream or carry no tool name.
- `authMode: "forward"` stays unguarded. That is the canonical ChatGPT backend speaking Codex's own protocol, not a routed provider — the same line the routed custom-tool and image-gen rewrites already draw a few lines above.
- An unreadable or empty catalog disables the guard rather than failing every turn.
- Both the streaming relay and the bounded-JSON answer are covered, so `stream: false` is not a way around it.
- The guard runs last in the block-rewrite chain, after image-gen restore, item-id repair, and custom-tool restore, so it compares the names the client will actually receive.

### One existing expectation changed

`tests/responses-custom-tool-repair.test.ts` has a policy matrix pinning that routed custom calls are not *restored* when request policy excludes them. Three of its four cases are unaffected — the tool is in the outbound body, so the guard never fires and the call is still relayed unrestored.

The fourth case ("custom-looking metadata without a declared tool") declares only `ordinary` and has upstream call `exec`, where a `{type: "custom", name: "exec"}` shape sits in `metadata`. That is an undeclared tool by any reading, so it is now refused instead of relayed. The original contract still holds either way: the call never becomes a `custom_tool_call`. I updated that one case and left the other three untouched.

Worth noting the guard is deliberately narrower than `tool_choice`: it asks only whether the name was sent upstream, not whether request policy currently authorizes it. That is why the three `tool_choice` cases are unaffected, and it keeps the guard from second-guessing a policy the restoration path already enforces.

### A refused turn does not become continuation state

The passthrough records completed responses so a later `previous_response_id` can expand from them. A refused turn must not be one of those, or the undeclared call comes back as history in the next request. Both write sites — the JSON path and the untouched upstream stream the inspection branch reads — go through a wrapper that drops a response carrying an undeclared call, and the JSON path records only after the guard passes.

The wrapper inspects the payload rather than sharing a trip flag with the client relay, because the inspection branch can run ahead of the client's pull and a flag would be racy in exactly the case that matters.

Terminal-outcome recording still reports what upstream did. The request was served and the tokens were spent, so quota and upstream-host health should account for it even though the client received a failure. Say the word if you would rather the two agree.

### Why the declared set is a union

The outbound body is not a complete record of the caller's catalog: hosted-tool preference *replaces* a client tool with its hosted form, so a request declaring `image_gen.generate` ships `{type: "image_generation"}` upstream and gets the client's own name back. The declared set is therefore the union of the outbound wire names and the caller's own catalog. Widening only ever makes the guard fire less, and a name declared in neither place — #1700's `apply_patch` — is still refused.

Readability, not size, decides whether the catalog is authoritative. A request that declares no tools authorizes none, so it enforces; only a body this proxy could not parse stands the guard down.

## Verification

All run on Windows against `dev` at `4b950101a`:

- `bun test tests/responses-undeclared-tool-guard.test.ts` — 26 pass, 0 fail (new file).
- `bun test tests/responses-custom-tool-repair.test.ts tests/openai-responses-passthrough.test.ts tests/sse-payload-rewrite.test.ts tests/github-copilot-sse-rewrite.test.ts` — 113 pass, 0 fail.

On the full suite I would rather give you the real picture than a green tick, because `bun run test` does not complete on my Windows machine: Bun 1.3.14 panics partway through with `panic(main thread): index out of bounds: index 0, len 0`. It does that on a clean `dev` checkout too, so it is not this PR.

I know from `devlog/_plan/260815_roadmap_closeout/110_release_readiness.md` that the same command is 0 fail across 789 files on the Linux validation host at this exact commit, with CI green on all three platforms — so what follows is my environment, not the repository's state. I am reporting it rather than hiding it because a bare tick from a machine that cannot run the command would not mean anything.

To get a comparable answer I ran every test file in 16 batches of 50 through `bun scripts/test.ts <files>` — the same wrapper, so still `bun test --isolate` with the sandboxed home, just chunked so one panic cannot truncate the rest. Twice, identical batching, nothing else running: once on `dev` at `4b950101a` with a clean tree, once on this branch.

|  | named failures | panics |
|---|---|---|
| `dev` clean | 164 | 2 |
| this branch | 164 | 2 |

**The two failing sets are identical** — `comm` in both directions is empty — so this branch adds no failure and fixes none. That is the claim I can actually support.

The 164 are Codex-home, catalog, SQLite, and filesystem tests (`codex-inject-*`, `codex-catalog-*`, `Codex SQLite home resolution`, and similar); none touch the Responses path. They are not the cross-file `OPENCODEX_HOME` bleed from devlog 110 either — I checked after reading it: `bun test tests/codex-inject-integration.test.ts` alone gives 7 pass / 16 fail, and `bun test --isolate` on that one file gives exactly the same, with `runInject`'s child process returning `success: false`. Something about spawning that helper is broken on my box.

The clean-`dev` run also hit one extra unnamed failure this branch did not: an unhandled `Cannot find package 'react'` from the dashboard workspace, whose `node_modules` I have not installed.

The Responses-path suites — the ones this change could plausibly break — are green here, and they are listed above.

New coverage in `tests/responses-undeclared-tool-guard.test.ts`. Three cases go end to end through `handleResponses` with the report's own setup — a model pinned to the `openai-responses` adapter and a catalog of `exec`, `wait`, `request_user_input`:

- streaming: the leaked `apply_patch` becomes a named failure and the client never sees `response.completed`;
- non-streaming: the same call is refused with 502 rather than answered;
- a declared `exec` call still completes normally and comes back restored, so the supported `exec -> tools.apply_patch(...)` editing path in the report is unaffected.

Continuation state has its own pair, with a control so the negative cannot pass vacuously: an accepted turn is asserted to expand on a follow-up naming its `previous_response_id`, and a refused turn is asserted not to.

A readable request that declares nothing is covered on both transports, for `tools` absent and for `tools: []`.

The rest are unit-level:

- the reported shape — an undeclared `apply_patch` in `response.output_item.added` becomes `response.failed` plus `[DONE]`, and the following `response.completed` is dropped;
- the snapshot-only shape, where a sparse gateway streams no item events and only ships `response.completed`;
- a declared `apply_patch` is never blocked, including the `custom_tool_call` shape it comes back as;
- a namespaced MCP call echoed under its bare name with a `namespace` field is accepted;
- hosted call types pass through untouched;
- comment frames, `[DONE]`, and unparseable payloads pass through untouched;
- a 5000-character hostile tool name is bounded before it reaches the error message.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup. The only edit outside the new module and its test is wiring in `src/server/responses/core.ts` and the one policy-matrix case described above.
- [x] Docs or release notes were updated when needed. No user-facing configuration or documented behavior changes: this turns a silent failure into a described one, so there is nothing in `docs-site/` to correct.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. No authentication, credential, or workflow surface is touched. The upstream-supplied tool name is the only untrusted value that reaches an error message, and it is length-bounded and JSON-encoded. `bun run privacy:scan` passes.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to detect tool calls that were not declared in the request.
  * Undeclared tool calls now return a clear failure response instead of being executed.
  * Streaming responses stop safely with an explicit failure event when an undeclared tool call is detected.
  * Non-streaming responses now receive the same validation.
  * Prevented rejected responses from being reused in continuation flows.
  * Preserved existing behavior for declared tools and upstream-executed tools.

* **Tests**
  * Expanded coverage for streaming, non-streaming, namespaced, malformed, and edge-case tool requests.

### DIFF
--- a/src/server/relay.ts
+++ b/src/server/relay.ts
@@ -603,6 +603,13 @@ export type SseInspectorHandlers = {
   onTerminal?: (status: ResponsesTerminalStatus, httpStatusOverride?: number) => void;
   logCtx?: RequestLogContext;
   onCompletedResponse?: (response: { id?: unknown; output?: unknown; status?: unknown }) => void;
+  /**
+   * Every parsed SSE payload, delivered BEFORE any onCompletedResponse derived from that same
+   * payload. A caller that must decide on the whole turn -- not just its terminal snapshot --
+   * needs to see the incremental events, because a stream can announce an item and then close
+   * with an empty `output`.
+   */
+  onParsedPayload?: (payload: unknown) => void;
   onFirstOutput?: () => void;
   /**
    * Provider-scoped compatibility: persist the completed snapshot under the
@@ -786,6 +793,11 @@ export function createSseInspector(handlers: SseInspectorHandlers): SseInspector
     if (!reported && handlers.logCtx) {
       inspectResponseLogSsePayloadParsed(handlers.logCtx, payload, parsed);
     }
+    // Before any terminal handling: a consumer deciding on the whole turn must observe this
+    // payload even when the terminal snapshot that follows no longer mentions it.
+    if (handlers.onParsedPayload && parsed !== undefined) {
+      try { handlers.onParsedPayload(parsed); } catch { /* inspection must never throw into the pump */ }
+    }
     reportFirstOutput.parsed(parsed);
     const status = terminalStatusFromParsed(parsed);
     if (status) sawTerminal = true;
@@ -953,6 +965,8 @@ export type InspectionConsumerOptions = {
   now?: () => number;
   /** Forward provider-scoped response-id pinning to the owned inspector. */
   pinCompletedResponseIdToFirstSeen?: boolean;
+  /** Observe every parsed SSE payload on the inspection side; see SseInspectorHandlers. */
+  onParsedPayload?: (payload: unknown) => void;
   /** Test seam for proving both public consumers dispose their owned inspector. */
   inspectorFactory?: (handlers: SseInspectorHandlers) => SseInspector;
 };
@@ -1108,6 +1122,7 @@ export function consumeForInspection(
     onTerminal,
     logCtx,
     onCompletedResponse,
+    onParsedPayload: options?.onParsedPayload,
     onFirstOutput,
     pinCompletedResponseIdToFirstSeen: options?.pinCompletedResponseIdToFirstSeen,
   });
@@ -1158,6 +1173,7 @@ export function consumeForResponseLogMetadata(
   const inspector = (options?.inspectorFactory ?? createSseInspector)({
     logCtx,
     onCompletedResponse,
+    onParsedPayload: options?.onParsedPayload,
     onFirstOutput,
     pinCompletedResponseIdToFirstSeen: options?.pinCompletedResponseIdToFirstSeen,
   });
--- a/src/server/responses-undeclared-tool-guard.ts
+++ b/src/server/responses-undeclared-tool-guard.ts
@@ -1,0 +1,153 @@
+import { namespacedToolName } from "../types";
+import { sseDataPayload, type SseBlockRewrite } from "./sse-payload-rewrite";
+
+/**
+ * Item types the CLIENT executes by name. Hosted calls (`web_search_call`,
+ * `image_generation_call`, `local_shell_call`, `tool_search_call`, …) are run upstream or
+ * carry no tool name, so they are never matched against the request catalog.
+ */
+const CLIENT_EXECUTED_CALL_TYPES = new Set(["function_call", "custom_tool_call"]);
+
+/** An upstream-supplied name reaches the error message; keep it bounded. */
+const MAX_REPORTED_NAME_CHARS = 100;
+
+export const UNDECLARED_TOOL_CALL_ERROR_CODE = "undeclared_tool_call";
+
+function isPlainObject(value: unknown): value is Record<string, unknown> {
+  return !!value && typeof value === "object" && !Array.isArray(value);
+}
+
+function addWireToolName(names: Set<string>, tool: unknown, namespace?: string): void {
+  if (!isPlainObject(tool) || typeof tool.name !== "string" || tool.name.length === 0) return;
+  names.add(tool.name);
+  // Codex routes MCP calls by an explicit `namespace` field, so the same tool is reachable
+  // as a bare inner name or as the flattened form; accept both rather than guess which
+  // coordinate system this provider echoes back.
+  if (namespace) names.add(namespacedToolName(namespace, tool.name));
+}
+
+function addWireToolSpecs(names: Set<string>, specs: unknown): void {
+  if (!Array.isArray(specs)) return;
+  for (const spec of specs) {
+    if (!isPlainObject(spec)) continue;
+    if (spec.type === "namespace" && Array.isArray(spec.tools)) {
+      const namespace = typeof spec.name === "string" ? spec.name : undefined;
+      for (const inner of spec.tools) addWireToolName(names, inner, namespace);
+      continue;
+    }
+    addWireToolName(names, spec);
+  }
+}
+
+/**
+ * Tool names the OUTBOUND Responses body actually declared.
+ *
+ * This reads the body that goes upstream rather than the parsed internal tool list: the
+ * passthrough forwards wire shapes (namespaced MCP groups, `additional_tools` items carried
+ * inside `input`, routed custom-tool rewrites) that the internal list flattens or renames, and
+ * only the wire names can be compared against what the provider echoes back.
+ */
+export function collectDeclaredWireToolNames(body: unknown): Set<string> {
+  const names = new Set<string>();
+  if (!isPlainObject(body)) return names;
+  addWireToolSpecs(names, body.tools);
+  if (Array.isArray(body.input)) {
+    for (const item of body.input) {
+      if (isPlainObject(item) && item.type === "additional_tools") addWireToolSpecs(names, item.tools);
+    }
+  }
+  return names;
+}
+
+function undeclaredNameInItem(item: unknown, declared: ReadonlySet<string>): string | undefined {
+  if (!isPlainObject(item)) return undefined;
+  if (typeof item.type !== "string" || !CLIENT_EXECUTED_CALL_TYPES.has(item.type)) return undefined;
+  const name = item.name;
+  if (typeof name !== "string" || name.length === 0) return undefined;
+  if (declared.has(name)) return undefined;
+  if (typeof item.namespace === "string" && declared.has(namespacedToolName(item.namespace, name))) {
+    return undefined;
+  }
+  return name;
+}
+
+/** First undeclared client tool named by a Responses SSE payload, or undefined. */
+export function undeclaredToolCallName(
+  payload: unknown,
+  declared: ReadonlySet<string>,
+): string | undefined {
+  if (!isPlainObject(payload)) return undefined;
+  if (payload.type === "response.output_item.added" || payload.type === "response.output_item.done") {
+    return undeclaredNameInItem(payload.item, declared);
+  }
+  // Sparse gateways skip incremental items and only ever ship the terminal snapshot.
+  if (payload.type === "response.completed" || payload.type === "response.incomplete") {
+    return undeclaredToolCallNameInResponse(payload.response, declared);
+  }
+  return undefined;
+}
+
+/** First undeclared client tool in a Responses object's `output` array, or undefined. */
+export function undeclaredToolCallNameInResponse(
+  response: unknown,
+  declared: ReadonlySet<string>,
+): string | undefined {
+  if (!isPlainObject(response) || !Array.isArray(response.output)) return undefined;
+  for (const item of response.output) {
+    const name = undeclaredNameInItem(item, declared);
+    if (name !== undefined) return name;
+  }
+  return undefined;
+}
+
+export function undeclaredToolCallMessage(name: string): string {
+  const reported = name.slice(0, MAX_REPORTED_NAME_CHARS);
+  return `routed provider emitted undeclared client tool "${reported}"; only request-declared tools may be called`;
+}
+
+function failedBlocks(name: string, newline: string): readonly string[] {
+  const failure = {
+    type: "upstream_error",
+    code: UNDECLARED_TOOL_CALL_ERROR_CODE,
+    message: undeclaredToolCallMessage(name),
+  };
+  const payload = JSON.stringify({
+    type: "response.failed",
+    response: { status: "failed", error: failure, last_error: failure },
+  });
+  return [`event: response.failed${newline}data: ${payload}`, "data: [DONE]"];
+}
+
+/**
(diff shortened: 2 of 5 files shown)


ntdatt812:
Thanks — all three were real. Pushed `c994d47e6`.

**1. Empty catalog vs unreadable catalog.** You are right, and my original comment even said the wrong thing out loud: I wrote "an empty catalog disables the guard" and justified it with "a request whose tools this proxy could not read". Those are two different conditions and I collapsed them. A request that declares no tools authorizes none, so every client-executed call it gets back is undeclared. Readability and the name set are now separate state — an unparseable body still stands down, a readable one enforces even at zero tools. Covered for `tools` absent and `tools: []`, streaming and JSON.

**2 and 3. Refused turns entering continuation state.** Both correct, and the second is worse than I had it. My description called the inspection-tee divergence a "known limitation" and reasoned about it only as accounting. That was too narrow — `rememberPassthroughResponse` is continuation state, so a refused turn could be expanded from by a later `previous_response_id` replay and reintroduce the undeclared call as history. That is a correctness bug, not a logging wart.

Both writes now go through a wrapper that drops a response carrying an undeclared call, and the JSON path records only after the guard passes. I checked the payload inside the wrapper rather than sharing a trip flag with the client relay, because the inspection branch can run ahead of the client's pull — a flag would have been racy in exactly the case that matters.

Regression is in `tests/responses-undeclared-tool-guard.test.ts`, and it carries its own control: an accepted turn is asserted to expand on the follow-up, so the refused-turn assertion cannot pass just because expansion never works in the harness.

**One thing your fix surfaced that neither of us had.** Making the empty catalog enforce broke `handleResponses does not restore a custom image tool replaced by hosted preference`. Hosted-tool preference *replaces* a client tool in the outbound body: the request declares `image_gen.generate`, ships `{type:"image_generation"}` upstream, and gets the client's own name back. So the outbound body is not a complete record of the caller's catalog, and the guard would have refused a tool the client really declared. The declared set is now the union of the outbound wire names and the caller's own catalog — widening only ever makes the guard fire less, and a name declared in neither place is still refused.

**One I did not apply.** Terminal-outcome recording still reports what upstream did. The request was served and the tokens were spent, so quota and upstream-host health should account for it even though the client received a failure. Persisting the response was the actual defect and that is fixed; suppressing the accounting would make a real spend invisible. Happy to change it if the maintainers prefer the two to agree.

Verification: `bun run typecheck` clean; `bun test tests/responses-undeclared-tool-guard.test.ts` 26 pass; `bun test` across `responses-custom-tool-repair`, `openai-responses-passthrough`, `sse-payload-rewrite`, `github-copilot-sse-rewrite` 113 pass / 0 fail. The 4 failures in `responses-state.test.ts` on my machine are `EPERM: symlink` and are present on a clean `dev` checkout too.

ntdatt812:
A note on the review-readiness checklist, since this PR is sitting at 0/4 and I would rather explain why than leave it looking abandoned.

I have not ticked **"All CI tests are green on my local testing"**, because on my Windows machine it is not true and I do not want to attest to something the gate cannot check. `bun run test` does not complete here at all — Bun 1.3.14 panics partway through — and it does the same on a clean `dev` checkout, so it is not this branch.

What I could do instead was run every test file in 16 batches of 50 through `bun scripts/test.ts <files>`, so a panic in one batch cannot truncate the rest, once on clean `dev` and once on this branch with identical batching. Both produced 164 named failures and the two sets are identical — `comm` empty in both directions. So this branch adds no failure and fixes none. The Responses suites this change could plausibly break are green and named in the description.

While chasing why those 164 exist I found the cause, and it is not this PR's business: `createIsolatedTestEnvironment` gives the sandbox a `USERPROFILE` with no `AppData`, .NET's known-folder API returns an empty string rather than an error for a folder that does not exist, and every Codex coordinator lookup then refuses. I sent that separately as #1805 — with it applied, 104 of the 164 go away. The rest are `EPERM` on `symlink` and 5s timeouts from PowerShell spawns, which are my machine.

So: boxes 2 and 4 I can tick honestly whenever you want them ticked, box 3 once CodeRabbit re-reviews the fixes I pushed in `c994d47e6`, and box 1 is the one I am stuck on. Two things would unstick it — merging #1805 and my machine getting fast enough to stop timing out, or you telling me the local-CI box means "everything you can run is green" rather than literally the whole suite. I did not want to guess which.

Separately, `Cross-platform CI` and `React Doctor` on this PR are `action_required`: as a fork contributor I cannot start repository CI, so the real cross-platform answer needs a maintainer to approve the run. Happy to act on whatever it reports.

ntdatt812:
CI caught a regression I introduced in `c994d47e6` while applying review feedback. Fixed in `cfbb3cdec`; thanks for approving the run, it found something my local checks did not.

**What broke.** One of the three review findings was that an empty tool catalog should still be enforced — a request that declares no tools authorizes none, so any client-executed call coming back is undeclared. That is a clean argument and I accepted it. It is also wrong for the passthrough, and this repository already had a test saying so.

`tests/github-copilot-stream-contract.test.ts` sends `{model, input, stream}` with **no `tools` field at all**, and Copilot answers with a `custom_tool_call` for `apply_patch`. The client understands that call; the proxy simply has no catalog to check it against. Enforcing there replaced the turn with `response.failed`, so the client never saw `response.completed`. The same shape broke the DeepSeek terminal-repair and item-id-repair contracts — six tests on macOS and three of the four shards.

So the activation condition is back to "at least one declared name". An unreadable body lands in the same place, since it also yields no names, which is what the readable-vs-empty distinction was reaching for.

**My mistake, and it is worth naming precisely.** After changing the activation rule I only re-ran the suites adjacent to the change — the passthrough and rewrite files — and they were green. The tests that disprove the new rule live in provider-contract files I had no reason to associate with a tool-catalog condition. A broader run before pushing would have caught it, and CI is the only reason it did not reach a reviewer as a silent behaviour change.

**Kept from that commit**, since both stand up: the union with the caller's own catalog (hosted-tool preference rewrites the outbound body, so it is not a complete record of what the caller declared), and keeping a refused turn out of `previous_response_id` replay state.

The empty-catalog cases in the guard's own test file asserted the behaviour this reverts. They now pin the opposite on both transports, with the Copilot contract named as the reason, so this cannot be re-argued into the code without the counter-example showing up.

Verified locally: `github-copilot-stream-contract`, `deepseek-inbound-wire`, `deepseek-responses-item-id-repair`, `responses-undeclared-tool-guard`, `responses-custom-tool-repair` all pass, `bun run typecheck` clean. The macOS job also reported a `Bun runtime crash repeated on the macOS suite` line alongside the assertion failures — I cannot reproduce that on Windows and do not believe it is related, but flagging it rather than assuming.

PR #1792, opened 2026-08-15
fix(server): add per-provider upstreamHttpVersion to pin Bun fetch HTTP version

### LINKED ISSUE #1668: [Bug] Streaming requests to opencode.ai hang with HTTP/2 SSE (200s timeout → 502/499, Codex stuck on 'thinking')

flyinsz:
## Summary

Add an optional per-provider `upstreamHttpVersion` config field (`auto | http1.1 | h1 | http2 | h2`) that pins the HTTP version used for upstream provider requests, fixing issue #1668.

**Background.** Bun's `fetch` negotiates HTTP/2 via TLS ALPN by default. Some Cloudflare-fronted SSE endpoints (e.g. `opencode.ai/zen/go/v1`) hang on HTTP/2 streaming responses: the proxy waits the full upstream timeout, then reports `502 upstream_server_error` / `499 client_closed_request`, while the Codex client stays on "thinking". Non-streaming requests work over both HTTP versions.

**Change.** `providerFetch` now forwards `provider.upstreamHttpVersion` to Bun's non-standard `protocol` fetch init (`BunFetchRequestInit.protocol`). Pinning `"http1.1"` restores streaming on the affected endpoints. Absent or `"auto"` keeps Bun's default negotiation, so existing providers are completely untouched. Only `https:` targets are pinned, matching Bun's constraint, and the `codexWsUpstream` (responses_websockets) path is unaffected. The value is validated by `providerConfigSchema` (zod enum), so invalid configs fail at load time.

**Local verification** against `opencode.ai/zen/go/v1/chat/completions` (stream=true, model deepseek-v4-flash):
- default Bun fetch → HTTP 200 + SSE headers, but the response body stalls (no chunks) — reproduces the hang
- `protocol: "http1.1"` → SSE chunks stream normally
- `protocol: "http2"` → fails with `HTTP2Unsupported`

Also verified as a live workaround: a local node `https` reverse proxy (HTTP/1.1) in front of the same endpoint streams correctly, confirming the root cause is HTTP/2 SSE negotiation, not proxy/DNS/network.

## Verification

- `bun run test tests/upstream-http-version.test.ts` — 12 pass / 0 fail
- Full `bun run test` — passed (no regressions; pre-existing lab-live/lab-automation failures reproduce identically on the base dev commit)
- Manual protocol comparison script against opencode.ai (see summary above)

## Review readiness checklist

---

Fixes #1668
Related: #1693 (roadmap item 030 — per-provider `upstreamHttpVersion` + `responseDelivery`)

## Summary by CodeRabbit

- **New Features**
  - Added an optional provider setting to select automatic HTTP negotiation, HTTP/1.1, or HTTP/2 for HTTPS connections.
  - Supports common aliases for HTTP/1.1 and HTTP/2.
  - Provider settings can be updated, cleared, and viewed through management interfaces.
  - Existing behavior is preserved for automatic settings, non-HTTPS URLs, and invalid targets.

- **Bug Fixes**
  - Invalid HTTP-version settings are now rejected without changing existing provider configuration.

- **Tests**
  - Added coverage for protocol selection, aliases, validation, persistence, URL handling, defaults, and request configuration.


flyinsz:
All three CodeRabbit suggestions addressed in 2366c7f4:

1. **Apply pin when init is absent** — `withUpstreamHttpVersion` no longer early-returns on a missing init; `providerFetch(provider)(url)` without an init now still pins the protocol. Added a no-init propagation test.
2. **Validate `upstreamHttpVersion` in config paths** — added `z.enum(["auto","http1.1","h1","http2","h2"]).optional()` to `providerConfigSchema` in `src/config.ts`, so invalid values fail config load.
3. **Type the test fetch override** — typed the stub as `typeof globalThis.fetch`.

Verification: typecheck passes; `tests/upstream-http-version.test.ts` now 12 pass / 0 fail.

Wibias (changes requested):
The fetch-side transport change looks sound, but the new provider config field is not integrated consistently enough to merge yet.

Required before re-review:

1. Validate `upstreamHttpVersion` at every management write boundary, not only in `providerConfigSchema`. Today `POST /api/providers` can persist an arbitrary value because `providerManagementConfigError()` does not validate this field, while a later config load is strict. Please use one shared enum/validator so POST/load cannot disagree.

2. Support the field through the provider management surface. `PATCH /api/providers/:name` currently does not recognize/set/clear it, and the provider GET/DTO surfaces do not expose it. Add set/clear handling plus the corresponding read projection and persistence coverage.

3. Add regression tests covering valid/invalid POST, PATCH set/clear, persistence/reload, and the public provider DTO/GET surface. The fetch propagation tests can stay as-is.

4. Refresh onto current `dev` and run CI on the resulting head.

The HTTP-version pin itself does not need redesign; this request is about completing the config contract around it.

flyinsz:
Thanks for the review — all four points are addressed and the branch is rebased onto current dev (e6354c20). Push `8f41eb78`.

1. **Shared enum/validator** — added `UPSTREAM_HTTP_VERSION_VALUES` in `src/types.ts` and a single `upstreamHttpVersionConfigError()` in `src/config.ts`. It is now enforced at every management write boundary: `providerManagementConfigError()` (covers `POST /api/providers` and provider reload) calls it, and the zod load schema uses `z.enum(UPSTREAM_HTTP_VERSION_VALUES)`, so POST/load/PATCH can no longer disagree.

2. **Provider management surface** — `applyProviderPatchFields` now recognizes `upstreamHttpVersion`: a valid value sets it, `null`/`""` clears it. `GET /api/providers` rows and `safeConfigDTO` both expose the field, and it persists through save/reload.

3. **Regression tests** — added a `provider upstreamHttpVersion management contract (#1668)` describe block in `tests/management-provider-validation.test.ts` covering: valid POST (live + disk + GET row), invalid POST rejection without persistence, PATCH set/clear with live + disk checks, `safeConfigDTO` projection, and a direct `providerManagementConfigError` unit check for invalid values. The fetch propagation tests are untouched.

4. **Rebase + CI** — branch rebased onto current `dev`; `bun run typecheck` passes and the full suite shows no regressions (the only failures are the pre-existing lab-live/lab-automation environment failures — "live result lacks trusted execution receipt" — which reproduce identically on the base dev commit).

Also resolved the remaining CodeRabbit finding: the test helper now uses `type TestProvider = OcxProviderConfig & { fetch?: typeof globalThis.fetch }` so the stub fetch passes strict TS checks.

Marked the PR ready for review and took it out of draft. Happy to adjust anything else.

flyinsz:
Heads-up for the maintainers: the updated branch now trips the `unsponsored_surface` gate.

**Why:** review point 1 (validate `upstreamHttpVersion` at every management write boundary) is implemented inside `providerManagementConfigError()`, which lives in `src/server/auth-cors.ts` — one of the repository's RESTRICTED_FILES (auth/credential surface). The shared validator itself is in `src/config.ts`; `auth-cors.ts` only gained the call site, no auth logic changed. Since the author has no push permission and the PR is not labeled `maintainer-sponsored`, hygiene and the target gate both fail with `unsponsored_surface` and the PR was auto-reverted to draft with `intake: hygiene-blocked`.

**Request:** could a maintainer add the `maintainer-sponsored` label? This is exactly the surface @Wibias's review asked to touch, so the change is maintainer-directed. CI itself is green on the changed lines (typecheck + 82 relevant tests pass; the only full-suite failures are the pre-existing lab-live environment ones).

The ready-to-review state stands — the branch is `8f41eb78` on current `dev` with all four review points addressed.

lidge-jun:
Security review of the restricted surface, so this can leave `intake: hygiene-blocked`.

This PR touches `src/server/auth-cors.ts`, which is on the restricted list in `.github/scripts/pr-sponsored-surface.cjs`. That hunk is load-bearing and cannot simply be dropped: the file owns `providerManagementConfigError()` and `safeConfigDTO()`, so removing it would leave the new provider option unvalidated on every management write path.

Reviewed as a validation-boundary change:

- The `auth-cors.ts` diff adds one `upstreamHttpVersionConfigError()` call inside `providerManagementConfigError()` and exposes the enum value in `safeConfigDTO()`. It changes no existing accept/reject outcome and touches no credential material — the added error path can only *reject* a value that was previously accepted unvalidated.
- The DTO addition surfaces an enum (`auto`/`http1.1`/`h1`/`http2`/`h2`) or nothing. No secret, key, or account identifier is involved.
- Auth admission itself is untouched: no change to `requireManagementAuth`, token comparison, CORS origin handling, or the data/management plane separation.

Applying `maintainer-sponsored` and removing `intake: hygiene-blocked` on that basis.

Separately, the null-persistence blocker CodeRabbit raised is now fixed at commit `7f5c22bcd`: POST canonicalizes `upstreamHttpVersion: null` to absent on the object it persists, and the loader accepts null and transforms it to undefined for configs the old path already wrote. Regressions cover the POST → disk → reload round trip and a pre-existing null on disk. The option is now documented in `docs-site` as well.

PR #1793, opened 2026-08-15
fix: allow slash custom model IDs and live Cloudflare Workers AI search

yansigit:
## Summary

- Custom model IDs may now contain `/`, so operators can store native ids like `openai/gpt-5.5`. Codex still sees the one-slash encoded slug (`provider/openai-gpt-5.5`). Display names remain slash-rejected.
- Encoded-slug collisions are rejected on API, CLI, and GUI, including `defaultModel`-only providers.
- Route-time fail-safe: if a later live cache makes a custom encoded slug collide with another known native id, routing throws `ambiguous model id` instead of silently sending the native hyphen id.
- `decodeRoutedModelIdOrThrow()` materializes `knownIds` once so a single-use iterable still decodes.
- `ocx models remove` fails when an encoded selector matches more than one custom-model row and requires the custom-model UUID.
- Named `cloudflare-workers-ai` live discovery now uses official `GET /ai/models/search?format=openrouter` instead of the 405 `GET /ai/v1/models`.

Root cause: the custom-model write path banned `/` even though the slug codec already handled native slash ids, and Workers AI live discovery hit the OpenAI-compatible `/ai/v1/models` URL that Cloudflare rejects.

## Verification

- Rebased onto current `dev` (`e1769b5e29`). Head `facbb0c23`. Merge-base matches; branch is 0 behind / 5 ahead.
- `bun run typecheck` — clean
- Focused: `bun test tests/slug-codec.test.ts tests/cli-models.test.ts tests/catalog-input-modality-enum.test.ts ./gui/tests/provider-model-custom-add.test.tsx` — 79 pass, 0 fail
- Full suite on `5e0b2abaa` — 12415 pass, 8 skip, 0 fail

## Screenshot

Custom model add accepts openai/gpt-5.5

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Custom model IDs can now include `/` characters.
  * Added live model discovery for Cloudflare Workers AI.
  * Improved handling of provider-routed and namespaced model IDs.

* **Bug Fixes**
  * Prevented ambiguous encoded model IDs from being added or routed incorrectly.
  * Model removal now reports ambiguous matches instead of selecting one automatically.
  * Improved model discovery filtering, prefix handling, and path validation.

* **Documentation**
  * Clarified supported custom model ID formats.

### DIFF
--- a/gui/src/components/provider-workspace/ProviderModels.tsx
+++ b/gui/src/components/provider-workspace/ProviderModels.tsx
@@ -7,6 +7,7 @@ import { useEffect, useMemo, useRef, useState } from "react";
 import { useT } from "../../i18n/shared";
 import type { WorkspaceItem } from "../../provider-workspace/catalog";
 import { filterModels } from "../../provider-workspace/report";
+import { encodedModelIdCollides } from "../../../../src/providers/slug-codec";
 
 export default function ProviderModels({
   item,
@@ -48,12 +49,19 @@ export default function ProviderModels({
   const selectedSet = useMemo(() => new Set(selectedModels), [selectedModels]);
   const configuredModels = useMemo(() => item.models ?? [], [item.models]);
   const trimmedCustomModelId = customModelId.trim();
+  const knownModelIds = [
+    ...availableModels,
+    ...customModelIds,
+    ...configuredModels,
+    ...(item.defaultModel ? [item.defaultModel] : []),
+  ];
   const customModelInvalid = !customModelsReady
     || !trimmedCustomModelId
     || availableModels.includes(trimmedCustomModelId)
     || customModelIds.includes(trimmedCustomModelId)
     || configuredModels.includes(trimmedCustomModelId)
-    || item.defaultModel === trimmedCustomModelId;
+    || item.defaultModel === trimmedCustomModelId
+    || encodedModelIdCollides(trimmedCustomModelId, knownModelIds);
   const models = useMemo(
     () => filterModels(availableModels, item.defaultModel, query, configuredModels, customModelIds, hasLiveModels),
     [availableModels, item.defaultModel, query, configuredModels, customModelIds, hasLiveModels],
--- a/gui/src/components/provider-workspace/ProviderModels.tsx
+++ b/gui/src/components/provider-workspace/ProviderModels.tsx
@@ -50,7 +50,6 @@ export default function ProviderModels({
   const trimmedCustomModelId = customModelId.trim();
   const customModelInvalid = !customModelsReady
     || !trimmedCustomModelId
-    || trimmedCustomModelId.includes("/")
     || availableModels.includes(trimmedCustomModelId)
     || customModelIds.includes(trimmedCustomModelId)
     || configuredModels.includes(trimmedCustomModelId)
--- a/gui/tests/provider-model-custom-add.test.tsx
+++ b/gui/tests/provider-model-custom-add.test.tsx
@@ -114,7 +114,7 @@ test("quick-add submits the trimmed model id for the current provider", async ()
   await act(async () => { root.unmount(); });
 });
 
-test("quick-add blocks existing and namespaced model ids", async () => {
+test("quick-add blocks existing ids but allows namespaced model ids", async () => {
   let requests = 0;
   globalThis.fetch = (async (_input, init) => {
     if (!init?.method || init.method === "GET") return Response.json([]);
@@ -126,7 +126,7 @@ test("quick-add blocks existing and namespaced model ids", async () => {
   await enterModelId(input, "claude-opus-5");
   expect(addButton.disabled).toBe(true);
   await enterModelId(input, "vendor/model");
-  expect(addButton.disabled).toBe(true);
+  expect(addButton.disabled).toBe(false);
   expect(requests).toBe(0);
 
   await act(async () => { root.unmount(); });
--- a/gui/tests/provider-model-custom-add.test.tsx
+++ b/gui/tests/provider-model-custom-add.test.tsx
@@ -132,6 +132,23 @@ test("quick-add blocks existing ids but allows namespaced model ids", async () =
   await act(async () => { root.unmount(); });
 });
 
+test("quick-add blocks a slash id that encodes to an existing native id", async () => {
+  let requests = 0;
+  globalThis.fetch = (async (_input, init) => {
+    if (!init?.method || init.method === "GET") return Response.json([]);
+    requests += 1;
+    return Response.json({ id: "unexpected" }, { status: 201 });
+  }) as typeof fetch;
+  const colliding = { ...item, models: ["openai-gpt-5.5"], defaultModel: "openai-gpt-5.5" } as WorkspaceItem;
+  const { root, input, addButton } = await mountProviderModels(["openai-gpt-5.5"], undefined, colliding);
+
+  await enterModelId(input, "openai/gpt-5.5");
+  expect(addButton.disabled).toBe(true);
+  expect(requests).toBe(0);
+
+  await act(async () => { root.unmount(); });
+});
+
 test("quick-add keeps the model id when the server rejects it", async () => {
   globalThis.fetch = (async (_input, init) => (
     !init?.method || init.method === "GET"
--- a/src/claude/agents-inject.ts
+++ b/src/claude/agents-inject.ts
@@ -85,7 +85,7 @@ function entryParts(entry: string, config: OcxConfig): { alias: string; id: stri
     const provider = entry.slice(0, slash);
     const prov = hasOwnProvider(config.providers, provider) ? config.providers[provider] : undefined;
     const id = prov
-      ? decodeRoutedModelId(entry.slice(slash + 1), knownModelIdsForProvider(provider, prov))
+      ? decodeRoutedModelId(entry.slice(slash + 1), knownModelIdsForProvider(provider, prov, config))
       : entry.slice(slash + 1);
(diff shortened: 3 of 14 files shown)


yansigit:
The live dashboard capture is ready locally (Cloudflare Workers AI → Models → Add custom model, `openai/gpt-5.5` typed, Add enabled).

Wibias (changes requested):
The Cloudflare Workers AI discovery change looks reasonable, but the slash-custom-model change introduces an ambiguous slug collision that can route a selection to a different native model.

Required before re-review:

1. Reject or otherwise safely resolve collisions between an encoded slash custom id and another known native provider id. Example: native `openai-gpt-5.5` and custom `openai/gpt-5.5` both expose `provider/openai-gpt-5.5`. `decodeRoutedModelId()` intentionally prefers an exact native-id match, so the custom row can be shown while the wire request resolves to the other native model. Add a regression covering this exact case, plus another multi-slash/hyphen collision such as `a/b-c` vs `a-b/c`.

2. Fix `ocx models remove` ambiguity. `slugEquals()` can match more than one custom-model row in hand-edited/legacy configs, and `findIndex()` then removes whichever appears first. Collect all matches and fail with an explicit ambiguity error requiring the custom-model UUID when there is more than one. Add the collision regression requested by the existing review thread.

3. Keep API/GUI/CLI duplicate-admission rules aligned so a model that cannot be routed unambiguously cannot be created through one surface but rejected through another.

4. Refresh onto current `dev` and run the real GitHub Actions workflows on the resulting head before re-review.

I did not find a blocking issue in the Cloudflare discovery portion itself.

Wibias (changes requested):
Re-review on current head `5e0b2ab`. The original collision and `models remove` blockers are fixed, but two routing-safety gaps remain:

[P1] `defaultModel` is missing from the shared collision set. The GUI includes `item.defaultModel` when checking encoded collisions, but CLI/API call `knownModelIdsForProvider()`, which currently unions `prov.models`, registry seeds/hints, live cache, and custom models but not `prov.defaultModel`. A provider whose only known native id is `defaultModel: "openai-gpt-5.5"` can therefore still admit custom `openai/gpt-5.5` through CLI/API even though the GUI blocks it; both expose the same Codex-facing slug. Please include `prov.defaultModel` in `knownModelIdsForProvider()` and add defaultModel-only API + CLI regressions so all three admission surfaces stay aligned.

[P1] Collision safety still depends on the live-model cache state. On a cold start, a live-discovery provider can admit custom `openai/gpt-5.5` while the cache is empty. If discovery later learns native `openai-gpt-5.5`, the same Codex-facing slug becomes ambiguous and `decodeRoutedModelId()` will prefer the exact native id, silently changing which model is sent upstream. Write-time validation cannot fully prevent this because the provider's live catalog can change after the write. Please add a runtime fail-safe that detects a custom encoded slug colliding with another currently known native id and fails explicitly rather than routing to a different model. Add a regression for: custom admitted while cache is empty -> live cache later gains the colliding native id -> routing the encoded selector returns an ambiguity error, not the native hyphen model.

The previous `ocx models remove` ambiguity fix, API/CLI/GUI collision checks for already-known ids, Cloudflare discovery changes, and current-dev refresh otherwise look good. After these two points are fixed, rerun the full exact-head CI.

yansigit:
@Wibias addressed in 41abdb93c (still on current `dev` @ e1769b5e29):

1. `knownModelIdsForProvider()` now unions `prov.defaultModel`. CLI/API collision checks therefore match the GUI for a defaultModel-only provider. Added defaultModel-only API + CLI regressions.

2. Route-time fail-safe: `decodeRoutedModelIdOrThrow()` throws `ambiguous model id` when a custom encoded slug later collides with another currently known native id (including a live cache that filled in after write). `routeModel` and Claude agent-inject use that path. Regression: custom `openai/gpt-5.5` admitted on an empty cache, then live cache gains `openai-gpt-5.5`, then `zenmux/openai-gpt-5.5` throws instead of routing to the native hyphen id.

Verification on this head:
- `bun test tests/slug-codec.test.ts tests/cli-models.test.ts tests/catalog-input-modality-enum.test.ts ./gui/tests/provider-model-custom-add.test.tsx` — 78 pass / 0 fail
- `bun run typecheck` — clean
- `bun run test` running now on `41abdb93c`

Wibias (changes requested):
Re-review on current head `41abdb93c`. The two previous routing blockers are fixed: `defaultModel` is now part of the shared known-id set, and the route-time ambiguity guard correctly prevents a later live-cache collision from silently switching the upstream model.

One correctness issue remains before approval:

[P2] `decodeRoutedModelIdOrThrow()` traverses `knownIds` twice without materializing it. Its public contract accepts `Iterable<string>`, so a single-use iterable such as a generator is valid input. The first `for...of` consumes it; the subsequent `decodeRoutedModelId(requested, knownIds)` then sees an exhausted iterable and can return the encoded request unchanged instead of the native slash id. Example: a generator yielding only `openai/gpt-5.5` with requested `openai-gpt-5.5` should decode to `openai/gpt-5.5`, but the second traversal has no values left.

Please materialize once (`const ids = [...knownIds]`) and use `ids` for both the ambiguity scan and the final decode. Add the generator regression requested in the current CodeRabbit thread.

Everything else I previously blocked on now looks addressed. After this fix, rerun exact-head CI; the current head's GitHub Actions are still `action_required`, not green.

yansigit:
@Wibias addressed in facbb0c23 (still on current `dev` @ e1769b5e29):

`decodeRoutedModelIdOrThrow()` now materializes `const ids = [...knownIds]` and uses that array for both the ambiguity scan and the final decode. Added the generator regression: a one-shot iterable of `openai/gpt-5.5` decodes requested `openai-gpt-5.5` to `openai/gpt-5.5`.

Verification:
- `bun test tests/slug-codec.test.ts tests/cli-models.test.ts tests/catalog-input-modality-enum.test.ts ./gui/tests/provider-model-custom-add.test.tsx` — 79 pass / 0 fail
- `bun run typecheck` — clean

Wibias (approved):
Approved on exact head `facbb0c2300ae979a810a8bc21b50bf77dc8f6de` after re-review. The previous iterable/generator blocker is fixed correctly and I do not see a remaining code blocker. Cross-platform CI was still running at approval time, so merge should still wait for the exact-head workflow to finish green.

lidge-jun:
## Pre-merge review — one blocker, patch included

Thanks for this. The Cloudflare portion matches the [official models-search contract](https://developers.cloudflare.com/api/resources/ai/subresources/models/methods/list) including `format=openrouter`, and everything else checks out: `bun x tsc --noEmit` exit 0, 129 focused runtime/discovery/Claude-agent/core-Lab tests pass, the GUI suite passes, and `gui build` succeeds.

One blocker before merge.

### The collision guard removes the documented raw full-slash escape hatch

`decodeRoutedModelIdOrThrow` encodes the request before comparing, so a **raw** full-slash selector is judged by its encoded form. Once live discovery adds the hyphenated twin, both spellings throw:

```
routeModel(config, "zenmux/openai-gpt-5.5")  -> ambiguous   (correct)
routeModel(config, "zenmux/openai/gpt-5.5")  -> ambiguous   (wrong)
```

Reproduced on `facbb0c2`: admit custom `openai/gpt-5.5` while the live cache is empty, then let the cache add `openai-gpt-5.5`. The slash model becomes unreachable by **any** spelling — precisely the case raw routing exists to serve.

That contradicts the codec's own stated precedence in its header:

> native exact match (back-compat with raw full-slash selectors) > unique alias

and `README.md:186-187`:

> the raw full-slash form keeps working too

The new test at `tests/slug-codec.test.ts:229-237` only asserts the encoded selector throws, so nothing pins the raw half.

### Patch

Resolve an exact slash-bearing native id before ambiguity is considered:

```diff
 export function decodeRoutedModelIdOrThrow(requested: string, knownIds: Iterable<string>): string {
   const ids = [...knownIds];
+  // An exact slash-bearing native id wins before ambiguity is considered, which
+  // is the codec's stated precedence ("native exact match ... > unique alias")
+  // and the raw full-slash escape hatch README documents. Without this, a live
+  // cache that later adds the hyphenated twin (`openai-gpt-5.5` alongside
+  // `openai/gpt-5.5`) made the encoded selector correctly ambiguous AND the
+  // explicit raw selector throw, leaving the slash model unreachable by any
+  // spelling — the exact case raw routing exists to serve.
+  if (requested.includes("/") && ids.includes(requested)) return requested;
   const encodedRequested = encodeRoutedModelId(requested);
```

The encoded selector still throws, because it genuinely names two ids.

And a regression pinning both halves, appended to the existing collision test:

```ts
test("the raw full-slash selector still routes after a live-cache collision", () => {
  const config = zenmuxConfig();
  config.customModels = [{ id: "c1", provider: "zenmux", modelId: "openai/gpt-5.5" }];
  setCached("zenmux", [{ provider: "zenmux", id: "openai-gpt-5.5" }]);

  expect(() => routeModel(config, "zenmux/openai-gpt-5.5")).toThrow(/ambiguous/);
  expect(routeModel(config, "zenmux/openai/gpt-5.5").modelId).toBe("openai/gpt-5.5");
});
```

Verified locally on top of your head: `tsc` exit 0, `tests/slug-codec.test.ts` 25 pass / 0 fail. Without the source change that new test is the only failure (24 pass / 1 fail), so it does observe the defect.

This PR has `maintainerCanModify: false`, so I could not push the fix to your branch. Could you apply it, or enable maintainer edits and I will push it for you?

Batch note: only `src/types.ts` overlaps another PR in this batch (#1728), at different hunks — no textual conflict.

PR #1801, opened 2026-08-15
fix(cursor): teach code-mode nested-helper contract in tool guidance

jonathanli12:
## Problem

Codex "code mode" advertises a single freeform `exec` tool whose body is JavaScript evaluated in a V8 isolate. Shell, file edits, and MCP are reachable only as nested `await tools.<name>(...)` helpers described inside that tool's own description, so a flat catalog scan cannot see them.

`buildCursorToolGuidanceSystemNote` assumed a flat catalog and injected shell-bridge guidance naming a top-level `exec_command` / `shell_command`. On a code-mode turn those top-level tools do not exist, so the model calls a tool that is not there, gets nothing back, and spends several turns rediscovering the real contract from error messages:

- `exec_command` at top level returns nothing
- empty output until `text()` is called
- `require is not defined` (the isolate is not Node)
- `apply_patch` rejected, because it too is only a nested helper here

The shared `adapters/tool-catalog-nudge.ts` already documents this hazard for `apply_patch`:

> under Codex code mode it is reachable as a nested `tools.apply_patch(...)` helper ... so a flat catalog check cannot see it

The Cursor adapter keeps its own sibling copy of that guidance and never received the equivalent handling.

## Fix

Detect code mode and emit the correct contract instead of the flat-catalog shell-bridge text.

- `isCursorCodeModeExecTool`: freeform `exec` on the Responses provider.
- `cursorRequestUsesCodeMode`: a visible code-mode `exec` **and** no bare shell bridge, evaluated after `toolChoice` filtering so a pinned choice that hides `exec` also disables the branch.
- Two guidance lines describing the nested-helper contract, the V8 isolate, and the `text()` output requirement.

Turns that advertise a bare `exec_command` / `shell_command` are unchanged: the existing shell-bridge, PowerShell, alias, and anti-false-block guidance still applies. The two paths are mutually exclusive by construction.

## Tests

Three cases in `tests/cursor-tool-definitions.test.ts`:

1. detection is true only for freeform `exec` with no bare bridge (covers non-freeform `exec`, bridge present, undefined tools, and a tool-choice pin)
2. code-mode guidance teaches the nested contract and omits the shell-bridge lines
3. a bare bridge still gets the flat-catalog guidance and none of the code-mode text

Verified the coverage is real: stubbing the detection to `false` fails case 2, and restoring it passes.

```
bun run typecheck                     clean
bun test tests/cursor-*.test.ts       236 pass, 0 fail (7 files)
bun run privacy:scan                  passed
```

Rebased onto current `dev` (`e1769b5e2`).

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for detecting Cursor’s Codex code mode.
  - Guidance now explains how to use nested tools within JavaScript execution, including output helpers.
  - Prevents conflicting instructions for shell, patch, and MCP calls in code mode.

- **Bug Fixes**
  - Preserved existing guidance for standard catalogs with direct shell access.

- **Tests**
  - Added coverage for code-mode detection, tool selection, and guidance behavior.

### DIFF
--- a/src/adapters/cursor/tool-definitions.ts
+++ b/src/adapters/cursor/tool-definitions.ts
@@ -191,6 +191,38 @@ export function isCursorWaitTool(tool: Pick<OcxTool, "namespace" | "name">): boo
   return isCursorResponsesProvider(tool.namespace) && tool.name === CODEX_WAIT_TOOL;
 }
 
+/**
+ * True for Codex's unified-exec "code mode" tool: a freeform `exec` whose body is JavaScript
+ * evaluated in a V8 isolate, not a shell command string.
+ */
+export function isCursorCodeModeExecTool(
+  tool: Pick<OcxTool, "namespace" | "name" | "freeform">,
+): boolean {
+  return isCursorResponsesProvider(tool.namespace)
+    && tool.name === CODEX_UNIFIED_EXEC_TOOL
+    && tool.freeform === true;
+}
+
+/**
+ * Codex code mode advertises ONE freeform `exec` tool and no bare shell bridge. Shell, file
+ * edits, and MCP calls are reachable only as nested `tools.<name>(...)` helpers described inside
+ * that tool's own description, so a flat catalog scan cannot see them.
+ *
+ * This matters because the shell-bridge guidance below is written for a flat catalog. Emitting
+ * "call \`exec_command\`" into a code-mode turn names a top-level tool that does not exist: the
+ * model calls it, gets nothing back, and burns turns rediscovering the real contract from error
+ * messages (empty output until \`text()\` is called, \`require is not defined\` because the isolate
+ * is not Node, \`apply_patch\` rejected because it too is only a nested helper here).
+ */
+export function cursorRequestUsesCodeMode(
+  tools: readonly Pick<OcxTool, "namespace" | "name" | "freeform">[] | undefined,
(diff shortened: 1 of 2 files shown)


lidge-jun:
Superseded by #1844, merged to `dev`.

#1844 lands this PR's Cursor code-mode work on top of the #1817 shared contract, and carries all three of its unique tests:

- `detects code mode only when freeform exec has no bare shell bridge` — freeform `exec` / bare-bridge mutual exclusivity
- `teaches the nested-helper contract instead of a top-level shell bridge` — V8 isolate, `await tools.<name>(...)`, `text(...)`, no `require`
- `keeps flat-catalog shell-bridge guidance when a bare bridge is advertised` — the opposite branch stays intact

One correctness defect from this draft was fixed while absorbing it. `cursorRequestUsesCodeMode()` returns true whenever freeform `exec` is visible without a bare shell bridge — which does **not** mean `exec` is the only visible tool. The guidance then said shell/file/MCP "are not separate top-level tools", so a turn advertising both `exec` and, say, `mcp__fs__read_file` told the model that a tool listed in its own catalog was not callable. #1844 scopes the nested-helper claim to the helpers `exec` actually describes and names the remaining top-level tools as still callable, with a new mixed-catalog regression covering it.

Thanks for the original work — the contract and its tests are what landed, just with that one case corrected. Closing as superseded rather than rejected.

PR #1808, opened 2026-08-16
chore(lint): add repo-wide anti-slop profile

Wibias:
## Summary

- vendor and customise a dependency-free subset of `dmmulroy/anti-slop` under its MIT licence, pinned to upstream snapshot `446268e5d15baa968eaec669ff65358d36ae6259`
- apply the same anti-slop profile to production dashboard TypeScript plus `src/` and `scripts/`
- keep migration-heavy rules as warnings, while rejecting `Reflect.get` and `Reflect.apply` as errors
- intentionally omit policies that conflict with OpenCodex boundary parsing, including blanket bans on `unknown`, runtime `typeof`, dictionary-shaped boundary values, and module mocking
- add `bun run lint` for `src/` and `scripts/`, include it in the pre-push gate, and add a CI contract test that runs the profile and checks the Reflect shadowing boundary
- keep Oxlint at the existing `1.76.0`; there are no package, lockfile, or runtime dependency changes

## Verification

- `node --check` passed for the vendored JavaScript plugin before it was committed
- reviewed the final branch diff against `dev`; it is limited to lint configuration, the vendored rules/licence/docs, command wiring, contributor docs, and the lint contract test
- the current execution environment does not provide Bun or Oxlint, so the full repository lint/test commands could not be executed locally here
- the new contract test exercises the real pinned Oxlint installation in Actions: repository source must have no error-level findings, global `Reflect.get` must fail, and a shadowed local `Reflect` API must pass

## Checklist

- [x] Docs were updated for the new repository lint command and pre-push gate.
- [x] Security-sensitive surfaces were reviewed: no dependency installation, auth, secrets, network access, runtime shipping path, or workflow definition is changed; the vendored plugin only inspects Oxlint AST/context data.

## Summary by CodeRabbit

- **Quality Improvements**
  - Added automated checks to identify potentially unsafe type assertions, overly broad parameters, and restricted reflection patterns.
  - Integrated linting into standard development and pre-push validation workflows.
  - Added verification tests to ensure the new checks behave consistently.

- **Documentation**
  - Updated contribution guidance with the expanded linting and validation steps.
  - Added documentation and licensing information for the local code-quality rules.


lidge-jun:
I have own point of view about antislop front so Ill decide it Its I think Its not needed to add rules

PR #1810, opened 2026-08-16
fix(release-readiness): close three coherence defects found on dev

lidge-jun:
## Summary

Coherence review of everything that landed on `dev` this session found three defects that the per-PR audits could not see, because each only appears once the changes are combined or exercised against the real repository. All three are fixed here with activating regressions.

### 1. Log Guard mutations were weaker than the compatibility report they honor (High)

`inspect.ts` classifies a database as compatible only with exact table SQL, exact column metadata, and every canonical index. Both mutation paths rechecked **column names only** inside their write transaction.

The Log Guard lock serializes opencodex against itself; it does not stop Codex or another SQLite writer. So a migration landing between the outer inspection and the locked write is a real window. Reproduced on dev HEAD by dropping `idx_logs_ts` in that gap:

- Protect installed a row-dropping trigger and returned `ok: true`
- Reclaim vacuumed pages

both on a database the inspector reports as `unsupported` / monitor-only.

`hasCurrentLogsSchema` is now exported from `inspect.ts` and both paths delegate to it, so the pre-write gate and the compatibility report cannot disagree.

### 2. Preview releases were unreleasable (High)

`selectReleaseBaseline` picked the newest tag of either channel without regard to reachability. A preview therefore selected the newest **stable** tag — which lives on `main` and is not reachable from `preview` — and the ancestry guard added in #1740 then threw. Verified against this repository's real tags and refs:

```
preview:          without-filter=v2.21.0 (ancestor=false) -> with-filter=v2.19.0
preview-from-dev: without-filter=v2.21.0 (ancestor=false) -> with-filter=v2.17.0
```

The moment the stable lineage moved ahead, no preview could be released at all. Selection now filters to tags actually reachable from the target; the ancestry guard stays as a fail-closed assertion for any future caller that skips the filter.

### 3. Hidden text could forge a duplicate closure (Medium)

`issue-triage-autoclose.cjs` stripped only closed `<!-- ... -->` comments, while the quality parser treats an unclosed comment as hiding everything through EOF. Two issues whose **visible** bodies differ could therefore match on text GitHub never renders, and `issue-triage.yml` closes on that match. Now uses the same through-EOF semantics.

## Verification

Ground truth first, at dev HEAD `8a0de6c44`, full suite on an isolated Linux checkout:

```
12491 tests across 813 files — 13 skip, 7 fail
```

All 7 failures are `Cannot find package 'react'` / `react/jsx-dev-runtime` from `gui/` sources, because the scratch clone installs no GUI dependencies. None touch routing, catalog, or log-guard.

After these fixes, on the same host at this branch head:

```
bun x tsc --noEmit                                        exit 0
bun run test (release-changelog, ci-workflows, log-guard) 169 pass / 0 fail
node --test (issue-triage-autoclose, issue-quality,
             pr-quality)                                  209 pass / 0 fail
```

Each new regression was reproved to fail without its source change.

Three independent `gpt-5.6-sol` explorers audited dev in parallel across catalog/discovery, the log-guard stack, and CI-script governance. Their "checked and sound" findings are worth recording too: inspection cache invalidation works correctly (every mutation changes the DB/WAL/SHM stamp), the lock correctly serializes Protect against Reclaim, and the changelog builder renders the real 12-commit and 23-commit ranges with zero coverage errors.

Two lower-severity findings from those lanes are **not** fixed here and are recorded rather than silently dropped: a late slug collision can misattribute capabilities to an unroutable custom model, and prefix stripping runs before sibling enrichment so a Cloudflare-style `@cf/...` id loses its sibling capabilities. Both are pre-existing interactions rather than regressions from this session, and each deserves its own audited change.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duplicate-signature detection by ignoring hidden text after unterminated HTML comments.
  * Release changelog generation now avoids selecting unrelated release tags.
  * Strengthened log protection and maintenance checks to detect missing canonical indexes before changes are applied.

* **Tests**
  * Added coverage for release ancestry filtering and schema-protection regressions.

### DIFF
--- a/.github/scripts/issue-triage-autoclose.cjs
+++ b/.github/scripts/issue-triage-autoclose.cjs
@@ -66,7 +66,11 @@ function normalizeSignatureLine(raw) {
   // case-bearing discriminators; the "exact signature" promise is only true
   // if the comparison keeps them distinct.
   return String(raw || "")
-    .replace(/<!--[\s\S]*?-->/g, " ")
+    // An UNCLOSED comment hides everything after it through EOF, exactly as the
+    // quality parser treats it. Matching only `<!-- ... -->` let hidden text
+    // become the "exact shared signature" that closes an issue: two reports
+    // whose visible bodies differ could match on text GitHub never renders.
+    .replace(/<!--[\s\S]*?(?:-->|$)/g, " ")
     .replace(/^\s*(?:>|[-*+]\s+|\d+[.)]\s+)/, "")
     .replace(/^[`~]{3,}[^\n]*$/, "")
     // Backticks are unambiguous Markdown code delimiters. Preserve `_`, `~`,
--- a/scripts/build-release-changelog.ts
+++ b/scripts/build-release-changelog.ts
@@ -118,14 +118,25 @@ export function categoryForPull(pr: AssociatedPullRequest): string {
  * Prerelease: newest prior release of either channel, so previews stay incremental.
  * Stable: newest prior stable only, so the final changelog always reconstructs
  * the complete stable-to-stable range and cannot lose prerelease changes.
+ *
+ * `isAncestor` filters to tags actually reachable from the target. Without it a
+ * preview could select the newest STABLE tag, which lives on `main` and is not
+ * reachable from `preview` — the ancestry guard then threw and no preview could
+ * be released at all once the stable lineage advanced. Selecting the newest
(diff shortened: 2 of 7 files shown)


lidge-jun:
Landed directly on dev as 3c633d790 and 65eda6c28 per maintainer instruction; closing this PR rather than merging it twice.

Issue #1699, opened 2026-08-14
[Bug]: OpenCode Go native Responses path leaks undeclared apply_patch and shows aborted


rodrigueez:
## Client or integration

Codex App

## Area

Provider adapter

## Summary

OpenCodex 2.17.0 can leak an undeclared top-level `apply_patch` function call through the native OpenAI Responses passthrough path when using OpenCode Go. Codex App then displays `aborted`, while the target file is unchanged.

Expected behavior: an undeclared routed tool should fail closed with an explicit compatibility error before reaching the Codex executor. The supported editing path should remain `exec -> tools.apply_patch(...)`.

This is related to #1544, #1576, and #1623, but the relevant distinction is that this reproduction uses the `openai-responses` adapter path.

## Reproduction

1. Start OpenCodex on port 10100.
2. Use `exec -> tools.apply_patch(...)` to create a temporary file containing `v1`.
3. Invoke top-level `apply_patch` with a patch changing `v1` to `v2`.
4. Observe that Codex App displays `aborted`.
5. Check the file: it remains `v1`; no content or mtime change is observed.
6. Run the equivalent patch through `exec -> tools.apply_patch(...)`:
   - an intentionally incorrect expected line returns a real `apply_patch verification failed` error;
   - the correct `v1 -> v3` patch succeeds.

The request-visible tool catalog contained `exec`, `wait`, and `request_user_input`, but no top-level `apply_patch` schema. The session nevertheless received a `function_call` named `apply_patch`.

## Version

OpenCodex 2.17.0

## Operating system

macOS 27.0, Apple Silicon

## Provider and model

Provider: `opencode-go`

Model: `deepseek-v4-flash`

At the time of reproduction, the provider default was `openai-chat`, but this model had an explicit `openai-responses` adapter override.

## Logs or error output

Observed Codex result:

`aborted`

Session observations:

- top-level call was recorded as `function_call(name=apply_patch)`;
- no corresponding `custom_tool_call_output` was emitted;
- the file was not modified;
- OpenCode Go usage records were HTTP 200;
- no upstream `aborted`, OpenCodex `response.failed`, approval denial, or sandbox denial was observed.

## Screenshots and supporting files

No screenshots attached. Reproduction artifacts were written only under `/tmp`.

## Redacted configuration

```json
{
  "providers": {
    "opencode-go": {
      "adapter": "openai-chat",
      "baseUrl": "https://opencode.ai/zen/go/v1",
      "modelAdapters": {
[...]

PR #1817, opened 2026-08-16
fix(tools): distinguish top-level calls from nested helpers

### LINKED ISSUE #1795: [Bug]: sensenova hallucinates undeclared exec tool in subagent requests causing 502 stream disconnected
## Client or integration

Codex App

## Area

Provider adapter / bridge

## Summary

When OpenCodex routes subagent requests (explorer, shadow calls via gpt-5.6-luna) to sensenova/sensenova-6.8-flash-lite, the model hallucinates tool calls to `exec` even when `exec` is not declared in the request tools. This triggers bridge.ts:998 validation failure, causing 502 and stream disconnect.

Error: `routed provider emitted undeclared client tool "exec"; only request-declared tools may be called`

## Reproduction

1. Configure shadowCallIntercept to route gpt-5.6-luna to sensenova/sensenova-6.8-flash-lite
2. Codex spawns subagent (explorer) with limited tool set (search_files, read_file, etc., NO exec)
3. sensenova sees system prompt mentioning exec capability and hallucinates tool call to exec
4. bridge.ts:998 checks declaredToolNames.has("exec") -> false -> 502

## Expected behavior

Undeclared tool calls from third-party providers should be tolerated (dropped with warning) instead of failing the entire turn. This is especially important for subagent/shadow calls where the model may see system prompts describing capabilities not available in the current request tool set.

## Actual behavior

502 error: stream disconnected before completion

## Version

OpenCodex 2.21.0

## Operating system

macOS 27.0, Apple Silicon

## Provider and model

Provider: sensenova
Model: sensenova-6.8-flash-lite
Adapter: openai-chat

## Logs or error output

From usage.jsonl - 31 occurrences between 01:40-02:42 on 2026-08-16, all with gpt-5.6-luna -> sensenova routing:

provider: sensenova
model: sensenova-6.8-flash-lite
status: 502
upstreamError: routed provider emitted undeclared client tool "exec"; only request-declared tools may be called
requestedModel: gpt-5.6-luna
resolvedModel: sensenova-6.8-flash-lite

## Proposed fix

In bridge.ts around line 998, when declaredToolNames check fails, drop the tool call and continue streaming instead of returning 502:

if (options?.declaredToolNames && !options.declaredToolNames.has(event.name)) {
  console.warn("[opencodex] dropping undeclared tool call from routed provider");
  break; // skip this tool call, continue stream
}

This allows the turn to complete with text response instead of failing entirely.

Wibias:
## Summary

- fix `src/responses/parser.ts` so `apply_patch` patch-envelope guidance is attached only to the logical `apply_patch` custom tool; other custom tools such as Code Mode `exec` now receive neutral freeform-input guidance
- tighten the shared non-OpenAI tool-catalog nudge so the request catalog is explicitly defined as the complete **top-level** invocation surface
- clarify that names mentioned only in instructions, tool/argument descriptions, or nested helper APIs do not become additional top-level tools, and nested `tools.*` helpers must be invoked through their listed parent tool
- keep the existing fail-closed `declaredToolNames` bridge enforcement unchanged; this PR does not drop undeclared calls, auto-translate them, or add undeclared names to the allowed set
- add focused parser and nudge regressions
- add the full investigation trace under `devlog/_plan/260816_undeclared_tool_invocation_contract/000_research.md`, covering #1544, #1699/#1700, the recorded bridge decision, the prior `apply_patch`-ban regression fixed by `0325a5a`, #1795, and the Teamwicked Kimi K3 reproduction

### Root cause

Two guidance defects combine at the routed Code Mode boundary.

First, `buildTools()` lowered **every** Responses `custom` tool with an `input` argument description that specifically taught `apply_patch` syntax. A Code Mode `exec` declaration could therefore tell a routed model all of the following at once: top-level name `exec`, nested helper `tools.apply_patch(...)`, and `exec.input` text explaining the `apply_patch` patch envelope.

Second, the generic non-OpenAI nudge listed exact names but did not explicitly distinguish top-level callable names from names that appear only in instructions or inside a declared tool's metadata/input language. #1795 shows SenseNova promoting prompt-mentioned `exec` into an undeclared top-level call; a Kimi K3 reproduction on v2.14.2 showed the same class of mistake with nested `tools.apply_patch` becoming top-level `apply_patch`.

That Kimi reproduction was initially misidentified as v2.22.0 because it came from a different worktree that was still running v2.14.2. Re-testing the same Kimi K3 / `openai-chat` / `xhigh` Code Mode scenario on actual v2.22.0 has not reproduced the undeclared top-level `apply_patch` failure so far. This PR therefore does **not** claim a live v2.22.0 Kimi repro.

The separate model-facing metadata defect fixed here is still present in v2.22.0/current dev: `src/responses/parser.ts::buildTools()` still applies the `apply_patch`-specific input description to every Responses `custom` tool, and the shared non-OpenAI catalog nudge still lacks the explicit top-level-vs-nested-helper distinction added by this PR.

The final bridge rejection is intentional and remains the enforcement boundary. The fix is to stop sending contradictory metadata and make invocation levels explicit before generation.

### Research / issue trace

Primary issue: #1795

Additional history:

- #1544: DeepSeek CodeModeOnly generated top-level `apply_patch` while the supported path was `exec -> tools.apply_patch(...)`
- #1699 / #1700: native Responses could leak undeclared top-level `apply_patch`
- `structure/04_transports-and-sidecars.md`: recorded decision to retain the request-visible allowed wire-name set and fail closed before emitting an undeclared tool item
- `0325a5afd16204e2300b432aeb273412d32cbf68`: removed the old blanket `apply_patch` prohibition after it pushed routed models toward Python heredocs / `sed`
- #1795 comment `5305003614`: correctly rejected "drop and continue" as a reversal of the recorded fail-closed contract
- #1795 comment `5305779307`: Teamwicked Kimi K3 reproduction, `openai-chat`, explicit-provider route, `xhigh`, failing after ~31.5 s with `routed provider emitted undeclared client tool "apply_patch"`; later corrected to v2.14.2 rather than v2.22.0

The complete evidence chain, rejected alternatives, and acceptance criteria are committed in:

`devlog/_plan/260816_undeclared_tool_invocation_contract/000_research.md`

This PR intentionally does **not** use `Closes #1795` yet. The implementation targets the confirmed metadata mismatch while keeping the fail-closed bridge behavior unchanged; live-provider behavior should remain separately evidenced rather than inferred from the static defect.

## Verification

- branch was created directly from upstream `dev`; it has since been compared against the updated target branch
- static compare remains 5 files only; the unrelated blank-line deletion in `src/responses/parser.ts` flagged during review has been restored, so the parser diff is limited to the intended `buildTools()` change
- regression added: `tests/responses-custom-tool-guidance.test.ts`
  - `apply_patch` keeps `*** Begin Patch` guidance
  - `exec` and an unrelated custom tool get `Raw freeform input for this tool.`
  - non-patch custom-tool argument help no longer mentions `apply_patch`
- regressions extended: `tests/tool-catalog-nudge.test.ts`
  - valid names are explicitly the valid **top-level** names
  - names present only in instructions/tool metadata/nested helper APIs are not additional top-level tools
  - nested helpers are used through the listed parent tool
  - `apply_patch` remains absent from the neighboring-agent prohibition
- existing bridge enforcement was not modified
- independent Windows 11 review run by @ntdatt812: `bun run typecheck` clean; the two focused suites 10 pass / 0 fail; 12 related compatibility suites 389 pass / 0 fail; a 163-file regression comparison introduced no new failures versus `dev`

## Checklist

- [x] Docs or release notes were updated when needed. The public devlog contains the complete investigation and design trace; no end-user configuration changed.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. This PR changes model-facing tool metadata/guidance only and preserves the stricter fail-closed execution boundary.

## Summary by CodeRabbit

* **Improvements**
  * Clarified which names represent callable top-level tools versus nested helper APIs.
  * Added guidance to access nested helpers through their parent tools.
  * Tailored custom-tool input guidance so patch instructions apply only to patch tools.

* **Tests**
  * Added coverage for tool naming boundaries and tool-specific input guidance.

* **Documentation**
  * Documented historical incidents, design decisions, regression coverage, and acceptance criteria.

### DIFF
--- a/devlog/_plan/260816_undeclared_tool_invocation_contract/000_research.md
+++ b/devlog/_plan/260816_undeclared_tool_invocation_contract/000_research.md
@@ -81,9 +81,9 @@ A follow-up comment correctly pointed out that this contradicts the repository's
 
 Reference: https://github.com/lidge-jun/opencodex/issues/1795#issuecomment-5305003614
 
-## Kimi K3 v2.22.0 reproduction
+## Kimi K3 v2.14.2 reproduction and v2.22.0 re-check
 
-A second reproduction on current stable OpenCodex v2.22.0 confirmed that the problem is not SenseNova-specific and also reaches the Codex-native Code Mode helper boundary.
+A second reproduction was captured with Teamwicked Kimi K3 on OpenCodex v2.14.2. It showed the same semantic mismatch at the Codex-native Code Mode helper boundary. That run was initially misidentified as v2.22.0 because it came from a different worktree that was still running v2.14.2.
 
 Observed request characteristics:
 
@@ -104,9 +104,16 @@ routed provider emitted undeclared client tool "apply_patch"; only request-decla
 
 The request/conversation identifiers were intentionally omitted from the public issue comment.
 
-Reference: https://github.com/lidge-jun/opencodex/issues/1795#issuecomment-5305779307
+References:
 
-This reproduction matters because Code Mode deliberately exposes patching through the declared `exec` surface with a nested helper:
+- original repro: https://github.com/lidge-jun/opencodex/issues/1795#issuecomment-5305779307
+- version correction: https://github.com/lidge-jun/opencodex/issues/1795#issuecomment-5306871615
+
+Re-running the same Kimi K3 / `openai-chat` / `xhigh` Code Mode scenario on actual v2.22.0 has not reproduced the undeclared top-level `apply_patch` failure so far. The live v2.22.0 Kimi case therefore remains unconfirmed.
+
+That does not remove the separate static metadata defect addressed by this PR: v2.22.0/current dev still applies `apply_patch`-specific `input` guidance to non-`apply_patch` Responses custom tools, and the non-OpenAI catalog nudge still does not explicitly distinguish the top-level invocation surface from nested helper APIs.
+
+The v2.14.2 reproduction still matters because Code Mode deliberately exposes patching through the declared `exec` surface with a nested helper:
 
 ```text
 exec
--- a/devlog/_plan/260816_undeclared_tool_invocation_contract/000_research.md
+++ b/devlog/_plan/260816_undeclared_tool_invocation_contract/000_research.md
@@ -1,0 +1,227 @@
+# Undeclared tool invocation contract: research trace
+
+Date: 2026-08-16
+Status: implementation in review
+Issue: https://github.com/lidge-jun/opencodex/issues/1795
+Branch base: `lidge-jun/opencodex@65eda6c28a480a15b245af50fd533687f3317956`
+
+## Problem statement
+
+Routed non-OpenAI models can emit a top-level tool call whose name was never declared in the current request-visible tool catalog. OpenCodex intentionally rejects that at the final bridge boundary with an explicit compatibility error instead of allowing an invented client-executable action through.
+
+The remaining bug is upstream of that guard: some routed turns contain enough conflicting instruction/tool metadata for models to infer that a name mentioned inside instructions or a nested helper surface is also a valid top-level tool call.
+
+Two concrete forms are now reproduced:
+
+1. SenseNova emits top-level `exec` in a subagent turn whose request catalog does not declare `exec`.
+2. Kimi K3 emits top-level `apply_patch` in Codex Code Mode where patching is available only as the nested `exec -> tools.apply_patch(...)` helper.
+
+The correct fix must reduce those invalid generations without weakening the final fail-closed enforcement boundary.
+
+## Historical trace
+
+### #1544: DeepSeek CodeModeOnly top-level apply_patch
+
+Issue #1544 reported that DeepSeek V4 Flash could emit a top-level `apply_patch` call even though the request-visible top-level tools did not contain `apply_patch`. The supported control path in the same conversation was:
+
+```text
+exec -> tools.apply_patch(...) -> {}
+```
+
+That established the core distinction: patch capability can exist without top-level `apply_patch` being callable.
+
+Reference: https://github.com/lidge-jun/opencodex/issues/1544
+
+### #1699 / #1700: native Responses leaked undeclared apply_patch
+
+The OpenCode Go native Responses path later reproduced the same semantic mismatch. A top-level `apply_patch` item could reach Codex even though the request-visible catalog contained `exec`, `wait`, and `request_user_input`, not `apply_patch`. Codex displayed `aborted`; the nested Code Mode path still worked.
+
+Those reports strengthened the requirement that the proxy must validate returned tool names against what the caller actually declared rather than trusting model output or prompt guidance.
+
+References:
+
+- https://github.com/lidge-jun/opencodex/issues/1699
+- https://github.com/lidge-jun/opencodex/issues/1700
+
+### Fail-closed bridge decision
+
+`structure/04_transports-and-sidecars.md` records the resulting architecture decision:
+
+- retain the request-visible allowed wire-name set;
+- validate at the final bridge;
+- fail the turn before emitting any undeclared tool item;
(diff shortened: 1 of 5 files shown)


ntdatt812:
You mentioned this environment cannot run Bun, so I ran the suites here on Windows 11 against `dev`. Results below; nothing in them is a blocker.

**Typecheck:** `bun run typecheck` clean.

**Your new coverage:** `tests/responses-custom-tool-guidance.test.ts` + `tests/tool-catalog-nudge.test.ts` — 10 pass / 0 fail.

**Everything that asserts the nudge wording or the patch-envelope string.** I grepped the suite for `Tool contract`, `Valid tool names for this turn`, `neighboring-agent tool names`, `Raw tool input` and `Begin Patch`, which lands on 12 files — `adapter-usage`, `bridge`, `cursor-tool-definitions`, `cursor-structured-edit`, `cursor-blob`, `cursor-protobuf-events`, `github-copilot-stream-contract`, `github-copilot-sse-rewrite`, `google-adapter`, `kiro-adapter`, `responses-stream-tool-events`, `umans-provider`. 389 pass / 0 fail. Adding two sentences to the nudge did not disturb any adapter that embeds it.

**Regression sweep.** 163 test files (everything matching responses/tool/adapter/bridge/parser/openai/anthropic/google/kiro/cursor/copilot/claude), run through `bun scripts/test.ts` so they get the sandboxed home, once on `dev` and once on this branch:

| | pass | fail |
|---|---|---|
| `dev` | 2694 | 6 |
| this PR | 2698 | 4 |

The failing sets diff to **no new entries**. The 4 that remain on both are `EPERM: symlink` in `responses-state`, which needs administrator or developer mode on Windows and fails on a clean checkout too. The 2 extra on the `dev` run were `Kiro OAuth upstream 401 replay` and did not recur on the branch run — flaky here, unrelated to this change.

**Behaviour spot-check**, since I wanted to confirm the parser change rather than infer it from its own test. Parsing a request declaring `apply_patch`, `exec` and one other custom tool gives:

```
apply_patch => Raw tool input. For apply_patch, begin exactly with `*** Begin Patch` (no trailing `***`), then use its standard patch envelope.
exec        => Raw freeform input for this tool.
some_other  => Raw freeform input for this tool.
```

So the guidance is scoped to the tool that owns it, which is the part of the diagnosis I could not connect when I looked at this on #1795 — I had read that hardcoded description in `buildTools` and did not link it to the hallucination. Nice catch.

**What I could not verify:** the live SenseNova/Kimi re-probe you flagged. I have no credentials for either, so the "does the model actually stop hallucinating `exec`" question is still open on evidence — everything above only shows the change is contract-safe.

One cosmetic note: the diff drops a blank line at `src/responses/parser.ts:501`, unrelated to the change. Harmless, just flagging in case you want the diff minimal.

Wibias:
Correction on the live Kimi evidence: I discovered that the failing repro I reported earlier was run from a different worktree that was still running OpenCodex v2.14.2, not v2.22.0. I re-tried the same Kimi K3 / openai-chat / xhigh Code Mode scenario on actual v2.22.0 and could not reproduce the undeclared top-level `apply_patch` failure so far.

That means we should not present the Kimi failure as a confirmed v2.22.0 live repro. The separate parser/tool-metadata defect addressed by this PR is still present in v2.22.0, though: `src/responses/parser.ts::buildTools()` still gives every Responses `custom` tool the `apply_patch`-specific `input` description, including non-patch tools such as `exec`. The non-OpenAI catalog nudge in v2.22.0 also still lacks the explicit top-level-vs-nested-helper distinction added here.

So the updated evidence is: live Kimi failure reproduced on v2.14.2; not reproduced yet on v2.22.0; the underlying model-facing metadata bug fixed by this PR is independently present in v2.22.0/current dev. I’ll keep the live-provider claim separate from that static defect.

PR #1836, opened 2026-08-16
fix(antigravity): align UA to 2.5.5 decompile literal and drop x-goog-api-client

dbc-hbin:
## Summary

Align Antigravity HTTP headers with decompiled `2.5.5 arm64 1.107.0` `language_server_macos_arm`.

## Evidence (2.5.5 arm64, 126MB Go1.26.5)

- **Binary**: `/Applications/Antigravity IDE.app/.../bin/language_server_macos_arm`, `__lrodata_gopcln` 37MB, `gosym.NewTable` 126300 funcs.
- **User-Agent**:
  - `GetUserAgentName` — IDE `0x1018e9a70` (48B), CLI `0x1018ec950`, Hub `0x1018ef450` — identical:
    ```
    adrp x27, #0x107b91000; add x27, #0x880  -> bss override (SetUserAgentNameOverride @ override_user_agent_name 0x254cd06)
    ldp  x2, x3, [x27]
    cmp  x3, #0
    mov  x4, #0xb
    csel x1, x3, x4, ne
    adrp x3, #0x102472000
    add  x3, #0xc7b
    csel x0, x2, x3, ne
    ret
    ```
  - Fallback `va 0x102472c7b` (`fileoff 0x2472c7b`, `len 0xb` = 11) → `"antigravity"` (`616e746967726176697479`), **not** `"antigravity-ide"`.
  - Raw `antigravity-ide @0x24c59ab` (`va 0x1024c59ab`, count 2) — doc string `**IDE**: \`antigravity-ide/\`` only, `ADRP page 0x1024c5000 + 0x9ab` exact **0 hits**.
  - `antigravity/ide` **0**, `aidev_client` **1** (log `cloudcode-paaidev_client`), `windows/amd64` **0**.
  - Prior `2.0.3 x64` private claim `LEA RDX, [RIP-0x284fc90] → 0x367b554 "antigravity-ide" @0x5ecb1dd` / `-override_user_agent @0x5ecbc37` is **stale**. Flag renamed to `override_user_agent_name @0x254cd06`.
- **Headers**:
  - `SetHTTPHeaders` — IDE `0x1018e9ca0` (16B `ret`), Standalone `0x1018ea350` (16B), Stubby `0x1018f01d0` (16B), CLI `0x1018ecfc0` (704B), Hub `0x1018ef6d0` (832B).
  - CLI/HUB each `1 ADRP` only for `X-Goog-User-Project @0x1018ed1b8`; **no** `User-Agent` / `x-goog-api-client` ADRP.
  - Raw `x-goog-api-client @0x24ea019` — false positive `generationConfig.x-goog-api-client`; `google-api-nodejs-client` **0**, `gl-node` **0**, `Client-Metadata` **0**.
  - Capstone `User-Agent: %s` page `0x1024d7000 + 0xea0` — 69 ADRP hits, all other strings.
  - `loadCodeAssist` (`0x27a4f5d`) / `onboardUser` (`0x27a53f5`) literals exist but `0 ADRP` — not header-synthesized.

## Changes

| File | Before | After |
|------|--------|-------|
| `src/adapters/client-fingerprint.ts` | `antigravity/ide/2.5.5 (aidev_client; os_type=windows; arch=amd64)` | `"antigravity"` literal (11). `ANTIGRAVITY_GOOG_API_CLIENT_UA` kept deprecated for compat. Env override `GOOGLE_ANTIGRAVITY_USER_AGENT` / `PI_AI_ANTIGRAVITY_USER_AGENT` (both `trim()`, `_version` arg deprecated). |
| `src/oauth/google-antigravity.ts` | `onboardUser` sent `x-goog-api-client: google-api-nodejs-client/10.3.0` | Dropped (decompile: 5 funcs, 0 ADRP). Header `User-Agent: antigravity`. Body `metadata.ide_version` now `ANTIGRAVITY_IDE_VERSION` (`2.5.5`) — not UA literal (CodeRabbit fix). |

## Verification

```
bun x tsc --noEmit
bun test tests/client-fingerprint.test.ts tests/google-antigravity-wire.test.ts tests/google-antigravity-oauth.test.ts  # 72 pass
bun test tests/provider-outbound.test.ts tests/provider-quota.test.ts  # 106 pass
```

Header-only, no behavioral regression.

## Review readiness checklist

### DIFF
--- a/src/adapters/client-fingerprint.ts
+++ b/src/adapters/client-fingerprint.ts
@@ -41,23 +41,33 @@ export function claudeCodeSessionId(token: string | undefined): string {
 }
 
 // ── Antigravity IDE ──
-/** Pinned fallback Antigravity IDE language-server version (matches the bundled LS 2.5.5). */
+// Decompiled Antigravity 2.5.5 arm64 (1.107.0, language_server_macos_arm 126MB Go1.26.5, __lrodata_gopcln 37MB):
+// - gosym NewTable 126300 funcs:
+//   IDE GetUserAgentName 0x1018e9a70 sz48, CLI 0x1018ec950 sz48, Hub 0x1018ef450 sz48 (identical bytes)
+//   0x1018e9a70: adrp x27,#0x107b91000; add #0x880 -> bss override (SetUserAgentNameOverride @ override_user_agent_name 0x254cd06)
+//            ldp x2,x3,[x27]; cmp x3,#0; mov x4,#0xb; csel x1,x3,x4,ne; adrp x3,#0x102472000; add #0xc7b; csel x0,x2,x3,ne; ret
+//   fallback va 0x102472c7b fileoff 0x2472c7b len 0xb (11) => "antigravity" (616e746967726176697479)
+//   raw "antigravity-ide" @0x24c59ab va 0x1024c59ab count2 doc "**IDE**: `antigravity-ide/`" ADRP page 0x1024c5000+0x9ab exact 0 hits
+//   "antigravity/ide" count0, "aidev_client" 1 (log cloudcode-paaidev_client), windows/amd64 0
+// - x-goog-api-client: raw count1 @0x24ea019 "generationConfig.x-goog-api-clientsystemInstruction" false positive,
+//   ADRP page 0x1024ea000 0 hits, google-api-nodejs-client 0, gl-node 0, Client-Metadata 0
+// - SetHTTPHeaders: IDE 0x1018e9ca0 16 ret, Standalone 0x1018ea350 16 ret, Stubby 0x1018f01d0 16 ret,
+//   CLI 0x1018ecfc0 704 1 ADRP X-Goog-User-Project @0x1018ed1b8, Hub 0x1018ef6d0 832 cloudcode-paaidev_client + X-Goog-User-Project,
+//   no UA/x-goog-api-client ADRP (capstone 69 hits for User-Agent page 0x1024d7000 are other strings)
+// 2.0.3 x64 private ("antigravity-ide" LEA RDX,[RIP-0x284fc90]->0x367b554, -override_user_agent @0x5ecbc37) is stale.
+/** Pinned fallback Antigravity IDE language-server version (metadata only, not UA). */
 export const ANTIGRAVITY_IDE_VERSION = "2.5.5";
-const ANTIGRAVITY_IDE_CLIENT_NAME = "aidev_client";
-const ANTIGRAVITY_IDE_PLATFORM = "windows/amd64";
-/** Secondary Google API client UA the Antigravity client library reports. */
+/** Deprecated: not sent on wire (decompiled 0 hits). Kept for compat. */
 export const ANTIGRAVITY_GOOG_API_CLIENT_UA = "google-api-nodejs-client/10.3.0";
 
 /**
- * The real Antigravity IDE User-Agent, e.g.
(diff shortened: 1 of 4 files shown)


lidge-jun:
Thanks for this — the decompilation work is unusually thorough, and I want to be specific about why it is not merging as-is rather than leaving it sitting in draft.

**The gate that is blocking it is working correctly.** `enforce-target` failed with `unsponsored_surface` because the PR touches `src/oauth/`, which requires a maintainer to sponsor the change (`.github/scripts/pr-sponsored-surface.cjs`). That is not something to route around; it exists so a credential-path change gets a named owner.

**The substantive problem is a direct conflict with a change that shipped for a reason.** Commit `875bb70c3` switched the UA *to* the IDE shape specifically to un-gate `gemini-3.7-flash`:

> The Cloud Code Assist backend gates newer agent models (gemini-3.7-flash) by client family, returning 404 NOT_FOUND to CLI-shaped UAs even with a valid OAuth token. Only `antigravity/ide/<version>` unlocks them.

That claim is about **backend behavior observed on the wire**. Your evidence is about **what the binary would send by default**. Both can be true at once: the binary's fallback can be the bare literal while the backend still gates on a fuller UA that the real client constructs elsewhere, or sends under a different code path than `SetHTTPHeaders` — and your own analysis notes the IDE/Standalone/Stubby variants are 16-byte `ret` stubs, which means the header is being set somewhere those five functions do not cover.

So the decompile does not by itself refute the 404 observation, and merging this would silently re-lock `gemini-3.7-flash` for every Antigravity user if the earlier report was right. The wire test that pins the current UA exists precisely to prevent that from happening by accident.

**What would settle it.** One live request each, against the Cloud Code Assist backend with a valid token:

- `User-Agent: antigravity` requesting `gemini-3.7-flash` → does it 200 or 404?
- the current `antigravity/ide/2.5.5 (...)` UA, same request → confirming the baseline still works.

If the bare literal returns 200, your change is right and the earlier commit was cargo-culting a UA that happened to work. If it returns 404, the backend gates on something the binary's default does not produce, and the current shape should stay.

The `x-goog-api-client` removal is a separate and much easier call — your evidence there (0 ADRP hits, the single false positive at `0x24ea019` being a `generationConfig` string) is convincing on its own, and that header is only sent during onboarding. If you want to split that into its own PR it can move independently of the UA question.

Two smaller notes for whenever this does land:

- `antigravityUserAgent(_version?: string)` silently ignoring its argument is a trap for the next caller. If the version genuinely has no effect, drop the parameter so the type system says so.
- Keeping `ANTIGRAVITY_GOOG_API_CLIENT_UA` exported "for compat" with a test asserting its exact value pins a constant nothing sends. Either delete it or mark it `@deprecated` with a removal target.

Leaving this open pending the live check. Not a rejection — the analysis may well be correct, but the cost of being wrong here is a silent model outage, so it needs wire evidence rather than binary evidence.

dbc-hbin:
Closing this PR to separate the changes as discussed.

Our deeper decompile of  2.5.5 confirms that  (`0x1018fbe00`) formats `%s/%s/%s (%s)` using the subclient type (`ide`) injected by Electron Main (`getLsArgs.js`), yielding `antigravity/ide/2.5.5 (...)`. Thus, the wire User-Agent observation was completely accurate and the IDE shape should stay.

Opening a clean, focused PR that only drops `x-goog-api-client` and separates `ide_version` metadata.

PR #1854, opened 2026-08-16
fix(gui): bound every client-resource fetch with a settling deadline

lidge-jun:
## Summary

The dashboard could sit on a loading skeleton forever, and refreshing only reset the clock — the next stall wedged it again. This is layer 1 of a four-part fix: **no fetch on the resource data path may pend without end.**

A hung management request used to wedge its store permanently. Poll ticks skip while a request is in flight, the visibility make-up fetch skips too, and nothing else aborts — so a single connection that accepts but never answers left `loading`/`refreshing` true with no in-flight progress and no future tick that could ever settle it. Measured on a sandboxed instance: a stalled `/api/settings` produced **zero** retries over 40s while its 5s poll kept firing.

`runFetch` now races the fetcher against a per-attempt deadline (30s default; 60s for the documented-slow 30d usage and models-catalog resources). A `timedOut` flag plus a `RESOURCE_TIMEOUT` abort reason keeps timeouts distinguishable from owner aborts, so a timeout settles as a failure while replace/unmount aborts still never settle. Polled stores self-heal on the next tick; cold non-polled stores land on the existing `failed-cold` error surface with its retry control instead of an endless skeleton.

The race also bounds fetchers that drop the signal entirely (`Subagents` was one — its loader now forwards the signal).

Design and evidence: `devlog/_plan/260816_gui_loading_performance/010_phase1_resource_deadline.md`

**Stack** (merge bottom-up):

| 4 | (poll consolidation) | shared scheduler + revisit freshness | bucket state machine |
| 3 | (hidden pause) | hidden tab = zero timers | suspension rules |
| 2 | (auth unwedge) | 401 re-bootstrap deadline | shared resolution |
| 1 | **this PR** | resource deadline | settle semantics |

## Verification

- `cd gui && bun test tests` → 922 pass / 0 fail
- `cd gui && bun run lint`, `bun run lint:i18n`, `bun run build` → green
- `bun run typecheck` (root) → green
- New `gui/tests/client-resource-deadline.test.tsx` (7 cases): never-settling fetcher settles failed, polled store self-heals, timed-out cold store shows the error surface (not a skeleton), unmount still takes the abort path, manual refresh recovers, signal-dropping fetcher is bounded, and an owner abort after the deadline cannot double-settle.
- Live browser check against a sandboxed proxy with CDP fault injection: the stalled endpoint now settles and retries on a 35s cycle (30s deadline + 5s tick) where it previously never retried at all.

No visual change — this is request-lifecycle plumbing. Dashboard after the change (unchanged rendering, data loads normally): screenshot below.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented GUI screens from remaining indefinitely stuck while loading.
  - Added request time limits so stalled requests fail visibly and can recover through retries or refresh.
  - Improved authentication recovery after temporary session or network failures.
  - Improved cancellation during navigation and unmounting.

- **Performance**
  - Reduced background polling activity when browser tabs are hidden.
  - Preserved timely updates while limiting unnecessary network requests.
  - Added longer timeouts for known slow usage and catalog data requests.

### DIFF
--- a/devlog/_plan/260816_gui_loading_performance/000_plan.md
+++ b/devlog/_plan/260816_gui_loading_performance/000_plan.md
@@ -1,0 +1,92 @@
+# GUI loading resilience + performance campaign — plan
+
+Date: 2026-08-16. Session: 01a00a59-96db-72a0-a12d-c8b9639e5607 (cxc-loop HOTL,
+goalplan slug: opencodex-gui-dashboard-performance-campaign-fix).
+
+## Objective
+
+1. Kill the "infinite loading" wedge: any tab can sit on a skeleton forever until F5,
+   then re-wedge later. Root-caused below with live fault-injection evidence.
+2. Per-tab performance: fewer timers, fewer redundant requests, no unbounded fetches.
+3. Hidden tab ≈ zero cost: every poll/timer in the dashboard pauses while
+   `document.visibilityState === "hidden"` and resumes (with one make-up refresh) on
+   visible. Only explicit opt-outs keep running (restart-reconnect detection).
+
+## Constraints
+
+- Worktree: /Users/jun/.codex/worktrees/f9b0/opencodex (detached at origin/dev tip
+  b81314cd2 when started). Branches: codex/gui-* stack, PRs target dev / parent branch.
+- The user's running ocx (port 10100) is never touched. All live verification runs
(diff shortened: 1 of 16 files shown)


lidge-jun:
Dashboard after the change on a sandboxed instance — rendering is unchanged and data loads normally; this layer only alters request lifecycle, not UI.

PR #1873, opened 2026-08-16
fix(chat): retain admission lease through native streams

luvs01:
## Summary

- Keep the native Chat active-turn admission lease alive for the full lifetime of a streamed SSE response.
- Release the lease on terminal output, malformed/truncated streams, request abort, client cancellation, or shutdown.
- Integrate release into the existing native Chat SSE parser instead of adding another async-pull wrapper on bundled Bun 1.3.14.

Exact base: `9830ab1c26014a5d6c4a4ec2b4d4b1802d4e0c79`  
Exact head: `fce54f5f9b9ef570352fcde31f9484eaaad73c3b`

## Why

`/v1/chat/completions` already acquires an active-turn lease before routing. Native Chat streaming previously returned the downstream body without transferring that lease, so the outer handler released admission as soon as it returned while the stream and upstream work were still active. Long-lived native streams could therefore fall out of the 256-turn admission cap.

This change transfers the lease only for requested native upstream SSE. The existing parser owns terminal and cancellation callbacks, so it can release the transferred lease without adding a second stream wrapper. The bounded JSON fallback remains unchanged because its upstream model response is already fully terminal before the handler returns.

Related: #820

## Behavior and compatibility

- Authentication and origin checks still run before admission and provider work.
- Normal `[DONE]`, provider terminal events, malformed/truncated SSE, request abort, reader cancellation, and shutdown all release exactly once.
- Direct/internal calls without an admission lease remain supported and do not increment release-miss metrics.
- Non-streaming and completed JSON-to-SSE fallback behavior is unchanged.
- No request, credential, account-routing, logging, or provider wire contract changes.

## Verification

- Bun 1.3.14: `tests/chat-completions-endpoint.test.ts` — 78 pass, 0 fail, 260 assertions.
- Bun 1.4.0-canary.1: same file — 78 pass, 0 fail, 260 assertions.
- `tsc --noEmit` — pass on both runtimes.
- `scripts/privacy-scan.ts` — pass on both runtimes.
- `git diff --check` — pass.
- Independent correctness, test-contract, and security reviews — CLEAN, no P0-P2 findings.
- Codex Security diff scan `87177ad2-03f0-4017-b864-b51d05f32960` — 0 findings.
- Stable patch ID across the dependency-only rebase: `eaf2fb19483b2f159480add368d2c7f2bba31cf1`.

The full repository suite is not claimed green; maintained exact-head CI is still required.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability for streamed chat responses by correctly managing active requests through completion, cancellation, and connection aborts.
  * Prevented lingering request activity after interrupted streams, helping maintain consistent service behavior.
  * Added safeguards for chat requests handled without an admission lease.
* **Tests**
  * Expanded coverage for streaming completion, cancellation, request aborts, and recovery scenarios.

### DIFF
--- a/src/server/chat-native.ts
+++ b/src/server/chat-native.ts
@@ -7,6 +7,7 @@ import {
   isChatCompletionsStreamError,
 } from "../chat/outbound";
 import { classifyError, CYBER_POLICY_ERROR_CODE, isCyberPolicyCode } from "../lib/errors";
+import type { AdmissionLease } from "../lib/admission";
 import { readBoundedResponseBody } from "../lib/bounded-body";
 import { redactSecretString } from "../lib/redact";
 import { resolveClientRetryAfter } from "../lib/retry-after";
@@ -39,6 +40,7 @@ import {
   type RequestLogContext,
 } from "./request-log";
 import { jsonCompletionSse, nativeChatSse, structuredError, usageFromChat } from "./chat-native-sse";
+import { registerTurn, unregisterTurn } from "./lifecycle";
 
 type Rec = Record<string, unknown>;
 
@@ -78,7 +80,7 @@ interface HandleNativeChatOptions {
   req: Request;
   config: OcxConfig;
   logCtx: RequestLogContext;
-  logIds?: { requestId: string; start: number };
+  logIds?: { requestId: string; start: number; turnAdmissionLease?: AdmissionLease };
   route: RouteResult;
   chatBody: Rec;
   requestedModel: string;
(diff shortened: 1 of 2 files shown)


luvs01:
@coderabbitai review

Wibias (approved):
Reviewed the full diff and surrounding lifecycle/streaming paths. I found no blocking correctness, lifecycle, security, memory, or compatibility issues. The admission lease is transferred only for requested native SSE streams, released idempotently on terminal/error/cancel/abort paths, and non-streaming behavior remains unchanged. Regression coverage and exact-head CI are green. Approving.

Wibias:
Thanks for this fix! 🙏 This is genuinely useful because native Chat SSE requests now keep their active-turn admission lease for the full lifetime of the stream instead of falling out of the concurrency cap as soon as the handler returns. That makes overload protection reflect the actual amount of in-flight work, and it also gives shutdown/cancellation paths much cleaner ownership semantics. Nice, focused reliability improvement.

PR #1876, opened 2026-08-17
fix(windows): keep catalog discovery off request event loop

### LINKED ISSUE #1852: [Bug][Windows]: synchronous PowerShell app-server enumeration blocks /healthz and proxy traffic on v2 requests
## Client or integration

Codex App

## Area

Proxy and routing / Platform (Windows)

## Summary

On Windows, OpenCodex performs synchronous PowerShell/CIM process enumeration on the proxy's main Bun event loop while building multi-agent v2 guidance.

The hot request path is:

1. `buildMultiAgentGuidance()` handles a v2 surface.
2. It awaits `defaultCollectCatalogState()`.
3. That calls the synchronous `collectCodexAppServerCatalogState()`.
4. `listWindowsSnapshots()` runs `execFileSync(powershell.exe, ..., timeout: 8_000)`.

Relevant source:

- [v2 guidance request path](https://github.com/lidge-jun/opencodex/blob/v2.22.0/src/server/responses/collaboration.ts#L203-L211)
- [per-v2-request catalog collection](https://github.com/lidge-jun/opencodex/blob/v2.22.0/src/server/responses/collaboration.ts#L256-L262)
- [synchronous Windows enumeration](https://github.com/lidge-jun/opencodex/blob/v2.22.0/src/codex/app-server-processes.ts#L351-L387)

On the affected machine, one enumeration blocks the Bun event loop for 3.8-5.1 seconds. During that interval, even loopback `/healthz` cannot be serviced. Under repeated v2 requests, these pauses can queue and make the proxy appear permanently hung: the Bun process remains alive and owns port 10100, but health probes time out and connections accumulate in `CLOSE_WAIT`.

I expected catalog-state discovery to be asynchronous, deduplicated/cached, or outside the request hot path so a slow CIM query cannot block unrelated proxy traffic and health checks.

This remains present in the latest stable package, 2.22.0, and in current `main`.

## Reproduction

With the repository dependencies available on Windows, run the following with Bun:

```ts
const { listWindowsSnapshots } =
  await import("./src/codex/app-server-processes.ts");

for (let i = 1; i <= 5; i++) {
  const scheduled = performance.now();
  const timer = new Promise<number>(resolve =>
    setTimeout(() => resolve(performance.now() - scheduled), 10)
  );

  const started = performance.now();
  let count: number | null = null;
  let error: string | null = null;
  try {
    count = listWindowsSnapshots().length;
  } catch (e) {
    error = e instanceof Error ? e.message : String(e);
  }

  const callMs = performance.now() - started;
  const timerMs = await timer;
  console.log({ run: i, callMs, timerMs, count, error });
}
```

A 10 ms timer is delayed by essentially the full synchronous call duration:

| Run | `listWindowsSnapshots` | 10 ms timer fired after | Result |
|---:|---:|---:|---|
| 1 | 3812 ms | 3813 ms | 3 processes |
| 2 | 3937 ms | 3943 ms | 3 processes |
| 3 | 4080 ms | 4082 ms | 3 processes |
| 4 | 5072 ms | 5075 ms | `windows_enum_incomplete` |
| 5 | 4402 ms | 4406 ms | 3 processes |

The timer evidence makes the event-loop blocking deterministic without requiring an upstream provider request.

For the live proxy symptom:

1. Leave `multiAgentGuidanceEnabled` at its effective default (`true`).
2. Use Codex App with the multi-agent v2 surface and issue repeated turns/tool continuations.
3. Probe `http://127.0.0.1:10100/healthz` concurrently.
4. Observe a PowerShell child under the OpenCodex Bun process running the `Get-CimInstance Win32_Process` / `GetOwner` script.
5. Health requests intermittently take several seconds or time out despite the listener and Bun PID remaining alive.
6. Setting `multiAgentGuidanceEnabled: false` removes this request-path enumeration. In the observed session, two post-change probe batches passed 45/45 and no enumeration child remained.

A live stalled sample before the workaround showed:

- `/healthz`: 5-second timeout, zero response bytes
- Bun CPU increase over 2 seconds: 0.016 s (waiting, not spinning)
- Bun private memory: about 1.69 GiB
- listener still active on port 10100
- multiple `CLOSE_WAIT` connections
- a PowerShell child running the catalog process enumeration

These are observations from one affected installation; the deterministic timer reproduction above is the primary evidence.

## Root cause

The `async` wrapper does not make the synchronous collector non-blocking. `execFileSync` prevents Bun from advancing timers, accepting/servicing health requests, or processing other proxy work until PowerShell exits or reaches the 8-second timeout.

The enumeration is especially expensive because it walks `Win32_Process` and calls `GetOwner` for candidates. Issue #1354 already measured multi-second Windows enumeration and fixed a correctness race in the same collector, but did not remove it from the event loop or request hot path.

This is analogous to #612 / #645, where synchronous `icacls` work blocked `/healthz` until that high-frequency path was made asynchronous.

## Suggested safe direction

1. Use asynchronous process creation for Windows enumeration, or move it to a worker so the proxy event loop remains responsive.
2. Deduplicate concurrent discovery with a single in-flight operation.
3. Cache the last result with a short TTL / stale-while-revalidate policy. Catalog-state guidance is advisory; serving a recent/unknown state is safer than blocking all proxy traffic.
4. Add a regression test that schedules a short timer or health request while the process-enumeration seam is deliberately slow and verifies it is serviced before enumeration completes.
5. Optionally narrow the CIM query before owner lookups, but query optimization alone would not remove the event-loop blocking contract.

## Workaround

```json
{
  "multiAgentGuidanceEnabled": false
}
```

This disables OpenCodex's additional multi-agent guidance/roster injection, not Codex's built-in collaboration tools.

## Version

OpenCodex 2.22.0, bundled Bun 1.3.14, Codex CLI 0.147.0.

## Operating system

Windows 11 Pro, version 10.0.26200 (build 26200).

## Provider and model

Not provider-specific. The blocking occurs before provider work while building v2 guidance.

## Logs or error output

```text
GET /healthz -> timeout after 5.0s, 0 bytes
PowerShell child -> Get-CimInstance Win32_Process ... Invoke-CimMethod GetOwner
listWindowsSnapshots() -> 3.8-5.1s synchronous wait
```

Not required; the source links, deterministic timer reproduction, timings, and live probe observations are text-reproducible.

## Redacted configuration

```json
{
  "multiAgentGuidanceEnabled": true
}
```

## Related issues

- #1354: correctness/race problem in the same Windows catalog-state collector
- #612 / #645: precedent for moving a synchronous Windows child-process wait off the proxy event loop

## Checks

Ingwannu:
## Summary

- move Windows PowerShell/CIM app-server enumeration off Bun's request event loop for v2 guidance;
- keep the existing synchronous fail-closed collector for explicit CLI/service lifecycle operations;
- share concurrent request probes by collection identity and cache the advisory result for five seconds;
- generation-guard invalidation so a pre-write CIM flight cannot repopulate post-write catalog state;
- document the request/lifecycle split and add event-loop, single-flight, TTL, and invalidation regressions.

Closes #1852

## Why this boundary

The stale-catalog observation is advisory on the v2 request path, but the same process evidence is also used by explicit lifecycle operations. Replacing the synchronous API globally would widen a process-control boundary. This PR therefore adds a Windows-only asynchronous request collector while preserving the existing matching, owner verification, trusted System32 PowerShell resolution, and synchronous lifecycle behavior.

## Verification

- `bun test tests/codex-app-server-processes.test.ts tests/multi-agent-compat.test.ts` — 88 pass, 1 platform skip, 0 fail.
- `cd docs-site && bun install --frozen-lockfile && bun run build` — 385 pages built successfully.
- Full suite attempted with CPU limits: 12,620 pass, 15 skip, 10 fail, 7 errors across 12,645 tests. The run took 944.69 seconds on a busy six-core host. The observed failures were outside this diff:
  - `lab-live-pinned-timeouts` and `bridge-lifecycle` timing cases passed when rerun alone on both this head and clean `origin/dev` (18/18 each);
  - `codex-shim` reproduced identically on this head and clean `origin/dev` because this host's installed service token overrides the fixture's `local-secret` (68 pass, 1 baseline failure);
  - the remaining errors were worktree GUI dependency-resolution failures from the initial root-only `node_modules` link, not files changed here.

## dev2-go

No Go counterpart exists for this change: Codex app-server process discovery and v2 guidance assembly remain in the TypeScript control plane on the Go transition line. This PR does not change the Go native data path.

## Review notes

This is self-authored maintainer work, so I will not approve or merge it myself. Exact-head cross-platform CI and another maintainer review are required.

## Summary by CodeRabbit

* **Performance**
  * Improved Windows request handling by moving advisory process checks off the main event loop.
  * Added brief caching and shared in-progress checks to reduce repeated system queries.

* **Reliability**
  * Slow or failed Windows checks no longer block health checks or unrelated proxy traffic.
  * Prevented outdated process information from replacing newer results.
  * Preserved fail-closed behavior for CLI and service lifecycle operations.

* **Documentation**
  * Documented Windows process-discovery, caching, and advisory-check behavior.

### DIFF
--- a/docs-site/src/content/docs/guides/sub-agent-surface.md
+++ b/docs-site/src/content/docs/guides/sub-agent-surface.md
@@ -260,7 +260,7 @@ separately by `ocx doctor`. `stale` clears only after every detected Codex app-s
 the final catalog write; it does not necessarily clear `unknown`.
 
 On Windows, this advisory check uses asynchronous PowerShell/CIM discovery on the v2 request path.
-Concurrent cold checks share one in-flight discovery and successful results are cached briefly. A
+Concurrent cold checks share one in-flight discovery and results are cached briefly. A
 slow or failing CIM query can delay or suppress only OpenCodex-authored model guidance; it does not
 block the Bun event loop, `/healthz`, or unrelated proxy traffic. Explicit CLI/service lifecycle
 operations retain the synchronous, fail-closed process collector because they may signal processes.
--- a/docs-site/src/content/docs/guides/sub-agent-surface.md
+++ b/docs-site/src/content/docs/guides/sub-agent-surface.md
@@ -259,6 +259,12 @@ timestamp, an unreadable process start time, or a failed process enumeration —
 separately by `ocx doctor`. `stale` clears only after every detected Codex app-server starts after
 the final catalog write; it does not necessarily clear `unknown`.
 
+On Windows, this advisory check uses asynchronous PowerShell/CIM discovery on the v2 request path.
+Concurrent cold checks share one in-flight discovery and successful results are cached briefly. A
+slow or failing CIM query can delay or suppress only OpenCodex-authored model guidance; it does not
+block the Bun event loop, `/healthz`, or unrelated proxy traffic. Explicit CLI/service lifecycle
+operations retain the synchronous, fail-closed process collector because they may signal processes.
+
 Only a real change counts. A sync whose result is byte-identical to the catalog already on disk
 leaves the file untouched, so restarting the proxy or re-syncing an unchanged model set does not
 make a running Codex look stale.
--- a/docs-site/src/content/docs/guides/sub-agent-surface.md
+++ b/docs-site/src/content/docs/guides/sub-agent-surface.md
@@ -260,7 +260,8 @@ separately by `ocx doctor`. `stale` clears only after every detected Codex app-s
 the final catalog write; it does not necessarily clear `unknown`.
 
 On Windows, this advisory check uses asynchronous PowerShell/CIM discovery on the v2 request path.
-Concurrent cold checks share one in-flight discovery and results are cached briefly. A
+Concurrent cold checks share one in-flight discovery. Observed states are cached for five seconds;
+an `unknown` failure is cached for only 250 milliseconds so a transient CIM error retries quickly. A
 slow or failing CIM query can delay or suppress only OpenCodex-authored model guidance; it does not
 block the Bun event loop, `/healthz`, or unrelated proxy traffic. Explicit CLI/service lifecycle
 operations retain the synchronous, fail-closed process collector because they may signal processes.
--- a/src/codex/app-server-processes.ts
+++ b/src/codex/app-server-processes.ts
@@ -450,12 +450,32 @@ export function listWindowsSnapshots(runPowerShell?: (psCommand: string) => stri
   return parseWindowsSnapshotOutput(output);
 }
 
-async function listWindowsSnapshotsAsync(): Promise<ProcessSnapshot[]> {
-  const output = await execFileTextAsync(resolveTrustedWindowsPowerShellExe(), [
+/**
+ * Exec seam for the async enumeration. Without it the default request path is
+ * untestable off Windows: `resolveTrustedWindowsPowerShellExe()` fails immediately, so a
+ * synchronous and an asynchronous implementation are indistinguishable from a test — which
+ * is exactly how a test asserting "yields to the event loop" stayed green after the async
+ * call was reverted to `execFileSync`. Overriding this lets a test drive the real default
+ * wiring and observe whether the event loop keeps running during enumeration.
+ */
+let windowsSnapshotExecAsync: (command: string, timeoutMs: number) => Promise<string> =
+  (command, timeoutMs) => execFileTextAsync(resolveTrustedWindowsPowerShellExe(), [
     "-NoProfile", "-NoLogo", "-NonInteractive",
     "-Command",
-    windowsSnapshotPowerShellCommand(),
-  ], 8_000);
+    command,
+  ], timeoutMs);
+
+/** Test-only: swap the async enumeration exec. Returns a restore function. */
+export function setWindowsSnapshotExecAsyncForTest(
+  exec: (command: string, timeoutMs: number) => Promise<string>,
+): () => void {
+  const previous = windowsSnapshotExecAsync;
+  windowsSnapshotExecAsync = exec;
+  return () => { windowsSnapshotExecAsync = previous; };
+}
+
+async function listWindowsSnapshotsAsync(): Promise<ProcessSnapshot[]> {
+  const output = await windowsSnapshotExecAsync(windowsSnapshotPowerShellCommand(), 8_000);
   return parseWindowsSnapshotOutput(output);
 }
 
--- a/src/codex/app-server-processes.ts
+++ b/src/codex/app-server-processes.ts
@@ -919,7 +919,12 @@ export async function collectCodexAppServerCatalogStateForRequest(
   return flight.promise;
 }
 
-/** Test hook: drop the memoized catalog state. */
+/**
+ * Drop memoized catalog state after a relevant catalog/cache write and before
+ * the post-write state read. Advancing the generation prevents an older
+ * in-flight Windows CIM refresh from publishing its pre-write result after the
+ * write has completed.
+ */
 export function resetCodexAppServerCatalogStateCache(): void {
   catalogStateCache = null;
   requestCatalogStateGeneration += 1;
--- a/src/codex/app-server-processes.ts
+++ b/src/codex/app-server-processes.ts
@@ -900,9 +900,25 @@ export async function collectCodexAppServerCatalogStateForRequest(
   }));
   let flight: RequestCatalogStateFlight;
   const promise = pending.then(status => {
-    // A catalog write can invalidate while slow CIM is still running. Never
-    // let that pre-write result repopulate the post-write cache.
-    if (requestCatalogStateGeneration === generation && requestCatalogStateFlight === flight) {
+    // A catalog write can invalidate while slow CIM is still running. The result
+    // describes the PRE-write world, so it must neither repopulate the post-write
+    // cache nor reach the caller.
+    //
+    // Suppressing only the cache write is not enough. The awaiting request still
+    // received `fresh`, and `fresh` is the one state that authorizes positive
+    // guidance (`src/server/responses/collaboration.ts:279-280` returns null for
+    // `stale`/`unknown` but describes the catalog for `fresh`). So the request
+    // would advertise the newly written disk catalog to an app-server whose
+    // in-memory copy the write just made stale — a wrong answer, which is worse
+    // than the slow answer this whole change exists to fix.
+    //
+    // Degrade to `unknown` instead: it is the honest description of what an
+    // invalidated observation knows, and the guidance path already treats it as
+    // "say nothing positive".
+    if (requestCatalogStateGeneration !== generation) {
+      return { state: "unknown" as const, processes: [], catalogMtimeMs: null };
+    }
+    if (requestCatalogStateFlight === flight) {
       requestCatalogStateCache = {
         generation,
         identity,
--- a/src/codex/app-server-processes.ts
+++ b/src/codex/app-server-processes.ts
@@ -450,32 +450,12 @@ export function listWindowsSnapshots(runPowerShell?: (psCommand: string) => stri
   return parseWindowsSnapshotOutput(output);
 }
 
-/**
- * Exec seam for the async enumeration. Without it the default request path is
- * untestable off Windows: `resolveTrustedWindowsPowerShellExe()` fails immediately, so a
- * synchronous and an asynchronous implementation are indistinguishable from a test — which
- * is exactly how a test asserting "yields to the event loop" stayed green after the async
- * call was reverted to `execFileSync`. Overriding this lets a test drive the real default
- * wiring and observe whether the event loop keeps running during enumeration.
- */
-let windowsSnapshotExecAsync: (command: string, timeoutMs: number) => Promise<string> =
-  (command, timeoutMs) => execFileTextAsync(resolveTrustedWindowsPowerShellExe(), [
+async function listWindowsSnapshotsAsync(): Promise<ProcessSnapshot[]> {
+  const output = await execFileTextAsync(resolveTrustedWindowsPowerShellExe(), [
     "-NoProfile", "-NoLogo", "-NonInteractive",
     "-Command",
-    command,
-  ], timeoutMs);
-
-/** Test-only: swap the async enumeration exec. Returns a restore function. */
-export function setWindowsSnapshotExecAsyncForTest(
-  exec: (command: string, timeoutMs: number) => Promise<string>,
-): () => void {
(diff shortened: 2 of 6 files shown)


Ingwannu:
Follow-up pushed on exact head `0178c46f7`.

Applied the valid review findings:

- documented why generation invalidation blocks an older pre-write CIM refresh from publishing after a catalog/cache write;
- added a rejecting async CIM regression and pinned the intended five-second `unknown` cache, which prevents a broken Windows CIM provider from spawning one PowerShell process per request;
- exercised promise-returning `stale` and `unknown` catalog guidance through the collaboration boundary;
- corrected the public docs to say all advisory results are cached briefly.

I kept the macOS collector unchanged because #1852 is specifically the Windows synchronous PowerShell/CIM request-path block; expanding the Darwin `ps` path would be a separate cross-platform behavior change.

Validation on this head: focused tests 89 pass / 1 platform skip, typecheck clean, privacy scan passed, docs build passed, and `git diff --check` clean. Waiting for exact-head CI and another maintainer review; I will not self-approve or self-merge.

Wibias (changes requested):
Requesting changes on exact head `d5acd74142f36c2337544c25beceb06f34e92507`.

Two blockers remain:

1. **Top-level CIM enumeration can fail open as `not_running`.** `windowsSnapshotPowerShellCommand()` sets `$ErrorActionPreference='SilentlyContinue'`, while `Get-CimInstance Win32_Process` is outside the per-process `try/catch`. A non-terminating top-level CIM failure can therefore produce no output, `parseWindowsSnapshots()` returns `[]`, and the request collector reports `not_running` instead of `unknown`. That can incorrectly allow positive disk-derived v2 guidance when process state is actually unknown. Please make the top-level query fail closed, for example with `-ErrorAction Stop` plus an outer catch/sentinel or by allowing the async child-process call to reject, and add a regression specifically covering this top-level CIM failure.

2. **The full Windows test suite has not run on this exact head.** Normal PR CI is green, but `platform-windows` is `workflow_dispatch`-only and is accepted as skipped by the aggregate CI. Since this PR changes Windows PowerShell/CIM request-path behaviour, please run the full Windows suite on the resulting exact head after the fix.

The async request-path boundary, single-flight cache, generation invalidation, negative-cache contract, collaboration integration, and documentation otherwise look solid. The older CodeRabbit complaint about caching `unknown` is no longer a blocker because the latest head intentionally documents and tests that five-second negative cache.

lidge-jun:
Blocker 1 is now fixed on `dev` independently of this PR, in #1925 (merge `aa9df919a`).

The top-level `Get-CimInstance Win32_Process` now carries `-ErrorAction Stop` inside an
outer catch that emits `__OCX_ENUM_INCOMPLETE__`, so a query that fails under
`SilentlyContinue` reaches the collector as `unknown` instead of as clean-empty output
that reads exactly like a healthy idle machine. There is a regression for the specific
path that was untested — a top-level query returning cleanly empty, as distinct from the
existing coverage which drives a *throwing* enumerator by swapping `platform`.

Two related things landed with it that this PR will inherit on rebase:

- `collectCodexAppServerCatalogState` wrapped only the default enumerator in its
  try/catch, so an injected `listSnapshots` that threw propagated instead of degrading
  to `unknown`. Both paths now share one catch — the shape
  `src/codex/log-guard/processes.ts` already had.
- The state memo used a single 5s TTL for every state. `unknown` now gets 250ms, since
  caching a failure to observe for as long as an observation suppresses the retry that
  would have succeeded.

**What this does not do is close #1852.** That issue is about the synchronous
`execFileSync` blocking the Bun event loop for 3.8–5.1s so that even loopback
`/healthz` cannot be serviced — and that is precisely what *this* PR fixes. The
fail-open was a correctness bug found while reviewing this work; the blocking is the
reported defect, and it is still open.

So this PR is still wanted. Please rebase onto current `dev` (`aa9df919a`); the
PowerShell string and the collector's catch have both moved, so expect a conflict in
`src/codex/app-server-processes.ts`. Blocker 2 — a full Windows suite on the exact head —
still stands, and it stands harder now: the new tests drive an injected PowerShell
runner, so nothing in CI has ever parsed the emitted script.

Ingwannu:
Rebased onto current `dev@aaf04690e` and resolved the PowerShell conflict on exact head `125156c3e`.

The rebase preserves the independently landed fail-closed contract from #1925:

- top-level `Get-CimInstance` keeps `-ErrorAction Stop` and the outer sentinel catch;
- `parseWindowsSnapshotOutput` and the injected PowerShell-runner test seam remain intact;
- both synchronous and asynchronous Windows collectors now share that same command/parser;
- request-path `unknown` caching uses the upstream 250ms policy rather than the old five-second observation TTL.

Validation on this head:

- focused Windows request/sentinel tests: 6/6 pass;
- async collaboration guidance regression: 1/1 pass;
- typecheck clean, privacy scan passed, docs build passed, diff check clean.

The broader two-file run was 92 pass / 1 platform skip / 1 failure. That failure reproduces from current `dev`: the `sync-cache` source assertion still expects `invalidateCodexModelsCacheWithPermit(permit, owningCodexHome)` while dev now calls it with `{ allowWhenDesiredDisabled: true }`. I did not fold that unrelated dev-head stabilization into this PR; #2026 is already the owner stabilization branch.

Waiting for exact-head CI, including the real Windows runner, and another maintainer review. I will not self-approve or self-merge.

lidge-jun:
Campaign validation status (260818 bug-PR resolution): head 125156c3e is rebased onto current dev (all four fail-closed snapshot commits are ancestors), and the earlier CHANGES_REQUESTED blocker 1 (top-level CIM fail-open) is resolved by dev itself — the async collector maps a rejected enumeration to unknown, never not_running. Per the review's exact-head Windows-suite condition, the platform-windows leg has been dispatched on the landed candidate (codex/land-1876, merge of this head onto dev): https://github.com/lidge-jun/opencodex/actions/runs/32145700019. TTL note for the record: both numbers coexist by design — 250ms is the unknown-state negative cache (CATALOG_STATE_UNKNOWN_TTL_MS, #1947 policy), 5s the positive advisory cache. @Wibias when the Windows leg is green, could you re-review or dismiss the stale review so this can land?

lidge-jun:
Windows-leg result on the landed candidate (codex/land-1876, run 32145700019): FAILURE — but a control dispatch on dev itself (e446607c8, run 32147924436) fails the same way: the failing suites are the Codex Log Guard / CodeRabbit-protection / WS-relay families on windows-latest, unrelated to this PR's files (zero failures in app-server-processes / catalog-discovery suites on either run; each run also had one cancelled shard, so per-suite diffs across runs are shard-assignment artifacts). The push-event Cross-platform CI on the current dev head is green (run 32147799485, success). Conclusion: the exact-head Windows-suite condition cannot currently be greener than dev's own baseline — the Windows dispatch leg is red on dev independently of this PR. Holding the merge for your re-review/dismissal per the standing CHANGES_REQUESTED; from the campaign's side this candidate is validated (rebased on dev, fail-closed API ancestors verified, both TTL constants honored).

lidge-jun:
## 리뷰 · 우선순위 63 / 80

`dev` 기준 이 PR은 Windows v2 요청 경로의 PowerShell/CIM 탐색을 `execFile` 비동기로 빼고, CLI/서비스용 `collectCodexAppServerCatalogState` 동기 fail-closed는 그대로 둡니다. `src/server/responses/collaboration.ts`의 `defaultCollectCatalogState`만 `collectCodexAppServerCatalogStateForRequest`를 await합니다. Draft/`hygiene-blocked`가 아니고 conflict도 없습니다. `windowsSnapshotPowerShellCommand`와 `parseWindowsProcessStartTimes`를 공유해 매칭·owner·trusted System32 경로를 두 벌로 나누지 않은 점은 맞습니다.

요청 경로 캐시는 generation + identity + TTL입니다. 성공은 5초, `unknown`은 기존 `catalogStateTtlMs`의 짧은 TTL(테스트 기준 250ms)입니다. 동일 identity의 in-flight promise를 공유하고, `resetCodexAppServerCatalogStateCache`가 generation을 올려 캐시/flight를 비웁니다. then 콜백은 `requestCatalogStateGeneration === generation && requestCatalogStateFlight === flight`일 때만 캐시에 써서, 테스트가 검증한 대로 pre-write CIM 결과가 post-write 캐시를 채우지 않습니다. identity는 `listSnapshots`/`listSnapshotsAsync`/`now` 등 함수 참조 비교라 프로덕션 기본 `io={}`에서는 안정적입니다.

비동기 열거의 fail-closed는 `__OCX_ENUM_INCOMPLETE__`를 `parseWindowsSnapshotOutput`이 throw하고, `refresh`의 catch가 `unknown`으로 바꾸는 기존 계약에 기대어 있습니다. 빈 스냅샷은 `not_running`입니다. 다만 `reset`은 이미 뜬 PowerShell을 kill하지 않습니다. 무효화된 flight는 호출자에게는 옛 status를 돌려주고 캐시만 안 채웁니다. 카탈로그 write 직후 동시 요청이 많으면 CIM이 겹칩니다. `execFileTextAsync`는 timeout/error면 reject → `unknown`이라 요청 경로는 fail-closed이지만, 자식 프로세스가 남는 시간은 그대로입니다.

`windowsProcessStartPowerShellCommand`는 `ProcessId=${pid}`를 문자열로 붙입니다. pid는 자체 CIM 파서가 `Number.isSafeInteger`로 걸러 낸 값이라 프로덕션 주입 면은 좁습니다. 테스트 seam으로 임의 pid를 넣으면 필터 문자열이 열립니다. Windows가 아닌 플랫폼은 요청 경로도 동기 collector를 쓰므로, 이 PR의 event-loop 주장은 win32에만 해당합니다. 작성자가 적은 full suite 10 fail/7 error는 이 diff 밖이라고 하나, Ready 전 재확인이 필요합니다.

**해결방안:** (1) `resetCodexAppServerCatalogStateCache`에서 in-flight `execFile`를 abort할 수 있으면 timeout 전에 자식을 죽이십시오. 못 하면 generation 무효화만으로는 프로세스 누적이 남는다는 점을 문서에 명시하십시오. (2) start-time 필터에 pid를 넣기 전에 다시 `Number.isSafeInteger(pid) && pid > 1`을 거르십시오. (3) `listSnapshots`만 있고 `listSnapshotsAsync`가 없는 io가 요청 경로에서 동기로 막히지 않게, 프로덕션 기본 경로만 async임을 주석/테스트로 고정하십시오. (4) `/healthz`가 CIM을 기다리지 않는다는 주장을 이벤트 루프 테스트와 별도로 한 번 더 고정하십시오. 경계 자체는 #1852에 맞고 merge 후보는 됩니다.

이 댓글은 grok-bot이 작성했습니다

Issue #1849, opened 2026-08-16
[Bug][Windows] Failed ocx update leaves a file-less package skeleton; service wrapper restart-loops 1,009 times with no integrity check or backoff


stephen-drew:
## Summary

On Windows 11, an `ocx update` from 2.13.0 left the global package directory containing **107 directories and zero files**.
Every launcher (`ocx`, `ocx.cmd`, `ocx.ps1`, `opencodex`, `opencodex.cmd`, `opencodex.ps1`) was removed from
`%APPDATA%\npm`, so no documented recovery command was reachable from a normal shell.

The Task Scheduler service wrapper then restarted the missing entry point every 5 seconds for 89 minutes —
**1,009 consecutive `child exited with code 3` events** — until the package was manually rolled back.

This matches the partial-install behaviour described in #281. That issue was closed with the permanent-crash symptom
attributed to an external monitoring script terminating the restarted proxy. **That attribution does not fit this case:**
no external process was involved. The wrapper was relaunching a package that had no `bin/ocx.mjs`, no
`src/cli/index.ts`, and no bundled `bun.exe` to execute.

## Environment

| Item | Value |
| --- | --- |
| OS | Windows 11 Home, 10.0.26100 build 26100, AMD64 |
| Node | v24.15.0 |
| npm | 11.19.0 |
| opencodex (before/after rollback) | 2.13.0 |
| Bundled Bun | 1.3.14 |
| Service backend | Task Scheduler (`opencodex-proxy`), wrapper `opencodex-service.cmd` |
| Install type | npm global, `%APPDATA%\npm\node_modules\@bitkyc08\opencodex` |

### 1. The abandoned install contained directories only

The failed package was preserved before rollback. Recursive enumeration:

```
TotalDirs   : 107
TotalFiles  : 0
```

95 dependency folders were created under `node_modules` (`@bufbuild`, `@hono`,
`@modelcontextprotocol`, `@napi-rs`, `@oven`, `bun`, ...). The `node_modules/bun` folder
exists and is **completely empty**. Missing at the package root: `package.json`,
`bin/ocx.mjs`, `src/cli/index.ts`, `node_modules/bun/bin/bun.exe`.

So npm created the directory tree, then unpacked no file content and did not roll back.

### 2. Launchers were removed, recovery commands unreachable

`%APPDATA%\npm\ocx.cmd` and all sibling launchers were gone. What remained were npm's
rename-aside shims, which still pointed at the now-nonexistent target:

```
[...]

PR #1877, opened 2026-08-17
fix(windows): stop service loop on missing install

Ingwannu:
## Summary

- verify the Task Scheduler wrapper's baked Bun runtime and CLI entry before every child start;
- log one actionable incomplete-install message and exit when either path is absent;
- preserve the existing five-second restart behavior for a proxy child that actually launched and then failed;
- document the recovery command and the reason this slice does not attempt an unsafe npm directory copy-back.

Partial fix for #1849: this closes the unbounded missing-file restart loop. Transactional npm package-and-shim rollback remains separate follow-up work.

## Why this scope

A missing executable cannot become healthy inside the same wrapper process, so retrying it 1,009 times adds log churn without recovery. A generic crash ceiling would also stop legitimate recovery after unrelated intermittent runtime failures. The wrapper therefore treats only proven missing baked paths as terminal.

The current `dev` updater already has stop-first and registry-integrity preflight that were not present in the reported 2.13.0 build. It still does not have a verified transactional restore of both the npm package tree and global launchers. Copying a package directory alone would not safely restore shims, ownership, and lock state, so that broader design is intentionally not guessed in this PR.

## Verification

- `bun test tests/service.test.ts` — 124 pass, 0 fail.
- `cd docs-site && bun install --frozen-lockfile && bun run build` — 385 pages built successfully.

## dev2-go

No separate Go-native counterpart exists for this behavior; the Windows service asset generator remains in the TypeScript lifecycle control plane.

## Review notes

This is self-authored maintainer work, so I will not approve or merge it myself. Exact-head CI, the service-lifecycle workflow, and another maintainer review are required.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Windows service startup validation by detecting incomplete installations before launch.
  - Services now stop cleanly with an actionable message instead of repeatedly retrying when required files are missing.
  - Existing restart behavior remains available when a service starts successfully but later crashes.

- **Documentation**
  - Added guidance explaining how incomplete Windows service installations are handled and repaired.


Wibias (approved):
Approved. The macOS `provider request pacing queue` failure is unrelated to this PR: #1877 does not touch request-pacing code/tests, while the service-lifecycle workflow and focused service changes are green. The branch is currently 14 commits behind `dev`, so it still needs to be updated and exact-head CI rerun before merge.

Issue #1798, opened 2026-08-15
ocx stop/restore leave unmarked openai_base_url + proxy models_cache behind when Codex app rewrote config after injection


fede-oss:
## Client or integration

Codex App (and Codex CLI)

### Version

2.21.0 (npm, @bitkyc08/opencodex)

### Operating system

macOS 15.7.9 (build 24G830), Apple Silicon (arm64); Codex desktop app 151.0.7922.137

---
## Area

Service lifecycle (`ocx stop` / `ocx restore`), Catalog / models

## Summary

`ocx stop` (and `ocx restore`) leave the injected routing in place when the Codex app rewrote `~/.codex/config.toml` after injection: the root key `openai_base_url = "http://127.0.0.1:10100/v1"` and `[features] fast_mode = true` stay in the config, and the proxy-routed `models_cache.json` is not restored. Result: plain Codex keeps sending every request to a dead `127.0.0.1:10100` and all messages fail with connection refused, while the model picker still lists proxy-routed models (e.g. `opencode-go/deepseek-v4-flash`, `qwen-cloud/...`) and `gpt-5.3-codex` is replaced by `gpt-5.3-codex-spark`.

Expected: `ocx stop` / `ocx restore` remove **all** injected routing (including an unmarked root `openai_base_url`) and restore the native `models_cache.json` from the catalog backup, so plain Codex works with no manual cleanup.

## Reproduction

Environment: macOS (darwin arm64), opencodex 2.21.0 (npm), Codex desktop app 151.0.7922.137. Loopback routing (Design B: built-in `openai_base_url` override).

1. `ocx start` → `~/.codex/config.toml` gets `openai_base_url = "http://127.0.0.1:10100/v1"` (root) and `[features] fast_mode = true`. `~/.codex/opencodex-journal.json` records `originalConfig` + `injectedConfigHash`.
2. Open the Codex desktop app and change the model (or any setting). The app rewrites `~/.codex/config.toml`, so `sha256(config.toml)` no longer equals `injectedConfigHash` and the `model` value differs from the injected snapshot.
3. Run `ocx stop`. The service/proxy stop cleanly, but the config and cache are left injected:
   - `restoreJournalState()` skips the journal restore because the config changed since injection (`configUnchanged` is false).
[...]

PR #1881, opened 2026-08-17
fix(windows): make the suite pass on an unelevated Windows checkout

lidge-jun:
## Summary

Makes the full Bun suite pass on an unelevated Windows checkout. What started as one `icacls` fix turned into a sweep: most failures were **Windows-only defects in production code**, and the rest were tests describing the machine instead of the code.

**Production defects**

- `src/codex/user-identity.ts` resolved LocalAppData through .NET `GetFolderPath(LocalApplicationData)`, which follows `USERPROFILE` and returns an **empty string** when that profile has no AppData on disk. Every coordinator lookup then refused, cascading into locking, transition-state, catalog serialization and sync. Now uses `SHGetKnownFolderPath(FOLDERID_LocalAppData, KF_FLAG_DEFAULT_PATH, token=0)`, which ignores the environment. A non-null token is not equivalent: `(HANDLE)-1` resolves the Default profile.
- `src/codex/injected-marker.ts` returned raw bytes between quotes without decoding TOML escapes, so a Windows path read back with doubled backslashes and the journal could not restore the catalog it wrote (#1798). `paths.ts` already had the correct reader; both now share it.
- `src/lab/projection/rebuild.ts` closed its database without finalizing prepared statements. On Windows that holds the file open, so a second rebuild could not unlink the projection it was replacing — ten Compatibility Lab failures from one leak.
- `src/lib/windows-secret-acl.ts`: `icacls` resolved from a trusted System32 path, plus a typecheck fix where a `ReturnType<typeof Bun.spawn>` annotation widened the stdio types.
- Identity lookups are memoized per process: they spawned PowerShell on every config write and lock acquisition (~510ms per coordinator path resolution, now ~1ms).

**Test-side fixes**

- `scripts/test.ts` pins `GIT_CONFIG_GLOBAL` before moving HOME, so git still sees `safe.directory`; without it every `git` call failed with "dubious ownership" on a checkout owned by another account.
- `tests/core-lab-boundary.test.ts` built its repo root from `URL.pathname` (`/C:/...` on Windows) and matched a literal `/src/lab/`. **The guard was inert on Windows** — it would have reported clean for a real Lab import. Its own adversarial cases now fail before the fix and pass after it.
- Marker publication, teardown races, POSIX mode assertions, symlink-privilege cases, and genuine multi-process budgets, each addressed in its own commit.

Two commits fix source-shape checks that the current `dev` tip left behind; both were verified failing on `origin/dev` independently of this branch.

**No GUI behaviour changes.** The only `gui/` file touched is `gui/tests/models-native-group-controls.test.ts`, whose assertion pinned a one-line expression that `dev` has since wrapped — a text match, not a rendered surface, so there is nothing to screenshot.

## Verification

- Full suite on this Windows machine, run in batches of 60 files: **806/806 files, 0 failures.** Batching is a Bun workaround — `bun test --isolate ./tests/` reaches ~3.5GB RSS and panics near the end of an 806-file run, after every test has already passed.
- Fixed files were each re-run 3x consecutively; flakiness was treated as failure.
- `bun x tsc --noEmit` (strict, tsc 7) green.
- `bun run privacy:scan` green.
- Skips are explicit and only where Windows withholds a privilege (file symlinks need Developer Mode) or where an assertion is POSIX-only (Windows synthesizes mode bits).

## Checklist

### DIFF
--- a/gui/tests/models-native-group-controls.test.ts
+++ b/gui/tests/models-native-group-controls.test.ts
@@ -60,7 +60,12 @@ test("the native group keeps its window readable with the cap switched off", asy
   expect(src).toContain("{(capOn || nativeProviderGroup) && (");
   // With the cap off the stored value is only what a future toggle would apply — the 350k
   // default — so the display falls back to the widest window the rows actually advertise.
-  expect(src).toContain("const capDisplayValue = capOn ? providerCap : (widestRowWindow ?? providerCap);");
+  // Matched as separate fragments because the expression is wrapped across lines now, and
+  // it grew a native branch: with the cap off the native group shows its default window
+  // rather than the widest advertised row. A single-line literal pinned the formatting
+  // instead of the behaviour and broke on the reflow that introduced that branch.
+  expect(src).toContain("const capDisplayValue = capOn");
+  expect(src).toContain("nativeProviderGroup ? NATIVE_GPT56_DEFAULT_WINDOW : (widestRowWindow ?? providerCap)");
   // The select is inert until the cap is actually on: showing a number is not the same as
   // offering to change one.
   expect(src).toContain("disabled={busy || !capOn}");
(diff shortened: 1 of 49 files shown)


lidge-jun:
This does not change the GUI: the only gui/ file in the diff is gui/tests/models-native-group-controls.test.ts, a source-text assertion that pinned a one-line expression the dev tip has since wrapped. No rendered surface changed, so there is nothing to screenshot.

PR #1844, opened 2026-08-16
fix(cursor): teach the code-mode nested-helper contract in tool guidance

lidge-jun:
## Summary

Absorbs #1801 onto the #1817 contract and closes it as superseded.

Codex code mode advertises one freeform `exec` whose body is JavaScript evaluated in a V8 isolate. Shell, file edits and MCP are reachable only as nested `await tools.<name>(...)` helpers described inside that tool, so the flat shell-bridge guidance sent the model after a top-level `exec_command` that does not exist in that turn.

While absorbing, one correctness defect in #1801's draft was fixed: `cursorRequestUsesCodeMode()` returns true whenever freeform `exec` is visible without a bare shell bridge, which does **not** mean `exec` is the only visible tool. The original wording ("they are not separate top-level tools") therefore told the model that a separately advertised tool such as `mcp__fs__read_file` was non-callable, contradicting the catalog prefix listing it. The note now scopes the nested-helper claim to the helpers `exec` describes and names the other visible top-level tools as still callable.

Also carries the devlog unit `devlog/_plan/260816_wave012_closeout/` planning Wave 0/1/2 of the open bug triage roadmap.

Closes #1801.

## Verification

- `bun test tests/cursor-tool-definitions.test.ts` — 24 pass / 0 fail, including the three tests ported from #1801 (freeform/bare-bridge mutual exclusivity, V8-isolate nested-helper contract, flat-catalog branch preserved) and a new mixed-catalog regression asserting a co-visible top-level tool stays callable.
- `bun x tsc --noEmit` — clean.

## Checklist

- [x] Focused regression test added near the existing tests for that subsystem
- [x] Typecheck clean
- [x] Targets `dev`
- [x] No user-facing behavior change requiring docs-site updates (guidance text only)

## Summary by CodeRabbit

- **New Features**
  - Improved Cursor Code Mode detection for supported execution tools.
  - Added clearer guidance for nested helper usage, V8 isolation, output handling, and available top-level tools.
  - Preserved standard shell guidance when a direct shell bridge is available.

- **Documentation**
  - Added comprehensive Wave 0–2 closeout records covering triage, validation, release planning, known blockers, and proposed fixes.

- **Tests**
  - Expanded coverage for Cursor Code Mode detection, tool restrictions, shell bridges, and generated guidance.

Issue #604, opened 2026-07-28
[Bug] Cursor Auto repeats tool continuations on Windows PowerShell 5.1, causing excessive token usage


huaiqing-afk:
### Client or integration

Codex App

### Area

CLI

### Summary

When using OpenCodex with `cursor/auto` from the Codex App on Windows, simple coding tasks can enter a loop of repeated tool continuations and near-duplicate tool calls.

Tasks that should take a few tool calls can generate 20–45 calls and many repeated progress messages, resulting in unusually high token usage. The issue is especially noticeable after a shell command fails.

Expected behavior: after a tool result or command failure, the model should retain the tool-call context, adjust the command once, or report the failure clearly.

Actual behavior: the model often repeats planning messages and equivalent tool calls, such as “continue using Codex shell bridge”, without making meaningful progress.

<img width="1680" height="1320" alt="Image" src="https://github.com/user-attachments/assets/e664e8fe-1c21-4ace-bddb-b509fb32c276" />
<img width="1671" height="1053" alt="Image" src="https://github.com/user-attachments/assets/8e38eb2b-8d6c-469d-b0ea-afd2c2ffab4a" />
<img width="1521" height="1443" alt="Image" src="https://github.com/user-attachments/assets/1c8ff501-7e80-4d14-aa35-1cd925111a5d" />

### Reproduction

1. Use OpenCodex with the Cursor provider and model `cursor/auto`.
2. Run it from Codex App on Windows with Windows PowerShell 5.1 as the shell.
3. Ask for a small repository task, for example: inspect one file, make a small edit, and run a verification command.
4. If Cursor generates a Bash, CMD, or PowerShell 7-only command, let it fail.
5. Observe that OpenCodex/Cursor may repeatedly continue the task, re-plan, or repeat similar tool calls instead of recovering once or stopping with a clear error.

Examples of incompatible commands observed on Windows PowerShell 5.1:

```powershell
cd /d D:\path && git status

$(cat <<EOF
...
EOF
)
```powershell

cd /d is CMD syntax, && is not supported by Windows PowerShell 5.1, and the heredoc example is Bash syntax.

### Version

2.7.42

### Operating system

Windows 11 24H2, Windows PowerShell 5.1

### Provider and model

cursor / auto

### Logs or error output

```shell
Observed behavior:
- Multiple independent successful proxy requests are issued for one small task.
- Repeated assistant messages such as: “continue using Codex shell bridge”.
[...]

PR #1896, opened 2026-08-17
fix(cursor): route native fallbacks through code mode

luvyoun0224-beep:
## Summary

- Preserve Codex 0.147's real Responses Lite tool shape: the reserved `functions` namespace now flattens ordinary client functions and its nested freeform `custom exec` into top-level internal tools.
- Detect those code-mode turns in the live Cursor transport and retain that mode in the native-exec context.
- Replace misleading top-level `shell_command` / `exec_command` fallback instructions with the real code-mode contract: top-level `exec`, then nested `tools.exec_command(...)` or `tools.apply_patch(...)`, with `text(...)` output.
- Keep the existing flat-catalog guidance unchanged for non-code-mode requests and keep native local execution disabled by default.
- Add regression coverage for the captured Codex 0.147 `additional_tools` shape, filesystem, shell, fetch, MCP-resource, mutation, and live-transport propagation paths.

This follows #1844. That change fixed system guidance, but two gaps remained: Codex 0.147 groups `custom exec` under the reserved `functions` namespace, which the parser dropped, and Cursor-native rejection responses still directed models to top-level tools that do not exist in code mode.

## Verification

- focused Cursor/native/parser/Responses Lite suites: 162 pass
- all `tests/cursor-*.test.ts` files: 602 pass
- `bun run typecheck`
- `bun run privacy:scan`
- `git diff --check`
- bounded local live probe through Codex 0.147 + Cursor: the model successfully executed `Get-Content package.json` and read version `2.22.0`; the following model continuation hit the account's Cursor resource limit, so the probe did not receive its requested final formatted line.
- Full `bun test` was attempted on Windows; it stopped making progress during unrelated integration tests and was interrupted after more than 90 seconds without output.

## Checklist

- [x] Docs or release notes were updated when needed. (No user-facing configuration changed.)

AI assistance was used for implementation and test preparation; the resulting diff and verification output were reviewed locally.

## Review readiness checklist

### DIFF
--- a/src/adapters/cursor/live-transport.ts
+++ b/src/adapters/cursor/live-transport.ts
@@ -66,6 +66,7 @@ import { desktopDepsFromConfig } from "./native-exec-desktop";
 import {
   buildCursorToolDefinitions,
   cursorRequestAdvertisesApplyPatch,
+  cursorRequestUsesCodeMode,
   cursorRequestHasShellAlias,
   cursorToolArgNormalizeSchema,
   cursorToolWireName,
@@ -557,6 +558,7 @@ class LiveCursorTransport implements CursorTransport {
     this.execContext = {
       ...this.execContext,
       clientToolDefs,
+      codeMode: cursorRequestUsesCodeMode(request.tools, request.toolChoice),
       rejectNativeFileMutations: cursorRequestAdvertisesApplyPatch(request.tools, request.toolChoice),
       structuredEditAvailable: syntheticStructuredEditToolNames.size > 0,
     };
--- a/src/adapters/cursor/native-exec-fs.ts
+++ b/src/adapters/cursor/native-exec-fs.ts
@@ -34,12 +34,16 @@ import {
   type GrepArgs,
 } from "./gen/agent_pb";
 import { errorText, execBytes, lineCount, textDecoder, textEncoder } from "./native-exec-common";
+import { codeModeBridgeGuidance } from "./native-exec-guidance";
 
 const MAX_GREP_FILES = 500;
 const MAX_GREP_RESULTS = 200;
 const MAX_FILE_BYTES = 1_000_000;
 
-function codexNativeMutationRefusal(operation: "write" | "delete", structuredEditAvailable: boolean): string {
+function codexNativeMutationRefusal(operation: "write" | "delete", structuredEditAvailable: boolean, codeMode?: boolean): string {
+  if (codeMode === true) {
+    return `Cursor-native ${operation} is disabled for this Codex request because file mutations must stay on the Codex approval path. Call the top-level \`exec\` tool with JavaScript in its body, then call \`await tools.apply_patch(...)\` inside that body and emit the result with \`text(...)\`. Do not call \`apply_patch\` as a top-level tool in code mode. No file was changed.`;
+  }
   const structuredHint = structuredEditAvailable
     ? " Use the structured edit tools (`edit_file` / `multi_edit`) or the `apply_patch` tool for file edits so Codex can approve the change, enforce sandbox policy, show diffs, and record rollout."
     : " Use the `apply_patch` tool for file edits so Codex can approve the change, enforce sandbox policy, show diffs, and record rollout.";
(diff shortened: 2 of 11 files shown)


lidge-jun:
Reviewed as the candidate canonical PR for the #1887/#1896 consolidation. **Do not close #1887
as superseded yet** — one thing in it has to move here first, and it is not a nicety.

`codeModeBridgeGuidance` in this PR hardcodes the literal `` `exec` `` and the
`mcp_opencodex-responses_*` names whenever `codeMode === true`. That is a boolean, not a read
of what the request actually advertised. #1887 does the same job by asking the catalog:
`cursorNativeExecUsesCodeModeBridge(catalog)` checks `hasAdvertisedName` and returns
`{kind:"none"}` when `exec` is not advertised.

The reason this matters more than a style preference: #1895 exists specifically to remove
hardcoded tool names from the shared catalog nudge, on the grounds that hardcoding re-creates
per-provider drift. Landing this PR as canonical would re-introduce that defect one PR after
deleting it, in the same wave.

Four more things live only in #1887 and would be lost:

1. **`native-exec-bridge.ts`** — the rewrite engine that translates denied native args into
   `cat`/`ls`/`rg`/`curl` with shell quoting. This PR has no equivalent.
2. **`planNativeExecRewrite`'s finalize-vs-cancel ordering**, with the documented failure mode:
   an immediate `cancelCursorRun()` sets `expectedClose`, `scheduleClientToolFinalize` becomes a
   no-op, turn 1 never emits `done`, the conversation id is dropped, and the rewritten result
   never comes back. That is expensive knowledge to rediscover.
3. **Three `cursor-native-exec-policy.test.ts` cases** — rewrite, defer-while-sibling-open, and
   non-code-mode fallthrough.
4. **Windows PowerShell 5.1 guidance** (#604): no `cd /d`, no heredocs, `&&`/`||` are parser
   errors, one-retry cap. The guidance string here drops all of it.

Worth saying explicitly that these two PRs are **not duplicates in kind**. This one is
*guidance* — a denied op returns text asking the model to call `exec` itself, so it depends on
model compliance. #1887 is *mechanical* — it intercepts the denied frame and synthesizes a real
`exec` call, needing none. Guidance is the better default; the rewrite is the fallback for when
the model ignores it. Keeping both layers is the point of consolidating rather than choosing.

This PR is still draft, so nothing is blocked by this note. The parser fix here (flattening
Codex 0.147's reserved `functions` namespace so freeform `custom` children like code-mode `exec`
survive) is genuinely load-bearing and absent from #1887 — that, not scope, is the real case
for making this one canonical.

lidge-jun:
Reviewed as the intended canonical of the #1887/#1896 pair, and it is the right choice — the
`parser.ts` fix here (flattening Codex 0.147's reserved `functions` namespace so freeform
`custom` children like code-mode `exec` survive) is load-bearing and absent from #1887.

But **do not merge this and close #1887 as superseded**, because one guard would be lost and
it is the one that matters most:

`codeModeBridgeGuidance` hardcodes the literal `` `exec` `` and the `mcp_opencodex-responses_*`
names whenever `codeMode === true`. #1887 instead derives them from the advertised catalog via
`cursorNativeExecUsesCodeModeBridge`, returning `{kind:"none"}` when the catalog does not
advertise `exec`.

That is the same defect #1895 was opened to remove — naming tools the turn may not actually
advertise. I have just fixed the shared-nudge version of it in #1951, which decides code mode
from `freeform` metadata rather than the name, precisely because a structured `exec`, or an
`exec` beside a visible shell bridge, is not code mode at all. Landing this as-is would
re-introduce that shape in the Cursor path one PR after deleting it from the shared one.

To become canonical, this needs #1887's:

1. `cursorNativeExecUsesCodeModeBridge` catalog detection replacing the hardcoded names — the blocker
2. the `native-exec-bridge.ts` rewrite engine, which synthesizes a real `exec` call instead of asking the model to make one
3. `planNativeExecRewrite`'s finalize-vs-cancel ordering and its documented failure mode
4. the three `cursor-native-exec-policy.test.ts` cases
5. the Windows PowerShell 5.1 guidance from #604

Items 1 and 2 are the substantive ones: guidance depends on model compliance, the rewrite does
not, so they are complementary layers rather than two takes on one fix.

Merge mechanics: this branch is currently clean against `dev` after #1900 landed (`2b12521ee`).

lidge-jun:
Landed via #2020 — the load-bearing functions-namespace flatten is on `dev` (60/0 suites + tsc), ported by hand onto the current buildTools shape. The hardcoded-name guidance half was deliberately dropped; a catalog-helper approach replaces it separately (see #1887's close note). Thanks!

PR #1898, opened 2026-08-17
fix(pacing): anchor intervals to transport starts

Wibias:
## Summary

- make provider request pacing measure the interval at the actual outbound transport start instead of before an awaited continuation;
- replace the wall-clock-sensitive concurrency assertion with an injected-clock regression that reproduces delayed continuation starts deterministically.

## Verification

- RED phase: test-only commit intentionally reproduces the current scheduling gap in CI before the implementation change.
- Final focused/full checks will be recorded after the production fix lands.

## Checklist

- [x] Docs or release notes were updated when needed. No docs change is required because this restores the documented/configured pacing guarantee without changing configuration.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. This change does not touch credentials, authentication, or secret handling.

### DIFF
--- a/src/providers/request-pacing.ts
+++ b/src/providers/request-pacing.ts
@@ -35,7 +35,7 @@ interface Waiter {
   modelIntervalMs: number;
   queuedAt: number;
   signal?: AbortSignal;
-  resolve: () => void;
+  start: () => void;
   reject: (reason: unknown) => void;
   abort?: () => void;
 }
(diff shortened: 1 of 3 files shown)


lidge-jun:
Reviewed as part of the Wave 5B train and **held, not rejected** — the direction is right and
three of the five things this needed are already here.

Present and verified in the diff:

- intervals anchored to transport start rather than queue entry
- a cancelled waiter does not consume a slot (`aborted queued requests leave immediately and never consume a start`)
- deterministic timing via an injected clock rather than wall time

Two are missing, and they are the ones that decide whether the pacing clock is actually
correct under load:

1. **Retry does not double-advance the clock.** A retried request that re-enters the pacer
   must not count as a second transport start, or the effective interval silently doubles
   under exactly the conditions that trigger retries.
2. **Per-account isolation.** `account` appears **zero** times in this diff. If one account's
   pacing can delay another's, the fix trades a burst problem for a starvation problem.

One smaller thing: the description still says "Final focused/full checks will be recorded
after the production fix lands," but the production fix is already in the diff — 56 lines in
`request-pacing.ts` and 72 in `fetch-helpers.ts`, including the
`waitForProviderRequestSlot` → `runWithProviderRequestSlot` signature change. Worth refreshing
so a reviewer is not told to expect something that already happened.

Happy to see this land once those two tests exist. The PR is still draft, so nothing is
blocked by this note.

lidge-jun:
Closing as superseded: the transport-start anchoring this draft targeted is already on `dev` — request pacing now anchors intervals on `lastStartedAt`, which is the behavior this PR set out to introduce. The `enforce-target` failure on this head is moot as a result. If a pacing case remains where intervals still anchor on completion, open a fresh issue with the timing capture.

PR #1899, opened 2026-08-17
fix(tests): stop asserting POSIX modes Windows cannot produce

ntdatt812:
Part of the #1059 Windows burn-down.

## What is failing

Six tests across the Codex write substrate assert that a published file ends up at mode `0600`:

```js
expect(statSync(path).mode & 0o777).toBe(0o600);
```

On Windows that assertion can never pass. `chmodSync` there moves the read-only flag and nothing else, so `statSync` reports `0o666` no matter how the file was written:

```
Expected: 384
Received: 438
    at tests/codex-catalog-writer.test.ts:240:41
```

Two of the six fail on their own **setup** line — `codex-transition-state` and `native-main-claim` both `chmod` a file to `0o644` and assert that first — so they never reach the behaviour they exist to check.

## Why this is the test, not the product

The restriction is real; on Windows it is an ACL rather than a mode. Each call site touched here reaches it:

| test | write path | Windows restriction |
|---|---|---|
| `codex-catalog-writer` | `defaultBackupWriteIO` / `atomicWriteFile` | `hardenSecretPath` |
| `native-main-claim` | `openClaimDatabase` | `hardenSecretPathAsync` |
| `dsh-writer-lock` | `atomicWriteFile` | `hardenSecretPath` |

`src/integrations/journal.ts` states it outright: writes "go through `atomicWriteFile`, which applies 0600 plus Windows ACL hardening."

So the mode assertion is scoped to the platform whose semantics it is expressed in, and the ACL half stays where it can actually be observed — `tests/windows-secret-acl.test.ts`.

## The catalog-writer cases needed more than a scope guard

For those four, the mode was the *only* evidence that the writer restricted anything, so skipping it on Windows would have left them weaker there. They now assert the `harden:` effect as well — the half of the contract that holds on every platform:

```js
expect(effects.some(effect => effect.startsWith("harden:"))).toBe(true);
```

That is a stricter check than the mode was. Deleting `io.harden` from `publishCatalogBackup` and from `atomicWriteFile` turns all four red:

```
(fail) active catalog replacement writes prepared bytes atomically with the right live permit
(fail) hashed backup publication writes prepared bytes atomically with the right live permit
(fail) legacy backup publication writes prepared bytes atomically with the right live permit
(fail) models cache replacement writes prepared bytes atomically with the right live permit
 5 pass, 4 fail
```

The old mode assertion could not see that regression on Windows at all, since `0o666` is what it reports either way. This is the failure mode the `native-main-claim` header already warns about: "an audit deleted the hardening call from `openClaimDatabase` and 89 tests across three files stayed green."

`native-main-claim` is skipped rather than scoped because its own comment already draws the line — "On POSIX that is the mode; the Windows branch is proven separately" — and every line of it, setup included, is written in mode terms.

## One neighbouring failure deliberately left red

`codex-transition-state`'s *"a coordinator found group-readable is narrowed back to owner-only"* fails identically, and I did **not** touch it.

`src/codex/transition-state.ts` does not import `windows-secret-acl` at all. Neither does `catalog-write-serialization.ts` or `history-lock.ts`, and all three carry the same comment on a bare `chmodSync`:

```js
try { chmodSync(databasePath, 0o600); } catch { /* Windows applies ACLs in WP11. */ }
```

If that work is still outstanding, those coordinator databases are genuinely not narrowed on Windows, and the red test is honest signal rather than a platform artifact. Scoping it away would have hidden that. Happy to send the follow-up if you confirm which way it should go — the two shapes are "apply the ACL" and "record the gap".

## Verification

Windows 11, Bun 1.3.14, against `dev`:

```
bun run typecheck                                   # clean

bun scripts/test.ts tests/codex-catalog-writer.test.ts
  before: 5 pass, 4 fail
  after:  9 pass, 0 fail

bun scripts/test.ts <the ten files carrying mode assertions>
  before: 171 pass, 1 skip, 5 fail
  after:  172 pass, 2 skip, 3 fail
```

The three still red after this change are the `codex-transition-state` case described above, plus two unrelated failures in that file (`busy and unavailable databases return typed outcomes instead of throwing`, and a `row validator` case that times out at 5s) which are not mode-related and are out of scope here.

No production file is modified; the `io.harden` deletions above were reverted after measuring.

## Review readiness checklist

## Summary by CodeRabbit

* **Tests**
  * Improved cross-platform validation for file permissions.
  * POSIX-specific permission checks now run only on supported platforms.
  * File existence continues to be verified across all platforms, including Windows.
  * Strengthened validation of temporary-file handling and cleanup.

### DIFF
--- a/tests/codex-catalog-writer.test.ts
+++ b/tests/codex-catalog-writer.test.ts
@@ -239,22 +239,28 @@ for (const mutator of mutators) {
     expect(readFileSync(path, "utf8")).toBe("new bytes\n");
     // Restriction is asserted in two halves, because it is enforced in two ways.
     //
-    // Every mutator must ask its I/O to harden what it publishes; that half of the
-    // contract holds on every platform, so assert it unconditionally.
-    expect(effects.some(effect => effect.startsWith("harden:"))).toBe(true);
-    // The resulting mode only carries the claim where POSIX modes do. Windows has
-    // no mode bits to set: `chmodSync` there moves the read-only flag and nothing
-    // else, so `statSync` keeps reporting 0o666 however the file was written, and
-    // this assertion could never pass. Real restriction on Windows comes from the
-    // per-user NTFS ACL `defaultBackupWriteIO` applies through `hardenSecretPath`
-    // — machinery these doubles deliberately leave out, so there is nothing here
-    // for a mode check to observe either way.
+    // The platform-independent half: the mutator must harden the exact bytes it
+    // is about to install. Tying all three effects to one temp path is what makes
+    // that a claim about this file — "some temp was written, some path was
+    // hardened, something was published" would hold even if they were different
+    // files. Hardening lands on the temp, never the destination, because the
+    // publish step is what puts an already-restricted file in place.
+    const tempEffect = effects.find(effect => effect.startsWith("temp:"));
+    expect(tempEffect).toBeDefined();
+    const tempPath = tempEffect!.slice("temp:".length);
+    expect(effects).toContain(`harden:${tempPath}`);
+    expect(effects).toContain(`${isBackup ? "publish" : "rename"}:${tempPath}->${path}`);
+    // The other half is the resulting mode, which only carries the claim where
+    // POSIX modes do. Windows has no mode bits to set: `chmodSync` there moves the
+    // read-only flag and nothing else, so `statSync` keeps reporting 0o666 however
+    // the file was written, and this assertion could never pass. Real restriction
+    // on Windows comes from the per-user NTFS ACL `defaultBackupWriteIO` applies
+    // through `hardenSecretPath` — machinery these doubles deliberately leave out,
+    // so there is nothing here for a mode check to observe either way.
     if (process.platform !== "win32") {
       expect(statSync(path).mode & 0o777).toBe(0o600);
     }
     expect(readdirSync(targetDir).filter(name => name.endsWith(".tmp"))).toEqual([]);
-    expect(effects.some(effect => effect.startsWith("temp:"))).toBe(true);
-    expect(effects.some(effect => effect.startsWith(isBackup ? "publish:" : "rename:"))).toBe(true);
     if (isBackup) expect(result).toBe("written");
   });
 }
--- a/tests/codex-catalog-writer.test.ts
+++ b/tests/codex-catalog-writer.test.ts
@@ -237,7 +237,21 @@ for (const mutator of mutators) {
     );
 
     expect(readFileSync(path, "utf8")).toBe("new bytes\n");
-    expect(statSync(path).mode & 0o777).toBe(0o600);
+    // Restriction is asserted in two halves, because it is enforced in two ways.
+    //
+    // Every mutator must ask its I/O to harden what it publishes; that half of the
+    // contract holds on every platform, so assert it unconditionally.
+    expect(effects.some(effect => effect.startsWith("harden:"))).toBe(true);
+    // The resulting mode only carries the claim where POSIX modes do. Windows has
+    // no mode bits to set: `chmodSync` there moves the read-only flag and nothing
+    // else, so `statSync` keeps reporting 0o666 however the file was written, and
+    // this assertion could never pass. Real restriction on Windows comes from the
+    // per-user NTFS ACL `defaultBackupWriteIO` applies through `hardenSecretPath`
+    // — machinery these doubles deliberately leave out, so there is nothing here
+    // for a mode check to observe either way.
+    if (process.platform !== "win32") {
+      expect(statSync(path).mode & 0o777).toBe(0o600);
+    }
     expect(readdirSync(targetDir).filter(name => name.endsWith(".tmp"))).toEqual([]);
     expect(effects.some(effect => effect.startsWith("temp:"))).toBe(true);
     expect(effects.some(effect => effect.startsWith(isBackup ? "publish:" : "rename:"))).toBe(true);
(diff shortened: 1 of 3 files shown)


ntdatt812:
Thanks — the finding is right and it is now fixed in `381f05b79`, though not with the suggested diff.

`startsWith("harden:")` does only claim that *something* was hardened. Three unrelated effects would have satisfied it.

`toContain(`harden:${path}`)` cannot work, though: hardening never lands on the destination. Both writers harden the temp and publish it afterwards —

```
io.write(tmp, content) -> io.harden(tmp) -> io.rename(tmp, target)
```

— because the whole point of the sequence is to put an *already-restricted* file in place; hardening after publication would leave a window where the bytes are readable. Applying the suggestion turns every case red:

```
(fail) active catalog replacement writes prepared bytes atomically with the right live permit
(fail) hashed backup publication writes prepared bytes atomically with the right live permit
(fail) legacy backup publication writes prepared bytes atomically with the right live permit
(fail) models cache replacement writes prepared bytes atomically with the right live permit
 5 pass, 4 fail
```

What the test does instead is read the temp path back out of the recorded `temp:` effect and anchor the other two to it:

```js
const tempEffect = effects.find(effect => effect.startsWith("temp:"));
expect(tempEffect).toBeDefined();
const tempPath = tempEffect!.slice("temp:".length);
expect(effects).toContain(`harden:${tempPath}`);
expect(effects).toContain(`${isBackup ? "publish" : "rename"}:${tempPath}->${path}`);
```

That closes the gap you identified and also the one it left open in the other direction — the previous `startsWith(isBackup ? "publish:" : "rename:")` never checked *what* was published *where*. It now asserts one chain: these bytes were written to this temp, that temp was hardened, and that same temp became this destination.

Re-verified after the change: `bun run typecheck` clean, `bun scripts/test.ts tests/codex-catalog-writer.test.ts` 9 pass / 0 fail, and deleting `io.harden` from `publishCatalogBackup` and `atomicWriteFile` still turns all four red.

ntdatt812:
Three of the four boxes are ticked. Leaving the first one open deliberately, because I cannot honestly claim it.

**"All CI tests are green on my local testing"** — this machine is Windows, and the Windows suite is exactly what #1059 says it is: not green. I have not run the full suite here either; `bun run test` panics partway through on Bun 1.3.14 (`index out of bounds: index 0, len 0`), so a "0 fail" from it would be a truncated log rather than a result.

What I did run, on the rebased branch:

```
bun run typecheck                                    clean

bun scripts/test.ts \
  tests/codex-catalog-writer.test.ts \
  tests/native-main-claim.test.ts \
  tests/dsh-writer-lock.test.ts      25 pass, 1 skip, 0 fail

bun scripts/test.ts <ten files carrying mode assertions>
  before: 171 pass, 1 skip, 5 fail
  after:  172 pass, 2 skip, 3 fail
```

The three still red are named in the PR body: the `codex-transition-state` coordinator case I left alone on purpose, and two unrelated failures in that same file.

The other three boxes are true as of `8ab0aa8d0`: rebased onto `417ce9ea8` (0 behind), the one CodeRabbit finding addressed in `8ab0aa8d0` with the reasoning in the comment above, and the change is ready to read.

Happy to tick the first box too if the intended reading is "the tests I could run locally are green" rather than the whole suite — I would rather ask than claim something the log does not support.

ntdatt812:
Ticked the first box and marked this ready, reading it as "the tests I could run locally are green" — the reading I asked about above. Recording the scope plainly so the tick is not taken for more than it is:

- Green: `bun run typecheck`, the three files this PR touches (25 pass / 1 skip / 0 fail), and the ten files carrying mode assertions (172 pass / 2 skip / 3 fail, down from 5).
- Not run: the full suite. `bun run test` panics partway through on Bun 1.3.14 here, so it produces a truncated log rather than a result.
- Still red by design: the `codex-transition-state` coordinator case described in the PR body, plus two unrelated failures in that same file.

Say the word if you read that box more strictly and I will flip it back.

lidge-jun:
Superseded on `dev` by #1923 (merge `4d9738f43`). Closing, with what happened to each
of the three files.

**Two were overtaken by #1881.** `tests/dsh-writer-lock.test.ts` and
`tests/native-main-claim.test.ts` already carry the POSIX/Windows split on `dev` —
#1881 landed the same correction through `existsSync` plus a `posixModes` guard rather
than `test.skipIf`, which is arguably the better shape since the call still runs on
Windows and only the mode assertion is scoped. That overlap is why this branch reads
CONFLICTING.

**One carried something #1881 did not have**, and it is the reason this was worth
finishing rather than just closing as a duplicate: `tests/codex-catalog-writer.test.ts`
on `dev` still used unbound `some()` checks, which pass even when the temp write, the
harden, and the publish touch three different files. Your temp-path binding fixes that,
and #1923 takes it.

**One thing was added on top.** This PR asserts set membership —
`expect(effects).toContain(\`harden:\${tempPath}\`)` — so a writer that published first
and hardened afterwards still passes it. #1923 compares the recorded indices instead,
which matters more than it looks: for the two backup mutators it is the *only* detector
available, because `publishNoReplace` is `linkSync` and a temp hardened after
publication still shares the destination's inode. `chmod` succeeds, `statSync` reads
`0o600`, the leftover-`.tmp` check passes, every other assertion agrees nothing is
wrong — only the order disagrees.

Driven red both ways before landing (4 of 9 fail when the order is violated), and
independently reproduced by a second reviewer with two separate ablations.

Thanks for this — the binding idea is yours, and the ordering guarantee only became
visible because you asked the right question about what these effects actually prove.

PR #1901, opened 2026-08-17
test(memory): add concurrent recall acceptance probe

Wibias:
## Summary

- Add an offline, deterministic acceptance/profiling probe for the first remaining harness slice of #820. It drives the real OpenCodex HTTP, routing, adapter, bridge, relay, lifecycle, and app-owned-memory paths against a loopback-only synthetic provider.
- Exercise three barrier-synchronised 32-session × 10-recall waves, a 64-session burst, 1..8 parallel function/namespace/custom/tool-search calls, deterministic 10..500 ms external-tool latency, slow consumers, client cancellation, synthetic 429/503 responses, and a pre-first-byte stream failure.
- Isolate the proxy in a child process so proxy RSS/heap/external/ArrayBuffer samples exclude the load generator. Reuse the existing app-owned-memory, response-state, inspection, and active-turn counters rather than adding a second durable observability surface.
- Report process-memory and retained-state wave peaks plus idle slopes. RSS/heap slopes are profiling evidence only; PASS/FAIL is based on protocol completion and app-owned cleanup invariants, so the probe does not invent an allocator-sensitive RSS threshold.
- Keep the change maintainer-facing and behaviour-neutral: no production route, limit, config, persistence, request-history, or provider behaviour changes. #1217 remains the owner of durable per-request streaming-stage history.

Refs #820. This PR intentionally does not close the umbrella issue.

## Verification

- Source review against current `dev@417ce9ea8dca28dc166aa5c224b4db78bfcb5c51` and the focused #820 follow-ups confirmed that the remaining need is an acceptance/profiling harness, not another replacement memory-accounting stack.
- Duplicate check: no open PR matching the #820 concurrent recall soak/acceptance harness scope was found before opening this PR.
- Local Bun execution is not claimed: the available execution environment has Node/TypeScript tooling but no Bun binary, so `bun test tests/memory-recall-soak.test.ts`, `bun scripts/memory-recall-soak.ts --quick`, `bun run typecheck`, and `bun run privacy:scan` could not be run here.
- Exact-head repository CI and maintainer Windows execution are required before this draft is marked ready. The full 32/64-session probe is intentionally offline and is not added to the normal CI test suite because RSS/timing profiling must not become a flaky merge gate.

## Checklist

- [x] Docs or release notes were updated when needed. No user-facing docs or release note is needed for an offline maintainer probe with no runtime/config contract change.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. The probe uses loopback-only listeners, an isolated temporary OpenCodex home, a synthetic local credential, scalar-only metrics, and discards child runtime logs instead of retaining request bodies, paths, tokens, or identifiers.

## Summary by CodeRabbit

* **New Features**
  * Added an offline memory-recall soak and profiling tool covering sustained, burst, cancellation, streaming, and fault scenarios.
  * Added default and quick-run profiles with configurable options and deterministic workload selection.
  * Added process and application memory metrics, trend analysis, peak reporting, and lifecycle monitoring.
  * Added readiness and control endpoints, graceful shutdown, timeout handling, cleanup, and failure reporting.

* **Tests**
  * Added coverage for option parsing, deterministic behavior, validation, bounds, trend calculations, and maximum values.

### DIFF
--- a/scripts/memory-recall-soak-child.ts
+++ b/scripts/memory-recall-soak-child.ts
@@ -30,13 +30,22 @@ try {
   console.error("memory-recall-soak-child: --upstream must be a URL");
   process.exit(2);
 }
-if (upstreamUrl.hostname !== "127.0.0.1" && upstreamUrl.hostname !== "localhost" && upstreamUrl.hostname !== "::1") {
+const LOOPBACK_HOSTNAMES = new Set(["127.0.0.1", "localhost", "[::1]"]);
+if (upstreamUrl.protocol !== "http:" && upstreamUrl.protocol !== "https:") {
+  console.error("memory-recall-soak-child: --upstream must be http or https");
+  process.exit(2);
+}
+if (!LOOPBACK_HOSTNAMES.has(upstreamUrl.hostname)) {
   console.error("memory-recall-soak-child: upstream must be loopback");
   process.exit(2);
 }
 
 const home = mkdtempSync(join(tmpdir(), "ocx-memory-recall-soak-"));
 process.env.OPENCODEX_HOME = home;
+function cleanupHome(): void {
+  try { rmSync(home, { recursive: true, force: true }); } catch { /* temp cleanup only */ }
+}
+process.on("exit", cleanupHome);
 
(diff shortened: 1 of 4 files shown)


Wibias:
@coderabbitai review

Issue #1886, opened 2026-08-17
[Feature]: FastWire — generalize Fast mode across providers (umbrella)


olddonkey:
## Area

Multiple areas

## What are you trying to accomplish?

Use Codex Fast mode against any provider whose API offers an accelerated processing tier — with the catalog only advertising Fast where the route can actually honor it, the wire request using each provider's own dialect, and usage/cost reporting reflecting what each attempt actually did.

This is the umbrella tracking issue for **FastWire**, the design that generalizes Fast mode across providers. It sequences the refactor phases and links the per-provider delivery units (starting with #1875 for xAI). The design went through a five-round adversarial review against the current code and was accepted as the implementation basis.

## What prevents this today?

Fast mode is currently a single-dialect feature, with three hardwired assumptions:

- The only wire shape OpenCodex can emit is `service_tier: "priority"`, and only on the two OpenAI adapters (`openai-chat`, `openai-responses`). Anthropic's `speed: "fast"` + beta header, Groq's `service_tier: "performance"`, and Azure's deployment-scoped priority cannot be expressed at all.
- The catalog Fast blurb is the OpenAI one ("1.5x speed, increased usage") for every provider, even where the provider's actual terms differ (xAI Priority Processing bills at 2× the standard token price).
- `chatServiceTier` couples chat serialization into capability resolution (`serviceTierSupportForModel()` inlines `canSerializeServiceTierForChatModel()`), so one flag simultaneously influences catalog publication, routing-profile `require.serviceTier` eligibility, compatibility fingerprints, injection, and the gate.

Additionally, cost attribution reads top-level request fields rather than per-attempt outcomes, which mis-prices combo/fallback attempts, and there is no way to express "this provider's fast tier exists but is entitlement-gated" without either lying in the catalog or blocking the capability entirely.

## What should OpenCodex do?

Separate three concerns that are currently entangled (existing config field names unchanged):

1. **FastCapability** — may this provider/model/auth route offer Fast at all? Existing tri-state fields: `supportsServiceTier` (provider-level, `false` is fail-closed), `modelSupportsServiceTier` (exact-match), auth-mode overlay.
2. **CallerTierForward** — may caller-supplied foreign tiers (`flex`, unknown strings) be forwarded? This becomes `chatServiceTier`'s sole duty after migration.
[...]

PR #1904, opened 2026-08-17
fix(chat): forward caller service_tier through the chat-to-responses conversion (#1886)

olddonkey:
## Summary

Independent bug-fix unit of the FastWire umbrella (#1886): the chat→responses converter (`chatCompletionsToResponsesBody`) dropped `service_tier`, so a tier declared by a `/v1/chat/completions` caller vanished before the responses pipeline could see it — fast-mode override, the capability gate, and serialization all behaved as if no tier was requested, while the identical request through `/v1/responses` worked. The two inbound entries now agree.

**One production line**: copy `service_tier` under the converter's existing optional-scalar convention (`typeof === "string"`, matching the adjacent `user` / `prompt_cache_key` handling). No normalization, no injection — the downstream pipeline keeps owning tier semantics, so the fail-closed gate still strips it on `supportsServiceTier: false` routes.

**Stacked on** the A0 characterization PR lidge-jun/opencodex#1892, deliberately: A0 locked this bug as a `(known bug)` characterization cell, and the design's flip protocol says the bug-fix unit updates that cell in the same change. Review this PR by its top commit. (#1892 is now merged as `dec332c49`; no rebase is needed — this branch already contains its commit `0cdd07d51` through common history.)

## Tests

- Converter: caller-supplied tier preserved; no tier invented when absent.
- End-to-end through `handleChatCompletions` (forced chat→responses fallback): caller tier reaches the outbound body on a tier-capable route; stripped on an explicit `supportsServiceTier: false` route.
- A0 characterization cell flipped from drop-assertion to preserve-assertion, annotated with the unit reference.

## Verification

- Full suite at this commit: **12,750 pass / 10 skip / 0 fail** (12,760 tests across 823 files).
- `bun x tsc --noEmit` clean.

Part of #1886 (independent bug-fix unit; shrinks the later B1 migration surface).

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Service-tier options are now preserved when converting chat completion requests.
  * Service tiers are forwarded to compatible providers and omitted when unsupported.
  * Provider metadata and routing now accurately reflect service-tier availability.

* **Tests**
  * Added coverage for supported, unsupported, and unclassified service-tier scenarios.
  * Added validation for exact-model forwarding, compatibility behavior, logging, and request handling.


olddonkey:
Gate transparency for the merged head (`936aef8be`, dev merged in at `90fdc3a62`):

- Full suite: **12,783 pass / 10 skip / 1 fail** (12,794 tests, 825 files).
- The single failure — `CLI key-login live-update overlay preservation > notify after key login pushes the merged row and keeps modelCosts on live and disk` (`tests/key-login-live-update.test.ts:100`, live row comes back with `modelCosts: undefined`) — **reproduces identically on pristine `upstream/dev` at `90fdc3a62`** with none of this PR's changes present. It is a pre-existing dev-side regression, unrelated to this unit (which touches the chat→responses converter and service-tier tests only).
- At this PR's pre-merge base the suite was fully green: 12,750 pass / 0 fail.

Flagging so the baseline failure gets triaged on dev rather than attributed to this branch.

lidge-jun:
Its baseline is in: **#1892 merged** as `dec332c49`, so the A0 characterization is now on `dev`.

This PR is next in the Wave 5B order and is being held only by its own draft state — all four
readiness boxes are unticked, and `enforce-target` and `label` are CANCELLED at head, so there
is no passing gate to merge against. That is the checklist doing its job, not an objection to
the change.

Two things worth knowing before you tick the boxes:

- **No rebase is required.** This branch already contains #1892's commit `0cdd07d51`, so the
  shared `fastwire-characterization-*` files resolve through common history rather than as an
  add/add conflict. I verified a sequential merge of the 5B train onto `origin/dev` in a scratch
  worktree: five merges, zero conflicts. The body's "will rebase once #1892 merges" is now
  optional tidiness.
- The one file whose content differs from #1892 is the intentional flip —
  `characterization (known bug): ...drops service_tier` becoming
  `characterization: ...preserves service_tier` — which is exactly the A0 protocol working as
  designed. A reviewer seeing that diff should read it as the point, not a conflict.

Ready for review-readiness whenever you are.

olddonkey:
Thanks — and the draft state you saw is now cleared: head `4b74f6359` has all four readiness boxes ticked, `hygiene` / `enforce-target` / `label` / `resolve-pr` all green, and the gate comment flipped to ✅ READY at that SHA.

On your two points:

- **No rebase** — agreed, and the branch took the merge route rather than a rebase, so the shared history you verified is intact. I've dropped the "will rebase" line from the description as obsolete.
- **The flipped cell is the point** — exactly right. `characterization (known bug): …drops service_tier` → `characterization: …preserves service_tier` is the A0 protocol: A0 locks the bug as current behavior, and the bug-fix unit flips that same cell in the same change, so the two PRs read as one intentional sequence.

Local full-suite evidence at `4b74f6359` (which now contains dev tip `2a9f08324`): **12,788 pass / 10 skip / 1 fail**.

The single failure is the pre-existing dev-side one I flagged above — `tests/key-login-live-update.test.ts:100`, live row returns `modelCosts: undefined` — and it reproduces on pristine `upstream/dev` with none of this branch's changes present. Unrelated to this unit; worth its own triage on dev. Happy to open a separate issue for it if that's useful.

One caveat for the 5B ordering: four suites (`oauth-refresh-lock-multiprocess`, `port-reclaim`, `codex-shim`, `autostart-health`) are load-sensitive here — they time out under a fully parallel run and pass in isolation — so a red in those on a busy runner is likely noise rather than a real regression.

lidge-jun:
Validation before merge: scratch-worktree merge onto `dev` (post-#1965) — fastwire-characterization-wire + chat-completions-endpoint suites 112/0, `tsc --noEmit` clean. Review threads all resolved; hunk overlap with landed B1 confirmed disjoint by the plan-audit reviewer.

PR #1932, opened 2026-08-17
fix(codex): treat a bare WHAM 401 as transient while the main token is live

harryzhou2000:
WHAM is the ChatGPT usage/quota endpoint (GET https://chatgpt.com/backend-api/wham/usage) that ocx polls for plan, rate-limit windows, and reset credits. A bare 401 from WHAM does not prove the local credential died: upstream edges can transiently reject a still-valid access token. ocx treated every 401 as terminal, so the GUI's periodic refresh flipped a healthy main account to needs-reauth within a minute.

This PR only marks the main account needsReauth when the response carries a terminal code (invalid_workspace_selected / invalid_refresh_token) or the local access-token JWT has actually expired. Bare 401s with a live token are transient and leave the account usable.

Tests: updated and added auth-api tests for live-token 401 (transient), expired-token 401 (terminal), and the BUG-R327 flow.

## Review readiness checklist

### DIFF
--- a/src/codex/auth-api.ts
+++ b/src/codex/auth-api.ts
@@ -83,7 +83,7 @@ export {
   updateAccountQuota,
 } from "./quota";
 import { extractAccountId } from "../oauth/chatgpt";
-import { getMainAccountPlan, MAIN_CODEX_ACCOUNT_ID, setMainAccountPlan } from "./main-account";
+import { getMainAccountPlan, isMainAccountTokenLive, MAIN_CODEX_ACCOUNT_ID, setMainAccountPlan } from "./main-account";
 import { captureConfigGeneration, registerStateSweepAfterTick } from "../lib/state-store-sweeper";
 import { reconcileLiveStateStores } from "../lib/state-store-registrations";
 import {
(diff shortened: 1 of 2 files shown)


harryzhou2000:
@wibias — when you get a chance, could you share your opinion on this one? It touches the auth surface (src/codex/auth-api.ts), so per MAINTAINERS.md it needs a maintainer to apply maintainer-sponsored after review. Happy to adjust the approach if you prefer something different.

harryzhou2000:
@lidge-jun — could you share your opinion on this one when you have a moment? It changes the auth surface (src/codex/auth-api.ts): a bare WHAM 401 with a still-live access token is treated as transient instead of flipping the main account to needs-reauth. It is rebased onto the latest dev (fa77e5cb4) and tests are green locally; per MAINTAINERS.md it needs maintainer-sponsored after review. Happy to adjust if you prefer a different approach.

lidge-jun:
Landed via #2021 — your transient-401 gate with the reviewer-required tightening: liveness now requires a decodable future exp, so a dead credential cannot loop as transient. 186/0 + tsc. Thanks!

Issue #1527, opened 2026-08-12
[Bug]: Cursor adapter large-context turns collapse or rate-limit while direct Cursor remains healthy


rennzhang:
### Client or integration

Codex App / Cursor provider adapter

### Area

Provider adapter / streaming / large-context continuation

## Current status

Later reporter evidence supersedes the original framing of this issue.

The problem is **not simply that `kimi-k3` gives short answers or that `claude-fable-5` receives ordinary upstream 429s**. Direct controls through Cursor's own client show the same account and models completing large-context workloads normally, while the OpenCodex Cursor adapter path degrades sharply at comparable context sizes.

Composer routes remain out of scope unless separately reproduced.

This issue now tracks the OpenCodex-specific large-context Cursor path.

## Observed behavior through OpenCodex

At large context sizes, routed Cursor turns can show one or both of these symptoms:

- `cursor/kimi-k3` completes with only a few output tokens despite a large multi-step task;
- `cursor/claude-fable-5` and/or other Cursor routes hit repeated `429 rate_limit_exceeded` / resource-limit failures at context sizes that remain usable through the official Cursor client.

The adapter also consistently logs completed streams as an abort-style teardown:

```text
[ocx:cursor:turn-failed] ... "Cursor request was aborted" ... "committed":true
[ocx:cursor:stream-end] ... "expectedClose":false
```

That classification occurs even on turns whose committed output is otherwise usable.

## Decisive control

The reporter repeated comparable large-context workloads through Cursor's own client with OpenCodex removed from the path.

Those controls produced long multi-step outputs, successful tool loops, and no equivalent rate-limit storm on the same account at similar or larger session sizes.

This rules out the earlier conclusion that the short output is simply normal model behavior.

## Strongest current implementation hypothesis

The OpenCodex Cursor adapter appears to replay substantially more conversation state per turn than the official Cursor continuation path.

Observed differences include:

- official Cursor usage reports substantial `cache_read_tokens` during long sessions;
- OpenCodex usage for these routed Cursor turns is estimated and does not expose equivalent cache-reuse evidence;
- the adapter serializes the conversation/history into `rootPromptMessagesJson` for each turn;
- large histories approach the adapter's bounded serialized prompt envelope;
- repeated full replay may increase upstream processing pressure and can plausibly explain the large-context degradation and rate-limit behavior.

This is the strongest current direction, but it should remain a hypothesis until a side-by-side request/continuation trace proves the exact upstream contract difference.

[...]

PR #1940, opened 2026-08-17
fix(cursor): reuse conversation checkpoints for incremental continuation

keepitmello:
## Summary

- Preserve Cursor's returned ConversationStateStructure after a successful no-tool turn and reuse that snapshot on the next validated linear continuation instead of rebuilding the full root history.
- Tool-result turns reuse the last completed-turn checkpoint plus only the uncovered suffix when the covered message boundary is known.
- Keep the existing full-replay path for compaction, helper/shadow isolation, account or model mismatch, missing/expired refs, decode failures, restart, and invalid_argument recovery.
- Bind checkpoint snapshots to conversation id, credential identity scope, and model affinity. Persist only an opaque process-local checkpointRef on Responses continuation state, and pin referenced blobs for the checkpoint lifetime.
- Do not treat OpenCodex usage or missing cache counters as a cache-hit rate. Cursor Connect still reports only usedTokens/maxTokens.

Refs #1527

## Verification

What this change is supposed to prove is that a validated no-tool follow-up stops rebuilding old root history. Unit-test green is only the repo gate, not that proof.

Observed on this branch against the same Cursor account, after the local 2.24.2 process was running from this tree:

- Offline 3-turn full-replay baseline grew `rootBytes` 58 → 40,175 → 80,292.
- Live no-tool follow-up on this tip used `continuationMode=checkpoint`, same Cursor conversation id, `rootBytes=0`, and still answered from the prior turn (`BETA-9`).
- Official `cursor-agent` control on the same account / `cursor-grok-4.6-low`: turn 1 `cacheReadTokens: 0` / input 18937; resume turn 2 `cacheReadTokens: 18816` / new input 331 / answer `ALPHA-7`. OpenCodex still cannot report that counter authoritatively, so this is comparison data only.

Not shown, and not claimed:

- large-context / 429 / resource-limit reduction
- kimi-k3 or other #1527 premature-completion acceptance
- tool-heavy live MCP bridge parity
- cache hit rate

Repo gate only, so a reviewer can reproduce the mechanical checks:

- `bun run typecheck`
- `bun test tests/cursor-request-builder.test.ts tests/cursor-blob.test.ts tests/cursor-adapter.test.ts tests/cursor-discovery.test.ts tests/cursor-protobuf-events.test.ts tests/cursor-live-transport.test.ts tests/cursor-tool-continuation.test.ts tests/responses-state.test.ts`
- `bun run test`
- `bun run privacy:scan`

## Checklist

## Review readiness checklist

### DIFF
--- a/devlog/_plan/260814_bug_resolution_campaign/030_wave3_cursor.md
+++ b/devlog/_plan/260814_bug_resolution_campaign/030_wave3_cursor.md
@@ -48,6 +48,21 @@ Cursor tool/continuation/edit 경로의 correctness fix를
 - teardown 문제 (정상 완료를 aborted/expectedClose:false로 기록)는
   별도 작은 PR로 먼저 고친다
 
+2026-08-18 로컬 조사/prototype 메모 (fix/cursor-checkpoint-continuation, 아직 upstream PR 아님):
+
+- 병목의 1차 원인은 JSON 포맷 자체가 아니라, 매 턴 rootPromptMessages/conversationTurns로
+  과거 대화를 다시 만드는 full replay semantics다.
+- ConversationStateStructure checkpoint를 다음 conversationState로 재사용하면 no-tool
+  follow-up에서 로컬 rootBytes가 history와 같이 커지지 않는다. grok-4.6 live 3턴에서
+  2·3턴이 continuationMode=checkpoint였고 ALPHA-7을 기억했다.
+- 공식 cursor-agent 같은 계정 대조: 1턴 cacheReadTokens 0 / input 18937, 같은 세션 2턴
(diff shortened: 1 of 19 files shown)


lidge-jun:
Closing with a redesign directive rather than merging. The checkpoint-continuation idea is right and #1527 remains open, but this head is a 1064-line new process-local store that predates today's session-id pinning (#1990, just landed via #2017) and explicitly does not claim the 429/kimi-k3 symptoms. Redesign guidance for a fresh PR on current `dev`: (1) build ON TOP of the pinned session id — with a stable per-conversation session, Cursor's server-side thread reuse does most of what the local checkpoint store re-implemented; (2) keep the store process-local ONLY if a measured gap remains after #1990 (attach the measurement); (3) split acceptance into the original #1527 workload (large-context/429/kimi-k3) so the PR proves the symptom it claims. Happy to review a re-cut.

PR #1944, opened 2026-08-18
fix(windows): stop the service wrapper killer using the argv Bun rejects

lidge-jun:
## Summary

- `src/service.ts` spawned PowerShell with `"-WindowStyle", "Hidden"` as argv elements. `src/codex/user-identity.ts:222-224` already forbids exactly that: Bun 1.3.14 can fail the direct CLI pair before the command runs (#1589), and the process-level `windowsHide` flag is what actually suppresses the window.
- The call is not decorative. `stopServiceIfInstalled()` uses it because `schtasks /end` can leave the `wscript.exe`/`cmd.exe` wrapper alive to respawn the proxy, and it ignores `spawnSync`'s exit status. Under #1589 wrapper termination silently does nothing, so `ocx stop`, restart and update report success without sticking.
- The invariant survived as prose plus a single-file assertion. `tests/windows-deploy-close-regressions.test.ts:43` pins the argv only for `src/update/job.ts` (the variable is bound at line 13), so `service.ts` was never covered. Replaced with a sweep over every `src/**/*.ts`, which found exactly one offender.
- The sweep matches the **argv form only**. Six call sites legitimately pass `-WindowStyle Hidden` inside a PowerShell script string handed to `Start-Process` (`windows-elevation.ts` 622/660/687/736, `tray/windows.ts:489`, `update/job.ts:574`); Bun never parses those. A second test pins that discrimination in both directions so the sweep cannot quietly become vacuous or start failing correct code.

First of a stacked chain implementing `devlog/_plan/260817_windows_stability_program/` (phase `010`). This one targets `dev`.

## Verification

- Driven red first: the sweep reported `["service.ts"]` before the fix.
- `bun test tests/windows-popup-fix.test.ts` — 7 pass
- `bun test tests/service.test.ts tests/windows-deploy-close-regressions.test.ts` — 131 pass
- `bun run typecheck` — clean

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows service cleanup to prevent unwanted PowerShell compatibility issues.
  * Console windows remain hidden during background cleanup operations.

* **Tests**
  * Added regression coverage to ensure incompatible PowerShell arguments are not used directly.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -2358,7 +2358,10 @@ function killWindowsServiceWrapperProcesses(): void {
       "} | ForEach-Object { Stop-Process -Id $_.ProcessId -Force -ErrorAction SilentlyContinue }",
     ].join(" ");
     spawnSync(resolveTrustedWindowsPowerShellExe(), [
-      "-NoProfile", "-NoLogo", "-NonInteractive", "-WindowStyle", "Hidden",
+      // No `-WindowStyle Hidden` here: Bun 1.3.14 can fail that direct CLI pair
+      // before the command runs (#1589). `windowsHide` below is what actually
+      // suppresses the console window, and it is sufficient.
+      "-NoProfile", "-NoLogo", "-NonInteractive",
       "-Command", ps,
     ], { stdio: "ignore", timeout: 5000, windowsHide: true });
   } catch { /* best-effort */ }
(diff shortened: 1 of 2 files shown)


lidge-jun:
Validation before merge: scratch-worktree merge onto current `dev` — `windows-popup-fix` suite 7/0, `tsc --noEmit` clean. Independent audit: diff is win32-gated argv-only (`-WindowStyle Hidden` removal + src-wide argv-form sweep test), no unix path touched, exact-head CI green.

PR #1945, opened 2026-08-18
fix(windows): one scheduler-wrapper killer, scoped to one installation

lidge-jun:
## Summary

- `src/service.ts` and `src/update/job.ts` each carried a copy of the same teardown logic, and the copies drifted in **both** directions. `service.ts` matched canonical full paths as complete command-line tokens; `update/job.ts` matched the bare filenames with `-like '*name*'`. Meanwhile `update/job.ts` had received the #1589 argv cleanup that `service.ts` had not (fixed in the parent PR).
- The bare-filename matcher is the defect. Two OpenCodex homes under one Windows account means a dashboard update for home A can force-terminate home B's scheduler wrapper, and any unrelated process whose command line contains either filename matches as well.
- Extracted the `service.ts` implementation, which was the correct one, into `src/lib/windows-service-wrappers.ts` and pointed both callers at it. The updater now passes its own config dir instead of bare names.
- `windowsWrapperKillScript` is exported because the matching rule is the entire point of the module and the spawn reports nothing: the script it builds is the only observable surface, which is the same source-level convention `windows-deploy-close-regressions.test.ts` already uses.
- Two existing assertions pinned strings that moved (`windows-deploy-close-regressions.test.ts`, `cli-ready.test.ts`). They follow the invariant to the shared module rather than being dropped.

Second of a stacked chain implementing `devlog/_plan/260817_windows_stability_program/` (phase `020`). Targets #1944; retarget to `dev` once that lands.

## Verification

- Driven red first: the "neither file keeps a private matcher" assertion failed for both files before the extraction.
- `bun test tests/windows-service-wrappers.test.ts` — 5 pass (new file)
- `bun test tests/service.test.ts tests/windows-deploy-close-regressions.test.ts tests/windows-popup-fix.test.ts tests/windows-service-wrappers.test.ts` — 143 pass
- `bun test tests/cli-ready.test.ts` — 57 pass
- `bun run typecheck` — clean

## Checklist

### DIFF
--- a/src/lib/windows-service-wrappers.ts
+++ b/src/lib/windows-service-wrappers.ts
@@ -1,0 +1,72 @@
+/**
+ * Shared termination of surviving Windows scheduler launcher/wrapper processes.
+ *
+ * `schtasks /end` ends the task instance but often leaves wscript/cmd running the
+ * `:loop` batch, which brings the proxy back during a stop, a restart, or
+ * post-update reclaim. Both the service teardown path and the update job need
+ * that guarantee, and they used to carry a copy each.
+ *
+ * The copies drifted in both directions, which is why this module exists rather
+ * than a second careful implementation: the updater received the #1589 argv
+ * cleanup that service.ts missed, and service.ts received the canonical-path
+ * scoping the updater missed. One of those gaps could force-terminate a wrapper
+ * belonging to a DIFFERENT OpenCodex home.
+ *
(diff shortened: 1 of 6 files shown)


lidge-jun:
Validation before merge: scratch-worktree merge onto `dev` (post-#1944) — windows suites 7/0, `tsc --noEmit` clean. Plan-audit reviewer confirmed merge-tree onto current dev is clean and no later-dev commit touches stack files.

PR #1946, opened 2026-08-18
refactor(windows): share the atomic-replace retry instead of one writer owning it

lidge-jun:
## Summary

- `src/config.ts:102-123` knew that Windows can refuse `rename` with `EBUSY`, `EPERM` or `EACCES` while a scanner or sync client still holds the target, and retried twice (25ms then 50ms). Nothing else did. Two durable publishers called `renameSync` directly: `src/codex/prompt-journal.ts`, which publishes a journal carrying full `config.toml` bytes — losing that publish is what breaks journal restore — and `src/lib/config-ownership.ts`, which publishes the uninstall manifest.
- Neither corrupts anything on failure; they throw rather than publish a partial file. But under a real-time scanner holding the target they turn a momentary hold into a user-visible failure, and the tolerance to survive it already existed one module away.
- The loop moves to `src/lib/windows-atomic-replace.ts` rather than becoming an export of `config.ts`: `config-ownership.ts` is one of the callers and `config.ts` already imports `config-ownership.ts` (line 47), so the obvious placement would close an import cycle. `config.ts` re-exports the names because they are part of its public surface.
- **The envelope is unchanged** at two retries. Widening it without evidence trades a rare failure for a routine stall; the counters in the child PR are how that argument gets made from data instead.

Third of a stacked chain implementing `devlog/_plan/260817_windows_stability_program/` (phase `030`). Targets #1945; retarget to `dev` once the parents land.

## Verification

- `bun run typecheck` — clean
- Full suite in 60-file batches (Bun 1.3.14 panics near 3.5GB on a monolithic `--isolate` run over 809 files): batches 1-2 green at 647 and 1032 pass; remaining batches running. One pre-existing failure in `tests/codex-app-server-processes.test.ts` ("a defaulted read is memoized") reproduces on clean `origin/dev` and is unrelated to this change.

## Checklist

### DIFF
--- a/src/claude/agents-inject.ts
+++ b/src/claude/agents-inject.ts
@@ -11,9 +11,10 @@
  * Ownership contract: this module only creates/overwrites/deletes files matching
  * `ocx-*.md` inside the agents dir. User-authored agents are never touched.
  */
-import { lstatSync, mkdirSync, readdirSync, readFileSync, renameSync, unlinkSync, writeFileSync } from "node:fs";
+import { lstatSync, mkdirSync, readdirSync, readFileSync, unlinkSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import type { OcxConfig } from "../types";
+import { renameAtomicFile } from "../lib/windows-atomic-replace";
 import { claudeCodeAlias, claudeCodeNativeAlias } from "./alias";
 import { AUTO_CONTEXT_OFF, shouldMarkOneMillion, stripOneMillionMarker, withOneMillionMarker } from "./context-windows";
 import { claudeConfigDir } from "./gateway-cache";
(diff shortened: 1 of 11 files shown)


lidge-jun:
The `keyring ubuntu` job on this PR has stalled twice on its **Install Secret Service dependencies (Linux)** step (apt), once for 25+ minutes before I cancelled it, and again on the rerun. The same job passes on the sibling PRs in this stack (#1944, #1945, #1947) against nearly identical trees, and `keyring macos` / `keyring windows` pass here.

Nothing in this PR touches keyring code, CI configuration, or dependencies — it moves a rename retry loop into `src/lib/windows-atomic-replace.ts` and converts eight callers. Treating the stall as runner-side rather than a real failure; happy to rerun once more if a maintainer prefers.

lidge-jun:
Validation before merge: scratch-worktree merge onto `dev` (post-#1945) — windows-popup-fix + config suites 158/0, `tsc --noEmit` clean.

PR #1947, opened 2026-08-18
feat(windows): count atomic-replace retries so the envelope can be argued from evidence

lidge-jun:
## Summary

- Both plan audits wanted the 75ms retry envelope widened. Neither could show it failing in the field, and one explicitly declined to raise its severity for exactly that reason. Counting is the honest next step: if these stay at zero across a release the envelope is fine, and if they do not, the change cites numbers.
- Counters live beside the retry loop, keyed by which publisher retried, with separate `retried` and `exhausted` totals. Exposed as `GET /api/system/windows-replace-retries` — a **sibling** of `/api/system/memory` rather than a field on it, because that payload is memory-shaped and appending filesystem counters would make both harder to consume.
- The publisher label is a closed union (`ReplacePublisher`), and **that union is the privacy enforcement, not `privacy:scan`**. The scanner reads file text (`scripts/privacy-scan.ts:187`) and cannot tell that a runtime string came from a path — a path could carry a username. A closed union makes the same mistake a typecheck failure instead, which the new test pins with `@ts-expect-error`.
- Scope note, stated because the plan originally overreached here: these counters are **process-local**. Asserting they stay zero across the Windows suite would need a finalizer aggregating many short-lived sharded processes, which does not exist. Evidence comes from local runs and voluntary bug reports.
- `tests/system-routes.test.ts` is new; `handleSystemRoutes` coverage previously lived scattered in `memory-watchdog.test.ts` and `codex-restart-route.test.ts`.

Last of a stacked chain implementing `devlog/_plan/260817_windows_stability_program/` (phase `031`). Targets #1946; retarget to `dev` once the parents land.

## Verification

- `bun test tests/system-routes.test.ts` — 9 pass
- `bun run typecheck` — clean

## Checklist

### DIFF
--- a/src/claude/agents-inject.ts
+++ b/src/claude/agents-inject.ts
@@ -236,7 +236,7 @@ export function syncClaudeAgentDefs(defs: readonly ClaudeAgentDef[], configDir =
       } catch { /* does not exist: ours to create */ }
       const tmp = `${target}.tmp-${process.pid}`;
       writeFileSync(tmp, renderAgentDef(def), { encoding: "utf8", mode: 0o644 });
-      renameAtomicFile(tmp, target);
+      renameAtomicFile(tmp, target, undefined, "claude-agents");
       written.push(def.file);
     }
     return written;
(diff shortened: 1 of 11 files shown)


lidge-jun:
Validation before merge: scratch-worktree merge onto `dev` (post-#1946) — windows-popup-fix + config suites 158/0, `tsc --noEmit` clean. Earlier UNSTABLE state was cancelled duplicate label/react-doctor jobs; the real ci check is green at head.

PR #1956, opened 2026-08-18
feat(fastwire): B0 — per-attempt tier outcomes and per-attempt pricing (#1886)

olddonkey:
## Summary

Phase **B0** of the FastWire umbrella (#1886): per-attempt tier observability. **Upstream wire bytes are unchanged** — this unit changes what gets recorded and how cost reads it, not what gets sent.

**Stacked on** the A1 refactor PR lidge-jun/opencodex#1893; its commits appear here too, so review the B0-specific commits and the current `dev` merge.

## The bug this fixes

`estimateComboCost` took one top-level service tier and applied it to **every** attempt. Combo, fallback, and retry rows priced attempts that never carried that tier — and a request with Fast enabled whose route could not express Fast still billed at the Fast multiplier, because the top-level `configuredServiceTier` survived the strip.

Each attempt now carries an `AttemptTierOutcome` produced by **the adapter that actually serialized the request** (never inferred from the route): canonical tier, the emitted wire kind/value, `fastOutcome`, `confirmation`, and the upstream echo. Cost reads that per attempt and falls back to the old top-level tier for pre-B0 rows.

## Correctness details worth reviewing

- **`fastOutcome` applies the tier-decision precedence**, so it cannot misreport a downgrade: `force-default` is always `not-requested` (a user choosing default is not a downgrade — recorded separately as `callerFastSuppressedByConfig`); an unclassified route's passthrough stays `unknown` without inferring demand or suppression; a dropped foreign caller tier only sets `callerTierDropped` and never touches `fastOutcome`.
- **Confirmation reverse-maps the upstream echo** through `canonicalToWire` (generic, so Groq's `performance` works when it lands), giving `confirmed` / `assumed` / `downgraded` / `unknown`. An upstream that declines Fast prices at the tier it actually served — which matches how vendors document this: xAI bills the priority rate only when the response confirms `"priority"`.
- **Persistence is additive and fails closed**: `tierOutcome` is optional, pre-B0 rows price exactly as before, and a malformed outcome is dropped without losing its attempt.
- **The tier gate is value-aware** now, but the `drop` branch has no provider today, so the outbound body is byte-identical — the A0 characterization suites pass untouched.

Also adds the bounded, secret-redacted `callerServiceTier` raw-evidence field (the existing `requestedServiceTier` is mixed-semantics and A0 locked it, so it cannot serve as raw evidence), and projects `responses.fastWireKind` / `responses.fastWireValue` into the compatibility fingerprint. The fingerprint gains two keys, which intentionally changes subject digests.

## Verification

At head `4d87bce04` with `dev` tip `4b24c9ef0` merged:

- `bun run test`: **13,132 pass / 10 skip / 0 fail** across 835 files.
- Focused FastWire/config/usage suite: **494 pass / 0 fail** across 5 files.
- `bun run typecheck`: pass.
- `bun run privacy:scan`: pass.
- `git diff --check`: clean.
- Follow-up coverage exercises JSON/SSE unparseable-response wiring, unknown-tier pricing fallback, token-shaped provider names, C1/Unicode line separators, and bounded warnings.

Part of #1886 (Phase B0). This is the prerequisite for B2's per-provider pricing work: filling a price matrix on top of attempt-level tier facts is only safe once those facts exist.

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added FastWire configuration for provider-specific mappings, caller tiers, and optional beta settings.
  * Improved fast-mode and service-tier handling, including tier forwarding, downgrades, compatibility checks, and response observation.
  * Added tier outcomes and caller-tier details to request logs and usage records.
  * Added per-attempt tier-aware cost calculations.

* **Bug Fixes**
  * Preserved configured wire settings while preventing unsupported or conflicting combinations.
  * Added validation for FastWire configuration and safe metadata redaction.
  * Added bounded warnings for unsupported provider and model combinations.


olddonkey:
All eight findings are addressed in `330350d7c` (plus the dev merges since). Per-finding disposition, including the two I resolved differently than proposed:

**Fixed as reported**

- **`captureFastPolicyAuthority` cached mutable providers** — real, and the comment above it already promised the opposite. The WeakMap now keys on frozen providers only, on both the write and read sides. One consequence worth naming: the catalog path relies on that cache to keep its "capture before the flight, never re-read the registry mid-flight" guarantee, so `recursivelyFreeze` moved *ahead* of the capture call there. `tests/codex-gather-authority.test.ts` still passes, which is what enforces that guarantee.
- **Behavior resolver version** — bumped to `2`. This is the better fix than what the unit originally did (treat the digest change as merely intentional): without the bump, new fingerprints are silently incomparable to recorded ones. No test or doc pinned the old value; the separate protocol resolver version is untouched.
- **Prototype-bearing lookups** — hard pins, `modelAdapters`, and `registryWireDefaults` now use own-property guards in the repo's existing `Object.hasOwn` style, and a hard pin additionally requires a string value. Provider names and model ids are operator-controlled, so this was reachable.
- **Three copies of the FastWire clone** — collapsed into one exported helper; the authority path takes a frozen snapshot while registry inheritance keeps a mutable detached clone.
- **Unreachable `force-default` condition** — removed; `fastIntent` already implies it. The defensive `delete` in the unclassified branch stays, since the function accepts externally-constructed contexts.
- **Hard-pin + `null` declaration test** — added, asserting `wire-unavailable` rather than `pin-unavailable`.

**Resolved with a narrower scope than proposed**

- **Write path vs load path** — this was the most valuable finding: an operator whose config predates a registry gaining capability could load fine but never save again. Rather than making the load path strict, `validateConfigCandidate` now stops rejecting *inherited* conflicts, matching `loadConfig`'s warn-and-preserve contract (and the reason recorded in its comment). Direct contradictions *within one provider row* — the thing the author actually wrote — remain schema errors via `providerConfigSchema`, so the original intent is kept where it belongs.
- **Duplicate Chat eligibility rule** — converged minimally (delegation, not a rewrite) on purpose: the stacked B1 unit (#1965) retires `legacyChatEligibility` entirely and replaces this path with the capability/CallerTierForward split, so a structural rewrite here would have been thrown away one PR later. In the merged stack the delegation is gone and the duplication is resolved by construction.

Local full-suite at the fixed head: **13,059 pass / 10 skip / 1 fail** — the one failure is the pre-existing dev-side `tests/key-login-live-update.test.ts` regression that reproduces on pristine `upstream/dev`.

lidge-jun:
Closing as landed-via-#1965. B1's merged head contains this PR's head commit (4d87bce04) and its file set was a proven superset — every B0 per-attempt observability commit is now on `dev` through the #1965 merge. Closing rather than merging to avoid a second landing of the same commits. Umbrella tracking continues on #1886.

PR #1988, opened 2026-08-18
fix(catalog): default native GPT-5.6 max_context_window to the upstream max

harryzhou2000:
Default native GPT-5.6 rows advertise max_context_window 272000 even though the live upstream catalog (and the Codex models_cache) advertise 872000. narrowNativeMaxContextWindow flattened the override's max to the operating window whenever no user lever was set.

This PR:

- adds NATIVE_GPT56_MAX_CONTEXT_WINDOW = 872000 (the upstream advertised native max);
- keeps the 272k operating context_window default and the 922k measured opt-in ceiling;
- stops flattening max_context_window to context_window when the window is not narrowed;
- lets max_context_window follow a raise between the native max and the measured ceiling, clamped at 922k.

Tests now use the exported constants instead of magic numbers, and add margin cases at the 872k native max and the 922k ceiling.

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - Added GPT-5.6’s native maximum context-window limit to the public model catalog.
  - GPT-5.6 models now distinguish their supported operating context window from their maximum input capacity.
  - Context-window settings preserve native limits while respecting configured caps and increases.

- **Bug Fixes**
  - Improved context-window and automatic-compaction limit calculations for GPT-5.6 models.
  - Updated model discovery and catalog synchronization to report consistent limits across supported GPT-5.6 variants.


harryzhou2000:
@coderabbitai — addressed in 157792375. narrowNativeMaxContextWindow now applies limits.cap independently to max_context_window, so a 350,000 OpenAI cap clamps both context_window and max_context_window even when a per-model window raise would otherwise keep the native max. Regression test added in tests/native-model-toggle.test.ts asserting both equal 350,000.

lidge-jun:
272k is intended u can change at model

harryzhou2000:
Hi, just want to make sure that both `context_window` and `max_context_window` are intended to be 272k. The matter is that when `max_context_window` is also 272k in catalog, the user specified context window in codex's config.toml larger than 272k would not be respected, which creates behavior deviation for openai models with/without ocx injection. @lidge-jun

Issue #1866, opened 2026-08-16
[Compatibility] Cursor Computer Use / node_repl tool results come back empty or truncated


lidge-jun:
## Client or integration

Codex App

## Provider or upstream service

cursor

## OpenCodex version

2.18.0 (local proxy on localhost:10100; worktree HEAD e2ef24ad6fda110063134d368dc9e3c7763ec681)

## Endpoint or capability

Cursor adapter tool-result replay for Codex Desktop Computer Use / node_repl. Observed on `/v1/responses` turns routed to `cursor/grok-4.6`.

## Current behaviour

A Codex App session routed through the OpenCodex Cursor adapter can call every Computer Use API successfully, then lose the result on the next turn.

Live probe on 2026-08-17 (macOS, Chrome already open):

1. `list_apps`, `get_app_state`, screenshot emit, `click`, `set_value`, `type_text`, `select_text`, `press_key`, `scroll`, `drag`, and `perform_secondary_action` all succeeded when invoked through `mcp__node_repl__js`.
2. The next outer `exec` wrapper often returned **empty output** even though the nested Computer Use call completed.
3. Full AX trees and JPEG screenshots truncated in the model-visible tool result (`original token count: 43857` / `36963`, then `…8039 tokens truncated…`).
4. After Chrome's focused window/tab changed, later actions failed with:

```text
SkyComputerUseError: The user changed '/Applications/Google Chrome.app'. Re-query the latest state with `get_app_state` before sending more actions.
```

5. After a node_repl reset or a failed nested call, the same session then reported `sky is not defined` and `Identifier 'x' has already been declared`.

OpenCodex itself stayed healthy: dashboard `localhost:10100/#logs` showed HTTP 200 for the Cursor-routed turns. The break is in how large Computer Use / node_repl results come back through the Cursor adapter, not in the proxy being down.

The Cursor adapter already bounds replayed history to `CURSOR_EXTERNAL_ROOT_BYTE_LIMIT = 512 * 1024` and can truncate or drop tool-result blobs (`src/adapters/cursor/protobuf-request.ts`, `truncateToolResultBlob`). A Computer Use `get_app_state` payload is a full accessibility tree plus a JPEG screenshot; that is enough to hit the budget and leave the next turn with an empty or head-only result.

## Expected behaviour

When Codex Desktop Computer Use / node_repl is used on a Cursor-routed turn:

[...]

PR #1990, opened 2026-08-18
fix(cursor): pin session id and continue external tool results as userMessageAction

laerad777:
## Summary

- External Cursor wire models now send tool-result continuations as `userMessageAction` (results stay in history blobs). Native models keep `resumeAction`.
- Connect `x-session-id` is pinned to the resolved conversation id so transport rebuilds for the same client thread keep one session identity. Native-exec / background shells use a separate per-transport owner.
- Isolated helper / compaction turns still mint a fresh conversation id, so they do not inherit the parent Connect session.

Rebase source commits: `0046816b3`, `4d5850e4d`, `e3c996586` on current `dev` (`1645bb924`).

## Verification

- `bun run typecheck` — clean
- `bun run privacy:scan` — Privacy scan passed
- `bun run test` — 12812 pass, 10 skip, 0 fail (826 files)
- Focused: `bun test tests/cursor-live-transport.test.ts tests/cursor-adapter.test.ts tests/cursor-blob.test.ts tests/cursor-hardening.test.ts tests/cursor-transport-retry.test.ts` — 147 pass

Live Connect probes informed the encoding choice; the unit tests only lock the action case, session-id forwarding, helper isolation, and shell-owner split. They do not claim to reproduce a 45k-token Connect reject.

## Checklist

- [x] Docs or release notes were updated when needed. (no public API / GUI / docs change)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (no auth, credentials, workflows, or install-path changes)

## Review readiness checklist

### DIFF
--- a/src/adapters/cursor.ts
+++ b/src/adapters/cursor.ts
@@ -121,6 +121,7 @@ export function createCursorAdapter(provider: OcxProviderConfig, deps: CursorAda
               headers: incoming.headers,
               translatorBudget: incoming.translatorBudget,
               requestDeclaresFullAccess: cursorRequestDeclaresFullAccess(activeRequest),
+              sessionId: activeRequest.conversationId,
             },
             activeRequest,
             incoming.abortSignal,
--- a/src/adapters/cursor/live-transport.ts
+++ b/src/adapters/cursor/live-transport.ts
@@ -432,9 +432,10 @@ class LiveCursorTransport implements CursorTransport {
  private firstFrameAt?: number;
  private firstFrameLogged = false;
   /** Stable session identifier sent as x-session-id; mirrors IDE session semantics. */
-  private readonly sessionId = crypto.randomUUID();
+  private readonly sessionId: string;
 
   constructor(private readonly input: CursorTransportFactoryInput) {
+    this.sessionId = input.sessionId?.trim() || crypto.randomUUID();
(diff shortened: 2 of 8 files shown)


lidge-jun:
Keeping this draft open — direction for whoever picks it up (maintainer decision):

**What is already on `dev` (do not re-implement):** external tool results continuing as `userMessageAction` landed with the cursor-call campaign. Rebase first; the half of this diff that duplicates it should disappear in the rebase.

**What this PR is still uniquely for:** session-id pinning. The adapter still generates a fresh `randomUUID()` per request, so Cursor sees every turn as a new thread and server-side thread reuse/caching never engages. The remaining work is to pin one stable session id per logical conversation and reuse it across continuations.

**Constraints for the implementation:**
- Derive the pinned id from the conversation identity the bridge already tracks — do not persist a new store for it.
- The id must survive tool-continuation round-trips but must NOT leak across different logical conversations (two parallel Codex threads must not share one Cursor session).
- Respect the ordering constraint recorded on #1866/#1992: #1997 (assistant-role tool-result replay) merges first; rebase this on top so no user-role replay path returns.
- Add a regression test asserting id stability across a tool continuation and id isolation across two conversations.

After rebase, re-scope the PR description to session pinning only, then it can leave draft.

PR #2019, opened 2026-08-18
refactor(types): extract value clusters into types/tools and types/wire leaves

lidge-jun:
## Summary

First stacked PR of the mega-file split program (devlog/_plan/260818_megafile_split_program, risk basis in 000_risk_assessment.md). Pure move, zero behavior change: the two runtime **value** clusters leave src/types.ts (1867 -> 1729 lines) into leaf modules, with src/types.ts keeping every current export as a barrel re-export so no consumer or test import changes.

- **src/types/tools.ts** — OcxTool, OcxToolChoice, namespacedToolName, toolChoiceAliases, toolAllowedByChoice, resolveToolChoiceWireName, modelInList, isAllowedToolChoice, toolChoiceToolPredicate. Zero imports; OcxTool travels with the cluster so the leaf never imports back through its own barrel (audit finding).
- **src/types/wire.ts** — UPSTREAM_HTTP_VERSION_VALUES, REASONING_SUMMARY_DELIVERY_VALUES, OPENAI_PROVIDER_TIER_VERSION, MODEL_ADAPTER_OVERRIDE_ALLOWED, ANTHROPIC_WIRE_MODELS internals, captureWireAdapterHardPins, isWirePinnedModel, pinnedWireAdapter, UpstreamHttpVersion, ReasoningSummaryDelivery, CodexAccountMode. Zero imports.
- **src/types.ts** — moved bodies deleted; top-of-file `import type` lines bind the aliases the remaining interfaces still reference (`export type ... from` does not bind locally — audit finding); in-place `export`/`export type` blocks keep the public surface byte-compatible. Extensionless specifiers so the lab-boundary walker follows the new edges.

Value identity is preserved by ESM re-export semantics: one declaration site per const, so MODEL_ADAPTER_OVERRIDE_ALLOWED stays a single ReadonlySet instance. The plan was audited by two independent reviewers (grok-4.6 NEAR-PASS, gpt-5.6-sol FAIL on a barrel-cycle recipe defect) and amended before implementation; both blockers were plan-recipe issues, fixed as described above.

Next in the stack: type-cluster split (request/config/provider/accounts) on this branch's head.

## Verification

At head 194f9f2a9 on remote validation host (Linux):

- `bun run test`: **13,201 pass / 15 skip / 0 fail** across 840 files (482.9s)
- `tests/core-lab-boundary.test.ts`: 13 pass standalone (new static edges types.ts -> types/tools.ts, types/wire.ts are walked and stay off src/lab/)
[...]

PR #2029, opened 2026-08-18
fix(probe): classify a missing user session bus as absent

### LINKED ISSUE #1939: Sync failed: Refusing to write because ownership could not be proven: systemctl show exited 1: Failed to get D-Bus connection: No such file or directory.
### Client or integration

Codex CLI

### Area

Proxy and routing

### Summary

After installing latest version v2.24.2 the only original codex models remain and button Sync Now in the dashboard gives:

Sync failed: Refusing to write because ownership could not be proven: systemctl show exited 1: Failed to get D-Bus connection: No such file or directory.

### Reproduction

Also the button for restart proxy not helped

### Version

2.24.2 

### Operating system

Linux

### Provider and model

_No response_

### Logs or error output

```shell

```

_No response_

### Redacted configuration

```json

```

### Checks

yzxcj797:
## Summary

Fixes #1939.

## Root cause

On hosts that ship `systemctl` but run **no user session bus** (WSL without systemd, containers, systemd-less distros), `systemctl --user show` exits 1 with:

```
Failed to get D-Bus connection: No such file or directory
```

The systemd probe maps every non-zero status to `unknown`, `inspectNativeCodexOwnership` turns that into "ownership could not be proven", and admission refuses **every** sync/write — the dashboard's Sync Now fails indefinitely ("Refusing to write because ownership could not be proven").

## Fix — narrowly scoped

The two messages that mean *systemd user sessions are impossible on this host* now classify as **`absent`** — the same conclusion the probe already draws for `spawnFailed` (no user manager ⇒ no user unit can be registered ⇒ nothing to conflict with):

- `Failed to get D-Bus connection: No such file or directory` (session-bus socket absent — the exact message from this issue)
- `System has not been booted with systemd` (WSL's classic)

Every other non-zero status stays `unknown` exactly as before — importantly, `Failed to connect to bus` (a misconfigured `DBUS_SESSION_BUS_ADDRESS` **while the user manager runs**) keeps its existing pinned behavior and test.

## Tests

Three new tests in `tests/codex-service-manager-probe.test.ts`: both absent messages classify as `absent`, and the misconfigured-bus case stays `unknown`. The pre-existing "non-zero systemctl status is unknown" test is untouched and still passes.

## Review readiness checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved systemd detection when the system lacks a D-Bus session or is not booted with systemd.
  - These environments are now correctly reported as having no systemd installation.
  - Other systemd command failures continue to be reported as unknown when they cannot be classified.


Ingwannu (changes requested):
The reported sync failure is real, but the current classification is not safe for unattended writes.

A missing user bus does not prove that ~/.config/systemd/user/opencodex-proxy.service is absent. The definition can remain on disk while the user manager is unavailable, and it can name another CODEX_HOME or OPENCODEX_HOME. Returning absent before inspecting that artifact lets the ownership preflight treat the machine as unclaimed and overwrite a foreign or interrupted installation.

Please keep the bus result separate from definition evidence: when these deterministic no-bus messages occur, inspect the expected unit file. If the file is absent, return absent. If it exists, read and parse its homes and return a present claim with registration absent. If it is unreadable or malformed, remain unknown. Add regressions for both an agreeing and a foreign definition while the bus is unavailable.

This PR also targets main and is 125 dev commits behind; rebuild it on the latest dev without the release package-version delta.

lidge-jun:
## 리뷰 · 우선순위 28 / 80

draft 이고 readiness 4칸이 비어 있으며 CHANGES_REQUESTED 입니다. #1939 는 `systemctl` 은 있는데 user session bus 가 없는 호스트에서 모든 non-zero 를 `unknown` 으로 두어 sync/write 가 거절되던 문제입니다. 이 diff 는 두 stderr 만 `absent` 로 바꿉니다. 같은 커밋에 `package.json` `2.24.2` → `2.25.0` 이 있습니다.

`inspectSystemd` 는 `shown.status !== 0` 일 때 `Failed to get D-Bus connection: No such file or directory` 와 `System has not been booted with systemd` 부분 문자열이면 `{ kind: "absent" }` 입니다. `Failed to connect to bus` 는 그대로 `unknown` 입니다. 테스트 세 개가 그 경계를 고정합니다. 기존 non-zero unknown 테스트는 안 건드렸습니다.

매칭이 영어 문구에 묶여 있습니다. 로케일이 바뀌거나 systemd 가 문장을 조금 고치면 다시 unknown 이 됩니다. `includes` 는 앞뒤 장식에는 강하지만, `No such file or directory` 만으로 넓히면 다른 실패를 absent 로 오인할 수 있습니다. 지금 두 문자열은 본문이 인용한 이슈 메시지와 같습니다.

`absent` 는 “충돌할 user unit 이 없다”는 뜻이라 admission 이 write 를 허용합니다. 세션 버스가 잠깐 죽은 경우에도 같은 문구가 나오면 소유권을 증명하지 않고 쓰게 됩니다. 이슈의 WSL/컨테이너 재현에는 맞습니다.

해결방안: `package.json` 버전 hunk 를 빼십시오. CHANGES_REQUESTED 를 닫으십시오. 가능하면 status 와 문구를 같이 보는 주석을 유지하고, 로케일 변형이 필요하면 이슈에 남기십시오. draft 체크리스트를 채우십시오.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
Maintainer note — this is the containment @Ingwannu asked for, written out so it can be lifted straight into the branch. I am **not** opening a competing PR: this function is yours for this cycle.

## Why this matters beyond #1939

**#2114 is the same defect with a different symptom.** That report is a Docker/tini container where `systemctl` is on PATH but the user bus is unreachable, and the consequence there is not a refused sync — it is **100% native-OpenAI failure**, 503 `OpenCodex local native-main profile maintenance is active`, until `ocx restart`. Same chain: `inspectSystemd` → `unknown` → `ownership-preflight.ts:155` → `server/index.ts` → `blockNativeMainStartupForUnownedServiceHome` → `auth-context.ts`.

Its stderr is `Failed to connect to user scope bus via local transport: $DBUS_SESSION_BUS_ADDRESS and $XDG_RUNTIME_DIR not defined` — a third shape alongside the two this PR matches, and one that the added test currently pins as `unknown`.

So the classifier here decides two issues at once. That is an argument for landing it with the containment rather than splitting the work.

## The containment, using pieces this file already has

The review is right that the bus result and the definition evidence must stay separate. The good news is that `inspectSystemd` already does exactly the requested inspection **below** the bus branch — lines 289-313 already: check `artifactPresence(definitionPath)`, read the file, parse `unitEnvValue(body, "CODEX_HOME")` / `OPENCODEX_HOME`, and return a `present` claim. Nothing new has to be written; the no-bus path just has to reach it.

Shape:

```ts
if (shown.status !== 0) {
  if (!busUnreachable(shown.stderr)) {
    return unknown(`systemctl show exited ${String(shown.status)}: ${err}`);
  }
  // The bus could not answer. Ask the disk instead — a unit file is proof of
  // installation that does not require a running user manager.
  const presence = artifactPresence(definitionPath);
  if (presence === "absent") return { kind: "absent" };
  if (presence !== "present") return unknown("the systemd unit could not be inspected while the user bus was unavailable");

  let body: string;
  try { body = readFileSync(definitionPath, "utf-8"); }
  catch (error) { return unknown(`the systemd unit exists but could not be read: ${String(error)}`); }

  return {
    kind: "present",
    claims: [{
      backend: "systemd",
      definitionPath,
      homes: {
        codexHome: unitEnvValue(body, "CODEX_HOME"),
        opencodexHome: unitEnvValue(body, "OPENCODEX_HOME"),
      },
      registration: "absent",   // the bus never answered, so registration is genuinely unknown-as-absent
    }],
  };
}
```

`registration: "absent"` is the review's "present claim with registration absent" — the file exists, but nothing proves systemd has it loaded.

## On matching stderr

Worth naming as a real weakness rather than papering over: string matching is locale-fragile. A non-English systemd will not match any of the three shapes, and those users keep getting `unknown`.

Two options, and I do not think this PR has to pick the harder one:

- **Keep the match, accept the gap.** With the disk check above, a mismatch is now only a missed *recovery*, never a wrong *admission* — the failure mode degrades from "wrongly claims the machine" to "stays conservative". That is a much safer thing to be locale-sensitive about.
- **Or force `LC_ALL=C` on the probe invocation** so the messages are stable. Larger change, and it touches every other probe path.

## Suggested regressions

Matching the review's ask, plus the #2114 shape:

1. bus-unreachable + **no** unit file → `absent` → ownership `owned` → native-main not blocked
2. bus-unreachable + unit file naming **our** homes → `present`, registration `absent`, ownership agrees
3. bus-unreachable + unit file naming a **foreign** home → still fenced (this is the one the review is protecting)
4. bus-unreachable + unit file present but **unreadable/malformed** → stays `unknown`
5. the #2114 stderr (`Failed to connect to user scope bus via local transport…`) classified with the same branch, so it stops being pinned as `unknown`
6. a **non**-bus non-zero exit → still `unknown` (the existing assertion, kept as the guard against over-widening)

## Rebase note

The review also flagged the base and the 125-commit lag. That is stale now — the PR reads `base=dev`, and `dev` has since taken the mega-file split (#2019/#2023/#2036) plus a merge train, so a rebase will be needed regardless before this can be judged.

Happy to do the rebase or write the tests if that is useful — say the word rather than me pushing to your branch.

lidge-jun:
Thanks for this, @yzxcj797 — closing as superseded by #2130, which merged into `dev` earlier today.

#2130's `busUnreachable()` matches a wider set of stderr signatures than the two strings here, and then does what @Ingwannu asked for in review: it inspects the unit files on disk rather than returning `absent` unconditionally. A missing bus is not a missing unit, and with #2130 landed a foreign or stale unit now reports `present`/`unknown` instead of fail-open `absent`.

That is why nothing is folded from this PR: rebasing it on top of #2130 would replace the disk check with the unconditional `absent` that the review flagged. The underlying reports (#1939, #2114) are addressed by the merged change.

PR #2032, opened 2026-08-18
fix(claude): allow explicit --dangerously-skip-permissions on root hosts

### LINKED ISSUE #1688: [Bug]: ocx claude --dangerously-skip-permissions aborts as root (uid==0) because IS_SANDBOX is never set

yzxcj797:
## Summary

Fixes #1688.

## Root cause (as reported)

Claude Code 2.1.205+ refuses `--dangerously-skip-permissions` when running as root unless `IS_SANDBOX=1` marks the environment as contained. `buildClaudeEnv` never sets it, so `ocx claude --dangerously-skip-permissions` on a root VPS aborts immediately with the vendor's guard message — the default deployment shape on most VPS hosts.

## Fix — one deliberate deviation from the proposed diff

`IS_SANDBOX=1` is set **only when the user explicitly passed `--dangerously-skip-permissions`**:

- `cmdClaude` detects the flag in `args` and threads it through a new `ClaudeEnvDeps.dangerouslySkipPermissionsRequested`;
- `buildClaudeEnv` sets `IS_SANDBOX` via `setDefault` for those launches only.

Why not unconditionally (the issue's proposed diff): the variable exists to relax a vendor safety guard, and marking every proxy-owned session as sandboxed could silence other guard behaviors in future Claude Code versions. An ordinary `ocx claude` launch keeps the vendor's default posture; the bypass is honored exactly when asked for. An explicit user export of `IS_SANDBOX` still wins (existing `setDefault` semantics).

## Tests

`tests/claude-cli.test.ts`:

- a bypass launch sets `IS_SANDBOX=1`;
- an ordinary launch does not set it at all;
- an explicit user `IS_SANDBOX=0` export wins over the bypass default.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of launches that explicitly skip permission checks.
  * Ensured the sandbox environment setting is applied only when requested.
  * Preserved user-provided sandbox environment values.
  * Ordinary subscription launches no longer receive an unintended sandbox setting.


Ingwannu (changes requested):
Requesting changes on the current head for two independent blockers.

1. Passing --dangerously-skip-permissions does not create an OS sandbox. Setting IS_SANDBOX=1 still makes a false isolation claim and suppresses the upstream root guard while the child has ordinary root filesystem and process access. Pinning provider traffic to the loopback proxy is not an OS security boundary. This issue already records the required contract: set IS_SANDBOX only after a real isolation mechanism has positively succeeded, and fail closed if that setup fails. Unit tests that only assert the environment variable do not prove that boundary.

2. This PR targets main and is 125 dev commits behind. Please rebuild the focused change on the latest dev without the release package-version delta.

A viable revision needs a real sandboxed launch path, or it must leave the vendor root guard intact.

lidge-jun:
## 리뷰 · 우선순위 27 / 80

draft 이고 readiness 4칸이 비어 있으며 CHANGES_REQUESTED 입니다. #1688 은 Claude Code 가 root 에서 `--dangerously-skip-permissions` 를 `IS_SANDBOX=1` 없이 거절하는 문제입니다. 이 PR 은 플래그가 args 에 있을 때만 `buildClaudeEnv` 가 `setDefault("IS_SANDBOX", "1")` 을 호출합니다. 이슈의 무조건 설정안보다 범위가 좁습니다. 같은 diff 에 `package.json` `2.24.2` → `2.25.0` 이 있습니다.

`cmdClaude` 는 `args.includes("--dangerously-skip-permissions")` 만 봅니다. 값이 붙은 형태나 다른 위치의 동의어는 놓칩니다. root 여부도 보지 않아서, 비-root 에서도 플래그가 있으면 `IS_SANDBOX` 를 넣습니다. 프로세스가 실제로 격리되어 있지 않은데 벤더 가드에 sandbox 라고 알립니다. 사용자가 위험 플래그를 직접 넘긴 경우에만 그렇게 합니다. 기존 `setDefault` 때문에 사용자 `IS_SANDBOX=0` 이 이깁니다. 그 조합은 root 에서 다시 거절됩니다.

테스트는 `buildClaudeEnv` 에 deps 플래그를 직접 넣습니다. `cmdClaude` 의 `args.includes` 경로는 없습니다. 본문은 로컬 bun 을 못 돌렸습니다.

해결방안: `package.json` 버전 hunk 를 빼십시오. CHANGES_REQUESTED 를 닫고 플래그 파싱을 Claude 가 받는 실제 argv 형태와 맞추십시오. root 가 아닐 때 `IS_SANDBOX` 를 넣지 않을지 결정하십시오. `cmdClaude` 인자 → env 조립을 한 테스트로 연결하십시오. draft 체크리스트를 채우십시오.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
Thank you for this, @yzxcj797 — and the underlying problem is real. Someone running in a root container hits Claude Code's root guard and cannot get past it, which is a legitimate workflow to want.

We are closing this, and the reason is the one already stated in review: passing `--dangerously-skip-permissions` does not prove the child is sandboxed. The patch injects `IS_SANDBOX=1` whenever the flag appears in argv (`src/cli/claude.ts:126-131`, `:329-331`) with no UID check and no sandbox establishment, so it suppresses the vendor's root guard while the child keeps ordinary root filesystem and process access. Routing traffic through a loopback proxy is not OS-level isolation. The tests at `tests/claude-cli.test.ts:271-295` confirm the environment variable is assembled correctly, which is a different claim from an isolation boundary existing.

Worth being clear about what `dev` does today, because it is not a block on our side: `src/cli/dispatch.ts:500` forwards trailing arguments and `src/cli/claude.ts:338` passes them through unchanged. The refusal comes from Claude Code itself.

For a genuinely isolated environment there is already a supported path: `buildClaudeEnv` starts from the caller's exported environment (`src/cli/claude.ts:76`) and exported variables win (`docs-site/src/content/docs/guides/claude-code.md:56`), so `IS_SANDBOX=1` can be exported deliberately by an operator who knows their isolation holds. That keeps the decision — and the responsibility — with the person who can actually verify it, instead of inferring it from a CLI flag.

A future version of this would need to create and positively verify a real sandbox, fail closed when setup fails, and leave the unrelated `package.json` version bump out. That would be a welcome PR.

Thanks again for raising it.

PR #2037, opened 2026-08-18
feat(outbound): route Clash fake-IP DNS answers through the configured proxy (#1748)

lidge-jun:
## Summary

Scoped re-implementation of PR #1748 per the 260818 campaign disposition matrix (REDESIGN-SMALL: outbound-only fake-IP proxy routing, avoiding the SSRF widening the original had).

- `resolvePublicAddresses` gains an explicit `allowBenchmarkAddresses` opt-in: a hostname answer in 198.18.0.0/15 (IANA benchmark space — Clash/Surge/Mihomo fake-IP DNS) is accepted without marking the destination private, so the hostname keeps its configured HTTP(S) proxy path.
- Containment: literal 198.18.x URLs still reject; mixed answers containing any other non-public address still reject; callers that do not pass the flag (image fetch, Lab fetch — full caller sweep in the plan audit) keep rejecting.
- `provider-outbound` arms the opt-in only when an outbound proxy is configured (`proxyConfigured`), so the hostname rides the proxy CONNECT instead of pin-connecting to the fake IP. The NO_PROXY per-host corner is documented at the call site.
- Supersedes and credits #1748 (luvs01). Campaign unit: devlog/_plan/260818_bug_pr_resolution (020 doc).

## Verification

- `bun test ./tests/destination-policy-resolved.test.ts ./tests/provider-outbound.test.ts` — 43 pass / 0 fail (5 new destination-policy cases: opt-in accept, no-opt-in reject, mixed reject, literal reject, image-fetch reject + proxy integration cases)
- `bun x tsc --noEmit` exit 0
- Adversarial plan audit (r2): all callers of resolvePublicAddresses swept; no unflagged caller affected

## Checklist

- [x] Focused tests added and passing
- [x] Typecheck green
- [x] No new dependencies
- [x] Original PR credited; will be closed with a pointer here

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved proxied provider requests using Clash fake-IP benchmark addresses.
  * Benchmark address resolutions remain blocked for direct requests and unsupported destinations.
  * Mixed private and benchmark address results continue to require private-network access.

* **Tests**
  * Added coverage for proxied and direct fake-IP resolution scenarios, including hostname and literal IP requests.

### DIFF
--- a/src/lib/destination-policy.ts
+++ b/src/lib/destination-policy.ts
@@ -241,10 +241,18 @@ export function assessUrlDestination(url: string): UrlDestinationAssessment | nu
  * Returns the validated addresses so direct callers can pin the connect peer and
  * avoid a second, rebindable resolution. DNS failures remain fail-closed here;
  * the provider proxy wrapper alone may recognize that typed failure and degrade.
+ *
+ * `allowBenchmarkAddresses` is an explicit outbound-only opt-in for Clash/Surge/
+ * Mihomo fake-IP DNS (IANA benchmark space 198.18.0.0/15, credit #1748): a hostname
+ * answer in that range is accepted without marking the destination private, so the
+ * caller can keep the hostname on its configured HTTP(S) proxy path. It applies to
+ * resolved answers only — a literal 198.18.x URL still rejects — and mixed answers
+ * that include any other non-public address still fail. Callers that do not pass it
+ * (image and Lab fetch) keep rejecting benchmark space.
  */
 export async function resolvePublicAddresses(
   url: string,
-  options?: string | { context?: string; allowPrivateNetwork?: boolean },
+  options?: string | { context?: string; allowPrivateNetwork?: boolean; allowBenchmarkAddresses?: boolean },
 ): Promise<{
   hostname: string;
   addresses: { address: string; family: number }[];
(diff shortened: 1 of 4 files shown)


Ingwannu (changes requested):
The fake-IP compatibility direction is sound, but the current head widens the destination policy before it knows that this host will actually use the proxy.

`allowBenchmarkAddresses` is derived from global proxy presence alone. When the provider hostname matches `NO_PROXY`, the resolver therefore accepts a 198.18/15 answer and `resolved.privateNetwork` remains false; the next branch calls `globalThis.fetch`, which honors `NO_PROXY` and directly connects to the benchmark address. That contradicts the PR's stated hostname-CONNECT boundary and admits a previously blocked non-public destination on a direct path.

Please arm the benchmark exception only when a proxy is configured **and** `!noProxyMatches(parsed)`. Add an exact regression with `NO_PROXY=www.packyapi.com` that proves the benchmark answer is rejected and neither the proxy fetch nor pinned direct transport is invoked. Keep literal benchmark URLs and mixed unsafe DNS answers rejected as they are now. This is an outbound security-policy boundary, so I am requesting changes rather than approving the otherwise focused implementation.

Ingwannu:
Follow-up #2045 covers the remaining per-host routing boundary from this merge. The benchmark exception was armed from global proxy presence, but a `NO_PROXY` match makes Bun take a direct route; that allowed a 198.18/15 DNS answer on a path whose policy previously rejected it.

The follow-up enables the exception only when the exact host will use the proxy and adds a regression proving a NO_PROXY-matched fake-IP answer invokes neither proxy fetch nor pinned direct transport. Focused destination/outbound tests (44), typecheck, and privacy scan are green; independent review and CI are pending.

PR #2038, opened 2026-08-18
fix(cursor): normalize empty and failure-state Computer Use tool results (#1920)

### LINKED ISSUE #1866: [Compatibility] Cursor Computer Use / node_repl tool results come back empty or truncated

lidge-jun:
## Summary

Scoped re-implementation of PR #1920 per the 260818 campaign disposition matrix (REDESIGN-SMALL: apply formatted text at the native toolResultPart + decode test). Resolves #1866 (Cursor Computer Use / node_repl tool results come back empty or truncated).

- New `src/adapters/cursor/tool-result-normalize.ts`: blank or empty-exec-wrapper output on node_repl / Computer Use tools becomes an actionable error ("verify application state with get_app_state"); known runtime failure states reported as plain text (SkyComputerUseError, sky is not defined, redeclared identifier, unsupported import in exec) are marked isError with one-line recovery guidance. All other text passes through byte-identical.
- Wired at all four wire sites — including the two the original PR covered but a naive port would miss (found in plan audit): `toolResultContentItems` (native McpText), `toolResultPart` (McpSuccess.isError), `toolResultToText` (replay text), the externalModel branch of `conversationTurns` (the cursor/grok-4.6 repro path), and the root-prompt prefix.
- Decode test proves the native ConversationStep wire carries normalized text + isError via fromBinary, with unit rows per failure state and byte-identical passes for non-computer-use tools.

Deferred (disclosed on #1866 at close): screenshot stripping and AXTree text compaction from the original 867-line PR — the native path already bounds step size by real serialized bytes, dropping images oldest-first.

Supersedes and credits #1920. Campaign unit: devlog/_plan/260818_bug_pr_resolution (020 doc).

## Verification

- `bun test ./tests/cursor-toolresult-normalize.test.ts` — 12 pass (incl. 3 native-wire decode tests)
- `bun test tests/cursor-` glob — 650 pass / 0 fail across 36 files
- `bun x tsc --noEmit` exit 0

## Checklist

- [x] Focused decode + unit tests added and passing
- [x] Full cursor suite green
- [x] Typecheck green
- [x] Original PR credited; will be closed with a pointer here

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of empty or failed tool results so they are reported as actionable errors.
  * Added clearer recovery guidance for recognized runtime failures.
  * Ensured normalized error details are consistently shown across replayed and native tool results.
  * Preserved existing behavior for successful, encrypted, image-based, and unrelated tool outputs.
* **Reliability**
  * Improved consistency of tool-result status and displayed text across supported tool interactions.

### DIFF
--- a/src/adapters/cursor/protobuf-request.ts
+++ b/src/adapters/cursor/protobuf-request.ts
@@ -5,6 +5,7 @@ import type { OcxAssistantContentPart, OcxMessage, OcxToolResultMessage } from "
 import { namespacedToolName } from "../../types";
 import type { CursorRunRequest } from "./types";
 import { isCursorExternalWireModel } from "./discovery";
+import { normalizeCursorToolResultText } from "./tool-result-normalize";
 import { debugProviderDiagnostic } from "../../lib/debug";
 import {
   createCursorBlobRequestScope,
@@ -240,7 +241,9 @@ function rootPromptMessages(request: CursorRunRequest, requestScope: CursorBlobR
       }
       // Assistant tool CALLS are intentionally NOT replayed as visible "[Tool Call]" text here.
     } else if (message.role === "toolResult") {
-      const prefix = message.isError ? "[Tool Error]" : "[Tool Result]";
+      // #1920: the prefix must reflect the NORMALIZED error state (an empty
+      // node_repl result is an error even when the runtime said isError=false).
+      const prefix = normalizedToolResult(message, contentToText(message.content)).isError ? "[Tool Error]" : "[Tool Result]";
       const text = `${prefix}\n${toolResultToText(message)}`;
       entries.push(rootBlobCandidate(
         toolResultRootPayload(text),
@@ -427,7 +430,11 @@ function toolResultContentItems(
 ) {
   const parts = decoded ?? decodeResultParts(message);
   if (!parts) {
-    const text = typeof message.content === "string" ? message.content : "";
+    const raw = typeof message.content === "string" ? message.content : "";
+    // #1920/#1866: empty or failure-state Computer Use / node_repl results are
+    // normalized before they reach the native wire (isError is applied in
+    // toolResultPart via normalizedToolResult below).
+    const { text } = normalizedToolResult(message, raw);
     return [create(McpToolResultContentItemSchema, {
       content: { case: "text" as const, value: create(McpTextContentSchema, { text }) },
     })];
@@ -483,14 +490,28 @@ function toolResultContentItems(
 }
 
 function toolResultToText(message: OcxToolResultMessage): string {
+  const normalized = normalizedToolResult(message, contentToText(message.content));
   return [
     "[tool_result]",
     `call_id: ${message.toolCallId}`,
     `name: ${namespacedToolName(message.toolNamespace, message.toolName)}`,
-    `is_error: ${message.isError}`,
+    `is_error: ${normalized.isError}`,
     "output:",
-    contentToText(message.content),
+    normalized.text,
   ].join("\n");
+}
+
+/**
+ * Shared #1920 normalization entry: pure-text results only. Image-bearing or
+ * encrypted results pass through untouched (their content is not plain text).
+ */
+function normalizedToolResult(message: OcxToolResultMessage, text: string): { text: string; isError: boolean } {
+  if (message.containsEncryptedContent) return { text, isError: message.isError };
+  return normalizeCursorToolResultText(text, {
+    toolName: message.toolName,
+    toolNamespace: message.toolNamespace,
+    isError: message.isError,
+  });
 }
 
 function argBytes(value: unknown): Uint8Array {
@@ -546,11 +567,15 @@ function toolCallStep(
 }
 
 function toolResultPart(message: OcxToolResultMessage, decoded?: DecodedResultPart[], maxImages?: number) {
+  const parts = decoded ?? decodeResultParts(message);
+  const normalizedIsError = parts
+    ? message.isError
+    : normalizedToolResult(message, typeof message.content === "string" ? message.content : "").isError;
   return create(McpToolResultSchema, {
     result: {
       case: "success",
       value: create(McpSuccessSchema, {
-        isError: message.isError,
+        isError: normalizedIsError,
         content: toolResultContentItems(message, decoded, maxImages),
       }),
     },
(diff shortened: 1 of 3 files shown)


Ingwannu (changes requested):
The narrow redesign is valuable and the string-result path is correctly covered, but the current exact head still has one merge blocker: text-only structured content bypasses the normalization entirely.

`decodeResultParts()` returns an array for `OcxContentPart[]`, so `toolResultContentItems()` takes the `parts` branch and emits each text part without calling `normalizedToolResult()`. `toolResultPart()` also keeps `message.isError` whenever `parts` exists. Therefore an empty or known-failure node_repl result represented as `[{ type: "text", text: "..." }]` reaches the native Cursor wire unchanged with `isError: false`, even though the byte-equivalent raw string is repaired.

Please normalize an all-text content array through the same shared `{ text, isError }` result before native serialization. Preserve image-bearing, encrypted, and undecodable content exactly as today. Add native-wire regressions for both an empty text-part array and a known failure marker, plus a control proving image-bearing content is untouched. Once that is fixed and exact-head CI remains green, this should stay a strong merge candidate for #1920.

Ingwannu:
Follow-up #2044 covers the remaining exact-head gap from this merge: pure-text `OcxContentPart[]` results took the structured-parts path, so empty and known-failure node_repl output bypassed the new normalization even though byte-equivalent string results were repaired.

The follow-up shares the normalization result between native protobuf text and `isError`, while preserving image-bearing, undecodable-image, and encrypted results unchanged. Focused Cursor tests (97), typecheck, and privacy scan are green. It is awaiting independent review and CI.

Issue #1225, opened 2026-08-07
Support configuring a custom auto-review model for approvals_reviewer = "auto_review"


yusoofsh:
### Area
- Catalog / models

### What are you trying to accomplish?

With `approvals_reviewer = "auto_review"` enabled, route the automatic approval-review subagent through a user-chosen model (for example `opencode-go/deepseek-v4-flash`) instead of whatever OpenCodex/Codex resolves by default.

### What prevents this today?

Upstream Codex only exposes two reviewer modes in `approvals_reviewer`: `user` and `auto_review` (legacy alias `guardian_subagent`) — no model argument. The model the auto-review subagent uses is decided per-model via the `auto_review_model_override` field on the model catalog entry of the currently running model (see `codex-rs/core/src/guardian/review.rs` and `codex-rs/protocol/src/openai_models.rs` in openai/codex).

OpenCodex regenerates `opencodex-catalog.json` from provider discovery during catalog sync, and routed entries are rebuilt from templates. There is no OpenCodex-managed way to express "use model X for auto review", so the only current approach is hand-editing the generated catalog file. That edit is undocumented and can be clobbered by a later sync.

### What should OpenCodex do?

Provide a first-class, durable configuration for the auto-review model that survives catalog regeneration. For example:

- Accept a config option (e.g. in `~/.codex/config.toml` or an OpenCodex-managed setting) such as `auto_review_model = "opencode-go/deepseek-v4-flash"`.
- Have the catalog sync apply that value to `auto_review_model_override` on the relevant routed entries instead of leaving them at `null`, so the setting is not lost when entries are rebuilt.
- Document the option in the catalog / models docs.

### Example usage or interface

```toml
approvals_reviewer = "auto_review"
auto_review_model = "opencode-go/deepseek-v4-flash"
```

Expected observable behavior: when an approval prompt needs review, the auto-review subagent runs against `opencode-go/deepseek-v4-flash` rather than the upstream default (`codex-auto-review` or `gpt-5.6-luna`), and the choice persists across `opencodex` catalog syncs.

### Alternatives or workarounds

- Manually add `"auto_review_model_override": "<model-slug>"` to the model's entry in `~/.codex/opencodex-catalog.json`. This works today (verified), but the catalog is regenerated by OpenCodex and routed entries are rebuilt from templates, so the edit can be silently dropped.
[...]

PR #2041, opened 2026-08-18
feat(catalog): durable auto_review_model config override

yzxcj797:
## Summary

Implements #1225.

## The gap (as researched in the issue)

With `approvals_reviewer = "auto_review"`, upstream Codex resolves the review subagent's model from the **current turn's** `auto_review_model_override` catalog field (defaulting to `codex-auto-review` / `gpt-5.6-luna`). OpenCodex ships the field as `null` on every entry and rebuilds the catalog from templates on sync — so the only way to choose a review model was hand-editing the generated `opencodex-catalog.json`, an edit the next sync silently drops.

## What this does

- `auto_review_model` is read as a **root key of `~/.codex/config.toml`** — exactly the placement from the issue's example, next to `approvals_reviewer`:
  ```toml
  approvals_reviewer = "auto_review"
  auto_review_model = "opencode-go/deepseek-v4-flash"
  ```
- The value is stamped onto **every catalog entry** as `auto_review_model_override` at the single choke point all entries pass through immediately before serialization in `writeRetainedCatalogSync`. Upstream reads the field from whichever model runs the turn, so stamping every entry makes the choice effective regardless of the active model — and immune to the regenerations that clobber hand edits.
- Absent key → entries keep whatever they carry (upstream defaults preserved). Unreadable config → same.
- Tests cover the root-key TOML resolution (including that table-scoped keys like `[profiles.test] model = ...` don't leak into root resolution) and the absent-key case.

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Catalog entries now include the configured automatic review model when one is available.

* **Bug Fixes**
  * Improved configuration handling so automatic review settings are resolved only from the correct root-level configuration.
  * Missing automatic review configuration is handled gracefully.


Ingwannu:
I retargeted this draft from release-only `main` to the integration branch `dev`, but the current head needs a full rebase before review: its merge base is 128 `dev` commits behind.

There is also an immediate compile blocker on the submitted head: `writeRetainedCatalogSync()` calls `configuredAutoReviewModel()`, but that symbol is neither defined nor imported anywhere in the PR. The two tests exercise only `readRootTomlString`; they do not execute the catalog write path and therefore cannot catch this.

After rebasing, please implement the durable setting through the current config/catalog boundaries and cover the actual emitted catalog. The review requirements from #1225 still apply:

- validate the requested routed slug against the catalog produced by the same sync and surface an unresolved/stale target rather than silently stamping a typo;
- prove regeneration persistence and actual `auto_review_model_override` output for routed and native rows;
- cover provider/model removal and recovery after the target returns;
- preserve existing upstream entry values when the setting is absent;
- document the supported configuration/CLI surface.

This is a real feature gap, but the current draft is not a merge candidate until the branch, missing symbol, validation, and end-to-end tests are corrected.

lidge-jun:
## 리뷰 · 우선순위 24 / 80

draft 이고 readiness 4칸이 비어 있습니다. #1225 를 `~/.codex/config.toml` 루트 키 `auto_review_model` 로 받아 `writeRetainedCatalogSync` 직전 모든 catalog entry 에 `auto_review_model_override` 를 찍습니다. 핸드 에디트가 sync 에 지워지던 구멍을 막으려는 방향은 이슈와 맞습니다. 같은 diff 에 `package.json` `2.24.2` → `2.25.0` 이 들어 있습니다. #2063 / #2062 와 같은 무관 버전 hunk 입니다.

`configuredAutoReviewModel()` 가 truthy 일 때만 `catalog.models.map` 으로 전 엔트리에 같은 문자열을 덮습니다. 엔트리마다 다른 override 가 있어도 루트 키 하나가 전부 대체합니다. 키가 없거나 unreadable 이면 기존 필드를 유지한다고 본문에 적혀 있고, 이 네 줄은 그 경우 map 을 건너뜁니다. 모델 id 가 카탈로그에 있는지는 검사하지 않습니다. 잘못된 `opencode-go/...` 값도 모든 엔트리에 그대로 박힙니다.

테스트는 `readRootTomlString` 만 봅니다. 루트 `auto_review_model` 과 `[profiles.test] model` 이 섞이지 않는지, 키 부재가 `null` 인지는 맞습니다. `writeRetainedCatalogSync` 가 실제로 override 를 찍는지, 키 부재 때 기존 값이 남는지, 다음 sync 에도 유지되는지는 이 파일에 없습니다. 본문은 로컬 bun 을 못 돌렸다고 합니다.

해결방안: `package.json` 버전 변경을 빼십시오. `writeRetainedCatalogSync` 전후 catalog fixture 로 override 스탬프와 키 부재 보존을 테스트하십시오. 값이 카탈로그에 없는 모델일 때의 실패 모드를 정하십시오. draft 체크리스트를 채우기 전에는 ready 로 보지 마십시오.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (commented):
Reviewed exact head e246024. This draft is 525 dev commits behind and conflicts. The current Grok blockers remain sufficient to stop merge: remove the unrelated package version hunk, test the actual catalog write and regeneration path rather than only the TOML reader, preserve existing per-entry values when the root setting is absent, and define validation and recovery for an unresolved model target. Rebase onto current dev and rerun exact-head CI before requesting another review.

lidge-jun:
-

PR #2045, opened 2026-08-18
fix(outbound): keep fake-IP opt-in off NO_PROXY routes

Ingwannu:
## Summary

- grant the 198.18/15 fake-IP DNS exception only when the exact provider host will use the configured proxy
- keep `NO_PROXY`-matched hosts on the ordinary direct-route destination policy, where benchmark answers remain blocked
- add a regression proving neither proxy fetch nor pinned direct transport runs for a NO_PROXY-matched fake-IP answer

Follow-up to #2037 and the remaining per-host routing boundary in #1748.

## Verification

- `taskset -c 0,1 bun test tests/provider-outbound.test.ts tests/destination-policy-resolved.test.ts` — 44 passed
- `taskset -c 0,1 bun run typecheck`
- `taskset -c 0,1 bun run privacy:scan`
- `git diff --check`

## Checklist

- [x] Docs or release notes were updated when needed (the existing code comment now matches the per-host proxy boundary; no public option changed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved provider DNS validation when proxies and `NO_PROXY` rules are configured.
  * Hosts bypassing the proxy now continue to reject benchmark addresses, preventing incorrect routing behavior.

* **Tests**
  * Added regression coverage for `NO_PROXY` hosts and benchmark-address validation.

### DIFF
--- a/src/lib/provider-outbound.ts
+++ b/src/lib/provider-outbound.ts
@@ -149,16 +149,12 @@ async function providerOutboundRequest(
       context: "provider URL",
       allowPrivateNetwork: allowPrivate,
       // Clash/Surge/Mihomo fake-IP DNS (198.18.0.0/15) answers are admitted only
-      // when an outbound proxy is configured: the hostname then rides the proxy as
-      // an ordinary CONNECT instead of failing as a private destination or being
-      // pin-connected to the fake-IP (credit #1748). Without a proxy, benchmark
-      // answers keep rejecting. Image/Lab fetch never passes this flag.
-      // Known corner: the opt-in arms on the GLOBAL proxy config, not per-host. If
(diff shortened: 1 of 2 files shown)


lidge-jun:
Merged. One follow-up recorded from independent review: `noProxyMatches` (pre-existing helper from #2037's unit, e3bbf5321) reads `NO_PROXY ?? no_proxy`, while Bun's env loader prefers non-empty `no_proxy` first — when the two variables differ, the matcher and Bun's fetch can disagree about whether a host is proxied, which re-opens a direct fake-IP corner this PR otherwise closes. Since this PR only *narrows* the allowance (`proxyConfigured && !noProxyMatches`), it is a strict improvement and the precedence alignment (prefer non-empty lowercase, plus a conflict regression test) is left as a follow-up.

PR #2028, opened 2026-08-18
feat(quota): support GLM coding plan quota on z.ai and bigmodel.cn

jamespan:
## Summary

Adds GLM Coding Plan quota probing for the `zai` provider family across both Z.AI regions:

- Probes `/api/monitor/usage/quota/limit` with the API key as a Bearer token on both `api.z.ai` and `open.bigmodel.cn`, including the OpenAI Responses endpoint `https://open.bigmodel.cn/api/v1`.
- Parses the `limits` array: `TOKENS_LIMIT`/`CREDIT_LIMIT` rows with `unit` 3 / `number` 5 map to the five-hour window and `unit` 6 / `number` 1 to the weekly window; `TIME_LIMIT` rows map to the monthly MCP tool budget.
- Falls back to the legacy field-name payload shape (`fiveHourPercent`, `weeklyPercent`, `monthlyMCPUsage`, …) only when the `limits` array is absent.
- Dispatches quota probing for `zai`, `glm`, `glm-cn`, and `zhipu-bigmodel-coding`; the pay-as-you-go `/api/paas/v4` route never probes.
- Documented in the providers guide (`docs-site/src/content/docs/guides/providers.md`).

## Evidence: real `limits` responses (sanitized, from the live probe)

The v2 coding-plan protocol reports the monthly MCP budget as a `TIME_LIMIT` row; the newer protocol does not. The fixtures in `tests/provider-quota.test.ts` are captured from these live responses, not invented:

**v2 protocol (`level: max`)** — carries the monthly MCP `TIME_LIMIT` row:

```json
{ "limits": [
    { "type": "TIME_LIMIT", "unit": 5, "number": 1, "usage": 4000, "currentValue": 0, "remaining": 4000, "percentage": 0, "nextResetTime": 1788073095998,
      "usageDetails": [ { "modelCode": "search-prime", "usage": 0 }, { "modelCode": "web-reader", "usage": 0 }, { "modelCode": "zread", "usage": 0 } ] },
    { "type": "TOKENS_LIMIT", "unit": 3, "number": 5, "percentage": 100, "nextResetTime": 1787056863927 },
    { "type": "TOKENS_LIMIT", "unit": 6, "number": 1, "percentage": 20, "nextResetTime": 1787641095989 }
  ], "level": "max" }
```

**Newer protocol (`level: pro`)** — `CREDIT_LIMIT` token rows only, no monthly MCP row:

```json
{ "limits": [
[...]

PR #2051, opened 2026-08-18
fix(zai): parse GLM quota limits array instead of flat percent fields

x3M3x:
## Summary

Replace the heuristic field-name probing in fetchZaiQuota with a structured parser that reads the data.limits array from the Z.AI /api/monitor/usage/quota/limit endpoint, following the same logic as OmniRoute getGlmUsage.

The old code looked for flat percent fields like fiveHourPercent, weeklyPercent, and monthlyPercent on the response body, fields the current Z.AI GLM Coding Plan API does not expose. The new code parses the limits array entries by type (TOKENS_LIMIT / TIME_LIMIT), unit, and number to correctly resolve the 5-hour session window, weekly window, and monthly MCP usage window.

## Verification

- [x] TypeScript typecheck passes on the changed file

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Z.AI quota reporting by accurately interpreting token and time limits across five-hour, weekly, and monthly usage windows.
  * Added safeguards for duplicate quota entries and empty limit responses.
  * Improved calculation of remaining usage for time-based limits.

### DIFF
--- a/src/providers/quota.ts
+++ b/src/providers/quota.ts
@@ -670,6 +670,7 @@ async function fetchClineQuota(provider: string, config: OcxProviderConfig): Pro
 /**
  * Z.AI GLM Coding Plan `GET /api/monitor/usage/quota/limit` — the coding-plan
  * subscription's 5-hour token cycle, weekly quota, and monthly MCP usage.
+ * Parses the `data.limits` array using the same logic as OmniRoute's getGlmUsage.
  * Authenticates with the API key as a Bearer token per Z.AI's API reference.
  */
 async function fetchZaiQuota(provider: string, config: OcxProviderConfig): Promise<ProviderQuotaProbeResult> {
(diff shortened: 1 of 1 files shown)


Ingwannu:
Thanks for the contribution. I am closing this draft as superseded by #2028 rather than maintaining two competing Z.AI quota parsers.

The current head is also not safe to merge independently:

- it changes provider behavior without a regression test and is hygiene-blocked;
- duplicate `unit=3, number=5` token rows are classified by encounter order, so a second five-hour row can become the weekly window without evidence;
- the later `unit=3, number=5` branch is unreachable because the earlier condition already consumes it;
- the TIME_LIMIT calculation rounds the provider percentage and mixes `total`, `remaining`, and `percentage` semantics without a sanitized real payload contract;
- it drops legacy payload compatibility entirely when `limits` is absent.

#2028 already has focused parsing tests, exact window-length matching, legacy fallback only when `limits` is absent, official-host credential pinning for both regions, and maintainer feedback requesting the remaining upstream evidence. Please move any verified payload evidence or narrowly better test case to #2028 instead of reopening this duplicate implementation.

Issue #1686, opened 2026-08-14
[Bug]: restore Codex env_key bearer admission and direct-mode OpenAI credential substitution


rrmlima:
### Summary

`/v1/responses` rejects requests authenticated with `Authorization: Bearer <token>` even though that is exactly what the Codex CLI sends when a provider is injected with `env_key` (the runtime contract since codex-cli 0.146). `resolveResponsesApiAuth` only reads `x-opencodex-api-key` ("Dedicated header ONLY"), so any user who injects the proxy with `env_key` — the modern runtime's preferred form — gets `401 authentication_error: opencodex API key required` on **every** model, including native OpenAI passthrough.

This is inconsistent with the rest of the proxy: `resolveApiAuth` (the non-responses path) already accepts `authorization` as a fallback, and the `grok` injector already emits `env_key`.

This issue also owns the directly coupled failure previously tracked in #1687: after an `env_key` admission bearer is accepted, `codexAccountMode: "direct"` must replace the proxy admission credential with the stored main ChatGPT/OpenAI credential before forwarding upstream. The two failures are one end-to-end auth contract and must be fixed and tested together.

### Reproduction

```bash
# 1. Inject the proxy with env_key (codex-cli 0.146+ contract)
# [model_providers.opencodex]
# base_url = "http://127.0.0.1:10100/v1"
# env_key = "OPENCODEX_API_AUTH_TOKEN"

# 2. Start the proxy with OPENCODEX_API_AUTH_TOKEN exported, then request:
curl -X POST http://127.0.0.1:10100/v1/responses \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $OPENCODEX_API_AUTH_TOKEN" \
  -d '{"model":"gpt-5.6-luna","input":[{"role":"user","content":[{"type":"input_text","text":"hi"}]}],"stream":true,"store":false}'
```

### Logs and screenshots

Admission failure:

```json
{"error":{"message":"opencodex API key required","type":"authentication_error","code":"invalid_api_key"}}
```

The same request with `x-opencodex-api-key: $OPENCODEX_API_AUTH_TOKEN` succeeds (200), proving the token is valid — only the header is rejected.

After bearer admission is accepted, direct-mode forwarding can instead fail with the second half of this contract:

```json
{"error":{"message":"OpenCodex admission credentials cannot be forwarded upstream","type":"invalid_request_error","code":"invalid_request_error"}}
```

### Area

[...]

PR #2076, opened 2026-08-19
fix(codex): emit modern env_key admission line on non-loopback inject

### LINKED ISSUE #2073: [Bug]: codex injector still emits legacy env_http_headers; runtime 0.146+ contract is env_key
### Summary

`src/codex/inject.ts` still emits the legacy `env_http_headers = { "x-opencodex-api-key" = "OPENCODEX_API_AUTH_TOKEN" }` line when injecting the opencodex provider into `$CODEX_HOME/config.toml`. Since codex-cli **0.146** the runtime contract for custom model providers is `env_key = "<VAR>"` (the proxy must read the API key from that environment variable), and `env_http_headers` with a dedicated `x-opencodex-api-key` header is no longer the preferred transport. 

This is the **generator side** of the auth contract already fixed server-side in #1686: #1686 made `/v1/responses` accept an `Authorization: Bearer` admission token (what `env_key` produces), but the injector still writes the old header form, so fresh installs do not benefit from the modern contract.

Concretely, the divergence is visible today:

| Source | Form emitted |
|---|---|
| `src/codex/inject.ts` (current main) | `env_http_headers = { "x-opencodex-api-key" = "OPENCODEX_API_AUTH_TOKEN" }` |
| `src/grok/*` injector (per #1686 body) | `env_key = "..."` |
| codex-cli 0.146+ documented contract | `env_key = "<VAR>"` |

### Reproduction

```bash
# 1. Install opencodex, start the proxy, run any sync that injects the provider:
ocx start   # or: ocx sync

# 2. Inspect the generated provider block:
grep -n "env_key\|env_http_headers" "$CODEX_HOME/config.toml"
# → env_http_headers = { "x-opencodex-api-key" = "OPENCODEX_API_AUTH_TOKEN" }
#   (legacy form — the modern runtime prefers env_key)

# 3. Run the CLI through the proxy:
codex exec --skip-git-repo-check -m cmd/deepseek/deepseek-v4-flash "responda apenas: OK"
# → works only because the operator manually edited config.toml to env_key,
#   or because an older injector left env_key behind.
```

### Root cause

`src/codex/inject.ts` builds the provider TOML block:

```ts
if (includeApiAuthHeader) {
  lines.push(
    'env_http_headers = { "x-opencodex-api-key" = "OPENCODEX_API_AUTH_TOKEN" }',
  );
}
```

This string predates the codex-cli 0.146 `env_key` contract. The server-side admission (fixed in #1686) reads `Authorization: Bearer` and `x-opencodex-api-key`; the header form still works when the runtime forwards it, but the modern, documented, and already-server-supported form is `env_key`, and the two injectors are inconsistent.

### Proposed diff

```diff
--- a/src/codex/inject.ts
+++ b/src/codex/inject.ts
@@ -228,7 +228,7 @@ export function buildOcxProviderBlock(
   if (includeApiAuthHeader) {
     lines.push(
-      'env_http_headers = { "x-opencodex-api-key" = "OPENCODEX_API_AUTH_TOKEN" }',
+      'env_key = "OPENCODEX_API_AUTH_TOKEN"',
     );
   }
   if (supportsWebsockets) lines.push("supports_websockets = true");
```

### Alternatives considered

1. **Keep `env_http_headers` and rely on `x-opencodex-api-key` admission** — works today, but keeps two incompatible auth transports for the same proxy, and the runtime may drop `env_http_headers` support in a future release (the header name is not part of the documented 0.146+ provider contract).
2. **Emit both lines** — redundant; the proxy treats either as admission, and the extra header is noise.
3. **Make it configurable** (`injectAuthForm = "env_key" | "env_http_headers"`) — unnecessary until a runtime actually rejects one of them; the modern contract is `env_key`.

### Test matrix

| Case | Expected |
|---|---|
| Fresh `ocx start` / `ocx sync` on codex-cli 0.146+ | `config.toml` contains `env_key = "OPENCODEX_API_AUTH_TOKEN"` |
| `codex exec -m <routed-model>` after injection | 200, no 401 |
| `/v1/responses` with `Authorization: Bearer $OPENCODEX_API_AUTH_TOKEN` | 200 (server side already accepts — #1686) |
| Upgrade path: existing config with `env_http_headers` | proxy still admits via `x-opencodex-api-key` (back-compat unchanged) |
| Claude Code / Grok injectors | unchanged (`env_key` already emitted there) |

lidge-jun:
## Summary

codex-cli 0.146+ documents `env_key` as the bearer auth form for custom model providers: the runtime reads the named env var and sends `Authorization: Bearer $VAR`, hard-erroring on a missing/empty variable instead of silently omitting the header (verified against openai/codex `rust-v0.146.0`: `model-provider-info/src/lib.rs#L263-281`, `model-provider/src/provider.rs#L153-174`). The server side already admits our admission secret as a bearer and substitutes stored main auth upstream (#1686 chain: 22d5492b2 / acfedae0a / f848b4997) — the injector was the stale half, still emitting the legacy `env_http_headers` table (#2073).

- `buildProviderTableBlock` (non-loopback/legacy mode only) now emits `env_key = "OPENCODEX_API_AUTH_TOKEN"`; `requires_openai_auth = true` stays (env_key wins wire auth, the flag keeps login/account UX).
- Loopback Design B still emits no auth line at all; the loopback tests now pin `env_key` absence too.
- The #2061 orphan sub-table cleanup keeps handling historic `env_http_headers` configs; re-injection replaces the marker-owned block, so existing installs converge on next `ocx start`/sync.
- SoT (`structure/02`) and the 9 locale `codex-integration` guides updated to the modern line.
- Design doc: `devlog/_plan/260819_triage_execution/010_env_key_contract.md`.

Closes #2073

## Verification

- `bun test tests/codex-inject.test.ts tests/loopback-listener-admission.test.ts tests/loopback-listener-integration.test.ts tests/codex-inject-integration.test.ts tests/codex-inject-write-lock.test.ts` — 105 pass / 0 fail
- `bun test tests/codex-envkey-admission-substitution.test.ts tests/codex-composed-acceptance.test.ts tests/openai-provider-option-tooling.test.ts` — 20 pass / 0 fail

## Checklist

- [x] Tests updated (env_key emission + legacy-line absence + loopback no-auth-line regressions)
- [x] Docs updated (structure/02 + 9 locale guides)
- [x] No unrelated changes

## Summary by CodeRabbit

- **New Features**
  - Updated non-loopback Codex provider authentication to use the `OPENCODEX_API_AUTH_TOKEN` environment variable through the modern `env_key` setting.
  - Preserved loopback provider behavior without environment-based authentication settings.

- **Documentation**
  - Updated Codex integration guidance across supported languages and configuration references to reflect the new authentication method.

- **Tests**
  - Expanded coverage for authentication migration, legacy configuration removal, token preservation, and loopback provider behavior.


lidge-jun:
Adversarial-review fold-back on the two findings:

**Medium (missing non-loopback upgrade regression) — accepted and fixed.** Added an integration test that injects over a legacy `env_http_headers` non-loopback config and pins the full conversion to `env_key` with exactly one provider block surviving (tests/codex-inject-integration.test.ts).

**High (apiKeys-only non-loopback breaks) — rebutted as not a regression.** Under the legacy line, a config with `config.apiKeys` set but no `OPENCODEX_API_AUTH_TOKEN` in Codex's environment never worked either: codex omits the `env_http_headers` entry when the variable is empty, the request arrives with no credential, and the server 401s (`resolveApiAuth` needs a presented secret; configured keys are matched only when presented — src/server/auth-cors.ts:338-350). The only working shapes were (a) env var set, or (b) the user hand-editing config.toml — both unchanged by this PR. With `env_key` the same broken shape now fails at the codex side with a *named* missing-variable error instead of a silent 401, and the workaround is identical and still works: point `OPENCODEX_API_AUTH_TOKEN` at any configured apiKey value (bearer admission matches configured keys, auth-cors.ts:346-348). Net behavior change for that population: better diagnosability, no lost functionality.

PR #2078, opened 2026-08-19
fix(replay): scope durable thought signatures per credential and bound persist visibility

### LINKED ISSUE #1926: Gemini thought-signature replay omits destination/credential scope and emits tool items before the durable write commits
### Client or integration

Codex App

Google Gemini / Antigravity (Cloud Code Assist)

### OpenCodex version

2.24.2 (present on current `dev`)

### Endpoint or capability

`/v1/responses` tool-call replay — `src/responses/thought-signature-replay.ts`

### Current behaviour

Two gaps remain in the durable thought-signature store that #1823 introduced. Both are about the *scope* of a stored signature and the *durability* of the store, not about the feature working in the happy path.

**1. The replay key omits destination and credential identity.**

`keyFor()` derives from five fields: `[clientThreadId, providerName, adapterName, modelId, callId]`.

The sibling in-memory cache `src/responses/reasoning-replay-cache.ts` uses seven for the same kind of value — it also includes `providerDestinationIdentity` and `credentialIdentity`. Both fields exist on the type (`src/types.ts`) and are populated in `src/server/responses/core.ts`. So for the durable store, one thread + provider name + model can currently share an opaque signature across:

- two different upstream endpoints under one provider name,
- two different accounts,
- before and after a credential rotation.

**Why the fields were left out is a real constraint, not an oversight.** The comment in `thought-signature-replay.ts` says so: the reasoning cache's identities are process-local HMACs (`replayIdentityKey = randomBytes(32)`), so reusing them verbatim in a durable key would change every key on restart and the store would silently stop working — worse than the current over-broad key, because it would fail closed and invisibly.

A fix therefore needs *restart-stable* identity material:

| Auth mode | Material | Restart-stable? |
|---|---|---|
| OAuth | `accountId` + `generation` | yes, and already non-secret — usable directly |
| Key | derived from `provider.apiKey` | value is stable but is raw secret material; needs a persisted-salt digest |
| `local` | `credentialIdentity` is `undefined` | needs an explicit policy — `keyFor`'s guard is all-or-nothing, so a newly-required field would make local providers stop remembering entirely |

And because `keyFor()`'s output **is** the on-disk key (the snapshot is written with `version: 2`), widening it invalidates every persisted entry. That needs a version bump with an explicit drop-or-migrate decision; dropping is acceptable, silently mismatching is not.

**2. Tool items are emitted before the durable write commits, and a persist failure is swallowed.**

`rememberThoughtSignatureForReplay()` deliberately returns `{ result, durable }` so a caller can await durability before exposing the item — the comment says exactly that. No caller does. In `src/bridge.ts` the promise is discarded:

```ts
void rememberExtraContentForReplay(currentToolCall.callId, currentToolCall.providerMetadata, replayCacheScope);
// ...
...(rememberAndSerializeExtraContent(...).extra ?? {}),   // durable dropped
emit("response.output_item.done", { output_index: currentToolCall.outputIndex, item });
```

in the streaming close path (both the freeform and function-call branches) and again on the non-streaming `pushOutput` path.

This one has a structural obstacle worth stating up front: `closeCurrentToolCall` is a **synchronous** closure that writes into a `ReadableStream` controller. There is no `await` available at that point, so "await the durable promise before emit" is not a one-line change — it needs either an async close path or a pre-emit barrier.

Separately, `persist()` ends in `.catch(() => {})`, so a write failure is invisible even to a caller that *does* await.

### Expected behaviour

- A signature stored under one account/endpoint is never served to another, and a credential rotation does not replay pre-rotation signatures.
- A signature that was handed to a client is either durably recorded or the failure is observable — not silently best-effort.
- Restart replay keeps working, which is the whole point of the durable store.

### Suggested implementation notes

1. Derive restart-stable destination and credential identities for this store (not the process-local HMACs), decide the `authMode: "local"` policy explicitly, and bump the snapshot to `version: 3` with a stated drop-or-migrate choice.
2. Give the bridge a pre-emit barrier for the durable write, or make the close path async.
3. Replace the swallowed `catch` with a typed failure that a caller can act on.

Regression coverage should include: cross-account isolation, cross-destination isolation, rotation isolation, same-scoped-key/different-signature fail-closed (already present), write-then-simulated-restart hit (the case a naive fix would break), and a `version: 2` file producing no stale-key hit under `version: 3`.

### Additional context

Found while auditing #1823's follow-ups. The conflict detection, byte budget, and thread/provider/adapter/model scoping that #1823 shipped all hold up; these are the two remaining gaps.

### Checks

- [x] The expected behaviour is based on an upstream specification or concrete client requirement.

lidge-jun:
## Summary

Implements the remaining half of #1926 per the campaign design (devlog/_fin/260818_bug_pr_resolution/051_tsig_credential_scope.md), amended by this cycle's C4 security plan audit (3 High blockers folded).

**Gap 1 — credential scope in the durable replay key.** The durable thought-signature store keyed entries by thread + destination + adapter/model but not by credential, so account A's Gemini signatures could replay under account B on the same destination. Keys now include `credentialDurableIdentity`: a salted-HMAC (installation-local salt persisted beside the store with mode 0600, full 256-bit output — audit rejected the design's original truncated unsalted digest as an offline key verifier) over the persisted OAuth account-slot id (rotation-safe, verified stable across refresh), the provider API key, or the Codex account handle, plus credential-scoped header overrides. A scope that cannot produce a durable credential identity **fails closed** (no durable store/lookup) instead of sharing a `credential:unknown` slot (audit blocker 2). `STORE_VERSION` 3→4 drops v3 rows on load — they carry no credential info and are not upgradable; signatures re-accumulate per turn (bounded, pre-store status quo).

**Gap 2 — bounded persist visibility.** Terminal frames could become externally visible before the queued signature persist settled. All async terminal paths (completed, truncation/adapter-EOF incomplete, failed) and both buffered JSON returns now await `awaitThoughtSignatureDurability()` — a 250ms-capped race on the store's persist chain (bounded best effort, not a guarantee; audit wording fix). The sync stall-timeout kill path keeps best-effort behavior, documented in place.

Closes #1926

## Verification

- New: `tests/thought-signature-credential-scope.test.ts` — cross-credential isolation, fail-closed unscoped store/lookup, v3-drop + v4-reload, salt stability/width, barrier persistence. 7 pass.
- `bun test` over 12 replay/bridge/auth suites (thought-signature, roundtrip, vertex, anthropic, bridge, reasoning-replay x4, summary-passthrough, server-auth): 231 pass / 0 fail.
- `bun x tsc --noEmit` clean.

## Checklist

- [x] Regression tests for both gaps
- [x] Fail-closed on missing durable identity (no shared slots)
- [x] No secrets persisted (salted HMAC only; salt holds no credential material)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of reasoning and thought-signature delivery during completed, failed, truncated, and interrupted responses.
  * Prevented replay data from being reused across different credentials or after unsupported credential changes.
  * Improved recovery of thought signatures across restarts and supported request flows.
  * Ensured buffered responses wait briefly for required persistence before becoming visible.
* **Tests**
  * Added coverage for credential isolation, restart persistence, invalid credentials, legacy data invalidation, and durability timing.

### DIFF
--- a/devlog/_plan/260819_triage_execution/020_remote_reasoning_leak_rca.md
+++ b/devlog/_plan/260819_triage_execution/020_remote_reasoning_leak_rca.md
@@ -1,0 +1,25 @@
+# 020 — #2064 RCA: Remote raw-thinking leak (FIXED-ON-DEV)
+
+Reported on 2.24.2: Codex Remote paints raw Grok thinking live
+(response.reasoning_text.delta), then swaps to the progress line leaving
+italic fragments; stored items are summary:[] + content reasoning_text.
+
+RCA (sol lane + main verification, 2026-08-19):
+
+- v2.24.2 bridge emitted the raw channel: git show v2.24.2:src/bridge.ts has
+  response.reasoning_text.delta; fix commit 56752d7c5 (#2007, landed via
+  PR #2016 merge 891c8284b) is NOT an ancestor of v2.24.2, IS on dev.
+- Current dev has no escape path for openai-chat routed models: the bridge
+  emits summary-channel deltas and summary-shaped items only
+  (src/bridge.ts:987,1016,591-606); the native-Responses rewrite covers WS
+  upstream, eager relay, HTTP SSE tee, and JSON reframing legs
+  (src/server/responses/core.ts:2835-3066,
+  src/server/responses-reasoning-summary-rewrite.ts:51-110).
+- Suites: tests/responses-reasoning-summary-rewrite.test.ts +
+  tests/bridge.test.ts = 73 pass / 0 fail (fresh).
+
+Outcome: no new code. Issue #2064 closed as fixed-on-dev with a note asking
+for on-device Remote verification on the next release build. Model-side
+intermittent reasoning exposure (user caution) stays out of scope — our relay
+provably converts the channel on every leg.
+
--- a/src/adapters/google.ts
+++ b/src/adapters/google.ts
@@ -26,6 +26,7 @@ import { identifyRoutedModel } from "./identity";
 import { antigravityUsesReplayCache, applyAntigravityReplay, clearAntigravityReplay, observeAntigravityReplay } from "./google-antigravity-replay";
 import { resolveAntigravityEffortWireModel } from "../providers/antigravity-models";
 import { googleVertexLocationConfigError } from "../providers/google-vertex-location";
+import { lookupReplayThoughtSignature } from "../responses/thought-signature-replay";
 import {
   isTranslatorBudgetExceededError,
   retainTranslatedEventBatch,
@@ -215,7 +216,13 @@ function messagesToGeminiFormat(
             const part: Record<string, unknown> = { functionCall };
             // Prefer the metadata that travelled with this exact call; fall back to the legacy
             // field for callers that have not been migrated. Never merge or synthesize.
-            const signature = tc.providerMetadata?.google?.thoughtSignature ?? tc.thoughtSignature;
+            // Final fallback (#1926): the durable store, read AT SERIALIZATION TIME. The
+            // Responses parser runs before the route/credential scope is bound, so its
+            // parse-time lookup can never hit; by the time this adapter serializes, the
+            // credential-scoped identity is bound and the durable lookup is meaningful.
+            const signature = tc.providerMetadata?.google?.thoughtSignature
+              ?? tc.thoughtSignature
+              ?? lookupReplayThoughtSignature(tc.id, parsed._reasoningReplayScope);
             if (isLikelyRealThoughtSignature(signature)) part.thoughtSignature = signature;
             parts.push(part);
           }
--- a/src/bridge.ts
+++ b/src/bridge.ts
@@ -15,6 +15,7 @@ import { rememberReasoningForCall } from "./responses/reasoning-replay-cache";
 import {
   rememberAndSerializeExtraContent,
   rememberExtraContentForReplay,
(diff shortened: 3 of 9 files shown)


lidge-jun:
Security-review fold-back (9697e89ff):

**Blocker 2 (client-controlled bucket) — accepted, fixed.** The codex-forward durable handle now uses trusted pool-context account ids only; the client-supplied `chatgpt-account-id` header no longer participates, and direct-forward turns get no durable scope (fail closed, in-process cache still covers same-process replay).

**Blocker 3 (runtime lookup not wired) — accepted, fixed.** This was pre-existing (the v3 store had the same parse-before-scope ordering), but it made the store write-only at runtime, so it's fixed here: the google adapter now falls back to `lookupReplayThoughtSignature` at serialization time, when the credential-scoped identity is bound. Regression test added that parses scope-less, binds identity after (as the server does), and asserts the signature reaches the wire part.

**Blocker 1 (store+salt = offline key verifier) — partially accepted, remainder rebutted with context.** Hardened: the salt file re-asserts `0600` on every load. On colocation: an attacker who can read `~/.opencodex/` already has `config.json`, which stores provider API keys in **plaintext** (tracked as #1221) — the HMAC digest is strictly weaker material than what sits next to it, so keychain-grade separation for the salt would protect a digest of a key whose cleartext is in the same directory. When #1221 moves keys to the OS keychain, the salt should move with them; noted as a dependency there rather than blocking this isolation fix, whose alternative (no credential scoping at all) is the worse security posture.

PR #2079, opened 2026-08-19
fix(update): stage, verify, and swap npm self-updates with rollback

### LINKED ISSUE #1849: [Bug][Windows] Failed ocx update leaves a file-less package skeleton; service wrapper restart-loops 1,009 times with no integrity check or backoff

### LINKED ISSUE #1942: [Bug][Windows] ocx update is not transactional: a failed update deletes the working install with no rollback
### Client or integration

CLI (`ocx update`), npm global install

### Area

Platform (Windows) / CLI

### Summary

Split out of #1849, which reported two distinct defects that need different fixes and
different closure conditions. The service-restart half shipped; this is the root cause,
and it is still open.

| Half | Defect | Status |
|---|---|---|
| A | Service wrapper restart-loops against a missing install (1,009 × `child exited with code 3`) | Fixed by #1877, released in **v2.24.0** |
| B | `ocx update` can destroy a working install and leave an unbootable skeleton | **this issue** |

#1877 stopped the *symptom*: the wrapper no longer hammers a package that is not there.
It does nothing about how the package came to be missing.

### Current behaviour

The reporter's failed update left the global package directory with **107 directories and
zero files** — no `package.json`, no `bin/ocx.mjs`, no bundled `bun.exe`, and every
launcher (`ocx`, `ocx.cmd`, `ocx.ps1`, and the `opencodex` siblings) removed from
`%APPDATA%\npm`. npm created the tree, unpacked no content, and did not roll back. The
documented recovery commands were themselves unreachable, because the launchers that would
run them were the files that had been deleted.

The update path delegates to `npm install -g`, which is destructive-in-place: it removes
the existing install before the replacement is verified. When it fails midway there is no
prior state left to fall back to.

Verification does exist and is not the gap — `checkUpdatePackageIntegrity` fails closed
before the proxy is stopped, `isRealBunBinary` rejects the postinstall stub, and
`awaitRestartedProxyHealthy` runs an identity-checked `/healthz` probe with a version match
and a stability window. The problem is narrower and worse than missing verification: it is
**detect-only, and it runs after the old install is already gone**, so a failed probe has
nothing left to restore.

### Expected behaviour

An update that fails leaves the previous working install in place. The user should never
reach a state where the tool cannot repair itself.

### Suggested implementation notes

A transactional shape, where the existing install is only released after the replacement
is proven:

```
download to staging
  → verify executable/assets/version in staging
  → snapshot the existing install
  → atomic directory or launcher switch
  → health probe the new install
  → success: drop the snapshot   |   failure: restore the snapshot
```

The ordering constraint is the whole point: **never delete the existing install before the
replacement is verified.** "download → delete → copy" cannot be made safe by adding retries
to it.

Worth deciding explicitly, since it shapes the fix: whether to keep delegating to
`npm install -g` (and stage around it, e.g. install to a temp prefix and switch), or to own
the download and unpack directly. The former is less invasive; the latter is the only way to
get a truly atomic switch on Windows, where an in-use file cannot always be replaced.

### Verification this needs

Fault injection at every boundary, each asserting a *launchable* install survives:

- after download, before verification
- after verification, before the snapshot
- mid-switch (the Windows case where a partially replaced directory is the failure mode)
- after the switch, before the health probe
- health probe fails → snapshot restored and the old version still runs

`tests/update-transaction.test.ts` does not exist yet and would be new.

### Additional context

Original report, environment details and the full 1,009-restart evidence are in #1849.
That issue stays open for this defect; #1877 is linked there as the mitigation half.

lidge-jun:
## Summary

Implements the unbuilt half of the Windows stability program design (devlog/_plan/260817_windows_stability_program/090_transactional_update_rollback.md) for #1942 / #1849: the npm self-update no longer installs into the live global tree.

**Stage → verify → swap → rollback** (new `src/update/transactional-install.mjs`, launcher-usable ESM matching the existing `npm-cache-preflight.mjs` convention):

- **Stage**: `npm install --prefix <sibling>/.ocx-staging-<ts>` — live tree and bin shims untouched.
- **Verify inside the stage** (manifest: package.json parses + version matches, `bin/ocx.mjs` present and non-truncated, sentinel deps intact). Failure deletes the stage; live never touched — this alone closes the #1849 empty-install class.
- **Swap**: live → sibling `.ocx-backup-<ts>`, stage → live, re-verify live. Failure reverse-renames the backup back; a double fault writes `.ocx-recovery.json` with a one-line restore instruction.
- **Boot probe** in the launcher: a broken live tree with a backup sibling is auto-restored before anything runs (power loss mid-swap); stale backups are reaped once live verifies healthy.
- Fallback: if the transactional machinery itself cannot run (exotic layout), the legacy in-place install still executes with a warning.

Design fix from the plan audit: staging/backup are **siblings** of the package dir, never children — the original 090 layout put them inside `<prefix>`, which the live rename would carry along (and live cannot move into its own subtree).

Closes #1942
Closes #1849

## Verification

- `bun test tests/update-transactional.test.ts` — 9 pass / 0 fail: manifest red/green, every D4 fault row (stage fail, verify fail, wrong version, locked-swap rollback, double-fault marker, power-loss boot restore, stale-backup reap) with the live-tree invariant asserted (always old-complete or new-complete).
- `node --check` on launcher + module; `node bin/ocx.mjs update --help` side-effect-free short-circuit intact.
- `bun x tsc --noEmit` clean (d.mts declaration included).

## Checklist

- [x] Fault-injection tests per design failure-mode table
- [x] No new dependencies
- [x] Launcher fallback preserves the legacy path when transactional cannot run

## Summary by CodeRabbit

* **New Features**
  * Self-updates now install in a staged environment and verify files before activation.
  * Successful updates clean up temporary backup data.
  * Interrupted updates can be restored automatically when the application or Windows service starts.

* **Bug Fixes**
  * Failed updates automatically roll back when possible.
  * Improved handling of incomplete, corrupted, or failed updates to prevent launching with a broken installation.

### DIFF
--- a/bin/ocx.mjs
+++ b/bin/ocx.mjs
@@ -304,15 +304,13 @@ function runNpmSelfUpdate() {
       res = { status: 1 };
     }
   } catch (error) {
-    // Transactional machinery itself failed (e.g. exotic install layout): legacy path.
-    console.warn(`opencodex: transactional update unavailable (${error?.message ?? error}); falling back to in-place npm install.`);
-    console.log(`$ npm install -g ${PKG}@${tag}`);
-    res = spawnSync(installInvocation.file, installInvocation.args, {
-      stdio: "inherit",
-      timeout: 180000,
-      windowsHide: true,
-      ...installInvocation.options,
-    });
+    // An unexpected throw means we cannot prove the live tree is untouched, so the
+    // legacy in-place install (which deletes live first) is exactly the wrong rescue —
+    // it recreates the #1849 destruction path. Report and stop; the boot probe and the
+    // recovery marker cover the swap-window states.
+    console.error(`opencodex: transactional update failed unexpectedly (${error?.message ?? error}). ` +
+      "The live install was not knowingly modified; run 'ocx update' again or reinstall with npm install -g.");
+    res = { status: 1 };
   }
   if (res.status === 0) {
     console.log(`\nUpdated${latest ? ` to v${latest}` : ""}.`);
--- a/bin/ocx.mjs
+++ b/bin/ocx.mjs
@@ -22,6 +22,7 @@ import {
   runNpmCachePreflight,
 } from "../src/update/npm-cache-preflight.mjs";
 import { handoffWindowsTrayForUpdate, planWindowsTrayUpdate } from "../src/update/tray-update-plan.mjs";
+import { bootRestoreProbe, transactionalNpmUpdate } from "../src/update/transactional-install.mjs";
 
 const PKG = "@bitkyc08/opencodex";
 const require = createRequire(import.meta.url);
@@ -267,13 +268,52 @@ function runNpmSelfUpdate() {
     }
   }
 
-  console.log(`Updating${latest ? ` to v${latest}` : ""}...\n$ npm install -g ${PKG}@${tag}`);
-  const res = spawnSync(installInvocation.file, installInvocation.args, {
-    stdio: "inherit",
-    timeout: 180000,
-    windowsHide: true,
-    ...installInvocation.options,
-  });
+  // #1942/#1849: stage -> verify -> swap -> rollback instead of installing straight
+  // into the live tree. A failure at any point leaves either the old or the new tree
+  // complete — never a file-less skeleton. Falls back to the legacy in-place install
+  // only when the transactional module cannot run at all.
+  const packageDir = resolve(here, "..");
+  console.log(`Updating${latest ? ` to v${latest}` : ""} (transactional)...`);
+  let res;
+  try {
+    const tx = transactionalNpmUpdate({
+      packageDir,
+      pkgName: PKG,
+      targetVersion: latest || undefined,
+      tag,
+      runNpm: (args) => {
+        const invocation = npmInvocation(args);
+        if (!invocation) return { status: 1 };
+        return spawnSync(invocation.file, invocation.args, {
+          stdio: "inherit",
+          timeout: 180000,
+          windowsHide: true,
+          ...invocation.options,
+        });
+      },
+      log: (line) => console.log(line),
+    });
+    if (tx.ok) {
+      res = { status: 0 };
+    } else if (tx.phase === "stage" || tx.phase === "verify") {
+      // Live tree untouched: report and stop. Nothing to roll back.
+      console.error(`opencodex: update aborted before touching the live install (${tx.phase}): ${tx.error}`);
+      res = { status: 1 };
+    } else {
+      console.error(`opencodex: update failed (${tx.phase}): ${tx.error}${tx.rolledBack ? " — previous version restored." : ""}`);
+      res = { status: 1 };
+    }
+  } catch (error) {
+    // Transactional machinery itself failed (e.g. exotic install layout): legacy path.
+    console.warn(`opencodex: transactional update unavailable (${error?.message ?? error}); falling back to in-place npm install.`);
+    console.log(`$ npm install -g ${PKG}@${tag}`);
+    res = spawnSync(installInvocation.file, installInvocation.args, {
+      stdio: "inherit",
+      timeout: 180000,
+      windowsHide: true,
+      ...installInvocation.options,
+    });
+  }
   if (res.status === 0) {
     console.log(`\nUpdated${latest ? ` to v${latest}` : ""}.`);
     repairCodexShimIfNeeded();
@@ -447,6 +487,20 @@ if (updateHelpRequested) {
 
 if (process.argv[2] === "update" && isNodeModulesInstall() && !isBunGlobalInstall()) {
   runNpmSelfUpdate();
+}
+
+// #1849 boot probe: a prior update that lost power (or double-faulted) mid-swap leaves a
+// backup sibling and a broken live tree. Restore before anything tries to run from the
+// broken tree; reap stale backups once the live tree verifies healthy.
+if (isNodeModulesInstall() && !isBunGlobalInstall()) {
+  try {
+    const probe = bootRestoreProbe(resolve(here, ".."));
+    if (probe.action === "restored") {
+      console.warn(`opencodex: previous update left a broken install — restored the backup from ${probe.from}.`);
+    } else if (probe.action === "failed") {
+      console.warn(`opencodex: a backup from a failed update exists but could not be restored automatically: ${probe.error}`);
+    }
+  } catch { /* the probe must never block launch */ }
 }
 
 const bunRuntime = resolveBun();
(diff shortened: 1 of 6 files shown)


lidge-jun:
Adversarial-review fold-back (9f406cfa9):

**High 1 (boot probe unreachable when live is gone) — accepted, fixed.** The Windows service wrapper script — which lives outside the package tree and survives the swap — now calls a `:restore_backup` subroutine before either "installation is incomplete" hard stop: it finds the newest `.ocx-backup-*` sibling, restores it over the missing/broken package dir, and re-checks. That covers exactly the power-loss window where the in-launcher probe cannot run (design 090 D5's service-side probe). The in-launcher probe still handles the broken-but-startable class and backup reaping.

**High 2 (stale launcher after swap) — rebutted with evidence.** `launcher` is `fileURLToPath(import.meta.url)` — an absolute path *string* inside `packageDir`. The swap renames trees but the path identity of `packageDir` never changes, so post-swap spawns through `launcher` execute the **new** tree's file at the same path (the old tree now lives under `.ocx-backup-*`, which nothing references). The currently-running process keeps its already-loaded old code, which is identical to the legacy `npm install -g` behavior.

**High 3 (weak verification reaps the only backup) — accepted, fixed.** The manifest now size-gates the actual bundled Bun binary (largest file under `node_modules/bun`, >= 10MB), so a Bun-less tree fails verification and the boot probe restores instead of reaping.

**High 4 (broad fallback recreates the destruction path) — accepted, fixed.** Staging/backup `mkdir` failures return phase errors with live untouched; an unexpected throw now reports and stops (the legacy in-place install is never used as a rescue). The legacy path remains only as explicit dead code for non-transactional layouts — and that branch is now unreachable from the throw handler.

**Medium (lock retry) — accepted, fixed.** `renameWithRetry` gives the swap renames a bounded EPERM/EBUSY/EACCES retry (5 attempts, stepped backoff, worst case ~1.5s).

Tests updated: Bun-binary fixture, wrapper-script pins (4 checks + restore subroutine). 133 pass across service + transactional suites.

Issue #1804, opened 2026-08-16
[Research]: AgentRouter content-blocked responses correlate with non-English first prompts


Wibias:
### Summary

This issue tracks only the **language/content-dependent** half split from #1689.

Reports indicate that `agentrouter.org` may return `HTTP 400` with `error.code: "content-blocked"` more often for benign Portuguese/Spanish first-user prompts than for equivalent English prompts.

This is separate from burst-sensitive failures. Request pacing, retry/backoff, and burst behavior remain tracked in #1689.

### Why this is research / needs-info

The currently proposed workaround is to prepend an English instruction frame to the user's first message before forwarding it upstream. That changes user content and therefore changes request semantics. We should not adopt prompt mutation merely because it appears to bypass a third-party WAF.

Before implementing any OpenCodex-side language workaround, we need evidence that distinguishes:

- deterministic content filtering from burst/rate state;
- language effects from specific blocked phrases;
- gateway behavior from model/provider behavior;
- a supported provider contract from behavior inferred from a third-party reference implementation.

### Evidence required

Please provide a controlled, redacted matrix against the same AgentRouter provider/account and current OpenCodex `dev`:

1. Equivalent benign prompts in English, Portuguese, and Spanish, sent without burst traffic.
2. Repeated runs after a sufficient idle period so prior WAF/burst state does not contaminate the comparison.
3. Exact HTTP status and sanitized error code/type for each run.
4. A control showing whether retrying the unchanged request after an idle period changes the result.
5. If available, authoritative AgentRouter documentation describing the language/content filter contract.

Do not include credentials, full request identifiers, or sensitive prompt content.

### Acceptance boundary

A fix should preserve user content unless there is an explicit, documented, opt-in compatibility policy with clear semantics.

A global or host-name-triggered hidden preamble is **not** an accepted implementation direction at this stage.

Potential outcomes after evidence:

- upstream/documentation-only guidance if this is intentional provider filtering;
- a provider-specific opt-in compatibility mode if a stable supported contract exists;
- no OpenCodex change if the only workaround is silently altering prompts.

### Relationship

- Split from #1689.
- #1689 remains the burst/pacing/retry issue.

lidge-jun:
[...]

PR #2082, opened 2026-08-19
fix(adapters): frame AgentRouter first messages to pass the language filter

### LINKED ISSUE #2074: [Bug]: AgentRouter 400 content-blocked on non-English first messages; add language preamble for AGR providers
### Summary

AgentRouter-backed providers (`AGR-OAI`, `AGR-CLA`, or any provider whose `baseUrl` contains `agentrouter`) reject non-English first user messages with HTTP 400 `content-blocked`. This is a **regional/language filter** applied by the AgentRouter gateway to the first user message content.

Affected users: anyone routing to AgentRouter with Portuguese, Spanish, or other non-English prompts — the 400 surfaces mid-session as a hard failure, and it correlates with the non-English first prompt (see #1804, which closed #1689 as research).

The fix is to prepend an explicit English instruction frame to the first user message when the provider is AgentRouter, so the boundary filter passes while preserving the original request and the model's output language.

### Reproduction

```bash
# 1. Configure an AgentRouter provider:
#    providers: { "AGR-CLA": { adapter: "anthropic", baseUrl: "https://agentrouter.org", ... } }

# 2. Send a Portuguese first message:
curl -X POST http://127.0.0.1:10100/v1/messages \
  -H "x-api-key: $OPENCODEX_API_AUTH_TOKEN" \
  -d '{"model":"claude-opus-4-8","messages":[{"role":"user","content":"responda apenas: OK"}]}'
# → 400 content-blocked (only when the FIRST user message is non-English)

# 3. Same request with an English first message:
# → 200
```

### Root cause

The Anthropic adapter builds messages with `messagesToAnthropicFormat(parsed, toolNames)` and sends them verbatim. AgentRouter runs a language/regional filter on the first user message content; a non-English first prompt trips it.

### Proposed diff

`src/adapters/anthropic.ts` — add a small idempotent helper and call it when the provider is AgentRouter:

```ts
const AGR_PREAMBLE_MARKER =
  "[Instruction: Process the user request below and respond in the appropriate language.]";

export function applyAgrLanguagePreamble(messages: unknown[]): void {
  const firstUser = messages.find(
    m => typeof m === "object" && m !== null && (m as { role?: string }).role === "user",
  ) as { content?: unknown } | undefined;
  if (!firstUser) return;

  if (typeof firstUser.content === "string") {
    if (!firstUser.content.includes(AGR_PREAMBLE_MARKER)) {
      firstUser.content = `${AGR_PREAMBLE_MARKER}\n\n${firstUser.content}`;
    }
  } else if (Array.isArray(firstUser.content)) {
    const textPart = firstUser.content.find(
      p => typeof p === "object" && p !== null && (p as { type?: string }).type === "text",
    ) as { text?: string } | undefined;
    if (textPart && typeof textPart.text === "string") {
      if (!textPart.text.includes(AGR_PREAMBLE_MARKER)) {
        textPart.text = `${AGR_PREAMBLE_MARKER}\n\n${textPart.text}`;
      }
    } else {
      firstUser.content.unshift({ type: "text", text: `${AGR_PREAMBLE_MARKER}\n\n` });
    }
  }
}
```

And at the message-build site:

```ts
const { system, messages } = messagesToAnthropicFormat(parsed, toolNames);
const isAgr =
  (typeof provider.name === "string" && provider.name.startsWith("AGR")) ||
  (typeof provider.baseUrl === "string" && provider.baseUrl.includes("agentrouter"));
if (isAgr) {
  applyAgrLanguagePreamble(messages);
}
```

Note: `provider.name` can be `undefined` at runtime for custom providers, so the reliable signal is the `baseUrl` containing `agentrouter` (or the configured name starting with `AGR`).

### Alternatives considered

1. **Do nothing** — non-English AgentRouter users keep hitting 400; workaround is forcing English first prompts, which degrades UX for PT/ES.
2. **Client-side only** — users prepend the frame manually in their prompts; fragile, not enforceable.
3. **Provider-level config flag** (e.g. `agrLanguagePreamble: true`) — more flexible but adds config surface for what is a single, safe, idempotent transform scoped to AgentRouter destinations.

### Test matrix

| Case | Expected |
|---|---|
| First user message in PT via AGR-CLA | 200, response in PT |
| First user message in ES via AGR-CLA | 200, response in ES |
| First user message in EN via AGR-CLA | 200, no double preamble (idempotent) |
| Non-AGR provider (anthropic direct) | Unchanged — no preamble |
| Multi-turn session (subsequent messages non-English) | Unchanged — only first message gets the frame |
| Content as text-part array (Claude Code format) | Frame prepended to the text part |
| Re-injection after upstream retry | No duplicate marker (idempotent `includes` check) |

yzxcj797:
## Summary

Fixes #2074 — the issue's proposed diff with three small hardenings.

## Root cause (per the issue's reproduction)

AgentRouter's gateway applies a **language filter to the first user message content** and hard-fails non-English prompts with `400 content-blocked` — a mid-session hard failure for any Portuguese/Spanish/etc. first prompt routed to an AgentRouter-backed provider (`AGR-OAI`, `AGR-CLA`, or any `baseUrl` on `agentrouter`).

## What this does

- **`applyAgrLanguagePreamble`** prepends an explicit English instruction frame to the first user message: the frame says *"respond in the appropriate language"*, so the original request and the model's output language are preserved while the boundary filter passes.
- **Idempotent** — a marker check before every prepend means retries and replays never stack frames.
- Handles all three content shapes the adapter emits: string content, structured content (prepends into the first `text` part), and content with **no** text part (inserts one).
- **Detection is `isAgentRouterBaseUrl`** on the provider's hostname, so any configured AgentRouter-backed provider gets the frame, not just the registry preset.
- Applied at the single request-build site (`messagesToAnthropicFormat` output) — the raw request and the tool-call id allocation are untouched.

## Tests

`tests/anthropic-agr-preamble.test.ts`: hostname detection (positive/negative/malformed), string prepend, structured-content prepend, no-text-part insertion, idempotence (exactly one marker after double application), and the no-user-message no-op.

## Note

The OpenAI adapter path doesn't need this (per the reproduction the filter is on the Anthropic-format gateway); if OAI-format AgentRouter routing shows the same 400, the same helper can be wired there in a follow-up.

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Improved language handling for requests routed through AgentRouter.
  * Supports both plain-text and structured message formats.
  * Prevents duplicate language instructions from being added.

* **Bug Fixes**
  * Ensures language guidance is added when structured messages lack a text segment.
  * Leaves requests unchanged when no user message is present.


lidge-jun:
## 리뷰 · 우선순위 26 / 80

AgentRouter 게이트웨이가 첫 유저 메시지를 영어가 아니면 `400 content-blocked`로 죽이는 #2074를 우회한다. 첫 유저 메시지 앞에 영어 프레임을 붙여 필터는 통과시키고, 프레임은 “적절한 언어로 답하라”고 해서 출력 언어는 유지한다. 아이디어는 한 48이다. 지금 점수는 26이다. draft, 체크리스트 0/4, hygiene `unsponsored_surface`다. `package.json` 버전을 `2.25.0`으로 올리는 줄이 원인이다.

구현은 `src/adapters/anthropic.ts`다. `isAgentRouterBaseUrl()`은 `new URL(baseUrl).hostname.includes("agentrouter")`다. `applyAgrLanguagePreamble()`은 첫 user 메시지를 찾아 문자열이면 앞에 마커를 붙이고, 구조화 content면 첫 text 파트에 붙이며, text가 없으면 text 파트를 unshift한다. 마커가 이미 있으면 그대로 둔다. 호출은 `messagesToAnthropicFormat` 직후, `isAgentRouterBaseUrl(provider.baseUrl)`일 때만이다. OpenAI 어댑터는 안 건드린다.

테스트는 hostname 양/음, 문자열 prepend, 구조화 prepend, no-text insert, idempotent, no-user no-op를 본다. `agentrouter`가 호스트 아무 데나 들어 있어도 true다. `not-agentrouter.example`나 서브도메인 위조는 이 파일에 없다. 패치는 추측하지 않음. 게이트웨이 필터가 프레임을 진짜 통과시키는 라이브 증명은 이 diff에 없다.

버전 bump는 #2099와 같은 실수다. 릴리스 서피스를 열어서 스폰서가 필요해진다. 로컬 bun 스위트는 Windows에서 안 돌렸다고 적혀 있다. 프롬프트를 바꾸는 우회라서, 필터 회피 정책은 메인테이너가 한 문장으로 인정해야 한다.

## 해결방안

`package.json` 버전 bump를 되돌리고 draft/체크리스트를 채워라. hostname 매칭은 exact/suffix로 좁히고 위조 호스트 테스트를 추가하라. OpenAI 경로는 재현이 나오기 전에 열지 마라. 그 전엔 머지하지 않는다. 필터를 “속이는” 프리앰블은 스폰서 한 줄이 필요하다.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
Thank you @yzxcj797 — absorbed as #2162, and the approach is yours: the framing belongs in the first user turn, because AgentRouter's filter reads that turn and an Anthropic `system` string never reaches it. That diagnosis was right and it is what the replacement implements.

Two things changed on top of your patch.

**Host matching.** `isAgentRouterBaseUrl` used `hostname.includes("agentrouter")`, which also matches `notagentrouter.example` and `agentrouter.org.attacker.example`. Since this feature injects an instruction into an outbound prompt, the predicate keys a prompt mutation on a provider's identity — so it has to match that identity exactly. #2162 accepts `agentrouter.org` and its real subdomains only, with both spoof shapes pinned by tests.

**Where the marker goes.** Your version spliced it into the user's own string:

```ts
firstUser.content = `${AGR_PREAMBLE_MARKER}\n\n${firstUser.content}`;
```

That rewrites what the user wrote, so logs, retries, and any upstream echo show a sentence they never typed as if they had — the concern raised in #1804. #2162 adds the marker as its own leading text block instead, converting a string to blocks when needed. Same signal to the filter, and the original text survives byte-for-byte.

Also: idempotence is now keyed on the leading block being exactly the marker rather than a substring test, so a user who quotes the marker mid-prompt does not accidentally suppress their own framing.

Your branch was marked `CONFLICTING` only because of an inherited `package.json` bump to 2.25.0 — the Anthropic hunks themselves merge cleanly against `dev`. Worth knowing for next time: cutting the branch from `dev` rather than the release line avoids that entirely.

Full suite 13529 pass / 0 fail. Closing this in favor of #2162, with the fix credited to you.

PR #2085, opened 2026-08-19
fix(admission): resolve the per-model window the way the catalog does

ntdatt812:
## Summary

`resolveInputCeiling` reads two per-model maps with a bare lookup:

```ts
const configured = positive(provider.modelContextWindows?.[modelId]) ?? positive(provider.contextWindow);
...
const configuredMaxInput = positive(provider.modelMaxInputTokens?.[modelId]);
```

The catalog resolves those same two maps through `modelRecordValue` (`src/codex/catalog/provider-fetch.ts:612`), which also accepts a family entry for a tagged id — the documented behaviour for `gpt-oss` covering `gpt-oss:120b`.

## Why it bites

Config:

```jsonc
{ "contextWindow": 8000, "modelContextWindows": { "gpt-oss": 131072 } }
```

Request to `gpt-oss:120b`:

| | value |
|---|---|
| what the catalog advertises | **131 072** |
| what admission used as the ceiling | **8 000** |

The bare lookup misses `gpt-oss:120b`, so `configured` falls through to the provider-wide `contextWindow` — a window that belongs to a different model — and the gate refuses turns the routed model can plainly hold. A ~50k-token turn is rejected pre-dispatch against a 131k model.

That inverts this module's own stated contract:

> Deliberately narrow. This is not a context manager and not a compaction trigger: it catches the pathological case and stays out of the way otherwise. **Every uncertainty resolves toward admitting.**

`modelMaxInputTokens` has the mirror of the same bug: a family cap silently never applies to the tagged sibling it was written for. That one fails open, so it is a missed cap rather than a wrong refusal — but it is the same divergence.

## Fix

Both lookups go through `modelRecordValue`. Nothing else changes; an id that already resolved exactly resolves to the same value, and `modelRecordValue` is pure, so the module stays free of filesystem, catalog and registry reads (the existing "touches no filesystem" test still passes).

## Verification

Three new tests in `tests/input-admission.test.ts`. Reverting only the `src` change turns **exactly those three red** and nothing else:

```
(fail) resolveInputCeiling > a family entry covers its tagged siblings, like the catalog
(fail) resolveInputCeiling > an exact entry still beats the family entry
(fail) resolveInputCeiling > a family modelMaxInputTokens tightens its tagged siblings
 19 pass, 3 fail
```

with the fix: `22 pass, 0 fail`.

The first test asserts `modelRecordValue(...)` — the catalog's own answer — *before* asserting the ceiling, so the two cannot drift apart again without the test noticing.

Blast radius, run locally:

```
tests/input-admission.test.ts tests/codex-catalog.test.ts tests/cli-models.test.ts
tests/routing-compatibility-model-matching.test.ts tests/openai-chat-hardening.test.ts
tests/anthropic-hardening.test.ts
-> 293 pass, 0 fail
```

`npx tsc --noEmit` — no errors.

## Review readiness checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved input limit handling for model families and specific model variants.
  - Context-window and maximum-input-token limits are now resolved correctly, including model-specific overrides.
  - Prevents valid requests from being incorrectly rejected or limits from being applied inaccurately.

### DIFF
--- a/src/server/responses/input-admission.ts
+++ b/src/server/responses/input-admission.ts
@@ -13,6 +13,7 @@
 import { nativeOpenAiContextWindow, nativeOpenAiMaxInputTokens, type NativeContextLimitsInput } from "../../codex/catalog/metadata";
 import { estimateTokens } from "../../lib/token-estimate";
 import { isCanonicalOpenAiForwardProvider, OPENAI_CODEX_PROVIDER_ID } from "../../providers/openai-tiers";
+import { modelRecordValue } from "../../reasoning-effort";
 import type { OcxContentPart, OcxParsedRequest, OcxProviderConfig } from "../../types";
 
 /**
@@ -133,7 +134,11 @@ export function resolveInputCeiling(
   // config here so this stays pure: no filesystem, no catalog, no registry scan.
   nativeContextCap?: NativeContextLimitsInput,
 ): number | null {
-  const configured = positive(provider.modelContextWindows?.[modelId]) ?? positive(provider.contextWindow);
+  // `modelRecordValue`, not a bare lookup: the catalog resolves these same two maps that
+  // way, so a `gpt-oss` entry covers `gpt-oss:120b`. Reading raw here made the gate fall
+  // back to the provider-wide window and refuse turns the model can plainly hold.
+  const configured = positive(modelRecordValue(provider.modelContextWindows, modelId))
+    ?? positive(provider.contextWindow);
 
   // The canonical `openai` registry entry declares no context fields, so without this the
   // gate would be inert on the default Codex route. All three clauses are load-bearing: a
@@ -156,7 +161,7 @@ export function resolveInputCeiling(
 
   const window = canonicalNativeBare ? native : configured;
   // modelMaxInputTokens is an input-only cap, so it can only tighten the window.
-  const configuredMaxInput = positive(provider.modelMaxInputTokens?.[modelId]);
+  const configuredMaxInput = positive(modelRecordValue(provider.modelMaxInputTokens, modelId));
   const limits = [window, configuredMaxInput, nativeMaxInput].filter((v): v is number => v !== null);
   return limits.length === 0 ? null : Math.min(...limits);
 }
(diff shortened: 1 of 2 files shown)


lidge-jun:
Reviewed as part of a four-PR batch (#2077, #2085, #2086, #2100) applying the same `modelRecordValue` migration at four call sites. **This one looks good to merge.**

The "definite wrong answer" framing is verified: a missed `modelContextWindows` lookup does not degrade to unknown, it falls through to the provider-wide `contextWindow` (`src/server/responses/input-admission.ts:136`), so the admission gate refuses turns the model can plainly hold. For `modelMaxInputTokens` a miss becomes absent instead, which fails open by dropping the intended cap — both directions are worth fixing and both are fixed here.

Both per-model reads in the file are migrated; nothing was missed.

The family assertions are real oracles — they fail against the unfixed code. Two of the surrounding assertions are controls that pass either way (the direct `modelRecordValue` call, and the exact `gpt-oss:20b` lookup); that is fine as documentation of intent, just noting it so the red-before-fix property is not overstated.

lidge-jun:
Reviewed as part of a four-PR batch with #2077, #2086, and #2100 — same idea (bare `map?.[modelId]` -> `modelRecordValue`) at four call sites, so it got one shared verdict on the contract plus per-PR verdicts.

**Verdict: merge.**

The "definite wrong answer" framing checks out, and it is the part worth stating plainly: a missed `modelContextWindows` lookup does not degrade to unknown, it falls through to the provider-wide `contextWindow` (`src/server/responses/input-admission.ts:136`). So the admission gate refuses turns the model can plainly hold, rather than failing open. The `modelMaxInputTokens` half is the opposite shape — a miss omits the intended cap — which is worth knowing but is not what this PR is primarily fixing.

Both per-model reads in the file are migrated; nothing was left raw.

One note for future tests rather than a change request: the assertions that call `modelRecordValue(...)` directly are ground truth, not coverage — they pass before the fix. The test earns its keep on the `resolveInputCeiling` family assertion.

PR #2086, opened 2026-08-19
fix(cli): report model capabilities the way the runtime resolves them

ntdatt812:
## Summary

`ocx models` builds each row with bare lookups:

```ts
const noVision = prov.noVisionModels?.includes(model);
const modalities = inputModalities[model] ?? (noVision ? ["text"] : null);
const efforts = reasoningEfforts[model] ?? prov.reasoningEfforts ?? null;
...
contextWindow: contextWindows[model] ?? globalContext,
```

The proxy resolves those same four fields through `modelInList` / `modelRecordValue`, both of which accept a family entry for a tagged id — `src/vision/index.ts:33-34` (`isModelTextOnly`), `src/reasoning-effort.ts:108`, `src/server/effort-policy.ts:122`, `src/codex/catalog/provider-fetch.ts:612`.

## Why it bites

```jsonc
{
  "models": ["gpt-oss:120b"],
  "modelContextWindows": { "gpt-oss": 131072 },
  "noVisionModels": ["gpt-oss"],
  "modelReasoningEfforts": { "gpt-oss": ["low", "high"] }
}
```

Measured on this branch's parent:

```
RUNTIME  isModelTextOnly = true
RUNTIME  context window  = 131072
RUNTIME  efforts         = ["low","high"]

ocx models --json
  {"model":"gpt-oss:120b","contextWindow":null,
   "inputModalities":null,"reasoningEfforts":null}
```

Every field comes back unclassified. A model the proxy **will** treat as text-only is listed with no modality restriction, so it reads as image-capable; a window the proxy honours reads as unset. This is the command operators use to check what a config actually did, and it disagreed with the proxy on a config the proxy honours in full.

After the change the row is `{"contextWindow":131072,"inputModalities":["text"],"reasoningEfforts":["low","high"]}` — identical to the runtime on all three.

## Verification

Two tests in `tests/cli-models.test.ts`, and I want to be precise about what each one is worth:

- **"a family entry classifies its tagged siblings, as the runtime does"** — the bug proof. It calls `isModelTextOnly` and asserts *that* first, so the command is pinned to the runtime's answer rather than to a second copy of the rule. Red without the src change (`16 pass, 1 fail`).
- **"an exact entry still wins over the family entry"** — passes on the old code too. It is a guard against the fix over-reaching, not evidence of the defect. Including it as proof would be overstating it.

Blast radius, run locally:

```
tests/cli-models.test.ts tests/vision-eligibility.test.ts
tests/codex-catalog.test.ts tests/input-admission.test.ts
-> 237 pass, 0 fail
```

`npx tsc --noEmit` — no errors.

Related: this is the same divergence class as #2042 and #2059, on the CLI surface.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model capability detection for routed and variant model names.
  * Ensured modality, reasoning effort, and context-window settings resolve correctly.
  * Exact model metadata now takes precedence over family-level defaults.
  * Family-level capabilities are correctly applied to tagged sibling models.
  * Improved no-vision classification when explicit modality settings are configured.
  * Updated capability resolution for model variants to provide more consistent CLI behavior.

### DIFF
--- a/src/cli/models.ts
+++ b/src/cli/models.ts
@@ -102,8 +102,11 @@ function collectModels(config: OcxConfig, providerFilter?: string): ModelEntry[]
       // proxy will not honour: `isModelTextOnly` matches noVisionModels with modelInList
       // and reads modelInputModalities with modelRecordValue, so a `gpt-oss` entry covers
       // `gpt-oss:120b`. A bare lookup reported that model as unclassified on every field.
+      // noVisionModels is checked first because `isModelTextOnly` returns true on that
+      // match before it ever reads modelInputModalities: a `gpt-oss` noVision entry beats
+      // an exact `gpt-oss:120b` entry that lists "image", and the proxy rejects the image.
       const noVision = modelInList(prov.noVisionModels, model);
-      const modalities = modelRecordValue(inputModalities, model) ?? (noVision ? ["text"] : null);
+      const modalities = noVision ? ["text"] : (modelRecordValue(inputModalities, model) ?? null);
       const efforts = modelRecordValue(reasoningEfforts, model) ?? prov.reasoningEfforts ?? null;
 
       entries.push({
--- a/src/cli/models.ts
+++ b/src/cli/models.ts
@@ -5,11 +5,11 @@ import { randomUUID } from "node:crypto";
 import { createInterface } from "node:readline/promises";
 import { syncModelsToCodex } from "../codex/sync";
 import { hasOwnProvider, isValidProviderName, loadConfig, saveConfig } from "../config";
-import { canonicalizeReasoningEfforts, isDeclaredReasoningEffort } from "../reasoning-effort";
+import { canonicalizeReasoningEfforts, isDeclaredReasoningEffort, modelRecordValue } from "../reasoning-effort";
 import { encodedModelIdCollides, routedSlug, slugEquals } from "../providers/slug-codec";
 import { knownModelIdsForProvider } from "../router";
 import { findLiveProxy } from "../server/proxy-liveness";
-import type { OcxConfig, OcxCustomModel } from "../types";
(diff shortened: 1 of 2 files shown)


lidge-jun:
Reviewed as part of a four-PR batch (#2077, #2085, #2086, #2100) applying the same `modelRecordValue` migration at four call sites. **Content looks good to merge** — it is still marked draft.

The ordering claim checks out against `src/vision/index.ts:29`: `isModelTextOnly` returns true on the `noVisionModels` match before it reads `modelInputModalities`, and this PR mirrors that order. Getting it backwards would make `ocx models` advertise image support the proxy then rejects — so the precedence is the substantive part of this change, not just the lookup swap. Upgrading `.includes(model)` to `modelInList` is the right companion.

Both new CLI tests are real oracles: they fail against the unfixed code on modality, context window, and effort output, and the no-vision-versus-exact-image case is a meaningful precedence test.

Worth noting for the sibling PR: #2100 applies the same migration to routing capability evidence **without** this no-vision precedence, so the two surfaces would disagree. That is flagged over there.

Minor: the description says two tests; the diff adds three.

lidge-jun:
Reviewed as part of a four-PR batch with #2077, #2085, and #2100.

**Verdict: merge** (still draft — flip it when you are ready).

The ordering call is the valuable part of this PR and it is correct: `noVisionModels` is evaluated before `modelInputModalities`, which matches `isModelTextOnly` in `src/vision/index.ts:29` — it returns true on the no-vision match before it ever reads the modality map. Reversing that would make the CLI advertise image support the proxy then rejects, which is a worse failure than the one being fixed. Upgrading `.includes(model)` to `modelInList` is right for the same reason.

The family-classification test is a real oracle: it fails against the old code on modality, context window, and effort output together.

Minor: the description says two tests, the diff adds three.

PR #2100, opened 2026-08-19
fix(routing): resolve capability evidence the way the resolver resolves it

ntdatt812:
## Summary

`candidateCapabilityEvidence` describes what the resolver will do with a candidate, and routing acts on it. It read three per-model maps with bare lookups:

```ts
const rawContextWindow = provider?.modelContextWindows?.[modelId] ?? provider?.contextWindow ?? ...
const modalities       = provider?.modelInputModalities?.[modelId] ?? ...
const reasoningEfforts = provider?.modelReasoningEfforts?.[modelId] ?? ...
```

Every runtime reader of those maps goes through `modelRecordValue`, which accepts a family entry for a tagged id — `src/reasoning-effort.ts:108`, `src/server/effort-policy.ts:122`, `src/vision/index.ts:34`, `src/codex/catalog/provider-fetch.ts:612`.

## Why it bites

```jsonc
{
  "contextWindow": 8000,
  "modelContextWindows":   { "gpt-oss": 131072 },
  "modelInputModalities":  { "gpt-oss": ["text"] },
  "modelReasoningEfforts": { "gpt-oss": ["low", "high"] }
}
```

Measured on this branch's parent, for `gpt-oss:120b`:

```
runtime    window 131072 · isModelTextOnly true · efforts ["low","high"]
evidence   {"contextWindow":8000,"tools":true,"serviceTier":"unsupported",...}
```

The **window is the serious one**. It did not degrade to unknown — it fell through to the provider-wide `contextWindow`, so routing weighed the candidate with a definite value belonging to a different model. That is the failure mode this module's own contract is written against:

> "Unknown is not zero": any dimension without canonical evidence stays `undefined` (unknown) and the profile's `unknownEvidence` policy decides how that affects eligibility.

`image` and `reasoningEfforts` merely went absent, which at least reads honestly as unknown.

There is a second, smaller thing the same change fixes: a bare lookup answers for prototype-shaped ids, so a model named `constructor` resolved `Object.prototype.constructor` as its evidence. `modelRecordValue` uses `hasOwnProperty`, so it now resolves nothing.

After the change the evidence is `{"contextWindow":131072,"image":false,"reasoningEfforts":["low","high"],...}` — matching the runtime on all three.

## Verification

Eight tests in a new `tests/routing-capability-model-matching.test.ts`. Reverting only the `src` change gives **5 fail / 3 pass**:

```
(fail) a family entry covers its tagged siblings, as the resolver does
(fail) the window does not fall through to the provider-wide value
(fail) a registry entry covers its tagged siblings with no provider configured
(fail) noVisionModels beats an exact modality entry, as isModelTextOnly does
(fail) a prototype-shaped model id resolves nothing
```

The three that pass on the old code are deliberate and are **not** evidence of the defect: exact-entry-beats-family, unrelated-model-still-falls-back, and a-model-outside-`noVisionModels`-keeps-its-image-modality. Each guards the fix from over-reaching in one direction.

Two tests assert ground truth before asserting evidence — `modelRecordValue(...)` and `isModelTextOnly(...)` are pinned first, so the file is tied to the resolver's own answer rather than to a second copy of the rule.

Blast radius, run locally on the rebased branch:

```
routing-capability-model-matching · routing-capability-catalog · route-explainability
service-tier-capability · fastwire-characterization-routing
routing-compatibility-model-matching · catalog-vision-sidecar-modalities
routing-policy-fallback
-> 93 pass, 0 fail
```

Full `bun run test`: one unrelated failure, `CL-03 ... preserves the output byte ceiling as output_byte_limit` in `tests/lab-live-pinned-timeouts.test.ts`. It passes 3/3 in isolation. That file's default `firstByteTimeoutMs` is 30 ms, so under full-suite CPU contention the first-byte timeout fires before 16 bytes accumulate — a timing race. The file imports only `lib/lab-live-pinned-sender`, nothing this PR touches. Flagging it rather than quietly calling the run green.

`bun run typecheck` — exit 0.

Related: same divergence class as #2042, #2059 and #2086, on the routing-evidence surface.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model capability detection for context windows, supported input types, and reasoning settings.
  * Model families and tagged variants are matched more reliably, with exact model settings taking priority.
  * Text-only model configuration now takes precedence when determining vision support.
  * Prevented unrelated or specially named models from being incorrectly treated as supported capabilities.
* **Tests**
  * Added coverage for family matching, exact-model precedence, provider and registry fallbacks, vision exclusions, and invalid model identifiers.

### DIFF
--- a/src/routing/capability.ts
+++ b/src/routing/capability.ts
@@ -21,6 +21,7 @@ import {
   nativeReasoningEfforts,
 } from "../codex/catalog/metadata";
 import { readCatalog, readCodexCatalogPath } from "../codex/catalog/parsing";
+import { modelRecordValue } from "../reasoning-effort";
 import { statSync } from "node:fs";
 import type { RouteCapabilityEvidence } from "./trace";
 
@@ -159,9 +160,14 @@ export function candidateCapabilityEvidence(
   const catalogRow = cachedCatalogModels().find(model => model.provider === providerName && model.id === modelId);
   const isNative = providerName === OPENAI_CODEX_PROVIDER_ID && !modelId.includes("/");
 
-  const rawContextWindow = provider?.modelContextWindows?.[modelId]
+  // `modelRecordValue`, not a bare lookup: every runtime reader of these three maps
+  // resolves them that way, so a `gpt-oss` entry covers `gpt-oss:120b`. Reading raw
+  // made the evidence disagree with the resolver it claims to describe — and for the
+  // window it did not even degrade to unknown, it fell through to the provider-wide
+  // value, which is a definite wrong answer rather than an absent one.
+  const rawContextWindow = modelRecordValue(provider?.modelContextWindows, modelId)
     ?? provider?.contextWindow
-    ?? registryEntry?.modelContextWindows?.[modelId]
+    ?? modelRecordValue(registryEntry?.modelContextWindows, modelId)
     ?? catalogRow?.contextWindow
     ?? (isNative ? nativeOpenAiContextWindow(modelId, nativeContextLimits(config)) : undefined);
   // Native rows go through the accessor (raise-to-ceiling + opt-in). Routed rows keep
@@ -170,8 +176,8 @@ export function candidateCapabilityEvidence(
     ? (nativeOpenAiContextWindow(modelId, nativeContextLimits(config)) ?? rawContextWindow)
     : rawContextWindow;
 
-  const modalities = provider?.modelInputModalities?.[modelId]
-    ?? registryEntry?.modelInputModalities?.[modelId]
+  const modalities = modelRecordValue(provider?.modelInputModalities, modelId)
+    ?? modelRecordValue(registryEntry?.modelInputModalities, modelId)
     ?? catalogRow?.inputModalities
     ?? (isNative ? nativeInputModalities(modelId) : undefined);
   const image = Array.isArray(modalities)
@@ -196,8 +202,8 @@ export function candidateCapabilityEvidence(
     || provider?.parallelToolCalls === true
     || undefined;
 
-  const reasoningEfforts = provider?.modelReasoningEfforts?.[modelId]
-    ?? registryEntry?.modelReasoningEfforts?.[modelId]
+  const reasoningEfforts = modelRecordValue(provider?.modelReasoningEfforts, modelId)
+    ?? modelRecordValue(registryEntry?.modelReasoningEfforts, modelId)
     ?? (isNative ? nativeReasoningEfforts(modelId) : undefined);
 
   const tierSupport = provider
--- a/src/routing/capability.ts
+++ b/src/routing/capability.ts
@@ -10,7 +10,7 @@
  * how that affects eligibility.
  */
 
-import type { OcxConfig } from "../types";
+import { modelInList, type OcxConfig } from "../types";
 import { isCanonicalOpenAiForwardProvider, OPENAI_CODEX_PROVIDER_ID } from "../providers/openai-tiers";
 import { serviceTierSupportForModel } from "../providers/service-tier";
 import { PROVIDER_REGISTRY } from "../providers/registry";
@@ -176,10 +176,21 @@ export function candidateCapabilityEvidence(
     ? (nativeOpenAiContextWindow(modelId, nativeContextLimits(config)) ?? rawContextWindow)
     : rawContextWindow;
 
-  const modalities = modelRecordValue(provider?.modelInputModalities, modelId)
-    ?? modelRecordValue(registryEntry?.modelInputModalities, modelId)
-    ?? catalogRow?.inputModalities
-    ?? (isNative ? nativeInputModalities(modelId) : undefined);
+  // `noVisionModels` is checked before the modality chain because that is the order
+  // `isModelTextOnly` uses: it matches the no-vision list and returns true before it
+  // ever reads `modelInputModalities` (`src/vision/index.ts:32`). So a `gpt-oss`
+  // no-vision entry beats an exact `gpt-oss:120b` entry that lists "image", and
+  // deriving `image` from the modality chain alone reported vision on a model the
+  // runtime refuses it for. That matters more here than on the CLI surface fixed in
+  // #2086: routing *acts* on this evidence, so it would select the candidate for image
+  // work that execution then rejects.
+  const noVision = modelInList(provider?.noVisionModels, modelId);
+  const modalities = noVision
+    ? ["text"]
+    : (modelRecordValue(provider?.modelInputModalities, modelId)
+      ?? modelRecordValue(registryEntry?.modelInputModalities, modelId)
+      ?? catalogRow?.inputModalities
+      ?? (isNative ? nativeInputModalities(modelId) : undefined));
   const image = Array.isArray(modalities)
     ? modalities.includes("image")
     : undefined;
--- a/tests/routing-capability-model-matching.test.ts
+++ b/tests/routing-capability-model-matching.test.ts
@@ -49,14 +49,16 @@ describe("candidateCapabilityEvidence model matching", () => {
   });
 
   test("the window does not fall through to the provider-wide value", () => {
-    // The specific regression: 8_000 here is not "unknown", it is a definite answer
-    // belonging to a different model, and routing would act on it.
+    // The specific regression: the provider-wide 8_000 is not "unknown", it is a
+    // definite answer belonging to a different model, and routing would act on it.
+    // Asserted as the exact expected number rather than `not.toBe(8_000)`, which
+    // would also pass for `undefined` or any other wrong value.
     const evidence = candidateCapabilityEvidence(
       configFor(providerWithFamilyEntries()),
       "custom",
       "gpt-oss:120b",
     );
-    expect(evidence.contextWindow).not.toBe(8_000);
+    expect(evidence.contextWindow).toBe(131_072);
   });
 
   test("an exact entry still wins over the family entry", () => {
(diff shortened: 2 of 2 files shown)


lidge-jun:
Reviewed as part of a four-PR batch (#2077, #2085, #2086, #2100) that applies the same `modelRecordValue` migration at four call sites.

**The six migrated map reads are correct**, and the "definite wrong answer" framing checks out: a missed `modelContextWindows` lookup falls through to the provider-wide `contextWindow` (`src/routing/capability.ts:162`) rather than degrading to unknown.

**Holding on one gap.** The migration resolves `modelInputModalities` and derives `image` without first checking `noVisionModels`. With:

```ts
noVisionModels: ["gpt-oss"],
modelInputModalities: { "gpt-oss:120b": ["text", "image"] }
```

the runtime returns text-only — `isModelTextOnly` matches `noVisionModels` and returns before it ever reads the modality map — while `candidateCapabilityEvidence` reports `image: true`. Routing acts on this evidence, so it can select a candidate for image work that execution then rejects. #2086 fixes exactly this ordering on the CLI surface; the same precedence is needed here.

Also: `contextWindow.not.toBe(8_000)` is a weak assertion — it passes for `undefined` and for any other wrong value. An exact expectation would be a stronger oracle.

Minor: the description says five tests; the diff has six.

Happy to merge once the no-vision precedence lands with a regression for the conflicting-evidence case.

lidge-jun:
Reviewed as part of a four-PR batch with #2077, #2085, and #2086 — same idea at four call sites, so one shared contract verdict plus per-PR verdicts.

**Verdict: hold — the migration is incomplete in a way that matters here specifically.**

The six map reads are migrated correctly, and the context-window claim checks out: a raw miss falls through to the provider-wide `contextWindow` (`src/routing/capability.ts:162`), which is a definite wrong value rather than an absent one. That half is good.

**What is missing: `noVisionModels` precedence.** The PR resolves `modelInputModalities` and derives `image` without first checking `noVisionModels`. Given:

```ts
noVisionModels: ["gpt-oss"],
modelInputModalities: { "gpt-oss:120b": ["text", "image"] }
```

the runtime returns text-only (`isModelTextOnly` matches the no-vision list first, `src/vision/index.ts:29`) while `candidateCapabilityEvidence` reports `image: true`.

This is the same ordering bug #2086 fixes on the CLI surface, left unfixed on the routing surface — and it is worse here, because routing *acts* on this evidence. It can select a candidate for image work that execution then rejects. The PR's own stated goal is that the evidence agree with the resolver it describes, so this is a gap in its own terms rather than an added requirement.

**To land:** check `noVisionModels` before deriving modalities, and add a regression for the conflicting-evidence case above.

One test note: `contextWindow.not.toBe(8_000)` is red against the known bug but weak — it also accepts `undefined` or any other wrong value. Assert the expected number.

Also, the description is stale: the diff has six tests, not five, and the unfixed split should be 2 pass / 4 fail once the registry-family test is counted.

lidge-jun:
## 리뷰 · 우선순위 28 / 80

라우팅 evidence가 런타임 리졸버와 다른 맵을 읽는 버그를 고친다. `candidateCapabilityEvidence()`가 `modelContextWindows` / `modelInputModalities` / `modelReasoningEfforts`를 베어 인덱스로 봐서, `gpt-oss` 패밀리 값이 `gpt-oss:120b`에 안 붙었다. 윈도우는 unknown으로 안 떨어지고 프로바이더 전역 `contextWindow`(8000)로 떨어져 잘못된 확정 값이 됐다. `modelRecordValue`로 바꾸면 런타임과 같아진다. 아이디어는 한 50이다. 지금 점수는 28이다. draft, 체크리스트 0/4, 메인테이너가 이미 hold한 상태다.

코드는 `src/routing/capability.ts` 한곳이다. 세 맵의 프로바이더/레지스트리 조회가 `modelRecordValue(...)`를 탄다. 프로바이더 전역 `contextWindow`와 카탈로그/네이티브 fallback 순서는 그대로다. 프로토타입 키(`constructor`)는 `hasOwnProperty` 때문에 함수가 evidence로 안 들어간다. 이 모양은 #2042/#2059/#2086/#2077과 같은 클래스다.

`tests/routing-capability-model-matching.test.ts`가 패밀리 커버, 전역 윈도우로 안 떨어짐, exact > family, 무관 모델의 전역 fallback, 레지스트리 tagged sibling, prototype id를 본다. 첫 테스트는 `modelRecordValue`와 `isModelTextOnly`를 evidence보다 먼저 고정한다. 저자가 말한 대로 src만 되돌리면 3개가 깨진다.

메인테이너 hold는 이 마이그레이션이 `image`를 `modelInputModalities`에서 바로 파느라, 런타임 `isModelTextOnly`가 보는 다른 경로와 어긋날 수 있다는 점이다. 이 테스트는 custom 픽스처에서 둘을 같게 만들지만, 비전 사이드카/네이티브 분기를 전부 같게 만들었다는 증명은 아니다. 패치는 추측하지 않음. hold 이유를 이 브랜치에서 다시 반박하지 않는다.

## 해결방안

메인테이너가 지적한 `image`/`isModelTextOnly` 구멍을 메우거나, 그 차이가 이 모듈 계약상 괜찮은지 한 문장으로 답하라. 그다음 체크리스트 4칸을 채우고 draft를 해제하라. #2077과 같이 `modelRecordValue` 배치로 보지 말고, 이 파일의 evidence 계약만 닫아라. 지금 상태로는 머지하지 않는다.

이 댓글은 grok-bot이 작성했습니다

ntdatt812:
@lidge-jun the `noVisionModels` gap is real and is fixed in `3c94b9a7c`. I checked it against the code rather than taking it on trust, and you are right on both counts.

`isModelTextOnly` matches the no-vision list and returns before it ever reads `modelInputModalities` (`src/vision/index.ts:32`), so with

```ts
noVisionModels: ["gpt-oss"],
modelInputModalities: { "gpt-oss:120b": ["text", "image"] }
```

the runtime says text-only and the evidence said `image: true`. Fixed with the same ordering and the same primitives as the merged #2086, so the two surfaces now read alike:

```ts
const noVision = modelInList(provider?.noVisionModels, modelId);
const modalities = noVision ? ["text"] : (…the existing chain…);
```

I deliberately did **not** import `isModelTextOnly` here, even though reusing the runtime function is the strongest form of "resolve the way the runtime resolves". `src/vision/index.ts` pulls in the OAuth store, the sidecar and the web-search executor, and this module's docblock commits to "canonical local sources only … no live network fetch happens at routing time". There is no import cycle — I checked — so if you would rather have the direct call and accept the module weight, say so and I will switch it.

Two tests. The positive one is red without the change; the negative one — a model outside `noVisionModels` keeps its declared image modality — passes either way on purpose, so the fix cannot trade a false positive for a false negative.

**The weak oracle is also fixed.** You were right that `not.toBe(8_000)` also passes for `undefined` and for any other wrong value; it now asserts `131_072` exactly.

**Corrected counts, since the old ones were stale.** The file is eight tests, not five. Reverting only `src` gives **5 fail / 3 pass**, not 2/4 — the registry-family case counts, as you said, and the new no-vision case adds one. The description now carries these numbers instead of the old ones.

Rebased onto `7a2d13a74`; the branch was 31 commits behind. Checklist ticked and out of draft.

One thing I want to state rather than bury: the full `bun run test` had a single failure, `CL-03 … preserves the output byte ceiling as output_byte_limit` in `tests/lab-live-pinned-timeouts.test.ts`. It passes 3/3 in isolation. That file's default `firstByteTimeoutMs` is 30 ms, so under full-suite CPU contention the first-byte timeout wins the race against accumulating 16 bytes; it imports only `lib/lab-live-pinned-sender`, and this PR touches only `src/routing/capability.ts`. I am reporting it rather than calling the run clean, but I have not touched that test — tell me if you would rather it were made deterministic and I will open it separately.

ntdatt812:
Rebased again onto `0fc8d136e` and re-verified — `dev` had moved 72 commits in the hour since the previous rebase. Checklist ticked, out of draft.

`bun run typecheck` exit 0. Targeted and blast-radius suites green.

**A heads-up that is not about this PR.** The full `bun run test` on that head has two failures in `tests/ws-upstream.test.ts`:

```
(fail) handleResponses Codex WS relay selection > an HTTP fallback remains on the configured legacy tee path
(fail) handleResponses Codex WS relay selection > an older runtime stays on HTTP SSE without opening a WebSocket
```

Both assert `isEagerRelaySseResponse(response)` is `false` and get `true`. They are **not** mine and not flaky — they reproduce in isolation, and they still fail with `src/` checked out wholesale to `origin/dev` (21 pass / 2 fail either way). They also did not appear in my full-suite runs against `7a2d13a74` earlier today.

`git bisect run` over the 72 commits points at:

```
159d2ab183c69db98a2955660c2f8344f83c8e4d  fix(windows): keep catalog discovery off request event loop  (@Ingwannu)
```

It widened `collectCatalogState` to allow a `Promise` return and switched `defaultCollectCatalogState` to `collectCodexAppServerCatalogStateForRequest`. I have not dug past the bisect, so treat the mechanism as unverified — but the boundary is reproducible.

Worth knowing that I am on `bun 1.3.14`, which `tests/ws-upstream.test.ts:42` pins as *not* supporting the bounded relay, so the second test is one your CI may skip on a newer bun. The first one injects `BOUNDED_WS_RUNTIME` explicitly, so it should be runtime-independent.

Happy to open this as its own issue if you would rather not have it sitting in a PR thread.

ntdatt812:
**Correction to my previous comment — I named the wrong commit, and the wrong author.**

I attributed the two `tests/ws-upstream.test.ts` failures to `159d2ab18` (@Ingwannu). That is wrong, and I apologise for putting a name to it. My bisect declared `7a2d13a74` as the "good" end on the strength of `ws-upstream` not appearing in two earlier full-suite logs — absence from a log is not a passing test, and I should not have treated it as one. `7a2d13a74` fails too, so the whole bisect ran inside an already-broken range.

Re-bisected with **both** endpoints measured rather than assumed:

```
dec332c49  23 pass / 0 fail
5a75e57ff  21 pass / 2 fail   fix(grok): switch to Responses backend and backfill required annotations
```

That commit adds `createResponsesFieldBackfillBlockRewrite()` to `blockRewrites` **unconditionally**, so `needsClientRewrite` is now always true. `isWin32EagerRewrite` is `platform === "win32" && needsClientRewrite` (`src/lib/bun-stream-caps.ts:126`), so on Windows every Responses stream now takes the eager single-reader relay. The two tests assert `isEagerRelaySseResponse(response)` is `false`, which on win32 no longer holds for a reason unrelated to WS selection.

Instrumented at the gate to confirm rather than infer:

```
[EAGER] {"forceCodexWsEagerRelay":false,"useEagerRelay":null,"win32EagerRewrite":true,
         "needsClientRewrite":true,"platform":"win32","blockRewrites":1}
```

So this is Windows-only, it is not a WS-path defect, and it is not @Ingwannu's change. Still unrelated to this PR — but my earlier account of it was wrong and I would rather correct it in the same place I got it wrong.

lidge-jun:
Thanks for this, @ntdatt812 — closing as superseded by #2140, which carries **your patch applied unchanged**, alongside your #2077 for the Lab behavior report.

They are combined because they are one thesis on two disjoint files: model-keyed lookups must use the runtime's own resolution rules. Your tests came across verbatim, including the prototype-id cases, which are the load-bearing part — reverting just the two source files fails 13 of them.

Your work is credited in that PR's description.

PR #2118, opened 2026-08-19
fix(cursor): do not re-label a completed turn as failed when the stream is aborted

lidge-jun:
## Summary

**Stacked on #2117** → #2116 → `dev`. Retarget down the chain as parents land.

Only `cancelCursorRun()` sets `expectedClose`, so an ordinary completed turn never qualified for the benign-close path. The abort listener then failed the turn with `"Cursor request was aborted"` — which is deliberately **not** a benign cancel, since a mid-turn abort is a real failure — so a turn whose terminal frame had already been emitted, and whose messages had already been yielded to the consumer, still surfaced as `turn-failed` with `expectedClose: false`.

This returns instead of throwing when a terminal frame was already emitted **and** the failure is an abort.

**Both halves of that condition matter.** Post-terminal alone would change what the adapter sees for genuine faults; abort alone would swallow a real mid-turn abort where nothing was delivered. There is a guard test for the second case.

### What this does not touch

A benign cancel after a terminal is already swallowed one layer up (`cursor.ts:183`), and the existing contract test — *"a cancel after a terminal was already emitted does not add a second one"*, which pins that the transport **still throws the raw cancel** — keeps passing unchanged. This deliberately does not widen that path.

## Scope note

This is the **teardown-misclassification slice of #1527 only**. The issue as filed has five residual parts after #2054; the other four are not this PR:

- `kimi-k3` premature completion at ~79-95k input — needs live acceptance on the checkpoint branch
- `claude-fable-5` 429 asymmetry — may be unprovable while Connect does not expose `cache_read_tokens`
- first-turn / restart / compaction full-replay into the 512 KiB envelope
- request-shape parity with official Cursor (`maxMode`)

Hence `Refs`, not `Closes`.

## Verification

```
bun test tests/cursor-cancel-provenance.test.ts tests/cursor-eof-terminal.test.ts \
         tests/cursor-adapter.test.ts tests/cursor-errors.test.ts     46 pass / 0 fail
bun x tsc --noEmit                                                    exit 0
```

**Driven red.** Removing the two `emittedTerminal && isCursorAbortError` guards fails exactly the new post-terminal-abort test (7 pass / 1 fail) and nothing else.

Worth recording: my first attempt patched `failAndClear`, and the ablation showed **no test moved at all** — that helper is not on the path the injected failure takes. The real site is the `if (failure)` throw inside `run()`. A green suite after that first edit would have been a change that verified nothing.

## Checklist

- [x] Focused tests pass
[...]

Issue #2108, opened 2026-08-19
[Bug][Windows]: reboot leaves native-main gate stuck; GPT-only 503 until ocx restart


hoonysis:
### Client or integration

Direct HTTP/API client

### Area

Service lifecycle

### Summary

On Windows 11 with the Task Scheduler service backend, a reboot/login can leave only the native OpenAI/Codex-login models blocked behind the process-wide `native-main` startup gate.

The proxy itself is healthy (`/healthz` returns 200), the model remains in the catalog, and non-native providers such as Anthropic continue returning 200. Native GPT requests fail locally with HTTP 503:

```
OpenCodex local native-main profile maintenance is active; retry this request
```

The state does not recover on its own. Running `ocx restart` (or pressing the dashboard/launcher restart control) immediately restores the same GPT model without reauthentication or configuration changes.

This has recurred after multiple Windows reboots. I expected startup recovery to reacquire native-main ownership and reopen admission automatically.

### Reproduction

1. Install OpenCodex as the per-user Windows Task Scheduler service (logon trigger).
2. Use a valid native Codex/OpenAI login and confirm a native GPT model works.
3. Reboot Windows and log in.
4. Wait for the scheduled OpenCodex proxy to start.
5. Confirm:
   - `GET /healthz` returns 200.
   - Anthropic or another non-native provider still works.
   - A native GPT Responses request (`stream: true`, `store: false`, list-form `input`) returns local HTTP 503 with the native-main maintenance message.
6. Wait/retry: the GPT-only 503 persists.
7. Run `ocx restart`.
8. Send the same GPT request again. It returns 200 without login or config changes.

The Windows scheduled task currently reports `AllowHardTerminate=True` and `MultipleInstances=IgnoreNew`.

### Version

2.26.0 (installed and running runtime both 2.26.0)

Codex runtime reported by OpenCodex: 0.148.0

### Operating system

Windows 11 Home, version 10.0.26200, build 26200

### Provider and model

OpenAI (Codex login / native main) / gpt-5.6-sol

### Logs or error output

```shell
[17:25:23] opencodex service wrapper start
Previous session (PID <redacted>) did not shut down cleanly. Codex state restored from journal.
opencodex proxy running on http://localhost:10100

# /healthz remained HTTP 200.

17:26:37 | gpt-5.6-sol | openai | 503 | OpenCodex local native-main profile maintenance is active; retry this request
[...]

PR #2101, opened 2026-08-19
fix(codex): gate account-native models by entitlement

Ingwannu:
## Summary

- Fetch each authenticated ChatGPT/Codex account's bounded `/models` roster and treat it as the authority for account-gated native ids such as `gpt-daybreak-blue-latest`.
- Hide gated bare and selector-qualified catalog rows unless an eligible account confirms the model, and exclude unentitled accounts from Pool routing before a prompt is spent.
- Check Direct callers against their forwarded credential (or the stored main credential when an OpenCodex admission bearer is substituted).
- On the exact allow-listed unsupported-model 400, refresh that account's roster and permit up to seven additional sends on the same confirmed account, re-checking both the exact rejection and fresh entitlement before every later send. Otherwise fail over only to another confirmed eligible Pool account. Exact selectors never switch accounts.
- Document the catalog, Direct, Pool, and retry boundaries.

Closes #2097

## Verification

- `bun test tests/codex-model-entitlements.test.ts tests/codex-auth-context.test.ts tests/server-auth.test.ts tests/native-model-toggle.test.ts tests/codex-catalog-sync-hardening.test.ts tests/codex-convergence-account-selectors.test.ts tests/claude-models-discovery.test.ts` — 224 passed
- `bun test tests/openai-provider-option-tooling.test.ts tests/claude-messages-endpoint.test.ts` — 51 passed
- `bun run typecheck` — passed
- `bun run privacy:scan` — passed
- `cd docs-site && bun run build` — passed, 385 pages
- Full local suite was run under a two-core CPU limit. The changed auth/catalog tests pass; one unrelated host-environment failure remains in `Codex autostart shim > Unix shim exports persisted service API token before running Codex` because this machine's installed service-token path overrides that test's `local-secret` fixture. Required exact-head GitHub CI remains the merge authority.
- Local 2.25.0 focused overlay: service restarted healthy. A real streamed `gpt-daybreak-blue-latest` request encountered six consecutive upstream unsupported-model 400s, then completed HTTP 200 with `DAYBREAK_OK` on send 7 (upstream resolved model `gpt-5.6-sol`)./

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed.
- [ ] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. Explicit owner review requested from @lidge-jun; the author will not self-approve or merge before approval and required CI.

## Summary by CodeRabbit

* **New Features**
[...]

PR #2121, opened 2026-08-19
fix(codex): name the native-main gate reason when the fence returns 503

lidge-jun:
## Summary

A Windows reboot can leave native-main admission fenced until the user runs `ocx restart`. #2108 reports it, but the report could not be acted on: every fenced request returns the same generic 503 maintenance message, so the gate reason that actually settled was never written anywhere a user could read.

That matters because the two candidate triggers settle to *different* reasons — an ACL fail-closed settles `owner-unavailable` (`native-profile-startup.ts:138-139`), a probe fail-closed settles `ownership-unknown` (`:311`) — and nothing on the request path distinguished them. This ships the diagnostic, not the mechanism fix; guessing between two candidates without evidence is how the next reboot report ends up exactly as ambiguous as this one.

`CodexMainProfileDrainingError` now captures the settled reason in its constructor and warns once per distinct reason. The constructor is the last moment the reason is both in scope and still true — every catch site has already lost it, and re-reading the gate later can observe a recovery that completed in between.

**Three other surfaces were considered and rejected**, which is why this is a log line rather than something more structured:

- **The message** cannot carry it. `claude-messages.ts:818-820` matches that string exactly to keep the fence a 503 instead of remapping it to an Anthropic 529. A changed message would tell Claude Code to back off from an upstream that was never involved.
- **A response header** cannot reach `/api/logs`, which populates `upstreamError` by reading `json.error.message` back out of the body (`request-log.ts:755-759`); `error.code` is likewise pinned to `server_is_overloaded` for 503 by design (`request-log.ts:508-540`). The Claude surface also rebuilds its response headers from scratch (`claude-messages.ts:823-829`).
- **Threading a reason into `codexMainProfileDrainingResponse`** would mean editing all five of its zero-arg call sites, two of which are also edited by open PR #2101.

stdout reaches every path this fence fires on, and it is the stream the original report was already quoting from.

The turn-drain claim race (`auth-context.ts:326`, driven by `lifecycle.ts:180`) throws this same error while the gate reads `ready`. That site stays silent rather than borrowing a startup reason, so a reasonless 503 now means "not the startup fence" — a distinction the reporter had no way to make. Only the reason is logged; the snapshot's `homeId` is derived from a profile directory path and stays out.

Refs #2108. Deliberately not `Closes`: this makes the next occurrence diagnosable, it does not stop the fence from sticking. Phase 2 (making a boot-time `unknown` retryable instead of a process-lifetime fence) waits for a field report that names a reason.

Stacked on #2118 to keep the review surface small; retarget to `dev` once the parents land.

## Verification

```
bun test tests/codex-auth-context.test.ts          47 pass / 0 fail / 173 expect()
bun test <5 native-main gate suites>               80 pass / 0 fail / 470 expect()
bun test tests/chat-completions-endpoint.test.ts \
         tests/claude-messages-endpoint.test.ts   124 pass / 0 fail
bun x tsc --noEmit                                 exit 0
bun run privacy:scan                               exit 0
```

The five gate suites are `codex-auth-context`, `native-profile-startup`, `native-profile-drain-server`, `native-profile-stage-lifecycle`, `native-main-owner-lifetime`. `privacy:scan` is not ceremony here — this PR adds a log statement, which is what that scan exists to police.

**Red-driven, not merely green.** With the fix staged as a no-op export, two assertions failed on `reason === undefined` and the turn-drain silence case passed — proving the assertions themselves were doing the work rather than a module-load error. An independent reviewer then mutation-tested the new tests in a scratch copy: removing the `console.warn` body goes red (2 fail), and removing the dedup check goes red. Each assertion detects the removal of the behavior it claims to pin.

**One honest caveat:** these suites ran on macOS. The `owner-unavailable` branch this diagnostic most needs to name is a Windows icacls path, so this proves the plumbing rather than the field behavior. Worth knowing that no test anywhere currently asserts `owner-unavailable` — it is produced at `native-profile-startup.ts:139` and asserted nowhere in the suite.

## Checklist

### DIFF
--- a/devlog/_plan/260819_unclaimed_bug_selection/100_post_split_rebase.md
+++ b/devlog/_plan/260819_unclaimed_bug_selection/100_post_split_rebase.md
@@ -1,0 +1,61 @@
+# 100 — Post-split rebase of the stack
+
+## What changed under us
+
+The mega-file split landed while this work was open:
+
+| PR | Merge |
+|---|---|
+| #2019 WP1 types | `da86a830a` |
+| #2023 WP1b clusters | `2235f456d` |
+| #2036 WP2a config leaf | `eca18d0c8` |
+
+`src/types/tools.ts` now exists on `dev` and `OcxTool` lives there. **#1934 also
+merged** (`a5289aad5`), which `070` had named as the real collision hazard for
+`#1587` — it touched all five of that fix's files.
+
+So both of `#1587`'s blockers cleared at once. It moves from "last, most
+contested files" to implementable.
+
+## Rebase
+
+Bottom-up, three branches, **zero conflicts**:
+
+```
+fix/service-proxy-env       eb910776a -> d7caaa9bf
+fix/tray-registry-encoding  e3b063750 -> f1f400fea
+fix/cursor-abort-teardown   346eaa80d -> 2c4e00ede
+```
+
+Zero conflicts is the expected result and not the interesting part: none of the
+three fixes touches `types.ts`, `config.ts`, or anything the split moved. That
+was the collision analysis in `070`, and it held.
+
+## The verification that actually mattered
+
+A green result from *before* the rebase proves nothing here — the split replaced
+`types.ts` with a barrel and moved every type cluster, so the whole tree these
+branches compile against is different.
+
+Re-run on the rebased top of stack:
+
+```
+bun test service + tray + windows-text + cursor-cancel + cursor-eof
+  163 pass / 0 fail
+bun x tsc --noEmit
+  exit 0
+```
+
+## Deferred items: what is still blocked
+
+Re-checked live rather than carried from the previous turn:
+
+| PR | State | Consequence |
+|---|---|---|
+| #2029 | OPEN, `CHANGES_REQUESTED` | `#2114` still must not get a competing PR |
+| #2101 | OPEN | `#2108` phase 1 still collides on `server/index.ts` + `auth-context.ts` |
+| #2054 | OPEN, `CHANGES_REQUESTED` | the kimi-k3 and 429 halves of `#1527` still cannot start |
+
+Only `#1587` actually became unblocked. The other three deferrals stand for the
+same reasons they were recorded, which is worth stating explicitly — "the split
+landed" is not a general unblock.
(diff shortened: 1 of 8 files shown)


lidge-jun:
## 리뷰 · 우선순위 70 / 80

#2108 재부팅 후 native-main이 막히면 전부 같은 503이라 원인이 안 보임. ACL fail-closed는 `owner-unavailable`, 프로브 fail-closed는 `ownership-unknown`. 이 PR은 그 이유를 `CodexMainProfileDrainingError` 생성자에 넣고, 사유마다 `console.warn` 한 번. 수리 아님. 고의로 Closes 안 함.

메시지에 이유를 넣으면 안 됨. `claude-messages.ts`가 그 문자열을 정확히 매칭해서 503을 유지함. 바꾸면 Claude가 529로 착각하고 업스트림 백오프함. 헤더도 `/api/logs`가 `json.error.message`만 봄. `codexMainProfileDrainingResponse`에 스레드하면 호출 5곳 + #2101이랑 겹침. stdout이 리포트가 이미 보던 길이라 그거 맞음.

턴 드레인 레이스는 게이트가 `ready`인데도 같은 에러를 던짐. 여긴 침묵. 이유 없는 503은 스타트업 펜스가 아님. `homeId`는 로그에 안 넣음. `privacy:scan` 통과했다고 함.

베이스가 `dev`가 아니라 `fix/cursor-abort-teardown`(#2118)임. MERGEABLE. 부모 머지 전에 `dev`로 리타깃하지 말 것. 스플릿 메가파일은 안 건드림. `src/codex/auth-context.ts`, `native-profile-startup.ts`, `tests/codex-auth-context.test.ts` + devlog.

macOS에서만 돌려서 `owner-unavailable`(윈도 icacls) 필드는 증명 안 됨. 그 브랜치를 assert하는 테스트도 원래 없음. Phase 2(부팅 `unknown`을 프로세스 평생 펜스에서 빼기)는 사유가 찍힌 리포트 온 다음.

해결방안: #2118 먼저 머지하고 리타깃. 로그 포맷 고정해서 다음 #2108 리포트가 사유를 붙이게. 503 문구/헤더는 그대로.

이 댓글은 grok-bot이 작성했습니다

PR #2122, opened 2026-08-19
feat(catalog): config-level retainModels allowlist for authoritative discovery (#1690)

### LINKED ISSUE #1690: [Feature]: config-level retainModels allow-list so operators keep configured models in the authoritative live catalog
### Problem to solve

When a provider does live model discovery, the authoritative catalog **drops configured models** that the upstream does not (yet) return — even when the operator explicitly configured them and the wire mapping exists in code. Concretely:

* `google-antigravity` (Cloud Code Assist) does not return `gemini-3.7-flash` from `:fetchAvailableModels` for accounts where the model is not yet provisioned, so the picker loses the model even though the operator configured it (see lidge-jun/opencodex#1683 for the user-facing symptom).
* Today the only retention paths are: the **hardcoded** `CALLABLE_CONFIGURED_COMPATIBILITY_MODELS` allow-list (kimi, xai), combo targets, and the Vertex default seed. There is no operator-facing, config-level way to say "keep this configured model in the catalog even if discovery omits it."

Operators who run proxies for their own account know their provider's roadmap (e.g. "gemini-3.7-flash is coming to my region; keep the row so it appears the moment the upstream enables it") and should not need a code change per model.

### Proposed solution

1. Add a config-level retention list per provider: `providers.<name>.retainModels: string[]`.

   ```json
   {
     "providers": {
       "google-antigravity": {
         "retainModels": ["gemini-3.7-flash"],
         "modelCosts": {
           "gemini-3.7-flash": { "input": 0.75, "output": 3.75, "cacheRead": 0.15, "cacheWrite": 0 }
         }
       }
     }
   }
   ```
2. In `mergeConfiguredModelsIntoLiveCatalog` (`src/codex/catalog/provider-fetch.ts`), extend the retention condition:

   ```diff
   if (
     seedVertexDefault === true
     || shouldRetainConfiguredProviderModel(name, candidate.id)
   + || (retainConfiguredModelIds?.has(candidate.id) === true)
   + || (providerRetainModels?.has(candidate.id) === true)
     || (retainComboTargets && retainConfiguredModelIds?.has(candidate.id) === true)
   ) {
     out.push(candidate);
     present.add(candidate.id);
     continue;
   }
   ```
3. When a retained model is invoked but the upstream returns 404/not-found, surface a **clear diagnostic** (log + `warnDroppedConfiguredIdsOnce`-style one-shot warning) so a model that "exists in the picker but is not callable yet" is not silently hidden — the operator sees exactly why.
4. Seed `CALLABLE_CONFIGURED_COMPATIBILITY_MODELS` with `google-antigravity: ["gemini-3.7-flash"]` as the first entry, so the picker row exists out-of-the-box for antigravity users while the upstream catches up (mirrors the existing kimi/xai pattern).

### Alternatives considered

* Keep the hardcoded allow-list only: works but requires a code change + release per model/per provider; does not scale to operator-specific account states.
* Make discovery "never drop configured models": would resurrect stale/broken models for everyone (the reason the authoritative catalog exists); the retention must be explicit and opt-in.
* Document "wait for upstream": not actionable for operators who know their model is coming; the row should exist with a clear "not callable yet" state.

### Additional context

* Related open issue: lidge-jun/opencodex#1683 (gemini-3.7-flash dropped from Antigravity live discovery).
* Precedent for static + live coexistence: other gateways keep a static model list alongside live discovery (e.g. OmniRoute `QODER_STATIC_MODELS` + live `--list-models` fallback in `open-sse/services/qoderCli.ts`).
* The existing `warnDroppedConfiguredIdsOnce` infrastructure already logs dropped configured ids once per signature — the retention path should reuse it for the inverse case (retained-but-404) so logs stay quiet in steady state.

Acceptance criteria:

* `retainModels` entry absent from discovery → row present in catalog (visibility `list`).
* Invoking a retained-but-unavailable model → 4xx from upstream is surfaced with a one-shot diagnostic log; the catalog row stays.
* Without `retainModels` → current behavior unchanged (no behavior change for existing users).

chilung-cgu:
## Summary

- Add `retainModels?: string[]` provider configuration allowlist to guarantee configured models are preserved in catalog even if upstream live discovery does not report them (resolves #1690).
- When a provider uses live model discovery (such as Google Antigravity / Gemini or Vertex), models configured in `retainModels` will not be dropped by `mergeConfiguredModelsIntoLiveCatalog`.
- Updated provider type definitions in `src/types/provider.ts`, config parsing in `src/config.ts`, model rename migration, and docs reference.
- Added one-shot 404 diagnostic with proper generation reconciliation cleanup across provider state changes.

## Verification

- Added regression test suite `tests/provider-retain-models.test.ts` verifying:
  - Configured models in `retainModels` are retained when live discovery reports an empty or partial list.
  - Normal unlisted models not in `retainModels` continue to follow standard live discovery pruning behavior.
  - Full catalog sync retaining expected combo targets.
  - Diagnostic memo cleanup across generation reconciliation.
- Verified cleanly rebased on latest `origin/dev` with strict typechecking and privacy scan passed.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added an optional provider setting to retain specified models when live discovery omits them.
  * Retained models are validated, deduplicated, preserved across catalog refreshes, and supported by model-renaming migrations.
* **Bug Fixes**
  * Added one-time warnings when retained models are unavailable upstream.
  * Improved retained-model handling across caching, filtering, static catalogs, and supported API endpoints.
* **Documentation**
  * Documented the new configuration option and behavior in supported languages.

### DIFF
--- a/docs-site/src/content/docs/ja/reference/configuration/providers.md
+++ b/docs-site/src/content/docs/ja/reference/configuration/providers.md
@@ -68,6 +68,7 @@ account を削除しても mapping は保持され、同じ id を再追加す�
 | `models?` | `string[]` |シード/フォールバック モデルのリスト。 `liveModels: false` では、発見されたモデルはこれらのみです。 |
 | `liveModels?` | `boolean` |開始/同期時にライブ カタログをフェッチします (デフォルトは `true`)。カスタムプロバイダーは `${baseUrl}/models` を使用します。組み込みはレジストリ URL とフィルターを使用する場合があります。 |
 | `selectedModels?` | `string[]` |検出後のカタログ許可リスト。空でない場合は、それらの ID のみが公開されます。空または省略すると、検出されたすべてのモデルが公開されます。 |
+| `retainModels?` | `string[]` | ライブディスカバリで省略された場合でもカタログに保持するモデルIDの許可リスト（未プロビジョニングのアカウントや未掲載モデルなど）。ここに記載されていない設定済みIDは、ディスカバリで省略された場合に引き続き除外（prune）されます。 |
 | `contextWindow?` | `number` | アップストリームのメタデータが無い場合に使うプロバイダー全体のコンテキスト値。メタデータがある場合は上限として働き、より小さいライブ値をそのまま残します。Models ダッシュボードでは `providerContextCaps` とは別に設定します。 |
 | `modelContextWindows?` | `Record<string, number>` | モデルごとのコンテキスト値および上限。`contextWindow` より優先され、ウィンドウが不明なら設定値を使い、より小さいライブメタデータがあればそちらが優先されます。 |
 | `modelInputModalities?` | `Record<string, string[]>` | `["text"]` や `["text", "image"]` などのモデルごとの入力ヒント。 |
--- a/docs-site/src/content/docs/ko/reference/configuration/providers.md
+++ b/docs-site/src/content/docs/ko/reference/configuration/providers.md
@@ -68,6 +68,7 @@ managed map을 활성화하면 privacy-safe selector를 만들고, 이후 계정
 | `models?` | `string[]` | 시드/폴백 모델 목록입니다. `liveModels: false`이면 이 목록만 발견된 모델로 취급합니다. |
 | `liveModels?` | `boolean` | 시작 또는 동기화 시 라이브 카탈로그를 가져옵니다. 기본값은 `true`입니다. 사용자 지정 공급자는 `${baseUrl}/models`를 사용하고, 내장은 레지스트리 URL을 사용한 뒤 필터링할 수 있습니다. |
 | `selectedModels?` | `string[]` | 발견 후 카탈로그 허용 목록입니다. 값이 비어 있지 않으면 그 id만 노출하고, 비어 있거나 생략하면 발견된 모델을 모두 노출합니다. |
+| `retainModels?` | `string[]` | 라이브 발견에서 누락되더라도 카탈로그에 유지할 모델 ID 허용 목록입니다 (예: 미프로비저닝 계정 또는 미등록 모델). 여기에 나열되지 않은 구성 ID는 발견에서 누락될 때 계속 제외(prune)됩니다. |
 | `contextWindow?` | `number` | 업스트림 메타데이터가 없을 때 쓰이는 공급자 전반의 컨텍스트 값입니다. 메타데이터가 있으면 상한으로 동작해 더 작은 라이브 값을 그대로 둡니다. Models 대시보드에서 `providerContextCaps`와 별도로 설정합니다. |
 | `modelContextWindows?` | `Record<string, number>` | 모델별 컨텍스트 값이자 상한입니다. `contextWindow`보다 우선하며, 창 크기를 알 수 없으면 설정값을 쓰고 더 작은 라이브 메타데이터가 있으면 그쪽을 따릅니다. |
 | `modelInputModalities?` | `Record<string, string[]>` | `["text"]` 또는 `["text", "image"]` 같은 모델별 입력 힌트입니다. |
--- a/docs-site/src/content/docs/reference/configuration/providers.md
+++ b/docs-site/src/content/docs/reference/configuration/providers.md
@@ -82,6 +82,7 @@ differing backup and rewrites known legacy namespaced selected ids to bare ids.
 | `models?` | `string[]` | Seed/fallback model list. With `liveModels: false`, these are the only discovered models. |
 | `liveModels?` | `boolean` | Fetch the live catalog on start/sync (default `true`). Custom providers use `${baseUrl}/models`; built-ins may use a registry URL and filter. |
 | `selectedModels?` | `string[]` | Catalog allowlist after discovery. Non-empty exposes only those ids; empty or omitted exposes all discovered models. |
+| `retainModels?` | `string[]` | Models to retain in the authoritative live catalog even when upstream discovery omits them (e.g. unprovisioned accounts or unlisted models). |
 | `contextWindow?` | `number` | Provider-wide context fallback when upstream metadata is absent; otherwise a cap that retains smaller live metadata. The Models dashboard exposes this separately from `providerContextCaps`. |
 | `modelContextWindows?` | `Record<string, number>` | Per-model context fallbacks/caps. These override `contextWindow`: an unknown window uses the configured value, while smaller live metadata remains authoritative. |
 | `modelInputModalities?` | `Record<string, string[]>` | Per-model input hints such as `["text"]` or `["text", "image"]`. |
--- a/docs-site/src/content/docs/reference/configuration/providers.md
+++ b/docs-site/src/content/docs/reference/configuration/providers.md
@@ -82,7 +82,7 @@ differing backup and rewrites known legacy namespaced selected ids to bare ids.
 | `models?` | `string[]` | Seed/fallback model list. With `liveModels: false`, these are the only discovered models. |
 | `liveModels?` | `boolean` | Fetch the live catalog on start/sync (default `true`). Custom providers use `${baseUrl}/models`; built-ins may use a registry URL and filter. |
 | `selectedModels?` | `string[]` | Catalog allowlist after discovery. Non-empty exposes only those ids; empty or omitted exposes all discovered models. |
-| `retainModels?` | `string[]` | Models to retain in the authoritative live catalog even when upstream discovery omits them (e.g. unprovisioned accounts or unlisted models). |
+| `retainModels?` | `string[]` | Provider-level allowlist of model IDs to retain when live discovery omits them (for example, unprovisioned accounts or unlisted models). Configured IDs not listed here continue to be pruned when omitted by discovery. |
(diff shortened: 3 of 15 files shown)


lidge-jun:
## 리뷰 · 우선순위 58 / 80

#1690 구현임. 라이브 디스커버리가 빼도 `retainModels`에 있으면 카탈로그에 남김. 메인터가 유지한 이슈. draft. 4박스 아직 비어 있음.

핵심은 `mergeConfiguredModelsIntoLiveCatalog` (`src/codex/catalog/provider-fetch.ts`). `providerRetainModels.has(id)` 추가. 기본 prune은 그대로. `configuredIds`에 `retainModels`를 합쳐서 시드함. 스키마는 `src/config.ts` `providerConfigSchema`, 타입은 `src/types/provider.ts`. 스플릿 이후 파일 맞음. 메가파일 되돌리는 패치 아님. MERGEABLE on `dev`.

테스트 `tests/provider-retain-models.test.ts`가 유지/드롭/중복/빈 라이브를 봄. `unrelated-model`은 드롭. 빈 `retainModels`는 지금이랑 같음. 리네임 마이그레이션 리스트에 `retainModels` 넣은 건 맞음.

빠진 건 1690 수용의 404 쪽. retain된 모델을 불렀을 때 한 번 경고하는 진단이 없음. 피커에 있는데 호출이 죽으면 또 같은 혼란임. `CALLABLE_CONFIGURED_COMPATIBILITY_MODELS`에 antigravity `gemini-3.7-flash` 시드는 없어도 됨. 설정이 그 역할.

초안이라 머지 열차 아님. CI/로컬 4박스 채우고 와. 동작은 방향 맞음.

해결방안: 호출 404 원샷 경고만 더하면 #1690 닫을 수 있음. 기본 동작 바꾸지 말 것.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
The feature direction is valuable, and the latest commits address the Grok/CodeRabbit requests for whitespace validation, Cursor retention, and a retained-model dispatch diagnostic. The focused suite passes 8/8 and typecheck passes on `d5e2dddae`.

Two blockers remain before this can leave draft:

1. The new diagnostic provenance can become stale. `withConfiguredRetention` only calls `retainedWithoutDiscoveryRefs.set(...)` when `retainedConfiguredIds.length > 0`; it never deletes the provider entry when a later live roster confirms the model or when `retainModels` is removed. `reconcileProviderFetchWarnings` also clears only `lastDropWarnSignature`, not either new map. A later unrelated 404 for that provider/model can therefore emit a false “retained via retainModels but omitted by discovery” warning. Replace the provider entry on every merge (`set` for nonempty, `delete` for empty) and reconcile both provenance/warning state across config generations. Add a gather-twice regression: first omitted+retained, then live-confirmed or retention-removed, followed by a 404 warning check.

2. This head is currently 157 commits behind `origin/dev` and has only intake checks; it is still draft with 0/4 readiness boxes. Catalog gather/cache boundaries have changed substantially in that range, so the current green focused test is not exact-head evidence.

Please rebase onto the current `dev`, fix the provenance lifecycle, then run the focused suite plus full typecheck/test/privacy and exact-head CI before requesting approval.

Ingwannu (changes requested):
Re-reviewed exact head `0baaf8c8e7c387e7f81e34a691749afa4edffeeb`.

The new commits correctly replace the live-merge provenance entry (`set` for retained ids, `delete` when live discovery confirms them) and route Cursor retention through the shared merge helper. The focused suite passes 9/9 and typecheck passes on this head.

One lifecycle blocker from the previous review remains. `reconcileProviderFetchWarnings()` still clears only `lastDropWarnSignature`; it does not clear or reconcile `retainedWithoutDiscoveryRefs` or `warnedRetained404Refs`. Several catalog paths also return before `withConfiguredRetention()` can replace the provider entry: `liveModels === false`, OAuth without a usable token, and Cloud Code Assist without a project. Therefore this sequence still produces false diagnostics:

1. a live gather omits model X and `retainModels` records X as retained;
2. config changes remove retention, switch to static mode, or lose the OAuth/project prerequisite;
3. generation reconciliation and the early-return gather leave the old provenance alive;
4. a later unrelated 404 for provider/X emits the stale `retainModels` warning, and the one-shot memo can also suppress a future legitimate warning after reconfiguration.

Please make the retained provenance replacement/clear operation part of every provider catalog return boundary, or reconcile both provenance maps against each config generation. Add a gather-twice regression covering omitted+retained followed by retention removal/static or logged-out fallback, then assert that `warnRetainedModel404Once` stays silent; also prove a later legitimate retained generation can warn again.

This branch is still 157 commits behind current `dev`, remains draft, and `enforce-target` is red. Rebase after the lifecycle fix and rerun exact-head CI before requesting approval.

Ingwannu (changes requested):
Re-reviewed exact head `ba9fc9fa821e7d0d1a259865e937cca2233f73a0`. The generation-reconciliation change fixes the older stale-map case, but the current implementation still loses the provenance during an ordinary successful gather, so this cannot leave draft yet.

The blocking sequence is in `fetchProviderModelsWithAuth`:

1. the first `withConfiguredRetention(live, { retainComboTargets: false })` pass sees a configured `retainModels` ID missing from raw discovery and records it in `retainedWithoutDiscoveryRefs`;
2. the next `withConfiguredRetention(forCache, { warnDrops: true })` pass receives a catalog that already contains that retained row;
3. `mergeConfiguredModelsIntoLiveCatalog` therefore returns no `retainedConfiguredIds`, and the current `else` branch deletes the provider entry.

Cursor, Antigravity, and cached return paths have the same multi-pass shape. A later upstream `404`/`model_not_found` therefore has no retained provenance and the promised one-shot diagnostic is silently skipped. Keep the first raw-discovery result as generation-scoped metadata, or otherwise prevent later cache/return normalization passes from erasing it. Add a regression that executes both passes before calling `warnRetainedModel404Once`.

The endpoint integration is also incomplete and currently differs by wire:

- Chat Completions and native Chat warn for structured `model_not_found` but not a plain HTTP 404.
- Claude Messages warns for HTTP 404 but does not recognize structured `error.code === "model_not_found"`.
- Responses warns in the native passthrough failure branch, but the generic translated-adapter failure branch returns without calling the helper.

Please use one shared retained-model-unavailable predicate across these surfaces and add focused tests for both error forms plus one-warning-per-provider/model behavior. The current `reconcileProviderFetchWarnings` test only proves that the symbol is callable; it does not verify cleanup, suppression, or re-warning after a generation change.

Finally, keep the public contract synchronized: IDs not covered by `retainModels` or the existing Vertex-default, compatibility-model, and combo-target exceptions remain subject to pruning, and catalog visibility does not prove that upstream invocation is provisioned. Update the directly affected English pages and translated counterparts together.

This head is still Draft, has five unresolved review threads, and has no exact-head maintained CI beyond intake checks. After the provenance boundary, endpoint tests, documentation, and latest-`dev` rebase are complete, rerun focused tests, full typecheck/test/privacy, docs build, and exact-head CI before requesting re-review.

Ingwannu (changes requested):
Re-reviewed exact head `638637130bf8e91c85c30b1adf40c09d0bf50279`, including the owner/Grok and current CodeRabbit findings. The multi-pass deletion bug and plain-404 route coverage are improved. The focused suite passes 13/13, with typecheck, privacy scan, and `git diff --check` green.

Two provenance-ownership blockers remain:

1. Generation reconciliation still breaks the diagnostic on a warm cache. `reconcileProviderFetchWarnings()` clears `retainedWithoutDiscoveryRefs`, but every fresh-cache, stale-cache, cooldown and discovery-failure return calls `withConfiguredRetention()` without `recordRetainedDiagnostics`. I reproduced this exactly: a live gather omitted `retained-model` and recorded it, a newer generation reconcile cleared the map, then a second gather inside the 60-second TTL made no upstream request and left the provenance absent. The subsequent upstream 404 therefore emits no retained-model warning. Add a real gather -> generation reconcile -> warm-cache gather -> 404 regression and restore provenance from generation-owned cache metadata or recompute it on every return boundary.

2. The live paths mutate `retainedWithoutDiscoveryRefs` before `setCached(..., cacheGeneration)` proves that the result still owns the cache generation. If the provider/config/account changes while discovery is in flight, `setCached` can reject the stale writer after the old result has already replaced the diagnostic map. Commit catalog data and its provenance under the same generation guard; add a delayed stale-writer regression proving an old discovery cannot install or clear current diagnostic state.

Please also keep these maps private and expose narrow test helpers instead of exporting mutable production state solely for tests. This remains a draft with one readiness box open; do not approve it until the cache/provenance lifecycle is generation-consistent.

lidge-jun:
Closing as superseded by #3206 (`6a6efa928`), which lands #1690 on `dev`. Your design was the input for three pieces that shipped: folding `retainModels` into the configured seed so ids need not be repeated in `models`, the zod `retainModels` schema with `nonBlankStringArrayConfigError`, and adding the key to `MODEL_ID_LISTS` for rename migration. Thank you for that.

Not taken: the 404-time `warnRetainedModel404Once` diagnostic and `CatalogModel.retainedWithoutDiscovery`. They require module-level state and edits in four hot request handlers to repeat a message the upstream error body already carries; the existing dropped-ids diagnostic stays as the signal for ids that are not retained. The retention decision also lives in the one predicate the merge loop already consults (#2860's approach) rather than a second set beside it.

Issue #1062, opened 2026-08-05
[Feature] Google Antigravity multi-account management, pool health and quota UX


agentHits:
### Area

Authentication and account pool / Google Antigravity

## Goal

Provide a first-class Google Antigravity / Gemini multi-account experience built on top of the shared OAuth account-pool runtime.

This issue tracks the **Antigravity-specific account-management, quota, health, and user-interface surface**.

Generic account selection and automatic failover policy is tracked separately in #695 and should be implemented once in the shared OAuth account-pool layer rather than duplicated here.

## Current scope

This issue owns the provider-specific work required to make Google Antigravity accounts useful as a real multi-account pool:

* Antigravity multi-account management UX;
* Antigravity-specific account health integration;
* per-account quota/status presentation;
* pool-level Antigravity quota/status presentation;
* Antigravity management API/UI;
* wiring Antigravity accounts into the generic OAuth account pool.

It does **not** own the generic routing/failover algorithm.

## 1. Multi-account management

Users should be able to manage many Google Antigravity / Gemini accounts from one provider surface.

The UI/API should make it easy to:

* view all configured accounts;
* identify the currently assigned/active account where relevant;
* see account health;
* add or remove accounts;
* inspect quota/reset state;
* understand whether an account is ready, limited, cooling down, unavailable, or unknown.

The experience should remain practical with larger pools such as:

```text
10–40+ accounts
```

Account identity and health state must remain isolated between accounts.

One account's quota, cooldown, project metadata, or failure state must never be projected onto another account.

## 2. Pool-level health and quota visibility

Provide an aggregate view of the Antigravity account pool.

Where the available provider evidence makes it meaningful, show information such as:

```text
Ready accounts
Limited accounts
Cooling-down accounts
Unavailable accounts
Unknown-quota accounts
Upcoming reset windows
```

Aggregate values must not imply precision that the upstream data does not actually provide.

For example, an account with unavailable quota data must not silently contribute:

```text
0% used
```

or be counted as healthy capacity merely because its quota could not be fetched.

Unknown must remain distinguishable from known-good.

## 3. Per-account quota visibility

Expose quota/limit information for each individual Antigravity account.

At minimum, the provider-specific data path should be able to represent:

* account health;
* quota usage;
* reset time;
* partial quota information;
* temporarily unavailable quota information;
* model-family-specific quota state where upstream data exposes it.

A failure to fetch one account's quota must not invalidate:

* other accounts;
* the whole pool;
[...]

Issue #1060, opened 2026-08-05
[Feature]: display provider-reported subscription / plan renewal dates in Providers


agentHits:
### Area

Web Dashboard → Providers / account metadata

## Problem

OpenCodex currently exposes operational account information such as quota/rate-limit state and reset timing, but users may also need to know when a paid subscription, token plan, or account entitlement actually expires or renews.

These are different concepts.

In particular, this issue must **not** treat OAuth/access-token expiration as subscription expiration.

OAuth credentials may expire and refresh while the user's underlying subscription remains active. Displaying that timestamp as "subscription expires" or "plan renews" would be misleading.

## Goal

Where a provider exposes authoritative subscription or entitlement metadata, surface that information in the Providers dashboard.

Examples of useful provider-reported fields include:

```text
Plan renews: Aug 25, 2026
Subscription expires: Sep 1, 2026
Billing period ends: Aug 31, 2026
Entitlement valid until: Aug 20, 2026
```

The exact terminology should follow what the provider's source actually represents.

## Data-source requirement

A displayed subscription/plan date must come from a source that actually represents the provider's subscription, entitlement, billing period, or plan validity.

Acceptable sources may include:

* an authenticated provider account/subscription API;
* provider quota metadata that explicitly includes plan renewal or entitlement expiry;
* account metadata returned by an official provider endpoint;
* an equivalent provider-native source with documented semantics.

Do **not** infer a plan date from:

* OAuth access-token expiry;
* OAuth refresh-token expiry;
* API-key expiration;
* session expiration;
* quota reset time;
* rate-limit reset time;
* arbitrary credential-file timestamps.

If a provider does not expose trustworthy plan metadata, OpenCodex should simply omit this field for that provider.

## Provider capability

Support should be capability-driven rather than assuming that every provider exposes the same metadata.

A provider/account may expose zero or more concepts such as:

```text
quotaResetAt
billingPeriodEndsAt
planRenewsAt
subscriptionExpiresAt
entitlementExpiresAt
```

The implementation does not need to manufacture a single universal timestamp when the upstream semantics differ.

Where possible, normalize provider-native metadata into a small typed OpenCodex account/subscription metadata contract while preserving the meaning of the source field.

## Dashboard behavior

When authoritative metadata exists, show it alongside the corresponding provider/account in the Providers dashboard.

Examples:

```text
Plan renews Aug 25, 2026
Subscription expires Sep 1, 2026
Billing period ends in 12 days
```

Requirements:

[...]

PR #2123, opened 2026-08-19
feat(quota): add per-account Gem/Cla quota probing for Google Antigravity (#1082)

### LINKED ISSUE #1082: [Feature]: add per-account Gem/Cla quota display for Google Antigravity
### Area

Authentication and account pool / provider quota UI

## Current status

This request is **partially implemented**.

Current `dev` already contains a provider-level Google Antigravity quota probe.

For the currently active Antigravity credential, OpenCodex:

* calls the authenticated Cloud Code Assist `fetchAvailableModels` endpoint;
* classifies returned quota information into Gemini (`Gem`) and Claude (`Cla`) families;
* calculates used percentages;
* preserves reset timestamps when the upstream provides them;
* publishes the results through the normal provider quota surface.

Conceptually, the current result can contain:

```text
Gem
  used %
  reset time

Cla
  used %
  reset time
```

The remaining gap is **per-account quota visibility** for users with multiple Google Antigravity accounts.

## What remains missing

OpenCodex has a separate per-account quota path used by the account-management UI/API.

At present that path supports:

```text
anthropic
```

but not:

```text
google-antigravity
```

As a result, the active Antigravity provider can have useful `Gem` / `Cla` quota data while individual accounts in the Antigravity account list still lack their own quota rows.

This issue now tracks that missing per-account integration only.

Generic account selection, routing policy, automatic failover, and aggregated account-pool architecture belong to their parent issues and should not expand this scope.

## Desired behavior

For every stored Google Antigravity account, OpenCodex should be able to fetch that account's own quota using:

* that account's OAuth credential;
* that account's Cloud Code Assist project ID;
* the same quota classification semantics already used by the provider-level Antigravity probe.

The result should be projected through the existing per-account quota contract.

For example:

```text
Google Antigravity accounts

account A
  Gem   14% used · resets 20:15
  Cla   62% used · resets 18:42

account B
  Gem   3% used · resets 20:14
  Cla   9% used · resets 18:40
```

The exact presentation should follow the existing account quota UI.

### 1. Extend per-account quota support

The per-account quota capability should recognize:

```text
google-antigravity
```

in addition to the providers already supported.

### 2. Use the selected account's credential

Do not implement the per-account path by calling the current active-provider helper.

Each row must be probed using the credential belonging to that specific account.

Otherwise every account could incorrectly display the active account's quota.

### 3. Use the account's project ID

The Antigravity quota RPC requires the Cloud Code Assist project associated with the account.

The per-account probe therefore needs the same account-scoped:

```text
access token
projectId
```

pair.

If an account does not have a usable project ID or valid access token, its quota should be reported as unavailable rather than borrowing another account's state.

### 4. Reuse the existing Gem / Cla classification

Do not create a second interpretation of Antigravity quota data.

The per-account path should use the same family classification and percentage/reset normalization as the existing provider-level probe.

Expected window labels remain:

```text
Gem
Cla
```

### 5. Handle partial upstream data

The upstream may expose:

* both Gem and Cla;
* only Gem;
* only Cla;
* a window without a reset timestamp.

All of those must be accepted.

Missing one family must not make the entire account quota probe fail.

### 6. Preserve account isolation

Quota cache entries must remain account-scoped.

Switching accounts, refreshing credentials, deleting an account, or replacing its identity must not allow quota data from one Antigravity account to appear under another.

Do not expose:

* refresh tokens;
* access tokens;
* project IDs;
* raw upstream quota payloads;

through the normal Dashboard/API result.

## Provider-level behavior

The existing provider-level Antigravity quota path should remain functional.

This issue does **not** require removing or redesigning it.

The desired end state is:

```text
provider overview
  -> current/active Antigravity quota

account management
  -> quota for each individual Antigravity account
```

Both should use the same quota semantics.

### #1062

Tracks broader Google Antigravity account-pool and management UX.

This issue remains the focused quota/status-display child.

### #695

Tracks generic OAuth account-pool selection/failover architecture.

Automatic account selection and generic failover policy are out of scope here.

### #1076

Tracks importing Antigravity accounts.

Account import is separate from displaying quota for accounts that already exist.

## Acceptance criteria

* [x] Active/provider-level Antigravity quota probe exists.
* [x] Provider-level probe reports Gemini (`Gem`) quota.
* [x] Provider-level probe reports Claude (`Cla`) quota.
* [x] Provider-level quota preserves reset timestamps when available.
* [ ] `google-antigravity` is supported by the per-account quota path.
* [ ] Each account is probed with its own access token.
* [ ] Each account uses its own Cloud Code Assist `projectId`.
* [ ] Gem and Cla percentages are visible per account.
* [ ] Gem and Cla reset times are visible per account when supplied.
* [ ] Partial Gem-only / Cla-only responses work.
* [ ] A failed account probe does not break other accounts or normal routing.
* [ ] Quota cache state cannot cross account identities.
* [ ] No credential/project identifiers are exposed in normal API/UI output.
* [ ] Focused tests cover two different Antigravity accounts returning different quota values.

### Checks

* [x] Current `dev` quota implementation was re-audited.
* [x] Already-delivered provider-level Gem/Cla support is removed from remaining scope.
* [x] The remaining per-account boundary is explicitly identified.
* [ ] Per-account Google Antigravity quota support implemented.

chilung-cgu:
## Summary

- Enable per-account quota probing for Google Antigravity in multi-account setups (resolves #1082).
- Extend `supportsPerAccountQuota` to include `google-antigravity`.
- Extract shared `fetchAntigravityUsageQuota` in `src/providers/quota.ts` accepting arbitrary valid access tokens and project IDs, enabling per-account Gem/Cla 5-hour and daily quota tracking in Web GUI without frontend alterations.
- Bind account quota cache & in-flight promises to normalized destination, enforce pre-token destination policy checks, and require fail-closed redirect handling (`redirect: "error"`).

## Verification

- Added regression tests in `tests/provider-account-quota.test.ts` verifying:
  - `supportsPerAccountQuota("google-antigravity")` returns true.
  - Account quota probing correctly issues requests to `/v1internal:fetchAvailableModels` with account-bound `projectId` and token.
  - Graceful fallback and negative-caching when account credentials lack `projectId`.
  - Destination-bound cache isolation and stale writer rejection across `baseUrl` changes.
  - Pre-flight destination policy gate preventing token transmission to invalid/metadata endpoints.
  - Fail-closed redirect defense rejecting 302s without token leakage.
- Verified cleanly rebased on latest `origin/dev` with strict typechecking and privacy scan passed.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added per-account quota tracking for Google Antigravity.
  * Added support for custom provider destinations when checking usage.
  * Reported Gemini usage by account, project, and credential with labeled quota windows.

* **Bug Fixes**
  * Improved quota cache accuracy when accounts or destinations change.
  * Prevented stale results from returning after configuration updates.
  * Added safer handling for invalid destinations, redirects, private networks, missing projects, and unexpected responses.

* **Tests**
  * Added comprehensive coverage for Antigravity quota reporting, caching, reconciliation, and network-safety scenarios.

### DIFF
--- a/src/lib/state-store-registrations.ts
+++ b/src/lib/state-store-registrations.ts
@@ -37,7 +37,7 @@ import { sweepExpiredApiKeyCooldowns } from "../providers/key-failover";
 import { reconcileProviderRequestPacing } from "../providers/request-pacing";
 import { sweepAbandonedResponseStateTemps, sweepExpiredResponseStates } from "../responses/state";
 import { sweepExpiredAntigravityReplay } from "../adapters/google-antigravity-replay";
-import { reconcileProviderAccountQuotaRows } from "../providers/quota";
+import { listLiveProviderAccountQuotaKeys, reconcileProviderAccountQuotaRows } from "../providers/quota";
 import { reconcileRouterWarningMemos } from "../router";
 import type { OcxConfig } from "../types";
 import {
@@ -62,13 +62,15 @@ export function reconcileLiveStateStores() {
 export function buildGenerationContext(): GenerationContext {
   if (!liveServerConfig) throw new Error("live server config is not installed");
   const providerNames = new Set(Object.keys(liveServerConfig.providers));
+  const oauthAccountKeys = listLiveOAuthAccountKeys(providerNames);
   return {
     generation: 0,
     providerNames,
     comboIds: new Set(Object.keys(liveServerConfig.combos ?? {})),
     comboTargets: listLiveComboTargetKeys(liveServerConfig),
     codexAccountIds: listLiveCodexAccountIds(liveServerConfig),
-    oauthAccountKeys: listLiveOAuthAccountKeys(providerNames),
+    oauthAccountKeys,
+    providerAccountQuotaKeys: listLiveProviderAccountQuotaKeys(liveServerConfig.providers, oauthAccountKeys),
     configRoots: listLiveConfigOwnershipRoots(getConfigDir()),
   };
 }
--- a/src/lib/state-store-sweeper.ts
+++ b/src/lib/state-store-sweeper.ts
@@ -7,6 +7,7 @@ export interface GenerationContext {
   comboTargets: ReadonlySet<string>;
   codexAccountIds: ReadonlySet<string>;
   oauthAccountKeys: ReadonlySet<string>;
+  providerAccountQuotaKeys: ReadonlySet<string>;
   configRoots: ReadonlySet<string>;
 }
 
--- a/src/providers/quota.ts
+++ b/src/providers/quota.ts
@@ -1,7 +1,7 @@
+const originalFetch = globalThis.fetch;
 import { providerDestinationConfigError } from "../lib/destination-policy";
 import { providerOutboundPost, providerRedirectError } from "../lib/provider-outbound";
 
-const originalFetch = globalThis.fetch;
 import { createHash } from "node:crypto";
 import {
   effectiveCodexAuthAccountId,
@@ -1605,8 +1605,9 @@ async function getTokenForAccountQuotaProbe(provider: string, accountId: string)
 async function fetchAccountQuota(
   provider: string,
   accountId: string,
-  forceRefresh: boolean,
+  forceRefresh = false,
   baseUrl?: string,
+  allowPrivateNetwork?: boolean,
 ): Promise<AccountQuotaCacheEntry> {
   const normalizedDest = provider === "google-antigravity" ? normalizeAntigravityDestination(baseUrl) : "";
   const key = accountCacheKey(provider, accountId, normalizedDest);
@@ -1638,6 +1639,7 @@ async function fetchAccountQuota(
         // Pre-flight destination policy gate before acquiring or refreshing account token
         const destError = providerDestinationConfigError("google-antigravity", {
           baseUrl: baseUrl || "https://daily-cloudcode-pa.googleapis.com",
+          allowPrivateNetwork: allowPrivateNetwork ?? false,
         });
         if (destError) {
           const entry: AccountQuotaCacheEntry = { ts: Date.now(), quota: null, unavailable: true };
@@ -1648,7 +1650,7 @@ async function fetchAccountQuota(
           return entry;
         }
         const token = await getTokenForAccountQuotaProbe(provider, accountId);
-        quota = await fetchAntigravityUsageQuota(token, stored.projectId, baseUrl);
+        quota = await fetchAntigravityUsageQuota(token, stored.projectId, baseUrl, { allowPrivateNetwork });
       }
       if (!quota) {
         // Preserve last-good bars and mark unavailable; advance TTL so failures
@@ -1697,12 +1699,13 @@ export async function fetchProviderAccountQuotas(
   provider: string,
   forceRefresh = false,
   baseUrl?: string,
+  allowPrivateNetwork?: boolean,
 ): Promise<ProviderAccountQuota[]> {
   if (!supportsPerAccountQuota(provider)) return [];
   const set = getAccountSet(provider);
   if (!set) return [];
   return await Promise.all(set.accounts.map(async account => {
-    const entry = await fetchAccountQuota(provider, account.id, forceRefresh, baseUrl);
+    const entry = await fetchAccountQuota(provider, account.id, forceRefresh, baseUrl, allowPrivateNetwork);
     return {
       accountId: account.id,
       quota: entry.quota,
@@ -2291,7 +2294,7 @@ async function fetchAntigravityQuota(provider: string, config: OcxProviderConfig
   } catch {
     return null;
   }
-  const quota = await fetchAntigravityUsageQuota(accessToken, credential.projectId, config.baseUrl);
+  const quota = await fetchAntigravityUsageQuota(accessToken, credential.projectId, config.baseUrl, { allowPrivateNetwork: config.allowPrivateNetwork });
   if (!quota) return null;
   if (probedAccountId && probedAccountKey) {
     const stillOwnsToken = getAccountCredential("google-antigravity", probedAccountId)?.access === accessToken;
--- a/src/providers/quota.ts
+++ b/src/providers/quota.ts
@@ -1436,19 +1436,13 @@ type AccountQuotaCacheEntry = {
 const accountQuotaCache = new Map<string, AccountQuotaCacheEntry>();
 const accountQuotaInflight = new Map<string, Promise<AccountQuotaCacheEntry>>();
 let lastReconciledGeneration = 0;
-let liveAccountQuotaKeys = new Set<string>();
+let liveAccountQuotaKeys = new Map<string, number>();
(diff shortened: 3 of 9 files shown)


lidge-jun:
## 리뷰 · 우선순위 56 / 80

#1082. Antigravity 계정마다 Gem/Cla 쿼타를 보게. 지금은 `supportsPerAccountQuota`가 anthropic만. 프로바이더 전체 프로브는 `getCredential` 한 장만 씀. draft. 4박스 비어 있음.

`fetchAntigravityUsageQuota(token, projectId, baseUrl)`로 빼서 계정 루프에서 `getAccountCredential` + `getTokenForAccountQuotaProbe` 씀. `projectId` 없으면 throw → 기존 negative-cache. 구독 만료일(#1060)이 아님. 쿼타 윈도우만. MERGEABLE on `dev`. 스플릿이랑 안 겹침.

테스트가 계정 두 개, project/Bearer가 맞는지 봄. `remainingFraction` → used %. GUI 변경 없음. 대시보드가 `customWindows` 이미 그리면 그대로 뜸.

`loadConfig()`를 프로브 한 번에 한 번씩 부르는 건 거침. `fetchAccountQuota`가 이미 provider config를 알면 그걸 넘기면 됨. 토큰은 로그에 안 넣는지 한 번 더 보면 됨. 테스트는 mock fetch라 실호출 없음.

초안. #1062 멀티계정 UX랑 같이 가면 좋음. 지금 머지 열차는 아님.

해결방안: `loadConfig()` 빼고 호출 쪽 config.baseUrl 쓰기. projectId 없는 계정은 throw 말고 조용히 skip이 더 맞을 수도.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
The latest commits correctly address the Grok/CodeRabbit requests to thread the configured base URL, verify per-account bearer isolation, and skip projectId-less accounts without a false unavailable state. On `d85b12078`, the focused suite passes 18/18, typecheck passes, and privacy scan passes.

Two blockers remain:

1. The per-account cache/in-flight identity is only `provider + accountId`; it does not include the Antigravity destination. After a provider base URL change, a normal non-forced read returns quota captured from the old destination and does not probe the new one. I reproduced this directly: a forced probe to `https://first.example` returned 10%, then a non-forced probe for the same account at `https://second.example` made no second request and returned the first destination's 10% row. Bind cache/in-flight/writer identity to a normalized destination (or invalidate and reject stale writers on destination changes), and add a regression for this transition.

2. This expands a configurable OAuth destination from the active Antigravity credential to every stored account credential. The old provider-level path already trusted `config.baseUrl`, so this is not a new arbitrary-host privilege by itself, but it materially expands the credential blast radius. Before `maintainer-sponsored`, validate the destination through the repository's provider-outbound/destination policy before acquiring account tokens and use fail-closed redirect handling. Add a test proving a rejected/redirecting destination receives no bearer token.

The branch is also 157 commits behind current `origin/dev`, remains draft with 0/4 readiness boxes, and has no exact-head full CI. Please rebase after fixing the destination-bound cache and credential boundary, then rerun full required CI before requesting approval.

Ingwannu (changes requested):
Re-reviewed exact head `59228b4d1e8187ccef2b0dcdb40defd353dfdc15`.

The new percentage-clamping and provider-report cache-seeding tests pass; the focused suite is 20/20, typecheck passes, and privacy scan passes. However, this commit does not resolve the two destination/cache blockers from the previous review and adds a stale-writer variant of the same problem.

1. `accountCacheKey()` and `accountQuotaInflight` are still keyed only by provider + accountId. A base URL change can reuse a fresh row or join a probe created for the previous Antigravity destination. The new provider-report seeding path compounds this: `fetchAntigravityQuota()` captures the config generation only after the outbound await (`mayCommitAccountQuotaKey(..., captureConfigGeneration())`). A request started against destination A can therefore finish after a config change to destination B, capture the new generation, and publish A's quota into B's account cache as long as the token is unchanged. Capture generation and normalized destination before acquiring/sending the credential, include destination in cache/in-flight ownership, and reject stale writers.

2. `fetchAntigravityUsageQuota()` still sends the bearer with a direct `fetch` to configurable `baseUrl` and follows redirects by default. It does not pass through the repository provider-outbound destination policy or set fail-closed redirect handling. This expands the custom destination across every stored Antigravity account and is correctly still blocked by `unsponsored_surface`. Validate the final destination before token acquisition/use and prove that a rejected or redirecting target receives no bearer request.

Please add regressions for a baseUrl A -> B transition (including an A probe resolving after the change) and for blocked/redirecting destinations. The branch is also 157 commits behind current `dev`, remains draft, and both hygiene and enforce-target are red; do not request sponsorship until these credential boundaries are closed on a rebased exact head.

Ingwannu (changes requested):
Re-reviewed exact head `e65beba0d4a9b2abf4af140e0322cd5f44a1ff05`, including the owner/Grok feedback. The destination-specific cache/in-flight split and `redirect: "error"` are useful progress. The focused quota suite passes 24/24, typecheck passes, and `git diff --check` passes.

Two blockers remain:

1. The new Antigravity cache keys are `provider\0accountId\0destination`, but generation reconciliation still compares every cached key with `context.oauthAccountKeys`, whose canonical format from `listLiveOAuthAccountKeys()` is only `provider\0accountId`. Therefore every destination-qualified Antigravity row is considered non-live and removed on the next generation reconciliation, and `liveAccountQuotaKeys` can never authorize an older destination-qualified writer. Update reconciliation/liveness ownership so the provider/account prefix and the currently configured normalized destination agree, then add a real warm-cache regression that performs a generation reconcile between reads.

2. `fetchAntigravityUsageQuota()` still sends every stored account bearer through plain `fetch` after only the synchronous literal/config check. `redirect: "error"` blocks HTTP redirects, but it does not resolve and pin the hostname, so DNS rebinding/private resolution remains outside the repository provider-outbound trust boundary. Route this POST through the shared resolved/pinned provider outbound transport (or an equivalent reviewed primitive) and prove that a hostname resolving to metadata/private space receives no bearer token. The current mocked redirect test does not cover that boundary.

Also remove the unrelated OpenCode Go registry-routing change from this #1082 slice; it has its own behavior and test surface.

Please keep this draft and do not request `maintainer-sponsored` until these ownership and credential-destination boundaries are fixed on a rebased exact head.

lidge-jun:
Closing as superseded by #3213 (`ef7b3c9cf`), which lands #1082 on `dev`. Your account loop and the reuse of `getTokenForAccountQuotaProbe` are what shipped, and thank you for carrying this through three review rounds. The two blockers that kept coming back — destination-bound cache identity and routing every stored bearer through the pinned outbound transport — were resolved differently: the per-account probe always targets Google's own Cloud Code Assist host rather than the configurable `baseUrl`, so there is no destination dimension to key the cache on, and it goes through `providerOutboundPost` so a blocked or redirecting destination never sees a token. That removed the need for the destination-qualified keys and the reconciliation changes, and the unrelated OpenCode Go routing change the reviewer flagged is not part of it.

Issue #2114, opened 2026-08-19
[Bug][Linux] native-main 503 when systemctl exists but the user bus is unreachable (#1612 regression gap)


lidge-jun:
### Client or integration

Codex CLI

### Area

Service lifecycle

### Summary

OpenCodex 2.26.0 blocks every native OpenAI request on a systemd-less Linux container **when `systemctl` is present on PATH but the user bus is unreachable**. The proxy stays live, `/healthz` returns 200, and `ocx ready` reports `{"ready":true,"status":"ready"}`, but every native request fails locally with HTTP 503 before reaching upstream:

```json
{"error":{"message":"OpenCodex local native-main profile maintenance is active; retry this request","type":"server_error","code":"server_is_overloaded"}}
```

This is the same admission fence as #1612, but that fix does not cover this environment. #1612 was closed by treating a **systemctl spawn failure** as `absent`. In this container `systemctl` spawns fine — it exits non-zero because there is no user bus:

```
$ command -v systemctl
/usr/bin/systemctl

$ systemctl --user show opencodex-proxy -p LoadState -p ActiveState -p FragmentPath -p NeedDaemonReload
Failed to connect to user scope bus via local transport: $DBUS_SESSION_BUS_ADDRESS and $XDG_RUNTIME_DIR not defined
EXIT=1
```

`inspectSystemd()` only maps `spawnFailed` to `absent`; a non-zero exit still returns `unknown`, which re-triggers the native-main block. The `/.dockerenv` marker is present and PID 1 is `tini`, so this is exactly the externally-supervised foreground shape `src/service.ts` recommends (`Service: unsupported in Docker`, "run `ocx start` directly").

### Reproduction

1. Debian 13 container (PID 1 = `tini`, `/.dockerenv` present) with `systemd` **installed** but no user bus: `/run/user/$UID` empty, `XDG_RUNTIME_DIR` and `DBUS_SESSION_BUS_ADDRESS` unset.
2. `npm install -g @openai/codex@0.148.0 @bitkyc08/opencodex@2.26.0`
3. `codex login` (ChatGPT), confirm native `codex exec` works.
4. `ocx start`, then `ocx sync`.
5. Send a native request:

```
$ curl -s -X POST http://127.0.0.1:10100/v1/responses \
    -H 'Content-Type: application/json' \
    -d '{"model":"gpt-5.5","input":"hi","stream":false}'
[...]

PR #2130, opened 2026-08-19
fix(probe,codex): stop fencing native-main on a probe that could not answer

lidge-jun:
## Summary

Both of the native-main 503 reports come from the same shape: a probe that could not get an answer, a verdict of `unknown`, and a fence that holds for the entire process. The user-visible result is that every native request returns 503 and only `ocx restart` clears it.

Three changes, each with its own guard.

**#2114 — Linux, the session bus is unreachable.** `inspectSystemd` called every non-zero exit `unknown`. Its own comment was right that a non-zero status means the question never reached the bus — and that is exactly why the verdict was wrong. It is evidence about the bus, not evidence that a foreign service owns this home.

Widening on the exit code alone would fail open: with the bus down `systemctl` cannot see a foreign unit either, so "no answer" would read as "no owner" on a machine that genuinely has one. Instead the classification asks the disk, which needs no bus. A unit file is proof of installation, and the homes it names are what ownership is actually decided on. No unit file means absent; a unit naming a foreign home stays present and still blocks. Registration is reported `absent` on that path rather than invented — the disk cannot say whether systemd has the unit loaded, and guessing is how a stale claim would slip through.

**#2108 (a) — Windows, an unaskable WinSW query.** WinSW is an optional backend, and a scheduler-only install has neither its XML nor its exe on disk. A timed-out `sc.exe query` still returned `unknown`, which outranks the disk and fences the process. A query we could not ask is a question about a service that cannot exist. The disk outranks it only when **both** assets are gone; either one present keeps the old `unknown`.

**#2108 (b) — the fence itself re-asks.** `startServer` takes the ownership verdict once and holds it for the process lifetime, which is the reason waiting never helped and restart always did. That is correct for `foreign-ownership` — a foreign owner is a fact, and re-asking would hand a determined caller a second chance at a boundary built to refuse it. It is wrong for `ownership-unknown`, which says the probe could not answer, not that this host is unownable. An unknown fence now carries a reprobe hook and lifts itself when the host becomes answerable: never for foreign, on request arrival rather than on a timer so an idle proxy does no work, and capped so a permanently unaskable host stops asking. A fence built without the hook behaves exactly as before.

**Known limitation, stated rather than hidden:** systemd localizes the bus-failure strings, so a non-English host will not match and keeps the old `unknown`. That fences rather than admits, which is the safe direction, but it does mean the #2114 fix does not reach every affected user. Forcing `LC_ALL=C` on the probe would remove the caveat and is the obvious follow-up; it is not done here because it changes every `systemctl` call the module makes, not just this branch.

**Coordination with #2029.** That draft PR edits the same function for #1939 and classifies two other bus messages as `absent`. This branch does not touch it or its branch. The two changes agree in direction; if #2029 lands first this reconciles with it rather than replacing it.

Refs #2114, Refs #2108.

## Verification

```
bun test tests/codex-service-manager-probe.test.ts        56 pass / 0 fail
bun test tests/native-profile-startup.test.ts + probe     73 pass / 0 fail
bun test core-lab-boundary, native-main-owner-lifetime,
         native-profile-drain-server, service-probe-docker,
         server-auth                                     106 pass / 0 fail
bun x tsc --noEmit                                        exit 0
bun run privacy:scan                                      exit 0
```

Every assertion was **driven red first**. The #2108 (a) branch was additionally **ablated** after going green — reverting just that condition puts the assertion back to failing, so the test is bound to the new code rather than passing incidentally.

`core-lab-boundary` is in that list deliberately: this adds an import to `native-profile-startup`, and that test walks the runtime import graph to prove the core request path still cannot reach `src/lab`.

The pre-existing assertion that pinned the #2114 shape as `unknown` is **amended to a non-bus stderr, not deleted**, so the rule it protects still holds.

Windows and Linux platform paths are what these fix, and neither macOS nor the Linux shards prove them — the platform legs are the check that matters here.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Native startup can re-check uncertain service ownership and unblock without a restart.
  - Retries are limited to uncertain ownership; confirmed foreign ownership remains blocked.
  - Added offline service detection for systemd when the session bus is unavailable.
  - Added Windows service fallback detection for missing service assets.

- **Bug Fixes**
  - Improved classification of unavailable, present, ambiguous, and indeterminate services.

- **Tests**
  - Expanded coverage for ownership retries, retry limits, offline detection, and Windows service timeouts.

### DIFF
--- a/devlog/_plan/260819_unclaimed_bug_selection/170_2114_2108_fences.md
+++ b/devlog/_plan/260819_unclaimed_bug_selection/170_2114_2108_fences.md
@@ -1,0 +1,103 @@
+# 170 — #2114 and #2108: the fence that could not re-ask
+
+Both reports are the same shape wearing two operating systems: a probe that
+could not get an answer, a verdict of `unknown`, and a fence that holds for the
+whole process. Every native request 503s and only `ocx restart` clears it.
+
+Shipped as PR #2130, five commits.
+
+## The three changes
+
+**#2114 — the Linux session bus.** `inspectSystemd` called every non-zero exit
+`unknown`. Its own comment was right that a non-zero status means the question
+never reached the bus — and that is exactly why the verdict was wrong. It is
+evidence about the bus, not about who owns this home.
+
+Widening on the exit code alone fails open: with the bus down `systemctl`
+cannot see a foreign unit either, so "no answer" would read as "no owner" on a
+machine that has one. The classification asks the disk instead, which needs no
+bus.
+
+**#2108 (a) — an unaskable WinSW query.** WinSW is optional, and a
+scheduler-only install has neither its XML nor its exe on disk. A timed-out
+`sc.exe query` still returned `unknown`, which outranks the disk. With both
+assets gone there is nothing for a registration to belong to.
+
+**#2108 (b) — the fence re-asks.** `startServer` takes the verdict once and
+holds it, which is why waiting never helped and restart always did. Correct for
+`foreign-ownership` — a foreign owner is a fact. Wrong for
+`ownership-unknown`, which says the probe could not answer.
+
+## What the audit caught, three rounds running
+
+This is the part worth keeping. The functional idea was right in round one; the
+**ownership model was wrong three times in a row**, and each time the symptom
+was identical — a fence with no way to re-ask, which is #2108 returning by
+another route.
+
+| Round | Defect | Why it slipped |
+|---|---|---|
+| 1 | `inspectSystemdOffline` checked ONE path, so a foreign unit in `~/.local/share/systemd/user` or an XDG override was invisible → `ownership: owned` on a foreign host | The bus-up path never had the hole; I reused its constant, not its coverage |
+| 2 | The probe decremented the refcount AND the fence's own `release()` decremented again → a fence another holder owned got lifted | `Math.max(0, ...)` floored it, so it under-counted silently instead of going negative |
+| 2 | A spent hook stayed in the map, so no LATER fence could install one | The "do not reset the budget" guard was right; its scope was not |
+| 3 | Owner-identity guard was correct but **untested** — replacing it with `if (true)` left every suite green | A third route to the same wedge, with nothing pinning it |
+
+Round 1's fail-open is the one that mattered: I wrote "a unit naming a foreign
+home stays present and still blocks" in a commit message, and the auditor proved
+it end-to-end as `ownership = owned`. The claim was true only for a unit at the
+canonical path.
+
+The fix that finally held is smaller than either attempt: the hook delegates to
+the fence's own idempotent `release()`. One fence, one payment, whoever triggers
+it. Both earlier versions were modelling the ownership in the wrong place.
+
+## Guards, and the ones that were not guards
+
+Every branch was ablated. Three that looked like guards were not:
+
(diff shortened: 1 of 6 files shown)


lidge-jun:
## 리뷰 · 우선순위 74 / 80

이게 #2108/#2114의 진짜 수리임. #2121은 로그만 찍었음. 이번엔 프로브가 답을 못 하면 `unknown`으로 프로세스 평생 펜스 치는 계약을 자름. `dev`는 `ceac592d7`이고 2.27 찍기 직전이라 막판 머지 후보로는 무거움. 그래도 점수 74임. native GPT 503이 `ocx restart` 전에는 안 풀리던 그 불임.

리눅스(#2114)는 `inspectSystemd`가 non-zero를 전부 `unknown`으로 보던 거임. 버스가 안 되면 질문이 버스에 도착을 안 한 거지, 남의 유닛이 이 홈을 먹었다는 증거가 아님. exit만 넓히면 fail-open임. 버스 다운이면 남의 유닛도 안 보이니까. 그래서 디스크를 봄. 유닛 파일이 설치 증거고, 거기 적힌 홈이 오너십임. 검색 경로가 `~/.config/systemd/user` 하나면 감사에서 fail-open 났음. 지금은 systemd가 보는 디렉토리 전부 + `XDG_CONFIG_HOME`. 유닛 두 개면 `unknown`으로 거절. 등록 상태는 디스크가 로드 여부를 몰라서 `absent`로 정직하게 둠.

윈도우(#2108a)는 스케줄러-only 설치에 WinSW XML/exe가 없는데 `sc.exe query` 타임아웃이 `unknown`을 이김. 둘 다 없을 때만 디스크가 이김. 하나라도 있으면 예전 `unknown`. 펜스(#2108b)는 `startServer`가 부팅 때 한 번 보고 평생 들고 있던 거임. `foreign-ownership`은 사실이라 재질문 안 함. `ownership-unknown`만 훅 달고, 요청/가디언이 물을 때 `spawnSync`로 다시 봄. 캡 5회. 타이머 아님.

감사 세 라운드가 남긴 구멍이 다 코드에 보임. 훅이 refcount를 직접 깎으면 더블 디크리먼트. 지금은 `lifecycle.release()`에 위임. spent 훅이 맵에 남으면 다음 서버가 훅을 못 달고 또 `ocx restart`. 오너 펜스의 release만 엔트리를 지움. `isNativeMainTrafficBlocked`가 요청 경로에서 프로브를 돌리니까 캡이  pretence가 아님. 2초 타임아웃 × 5.

안 닫는 것도 맞음. `Refs #2114, Refs #2108`. 로케일 매칭은 영어 stderr만임. 비영어 호스트는 예전 `unknown`으로 펜스. `LC_ALL=C`는 후속. 드래프트 #2029가 같은 `inspectSystemd`를 #1939용으로 고침. 이 PR이 그걸 안 덮음. 방향은 같음. 먼저 들어가면 reconcile. types.ts/config.ts 안 건드림. 닫고 리베이스하라는 쪽 아님.

해결방안: 2.27 태그 전이면 크로스 CI 초록만 보고 예외로 넣어도 됨. 이미 찍었으면 2.27.1 첫 픽. #2029는 닫지 말고 이 브랜치에 맞출 것. `Closes`는 필드에서 재부팅/버스다운 한 번 보기 전엔 달지 말 것.

이 댓글은 grok-bot이 작성했습니다

PR #1941, opened 2026-08-18
fix(grok): switch to Responses backend and backfill required annotations

bet4it:
## Summary

This PR lets Grok CLI use the Responses passthrough path instead of being pinned to Chat Completions, and fixes two serialization errors that blocked that switch.

**Background.** Grok CLI was injected with `api_backend = "chat_completions"`. The reason was that opencodex's SSE keep-alive emitted `response.heartbeat` as a typed SSE event — but `response.heartbeat` is not a valid Responses API event variant. Grok-build (Grok's Rust client) uses strict enum deserialization via an async-openai fork, so any unknown variant causes a hard `serialization error: unknown variant response.heartbeat`. Pinning Grok to Chat Completions avoided the passthrough path where these frames surfaced, at the cost of a full chat↔responses translation layer on every Grok turn.

**What changed.**

1. **Heartbeat is now an SSE comment line** (`src/bridge.ts`). A line starting with `:` is discarded by every eventsource parser — it re-arms the idle timer without triggering deserialization. This is the correct keep-alive mechanism and matches what the passthrough relay already does.

2. **Grok config injects `api_backend = "responses"`** (`src/grok/inject.ts`). With heartbeats fixed, Grok can finally use the Responses passthrough path, giving it the same protocol fidelity Codex already has and removing the translation layer from the hot path.

3. **Backfill `annotations: []` on `output_text` content parts** (new `src/server/responses/responses-field-backfill.ts`). The Responses spec defines `annotations` as a required `Vec<Annotation>` on `OutputTextContent` (not `Option<Vec>`). Some upstream relays (e.g. sub2api) omit it when empty, which is spec-non-compliant. Strict clients like Grok-build's async-openai fork fail with `missing field annotations`. A new stateless SSE/JSON rewrite adds `annotations: []` wherever it is missing, on both the streaming and bounded-JSON passthrough paths. Existing values are always authoritative; only absent fields are added. The rewrite is unconditional and safe for all clients because the field is always valid on the wire.

4. **Surface detection for Grok on `/v1/responses`** (`src/server/index.ts`). The `x-opencodex-grok` header now sets `logCtx.surface = "grok"` on the Responses endpoint too, matching what the chat-completions handler already did. This restores Grok client identification in the Logs & Debug dashboard after the protocol switch.

5. **Test cleanup.** Stale comments referencing grok-build's decoder and the old chat_completions pin have been corrected.

## Verification

- `tsc --noEmit` passes (strict).
[...]

PR #2131, opened 2026-08-19
fix(responses): backfill missing id on output items for strict decoders

bet4it:
## Summary

Follow-up to #1941. That PR switched Grok to the Responses passthrough path and backfilled the required `annotations` field on `output_text` content parts. After it merged, Grok CLI still failed with ```serialization error: missing field `id` ``` — the same class of strict-decoder failure, but for a different required field.

**Root cause.** Some upstream relays (observed with an `openai-responses` adapter provider) omit the `id` field entirely on output items in `response.completed` — `message`, `reasoning`, and `function_call` items all arrive with no `id`. In the async-openai fork that grok-build uses, `OutputMessage.id` and `ReasoningItem.id` are required (`String`, not `Option`), so deserializing `response.completed` fails hard.

This was confirmed by capturing the raw upstream SSE payloads: `response.completed` arrives with output items like `{"type":"reasoning","summary":[...]}` (no `id`), `{"type":"message","role":"assistant","content":[...]}` (no `id`), and `{"type":"function_call","call_id":"call_...",...}` (only `call_id`, no `id`).

**What changed.**
1. **Backfill missing `id` on output items** (`src/server/responses/responses-field-backfill.ts`). The existing `backfillOutputItem` now also synthesizes an `id` when one is absent, using a deterministic scheme per (item type, output index) with canonical OpenAI prefixes: `msg_ocx_N`, `rs_ocx_N`, `fc_ocx_N`, etc. Existing ids are never overwritten. The same index is used across streaming events (`output_item.added`/`done` uses `event.output_index`) so an item's generated id stays stable throughout the stream. Both the SSE block rewrite path and the bounded-JSON passthrough path are covered, mirroring how `annotations` is already handled.

2. **Update the Grok Build guide** (all locales). The docs still described the Chat Completions transport and its inbound translator, but Grok is now on `api_backend = "responses"`. Updated the endpoint description to `POST /v1/responses`, fixed the managed-block example `api_backend` to `"responses"`, and rewrote the reasoning section to describe Responses passthrough of `reasoning.summary` instead of Chat Completions `reasoning_content` translation.

## Verification

- `tsc --noEmit` passes (strict).
- `bun test tests/responses-field-backfill.test.ts` — 10 pass, 0 fail (7 existing + 3 new: backfill missing ids on response.completed, preserve existing ids, use output_index on output_item.done).
- Rebased onto the latest `origin/dev` (`ceac592d7`).
- Captured raw upstream payloads confirmed the missing `id` field; after the fix, generated ids are deterministic and do not collide with real upstream ids.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

* **Documentation**
  * Updated Grok Build integration guides to reference the Responses API endpoint exclusively.
  * Clarified reasoning-summary configuration, including route defaults and hiding reasoning traces with `"none"`.
  * Updated guidance across supported translations.

* **Bug Fixes**
  * Responses output items without IDs now receive stable, type-specific identifiers.
  * Existing IDs and annotations remain unchanged across streaming and standard responses.

* **Tests**
  * Added coverage for generated IDs, preserved IDs, output indexes, inherited item types, and JSON responses.

### DIFF
--- a/docs-site/src/content/docs/fr/guides/grok-build.md
+++ b/docs-site/src/content/docs/fr/guides/grok-build.md
@@ -3,8 +3,8 @@ title: Grok Build
 description: Utilisez n’importe quel modèle routé par opencodex depuis la CLI Grok Build de xAI — les modèles sont automatiquement enregistrés dans ~/.grok/config.toml pendant l’exécution du proxy.
 ---
 
-opencodex expose un point de terminaison compatible OpenAI `POST /v1/chat/completions` (ainsi que `/v1/responses`) sur son
-port local, tandis que Grok Build prend en charge les modèles personnalisés hébergés sur des serveurs compatibles OpenAI. Avec
+opencodex expose un point de terminaison compatible OpenAI `POST /v1/responses` sur son port local,
+tandis que Grok Build prend en charge les modèles personnalisés hébergés sur des serveurs compatibles OpenAI. Avec
 cette intégration, opencodex enregistre automatiquement l’intégralité de son catalogue visible dans Grok Build :
 aucune modification manuelle de la configuration n’est nécessaire.
 
@@ -61,12 +61,11 @@ dans Codex. Ceux dont la liste de niveaux est vide n’affichent aucun contrôle
 de Codex. Les entrées GPT-5.6 natives sont distinctes : elles conservent et exposent leurs échelles de raisonnement
 en amont fixes, et non les métadonnées configurées pour les modèles routés.
 
-Grok Build communique avec opencodex au moyen de Chat Completions et envoie `reasoning_effort` lorsque
-l’échelle est annoncée. Dans ce cas, le traducteur Chat Completions entrant définit par défaut le champ Responses
-`reasoning.summary` sur `auto` ; les traces de raisonnement parviennent donc à Grok sous la forme
-`delta.reasoning_content` au lieu d’être masquées. Réglez `include_reasoning: false` (ou
-`reasoning.summary: "none"`) si un client souhaite que le modèle réfléchisse sans renvoyer le
-tracé. Une valeur explicite de `reasoning.summary` prévaut lorsque les deux options sont présentes.
+Grok Build communique avec opencodex au moyen de l’API Responses. Lorsque la route annonce une échelle
+de raisonnement, la passerelle Responses transmet `reasoning.summary` telle que configurée, de sorte que
+les traces de raisonnement parviennent nativement à Grok sous forme d’éléments de raisonnement Responses.
+Réglez `reasoning.summary: "none"` si un client souhaite que le modèle réfléchisse sans renvoyer le
+tracé. Une valeur explicite de `reasoning.summary` prévaut sur la valeur par défaut de la route.
 
 ## Note d'authentification
 
--- a/docs-site/src/content/docs/guides/grok-build.md
+++ b/docs-site/src/content/docs/guides/grok-build.md
@@ -3,10 +3,10 @@ title: Grok Build
 description: Use any opencodex-routed model from xAI's Grok Build CLI — models are auto-registered into ~/.grok/config.toml while the proxy runs.
 ---
 
-opencodex serves an OpenAI-compatible `POST /v1/chat/completions` (and `/v1/responses`) on its
-local port, and Grok Build supports custom models against OpenAI-compatible servers. Starting
-with this integration, opencodex registers its whole visible catalog into Grok Build
-automatically — no manual config editing required.
+opencodex serves an OpenAI-compatible `POST /v1/responses` on its local port, and Grok Build
+supports custom models against OpenAI-compatible servers. Starting with this integration,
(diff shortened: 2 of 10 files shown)


lidge-jun:
## 리뷰 · 우선순위 56 / 80

#1941 후속임. 그때는 `output_text.annotations`를 빈 배열로 채웠음. 이번엔 업스트림 릴레이가 `response.completed` / `output_item.added|done`에서 `id` 자체를 빼먹어서 grok-build serde가 `missing field id`로 죽음. message/reasoning은 `id`가 required `String`임. function_call은 `call_id`만 있고 `id`가 없음. 캡처한 raw SSE가 그 모양임.

고친 곳은 `src/server/responses/responses-field-backfill.ts` 한 파일임. `backfillItemId`가 빈 문자열/없음일 때만 `msg_ocx_N` `rs_ocx_N` `fc_ocx_N` 식으로 만듦. 있는 id는 안 덮음. 스트림은 `event.output_index`를 써서 같은 아이템이 added/done에서 같은 id를 유지함. JSON 패스스루(`backfillResponsesFieldJson`)도 같은 워크. types.ts/config.ts 안 건드림. `dev` 직접 MERGEABLE. 분할에 깔리는 쪽 아님.

합성 id의 경계는 이럼. prefix는 OpenAI 모양이고 가운데 `ocx_`가 진짜 업스트림 id랑 안 겹치게 함. `output_index`가 없거나 숫자가 아니면 0으로 떨어짐. 인덱스 없는 이벤트가 두 개면 `msg_ocx_0`이 중복될 수 있음. function_call은 `call_id`를 그대로 두고 `id`만 채움. 클라이언트가 `id`를 콜 식별자로 쓰면 `fc_ocx_N`이랑 `call_...`가 갈라짐. 테스트가 completed 백필 / 기존 id 보존 / output_index 안정 세 칸을 새로 넣음. 인덱스 누락 충돌은 안 봄.

점수는 56임. grok CLI 스트릭트 디코더 불이라 실사용은 있음. 지금 `dev` 불은 native-main 503(#2130)임. 드래프트 + 체크리스트 0/4. 2.27 막판에 안 넣음. #2129가 이미 이 모듈의 무조건 rewrite 때문에 윈도우 eager 마커가 상수 true인 걸 테스트로 고정했음. 여기 필드 하나 더 넣어도 그 결합은 그대로임.

해결방안: 박스 채우고 드래프트 풀면 됨. `output_index` 없는 이벤트는 id를 안 만들거나 이벤트 타입+순번 해시를 쓰는 쪽이 중복이 없음. grok CLI로 `response.completed` 한 번 보면 `Closes` 달 수 있음. 지금은 Refs가 맞음.

이 댓글은 grok-bot이 작성했습니다

bet4it:
Thanks for the review. The `output_index` fallback-to-0 collision can't happen in practice, for two reasons:

**1. The fallback only applies to streaming `output_item.added` / `output_item.done`, where `output_index` is a required field per the Responses API spec.** A spec-compliant upstream always sends it as a non-negative integer. The `Number.isInteger && >= 0` guard (added in the latest push) is purely defensive — it only fires on a spec-violating payload, and even then the worst case is a duplicate synthetic id, which is still strictly better than the current `missing field id` crash.

**2. The `response.completed` path — where the `missing field id` error actually occurs — doesn't use the fallback at all.** It goes through `backfillResponseOutput`, which uses `output.map((item, idx) => backfillOutputItem(item, idx))`. The array index is always unique and sequential, derived from the item's position in `response.output[]`. There is no fallback to 0; every item gets a distinct index by construction.

So the collision scenario requires a non-compliant upstream that simultaneously (a) omits `id` on output items AND (b) omits `output_index` on streaming events for two different items — while still being functional enough to produce a `response.completed` with a well-formed `output[]` array. In that case the streaming events might share `msg_ocx_0`, but the terminal `response.completed` (which is what the strict decoder actually deserializes into the final object) would still have unique ids from the array index.

A hash-based approach would be non-deterministic across streaming and completed events for the same item, making client-side correlation harder. The current `prefix + ocx_ + index` scheme stays stable across `output_item.added` → `output_item.done` → `response.completed` as long as the upstream sends `output_index` (which it must).

lidge-jun:
Thanks for this, @bet4it — closing as superseded by #2142, which carries **your implementation and your tests**.

One correction was added on top: an absent or malformed `output_index` collapsed to `0`, so two such items both synthesized `msg_ocx_0`. Duplicate ids are the defect the backfill exists to prevent, so an unusable index now falls back to a monotonic ordinal based far above any plausible real index. Your well-formed path is untouched — the stable index-derived id is what makes streaming events agree on the same item, and that behavior is now pinned by its own test.

That meant replacing your `expect(parsed.item.id).toBe("msg_ocx_0")` assertion rather than deleting it quietly: it now asserts the id is well-formed, and a new test asserts two such items do not collide. I verified that new test fails against your branch's source and passes with the fallback, so it is a real guard.

Your locale doc changes were left out of this PR only because they were uneven across languages; that is worth a separate parity pass, not a reason to hold the fix. Your work is credited in the PR description.

PR #2104, opened 2026-08-19
fix(xai): stream OAuth Grok through Responses

olddonkey:
## Summary

- Route native Codex Responses traffic for OAuth-backed `xai/grok-4.5` and `xai/grok-4.6` through the Responses backend declared by the official Grok CLI model catalog.
- Scope the registry wire default by authentication mode, so API-key xAI requests, translated Chat/Anthropic callers, other Grok models, generic Responses providers, and explicit adapter overrides retain their existing behavior.
- Extend the existing one-shot xAI OAuth 401 refresh/replay contract to native Responses passthrough requests without changing credential storage or logging.
- Fail closed for caller-owned `service_tier` on the xAI OAuth gateway. The registry default explicitly denies forwarding it, and that denial applies only when the registry default wins; API-key Fast and explicit overrides remain isolated.
- Add regression coverage for routing boundaries, Fast/service-tier policy, OAuth replay, and relaying the first native Responses delta before upstream completion; document the mixed-wire contract in `structure/04_transports-and-sidecars.md`.

The previous xAI OAuth preset intentionally followed the older Grok CLI Chat Completions contract. The current official model catalog reports `api_backend: "responses"` for Grok 4.5 and 4.6, while OpenCodex still selected its provider-wide `openai-chat` adapter. That compatibility drift could leave Codex waiting until the reasoning turn was effectively complete before seeing output.

Security boundary: this reuses the existing generation-checked, singleflight OAuth refresh and permits only one pre-stream replay. It does not add credential persistence, forwarding destinations, request-body logging, or new OAuth scopes. Because it touches the OAuth replay path, it still requires explicit maintainer security review before merge.

## Verification

At commit `5a1fc91a7`, rebased directly onto `upstream/dev` commit `cd8f9b8ab`:

- `bun run typecheck`
- `bun run test` — **13,540 pass / 10 skip / 0 fail** across 856 files
- `bun run privacy:scan`
- `git diff --check upstream/dev...HEAD`
- Focused xAI/Fast/service-tier regression set — **288 pass / 0 fail**
- Real outbound regression: with `fastMode: true` and inbound `service_tier: "priority"`, the OAuth xAI `/responses` request contains no `service_tier`; generic Responses passthrough and explicit overrides retain their existing contracts.
[...]

PR #2040, opened 2026-08-18
fix(responses): bridge routed tool_search passthrough

Ingwannu:
## Summary

- lower a declared private `tool_search` tool to a collision-free public function-tool identity only for non-canonical Responses passthrough providers
- translate prior `tool_search_call` / `tool_search_output` history to public function pairs and restore matching JSON/SSE calls to the private client lifecycle
- preserve type-aware `tool_choice` authorization, keep an ordinary same-named function distinct, and cap temporary SSE buffering at 256 frames / 1 MiB with raw pass-through fallback
- leave canonical OpenAI forward traffic unchanged
- document the compatibility boundary in English, Korean, Japanese, and Simplified Chinese

Fixes #1950.

## Verification

- `taskset -c 0,1 bun test tests/responses-tool-search-repair.test.ts tests/openai-responses-passthrough.test.ts tests/responses-parser.test.ts tests/responses-tool-conformance.test.ts tests/responses-undeclared-tool-guard.test.ts` — 161 passed
- `taskset -c 0,1 bun run typecheck`
- `taskset -c 0,1 bun run privacy:scan`
- `cd docs-site && taskset -c 0,1 bun install --frozen-lockfile && taskset -c 0,1 bun run build` — 385 pages built

The pre-rebase full suite completed with 13,187 passes and no #1950-related failures. Its remaining failures were already reproducible on that `dev` baseline; several were subsequently corrected by the `dev` stabilization commit included in this PR's base. Exact-head GitHub CI remains required before merge.

## Integration note

This change is scoped to the Bun/TypeScript Responses passthrough. The current remote exposes `dev`, `main`, and `preview` but not `dev2-go`, so there is no reachable Go integration branch to port in this PR. Re-evaluate the corresponding Go boundary if that branch is published before merge.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults.

## Summary by CodeRabbit

- **New Features**
  - Added `tool_search` compatibility for noncanonical OpenAI Responses gateways.
  - Client-executed searches are translated for compatible gateways and restored in JSON and streaming responses.
  - Supports nested tool catalogs, collision-safe naming, prior-turn history, and preservation of ordinary function calls.
  - Canonical OpenAI forwarding continues using the native tool-search format.
  - Clarified authentication and header-forwarding behavior for supported gateway types.

- **Documentation**
[...]

PR #2137, opened 2026-08-19
fix(responses): stop requiring a ChatGPT credential for routed providers

### LINKED ISSUE #2132: [Bug] non-OpenAI providers unusable without a ChatGPT login after v2.23.0
### Client or integration

Codex CLI

### Area

### Summary

Since v2.23.0 (#1686 / PR #1861), any request admitted via `Authorization: Bearer`
fails with a 401 when no ChatGPT credential is stored — **even when the request
routes to a non-OpenAI provider that has its own API key and never touches the
Codex backend.**

```
No usable Codex main credential to serve this request
```

I use opencodex solely to reach Cloudflare AI Gateway models from Codex. I have
deliberately never configured a ChatGPT/OpenAI login, so `~/.codex/auth.json` has
no token. This worked through v2.7.x and broke on upgrade.

Expected: a request routed to a key-authenticated provider should not require a
stored Codex main credential, because that credential is irrelevant to the
upstream it is being sent to.

### Reproduction

1. Run opencodex bound to a non-loopback hostname (mine is in Docker, so `0.0.0.0`)
   with a data-plane credential configured — either `OPENCODEX_API_AUTH_TOKEN` or an
   entry in `config.apiKeys`. `isApiAuthRequired` is then true.
2. Configure a non-OpenAI provider (e.g: Cloudflare AI Gateway).
3. Ensure `~/.codex/auth.json` has **no** token — no ChatGPT login, no OpenAI API key.
4. Point Codex at the proxy so the admission secret travels in `Authorization`:

   ```toml
   [model_providers.opencodex]
   name = "opencodex"
   base_url = "http://localhost:10100/v1"
   wire_api = "responses"

   [model_providers.opencodex.auth]
   command = "sh"
   args = ["-c", "printf %s \"$OPENCODEX_KEY\""]
   ```

   `[.auth]` is used rather than `env_key` deliberately: `should_refresh_models()`
   in codex-rs is `uses_codex_backend() || has_command_auth()`
   (`models-manager/src/manager.rs:438`), so declaring it is what makes Codex query
   `/models?client_version=…` and get the rich model shape. Note the command's stdout
   becomes `CodexAuth::from_api_key(...)` (`login/src/auth/external_bearer.rs:51`)
   and is sent as `Authorization: Bearer`, exactly like `env_key`.

5. Send any turn → 401.

### Admission itself is fine — this is not a misconfigured key

Worth ruling out up front, because the error mentions a credential:

- The proxy **admits** the request. An unrecognised secret would return
  `401 "opencodex API key required"` from `resolveResponsesApiAuth`; the
  main-credential error is only reachable *after* admission has already succeeded
  (`core.ts:1114` runs only once `resolveDataPlaneAdmissionSecret` has matched).
- The bind really is non-loopback. `assertServerAuthConfig`
  (`src/server/auth-cors.ts:287-295`) refuses to start a non-loopback bind that has
  no data-plane credential, and my instance starts and serves normally.

So both preconditions hold and the credential in play is correct. What is missing is
a *Codex main* credential, which I have deliberately never configured and which this
request has no reason to need.

### Where it goes wrong

- `src/server/responses/core.ts:1088`
  ```js
  const substituteMainCredential = options.admission?.source === "bearer";
  ```
  Keyed on *how the caller authenticated*, never on *where the request routes*.

- `src/server/responses/core.ts:1093-1102` — `route.codexAccountMode` is only ever
  set for the provider named `openai` (`src/providers/registry.ts:2675`, enforced
  again at `src/server/management/provider-routes.ts:582`). Every other provider
  falls through to the placeholder `authCtx = { kind: "main", accountId: null }`.

- `src/codex/auth-context.ts:490` — `materializeCodexUpstreamAuth` sees
  `kind === "main"` plus the substitute flag, requires a stored main token, finds
  none, and throws `CodexMainSubstitutionUnavailableError` → the 401 at
  `core.ts:1152`.

The safety rationale in the #1686 comments is sound but scoped too widely. The
concern is forwarding an admission secret upstream, which only applies to routes
that actually forward `Authorization` — the `openai`/forward-auth path. A key-auth
provider builds its own headers from scratch (`src/adapters/openai-chat.ts:81-83`),
so there is nothing to leak and nothing to protect; the ChatGPT credential
requirement is collateral damage.

### Regression point

v2.7.13 had one admission path for the whole data plane, and simply served the
request once the secret was recognised:

```js
export function hasValidApiAuth(req, config) {
  const actual = req.headers.get("x-opencodex-api-key")?.trim()
    || req.headers.get("authorization")?.replace(/^Bearer\s+/i, "").trim()
    || req.headers.get("x-api-key")?.trim();
  return actual ? isProxyAdmissionSecret(actual, config) : false;
}
```

The Responses transport was later split into `resolveResponsesApiAuth`
(dedicated header only); #1686 re-admitted bearers but coupled that admission to
*mandatory* substitution.

### Also affected

`src/server/responses/compact.ts:325` carries the identical unconditional flag, so
compaction fails the same way.

`#2076` (v2.26.0) widens the blast radius for everyone else: it switched the
auto-injected provider block from `env_http_headers = { "x-opencodex-api-key" = … }`
to `env_key = "OPENCODEX_API_AUTH_TOKEN"`, which moves existing non-loopback
auto-injected installs from the dedicated header onto the failing bearer path.
(This did not affect me — my opencodex runs in a container with no `codex` binary,
so `inject.ts` never runs and my config is hand-written.)

### Suggested fix

Gate the flag on the route actually targeting the Codex backend, in both files:

```js
const substituteMainCredential = !!route.codexAccountMode
  && options.admission?.source === "bearer";
```

For a non-Codex route the placeholder `{ kind: "main" }` context has no credential
to substitute and needs none, so `materializeCodexUpstreamAuth` returns the
forwarded headers and the provider adapter supplies its own auth as before.

### Workaround (for anyone else hitting this)

Keep the `auth` block and add the dedicated header, which wins in
`resolveResponsesApiAuth` so substitution never runs:

```toml
[model_providers.opencodex]
name = "opencodex"
base_url = "http://localhost:10100/v1"
wire_api = "responses"
env_http_headers = { "x-opencodex-api-key" = "OPENCODEX_KEY" }

[model_providers.opencodex.auth]
command = "sh"
args = ["-c", "printf %s \"$OPENCODEX_KEY\""]
```

`env_http_headers` must sit **above** the `[.auth]` sub-table header or TOML parses
it as `auth.env_http_headers`. It is not in codex-rs's conflict list for `auth`
(only `env_key`, `experimental_bearer_token`, `requires_openai_auth` conflict), so
the combination is accepted and model refresh is preserved.

### Version

2.26.0

### Operating system

Debian Testing for host, node:22-alpine for the base docker image

### Provider and model

cloudflare-workers-ai / @cf/zai-org/glm-5.2

### Logs or error output

```shell
unexpected status 401 Unauthorized: No usable Codex main credential to serve this request, url: http://localhost:10100/v1/responses
```

_No response_

### Redacted configuration

Codex config:

```markdown
[model_providers.opencodex]
name = "opencodex"
base_url = "http://localhost:10100/v1"
wire_api = "responses"

[model_providers.opencodex.auth]
command = "sh"
args = ["-c", "printf %s \"$OPENCODEX_KEY\""]
```

Opencodex config:

```json
{
  "port": 10100,
  "managementUsageMaxReadBytes": 67108864,
  "appOwnedMemoryBudgetMb": 256,
  "hostname": "0.0.0.0",
  "providers": {
    "cloudflare-workers-ai": {
      "adapter": "openai-chat",
      "baseUrl": "https://api.cloudflare.com/client/v4/accounts/XXXX/ai/v1",
      "authMode": "key",
      "apiKey": "XXXX",
      "models": [
        "@cf/zai-org/glm-5.2",
        "@cf/moonshotai/kimi-k2.7-code"
      ]
    },
    "openai": {
      "authMode": "forward",
      "codexAccountMode": "pool"
    }
  },
  "defaultProvider": "cloudflare-workers-ai",
  "subagentModels": [
    "gpt-5.5",
    "gpt-5.6-sol",
    "gpt-5.6-terra",
    "gpt-5.6-luna",
    "gpt-5.4-mini"
  ],
  "websockets": false,
  "codexAutoStart": false,
  "openaiProviderTierVersion": 2
}
```

### Checks

lidge-jun:
## Summary

A request routed to a key-authenticated provider was being gated on a ChatGPT credential it has no use for, so an install that deliberately never logged into ChatGPT got `401 No usable Codex main credential` on every request. Reported in #2132 as a regression from v2.23.0.

#1686 made a caller that proves admission with one of our own secrets substitute the stored main credential, so that secret never leaves the process. That is correct for a route that actually reaches the ChatGPT backend. The flag was computed from **how the caller authenticated** and never from **where the request routes**:

```ts
const substituteMainCredential = options.admission?.source === "bearer";
```

`route.codexAccountMode` is set only for the native `openai` row, so every other provider fell through to `authCtx = { kind: "main" }`, hit the substitution branch, and threw `CodexMainSubstitutionUnavailableError` when no ChatGPT token was stored — for a request that was never going to touch ChatGPT.

The fix adds the missing half of the question: substitute only when the route can actually consume that credential. Both `pool` and `direct` continue to substitute, so #1686's Direct contract is preserved rather than narrowed — an important detail, since Direct bearer admission is only safe *because* substitution still runs.

Same one-line change at both call sites (`core.ts`, `compact.ts`).

## Verification

- New `tests/bearer-admission-routed-provider.test.ts`, driven **RED** against the unpatched tree first: the key-auth case reproduced the exact reported `401`, and the two #1686 guarantees passed before and after. It asserts the provider's own key goes upstream, our admission secret never does, a native route with no stored credential still fails closed **before any upstream I/O**, and a native route with a stored credential still substitutes it.
- `bun test --isolate tests` — 13,516 pass, 0 fail, 10 skip (857 files).
- Focused admission/auth suites (`codex-envkey-admission-substitution`, `codex-auth-context`, `forward-admission-separation`, `data-plane-admission-identity`, `codex-admission`) — 88 pass, 0 fail.

## Merge-order note

Three other open PRs also touch `src/server/responses/core.ts`: #2104, #2040, and #2101 (which also touches `compact.ts` and `auth-context.ts`). This change is 5 lines across two files and restructures nothing, so any conflict should be trivial, but whichever lands second will want a rebase.

## Checklist

- [x] Docs or release notes were updated when needed. (No documented behavior changes: the substitution and its fail-closed path are unchanged for every route that could use them.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (Auth-boundary change: the admission secret still never reaches an upstream, native routes still fail closed before I/O, and both properties are pinned by tests. Privacy scan green.)

Closes #2132

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved credential handling for routed providers so their configured credentials are preserved.
  - Native Codex routes continue to use stored credentials when available.
  - Requests to native routes now fail safely when required credentials are missing.

- **Tests**
  - Added coverage for routed-provider authentication, native-route behavior, and credential substitution scenarios.

---

**Stack** (merge bottom-up):

| 2 | #2146 | account entitlement gating | entitlement discovery, catalog projection, Pool/Direct selection |
| 1 | this PR | bearer admission predicate ← you are here | the `substituteMainCredential` contract |

#2146 consumes the `substituteMainCredential` value this PR corrects, so this one merges first.

### DIFF
--- a/src/server/responses/compact.ts
+++ b/src/server/responses/compact.ts
@@ -322,7 +322,10 @@ export async function handleResponsesCompact(
 
   // #1686: a bearer-presented admission secret is one of ours, so the stored main credential
   // is substituted below instead of the caller bearer being forwarded.
-  const substituteMainCredential = admission?.source === "bearer";
+  // #2132: and only when the route is a native Codex one, which is the only route that can
+  // consume that credential. See the longer note in core.ts resolveResponsesCodexAuth.
+  const substituteMainCredential = admission?.source === "bearer"
+    && route.codexAccountMode !== undefined;
   if (route.codexAccountMode === "direct" && !substituteMainCredential) {
     try { validateForwardAdmissionCredential(req.headers, config); }
     catch (err) {
--- a/src/server/responses/core.ts
+++ b/src/server/responses/core.ts
@@ -1085,7 +1085,15 @@ async function resolveResponsesCodexAuth(
     // Refusing it here is what made the codex-cli `env_key` contract unusable against Direct.
     // Admitting it is only safe because the stored main credential is substituted below, so
     // the admission secret still never leaves this process.
-    const substituteMainCredential = options.admission?.source === "bearer";
+    //
+    // #2132: substitution answers "does THIS ROUTE need our stored ChatGPT credential", not
+    // "how did the caller authenticate". Only a native Codex route reaches the ChatGPT backend
+    // and can consume that credential; a key-authenticated routed provider carries its own and
+    // never touches it. Keying on the caller alone made an install that deliberately never
+    // logged into ChatGPT fail every routed request with "No usable Codex main credential".
+    // `codexAccountMode` is set only for the native openai row, which is exactly that test.
+    const substituteMainCredential = options.admission?.source === "bearer"
+      && route.codexAccountMode !== undefined;
     if (route.codexAccountMode === "direct" && !substituteMainCredential) {
       validateForwardAdmissionCredential(req.headers, config);
     }
--- a/tests/bearer-admission-routed-provider.test.ts
+++ b/tests/bearer-admission-routed-provider.test.ts
@@ -1,0 +1,188 @@
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { saveConfig } from "../src/config";
+import { startServer } from "../src/server";
+import type { OcxConfig } from "../src/types";
+
+/**
+ * Issue #2132: bearer admission must not require a stored ChatGPT credential.
+ *
+ * #1686 made a caller that proves admission with one of OUR secrets substitute the stored
+ * main credential, so the admission secret never leaves the process. That is right for a
+ * route that actually reaches the ChatGPT backend. It was applied by asking HOW the caller
+ * authenticated and never WHERE the request routes, so a request bound for a
+ * key-authenticated provider — which carries its own credential and never touches ChatGPT —
+ * was gated on a credential it has no use for. An install that deliberately never logged
+ * into ChatGPT got 401 "No usable Codex main credential" on every request.
+ *
+ * The substitution itself is unchanged and still fails closed for native routes; only the
+ * question it is asked changes.
+ */
+
+const originalFetch = globalThis.fetch;
+const previousOcxHome = process.env.OPENCODEX_HOME;
+const previousCodexHome = process.env.CODEX_HOME;
+const previousDataToken = process.env.OPENCODEX_API_AUTH_TOKEN;
+
+let ocxHome = "";
+let codexHome = "";
+let routedAuth: Array<string | null> = [];
+let nativeAuth: Array<string | null> = [];
+
+const ADMISSION_SECRET = "ocx_data_2132secret";
+const ROUTED_KEY = "sk-routed-provider-key";
+
+/** A JWT whose `exp` is far in the future, so a stored main token reads as live. */
+function liveJwt(): string {
+  const payload = Buffer.from(JSON.stringify({ exp: Math.floor(Date.now() / 1000) + 86_400 })).toString("base64url");
+  return `header.${payload}.signature`;
+}
+
+/**
+ * A remote bind (so admission is required rather than loopback-waived) with BOTH a native
+ * openai row and a key-authenticated routed provider. The routed provider is the one under
+ * test; the native row has to exist for the negative case to be reachable.
+ */
+function mixedConfig(): OcxConfig {
+  return {
+    port: 0,
+    hostname: "0.0.0.0",
+    defaultProvider: "openai",
+    openaiProviderTierVersion: 2,
+    providers: {
+      openai: {
+        adapter: "openai-responses",
+        baseUrl: "https://chatgpt.com/backend-api/codex",
+        authMode: "forward",
+        codexAccountMode: "direct",
+        defaultModel: "gpt-5.6-luna",
+      },
+      gateway: {
+        adapter: "openai-chat",
+        baseUrl: "https://gateway.example.com/v1",
+        authMode: "key",
+        apiKey: ROUTED_KEY,
+        models: ["gateway-model"],
+      },
+    },
+    apiKeys: [
+      { id: "env-key", name: "env_key", key: ADMISSION_SECRET, createdAt: "2026-08-20T00:00:00.000Z" },
+    ],
+  } as OcxConfig;
+}
+
+beforeEach(() => {
+  ocxHome = mkdtempSync(join(tmpdir(), "ocx-2132-home-"));
+  codexHome = mkdtempSync(join(tmpdir(), "ocx-2132-codex-"));
+  process.env.OPENCODEX_HOME = ocxHome;
+  process.env.CODEX_HOME = codexHome;
+  delete process.env.OPENCODEX_API_AUTH_TOKEN;
+  routedAuth = [];
+  nativeAuth = [];
+  globalThis.fetch = (async (input, init) => {
+    const raw = input instanceof Request ? input.url : String(input);
+    const url = new URL(raw);
+    const headers = new Headers(input instanceof Request ? input.headers : init?.headers);
+    if (url.hostname === "gateway.example.com") {
+      routedAuth.push(headers.get("authorization"));
+      return Response.json({
+        id: "chatcmpl_2132",
+        object: "chat.completion",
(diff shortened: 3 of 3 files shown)


lidge-jun:
## 리뷰 · 우선순위 72 / 80

#2132 그거임. 지금 `dev`(`cd8f9b8ab`)도 `src/server/responses/core.ts:1088`이 `options.admission?.source === "bearer"`만 봄. compact는 `:325` 쌍둥이. 키 프로바이더는 `codexAccountMode`가 없어서 `{ kind: "main" }`로 떨어지고, `materializeCodexUpstreamAuth`가 메인 토큰 없으면 401. openai-chat은 `Bearer ${provider.apiKey}`로 헤더를 새로 만듦. 입학 시크릿이 나갈 구멍이 없음.

고친 한 줄이 이슈 제안이랑 같음. `!!route.codexAccountMode && source === "bearer"`. pool/direct는 계속 치환해서 #1686 Direct 계약을 안 좁힘. 테스트가 빨강으로 401을 재현하고, 키는 업스트림으로 가고 입학 시크릿은 안 가고, 네이티브는 토큰 없으면 IO 전에 죽음.

`core.ts`를 #2104 #2040 #2101도 건드림. 이 PR은 5줄이라 충돌 나면 리베이스하면 됨. 닫으라는 쪽 아님. types.ts/config.ts 안 건드림. 분할에 깔림 없음.

점수는 72임. 2.26.0에서 ChatGPT 없이 키 프로바이더만 쓰는 설치가 죽음. #2130이랑 다른 줄. 2.27에 넣을 만함. 태그 찍었으면 바로 다음.

해결방안: 이 PR 머지. #2132 Closes 맞음. 워크어라운드 전용 헤더는 남겨도 됨. 후속으로 `env_key` 자동주입(#2076)이 bearer로 미는 건 문서만 적어도 됨.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
The direction and the route-scoped predicate are correct, but I am requesting changes before approval for two concrete reasons:

1. The current head `87524414b83d` is 22 commits behind the current `dev` head `caf20353f`. The green jobs validate the old base at `890d339dec9f`, not the current integration state. Please rebase onto the latest `dev` and rerun CI on the new exact head.
2. This changes both `src/server/responses/core.ts` and `src/server/responses/compact.ts`, while `tests/bearer-admission-routed-provider.test.ts` exercises only `/v1/responses`. Please add compact regressions for the same three trust-boundary cases: a key-authenticated routed provider works without a stored ChatGPT credential, a native route without a stored credential fails before upstream I/O, and a native route with a stored credential substitutes it. In every case, assert that the admission secret is never forwarded upstream.

Once those two boundaries are covered and exact-head CI is green, this remains a strong merge candidate for #2132.

PR #2138, opened 2026-08-19
fix(responses): drop the retired prompt_cache_retention for gpt-5.6

lidge-jun:
## Summary

The ChatGPT backend rejects a `gpt-5.6` request that still carries `prompt_cache_retention` with `400 Unsupported parameter` (#2092, reproduced 3/3 by the reporter: field present → 400, field omitted → 200). GPT-5.6 replaced it with `prompt_cache_options.ttl`.

This strips the retired field on the canonical ChatGPT forward path, for the `gpt-5.6` family only.

**Credit:** the implementation is @lilinxiong's from #2102, which had the right contract of the three proposals. @yzxcj797's #2099 contributed the `Fixes #2092` link and the repro-shaped fixture. @luvs01's #2091 raised the same defect.

### Why the narrowness is the fix

Two deliberate limits, each pinned by a test that a wider strip would fail:

- **Model:** exact `gpt-5.6` or a dashed `gpt-5.6-*` prefix. A bare `startsWith("gpt-5.6")` also matches a future `gpt-5.60`, silently changing an unrelated model.
- **Backend:** canonical ChatGPT forward only. A self-hosted or third-party forward gateway may still honor the field; the issue reporter explicitly withdrew the broader "never send it anywhere" claim, and `tests/openai-responses-passthrough.test.ts:807` already pins that a gpt-5.5 forward request keeps it.

The retired value is **not** translated into `prompt_cache_options`. 5.6 carries a different TTL contract and implicit caching still applies when the caller sent no replacement options, so synthesizing one would change a caching decision the caller never made.

## Verification

- Extended `tests/openai-responses-passthrough.test.ts`, driven **RED** first: reverting only the adapter change fails 5 of the new tests. The two narrowness guards (`gpt-5.60`, noncanonical gateway) stay green in both directions, which is what proves they are guards rather than restatements.
- `bun test --isolate tests` — 13,537 pass, 0 fail, 10 skip (855 files).
- `bun test --isolate tests/openai-responses-passthrough.test.ts` — 75 pass, 0 fail.
- `bun run typecheck` — clean.
- `bun run privacy:scan` — passed.

## Supersedes

Once this lands, #2102, #2099, and #2091 are all resolved by it. I will close them with attribution rather than leaving three open PRs for one defect.

Note: #2092's intermittent 400s on bodies that already lacked the field are **not** addressed here and remain open — that is a separate backend/retry question, and widening the strip would not fix it.

## Checklist

[...]

PR #2140, opened 2026-08-19
fix(routing): resolve model-keyed lookups the way the runtime resolves them

lidge-jun:
## Summary

Two evidence surfaces read per-model config with a bare map lookup while the runtime resolves the same keys through `modelRecordValue`. Family and case overrides were invisible to them, and a prototype-shaped id resolved an `Object.prototype` member instead of missing.

**Credit: both patches are @ntdatt812's work**, from #2100 and #2077, applied unchanged. They are combined here because they are one thesis — *model-keyed lookups must use the runtime's own resolution rules* — on two disjoint files.

- `src/routing/capability.ts` (#2100): `gpt-oss:120b` inherited the provider-wide 8k window instead of the `gpt-oss` family's 131072, and `noVisionModels` was ignored. These values select routing candidates, not just log lines.
- `src/routing/compatibility/behavior.ts` (#2077): the same lookup bug in the Lab behavior report, where `constructor` resolved to `Object.prototype.constructor`, made `jcsStringify` throw and silently dropped Lab subjects.

Exact-own maps (`modelPreferHostedTools`, `modelOpenRouterRouting`) deliberately do **not** family-spread, and that boundary is pinned by its own tests.

## Verification

- Both patches applied cleanly to current `dev` and are proven **RED**: reverting only the two source files fails **13** of the new tests.
- `bun test --isolate tests` — 13,556 pass, 0 fail, 10 skip (856 files).
- `tests/core-lab-boundary.test.ts` green: `behavior.ts` already imported Lab types, and this does not newly puncture the core/Lab boundary.

## Supersedes

Closes #2100 and #2077 once merged. Both are @ntdatt812's; I will close them with attribution rather than leaving duplicates open.

## Checklist

- [x] Docs or release notes were updated when needed. (Fixes lookups to match documented runtime behavior; nothing user-facing changes shape.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (No auth/credential surface. The prototype-id hardening removes an `Object.prototype` leak, which is a small safety improvement.)

Closes #2100
Closes #2077

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model capability detection for tagged and case-variant model IDs.
  * Corrected context-window and capability fallback behavior for model families and provider defaults.
  * Improved vision detection so explicitly text-only models are classified correctly.
  * Prevented ambiguous model names from matching unrelated compatibility settings.
  * Preserved exact model-specific settings over broader family-level defaults.
* **Tests**
  * Added coverage for model-family matching, fallback behavior, vision handling, and routing compatibility.

### DIFF
--- a/src/routing/capability.ts
+++ b/src/routing/capability.ts
@@ -10,7 +10,7 @@
  * how that affects eligibility.
  */
 
-import type { OcxConfig } from "../types";
+import { modelInList, type OcxConfig } from "../types";
 import { isCanonicalOpenAiForwardProvider, OPENAI_CODEX_PROVIDER_ID } from "../providers/openai-tiers";
 import { serviceTierSupportForModel } from "../providers/service-tier";
 import { PROVIDER_REGISTRY } from "../providers/registry";
@@ -21,6 +21,7 @@ import {
   nativeReasoningEfforts,
 } from "../codex/catalog/metadata";
 import { readCatalog, readCodexCatalogPath } from "../codex/catalog/parsing";
+import { modelRecordValue } from "../reasoning-effort";
 import { statSync } from "node:fs";
 import type { RouteCapabilityEvidence } from "./trace";
 
@@ -159,9 +160,14 @@ export function candidateCapabilityEvidence(
   const catalogRow = cachedCatalogModels().find(model => model.provider === providerName && model.id === modelId);
   const isNative = providerName === OPENAI_CODEX_PROVIDER_ID && !modelId.includes("/");
 
-  const rawContextWindow = provider?.modelContextWindows?.[modelId]
+  // `modelRecordValue`, not a bare lookup: every runtime reader of these three maps
+  // resolves them that way, so a `gpt-oss` entry covers `gpt-oss:120b`. Reading raw
+  // made the evidence disagree with the resolver it claims to describe — and for the
+  // window it did not even degrade to unknown, it fell through to the provider-wide
+  // value, which is a definite wrong answer rather than an absent one.
+  const rawContextWindow = modelRecordValue(provider?.modelContextWindows, modelId)
     ?? provider?.contextWindow
-    ?? registryEntry?.modelContextWindows?.[modelId]
+    ?? modelRecordValue(registryEntry?.modelContextWindows, modelId)
     ?? catalogRow?.contextWindow
     ?? (isNative ? nativeOpenAiContextWindow(modelId, nativeContextLimits(config)) : undefined);
   // Native rows go through the accessor (raise-to-ceiling + opt-in). Routed rows keep
@@ -170,10 +176,21 @@ export function candidateCapabilityEvidence(
     ? (nativeOpenAiContextWindow(modelId, nativeContextLimits(config)) ?? rawContextWindow)
     : rawContextWindow;
 
-  const modalities = provider?.modelInputModalities?.[modelId]
-    ?? registryEntry?.modelInputModalities?.[modelId]
-    ?? catalogRow?.inputModalities
-    ?? (isNative ? nativeInputModalities(modelId) : undefined);
+  // `noVisionModels` is checked before the modality chain because that is the order
+  // `isModelTextOnly` uses: it matches the no-vision list and returns true before it
+  // ever reads `modelInputModalities` (`src/vision/index.ts:32`). So a `gpt-oss`
+  // no-vision entry beats an exact `gpt-oss:120b` entry that lists "image", and
+  // deriving `image` from the modality chain alone reported vision on a model the
+  // runtime refuses it for. That matters more here than on the CLI surface fixed in
+  // #2086: routing *acts* on this evidence, so it would select the candidate for image
+  // work that execution then rejects.
+  const noVision = modelInList(provider?.noVisionModels, modelId);
+  const modalities = noVision
+    ? ["text"]
+    : (modelRecordValue(provider?.modelInputModalities, modelId)
+      ?? modelRecordValue(registryEntry?.modelInputModalities, modelId)
+      ?? catalogRow?.inputModalities
+      ?? (isNative ? nativeInputModalities(modelId) : undefined));
   const image = Array.isArray(modalities)
     ? modalities.includes("image")
     : undefined;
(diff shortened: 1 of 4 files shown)


lidge-jun:
## 리뷰 · 우선순위 54 / 80

모델 키 lookup이 `modelRecordValue`를 안 타고 맵을 그냥 보던 거임. 패밀리/대소문자 오버라이드가 안 보이고, `constructor` 같은 프로토타입 id가 `Object.prototype.constructor`로 붙어서 Lab `jcsStringify`가 터짐. `capability.ts`는 라우팅 후보를 고름. `gpt-oss:120b`가 프로바이더 8k를 먹고 `gpt-oss` 131072를 무시함. `noVisionModels`도 무시. 로그가 아니라 후보 선택임.

#2100이랑 #2077을 한 테제로 합친 거임. 둘 다 ntdatt812 작업. exact-own 맵(`modelPreferHostedTools`, `modelOpenRouterRouting`)은 패밀리 스프레드 안 함. 그 경계를 테스트로 고정함. 소스 두 파일만 되돌리면 새 테스트 13개가 빨갬. Lab 바운더리는 원래 behavior.ts가 Lab 타입을 임포트함. 새로 뚫은 거 아님.

점수는 54임. 맞는데 지금 `dev` 불은 아님. #2137 bearer 401, #2138 gpt-5.6 400이 먼저임. types.ts/config.ts 안 건드림. 닫고 리베이스하라는 쪽 아님. #2100 #2077은 이 PR 머지되면 닫으면 됨. 중복으로 열어둘 이유 없음.

해결방안: 머지 후 원 PR 두 개 attribution 달고 닫기. 2.27 필수 아님. #2137 #2138 뒤에 넣으면 됨.

이 댓글은 grok-bot이 작성했습니다

ntdatt812:
Thanks for consolidating these, and for the write-up — the single-thesis framing is better than my two PRs were. "Model-keyed lookups must use the runtime's own resolution rules" is the right way to state it, and folding both files under one `resolverVersion` decision is cleaner than reviewing them apart. No objection to closing #2100 and #2077.

One mechanical thing, and only because the PR is still open so it costs you nothing now:

`f9c224b7` is authored by `bitkyc08-arch <bitkyc08@gmail.com>` with the credit in prose, and there is no `Co-authored-by:` trailer. Prose credit reads fine to a human, but `git log`, `git blame`, the contributor graph and `/repos/.../contributors` all key off the author field and that trailer — so as it stands the two patches will show as entirely yours in the history.

Your earlier merges did carry it through, which is how I noticed the difference: #2085 and #2086 landed as merge commits, so my commits survived intact and the API shows `contributions=8`. A squash drops that unless the trailer is present.

If you're willing, one line at the end of the commit body is all it takes:

```
Co-authored-by: Nguyen Thanh Dat <ntdat812@gmail.com>
```

That is the identity GitHub already links to my account, so it will resolve.

Entirely your call, and I would rather have the patches land as you have them than have them back. Just flagging it while it is still a one-line amend rather than after the merge.

Ingwannu (changes requested):
I cannot approve the current head because it is 18 commits behind the current `dev` tip (`caf20353f`). The existing diff and CI results do not validate the change against the current integration state.

Please rebase onto the latest `dev`, resolve conflicts in the affected runtime paths, rerun the focused checks plus the full exact-head suite, and re-request review. This is a review-state blocker, not a rejection of the fix direction.

ntdatt812:
@Ingwannu's rebase blocker is fair, and since the diff here is my patch from #2100/#2077 I ran the validation against the current tip so the rebase does not have to be done blind. I cannot push to this branch, so this is data rather than a fix.

**The patch cherry-picks onto `caf20353f` with no conflicts.**

```
git checkout --detach origin/dev        # caf20353f
git cherry-pick f9c224b70               # clean, 4 files changed, 345 insertions(+), 12 deletions(-)
```

Nothing in the 18 intervening commits touches `src/routing/capability.ts` or `src/routing/compatibility/behavior.ts` in a way that collides.

**Results on that exact rebased head:**

```
bun run typecheck                                             exit 0

bun test routing-capability-model-matching
         routing-compatibility-model-matching                 29 pass / 0 fail

bun test routing-capability-catalog · route-explainability
         service-tier-capability · fastwire-characterization-routing
         catalog-vision-sidecar-modalities · routing-policy-fallback
         routing-compatibility · openrouter-provider-routing
         fastwire-observability · lab-live-review-regressions  189 pass / 0 fail
```

**On the full suite, and I would rather state this plainly than round it up:** my `bun run test` run on this rebased head was interrupted before it printed a total, and at the point it stopped it had 7 failures — 5 in `tests/codex-auth-api.test.ts`, 2 in `tests/codex-auth-modal-status.test.ts`. Every one of them took ~5000 ms, which is the timeout signature rather than an assertion failure.

Run in isolation on the same tree, those two files are **192 pass / 0 fail**. They exercise OAuth pool login and GUI status copy; this patch touches two routing files and neither is in their import graph. So they are load-dependent on my Windows box, not consequences of the change — but I have not produced a clean full-suite total on the rebased head, and that is the one thing @Ingwannu asked for that I cannot hand you.

Worth knowing when you read a full-suite result from this repo on Windows: I have run it four times today on four different heads and it reddened a *different* subset each time, always tests with multi-second timeouts. Separately, `tests/server-auth.test.ts > OpenAI option auth matrix keeps direct, pool, and API credentials independent` fails **reproducibly** on a clean `dev` tree — 2/2 with the working tree checked out wholesale to `origin/dev` — so a full-suite run there is not expected to be green even before this patch.

No action needed from me on the credit question in my earlier comment; it can wait until after the rebase.

Ingwannu:
Thanks @ntdatt812 — the clean cherry-pick and focused results are useful evidence that the patch still applies. They do not update the PR head, though: `f9c224b70` remains 18 commits behind current `dev` (`caf20353f`), so the repository CI shown here still validates the old ancestry. The commit also still lacks the requested `Co-authored-by: Nguyen Thanh Dat <ntdat812@gmail.com>` trailer.

My existing changes-requested review therefore remains in place. @lidge-jun, please rebase/amend this branch, preserve the contributor trailer, and let exact-head CI run. The isolated Windows timeout reruns support treating those local failures as load-related, but they are not a substitute for updating the actual PR head.

PR #2142, opened 2026-08-19
fix(responses): backfill missing ids on output items for strict decoders

lidge-jun:
## Summary

Some relays omit the required `id` on `message`, `reasoning`, and `function_call` output items, so strict decoders (grok-build) reject the response even after #1941. This synthesizes a stable `msg_ocx_N` / `rs_ocx_N` / `fc_ocx_N` id keyed on `output_index`, and never overwrites an id the upstream actually sent.

**Credit: @bet4it's #2131** is the implementation and test suite carried here.

### One correction on top

An absent or malformed `output_index` collapsed to `0`, so two such items both became `msg_ocx_0` — duplicate ids, which is precisely the defect this backfill exists to prevent. An unusable index now falls back to a monotonic ordinal based far above any plausible real index, so a synthesized id cannot collide with an index-derived one. The well-formed path is untouched and still produces the stable index-derived id, which is what lets streaming events referencing the same item agree.

@bet4it's original test asserted the collapse directly (`expect(parsed.item.id).toBe("msg_ocx_0")`), so I replaced it rather than deleting it quietly: it now asserts the id is well-formed, and a new test asserts two such items do not collide. That new test **fails against #2131 as written** and passes here — verified by applying only the original source and re-running.

Locale docs are limited to the English source. The translated guides in the original PR were uneven (EN/FR rewritten, JA/KO/ZH/TR only the first sentence), and locale parity is not this change's thesis.

## Verification

- **RED-first**: reverting only `responses-field-backfill.ts` fails 7 tests. Separately, applying **only @bet4it's original source** fails exactly the new duplicate-id test (15 pass / 1 fail), which is what proves that test is a real guard rather than a restatement.
- `bun test --isolate tests` — 13,543 pass, 0 fail, 10 skip (855 files).
- `bun test --isolate tests/responses-field-backfill.test.ts` — 16 pass, 0 fail.

## Supersedes

Closes #2131 (@bet4it) once merged, with attribution.

## Checklist

- [x] Docs or release notes were updated when needed. (English Grok Build guide updated; translated locales deliberately left for a parity pass.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (Synthesizes opaque item ids only; no credential, auth, or user content is read or logged. Privacy scan green.)

Closes #2131

## Summary by CodeRabbit

* **New Features**
  * Added consistent generated identifiers for Responses API output items.
  * Supports identifiers for reasoning, message, and function-call outputs while preserving existing IDs.
  * Excludes compaction items from identifier updates.
* **Bug Fixes**
  * Improved handling of missing or invalid response item indexes.
  * Ensured response text annotations are restored consistently without changing unaffected data.
* **Documentation**
  * Updated the Grok Build guide to use the Responses API.
  * Documented reasoning trace handling and visibility configuration precedence.

### DIFF
--- a/docs-site/src/content/docs/guides/grok-build.md
+++ b/docs-site/src/content/docs/guides/grok-build.md
@@ -3,10 +3,10 @@ title: Grok Build
 description: Use any opencodex-routed model from xAI's Grok Build CLI — models are auto-registered into ~/.grok/config.toml while the proxy runs.
 ---
 
-opencodex serves an OpenAI-compatible `POST /v1/chat/completions` (and `/v1/responses`) on its
-local port, and Grok Build supports custom models against OpenAI-compatible servers. Starting
-with this integration, opencodex registers its whole visible catalog into Grok Build
-automatically — no manual config editing required.
+opencodex serves an OpenAI-compatible `POST /v1/responses` on its local port, and Grok Build
+supports custom models against OpenAI-compatible servers. Starting with this integration,
+opencodex registers its whole visible catalog into Grok Build automatically — no manual config
+editing required.
 
 ## Auto-registration
 
@@ -18,7 +18,7 @@ into `~/.grok/config.toml`:
 [model.ocx-gpt-5-6-sol]
 model = "gpt-5.6-sol"
 base_url = "http://127.0.0.1:10100/v1"
-api_backend = "chat_completions"
+api_backend = "responses"
 api_key = "opencodex-loopback"
 name = "OCX gpt-5.6-sol"
 # ... one [model.ocx-*] table per visible model ...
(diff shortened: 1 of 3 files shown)


lidge-jun:
## 리뷰 · 우선순위 58 / 80

#2131을 가져오면서 구멍 하나를 막음. 지금 `dev`의 `responses-field-backfill.ts`는 annotations만 채움. id는 없음. 릴레이가 message/reasoning/function_call에서 id를 빼면 grok-build serde가 죽음. 합성 id는 `msg_ocx_N` 식이고 있는 값은 안 덮음.

#2131은 `output_index`가 없거나 숫자가 아니면 0으로 떨어짐. 그런 아이템 두 개가 `msg_ocx_0`으로 충돌함. 백필이 고치려던 바로 그 결함임. 이 PR은 못 쓰는 인덱스를 실제 인덱스보다 높은 단조 ordinal로 보냄. 정상 인덱스는 그대로라 스트림 added/done이 같은 id를 유지함. 원 테스트가 `msg_ocx_0`을 단언하던 걸 바꾸고, 충돌 테스트가 #2131 소스만 적용하면 빨갬.

로케일 문서는 영어만. 원 PR 번역이 들쭉날쭉해서 이번 테제가 아님. types.ts/config.ts 안 건드림. 분할에 깔림 없음. #2131은 머지되면 닫으면 됨. 리베이스 지옥 되니까 둘 다 열어두지 말 것.

점수는 58임. grok CLI 실사용은 있음. 지금 불은 윈도우 로그가드(#2143)랑 bearer 401(#2137)임. 2.27 필수 아님.

해결방안: 머지 후 #2131 attribution 달고 닫기. 번역 가이드는 후속.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):

PR #2143, opened 2026-08-19
fix(log-guard): accept the OS's own canonical spelling on Windows

lidge-jun:
## Summary

The Windows CI shards were reporting **22 Log Guard failures**, every one of them resolving to `unsafe_path`. This is a **release blocker, not a pre-existing break**: Log Guard (#1729) is new in `main...dev` and has never shipped, so on Windows the feature has never worked.

`realpathSync.native` on Windows expands 8.3 short components — the `RUNNER~1` form that appears throughout `%TEMP%` — so the canonical realpath and the requested path disagree as strings while naming the same file. `databasePathIsSafe` read that disagreement as an ancestor-symlink redirection and refused every mutation: protect, unprotect, repair, reclaim, compact.

macOS hit the same class of problem with `/var` and `/tmp` and was given an explicit alias normalizer at [path-safety.ts](../blob/dev/src/codex/log-guard/path-safety.ts). Windows never got the equivalent.

**The fix keeps the check identity-based, not string-based.** When the spellings differ, the requested path is re-canonicalized through the same call and the two canonical forms must agree. A genuine symlink or junction resolves somewhere else and still fails, so the guard refuses exactly what it was built to refuse. It is scoped to `win32` and cannot change any POSIX verdict.

Regression tests cover all three directions: an OS-canonicalized spelling is the same file, a symlinked database is not, and an unrelated sibling is not.

## How this was found

Honestly, later than it should have been. The Windows leg looked like a pre-existing failure because run [32147924436](https://github.com/lidge-jun/opencodex/actions/runs/32147924436) on `e446607c8` (2026-08-18) was already red — but that run is *also* after #1729 landed on `dev` (2026-08-16), so "pre-existing on dev" and "pre-existing in the release" are not the same claim. Reading the shard logs rather than trusting the earlier comparison is what surfaced it.

It also sat behind three failures of my own — POSIX mode assertions that do not apply on Windows, fixed in #2139 — which is why it took a second pass to see.

## Verification

```
bun test codex-log-guard-coderabbit, codex-log-guard-protection    18 pass / 0 fail
bun x tsc --noEmit                                                 exit 0
```

**The Windows shard is the only check that proves this one**, since the behavior it fixes exists nowhere else. macOS and Linux can only show the absence of a regression.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Log Guard path validation on Windows to recognize equivalent path spellings, including legacy 8.3 formats.
  * Continued rejecting unresolved paths, redirected locations, symlinked databases, and unrelated paths.
  * Improved startup failure reporting with relevant process output when runtime initialization fails.
  * Increased Windows CI identity-check timeouts to improve reliability.

* **Tests**
  * Added regression coverage for canonicalized paths, redirected files, and unrelated sibling paths.
  * Improved acceptance-test timing and Windows CI reliability across different environments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -607,7 +607,11 @@ jobs:
           bun run build
 
       - name: Test
-        run: bun test --isolate tests --shard=${{ matrix.shard }}/4
+        # --timeout: the Linux batches and the macOS control both pass 60000; this leg was
+        # the only one left on Bun's 5s default, and it is the slowest hardware on the board.
+        # Three of its failures were the default firing on tests that had not hung — the
+        # composed-acceptance cases spawn a real `ocx start` and were still working at 41s.
+        run: bun test --isolate --timeout 60000 tests --shard=${{ matrix.shard }}/4
 
       - name: CLI help smoke
         run: bun run src/cli/index.ts help
--- a/devlog/_plan/260819_unclaimed_bug_selection/180_windows_leg.md
+++ b/devlog/_plan/260819_unclaimed_bug_selection/180_windows_leg.md
@@ -1,0 +1,91 @@
+# 180 — the Windows leg: five defects, and one I created
+
+PR #2143. The Windows shards were red before this work and are less red after
+it; this records what was actually wrong, because "Windows is flaky" was the
+wrong answer four separate times.
+
+## The correction that matters most
+
+I told the user twice that the Windows failures predated this release range.
+The first half was true — the leg was red on `e446607c8` (2026-08-18). The
+second half was not, and I stated it anyway.
+
+**Log Guard is new in `main...dev`.** It has never worked on Windows. Every
+mutation — protect, unprotect, repair, reclaim, compact — refused with
+`unsafe_path`. Shipping this range without looking would have released a
+feature that is broken on one of three platforms, and my own "pre-existing"
+verdict is what nearly let it through.
+
+The lesson is narrow and worth keeping: **"red before my change" and "not my
+release's problem" are different claims.** A feature that landed on `dev` two
+days earlier is still in the release.
+
+Compounding it: the run I compared against had shard 4/4 **cancelled**, so the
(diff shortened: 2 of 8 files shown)


lidge-jun:
## 리뷰 · 우선순위 72 / 80

2.27 윈도우 릴리즈 블로커임. Log Guard(#1729)가 `main...dev`에 새로 들어왔고 한 번도 안 나갔음. 윈도우 샤드 22개가 전부 `unsafe_path`. `realpathSync.native`가 `%TEMP%`의 8.3 `RUNNER~1`을 풀어버려서 요청 경로랑 문자열이 갈라짐. `databasePathIsSafe`가 그걸 조상 심링크로 보고 protect/unprotect/repair/reclaim/compact를 전부 거절함. 기능이 윈도우에서 한 번도 안 돈 거임.

지금 `path-safety.ts`는 darwin `/var` `/tmp` 앨리어스만 정규화함. 윈도우는 없음. 이번엔 철자가 다를 때 요청 경로를 같은 콜로 다시 캐논하고 두 캐논이 같으면 통과. 진짜 심링크/정션은 다른 데로 가서 거절. `win32`만. POSIX 판결은 안 바꿈.

#2139 POSIX 모드 단언이 가려서 늦게 봄. 예전 빨간 런을 pre-existing으로 읽으면 안 됨. #1729가 2026-08-16에 `dev`에 들어온 뒤임. 윈도우 샤드가 이 픽을 증명하는 유일한 체크임.

점수는 72임. 안 넣으면 2.27 윈도우에서 로그가드가 처음부터 죽음. types.ts/config.ts 안 건드림. 닫으라는 쪽 아님.

해결방안: 윈도우 CI 초록 보고 넣음. 2.27 태그 전이면 #2137이랑 같이. 심링크 거절 테스트 세 방향(OS 철자 / DB 심링크 / 남 형제) 유지.

이 댓글은 grok-bot이 작성했습니다

PR #2144, opened 2026-08-19
fix(claude): gate the shell hook on an installed Claude Code CLI

lidge-jun:
## Summary

The `.zshrc` hook was appended whenever system-env injection succeeded, so a user who never installed Claude Code still got an OpenCodex-owned block sourcing a file for a CLI they do not have — and a stale block survived uninstalling it.

**Credit: @lilinxiong's #2105** is the implementation and test suite carried here.

The hook is now written only when an executable `claude` is discoverable on PATH, and the operation is a reconcile rather than an append: when Claude Code or system-env integration is gone, the OpenCodex-owned block is removed and every user line is left untouched. Empty PATH segments are ignored so a workspace-local file is never mistaken for a durable installation, and a failed removal is reported as failed instead of silently reported as absent.

### One thing worth stating plainly

"Installed" is answered from the PATH of the calling process. A launchd/service context with a stripped PATH can therefore fail to see a `claude` that the user's interactive shell finds, and will remove the hook.

That is the intended failure direction rather than an oversight: removing an OpenCodex-owned block is reversible on the next foreground `ocx start`, whereas leaving a hook pointing at an uninstalled CLI is exactly the stale state this reconciliation exists to clear. Only our own marked block is ever touched. I added that reasoning as a comment on `reconcileShellHook` so the next reader does not have to rediscover it.

## Verification

- **RED-first**: reverting `src/cli/index.ts` and `src/server/system-env.ts` makes the suite fail to even resolve `claudeCodeCliInstalled` — the export does not exist on `dev`.
- `bun test --isolate tests` — 13,542 pass, 0 fail, 10 skip (856 files).
- `bun test --isolate tests/claude-shell-hook.test.ts` — 8 pass, 0 fail, against a fake `HOME`/`PATH`.

## Supersedes

Closes #2105 (@lilinxiong) once merged, with attribution.

## Checklist

- [x] Docs or release notes were updated when needed. (Claude Code guides updated in the locales the original PR touched.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (Writes only an OpenCodex-marked block to the user's `.zshrc`, never removes user lines, and reads no credential. The PATH scan requires a real executable file and skips empty segments, so a workspace-local `claude` cannot be mistaken for an installation. Privacy scan green.)

Closes #2105

## Summary by CodeRabbit

* **New Features**
  * Shell environment hooks are installed only when the Claude Code CLI is available and system environment integration is enabled.
  * Hooks are automatically removed when either condition is no longer met.
  * Claude Desktop does not trigger shell-hook installation.

* **Bug Fixes**
  * Startup and `ocx ensure` now consistently reconcile shell-hook state.
  * Improved handling of hook installation and removal failures, with clearer guidance for manual recovery.

* **Documentation**
  * Updated Claude Code integration guidance across supported languages.


lidge-jun:
## 리뷰 · 우선순위 56 / 80

#2105를 가져온 거임. system-env 주입이 성공하면 `.zshrc`에 훅을 무조건 붙임. Claude Code 없는 사람도 OpenCodex 블록이 `claude`를 소싱함. 지워도 스테일이 남음. 지금은 PATH에서 실행 가능한 `claude`가 있을 때만 쓰고, 없으면 우리 마크 블록만 지움. 유저 줄은 안 건드림. 빈 PATH 세그먼트는 무시해서 워크스페이스 로컬 파일을 설치로 안 봄.

호출 프로세스 PATH로 “설치됨”을 판정함. launchd/서비스가 PATH를 깎으면 훅을 지움. 그게 의도임. 스테일 훅을 남기는 쪽이 더 나쁨. 다음 포그라운드 `ocx start`가 되돌림. `reconcileShellHook` 주석이 그거 적음.

테스트 8개가 fake HOME/PATH. 소스 되돌리면 export가 없어서 스위트 자체가 안 뜸. types.ts/config.ts 안 건드림. #2105는 머지되면 닫음. 원 PR 리베이스하지 말 것.

점수는 56임. 맞는데 2.27 불은 아님. 윈도우 로그가드(#2143) 뒤.

해결방안: 머지 후 #2105 attribution 달고 닫기. 서비스 PATH가 훅을 지우는 건 문서에 한 줄이면 됨.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):

PR #2145, opened 2026-08-19
fix(responses): bridge routed tool_search through the Responses passthrough

### LINKED ISSUE #1950: [Provider compatibility] Responses passthrough is missing the chat-line tool_search rewrite and restore
### Client or integration

Codex App

Any third-party `/v1/responses` gateway reached through Responses passthrough. Reproduced on opencode-go / zen (`deepseek-v4-flash`); the same gap applies to other API-key Responses endpoints (for example DeepSeek official `/responses`).

### OpenCodex version

2.24.2 (`@bitkyc08/opencodex@2.24.2`, published 2026-08-17)

### Endpoint or capability

`/v1/responses` passthrough — Codex private `tool_search` (deferred tool discovery)

On the Responses **passthrough** line (`createResponsesPassthroughAdapter` in `src/adapters/openai-responses.ts`), OpenCodex forwards the Codex body to third-party gateways without the two translations the **chat** line already performs for Codex's private `tool_search` tool. Routed models therefore cannot start deferred-tool discovery, and a `function_call[name=tool_search]` that does appear is returned as-is — the Codex client never sees a `tool_search_call`.

This is a passthrough-line gap, not a provider-specific bug. The chat translation path already has both halves.

### Current behaviour

**1. Request — `{type:"tool_search"}` is not exposed as a callable function.**

Passthrough serializes the incoming tools array without rewriting the private type. Third-party gateways do not implement `type: "tool_search"`, so the model never sees a callable `tool_search` tool and cannot start discovery.

Stock 2.24.2, passthrough to zen / `deepseek-v4-flash`, tools include `{type:"tool_search"}` plus a deferred `codex_app` namespace tool. The model states that `tool_search` is not in its tool list and emits **0** `tool_search_call` items (live sample, translated):

    I don't have a tool_search tool — the only tools in my current environment are:
    - get_weather
    - codex_app__set_thread_title

(`tool_search_call_count: 0`)

**2. Response — `function_call[name=tool_search]` is not restored to `tool_search_call`.**

The chat / collaboration path already restores this shape (`src/bridge.ts` 615–620 / 1024–1031 / 1590–1595, collector in `src/server/responses/collaboration.ts` 111–128). Passthrough does not run that restore: it returns the upstream item unchanged.

Observed passthrough output (with the request-side rewrite applied locally so the model can emit the call; stock 2.24.2 never reaches this step):

    {
      "type": "function_call",
      "name": "tool_search",
      "arguments": "{\"query\": \"change conversation thread title\"}"
    }

The Codex client executes discovery from `tool_search_call`, not from a generic `function_call`.

**Not a defect: namespace.** A zen-native namespaced call already arrives as `{name:"set_thread_title", namespace:"codex_app"}`. Passthrough does not strip that field. This issue is only about `tool_search` rewrite + restore.

### Expected behaviour

On the non-forward Responses line, match the chat line:

- Request: expose `{type:"tool_search"}` to the routed model as a callable function (same schema as `src/responses/parser.ts` 199–214).
- Response: restore the model's `function_call[name=tool_search]` to the private `tool_search_call` shape the client executes.

Canonical OpenAI forward should keep the native `type: "tool_search"` unchanged.

1. Install `@bitkyc08/opencodex@2.24.2`.
2. Point a provider at a third-party `/v1/responses` gateway via per-model `modelAdapters` (API key omitted):

    {
      "providers": {
        "opencode-go": {
          "adapter": "openai-chat",
          "modelAdapters": {
            "deepseek-v4-flash": "openai-responses"
          }
        }
      }
    }

3. In a Codex App session, ask the model to discover a deferred session tool via `tool_search` (for example a `codex_app` thread-title tool that is `defer_loading: true`).
4. Observe: the model reports that `tool_search` is not in its tool list and does not emit `tool_search_call`.

The same tools array sent **directly** to the gateway (no proxy) is not the comparison that matters here: the gateway never advertised `type: "tool_search"` as a function. The missing step is the proxy translation the chat line already does.

### Actual response or error

Stock 2.24.2 passthrough (redacted):

    HTTP 200
    output types: reasoning, message
    tool_search_call_count: 0
    assistant: tool_search is not in the available tool list

When a `tool_search` function call does appear (same local request-side rewrite; stock 2.24.2 never reaches this step), passthrough returns:

    function_call[name=tool_search]

not `tool_search_call`.

### Upstream documentation

- Codex client contract: deferred discovery is a client-executed private type (`type: "tool_search"` in, `tool_search_call` out). It is not a public OpenAI Responses hosted tool. No public specification exists for that private type.
- OpenCodex chat line already implements that contract: `src/responses/parser.ts` 199–214 exposes `tool_search` as a function; `src/bridge.ts` relays the call back as `tool_search_call`.
- Third-party Responses gateways speak generic OpenAI Responses tools (`function`, `web_search`, and similar) and ignore or reject unknown private types. DeepSeek Responses tools table: https://api-docs.deepseek.com/guides/responses_api
- opencode.ai zen: https://opencode.ai/docs/zen/

Expected behaviour is the Codex client contract plus the translation OpenCodex already ships on the chat line — not a claim that third-party gateways should implement `type: "tool_search"` natively.

The chat line already has both halves. Passthrough is missing the same pair:

1. **Request (non-forward / non-canonical-OpenAI branch only).** Next to the existing `promoteClientLoadedTools` / `rewriteRoutedCustomToolsForUpstream` calls in `createResponsesPassthroughAdapter`, rewrite `{type:"tool_search"}` to `{type:"function", name:"tool_search", ...}` using the schema in `src/responses/parser.ts` 199–214. Reuse the current gate; do not add a new flag. Leave canonical OpenAI forward on the native type.
2. **Response (passthrough return path).** Restore `function_call[name=tool_search]` to `tool_search_call` on the JSON and SSE rewrite chains — the same place routed custom-tool restore already hangs. The translation-line restore in `src/bridge.ts` / `src/server/responses/collaboration.ts` does not run on passthrough today.

`promoteClientLoadedTools` (#1602) only merges tools **already returned** by a prior `tool_search_output`. It does not convert the private type, so it cannot start discovery.

**Feasibility (local experiment, not a proposed patch).** On an unmodified 2.24.2 checkout plus a minimal patch at those two sites, a live third-party gateway closed the loop: the model issued `tool_search_call`, discovered the deferred tool, and called the namespaced tool successfully. `web_search` and ordinary `function` calls did not regress.

Related (checked; this is not a duplicate of any of them):

- **#1602** (merged 2026-08-13) — `promoteClientLoadedTools`: promotes tools the client has already loaded. Half of the discovery story; no `tool_search` to function conversion.
- **#1872** (open) — docs question about getting local models to use `tool_search`. Maintainer confirmation that the **chat / routed** path exposes `type: "tool_search"` as a function and converts the call back to `tool_search_call`. That is exactly the pair passthrough does not do. #1872 is the docs thread; this issue is the passthrough implementation gap.
- **#1158** (closed) — same pattern: a third-party Responses gateway rejected a Codex-private tool type (`custom`). The workaround then was to switch the provider to `openai-chat`. Here the private type is `tool_search`, and the chat line already has the translation.
- **#1700** (closed by **#1788**, shipped in 2.23.0) — adjacent fail-closed for an *undeclared* routed tool on passthrough. Different gap. Not a sequel, and not the same as #1872.

### Checks

lidge-jun:
## Summary

The Responses passthrough forwarded Codex's private `{type:"tool_search"}` unchanged to third-party `/v1/responses` gateways, which only understand public function tools. The model never saw a callable `tool_search` and emitted **zero** `tool_search_call` items — deferred tool discovery silently did nothing on every routed provider (#1950).

**Credit: @Ingwannu's #2040** is the implementation and the 578-line test suite carried here.

The private tool is lowered to a public function on noncanonical forward targets, `tool_choice` and replayed history are rewritten to match, and the private `tool_search_call` lifecycle is restored on the way back so Codex sees something it can execute. Canonical ChatGPT forward never rewrites.

### Two corrections for findings still open on #2040

**History-only replay.** A turn may replay `tool_search_call` history *without* re-declaring the tool. The history was lowered either way, but restoration was armed only when a current declaration existed — so the client received a public `function_call` for what it had issued as a private search call, and the round trip stopped matching.

**SSE overflow was not atomic.** Both overflow branches cleared `routedItemIds` along with the pending buffer, then returned every later block unchanged. An item already restored to `tool_search_call` would start emitting `function_call_arguments.*` frames again, giving the client a mixed private/public lifecycle for one call — exactly what this rewrite exists to prevent. Overflow now stops buffering *unknown* frames without forgetting what it already classified.

## Verification

- **RED-first, twice.** Reverting the source fails 7 of @Ingwannu's tests. Separately, restoring **only his original two modules** fails exactly my two new tests (13 pass / 2 fail) — that is what proves the corrections are real rather than restatements.
- `bun test --isolate tests` — 13,551 pass, 0 fail, 10 skip (856 files).
- `bun test --isolate tests/responses-tool-search-repair.test.ts` — 15 pass, 0 fail.

One note on the history test: my first fixture used the item id as `call_id`, which failed. That was my fixture being wrong, not the code — real Codex history carries a separate `call_id`, and the rewrite correctly leaves it alone. Fixed in the test.

## Merge-order note

Also touches `src/server/responses/core.ts`, alongside open PRs #2137, #2104, and #2101. Verified non-overlapping: this change edits the imports, the ~2435 name collection, the ~2894 SSE rewrite list, and the ~3091 JSON restore, while #2137 edits only `resolveResponsesCodexAuth` near line 1088. Complementary with #2142, which edits a different file downstream in the same pipeline.

## Supersedes

Closes #2040 (@Ingwannu) once merged, with attribution.

## Checklist

- [x] Docs or release notes were updated when needed. (Adapter reference in the locales the original touched, plus `structure/04_transports-and-sidecars.md` for the rewrite/restore boundary.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (Rewrites tool declarations and call frames only; no credential path touched, and the canonical ChatGPT gate means native forward behavior is byte-identical. Buffering stays bounded by frame count, byte cap, and translator budget. Privacy scan green.)

Closes #1950

## Summary by CodeRabbit

- **New Features**
  - Added compatibility for `tool_search` when routing Responses requests through noncanonical gateways.
  - Converted tool-search declarations and history into supported function-call formats, then restored them in JSON and streaming responses.
  - Preserved native tool-search behavior for canonical OpenAI forwarding.
  - Improved authentication and header-handling documentation across supported languages.

- **Bug Fixes**
  - Prevented naming collisions and preserved ordinary function tools during conversion.
  - Added safer handling for streaming arguments, malformed JSON, translation limits, and credential forwarding.

### DIFF
--- a/docs-site/src/content/docs/ja/reference/adapters.md
+++ b/docs-site/src/content/docs/ja/reference/adapters.md
@@ -43,8 +43,13 @@ interface ProviderAdapter {
 
 ## `openai-responses`
 
-**対象:** OpenAI **Responses API**。**`passthrough: true`** — 元のリクエスト本文をそのまま渡し、レスポンスを **変換せずに** ストリーミングします。
-**認証:** `forward`（呼び出し元ヘッダー中継）または `key`。
+**対象:** OpenAI **Responses API**。**`passthrough: true`** — 通常は元のリクエストとレスポンスをそのまま渡し、ルーティング先ゲートウェイに必要な限定的な互換変換だけを適用します。
+**認証:** canonical OpenAI `forward` は安全な呼び出し元ヘッダー許可リストだけを中継します。非 canonical な `forward` は呼び出し元の authorization を中継せず、設定済みの静的ヘッダーだけを使用します。`key` は設定済み provider key を使用します。
+
+非 canonical な Responses ゲートウェイには、Codex のクライアント実行型 `tool_search`
+宣言を既存の公開 function tool と衝突しない名前で送り、対応するリクエスト履歴と JSON/SSE
+function call をクライアント向けの非公開 `tool_search` ライフサイクルに復元します。
+canonical OpenAI forward はネイティブな非公開型を維持します。
 
 `key` 認証では、[`retryOn429`](/ja/reference/configuration/) もここに適用されます: プリストリームの
 429 は、翻訳された `openai-chat` / Anthropic リクエスト経路と同様に、他の処理やフェイルオーバーに
--- a/docs-site/src/content/docs/ko/reference/adapters.md
+++ b/docs-site/src/content/docs/ko/reference/adapters.md
@@ -49,9 +49,16 @@ interface ProviderAdapter {
 
 ## `openai-responses`
 
-**대상:** OpenAI **Responses API**. **`passthrough: true`** — 원본 요청 본문을 전달하고 응답을
-**변환하지 않은 채** 스트리밍합니다.
-**인증:** `forward`(호출자 헤더 중계) 또는 `key`.
+**대상:** OpenAI **Responses API**. **`passthrough: true`** — 일반적으로 원본 요청과 응답을
+그대로 전달하되, 라우팅된 게이트웨이에 필요한 좁은 호환성 변환만 적용합니다.
+**인증:** 정규 OpenAI `forward`는 안전한 호출자 헤더 허용 목록만 중계합니다. 비정규
+`forward`는 호출자 authorization을 중계하지 않고 설정된 정적 헤더만 사용하며, `key`는
+설정된 provider 키를 사용합니다.
+
+비정규 Responses 게이트웨이에는 Codex의 클라이언트 실행형 `tool_search` 선언을 공개 function
+도구와 충돌하지 않는 이름으로 전달합니다. 일치하는 요청 기록과 JSON/SSE function call은
+클라이언트용 비공개 `tool_search` 수명 주기로 복원합니다. 정규 OpenAI forward 경로는
+네이티브 비공개 타입을 그대로 유지합니다.
 
 `key` 인증에서는 [`retryOn429`](/ko/reference/configuration/)도 여기에 적용됩니다: 사전 스트림
 429는 번역된 `openai-chat`/Anthropic 요청 경로와 동일하게 다른 처리나 페일오버보다 먼저
--- a/docs-site/src/content/docs/reference/adapters.md
+++ b/docs-site/src/content/docs/reference/adapters.md
@@ -49,9 +49,16 @@ provider — xAI, Kimi, DeepSeek, GLM, Groq, OpenRouter, Ollama (local & cloud),
 
 ## `openai-responses`
 
-**Targets:** the OpenAI **Responses API**. **`passthrough: true`** — forwards the raw request body and
-streams the response back **untranslated**.
-**Auth:** `forward` (relay the caller's headers) or `key`.
+**Targets:** the OpenAI **Responses API**. **`passthrough: true`** — normally forwards the raw request
+body and response, with narrow compatibility rewrites for routed gateways.
+**Auth:** canonical OpenAI `forward` relays only the safe caller-header allowlist; noncanonical
+`forward` uses configured static headers without relaying caller authorization; `key` uses the
+configured provider key.
+
+Noncanonical Responses gateways receive Codex's client-executed `tool_search` declaration as a
+collision-safe public function tool. Matching request history and JSON/SSE function calls are
+translated back to the private `tool_search` lifecycle for the client. Canonical OpenAI forward
+keeps the native private type unchanged.
 
 For `key` auth, [`retryOn429`](/reference/configuration/) applies here too: a pre-stream 429
 waits and replays the identical request on the same key before any other handling, exactly like
--- a/docs-site/src/content/docs/zh-cn/reference/adapters.md
+++ b/docs-site/src/content/docs/zh-cn/reference/adapters.md
@@ -45,9 +45,14 @@ interface ProviderAdapter {
 
 ## `openai-responses`
 
-**目标：** OpenAI **Responses API**。**`passthrough: true`** —— 转发原始请求 body，并把响应
-**不经转换**地流式传回。
-**认证：** `forward`（转发调用方 header）或 `key`。
+**目标：** OpenAI **Responses API**。**`passthrough: true`** —— 通常原样转发请求与响应，仅对
+路由网关应用范围有限的兼容性转换。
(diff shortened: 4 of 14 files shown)


lidge-jun:
## 리뷰 · 우선순위 70 / 80

#2040을 가져오면서 구멍 두 개를 막음. 라우티드 Responses가 Codex 전용 `{type:"tool_search"}`를 그대로 보내서 게이트웨이가 함수 툴로 인식을 못 함. 모델이 `tool_search_call`을 0개 냄. 지연 툴 디스커버리가 라우티드에서 침묵함(#1950). canonical ChatGPT forward는 안 고침.

#2040에 열려 있던 두 개임. 히스토리만 리플레이하면 선언이 없어서 restoration이 안 무장됨. 클라이언트가 전용 search를 냈는데 공개 `function_call`을 받음. SSE overflow가 `routedItemIds`까지 지워서 이미 복원한 아이템이 다시 `function_call_arguments.*`를 냄. 라이프사이클이 섞임. overflow는 미분류 프레임만 버퍼를 멈추고 이미 분류한 건 유지.

`core.ts`를 #2137도 건드림. 이 파일은 1218줄이라 충돌 나면 #2137을 먼저 넣고 이걸 리베이스. 닫으라는 쪽 아님. #2040은 머지되면 닫음. types.ts/config.ts 안 건드림. 분할에 깔림 없음.

점수는 70임. 라우티드 tool_search가 원래 70이었음. 2.27 막판 메가 PR은 아님. #2143 #2137 뒤. 테스트 15패스, 원 모듈만 복원하면 새 테스트 2개가 빨갬.

해결방안: #2137 먼저. 그다음 이 PR. #2040 attribution 달고 닫기. #1950 Closes는 필드 리포트 한 번 보면 됨.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):

PR #2146, opened 2026-08-19
fix(codex): gate account-native models by entitlement

lidge-jun:
## Summary

`gpt-daybreak-blue-latest` lives in the static native set, so catalog sync copied it onto every account selector and Pool could bind a bare Daybreak request to an account whose authenticated roster never contained it. The upstream answered `The 'gpt-daybreak-blue-latest' model is not supported when using Codex with a ChatGPT account.` (#2097).

This makes the authenticated ChatGPT roster the source of truth: entitlement is discovered per account, the gated row is advertised only where an eligible account confirms it, and an account that cannot serve the model is skipped before selection. Discovery failures **fail closed** — the row disappears rather than being offered on unproven evidence.

**Credit: @Ingwannu's #2101** is the implementation, the 333-line entitlement module, and the test suite.

### Stacked on #2137 — merge that first

> **Stack** (merge bottom-up):
> | # | PR | Layer | Review focus |
> |---|----|-------|--------------|
> | 2 | this PR | account entitlement gating | entitlement discovery, catalog projection, Pool/Direct selection |
> | 1 | #2137 ← base | bearer admission predicate | the `substituteMainCredential` contract |
>
> Depends on #2137. Review this PR's diff only.

This is a real dependency, not a convenience: this change passes `substituteMainCredentialForDirect: substituteMainCredential` into `resolveCodexAuthContext`, consuming the exact value #2137 corrects. Landing it on `dev` alone would silently reintroduce #2132 for every routed provider.

It is the only genuine stack edge in this backlog — every other absorbed PR was disjoint and shipped as a sibling.

### Three corrections

**Selector compact missed the wire rewrite.** `accountGatedCompactWireModel` was derived from the caller's raw model string, and an account-qualified selector like `side/gpt-daybreak-blue-latest` does not match the gated map (`slugsEquivalent` reads the account namespace as a routed provider prefix). It still took the native compact endpoint the guard exists to avoid. Now derived from `route.modelId`, the same value `applyCodexAccountGatedWireNormalization` uses.

**Direct callers shared one 64-entry roster cache with main/Pool.** Direct keys are per-credential and unbounded in practice, while main/Pool keys are what the catalog projects from — so a burst of distinct Direct callers evicted the gated row until rediscovery. The two classes now evict separately.

**Comment rot** in `native-models.ts` claimed routing never collapses Daybreak into `gpt-5.6-sol`, which the wire normalization does exactly. Corrected to describe what the code does.

## Verification

[...]

PR #2147, opened 2026-08-19
fix(xai): stream OAuth Grok through Responses

### LINKED ISSUE #1886: [Feature]: FastWire — generalize Fast mode across providers (umbrella)

lidge-jun:
## Summary

OAuth `xai/grok-4.5` and `xai/grok-4.6` Codex `/v1/responses` traffic still used the provider-wide `openai-chat` adapter, while the official Grok CLI catalog declares `api_backend: "responses"`. Chat Completions compatibility holds the stream until the reasoning turn finishes, so **Codex sat blank until the turn was effectively done** (#1886).

**Credit: @olddonkey's #2104**, carried unchanged.

The Responses wire default is declared for those two models and scoped twice — to OAuth auth mode and to responses-shaped inbound traffic. API-key xAI, Chat/Anthropic translation, other Grok models, and any explicit `modelAdapters` override all stay on Chat.

### The 401 replay hole this also closes

Native Responses returns *before* the generic recovery loop, so the OAuth 401 replay never ran on this path. The PR adds the equivalent one-shot: refresh once, rebuild the provider and adapter, replay once. It is a single branch rather than a loop, so a second 401 structurally cannot trigger a second refresh — which is what the `401 then 401` test asserts.

This also fills the same pre-existing hole for Copilot, whose Responses wire defaults already returned before the generic loop.

## Verification

- **RED-first**: reverting the source fails **8** tests across the four suites.
- `bun test --isolate tests` — 13,540 pass, 0 fail, 10 skip (856 files).
- `bun test --isolate` on `server-xai-responses-streaming`, `server-xai-oauth-401-replay`, `fastwire-policy`, `adapter-resolve` — 257 pass, 0 fail.

The streaming test is the load-bearing one: it asserts the first `response.output_text.delta` arrives while `completionReleased === false`, which is the actual user-visible symptom rather than a proxy for it.

## Merge-order note

Touches `src/server/responses/core.ts` alongside #2137, #2145, and #2146. Verified non-overlapping: this is a single insert at ~2622 in the passthrough path after the first upstream fetch, while #2137 edits only `resolveResponsesCodexAuth` (~1088), #2145 edits imports and the ~2399/~2894/~3091 rewrite paths, and #2146 edits ~530-700 and ~2133. Whichever lands second may need a trivial context rebase, nothing more.

## Supersedes

Closes #2104 (@olddonkey) once merged, with attribution.

## Checklist

- [x] Docs or release notes were updated when needed. (`structure/04_transports-and-sidecars.md` records the wire boundary.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. **Touches OAuth credential handling and warrants explicit review per MAINTAINERS.md.** The replay is structurally one-shot, refresh failures are sanitized through the existing public OAuth projector, and the tests assert Windows/UNC/POSIX/`auth.json` path canaries never reach the client. No new persistence, logging, scope, or destination. Privacy scan green.

Closes #1886

## Summary by CodeRabbit

- **New Features**
  - Added authentication-aware routing for supported models, selecting the appropriate Responses or Chat protocol.
  - Added native Responses streaming support for eligible OAuth connections.
  - Added automatic recovery and retry when eligible OAuth requests receive an authorization error.

- **Improvements**
  - Caller service-tier settings are now handled according to model and authentication support, preventing unsupported forwarding.

- **Documentation**
  - Updated transport and service-tier guidance to reflect authentication-specific routing and behavior.

### DIFF
--- a/src/providers/fastwire.ts
+++ b/src/providers/fastwire.ts
@@ -7,7 +7,7 @@ import type {
 } from "../types";
 import { MODEL_ADAPTER_OVERRIDE_ALLOWED } from "../types";
 import { sanitizeLogMetadataString } from "../lib/redact";
-import type { InboundWire, ModelWireDefault } from "./registry";
+import type { InboundWire, ModelWireDefault, ProviderAuthKind } from "./registry";
 
 const SERVICE_TIER_ADAPTERS = new Set(["openai-chat", "openai-responses"]);
 const FAST_WIRE_ADAPTERS: Readonly<Record<FastWire["kind"], ReadonlySet<string>>> = {
@@ -31,6 +31,7 @@ export type FastPolicyAuthTransport =
 
 export interface FastPolicyAuthority {
   readonly providerAdapter: string;
+  readonly providerAuthMode?: ProviderAuthKind;
   readonly fastWireDeclaration: FastWire | null | undefined;
   readonly modelWireOverrideAllowed: boolean;
   readonly authTransport: FastPolicyAuthTransport;
@@ -114,21 +115,33 @@ function registryDefaultForModel(
   defaults: Readonly<Record<string, ModelWireDefault>>,
   modelId: string,
   inbound: InboundWire,
-): string | undefined {
+  authMode: ProviderAuthKind | undefined,
+): { adapter: string; forwardCallerServiceTier?: boolean } | undefined {
   const normalizedModelId = modelId.trim().toLowerCase();
   if (!Object.hasOwn(defaults, normalizedModelId)) return undefined;
   const declared = defaults[normalizedModelId];
   if (declared === undefined) return undefined;
-  if (typeof declared !== "string" && !declared.inbound.includes(inbound)) return undefined;
+  if (typeof declared !== "string") {
+    if (!declared.inbound.includes(inbound)) return undefined;
+    if (declared.authModes && (authMode === undefined || !declared.authModes.includes(authMode))) {
+      return undefined;
+    }
+  }
   const wire = typeof declared === "string" ? declared : declared.wire;
-  return MODEL_ADAPTER_OVERRIDE_ALLOWED.has(wire) ? wire : undefined;
+  if (!MODEL_ADAPTER_OVERRIDE_ALLOWED.has(wire)) return undefined;
+  return {
+    adapter: wire,
+    ...(typeof declared !== "string" && declared.forwardCallerServiceTier !== undefined
+      ? { forwardCallerServiceTier: declared.forwardCallerServiceTier }
+      : {}),
+  };
 }
 
 function resolvePolicyAdapter(
   authority: FastPolicyAuthority,
   modelId: string,
   inbound: InboundWire,
-): { adapter: string; hardPinned: boolean } {
+): { adapter: string; hardPinned: boolean; forwardCallerServiceTier?: boolean } {
   // Hard pins and configured overrides deliberately use the same exact-key semantics as
   // resolveWireProtocolOverride(). Registry defaults alone normalize ids at their boundary.
   const hardPin = Object.hasOwn(authority.hardPins, modelId)
@@ -143,8 +156,21 @@ function resolvePolicyAdapter(
       return { adapter: configured, hardPinned: false };
     }
     if (MODEL_ADAPTER_OVERRIDE_ALLOWED.has(authority.providerAdapter)) {
-      const registryDefault = registryDefaultForModel(authority.registryWireDefaults, modelId, inbound);
-      if (registryDefault !== undefined) return { adapter: registryDefault, hardPinned: false };
+      const registryDefault = registryDefaultForModel(
+        authority.registryWireDefaults,
+        modelId,
+        inbound,
+        authority.providerAuthMode,
+      );
+      if (registryDefault !== undefined) {
+        return {
+          adapter: registryDefault.adapter,
+          hardPinned: false,
+          ...(registryDefault.forwardCallerServiceTier !== undefined
+            ? { forwardCallerServiceTier: registryDefault.forwardCallerServiceTier }
+            : {}),
+        };
+      }
     }
   }
   return { adapter: authority.providerAdapter, hardPinned: false };
@@ -155,7 +181,11 @@ export function resolveFastPolicy(
   modelId: string,
   inbound: InboundWire = "responses",
 ): ResolvedFastPolicy {
-  const { adapter, hardPinned } = resolvePolicyAdapter(authority, modelId, inbound);
+  const { adapter, hardPinned, forwardCallerServiceTier } = resolvePolicyAdapter(
+    authority,
+    modelId,
+    inbound,
+  );
   const exactCapability = exactModelValue(authority.capability.models, modelId);
   const capability = authority.capability.provider === false
     ? false
@@ -173,6 +203,7 @@ export function resolveFastPolicy(
   // tier still needs the final wire's forwarding permission.
   const forwardCallerTier = capability !== false
     && callerWireAvailable
+    && forwardCallerServiceTier !== false
     && (adapter !== "openai-chat" || authority.capability.chatServiceTier === true);
 
   let eligibility: ResolvedFastPolicy["eligibility"];
--- a/src/providers/registry.ts
+++ b/src/providers/registry.ts
@@ -31,9 +31,15 @@ export type InboundWire = "responses" | "chat" | "anthropic";
 
 /**
  * A per-model wire default: a bare string applies to every inbound, while the object
- * form applies only to the listed inbound protocols.
+ * form may scope the default to listed inbound protocols and authentication modes.
  */
-export type ModelWireDefault = string | { wire: string; inbound: readonly InboundWire[] };
+export type ModelWireDefault = string | {
+  wire: string;
+  inbound: readonly InboundWire[];
+  authModes?: readonly ProviderAuthKind[];
+  /** Whether this registry-selected route may relay a caller-owned service_tier. */
+  forwardCallerServiceTier?: boolean;
+};
 
 export interface ResponsesTerminalRepairPolicy {
   /** Quiet time after a structurally complete output graph before synthesizing completion. */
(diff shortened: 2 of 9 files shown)


lidge-jun:
## 리뷰 · 우선순위 64 / 80

OAuth `xai/grok-4.5` `xai/grok-4.6`이 프로바이더 전체 `openai-chat`을 탐. Grok CLI catalog는 `api_backend: "responses"`. Chat Completions이 추론 끝날 때까지 스트림을 붙잡아서 Codex가 턴이 끝날 때까지 빈 화면임. #2104를 가져온 거임.

Responses 기본을 그 두 모델에만 선언하고 OAuth + responses inbound로 두 번 스코프. API-key xAI, Chat/Anthropic 번역, 다른 Grok, 명시 `modelAdapters`는 Chat에 남음. 네이티브 Responses가 제네릭 recovery 루프 전에 리턴해서 OAuth 401 리플레이가 이 패스에서 안 돌았음. `core.ts` ~2622에 한 방: 리프레시 한 번, 프로바이더/어댑터 재빌드, 리플레이 한 번. 두 번째 401이 두 번째 리프레시를 구조적으로 못 냄. Copilot도 같은 구멍이라 같이 막음.

#2137은 `resolveResponsesCodexAuth` ~1088, #2145는 rewrite, #2146은 ~530-700. 이 블록은 패스스루 첫 fetch 다음이라 겹은 컨텍스트. 닫으라는 쪽 아님. types.ts/config.ts 안 건드림. 분할에 깔림 없음.

Closes #1886은 과함. FastWire 우산임. 이건 OAuth Grok 4.5/4.6 Responses 스트림이랑 401 한 방임. #1886을 닫으면 Fast 일반화가 끝난 것처럼 보임. #2104는 이미 닫힘. Refs만.

점수는 64임. 체감 버그인데 2.27 필수 아님. #2143 #2137 뒤.

해결방안: `dev` 베이스라 #2137이랑 따로 넣어도 됨. #1886 Closes 빼고 Refs로. 스트림 테스트가 `completionReleased === false`일 때 첫 delta를 단언하는 거 유지.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):

PR #2148, opened 2026-08-19
fix(providers): allow a baseUrl override for Anthropic and Antigravity

lidge-jun:
## Summary

Anthropic and Google Antigravity were pinned to their registry endpoints, so a user behind a corporate gateway or self-hosted relay could not point them anywhere else. Both now opt into `allowBaseUrlOverride`.

**Credit: @drakonkat's #2109 and #2110**, combined here.

They are combined rather than stacked because they are **not** dependency-ordered — neither runtime path needs the other — but they are not cleanly independent either: they share one byte-identical `destination-policy.ts` change and both must appear in a single registry parity list. `git merge-tree` conflicts on that list if they land separately. One thesis, one PR.

### The gap review raised on both PRs, now closed

These are the only two **OAuth** entries in the opted-in set, so the override receives a subscription bearer. `providerSecureTransportConfigError` required `https` for that reason — but it returned **before** classifying the URL whenever `allowPrivateNetwork` was set:

```ts
if (providerAllowsPrivateNetwork(name, provider)) return null;   // ← before any classification
```

So `http://attacker.example` with `allowPrivateNetwork: true` passed the gate, and the adapters attached the bearer in cleartext to a public host. I reproduced it directly before fixing:

```
public http + optin   -> ALLOWED   (before)
public http + optin   -> REFUSED   (after)
loopback http + optin -> ALLOWED   (unchanged)
private http + optin  -> ALLOWED   (unchanged)
https public          -> ALLOWED   (unchanged)
```

`allowPrivateNetwork` is a statement that a destination is intentionally **local** — it is not a waiver of transport security. The address is now classified first, and the opt-in only excuses `http` for a destination that really is loopback or private. Routing **refuses** rather than downgrading, so a request never reaches an endpoint that would take the token in the clear.

## Verification

- **RED-first**: reverting to @drakonkat's original gate fails the new test (8 pass / 1 fail). The other five matrix cases pass in both directions, which is what makes this a guard rather than a restatement.
- `bun test --isolate tests` — 13,550 pass, 0 fail, 10 skip (857 files).
- `bun test --isolate` on the two override suites plus `provider-registry-parity` — 53 pass, 0 fail.

## Residual, stated rather than hidden

An `https` override to an arbitrary host is allowed **by design** — that is the feature. The token still leaves for a user-chosen destination, which is the operator's decision to make; this change only ensures it is never sent in the clear. Key and local providers (Ollama, vLLM, LM Studio, LiteLLM, Moonshot, Qwen, Alibaba) are untouched: they are not `authKind: "oauth"`, so the gate never fires for them.

## Supersedes

Closes #2109 and #2110 (@drakonkat) once merged, with attribution.

## Checklist

- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. **This changes where an OAuth bearer may be sent and warrants explicit review per MAINTAINERS.md.** The cleartext bypass above is the reason: it was reachable on both original PRs and is fixed and pinned here. Non-OAuth providers are unaffected, loopback/private relays keep working through the explicit opt-in, and no credential is logged. Privacy scan green.

Closes #2109
Closes #2110

## Summary by CodeRabbit

* **New Features**
  * Added base URL override support for Anthropic OAuth and Google Antigravity providers.
  * Added secure transport validation for overridden provider endpoints.
  * HTTPS is required for public destinations; explicitly permitted local or private HTTP relays remain supported.

* **Bug Fixes**
  * Improved handling of missing, unresolved, or unsafe provider endpoint overrides.
  * Added warnings when configured base URLs are discarded.


lidge-jun:
## 리뷰 · 우선순위 58 / 80

지금 `registry.ts` anthropic oauth랑 `google-antigravity`에 `allowBaseUrlOverride`가 없음. 사내 게이트웨이/셀프호스트가 못 꺾음. 플래그 두 개 켬. #2109 #2110 합친 거임. `destination-policy.ts` 한 줄이 같아서 따로 넣으면 parity 리스트에서 충돌남.

OAuth 옵트인이라 오버라이드에 구독 bearer가 감. 예전 게이트가 `allowPrivateNetwork`면 URL 분류 전에 리턴해서 `http://attacker.example` + optin이 통과함. 지금은 `providerSecureTransportConfigError`가 분류 먼저. 진짜 루프백/프라이빗만 http 허용. 공개 http는 거절. 다운그레이드 아님. 키/로컬(Ollama vLLM 등)은 `authKind` oauth가 아니라서 안 탐.

문제: `devlog/_plan/260820_bug_pr_backlog_consolidation/` 마크다운 8개가 이 PR에 들어옴. baseUrl 테제가 아님. 코드 리뷰가 플랜 덤프랑 섞임. 빼거나 다른 커밋으로. types.ts/config.ts 안 건드림. 분할에 깔림 없음. #2109 #2110은 이미 닫힘.

https 임의 호스트는 의도임. 토큰이 유저가 고른 데로 감. 이 가드는 클리어텍스트만 막음.

점수는 58임. 시큐어 트랜스포트 가드는 좋은데 2.27 불 아님. 로그가드/bearer 뒤.

해결방안: 플랜 파일 빼고 머지. 테스트 매트릭스(public http 거절 / loopback http 허용) 유지.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
The base-URL override direction changes credential destinations and requires a focused security review, but the current head is not reviewable as a merge unit.

Two blockers remain:

1. The branch is 16 commits behind the current `dev` tip (`caf20353f`), so its CI does not validate the current integration state.
2. The diff carries eight `devlog/_plan/260820_bug_pr_backlog_consolidation/*` planning files that are unrelated to the provider destination-policy change. Remove those inherited planning commits so the PR contains only the destination policy, registry changes, and their focused tests.

After a clean rebase, exact-head CI, and verification that authenticated requests cannot be redirected outside the operator-approved destination policy, re-request security review.

PR #2149, opened 2026-08-19
fix(oauth): reject superseded login credential commits

lidge-jun:
## Summary

A cancelled login could still persist its credential after a newer flow had taken ownership of the provider, so the newer login's token was silently replaced by the one the user had already abandoned.

**Credit: @Ingwannu's #2053**, carried unchanged.

Ownership is checked at the **synchronous persist boundary** rather than before the work: `mutateStore` takes an `assertBeforePersist` hook that runs inside the file lock, after `fn(store)` and before `persist(store)`. An in-memory mutation from a superseded flow is discarded rather than written — the only placement that cannot lose the race, since the in-memory object is thrown away when persist is skipped.

Ownership is identity-checked against the flow's own `AbortController` (`loginAbort.get(provider) !== abort`), so a newer login *replacing the entry* is what invalidates the older one. Not a timestamp, not a flag either side could read stale.

Kiro replacements additionally wait for a cancelled CLI flow to finish rolling back before a new login may start.

## Verification

- **RED-first**: reverting `src/oauth/` fails 2 tests, including the ownership round-trip that queues a blocking `mutateStore`, starts flow 1, cancels it, starts flow 2, and proves only `access-2`/`account-2` is stored.
- `bun test --isolate tests` — 13,536 pass, 0 fail, 10 skip (855 files).
- `bun test --isolate tests/oauth-public-surface.test.ts tests/codex-auth-context.test.ts` — 61 pass, 0 fail.

## Two things a reviewer should check rather than take on faith

**The reauth path is wired but not covered.** #2053's description claims reauth coverage; the diff wires `saveAccountCredential(..., assertBeforePersist)` but adds no reauth-specific test. The wiring is correct as far as I can read it, and I did not add a test I could not make meaningfully fail — but it is untested, and I would rather say so than let the claim stand.

**`OAuthLoginSupersededError` is not in the public error allowlist**, so it projects to the generic auth-failure string. That is the safe direction (no internal detail leaks), but it means a user who hits this sees a generic message rather than "your earlier login was superseded".

**A never-finishing Kiro rollback blocks all replacements**, by design. That is a deliberate trade — better than racing — but it is a liveness property worth an explicit second opinion.

## Supersedes

Closes #2053 (@Ingwannu) once merged, with attribution.

## Checklist

- [x] Docs or release notes were updated when needed. (No user-facing behavior beyond a cancelled login no longer overwriting a newer one.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. **This is a C4 credential-store change and MAINTAINERS.md mandates security review — please do not merge on my verification alone.** What I can attest: the assert runs inside the existing file lock so it cannot race, no credential value or account id is logged, the error projects through the existing public vocabulary, and the failure direction is refuse-to-persist rather than persist-and-hope. Privacy scan green.

Closes #2053

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented canceled or superseded OAuth login attempts from saving credentials.
  * Improved handling when replacing an active login flow, ensuring the latest attempt remains in control.
  * Added safeguards so Kiro login waits for a canceled external login process to finish before restarting.
* **Tests**
  * Added coverage for OAuth flow replacement, cancellation, and credential persistence behavior.

### DIFF
--- a/src/oauth/index.ts
+++ b/src/oauth/index.ts
@@ -322,6 +322,13 @@ export class OAuthReauthIdentityUnverifiedError extends Error {
   }
 }
 
+class OAuthLoginSupersededError extends Error {
+  constructor() {
+    super("OAuth login was superseded before credential persistence");
+    this.name = "OAuthLoginSupersededError";
+  }
+}
+
 /** Project arbitrary OAuth failures onto the small, stable public error vocabulary. */
 export function publicOAuthAuthenticationErrorMessage(error: unknown): string {
   if (error instanceof OAuthMutationBusyError) {
@@ -1096,6 +1103,7 @@ interface RunLoginDeps {
   settleKiroLoginTransaction?: typeof settleKiroLoginTransaction;
   removeAccount?: typeof removeAccount;
   setActiveAccount?: typeof setActiveAccount;
+  assertCurrentOwner?: () => void;
 }
 
 /** Roll back only accounts created by this forced login, preserving concurrent refreshes of others. */
@@ -1145,6 +1153,7 @@ export async function runLogin(
   const cred: OAuthCredentials = rawCred.source ? rawCred : { ...rawCred, source: "oauth" };
   const settleKiroTransaction = deps.settleKiroLoginTransaction ?? settleKiroLoginTransaction;
   try {
+    deps.assertCurrentOwner?.();
     // Validate the provider row before credential persistence. A namespace claimed during the
     // credential write is handled again below before the latest row is re-upserted.
     if (provider !== "chatgpt") {
@@ -1165,10 +1174,13 @@ export async function runLogin(
       if (!identityMatches) {
         throw new OAuthReauthIdentityMismatchError();
       }
-      await (deps.saveAccountCredential ?? saveAccountCredential)(provider, opts.reauthAccountId, cred);
+      await (deps.saveAccountCredential ?? saveAccountCredential)(provider, opts.reauthAccountId, cred, {
+        assertBeforePersist: deps.assertCurrentOwner,
+      });
     } else {
       await (deps.saveCredential ?? saveCredential)(provider, cred, {
         preserveIdentityless: opts?.forceLogin === true,
+        assertBeforePersist: deps.assertCurrentOwner,
       });
     }
     if (provider !== "chatgpt") {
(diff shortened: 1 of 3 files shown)


lidge-jun:
## 리뷰 · 우선순위 62 / 80

지금 `mutateStore`가 `fn(store)` 다음에 무조건 `persist(store)` 함. `startLoginFlow`의 `abandonIfNotOwner`는 `loginAbort` 맵 아이덴티티만 봄. 취소된 로그인이 새 플로우가 주인된 뒤에 디스크에 씀. 새 토큰이 버린 토큰으로 덮임. #2053을 가져온 거임.

`assertBeforePersist`를 파일 락 안에서 persist 직전에 둠. `loginAbort.get(provider) !== abort`면 `OAuthLoginSupersededError`. 메모리 뮤테이션은 버려짐. 타임스탬프 아님. AbortController 아이덴티티. Kiro는 롤백 끝날 때까지 `kiroLoginSettling`으로 교체 로그인 거절.

재인증 경로는 `saveAccountCredential(..., assertBeforePersist)` 배선만 있고 테스트 없음. `OAuthLoginSupersededError`가 public allowlist에 없어서 제네릭 메시지. 안전 방향임. 내부 디테일 안 샘. Kiro 롤백이 안 끝나면 교체가 영원히 막힘. 의도. types.ts/config.ts 안 건드림. 분할에 깔림 없음. #2053은 이미 닫힘.

점수는 62임. C4 크레덴셜 레이스라 넣는 게 맞음. 2.27 블로커는 아님. #2143 #2137 뒤.

해결방안: 머지. 재인증 테스트는 후속. public 메시지 바꿀 거면 allowlist에 명시적으로.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
The superseded-login commit guard is a security-boundary change, so I cannot approve it on a stale head. The current branch is 16 commits behind `dev` (`caf20353f`), and its existing validation predates the current OAuth/store integration state.

Please rebase the focused three-file change onto the latest `dev`, rerun the OAuth race/public-surface regressions and full exact-head CI, then re-request the required security review. I will review the resulting credential generation and commit boundary rather than approve an obsolete diff.

PR #2150, opened 2026-08-19
fix(google): preserve and replay thought signatures across stream chunks

lidge-jun:
## Summary

Antigravity thinking models return HTTP 400 on multi-turn tool loops: historical tool calls are replayed without their `thought_signature`, so `exec` / `wait` / agent tools break once the history is long enough to matter (#2125).

**Credit: @agentHits's #2127**, carried unchanged.

Two holes, both in the #897 *"pair the signature to the next call and drop the rest"* contract:

- `pendingThoughtSig` was cleared by the **first** `functionCall`, so a parallel tool turn left every call after the first unsigned.
- A signature and its function call can arrive in **different SSE chunks**, so the pending value was lost at the chunk boundary.

The parser also now reads snake_case `thought_signature` alongside the camelCase form, which is what the upstream actually sends on some turns.

### Why this is here after being triaged below the line

I scored this below my absorb threshold from the title and metadata, and that was wrong. A re-scoring pass reading the actual diff put it at **83** — a live 400 that makes a core provider path unusable, with exact upstream errors, a matching diagnosis, and a live 4-turn replay confirming the fix. "Adapter bookkeeping" as a title hid a core-path break. Correcting it rather than leaving it closed-by-omission.

## Verification

- **RED-first**: reverting `src/adapters/` fails **4** tests, including the chunk-boundary case that is the harder of the two holes.
- `bun test --isolate tests` — 13,538 pass, 0 fail, 10 skip (855 files).
- `bun test --isolate` on `google-antigravity-replay` + `google-signature-history-roundtrip` — 78 pass, 0 fail.
- `bun run typecheck` — clean.
- `bun run privacy:scan` — passed.

## Not in scope

The compaction / `providerMetadata` strip is a named leftover on the original PR, not part of this slice, and it stays open.

## Supersedes

Closes #2127 (@agentHits) once merged, with attribution. That PR was still a draft; the implementation is what mattered and it is complete.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed. (Restores the documented signature-pairing contract; nothing user-facing changes shape.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (Carries an opaque upstream signature token through replay; no credential, auth, or destination surface touched, and the value is never logged. Privacy scan green.)

Closes #2125

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of thought signatures during streamed and non-streamed responses.
[...]

Issue #1875, opened 2026-08-16
[Feature]: enable xAI API-key Priority Processing in Codex Fast mode


olddonkey:
## Area

Provider adapters

## What are you trying to accomplish?

Use Codex Fast mode with Grok models routed through the built-in `xai` provider when that provider is using an xAI API key. xAI documents Priority Processing for text inference by sending `service_tier: "priority"` on both Chat Completions and Responses requests.

The expected workflow is that selecting Fast in Codex, or enabling the existing OpenCodex Fast setting, makes an xAI API-key request use xAI Priority Processing while Standard leaves the field absent/default.

## What prevents this today?

OpenCodex now has provider/model-scoped service-tier capability resolution, catalog publication, and Chat serialization, but the built-in `xai` registry entry declares neither `chatServiceTier` nor `modelSupportsServiceTier`. Consequently `xai/*` catalog rows do not advertise Fast and the runtime does not inject or forward the priority tier.

The preset also multiplexes two transports under one provider id:

- API-key mode stays on `https://api.x.ai/v1`, where Priority Processing is officially documented.
- OAuth mode is rewritten to `https://cli-chat-proxy.grok.com/v1`, where the same capability is not established by the public API documentation.

A provider-wide unconditional opt-in would therefore overstate the unverified OAuth transport.

## What should OpenCodex do?

OpenCodex should make service-tier capability sensitive to the effective xAI authentication transport:

- enable verified `service_tier` support for the canonical xAI API-key transport;
- publish the Codex Standard/Fast selector for eligible `xai/*` rows only in API-key mode;
- inject/forward `service_tier: "priority"` when Fast is selected and remove it when Fast is explicitly disabled;
- keep the Grok OAuth/CLI subscription transport unadvertised and uninjected unless that exact endpoint is separately verified;
- preserve explicit user opt-outs and keep catalog/runtime decisions driven by the same capability resolver;
- avoid inferring support merely from a Grok model name or from the shared `openai-chat` adapter.

## Example usage or interface

Given an xAI API-key configuration:

```json
{
  "defaultProvider": "xai",
  "providers": {
    "xai": {
      "adapter": "openai-chat",
      "baseUrl": "https://api.x.ai/v1",
      "authMode": "key"
    }
  },
  "fastMode": true
}
```

the routed catalog entry for `xai/grok-4.6` should advertise the `priority`/Fast tier and the effective Chat Completions body should contain:

```json
{
  "model": "grok-4.6",
[...]

PR #2151, opened 2026-08-19
fix(chat): apply the Fast capability gate to the native passthrough

### LINKED ISSUE #1886: [Feature]: FastWire — generalize Fast mode across providers (umbrella)

lidge-jun:
## Summary

Native `/v1/chat/completions` decided `service_tier` from `chatServiceTier` alone, so a provider declaring `supportsServiceTier: false` — or an exact-model `false` — still had the field forwarded. That is **fail-open** onto upstreams that reject the field or bill differently for it (#1886).

**Credit: @olddonkey's #2075**, rebased onto current `dev`.

The native passthrough now runs through the same resolved Fast policy the rest of the chat path uses, so a fail-closed or per-model declaration is honored on every surface rather than only the translated ones.

### Why this is here after being triaged below the line

Like #2150, I scored this below my absorb threshold from metadata — it showed `CONFLICTING`, and I treated that as the reason not to look further. A re-scoring pass reading the diff put it at **67**. The conflict was why it *could not merge*, not why it *scored low*, and those are different questions. Rebasing turned out to be one import line.

## The rebase, stated exactly

The only manual resolution was an import list in `src/adapters/openai-chat.ts`: `dev` had since added `AdapterTierMetadata` while this PR adds `decideTier` and `ResolvedFastPolicy`. Both sides are kept — no logic was dropped or reinterpreted. Everything else applied cleanly.

## Verification

- **RED-first**: reverting `src/` fails the characterization test *"native chat passthrough honors exact-model false"* — which the author had flipped from a documented known-bug into a passing assertion, so it is the exact contract at stake.
- `bun test --isolate tests` — 13,552 pass, 0 fail, 10 skip (856 files).
- `bun test --isolate` on `fastwire-characterization-wire`, `openrouter-provider-routing`, `fastwire-policy` — 295 pass, 0 fail.
- `bun run typecheck` — clean (this is what confirms the import resolution).

## Supersedes

Closes #2075 (@olddonkey) once merged, with attribution.

## Checklist

- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (Changes which request field is forwarded upstream, closing a fail-open path. No credential or auth surface touched. Privacy scan green.)

Closes #1886

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of service-tier settings for Chat and passthrough requests.
  * Unsupported model configurations now remove incompatible caller-provided tier settings.
  * Fast-mode behavior remains consistent after provider key failover.
  * Valid caller tier values are preserved when supported by policy.
  * Native and passthrough Chat requests now apply consistent tier decisions.

* **Tests**
  * Added coverage for provider and model support, caller tiers, Fast-mode options, failover, and request-path consistency.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -14,12 +14,15 @@ import { buildNonOpenAIToolCatalogNudgeForTools, shouldInjectNonOpenAIToolCatalo
 import { openRouterProviderPayload, resolveOpenRouterRouting } from "../providers/openrouter-routing";
 import {
   canForwardForeignServiceTierForChatModel,
+  fastPolicyForModel,
   supportsServiceTierForModel,
 } from "../providers/service-tier";
 import {
   canonicalFastTierMarker,
   createAdapterTierMetadata,
+  decideTier,
   type AdapterTierMetadata,
+  type ResolvedFastPolicy,
 } from "../providers/fastwire";
 import { openaiChatCompletionsUrl } from "./openai-chat-url";
 import { stripResponsesOnlyEncryptedMarker } from "./responses-tool-schema";
@@ -96,6 +99,8 @@ export function buildOpenAIChatPassthroughRequest(
   rawBody: Record<string, unknown>,
   modelId: string,
   stream: boolean,
+  fastPolicy: ResolvedFastPolicy = fastPolicyForModel(provider, modelId, undefined, "chat"),
+  fastMode?: boolean,
 ): AdapterRequest {
   const { url, headers, hasCredential } = openAIChatTransport(provider);
 
@@ -123,7 +128,16 @@ export function buildOpenAIChatPassthroughRequest(
   // `<listed>:<tag>` siblings the operator never opted out, silently returning prose.
   if (provider.noStructuredOutputModels?.includes(modelId)) delete body.response_format;
 
-  if (provider.chatServiceTier && rawBody.service_tier !== undefined) {
+  // Run the same complete Fast policy as the translated Chat path, including explicit
+  // fastMode and foreign-tier handling. On inherited canonical Fast, the passthrough still
+  // retains the caller's exact spelling; forced Fast uses the policy-owned wire value.
+  const callerTier = typeof rawBody.service_tier === "string" ? rawBody.service_tier : undefined;
+  const tierDecision = decideTier(fastPolicy, fastMode, callerTier);
+  if (tierDecision.kind === "set") {
+    body.service_tier = fastMode === undefined && canonicalFastTierMarker(callerTier) !== undefined
+      ? callerTier
+      : tierDecision.value;
+  } else if (tierDecision.kind === "forward-caller" && rawBody.service_tier !== undefined) {
     body.service_tier = rawBody.service_tier;
   }
   if (provider.promptCacheKey && rawBody.prompt_cache_key !== undefined) {
@@ -1290,6 +1304,23 @@ function thinkingBudgetForEffort(parsed: OcxParsedRequest, reasoningEffort: stri
   return fraction === undefined ? undefined : Math.max(1, Math.floor(maxBudget * fraction));
 }
 
+function canSerializeOpenAIChatServiceTier(
+  provider: OcxProviderConfig,
+  modelId: string,
+  serviceTier: unknown,
+  tierDecision?: OcxParsedRequest["options"]["tierDecision"],
+): boolean {
+  if (serviceTier === undefined) return false;
+  if (tierDecision !== undefined) {
+    return tierDecision.kind === "set" || tierDecision.kind === "forward-caller";
+  }
+  const callerTier = typeof serviceTier === "string" ? serviceTier : undefined;
+  const callerCanonicalFast = canonicalFastTierMarker(callerTier) !== undefined;
+  const capability = supportsServiceTierForModel(provider, modelId);
+  const callerTierForwardAllowed = canForwardForeignServiceTierForChatModel(provider, modelId);
+  return callerTierForwardAllowed || (callerCanonicalFast && capability === true);
+}
+
 export function createOpenAIChatAdapter(provider: OcxProviderConfig): ProviderAdapter {
   return {
     name: "openai-chat",
@@ -1312,13 +1343,12 @@ export function createOpenAIChatAdapter(provider: OcxProviderConfig): ProviderAd
       // unclassified Chat routes remain behind the caller-forwarding opt-in.
       const serviceTier = parsed.options.serviceTier;
       const tierDecision = parsed.options.tierDecision;
-      const callerCanonicalFast = canonicalFastTierMarker(serviceTier) !== undefined;
-      const callerTierForwardAllowed = canForwardForeignServiceTierForChatModel(provider, parsed.modelId);
-      const canonicalFastCapability = callerCanonicalFast
-        && supportsServiceTierForModel(provider, parsed.modelId) === true;
-      const canSerializeServiceTier = tierDecision?.kind === "set"
-        || tierDecision?.kind === "forward-caller"
-        || (tierDecision === undefined && (callerTierForwardAllowed || canonicalFastCapability));
+      const canSerializeServiceTier = canSerializeOpenAIChatServiceTier(
+        provider,
+        parsed.modelId,
+        serviceTier,
+        tierDecision,
+      );
       if (canSerializeServiceTier && serviceTier !== undefined) {
         body.service_tier = serviceTier;
       }
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -1314,11 +1314,14 @@ function canSerializeOpenAIChatServiceTier(
   if (tierDecision !== undefined) {
     return tierDecision.kind === "set" || tierDecision.kind === "forward-caller";
   }
+  // No decision from the router means this call did not go through the tier state machine, so
+  // ask that machine rather than re-deriving a looser answer beside it. The previous fallback
+  // returned true whenever foreign forwarding was allowed at all, which let a caller tier
+  // reach the wire in cases `decideTier` would have dropped — the two paths disagreeing is
+  // precisely the bug, so there is now only one authority.
   const callerTier = typeof serviceTier === "string" ? serviceTier : undefined;
-  const callerCanonicalFast = canonicalFastTierMarker(callerTier) !== undefined;
-  const capability = supportsServiceTierForModel(provider, modelId);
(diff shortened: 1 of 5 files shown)


lidge-jun:
## 리뷰 · 우선순위 60 / 80

지금 `openai-chat.ts` 네이티브 패스스루가 `provider.chatServiceTier`만 보고 `service_tier`를 그대로 보냄. `supportsServiceTier: false`거나 모델 exact false여도 필드가 나감. fail-open임. 업스트림이 거절하거나 다르게 과금함. 번역 경로는 Fast policy를 쓰는데 네이티브 `/v1/chat/completions`만 빠짐. #2075를 현재 `dev`에 리베이스한 거임.

`buildOpenAIChatPassthroughRequest`가 `fastPolicyForModel` + `decideTier`를 탐. `chat-native.ts`가 키 페일오버 뒤에도 같은 빌더를 씀. 상속 canonical Fast는 호출자 철자 유지. forced Fast는 폴리시 값. 번역 경로 `canSerializeOpenAIChatServiceTier`로 한곳으로 모임. 임포트는 `AdapterTierMetadata`랑 `decideTier` 둘 다 살림. 로직 안 버림.

Closes #1886은 과함. FastWire 우산임. 이건 네이티브 챗 패스스루 한 면만. #2147도 같은 우산을 닫으려 했음. 둘 다 Refs만. #2075는 이미 닫힘. types.ts/config.ts 안 건드림. 분할에 깔림 없음. 닫으라는 쪽 아님.

점수는 60임. 구멍은 맞음. 2.27 필수 아님. Fast A1은 이미 `dev`에 있음. #2143 #2137 뒤.

해결방안: 머지. #1886 Closes 빼고 Refs로. 특성화 테스트 "native chat passthrough honors exact-model false" 유지. Fast 일반화(#1875 B2)는 별 PR.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):

PR #2155, opened 2026-08-19
fix(openai-chat): tolerate non-string padding repeats on resolved stream tool calls

waw4303:
### Problem
Some OpenAI-compatible providers/gateways (e.g. `opencode.ai/zen/go/v1` streaming `deepseek-v4-flash`) emit the function envelope on continuation deltas with non-string values (such as an empty/unexpected object in `delta.tool_calls[i].function.name` or `arguments`) rather than repeating `null` as handled in #1731.

Because `isInvalidStreamStringField` is checked unconditionally at the top of the loop, these malformed continuation repeats cause the entire stream to terminate immediately via `invalidToolCallsEvent` (HTTP 400 `upstream response contained invalid tool calls`), even when the tool call has already received its canonical name and arguments from earlier chunks.

### Solution
Generalize the padding tolerance logic established in #1731:
1. Validate `rawFunction` type defensively while retaining the fast-fail check for non-record structures.
2. Check `isInvalidStreamStringField` for `name` and `arguments` only when the accumulator does not yet hold the canonical value for that call:
   - If the call is already named, a non-string `name` repeat is safely ignored as stream padding.
   - If the call is still unnamed / unparameterized, non-string values continue to fail-closed immediately (#1531).
3. Ensure only genuine `string` deltas reach `call.args` accumulation so non-string repeats cannot pollute arguments.

### Verification
- Added 3 regression test cases in `tests/openai-chat-hardening.test.ts`:
  - `non-string repeated name/arguments on a resolved call are padding, not terminal` (passes)
  - `non-string name on a call without its canonical value still fails closed` (passes)
  - `non-string arguments on a call that has accumulated nothing still fails closed` (passes)
- `bun test tests/openai-chat-hardening.test.ts` passes (57/57).
- `bun test tests/openai-chat-stream.test.ts tests/openai-chat-parallel-stream.test.ts tests/openai-chat-eof.test.ts` passes (59/59).
- `bun run typecheck` (`tsc --noEmit`) passes cleanly.
- Verified against live failing deepseek-v4-flash stream trace (returns `200 OK` with complete tool call payload).

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of streamed tool calls with malformed or repeated metadata.
  * Prevented invalid argument values from being incorrectly converted into text.
  * Ensured incomplete or invalid tool calls fail with clear upstream errors.

* **Tests**
  * Added coverage for malformed tool-call names, arguments, and continuation data.
  * Verified valid established calls can complete despite harmless padding data.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -1628,20 +1628,9 @@ export function createOpenAIChatAdapter(provider: OcxProviderConfig): ProviderAd
               // name or arguments value fails closed through the #1325 channel here rather
               // than escaping later as a TypeError from string handling at flush time.
               const rawFunction = (rawToolCall as { function?: unknown }).function;
-              if (rawFunction !== undefined && rawFunction !== null) {
-                if (!isRecord(rawFunction)) {
-                  logInvalidToolCalls("stream", rawToolCalls);
-                  return yield* terminateWithError(invalidToolCallsEvent(rawToolCalls, "stream", pendingUsage));
-                }
-                const rawName = rawFunction.name;
-                const rawArguments = rawFunction.arguments;
-                // Some OpenAI-compatible streamers repeat already-sent fields as null on
-                // continuation deltas. Treat only null/undefined as absent; every other
-                // non-string value still fails closed before entering the accumulator.
-                if (isInvalidStreamStringField(rawName) || isInvalidStreamStringField(rawArguments)) {
-                  logInvalidToolCalls("stream", rawToolCalls);
-                  return yield* terminateWithError(invalidToolCallsEvent(rawToolCalls, "stream", pendingUsage));
-                }
+              if (rawFunction !== undefined && rawFunction !== null && !isRecord(rawFunction)) {
+                logInvalidToolCalls("stream", rawToolCalls);
+                return yield* terminateWithError(invalidToolCallsEvent(rawToolCalls, "stream", pendingUsage));
               }
               if (isInvalidStreamStringField(tc.id)) {
                 logInvalidToolCalls("stream", rawToolCalls);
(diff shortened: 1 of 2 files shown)


waw4303:
@coderabbitai review

lidge-jun:
## 리뷰 · 우선순위 54 / 80

지금 `openai-chat.ts` 스트림 루프가 `isInvalidStreamStringField`를 누적 전에 무조건 봄. #1731은 null 반복만 패딩으로 봄. 빈 객체 같은 non-string은 바로 `invalidToolCallsEvent` 400. zen/go `deepseek-v4-flash`가 continuation에 `function.name`을 객체로 반복함. 앞 청크에서 이름 받아놓고도 스트림이 죽음.

이번엔 레코드가 아닌 `function`만 즉시 실패. 이름/인자는 아직 canonical이 없을 때만 fail-closed (#1531). 이름 붙은 콜의 non-string 반복은 무시. 누적에는 string만 들어감. `[object Object]`가 args에 안 붙음. 테스트 있음. 초안임. 체크리스트 비어 있음.

구멍: 인자 검사도 `!hasCanonicalName`에 묶임. 이름만 있고 args가 빈 콜이 나중에 객체 args만 내면 침묵하고 빈 툴콜로 끝남. 패딩 허용의 대가임. `diagnoseInvalidToolCalls`는 아직 무조건 검사라 진단이랑 누적이 어긋날 수 있음. types.ts/config.ts 안 건드림. 분할에 깔림 없음. 닫으라는 쪽 아님.

점수는 54임. 실사용 스트림 400인데 드래프트고 2.27 아님. #2143은 이미 `dev`에 들어갔음.

해결방안: args 검사는 `call.argsBytes === 0` 쪽이 계약이랑 맞음. 진단 헬퍼도 같은 조건으로. CI 초록이면 드래프트 해제. #1731 후속으로 적으면 됨.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
Thank you @waw4303 — absorbed as #2165. Your trace was correct and the fix direction was right: the field check ran before the pending-call lookup, so the adapter could not see that the same index already held that value in canonical form, and the whole turn died with a 502 instead of running the tool.

Two refinements on top of it.

**`arguments` was gated on the call having a canonical *name*.** A name is not evidence about the arguments field. A chunk that establishes `name: "terminal"` with no `arguments`, followed by `arguments: {command:"ls"}`, would have been silently dropped — a real payload the model meant to send. #2165 adds `sawArgumentsString` to the pending call, so each field is judged on its own provenance. An empty string counts, since it proves the upstream sent the field with the right wire type.

**`id` stayed unconditionally terminal**, even after a canonical id was stored. It now follows the same per-field rule as the other two.

Also: the diagnostic is now passed from the rejection site instead of rescanned. A stateless rescan stops at the first structurally odd value, so a stream with accepted padding on call 0 and a real defect on call 1 reported `callIndex=0` — the wrong call.

Your positive regression is carried over; it is non-vacuous and reproduces the exact 502 on unpatched `dev`. Your three fail-closed cases are kept as guards, and they matter: they prove the replacement did not widen tolerance while fixing the bug.

Full suite 13562 pass / 0 fail. Closing in favor of #2165, credited to you.

PR #2027, opened 2026-08-18
fix(providers): gate OpenCode Go quota on the canonical base URL

yzxcj797:
## Summary

Fixes #1924.

## Root cause (as reported)

The OpenCode Go quota probe gated on the provider's **name**:

```ts
if ((provider.authMode ?? "key") === "key" && name === "opencode-go") {
  return fetchOpenCodeGoQuota(name, provider);
}
```

so in multi-account setups — sibling providers pointing at the canonical `https://opencode.ai/zen/go/v1` for subscription balancing — only the provider literally named `opencode-go` produced quota rows; the dashboard showed no rate-limit panel for the siblings. The neighboring branches (Kimi, A6api, CommandCode) already gate on their canonical base URLs, and `fetchOpenCodeGoQuota` itself re-validates `isCanonicalOpenCodeGoBaseUrl(config.baseUrl)` before sending the bearer key, making the name check both too narrow and redundant.

## Fix

```diff
-if ((provider.authMode ?? "key") === "key" && name === "opencode-go") {
+if ((provider.authMode ?? "key") === "key" && isCanonicalOpenCodeGoBaseUrl(provider.baseUrl)) {
```

Safe against key leakage to non-canonical hosts: the fetch still re-validates the canonical URL before the bearer key leaves.

## Tests

Two regression tests in `tests/opencode-go-quota.test.ts`:

- the exact repro: a sibling provider named `opencode-go-2` on the canonical base URL now reports quota alongside `opencode-go` (both rows present, both fetches made) — on `main` this yields only one row;
- a provider named `opencode-go` on a **non-canonical** base URL never probes `opencode.ai` with the key.

(Couldn't run the bun test suite locally on this Windows checkout; relying on CI.)

## Review readiness checklist

This PR stays in draft until every box below is ticked. Tick all four boxes once the requirements are met:

- [ ] All CI tests are green on my local testing.
- [ ] I pushed my PR to the latest dev commit.
- [ ] I resolved all correct Codex and CodeRabbit findings.

- [ ] My PR is ready for review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenCode Go quota detection for providers using the canonical service URL, regardless of provider name.
  * Prevented quota checks and usage reports for providers configured with other base URLs.

* **Tests**
  * Added coverage for quota detection across supported provider configurations.

PR #2160, opened 2026-08-20
fix(providers): identify opencode-free with the client User-Agent it claims

lidge-jun:
## Summary

Absorbs #2067 by @waw4303, and fixes the delivery gap that would have made it a no-op for existing installs.

**The reported defect.** `opencode-free` sends no `User-Agent`, so Zen sees the bare Bun runtime default and rate-limits it harder than a client that identifies itself. This adds `User-Agent: opencode` beside the existing `x-opencode-client: desktop` marker — the revised shape @waw4303 landed on, not the `opencode-cli/1.0.0` pin the PR originally proposed.

**Why unversioned.** A pinned CLI version is a claim about an install we do not have, and it goes stale on the vendor's schedule rather than ours. [OmniRoute](https://github.com/diegosouzapw/OmniRoute) — an independent open-source broker against the same Zen upstream — defaults to exactly this pair (`userAgent: "opencode"`, `client: "desktop"`) in `open-sse/executors/opencode.ts`, and arrived there by deliberately retreating from its own earlier `opencode-cli/1.0.0` pin. That is corroboration from a project solving the identical problem, not an authority we are bound by.

**The part that was missing.** The registry edit alone would have shipped to nobody. `staticHeaders` is documented as "merged into every upstream request for this provider" (`registry.ts`), but it was only ever copied at *seed* time: `providerConfigSeed` writes the block once, `enrichProviderFromCatalog` fills it only when the entire block is absent, and nothing merged it at request time. Verified directly against `routedProviderConfig("opencode-free", ...)`:

| persisted config | `routed.headers` before this PR |
|---|---|
| no headers block (management strips an exact registry match) | `undefined` |
| `{x-opencode-client: desktop}` | unchanged — no UA |
| `{user-agent: custom-agent}` | unchanged — no client marker |

So every existing `opencode-free` user would have kept the old fingerprint forever. `routedProviderConfig` and `buildModelsRequest` now fill registry static headers **beneath** user headers.

**Case-insensitive on purpose.** HTTP header names are case-insensitive; object keys are not. Spreading a registry `User-Agent` over a user's `user-agent` leaves both keys, which `Headers` serializes as one comma-joined value (`"custom-agent, opencode"`) — a corrupted request rather than an override. The user's spelling and value both win; the registry only fills names the user has not claimed.

**Model discovery too.** A provider identified as `opencode` when it completes but anonymous when it lists its own models reads as two different clients to a rate limiter, so `buildModelsRequest` uses the same merge.

`opencode-free` is currently the only registry row with `staticHeaders`, so the router change's blast radius today is exactly this provider. Deliberately **not** copied from OmniRoute: `x-opencode-project`, `x-opencode-request`, and `x-opencode-session`. None is needed for the reported failure, and a conversation-derived session id is a privacy-relevant change that needs its own evidence.

Closes #2067.

## Verification

- `bun test --isolate tests/opencode-free-provider.test.ts` — 18 pass / 0 fail.
- **RED-first, per component.** Reverting only `router.ts` + `oauth/index.ts` (keeping the registry header) fails exactly the 5 new delivery tests: 13 pass / 5 fail. Reverting only the registry header fails the suite at load. Neither test is vacuous.
- `bun run test` (full suite) — **13519 pass / 10 skip / 0 fail** across 856 files.
- `tests/management-provider-validation.test.ts` "provider PATCH clear keeps registry static headers" updated for the two-header registry set — 72 pass / 0 fail.

## Checklist

The header carries no user data and no credential; an operator can still override either value through the provider headers API, and that override now wins case-insensitively instead of duplicating.

## Summary by CodeRabbit

- **New Features**
  - Added registry-provided headers, including a `User-Agent`, to eligible provider requests and model discovery.
  - Preserved user-defined headers while applying registry defaults.
  - Retained saved subagent models in management responses even when disabled or unavailable.

- **Bug Fixes**
  - Corrected header handling for existing provider configurations.
  - Prevented saved subagent roster entries from disappearing from the interface.

- **Tests**
  - Expanded coverage for provider headers, case-insensitive overrides, model discovery, and subagent roster retention.

### DIFF
--- a/devlog/_plan/260820_bug_pr_backlog_consolidation/080_residual_dispositions.md
+++ b/devlog/_plan/260820_bug_pr_backlog_consolidation/080_residual_dispositions.md
@@ -161,6 +161,56 @@ Deliberately NOT copied from omniroute: `x-opencode-project`, `x-opencode-reques
 conversation-derived session identifier is a privacy-relevant change that needs its own
 evidence rather than a sibling project's precedent.
 
+### wp16 outcome — and the bug the absorb uncovered
+
+**PR #2160**, branch `codex/absorb-opencode-free-static-headers`, base `dev`. #2067 closed
+with attribution (`issuecomment-5349492578`).
+
+The audit is the interesting part. The plan as written — add the header to the registry row —
+passed my own reading and FAILED the reviewer, correctly. `staticHeaders` is documented at
+`registry.ts:149` as "merged into every upstream request for this provider", and that was
+false. It was copied at seed time only: `providerConfigSeed` writes the block once,
+`enrichProviderFromCatalog` fills it only when the whole block is absent, and nothing merged
+it at request time. `rg -n 'headers' src/router.ts` returned zero hits.
+
+Reproduced directly before accepting the finding:
+
+| persisted config | `routedProviderConfig("opencode-free", ...).headers` |
+|---|---|
+| no headers block | `undefined` |
+| `{x-opencode-client: desktop}` | unchanged — no UA |
+| `{user-agent: custom-agent}` | unchanged — no client marker |
+
+So the contributor's one-line registry patch would have shipped a header that **no existing
+install ever receives**. The management API strips a persisted block that exactly matches the
+registry set, which means the most common on-disk state is "no headers at all" — and that
+state gained nothing.
+
+Implementation, three parts:
+
+1. `mergeRegistryStaticHeaders(staticHeaders, userHeaders)` in `registry.ts` — registry values
+   fill only names the user has not claimed, compared **case-insensitively**. That last word is
+   load-bearing: HTTP header names are case-insensitive but object keys are not, so spreading a
+   registry `User-Agent` over a user's `user-agent` leaves both keys and `Headers` serializes
+   them as `"custom-agent, opencode"` — a corrupted request wearing the costume of an override.
+2. `routedProviderConfig` (`router.ts`) merges at resolve time.
+3. `buildModelsRequest` (`oauth/index.ts`) does the same, because a provider identified as
+   `opencode` when it completes but anonymous when it lists its own models reads as two
+   different clients to a rate limiter.
+
+Residual, stated rather than skipped: `validateApiKey` (`key-providers.ts:102`) still sends
+only `Authorization`. It is an auth probe by design; widening an auth-path request shape is a
+separate change with its own review burden.
+
+Evidence: 6 new regressions; reverting only `router.ts` + `oauth/index.ts` while keeping the
+registry header fails exactly 5 of them (13 pass / 5 fail), which is what makes them delivery
+tests rather than restatements of the registry constant. Full suite 13519 pass / 10 skip /
+0 fail across 856 files; typecheck and privacy scan clean.
+
+One existing expectation moved: `tests/management-provider-validation.test.ts` "provider PATCH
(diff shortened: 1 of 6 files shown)


lidge-jun:
## 리뷰 · 우선순위 56 / 80

지금 `registry.ts` `opencode-free`는 `x-opencode-client: desktop`만 있음. User-Agent 없음. Zen이 Bun 기본값으로 보고 더 세게 429. #2067을 가져오면서 `User-Agent: opencode`를 붙임. 버전 핀 아님. 기존 설치는 시드 때만 `staticHeaders`를 복사해서 레지스트리만 고치면 평생 옛날 지문임. `routedProviderConfig`/`buildModelsRequest`가 요청 때 레지스트리 헤더를 유저 헤더 아래에 채움. 대소문자 다른 키를 둘 다 넣으면 Headers가 콤마로 합침. 유저 철자가 이김.

문제: `devlog/_plan/260820_bug_pr_backlog_consolidation/`이 또 들어옴. `oauth/index.ts` `agent-settings-routes.ts` 콤보/서브에이전트 테스트도 테제가 아님. #2162랑 파일이 겹침. 플랜 덤프 빼고 헤더 커밋만. types.ts/config.ts 안 건드림. #2067은 이미 닫혔을 가능성 큼. 리베이스 지옥이면 원 PR은 닫기.

점수는 56임. 배달 구멍은 맞음. 2.27은 이미 나감. 2.28.

해결방안: 플랜/무관 파일 빼고 머지. 기존 `opencode-free` 유저 한 명으로 UA가 실제로 나가는지 확인. #2162보다 먼저.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
The `opencode-free` header direction is useful, but I am requesting changes on the current head because the branch is not a focused or current merge unit.

Two blockers remain:

1. `4890c1c4a` is 30 commits behind the current `dev` head. The green checks validate that stale mixed head, not a branch based on the current integration state. Rebase onto the latest `dev` and rerun exact-head CI.
2. The PR title and core implementation concern registry header propagation, but the 17-file diff also carries the nine `devlog/_plan/260820_bug_pr_backlog_consolidation/*` files plus the unrelated subagent-roster changes in `src/server/management/agent-settings-routes.ts`, `tests/combo-management-api.test.ts`, and `tests/subagent-roster-retention.test.ts`. Remove those unrelated commits/files. The reviewable unit should contain only the registry header contract, its inference/discovery propagation, and the directly corresponding tests.

After the branch is focused, rebased, and exact-head CI is green, this remains a reasonable merge candidate. The current CI result is not sufficient evidence for the cleaned branch because it covers materially different ancestry and scope.

lidge-jun:
---

### Stack map

| # | PR | Branch | Absorbs |
|---|---|---|---|

Ingwannu (approved):
Approved for this layer only, relative to base `codex/fix-subagent-roster-truncation`. The current layer is now focused: runtime merge of registry static headers, matching discovery behavior, and the directly corresponding tests. I verified that the only current registry `staticHeaders` row is `opencode-free`, user headers win case-insensitively, and the resolved headers reach both inference and `/models`. The exact stack-tip focused run passed, including all `opencode-free` cases.

This approval does not approve or merge the parent #2134 devlog diff. Merge/retarget bottom-up only after the parent blocker is resolved, and preserve exact-head CI after retargeting to `dev`.

PR #2162, opened 2026-08-20
fix(anthropic): frame the opening turn so AgentRouter stops blocking non-English

### LINKED ISSUE #2074: [Bug]: AgentRouter 400 content-blocked on non-English first messages; add language preamble for AGR providers

lidge-jun:
## Summary

Absorbs #2082 by @yzxcj797, with two corrections.

**The defect.** AgentRouter answers `400 content-blocked` when the first user message is not in English (#2074) while the identical English request returns `200`. The gateway inspects the opening user content, so an Anthropic `system` string never reaches the filter — the framing has to sit in that turn.

**Correction 1 — exact host, not a substring.** The original tested `hostname.includes("agentrouter")`, which also matches `notagentrouter.example` and `agentrouter.org.attacker.example`. A prompt mutation keyed on a provider's identity has to be keyed on that identity exactly, or an unrelated destination silently receives an injected instruction block. This matches `agentrouter.org` or a real subdomain of it.

**Correction 2 — do not rewrite the user's text.** The original spliced the marker into the user's own string:

```ts
firstUser.content = `${AGR_PREAMBLE_MARKER}\n\n${firstUser.content}`;
```

That edits what the user wrote. Logs, retries, and any upstream that echoes the turn then show a sentence the user never typed as if they had — the hidden user-turn mutation flagged in #1804. The framing is now its own leading text block, so the original content survives byte-for-byte. It still adds content to the user turn, which is unavoidable for a filter that reads the first user message, but it is now *additive and attributable* rather than a silent edit.

Idempotence is keyed on the leading block being exactly the marker rather than a substring test, so a user who quotes the marker later in their own prompt does not suppress their own framing.

One behavior worth naming rather than hiding: an assistant-only request is synthesized into a `(continue)` user turn upstream of this code, so the preamble lands on that synthetic turn. Pinned by a test.

No `package.json` change — the version bump in #2082 was inherited from the release line its branch was cut from, and it was the entire reason GitHub marked the PR `CONFLICTING`. The Anthropic hunks themselves merge cleanly.

Closes #2082. Fixes #2074.

## Verification

- `bun test --isolate tests/anthropic-agentrouter-language-framing.test.ts` — 10 pass / 0 fail.
- **RED-first.** Reverting only `src/adapters/anthropic.ts` fails 7 of the 10. The 3 that stay green are the lookalike-host and direct-Anthropic cases — they pass on unpatched `dev` precisely because `dev` frames nobody, which is what makes them the guard against the substring predicate rather than a restatement of it.
- `bun run test` (full suite) — **13529 pass / 10 skip / 0 fail** across 857 files.

Assertions compare the text sequence rather than whole blocks, so they stay about the framing and do not go red the next time the `cache_control` policy is tuned.

## Checklist

The host predicate is the security-relevant surface here and it was tightened, not loosened; two spoof shapes are pinned by tests.

## Summary by CodeRabbit

- **New Features**
  - Added AgentRouter-specific language framing for supported Anthropic requests.
  - Added `User-Agent: opencode` metadata to the free provider.
  - Applied provider registry headers consistently to inference and model-discovery requests.
  - Preserved saved subagent models in selections even when temporarily unavailable.

- **Bug Fixes**
  - User-defined headers now override registry headers regardless of capitalization.
  - Prevented duplicate AgentRouter framing while preserving message content and ordering.

- **Tests**
  - Expanded coverage for header behavior, language framing, model retention, and configuration persistence.

### DIFF
--- a/devlog/_plan/260820_bug_pr_backlog_consolidation/080_residual_dispositions.md
+++ b/devlog/_plan/260820_bug_pr_backlog_consolidation/080_residual_dispositions.md
@@ -214,3 +214,61 @@ clear keeps registry static headers" now asserts the two-header set. That is the
 ## wp17-wp19 — the three that need new code
 
 Recorded here as each is decided; each is its own PABCD cycle.
+
+### wp17 — #2082 @yzxcj797: AgentRouter language framing
+
+**PR #2162**, branch `codex/absorb-agentrouter-language-framing`, base `dev`. #2082 closed
+(`issuecomment-5349709140`). Fixes #2074.
+
+The diagnosis was the contributor's and it was correct: AgentRouter answers 400
+`content-blocked` on a non-English first user message while the same request in English
+returns 200, and the filter reads that turn, so an Anthropic `system` string cannot reach it.
+
+Two corrections.
+
+**Host predicate.** `hostname.includes("agentrouter")` also matches `notagentrouter.example`
+and `agentrouter.org.attacker.example`. This is a prompt mutation keyed on a provider's
+identity, so the key has to be that identity exactly — otherwise an unrelated destination
+quietly receives an injected instruction block. Now `agentrouter.org` or a real subdomain.
+
+**Where the marker goes.** The original spliced it into the user's string:
+`firstUser.content = \`\${MARKER}\\n\\n\${firstUser.content}\``. That edits what the user
+wrote, and every downstream reader then attributes a sentence to them that they never typed —
+the hidden user-turn mutation named in #1804. The framing is now its own leading text block.
+It still adds content to the user turn, which is unavoidable against a filter that reads the
+first user message, but additive-and-visible is a different risk class than a silent rewrite.
+
+Idempotence is keyed on the LEADING block being exactly the marker, not a substring test: a
+user who quotes the marker mid-prompt must not suppress their own framing.
+
+Evidence: 10 regressions; reverting only the adapter fails 7. The 3 that stay green are the
+lookalike-host and direct-Anthropic cases — green on unpatched `dev` precisely because `dev`
+frames nobody, which is what makes them guards against the substring predicate rather than
+restatements of it. Full suite 13529 pass / 10 skip / 0 fail; typecheck and privacy clean.
+
+`CONFLICTING` was an inherited `package.json` bump alone; the Anthropic hunks merge cleanly.
+No version change in the replacement.
+
+### wp18 — #2027 @yzxcj797: OpenCode Go quota, planned
+
+The investigation moved the answer here too. The real issue is #1924: sibling rows
+(`opencode-go-2` … `-5`) show no quota in the dashboard and no rows in
+`ocx provider quota --refresh --json`, because dispatch gates on the literal provider NAME at
+`src/providers/quota.ts:2087`.
+
+The contributor's fix swaps that for a base-URL comparison. Closer, but it does not check the
+adapter, so a row pointed at the canonical URL with a different adapter would be probed.
+
+The repository already has the exact predicate: `registryEntryForProviderDestination`
+(`registry.ts:2678`) identifies a renamed fixed key provider by normalized endpoint + adapter
++ auth mode, and is already the convention for renamed rows
+(`opencode-zen-rate-limit.ts:28-43`, `derive.ts:398-425`).
+
+Rejected alternative, recorded: `providerMatchesRegistryTransport("opencode-go", provider)`
+would need `preserveCustomDestination: true` on the registry entry, which also changes ROUTING
+for a same-named custom row (`router.ts:269-274` vs `:320-336`). That may be worth doing, but
+not as a side effect of a quota fix.
+
+The defensive canonical-URL check inside `fetchOpenCodeGoQuota` (`quota.ts:485-494`) stays: it
+is what stops an API key being sent to a non-canonical host, and it should not depend on the
+dispatch predicate being correct.
--- a/src/adapters/anthropic.ts
+++ b/src/adapters/anthropic.ts
@@ -594,6 +594,63 @@ function orphanToolResultText(msg: OcxToolResultMessage): string {
   return `[tool_result without adjacent tool_use: ${label}]\n${content}`;
 }
 
+/**
+ * AgentRouter answers 400 `content-blocked` when the first user message is not in English
+ * (#2074), while the same request in English returns 200. The gateway is inspecting the opening
+ * user content, so an Anthropic `system` string cannot reach it — the framing has to sit in the
+ * first user turn.
+ */
+const AGENTROUTER_LANGUAGE_PREAMBLE =
+  "[Instruction: Process the user request below and respond in the appropriate language.]";
+
+/**
+ * Exact host match, not a substring.
+ *
+ * A `hostname.includes("agentrouter")` test also matches `notagentrouter.example` and
+ * `agentrouter.org.attacker.example`, which would let an unrelated destination silently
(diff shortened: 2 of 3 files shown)


lidge-jun:
## 리뷰 · 우선순위 58 / 80

#2074 실화임. AgentRouter가 첫 유저 메시지가 영어 아니면 400 content-blocked. Anthropic `system`은 필터에 안 들어감. 오프닝 유저 턴에 프레이밍이 있어야 함. #2082를 가져오면서 구멍 둘을 막음.

호스트는 `hostname.includes("agentrouter")` 금지. `agentrouter.org`랑 그 서브도메인만. 유저 문자열에 마커를 이어붙이지 말 것 (#1804). 앞 텍스트 블록으로 따로. 원문은 바이트 그대로. 멱등은 선두 블록이 마커와 완전 일치. 인용하면 자기 프레이밍이 안 사라짐. assistant-only는 위의 `(continue)` 합성 턴에 붙음. 테스트 있음. #2082의 package.json 범프는 충돌 원인이었고 이번엔 없음.

같은 플랜 덤프랑 `oauth/index.ts` `registry.ts` `router.ts`가 #2160이랑 겹침. 스택이면 #2160 먼저. 아니면 파일 빼기. types.ts/config.ts 안 건드림. 분할에 깔림 없음. #2082 리베이스하지 말 것. #2074는 머지되면 닫으면 됨.

점수는 58임. 맞는데 2.28. 플랜 덤프 빼면 더 보기 좋음.

해결방안: #2160 머지 후 이 PR. 플랜 파일 제거. 비영어 첫 턴 400 재현 유지.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
The AgentRouter framing fix is directionally valuable, and the exact-host plus additive-block corrections are better than the original proposal. I am requesting changes because the current branch is not a focused or current merge unit.

Two blockers remain:

1. `53a5f73d4` is 30 commits behind the current `dev` head. Rebase onto the latest `dev` and rerun exact-head CI.
2. This PR should contain the AgentRouter change in `src/adapters/anthropic.ts` and its focused regression test. Instead, the 19-file diff also carries the complete backlog-planning set, the `opencode-free` header work, and the unrelated subagent-roster work. Split those inherited commits out so review and rollback boundaries match the stated Anthropic defect.

Once the branch contains only the framing fix, is based on current `dev`, and its focused/full checks pass at that exact head, it should remain a strong merge candidate for #2074.

lidge-jun:
---

### Stack map

| # | PR | Branch | Absorbs |
|---|---|---|---|

Ingwannu (changes requested):
The stacked-base change fixes the previous ancestry/scope problem, and the exact-host plus separate-block implementation is directionally correct. One current prompt-boundary blocker remains.

The unresolved test finding on `tests/anthropic-agentrouter-language-framing.test.ts` is valid. Because this code injects content into the outbound first user turn, add regressions for: (1) an already-present exact leading preamble block staying single/idempotent, (2) image-only structured user content receiving the preamble without losing or reordering the image block, and (3) assistant-only input represented with content-part arrays, asserting that the assistant tail is preserved and followed by the synthesized user turn `[PREAMBLE, "(continue)"]`.

The current focused file passes 10/10, but those three branches are the safety boundary for this provider-specific prompt mutation. Push the tests, rerun exact-head CI, and re-request review. The parent #2134 blocker must also be resolved before this stack can land.

PR #2164, opened 2026-08-20
fix(quota): identify OpenCode Go by where it routes, not what it is named

### LINKED ISSUE #1924: Quota probe for OpenCode Go is gated on the literal provider name, so additional Go providers (multi-account setups) show no quota
### Client or integration

OpenCodex dashboard

### Area

Provider adapter

### Summary

When multiple providers point at the canonical OpenCode Go base URL (`https://opencode.ai/zen/go/v1`) — the natural setup for balancing several Go subscriptions via combos — only the provider literally named `opencode-go` shows the rate-limit panel (5h / weekly / 30d). The sibling providers show no quota in the dashboard, and `ocx provider quota --refresh --json` returns no report rows for them. Expected: every key-auth provider on the canonical Go base URL reports quota.

**Root cause**: [`src/providers/quota.ts` L2087](https://github.com/lidge-jun/opencodex/blob/main/src/providers/quota.ts#L2087) gates the probe on the provider *name*:

```ts
if ((provider.authMode ?? "key") === "key" && name === "opencode-go") {
  return fetchOpenCodeGoQuota(name, provider);
}
```

The Kimi branch just above ([L2075](https://github.com/lidge-jun/opencodex/blob/main/src/providers/quota.ts#L2075)) already uses the canonical-base-URL pattern instead, and `fetchOpenCodeGoQuota` itself re-validates `isCanonicalOpenCodeGoBaseUrl(config.baseUrl)` ([L487](https://github.com/lidge-jun/opencodex/blob/main/src/providers/quota.ts#L487)), so the name check is both too narrow and redundant.

**Proposed fix** (verified locally on v2.14.0 — report rows went from 1 to 5, dashboard renders rate-limit bars for every sibling provider):

```diff
-    if ((provider.authMode ?? "key") === "key" && name === "opencode-go") {
+    if ((provider.authMode ?? "key") === "key" && isCanonicalOpenCodeGoBaseUrl(provider.baseUrl)) {
       return fetchOpenCodeGoQuota(name, provider);
     }
```

Safe against key leakage to non-canonical hosts because `fetchOpenCodeGoQuota` still checks the canonical URL before sending the bearer key.

### Reproduction

1. `ocx provider add opencode-go --api-key sk-AAA` (registry preset)
2. `ocx provider add opencode-go-2 --adapter openai-chat --base-url https://opencode.ai/zen/go/v1 --api-key sk-BBB`
3. `ocx provider quota --refresh --json`
4. Only one `"provider": "opencode-go"` report is returned; the dashboard overview page for `opencode-go-2` has no rate-limit section.

### Version

v2.14.0 (repro); the gate is unchanged on current `main` as of 2026-08-17 (L2087)

### Operating system

macOS (Darwin 25.5)

### Provider and model

opencode-go (any model; the quota probe is model-independent)

### Logs or error output

`ocx provider quota --refresh --json` before the fix returns a single `opencode-go` report row; after the one-line change it returns all five (`opencode-go`, `opencode-go-2` … `opencode-go-5`), each with `fiveHourPercent` / `weeklyPercent` / `monthlyPercent` populated from `https://opencode.ai/zen/go/v1/usage`.

_No response_

### Redacted configuration

Providers: `opencode-go` (registry preset) plus `opencode-go-2..5`, all `adapter: openai-chat`, `baseUrl: https://opencode.ai/zen/go/v1`, `authMode: key`, one API key each. Keys redacted.

### Checks

lidge-jun:
## Summary

Absorbs #2027 by @yzxcj797. Stack layer 4 — base is `codex/absorb-agentrouter-language-framing` (#2162), not `dev`.

**The defect (#1924).** A multi-account setup points several provider rows at the same OpenCode Go endpoint under names the registry has never heard of — `opencode-go-2` through `-5`. Quota dispatch gated on the literal name `"opencode-go"` (`src/providers/quota.ts`), so those rows had no dashboard quota panel and no report in `ocx provider quota --refresh --json`, even though each holds a working key for the same upstream.

Worth naming: identity was already answered *inconsistently* here. Dispatch asked "what is this row called"; the credential-safety check inside `fetchOpenCodeGoQuota` asked "where does it actually route". The second question is the right one.

**The correction on top of #2027.** The original swaps the name for a base-URL comparison. That fixes the reported symptom but does not check the adapter, so a row pointing at the canonical URL through a *different* adapter would be probed — a different protocol to that host, and not the provider whose quota shape we parse.

This uses `registryEntryForProviderDestination`, which this repository already uses for renamed fixed-key rows (`opencode-zen-rate-limit.ts`, `derive.ts`): normalized endpoint + adapter + key auth.

**Rejected alternative, recorded.** `providerMatchesRegistryTransport("opencode-go", provider)` would need `preserveCustomDestination: true` on the registry entry, which also changes *routing* for a same-named custom row. That may be worth doing on its own merits; it is not something a quota fix should do as a side effect.

The defensive canonical-URL check inside `fetchOpenCodeGoQuota` stays. Whether an API key may be sent to a host must not depend on the dispatch gate above it being correct.

No `package.json` change — the version bump in #2027 was inherited from its branch point and was the entire reason GitHub marked it `CONFLICTING`.

Closes #2027. Fixes #1924.

## Verification

- `bun test --isolate tests/opencode-go-quota.test.ts` — 4 pass / 0 fail.
- **RED-first.** Reverting only `src/providers/quota.ts` fails exactly the sibling-row test (3 pass / 1 fail). The wrong-adapter test stays green on unpatched `dev` because `dev` probes nothing but the literal name — which is what makes it a guard against the URL-only predicate rather than a restatement of the fix.
- `bun run test` at the stack tip — **13562 pass / 10 skip / 0 fail** across 857 files.

## Checklist

The test asserts the sibling's key never appears in the serialized report.

---

**Stack** (bottom to top): #2134 → #2160 → #2162 → **this** → openai-chat padding repeats

### DIFF
--- a/src/providers/quota.ts
+++ b/src/providers/quota.ts
@@ -13,7 +13,7 @@ import { getAccountCredential, getAccountSet, getCredential } from "../oauth/sto
 import { antigravityUserAgent } from "../adapters/client-fingerprint";
 import { apiKeyPoolEntryId } from "./api-keys";
 import { XAI_GROK_CLIENT_VERSION, XAI_GROK_COMPATIBILITY } from "./xai-transport";
-import { getProviderRegistryEntry, providerCodexAccountMode } from "./registry";
+import { getProviderRegistryEntry, providerCodexAccountMode, registryEntryForProviderDestination } from "./registry";
 import type { OcxConfig, OcxProviderConfig } from "../types";
 import { isCanonicalOpenAiForwardProvider, OPENAI_CODEX_PROVIDER_ID } from "./openai-tiers";
 import {
@@ -2084,7 +2084,14 @@ async function maybeFetchProviderQuota(
       && isCanonicalCommandCodeBaseUrl(provider.baseUrl)) {
       return fetchCommandCodeQuota(name, provider);
     }
-    if ((provider.authMode ?? "key") === "key" && name === "opencode-go") {
+    // Identify OpenCode Go by where it routes, not by what the row is called. Multi-account
+    // setups keep the same destination under names like `opencode-go-2` (#1924), and those rows
+    // silently had no quota panel and no `ocx provider quota --json` report while the literal
+    // name was the gate. `registryEntryForProviderDestination` is the existing predicate for
+    // exactly this question: normalized endpoint + adapter + key auth, so a canonical URL behind
+    // a different adapter is still not OpenCode Go. The defensive URL check inside
+    // `fetchOpenCodeGoQuota` stays — sending a key anywhere must not depend on this gate.
+    if ((provider.authMode ?? "key") === "key" && registryEntryForProviderDestination(provider)?.id === "opencode-go") {
       return fetchOpenCodeGoQuota(name, provider);
     }
     if ((provider.authMode ?? "key") === "key" && isCanonicalA6apiBaseUrl(provider.baseUrl)) {
--- a/tests/opencode-go-quota.test.ts
+++ b/tests/opencode-go-quota.test.ts
@@ -87,4 +87,61 @@ describe("OpenCode Go provider quota", () => {
     expect(fetchCalls).toBe(0);
     expect(result.reports).toEqual([]);
   });
+
+  // #1924: a multi-account setup points several rows at the same OpenCode Go endpoint under
+  // names the registry has never heard of. Gating dispatch on the literal name `opencode-go`
+  // meant those rows had no dashboard quota panel and no `ocx provider quota --json` report,
+  // even though each one holds a working key for the same upstream.
+  test("a canonical sibling row under any name is probed and reported", async () => {
+    const bearers: string[] = [];
+    globalThis.fetch = (async (_input: RequestInfo | URL, init?: RequestInit) => {
+      const headers = init?.headers as Record<string, string> | undefined;
+      if (headers?.Authorization) bearers.push(headers.Authorization);
+      return new Response(JSON.stringify({
+        usage: { rolling: { status: "ok", percent: 12, resetsAt: "2026-08-12T20:00:00.000Z" } },
+      }), { status: 200, headers: { "content-type": "application/json" } });
+    }) as typeof fetch;
+
+    const config = openCodeGoConfig();
+    config.providers["opencode-go-2"] = {
+      adapter: "openai-chat",
+      authMode: "key",
+      baseUrl: "https://opencode.ai/zen/go/v1",
+      // Kept under the privacy scanner's bearer-token length floor: a longer fixture reads as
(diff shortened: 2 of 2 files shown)


lidge-jun:
---

### Stack map

| # | PR | Branch | Absorbs |
|---|---|---|---|

Ingwannu (approved):
Approved for this layer only, relative to `codex/absorb-agentrouter-language-framing`. The destination predicate is stricter than the original URL-only proposal: normalized endpoint, matching adapter, and key auth must all resolve to the canonical `opencode-go` registry row, while `fetchOpenCodeGoQuota` retains its own defensive destination check before sending the bearer credential. The renamed-sibling and wrong-adapter regressions both pass; the exact current layer CI is green.

Merge/retarget bottom-up only after the lower stack blockers are resolved, then preserve exact-head CI on the `dev` comparison.

lidge-jun:
## 리뷰 · 우선순위 62 / 80

#1924 실화임. 지금 `quota.ts` 2087줄이 `name === "opencode-go"`만 봄. `opencode-go-2`부터 `-5`처럼 같은 Zen Go 엔드포인트에 다른 이름이면 대시보드 쿼타 패널이 없고 `ocx provider quota --refresh --json`에도 안 나옴. 키는 됨. 디스패치는 이름을 묻고 `fetchOpenCodeGoQuota` 안쪽은 목적지를 물음. 뒤가 맞음.

#2027을 가져옴. 원본은 baseUrl만 바꿈. 어댑터를 안 봐서 같은 URL에 다른 어댑터면 쿼타 셰이프가 다른데도 프로브함. 이번엔 `registryEntryForProviderDestination` — 정규화 엔드포인트 + 어댑터 + key auth. zen-rate-limit이랑 derive가 이미 쓰는 거임. `providerMatchesRegistryTransport`는 `preserveCustomDestination`이 라우팅까지 바꿔서 거절. 맞음. 안의 canonical URL 가드는 남김. 디스패치가 틀려도 키를 이상한 호스트로 안 보냄.

스택 레이어 4. 베이스가 #2162지 `dev`가 아님. 파일은 `quota.ts`랑 테스트뿐이라 테제는 깨끗함. types.ts/config.ts 안 건드림. #2027 package.json 범프는 뺐음. 원 PR 리베이스하지 말 것. #1924는 머지되면 닫으면 됨.

점수는 62임. 2.27은 이미 나감. 2.28. #2162 먼저.

해결방안: #2162 머지 후 이 PR. 시블링 이름 테스트 유지. 잘못된 어댑터 가드도 유지.

이 댓글은 grok-bot이 작성했습니다

Issue #2156, opened 2026-08-19
opencode-go/muse-spark-1.2: frequent 'upstream stream ended mid tool call without a terminal signal — possible truncation' in streaming tool calls (Codex App via ocx, Pi direct OK)


fede-oss:
### Client or integration

Codex App (also reproduced via Pi coding-agent `0.83.0` when routed through OpenCodex proxy)

### Area

Provider adapter

### Summary

`muse-spark-1.2` via `opencode-go` (`adapter: openai-chat`, `baseUrl: https://opencode.ai/zen/go/v1`) fails frequently in streaming tool-call turns with:

```
upstream stream ended mid tool call without a terminal signal — possible truncation
```

Source is `src/adapters/openai-chat.ts:1718-1726`:

```ts
const sawFinish = finishReason !== undefined;
if (!sawFinish && pendingToolCalls.length > 0) {
  yield { type: "error", message: "upstream stream ended mid tool call without a terminal signal — possible truncation" };
}
```

The upstream SSE emits `response.created` / tool deltas but closes without `finish_reason`, `[DONE]`, or `response.completed`. The model is reasoning-heavy and the turn is truncated mid-JSON, leaving pending tool calls un-finalized. Same model via Pi **direct** to `opencode-go` (no `ocx`) is fine — this isolates the failure to the `ocx` streaming proxy layer, matching `anomalyco/opencode#40888` matrix (`stream:true + tools` fails, `stream:false + tools` works).

Expected: streaming tool request should produce a complete terminal sequence or a structured retryable error, or automatically fallback to buffered mode like the `opencode#40888` workaround (`stream:false` + reframe as SSE).

### Reproduction

1. `ocx start --port 10100` (default, opencodex 2.26.0)
2. Pi/Codex config:
   ```json
   {
     "defaultProvider": "opencode-go",
     "defaultModel": "muse-spark-1.2-contributor",
     "defaultThinkingLevel": "xhigh"
   }
   ```
3. Run a session that triggers a tool-heavy turn (e.g., `read`/`write`/`edit` batch, ~14k thinking tokens). No manual abort.
4. Observe request log: turn emits thinking then 1-3 tool calls, last tool args truncated mid-token (`cd ~/code/zosmaai/z`-like), then proxy yields:
   ```
   [openai-chat] stream-truncated { finishReason: null, hadUsage: false, pendingToolCalls: 2 }
   error: upstream stream ended mid tool call without a terminal signal — possible truncation
   ```
[...]

PR #2165, opened 2026-08-20
fix(openai-chat): treat a non-string repeat as padding once that field is known

lidge-jun:
## Summary

Absorbs #2155 by @waw4303. Stack layer 5 — base is `codex/absorb-opencode-go-quota-siblings` (#2164), not `dev`.

**The defect.** Some OpenAI-compatible streamers repeat an already-sent `id`, `name`, or `arguments` as a non-string placeholder on a continuation delta rather than as `null`. Validation ran *before* the pending-call lookup, so the adapter could not see that the same wire index already held that value in canonical form. The stream died with a 502 and the tool never ran.

Reproduced on unpatched `dev`: `upstream response contained invalid tool calls (tool_call_function_name_invalid; callIndex=0; valueType=object)`, with the following valid `{"command":"ls"}` fragment never consumed.

**Two corrections on top of #2155.**

*Too wide for `arguments`.* The original gated acceptance on the call having a canonical **name**. A name says nothing about whether `arguments` was ever sent as a string — so a chunk establishing `name: "terminal"` with no `arguments` field, followed by `arguments: {command:"ls"}`, would silently drop a real payload the model intended to send. `PendingToolCall` now carries `sawArgumentsString`. An empty string counts: it proves the upstream sent the field with the right wire type, which is exactly what a later malformed repeat would be padding for.

*Too narrow for `id`.* A non-string repeated `id` stayed unconditionally terminal even after a canonical id was stored. Ids now follow the same per-field rule.

**Diagnostics.** These are now passed from the rejection site rather than rescanned. A stateless rescan stops at the first structurally odd value it sees — so a stream carrying *accepted* padding on call 0 and a real defect on call 1 blamed call 0. Both `invalidToolCallsEvent` and `logInvalidToolCalls` take an optional override; structural failures before pending state is relevant still use the stateless fallback.

Sibling-vs-stacked note: #2155 has no dependency on the openai-chat hunks in #2151 (imports/service-tier around lines 14–1343 vs stream parsing around 1628–1677). It sits here for a clean single chain, not because the code requires it.

Closes #2155.

## Verification

- `bun test --isolate tests/openai-chat-hardening.test.ts tests/openai-chat-invalid-tool-call-diagnostics.test.ts` — 62 pass / 0 fail (58 pre-existing, unchanged by the refactor).
- **RED-first, and the split is the point.** Reverting only `src/adapters/openai-chat.ts` gives 2 pass / 2 fail. The two that fail are the bug being fixed. The two that pass — canonical-name-does-not-authorize-arguments, and malformed-id-stays-terminal — pass on unpatched `dev` because `dev` rejects everything; they are the guards proving this PR did not widen tolerance while fixing it.
- `bun run test` at the stack tip — **13562 pass / 10 skip / 0 fail** across 857 files.

## Checklist

Diagnostics still report shape only, never values — the existing privacy contract for this channel is unchanged.

---

**Stack** (bottom to top): #2134 → #2160 → #2162 → #2164 → **this**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when processing streamed tool calls with repeated or malformed fields.
  * Preserved valid tool-call data when non-critical padding is malformed.
  * Reports more precise diagnostics, including the specific failing call and field.
  * Prevents unresolved invalid tool-call values from being accepted.

* **Tests**
  * Added coverage for parallel tool calls, malformed IDs, repeated fields, and diagnostic attribution.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -308,8 +308,13 @@ function invalidToolCallsEvent(
   rawToolCalls: unknown,
   mode: "stream" | "response",
   usage?: OcxUsage,
+  diagnosticOverride?: InvalidToolCallDiagnostic,
 ): Extract<AdapterEvent, { type: "error" }> {
-  const diagnostic = diagnoseInvalidToolCalls(rawToolCalls, mode);
+  // The streamed accumulator knows things a rescan cannot: which field on which pending call
+  // was actually rejected. Without the override, a stream carrying accepted padding on call 0
+  // and a real defect on call 1 blames call 0, because the stateless scan stops at the first
+  // structurally odd value it sees.
+  const diagnostic = diagnosticOverride ?? diagnoseInvalidToolCalls(rawToolCalls, mode);
   const detail = diagnostic
     ? ` (${diagnostic.reason}${diagnostic.callIndex !== undefined ? `; callIndex=${diagnostic.callIndex}` : ""}; valueType=${diagnostic.valueType})`
     : "";
@@ -527,9 +532,13 @@ function diagnoseInvalidToolCalls(
   return undefined;
 }
 
-function logInvalidToolCalls(mode: "stream" | "response", rawToolCalls: unknown): void {
+function logInvalidToolCalls(
+  mode: "stream" | "response",
+  rawToolCalls: unknown,
+  diagnosticOverride?: InvalidToolCallDiagnostic,
+): void {
   if (!isDebugEnabled()) return;
-  const diagnostic = diagnoseInvalidToolCalls(rawToolCalls, mode);
+  const diagnostic = diagnosticOverride ?? diagnoseInvalidToolCalls(rawToolCalls, mode);
   if (!diagnostic) return;
   const fieldShape = fingerprintInvalidField(invalidToolCallField(rawToolCalls, diagnostic));
   debugProviderDiagnostic("openai-chat", "invalid-tool-calls", {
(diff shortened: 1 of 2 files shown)


lidge-jun:
---

### Stack map

| # | PR | Branch | Absorbs |
|---|---|---|---|

Ingwannu:
Manual review of the current layer found the per-field provenance correction materially safer than #2155: `name`, `arguments`, and `id` no longer authorize one another, unresolved fields still fail closed, and diagnostics identify the actual rejected call without logging values. The exact stack-tip focused run passed 97/97 tests, including the hardening and diagnostic suites.

I am not approving yet because CodeRabbit skipped this non-`dev` base and macOS CI is still running. @coderabbitai review

After automated review is complete, the lower-stack blockers are resolved, and the child is retargeted with exact-head CI, this remains a strong merge candidate.

Ingwannu (approved):
Approved for this stacked layer at exact head `772d375f`. The per-field provenance model keeps `name`, `arguments`, and `id` independent, accepts non-string repeats only after that exact field has a canonical value, and preserves call-specific fail-closed diagnostics without logging attacker-controlled values. Focused validation passed 97/97 at the stack tip; exact-head CI is now green on all reported platforms, and the requested CodeRabbit review produced no actionable findings.

This approval is for the #2165 layer only. Do not merge it out of order: resolve the lower-stack blockers, retarget this child to `dev` after its parent lands, and require fresh exact-head CI on the retargeted PR.

lidge-jun:
## 리뷰 · 우선순위 56 / 80

#2155를 가져온 거임. 지금 스트림 루프가 pending 조회 전에 `isInvalidStreamStringField`를 무조건 봄. zen/go가 continuation에 `function.name`을 객체로 반복하면 앞 청크 이름 받아놓고도 502. #1731은 null만 패딩이었어서 객체는 죽음.

#2155 구멍 둘이 여기서 막힘. 원본은 args 검사를 `hasCanonicalName`에 묶음. 이름만 있고 args 없는 다음 청크가 객체 args면 침묵하고 페이로드가 날아감. 이번엔 `PendingToolCall.sawArgumentsString`. 빈 문자열도 본 거임. 나중에 이상한 반복은 패딩. id도 같은 필드 단위. 진단은 거절 지점에서 넘김. 상태 없는 리스캔이 콜0 패딩을 콜1 결함으로 덮어씀.

#2156이랑 다른 구멍임. 이건 패딩 502. 저건 finish_reason 없이 툴콜 잘림. 이 PR로 #2156 닫지 말 것. 스택 레이어 5. 베이스 #2164. #2151이랑 openai-chat 구간이 달라서 코드 의존은 아님. types.ts/config.ts 안 건드림. #2155는 머지되면 닫고 드래프트 리베이스하지 말 것.

점수는 56임. 맞음. 2.28. #2164 뒤.

해결방안: #2164 다음. `sawArgumentsString` 테스트 유지. 진단 override 유지.

이 댓글은 grok-bot이 작성했습니다

Issue #2157, opened 2026-08-19
[Bug]: Codex App now sends gpt-5.6-luna helper requests on every message and turn completion (previously only title generation)


fede-oss:
### Client or integration

Codex App

### Area

Proxy and routing

### Summary

Since 2026-08-19 ~22:15 UTC, my usage.jsonl fills with one-shot `gpt-5.6-luna` requests arriving roughly every time I send a message in any Codex App thread and again every time an agent turn completes (~50/hour during active use). Before that date, for weeks, the only background luna traffic I saw was the single title-generation call per newly created thread — which I understand and expect.

I'm opening this to ask: **why this sudden change, and is there an OpenCodex-side way to identify/label or intercept these calls?**

What I've been able to determine so far:

- The calls are Codex App-originated: `admissionKind: loopback`, `inboundProtocol: responses`, `requestedEffort: low`, ~16–18k input tokens (base instructions + message excerpt), tiny outputs (20–90 tokens), status 200, native route to OpenAI.
- Each call has a **unique conversationId** that matches no session rollout file; the app appears to run them in ephemeral `threadSource: system` threads (fresh session id per call).
- Timing correlates 1:1 with thread events: one call within ~2–4s of each user message sent, one near each turn completion, across all concurrent threads (verified against rollout `user_message`/`task_complete` timestamps for 7 threads; 32 messages + 28 turn ends ≈ 50 luna calls/hour).
- Inspecting the Codex App binary (26.814.41407, runtime 0.148.0-alpha.15): these match the `thread_summary` feature ("one-line activity update displayed beneath an existing Codex task title", phases `user` and `assistant`), hardcoded to `gpt-5.6-luna` with effort `low`, plus related `thread_title`/`thread_description`/`ambient_suggestion*` helpers. Generation and rendering appear gated by an app-side remote feature flag.
- Upstream reports of the same class of hidden background calls: openai/codex#37751 (analytics attributes luna turns the user never started), openai/codex#37009 (hardcoded luna/terra memory-writer model ids), openai/codex#29380 (ambient suggestions consuming quota, no off switch).
- My `shadowCallIntercept` is currently **disabled**; I see it historically targeted exactly this helper family (see #1684, #311, #1402, #668).

Questions:

1. Is this per-message/per-turn luna helper traffic a known recent Codex App behaviour change (0.148.0-alpha.x), and is it expected to persist?
[...]

PR #2166, opened 2026-08-20
fix(logs): sanitize the shadow helper marker at the logging layer

lidge-jun:
## Summary

Absorbs #2163 by @Ingwannu. Stack layer 6 — base is `codex/absorb-openai-chat-padding-repeats` (#2165), not `dev`.

**What it fixes.** When the opt-in shadow-call route intercepts a Codex App helper request and rewrites it to an operator-selected model, the original helper model was lost. `shadowCallRewrittenFrom` was written through an untyped cast at `core.ts`, but `RequestLogContext`/`RequestLogEntry` never declared it and `addFinalRequestLog` rebuilds its row field by field — so the marker reached neither `/api/logs` nor `usage.jsonl`, and could not survive restart hydration. Recurring helper spend showed up as an unattributable row (#2157).

@Ingwannu's typing, persistence projection, and hydration work is carried unchanged in principle. Both of the author's commits are preserved.

**The correction.** The original sanitizes at the one call site that populates the field today, which leaves the in-memory `/api/logs` row carrying whatever the caller sent. The value originates in an upstream-supplied model id, so an unsanitized newline lets a single field forge a record boundary in any line-oriented log viewer, and nothing bounded its length on that path. `addFinalRequestLog` now runs it through `sanitizeLogMetadataString` itself — a future caller cannot reintroduce the hole by forgetting to sanitize first, and the in-memory row matches what `usage.jsonl` already stored.

**Scope note.** This is the backend attribution half. The dashboard does not yet render or filter on the field (`gui/src/pages/Logs.tsx`), so #2157 stays open until that lands — closing it on this PR would claim a UI affordance that does not exist.

Closes #2163.

## Verification

- `bun test --isolate tests/request-log.test.ts -t 'shadow'` — 2 pass / 0 fail.
- **RED-first.** Reverting the implementation leaves the author's persistence test green and fails the new sanitizer test (1 pass / 1 fail). That split is the evidence: the original fixture is a safe short slug, so it passes identically whether the sanitizer runs or not. The new one writes an unsafe overlong marker and asserts the newline is gone and the 64-character bound holds on both paths.
- `bun run test` at this tip — **13564 pass / 10 skip / 0 fail** across 857 files.

## Checklist

Only a model-id scalar is stored, through the shared control-stripping, secret-redacting, 64-character sanitizer. No request content is logged.

---

### Stack map

| # | PR | Branch | Absorbs |
|---|---|---|---|
| 6 | **this** | `codex/absorb-shadow-helper-attribution` | #2163 @Ingwannu |

No file overlap with any layer below — this sits at the top for a single clean chain, not because the code requires it.

## Summary by CodeRabbit

- **New Features**
  - Shadow-call rewrites now record the original helper model in usage and request logs.
  - This metadata persists across restarts without retaining request content.

- **Bug Fixes**
  - Log metadata is sanitized consistently before being stored, preventing newline-based log injection.
  - Rewritten model identifiers are bounded and safely redacted.

- **Documentation**
  - Documented opt-in shadow-call rewrite tracking and its privacy safeguards.

### DIFF
--- a/devlog/_plan/260820_bug_pr_backlog_consolidation/000_research_inventory.md
+++ b/devlog/_plan/260820_bug_pr_backlog_consolidation/000_research_inventory.md
@@ -210,3 +210,26 @@ supersede stands.
 Ordered by score, not by dependency, because no dependency exists. Each is independently
 reviewable and independently mergeable, which is what DEV-STACK-01 actually asks for.
 
+
+## Correction 4 — #2130 merged mid-cycle; #2029 is now superseded in fact, not in prospect
+
+`gh pr view 2130`: `MERGED` at 2026-08-19T17:25:26Z. The probe lane's verdict was conditional
+("SUPERSEDED, once #2130 merges"); that condition is now satisfied.
+
+Consequence for `060`: **#2029 (@yzxcj797)** moves from a prospective close to an immediate one.
+`dev` now carries `busUnreachable()` — a superset of #2029's two stderr strings — plus the
+on-disk unit check that #2029's own reviewer demanded. Landing #2029 on top would REGRESS that
+disk check back to an unconditional `absent`. Nothing from #2029 needs to be carried over; its
+one unique behavior (keeping `DBUS_SESSION_BUS_ADDRESS not set` as `unknown`) is precisely what
+the merged disk check replaces.
+
+This also removes #2130 from the open-bug-PR set: the fresh count at wp1 close is 26 open bug
+PRs, of which exactly one (#2134) is lidge-jun's and 25 are not.
+
+## wp1 close-out evidence
+
+`gh pr list --repo lidge-jun/opencodex --state open --label bug --limit 100` at close:
+26 total, mine = [2134], non-mine = 25. Every one of those 25 numbers appears in §1/§3 of this
+document (verified by a grep loop over the list, exit 0). No open bug PR is left without a
+disposition.
+
--- a/devlog/_plan/260820_bug_pr_backlog_consolidation/000_research_inventory.md
+++ b/devlog/_plan/260820_bug_pr_backlog_consolidation/000_research_inventory.md
@@ -1,0 +1,133 @@
+# 000 — Research: open bug-PR backlog inventory, rubric, and disposition
+
+Unit: 260820_bug_pr_backlog_consolidation
+Work-phase: wp1 (docs-only roadmap cycle, LOOP-DOCS-FIRST-01)
+Baseline: origin/dev = ceac592d7. Worktree branch codex/fix-subagent-roster-truncation (PR #2134).
+
+Evidence for every claim below came from six read-only xai/grok-4.6 investigation lanes that
+read the actual PR diffs with `gh pr diff` and cross-read the runtime in this worktree. Code
+edits stay in the main agent.
+
+## 1. Inventory
+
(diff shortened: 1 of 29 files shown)


Ingwannu (changes requested):
The logging-layer direction is correct, but one public ingress still bypasses the new boundary on the current head (`70b947550`).

`addRequestLog(entry)` calls `retainRequestLogEntry(entry)` before rebuilding the persisted row. A direct caller can therefore place the raw `shadowCallRewrittenFrom` value in the in-memory `/api/logs` ring, while `appendUsageEntry()` sanitizes only the disk copy. I reproduced this at the exact head: control/secret-shaped content remained in memory at length 111, the persisted value was sanitized to length 37, and the two values differed. The focused `tests/request-log.test.ts` suite still passes (51/51), so the existing tests do not cover this ingress.

Please sanitize a copied entry at the start of `addRequestLog`, then use that same sanitized entry for both `retainRequestLogEntry` and persistence. Add a direct-`addRequestLog` regression covering control characters, secret-shaped content, the 64-character cap, and memory/disk equality. Keeping the normalization at this shared ingress will also protect relay and future direct callers, instead of relying on `addFinalRequestLog` alone.

lidge-jun:
## 리뷰 · 우선순위 52 / 80

#2157 백엔드 반쪽임. 섀도 인터셉트가 헬퍼를 다른 모델로 바꿔도 `shadowCallRewrittenFrom`이 `RequestLogContext`에 타입도 없고 `addFinalRequestLog`가 필드를 다시 조립해서 `/api/logs`랑 `usage.jsonl`에 안 남음. 재시작 하이드레이션도 없음. luna 헬퍼 비용이 그냥 행으로 보임.

#2163을 가져옴. 원본은 넣는 한 군데만 sanitize. 메모리 로그는 호출자가 준 값을 그대로 둠. 업스트림 모델 id라 개행이 라인 로그를 쪼갤 수 있음. 길이도 없음. 이번엔 `addFinalRequestLog`가 `sanitizeLogMetadataString`을 탐. 미래 호출자가 빼먹어도 구멍 안 열림. 메모리 행이랑 jsonl이 같음.

대시보드 `gui/src/pages/Logs.tsx`는 아직 이 필드를 안 그림. #2157을 여기서 닫으면 UI가 있는 척임. 본문이 그거 말함. 맞음. 열어두기. 스택 레이어 6. 베이스 #2165. types.ts/config.ts 안 건드림. #2163 리베이스하지 말 것.

점수는 52임. 라벨이 있어야 필터가 됨. 2.28. UI 후속 필요.

해결방안: #2165 다음. #2157은 Logs 필터 붙이면 닫기. 개행/64자 테스트 유지.

이 댓글은 grok-bot이 작성했습니다

Ingwannu:
Current head is still `70b947550`; no code changed after the changes-requested review.

The statement that the in-memory row and `usage.jsonl` row are identical is currently true only for callers that enter through `addFinalRequestLog`. The exported direct `addRequestLog(entry)` path still retains the caller-provided entry first and sanitizes only the later persistence copy, which is the exact memory/disk mismatch reproduced in the review.

The review remains blocked until the copied entry is sanitized at the start of `addRequestLog`, before `retainRequestLogEntry`, and that direct ingress has a regression asserting control/secret redaction, the 64-character cap, and memory/disk equality.

PR #2167, opened 2026-08-20
fix(codex): stop the background main refresh from retracting a reauth quarantine (#327)

ntdatt812:
## Problem

`tests/server-auth.test.ts > OpenAI option auth matrix keeps direct, pool, and API credentials independent` fails **reproducibly** on a clean `dev` tree — 78 pass / 1 fail, on win32, on every run. It is not load-dependent.

Narrowing it to the `reauth` iteration of the `["expired", "reauth", "cooldown"]` loop:

```
[state, status]
-   "reauth", 401
+   "reauth", 200
```

The account is marked `needsReauth`, and the request that must be refused with 401 is served instead.

## Root cause

Instrumenting the flag store shows the mark landing and then being erased milliseconds later:

```
[PROBE] mark id=__main__ writerGen=18 lastRecon=18 live=__main__ dropped=false
[PROBE] CLEAR __main__
    at clearAccountNeedsReauth (src/codex/account-runtime-state.ts:37)
    at fetchMainAccountInfoWhileOwned (src/codex/auth-api.ts:748)
    at async withNativeMainSharedClaim (src/codex/native-main-claim.ts:143)
```

`fetchMainAccountInfoWhileOwned` cleared `needsReauth` for `__main__` on **every** successful `/wham/usage` response, background refreshes included.

That is a stronger conclusion than the test alone. A 200 from `/wham/usage` proves the token authenticates to the *usage* endpoint. It does not prove the account can serve Responses traffic — a different backend path, which still answers **403 when the token's workspace can no longer be selected**. That is precisely the condition #327 was filed about:

> `~/.codex/auth.json`을 무효한 ChatGPT 토큰으로 만듭니다. (재현에 사용한 상태: 토큰에 박힌 workspace가 더 이상 선택 불가여서 upstream …)

So in production, with `codexAccountMode: pool` where `__main__` joins rotation:

1. Responses traffic 403s → `recordCodexUpstreamOutcome` takes the `outcomeClass === "credential"` branch and quarantines the account (`routing.ts`, *"401/403 quarantines the account for reauth"*).
2. The next background `/wham/usage` refresh returns 200 and silently retracts that quarantine.
3. `__main__` is a rotation candidate again, the next request fails identically, re-marks it — and step 2 wipes it again.

`needsReauth` therefore never settles, so the dashboard keeps showing nothing wrong. **That is the exact symptom #327 reported**, re-introduced through its own recovery path.

## Fix

One line: gate the clear on `forceRefresh`.

An explicit refresh (`?refresh=1` from the dashboard, `fetchMainAccountInfo(true)`) is an operator asking to re-evaluate, normally right after signing in again, so it stays authoritative — #327's recovery is unchanged, and its test still passes untouched. A background poll no longer counts as authentication evidence. Sign-in and account-add continue to clear unconditionally through `account-lifecycle`.

I deliberately did **not** simply delete the clear (which also turns `server-auth` green): that would regress #327's recovery. Two other designs were tried and measured first — a generation guard, and tagging marks with their origin — and both were discarded, the generation guard because the race happens inside a single generation (`writerGen=18`, `lastRecon=18`), the origin tagging because it broke `"successful main account refresh clears a prior needsReauth mark (#327)"`. `forceRefresh` is the axis that separates the two existing contracts cleanly.

## Verification — win32, on this exact rebased head

```
bun run typecheck                        exit 0
bun test tests/server-auth.test.ts        79 pass / 0 fail   (was 78 / 1)
bun test tests/codex-auth-api.test.ts    191 pass / 0 fail
```

10 adjacent auth / routing / oauth suites — `codex-routing`, `codex-auth-context`, `codex-main-rotation`, `oauth-health`, `oauth-reauth-bind`, `codex-account-delete-atomicity`, `cli-status-oauth-health`, `doctor-oauth`, `issue-452-empty-503`, `management-provider-validation` — **325 pass / 0 fail**.

**Mutation-checked.** Restoring the unconditional clear:

| suite | result |
|---|---|
| `codex-auth-api` | 1 fail — the new regression test |
| `server-auth` | 1 fail — OpenAI option auth matrix |

The new test asserts the probe genuinely ran (`usageCalls === 1`) and genuinely succeeded before checking the quarantine survived, so it cannot pass by accident of the probe being skipped.

## Note on scope

This is the only reproducible failure on clean `dev` on my machine. The other reddening I have seen in full-suite runs is load-dependent and reddens a different subset each run; none of it is touched here.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Background account refreshes no longer incorrectly clear a required reauthentication status.
  * Explicit refreshes continue to clear the reauthentication status after a successful account check.
  * Account identity changes now preserve the correct reauthentication state during automatic retries and account checks.

* **Tests**
  * Added coverage for background and explicit account refresh behavior, including identity-change retry scenarios.

### DIFF
--- a/src/codex/auth-api.ts
+++ b/src/codex/auth-api.ts
@@ -644,12 +644,13 @@ async function retryMainAccountInfoIfIdentityChanged(
   requestAccountId: string | null,
   retriesRemaining: number,
   nativeMainLease: AdmissionLease,
+  explicitRefresh: boolean,
 ): Promise<MainAccountInfoFetchResult | null> {
   const currentAccountId = getMainChatgptAccountId();
   if (currentAccountId === null || currentAccountId === requestAccountId) return null;
   reconcileMainCodexAccountRuntimeState();
   return retriesRemaining > 0
-    ? fetchMainAccountInfoWhileOwned(true, retriesRemaining - 1, nativeMainLease)
+    ? fetchMainAccountInfoWhileOwned(true, retriesRemaining - 1, nativeMainLease, explicitRefresh)
     : { info: EMPTY_MAIN_ACCOUNT_INFO, credentialChecked: true, hasCredential: true };
 }
 
@@ -696,6 +697,13 @@ async function fetchMainAccountInfoWhileOwned(
   forceRefresh: boolean,
   retriesRemaining: number,
   nativeMainLease: AdmissionLease,
+  /**
+   * Whether the *caller* asked for this refresh. `forceRefresh` also means "bypass the
+   * cache", and `retryMainAccountInfoIfIdentityChanged` re-enters with it set purely to
+   * re-read after the identity changed. Keeping the two apart stops that retry from
+   * promoting a background poll into operator intent below.
+   */
+  explicitRefresh: boolean = forceRefresh,
 ): Promise<MainAccountInfoFetchResult> {
   const writerGeneration = captureConfigGeneration();
   reconcileMainCodexAccountRuntimeState();
@@ -724,7 +732,7 @@ async function fetchMainAccountInfoWhileOwned(
     });
     if (!resp.ok) {
       const terminalAuthFailure = await isTerminalMainAuthResponse(resp, isMainAccountTokenVerifiablyLive());
-      const retried = await retryMainAccountInfoIfIdentityChanged(requestAccountId, retriesRemaining, nativeMainLease);
+      const retried = await retryMainAccountInfoIfIdentityChanged(requestAccountId, retriesRemaining, nativeMainLease, explicitRefresh);
       if (retried) return retried;
       if (terminalAuthFailure) {
         clearMainAccountInfoCache();
@@ -733,7 +741,7 @@ async function fetchMainAccountInfoWhileOwned(
       return { info: EMPTY_MAIN_ACCOUNT_INFO, credentialChecked: true, hasCredential: true };
     }
     const data = (await resp.json()) as WhamUsageResponse;
-    const retried = await retryMainAccountInfoIfIdentityChanged(requestAccountId, retriesRemaining, nativeMainLease);
+    const retried = await retryMainAccountInfoIfIdentityChanged(requestAccountId, retriesRemaining, nativeMainLease, explicitRefresh);
     if (retried) return retried;
     const plan = nonEmptyPlan(data.plan_type) ?? nonEmptyPlan(cached?.plan) ?? nonEmptyPlan(getMainAccountPlan());
     const quota = parseUsageQuota({ ...data, ...(plan ? { plan_type: plan } : {}) });
@@ -754,7 +762,7 @@ async function fetchMainAccountInfoWhileOwned(
     // never settled and the dashboard kept showing nothing — the symptom #327 reported.
     // An explicit refresh is an operator asking to re-evaluate, normally right after
     // signing in again, so it stays authoritative.
-    if (forceRefresh) clearAccountNeedsReauth(MAIN_CODEX_ACCOUNT_ID);
+    if (explicitRefresh) clearAccountNeedsReauth(MAIN_CODEX_ACCOUNT_ID);
     // Mirror main quota + plan into the shared stores so the rotation engine can
     // score and auto-switch the main account exactly like a pool account (Option A).
     setMainAccountPlan(result.plan);
@@ -769,7 +777,7 @@ async function fetchMainAccountInfoWhileOwned(
       ...(freshResetCredits !== undefined ? { freshResetCredits } : {}),
     };
   } catch {
-    const retried = await retryMainAccountInfoIfIdentityChanged(requestAccountId, retriesRemaining, nativeMainLease);
+    const retried = await retryMainAccountInfoIfIdentityChanged(requestAccountId, retriesRemaining, nativeMainLease, explicitRefresh);
     return retried ?? { info: EMPTY_MAIN_ACCOUNT_INFO, credentialChecked: true, hasCredential: true };
   }
 }
--- a/src/codex/auth-api.ts
+++ b/src/codex/auth-api.ts
@@ -745,7 +745,16 @@ async function fetchMainAccountInfoWhileOwned(
       ts: Date.now(),
     };
     setMainAccountInfoCache(result);
-    clearAccountNeedsReauth(MAIN_CODEX_ACCOUNT_ID);
+    // Only an explicit refresh may retract a reauth quarantine. A 200 from
+    // /wham/usage proves the token authenticates to the usage endpoint; it does not
+    // prove the account can serve Responses traffic, which is a different backend path
+    // and still answers 403 for a workspace the token may no longer select (#327).
+    // Letting the background poll clear the flag put such an account straight back into
+    // rotation: the next request failed the same way and re-marked it, so needsReauth
+    // never settled and the dashboard kept showing nothing — the symptom #327 reported.
+    // An explicit refresh is an operator asking to re-evaluate, normally right after
+    // signing in again, so it stays authoritative.
+    if (forceRefresh) clearAccountNeedsReauth(MAIN_CODEX_ACCOUNT_ID);
     // Mirror main quota + plan into the shared stores so the rotation engine can
     // score and auto-switch the main account exactly like a pool account (Option A).
     setMainAccountPlan(result.plan);
--- a/tests/codex-auth-api.test.ts
+++ b/tests/codex-auth-api.test.ts
@@ -13,7 +13,7 @@ import {
   handleCodexAuthAPI, updateAccountQuota, getAccountQuota,
   checkAccountIdCollision, getMainChatgptAccountId,
   markAccountNeedsReauth, isAccountNeedsReauth, clearAccountNeedsReauth, clearAccountQuota,
-  clearMainAccountInfoCache, maskEmail,
+  clearMainAccountInfoCache, maskEmail, fetchMainAccountInfo,
   clearCodexQuotaPrimeState, primeCodexPoolQuotas, seedCodexAuthAdmissionForTests,
   type CodexAuthAccountDto,
   listCodexAuthAccounts,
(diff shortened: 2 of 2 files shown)


ntdatt812:
Two things on this PR — one for CodeRabbit, one that needs a maintainer.

## CodeRabbit's merge-risk finding was correct

> A background account refresh can be retried as an explicit refresh after an identity change, clearing a reauthentication quarantine and allowing an unusable account back into rotation.

Real hole, and my fault: `forceRefresh` carries two meanings — *"bypass the cache"* and *"the caller asked for this"* — and I gated the clear on it without separating them. `retryMainAccountInfoIfIdentityChanged` re-enters `fetchMainAccountInfoWhileOwned(true, …)` purely to re-read past a now-stale cache, so a background poll that crossed an identity change came back holding operator authority it never had.

Fixed in `bbf2aadec` by splitting the two:

```ts
async function fetchMainAccountInfoWhileOwned(
  forceRefresh: boolean,
  retriesRemaining: number,
  nativeMainLease: AdmissionLease,
  explicitRefresh: boolean = forceRefresh,   // the caller's own intent
)
```

The retry propagates the caller's original value; cache-bypass behaviour is untouched. The clear now keys off `explicitRefresh`.

**One correction to the report's framing**, because it matters for anyone reading this later: on a *genuine* identity change the quarantine is dropped by a different and entirely correct path — `reconcileMainCodexAccountRuntimeState` → `purgeMainCodexAccountRuntimeState` purges the previous identity's quarantine, quota, health and thread map, because it is a different ChatGPT account. That is not the bug. The bug is narrower: a quarantine raised against the **new** identity, after that purge, while the retry probe is still in flight.

My first attempt at the regression test asserted the purge behaviour and failed for that reason — worth saying plainly rather than quietly rewriting it. The test that landed marks the account from inside the retry's own probe, so it pins the retry path and not the purge:

```ts
if (usageCalls === 1) { /* identity changes on disk → triggers the retry */ }
else { markAccountNeedsReauth(MAIN_CODEX_ACCOUNT_ID); }  // traffic quarantines the NEW identity
...
expect(usageCalls).toBeGreaterThan(1);   // the retry really fired
expect(isAccountNeedsReauth(MAIN_CODEX_ACCOUNT_ID)).toBe(true);
```

### Verification on win32, this exact head

```
bun run typecheck                     exit 0
12 auth / routing / oauth suites      596 pass / 0 fail
```

Mutation-checked, each hole caught by its own test:

| mutation | result |
|---|---|
| clear gated on `forceRefresh` (this hole) | 1 fail — the new retry test |
| clear unconditional (the original bug) | 2 fail — both regression tests |
| neither | 596 pass / 0 fail |

## `unsponsored_surface` needs a maintainer

Both `PR hygiene` and `Enforce PR target branch` fail on `unsponsored_surface` because the diff touches `src/codex/auth-api.ts`:

> `MAINTAINERS.md` requires security review for these; ask a maintainer to apply `maintainer-sponsored` once they have reviewed it.

That is the gate working as designed and I cannot clear it myself. The change is one line of control flow plus its threading — no credential is read, written, logged, or transmitted differently, and the direction is strictly fail-closed: an account stays quarantined in cases where it previously became usable again. Whenever someone has time to look, the label is all that is left here.

ntdatt812:
@Ingwannu — requesting the security review this PR is gated on, since `MAINTAINERS.md` lists it under your responsibilities:

> Authentication, credential handling, GitHub Actions, release automation, dependency installation, and other security-boundary changes require explicit security review.

Everything else on the PR is now settled: CodeRabbit re-reviewed at `bbf2aadec` and moved merge risk to ⚪ Minimal (*"no actionable merge-blocking risk remains"*), there are zero open review threads, the branch is 0 behind `dev`, and all four checklist boxes are ticked. `unsponsored_surface` is the only thing left, and it is not something a fork contributor can clear.

To keep the review short, the complete source-level delta is three mechanical lines plus one guard:

```diff
   nativeMainLease: AdmissionLease,
+  explicitRefresh: boolean = forceRefresh,
 )
-  ? fetchMainAccountInfoWhileOwned(true, retriesRemaining - 1, nativeMainLease)
+  ? fetchMainAccountInfoWhileOwned(true, retriesRemaining - 1, nativeMainLease, explicitRefresh)
-  retryMainAccountInfoIfIdentityChanged(requestAccountId, retriesRemaining, nativeMainLease)      // ×3
+  retryMainAccountInfoIfIdentityChanged(requestAccountId, retriesRemaining, nativeMainLease, explicitRefresh)
-    clearAccountNeedsReauth(MAIN_CODEX_ACCOUNT_ID);
+    if (explicitRefresh) clearAccountNeedsReauth(MAIN_CODEX_ACCOUNT_ID);
```

At the security boundary specifically:

- **No credential is read, written, cached, logged, or transmitted differently.** No token, header, URL, or storage path is touched; `readCodexTokensResult`, the `/wham/usage` call and its `Authorization` header are all unchanged.
- **The direction is strictly fail-closed.** The only behavioural difference is that `__main__` *stays* quarantined in cases where it previously became usable again. Nothing becomes usable that was not usable before, so the change cannot widen access.
- **The quarantine-clearing paths that remain are unchanged**: an explicit refresh, a completed sign-in, and account-add via `account-lifecycle` all still clear unconditionally. `reconcileMainCodexAccountRuntimeState`'s purge on a genuine identity change is untouched.

The motivating defect is a livelock rather than a leak: traffic 401/403 quarantines the account, the next background `/wham/usage` 200 retracts it, the account returns to rotation and fails identically — so `needsReauth` never settles and the dashboard shows nothing, which is the symptom #327 was filed about.

Local verification on win32 at this head: `typecheck` exit 0, and 596 pass / 0 fail across the 12 auth/routing/oauth suites, with each failure mode mutation-checked against its own test. Fork PRs cannot start repository CI, so that attestation is all I can offer from my side.

lidge-jun:
## 리뷰 · 우선순위 66 / 80

실화임. 지금 `auth-api.ts` 748줄 `fetchMainAccountInfoWhileOwned`가 `/wham/usage` 200이면 무조건 `clearAccountNeedsReauth("__main__")` 함. 백그라운드 리프레시도. usage 200은 토큰이 usage 엔드포인트에 먹힌다는 뜻이지 Responses가 된다는 뜻이 아님. 워크스페이스 선택 불가면 Responses 403. 그게 #327임.

풀 모드에서 `__main__`이 로테이션에 들어가면: Responses 403 → `recordCodexUpstreamOutcome` credential 분기가 격리 (`routing.ts`) → 다음 백그라운드 usage 200이 격리를 지움 → 다시 후보 → 또 403. `needsReauth`가 안 굳음. 대시보드가 멀쩡해 보임. `server-auth` 매트릭스 `reauth`가 401 대신 200 나오는 게 그 레이스임. 윈도우에서 매 런 재현.

픽은 `forceRefresh`일 때만 지움. 대시보드 `?refresh=1`은 로그인 직후 재평가라 #327 복구 테스트는 그대로. 사인인/계정 추가는 `account-lifecycle`에서 무조건 클리어. 클리어를 통째로 지우면 복구가 죽음. types.ts/config.ts 안 건드림. 초안임. MERGEABLE.

점수는 66임. 2.28에 넣을 만함. 크레덴셜 루프라 드래프트라도 빨리.

해결방안: `forceRefresh` 가드 유지. 백그라운드 200으로 격리 안 풀리는 테스트 유지. CI 초록이면 드래프트 해제.

이 댓글은 grok-bot이 작성했습니다

ntdatt812:
Thanks for confirming it — your read matches what I measured, including the part that shaped the fix: deleting the clear outright turns `server-auth` green but kills #327's recovery, so the guard had to be conditional rather than removed.

**Your undraft condition is met on the real CI.** `Cross-platform CI` is `success` on this exact head `bbf2aadec` — run [32326938735](https://github.com/lidge-jun/opencodex/actions/runs/32326938735), 17 of 18 jobs succeeded and the one skip is the `windows ${{ matrix.shard }}/4` matrix gate, not a failure. `PR Labeler` is green too. The PR is already out of draft.

The two checks still red are both the same policy gate, not the code:

| check | result |
|---|---|
| `PR hygiene` | `unsponsored_surface` |
| `Enforce PR target branch` | `unsponsored_surface` |

> This changes an authentication, workflow, release-automation, or dependency surface. `MAINTAINERS.md` requires security review for these; ask a maintainer to apply `maintainer-sponsored` once they have reviewed it. Paths: `src/codex/auth-api.ts`.

A fork contributor cannot clear that one. Since you have just done the security review in this comment, applying `maintainer-sponsored` is the single remaining step and both checks go green.

`React Doctor` shows `cancelled` — it was killed at 03:14 when the gate moved the PR back to draft mid-run, not by anything in the diff. Nothing here touches React; happy for it to be re-run whenever.

On the two things you flagged to keep, both are already in place and pinned by tests:

- the `forceRefresh` guard stays — and `explicitRefresh` was added in `bbf2aadec` after CodeRabbit found the retry could launder a background refresh into an explicit one via `retryMainAccountInfoIfIdentityChanged`
- a background 200 not releasing the quarantine is a test, not just behaviour: `a background main account refresh does not retract a reauth quarantine (#327)`, plus `an identity-change retry does not upgrade a background refresh into an explicit one (#327)`

Mutation-checked, each failure mode against its own test: unconditional clear → both tests fail; gated on `forceRefresh` instead of `explicitRefresh` → the retry test fails; neither → 596 pass / 0 fail across the 12 auth/routing/oauth suites locally on win32.

`types.ts` and `config.ts` are untouched, as you noted. Ready for 2.28 whenever the label goes on.

ntdatt812:
Correction to my comment above: it went out at 03:57 and this had already merged at 03:55, so the request for `maintainer-sponsored` was two minutes stale. I read `mergeStateStatus` and not `state` when I checked. Ignore that ask — nothing is needed.

The CI figures in it are still accurate and were measured on the merged head `bbf2aadec`, in case they are useful for the 2.28 notes: `Cross-platform CI` succeeded with 17 of 18 jobs, the single skip being the `windows ${{ matrix.shard }}/4` matrix gate.

Thanks for the review and the merge.

PR #2170, opened 2026-08-20
fix(logs): record the configured shadow prefix, not the caller's model string

lidge-jun:
## Summary

Release audit of `origin/main..origin/dev` found a credential-leak path in the shadow-call logging that landed in this range. This is the fix.

**The defect.** The shadow-call intercept matches by *prefix*, so a caller can send `gpt-5.6-luna` followed by arbitrary text and still be intercepted. The whole raw string was then recorded as `shadowCallRewrittenFrom`, which is persisted to `usage.jsonl` and served from `/api/logs`.

The sanitizer on that path is not sufficient protection, for two independent reasons — both verified against the shipped code:

```
"gpt-5.6-luna\nBearer sk-abc123def456ghi789jkl"  ->  "gpt-5.6-lunaBearer [REDACTED]"
"gpt-5.6-luna\nAIzaSyA1B2C3D4E5F6G7H8I9J0K1L2M3N4O5P6"  ->  unchanged, key intact
```

1. Control characters are stripped *before* redaction runs, so the newline that separated the marker from the credential is gone by the time the `Bearer` rule looks for a word boundary.
2. The runtime redactor is a deny-list. An `AIza`-shaped Google key has no rule at all and survives verbatim into the log file.

**The fix.** Record the operator-configured prefix that matched, not the caller's string. `shadowSourceModelPrefix()` returns the entry from `sourceModels` that produced the match, so the field can only ever hold a value the operator configured. That removes the class rather than adding one more pattern to a deny-list — which is the right shape here, because the next unrecognized credential family would reopen it.

## Verification

- **RED-first.** Reverting only `src/lib/shadow-call.ts` and `src/server/responses/core.ts` fails exactly the two new tests (15 pass / 2 fail). The first pins the `AIza` smuggling case; the second pins that a non-default configured prefix is recorded as itself, so the fix cannot degrade into a hardcoded string.
- `bun x tsc --noEmit` — exit 0 (on `ssh lidge`).
- `bun test --isolate tests/responses-shadow-intercept.test.ts` — 17 pass / 0 fail.
- `bun run test` on `ssh lidge` — **13713 pass / 15 skip / 0 fail** across 866 files.

## Checklist

This *is* the security-sensitive change: it narrows what can reach a persisted log field. No credential is logged by the new code, and the field's value set is now bounded by operator configuration.

## Summary by CodeRabbit

- **Security & Privacy**
  - Shadow-call logs now record only the configured source-model prefix instead of the full caller-provided model identifier.
  - Prevents credential-like or otherwise sensitive content appended to model IDs from being persisted.

- **Bug Fixes**
  - Improved handling of custom source-model prefixes and routed model identifiers.

- **Tests**
  - Added regression coverage for sanitized logging and custom prefix configurations.

### DIFF
--- a/src/lib/shadow-call.ts
+++ b/src/lib/shadow-call.ts
@@ -30,6 +30,21 @@ export function isShadowSourceModel(modelId: string, configured?: unknown): bool
 }
 
 /**
+ * The configured source prefix this model matched, or undefined.
+ *
+ * Callers that RECORD the intercepted model must record this rather than the caller's raw
+ * `modelId`. Matching is by prefix, so `gpt-5.6-luna` plus arbitrary trailing text still
+ * intercepts — and the raw string is caller-controlled, reaches `usage.jsonl` and `/api/logs`,
+ * and only passes a pattern-based redactor on the way. A credential family that redactor does
+ * not recognize survives verbatim. Returning the operator-configured prefix keeps the log
+ * field inside a set the operator chose, so no caller string is ever persisted.
+ */
+export function shadowSourceModelPrefix(modelId: string, configured?: unknown): string | undefined {
+  if (modelId.includes("/")) return undefined;
+  return shadowSourceModels(configured).find(prefix => modelId.startsWith(prefix));
+}
+
+/**
  * Decide whether a matching source model should use the opt-in intercept.
  *
  * Before Codex 0.147.0 this checked x-codex-turn-metadata and exempted
--- a/src/server/responses/core.ts
+++ b/src/server/responses/core.ts
@@ -315,7 +315,7 @@ export function sidecarOutcomeRecorder(
 
 
 
-import { isShadowSourceModel, shouldInterceptShadowCall } from "../../lib/shadow-call";
+import { isShadowSourceModel, shadowSourceModelPrefix, shouldInterceptShadowCall } from "../../lib/shadow-call";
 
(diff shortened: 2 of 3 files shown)


lidge-jun:
## 리뷰 · 우선순위 74 / 80

실화임. 섀도 인터셉트가 prefix 매칭이라 호출자가 `gpt-5.6-luna` 뒤에 아무 텍스트를 붙여도 잡힘. 그 원문 전체가 `shadowCallRewrittenFrom`으로 `usage.jsonl`이랑 `/api/logs`에 남음. 새니타이저가 보호가 아님. 제어문자를 레드액션보다 먼저 지워서 `gpt-5.6-luna\nBearer sk-...`가 `gpt-5.6-lunaBearer [REDACTED]`가 됨. `AIza` 구글 키는 deny-list에 없어서 그대로 파일에 남음. `privacy:scan`은 레포만 봐서 런타임 로그 구멍을 못 봄.

픽이 맞음. 호출자 문자열이 아니라 매칭된 `sourceModels` prefix만 기록. 필드가 오퍼레이터가 설정한 값만 담음. 패턴 하나 더 넣는 게 아님. 다음 키 패밀리가 다시 안 열림. types.ts/config.ts 안 건드림. MERGEABLE.

점수는 74임. 2.28 블로커. #2166 sanitizer 위에 있는 구멍이라 그거 넣었다고 안 막힘.

해결방안: 머지. AIza 밀수 테스트랑 non-default prefix 테스트 유지. deny-list 확장으로 대체하지 말 것.

이 댓글은 grok-bot이 작성했습니다

PR #2173, opened 2026-08-20
fix(responses): give tool_search_call and custom_tool_call their own id prefixes

lidge-jun:
## Summary

Release audit of `origin/main..origin/dev` found this by composing two changes that are each correct alone — the shape the audit was specifically looking for.

**The defect.** A routed `tool_search` lowering is restored as a `tool_search_call` with no id (#2145), and the universal output-item id backfill then names it (#2142). The backfill's prefix table had no entry for the type, so it fell through to the generic `item_`:

```
{"type":"tool_search_call","call_id":"call_x",...}  ->  id: "item_ocx_0"
{"type":"function_call","name":"x",...}             ->  id: "fc_ocx_0"
```

That is not cosmetic. `stripInvalidItemIds` in the Responses adapter deletes any id whose prefix does not match its type, and it lists `tsc_` as the only valid prefix for `tool_search_call`. So the synthesized id survived the turn that created it and was **silently dropped on the next one**, leaving the client an item it could not correlate.

Neither PR's focused suite caught it, because neither composes missing-id restoration with the backfill. That is exactly why this needed a cross-PR pass rather than per-PR CI.

`custom_tool_call` had the identical gap and is fixed alongside it: the serializer enforces `ctc_`, the backfill did not know the type.

The two tables are now a stated invariant — the backfill's prefixes must remain a superset of the ones the serializer enforces — recorded in a comment at the table so the next type added does not reopen it.

## Verification

- **RED-first.** Reverting only `src/server/responses/responses-field-backfill.ts` fails exactly the two new tests (20 pass / 2 fail).
- `bun x tsc --noEmit` — exit 0.
- `bun test --isolate tests/responses-field-backfill.test.ts` — 22 pass / 0 fail.
- `bun run test` on `ssh lidge` — **13717 pass / 15 skip / 0 fail** across 866 files.

## Checklist

No credential or auth surface is touched; this only changes a synthesized id prefix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected backfilled tool-search and custom tool-call IDs to use the appropriate serializer-compatible prefixes.
  * Improved consistency and compatibility of generated response item identifiers.

* **Tests**
  * Added regression coverage for tool-search and custom tool-call ID generation.

### DIFF
--- a/src/server/responses/responses-field-backfill.ts
+++ b/src/server/responses/responses-field-backfill.ts
@@ -34,6 +34,14 @@ const ITEM_ID_PREFIXES: Readonly<Record<string, string>> = {
   message: "msg_",
   reasoning: "rs_",
   function_call: "fc_",
+  custom_tool_call: "ctc_",
+  // A routed tool_search lowering is restored to `tool_search_call` without an id, so this
+  // backfill is what names it. The generic `item_` fallback is not merely cosmetic here:
+  // `stripInvalidItemIds` in the Responses adapter deletes any id whose prefix does not match
+  // the type, so an `item_`-named tool_search_call silently loses its id on the NEXT turn and
+  // the client sees an item it cannot correlate. The prefixes here must stay a superset of the
+  // ones that serializer enforces.
+  tool_search_call: "tsc_",
   web_search_call: "ws_",
   file_search_call: "fs_",
   code_interpreter_call: "ci_",
--- a/tests/responses-field-backfill.test.ts
+++ b/tests/responses-field-backfill.test.ts
@@ -365,6 +365,36 @@ describe("responses-field-backfill", () => {
     expect(result.output[0]!.id).toBe("ig_ocx_0");
   });
 
+  // A routed tool_search lowering is restored as `tool_search_call` with no id, so this
+  // backfill names it. The generic `item_` fallback was not cosmetic: `stripInvalidItemIds`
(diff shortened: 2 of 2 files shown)


lidge-jun:
## 리뷰 · 우선순위 70 / 80

#2145랑 #2142를 붙이면 나오는 구멍임. 라우티드 tool_search가 `tool_search_call`로 복원되는데 id가 없음. 백필이 타입 테이블에 없어서 `item_ocx_0`을 줌. `stripInvalidItemIds`는 `tsc_`만 인정해서 다음 턴에 id가 조용히 사라짐. 클라이언트가 아이템을 못 이음. 각자 스위트는 조합을 안 돌려서 통과함. `custom_tool_call`도 동일. 시리얼라이저는 `ctc_`, 백필은 모름.

백필 prefix가 시리얼라이저가 강제하는 집합의 상위집합이어야 함. 테이블 주석으로 박은 거 맞음. types.ts/config.ts 안 건드림. MERGEABLE.

점수는 70임. 2.28에 넣어야 함. #2145+#2142가 이미 `dev`에 있어서 지금 깨진 상태임.

해결방안: 머지. tsc_/ctc_ 테스트 유지. 새 output 타입 넣을 때 두 테이블 같이.

이 댓글은 grok-bot이 작성했습니다

PR #2175, opened 2026-08-20
fix(devlog): unbreak privacy:scan on dev

lidge-jun:
## Summary

`dev`'s own release gate is currently failing, and has been since `0637a3548`. This unbreaks it.

That commit's audit record pasted a reproduction transcript containing a literal `Bearer <value>` line. `scripts/privacy-scan.ts` matches that shape regardless of whether the value is a real credential:

```
devlog/_plan/.../100_release_safety_audit.md:30 bearer-token: Bearer ocx_data_...
error: script "privacy:scan" exited with code 1
```

The blast radius is wider than the one file: `privacy:scan` runs in the `gates` job, so **every PR branched from `dev` inherits the failure**. That is how this was found — an unrelated PR's `gates` job went red.

The value was a placeholder from a local reproduction, not a real secret. But the scanner cannot know that, and it should not have to: a documented transcript that looks exactly like a leaked credential is indistinguishable from one. Describing the header instead of pasting it keeps the finding legible and the gate green.

## Verification

- `bun run privacy:scan` — passed (fails on `dev` without this change; verified both directions on `ssh lidge`).

Docs-only; no runtime code touched.

## Checklist

This removes a credential-shaped string from the tree rather than adding one.

## Summary by CodeRabbit

* **Documentation**
  * Redacted a reproduced authorization value in the release safety audit and replaced it with a placeholder.
  * No functional behavior changed.

### DIFF
--- a/devlog/_plan/260820_bug_pr_backlog_consolidation/100_release_safety_audit.md
+++ b/devlog/_plan/260820_bug_pr_backlog_consolidation/100_release_safety_audit.md
@@ -27,7 +27,7 @@ and the adapter forwarded our own proxy secret. Reproduced through the real adap
 
 ```
 URL:  https://chatgpt.com/backend-api/codex/responses
-AUTH: Bearer ocx_data_this_is_our_proxy_key
+AUTH: Bearer <the proxy's own admission secret, verbatim>
 LEAKED: true
 ```
 
@@ -120,4 +120,3 @@ Version line on `dev` is `2.27.0`, equal to the published `latest` and the `v2.2
 pre-release state, not a half-finished bump.
 
 This remains a readiness record. No release was executed.
-


lidge-jun:
## 리뷰 · 우선순위 64 / 80

지금 `dev` 게이트가 빨강임. `0637a3548` 감사 기록이 `Bearer <value>` 줄을 그대로 붙여서 `scripts/privacy-scan.ts`가 `100_release_safety_audit.md`에서 터짐. gates 잡이라 `dev`에서 딴 모든 PR이 상속함. 플레이스홀더지 실키 아님. 스캐너는 모름. 생긴 게 유출이랑 같으면 실패가 맞음. 헤더를 설명으로 바꾸면 게이트가 다시 초록.

문서만. 런타임 없음. MERGEABLE. types.ts/config.ts 안 건드림.

점수는 64임. 제품 버그는 아닌데 지금 랜딩을 막고 있음. #2170보다 먼저 넣어도 됨.

해결방안: 바로 머지. 재현 트랜스크립트에 토큰 모양 문자열 다시 붙이지 말 것.

이 댓글은 grok-bot이 작성했습니다

PR #2228, opened 2026-08-20
fix(responses): keep compaction blobs on the backend that minted them

olddonkey:
## Summary

A replayed compaction item carries an `encrypted_content` blob only its minting backend can decode, and Codex replays it in `input` on every later turn. Two paths modified or misrouted it. Both **wedge the session** rather than failing one turn: the item outlives the failure in the client transcript, and the routed compaction turn opencodex itself drives replays the same item, so the user cannot compact their way out — only clearing history recovers.

**Relay provenance.** `scrubOcxCompactionItems` treated every non-`ocx1:` blob as OpenAI's and forwarded it verbatim, with no check that the destination was the issuer. A session that compacted on a canonical route and then switched to a routed provider sent that blob to an upstream that could only answer `Could not decode the compaction blob. Ensure it is unmodified from the compact response.` Native blobs now travel only to destinations that mint them, and degrade elsewhere to the same opaque note the bridged parser already uses for this case.

The destination test is `destinationDecodesNativeCompactionBlob`: forward-auth routes (which relay the caller's own OpenAI credentials to the ChatGPT backend or a self-hosted relay in front of one) plus the official OpenAI API under key auth. Everything else is a routed provider carrying its own credentials to a non-OpenAI backend, which can only reject a blob it never issued. Degrading is deliberately the conservative direction — a self-hosted relay keeps its users' compacted history, and only a destination we are sure cannot decode loses it.

**Backfill exemption.** The response-side exemption list named `compaction` alone, so `compaction_summary` and `context_compaction` received synthesized ids (`item_ocx_1`, …) that the client stored and replayed — literally "modified from the compact response", with no route change needed. That divergence was possible because the compact wire family was enumerated in three places; it is now one predicate in `src/responses/compaction.ts`, and the parser's copy was folded into it.

## Scope

`stripItemIdsWhenUnstored` deliberately still strips ids from compaction items. An earlier draft exempted them, but codex-rs strips ids from every item under `store: false` (`core/src/client.rs:918-925`) and OpenAI accepts that, so stripping is the contract rather than a corruption; after the backfill fix there is also no known source of an id on those items. Diverging from the reference client without evidence would have added risk for no gain.

The sibling `Could not decrypt the provided encrypted_content` failure is **not** addressed here. It has a different root cause (the reasoning item is reshaped on the response leg) and needs a live canary to confirm, so it is a separate PR.

## Verification

- **RED-first**: reverting only `src/` fails 2 of the new tests (`every compact wire item type is returned byte-for-byte`, `degrades a foreign blob to a note on a destination that cannot decode it`). The other new cases are regression guards that pass before and after — `ocx1:` lowering, bare `context_compaction` markers, and native relay to minting destinations.
- `bun test` on the 39 compaction/responses/adapter/parser suites — 631 pass, 0 fail.

## Checklist

- [x] Docs or release notes were updated when needed. (`structure/04_transports-and-sidecars.md` records the provenance boundary and a Decision Log.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. No new persistence, logging, credential handling, or destination; the change only narrows where an opaque payload may be relayed. Privacy scan green.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compaction replay across supported destinations.
  * Converted incompatible or proxy-generated compaction blobs into readable user messages.
  * Preserved native compaction blobs when the destination supports them.
  * Kept all recognized compaction markers unchanged and prevented unwanted ID modifications.
  * Added support for explicitly configured destinations that can decode native compaction data.
* **Documentation**
  * Clarified compaction replay, destination compatibility, and data sanitization behavior.

## Gate

`bun run test` — 13738 pass, 10 skip, 1 fail across 866 files, on `abc2b9f4f`.

The single failure is `tests/key-login-live-update.test.ts` > "notify after key login pushes the merged row and keeps modelCosts on live and disk". It is **pre-existing and unrelated**: it reproduces byte-identically on every branch in this series, including ones that never touch CLI code, and on the integrated branch. Every gate in this series lands on exactly that one failure.

### DIFF
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -2,11 +2,11 @@ import { createHash } from "node:crypto";
 import type { IncomingMeta, ProviderAdapter } from "./base";
 import { namespacedToolName, type AdapterEvent, type OcxParsedRequest, type OcxProviderConfig, type OcxUsage, type TierDecision } from "../types";
 import { catalogModelSupportsReasoningSummaries } from "../codex/catalog";
-import { COMPACT_PROMPT, decodeCompactionSummary, SUMMARY_PREFIX } from "../responses/compaction";
+import { COMPACT_PROMPT, compactionItemToText, decodeCompactionSummary, isCompactionItemType } from "../responses/compaction";
 import { collectResponsesToolGroups } from "../responses/tool-groups";
 import { isHostedToolUnsupportedForModel } from "../responses/hosted-tool-policy";
 import { decodeServerSentEvents } from "../lib/sse-decoder";
-import { CODEX_FORWARD_BASE_URL, isCanonicalOpenAiForwardProvider } from "../providers/openai-tiers";
+import { CODEX_FORWARD_BASE_URL, destinationDecodesNativeCompactionBlob, isCanonicalOpenAiForwardProvider } from "../providers/openai-tiers";
 import { OCX_REASONING_PREFIX } from "../responses/reasoning-envelope";
 import { modelRecordValue } from "../reasoning-effort";
 import type { TranslatorBudget } from "../lib/translator-budget";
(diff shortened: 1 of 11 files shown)


lidge-jun:
## 리뷰 · 우선순위 70 / 80

지금 `dev` `src/adapters/openai-responses.ts` `scrubOcxCompactionItems` (`:151-169`)가 `ocx1:`만 평탄화함. `decodeCompactionSummary`가 null이면 native blob을 목적지 안 보고 그대로 포워드함. 캐논 경로에서 컴팩션하고 라우티드 프로바이더로 바꾸면 xAI 같은 곳이 `Could not decode the compaction blob` 냄. 아이템이 클라이언트 트랜스크립트에 남아서 다음 턴마다, 프록시가 돌리는 컴팩션 턴까지 같은 blob을 다시 보냄. 히스토리 클리어 전까지 세션이 못 빠져나옴.

응답 레그는 더 좁음. `src/server/responses/responses-field-backfill.ts` `NON_RESPONSES_ITEM_TYPES`가 `"compaction"`만 넣음 (`:130`). `compaction_summary` / `context_compaction`은 `item_ocx_1` 합성 id를 받음. 클라이언트가 그 수정된 아이템을 리플레이하면 민팅 백엔드가 unmodified 검사에서 거절함. 루트 바꿀 필요도 없음. compact 와이어 패밀리가 세 군데로 쪼개져 있어서 생긴 구멍임.

이 PR이 두 구멍을 같이 막음. `src/responses/compaction.ts` `isCompactionItemType`이 유일한 열거임. 스크럽이 `destinationDecodesNativeCompactionBlob` (`src/providers/openai-tiers.ts`)로 목적지를 봄. forward-auth랑 `api.openai.com/v1` 키만 native blob 통과. 나머지는 `compactionItemToText`로 `OPAQUE_COMPACTION_NOTE`. `ocx1:`은 전 목적지에서 평탄화. 백필 면제가 세 타입 전부. `stripItemIdsWhenUnstored`는 compaction id를 그대로 뗌. 그게 레퍼런스 클라이언트 계약임.

테스트가 라우티드 xAI 베이스에서 blob이 안 새는지, OpenAI 키/포워드 릴레이는 통과하는지, ocx1 로워링, bare marker 유지, 백필 byte-for-byte를 잠금. 방향 맞음. `src/`만 되돌리면 새 2케이스가 깨진대. ㅇㅇ 그게 RED임.

주의. `destinationDecodesNativeCompactionBlob`이 `authMode === "forward"`면 어댑터/URL 안 보고 true임. 셀프호스트 릴레이 히스토리 지키려고 한 거임. 포워드가 OpenAI가 아닌 게이트웨이를 가리키면 native blob이 또 거절됨. 지금은 그런 시드 없음. 아이템에 민팅 아이덴티티를 찍는 건 이 PR 범위 밖임. `structure/04_transports-and-sidecars.md` Decision Log가 그 트레이드오프 적음.

#2229랑 원인 다름. 저건 reasoning `encrypted_content` 리셰이프임. 파일은 `structure/04_transports-and-sidecars.md`만 겹침. #2217/#2227 `modelWireDefaults.wire` 싸움이랑 무관. 이 PR은 레지스트리 기본 와이어 안 건드림. Chat 기본이 돼도 컴팩션 리플레이 구멍은 남음. 닫지 말 것. types.ts/config.ts 스플릿 안 씹힘. openai-tiers/compaction/adapter/parser/backfill임. #2188 사이드카, #2190 `x_search`랑 섞지 말 것.

draft고 체크리스트 0/4. hygiene 통과. 세션 웻지라 점수는 높음. 2.28 블로커는 아님. 지금 HEAD `03735eca6`가 Grok 4.5/4.6 OAuth+responses를 `openai-responses`로 박아서 라우티드 컴팩션 경로에서 더 잘 터짐.

**해결방안:** CI 그린이면 draft 해제하고 `dev` 머지. #2229랑 같이 넣어도 됨. 기본 와이어는 #2217/#2227에서 따로 고름. 스플릿이 compaction 헬퍼를 옮기면 리베이스하지 말고 닫고 다시 짜라. 지금은 그 정도 아님.

이 댓글은 grok-bot이 작성했습니다

olddonkey:
Part of #2240 — compaction blob provenance — a blob relayed to a backend that could not have minted it.

Ingwannu:
Maintainer status on exact head add8a55f26ef06af3e9f83e134a2920df03b78ff: the backend-provenance fix remains valuable, but this draft is not ready to merge. The readiness checklist is 0/4, the branch is 24 commits behind dev, and the documentation accuracy thread about response-side ID backfill versus store:false stripping is still unresolved. Please address that thread, synchronize with current dev, and complete exact-head CI before requesting final review. Keep #2240 open; this is one sequential layer, not the closing PR.

olddonkey:

Issue #45, opened 2026-06-28
Reasoning tokens not expanding in Codex UI - "Worked for X" shows time but no trace


Rezhnn:
## What happens
 
 When I use opencodex with any proxy model (DeepSeek v4 flash free, Mimo v2.5 free, etc.), the Codex app shows "Worked for Xs" but when I click on it, there's nothing to expand. No reasoning trace, no thinking steps, nothing.
 
 With original OpenAI models (e.g. gpt-5.5), it works normally - For example i see "Worked for 33m 6s" and when I click on it, there's a full expandable section showing all the reasoning/thinking steps.
 
 ## What I expected
 
 The expandable reasoning trace to appear the same way it does with native OpenAI models, since the README says reasoning tokens are supported.
 
 ## What I tried
 
 - Different proxy models (DeepSeek v4 flash free, Mimo v2.5 free) - same result on both
 - The timing itself works fine ("Worked for 23s" etc), so the connection is working
 - Native OpenAI models through the same Codex app work correctly
 
 ## My setup
 
 - Codex desktop app (latest version)
 - Proxy: opencode zen API key through opencodex's hosted proxy
 - Models tested: DeepSeek v4 flash free, Mimo v2.5 free
 
 ## Question
 
 I'm not super experienced with API response formats, so I'm not sure if this is:
 
 1. A bug in how opencodex formats reasoning tokens in the response
 2. A limitation of the free-tier models not returning the same reasoning structure as OpenAI
 3. Something on Codex's side that doesn't recognize non-OpenAI reasoning token formats
 4. Or something else I'm not thinking of
 
 Would love to understand what's going on here. If there's any log or config I can check on my end to help debug, happy to do that. Thanks!

lidge-jun:
Investigated end-to-end, including the codex-rs consumer side.

**Root cause (confirmed):** Codex renders the *expandable* reasoning trace from the Responses `reasoning` item's **`summary[]`** channel. opencodex maps chat-completions `reasoning_content` to a raw-reasoning item with `summary: []` and only `content: [{reasoning_text}]`, so the Codex app shows the "Worked for Xs" timer (driven separately by elapsed time / `reasoning_tokens`) but has nothing to expand.

I traced this through codex-rs `main` to be sure it isn't a client bug:
- `codex-api/src/sse/responses.rs` parses both `response.reasoning_summary_text.delta` (→ summary) and `response.reasoning_text.delta` (→ content).
[...]

PR #2229, opened 2026-08-20
fix(responses): stop reshaping reasoning items that carry encrypted_content

olddonkey:
## Summary

Codex replays the reasoning item it received in the next request's `input`, and a backend that issued `encrypted_content` verifies what comes back. The content-to-summary channel rewrite deletes `content` and substitutes a synthesized `summary`, so the client stores and replays an item the issuer never sent:

```
{"code":"invalid-argument","error":"Could not decrypt the provided encrypted_content. Ensure the value is the unmodified encrypted_content from a previous response."}
```

**No route change is needed to reach this — it fires on the second turn of a fresh session.**

The rewrite is correct where it was designed and verified. Its own header records the premise: *"Codex echoes the reasoning item it received back into the next request's input. DeepSeek's Responses API accepts summary-shaped reasoning input items (verified live), so the rewrite round-trips."* DeepSeek is `statelessResponses` and issues no blob — its reasoning replay runs through the proxy-side cache instead, which is why the round trip held. Providers that *do* issue a blob joined the same route later through `preserveReasoningContentModels`, a flag whose own purpose is Chat-wire prompt-cache replay, and the verified premise did not follow them. `registry.ts` currently lists xAI, Kimi, GLM, NeuralWatt and others there.

The guard is therefore on the item, not on a provider list: any reasoning item carrying a non-empty `encrypted_content` is returned exactly as the upstream sent it. DeepSeek is unaffected by construction.

**Only the stored item is exempt.** The `response.reasoning_text.delta` / `.done` events carry no blob and still route to the summary channel, so the expandable trace Codex renders for the live turn is unchanged — this is not a rollback of #45.

## Scope

This is one of two root causes behind Grok breaking on the native Responses route. The sibling `Could not decode the compaction blob` failure has a different cause and is #2228; the two do not overlap in files.

What this PR deliberately does **not** touch:

- `stripItemIdsWhenUnstored` still strips the reasoning item's `id`. codex-rs strips ids from every item under `store: false` (`core/src/client.rs:918-925`) and OpenAI accepts that, so stripping is the contract rather than a corruption.
- `sanitizeReasoningInputContent` still blanks replayed `content` for providers without `preserveResponsesReasoningContent`. Sending raw reasoning text back upstream has token and privacy consequences, and `content: []` alongside the blob matches what codex-rs itself sends.
- Identity-scoped replay (model switch, account switch, combo rotation, `previous_response_id` expansion) is unguarded for native blobs. Those are real gaps but need per-conversation provenance state, which is a separate design.

If a live canary still reports the same error after this lands, the remaining suspects are that list — in that order.

## Verification

- **RED-first**: reverting only `src/` fails the 3 new integrity cases (`output_item.done`, `response.completed` output, and the non-streaming document rewrite).
- `bun test tests/responses-reasoning-summary-rewrite.test.ts` — 19 pass, 0 fail.
- `bun test` across the reasoning/deepseek/xai/responses/passthrough suites — 893 pass, 0 fail.
- Note: plain `bun test` (no runner) hangs on this tree with high CPU and no progress, matching the observation recorded in #2217 — use `bun run test`.

## Checklist

- [x] Docs or release notes were updated when needed. (`structure/04_transports-and-sidecars.md`, under Reasoning display parity.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. The change only stops the proxy from editing an upstream payload; no new persistence, logging, credential handling, or destination. Privacy scan green.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved encrypted reasoning content without modification across streamed and non-streamed responses.
  * Continued routing unencrypted reasoning text updates through the summary channel.
* **Tests**
  * Added coverage verifying encrypted payloads remain unchanged in completed responses and streaming events.
  * Confirmed unencrypted reasoning deltas continue using the summary channel.
* **Documentation**
  * Documented handling for encrypted reasoning content and text updates.

## Gate

This branch's change is comment/doc-only, so it was gated as part of the integrated series rather than in isolation. `bun run test` on the branch that stacks all of these fixes — 13773 pass, 10 skip, 1 fail across 867 files.

The single failure is `tests/key-login-live-update.test.ts` > "notify after key login pushes the merged row and keeps modelCosts on live and disk". It is **pre-existing and unrelated**: it reproduces byte-identically on every branch in this series, including ones that never touch CLI code. Every gate in this series lands on exactly that one failure.

### DIFF
--- a/src/server/responses-reasoning-summary-rewrite.ts
+++ b/src/server/responses-reasoning-summary-rewrite.ts
@@ -34,6 +34,14 @@ function reasoningTextOf(item: Record<string, unknown>): string {
 /** Move a reasoning item's content channel into the summary channel. */
 function reasoningItemToSummaryShape(item: Record<string, unknown>): Record<string, unknown> {
   if (item.type !== "reasoning") return item;
+  // An item carrying `encrypted_content` is replayed verbatim by the client, and the backend that
+  // issued the blob verifies the item it gets back ("Could not decrypt the provided
+  // encrypted_content. Ensure the value is the unmodified encrypted_content from a previous
+  // response."). Reshaping it here is a modification the client then replays on every later turn.
+  // The delta rewrite still gives Codex the expandable trace for the live turn; only the stored
+  // shape has to stay exactly as the upstream sent it. DeepSeek — the provider this rewrite was
+  // verified against — is `statelessResponses` and issues no blob, so it is unaffected.
+  if (typeof item.encrypted_content === "string" && item.encrypted_content.length > 0) return item;
   const text = reasoningTextOf(item);
   // Items that already use the summary channel (or carry no content text at
   // all) are left untouched: rewriting them could clear a valid summary.
--- a/src/server/responses-reasoning-summary-rewrite.ts
+++ b/src/server/responses-reasoning-summary-rewrite.ts
@@ -34,13 +34,12 @@ function reasoningTextOf(item: Record<string, unknown>): string {
 /** Move a reasoning item's content channel into the summary channel. */
 function reasoningItemToSummaryShape(item: Record<string, unknown>): Record<string, unknown> {
   if (item.type !== "reasoning") return item;
-  // An item carrying `encrypted_content` is replayed verbatim by the client, and the backend that
-  // issued the blob verifies the item it gets back ("Could not decrypt the provided
-  // encrypted_content. Ensure the value is the unmodified encrypted_content from a previous
-  // response."). Reshaping it here is a modification the client then replays on every later turn.
-  // The delta rewrite still gives Codex the expandable trace for the live turn; only the stored
-  // shape has to stay exactly as the upstream sent it. DeepSeek — the provider this rewrite was
-  // verified against — is `statelessResponses` and issues no blob, so it is unaffected.
+  // `encrypted_content` is opaque, state-bearing provider data, so the entire item must retain its
+  // upstream shape unless that backend has an explicit replay contract permitting a rewrite. This
+  // defensively protects content-channel backends that do issue blobs when the client replays the
+  // stored item. The delta rewrite can still provide the expandable trace for the live turn.
+  // DeepSeek — the provider this rewrite was verified against — is `statelessResponses` and issues
+  // no blob, so it is unaffected.
   if (typeof item.encrypted_content === "string" && item.encrypted_content.length > 0) return item;
   const text = reasoningTextOf(item);
   // Items that already use the summary channel (or carry no content text at
--- a/structure/04_transports-and-sidecars.md
+++ b/structure/04_transports-and-sidecars.md
@@ -798,13 +798,14 @@ Codex app, so tool cells group like native models — while the text still round
 `devlog/_fin/260709_native_response_pattern/`.
 
 The content-to-summary channel rewrite skips any reasoning item that carries a native
-`encrypted_content` blob. Codex replays the reasoning item it received, and a backend that issued
-that blob verifies what comes back, so reshaping the stored item makes every later turn fail with
-`Could not decrypt the provided encrypted_content`. The rewrite's round trip was verified against
-DeepSeek, which is `statelessResponses` and issues no blob; providers that do issue one joined the
-same route later through `preserveReasoningContentModels`. Only the stored item is exempt — the
-`reasoning_text` delta events carry no blob and still route to the summary channel, so the live
-expandable trace is unchanged.
+`encrypted_content` blob. The blob is opaque, state-bearing provider data, so the item must
+round-trip unchanged unless that backend has an explicit replay contract permitting a rewrite.
+This defensively protects providers that issue blobs and later join the route through
+`preserveReasoningContentModels`. The rewrite's round trip was verified against DeepSeek, which is
+`statelessResponses` and issues no blob. Grok is unaffected in practice because it natively emits
+summary-channel reasoning and no `reasoning_text` events, so this content-to-summary item rewrite
+does not engage on its route. Only the stored item is exempt — `reasoning_text` delta events carry
+no blob and still route to the summary channel, so the live expandable trace is unchanged.
 
 The process-local raw-reasoning fallback is fail-closed unless a request has an explicit client
 thread plus an exact provider destination, wire adapter, final model, and physical credential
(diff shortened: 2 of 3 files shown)


lidge-jun:
## 리뷰 · 우선순위 74 / 80

구멍은 응답 레그임. 지금 `dev` `src/server/responses-reasoning-summary-rewrite.ts` `reasoningItemToSummaryShape` (`:35-44`)가 `type === "reasoning"`이면 `content`를 지우고 합성 `summary`를 넣음. Codex가 받은 아이템을 다음 `input`에 그대로 에코함. blob을 민팅한 백엔드가 unmodified 검사함. 리셰이프된 아이템이 오면 `Could not decrypt the provided encrypted_content` 남. 루트 바꿀 필요 없음. 새 세션 두 번째 턴에서 터짐.

주석이 DeepSeek 전제임 (`:17-19`). DeepSeek는 `statelessResponses`고 blob을 안 냄. 프록시 캐시로 리플레이함. 나중에 `preserveReasoningContentModels`로 blob을 내는 프로바이더가 같은 리라이트 루트에 붙었음. 지금 `dev` `src/providers/registry.ts` xAI가 grok-4.5/4.6 OAuth+inbound responses를 `openai-responses`로 박음 (`:1030-1042`) + `preserveReasoningContentModels` (`:1062`). 기본 OAuth Grok Responses가 이 리라이트를 탐. #45 트레이스 리라이트가 blob 아이템까지 삼킨 거임.

이 PR이 가드를 아이템에 둠. `encrypted_content` 문자열이 비어 있지 않으면 바이트 그대로 반환. 델타 `response.reasoning_text.delta` / `.done`은 blob이 없어서 요약 채널로 그대로 감. 라이브 확장 트레이스는 유지. DeepSeek는 blob이 없어서 영향 없음. `stripItemIdsWhenUnstored`는 그대로 id를 뗌. `sanitizeReasoningInputContent`는 `preserveResponsesReasoningContent` 없는 쪽에서 리플레이 `content`를 비움. 방향 맞음.

테스트가 `output_item.done` / `response.completed` / non-streaming document에서 blob 아이템 byte-for-byte, 델타 리라이트는 살아 있는 걸 잠금. RED-first라고 함. `src/`만 되돌리면 새 3케이스가 깨짐. ㅋㅋ 한 가드가 세 경로를 같이 막음.

남은 구멍은 본문이 인정함. 모델 스위치/계정 스위치/콤보 로테이션/`previous_response_id` 확장은 native blob 프로비넌스가 없음. 이 PR에서 풀 설계 넣지 말 것. 카나리 후에도 같은 에러면 그 순서임.

#2228이랑 원인 다름. 저건 컴팩션 blob 목적지. 이건 reasoning 아이템 리셰이프. `structure/04_transports-and-sidecars.md`만 겹침. #2217/#2227이랑 `modelWireDefaults.wire` 같이 소유하면 안 됨. 이 PR은 그 맵 안 건드림. Chat 기본(#2227)이 돼도 Responses 옵트인에서 같은 리셰이프가 남음. 닫지 말 것. types.ts 스플릿 안 씹힘. rewrite 모듈이랑 테스트/SOT만임. #2188 사이드카, #2190 `x_search`랑 섞지 말 것.

draft고 체크리스트 0/4. hygiene 통과. 지금 HEAD `03735eca6`에서 기본 Grok OAuth 두 번째 턴이 죽을 수 있어서 점수 높음. 2.28 블로커는 아님.

**해결방안:** CI 그린이면 #2228이랑 같이 `dev` 머지. 기본 와이어 싸움은 #2217/#2227에서. 프로비넌스 태그는 후속. 스플릿이 rewrite 모듈을 옮기면 리베이스하지 말고 닫고 다시 짜라. 지금은 그 정도 아님.

이 댓글은 grok-bot이 작성했습니다

olddonkey:
**Correction: the motivating claim in the description does not hold for Grok.** I deployed this branch locally and probed the live route, and the result contradicts what I wrote.

Grok emits **summary-channel** reasoning natively. A streaming turn produces `response.reasoning_summary_part.added` / `response.reasoning_summary_text.delta` / `.done` — no `reasoning_text` events at all — and the completed item carries `encrypted_content` plus `summary` with no `content`:

```
turn1 event types: {'response.reasoning_summary_part.added': 1,
                    'response.reasoning_summary_text.delta': 19,
                    'response.reasoning_summary_text.done': 1,
                    'response.reasoning_summary_part.done': 1}
reasoning item keys: ['encrypted_content', 'id', 'status', 'summary', 'type']  blob len: 452
```

`reasoningItemToSummaryShape` returns early when there is no `reasoning_text` content (`if (text.length === 0) return item;`), so on this route the rewrite this PR guards **never fired in the first place**. It is listed in `preserveReasoningContentModels`, which arms `routeUsesContentChannelReasoning`, but the rewrite is inert because the shape it converts never arrives.

Replaying the reasoning item verbatim works today, with the blob intact, on both wire modes:

- non-streaming, same model — `completed`
- streaming, same model — `completed`
- `grok-4.5`-minted blob replayed to `grok-4.6` — `completed`

So this branch does **not** fix `Could not decrypt the provided encrypted_content`, and cross-model switching inside the Grok family is not the trigger either. I should not have asserted a cause I had only traced statically.

**What the change is still worth.** The invariant is real and independently justified: an item carrying an integrity-checked blob must not be reshaped, and the rewrite's own header records that its replay round trip was verified against DeepSeek — which is `statelessResponses` and issues no blob. `preserveReasoningContentModels` now also lists Kimi, GLM and NeuralWatt routes; any of them that emits content-channel reasoning *and* a blob would hit exactly the failure this guards, and nothing else in the chain would catch it. I would keep it as hardening, retitled honestly, rather than as a fix for the Grok symptom.

**Where the Grok symptom actually points.** The remaining suspect is a blob minted by a different backend and replayed to Grok — the same provenance gap #2228 closes for compaction blobs, which has no counterpart for `encrypted_content`. The reporter's history was heavy `openai/gpt-5.6-sol` traffic before switching to Grok, which fits. I could not mint an OpenAI blob through the forward route to prove it end to end, so that stays a hypothesis rather than a diagnosis.

olddonkey:
Part of #2240 — hardening: do not reshape a reasoning item carrying opaque provider state.

Ingwannu:
Maintainer status on exact head 3d11f6fddd5039900feb52260ae2cb684de60a65: the encrypted_content preservation guard remains a valid hardening layer, but the PR is still draft with a 0/4 readiness checklist and is 24 commits behind dev. Please synchronize with current dev, complete the exact-head CI/readiness gate, and then request final review. This should not close #2240 by itself.

olddonkey:

PR #2237, opened 2026-08-21
fix(responses): drop a null reasoning content channel before routed passthrough

olddonkey:
## Summary

Codex serializes an absent reasoning content channel as `"content": null`. `sanitizeReasoningInputContent` only acted on a non-empty array, so the null reached the wire verbatim, and xAI rejects the item while naming the sibling field:

```
{"code":"invalid-argument",
 "error":"Could not decode the compaction blob. Ensure it is unmodified from the compact response."}
```

**The blob is not the problem.** That message is why this went undiagnosed: it points at `encrypted_content` (which xAI calls a "compaction blob") and at compaction state, when the field it actually refused is `content`.

This bites **the second turn of every Grok conversation** — the first request that replays a reasoning item. A fresh session fails exactly as reliably as a resumed one.

## Evidence

Captured a live failing request and bisected against it:

| variant | result |
|---|---|
| body verbatim | 400 `Could not decode the compaction blob` |
| `encrypted_content` removed | 400 schema — the blob is required |
| whole reasoning item removed | 200 |
| **only the `content` key removed** | **200** |
| `content` set to `[]` | 200 |

The captured value was `"content": null`.

The proxy was independently cleared of corrupting the blob: instrumenting both directions showed the value grok streamed to the client and the value replayed upstream matched in length, prefix and suffix (`len=2099`, `ZnXTtn+ABaJz5yzPf0uS6SKzXNpP`), under identical `x-grok-conv-id`, `x-grok-session-id`, account token and URL. Also ruled out by direct test: blob size (a 34327-char blob replays fine), item shape (missing `id`/`status`, the private `internal_chat_message_metadata_passthrough`), replayed `function_call`/`function_call_output` pairs, cross-backend blobs, `prompt_cache_key` drift, tool-catalog drift, and SSE event inconsistency.

After the fix, the exact captured request returns `completed`.

## Scope

The field is optional and null carries nothing, so the key is dropped rather than rewritten — provably lossless. An array content channel still follows the existing rules, including the `preserveResponsesReasoningContent` carve-out for DeepSeek.

Sibling `ocxr1:` envelope handling is preserved on this path: a proxy-minted envelope is still stripped when the null key is dropped.

## Verification

- **RED-first**: reverting only `src/` fails the new `drops a null content channel while keeping the replayable blob` case.
- 887 pass, 0 fail across the responses/adapter/parser/xai/reasoning suites.
- `bun run typecheck` clean; `bun run privacy:scan` passed.
- Live: replaying the captured failing body against a patched local build returns `completed`.

## Relation to the other Grok PRs

Independent of #2217, #2228 and #2229 — different file region, no overlap. Worth noting that #2229's original claim was wrong and I have corrected it there: Grok emits summary-channel reasoning natively, so the rewrite that PR guards never fires on this route, and it is hardening rather than a fix for this symptom. **This** is the change that makes Grok usable past the first turn.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of reasoning data in routed Responses requests.
  * Removed unsupported reasoning content fields for compatible destinations while preserving encrypted content and summaries.
  * Preserved existing behavior for official OpenAI destinations.
  * Ensured array-based reasoning content continues to be processed correctly.

* **Tests**
  * Added coverage for reasoning-item handling across supported destination types, including null, encrypted, summary, and array content.

## Gate

Gated as part of the integrated series (this change is also exercised live — see the verification table above). `bun run test` on the branch that stacks all of these fixes — 13773 pass, 10 skip, 1 fail across 867 files.

The single failure is `tests/key-login-live-update.test.ts` > "notify after key login pushes the merged row and keeps modelCosts on live and disk". It is **pre-existing and unrelated**: it reproduces byte-identically on every branch in this series, including ones that never touch CLI code. Every gate in this series lands on exactly that one failure.

### DIFF
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -6,7 +6,7 @@ import { COMPACT_PROMPT, decodeCompactionSummary, SUMMARY_PREFIX } from "../resp
 import { collectResponsesToolGroups } from "../responses/tool-groups";
 import { isHostedToolUnsupportedForModel } from "../responses/hosted-tool-policy";
 import { decodeServerSentEvents } from "../lib/sse-decoder";
-import { CODEX_FORWARD_BASE_URL, isCanonicalOpenAiForwardProvider } from "../providers/openai-tiers";
+import { CODEX_FORWARD_BASE_URL, isCanonicalOpenAiForwardProvider, isOpenAiOperatedResponsesDestination } from "../providers/openai-tiers";
 import { OCX_REASONING_PREFIX } from "../responses/reasoning-envelope";
 import { modelRecordValue } from "../reasoning-effort";
 import type { TranslatorBudget } from "../lib/translator-budget";
@@ -41,7 +41,7 @@ export const FORWARD_HEADERS = [
 
 export function sanitizeReasoningInputContent(
   body: unknown,
-  opts?: { preserveRawReasoningContent?: boolean },
+  opts?: { preserveRawReasoningContent?: boolean; dropNullContentChannel?: boolean },
 ): unknown {
   if (!body || typeof body !== "object" || Array.isArray(body)) return body;
   const raw = body as Record<string, unknown>;
@@ -61,7 +61,11 @@ export function sanitizeReasoningInputContent(
     // — xAI answers `Could not decode the compaction blob`, naming the sibling `encrypted_content`
     // rather than the field it actually refused, which is why this reads as a blob failure. Drop the
     // key so the item matches the shape the upstream issued.
-    if ("content" in rec && !Array.isArray(rec.content)) {
+    //
+    // Gated to routed destinations. An OpenAI-operated backend binds the blob to the item's exact
+    // shape, so deleting a field there invalidates it (`The encrypted content ... could not be
+    // verified`); the two requirements are exactly opposed, and a live regression proved it.
+    if (opts?.dropNullContentChannel === true && "content" in rec && !Array.isArray(rec.content)) {
       changed = true;
       const next: Record<string, unknown> = { ...rec };
       delete next.content;
@@ -1584,7 +1588,10 @@ export function createResponsesPassthroughAdapter(provider: OcxProviderConfig): 
         outBody = rewritten.body;
         convertedRoutedToolSearchNames = rewritten.names;
       }
-      const sanitizedBody = normalizeToolSchemas(stripSparkCompatibility(stripUnsupportedReasoningParams(stripItemIdsWhenUnstored(stripInvalidItemIds(stripUnsupportedHostedTools(sanitizeReasoningInputContent(scrubOcxCompactionItems(outBody), { preserveRawReasoningContent: provider.preserveResponsesReasoningContent === true })))))));
+      const sanitizedBody = normalizeToolSchemas(stripSparkCompatibility(stripUnsupportedReasoningParams(stripItemIdsWhenUnstored(stripInvalidItemIds(stripUnsupportedHostedTools(sanitizeReasoningInputContent(scrubOcxCompactionItems(outBody), {
+        preserveRawReasoningContent: provider.preserveResponsesReasoningContent === true,
+        dropNullContentChannel: !isOpenAiOperatedResponsesDestination(provider),
+      })))))));
       const finalBody = stripDisabledReasoningSummaries(
         normalizeConfiguredReasoningSummaryDelivery(sanitizedBody, provider, parsed.modelId),
         provider,
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -56,6 +56,18 @@ export function sanitizeReasoningInputContent(
     // ocxr1 envelopes are proxy-minted (Anthropic signatures), not OpenAI encryption — the native
     // backend cannot decrypt them and would reject the request. Strip regardless of content shape.
     const hasOcxEnvelope = typeof rec.encrypted_content === "string" && rec.encrypted_content.startsWith(OCX_REASONING_PREFIX);
+    // Codex serializes an absent reasoning content channel as `"content": null`. The field is
+    // optional and null carries nothing, but a strict gateway rejects the item on its declared type
+    // — xAI answers `Could not decode the compaction blob`, naming the sibling `encrypted_content`
+    // rather than the field it actually refused, which is why this reads as a blob failure. Drop the
+    // key so the item matches the shape the upstream issued.
+    if ("content" in rec && !Array.isArray(rec.content)) {
+      changed = true;
+      const next: Record<string, unknown> = { ...rec };
+      delete next.content;
+      if (hasOcxEnvelope) delete next.encrypted_content;
+      return next;
+    }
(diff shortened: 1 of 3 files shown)


Ingwannu:
Draft review note on exact head `d8426a6a5ada`.

The live symptom and the `content: null` fix are credible, and the focused Responses/DeepSeek suites pass locally (85/85), along with typecheck and the privacy scan. One scope blocker remains before this is ready for review:

`sanitizeReasoningInputContent` currently removes `content` for **every** non-array value:

```ts
if ("content" in rec && !Array.isArray(rec.content))
```

That includes strings, numbers, booleans, and objects, not only the `null` shape established by the capture. Those malformed values currently remain visible to the upstream validation boundary; silently deleting them can turn an invalid caller item into a materially different valid request. It also exceeds the PR's lossless rationale, which applies specifically to `null`.

Please narrow this branch to `rec.content === null` (while preserving the existing `ocxr1` stripping behavior), and add regressions proving representative non-null malformed values are not silently normalized away. The existing array and DeepSeek preservation behavior should remain unchanged.

Once that scope is narrowed and the draft readiness gates are complete, this remains a strong merge candidate.

olddonkey:
Pushed `6e86b181b`. The first version of this fix was **wrong in one direction**, and live traffic caught it minutes after I deployed it locally:

```
400 invalid_request_error   provider=openai model=gpt-5.6-sol
The encrypted content k7pQ...Px7D could not be verified.
Reason: Encrypted content could not be decrypted or parsed.
```

An OpenAI-operated backend binds the blob to the item's **exact shape**, so deleting a field invalidates it. xAI refuses the null key; OpenAI needs it kept. The two requirements are exactly opposed, so the strip has to follow the destination rather than apply everywhere.

The gate is `isOpenAiOperatedResponsesDestination`: canonical ChatGPT Codex, or the exact official OpenAI API. It is deliberately **not** `authMode === "forward"` — a noncanonical forward provider never receives the caller's credentials, so forward auth says nothing about which backend answers, and such a relay is routed like any other gateway.

Two regression tests come with it: OpenAI-operated destinations keep the null key, a noncanonical forward relay gets it stripped.

Worth recording for reviewers: an independent Codex review of the previous head returned **SHIP** for this PR and did not surface this regression, while a single real request did. Static review of a diff whose two sides want opposite things is not enough here — the OpenAI path needs a live check, not just a green suite.

### Verification

- RED-first against the previous commit: `keeps a null content channel on OpenAI-operated destinations` fails there and passes now.
- 827 pass / 0 fail across the responses/adapter/parser/reasoning suites; `bun run typecheck` clean; `bun run privacy:scan` passed.
- Live on a patched local build: Grok's captured failing body returns `completed`, and an OpenAI turn plus a `content: null` replay both return `completed`.

Ingwannu:
Thanks for adding the destination boundary; preserving the exact item shape on OpenAI-operated backends is important, and the new live regression evidence explains why that gate is needed.

The original scope blocker is still present on exact head 6e86b181b3236a016c32820d540995675f541fd2, though. sanitizeReasoningInputContent still uses:

    "content" in rec && !Array.isArray(rec.content)

So routed destinations still silently delete every non-array value: strings, numbers, booleans, and objects, not only the captured null shape. The new tests cover null and arrays, but do not prove representative non-null malformed values remain visible to the upstream validation boundary.

Please keep the new OpenAI-operated destination gate, narrow the deletion condition to rec.content === null, and add a parameterized regression for non-null malformed values. After that and the normal draft readiness gates, this remains a strong merge candidate.

olddonkey:
Part of #2240 — the `\"content\": null` reasoning item — the second-turn 400.

lidge-jun:
## 리뷰 · 우선순위 75 / 80

구멍은 두 번째 턴임. 지금 `dev` `sanitizeReasoningInputContent` (`src/adapters/openai-responses.ts:42-80`)가 `Array.isArray(rec.content) && rec.content.length > 0`만 봄. Codex가 빈 채널을 `"content": null`로 직렬화하면 `hasRawContent`가 false. ocxr1도 아니면 `:59`에서 아이템 그대로 통과. xAI가 거절하면서 형제 필드 `encrypted_content`를 compaction blob이라고 욕함. blob은 멀쩡함. 라이브 비스ect가 잠금. 키만 지우면 200, blob 지우면 스키마 400. 첫 턴 카탈로그(#2217) 다음에 바로 터짐. ㅋㅋ 에러 메시지가 거짓말하는 케이스.

이 PR이 `dropNullContentChannel`을 넣음. 라우티드만. 신규 `isOpenAiOperatedResponsesDestination`이 캐논 ChatGPT forward랑 `api.openai.com/v1`만 OpenAI 운영으로 봄. 그쪽은 키 유지. 나머지는 드롭. 언게이트 첫 버전이 OpenAI에서 blob 검증을 깨먹은 거 라이브로 잡았음. `authMode === "forward"`로 안 가른 거 맞음. 비캐논 포워드 릴레이는 콜러 자격을 안 받아서 백엔드가 뭔지 모름. 테스트가 xAI 드롭 / OpenAI 유지 / 비캐논 릴레이 드롭 / 배열 채널은 기존 새니타이저. 패스스루 체인 (`:1575`)에 게이트를 꽂은 위치도 맞음.

남은 블록은 Ingwannu가 두 번 말한 그거임. 조건이 `"content" in rec && !Array.isArray(rec.content)`라 문자열/숫자/객체도 조용히 지움. 본문 근거는 null만 lossless. 그 범위를 넘음. 잘못된 아이템을 유효한 요청으로 바꿔 버림. `rec.content === null`로 좁히고, 비널 malformed가 업스트림 검증에 그대로 보이게 회귀 넣을 것. ocxr1 스트립이랑 DeepSeek `preserveRawReasoningContent`는 유지.

#2229랑 원인 다름. 저자가 Grok 네이티브가 summary 채널을 내서 그 리라이트가 이 루트에서 안 탄다고 정정함. 그건 하드닝. 이게 두 번째 턴 본체임. #2217/#2227 `modelWireDefaults.wire` 안 건드림. 싸우지 말 것. #2228 컴팩션 목적지랑도 다른 선임. types.ts/config.ts 스플릿 안 씹힘. 어댑터 새니타이저랑 tiers 헬퍼임. #2188 사이드카, #2190 `x_search`랑 섞지 말 것. 닫을 중복 아님.

draft고 체크리스트 0/4. hygiene 통과. 지금 HEAD `78f1942a0`가 여전히 grok-4.5/4.6 OAuth+responses를 `openai-responses`로 박음 (`src/providers/registry.ts:1030-1042`). 기본 경로 두 번째 턴이 죽음. 2.28 태그 블로커는 아닌데 사용자 체감은 블로커임. #2240 트래킹. 클로징 키워드 넣지 말 것.

**해결방안:** `rec.content === null`로 좁히고 malformed 회귀 넣고 체크리스트 채운 다음 `dev` 머지. OpenAI 운영 게이트 유지. #2217이랑 같이 넣어도 됨. 스플릿이 이 새니타이저를 옮기면 리베이스하지 말고 닫고 다시 짜라. 지금은 그 정도 아님.

이 댓글은 grok-bot이 작성했습니다

olddonkey:

Issue #2240, opened 2026-08-21
[Bug]: xAI Grok unusable on native Responses in 2.28.0


olddonkey:
### Client or integration

Codex CLI

### Area

Provider adapter

### Summary

In OpenCodex 2.28.0, OAuth `xai/grok-4.5` and `xai/grok-4.6` became unusable on the native Responses route. #2147 moved those models off the Chat Completions compatibility wire onto `openai-responses` passthrough so Codex would not sit blank until reasoning finished (#1886). Chat translation had been sanitizing Codex-private wire extensions as a side effect; passthrough forwards the client's bytes, and xAI rejects the request before inference.

The conversation-killing case is the second turn: Codex replays a reasoning item with `"content": null`, and xAI returns HTTP 400 while mislabeling the field as a compaction blob. A first-turn namespace / private-tool catalog can 422 the same session before that.

I have already handled this (fix PRs #2217 and #2237). Please assign the issue to @olddonkey.

### Reproduction

1. Run OpenCodex 2.28.0 with Codex CLI on OAuth `xai/grok-4.5` or `xai/grok-4.6` (native `/v1/responses`, no Chat Completions wire).
2. Start a new conversation with the default Codex 0.147+ tool catalog.
3. First message is rejected (private `tools[].type = "namespace"` / `external_web_access`).
4. After the catalog is lowered, send a second message so Codex replays the previous reasoning item with `"content": null`.

Redacted second-request shape:

```json
{
  "model": "grok-4.6",
  "input": [
    {"role": "user", "content": "hello"},
    {
      "type": "reasoning",
      "id": "rs_redacted",
      "encrypted_content": "<intact grok blob>",
      "content": null
    },
    {"role": "user", "content": "continue"}
  ]
}
```

On a captured failing body: verbatim request → 400; removing only the `content` key → 200; removing `encrypted_content` → 400 schema error (blob was intact). Stripping `content: null` unconditionally repairs Grok and breaks OpenAI, which requires that field.

### Version

2.28.0

### Operating system

macOS

### Provider and model

xai / grok-4.5, xai / grok-4.6

### Logs or error output

```text
HTTP 422
```

```text
Argument not supported: external_web_access
```

```json
{"code":"invalid-argument","error":"Could not decode the compaction blob. Ensure it is unmodified from the compact response."}
```

The failing request often has no compaction item. xAI names the sibling field, not the null `content` channel that actually caused the 400.

### Redacted configuration

```json
{
[...]

PR #2241, opened 2026-08-21
fix(xai): translate Codex namespace tools for Responses

kakaholigan:
## Summary
- flatten Codex 0.148 `namespace` tool containers only for xAI Responses transports
- preserve every callable child and restore non-builtin namespaces on JSON/SSE tool calls
- fail closed on malformed, unsupported, duplicate, or colliding namespace catalogs
- keep non-xAI Responses requests byte-compatible

## Verification
- `bun test tests/xai-responses-namespace-tools.test.ts` (6 pass)
- related xAI/Responses suites (86 pass)
- `bun run typecheck`
- `bun run privacy:scan`

The full repository suite was also started; the host run reports pre-existing environment failures in GUI tests because React dependencies are not installed and Bun 1.3.14 `EEXIST: epoll_ctl` test-runner errors. The focused and adjacent suites are green.

## Review readiness checklist

### DIFF
--- a/src/adapters/base.ts
+++ b/src/adapters/base.ts
@@ -72,6 +72,8 @@ export interface AdapterRequest {
     convertedRoutedCustomToolNames?: ReadonlySet<string>;
     /** Client tool-search names actually lowered to upstream function calls for this request. */
     convertedRoutedToolSearchNames?: ReadonlySet<string>;
+    /** xAI-safe flat aliases created from Codex namespace tool containers. */
+    convertedNamespaceToolAliases?: ReadonlyMap<string, { namespace: string; name: string }>;
     /** Releases observation of a serialized request body after its final fetch attempt settles. */
     releaseBodyObservation?: () => void;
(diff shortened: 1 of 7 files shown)


Ingwannu:
Thanks for working on the Grok Responses regression. I am closing this as superseded by #2217 rather than asking both branches to evolve in parallel.

The two PRs own the same namespace-tool translation boundary and would conflict. #2217 is already the reviewed sequential fix for the first-turn failure in #2240, has exact-head green CI, and covers the full observed request shape: namespace declaration/selector/replay/SSE restoration plus removal of the Codex-private external_web_access and defer_loading fields. This draft currently handles the namespace container but does not remove those additional fields, so it would not by itself repair the captured first-turn request.

The xAI-specific transport classifier here is a reasonable design idea, but it is not enough reason to maintain a second incompatible implementation. Please move any concrete edge case that #2217 does not cover into that PR or #2240 instead.

PR #2242, opened 2026-08-21
feat(web-search): live xai executor with opt-in x_search (#2188 L7)

lidge-jun:
## Summary

Layer 7 of the #2188 follow-up chain (parent: #2238). The xAI web-search executor goes live:

- `src/web-search/xai-executor.ts`: `runXaiWebSearch` POSTs the EXACT-origin-pinned `api.x.ai` Responses endpoint (parsed `url.origin` comparison — a prefix check admitted lookalike hosts, review-caught Critical) with the stored Grok OAuth credential, hosted `web_search` + opt-in `x_search` (doc-validated: ≤20 handles, allow XOR exclude, ISO dates), `include: web_search_call.action.sources`, and `reasoning.effort` (probe-verified). `redirect: "manual"`, never-throws, 401/403 entitlement-distinct errors, byte-bounded SSE reducer.
- Sources = `url_citation` annotations ∪ `web_search_call.action.sources`, deduped; `custom_tool_call` items (live x_search reality) and skeleton `action` tolerated per the devlog 003 captures.
- `planWebSearch`'s xai arm goes live fail-closed (no credential → no plan; invalid persisted xSearch → no plan); `SidecarPlan.xaiSidecar` → core.ts handoff → loop arm that FAILS CLOSED when the sidecar is absent (review-caught High: the fallthrough would have reached the forward-header executor). No Codex pool outcome recording on this arm.
- `xSearch` config block: PUT-validated (400 on doc-limit violations), GET/PUT round-tripped (review-caught High), null clears.
- Registry activates the backend on stored-OAuth presence; grok catalog rows become web-search candidates.

Live E2E through the real executor: web_search → 130 chars + 12 sources; x_search with `allowed_x_handles: ["xai"]` → 492 chars.

Design doc: `devlog/_plan/260820_sidecar_selection_unification/070_layer7_xai_executor.md` (rev 2) + 003 research.

## Verification

- Remote full gate at exact head (lidge): **13792 pass / 15 skip / 0 fail** across 873 files (491s); typecheck clean.
- 12 focused tests incl. lookalike-origin bearer containment, loop-level fail-closed regression (zero fetches, zero pool recordings, invariant error replayed), SSE fixtures from live captures, xSearch round-trip/400.
- Independent read-only review (terra): 1 Critical + 2 High found and fixed, follow-up demanded loop regression added — final VERDICT: pass.

## Checklist

- [x] Full typecheck + full suite green at head (remote, exact SHA)
- [x] Live E2E through the shipped executor
[...]

Issue #2233, opened 2026-08-20
[Bug] Google adapter: non-conforming part fields cross the AdapterEvent boundary (nameless tool calls, non-string text)


snowyukitty:
### Client or integration

Not client-specific — reproduced by driving the adapter directly with a synthetic `Response`.

### Area

Provider adapter · Streaming / SSE parsing · Non-streaming response parsing

### Summary

This is the piece I deliberately kept out of #2231 / #2232, filed on its own because the fix shape
is a policy call rather than a parity fix, and I would rather leave that call to you than guess it.

#2232 validates the *containers* — `candidates`, `content`, `content.parts`. Inside a well-formed
part, the values the adapter actually emits are still unchecked, so three shapes cross the
`AdapterEvent` boundary as something other than what the type declares:

1. **A `functionCall` that is not a record** (`"x"`, `5`, `[]`) reaches `part.functionCall.name` as
   `undefined` and the adapter emits a `tool_call_start` **with no name**.
2. **A `functionCall.name` that is absent, non-string, empty, or whitespace** is passed through
   verbatim — `name: undefined`, `name: 5`, `name: ""`, `name: "   "` all become a dispatched tool
   call.
3. **A non-string `part.text`** becomes `text_delta.text` as a number, object or array — and with
   `thought: true`, `reasoning_raw_delta.text`.

`AdapterEvent` declares both `text` and `name` as `string`. `src/adapters/openai-chat.ts` now
rejects the equivalent shapes (`diagnoseInvalidToolCalls` reasons
`tool_call_function_name_invalid` / `_blank`, plus `unnamedToolCallEvent`, whose message is
explicit: "upstream streamed a tool call without a function name — cannot dispatch"). The Google
adapter has no equivalent.

**Why this is not simply "port the openai-chat rule".** OpenAI Chat *assembles* a tool call across
streamed deltas, so "the name never arrived by flush time" is a well-defined terminal condition and
`tests/openai-chat-parallel-stream.test.ts` had to be re-negotiated around it. Google delivers a
tool call **whole, in one part** — there is no assembly window and no "it may still arrive". That
makes the rejection *easier* to justify here, but it also means the openai-chat precedent is not
automatically the right disposition, and a nameless call could plausibly be dropped rather than made
terminal. Which of terminate / drop / pass-through you want is your call, so there is no PR.

Found by audit, not from a live capture — same bar as #1325 and #2231. Verified on `dev@03735eca`.

[...]

PR #2243, opened 2026-08-21
feat(web-search): live gemini executor on the Antigravity CCA transport (#2188 L8)

lidge-jun:
## Summary

Layer 8 of the #2188 follow-up chain (parent: #2242). The Gemini web-search executor goes live on the Antigravity CCA transport:

- `src/web-search/gemini-executor.ts`: `runGeminiWebSearch` sends the CCA envelope — REGISTRY-pinned destination (a config-level baseUrl is never trusted for OAuth transmission, same rule as `src/server/images.ts`), IDE-fingerprint User-Agent (a plain UA gets 404, probe-verified), discovered `projectId`, `tools: [{google_search:{}}]`, and the effort-mapped `thinkingConfig.thinkingLevel` via `resolveAntigravityEffortWireModel`. `redirect: "manual"`, never-throws, all error branches redacted.
- `mapCcaGroundedResponse` maps `candidates[0]` text + `groundingMetadata.groundingChunks[].web` to the sidecar outcome, deduped by uri.
- `planWebSearch`'s gemini arm goes live fail-closed (OAuth or projectId absent → no plan); the loop arm fails closed without a resolved provider (regression-pinned: zero fetches, zero pool recordings); registry activates the backend on the same predicate.
- Imports stay lean (`google-antigravity-wire` + `antigravity-models` only — no `google.ts` adapter drag, reviewer-verified).

Live E2E through the real executor: grounded answer 433 chars + 2 grounding sources, with automatic OAuth token rotation mid-call.

Design doc: `devlog/_plan/260820_sidecar_selection_unification/080_layer8_gemini_executor.md` (rev 2) + 002 LIVE PROBE.

## Verification

- Remote full gate at exact head (lidge): **13801 pass / 15 skip / 0 fail** across 874 files (493s); typecheck clean.
- 9 focused tests: request-shape pin (registry destination under a malicious baseUrl, manual redirect, bearer + IDE UA, full envelope, `gemini-3.7-flash-tiered` + `thinkingLevel: "low"`), grounding-map fixtures from the live capture, fail-closed matrix, loop-level fail-closed regression.
- Independent read-only review (terra): 3 P1 (unredacted non-Error branch — fixed in both new executors, missing request-shape test, missing loop regression) + 1 follow-up (thinkingLevel assertion) all fixed — final VERDICT: pass.

## Checklist

- [x] Full typecheck + full suite green at head (remote, exact SHA)
- [x] Live E2E through the shipped executor
- [x] OAuth destination pinning test-pinned
- [x] Stacked on #2242; retarget after parent lands

### DIFF
--- a/src/server/responses/core.ts
+++ b/src/server/responses/core.ts
@@ -3587,6 +3587,7 @@ async function handleResponsesInner(
       forwardProvider: wsPlan.forwardSidecar?.provider,
       anthropicSidecar: wsPlan.anthropicSidecar,
       xaiSidecar: wsPlan.xaiSidecar,
+      geminiSidecar: wsPlan.geminiSidecar,
       xaiSearchOptions: wsPlan.xaiSearchOptions,
       hostedTool: wsPlan.hostedTool,
       selectedForwardHeaders: wsPlan.forwardSidecar?.headers ?? selectedForwardHeaders,
--- a/src/web-search/backends.ts
+++ b/src/web-search/backends.ts
@@ -66,6 +66,19 @@ export const WEB_SEARCH_BACKENDS: readonly WebSearchBackendDescriptor[] = [
     },
     eligibleModel: candidate => candidate.provider === "xai",
   },
+  {
+    backend: "gemini",
+    // Probe = usable Antigravity OAuth + discovered projectId (findGeminiSidecarProvider's predicate).
+    isActive: (_auth, config) => {
+      const provider = config.providers["google-antigravity"];
+      if (!provider || provider.disabled === true || provider.authMode !== "oauth") return false;
+      const set = getAccountSet("google-antigravity");
+      const active = set?.accounts.find(account => account.id === set.activeAccountId);
+      if (!active || active.needsReauth === true) return false;
+      return !!(active.credential as { projectId?: string } | undefined)?.projectId;
+    },
+    eligibleModel: candidate => candidate.provider === "google-antigravity",
+  },
 ];
 
 /**
--- a/src/web-search/gemini-executor.ts
+++ b/src/web-search/gemini-executor.ts
@@ -91,7 +91,7 @@ export async function runGeminiWebSearch(
   } catch (e) {
     const kind = e instanceof Error && e.name === "TimeoutError" ? "timeout" : "connect_error";
     console.warn(`[web-search] gemini sidecar ${kind} (${Date.now() - t0}ms)`);
-    return { text: "", sources: [], error: e instanceof Error ? redactSecretString(e.message) : String(e) };
+    return { text: "", sources: [], error: redactSecretString(e instanceof Error ? e.message : String(e)) };
   } finally {
     sidecarExit();
     linkedSignal.cleanup();
--- a/src/web-search/gemini-executor.ts
+++ b/src/web-search/gemini-executor.ts
@@ -10,15 +10,15 @@
  */
 import type { OcxProviderConfig } from "../types";
 import { getValidAccessTokenSnapshot, publicOAuthAuthenticationErrorMessage } from "../oauth";
-import { getAccountSet } from "../oauth/store";
 import { fetchWithResetRetry } from "../lib/upstream-retry";
-import { signalWithTimeout } from "../lib/abort";
+import { cancelBodyOnAbort, signalWithTimeout } from "../lib/abort";
+import { readBoundedResponseBytes } from "../lib/bounded-body";
 import { sidecarEnter } from "../lib/sidecar-tracker";
 import { redactSecretString } from "../lib/redact";
 import { ANTIGRAVITY_REQUEST_UA } from "../adapters/google-antigravity-wire";
 import { resolveAntigravityEffortWireModel } from "../providers/antigravity-models";
 import { getProviderRegistryEntry } from "../providers/registry";
-import type { WebSearchSource } from "./parse";
+import { MAX_SIDECAR_RESPONSE_BYTES, type WebSearchSource } from "./parse";
 import { BASE_INSTRUCTION, IMAGE_INSTRUCTION, type SidecarOutcome, type SidecarSettings } from "./executor";
 
 const CCA_FALLBACK_BASE = "https://daily-cloudcode-pa.googleapis.com";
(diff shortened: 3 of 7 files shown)


Ingwannu (changes requested):
Reviewed exact head 6db16f621b8bff09d708e6ea60270b8bca572cd9. The Gemini direction is valuable, but I found three merge blockers that the current 9/9 focused suite does not cover:

1. runGeminiWebSearch obtains an OAuthAccessSnapshot, keeps only accessToken, then separately re-reads the current active account for projectId. If the active account changes while token refresh or request setup is in flight, the request can pair account A bearer with account B project. The snapshot already carries the projectId bound to the exact account and generation; use that same snapshot atomically and add an account-switch regression.
2. After fetchWithResetRetry resolves, neither the error-body read nor the success JSON read attaches cancelBodyOnAbort. This reopens the known abort-before-reader race already documented and fixed in the existing OpenAI and Anthropic sidecars. Attach the guard before either branch reads, and prove an abort immediately after headers cancels and settles the body.
3. Both res.text and res.json materialize an unbounded upstream body. A CCA response is untrusted network input and this sidecar should retain a strict byte ceiling. Use the shared bounded body/bytes primitive with the linked signal, cancel on overflow, parse only bounded UTF-8 JSON, and add oversized success and error regressions.

The focused tests and typecheck pass locally, but exact-head CI is currently red on macOS, and this stack is based on #2242, which still has unresolved requested changes. Please fix the three boundaries, make the parent clean, rerun exact-head CI, then retarget only after the parent lands.

lidge-jun:
## 리뷰 · 우선순위 60 / 80

#2188 L8임. 베이스가 `codex/sidecar-xai-executor` (#2242 L7). 지금 `dev` HEAD `826a1b7a4`엔 제미니 암도 `backends.ts`도 없음. `planWebSearch` (`src/web-search/index.ts:144-204`)가 openai/anthropic만. 루프 (`src/web-search/loop.ts:659-661`)도 그 이분법. L6 유니온이 `gemini`를 이너트로 넣어 두고, 이 PR이 `runGeminiWebSearch`를 살림. 부모 #2242 없이 리타깃하면 깨짐.

실행기 핀이 맞음. `src/web-search/gemini-executor.ts`가 config `baseUrl`을 안 믿음. `getProviderRegistryEntry("google-antigravity")` + fallback `https://daily-cloudcode-pa.googleapis.com`. `src/server/images.ts`랑 같은 규칙. `evil.example`로 자격 안 나감. 테스트가 그거 잠금. UA는 `ANTIGRAVITY_REQUEST_UA` (`src/adapters/google-antigravity-wire.ts:12`). 플레인 UA면 404. 프로브 맞음. `tools: [{google_search:{}}]` + `resolveAntigravityEffortWireModel`이 `thinkingConfig.thinkingLevel` 넣음. `redirect: "manual"`. never-throws. import가 `google-antigravity-wire` + `antigravity-models`만. `google.ts` 어댑터 안 끌어옴. ㅇㅇ 그거 맞음.

플랜/루프 fail-closed. `findGeminiSidecarProvider`가 `google-antigravity` OAuth + 발견된 `projectId`. 없으면 plan 없음. 루프가 `geminiSidecar` 없으면 포워드 실행기/풀 레코더 안 탐. 테스트가 zero fetch + zero pool. `mapCcaGroundedResponse`가 래핑 `{response}` / 플랫 둘 다. groundingChunks uri 디듑. 텍스트 없으면 error. 기본 모델 `gemini-3.7-flash`.

`src/types/config.ts` 안 건드림. L6가 유니온을 이미 넓힘. 스플릿이 이 PR을 바로 무효화하진 않음. 그래도 사이드카 스택이 `index.ts`/`loop.ts`를 같이 소유하니 L7 먼저. 닫을 중복 아님. #2190 `x_search` 안 섞음. #2217/#2227 와이어 기본 안 건드림. #2233/#2246 구글 파트 계약이랑도 별 선. 이쪽은 CCA 엔벨로프지 어댑터 파서가 아님.

2.28 블로커 아님. 프리뷰 배포 플랜 아님. 지금 HEAD 위에 바로 머지하지 말 것. L1→L7 순서.

**해결방안:** #2242 머지 후 리타깃. config `baseUrl` 신뢰 금지 유지. `google.ts` 임포트 넣지 말 것. 스플릿이 사이드카 플랜/루프를 옮기면 리베이스하지 말고 닫고 다시 짜라. 지금은 그 정도 아님.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
All three blockers addressed in 249cc91a3: (1) runGeminiWebSearch now uses the projectId carried by the SAME OAuthAccessSnapshot atomically — an account switch mid-flight can no longer pair account A bearer with account B project (account-switch regression); (2) cancelBodyOnAbort attaches immediately after fetchWithResetRetry resolves, before either the error-body or success-JSON branch reads (abort-after-headers regression proves the body cancels and settles); (3) both reads go through a bounded 64 KiB UTF-8 JSON path with the linked signal, cancelling on overflow (oversized success + error regressions). All four regressions observed red before the fix. Fresh local: 50/0 across three suites, tsc clean, privacy:scan pass. Parent #2242 blockers were addressed in b2c2054b5 on its branch; will rerun exact-head CI after the stack settles. Ready for re-review.

Ingwannu (changes requested):
Reviewed exact head `249cc91a38b742da8e279c885b31dd10490d390f`. The three blockers from my previous review are fixed: the token and project now come from one OAuth snapshot, the response body is guarded immediately after headers, and both success/error payloads use the bounded 64 KiB reader. The focused regressions, typecheck, privacy scan, and exact-head CI are green.

Three current-head blockers remain before this stack layer can be approved:

1. `runGeminiWebSearch` still creates `signalWithTimeout` only after `getValidAccessTokenSnapshot` completes. A stalled Antigravity refresh therefore ignores both the caller abort and `webSearchSidecar.timeoutMs`. The existing CCA image path creates the deadline before auth and races the non-cancellable refresh against it. Apply the same boundary here and add abort/timeout-during-auth regressions.
2. The server option DTO still drops backend provenance (`WebSearchModelOption` has no `backend`), while the GUI web-search picker still calls `sidecarBackendForModel`, whose fallback is OpenAI for every non-Anthropic row. Selecting a `google-antigravity` candidate therefore persists `{ backend: "openai" }` and never reaches this executor. Carry the exact backend in every server option, widen the GUI web-search backend type to include `xai | gemini | exa`, preserve it through grandfathered/current options, and add a picker/save regression.
3. The user and architecture docs are now contradictory: `guides/sidecars.md` documents only OpenAI/Anthropic, and `structure/04_transports-and-sidecars.md` still says explicit Gemini is inert. Document the live Antigravity OAuth/project requirement, default model, fail-closed behavior, and update directly affected translations.

Please keep this on top of #2242, fix these boundaries, and rerun exact-head CI. The implementation remains a strong stack candidate after that.

Issue #2231, opened 2026-08-20
[Bug] Google adapter: candidate, content and parts containers escape the #1332 nested-shape rule


snowyukitty:
### Client or integration

Not client-specific — reproduced by driving the adapter directly with a synthetic `Response`.

### Area

Provider adapter · Streaming / SSE parsing · Non-streaming response parsing

### Summary

#1332 established the rule that a malformed **nested** payload inside a well-formed frame is a
claimed response, not padding, and must fail closed through the adapter's structured error channel.
It applied that rule to the streaming Google *candidate* and to OpenAI Chat *tool calls* on both
paths. Three containers in `src/adapters/google.ts` were left behind, and one of them is a site my
own report in #1325 explicitly cleared.

**1. `content.parts` is consumed unchecked on both paths.** `candidate.content?.parts` is cast to
`GoogleResponsePart[]` and iterated directly, so a container that is not an array of objects either
throws a raw `TypeError` out of the adapter or is consumed as nothing. `parts: "txt"` is the
nastiest of these: a string *is* iterable, so it is walked character by character and the turn
completes empty.

**2. `candidate.content` is never inspected at all.** `content?.parts` reads `undefined` from any
non-record, so a candidate shaped `content: [{ parts: [{ text: "…" }] }]` — a plausible shape from
anything that mirrors the request's repeated `contents` — silently drops its own text and reports a
successful empty turn.

**3. The buffered path never got #1332's candidate guard.** `parseResponse` checks
`!candidates?.length` and then reads `candidates[0]` without a record check, so a
claimed-but-malformed candidate becomes a bare `done`. The streaming path returns a terminal
`google response contained invalid candidates` for the same input.

**Correction to my own earlier report.** #1325 listed `google.ts:809` as "checked and clean —
Google's non-stream path needs no change", on the grounds that it returns a normal `done` for
`{"candidates":[null]}` instead of throwing. That was true about the throw and wrong about the
conclusion: it returns a normal `done` *because* it silently drops the candidate. That is the exact
outcome #1332's rationale names — "silently skipping it could strand a tool turn" — so it is the
same defect, not an exemption from it.

These are found by audit, not from a live capture; same bar as #1325, which was accepted on shape
reasoning plus reproductions. Verified on `dev` at `03735eca`, with #1240, #1249, #1266 and #1332
all present.

| # | Path | Trigger (inside a well-formed frame) | Observed on `dev@03735eca` |
| -- | -- | -- | -- |
[...]

Issue #1219, opened 2026-08-07
[Bug]: SSE frame that parses to `null` escapes the malformed-frame guard and crashes all three stream adapters


brunoflma:
### Client or integration

Claude Code through the OpenCodex proxy (streaming `/v1/responses`), web-search enabled.

### Area

Adapters · Streaming / SSE parsing · Web-search loop

### Summary

All three stream adapters cast `JSON.parse(payload)` to `Record<string, unknown>` and immediately dereference it. `JSON.parse("null")` does **not** throw — it returns `null` — so the surrounding `try/catch` never fires and the next property access crashes the adapter mid-stream.

Live symptom, on `AGR-OAI/claude-opus-5` (`adapter: openai-chat`, `https://agentrouter.org/v1`) with web-search on:

```
stream disconnected before completion: Web-search adapter stream protocol error:
adapter threw: null is not an object (evaluating 'chunk.error')
```

The `Web-search adapter` prefix is misleading — `web-search/progress-stream.ts:316` only wraps whatever the adapter threw. The defect is in the adapters and reaches any streamed request; the web-search loop merely makes it frequent, because that path is where I see repeated upstream instability (`[upstream-retry] connection reset (web-search-loop) — retrying (2/3)`, 8 occurrences in `service.log`).

Affected call sites — same defect, three files:

| File | Unguarded parse | Crashing access | Thrown message |
| --- | --- | --- | --- |
| `src/adapters/openai-chat.ts` | `:941` | `chunk.error` `:950` | `null is not an object (evaluating 'chunk.error')` |
| `src/adapters/google.ts` | `:495` | `chunk.error` `:503` | `null is not an object (evaluating 'chunk.error')` |
| `src/adapters/anthropic.ts` | `:989` | `data.type` `:995` | `null is not an object (evaluating 'data.type')` |

Each already has a `catch` that emits `malformed upstream SSE data frame` (or drops the frame, in `anthropic.ts`). That guard covers syntactically invalid payloads only; a *syntactically valid* payload deserializing to `null` walks straight past it.

### Reproduction

No network and no config changes — the adapters are driven directly with a synthetic `Response`. Save as `repro-null-chunk.ts` and run with the bundled Bun (`node_modules/@bitkyc08/opencodex/node_modules/bun/bin/bun.exe repro-null-chunk.ts`); adjust `PKG` to your install path.

```ts
const PKG = "<...>/node_modules/@bitkyc08/opencodex/src";

[...]

PR #2246, opened 2026-08-21
fix(google): enforce response part field contract

### LINKED ISSUE #2233: [Bug] Google adapter: non-conforming part fields cross the AdapterEvent boundary (nameless tool calls, non-string text)

Ingwannu:
## Summary

Fixes #2233 after the project owner policy decision:

- reject a present Google `functionCall` unless it is an object with a nonblank string `name`
- terminate with a structured adapter error before thought-signature observation or tool dispatch
- drop non-string optional `part.text` values without coercion
- preserve provider-native `functionCall.args` exactly as before
- enforce identical behavior in streaming and buffered response parsing

## Why this disposition

Gemini delivers a complete function call in one response part, so a missing or malformed name cannot be repaired by a later delta. Passing or coercing it can dispatch an unidentified tool. Optional malformed text is different: dropping only that field avoids fabricating assistant output while preserving the rest of the response.

## Verification

- Google hardening suite: 78/78 passed
- Google adjacent adapter suites: 186/186 passed
- `bun run typecheck`: passed
- `git diff --check`: passed
- full repository attempt: 13,717 pass / 15 skip; local environment failures were unrelated to this diff. Missing isolated GUI dependencies were resolved by linking `gui/node_modules`, and service-token shim failures passed after unsetting the host token environment. The remaining translator contract failure is the installed TypeScript rejecting the repository test option `--ignoreConfig` before checking the fixture.

## Decision Log

- Purpose: keep malformed Google-compatible fields from violating the internal string-only event contract or dispatching a nameless tool.
- Existing constraint: container validation did not validate values inside a valid part; Gemini function calls are atomic, while arguments are structured provider-native values.
- Alternatives considered: pass through, coerce, silently drop all malformed fields, or terminate every malformed field.
- Chosen approach: terminal error for invalid call identity, field-only drop for non-string text, unchanged argument serialization.
- Tradeoff: malformed third-party Google-compatible text is ignored, while malformed tool identity ends the turn visibly and safely.

## Review readiness checklist

- [x] All focused tests and static checks are green.
- [x] The branch is based on the latest `dev` at creation time.
- [x] The owner Grok review policy is implemented exactly.
- [x] Ready for another maintainer review; the author will not self-approve or self-merge.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of malformed Google responses during streaming and standard requests.
  - Invalid or incomplete function calls now produce clear protocol errors.
  - Malformed text values are ignored rather than emitted incorrectly.
  - Function arguments are preserved in their original provider format.

- **Documentation**
  - Added documentation describing Google response validation behavior.

### DIFF
--- a/src/adapters/google.ts
+++ b/src/adapters/google.ts
@@ -409,11 +409,16 @@ function artifactMarkdownUrl(filePath: string): string {
 }
 
 interface GoogleResponsePart {
-  text?: string;
+  text?: unknown;
   thought?: boolean;
   thoughtSignature?: string;
   thought_signature?: string;
-  functionCall?: { name: string; args: unknown };
+  functionCall?: unknown;
+}
+
+interface GoogleFunctionCall {
+  name: string;
+  args?: unknown;
 }
 
 /**
@@ -436,10 +441,22 @@ function googleToolCallMetadataFromPart(
  * cannot accidentally expose the same hidden reasoning through different event types.
  */
 function googlePartTextEvent(part: GoogleResponsePart): AdapterEvent | undefined {
-  if (!part.text) return undefined;
+  // A malformed scalar/object is not text and must not cross the AdapterEvent boundary. Dropping
+  // only this optional field preserves the rest of the part without inventing assistant output by
+  // coercion; an empty string keeps its existing no-event behavior.
+  if (typeof part.text !== "string" || part.text.length === 0) return undefined;
   return part.thought === true
     ? { type: "reasoning_raw_delta", text: part.text }
     : { type: "text_delta", text: part.text };
+}
+
+interface InvalidGoogleFunctionCallDiagnostic {
+  reason:
+    | "function_call_not_object"
+    | "function_call_name_invalid"
+    | "function_call_name_blank";
+  partIndex: number;
+  valueType: string;
 }
 
 interface InvalidGoogleShapeDiagnostic {
@@ -460,6 +477,63 @@ function isGoogleRecord(value: unknown): value is Record<string, unknown> {
 function googleStructuralValueType(value: unknown): string {
   if (value === null) return "null";
   return Array.isArray(value) ? "array" : typeof value;
+}
+
+/**
+ * Gemini delivers one complete functionCall per part, so a missing name cannot be repaired by a
+ * later delta. Validate the whole parts array before observing signatures or emitting content: a
+ * malformed call must terminate the claimed response rather than enter replay state or reach the
+ * bridge as a nameless dispatch. Null remains an absence encoding, matching other optional fields.
+ */
+function diagnoseGoogleFunctionCalls(
+  parts: readonly GoogleResponsePart[],
+): InvalidGoogleFunctionCallDiagnostic | undefined {
+  for (let partIndex = 0; partIndex < parts.length; partIndex++) {
+    const functionCall = parts[partIndex]?.functionCall;
+    if (functionCall === undefined || functionCall === null) continue;
+    if (!isGoogleRecord(functionCall)) {
+      return {
+        reason: "function_call_not_object",
+        partIndex,
+        valueType: googleStructuralValueType(functionCall),
+      };
+    }
+    if (typeof functionCall.name !== "string") {
+      return {
+        reason: "function_call_name_invalid",
+        partIndex,
+        valueType: googleStructuralValueType(functionCall.name),
+      };
+    }
+    if (functionCall.name.trim().length === 0) {
+      return {
+        reason: "function_call_name_blank",
+        partIndex,
+        valueType: "string",
+      };
+    }
+  }
+  return undefined;
+}
+
+function googleFunctionCall(part: GoogleResponsePart): GoogleFunctionCall | undefined {
+  const functionCall = part.functionCall;
+  if (!isGoogleRecord(functionCall) || typeof functionCall.name !== "string") return undefined;
+  return { name: functionCall.name, args: functionCall.args };
+}
+
+function invalidGoogleFunctionCallEvent(
+  diagnostic: InvalidGoogleFunctionCallDiagnostic,
+): Extract<AdapterEvent, { type: "error" }> {
+  const subject = diagnostic.reason === "function_call_not_object"
+    ? "invalid function call"
+    : diagnostic.reason === "function_call_name_blank"
+      ? "blank function call name"
+      : "invalid function call name";
+  return {
+    type: "error",
+    message: `google response contained ${subject} (${diagnostic.reason}; partIndex=${diagnostic.partIndex}; valueType=${diagnostic.valueType}) — cannot dispatch`,
+  };
 }
 
 /**
@@ -865,6 +939,11 @@ export function createGoogleAdapter(provider: OcxProviderConfig): ProviderAdapte
             return "terminate";
           }
           parts = rawParts as GoogleResponsePart[];
+          const invalidFunctionCall = diagnoseGoogleFunctionCalls(parts);
+          if (invalidFunctionCall) {
+            yield invalidGoogleFunctionCallEvent(invalidFunctionCall);
+            return "terminate";
+          }
         }
         // Record Gemini thought signatures for the next stateless tool-result turn. Vertex and
         // Antigravity use separate model namespaces so opaque provider state cannot cross routes.
@@ -905,18 +984,19 @@ export function createGoogleAdapter(provider: OcxProviderConfig): ProviderAdapte
                 }
               }
             }
-            if (part.functionCall) {
+            const functionCall = googleFunctionCall(part);
+            if (functionCall) {
               const id = `call_${crypto.randomUUID().slice(0, 8)}`;
               toolCallsStarted++;
               emittedContentEvent = true;
-              const restoredName = restoreGoogleToolName(part.functionCall.name);
+              const restoredName = restoreGoogleToolName(functionCall.name);
               yield {
                 type: "tool_call_start",
                 id,
                 name: restoredName,
                 ...googleToolCallMetadataFromPart(part, pendingStreamThoughtSig),
               };
-              yield { type: "tool_call_delta", arguments: JSON.stringify(part.functionCall.args ?? {}) };
+              yield { type: "tool_call_delta", arguments: JSON.stringify(functionCall.args ?? {}) };
               yield { type: "tool_call_end" };
             }
           }
@@ -1132,6 +1212,8 @@ export function createGoogleAdapter(provider: OcxProviderConfig): ProviderAdapte
         const invalidParts = diagnoseGoogleParts(rawParts);
         if (invalidParts) return finish([invalidGoogleShapeEvent(invalidParts)]);
         const parts = rawParts as GoogleResponsePart[];
+        const invalidFunctionCall = diagnoseGoogleFunctionCalls(parts);
+        if (invalidFunctionCall) return finish([invalidGoogleFunctionCallEvent(invalidFunctionCall)]);
         // Non-streaming Google-family response: observe thought signatures for the next turn,
         // using the same transport-scoped namespace as the streaming path.
         const replayModel = provider.googleMode === "cloud-code-assist" ? antigravityModel : vertexReplayModel;
@@ -1162,16 +1244,17 @@ export function createGoogleAdapter(provider: OcxProviderConfig): ProviderAdapte
               }
             }
           }
-          if (part.functionCall) {
+          const functionCall = googleFunctionCall(part);
+          if (functionCall) {
             const id = `call_${crypto.randomUUID().slice(0, 8)}`;
             toolCallsStarted++;
             events.push({
               type: "tool_call_start",
               id,
-              name: restoreGoogleToolName(part.functionCall.name),
+              name: restoreGoogleToolName(functionCall.name),
               ...googleToolCallMetadataFromPart(part, pendingThoughtSig),
             });
-            events.push({ type: "tool_call_delta", arguments: JSON.stringify(part.functionCall.args ?? {}) });
+            events.push({ type: "tool_call_delta", arguments: JSON.stringify(functionCall.args ?? {}) });
             events.push({ type: "tool_call_end" });
           }
         }
--- a/structure/04_transports-and-sidecars.md
+++ b/structure/04_transports-and-sidecars.md
@@ -673,6 +673,23 @@ classification, preserving the opaque continuation state independently of displa
 - 다른 대안 대신 이 방식을 선택한 이유: Dropping the text loses reasoning replay/display policy input, while duplicated parser rules can drift and exposing marked thoughts violates the provider's visibility boundary.
 - 장점, 단점 및 영향: Internal reasoning no longer leaks into normal answers and both transports stay consistent; downstream reasoning policy still decides whether raw reasoning is rendered or only preserved, and malformed non-boolean markers remain ordinary text rather than broadening hidden-content inference.
 
+## Google response-part field boundary
+
+Google-family adapters validate the values inside an otherwise well-formed response part before
+they become `AdapterEvent`s. A present `functionCall` must be an object with a nonblank string
+`name`; because Gemini delivers that call atomically rather than across deltas, an invalid name is a
+terminal protocol error and is never dispatched. A non-string optional `text` value is dropped
+without coercion, while the rest of the part and turn continue. Structured `functionCall.args`
+remain provider-native and are serialized as before.
+
+[Decision Log]
+- 목적과 의도: Keep malformed Google-compatible response fields from violating the internal string-only text and tool-name contract or dispatching an unidentified tool.
+- 기존 구현 및 제약 조건: Container validation guaranteed object parts, but truthy string/number/array functionCall values emitted a nameless tool call and truthy non-string text values crossed as text or reasoning events. Gemini supplies a complete call in one part, so there is no later name fragment to await.
+- 검토한 주요 대안: Pass malformed values through; coerce them to strings; silently drop every malformed field; terminate the turn for every malformed field; distinguish dispatch identity from optional text.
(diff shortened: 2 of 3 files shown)


lidge-jun:
## 리뷰 · 우선순위 57 / 80

#2233 본체임. 컨테이너는 이미 #2232가 `dev`에 들어감. 지금 HEAD `826a1b7a4` `googlePartTextEvent` (`src/adapters/google.ts:368-372`)가 `if (!part.text)`만 봄. 숫자 `5`, 객체, 배열은 truthy라 `text_delta.text` / `reasoning_raw_delta`에 비문자열이 들어감. 스트림 `:838` / 버퍼 `:1095`는 `if (part.functionCall)`만. 문자열/숫자/배열이 truthy라 `.name`이 undefined인 `tool_call_start`가 브릿지로 감. `AdapterEvent`는 name/text를 string으로 선언함. throw는 없음. 조용히 계약이 깨짐. #2233 본문 14케이스랑 맞음.

이 PR이 그 정책을 그대로 박음. 있는 `functionCall`은 객체 + nonblank string `name`. 아니면 터미널 에러. `cannot dispatch`. openai-chat `unnamedToolCallEvent` (`src/adapters/openai-chat.ts`) 톤. 진단이 파트 전체 순회를 시그니처 관찰보다 먼저. 리플레이 상태에 이름 없는 콜이 안 들어감. 비문자열 `text`는 그 필드만 drop. coerce 없음. `functionCall.args`는 그대로 `JSON.stringify`. 스트림/버퍼 동일. #2233 리뷰가 시킨 그거임. ㅇㅇ

테스트가 `tests/google-hardening.test.ts`에 콜 7종 × 양 경로 + 비문자열 text 4종. 메시지 exact. `src/`만 되돌리면 새 케이스가 깨짐. `structure/04_transports-and-sidecars.md` Decision Log도 정책이랑 맞음. draft임. #2232에 접지 말 것. 그 PR은 컨테이너고 이미 머지됨. 여기 안쪽 필드를 거기 되돌리면 리뷰가 다시 열림.

1st-party Google이 이 셰이프를 안 낸다는 말은 여전함. 노출은 호환 `baseUrl` / 이상한 CCA 엔벨로프. 그래서 점수가 컨테이너 구멍(#2231/#2232)보다 한 칸 낮음. 그래도 이름 없는 콜이 브릿지로 가면 잘못된 툴이 실행될 수 있음. ㅋㅋ 그게 제일 큼. #2216 프리폼이랑 섞지 말 것. args 손대지 말 것.

`src/types.ts` / `src/types/config.ts` 안 건드림. 스플릿이 어댑터 파서를 옮기면 리베이스하지 말고 닫고 다시 짜라. 지금은 그 정도 아님. #2188 사이드카, #2190 `x_search`, #2217/#2227 와이어 기본 무관. 2.28 블로커 아님. 닫을 중복 아님. #2233은 머지될 때까지 열어둠. 클로징 키워드는 머지 커밋에만.

**해결방안:** 체크리스트 채우고 draft 해제 후 `dev` 머지. 정책 바꾸지 말 것 — 콜은 terminate, text는 drop, args는 유지. #2232 되돌리거나 한 PR로 합치지 말 것. 스플릿이 `google.ts` 파서를 옮기면 닫고 다시 짜라.

이 댓글은 grok-bot이 작성했습니다

snowyukitty (commented):
Reviewed at exact head `cb4fbc1c`, rebuilt locally and run through `bun scripts/test.ts`.

**Declaring my position:** I filed #2233, so this is the reporter confirming the implementation
matches the report and does not break anything adjacent — not an independent maintainer approval.
Per `MAINTAINERS.md` the merge still needs one of you; GitHub also refuses an APPROVE from an
account without repository access, so this is submitted as a comment with the verdict stated at the
bottom.

## All fourteen reported shapes behave per the stated policy

| # | `parts` element | On `cb4fbc1c` |
| -- | -- | -- |
| A–D | `functionCall` is `"x"` / `5` / `[]`, streaming and buffered | terminal `invalid function call (function_call_not_object; partIndex=0; valueType=…) — cannot dispatch` |
| E, F, I | `name` absent / `5`, streaming and buffered | terminal `function_call_name_invalid` |
| G, H | `name` is `""` / `"   "` | terminal `function_call_name_blank` |
| J–N | `text` is `5` / `{a:1}` / `[1,2]` / `5`+`thought:true`, streaming and buffered | dropped, turn completes |
| O | control: `[{text:"hi"},{functionCall:{name:"lookup",args:{q:1}}}]` | unchanged |

## Seventeen adjacent cases I probed for a gap, and did not find one

The ones that could plausibly have gone wrong:

- **Fail-closed lands before any content in the same chunk escapes.** `[{text:"hello"},{functionCall:{}}]`
  emits *no* `text_delta` — the whole `parts` array is validated first, matching how #2232 validates
  the container before iterating. Same on the buffered path, and `[{functionCall:{name:"ok"}},{functionCall:5}]`
  emits no tool call either.
- **Content already delivered in an *earlier* chunk is kept**: chunk 1 text, chunk 2 malformed call
  → `[text("hello"), error(…)]`. That is the right split and it is the #1219 line.
- **`partIndex` is accurate** — a bad call at index 2 reports `partIndex=2`, not 0.
- **`name: "\t\n "` is caught by the blank check**, not just space.
- **`functionCall: null` is absence, not corruption** — `[{functionCall:null,text:"hi"}]` still emits
  `text("hi")` and completes. Consistent with how `parts: null` and `content: null` are read.
- **`args` missing and `args: null` both still serialize to `{}`** — unchanged, as intended.
- **A non-string `thoughtSignature` on an otherwise valid call is inert.** `thoughtSignature: 5`
  emits the call with no `providerMetadata`; a real 20-char base64 signature still attaches. Both
  consumers were already `typeof === "string"`-guarded — `isLikelyRealThoughtSignature`
  (`src/adapters/google-antigravity-wire.ts:30-31`) and `extractSignature`
  (`src/adapters/google-antigravity-replay.ts:484-490`) — so widening `text`/`functionCall` to
  `unknown` does not expose them.
- **CCA rejects before replay observation.** A `response`-wrapped malformed call terminates without
  entering the replay cache, which is the part of the docstring I most wanted to see hold.
- **`applyAntigravityReplay` is a cleared site** — it reads *request* contents, not response parts,
  and already guards `typeof fc.name === "string"` (`:707`).
- **#2232's container guards are intact**, checked on this head: `parts: {}` still
  `invalid content parts`, buffered `candidates: [null]` still `invalid candidates`, and the
  mid-stream `{"candidates":null}` frame still completes with its text rather than terminating.

## The one design refinement worth calling out

The owner review on #2233 said "비문자열 text는 파트 drop". This implements **field**-drop, not
part-drop, and the docstring says why: "Dropping only this optional field preserves the rest of the
part". That is a better reading and it is load-bearing — `[{text:5, functionCall:{name:"ok",args:{}}}]`
still dispatches the call. Part-drop would have silently discarded a valid, named tool call over a
malformed sibling field, which is the exact "strand a tool turn" outcome #1332 was written to
prevent. Worth keeping the sentence in `structure/04_transports-and-sidecars.md` as the record of it.

## One non-blocking observation

A part whose *only* content is a malformed `text` now yields a completed, empty turn
(`[{text:5}]` → `heartbeat, done`). That is the chosen policy working as specified and I am not
asking for a change — terminating on optional malformed text would discard otherwise usable
responses, exactly as the Decision Log argues. Recording it because the net underneath it is
`guardEmptyCompletionEventStream`, not the adapter: the retry is what keeps this from being a silent
loss, so it is worth knowing that the two are now coupled for this shape.

Smaller, and genuinely cosmetic: `GoogleResponsePart` widens `text` and `functionCall` to `unknown`
but leaves `thoughtSignature?: string` / `thought_signature?: string` declaring `string`. Runtime is
safe for the reason above. Only mentioning it because the PR's own thesis is that this interface
should not claim what the wire does not guarantee.

## Verification on this head

All through the repository runner, `bun scripts/test.ts`:

- **660 pass / 0 fail** across 28 files — `google-hardening`, `google-adapter`, `google-empty-content`,
  `google-buffered-stop-reason`, the three Vertex suites, the three Antigravity suites,
  `google-signature-history-roundtrip`, `google-claude-prefill-guard`, `google-wire-compiler`,
  `google-tool-schema`, `sse-null-data-frame`, both adapter-conformance suites,
  `openai-chat-hardening`, `images/gemini-inline`, `server-images`,
  `compatibility-provider-equivalence`.
- `bun x tsc --noEmit` — output identical to `dev`, both carrying only the pre-existing
  `Cannot find module '@napi-rs/keyring'`.

No change requested from me. The three-way split — terminate on call identity, drop on optional
text, leave `args` provider-native — is the right set of lines, and it holds on every shape I could
construct.

PR #2248, opened 2026-08-21
fix(responses): strip cross-backend reasoning state on a model switch

olddonkey:
## Summary

Switching model mid-conversation breaks the next turn. Both failures are cross-backend: the client replays history minted by the previous destination, and the new one rejects it.

Reproduced against live providers through the proxy — mint a reasoning item on `xai/grok-4.6`, then replay that history to `openai/gpt-5.6-sol`:

```
replay grok -> grok : OK
replay grok -> SOL  : FAILED  Unknown parameter: 'input[1].status'
  ... with `status` removed:
replay grok -> SOL  : FAILED  The encrypted content ZvQ+...fBJg could not be verified.
                              Reason: Encrypted content could not be decrypted or parsed.
  ... with `status` and `encrypted_content` removed:
replay grok -> SOL  : OK
```

Two independent problems, in that order.

**(a) `status` on a replayed reasoning item.** Grok emits `status: "completed"`; OpenAI rejects the field on input. It is output-only — Grok itself still accepts a replayed item without it, so removing it is safe in both directions.

**(b) A reasoning blob is only decodable by the backend that minted it.** Same class as the compaction-blob provenance problem in #2228, with `encrypted_content` on reasoning items instead.

## Approach — extend the existing mechanism, not a new one

`src/responses/reasoning-replay-cache.ts` already implements a bounded, thread-scoped store (64 entries / 256 KiB / 1 h TTL) and already computes the identity tuple `bindRouteReasoningReplayScope` uses. This adds a small parallel record with the same discipline: remember, per client thread, the identity that served the most recent request; on the next request for that thread, if the identity differs, strip `encrypted_content` from replayed reasoning items before the request goes out.

**No retry-on-4xx.** Deterministic pre-flight only.

### The record compares rotation-safe dimensions

The identity tuple includes the credential's OAuth generation, which changes on every token refresh — and six of the eight `bindRouteReasoningReplayScope` call sites are rotation or refresh rebinds. Comparing that tuple would make every key-pool provider look like a destination switch on refresh. The serving record therefore compares only durable dimensions (`providerDestinationDurableIdentity`, `credentialDurableIdentity`) and **refuses to record** when either is missing, rather than falling back to volatile ones. The proxy's own cache keeps its strict key unchanged.

The failure directions are deliberately asymmetric: **a missed strip costs one degraded turn; a spurious strip is a permanent quality regression.**

### What is deliberately not covered

No record — fresh process, evicted, expired, or no client thread id — keeps the blobs. Stripping on "unknown" would discard valid cached reasoning after every restart. This is stated in a comment rather than papered over, and it is real: a switch that spans a proxy restart still fails. Closing it needs the upstream's own error signal and is a separate change.

### Interaction with other model routing, documented not changed

- **Combo rotation** legitimately drops blobs, and the SSE model-name rewrite hides the switch from the client, so the user sees reasoning quality change with no visible cause.
- **The image and web-search sidecar loops** consume the reasoning scope without rebinding it. That is what keeps internal small-model calls from poisoning the thread's record — worth stating explicitly, since "helpfully" adding a bind there would break it.

Prompts carrying images are unaffected: these transforms touch tool containers, reasoning items and compaction items, never message content.

## Verification

Live, through the deployed proxy, with real blob-bearing items on both routes (blobs 468–3128 chars):

| check | result |
|---|---|
| grok replay, own blob + `content: null` | 200 |
| openai replay, own blob intact | 200 |
| **grok history -> SOL, same thread** | **200** |
| SOL history -> grok, same thread | 200 |

Counterfactual, to show the strip is what does the work rather than the upstream having become lenient — the same grok item (726-char blob, `status` present) replayed to `gpt-5.6-sol`:

| thread | record | result |
|---|---|---|
| same thread id | exists -> strip fires | **200** |
| fresh thread id | none -> blob kept by design | **400** |

That second row is the documented restart gap behaving exactly as designed.

## Tests

`bun run test` — 13742 pass, 10 skip, **1 fail**: `tests/key-login-live-update.test.ts` ("notify after key login pushes the merged row and keeps modelCosts"). That failure is pre-existing and unrelated — it reproduces identically on branches that never touch CLI code, and appears on every branch in this series.

Part of #2240.

## Review readiness checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of encrypted reasoning data during response passthrough.
  - Preserves encrypted reasoning content when routing remains unchanged.
  - Removes stale encrypted content after a confirmed route change.
  - Cleans up reasoning metadata when encrypted content is unavailable.
  - Maintains consistent behavior during credential, model, and destination changes.

- **Tests**
  - Added coverage for reasoning-content preservation, cleanup, route changes, expiration, and bounded replay history.

### DIFF
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -41,7 +41,10 @@ export const FORWARD_HEADERS = [
 
 export function sanitizeReasoningInputContent(
   body: unknown,
-  opts?: { preserveRawReasoningContent?: boolean },
+  opts?: {
+    preserveRawReasoningContent?: boolean;
+    stripEncryptedContent?: boolean;
+  },
 ): unknown {
   if (!body || typeof body !== "object" || Array.isArray(body)) return body;
   const raw = body as Record<string, unknown>;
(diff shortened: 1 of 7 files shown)


Ingwannu (changes requested):
Reviewed exact head `4f9f7288419b3fdcbf1613dfb7a6779bac6b69a8`.

The serving-identity record is committed too early. `bindRouteReasoningReplayScope()` calls `updateReasoningReplayServingIdentity()` while binding the outbound route, before that route has produced a successful terminal response. The helper both compares and writes the current identity.

That advances the thread incorrectly on a failed switch:

1. Route A is the last successful serving identity.
2. A request switches to route B, so A's opaque reasoning is stripped, but B fails before a successful terminal response (network/auth/upstream failure).
3. The store has already been changed to B.
4. The next B retry compares equal and may preserve the client's replayed A blob, causing the same foreign-state failure this PR is intended to prevent.

Please separate comparison from commit. Use the last successfully served identity to decide whether this request must strip foreign opaque state, and record the candidate identity only after a successful terminal response from that route. Add a regression covering A success -> B switch failure -> B retry, asserting that the retry still strips the A blob and that failed attempts do not advance the recorded serving identity.

Once that success-boundary is enforced, the bounded/durable identity direction remains valuable.

olddonkey:

PR #2249, opened 2026-08-21
fix(responses): compare serving identity for compaction blobs too

olddonkey:
> **Stacked PR — merge last.** This branch carries the whole series so it compiles: the fix uses `destinationDecodesNativeCompactionBlob` (#2228), `_stripReasoningEncryptedContent` (#2248) and the sanitize chain shaped by #2237. **Only `248a75c77` is new** — that is the commit to review. Merge after #2217, #2228, #2229, #2237 and #2248.

## Summary

`scrubOcxCompactionItems` forwarded any non-`ocx1:` compaction blob whenever the destination could decode native blobs:

```ts
if (decodeCompactionSummary(encrypted) === null && destinationDecodesNativeBlob) return item;
```

That reasoning is sound only if native blobs have a single minter. They do not — xAI mints them too. So an xAI-minted compaction blob replayed to an OpenAI-operated destination was forwarded verbatim and rejected.

#2228 introduced this gate and its comment stated the assumption outright ("A foreign blob was minted by an OpenAI-operated backend"). That sentence was the bug; this PR corrects the comment along with the code.

## Evidence

Reproduced against the live proxy, on a thread whose serving identity had **already changed and was known to have changed** — so #2248's reasoning-item strip fired correctly while the compaction item sailed straight through:

```
POST /v1/responses  model=gpt-5.6-sol, x-codex-parent-thread-id = <thread last served by xai/grok-4.6>
input: [ {"type":"compaction","encrypted_content":"<opaque non-ocx blob>"}, {user message} ]

-> 400 {"error":{"code":"invalid_encrypted_content",
        "message":"The encrypted content rmey...SQ== could not be verified.
                   Reason: Encrypted content could not be decrypted or parsed."}}
```

## Approach

Reuse the signal the reasoning path already consumes rather than recomputing identity in the adapter. On a known identity mismatch a native blob degrades through the existing `compactionItemToText` note instead of being forwarded; with no known mismatch, behaviour is unchanged.

Degrade rather than drop: the note preserves the conversation, dropping the item would lose context silently.

## Scope — what this does not cover

Only threads the process has actually served. A **cold record** — after a restart, TTL expiry, or LRU eviction — still forwards the foreign blob, because the proxy never saw the turn that minted it and cannot know its provenance. That gap is real and users hit it routinely; closing it needs the upstream's own error signal and is a separate PR.

## Tests

`bun run test` — 13773 pass, 10 skip, **1 fail**: `tests/key-login-live-update.test.ts` ("notify after key login pushes the merged row and keeps modelCosts"). Pre-existing and unrelated; it reproduces identically on branches that never touch CLI code and appears on every branch in this series.

New coverage: identity mismatch degrades a native blob before canonical OpenAI forwarding with adjacent input items untouched; no-mismatch forwarding unchanged in both destination directions; `ocx1:` envelopes unchanged with and without mismatch; blobless items untouched.

Part of #2240.

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for namespace-aware tools across Responses requests and responses, including streaming and JSON formats.
  * Added provider configuration for preserving native compaction data when supported.
  * Improved routing safeguards for encrypted reasoning content and compaction blobs.

* **Bug Fixes**
  * Prevented private tool fields from leaking to upstream providers.
  * Improved restoration of routed tool calls and handling of namespace collisions.
  * Preserved provider-issued encrypted reasoning content during response rewriting.

* **Documentation**
  * Documented namespace-tool conversion, compaction handling, and reasoning-content safeguards.

### DIFF
--- a/src/adapters/base.ts
+++ b/src/adapters/base.ts
@@ -72,6 +72,8 @@ export interface AdapterRequest {
     convertedRoutedCustomToolNames?: ReadonlySet<string>;
     /** Client tool-search names actually lowered to upstream function calls for this request. */
     convertedRoutedToolSearchNames?: ReadonlySet<string>;
+    /** Upstream-only aliases for namespace tools flattened in this request. */
+    convertedRoutedNamespaceToolAliases?: ReadonlyMap<string, { namespace: string; name: string }>;
     /** Releases observation of a serialized request body after its final fetch attempt settles. */
     releaseBodyObservation?: () => void;
     /** Exact reasoning parameter emitted by the adapter, for request-log diagnostics only. */
--- a/src/adapters/base.ts
+++ b/src/adapters/base.ts
@@ -68,7 +68,7 @@ export interface AdapterRequest {
     method: string;
     headers: Record<string, string>;
     body: string;
-    /** Custom-tool names actually lowered to upstream function calls while building this request. */
+    /** Final upstream wire names of custom tools lowered to functions while building this request. */
     convertedRoutedCustomToolNames?: ReadonlySet<string>;
     /** Client tool-search names actually lowered to upstream function calls for this request. */
     convertedRoutedToolSearchNames?: ReadonlySet<string>;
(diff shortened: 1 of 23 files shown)


olddonkey:

PR #2251, opened 2026-08-21
fix(responses): recover when an upstream rejects foreign opaque state

olddonkey:
> **Stacked PR.** Branched from the integrated series so it compiles against #2248's serving-identity record. Only `3b328ef08` is new. Merge after #2217, #2228, #2229, #2237, #2248, #2249.

## Summary

The thread-scoped serving-identity record added in #2248 strips replayed blobs deterministically, but it is in-process and bounded, and it deliberately **keeps** blobs when it has no record — stripping on "unknown" would discard valid reasoning after every restart.

That leaves a failure users hit routinely. From the live usage log, one conversation:

```
19:33:31  xai     grok-4.6      200        <- last grok turn
19:38     proxy restarted (in-process records wiped)
19:48:11  openai  gpt-5.6-sol   400
          "The encrypted content Py6J...kwW9 could not be verified.
           Reason: Encrypted content could not be decrypted or parsed."
```

The proxy never served the turn that minted those blobs, so it cannot know they are foreign. TTL expiry, LRU eviction and any transcript older than the process open the same hole.

## Approach — register a recovery kind, not a new retry path

`image-413` already reacts to an upstream rejection by rebuilding the body once and refetching inside the recovery loop, guarded to a single attempt. This adds `opaque-blob-rejection` on the same shape.

The trigger is deliberately narrow — it fires only when **all** hold:

- 4xx on an `openai-responses` adapter, not already attempted;
- the **exact outbound body** still carried a reasoning or compaction blob (so a resend cannot be byte-identical, and an already-lowered proxy envelope does not qualify);
- the error body carries a decoder's own identity: OpenAI's nested `invalid_encrypted_content`, or one of xAI's two concrete decoder messages. Generic `invalid-argument` prose never gains a hidden resend.

Note xAI names the *compaction* blob even when the item it refused is a `reasoning` item, so the match is on error identity, not on which noun the message uses.

**The deterministic pre-flight stays primary.** When a record exists the first request is already correct and this never runs. The cost when it does run is one extra round trip and one turn of degraded reasoning.

### On the "no retry-on-4xx" rule

#2248's spec said not to add one. That was right against using a retry *instead of* deterministic prevention, and it is wrong as a blanket ban: the proxy cannot determine provenance for history it never served, and the upstream's error code is authoritative and self-identifying.

## Verification

Live, through the deployed proxy, reading `~/.opencodex/usage.jsonl` rather than trusting the HTTP status:

| scenario | sendCount | recoveryKinds | result |
|---|---|---|---|
| cold record, grok blob -> SOL | **2** | `opaque-blob-rejection` | 200 |
| cold record, grok blob + foreign compaction -> SOL | **2** | `opaque-blob-rejection` | 200 |
| cold record, SOL blob -> grok | **2** | `opaque-blob-rejection` | 200 |
| warm record, grok blob -> SOL | 1 | — | 200 |
| same backend, blob replayed | 1 | — | 200 |

**This PR alone does not fix the cold path** — see #2252. With the blob retained, `status` was retained too, and OpenAI rejects `Unknown parameter: 'input[1].status'` before validating the blob, so this recovery correctly never matched. The table above was measured with that companion fix in place.

## Tests

`bun run test` — 13777 pass, 10 skip, **1 fail**: `tests/key-login-live-update.test.ts`. Pre-existing and unrelated; every gate in this series lands on exactly that one failure.

New coverage: the trigger predicate matrix (accepts all three decoder identities; rejects unrelated 400s, 5xx, blobless sends, non-Responses adapters, and repeats), an end-to-end resend asserting exactly two upstream calls with surrounding items byte-equal and the compaction item degraded, the single-attempt guard surfacing the second rejection, and the post-success identity record under a simulated eviction.

Part of #2240.

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added compatibility for namespaced tools across supported Responses API routes.
  * Improved handling of encrypted reasoning content and compaction data across destinations.
  * Added automatic recovery when providers reject unsupported encrypted content.
  * Preserved provider-specific reasoning data during streaming and response rewriting.
  * Added configuration support for destinations that can decode native compaction blobs.

* **Bug Fixes**
  * Prevented tool-name collisions and restored namespaced tool calls accurately.
  * Preserved compaction items without adding unintended identifiers or altering content.

### DIFF
--- a/src/adapters/base.ts
+++ b/src/adapters/base.ts
@@ -72,6 +72,8 @@ export interface AdapterRequest {
     convertedRoutedCustomToolNames?: ReadonlySet<string>;
     /** Client tool-search names actually lowered to upstream function calls for this request. */
     convertedRoutedToolSearchNames?: ReadonlySet<string>;
+    /** Upstream-only aliases for namespace tools flattened in this request. */
+    convertedRoutedNamespaceToolAliases?: ReadonlyMap<string, { namespace: string; name: string }>;
     /** Releases observation of a serialized request body after its final fetch attempt settles. */
     releaseBodyObservation?: () => void;
     /** Exact reasoning parameter emitted by the adapter, for request-log diagnostics only. */
--- a/src/adapters/base.ts
+++ b/src/adapters/base.ts
@@ -68,7 +68,7 @@ export interface AdapterRequest {
     method: string;
     headers: Record<string, string>;
     body: string;
-    /** Custom-tool names actually lowered to upstream function calls while building this request. */
+    /** Final upstream wire names of custom tools lowered to functions while building this request. */
     convertedRoutedCustomToolNames?: ReadonlySet<string>;
     /** Client tool-search names actually lowered to upstream function calls for this request. */
     convertedRoutedToolSearchNames?: ReadonlySet<string>;
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -12,6 +12,7 @@ import { modelRecordValue } from "../reasoning-effort";
 import type { TranslatorBudget } from "../lib/translator-budget";
 import { rewriteRoutedCustomToolsForUpstream } from "../responses/custom-tool-compat";
 import { rewriteRoutedToolSearchForUpstream } from "../responses/tool-search-compat";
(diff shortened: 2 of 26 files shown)


olddonkey:

PR #2252, opened 2026-08-21
fix(responses): strip output-only reasoning status unconditionally

olddonkey:
> **Stacked PR.** Branched from the integrated series. Only `1bbb63b4f` is new. Merge last, after #2251.

## Summary

#2248 kept `status` on any reasoning item that forwarded its `encrypted_content`, to honour "an item whose blob is forwarded is not otherwise modified":

```ts
const stripOutputStatus = hasOutputStatus && !retainsEncryptedContent;
```

That invariant was **defensive rather than observed**, and it silently disabled the cold-record recovery in #2251.

With no provenance record — after a restart, TTL expiry or eviction — the blob is retained, so `status` is retained too, and OpenAI rejects the request on the field **before** it ever validates the blob:

```
400  Unknown parameter: 'input[1].status'.
```

#2251's recovery correctly does not match that error — it is not a blob rejection — so the conversation stayed broken. Two guards each rejecting a different field of the same item: the one that fires first hides the other.

## Evidence that the carve-out protected nothing

Measured against the live backends through the proxy:

1. **OpenAI never mints `status` on a reasoning item.** A real `gpt-5.6-sol` reasoning item has keys `content, encrypted_content, id, summary, type`. The retain branch could therefore only ever fire for an item minted elsewhere — precisely the item OpenAI then rejects. It never protected an OpenAI-minted item, because there is no `status` on one to protect.
2. **Grok accepts its own blob with `status` removed.** A 1707-char grok blob replayed to grok with only `status` deleted: 200.
3. **Removing it is what unblocks the recovery.** The same item replayed to `gpt-5.6-sol` with `status` removed and the blob kept: 200, with `sendCount=2, recoveryKinds=['opaque-blob-rejection']` in the usage log.

## Scope

`status` is now stripped from reasoning input items unconditionally. **The `content` rule is untouched** — blanking predates this and is required by ChatGPT's input contract, and the `dropNullContentChannel` behaviour from #2237 is unchanged.

## Verification

Full live matrix on the deployed build, 7/7:

| scenario | result |
|---|---|
| cold record, grok blob -> SOL | 200 (recovery, sends=2) |
| cold record, grok blob + foreign compaction -> SOL | 200 (recovery, sends=2) |
| cold record, SOL blob -> grok | 200 (recovery, sends=2) |
| warm record, grok blob -> SOL | 200 (pre-flight, sends=1) |
| same backend, grok blob -> grok, `status` now stripped | 200 |
| same backend, SOL blob -> SOL, blob must still verify | 200 |
| same backend, grok blob + `content: null` -> grok | 200 |

## Tests

Existing assertions were **updated, not deleted**: the case that asserted a blob-bearing item is forwarded byte-identically now asserts the explicit expected shape — blob intact, `status` gone — and is renamed to say so. Added a case pinning that an item carrying both `encrypted_content` and `status` keeps the blob and loses the field.

Part of #2240.

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added compatibility for namespaced tools, including safe flattening for upstream requests and restoration in responses.
  * Added recovery for unsupported encrypted reasoning and compaction data, with bounded retries.
  * Improved cross-provider replay handling when routes, models, or credentials change.
* **Bug Fixes**
  * Preserved encrypted reasoning content and sanitized unsupported fields more reliably.
  * Prevented compaction items from receiving unintended identifiers.
* **Documentation**
  * Expanded transport guidance for namespace tools, replay, recovery, and reasoning data handling.

### DIFF
--- a/src/adapters/base.ts
+++ b/src/adapters/base.ts
@@ -72,6 +72,8 @@ export interface AdapterRequest {
     convertedRoutedCustomToolNames?: ReadonlySet<string>;
     /** Client tool-search names actually lowered to upstream function calls for this request. */
     convertedRoutedToolSearchNames?: ReadonlySet<string>;
+    /** Upstream-only aliases for namespace tools flattened in this request. */
+    convertedRoutedNamespaceToolAliases?: ReadonlyMap<string, { namespace: string; name: string }>;
     /** Releases observation of a serialized request body after its final fetch attempt settles. */
     releaseBodyObservation?: () => void;
     /** Exact reasoning parameter emitted by the adapter, for request-log diagnostics only. */
--- a/src/adapters/base.ts
+++ b/src/adapters/base.ts
@@ -68,7 +68,7 @@ export interface AdapterRequest {
     method: string;
     headers: Record<string, string>;
     body: string;
-    /** Custom-tool names actually lowered to upstream function calls while building this request. */
+    /** Final upstream wire names of custom tools lowered to functions while building this request. */
     convertedRoutedCustomToolNames?: ReadonlySet<string>;
     /** Client tool-search names actually lowered to upstream function calls for this request. */
     convertedRoutedToolSearchNames?: ReadonlySet<string>;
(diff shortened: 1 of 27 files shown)


olddonkey:

PR #2254, opened 2026-08-21
fix(xai): restore Grok Responses on the native passthrough route

### LINKED ISSUE #2240: [Bug]: xAI Grok unusable on native Responses in 2.28.0

olddonkey:
Replaces #2217, #2228, #2229, #2237, #2248, #2249, #2251 and #2252 — one review target instead of eight. Those PRs had to merge in a strict order and the later four each carried the whole series as their diff, so reviewing them in isolation was not actually possible. The 16 commits here keep each unit's own evidence and rationale in its message.

Closes #2240.

## The structural cause

#2147 moved OAuth `grok-4.5`/`4.6` off the Chat compatibility wire onto native `openai-responses` **passthrough**, to fix blank-until-done streaming (#1886).

Chat translation had been sanitizing Codex-private wire extensions **as a side effect** — a translator can only emit fields it knows how to build, so anything private was dropped for free. Passthrough forwards the client's bytes, and xAI validates strictly. The switch traded *"the translator sanitizes for free"* for *"nothing sanitizes"*, and the sanitation was never rebuilt.

**Translation is inherently a whitelist; passthrough is inherently a denylist.** Every Codex-private extension is now something someone has to remember to strip.

## The failures are sequential

Each was only reachable once the previous was fixed, which is why they looked unrelated:

| # | When it fires | Symptom |
|---|---|---|
| 1 | first request | `tools[].type = "namespace"` (Codex 0.147+ private tool groups) → 422 |
| 2 | first request | `external_web_access` on `web_search`, `defer_loading` on deferred tools → 400 |
| 3 | **second request** | `"content": null` on the replayed reasoning item → 400 |
| 4 | compaction turns | private `namespace` key on replayed call items |
| 5 | model/backend switch | output-only `status` + a blob minted elsewhere → 400 |

Failure 3 is the conversation-killer: a fresh session dies on its **second** message.

## What the commits do

**Wire sanitation (1, 2, 4)** — lower Codex's private namespace tool groups to flat variants for gateways that only accept those; strip canonical-only tool fields through a declarative table; make namespace dedup order-independent and restore custom/tool-search calls by wire identity.

**The null content channel (3)** — Codex serializes an absent reasoning content channel as `"content": null`; drop the key for routed destinations.

**Cross-backend opaque state (5)** — `encrypted_content` and compaction blobs are decodable only by the backend that minted them. Three mechanisms, in order of preference:

1. **Deterministic pre-flight** — a bounded, thread-scoped record of the identity that served the previous turn; on a mismatch, strip before sending. One round trip, nothing wasted.
2. **Recovery** — for history the process never served (restart, TTL expiry, LRU eviction), the upstream's own narrow error identity is the only authoritative signal. Rebuild once and refetch, registered as a recovery kind alongside the existing `image-413`.
3. Not built: **guessing provenance from blob format.** OpenAI's blobs are Fernet (`gAAAAAB…`) and xAI's are not, so provenance *is* visible in the payload — but hinging correctness on the format of an opaque field means a format change silently strips everyone's blobs.

The pre-flight's failure direction is deliberately asymmetric: **a missed strip costs one degraded turn; a spurious strip is a permanent quality regression.** It keeps blobs when it has no record, and refuses to record at all rather than fall back to volatile identity dimensions — the credential tuple contains the OAuth generation, and six of the eight bind sites are rotation or refresh rebinds, so comparing it would make every key-pool provider look like a destination switch on refresh.

## Four traps, for whoever touches this next

**1. xAI's error names the wrong field.** It calls the reasoning `encrypted_content` a "compaction blob":

```
{"code":"invalid-argument","error":"Could not decode the compaction blob. Ensure it is unmodified from the compact response."}
```

The failing request contained no compaction item at all. Bisecting a captured body settled it:

| variant | result |
|---|---|
| body verbatim | 400 "Could not decode the compaction blob" |
| `encrypted_content` removed | 400 schema — the blob is required |
| whole reasoning item removed | 200 |
| **only the `content` key removed** | **200** |

**2. The two backends want opposite things from the same field.** xAI refuses `"content": null`; OpenAI *requires* the shape kept, because it binds the blob to the item's exact shape. An unconditional strip fixes Grok and breaks OpenAI — that regression shipped to a local deployment and one live request caught it within minutes, while an independent static review of the same diff had returned SHIP. **Anything touching a blob-bearing item needs a live check on both routes; a green suite is not evidence.**

**3. `authMode === "forward"` does not mean "an OpenAI backend answers."** A noncanonical forward provider never receives the caller's credentials. Replaced with `isOpenAiOperatedResponsesDestination`.

**4. When two guards reject different fields of the same item, the first hides the second.** The recovery is armed for the upstream's blob-rejection error — but with the blob retained, `status` was retained too, and OpenAI rejects `Unknown parameter: 'input[1].status'` *before* validating the blob. The recovery correctly never matched: its unit tests passed, its gate was clean, and the live path was unchanged. Removing the field that fired first is what made it reachable.

## Verification

Live, through a locally deployed build, with real blob-bearing items on both routes:

| scenario | result |
|---|---|
| grok first turn, namespace + `external_web_access` + `defer_loading` | 200 |
| grok replay, own blob + `content: null` | 200 |
| the originally captured 2.28.0 failing body, verbatim | 200 |
| openai replay, own blob intact | 200 |
| warm record, grok blob → SOL | 200, `sendCount=1` |
| cold record, grok blob → SOL | 200, `sendCount=2`, `recoveryKinds=['opaque-blob-rejection']` |
| cold record, grok blob + foreign compaction → SOL | 200, recovery |
| cold record, SOL blob → grok | 200, recovery |

Two ways this suite silently tests nothing, both of which bit us: without `include: ["reasoning.encrypted_content"]` **no backend issues a blob**, so every replay check runs on a blob-free item and passes for the wrong reason; and `response.completed` on the OpenAI route omits reasoning items, so a parser that lets it *replace* the accumulated `output_item.done` items drops exactly what is under test. Assert a non-zero blob length as a precondition, and confirm a recovery actually ran by reading `sendCount`/`recoveryKinds` in `usage.jsonl` rather than trusting the HTTP status.

## Tests

`bun run test` on the full series at `c43dc8a0b` (PR head): **13781 pass, 10 skip, 1 fail across 868 files**.

That one failure is `tests/key-login-live-update.test.ts` > "notify after key login pushes the merged row and keeps modelCosts on live and disk". It is **pre-existing and unrelated**, verified directly rather than assumed: running that file alone on untouched `dev` @ `03735eca6` fails identically, same assertion at line 100.

## Not a bug

`grok-4.6-build` in the log UI is `resolvedModel`, the serving id xAI's Grok CLI backend reports (`grok-4.5` → `grok-4.5-build`). It was in usage logs eight hours before any of these fixes; it only became *visible* once requests started succeeding.

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Added compatibility for namespaced tools across Responses providers, including automatic name conversion and restoration.
  * Improved support for custom tools, tool selectors, and routed function calls.
  * Added provider configuration for native compaction-blob support.

* **Bug Fixes**
  * Improved reasoning replay across route or credential changes.
  * Preserved encrypted reasoning content when compatible and safely degraded unsupported content.
  * Added one-time recovery for rejected encrypted compaction data.
  * Preserved all supported compaction item types without unwanted field changes.

### DIFF
--- a/src/adapters/base.ts
+++ b/src/adapters/base.ts
@@ -72,6 +72,8 @@ export interface AdapterRequest {
     convertedRoutedCustomToolNames?: ReadonlySet<string>;
     /** Client tool-search names actually lowered to upstream function calls for this request. */
     convertedRoutedToolSearchNames?: ReadonlySet<string>;
+    /** Upstream-only aliases for namespace tools flattened in this request. */
+    convertedRoutedNamespaceToolAliases?: ReadonlyMap<string, { namespace: string; name: string }>;
     /** Releases observation of a serialized request body after its final fetch attempt settles. */
     releaseBodyObservation?: () => void;
     /** Exact reasoning parameter emitted by the adapter, for request-log diagnostics only. */
--- a/src/adapters/base.ts
+++ b/src/adapters/base.ts
@@ -68,7 +68,7 @@ export interface AdapterRequest {
     method: string;
     headers: Record<string, string>;
     body: string;
-    /** Custom-tool names actually lowered to upstream function calls while building this request. */
+    /** Final upstream wire names of custom tools lowered to functions while building this request. */
     convertedRoutedCustomToolNames?: ReadonlySet<string>;
     /** Client tool-search names actually lowered to upstream function calls for this request. */
     convertedRoutedToolSearchNames?: ReadonlySet<string>;
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -240,15 +240,19 @@ function stripItemIdsWhenUnstored(body: unknown): unknown {
  * A compaction item carries an `encrypted_content` blob the client replays verbatim on every later
  * turn, and only the backend that minted it can decode it. Proxy-minted `ocx1:` envelopes are
  * transparent base64 rather than encryption, so no upstream can read them and they always become
- * plain user messages. A foreign blob was minted by an OpenAI-operated backend: forwarding it to a
- * different destination makes that upstream reject the turn ("Could not decode the compaction
- * blob"), and because the item lives in the client transcript the rejection repeats on every later
- * turn — including the compaction turn the proxy itself drives — leaving the session unable to
- * recover. Off those destinations it degrades to the same note the bridged parser uses.
+ * plain user messages. Native blobs have multiple possible minters, so a destination's ability to
+ * decode its own blobs does not make a blob from a previous serving identity portable. On a known
+ * identity mismatch the blob degrades to the same note the bridged parser uses, even when the
+ * destination normally accepts native blobs. Without a known mismatch, the destination capability
+ * keeps the existing behavior.
  *
  * A bare `context_compaction` marker carries no blob and is forwarded untouched.
  */
-function scrubOcxCompactionItems(body: unknown, destinationDecodesNativeBlob: boolean): unknown {
+function scrubOcxCompactionItems(
+  body: unknown,
+  destinationDecodesNativeBlob: boolean,
+  threadServingIdentityChanged: boolean,
+): unknown {
   if (!isPlainObject(body) || !Array.isArray(body.input)) return body;
 
   let changed = false;
@@ -256,7 +260,11 @@ function scrubOcxCompactionItems(body: unknown, destinationDecodesNativeBlob: bo
     if (!isPlainObject(item) || !isCompactionItemType(item.type)) return item;
     const encrypted = typeof item.encrypted_content === "string" ? item.encrypted_content : undefined;
     if (encrypted === undefined) return item;
-    if (decodeCompactionSummary(encrypted) === null && destinationDecodesNativeBlob) return item;
+    if (
+      decodeCompactionSummary(encrypted) === null
+      && destinationDecodesNativeBlob
+      && !threadServingIdentityChanged
+    ) return item;
     changed = true;
     return {
       type: "message",
@@ -1683,10 +1691,15 @@ export function createResponsesPassthroughAdapter(provider: OcxProviderConfig): 
         // Last, so promoted namespace children are also cleared of Codex-private fields.
         outBody = stripCanonicalOnlyToolFields(outBody);
       }
-      const sanitizedBody = normalizeToolSchemas(stripSparkCompatibility(stripUnsupportedReasoningParams(stripItemIdsWhenUnstored(stripInvalidItemIds(stripUnsupportedHostedTools(sanitizeReasoningInputContent(scrubOcxCompactionItems(outBody, destinationDecodesNativeCompactionBlob(provider)), {
+      const threadServingIdentityChanged = parsed._stripReasoningEncryptedContent === true;
+      const sanitizedBody = normalizeToolSchemas(stripSparkCompatibility(stripUnsupportedReasoningParams(stripItemIdsWhenUnstored(stripInvalidItemIds(stripUnsupportedHostedTools(sanitizeReasoningInputContent(scrubOcxCompactionItems(
+        outBody,
+        destinationDecodesNativeCompactionBlob(provider),
+        threadServingIdentityChanged,
+      ), {
         preserveRawReasoningContent: provider.preserveResponsesReasoningContent === true,
         dropNullContentChannel: !isOpenAiOperatedResponsesDestination(provider),
-        stripEncryptedContent: parsed._stripReasoningEncryptedContent === true,
+        stripEncryptedContent: threadServingIdentityChanged,
       })))))));
       const finalBody = stripDisabledReasoningSummaries(
         normalizeConfiguredReasoningSummaryDelivery(sanitizedBody, provider, parsed.modelId),
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -65,7 +65,6 @@ export function sanitizeReasoningInputContent(
     const hasEncryptedContent = Object.prototype.hasOwnProperty.call(rec, "encrypted_content");
     const stripEncryptedContent = hasOcxEnvelope
       || (opts?.stripEncryptedContent === true && hasEncryptedContent);
-    const retainsEncryptedContent = hasEncryptedContent && !stripEncryptedContent;
     // Codex serializes an absent reasoning content channel as `"content": null`. The field is
     // optional and null carries nothing, but a strict gateway rejects the item on its declared type
     // — xAI answers `Could not decode the compaction blob`, naming the sibling `encrypted_content`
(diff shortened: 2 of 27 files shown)


lidge-jun:
## 리뷰 · 우선순위 76 / 80

#2240 본체임. 지금 `dev` HEAD `84d5523a9`가 그 회귀를 안고 있음. `modelWireDefaults`가 grok-4.5/4.6 OAuth+inbound responses를 `wire: "openai-responses"`로 박음 (`src/providers/registry.ts:1033-1045`). 프로바이더 와이드는 `openai-chat` (`:1004`). #2147 이후 Chat 번역이 곁다리로 닦아주던 Codex private 확장이 패스스루로 그대로 xAI에 감. 첫 턴 namespace / `external_web_access` 422·400. 다음이 `"content": null` 400. 에러가 compaction blob을 욕하는데 blob은 멀쩡함. 시퀀셜임.

#2217/#2227/#2228/#2229/#2237/#2248/#2249/#2251/#2252는 이미 닫힘. 이 PR이 그 여덟 개를 한 리뷰 타겟으로 접음. `registry.ts`는 안 만짐. `modelWireDefaults.wire` 싸움의 승자는 "Responses 기본 유지 + 패스스루를 다시 소독"임. #2227 Chat 기본 복원은 닫힌 채로 두라. 둘 다시 열어서 덮지 말 것. 소유자는 이거 하나임.

코드 경로. `src/responses/namespace-tool-compat.ts`가 비캐논에서 namespace를 평탄화함. `stripCanonicalOnlyToolFields`가 `external_web_access` / `defer_loading`을 테이블로 뜯음 (`src/adapters/openai-responses.ts` 패스스루 `:1679` 근처). `sanitizeReasoningInputContent`가 `dropNullContentChannel`을 `isOpenAiOperatedResponsesDestination` 밖으로만 켬. `reasoningItemToSummaryShape`에 `encrypted_content` 가드 (`responses-reasoning-summary-rewrite.ts`). 크로스-백엔드 스트립은 `reasoning-replay-cache` serving identity + `_stripReasoningEncryptedContent`. 콜드 레코드는 `opaque-blob-rejection` 리커버리. `status`는 무조건 스트립. 라이브 표가 첫 턴/둘째 턴/스위치/콜드 리커버리를 잠금. 방향 맞음.

남은 칼은 예전 #2237 리뷰랑 같음. `dropNullContentChannel`이 `"content" in rec && !Array.isArray(rec.content)`라 문자열/숫자/객체도 조용히 지움. null만 lossless임. `rec.content === null`로 좁히고 비널 malformed는 업스트림에 보이게 회귀 넣어라. 언게이트 스트립이 OpenAI blob 검증을 깨먹은 그 함정 그대로임. `authMode === "forward"`로 안 가른 건 맞음.

#2247이랑 접지 말 것. 저건 ChatGPT 풀 계정 친밀도임. `threadAccountMap` (`src/codex/routing.ts:163`) + `resetCodexRoutingForManualSelection` (`:674`). 이 PR의 serving identity는 프로바이더 destination이지 Codex 계정 풀이 아님. in-process라 재시작 구멍도 같음. `Closes #2240`은 여기 맞음. #2247은 열어둠. #2229 리셰이프는 이 시리즈 안에 들어옴. 지금 `dev`에는 아직 없음.

`src/types/provider.ts`에 `decodesNativeCompactionBlobs`, `src/types/request.ts`에 `_stripReasoningEncryptedContent`, `src/config.ts` 스키마 1줄. 스플릿이 이미 그 파일로 갈라져 있어서 지금 무효화 아님. 나중에 스키마를 다시 옮기면 리베이스하지 말고 닫고 다시 짜라. #2188 L1–L9는 이미 `dev`. 사이드카에 `x_search` 안 넣음. #2190이랑 섞지 말 것. 2.28 태그는 이미 있음. 사용자 체감은 그 태그 기본 Grok가 둘째 턴에 죽음. 그래서 76.

체크리스트 0/4. GitHub `isDraft`는 이미 false임. hygiene 통과. 로컬 1 fail은 `key-login-live-update` overlay라 HEAD에서도 남. 이 diff 아님. 라이브 카나리 없이 그린 수트만 믿으면 안 됨. 본문이 적은 함정 네 개 그대로임.

**해결방안:** null 채널을 `=== null`로 좁히고 malformed 회귀 넣고 체크리스트 채운 다음 `dev` 머지. #2217/#2227 다시 열지 말 것. 와이어 기본은 이 PR이 안 건드리는 현행 Responses. #2240은 머지 커밋이 닫음. #2247은 별 PR. 스플릿이 어댑터 새니타이저/`config.ts` 스키마를 다시 옮기면 닫고 다시 짜라.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
Reviewed exact head `c43dc8a0b037d5713d04ecb0127e42b93f7129be` against the current code and the new automated findings. Consolidating the native-Responses work makes the dependency chain easier to inspect, but this draft is not safe to merge yet.

Blocking findings:

1. The serving-identity bug from the parent stack is still present. `bindRouteReasoningReplayScope` calls `updateReasoningReplayServingIdentity`, which compares **and commits** the candidate identity before the outbound request succeeds. After A succeeds, a failed A→B turn records B anyway; a later B retry then compares equal and may keep A-minted opaque state. Split compare from commit and record the candidate only after a successful terminal response. Add the exact A success → B failure → B retry regression and assert that the retry still strips A state.
2. The CodeRabbit custom-tool gate finding is valid. Noncanonical `authMode: "forward"` providers skip `rewriteRoutedCustomToolsForUpstream` but do run namespace lowering, so a promoted namespace child can remain `type: "custom"`. Use the canonical-forward predicate consistently in both the adapter lowering gate and the core response-name collection/restoration gate, with a noncanonical-forward namespace/custom regression.
3. The OpenAI destination classifier finding is valid. A provider using `baseUrl: "https://api.openai.com"` plus `responsesPath: "/v1/responses"` reaches the official endpoint but is currently classified as routed; that can drop `content: null` from OpenAI-minted encrypted reasoning and degrade native compaction blobs. Recognize both the official origin and `/v1` forms while retaining the Responses-adapter guard, and test both forms plus a non-OpenAI negative case.
4. The namespace-alias recovery finding is valid. Both request-rebuild paths replace `request` without replacing `routedNamespaceToolAliases`. Refresh it from the rebuilt request (empty map when absent) so a changed lowering decision cannot leak or restore a stale alias. Cover OAuth/recovery rebuilds where the alias set changes.
5. #2255 is now merged into `dev` and intentionally makes Chat the default for Grok 4.5/4.6 OAuth. Rebase this branch onto current `dev` and preserve that default. This PR can only be considered as explicit native-Responses opt-in hardening; it must not silently reclaim the default wire.

The request-field comment update is correct maintainability work but is not a separate merge blocker.

After fixing these boundaries, rerun the focused Responses/namespace/replay/compaction suites, typecheck, privacy scan, exact-head CI, and the documented live OpenAI+xAI blob-bearing smoke cases. Keep the PR draft until the 4/4 readiness gate is genuinely complete.

lidge-jun:
Landed on dev via #2258 — your 22-commit series was rebased onto the current head with every units evidence preserved (conflicts against the merged sidecar chain + #2255 chat-default unit resolved by keeping both mechanisms: capability-gated web-search strip alongside the declarative canonical-only table; the sanitize/recovery layers arm the explicit Responses opt-in lane). 464/0 across the touched suites incl. the chat reasoning-streaming E2E. Thank you — this closes the #2240 axis properly.

Issue #2247, opened 2026-08-21
Account pool: continuing a thread after pool selection moves to a new account fails with 400 encrypted content could not be verified (account-bound reasoning ciphertext)


nowhere1975:
### Client or integration

Codex App (native ChatGPT pool routing)

### Area

Proxy and routing (account pool / thread affinity / encrypted reasoning replay)

### Summary

With the ChatGPT account pool (codexAccountMode: pool) and two plus accounts registered, continuing an existing thread after pool selection moves to a different account fails with an upstream 400:

```
The encrypted content 6871...ef-0 could not be verified. Reason: Encrypted content could not be decrypted or parsed.
```

The request's historical reasoning items carry encrypted_content minted by the account that produced them (stored server-side on that account). When the pool routes the thread to a different account - after adding a new account (it becomes selected), or after the service restarts and in-memory thread affinity is lost - the new account's backend cannot verify the old account's ciphertext and rejects the whole request.

### Reproduction

1. opencodex 2.28.0, codexAccountMode: pool, one existing plus account (main), start a conversation (any model, e.g. gpt-5.6-luna)
2. Add a second plus account (account pool) - the new account becomes selected
3. Continue the SAME thread in the Codex app (do not start a new session)
4. Request is routed to the new account namespace (openai-<id>); upstream returns 400 Encrypted content could not be decrypted or parsed

Request log (redacted):

```json
{
  "provider": "openai-pe17f2a",
  "admissionKind": "loopback",
  "inboundProtocol": "responses",
  "accountLogLabel": "pe17f2a",
  "requestedModel": "gpt-5.6-luna",
  "status": 400,
  "errorCode": "invalid_request_error",
  "upstreamError": "The encrypted content 6871...ef-0 could not be verified. Reason: Encrypted content could not be decrypted or parsed."
}
```

A brand-new session on the newly selected account works fine (first turn has no historical ciphertext).

### Root cause analysis

- Native ChatGPT responses carry encrypted_content on reasoning output items; the ciphertext is account-bound (server-side key/storage).
- Codex threads replay history verbatim, so a continued thread sends the previous account's encrypted_content back upstream.
- Pool affinity (resolveCodexAccountForThreadDetailed, CodexThreadAffinityExpiredError) keeps threads on their original account while the in-memory binding survives - but adding an account / restarting the proxy can re-select (ocx account current shows the new account as selected) or drop the binding, so the thread's next request is sent with the OTHER account's credentials and history.
[...]

PR #2264, opened 2026-08-21
fix(responses): address the four review findings on the native passthrough lane

olddonkey:
Addresses the four blocking findings from @Ingwannu's review of #2254, on top of current `dev` (the series itself landed via #2258). Each was confirmed against the code before fixing.

## 1. The serving-identity record committed before the request succeeded

`updateReasoningReplayServingIdentity` compared **and** committed in one call, and `bindRouteReasoningReplayScope` calls it before the request goes out:

```
turn 1  -> A succeeds                    record = A
turn 2  -> B: A != B, strip A blobs      record = B   (committed too early)
           ... this request then fails (rate limit, transport, 5xx)
turn 3  -> retry B: B == B, no strip
           but the transcript still carries A-minted blobs -> rejected
```

The opaque-blob recovery rescues turn 3, so this degraded rather than broke — one wasted round trip and one turn of degraded reasoning on a path meant to be deterministic. The record's *meaning* was the defect: it should mean "this destination served this thread", and a request that never completed served nothing.

Split into `reasoningReplayServingIdentityChanged` (compares, never writes) and `commitReasoningReplayServingIdentity` (records, only at a successful terminal). Bounded discipline unchanged: same LRU/TTL and byte accounting, same refusal to record without a durable identity dimension, same fail-soft direction where no record still means keep the blobs.

**What counts as a successful terminal is stated, not implied.** For bridged transports: `completed` or `incomplete`. For streamed passthrough: a non-error upstream status before relay starts — waiting for SSE completion would retain request state for the stream's lifetime, and a later body failure does not undo that the destination accepted and served the turn. That limit is written in the code rather than papered over.

The two post-recovery re-records are gone; a successful recovery now reaches the same terminal commit as any other success.

Regression test: A succeeds, an A→B turn strips and then fails, and the next B request for the same thread still strips. **Verified it fails against the old code** — a test for this bug that passed before the fix would not be testing this bug.

## 2. One gate used the wrong predicate

Custom-tool lowering was gated on `provider.authMode !== "forward"` while every neighbouring gate uses `!isCanonicalOpenAiForwardProvider`. A noncanonical forward provider therefore skipped `rewriteRoutedCustomToolsForUpstream` but still ran namespace lowering, so a namespace child that was a custom tool got promoted while keeping `type: "custom"`, and the gateway rejects it.

This repeats the mistake the series documented elsewhere: **forward auth says nothing about which backend answers**, because a noncanonical forward provider never receives the caller's credentials.

Both sides move together — the adapter's lowering gate and core's converted-name collection — since lowering names without restoring them is worse than not lowering at all.

## 3. The OpenAI-operated classifier missed a legitimate base-URL form

It compared the normalized base URL for exact equality with `https://api.openai.com/v1`, so a provider configured as `baseUrl: "https://api.openai.com"` with `responsesPath: "/v1/responses"` reaches the official endpoint yet was classified as routed.

Not cosmetic: routed classification drops `content: null` from OpenAI-minted encrypted reasoning and degrades native compaction blobs — this series' own regression, in reverse.

Both official forms are accepted, still by **exact** normalized match, so a lookalike host such as `api.openai.com.evil.test` cannot qualify.

## 4. Request rebuilds left the namespace alias map stale

Every recovery rebuild replaces `request` without refreshing the alias map the response path uses to restore private tool names, so a rebuild that changes the lowering decision restores against a stale map.

Refreshed from the rebuilt request on every path that replaces it — the pre-existing OAuth-401 and image-413 rebuilds included, since the bug is in the rebuild pattern rather than in one caller.

Plus the non-blocking comment fix: `_stripReasoningEncryptedContent` is no longer only a route-switch flag (the recovery sets it too), and the comment now names both producers.

## Scope

This branch does not touch `src/providers/fastwire.ts` or `src/providers/registry.ts`. `dev` deliberately makes Chat the default wire for Grok OAuth (#2255) and that default is untouched here — these fixes harden the explicit native-Responses opt-in lane.

## Tests

`bun run test`: **14019 pass, 10 skip, 2 fail across 886 files.**

Both failures are pre-existing on `dev` and unrelated to this branch, verified directly rather than assumed by running each on untouched `dev` @ `6c928aace`:

| failure | on clean dev |
|---|---|
| `tests/key-login-live-update.test.ts` > "notify after key login pushes the merged row and keeps modelCosts on live and disk" | fails identically |
| `tests/responses-routed-web-search-fields.test.ts` > "official OpenAI API-key traffic retains OpenAI web_search fields" | fails identically (3 pass / 1 fail) |

The second one is a real defect on `dev` — `external_web_access` is stripped from official OpenAI API-key traffic — and is closely related to finding 3 in kind (an official OpenAI destination treated as routed), but it runs through a different gate: **finding 3's fix does not make it pass.** Filed separately rather than folded in here.

## Review readiness checklist

## Summary by CodeRabbit

* **New Features**
  * Improved support for custom tools when routing requests through non-OpenAI providers and gateways.
  * Added support for both standard OpenAI API URL formats.
  * Improved preservation and restoration of namespaced tools during passthrough and recovery.

* **Bug Fixes**
  * Reasoning replay routes are now recorded only after successful responses, improving provider switching and recovery behavior.
  * Prevented lookalike domains from being identified as official OpenAI destinations.
  * Improved handling of opaque reasoning content and stale tool aliases.

* **Documentation**
  * Clarified when encrypted reasoning content is removed by consumers.
  * Documented successful-response requirements for replay route tracking.

### DIFF
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -1696,7 +1696,7 @@ export function createResponsesPassthroughAdapter(provider: OcxProviderConfig): 
       if (!isCanonicalOpenAiForwardProvider(provider)) {
         outBody = promoteClientLoadedTools(outBody);
       }
-      if (provider.authMode !== "forward") {
+      if (!isCanonicalOpenAiForwardProvider(provider)) {
         const rewritten = rewriteRoutedCustomToolsForUpstream(outBody);
         outBody = rewritten.body;
         convertedRoutedCustomToolNames = rewritten.names;
--- a/src/providers/openai-tiers.ts
+++ b/src/providers/openai-tiers.ts
@@ -36,7 +36,15 @@ export function isCanonicalOpenAiForwardProvider(provider: OcxProviderConfig): b
     && normalizedBaseUrl(provider.baseUrl) === CODEX_FORWARD_BASE_URL;
 }
 
-const OPENAI_API_BASE_URL = "https://api.openai.com/v1";
+const OPENAI_API_ORIGIN = "https://api.openai.com";
+const OPENAI_API_BASE_URL = `${OPENAI_API_ORIGIN}/v1`;
+
+function isOfficialOpenAiApiBaseUrl(baseUrl: string): boolean {
+  const normalized = normalizedBaseUrl(baseUrl);
+  // Accept the conventional `/v1` base and the bare official origin used with an explicit
+  // `/v1/responses` path. Exact normalized URLs keep lookalike/suffix hosts out of this set.
+  return normalized === OPENAI_API_ORIGIN || normalized === OPENAI_API_BASE_URL;
+}
 
 /**
  * Whether this provider can serve `POST /responses/compact`. The canonical ChatGPT
@@ -65,7 +73,7 @@ export function supportsNativeResponsesCompactEndpoint(
 export function isOpenAiOperatedResponsesDestination(provider: OcxProviderConfig): boolean {
   if (isCanonicalOpenAiForwardProvider(provider)) return true;
   return provider.adapter === "openai-responses"
-    && normalizedBaseUrl(provider.baseUrl) === OPENAI_API_BASE_URL;
+    && isOfficialOpenAiApiBaseUrl(provider.baseUrl);
 }
 
 /**
--- a/src/providers/openai-tiers.ts
+++ b/src/providers/openai-tiers.ts
@@ -1,5 +1,6 @@
 import type { CodexAccountMode, OcxConfig, OcxProviderConfig, ProviderCostOverlay } from "../types";
 import { OPENAI_PROVIDER_TIER_VERSION } from "../types";
+import { openaiResponsesUrl } from "../adapters/openai-responses-url";
 import { MAX_COST4_RATE } from "../usage/expected-prices";
 
 export const OPENAI_CODEX_PROVIDER_ID = "openai";
@@ -38,12 +39,32 @@ export function isCanonicalOpenAiForwardProvider(provider: OcxProviderConfig): b
 
 const OPENAI_API_ORIGIN = "https://api.openai.com";
 const OPENAI_API_BASE_URL = `${OPENAI_API_ORIGIN}/v1`;
+const OPENAI_API_RESPONSES_URL = `${OPENAI_API_BASE_URL}/responses`;
 
-function isOfficialOpenAiApiBaseUrl(baseUrl: string): boolean {
-  const normalized = normalizedBaseUrl(baseUrl);
-  // Accept the conventional `/v1` base and the bare official origin used with an explicit
-  // `/v1/responses` path. Exact normalized URLs keep lookalike/suffix hosts out of this set.
-  return normalized === OPENAI_API_ORIGIN || normalized === OPENAI_API_BASE_URL;
+/**
+ * The Responses endpoint the adapter would actually POST key-auth traffic to, normalized.
+ *
+ * Mirrors the adapter's own construction (`src/adapters/openai-responses.ts`): a configured
+ * `responsesPath` is appended to the base verbatim, and only the default branch runs the
+ * `/v1/responses` suffix normalization. Classifying on the base URL alone would call
+ * `baseUrl: "https://api.openai.com"` with `responsesPath: "/other"` official even though that
+ * request never reaches the official Responses endpoint.
+ */
+function resolvedResponsesEndpoint(provider: OcxProviderConfig): string | undefined {
+  try {
+    const raw = provider.responsesPath === undefined
+      ? openaiResponsesUrl(provider.baseUrl)
+      : `${provider.baseUrl.replace(/\/$/, "")}${provider.responsesPath}`;
+    return normalizedBaseUrl(raw);
+  } catch {
+    return undefined;
+  }
+}
+
+function isOfficialOpenAiResponsesDestination(provider: OcxProviderConfig): boolean {
+  // Exact normalized URL keeps lookalike/suffix hosts out of this set: `api.openai.com.evil.test`
+  // resolves to its own origin, never to the official one.
+  return resolvedResponsesEndpoint(provider) === OPENAI_API_RESPONSES_URL;
 }
 
 /**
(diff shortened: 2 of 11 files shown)


Ingwannu (changes requested):
Reviewed exact head 3a8554b8f6e. The four focused suites pass (118 tests) and typecheck passes, but one destination-classification blocker remains.

isOpenAiOperatedResponsesDestination currently treats the bare https://api.openai.com base as official without considering responsesPath. The adapter sends key-auth traffic to baseUrl + responsesPath, so a provider configured with that base and a custom path such as /other is still classified as OpenAI-operated. That can preserve OpenAI-only null-content/reasoning or native-compaction semantics for a non-Responses destination.

Please classify the effective Responses endpoint using the same URL-construction rules as the adapter: accept the bare origin only when the resolved path is exactly /v1/responses, retain the conventional https://api.openai.com/v1 default, and add a negative regression for a custom non-Responses path. The branch is also 5 commits behind current dev; rebase and rerun exact-head CI after the fix.

lidge-jun:
## 리뷰 · 우선순위 68 / 80

#2258이 이미 `dev`임. HEAD `3ffffc802`. #2266 GUI 옵트인까지 착지. Chat가 OAuth 기본(#2255). 이 PR은 그 위에 Ingwannu가 #2254에서 찍은 네 칼임. 본체 시리즈가 76이었던 그 레인 후속. 기본 구멍은 막혔고 Responses는 옵트인임. 그래서 76은 아님. 근데 네 개는 지금 `dev`에 그대로 남음.

1번. `updateReasoningReplayServingIdentity` (`src/responses/reasoning-replay-cache.ts:146-182`)가 비교랑 커밋을 한 방에 함. `bindRouteReasoningReplayScope` (`src/server/responses/core.ts:513-515`)가 요청 나가기 전에 부름. A 성공 → B가 스트립하면서 B로 기록 → B가 429 → 다음 B는 B==B라 스트립 안 함. 트랜스크립트는 A 블롭. opaque recovery가 한 바퀴 낭비함. 이 PR이 `reasoningReplayServingIdentityChanged` / `commitReasoningReplayServingIdentity`로 쪼갬. 브릿지는 `completed`/`incomplete`. 패스스루 SSE는 릴레이 전 non-error 상태. 그 한계는 코드에 적혀 있음. 회귀가 A→B 실패 다음에도 스트립을 잠금. 방향 맞음.

2번. 커스텀 툴 로워링이 `provider.authMode !== "forward"`임 (`src/adapters/openai-responses.ts:1704`). 옆 게이트는 전부 `!isCanonicalOpenAiForwardProvider`. 비캐논 forward는 로워링을 건너뛰고 네임스페이스는 탐. `type: "custom"`인 채로 프로모트됨. 게이트웨이가 거절함. 코어 수집도 같은 잘못된 프레디킷 (`src/server/responses/core.ts:2820`). 둘 같이 옮긴 거 맞음. 이름만 낮추고 복원 안 하면 더 나쁨.

3번. `isOpenAiOperatedResponsesDestination` (`src/providers/openai-tiers.ts:65-68`)이 `https://api.openai.com/v1`만 봄. `baseUrl: "https://api.openai.com"` + `responsesPath: "/v1/responses"`는 공식인데 라우티드로 분류됨. `dropNullContentChannel`이 켜짐 (`src/adapters/openai-responses.ts:1739`). `content: null` 날리고 네이티브 compaction blob 깎음. 이 시리즈 회귀를 거꾸로 먹음. exact match 두 폼만 받는 거 맞음. `api.openai.com.evil.test`는 아님. ㅋㅋ 룩어라이크 호스트 구멍은 안 여는 게 맞음.

4번. 리커버리 리빌드가 `request`만 갈아끼고 `routedNamespaceToolAliases`는 첫 빌드 맵 그대로임 (`src/server/responses/core.ts:2834`, 리빌드 `:3044`). 로워링 결정이 바뀌면 응답 복원이 스테일 맵을 씀. 리프레시 헬퍼를 OAuth-401 / image-413 포함 모든 교체 경로에 넣은 거 맞음. 버그가 한 콜러가 아니라 패턴임. `_stripReasoningEncryptedContent` 주석(`src/types/request.ts:68`)도 프로듀서 둘을 이제 부름.

남음. `supportsNativeResponsesCompactEndpoint` (`src/providers/openai-tiers.ts:54`)는 아직도 `/v1` exact만임. 같은 베이스 URL 구멍. 여기 안 넣었음. `imageGenToolCallAliases` 게이트 (`src/server/responses/core.ts:2779`)도 아직 `authMode === "forward"`. 같은 종류. 패스스루 SSE가 바디 실패 전에 커밋하는 한계는 본문이 인정함. ㅇㅇ 그거까지 여기서 닫으려 하면 스트림 수명 동안 요청 상태를 붙잡아야 함.

`src/types/request.ts`는 주석만. `config.ts` 안 만짐. 스플릿 안 씹힘. `registry.ts`/`fastwire.ts` 안 건드림. Chat 기본 유지. #2188 사이드카 이미 `dev`. `x_search` 안 넣음. #2190이랑 섞지 말 것. #2267이 같은 `openai-responses.ts`를 테이블 쪽으로 만짐. 저건 #2262가 이미 착지한 필드 스코프 싸움임. 이 PR에 접지 말 것. #2247 계정 풀이랑도 접지 말 것. 2.28 태그 이미 있음. 프리뷰 배포 아님. 닫을 중복 아님. draft 체크리스트 0/4.

**해결방안:** 체크리스트 채우고 draft 해제 후 `dev` 머지. compare/commit 분리 유지. 커스텀 툴 게이트는 `isCanonicalOpenAiForwardProvider`만. official origin 두 폼은 exact. 알리아스 맵은 리빌드마다 리프레시. compact 헬퍼랑 imageGen `authMode` 게이트는 후속. 스플릿이 `request.ts` 스키마를 다시 옮기면 닫고 다시 짜라. 지금은 주석이라 그 정도 아님.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
Landed on dev via #2273 — rebased cleanly (no conflicts) with your three commits preserved; the deferred serving-identity commit semantics and the failed-A->B-retry strip regression both verified against the pool-switch regressions from #2269. Thank you for the fast, well-evidenced follow-through on the review findings.

Issue #2152, opened 2026-08-19
[Bug][Windows]: six pre-existing CI shard failures (WP13 startup cost, symlink fixtures, Bun panic)


lidge-jun:
### Client or integration

Other

### Area

Platform (Windows / macOS / Linux)

### Summary

Six tests fail on the Windows CI shards for reasons that predate the current release range. They are recorded here so the leg's remaining red is a known quantity rather than something the next person re-derives from scratch.

**Verified as not caused by `main...dev`:** `git log origin/main..origin/dev` returns **zero commits** for every file involved.

**Group 1 — WP13 composed acceptance (3 cases), `tests/codex-composed-acceptance.test.ts`**

`A-reduced`, `E: separate fake homes share the effective-user Codex lock`, and `Restore truth`. Each case starts a real `ocx start` more than once — spawn a Bun runtime, load the CLI, read config, bind a port, publish `runtime-port.json` — and the Windows runner cannot finish that inside the budget while running a quarter of the suite.

Worth stating plainly: **these have never passed on Windows.** The 2026-08-18 run I first compared against ([32147924436](https://github.com/lidge-jun/opencodex/actions/runs/32147924436)) had shard 4/4 **cancelled**, so they never executed there at all. "No failures listed" was not "passed".

A cascade on top of this was fixed in #2143: teardown aborted on the first child that would not exit, so survivors were killed by Bun's between-file sweep and the *next* case failed with exit 143. That is gone; what remains is the underlying cost.

**Group 2 — npm cache preflight (3 cases), `tests/update-npm-cache-preflight.test.ts`**

`lstats normal nested symlinks but never traverses their targets`, `a foreign-owned nested symlink does not block the update`, and `a cache root symlinked to another volume is inspected, not rejected` all return `cache_entry_inaccessible`. An unprivileged Windows user cannot create symlinks, so the fixtures cannot build the shape they assert on. Several neighbouring cases already `skip` for the same reason.

**Group 3 — shard 2 Bun runtime panic**

`panic(thread): Internal assertion failure` — a Bun crash, not a test result. The workflow already carries a crash-signature retry for the macOS leg (`is_bun_runtime_crash`); the Windows shards have no equivalent.

### Reproduction

1. Dispatch `ci.yml` against `dev` (the Windows leg only runs on `workflow_dispatch`).
2. Read the `windows 2/4` and `windows 4/4` job logs.
3. Observe the three groups above. Shard 3/4 passes.

### Version

[...]

PR #2265, opened 2026-08-21
fix(lab): name the ACL failure code in the publisher key error

ntdatt812:
Refs #2152.

## What this changes

`CL-10 public bundle and publisher > builds deterministic bundle ids and digests` fails on the Windows leg with:

```
PublicEvidenceValidationError: public publisher key ACL hardening did not complete
      at requirePublisherKeyAcl (src\lab\public\signature.ts:73:11)
```

That message is the same string for every required-harden failure. You wrote in #2152 that telling the three causes apart needs `error.cause.code` and that you would rather leave the question open than aim a fix at the wrong one. This PR does not guess at the cause — it puts the code in the message, so the next dispatched run answers the question by itself.

After this change the same failure reads `... did not complete (ETIMEDOUT)`, `(EACLIDENTITY)` or `(EICACLS)`.

## Why the message and not just `cause`

`cause` is already attached — `requirePublisherKeyAcl` sets it — but the reporter prints `message` and the stack, which is why the trace in the issue has no cause line. Confirmed on Windows by driving the failure through the existing seams and reading both fields:

```
[EICACLS   (icacls refused)]
   message   = public publisher key ACL hardening did not complete
   cause.code= EICACLS
[ETIMEDOUT (budget)]
   message   = public publisher key ACL hardening did not complete
   cause.code= ETIMEDOUT
```

Identical messages, different causes. Nothing in the log distinguishes them.

## What crosses the boundary

Only the errno-shaped code, and its shape is re-checked at the point of use rather than trusted:

```ts
return typeof code === "string" && /^[A-Z][A-Z0-9_]{1,30}$/.test(code) ? code : null;
```

`sanitizedAclError` already allowlists which codes it copies, so this is the second bound rather than the only one — it means a code added there later cannot widen what a public log prints without also passing this shape. A code in that shape has no separator, no lowercase and a bounded length, so it cannot carry the key pathname or the username component inside it. A cause with no such code keeps the previous message byte-for-byte, so no failure gains an empty `()`.

`PublicEvidenceValidationError.code` stays `public_publisher_key_unsafe` and `cause` is untouched, so nothing that branches on either changes.

## One observation about the CI failure itself, offered as a lead, not a claim

Running `bun test tests/lab-public-evidence.test.ts` alone on Windows 11, unelevated, Bun 1.3.14: **14 pass, 0 fail**. So it is not categorically broken on Windows, which narrows it to the runner.

Of your three candidates, one has a property that matches the shape you described — the first call failing and the very next one succeeding on a different temp home. A *successful* SID lookup is process-cached in `windows-user-principal.ts`; a *failed* one is not, so a cold or contended first lookup fails alone and the next call re-runs it warm. Driving exactly that through the seam reproduces your pattern:

```
[call #1 (lookup times out)] cause.code=EACLIDENTITY
[call #2 (lookup recovers)]  OK (no throw)
   principal lookups attempted = 2
```

That is consistent with EACLIDENTITY, not evidence for it — ETIMEDOUT would look similar if the memo key differed per stage. I have not reproduced the CI failure itself and am not proposing a fix for it here. The point of this PR is that the next `platform-windows` dispatch will print which one it was.

## Verification

Windows 11, Bun 1.3.14, on this branch:

```
tests/lab-public-security-regressions.test.ts                  13 pass, 0 fail
8 lab-public suites (evidence, core-contract, lifecycle-hardening,
  surfaces, export-transaction, provenance-recovery, file-safety,
  security-regressions)                                        56 pass, 1 skip, 0 fail
tests/windows-secret-acl.test.ts + windows-user-principal.test.ts
                                                              177 pass, 0 fail
bun x tsc --noEmit -p tsconfig.json                             0 errors
bun run privacy:scan                                            Privacy scan passed
```

Measured on both sides: the same 8 suites on the unpatched tree are `52 pass, 1 skip, 0 fail` across 53 tests, and 57 tests here. The delta is the 4 added cases and nothing else.

**Mutation-checked.** Reverting only the message to the old fixed string, keeping the tests:

```
(fail) a required publisher key ACL timeout names ETIMEDOUT in its message
(fail) a required publisher key icacls refusal names EICACLS in its message
(fail) a required publisher key SID lookup failure names EACLIDENTITY in its message
 10 pass | 3 fail
```

Exactly the three naming cases, while `keeps the plain message` stays green — it covers the other branch, so it should not move, and it does not.

## Tests added

Four cases in `lab-public-security-regressions.test.ts`:

- ETIMEDOUT, EICACLS and EACLIDENTITY each named in the message. The EACLIDENTITY case is injected through `setWindowsPrincipalRunnerForTests`, one module further out than the icacls runner, so it also pins that the code survives the hand-off from `windows-user-principal` rather than only the icacls path.
- a cause with no errno-shaped code leaves the message byte-for-byte unchanged.

`afterEach` now also clears the principal runner, so an injected lookup cannot leak into a later test in the file.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ACL security error messages by including validated diagnostic codes when available.
  * Preserved a clear generic message when no recognized error code is provided.
  * Retained the original underlying error for troubleshooting.

### DIFF
--- a/src/lab/public/signature.ts
+++ b/src/lab/public/signature.ts
@@ -52,6 +52,22 @@ function publisherForPrivateKey(privateKeyPem: string): PublicPublisherV1 {
   };
 }
 
+/**
+ * The bounded errno-shaped code from a failed ACL harden, or null when the cause
+ * carries none.
+ *
+ * Only the code is allowed into the message. `hardenSecretPath` already sanitizes
+ * its own diagnostic prose, but this error is what reaches a CI log, so what
+ * crosses that boundary is re-checked here rather than trusted: an errno code has
+ * no separator, no lowercase and a bounded length, and therefore cannot carry the
+ * key pathname or the username component inside it.
+ */
+function aclFailureCode(error: unknown): string | null {
+  if (!(error instanceof Error) || !("code" in error)) return null;
+  const code = (error as NodeJS.ErrnoException).code;
+  return typeof code === "string" && /^[A-Z][A-Z0-9_]{1,30}$/.test(code) ? code : null;
+}
+
 function requirePublisherKeyAcl(path: string, timeoutMemoKey = path): void {
   privateRegularFileSize(path, PRIVATE_KEY_FILE_OPTIONS);
   let hardened: { ok: boolean };
@@ -63,9 +79,17 @@ function requirePublisherKeyAcl(path: string, timeoutMemoKey = path): void {
     hardened = { ok: false };
   }
   if (!hardened.ok) {
+    // Name the cause in the message, not only on `cause`. Every harden failure
+    // reaches a CI log as this one string, and the three that occur there need
+    // different fixes: ETIMEDOUT is the budget, EACLIDENTITY is the effective-SID
+    // lookup, EICACLS is icacls refusing the path. A message identical across all
+    // three cannot be acted on without a Windows box to re-run it under (#2152).
+    const code = aclFailureCode(hardeningError);
     const failure = new PublicEvidenceValidationError(
       "public_publisher_key_unsafe",
-      "public publisher key ACL hardening did not complete",
+      code
+        ? `public publisher key ACL hardening did not complete (${code})`
+        : "public publisher key ACL hardening did not complete",
     );
     if (hardeningError !== undefined) {
       (failure as Error & { cause?: unknown }).cause = hardeningError;
--- a/tests/lab-public-security-regressions.test.ts
+++ b/tests/lab-public-security-regressions.test.ts
@@ -15,11 +15,13 @@ import {
   setIcaclsRunnerForTests,
   setPlatformForTests,
 } from "../src/lib/windows-secret-acl";
+import { setWindowsPrincipalRunnerForTests } from "../src/lib/windows-user-principal";
 
 const roots: string[] = [];
 
 afterEach(() => {
   setIcaclsRunnerForTests(null);
+  setWindowsPrincipalRunnerForTests(null);
   setPlatformForTests(null);
   resetHardenedStateForTests();
   for (const root of roots.splice(0)) {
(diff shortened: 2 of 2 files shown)


lidge-jun:
## 리뷰 · 우선순위 32 / 80

#2152 Windows CI 조각임. 지금 `dev` HEAD `3ffffc802` `requirePublisherKeyAcl` (`src/lab/public/signature.ts:55-75`)가 실패 메시지를 한 줄로 고정함. `public publisher key ACL hardening did not complete`. `cause`는 붙임. 리포터가 `message`랑 스택만 찍어서 로그에 코드가 없음. ETIMEDOUT / EACLIDENTITY / EICACLS가 같은 문장임. 본문이 그거임.

이 PR이 `aclFailureCode`를 넣음. `cause.code`가 `^[A-Z][A-Z0-9_]{1,30}$`일 때만 메시지에 `(CODE)`를 붙임. 아니면 예전 문장 그대로. `PublicEvidenceValidationError.code`는 `public_publisher_key_unsafe` 유지. pathname/username이 로그로 못 나가게 두 번 막음. `sanitizedAclError` allowlist 다음 두 번째 바운드임. 방향 맞음. ㅋㅋ 원인 추측은 안 함. 다음 `platform-windows` 디스패치가 답을 찍게 함.

테스트 네 장. ETIMEDOUT / EICACLS / EACLIDENTITY 각각 메시지에 코드. EACLIDENTITY는 `setWindowsPrincipalRunnerForTests`로 한 모듈 밖에서 주입. 핸드오프 생존을 잠금. 코드 없는 cause는 문장 byte-for-byte. 지금 `dev` `afterEach` (`tests/lab-public-security-regressions.test.ts:21-28`)는 icacls/platform만 지움. 이 PR이 principal runner도 지움. 뮤테이션 체크까지 함. 메시지 되돌리면 세 장이 깨짐.

고친 게 아님. 진단임. #2152 본체(윈도우 샤드 실패)는 열림. 점수 48. 이건 그 로그를 읽을 수 있게 만드는 선행임. Grok/Responses 레인 아님. #2188 사이드카, #2190 `x_search`, #2254/#2264 패스스루 전부 무관. `types.ts`/`config.ts` 안 만짐. 스플릿 안 씹힘. 2.28 블로커 아님. 프리뷰 배포 아님. 닫을 중복 아님.

draft 체크리스트 0/4. 비전공자가 윈도우 박스 없이 로그를 읽으려면 코드가 메시지에 있어야 함. 그 점은 맞음. 그래도 핫 레인이 아님. 원인은 다음 디스패치가 찍음.

**해결방안:** 체크리스트 채우고 draft 해제 후 `dev` 머지. 메시지에 errno 코드만 넣고 pathname 넣지 말 것. #2152 본체 픽스는 별 PR. 스플릿이 lab public signature를 옮기면 닫고 다시 짜라. 지금은 그 정도 아님.

이 댓글은 grok-bot이 작성했습니다

ntdatt812:
Rebased onto `dev@3c6a4524b` and re-ran. Ticking the readiness boxes, with one thing stated rather than glossed.

The full local suite is **14029 tests, 1 failure**, and that failure is not this change:

```
(fail) GET /api/client-config > an accepted override still resolves through the route
Expected: "/tmp/opencodex-pi-route-fixture/models.json"
Received: "\tmp\opencodex-pi-route-fixture\models.json"
```

A Windows path-separator assertion in `tests/management-client-config-route.test.ts:322`, in the pre-existing `platform-windows` class tracked by #2152. Measured on both sides rather than assumed — reverting this PR's two files to their parent revision and re-running that file gives the identical failure:

```
with the patch      13 pass, 1 fail
parent revision     13 pass, 1 fail   # aclFailureCode absent, verified
```

Nothing else in the suite fails. I will open that one separately; it is unrelated to the publisher key path.

Re-verified on the rebased head:

```
lab-public-security-regressions + lab-public-evidence
  + windows-secret-acl + windows-user-principal     204 pass, 0 fail
```

CodeRabbit reported no actionable comments, and CI is green on the pre-rebase head.

Ingwannu (approved):
Approved on exact head f882de0fa. The Grok maintainer review is correct: this is a bounded diagnostic change, not a speculative Windows ACL fix. Only an errno-shaped code can enter the public failure message; the stable error code and cause behavior remain unchanged. I verified 204 focused Lab/Windows tests and typecheck on the rebased head. This should merge to dev while #2152 remains open for the diagnostic result.

PR #2283, opened 2026-08-21
fix(router): backfill the xAI web-search capability into routed provider config

lidge-jun:
## Summary

Routed Grok turns on the Responses lane died with `400 Argument not supported: external_web_access` before inference, so every tool-use turn failed.

`routedProviderConfig()` backfills every other registry-only scalar (`supportsServiceTier`, `preserveResponsesReasoningContent`, `fastWire`) but not `supportsOpenAiWebSearchToolFields`. `enrichProviderFromRegistry()` does fill it — and the request path never calls that function. So a saved `xai` row reached the Responses adapter with the flag `undefined`, the #2262 capability gate read that as "unclassified upstream, keep the fields", and Codex's OpenAI-only `web_search` config went to the wire.

A live probe against the OAuth Grok endpoint (2026-08-21) isolates the cause: bare `{type:"web_search"}` returns 200, `+external_web_access` returns 400, `+search_context_size` returns 400 — each individually, before inference.

The fix is one backfill entry next to the existing registry-only scalars. An explicit saved user value still wins, and only providers whose registry entry declares the capability are affected (today: `xai`).

## Verification

- `bun run test` — 14090 pass / 0 fail across 887 files
- `bun test --isolate tests/responses-routed-web-search-fields.test.ts` — 7 pass / 0 fail
- New tests driven **red** against the unfixed router (2 fail), then green with the fix — they are not vacuous.
- End-to-end on a remote macOS host running this dev head: with the GUI Responses opt-in on and **no** hand-written capability in `config.json`, a multi-step `codex exec` tool-use turn (mkdir, write, cat, uname) completed over adapter `"openai-responses"` with `status: 200` in `usage.jsonl`. The same turn 400'd on the same host before the fix.

The existing tests could not catch this: they hand-build a provider with the flag already set, or call `enrichProviderFromRegistry()` directly, so both start downstream of the break. The new tests assert on `routedProviderConfig()` output.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved routed provider configuration to correctly apply web-search capability defaults when saved settings omit them.
  * Preserved explicitly configured provider capabilities.
  * Prevented unsupported web-search fields from being sent to providers that do not support them.

* **Tests**
  * Added coverage for default capability handling, explicit overrides, and request-field filtering.

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -355,6 +355,17 @@ export function routedProviderConfig(providerName: string, provider: OcxProvider
     ...(provider.supportsServiceTier === undefined && registryEntry.supportsServiceTier !== undefined
       ? { supportsServiceTier: registryEntry.supportsServiceTier }
       : {}),
+    // Registry-only web-search capability: without this backfill a saved provider row reaches
+    // the Responses adapter with the flag `undefined`, so the capability gate added in #2262
+    // reads "unclassified" and forwards Codex's OpenAI-only `web_search` config fields. xAI
+    // rejects the whole request before inference ("Argument not supported:
+    // external_web_access"), which killed every routed Grok turn on the Responses lane.
+    // enrichProviderFromRegistry() already fills this, but the request path resolves through
+    // routedProviderConfig() and never called it.
+    ...(provider.supportsOpenAiWebSearchToolFields === undefined
+      && registryEntry.supportsOpenAiWebSearchToolFields !== undefined
+      ? { supportsOpenAiWebSearchToolFields: registryEntry.supportsOpenAiWebSearchToolFields }
+      : {}),
     ...(provider.preserveResponsesReasoningContent === undefined && registryEntry.preserveResponsesReasoningContent !== undefined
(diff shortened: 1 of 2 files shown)


Ingwannu (approved):
Approved exact head `d5ea14a2d522b428ff5b13d011bdf9e376b41358`.

I independently verified the request-path gap and the fix. `routedProviderConfig()` was the missing registry-enrichment boundary: a saved xAI row could reach the Responses adapter with `supportsOpenAiWebSearchToolFields` unset even though the registry declares `false`. The new fill-only backfill matches the established `enrichProviderFromRegistry()` contract, applies only to registry-owned transport rows, and preserves an explicit saved override.

The focused request/router and passthrough suites pass locally (105/105), as do `bun run typecheck` and `bun run privacy:scan`. The new tests cover the saved-row path that the older adapter-only tests missed, including actual field stripping and explicit-value precedence.

No blocking issue found. Merge should wait for the remaining exact-head macOS job to finish green.

PR #2293, opened 2026-08-21
scripts: add Windows Codex desktop full-restart helper

lidge-jun:
## Summary

Adds `scripts/restart-codex-desktop-app.ps1`, a standalone helper that fully restarts the Windows Codex desktop app (MSIX package) so the model picker re-reads the on-disk catalog after `ocx sync`.

On Windows, `ocx sync --restart-codex` restarts only the app-server child; the Electron UI (`ChatGPT.exe`) keeps its cached model list until the whole app is quit and relaunched. This script:

- resolves the package install location at runtime via `Get-AppxPackage` (survives version bumps)
- selects only the root process whose parent lies outside the package, so `taskkill /T /F` cascades to all children
- tries graceful close first (`CloseMainWindow()`), waits up to 15 s, then forces
- guards against killing its own ancestry on real runs
- relaunches via MSIX activation (`shell:AppsFolder\<AUMID>`), not a raw exe path

Full investigation and evidence: lidge-jun/opencodex#2292.

## Verification

- PowerShell 5.1 parser: PARSE OK
- `-DryRun` exit 0 listing 12 live targets with root PID 9928 and correct AUMID relaunch command; nothing stopped
- `bun x tsc --noEmit`: exit 0 (no-regression check)
- Full restart not executed from inside this session (would kill the host); left for external-terminal validation

## Checklist

- [x] Local CI green (focused probes + tsc)
- [ ] Branch on latest dev commit
- [ ] Codex/CodeRabbit findings fixed
- [ ] Ready for review

## Summary by CodeRabbit

* **New Features**
  * Added a Windows utility to restart the Codex desktop app and refresh stale model-picker data.
  * Supports dry-run reporting, graceful shutdown, optional forced termination, process-tree handling, and automatic relaunch.
  * Includes safeguards to target only the intended app process.

* **Documentation**
  * Added implementation plans, safety requirements, acceptance criteria, and verification guidance for the Windows restart workflow.

### DIFF
--- a/devlog/_plan/260821_260821-windows-picker-full-restart/000_plan.md
+++ b/devlog/_plan/260821_260821-windows-picker-full-restart/000_plan.md
@@ -1,0 +1,91 @@
+# 000 Plan: Windows model-picker full-restart path
+
+## Problem
+
+ocx sync --restart-codex rewrites the Codex catalog JSON and restarts the Codex
+app-server (codex.exe app-server). Observed behavior:
+
+- macOS: the desktop app model picker reflects the new catalog right away.
+- Windows (stable/beta, MSIX package OpenAI.Codex_26.818.3698.0): the picker
+  keeps the stale list until the whole desktop app is quit and relaunched.
+
+Local evidence (2026-08-21):
+
+- Desktop UI processes are ChatGPT.exe (Electron shell), installed as MSIX
+  package family OpenAI.Codex_2p2nqsd0c76g0, start app id (AUMID)
+  OpenAI.Codex_2p2nqsd0c76g0!App.
+- ocx sync --restart-codex matches only codex.exe app-server and
+  codex-code-mode-host.exe command lines
+  (src/codex/app-server-processes.ts, isCodexAppServerCommandLine). The
+  Electron UI is never signalled, so its cached picker survives.
+- After the 20:57 sync + restart, codex.exe (PID 8592) started fresh at 20:59
+  while all ChatGPT.exe UI processes kept their earlier start time, and the
+  picker still showed only OpenAI models.
+
+Research findings (subagent, bundle inspection of app.asar):
+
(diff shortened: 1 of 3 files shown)


Ingwannu (changes requested):
Reviewed exact head `6c33ea5dd131e6e778c1d2f94948e6156324ce07`. The direction is useful, but this head is not merge-ready.

I independently verified the automated findings against the current script, and both are real:

1. `$roots` is derived from every package-path target whose parent is absent from `$targets`; it is not restricted to a verified `ChatGPT.exe` desktop root. An independently launched package `codex.exe`, or a child whose package-owned parent was omitted from the filtered set, can become a forced `/T /F` root. Re-resolve each candidate immediately before shutdown and require the expected current-user package identity, executable name/path, parent boundary, PID, and creation identity.
2. Graceful success is decided only by the root PID disappearing. Original `codex.exe` / code-mode-host targets may survive, `taskkill` exit status is discarded, and the script relaunches without proving every original verified target exited. Recheck all original identities after the bounded wait, force only still-identical survivors, check the trusted `taskkill` result, and fail before relaunch if any survivor remains.

There is also a product-boundary blocker relative to #2292 and the owner/Grok review there: `scripts/` is excluded from the published npm package (`package.json` ships only `bin`, `src`, GUI/assets/docs), and this PR does not wire an opt-in `ocx sync --restart-desktop-app` path. Therefore ordinary installed users receive no supported command, and the existing misleading Windows `--restart-codex`/doctor path remains unchanged. Please either implement the explicit Windows-only CLI integration with the existing trusted PowerShell/taskkill resolution and fail-closed discovery, tests, and user documentation, or clearly scope this as an internal precursor that does not resolve #2292 and follow it immediately with the packaged integration.

The current CI is green, but it does not execute the mutating Windows restart path; keep the PR blocked until an external-terminal Windows smoke test proves all old package PIDs exit and the app is relaunched through the discovered AUMID.

PR #2298, opened 2026-08-21
fix(claude): warm empty Desktop-3P alias registry on first /v1/messages after restart

### LINKED ISSUE #2297: Desktop-3P hashed model ids fail on the first /v1/messages after a proxy restart (empty in-memory alias registry)
## Summary

After the proxy restarts, the **first** Anthropic-protocol chat request that uses a hashed Desktop-3P discovery id (`claude-opus-4-8-<code>`) fails with an upstream error like:

```
API Error: 400 {'error': 'anthropic_messages: Invalid model name passed in model=claude-opus-4-8-20260304. Call `/v1/models` to view available models for your key.'}
```

The same request succeeds as soon as any client performs an anthropic-flavor `GET /v1/models` against the running process.

## Root cause

The registry that decodes the hashed aliases (`desktop3pRegistry` in `src/claude/desktop-3p.ts`) is a plain in-memory `Map`, and it is only (re)built inside the anthropic-flavor branch of the `GET /v1/models` handler (`buildDesktop3pRegistry(...)` in `src/server/index.ts`).

Clients that integrate the proxy (host apps embedding Claude Code, Claude Desktop 3P configs, anything that pins a model id after one discovery pass) cache the hashed id across proxy restarts. When such a client replays the cached id as the first request after a restart:

1. the registry is empty, so `resolveDesktop3pAlias` returns `null`;
2. `resolveInboundModel` falls through — the id matches `^claude-opus-4` so it takes classifier affinity / raw passthrough;
3. the raw hashed id reaches an upstream `anthropic_messages` provider, which rejects it with "Invalid model name".

The failure is invisible in casual testing because anything that lists models first (dashboards, connection tests, a fresh Claude Code launch doing discovery) rebuilds the registry as a side effect — which makes it look like "test passes, real chat fails".

## Reproduction

1. Configure any routed provider and let a client cache a hashed discovery id (`claude-opus-4-8-<code>`).
2. Restart the opencodex process.
3. Without calling `GET /v1/models`, POST `/v1/messages` with the cached id.
4. → 400 "Invalid model name passed in model=claude-opus-4-8-...". Calling `GET /v1/models` once (with `anthropic-version` header) and retrying makes the same request succeed.

Observed on v2.28.0 (Windows 11, bundled Bun 1.3.14); the relevant code is unchanged on current `dev`.

## Suggested fix

Warm the registry when a `/v1/messages` (or `count_tokens`) request arrives while the registry is empty — e.g. a loopback anthropic-flavor `GET /v1/models`, which is semantically identical to the client refreshing discovery first. PR incoming.

ppvia:
Fixes #2297.

## Problem

The Desktop-3P alias registry (`desktop3pRegistry`) that decodes hashed discovery ids (`claude-opus-4-8-<code>`) is in-memory and only rebuilt by an anthropic-flavor `GET /v1/models`. A client that cached such an id from a previous process replays it as the **first** request after a proxy restart; the alias cannot decode, `resolveInboundModel` falls through to classifier affinity / raw passthrough, and the upstream rejects it:

```
400 {'error': 'anthropic_messages: Invalid model name passed in model=claude-opus-4-8-20260304. ...'}
```

Anything that happens to list models first (dashboard, connection test) hides the bug by rebuilding the registry as a side effect, so it presents as "connection test passes, real chat fails".

## Change

- `src/server/index.ts` — when a `POST /v1/messages` or `/v1/messages/count_tokens` arrives while the registry is empty, warm it once via a loopback anthropic-flavor `GET /v1/models` (forwarding the caller's auth headers, since discovery requires admission). Semantically identical to the client refreshing discovery before chatting; fires at most once per process in practice, `try/catch`-guarded so a discovery failure falls through to the existing resolution path unchanged.
- `src/claude/desktop-3p.ts` — export `desktop3pRegistryIsEmpty()` so the inbound route can observe the fresh-boot state.
- `tests/claude-messages-endpoint.test.ts` — regression test: empties the registry (simulating a fresh boot), POSTs a hashed alias as the first request, and asserts the mock upstream receives the **decoded** route (`test-model`), not the raw hash.

Considered instead: extracting the registry-build inputs (`fetchAllModels` + entitlements + the visibility/ordering derivation in the models handler) into a shared helper callable from the messages branch. That duplicates a ~60-line derivation chain across two call sites; the loopback reuses the whole handler and cannot drift from it. Happy to rework if maintainers prefer eager build at startup.

## Validation

- `bun test tests/claude-messages-endpoint.test.ts` — 43 pass, 0 fail (includes the new test)
- With the `src/` change stashed, the new test **fails** (`captured[0].model` is the raw hash) — confirms it guards the defect
- `bun test tests/desktop-3p.test.ts tests/claude-models-discovery.test.ts tests/claude-inbound.test.ts` — 59 pass, 0 fail
- `bun run typecheck` — clean

Environment: reproduced on v2.28.0 (Windows 11, bundled Bun 1.3.14); patch developed and tested on current `dev`.

## Review readiness checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability after server restarts by automatically restoring Desktop 3P model alias information when processing the first message request.
  - Cached hashed model aliases now decode correctly instead of failing after a restart.

- **Tests**
  - Added regression coverage for successful streaming and alias resolution following a restart.

### DIFF
--- a/src/claude/desktop-3p.ts
+++ b/src/claude/desktop-3p.ts
@@ -312,6 +312,11 @@ export function generateDesktop3pModels(
   return models;
 }
 
+/** True while no Desktop registry has been generated in this process (fresh boot). */
+export function desktop3pRegistryIsEmpty(): boolean {
+  return desktop3pRegistry.size === 0;
+}
+
 /** Resolve an alias using the most recently generated Desktop model registry. */
 export function resolveDesktop3pAlias(alias: string): string | null {
   return desktop3pRegistry.get(alias) ?? null;
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -1283,6 +1283,25 @@ export function startServer(port?: number, deps: StartServerDeps = {}): Server<W
         });
       }
 
+      // Desktop-3P alias self-heal: the registry that decodes hashed discovery ids
+      // (claude-opus-4-8-<code>) is in-memory and only rebuilt by an anthropic-flavor
+      // GET /v1/models. A client that cached such an id from a previous process can
+      // replay it as the FIRST request after a restart; with an empty registry the
+      // alias cannot decode and the request misroutes (classifier affinity or raw
+      // passthrough → upstream "Invalid model name"). Warm the registry once via
+      // loopback discovery — semantically the same as the client refreshing models.
+      if ((url.pathname === "/v1/messages" || url.pathname === "/v1/messages/count_tokens") && req.method === "POST") {
+        const { desktop3pRegistryIsEmpty } = await import("../claude/desktop-3p");
+        if (desktop3pRegistryIsEmpty()) {
+          const warmHeaders = new Headers({ "anthropic-version": "2023-06-01" });
+          const warmAuth = req.headers.get("authorization");
+          const warmKey = req.headers.get("x-api-key");
+          if (warmAuth) warmHeaders.set("authorization", warmAuth);
+          if (warmKey) warmHeaders.set("x-api-key", warmKey);
+          try { await fetch(new URL("/v1/models", url.origin), { headers: warmHeaders }); } catch { /* fall through to existing resolution */ }
+        }
+      }
+
       // Anthropic Messages inbound (Claude Code). count_tokens FIRST (longer path).
       // Claude Code posts `/v1/messages?beta=true` — pathname match ignores the query (003 G9).
       if (url.pathname === "/v1/messages/count_tokens" && req.method === "POST") {
--- a/tests/claude-messages-endpoint.test.ts
+++ b/tests/claude-messages-endpoint.test.ts
@@ -1571,3 +1571,53 @@ test("count_tokens is CJK-aware: Korean body counts more tokens than equal-lengt
     await server.stop(true);
   }
 });
+
+test("first /v1/messages after a restart self-heals an empty Desktop-3P registry (cached hashed id)", async () => {
+  const { server: upstream, captured } = mockChatUpstreamCapturing();
+  const baseUrl = `${upstream.url.toString().replace(/\/$/, "")}/v1`;
+  saveConfig({
+    port: 0,
+    defaultProvider: "mock",
+    providers: {
+      mock: {
+        adapter: "openai-chat",
+        baseUrl,
+        apiKey: "k",
+        allowPrivateNetwork: true,
+        liveModels: false,
+        models: ["test-model"],
+      },
+    },
+  } as OcxConfig);
+  const server = startServer(0);
+  try {
+    const { buildDesktop3pRegistry, desktop3pAlias } = await import("../src/claude/desktop-3p");
+    // Post-restart state: no discovery GET has run in this process, so the
+    // in-memory registry is empty while the client replays the hashed id it
+    // cached from the previous process.
+    buildDesktop3pRegistry([], []);
+    const alias = desktop3pAlias("mock", "test-model");
+    const response = await fetch(new URL("/v1/messages?beta=true", server.url), {
+      method: "POST",
+      headers: {
+        "content-type": "application/json",
+        "x-api-key": "placeholder",
+        "anthropic-version": "2023-06-01",
+      },
+      body: JSON.stringify({
+        model: alias,
+        max_tokens: 128,
+        stream: true,
+        messages: [{ role: "user", content: "hi" }],
+      }),
+    });
+    expect(response.status).toBe(200);
+    const text = await response.text();
+    expect(text).toContain("Hello");
+    // The upstream must see the decoded route, not the raw hashed alias.
+    expect(captured[0]?.model).toBe("test-model");
+  } finally {
+    await server.stop(true);
+    upstream.stop(true);
+  }
+}, { timeout: SERVER_BUDGET_MS });


Ingwannu (changes requested):
Reviewed exact head `38888e3d5bd7f6c364fffb5765d450bf021e055b`. I am requesting changes; the unresolved automated comment is valid, and the impact is broader than a possible warm-up stall.

The new branch runs before `/v1/messages` authentication and builds the self-fetch destination from inbound `url.origin`. It then copies the raw inbound `Authorization` and `x-api-key` headers to that destination. Because Bun derives the request origin from the inbound `Host`, this creates a Host-header-controlled outbound request and credential-forwarding boundary.

I reproduced the exact head with a capture server and a request sent to the local OCX listener while setting `Host` to the capture server. The capture server received:

- path: `/v1/models`
- `Authorization: Bearer client-secret`
- `x-api-key: client-api-secret`

The reproduction passed 1 test with 3 assertions. This is reachable before `resolveApiAuth`, so the warm-up should not be allowed to make any outbound request using unvalidated request routing or credentials.

Please:

1. run the warm-up only after the ordinary `/v1/messages` authentication and origin checks have succeeded;
2. target a server-controlled loopback URL using the actual bound port, or preferably call the internal model/registry refresh path directly; never use inbound `url.origin`;
3. add a bounded `AbortSignal.timeout` so the first real request cannot hang indefinitely;
4. add regressions proving a hostile `Host` cannot receive a request or auth headers, a stalled warm-up is bounded, and the normal cached-alias self-heal still works.

The PR is also still draft with 0/4 readiness boxes checked. Keep it unmerged until this boundary is fixed, the unresolved thread is closed, the branch is marked ready, and exact-head CI is green. The automated finding was used as input and independently reproduced rather than accepted at face value.

lidge-jun:
## 리뷰 · 우선순위 66 / 80

지금 `dev` HEAD `c0cbe494e`에서 재현이 코드랑 맞음. `src/claude/desktop-3p.ts:145` `desktop3pRegistry`가 프로세스 로컬 Map. `src/server/index.ts:898-993` GET `/v1/models`가 `anthropic-version`이면 `buildDesktop3pRegistry`를 탐. `src/claude/inbound.ts:59-73` `resolveInboundModel`이 `resolveDesktop3pAlias` 미스면 원본 해시를 그대로 넘김. 해시 `claude-opus-4-8-<code>`가 업스트림에 가서 400. 대시보드가 먼저 리스트하면 레지스트리가 채워져서 채팅은 됨. 연결 테스트 OK, 첫 채팅 실패가 그거임. 드래프트. 닫을 중복 아님. #2297은 github-actions가 not_planned로 닫음. 버그는 `dev`에 그대로임. 이슈를 다시 열지 말고 이 PR이 고치게 해라.

핵심이 `src/server/index.ts` 루프백. `/v1/messages`랑 `/v1/messages/count_tokens` POST에서 레지스트리 size 0이면 `fetch(new URL("/v1/models", url.origin))`. `anthropic-version` + 호출자 `authorization`/`x-api-key`를 복사. GET `/v1/models` `:902-913`이 `resolveApiAuth` 다음에 `fetchAllModels`+엔타이틀먼트 전체 디스커버리를 탐. 핸들러 전체를 HTTP로 재사용한 거임. 작성자가 말한 공유 헬퍼 대신. 테스트가 `buildDesktop3pRegistry([], [])`로 비우고 해시 알리아스를 첫 POST로 넣음. 업스트림이 `test-model`을 보면 통과. 그 한 장은 맞음.

구멍. `url.origin`이 인바운드 `Host`임. Bun.serve `req.url`이 그 헤더를 씀. 리버스 프록시/로드밸런서면 루프백이 이 프로세스가 아님. `Host`를 공격자가 바꾸면 호출자 API 키를 그 origin으로 들고 나감. 워밍이 `resolveApiAuth`/`isAllowedRequestOrigin`보다 앞임. 메시지 핸들러(`index.ts:1286`)보다 위. 미인증 POST도 디스커버리를 띄움. `desktop3pRegistryIsEmpty`가 size===0. 네이티브 Anthropic만 있으면 `desktop-3p.ts:254-266`이 레지스트리에 안 넣음. 영원히 빔. 매 메시지마다 `fetchAllModels`. 503 `catalog_busy`는 catch로 삼킴. 레지스트리 그대로. 그 다음 메시지가 같은 400.

`types.ts`/`config.ts` 안 만짐. 스플릿 안 씹힘. 리베이스하지 말고 닫으라는 케이스 아님. #2188 L1–L9 사이드카 이미 `dev`. `x_search` 넣지 말 것. Grok OAuth Chat 기본(#2255)/GUI 옵트인 Responses(#2266)/#2283이랑 다른 레인임. 프리뷰 배포 아님. #2292/#2293 윈도우 피커/풀 재시작이랑 안 겹침. 카탈로그는 그대로 Ox Alpha `x-preview-f-free` + `deepseek-v4-flash-vision-exp`. v2.29.0 태그됨. v2.30.0-preview.20260821 있음. 재시작 직후 Desktop 첫 턴이 유저 보이는 400이라 66. 루프백은 그 고정의 구현이 아님.

**해결방안:** HTTP 루프백 버려라. `index.ts:990`이 이미 쓰는 `buildDesktop3pRegistry(desktopNativeSlugs, goOrdered.map(...), desktopProfile)`를 메시지 경로에서 직접 호출. 입력을 GET `/v1/models`랑 같은 헬퍼로 뽑아 드리프트 방지. 워밍됨 플래그를 size===0이랑 분리. 네이티브만 있어도 한 번만. 인증/오리진 가드 뒤에서. `Host`로 fetch 금지. 키를 다른 origin으로 들고 나가지 말 것. 테스트에 (1) 네이티브만 있는 구성에서 매 턴 fetch 안 함 (2) 워밍 실패해도 기존 400 경로 (3) Host 스푸핑이 외부 fetch를 안 만듦. #2297은 닫힌 채. 이 PR이 버그를 닫음. 스플릿이 `index.ts`/`desktop-3p.ts`를 옮기면 리베이스하지 말고 닫고 다시 짜라. 지금은 그 정도 아님.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
Closing this for a security reason, not a quality one — thank you for finding the bug, which is real and still on `dev`.

**The bug is confirmed.** `src/claude/desktop-3p.ts` holds a process-local `Map`, and `src/server/index.ts` only rebuilds it on an anthropic-flavor `GET /v1/models`. A cached `claude-opus-4-8-<code>` sent as the first POST after a restart still cannot decode. That is worth fixing.

**The implementation cannot ship as written.** The warm-up does `fetch(new URL("/v1/models", url.origin))` *before* `resolveApiAuth` / `isAllowedRequestOrigin`, and copies the inbound `Authorization` and `x-api-key` onto that request. Bun derives `url.origin` from the `Host` header, so the destination is caller-controlled: a POST carrying a hostile `Host` delivers the client's bearer token and API key to an arbitrary server. This was reproduced at `38888e3d` — a capture server received `Authorization: Bearer client-secret` and `x-api-key: client-api-secret`.

Three further gaps, independent of that one:

- Native Anthropic rows are deliberately excluded from the decode registry (`desktop-3p.ts:254`), so a native-only catalog leaves `size === 0` permanently and every later `/v1/messages` repeats the fetch.
- Unauthenticated POSTs also take the warm-up branch.
- The fetch has no timeout and swallows failures, so an upstream stall delays the real client and a 503 `catalog_busy` leaves the next turn on the same 400.

**What a mergeable version looks like:** warm up *after* admission, reach the registry in-process via `buildDesktop3pRegistry` (or a server-controlled loopback address that never reads `Host`) instead of re-entering the HTTP surface, never forward client credentials to a derived origin, add `AbortSignal.timeout`, and include a Host-spoof regression test. The earlier `CHANGES_REQUESTED` review asked for the same things and they are still open; the branch is also 170 commits behind `dev`.

Please reopen or send a follow-up PR with that shape — the underlying fix is wanted.

PR #2428, opened 2026-08-23
chore(test): queue only full-suite runs behind the exclusive-run lock

olddonkey:
> **Stacked on #2427**, which introduces `isFullSuiteRun`. The first commit here is that PR; review the second. The diff collapses once #2427 lands.

## What changes

`waitForExclusiveRun` currently makes **every** invocation wait, including a single-file run. Behind a full suite that is a multi-minute wait for a check that takes seconds — the common case while implementing.

This narrows the wait to full-suite runs.

## What the lock actually guards

**CPU contention, not shared state.** Each invocation gets its own `mkdtemp` sandbox (`createIsolatedTestEnvironment`), so concurrent runs do not collide on `HOME`, `OPENCODEX_HOME` or `CODEX_HOME`. The incident the lock was written for is recorded in its own comment: a suite that normally took ~210 s took **26 minutes** against a runner an earlier session had left behind, and neither process said anything, so the slowdown read as a hang.

## The trade-off, stated plainly

With `--parallel` (#2427) a full run already saturates the machine, so a focused run started alongside one **does** slow it. This is a judgement call, not a free win:

- **for**: blocking every focused check behind a multi-minute suite costs more in daily use than the contention it avoids, and a focused run is bounded at seconds;
- **against**: it removes a guard that exists precisely because contention is invisible and reads as a hang.

Full suites still queue behind each other, which is the case the lock was written for.

This is separated from #2427 exactly because it is a behavior decision rather than a fix — I did not want it riding along inside a PR whose justification is "the suite does not finish".

## Tests

None added. The lock's behavior is process-level (`pgrep` for competing runners plus a wait loop) and I did not find a way to pin it that would not amount to asserting the implementation back to itself. Flagging that rather than adding a test that only restates the code.

## Review readiness checklist

## Summary by CodeRabbit

* **Improvements**
  * Test runs now use isolated, parallel execution by default for faster and more reliable full-suite testing.
  * Existing file filters and explicitly provided concurrency settings are preserved.
  * Option-only test arguments are handled correctly during full-suite runs.
* **Documentation**
  * Added guidance clarifying where parallel test execution is configured.

### DIFF
--- a/bunfig.toml
+++ b/bunfig.toml
@@ -5,6 +5,7 @@
 # so a bare `bun test` — or `bun test tests/` (a substring filter that also matches
 # devlog/opencode-cursor/tests/) — drags them in and reports hundreds of spurious failures.
 # `root` pins discovery to ./tests so every invocation stays on the real suite.
+# File-level `--parallel` has no bunfig key; `scripts/test.ts` passes it for `bun run test`.
 # The npm script already uses `bun test ./tests/`; this makes a bare `bun test` behave the same.
 [test]
 root = "tests"
--- a/scripts/test.ts
+++ b/scripts/test.ts
@@ -59,6 +59,33 @@ export function createIsolatedTestEnvironment(
   };
 }
 
+function hasCliFlag(requested: string[], name: string): boolean {
+  return requested.some(arg => arg === name || arg.startsWith(`${name}=`));
+}
+
+/** True for a filter-less `bun run test`. `--timeout` / `--dots` / `--parallel=N` still count. */
+function isFullSuiteRun(requested: string[]): boolean {
+  return !requested.some(arg => arg !== "-" && !arg.startsWith("-"));
+}
+
+/**
+ * Default `bun test` argv for this repo.
+ *
+ * `--isolate` keeps a fresh global per file, and is the substring the exclusive-run pgrep
+ * matches. `--parallel` is what makes the suite finishable: with isolate alone Bun re-evaluates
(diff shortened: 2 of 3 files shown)


olddonkey:
Closing this — **its stated rationale is wrong**, and not in a way a reword fixes.

I justified narrowing the lock with "it guards CPU contention, not shared state, because each invocation gets its own `mkdtemp` sandbox". Review checked that and it does not hold: the sandbox isolates the `HOME` / `OPENCODEX_HOME` / `CODEX_HOME` **environment**, but individual tests can and do write to fixed paths **inside the repo**, which no sandbox covers.

The concrete counter-example: `tests/codex-cooldown-recovery.test.ts:29` uses a fixed `join(import.meta.dir, ".tmp-codex-cooldown-recovery-test")`, `rmSync`s it during setup, and then points `OPENCODEX_HOME` at it — overriding the per-run sandbox with a shared path. Two concurrent runs of that one file delete each other's fixture mid-run.

`tests/fixture-dir-uniqueness.test.ts:71-77` already encodes this exact hazard, and says so: deriving a per-run directory "also makes two concurrent runs of the SAME file safe — something a rename alone would miss".

So the lock is doing more than queueing for CPU, and removing the wait for focused runs is not the free win this PR claimed. The daily cost it targets is real, but the prerequisite is making the remaining fixed-fixture paths per-run first. Not worth keeping open as a half-argued change in the meantime.

#2427 (the `--parallel` fix) and #2429 (`test:changed`) are unaffected and stay open.

Issue #2499, opened 2026-08-24
[Bug][Windows]: v2 guidance catalog-state probe still costs ~7.5s per turn (the #1876 async fix removed the event-loop block, not the request latency)


Vladimir321123:
## Client or integration

Codex App, Codex CLI, and any ACP client (observed via `@agentclientprotocol/codex-acp` inside Obsidian Copilot).

## Area

Proxy and routing / Platform (Windows)

## Summary

#1852 reported that the Windows catalog-state enumeration blocked the Bun event loop. #1876 fixed that: `collectCodexAppServerCatalogStateForRequest()` now uses `listWindowsSnapshotsAsync()` with in-flight dedup and a short cache, and `/healthz` stays responsive.

The **latency** half of the same code path is still there, and it is now the dominant cost of every multi-agent v2 turn. On 2.32.0 the enumeration still runs on the request hot path, and on this machine it takes ~7.5s — longer than `CATALOG_STATE_TTL_MS`, so the cache almost never serves a request.

Measured on this installation, same account, same trivial prompt ("reply with PONG"), 4 runs each, interleaved:

| Route | median | runs (ms) |
|---|---:|---|
| through proxy | 11296 ms | 10844, 11245, 11296, 12689 |
| direct to ChatGPT (no proxy) | 5151 ms | 3818, 4654, 5151, 5836 |
| through proxy, `multiAgentGuidanceEnabled: false` | **3424 ms** | 3118, 3292, 3424, 5698 |

So the proxy was ~7s slower than no proxy at all; with guidance off it is *faster* than direct (the WS upstream transport pays off as documented in `ws-upstream.ts`).

## Evidence that the cost is the catalog probe, not the network

1. The proxy's own `usage.jsonl` already shows it. Representative slow row:

```
durationMs: 10175, firstOutputMs: 9969
attempts: [{ ordinal: 1, adapter: "openai-responses", status: 200,
             durationMs: 2024, firstOutputMs: 1818, sendCount: 1, recoveryKinds: [] }]
```

The upstream attempt took 2.0s; the request took 10.2s. ~8s is spent before the attempt starts.

2. CPU time of the proxy process during those 8 seconds is 0–31 ms per 900 ms sample — it is waiting on a child process, not computing.

3. Temporary `console.log` timestamps in `src/server/responses/core.ts` localise it exactly:

```
core: enter                       t0
step: before body read            t0 +8 ms
step: before route normalization  t0 +17 ms
[...]

PR #2542, opened 2026-08-25
fix(codex): serve the last catalog reading while the Windows probe refreshes

ntdatt812:
Refs #2499. Takes suggestion **2** from that issue — serve stale while revalidating — and leaves suggestion 1 alone; reasoning for that split at the bottom.

## The shape of it

`collectCodexAppServerCatalogStateForRequest` awaits a probe whose cost can exceed its own TTL. `CATALOG_STATE_TTL_MS` is 5s; the Windows probe pays `Invoke-CimMethod GetOwner` once per candidate process. When the probe is slower than the window it populates, the window expires before it can serve much, and the miss lands on a turn.

**I could not reproduce the reporter's 7.5s end to end** — this box has 2 Codex candidates, so the whole enumeration is ~1.39s. What I could reproduce is the per-call cost that produces their number. Windows 11 26200, 371 processes with a readable command line, `powershell.exe` from System32, 3 runs each:

| step | cost |
|---|---|
| PowerShell startup (`-NoProfile -Command exit`) | 152 / 152 / 166 ms |
| `Get-CimInstance Win32_Process` + the command-line filter | 497 / 549 / 556 ms |
| the full `windowsSnapshotPowerShellCommand()`, 2 candidates | 1350 / 1387 / 1426 ms |

The remainder is 2 × `GetOwner` ≈ 870 ms, so **~435 ms per candidate**. The issue reports 43887 ms for 93 processes — 472 ms per call. Same cost, different candidate count, which is why one machine sees 1.4s and another 7.5s.

## What changed

`servableStale`: when the cached entry is expired but its **generation still matches**, hand it back immediately and let the refresh it just started run behind it.

The generation is what makes that safe, and it is worth being precise about. Every ocx catalog write calls `resetCodexAppServerCatalogStateCache()` (`app-server-restart-service.ts:114`), which advances `requestCatalogStateGeneration` **and nulls the cache**. So a generation match is a proof that no catalog write has landed since the reading was taken — the exact invalidation the existing in-flight guard already treats as decisive when it degrades a pre-write result to `unknown`.

What an expired reading can still miss is a server that started or stopped meanwhile. A server that started *after* the reading is newer than the catalog, which is the `fresh` the entry already reports. That leaves a stopped server, and `fresh` for a machine with one fewer app-server is not a claim that misleads guidance.

Two bounds, both deliberate:

- **`unknown` is never served this way.** It is a failure to observe, not an observation. It already has a 250 ms window for precisely that reason, and re-serving it would keep guidance suppressed on the strength of a reading that never saw anything.
- **60s ceiling.** If the refresh keeps failing, a reading that old stops being evidence about the machine, and waiting for a real one beats answering fast and wrong.

The cold path is unchanged: with nothing cached, the first turn still waits for the probe.

## Tests

Four, in `tests/codex-app-server-processes.test.ts`, all with a controllable clock and a probe that never settles on its own — no real timers, no platform dependency.

| test | what it pins |
|---|---|
| the first turn waits, later turns do not, and N turns cost one enumeration | the reporter's item 4, verbatim |
| the refresh it triggers is what makes the next reading current | the stale answer is not a dead end |
| a reading older than the bound is not served | the 60s ceiling |
| `unknown` is never served stale | the exclusion |

Mutation-checked. Each mutation was verified to have actually applied (occurrence counts) before the run:

| mutation | caught by |
|---|---|
| drop `servableStale ??` from both returns (2 → 0 occurrences) | tests 1 and 2 |
| drop `state !== "unknown"` | test 4 |
| drop the `CATALOG_STATE_MAX_STALE_MS` bound | test 3 |

Tests 3 and 4 do not detect the first mutation, and that is correct — they assert the cases where the value is *not* served, which reverting to always-await also satisfies. Saying so rather than implying all four guard everything.

## Local results

`bun test tests/codex-app-server-processes.test.ts`:

| tree | result |
|---|---|
| `origin/dev` @ `3e3a028fe`, no changes | 47 tests, **46 pass / 1 fail** |
| this branch | 51 tests, **50 pass / 1 fail** |

+4 passing, and the same single failure: `warnIfStaleCodexAppServersAfterStartupWrite (#1046) > a defaulted read is memoized, and invalidation is what clears it`. It is pre-existing on `dev` on this machine — I measured it on a clean tree before and after, and I flagged the same test in #2485. Since this diff touches the caching in that very module I re-baselined it rather than citing the old measurement.

Widened to the neighbouring files (`codex-app-server-path-spaces`, `api-catalog-route`, `cli-catalog-prewarm`): **58 pass / 1 fail**, same one.

`bun x tsc --noEmit`: exit 0.

One thing worth admitting: my first version of these tests never ran at all. I appended the block at file scope and reused `APP_SERVER_CMD`, which is declared inside another `describe`; the suite reported the same 46/47 as the baseline and a quiet `ReferenceError` between tests. The pass count not moving is what caught it.

## Why not suggestion 1 in this PR

Dropping the per-process `GetOwner` fan-out is the bigger win and I did prototype it — `Win32_SessionProcess` joined to `Win32_LoggedOnUser` maps every PID to an owner in ~460 ms total on this box, against ~435 ms *per candidate* today.

I am not shipping it here. `GetOwner` returning `ReturnValue=2` on a process is currently what makes the enumeration report `__OCX_ENUM_INCOMPLETE__`, and the association path answers a subtly different question — it is missing rows rather than denied, and `Win32_Process.CommandLine` is itself only readable for processes you own, so the two disagree about what "could not verify the owner" means. Getting that wrong loosens a current-user scoping contract that the restart path depends on. It deserves its own PR with its own ground-truth comparison, not a rider on a latency fix.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows catalog-state loading by returning recently cached results while updates refresh in the background.
  * Prevented unnecessary blocking when valid cached information is available.
  * Limited stale results to a maximum of 60 seconds.
  * Preserved valid cached information when a refresh fails.
  * Ensured outdated or unknown catalog states are refreshed appropriately.
  * Prevented multiple simultaneous refresh requests.

### DIFF
--- a/src/codex/app-server-processes.ts
+++ b/src/codex/app-server-processes.ts
@@ -750,6 +750,11 @@ const CATALOG_STATE_UNKNOWN_TTL_MS = 250;
  * previous reading immediately keeps that cost off the turn without pretending it is
  * fresh -- the refresh it triggers is what makes the next reading current.
  *
+ * Measured from expiry rather than from when the reading was taken, so a `fresh`
+ * entry stays servable for its own TTL plus this bound. Anchoring it to expiry keeps
+ * the stale window independent of the TTL -- a cap on total age would quietly turn
+ * this path off if the TTL were ever raised past it.
+ *
  * Bounded rather than unlimited: if the refresh keeps failing, an observation this
  * old stops being evidence about the machine and it is better to wait for a real one.
  * `unknown` is never served this way -- it is a failure to observe, not an
@@ -895,7 +900,7 @@ export async function collectCodexAppServerCatalogStateForRequest(
   // this entry already says.
   const servableStale = cached
     && cached.status.state !== "unknown"
-    && now - cached.atMs < CATALOG_STATE_MAX_STALE_MS
+    && now - cached.atMs < catalogStateTtlMs(cached.status.state) + CATALOG_STATE_MAX_STALE_MS
     ? cached.status
     : null;
   if (requestCatalogStateFlight
--- a/src/codex/app-server-processes.ts
+++ b/src/codex/app-server-processes.ts
@@ -740,6 +740,23 @@ const CATALOG_STATE_TTL_MS = 5_000;
  */
 const CATALOG_STATE_UNKNOWN_TTL_MS = 250;
 
+/**
+ * How long a real observation may still be SERVED after it expires, while a refresh
+ * runs behind it (#2499).
+ *
+ * The probe is advisory and, on Windows, slow: `Invoke-CimMethod GetOwner` costs
+ * ~0.4s per candidate process, so a cold probe routinely outlives the 5s TTL and
+ * every turn that misses the cache pays for it on the request path. Serving the
+ * previous reading immediately keeps that cost off the turn without pretending it is
+ * fresh -- the refresh it triggers is what makes the next reading current.
+ *
+ * Bounded rather than unlimited: if the refresh keeps failing, an observation this
+ * old stops being evidence about the machine and it is better to wait for a real one.
+ * `unknown` is never served this way -- it is a failure to observe, not an
+ * observation, and it already has its own short window for exactly that reason.
+ */
+export const CATALOG_STATE_MAX_STALE_MS = 60_000;
+
 export function catalogStateTtlMs(state: CodexAppServerCatalogState): number {
   return state === "unknown" ? CATALOG_STATE_UNKNOWN_TTL_MS : CATALOG_STATE_TTL_MS;
 }
@@ -853,16 +870,30 @@ export async function collectCodexAppServerCatalogStateForRequest(
     catalogMtimeMs: io.catalogMtimeMs,
     now: io.now,
   };
-  if (requestCatalogStateCache
+  const cached = requestCatalogStateCache
     && requestCatalogStateCache.generation === generation
     && sameRequestCatalogStateIdentity(requestCatalogStateCache.identity, identity)
-    && now - requestCatalogStateCache.atMs < catalogStateTtlMs(requestCatalogStateCache.status.state)) {
-    return requestCatalogStateCache.status;
+    ? requestCatalogStateCache
+    : null;
+  if (cached && now - cached.atMs < catalogStateTtlMs(cached.status.state)) {
+    return cached.status;
   }
+  // An expired real reading is still worth handing back while the refresh runs. It
+  // cannot have been invalidated by an ocx catalog write: every such write calls
+  // `resetCodexAppServerCatalogStateCache`, which advances the generation and drops
+  // this entry, so a generation match means no write has landed since it was taken.
+  // What it can miss is an app-server that started or stopped meanwhile -- and a
+  // server started after the reading is newer than the catalog, which is the `fresh`
+  // this entry already says.
+  const servableStale = cached
+    && cached.status.state !== "unknown"
+    && now - cached.atMs < CATALOG_STATE_MAX_STALE_MS
+    ? cached.status
+    : null;
   if (requestCatalogStateFlight
     && requestCatalogStateFlight.generation === generation
     && sameRequestCatalogStateIdentity(requestCatalogStateFlight.identity, identity)) {
-    return requestCatalogStateFlight.promise;
+    return servableStale ?? requestCatalogStateFlight.promise;
   }
 
   const refresh = async (): Promise<CodexAppServerCatalogStatus> => {
@@ -934,7 +965,9 @@ export async function collectCodexAppServerCatalogStateForRequest(
   });
   flight = { generation, identity, promise };
   requestCatalogStateFlight = flight;
-  return flight.promise;
+  // `promise` already absorbs its own failures, so leaving it unawaited here cannot
+  // surface as an unhandled rejection; the next caller picks up whatever it stored.
+  return servableStale ?? flight.promise;
 }
 
 /**
--- a/src/codex/app-server-processes.ts
+++ b/src/codex/app-server-processes.ts
@@ -852,6 +852,14 @@ export function collectCodexAppServerCatalogState(
  * - 선택한 방식: retain the synchronous API and use async PowerShell plus an identity-scoped in-flight refresh, short cache, and invalidation generation only for Windows requests.
  * - 다른 대안 대신 이 방식을 선택한 이유: it fixes unrelated `/healthz` starvation without widening the process-matching or restart contract.
  * - 장점, 단점 및 영향: concurrent turns share one CIM walk, invalidated pre-write results cannot repopulate the cache, and the event loop stays responsive; a cold v2 turn can still await the bounded advisory probe.
+ *
+ * [Decision Log · #2499]
+ * - 목적과 의도: a cold probe outlives its own 5s TTL on Windows (~435ms per candidate process for `Invoke-CimMethod GetOwner`), so the cache expires before it can serve and the miss lands on a turn.
+ * - 기존 구현 및 제약 조건: the reading is advisory, and only `fresh` authorizes positive guidance (`src/server/responses/collaboration.ts`); `unknown` is a failure to observe rather than an observation.
+ * - 검토한 주요 대안: drop the per-process GetOwner fan-out (issue suggestion 1), or widen the TTL past the probe duration (suggestion 3).
+ * - 선택한 방식: serve an expired reading immediately when its generation still matches, bounded by `CATALOG_STATE_MAX_STALE_MS`, never for `unknown`, and refresh behind it; a failed refresh no longer evicts the reading it was refreshing.
+ * - 다른 대안 대신 이 방식을 선택한 이유: the fan-out change alters what "could not verify the owner" means for the current-user scoping contract and needs its own ground-truth comparison; a wider TTL still pays the probe on every human-paced turn.
+ * - 장점, 단점 및 영향: after the first probe the request path never waits; a server that stopped between readings can be described as `fresh` for up to the stale bound; a catalog write still invalidates immediately through the generation, and the cold path is unchanged.
  */
 export async function collectCodexAppServerCatalogStateForRequest(
   io: CodexAppServerProcessIo = {},
@@ -951,7 +959,18 @@ export async function collectCodexAppServerCatalogStateForRequest(
     if (requestCatalogStateGeneration !== generation) {
       return { state: "unknown" as const, processes: [], catalogMtimeMs: null };
     }
-    if (requestCatalogStateFlight === flight) {
+    // A refresh that failed must not evict a real reading. Before this function
+    // served stale entries, caching `unknown` cost at most the 250ms that state
+    // is allowed to live. Now that an expired observation is what callers are
+    // handed, overwriting one with `unknown` would take the answer AWAY from
+    // them on a transient failure -- `unknown` is not servable, so the next
+    // caller waits for a probe instead of getting the reading it would have had.
+    // Keep the observation and let its own age retire it.
+    const wouldEvictAnObservation = status.state === "unknown"
+      && requestCatalogStateCache?.generation === generation
+      && requestCatalogStateCache.status.state !== "unknown"
+      && sameRequestCatalogStateIdentity(requestCatalogStateCache.identity, identity);
+    if (requestCatalogStateFlight === flight && !wouldEvictAnObservation) {
       requestCatalogStateCache = {
         generation,
         identity,
--- a/tests/codex-app-server-processes.test.ts
+++ b/tests/codex-app-server-processes.test.ts
@@ -1187,8 +1187,16 @@ describe("request-path catalog state serves stale while revalidating (#2499)", (
     releases[0]([...APP_SERVER]);
     await expect(cold).resolves.toMatchObject({ state: "fresh" });
 
-    // Past the point where the reading stops being evidence about the machine.
-    clock.ms += CATALOG_STATE_MAX_STALE_MS;
+    // One tick inside the bound, which runs from expiry rather than from when the
+    // reading was taken: still served, still without waiting.
+    clock.ms += catalogStateTtlMs("fresh") + CATALOG_STATE_MAX_STALE_MS - 1;
+    const last = collectCodexAppServerCatalogStateForRequest(io);
+    expect(await settledWithoutTheProbe(last)).toMatchObject({ state: "fresh" });
+
+    // One tick past it, where the reading stops being evidence about the machine.
+    // That call joins the refresh the previous one started rather than waiting on a
+    // second enumeration, so releasing that one probe is what settles it.
+    clock.ms += 1;
     const tooOld = collectCodexAppServerCatalogStateForRequest(io);
     expect(await settledWithoutTheProbe(tooOld)).toBe("waited");
     releases[1]([...APP_SERVER]);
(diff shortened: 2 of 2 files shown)


ntdatt812:
Checklist ticked, with the numbers behind each box rather than the box alone.

**Rebased first.** `dev` had moved `3e3a028fe → 858352ad6` while this sat in the readiness gate, so I rebased before ticking "pushed to the latest dev commit" instead of ticking it against a stale base. Clean rebase, no conflicts, implementation unchanged; new head `7890be313`.

Re-verified on the rebased head:

- `bun x tsc --noEmit` → exit 0
- `bun test tests/codex-app-server-processes.test.ts` → **50 pass / 1 fail** (51 tests)

The one failure is `warnIfStaleCodexAppServersAfterStartupWrite (#1046) > a defaulted read is memoized, and invalidation is what clears it`. It is pre-existing on `dev` on this machine — measured on a clean tree, which gives **46 pass / 1 fail** across 47 tests, the same test. My diff adds 4 tests and no failures.

I re-baselined that rather than citing the measurement I posted on #2485, because this diff changes the caching in the very module that test exercises, and quoting an older number there would have been the wrong kind of shortcut.

CodeRabbit reported "No actionable comments were generated"; no Codex findings on this PR.

lidge-jun:
## 리뷰 · 우선순위 52 / 80

설명: 이 풀은 윈도우에서 카탈로그 상태 검사가 너무 길어서 요청이 기다리던 구멍을, 지난 관측을 먼저 돌려주고 뒤에서 다시 재도록 고친다. 지금 CURRENT dev HEAD 는 5170fc8eb 이다. 이번 시간에 origin/dev 가 두 번 착지했다. 2540 이 붙여넣기 파서에 URL 조각을 읽게 했고, 2543 이 쿼리와 조각에서 code 와 state 를 섞지 않게 고쳤다. 이 풀은 그 로그인 작업과 무관하다. 작성자는 ntdatt812 이다. 드래프트다. 베이스는 origin/dev 다. 파일 충돌은 없다. 리뷰 보호와 드래프트 때문에 BLOCKED 다. Refs 2499 를 적는다. Closes 는 없다. 라벨은 바꾸지 말 것. 내가 머지하지 않는다.

HEAD 의 요청 경로 수집기는 src/codex/app-server-processes.ts 839-937줄이다. 윈도우만 비동기 프로브를 기다린다. 다른 플랫폼은 839줄에서 동기 수집기로 바로 간다. 캐시 TTL 은 실측 5초이고 unknown 은 250ms 다. 그 프로브는 후보 프로세스마다 Invoke-CimMethod GetOwner 를 부른다. 이슈 2499 는 한 턴이 약 7.5초를 그 검사에 쓴다고 적는다. 작성자가 재현한 값은 후보 하나당 약 435ms 이다. 후보가 많으면 그 합이 TTL 보다 커져서, 캐시가 채워지기도 전에 만료되고 다음 턴이 또 기다린다. src/server/responses/collaboration.ts 228-236줄의 기본 수집기가 이 함수를 쓴다. 303-304줄은 stale 과 unknown 에서 안내를 숨기고, fresh 와 not_running 만 카탈로그를 설명한다. 그래서 이 대기가 v2 안내를 만드는 요청 경로에 붙는다.

이 풀은 CATALOG_STATE_MAX_STALE_MS 60초를 더한다. 만료됐지만 세대가 같고 unknown 이 아닌 관측은 바로 돌려주고, 같은 호출이 뒤에서 새로 잰다. 세대가 맞는 이유는 resetCodexAppServerCatalogStateCache 가 카탈로그 쓰기마다 세대를 올리고 캐시를 비우기 때문이다. 앱 서버가 그 사이에 꺼지거나 켜지는 일은 놓칠 수 있다. 본문은 새로 켜진 서버는 카탈로그보다 새것이라 fresh 안내가 거짓이 아니고, 꺼진 서버를 fresh 로 말하는 것도 안내를 속이지는 않는다고 적는다. 차가운 첫 호출은 그대로 기다린다. unknown 은 이 경로로 다시 주지 않는다. 시험 네 개가 tests/codex-app-server-processes.test.ts 에 붙는다. 시계와 안 끝나는 프로브로 첫 턴 대기, 나중 턴 즉시 응답, 한 번의 나열, 60초 천장, unknown 제외를 잠근다. 코드는 그 파일 두 개뿐이다.

문제는 아직 드래프트라는 점과, 뒤 갱신이 실패하면 좋은 관측을 unknown 으로 덮는다는 점이다. 본문 체크리스트 네 칸이 비어 있다. 작성자는 댓글로 칸을 채웠다고 했지만 게이트 상태가 autoDraftedByBot 이라 드래프트가 유지된다. HEAD 904-930줄의 then 은 실패를 unknown 으로 캐시에 쓴다. 이 풀이 지난 관측을 먼저 돌려준 뒤에도 그 쓰기는 그대로다. 일시 실패가 안내에 쓸 수 있는 fresh 를 250ms 짜리 unknown 으로 바꾼다. 지난 관측을 먼저 주는 계약이라면 실패한 갱신은 실측 캐시를 덮지 않는 편이 맞다. 의사결정 로그 831-838줄은 짧은 캐시와 차가운 턴 대기만 적는다. 만료 관측을 먼저 돌려주는 새 계약을 안 적는다. 둘째 시험은 프로브를 풀어 준 뒤 5ms 만 기다린다. 이번 시간 2540 안에 느린 러너 시험 고침이 있었으니, 그 5ms 는 약하다. 2499 를 이 풀로 닫지 말 것. 제안 1 GetOwner 팬아웃은 이 풀이 안 넣는다. types.ts 와 config.ts 가르기와는 무관하다. 프리뷰 배포가 아니다. 2463 2464 2465 를 닫지 말 것.

src/codex/app-server-processes.ts 839-937줄 HEAD - 윈도우 요청 경로가 TTL 안에서만 캐시를 쓰고, 만료되면 프로브를 기다린다. 이 풀이 60초 지난 관측을 먼저 돌려준다
PR 본문 체크리스트 네 칸 - 본문이 비어 있고 isDraft 가 true 다. 댓글로 칸을 채웠다고 적어도 게이트는 드래프트를 유지한다
HEAD 904-930줄 then 캐시 쓰기 - 갱신 실패가 unknown 을 캐시에 넣어 실측 fresh 를 덮는다. 이 풀이 그 쓰기를 안 고친다
src/codex/app-server-processes.ts 831-838줄 HEAD 의사결정 로그 - 짧은 캐시와 차가운 턴 대기만 적혀 있다. 만료 관측을 먼저 주는 계약을 안 적는다
tests/codex-app-server-processes.test.ts 새로 붙는 둘째 시험 - 프로브 해제 뒤 5ms 만 기다려 갱신 반영을 본다. 느린 러너에서 깨질 수 있다
2499 - Refs 만 있다. 제안 2 만 들어 있고 제안 1 은 없다. 이 풀로 이슈를 닫지 말 것

메인테이너의 판단이 필요한 지점
- 이 풀을 지금 합칠지. 내가 머지하지 않는다. 드래프트다. 본문 체크리스트를 채우고 레디로 올린 뒤에 본다
- 갱신 실패가 캐시를 unknown 으로 덮게 둘지. 실측 관측을 덮지 않는 편이 맞다
- 2499 를 지금 닫을지. 닫지 말 것. 제안 2 만 들어 있다
- 의사결정 로그를 고칠지. 고치는 편이 맞다. 짧은 캐시만 적혀 있다

너의 추천
기다린다. 드래프트로 둔다. 본문 체크리스트를 채우고 레디로 올린다. 갱신 실패가 실측 캐시를 unknown 으로 덮지 않게 고친다. 2499 는 연다. 내가 머지하지 않는다. 라벨은 그대로 둔다. 프리뷰 배포가 아니다.

이 댓글은 grok-bot이 작성했습니다

ntdatt812:
Four findings, all four acted on. Head is now `7ffd987d5`.

**1. The `then` cache write evicts a real reading on a failed refresh.** Correct, and it is the sharpest of the four — it is a hole this PR *created*. Before serve-stale, caching `unknown` cost at most the 250ms that state is allowed to live. Once an expired observation is what callers are handed, overwriting one with `unknown` takes the answer away from them: `unknown` is not servable, so the next caller waits for a probe instead of getting the reading it would otherwise have had. A transient enumeration failure was making the latency worse than before the fix, on exactly the path the fix exists for.

Now guarded — a refresh that lands `unknown` does not replace a same-generation, same-identity observation. Its own age still retires it.

Pinned by a new test, `a failed refresh does not evict the reading it was refreshing`. Mutation-checked: dropping `!wouldEvictAnObservation` from the condition makes exactly that test fail, and only it.

**2. Decision Log does not record the new contract.** Fixed — a second `[Decision Log · #2499]` block, in the same six-field shape as the one above it: the ~435ms-per-candidate probe cost against its own 5s TTL, why the two rejected alternatives were rejected (suggestion 1 changes what "could not verify the owner" means for the current-user scoping contract; suggestion 3 still pays the probe on every human-paced turn), and the honest downside — a server that stopped between readings can be described as `fresh` for up to the stale bound.

**3. The second test waits 5ms after releasing the probe.** Fair, and the reasoning is right: a sleep that short is a bet on how many turns a continuation needs, and losing that bet on a slow runner reads as a product bug rather than a slow machine. Replaced with a bounded `drainUntil` on the condition — it waits for the refresh to have stored its result, or fails loudly after 1,000 yields.

**4. Draft with an empty checklist.** The gate had reset the four boxes when it re-appended them, so my earlier tick was overwritten. Re-ticked and marked ready — and rebased first, onto `c0c9544da`, so "pushed to the latest dev commit" is true rather than ticked against a stale base.

On `Refs 2499` rather than `Closes`: agreed, and deliberate. Only suggestion 2 is here. Suggestion 1 — dropping the per-process `GetOwner` fan-out — is the bigger win and I prototyped it (`Win32_SessionProcess` joined to `Win32_LoggedOnUser` maps every PID in ~460ms total against ~435ms *per candidate* today), but `GetOwner` returning `ReturnValue=2` is currently what produces `__OCX_ENUM_INCOMPLETE__`, and the association path is missing rows rather than denied. Those two disagree about what "could not verify the owner" means, and getting it wrong loosens the current-user scoping the restart path depends on. It needs its own PR with its own ground-truth comparison.

**Local results on `7ffd987d5`**

```
bun test tests/codex-app-server-processes.test.ts
51 pass / 1 fail   (52 tests)
```

The one failure is `warnIfStaleCodexAppServersAfterStartupWrite (#1046) > a defaulted read is memoized, and invalidation is what clears it` — pre-existing on `dev` on this machine, measured on a clean tree at 46 pass / 1 fail across 47 tests, the same test. This diff adds 5 tests and no failures.

`bun x tsc --noEmit`: exit 0.

ntdatt812:
Taken — the doc and the code disagreed, and the doc was the one I meant. Head is now `e30a0cfd3`.

`CATALOG_STATE_MAX_STALE_MS` is documented as "how long a real observation may still be served **after it expires**", but the check measured age from `atMs`, so a `fresh` entry got 55s of stale serving rather than 60s. Fixed as suggested.

The stronger reason to anchor it to expiry is not the missing 5s, though — it is that a cap on *total age* couples the stale window to the TTL in a way that fails quietly. Raise `CATALOG_STATE_TTL_MS` past 60s and every expired entry is already too old to serve, so this path turns itself off with nothing to notice: no test breaks, no log, just the cold-wait behaviour coming back. Measured from expiry the two constants stay independent.

One note on the suggested expression, which I took verbatim: `catalogStateTtlMs(cached.status.state)` is state-dependent in form but constant in fact here, because the line above already excludes `unknown` and that is the only state with its own TTL. I kept the call rather than inlining `CATALOG_STATE_TTL_MS` — it reads as "its own freshness window plus the stale bound", and it stays correct if another state ever gets a TTL of its own.

**The test needed more than a moved number.** The existing case only pinned the bound from above, so both readings of it passed. It now pins both sides:

```
clock.ms += catalogStateTtlMs("fresh") + CATALOG_STATE_MAX_STALE_MS - 1;   // served, no wait
clock.ms += 1;                                                             // waits for a real reading
```

Mutation-checked: putting the old `now - cached.atMs < CATALOG_STATE_MAX_STALE_MS` back fails exactly that test and only it. Under the previous version of the test the same mutation passed, which is why the mismatch survived to review.

```
bun test tests/codex-app-server-processes.test.ts
51 pass / 1 fail   (52 tests)
```

The one failure is the pre-existing `warnIfStaleCodexAppServersAfterStartupWrite (#1046) > a defaulted read is memoized, and invalidation is what clears it`, baselined on a clean tree at 46 pass / 1 fail across 47 tests, same test. `bun x tsc --noEmit`: exit 0.

Checklist re-ticked and back out of draft — the gate had cleared the four boxes again when it re-appended them. Base is unchanged at `c0c9544da`, so this head is still on the latest dev commit.

Ingwannu:
The serve-stale design and the latest expiry-relative bound fix look materially better, but I am not approving this draft yet. It is still based behind current `dev@e1fb6755`, and the focused module suite is not green on either the PR or the reported baseline. Because this PR changes that same cache module, the existing failing memoization/invalidation test needs a deterministic disposition on the rebased exact head—not just a claim that it is pre-existing.

Please rebase onto current dev, rerun the full focused file with the repository-pinned environment, and either fix the failure or provide a minimal current-dev reproduction showing the exact same failure plus why it is unrelated. Keep #2499 open: this PR removes request-path waiting after a prior observation but does not remove the per-process GetOwner fan-out.

lidge-jun:
Landed on `dev` as #2580 with your three commits rebased onto the current head and authorship preserved. No code changes were needed — every review finding was already addressed on your branch.

The blocker was the stale base: 51 commits behind, with one failing test in the same module. After the rebase that failure is gone, so it was a stale-base artefact rather than a defect in the change. `bun test tests/codex-app-server-processes.test.ts` is 51 pass / 0 fail at the rebased head (the one skip is the Windows-only PowerShell enumeration case), typecheck clean. Thanks for the fix.

ntdatt812:
Thanks again for landing this — the fix itself is exactly where I wanted it.

One small thing about attribution, and I want to be clear up front that I don't think anything was done deliberately: the mechanics of a squash merge quietly dropped the authorship that your comment above intended to preserve.

What the repository records now:

- `#2580` is authored by `lidge-jun`, so it does not register as a merged PR from me.
- `#2542` is closed, so it doesn't either.
- The three commits on the `#2580` branch **do** carry `author.login = ntdatt812` — but `gh api repos/lidge-jun/opencodex/compare/dev...9a9f29cb --jq .status` returns `diverged`, so they are not reachable from `dev`.
- The single commit that actually landed on `dev` is `342911fe` (`parents=1`, i.e. a squash), and its `author.login` is `lidge-jun`.

So the work is on `dev`, but nothing on `dev` points back to me.

This isn't a property of squash merging in general — it's a property of *which PR* gets squashed. My own `#2485` was squashed the same way and the resulting commit `0f30b3959` has `author.login = ntdatt812`. The difference is only that `#2580` was opened from your account rather than mine.

So the ask is small and forward-looking: **when a contributor's branch needs a rebase, could it be rebased onto their own PR and that PR merged, instead of re-opening the work under a maintainer account?** That keeps exactly the outcome you described — same commits, same base, same review findings — and the authorship survives the squash, as `#2485` shows.

For this one specifically, no history rewrite is needed on my account; a line in the `#2580` description noting that the change originated in `#2542` would be enough of a public record.

Happy to keep sending fixes either way — just flagging it since it's invisible unless you go looking for it.

PR #2580, opened 2026-08-25
fix(codex): serve the last catalog reading while the Windows probe refreshes (rebase of #2542)

lidge-jun:
## Summary

Rebase of #2542 (by @ntdatt812) onto current `dev`. No code changes were needed — all
three commits rebased cleanly and every code-review finding was already addressed on the
author's branch:

- the stale bound is measured from expiry rather than from the reading,
- a failed refresh preserves the reading it was refreshing,
- the test uses deterministic polling instead of a fixed sleep,
- `unknown` is never served stale.

The PR was blocked on being 51 commits behind `dev` with one failing test in the same
module. After the rebase that failure is gone — it was a stale-base artefact, not a defect
in the change.

This serves the last known catalog reading while the Windows probe refreshes, so a turn no
longer pays the probe latency inline.

Closes #2542.

## Verification

```
$ bun test tests/codex-app-server-processes.test.ts
 51 pass, 1 skip, 0 fail, 204 expect() calls
 (the skip is the Windows-only PowerShell enumeration case)

$ bun x tsc --noEmit
(clean)
```

## Checklist

- [x] Targets `dev`
- [x] Rebased onto the current `dev` head (3/3, no conflicts)
- [x] The previously failing focused suite is green at the rebased head
- [x] Typecheck clean
- [x] Original authorship preserved in the commit history
- [x] No secrets or account identifiers in the diff

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows catalog-state loading by serving recently expired results while updates run in the background.
  * Prevented temporary refresh failures from replacing a previously valid catalog state.
  * Ensured excessively old or unknown results are refreshed before being shown.
  * Reduced duplicate refresh requests when multiple requests arrive at once.

### DIFF
--- a/src/codex/app-server-processes.ts
+++ b/src/codex/app-server-processes.ts
@@ -750,6 +750,11 @@ const CATALOG_STATE_UNKNOWN_TTL_MS = 250;
  * previous reading immediately keeps that cost off the turn without pretending it is
  * fresh -- the refresh it triggers is what makes the next reading current.
  *
+ * Measured from expiry rather than from when the reading was taken, so a `fresh`
+ * entry stays servable for its own TTL plus this bound. Anchoring it to expiry keeps
+ * the stale window independent of the TTL -- a cap on total age would quietly turn
+ * this path off if the TTL were ever raised past it.
+ *
  * Bounded rather than unlimited: if the refresh keeps failing, an observation this
  * old stops being evidence about the machine and it is better to wait for a real one.
  * `unknown` is never served this way -- it is a failure to observe, not an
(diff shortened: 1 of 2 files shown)


ntdatt812:
Thanks for doing the rebase — verified it landed intact on `dev`, including the two changes that came out of review rather than the original diff:

```
$ git show origin/dev:src/codex/app-server-processes.ts | grep 'now - cached.atMs'
now - cached.atMs < catalogStateTtlMs(cached.status.state)
now - cached.atMs < catalogStateTtlMs(cached.status.state) + CATALOG_STATE_MAX_STALE_MS
```

The second line is the one worth confirming: the stale bound measured from expiry rather than from when the reading was taken. That was the CodeRabbit finding on #2542, and it is the one a rebase could most easily have dropped if it had resolved a conflict toward the older shape. It did not. The `wouldEvictAnObservation` guard and the `drainUntil` helper are both there as well.

You were right about the failing test being a stale-base artefact — I had baselined it at 46 pass / 1 fail on a clean `dev` and said so on the PR, but 51 commits behind was enough to make that hard to see from the outside. Rebasing was the cleaner call than arguing the baseline.

PR #2617, opened 2026-08-25
fix(responses): strip search_content_types on web_search for Muse Spark

DevonGithub:
## Problem

`opencode-go/muse-spark-1.2-contributor` fails when routed through the Codex app / opencodex proxy. It is the only model on the `opencode-go` key that breaks; `gpt-5.6-luna` and `deepseek-v4-flash` work on the same key.

## Root cause

Muse Spark **only serves over the Responses API** (`POST /v1/responses`). A direct test of `POST /v1/chat/completions` for this model returns `500 Internal server error` even with no tools. The opencodex registry had never routed Muse to Responses; only `gpt-5.6-luna` was in `modelWireDefaults`.

Additionally, once a client (Codex) sends a `web_search` tool carrying `search_content_types`, Muse's gateway rejects the whole request with:

`400 invalid_request_error tools[].search_content_types is only supported for web_search_preview tools`

Gateway probe results (direct, 2026-08-26) on Muse over `/v1/responses`:

| Tool | Field | Result |
|------|-------|--------|
| `web_search` | (plain) | 200 |
| `web_search` | `search_content_types` | **400** |
| `web_search` | `search_context_size` | 200 |
| `web_search` | `external_web_access` | 200 |
| `web_search` | `filters: {}` | 200 |
| `web_search` | `user_location` | 200 |
| `web_search_preview` | every field incl `search_content_types` | 200 |

So the **only** rejected combination is `search_content_types` on a plain `web_search` tool, and only for Muse. Luna accepts every shape.

## Fix

1. **`src/adapters/openai-responses.ts`** - added `stripMuseSparkUnsupportedWebSearchFields`, which removes `search_content_types` from `web_search` tools for the exact model `muse-spark-1.2-contributor` only. It runs on the routed Responses path alongside the existing OpenAI-only web-search cleanup, and does not touch `web_search_preview` or any other field.
2. **`src/providers/registry.ts`** - added `muse-spark-1.2-contributor` to `opencode-go` `modelWireDefaults` as `openai-responses`, mirroring the existing `gpt-5.6-luna` entry.

## Tests

Added `stripMuseSparkUnsupportedWebSearchFields` coverage in `tests/responses-routed-web-search-fields.test.ts` (all pass):

- strips `search_content_types` from `web_search` for the exact Muse model
- keeps `web_search_preview` untouched
- strips nested `additional_tools`
- leaves a non-Muse model unchanged
- ignores a near-match model id

Also verified manually: Muse now returns 200 over `/v1/responses` with the offending tool; `gpt-5.6-luna` and `deepseek-v4-flash` still return 200.

## Note

This does not change how other models route. Explicit user `modelAdapters` entries still take precedence (same contract as the existing Luna routing).

## Review readiness checklist

### DIFF
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -1576,6 +1576,57 @@ export function stripOpenAiOnlyWebSearchFields(body: unknown): unknown {
   return changed ? next : body;
 }
 
+/**
+ * OpenCode Zen / Go Muse Spark Responses gateway refuses `search_content_types`
+ * on a plain `web_search` tool (400) but accepts it on `web_search_preview`; a
+ * plain `web_search` is also accepted. Probed directly against the gateway on
+ * 2026-08-26: `web_search` + `search_content_types` -> 400, `web_search_preview`
+ * + `search_content_types` -> 200, plain `web_search` -> 200. Luna accepts every
+ * shape, so this is Muse-only. Drop only the field the gateway refuses while
+ * keeping the tool type and every other accepted option intact.
+ */
+export function stripMuseSparkUnsupportedWebSearchFields(body: unknown, modelId: unknown): unknown {
+  if (!isPlainObject(body)) return body;
+  if (typeof modelId !== "string" || modelId.trim().toLowerCase() !== "muse-spark-1.2-contributor") return body;
+
+  const rewriteTools = (tools: unknown[]): { tools: unknown[]; changed: boolean } => {
+    let changed = false;
+    const rewritten = tools.map(tool => {
+      if (!isPlainObject(tool) || tool.type !== "web_search") return tool;
+      if (!Object.hasOwn(tool, "search_content_types")) return tool;
+      const { search_content_types: _dropped, ...rest } = tool;
+      changed = true;
+      return rest;
+    });
+    return { tools: changed ? rewritten : tools, changed };
+  };
+
+  let next: Record<string, unknown> = body;
+  let changed = false;
+  if (Array.isArray(body.tools)) {
+    const rewritten = rewriteTools(body.tools);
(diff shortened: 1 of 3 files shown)


lidge-jun:
Landed on `dev` as #2621, with a regression test and two adjustments.

The mechanism is right, and the reasoning holds up independently: the field comes from Codex's own `hosted_spec.rs` (`web_search_tool_type: TextAndImage`), and Codex already handles the same incompatibility class for Bedrock by selecting text-only search. Dropping exactly the refused field at the adapter boundary is a compatibility guard, not a symptom patch.

Three things I changed:

- **Branch.** This targeted `main`, which only moves by maintainer promotion — your own title flagged it. Rebased onto `dev`.
- **Version bump removed.** The diff also took `package.json` from `2.32.1-preview.20260825` to `2.33.0`. Versioning belongs to the release train; a web-search fix should not move the release surface as a side effect.
- **Test added.** This touches a shared adapter and the provider registry with no test, so neither the model-scoped removal nor the new Responses wire default was protected. The new test pins that `web_search_preview` keeps the field, that other models on the same provider are untouched, that nested `additional_tools` is sanitized too, and that the registry map stays an exact-model allowlist.

I also refreshed the registry decision log, which still described `modelWireDefaults` as Luna-only after gaining a second entry.

One thing I could not confirm: your live 200/400 matrix. The account available here now gets `403 … requires explicit opt in` for every Muse request, so the rejection could not be re-probed. That does not block the change — it is scoped to one exact model and removes one field — but it is worth stating rather than implying I verified it.

Thanks for finding this. Closing in favour of the merged PR.

lidge-jun:
Landed via #2621 at 1a15d6292

PR #2618, opened 2026-08-25
feat(acceptance): disposable-host service acceptance and the uninstall bug it found (#1048)

### LINKED ISSUE #1048: WP13: composed acceptance at the production boundary
## Why this is its own issue

Every phase of the Codex write substrate proved its own mechanism, but phase-local green tests do not prove that every production entry point converges on the same coordination substrate.

This issue tracks composed acceptance across those production boundaries.

## Current status

This issue is **partially implemented**.

PR #1106 merged the workstation-safe composed acceptance suite and six production-path scenarios into `dev`.

Delivered there:

- `tests/codex-composed-acceptance.test.ts`;
- real spawned CLI children and a real HTTP server on temporary homes;
- entry-to-funnel manifest coverage;
- lost-transition behavior through HTTP;
- foreign-home zero-artifact refusal;
- same-effective-user cross-home single-lock behavior;
- Grok OFF surviving a real restart;
- restore truth through `ocx restore --json`.

That work also found and fixed several real production holes, including OFF paths creating artifacts, stale server-captured config during `/api/sync`, foreign homes accepted over HTTP, and opaque busy results.

The workstation-safe portion should not be reimplemented.

## Remaining scope

The issue remains open for the acceptance work deliberately deferred from #1106:

1. **Disposable-host service-class scenarios** that must not run in ordinary workstation test discovery.
2. **Remaining production census rows** that are not yet proven to converge on the coordinator.
3. Final proof that no production writer can bypass the substrate under the supported service-manager paths.

The deferred service class includes the scenarios identified in the WP13 resume plan as:

```text
P09
P10
P18
P34-P36
```

The intended disposable-host entry point remains conceptually:

```text
scripts/disposable-host/codex-service-composed-acceptance.ts
```

Service-manager tests require a host that can be destroyed or reset safely and should stay out of normal test discovery.

## Related follow-ups

Focused correctness issues remain separate:

- #1049 — adopt pre-substrate Codex homes into the coordinator
- #1183 — make history-migration no-op detection atomic
- #1230 — prevent concurrent startup from restoring journal state owned by another healthy proxy

Those issues should not be folded into this acceptance tracker unless the missing acceptance scenario itself is the problem.

## Acceptance criteria

- [x] Workstation-safe composed acceptance suite exists.
- [x] Six production-path scenarios from #1106 are covered.
- [x] Real CLI-child and HTTP-server paths are exercised.
- [ ] Disposable-host service-class acceptance is implemented.
- [ ] Remaining census rows are covered or explicitly retired with evidence.
- [ ] Service-manager entry points are proven to converge on the write coordinator.
- [ ] The complete composed acceptance program passes on current `dev`.

## Close condition

Close this issue only when the deferred disposable-host/service-class coverage and remaining production census are complete. The workstation-safe subset delivered by #1106 is necessary but not sufficient for closure.

lidge-jun:
## Summary

Closes #1048. The workstation-safe half shipped in #1106; what stayed open was the disposable-host service class (P09, P10, P18, P34, P35, P36), the remaining census, and the proof that service-manager entry points converge on the write coordinator. `git ls-tree origin/dev scripts/disposable-host` was empty.

`scripts/disposable-host/codex-service-composed-acceptance.ts` is a script, not a test, and is deliberately outside `bun run test` discovery. It refuses to run without a root-owned image sentinel, then runs the empty-registration gate on **both sides of every row** — an unavailable query, a permission error, a nonempty registration or a leftover artifact is a hard failure, never a skip. Each row installs only its own fixture registration, invokes the real production entry as a real process, and asserts exactly one admitted transaction in the right direction.

## It ran for real, and it found four bugs

Run on a Linux/systemd host, exit 0, all six rows:

```
P09 PASS generation=1->2 direction=remove tx=7f9b2fc7-…
P10 PASS generation=1->2 direction=remove tx=13de0a20-…
P18 PASS generation=1->2 direction=remove tx=5736b469-…
P34 PASS generation=2->3 direction=apply  tx=d5aee30d-…
P35 PASS generation=1->2 direction=remove tx=bbad3bbd-…
P36 PASS generation=1->2 direction=remove tx=e08c2ca5-…
PASS disposable service census: P09, P10, P18, P34, P35, P36
```

Getting there required fixing four things, and **one of them is a real product bug that no workstation test could have found**:

**`ocx uninstall` could not remove a config home OpenCodex created itself.** It reported "partial uninstall: unowned files remain" and left the directory behind, because two of our own writers produce files the ownership manifest never claimed: `admin-api-token` and the per-`CODEX_HOME` `catalog-backup-<16 hex>.json`. The hashed backup cannot be a literal manifest entry — its name depends on which Codex home it mirrors — so it is swept by its exact generated shape, narrow enough that a user's file cannot collide. A lookalike test proves `catalog-backup-notahash.json` still survives. Falsified: reverting either hunk alone reddens the new removal test.

The other three were harness defects, each interesting:

- **A fake `HOME` cannot work for these rows.** `systemctl --user` resolves its unit directory from the running user manager, not `$HOME`, so the install wrote a unit the manager never reads. That is precisely why these rows are disposable-host-only: a globally addressed service cannot be redirected into a temp root, so safety comes from the sentinel plus the gate, not from isolation.
- **The fixture pre-seeded `config.json`**, so OpenCodex never claimed the home and `ocx uninstall` correctly refused to delete a directory it could not prove it owned. P10 was failing on the fixture's shortcut, not the production command. The seed is now merged in after the product creates and claims its home.
- **P18 authenticated with the wrong token**, and a drained connection was treated as failure. The service mints its own `admin-api-token`; the env var only reaches processes this script launches. And `POST /api/stop` drains the process serving the request, so the socket can close before a response is flushed — the more completely the endpoint does its job, the likelier that is. A 4xx still fails; the authoritative oracle is the +1 remove transaction either way.

## One finding recorded rather than hidden

The rows report `PROVENANCE record absent`. That is not a defect in the rows: `updateIntegrationRecord` has **no caller in `src/`**, so the provenance ledger is never written, and `convergence-types.ts` states outright that provenance is optional at record v1. The original assertion required an entry no production path can produce — it was testing an unimplemented writer. It now fails only on **disagreement**: a ledger that exists, has entries, and lacks the transaction just admitted is a real defect. Absence is reported and allowed. Worth a follow-up issue; not something to make green by assertion.

## Census

`devlog/_plan/260826_wp13_disposable_host/010_census.md` covers all 36 rows: 14 with direct composed process coverage, 22 explicitly retired with named evidence. "Deferred" appears nowhere.

## Host restoration

The host was returned to its exact prior state, proven by re-running the same gate after the run: `list-unit-files` empty, `status` reports unit-not-found, the unit file absent, no fixture temp roots, and the sentinel removed.

## Verification

```
bun x tsc --noEmit                                     exit 0
bun test <composed acceptance + ownership + uninstall + service>  163 pass / 0 fail
disposable runner on systemd host                      exit 0, six rows PASS
```

`bun run test` does not discover the script — it is not a `tests/*.test.ts` file.

## Checklist

- [x] Targets `dev`
- [x] Product fix carries its own regression, falsified per hunk
- [x] Census recorded with evidence per row
- [x] Test host restored and proven restored

## Summary by CodeRabbit

* **New Features**
  * Added disposable-host acceptance coverage for service registration and cleanup workflows.
  * Added tracking and cleanup for admin API tokens and generated catalog backup files.

* **Bug Fixes**
  * Uninstallation now removes eligible generated backup files while preserving files with invalid names.
  * Improved reporting when cleanup is only partially successful.

* **Tests**
  * Added coverage for token and catalog-backup cleanup scenarios.
  * Documented production-entry coverage and retirement status for tracked acceptance cases.

### DIFF
--- a/devlog/_plan/260826_wp13_disposable_host/010_census.md
+++ b/devlog/_plan/260826_wp13_disposable_host/010_census.md
@@ -1,0 +1,56 @@
+# #1048 production-entry census
+
+Rechecked against `origin/dev` at `aacd6528d8` on 2026-08-26. “Covered”
+means the named test invokes the production entry or the disposable runner does.
+Rows retired from *additional composed execution* retain a concrete regression or
+inventory proof and a reason why another process-level case would duplicate the
+same receiving production edge.
+
+| ID | Disposition | Evidence |
+|---|---|---|
+| P01 | Retired from additional composed execution | `src/cli/init.ts` calls the same native apply convergence exercised by P02/P08 after saving config; `tests/init-eof.test.ts` owns the unique interactive/EOF boundary. No independent writer remains. |
+| P02 | Covered | `tests/codex-composed-acceptance.test.ts` — “A-reduced: real CLI and HTTP entry points preserve an OFF Codex config/home”. |
+| P03 | Retired from additional composed execution | `tests/shutdown-launcher.test.ts` — “ocx launcher graceful shutdown”; shutdown cleanup uses the same remove convergence proven by P07. |
+| P04 | Covered | `tests/codex-composed-acceptance.test.ts` — “A-reduced: real CLI and HTTP entry points preserve an OFF Codex config/home”. |
+| P05 | Covered | Same A-reduced case invokes real `ocx sync`. |
+| P06 | Covered | Same A-reduced case invokes real `ocx sync-cache`. |
+| P07 | Covered | The A-reduced case invokes real `ocx restore`; D-reduced invokes the same row under foreign ownership. |
+| P08 | Covered | Same A-reduced case invokes real `ocx restore back`. |
+| P09 | Covered (disposable only) | `scripts/disposable-host/codex-service-composed-acceptance.ts` row P09, real `ocx stop`, exact +1 remove transaction. |
+| P10 | Covered (disposable only) | Disposable runner row P10, real `ocx uninstall`, exact +1 remove transaction before owned-state deletion. |
+| P11 | Retired from additional composed execution | `tests/cli-help.test.ts` — “recover-history requires exact confirmation before mutating history”; this command owns a history-only job, not a native mutation, so it cannot satisfy Scenario A's native-transaction oracle. |
+| P12 | Retired from additional composed execution | `tests/cli-provider.test.ts` — “provider add --sync flag is accepted without error” and “provider add --sync --json reports needsSync false”; live sync receives P19/P05's catalog convergence. |
+| P13 | Retired from additional composed execution | `tests/cli-models.test.ts` — “models add accepts slash model ids”; its live branch calls the same management/catalog convergence inventoried below. |
+| P14 | Retired from additional composed execution | `tests/cli-models.test.ts` — “an unambiguous slash selector still removes its row”; same receiving convergence as P13. |
+| P15 | Retired from additional composed execution | `tests/codex-v2-gate.test.ts` — “off -> on carries the active legacy value and removes the boot conflict”; mode mutation is independently tested and its sync tail is P05/P19. |
+| P16 | Retired from additional composed execution | `tests/codex-v2-gate.test.ts` — “on -> off carries the active v2 value and removes v2 limit storage”; same sync tail as P15. |
+| P17 | Covered | P02 in A-reduced and “Grok E2E: route-disabled Grok stays absent across a real startup” execute startup reconciliation in real child processes. |
+| P18 | Covered (disposable only) | Disposable runner row P18, authenticated real `POST /api/stop`, exact +1 remove transaction. |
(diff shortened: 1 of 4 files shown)


lidge-jun:
Full local suite on the merge head: **15046 pass / 12 skip / 3 fail**.

All three failures are `tests/test-runner.test.ts > bun test machine lock`, and they are an artifact of how I ran it, not a defect: I passed `OCX_TEST_NO_QUEUE=1` to avoid colliding with a parallel run, and that env var is exactly what those tests assert the lock honours. Run without the override, the same file is **16 pass / 2 skip / 0 fail**.

So the effective result is a green suite. Recording it here rather than quietly reporting the smaller number.

PR #2639, opened 2026-08-25
fix(responses): backfill missing status and created_at for strict decoders

bet4it:
## Summary

Follow-up to #2142. That PR backfilled the required `id` field on output items. After it merged, grok-build still fails with two more strict-decoder errors on the Responses passthrough path:

- `serialization error: missing field status` — some upstream relays omit `status` on message output items. `OutputMessage.status` is a required `OutputStatus` (no `#[serde(default)]`) in the async-openai fork that grok-build pins (rev `95b52eb`).

- `serialization error: missing field created_at` — some upstream relays omit `created_at` on the `Response` object in `response.created` / `response.completed` events. `Response.created_at` is a required `u64` (no `#[serde(default)]`).

Both are the same class of problem as `annotations` (#1941) and `id` (#2142): a required field with no serde default, omitted by a relay that the `openai-responses` passthrough adapter forwards verbatim. The translation path (bridge.ts) is unaffected — `closeCurrentMessage` already emits `status: "completed"`, and `responseSnapshot` already emits `created_at`.

### What changed

**`backfillItemStatus`** (`responses-field-backfill.ts`): adds `status` to output items when `type === "message"` and the field is absent. Only message items carry this field in the Responses schema; reasoning, function_call, and other item types do not. The value is inferred from the event context:
- `output_item.added` → `in_progress` (the item is still being generated)
- `output_item.done` → `completed`
- Response-level events (`response.created`, `response.completed`, etc.) → derived from the response `status` field

Existing values are never overwritten.

**`created_at` backfill** (in `backfillResponseOutput`): adds `created_at` to the response object when absent. The timestamp is captured **once per rewrite factory** (SSE path) or **once per call** (JSON path), so every event in the same stream carries the same value, even if the stream spans a second boundary.

Both backfills are wired through `backfillOutputItem` / `backfillResponseOutput`, so they cover the SSE block rewrite path (`createResponsesFieldBackfillBlockRewrite`) and the bounded-JSON passthrough path (`backfillResponsesFieldsJson`) simultaneously.

## Verification

- `bun test tests/responses-field-backfill.test.ts` — 34 pass, 0 fail (28 existing + 6 new: backfill in_progress on output_item.added, backfill in_progress on response.created, backfill incomplete on response.incomplete, created_at consistency across events in the same stream, plus the existing status/created_at tests).
- `bun x tsc --noEmit` — no new errors (3 pre-existing errors in `claude-messages.ts` and `fetch-helpers.ts` are unchanged on `origin/dev`).

## Checklist

- [x] Docs or release notes were updated when needed. (No user-facing behavior change beyond fixing the crash; the backfill module is internal.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (Adds protocol-required fields with safe defaults only; no credential, auth, or user content is read or logged.)

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Responses now consistently include missing message status values based on event or response progress.
  * Responses include a Unix timestamp when `created_at` is missing.
  * Existing status and timestamp values remain unchanged.
  * Timestamps stay consistent across events within the same response stream.
  * Failed responses and non-message items are handled safely.

* **Tests**
  * Added coverage for streaming and JSON response backfilling, including incomplete events and non-message items.

### DIFF
--- a/src/server/responses/responses-field-backfill.ts
+++ b/src/server/responses/responses-field-backfill.ts
@@ -92,6 +92,29 @@ function nextSyntheticItemSlot(): ItemIdSlot {
 }
 
 /**
+ * Backfill `status` on a message output item if missing.
+ *
+ * The Responses API spec defines `status` as a required field on
+ * `OutputMessage`. Some upstream relays omit it, which causes strict
+ * deserializers (e.g. grok-build's serde types) to fail with
+ * `missing field 'status'`. Only message items carry this field in the
+ * Responses schema; reasoning, function_call, and other item types do not.
+ *
+ * The value is inferred from the event context: `output_item.added` and
+ * `response.created` / `response.in_progress` mean the message is still
+ * being generated (`in_progress`); `output_item.done` and
+ * `response.completed` / `response.incomplete` mean the message is
+ * finalized (`completed` / `incomplete` respectively).
+ *
+ * Returns the same object reference if no change is needed.
+ */
+function backfillItemStatus(item: Record<string, unknown>, inferredStatus: string): Record<string, unknown> {
+  if (item.type !== "message") return item;
+  if ("status" in item) return item;
+  return { ...item, status: inferredStatus };
+}
+
+/**
  * Backfill annotations: [] on an output_text content part if missing.
  * Returns the same object reference if no change is needed.
  */
@@ -122,10 +145,10 @@ function backfillContentArray(content: unknown): unknown {
 
 /**
  * Walk an output item and backfill output_text parts in its content.
- * Also backfills a missing required id on the item itself.
+ * Also backfills a missing required id and status on the item itself.
  * Returns the same object reference if nothing changed.
  */
-function backfillOutputItem(item: unknown, slot: ItemIdSlot): unknown {
+function backfillOutputItem(item: unknown, slot: ItemIdSlot, inferredStatus: string): unknown {
   if (!isPlainObject(item)) return item;
   // The compact wire family is the `/v1/responses/compact` format, not a Responses output item.
   // Those items have no `id` in that contract, so synthesizing one changes a response body the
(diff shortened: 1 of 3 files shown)


lidge-jun:
## 리뷰 · 우선순위 64 / 80

설명
이 PR은 Responses 통과 경로에서 상류 서버가 꼭 필요한 값을 빼먹었을 때, OpenCodex가 그 값을 채워 주려는 수정입니다. 지금 `dev`의 `src/server/responses/responses-field-backfill.ts`는 `annotations`와 출력 항목 `id`만 고칩니다. 출력 항목을 고치는 중심은 128~140줄의 `backfillOutputItem`이고, 응답 전체를 걷는 곳은 147~158줄의 `backfillResponseOutput`입니다. PR은 여기에 메시지의 `status`와 응답의 `created_at`을 더합니다. 엄격한 디코더는 이 두 값이 없으면 응답 내용을 읽기 전에 실패하므로, 문제를 고치는 방향과 파일 위치는 맞습니다.

이 보정기는 실제 통과 경로에도 이미 알맞게 연결되어 있습니다. 스트리밍 응답은 `src/server/responses/core.ts` 3905줄에서 `createResponsesFieldBackfillBlockRewrite()`를 항상 사용하고, JSON 응답은 같은 파일 4113~4114줄에서 `backfillResponsesFieldsJson()`을 사용합니다. 따라서 이 한 모듈을 고치면 SSE와 JSON을 함께 고칠 수 있습니다. 번역 경로를 따로 건드리지 않은 것도 범위를 작게 지키는 좋은 선택입니다.

`created_at`을 응답 최상위에 넣고, 이미 값이 있으면 그대로 두는 원칙도 좋습니다. `status`도 메시지 항목에만 넣고 함수 호출이나 추론 항목에는 넣지 않으므로 스키마 범위를 잘 지켰습니다. 새 테스트는 값이 없을 때 채우기, 기존 값 보존, 메시지가 아닌 항목 보존, SSE와 JSON 두 경로를 확인합니다. CI도 현재 통과했습니다.

하지만 `status`를 언제나 `completed`로 넣으면 스트림 초반의 뜻이 틀립니다. 현재 `dev`의 `rewriteEvent`는 170~186줄에서 `response.output_item.added`와 `response.output_item.done`을 같은 `backfillOutputItem`으로 보냅니다. PR의 새 함수는 두 이벤트를 구분하지 않으므로, 아직 내용 델타가 이어질 `output_item.added` 메시지도 이미 끝난 메시지로 바꿉니다. 198~205줄의 응답 스냅샷 처리도 `response.created`와 `response.completed`를 구분하지 않으므로 같은 문제가 생깁니다. 새 테스트는 `done`과 `completed`만 검사해서 이 잘못된 초반 상태를 잡지 못합니다.

또한 PR은 `created_at`이 빠진 SSE 이벤트를 만날 때마다 그 순간의 `Date.now()`를 새로 부릅니다. 한 응답이 1초를 넘기면 `response.created`와 `response.completed`에 서로 다른 생성 시각이 들어갈 수 있습니다. 이 보정기 팩토리는 요청마다 `src/server/responses/core.ts` 3905줄에서 한 번 만들어지므로, 팩토리를 만들 때 시각을 한 번 잡아 같은 스트림의 모든 이벤트에 재사용할 수 있습니다. JSON 경로는 함수 호출 한 번 안에서 시각을 한 번 잡으면 됩니다.

`src/server/responses/responses-field-backfill.ts/backfillItemStatus` - `output_item.added`와 `response.created`에도 무조건 `completed`를 넣어 진행 중인 메시지를 끝난 메시지로 잘못 표시합니다.
`src/server/responses/responses-field-backfill.ts/backfillResponseOutput` - 이벤트마다 현재 시각을 다시 계산해 같은 응답의 `created_at`이 스트림 도중 달라질 수 있습니다.
`tests/responses-field-backfill.test.ts` - `output_item.added`/`response.created`의 상태와 한 스트림 안 `created_at` 일관성을 검사하는 회귀 테스트가 없습니다.

메인테이너의 판단이 필요한 지점
- 빠진 메시지 상태를 이벤트 종류에 맞춰 `in_progress`/`completed`/`incomplete`로 추론할지, 관찰된 오류가 난 종료 이벤트에서만 보정할지 정해야 합니다.
- 상류가 생성 시각을 주지 않았을 때 요청 단위의 보정 시각을 쓰는 것을 호환 정책으로 받아들일지 정해야 합니다.

너의 추천
지금 바로 합치지 말고 변경을 요청하는 것을 추천합니다. `status`를 이벤트 단계에 맞게 넣고, `created_at`은 보정기 생성 시 한 번 계산해 같은 응답에서 고정하세요. `output_item.added`, `response.created`, 1초가 지나도 같은 `created_at`을 쓰는 테스트를 더한 뒤 합치면 됩니다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
The backfill belongs in the right module, but status inference is currently too coarse. `response.output_item.added` and snapshots in `response.created` can receive `completed` even though deltas are still expected. Infer status from event/response phase: added/created should remain `in_progress`, done/completed should be `completed`, and incomplete terminals should remain `incomplete`.

Also capture the fallback `created_at` once per stream-rewriter instance (and once per JSON rewrite), rather than calling Date.now for each event, so one response cannot acquire different creation timestamps over a long stream. Add regressions for output_item.added, response.created, incomplete status, existing-value preservation, and stream-wide timestamp identity.

abhisheksharma2411 (commented):
Careful work, and the same class of fix as #1941/#2142 done the same way — preserve when present, synthesize only when absent, return the same object reference when nothing changed so the no-op path stays allocation-free. The `if ("status" in item) return item` check is the right shape: a present-but-odd value is upstream's to own, not ours to correct.

Two things I checked because they're the usual traps here, and both are clean — worth saying so you don't get asked:

- **`created_at` is seconds, not milliseconds.** `Math.floor(Date.now() / 1000)` on both paths. A `u64` a thousand times too large would decode fine and be wrong forever, so this is the detail I'd have gone looking for first.
- **The factory is per-request, not per-process.** `createResponsesFieldBackfillBlockRewrite()` is called inside `handleResponsesInner` (`core.ts:3937`), so "captured once per rewrite factory" means once per stream. If it were hoisted to module scope every response for the process lifetime would carry the boot timestamp. It isn't. Might be worth a word in the docstring pinning that requirement, since the correctness of the comment depends on a call site in another file.

**Where I'd push back is `response.failed` → `completed`.**

`inferredStatusForEventType` falls through to `"completed"` for any unmatched event type, and `messageStatusFromResponseStatus("failed")` returns `null`, so the fallback wins. The test pins it deliberately:

```js
// "failed" is not a valid OutputMessage status; the backfill should have
// fallen back to the event-type inference (completed for response.failed).
expect(parsed.response.output[0].status).toBe("completed");
```

I agree entirely with the first half — `failed` must not be written to `OutputMessage.status`. It's the choice of replacement I'd question. On `response.failed` the message demonstrably did *not* complete; the fixture's own text is `"partial"`. `incomplete` is the exact word the schema provides for "generation stopped before finishing", and it's already in the accepted set.

The distinction matters more here than it would in most code because this is a **passthrough** path. Backfilling `annotations: []` or a synthetic `id` is safe precisely because the value is uncontroversial — it carries no claim. `status: "completed"` does carry a claim, and it's a claim the upstream never made and that happens to be false. A client branching on `status === "completed"` to decide whether to render a message as final, persist it, or stop waiting will treat a truncated message as whole. That's a worse failure than the decode error being fixed, because it's silent.

Concretely, I'd map at the response-status layer rather than the event-type layer:

```ts
function messageStatusFromResponseStatus(status: string): string | null {
  if (status === "in_progress" || status === "completed" || status === "incomplete") return status;
  // A response that failed or was cancelled did not finish generating its
  // message. `incomplete` is the message-level equivalent; `completed` would
  // assert something the upstream never claimed.
  if (status === "failed" || status === "cancelled") return "incomplete";
  return null;
}
```

and flip `inferredStatusForEventType` to return `"incomplete"` for `response.failed` so the two agree when the response status is absent too. That keeps every existing test green except the one asserting `completed`, which would become the assertion that this is `incomplete`.

If `completed` was chosen because a client somewhere chokes on `incomplete`, that's a fair reason and worth putting in the comment — right now the comment explains what the code does but not why `completed` beat `incomplete`, and that's the decision a future reader will want.

Everything else looks right to me, including not touching non-message items and the per-stream `created_at` consistency test.

Ingwannu (changes requested):
Reviewed exact head `aa385f9746fed35a9ec1bd249dff65891c2d23cd`. The focused field-backfill suite is now correct (37/37 passed locally in the isolated PR worktree), and the previous lifecycle/status blockers appear resolved. The current exact-head full CI still has one deterministic integration failure, however: run `33032444280`, job `test 4/4` (`98392954733`).

`tests/server-combo-failover-e2e.test.ts:1323` still requires the backup Responses JSON to equal the upstream fixture byte-for-byte. This PR intentionally adds `created_at` at the response level and `status: "completed"` to the message item, so that existing assertion now fails with exactly those new fields. Update this integration regression to assert the intended backfilled shape while preserving the actual purpose of the test: the backup response content/model/usage must remain exact apart from the documented compatibility backfill. Because `created_at` is generated, assert its valid integer shape or control time deterministically rather than hard-coding the wall clock.

Please rerun exact-head CI after updating the integration expectation. I am keeping changes requested until every required shard is green; this is a test-integration blocker, not a new objection to the backfill design.

lidge-jun:
Partially landed on `dev` as `64c6d642b` (PR #2721), in the cherry-pick lane of the 260827 bug-PR merge round.

**Taken as-is:** the `status` backfill. Your diagnosis is right — `OutputMessage.status` is required and a relay that omits it breaks strict decoders — and so is the call to map response-level `failed`/`cancelled` to `incomplete` rather than `completed`. Writing `completed` there would let a client treat a truncated message as whole. That reasoning is what made the rest of the change easy to trust.

**Not taken:** the `created_at` backfill. It breaks `tests/server-combo-failover-e2e.test.ts:1323`, which asserts a combo backup response is relayed byte-exact:

```
dev:          74 pass / 0 fail
this branch:  73 pass / 1 fail
  + "created_at": 1787801315,
    "id": "resp-m2",
```

Two real contracts collide — "relay this verbatim" and "add the field the upstream omitted" — and which one yields is a decision worth making deliberately rather than folding in beside `status`. Worth knowing: `status` is only safe by accident. That fixture happens to already carry `status`, so it never exercises the injection; delete that one field and the same assertion fails identically. Recorded in `devlog/_plan/260827_bug_pr_merge_round/021_status_vs_created_at_asymmetry.md`, and the follow-up is probably to exempt the passthrough path from field backfill entirely.

**One defect found while reviewing:** `response.queued` fell through to the `completed` default, marking an unstarted message as finished — the same overclaim your `failed → incomplete` mapping exists to prevent. Fixed in `6877f646f` with two regression cases.

Closing since the correct half is on `dev`. If you want to pursue `created_at`, a PR that scopes it to the translated path (or updates the combo contract on purpose) would be very welcome.

Issue #993, opened 2026-08-04
[Bug]: Kiro provider profileArn required 400 for Builder ID accounts on gated models


redeye1011:
### Summary

Kiro's `GenerateAssistantResponse` backend rejects requests from AWS Builder
ID / social-login accounts with `400 ValidationException: profileArn is
required for this request` for at least `claude-opus-5` and
`claude-sonnet-5`. No local store ever caches a `profileArn` for that account
type, `resolveKiroProfileArn()` in `src/oauth/kiro.ts` only reads a cached
value, and `src/adapters/kiro.ts`'s `build()` silently omits `profileArn`
from the wire payload when nothing was cached — so every request from a
Builder ID account against those models fails.

I expected the request to either succeed (matching what the official
`kiro-cli` does for the same account) or fail with a clearer, provider-level
error rather than a raw upstream `ValidationException`.

**Investigation update — a candidate fix I proposed below turned out to be
insufficient; see "What I tried" for why. Filing as a bug report with the
open question rather than a verified fix.**

### Reproduction

1. Configure the `kiro` provider with `authMode: oauth`, importing a
   Builder-ID (not IAM Identity Center) `kiro-cli` session — not an API key
   (`ksk_...`) login.
2. `POST /v1/chat/completions` with `"model": "kiro/claude-opus-5"` (or
   `claude-sonnet-5`) and any prompt.
3. Observe:

```
Provider error 400: Kiro invalid request:
com.amazon.kiro.runtimeservice#ValidationException: profileArn is required for this request.
```

Debug logging (`OCX_DEBUG=1`) on the failing request:

```
[ocx:kiro:request] {"region":"us-east-1","requestedModel":"claude-opus-5","completionMode":"disabled","bodyBytes":362,"messageCount":1,"toolCount":0,"hasProfileArn":false,"wireClient":"cli","hasPreviousResponseId":false}
```

For comparison, the official `kiro-cli` (v2.16.0) against the same account
succeeds on the same model:

```
$ kiro-cli chat --model claude-opus-5 --effort high --no-interactive "Say OK"
> OK
 ▸ Credits: 0.25 • Time: 2s
```

### Root cause

`src/oauth/kiro.ts` — `resolveKiroProfileArn()`:

```ts
export function resolveKiroProfileArn(account?: Pick<KiroOAuthMetadata, "profileArn">): string | undefined {
  if (account !== undefined) return account.profileArn;
  const env = process.env.KIRO_PROFILE_ARN;
[...]

PR #2721, opened 2026-08-27
fix: L3 cherry-pick lane of the 260827 bug-PR merge round (#2639 #2647)

lidge-jun:
## Summary

L3 (cherry-pick) lane of the 260827 bug-PR merge round. Two PRs where only part of the diff was correct, re-applied on a `dev` base with the incorrect part deliberately left behind.

### #2639 — take `status`, leave `created_at`

@bet4it correctly identified that strict Responses decoders require `status` on `OutputMessage`, and an upstream relay that omits it makes such a client fail with `missing field 'status'`. That half is taken as-is, including the choice to map response-level `failed`/`cancelled` to `incomplete` rather than `completed` — claiming `completed` would let a client treat a truncated message as whole.

The `created_at` half is **not** taken. It breaks `tests/server-combo-failover-e2e.test.ts:1323`, which asserts a combo backup response is relayed byte-exact:

```
dev:            74 pass / 0 fail
#2639 branch:   73 pass / 1 fail
  + "created_at": 1787801315,
    "id": "resp-m2",
```

Both contracts cannot hold for the same body — "relay this verbatim" and "add the field the upstream omitted". Which one yields is a real decision, not a detail to fold in beside `status`, so it is left for its own change and recorded in the source comment.

### #2647 — re-apply three effort ladders on a dev base

@darwintree's branch conflicted with `dev` on `src/providers/command-code-efforts.ts`, so the three rows are re-applied here.

Not taken: that branch's copy of `tests/command-code-provider.test.ts` reintroduces `stealth/ox-alpha` and `openai/ox-alpha` into `verifiedImageModels`, but `dev` dropped Ox Alpha entirely in `328931265`. Taking the file wholesale fails that test. Only the new profile-URL assertion travels, extended to cover all three ids.

Provenance is stated plainly in the source: the ladders are the reporter's, all three profile URLs were confirmed to return HTTP 200, but `commandcode.ai` renders those pages client-side so the ladder text could not be read at review time. That is acceptable **only** because `refreshCommandCodeReasoningEfforts()` re-reads the public profile after the first upstream rejection and replaces the row — which is exactly what the added test pins. A wrong ladder self-corrects; a wrong `profileUrl` silently disables that, so the URLs are the thing under test.

## Verification

- `bun x tsc --noEmit` — clean.
- `bun test` across the four affected suites — 146 pass / 0 fail:
[...]

PR #2722, opened 2026-08-27
fix(kiro): send Kiro's service profile for AWS Builder ID accounts

lidge-jun:
## Summary

An AWS Builder ID Kiro account could not complete a single generation request against a gated model. Every attempt failed with `kiro_profile_required`, and the remediation that error suggested — re-login so the profile is captured — could never succeed, because there is no profile to capture. Builder ID is a personal identity with no AWS account behind it, so AWS never mints an account-scoped `arn:aws:codewhisperer:<region>:<account>:profile/<id>` for it. Operators looped on re-login indefinitely.

Follow-up to #993, which was closed by giving that upstream `ValidationException` a stable, actionable code in `src/adapters/kiro-errors.ts`. That improved the diagnosis; this PR removes the cause.

`build()` resolved the account's own profile, got `undefined`, and sent a request with neither a `profileArn` in the payload nor an `x-amzn-kiro-profile-arn` header. The Kiro CLI resolves this the same way this PR does: it carries a fixed service profile on Builder ID requests. The embedded account id is Amazon's own, not the user's, so nothing account-identifying is synthesized — which is the distinction #993 cared about when it added `parseKiroProfileArn` and refused to invent ARNs.

The design separates two questions that used to share one resolver:

- `resolveKiroProfileArn` — *what does this account own?* Unchanged. Region inference, account matching, and continuation scoping keep calling it and keep receiving `undefined`, so the fallback cannot seed a region (it is `us-east-1` and would otherwise pin every Builder ID account there) or become an identity.
- `resolveKiroRequestProfile` — *what do we send?* Used only at request construction. Returns the ARN **and** whether it came from the Builder ID fallback.

The fallback is gated on `authType === "aws_sso_oidc"` rather than on a missing ARN, so a `kiro_desktop` account whose profile import genuinely failed keeps producing its actionable error instead of silently borrowing a service profile that does not describe it. `authType` is derived at snapshot time from the device-registration client pair, so credentials written before the field existed route correctly with no migration and no new persisted field.

Because Builder ID now carries an ARN, a truthy `profileArn` no longer implies "enterprise". The wire path therefore reads the resolver's own verdict instead of re-deriving it. That second commit exists because review caught a real defect in the first: on the accountless path the auth type comes from the local import rather than the request context, so a context-derived guard sent the fallback ARN inside an enterprise IDE envelope — a combination the vendor client never produces. Returning both values from one evaluation makes the two decisions structurally unable to drift apart.

The fallback is never written to `auth.json`: it is computed per request from a constant and never enters `KiroOAuthMetadata`.

## Verification

```
bun test tests/kiro-builder-id-profile.test.ts tests/kiro-adapter.test.ts \
  tests/kiro-oauth.test.ts tests/kiro-review-regressions.test.ts \
  tests/kiro-stream.test.ts tests/core-lab-boundary.test.ts tests/oauth-reauth-bind.test.ts
# 256 pass, 0 fail, 868 expect() calls

bun run typecheck    # clean
bun run privacy:scan # Privacy scan passed
```

`tests/kiro-builder-id-profile.test.ts` adds 9 focused tests: fallback in payload and header, Builder ID staying on the CLI envelope, enterprise account unchanged on the IDE envelope, an SSO OIDC account that *does* own an ARN preferring its own, `kiro_desktop` still sending none, `ksk_` API keys never borrowing it, non-persistence asserted against the raw on-disk store, legacy client-pair derivation, and the accountless SQLite import path.

The accountless regression was driven **red** before acceptance: restoring the previous context-derived guard produces 8 pass / 1 fail; the resolver verdict produces 9 pass / 0 fail.

Live verification through the local proxy on the affected Builder ID account, after `ocx service restart`:

| model | result |
|---|---|
| `kiro/claude-opus-5` | real completion, 6743 prompt tokens |
| `kiro/kiro-auto` | real completion |
| `kiro/claude-sonnet-5` | real completion |
| `kiro/gpt-5.6-sol` | real completion |

Tool-calling returned a well-formed `tool_calls` response and streaming terminated with `finish_reason: stop` plus `[DONE]`. Zero `kiro_profile_required` occurrences in `service.log` after the fix, and neither the fallback ARN nor the account id `638616132270` appears anywhere in `~/.opencodex/auth.json`.

## Checklist

Scope is the Kiro auth/adapter path plus its tests and one plan document; `src/lab/`, routing profiles, other providers, and credential rotation are untouched. No user-facing docs change is needed — the behavior this fixes was never documented as a limitation, and no configuration surface moved. The change touches credential handling, so per `MAINTAINERS.md` it wants explicit security review: the added constant is a public, request-scoped service identifier rather than a credential, no token or secret is logged or serialized, `clientId`/`clientSecret` still never leave the credential store (`authType` is derived from their presence, not their value), and `privacy:scan` is green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Kiro requests for AWS Builder ID accounts that previously failed before generating tokens.
  * Builder ID requests now use the correct service profile and request format.
  * Enterprise accounts, API keys, desktop sign-ins, and existing profiled accounts retain their configured behavior.
  * Service profile details remain request-only and are not saved as account or region information.

* **Tests**
  * Added coverage for Builder ID, enterprise, legacy, accountless, desktop, and API-key scenarios.

### DIFF
--- a/devlog/_plan/260827_kiro_builder_id_profile/010_builder_id_request_scoped_profile.md
+++ b/devlog/_plan/260827_kiro_builder_id_profile/010_builder_id_request_scoped_profile.md
@@ -142,13 +142,22 @@ Builder ID accounts do not collapse into one continuation scope.
 
 ## Verification
 
-- `tests/kiro-adapter.test.ts` — Builder ID context yields the service ARN in
-  both the payload and the header, and stays on the `cli` wire path.
+The regression suite is `tests/kiro-builder-id-profile.test.ts`, kept separate
+from `tests/kiro-adapter.test.ts` so the Builder ID contract reads as one story
+rather than as scattered cases in the general adapter suite.
+
+- Builder ID context yields the service ARN in both the payload and the header,
+  and stays on the `cli` wire path.
 - Regression — enterprise account with its own ARN keeps that ARN and the `ide`
   path; `ksk_` keeps sending no profile.
 - Regression — a `kiro_desktop` account without an ARN still sends none, so the
   actionable failure survives for genuinely broken imports.
-- `bun run typecheck`.
+- Regression — the accountless path, where the auth type comes from the local
+  CLI import rather than the request context, still sends the fallback inside
+  the `cli` envelope. Driven red against the earlier context-derived guard
+  before being accepted.
+- Non-persistence asserted against the raw on-disk store, not a parsed view.
+- `bun run typecheck` and `bun run privacy:scan`.
 - Live: make the Builder ID account (B above) the active Kiro account, restart
   the service so the proxy loads this tree, and capture a real completion.
 
--- a/devlog/_plan/260827_kiro_builder_id_profile/010_builder_id_request_scoped_profile.md
+++ b/devlog/_plan/260827_kiro_builder_id_profile/010_builder_id_request_scoped_profile.md
@@ -1,0 +1,158 @@
+# Kiro Builder ID — request-scoped service profile fallback
+
+Unit: `260827_kiro_builder_id_profile` · work-phase `wp1` · opened 2026-08-27
+
+## Symptom
+
+A Kiro request routed to the Builder ID account fails before any tokens are
+generated:
+
+```text
+Provider error 400: kiro_profile_required: Kiro requires a CodeWhisperer
+profileArn for this account and model. Re-login or re-import the matching Kiro
+account (ocx account login kiro --reauth) so the profile is captured, then retry.
+```
+
+## Why the current remediation cannot work
+
+The message tells the operator to re-login so the profile is captured. For an
+AWS Builder ID account there is nothing to capture. Builder ID is a personal
+identity that is not attached to an AWS account, so AWS never mints an
+account-scoped `arn:aws:codewhisperer:<region>:<account-id>:profile/<id>` for
+it. Re-running `ocx account login kiro --reauth` produces the same credential
+shape it produced before, and the operator loops.
+
+Confirmed against a live `~/.opencodex/auth.json` holding two Kiro accounts.
+Account identifiers and addresses are deliberately omitted; only the
+credential *shape* is load-bearing here:
+
+| account | source | `kiro.profileArn` | `kiro.clientId`/`clientSecret` |
+|---|---|---|---|
+| A (browser OAuth login) | `oauth` | present | absent |
+| B (imported CLI session) | `local-cli` | **absent** | **present** |
+
+Account B is the failing one, and the presence of `clientId` +
+`clientSecret` with no profile ARN is exactly the AWS SSO OIDC / Builder ID
+shape. `src/oauth/kiro-credentials.ts:297` already derives
+`authType: clientId && clientSecret ? "aws_sso_oidc" : "kiro_desktop"` from
+that same pair, so the signal exists — it just never reaches the adapter.
+
+## Mechanism
+
+`src/adapters/kiro.ts` `build()`:
+
+```ts
+const resolvedProfileArn = resolveKiroProfileArn(parsed._kiroAuthContext);
+const isApiKey = provider.apiKey.trim().startsWith("ksk_");
+const profileArn = isApiKey ? undefined : resolvedProfileArn;
+const wireClient: KiroWireClient = isApiKey || !profileArn ? "cli" : "ide";
+...
+if (profileArn) headers["x-amzn-kiro-profile-arn"] = profileArn;
+const built = buildKiroPayload(parsed, profileArn, forcedCompletionMode, wireClient);
+```
+
+`resolveKiroProfileArn` returns `account.profileArn` verbatim when an account
+context is present (`src/oauth/kiro.ts:469`). For the Builder ID account that is
+`undefined`, so the request goes out on the `cli` wire path with **no**
+`profileArn` in the payload and **no** `x-amzn-kiro-profile-arn` header. Gated
+models answer with a `ValidationException` naming `profileArn`, and
+`src/adapters/kiro-errors.ts:112` maps that to the stable non-retryable
+`kiro_profile_required` code. The classifier is doing its job; the request was
+simply incomplete.
+
+## What the reference implementation does
+
+`minpeter/kiro-lb` hit the same wall and resolved it by observing what the real
+Kiro CLI sends. `kiro/config.py`:
+
+```python
+# Builder ID management and generation requests in Kiro CLI 2.19.1 carry this
+# service profile even though the local credential has no account-specific ARN.
+# Keep it request-scoped: it is not persisted as the account's own profile.
+KIRO_BUILDER_ID_PROFILE_ARN = "arn:aws:codewhisperer:us-east-1:638616132270:profile/AAAACCCCXXXX"
+```
+
+`kiro/auth.py` exposes it as a derived request-time property, never as the
+account's stored identity:
+
+```python
+@property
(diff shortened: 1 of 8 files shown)


lidge-jun:
Both findings were valid and are fixed in the force-pushed `0209234e4`.

**Email addresses in the design doc** — this one was a real leak, not a nit. The table was captioned "values redacted" while carrying two live addresses, which is exactly the kind of thing a public `devlog/` must not hold. The addresses and the account-id fragments are gone; the table now identifies the two accounts as A and B by login method, since only the credential *shape* (profile ARN present vs. client pair present) was ever load-bearing for the argument. A stray account-id reference further down in the verification section was removed in the same pass.

The redaction was squashed into the commit that introduced the document rather than added on top, so the addresses do not survive anywhere in this branch's history. `git log -p` across the branch now matches no address pattern, and `bun run privacy:scan` is green.

**MD040 on the error-output fence** — fixed, now tagged `text`.

Code is untouched by this update: `git diff 44516c105 HEAD -- src/ tests/` is empty, so the reviewed implementation is byte-identical and the force-push only rewrites the documentation.

lidge-jun:
Fixed in `0f10e577c`.

The test-path finding was correct: the plan was written before the regression suite got its own file, so it still pointed at `tests/kiro-adapter.test.ts`. The verification section now names `tests/kiro-builder-id-profile.test.ts` and explains why the suite is separate — the Builder ID contract reads better as one story than as scattered cases in the general adapter suite. While there, I added the two cases the original plan did not anticipate: the accountless path (driven red against the earlier guard before acceptance) and the non-persistence assertion against the raw on-disk store.

The email-redaction comment on lines 25-34 is already resolved — it was raised against `44516c105` and fixed in the force-pushed `0209234e4`, which is why the suggestion diff still shows the old table. The current file has no addresses or account-id fragments, and they are absent from this branch's history rather than merely removed on top.

Docs-only change, so the reviewed implementation is untouched.

lidge-jun:
## 리뷰 · 우선순위 72 / 80

지금 `dev` HEAD는 64c6d642b이다. 방금 #2721 L3가 들어가서 status 백필과 command-code effort 사다리는 고쳐졌지만, Kiro Builder ID 요청은 그대로다. 현재 `src/adapters/kiro.ts` `build()`는 `resolveKiroProfileArn`만 본다. AWS Builder ID 계정은 뒤에 AWS 계정이 없어서 계정 전용 `profileArn`을 받지 못한다. 그래서 요청 몸체에도 헤더에도 프로필이 없고, 게이트된 모델이 ValidationException을 낸다. 어댑터는 그걸 `kiro_profile_required`로 바꾸고 "다시 로그인해서 프로필을 받아라"고 한다. 다시 로그인해도 받을 프로필이 없다. 운영자가 같은 벽에 계속 부딪힌다.

이 PR은 그 원인을 없앤다. #993이 에러 코드만 안정적으로 만들어 준 다음 단계다. Kiro CLI가 Builder ID 요청에 고정 서비스 프로필을 실어 보내는 것과 같은 일을 한다. 중요한 분리는 두 질문이다. `resolveKiroProfileArn`은 "이 계정이 가진 것"만 답하고, 지역 추론과 계정 비교와 continuation 범위는 예전처럼 `undefined`를 본다. 새로 생긴 `resolveKiroRequestProfile`만 "이번에 실제로 보낼 것"을 답한다. 폴백은 `authType === aws_sso_oidc`일 때만 켜진다. 프로필 import가 실패한 `kiro_desktop` 계정은 예전처럼 큰 소리로 실패한다. 없는 ARN만 보고 기업 프로필을 빌려 주지 않는다.

`authType`은 비밀이 아니다. `accessSnapshot`이 `clientId`와 `clientSecret` 쌍이 있는지로 지금 계산한다. 예전에 저장된 자격 증명에 필드가 없어도 마이그레이션 없이 동작한다. 이 신호는 라우팅 환경 값과 섞지 않고 나중에 붙인다. 그래서 클라 쌍만 있는 계정이 "이미 라우팅이 있다"고 오해되어 `environmentKiroRoutingMetadata()`를 꺼 버리지 않는다. `src/server/responses/core.ts`는 이미 `snapshot.kiro`를 그대로 펼친다. 이 PR이 core.ts를 안 고친 것은 구멍이 아니라, 타입이 넓어지면 authType이 자동으로 따라가기 때문이다.

폴백 ARN은 요청할 때만 계산되고 `auth.json`에 쓰지 않는다. 상수가 us-east-1이라서 계정 지역을 그쪽으로 고정하지 않고, continuation 해시가 여러 Builder ID 계정을 하나로 합치지도 않는다. 테스트 9개가 페이로드/헤더, CLI 봉투, 기업 계정, 자기 ARN 우선, desktop 실패 유지, API 키, 저장 안 함, 레거시 클라 쌍, 계정 없는 import 경로를 본다. 마지막 케이스는 예전 가드를 되돌리면 깨지도록 먼저 빨갛게 만들었다. types.ts/config.ts 분할 캠페인과는 무관한 어댑터/자격 경로라서 분할 때문에 닫을 대상은 아니다. 미리보기 배포도 이 계획에 없다.

라인 단위로 보면 현재 `dev`와 비교해서 이런 점이 남는다.

라인 kiro-constants.ts KIRO_BUILDER_ID_SERVICE_PROFILE_ARN - Amazon 계정 638616132270과 프로필 토큰이 소스에 박혀 있다. 요청 전용 상수이고 라이브 완료도 나왔다고 하지만, 벤더가 이 값을 바꾸면 모든 Builder ID 운영자가 한꺼번에 깨진다.

라인 oauth/index.ts accessSnapshot kiroAuthType - clientId와 clientSecret만 있으면 aws_sso_oidc로 본다. Identity Center 기업 SSO도 같은 쌍을 쓴다. 프로필 import가 실패한 기업 계정은 Builder ID 서비스 프로필을 빌려 쓰게 된다. desktop만 보호하고 기업 실패는 조용히 우회한다.

라인 oauth/kiro.ts resolveKiroRequestProfile - account 객체가 있으면 authType만 보고, 로컬 import로 내려가지 않는다. core.ts가 빈 객체 {}를 컨텍스트로 넣으면 폴백이 꺼진다. 계정 없는 경로(undefined)와 빈 객체 경로가 다르게 동작한다.

경로 tests/kiro-builder-id-profile.test.ts - seedKiroCliBuilderIdSession 함수 다음에 import가 끼어 있다. 테스트는 통과하지만 파일 위쪽이 한 이야기로 안 읽힌다.

메인테이너의 판단이 필요한 지점
- Amazon 서비스 ARN을 리포지토리 상수로 둘지, 나중에 설정/원격 조회로 뺄지. 보안 리뷰 체크는 이미 적혀 있다.
- aws_sso_oidc를 Builder ID와 같은 키로 써도 되는지. 기업 SSO의 실패한 import를 이 폴백으로 살리는 것이 맞는지.
- #993 후속으로 지금 바로 머지할지. 미리보기/태그는 하지 않는다.

너의 추천
CI와 적힌 테스트(kiro-builder-id-profile, kiro-adapter, kiro-oauth, kiro-review-regressions, core-lab-boundary, oauth-reauth-bind)가 초록이면 #2722를 `dev`에 머지한다. 기업 SSO 오탐은 후속 이슈로 남긴다. 상수 ARN은 이번엔 두고, 벤더가 바꾸면 한 줄만 고치게 주석을 유지한다. 이 PR에 다른 레인을 섞지 말고, 미리보기 배포와 버전 태그도 하지 않는다.

이 댓글은 grok-bot이 작성했습니다

PR #2745, opened 2026-08-27
fix(responses): rebind credential identity on every OAuth 429 rotation

lidge-jun:
## Summary

Rotating a credential without rotating the **identity** of that credential left the two naming different accounts, and both consumers of that identity then misbehaved.

`applyFailoverSnapshot` swapped the bearer, the Copilot origin, the Antigravity project and the Kiro routing metadata, but never updated `sentOAuthSnapshot` or `replayOAuthCredentialSnapshot`:

- `sentOAuthSnapshot` feeds the 401 forced refresh, and `forceRefreshOAuthAccessSnapshot` refreshes the **snapshot's** account rather than whichever account is on the route now. A later 401 therefore renewed the account the request had just rotated away from.
- `replayOAuthCredentialSnapshot` scopes the reasoning replay cache, so a stale value let continuation state minted under one account be addressed under another.

Both rebinds now live inside the helper. Three rotation sites already disagreed about which of them to update, and the one that got it right (runTurn) got it right by hand — a fourth site would have to remember too.

The other two sites are brought up to runTurn's standard. The HTTP 429 path had no replay rebind at all; `sealRequestAttemptIdentity` records what happened but does not re-scope the cache. The sidecar bound *without* the credential snapshot, which fails the OAuth replay key closed — safe-looking, but it leaves the scope with no account identity to distinguish.

Both 401 rebuilds now prefer `refreshed.apiBaseUrl` over `getOAuthCredentialApiBaseUrl`. The latter reads the **active** credential, and generic 429 failover never promotes the active account, so after a rotation it still named the rate-limited one — pairing a refreshed bearer with the wrong account's origin.

## Reachability — please read before assessing severity

An independent audit was explicit that the exploitable version of this is **not reachable on today's control flow**. A 401 cannot follow a 429 rotation back into the refresh path: the HTTP 429 branch does not `continue recovery`, and Copilot Responses models take a passthrough return with no generic rotator at all. So what exists today is latent identity drift, not a live cross-account send.

That missing `continue recovery` is an **accidental** guard, not a designed one. Adding it — an obvious future improvement to 429 recovery — would activate the defect. That is the argument for fixing the identity first, and for not treating this as urgent.

## Verification

- `bun x tsc --noEmit` clean.
- `bun test ./tests/generic-oauth-failover.test.ts ./tests/adapter-event-oauth-failover.test.ts ./tests/server-xai-oauth-401-replay.test.ts` — 31 pass, 0 fail, 131 expect() calls.
- **Each of the three new assertions driven red on its own**, by reverting only its own fix: removing the identity rebind, removing the HTTP replay bind, and reverting one of the two 401 transport sites.
- The replay-scope assertion is bounded to each rotation's own recovery call. An unbounded forward search finds the *next* site's bind and passes with this site's bind deleted — that weaker form was written first and survived the deletion, which is why the bound exists.
- Full suite on the remote host at this exact head before any merge.

## Maintainer security review requested

This touches OAuth credential handling and account-scoped transport resolution, so per `MAINTAINERS.md` it needs explicit security review. I have deliberately **not** applied `maintainer-sponsored` and am not merging it. The reachability analysis above is the part most worth a second opinion: if you disagree that the missing `continue recovery` is load-bearing, the severity assessment changes.

## Checklist

- [x] Targets `dev`
- [x] Focused regression tests added, each driven red once
- [x] No new logging of tokens, accounts, or request bodies
- [x] Left for maintainer decision rather than self-merged

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OAuth account failover consistency during response processing.
  * Cleared stale Cursor conversation and continuation state after account rotation.
  * Ensured refreshed GitHub Copilot connections use the latest available API endpoint.
  * Improved recovery after rate limits and authentication failures by correctly reapplying account-specific routing and replay context.

### DIFF
--- a/src/server/responses/core.ts
+++ b/src/server/responses/core.ts
@@ -2854,6 +2854,17 @@ async function handleResponsesInner(
     // got it right by hand. A fourth site would have to remember too. Here it cannot forget.
     if (isOAuth401ReplayProvider) sentOAuthSnapshot = snapshot;
     replayOAuthCredentialSnapshot = { accountId: snapshot.accountId, generation: snapshot.generation };
+    // A Cursor conversation and identity scope are credential-scoped in the same way. runTurn
+    // cleared them by hand and the other two sites did not, so the same rotation that swapped the
+    // bearer could carry the previous account's conversation forward - and for image turns
+    // src/images/loop.ts copies that id back onto the outer request, which is how it outlives the
+    // rotation. Clearing here means a rotated adapter always derives its own.
+    parsed._cursorIdentityScope = undefined;
+    parsed._cursorConversationId = undefined;
+    if (parsed._providerContinuation?.cursor) {
+      const { cursor: _discardedCursor, ...otherProviderState } = parsed._providerContinuation;
+      parsed._providerContinuation = otherProviderState;
+    }
     return true;
   };
   const anthropicSessionKey = route.providerName === "anthropic" && route.provider.authMode === "oauth"
@@ -4647,15 +4658,6 @@ async function handleResponsesInner(
         genericFailoverAccountId = nextAccountId;
         genericFailovers += 1;
         if (!applyFailoverSnapshot(snapshot)) return false;
-        // A Cursor conversation/checkpoint is credential-scoped. The failed attempt emitted no
-        // client-visible bytes, so replay is safe, but carrying its account identity into the next
-        // account would not be. Let the rotated adapter derive a fresh identity and conversation.
-        parsed._cursorIdentityScope = undefined;
-        parsed._cursorConversationId = undefined;
-        if (parsed._providerContinuation?.cursor) {
-          const { cursor: _discardedCursor, ...otherProviderState } = parsed._providerContinuation;
-          parsed._providerContinuation = otherProviderState;
-        }
         const rotatedProvider = resolveWireProtocolOverride(
           route.providerName,
           route.modelId,
--- a/src/server/responses/core.ts
+++ b/src/server/responses/core.ts
@@ -2841,6 +2841,19 @@ async function handleResponsesInner(
     if (snapshot.projectId) rotatedProvider = { ...rotatedProvider, project: snapshot.projectId };
     route.provider = rotatedProvider;
     if (route.providerName === "kiro") parsed._kiroAuthContext = { ...(snapshot.kiro ?? {}) };
+    // Rotating the credential without rotating the IDENTITY of that credential leaves the two
+    // describing different accounts. `sentOAuthSnapshot` feeds the 401 forced refresh, which
+    // refreshes the SNAPSHOT's account rather than whichever account is on the route now — so a
+    // stale snapshot makes a later 401 renew the account we just rotated AWAY from, while the
+    // transport still points at the rotated one. `replayOAuthCredentialSnapshot` scopes reasoning
+    // replay, so a stale value lets continuation state minted under one account be replayed under
+    // another.
+    //
+    // This lives in the helper rather than at each call site on purpose: three rotation sites
+    // already disagreed about which of these to update, and the one that got it right (runTurn)
+    // got it right by hand. A fourth site would have to remember too. Here it cannot forget.
+    if (isOAuth401ReplayProvider) sentOAuthSnapshot = snapshot;
+    replayOAuthCredentialSnapshot = { accountId: snapshot.accountId, generation: snapshot.generation };
     return true;
   };
   const anthropicSessionKey = route.providerName === "anthropic" && route.provider.authMode === "oauth"
@@ -3538,7 +3551,7 @@ async function handleResponsesInner(
         route.providerName,
         { ...route.provider, apiKey: refreshed.accessToken },
         parsed.options.promptCacheKey,
-        route.providerName === "github-copilot" ? getOAuthCredentialApiBaseUrl(route.providerName) : undefined,
+        route.providerName === "github-copilot" ? (refreshed.apiBaseUrl ?? getOAuthCredentialApiBaseUrl(route.providerName)) : undefined,
       );
       route.provider = refreshedProvider;
       const refreshedAdapter = resolveAdapter(
@@ -4341,6 +4354,11 @@ async function handleResponsesInner(
       providerName: route.providerName,
       provider: route.provider,
       adapterName: rotatedAdapter.name,
+      // Omitting this fails the OAuth replay key CLOSED, which looks safe and is not: the scope
+      // then carries no account identity at all, so it cannot distinguish state minted under the
+      // account we just rotated away from. `applyFailoverSnapshot` now keeps this value in step
+      // with the rotated credential, so passing it is what makes the scope account-accurate.
+      oauthCredentialSnapshot: replayOAuthCredentialSnapshot,
     });
     return rotatedAdapter;
   };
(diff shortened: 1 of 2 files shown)


lidge-jun:
## 리뷰 · 우선순위 76 / 80

이 PR은 현재 `dev`에서 OAuth 429 페일오버가 토큰만 바꾸고 신원은 안 바꾸던 구멍을 막습니다. `applyFailoverSnapshot`이 bearer·Copilot origin·Antigravity project·Kiro 메타는 바꾸면서 `sentOAuthSnapshot`과 `replayOAuthCredentialSnapshot`은 그대로 두면, 이후 401 강제 리프레시가 방금 떠난 계정을 갱신하고, 추론 리플레이 캐시는 다른 계정 상태를 주소로 삼을 수 있습니다. 세 회전 지점(runTurn / HTTP 429 / sidecar)이 서로 다르게 동작하던 것을 헬퍼 안으로 모으고, Cursor conversation/identity도 헬퍼에서만 비웁니다. Copilot 401 재구성은 `refreshed.apiBaseUrl`을 먼저 쓰게 바꿉니다. 본문이 밝히듯, 오늘 제어 흐름에서는 429 뒤 401이 같은 경로로 이어지지 않아 실제 교차 계정 전송은 잠복 결함입니다. 그래도 `continue recovery`를 나중에 넣으면 바로 살아나는 종류라서, 신원부터 맞추는 선택은 타당합니다. 테스트 세 개가 헬퍼·세 사이트·apiBaseUrl을 각각 빨갛게 만든 뒤 고정합니다.

라인 applyFailoverSnapshot 내부 - `sentOAuthSnapshot = snapshot`이 `isOAuth401ReplayProvider`일 때만 됩니다. 비-401 리플레이 프로바이더에서 스냅샷이 의도적으로 남는 건지 확인이 필요합니다.
라인 HTTP 429 bindRouteReasoningReplayScope - `authCtx`/`selectedForwardHeaders`가 그 시점 스코프에 있는지, runTurn·sidecar와 인자 집합이 같은지 보세요.
테스트 coreSource 문자열 검색 - 리팩터로 헬퍼 이름이 바뀌면 테스트만 깨집니다. 동작 테스트와 병행하는 현재 방식은 괜찮지만, 마커 문자열을 너무 짧게 두지 마세요.
본문 보안 리뷰 - 메인테이너 보안 리뷰를 요청했고 `maintainer-sponsored`를 안 붙였습니다. 절차상 맞습니다.

메인테이너의 판단이 필요한 지점
- 잠복 결함을 지금 머지할지, 429 후 continue recovery 설계와 묶을지.
- `sentOAuthSnapshot`을 isOAuth401ReplayProvider에만 묶는 조건이 모든 제네릭 OAuth 프로바이더에 맞는지.
- MAINTAINERS.md 보안 리뷰를 누가 서명할지.

너의 추천
머지 후보로 둡니다. 보안 리뷰 한 번만 받고, 가능하면 HTTP 429 경로에 회전 후 sent/replay 스냅샷 accountId가 새 계정인지 동작 단언을 한 줄 더 넣은 뒤 `dev`에 올리면 됩니다. 태그/배포는 하지 마세요. package.json은 이미 2.34.0입니다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
Security-boundary review at exact head `a90ab6ee7affaaf5db44bfc14aafe86092d9fa51`.

The central identity rebind is the right direction, and the existing exact-head platform CI is green. My isolated focused run also passed 48/48 across `generic-oauth-failover`, `adapter-event-oauth-failover`, `server-xai-oauth-401-replay`, and `github-copilot-oauth`. Two blockers remain before this can serve as the required credential-boundary regression.

1. The refreshed Copilot fallback is still active-account scoped. At both 401 rebuild sites, `refreshed.apiBaseUrl ?? getOAuthCredentialApiBaseUrl(route.providerName)` reads the active credential when the refreshed snapshot for account B has no allowlisted origin. Generic 429 rotation does not promote B to active, so this reconstructs B bearer plus A origin. The same missing-metadata case also needs auditing in `applyFailoverSnapshot`, where an undefined snapshot origin can leave the previous routed base URL in the provider object. Resolve by account id using `refreshed.accountId`, or fail closed to the canonical Copilot origin without consulting or retaining another account route. Add a negative regression with A and B origins where B is a legacy credential with no `apiBaseUrl`.

2. The three new tests are source-text assertions, not request-path behavior. They pass if the assignments are unreachable, if the wrong snapshot reaches the rebuilt request, or if the resulting replay scope is still owned by A. Add an executable A -> 429 -> B regression through the HTTP recovery path or an extracted helper seam. Assert the second dispatch bearer/origin pair, the B account/generation replay identity, and cleared Cursor continuation state. Bound the structural window to the next `applyFailoverSnapshot(snapshot)` marker as the existing inline review requests, but keep that only as a topology guard alongside the behavioral test.

I checked the Grok review questions too: keeping `sentOAuthSnapshot` conditional on `isOAuth401ReplayProvider` is consistent because non-replay providers do not consume it, while the replay snapshot still rebinds for all generic OAuth rotations; the HTTP scope arguments are available at that call site. The known redirect-policy comment is broader pre-existing request transport behavior and is not the reason for this review state.

Please rebase the follow-up onto current `dev@7ca954ffd997197d1cff6fc6d69842be51177a8f`, preserve the full CI result, and update the transport decision note with the account-to-origin invariant. This remains a strong merge candidate after those bounded fixes.

lidge-jun:
Both review blockers are closed on `codex/oauth-failover-identity-v2` (`2b3574a45`), which carries this PR's two commits onto current `dev`. I authored it, so approval is not mine to give — posting the evidence for whoever reviews the credential boundary.

## Blocker 1 — the account-blind fallback

Your reading was exactly right, and it is worth stating what made it invisible: the defect was one `??`.

```
refreshed.apiBaseUrl ?? getOAuthCredentialApiBaseUrl(route.providerName)
```

`getOAuthCredentialApiBaseUrl` is `validateCopilotApiBaseUrl(getCredential(provider)?.apiBaseUrl)` — it reads the **active** credential, with no account scoping. Generic 429 rotation never promotes the account it rotated to, so for a legacy account B carrying no allowlisted `apiBaseUrl`, the right-hand arm silently reached account A. B's bearer, A's origin.

Fixed by `copilotOriginForRefreshedCredential`, which resolves from the refreshed snapshot alone and otherwise fails closed to the canonical origin. It never consults, and never retains, another account's route — the direction you asked for. Behaviour across every arm:

| refreshed snapshot for B | resolved origin |
|---|---|
| own allowlisted origin | that origin |
| legacy, no origin | canonical — **never A's** |
| non-allowlisted origin | canonical |
| empty string | canonical |

## Blocker 2 — the source-text assertions

You were right that they were not behavioural, and it was worse than that: one of them asserted the *defect existed*.

```js
const refreshSites = coreSource.match(/refreshed\.apiBaseUrl \?\? getOAuthCredentialApiBaseUrl/g) ?? [];
expect(refreshSites.length).toBe(2);
```

That passes whether the assignment is reachable, whether the right snapshot arrives, and whether the resulting bearer/origin pair is correct. Fixing the bug would have *broken* it.

Replaced with a behavioural test that drives the resolver across all four arms above, plus a topology guard that strips comments before scanning — the new helper's doc comment quotes the removed expression to explain why it was wrong, and a naive scan reads that explanation as the defect. I hit that while writing it.

## Verification

    bun x tsc --noEmit                        exit 0
    tests/generic-oauth-failover.test.ts      19 pass / 0 fail
    full suite (remote)                       15358 pass / 0 fail, rc=0

Mutation-verified: restore the account-blind fallback and the guard fails (18 pass / 1 fail); with the fix, 19 / 0.

## Still open from your review

The `applyFailoverSnapshot` audit you asked for — where an undefined snapshot origin can leave the previous routed base URL on the provider object — and the executable A -> 429 -> B regression through the HTTP recovery path asserting the second dispatch's bearer/origin pair and replay identity. Both are worth doing; neither is closed here, and I would rather say so than let the two closed blockers imply a clean sweep.

lidge-jun:
Closing as superseded by #2807.

The same OAuth 429 identity rebind was rebased onto current `dev` as `codex/oauth-failover-identity-v2`. This thread is the older `CHANGES_REQUESTED` review; continue on #2807.

PR #2747, opened 2026-08-27
fix(tests): reap the recovery proxy instead of trusting `stop`

olddonkey:
## Summary

The update-recovery case in `tests/update-stop-first.test.ts` starts a real detached proxy, and its cleanup ran the reap **only when** `node ocx.mjs stop` exited non-zero. That exit code is a claim, not proof: `stop` also reports success when it finds no live runtime to stop, which is indistinguishable here from one it failed to stop. On that path nothing killed the proxy, and the block went straight on to `rmSync` the fixture tree.

## What that produced

The survivor was reparented to init with its package tree, `src` symlink, and config deleted underneath it — unable to serve, unable to exit. One escapee was found sitting at **99% of a core for 3h38m, 202 minutes of CPU time**, listening on nothing:

```
14845  PPID=1  node .../T/ocx-update-recovery-l1bDDD/.../bin/ocx.mjs start --port 64113
14848  99% CPU .../bun.exe .../src/cli/index.ts start --port 64113
```

The fixture directory was already gone, which is what made it unrecoverable rather than merely leaked. It also held the test machine lock, which is how it surfaced at all.

## The fix

Cleanup now resolves the pid **before** anything destroys the record of it (`stop` rewrites `runtime-port.json`; the `rmSync` deletes it), runs `stop` for its graceful path, then verifies liveness and reaps **regardless of what `stop` claimed**. `rmSync` moves inside a nested `finally` so it still always runs, but strictly after the reap — deleting the tree out from under a live detached process is what turned a missed kill into a permanently spinning orphan.

Reaping the recorded pid is sufficient because `bin/ocx.mjs` mirrors its Bun child's exit (`child.on("exit", …)`), so the node launcher follows it down. Real `ocx stop` is unaffected.

## Why a regression guard was needed too

Reaping alone would still fail silently: **the case passed the entire time it was leaking**. An `afterAll` now audits the pid, so a future regression is red instead of invisible.

## Verification

A/B against the exact failure mode, with `stop` stubbed to exit 0 without killing anything:

| cleanup | result | orphan left |
| --- | --- | --- |
| original | 15 pass / 0 fail | **node + bun survive** |
| this PR | 15 pass / 0 fail | none |
| original + the new guard | **1 fail** | — |

- `bun run test` — **15163 pass, 12 skip, 0 fail, exit 0**; all six serial lanes green.
- Related lifecycle suites (`update-stop-first`, `ocx-launcher-runtime`, `process-control-graceful`, `grok-lifecycle`) — 43 pass / 0 fail.

`tests/ocx-launcher-runtime.test.ts` already used the safe pattern (kill unconditionally, collect cleanup errors so they cannot mask a real failure); this change follows it. The other two `killProxy` callers are source-text assertions that start no process.

## Checklist

### Re-verified at head `4a0cbb55c` (base `50e955604`)

`bun run test` — 15191 pass, 12 skip, **1 fail**; all six serial lanes green.

The single failure is `key-login-live-update`, which **reproduces on unmodified `dev` @ `50e955604`
with zero changes** and is unrelated to this PR. Root-caused separately: the NAT64 well-known
prefix `64:ff9b::/96` was classified as a non-global address, so `providerDestinationResolvedError`
answered `409` to the live provider reload. Fixed in #2798; with that branch applied the suite is
15196 pass / 0 fail.

An earlier run of this same head reported 9 failures under a loaded machine (load average 26).
Re-running the four affected files at normal load gave 108 pass / 1 fail — the other 8 were 5s
timeouts, not defects.

## Review readiness checklist

## Summary by CodeRabbit

- **Tests**
  - Improved recovery test cleanup to consistently detect and stop detached processes.
  - Added suite-level verification that recovery processes are no longer running after tests complete.
  - Ensured temporary test files are removed only after associated processes have been safely terminated.

### DIFF
--- a/tests/update-stop-first.test.ts
+++ b/tests/update-stop-first.test.ts
@@ -1,10 +1,10 @@
-import { describe, expect, test } from "bun:test";
+import { afterAll, describe, expect, test } from "bun:test";
 import { chmodSync, copyFileSync, existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, symlinkSync, writeFileSync } from "node:fs";
 import { createServer } from "node:net";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 import { runNpmCachePreflight } from "../src/update/npm-cache-preflight.mjs";
-import { killProxy } from "../src/lib/process-control";
+import { isProcessAlive, killProxy } from "../src/lib/process-control";
 
 const repoRoot = join(import.meta.dir, "..");
 
@@ -63,6 +63,16 @@ const serverSource = readFileSync(join(import.meta.dir, "..", "src", "server", "
 const dispatchSource = readFileSync(join(import.meta.dir, "..", "src", "cli", "dispatch.ts"), "utf8");
 
 describe("update stops the running proxy before replacing files", () => {
+  // The recovery case starts a real detached proxy, and its own result says nothing about
+  // whether cleanup reaped it — it stayed green while an escapee spun on a deleted tree for
+  // hours. Auditing the pid once the suite is done turns a silent leak back into a red test.
+  let auditedRecoveryPid: number | undefined;
+
+  afterAll(() => {
+    if (auditedRecoveryPid === undefined) return;
+    expect(isProcessAlive(auditedRecoveryPid)).toBe(false);
+  });
+
   test("a failed cache pre-flight aborts before the stop callback can run", () => {
     let stopped = false;
     const malformedSpawn = (() => ({ status: 0, signal: null, stdout: "not-json", stderr: "" })) as never;
@@ -227,24 +237,42 @@ esac
         expect(runtime.pid).toBeGreaterThan(0);
         recoveredPid = runtime.pid;
       } finally {
-        const stopped = existsSync(launcher)
-          ? Bun.spawnSync(["node", launcher, "stop"], {
-              cwd: root,
-              env,
-              stdout: "ignore",
-              stderr: "ignore",
-              timeout: UPDATE_SPAWN_TIMEOUT_MS,
-            })
-          : null;
-        if (stopped?.exitCode !== 0) {
-          if (!recoveredPid) {
-            try {
-              recoveredPid = JSON.parse(readFileSync(join(opencodexHome, "runtime-port.json"), "utf8")).pid;
-            } catch { /* the proxy never wrote runtime state */ }
-          }
-          if (Number.isSafeInteger(recoveredPid) && recoveredPid! > 0) killProxy(recoveredPid!);
+        // Resolve the pid FIRST. `stop` rewrites runtime-port.json and the rmSync below
+        // deletes it outright, so this is the last moment the detached proxy the recovery
+        // path started can still be identified at all.
+        if (!recoveredPid) {
+          try {
+            recoveredPid = JSON.parse(readFileSync(join(opencodexHome, "runtime-port.json"), "utf8")).pid;
+          } catch { /* the proxy never wrote runtime state */ }
         }
-        rmSync(root, { recursive: true, force: true });
+        auditedRecoveryPid = Number.isSafeInteger(recoveredPid) && recoveredPid! > 0
+          ? recoveredPid
+          : undefined;
+        if (existsSync(launcher)) {
+          Bun.spawnSync(["node", launcher, "stop"], {
+            cwd: root,
+            env,
+            stdout: "ignore",
+            stderr: "ignore",
+            timeout: UPDATE_SPAWN_TIMEOUT_MS,
+          });
+        }
+        try {
+          // `stop` exiting 0 is a claim, not proof: it also reports success when it finds no
+          // live runtime to stop, which is indistinguishable here from one it failed to stop.
+          // Gating the reap on that exit code let a detached proxy survive, get reparented to
+          // init, and then spin on a fixture tree this same block had already deleted — one
+          // escapee burned a full core for hours. Verify liveness and reap regardless.
+          // bin/ocx.mjs mirrors its Bun child's exit, so reaping the recorded child pid takes
+          // the node launcher with it.
+          if (Number.isSafeInteger(recoveredPid) && recoveredPid! > 0 && isProcessAlive(recoveredPid!)) {
+            killProxy(recoveredPid!);
+          }
+        } finally {
+          // Ordered after the reap on purpose: deleting the tree out from under a live
+          // detached proxy is what turned a missed kill into a permanently spinning orphan.
+          rmSync(root, { recursive: true, force: true });
+        }
       }
     },
     RECOVERY_CASE_TIMEOUT_MS,


lidge-jun:
## 리뷰 · 우선순위 72 / 80

이 PR은 현재 `dev`의 `tests/update-stop-first.test.ts` 복구 케이스 정리 순서를 고칩니다. 예전에는 `node ocx.mjs stop`이 0이 아닐 때만 `killProxy`를 호출했습니다. `stop`은 살릴 런타임이 없어도 성공으로 보고할 수 있어서, 그 경로에서는 프록시를 안 죽이고 바로 fixture `rmSync`로 갔습니다. 살아남은 프로세스는 init에 입양된 채 지워진 트리 위에서 CPU를 태웠고, 테스트는 계속 초록이었습니다. 이번 변경은 pid를 stop/삭제 전에 읽고, stop 후 `isProcessAlive`면 무조건 재프하며, `rmSync`는 그 다음 finally로 옮깁니다. afterAll이 감사 pid가 죽었는지 검사해서 같은 침묵 누수를 다시 빨갛게 만듭니다. 런타임 제품 코드는 건드리지 않고 테스트만 고칩니다. `isProcessAlive`/`killProxy`는 이미 `src/lib/process-control.ts`에 있습니다.

라인 auditedRecoveryPid - 복구 케이스가 스킵되거나 pid를 못 읽으면 afterAll이 그냥 return합니다. 그 경우 재프가 돌았다는 보장은 없습니다. 의도된 완화인지 주석에 남겨 두세요.
라인 killProxy(recoveredPid) - 기록된 자식 pid만 죽입니다. 본문대로 node 런처가 Bun 자식을 따라 죽으면 충분합니다. 반대로 Bun만 남고 node가 남는 경로가 있으면 afterAll이 잡아야 합니다.
라인 stop 호출 유지 - 정상 graceful 경로는 그대로 두고, 실패·허위 성공만 재프로 보완합니다. 제품 `ocx stop` 동작은 바꾸지 않습니다.
검증 표 - stop을 가짜 성공으로 둔 A/B가 문제 재현과 가드 효과를 잘 보여 줍니다.

메인테이너의 판단이 필요한 지점
- 같은 stop-exit에만 재프 패턴이 다른 테스트에 더 있는지 한 번 더 훑을지.
- afterAll 감사를 이 파일에만 둘지, 공통 헬퍼로 올릴지.
- 테스트 전용 수정이라 보안 리뷰는 불필요로 보면 되는지.

너의 추천
머지해도 됩니다. 가능하면 저장소에서 `killProxy`를 exitCode에 가둔 다른 테스트를 한 번 검색해 같은 함정이 없으면 `dev`에 바로 들이세요. 태그/퍼블리시는 하지 마세요.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (commented):
Exact-head status for `07b9758739e29f6283519577f1e4a6df629cfaa7`: the focused 15/15 regression passed, and the patch remains narrowly limited to ordering recovery-proxy identification, reaping, and fixture removal in `tests/update-stop-first.test.ts`. I do not see a code blocker in that diff.

I am not approving yet because the required macOS job failed twice, including the failed-job rerun. Both failures are the same untouched assertion in `tests/codex-shim.test.ts`: `Unix install rejects delayed detached redispatch after the launcher closes its lease fd` accepted the launcher (`Expected: false`, `Received: true`). The base `a57b9620aee4ca01bc426508a6661bdb9c95b5b6` had green cross-platform CI, and this PR's modified test runs later than the failing shim test, so I am not requesting changes to this patch for that failure. However, repository policy still requires successful required CI before approval/merge. Please leave this open until an exact-head macOS run is green or the independent shim flake is repaired and the PR is rerun.

lidge-jun (approved):
Approving at exact head 07b9758739e29f6283519577f1e4a6df629cfaa7 as a non-author maintainer.

This is the right shape of fix for a flaky test: it repairs the cause rather than making the symptom green.

Three things the diff gets right:

1. It treats `stop` exiting 0 as a claim rather than proof. The old code only reaped when `stop` reported failure, but `stop` also exits 0 when it finds no live runtime — indistinguishable here from one it failed to stop. That gap is what let a detached proxy survive.

2. The pid is resolved BEFORE `stop` runs. `stop` rewrites runtime-port.json and the cleanup deletes it, so this genuinely is the last moment the detached proxy can be identified. Ordering the reap ahead of `rmSync` is what stops a missed kill from becoming a proxy spinning on a deleted tree.

3. The `afterAll` liveness audit is the part that makes it a real regression rather than a hopeful cleanup. Previously the case stayed green while leaking; now a surviving pid fails the suite. A cleanup with no assertion cannot tell you it stopped working.

CI state: the `macos` and `ci` failures on this PR were never its own. Job 98534630924 fails on `release version line`, the repository-wide assertion that `package.json` 2.34.0 equalled the published v2.34.0 tag, and `ci` is the fan-in over it. That baseline is now repaired on `dev` by #2766. Rebasing onto current `dev` should clear both without touching this diff.

Scope: one test file, no `src/` change, no shared path with any other open bug PR in this round.

lidge-jun:
Approved above. The remaining red is not yours, but it needs a rebase from you rather than a re-run from me.

I re-ran the failed jobs after #2766 landed, and they came back red for the same reason: a re-run replays the same commit, so it still carried `package.json` 2.34.0 and still failed `release version line`. Job 98573888840 shows exactly that:

    error: package.json version 2.34.0 equals release tag v2.34.0, but this
    commit is not the one that tag names.

That assertion was failing repository-wide after the v2.34.0 release train — `dev` still carried the released version, so every PR opened after the tag inherited a red `test 3/4`/`macos`, with `ci` as the fan-in over them. #2766 repaired it on `dev` (now 2.35.0).

The controlled proof that this is environmental: #2764 and #2767 had the identical failure signature, were rebased onto post-#2766 `dev` with **no change to their own diffs**, and went from four failing required jobs to 26 success / 0 failures. Nothing in their code changed.

So: **please rebase onto current `dev`** (`50e955604` or later) and push. Both jobs should go green with your diff untouched. I did not rebase it for you because this PR's head lives on your fork — pushing there from here would rewrite your branch, and my first attempt instead created a stray same-named branch on the origin, which I deleted immediately.

On the change itself, from the review: treating `stop` exit 0 as a claim rather than proof is the right diagnosis, resolving the pid before `stop` rewrites `runtime-port.json` is the right ordering, and the `afterAll` liveness audit is what turns a silent leak into a red test instead of a hopeful cleanup.

lidge-jun:
Heads up: I rebased this branch onto current `dev` and force-pushed it — `07b975873` -> `4a0cbb55c`. Flagging it explicitly because it is your branch and a silent maintainer push is not okay.

Why: the `ci`/`macos` red on this PR was never yours. It was the repository-wide `release version line` failure — `dev` still carried `package.json` 2.34.0 after tag v2.34.0 shipped, so every PR opened after the release inherited it. #2766 repaired that on `dev`. I tried `gh run rerun --failed` first and it came back red, because a re-run replays the same commit; only a rebase picks up the repaired base.

What I did NOT do: nothing was squashed, reworded, dropped, or edited. Your commit is byte-identical after the rebase — `git diff` restricted to `tests/update-stop-first.test.ts` between the old and new head is empty, and the patch ID is unchanged at `efa23210f341`. The push used `--force-with-lease` pinned to your previous head, so it would have refused if you had pushed anything in the meantime.

Verified on the merged tree: `tsc --noEmit` exit 0, `tests/update-stop-first.test.ts` 15 pass / 0 fail.

The approval from my earlier review stands. Once CI is green at the new head this is ready to land.

lidge-jun:
One more note, because my rebase had a side effect I should own: force-pushing reset the four-box readiness checklist and moved this PR back to draft. That is `enforce-target` behaving correctly — a new commit invalidates an author's attestation about the old one — but it means my push created work for you.

Two of the four boxes are now objectively true and I can evidence them:

- **"I pushed my PR to the latest dev commit"** — the branch is now on `dev@50e955604`, 0 behind.
- **"I resolved all correct Codex and CodeRabbit findings"** — CodeRabbit is green at the new head, and the only outstanding review was my approval.

The other two are yours to attest, not mine: "All CI tests are green on my local testing" is an author attestation the gate deliberately never disproves, and "My PR is ready for review" is your call.

So: **tick the boxes and this lands.** Or reply here saying you are happy for me to proceed and I will mark it ready on your behalf and merge once the full matrix is green — I would rather ask than flip your attestation for you.

For reference, verified independently on the merged tree at the new head: `tsc --noEmit` exit 0, `tests/update-stop-first.test.ts` 15 pass / 0 fail.

olddonkey:
Re-checked the maintainer-rebased exact head `4a0cbb55c04190eb60cb02cbd16bdc751a607d14`.

- It is based directly on `dev@50e955604`.
- The PR patch is byte-identical to my independently rebased local commit.
- `bun test tests/update-stop-first.test.ts`: 15 pass / 0 fail.
- `bun run typecheck`: pass.
- `git diff --check`: pass.
- Full suite: 15,191 pass / 12 skip / 1 fail. The sole failure is `CLI key-login live-update overlay preservation`; it reproduces unchanged on a clean detached `upstream/dev@50e955604`, so it is not caused by this PR.

I restored the three objectively true readiness boxes: latest dev, review findings resolved, and ready for review. I intentionally left the local-all-green attestation unchecked because the exact current base is not locally green, even though this PR-specific validation is green. No review threads remain unresolved.

lidge-jun:
The full matrix is now green at `4a0cbb55c` — **20 success, 0 failures**, including `ci`, all four `test N/4` shards, `macos` and `gates`. That is the evidence your unticked box was waiting on, and it was never yours to produce: the red was the repository-wide `release version line` failure that #2766 repaired on `dev`.

Worth knowing why it took an extra step. Fork PRs here sit in `action_required` and run **no** product CI until a maintainer approves the workflow run — so "5 checks passing" on a fork PR means the matrix never started, not that it passed. I approved the run after the rebase; the 20 green checks above are that run.

Two boxes are now objectively true and evidenced:

- **"I pushed my PR to the latest dev commit"** — the branch is 0 behind `dev`.
- **"I resolved all correct Codex and CodeRabbit findings"** — CodeRabbit green, and my approval stands.

**"All CI tests are green on my local testing"** is the one left, and it is deliberately an author attestation the gate never disproves — I can tell you repository CI is green, but I cannot attest to your local run. So: tick it and this merges, or reply here that you are happy for me to proceed and I will mark it ready and merge on your behalf.

On the change itself, from the earlier review: treating `stop` exit 0 as a claim rather than proof is the right diagnosis, resolving the pid before `stop` rewrites `runtime-port.json` is the right ordering, and the `afterAll` liveness audit is what turns a silent leak into a red test instead of a hopeful cleanup.

PR #2751, opened 2026-08-27
test(core): lock the two invariants AGENTS.md claims are enforced

lidge-jun:
## Summary

The wp6 audit's useful negative result: the core invariants hold on `dev`, but two of them were held by nothing except the current code being correct. This adds the two guards that were promised and never written, and corrects one comment that has become false.

### Gap 1 — the no-await activation scan

`AGENTS.md` states the synchronous-activation guarantee as if it were enforced, and [`080_activation_is_synchronous.md:143`](devlog/_fin/260814_lab_core_decoupling/080_activation_is_synchronous.md) recorded it as a Phase-4 test. `tests/core-lab-boundary.test.ts` only ever held the two import-graph guards.

The failure it guards against is silent. Everything between `Bun.serve` and the return of `startServer` runs in one synchronous turn, which is what guarantees a policy route cannot be evaluated before its evidence provider is registered. Add one `await` there and the synchronous subagent-fallback chain observes an empty slot and routes subagents to a different model than the operator configured. Nothing goes red — the wrong model simply answers.

The scan blanks comments and string bodies before looking, which is load-bearing rather than tidy: the window itself contains two comments that say the word *await* (`src/server/index.ts:1853` "cannot await", `:1950` "nowhere to await"). A naive text match fails on correct code, and a guard that cries wolf on `dev` gets deleted rather than fixed. Template interpolations stay code, because `${await x()}` is a genuine violation. Brace-depth tracking distinguishes a function body from a block, so the three awaits inside the `server.stop` closure are tolerated while an await inside `if`/`try`/`for` is still reported.

### Gap 2 — the R3-1 profile-less dry-run assertion

[`090_audit_round3_closeout.md:47`](devlog/_fin/260814_lab_core_decoupling/090_audit_round3_closeout.md) required a behavioral assertion that a genuinely profile-less dry-run registers no slot. `tests/lab-activation.test.ts` only checked that a bare *config* never activates, which never exercises the management route where the activation call actually lives (`src/server/management/routing-profile-routes.ts:293` and `:368`).

This was exactly the "guard passes while the property is violated" path from audit round 3, so it is asserted in both directions: the same route, same process, one config field different must still fill the slot. Without that half, the negative case would keep passing if the route stopped activating altogether.

### Prose correction

`src/server/responses/core.ts:5194` still described generic OAuth failover as "opt-in and a strict no-op otherwise". Since #2568d presence is consent: rotation defaults ON at two or more eligible accounts (`src/oauth/generic-account-failover.ts:104-128`). Single-account installs are still a no-op and an explicit `oauthAccountFailover.enabled: false` still wins, so only the "opt-in" half was wrong. The behaviour is a recorded owner decision and is unchanged.

## Verification

Each new assertion was driven red before being kept:

| Injected regression | Result |
| --- | --- |
| `await Promise.resolve()` before the activation block + `startServer` marked `async` | 3 of 4 new boundary assertions fail |
| dry-run `activateLab` moved above the unknown-profile 404 and made unconditional | the R3-1 case fails, its anti-vacuity twin still passes |

Both injections reverted; `git status` confirmed clean afterwards.

At this head:

- `bun test tests/core-lab-boundary.test.ts tests/lab-activation.test.ts` — 32 pass, 0 fail
- full `bun run test` on the Linux CI host at this exact head

The scan's own behaviour is pinned by a guard-on-the-guard case covering comments, string literals, template interpolation, nested async callbacks, `for await`, and identifiers that merely contain the word.

## Checklist

- [x] Local CI green (focused tests, typecheck; full suite on the remote host at this head)
- [x] Branch is on the latest `dev` commit
- [x] All correct Codex and CodeRabbit findings fixed
- [x] Ready for review

## Summary by CodeRabbit

* **Documentation**
  * Clarified that generic OAuth account failover is enabled by default when multiple eligible accounts are available, with an opt-out setting. Certain providers remain excluded.
  * Documented safeguards that keep server activation synchronous and prevent unintended suspension.

* **Tests**
  * Added coverage for activation-window boundaries and synchronous server startup.
  * Added dry-run coverage for profile-less installations, including correct error handling and prevention of unintended activation records.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -62,6 +62,14 @@ subagent-fallback chain has nowhere to await, so an `await` added before the
 activation block would silently reroute subagents to a different model than the
 operator configured.
 
+That one is enforced too, in the same file: a scan reads the window between the
+`Bun.serve` call and the `labActivationRequired` check and fails on any `await`
+that would suspend `startServer` itself, plus on `startServer` being declared
+`async`. It has to ignore comments, string bodies, and nested functions to be
+usable, because the window legitimately contains three awaits inside the
+`server.stop` closure and two comments that mention the word. Until it existed,
+this paragraph was the only thing holding the guarantee.
+
 Design and audit history: `devlog/_fin/260814_lab_core_decoupling/`.
 
 ## The `devlog` directory
--- a/src/server/responses/core.ts
+++ b/src/server/responses/core.ts
@@ -5191,10 +5191,14 @@ async function handleResponsesInner(
           break;
         }
       }
-      // Generic OAuth account failover (#2568) for providers with no pool of their own. Opt-in
-      // and a strict no-op otherwise, so a single-account install and every existing config are
-      // unchanged. Codex and Anthropic are excluded by isGenericFailoverProvider — their pools
-      // own quota scopes, probe leases and affinity that this must not reimplement.
+      // Generic OAuth account failover (#2568) for providers with no pool of their own.
+      // Presence is consent since #2568d: rotation is ON by default once two or more eligible
+      // accounts are stored for the provider, because a second deliberate login is read as the
+      // operator asking for it. A single-account install is still a strict no-op, and an
+      // explicit `oauthAccountFailover.enabled: false` (global or per provider) still wins --
+      // see isGenericOAuthFailoverEnabled in src/oauth/generic-account-failover.ts. Codex and
+      // Anthropic are excluded by isGenericFailoverProvider: their pools own quota scopes,
+      // probe leases and affinity that this must not reimplement.
       while (
         upstreamResponse.status === 429
         && genericFailoverAccountId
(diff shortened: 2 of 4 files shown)


lidge-jun:
Independent review of this guard found it narrower than the invariant it claims to hold, and it was right. Fixed in `8be960f07`.

The window ran from `Bun.serve` to the `labActivationRequired` check, which left two blind spots: an `await` inside the `if (labActivationRequired(...))` body, and an `await` between activation and `return server`. Both `AGENTS.md` and `080_activation_is_synchronous.md` state the guarantee as covering everything from `Bun.serve` to the return, so the guard was holding less than it advertised.

Demonstrated rather than argued:

| Window | `await` injected inside the activation body | Result |
| --- | --- | --- |
| ends at the activation anchor (as landed in `af479e0b8`) | yes | 17 pass / 0 fail — invisible |
| ends at `return server;` (`8be960f07`) | yes | fails |
| ends at `return server;` | just before the return | fails |

The scan now also asserts the activation check sits *inside* the window, so it cannot stay green by drifting past the ordering it protects. Both injections reverted; tree confirmed clean.

One point from the review I want to record rather than act on: the unknown-profile dry-run case does 404 before reaching the activation call at `routing-profile-routes.ts:368`, so it is not by itself a lock on that line. That is why the block asserts both halves — the resolvable-profile case is the one that pins `:368`, and the unknown-profile case pins the weaker but still real property that no slot is filled on a profile-less install. I verified the asymmetry by making the dry-run activation unconditional and moving it above the 404: the unknown-profile case fails while its twin keeps passing.

PR #2775, opened 2026-08-27
fix(cli): stop reporting success for failed management calls (#2696 #2697 #2698)

lidge-jun:
## Summary

Stacked on #2773 (targets `codex/ocx-agentic-control-roadmap`, not `dev`).
Implements `devlog/_plan/260828_ocx_agentic_control/010_phase_transport_honesty.md`.

Three defects made the `ocx` CLI unusable for unattended operation, and they share a
theme: **a failed management call did not look like a failure.**

**#2697 — exit codes discarded.** The `provider` and `models` dispatch runners awaited
their handler then `return 0`, so `index.ts`'s single `process.exit(...)` erased the
code the handler had set. `ocx provider quota` against a stopped proxy printed an error
and exited 0 — silently green in any script or CI. Both runners now return
`Number(process.exitCode ?? 0)`, the pattern ten other runners already use, and reset
it first like the `service` runner does.

**#2698 — the server's diagnosis was thrown away.** `responseMessage` read only
`error`/`message`/`detail`, never `reason` or `hint` — the two fields the management
plane sets to say *why* it refused and *what to do*. Routes that return
`{ok:false, reason:"…"}` with no `error` key degraded to a generic
`Management request failed (409)`. The body was already attached to the thrown error,
so this was pure rendering loss. The account client was worse: it collapsed every
transport error into a bare status-0 sentinel inside `catch {}`, discarding the cause,
and exited 1 for everything including 404 and 409.

**#2696 — a management token could fence the whole management plane closed.** The
service exports its token file as `OPENCODEX_API_AUTH_TOKEN` before starting the proxy,
and nothing refused a management token in that slot. When the two match, the server
treats the admin credential as a data-plane admission secret and fails the entire
management plane closed at boot: every `/api/*` returns 503, even on a loopback install
that never needed a data-plane secret. The CLI could not recover it, because the fence
is decided server-side at startup.

`assertNotAdminToken` now refuses it at the write chokepoint and in
`assertServiceAuthEnvironment` — before the loopback short-circuit, since a loopback
install writes the token file too and returning early there is what let the broken
state through. For an install already in that state, `ocx doctor` names the cause and
the remedy without echoing the credential.

### The recurrence guard corrected its own plan

The plan specified an allowlist of `["login", "logout", "tray"]` for runners permitted
to return a literal 0. The guard contradicted it: the real offenders are `debug` and
`login`, and the pattern the plan proposed (`[^)]*`) matched **neither target runner**
[...]

PR #2778, opened 2026-08-27
feat: declare the management route surface and add ocx capabilities

### LINKED ISSUE #2701: [Bug]: PATH ocx package can be older than the running launchd proxy (help/features lie)
### Area

CLI / service lifecycle

The ocx binary on PATH should match the running launchd proxy, or doctor/status should scream when it does not.

### Current state

On macOS:
- PATH ocx -> ~/.bun/bin/ocx -> ~/node_modules/@bitkyc08/opencodex (2.30.0-preview.20260821)
- launchd com.opencodex.proxy runs nvm global 2.33.0

ocx usage --help from PATH still prints range 7d|30d|all and omits --provider/--model. 2.33.0 registry.ts already documents today|1d and those flags. Operators debug the running proxy with an older CLI.

ocx status reports proxy version via healthz (2.33.0) but does not warn that the invoking CLI package is 2.30.0-preview.

### Proposed

- ocx doctor / ocx status: compare process.argv CLI package version to GET /healthz version; warn on mismatch.
- Document that service install does not replace ~/.bun/bin/ocx.

### Version

CLI 2.30.0-preview.20260821 vs proxy 2.33.0

lidge-jun:
## Summary

Third PR in the ocx agentic-control stack. Targets **`codex/ocx-transport-honesty`** (PR #2775), not `dev` — retarget after the parent lands.

Closes #2701.

This is the keystone phase: it introduces the declared route registry and capability table that later phases register verbs into. Building verbs first would mean writing them twice.

Three things land:

1. **`src/server/management/route-registry.ts`** — 210 declared routes (192 resolved from source, 18 non-literal), with a mandatory-reason exemption vocabulary.
2. **`ocx capabilities`** — the machine-readable surface index, so an agent driving `ocx` does not have to parse help text.
3. **Version skew** (#2701) — `ocx status` and `ocx doctor` now report when the `ocx` on PATH is a different build from the running proxy.

### The registry is declared, not harvested

18 routes are unreachable by any text scan of their own file: four regexes over model paths, three over lab ids, an `endsWith`, a `pathname.slice`, a prefix decode, three path constants, and two negated `pathname !==` guards.

The negated pair is why this matters. `GET /api/routing-analytics` has no equality literal anywhere, so a scan built on `===` never sees it. And for `GET /api/storage` an equality scan finds **only the dead shadowed copy** in `logs-usage-routes.ts` — never the live guard in `storage-log-guard-routes.ts`. A gate built that way would have reconciled against the corpse and called the surface covered.

### The reconciliation counts routes, not grep hits

A count keyed on `rg` line hits cannot balance against this source:

- `/api/codex-auth/pool-strategy` registers **PUT and PATCH from one guard** spanning four lines.
- 19 path guards decide their method somewhere other than their own line.
- Both nesting orders occur and neither is inferable from the other: `lab-routes.ts:352` narrows to GET *before* eight path literals, while `storage-log-guard-routes.ts:112` opens with the path and puts the method guard *inside* at :113.

So the scanner walks statements in order maintaining a method-narrowing context, and when it cannot resolve a method it returns `null` and the test fails. It never assumes GET.

### The gate was driven red before being trusted

Injecting `POST /api/combos/red-first-probe` into a real handler failed two checks independently and named the route in both. A gate nobody has attacked is not evidence.

Three further defects were found **by the new tests** during implementation, not by inspection:

- the generator double-declared the negated-guard routes once the scanner learned to resolve them;
- the inertness assertion flagged the registry's own header prose;
- the capability table claimed `ocx status` drives `GET /api/status` — a route that does not exist. `collectStatus` probes `/healthz` and reads local config.

### Lab boundary

`route-registry.ts` is imported by `src/server/management-api.ts`, which `tests/core-lab-boundary.test.ts` protects. Route paths are strings, so declaring `/api/lab/status` creates no module edge. The module imports **nothing at all**, and a test asserts it.

The seven mutating `/api/lab` routes were going to be waved through as `local-transport` alongside the reads, but `ocx lab` reads local SQLite and cannot start an automation run or import a community bundle. They now carry `deferred-verb` with an owner phase and a **tracked** devlog doc the test verifies exists — deliberately a repository file and not the goalplan, which is gitignored, so the assertion means the same thing in CI as locally.

### Version skew (#2701)

No extra request: the identity probe already parsed the healthz body. Carrying the version took three hops, because `isOpencodexHealthz` is a boolean predicate and cannot return a value — `proxyIdentityAt`'s return type widened and all three `findLiveProxy` construction sites thread the field.

That broke 12 exhaustive `toEqual` assertions in `proxy-liveness.test.ts` (Bun's `toEqual` rejects an extra defined key). Each now asserts the version its own mock actually reports, rather than being loosened.

The warning gets its own `versionSkew` field rather than joining `warningParts`, which funnels into `codexRuntime.warning` and would print a stale-PATH fact under the Codex runtime heading. `schemaVersion` stays 1. Both placeholder versions (`0.0.0` server-side, `unknown` CLI-side) suppress the warning — a false stale-CLI warning sends an operator to reinstall a healthy install.

## Verification

- `bun test` over the 12 affected suites: **363 pass, 0 fail, 1216 expect() calls**.
- `tsc --noEmit`: clean.
- Reconciliation gate proven red-then-green by injecting a real route into `combo-routes.ts`.
- Fail-loud scanner path proven red by a synthetic unresolvable guard.
- Skew warning verified live, not only under test: this worktree's CLI at 2.35.0 against a running 2.32.1-preview proxy prints the warning.
- `tests/core-lab-boundary.test.ts` and `tests/repo-hygiene.test.ts` green.

Local full-suite and CI checks are deliberately deferred to the final stack-wide rebase and CI pass (wp9), per the stack plan.

## Checklist

- [x] `tsc --noEmit` clean
- [x] No request bodies, keys, or account identifiers logged

## Summary by CodeRabbit

* **New Features**
  * Added `ocx capabilities` with human-readable or JSON output, mutation filtering, and route lookup.
  * Added capability listings for status, providers, accounts, usage, and related commands.
  * `ocx status` and `ocx doctor` now detect and report CLI/proxy version mismatches.
  * Proxy discovery now preserves the running proxy’s reported version.

* **Documentation**
  * Updated CLI help to document the new capabilities command.

### DIFF
--- a/src/cli/capabilities-command.ts
+++ b/src/cli/capabilities-command.ts
@@ -1,0 +1,93 @@
+/**
+ * `ocx capabilities` -- the surface index an agent reads first.
+ *
+ * The point of this verb is that driving `ocx` programmatically should not require
+ * parsing help text. `--json` emits the capability table with the management routes each
+ * capability drives; `--route` answers the inverse question.
+ */
+import {
+  CAPABILITIES,
+  HEAD_CAPABILITIES,
+  capabilitiesForRoute,
+  capabilityInvocation,
+  type Capability,
+} from "./capabilities";
+import { takeFlag } from "./runtime-api";
+
+function takeValueFlag(args: string[], flag: string): string | undefined {
+  // Order-independent by construction: scan for the flag anywhere in argv rather than
+  // reading a fixed position. Positional flag handling is exactly why
+  // `ocx restore back --json` ignored its flag.
+  const idx = args.indexOf(flag);
+  if (idx === -1) return undefined;
+  const value = args[idx + 1];
+  args.splice(idx, value === undefined ? 1 : 2);
+  return value;
+}
+
+function renderHuman(caps: readonly Capability[], includeHead: boolean): void {
+  for (const cap of caps) {
+    const marker = cap.mutates ? "!" : " ";
+    console.log(`${marker} ${capabilityInvocation(cap)}`);
+    console.log(`    ${cap.summary}`);
+    if (cap.routes.length > 0) {
+      console.log(`    routes: ${cap.routes.map(r => `${r.method} ${r.path}`).join(", ")}`);
+    }
+    if (cap.flags.length > 0) {
+      console.log(`    flags:  ${cap.flags.map(f => f.name).join(" ")}`);
+    }
+  }
+  if (!includeHead) return;
+  for (const head of HEAD_CAPABILITIES) {
+    console.log(`  ocx ${head.invocations[0]}`);
+    console.log(`    ${head.summary}`);
+  }
+}
+
+export async function runCapabilities(argv: string[]): Promise<number> {
+  const args = [...argv];
+  const json = takeFlag(args, "--json");
+  const mutatingOnly = takeFlag(args, "--mutating-only");
+  const route = takeValueFlag(args, "--route");
+
+  if (route !== undefined && route.length === 0) {
+    console.error("Usage: ocx capabilities --route <path>");
+    return 64;
+  }
+
+  let selected: readonly Capability[] = route === undefined ? CAPABILITIES : capabilitiesForRoute(route);
+  if (mutatingOnly) selected = selected.filter(cap => cap.mutates);
+
+  if (route !== undefined && selected.length === 0) {
+    // An unmatched route is a real answer, not an error: the route may be exempt or may
+    // not exist. Say which, rather than exiting 0 with silence.
+    if (json) {
+      console.log(JSON.stringify({ schemaVersion: 1, route, capabilities: [] }, null, 2));
+    } else {
+      console.error(`No CLI capability drives ${route}.`);
+    }
+    return 4;
+  }
+
+  if (json) {
+    console.log(JSON.stringify({
+      schemaVersion: 1,
+      ...(route === undefined ? {} : { route }),
+      capabilities: selected.map(cap => ({
+        command: cap.command,
+        invocation: capabilityInvocation(cap),
+        summary: cap.summary,
+        routes: cap.routes,
+        flags: cap.flags,
+        mutates: cap.mutates,
+        json: cap.json,
+        ...(cap.details ? { details: cap.details } : {}),
+      })),
+      ...(route === undefined && !mutatingOnly ? { headCapabilities: HEAD_CAPABILITIES } : {}),
+    }, null, 2));
+    return 0;
+  }
+
+  renderHuman(selected, route === undefined && !mutatingOnly);
+  return 0;
+}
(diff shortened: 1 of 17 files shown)


lidge-jun:
## 리뷰 · 우선순위 66 / 80

이 PR은 ocx agentic-control 스택의 열쇠입니다. 베이스는 `dev`가 아니라 `codex/ocx-transport-honesty`(PR #2775)입니다. 부모를 먼저 머지한 뒤 리타깃해야 합니다. 지금 `dev`에는 `src/server/management/route-registry.ts`도, `src/cli/capabilities.ts`도 없습니다.

하는 일은 세 가지입니다. (1) 관리 API 라우트를 선언 목록으로 고정합니다. 평등 비교(`===`)만으로는 안 보이는 라우트가 18개라서, 수확(harvest)이 아니라 선언이 맞습니다. `GET /api/routing-analytics`처럼 부정 가드만 있는 경로가 대표입니다. (2) `ocx capabilities`로 기계가 읽을 수 있는 표면 목록을 줍니다. 에이전트가 help 텍스트를 파싱하지 않게 하려는 목적입니다. (3) #2701 — PATH의 `ocx`와 실행 중 프록시 빌드가 다르면 `status`/`doctor`가 말해 줍니다.

면제(exemption) 어휘에 `deferred-verb`가 있고, owner·ownerDoc·TRACKED 문서 존재를 테스트가 검사합니다. Lab 경계도 의식합니다. 레지스트리 모듈은 순수 데이터라서 `/api/lab/...` 경로 문자열만 적어도 `src/lab/` import는 생기지 않습니다. types/config 분할과는 무관합니다. 다만 이 PR만으로는 `dev`에 못 올라갑니다.

라인 src/server/management/route-registry.ts - MANAGEMENT_ROUTES 선언. mechanism·exempt·mutates 필드가 있습니다. handler를 import하면 안 됩니다.

라인 tests/management-route-registry.test.ts - 소스에서 (method, path)를 모아 선언과 맞춥니다. 빠지면 실패합니다.

라인 src/cli/capabilities.ts / capabilities-command.ts - 이후 동사들이 등록될 테이블의 시작점입니다.

라인 src/cli/version-skew.ts - PATH ocx와 프록시 빌드 불일치를 보고합니다. #2701 핵심입니다.

경로 base branch - `codex/ocx-transport-honesty` 위입니다. #2775가 먼저입니다.

메인테이너의 판단이 필요한 지점
- #2775 머지 후 이 PR을 `dev`로 리타깃할지, 스택 전체를 한 번에 올릴지
- deferred-verb 면제 문서가 CI에서 실제로 파일이 있는지
- 210개 라우트 전량이 맞는지(특히 non-literal 18개)

너의 추천
부모 #2775 머지·리타깃 후에 이 PR을 올리세요. 키스톤이라 동사 PR(#2779 이후)보다 먼저 가는 순서는 맞습니다. Lab-clean·선언-not-harvest 원칙은 유지하세요.

이 댓글은 grok-bot이 작성했습니다

PR #2779, opened 2026-08-27
fix(cli): make exit codes and --json usable from a script

lidge-jun:
## Summary

Fourth PR in the ocx agentic-control stack. Targets **`codex/ocx-capability-registry`** (PR #2778), not `dev` — retarget after the parent lands.

Three defects that made `ocx` hard to drive from a script. Each was verified in source before being changed.

### `doctor` always exited 0

It reports by direct `console.log` across roughly a dozen sections with no checks collection to inspect, so a `FAIL`-level condition printed and then the runner hard-returned 0. A diagnostic that cannot fail cannot gate anything, which defeats running it from a script at all.

It now returns 1 when any `FAIL`-level check fired, while still honouring the explicit `process.exitCode` its two special flag paths already set.

**Only `FAIL` fails the command.** `WARN` describes a degraded-but-working install, and this doctor emits `[WARN]` freely — a Codex app-server started before the catalog changed, a WHAM probe that could not reach chatgpt.com. Failing on those would break pipelines that are legitimately green, and the predictable response is that people stop running the gate.

The failure flag resets at the top of each `runDoctor` pass. That is not incidental: the suite drives it several times in one process, and a sticky flag would fail the second call because the first saw a problem — a test whose result depends on execution order, which is worse than no gate.

> ⚠️ **Breaking for pipelines** that ran `ocx doctor` and ignored the result. The alternative is a diagnostic that always claims success.

### `--json` was position-dependent in two commands

- `status` honoured it **only as the lone argument**, so `ocx status --json --anything` printed human output to a caller that explicitly asked for JSON.
- `restore` matched `args[1]`, so `ocx restore back --json` **ignored the flag entirely** — position 1 held `back`.

`status` now takes the flag out of argv and rejects whatever remains, which keeps the strict unknown-argument behaviour rather than trading one defect for another.

### `sync-cache` could not report a skipped write

`completed` with a falsy value means the cache was **not** rewritten; any other kind means the write never completed. Both exited 0, so a script could not tell a refreshed cache from a skipped one. It now exits 1 on an incomplete write, treats a deliberate skip (Codex integration off) as success — failing there would break a correctly configured machine that simply is not using Codex — and gained `--json` reporting `wrote` and `skipped` separately.

### The test generalises past the two known cases

Pinning only `status` and `restore` would leave the next command free to repeat the mistake, so the contract test fails on **any** `args[<n>] === "--json"` in `dispatch.ts`, `index.ts`, or `root.ts`. Both defective forms are also pinned as exact strings, so a revert is caught by a failing test rather than discouraged by a comment.

### Deferred deliberately

`025.3` also asks for `--json` on `login`, `logout`, `sync`, and `debug`. Those are interactive or streaming flows whose output is not a single structured result, and inventing an envelope without a consumer would be guesswork. They move to wp7, where an actual caller drives the shape. Recorded in `devlog/.../026` rather than dropped silently.

## Verification

- 9 focused suites: **125 pass, 0 fail, 618 expect() calls**.
- `tsc --noEmit`: clean.
- **Proven non-vacuous:** reverting the `restore` fix alone turns 3 of the 8 contract tests red.
- `ocx status --json` verified live; `ocx status --json --bogus` still rejects with usage.

Local full-suite and CI checks are deferred to the stack-wide final rebase and CI pass (wp9), per the stack plan.

## Checklist

- [x] `tsc --noEmit` clean
- [x] Breaking change called out above and in the devlog record
- [x] No request bodies, keys, or account identifiers logged

## Summary by CodeRabbit

- **New Features**
  - Added structured JSON output for `restore back` and `sync-cache`.
  - `--json` is now accepted in any argument position for supported commands.
  - JSON responses distinguish successful writes from skipped operations.
- **Bug Fixes**
  - Improved command exit statuses for failed doctor checks and genuine synchronization failures.
  - Prevented unnecessary process restarts when synchronization makes no changes.
  - Unknown command-line arguments are now rejected consistently.
- **Documentation**
  - Added an implementation record describing the updated CLI behavior and verification results.

### DIFF
--- a/devlog/_plan/260828_ocx_agentic_control/026_wp3b_implementation_record.md
+++ b/devlog/_plan/260828_ocx_agentic_control/026_wp3b_implementation_record.md
@@ -1,0 +1,66 @@
+# 026 — wp3b implementation record
+
+Branch `codex/ocx-uniform-contract` off `codex/ocx-capability-registry`. Implements `025`.
+
+## What landed
+
+| File | Change |
+|---|---|
+| `src/cli/doctor.ts` | per-pass failure flag + `doctorFailed()`; a `FAIL`-level OAuth check records a failure |
+| `src/cli/dispatch.ts` | `doctor` returns 1 on failure; `restore` scans argv for `--json`; `sync-cache` returns 1 when the cache write did not complete and gained a `--json` envelope |
+| `src/cli/index.ts` | `status` uses `takeFlag`, so `--json` works in any argv position |
+| `tests/cli-json-contract.test.ts` | 8 tests, new file |
+
+## The plan under-specified how `doctor` aggregates failures
+
+`025` says "return non-zero when any check fails." That reads as though there were a
+checks collection to inspect. There is not: the `checks.push` calls belong to
+`collectOAuthDoctorChecks`, a different function, and `runDoctor` reports by direct
+`console.log` across roughly a dozen sections with `ok `/`!! `/`[WARN]` prefixes.
+
+So the failure signal is a module-scoped flag reset at the top of each `runDoctor` pass.
+The reset is not incidental: the suite drives `runDoctor` several times in one process,
+and a sticky flag would fail the second call because the first saw a problem — a test
+that passes or fails depending on execution order, which is worse than no gate.
+
+## Only `FAIL` fails the command
+
+`025` does not distinguish the levels. It matters. `WARN` describes a
+degraded-but-working install, and this doctor emits `[WARN]` freely — a Codex app-server
+started before the catalog changed, a WHAM probe that could not reach chatgpt.com. Failing
+on those would break pipelines that are legitimately green, and the predictable response
+is that people stop running the gate. `FAIL` is documented in `doctor.ts:63` as the level
+for a surface that is unusable rather than degraded, which is exactly the line worth
+exiting non-zero on.
+
+## `sync-cache` needed a success definition, not just an exit code
+
+`invalidated.kind === "completed"` with a falsy value means the cache was **not**
+rewritten, and any other kind means the write never completed. Both previously exited 0.
+But a deliberate skip — Codex integration off — is not a failure, and treating it as one
+would make `ocx sync-cache` fail on a correctly configured machine that simply is not
+using Codex. Success is therefore `wrote || desiredDisabled`, and the `--json` envelope
+reports `wrote` and `skipped` separately so a caller can tell which happened.
+
+## The contract test generalises past the two known defects
+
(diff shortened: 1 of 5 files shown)


lidge-jun:
## 리뷰 · 우선순위 64 / 80

이 PR은 ocx agentic-control 스택의 네 번째입니다. 베이스는 `codex/ocx-capability-registry`(PR #2778)이지 `dev`가 아닙니다. 부모를 머지한 뒤 리타깃해야 합니다. 목표는 하나뿐입니다. 스크립트가 `ocx`를 신뢰할 수 있게 만드는 것입니다.

가장 큰 구멍은 `doctor`였습니다. 대략 열 개 넘는 섹션이 `console.log`로 FAIL을 찍은 뒤, 러너가 하드코드로 0을 반환했습니다. 실패할 수 없는 진단은 CI 게이트로 쓸 수 없습니다. 이 PR은 FAIL이 하나라도 있으면 1을 돌립니다. WARN은 실패가 아닙니다. Codex 앱 서버가 카탈로그 갱신 전에 떠 있거나 WHAM 프로브가 chatgpt.com에 못 닿는 같은 정상 저하까지 빨갛게 만들면, 사람들은 게이트 자체를 끄게 됩니다. `runDoctor` 시작마다 실패 플래그를 리셋하는 것도 중요합니다. 테스트가 같은 프로세스로 여러 번 돌리기 때문입니다.

둘째는 `--json` 계약입니다. 명령마다 성공/실패 JSON 모양이 달랐고, 스크립트가 분기하기 어려웠습니다. 셋째는 dispatch/index에서 종료 코드가 삼켜지지 않게 정리한 것입니다. 서버·config 스키마는 안 건드립니다. types.ts/config.ts 분할과도 무관합니다.

라인 src/cli/doctor.ts - FAIL만 exit 1. WARN은 통과. 매 패스 시작 때 실패 플래그 리셋.

라인 src/cli/dispatch.ts / index.ts - 하위 명령 exit code가 프로세스까지 전달됩니다.

라인 tests/cli-json-contract.test.ts - --json과 종료 코드를 함께 고정합니다.

경로 base branch - #2778 위. #2775→#2778 없이 `dev` 단독 머지 금지.

메인테이너의 판단이 필요한 지점
- WARN을 영원히 비실패로 둘지, 나중에 `--strict`를 둘지
- #2778 머지 직후 바로 이어서 올릴지

너의 추천
부모 스택이 `dev`에 붙은 뒤 머지하세요. FAIL-only 정책은 그대로 두는 편이 운영에 안전합니다.

이 댓글은 grok-bot이 작성했습니다

PR #2780, opened 2026-08-27
fix(cli): stop discarding fields the API already returns

### LINKED ISSUE #2700: [Feature]: ocx usage human report omits accounts[] that GET /api/usage already returns
### Area

CLI usage report

See per-account request/token/cost totals from the terminal without opening the dashboard.

### Current state

GET /api/usage already returns accounts[] (src/server/management/logs-usage-routes.ts, src/usage/summary.ts UsageAccount).

ocx usage / ocx observe usage human renderer (src/cli/usage-report.ts formatUsageReport) prints requests, tokens, est. cost, providers, models. It never prints accounts[]. --json dumps the API payload so accounts are visible only in JSON.

### Why not existing issues

- #1820 is dashboard Usage cost/cache metrics, not CLI accounts table.
- #1062 is Antigravity dashboard mixing.
- #2566 is quota bars, not usage.jsonl rollup.
- #2696 is 503 on the call; even when the API works, the human table still omits accounts.

### Proposed

Print an ACCOUNTS table in formatUsageReport when accounts[] is non-empty. Optional --account filter later.

### Version

2.33.0

lidge-jun:
## Summary

Fifth PR in the ocx agentic-control stack. Targets **`codex/ocx-uniform-contract`** (PR #2779) — retarget after the parents land.

Closes #2700, #2703, #2705. No server change, by design.

Three fields arrived on the wire, were declared in a type, and never reached the operator. That is the worst shape for this class of bug: the code typechecks, the renderer reads correctly, and only the output is wrong.

| Field | Where it was lost |
|---|---|
| `fiveHourPercent` / `fiveHourResetAt` | `projectQuota`'s whitelist, while the DTO declared both and two renderers read them |
| `paused` | never mapped at all, though the server sends it at `auth-api.ts:286` and `:1315` |
| `requests7d` / `totalRequests` / `lastUsedAt` | `access key list` formatted keys as exactly id + name + prefix |

### The test the plan proposed could not detect the bug it targeted

Worth reading even if you skim the rest. The plan asks for an assertion that `formatAccountTable` shows a 5h-only quota. **That passes with the defect fully present** — `formatAccountTable` takes a row directly, and the field was dropped one layer earlier in `projectQuota`.

Confirmed rather than reasoned: after the renderer test went green, I reverted the `projectQuota` fix and the renderer test **stayed green**. The coverage then moved to drive `fetchCodexRows` with a server payload, which does go red on the same revert. Both tests now exist, for different reasons.

A renderer test for a projection bug is the same category of mistake as the bug: the layer that looks responsible is not the one that is.

### `paused` prints additively, not exclusively

The plan says `paused` outranks `selected`. Implemented as **both**. A paused-but-selected account is the state an operator most needs named — requests route to it while the pool believes it is held out of rotation — and printing only the winner of a precedence rule hides exactly that case.

### Two quota dialects in one file

`refreshLine` gated its whole quota block on weekly/monthly, so a 5h-only account printed `quota: unknown` while `quotaParts` **five lines below in the same file** rendered the same DTO correctly. Teaching the second dialect a third window would leave the disagreement in place, so it delegates.

### The ambiguous union is a contract, not a hint

`ApiKeyUsage` is a discriminated union whose `{ambiguous:true}` variant carries **no** numbers, and the type comment says why: an optional marker beside `requests7d: 7` invites a consumer to render the 7. So the CLI prints `ambiguous` across the numeric columns and never a `0`. A test asserts the output contains no standalone `0` in that case — reporting zero requests for a key that may be in heavy use is the dangerous answer for someone deciding what to revoke.

`attributionSince` and `historyTruncated` print once as a footer, since they describe the data set. That is load-bearing: without `attributionSince`, an absent `lastUsedAt` cannot be read at all, because "never used" and "nothing is attributable yet" look identical.

### The withheld-rows case (#2700)

The server blanks `accounts` under any provider or model filter, because account rows are not provider-partitioned in a way the projection could honestly re-derive. "What did this provider cost me per account" is the most natural way to ask, so an empty table would answer *"no accounts used this provider"* — a different and wrong answer, and the same silently-wrong-output class this unit removes. The report now states the withholding, and the `usage` capability's `details[]` carries the same sentence so `ocx capabilities --json` reflects it.

ACCOUNT rows stay empty for xai and cursor until wp6 stamps their labels (#2699). The renderer lands first deliberately, so that phase's proof is visible immediately.

## Verification

- 7 focused suites: **202 pass, 0 fail, 775 expect() calls**.
- `tsc --noEmit`: clean.
- **Non-vacuous:** reverting `projectQuota` turns the projection test red while the renderer test does not move — which is why both exist.

Local full-suite and CI are deferred to the stack-wide final rebase and CI pass (wp9).

## Checklist

- [x] `tsc --noEmit` clean
- [x] No server-side change in this diff
- [x] No request bodies, keys, or account identifiers logged (key prefixes only, as the API already returns them)

### DIFF
--- a/devlog/_plan/260827_igwanu_bug_pr_merge_round/000_plan.md
+++ b/devlog/_plan/260827_igwanu_bug_pr_merge_round/000_plan.md
@@ -1,0 +1,116 @@
+# 000 — igwanu bug-PR merge round: plan
+
+Round base: `dev @ 8b1b65b8d` (local == `origin/dev`, verified 2026-08-27).
+Scope: the 13 open **bug**-labelled PRs. Four are Ingwannu's (#2767, #2766, #2764,
+#2761); nine are other authors' (#2747, #2745, #2740, #2733, #2729, #2726, #2693,
+#2638, #2497).
+
+Enhancement-labelled PRs are explicitly out of scope for this round.
+
+## The finding that orders the whole round
+
+Three PRs (#2767, #2764, #2747) show **failing required CI** — `ci`, `macos`,
+`test 3/4`, `gates` — while their own merged trees compile clean. The failure is
+not theirs. Every one of them fails the same repository-wide assertion:
+
+```
+error: package.json version 2.34.0 equals release tag v2.34.0, but this commit is
+not the one that tag names. The tree claims an already-published version.
+(fail) release version line > the in-tree version is never behind a released one
+```
+
+`dev` still carries `2.34.0` after tag `v2.34.0` shipped, so *every* PR opened
+after the release train inherits a red matrix. A second shared failure hits
+`gates`: `privacy:scan` reads the scp-style SSH remote principal recorded in
+`devlog/_plan/260827_release_train/020_preview_release.md` as an email address.
+
+**#2766 repairs both.** It is the keystone: until it lands, no other PR in this
+round can produce a trustworthy green matrix, and re-running their CI is wasted
+work. This is the inverse of the previous round's lesson — there, green checks
+were not evidence of health; here, red checks are not evidence of harm.
+
+Evidence: run `33081644562` job `98550259965` (#2767), run `33080634739` job
+`98546624127` (#2764), run `33059606933` job `98534630924` (#2747) — each shows
+`1 fail` and that one failure is `release version line`.
+
+## Merged-tree gate (this round's own evidence, not GitHub's)
+
+Every PR head was fetched, merged against `dev @ 8b1b65b8d` with
+`git merge-tree --write-tree`, committed as `mtp/<n>`, checked out to an isolated
+worktree sharing this repo's `node_modules`, and compiled.
+
+| PR | ahead | behind dev | merge-tree | tsc on MERGED tree |
+|---|---|---|---|---|
+| #2767 | 1 | 0 | CLEAN | OK |
+| #2766 | 2 | 0 | CLEAN | OK |
+| #2764 | 1 | 0 | CLEAN | OK |
+| #2761 | 1 | 2 | CLEAN | OK |
+| #2747 | 1 | 26 | CLEAN | OK |
+| #2745 | 2 | 26 | CLEAN | OK |
+| #2740 | 1 | 26 | CLEAN | OK |
+| #2733 | 1 | 43 | CLEAN | OK |
+| #2729 | 2 | 89 | CLEAN | OK |
+| #2726 | 1 | 63 | CLEAN | OK |
+| #2693 | 2 | 118 | CLEAN | OK |
+| #2638 | 2 | 179 | CLEAN | OK |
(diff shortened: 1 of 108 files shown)


lidge-jun:
## 리뷰 · 우선순위 68 / 80

이 PR은 ocx 스택 다섯 번째입니다. 베이스는 `codex/ocx-uniform-contract`(PR #2779)입니다. #2700·#2703·#2705를 닫습니다. 서버는 이미 보내던 필드를 CLI가 버리고 있었습니다. 타입은 맞고 렌더러도 맞아 보이는데 출력만 틀린, 가장 나쁜 형태의 버그입니다.

세 군데입니다. (1) `projectQuota` 화이트리스트에 `fiveHourPercent`/`fiveHourResetAt`이 없어서 5시간 창만 있는 계정이 unknown처럼 보였습니다. 지금 `dev`의 auth-api는 해당 필드를 보냅니다. (2) `paused`는 `auth-api.ts` 286·1315에서 오는데 CLI가 매핑을 안 해서, 일시정지 계정과 사용 가능 계정이 목록에서 같았습니다. (3) access key list가 id·name·prefix만 찍고 `requests7d`/`lastUsedAt` 등을 버렸습니다. usage report의 `accounts[]`도 같이 살립니다.

본문에 적힌 대로, 계획서의 `formatAccountTable` 테스트는 결함이 있어도 통과합니다. 필드가 그 앞 레이어에서 이미 잘리기 때문입니다. 이번 PR은 projection·매핑 레이어를 고치고, 그 레이어를 직접 잠그는 테스트를 둡니다. types/config 분할과 무관합니다.

라인 src/cli/account-api.ts projectQuota - fiveHour*를 화이트리스트에 넣습니다. DTO에 있는데 빼던 상태가 버그였습니다.

라인 src/cli/account-api.ts / account.ts - paused를 매핑하고, paused+selected를 같이 표시합니다. 둘 중 하나만 보면 헷갈리는 상태가 가려집니다.

라인 src/cli/access.ts - 키 사용량 필드를 출력합니다.

라인 src/cli/usage-report.ts - accounts[]를 그리고, provider/model 필터 때는 빈 표 대신 이유를 말합니다.

라인 tests/cli-dto-fidelity.test.ts - projection 레이어를 직접 검증합니다. 렌더러만 보면 안 됩니다.

경로 base - #2779 위. 서버 변경 없음은 의도입니다.

메인테이너의 판단이 필요한 지점
- #2700/#2703/#2705를 이 PR Closes로 묶어도 되는지(서버 불변 전제)
- 스택 중간이라 CI 빨강이 부모 미머지 때문인지 확인

너의 추천
부모 #2779까지 올린 뒤 머지하세요. DTO fidelity 테스트 패턴은 이후 CLI 버그에도 재사용하세요.

이 댓글은 grok-bot이 작성했습니다

PR #2783, opened 2026-08-27
feat(quota): detect usage-window resets and notify on them

lidge-jun:
## Summary

OpenCodex can now notice when a usage window resets and tell you about it — both the scheduled
rollover you could predict and an out-of-band reset you could not. Off by default: with no
`quotaResetNotify` section, no detection runs, no timer starts, and no state file is written.

```json
{
  "quotaResetNotify": {
    "enabled": true,
    "webhookUrl": "https://hooks.slack.com/services/...",
    "pollSeconds": 900
  }
}
```

Detection compares two consecutive snapshots of one window at both existing quota seams
(`src/codex/quota.ts`, `src/providers/quota.ts`), plus an opt-in idle poller — that poller is
load-bearing rather than a convenience, because `fetchProviderQuotaReports` has exactly one
caller, so with no dashboard open an overnight reset would never be observed, and the overnight
case is the whole point.

Delivery goes to a webhook, a local command, or both. Read recent detections with
`ocx provider resets` or `GET /api/quota-resets`.

**Payload:** closed-union labels and numbers only.

```json
{"type":"quota_reset","kind":"scheduled","scope":"codex","accountTag":"9rhlw1hu",
 "window":"weekly","percentBefore":96,"percentAfter":2,"detectedAt":1787857037499}
```

`accountTag` is a per-install salted hash — it distinguishes your accounts from each other
without telling the receiver who they are. The payload is built field by field rather than by
spreading the event, so the internal idempotence key (which encodes scope and tag) cannot leak
and a future detector field cannot silently start crossing a third-party boundary.
`detectedAt` is when the proxy noticed, not when the reset happened: observation is bounded by
the 5-minute provider cache and the 10-minute per-account cache, so the instant can only be
bracketed.

### Security-relevant choices

- **`webhookUrl` is treated as a credential.** For Slack and Discord, holding the URL is
  sufficient to post. It matched none of the existing `SECRET_KEYS` patterns, so
  `ocx config show` printed it and `config export` wrote it to disk in the clear. Now redacted;
  the test fails without that pattern entry.
- **The webhook refuses a private-network destination** unless `allowPrivateNetwork: true`. An
  operator-supplied URL is an SSRF surface — this process reaches loopback services and cloud
  metadata endpoints a browser cannot — so `assertUrlResolvesPublic` gates it by default.
- **The command sink is argv, never a shell string,** so an operator value cannot become an
  injection surface. Failures are reported as a closed union of reasons, never an upstream body
  or the URL.
- **No new auth surface.** `GET /api/quota-resets` inherits `requireManagementAuth` and
  mutates nothing.
- **No retry.** The observer claims the idempotence key before dispatch, so a retry could only
  duplicate a delivery the durable ledger already considers done.

### Boundary

Adds `tests/quota-reset-core-boundary.test.ts`. `tests/core-lab-boundary.test.ts` hardcodes
`/src/lab/`, so nothing kept `src/quota/` off the core path — and both seams are statically
reachable from `src/server/responses/core.ts`, so one static import would put the detector and
the `../config` barrel on the request path of a user who never enabled the feature. The walker
is extracted to `tests/helpers/import-graph.ts` and shared with the Lab guard rather than
copied, because that guard already records what a duplicated predicate costs.

### Defects found and fixed during the work

Several were in code that was already green, found by adversarial review and by attacking my
own guards. The ones worth calling out:

1. **Out-of-order observations manufactured resets.** Two awaits before the baseline swap, and
   Bun does not resolve concurrent `import()` calls in call order. A burst of 21 monotonically
   *rising* writes (10% → 90%, no reset) fired a false `surprise` on every run — and the false
   event burned the durable claim key, permanently suppressing the genuine reset on that window.
2. **The provider seam was dead on arrival.** `previous` binds only when `cache.key === key`,
   but the key digest includes the quota values, so it is empty precisely when a reset happened.
3. **Reauth of a used account fired a false reset**, and the regression test written to prevent
   it also cleared the observer store — which no production path does. It simulated a state that
   never occurs.
4. **The account key collapsed an API-key pool** onto one identity, and could be read after a
   mid-flight failover rewrote it.
5. **The enable gate could never turn on**: `captureConfigGeneration` only advances on
   account/provider reconciliation, so editing the notify section never bumped it.
6. **A rolling window's natural decay read as a surprise reset** (88% → 61% one hour into a 5h
   window) on the most common window in the system.
7. **The observed-window map evicted its hottest row**, because re-setting a key does not move
   it in a Map — so eviction removed the earliest-inserted rather than the least-recently-used.
8. **The debounced baseline write was starvable**: 75 observations at 40ms produced zero disk
   writes, so a SIGKILL lost the whole baseline.

Design record and full audit history: `devlog/_plan/260828_quota_reset_detection/`.

## Verification

```
$ bun x tsc --noEmit
(exit 0, no output)

$ bun run test
Ran 15307 tests across 958 files.
(exit 0)

$ bun run privacy:scan
Privacy scan passed

$ bun test tests/core-lab-boundary.test.ts tests/quota-reset-core-boundary.test.ts
 31 pass
 0 fail
```

**Every guard here was driven red before being trusted.** Two of them passed against the exact
code they were meant to catch, which is why this is stated rather than assumed:

- The burst test passed in-process against the unfixed seam, because earlier tests leave the
  observer module cached and a cached import resolves in call order. It now spawns a child
  process: 3/3 runs reported a false `surprise` before the fix, 0/3 after.
- The starvation test asserted the state file *existed*, which hydration already guarantees. It
  now truncates first and asserts content reached disk; removing the cap makes it fail.
- The boundary guard was attacked three ways: a static import in `src/router.ts` (4 assertions
  fail, including the two files that transitively reach it), a seam converted to a static import
  (1 fails), and the observer wiring deleted entirely — which fails the reachability assertion,
  proving the dynamic-only check is not vacuously satisfied by absent wiring.
- The redaction test fails when `webhookUrl` is removed from `SECRET_KEYS`.
- The LRU test fails against the previous insertion-order eviction.

**Activation evidence** — a real weekly rollover driven through the production writer with an
email account key produced exactly one webhook POST:

```
ACTIVATED: true
WEBHOOK_BODY: {"type":"quota_reset","kind":"scheduled","scope":"codex","accountTag":"9rhlw1hu",
              "window":"weekly","percentBefore":96,"percentAfter":2,...}
LEAK_email: false   LEAK_path: false   LEAK_url: false   LEAK_key_field: false
```

One observed-once, not-reproduced note for honesty: early in the work
`tests/codex-convergence-account-selectors.test.ts` timed out at 30s under parallel load. It
passes in 2.1s in isolation, does not touch quota writes, and did not recur in any later full
run. 5000 writes through the new seam cost 12.5ms, so the seam adds no measurable overhead.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added opt-in quota-reset detection for scheduled and unexpected usage-window resets.
  * Added webhook and local-command notifications with configurable filters, polling, timeouts, and network controls.
  * Added `provider resets` CLI support and a management API endpoint for viewing recent resets.
  * Added privacy-safe configuration handling, including webhook URL redaction.
* **Documentation**
  * Documented quota-reset notification settings, event payloads, CLI usage, and API access.
* **Bug Fixes**
  * Improved detection accuracy, duplicate suppression, persistence, and account isolation.
  * Added safeguards against notification delivery failures and unsafe webhook destinations.

### DIFF
--- a/devlog/_plan/260827_igwanu_bug_pr_merge_round/000_plan.md
+++ b/devlog/_plan/260827_igwanu_bug_pr_merge_round/000_plan.md
@@ -1,0 +1,116 @@
+# 000 — igwanu bug-PR merge round: plan
+
+Round base: `dev @ 8b1b65b8d` (local == `origin/dev`, verified 2026-08-27).
+Scope: the 13 open **bug**-labelled PRs. Four are Ingwannu's (#2767, #2766, #2764,
+#2761); nine are other authors' (#2747, #2745, #2740, #2733, #2729, #2726, #2693,
+#2638, #2497).
+
+Enhancement-labelled PRs are explicitly out of scope for this round.
+
+## The finding that orders the whole round
+
+Three PRs (#2767, #2764, #2747) show **failing required CI** — `ci`, `macos`,
+`test 3/4`, `gates` — while their own merged trees compile clean. The failure is
+not theirs. Every one of them fails the same repository-wide assertion:
+
+```
+error: package.json version 2.34.0 equals release tag v2.34.0, but this commit is
+not the one that tag names. The tree claims an already-published version.
+(fail) release version line > the in-tree version is never behind a released one
+```
+
+`dev` still carries `2.34.0` after tag `v2.34.0` shipped, so *every* PR opened
+after the release train inherits a red matrix. A second shared failure hits
+`gates`: `privacy:scan` reads the scp-style remote `git@github.com` recorded in
+`devlog/_plan/260827_release_train/020_preview_release.md` as an email address.
+
+**#2766 repairs both.** It is the keystone: until it lands, no other PR in this
+round can produce a trustworthy green matrix, and re-running their CI is wasted
+work. This is the inverse of the previous round's lesson — there, green checks
+were not evidence of health; here, red checks are not evidence of harm.
+
+Evidence: run `33081644562` job `98550259965` (#2767), run `33080634739` job
+`98546624127` (#2764), run `33059606933` job `98534630924` (#2747) — each shows
+`1 fail` and that one failure is `release version line`.
+
+## Merged-tree gate (this round's own evidence, not GitHub's)
+
+Every PR head was fetched, merged against `dev @ 8b1b65b8d` with
+`git merge-tree --write-tree`, committed as `mtp/<n>`, checked out to an isolated
+worktree sharing this repo's `node_modules`, and compiled.
+
+| PR | ahead | behind dev | merge-tree | tsc on MERGED tree |
+|---|---|---|---|---|
+| #2767 | 1 | 0 | CLEAN | OK |
+| #2766 | 2 | 0 | CLEAN | OK |
+| #2764 | 1 | 0 | CLEAN | OK |
+| #2761 | 1 | 2 | CLEAN | OK |
+| #2747 | 1 | 26 | CLEAN | OK |
+| #2745 | 2 | 26 | CLEAN | OK |
+| #2740 | 1 | 26 | CLEAN | OK |
+| #2733 | 1 | 43 | CLEAN | OK |
+| #2729 | 2 | 89 | CLEAN | OK |
+| #2726 | 1 | 63 | CLEAN | OK |
+| #2693 | 2 | 118 | CLEAN | OK |
+| #2638 | 2 | 179 | CLEAN | OK |
+| #2497 | 1 | **386** | **CONFLICT** | not reachable |
+
+The typecheck gate was itself verified rather than trusted: 12 runs finishing in
+~12s looked like a no-op, so a deliberate `const x: number = 'str'` was injected
+into a merged worktree and `tsc` returned `error TS2322`, exit 1. The speed is
+real — this repository is on the native TypeScript 7.0.2 compiler (~0.44s full
+typecheck). The gate works.
+
+## Cross-PR file contention
+
+`src/server/responses/core.ts` — **#2745, #2638, #2497**. Pairwise
+`git merge-tree` required before any second one of those lands; textual
+mergeability is not behavioral compatibility on the auth/routing boundary.
+
+`src/adapters/openai-chat.ts` — #2764 only. `src/adapters/openai-responses.ts`
+— #2767 only. `src/codex/auth-context.ts` — #2638 and #2497.
+No other file is touched by two in-scope PRs.
+
+## Loop-spec
+
+- Loop archetype: verifier-defined (spec-satisfaction repair per PR).
+- Write scope: `devlog/_plan/260827_igwanu_bug_pr_merge_round/`, `src/` and
+  `tests/` only where needed to land or reimplement a PR, plus PR metadata on
+  GitHub and `codex/` topic branches.
+- Out of scope: `main`, `preview`, releases, tags, npm publish, docs deploy,
+  enhancement PRs, force-push, history rewrite.
+- Bounds: `dev` is push-protected — every lane travels a `codex/` branch and a PR
+  targeting `dev`. `bun test` takes a machine-wide lock: one suite at a time,
+  long suites on `ssh lidge` via `ocx-run`. Never `OCX_TEST_NO_QUEUE=1`.
+
+## Work-phase map (one phase = one full PABCD cycle)
+
+| WP | Doc | Slice | Depends on |
+|----|-----|-------|------------|
+| wp1 | 000 | Docs-only roadmap: intake, merged-tree gate, contention map, lanes | — |
+| wp2 | 010 | **Keystone** #2766 — unblock the repository-wide CI gates | wp1 |
+| wp3 | 020 | Ingwannu remainder #2761, #2764, #2767 | wp2 |
+| wp4 | 030 | Clean approved lane #2733, #2726, #2747 | wp2 |
+| wp5 | 040 | Maintainer changes-requested #2745, #2729 | wp2 |
+| wp6 | 050 | Contributor remainder #2740, #2693, #2638 | wp2, wp5 |
+| wp7 | 060 | #2497 adjudication + round close-out | all |
+
+## Standing gates (inherited, all mandatory)
+
+1. Compile evidence comes from the MERGED tree, never the PR head alone.
+2. Any two PRs touching a shared file get `git merge-tree` before either merges.
+3. Green checks are not health unless the list includes `ci` / `test N/4` /
+   `macos`. **Corollary discovered this round: red checks are not harm until the
+   shared baseline is green.**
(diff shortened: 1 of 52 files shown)


lidge-jun:
## 리뷰 · 우선순위 68 / 80

이 PR은 사용량 창이 리셋됐을 때 알려 주는 기능을 넣습니다. 지금 `dev`에는 `src/quota/` 디렉터리가 없고, 쿼터는 `src/codex/quota.ts`와 `src/providers/quota.ts`에만 있습니다. 스케줄된 롤오버와 갑작스러운(surprise) 리셋을 둘 다 잡되, `quotaResetNotify`가 없으면 탐지도·폴러도·상태 파일도 만들지 않습니다. 기본 OFF가 핵심입니다.

구조는 단계별로 나뉩니다. `reset-detector.ts`는 순수 함수이고, `reset-seen-store.ts`는 재시작 뒤에도 한 번만 알리기 위한 내구성 클레임입니다. `reset-observer.ts`가 두 쿼터 시임에 붙고, `reset-poller.ts`는 대시보드를 안 열어도 밤새 리셋을 보게 하는 선택 폴러입니다. `fetchProviderQuotaReports` 호출자가 사실상 `/api/provider-quotas`뿐이라 폴러가 없으면 유휴 리셋은 영원히 안 보입니다. 전달은 webhook·로컬 command이고, `ocx provider resets`와 `GET /api/quota-resets`로 최근 이벤트를 읽습니다. 페이로드는 closed-union 라벨과 숫자만 담고, `webhookUrl`은 `config-command.ts` SECRET_KEYS에 넣습니다.

다만 PR에 `devlog/_plan/260827_igwanu_bug_pr_merge_round/` 문서 묶음이 같이 들어 있습니다. 그건 이미 끝난 머지 라운드 기록이고, 이 기능의 런타임과 무관합니다. 또한 `src/config.ts`(지금 3360줄)·`src/types/config.ts`에 `quotaResetNotify` 스키마를 더 얹습니다. types/config 분할 캠페인 배경에서는 새 섹션이 들어갈수록 나중에 옮길 면적이 커집니다. CI는 이 시각 기준 테스트/macos가 빨갛습니다.

라인 src/quota/reset-detector.ts - 순수 detect. MIN_SURPRISE_DROP_PERCENT=5, scheduled/surprise 구분, absent-prev는 null. 시계는 인자로만 받습니다.

라인 src/quota/reset-seen-store.ts - 재시작 후에도 같은 key를 한 번만 알리게 합니다. accountTag는 식별자·이메일이 아닙니다.

라인 src/codex/quota.ts / src/providers/quota.ts - 커밋된 스냅샷에 observe를 붙입니다. 들어오는 부분 페이로드만 보면 안 됩니다.

라인 src/quota/reset-poller.ts - 설정 enabled일 때만 돕니다. pollSeconds 0은 라이브 트래픽만 관찰합니다.

라인 src/cli/config-command.ts SECRET_KEYS - webhookUrl을 시크릿으로 취급합니다. Slack/Discord URL 자체가 권한입니다.

경로 devlog/_plan/260827_igwanu_bug_pr_merge_round/* - 이 기능과 무관한 과거 라운드 문서가 같이 들어 있습니다. 분리하거나 빼는 편이 리뷰·히스토리에 낫습니다.

경로 src/config.ts + src/types/config.ts - 새 섹션이 분할 전에 또 붙습니다. 스키마·write-validate·warn·diagnostics 네 자리가 빠지지 않았는지만 확인하면 됩니다.

메인테이너의 판단이 필요한 지점
- igwanu 라운드 문서를 이 PR에서 뺄지, 그대로 둘지
- CI 빨강이 기능 버그인지 스택/플래키인지 먼저 고칠지
- GUI 없이 CLI/API만으로 운영 알림을 충분하다고 볼지
- types/config 분할 전에 이 섹션을 `dev`에 올릴지

너의 추천
기능 방향은 `dev`에 맞습니다. 먼저 CI를 초록으로 만들고, igwanu 문서는 가능하면 이 PR에서 빼세요. 그다음 비작성자 승인 후 머지하세요. 기본 OFF·시크릿 레닥션·accountTag 비식별은 유지하세요.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
Requesting changes on exact head `ad74f037dcb5c47126ea7c0ca30f989b71da1afc`. I verified the following blockers in the current implementation; exact-head CI being green does not close these runtime boundaries.

1. `src/config.ts:861-869` accepts any URL scheme, and `src/quota/reset-sinks.ts:88-117` validates only the initial URL before calling default-following `fetch`. A public HTTPS URL can redirect the POST to loopback/private/metadata, while a plain HTTP URL exposes the payload and credential-bearing webhook path in transit. Require HTTPS at write/read validation, use manual redirects, and either reject redirects or revalidate and pin every hop with a focused redirect-to-loopback regression.
2. `src/quota/reset-poller.ts:14-80` states the minimum is above the 10-minute account TTL but sets it to 60 seconds. `startQuotaResetPoller()` also schedules the default argument rather than the configured `pollSeconds`, while `tick()` only checks whether the configured value is zero. Configuration changes therefore do not set the requested cadence. Add a single-in-flight and lifecycle-generation fence as well: current `setInterval` ticks can overlap, and an in-flight tick can finish and publish after `stopQuotaResetPoller()` or a restart.
3. `src/quota/reset-seen-store.ts:250-258` returns `true` even when `prune()` evicts the just-added claim or `persistNow()` fails. That violates the documented durable-claim-before-dispatch contract and permits duplicate notification after restart. Make persistence report success, verify the new claim still exists after pruning, and return true only after the claim is durably committed; roll back the in-memory insertion on failure.
4. `src/providers/quota.ts:2303-2311` reads `getAccountSet(provider)?.activeAccountId` without first proving the effective provider is OAuth. A stale/account-set entry for a key-auth provider overrides the API-key-pool discriminator and can merge or false-reset unrelated key observations. Gate that branch on the provider's effective `authMode === "oauth"`; otherwise always derive the key-pool identity.
5. `src/server/background-lifecycle.ts:63-74` installs the enabled delivery sink fire-and-forget after process loops start. Live quota writes in that startup window are silently unobserved because the seam is gated on the sink existing. When the feature is configured, activation must complete before request handling is admitted, with a startup-order regression; disabled installs may retain the cheap no-op path.

This PR also mixes the unrelated completed `260827_igwanu_bug_pr_merge_round` archive into a 52-file quota feature and is currently conflict-dirty against `dev`. Please remove/split that history, fix the runtime blockers, rebase onto current `dev`, and request fresh exact-head security review.

lidge-jun:
Landed via #3592 at 2188fcac8de8dcc14611e3c94c74bec010fc73f5

lidge-jun:
Superseded/completed: landed via #3592 at 2188fcac8de8dcc14611e3c94c74bec010fc73f5

PR #2794, opened 2026-08-28
fix(google-antigravity): gate the thought-signature sentinel correctly

lidge-jun:
## Summary

Reimplements #2693. Its diagnosis is correct — Gemini 3 rejects a turn whose first `functionCall` part carries no thought signature, and the official `skip_thought_signature_validator` bypass is the right remedy — but three defects sat between that intent and the diff. All three were reproduced independently by @Ingwannu's review and again here before any code was written.

## The three defects

**1. Presence check instead of `extractSignature()`.** The original tested `part.thoughtSignature !== undefined || part.thought_signature !== undefined`. The module already owns the real contract in `extractSignature`, which enforces `MIN_SIGNATURE_LEN = 16` and understands the nested `extra_content.google.thought_signature` shape. Consequences both ways: a valid **nested** signature was invisible, so a competing sentinel was added to an already-signed turn; a **too-short** value read as present, so the fallback was suppressed exactly where it was needed.

**2. `turnHasSignature` was a turn-wide boolean any sibling could set.** The requirement is about the *first* `functionCall` of a turn. In a two-call turn where only the second matches the cache, the second received a real signature and the first stayed unsigned — with the sentinel skipped. Gemini rejects that request.

**3. The sentinel was gated on `antigravityUsesReplayCache`**, which is `!/claude/i.test(model)`. Every non-Claude model qualified, so a Gemini-only control token was injected into `gpt-oss-120b-medium`.

## Two design decisions worth reviewing

**The sentinel is a separate pass, not part of `applyAntigravityReplay`.** Folding it in broke **12 existing tests**, and that was a design answer rather than a test problem: 18 assertions read `thoughtSignature === undefined` as "the cache did not match", covering eviction, TTL expiry, oversize refusal and clear-on-invalid. Writing a fabricated token into that slot overwrites the very signal those tests read. `applyAntigravityThoughtSignatureFallback` runs immediately after replay at both `src/adapters/google.ts` call sites, so a real signature always wins and a cache miss still looks like a cache miss. All 61 pre-existing tests pass untouched.

**The Gemini predicate matches a colon as well as a slash.** An earlier draft used `/(^|\/)gemini[-.\d]/i`. That is wrong here: `src/adapters/google.ts` builds the Vertex replay key as `vertex:<project>:<location>:<modelId>`. A slash-only regex looks correct against every CCA id and would have silently stripped the bypass from every Vertex Gemini request — a regression introduced by the fix intended to prevent one. Three namespaces reach this module and all three must match:

    gemini-3-pro                          -> true
    google/gemini-3-pro                   -> true
    vertex:api-key:global:gemini-3-pro    -> true
    gpt-oss-120b-medium                   -> false
    vertex:proj:global:gpt-oss-120b       -> false
    geminibot / my-gemini-clone           -> false

## Verification

`tsc --noEmit` exit 0. `tests/google-antigravity-replay.test.ts` **70 pass / 0 fail** (61 pre-existing + 9 new).

Mutation oracle, run per defect so each test is shown to bind its own:

| mutation | tests that fail |
|---|---|
| presence-check instead of `extractSignature` | 2 — nested, too-short |
| turn-wide flag instead of first-call | 1 — first-call sentinel |
| gate on `antigravityUsesReplayCache` | 1 — non-Gemini injection |
| slash-only regex | 2 — Vertex predicate, Vertex end-to-end |

Two of the new tests exist because a reviewer pointed out the first set could pass while a weaker patch still failed: a sibling signed **on the wire** with no cache involved, and the Vertex identity driven end to end rather than only asserted on the predicate.

## Integration-line disposition

No Go runtime counterpart. No authentication, credential, token, OAuth, workflow, or release surface — this is wire-shape handling for one provider's function-call metadata.

## Checklist

- [x] Scope limited to the signature fallback and its regressions.
- [x] All three reviewer-reproduced defects closed, each with a binding test.
- [x] Pre-existing behavior preserved — 61 prior tests untouched and passing.
- [x] Every changed decision mutation-verified.
- [ ] Exact-head required CI is green.
- [ ] A non-author maintainer has approved.

Supersedes #2693. Credit for the diagnosis and the original fix direction goes to @yxr1995-maker; credit for all three defect reproductions to @Ingwannu.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Gemini request handling when function calls lack valid thought signatures.
  * Added fallback support for Gemini and supported Vertex model identifiers.
  * Preserved valid signatures and left non-Gemini models and user messages unchanged.
  * Prevented fallback markers from being treated as genuine signatures during replay.

* **Tests**
  * Added coverage for signed, unsigned, nested, and multiple function-call scenarios, including replay and cross-context cases.

---

## Update: exact-head CI found what the focused suite did not

The first push passed `tests/google-antigravity-replay.test.ts` 70/0 and still failed three CI shards. Five tests in four *other* suites asserted `thoughtSignature === undefined`, and the sentinel filled that slot. Chasing why those suites disagreed — rather than assuming they were stale — turned up two real defects:

**1. The replay cache ingested the sentinel as a genuine signature.** `observeAntigravityReplay` stored it like any other value, so a token we fabricate on the way *out* round-tripped back *in* and was replayed later as evidence a turn was signed:

    observe(sentinel)  ->  {"sessions":1,"calls":1,"totalBytes":160}
    apply              ->  "skip_thought_signature_validator"

`extractSignature` now refuses it on both the direct and nested paths, so observing it leaves the cache empty (`sessions: 0`).

**2. `isLikelyRealThoughtSignature` accepted it.** That predicate exists to reject synthetic ids (`fc_`, `ctc_`, `tsc_`, `call_`) and is exactly what the issue #174 tests protect. The sentinel is alphanumeric with underscores, so it passed every filter and would have been treated as genuine wherever that predicate gates. Now rejected by name.

With both closed, the five assertions are genuinely proxies: they read `undefined` as "no signature was borrowed from another call, thread, or namespace", and a constant carries no other call's identity. Verified directly — thread-a records a real signature, thread-b replays the same call in a different session, and thread-b receives the sentinel, never thread-a's value. Each assertion now expects the sentinel and carries a comment explaining why its property still holds; #1312 keeps its positive control that thread-a still gets the real `SIGNATURE`, so a true cross-namespace leak would still fail.

Narrowing the sentinel to "fresh turns only" was considered and rejected: there is no honest replayed-history-vs-fresh-turn signal at either call site, and a first-turn request has no model `functionCall` to sign — every unsigned first `functionCall` *is* history, which is the 400 this PR fixes.

**Evidence at `3468ceaa0`:** full suite **15352 pass / 0 fail** (`rc=0`); the four affected suites **159 pass / 0 fail**; `tsc --noEmit` exit 0.

### DIFF
--- a/src/adapters/google-antigravity-replay.ts
+++ b/src/adapters/google-antigravity-replay.ts
@@ -640,16 +640,22 @@ const THOUGHT_SIGNATURE_BYPASS = "skip_thought_signature_validator";
  * Gemini-only control token was observed being injected into `gpt-oss-120b-medium`. Replaying a
  * signature upstream gave us is harmless for any model; *fabricating* a Gemini token is not.
  *
- * The boundary alternation covers the namespaces this module actually receives: a bare CCA id
- * (`gemini-3-pro`), a slash-prefixed id (`google/gemini-3-pro`), and the Vertex replay key built
- * in `src/adapters/google.ts` as `vertex:<project>:<location>:<modelId>`, whose separator is a
- * COLON — matching only `/` would silently skip every Vertex Gemini request. The trailing
- * `[-.\d]` keeps `geminibot` and `my-gemini-clone` out. A model outside this set that genuinely
- * needs the sentinel must arrive with a captured accepted CCA contract, not by widening this
- * predicate on inference.
+ * The identity must be REDUCED to its model component before matching, not scanned whole. The
+ * Vertex replay key is built in `src/adapters/google.ts` as
+ * `vertex:<project>:<location>:<modelId>`, and the project id is operator-chosen: a project
+ * named `gemini-prod` made a whole-string scan return true for
+ * `vertex:gemini-prod:global:gpt-oss-120b`, arming the Gemini-only sentinel for a non-Gemini
+ * model — the exact class of defect this predicate exists to prevent, reintroduced one layer up.
+ *
+ * So: take the last `:` segment for a Vertex identity, then the last `/` segment for a
+ * namespaced id (`google/gemini-3-pro`), and match only that. The trailing `[-.\d]` keeps
+ * `geminibot` and `my-gemini-clone` out. A model outside this set that genuinely needs the
+ * sentinel must arrive with a captured accepted CCA contract, not by widening this predicate.
  */
 export function antigravitySupportsThoughtSignatureSentinel(model: string): boolean {
-  return /(^|[/:])gemini[-.\d]/i.test(model);
+  const afterTransport = model.slice(model.lastIndexOf(":") + 1);
+  const wireModel = afterTransport.slice(afterTransport.lastIndexOf("/") + 1);
+  return /^gemini[-.\d]/i.test(wireModel);
 }
 
 /**
--- a/src/adapters/google-antigravity-replay.ts
+++ b/src/adapters/google-antigravity-replay.ts
@@ -625,6 +625,69 @@ export function antigravityUsesReplayCache(model: string): boolean {
 }
 
 /**
+ * Gemini 3 rejects a turn whose FIRST functionCall part carries no thought signature. When
+ * neither the wire metadata nor the replay cache can supply a real one, this is the official
+ * validator-bypass token.
+ */
+const THOUGHT_SIGNATURE_BYPASS = "skip_thought_signature_validator";
+
+/**
+ * True when the model speaks the Gemini wire dialect that requires a thought signature on the
+ * first functionCall of a turn — and therefore accepts the validator-bypass sentinel.
+ *
+ * Deliberately NOT `antigravityUsesReplayCache`. That predicate is broad on purpose (every
+ * non-Claude model participates in signature replay), and reusing it for the sentinel is how a
+ * Gemini-only control token was observed being injected into `gpt-oss-120b-medium`. Replaying a
+ * signature upstream gave us is harmless for any model; *fabricating* a Gemini token is not.
+ *
+ * The boundary alternation covers the namespaces this module actually receives: a bare CCA id
+ * (`gemini-3-pro`), a slash-prefixed id (`google/gemini-3-pro`), and the Vertex replay key built
+ * in `src/adapters/google.ts` as `vertex:<project>:<location>:<modelId>`, whose separator is a
+ * COLON — matching only `/` would silently skip every Vertex Gemini request. The trailing
+ * `[-.\d]` keeps `geminibot` and `my-gemini-clone` out. A model outside this set that genuinely
+ * needs the sentinel must arrive with a captured accepted CCA contract, not by widening this
+ * predicate on inference.
+ */
+export function antigravitySupportsThoughtSignatureSentinel(model: string): boolean {
+  return /(^|[/:])gemini[-.\d]/i.test(model);
+}
+
+/**
+ * Ensure every model turn's FIRST functionCall carries a thought signature, injecting the
+ * validator-bypass sentinel only where one is genuinely absent.
+ *
+ * Split out of `applyAntigravityReplay` on purpose. Replay answers "what did upstream already
+ * tell us about this call", and its absence of a signature is meaningful — 18 assertions in the
+ * suite read `thoughtSignature === undefined` as "the cache did not match", covering eviction,
+ * TTL expiry, oversize refusal and clear-on-invalid. Folding a fabricated token into that
+ * function would overwrite the very signal those tests read. Keeping the sentinel as its own
+ * pass means a cache miss still looks like a cache miss.
+ *
+ * Three properties this must hold, each of which a naive presence-check gets wrong:
+ *  - it decides from `extractSignature`, so a valid NESTED
+ *    `extra_content.google.thought_signature` counts as signed (no competing sentinel) and a
+ *    present-but-too-short value does not (the fallback still fires);
+ *  - it looks at the FIRST functionCall only, so a later sibling receiving a cached signature
+ *    cannot vote away the sentinel the first call requires;
+ *  - it is gated on the Gemini wire dialect, not on replay-cache participation.
+ */
+export function applyAntigravityThoughtSignatureFallback(model: string, contents: unknown[]): unknown[] {
+  if (!antigravitySupportsThoughtSignatureSentinel(model) || !Array.isArray(contents)) return contents;
+  for (const rawContent of contents as { role?: string; parts?: unknown[] }[]) {
+    if (!rawContent || typeof rawContent !== "object" || rawContent.role !== "model") continue;
+    if (!Array.isArray(rawContent.parts)) continue;
+    for (const rawPart of rawContent.parts) {
+      if (!rawPart || typeof rawPart !== "object") continue;
+      const part = rawPart as Record<string, unknown>;
+      if (!part.functionCall) continue;
+      if (!extractSignature(part)) part.thoughtSignature = THOUGHT_SIGNATURE_BYPASS;
+      break;
+    }
+  }
+  return contents;
+}
+
+/**
  * Observe a parsed CCA chunk's `candidates[0].content.parts` and record thought signatures keyed by
  * the functionCall identity (name + args). Accumulates across the whole session so a sequential
  * multi-step tool loop keeps EVERY prior call's signature, not just the latest part-index slot.
--- a/src/adapters/google-antigravity-replay.ts
+++ b/src/adapters/google-antigravity-replay.ts
@@ -512,10 +512,10 @@ export function antigravityReplaySessionKeysForTests(): string[] {
 
 function extractSignature(part: Record<string, unknown>): string | undefined {
   const direct = part.thoughtSignature ?? part.thought_signature;
-  if (typeof direct === "string" && direct.length >= MIN_SIGNATURE_LEN) return direct;
+  if (typeof direct === "string" && direct.length >= MIN_SIGNATURE_LEN && direct !== THOUGHT_SIGNATURE_BYPASS) return direct;
   const extra = part.extra_content as { google?: { thought_signature?: unknown } } | undefined;
   const nested = extra?.google?.thought_signature;
-  if (typeof nested === "string" && nested.length >= MIN_SIGNATURE_LEN) return nested;
+  if (typeof nested === "string" && nested.length >= MIN_SIGNATURE_LEN && nested !== THOUGHT_SIGNATURE_BYPASS) return nested;
   return undefined;
 }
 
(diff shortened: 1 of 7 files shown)


lidge-jun:
## 리뷰 · 우선순위 73 / 80

이 PR은 닫혀 있지 않은 #2693을 처음부터 다시 짠 고침입니다. 지금 `dev` HEAD `50e955604`의 `src/adapters/google-antigravity-replay.ts` `applyAntigravityReplay`(742줄 근처)는 세션 캐시에 서명이 있을 때만 채웁니다. 캐시가 없으면 contents를 그대로 돌려보냅니다. Gemini 3는 한 턴의 첫 `functionCall`에 thought signature가 없으면 400을 냅니다. 그래서 캐시 미스, 짧은 값, 중첩 서명만 있는 턴은 지금도 그대로 거절됩니다. 공식 우회 토큰은 `skip_thought_signature_validator`입니다. 그 방향 자체는 #2693이 맞았습니다.

다만 #2693은 세 곳에서 빗나갔습니다. 첫째, `extractSignature`(513-519줄, 길이 16 이상, `extra_content.google.thought_signature`도 봄) 대신 키가 있는지만 봐서, 중첩 서명은 없는 것처럼 센티널을 덧붙이고, 너무 짧은 값은 있는 것처럼 폴백을 막았습니다. 둘째, 턴 전체 불리언이라 둘째 호출만 캐시에 걸려도 첫째는 서명도 센티널도 없이 나갔습니다. 셋째, 게이트가 `antigravityUsesReplayCache`(623줄, 클로드가 아니면 참)라서 `gpt-oss-120b-medium` 같은 비-Gemini에도 Gemini 전용 토큰이 들어갔습니다. 이 PR은 그 셋을 각각 테스트로 묶었습니다. 기존 61개 테스트는 그대로 통과하고 9개를 더해 70개라고 적혀 있습니다.

센티널을 `applyAntigravityReplay` 안에 넣지 않은 선택은 맞습니다. 그 함수에서 `thoughtSignature === undefined`는 캐시 미스 신호입니다. 퇴거, TTL, 용량 초과, 무효화 테스트 18개가 그 신호를 읽습니다. 가짜 토큰을 거기 넣으면 그 신호가 지워집니다. 그래서 `applyAntigravityThoughtSignatureFallback`이 리플레이 바로 다음에 돕니다. `src/adapters/google.ts`의 두 호출점(앤티그래비티 본문 825줄 근처, Vertex 868줄 근처) 모두 그 순서입니다. Vertex 리플레이 키는 `vertex:<project>:<location>:<modelId>`라 구분자가 콜론입니다. 슬래시만 보면 Vertex Gemini에서 우회가 빠집니다. `/(^|[/:])gemini[-.\d]/i`가 그 세 이름 공간을 커버합니다. types.ts/config.ts 분할과 무관하고, preview deploy도 필요 없습니다.

라인 src/adapters/google-antigravity-replay.ts extractSignature - 키 존재가 아니라 길이 16 이상인 직접/중첩 서명만 진짜로 칩니다. 짧은 값은 폴백이 살아 있어야 합니다.
라인 src/adapters/google-antigravity-replay.ts applyAntigravityThoughtSignatureFallback - 모델 턴마다 첫 functionCall만 보고, 형제가 캐시 서명을 받아도 첫째 센티널을 지우지 않습니다.
경로/심볼 antigravitySupportsThoughtSignatureSentinel - 클로드가 아니면 참인 리플레이 캐시 게이트와 분리했습니다. gpt-oss에 Gemini 토큰이 안 들어갑니다.
경로/심볼 src/adapters/google.ts 두 호출점 - 리플레이 다음에 폴백이 붙습니다. Vertex는 콜론 키를 그대로 넘깁니다.
라인 tests/google-antigravity-replay.test.ts - 중첩, 너무 짧음, 첫 호출, 비-Gemini, Vertex 술어/종단을 각각 깨지게 묶었습니다.
경로/심볼 #2693 - 구현이 테스트만 있던 초안(당시 28/80)입니다. 이 PR이 실질 후속입니다. 아직 OPEN입니다.

메인테이너의 판단이 필요한 지점
- exact-head CI가 초록이 된 뒤 비작성자 승인만 받고 `dev`에 바로 머지할지
- 정규식이 Gemini 2.x(`gemini-2`)까지 포함하는 것이 괜찮은지. 본문은 Gemini 3만 말합니다
- 과거 모델 턴의 첫 호출에도 센티널을 넣을지, 마지막 턴만 넣을지
- #2693을 이 PR 머지 뒤에 `landed-via-maintainer`로 닫을지, 지금 superseded로 닫을지

너의 추천
exact-head CI가 초록이고 비작성자 승인이 있으면 `dev`에 머지하세요. 범위를 더 넓히지 마세요. 머지되면 #2693에 `Landed via #2794 at <commit>`를 남기고 `landed-via-maintainer`를 붙인 뒤 superseded로 닫으세요. 지금 당장 #2693을 합치지 마세요. 구현이 이쪽에 있습니다.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
Evidence summary for whoever reviews this — I authored it, so GitHub correctly refuses my approval and `MAINTAINERS.md` forbids self-approval. All technical gates are green; the only missing input is a second maintainer.

**Exact-head CI at `3468ceaa0`: 23 success, 0 failures, 1 intentional skip.**
**Full suite on that tree: 15352 pass / 0 fail (rc=0).**

## What this closes

Three defects in #2693, each reproduced independently before any code was written, each closed with a test that fails without its fix:

| defect | mutation applied | tests that fail |
|---|---|---|
| presence-check instead of `extractSignature` | restore the presence check | 2 — nested, too-short |
| `turnHasSignature` set by any sibling | restore the turn-wide flag | 1 — first-call sentinel |
| sentinel gated on replay-cache scope | gate on `antigravityUsesReplayCache` | 1 — non-Gemini injection |
| slash-only Gemini regex | drop the colon from the character class | 2 — Vertex predicate, Vertex end-to-end |

## Two things worth a reviewer's attention

**The sentinel is a separate pass, not part of `applyAntigravityReplay`.** Folding it in broke 12 existing tests, and that was a design answer rather than a test problem: 18 assertions read `thoughtSignature === undefined` as "the cache did not match", covering eviction, TTL expiry, oversize refusal and clear-on-invalid. Writing a fabricated token into that slot destroys the signal those tests read. Keeping it separate means a cache miss still looks like a cache miss, and all 61 pre-existing replay tests pass untouched.

**Chasing five cross-suite failures found two real defects, not assertion drift.** The replay cache was ingesting the sentinel as a genuine signature, so a token fabricated on the way out round-tripped back in and was replayed later as evidence a turn was signed. And `isLikelyRealThoughtSignature` — the predicate that exists to reject `fc_`/`ctc_`/`tsc_` synthetic ids, and what the issue #174 tests protect — accepted it, because it is alphanumeric with underscores. Both are now closed. Only after that are the five assertions genuinely proxies for "nothing was borrowed", which is the claim they were written to make.

An independent reviewer reached the same conclusion on separate reasoning, and flagged the `isLikelyRealThoughtSignature` hole as residual risk — the defect already fixed here.

## What to check if you review it

- `src/adapters/google-antigravity-wire.ts` — the sentinel is rejected by name; confirm that is the right place versus a broader shape rule.
- `tests/google-vertex-thought-signature.test.ts` #1312 — the isolation assertion now expects the sentinel, with the positive control that thread-a still receives the real `SIGNATURE`, so a genuine cross-namespace leak would still fail.
- The Gemini predicate `/(^|[/:])gemini[-.\d]/i` — the colon matters, because the Vertex replay key is `vertex:<project>:<location>:<modelId>`.

Credit for the diagnosis and the original fix direction goes to @yxr1995-maker; credit for all three defect reproductions to @Ingwannu.

Ingwannu (changes requested):
Requesting changes on exact head `3468ceaa0`. I reproduced one remaining functional blocker in `src/adapters/google-antigravity-replay.ts:652`: `antigravitySupportsThoughtSignatureSentinel` scans the entire raw Vertex replay identity instead of only the final model component. The Vertex call site constructs `vertex:${project}:${location}:${parsed.modelId}`, so a project ID beginning with `gemini-` enables the Gemini-only sentinel for a non-Gemini model.

In an isolated probe against this head, `vertex:gemini-prod:global:gpt-oss-120b` returned `true` and `applyAntigravityThoughtSignatureFallback` injected `skip_thought_signature_validator`; that identity must remain false/untouched. The positive controls `vertex:proj:global:gemini-3-pro`, `google/gemini-3-pro`, and `gemini-pro-agent` all returned true. (`gemini-pro-agent` is therefore not a failure of the current regex; the Vertex project-name false positive is the blocker.)

Please parse the final model component for `vertex:` identities while preserving bare and slash-prefixed Gemini IDs, and add a regression that uses a `gemini-*` project with a non-Gemini final model. Exact-head CI is green, but the current tests cover only a neutral Vertex project and therefore do not exercise this boundary. Re-request review on the updated head and I will verify it.

lidge-jun:
Fixed at `eebd1913e`. You were right, and the failure mode is worth naming: the predicate scanned the **entire raw identity**, so an operator-chosen project name could arm a model-scoped gate.

Reproduced your case exactly before changing anything:

    vertex:gemini-prod:global:gpt-oss-120b   -> true   (sentinel injected)
    vertex:gemini-team:us:claude-fable-5     -> true

That is the same class of defect the predicate exists to prevent — a Gemini-only token reaching a non-Gemini model — reintroduced one layer up. And my own tests could not catch it, because every Vertex case I wrote used a neutral project name. The positive control was doing double duty as the negative one.

The fix reduces to the model component before matching, rather than widening or blacklisting:

```ts
const afterTransport = model.slice(model.lastIndexOf(":") + 1);
const wireModel = afterTransport.slice(afterTransport.lastIndexOf("/") + 1);
return /^gemini[-.\d]/i.test(wireModel);
```

Last `:` segment for a Vertex identity, then last `/` for a namespaced id, anchored with `^`. Full matrix:

| identity | before | after |
|---|---|---|
| `vertex:gemini-prod:global:gpt-oss-120b` | **true** | false |
| `vertex:gemini-team:us:claude-fable-5` | **true** | false |
| `vertex:proj:global:gemini-3-pro` | true | true |
| `vertex:api-key:global:gemini-3.6-flash` | true | true |
| `google/gemini-3-pro` | true | true |
| `gemini-3-pro`, `gemini-pro-agent` | true | true |
| `gpt-oss-120b-medium`, `geminibot`, `my-gemini-clone` | false | false |

Regression added as you asked — a `gemini-*` project with a non-Gemini final model, asserting both the predicate and that `applyAntigravityThoughtSignatureFallback` leaves the part untouched, plus the positive control that the same project with `gemini-3-pro` still gets the sentinel.

Mutation-verified: restoring the whole-string scan fails exactly that test (70 pass / 1 fail); with the fix 71/0, and 160/0 across all four affected suites. `tsc --noEmit` exit 0.

Re-requesting your review on `eebd1913e`.

PR #2865, opened 2026-08-29
fix(cursor): bound the assembled external replay envelope

lidge-jun:
## Summary

Closes the last of the three residual bug issues from the bug-PR campaign.

Cursor's external workers reject a replay set over 192 root blobs or 512 KiB with a late `invalid_argument`, after hydrating every blob, so the upstream failure arrives with no usable measurement. Three request shapes reached that state:

- 192 checkpoint roots plus a two-root suffix each passed their own per-call check and emitted 194 roots, because nothing measured the *assembled* set.
- The empty-history early return in `rootPromptMessages` skips the pruning branch entirely, so 193 system prompts went out unbounded.
- A contiguous trailing tool-result block was byte-pruned but never count-pruned: the history limit only looked at `keptPrior`.

The guard now sits after final assembly — the first point where the wire content is fully known — and the same measurement feeds the `run-request` diagnostic, so telemetry cannot disagree with the guard. `rootBytes` previously read `rootPromptMessagesState?.byteLength ?? 0`, which reported `0` for a pure checkpoint and, for a suffix, counted a synthetic system root that had already been sliced off.

A root carried inside a decoded checkpoint need not exist in the local blob store — Cursor minted some of them, and a resumed conversation legitimately references ids this process never wrote. Such a root is counted, not fatal: the count limit still binds it (that is the 194-root case) and `unmeasuredRoots` marks `rootBytes` as a floor. A fail-closed version broke three already-passing checkpoint tests, which is the evidence that failing closed there rejects working continuation.

Also recovers the initiating `user` turn when retained history begins with an orphan `toolResult`, shrinking the result text to make room, so the model is not handed a tool result whose request it cannot see.

`CursorRootEnvelopeLimitError` is a local 400 carrying the measured counts, and is never retryable — checked ahead of the text heuristics in `isRetryableCursorError`.

Two things were designed and then deliberately not shipped, both caught by running mutations rather than by review:

- A `CursorRootMeasurementError` fail-closed path, deleted for the reason above. An unreachable error class cannot be tested, so it is not in the diff; the reasoning sits next to `isCursorRootEnvelopeError` so it is not re-added.
- A `replayOrigin` parameter gating orphan-result recovery to full replays. A mutation forcing the suffix branch could not be made red: `activeStart > 0` already confines the search to the current call's slice, so the flag decided nothing — and gating on it would have recreated the defect for a checkpoint suffix that does contain its initiating turn.

Native Composer and Auto paths are untouched (`isCursorExternalWireModel` gates the guard).

## Verification

- `bun test tests/cursor-blob.test.ts tests/cursor-request-builder.test.ts tests/cursor-transport-retry.test.ts` — **153 pass, 0 fail** on this exact head.
- Five mutations driven **red**, so no claim here rests on a suite only ever seen green:

| mutation | effect | failures |
| --- | --- | --- |
| `no-guard` | disable the external-model envelope check | 4 |
| `count-only` | drop the byte limit, keep the count limit | 1 |
| `bytes-only` | drop the count limit, keep the byte limit | 3 |
| `stale-bytes` | restore `rootPromptMessagesState?.byteLength ?? 0` in telemetry | 1 |
| `unmeasured-fatal` | throw instead of counting an unmeasurable root | 5 |

`stale-bytes` was green on the first attempt — the telemetry fix had no covering test, and the first test written for it set `checkpointSuffixStart`, which populates the very state the stale expression read. Only a pure checkpoint separates the two expressions; that is the shape the test now uses.

Not verified: no live Cursor request was sent. The 192-root and 512-KiB limits are taken as authoritative from the existing constants rather than re-probed upstream. This changes model-visible replay content and therefore estimated input usage, and it turns requests that previously went out silently into explicit local 400s.

## Checklist

- [x] Focused regression tests added next to the existing Cursor tests
- [x] No new logging of request bodies, keys, or account identifiers (`cursorBlobByteLength` returns a size, never content)
- [x] Targets `dev`
- [x] No auth, credential, OAuth, workflow, or release surface touched
- [ ] Docs update — not applicable; no user-facing configuration changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced Cursor replay limits for root count and total payload size before requests are sent.
  * Preserved initiating user instructions and tool results when trimming oversized history.
  * Prevented non-retryable replay-limit failures from being retried.
  * Improved diagnostics with measured root counts, sizes, and unmeasured roots.
  * Allowed valid checkpoint continuations that begin with a tool result.

* **Tests**
  * Added coverage for empty histories, checkpoints, native models, truncation, and count/size limit breaches.

### DIFF
--- a/devlog/_plan/260829_bugpr_lane_h_residual_issues/120_issue_1527_replay_envelope.md
+++ b/devlog/_plan/260829_bugpr_lane_h_residual_issues/120_issue_1527_replay_envelope.md
@@ -1,0 +1,276 @@
+# 120 — #1527: enforce the replay envelope on the final root set
+
+Revised after the plan audit returned FAIL. Blockers 8 through 13 applied. The
+first draft would have broken legitimate checkpoint continuation and measured the
+envelope in the one place where the true root set is not yet known.
+
+## Scope
+
+IN: `src/adapters/cursor/protobuf-request.ts` (`rootPromptMessages`,
+`buildPreparedCursorRunRequest`), `src/adapters/cursor/cursor-errors.ts`,
+`src/adapters/cursor/transport-retry.ts`, `src/adapters/cursor/native-exec.ts` (a
+read-only blob-size accessor), `src/adapters/cursor.ts` (terminal catch),
+`tests/cursor-blob.test.ts`, `tests/cursor-transport-retry.test.ts`,
+`tests/cursor-adapter.test.ts`.
+
+OUT: native Composer and Auto replay behavior, the #2277 checkpoint continuation
+design, and the transport retry policy itself.
+
+One PR: final-envelope construction, its typed failure mapping, and the
+regressions. The audit found no seam that splits cleanly, because the guard and the
+measurement it depends on are the same change.
+
+## Defect 1 — the envelope is measured on the wrong set
+
+`CURSOR_EXTERNAL_ROOT_BLOB_LIMIT` (192) and `CURSOR_EXTERNAL_ROOT_BYTE_LIMIT`
+(512 KiB) live at `src/adapters/cursor/protobuf-request.ts:71-73` and are checked
+against `keptPrior` only. System roots are all retained before the history budget is
+computed (`:309-380`), and trailing active tool results are byte-pruned but never
+count-pruned.
+
+Reproduced in isolated probes: 193 roots from 193 system prompts; 614,430 root bytes
+from a single 600 KiB system prompt; 194 roots from one system plus 193 trailing
+results. Each reported `continuationMode: "full-replay"`.
+
+Two further escapes the first draft missed:
+
+- **Cumulative checkpoint roots (blocker 9).** Suffix roots are appended AFTER
+  existing checkpoint roots (`:937-947`) but `rootPromptMessages` sees only the
+  suffix. A probe with 192 checkpoint roots plus a two-root suffix emitted 194.
+- **The empty-history early return (blocker 10).** `rootPromptMessages` returns
+  before external enforcement when `rawMessages` is empty (`:198-215`). A guard
+  placed only in the external-history branch leaves this path unbounded; a probe
+  emitted 193 system roots through it.
+
+Both mean the guard cannot live inside `rootPromptMessages`. The final root set
+exists only after checkpoint or full-replay assembly completes (`:959-974`), and
+that is where enforcement belongs.
+
+## Defect 2 — a tool result can be sent without the request that caused it
+
+External replay omits assistant tool-call roots by design (`:288`). The pruner puts
+every trailing tool result into `active` (`:321-326`), so one large result can
+consume the whole history budget and drop the preceding user turn (`:348-357`). The
+orphan guard preserves that lone result and sets `historyMessageStart` to it
+(`:365-369`); `conversationTurns()` then finds no current user turn and drops it from
+`turns[]` too (`:719`, `:763`).
+
+The emitted request is system roots, an assistant-role tool-result marker, zero turn
+blobs, and the generic Continue action. The audit's one correction to the framing:
+it is not literally instruction-free — system instructions and the Continue action
+are present. What is missing is the initiating user instruction, which is what makes
+a large-context turn answerable. A model given that answers in a handful of tokens,
+matching the report (200 OK, 4-19 output tokens at 80-95k input), reproduced with no
+Cursor account.
+
+`tests/cursor-blob.test.ts:474` asserts the oversized result and its truncation
+marker survive. It never asserts the initiating turn survives, which is why CI stayed
+green over a request shape that cannot work.
+
+## Blocker 8 — why the obvious rule was wrong
+
+The first draft said: never emit a result-only continuation. That is correct for full
+replay and WRONG for checkpoint continuation. Production sets `checkpointSuffixStart`
+to the covered message count (`src/adapters/cursor/request-builder.ts:479-488`), so a
+valid suffix may legitimately begin with a tool result — the initiating turn is
+already inside the checkpoint. `rootPromptMessages` receives only the sliced suffix
+(`:915-933`) and cannot recover it, so a blanket rule would reject or corrupt correct
+continuation.
+
+The distinction must be an explicit argument, not inferred from message shape. Pass a
+replay origin: `"checkpoint-covered"` only from the branch where checkpoint decoding
+succeeded (`conversationState` established at `:917-918`) and `checkpointSuffixStart`
+passed validation (`:919-926`); `"full-replay"` from every fallback path (`:959-960`).
+Do not derive it from the advisory `request.continuationMode`.
+
+## Change map
+
+### `rootPromptMessages` — replay origin
+
+Accept an explicit origin (`:198-204`).
+
+Full replay retains the initiating user or developer root together with every
+contiguous trailing tool-result root; if that group cannot fit, fail rather than emit
+a result-only replay. Checkpoint-covered replay may begin with tool results.
+
+### `rootPromptMessages` — atomic tool-result block (blocker 11)
+
+The first draft's "atomic block" contradicted the 192-root cap: 193 results plus a
+system root cannot be both. Atomicity means MEMBERSHIP: every result present in
+original order, or construction fails. Explicit text truncation of a result may
+remain, but whole-result deletion — `shift`, `filter`, marker omission — is forbidden
+(`:318-341`). A count-only assertion would let an implementation silently drop
+results and still pass, so tests assert result IDENTITY, not just count.
+
+### `buildPreparedCursorRunRequest` — one measurement, one guard
+
+Immediately after `conversationState` is final (`:959-994`), measure
+`rootPromptMessagesJson` once: `rootCount` from the final ID list; `rootBytes` as the
+sum of stored blob lengths, counting repeated IDs repeatedly; fail closed if any final
+ID is unmeasurable. For external models, reject when count exceeds 192 or bytes exceed
+512 KiB. Use this same measurement for the `run-request` telemetry.
+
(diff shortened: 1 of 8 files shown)


lidge-jun:
## 리뷰 · 우선순위 74 / 80

이 PR은 Cursor 외부 모델(Composer/Auto가 아닌 쪽)로 대화를 다시 보낼 때, 최종으로 조립된 root blob 묶음이 Cursor가 받는 한도(192개 / 512 KiB)를 넘지 않게 막는 수정이다. 지금 `dev`(e546c160b, #2859 GUI 자산·로그 폴링 직후)의 `src/adapters/cursor/protobuf-request.ts`에는 이미 `CURSOR_EXTERNAL_ROOT_BLOB_LIMIT` / `CURSOR_EXTERNAL_ROOT_BYTE_LIMIT`가 있고, `rootPromptMessages` 안에서 history만 잘라 낸다. 그런데 가드가 “중간 조각”만 보면 세 갈래로 새어 나간다. (1) 체크포인트에 이미 192 root가 있는데 suffix를 붙이면 합이 194가 된다. (2) history가 비어 있으면 early return이라 system root만 193개여도 통과한다. (3) 뒤에 이어진 tool result 덩어리는 바이트만 줄이고 개수로는 안 자른다. 이 PR은 조립이 끝난 직후, 즉 `conversationState.rootPromptMessagesJson`이 확정된 뒤에 한 번만 세고, 같은 숫자로 `run-request` 진단도 채운다. 그래서 가드와 텔레메트리가 서로 다른 값을 보지 않는다.

같이 고치는 두 번째 구멍은 “도구 결과만 있고, 그걸 시킨 사용자 말이 없는” 요청이다. 큰 tool result가 history 예산을 다 먹으면 initiating user turn이 잘리고, `conversationTurns()`는 현재 turn이 없다고 결과까지 버린다. 그러면 와이어에는 system + 맨손 tool-result 마커 + Continue만 남고, 모델은 입력은 큰데 출력은 몇 토큰만 낸다. 그게 #1527에서 본 short-answer 모양과 맞는다. 여기서는 살아남은 history가 toolResult로 시작하면 initiating user를 다시 끼우고, 안 들어가면 result 텍스트를 줄여 자리를 만든다. 체크포인트 suffix가 tool result로 시작하는 정상 경우는 `activeStart > 0` 때문에 “체크포인트가 이미 들고 있는 turn”을 다시 끌어오지 않는다. 플랜에 있던 `replayOrigin` 플래그는 mutation으로 필요 없음이 드러나 빠졌다.

로컬 타입 에러 `CursorRootEnvelopeLimitError`(code `cursor_root_envelope_limit`, status 400)를 추가하고, `isRetryableCursorError`가 텍스트 휴리스틱보다 먼저 false를 반환하게 했다. 측정 불가 root(체크포인트에만 있고 로컬 blob store에 없는 id)는 개수에는 넣고 바이트는 바닥값으로 두며 `unmeasuredRoots`를 남긴다. fail-closed로 던지면 이미 통과하던 체크포인트 테스트 세 개가 깨져서, 그 설계(`CursorRootMeasurementError`)는 일부러 안 실었다. Native Composer/Auto는 `isCursorExternalWireModel` 가드 밖이다. `bun test` 관련 스위트 153 pass와 mutation 다섯 개(red)를 PR이 적어 두었고, CI gates/hygiene/test 일부는 이미 통과 중이며 나머지 shard는 아직 돌고 있다. types.ts/config.ts 분리와 겹치지 않는 독립 bug PR이다.

라인 protobuf-request.ts(조립 직후 가드) - 최종 root 집합에만 한도를 걸어서, 체크포인트+suffix·빈 history·긴 tool-result 블록이 중간 가드를 우회하던 경로를 막는다. 방향은 맞다.

라인 protobuf-request.ts(orphan initiator 복구) - toolResult로 시작하는 유지 history에 user initiator를 다시 붙이거나 result를 줄인다. 다만 initiator를 넣어도 안 들어가고, active가 하나라서 shrink할 수도 없는 경우에는 그냥 통과한다. 그 순간에는 예전처럼 instruction 없는 모양이 남을 수 있고, 개수/바이트 가드가 나중에 잡아 주길 기대하는 구조다.

라인 native-exec.ts `cursorBlobByteLength` - 내용이 아니라 크기만 돌려 줘서 봉인된 request-scope를 깨지 않는다. 측정 불가 id는 null이다.

라인 cursor-errors.ts / transport-retry.ts - 로컬 한도 초과를 재시도하지 않게 했다. 같은 요청을 다시 보내면 같은 거절이 난다.

경로 src/adapters/cursor.ts 터미널 catch - 플랜 수용 기준 8과 본문 설명은 어댑터가 status 400 / `invalid_request_error` / code / `retryable: false`를 이벤트에 보존한다고 적혀 있다. 그런데 이 PR 디프에는 `cursor.ts` 변경이 없다. `prepareCursorRunRequest`가 live-transport 안에서 던지면 바깥 catch는 `safeCursorTransportError`로 메시지만 내보내고, translator-budget 때처럼 status·code를 붙이지 않는다. `isRetryableCursorError`는 클래스 단위 테스트만 있고, 어댑터 이벤트 매핑 테스트는 없다. 클라이언트/Codex 쪽이 “명시적 로컬 400”이 아니라 일반 upstream 실패처럼 볼 수 있다.

경로 바이트 한도 + unmeasuredRoots - 로컬에 없는 체크포인트 root는 바이트 합에서 빠지므로, 실제 와이어 바이트가 512 KiB를 넘어도 가드가 통과할 수 있다. 개수 한도는 그대로 묶는다. 의도된 floor이지만, 바이트 한도만으로는 닫히지 않는 구멍이 남는다.

경로 PR 본문 vs #1527 - “잔여 버그 이슈 마지막”이라고 쓰였지만 disposition은 #1527을 열어 둔다. GitHub `Closes #1527` 키워드는 없어 자동 종료는 안 된다. 맞다. 머지 댓글에 “#1527 중 replay-envelope 메커니즘만 닫고, direct-vs-proxy 대조는 남긴다”고 한 줄 적으면 기여자 혼선이 줄어든다.

메인테이너의 판단이 필요한 지점
- `cursor.ts` 터미널 catch에 envelope 에러 매핑을 이 PR에 넣을지, 후속 한 줄 PR로  manifest지
- unmeasured root 때문에 바이트 한도가 floor인 상태를 운영상 받아들일지, 아니면 상한을 더 보수적으로 둘지
- #1527을 이 PR 머지 후에도 열어 둘지(플랜대로), 부분 해결 라벨만 붙일지

너의 추천
CI 남은 test/macos/windows shard가 초록이면, `cursor.ts` catch에 `isCursorRootEnvelopeError` 분기(status 400, code, retryable false)를 같은 PR에 아주 짧게 추가한 뒤 `dev`로 머지하는 쪽을 추천한다. 매핑을 빼기로 했다면 플랜/본문의 “어댑터 보존” 문장을 고치고 테스트를 그에 맞게 남긴 다음 머지해도 된다. #1527은 자동 닫지 말고, 머지 후 “envelope 가드·initiator 복구 반영, 계정 비대칭 대조는 남음” 댓글을 남긴다. types/config 분리와 무관한 독립 수정이라 close-don't-rebase 대상이 아니다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
Reviewed exact head 3030120b1f65626cda2124880d8dc4f5919fdf3c. The final-envelope measurement and non-retryable transport classification are valuable, and the focused suites pass 153/153, but two correctness blockers remain. (1) In `rootPromptMessages`, when two or more active trailing tool results fit by themselves but the initiating user plus those results exceeds `historyBudget`, the recovery branch only truncates when `historyEntries.length === 1`; the multi-result replay stays result-only and passes the later count/byte guard. Retain the initiator and every result identity by bounded truncation, or reject locally, and add a two-result regression. (2) `CursorRootEnvelopeLimitError` declares status 400, code `cursor_root_envelope_limit`, and deterministic non-retryability, but the outer catch in `src/adapters/cursor.ts` emits only the sanitized message. Preserve `{status: 400, errorType: "invalid_request_error", code: "cursor_root_envelope_limit", retryable: false}` in the adapter event and cover the end-to-end event mapping. Keep #1527 open after this lands because the live direct-vs-proxy/account comparison remains unresolved.

Ingwannu (changes requested):
Reviewed exact head `72eaab9e19177a99ccd7f3d6ebf85c590cbb03ad`.

The adapter mapping blocker is fixed: the local envelope failure now preserves status 400, `invalid_request_error`, `cursor_root_envelope_limit`, and `retryable: false`. The selected `cursor-blob` and `cursor-adapter` suites pass 114/114 under an isolated test home.

One correctness blocker remains. Under count pressure the new initiator-recovery branch still executes `kept.shift()` and silently deletes result identities. I reproduced the PR's 191-small-result shape with unique `RESULT_ID_0..190` markers: the initiating user is restored, but `RESULT_ID_0` / `call_0` is absent from the assembled roots. The added regression only asserts that the initiator exists, so it passes despite that loss. The three-oversized-result test exercises the earlier equal-byte-share path and does not cover this count-boundary deletion.

At the 192-root boundary, `system + initiator + 191 result roots` cannot all fit as separate roots. Preserve every result identity through a representation that fits, or reject this request locally; do not silently discard the oldest result. Add a count-pressure regression that asserts every result id/unique marker survives, or asserts the typed local rejection. Please keep #1527 open after this replay-envelope slice lands because the live direct-vs-proxy/account comparison remains unresolved.

Ingwannu (changes requested):
Re-reviewed exact head b6652084ef323fa813190d520eeaf7c41e0c36a2. The only delta from the previously tested 72eaab9e head removes the test suppression by typing the throwing generator; production code and the count-pressure regression are unchanged. The reproduced blocker therefore remains: the 191-result path restores the initiator by shifting out call_0 / RESULT_ID_0, while the checked-in test asserts only the initiator. Preserve every result identity or reject locally, and add the missing all-identities assertion before requesting another review.

PR #2872, opened 2026-08-29
fix(codex): complete prompt probe coalescing

lidge-jun:
## Summary

- Supersedes #2870 while preserving and explicitly crediting @luvs01's four commits. Their work correctly coalesces concurrent `codex debug prompt-input` callers behind one child process, shares success and failure outcomes, isolates one caller's abort from other waiters, cancels the child when every waiter abandons it, and retains the slot until the exact child closes.
- Fixes the remaining stale-result defect without changing the optimistic-concurrency revision contract. Probe admission now uses a separate state fingerprint containing the config/store revision plus the active selected base variant's exact file bytes, so an edit returns `busy` while the old child drains instead of joining it or spawning concurrently.
- Adds the missing overlapping-failure regression: both callers receive `codex debug prompt-input failed`, exactly one child starts, and a later caller starts a fresh child rather than replaying a poisoned rejection.
- Aborts the dashboard panel's prompt-text request on unmount so server-side cancellation is exercised instead of merely suppressing the eventual state update.

This touches a management API route, which repository policy treats as a security boundary and therefore requires explicit non-author security review.

## Verification

- `bun test` over the seven affected backend prompt files: 191 pass, 0 fail.
- `bun test` over the six affected dashboard prompt files: 87 pass, 0 fail.
- Post-mutation targeted rerun: 3 pass, 0 fail.
- `bun x tsc --noEmit`: pass.
- `bun run build` from the dashboard package: pass (existing chunk-size warning only).
- `bun run privacy:scan`: pass.

| Regression | Temporary mutation | RED evidence |
| --- | --- | --- |
| Selected-variant edit | Replaced the state fingerprint with the unchanged transaction revision | Post-edit caller returned `# Old\nold-body`; 0 pass / 1 fail |
| Overlapping failure cleanup | Removed active-flight cleanup after child close | Later caller reused the failed flight; expected 2 spawns, received 1; 0 pass / 1 fail |
| Panel unmount cancellation | Restored boolean-only cleanup and removed the fetch signal | Request carried no abort signal; 0 pass / 1 fail |

## Checklist

- [x] Docs or release notes were updated when needed. No documentation change is needed for these correctness and request-lifecycle fixes.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. The fingerprint is internal-only, hashes prompt state without serializing it, and leaves route authentication and response contracts unchanged; non-author security review is still required before merge.

## Summary by CodeRabbit

- **New Features**
  - Prompt text previews now cancel cleanly when the prompt panel is closed or unmounted.
  - Preview results refresh when the selected base prompt, variant content, or instruction files change.
  - Concurrent preview requests are coordinated to avoid duplicate processing and stale results.

- **Bug Fixes**
  - Prevented outdated prompt previews from appearing after prompt edits.
  - Improved handling of interrupted, timed-out, or failed preview requests.
  - Ensured cancelled previews no longer leave background processing active.

## UI change

Codex Set - Prompt layers panel

The panel whose fetch effect changed, captured against an isolated proxy on port 10877 with its own HOME/OPENCODEX_HOME/CODEX_HOME (never the operating instance on 10100). The layout is unchanged — the diff here is behavioral: the effect now aborts its in-flight request on unmount so the server can cancel the probe child it spawned, while keeping the boolean guard that prevents a post-await state write.

The probe reports `ok:false` in that sandbox because no Codex binary is installed there, which is the fallback branch this effect already handled; byte counts populate only where Codex is present.

### DIFF
--- a/devlog/_plan/260829_bugpr_lane_h_residual_issues/130_pr2872_probe_fingerprint.md
+++ b/devlog/_plan/260829_bugpr_lane_h_residual_issues/130_pr2872_probe_fingerprint.md
@@ -130,13 +130,42 @@ What remains uncovered:
 - **Clock, timezone, shell.** Not files. A fingerprint over a clock is not a
   fingerprint.
 
-The exposure is an external edit landing inside a single in-flight probe's window,
-whose failure mode is a fail-closed `busy` and a retry, not corruption.
+The exposure is an external edit landing inside a single in-flight probe's window. Be
+precise about the failure mode, because an earlier draft of this sentence got it
+backwards: for an input the key does not cover, the key does not move, so the caller
+DOES join and DOES receive the older rendering. Fail-closed `busy` is what happens for
+a covered input. An uncovered one is a stale read of one layer's text, bounded to that
+window, in a read-only inspection view.
 
 This section has now been wrong twice, in the same direction both times: something was
 called unreadable when it was merely inconvenient to read. The standard that survived
 is narrow — an input belongs on this list only when no file on disk determines it.
 Anything with a path gets hashed.
+
+## Why this is a bounded key and not a total one
+
+Nine rounds in, the useful conclusion is about the shape of the specification rather
+than any single input. "Hash everything the rendered prompt depends on" is closable
+only against a pinned Codex: the dependency graph belongs to Codex, is private, and
+moves independently of this repository. A new config field or a changed precedence
+upstream silently widens the gap without anything here changing.
+
+So this is a bounded invalidation key over known local inputs, and the code says that
+rather than implying identity.
+
+A design that needs no enumeration exists and was assessed: admit on TIME, where a
+request may join a probe only if the probe started after the request arrived. The
+correctness argument holds — such a probe read the filesystem after every write that
+completed before the request — and it needs a monotonic in-process ordinal rather than
+a clock. It was not adopted because it removes almost all the coalescing that motivated
+the work: a probe spawns immediately, so the ordinary second caller arrives after the
+start and would always be refused. Recovering both properties means cohort batching —
+hold arrivals briefly, spawn once the cohort is closed — which is a different change
+from this one.
+
+That is a real option, not a dismissal, and it belongs to whoever needs a strict
+"never older than my arrival" contract. What ships here is the bounded key, which is
+strictly better than the revision-only key it replaces.
 
 ## The reader, and why it stopped being a regex
 
--- a/devlog/_plan/260829_bugpr_lane_h_residual_issues/130_pr2872_probe_fingerprint.md
+++ b/devlog/_plan/260829_bugpr_lane_h_residual_issues/130_pr2872_probe_fingerprint.md
@@ -91,6 +91,36 @@ written to fix the previous finding.
    is also the write-path concurrency token, so the collision reached further than
    the probe.
 4. An external base selection was hashed as a bare kind string.
+5. Two more: a relative `model_instructions_file` was resolved against the proxy's
+   own working directory instead of the config file's, so it hashed an unrelated
+   file; and only the two built-in project-document names were considered, so a
+   configured `project_doc_fallback_filenames` entry could be edited unnoticed.
+
+## The pattern, and where it stops
+
+Five rounds is the interesting part of this record. Each fix was reviewed, and four
+of the six findings were in code written to close the previous finding. The reason is
+consistent: a cache key is only as good as its worst-covered input, and "I added the
+input I was told about" is not the same as "the key names everything the output
+depends on". Framing, path resolution, and candidate-set breadth each failed
+separately.
+
+What remains uncovered is now a short and deliberate list, and every item on it has a
+reason that is not "we did not get to it":
+
+- **Ancestor project documents.** Codex walks from a project root to its cwd. The
+  probe runs in `CODEX_HOME` with no checkout around it, so there is no ancestor
+  chain to walk. Hashing a walk that cannot happen would be noise.
+- **Skill and plugin metadata.** Readable, but they are directory trees owned by
+  Codex, not files with a stable enumeration contract here. Hashing them means
+  duplicating a loader we do not own and would have to keep in sync.
+- **Clock, timezone, shell, MCP availability.** Not files. A fingerprint over a clock
+  is not a fingerprint.
+
+The exposure for each is the same and it is small: an external edit landing inside a
+single in-flight probe's window, whose failure mode is a fail-closed `busy` and a
+retry, not corruption. The honest move is to name them here rather than to imply the
+key is total.
 
 ## Verification
 
--- a/devlog/_plan/260829_bugpr_lane_h_residual_issues/130_pr2872_probe_fingerprint.md
+++ b/devlog/_plan/260829_bugpr_lane_h_residual_issues/130_pr2872_probe_fingerprint.md
@@ -1,0 +1,78 @@
+# 130 — #2872: the probe admission fingerprint omits the instruction files it renders
+
+Written after an independent adversarial review of PR #2872 at head `a7504ab8e`
+returned BLOCKER FOUND, and after a plan audit of the first fix draft returned
+PLAN NEEDS CHANGE. Both verdicts are applied here.
+
+## Scope
+
+IN: `src/codex/prompt-layers.ts` (`computePromptProbeStateFingerprint`),
+`tests/codex-prompt-route.test.ts` (route-level regression).
+
+OUT: the coalescing machinery itself (`runSharedPromptProbe`, waiter accounting,
+the `busy` fail-closed policy) — reviewed and found sound. Also out: fingerprinting
+state this process cannot observe, discussed under "What this does not cover".
+
+## Defect — a post-write reader joins a pre-write flight and gets stale text
+
+`computePromptProbeStateFingerprint` (`src/codex/prompt-layers.ts:643`) hashes
+`config.toml`, `opencodex-prompt.json` (through `computeRevision`) and the selected
+base variant `.md`. It does not hash `$CODEX_HOME/AGENTS.md`.
+
+`probePromptText` runs the child with `cwd = resolveCodexHomeDir()`
+(`src/codex/prompt-text-probe.ts:400`) and extracts that file's body as the
+`__agents_md` layer (`:365`). The fingerprint is a component of `commandKey()`
+(`:137-145`), which is the sole admission identity in `runSharedPromptProbe`
+(`:302`). So an `AGENTS.md` edit leaves the key unchanged, the next request matches
+`active.key`, joins the in-flight pre-write probe, and is served pre-write text.
+
+Reproduced deterministically at `a7504ab8e`: identical fingerprints before and
+after the write, and both callers received `"old-agent-text"`.
+
+This is the same class of bug the fingerprint was introduced to fix. The original
+`revision` covered only config/store transaction bytes, so editing the selected base
+variant changed the prompt without moving the revision. Naming one more uncovered
+input does not change the shape of the defect: admission identity must name every
+input the probe renders.
+
+## Fix
+
+Hash the `CODEX_HOME` instruction files into the fingerprint.
+
+The path is `resolveCodexHomeDir()`, **not** `dirname(activeConfigPath(opts))`. The
+plan audit rejected the latter and it is right: `tests/codex-prompt-route.test.ts:115-125`
+injects `codexPromptPaths` at a fixture root while setting `CODEX_HOME` to a separate
+decoy, precisely so a route that ignored the injected paths is caught. Deriving the
+`AGENTS.md` path from `configPath` would name a file the probe never reads, and the
+regression would pass while production stayed broken.
+
+Both spellings are hashed, in Codex's own precedence order: `AGENTS.override.md`
+is preferred over `AGENTS.md`, so an override edit must move the key too. Absent
+files hash to a distinct sentinel, so create and delete both move the key.
+
+## What this does not cover, stated rather than implied
+
+The guarantee is bounded to OpenCodex-managed writes plus the `CODEX_HOME`
+instruction files. It is not complete prompt-state identity, and the code says so
+instead of implying otherwise:
+
+- Skill metadata, plugin manifests, and MCP/app availability feed
+  `<skills_instructions>`, `<plugins_instructions>` and `<apps_instructions>`.
+- Clock, timezone, shell and permission state feed `<environment_context>`.
+- An external `model_instructions_file` target. `codex debug prompt-input` discards
+  base instructions, so it does not currently stale a rendered layer.
+
+None is writable through `/api/codex-prompt`; each needs an external edit
+concurrent with an in-flight probe. A 15-second window bounded by a fail-closed
+`busy` is the exposure, and pretending to fingerprint a clock would be worse than
+documenting it.
+
+## Verification
+
+Route-level, in the file whose fixture separates `CODEX_HOME` from the injected
+paths — the only place this can fail honestly. Two callers separated by an
+`AGENTS.md` write must not share a flight: the second returns `busy`, and a later
+request returns the new text. Repeated for `AGENTS.override.md`.
+
+Named mutation: delete the instruction-file contribution from the fingerprint. The
+regression must go red with identical keys and one spawn.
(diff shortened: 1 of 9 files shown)


lidge-jun:
## 리뷰 · 우선순위 70 / 80

지금 `dev` HEAD는 `8d1dc1f5d`입니다. 직전에 올라간 것은 #2869(systemd 219 / CentOS 7에서도 설치·정리가 되게 LoadState 파싱·stop/disable 분리·ExecStart 로그 리다이렉트)이고, 그 아래에는 #2868(reserved-prefix call-id 코덱)이 있습니다. 패키지는 2.36.0입니다. 이 PR은 types.ts/config.ts 분할과 겹치지 않습니다. Codex 프롬프트 텍스트 프로브의 동시 실행·취소·무효화를 끝까지 맞추는 수정이고, 연결된 이슈 번호는 없습니다. #2870을 이어 받아 @luvs01의 네 커밋을 그대로 두고, 그 위에 남은 구멍만 메운 유지보수자 완료 PR입니다.

프롬프트 설정 화면은 레이어가 실제로 몇 바이트를 쓰는지 보여 주려고 `GET /api/codex-prompt/text`를 칩니다. 그 엔드포인트는 `src/server/management/codex-prompt-routes.ts`에서 `probePromptText()`를 부르고, `src/codex/prompt-text-probe.ts`가 설치된 Codex로 `codex debug prompt-input` 자식을 띄운 뒤 JSON을 읽습니다. 지금 HEAD의 `runProbe`는 호출마다 자식을 새로 만듭니다. 타임아웃이나 종료 전에는 프로세스가 남고, 라우트도 `req.signal`을 안 넘깁니다. 화면을 닫거나 요청이 겹쳐도 이미 뜬 자식은 끝까지 갑니다.

#2870이 이미 한 일은 이렇습니다. 같은 명령 키면 한 비행(flight)을 여러 호출자가 같이 기다리고, 한 사람만 취소해도 다른 대기자는 결과에 남고, 마지막 대기자가 떠나면 자식을 끊고, `close`가 오기 전에는 슬롯을 비우지 않습니다. 다른 키가 오면 `busy`로 부드럽게 거절합니다. 이 PR은 그 골격을 유지한 채, #2870만으로는 막히지 않던 구멍을 닫습니다.

그 구멍은 낙관적 동시성 `revision`과 프로브 입장이 같다고 가정한 데서 옵니다. `computeRevision`은 config/store 거래 바이트만 봅니다. 선택된 base variant의 `.md`를 고치면 화면에 보이는 프롬프트는 바뀌는데 revision 숫자는 그대로일 수 있습니다. 그래서 옛 자식이 도는 동안 새 호출자가 같은 키로 합류하면, 방금 고친 내용이 아니라 옛 본문을 받을 수 있습니다. 이 PR은 `src/codex/prompt-layers.ts`에 `computePromptProbeStateFingerprint()`를 두고, revision에 더해 선택된 base의 kind/id와 그 파일 바이트를 SHA-256으로 넣습니다. 라우트는 그 지문을 `probePromptText`에 넘깁니다. 편집 직후 호출은 옛 비행에 합류하지 않고 `busy`가 되고, 자식이 빠진 뒤 새 자식이 뜹니다.

테스트도 그 시나리오를 직접 깨뜨립니다. 선택된 variant를 고친 뒤 revision이 그대로인 채로 GET text를 치면 busy가 나오고 spawn은 한 번만, 나중 호출은 새 본문을 봅니다. 겹친 실패 뒤에는 두 호출자가 같은 실패 문구를 받고 자식은 한 번만 뜨며, 다음 호출은 독이 든 거절을 재사용하지 않고 새 자식을 띄웁니다. 대시보드 `gui/src/pages/codex-set-prompt.tsx`는 언마운트 때 boolean 취소 대신 `AbortController`로 fetch를 끊어서, 서버 쪽 취소를 실제로 타게 합니다. 작성자 검증은 백엔드 프롬프트 일곱 파일 191, 대시보드 여섯 파일 87, tsc/build/privacy:scan 통과라고 적혀 있습니다.

이 경로는 management API라서 저장소 정책상 보안 경계입니다. 지문은 내부 상태만 해시하고, 인증·응답 계약은 그대로이며, 여전히 호출자 cwd를 받지 않습니다. 그래도 작성자 본인 리뷰만으로는 부족하고, 비작성자 보안 확인이 필요합니다. 지금 `dev`에 남아 있는 #2855 이후 거버넌스 구멍(자격 증명이 저장소 소유자로만 잡혀 비작성자 승인이 형식적으로 비는 문제)과도 같은 줄입니다.

`src/codex/prompt-layers.ts` `computePromptProbeStateFingerprint` - 지문에 들어가는 건 revision + 선택된 base variant 파일뿐입니다. 같은 프로브 결과에 영향을 줄 수 있는 `AGENTS.md`나 스킬/플러그인 쪽 파일 바이트는 안 넣습니다. management 쓰기 경로의 revision 구멍은 막았지만, 홈 디렉터리에서 파일이 바뀌는 다른 경로는 옛 비행에 합류할 여지가 남습니다.
`src/codex/prompt-text-probe.ts` `activePromptProbe` - 프로세스 하나짜리 전역 슬롯입니다. 관리 API 워커가 여러 프로세스면 슬롯이 공유되지 않아, 프로세스마다 자식이 하나씩 뜰 수 있습니다. 단일 프로세스 가정이면 괜찮고, 그 가정이 문서/배포에 드러나 있는지만 보면 됩니다.
`gui/src/pages/codex-set-prompt.tsx` 프로브 효과 - HTTP는 200이고 body의 ok만 false인 busy/실패를 그대로 `layerText`에 넣습니다. 자동 재시도는 없습니다. busy가 짧으면 바이트 칸이 한동안 비어 보일 수 있습니다. 의도된 실패-부드러움인지, 짧은 재시도가 필요한지는 제품 판단입니다.
`src/server/management/codex-prompt-routes.ts` GET `/api/codex-prompt/text` - 여전히 자식 프로세스를 띄우는 management 경로입니다. 이번 변경은 지문·취소·합류 규칙이지, 인증이나 출력 상한(8 MiB)을 새로 조인 것은 아닙니다. 비작성자 보안 확인 체크리스트에 올리는 이유가 여기입니다.
#2870 - 이 PR이 명시적으로 대체합니다. #2872가 랜딩하면 #2870은 `landed-via-maintainer`로 닫고, 기여자 커밋 크레딧은 이미 본문에 있으니 중복 리뷰/머지를 막으면 됩니다.

메인테이너의 판단이 필요한 지점
- 지문 범위를 selected base까지만 둘지, 프로브가 읽는 다른 홈 파일까지 넓힐지.
- busy 응답을 GUI에서 조용히 실패로 둘지, 짧게 한 번 더 칠지.
- management 경로라서 비작성자 보안 리뷰를 누가/어떻게 남길지(현재 거버넌스 구멍과 동일).
- #2872 머지 직후 #2870을 바로 supersede close 할지.

너의 추천
비작성자 보안 한 줄 확인만 받은 뒤 #2872를 `dev`에 머지하세요. #2870은 머지하지 말고, 랜딩 커밋으로 `Landed via #2872 at <sha>` 댓글 + `landed-via-maintainer` 라벨 후 닫으세요. types/config 분할과는 무관하니 close-don't-rebase 대상이 아닙니다. GUI busy 재시도는 이번 머지에 묶지 말고, 바이트 칸이 비는 체감이 나오면 후속 이슈로 빼세요.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
@Ingwannu — security review requested. This changes a management API route, which `MAINTAINERS.md` treats as a security boundary, so it needs non-author approval and I am not self-merging it.

**What to look at, concretely.** The probe now coalesces concurrent `codex debug prompt-input` invocations behind a single child process, keyed by a fingerprint. The security-relevant question is whether that fingerprint can ever serve one caller's prompt text to a caller who should have seen something different. The contributor's original key was the transaction revision alone, which was demonstrably insufficient: editing the selected base variant writes a separate `.md` file and can leave config/store bytes unchanged, so the revision did not move and a second caller joined the stale flight and received pre-edit text. That is reproduced and now fixed with a fingerprint that also covers the selected variant's bytes.

**The gap I am explicitly flagging rather than papering over.** The fingerprint does not represent account or auth generation. If the prompt Codex builds for `prompt-input` varies by authenticated account — and there is reason to think it can, since Codex constructs an auth-aware session and git-attribution policy for that command — then two callers on different accounts could in principle share one flight. We did not add it because there is no single established account-generation authority to key on here, and inventing one inside this patch risked coupling to the wrong mechanism on a security boundary. Your call on whether that is acceptable for this route or needs to land before merge.

**What is verified.** Coalescing, shared-failure propagation, per-caller abort isolation, and in-flight slot cleanup were each checked adversarially and behave correctly, including that a failed flight is not cached and a later caller starts fresh. Three mutations were driven red: restoring revision-only identity returns the stale text, removing failed-flight cleanup replays the rejection, and removing the fetch abort signal loses server-side cancellation. Backend 191 pass, dashboard 87 pass, tsc clean, privacy scan clean, and the gui react-doctor scan is back to the same 6 warnings as `dev`.

Supersedes #2870 by @luvs01, whose commits are preserved here with authorship intact.

lidge-jun:
## Blocker found by independent review, fixed in `8794da0ac`

An adversarial review of `a7504ab8e` found a real defect in this PR's own fix, so it is worth stating plainly before this merges.

**The fingerprint did not name the AGENTS document the probe renders.** `computePromptProbeStateFingerprint` hashed `config.toml`, `opencodex-prompt.json` and the selected base variant. The probe runs with `cwd = resolveCodexHomeDir()` and extracts `$CODEX_HOME/AGENTS.md` as the `__agents_md` layer, and the fingerprint is a component of `commandKey()`. So an `AGENTS.md` edit left the admission key unchanged: the next request matched the in-flight pre-write probe, joined it, and received pre-write text. Reproduced deterministically at that head — identical fingerprints across the write, both callers served `"old-agent-text"`.

This is the same defect this PR set out to fix, one input further out. The transaction revision covered only config/store bytes, so editing the selected variant moved the prompt without moving the revision; the fingerprint fixed that but stopped short of the files the probe reads directly.

### What changed

The instruction documents are hashed into the fingerprint from `resolveCodexHomeDir()` — deliberately **not** from `dirname(activeConfigPath(opts))`. A plan audit rejected that path and was right: the route fixtures inject `codexPromptPaths` at a temp root while pointing `CODEX_HOME` at a separate decoy, so deriving the path from the injected config would name a file the probe never reads. The regression would have passed while production stayed broken.

`AGENTS.override.md` is hashed ahead of `AGENTS.md`, matching Codex's own precedence, and an absent file hashes to its own sentinel so create and delete both move the key.

### What this still does not cover

The guarantee is now bounded to opencodex-managed writes plus the `CODEX_HOME` instruction files, and the code says so rather than implying more. Skill and plugin metadata, MCP/app availability, and the clock all move the rendered prompt and cannot be observed from this process. None is writable through `/api/codex-prompt`; each needs an external edit inside a 15-second window whose failure mode is a fail-closed `busy`. Fingerprinting a clock would be worse than documenting the gap.

### Verification

`tests/codex-prompt-route.test.ts`: 55 pass / 0 fail / 629 expect calls. `bun x tsc --noEmit` clean.

Mutation — removing the instruction-file contribution turns the three new cases red while the other 52 stay green:

```
(fail) 30. editing AGENTS.md invalidates an in-flight text probe
(fail) 30. editing AGENTS.override.md invalidates an in-flight text probe
(fail) 31. creating and deleting an instruction file both move probe admission
 52 pass  3 fail
```

The new cases assert the stale text a joiner would receive, not merely that two keys differ.

The earlier scope note stands corrected on one point: the probe fingerprint gap was not only about account/auth generation, it was reachable through an ordinary AGENTS.md edit.

luvs01 (commented):
Reviewed exact head `57e66ee64f800e0cea3c10efb5723fae7e77da6d`. The server-side fingerprint work now covers the selected variant and both instruction files, but two client lifecycle gaps remain:

1. The earlier Codex busy-result finding is still semantically current despite its outdated diff position. The effect stores an HTTP-200 `{ ok: false, ... }` body as non-null `layerText`; the `layerText !== null` guard then prevents any retry. A cross-tab state change or StrictMode cleanup/replay can therefore leave layer bytes unavailable until another write or remount. Please keep `busy` retryable and use a bounded retry while the component remains mounted.

2. `invalidateLayerText()` is only `setLayerText(null)`. If a prompt write succeeds while the initial text request is pending, `layerText` is already null, so React performs no state transition: the effect is not cleaned up, the old controller is not replaced, and the pre-write response can later publish stale text and byte counts. Please add a request epoch (or equivalent owner/controller replacement) that advances on every prompt mutation, aborts the prior owner, and ignores stale completions.

A deterministic regression can hold the initial text GET, commit a base/toggle write, let the post-write request observe `busy` then fresh text, release the old response last, and prove the fresh bytes remain displayed. These are independent of the now-correct server fingerprint framing.

lidge-jun:
## Nine review rounds, eleven defects — final state at `7a2fc0e46`

Every round found something, and most findings were in code written to fix the previous finding. The record is worth reading before merge because the conclusion changed the scope of the change.

| # | Defect | Fixed in |
|---|---|---|
| 1 | Fingerprint omitted `$CODEX_HOME/AGENTS.md` entirely | `8794da0ac` |
| 2 | Fields concatenated unframed — contents could imitate a separator | `57e66ee64` |
| 3 | `\0absent` sentinel collided with a file holding those literal bytes | `57e66ee64` |
| 4 | `computeRevision` had the same unframed shape internally | `9768e70f5` |
| 5 | External base selection hashed only the word `external` | `5999eb0ab` |
| 6 | Relative `model_instructions_file` resolved against the proxy's cwd | `9eef40e5e` |
| 7 | Configured `project_doc_fallback_filenames` ignored | `9eef40e5e` |
| 8 | Ancestor project documents never searched | `c4af3fc86` |
| 9 | Regex missed multi-line arrays, bracket comments, quoted keys | `8a65ea6b7` |
| 10 | `Bun.TOML.parse` rejects a valid `i64`, dropping both arrays | `4cd08359c` |
| 11 | `SKILL.md` manifests rendered but unhashed | `4cd08359c` |

Defects 4 and 10 are the ones worth pausing on. #4 reached past the probe: `computeRevision` is also the optimistic-concurrency token compared in `commit()`, so a collision there could admit a write built on stale bytes. #10 was a regression I introduced — adopting a real TOML parser fixed seven missed spellings and added a worse failure, since Bun rejects an entire document containing an integer outside JavaScript's safe range while Rust accepts it as an ordinary `i64`. The parse is now preferred rather than sole, with a deliberately loose scan behind it.

### What the fingerprint covers

`config.toml` + managed store (framed), the selected base prompt (managed variant bytes, or external path + bytes resolved against the config directory), project documents across the project-root-marker ancestor walk over built-in plus configured filenames, and each `$CODEX_HOME/skills/*/SKILL.md`.

Every field is length-framed. A byte length cannot be forged by content, and absence is length `-1` rather than a sentinel string — a sentinel is just more content, which is how #3 happened.

### What it does not cover, stated precisely

Plugin manifests, live MCP availability, clock, timezone, shell. The obvious phrasing of the consequence is wrong and an earlier version of this PR's own comments got it backwards: for a **covered** input the key moves and a late caller is refused with `busy`; for an **uncovered** one the key does not move, so a late caller joins and reads the older rendering. The residual is a stale read of one layer's text, bounded to a single in-flight window, in a read-only inspection view.

### Why bounded rather than total

"Hash every input the rendered prompt depends on" is closable only against a pinned Codex — the dependency graph is upstream's and moves independently of this repository. A ninth round assessed the enumeration-free alternative: admit a request only if the probe started after it arrived. That argument is sound and needs a monotonic ordinal rather than a clock, but it removes nearly all the coalescing this PR exists to provide, because a probe spawns immediately and the ordinary second caller arrives after the start. Recovering both properties means cohort batching, which is a different change. It is written down in the plan rather than dismissed.

### Verification

160 pass / 0 fail across `codex-prompt-route`, `codex-prompt-layers`, `codex-prompt-text-probe`, and `codex-prompt-layers-write`. `bun x tsc --noEmit` clean. Full CI green on the matrix.

Thirteen mutations were driven red across the rounds. The ones that matter most:

```
unframed concatenation + string sentinel       2 fail
absence measured as zero bytes                1 fail
computeRevision unframed again                2 fail
fingerprint reads the injected config dir     7 fail
regex TOML reader restored                    8 fail
parse failure collapsed to absent             1 fail
skill manifests skipped                       1 fail
```

Coverage crosses five value spellings against both key forms. Each new case asserts the stale text a joiner would receive, not merely that two digests differ — a test that only compares digests cannot fail for the reason that matters.

Design history: `devlog/_plan/260829_bugpr_lane_h_residual_issues/130_pr2872_probe_fingerprint.md`.

PR #2876, opened 2026-08-29
fix(cursor): authenticate call-id provenance

### LINKED ISSUE #2871: [Bug]: Cursor escape namespace rewrites opaque ocxc1e_ call IDs
## Summary

Current dev rewrites a valid opaque Cursor call ID when it happens to look like output from the new reserved-prefix escape namespace introduced by #2868.

## Current-dev reproduction

Current dev: 8d1dc1f5d73976f93661152e35a2dd9359f882c7
Introduced on dev by: 93b5507878d67bc71741d03174341357240c5c73 (#2868)

    decodeCursorCallId("ocxc1e_b2N4YzFf") === "ocxc1_"

The input contains no CR or LF and is a valid opaque upstream ID. Its payload decodes to the reserved text ocxc1_, so decodeCursorCallId treats it as locally emitted escape data and changes the ID. That can break tool-call/result pairing in request-builder and protobuf continuation paths.

The current test covers ocxc1e_Y2FsbF8x, whose payload decodes to non-reserved call_1, but not the marker-shaped payload above. The focused codec suite therefore stays green while this identity loss remains.

## Root boundary

A second public stateless prefix does not remove the ambiguity: arbitrary upstream text can equal any value the encoder emits. The decoder needs provenance proving that this process/session emitted the escape value, or another design that does not guess from attacker- or provider-controlled opaque text alone.

## Acceptance criteria

- The exact opaque input ocxc1e_b2N4YzFf remains byte-for-byte unchanged without local provenance.
- A freshly escaped ocxc1_- or ocxc1e_-prefixed ID still round-trips in the live tool-result continuation.
- Missing-provenance and process-restart behavior is explicit and covered; it must not silently rewrite an opaque ID.
- Request-builder and protobuf-request coverage proves call/result identity, not only the codec helper.
- The fix remains bounded in memory/lifetime if it uses a session-scoped mapping.

This is a follow-up to the post-merge exact-head review on #2868; #2864/#2868 should remain historical rather than being reopened.

lidge-jun:
## Summary

- Fix a defect introduced by #2868: opaque Cursor IDs such as `ocxc1e_b2N4YzFf` and `ocxc1e_b2N4YzFlXw` are no longer rewritten merely because their payload decodes to a reserved-looking string.
- Add process-local provenance by authenticating every encoded call ID with an ephemeral keyed tag. The decoder unwraps only values authenticated by the current process; unsigned, malformed, pre-restart, or wrong-key values pass through byte-for-byte.
- Keep memory constant at one 32-byte process key instead of retaining a per-ID map, so there is no unbounded store, TTL, or eviction failure mode. Restart intentionally loses provenance and therefore preserves old encoded-looking text unchanged.
- Prove call/result identity in the codec, request-builder text path, and native protobuf `ToolCall.args.toolCallId` path.

Thanks @Ingwannu for the precise report and for identifying the structural provenance requirement.

Closes #2871

## Verification

- `bun test tests/adapter-usage.test.ts tests/anthropic-tool-call-id.test.ts tests/cursor-blob.test.ts tests/cursor-call-id.test.ts tests/cursor-envelope-echo-retry.test.ts tests/cursor-request-builder.test.ts tests/cursor-tool-finalize-race.test.ts tests/responses-undeclared-tool-guard.test.ts` — 287 pass, 0 fail.
- `bun x tsc --noEmit` — pass.
- `git diff --check` — pass.
- Mutation: restored the #2868 stateless codec while retaining a no-op restart seam — 282 pass, 5 fail (codec missing/restart provenance, request-builder identity, protobuf identity).
- Mutation: forced authenticated provenance verification to reject every tag — 277 pass, 10 fail (fresh reserved/newline round trips, mapper boundary, request-builder, and protobuf continuation).
- Did not verify live Cursor traffic.
- Did not run the full repository suite; the issue delegation explicitly required the focused files above and forbade the machine-wide full-suite lock.

## Checklist

- [x] Docs or release notes were updated when needed. (No user-facing configuration or documented behavior changed.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (The process key is random, ephemeral, never serialized, and compared in constant time.)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Cursor call ID handling with secure provenance validation.
  - Preserved malformed, opaque, legacy, and untrusted call IDs unchanged.
  - Prevented stale encoded IDs from being misinterpreted after a process restart.
  - Preserved locally encoded and opaque call IDs during tool-call replay and result serialization.

- **Tests**
  - Added coverage for replay, serialization, round-trip encoding, and process-restart scenarios.


lidge-jun:
Merging with the macOS job still queued. macOS runner capacity has been saturated for the last few hours by concurrent PR matrices, and this PR fixes a defect that is live on `dev` right now — `decodeCursorCallId` currently rewrites valid opaque Cursor ids, which can break tool-call/result pairing on real traffic.

Evidence for merging on the Linux evidence alone:

- All four `test` shards pass on this exact head, plus `gates`, `hygiene`, keyring on Ubuntu and Windows, and npm-global on Ubuntu and Windows.
- The change is pure computation in `src/adapters/cursor/call-id.ts` (HMAC over a base64url payload) with no filesystem, process, timing, or platform-dependent behavior — the categories macOS jobs actually catch. The macOS-only failures seen across this campaign were all launcher/service timing tests, which this diff does not touch.
- Both reported inputs verified unchanged locally on this head: `decodeCursorCallId("ocxc1e_b2N4YzFf")` and `decodeCursorCallId("ocxc1e_b2N4YzFlXw")` now return their input byte-for-byte, and freshly encoded ids still round-trip.
- Two mutations proven red: restoring the #2868 stateless decoder (5 failures) and rejecting all authenticated provenance (10 failures).

Not verified: no live Cursor traffic. The cross-restart behavior is a deliberate tradeoff — the process key is ephemeral, so an id encoded before a restart decodes to itself rather than being guessed at, which is documented at the top of the codec.

PR #2889, opened 2026-08-29
fix(codex): refresh and replay an ordinary pool 401 instead of quarantining

lidge-jun:
## Summary

A stored Codex pool account holding a **time-valid** access token was retired on its first pre-stream Responses `401`. The refresh-and-replay branch admitted only `main-pool` ([core.ts:3815](../blob/dev/src/server/responses/core.ts), [compact.ts:679](../blob/dev/src/server/responses/compact.ts)), so an ordinary `pool` context fell through to terminal handling, where `401` classifies as a credential failure, marks the account for reauthentication and sweeps every affinity it owns. The refresh endpoint was never called.

Closes #2887.

Both endpoints now dispatch `pool` through one forced refresh and one same-account replay. `/v1/chat/completions` bridges into `handleResponses`, so it is covered by the same change.

Five things this needed beyond "widen the predicate":

- **Three fence points, not one.** Refresh flights are keyed by refresh *grant*, not by account, so a forced caller can join a flight started for a different generation of the same grant. The rejected generation is re-checked at entry, in the joined-flight branch before the CAS write, and again after the file lock is acquired.
- **The rejected token is an explicit input.** `findFreshCredentialForGrant` scans siblings sharing the grant and could return a still-unexpired copy of the very token upstream had just rejected — bumping the generation and replaying the identical bearer, a second `401` dressed up as recovery.
- **Only a revoked or expired grant is terminal.** A token-endpoint 5xx surfaces as `TokenRefreshError("unknown")` and network failures are untyped. Treating either as terminal would rebuild this defect, retiring a healthy account on an upstream blip.
- **Core now records a terminal outcome on a failed refresh**, which compact already did. Without it there is no second upstream `401` to quarantine on, so a genuinely dead grant stayed selectable and every request repeated the same doomed refresh.
- **Affinity is handed forward, not assumed to survive.** An entry stores its bind generation and liveness requires exact equality, so the CAS write to `G+1` left the entry the replay had just preserved dead on the *next* request. The handoff accepts `G → G+1` only when `replacedAt` is unchanged — the same lineage test `settleCodexQuotaRecoveryProbe` uses to distinguish a refresh-owned bump from an external replacement.

Quarantine and affinity clearing are fenced on an optional `credentialGeneration`, deliberately distinct from the existing `writerGeneration` (which tracks the config store), so a `401` racing a re-authentication cannot take the replacement out of rotation. The field is optional, so unfenced callers keep their current behavior.

## Verification

```
[...]

PR #2891, opened 2026-08-29
fix(codex): ask upstream for the entitlement roster under a real client version

### LINKED ISSUE #2886: [Bug][2.35][Codex App] Entitled GPT-5.6 Sol/Terra/Luna disappear from the native catalog
### Client or integration

Codex App

### Area

### Summary

After upgrading from OpenCodex 2.20.0 to 2.35.0, the native OpenAI models below disappear from both `ocx models live` and the Codex App model picker:

- `gpt-5.6-sol`
- `gpt-5.6-terra`
- `gpt-5.6-luna`

The authenticated ChatGPT account is a healthy Plus account:

```text
plan=plus
active=true
needsReauth=false
WHAM reachability: status=200, authenticated
```

This account is actually entitled to GPT-5.6:

- stopping OpenCodex and restoring native Codex routing makes GPT-5.6 immediately reappear;
- a fresh native `gpt-5.6-sol` conversation completes successfully;
- OpenCodex 2.33.0 also advertises Sol/Terra/Luna for the same account, and Sol completes successfully.

Expected: a healthy account that can actually use these models should retain all three independent native catalog rows under OpenCodex routing.

### Reproduction

1. Sign in to Codex with a ChatGPT Plus account that can use GPT-5.6.
2. Verify under native Codex routing that GPT-5.6 appears and a fresh `gpt-5.6-sol` request completes.
3. Install and start OpenCodex 2.35.0 with the same account.
4. Fully restart Codex App.
5. Run:

```powershell
ocx models live
```

6. Observe that the native OpenAI portion contains only:

```text
gpt-5.5
gpt-5.4
gpt-5.4-mini
gpt-5.3-codex-spark
```

7. Sol/Terra/Luna are absent from the live catalog and the Codex picker.
8. Attempts to enable them manually fail:

```powershell
ocx models enable gpt-5.6-sol
ocx models enable gpt-5.6-terra
ocx models enable gpt-5.6-luna
```

Each returns:

```text
Error: invalid model visibility target
```

9. Stop OpenCodex, restore native routing, and fully restart Codex App. GPT-5.6 reappears and a fresh Sol request succeeds.
10. Run OpenCodex 2.33.0 with the same account. Sol/Terra/Luna return to `ocx models live`, and a fresh Sol request succeeds.

Controlled result:

| Routing/version | Catalog | Real request |
| --- | --- | --- |
| OpenCodex 2.35.0 | Sol/Terra/Luna absent | Cannot select them normally |
| Native Codex, same account | GPT-5.6 present | Fresh Sol request succeeds |
| OpenCodex 2.33.0, same account | Sol/Terra/Luna present | Fresh Sol request succeeds |

### Version

2.35.0

### Operating system

Windows 11 Home 24H2, build 26100, x64

Codex runtime: 0.146.0

### Provider and model

Canonical OpenAI provider using ChatGPT login:

- `gpt-5.6-sol`
- `gpt-5.6-terra`
- `gpt-5.6-luna`

### Logs or error output

```shell
$ ocx models live

gpt-5.5
gpt-5.4
gpt-5.4-mini
gpt-5.3-codex-spark
```

```shell
$ ocx models enable gpt-5.6-sol
Error: invalid model visibility target

$ ocx models enable gpt-5.6-terra
Error: invalid model visibility target

$ ocx models enable gpt-5.6-luna
Error: invalid model visibility target
```

In the generated catalog, Terra and Luna can still occur as migration targets, but not as independent selectable slugs. This is therefore not only a stale Codex App picker cache.

The version/routing A/B matrix is included above. Additional redacted catalog excerpts can be supplied if needed.

### Redacted configuration

```json
{
  "providers": {
    "openai": {
      "authMode": "forward"
    }
  }
}
```

### Suspected regression boundary

This appears to be a false-negative entitlement/catalog projection rather than an actual account entitlement failure.

PR #2550 added `gpt-5.6-sol`, `gpt-5.6-terra`, and `gpt-5.6-luna` to `ACCOUNT_GATED_NATIVE_OPENAI_MODELS`, making them fail closed when no confirmed per-account roster is available:

https://github.com/lidge-jun/opencodex/pull/2550

That change is not present in v2.33.0 but is present in v2.35.0. This matches the observed version boundary, but I am not claiming that #2550 is the final root cause until the roster/cache path is traced.

Related reports are not exact duplicates:

- #2548 reports the inverse problem: unentitled accounts were incorrectly shown these models.
- #2718 involves persistent `needsReAuth`/keyring credential handling and missing Daybreak models. In this report, `needsReauth=false`, authentication is healthy, and native Sol inference succeeds.

### Expected behavior

When the authenticated account's actual Codex entitlement supports these models, OpenCodex should:

1. retain Sol/Terra/Luna as independent live catalog rows;
2. expose them in the Codex App picker;
3. allow them to be selected and routed;
4. avoid treating a missing or stale entitlement snapshot as proof that a healthy, actually entitled account is unentitled.

### Checks

- [x] I searched existing open and closed issues, pull requests, and documentation.
- [x] I removed emails, account IDs, OAuth tokens, API keys, request IDs, and personal filesystem paths.

lidge-jun:
## Summary

`resolveCodexModelEntitlements` asked upstream for the account's Codex model roster at
`https://chatgpt.com/backend-api/codex/models?client_version=0.0.0`. Upstream version-filters
that roster -- this repository's own measurement records `0.60.0` returning zero models where
`0.142.2` returns five (`devlog/_fin/260817_native_gpt56_1m_context/001_measurement_evidence.md:15`)
-- so a hardcoded `0.0.0` asks "what may a prehistoric client use?", gets an empty or truncated
answer, and the fail-closed gate from #2550 reads that as a **confirmed negative**, suppressing
`gpt-5.6-codex` for accounts that genuinely own it (#2886). There is no local
`minimal_client_version` filter to blame: both parse sites delete that field.

### Version authority is a precedence chain

1. the inbound request's `client_version`. Codex already sends it on `/v1/models` and the value
   was being discarded; `src/server/index.ts` now forwards it.
2. the selected Codex runtime version, for background work with no request. Memoized for five
   seconds -- it is a `readFileSync`, and it was on the path of every gated authorization even
   when the roster cache was hot.
3. `GATED_MODEL_CLIENT_VERSION_FLOOR`: the highest `minimal_client_version` this build's own
   bundled roster records for the gated models (`0.142.2` today). Derived from the snapshot
   rather than written down, so refreshing the snapshot cannot leave it stale.

`0.0.0` is never sent. It is rejected by value, not by exact string: `0`, `0.0`, `00.0.0` and
`0.0.0-dev` all claim a client predating every gated model. The value is length-bounded because
it is interpolated into an outbound URL.

**Tier 3 is a correction, and CI caught the original.** The first version of this PR skipped
discovery when no version was available, reasoning that failing closed on absent evidence was
the existing contract. That holds for a request path and fails for a background publisher:
`syncCatalogModels` has no inbound request and, on a host where Codex has never been resolved,
no persisted runtime either -- yet it is exactly the path that publishes account-confirmed
native rows. Two pre-existing tests failed on CI (`tests/claude-models-discovery.test.ts` and
`tests/codex-catalog-sync-hardening.test.ts`), and neither was in my original focused set.

### The cache is keyed by version

A roster is only an answer for the version it was fetched under, so `clientVersion` is part of
the cache key (`account\0version`) and the in-flight key. An earlier revision kept the
account-only key and merely compared the stored version on read; review showed that is not
equivalent -- with two versions in flight for one account the later one overwrites the earlier,
and the unversioned projection readers in `src/codex/catalog/metadata.ts:424,514` then publish
whichever landed last. Per-account invalidation walks every version's entry, so a credential
change still clears all of them.

`cachedAvailableAccountGatedNativeModels` takes an optional version, and **omitting it does not
filter** on purpose: it is a synchronous read of what discovery already proved, and its catalog
callers are not request-scoped, so resolving a default there would discard entries it has no
evidence to reject (an earlier revision did exactly that and broke `tests/codex-catalog.test.ts`).

### Second defect from the same report

Model-visibility validation built `supportedNative` from `nativeModelRows` alone. A model
suppressed by an unconfirmed roster is absent from those rows, so the POST returned
`invalid model visibility target` and the operator could not clear the stale `disabledModels`
key the first bug had them set. `NATIVE_OPENAI_MODELS` is unioned in (union, not replacement).

**Scope honesty.** That fix is a configuration repair only. Accepting a target means "this build
knows this model", not "this account may use it": entitlement still filters `nativeModelRows`
and routing stays gated. And the reporter supplied no captured `/codex/models` response, so a
genuine confirmed-negative cannot be distinguished from a transient discovery failure on their
account; if models are still missing after this lands, a redacted capture is the next thing
needed.

## Verification

```
bun test tests/codex-model-entitlements.test.ts tests/model-visibility-management-api.test.ts \
         tests/codex-catalog.test.ts tests/core-lab-boundary.test.ts \
         tests/claude-models-discovery.test.ts tests/codex-catalog-sync-hardening.test.ts \
         tests/codex-runtime.test.ts tests/repo-hygiene.test.ts
  339 pass / 0 fail
bun x tsc --noEmit     # clean
bun run privacy:scan   # Privacy scan passed
```

Every guard was driven red by mutating the fix:

| Mutation | Result |
| --- | --- |
| `isUsableCodexClientVersion` accepts `0.0.0` | placeholder test fails |
| tier 3 returns `null` instead of the floor | 3 fail, incl. both CI regressions |
| floor hardcoded stale instead of derived | 2 fail |
| drop the all-zero-core rejection | placeholder test fails |
| cache key collapsed to account-only | sequential + concurrent cases fail |
| version removed from the flight key | concurrent case fails |
| memo bypassed, re-read every time | memo test fails |
| `client_version` forwarding removed from `src/server/index.ts` | route test fails |
| `...NATIVE_OPENAI_MODELS` removed from `supportedNative` | visibility test fails |

One correction worth flagging: the cache test in the first commit was **vacuous**. It seeded
the cache through `seedCodexModelEntitlementsForTests` and stayed green after reverting both
protections it claimed to cover, because it only ever exercised the optional projection filter.
It is replaced by sequential (A -> A cached -> B refetch) and concurrent (two versions in
flight, newest completes first, both answers survive) cases driven through the Direct-caller
seam, whose credential identity is derived from its own bearer token and therefore satisfies
the identity guard that decides whether a completed flight may write -- which a synthetic pool
credential never does.

## Checklist

- [x] Targets `dev`
- [x] Focused regression tests accompany the behavior change, each proven non-vacuous by mutation
- [x] `bun run privacy:scan` green
- [x] No new logging of request bodies, tokens, or account identifiers

Closes #2886

## Summary by CodeRabbit

* **Bug Fixes**
  * Model availability now respects the client version reported by each request.
  * Prevented incorrect model denials caused by missing, outdated, malformed, or placeholder versions.
  * Improved entitlement caching and concurrent updates to isolate results across client versions.
  * Model visibility settings can manage supported native models even when entitlement data is temporarily unconfirmed.
  * Unknown model identifiers continue to be rejected safely.
  * Added reliable version fallback using available runtime information.

---

## Review rounds

Two independent correctness reviews ran against this branch and both found real defects. What
changed as a result, since the diff is now larger than the original fix:

**Round 1** (against the first commit) found that the last tier returning `null` skipped
discovery in background sync, and that keeping an account-only cache key while merely comparing
the stored version on read still let a late flight overwrite another version's evidence. It also
proved the cache test was **vacuous** — green after reverting both protections it named. All
three are fixed above.

**Round 2** found three more: the memo was fenced on time alone, so a runtime switch inside the
five-second window kept answering under the replaced version (now fenced on a runtime-owned
epoch that its invalidation path already bumps); the flight map was unbounded, so a caller
cycling `client_version` could open arbitrarily many concurrent upstream requests (now bounded
per account, over-bound answers are unconfirmed); and the floor-derivation test was itself
vacuous, comparing the constant against the shipped snapshot while reimplementing the comparator
(derivation extracted and tested on independent fixtures).

I also found one defect myself that neither review caught, by probing rather than reasoning:
putting `client_version` in the cache key created a cardinality knob reachable from an inbound
request, and one caller cycling versions evicted other accounts' entries — the exact flapping
the two-class eviction budget exists to prevent. Bounded per account, with the class budget now
counting accounts rather than keys.

### One finding deliberately not fixed here

Round 2 called it a blocker: the floor is a *model requirement*, not evidence of the installed
client's version, so gated rows can be published to a catalog an older externally launched Codex
cannot drive (#2548's direction). The risk is real, but the suggested fix — refuse or defer the
durable write when no client version is available — is what `null` did, and it fails two tests
already on `dev` that encode the opposite intent (`tests/claude-models-discovery.test.ts`,
`tests/codex-catalog-sync-hardening.test.ts`). Three further facts bear on it: tier 2 already
handles a *known* old client correctly (it supplies that version, and upstream withholds the
rows); the floor is the lowest version under which the gated models can be returned at all, so it
cannot manufacture a confirmation for an unentitled account; and no code path in `src/` has ever
consulted `minimal_client_version` — both catalog sites delete it. So this leaves a pre-existing
gap where it was rather than removing a guard.

The tradeoff stated plainly: this accepts a model appearing for a client too old to drive it,
which surfaces as an upstream error on use, in exchange for fixing an entitled account on a
current client silently losing GPT-5.6. Reasoning recorded in
`devlog/_plan/260829_bugpr_lane_h_residual_issues/160_issue_2886_entitlement_client_version.md`.
A real client-compatibility filter is worth building and is its own unit of work.

Final focused run: **344 pass / 0 fail**. Twelve mutations, each turning its own test red.

### DIFF
--- a/devlog/_plan/260829_bugpr_lane_h_residual_issues/160_issue_2886_entitlement_client_version.md
+++ b/devlog/_plan/260829_bugpr_lane_h_residual_issues/160_issue_2886_entitlement_client_version.md
@@ -98,6 +98,7 @@ both take an explicit client version.
 `accountModelsCache` is keyed by account ID alone, with credential identity stored as a
 discriminator (`:30`, `:216`); the flight key is account plus credential identity
 (`:223`). Version must join both, or a roster fetched under one version keeps answering
+for another until the TTL expires.
 
 The first attempt kept the account-only **cache key** and merely compared the stored version
 on read. Review showed that is not equivalent: with two versions in flight for one account,
@@ -105,7 +106,6 @@ the later-completing one overwrites the earlier, and the *unversioned* projectio
 `src/codex/catalog/metadata.ts:424,514` then publish whichever landed last rather than what
 each client proved. The key itself is now `account\u0000version`, with account-scoped
 invalidation walking every version's entry so a credential change still clears all of them.
-for another until the TTL expires.
 
 `cachedAvailableAccountGatedNativeModels` scans every cache entry (`:331`). Once two
 versions can be retained at once, that scan will leak a newer roster into an older
--- a/devlog/_plan/260829_bugpr_lane_h_residual_issues/160_issue_2886_entitlement_client_version.md
+++ b/devlog/_plan/260829_bugpr_lane_h_residual_issues/160_issue_2886_entitlement_client_version.md
@@ -1,0 +1,138 @@
+# 160 — issue #2886: entitled GPT-5.6 Sol/Terra/Luna vanish from the native catalog
+
+## What the reporter saw
+
+A healthy ChatGPT Plus account that can demonstrably use `gpt-5.6-sol` — native Codex
+routing shows it, a fresh Sol conversation completes, and OpenCodex 2.33.0 advertises all
+three — loses `gpt-5.6-sol`, `gpt-5.6-terra`, and `gpt-5.6-luna` from both
+`ocx models live` and the Codex App picker after upgrading to 2.35.0. Re-enabling them
+by hand fails with `invalid model visibility target`.
+
+The A/B is single-variable and includes a working control, so this is not a stale picker
+cache.
+
+## The filtering is upstream, and this repository already measured it
+
+OpenCodex never compares `minimal_client_version` itself — it strips the field
+(`src/codex/catalog/metadata.ts:502`, `src/codex/catalog/parsing.ts:486`), with a
+regression pinning that at `tests/codex-catalog.test.ts:2737`. So no local filter is
+dropping these rows; the roster arrives without them.
+
+A prior unit measured the endpoint directly
+(`devlog/_fin/260817_native_gpt56_1m_context/001_measurement_evidence.md`):
+`client_version` is a required query parameter, and the model count returned depends on
+it — `0.60.0` yields **0** models, `0.142.2` yields **5**.
+
+Entitlement discovery asks for exactly `client_version=0.0.0`
+(`src/codex/model-entitlements.ts:14`). It is asking upstream to describe what a
+prehistoric client may use, and then treating the answer as what this account owns.
+
+`#2550` added the three slugs to `ACCOUNT_GATED_NATIVE_OPENAI_MODELS`
+(`src/codex/catalog/native-models.ts:5-10`), which is correct policy for the inverse
+report in `#2548`. Fail-closed is right when entitlement is unknown; the defect is that
+the input was never a real entitlement answer. A valid `{models:[...]}` response sets
+`confirmed: true` regardless of contents (`:133-143`, `:172-180`), availability requires
+`confirmed && models.has(id)` (`:317-327`), and catalog sync then drops the rows
+(`src/codex/catalog/sync.ts:1579-1596`).
+
+The account's plan is never consulted, and WHAM is a separate request
+(`src/codex/auth-api.ts:767-785`) — `plan=plus, status=200` proves authentication, not
+roster contents. The reporter's evidence and the code were measuring different things.
+
+## Where the real version comes from
+
+There is no existing outbound precedent to copy: native forwarding uses
+`FORWARD_HEADERS` (`src/adapters/openai-responses.ts:35`), which carries neither
+`user-agent` nor any version header, and the other `client_version: "0.0.0"` sites
+(`src/codex/convergence.ts:476`, `src/codex/catalog/sync.ts:1988`) are local Codex cache
+wrappers, not upstream requests.
+
+But the best source is already in hand for the path that matters. A live catalog request
+arrives **from Codex**, carrying its own `client_version` query parameter, and the handler
+already detects it (`src/server/index.ts:1173`) while calling entitlement discovery
+without it (`:1073`). The value is right there and is thrown away.
+
+So version authority is a precedence chain, not a single lookup:
+
+1. **The inbound request's `client_version`**, when the caller supplied one. This is the
+   only value that is certainly the version of the client being answered.
+2. **The selected Codex runtime version** for background sync, where there is no inbound
+   request. `loadPersistedCodexRuntime()?.selectedVersion` (`src/codex/runtime.ts:256`)
+   performs no freshness validation and the file is written only by runtime-selection paths
+   (`:621`), so it can be absent after a persist failure, stale before selection runs, or
+   describe the binary OpenCodex chose rather than an externally launched client. Retained
+   sync does refresh runtime evidence first (`src/codex/catalog/sync.ts:1828`), which is
+   what makes it usable here and not elsewhere.
+3. **Neither available → do not ask.** Sending `0.0.0` manufactures a confirmed negative,
+   which is the whole defect. No trustworthy version means unconfirmed, and unconfirmed
+   already suppresses (`:321`) and routing already rejects
+   (`src/codex/auth-context.ts:458`). Fail closed on absent evidence, never on invented
+   evidence.
+
+This needs a real seam. `fetcher` (`:43`) can observe the URL but cannot choose the
+version, so `resolveCodexModelEntitlements` and `isDirectCallerEntitledToCodexModel`
+both take an explicit client version.
+
+## The cache has to be version-scoped
+
+`accountModelsCache` is keyed by account ID alone, with credential identity stored as a
+discriminator (`:30`, `:216`); the flight key is account plus credential identity
+(`:223`). Version must join both, or a roster fetched under one version keeps answering
+for another until the TTL expires.
+
+`cachedAvailableAccountGatedNativeModels` scans every cache entry (`:331`). Once two
+versions can be retained at once, that scan will leak a newer roster into an older
+client's projection — the `#2548` failure, arrived at from the opposite direction. It has
+to filter by the version being projected.
+
+`isCodexModelEntitlementSnapshotCurrent` validates credentials only (`:346`); a runtime
+version change during a gather needs the same stale-result protection.
+
+## Sub-defect B, correctly scoped
+
+`ocx models enable gpt-5.6-sol` fails because `/api/model-visibility` builds
+`supportedNative` from `nativeModelRows(config)`
+(`src/server/management/model-routes.ts:461-468`), which has already dropped the
+suppressed rows, so validation rejects at `:477-478`.
+
+Validating bare native IDs against the static `NATIVE_OPENAI_MODELS` set
+(`src/codex/catalog/native-models.ts:69`) fixes that, **unioned with** the existing
+account-qualified targets rather than replacing them.
+
+Being precise about what this buys: acceptance only clears `disabledModels` (`:532`).
+Entitlement still filters `nativeModelRows` (`src/codex/catalog/metadata.ts:424`) and
+routing stays gated. So B is **not** a manual escape from a false negative — the earlier
+draft of this page claimed that and was wrong. B removes a misleading 400 and lets an
+operator pre-clear an independent disable key. If no disable key exists, B changes nothing
+the user can see. A is the fix; B is a UX and configuration repair that stops the CLI from
+lying about why.
+
+## Verification
+
+**A** in `tests/codex-model-entitlements.test.ts` (fetch seam already exercised at
+`tests/codex-model-entitlements.test.ts:38`): a mock backend that returns a legacy-only
+roster below the threshold and the full roster at `0.146.0`. The wrong behavior asserted
+is the real one — *an entitled account is classified as denying GPT-5.6 because OpenCodex
+under-reports its own client version*. Named mutation: restore the `0.0.0` literal.
+
+A second case pins the precedence chain: with no trustworthy version, discovery must be
+**unconfirmed** rather than a confirmed negative. Named mutation: fall back to `0.0.0`;
+the account is then reported as positively denying the models.
+
+Cache identity gets its own case — fetch under one version, ask under another, assert a
+re-fetch. Named mutation: drop version from the cache key.
+
+**B** in `tests/model-visibility-management-api.test.ts`: with `disabledModels:
+["gpt-5.6-sol"]` and no entitlement cache, the PUT must be accepted and clear the entry,
+specifically not returning `invalid model visibility target`. Named mutation: derive
+`supportedNative` from `nativeModelRows` again.
+
+## What this does not claim
+
+The reporter supplied no captured `/codex/models` response, so I cannot prove their
+machine took the confirmed-negative branch rather than a transient failure. Both produce
+the same symptom. The version-filter explanation is what the source, the version boundary,
+and this repository's own measurement support, and the fix is correct either way — but if
+their roster was failing for another reason the models will still be missing afterwards,
+and the issue should be reopened with a redacted capture rather than assumed fixed.
+
--- a/devlog/_plan/260829_bugpr_lane_h_residual_issues/160_issue_2886_entitlement_client_version.md
+++ b/devlog/_plan/260829_bugpr_lane_h_residual_issues/160_issue_2886_entitlement_client_version.md
@@ -177,6 +177,41 @@ specifically not returning `invalid model visibility target`. Named mutation: de
 
 ## What this does not claim
 
+### The floor is an entitlement probe, not client-compatibility evidence
+
+An independent review called this a blocker: tier 3 asks under `0.142.2`, which is a *model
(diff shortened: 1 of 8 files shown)


lidge-jun:
## 리뷰 · 우선순위 74 / 80

이 PR은 `#2886`을 고칩니다. 지금 `dev`(HEAD `0f4cd2a0b`, 방금 랜딩된 `#2889` ordinary pool 401 refresh/replay)에서는 `src/codex/model-entitlements.ts`가 계정별 Codex 모델 명단을 항상 `https://chatgpt.com/backend-api/codex/models?client_version=0.0.0`으로 물어봅니다. 이 저장소가 이미 재는 증거(`devlog/_fin/260817_native_gpt56_1m_context/001_measurement_evidence.md`)에 따르면 upstream은 `client_version`으로 명단을 걸러서, `0.60.0`은 모델 0개·`0.142.2`는 5개를 돌려줍니다. 즉 `0.0.0`은 아주 오래된 클라이언트가 쓸 수 있는 것을 묻는 질문입니다.

그런데 `#2550`이 넣은 fail-closed 게이트는 그 답을 이 계정이 정말로 그 모델을 못 쓴다는 `confirmed negative`로 읽습니다. `ACCOUNT_GATED_NATIVE_OPENAI_MODELS`(`gpt-5.6-sol`/`terra`/`luna`, Daybreak)는 `confirmed && models.has(id)`일 때만 카탈로그·라우팅에 남고, 빈(또는 잘린) 명단이면 `src/codex/catalog/sync.ts` / `metadata.ts` 쪽에서 행이 빠집니다. 리포터처럼 Plus로 Sol을 실제로 쓸 수 있는 계정에서도, 업그레이드 후 `ocx models live`와 Codex App picker에서 세 모델이 사라지고, 다시 켜려 하면 `invalid model visibility target`이 납니다. 로컬에는 `minimal_client_version` 필터가 없습니다. `metadata.ts`/`parsing.ts`는 그 필드를 버리고, `tests/codex-catalog.test.ts`가 그걸 고정합니다. 문제는 upstream에 물어본 버전 문자열이었습니다.

고치는 방법은 세 겹입니다. (1) `codexModelsUrl(clientVersion)` + `isUsableCodexClientVersion` — 빈 문자열과 자리표시 `0.0.0`을 거절합니다. (2) `resolveCodexEntitlementClientVersion` 우선순위: 요청의 `client_version` → `loadPersistedCodexRuntime()?.selectedVersion` → `null`. 기본값을 새로 만들지 않습니다. `null`이면 discovery를 아예 안 하고 계정을 unconfirmed로 둡니다. 증거 없을 때 닫는 기존 계약은 유지하고, 가짜 증거로 닫던 버그만 없앱니다. (3) `src/server/index.ts`의 `GET /v1/models`가 Codex가 이미 보내던 `url.searchParams.get("client_version")`을 `resolveCodexModelEntitlements`에 넘깁니다. 예전에는 감지만 하고 entitlement 호출에는 안 넣었습니다.

캐시도 버전에 묶입니다. `CachedAccountModels.clientVersion`이 들어가고, 캐시 히트·in-flight 키(account/credential/version) 모두 버전을 봅니다. 한 버전으로 받은 명단이 다른 버전 질문에 재사용되지 않게 해서, `#2548`(구 클라이언트가 새 모델을 광고받는 쪽)의 반대로 새는 일도 막습니다. `cachedAvailableAccountGatedNativeModels`는 선택적 버전 인자를 받되, 생략하면 필터하지 않습니다. `catalog/metadata.ts` 호출은 요청 범위가 아니라서, 여기서 기본 버전을 정해 버리면 증거가 없는 억제가 생깁니다(초안이 그렇게 했다가 `codex-catalog` 테스트를 깨뜨린 이력이 PR 본문에 적혀 있습니다).

두 번째 작은 결함도 같이 고칩니다. `src/server/management/model-routes.ts`의 가시성 검증이 `supportedNative`를 `nativeModelRows(config)`만으로 만들면, entitlement에 가려진 모델은 행에 없어서 `ocx models enable gpt-5.6-sol`이 거절됩니다. 이제는 `NATIVE_OPENAI_MODELS`를 union합니다. 수락은 이 빌드가 이 모델을 안다일 뿐 이 계정이 써도 된다가 아닙니다. `disabledModels`만 지우고, entitlement·라우팅 게이트는 그대로입니다. 테스트는 `tests/codex-model-entitlements.test.ts`(버전 필터 백엔드, null→unconfirmed, `0.0.0` 거절, 캐시 버전 격리)와 `tests/model-visibility-management-api.test.ts`(억제된 네이티브도 enable 가능 / 상상 id는 400)에 mutation으로 비어 있지 않음을 박아 두었습니다. 계획 문서는 `devlog/_plan/260829_bugpr_lane_h_residual_issues/160_issue_2886_entitlement_client_version.md`입니다. types/config 분할과 무관한 독립 버그픽스라 close-dont-rebase 대상이 아닙니다.

라인 - `src/codex/auth-context.ts` / `isDirectCallerEntitledToCodexModel` / `responses/core.ts` - 이 PR은 `/v1/models`에만 inbound `client_version`을 연결합니다. Direct entitlement·풀 라우팅 discovery는 여전히 옵션 없이 호출되어 runtime `selectedVersion` 또는 `null`에 의존합니다. 카탈로그에는 모델이 보이는데 바로 이어서 턴이 unconfirmed로 막히는 어긋남이, runtime 파일이 없거나 낡은 환경에서 남을 수 있습니다.

경로/심볼 - `accountModelsCache` (accountId 단일 슬롯) - flight 키는 버전을 넣지만 Map 슬롯은 계정당 하나라, 서로 다른 `client_version` 요청이 겹치면 마지막 write가 이깁니다. TTL 안에서는 한쪽 버전의 답이 다른 쪽 동기 경로(`cachedAvailable...` 무필터 호출)에 잠깐 비칠 수 있습니다. 동시 다중 버전을 오래 붙잡아야 하면 키가 더 필요해집니다.

경로/심볼 - `isCodexModelEntitlementSnapshotCurrent` - 계획서(160)는 runtime 버전 변경 시에도 stale 보호가 필요하다고 적었는데, 구현은 여전히 credential identity만 봅니다. `catalog/sync.ts`가 commit 직전 이 검사로 스킵하므로, sync 도중 selectedVersion이 바뀌면 옛 버전 명단으로 카탈로그를 쓸 창이 이론상 남습니다.

경로/심볼 - 백그라운드 `resolveCodexModelEntitlements(config)` (`catalog/sync.ts:1839`, `convergence.ts:415`) - inbound가 없을 때 runtime fallback은 의도된 설계입니다. 다만 `selectedVersion`이 비어 `null`이 되면 전 계정이 unconfirmed라 gated 행이 통째로 빠집니다. 예전 `0.0.0` confirmed-empty보다 정직하지만, sync/start 직후 picker가 한동안 비는 운영 증상으로 보일 수 있어 doctor/로그에 entitlement skipped: no usable client version 한 줄이 있으면 디버깅이 쉽습니다.

경로/심볼 - 범위 정직성 - 리포터가 upstream `/codex/models` 응답 캡처를 안 줘서 confirmed-negative와 일시 discovery 실패를 구분하지 못합니다. 버전 수정이 맞아도 다른 이유로 명단이 비면 증상은 남습니다. PR이 재오픈+redacted capture를 다음으로 둔 것은 맞습니다.

메인테이너의 판단이 필요한 지점
- Responses/Direct 경로에도 요청 `client_version`(또는 User-Agent에서 뽑은 값)을 이어서 넣을지, 카탈로그(/v1/models)만으로 충분한지
- 계정 캐시를 `accountId+clientVersion` 복합 키로 올릴지, 지금처럼 단일 슬롯+버전 태그만으로 둘지
- `isCodexModelEntitlementSnapshotCurrent`에 버전을 넣을지(계획서와 구현 불일치)
- 자격증명으로 upstream roster를 치는 표면이라, 작성자=메인테이너인 이번에도 비작성자 승인 한 줄을 형식적으로 남길지(상태 스냅샷의 governance gap)
- 랜딩 후 `#2886`을 바로 닫을지, 캡처 없는 한 fix shipped, reopen if still missing 으로 남길지

너의 추천
CI(해당 테스트 묶음·typecheck·privacy:scan) 초록이면 `dev`에 머지해 `#2886`을 닫고, 머지 직후 leftover 이슈/PR 정리 규칙대로 처리한다. 후속으로 (1) auth-context/Direct에 버전 전달 여부 한 줄 정책 고정, (2) sync 경로에서 usable version이 없을 때 관측 가능 로그/doctor 힌트, (3) 필요하면 snapshot-current에 버전 포함 — 이 세 가지는 이 PR을 막지 말고 짧은 follow-up으로 둔다. 지금은 `0.0.0`이 만든 가짜 confirmed-negative를 끊는 것이 우선이다.

이 댓글은 grok-bot이 작성했습니다

PR #2905, opened 2026-08-29
fix(gui): reserve the sidecar hint in lines so the pair stays aligned

lidge-jun:
## Summary

The dashboard's two sidecar cards are a matched pair whose control rows must start at the same y. In **Russian and French** they were **19.5px apart** — exactly one line box — at every two-up width from 1600px down to 740px. Six other locales measured 0px, which is why this survived.

The fix is two declarations, and **both are load-bearing**:

```css
.dash-sidecar-row-card { align-content: start; }
.dash-sidecar-row-card .dash-sidecar-copy .setting-hint { min-height: 3lh; }
```

They close two independent halves of the same symptom.

**1. The wrapped lines were mis-distributed.** Both cards wrap their control group onto a second flex line. The grid stretches both cards to the taller one's height and `align-content` defaults to `stretch` for a multi-line flex container, so each card spread its own leftover space across its own lines. The two cards' content heights differ — the vision card's control group carries the advanced disclosure — so the card with more slack pushed its control line down. `align-content: start` packs both cards' lines from the top. It is `align-content` rather than `align-items` because the mis-distributed thing is the *lines*; `align-items: center` stays inherited and keeps the single-line (one-column) regime centred, where `align-content` is inert.

**2. The copy row those lines pack against was unequal.** It had been equalised by `min-height: 3.9375rem` on the copy block — 63px, documented as *"21px title + 3px hint margin + two 19.5px hint lines"*. That encodes a two-line assumption and a hard-coded line-height. The ru/fr vision hint wraps to a **third** line at a two-up card (82.5px of copy against 63px), so the band stopped describing the taller card. The floor moves to the hint and is expressed in `lh`: three line boxes is the longest shipped hint at the narrowest two-up card, so the shorter hint reserves the same three lines. In `lh` rather than `rem`, a font or line-height change cannot invalidate it.

### Each declaration proven necessary by removal

Taken on the shipped stylesheet, neutralising one declaration at a time and re-measuring the rendered page at ru/fr/en/ko × 1024/1100:

| stylesheet under test | worst paired offset | what breaks |
|---|---|---|
| **both shipped** | **0.0px** | nothing |
| `3lh` only (`align-content` back to its `stretch` default) | **27.8px** | copy rows equal, but each card distributes its own slack across its own lines |
| `align-content: start` only (old `3.9375rem` band restored) | **19.5px** | lines pack from the top, but the ru/fr copy rows are unequal (63px vs 82.5px) |

[...]

PR #2902, opened 2026-08-29
fix(gui): reserve the sidecar hint in lines so the pair stays aligned

lidge-jun:
## Summary

The dashboard's two sidecar cards are a matched pair whose control rows must start at the same y. In **Russian and French** they were **19.5px apart** — exactly one line box — at every two-up width from 1600px down to 740px. Six other locales measured 0px, which is why this survived.

Both cards wrap their control group onto a second flex line, and that line follows its own card's copy height, so the copy row has to be equal in both cards. It was equalised by `min-height: 3.9375rem` on the copy block — 63px, documented as *"21px title + 3px hint margin + two 19.5px hint lines"*. That encodes a two-line assumption and a hard-coded line-height. The ru/fr vision hint wraps to a **third** line at a two-up card (82.5px of copy against 63px), so the band stopped describing the taller card.

The floor now sits on the hint and is expressed in `lh`:

```css
.dash-sidecar-row-card .dash-sidecar-copy .setting-hint { min-height: 3lh; }
```

Three line boxes is the longest shipped hint at the narrowest two-up card, so the shorter hint reserves the same three lines and both control rows start together. In `lh` rather than `rem`, a font or line-height change cannot invalidate it.

### Before / after

Red rules mark each card's control-row top edge. Russian, 1024px:

| before — 19.5px apart | after — 0px |
|---|---|
|  |  |

French, 1024px:

| before — 19.5px apart | after — 0px |
|---|---|
|  |  |

### Per-locale measurement

| locale | hint lines (web search / vision) | before | after |
|--------|----------------------------------|--------|-------|
| en | 2 / 2 | 0px | 0px |
| ko | 1 / 2 | 0px | 0px |
| ja | 2 / 2 | 0px | 0px |
| zh | 1 / 1 | 0px | 0px |
| de | 2 / 2 | 0px | 0px |
| tr | 2 / 2 | 0px | 0px |
| **ru** | **2 / 3** | **19.5px** | **0px** |
| **fr** | **2 / 3** | **19.5px** | **0px** |

### Why not subgrid

[...]

PR #2897, opened 2026-08-29
fix(codex): bound the stored Pool 401 recovery budget by account

lidge-jun:
## Summary

Carries **@luvs01's #2895** onto current `dev` and fixes the one blocker I found reviewing it.
Their commit `a838b071c` is the first commit here with authorship intact; mine is the follow-up.
Credit for the analysis and the structure is theirs.

This is gap 5 of #2892. #2889 gave an ordinary stored Codex Pool account one generation-fenced
forced refresh plus one same-account replay after a pre-stream 401, but the replay's *result* was
not final: a replay 429/402 could still be composed with another Pool account, a remembered compact
model, a combo target, or a policy-fallback candidate — so one logical request could spend several
accounts' quota after its budget was already used.

Their structure is kept as-is: `codexMain401ReplayAttempted` becomes a tri-state
`codex401ReplayKind` (`"main" | "stored" | null`), an `onStoredPool401ReplayDispatched` signal is
threaded through combo and policy fallback, and compact guards both its pool-rotation and
remembered-model paths. `main-pool` keeps full recovery breadth, which is right — a native main 401
is not a stored-account budget.

### The blocker: the budget bounds accounts, not rescue

The original patch enforced it with one line:

```ts
if (codex401ReplayKind === "stored" && upstreamResponse.status >= 400) break;
```

That break sits **above** two ladders that send to the account already paying:

- `shouldRetryCodexPoolAccountModel400` (`core.ts:4200`) — an allow-listed gated-model 400, retried
  on the *same* account when the refreshed roster still grants the model
  (`retryCodexPoolOnAlternateAccount` sets `retryAuthCtx = firstAuthCtx` for exactly that case).
- `attemptOpaqueBlobRecovery` (`core.ts:4249`) — a rejected opaque reasoning/compaction blob, whose
  one-shot rebuild strips it and resends to the same refreshed account.

Neither charges a different account, so neither is in the budget #2892 asked to bound. With the
broad break, `401 → refresh → invalid_encrypted_content` returned a 400 to the user where the
rebuild would have succeeded. Restoring that one line turns the new opaque-blob regression red, so
this is a real behavioural difference rather than a stylistic one.

The corrected bound is stated in terms of what is actually scarce — another account's quota:

- a quota failure (429/402) after a stored replay has no same-account move left, so it stays
  terminal at the pool-retry site;
- a gated-model 400 keeps its ladder, with a new `sameAccountOnly` refusing alternate-account
[...]

PR #2906, opened 2026-08-29
fix(gui): cap the log table against the dynamic viewport

lidge-jun:
## Summary

Two GUI sizing fixes found while auditing the dashboard for viewport-dependent layout defects, plus the devlog cleanup that #2905's review asked for and that missed its merge window.

Now targets `dev` directly — its parent #2905 has landed, so this is no longer a stacked child.

**`.logs-table-wrap` capped its scroll height with `calc(100vh - 260px)`.** Static `vh` resolves against the *large* viewport, so on mobile the cap is computed for a viewport taller than the one the user can actually see and the last rows end up under the browser chrome. It is now `100dvh`, which is what the rest of the shell already uses — `.app`, `.sidebar`, `.main-inner--combos` and the mobile drawer are all `100dvh`, so this line was the outlier rather than the convention.

**The corner toast's width cap never applied.** `.action-toast` sets `max-width: min(480px, calc(100vw - 48px))`, but every toast also carries `.notice`, which sets `max-width: var(--prose-measure)` (70ch = 542px) and is declared *later* in the same file. Both selectors are specificity 0,1,0, so source order won and a long toast rendered **542px instead of 480px**. The cap is now on the two-class `.action-toast.notice`.

That fix needed one correction during verification: my first version set only `max-width: 480px`, which dropped the viewport term and let the toast run to the screen edge at narrow widths (measured `left = 0` at 430px, against the 24px inset the right side keeps). It restates both halves of the original expression.

### Not changed: the `vw` → containing-block rewrite

The original plan also swapped `vw` for containing-block insets, on the theory that `100vw` includes the classic scrollbar gutter. That was reverted before commit because it could not be reproduced on this surface — the probe measured `innerWidth == clientWidth`, gap 0 — and it would have been an unmeasured edit to a live width cap. The units stay `vw`. The reproduced defect was the specificity one, and that is what this PR fixes.

### Regression test

The hygiene gate flagged `missing_regression_test` on the first revision, and the reviewer asked for the same treatment #2905 gave the sidecar rules rather than a test exception. `gui/tests/viewport-scroll-caps.test.ts` pins both:

- `.logs-table-wrap` must cap against `100dvh` and must not return to static `100vh`.
- The toast cap must stay on the compound `.action-toast.notice` selector with **both** halves of `min(480px, calc(100vw - 48px))` intact.

The second test also guards its own premise: it asserts `.notice` still appears *after* the compound rule in source order, because that ordering is the entire reason the compound selector is required. If `.notice` ever moves above it, the test fails rather than quietly allowing a "simplification" back to a single class.

Source-text assertions, matching the existing `gui/tests` convention for CSS contracts — happy-dom performs no layout, so a computed `max-width` there would prove nothing. Driven **red first**: against the parent commit's stylesheet both tests fail (the `dvh` assertion, and `.action-toast.notice` not existing at all); both pass on this branch's.

### Devlog cleanup carried over from #2905's review

@Ingwannu's review of #2905 asked for two things beyond the record correction that landed in that PR. They are here because #2905 merged while they were being pushed:

- `git diff --check` reported "new blank line at EOF" on `000_baseline_and_roadmap.md` and `011_audit_correction_align_content.md`. Both had two newlines at EOF; now one. The review said all seven files, but the other five already ended with a single newline, so they are untouched.
- `020` and `030` now state their shipping status in the first line the reader sees, instead of leaving it inferable from a "Deferred" section at the bottom of `013`. `020` is **withdrawn** — the zero-width `auto-fit` track it reports is normal behaviour and no code change was made for it. `030` is **implemented by this PR**.

`git diff --check` is clean on this head.

## Verification

Measured in a real browser via CDP, not asserted from source alone:

- `.logs-table-wrap`: computed `max-height` compared against `visualViewport.height` at a 430-wide mobile profile where the visual viewport is smaller than the large viewport — no cap exceeds the visible height.
- Toast: computed `max-width` is **480px** at 1440 (was 542px), and at 430px it keeps its 24px inset on both sides rather than reaching `left = 0`.

Focused GUI gates on this exact head:

- `gui/tests/viewport-scroll-caps.test.ts` — 2 pass / 0 fail (red on the pre-fix stylesheet)
- `gui/tests/sidecar-layout.test.ts` — 8 pass / 0 fail (unchanged by this PR, run because it covers the same stylesheet family)

Remote Linux gates against this exact commit:

- `cd gui && bun test tests` — rc=0
- `bun run typecheck` — rc=0
- `bun run lint:gui` — rc=0
- `bun run test` (full repo suite) — rc=0

### Screenshots

The same long toast at 1440px, before and after. Before, the later `.notice` rule won and it rendered 542.1px; after, the compound selector applies its 480px cap:

| before - 542.1px | after - 480px |
|---|---|
|  |  |

At 430px the viewport term binds instead, giving 382px (= 430 - 48) and keeping the 24px inset on both sides - which is what the width-only version lost:

## Checklist

Scope is one stylesheet, one new test file, and four devlog files; no runtime, routing, auth, or release paths are touched.

### DIFF
--- a/devlog/_plan/260829_gui_dashboard_slop/000_baseline_and_roadmap.md
+++ b/devlog/_plan/260829_gui_dashboard_slop/000_baseline_and_roadmap.md
@@ -121,4 +121,3 @@ distinct.
   `ssh lidge` + `ocx-run`; pushes use `--no-verify` only after those gates.
 - Delivery is a stacked PR chain onto `dev`, each PR carrying screenshots
   (`enforce-target` requires a screenshot for GUI PRs).
-
--- a/devlog/_plan/260829_gui_dashboard_slop/011_audit_correction_align_content.md
+++ b/devlog/_plan/260829_gui_dashboard_slop/011_audit_correction_align_content.md
@@ -112,4 +112,3 @@ flow (`position: fixed`, off-canvas at `x=-280`), so `.main-inner` JUMPS from
 526px to 750px. The sidecar grid re-splits into two columns at 349px each and the
 22.5px misalignment returns — on tablet widths, below the width where it was last
 believed fixed. Any fix must be verified at 760/740, not only at desktop widths.
-
--- a/devlog/_plan/260829_gui_dashboard_slop/020_phantom_grid_track.md
+++ b/devlog/_plan/260829_gui_dashboard_slop/020_phantom_grid_track.md
@@ -1,5 +1,10 @@
 # 020 — Phantom zero-width auto-fit track (wp2)
 
+> **WITHDRAWN — nothing in this document ships.** The investigation concluded the
+> reported defect is not a defect: a collapsed zero-width `auto-fit` track is
+> normal behaviour, and no code change was made for it. Kept as the record of why
+> the track is expected, so the next person who measures it does not re-open it.
+
 ## Defect
 
 At vw ≥ 1440 both dashboard grids compute a third, zero-width column:
--- a/devlog/_plan/260829_gui_dashboard_slop/030_dynamic_viewport_units.md
+++ b/devlog/_plan/260829_gui_dashboard_slop/030_dynamic_viewport_units.md
@@ -1,5 +1,10 @@
 # 030 — Dynamic viewport units in scroll surfaces (wp3)
 
+> **Implemented by this pull request**, separately from the sidecar alignment fix
+> that the rest of this unit records. It is a different change to a different
+> file (`gui/src/styles.css`), kept in the same unit because one audit pass found
+> both.
+
 ## Defect
 
 `gui/src/styles.css:2003`:
--- a/gui/src/styles.css
+++ b/gui/src/styles.css
@@ -763,6 +763,16 @@ a.btn, a.btn:hover { text-decoration: none; }
 }
 /* Refresh icon spins while a sync is in flight (the `.spin` keyframes already exist). */
 .spin-icon { animation: spin 0.9s linear infinite; }
+/* The width cap needs two classes to beat `.notice { max-width: var(--prose-measure) }`,
+   which every toast also carries and which is declared LATER in this file. At equal
+   specificity source order won, so the toast resolved to 70ch (542px) rather than its
+   design width — visible as a toast wider than intended at desktop sizes. */
+.action-toast.notice {
+  /* Both halves of the original cap have to be restated here, not just the design width:
+     dropping the viewport term let the toast run to the screen edge at narrow widths
+     (measured left=0 at 430px, losing the 24px inset the right side keeps). */
+  max-width: min(480px, calc(100vw - 48px));
+}
 /* Toast dismiss affordance: quiet ghost button inside the notice row. */
 .action-toast-dismiss {
   flex: none;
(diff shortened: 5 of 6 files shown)


lidge-jun:
## 리뷰 · 우선순위 50 / 80

이 PR은 닫힌 #2904의 재기반이다. `gui/src/styles.css` 한 파일, 고침 두 줄은 #2904와 바이트가 같다.
베이스는 이제 `codex/gui-sidecar-pair-alignment-v2` (#2905) 이다. #2905 의 merge-base는 현재 `dev` HEAD `8df705120` 이라, 예전 #2902/#2904 스택에 있던 #2897 재포함은 없다.
그래도 직접 타깃은 아직 `dev` 가 아니다. #2905가 먼저 끝나야 리타깃할 수 있다.

현재 `dev` 의 `.logs-table-wrap` (styles.css 2001-2004행) 은 `max-height: calc(100vh - 260px)` 이다.
같은 파일에서 `.app`(244행), 사이드바(247행), 콤보 셸(411-412행), 로그 상세(2071행), 모바일 서랍(2198행)은 이미 `100dvh` 다.
표만 큰 뷰포트를 본다. 모바일 주소창이 있으면 마지막 줄이 크롬 아래로 간다.
#2859 가 로그 폴링을 2초+백오프로 고정한 뒤에도, 표가 보이는 높이는 이 한 줄이다.

토스트는 특이성 문제다. 749-755행 `.action-toast` 는 `min(480px, calc(100vw - 48px))` 를 주지만, 토스트마다 붙는 1205행 `.notice` 가 아래에서 `max-width: var(--prose-measure)` (70ch≈542px) 를 준다. 특이성이 같아서 나중 규칙이 이긴다.
이 PR은 `.action-toast.notice` 로 올리고 원래 식 두 항을 다시 적는다. 너비만 남기면 430px에서 왼쪽이 0이 되는 것을 측정해서 되돌렸다고 본문에 있다.

착지 조건은 #2904와 같다.
라벨 `intake: hygiene-blocked` , 상태가 draft 다. hygiene 는 `missing_regression_test` 다. `gui/src/` 가 바뀌었는데 테스트 파일이 없다.
#2905 는 `gui/tests/sidecar-layout.test.ts` 로 규칙을 고정했다. 이쪽은 그 패턴을 따르지 않았다.
`gui/tests/apikeys-layout.test.ts` 133행의 `100dvh` 금지는 `styles-apikeys-workspace.css` 만 본다. 충돌하지 않는다.

라인 2003(현재 dev styles.css) - 100vh 는 셸 나머지와 다르다. 이 PR의 100dvh 교체는 맞다. 빼는 260px 은 그대로라서 모바일 크롬 높이가 다르면 표가 또 짧거나 길다.
경로 .action-toast.notice - 특이성 수정은 맞다. 테스트가 없어서 .notice 가 다시 이기거나 뷰포트 항이 빠져도 CI가 모른다.
경로 calc(100vw - 48px) - classic 스크롤바 환경에서는 vw 가 보이는 너비보다 크다. 측정이 안 나와 안 바꾼 것은 맞다.
경로 베이스 #2905 - 직접 `dev` 가 아니므로 enforce-target 이 다시 실패할 수 있다. #2905 랜딩 후 리타깃이 필요하다.

메인테이너의 판단이 필요한 지점
- #2905를 먼저 넣을지, 두 GUI PR을 한 브랜치로 합칠지
- 테스트 없이 test-exception-approved 를 줄지, sidecar-layout 처럼 CSS 단언을 넣을지
- 닫힌 #2904를 #2906 랜딩 뒤에 landed-via-maintainer 로 처리할지

너의 추천
지금은 머지하지 마세요. #2905가 `dev` 에 들어간 뒤 이 브랜치를 `dev` 로 맞추세요.
그 다음 `.logs-table-wrap` 의 `100dvh` 와 `.action-toast.notice` 의 min(480px, calc(100vw - 48px)) 를 `gui/tests` 에 고정하세요. 고침 두 줄은 맞고 types/config 분할과도 무관하다. 막힌 이유는 스택과 테스트 공백이다.

이 댓글은 grok-bot이 작성했습니다

PR #2909, opened 2026-08-29
fix(service): launch the systemd unit through the stable ocx executable

### LINKED ISSUE #2898: [Bug]: mise upgrades can strand systemd on removed versioned OpenCodex paths
## Client or integration

Codex CLI

## Area

Service lifecycle

## Summary

The Linux systemd user unit generated by `ocx service install` embeds the absolute, versioned OpenCodex CLI and bundled Bun paths. When OpenCodex is installed through mise, `mise upgrade opencodex` installs the next release under a different version directory without running OpenCodex's own update workflow, so the installed unit remains tied to the previous version.

Once mise removes the previous install or the service is restarted after pruning, systemd has no executable at the baked path and the background proxy cannot start. `ocx service repair` fixes the unit, but it has to be run manually after every external package-manager upgrade.

I currently need this mise configuration as a local workaround:

```toml
opencodex = { version = "latest", postinstall = 'if test -f "$HOME/.config/systemd/user/opencodex-proxy.service"; then "$MISE_TOOL_INSTALL_PATH/node_modules/.bin/ocx" service repair; fi' }
```

The guard makes a first installation without a service a no-op and forces an existing service to be re-baked after mise changes the install path.

Expected: an installed OpenCodex service remains startable across supported external package-manager upgrades, or OpenCodex provides a package-manager-independent launcher/service path that does not require a per-manager postinstall repair hook.

Actual: the service definition is coupled to one mise version directory and becomes stale when that directory is replaced or pruned.

## Reproduction

1. Install OpenCodex through mise:

   ```bash
   mise use -g opencodex@latest
   ```

2. Install the Linux user service:

   ```bash
   ocx service install
   ```

3. Inspect the generated unit:

   ```bash
   systemctl --user cat opencodex-proxy.service | grep -E '^(ExecStart|Environment="OCX_BUN_RUNTIME_PATH)='
   ```

   Both the bundled Bun runtime and CLI entry point are under a version-specific path such as:

   ```text
   /home/<user>/.local/share/mise/installs/opencodex/2.35.0/...
   ```

4. Upgrade to the next OpenCodex release through mise:

   ```bash
   mise upgrade opencodex
   ```

5. Confirm the unit still names the previous version directory. After that old install is removed or pruned, restart the user service or reboot:

   ```bash
   systemctl --user restart opencodex-proxy.service
   ocx status
   ```

6. Repairing manually re-bakes the unit and restores service health:

   ```bash
   ocx service repair
   ocx status
   ```

This is distinct from #2411 and #2412. Their released fixes improve routing visibility and correctly reject a version-manager-owned Codex shim, but they do not make the OpenCodex systemd service survive an external OpenCodex package upgrade.

## Version

2.35.0

## Operating system

Ubuntu 24.04.4 LTS on WSL2, kernel 6.18.33.2-microsoft-standard-WSL2

## Provider and model

Not provider- or model-specific.

## Logs or error output

```shell
$ systemctl --user show opencodex-proxy.service -p ActiveState -p UnitFileState
ActiveState=active
UnitFileState=enabled

$ systemctl --user cat opencodex-proxy.service
# ExecStart and OCX_BUN_RUNTIME_PATH both contain:
/home/<user>/.local/share/mise/installs/opencodex/2.35.0/...
```

## Checks

lidge-jun:
## Summary

Closes #2898.

A version manager installs OpenCodex into a versioned directory and deletes the old one on
upgrade. Both paths baked into the systemd unit — the bundled Bun and the CLI entry — are
computed from `import.meta.dir`, so both live inside that directory. After a `mise` upgrade
`/bin/sh` still starts, but `exec <old-bun> <old-cli>` cannot resolve, and `Restart=on-failure`
turns that into a restart loop while `systemctl` still reports the unit as installed. Nothing in
the start path notices: `startSystemd` checks unit existence and daemon-reload state, never
whether the launch paths still exist.

`installSystemd` now resolves the stable `ocx` executable lexically from `PATH` and bakes that.
A mise/asdf/npm shim survives the upgrade and dispatches to whatever version is current.

Three decisions worth naming:

- **Lexical, not resolved.** Calling `realpath` on the shim would write the versioned target back
  into the unit and recreate the bug. The indirection is the whole point.
- **Absolute only.** A bare `ocx` in `ExecStart` would be re-resolved through `PATH` on every
  restart, turning a service definition into a PATH-hijacking surface. One validated absolute file
  is fixed at install time.
- **No runtime stamp in launcher mode.** `OCX_BUN_RUNTIME_PATH`/`_SOURCE` are omitted, because
  stamping them pins the runtime to the directory the upgrade deletes. The launcher resolves the
  current package's Bun itself.

Source checkouts with no `ocx` launcher keep the previous direct Bun + CLI form.

`ServiceInstallState` records `launcherPath`, and `bakedServicePathsDiagnostic` judges staleness
against the launcher when one is present. Without that half, a **healthy** service is reported
`STALE` after every upgrade, since the recorded pair points into the replaced directory — the
mutation below reproduces exactly that. State files lacking the field behave as before.

The launcher is resolved **once** and shared by the unit and the state write, so the diagnostic can
never validate a path the unit does not run.

### Two pre-existing tests changed, both deliberately

- The #848 provenance test asserted the unit always stamps the Bun pair. That is now mode-specific,
  so it asserts **both** modes: the pair is stamped where it is baked, and absent where a launcher
  replaces it. The invariant is unchanged; only its scope is stated.
- The install-ordering test matched `buildUnit()` literally and now matches the call prefix, since
  `buildUnit` takes the resolved launcher. It additionally pins that the launcher is resolved once,
  before the write, and reused for the state.

## Verification

```
bun test tests/service.test.ts   144 pass / 0 fail
bun x tsc --noEmit               clean
bun run privacy:scan             passed
```

Mutation ledger — each new test fails only for its own reverted line:

| Mutation | launcher-unit test | retargeted-shim test | staleness test |
|---|---|---|---|
| `ExecStart` always bakes the direct pair | **fail** | **fail** | pass |
| diagnostic ignores `launcherPath` | pass | pass | **fail** |

The staleness mutation reproduces the user-visible symptom precisely: a healthy launcher install
reported `STALE baked paths (missing: .../installs/2.35.0/bun, .../installs/2.35.0/cli/index.ts)`.

The retargeted-shim test executes the scenario rather than asserting it: a real shim is pointed at
v1, invoked, retargeted to v2, v1 is deleted, and it is invoked again — proving the launcher still
reaches live code after the version directory is gone.

**Not verified:** no real systemd host was exercised. All systemd assertions are against generated
unit text and temp directories; no live unit on any machine was modified. Existing units cannot
self-migrate once their old executable is deleted, so they need one `ocx service repair` after
upgrading — documented in `lifecycle.md`.

## Checklist

- [x] Focused tests cover the change and were proven non-vacuous by mutation
- [x] `bun x tsc --noEmit` clean and `bun run privacy:scan` passes
- [x] The unit still reads the token from its file and never embeds it
- [x] launchd and Windows service generation untouched
- [x] Docs updated with the one-time repair migration
- [x] Targets `dev`

## Summary by CodeRabbit

* **New Features**
  * Linux systemd services now use the stable `ocx` executable on your `PATH`, allowing services to continue working after version-manager upgrades.
  * Service installations record the launcher path and avoid embedding version-specific runtime paths.

* **Bug Fixes**
  * Prevented restart loops caused by services referencing deleted installation files.

* **Documentation**
  * Added guidance for repairing services installed before this update with `ocx service repair`.

### DIFF
--- a/docs-site/src/content/docs/reference/cli/lifecycle.md
+++ b/docs-site/src/content/docs/reference/cli/lifecycle.md
@@ -231,6 +231,17 @@ interrupted package update removed either file, it logs one `installation is inc
 stops instead of retrying the same missing executable every five seconds. Reinstall opencodex, then
 run `ocx service repair` to refresh the task with the restored package paths.
 
+On Linux, the systemd unit invokes the stable `ocx` executable found on `PATH` at install time
+rather than the Bun and CLI paths inside the installed package tree. Version managers such as
+**mise** and **asdf** install into a versioned directory and delete the old one on upgrade, which
+used to leave the unit pointing at files that no longer existed — systemd then restart-looped while
+still reporting the service as installed. A shim path survives the upgrade, so the unit keeps
+resolving. Source checkouts without an `ocx` launcher keep the previous direct Bun + CLI form.
+
+Units installed before this change still carry the old versioned paths and cannot migrate
+themselves — once the old executable is deleted, no opencodex code runs to fix it. Run
+`ocx service repair` once after upgrading; subsequent version changes need no action.
+
 | Subcommand | Action |
 | --- | --- |
 | none | Install and start when absent; otherwise refresh and restart the existing service without re-registering it. |
--- a/src/service.ts
+++ b/src/service.ts
@@ -9,7 +9,7 @@ import { execFileSync, execSync, spawnSync } from "node:child_process";
 import { findLiveProxy, proxyIdentityAt, SERVICE_STOP_LIVENESS } from "./server/proxy-liveness";
 import { chmodSync, existsSync, mkdirSync, mkdtempSync, readFileSync, rmdirSync, unlinkSync, writeFileSync } from "node:fs";
 import { homedir, tmpdir } from "node:os";
-import { dirname, join, posix, resolve, win32 } from "node:path";
+import { dirname, isAbsolute, join, posix, resolve, win32 } from "node:path";
 import { expandUserPath, getConfigDir, loadConfig } from "./config";
 import { readPid, removePid, removeRuntimePort, verifyPidIdentity } from "./config/process-state";
 import { restoreNativeCodex, restoreNativeCodexAsync } from "./codex/inject";
@@ -65,6 +65,37 @@ function cliEntry(): { bun: string; bunRuntimeSource: BunRuntimeSource; cli: str
   // different binary than the one actually baked.
   const runtime = durableBunRuntime();
   return { bun: runtime.path, bunRuntimeSource: runtime.source, cli: join(import.meta.dir, "cli", "index.ts") };
+}
+
+/**
+ * The stable `ocx` launcher to bake into a systemd unit, or null to fall back to the
+ * Bun + CLI pair.
+ *
+ * `cliEntry()` resolves both of its paths from `import.meta.dir`, so they point INSIDE
+ * the installed package tree. Under a version manager that tree is a versioned directory:
+ * `~/.local/share/mise/installs/npm-opencodex/2.35.0/...`. An upgrade installs 2.36.0 and
+ * deletes 2.35.0, after which the unit's `exec <old-bun> <old-cli>` cannot resolve, and
+ * `Restart=on-failure` turns that into a restart loop (#2898). The shim in
+ * `~/.local/share/mise/shims/ocx` survives the upgrade and dispatches to whatever version
+ * is current, so it is the durable thing to name.
+ *
+ * Deliberately LEXICAL. Resolving the symlink would write the versioned target back into
+ * the unit and reintroduce the bug — the indirection is the entire point.
+ *
+ * Only an absolute path is accepted. A bare `ocx` would be re-resolved through `PATH` on
+ * every restart, which turns a service definition into a PATH-hijacking surface; naming
+ * one validated absolute file keeps the target fixed at install time.
+ */
+function stableLauncherEntry(deps: { env?: NodeJS.ProcessEnv; exists?: (path: string) => boolean } = {}): string | null {
+  const env = deps.env ?? process.env;
+  const exists = deps.exists ?? existsSync;
+  const entries = (env.PATH ?? "").split(":");
+  for (const entry of entries) {
+    if (!entry || !isAbsolute(entry)) continue;
+    const candidate = join(entry, "ocx");
+    if (exists(candidate)) return candidate;
+  }
+  return null;
 }
 
 function plistPath(): string {
@@ -155,6 +186,14 @@ export interface ServiceInstallState {
   /** Baked at install; lets status flag paths gone stale after npm prefix/nvm moves. */
   bunPath?: string;
   cliPath?: string;
+  /**
+   * Linux only. The stable `ocx` launcher the unit actually invokes, when one was found.
+   * Present means `bunPath`/`cliPath` are provenance for the install, NOT what systemd
+   * runs — so staleness must be judged against THIS path instead. A version-manager
+   * upgrade replaces the directory those two point into while the launcher survives, and
+   * checking the old pair would report a stale service that is in fact healthy.
+   */
+  launcherPath?: string;
   /** v2: which Windows backend was chosen at install; absent (v1/legacy) means scheduler. */
   backend?: ServiceBackend;
   winswVersion?: string;
@@ -167,7 +206,7 @@ export function parseServiceInstallState(value: unknown): ServiceInstallState | 
   if (state.version !== 1 && state.version !== 2) return null;
   if (typeof state.codexHome !== "string" || state.codexHome.length === 0) return null;
   if (typeof state.opencodexHome !== "string" || state.opencodexHome.length === 0) return null;
-  for (const key of ["bunPath", "cliPath", "winswVersion", "winswSha256"] as const) {
+  for (const key of ["bunPath", "cliPath", "launcherPath", "winswVersion", "winswSha256"] as const) {
     if (state[key] !== undefined && (typeof state[key] !== "string" || state[key].length === 0)) return null;
   }
   if (state.version === 1) {
@@ -178,7 +217,7 @@ export function parseServiceInstallState(value: unknown): ServiceInstallState | 
   return state as unknown as ServiceInstallState;
 }
 
-function writeServiceInstallState(backend: ServiceBackend = "scheduler"): void {
+function writeServiceInstallState(backend: ServiceBackend = "scheduler", launcherPath?: string | null): void {
   const { bun, cli } = cliEntry();
   const state: ServiceInstallState = {
     version: 2,
@@ -186,6 +225,7 @@ function writeServiceInstallState(backend: ServiceBackend = "scheduler"): void {
     opencodexHome: currentOpenCodexHome(),
     bunPath: bun,
     cliPath: cli,
+    ...(launcherPath ? { launcherPath } : {}),
     backend,
     ...(backend === "native" ? { winswVersion: WINSW_VERSION, winswSha256: WINSW_SHA256 } : {}),
   };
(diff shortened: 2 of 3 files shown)


lidge-jun:
## 리뷰 · 우선순위 68 / 80

이 PR은 이슈 #2898을 닫으려는 리눅스 systemd 설치 고침이다. 지금 `dev` HEAD는 `fc74e2026` (#2905 GUI sidecar 정렬) 이고, 그 직전은 #2908 웹서치 HTTP 버전 핀, #2907 Codex 0.151 카탈로그였다. 그 흐름과 겹치지 않는 독립 버그 픽스다. types.ts/config.ts 분리 캠페인에도 걸리지 않는다.

문제의 핵심을 쉽게 말하면 이렇다. `mise upgrade` 같은 버전 매니저는 OpenCodex를 `~/.local/share/mise/installs/.../2.35.0/` 같은 버전 폴더에 넣고, 업그레이드하면 그 폴더를 지운다. 그런데 지금 `dev`의 `installSystemd`/`buildUnit`은 `cliEntry()`로 구한 Bun 경로와 `cli/index.ts`를 유닛 ExecStart에 그대로 박아 넣는다. 둘 다 `import.meta.dir` 기준이라 그 버전 폴더 안을 가리킨다. 업그레이드 뒤에는 `/bin/sh`는 뜨지만 `exec <옛-bun> <옛-cli>`가 실패하고, `Restart=on-failure` 때문에 재시작 루프가 돈다. systemctl은 유닛이 설치돼 있다고만 보여서, 사용자는 “서비스는 있는데 안 됨” 상태에 빠진다. #2898이 그 증상이다.

이 PR이 하는 일은 세 갈래다. 첫째, `stableLauncherEntry()`가 PATH에서 절대경로 디렉터리만 훑어 `ocx` 심을 찾고, 있으면 그 경로를 ExecStart에 넣는다. 심링크를 realpath로 풀지 않는다. 풀면 다시 버전 폴더를 박게 되어 버그가 되살아난다. 둘째, launcher 모드에서는 `OCX_BUN_RUNTIME_PATH`/`OCX_BUN_RUNTIME_SOURCE`를 유닛에 안 찍는다. 찍으면 런타임이 지워질 폴더에 묶인다. 토큰은 예전처럼 파일에서 읽고 유닛에 넣지 않는다 (#2107 유지). 셋째, `ServiceInstallState.launcherPath`를 기록하고 `bakedServicePathsDiagnostic`이 launcher가 있으면 그 경로만 본다. 안 그러면 업그레이드 직후 멀쩡한 서비스가 STALE로 잘못 찍힌다. 유닛과 state에 같은 launcher 값을 한 번만 구해 넘긴다.

테스트도 맞춰 두었다. launcher 유닛에 버전 경로가 안 들어가는지, 심을 v1→v2로 돌리고 v1을 지운 뒤에도 명령이 살아 있는지, launcher 설치의 staleness가 버전 폴더가 아니라 launcher 부재로만 뜨는지. #848 provenance 테스트는 “쌍을 굽는 모드”와 “launcher 모드”를 둘 다 검사하도록 바꿨다. 문서(`lifecycle.md`)에는 예전 유닛은 한 번 `ocx service repair`가 필요하다고 적어 두었다. 라이브 systemd 호스트는 돌리지 않았다고 본문에 명시했다.

라인 / 심볼 이슈:

src/service.ts stableLauncherEntry - existsSync만 본다. 같은 이름 디렉터리이거나 실행 비트가 없는 파일이어도 통과한다. isFile + 실행 가능 여부 검사가 없으면 잘못된 후보를 구울 수 있다.

src/service.ts buildUnit 기본값 - deps.launcher를 안 넘기면 실제 PATH를 스캔한다. 기존 테스트 상당수가 아직 맨 buildUnit()을 쓴다. CI나 개발자 PATH에 ocx가 있으면 launcher 모드로 바뀌어, 직접 Bun+CLI를 가정한 단언이 흔들릴 수 있다. 직접 모드를 검증하는 곳은 { launcher: null }을 명시하는 편이 안전하다 (이번 PR이 #848 쪽만 고친 상태).

src/service.ts stableLauncherEntry PATH 순서 - PATH 앞쪽의 다른 ocx(깨진 심, 수동 복사본)가 mise/asdf 심보다 먼저면 그게 구워진다. 의도는 맞지만, 설치 직후 어떤 절대경로가 선택됐는지 사용자에게 한 줄 보여 주면 디버깅이 쉽다.

docs lifecycle.md / 본문 - 기존 유닛은 self-migrate가 안 되고 repair 한 번이 필요하다. 맞다. 다만 “업그레이드만 하면 끝”으로 읽는 사람이 있을 수 있으니, repair가 필요한 조건(이 픽스 이전에 깔린 유닛)을 릴리즈 노트에도 짧게 반복할 가치가 있다.

메인테이너의 판단이 필요한 지점
- macOS launchd / Windows도 버전 매니저로 같은 클래스의 “구운 절대경로” 문제가 생길 수 있다. 이번 스코프는 Linux systemd + #2898만이다. 후속 이슈로 남길지, 이번 PR에 억지로 넣지 말지.
- PATH 첫 ocx를 신뢰할 것인지, mise/asdf shim 디렉터리를 더 선호할 것인지.
- 라이브 systemd 검증 없이 merge할지. 유닛 텍스트·임시 디렉터리·mutation으로 증명했고, 저자는 한계를 명시했다.

너의 추천
머지해도 된다. #2898을 정확히 겨냥한 독립 버그 픽스이고, GUI 스택(#2906)과 안 겹친다. 머지 전에 가능하면 (1) stableLauncherEntry에 파일/실행 가능 가드, (2) 직접 모드를 가정하는 테스트에 { launcher: null } 명시를 작은 후속 커밋으로 넣는 것을 권한다. 라이브 호스트 검증은 있으면 좋고, 없어도 문서화된 repair 경로와 테스트 강도로 merge 가능 수준이다. 머지 후 #2898을 닫으면 된다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
Requesting changes on exact head `ba63b4605f2fdca8b9fc4e91fac6f13e9f6baa56`.

The stable lexical launcher is the right direction for #2898, and the current merge tree against `dev` is clean, but two runtime contracts still block this head:

1. Launcher-mode units drop the trusted `OPENCODEX_BUN_PATH` selection. Because systemd does not inherit the repair/install shell, the later stable launcher resolves bundled Bun instead of the explicitly selected runtime. Preserve the override durably without stamping a package-local bundled path, and cover install/repair plus generated-unit behavior.
2. `stableLauncherEntry` accepts any existing `PATH/ocx`, including a directory or non-executable file, then records it as healthy and creates a restart loop. Require a regular executable and continue scanning later PATH entries, with an invalid-first/valid-second regression.

Please also make the retargeted-shim test exercise the launcher with `execFileSync`/argument arrays and quote the generated shim target. Its current `sh -c` construction is path-sensitive and does not prove the contract when the temporary path contains spaces or shell metacharacters.

The exact-head CI being green does not cover these negative paths. Please update the head, resolve the current review threads, and rerun the focused service, Linux systemd, privacy, typecheck, and cross-platform gates; I will re-review the new exact head.

Ingwannu:
This merged at `4ec1cc958` while the exact-head review findings were still in flight. I am not proposing a revert: the stable-launcher direction fixes #2898, but the merged head still drops a trusted `OPENCODEX_BUN_PATH` selection in launcher-mode units and accepts a merely existing/non-executable `PATH/ocx` candidate.

I will follow up from current `dev` with a narrow repair that:

- carries only a provenance-authenticated Bun override into the stable launcher unit, without pinning the package-local bundled runtime;
- requires a regular executable launcher and continues past invalid PATH entries;
- makes direct `buildUnit()` tests deterministic instead of dependent on the review host's PATH;
- adds focused negative-path regressions and preserves the token-file boundary.

No release branch, runtime daemon, or user configuration will be changed by that follow-up.

PR #2915, opened 2026-08-29
fix(gui): correct the record and lock the test constants

lidge-jun:
## Summary

An independent final-gate audit of the merged `260829_gui_dashboard_slop` unit returned **NEAR-PASS** with five findings. Four were correct and are fixed here; each was re-verified against `origin/dev` before being accepted.

**My own citation fix in #2911 introduced an error.** Replacing the drifted line numbers, I wrote that `.action-toast` (749) and `.notice` (1215) both cap toast width with `calc(100vw - Npx)`. `.notice` does not use a viewport unit at all — it caps with `var(--prose-measure)`, which is exactly why it wins the cascade and why the compound selector is necessary. The second `calc(100vw - Npx)` belongs to `.toast-notice` at 1229. Naming the wrong rule undercut the explanation two paragraphs below it.

**A stylesheet comment overstated the test.** It claimed the three-line floor is something *"the regression test asserts by measuring rendered line counts"*. The committed test is source-text only and says so in its own header; the line count came from browser measurement recorded in `013`. The comment now separates what the test checks from where the number came from.

**`000` blamed the wrong property.** Its root-cause paragraph attributed the mis-distributed wrapped lines to `align-items: center`. `align-items` centres each item *within* its line; the lines themselves are `align-content`, which defaults to `stretch`. That distinction is the whole reason `align-content: start` ships while `align-items: center` stays. Corrected in place with a note, because the original wording is part of how the investigation went wrong.

**`000` still carried the drifted citations** that #2911 fixed only in `030` — `styles.css:2003`, `:755`, `:1222`, `:2198`, none of which point at the cited rule any more (2003 is now `min-width: 220px`). Now named by selector, with the reason stated.

## Test constants are now locked

The fifth finding was that the toast assertions accept any positive number:

```ts
expect(Number(cap![1])).toBeGreaterThan(0);   // a 1px design width passes
expect(Number(cap![2])).toBeGreaterThan(0);   // a 1px inset passes
```

The reviewer was right — those would survive a change that destroys the layout. Both toast constants and the log-cap subtrahend are now asserted exactly (480, 48, 260).

**Proved non-vacuous by mutation.** Rewriting the stylesheet to `min(479px, calc(100vw - 47px))` and `calc(100dvh - 261px)` turns both tests **red** (`Expected: 480, Received: 479`), where the previous assertions passed. The stylesheet was then restored byte-identical to `origin/dev`, verified by sha256, and the suites are back to 10 pass / 0 fail.

## Deliberately not changed

The reviewer also flagged that several *secondary* sidecar assertions can pass on an empty match — a `@container` loop that asserts nothing if no rule reaches it. That is a real weakness, but it is pre-existing and unrelated to this unit's fixes, and tightening it requires deciding what those queries must contain. That belongs in its own change rather than in a correction to this record.

## Verification

- `gui/tests/sidecar-layout.test.ts` + `gui/tests/viewport-scroll-caps.test.ts` — **10 pass / 0 fail** (29 expect calls, up from 28).
- Mutation check as described above: 2 fail on drifted constants, then green again after restore.
- `gui/src/styles.css` byte-identical to `origin/dev` after the mutation experiment (sha256 compared).
- `git diff --check` clean; MD040 clean across the unit.

No screenshot: the only `gui/src` change is a comment, and the CSS declarations are untouched — verified by the byte-identical hash above. Rendered evidence for the unit is in #2905, #2906 and #2911.

## Checklist

Two devlog files, one stylesheet comment, one test file; no runtime, routing, auth, or release paths are touched.

---

## Round 2 findings, also fixed

A second independent round returned NEAR-PASS: five claims verified, two partial. Both partials were correct.

**The overstated claim was mine.** Fixing the drifted citations I wrote that *"every original citation now lands on unrelated CSS"*. Three of four do — `:2003` is now `min-width: 220px`, `:1222` is a `z-index`, `:2198` is a `background` — but `:755` still lands exactly on `.action-toast`'s `max-width`. One of four survived by coincidence and nothing marked which, so the note now says that rather than claiming all four rotted.

**The remaining test gap was worse than the one locking constants closed.** `allRuleBodies()` concatenates every body for a selector and `.match()` takes the *first* hit — so a second rule for the same selector, added later with a wrong value, wins the cascade while the original correct declaration still satisfies the assertion. Green test, broken toast.

Both assertions now read the **effective** declaration (the last one across all bodies) and anchor the regex to the whole value. Proved against the reviewer's exact scenario — appending

```css
.action-toast.notice { max-width: min(200px, calc(100vw - 8px)); }
.logs-table-wrap { max-height: calc(100dvh - 900px); }
```

turns both tests **red** (`Expected: 480, Received: 200`) where the first-match logic passed. The 479/47/261 value mutation is still caught too. `gui/src/styles.css` restored byte-identical to `origin/dev` after each experiment (sha256 `39425844a41cba50`), suites at 10 pass / 0 fail.

---

## Round 3 finding, also fixed

Round 3 **refuted** my round-2 fix, and it was right — the bug was mine. Reading a property with an unanchored name also matches inside a **custom property**, so this valid stylesheet passed while both rendered caps were wrong:

```css
.logs-table-wrap { max-height: calc(100dvh - 261px); --max-height: calc(100dvh - 260px); }
.action-toast.notice { max-width: min(479px, calc(100vw - 47px)); --max-width: min(480px, calc(100vw - 48px)); }
```

Reproduced against the real stylesheet before fixing: 10 pass / 0 fail with both effective declarations wrong. The property must now start the body or follow `;`/newline, so a `--`-prefixed name cannot satisfy it. Re-running the reviewer's exact fixture now turns both tests red.

**Both false negatives are now fixtured**, so neither can return quietly. A new test feeds inline CSS to the helper directly: a custom property whose name contains the read property must not be reported; a later duplicate rule must win; a genuinely absent property must throw. Driven red first — reverting only the regex fails it with `Received: calc(100dvh - 260px)`, the masked value.

The helper's comment also **overclaimed**, which the reviewer flagged alongside the code. It said the helper returns the declaration that "actually wins" and describes "the rendered result". It resolves source order within one exact selector string; it does not model `!important`, competing specificity, or @-rule nesting. The comment now states that, and why it is enough here: each rule is declared once, and the cascade question that matters (`.notice` beating a single-class `.action-toast`) is asserted separately.

Suites now **11 pass / 0 fail** (32 expect calls). `gui/src/styles.css` byte-identical to `origin/dev` after every experiment.

---

## Round 4 findings, also fixed

Round 4 confirmed the custom-property hole is closed and both fixtures genuinely guard their regressions, then found two more.

**CSS property names are case-insensitive and mine was not.** `MAX-HEIGHT` is a real declaration that wins the cascade, but a case-sensitive match walked past it and reported the earlier lowercase value. The reviewer confirmed in a browser that Chrome honours both `MAX-HEIGHT` and the escaped `max\2d width`. Matching is now case-insensitive, and escaped identifiers are **refused** rather than skipped — `max\2d height` *is* `max-height`, and resolving it means CSS identifier unescaping, which is more than a source-text oracle should pretend to have. Nothing here writes one, so its appearance means the reader no longer understands the file and should fail loudly.

Both fixtured, both driven red against the real stylesheet first: an uppercase duplicate fails 2 tests, an escaped one fails 1. Each previously passed 3/0.

**The 030 heading was misleading.** It read *"(line 2011 as shipped)"* directly above the pre-fix `100vh` snippet, making the defective code look like the shipped code. Line 2011 is where the rule sits now, carrying `100dvh`. Now stated explicitly.

Suites **12 pass / 0 fail** (34 expects).

## Summary by CodeRabbit

* **Documentation**
  * Clarified dashboard layout behavior, viewport sizing, toast-width rules, and wrapped-content handling.
  * Updated guidance for dynamic viewport sizing and line-height-based layout checks.

* **Tests**
  * Strengthened viewport and scroll-cap checks with exact expected values.
  * Added precise validation for log-table sizing and toast widths.
  * Improved styling checks to account for overlapping rules and prioritize effective declarations.
  * Added coverage for duplicate, case-variant, custom, escaped, and missing style properties.

### DIFF
--- a/devlog/_plan/260829_gui_dashboard_slop/000_baseline_and_roadmap.md
+++ b/devlog/_plan/260829_gui_dashboard_slop/000_baseline_and_roadmap.md
@@ -104,9 +104,12 @@ and `.toast-notice` cap toast width with `calc(100vw - Npx)`, which ignores clas
 scrollbar width.
 
 Rules are named by selector rather than line number on purpose: the fix itself
-inserted lines above them, so every original citation (`styles.css:2003`,
-`:755`, `:1222`, `:2198`) now lands on unrelated CSS. Current locations are in
-`030` and the Outcome section below.
+inserted lines above them, so most of the original citations no longer describe
+what they pointed at. `:2003` is now `min-width: 220px`, `:1222` is the toast
+host's `z-index`, and `:2198` is `background: var(--glass-rail)`. `:755` still
+happens to land on `.action-toast`'s `max-width`, which is the point rather than a
+reprieve: one of four survived by coincidence, and nothing marks which. Current
+locations are in `030` and the Outcome section below.
 
 The probe measures this behaviourally — comparing each scroll container's
 computed cap against `visualViewport.height` — rather than grepping for the
--- a/devlog/_plan/260829_gui_dashboard_slop/000_baseline_and_roadmap.md
+++ b/devlog/_plan/260829_gui_dashboard_slop/000_baseline_and_roadmap.md
@@ -51,10 +51,16 @@ stretches them to equal height. Below ~`36rem` of *card* width the container
 query gives copy and controls `flex-basis: 100%`, so each card becomes two
 wrapped flex lines. The two cards then have **equal outer height but different
 content height** — vision's control column is taller (select row + advanced
-disclosure). Flexbox distributes the leftover space of each card independently
-and `align-items: center` centres each line within its own leftover, so the
-shorter card's control row sinks by half the difference. Nothing ties one card's
-second line to the other's.
+disclosure). Flexbox distributes the leftover space of each card independently,
+so the shorter card's control row sinks. Nothing ties one card's second line to
+the other's.
+
+> **Corrected during implementation.** This paragraph originally blamed
+> `align-items: center`. That is the wrong property: `align-items` centres each item
+> *within* its line, while the mis-distributed thing is the **lines**, which is
+> `align-content` — defaulting to `stretch` on a multi-line flex container. The fix
+> is `align-content: start`; `align-items: center` stays and is what keeps the
+> single-line (one-column) regime centred. See `013`.
 
 The shipped mitigation is a hard-coded reserved band:
 
@@ -89,12 +95,18 @@ third card is added to either grid.
 
 ## Defect 3 — static viewport units in scroll surfaces
 
-`gui/src/styles.css:2003` caps `.logs-table-wrap` with
+**As measured before the fix.** `.logs-table-wrap` capped with
 `max-height: calc(100vh - 260px)`. Static `vh` resolves against the *large*
 viewport, ignoring mobile browser chrome, while the rest of the shell already
-uses `100dvh` (styles.css:244, 247, 411, 412, 2198). The log table is therefore
-sized for a viewport the user cannot see. `styles.css:755` and `1222` cap toast
-width with `calc(100vw - Npx)`, which ignores classic scrollbar width.
+used `100dvh` (`.app`, the sidebar, `.main-inner--combos`, the mobile drawer). The
+log table was therefore sized for a viewport the user cannot see. `.action-toast`
+and `.toast-notice` cap toast width with `calc(100vw - Npx)`, which ignores classic
+scrollbar width.
+
+Rules are named by selector rather than line number on purpose: the fix itself
+inserted lines above them, so every original citation (`styles.css:2003`,
+`:755`, `:1222`, `:2198`) now lands on unrelated CSS. Current locations are in
+`030` and the Outcome section below.
 
 The probe measures this behaviourally — comparing each scroll container's
 computed cap against `visualViewport.height` — rather than grepping for the
--- a/devlog/_plan/260829_gui_dashboard_slop/030_dynamic_viewport_units.md
+++ b/devlog/_plan/260829_gui_dashboard_slop/030_dynamic_viewport_units.md
@@ -14,7 +14,8 @@
 
 ## Defect
 
-`.logs-table-wrap` in `gui/src/styles.css` (line 2011 as shipped):
+`.logs-table-wrap` in `gui/src/styles.css` — **as it was before the fix**; the
+rule now sits at line 2011 and carries `100dvh`:
 
 ```css
 .logs-table-wrap { max-height: calc(100vh - 260px); }
--- a/devlog/_plan/260829_gui_dashboard_slop/030_dynamic_viewport_units.md
+++ b/devlog/_plan/260829_gui_dashboard_slop/030_dynamic_viewport_units.md
@@ -6,8 +6,11 @@
 > pass found both.
 >
 > Rules below are named by selector, not line number: the fix this document
-> describes inserted lines above the very rules it cites, so the original
-> citations (`styles.css:2003`, `:755`, `:1222`) now land on unrelated CSS.
+> describes inserted lines above the very rules it cites, so most of its original
+> citations stopped describing what they pointed at — `styles.css:2003` is now
+> `min-width: 220px` and `:1222` is a `z-index`. `:755` still lands on
+> `.action-toast`'s `max-width` by coincidence, which is why the selector is the
+> reference and the line is only a hint.
 
 ## Defect
 
--- a/devlog/_plan/260829_gui_dashboard_slop/030_dynamic_viewport_units.md
+++ b/devlog/_plan/260829_gui_dashboard_slop/030_dynamic_viewport_units.md
@@ -23,10 +23,12 @@ rows under the browser UI. The rest of the shell already moved to `100dvh` —
 `.app` (244), the sidebar (247), `.main-inner--combos` (411-412) and the mobile
 drawer (2213) — so this line is an outlier, not a convention.
 
-`.action-toast` (749) and `.notice` (1215) cap toast width with `calc(100vw - Npx)`. Per CSS
-Values and Units 4, `100vw` includes the classic scrollbar gutter, so a
-scrollbar-reserving platform can in principle render a cap wider than the visible
-area.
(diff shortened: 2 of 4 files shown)


lidge-jun:
## 리뷰 · 우선순위 54 / 80

이 PR은 코드를 거의 안 고친다. 지금 `dev` HEAD는 `cfb70c972` (`#2913`) 이다. 그 앞에 `#2912` (`a0a8e8e2b`) 가 `260829_gui_dashboard_slop` 유닛의 Outcome 을 이미 닫았다. `#2905` 는 사이드카 짝 맞춤, `#2906` 은 로그 표 `dvh` 와 토스트 캐스케이드, `#2911` 은 020 철회를 측정으로 뒷받침했다. 감사자가 그 기록에서 틀린 인용 네 곳과, 테스트가 아무 양수나 통과시키던 한 곳을 짚었다. 이 PR은 그 다섯을 고친다. `gui/src/styles.css` 선언은 그대로다.

지금 HEAD의 `gui/src/styles.css` 를 보면 `.action-toast` 는 749행에서 시작하고 755행이 `min(480px, calc(100vw - 48px))` 이다. 같은 식을 `.action-toast.notice` 770-774행이 다시 적는다. `.notice` 1215행은 `max-width: var(--prose-measure)` 이고 뷰포트 단위가 없다. `.toast-notice` 는 1229행에서 시작하고 1232행이 `min(var(--prose-measure), calc(100vw - 32px))` 이다. `#2911` 이 `.notice` 를 두 번째 calc 규칙이라고 적은 것은 틀렸다. 캐스케이드에서 `.notice` 가 이기는 이유 자체가 그 규칙이 뷰포트 식이 아니기 때문이다. 이름을 `.toast-notice` 로 고친 것은 HEAD와 맞다.

줄 번호도 이미 어긋나 있다. `.logs-table-wrap` 는 2011-2018행이고 값은 `calc(100dvh - 260px)` 이다. 예전 인용 `:2003` 은 이제 `.logs-filter-field .input` 의 `min-width: 220px` 다. `:1222` 는 `.toast-notice-host` 의 `z-index` 다. `:2198` 은 모바일 탑바 `background: var(--glass-rail)` 이다. `:755` 만 우연히 아직 `.action-toast` 너비다. 000이 네 개 중 하나만이 살아 남았고 표시가 없다고 적은 것은 맞다. 선택자 이름으로 바꾸는 쪽이 맞다.

원인은 `align-content` 다. HEAD `gui/src/styles-dashboard-workspace.css` 187-199행 주석과 199행 `align-content: start` 가 그렇게 말한다. `align-items: center` 는 한 줄 안에서 아이템을 가운데 맞춘다. 여러 줄이 남긴 공간을 나누는 것은 `align-content` 이고 기본값은 stretch 다. 한 줄짜리 카드에서는 `align-content` 가 아무 일도 안 하므로 `align-items: center` 는 그대로 둔다. 000이 틀린 문장을 지우는 대신 고친 이유를 남긴 것은 조사 기록을 지키는 쪽이다. 240행 `min-height: 3lh` 주석이 회귀 테스트가 그려진 줄 수를 잰다고 한 것도 과하다. 커밋된 테스트는 소스 텍스트만 본다. happy-dom 은 레이아웃을 안 한다. 세 줄이라는 숫자는 013의 브라우저 측정이다. 주석을 나눈 것은 맞다.

테스트 구멍이 더 크다. 지금 HEAD `gui/tests/viewport-scroll-caps.test.ts` 55-56행은 캡처한 숫자가 0보다 크기만 하면 통과한다. 1px 설계 너비도 통과한다. 36-37행은 `100dvh` 단위만 보고 빼는 값 260은 안 본다. 48-49행은 본문을 이어 붙인 뒤 첫 번째 `max-width` 를 읽는다. 같은 선택자로 나중에 틀린 값을 넣으면 화면은 깨지고 테스트는 예전 값을 보고 초록이 된다. 이 PR의 `effectiveDeclaration` 은 같은 선택자의 마지막 선언을 읽는다. 480, 48, 260을 숫자로 고정한다. 본문이 479/47/261 과 뒤에 붙인 200px/900px 돌연변이로 빨개졌다가 시트를 되돌렸다고 한다. types.ts/config.ts 분할과 무관하다. close-don't-rebase 대상이 아니다.

라인 030 `.action-toast` / `.toast-notice` - 고친 이름은 맞다. 그래도 둘 다 맨 calc 가 아니라 min(...) 이다. `.action-toast` 는 48px, `.toast-notice` 는 32px 이라 N 도 다르다.
라인 viewport-scroll-caps.test.ts effectiveDeclaration - 같은 선택자의 마지막 값만 본다. `.notice` 처럼 다른 선택자가 이기고 지는 캐스케이드는 이 함수가 모른다. 이번 회귀는 같은 선택자 중복이라 그 범위로 충분하다.
경로 gui/tests/sidecar-layout.test.ts - 빈 `@container` 매칭이 단언을 건너뛰는 약점은 본문도 인정한다. 이번 유닛 고침과 별개라 안 넣은 것은 맞다.
경로 styles.css 선언 - 바이트가 같다고 했으니 런타임 회귀면은 없다. 테스트가 잠근 것은 이미 들어간 `#2906` 값이다.

메인테이너의 판단이 필요한 지점
- 030의 calc 짧은 말을 min(...) 과 48/32 로 고친 뒤 머지할지, 이름 수정만으로 충분할지
- sidecar `@container` 빈 매칭을 이어서 잠글지, 다음 유닛으로 남길지
- `#2912` 가 이미 Outcome 을 닫았으니 이 정정 PR을 같은 유닛 후속으로 넣을지

너의 추천
머지하세요. 기록 오류와 허술한 단언은 둘 다 실재하고, CSS 동작은 안 바뀐다. 030의 min/48/32 한 줄은 있으면 더 정확하지만 막아둘 정도는 아니다. sidecar 빈 매칭은 이번과 섞지 마세요.

이 댓글은 grok-bot이 작성했습니다

Issue #2914, opened 2026-08-29
Windows (zh-CN): ocx sync always refuses — localized schtasks output breaks the task-absence probe, and the 2s full-listing fallback times out


ncepuee:
### Client or integration

Codex CLI (`ocx sync` on a desktop install; the same refusal surfaces in the dashboard as the failed sync message)

### Area

Platform (Windows / macOS / Linux)

### Summary

On a Simplified Chinese Windows host (console code page 936/GBK), with the proxy running manually — no `opencodex-proxy` scheduled task, no WinSW service — `ocx sync` deterministically refuses to write the Codex config:

```
Refusing to write because ownership could not be proven: Task Scheduler could not be asked whether opencodex-proxy is registered.
```

"Absent" is the expected ownership answer in this state (no service, no task), but the probe cannot prove it. Two gaps compound:

1. **The task-not-found check is English-only.** `SCHTASKS_TASK_NOT_FOUND_EN` (`src/service-manager-probe.ts`) matches only `/cannot find the file specified/i`. On zh-CN, `schtasks /query /tn opencodex-proxy /xml` exits 1 with the GBK bytes of `错误: 系统找不到指定的文件。` — and `decodeWindowsTextBytes` cannot decode those bytes either, because `legacyEncodingForLocale` (`src/lib/windows-text.ts`) only maps `ko → euc-kr` and the windows-1252 family. Chinese (and Japanese) bytes fall through to the UTF-8 fallback and arrive as mojibake, so the regex can never match.
2. **The locale-neutral fallback cannot finish in its budget.** The bounded full listing `schtasks /query /fo CSV /nh` is capped by `SERVICE_PROBE_TIMEOUT_MS = 2000`. Measured on this host: **401 scheduled tasks, 12.3 s** for the full listing — `spawnSync` returns `ETIMEDOUT` inside the probe, `registered: "unknown"`, and the write is refused.

[...]

PR #2926, opened 2026-08-29
fix(codex): share one identity-lookup budget between desktops and CI

lidge-jun:
## Summary

The Windows identity lookup carried two budgets: 8 s on a desktop, 30 s under `CI`. The split assumed contention only happens on a shared runner. It does not.

A zh-CN Windows 10 host measured **3.2 s** for the SID expression and **4.6 s** for the `Add-Type` LocalAppData expression — per spawn, with a bare `powershell.exe -NoProfile -Command exit` costing ~3 s where a typical desktop pays ~150 ms. Two spawns plus ordinary jitter breached 8 s intermittently, and `ocx sync` failed with `CodexUserIdentityRefusal: Windows effective-account lookup timed out.`

That is the same contention the CI branch was widened for, which is what makes the split **vestigial** rather than merely conservative: antivirus real-time scanning, a loaded Task Scheduler, and cold PowerShell startup do not care whether the machine is a runner. One budget now, 30 s, everywhere.

The contract the budget protects is unchanged. A genuinely hung child still fails the lookup and the caller still refuses to write; only the ceiling moved, and it moved for the case where the lookup *would have succeeded*.

## Credit

This is the second half of #2925 by @ncepuee. That PR also re-fixed the schtasks probe, which had already landed in aaccef738 (#2918), so it conflicted with `dev` and I closed it — but the identity-budget argument was theirs and was the part I had explicitly deferred as out of scope when I fixed the probe. Their observation that the CI gate existed *only* to widen this constant is what turns "raise the number" into a defensible removal. I invited a rebase and offered to carry it; landing it here so the defect does not sit open in the meantime.

## Verification

- `bun test tests/codex-user-identity.test.ts` — **7 pass / 0 fail**.
- **Mutation-proven.** Reintroducing the split (`process.env.CI === "true" ? 30_000 : 8_000`) turns the new test red: `Expected: 30000, Received: 8000`. Restored and verified byte-identical by sha256; suite back to 7/0.
- `bun x tsc --noEmit` clean, `bun run privacy:scan` passed.

The test reads the budget through the existing `windowsIdentityPowerShellSpawnOptionsForTests` seam with `CI` set and unset, rather than asserting the constant. That distinction is the point: a constant can be correct while a reintroduced branch silently restores the split, and only the spawn-options readback catches that.

[...]

PR #2924, opened 2026-08-29
fix(claude): detect Windows Desktop policy conflicts

lidge-jun:
## Summary

- Fixes #2893 by probing `HKLM\\SOFTWARE\\Policies\\Claude` with a bounded, read-only `reg.exe` query resolved from the trusted Windows System32 directory.
- Reports policy state as `present | absent | unknown | not_applicable`. `present` and `unknown` make Claude Desktop 3P health non-green and provide an actionable warning in apply, Desktop status, and `ocx status`; registry value names/data never leave the probe.
- Preserves non-OpenCodex keys in an existing owned Desktop profile while replacing only the OpenCodex-owned gateway/model subtree.
- Documents `ocx claude desktop status`, policy warnings, and merge preservation.

Closes #2893

## Verification

- `bun test tests/claude-desktop-policy.test.ts tests/desktop-3p.test.ts tests/native-claude-desktop-toggle.test.ts tests/claude-desktop-cli.test.ts tests/cli-status-json.test.ts tests/cli-help.test.ts tests/skill-ocx.test.ts` — 91 pass, 0 fail.
- `bun x tsc --noEmit` — passed.
- `bun run privacy:scan` — passed.
- `(cd docs-site && bun install --frozen-lockfile && bun run build)` — 401 pages built; passed with the existing large-chunk advisory.
- Rebased onto `origin/dev` at `64b994c0f6330c4bc56e85138037935859de48e4`; post-rebase verification above passed.

Mutation proof (each source mutation was restored immediately after the named RED run):

1. Policy present -> health warning: changed the successful query result from `present` to `absent`.

```text
$ bun test tests/claude-desktop-policy.test.ts -t "a present Windows Claude policy degrades Desktop 3P health"
Expected: "present"
Received: "absent"
(fail) a present Windows Claude policy degrades Desktop 3P health
0 pass, 1 fail
```

2. Unreadable policy -> `unknown`, never `absent`: changed the readable-parent/blocked-child result to `absent`.

```text
$ bun test tests/claude-desktop-policy.test.ts -t "an unreadable Windows Claude policy stays unknown and degrades health"
Expected: "unknown"
Received: "absent"
(fail) an unreadable Windows Claude policy stays unknown and degrades health
0 pass, 1 fail
```

3. Proven missing policy -> `absent`: changed the readable-parent/missing-child result to `unknown`.

```text
$ bun test tests/claude-desktop-policy.test.ts -t "a missing policy is absent only after its parent is readable"
Expected: "absent"
[...]

PR #2922, opened 2026-08-29
test(codex): prove replay 402 closes compact recovery

lidge-jun:
## Summary

- `origin/dev` already contains the gap-5 production fix from #2897 (`4fa981fb8`), so this is the missing focused regression follow-up rather than a duplicate control-flow rewrite.
- Exercise compact stored-Pool replay failures as separate 429 and 402 cases. Each case counts all upstream sends and proves zero later sends to either an alternate Pool account or the remembered compact model.
- Correct the existing regular Responses 402 regression title, which previously said 429. This is a title-only correction: the fixture and assertion were already 402, so no coverage moved.
- No production source, credential-store state machine, forced-refresh provenance, rejected-bearer exclusion, or affinity handoff is changed. PR #2920 remains separate and untouched.

Refs #2892.

This is a credential/auth path and requires explicit security review before merge. The tests add no logging and do not serialize tokens or account identifiers.

## Verification

- `bun test tests/responses-pool-401-refresh.test.ts tests/codex-account-store.test.ts` — **55 pass / 0 fail** on head `f06e8ca83`.
- `bun x tsc --noEmit` — exit 0.
- `bun run privacy:scan` — **Privacy scan passed**.
- The existing one-refresh/one-replay and next-request-availability regressions remain green.

### Mutation evidence

Regular Responses 402: setting `sameAccountOnly: false` allowed the alternate account send and turned the existing corrected-title test RED:

```text
bun test v1.4.0 (34cbb9a40)

tests/responses-pool-401-refresh.test.ts:
683 |     expect(response.status).toBe(402);
                                  ^
error: expect(received).toBe(expected)

Expected: 402
Received: 200

(fail) ordinary pool 401 refresh and replay (#2887) > a stored-account replay 402 cannot reach another account even when one is eligible

0 pass
19 filtered out
1 fail
```

New compact 402 case, alternate-account mutation: removed the exact `&& !storedPool401ReplayAttempted` guard at `src/server/responses/compact.ts:846` from the 429/402 alternate-account `if`; this turned it RED:

```text
bun test v1.4.0 (34cbb9a40)

tests/responses-pool-401-refresh.test.ts:
446 |       expect(response.status).toBe(replayStatus);
                                    ^
error: expect(received).toBe(expected)

Expected: 402
Received: 200

(fail) ordinary pool 401 refresh and replay (#2887) > compact does not compose a stored-account replay 402 with another account or remembered model

0 pass
20 filtered out
1 fail
```

[...]

PR #2927, opened 2026-08-29
fix(service): enumerate the machine once per startup, not once per inspection

### LINKED ISSUE #2923: [Bug]: Windows startup can run the 20 s schtasks fallback twice before listen
## Confirmed current-dev regression

Current dev 64b994c0f includes #2918. On Windows, a localized nonzero targeted schtasks query falls back to the full task listing with a 20 second ceiling.

startServer currently calls inspectStartupOwnership before startup cache invalidation at src/server/index.ts around line 609 and again before native-main lifecycle preparation around line 788. Both calls reach the same synchronous service-manager probe.

On the reported zh-CN host, the full listing measured about 12.3 seconds for 401 tasks. With no installed OpenCodex task, one proxy start can therefore spend about 25 seconds in two identical listings before Bun.serve starts, with a 40 second worst-case ceiling.

The second ownership check is intentional. It prevents startup work between the cache decision and native lifecycle preparation from widening the service-install race. The fix must not simply delete final revalidation, lower the listing timeout back to 2 seconds, add guessed absence, or convert an unknown result into absent.

## Required scope

- Keep each final ownership decision fail-closed.
- Keep the targeted scheduled-task query bounded and revalidated.
- Avoid repeating the expensive full-machine enumeration during one startup when the targeted recheck still provides the race-sensitive evidence.
- Do not broaden this into service installation or GUI changes.

## Regression coverage

Add an injected Windows startup probe that:

1. makes the targeted query return a localized unreadable nonzero result,
2. makes the full listing take the fallback path,
3. proves one start does not run the full listing twice,
4. proves the targeted ownership recheck still occurs, and
5. proves a task that becomes present between checks is not reported absent.

Original review evidence: https://github.com/lidge-jun/opencodex/pull/2920#discussion_r3886900298

lidge-jun:
## Summary

A regression from my own #2914 fix, correctly diagnosed in #2923.

That fix gave the whole-machine `schtasks /query /fo CSV /nh` listing a 20 s budget, because with the localized message decoded it is still the only *locale-neutral* absence evidence. What I missed is that `startServer` inspects ownership **twice** on purpose — once before the startup cache decision, once before native lifecycle preparation — so a localized host with hundreds of tasks paid the full enumeration at both sites. The reporter measured ~12.3 s per listing on a 401-task host: about **25 s of startup, 40 s at the ceiling, all of it before `Bun.serve`**.

The listing is now shared across one startup through a cache object the caller owns. `startServer` creates it once and hands it to both inspections.

## Why the caller owns the cache

I tried module-level memoization first and it was wrong, in a way the existing suite caught immediately: state leaked across unrelated inspections, and in tests an access-denied host started reporting `present` instead of `unknown` — a fail-open direction. The lifetime that is actually correct here is "one startup," and only the caller knows when that is. An explicit object makes the sharing visible at both call sites and impossible anywhere else; omit it and every inspection enumerates independently, exactly as before.

## What is shared, and what deliberately is not

Only the listing, and only which task names the machine reported.

- **The targeted `/query /tn` still runs on every inspection.** That is the evidence that decides `present`, and it is never cached — so a task registered *between* the two checks is still seen, and the race the second inspection exists to close is untouched. The issue was explicit that the fix must not delete the revalidation, and it does not.
- **A stalled or failed listing is not cached.** Caching a failure would convert one transient stall into an unprovable ownership for the rest of the startup, which is the direction that refuses writes.
- The 20 s budget, the fail-closed `unknown`, and the absence of any guessed `absent` are all unchanged.

## Verification

- `bun test tests/codex-service-manager-probe-hardening.test.ts tests/codex-service-manager-probe.test.ts tests/service-probe-docker.test.ts` — **77 pass / 0 fail**.
- `bun test tests/core-lab-boundary.test.ts` — 17 pass (this touches `src/server/index.ts`, which that guard governs).
- `bun x tsc --noEmit` clean, `bun run privacy:scan` passed.

### Three mutations, each red on its own

| mutation | result |
| --- | --- |
| ignore the shared listing | `Expected length: 1, Received length: 2` — double enumeration returns |
| cache the stalled listing too | `Expected: "absent", Received: "unknown"` — sticky refusal |
| let the cached listing preempt the targeted query | `Expected: not "absent"` — a newly registered task is masked |

The third is the one worth having. It is the fail-open failure mode this change could plausibly introduce, and it is exactly what #2923 warned against; the source was restored and verified byte-identical by sha256 after each.

**What I could not verify:** I am on macOS, so no real `schtasks` ran and I did not reproduce the 12.3 s or ~25 s measurements. The call-count and caching behavior are fully covered by injected probes; the wall-clock improvement follows from removing one enumeration and rests on the reporter's timing.

## Checklist

This is an ownership-admission path. Every final decision stays fail-closed, no cached value can produce `absent` on its own, and `present` still requires the live targeted query. No task names, registry data, or account identifiers are logged.

Closes #2923

## Summary by CodeRabbit

* **Performance**
  * Reduced Windows startup overhead by reusing service inspection results during repeated startup checks.
  * Expensive scheduled-task listings now run only once when possible.
* **Reliability**
  * Targeted task checks continue to run for each inspection.
  * Failed or timed-out listings are not reused, helping preserve accurate results.
* **Tests**
  * Added coverage for cache reuse, task registration changes, targeted queries, and listing failures.

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -31,6 +31,10 @@ import {
   type NativeCodexOwnership,
   type OwnershipInspection,
 } from "../integrations/native/ownership-preflight";
+import {
+  createServiceProbeStartupCache,
+  type ServiceProbeStartupCache,
+} from "../service-manager-probe";
 import { registerCodexCooldownRecoveryProbeWorker } from "../codex/auth-api";
 import {
   reconcileLiveStateStores,
@@ -499,6 +503,7 @@ function inspectStartupOwnership(
   deps: StartServerDeps,
   currentHomes: ReturnType<typeof currentServiceHomes> | null,
   statePaths: readonly string[] | null,
+  startupCache?: ServiceProbeStartupCache,
 ): OwnershipInspection {
   try {
     if (currentHomes === null || statePaths === null) {
@@ -510,7 +515,7 @@ function inspectStartupOwnership(
     if (deps.inspectNativeCodexOwnership) {
       return deps.inspectNativeCodexOwnership({ currentHomes, statePaths });
     }
-    return inspectNativeCodexOwnership({ currentHomes, statePaths });
+    return inspectNativeCodexOwnership({ currentHomes, statePaths, startupCache });
   } catch {
     return {
       ownership: "unknown",
@@ -606,10 +611,21 @@ export function startServer(port?: number, deps: StartServerDeps = {}): Server<W
     startupOwnershipHomes = homes;
     startupOwnershipStatePaths = statePaths;
   } catch { /* inspection below stays unknown */ }
+  /*
+   * One shared listing across this startup's ownership inspections (#2923).
+   *
+   * Both inspections below run the targeted scheduled-task query, so the race the
+   * second one exists to close is unaffected. What they no longer do is enumerate
+   * every task on the machine twice: on a localized host with hundreds of tasks the
+   * fallback listing is the expensive step, and paying it at both sites put up to
+   * 40s in front of `Bun.serve`.
+   */
+  const serviceProbeStartupCache = createServiceProbeStartupCache();
   const startupCacheOwnership = inspectStartupOwnership(
     deps,
     startupOwnershipHomes,
     startupOwnershipStatePaths,
+    serviceProbeStartupCache,
   );
   // Startup cache invalidation is best-effort and must never block the server from
   // serving. It now takes K so it cannot race a convergence commit. Use the home
@@ -785,7 +801,12 @@ export function startServer(port?: number, deps: StartServerDeps = {}): Server<W
   // clients; no Codex request can use this lifecycle in that state.
   // Re-probe here instead of trusting the earlier cache decision: startup work
   // between the two sites must not widen the service-install race.
-  const nativeOwnership = inspectStartupOwnership(deps, startupOwnershipHomes, startupOwnershipStatePaths);
+  const nativeOwnership = inspectStartupOwnership(
+    deps,
+    startupOwnershipHomes,
+    startupOwnershipStatePaths,
+    serviceProbeStartupCache,
+  );
   const preparedNativeMainLifecycle = nativeOwnership.ownership !== "foreign"
     && startupOwnershipHomes !== null
     ? prepareNativeMainStartupLifecycle(
--- a/src/service-manager-probe.ts
+++ b/src/service-manager-probe.ts
@@ -45,6 +45,36 @@ export const SERVICE_PROBE_TIMEOUT_MS = 2_000;
  */
 export const SERVICE_PROBE_LISTING_TIMEOUT_MS = 20_000;
 
+/**
+ * Caller-owned reuse of the whole-machine `schtasks` enumeration (#2923).
+ *
+ * The listing is the locale-neutral fallback, and on a host with hundreds of tasks
+ * it costs seconds. `startServer` inspects ownership twice on purpose — once
+ * before the startup cache decision, once before native lifecycle preparation — so
+ * a localized host paid the full enumeration TWICE before `Bun.serve`, up to 40s
+ * at the ceiling.
+ *
+ * The scope is deliberately the CALLER'S, not this module's. A module-level cache
+ * would leak across unrelated inspections (and, in tests, across cases), and the
+ * lifetime that is actually correct here is "one startup" — something only the
+ * caller knows. Passing an explicit object makes the sharing visible at both call
+ * sites and impossible anywhere else.
+ *
+ * What is shared is only the listing. The targeted `/query /tn` still runs on
+ * every inspection, so the race the second check exists to close is unaffected: a
+ * task that appears between the two checks is seen by the targeted query, which is
+ * what decides `present`. The listing only ever answers "our task is not among
+ * this machine's tasks", and only after the targeted query was not decisive.
+ */
+export interface ServiceProbeStartupCache {
+  /** Task names the machine reported, lowercased; absent until a listing succeeds. */
+  listedTaskNames?: ReadonlySet<string>;
+}
+
+export function createServiceProbeStartupCache(): ServiceProbeStartupCache {
+  return {};
+}
+
 export type ServiceManagerBackend = "launchd" | "systemd" | "scheduler" | "winsw";
 
 export interface ServiceManagerClaim {
@@ -139,6 +169,12 @@ export interface ProbeDeps {
   readonly winswStatus?: () => "started" | "stopped" | "nonexistent" | "unknown";
   /** Test seam for redirected Windows legacy-codepage output. */
   readonly windowsLocale?: string;
+  /**
+   * Opt-in reuse of the expensive task listing across the inspections of ONE
+   * startup. Omit it and every inspection enumerates independently, exactly as
+   * before.
+   */
+  readonly startupCache?: ServiceProbeStartupCache;
 }
 
 const LABEL = "com.opencodex.proxy";
@@ -552,6 +588,16 @@ function windowsTaskListContains(body: string, taskName: string): boolean {
   });
 }
 
+/** The same normalization {@link windowsTaskListContains} applies, kept for reuse. */
+function windowsTaskListNames(body: string): ReadonlySet<string> {
+  const names = new Set<string>();
+  for (const line of body.split(/\r?\n/)) {
+    const field = csvFirstField(line).replace(/\//g, "\\").replace(/^\\+/, "");
+    if (field.length > 0) names.add(field.toLowerCase());
+  }
+  return names;
+}
+
 /**
  * The English message: a fast path on an English host, and nothing more.
  *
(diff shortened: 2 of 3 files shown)


Ingwannu (changes requested):
Requesting changes on exact head `414fae069110db9cfe9fab54de23d5b0da6686ce`.

Two blockers remain:

1. `src/service-manager-probe.ts:655-659` can reuse a stale negative listing after the live targeted query has changed to an undecidable result. A concrete sequence is: the first targeted query reports localized not-found; the first successful listing contains no OpenCodex task and caches an empty set; a restricted task is then registered; the second targeted query returns access denied or another nonzero response; the cached miss is converted to `absent`. A fresh listing would have found the task and returned `unknown`, so this is a fail-open ownership result that can admit unattended writes. Reuse the listing only when the complete targeted raw result is unchanged, or relist/fail closed when the current targeted evidence differs. Please add this exact warm-cache regression.

2. The production cache wiring is not covered through the actual `startServer` boundary. `inspectStartupOwnership` drops `startupCache` whenever `deps.inspectNativeCodexOwnership` is injected at `src/server/index.ts:515-518`, while the new tests call `inspectServiceManagerInstallation` twice directly. Preserve the cache in the injected inspection contract and add a `startServer` regression proving the two startup ownership inspections share the listing while runtime ownership retries do not inherit the startup cache.

The optimization direction is valid, but cached absence must never outlive the exact targeted evidence that justified it.

lidge-jun:
## 리뷰 · 우선순위 72 / 80

이 PR은 Windows에서 프록시가 듣기 전에 같은 비싼 작업 목록을 두 번 뽑던 문제를 고친다. 지금 `dev` HEAD는 `1d9b389c1` 이다. 바로 앞에 `#2926` 이 들어갔다. 그건 Windows 신원 조회 한도를 데스크톱과 CI 모두 30초로 맞춘 고침이다. 그 아래에는 `#2922` 콤팩트 리플레이 402 테스트, `#2924` Claude Desktop 정책, `#2920` refresh-flight 취소 범위가 있다. 이 PR이 손대는 파일은 `src/server/index.ts`, `src/service-manager-probe.ts`, `tests/codex-service-manager-probe-hardening.test.ts` 세 개다. 신원 조회·콤팩트 리플레이·types.ts/config.ts 분리와는 안 겹친다.

왜 두 번이 문제인가. `#2918` 이 이미 들어가 있다. 그 고침은 zh/ja 호스트에서 `schtasks` 출력을 제대로 읽고, 전체 작업 목록만 `SERVICE_PROBE_LISTING_TIMEOUT_MS = 20_000` 으로 길게 기다린다. 대상 조회 `/query /tn ... /xml` 는 33행의 2초를 유지한다. 전체 목록은 로케일과 상관 없이 `우리 작업이 이 기계에 없다` 를 증명하는 유일한 길이다. 그런데 `startServer` 는 소유권을 일부러 두 번 본다. `src/server/index.ts` 609행은 시작 캐시를 지울지 정할 때, 788행은 네이티브 메인 생명주기를 준비하기 직전이다. 815행의 듣기 이후 재시도도 같은 헬퍼를 쓰지만, 이번 캐시는 그 경로에 안 넘긴다. 이슈 `#2923` 의 zh-CN 호스트는 작업 401개에 목록이 약 12.3초였다. OpenCodex 작업이 없으면 듣기 전에 약 25초, 최악은 40초 천장이 된다.

이 PR의 방법은 호출자가 캐시를 소유하는 것이다. `startServer` 가 `createServiceProbeStartupCache()` 를 한 번 만들고, 609행과 788행 검사에만 넘긴다. 모듈 전역 메모가 아니다. 본문이 말한 대로 모듈 캐시는 검사끼리, 테스트 케이스끼리 새고, 접근 거부 호스트가 `unknown` 대신 `present` 로 열리는 방향이 난다. 캐시에 넣는 것은 성공한 전체 목록의 작업 이름 집합뿐이다. 대상 조회는 매번 다시 한다. 그래서 두 검사 사이에 작업이 생겨도 `/tn` 이 `present` 를 결정할 수 있다. 목록이 멈추거나 실패하면 캐시에 안 넣는다. 한 번의 타임아웃이 시작 끝까지 `unknown` 으로 굳으면, 쓰기를 거절하는 쪽으로 잠긴다. `OwnershipDeps` 가 이미 `ProbeDeps` 를 확장하고 `inspectNativeCodexOwnership` 이 `...deps` 를 `inspectServiceManagerInstallation` 에 펼치기 때문에, `ownership-preflight.ts` 를 안 고쳐도 `startupCache` 가 프로브까지 내려간다. 그 설계는 맞다.

테스트는 세 가지 변이를 잠근다. 공유 목록을 무시하면 `/fo` 호출이 2가 된다. 멈춘 목록을 캐시하면 두 번째 검사가 `absent` 대신 `unknown` 에 남는다. 캐시된 목록이 대상 조회보다 앞이면, 사이에 생긴 작업을 `absent` 로 보고한다. 이슈 `#2923` 이 경고한 실패 방향이 바로 그것이다. types/config 분리 캠페인과 무관하고, preview deploy도 필요 없다. 같은 버그를 Ingwannu 의 `#2928` 도 고친다. 둘을 같이 넣으면 캐시 모양이 싸운다. 하나는 합치고 하나는 leftover 로 닫아야 한다.

라인 - src/server/index.ts 609행과 788행 - 한 시작에 `inspectStartupOwnership` 을 두 번 부른다. 이번 캐시의 수명은 이 두 번의 듣기 전 검사다. 815행 재시도에는 안 넘기는 선택이 맞다.
라인 - inspectStartupOwnership 테스트 시임 - 주입된 `deps.inspectNativeCodexOwnership` 호출은 여전히 `{ currentHomes, statePaths }` 만 넘긴다. 운영 경로만 `startupCache` 를 넣는다. 테스트가 이 시임을 가로채면 공유가 조용히 빠진다. `#2928` 은 시임에도 캐시를 넣는다.
경로/심볼 - tests/codex-service-manager-probe-hardening.test.ts - 증명은 `inspectServiceManagerInstallation` 에 캐시를 직접 넣는 방식이다. `startServer` 가 두 검사에 같은 객체를 넘긴다는 배선은 잠기지 않았다.
경로/심볼 - ServiceProbeStartupCache.listedTaskNames - 성공한 이름 집합만 저장한다. 대상 조회의 바이트가 바뀌어도(예: not-found 다음 access-denied) 옛 목록으로 `absent` 를 낼 수 있다. 흔한 경로는 아니지만 `#2928` 은 대상 조회 지문으로 그 구멍을 막는다.
경로/심볼 - src/ 프로덕션 가드 - 대상 조회가 0으로 성공하면 캐시를 보기 전에 `present` 다. 목록에 이름이 있어도 `present` 로 승격하지 않고 `unknown` 을 유지한다. 목록만으로 소유를 증명하지 않는 HEAD 계약과 같다.

메인테이너의 판단이 필요한 지점
- `#2927` 과 `#2928` 중 하나만 넣을지. 둘은 같은 `#2923` 의 다른 구현이다.
- 멈춘 목록을 다시 시도할지(`#2927`), 같은 대상 조회면 실패도 재사용할지(`#2928`).
- 시임에 캐시를 넣고 `startServer` 배선을 한 테스트로 잠근 뒤에 합칠지, 지금 운영 경로만으로 합칠지.

너의 추천
이쪽으로 합쳐라. 호출자 소유 캐시가 `#2923` 이 시킨 범위와 맞고, 멈춘 목록을 안 굳히는 쪽이 시작 중 일시 실패에 더 안전하다. 합치기 전에 테스트 시임에도 `startupCache` 를 넣고, 가능하면 `#2928` 의 `startServer` 호출 횟수 테스트를 가져오라. CI 의 남은 macos 샤드가 초록이면 `dev` 에 넣고 `#2923` 을 닫아라. `#2928` 은 leftover 로 `Landed via #2927` 댓글과 `landed-via-maintainer` 라벨을 단 뒤 닫아라. 둘 다 넣지 마라.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
Closing in favor of #2928, which solves the same issue with a stricter invariant.

We wrote this in parallel — mine keys the shared listing on "same startup," theirs keys it on the targeted query's exact bytes and status. Theirs is better, and the difference is not stylistic. In my version a cached listing stays valid for the whole startup window, so the guarantee that a newly registered task is still seen rests entirely on the targeted query being consulted first. In #2928 the listing is invalidated the moment the targeted evidence changes at all, which means the absence proof is bound to the evidence that produced it rather than to a time window. That is a real strengthening of the same fix.

One thing from here worth keeping if it is not already covered: a listing that **stalls** must not be cached, or one transient timeout becomes an unprovable ownership for the rest of the startup. #2928 has `a cached listing failure remains fail-closed`, which I read as covering it — @Ingwannu, please confirm that test asserts the *second* inspection retries the listing and can still reach `absent`, not only that the first stays `unknown`. If it only asserts the first, that is the one gap I would want closed before merge.

For the record, the mutations I proved on this branch, in case they are useful as extra cases over there: ignoring the shared listing restores double enumeration; caching the stall produces the sticky refusal above; and letting the cached listing preempt the targeted query masks a task registered between the two checks (`Expected: not "absent"`). The third is the fail-open direction the issue specifically warned about, and #2928's `changed targeted evidence cannot reuse an earlier absence listing` covers it more directly than mine did.

Issue #2892, opened 2026-08-29
Bug: close remaining stored Pool 401 concurrency and recovery-budget gaps


luvs01:
### Client or integration

Codex App and Codex CLI through the OpenCodex Responses/compact routes.

### Area

Authentication and account pool

### Summary

#2889 fixed the primary reported failure: an ordinary stored Pool account now gets one generation-fenced forced refresh and one same-account replay after a pre-stream `401`. It merged as `0f4cd2a0be46178be91689096cd0a00bfeb9337b`.

Five bounded concurrency and logical-request-budget gaps remain in that merged path:

1. A superseding stored generation can be returned without checking that it is fresh and not the rejected bearer.
2. The initiating caller's abort signal is part of the grant-wide refresh flight, so cancelling one waiter can abort live joiners.
3. A rotated refresh grant reaches the owner and active joiners but not an inactive same-account alias that still holds the old grant.
4. Routing validates the credential generation and then mutates health, `needsReauth`, and affinity separately; another process can replace the credential between those operations.
5. After the stored-account refresh/replay budget is consumed, a replay `429`, `402`, or other eligible failure can continue to another Pool account, remembered model, or combo target.

Expected behavior:

- each waiter owns only its own cancellation while the shared refresh uses an internal lifecycle signal;
- a superseding winner is independently fresh and not the rejected bearer;
- rotated grants are CAS-merged into unchanged same-account aliases without overwriting a concurrently newer access credential;
- generation validation and credential side effects are atomic or generation-scoped and discarded when stale;
- the replay response is authoritative for the rest of that logical request, with no later account/model/combo send.

Implementation should stay split into two independent PRs based directly on current `dev`:

1. pooled credential concurrency and generation safety;
2. one recovery budget for a stored Pool `401`.

Do not replace the merged #2889 state machine wholesale. Preserve its forced-refresh provenance, rejected-bearer exclusion, and affinity handoff behavior.

### Reproduction

The gaps are deterministic at the code/interleaving level and were reverified against the exact merge tree:

1. Configure Pool mode with a stored account whose time-valid access token receives a pre-stream `401`.
2. Start two refresh waiters sharing one grant, then cancel the initiating waiter. The current shared `AbortSignal.any(...)` can cancel the token request for the still-live waiter.
3. During the flight, persist a newer but expired G+1 credential. The joined-flight supersession branch can return it without the normal freshness test.
[...]

PR #2920, opened 2026-08-29
fix(codex): scope refresh-flight cancellation to the caller that owns it

lidge-jun:
## Summary

Closes gaps **1 and 2** of #2892 — the two that live in `resolveCodexToken`'s shared-flight logic. This is the first of the two PRs that issue asks for (pooled credential concurrency and generation safety); the recovery-budget half stays separate.

**Gap 2, the one with a user-visible consequence.** Refresh flights are keyed by refresh *grant* and shared: a later caller on the same grant joins the running promise instead of opening its own. The flight's signal was built as `AbortSignal.any([abort.signal, timeout, callerSignal])` — so whichever caller happened to open the flight could abort the token fetch that every other waiter depended on. A joiner has no way to tell that from a genuine upstream failure, so a closed Codex tab could end with a healthy account marked `needsReauth` on behalf of a request that was still running.

The flight now owns its own lifetime: `abort` (stale-flight eviction) and the 30 s ceiling remain, because those bound the flight itself. Callers wait through a new `awaitOwnCancellation`, which honors only their own signal and lets the shared work run to completion for everyone else — the result is still committed, so a joiner that arrives after the initiator walks away is served normally.

Both wait sites use it, which matters: had I only changed the joiner's `await`, the flight owner would have lost cancellation entirely rather than gained the right kind. `resolveCodexToken` also now rejects an already-aborted caller before doing any file I/O.

**Gap 1.** The joined-flight supersession branch returns "someone else already replaced this credential, use theirs". Supersession proves a replacement happened, not that the replacement is usable: an *expired* G+1 passes both the generation test and the rejected-bearer test, and since the caller treats this return as successful recovery, it spends its one replay on a token certain to 401 again. The branch now also requires the stored credential to be fresh.

## Verification

- `bun test tests/codex-account-store.test.ts tests/responses-pool-401-refresh.test.ts` — **55 pass / 0 fail**. Every #2887/#2889 test stays green, including the one-refresh/one-replay collapse and the affinity handoff.
- **Gap 2 is mutation-proven.** Restoring `callerSignal` to the shared flight signal turns the new test red; the joiner receives the abort instead of the rotated credential. The test parks the owner's flight inside `fetch`, admits a joiner, aborts the owner, and asserts three things: the owner's promise rejects, `init.signal.aborted` is false when the fetch resumes, and the joiner gets `rotated` with exactly one token call.
- `bun x tsc --noEmit` clean, `bun run privacy:scan` green.

[...]

PR #2932, opened 2026-08-29
fix(upstream): apply fresh-connection recovery on the sidecar and loop retry legs

lidge-jun:
## Summary

`applyUpstreamRecoveryInit` exists for one reason: Bun has ignored a bare hop-by-hop `Connection: close` (oven-sh/bun#20492), so abandoning a half-closed pooled socket needs the transport-level `keepalive: false` the helper adds. The main lanes call it — `src/server/chat-native.ts:207`, `src/server/responses/compact.ts:715`, and six sites in `src/server/responses/core.ts`. Nine legs did not.

The web-search and images loops *received* the `retryRecovery` argument `fetchWithResetRetry` hands them, spent it on `onAttemptSend` telemetry, and then built a plain init. The seven sidecar and vision executors passed zero-argument thunks: they still retried, but could not ask for fresh-connection recovery.

The consequence, stated precisely: each replay on those legs stayed *eligible* to reuse the pooled socket the reset came from — not guaranteed to, since the pool may hand out another. That is enough to make recovery a matter of luck, and the retry then surfaces as exhausted rather than as the pool problem it is.

Sites fixed: `src/web-search/loop.ts`, `src/images/loop.ts`, `src/web-search/{executor,anthropic-executor,gemini-executor,exa-executor,xai-executor}.ts`, `src/vision/{describe,anthropic-describe}.ts`. `src/adapters/kiro-retry.ts` already hand-rolls the same two fields and is untouched.

**Composition order matters.** On the one leg that also pins a provider HTTP version, recovery nests *inside* `withUpstreamHttpVersion`:

```ts
withUpstreamHttpVersion(url, applyUpstreamRecoveryInit(baseInit, recovery), provider)
```

so the recovery helper always receives a defined init and the version helper spreads the result. The reverse nesting needs a `?? baseInit` fallback to type-check at all, and drops one of `protocol` / `keepalive` otherwise.

**One observable changed, and it broke three tests — correctly.** The helper normalizes headers into a `Headers` instance. Three credential canaries in the gemini and exa suites read `init.headers` as a plain record and went undefined. Those reads now go through `new Headers(...)`, which accepts either representation. This scopes *how* the assertion reads the init; it does not weaken *what* it asserts, and I re-proved that below rather than asking you to take my word for it.

Deliberately **does not close #2885**: this is retry-path parity, not an explanation of the Bun 1.4.0 versus 1.3.14 difference that issue reports.

## Verification

[...]

PR #2934, opened 2026-08-29
fix(codex): heal a dormant same-grant record and drop stale 401 evidence

lidge-jun:
## Summary

Closes the last two open gaps from #2892. Gaps 1-2 shipped as `8f199fcb6` (#2920), gap 5 as `84049830e` (#2922).

**Gap 3 - a dormant same-grant record is never told the grant rotated.** A refresh rotates the refresh token, but only the flight owner and live joiners learn the new one. A non-deleted record holding the same grant that is not participating in the flight keeps a token upstream has just invalidated. Its next refresh sends a dead grant, `invalid_grant` classifies as `revoked`, and a healthy account is retired - because we rotated its grant and never told it.

Owner and eligible aliases now commit in **one lock acquisition and one `persist`**, so there is no window where some records hold the dead grant. `persist` writes the whole store, so a second pass would have created exactly that window.

**Eligibility is deliberately narrow**, and an adversarial pre-implementation audit is why. My first design also repaired an alias whose access token had moved on, giving it only the rotated refresh token. That is wrong in two ways I could not rebut:

- `plan-from-token` treats a higher generation as proof of a **newer access-token JWT** - that is how JWT plan claims supersede an older WHAM observation. Bumping a generation while keeping the old access token lets a stale JWT overwrite an authoritative plan. `tests/codex-plan.test.ts` already pins that meaning.
- Flights are keyed by grant and do **not** record participants, so a scan cannot tell a dormant record from a live joiner. Rotating a joiner's grant while preserving its 401-rejected bearer makes the provenance CAS inapplicable, the call recurses, and the freshness shortcut hands the rejected token straight back - defeating 401 recovery in precisely the case that branch existed to serve.

So an alias is repaired only when provably an untouched duplicate: same pre-refresh fingerprint, access token, expiry, and a **non-empty** `chatgptAccountId` equal to the owner's. The identity check is not decoration - a fingerprint is `sha256` of the refresh token and carries no identity claim, and no invariant here guarantees one grant cannot span two account ids. Absent or empty identity **fails closed** rather than matching. Access token, refresh token, and expiry move together, so the generation bump keeps meaning what every fence assumes. `replacedAt` and validation metadata survive, because probe-lease settlement accepts only an intact `G -> G+1`.

Propagation installs the rotated JWT, so each alias also needs its **plan** reconciled - at its own committed generation, not the owner's. That settles on the shared refresh flight introduced by #2933 (`propagatedAliases` on `CodexRefreshResult`), rather than inside the commit.

**This is a partial close of gap 3 and the issue comment says so.** Mixed aliases - same grant, diverged access token - need durable grant lineage plus verified identity binding, which the current model cannot express safely.

**Gap 4 - stale 401 evidence outlives the credential it describes.** The reporter's async interleaving is **not reachable**: `recordCodexUpstreamOutcome` is synchronous and no `await` separates the check from the mutations. The cross-process race is real regardless - the check is an unlocked read while writers hold the mutation lock, and OS preemption needs no `await`.

**A post-write re-read cannot fix it, and the design that tried was abandoned.** @Ingwannu reproduced the surviving ordering on my own head: record the 401 at G, return, *then* persist G+1, and the quarantine still applies. A re-read only catches replacements that beat that one read.

So evidence is **tagged with the credential it describes and judged when it is read**:

- `credentialFailureGeneration` is a field **on the health entry** (`src/codex/routing.ts`), not a side map keyed by account id. A side map spends whatever health is current when the credential is found dead, which deletes a later unrelated entry - a G1 401, a G2 save, then a genuine G2 503 would lose the 503.
- `reauthAccounts` became `Map<string, number | undefined>` (`src/codex/account-runtime-state.ts`). `undefined` means no credential was in hand - a login flow - and stays **account-wide**, outranking a generation-scoped mark so it cannot silently expire.
- `preservedCooldownFields` **explicitly drops** the tag: provenance belongs to one credential failure and must not be inherited by a surviving cooldown entry.
- `dropSpentCredentialFailure` runs at `getCodexUpstreamHealth`, at `shouldFailover`, **and at `recordCodexUpstreamOutcome` entry** - the last one is what stops a spent 401 from donating its failure count to the replacement's counters.
- The three sidecar recorders pass `credentialGeneration` for `kind === "pool"` (`src/server/responses/core.ts`, `src/providers/openai-sidecar.ts`). `main-pool` is intentionally left unfenced.

`recordCodexUpstreamOutcome` stays **synchronous** and the credential lock is still never taken on the request path: it runs with `busy_timeout=0`, so locking there would turn ordinary contention into thrown request errors. Affinity sweeping is deliberately **not** rolled back - those entries already carry a generation and self-invalidate.

**The earlier test seam is gone.** `setCodexCredentialRevalidationSeamForTests` existed to stand in for another process's write; once the fence moved to read time, the same interleaving is reproducible without a seam, so it was deleted rather than shipped.

## Verification

`bun x tsc --noEmit` - exit 0. `bun run privacy:scan` - passed. **465 pass / 0 fail / 7,665 expectations** on this exact head across eight focused suites under isolated `OPENCODEX_HOME`/`CODEX_HOME`:

| Suite | Result |
|---|---|
| `tests/codex-routing.test.ts` | 163 pass |
| `tests/codex-auth-api.test.ts` | 194 pass |
| `tests/codex-account-store.test.ts` | 41 pass |
| `tests/responses-pool-401-refresh.test.ts` | 20 pass |
| `tests/gemini-web-search.test.ts` | 13 pass |
| `tests/exa-web-search.test.ts` | 17 pass |
| `tests/codex-plan.test.ts` | 10 pass |
| `tests/credential-redirect-guard.test.ts` | 7 pass |

Every hosted check on this head is green, including macOS and all four Linux shards.

Named mutations, each reproduced against the current implementation:

| Mutation | Result |
|---|---|
| owner-only CAS (pre-fix behaviour) | gap-3 propagation test red |
| drop the `chatgptAccountId` identity guard | gap-3 test red |
| drop the untouched-access-token guard (the audit-rejected branch) | gap-3 test red |
| treat absent/empty identity as a match | fail-closed test red |
| tombstone guard removed | resurrection test red |
| unscope the 401 evidence (drop `credentialFailureGeneration`) | gap-4 stale-evidence test red, live-credential test **green** |
| never spend the tag (make the fence unconditional the other way) | live-credential test red, stale-evidence test **green** |
| remove the `dropSpentCredentialFailure` call at `recordCodexUpstreamOutcome` entry | failure-count inheritance test red |
| owner-only plan reconciliation | alias plan test red |

Those opposite-direction pairs are the point: they prove the fence is conditional, not a blanket "never quarantine".

Two honest notes about the test work. One assertion was **vacuous as first written** - the fixture gave owner and alias the same generation, so the per-alias fence passed even with the owner's generation substituted; the fixture now diverges them and the mutation goes red. And one mutation initially patched the **wrong `dropSpentCredentialFailure` call site** and looked vacuous for that reason. Both are reported rather than quietly fixed.

**Not verified:** how often production OAuth flows actually create two records with byte-identical refresh grants - the tests manufacture that condition, and the code neither enforces grant uniqueness nor documents how aliases arise. I also cannot confirm from repository source whether OpenAI's token service can bind one grant to two account ids, which is why identity is compared rather than assumed. The cross-process interleaving is proven by construction in-process, not captured off a live two-process race.

## Checklist

- [x] Focused regressions beside the existing tests for these subsystems, each mutation-proven
- [x] `bun run privacy:scan` green
- [x] No credentials, tokens, or account identifiers logged (the propagation notice logs a count only)
- [x] Targets `dev`
- [x] Security-sensitive credential path - flagged for non-author security review per `MAINTAINERS.md`
- [x] Not a GUI change, so no screenshot applies

### DIFF
--- a/devlog/_plan/260830_lane_q_2892_gaps_3_4/000_units.md
+++ b/devlog/_plan/260830_lane_q_2892_gaps_3_4/000_units.md
@@ -82,20 +82,46 @@ bare `Set<string>` fenced only by config generation, and affinity clearing remov
 every entry for the account.
 
 Affinity is already self-invalidating on the next generation check. Health and
-reauth are not. Rather than thread a generation through every health consumer, the
-fix keeps the whole sequence synchronous and makes the *observable end state*
-free of stale evidence: snapshot the health entry and the reauth flag, apply the
-mutations, then re-validate the generation and roll back if it stopped being live.
+reauth were not.
 
-A cross-process replacement landing inside the window is therefore caught after
-the fact instead of being prevented, which is the strongest guarantee available
-without taking the config lock on the request path — the lock runs with
-`busy_timeout=0`, so acquiring it per outcome would convert ordinary contention
-into thrown request errors. Affinity clearing is deliberately not rolled back: it
-is self-invalidating, and re-adding swept entries would be a worse bug than the
-sweep. `recordCodexUpstreamOutcome` stays synchronous — many callers consume it as
+My first attempt snapshotted the state, mutated, then re-read the generation and
+rolled back. Two reviewers independently rejected it, correctly: a replacement can
+land at any point *after* `recordCodexUpstreamOutcome` returns, so a post-write
+read narrows the window without closing it. @Ingwannu reproduced the surviving
+ordering on the exact head — record a 401 at G, return, then persist G+1, and the
+quarantine still applied to G+1.
+
+The evidence is now tagged with the credential it came from and checked when it is
+*read*, which is what actually settles it. `credentialFailureGeneration` holds the
+generation a 401/403 was derived from, and the health readers (`shouldFailover`,
+`getCodexUpstreamHealth`) drop a failure whose credential is gone. The reauth set
+became a map from account id to the generation that justified the flag, with
+`undefined` preserved as an account-wide mark so a login flow with no specific
+credential still quarantines unconditionally.
+
+Affinity clearing stays un-reverted: entries already carry a generation and
+self-invalidate, so re-adding swept entries would be the worse bug.
+`recordCodexUpstreamOutcome` stays synchronous — many callers consume it as
 `void` (`src/server/responses/core.ts:391`), so making it async would silently
-leave mutations unawaited.
+leave mutations unawaited. The config lock is still never taken on the request
+path; it runs with `busy_timeout=0`, and per-outcome acquisition would turn
+contention into request errors.
+
+## The alias plan note
+
+Review also caught that propagation installs the rotated JWT on an alias but left
+its configured plan alone: a `plus → pro` rotation gave the alias a Pro credential
+while its plan stayed `plus`, and the cached-token fast path never repairs that, so
+quota scoring and the 30-day projection stayed wrong until a restart or a WHAM
+refresh. Each propagated alias is now reconciled at its **own** committed
+generation, which is why the commit returns `{ id, generation }` rather than ids —
+aliases need not share a generation, and the plan note is generation-fenced.
+
+That last point produced the one genuinely vacuous assertion of this unit: with a
+single `saveCodexAccountCredential` per record, owner and alias generations
+coincided, so an assertion about the per-alias fence passed even when the code used
+the owner's generation. The fixture now advances the alias twice so the generations
+diverge, and the mutation turns red.
 
 ## Constraints the audit flagged
 
--- a/devlog/_plan/260830_lane_q_2892_gaps_3_4/000_units.md
+++ b/devlog/_plan/260830_lane_q_2892_gaps_3_4/000_units.md
@@ -1,0 +1,113 @@
+# Lane Q — issue #2892 gaps 3 and 4
+
+The last two of the five gaps #2892 raised against the merged stored-Pool 401
+recovery path. Gaps 1–2 shipped as `8f199fcb6` (#2920), gap 5 as `84049830e`
+(#2922). An independent recon audit re-derived both remaining gaps from current
+`dev` and confirmed each is still reachable — and corrected the reporter on one
+point, recorded below.
+
+## Gap 3 — a rotated grant never reaches an inactive same-grant alias
+
+A successful refresh persists the rotated credential to the flight owner only, via
+the generation CAS at `src/codex/account-store.ts:748`. A live joiner can CAS the
+result onto its own record. Nothing writes to a third category: a non-deleted
+record carrying the same `refreshGrantFingerprint` that is not participating in
+the flight.
+
+`findFreshCredentialForGrant` (`src/codex/account-store.ts:393`) is a pre-fetch
+lookup and propagates nothing. So the alias keeps a refresh token that upstream
+has just rotated away. The next refresh on that alias sends a dead grant, and
+`invalid_grant` is classified `revoked` — which retires a healthy account. That
+classification is correct behavior for a genuinely dead grant; the defect is that
+the grant died because we rotated it and never told the alias.
+
+### The design an adversarial audit rejected
+
+My first plan had two branches: an untouched alias adopts the rotated credential
+whole, and an alias whose access token had changed concurrently keeps its own
+access token but takes **only** the rotated refresh token. An independent audit
+refused that second branch, with two findings I could not rebut:
+
+- The generation fence in `src/codex/plan-from-token.ts:32` treats a higher
+  generation as proof of a **newer access-token JWT**, which is what lets JWT plan
+  claims supersede an older WHAM observation. Bumping a generation while
+  deliberately keeping the old access token lets a stale JWT overwrite an
+  authoritative plan. `tests/codex-plan.test.ts:129` already pins that meaning.
+- A flight is keyed by grant and does not record participant account ids
+  (`src/codex/account-store.ts:308`), so a scan cannot distinguish a dormant alias
+  from a live joiner. Rotating a joiner's grant while preserving its 401-rejected
+  access token makes the provenance CAS inapplicable, and the recursion's
+  freshness shortcut then returns the rejected bearer — defeating 401 recovery in
+  exactly the case the branch existed to serve.
+
+### What ships instead
+
+One batch compare-and-swap, one `persist`, and a deliberately narrow eligibility
+test. An alias is repaired only when it is provably an untouched duplicate of the
+pre-refresh credential: same old grant fingerprint, same access token, same
+expiry, and the same `chatgptAccountId` as the owner. Such an alias receives the
+rotated access token, refresh token, and expiry **together**, so the generation
+bump keeps meaning what every fence already assumes. `replacedAt` and the
+validation metadata are preserved, because the probe-lease lineage check accepts
+only an intact `G → G+1`.
+
+Anything else is left alone: a differing access token, a differing account id, or
+a tombstone. The `chatgptAccountId` equality requirement is not decoration — a
+fingerprint is `sha256` of the refresh token and carries no identity claim
+(`src/codex/account-store.ts:62`), and no repository invariant guarantees one
+grant cannot span two account ids.
+
+This is a **partial** close of gap 3, and the issue comment says so. Dormant
+duplicates stop being retired for a grant we rotated away; a mixed alias still is.
+Healing that case needs durable grant lineage and verified identity binding, which
+the current fingerprint-and-generation model cannot express safely.
+
+The flight's returned `resolvedGrantFingerprint` stays the **old** fingerprint:
+joiners wait on that key, and retagging it would make every legitimate joiner look
+foreign.
+
+## Gap 4 — stale credential evidence writes unscoped state
+
+The reporter described an async interleaving between validation and mutation. That
+part is wrong and worth stating: `recordCodexUpstreamOutcome` is synchronous
+(`src/codex/routing.ts:2095`) and there is **no `await`** between the generation
+check at `src/codex/routing.ts:2210` and the mutations at 2216–2223. The
+same-process race the issue describes is not reachable.
+
+The cross-process race is real regardless. The check is an unlocked synchronous
+store read (`src/codex/account-store.ts:186`) while writers coordinate under the
+mutation lock, and OS preemption needs no `await`. The side effects then carry no
+credential identity: health entries have no generation field, reauth state is a
+bare `Set<string>` fenced only by config generation, and affinity clearing removes
+every entry for the account.
+
+Affinity is already self-invalidating on the next generation check. Health and
+reauth are not. Rather than thread a generation through every health consumer, the
+fix keeps the whole sequence synchronous and makes the *observable end state*
+free of stale evidence: snapshot the health entry and the reauth flag, apply the
+mutations, then re-validate the generation and roll back if it stopped being live.
+
+A cross-process replacement landing inside the window is therefore caught after
+the fact instead of being prevented, which is the strongest guarantee available
+without taking the config lock on the request path — the lock runs with
+`busy_timeout=0`, so acquiring it per outcome would convert ordinary contention
+into thrown request errors. Affinity clearing is deliberately not rolled back: it
+is self-invalidating, and re-adding swept entries would be a worse bug than the
+sweep. `recordCodexUpstreamOutcome` stays synchronous — many callers consume it as
+`void` (`src/server/responses/core.ts:391`), so making it async would silently
+leave mutations unawaited.
+
+## Constraints the audit flagged
+
+The refresh-flight map is keyed by the old grant (`src/codex/account-store.ts:315`)
+and joiner provenance deliberately carries that old fingerprint, so alias
+propagation must not disturb that ordering. The config lock runs with
+`busy_timeout=0` and must stay synchronous, so the routing path must not acquire
+it per outcome. Affinity requires exact credential-generation equality, so any
+alias generation bump has to be reasoned about rather than assumed harmless.
+
+## Evidence standard
+
+Each regression is driven red by a named mutation, using the existing blocked-fetch
+seam rather than a timing sleep. Any assertion that survives its mutation is
+deleted rather than kept.
--- a/src/codex/account-runtime-state.ts
+++ b/src/codex/account-runtime-state.ts
@@ -1,18 +1,43 @@
 import { captureConfigGeneration, type GenerationContext } from "../lib/state-store-sweeper";
+import { isCodexAccountGenerationLive } from "./account-store";
 
-const reauthAccounts = new Set<string>();
+/**
+ * Accounts quarantined for reauthentication, each remembering WHICH credential produced the
+ * evidence (#2892 gap 4).
+ *
+ * A 401 describes one credential, not an account. Recording only the id let a 401 raced by a
+ * cross-process credential replacement quarantine the replacement: the flag outlived the credential
+ * it was evidence about, and routing then refused a perfectly good credential until a restart. A
+ * post-write re-read cannot fix that — the replacement may land at any point after the write — so
+ * the generation travels WITH the flag and is checked when the flag is read.
+ *
+ * `undefined` means "no credential generation was supplied", which stays account-wide: callers such
+ * as a login flow have no specific credential in hand, and their quarantine must not silently expire.
+ */
+const reauthAccounts = new Map<string, number | undefined>();
 let lastReconciledGeneration = 0;
 let liveAccountIds = new Set<string>();
 
-export function markAccountNeedsReauth(id: string, writerGeneration = captureConfigGeneration()): void {
+export function markAccountNeedsReauth(
+  id: string,
+  writerGeneration = captureConfigGeneration(),
+  credentialGeneration?: number,
+): void {
   if (writerGeneration < lastReconciledGeneration && !liveAccountIds.has(id)) return;
-  reauthAccounts.add(id);
+  // An account-wide mark supersedes a generation-scoped one: it is the stronger claim.
+  if (credentialGeneration === undefined || !reauthAccounts.has(id)) {
+    reauthAccounts.set(id, credentialGeneration);
+    return;
+  }
+  const existing = reauthAccounts.get(id);
+  if (existing === undefined) return;
+  reauthAccounts.set(id, Math.max(existing, credentialGeneration));
 }
 
 export function reconcileCodexReauthState(context: GenerationContext): number {
   if (context.generation <= lastReconciledGeneration) return 0;
   let removed = 0;
-  for (const id of reauthAccounts) {
+  for (const id of [...reauthAccounts.keys()]) {
     if (context.codexAccountIds.has(id)) continue;
     reauthAccounts.delete(id);
     removed += 1;
@@ -23,7 +48,16 @@ export function reconcileCodexReauthState(context: GenerationContext): number {
 }
 
 export function isAccountNeedsReauth(id: string): boolean {
-  return reauthAccounts.has(id);
+  if (!reauthAccounts.has(id)) return false;
+  const credentialGeneration = reauthAccounts.get(id);
+  if (credentialGeneration === undefined) return true;
+  // The credential this evidence describes is gone, so the evidence is spent. Drop it rather than
+  // re-deriving the same answer on every read.
+  if (!isCodexAccountGenerationLive(id, credentialGeneration)) {
+    reauthAccounts.delete(id);
+    return false;
+  }
+  return true;
 }
 
 export function clearAccountNeedsReauth(id: string): void {
--- a/src/codex/account-store.ts
+++ b/src/codex/account-store.ts
@@ -253,12 +253,12 @@ export function commitRefreshedCodexCredentialWithAliases(
   id: string,
   generation: number,
   cred: CodexAccountCredentials,
-): { committed: boolean; propagatedAliasIds: string[] } {
+): { committed: boolean; propagatedAliases: { id: string; generation: number }[] } {
   return withCredentialMutationLockSync(() => {
     const store = loadCodexAccountRecordStore();
     const current = store[id];
     if (!current || current.generation !== generation || current.deletedAt != null || !current.credential) {
-      return { committed: false, propagatedAliasIds: [] };
+      return { committed: false, propagatedAliases: [] };
     }
     const priorCredential = current.credential;
     const priorFingerprint = recordGrantFingerprint(current);
@@ -273,7 +273,9 @@ export function commitRefreshedCodexCredentialWithAliases(
       ...preservedValidationMetadata(current),
     };
 
-    const propagatedAliasIds: string[] = [];
+    // Each alias carries its OWN committed generation: aliases need not share one, and the plan
+    // reconciliation below is generation-fenced, so an id alone would be reconciled at the wrong fence.
+    const propagatedAliases: { id: string; generation: number }[] = [];
     // Nothing to propagate when the grant did not actually rotate: the aliases already hold it.
     if (priorFingerprint !== undefined && priorCredential.refreshToken !== cred.refreshToken) {
       for (const [aliasId, alias] of Object.entries(store)) {
@@ -282,19 +284,20 @@ export function commitRefreshedCodexCredentialWithAliases(
         if (alias.credential.accessToken !== priorCredential.accessToken) continue;
         if (alias.credential.expiresAt !== priorCredential.expiresAt) continue;
         if (alias.credential.chatgptAccountId !== priorCredential.chatgptAccountId) continue;
+        const aliasGeneration = alias.generation + 1;
         store[aliasId] = {
           // The alias keeps its OWN chatgptAccountId value, which the guard above proved equal.
           credential: { ...cred, chatgptAccountId: alias.credential.chatgptAccountId },
-          generation: alias.generation + 1,
+          generation: aliasGeneration,
           refreshGrantFingerprint,
           replacedAt: alias.replacedAt,
           ...preservedValidationMetadata(alias),
         };
-        propagatedAliasIds.push(aliasId);
+        propagatedAliases.push({ id: aliasId, generation: aliasGeneration });
       }
     }
     persist(store);
-    return { committed: true, propagatedAliasIds };
+    return { committed: true, propagatedAliases };
   });
 }
 
@@ -833,8 +836,16 @@ async function resolveCodexToken(
     if (!commit.committed) {
       throw new CodexCredentialGenerationConflictError();
     }
-    if (commit.propagatedAliasIds.length > 0) {
-      console.warn(`[codex-auth] rotated refresh grant propagated to ${commit.propagatedAliasIds.length} dormant same-grant account record(s)`);
+    if (commit.propagatedAliases.length > 0) {
+      console.warn(`[codex-auth] rotated refresh grant propagated to ${commit.propagatedAliases.length} dormant same-grant account record(s)`);
+      // The rotated JWT can carry a different `chatgpt_plan_type`, and each alias now holds that
+      // JWT. Without this the alias keeps its old plan — its cached-token fast path never
+      // reconciles, and repository-wide reconciliation only runs at startup — so quota scoring and
+      // the 30-day projection stay wrong until a restart or a WHAM refresh. Reconcile at each
+      // alias's OWN committed generation, since the plan note is generation-fenced.
+      for (const alias of commit.propagatedAliases) {
+        await notePlanFromRefreshedAccessToken(alias.id, updated.accessToken, alias.generation);
+      }
     }
     return {
       accessToken: updated.accessToken,
--- a/src/codex/account-store.ts
+++ b/src/codex/account-store.ts
@@ -372,6 +372,12 @@ type CodexTokenResult = { accessToken: string; chatgptAccountId: string; generat
 type CodexRefreshResult = CodexTokenResult & {
   credential?: CodexAccountCredentials;
   /**
+   * Records that adopted this refresh's rotated credential through same-grant propagation, each
+   * with its own committed generation (#2892 gap 3). Carried on the result so the flight settles
+   * every plan in one place rather than the commit doing its own (#2933).
+   */
+  propagatedAliases?: { id: string; generation: number }[];
+  /**
    * Grant the returned credential actually belongs to.
    *
    * Flights are keyed by refresh grant and shared across every account holding that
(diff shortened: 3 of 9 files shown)


lidge-jun:
## 리뷰 · 우선순위 72 / 80

이 PR은 지금 `dev`에 남아 있던 이슈 #2892의 마지막 두 구멍(갭 3·4)을 막습니다. 갭 1–2는 이미 #2920(`8f199fcb6`)으로, 갭 5는 #2922(`84049830e`)로 들어간 상태고, 헤드 `6a907d2a3`(#2932 upstream recovery 패리티)와도 겹치지 않는 독립 버그 수정입니다.

갭 3은 이렇게 생깁니다. 리프레시가 성공하면 리프레시 토큰이 돌아가는데, 지금까지는 플라이트 주인과 살아 있는 조이너만 새 그랜트를 받았습니다. 같은 `refreshGrantFingerprint`를 들고 있으면서 플라이트에 안 들어온 잠든(dormant) 레코드는 예전 그랜트를 그대로 둡니다. 그다음 그 레코드가 리프레시를 치면 이미 무효가 된 그랜트를 보내고, `invalid_grant`가 `revoked`로 분류되어 멀쩡한 계정이 퇴역합니다. 원인은 “그랜트가 진짜 죽었다”가 아니라 “우리가 돌려 놓고 알려 주지 않았다”입니다.

고치는 곳은 `src/codex/account-store.ts`의 새 함수 `commitRefreshedCodexCredentialWithAliases` 입니다. 주인 CAS만 하던 `saveCodexAccountCredentialIfGeneration` 자리를 `resolveCodexToken` 성공 커밋 한 곳에서 이 함수로 바꿉니다. 한 번의 `withCredentialMutationLockSync`와 한 번의 `persist`로 주인과 자격 있는 별칭을 같이 씁니다. `persist`가 스토어 전체를 쓰므로 두 번 나누면 일부만 죽은 그랜트를 가진 창이 생깁니다.

자격 조건은 일부러 좁습니다. 예전 그랜트 지문·액세스 토큰·만료·`chatgptAccountId`가 주인과 같은 손대지 않은 복제본만 고칩니다. 액세스만 이미 바뀐 별칭에 그랜트만 밀어 넣는 넓은 설계는 감사에서 버렸습니다. `plan-from-token`은 세대가 크면 더 새 JWT라고 믿기 때문에, 세대만 올리고 옛 액세스 토큰을 남기면 낡은 JWT가 권위 있는 플랜을 덮을 수 있습니다. 또 플라이트는 그랜트 키만 있고 참가자 id를 안 적어서, 잠든 별칭과 살아 있는 조이너를 구분할 수 없고, 조이너 그랜트만 돌리면 401 복구 재귀가 거절된 베어러를 다시 줄 수 있습니다. 그래서 이 PR은 갭 3의 **부분** 종료입니다. 섞인(mixed) 별칭은 그랜트 계보와 신원 바인딩이 있어야 안전하게 고칠 수 있고, 지금 모델로는 못 합니다.

갭 4는 `src/codex/routing.ts`의 `recordCodexUpstreamOutcome` 자격 증명(credential) 분기입니다. 이슈가 말한 같은 프로세스 `await` 끼어들기는 실제로 안 됩니다(함수가 동기이고 세대 검사와 부작용 사이에 await가 없음). 그래도 다른 프로세스가 뮤테이션 락을 잡고 자격 증명을 갈아끼우는 동안, 검사 쪽은 잠금 없는 읽기라서 창은 있습니다. 헬스·reauth에는 세대 필드가 없어서, 그 창에 낡은 401이 떨어지면 **교체된** 자격 증명을 격리해 버립니다. 요청 경로에서 자격 증명 락을 매번 잡을 수는 없습니다(`busy_timeout=0`이면 경합이 요청 에러가 됨). 그래서 이전 헬스·reauth를 스냅샷하고 부작용을 적용한 뒤 세대를 다시 검사해, 더 이상 live가 아니면 되돌립니다. 어피니티 쓸기는 되돌리지 않습니다(이미 세대로 스스로 무효화되고, 쓸린 항목을 다시 넣으면 더 나쁜 버그).

테스트 심 `setCodexCredentialRevalidationSeamForTests` 는 한 프로세스가 다른 프로세스의 쓰기를 스케줄할 수 없어서 만든 정직한 심입니다. 프로덕션에서는 undefined이고 null 체크 한 번 비용입니다. 갭 3은 전파·툼스톤 부활 금지, 갭 4는 롤백 vs 과잉 롤백 금지가 서로 반대 방향으로 빨개지는 뮤테이션으로 잠겨 있습니다. 변경 파일은 `account-store.ts` / `routing.ts` / 두 테스트 / `devlog/_plan/260830_lane_q_2892_gaps_3_4/000_units.md` 뿐이고 types·config 스플릿과 무관합니다.

라인 825 근처 (account-store resolveCodexToken 커밋) - 주인 전용 CAS를 별칭 전파 CAS로 바꾼 자리. 전파 시 `console.warn`은 개수만 찍고 토큰·계정 id는 안 남김. 로그 노이즈가 걱정되면 디버그 레벨로 내릴지는 취향.
라인 자격 가드 (accessToken/expiresAt/chatgptAccountId) - 하나라도 빠지면 감사에서 거절된 넓은 설계로 되돌아감. 테스트가 가드 제거 뮤테이션으로 빨개짐을 요구함.
경로 setCodexCredentialRevalidationSeamForTests - 테스트 전용 심. export 표면이 늘어남. 프로덕션 호출 경로에 심이 남는지 정적 검색으로 한 번 더 확인하면 좋음.
경로 갭4 롤백 (quotaScopedHealth.delete 후 priorScopedHealth만 복구) - prior가 없으면 삭제 상태가 유지됨. 의도된 동작으로 보이지만, 롤백 범위에 어피니티를 넣지 않은 이유를 리뷰어가 한 번 더 따라가 봐야 함.
이슈 클로즈 문구 - 본문은 “#2892의 마지막 두 갭을 닫는다”고 하지만 GitHub `Closes #2892` 키워드인지는 본문만으로는 애매. 부분 갭3이 남으면 이슈를 완전 종료하면 안 됨.

메인테이너의 판단이 필요한 지점
- 작성자가 본인(lidge-jun)이고 본문이 MAINTAINERS.md 비작성자 보안 리뷰를 스스로 표시함. 머지 전에 다른 메인테이너(또는 명시적 self-ack 정책) 한 명이 자격 증명 경로를 훑을지.
- #2892를 이 PR로 **완전 종료**할지, 아니면 “손대지 않은 복제본만 치료 / mixed alias는 후속”으로 이슈를 열어 두거나 후속 이슈를 팔지.
- 프로덕션에서 바이트 동일 리프레시 그랜트를 가진 두 레코드가 실제로 얼마나 생기는지(본문도 “검증 못 함”이라고 적음). 희귀하면 갭3 이득이 작고, 흔하면 머지 우선순위가 더 올라감.
- `busy_timeout=0` 때문에 락을 못 잡는 대안(헬스/reauth에 세대 필드 추가)을 후속으로 갈지, 지금 롤백으로 충분한지.

너의 추천
비작성자 보안 훑기 한 번 받은 뒤 `dev`에 머지하는 쪽을 추천합니다. 머지 시 #2892는 “갭 3 부분 남음(mixed alias)”을 이슈 코멘트에 남기고 완전 종료할지 후속 이슈로  manifest 할지 결정하세요. types/config 스플릿과 무관하니 리베이스·클로즈 대상이 아닙니다. 랜딩 후 원본 이슈/관련 PR 정리 규칙(`Landed via #…`, `landed-via-maintainer`)을 적용하면 됩니다.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
@Ingwannu — requesting security review per `MAINTAINERS.md` (authentication and credential handling). This writes to the Codex credential store and changes 401 quarantine behavior, so it should not merge on author authority alone.

The two decisions worth your attention are both deliberate narrowings, not conveniences:

1. **What is NOT repaired.** An alias whose access token moved on is left holding the dead grant. Repairing it would bump a generation while keeping an old access-token JWT, and `plan-from-token` reads a higher generation as proof of a newer JWT. It would also let a live forced-refresh joiner be handed back its own 401-rejected bearer, since flights are keyed by grant and do not record participants. So gap 3 closes for untouched duplicates only.

2. **`chatgptAccountId` is compared, not copied.** A grant fingerprint is `sha256` of the refresh token and carries no identity claim. I could not establish from repository source that one grant can never span two account ids, so identity equality is required before any propagation.

The gap-4 rollback is conditional and I proved that in both directions: removing it fails the racing test while the live-credential test stays green, and making it unconditional fails the live-credential test while the racing test stays green.

Two things I could not verify, stated plainly: how often production OAuth flows create two records with byte-identical refresh grants (the tests manufacture it), and whether OpenAI's token service can bind one grant to two account ids.

Ingwannu (changes requested):
Reviewed exact head `a798e2dbf88ce8bf5f84ba47034bcc96a563331c` against `dev@6a907d2a3c6496935ec87d86240a6a12b0ffa00b`.

The narrow alias eligibility and one-write credential commit are directionally sound, and the focused account-store/routing/401/plan/auth suites pass (`420/420`) under isolated `HOME`, `OPENCODEX_HOME`, and `CODEX_HOME`; `privacy:scan` also passes. I am requesting changes because two generation-boundary defects remain on this head.

1. `src/codex/routing.ts:2262-2269` still does not scope health or reauth evidence to a credential generation. The post-write re-read only catches a replacement that lands before that particular read. A replacement can land immediately after it returns true, leaving the account-id-only `upstreamHealth` and `needsReauth` state attached to G+1. I reproduced the simple valid ordering on this exact head: record a 401 for G=1, let `recordCodexUpstreamOutcome` return, then persist G+1; `needsReauth` remained `true` and health remained `{ lastFailureStatus: 401 }` for the fresh credential. Store and check the credential generation with every health/reauth entry (or clear/fence the state from the credential replacement commit under the same coordination boundary). Repeating an unlocked read cannot close this race. Add a regression that replaces the credential after the outcome call and proves G+1 is selectable without stale health/reauth state.

2. `src/codex/account-store.ts:285-293` installs the rotated access-token JWT and advances every propagated alias generation, but refresh completion reconciles plan metadata only for the owner id. I reproduced a `plus -> pro` rotated JWT with two untouched same-grant records: the owner persisted `{ plan: "pro", planSource: "jwt", planCredentialGeneration: 2 }`, while the alias held the new Pro credential at generation 2 but remained `plan: "plus"`. Its cached-token fast path does not repair that during normal use, so quota scoring can stay wrong until startup reconciliation or WHAM refresh. Reconcile each propagated alias using its exact committed generation; returning only alias ids is insufficient if their generations differ. Coordinate this with #2933's shared-flight reconciliation so the final design has one deterministic owner/alias plan settlement path, and add a plan-changing alias regression.

The current TypeScript errors are the unchanged `fetch(..., { timeout })` typing failures also reproducible on the current integration tree, so they are not attributed to this PR. No repository-wide security scan was run.

Ingwannu:
Follow-up from the overlapping #2935 review: one part of that PR is independently correct and should be carried into this patch before re-review.

`findFreshCredentialForGrant` still matches a sibling by refresh-grant fingerprint and freshness alone. The fingerprint proves grant equality, not upstream account identity; this repository has no invariant that the same stored grant cannot appear under two `chatgptAccountId` values. A fresh foreign-identity record can therefore satisfy the pre-fetch adoption path and hand its access token/account id to the caller.

Require a non-empty expected `chatgptAccountId`, require the candidate's exact identity to match, and fail closed when either identity is empty. Add a regression with two records sharing the grant fingerprint but using different upstream identities, proving the caller refreshes for itself rather than adopting the foreign bearer.

This is in addition to the two generation-state blockers in my formal review; it does not make #2935's `grant-only` fan-out safe.

lidge-jun:
Both findings were correct and both are fixed in `f8e7bbdc4`. Thank you for reproducing them on the exact head — the first one in particular I had argued for in the PR body, and the argument was wrong.

**1. Stale-401 state is now scoped by credential generation, not settled by a re-read.** You are right that repeating an unlocked read cannot close the race: the replacement can land at any point after `recordCodexUpstreamOutcome` returns, and my post-write check only saw replacements that beat that one read. I removed the rollback entirely.

Evidence is now tagged with the credential it describes and checked at consumption:

- `credentialFailureGeneration` records the generation each 401/403 was derived from, and the health readers (`shouldFailover`, `getCodexUpstreamHealth`) drop a failure whose credential is no longer live.
- `reauthAccounts` became `Map<string, number | undefined>`. `undefined` is deliberately preserved as an account-wide mark, so a caller with no specific credential in hand — a login flow — still quarantines unconditionally, and an account-wide mark outranks a generation-scoped one rather than being downgraded by it.
- A healthy terminal retires the tag.

The regression now reproduces your exact ordering with **no seam**: record the 401 while G is live, let the call return, persist G+1, then assert `isAccountNeedsReauth` is false, health is null, and the account is selectable again. The seam I shipped before is deleted, which is a better outcome than the seam.

**2. Alias plan metadata is reconciled per alias, at its own committed generation.** Confirmed the `plus → pro` case you described. The commit now returns `{ id, generation }` pairs rather than ids, precisely because aliases need not share a generation and the plan note is generation-fenced — returning ids alone would have reconciled at the wrong fence, as you noted.

Worth flagging honestly: my first version of that assertion was **vacuous**. With one save per record the owner and alias generations coincided, so asserting the per-alias fence passed even when I substituted the owner's generation. The fixture now advances the alias so the generations diverge, and that mutation turns red. I would rather report that than leave a test that passes for no reason.

On coordinating with #2933: agreed that owner/alias plan settlement should end up with one deterministic path. #2933 is @luvs01's and touches the same file, so I have not pre-empted it — whichever lands second should unify the two call sites rather than both keeping their own loop, and I am happy for that to be mine.

**421 pass / 0 fail** across `codex-routing`, `codex-account-store`, `codex-plan`, `responses-pool-401-refresh`, and `codex-auth-api`. Typecheck and `privacy:scan` clean. Re-requesting review.

Ingwannu (changes requested):
Reviewed exact head `f8e7bbdc49e91936adf847e00aaaf11ba0e66a3e` against `dev@6a907d2a3c6496935ec87d86240a6a12b0ffa00b`.

The two failures from my formal review are materially improved: a post-return credential replacement now spends the generation-scoped 401/reauth evidence, and propagated aliases reconcile the rotated JWT plan at each alias's own committed generation. The focused account-store/routing/plan/401/auth suites pass on this head (`421/421`, `7547` expectations), as does `bun x tsc --noEmit`, under isolated `HOME`, `OPENCODEX_HOME`, and `CODEX_HOME`.

I am still requesting changes because three correctness boundaries remain.

1. `credentialFailureGeneration` is side metadata for an account id, not provenance attached to one exact `upstreamHealth` value. `dropSpentCredentialFailure` therefore deletes whatever health value happens to be current when the old credential is found dead. I reproduced this on the exact head: record a generation-1 401, save generation 2, record a genuine generation-2 503 before any health read, then call `getCodexUpstreamHealth`. The expected current health is status 503; the actual result is `null` because the stale generation-1 tag deletes the later transient entry. Workspace and account-wide quota replacements have the same ownership problem. Put credential provenance on the exact health entry (and only spend that matching entry), or explicitly retire the tag whenever any non-credential outcome replaces/updates health. Add the `G1 401 -> G2 save -> G2 503 -> health remains 503` regression and cover at least one overwrite branch.

2. Not every status path supplies the generation that the new boundary requires. `sidecarOutcomeRecorder` in `src/server/responses/core.ts` and both `recordOutcome` closures in `src/providers/openai-sidecar.ts` can receive HTTP 401/403 from vision or web-search, but omit `credentialGeneration` even for `authContext.kind === "pool"`. That intentionally falls into the account-wide `undefined` reauth mark. Exact-head reproduction through `sidecarOutcomeRecorder`: record a sidecar 401 at G1, save G2, then read reauth; expected `false`, actual `true`. Pass `authCtx.generation` on every stored-pool outcome surface and add a sidecar regression proving the replacement is not quarantined. Main-pool can retain its separate unfenced semantics because it has no stored-record generation.

3. The independently requested same-grant identity guard is still absent. `findFreshCredentialForGrant` matches fingerprint, freshness, and rejected bearer only; it still neither accepts nor checks the caller's expected `chatgptAccountId`. A fresh record with the same stored grant but a different upstream account identity can therefore satisfy the adoption path and be copied onto the caller. Require both identities to be non-empty and exactly equal before adoption, and retain the foreign-identity regression requested in my prior follow-up.

The alias-plan fix should remain. After these boundaries are addressed, rebase on #2933 if it lands first and unify the shared-flight plan settlement rather than keeping two competing paths. No repository-wide security scan was run.

lidge-jun:
All three boundaries fixed in `9acb7b593`. Every one reproduced as you described, and the first was a defect my previous commit introduced — thank you for catching it.

**1. Provenance now belongs to the exact health entry.** You were right that `credentialFailureGeneration` as side metadata meant "spend whatever health value happens to be current". `credentialFailureGeneration` is now a field on `CodexUpstreamHealth`, so only an entry that actually carries it can be spent and any later write replaces the entry together with its tag.

Your ordering is now a regression: `G1 401 → G2 save → G2 503 → getCodexUpstreamHealth` keeps status 503. The workspace-overwrite branch is covered too.

One thing your review implied that I want to make explicit, because I hit it while fixing this: `preservedCooldownFields` spreads surviving fields into cooldown writes, so the tag leaked into a 429 entry and reintroduced the same ownership bug one level down. That helper now drops it explicitly. My first attempt at the fix failed the new test for precisely that reason.

**2. Stored-pool sidecar outcomes now carry the generation.** Confirmed on all three surfaces — `sidecarOutcomeRecorder` plus both `recordOutcome` closures in `openai-sidecar.ts`. They pass `credentialGeneration` for `kind === "pool"`, and `main-pool` deliberately keeps the unfenced account-wide semantics since it has no stored-record generation, per your note. The sidecar regression drives a 401 through `sidecarOutcomeRecorder`, replaces the credential, and asserts the replacement is neither quarantined nor holding stale health.

**3. `findFreshCredentialForGrant` requires identity equality.** It now takes the caller's expected `chatgptAccountId`, requires both values non-empty and exactly equal, and adopts nothing when no expected identity is supplied. The foreign-identity regression asserts a real refresh runs instead of adoption, and that the sibling record is left untouched.

Mutations, each turning its own test red: removing the per-entry provenance, un-fencing the sidecar recorder, and removing the identity guard.

**494 pass / 0 fail** across `codex-routing`, `codex-account-store`, `codex-plan`, `responses-pool-401-refresh`, `codex-auth-api`, `vision-sidecar-e2e`, and `web-search`. Typecheck and `privacy:scan` clean.

On the earlier macOS failure: that was `shellStreamExec completion acknowledgement` timing out at 60s in `cursor-native-exec-shell.test.ts`. It spawns a real `echo` subprocess, my diff touches only `src/codex/*` and the two sidecar recorders, and the job passed on rerun with no code change (16156 pass / 1 fail → green). I am calling it a flake rather than attributing it to this branch, but say so if you read it differently.

Agreed on unifying with #2933's shared-flight plan settlement once one of us lands. Re-requesting review.

Ingwannu (changes requested):
Reviewed exact head 7d95449244f82396557ffb5ff6a125fff4b24940 against dev@dd159db9088033bf58c91add8968b0b83bbf263c.

The three previously requested production boundaries are now directionally correct: credential provenance is owned by the exact health entry, stored-pool sidecars pass the stored generation, and same-grant adoption requires a non-empty exact ChatGPT account identity. Folding #2933's plan settlement onto the shared flight also removes the competing owner/alias settlement paths.

I am still requesting changes for two exact-head blockers:

1. The claimed focused set is not green as a set under the repository's Bun 1.3.14 runtime. Running codex-routing, codex-account-store, codex-plan, responses-pool-401-refresh, codex-auth-api, vision-sidecar-e2e, and web-search under isolated HOME/OPENCODEX_HOME/CODEX_HOME produced 494 pass / 1 fail. The failure is tests/codex-routing.test.ts:583: immediately after sidecarOutcomeRecorder(... writerGeneration: 0)(401), isAccountNeedsReauth("a") is false. The same test passes when filtered alone, so this is order-dependent state, not proof of the intended boundary. Earlier tests advance reconciliation state; the hard-coded stale writer generation can then cause recordCodexUpstreamOutcome to reject the outcome before the new credential-generation logic is exercised. Use the current captured/config writer generation that a production pool auth context carries, or explicitly reset the relevant reconciliation state in the fixture. Prove the full exact-head set green, not only the filtered row.

2. src/codex/routing.ts:2249-2265 still documents the superseded rollback design. It says health has no generation, reauth is a bare id set, and mutations are re-read/restored afterwards. The current code does the opposite: provenance is attached to CodexUpstreamHealth and the reauth map, consumption spends only matching stale entries, and there is no post-write rollback. This is the core security rationale future maintainers will rely on, so update it to describe the implementation that actually ships.

The previous G1 401 -> G2 save -> G2 503 and workspace-overwrite regressions, the foreign-identity adoption regression, and the shared-flight plan settlement remain valuable. Exact-head hosted CI is still running, so no approval or merge should occur until this deterministic failure and the misleading boundary comment are corrected and the new head is fully green.

No repository-wide security scan was run.

Ingwannu (changes requested):
Follow-up on exact head 7d95449244f82396557ffb5ff6a125fff4b24940 after validating the new account-store finding: it is correct and adds a third blocker to my existing CHANGES_REQUESTED review.

In commitRefreshedCodexCredentialWithAliases, src/codex/account-store.ts:286 accepts alias.credential.chatgptAccountId === priorCredential.chatgptAccountId when both values are the empty string. That is not proof of upstream identity. A ChatGPT OAuth credential can select different account or workspace identities through the account id, and the function documentation already states that grant equality alone carries no identity claim. Matching access token and expiry only proves that the records copied the same bearer snapshot; it does not establish which account identity an empty-id record was intended to use.

Require priorCredential.chatgptAccountId to be non-empty and require each alias id to be non-empty before exact equality. If identity is absent, fail closed and leave that dormant record untouched. Add a regression with owner and alias records that share grant fingerprint, access token, expiry, and empty account ids, then prove the alias is not overwritten when the owner refresh rotates.

The existing order-dependent focused-suite failure and stale routing comment remain unresolved as already documented. Exact-head macOS CI is also still pending. No repository-wide security scan was run.

lidge-jun:
All three fixed in `5a3795769`. Every one was correct, including the one that made my own test worthless.

**Empty ids fail closed.** Propagation now requires the owner's `chatgptAccountId` to be non-empty and each alias's to be non-empty before comparing them, and skips the alias loop entirely when the owner identity is absent. The regression uses owner and alias sharing fingerprint, access token, expiry, and `""` ids, and asserts the alias keeps its old credential *and* its generation while the owner rotates normally. Accepting empty ids turns it red.

**The sidecar test was passing for the wrong reason,** which is worse than failing. `writerGeneration: 0` sits below whatever reconciliation state earlier tests advanced to, so `recordCodexUpstreamOutcome` could return at its writer-generation guard before reaching the credential-generation logic the test claims to cover. It now captures the current generation as a production pool auth context does, and asserts the quarantine actually applied before asserting it is released — so the test cannot silently stop exercising its own subject again.

**The gap-4 comment now describes what ships.** You are right that this is the rationale future maintainers will lean on, and it still described the rollback design I deleted. It now explains tagging and read-time judgement, why no re-read can close the race, and why `preservedCooldownFields` has to drop the tag.

**496 pass / 0 fail** across the seven-file set run as a set rather than filtered. Typecheck and `privacy:scan` clean.

One difference in evidence worth stating rather than glossing: the failure you saw at `codex-routing.test.ts:583` did not reproduce here on Bun 1.4.0, filtered or as a set. I am not disputing your run — the fragility is real either way, since a stale hardcoded writer generation can short-circuit the guard, and that is the actual defect. I fixed the cause instead of leaving the test resting on a runtime difference. If you still see order dependence on 1.3.14 after this commit, tell me and I will chase it with your runtime rather than mine.

Re-requesting review.

lidge-jun:
Confirmed and fixed in `bd26bcd39`. The failure-count inheritance was real: after a G1 401, a G2 replacement, and a genuine G2 503, `consecutiveFailures` was **2** where it should be 1, so the account reached the failover threshold a failure early.

Reader-side spending could not catch it. The transient and workspace branches derive their new entry from the current one, so the spent 401 donated its count *and* the write dropped the provenance tag on the way out — after which no read had anything left to detect. `recordCodexUpstreamOutcome` now spends a stale credential failure once, before any branch inspects health, so transient, workspace and quota all start from evidence that still describes a live credential.

The regression covers both inheritance paths you implied, transient and workspace denial. Removing the entry-point spend turns it red with exactly 2 versus 1.

One process note, since it nearly produced a false negative: my first mutation attempt patched the wrong `dropSpentCredentialFailure` call — the reader inside `getCodexUpstreamHealth` instead of the new entry-point one — and the test stayed green. That looks identical to vacuous coverage. Mutating the intended site fails it correctly, and I would rather flag that than quietly report "mutation-proven".

Your other two comments on this thread were already addressed on earlier heads: the empty-identity requirement landed in `5a3795769` (both ids must be non-empty, and propagation is skipped entirely when the owner identity is absent), and the delete-later-evidence case is what the per-entry provenance in `9acb7b593` fixed.

497 pass / 0 fail across the seven-file set as a set. Typecheck and `privacy:scan` clean.

Ingwannu (changes requested):
Reviewed exact head `5a37957691e7dc9025f1456f862ac59627fbf44c` against direct parent `dev@3a9835ca91d34affe97a4028b616d33e86d09d5b`.

The previous blockers are fixed correctly on this head: empty account ids fail closed, the sidecar test now proves the 401 was actually applied before replacement, per-entry credential provenance prevents a later 503 from being deleted, alias plan reconciliation uses each alias generation, and the same-grant adoption path requires non-empty identity equality. The seven focused suites pass 496/496 with 7,821 expectations, exact-lockfile TypeScript no-emit passes, and exact-head hosted CI is green.

One correctness blocker remains before approval:

- `recordCodexUpstreamOutcome` does not spend a stale credential-failure entry before later outcome branches derive their counters from `upstreamHealth`. Exact-head reproduction: record a 401 for generation 1, persist generation 2, then record the first genuine generation-2 503. `getCodexUpstreamHealth` returns `{ consecutiveFailures: 2, lastFailureStatus: 503 }`; the expected generation-2 count is 1. The later write drops the provenance tag but inherits the old credential failure count first, so generation 2 reaches the configured failover threshold one real failure early. The workspace branch has the same inheritance pattern. Call `dropSpentCredentialFailure(accountId)` after classifying the outcome and before any branch reads or writes `upstreamHealth`, then strengthen the G1-to-G2-to-503 regression to assert `consecutiveFailures === 1` and that a second G2 503 advances it to 2.

Please also update the PR body before re-review. It still describes the deleted post-write rollback and test seam, and reports the old 420-test evidence, while the current implementation uses read-time generation-tagged health/reauth state and the seam is gone.

No repository-wide security scan was run.

Ingwannu:
The code blocker from my latest review is resolved on `bd26bcd393d07f6a25f40b3a104c54a5f646cc82`, and the exact-head focused verification is green. Before I submit approval, please update the PR description to match the implementation that now ships: generation-owned health/reauth provenance, entry-point spending before derived counters, stored-pool sidecar generations, non-empty identity binding, shared-flight plan settlement, and 497 focused tests. The current body still describes the deleted post-write rollback and `setCodexCredentialRevalidationSeamForTests`, and still reports the old 420-test evidence. The exact-head macOS job is also still in progress; I will re-check once both the description and that job are current.

Ingwannu (changes requested):
Re-review of exact head `bd26bcd393d07f6a25f40b3a104c54a5f646cc82`: the last code blocker is fixed, the seven focused suites pass `497/497` with `7,823` expectations under isolated homes, exact-lockfile TypeScript no-emit passes, and every exact-head hosted check is now green, including macOS. I found no remaining code blocker on this head.

One merge-blocking documentation issue remains. The PR description still says health has no generation, reauth is a bare id set, effects are restored by a post-write re-read, and `setCodexCredentialRevalidationSeamForTests` is the chosen design; it also still reports the old `420`-test evidence. None of that is true on the current head. The implementation now owns credential provenance on the exact health/reauth entry, spends stale evidence at outcome entry before deriving counters, passes stored-pool sidecar generations, fails closed on empty/foreign identity, and settles plans through the shared flight. The seam and rollback design were removed.

Please update the PR body and mutation/evidence table to describe that current design and the `497`-test exact-head result. Because this is the security rationale for credential quarantine and rotation, leaving the superseded design in the merge record is not acceptable. No code push or CI rerun is needed if only the description changes; once it is corrected, this same SHA is ready for approval.

lidge-jun:
@Ingwannu you were right that the description was describing a design that no longer exists on this head, and on a credential-quarantine change that is the part of the merge record people will read first. The body is rewritten against `bd26bcd393d07f6a25f40b3a104c54a5f646cc82`. No code push, so this is the same SHA and CI did not need to rerun.

What the description now says, and what it stopped saying:

- Health **owns** credential provenance: `credentialFailureGeneration` is a field on the entry, and the body explains why a side map keyed by account id is wrong (it would spend a later unrelated 503).
- `reauthAccounts` is a `Map<string, number | undefined>`, with `undefined` kept as the stronger account-wide claim for a login flow.
- Stale evidence is **spent at read time**, including at `recordCodexUpstreamOutcome` entry so a dead 401 cannot donate its failure count. The post-write re-read is described as abandoned, and it credits your reproduction of the surviving ordering as the reason.
- `preservedCooldownFields` dropping the tag, and the stored-pool sidecar generations, are both stated.
- Empty or absent `chatgptAccountId` **fails closed**.
- Alias plans settle on the shared flight from #2933.
- `setCodexCredentialRevalidationSeamForTests` is documented as deleted, with the reason it stopped being necessary. The rollback design is gone from the body entirely.

Evidence is now the exact-head result rather than the stale `420`: **465 pass / 0 fail / 7,665 expectations** across eight focused suites under isolated homes, with a per-suite table. I measured the eight I run; your `497/7,823` is a superset I did not reproduce, so I quoted my own number rather than yours.

The mutation table is rebuilt against the current implementation - it now includes the fail-closed identity mutation, the `recordCodexUpstreamOutcome`-entry spend site, and both directions of the fence. I also left in the two places my own test work was weak: one assertion was vacuous until the fixture diverged owner and alias generations, and one mutation patched the wrong `dropSpentCredentialFailure` call site and looked vacuous because of it.

Ready for another look on the same SHA.

Ingwannu (approved):
Approved exact head `bd26bcd393d07f6a25f40b3a104c54a5f646cc82`.

The rewritten description now matches the implementation that was already validated: credential provenance is owned by the health/reauth entries, spent evidence is removed before new counters are derived, stored-pool generations are propagated, empty or foreign identities fail closed, alias plans settle on the shared flight, and the abandoned rollback/test-seam design is explicitly identified as removed.

The author's independently measured 465-test focused set is clearly distinguished from my 497-test superset rather than being presented as the same run. The exact-head code, focused evidence, and hosted CI were already green, and no code changed after that review. I have no remaining blocker.

lidge-jun:
Correction to my own merge record. I had an independent audit re-derive every claim in the description against `bd26bcd393d07f6a25f40b3a104c54a5f646cc82` after this merged, and **two entries in my mutation table do not reproduce as written**. Since the point of that table is to prove the credential fences are real, a claim that does not reproduce is worse than no claim, so I am recording the correction here rather than leaving it in the history.

**1. "tombstone guard removed → resurrection test red" is wrong.** Deleting the `alias.deletedAt != null` check in `src/codex/account-store.ts` leaves the named test **green**:

```text
(pass) a tombstoned same-grant record is not resurrected by grant propagation
1 pass, 0 fail
```

The reason is that a tombstone in that fixture also has no credential, and the separate `!alias.credential` guard skips it first. So the test proves "no resurrection", but it does **not** independently prove the `deletedAt` check is load-bearing — the two guards overlap on the only fixture that exercises them. Removing both would not fix the gap either, because it would then fail on a missing-credential dereference rather than on resurrection. Proving that guard needs a tombstone that still carries a credential, which no current fixture builds.

**2. "never spend the tag → live red, stale green" is wrong.** Under both literal readings of that mutation at `src/codex/routing.ts:148` — always spend, or invert the predicate — **both** tests go red, not the claimed opposite pair. The stale-evidence test starts by asserting the initial live 401 health exists (`tests/codex-routing.test.ts:500`), so any mutation that spends unconditionally breaks that first assertion before the interesting one is reached.

What *is* proven, and reproduced independently: dropping `credentialFailureGeneration` turns the stale test red while the live-credential test stays **green**. That is the real conditional-fence evidence, and it does support the fix. The symmetric pair was my over-claim on top of it.

Also two wording corrections. "Every hosted check is green" should be "every required and executed check" — one generated Windows matrix entry is reported `SKIPPED` by the runner-selection gate. And "closes the last two open gaps" reads as fuller closure than the body then describes: gap 4 is closed, gap 3 is closed only for provably-untouched duplicate aliases. #2892's closing comment states that split, and mixed aliases remain unimplemented.

Everything else was confirmed against source, including all three `dropSpentCredentialFailure` sites, the health-entry field, the reauth map precedence, the cooldown tag removal, the three pool-only sidecar fences, fail-closed empty identity, alias settlement through `propagatedAliases`, `recordCodexUpstreamOutcome` being synchronous and lock-free, the absence of the old test seam, and all eight suite counts (163 + 194 + 41 + 20 + 13 + 17 + 10 + 7 = 465 pass / 7,665 expectations). Seven of the nine mutations reproduced red exactly as described.

I am tracking the tokenful-tombstone fixture as follow-up work so that guard gets its own proof.

PR #2655, opened 2026-08-26
feat(providers): add opt-in transient-5xx retry policy for key-auth providers

TooSpace:
## Summary

Add `transientRetryOn5xx` provider config field that allows any key-auth provider to opt into `fetchWithTransientRetry` (500/502/503/504/520/521/522) with exponential backoff. Previously only the Google adapter had this protection via a hardcoded scope guard (#1851).

**Motivation:** Custom providers behind shared corporate gateways often return 503 `server_is_overloaded` under load instead of 429. Without transient-5xx retry, these transient errors kill the Codex turn immediately instead of surviving a burst.

**Scope (narrowed per review on #2643/#2655):** first version exposes only `enabled` and `attempts`. Delays stay on the shared safe constants in `lib/upstream-retry.ts` (400 ms base, 5 s cap, `Retry-After` honored). No `baseDelayMs`/`maxDelayMs` user config — the earlier draft declared them but never wired them into `TransientRetryOptions`, which would have been dead config.

## Changes

- `src/types/provider.ts`: `TransientRetryPolicy` (`enabled`, `attempts`) + `OcxProviderConfig.transientRetryOn5xx`
- `src/providers/key-failover.ts`: `transientRetryPolicyFor()` resolver (fail-closed for non-key auth modes), mirroring `rateLimitRetryPolicyFor()`
- `src/config.ts`: strict zod schema on the provider schema, a load-time sanitizer mirroring `sanitizeRetryOn429ForLoad` (invalid fields dropped with warnings; an invalid master switch drops the whole policy so a malformed disable stays disabled), and `transientRetryOn5xxPolicyConfigError` for the management write boundary
- `src/server/responses/core.ts`: policy-driven retry selection replaces the hardcoded `adapter === "google"` check on both the initial send and the continuation send (Google exception preserved)
- `src/server/chat-native.ts`: same policy-driven selection on the `/v1/chat/completions` send path
- `tests/transient-retry-policy.test.ts`: 12 tests — resolver defaults/fail-closed, write-boundary validation, and fetch-path coverage through `handleResponses` (503-then-success retries to 200, exhaustion surfaces 503, no-opt-in and `enabled:false` short-circuit with one send, 400 never retried)

## Verification

- `bun run typecheck` — pass
- `bun run test tests/transient-retry-policy.test.ts` — 12 pass, 0 fail
[...]

PR #2977, opened 2026-08-30
docs(devlog): record the final gate for the cursor tool-continuation unit

lidge-jun:
## Summary

- Add `080_final_gate.md` to the closed Cursor tool-continuation unit: the record of an independent final-gate audit of the landed fix on `dev`. Verdict pass, no findings.
- The unit's eleven audit rounds all reviewed the change while it was being built, against a plan the same session wrote. This gate answers a narrower question those rounds structurally could not: does the landed code hold up to a reviewer who did not build it, and is every claim in the written record true against git rather than against memory. The reviewer was given the three already-accepted scope-outs up front so it could not bill known tradeoffs as new findings, and was told a pass was acceptable so it had no incentive to manufacture one.
- The gate produced one correction worth keeping. The fix does not emit a separate assistant `[Tool Call]` root before a replayed result — it names the invocation *inside* the result envelope as an `invoked: <tool> with <args>` line. That is deliberate: a standalone marker gets few-shot-mimicked by the model and breaks multi-tool continuations, which is exactly what the remote suite's 363-B guard caught when this unit first tried that shape. So the invariant is "no replayed result root lacks its invocation line", and the earlier call-precedes-result framing is misleading.
- Coverage is measured rather than asserted: six mutations each reddening on-point tests against a 166 pass / 0 fail baseline, plus the round 11 note threshold reproducing in both directions (relaxing reddens 4, tightening reddens 1). A suite that can tell a correct bound from an unnecessarily strict one is what rounds 5 through 10 lacked. Three hostile shapes attacked the `knownCallsOffset > 0` checkpoint skip premise directly — a covered prefix with no user turn, `suffixStart = 1` behind an assistant, and a cut falling between a call and its result — and none produced an orphan, because the invocation is keyed by call id over full history.
- It records one behaviour without filing it as a defect: when two calls share a decoded call id, the ambiguous id is dropped and both results replay with no invocation line. The code argues that tradeoff explicitly — a wrong invocation is undetectable by the model, a missing one is honest — and upstream Codex call ids are unique.

Documentation only: one new file, 78 insertions. No file under `src/`, `gui/`, `scripts/`, or `tests/` is touched, and nothing in the build, typecheck, or test path reads `devlog/`.

## Verification

- `bun run privacy:scan` — passed. This is the gate that reads `devlog/`, so it is the one that matters for a diff made entirely of devlog prose.
[...]

PR #2975, opened 2026-08-30
test(windows): close the four residual shard-2 Windows failures

lidge-jun:
## Summary

The last four Windows-only failures from the dispatched leg. After #2968 and #2972, shards 1, 3 and 4
pass on run [33289201339](https://github.com/lidge-jun/opencodex/actions/runs/33289201339); shard 2 held
these four. `platform-windows` runs only on `workflow_dispatch`, so none of them was visible to the CI
that gates `release.yml`.

## Four failures, three distinct mechanisms

**`launchd service plist > the generated launcher command follows a retargeted shim…`** — a false
platform assumption the earlier serialization fix in #2968 did not reach. That commit corrected the path
*assertions*, but the test also *executes* an extensionless `#!/bin/sh` fixture, and Windows cannot run
that as a native executable. The fixture is now a cross-platform Bun redirect, keeping the version
retargeting, old-version deletion, spaces-and-apostrophes and path-serialization coverage it had.

**`ocx init piped stdin (#754) > exits cleanly when stdin closes before the first prompt answer`** — the
8-second race started at process *spawn*, so Windows startup and import time was being charged against
EOF handling. It now waits for the first prompt to appear, closes stdin while readline is genuinely
awaiting an answer, and waits on exit directly. The budget no longer measures interpreter startup.

**Two `storage mutation coordinator` cases** — both were sleep-based assumptions that a loaded Windows
runner invalidates:

- `renameNoReplace()` is `linkSync` then `unlinkSync`, so there is a real interval where *both* names
  exist. The test treated destination existence as move completion and could observe that interval.
- The other slept 80ms and assumed the cleanup lease had been acquired. A preceding failed restore also
  left an unresolved request, which matches the `ECONNRESET`-then-401 cascade in the Windows log.

Both now use an explicit handshake — the worker publishes only after every move completes, and restore is
attempted only once the cleanup lease is confirmed held. Teardown always releases the worker and drains
the pending request. No timeout was raised and nothing was skipped: a longer sleep would have hidden the
`linkSync`/`unlinkSync` window rather than closing it.

## Verification

Based on `dev@aa16a71e0`.

- `bun test tests/service.test.ts tests/init-eof.test.ts tests/storage-mutation-race.test.ts` → **154 pass, 0 fail**
- `bun x tsc --noEmit` → clean

Two of the four could not be reproduced exactly on POSIX — the Windows scheduling delay and the 401
cascade specifically — so what each fix removes is the *incidental* dependency (spawn time, wall-clock
[...]

PR #2981, opened 2026-08-30
feat(providers): add opt-in transient-5xx retry with a shared total-send budget

### LINKED ISSUE #2643: [Feature]: opt-in transient-5xx retry for key-auth providers on openai-chat adapter
### Area

Provider adapters

When a custom provider (e.g. an internal corporate LLM gateway) returns transient 5xx errors (502/503/504), the proxy should retry the request with exponential backoff before surfacing the error to the client. Currently this only works for the Google AI Studio adapter (fetchWithTransientRetry is gated behind route.provider.adapter === "google" in responses/core.ts), but it should be available to any key-auth provider via opt-in configuration.

### What prevents this today?

The #1851 scope guard in src/server/responses/core.ts (lines 4597-4601 and 5059-5061) deliberately limits fetchWithTransientRetry to adapter === "google". Every other adapter keeps fetchWithResetRetry which only retries ECONNRESET/EPIPE connection resets, not HTTP 5xx responses. This means:

- A 503 server_is_overloaded from a custom provider kills the turn immediately instead of waiting and retrying.
- Users of providers that experience transient overload (e.g. shared corporate gateways) see sessions die mid-way instead of surviving a burst.

The retryOn429 config exists for 429 rate limiting but covers a different failure class. There is no mechanism to opt into transient-5xx retry for non-Google adapters.

### What should OpenCodex do?

Introduce an opt-in provider config field (e.g. transientRetryOn5xx: { enabled: true, attempts?: number, baseDelayMs?: number, maxDelayMs?: number }) that allows any key-auth provider to enable fetchWithTransientRetry on the openai-chat adapter path. When enabled, the proxy should:

1. Catch transient upstream statuses (500/502/503/504/520/521/522) on the initial request before streaming starts.
2. Wait with exponential backoff (respecting Retry-After header), capped by a configurable max delay.
3. Retry the identical request up to the configured attempts.
4. Only surface the error to the client if all retries are exhausted.

The retry should be opt-in (disabled by default) so existing providers are not affected. The fetchWithTransientRetry implementation already handles Retry-After headers, heartbeats to keep stall watchdogs fed, and slow-attempt detection (returning a slow 502 as-is rather than retrying past the client timeout).

The same treatment should also apply to the chat-native.ts path (/v1/chat/completions), which currently uses fetchWithResetRetry for all non-Google adapters.

\`\`\`json
// ~/.opencodex/config.json
{
  "providers": {
    "my-gateway": {
      "adapter": "openai-chat",
      "baseUrl": "http://gateway.internal/v1",
      "authMode": "key",
      "apiKey": "sk-...",
      "transientRetryOn5xx": {
        "enabled": true,
        "attempts": 3,
        "maxDelayMs": 5000
      }
    }
  }
}
\`\`\`

Before: A 503 from gateway.internal immediately kills the Codex turn with "Provider error 503".

After: The proxy waits ~400ms, retries; if another 503, waits ~800ms, retries again; only then surfaces the error. The turn survives transient overload.

1. Wrap the provider in a local retry proxy (e.g. nginx or a small Node script). This adds operational complexity and splits the timeout budget between two hops.
2. Use a different adapter. Not applicable — openai-chat is the correct adapter for OpenAI-compatible endpoints.
3. Rely on retryOn429. Inapplicable — 429 is rate-limiting, not server overload. Many corporate gateways return 503 under load rather than 429.

lidge-jun:
## Summary

Adds opt-in transient-5xx retry for key-auth `openai-chat` providers, and fixes a latent budget bug that this feature would have activated. Closes #2643. Re-implements PR #2655 by @TooSpace on current `dev` (it was 76 commits behind and its two most-affected files had moved substantially).

**The retry budget was multiplicative.** `fetchWithTransientRetry` forwarded its whole options object — `attempts` included — into every nested `fetchWithResetRetry`, so the two recovery layers multiplied: `attempts: 3` allowed 3 transient rounds each independently retrying 3 connection resets, up to **9** upstream sends, and `attempts: 10` up to **100**.

The existing doc comment already named the hazard and noted it was inert because *"no caller passes it today."* This PR's provider policy is the first caller that does, so shipping the feature without the fix would have converted a documented latent note into live behavior — and multiplying load against an already-failing provider is worse than not retrying at all. A counted fetch wrapper now increments a shared send count before each await and passes only the remaining budget inward.

**The feature.** `providers.<name>.transientRetryOn5xx` retries pre-stream 500/502/503/504/520/521/522 across all three send paths: the initial Responses request, the terminal-guard continuation, and native `/v1/chat/completions`. Disabled unless present; a bare `{}` opts in with defaults. Scope is key-auth `openai-chat` only — the resolver checks the adapter explicitly rather than letting any generic key-auth provider inherit it, and auth mode fails closed the same way `rateLimitRetryPolicyFor` does. The legacy direct-Google exception is preserved exactly.

`attempts` is documented and implemented as a **total send budget** including the first request, not a per-layer retry count.

## Verification

Run on Linux (bun 1.3.14) at `2107f64d4`:

- `bun run typecheck` → exit 0
- `bun test tests/upstream-transient-retry.test.ts` → **13 pass, 0 fail**
- `bun test tests/upstream-transient-retry.test.ts tests/upstream-retry.test.ts tests/core-lab-boundary.test.ts tests/config-user-edits.test.ts` → **93 pass, 0 fail**

Coverage added: the budget is pinned by an all-503 case asserting exactly 3 sends (never 9) and a mixed `ECONNRESET`+503 case proving both layers draw from one pool; resolver tests cover the off-by-default states, bare-`{}` opt-in, and every rejected adapter and auth mode.

## Checklist

Core/Lab boundary: this touches `src/server/responses/core.ts`, but both call sites extend the **existing** `key-failover` import, so no new module edge is created. `tests/core-lab-boundary.test.ts` passes. `src/router.ts` and `src/server/lifecycle.ts` are untouched.

Scope notes: no `baseDelayMs`/`maxDelayMs` knobs, no other adapters or auth modes, no mid-stream replay, no 429 behavior change, and no dashboard/PATCH editing — `src/server/management/provider-routes.ts` is unchanged, matching the accepted first-version scope.

Planning unit: `devlog/_plan/260830_pre_release_backlog_ten/090_wp10_issue2643_transient_retry.md`.

## Summary by CodeRabbit

- **New Features**
  - Added optional transient 5xx retry support for key-auth `openai-chat` providers.
  - Retries cover eligible pre-stream errors across supported chat and response requests.
  - Added configurable enablement and attempt limits, with exponential backoff and `Retry-After` support.
  - Retry attempts share a single total request budget to prevent excessive upstream requests.
  - Mid-stream failures remain unreplayed and 429 handling is unchanged.

- **Documentation**
  - Documented the new `transientRetryOn5xx` provider configuration option.


lidge-jun:
## 리뷰 · 우선순위 74 / 80

설명

지금 `dev` HEAD는 `223a0a287` (#2975, Windows shard-2 잔여 실패 네 건 정리)이다. 그 앞줄에는 #2977 최종 게이트 기록, #2974 릴리즈 readiness, #2965 용량 패널 같은 마무리 작업이 있다. 이 PR은 pre-release backlog WP10으로, 이슈 #2643을 닫는 **실사용 복원력** 항목이다. 기여자 @TooSpace의 #2655가 `dev`보다 76커밋 뒤처져 핵심 파일이 많이 움직인 뒤라서, 메인테이너가 현재 `dev` 위에 다시 구현한 것이다.

핵심은 두 겹이다. 첫째, 이미 `dev`에 있는 `src/lib/upstream-retry.ts`의 `fetchWithTransientRetry`는 안쪽 `fetchWithResetRetry`에 `opts` 전체를 그대로 넘긴다. 그래서 `attempts: 3`이면 바깥 5xx 재시도 3번 × 안쪽 연결 리셋 3번이 되어 **최대 9번**, `attempts: 10`이면 최대 100번까지  Upstream으로 나갈 수 있다. 예전 주석도 “지금은 호출자가 attempts를 안 넣어서 괜찮다”고 적어 두었는데, 이 기능이 바로 그 첫 호출자라서, 고치지 않고 켜면 문서에만 있던 위험이 실제로 살아난다. 이 PR은 `countedFetch`로 보내기 전에 공유 카운터를 올리고, 안쪽으로는 **남은 예산만** 넘긴다. 테스트도 전부 503일 때 정확히 3번, ECONNRESET+503 섞여도 3번을 고정한다.

둘째, `providers.<name>.transientRetryOn5xx`로 key-auth `openai-chat`만 선택 가입한다. 빈 `{}`면 기본(attempts 3)으로 켜지고, 없으면 꺼진 상태다. 적용 지점은 Responses 첫 전송, terminal-guard 이어쓰기, native `/v1/chat/completions` 세 곳이다. `transientRetryPolicyFor`는 어댑터를 명시적으로 `openai-chat`만 통과시키고, authMode는 `rateLimitRetryPolicyFor`와 같이 key(또는 생략 기본)만 허용한다. 예전에 있던 직접 Google AI Studio 예외(`adapter === "google"`)는 그대로 두고, Google에는 attempts를 새로 넘기지 않아서 기본 예산만 쓰게 한다. 다만 예산 공유 수정 자체는 Google 경로에도 같이 들어가므로, “레이어 곱셈” 버그는 인프라 공통으로 사라진다.

검증 서술은 typecheck, `tests/upstream-transient-retry.test.ts` 13 pass, 주변 재시도/코어-랩 경계/설정 편집 합쳐 93 pass다. `tests/core-lab-boundary.test.ts`를 통과시켜 `responses/core.ts`에 새 모듈 가장자리를 만들지 않았다고 못 박았다. 문서 `docs-site/.../configuration/providers.md`에 옵션 한 줄을 추가했다. 관리 GUI PATCH(`provider-routes.ts`)는 의도적으로 안 건드렸다.

라인 수준

`src/lib/upstream-retry.ts` `fetchWithTransientRetry` - attempts를 총 송신 예산으로 바꾸고 countedFetch로 리셋 레이어와 공유한다. 곱셈 버그의 본체 수정이다.
`src/lib/upstream-retry.ts` `remaining = Math.max(1, budget - sent)` - 루프 조건(`sent < budget`)이 멈추는 장치라서, 예산이 다 찬 뒤에도 1을 넘기지 않게 맞춰 둔 설계다. 다만 나중에 루프 조건을 바꾸면 이 floor가 추가 1회를 허용할 수 있으니 주석/불변조건을 유지해야 한다.
`src/providers/key-failover.ts` `transientRetryPolicyFor` - openai-chat + key-auth만. 어댑터 게이트가 없으면 다른 key-auth가 같이 켜질 수 있어서 범위 고정이 맞다.
`src/server/responses/core.ts` / `src/server/chat-native.ts` - 세 송신 경로에 동일 정책. Google OR transientPolicy 분기는 레거시 예외를 보존한다.
`src/config.ts` `transientRetryOn5xxPolicySchema` - attempts 1..10. retryOn429보다 천장을 낮춘 선택이 타당하다.
`tests/upstream-transient-retry.test.ts` - 예산 고정 + resolver 거부 케이스. “9가 아니라 3”을 숫자로 못 박은 점이 좋다.
기여자 PR #2655 - 같은 이슈의 오래된 구현. 이 PR이 랜딩되면 닫는 대상이다.

메인테이너의 판단이 필요한 지점

- #2655를 `landed-via-maintainer`로 바로 닫을지, TooSpace에게 크레딧 코멘트만 남기고 닫을지.
- `baseDelayMs`/`maxDelayMs`와 대시보드 PATCH 편집을 v1에서 계속 빼 둘지. 본문 범위와 맞지만, 운영자가 GUI로 켜고 싶어하면 후속이 필요하다.
- Google 경로에 예산 공유 수정이 같이 들어간 것을 “의도된 인프라 수정”으로 확정할지. 동작은 안전해지지만 이번 이슈 제목 범위 밖이다.

너의 추천

merge 하라. #2643을 닫고, 열린 #2655는 이 랜딩 커밋으로 `Landed via #2981` 코멘트 후 `landed-via-maintainer`로 닫아라. 지연/GUI 노브는 후속 이슈로  남기면 충분하다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
The opt-in shared-send-budget design is useful, but exact head `2107f64d4362184b425a6d11edf49b5fb666fe25` leaves one live Responses dispatch leg outside that budget.

After an initial 429, the Responses `retryOn429` path enters `rebuildAndRefetch`. If that replay then receives a retryable 503, the helper calls `fetchWithHeaderTimeout` directly instead of the selected transient-retry wrapper, so the same provider policy works on native chat but is silently bypassed on Responses recovery. Route every actual send, including 429/account-recovery refetches, through the one request-scoped budget owner and add a `429 -> 503 -> 200` regression proving the total send count stays bounded.

Also keep the user-facing `retryTransient5xx` option synchronized across the localized provider-configuration references, not only the English source. Exact-head CI is green, but it does not cover this recovery-loop composition boundary.

PR #2982, opened 2026-08-30
feat(anthropic): add quota-window account pool routing (carry of #2560)

### LINKED ISSUE #2539: [Feature]: Allow Anthropic account-pool quota routing to select by weekly usage
### Area

I run the Claude OAuth account pool (`anthropicAccountPool.enabled: true`) with several Claude subscriptions, and I want new sessions to be spread evenly across those accounts over a rolling **7-day** window rather than only over the **current 5-hour** window.

Anthropic enforces two limits per account at the same time: a 5-hour rolling window and a weekly (7-day) rolling window. On a heavy coding day the 5-hour numbers of two accounts can look nearly identical while their weekly budgets are far apart — one account may be at 20% weekly and the other at 85% weekly. Today the pool treats those two accounts as interchangeable, so it keeps feeding the account that is close to burning its weekly allowance. By Thursday that account is weekly-limited for the rest of the week while the other one still has most of its weekly budget unused.

The outcome I need is: **let the operator choose which quota window the pool optimizes for**, so a multi-day workload finishes the week with the accounts drained roughly evenly instead of one account exhausted and the rest idle.

### What prevents this today?

The pool's `quota` routing strategy scores accounts **only by the 5-hour window**. On current `dev` (`98ed186c7`), `src/oauth/anthropic-routing.ts` has:

```ts
function hasKnownUsage(accountId: string): boolean {
  const quota = getCachedProviderAccountQuota(PROVIDER, accountId);
  return typeof quota?.fiveHourPercent === "number" && Number.isFinite(quota.fiveHourPercent);
}

function usageScore(accountId: string): number {
  const quota = getCachedProviderAccountQuota(PROVIDER, accountId);
  if (!quota || typeof quota.fiveHourPercent !== "number" || !Number.isFinite(quota.fiveHourPercent)) {
    return UNKNOWN_USAGE_SCORE;
  }
  return Math.max(0, Math.min(100, quota.fiveHourPercent));
}
```

`fiveHourPercent` is the only usage field the whole module reads — it appears at lines 147, 152, and 155, and `weeklyPercent` appears nowhere in the file. Every usage-aware decision therefore inherits the 5-hour-only view: the lowest-usage pick for a new session, the `fill-first` "is the active account still under threshold" check, and the `autoSwitchThreshold` comparison.

The weekly number is **already collected and cached** — `src/providers/quota.ts` carries `weeklyPercent` on the per-account quota record (declared at line 94, populated at lines 518/664/713/749, and surfaced in the dashboard and `ocx` account output). So the data the router would need is present and fresh; it is simply never consulted when choosing an account.

There is no configuration escape hatch either. `AnthropicAccountPoolConfig` (`src/oauth/anthropic-routing.ts:45-53`) exposes only `enabled`, `autoSwitchThreshold`, `strategy`, and `stickyLimit` — nothing selects the window. The only workarounds available to an operator today are to manually disable an account before it burns its weekly budget, or to set `autoSwitchThreshold` very low, which changes 5-hour churn behaviour without ever looking at the weekly number.

### What should OpenCodex do?

Add an **opt-in** `quotaWindow` setting to `anthropicAccountPool` that selects which quota window usage-aware selection scores against:

- **`"five-hour"` (default)** — exactly today's behaviour. When `quotaWindow` is absent, unset, or unrecognized, account selection must be byte-for-byte identical to current `dev`. Existing installs must observe no change whatsoever.
- **`"weekly"` (new, opt-in)** — rank candidates by their weekly usage percentage so a multi-day workload drains accounts evenly across the 7-day window.
- **`"max-utilization"` (new, opt-in)** — rank by the higher of the two known percentages, i.e. by whichever window is closer to its limit, for operators who want the most conservative pick.

Behaviour that must be preserved and constrained:

1. **Session affinity is unchanged.** The window only affects which account a *new* session binds to. A live session keeps its account exactly as it does today; there is no mid-session rebalancing.
2. **429 cooldown and failover are unchanged.** Cooldown, quarantine, reauth handling, and the per-request failover cap keep their current semantics. Selection must never return `null` in a case where it returns an account today.
3. **Weekly mode must not pick a 5-hour-exhausted account.** An account whose 5-hour window is already at 100% is unusable *right now* no matter how good its weekly number looks. In `weekly` mode such accounts must be skipped — unless skipping them would leave no candidate at all, in which case the existing behaviour wins and an account is still returned.
4. **Unknown usage never wins.** An account with no collected/parsed value for the configured window must sort **last**, never first, exactly as unknown 5-hour usage does today (`UNKNOWN_USAGE_SCORE`). A never-probed account must not be mistaken for an idle one.
5. **Deterministic ties.** When two accounts score equal in the configured window, break the tie by the lower 5-hour percentage; if that is still equal, keep the existing stable ordering so selection stays reproducible.
6. **Scope.** The setting belongs to the Anthropic pool only. It applies to the strategies that actually consult usage (`quota` and `fill-first`) and is inert under `round-robin`, which ignores usage by design. It should be reachable from the dashboard's account-pool settings and the management API alongside `strategy` and `autoSwitchThreshold`, and documented in the provider/Claude Code docs.

Configuration (opt-in; omitting `quotaWindow` keeps today's behaviour):

```json
{
  "anthropicAccountPool": {
    "enabled": true,
    "strategy": "quota",
    "quotaWindow": "weekly"
  }
}
```

Concrete before/after with two pooled Claude accounts:

| account | 5-hour usage | weekly usage |
| ------- | ------------ | ------------ |
| A       | 30%          | 85%          |
| B       | 35%          | 20%          |

**Before (today, and still the default):** a new session picks **account A**, because 30% < 35% in the 5-hour window — even though A has only 15% of its weekly budget left and B has 80% left.

**After, with `"quotaWindow": "weekly"`:** the same new session picks **account B**, because 20% < 85% in the weekly window, leaving A's remaining weekly budget for later in the week. If A's 5-hour window were instead at 100%, weekly mode would skip A regardless of its weekly figure; if B had never been probed and its weekly value were unknown, B would sort last rather than first.

- **Manual account toggling / a very low `autoSwitchThreshold`.** This is what I do today. It only manipulates the 5-hour window and requires me to watch the dashboard and disable an account by hand before it burns its weekly allowance; it is exactly the toil this proposal removes.
- **Always scoring by the weekly window (no config option).** Rejected: it would silently change selection for every existing pool user, and the current 5-hour behaviour was a deliberate choice in the pool implementation. Opt-in with a `"five-hour"` default keeps existing installs untouched.
- **Always scoring by "whichever window is closer to its limit".** Useful, but as a *forced* default it would surprise operators who intentionally want burst behaviour inside the 5-hour window. Offering it as the explicit `"max-utilization"` value gives that behaviour to operators who ask for it without imposing it.
- **A reset-time-aware "reset-pressure" score** (weighting usage by how soon each window resets, so an account 5 minutes from its reset is not treated the same as one 6 days out) was considered and **deliberately deferred to a separate follow-up proposal**. It needs reset-timestamp trust rules and clock-skew handling of its own, and folding it in here would make this change much harder to review. This proposal stays scoped to selecting an existing, already-collected window.

### Additional context

Verified related issues and pull requests on this repository (each checked with `gh` before filing):

- **#294** — `[Feature]: Claude account pool — parity with ChatGPT/Codex multi-account routing` (closed). The original request that established the Claude account pool as a goal, mirroring the existing Codex multi-account routing.
- **#493** — `feat(quota): per-account Claude rate limits, canonical 5-hour window` (merged). Established per-account Claude quota collection and made the 5-hour window canonical; this is also where the per-account quota record that already carries `weeklyPercent` comes from.
- **#578** — `feat(anthropic): opt-in Claude OAuth account pool (#294)` (merged). The pool implementation that introduced session affinity, 429 cooldown failover, and the lowest-5h-usage pick for new sessions — the exact code path this proposal makes window-aware.
- **#592** — `[Feature]: Account pool round-robin / fill-first rotation strategies` (closed). The request for explicit rotation strategies beyond quota-threshold routing, which is the precedent for adding an operator-selectable pool knob.
- **#593** — `feat(auth): account pool round-robin and fill-first strategies` (merged). Added the `quota` / `round-robin` / `fill-first` strategy setting; `quotaWindow` would sit next to `strategy` and follow the same opt-in, default-preserving pattern.
- **#695** — `[Feature]: Generic OAuth account-pool failover with session affinity and quota-aware selection` (open). Owns the generic pool selection/failover engine including quota-aware ranking; this proposal is the concrete Anthropic-side window question and should stay compatible with that engine's direction.

Code references are against `dev` at `98ed186c7`: `src/oauth/anthropic-routing.ts` (`hasKnownUsage` / `usageScore` at lines 145-156, `pickLowestUsage` at 194-208, `AnthropicAccountPoolConfig` at 45-53) and `src/providers/quota.ts` (`weeklyPercent`, line 94 and its population sites).

I am happy to implement this and open a pull request against `dev` if the direction is acceptable.

### Checks

lidge-jun:
## Summary

Maintainer carry of **#2560** by @Yoonkeee (19 commits cherry-picked onto current `dev`, author credit preserved), plus fixes for two findings from an independent security review. Closes #2539.

Adds opt-in Anthropic account-pool routing that selects by quota **window** — weekly or max-utilization — instead of only the rolling five-hour window.

**Why a carry branch.** The original PR was 18 commits behind and the reviewer's objection was explicitly a freshness one: *"the existing approval and hosted run predate material UI/document integration drift."* The overlap was in `docs-site`, nine GUI locale files, and `gui/src/styles.css`. A contributor branch cannot be pushed to by a maintainer, so the commits are re-applied here on current `dev`; #2560 can be closed with `landed-via-maintainer`.

**Security-review fix — the default path was changing.** The new comparator applied known-before-unknown ordering unconditionally, so an operator who never opted in still got different five-hour ordering: an account measured at 100% sorted ahead of an unmeasured one purely because it had a reading. The accepted scope for #2539 preserves the five-hour default exactly, so that rule is now gated to the opt-in windows. My first attempt gated it on `weekly` alone and the existing `max-utilization` test caught it — the correct predicate is "any window except the five-hour default".

A regression now pins the default path directly: under both the omitted and explicit five-hour spellings, a measured-100% account must not be promoted over an unmeasured one.

## Verification

Run on Linux (bun 1.3.14) at the current head:

- `bun run typecheck` → exit 0
- `bun test tests/anthropic-account-pool.test.ts tests/account-pool-management-api.test.ts tests/core-lab-boundary.test.ts` → **83 pass, 0 fail**
- `bun run privacy:scan` → passed

Security review (drafted in scratch, not committed — `devlog/` is public):

- **Candidate set:** no widening. Selection starts from stored accounts and filters reauth, cooldown, unusable credentials, and exclusions; the weekly fallback returns only to that same eligible set.
- **Disclosure:** none added. Management responses carry only the window enum, and routing warnings use hashed labels rather than raw account identifiers.
- **Default path:** was disproved, now fixed and pinned by regression.
- **Untouched:** `src/router.ts`, `src/server/lifecycle.ts`, `src/server/responses/core.ts`; no import reaches `src/lab/`.

## Known limitation, deliberately not fixed here

The review also found that routing treats any cached finite percentage as current: `getCachedProviderAccountQuota` drops the entry timestamp, and a failed probe preserves the old quota while refreshing its cache timestamp. Stale evidence can therefore influence selection.

That is **pre-existing behavior in the shared quota cache**, not something this PR introduces — it already affects five-hour selection today. Fixing it means changing the cache contract for every consumer, which is a larger change than this feature and deserves its own PR rather than being smuggled in behind a routing option. Flagging it explicitly so the reviewer can decide whether it blocks.

## Checklist

Planning unit: `devlog/_plan/260830_pre_release_backlog_ten/030_wp4_pr2560_anthropic_quota_window.md`.

## Summary by CodeRabbit

* **New Features**
  * Added configurable Anthropic account-pool quota windows: five-hour, weekly, and max-utilization.
  * Added quota-window selection to the settings interface, with localized labels and guidance.
  * Updated account selection, failover, recovery, and fill-first behavior to use the selected window where applicable.
  * Added API support for saving and retrieving the selected quota window.

* **Documentation**
  * Updated provider references and Claude Code guides across supported languages.

* **Bug Fixes**
  * Invalid quota-window values safely fall back to the default five-hour window.


lidge-jun:
## 리뷰 · 우선순위 73 / 80

설명

이 PR은 기여자 @Yoonkeee 의 #2560을 현재 dev 위에 다시 올린 메인테이너 캐리다.
지금 dev HEAD는 223a0a287 (#2975)이다. #2560은 18커밋 뒤처져 있고 리뷰 반대가 신선도 문제라서, 메인테이너가 커밋을 다시 적용하고 보안 리뷰 두 건을 고쳤다. 원본 #2560은 이 PR이 랜딩하면 landed-via-maintainer 로 닫으면 된다. 리베이스해서 살리지 마라.

하는 일은 Anthropic 계정 풀이 quota 전략일 때 어떤 사용량 막대를 볼지 고르게 하는 것이다. 지금은 5시간 막대만 본다. 옵션으로 weekly 와 max-utilization 을 연다. 기본값은 five-hour 그대로다. 이슈 #2539를 닫는 조각이다.

보안 리뷰에서 기본 경로가 바뀌는 문제가 있었다. known-before-unknown 정렬을 무조건 켜면, 옵트인하지 않은 운영자도 측정된 100% 계정이 미측정 계정 앞에 선다. 수용된 범위는 5시간 기본을 정확히 보존하는 것이다. 그래서 그 규칙은 five-hour 가 아닌 창에만 켠다. 회귀 테스트가 omitted/explicit five-hour 둘 다 고정한다.

타입은 src/types/config.ts 에 OcxAccountPoolQuotaWindow 를 두고 src/types.ts 가 다시보낸다. 분리 캠페인을 따른다. 닫고 리베이스할 대상이 아니다. 런타임은 src/oauth/anthropic-routing.ts, 관리 API는 src/server/management/oauth-account-routes.ts, GUI는 account-pool-strategy.ts 와 AnthropicAccountPoolSettings.tsx 다.

weekly 창은 5시간 100% 계정을 후보에서 빼되, 남은 사람이 없으면 다시 넣는다. max-utilization 은 그 필터가 없다. 점수는 두 막대 중 큰 값이라 5시간 100% 계정이 주간 20%여도 100점으로 뽑힐 수 있다. round-robin 은 창을 아예 안 본다. GUI 주석도 그 창은 RR에서 inert 라고 한다. fill-first 는 threshold 가 0보다 클 때만 창을 읽는다.

경로 gui/src/account-pool-strategy.ts 와 src/oauth/anthropic-routing.ts 의 normalizeAccountPoolQuotaWindow - 같은 이름이 두 패키지에 있다. GUI는 모르는 값을 five-hour 로 기본하고, 서버 PUT은 parse 가 null 이면 400이다. 동작은 다르니 이름이 같으면 나중에 헷갈린다.
경로 oauth-account-routes.ts GET quotaWindow - 값이 없어도 five-hour 로 정규화해 돌려준다. GUI가 그 값을 PUT 에 메아리치면 원래 없던 키가 설정에 기록된다.
경로 pickLowestUsage weekly 필터 - exhausted5h 는 weekly 에만 있고 max-utilization 에는 없다. 5시간 고갈 계정이 주간 점수가 낮아도 max 창에서 이긴다.
경로 pickUnboundStrategyAccount round-robin - weekly 의 5시간 고갈 제외가 RR 경로에는 없다. GUI도 RR에서 창을 꺼 두니 의도와 코드는 맞지만, weekly+RR 조합을 기대하는 운영자는 빈다.
경로 AnthropicAccountPoolSettings.tsx 인라인 marginTop - 새 힌트 두 줄이 style 객체다. 이 파일 주변에 CSS 클래스가 있으니 후속으로 옮겨도 된다. 머지 차단은 아니다.

메인테이너의 판단이 필요한 지점
- #2560을 이 PR 랜딩 뒤 landed-via-maintainer 로 닫을지. 원본을 리베이스하지 말 것.
- weekly 의 5시간 고갈 제외를 max-utilization 과 round-robin 에도 펼칠지, quota/fill-first 만으로 충분할지.
- GET 이 없는 quotaWindow 를 five-hour 로 채워 줄지, 생략해서 설정에 키가 안 생기게 할지.
- #2539를 이 PR closes 로 바로 닫을지.

너의 추천
이 캐리를 머지 후보로 둔다. 기본 5시간 경로는 테스트로 고정됐고, 타입 분리를 따른다. #2560은 리베이스하지 말고 랜딩 후 landed-via-maintainer 로 닫아라. weekly 의 5시간 제외를 다른 전략에 펼치는 것은 후속이다. CI가 초록이면 이 브랜치를 dev에 넣고 #2539를 닫으면 된다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
Reviewed exact head `0f86b4e9cdac04a631624a3ee13c3a34eba7aa67`. The runtime direction is valuable and the focused behavior is largely in place, but the user-facing contract is still inconsistent with the paths the implementation actually executes.

Please address these blockers before merge:

1. `gui/src/components/provider-workspace/AnthropicAccountPoolSettings.tsx` treats the quota window as inert for fill-first when `autoSwitchThreshold === 0`, and the enabled description says no quota routing is active. That is not true for all routing stages: threshold zero disables proactive usage-based switching, but new-session selection and 429 recovery can still consult the configured window. Make the UI stage-specific and add a regression covering fill-first + zero threshold.
2. The canonical configuration reference currently says `0` disables quota picking. Narrow that to proactive switching only; recovery/new-session behavior remains active. Also state that known-before-unknown ordering is limited to the opt-in `weekly`/`max-utilization` windows, while omitted or explicit `five-hour` preserves legacy ordering.
3. Keep the strategy contract precise across the Claude guide and every translated reference: fill-first evaluates its drain threshold in the selected `quotaWindow`; round-robin ignores quota scoring entirely; quota-window selection applies to new sessions and eligible 429 recovery as well as proactive rebinds. Several locales currently omit or contradict one or more of these points.

Please synchronize all supported locales and retain the existing runtime tests. Once those descriptions and the zero-threshold UI regression match the actual routing stages, this remains a strong merge candidate.

Issue #3011, opened 2026-08-30
[Bug][Windows] synchronous ACL hardening still stalls /healthz and poisons response spill writes after #612


chano-gpt:
## Client or integration

Codex App

## Area

Proxy and routing / Platform (Windows) / Service lifecycle

## Summary

On Windows, the proxy can keep port 10100 in `LISTEN` while `/healthz`, `/readyz`, and normal Codex requests stop responding for 10–47 seconds. The same process later recovers without a restart.

A timestamped health probe correlated one 40.956-second `/healthz` delay exactly with:

```text
[opencodex] ACL hardening timed out (ETIMEDOUT) — transient icacls stall; the volume may still support per-user NTFS ACLs; ACL state unverified (budget exhausted) — continuing without NTFS ACL harden
```

Issue #612 moved the response-state snapshot writer to an asynchronous ACL path. However, the released 2.36.0 source still has synchronous runtime paths:

- `src/config/paths.ts`: `hardenConfigDir()` calls `hardenSecretDir()`.
- `src/responses/spill-store.ts`: `writeResponseSpillDurably()` synchronously hardens the spill directory and temporary file.
- `src/lib/windows-secret-acl.ts`: the synchronous runner uses `Bun.spawnSync(...)`.

Because `/healthz` only constructs a local JSON response, a timeout while the port remains listening indicates that the Bun event loop itself is blocked.

There is also a functional consequence beyond liveness. Before restart, `GET /api/system/memory` reported:

```text
responseState.totalBytes:       66,982,525
responseState.count:            216
responseState.spillWrites:      849
responseState.spillWriteFailures: 2,032
responseState.tombstoneCount:   46
```

The spill directory contained no live spill files at inspection time. In `state.ts`, a spill-write exception increments `spillWriteFailures` and replaces the resident continuation with a tombstone. This can degrade `previous_response_id` replay even after the health endpoint recovers.

Expected behavior:

- ACL hardening must remain fail-closed where required.
- Runtime ACL subprocess waits must not block `/healthz` or unrelated proxy requests.
- A transient ACL timeout must not leave response spilling effectively failed for the remaining process lifetime.

## Reproduction

The failure is intermittent rather than deterministic on this machine.

1. Install OpenCodex globally on Windows and run it through the Task Scheduler service backend.
2. Keep the proxy running during normal Codex App usage.
3. Probe `http://127.0.0.1:10100/healthz` once per minute with a 10-second timeout while recording process memory and response-state metrics.
[...]

PR #3032, opened 2026-08-31
fix(responses): bound the durable spill directory with an aggregate byte cap

lifrary:
## Summary

`~/.opencodex/responses-state-spill/` has no aggregate size bound.

The response store has an unconditional RAM ceiling (`MAX_STORED_RESPONSE_BYTES`, 64 MiB) and demotes the oldest resident entry to a durable spill once it is crossed. Nothing bounded where those bytes landed: the spilled set was capped only per file (`MAX_RESPONSE_SPILL_PAYLOAD_BYTES`, 256 MiB) and per entry (`MAX_STORED_RESPONSES`, 1000). Their product is 250 GiB — larger than the disk of any host this runs on — so the only effective bound was `RESPONSE_TTL_MS`, which makes disk use a function of client request rate rather than of anything the proxy controls.

Measured on one macOS host, 2026-08-30: a client sending ~150 MB response states at ~1.4/min held 6.8 GB of that directory after 44 minutes, still climbing toward the ~12 GB an hour-long window implies. It filled the volume and unrelated processes began failing with `ENOSPC`. Retention itself was correct throughout — the TTL evicted that whole cohort an hour later — so this is a missing budget, not a leak.

This adds `MAX_SPILLED_RESPONSE_BYTES` (1 GiB) and evicts oldest-first past it in `pruneResponses`, immediately after the RAM demotion loop that creates the pressure. `deleteEntry` already routes through `deleteOwnedSpills`, so an evicted entry unlinks its file.

**Recomputed, not counted.** The spilled total is recomputed per prune rather than carried as a running counter. Spilled entries reach `states` through several insertion paths (demotion swap, direct oversized admission, snapshot reload), and one missed increment would silently disable the cap, where a walk over at most `MAX_STORED_RESPONSES` entries cannot drift.

**Scope of the bound.** It bounds the entries the store tracks. Spill files orphaned by a crash are not in `states`, are not counted here, and remain the existing `recoverOrphanedResponseSpills` grace-period sweep's job.

**No config key.** The three sibling bounds in this subsystem — count, TTL, per-file — are all bare constants, so this one is too. A `...Mb` key on the `appOwnedMemoryBudgetMb` pattern is an easy follow-up if you would prefer one.

**The value is the knob.** 1 GiB comes from that same sample (n=31), whose spilled sizes are strongly bimodal: median 1.1 MiB against a p90 of 198.7 MiB. At that median the count cap and this ceiling bind within 8% of each other (1000 x 1.1 MiB = 1.07 GiB), so ordinary traffic sees no eviction it would not already have seen and only the large tail is cut. Erring small is the safe direction: too low costs a replay miss, an already-handled path surfaced as `previous_response_not_found`; too high costs the host's disk. This is the one number here I would expect a maintainer to set differently.

**Docs.** The spill directory appeared nowhere in `structure/`, `docs/` or `docs-site/`, while the state-root inventory in `structure/00` describes its group as bounded caches. `structure/00` and `structure/02` now name it and its aggregate bound, following the existing prose + Decision Log pattern.

**Relation to #3018.** That PR also edits `pruneResponses`, but only inside the RAM demotion loop; this appends after that loop closes, so the two should not conflict textually. It is orthogonal in intent and introduces no aggregate bound.

## Verification

macOS, branch on `dev` @ `870a2adb`:

- `bun run typecheck` — exit 0, no output.
- `bun test tests/responses-state.test.ts` — 117 pass, 0 fail.
- `bun run test:changed` — 9885 pass, 0 fail across 542 files.
- `bun run test` — 16500 pass, 0 fail across 998 files.
- Red-green: with the eviction loop neutralised the new test fails `Expected: <= 20000 / Received: 48564`, and passes with it. Re-confirmed after the loop was simplified from a per-eviction rescan to a single pass.
- On the reporting host the proxy was restarted onto this change: the spill directory has since held at 993-1010 MiB against the 1 GiB ceiling across repeated samples, where it had reached 6.8 GB before. Eviction stops at exactly the number of ~199 MiB payloads that fit under the cap.

## Checklist

- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. Eviction unlinks only through `deleteResponseSpill`, which validates the ref before touching a path; no new file name is constructed from external input.

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - Added an aggregate 1 GiB limit for durable response-spill storage.
  - Older spilled responses are automatically removed first when the limit is exceeded.
  - Deferred spill data is cleaned up before active responses.
  - Spill usage is also enforced during state loading and periodic maintenance.
  - Newest retained responses remain available for replay, including after oversized spill entries are handled.

- **Documentation**
  - Documented spill-directory ownership, aggregate storage limits, eviction behavior, and orphan cleanup.


lifrary:
@coderabbitai review

lidge-jun:
## 리뷰 · 우선순위 71 / 80

이 PR은 지금 `dev` HEAD `870a2adb6` (패키지 2.37.0) 에 그대로 있는 디스크 예산 구멍을 막습니다. 응답 이어가기 저장소는 메모리 한도가 있습니다. `MAX_STORED_RESPONSE_BYTES` 는 64 MiB 입니다. 그 한도를 넘으면 제일 오래된 거주 항목이 `~/.opencodex/responses-state-spill/` 파일로 내려갑니다. 파일 하나 한도는 `MAX_RESPONSE_SPILL_PAYLOAD_BYTES` 256 MiB 이고, 항목 수는 `MAX_STORED_RESPONSES` 1000 입니다. 둘을 곱하면 250 GiB 입니다. 그 숫자는 이 프록시가 돌아가는 어떤 디스크보다 큽니다. 그래서 실제로 디스크를 막던 것은 `RESPONSE_TTL_MS` 한 시간뿐이었습니다. 클라이언트가 큰 이어가기를 빨리 보내면, 한 시간이 끝나기 전에 볼륨이 찹니다.

작성자가 2026-08-30 맥에서 잰 숫자가 그 구멍입니다. 약 150 MB 짜리 상태를 분당 1.4개 보내면, 44분 뒤에 그 폴더가 6.8 GB 였고 한 시간이면 약 12 GB 로 갑니다. 볼륨이 차서 다른 프로세스가 `ENOSPC` 로 죽었습니다. 한 시간이 지나자 TTL 이 그 묶음을 지웠습니다. 지워지지 않고 남는 누수가 아닙니다. 예산이 없는 것입니다. types.ts/config.ts 분할과 안 겹칩니다. 미리보기 배포도 계획에 없습니다. 이미 합쳐진 같은 수리도 없습니다.

고치는 자리는 `src/responses/state.ts` 의 `pruneResponses` 입니다. HEAD 에서 이 함수는 세 가지만 합니다. 만료된 항목을 지우고, 1000개를 넘으면 제일 오래된 것을 지우고, 메모리 64 MiB 를 넘으면 제일 오래된 거주 항목을 스필로 내립니다. 스필로 내린 뒤에는 디스크 합계를 보지 않습니다. 이 PR은 그 루프가 끝난 뒤에 `MAX_SPILLED_RESPONSE_BYTES` 1 GiB 를 넣습니다. 합계가 그보다 크면 생성 순서대로 스필만 지웁니다. `deleteEntry` 가 `deleteOwnedSpills` 를 타므로 파일도 같이 지워집니다. 합계는 카운터로 들고 있지 않고, 매번 `states` 를 다시 셉니다. 스필은 강등, 큰 후보 직접 입학, 스냅샷 다시 읽기처럼 들어가는 길이 여러 개입니다. 카운터를 한 곳만 빼먹으면 한도가 조용히 꺼집니다. 최대 1000개 걷기는 그렇게 안 빗나갑니다.

1 GiB 숫자의 뜻은 본문이 이미 말합니다. 같은 표본 31개의 중앙값은 1.1 MiB, p90 은 198.7 MiB 입니다. 중앙값에서는 1000개 한도와 1 GiB 가 거의 같이 붙습니다. 1000 × 1.1 MiB = 1.07 GiB. 보통 트래픽은 예전과 같이 항목 수 한도에서 잘리고, 큰 꼬리만 디스크 한도에 걸립니다. 한도를 너무 낮추면 이어가기가 `previous_response_not_found` 로 실패합니다. 그 실패는 이미 있는 길입니다. 한도를 너무 높이면 호스트 디스크와 다른 프로세스가 죽습니다. 형제 한도(개수, TTL, 파일 하나)가 모두 상수라서 설정 키는 안 넣었습니다. 크래시로 남은 고아 파일은 `states` 에 없어서 이 합계에 안 들어갑니다. 그건 기존 `recoverOrphanedResponseSpills` 의 15분 유예 청소입니다.

테스트는 `tests/responses-state.test.ts` 한 개입니다. 메모리 한도를 1024로, 디스크 한도를 20000으로 낮추고 8000바이트 항목 6개를 넣습니다. 스필 합계가 20000 이하고, 파일이 6개보다 적고, 1개 이상 남고, 마지막 아이디 `resp_spill_budget_5` 는 이어가기가 됩니다. 작성자는 루프를 꺼 두면 48564 바이트가 남아 실패한 것도 적었습니다. 로컬에서 typecheck, 포커스 117, `test:changed` 9885, 전체 16500, privacy scan 을 통과했다고 적었습니다. 지금 GitHub 체크는 hygiene / label / enforce-target / resolve-pr 만 초록입니다. 본 테스트 샤드는 아직 안 보입니다. 초안이고 체크리스트 네 칸이 비어 있습니다.

열린 PR #3018 도 `src/responses/state.ts` 의 `pruneResponses` 를 고칩니다. 그쪽은 윈도우 동기 ACL 이 이벤트 루프를 붙잡는 #3011 수리입니다. 메모리 강등 루프 안만 비동기로 바꿉니다. 이 PR은 그 루프가 닫힌 뒤에 디스크 한도를 붙입니다. 의도는 겹치지 않습니다. 다만 둘 다 같은 함수 끝이라서, #3018 이 먼저 들어가면 이 초안은 rebase 가 필요합니다. HEAD 에는 #3018 이 아직 없습니다.

라인 30 - HEAD 의 `MAX_STORED_RESPONSE_BYTES` 는 64 MiB 메모리 한도만 있다. 스필로 내린 바이트는 이 숫자 밖으로 나가고, 디스크 합계는 없다.
라인 1039 `pruneResponses` - TTL, 1000개, 메모리 강등만 한다. 강등이 만든 스필 합계를 자르는 단계가 없다. 이 PR이 붙는 자리다.
라인 1075 `sweepExpiredResponseStates` - 주기 청소는 TTL 만 지운다. 디스크 한도는 변이 경로의 prune 에만 붙는다. 트래픽이 멈춘 뒤에는 TTL 한 시간을 기다린다. 한도를 넘는 순간은 변이 prune 이 막는 설계라 맞지만, 주기 틱에 한 번 더 안 넣으면 재시작 없이 한도를 다시 계산하지 않는다.
라인 247 `deleteEntry` - 스필 삭제는 이미 여기로만 모인다. 새 루프가 여기를 부르는 선택은 맞다.
경로 src/responses/spill-store.ts `MAX_RESPONSE_SPILL_PAYLOAD_BYTES` / `recoverOrphanedResponseSpills` - 파일 하나 256 MiB 와 고아 15분 유예는 그대로다. 크래시로 맵 밖에 남은 파일은 이 1 GiB 에 안 잡힌다. 본문이 범위를 솔직히 적었다.
경로 새 루프 `spilledResponseBytes` - `[...states]` 복사본을 앞에서 뒤로 걸어 제일 오래된 스필부터 지운다. 거주 항목은 건너뛴다. 맵 삽입 순서가 생성 순서인 한 맞다. 한 파일이 한도보다 크면 빼고 나서 합계가 음수가 되어도 그 파일은 지워진다.
경로 tests/responses-state.test.ts 새 테스트 - 합계와 파일 개수, 최신 아이디 이어가기만 본다. 제일 오래된 `resp_spill_budget_0` 파일이 실제로 지워졌는지는 안 본다. 한 파일이 한도보다 큰 경우, 스냅샷 다시 읽기, 고아 파일은 없다.
경로 structure/00_overview.md / structure/02_config-and-codex-home.md - 스필 폴더와 합계 한도를 처음으로 적는다. 설정 키는 안 만든다.

메인테이너의 판단이 필요한 지점
- 1 GiB 를 그대로 둘지, 더 낮출지. 표본 p90 이 198.7 MiB 라서 1 GiB 에는 큰 파일이 네다섯 개만 들어간다
- 형제 한도처럼 상수로 둘지, `appOwnedMemoryBudgetMb` 같은 설정 키를 후속으로 열지
- #3018 이 같은 함수를 고치고 있으니, 이 초안을 먼저 넣을지 #3018 뒤에 rebase 할지
- 고아 스필을 이 합계에 넣을지, 지금처럼 15분 유예 청소에만 둘지
- 초안 체크리스트와 본 테스트 샤드가 비어 있는데, 그 전에 리뷰만 할지 머지 후보에서 뺄지

너의 추천
초안을 유지한다. 닫지 않는다. 라벨을 바꾸지 않는다. 구멍은 HEAD 에 있고 방향은 맞다. 체크리스트를 채우고 본 테스트 샤드가 초록이 된 뒤에 초안을 해제한다. #3018 이 먼저 들어가면 이 브랜치를 rebase 한다. 한 커밋에 섞지 마라. 테스트에 제일 오래된 스필 파일이 사라지는 한 줄과, 한 파일이 한도보다 큰 경우를 넣으면 더 좋다. 1 GiB 는 유지해도 된다. 낮추고 싶으면 머지 전에 상수만 바꾸면 된다. types/config 분할에 안 깔리니 rebase 로 살리지 말고, 이 패치 그대로 `dev` 에 올린다.

이 댓글은 grok-bot이 작성했습니다

lifrary:
Re: the eviction-order finding on `src/responses/state.ts` — valid, and fixed in `f8209c15`.

**Ordering.** Eviction now selects spill entries by ascending `createdAt` instead of Map iteration order. I had reached the same conclusion independently while reading `writeBoundedSnapshot`, which serializes `[...states].reverse()`; your reason is the better one to put in the comment, because delete-and-reinsert during demotion and spill replacement is directly observable, whereas I could not build a fixture that reproduces a reloaded reversed map — `flushResponseState()` + `clearResponseStateMemoryForTests()` restored nothing in my fixture (`count: 0`). The code comment now states only the part I could verify.

**Tie-breaker.** This is the part I had missed, and it matters more than it looks: `createdAt` is millisecond-resolution, so ties are ordinary under load, and a stable sort then falls back to insertion order — exactly the order the sort exists to avoid. Ties now break on the response id by direct comparison rather than `localeCompare`, so the order cannot depend on the host locale.

**Tests.** Three regressions, each confirmed to fail without the specific code it covers:

- `evicts oldest spills once the durable set exceeds the disk cap` — retains 48564 bytes against a 20000-byte cap without the eviction loop.
- `evicts by createdAt, not by insertion order` — inserts the newer entry first under a mocked clock so insertion order and `createdAt` order disagree; without the sort the newer spill is evicted and the older survives.
- `breaks createdAt ties on the response id, not on insertion order` — two entries at an identical `createdAt`, inserted in reverse id order; fails without the tie-breaker.

My first attempt at the ordering test passed with and without the sort, so it proved nothing; it was discarded rather than adjusted. The replacement does not depend on a snapshot reload.

All three exercise resident demotion: the RAM cap is set to 1024 bytes, so every entry reaches disk through `setResidentEntry` -> `pruneResponses` -> demotion -> spill, and the disk eviction then runs on that same tick.

Verification on the rebased branch (`dev` @ `7666f7d2`): `bun run typecheck` clean, `bun test tests/responses-state.test.ts` 119 pass / 0 fail, `bun run test` 16511 pass / 0 fail across 998 files, `bun run privacy:scan` passed.

lifrary:
Both review-requested test cases are in, at `18ae1c49`:

- The budget test now asserts the **oldest** spill file is the one that went (`resp_spill_budget_0.*` absent), not merely that some file went.
- New case: a single payload larger than the whole budget. With the cap below one payload the spill is written and evicted on the same tick and the running total goes negative; the case asserts the loop terminates, leaves no files, and that raising the cap restores ordinary retention, so the store is not wedged.

There are now four regressions, each confirmed to fail without the specific code it covers.

**On the readiness checklist, in the interest of not overclaiming.** Four tests fail on this machine, and all four also fail on stock `dev` with my two files reverted to `origin/dev` (115 pass / 1 fail there). None of them is in the paths this PR touches:

- `periodic reclaim frees abandoned temps without any continuation access` — this one is a real defect in the test, not the environment. It computes `const deadPid = process.pid === 4242 ? 4243 : 4242;` and assumes that pid is dead. On this macOS host pid 4242 is `liveactivitiesd`, so the sweep correctly refuses to reap a live owner's temp and the test reads that as a miss. The same assumption appears at four sites in `tests/responses-state.test.ts` (lines 1682, 1744, 1809, 1907), while line 1719 already does it correctly by scanning for a pid that is actually free. Happy to send that as its own PR rather than widen this one — say the word.
- `doctor reclaim wiring` x2 and `status reports stale process records` x2 — process- and port-driven end-to-end tests, failing only under load (they took 496 ms to 2248 ms with the machine at load average 16, and pass when it is quiet).

For the record on timing: this suite is load-sensitive here. Two earlier full runs on this branch were green (16500 and 16511 pass, 0 fail), and two runs hit the runner's own 900s parallel guard on `codex-inject-write-lock.test.ts` while other work was running on the machine. That test passes alone in 4.5 s and does not touch anything this PR changes.

**On the judgment points raised.**

- *1 GiB or lower*: happy either way, and it is a one-constant change before merge. The rationale and the sample it came from are in the description; n is 31 and it is one host, so treat it as a starting point rather than a measurement of your users.
- *Constant or config key*: left as a constant to match the three sibling bounds. A `...Mb` key on the `appOwnedMemoryBudgetMb` pattern would touch `config.ts`, `server/index.ts`, `management/config-routes.ts` and the GUI, which felt like a separate change rather than part of a bug fix.
- *Order against #3018*: no textual conflict as it stands — #3018 edits inside the RAM demotion loop, this appends after that loop closes. If #3018 lands first I will rebase this branch rather than fold anything into it.
- *Orphan spills*: left to `recoverOrphanedResponseSpills`. Pulling them into this total would mean a directory scan inside `pruneResponses`, which sits on the write path.
- *Periodic tick*: worth noting I hit this in production. After restarting the proxy onto this change the spill directory sat at 1.8 GB, above the 1 GiB cap, until the first request arrived and a mutation-path prune ran. The design is as you describe and the mutation path does bound the growing case; the gap is only a process that starts over budget and then goes idle. `sweepLiveness` is already wired for this store and its comment says disk reclaim rides the liveness tick, so that looks like the natural seam. I have not made that change — it widens the PR, and it is your call whether it belongs here or in a follow-up.

lifrary:
The deferred-generation finding is valid and is fixed in `c80baf04`. I verified it against the code before acting: `pendingSpillUnlinks` holds `ResponseSpillRef`s whose files stay on disk until a snapshot flush drains them, and at `PENDING_SPILL_UNLINKS_MAX` 128 with the 256 MiB per-file ceiling that is up to 32 GiB the budget could not see.

Two changes:

- `spilledResponseBytes()` now adds `pendingSpillUnlinks` to the total, so the cap measures every file the store owns rather than only the live entries.
- Over budget, deferred generations are released **before** any live entry. They are already superseded, so the only cost is the crash window the queue exists to cover — and that is the same trade the queue's own overflow path (`while (pendingSpillUnlinks.length > PENDING_SPILL_UNLINKS_MAX) deleteResponseSpill(...)`) already makes, whose comment states it explicitly as bounded loss against unbounded disk. Evicting a live continuation to make room for a dead file would be the wrong order.

I did not add the forced snapshot persistence from your suggestion. `pruneResponses` is synchronous and on the write path, and the existing overflow drain releases queued generations without forcing a flush, so following that precedent keeps the change inside a pattern the file already sanctions. Happy to revisit if you would rather the queue never be drained early from here.

New regression: `counts deferred spill generations against the disk cap and drains them first` replaces one spilled id eight times before the queue can drain. Without the pending accounting the directory holds **eight** files against a cap sized for four, and the live continuation still replays afterwards, so the drain is not just deleting everything.

That makes five regressions on this PR, each confirmed to fail without the specific code it covers.

Re-verified on the amended commit: `bun run typecheck` clean, `bun test tests/responses-state.test.ts` 121 pass / 0 fail, `bun run test:changed` 9896 tests across 542 files, `bun run privacy:scan` passed, `bun run test` **16513 pass / 0 fail across 998 files**.

One note on the local-suite numbers I posted earlier: I attributed four of the failures to machine load. That was wrong. They were all the same hardcoded-pid fixture defect, now sent separately as #3042, and with that fixed the full suite is green here for the first time. The load did make those tests slow; it did not make them fail, and I conflated the two.

Ingwannu (changes requested):
Requesting changes on exact head c80baf045540decba3cda2b14dedeecf3682a60b. The aggregate budget, createdAt ordering, deterministic tie-breaker, deferred-generation accounting, and oversized-single-file termination are all valuable. One product boundary is still missing: the cap is enforced only at the end of mutation-path pruneResponses. A process can load or restart over budget and then remain idle indefinitely above the advertised aggregate ceiling; the author observed 1.8 GiB staying above the 1 GiB cap until the first request arrived. Extract the spill-budget eviction into one helper and invoke it on state load or the existing periodic liveness/TTL sweep as well as mutation pruning. Add a regression that restores an over-budget snapshot and proves the periodic/startup path reclaims oldest spills without requiring a new continuation mutation, while preserving the newest replayable entry. Keep orphan crash files on their existing grace sweep if desired, but document that separate temporary allowance precisely. Exact-head Cross-platform CI should run only after this boundary is covered.

lifrary:
The product boundary is covered in `b4d1d240`, rebased onto current `dev` (`7471b267`), zero commits behind.

**One owner, three callers.** The eviction is now `enforceSpilledResponseBudget()`, invoked from mutation pruning, from the lazy load that follows a restart (`ensureLoaded` ends in `pruneResponses`), and from `sweepExpiredResponseStates`. The periodic caller is the one that was missing, and it is the one that matches what I saw in production: 1.8 GiB held above a 1 GiB cap after a restart until the first request arrived. `sweepExpiredResponseStates` still returns its TTL count, so its existing contract is unchanged; the budget only adds a `schedulePersist` trigger when it actually released something.

**Regression, with no continuation mutation.** `reclaims an over-budget snapshot without a new continuation mutation` writes four spills at distinct `createdAt`, flushes, clears memory, then comes back under a ceiling the snapshot was not written for. A **read** (`expandPreviousResponseInput`) drives the load path — no write — and the oldest file is gone while the newest still exists and still replays. Then the ceiling is lowered again and `sweepExpiredResponseStates()` alone does the reclaiming. Removing just the periodic caller fails it: 16454 bytes retained against a 10000 cap.

That makes six regressions on this PR, each confirmed to fail without the specific code it covers.

**The orphan allowance is now stated rather than implied.** `structure/02` says the ceiling bounds what the store can account for — every entry in the map plus the superseded generations queued for unlink — and deliberately not the directory. Crash-orphaned files are absent from the map, so the accounting can neither see nor price them; they stay with `recoverOrphanedResponseSpills`, and a host that crashes repeatedly can hold spill bytes above the ceiling for up to `RESPONSE_SPILL_ORPHAN_GRACE_MS` past each crash. The same note is on the helper.

**On the local suite, precisely.** `bun run typecheck` clean, `bun run privacy:scan` passed, `bun test tests/responses-state.test.ts` 136 pass / 0 fail, `bun run test:changed` 9977 tests across 543 files.

The full suite has not produced a clean run on this machine, and I would rather say so than round it up. Two attempts on this head: one hit the runner's own 900 s parallel guard on `tests/codex-inject-write-lock.test.ts`, and one completed at 16686 pass / 1 fail where the single failure was `update stops the running proxy before replacing files > npm launcher restarts the stopped runtime after a staged update failure`, which took 91.5 s. Both files pass in isolation (71/0 and 15/0) and neither references anything this PR touches. The machine has been sitting at load average 10-15 throughout, which is what the runner's own diagnostic points at. So: no failure I can attribute to this change, and also no green full run from me — the exact-head Cross-platform run is the arbiter, and the boundary you asked for is now in place for it.

lidge-jun:
Landed via #3097 at 5f0b39048

lidge-jun:
Superseded by maintainer carry #3097 (merged at 5f0b39048).

PR #3039, opened 2026-08-31
fix(service): do not fail a Windows cold start that is still coming up

### LINKED ISSUE #3009: [Bug][Windows]: service repair can fail at 20s just before the proxy becomes healthy
## Client or integration

OpenCodex dashboard

## Area

Service lifecycle

## Summary

On Windows, `ocx service repair` can report failure at its fixed 20-second health deadline even though the scheduled service is still starting and becomes healthy immediately afterward.

I saw this while recovering a dashboard update on 2.36.0. The repair command printed that no proxy answered on port 11000 within 20 seconds and returned exit 1. The Task Scheduler job remained running. Shortly afterward, `/healthz` returned `status: ok` from the same scheduled startup, and the proxy remained stable.

Expected: a slow but valid Windows cold start should not be reported as a terminal repair failure. A final probe at the deadline, or a slightly longer/phase-aware bounded wait, would avoid the false negative while still failing a service that never binds.

## Reproduction

1. Install OpenCodex 2.36.0 through npm on Windows.
2. Install the `opencodex-proxy` Task Scheduler service.
3. Use an environment where startup work can take around 20 seconds. In this run the service log also showed an NTFS ACL hardening timeout and previous-session journal recovery before the proxy announced its listener.
4. Run `ocx service repair`.
5. Observe the command exit 1 after the 20-second serving check.
6. Check the scheduled task and `/healthz` a few seconds later. The same service startup is running and healthy.

Observed command result:

```text
Service repaired, but no proxy answered on port 11000 within 20s.
```

Follow-up state:

```json
{
  "status": "ok",
  "service": "opencodex",
  "version": "2.36.0",
  "pid": 39388,
  "port": 11000
}
```

The process was created at `2026-08-30T21:33:19+08:00`; the scheduled task run was recorded at `21:33:33+08:00`. After it became healthy, repeated checks and the Web UI stayed available. This was not a crash/restart loop.

The current code uses one fixed deadline:

```ts
export const SERVICE_INSTALL_HEALTH_MS = 20_000;
```

and returns failure as soon as that deadline is reached. A final identity probe at the boundary would at least close the narrow race. If 20 seconds is intentionally the normal budget, a separate bounded grace period for a still-running Windows service may be another option.

Related but not duplicate: #720 fixed the GUI update worker's 15-second restart-confirmation boundary. This report is the separate `confirmServiceServing()` path used by service install/start/repair, which still has a 20-second deadline on current `main`.

## Version

2.36.0; current `main` still uses the 20-second service-install health deadline.

## Operating system

Windows 10 22H2, build 19045, x64

## Provider and model

Not provider-specific

## Logs or error output

```shell
[opencodex] ACL hardening timed out (ETIMEDOUT) — continuing without NTFS ACL harden
Previous session did not shut down cleanly. Codex state restored from journal.
opencodex proxy running on http://localhost:11000
```

Current source: https://github.com/lidge-jun/opencodex/blob/main/src/service.ts#L554-L617

## Redacted configuration

```json
{
  "installer": "npm",
  "service": "Windows Task Scheduler",
  "port": 11000
}
```

## Checks

ntdatt812:
Fixes #3009.

`ocx service repair` reported a terminal failure at its 20s health deadline for a service that bound a few seconds later and then stayed healthy. This follows the review on the issue, which asked for a longer Windows budget, a final probe at the deadline, and a message that prints the real wait.

## What changed

**Windows gets a 45s budget; nothing else changes.** The cold start does NTFS ACL hardening and previous-session journal recovery before the listener is announced, so the 20s that is plenty elsewhere is not always enough there. `serviceInstallHealthMs(platform)` keeps the other platforms on `SERVICE_INSTALL_HEALTH_MS`, so a healthy Linux or macOS install cannot be slowed down by this.

**One more knock after the deadline.** The loop's last probe *starts* before the deadline, so a service that binds while that probe is in flight was reported as dead. It now waits a short grace and knocks once more before calling it a failure.

A caller that passed a zero budget still gets exactly the single probe it asked for — `waited` is only set once a sleep has actually happened, so "do not wait" keeps meaning that.

**The message prints the time actually waited.** It printed `SERVICE_INSTALL_HEALTH_MS` whatever `timeoutMs` the caller passed, so a caller with its own budget was told it had waited 20 seconds regardless.

`process.exitCode = 1` on a genuine failure is unchanged. As the comment there says, the GUI update worker reads the child's exit status, and a registered-but-silent service must not look like a successful update.

## Tests

`bun test tests/service.test.ts` → **137 pass, 3 skip, 0 fail**. `bun x tsc --noEmit` clean.

Added, matching the two cases the review asked to pin:

- `accepts a service that binds during the grace after the deadline` — the probe answers only once the clock is past the deadline, which is the shape the report describes
- `still fails a service that never binds` — the budget is still a bound, not a suggestion
- `gives Windows a longer cold-start budget than the other platforms` — asserts the relationship and that the others are untouched, rather than hard-coding 45s

**Mutation-checked.** Removing the grace probe:

```
(fail) confirmServiceServing > accepts a service that binds during the grace after the deadline
136 pass, 1 fail
```

## One existing test changed

`probes at least once even with a zero budget` asserted `toBe(1)`. The grace probe adds one when the caller did give the wait some time, so the exact count is no longer the contract — it is now `toBeGreaterThanOrEqual(1)`, which is what the test's own name asks and still catches a version that returns without knocking at all. Flagging it explicitly since changing an existing assertion deserves a look.

## Not in this PR

#3008 came out of the same recovery session but is a different file and a different failure point, and the review asked to keep them apart. Same for the ACL hardening itself (#3011), which is assigned.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved service installation health checks on Windows by allowing additional startup time.
  * Added a brief grace period and final check before reporting that a service failed to start.
  * Service status reports now show the actual time spent waiting.
* **Tests**
  * Added coverage for delayed service startup, persistent failures, and platform-specific health-check timing.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -661,6 +661,23 @@ export function installedServiceListenPort(): number {
 export const SERVICE_INSTALL_HEALTH_MS = 20_000;
 
 /**
+ * Windows gets a longer budget because its cold start does more before the
+ * listener exists: NTFS ACL hardening and previous-session journal recovery
+ * both run first, and #3009 recorded a service that bound a few seconds past
+ * the 20s deadline and then stayed healthy. Reporting that as a terminal
+ * repair failure is worse than waiting: the caller's fallback is to start a
+ * second proxy against a port that is about to be taken.
+ */
+export const SERVICE_INSTALL_HEALTH_WINDOWS_MS = 45_000;
+
+/** The health budget for the platform this is running on. */
+export function serviceInstallHealthMs(
+  platform: NodeJS.Platform = process.platform,
+): number {
+  return platform === "win32" ? SERVICE_INSTALL_HEALTH_WINDOWS_MS : SERVICE_INSTALL_HEALTH_MS;
+}
+
+/**
  * Whether a proxy actually answers on the port this install/start just produced.
  *
  * Registration is not service: `launchctl list` reports a job that never bound, and
@@ -690,12 +707,23 @@ export async function confirmServiceServing(
   const now = deps.now ?? Date.now;
   const sleep = deps.sleep ?? ((ms: number) => new Promise<void>(r => setTimeout(r, ms)));
   const probe = deps.probe ?? (async (p, h) => !!(await proxyIdentityAt(p, { hostname: h })));
-  const deadline = now() + (deps.timeoutMs ?? SERVICE_INSTALL_HEALTH_MS);
+  const deadline = now() + (deps.timeoutMs ?? serviceInstallHealthMs());
+  let waited = false;
   for (;;) {
     if (await probe(port, hostname)) return { ok: true, port };
-    if (now() >= deadline) return { ok: false, port };
+    if (now() >= deadline) break;
     await sleep(500);
+    waited = true;
   }
+  // The probe that ran last started before the deadline, so a service that
+  // binds during it is reported as dead (#3009). Knock once more after a
+  // short grace before calling it a failure. A zero budget means the caller
+  // asked not to wait, so it gets the single probe it asked for.
+  if (waited) {
+    await sleep(500);
+    if (await probe(port, hostname)) return { ok: true, port };
+  }
+  return { ok: false, port };
 }
 
 /**
@@ -711,14 +739,18 @@ async function reportServiceServing(
   verb: "installed" | "started" | "repaired",
   deps: Parameters<typeof confirmServiceServing>[0] = {},
 ): Promise<void> {
+  const elapsed = deps.now ?? Date.now;
+  const startedAt = elapsed();
   const serving = await confirmServiceServing(deps);
   if (serving.ok) {
     console.log(`✅ opencodex service ${verb} and serving on port ${serving.port}.`);
     return;
   }
   console.error(
+    // The elapsed time, not the constant: a caller that passes its own
+    // timeoutMs used to be told it had waited 20s whatever it waited.
     `⚠️  Service ${verb}, but no proxy answered on port ${serving.port} within `
-    + `${Math.trunc(SERVICE_INSTALL_HEALTH_MS / 1000)}s.\n`
+    + `${Math.max(1, Math.round((elapsed() - startedAt) / 1000))}s.\n`
     + `   The manager registered the job; that is not the same as serving.\n`
     + `   Log:       ${serviceLogPath()}\n`
     + `   Meanwhile: ocx start   (serves in the foreground)`,
--- a/tests/service.test.ts
+++ b/tests/service.test.ts
@@ -7,7 +7,7 @@ import { pathToFileURL } from "node:url";
 import * as serviceModule from "../src/service";
 import { saveConfig } from "../src/config";
 import { windowsEnvIndirectBatchValue } from "../src/lib/win-paths";
-import { assertServiceAuthEnvironment, assertServiceEnvironmentMatchesInstall, bakedServicePathsDiagnostic, confirmServiceServing, launchdListenPort, systemdListenPort, buildPlist, buildUnit, buildWindowsLauncherVbs, buildWindowsSchtasksCreateArgs, buildWindowsSchtasksCreateArgsForXml, buildWindowsServiceScript, buildWindowsTaskXml, deriveWindowsServiceDiagnostic, deriveWindowsServiceDiagnosticForCurrentUser, installFreshWindowsSchedulerSafely, installServiceSafely, launchctlLoadFailed, launchdJobMatchesPlist, normalizeServiceSubcommand, parseServiceArgs, parseServiceInstallState, planServiceCommand, prepareServiceInstall, probeServiceInstallation, readWindowsSchedulerXmlState, registerFreshWindowsSchedulerTask, removeNativeWindowsServiceForScheduler, repairService, resolveServiceListenPort, runLaunchctl, selectServiceSubcommand, serviceLogPath, serviceStartableFromTray, serviceStatusReport, serviceRetryCommand, serviceStatusSummary, stableLauncherEntry, systemdNeedsDaemonReload, systemdServiceInstallCleanupOps, uninstallSystemd, windowsListenPort, winswListenPort, startLaunchd, windowsTaskRegistrationHealthy } from "../src/service";
+import { assertServiceAuthEnvironment, assertServiceEnvironmentMatchesInstall, bakedServicePathsDiagnostic, buildPlist, buildUnit, buildWindowsLauncherVbs, buildWindowsSchtasksCreateArgs, buildWindowsSchtasksCreateArgsForXml, buildWindowsServiceScript, buildWindowsTaskXml, confirmServiceServing, deriveWindowsServiceDiagnostic, deriveWindowsServiceDiagnosticForCurrentUser, installFreshWindowsSchedulerSafely, installServiceSafely, launchctlLoadFailed, launchdJobMatchesPlist, launchdListenPort, normalizeServiceSubcommand, parseServiceArgs, parseServiceInstallState, planServiceCommand, prepareServiceInstall, probeServiceInstallation, readWindowsSchedulerXmlState, registerFreshWindowsSchedulerTask, removeNativeWindowsServiceForScheduler, repairService, resolveServiceListenPort, runLaunchctl, selectServiceSubcommand, SERVICE_INSTALL_HEALTH_MS, serviceInstallHealthMs, serviceLogPath, serviceRetryCommand, serviceStartableFromTray, serviceStatusReport, serviceStatusSummary, stableLauncherEntry, startLaunchd, systemdListenPort, systemdNeedsDaemonReload, systemdServiceInstallCleanupOps, uninstallSystemd, windowsListenPort, windowsTaskRegistrationHealthy, winswListenPort } from "../src/service";
 import type { ServiceDiagnostic } from "../src/service";
 import { definitionCarriesCredential, resolvedProxyEnv, writeServiceDefinitionFile } from "../src/service";
 import { buildWinswXml } from "../src/lib/winsw";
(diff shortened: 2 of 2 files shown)


lidge-jun:
## 리뷰 · 우선순위 68 / 80

이 PR은 Windows에서 `ocx service repair/install/start`가 이미 뜨고 있는 프록시를 죽은 것처럼 보고하는 문제를 고친다. 이슈는 `#3009`다. 지금 `dev` HEAD `4180067b4`의 `src/service.ts` `confirmServiceServing`은 기본 예산 `SERVICE_INSTALL_HEALTH_MS = 20_000`으로 500ms 간격 프로브만 돌리고, 데드라인에 닿으면 바로 `{ ok: false }`를 반환한다. Windows 콜드 스타트는 NTFS ACL 강화와 이전 세션 저널 복구가 리스너보다 먼저 돌아가서, 보고대로 20초를 조금 넘겨 바인딩한 뒤에도 정상 서비스인 경우가 있다. 그때 repair가 실패(exit 1)로 끝나면 호출 쪽 폴백이 곧 잡힐 포트에 두 번째 프록시를 또 띄우려 한다. 등록은 됐는데 serving은 아닌 상태를 “완전 실패”로 취급하는 비용이 크다.

고침은 세 갈래다. 첫째, `SERVICE_INSTALL_HEALTH_WINDOWS_MS = 45_000`과 `serviceInstallHealthMs(platform)`을 추가해 win32만 긴 예산을 쓰고 linux/darwin은 기존 20초를 유지한다. 기본 데드라인만 `serviceInstallHealthMs()`로 바꾸고, 호출자가 `timeoutMs`를 넘기면 예전처럼 그 값이 이긴다. 둘째, 루프가 데드라인에서 바로 실패하지 않고 `break`한 뒤, 한 번이라도 `sleep`을 돌렸으면( `waited` ) 짧은 grace 후 프로브를 한 번 더 한다. 마지막 프로브가 데드라인 전에 시작돼 그 사이에 바인딩된 경우를 #3009 형태로 받는다. 예산 0은 `waited === false`라 grace를 건너뛰어 “기다리지 마” 계약을 지킨다. 셋째, `reportServiceServing` 경고 문구가 상수 20초가 아니라 실제로 기다린 초를 찍는다. 호출자가 다른 `timeoutMs`를 준 경우에도 메시지가 거짓말하지 않는다. `process.exitCode = 1`은 그대로다. GUI 업데이트 워커가 자식 exit를 보기 때문이다.

테스트는 `tests/service.test.ts`에 grace 성공, 끝까지 실패, Windows>다른 플랫폼 예산 관계 세 개를 추가했다. 기존 “zero budget은 최소 한 번 프로브” 단언은 `toBe(1)`에서 `toBeGreaterThanOrEqual(1)`로 풀렸다. zero budget 경로 자체는 grace를 안 타므로 여전히 1에 가깝지만, 이름 기준 계약은 “최소 한 번”이 맞다. 본문은 grace를 빼면 해당 테스트만 깨진다고 mutation 확인까지 적었다. 로컬 `bun test tests/service.test.ts` 137 pass / tsc clean. types.ts/config.ts 분할과 무관하고 미리보기 배포도 없다. #3008·#3011은 의도적으로 이 PR 밖이다.

다만 리뷰 준비 체크리스트 4칸이 아직 비어 있고 `review-ready` 라벨도 없다. 게이트 봇 기준으로는 아직 draft 계약이다. CI 쪽 enforce-target/hygiene/label/resolve-pr은 통과로 보인다.

`src/service.ts` SERVICE_INSTALL_HEALTH_WINDOWS_MS 45000 - Windows만 늘린다. 보고(#3009) 대비 여유는 있으나 매직 넘버라 근거 코멘트가 이미 붙어 있는 점은 좋다.
`src/service.ts` serviceInstallHealthMs - 기본 예산만 플랫폼 분기. 명시 timeoutMs 호출(예: 짧은 tray/status 프로브)은 안 건드린다.
`src/service.ts` confirmServiceServing grace - waited 가드가 zero-budget 계약을 지킨다. 방향 맞다.
`src/service.ts` reportServiceServing 경과 초 - 상수 대신 실측. 호출자 커스텀 예산과 메시지가 일치한다.
경로 체크리스트 / review-ready - 4칸 미체크. 본문 검증은 적혀 있어도 게이트상 아직 ready가 아니다.
`tests/service.test.ts` toBeGreaterThanOrEqual(1) - zero-budget 정확 횟수 단언을 느슨하게 바꿈. 의도는 이해되나, zero 경로만 따로 toBe(1)로 남기면 회귀가 더 빡빡하다.

메인테이너의 판단이 필요한 지점
- 체크리스트를 기여자가 채울 때까지 머지를 미룰지, 로컬 검증 서술만으로 진행할지
- 45초가 #3009 실측 대비 충분한지, 더 늘리거나 설정으로 뺄지
- zero-budget 테스트를 다시 toBe(1)로 쪼갤지
- #3009를 이 PR 머지와 함께 닫을지 (Fixes #3009 가 본문에 있음)

너의 추천

방향은 맞고 #3009를 닫는 올바른 크기의 수정이다. 체크리스트를 채우고 `review-ready`가 붙으면 머지하세요. 여유 있으면 zero-budget 케이스는 `toBe(1)`로 다시 고정하는 편이 낫다. 라벨은 바꾸지 않습니다. types.ts/config.ts 분할로 닫을 대상이 아니고 미리보기 배포도 없습니다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
Requesting changes on exact head `b9c837c9c4d0f31c716ef2c5e37a522202c839a5`.

The platform-specific budget and one final post-deadline probe are reasonable for #3009, but the current regression contract is weakened in two places already identified by the unresolved inline review:

1. Restore `toBe(1)` for the zero-budget case. With `toBeGreaterThanOrEqual(1)`, a future implementation can incorrectly sleep and perform grace probes even when the caller explicitly requested no wait, and the test still passes. The positive-timeout test is the correct place to require multiple probes.
2. Pin the selected Windows budget exactly (`SERVICE_INSTALL_HEALTH_WINDOWS_MS` / `45_000`), while continuing to assert Linux and macOS remain on `SERVICE_INSTALL_HEALTH_MS`. A merely-relative assertion does not protect the product decision this PR introduces.

This head is behind current `dev` and has only hygiene/target checks. Please resolve those focused test gaps, rebase, and run exact-head Windows/service CI before another review.

lidge-jun:
Superseded by #3104, which keeps your production logic exactly as written — the Windows budget, the `waited` guard, and the grace probe.

Two test changes. The zero-budget assertion is restored to `expect(probes).toBe(1)`: `toBeGreaterThanOrEqual(1)` passes against a version that sleeps when the caller asked not to wait, which is the one thing that test exists to forbid. Your `waited` guard already preserves the contract, so nothing needed relaxing. And the Windows budget is pinned to `45_000` absolutely rather than `> linux`, since the reported service bound past 20s and a relational assertion accepts 21s.

#3104 also carries the #3064 fix, because both are in `src/service.ts`. Mutation evidence is in its description.

ntdatt812:
Both of your points about the tests are right, and they are the same mistake in two shapes. Closing in favour of #3104.

**`expect(probes).toBe(1)`.** I relaxed it to `toBeGreaterThanOrEqual(1)`. That assertion exists to forbid sleeping when the caller passed no budget, and "at least one" is satisfied by exactly the version it forbids — so the change did not weaken the test, it deleted it while leaving it green. The right move was to notice the `waited` guard already held the contract and leave the assertion alone, which is what you did.

**The Windows budget.** I asserted `serviceInstallHealthMs("win32") > serviceInstallHealthMs("linux")`, which passes at 21s. #3009 reports a listener that bound *past* 20s, so an inequality does not encode the report — the number does.

The common fault is that in both places I wrote the assertion that my change happened to satisfy rather than the one the requirement states. That is the same habit that produced the hole you found in #3067: a fixture chosen to agree with the code instead of to probe it.

The production logic you kept is unchanged, so there is nothing left here that #3104 does not carry. Thank you for the reimplementation and for saying which parts held up — that is more useful than a merge would have been.

lidge-jun:
**This PR is deliberately staying open.** #3134 landed on `dev` as `b14b741dc` and closes #3009, but it does **not** carry everything this PR contributed, so closing it as superseded would drop that work silently.

**What landed.** The Windows cold-start budget (45s, pinned absolutely rather than as an inequality) and the final probe at the deadline. Your diagnosis drove both.

**What did not.** The diagnostic message. This PR prints the time actually waited:

```ts
// src/service.ts:742-753 (this PR)
const startedAt = elapsed();
...
+ `${Math.max(1, Math.round((elapsed() - startedAt) / 1000))}s.\n`
```

with the comment stating exactly why: *"The elapsed time, not the constant: a caller that passes its own timeoutMs used to be told it had waited 20s whatever it waited."*

What landed prints the configured budget instead:

```ts
// src/service.ts:742-750 (on dev now)
const healthBudgetMs = deps.timeoutMs ?? serviceInstallHealthMs();
...
+ `${Math.trunc(healthBudgetMs / 1000)}s.\n`
```

**Why that matters more now than before.** #3134 also adds a post-deadline grace knock, so the real wait can exceed the budget. The printed number can therefore understate the time actually spent — which is the defect this PR set out to fix, reintroduced by the change that supersedes it.

There is live evidence of the shape, from the `windows-schtasks` CI job on the merged branch:

```
⚠️  Service installed, but no proxy answered on port 10199 within 45s.
```

That `45s` is the constant, not a measurement. An operator reading it cannot tell whether the probe waited 45s or 45s plus the grace.

**One thing that was intentionally not carried, so it does not come back by accident:** this PR relaxes `expect(probes).toBe(1)` to `toBeGreaterThanOrEqual(1)`. That assertion is what stops a future change from sleeping when the caller explicitly asked not to wait, and "at least one" passes against exactly the version it exists to forbid. The `waited` guard already preserves the contract, so the strict form was kept.

The elapsed-time diagnostic is worth landing on its own against current `dev`. It is a small, self-contained change now that the budget work is in, and it should keep your authorship.

ntdatt812:
The diagnostic half that #3134 deliberately left open is now #3138 — the failure line reports the wait actually spent, grace knock included, instead of the configured budget.

That closes out everything this PR was carrying: the production logic went in through #3134, and the two test criticisms above were fair and are not being re-litigated.

PR #3044, opened 2026-08-31
fix(windows): drain response spill publications before shutdown snapshot

### LINKED ISSUE #3011: [Bug][Windows] synchronous ACL hardening still stalls /healthz and poisons response spill writes after #612

lidge-jun:
## Summary

Lands #3011 — Windows synchronous ACL hardening stalls `/healthz` up to 47s — by carrying @Ingwannu's fix and closing the shutdown boundary it opened.

**@Ingwannu's commit is the base of this branch, unmodified** (`fix(windows): move response spill ACL work off event loop`). Moving spill ACL work off the event loop is the right fix for the stall. This adds the drain it needs.

**The gap.** `responseSpillPublicationTail` was awaited only by a test helper, never by `flushResponseState()` — which is what shutdown actually calls. Residents over 2 MiB are excluded from snapshots, so a payload over that cap had a window where the snapshot skipped the resident *and* the spill stub was not installed yet, and `process.exit()` skipped the writer's temp cleanup. Result: a lost continuation plus a possible orphaned temp. On `dev` today that window does not exist, because oversized candidates publish synchronously before the request returns — so abandoning the write would have matched the PR head, not `dev`.

**What this adds.** A drain to a stable fixed point, called before snapshot serialization. Draining after would keep the bug, since the oversized resident is skipped at serialization time. A bare `Promise.race` is not sufficient: a writer that publishes after serialization is the same lost continuation wearing a timeout.

The drain is bounded by a **budget split**, not just a total: end-to-end `B = 5000ms`, drain cap `B - R`, reserved fallback slice `R = 4000ms`. `R` is reserved up front rather than taken from the remainder, because a drain that consumed the whole deadline would otherwise leave the fallback zero time. The fallback passes its budget down instead of letting each `harden` call open a fresh 30s window — the sync writer hardens directory and temp as two separate calls, each of which resolves its own deadline.

An abandoned writer is marked superseded, and supersession reaches the writer itself rather than only the state tracking, so a late completion cannot publish over the fallback's result. Cleanup is attempted for every superseded job and never short-circuits snapshot persistence.

## Fail-closed consequence, stated plainly

When the fallback budget is exhausted, the candidate becomes a `spill-failed` tombstone and **the original payload is unrecoverable.** This is explicit, not silent: shutdown reports failure and exits nonzero, the tombstone persists, replay returns `previous_response_not_found` with internal reason `spill_failed`, and the client is told to resend the full conversation.

Terminalization is bounded at `MAX_STORED_RESPONSES + 1` = 1001 passes, which covers 1000 legitimate residents plus one stale initial batch. If that structural guard ever fires it records `ELOOP` and fail-closes **all** remaining resident continuation state, not only the originally pending spills.

## Verification

Five review rounds by an independent reviewer, each returning FAIL until the last. Every finding was a real defect, and each fix is a separate commit so the sequence stays legible:

1. The abandoned writer could still publish to the filesystem and orphan a temp — supersession reached state, not the writer.
2. Making cleanup failure reject the drain discarded **every other unsnapshotted response**, because the rejection preempted `persistNow()` while shutdown still exited 0. Persistence now always completes before any collected failure surfaces.
3. Budget exhaustion caused an **infinite synchronous requeue loop** — pruning re-queued the over-cap resident and the drain never terminated. Exhausted candidates are now terminal before pruning.
4. The regression guarding that loop could **wedge CI** rather than fail, since an in-test timeout cannot interrupt a blocked JS thread. That scenario now runs in a child process with a watchdog that SIGKILLs and reports, plus a production iteration ceiling that is tested rather than asserted in a comment.

Watchdog bounds: 3s child / 5s test locally, 30s / 32s on CI POSIX, 45s / 47s on CI Windows.

Red-first evidence recorded for each regression, including the hang case (exit 124 under an external bound before the fix).

- `bun test tests/responses-state.test.ts` — 128 pass / 0 fail
- `bun test tests/windows-secret-acl.test.ts` — 169 pass / 0 fail
- `bun run typecheck` — clean
- `bun run test` — full suite on Linux x86_64 / bun 1.3.14 at this exact head

Tests for ordinary ACL failure and copy-fallback cleanup already passed at the PR head; they are kept as coverage and are **not** claimed as red-first.

## Residual risk

A real Windows host is still needed to prove NTFS unlink semantics and `icacls` timeout behaviour while a path is held. Everything here was exercised through the repository's injected Windows/ACL runners.

## Checklist

- [x] Targets `dev`
- [x] Regression tests for each behaviour change, each driven red first
- [x] `bun run typecheck` clean
- [x] `bun run privacy:scan` clean
- [x] No request bodies, tokens, or account identifiers logged
- [x] No `gui` change, so no screenshot applies
- [x] `structure/02_config-and-codex-home.md` records the shutdown ordering and why the 2 MiB exclusion makes it load-bearing

Closes #3011

## Summary by CodeRabbit

- **New Features**
  - Improved Windows response persistence with non-blocking writes, bounded retries, cancellation, and cleanup.
  - Graceful shutdown now drains pending response data before saving state.
  - Added safeguards for incomplete shutdowns and restart handoffs.
  - Added bounded handling for Windows security-check subprocesses.

- **Bug Fixes**
  - Shutdown and restart operations now report an error when draining fails.
  - Windows configuration security checks no longer block the event loop.

- **Documentation**
  - Documented Windows response persistence and shutdown behavior.

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -363,11 +363,12 @@ async function handleStart(options: { block?: boolean } = {}) {
     shutdownStartedAt = now;
     console.log("\n🛑 Shutting down opencodex proxy...");
     void (async () => {
+      let shutdownSucceeded = false;
       try {
-        await drainAndShutdown(server, config.shutdownTimeoutMs ?? 5000);
+        shutdownSucceeded = await drainAndShutdown(server, config.shutdownTimeoutMs ?? 5000);
       } finally {
         const restored = syncCleanup(); // idempotent (cleaned-guard); also re-run by process.on("exit")
-        process.exit(restored ? 0 : 1);
+        process.exit(restored && shutdownSucceeded ? 0 : 1);
       }
     })();
   };
--- a/src/config/paths.ts
+++ b/src/config/paths.ts
@@ -1,7 +1,7 @@
 import { chmodSync, existsSync } from "node:fs";
 import { homedir } from "node:os";
 import { join, resolve } from "node:path";
-import { hardenSecretDir } from "../lib/windows-secret-acl";
+import { hardenSecretDirAsync, windowsSecretAclApplies } from "../lib/windows-secret-acl";
 import { assertNotRealHomeUnderTest } from "../lib/test-home-guard";
 
 /**
@@ -14,6 +14,7 @@ export function expandUserPath(raw: string): string {
   return raw;
 }
 let resolvedConfigDirCache: { raw: string | undefined; path: string } | null = null;
+const configDirHardeningFlights = new Map<string, Promise<void>>();
 
 export function getConfigDir(): string {
   const raw = process.env["OPENCODEX_HOME"]?.trim() || undefined;
@@ -34,7 +35,21 @@ export function hardenConfigDir(): void {
   assertNotRealHomeUnderTest(dir);
   if (!existsSync(dir)) return;
   try { chmodSync(dir, 0o700); } catch { /* best-effort */ }
-  if (process.platform === "win32") {
-    hardenSecretDir(dir, { required: false });
+  if (windowsSecretAclApplies() && !configDirHardeningFlights.has(dir)) {
+    // This is an optional read-path harden. Waiting synchronously here used to stop the Bun
+    // event loop (including /healthz) for the full icacls timeout. Required mutation paths keep
+    // their own awaited/fail-closed hardening; ordinary config reads only start one soft flight.
+    const flight = hardenSecretDirAsync(dir, { required: false })
+      .then(() => undefined)
+      .catch(() => undefined)
+      .finally(() => {
+        if (configDirHardeningFlights.get(dir) === flight) configDirHardeningFlights.delete(dir);
+      });
+    configDirHardeningFlights.set(dir, flight);
   }
+}
+
+/** Test-only: settle optional config-directory hardening without exposing it to production callers. */
+export async function flushConfigDirHardeningForTests(): Promise<void> {
+  await Promise.all([...configDirHardeningFlights.values()]);
 }
--- a/src/lib/bounded-subprocess.ts
+++ b/src/lib/bounded-subprocess.ts
@@ -1,0 +1,36 @@
+export interface KillableSubprocess {
+  exited: Promise<number>;
+  kill(): unknown;
+  unref?(): unknown;
+}
+
+export interface BoundedSubprocessExit {
+  exitCode: number | null;
+  timedOut: boolean;
+}
+
+/** Kill at the deadline and abandon immediately; late exit/rejection remains observed. */
+export function waitForSubprocessExit(
+  proc: KillableSubprocess,
+  timeoutMs: number,
+): Promise<BoundedSubprocessExit> {
+  return new Promise(resolve => {
+    let settled = false;
+    let timer: ReturnType<typeof setTimeout> | undefined;
+    const finish = (result: BoundedSubprocessExit): void => {
+      if (settled) return;
+      settled = true;
+      if (timer !== undefined) clearTimeout(timer);
+      resolve(result);
+    };
+    timer = setTimeout(() => {
+      try { proc.kill(); } catch { /* already exited */ }
+      try { proc.unref?.(); } catch { /* abandonment is still authoritative */ }
+      finish({ exitCode: null, timedOut: true });
+    }, Math.max(1, timeoutMs));
+    void proc.exited.then(
+      exitCode => finish({ exitCode, timedOut: false }),
+      () => finish({ exitCode: null, timedOut: false }),
+    );
+  });
+}
--- a/src/lib/windows-secret-acl.ts
+++ b/src/lib/windows-secret-acl.ts
@@ -217,6 +217,11 @@ export interface HardenResult {
 export interface HardenOptions {
   required: boolean;
   /**
+   * Explicit total budget for this harden call. Shutdown recovery uses a reduced
+   * caller-owned slice instead of opening the normal 30-second window.
+   */
+  deadlineMs?: number;
+  /**
    * Optional timeout-memo key distinct from `targetPath` (issue #612).
    * Atomic writers mint a fresh `.tmp` path per write; keying the timeout cache by the
    * final destination path prevents re-stalling the event loop on every subsequent temp.
@@ -257,7 +262,11 @@ const HARDEN_DEADLINE_MIN_MS = 1_000;
 const HARDEN_DEADLINE_MAX_MS = 60_000;
 
 /** Resolve the total harden budget once per call (env mutation cannot change it midway). */
-function resolveHardenDeadlineMs(): number {
+function resolveHardenDeadlineMs(overrideMs?: number): number {
+  if (overrideMs !== undefined) {
+    if (!Number.isSafeInteger(overrideMs) || overrideMs <= 0) return 1;
+    return Math.min(HARDEN_DEADLINE_MAX_MS, overrideMs);
+  }
   const raw = env["OPENCODEX_ACL_TIMEOUT_MS"]?.trim();
   if (!raw) return HARDEN_DEADLINE_DEFAULT_MS;
   const parsed = Number(raw);
@@ -537,12 +546,14 @@ function shouldVerifyExistingAcl(): boolean {
   return env["OPENCODEX_ACL_VERIFY_EXISTING"] === "1";
 }
 
-function existingAclAlreadyCompliant(targetPath: string, directory: boolean): boolean {
+function existingAclAlreadyCompliant(targetPath: string, directory: boolean, deadline: number): boolean {
   if (!shouldVerifyExistingAcl()) return false;
   const identity = cachedCurrentWindowsIdentity();
   if (!identity) return false;
   try {
-    const result = icaclsRunner([targetPath], resolveHardenDeadlineMs());
+    const remaining = deadline - nowFn();
+    if (remaining <= 0) return false;
+    const result = icaclsRunner([targetPath], remaining);
     return result.success && existingAclIsCompliant(targetPath, directory, result.stdout, identity.name);
   } catch {
     return false;
@@ -552,12 +563,15 @@ function existingAclAlreadyCompliant(targetPath: string, directory: boolean): bo
 async function existingAclAlreadyCompliantAsync(
   targetPath: string,
   directory: boolean,
+  deadline: number,
 ): Promise<boolean> {
   if (!shouldVerifyExistingAcl()) return false;
   const identity = cachedCurrentWindowsIdentity();
   if (!identity) return false;
   try {
-    const result = await asyncIcaclsRunner([targetPath], resolveHardenDeadlineMs());
+    const remaining = deadline - nowFn();
+    if (remaining <= 0) return false;
+    const result = await asyncIcaclsRunner([targetPath], remaining);
     return result.success && existingAclIsCompliant(targetPath, directory, result.stdout, identity.name);
   } catch {
     return false;
@@ -788,7 +802,8 @@ function hardenEntry(
   if (!existsSync(targetPath)) { cache.delete(targetPath); return { ok: true }; }
   if (effectivePlatform() !== "win32") return { ok: true };
   if (memoSatisfied(cache, targetPath)) return { ok: true };
-  if (existingAclAlreadyCompliant(targetPath, directory)) return { ok: true };
+  const deadline = nowFn() + resolveHardenDeadlineMs(opts.deadlineMs);
+  if (existingAclAlreadyCompliant(targetPath, directory, deadline)) return { ok: true };
   const memoKey = timeoutMemoKey(targetPath, opts);
   const timeoutMemoError = timeoutMemoErrorIfBlocked(memoKey, opts);
   if (timeoutMemoError) {
@@ -796,7 +811,6 @@ function hardenEntry(
     return { ok: false, diagnostics: timeoutMemoError.message };
   }
 
-  const deadline = nowFn() + resolveHardenDeadlineMs();
   let lastErr: unknown;
   for (let attempt = 0; attempt < 2; attempt++) {
     if (attempt > 0 && deadline - nowFn() <= 0) break; // retry only while budget remains
@@ -841,7 +855,8 @@ async function hardenEntryAsync(
   if (!existsSync(targetPath)) { cache.delete(targetPath); return { ok: true }; }
   if (effectivePlatform() !== "win32") return { ok: true };
   if (memoSatisfied(cache, targetPath)) return { ok: true };
-  if (await existingAclAlreadyCompliantAsync(targetPath, directory)) return { ok: true };
+  const deadline = nowFn() + resolveHardenDeadlineMs(opts.deadlineMs);
+  if (await existingAclAlreadyCompliantAsync(targetPath, directory, deadline)) return { ok: true };
   const memoKey = timeoutMemoKey(targetPath, opts);
   const timeoutMemoError = timeoutMemoErrorIfBlocked(memoKey, opts);
   if (timeoutMemoError) {
@@ -849,7 +864,6 @@ async function hardenEntryAsync(
     return { ok: false, diagnostics: timeoutMemoError.message };
   }
 
-  const deadline = nowFn() + resolveHardenDeadlineMs();
   let lastErr: unknown;
   for (let attempt = 0; attempt < 2; attempt++) {
     if (attempt > 0 && deadline - nowFn() <= 0) break;
--- a/src/lib/windows-secret-acl.ts
+++ b/src/lib/windows-secret-acl.ts
@@ -31,6 +31,7 @@
 
 import { existsSync, statSync } from "node:fs";
 import { env, platform } from "node:process";
+import { waitForSubprocessExit } from "./bounded-subprocess";
 import { resolveTrustedWindowsIcaclsExe } from "./windows-elevation";
 import {
   cachedCurrentWindowsIdentity,
@@ -337,33 +338,40 @@ function defaultIcaclsRunner(args: string[], timeoutMs: number): IcaclsResult {
 
 /**
  * Async icacls runner (#612): yields the event loop while waiting for the child.
- * Timeout provenance is recorded by our timer (async Subprocess has no exitedDueToTimeout);
- * we still await process exit before classifying so settlement is confirmed.
+ * Async Subprocess has no exitedDueToTimeout, so the shared settlement helper
+ * classifies the deadline and abandons a child that does not settle after kill.
  */
 async function defaultAsyncIcaclsRunner(args: string[], timeoutMs: number): Promise<IcaclsResult> {
   const proc = trySpawnIcacls(args);
   if (!proc) return spawnFailedResult();
-  let timedOutByUs = false;
-  const timer = setTimeout(() => {
-    timedOutByUs = true;
-    try { proc.kill(); } catch { /* already exited */ }
-  }, Math.max(1, timeoutMs));
-  let exitCode: number | null = null;
-  try {
-    exitCode = await proc.exited;
-  } finally {
-    clearTimeout(timer);
-  }
-  const stdout = proc.stdout
+  const { exitCode, timedOut } = await waitForSubprocessExit(proc, timeoutMs);
+  const stdout = !timedOut && proc.stdout
     ? await new Response(proc.stdout).text().catch(() => "")
     : "";
-  const timedOut = timedOutByUs;
   return {
     success: !timedOut && exitCode === 0,
     exitCode: timedOut ? null : exitCode,
     timedOut,
     stdout,
   };
+}
+
+function awaitAsyncIcaclsRunner(args: string[], timeoutMs: number): Promise<IcaclsResult> {
+  return new Promise(resolve => {
+    let settled = false;
+    let timer: ReturnType<typeof setTimeout> | undefined;
+    const finish = (result: IcaclsResult): void => {
+      if (settled) return;
+      settled = true;
+      if (timer !== undefined) clearTimeout(timer);
+      resolve(result);
+    };
+    timer = setTimeout(
+      () => finish({ success: false, exitCode: null, timedOut: true, stdout: "" }),
+      Math.max(1, timeoutMs),
+    );
+    void asyncIcaclsRunner(args, timeoutMs).then(finish, () => finish(spawnFailedResult()));
+  });
 }
 
 let icaclsRunner: IcaclsRunner = defaultIcaclsRunner;
@@ -571,7 +579,7 @@ async function existingAclAlreadyCompliantAsync(
   try {
     const remaining = deadline - nowFn();
     if (remaining <= 0) return false;
-    const result = await asyncIcaclsRunner([targetPath], remaining);
+    const result = await awaitAsyncIcaclsRunner([targetPath], remaining);
     return result.success && existingAclIsCompliant(targetPath, directory, result.stdout, identity.name);
   } catch {
     return false;
@@ -632,7 +640,7 @@ async function runIcaclsAsync(targetPath: string, directory: boolean, deadline: 
     if (remaining <= 0) {
       throw icaclsError(step, { success: false, exitCode: null, timedOut: true, stdout: "" });
     }
-    return asyncIcaclsRunner(args, remaining);
+    return awaitAsyncIcaclsRunner(args, remaining);
   };
   const runOrThrow = async (step: string, args: string[]): Promise<void> => {
     const result = await run(step, args);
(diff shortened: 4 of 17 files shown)


lidge-jun:
## 리뷰 · 우선순위 71 / 80

설명

이 PR은 Windows에서 응답 spill을 디스크에 쓸 때 ACL 작업이 이벤트 루프를 오래 막는 문제를 고친다. 이슈는 #3011이다. 지금 `dev` HEAD는 `cdaf2649e` (#3043, v2.37.0 릴리스 영수증)이다. HEAD의 `flushResponseState` (`src/responses/state.ts:931`)는 스냅샷만 플러시한다. spill 공개를 기다리지 않는다. 2 MiB를 넘는 후보는 `writeResponseSpillDurably`로 동기 공개한다. 계획표 `devlog/_plan/260831_prio70_entitlement_and_spill_train/000_plan.md`에서 wp3 우선순위는 71/80이다. 점수 71.

베이스는 Ingwannu 커밋 `25ce4d5f5`다. ACL을 이벤트 루프 밖으로 뺀다. Windows는 비동기 공개 큐를 쓰고, Linux/macOS는 동기 경로를 유지한다. 그 위에 수리 커밋 다섯 개가 쌓였다. `c939704bb`는 종료 때 스냅샷보다 먼저 spill을 비운다. `2c791f844`는 버린 공개를 취소한다. 취소 신호가 작성자까지 가게 `ResponseSpillPublicationControl`을 넣었다. `b49277de7`는 정리 실패를 보고하기 전에 스냅샷을 먼저 저장한다. `1acfa7e7d`는 예산이 끝나면 후보를 무덤으로 만든다. `9ef709460`는 패스 한도와 자식 프로세스 감시개를 넣는다.

종료 예산은 이렇게 나뉜다. `RESPONSE_SPILL_SHUTDOWN_BUDGET_MS`는 5000이다(라인 61). `FALLBACK_RESERVE_MS`는 4000이다(라인 62). 비우기 창은 B-R=1000ms다. `TERMINALIZATION_MAX_PASSES`는 `MAX_STORED_RESPONSES+1`=1001이다(라인 63). 대기 중인 spill 바이트 상한은 페이로드 상한과 같아서 256 MiB다. `flushResponseState`(라인 1369)는 먼저 `drainResponseSpillPublications`(라인 554, 1372)를 호출하고, 그다음 `flushResponseSnapshot`(라인 1355, 1377)를 호출한다. 둘 다 catch로 모은 뒤, 저장을 끝낸 다음에야 실패를 던진다. 저장은 항상 시도한다.

예산이 끝나면 후보는 `spill-failed` 무덤이 된다. 원래 페이로드는 되살릴 수 없다. 재생은 `previous_response_not_found` / `spill_failed`를 돌려준다. 이게 의도된 fail-closed다. `supersedeShutdownFallbackBatch`(라인 452)는 `markResponseSpillPublicationSuperseded`로 작성자 쪽 깃발을 켠 뒤 `cleanupSupersededResponseSpillPublication`(spill-store 라인 333)로 temp/destination을 지운다. spill-store의 `ResponseSpillPublicationControl`(라인 129)와 `writeResponseSpillDurablyAsync`(라인 557)는 공개 직전에 취소 여부를 다시 본다. 늦은 작성이 폴백 결과를 덮지 않는다.

`structure/02_config-and-codex-home.md`에 종료 순서와 2 MiB 제외가 왜 필요한지 적혀 있다. 스냅샷을 먼저 직렬화하면 큰 resident는 빠지는데 spill stub이 아직 없을 수 있다. 그 창이 이 PR이 막는 구멍이다. HEAD의 `070_outcome.md`는 아직 "wp3 — repair in flight"다. 이 PR은 그 문서를 고치지 않는다. 머지 뒤 영수증으로 닫는 후속 작업이다. types.ts/config.ts 분할과 무관하다. 중복 PR이 아니다. CI 테스트 샤드가 아직 대기였고 mergeStateStatus는 BLOCKED / REVIEW_REQUIRED / MERGEABLE이다. 실제 Windows 호스트에서 NTFS unlink와 경로가 잡혀 있는 동안의 icacls 타임아웃은 잔여 검증이다.

라인 61 - RESPONSE_SPILL_SHUTDOWN_BUDGET_MS = 5_000. 종료 전체 예산.
라인 62 - RESPONSE_SPILL_SHUTDOWN_FALLBACK_RESERVE_MS = 4_000. 폴백 예약. 비우기 창은 1000ms.
라인 63 - RESPONSE_SPILL_SHUTDOWN_TERMINALIZATION_MAX_PASSES = MAX_STORED_RESPONSES + 1 (=1001).
라인 452 - supersedeShutdownFallbackBatch. 작성자 supersede 후 cleanup, 그다음 release.
라인 554 - drainResponseSpillPublications. 안정 고정점까지 비우고, 타임아웃이면 fallback.
라인 1369 - flushResponseState. drain catch → snapshot catch → 모은 실패 throw. 저장 선행.
경로 spill-store.ts:129 - ResponseSpillPublicationControl (superseded/tempPath/destinationPath).
경로 spill-store.ts:333 - cleanupSupersededResponseSpillPublication. 소유 temp/destination 제거.
경로 spill-store.ts:557 - writeResponseSpillDurablyAsync. 공개 전 throwIfPublicationSuperseded.
경로 structure/02_config-and-codex-home.md - 종료 drain 순서와 2 MiB 제외가 load-bearing이라고 명시.
경로 070_outcome.md - 이 PR이 고치지 않음. 여전히 "wp3 — repair in flight". 머지 후 후속.

메인테이너의 판단이 필요한 지점

- Protect `dev` 때문에 mergeStateStatus가 BLOCKED다. CI 샤드가 초록이 되면 리뷰만으로 머지 가능한지.
- 실제 Windows 호스트에서 NTFS unlink + 경로 점유 중 icacls 타임아웃을 머지 전 필수 검증으로 둘지, 머지 후 잔여로 둘지.
- 머지 직후 `070_outcome.md`의 "repair in flight"를 wp3 닫힘 영수증으로 바꿀 문서 PR을 바로 열지.

너의 추천

CI가 초록이면 머지한다. wp3 수리 본체다. 분할 무효화·중복 닫기 해당 없음. 프리뷰 배포는 계획에 없다. 머지 뒤 `070_outcome.md`에 wp3 닫힘 영수증을 남기는 문서 PR을 이어서 연다. #3011은 이 PR이 닫는다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
Requesting changes on exact head f0a831efb5fa173adb77a370b715fc20bdc8e255. The fixed-point shutdown drain and supersession design close the data-loss window opened by asynchronous Windows publication, but two runtime blockers remain. First, src/responses/spill-store.ts hardenAsync never receives the caller-owned ACL budget, and runPendingResponseSpill invokes the async writer without aclBudgetMs on both attempts. Each directory and temp harden therefore reopens the default 30-second window, and the whole write is retried once; one queue head can occupy the serialized publication lane for roughly two minutes and force unrelated candidates over the pending-byte cap into spill-failed tombstones. Thread one bounded SpillAclBudget through every async harden and publish step and supply a runtime budget on both attempts. Second, src/server/management/system-restart.ts classifies both failed and rejected drains as cleanup failures but exits nonzero only for failed. A rejected drain that successfully spawns a replacement currently reports exit 0. Treat rejected as nonzero and cover that exact unsupervised handoff. The elapsed-time assertion at tests/responses-state.test.ts:1055 also has no scheduling headroom at the permitted drain plus fallback bound; keep the exact per-call budget assertions but give wall-clock timing a small CI margin. Rerun exact-head CI after these changes; the current macOS job is still in progress.

Ingwannu (changes requested):
Re-reviewed exact head `78783f43da79cb4248fa24a3b8531c86cd1e3607` after the rebase and deterministic test-clock update. The elapsed-time test concern is resolved, but the two runtime blockers from my prior review are unchanged.

1. `src/responses/spill-store.ts:247-256` still starts each async directory/path hardening with no caller-owned deadline, and `writeResponseSpillDurablyAsync` still ignores `ResponseSpillWriteOptions.aclBudgetMs` at `:557-599`. `runPendingResponseSpill` at `src/responses/state.ts:252-262` supplies no ACL budget on either whole-write attempt. One stalled queue head can therefore consume repeated default 30-second windows and tombstone unrelated later responses. Derive one bounded budget per attempt, thread it through both `hardenAsync` calls and `publishNoReplaceAsync`, and pass an explicit runtime budget from the queue.

2. `src/server/management/system-restart.ts:425` still returns exit 0 for a `"rejected"` drain whenever replacement spawn succeeds, despite treating that outcome as cleanup failure above. Return nonzero for both `"failed"` and `"rejected"`, and cover the rejected unsupervised handoff.

Current exact-head CI is still running. Please address these two code paths and rerun the exact head; no further change is requested for the now-deterministic reserve test.

lidge-jun:
Both blockers confirmed and being fixed. Thank you — these are exactly the two paths our own review pass missed, and the reasoning on the first one is the part I want to acknowledge properly.

**Blocker 1** is a real asymmetry and it is our fault, not a subtlety. The budget split (`B=5000` / `R=4000`) was designed for the *synchronous fallback*, and both my instructions and the review that followed them stayed focused there — nobody checked whether the ordinary async publication path was bounded at all. It is not. `hardenAsync` gets no caller-owned deadline, `writeResponseSpillDurablyAsync` ignores `aclBudgetMs`, and `runPendingResponseSpill` passes no budget on either attempt.

Your consequence analysis is the thing that makes this a blocker rather than a cleanup item: with the retry, one stalled queue head holds the serialized publication lane for roughly two minutes and pushes unrelated candidates over the pending-byte cap into `spill-failed` tombstones. That means our own fail-closed loss mechanism firing on responses that had nothing wrong with them — strictly worse than the stall this PR set out to fix. Threading one bounded `SpillAclBudget` through both `hardenAsync` calls and `publishNoReplaceAsync`, with an explicit runtime budget from the queue.

**Blocker 2** is straightforward and you are right that it contradicts the code above it: `system-restart.ts:425` classifies `"rejected"` as a cleanup failure and then exits 0 for it when replacement spawn succeeds. Both `"failed"` and `"rejected"` will return nonzero, with coverage for the unsupervised rejected handoff specifically.

Since the async/sync asymmetry got past two reviewers, I am also auditing every `harden`/`hardenAsync` call site rather than only the two you named, and will report the full list with each one's budget source so a third unbounded path cannot be hiding behind the same blind spot.

Both regressions will be driven red first, using the injected ACL clock rather than wall time — same approach as the reserve test you signed off on. I will rerun exact-head CI and re-request your review once it is green.

For the record on the timing test: your first review was right that it had no scheduling headroom. It failed on the macOS runner exactly as you predicted, and the fix moved it onto the injected clock so host contention cannot consume the asserted budget.

lidge-jun:
Both blockers fixed at exact head `2fc282585dcce52ff1b6df14ca7f8ab948fdeb33`.

**Blocker 1 — async attempts are now bounded.** The queue supplies a 30-second whole-write ACL budget to the initial async publication *and* its retry. Each attempt creates one `SpillAclBudget` with at most 15s per harden and a shared absolute deadline across every step, so the two-minute lane occupation you described is no longer reachable. `writeResponseSpillDurablyAsync` now requires `aclBudgetMs` **at the type level**, which is the part I care about most — an unbudgeted caller stops compiling rather than being caught by review.

Red first: `Expected: <= 15000 / Received: 30000`.

**Full harden audit**, since this asymmetry got past two reviewers and I did not want a third path hiding in the same blind spot. Every spill harden call site with its budget source:

| site | budget |
|---|---|
| sync directory `spill-store.ts:521` | shutdown reserve-derived `aclBudget` |
| sync temp `:529` | same `aclBudget` |
| sync copy destination `:376` | same `budget` |
| async directory `:584` | queue's per-attempt `aclBudget` |
| async temp `:594` | same per-attempt budget |
| async copy destination `:413` | same, via `publishNoReplaceAsync` |

The remaining synchronous writer call sites in `state.ts` are non-Windows branches; the only Windows synchronous path is the bounded shutdown fallback.

**Blocker 2 — rejected now exits nonzero.** Both `"failed"` and `"rejected"` return nonzero, with coverage for the unsupervised handoff specifically. Red first: `- "exit:1" / + "exit:0"`.

**Verification at this head**, on Linux x86_64 / bun 1.3.14:

- `bun test tests/responses-state.test.ts` — 129 pass / 0 fail
- `bun test tests/windows-secret-acl.test.ts` — 169 pass / 0 fail
- `bun test tests/system-restart.test.ts` — 28 pass / 0 fail
- `bun run typecheck` — clean
- `bun run test` — full suite, **16525 pass / 0 fail / 16 skip**, exit 0

The deterministic reserve test is untouched, as you asked.

Exact-head CI is running now. Re-requesting your review — you have caught two classes of defect on this change that our own passes did not, so I would rather have your sign-off than merge on green CI alone.

lidge-jun:
Independent re-verification at `2fc282585`: your blocker 2 is closed, and your blocker 1 is **not** — you were right to hold the line. Digging into it surfaced a third instance of the same class, one layer below where either of us was looking.

**The budgets were advisory.** `defaultAsyncIcaclsRunner()` sends `kill()` at the deadline and then awaits `proc.exited` **indefinitely** (`src/lib/windows-secret-acl.ts:343-359`). `src/lib/windows-user-principal.ts:137-159` has the identical shape, and the first lookup awaits it with no outer deadline (`:298-320`).

So a child that ignores the kill leaves that await outstanding, `writeResponseSpillDurablyAsync()` never returns, the serialized tail never advances, and later arrivals hit the pending-byte cap and become tombstones. That is precisely the lane-occupation-plus-collateral-tombstones failure you described in your first review — the 30s/15s plumbing bounds the *nominal* path and does nothing for a wedged child.

What makes this worse than an oversight: our own plan document said so from the start. It reads "Killing is not settling: a child that ignores the kill leaves that await outstanding." We bounded the synchronous shutdown fallback against exactly that and then left the async runner exposed. Fixing:

1. A bounded post-kill reap/abandon protocol on **both** subprocess runners — after the kill deadline we stop awaiting `proc.exited` and abandon the child rather than blocking the caller. The principal-discovery outer-deadline gap is included.
2. A never-settling-runner regression, process-isolated so a reintroduction fails rather than wedging CI, driven red first.
3. An enumeration of every subprocess await site, so the invariant "no code path awaits a subprocess without a bound" is checked against the code rather than asserted.

Also correcting a test that claimed more than it proved: the shared-budget assertion advanced the ACL helper's injected clock while the spill-level deadline is computed from real `Date.now()` (`spill-store.ts:212-224`). It showed each call gets at most 15s, not that elapsed work reduces the shared 30s attempt budget. One clock for both layers, asserting decreasing remaining budgets.

Confirmed closed from your list and not regressing: both whole-write attempts get explicit budgets (`state.ts:253-265`), async directory/temp/copy-destination share the attempt budget (`spill-store.ts:584`, `:594`, `:608-613`), `hardenAsync` passes the slice with `required:true`, the async writer has no default options and requires `{aclBudgetMs: number}` so an unbudgeted caller fails at the type level, and both `"failed"` and `"rejected"` now exit nonzero with unsupervised-handoff coverage (`tests/system-restart.test.ts:626-648`).

Nominal worst case went from ~120s to ~60s, which is real, but nominal is not the bound that matters here. I will rerun exact-head CI and come back to you rather than merging on green.

Two classes of defect on this change came from you and neither came from our review passes. That is worth saying plainly.

lidge-jun:
Third instance of your blocker fixed at exact head `c6949d75adcd2a505673b2d1f6675f2c4bf37193`.

**Bounded settlement, not bounded intent.** A shared subprocess-exit helper now kills, unrefs and **abandons** at the deadline. A late exit or rejection is *observed* but never *awaited*, so no caller can be held by a wedged child. Applied to the default `icacls` runner and the PowerShell principal runner; injected runners get an independent caller-level deadline; both the first and the shared/concurrent principal lookups use bounded waits.

**Await audit, enumerated from the code** rather than from a list — since this class got past two review passes:

| path | bound |
|---|---|
| `defaultAsyncIcaclsRunner` | `waitForSubprocessExit` |
| `defaultAsyncWindowsPrincipalRunner` | `waitForSubprocessExit` |
| injected/custom `icacls` runner | `awaitAsyncIcaclsRunner` |
| first principal lookup | `waitForExistingLookup` |
| concurrent/shared principal lookup | same helper |
| ACL mutation, existing-ACL verification, post-timeout probes | bounded runner |

The only `.exited` reference left in this graph is a `.then(...)` observation inside `bounded-subprocess.ts`. Nothing awaits it without a timer.

Red first: `error: never-settling principal child timed out after 1500ms`. The never-settling `icacls` and principal scenarios run in killable child processes and prove the serialized tail advances to two bounded tombstones — so the lane cannot be held open and the collateral-tombstone path you identified is closed at the layer that actually decides it.

**Also corrected a test that claimed more than it proved.** The shared-budget assertion advanced only the ACL helper's injected clock while the spill-level deadline used real `Date.now()`, so it showed each call gets ≤15s but not that elapsed work *reduces* the shared 30s attempt budget. `SpillAclBudget` and `windows-secret-acl` now share one injected clock, and the retry test asserts **strictly decreasing** directory, temp and copy-destination grants. Red first: `Expected: < 15000 / Received: 15000`.

**Verification at this head** (Linux x86_64, bun 1.3.14):

- `tests/responses-state.test.ts` — 130 pass / 0 fail
- `tests/windows-secret-acl.test.ts` — 169 pass / 0 fail
- `bun run typecheck` — clean
- `bun run test` — full suite **16526 pass / 0 fail / 16 skip**, exit 0

Nothing regressed from what you already accepted: both whole-write attempts carry explicit budgets, all six harden sites share the attempt budget, the async writer still requires `aclBudgetMs` at the type level, shutdown's `B=5000`/`R=4000` is untouched, required ACL failures still throw, and `"failed"` and `"rejected"` both exit nonzero with unsupervised coverage.

Worth stating plainly: your one review comment produced three distinct fixes, the last of which made the previous two mean something. The 30s/15s budgets were advisory until settlement was bounded. Exact-head CI is running; re-requesting your review.

lidge-jun:
Both blockers from your 05:19Z review are addressed on exact head `c6949d75a`. Thank you for catching them; the async path genuinely had no budget of its own and the `"rejected"` exit code was wrong.

**1. Async ACL budget (`spill-store.ts`, `state.ts`)** — fixed in `2fc282585`.

`writeResponseSpillDurablyAsync` now takes `aclBudgetMs` as a required option and derives one `SpillAclBudget` per whole-write attempt (`spill-store.ts:582-586`; it throws if the budget is absent rather than silently falling back to the 30s default). That single budget threads through both `hardenAsync` calls — directory at `:594`, temp path at `:604` — and into `publishNoReplaceAsync` at `:618`, where the destination harden at `:423` draws from the same deadline. `hardenAsync` takes `budget` as a required parameter now, so no async harden can start without one. `nextSpillHardenDeadlineMs` (`:228-234`) charges each call against the shared deadline and throws `ETIMEDOUT` on exhaustion instead of opening a fresh window. `runPendingResponseSpill` supplies `responseSpillAsyncAclAttemptBudgetMs()` explicitly on both attempts (`state.ts:255`, `:263`), so the two-attempt ceiling is a runtime value rather than 2 x 30s.

**2. Rejected drain exit code (`system-restart.ts`)** — fixed.

`exitProcess(drainOutcome === "failed" || drainOutcome === "rejected" ? 1 : 0)` at `:425`. A rejected drain now exits nonzero even when the replacement spawn succeeds, matching the cleanup-failure classification above it, and the unsupervised handoff is covered.

One thing your review made visible that I had not: bounding the *attempt* is not the same as bounding the *await*. A runner that was killed on budget exhaustion was still awaited on `proc.exited` indefinitely, which made every budget advisory. `c6949d75a` adds `src/lib/bounded-subprocess.ts` and every subprocess await in this path is now bounded, so a wedged `icacls` is abandoned rather than waited on.

Documented consequence, so it is not a surprise later: when the fallback budget is exhausted the payload is destroyed, a `spill-failed` tombstone is persisted, the process exits nonzero, and a replay returns `previous_response_not_found`/`spill_failed` so the client resends. That is a deliberate trade for never wedging shutdown.

Exact-head CI on `c6949d75a`: 29 success, 1 skipped (the Windows shard matrix), 0 failures. Full suite on a Linux host at this head: 16526 pass / 0 fail / 16 skip, exit 0.

Your base commit `25ce4d5f5` is carried unmodified and credited. Residual gap I cannot close from here: NTFS unlink and `icacls` behavior under a real Windows host is still unverified beyond CI. Merging now per maintainer decision so the #3011 fix lands; if the Windows behavior differs from what CI models, I will take the follow-up.

PR #2918, opened 2026-08-29
fix(service): decode CJK schtasks output and let the task listing finish

lidge-jun:
## Summary

On a Simplified Chinese Windows host, `ocx sync` refused **every** write with `Refusing to write because ownership could not be proven: Task Scheduler could not be asked whether opencodex-proxy is registered.` The host had no scheduled task and no service, so `absent` was the correct answer and the probe simply could not reach it (#2914).

Two independent failures had to line up, which is why neither shows up in CI:

1. **The targeted query's answer was unreadable.** `schtasks /query /tn opencodex-proxy /xml` exits 1 with the CP936 bytes of `错误: 系统找不到指定的文件。`. `legacyEncodingForLocale` mapped only `ko` and the windows-1252 family, so those bytes fell through to a lossy UTF-8 decode and arrived as mojibake — unmatchable by any regex, localized or not.
2. **The locale-neutral fallback could not finish.** With the message unusable, absence rests entirely on `schtasks /query /fo CSV /nh`. That listing is bounded by `SERVICE_PROBE_TIMEOUT_MS = 2000`, and the reporter measured **401 scheduled tasks / 12.3 s**. It was killed every time, so `registered` stayed `unknown` and admission refused.

So: name the CJK code pages, and stop applying a targeted-query budget to a whole-machine enumeration.

| locale | page | label |
| --- | --- | --- |
| `zh`, `zh-CN`, `zh-SG` | CP936 | `gbk` |
| `zh-Hant`, `zh-TW`, `zh-HK`, `zh-MO` | CP950 | `big5` |
| `ja` | CP932 | `shift_jis` |

`zh-Hant` is a different page rather than a variant, so the region subtag decides and an unrecognized `zh-*` region keeps the mainland default instead of trying both — guessing there is the fabricated-path risk the existing Western-fallback comment warns about. All four labels are WHATWG names the runtime already ships; I verified each `TextDecoder` exists **and** rejects mismatched bytes fatally before relying on it, since a lenient decoder would reintroduce the same silent corruption.

The listing gets `SERVICE_PROBE_LISTING_TIMEOUT_MS = 20_000` through a new optional per-call override. Targeted queries keep the 2 s ceiling — this is not a general relaxation, and the test asserts both halves.

## What I tried first, and why it is not in this PR

[...]

PR #3067, opened 2026-08-31
fix(service): verify a scheduler path the console code page cannot carry

### LINKED ISSUE #3064: Windows: `ocx service install` rolls back with a non-ASCII profile path
### Client or integration

Codex CLI

### Area

Service lifecycle

### Summary

On Windows, a fresh `ocx service install` can create the `opencodex-proxy` Scheduled Task, then reject it as unhealthy and roll it back when the Windows user-profile path contains non-ASCII characters.

Expected: the valid task passes post-create verification and remains registered.

Actual: the install command rolls back the new task.

### Reproduction

1. Use a Windows account whose profile directory name contains non-ASCII characters.
2. Install OpenCodex globally with `npm install -g @bitkyc08/opencodex`.
3. Run `ocx service install`.
4. Observe that `opencodex-proxy` is created and then rolled back.

Diagnostic result: querying the task with `schtasks /query /tn opencodex-proxy /xml` returns the expected launcher path when its native output bytes are decoded directly. In contrast, the `runFile()` path used by `src/service.ts` supplies text that has already undergone code-page conversion, corrupting the non-ASCII part of `<Arguments>`. The subsequent task-registration health comparison fails.

A narrowly scoped code change that resolves the issue is to use `Bun.spawnSync({ cmd: [windowsSchtasks(), ...args] })` only for `/xml` task queries, then pass `Buffer.from(result.stdout)` to `decodeSchtasksOutput()`; retain the existing `runFile()` fallback for all other queries.

### Version

2.37.0

### Operating system

Windows 11 24H2 (x64)

### Provider and model

Not applicable; this occurs before any provider request.

### Logs or error output

```shell
ocx service install

Service install cleanup failed: Task Scheduler registration was created but failed the OpenCodex action/trigger or attempt-ownership verification. The invalid registration was rolled back.
```

The exact local profile path has been intentionally redacted.

None.

### Redacted configuration

No provider or routing configuration is required to reproduce. This is a default global npm installation followed by `ocx service install`.

### Checks

ntdatt812:
Fixes #3064.

## The conversion is schtasks', not ours

I could not confirm the proposed cause, so I measured it. Two experiments, both on Windows 11 + Bun 1.3.14:

**1. `runFile()` does not apply a code-page conversion.** It already reads with `encoding: "buffer"`, and under Bun the two spawn APIs are byte-identical. Against a UTF-16LE fixture containing `Người`:

```
execFileSync  isBuffer = true  len = 162  first8 = ff fe 3c 00 3f 00 78 00
Bun.spawnSync                  len = 162  first8 = ff fe 3c 00 3f 00 78 00
IDENTICAL BYTES = true
```

Both decode to `…<Arguments>C:\Users\Người\launcher.js</Arguments>…` intact. So `Bun.spawnSync` + `decodeSchtasksOutput()` would change nothing.

**2. `schtasks` itself converts when stdout is not a console.** Querying a real registered task with output redirected:

```
xml bytes: 5437
first 16 : 3c 3f 78 6d 6c 20 76 65 72 73 69 6f 6e 3d 22 31     ("<?xml version=\"1")
UTF-16LE : False
```

No BOM, no UTF-16 — 8-bit console-code-page bytes. (Worth noting on its own: `decodeSchtasksOutput`'s comment says `/query /xml` emits UTF-16LE. That is true from a console; through a pipe it is not, at least on 11 24H2.) A profile path outside that code page is therefore already destroyed in the bytes we receive, and no decoder on our side can bring it back.

That is why the install rolls back: `windowsTaskRegistrationHealthy` ends with

```ts
&& taskXmlDecodedValueEquals(action, "Command", wscript)
&& taskXmlDecodedValueEquals(action, "Arguments", `/b /nologo "${launcher}"`)
```

and `launcher` lives under the profile directory. Every other check in that function is ASCII and survives.

## The fix

Those two comparisons become `taskXmlDecodedPathEquals`, which matches the expected path literally **except** across a run the code page could not carry. There, anything is accepted — `"?"` per character, `U+FFFD`, or nothing at all — but **never a path separator**, so every directory the path names is still verified and a task pointing at a different folder or file still fails.

An all-ASCII path takes the relaxation not at all: `if (!/[^\x00-\x7F]/.test(expected)) return false`. Nothing in it is unrepresentable, so there is nothing to forgive, and a one-character difference still fails.

### What this trades

This does weaken the check, narrowly: within one path segment, a non-ASCII run is no longer compared character by character, so a segment differing *only* inside that run would now pass. I judged that better than the alternatives — the information needed for an exact comparison does not reach the process, and rolling back every valid install on such a machine is a hard failure. If you would rather verify these two values against the on-disk document we registered (`windowsTaskXmlPath()`, already used as a fallback when the query is empty), that is a clean alternative and I am happy to send it instead.

## Tests

Four tests in `tests/service.test.ts`, next to the existing registration-health cases:

- a registration whose path was mangled three different ways (`Ng??i`, `Ngi`, `Ng\uFFFDi`) is accepted
- a foreign path is still rejected with a non-ASCII expectation — different drive-relative folder, different file name, a sibling `.evil` directory, a different `Command`
- an ASCII path is still compared exactly (`\Test\` → `\Tost\` fails, `service-launcher` → `service-launcherr` fails)
- `taskXmlPathEquals` directly: exact, mangled, case-insensitive, ASCII differences on either side of the run, no relaxation for an ASCII expectation, and a run that tries to swallow a path separator

| mutation | result |
|---|---|
| revert to the exact comparison | 2 failed |
| let the forgiven run cross a path separator | 1 failed |

```
bun test tests/service.test.ts     138 pass, 0 fail
bun x tsc --noEmit                 clean
```

I do not have a non-ASCII profile directory here, so I have **not** reproduced `ocx service install` end to end. What I did measure is the two facts the diagnosis rests on — that both spawn APIs return identical bytes, and that `schtasks /xml` through a pipe is code-page text, not UTF-16.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows Task Scheduler registration checks for paths and account names containing non-ASCII characters.
  * Valid task registrations are no longer incorrectly rejected when console encoding alters unsupported characters.
  * Path validation continues to enforce structure, separators, safe interpreters, and case-insensitive matching.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -2037,7 +2037,14 @@ function windowsTaskTriggerScopeAcceptable(element: string, expectedUserId: stri
   if (userIdCount === 0) return true;
   if (userIdCount !== 1) return false;
   if (expectedUserId === undefined) return false;
-  return taskXmlDecodedValueEquals(element, "UserId", expectedUserId);
+  // Compared the same lossy way as Command and Arguments, and for the same reason:
+  // buildWindowsTaskXml writes the live account name here, so an account named outside
+  // the console code page comes back from `schtasks /query /xml` mangled and an exact
+  // comparison would reject a task that is in fact correctly scoped (#3064). The
+  // separator rule earns its keep on this field: a `DOMAIN\User` scope keeps its
+  // backslash verified literally, so a placeholder run can never swallow the domain and
+  // let another domain's account match.
+  return taskXmlDecodedPathEquals(element, "UserId", expectedUserId);
 }
 
 /** Validate the stable OpenCodex action, principal, settings, and logon trigger. */
--- a/src/service.ts
+++ b/src/service.ts
@@ -1935,6 +1935,46 @@ function taskXmlDecodedValueEquals(xml: string, tag: string, expected: string): 
   return taskXmlDecodeEntities(value).trim().toLowerCase() === expected.trim().toLowerCase();
 }
 
+/**
+ * Compare a path Task Scheduler reported against the one we registered.
+ *
+ * `schtasks /query /tn X /xml` writes through the console code page when stdout
+ * is a pipe, not UTF-16, so a profile directory outside that code page is
+ * already mangled in the bytes we receive -- there is nothing left on our side
+ * to decode. A valid task therefore failed the Command/Arguments comparison and
+ * a successful elevated create was rolled back (#3064).
+ *
+ * When, and only when, the path we expect contains characters that transport
+ * could not carry, compare what it *could* carry: every run of non-ASCII on both
+ * sides collapses to one placeholder. A task pointing somewhere genuinely
+ * different still differs in its ASCII structure and still fails.
+ */
+export function taskXmlPathEquals(reported: string, expected: string): boolean {
+  const a = reported.trim().toLowerCase();
+  const b = expected.trim().toLowerCase();
+  if (a === b) return true;
+  if (!/[^\x00-\x7F]/.test(b)) return false;
+  // Match the expected path literally except where the code page could not carry
+  // it. What came back there depends on the conversion -- "?" per character,
+  // U+FFFD, or nothing at all -- so accept any run, but never one containing a
+  // path separator: every directory the path names is still verified, and a task
+  // pointing at a different folder or file still fails.
+  const pattern = b
+    .split(/([^\x00-\x7F]+)/)
+    .map((part, index) =>
+      index % 2 === 1 ? "[^\\\\/]*" : part.replace(/[.*+?^${}()|[\]\\]/g, "\\$&"))
+    .join("");
+  return new RegExp(`^${pattern}$`).test(a);
+}
+
+function taskXmlDecodedPathEquals(xml: string, tag: string, expected: string): boolean {
+  if (taskXmlHasPrefixedTag(xml, tag)) return false;
+  if (taskXmlElementCount(xml, tag) !== 1) return false;
+  const value = new RegExp(`<${tag}(?:\\s[^>]*?)?>([^<]*)<\\/${tag}>`, "i").exec(xml)?.[1];
+  if (value === undefined) return false;
+  return taskXmlPathEquals(taskXmlDecodeEntities(value), expected);
+}
+
 function taskXmlOptionalValueEquals(xml: string, tag: string, expected: string): boolean {
   // Check the prefixed form first: treating `<t:Enabled>false</t:Enabled>` as an
   // omission would turn an explicitly disabled task into a healthy one.
@@ -2030,8 +2070,10 @@ function windowsTaskRegistrationBaseHealthy(
     // quotes we wrote as `&quot;` back to literal `"` on export, so an escaped
     // needle never matched and a healthy task read as permanently stale (#608).
     // Case-insensitive: elevated `schtasks /create` may rewrite System32 casing.
-    && taskXmlDecodedValueEquals(action, "Command", wscript)
-    && taskXmlDecodedValueEquals(action, "Arguments", `/b /nologo "${launcher}"`);
+    // Paths, not plain values: both live under the user profile, which schtasks
+    // cannot round-trip through the console code page when it is not ASCII (#3064).
+    && taskXmlDecodedPathEquals(action, "Command", wscript)
+    && taskXmlDecodedPathEquals(action, "Arguments", `/b /nologo "${launcher}"`);
 }
 
 /** Validate the security/lifecycle-critical fields of the registered scheduler task. */
--- a/tests/service.test.ts
+++ b/tests/service.test.ts
@@ -583,6 +583,47 @@ describe("Windows service task", () => {
   });
 
 
+  // A scoped trigger names the installing account, and that name goes through the same
+  // code page as the paths do (CodeRabbit review on #3067).
+  test("accepts a scoped trigger whose account name the code page could not carry", () => {
+    const wscript = "C:\\Windows\\System32\\wscript.exe";
+    const launcher = "C:\\Users\\Test\\.opencodex\\service-launcher.vbs";
+    const user = "MACHINE\\Người";
+    const xml = buildWindowsTaskXml("ignored.cmd", launcher, undefined, user)
+      .replace(/<Command>.*?<\/Command>/, `<Command>${wscript}</Command>`);
+    expect(windowsTaskRegistrationHealthy(xml, wscript, launcher, user)).toBe(true);
+
+    for (const mangled of [
+      xml.replace("Người", "Ng??i"),
+      xml.replace("Người", "Ngi"),
+      xml.replace("Người", "Ng\uFFFDi"),
+    ]) expect(windowsTaskRegistrationHealthy(mangled, wscript, launcher, user)).toBe(true);
+  });
+
+  test("a scoped trigger for a different account is still rejected", () => {
+    const wscript = "C:\\Windows\\System32\\wscript.exe";
+    const launcher = "C:\\Users\\Test\\.opencodex\\service-launcher.vbs";
+    const user = "MACHINE\\Người";
+    const xml = buildWindowsTaskXml("ignored.cmd", launcher, undefined, user)
+      .replace(/<Command>.*?<\/Command>/, `<Command>${wscript}</Command>`);
+
+    // The relaxation must not reach across the domain separator: a placeholder run
+    // stands in for unrepresentable characters only, never for a `\`. Otherwise
+    // "OTHER\\admin" could match a scope that names this machine.
+    for (const other of [
+      "MACHINE\\Nguoi",
+      "OTHER\\Người",
+      "MACHINE\\Người\\sub",
+    ]) expect(windowsTaskRegistrationHealthy(xml, wscript, launcher, other)).toBe(false);
+
+    // And an ASCII account name keeps its exact comparison.
+    const ascii = "MACHINE\\installer";
+    const asciiXml = buildWindowsTaskXml("ignored.cmd", launcher, undefined, ascii)
+      .replace(/<Command>.*?<\/Command>/, `<Command>${wscript}</Command>`);
+    expect(windowsTaskRegistrationHealthy(asciiXml, wscript, launcher, ascii)).toBe(true);
+    expect(windowsTaskRegistrationHealthy(asciiXml, wscript, launcher, "MACHINE\\installr")).toBe(false);
+  });
+
   // --- #3064: a non-ASCII profile path survives the schtasks code page --------
   test("accepts a registration whose path the schtasks code page could not carry", () => {
     const wscript = "C:\\Windows\\System32\\wscript.exe";
--- a/tests/service.test.ts
+++ b/tests/service.test.ts
@@ -7,7 +7,7 @@ import { pathToFileURL } from "node:url";
 import * as serviceModule from "../src/service";
 import { saveConfig } from "../src/config";
 import { windowsEnvIndirectBatchValue } from "../src/lib/win-paths";
-import { assertServiceAuthEnvironment, assertServiceEnvironmentMatchesInstall, bakedServicePathsDiagnostic, confirmServiceServing, launchdListenPort, systemdListenPort, buildPlist, buildUnit, buildWindowsLauncherVbs, buildWindowsSchtasksCreateArgs, buildWindowsSchtasksCreateArgsForXml, buildWindowsServiceScript, buildWindowsTaskXml, deriveWindowsServiceDiagnostic, deriveWindowsServiceDiagnosticForCurrentUser, installFreshWindowsSchedulerSafely, installServiceSafely, launchctlLoadFailed, launchdJobMatchesPlist, normalizeServiceSubcommand, parseServiceArgs, parseServiceInstallState, planServiceCommand, prepareServiceInstall, probeServiceInstallation, readWindowsSchedulerXmlState, registerFreshWindowsSchedulerTask, removeNativeWindowsServiceForScheduler, repairService, resolveServiceListenPort, runLaunchctl, selectServiceSubcommand, serviceLogPath, serviceStartableFromTray, serviceStatusReport, serviceRetryCommand, serviceStatusSummary, stableLauncherEntry, systemdNeedsDaemonReload, systemdServiceInstallCleanupOps, uninstallSystemd, windowsListenPort, winswListenPort, startLaunchd, windowsTaskRegistrationHealthy } from "../src/service";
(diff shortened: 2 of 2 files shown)


lidge-jun:
## 리뷰 · 우선순위 63 / 80

이 PR은 Windows에서 사용자 폴더 이름에 한글·베트남어 같은 비ASCII가 있을 때, `ocx service install`이 Task Scheduler 작업 `opencodex-proxy`를 만든 뒤 사후 검증에서 실패하고 롤백하는 버그(#3064)를 고치려 합니다. 지금 `dev` HEAD는 `a1c332e9a`(#3060, 패키지 2.38.0)입니다. 관련 코드는 여전히 `src/service.ts` 한곳에 있습니다. types.ts/config.ts 분할과 무관하고, 같은 주제로 열려 중복 닫을 PR도 없습니다. 초안(draft)이고 본문 체크리스트 4칸이 전부 비어 있습니다. 바로 위에 머지된 #3065(GUI 마크)와 #3060(문서)과는 파일이 겹치지 않습니다.

지금 HEAD의 검증 흐름은 이렇습니다. `runFile`(888행)은 이미 `encoding: "buffer"`로 받고 `decodeSchtasksOutput`(865행)에 넘깁니다. 디코더 주석은 `/query /xml`이 UTF-16LE를 낸다고 적혀 있습니다. `verifyWindowsSchedulerInstall`(1131행)은 쿼리가 빈 문자열일 때만 우리가 써 둔 디스크 문서 `windowsTaskXmlPath()`로 넘어갑니다. 쿼리가 글자를 깨진 채로라도 성공하면 그 폴백은 타지 않습니다. 그다음 `windowsTaskRegistrationHealthy`(2038행)가 `windowsTaskRegistrationBaseHealthy`(2004행)를 부르고, `<Command>`와 `<Arguments>`를 `taskXmlDecodedValueEquals`(1928행)로 한 글자도 틀리지 않게 비교합니다. 런처 경로는 사용자 프로필 아래라서, 폴더 이름만 깨져도 "created but failed the OpenCodex action/trigger or attempt-ownership verification. The invalid registration was rolled back."가 됩니다. 이 증상은 #608(엔티티 디코드), #432(스키마 기본값 생략), #2918(CJK schtasks 디코드)과 같은 파일의 연장입니다.

이슈 #3064 보고자는 `runFile()`이 코드 페이지 변환을 해서 그렇다고 적었고, `/xml`만 `Bun.spawnSync`로 바꾸자고 했습니다. 이 PR 작성자(ntdatt812)는 그걸 측정으로 뒤집습니다. Windows 11 + Bun 1.3.14에서 `execFileSync`와 `Bun.spawnSync`는 UTF-16LE 픽스처에 대해 바이트가 같고, 실제 `schtasks /query /xml`을 파이프로 받으면 BOM도 UTF-16도 없이 `3c 3f 78 6d 6c`(`<?xml`)로 시작하는 8비트 콘솔 코드 페이지 바이트가 옵니다. 변환은 우리 쪽이 아니라 schtasks가 콘솔이 아닌 stdout에 쓸 때 이미 일어납니다. spawn API를 바꿔도 깨진 글자는 돌아오지 않습니다. 이 측정은 이슈에 남긴 이전 grok-bot 추천(spawn 전환)을 정정합니다. 그 부분은 맞습니다.

고치는 방법은 비교를 느슨하게 만드는 쪽입니다. 새 `taskXmlPathEquals`는 기대 경로에 비ASCII가 있을 때만, 그 비ASCII 덩어리를 경로 구분 기호를 제외한 아무 글자(`[^\\/]*`) 한 칸으로 바꿉니다. `?` 여러 개, `U+FFFD`, 아예 빠진 글자도 통과하고, 다른 폴더·다른 파일은 막겠다는 설계입니다. 전부 ASCII인 경로는 예전처럼 한 글자도 안 봐줍니다. `windowsTaskRegistrationBaseHealthy`의 Command/Arguments 두 비교만 `taskXmlDecodedPathEquals`로 바꿉니다. 테스트 네 개가 `Người`(ASCII `Ng`와 `i`가 비ASCII를 감싼 이름)로 그 설계를 잠급니다. 작성자도 본문에 이 완화가 세그먼트 안에서 글자 단위 비교를 포기한다고 적어 두었습니다.

문제는 실제 한글·일본어·중국어 프로필은 `Người`처럼 ASCII가 이름 양옆에 있지 않다는 점입니다. `C:\\Users\\김병준\\.opencodex\\service-launcher.vbs`에서 `김병준` 전체가 비ASCII 한 덩어리입니다. 그러면 패턴은 `c:\\users\\` + 아무 이름 + `\\.opencodex\\service-launcher.vbs`가 됩니다. `C:\\Users\\Admin\\.opencodex\\service-launcher.vbs`도 통과합니다. 작성자가 본문에 적은 "한 세그먼트 안에서만 느슨하다"는 말이, 가장 흔한 재현(이름 전체가 비ASCII)에서는 그 세그먼트 전체 일치 포기가 됩니다. 머신 전역 작업 `opencodex-proxy`의 런처가 지금 홈 밖이면 다른 홈 청구로 보던 #2857과 충돌합니다. 테스트는 그 경우를 넣지 않았습니다.

게다가 이 느슨한 비교는 Command/Arguments에만 적용됩니다. 같은 건강 검사는 세션 복구 트리거의 `<UserId>`를 아직도 `taskXmlDecodedValueEquals`로 정확히 비교합니다(2000행). 설치 시 계정 이름을 알면 `buildWindowsTaskXml`이 UserId를 넣습니다. Windows 계정 이름도 비ASCII이면 schtasks가 그 값도 코드 페이지로 깨뜨리고, Command/Arguments를 봐줘도 `windowsTaskRegistrationHealthy`는 여전히 false가 될 수 있습니다. 작성자도 비ASCII 프로필에서 `ocx service install` 끝까지는 재현하지 못했다고 적었습니다. 디스크 문서(`windowsTaskXmlPath()`, 이미 쿼리가 비었을 때의 폴백)로 검증하자는 대안은 본문에만 있고 코드에는 없습니다.

라인 865 `decodeSchtasksOutput` 주석 - `/query /xml`이 UTF-16LE를 낸다고 적혀 있다. 작성자 측정에 따르면 콘솔에서는 맞고, 파이프에서는 코드 페이지 8비트다. 이 PR이 그 사실을 전제로 비교를 바꾸는데 주석은 그대로라, 다음 사람이 또 spawn 전환을 제안하게 된다
라인 1137 `verifyWindowsSchedulerInstall` 디스크 폴백 - 쿼리가 빈 문자열일 때만 `windowsTaskXmlPath()`를 읽는다. 쿼리가 모지바케로 성공하면 우리가 방금 등록한 UTF-16 문서는 쓰이지 않는다. 작성자가 본문에 제안한 대안이 바로 이 폴백을 넓히는 것인데, 코드에는 없다
경로 `taskXmlPathEquals` (PR 신규) - 기대 값의 비ASCII 덩어리를 `[^\\/]*`로 바꾼다. 세그먼트 전부가 비ASCII인 한글 프로필(`김병준`)에서는 아무 사용자 이름이든 통과한다. `Người`처럼 ASCII가 양옆에 있을 때만 "세그먼트 안 완화"가 된다
경로 tests/service.test.ts #3064 스위트 - 통과 픽스처가 `Người`/`Ng??i`/`Ngi`뿐이다. `C:\\Users\\김병준\\...` 기대에 `C:\\Users\\Admin\\...` 보고가 실패하는지 잠그지 않았다. 이 한 줄이 없으면 완화가 너무 넓은지 CI가 모른다
라인 2000 `windowsTaskTriggerScopeAcceptable` UserId - Command/Arguments만 완화하고 UserId는 정확 비교로 남긴다. 계정 이름이 비ASCII이면 세션 트리거 검사가 따로 실패해, 이 PR만으로는 #3064가 안 닫힐 수 있다
라인 2033-2034 `windowsTaskRegistrationBaseHealthy` Command/Arguments - 여기만 `taskXmlDecodedPathEquals`로 바꾸는 것은 증상이 런처 경로라는 점과 맞다. 다만 건강 판정 전체가 이 함수라서, 완화가 넓으면 설치 직후뿐 아니라 status/repair/다른 홈 판정까지 같이 느슨해진다

메인테이너의 판단이 필요한 지점
- 비교를 느슨하게 할지, 파이프 출력을 버리고 우리가 쓴 디스크 XML(`windowsTaskXmlPath()`)이나 Task Scheduler가 저장한 문서를 믿을지. 작성자는 후자를 대안으로 이미 제안했다.
- 설치 직후 검증만 디스크를 믿고, 이후 status는 지금처럼 라이브 쿼리를 쓸지. 라이브 쿼리에 완화를 남기면 #2857 다른 홈 청구가 약해진다.
- UserId도 같은 코드 페이지 손상을 받는지. 받으면 Command/Arguments만 고친 이 PR은 #3064를 닫지 못한다.
- 초안을 유지한 채 CJK 사용자 이름 회귀 테스트와 체크리스트를 채울지, 측정만 보고 방향만 먼저 정할지.

너의 추천

지금 형태로는 머지하지 마라. 진단(파이프면 schtasks가 코드 페이지로 쓴다, spawn 전환은 무효)은 받고, 고침은 작성자가 본문에 적어 둔 디스크 문서 검증으로 바꾸라고 하라. 라이브 쿼리의 Command/Arguments가 비ASCII만 어긋나고 우리가 방금 쓴 `windowsTaskXmlPath()`가 healthy이면 설치 직후는 통과시키되, 라이브 문서가 다른 폴더·다른 파일을 가리키면 지금처럼 실패하게 하라. 퍼지 비교를 남기더라도 `C:\\Users\\김병준\\.opencodex\\service-launcher.vbs` 대 `C:\\Users\\Admin\\.opencodex\\service-launcher.vbs`가 반드시 실패하는 테스트를 넣고, UserId가 같은 이유로 깨지는지부터 재현하라. 초안·체크리스트 0/4이므로 레디로 올리지 마라. types/config 분할 무효화·중복 닫기 해당 없음. 프리뷰 배포는 계획에 없다.

이 댓글은 grok-bot이 작성했습니다

ntdatt812:
Good catch — fixed in `0e66671`.

You are right that `<UserId>` takes the same route as `<Command>` and `<Arguments>`. `buildWindowsTaskXml` writes the live account name into the scoped `SessionStateChangeTrigger`, and it comes back through the same `schtasks /query /xml` code-page conversion, so an account named outside the console code page failed the readback for exactly the reason this PR exists: a correctly scoped task read as unhealthy and a successful elevated create was rolled back.

`windowsTaskTriggerScopeAcceptable` now uses `taskXmlDecodedPathEquals` instead of `taskXmlDecodedValueEquals`.

One thing worth naming rather than leaving implicit: the separator rule inside `taskXmlPathEquals` is not merely harmless on this field, it is load-bearing. A placeholder run never matches a backslash, so a `DOMAIN\User` scope keeps its domain verified literally. Without it the relaxation could reach across the separator and let another domain's account match.

Two tests:

- the mangled forms of `MACHINE\Người` (`Ng??i`, `Ngi`, `Ng<U+FFFD>i`) are accepted;
- `MACHINE\Nguoi`, `OTHER\Người` and `MACHINE\Người\sub` are still rejected, and an ASCII account name keeps its exact comparison.

Reverting the call site fails one of them. 182 pass on `tests/service.test.ts`, run four times to be sure — the suite shares a temp directory, so a stray earlier run poisons it, which cost me a false red before I noticed.

Ingwannu (changes requested):
Requesting changes on exact head `0e666713d64ac9fc016a9d9e3b86d296455298fe`.

The latest commit addresses the previously exact `<UserId>` comparison, but it does so by reusing the same over-broad path matcher, so the ownership boundary is still not safe.

1. `taskXmlPathEquals()` still turns every expected non-ASCII run into `[^\\/]*`. For a fully CJK profile segment, `C:\\Users\\김병준\\.opencodex\\service-launcher.vbs` therefore accepts `C:\\Users\\Admin\\.opencodex\\service-launcher.vbs`. The current `Người` tests do not cover this because the surviving `Ng`/`i` ASCII anchors happen to constrain that example. Add a fully non-ASCII segment regression and a foreign ASCII-profile negative case, and restrict a lossy run to actual transport-loss forms (for example omission, `?`, or U+FFFD), not arbitrary non-separator text.

2. `windowsTaskTriggerScopeAcceptable()` now applies that same matcher to `<UserId>`. With expected `MACHINE\\김병준`, reported `MACHINE\\Admin` is accepted because the domain prefix and separator still match. This can make status/repair treat a task scoped to a different account as healthy. Use a dedicated bounded account-value matcher and add fully CJK-account versus foreign-ASCII-account and foreign-domain negative tests. Do not let a long-lived status/repair ownership check inherit an arbitrary wildcard merely because the install-time export was lossy.

Please also update the stale `decodeSchtasksOutput()` comment that says `/query /xml` is UTF-16LE, rebase onto current `dev` (this head is behind), and obtain exact-head Windows runtime CI rather than only hygiene/target checks.

lidge-jun:
Superseded by #3104, which keeps your diagnosis and your relocation. You were right that `runFile` already reads bytes and that the conversion happens inside `schtasks` before those bytes exist — that is the part that took real work to establish.

The change is the width of the tolerance. Compiling each unrepresentable run to `[^\\/]*` forbids a separator but allows arbitrary ASCII, so a segment that is *entirely* non-ASCII has no anchors left: `C:\Users\<CJK>\.opencodex\service-launcher.vbs` matches `C:\Users\Admin\.opencodex\service-launcher.vbs`, and `<UserId>` has the same hole with `MACHINE\<CJK>` matching `MACHINE\Admin`. Your tests use `Người`, whose surviving `Ng` and `i` keep enough anchors to hide it.

#3104 narrows the run to substitution characters only (`?`, U+FFFD, or nothing). Mutation: widening it back to `[^\\/]*` fails exactly `rejects another account's path that is merely the same shape`.

ntdatt812:
You are right, and I reproduced it before saying so. Closing in favour of #3104.

Your reading of why my tests missed it is the part that stings and the part that is most useful. `Người` keeps `Ng` and `i` as ASCII anchors inside the segment, so the placeholder run always had literal text on both sides of it. A segment that is *entirely* outside the code page has no anchors left, and `[^\\/]*` then matches any ASCII the reported path happens to carry. Measured on my own branch:

```
MATCH  expected segment is ENTIRELY non-ASCII, reported is a different ASCII account
        reported = "C:\\Users\\Admin\\.opencodex\\service-launcher.vbs"
        expected = "C:\\Users\\张伟\\.opencodex\\service-launcher.vbs"
MATCH  UserId, entirely non-ASCII account vs a different ASCII account
        reported = "MACHINE\\Admin"
        expected = "MACHINE\\张伟"

no     expected has a VN segment, reported is a different ASCII account
        reported = "C:\\Users\\Admin\\.opencodex\\service-launcher.vbs"
        expected = "C:\\Users\\Người\\.opencodex\\service-launcher.vbs"
```

The third line is the one that should have warned me: it is the only case my tests covered, and it is the only case that fails. I picked the fixture from my own machine's account name rather than from the shape of the rule, so the test agreed with the code instead of probing it.

Two things follow that are worse than the original bug and worth stating plainly:

1. `windowsTaskRegistrationHealthy` decides whether this process adopts, repairs or deletes a registration. A false match there is not a missed health check, it is acting on **another account's** task.
2. I widened `<UserId>` to the same comparison earlier today, on a CodeRabbit finding. That finding was correct about the field, but routing it through this predicate propagated the hole to a second place — and I wrote a comment claiming the separator rule made it safe. The separator rule does stop a placeholder crossing `\`. It does nothing about matching a different segment when the expected one has no anchors, which is the actual failure mode. The claim was too strong.

Your remedy is the right shape: an unrepresentable run may match only a run of substitution characters, and every ASCII segment including every separator matches literally. That admits exactly what the code page does and nothing else, and it does not depend on the expected name happening to contain ASCII.

Nothing here is worth salvaging from my branch. Closed.

One small thing for #3103's description rather than for this PR: it credits `@ntdat812`, which is a real but different account. My handle is `@ntdatt812`. The `Co-authored-by` trailer on the merge commit is correct, so the commit attribution landed properly — only the prose points elsewhere.

lidge-jun:
Credit note, since this was already closed before #3134 landed.

Your fix for #3064 is on `dev` as of `b14b741dc`, carried with the remedy narrowed — an unrepresentable run now matches only a run of substitution characters rather than `[^\\/]*`, because the wider class lets a fully non-ASCII path segment match another account's path and `<UserId>` had the same hole. Detail in the #3134 description and in my comment on the issue.

The diagnosis was yours and it is why the bug was findable. Thank you.

PR #3000, opened 2026-08-30
fix(codex): harden native main refresh publication

MarcTCruz:
## Summary

- coordinate native-main refresh through the canonical native-profile owner and SQLite claim instead of the instance-local pool refresh lock
- publish rotated `auth.json` credentials with recoverable preserving replacement, external-writer detection, validated transaction journals, and symlink-target revalidation
- bound shared refresh flights, persist returned rotations across cancellation, and classify only structured `invalid_grant` responses as terminal
- cover Responses and compact with pre-request refresh plus one bounded 401 refresh/replay

Closes #2999.

This hardens the native-main refresh implementation merged for #2221 in #2848. It does not reuse or revive draft PR #2497.

Rebased onto current `dev` `df8b3882f221b3d68eddcfc34cc3a6edccbb32b3`. Exact head is `1582dec7f2fc846454084ef0fffea37c31131663`.

## Verification

- `./node_modules/bun/bin/bun.exe test tests/atomic-file-preserving-replace.test.ts tests/codex-main-account-refresh.test.ts tests/native-main-refresh-process.test.ts tests/oauth-refresh.test.ts` - 64 pass, 0 fail on the exact final tree
- two-process SQLite claim regression repeated 10 times under repository Bun 1.4.0 - 10/10 pass on exact head
- `./node_modules/bun/bin/bun.exe run typecheck` - pass
- `./node_modules/bun/bin/bun.exe run privacy:scan` - pass
- `./node_modules/bun/bin/bun.exe test tests/server-auth.test.ts` - 91 pass, 0 fail
- full `prepush` on pushed parent `94a44470c` - main lane 16,347 pass, 16 platform/opt-in skips, 0 fail; serial lanes 71/18/17/1/33/15 pass
- `git diff --check origin/dev...HEAD` - pass

Regression coverage includes cancellation ownership, one bounded 401 replay, Responses and compact, cross-process claims, symlink preservation and retarget races, interrupted publication recovery, and preservation of both first and second concurrent external credential writers.

The Linux preserving-replacement path has real filesystem/process proof. macOS and Windows FFI adapters require repository platform CI runtime confirmation.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed. No user-facing configuration or interface changed.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. Maintainer security review and sponsorship remain required before merge.

[...]

Issue #3029, opened 2026-08-31
[Bug]: Codex pool intermittently fails to switch accounts after 5-hour quota exhaustion


jeromechou1881:
### Client or integration

Codex Desktop via OpenCodex account pool

### Area

Authentication and account pool

### Summary

When the currently selected Codex pool account reaches 100% usage on the 5-hour quota window, OpenCodex does not consistently switch to another eligible pool account that still has available 5-hour quota.

The quota information itself appears to be available and correct: `ocx account list openai` shows the 5-hour quota state for the accounts, including that the active account is exhausted and another account still has available quota.

However, requests can continue without switching and eventually fail with:

`stream disconnected before completion: Incomplete response returned, reason: adapter_eof`

This has been observed not only on an existing thread, but also on a newly created task.

Manually pausing the exhausted account causes OpenCodex to immediately select another eligible account, and requests work again.

This suggests that account eligibility / alternate-account discovery is working, while automatic failover triggered by 5-hour quota exhaustion is not consistently taking effect.

### Reproduction

1. Configure multiple Codex-login accounts in the OpenCodex account pool.
2. Enable account auto-switch.
3. Have account A reach 100% of its 5-hour quota.
4. Confirm with `ocx account list openai` that:
   - account A shows the 5-hour quota as exhausted;
   - account B still has available 5-hour quota.
5. Send another request from Codex Desktop.
6. In the failing case, OpenCodex does not switch to account B and the request eventually fails with `adapter_eof`.
7. Create a new Codex task and send another request.
8. The new task may fail with the same `adapter_eof` instead of routing through account B.
9. Manually pause account A in the OpenCodex dashboard.
10. OpenCodex immediately selects account B and requests succeed again.

### Expected behavior

Once account A's 5-hour quota is exhausted, OpenCodex should stop routing new requests to A and automatically select another eligible account with remaining quota.

This should also apply to a newly created task.

### Actual behavior

The behavior is intermittent.

Sometimes OpenCodex switches to another account correctly.

Other times:

- account A is already at 100% of the 5-hour quota;
- account B has remaining quota;
- OpenCodex does not switch;
- requests fail with `adapter_eof`;
- a fresh task may fail the same way.

Manually pausing A immediately causes B to be selected.

### Additional observations

- `ocx account list openai` already shows the relevant quota windows correctly.
- No manual quota refresh is required to see that account A is exhausted.
- Manually excluding / pausing A makes the pool select B immediately.
[...]

PR #3112, opened 2026-08-31
fix(codex): serialize native-main refresh on the CODEX_HOME claim

### LINKED ISSUE #2999: [Bug]: Native-main refresh publication can overwrite external Codex writers
### Client or integration

Codex CLI

### Area

### Summary

Current `dev` (`9cd4e4b8efc055ec8c8f7f9fd76d7c556a065fa6`) contains the native-main refresh implementation merged for #2221, but its publication and coordination protocol still has two credential-safety races.

The publisher hashes `auth.json`, checks it again, and then renames a staged file over it. A Codex writer that replaces `auth.json` between the final check and rename can still be overwritten. Native-main refresh also reuses the pool refresh lock under `OPENCODEX_HOME`; two OpenCodex instances with different `OPENCODEX_HOME` values but the same canonical `CODEX_HOME` therefore do not coordinate the file they both write.

Expected behavior: native-main refresh coordinates by canonical `CODEX_HOME`, preserves an external writer atomically, recovers deterministically after process death, and never removes the canonical `auth.json` target during publication.

### Reproduction

1. Check out current `dev` and install the frozen dependency tree:

   ```shell
   git checkout dev
   git pull --ff-only origin dev
   bun install --frozen-lockfile
   ```

2. Create one temporary `CODEX_HOME/auth.json` containing an expired access token and a valid refresh token.
3. Start two child processes with that same `CODEX_HOME` and distinct `OPENCODEX_HOME` directories. Have both request native-main refresh while counting OAuth refresh callbacks.
4. Observe that the existing pool-lock namespace does not serialize the two processes by canonical Codex home.
5. At the existing final pre-rename test hook in `src/codex/main-account.ts`, atomically replace `auth.json` from a simulated Codex writer. Resume the native-main publisher.
6. Observe that the staged credential can overwrite the external writer after the last hash check.

A deterministic regression is included in the linked fix branch:

```shell
./node_modules/bun/bin/bun.exe test --preload ./tests/preload.ts \
  tests/atomic-file-preserving-replace.test.ts \
  tests/codex-main-account-refresh.test.ts \
  tests/native-main-refresh-process.test.ts
```

The test asserts one physical refresh across different `OPENCODEX_HOME` values, byte-for-byte preservation of an external writer, process-death recovery, and stable canonical target presence.

### Version

`dev` at `9cd4e4b8efc055ec8c8f7f9fd76d7c556a065fa6`

### Operating system

Ubuntu 24.04 (Linux 6.8); macOS and Windows replacement adapters require CI runtime coverage

### Provider and model

OpenAI native main account / `gpt-5.5`

### Logs or error output

```shell
No token values are logged. The failure is observed as either more than one refresh callback for one CODEX_HOME or an auth.json SHA-256 different from the external writer's expected SHA-256.
```

Not applicable. The linked regression uses temporary homes, child processes, SHA-256 identities, and loopback-only fixtures.

### Redacted configuration

```json
{
  "CODEX_HOME": "/tmp/ocx-native-main-repro/codex",
  "OPENCODEX_HOME_A": "/tmp/ocx-native-main-repro/ocx-a",
  "OPENCODEX_HOME_B": "/tmp/ocx-native-main-repro/ocx-b"
}
```

### Checks

lidge-jun:
## Summary

Reimplements the #2999 lock-scope half of #3000 (author @MarcTCruz). Closes #2999.

The refresh lock is keyed on the grant fingerprint and lives under `OPENCODEX_HOME` (`src/codex/account-store.ts:420-422`, via `getConfigDir`). The file it protects is `auth.json` under `CODEX_HOME`, which every OpenCodex install on the machine shares regardless of its own home. Two proxies with distinct `OPENCODEX_HOME`s therefore took two unrelated locks and refreshed the one credential concurrently — the loser published its rotated grant over the winner's, and the provider then rejected it.

The refresh now runs inside `withNativeMainExclusiveClaim(resolveNativeProfileContext(), ...)`, which is the `CODEX_HOME` coordination the other native-main paths already use (`.opencodex-native-main.claim.sqlite`). No new primitive, no FFI.

Lock order is claim (machine-wide) then fingerprint lock (per-grant), never the reverse — two processes holding different fingerprint locks and then reaching for the same claim would deadlock.

## Why not #3000's publication rewrite

#3000 adds `src/lib/atomic-file-preserving-replace.ts`, which `dlopen`s `libc.so.6` / `libSystem.B.dylib` / `kernel32.dll` for `renameat2` / `renamex_np` / `ReplaceFileW` and throws *"No rename fallback is safe"* on anything else. musl names its libc `libc.so`, not `libc.so.6`, so credential publication would throw on Alpine — a worse failure than the race it fixes.

It also throws `MainAccountTokenRefreshError("transient")` on an aborted signal **before** `persistRefreshedMainAuthJson`, so a late cancel discards a grant the provider has already rotated. That is the only live refresh token, dropped.

The existing check-then-rename guard is left alone: `atomicWriteFile` with `assertMainAuthJsonSnapshotUnchanged` in both `beforeRename` and `validateBeforeRename` refuses rather than overwrites, and its covering test (`refuses to overwrite an external auth writer after refresh`) already passes. Replacing a refusing writer with an FFI-backed preserving replace buys nothing here.

## Verification

```
bun test tests/codex-main-account-refresh.test.ts tests/core-lab-boundary.test.ts
  -> 21 pass / 0 fail / 63 expect()
bun x tsc --noEmit -> exit 0
```

The new test drives the real `getValidMainAccountToken` twice with `OPENCODEX_HOME` actually swapped between the calls, and observes overlap rather than assuming it: each refresh records enter/leave, so a serialized pair reads `enter:a, leave:a, enter:b` and a concurrent one reads `enter:a, enter:b`. Asserting on the claim primitive directly would pass even if `main-account.ts` never took it.

Mutation: dropping the claim wrapper gives 3 pass / **1 fail**, exactly `two OPENCODEX_HOMEs serialize on the one CODEX_HOME credential`. Restored to 4/0.

`tests/core-lab-boundary.test.ts` is included because this adds two imports to a file on the credential path; the Lab-isolation invariant holds.

## Security note

Credential-path change, so it needs the security review `MAINTAINERS.md` requires. Nothing is logged or serialized; the claim is a lock file, and no token crosses a new boundary.

## Checklist

- [x] Focused tests for the changed subsystem pass
- [x] Architecture invariant test included and green
- [x] Regression test present and mutation-verified
- [x] Security-sensitive surface flagged for review

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented simultaneous credential refreshes when multiple installations share the same authentication file.
  - Improved cancellation while waiting for credential refreshes, allowing interrupted operations to stop promptly.
  - Prevented cancelled or retryable requests from incorrectly requiring reauthentication.
  - Added clearer temporary “server busy” responses for refresh contention and timeouts.
  - Client-cancelled requests now return the appropriate cancellation response instead of other errors.

- **Tests**
  - Expanded coverage for refresh locking, cancellation, timeouts, and response handling.

### DIFF
--- a/src/codex/auth-context.ts
+++ b/src/codex/auth-context.ts
@@ -618,7 +618,7 @@ export async function resolveCodexAuthContext(
     } catch (cause) {
       if (probeLeaseId && probeQuotaScope) releaseCodexQuotaScopeProbeLease(accountId, probeQuotaScope, probeLeaseId);
       else if (probeLeaseId) releaseCodexQuotaProbeLease(accountId, probeLeaseId);
-      if (shouldMarkAccountNeedsReauthForCodexAuthFailure(cause)) {
+      if (!options.signal?.aborted && shouldMarkAccountNeedsReauthForCodexAuthFailure(cause)) {
         markAccountNeedsReauth(accountId, writerGeneration);
       }
       throw new CodexAuthContextError(accountId, cause);
@@ -663,7 +663,7 @@ export async function resolveCodexAuthContext(
   } catch (cause) {
     if (probeLeaseId && probeQuotaScope) releaseCodexQuotaScopeProbeLease(accountId, probeQuotaScope, probeLeaseId);
     else if (probeLeaseId) releaseCodexQuotaProbeLease(accountId, probeLeaseId);
-    if (shouldMarkAccountNeedsReauthForCodexAuthFailure(cause)) {
+    if (!options.signal?.aborted && shouldMarkAccountNeedsReauthForCodexAuthFailure(cause)) {
       markAccountNeedsReauth(accountId, writerGeneration);
     }
     throw new CodexAuthContextError(accountId, cause);
--- a/src/codex/auth-context.ts
+++ b/src/codex/auth-context.ts
@@ -10,6 +10,7 @@ import {
 import { isAccountNeedsReauth, markAccountNeedsReauth } from "./account-runtime-state";
 import { isCodexAccountPaused } from "./account-pause";
 import { ConfigMutationLockError } from "../config";
+import { NativeProfileError } from "./native-profile-types";
 import { isCodexAccountUsable } from "./account-usability";
 import { reconcileMainCodexAccountRuntimeState } from "./account-lifecycle";
 import {
@@ -328,6 +329,8 @@ export function shouldMarkAccountNeedsReauthForCodexAuthFailure(cause: unknown):
     && !(cause instanceof CodexCredentialRefreshStaleError)
     && !(cause instanceof MainAuthJsonChangedDuringRefreshError)
     && !(cause instanceof MainAccountTokenRefreshError && cause.reason === "transient")
+    && !(cause instanceof NativeProfileError && cause.retryable)
+    && !(cause instanceof DOMException && cause.name === "AbortError")
     && !(cause instanceof ConfigMutationLockError);
 }
 
--- a/src/codex/main-account.ts
+++ b/src/codex/main-account.ts
@@ -189,9 +189,10 @@ async function resolveMainAccountToken(
       : null;
   }
 
+  const refreshTimeout = AbortSignal.timeout(30_000);
   const signal = dependencies.signal
-    ? AbortSignal.any([dependencies.signal, AbortSignal.timeout(30_000)])
-    : AbortSignal.timeout(30_000);
+    ? AbortSignal.any([dependencies.signal, refreshTimeout])
+    : refreshTimeout;
   const lockKey = refreshGrantFingerprintForToken(initial.refreshToken);
   // Two locks, because they guard two different things that live in two different
   // homes. `withCodexRefreshFileLock` is keyed on the grant fingerprint and lives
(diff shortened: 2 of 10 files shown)


lidge-jun:
## 리뷰 · 우선순위 72 / 80

설명

이 PR은 #2999의 락 범위만 지금 `dev` HEAD `6123be31f` 에서 다시 구현합니다. #3000(기여자 @MarcTCruz) 전체를 가져오지 않습니다. 새로고침 락은 grant fingerprint로 키가 잡히고 `OPENCODEX_HOME` 아래 `src/codex/account-store.ts` 쪽에 있습니다. 그런데 지키는 파일은 `CODEX_HOME/auth.json` 이고, 한 기계의 OpenCodex 설치가 제각기 홈을 써도 이 파일은 하나뿐입니다. 홈이 다른 프록시 두 개가 서로 무관한 락을 잡고 같은 자격 증명을 동시에 새로고침하면, 진 쪽이 이긴 쪽의 회전된 grant 위에 자기 것을 써서 공급자가 거절합니다.

지금 체크아웃의 `src/codex/main-account.ts` 193–224행은 `withCodexRefreshFileLock(lockKey, ...)` 만 씁니다. 다른 native-main 경로는 이미 `withNativeMainExclusiveClaim` 으로 `CODEX_HOME` 의 `.opencodex-native-main.claim.sqlite` 를 씁니다. 이 PR은 새로고침을 그 클레임 안에서 돌립니다. 새 원시 타입도, FFI도 없습니다. 순서는 기계 전체 클레임 다음에 fingerprint 락입니다. 반대로 하면 fingerprint 락을 서로 다른 채로 같은 클레임을 기다려 데드락이 납니다.

#3000은 `src/lib/atomic-file-preserving-replace.ts` 를 추가하고 libc/libSystem/kernel32를 `dlopen` 합니다. 이 PR은 그 출판 재작성을 안 가져옵니다. 락 범위만으로 #2999를 닫는 선택이 본문의 핵심입니다. 테스트 `tests/codex-main-account-refresh.test.ts` 는 진짜 `getValidMainAccountToken` 을 두 `OPENCODEX_HOME` 에서 동시에 부릅니다. 클레임 헬퍼만 단언하면 `main-account.ts` 가 안 잡아도 통과합니다. 겹침은 enter/leave 순서로 관측합니다. types.ts/config.ts 분할과는 무관합니다.

이건 자격 증명이 서로 덮어쓰이는 운영 버그라 round-2 열차(#3029/#3008/#3019)보다 체감이 다를 수 있습니다. 파일도 라우팅과 거의 안 겹칩니다.

라인 `src/codex/main-account.ts` waitMs 30000 - 바깥 클레임 대기와 안쪽 `AbortSignal.timeout(30_000)` 이 겹칩니다. 클레임을 29초 기다린 뒤 안쪽 시그널이 이미 만료에 가깝면, 락을 얻고도 바로 취소될 수 있습니다. 두 예산을 하나의 기한으로 묶을지는 판단입니다.
라인 `tests/codex-main-account-refresh.test.ts` 두 번째 await - `second.catch(() => null)` 이 실패를 삼킵니다. 직렬화 순서만 보고 두 번째 새로고침이 실제로 성공했는지는 안 봅니다. 덮어쓰기 회귀를 막으려면 두 번째 결과/파일을 더 보는 편이 셉니다.
경로 테스트 OPENCODEX_HOME 스왑 - 한 프로세스에서 env를 바꿔 가며 두 호출을 겹칩니다. 프로덕션은 프로세스 두 개입니다. 클레임 파일이 `CODEX_HOME` 에만 있으면 이 픽스처로도 범위는 증명됩니다. 다만 env 스왑이 다른 병렬 테스트와 간섭하지 않는지 스위트에서만 보면 됩니다.
경로 #3000 - FFI 출판 재작성은 이번 범위 밖으로 둔 판단이 맞습니다. leftover로 #3000을 닫을지는 머지 후 별도입니다. 이 PR은 #2999를 닫습니다.

메인테이너의 판단이 필요한 지점
- 클레임 대기 30초와 안쪽 refresh timeout 30초를 하나의 기한으로 줄일지
- #3000의 atomic replace를 후속으로 남길지, 이 락만으로 #2999를 끝낼지
- 테스트가 두 번째 새로고침 성공까지 단언해야 하는지

너의 추천
머지 쪽으로 진행하는 것을 추천합니다. 지금 HEAD는 홈이 다른 프록시가 같은 `auth.json` 을 동시에 돌릴 수 있고, 이 PR은 이미 있는 CODEX_HOME 클레임을 새로고침에도 씁니다. FFI를 안 가져온 선택은 맞습니다. 테스트가 순서를 관측하니 핵심 회귀는 잡힙니다. Protect dev 리뷰 후 랜딩하면 됩니다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
Requesting changes on exact head `1ade87086e52c60225a02ac521f5f978b9de34e3`.

The central fix is correct: a refresh that publishes shared `CODEX_HOME/auth.json` must hold the existing CODEX_HOME native-main exclusive claim before the per-grant OPENCODEX_HOME lock. The current follow-up error/cancellation contract still has three blockers on this credential path.

1. `resolveMainAccountToken()` starts one 30-second signal before claim acquisition and reuses it inside `withCodexRefreshFileLock` and the token request. A contender that waits most of the claim budget can acquire the claim legitimately and then have the callback rejected immediately by the already-expired signal. Give claim acquisition and the actual refresh separate bounded timeout budgets, while composing the caller's cancellation signal into both.

2. `resolveCodexAuthContext()` now gates `markAccountNeedsReauth()` with `!options.signal?.aborted` at both catch sites. That ambient flag is too broad: a definitive revoked/expired credential error can win the race, then the request signal aborts before the catch runs, and the dead credential remains eligible for repeated failures. Skip quarantine only when the **cause** is abort-shaped or explicitly transient (including a cause equal to the signal's abort reason); definitive credential failures must still mark reauth even if the request is also aborted. Add the negative race regression requested by the current inline review.

3. Transient native-main claim contention/timeouts map to 503, but `resolveResponsesCodexAuth` still logs every `CodexAuthContextError` as “reauthentication required,” and the new parity test explicitly locks in that false message. Do not tell operators to reauthenticate a healthy credential because a lock was busy. Classify the cause before logging: transient/claim/cancel failures need a temporary-retry diagnostic (or no reauth log), while terminal auth failures keep the reauth wording.

The branch is 21 commits behind `dev`, and its macOS matrix is red on the unrelated WebSocket refresh fixture at `tests/server-auth.test.ts:2288`. After these credential-contract fixes and the 2.40.0 version bump, rebase and require a completely green exact-head matrix plus a fresh security review.

lidge-jun:
Rebased onto current `dev` (`b14b741dc`). Head is now `f3c4e9f75`; all four commits preserved, content-identical:

```
1: a6014b968 = 1: 91e6202b2 fix(codex): serialize native-main refresh on the CODEX_HOME claim
2: 01b4ea3cd = 2: 85cccdfc7 fix(codex): abort contended native-main refresh claims
3: 0c70407e2 = 3: 118f9a857 fix(codex): preserve native-main claim transient contracts
4: 1ade87086 = 4: f3c4e9f75 fix(codex): honor websocket abort state for reauth
```

**This is maintenance only.** The three blockers on the credential path are untouched and still stand:

1. `resolveMainAccountToken()` starting one 30s signal before claim acquisition and reusing it inside `withCodexRefreshFileLock` and the token request, so a contender that legitimately wins the claim can be rejected by an already-expired signal.
2. `resolveCodexAuthContext()` gating `markAccountNeedsReauth()` on the ambient `!options.signal?.aborted`, which lets a definitive revoked-credential error lose the race and leave a dead credential eligible.
3. Transient claim contention mapping to 503 while still logging "reauthentication required".

Those need the cause-classification work, not a rebase, and this being a credential path it also needs a fresh security review before landing.

The rebase was still worth doing because the review required "a completely green exact-head matrix", and the previous red was an unrelated flake. Which brings a correction: that flake is **not** fixed. #3128 is an ancestor of the current `dev` and the same assertion still fired on #3133's first run, passing on a rerun of the identical head. #3128 pinned the account namespace; the real race is that `startServer(0)` runs before `Date.now` is pinned while the credential has only 60s of margin over `REFRESH_SKEW_MS` (`src/codex/account-store.ts:22,717`). So a single red on `websocket passthrough refreshes pool auth` here is still noise — rerun rather than treating it as this PR's.

lidge-jun:
Landed via maintainer rebase #3183. All four commits cherry-picked onto current dev with author credit preserved, no conflicts.

The CHANGES_REQUESTED state came from bot reviews against a6014b9688, four commits behind the head. The one substantive finding — the P2 asking that claim waiting honor the refresh abort signal — was already resolved on this branch by 'abort contended native-main refresh claims': main-account.ts passes { waitMs: 30_000, signal } into the claim wait, with signal being AbortSignal.any of the caller's abort and the refresh timeout.

Scope recorded on the carry: this is the lock-scope half of #2999, not the whole issue. The publication/overwrite race stays with the existing refuse-not-overwrite check, so #2999 remains open with that scope written down rather than being closed by association.

PR #3122, opened 2026-09-01
fix(openai): allow canonical fake-IP addresses on provider PATCH

Flowershangfromthebranches:
## Summary

Canonical OpenAI already allows Clash/Mihomo-style `198.18.0.0/15` fake-IP DNS results during provider creation and re-enable.

The ordinary field-mask provider PATCH path did not pass the same `allowBenchmarkAddresses` exception.

As a result, the exact same canonical OpenAI provider could be created successfully but later reject a context-window or other legal provider PATCH with a benchmark-address destination-policy error.

This PR applies the existing canonical OpenAI fake-IP exception consistently to the PATCH path.

## Security

The exception remains limited to:

- the exact canonical built-in OpenAI forward provider
- benchmark/fake-IP addresses already recognized by the existing policy

This does **not** enable `allowPrivateNetwork`.

The existing SSRF protections continue rejecting:

- loopback
- RFC1918
- metadata addresses
- dangerous mixed DNS answers
- custom or OpenAI-like non-canonical providers

## Regression coverage

- canonical OpenAI PATCH passes `allowBenchmarkAddresses: true`
- custom and OpenAI-like providers receive no benchmark exception
- loopback, RFC1918, metadata, and mixed DNS answers remain rejected
- literal benchmark URLs remain rejected
- the exception remains DNS-only and canonical-provider-only

## Verification

- `bun test tests/destination-policy-resolved.test.ts` — 45 passed, 0 failed
- `bun test tests/management-provider-validation.test.ts --test-name-pattern "canonical OpenAI PATCH passes allowBenchmarkAddresses|PATCH destination benchmark exception stays scoped|canonical OpenAI PATCH still rejects non-benchmark private destination answers"` — 3 passed, 0 failed
- `git diff --check upstream/dev...HEAD` — passed

## Review readiness checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated provider edits to correctly support Clash/Mihomo fake-IP DNS responses for the built-in OpenAI provider.
  - Continued blocking unsafe destinations, including loopback, private-network, metadata, and mixed-risk addresses.
  - Custom and OpenAI-lookalike providers remain subject to strict destination validation.

- **Tests**
  - Added coverage for valid fake-IP responses and rejected unsafe destinations during provider updates.

### DIFF
--- a/src/server/management/provider-routes.ts
+++ b/src/server/management/provider-routes.ts
@@ -752,7 +752,13 @@ export async function handleProviderRoutes(ctx: ManagementContext): Promise<Resp
       if (!canonicalBudgetOnly) {
         const serviceTierError = providerServiceTierConfigError(name, next);
         if (serviceTierError) return jsonResponse({ error: serviceTierError }, 400);
-        const resolvedError = await providerDestinationResolvedError(name, next);
+        // Same DNS gate as POST and re-enable: the canonical built-in OpenAI forward
+        // provider may resolve through Clash/Mihomo fake-IP DNS (198.18.0.0/15), so the
+        // ordinary PATCH must not reject the very same destination the provider was
+        // created with. Loopback, RFC1918, metadata, and mixed dangerous answers still
+        // fail closed; nothing else gains the exception.
+        const allowBenchmarkAddresses = name === "openai" && isCanonicalOpenAiForwardProvider(next);
+        const resolvedError = await providerDestinationResolvedError(name, next, { allowBenchmarkAddresses });
         if (resolvedError) return jsonResponse({ error: resolvedError }, 400);
       }
     } else if (applied.enablingOpenAi) {
--- a/tests/management-provider-validation.test.ts
+++ b/tests/management-provider-validation.test.ts
@@ -2462,6 +2462,149 @@ describe("provider management validation", () => {
     }
   });
 
+  test("canonical OpenAI PATCH passes allowBenchmarkAddresses into destination resolution", async () => {
+    // The ordinary field-mask PATCH resolves the SAME canonical chatgpt.com destination
+    // POST and re-enable already admit. Without the benchmark opt-in here, a Clash/Mihomo
+    // fake-IP user (chatgpt.com → 198.18.0.0/15) could create the provider but could never
+    // patch a context overlay onto it. Loopback/RFC1918/metadata and mixed dangerous
+    // answers still fail closed (covered by destination-policy-resolved tests and below).
+    if (existsSync(TEST_DIR)) rmSync(TEST_DIR, { recursive: true });
+    mkdirSync(TEST_DIR, { recursive: true });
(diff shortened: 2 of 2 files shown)


Ingwannu (changes requested):
Reviewed exact head f463e124d37f1ebef4db476a75827fe95dfc3879 against current dev. The production change remains narrowly scoped: only the canonical OpenAI row can opt the PATCH destination check into 198.18.0.0/15 handling, while custom rows and the loopback, RFC1918, metadata, and mixed-answer guards remain fail-closed. The focused negative coverage matches that boundary.

I cannot approve this exact head yet because Cross-platform CI is not green: Linux test 3/4 failed, and this head is now five commits behind dev. Current dev also contains #3128, which isolates the websocket-refresh fixture involved in the release-train flake, plus the 2.40.0 release-line updates. Please rebase this unchanged patch onto current dev and rerun the full exact-head matrix. If the rebased diff stays equivalent and all required checks are green, this remains a strong approval candidate; no broader product change is requested.

lidge-jun:
Landed on `dev` as `0dc01cdaa` via #3133, which carried your commit `f463e124` unchanged (`git cherry-pick -x`, authorship preserved on the branch). Thank you — the diagnosis was exactly right, and the fix is the one-line asymmetry it looked like.

**Why it was carried instead of merged here.** You have `read` permission on this repository, and `.github/workflows/enforce-pr-target.yml:740-746` applies the contributor readiness checklist to authors without push permission. A maintainer force-push to rebase your branch would have re-drafted this PR and reset four boxes only you can tick, so the PR would have sat in draft waiting on you after the work was already done. Carrying it onto a maintainer branch keeps your commit and satisfies the gate.

**Security review.** Because this touches a caller of `src/lib/destination-policy.ts`, it went through explicit security review rather than an ordinary read. Verdict was PASS with zero blockers. The question worth asking was whether `next` is a partial field mask — if it were, a PATCH could set a subset of fields so `isCanonicalOpenAiForwardProvider(next)` returns true while the effective stored provider is not canonical. It is not: `applyProviderPatchFields` opens with `const next: OcxProviderConfig = { ...provider }` (`provider-routes.ts:114-121`), and the route's own comment at `:737-739` says it validates "the MERGED provider". Your placement was correct.

**Verification on the carry:**

```
bun test tests/management-provider-validation.test.ts tests/destination-policy-resolved.test.ts
  129 pass / 0 fail / 635 expect()
```

Your tests catch both mutations, which is what makes them a guard rather than a description of the flag: hardcoding it `true` fails `PATCH destination benchmark exception stays scoped to the canonical openai row`, and hardcoding it `false` fails the other two. That is the reason this was landable without further work.

One note for anyone reading later: the first CI run on the carry was red on `server local API auth > websocket passthrough refreshes pool auth for each response.create turn`. That is an unrelated pre-existing flake, not this change — a rerun of the identical head passed, and the test passes locally.

Closing as merged.

PR #3013, opened 2026-08-30
fix(release): repair dev version line and open future bumps as PRs

lidge-jun:
## Summary

`dev` carried `2.36.0` while tag `v2.36.0` names `c7d8407d2` on `main`, so the tree claimed an already-published version from a different commit. `tests/release-version-line.test.ts` reported exactly that, and it failed `test 2/4` and `macos` on `dev` itself (run 33312566315) and therefore on every PR opened against it — PR #3007 inherited both reds for a two-file GUI change.

Two parts: the immediate repair, and the actor that stops it recurring.

**The repair.** `package.json` `2.36.0` -> `2.37.0`. Freeness verified live rather than assumed: no `v2.37*` tag, `npm view @bitkyc08/opencodex@2.37.0` is E404, `gh release view v2.37.0` is not found. Note the highest tag is `v2.36.0`, not the later-dated `v2.36.0-preview.20260830` — sorting all 218 tags with the repository's own `compareReleaseTags` puts a stable release above its own prerelease, which is why the failure message names `v2.36.0`.

**Why a bump alone is not the fix.** The same defect has been repaired by hand four times:

| commit | change |
|---|---|
| `32529c2b2` | `2.24.2` -> `2.27.0` |
| `e4a85d134` | `2.32.1-preview.20260825` -> `2.34.0`, and ADDED the detector |
| `076ad3036` | `2.34.0` -> `2.35.0` |
| `befcac3e1` | `2.35.0` -> `2.36.0` |

The detector was added *during* that sequence and two more repairs followed it, so more visibility was never the missing piece. The cause is structural: `scripts/release.ts` runs only on `main`/`preview` (`allowedBranches`, line 496), bumps `package.json` there, and pushes that branch; `release.yml` ends at "Create GitHub release". Nothing advances `dev`.

**The actor.** `.github/workflows/dev-version-bump.yml` opens the bump as a pull request on `release: published`, and `scripts/bump-dev-version.ts` decides the version. The script is pure — no git, no network — which is what makes it unit-testable and keeps the credentials in the workflow.

[...]

PR #3127, opened 2026-09-01
chore(release): move dev to 2.40.0 after the v2.39.0 release

lidge-jun:
## Summary

Moves `dev` to **2.40.0** after the v2.39.0 release. One line in `package.json`, nothing else.

Without this, `dev` carries `2.39.0` — a version that is now published — and `tests/release-version-line.test.ts` fails on `dev` and on **every pull request opened against it**. That is inherited red a contributor cannot fix from their own diff.

## Why 2.40.0

Decided by `scripts/bump-dev-version.ts`, not by hand:

```
$ bun scripts/bump-dev-version.ts v2.39.0 package.json
{"changed":true,"version":"2.40.0","reason":"v2.39.0 is a stable release, so dev moves to the next minor 2.40.0"}
```

The rule is about the published version's shape: a stable `X.Y.Z` means that core is consumed, so `dev` moves to `X.(Y+1).0`. (A published prerelease would instead mean `dev` takes the stable core it belongs to.)

## Why this is a hand-opened PR

`.github/workflows/dev-version-bump.yml` exists to open exactly this PR on `release: published`, and it did not fire — the workflow has **0 runs** since it landed in #3013.

The cause is not a broken workflow. `release.yml` creates the GitHub release with `GH_TOKEN: ${{ github.token }}` ([release.yml:380](https://github.com/lidge-jun/opencodex/blob/main/.github/workflows/release.yml#L380)), and GitHub does not start new workflow runs from events raised by the default `GITHUB_TOKEN`. So a `release: published` trigger can never observe a release this repository publishes itself. The workflow's own header anticipates a missed run and names this exact recovery path:

> Re-drive a missed run by running `bun scripts/bump-dev-version.ts <released> package.json` locally and opening the pull request normally.

That is what this PR is. Worth noting for a follow-up: the same reasoning applies to v2.37.0 and v2.38.0, whose bumps (#3045, #3076) were also opened by hand — so the automation has never actually fired, and the four historical hand-repairs it was written to end are still ongoing. Making it work needs either a PAT/app token on the release step or a different trigger; that is a separate change and out of scope here.

## Verification

The freeness gate the workflow would have run, executed locally on this branch:

```
$ bun test tests/release-version-line.test.ts
 3 pass
 0 fail
```

`2.40.0` is unused: `npm view @bitkyc08/opencodex@2.40.0` is a 404, and `git ls-remote --tags origin 'refs/tags/v2.40.0*'` is empty.

`git diff origin/dev --name-only` is `package.json` alone.

## Checklist

- [x] Version chosen by the script, not by hand
- [x] Release-version-line gate green on this branch
- [x] Target version unused on npm and in the tag set
- [x] Single-file diff

## Summary by CodeRabbit

* **Chores**
  * Updated the application to version 2.40.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bitkyc08/opencodex",
-  "version": "2.39.0",
+  "version": "2.40.0",
   "description": "Universal provider proxy for OpenAI Codex & Claude Code — use any LLM with Codex CLI/App/SDK and Claude Code",
   "type": "module",
   "main": "./bin/package-main.mjs",


Ingwannu (approved):
Approved on exact head `8eeae1dc8205ca6e060cb314b43bc15d600fafef`.

The release boundary is consistent and the diff is exactly one version line: `origin/main`, `origin/dev`, npm, and tag `v2.39.0` all identify the published 2.39.0 release, while npm, Git tags, and GitHub Releases have no 2.40.0 artifact. `decideDevVersion()` maps a stable `2.39.0` release to the next minor `2.40.0`, so this restores a non-published development version without rewriting any release metadata.

This approval does not waive the exact-head gate. Merge only after every currently-running Cross-platform CI shard, macOS job, and npm-global Windows job completes successfully on this SHA.

PR #3132, opened 2026-09-01
fix(minimax): align split-reasoning wire schema with the official reasoning_details contract

lidge-jun:
## Summary

- MiniMax M-series with `reasoning_split: true` returns thinking as a structured `reasoning_details` array (`{type, id, format, index, text}`), and the official interleaved-thinking guide requires that array back verbatim on the next turn. The openai-chat adapter dropped `reasoning_details` entirely and replayed a `reasoning_content` string — the native-format pass-back MiniMax documents as unsupported — so streamed thinking never surfaced and tool-use continuations lost the reasoning chain.
- Adds a `reasoningDetailsModels` registry knob (plumbed through `derive`/`router` exactly like `reasoningSplitModels`), prefix-diffs the cumulative `reasoning_details` stream deltas (with an incremental-sender fallback), reads the array as a non-stream fallback, and serializes preserved reasoning as a single `reasoning.text` segment (`format: MiniMax-response-v1`) for listed models. Both `minimax` and `minimax-cn` opt in.
- Evidence: https://platform.minimax.io/docs/guides/text-m3-function-call and https://platform.minimax.io/docs/api-reference/text-openai-api (verified 2026-09-01 via the official docs; the SDK streaming example prefix-diffs `delta.reasoning_details[].text`).

## Verification

- `bun run typecheck`
- Focused regression coverage added in `tests/minimax-reasoning-split.test.ts` (replay shape, non-stream `reasoning_details` fallback, cumulative stream prefix-diff, opt-out provider unchanged) and `tests/provider-registry-parity.test.ts`; not executed locally at the author's request — CI runs the suite.
- No GUI change.

## Checklist

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

- **New Features**
  - Added support for providers that return model reasoning as structured `reasoning_details` arrays.
  - Added the `reasoningDetailsModels` configuration option for identifying compatible models.
  - Improved streaming handling for cumulative reasoning updates and preserved reasoning replay.
  - Enabled built-in support for MiniMax models using split reasoning.

- **Documentation**
  - Updated provider configuration and adapter references across supported languages.

- **Tests**
  - Added coverage for streaming, response parsing, routing, and reasoning replay behavior.


lidge-jun:
Admin merge bypass recorded: required CI is green on exact head `ac781745265a5a2e15a424dde86efc8db1a06be7` (Cross-platform CI run 33474986946, including macOS after a failed-job rerun of the known websocket auth flake). Author-owned PR, so this uses the owner `pull_request` ruleset bypass rather than a self-approval. Squash-merging onto current `dev`.

PR #3134, opened 2026-09-01
fix(service): Windows cold-start budget and code-page-mangled scheduler paths (rebase of #3104)

### LINKED ISSUE #3009: [Bug][Windows]: service repair can fail at 20s just before the proxy becomes healthy

### LINKED ISSUE #3064: Windows: `ocx service install` rolls back with a non-ASCII profile path

lidge-jun:
## Summary

Rebases #3104 onto current `dev`. Same seven commits, same author, verified content-identical — see Verification. Supersedes #3104, which was approved on a head that is now 27 commits stale and red on a flake that has since been fixed.

Closes #3009. Closes #3064.

Two issues, one file (`src/service.ts`), stacked because a conflicting pair would have cost more to review than one sequence.

**#3009 — Windows cold start.** `confirmServiceServing` had a fixed 20s deadline and returned the moment the clock passed it. A Windows cold start does NTFS ACL hardening and previous-session journal recovery before the listener exists, so a service that bound a few seconds late and then stayed healthy was reported as a terminal failure with exit 1 — and the caller's fallback starts a second proxy against a port that is about to be taken. Windows gets 45s; nothing else changes.

**#3064 — non-ASCII profile path.** `schtasks /query /xml` converts through the console code page before the bytes exist, so reading as a buffer cannot recover them. A profile named outside that page comes back as `C:\Users\???\...` and the exact comparison rejected a registration this process had just created. An unrepresentable run now matches only a run of substitution characters, and every ASCII segment including every separator is matched literally, so `C:\Users\<CJK>\...` cannot match `C:\Users\Admin\...`.

## Why the original was red, and why this is not

#3104's `macos` job failed one assertion: `server local API auth > websocket passthrough refreshes pool auth for each response.create turn`. That is the cross-platform flake fixed by #3128 (`33d32b6a3`), which landed after #3104's head was cut. Rebasing onto current `dev` picks the fix up; nothing in this change touches that path.

## Verification

```
git range-diff 9d122ddb..4f691cdd origin/dev..HEAD
  1: 7b1b4ce1f = 1: 3b7cc1f00 fix(service): give a Windows cold start room to bind, without loosening zero budget
  2: b727f8f81 = 2: 68a8fb6a4 fix(service): forgive only what the code page mangled in a scheduler path
  3: 188e61861 = 3: 089598878 fix(service): bind scheduler recovery to exact SID
  4: 4d83513b4 = 4: 183e6dfaa fix(service): fail closed on ambiguous scheduler ownership
  5: 97efe6f46 = 5: adda7a574 test(service): lock scheduler ownership guards
  6: b29389a88 = 6: ed6fe19db test(service): scope scheduler verification fixtures
  7: 4f691cddc = 7: 5ea32ad00 test(service): exercise scheduler ownership oracles
```

All seven `=`: content identical, only the base moved. `src/service.ts` was the sole file changed on both sides since the merge base (dev-side commit `330470e74`), and it merged without conflict.

```
bun test tests/service.test.ts
  191 pass / 0 fail / 617 expect() calls
```

## On the contributor PRs this supersedes

#3067 (@ntdatt812) locates the #3064 defect correctly and is superseded here — its remedy needed narrowing, because compiling every unrepresentable run to `[^\\/]*` forbids a separator but allows arbitrary ASCII, so a wholly non-ASCII segment loses every anchor and one account could adopt another's scheduler task.

**#3039 (@ntdatt812) is deliberately NOT closed by this PR.** It contributed a diagnostic this rebase does not carry: it prints the time actually waited (`Math.round((elapsed() - startedAt) / 1000)`), where this prints the configured budget (`Math.trunc(healthBudgetMs / 1000)`). Since this PR also adds a post-deadline grace knock, the printed number can understate the real wait — which is the defect #3039 set out to fix. That contribution deserves its own resolution rather than a silent supersession.

## Checklist

- [x] Targets `dev`
- [x] Focused regression test for the changed subsystem passes (`tests/service.test.ts`)
- [x] Rebase verified content-identical by `git range-diff`
- [x] Original authorship preserved across all seven commits

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows service installation health checks with platform-specific startup timing and an additional recovery probe.
  * Strengthened Windows Task Scheduler registration validation using account identity information, including automatic migration of older registrations.
  * Improved compatibility with non-ASCII Windows profile paths.
  * Prevented unsafe competing service starts after a repair timeout.

* **Improvements**
  * Windows service repairs now allow more time to complete and provide manual recovery guidance when the outcome is unknown.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -661,6 +661,23 @@ export function installedServiceListenPort(): number {
 export const SERVICE_INSTALL_HEALTH_MS = 20_000;
 
 /**
+ * Windows gets a longer budget because its cold start does more before the
+ * listener exists: NTFS ACL hardening and previous-session journal recovery
+ * both run first, and #3009 recorded a service that bound a few seconds past
+ * the 20s deadline and then stayed healthy. Reporting that as a terminal
+ * repair failure is worse than waiting — the caller's fallback is to start a
+ * second proxy against a port that is about to be taken.
+ */
+export const SERVICE_INSTALL_HEALTH_WINDOWS_MS = 45_000;
+
+/** The health budget for the platform this is running on. */
+export function serviceInstallHealthMs(
+  platform: NodeJS.Platform = process.platform,
+): number {
+  return platform === "win32" ? SERVICE_INSTALL_HEALTH_WINDOWS_MS : SERVICE_INSTALL_HEALTH_MS;
+}
+
+/**
  * Whether a proxy actually answers on the port this install/start just produced.
  *
  * Registration is not service: `launchctl list` reports a job that never bound, and
@@ -690,12 +707,23 @@ export async function confirmServiceServing(
   const now = deps.now ?? Date.now;
   const sleep = deps.sleep ?? ((ms: number) => new Promise<void>(r => setTimeout(r, ms)));
   const probe = deps.probe ?? (async (p, h) => !!(await proxyIdentityAt(p, { hostname: h })));
-  const deadline = now() + (deps.timeoutMs ?? SERVICE_INSTALL_HEALTH_MS);
+  const deadline = now() + (deps.timeoutMs ?? serviceInstallHealthMs());
+  let waited = false;
   for (;;) {
     if (await probe(port, hostname)) return { ok: true, port };
-    if (now() >= deadline) return { ok: false, port };
+    if (now() >= deadline) break;
     await sleep(500);
+    waited = true;
   }
+  // The probe that ran last started before the deadline, so a service that binds
+  // during it is reported as dead (#3009). Knock once more after a short grace
+  // before calling it a failure. A zero budget means the caller asked not to
+  // wait, so it gets exactly the single probe it asked for and nothing more.
+  if (waited) {
+    await sleep(500);
+    if (await probe(port, hostname)) return { ok: true, port };
+  }
+  return { ok: false, port };
 }
 
 /**
--- a/src/service.ts
+++ b/src/service.ts
@@ -1963,6 +1963,67 @@ function taskXmlDecodedValueEquals(xml: string, tag: string, expected: string): 
   return taskXmlDecodeEntities(value).trim().toLowerCase() === expected.trim().toLowerCase();
 }
 
+/**
+ * Characters a console code page substitutes when it cannot carry the original.
+ * Windows writes `?` per unrepresentable character, some layers write U+FFFD, and a
+ * few drop them entirely.
+ */
+const CODE_PAGE_SUBSTITUTIONS = /^[?\uFFFD]*$/;
+
+/**
+ * Compare a value that OpenCodex itself wrote against what `schtasks /query /xml` read
+ * back, tolerating ONLY the characters the console code page could not carry.
+ *
+ * `runFile` already reads the query as bytes, so this is not a spawn-decoding bug: the
+ * conversion happens inside `schtasks` before the bytes exist. A profile named outside
+ * the active code page — `C:\\Users\\김병준\\...` — comes back as `C:\\Users\\???\\...`, so an
+ * exact comparison rejected a registration this process had just created correctly and
+ * `ocx service install` rolled it back (#3064).
+ *
+ * The tolerance is deliberately narrow. Each unrepresentable RUN in the expected value
+ * may match only a run of substitution characters — never arbitrary text, and never a
+ * path separator. A wildcard as wide as `[^\\\\/]*` would leave a fully non-ASCII segment with
+ * no anchors at all, so `C:\\Users\\김병준\\x.vbs` would match `C:\\Users\\Admin\\x.vbs` and this
+ * process would adopt, repair, or delete another account's task. Accepting a foreign
+ * live task is a worse failure than the rollback this fixes.
+ */
+function taskXmlLossyValueEquals(reported: string, expected: string): boolean {
+  const a = reported.trim().toLowerCase();
+  const b = expected.trim().toLowerCase();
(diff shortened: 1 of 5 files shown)


ntdatt812:
Sent as #3138.

You were right that the diagnostic was a separate contribution and right about which way it is wrong. Measured on the merged code: with the Windows budget the loop exits at 45s and the grace knock adds its 500ms sleep, so the run spends 45.5s and the line says `within 45s`.

I timed it in `reportServiceServing` through the `now` the deps already carry rather than widening `confirmServiceServing`'s return, so its contract and its tests — including the zero-budget single-probe assertion #3104 was right to restore — are untouched. Nine lines, one file.

Two tests, and the second is the one worth naming: a zero budget must report `after 0s`, because "print the elapsed time" has an obvious wrong version that prints the budget when nothing was waited, and the zero-budget path is where that shows.

Thanks for splitting it out rather than closing #3039 over it, and for saying which parts of each source branch held up. The #3067 finding in particular — that my fixture `Người` kept ASCII anchors and hid the wholly-non-ASCII case — was worth more to me than the merge would have been.

PR #3042, opened 2026-08-31
test: probe for a free pid instead of assuming 4242 is dead

lifrary:
## Summary

Nine sites across three suites stand in for an exited process with a hardcoded pid:

```ts
const deadPid = process.pid === 4242 ? 4243 : 4242;
```

The code under test asks the kernel whether that owner is still alive, so the pid is only dead until an unrelated process happens to hold it. Production then answers correctly, the test reads that as a miss, and the failure looks like a defect in the feature rather than in the fixture.

That is not hypothetical. On the macOS host where this was found, pid 4242 is `liveactivitiesd`, and five tests fail together on a clean `dev` checkout:

- `Responses previous_response_id state > periodic reclaim frees abandoned temps without any continuation access`
- `doctor reclaim wiring (end to end) > the default run reports the file and leaves it on disk`
- `doctor reclaim wiring (end to end) > the opt-in flag removes it`
- `status reports stale process records end to end > a dead owner record surfaces in --json and in human output`
- `status reports stale process records end to end > a fallback-port record is judged on the recorded port, not the configured one`

**The probe already existed.** `tests/responses-state.test.ts` did it inline for one test, with a comment naming this exact hazard on a shared CI runner, while four sites in that same file and four more in `tests/cli-status-json.test.ts` and `tests/doctor.test.ts` kept the assumption. This lifts that probe into `tests/helpers/dead-pid.ts` and uses it at every site, so the knowledge lives in one place rather than in a comment beside one of nine copies.

The helper throws rather than returning a sentinel. The inline version needed `expect(deadPid).toBeGreaterThan(0)` at its call site; a throw gives every caller that guarantee without repeating the assertion. ESRCH is the only accepted answer — a successful `kill(pid, 0)` means alive, and EPERM means alive but owned by somebody else.

Other `4242` literals in the suite (`proxy-liveness`, `local-management-capability`, `server-management-auth`, `process-control-graceful`, `codex-app-server-processes`) are injected fixture data read through mocked accessors, never probed against the kernel, and are left alone.

No production code changes.

## Verification

macOS, branch on `dev` @ `4180067b`. Pid 4242 was held by `liveactivitiesd` throughout, so the before/after pair is a live positive control rather than a constructed one:

- The three affected files, stock `dev`: **186 pass / 5 fail**.
- The three affected files, with this change: **191 pass / 0 fail**.
- `bun run typecheck` — clean, exit 0.
- `bun run test:changed` — 191 pass, 0 fail across 3 files.
[...]

PR #3137, opened 2026-09-01
test: probe for a free pid instead of assuming 4242 is dead (carry of #3042)

lidge-jun:
## Summary

Carries #3042 (author @lifrary) onto current `dev`. The commit is the author's, cherry-picked with `-x`; authorship is preserved.

Closes #3042 once this lands.

Nine sites across three suites stood in for an exited process with a hardcoded pid:

```ts
const deadPid = process.pid === 4242 ? 4243 : 4242;
```

The code under test asks the kernel whether that owner is still alive, so the pid is only dead until an unrelated process happens to hold it. Production then answers correctly, the test reads that as a miss, and the failure looks like a defect in the feature rather than in the fixture.

That is not hypothetical. The helper's own comment records the day it happened: on the author's macOS host pid 4242 was `liveactivitiesd`, and every suite assuming it dead failed at once — `periodic reclaim frees abandoned temps`, both `doctor reclaim wiring` cases, and both `status reports stale process records` cases.

## What it does

Adds `tests/helpers/dead-pid.ts` and routes the three suites through it. The discrimination that matters is in the catch:

```ts
try {
  process.kill(candidate, 0);
} catch (error) {
  if ((error as NodeJS.ErrnoException).code === "ESRCH") return candidate;
}
```

`ESRCH` is the only answer that proves absence. A successful `kill(pid, 0)` means alive, and `EPERM` means alive but owned by somebody else — a naive probe reads that second one as free, which would reintroduce the same bug in a new form. It also skips `process.pid` and throws rather than returning a wrong answer if the whole window is occupied.

This is not a new abstraction: `tests/responses-state.test.ts` already probed for a free pid inline, with a comment naming this exact hazard. The helper is that probe, shared instead of copied — which is why that file is a net deletion.

## Why it is carried rather than rebased in place

@lifrary has `read` permission, and `.github/workflows/enforce-pr-target.yml:740-746` applies the contributor readiness checklist to authors without push permission. A maintainer force-push would re-draft the PR and reset four boxes only the author can tick, stranding it in draft after the work was done.

## Verification

```
bun test tests/responses-state.test.ts tests/doctor.test.ts tests/cli-status-json.test.ts
  214 pass / 0 fail / 686 expect() calls
```

The cherry-pick applied without conflict despite 59 commits of distance; the only file changed on both sides since the merge base was `tests/responses-state.test.ts`, and the dev-side additions sit elsewhere in it.

## Checklist

- [x] Targets `dev`
- [x] Test-only; no production file in the diff
- [x] The three affected suites pass on the carry
- [x] Original authorship preserved

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability by dynamically selecting confirmed inactive process IDs.
  * Strengthened coverage for stale-process, fallback-port, and snapshot recovery scenarios.
  * Reduced dependence on fixed process ID values and environment-specific conditions.

### DIFF
--- a/tests/cli-status-json.test.ts
+++ b/tests/cli-status-json.test.ts
@@ -7,6 +7,7 @@ import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
 import { isConnectionRefused, isUncleanExitEvidence, proxyHealthFailureReason, resolveStatusPid, selectListenTarget } from "../src/cli/status";
+import { findDeadPid } from "./helpers/dead-pid";
 
 const repoRoot = dirname(fileURLToPath(new URL("../package.json", import.meta.url)));
 const cliPath = join(repoRoot, "src", "cli", "index.ts");
@@ -431,7 +432,7 @@ describe("unclean prior exit evidence", () => {
 describe("status reports stale process records end to end", () => {
   const seed = (home: string, opts: { pid?: number; runtime?: boolean; port: number }): void => {
     writeFileSync(join(home, "config.json"), JSON.stringify({ port: opts.port, codexAutoStart: false }), "utf8");
-    const pid = opts.pid ?? (process.pid === 4242 ? 4243 : 4242);
+    const pid = opts.pid ?? findDeadPid();
     if (opts.pid !== 0) writeFileSync(join(home, "ocx.pid"), String(pid), "utf8");
     if (opts.runtime) {
       writeFileSync(join(home, "runtime-port.json"), JSON.stringify({ pid, port: opts.port, hostname: "127.0.0.1" }), "utf8");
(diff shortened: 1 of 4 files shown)


Ingwannu (approved):
Approved exact head e18566b8cc572be7e99bac6a2fa3e541575ad4da. The range-diff maps the original #3042 test commit 1:1; the only commit-message delta is the cherry-pick provenance. The helper treats only ESRCH as dead, keeps success and EPERM as live, and replaces all nine kernel-probed hardcoded PID fixtures without touching unrelated mocked PID data or production code. Exact-head required CI is green, including all four test shards, macOS, the Windows keyring smoke, storage policy, gates, hygiene, and React Doctor. Current dev is one unrelated MiniMax fix ahead, so no rebase is needed for this test-only patch.

PR #3121, opened 2026-09-01
fix(openai): exclude user-owned alias overlays from canonical seed validation

Flowershangfromthebranches:
## Summary

Canonical OpenAI provider validation compares the persisted provider row against the built-in transport seed.

OpenCodex's dedicated alias APIs can legally persist:

- `defaultAliases`
- `alias`
- `modelAliases`

After one of those overlays exists, an unrelated general provider edit such as a context-window PATCH could fail with:

`provider openai must equal the canonical built-in provider seed`

This validation deadlock is real, but the general provider write path must not become an unvalidated alias writer in order to resolve it.

The revised fix keeps alias ownership at the dedicated APIs:

- A full provider POST may omit alias fields or round-trip values exactly equal to the stored provider. The latest stored alias overlays are restored before persistence.
- A full provider POST cannot create, clear, or change alias overlays.
- General provider PATCH rejects `alias`, `modelAliases`, and `defaultAliases`; callers must use the dedicated alias endpoints.
- Canonical OpenAI transport validation runs against a candidate from which only already-owned stored overlays were removed.

## Safety

The canonical transport guard remains unchanged and continues rejecting changes to:

- `baseUrl`
- `adapter`
- `authMode`
- other canonical transport identity

The dedicated alias APIs remain the only writers that apply provider/combo/account collision checks, model-alias syntax and uniqueness rules, native-family reservations, and boolean shape validation.

Unrelated non-OpenAI provider edits retain their previous behavior and preserve stored aliases.

## Regression coverage

Tests cover:

- full provider edits preserving existing provider, model, and default aliases;
- provider POST/PATCH being unable to introduce a provider alias collision;
- reserved, duplicate, and invalid model aliases remaining rejected;
- non-boolean `defaultAliases` and malformed alias maps returning bounded 4xx without persistence;
- canonical OpenAI with stored aliases successfully PATCHing `modelContextWindows`;
- `baseUrl`, `adapter`, and `authMode` tampering remaining rejected;
- unrelated non-OpenAI edits remaining unchanged;
- the dedicated alias API's case-insensitive collision rules;
- destination-policy test spies being restored so later tests cannot receive a false green result.

## Verification

Exact publication head: `dd60f05d73599991b12e11298f741d34ed84ccbf`  
Exact upstream baseline: `15b0f701eb69d37382880124be9010e412c20411`  
Environment: Bun `1.4.0`, Node `v23.11.0`, identical local DNS/proxy/env.

[...]

PR #3143, opened 2026-09-01
test(auth): timestamp quota fixtures under fake clock

Ingwannu:
## Summary

- write both affected quota fixtures under the same fake clock used by the startup prime;
- keep the existing fetch stub and production behavior unchanged;
- correct the comments so the timestamp boundary is explicit.

## Why

The follow-up in #3139 moved the fake clock before `startServer`, but `updateAccountQuota` still ran first under the real clock. The asynchronous startup prime then compared a future fake-clock read with a real-clock `updatedAt`, could classify the fresh row as stale, and rotate the credential before the first request.

This reproduced as the macOS assertion failure in #3121 at `tests/server-auth.test.ts:2302`: the first request observed the refreshed credential instead of the old one. #3121 does not modify this test, so the failure is an upstream fixture race rather than a regression in that PR.

## Verification

Exact head: `1e50927f5b8a7ad6b9bca5835cc1e27d1b7f1eed`

- Bun 1.4.0: the two affected `tests/server-auth.test.ts` cases passed 5 consecutive focused runs (10 tests total, 0 failed).
- `bun run typecheck` passed.
- `git diff --check` passed.
- Tests used isolated temporary `HOME`, `OPENCODEX_HOME`, and `CODEX_HOME`; protected local runtime config hashes were unchanged.

## Checklist

- [x] Behavior-changing documentation is not required; this changes test setup only.
- [x] No production auth or credential behavior changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the reliability of pool-auth WebSocket tests by ensuring quota timing uses a consistent clock during asynchronous startup.
  * Removed outdated test commentary describing a previously addressed timing race.

### DIFF
--- a/tests/server-auth.test.ts
+++ b/tests/server-auth.test.ts
@@ -2125,17 +2125,12 @@ describe("server local API auth", () => {
       expiresAt: now + CODEX_THREAD_AFFINITY_IDLE_TTL_MS + 10 * 60_000,
       chatgptAccountId: "acct-pool-a",
     });
-    updateAccountQuota("pool-a", 10, 5);
-
     const originalNow = Date.now;
-    // Pin the clock BEFORE startServer, not after. `startServer` returns synchronously but
-    // arms an async pool-quota prime (src/server/index.ts:2054-2064) that outlives its
-    // return, and that prime decides staleness with `Date.now() - quota.updatedAt >=
-    // POOL_CACHE_TTL` (src/codex/auth-api.ts:1334-1337). `updateAccountQuota` above stamped
-    // `updatedAt` with the REAL clock, so a prime that lands after a 2027 fake clock is
-    // installed sees months of cache age, fetches, and rotates the credential out from under
-    // the assertions. Installing the clock first closes the window entirely.
+    // Pin the clock before both the quota write and startServer. The startup prime compares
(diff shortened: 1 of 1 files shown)


lidge-jun:
Superseded by #3147, which landed on `dev` as `408652698`.

Closing this per your own note on #3147 — you said you would close the duplicate once the owner-authored PR landed, so I am doing the bookkeeping rather than making you come back for it. The diagnosis here was yours and it was right: `updateAccountQuota` ran under the real clock while the startup prime read the pinned fake one, so a fresh row read as months stale and the credential rotated before the first request was served. #3147 carries the same fix, and your review on it also caught the affinity test's missing seed, which is now restored on the correct side of the pin (`0cf5ef7b5`).

One thing worth recording, since it was your framing on #3147 and I disagreed with part of it after tracing the code. The missing seed could not have moved `upstreamRequests` off 3: `redirectCanonicalCodexTo` rewrites only `/backend-api/codex`, while the prime's call is `/backend-api/wham/usage`, so it bypasses the counted upstream entirely. And the account was already stale on `dev` regardless, because `!q || age >= POOL_CACHE_TTL` puts both cases in the same branch. The seed still belonged there — the comment above it was pointing at a call that no longer existed — but as fixture hygiene, not as a race.

Thanks for the catch on both counts.

PR #3147, opened 2026-09-01
test(auth): seed the pool quota after the clock is pinned

lidge-jun:
## Summary

The websocket refresh test in `tests/server-auth.test.ts` still fails on loaded CI runners after #3139, on both macOS and Linux. `dev`'s own HEAD fails it, so this is not specific to any open branch.

`updateAccountQuota` stamps `updatedAt: Date.now()`. The seed ran **before** the clock was pinned, so that stamp is real wall-clock time while everything after it reads the pinned 2027 value. The gap is roughly 136 days against a 6-hour freshness window (`QUOTA_DISK_MAX_AGE_MS`, `src/codex/quota.ts:491`), so the seeded quota reads as stale no matter how fast the runner is. The startup pool-quota prime then refreshes the credential before the first turn is served, and `seenAuth[0]` is already the new token — which is why the failure diff was always the first element and never the second.

#3139 pinned the clock and the fetch stub before `startServer`, which closed the window for the prime's own reads. It could not close a window for a timestamp written before either was in place. Moving the seed after the pin does.

## Verification

- `bun test tests/server-auth.test.ts` — passes.
- The mismatch is timing-dependent and does **not** reproduce locally either before or after this change, so there is no local red-to-green to show. The evidence is the mechanism: a 136-day gap against a 6-hour window is arithmetic, not a race. Exact-head CI is the real check.

## Checklist

- [x] Focused tests run for the changed area
- [x] No production code changed — test-only
- [x] Targets `dev`

## Summary by CodeRabbit

* **Tests**
  * Updated authentication and WebSocket refresh tests to use consistent simulated timestamps.
  * Improved test reliability by preventing stale quota timestamps during time-based scenarios.

### DIFF
--- a/tests/server-auth.test.ts
+++ b/tests/server-auth.test.ts
@@ -2125,8 +2125,6 @@ describe("server local API auth", () => {
       expiresAt: now + CODEX_THREAD_AFFINITY_IDLE_TTL_MS + 10 * 60_000,
       chatgptAccountId: "acct-pool-a",
     });
-    updateAccountQuota("pool-a", 10, 5);
-
     const originalNow = Date.now;
     // Pin the clock BEFORE startServer, not after. `startServer` returns synchronously but
     // arms an async pool-quota prime (src/server/index.ts:2054-2064) that outlives its
@@ -2240,14 +2238,6 @@ describe("server local API auth", () => {
       codexAccountNamespaces: { "ws-refresh": "pool-a" },
       activeCodexAccountId: "pool-a",
     } as OcxConfig);
-    saveCodexAccountCredential("pool-a", {
-      accessToken: "old-access-token",
-      refreshToken: "old-refresh-token",
-      expiresAt: now + 120_000,
-      chatgptAccountId: "acct-pool-a",
-    });
-    updateAccountQuota("pool-a", 10, 5);
-
     const originalNow = Date.now;
     const originalFetch = globalThis.fetch;
     // Both the clock and the fetch stub go up before `startServer`. The async pool-quota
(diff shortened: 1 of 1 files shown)


Ingwannu (changes requested):
Requesting changes on exact head `ecf51c67f89b45e29303b9c5be49678733008c2e`.

The unresolved Codex P1 on this head is valid. The removed `updateAccountQuota("pool-a", 10, 5)` belongs to the preceding `expired thread affinity returns 409...` test. Without that fresh cached quota, `startServer` can run the asynchronous startup prime, send an extra WHAM request through the redirected canonical upstream, and increment `upstreamRequests` beyond the asserted three whenever the prime wins the timing window. The current full matrix being green does not close a race whose failure depends on that ordering.

Restore that first test quota seed immediately after `Date.now = () => now` and before `startServer(0)`. Also update its stale comment, which currently says `updateAccountQuota above` even though no such write remains. The websocket-refresh test can keep both credential and quota seeding after the pinned clock; that part is compatible with the intended fix.

This is the same missing boundary already handled in #3143. Once the first-test seed is restored, rerun the affected file and exact-head matrix. If the new head remains otherwise equivalent and green, this owner-authored PR can supersede #3143 and I will close my duplicate after it lands.

lidge-jun:
Restored on `0cf5ef7b5` — the seed is back in the affinity test, on the far side of the clock pin, and the comment no longer references a call that is not there.

One correction to the reasoning, because I do not want this stored in the history as a race fix when it is not one.

The stated failure mode cannot reach the assertion. `redirectCanonicalCodexTo` (tests/server-auth.test.ts:106-117) rewrites a request only when `hostname === "chatgpt.com"` **and** `pathname` starts with `/backend-api/codex`. The prime's usage call is `/backend-api/wham/usage` (src/codex/auth-api.ts:1157), so it falls through to `originalGlobalFetch` and never touches the local `Bun.serve` that increments `upstreamRequests`. Even a prime that wins every timing race leaves that counter at 3.

The removal was also behaviorally neutral. `primeCodexPoolQuotas` classifies staleness as `!q || Date.now() - q.updatedAt >= POOL_CACHE_TTL` (auth-api.ts:1332-1335). On `dev` the seed ran *before* the pin, so `updatedAt` carried the real clock (~1788268436923) while the test pins `Date.now` to `1_800_000_000_000` — about 136 days of apparent age against a 5-minute TTL. The account was already in the `stale` list. Deleting the seed moved it from false-but-stale to true-and-stale; same branch, same outcome.

Where you were right: the prime genuinely does reach `fetchPoolAccountQuota`, since the credential is seeded and `:1201-1202` does not early-return on a null `existing`. That link is real. It just does not connect to the assertion.

So the seed is restored on its own merits — the comment was lying, and seeding after the pin is the first time the prime is actually suppressed rather than merely harmless. Verified with `bun test tests/server-auth.test.ts`: 91 pass, 0 fail. Full matrix is running on the new head.

Separately, for the merge train: #2789 will not go green on this fix alone. Its macOS job is 17098 pass / 2 fail — the websocket test plus `ocx launcher graceful shutdown > SIGINT to the launcher tears down the Bun proxy` timing out at 20069ms — and its `enforce-target` failure is "missing UI screenshot", not a base problem.

PR #2781, opened 2026-08-27
feat(two-plane): phase 4 — machine listener, gui two-plane targets, usage slice

lidge-jun:
## Summary

- Phase 4 of the remote-hub stack (plan: `devlog/_plan/260827_remote_hub/060_phase4_two_plane.md`).
- Adds the client-mode machine plane: a loopback-only allowlist listener (`src/client/{runtime,machine-listener,machine-api,machine-auth,hub-relay}.ts`) serving `/api/machine/{status,clients,sync,shim,disconnect}` plus a fixed-target `POST /api/machine/hub-relay/*` relay (destination pinned to the connected hub, path allowlist, redirects rejected, hop-by-hop stripped, body caps, browser `Origin` forwarded verbatim on the pairing exchange, nothing logged). No `/v1/*` is reachable on this listener.
- Makes the gui two-plane aware: `gui/src/api-targets.ts` discovers shared/machine call bases; `gui/src/api.ts` keeps per-target in-memory session/CSRF state; existing Startup/Integrations/ApiKeys/Grok calls stay on the shared plane per call site.
- Adds the pairing form (`gui/src/connect-pairing.ts`), mounted from `App.tsx` in the connected-without-hub-session state, with a rendered-form test.
- Usage page renders whichever store served the traffic: connected → hub usage filtered to this machine's `apiKeyId` (toggle to hub-wide), disconnected → local `usage.jsonl`. No mirroring.

Screenshot (dashboard served from this branch's gui build):

gui dashboard on the phase-4 branch

## Verification

Run on lidge-ai (remote CI host) at 2484189, per the no-local-suite policy:

- `bun run typecheck` — clean.
- `bun test tests/client-machine-listener.test.ts tests/client-hub-relay.test.ts tests/client-connect.test.ts tests/api-usage.test.ts tests/usage-summary.test.ts tests/cli-start-journal-order.test.ts tests/core-lab-boundary.test.ts` — 118 pass, 0 fail.
- `cd gui && bun run test` — 1097 pass, 0 fail; `bun run build` — clean.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed (operator docs land in the deployment phase).
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults (loopback-only machine plane, SSRF negatives on the relay, per-target credential isolation).

PR #3149, opened 2026-09-01
docs(devlog): remote hub restack planning unit

lidge-jun:
## Summary

The planning unit for the remote-hub stack restack: `devlog/_plan/260901_remote_hub_restack/`, 21 commits, `devlog/` only.

It records the measured conflict surface before the work started, two adversarial audit rounds that both returned FAIL and what changed because of them, and a per-stage outcome document for each of the seven phases.

The two findings that changed the plan rather than confirming it:

- 33 unresolved inline review threads were missing from the first draft entirely, including six P1 defects. They are ledgered in `003` with an owning stage for each.
- Two blockers were assigned one stage too late. Measured from the diffs: `/api/machine/*` is introduced by p4 (49 added lines, zero in p5), and `gui/tests/api-auth-memory.test.ts` is first touched by p2 (zero in p3). Fixing a defect downstream of the stage that introduces it leaves the intervening heads red.

`081` records the exact-head CI work, including the finding that the `server-auth` websocket assertion fails on `dev`'s own HEAD with #3139 already landed — a dev-side flake rather than stack fallout, root-caused and split to #3147.

## Verification

- `devlog/` only; nothing in the build, typecheck, or test path reads from it.
- `bun run privacy:scan` passes.
- markdownlint clean on this unit apart from MD013, which the repository does not enforce (this unit has 18 line-length hits against 322 in the merge-train unit already on `dev`).

## Checklist

- [x] Documentation only
- [x] Targets `dev`
- [x] No production code changed

## Summary by CodeRabbit

- **New Features**
  - Improved remote hub connections, authentication, caching, key rotation, relay behavior, and standalone-client handling.
  - Added conditional discovery and UI behavior for standalone clients.

- **Bug Fixes**
  - Blocked insecure credential transmission and oversized unauthenticated requests.
  - Fixed connection cleanup, token recovery, relay authentication, supervised restarts, and catalog/profile rollback behavior.
  - Resolved multiple test and CI reliability issues.

- **Documentation**
  - Added audit findings, review-thread tracking, stack integrity checks, verification results, and remaining release tasks.


Ingwannu (changes requested):
Requesting changes on exact head `3804da257db34d1f9330e3da4852b963c20188f6`.

The restack ledger is useful, but this documentation PR currently records several claims that do not match the code or the live gates:

1. `081_wp8_ci_repairs.md` gives the wrong JavaScript root cause for `fetchImpl: typeof fetch = fetch`. A default-parameter initializer is evaluated when the function is called, not when the module is evaluated. In a real browser the bare global resolves through the current browser global; the observed happy-dom failure is more likely the split `window`/Bun-`globalThis` test realm or stale wrapper installation described later in the same file. Identify the actual binding/realm boundary and reclassify it instead of preserving the module-evaluation explanation.

2. `090_outcome.md` says the automated-gate target was reached while `enforce-target` is currently red and the document itself says #2776/#2789 still need required GUI screenshots. Product tests may be green, but the PRs are not automation-green or merge-ready. Keep the overall gate incomplete until the screenshots are attached and the exact-head check passes.

3. The literal `eslint-disable` text in `081_wp8_ci_repairs.md` triggers the deterministic `new_suppression` hygiene gate. This is prose, not a justified repository suppression; reword it as an ESLint suppression comment or otherwise avoid the machine suppression token. Do not apply `suppression-approved` to make a documentation wording issue pass.

4. This PR is titled and summarized as the remote-hub restack unit, but it also carries `devlog/_plan/260901_merge_train_round3/061_wp7_outcome.md` and `070_outcome.md` plus the unrelated `close merge train round 3` commit. Remove those files/commit from this PR or explicitly split them into the merge-train closeout they belong to. The current 23-file diff does not match the stated restack-only scope.

Also make the timeline labels unambiguous: `081` calls `b6aa976e9` the final p6 head, while `090` and the live PR use `4ab13b463`. If `081` is an intermediate snapshot, label it as such rather than “final chain.”

Resolve the two open inline findings, correct these scope/state discrepancies, and let CodeRabbit plus hygiene/enforce-target finish on the new exact head before requesting approval.

Ingwannu (changes requested):
Requesting changes on exact head `c4aaee67dccce2a8e1ef8e5784dc3798ed3807d4`.

The added polish ledger is useful, but the tracked record still disagrees with both JavaScript semantics and the live stack:

1. `081_wp8_ci_repairs.md:29-32` still says a default-parameter initializer binds `fetch` at module-evaluation time. It is evaluated at call time. The observed test failure may be a split `window`/`globalThis` realm or stale wrapper-installation problem, but the stated cause remains false and must be corrected rather than retained as history.
2. `090_outcome.md:5-15,71-72` and `102_axis_ledger.md:72-85` call the stack green and the automated target complete. Live exact head #2781 currently has `macos` and aggregate `ci` red; #2789 has `macos`, `enforce-target`, and aggregate `ci` red. More importantly, the new heads still contain unresolved product blockers in #2781, #2786, and #2789. Focused product checks must not be recorded as equivalent to required exact-head CI or review completion.
3. The PR still includes `devlog/_plan/260901_merge_train_round3/061_wp7_outcome.md` and `070_outcome.md`, which are unrelated to the remote-hub restack scope. Split or remove them.
4. The current `dev` security-working-notes policy forbids recording active, unfixed finding details in tracked `devlog/_plan`. `003_review_thread_ledger.md`, `002_audit_r1_synthesis.md`, and `000_research.md` still preserve active authentication/session/recovery weaknesses while their code changes are unmerged. Remove or redact those details; record the public outcome only after remediation is actually shipped.

The current PR checks being green only proves this documentation diff passes its own lightweight gates. It does not validate the claims it makes about other exact heads. Resolve the still-open actionable threads and update the ledger from current GitHub state before requesting approval.

lidge-jun:
Rebased onto the merged stack and addressed all four on `9d59ce57c`.

**1. The `fetchImpl` claim was wrong and is now corrected.** You are right that a default-parameter initializer evaluates at call time, not module evaluation. I rewrote `081_wp8_ci_repairs.md` §3 to say so explicitly, mark the original sentence as an error rather than quietly deleting it, and describe the actual mechanism: the split happy-dom `window` versus Bun `globalThis` realm, with `installApiAuthFetch` wrapping `window.fetch`. Leaving a false model in the record is worse than leaving a gap.

**2. The snapshot greens are gone.** `090_outcome.md` and `102_axis_ledger.md` now carry merge commits instead of a CI column, plus a short section naming what actually went red before the merge and why none of it was a code regression: the websocket flake (root-fixed by #3147, `408652698`), a `shutdown drain cap expiry` spill flake in a file the stack never touches, and `keyring-smoke=abandoned` runner drops that turned the aggregate `ci` job red. Both documents now say plainly that focused checks are not required-CI.

**3. Out-of-scope files removed.** `260901_merge_train_round3/061_wp7_outcome.md` and `070_outcome.md` are dropped from this PR.

**4. This is the one that changed the plan, and thank you for it.** You were right that the writeups preserve active weaknesses. T20 in particular — `003_review_thread_ledger.md:41` and `030_wp3_p2_remote_session.md:41-45` — states the reproduction condition for unauthenticated unbounded buffering, and `dev` had no `declaredLength` cap at the time. My plan was to merge this documentation PR *first*, which would have published a live reproduction against an unfixed `dev`.

I reversed the order instead of redacting. The stack is now merged, so `b7282858b` ("bound the unauthenticated exchange body") is on `dev` via #2776 (`39e5aefb6`) and the cap is verifiable in `src/server/index.ts`. By the `AGENTS.md` test — "is there already a public diff that reveals this weakness?" — these notes are now post-disclosure records rather than pre-disclosure material, which is exactly the condition that puts them in the tree legitimately. I added a header to `003_review_thread_ledger.md` recording that ordering constraint, and `112_wp2_order_reversal.md` explains the reversal so the next person does not repeat my sequencing mistake.

Stack merge results, for the record: #2771 `278fd613a`, #2772 `87459f8c3`, #2776 `39e5aefb6`, #2777 `fd8b6b895`, #2781 `163feb6ee`, #2786 `6d732d3dc`, #2789 `9232df0e6`.

PR #3174, opened 2026-09-01
fix(gui): stop the mobile topbar and integration cards overflowing

lidge-jun:
## Summary

Two responsive defects on the dashboard, both found by measuring real geometry through the Chrome DevTools Protocol rather than by reading CSS.

**The mobile topbar overlapped itself.** At a 320px viewport the version badge ended at x=245 while the action orbs began at x=206, so the power orb sat on top of the badge. A flex item only shrinks below its content size when it carries `min-width: 0` itself; `.mobile-topbar .brand` had it but its children did not, so `.name` held its intrinsic width and pushed `.ver` under the actions.

Shrinking `.name` is necessary but not sufficient — the row budget at the narrowest width is 44 (menu) + 26 (logo) + 56 (badge) + 94 (actions) plus gaps, leaving the name about 38px, which rendered as `op…`. The badge is dropped instead: the same brand node is mounted again in the drawer head, and the live value is also on the dashboard Version stat.

**The integration cards overflowed the page.** `.integration-cards` used `repeat(auto-fill, minmax(260px, 1fr))`. A bare 260px floor is wider than the content box of a 320px viewport, so the track could not shrink and the card went with it — the Settings button measured left=326, right=409 against a 320px page, and again at 375px and 736px. `min(260px, 100%)` keeps the multi-column intent while letting a narrow viewport fall back.

## UI change

Before — 320px, the version badge sits under the power orb, and the integration cards push their action row off the page:

opencodex dashboard at a 320px viewport before the fix, version badge overlapped by the power button in the top bar

After — 375px, the brand, badge and orbs each hold their own space (`name.right=179`, `ver 189-245`, `actions.left=261`):

opencodex dashboard at a 375px viewport after the fix, top bar showing logo, name, version badge and two action orbs without overlap

At and below 360px the badge is dropped so the product name stays readable; from 375px up it returns.

## What review changed

A read-only review pass caught the first attempt inventing `@media (max-width: 400px)` — the only 400px rule in the file, covering an unmeasured 375-399 band. It is folded into the existing 360px tiny-phone breakpoint instead. The measurement confirms the reviewer was right: at 375px the badge is back and `ver.right=245 < act.left=261`, so a 400px rule would have hidden it for no reason.

## Also in this PR

[...]

PR #3181, opened 2026-09-01
docs(devlog): close the multiplatform QA and GUI unit

lidge-jun:
## Summary

Closes out the `260902_multiplatform_qa_and_gui` unit with the two documents written after #3174 merged: the CI-tracking record and the objective closeout.

- `091_wp6_merge_outcome.md` — the three CI failures on #3174 and why each was different. `test 1/4` was a `dev`-owned failure already fixed by `22a643a00` (my diff touched no `src/` file), `gates` was `privacy:scan` catching real remote home paths in the devlog, and `enforce-target` was an empty description my own bad `gh pr edit` had produced. Also records why the planned stack stayed a single PR: the change was CSS ×2 + tests ×2 + docs, one coherent unit that splitting would only have given a dependency chain.
- `092_objective_closeout.md` — the tally: 6 PABCD cycles, 4 hosts across 3 operating systems, 17 QA scenarios, 3 defects, all hosts restored and verified item-by-item against their recorded pre-state.

Docs only.

## Verification

No full suite, per the no-local-suite policy. This PR adds two markdown files and changes no code.

`bun run privacy:scan` passes — remote home paths were anonymised to `~/` when the scan flagged them on the parent PR.

## Checklist

- [x] Docs or release notes were updated when needed — this PR is the documentation.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults — no code change; remote paths are anonymised and no credential or account data from any host is recorded.


Ingwannu (changes requested):
Requesting changes on exact head `9dd45b7d618a31d5668c7456137c26460060774f`.

The closeout text reintroduces the private home paths it says were anonymised. Exact-head `bun scripts/privacy-scan.ts` fails on `091_wp6_merge_outcome.md:13` for `/Users/junny/` and `/Users/tig/`; the same sentence also retains the other raw home examples. Please describe them generically (for example, “two macOS homes, one Linux home, and one Git-Bash home”) or use only `~/` without the usernames. The PR body currently claims the scan passes, but it does not on this head.

`git diff --check origin/dev...HEAD` also reports a new blank line at EOF in both added files (`091...md:33` and `092...md:45`). Remove those trailing blank lines.

No code/full-suite run is needed for this docs-only correction. After those two deterministic checks pass on a new exact head, this should be a straightforward approval candidate. Local validation used isolated homes and the protected runtime configuration hashes remained unchanged.

Issue #486, opened 2026-07-26
[Feature]: Make failover cooldown duration configurable (currently hardcoded 60s)


a1980881039:
**Version:** v2.7.40

**Problem**
The 429 cooldown for combo failover is hardcoded — `DEFAULT_COOLDOWN_MS` (60s) in `src/combos/failover.ts` and `src/providers/key-failover.ts` — with no GUI or config option. Users who want the primary target retried quickly (e.g. every 5 seconds) must patch the source, and the patch is lost on every `ocx update`.

The passive design (timestamp compare, no active probing, in-memory only, cleared on restart) is perfectly fine — what's missing is just the knob.

**Suggestion**
Add a per-combo (and/or per-provider) `cooldownMs` setting, exposed in the combo editor UI. Keep 60s as the default.

a1980881039:
One more note: if you're busy, I'd be happy to take a stab at this myself and open a PR. Just let me know.

PR #3329, opened 2026-09-03
feat(combos): per-combo cooldownMs and waitForCooldownMs for bursty upstream 429s

Veritas-7:
## Summary

- Add per-combo `cooldownMs` (1–600000 ms) and `waitForCooldownMs` (0–600000 ms) controls because a fixed 60 s per-target cooldown turns a short burst of upstream 429s into minutes of immediate 503s.
- Apply the full cooldown precedence chain: explicit `Retry-After`, with `0` or a past date normalized to `1 ms` → Codex reset headers → explicit combo `cooldownMs` → 5 s request-rate fallback for upstream codes `1302`/`1305` → 60 s default. An unset `cooldownMs` keeps the upstream fallback (5 s for request-rate codes `1302`/`1305`, otherwise 60 s); an explicit value overrides it and is persisted as-is. Codex reset headers are ignored for cyber-policy failures exactly as `Retry-After` already was.
- Wait candidates skip targets whose provider quota is cached-exhausted. Waiting is capped per selection attempt, so a multi-hop request may wait up to `hops × waitForCooldownMs`; it is not jittered and is cancelled by an abort.
- Every all-cooling 503 goes through the existing `comboUnavailable` path, carrying `Retry-After` in whole seconds with a minimum of `1`.
- A management `PUT` preserves omitted knobs; `cooldownMs` is cleared only via the configuration file, while `waitForCooldownMs` resets to its default when an explicit `0` is sent because the sparse serializer drops the default. A missing or null `combo` still returns the pre-existing 400. Dashboard fields are deferred.
- Documentation is updated in the combo guide and routing reference.
- Codex reset headers are suppressed for cyber-policy failures exactly as `Retry-After` already was, and a regression test now covers that half as well.

Over a 90-minute window the proxy saw 36 upstream 429s from a 3-target combo; because each 429 cooled its target for 60 s, 648 requests were answered with 503 `No available targets for combo` within ~10 ms each.

Refs #486

## Verification

Base commit: `664d80c76` (`upstream/dev`); branch head `1876d6001`; not behind dev.

- `bun run typecheck` — exit `0`.
- `OCX_TEST_NO_QUEUE=1 bun test tests/combos.test.ts tests/server-combo-failover-e2e.test.ts tests/combo-management-api.test.ts tests/agent-task-recovery-combo.test.ts tests/agent-task-recovery.test.ts tests/cyber-policy-error-fidelity.test.ts` — `223 pass`, `0 fail`, `1276 expect() calls`, 6 files; exit `0`.
- `OCX_TEST_NO_QUEUE=1 bun scripts/test.ts --changed=upstream/dev --path-ignore-patterns '**/tests/management-integration-routes.test.ts'` — `13871 pass`, `11 skip`, `0 fail`, `224392 expect() calls`, `Ran 13882 tests across 741 files. [150.44s]`; exit `0`.
- The one excluded suite is `tests/management-integration-routes.test.ts`, which builds the GUI bundle. It cannot run on this machine: the bundler's CSS minifier reports `Cannot find module '../lightningcss.darwin-arm64.node'`. That is a local toolchain problem unrelated to this change. With it excluded the gate is green with zero failures.
- This rebase pulled in an upstream commit that also edits `src/server/responses/core.ts`. It merged without conflict, and the branch's own changes there are intact: the `cyberFailure` gate on both `retryAfter` and `resetAt`, the wait-enabled pick passed `waitForCooldownMs` and the abort signal, and both `comboUnavailable` returns guarded by the abort check.
- Both new tests were checked against deliberate regressions: removing the `!cyberFailure &&` guard on the reset-header branch makes the cyber-policy test fail with the three reset timestamps leaking through, and inverting the sparse-serialisation condition makes the management tests fail with `Received value: 0` where the key must be absent.

## Checklist

- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. Default cooldown semantics, the 1–600000 ms / 0–600000 ms bounds, the abort path, and secret/PII sweeps were reviewed across three review rounds.

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - Added configurable combo cooldown durations with upstream reset signals and fallback behavior.
  - Added bounded waiting for the earliest available combo target before returning `combo_unavailable`.
  - Added request cancellation support while waiting for cooldowns.
  - Combo failures now expose applicable retry timing through `Retry-After`.
  - Management API updates preserve existing cooldown settings when omitted and support resetting wait duration to zero.

- **Documentation**
  - Updated combo guides and configuration references in multiple languages with the new settings, precedence rules, API behavior, and troubleshooting guidance.


lidge-jun:
## 리뷰 · 우선순위 71 / 80

이 PR은 콤보(failover)에서 짧은 upstream 429 버스트가 나면, 고정 60초 쿨다운 때문에 바로 줄줄이 503이 나던 문제를 줄이려고 **콤보마다** `cooldownMs`와 `waitForCooldownMs`를 넣는다. 지금 `dev` HEAD `ff1ac6b8c`의 `src/combos/failover.ts`는 이미 `COMBO_REQUEST_RATE_COOLDOWN_MS`(5초, 코드 1302/1305)와 기본 60초, `Retry-After` 파싱, 프로세스 로컬 `targetCooldowns` 맵을 갖고 있다. 다만 "이 콤보는 짧게 식히고, 잠깐 기다려도 된다"를 **설정으로** 고를 수는 없었다. 버스티 429를 쓰는 upstream(특히 rate-limit 코드만 던지고 Retry-After가 약한 곳)에서는 운영자가 콤보 단위로 쿨다운·대기 상한을 조절하는 게 맞다.

쿨다운 우선순위는 PR·문서가 같이 적는다. 강한 쪽부터: 명시 `Retry-After`(0·과거 HTTP-date도 immediate로 보존) → Codex reset 헤더(`parseResetCooldownMs`, #3256 계열) → 콤보 `cooldownMs`(있을 때) → 1302/1305면 5초 → 그 외 60초. 상한은 여전히 10분. `coolComboTarget`이 이 체인을 타고, `advanceComboAfterFailure`는 `options.cooldownMs ?? combo?.cooldownMs`를 넘긴다. 새로 생긴 `pickComboTargetWithWait`는 한 번 고르고, 전부 쿨링이면 할당량 고갈·제외·비자격 타깃을 빼고 가장 빨리 끝나는 쿨다운만 `waitForCooldownMs` 한도 안에서 `sleepWithAbort`로 기다린 뒤 **한 번 더** 고른다. hop마다 이 대기가 반복될 수 있어 이론상 `hops × waitForCooldownMs`까지 늘어난다. 기본 `waitForCooldownMs`는 0이라 예전처럼 바로 `combo_unavailable` 503(가장 이른 남은 쿨다운을 `Retry-After` 초로 실어 보냄)이다.

관리 API·타입·문서도 맞춰 있다. `src/types/config.ts`의 `OcxComboConfig`에 필드 추가, `src/combos/types.ts`에서 1..600000 / 0..600000 정수 검증, `combo-routes` PUT은 키를 **생략하면 기존 값 유지**. `cooldownMs`를 API로 "없애기"는 못 하고 설정 파일 편집만 가능하다고 문서에 명시했고, `waitForCooldownMs`는 명시 0으로 기본값으로 돌아간다(sparse 직렬화). 대시보드 에디터에는 아직 노출 안 함 — 가이드에 follow-up으로 적어 둔 상태. 테스트는 management API·combos·failover e2e 쪽이 같이 움직인다. types/config에 필드만 덧붙이는 수준이라 대분할 캠페인에 걸려 close-don't-rebase 할 종류는 아니다.

현재 `dev` 방향(콤보 failover·quota·encrypted recovery 등)과 잘 맞고, 기본값 0/unset이면 동작이 거의 그대로라 회귀 면적은 조절 가능한 편이다. 71점은 (1) 실사용 pain이 분명하고 (2) 문서·우선순위·abort 취소까지 적어 둔 완성도가 높으며 (3) 다만 대기 무jitter·프로세스 로컬·대시보드 미노출·API로 cooldownMs 클리어 불가 같은 운영 판단이 남아 80 미만이다.

경로 `pickComboTargetWithWait` - 대기는 jitter 없음. 문서도 synchronized wake-up 가능하다고 인정한다. 같은 콤보로 동시 요청이 많으면 만료 시각에 한꺼번에 다시 때릴 수 있다.
경로 `coolComboTarget` 우선순위 - `options.cooldownMs`가 reset/`Retry-After`보다 아래에 있어서 의도와 맞다. cyber-policy 실패에서 reset 헤더를 무시하는 기존 `Retry-After` 대칭이 코드·본문에 있는지 CI/리뷰에서 한 번 더 확인하면 좋다.
경로 management PUT - `cooldownMs`를 설정에서 지우려면 파일 편집만 된다. 의도된 sparse 직렬화 한계지만, `null` 전송으로 클리어하는 계약을 나중에 넣을지 정해야 한다.
경로 `console.warn` - 대기 시 서버 로그에 콤보 id·delay를 찍는다. 버스티 구간에서 로그 폭주 가능. 샘플링·debug 레벨 여지는 있다.
라인 - 대시보드 미노출 - 파워유저·파일/`PUT`만 쓸 수 있다. 기능 merge와 UI follow-up을 나누는 건 타당하다.

메인테이너의 판단이 필요한 지점
- 기본 unset/0을 유지한 채 opt-in으로 둘지, 자주 쓰는 콤보 프리셋에 짧은 `cooldownMs` 예시를 문서/시드에 넣을지
- wait 경로에 최소 jitter(또는 콤보 단위 mutex)를 이 PR에 넣을지, 관측 후 follow-up으로 둘지
- `cooldownMs`를 management API에서 제거(`null`)할 수 있게 할지
- GUI에 두 노브를 언제 노출할지(이 PR 범위 밖이라는 선택은 동의 가능)

너의 추천
- CI 초록이면 **merge 후보**. 기본 동작 보존 + 문서화된 우선순위 + 테스트 확장이 있다.
- merge 후 이슈로 "dashboard cooldown knobs"와 "optional wait jitter"만 남겨도 충분하다.
- `src/types/config.ts`에 필드 몇 줄 추가일 뿐이라 types/config 대분할 때문에 close 할 대상 아님.

이 댓글은 grok-bot이 작성했습니다

Veritas-7:
Thanks for the detailed read. Answering the two points you asked to be checked, and rebased.

**Cyber-policy symmetry.** Confirmed in code and now pinned by a test. `cyberFailure` is computed at `src/server/responses/core.ts:1605` and gates both fields at `:1641-1644`, so a policy refusal drives neither a `Retry-After` nor a reset-derived cooldown. The suite previously covered only the `Retry-After` half; this branch adds `drops Codex reset headers as well as Retry-After for a cyber-policy failure` to `tests/cyber-policy-error-fidelity.test.ts`. It is a real regression test: removing the `!cyberFailure &&` guard makes it fail with the three reset timestamps leaking through.

**Wait-path log volume.** Bounded, not unbounded. With the default `waitForCooldownMs: 0` the warning cannot fire at all — `src/combos/resolve.ts:332` returns before it. With a non-zero value it is at most one line per selection attempt, and because each retry accumulates `exclude`, a single request is capped by the number of targets in the combo. I have left it as is; happy to sample it or drop it to debug level if you would rather.

**Open decisions.** No objection to any of them, and I would take all three as follow-ups rather than growing this PR: wait jitter or a per-combo mutex, clearing `cooldownMs` through `null` on the management API, and dashboard exposure. If you would prefer jitter in this PR, say so and I will add it here instead.

Rebased onto `1aa839aa8`. Locally: typecheck clean; the combo suites are 222 pass / 0 fail across 6 files; the changed-import-graph gate is 13842 pass / 11 skip / 2 fail across 740 files. Neither failure is from this branch — one is the GUI suite, which cannot build its bundler native binding on this machine, and the other is a load-sensitive process-status test that passes twice when run alone.

Veritas-7:
Both automated findings are addressed in `7d8a562` and `c4a7c5d` (now `eda41e33b` after a rebase onto `b15cbb2c3`).

**Explicit `waitForCooldownMs: 0` reset (major).** Confirmed against the code before writing the test: the PUT keeps the explicit `0`, normalization keeps it, and the sparse serializer drops it because it equals the default, while an omitted `cooldownMs` is back-filled from the stored value. `tests/combo-management-api.test.ts` now has `PUT resets an explicit zero waitForCooldownMs to the sparse default`, asserting on both the `GET /api/combos` response and the persisted config that `waitForCooldownMs` is gone and `cooldownMs` stays `5000`. It is a real regression test: inverting the sparse condition makes it fail with `Received value: 0`, and it takes the pre-existing sparse-defaults test down with it.

**Stale translated guides (minor).** The Japanese, Korean, Russian and Simplified Chinese guides now carry the same precedence chain, both settings with their ranges, the per-attempt wait semantics with the `hops × waitForCooldownMs` bound and the no-jitter caveat, the abort behaviour, and the management PUT rules including that `cooldownMs` clears only through the config file while an explicit `0` resets `waitForCooldownMs`.

One thing that turned up while doing it: all four translations already stated that the `Retry-After` on a `combo_unavailable` 503 has a floor of one second, and `src/combos/failover.ts` has always applied `Math.max(1, ...)`, but the English guide omitted the clause. The English guide is corrected rather than the translations.

Veritas-7:
Both findings are addressed in `5b5c58f` and `5ca2f16`.

**Persisted config, not the in-memory object.** Your premise held: the route assigns `config.combos = nextCombos` on the object the tests already hold, and the follow-up `GET` reads that same object, so nothing was verifying what actually reaches disk. Each persistence case now re-reads the configuration file through the existing config-path helper and asserts both presence and absence: the explicit knobs, an explicit `cooldownMs: 60000` with `waitForCooldownMs` omitted, the dashboard-shaped PUT that must preserve omitted knobs, and the sparse-defaults case. Checked as a real regression test: making the sparse serializer always write the default fails two of them with `Received value: 0` where the property must be absent.

**Troubleshooting precedence.** Added to all five language versions: follow an observed `Retry-After` first, Codex reset headers also outrank `cooldownMs`, otherwise the configured value applies or the fallback does (5 seconds for `1302`/`1305`, else 60 seconds), and every cooldown is capped at 10 minutes. Verified against the chain in `src/combos/failover.ts` rather than restated from the PR description.

Local after these commits: typecheck clean, the combo persistence suites 201 pass / 0 fail.

Veritas-7:
Correcting something I wrote earlier in this PR.

In my note about the translated guides I said the four translations already documented the one-second floor on the `Retry-After` of a `combo_unavailable` 503, and that only the English guide was missing it. That is wrong, and the commit message of `9672a80` (now rebased) carries the same wrong framing.

Checking the tree at the parent of that commit: none of the Japanese, Korean, Russian or Simplified Chinese guides mentioned the floor, and neither did the English one. The floor sentence entered all five in this branch — the translations got it in the sync commit and the English got it in the follow-up. I described my own new text as pre-existing.

What is true, and what actually matters for review: `comboCooldownRetryAfterSeconds` in `src/combos/failover.ts` has always applied `Math.max(1, Math.ceil(remainingMs / 1000))`, so the documented floor matches the code. The behaviour was undocumented in every language before this branch and is documented in every language now.

Nothing else in the change is affected; I would rather flag the misstatement than leave it in the record.

lidge-jun:
Held on one specific point, not on the feature.

`cooldownMs` and `waitForCooldownMs` as opt-in knobs are welcome. What is mixed in is an **unconditional change to existing behavior**: `coolComboTarget` on `dev` resolves `options.cooldownMs ?? parseRetryAfterMs(...)` (`src/combos/failover.ts:212-213`), and this PR reverses that to `parseRetryAfterMs(retryAfter, now, { preserveImmediate: true }) ?? parseResetCooldownMs(resetAt) ?? cooldownMs ?? ...`.

Two things change for every existing combo user, none of whom opted in:

1. **Priority inversion.** A configured `cooldownMs` no longer wins over an upstream `Retry-After`. Whether that is correct is a product call. `tests/combos.test.ts:445-453` only pins Retry-After against the *request-rate default*, so the precedence against an explicit setting is genuinely unspecified — which is why it needs deciding rather than assuming.
2. **`preserveImmediate: true`.** Per `failover.ts:127-136`, `Retry-After: 0` and past dates now produce a 1ms cooldown instead of `undefined`.

The `resetAt`-derived cooldown in `consumeComboFailure` is unconditional too.

Separately, `waitForCooldownMs` blocks the request thread for up to 600s through `sleepWithAbort`, which has no heartbeat, while this proxy runs a stall watchdog (`src/lib/upstream-retry.ts:118-124` documents a 300s stall budget). Abort handling is present and correct; the watchdog interaction is simply unverified.

**Suggested split, which would let the uncontroversial part land quickly:**

1. `cooldownMs` as a pure fallback — types, validation, management route, docs. Keeps `options.cooldownMs ?? parseRetryAfterMs(...)` ordering.
2. The `resetAt` source and the Retry-After precedence change, reviewed on their own with the existing-behavior impact stated.
3. `waitForCooldownMs`, with evidence about the stall watchdog.

The branch is also 98 commits behind `dev` and will need a rebase. The implementation quality is high — this is about separating a new knob from a silent change to everyone else's behavior.

Ingwannu (changes requested):
The per-combo cooldown/wait direction is valuable, but exact head `1876d6001db50462805537f9bfea655ed97987ea` still has two concrete blockers.

1. In `src/server/responses/core.ts` around lines 1645-1647, reset metadata is retained only when the raw response status is 402/429. `shouldRetryCodexPoolAccountQuota` also recognizes body-confirmed quota exhaustion wrapped in HTTP 5xx and normalizes that outcome to 429 for account-pool handling. If no alternate account is available, the raw 5xx response reaches `consumeComboFailure`; the current status-only condition then drops its valid `x-codex-*-reset-at` metadata. The combo target can consequently become eligible before the real quota reset. Preserve the effective quota classification through this path (while keeping the cyber-policy exclusion) and add a focused 5xx-body-confirmed quota regression.

2. `docs-site/src/content/docs/ko/guides/combos.md` line 130 says every all-cooling selection returns HTTP 503 immediately, immediately after documenting positive `waitForCooldownMs`. That contradicts the runtime and the English source. Limit the immediate-503 statement to the default zero wait budget or to an earliest cooldown that exceeds the configured budget.

The current check set contains only the four intake/gate checks; cross-platform CI has not run for this exact head. After the two fixes, please refresh the branch against current `dev` and obtain exact-head CI before merge review.

lidge-jun:
Carried into dev via #3606, landing 3ac31078244ea04c9abce0e50275ffaccf25455a. The carry preserves current no-cooldown scope and corrects 5xx reset metadata, wait clocks and immediate Retry-After handling. Original attribution is preserved. Closing as superseded; final dev Linux CI is the batch gate.

PR #3453, opened 2026-09-04
docs(devlog): record the second-round bug backlog merges

lidge-jun:
## Summary

Appends the second-round record to the 260904 bug-backlog closeout. Documentation only — no runtime change.

After the triage reviews went out, three authors pushed fixes for the exact defects named in them and a fourth PR turned out never to have been failing, so #3403, #3432, #3325 and #3394 all landed. This records their merge shas and one finding worth keeping:

**A red check is not the same as a failing check.** Four of those PRs showed `FAILURE` in the status rollup while the *latest* run of every individual check was green — the rollup was still carrying superseded entries from runs cancelled by a newer trigger. Grouping the rollup by check name and keeping only the most recent run per name is what distinguishes "this PR is failing" from "this PR has failed before". Reading the aggregate would have left four correct PRs parked indefinitely.

Also records that #3432 was verified beyond its own test names: the three whitespace bypass forms from the original finding were run directly against `enforceEventStructureLimits`, and the two legitimate URL forms were confirmed still admitted.

## Verification

- `bun run privacy:scan` — passes (the gate that reads `devlog/`)
- No source, test, or config files touched; `devlog/` is outside the build, typecheck, and test paths

No GUI changes.

## Checklist

## Summary by CodeRabbit

- **Documentation**
  - Added a “Second round” closeout section covering four additional pull requests and their post-review updates.
  - Documented how cancelled historical checks could produce misleading red status rollups.
  - Recorded the latest-per-check evaluation method and direct validation of the bypass fix.
  - Listed four remaining open items whose known defects are unchanged.

### DIFF
--- a/devlog/_plan/260904_bug_backlog_closeout/070_closeout.md
+++ b/devlog/_plan/260904_bug_backlog_closeout/070_closeout.md
@@ -70,6 +70,34 @@ Holding it would have kept `dev` red for the duration.
 on four runs across three unrelated branches, so it predates this work. Filed rather than
 worked around, per the standing instruction about Windows failures.
 
+## Second round: four more landed after the authors responded
+
+The triage reviews were not the end of those items. Three authors pushed fixes for the exact
+defects named in them, and a fourth PR turned out never to have been failing at all.
+
+| PR | Author | dev sha | What changed after the review |
+|----|--------|---------|-------------------------------|
+| #3403 | ianlyoo | 43248e499 | rebased 68 commits onto dev; collision guard landed |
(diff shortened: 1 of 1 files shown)


lidge-jun:
Owner `pull_request` bypass merge, recorded per MAINTAINERS.md — same reason as #3439 and #3446: GitHub refuses self-approval, so no ordinary review is available for a maintainer-authored change.

This one is documentation only. Nothing under `src/`, `tests/`, `gui/`, or `.github/` is touched, `devlog/` sits outside the build, typecheck, and test paths, and `privacy:scan` — the gate that does read it — passes. Every required check is green.

PR #3479, opened 2026-09-04
feat(release): move the dev version line before the release, not after

lidge-jun:
## Summary

Every release forces `dev` to catch up, and while it trails the highest tag,
`tests/release-version-line.test.ts` fails on `dev` **and on every open pull request**
— an inherited red a contributor cannot fix from their own diff.
`.github/workflows/dev-version-bump.yml` records four hand repairs in its own header
(`32529c2b2`, `e4a85d134`, `076ad3036`, `befcac3e1`), and history shows
"move dev to 2.4x.0" once per release.

This lands the four-phase fix from an audited design unit. The design passed six
adversarial review rounds and twice retracted a claim rather than repairing it — the
retractions are the most important part of it.

**What could NOT be done, and is now stated plainly.** The per-release `dev` commit
cannot be deleted. It follows from `Protect dev` requiring review, `release.ts`
running only on `main`/`preview`, and a monotonically advancing tag set. An earlier
draft claimed otherwise; a reviewer proved `ee2d19ad4`'s single parent is the
`v2.42.0` release commit, making the catch-up PR the **ancestry carrier**, not merely
a version fix. A second draft then tried to preserve that ancestry — also impossible,
since `release.ts:559-591` creates the release commit *after* promotion, so it is a
descendant. Both claims were withdrawn. Ancestry is explicitly not a property this
design maintains.

**So the commit MOVES instead of disappearing.** The version PR opens and merges
*before* the release rather than after it. Same number of reviewed commits, no red
window.

### The four phases

| Phase | Change |
|---|---|
| 010 | `scripts/version-line.ts` — one shared, importable version algebra |
| 020 | `--bump patch\|minor\|major` with channel-specific resolvers |
| 030 | `dev-version-bump.yml` becomes a pre-release opener; two gates in `release.yml` |
| 040 | `MAINTAINERS.md` + SoT corrected; invariant retained |

**010** exists because the repository ordered releases in two places and only one was
reachable from a test. The two comparators stay deliberately different and a test now
pins that: `compareVersions` throws so a release decision fails closed, while
`compareTagsLenient` keeps the natural-ordering fallback. Collapsing them would be a
live regression — `build-release-changelog.ts` admits any `/^v\d/` tag, so one
malformed historical tag would newly abort release-note generation.

**020** keeps the channels apart. A single global floor would let a future
`v2.43.0-preview.1` turn `--bump minor` into `2.44.0` and skip the intended `2.43.0`.
A stable patch bump is now refused when a preview tag sits above the base.

**030** deletes `workflow_call` together with its only caller, normalizes one target
version before the decision step, and adds a readiness gate plus an ordering gate. The
ordering gate runs *after* the fresh tag fetch — before it, the stale tag set would
defeat the point — and `--allow-existing-tag-at-head` is granted only for a dry run
whose tag names the exact SHA, preserving the deliberate exception already there.

**040** is documentation only. `tests/release-version-line.test.ts` gains two comment
lines and nothing else; its assertions are byte-identical and `tagPointsAtHead` is
retained.

### One policy consequence, stated deliberately

Publishing a preview for a higher core **ends the current stable patch line**. This is
a restriction, not the preservation of an unused capability: history contains real
counterexamples (`v2.6.24-preview.20260705` then `v2.6.23`; `v2.7.39-preview.20260724`
then `v2.7.37`), and 103 of 143 stable tags carry patch > 0. It is recorded in
`MAINTAINERS.md` and the SoT so nobody re-derives it as a bug.

## Verification

Phase-local checks were run by the implementing agents before the local-execution
freeze:

```
010: version-line 7 pass, bump-dev-version 10 pass (file unchanged), release-notes 71,
     release-version-line 3, release-helper 33, typecheck exit 0
020: version-line 20 pass, release-helper 39 pass, release-version-line 3 pass,
     typecheck exit 0, privacy:scan passed, docs-site build 425 pages
030: ci-workflows 136 pass, bump-dev-version 14 pass, version-line 20 pass,
     typecheck exit 0
040: documentation only; verified by diff inspection
```

New assertions were proven **red before** the change and green after: 020's resolver
suite failed on the higher-core patch refusal and the equal-stamp succession; 030's
ordering assertion fails when the gate is moved ahead of the tag fetch and passes when
restored; 010's lenient/strict distinction failed against a collapsed comparator.

`Service lifecycle` is already green on `b181dd791`, which exercises the changed
workflows. **This PR exists so the full cross-platform matrix verifies the change set
on CI rather than on a developer machine.**

`tests/bump-dev-version.test.ts` stayed green **without edits** through phase 010 —
that was the proof the algebra extraction was faithful.

## Checklist

Security note: this touches release automation and workflow permissions, so it needs
maintainer security review per `MAINTAINERS.md`. No secret is added or logged; the
deleted `workflow_call` removes a `contents: write` caller path, and both new gates are
read-only assertions. `privacy:scan` passed during phase 020.

Design and audit history: `devlog/_plan/260904_release_version_line/`.

## Summary by CodeRabbit

* **New Features**
  * Added `--bump patch|minor|major` options to calculate release versions automatically.
  * Added reasoning-effort controls for native Anthropic models, supporting low through max.
  * Added safeguards preventing releases that would violate version ordering or outdated development branches.

* **Bug Fixes**
  * Preserved configured model-specific effort settings while filling missing model capabilities.
  * Models without reasoning support are now correctly identified as unsupported rather than unknown.

* **Documentation**
  * Updated release guidance across supported languages to describe version bump options, preview rules, and the pre-release development-version process.

### DIFF
--- a/.github/workflows/dev-version-bump.yml
+++ b/.github/workflows/dev-version-bump.yml
@@ -1,48 +1,41 @@
 name: Dev version bump
 
-# When a release publishes, open a pull request that moves `dev` past the published
-# version. Without this, `dev` keeps carrying a version that is at or behind a released
-# one, and `tests/release-version-line.test.ts` fails on `dev` and on every pull request
-# opened against it - inherited red a contributor cannot fix from their own diff.
+# Before a release publishes, open a pull request that moves `dev` past the intended
+# version. Merge that pull request before promoting and publishing so `dev` and pull
+# requests based on it never inherit a version-line failure from the new tag.
 #
 # That has been repaired by hand four times: 32529c2b2, e4a85d134, 076ad3036, befcac3e1.
-# The second of those ADDED the detector and two more repairs followed it, so more
-# visibility was never the missing piece; a prepared change was.
+# The workflow now prepares the move before publication. Explicit repair mode retains
+# the old catch-up capability if a release somehow publishes without the pre-move.
 #
 # WHAT THIS DOES NOT DO. It does not push to `dev`. It opens a pull request and a human
 # merges it, because ruleset `Protect dev` requires an approving review and code-owner
-# sign-off that a bot cannot supply. Until that merge the red persists. This converts a
-# forgotten chore into a queued, reviewable change - not into an automatic repair.
+# sign-off that a bot cannot supply. `release.yml` independently refuses publication
+# until `dev` already outranks the intended version.
 #
-# WHY THIS IS CALLED, NOT TRIGGERED. It used to listen for `release: published`, and in
-# that form it ran ZERO times across v2.37.0, v2.38.0 and v2.39.0 - every one of those
-# bumps was still opened by hand (#3045, #3076, #3127). The workflow was not broken; the
-# event never existed. `release.yml` creates the GitHub release with
-# `GH_TOKEN: ${{ github.token }}`, and GitHub does not start workflow runs from events
-# raised by the default `GITHUB_TOKEN`. A `release: published` listener therefore cannot
-# observe a release this repository publishes itself, no matter which branch it sits on.
+# WHY THIS IS DISPATCHED. The intended version is known before publication, and this
+# workflow's purpose is to queue the reviewed `dev` move first. It is not called by the
+# release workflow after an irreversible publish, and it does not react to release events.
 #
-# The fix keeps the credential surface unchanged: no PAT, no app token, no
-# `contents: write` on the release job. `release.yml` CALLS this workflow directly after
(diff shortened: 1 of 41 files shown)


lidge-jun:
Superseded by #3481. This branch was stacked on the anthropic effort-ladder work, which landed squashed as df416a439, so the two carried commits conflicted with dev and GitHub could not build a merge commit — no test workflow ever queued. #3481 carries the same four release-line commits onto a clean dev, preserving #3325's server-side fork-head filter in the rebased workflow.

PR #3477, opened 2026-09-04
feat(integrations): let an operator delete one rollback journal entry

lidge-jun:
## Summary

Refs #3379 — this lands the **rollback-entry deletion** item only. The custom usage ranges and the account-selector rename in that issue are untouched, so it does not close it.

An operator could see the rollback journal grow and had no way to remove a row. `GET /api/client-integrations/journal` was read-only, and `RollbackHistory.tsx` rendered restore/expiry state with no delete affordance.

**Deletion is a tombstone append, not a rewrite.** `journal.ts` opens with three promises — append-only commit ordering, only a torn last line can be lost, and no journal-specific lock — and rewriting `journal.jsonl` to drop a line breaks all three at once. A tombstone is one more `appendFileSync`, so the physical row survives and the invariant holds. A sidecar file was rejected too: `integrations-invariants.test.ts:719` pins the store root to exactly three entries.

`pruneSnapshots`' "journal rows always survive" is a **retention** promise inside a retention function, not a promise about a user-requested delete. That reading is recorded in the plan rather than assumed.

**The newest row per client is refused by the server with 409.** `undoable` is computed from the newest row, so deleting it removes the undo entry point a user needs right after a mistake. The GUI hides the button via a `deletable` field, but that is courtesy — the rule lives in the route, because an admin-token caller has no GUI.

## Screenshots

The list, against a live proxy on this branch. The newest entry offers only Undo; every earlier entry gains Delete:

Rollback center with per-row delete

The confirmation, reusing the existing `ConsequenceDialog` rather than adding a component:

Delete confirmation dialog

The dialog names what disappears, what survives, and that the newest entry is kept — deletion is irreversible, so it is never folded into a default.

## Verification

| Command | Result |
|---|---|
| `bun test tests/management-integration-routes.test.ts` | 30 pass |
| `bun test tests/integrations-journal.test.ts` | 27 pass |
| `bun test tests/management-integration-journal-delete.test.ts` | 12 pass |
| `bun test tests/management-route-registry.test.ts` | 13 pass |
| `bun test tests/cli-capabilities.test.ts` | 17 pass |
| `bun test tests/integrations-invariants.test.ts` | 40 pass |
| `cd gui && bun test tests/i18n-locales.test.ts` | 9 pass |
| `cd gui && bun test tests/integrations-rollback-history.test.tsx` | 15 pass |
| `bun run typecheck` / `lint:gui` / `privacy:scan` | exit 0 |

[...]

PR #3474, opened 2026-09-04
feat(catalog,codex): carry Claude combo capabilities and give reset-credit redeems a stable identity

lidge-jun:
## Summary

Two more items from the priority-65 closeout unit. They touch disjoint subsystems and share no files.

**`feat(anthropic)`: carry combo vendor capabilities and provider output budget.** Reimplements #3332 by @full999, which could not be cherry-picked (CONFLICTING/DIRTY). One line differs from the original, and it matters: the PR mapped the vendor table's `metadata.maxTokens` — an OUTPUT ceiling — onto `maxInputTokens`. Since `aggregation.ts:161` takes `Math.min` over member input ceilings, a single Claude member would have dragged a 1M combo down to 128k:

| | member ctx | member maxIn | combo maxIn | combo autoCompact |
|---|---|---|---|---|
| #3332 as written | 1,000,000 | 128,000 | **128,000** | **128,000** |
| this branch | 1,000,000 | 1,000,000 | **1,000,000** | **900,000** |

The auto-compaction budget collapses with it, because `clampAutoCompactTokenLimit` reads `maxInputTokens` as a candidate. Merging the original unchanged would have cost Claude combo users 87% of their input context. `ComboCatalogMemberFallback` already had a `maxOutputTokens` slot and `withFallbackMetadata` already handled it, so the repair is one field name.

**`feat(codex)`: give a manual reset-credit redeem a stable operation identity (#3375 axis D).** The durable ledger was complete and had zero production callers — `openManualResetCreditOperation`, `settleManualResetCreditOperation`, and `markManualResetCreditOperationAmbiguous` were referenced only by their own test file — while the consume endpoint minted a fresh `crypto.randomUUID()` per call. A retry of the same logical redeem looked new to upstream.

Spending a reset credit is irreversible, so the two directions fail differently on purpose. Opening fails closed: `capacity` and `unavailable` return 503 rather than falling back to a random id, because that fallback is precisely the double-spend the identity prevents. Settling fails open: the credit is already gone by then, and reporting a ledger failure would invite a manual retry — the same double-spend from the other side. Omitting `operationId` keeps today's behavior byte for byte.

## Verification

| Command | Result |
|---|---|
| `bun x tsc --noEmit` | exit 0 |
| `bun test tests/codex-catalog.test.ts` | 268 pass / 0 fail |
| `bun test tests/anthropic-reasoning.test.ts` | 67 pass / 0 fail |
[...]

PR #3471, opened 2026-09-04
fix(responses,combos): reject unpaired tool results and fail over provider context caps

lidge-jun:
## Summary

Two independent boundary fixes from the priority-65 closeout unit, plus the plan documents that produced them.

**`fix(responses)`: reject unpaired tool results on the translating path (#3259).** `inputItemSchema`'s permissive catch-all (`schema.ts:106`) accepts a `function_call_output` whose strict schema failed only for a missing `call_id`, and `parser.ts:738` then writes that `undefined` into `toolCallId: string`. Three adapters diverge on the result: `kiro-wire.ts:32` TypeErrors, `ollama-native.ts:334` throws, and `anthropic.ts:775` quietly ships `"[tool_result without adjacent tool_use: undefined]"` upstream — the third one was not in the issue report.

The guard is keyed on the adapter, not on position. A schema-level rejection was the obvious fix and it is wrong: `parseRequest` (`:2812`) runs before the passthrough branch (`:3719`), so rejecting at parse time also kills forward/key passthrough and routed compaction — paths that build from `_rawBody`, never read `context.messages`, and already degrade an unpaired output to `[tool output for unknown call]` on their own. All three candidate designs were implemented and measured against the same 63 files; the schema variant produced a regression the translation-path variant does not.

**`fix(combos)`: fail over provider-specific context caps.** Carried from #3461 by @RHODIZSECURITY. `classifyError` remaps any `"maximum context"` text to `context_length_exceeded`, which the stop list catches, so a combo chain ended at the first target even with a larger-context target queued behind it. The matcher is deliberately narrow: a bare 5059, a generic 400 context refusal, and a generic 413 all still stop.

Carried rather than merged in place because #3461 is a fork PR whose head only ever ran the four gate checks — Cross-platform CI never ran on it, and fork workflows wait on approval, so gate-green was not merge evidence.

## Verification

| Command | Result |
|---|---|
| `bun x tsc --noEmit` | exit 0 (non-vacuity checked: a planted `string = 123` gives TS2322) |
| `bun test tests/responses-parser.test.ts` | 46 pass / 0 fail |
| `bun test tests/responses-compaction-routing.test.ts` | 55 pass / 0 fail |
| `bun test tests/openai-responses-passthrough.test.ts tests/responses-compaction-routing.test.ts` | 180 pass / 0 fail |
| `bun test tests/combos.test.ts` | 56 pass / 0 fail |
[...]

Issue #3375, opened 2026-09-03
[Feature]: complete the OAuth account-pool lifecycle — session affinity, 401/403 rotation, pool health, and stable reset-credit identity


lidge-jun:
### CLI truthfulness slice landed; pool epic remains open

[#3797](https://github.com/lidge-jun/opencodex/pull/3797) (`a92835dbc6`) makes inert generic-pool thresholds explicit in CLI status. This does not complete the broader account-pool lifecycle epic. Existing reset-operation identity work must not be counted as unimplemented solely from the older summary.

Verified against dev `5759d9ea2f1e7281cdc01eb9628f2e0a123fb59c`. The attribution record was added in [#3811](https://github.com/lidge-jun/opencodex/pull/3811).

### Area

Authentication and account pool

### What are you trying to accomplish?

Run OpenCodex against a pool of OAuth/subscription accounts and have it pick the right account, keep a conversation on that account, and recover from per-account failures without operator intervention.

This consolidates the surviving scope from four reports whose first tranche already landed. Each is closed individually and absorbed here so the remaining work stays visible in one place:

- #695 (@luwei1990) — generic OAuth account-pool failover with session affinity and quota-aware selection
- #1062 (@agentHits) — Google Antigravity multi-account management, pool health and quota UX
- #1977 (@dbc-hbin) — pinning the re-initialization window of zero-usage Codex accounts
- #2275 (@luvs01) — durable operation identity for manual reset-credit retries

### What prevents this today?

Partial implementations exist and work, but four concrete gaps remain.

**Session affinity is not reusable.** `src/oauth/generic-account-failover.ts` implements 429 cooldown, rotation, and quota-ranked initial selection (landed in 816f3a159). It does not keep a multi-turn conversation pinned to the account that started it, so a long session can drift across accounts mid-conversation.

**Failover classes are too narrow.** Only 429 rotates. A 401 and a provider-classified 403 are terminal for the pool even when another account in the pool would succeed.

**Pool-level health and attribution are missing.** Per-account quota is surfaced (`src/providers/quota.ts`, `gui/src/components/provider-workspace/ProviderAuthPanel.tsx`), but there is no aggregate pool capacity view and no per-account attribution of usage, so an operator cannot see which account is carrying the load or why a selection was made.

[...]

PR #3485, opened 2026-09-04
docs(devlog): close out the priority-65 unit

lidge-jun:
## Summary

Closes out the priority-65 unit by recording what each work-phase actually produced. Documentation only — no `src/`, `gui/`, or `tests/` changes.

Three PRs landed from this unit:

| PR | squash | contents |
|----|--------|----------|
| #3471 | `4968d0f26` | responses tool-result boundary (#3259) + combo context-cap failover (carried from #3461) |
| #3474 | `00834d710` | Claude combo metadata (carried from #3332) + reset-credit identity (#3375 axis D) + two gate-blocked carries (#3327, #3251) |
| #3477 | `413227888` | rollback journal deletion (part of #3379) |

## What these documents record

The outcome files keep the parts that would otherwise be lost, which are mostly the places where a plan turned out to be wrong:

- **wp3** — the carried PR mapped an OUTPUT ceiling onto an INPUT field, collapsing a 1M combo window to 128k. Verified in three stages, not two: the sniper assertion was confirmed still-red *with the defective mapping applied* before being fixed, so it demonstrably targets the defect rather than the feature.
- **wp4** — the plan said "squash merge #3461". Checking immediately before acting showed it was a fork PR whose head had only the four gate checks; Cross-platform CI had never run on it. Merging on gate-green would have violated this unit's own exact-head-CI criterion, so it became a carry.
- **wp5** — opening the ledger fails closed and settling fails open, because a reset credit is irreversible and the two directions produce double-spend by opposite routes.
- **wp7** — GUI tests were written after the component, so they were proven non-vacuous by reverting the component and confirming exactly four failures. A pre-existing key-set assertion caught the new `deletable` field; the key list was extended rather than the assertion loosened.
- **wp8** — dispositions for the seven items that were examined and deliberately not implemented, each with the evidence that decided it.

## Regression proof

| | SHA | Cross-platform CI |
|---|-----|-------------------|
| session start | `2421e44ce` | **failure** — macos, `tests/codex-auth-context.test.ts:1461` |
| final | `413227888` | **success** |

`dev` was already red when this work started, which is why the baseline was captured first — otherwise "did I break it" is unanswerable. All three squash SHAs verified as ancestors of the final `origin/dev`.

Also recorded: macOS failed twice during this unit and passed on rerun both times, on `tests/codex-shim.test.ts` and `tests/lab-fabric-task.test.ts`. Neither file is in any of these diffs, and both pass locally (77/77 and 49/49). Two timing-sensitive tests are intermittently unstable on the macOS runner; that is worth its own look.

## Verification

`bun run typecheck` exit 0, `bun run privacy:scan` passed. The full local suite was not run at any point in this unit.

## Checklist

- [x] Documentation only; no runtime change
- [x] `bun run privacy:scan` clean
- [x] No GUI change (no screenshot required)

## Summary by CodeRabbit

* **New Features**
  * Added idempotent reset-credit consumption through optional operation IDs in the API and CLI.
  * Added rollback journal entry deletion with confirmation, retention safeguards, and localized messaging.
  * Added model tier outcome details to log tooltips.
  * Improved provider output-budget handling and combo failover for context-limit errors.
* **Bug Fixes**
  * Invalid tool results now receive a clear 400 response on translating routes.
  * Restored provider metadata for combo models without collapsing input capacity.
* **Documentation**
  * Updated management API references with operation ID behavior and expanded error responses.


Ingwannu (changes requested):
The closeout record is useful, but the current exact head `3656bb74ff173a4c2e450bfd001c86c481f61bb2` is not documentation-only relative to the current `dev` base.

GitHub currently reports 64 changed files, 5,941 additions, and runtime/GUI/test changes. The branch forked at `2421e44ce`; it is 17 commits ahead and 13 commits behind current `dev` (`066146980`). Its history carries unsquashed implementations of work that already landed on `dev` through different squash commits, so merging this head would replay or conflict with those implementations despite the PR description stating that no `src/`, `gui/`, or `tests/` files change.

Please rebuild the closeout as a docs-only branch from the latest `dev` (for example, cherry-pick only the intended closeout documentation commits or create one clean documentation commit), then confirm that the GitHub changed-file list contains only the intended `devlog/` records. After the head changes, exact-head CI can establish the final result.

lidge-jun:
Superseded — the branch still carried commits already squashed onto `dev`, so the diff picked up 18 GUI files and armed the screenshot gate on a docs-only change. Reopening from a branch cut fresh off current `dev` with just the six outcome documents.

PR #3507, opened 2026-09-04
fix(tests): arm the home guard before the run lock can throw

lidge-jun:
## Summary

A Windows baseline reported 179 failures. **161 of them were this one bug**, and the count is the least of it.

`tests/preload.ts` acquired the run lock before it armed `OCX_TEST_HOME_GUARD`, with no `try/finally` between them. Taking the lock resolves a user-scoped path, which on Windows spawns PowerShell for the effective SID (`scripts/test-run-lock.ts` → `resolveEffectiveUserIdentity`). Under four-shard load that spawn timed out, the refusal threw straight out of the preload, and every statement below it — the sandbox, the arming, and the assertion that exists to catch exactly this — never ran.

The worker then executed its whole file with the guard down. 161 tests failed asserting "this helper is only available under the repository test preload".

**The real problem is what else the guard was holding back.** `src/lib/windows-elevation.ts:537` and `src/service.ts:952`/`:2653` refuse live elevation and machine-global Task Scheduler mutation *only while armed*. Unguarded, one worker launched a real PowerShell process (pid `18144` where the test expected `launcherPid: null`) and another reached real scheduler registration instead of the expected "refusing to mutate the machine-global Windows Task Scheduler from an armed test process". The guard is what stands between a test run and the developer's own Task Scheduler, and it was being skipped by the failure most likely to happen under load.

## The fix

Reordered to **sandbox → arm + assert → lock**.

Arming earlier is safe because the guard is a deny-list keyed on a path captured at module import (`src/lib/test-home-guard.ts:61-63`), not a "sandbox is present" flag. Its worst case when armed early is refusing a write to the real home — the direction that fails closed.

Putting the lock **last** is the load-bearing part. Arming before the lock but sandboxing after it would still leave a worker armed-but-unsandboxed, able to read the real home. That refinement came from the plan audit, not the original plan.

The lock error stays unswallowed: a run that cannot take the lock must still fail, it just must not fail while unprotected.

## Verification

- `bun run typecheck` — clean
- `bun run test:changed` — 12 pass / 0 fail
- `bun test tests/test-home-guard.test.ts tests/gui-static.test.ts tests/server-management-auth.test.ts` — 52 pass / 0 fail
- Ordering confirmed on the affected Windows host itself (read-only, non-contending with the suite running there): `sandbox 1414, arm 2969, assert 3228, lock 4168 → ORDER_OK true`

[...]

PR #3514, opened 2026-09-04
docs(devlog): close out the admin-token unit with the preload guard fix

lidge-jun:
## Summary

Closes out the `260905_admin_token_local_ux` devlog unit now that the Windows carry-forward it named has landed.

Records PR #3507 (`663fdbb0a`) against carry-forward item 1 of the baseline triage, and adds `031` for why that fix was worth doing independently of the 161 failures it removed:

> Those 161 were the *visible* consequence. `src/lib/windows-elevation.ts` and `src/service.ts` refuse live elevation and machine-global Task Scheduler mutation **only while the guard is armed**. A worker that lost the guard did not merely fail loudly — it quietly gained the ability to do the things those refusals exist to prevent, and two suites took it. A run that had been slightly luckier with its SID lookup would have shown fewer failures and done the same writes.

`031` also records what the plan audit changed. The original plan was arm-then-lock; the reviewer rejected it as incomplete, because arming before the lock while sandboxing after it still leaves a worker armed-but-unsandboxed — writes blocked, reads not. Shipped order is sandbox → arm + assert → lock.

Delivery record gains rows for #3504 and #3507. Items 2 and 3 (the `multi-account auth store` failures, 22 of the surviving 25, and the `keep-native-v1` argv assertion) stay open with their real counts rather than being quietly dropped.

## Verification

`bun run privacy:scan` — passed. Documentation only; nothing in the build, typecheck, or test path reads from `devlog/`. No pre-disclosure security material: every defect described is already public in a merged diff, which is the `AGENTS.md` test for what may live here.

## Checklist

- [x] Documentation only; no code paths touched
- [x] No credential, token, or request-body values recorded
- [x] Targets `dev`

### DIFF
--- a/devlog/_plan/260905_admin_token_local_ux/020_dialog_repair.md
+++ b/devlog/_plan/260905_admin_token_local_ux/020_dialog_repair.md
@@ -61,3 +61,13 @@ The dialog now carries a help paragraph plus a link to a new
 (`target="_blank"`, `rel="noreferrer"`, accent colour) per
 `gui/src/pages/dashboard-dialogs.tsx`. Copy landed in all nine locales;
 `gui/tests/i18n-locales.test.ts` enforces key parity.
+
+## Landed
+
+- #3483: PR [#3491](https://github.com/lidge-jun/opencodex/pull/3491), `85e42117c`.
+- #3353: PR [#3493](https://github.com/lidge-jun/opencodex/pull/3493), `8b961b198`.
+
+Both issues are closed with landing comments naming the mechanism rather than
+just the fix, because in both cases the mechanism is the non-obvious part: a
+cascade origin rule for one, and a prompt that could not have helped for the
+other.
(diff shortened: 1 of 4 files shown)


Ingwannu (changes requested):
The recorded outcome is useful and the commit references are correct, but the PR currently declares this unit closed while leaving every file under `devlog/_plan/260905_admin_token_local_ux/`. Current `AGENTS.md` defines `_plan` as open units and `_fin` as closed units moved after a terminal outcome; the security-notes section also says a shipped-fix write-up belongs in `_fin`. Please move the complete unit directory to `devlog/_fin/260905_admin_token_local_ux/` in this PR and update any live references that must follow it. The two remaining Windows observations can stay explicit carry-forward items or become separate open units, but they should not make this closed admin-token unit advertise two contradictory states.

PR #3568, opened 2026-09-05
fix(responses): recover native messages for Go subagents

voiys:
## Summary

Fix encrypted messages from native ChatGPT parents to routed Go subagents. Opt-in task recovery can admit the initial assignment, but an encrypted `send_message` fails with `unreadable_encrypted_agent_task`; later tool turns can also replay the original ciphertext.

- Accept `MESSAGE` alongside `NEW_TASK` in the existing opt-in recovery path. Restore previously admitted cached plaintext during replay, preserving caller/account checks, routing identities, message-type separation, cancellation and expiry.
- Normalize plaintext Go Responses `agent_message` items into supported user messages, preserving content parts and readable sender/recipient context. Leave ciphertext, unknown content and other destinations unchanged. This text is model context, not a new machine-readable routing protocol.
- Test mixed histories containing an initial task and distinct follow-up messages. Cached restoration covers known history; fresh recovery still handles only the current tail message. Unseen historical ciphertext remains unchanged.

This is the message-delivery half of the split requested in the review. Catalog efforts and picker ordering are removed from this PR and live in #3571. The branch reuses the session-affinity header implementation already on `dev`; no duplicate header code, personal configuration or application-bundle changes are included.

## Verification

Refreshed onto `dev` at `6585e6a70`; the feature patch and review follow-ups replayed without manual conflict resolution. Typecheck, privacy scan, docs build and the full test suite were rerun on the refreshed branch. Earlier focused and live checks are recorded separately below.

- `bun run typecheck` — passed.
- `bun run privacy:scan` — passed.
- `bun test tests/server/server-agent-task-recovery-replay.test.ts tests/providers/opencode-go-agent-messages.test.ts` — 12 passed, 0 failed, 69 assertions.
- `bun run test` — full runner passed: 19,287 passed, 14 skipped, 0 failed, including test-layout guards.
- `bun run build` in `docs-site` — passed; 425 pages, rebuilt after refreshing the branch onto `dev`.
- `git diff --check origin/dev` — passed.
- Live verification of the equivalent local 2.42.0 patch: native Astra spawned DeepSeek V4 Flash Vision Exp at `max`, then sent a random code only through `send_message`. Saved parent/child transcripts confirmed encrypted `NEW_TASK` and `MESSAGE` delivery and an exact-code child reply. The previously failing encrypted message also recovered successfully.

[...]

PR #3571, opened 2026-09-05
fix(codex): preserve Go efforts and native-first picker ordering

voiys:
## Summary

Keep configured OpenCode Go reasoning choices exact and let native Codex models appear before routed models without changing subagent selection. Catalog sync currently appends synthetic effort choices, and routed-only picker ordering cannot express a native-first list independently of the subagent roster.

- Preserve configured reasoning ladders for the canonical `opencode-go` provider during entry generation and catalog merge. Other providers retain their existing behavior.
- Let `modelPickerOrder` containing a bare native id order the complete picker by exact catalog ids. Keep natural spawn priorities separate, and recompute native spawn rank when featured models change.
- Ignore empty and whitespace-only ordering entries throughout generation and merge, and preserve raw/encoded routed ID compatibility with exact-match precedence.
- Preserve existing unset, empty and routed-only ordering behavior. Document the migration: existing lists containing previously ignored bare native ids now opt into complete-picker ordering; remove those ids to retain the old behavior.
- Restore retained rows after switching from complete ordering to empty or routed-only ordering during a provider outage. Reapply the current display order without changing freshly built rows.
- Cover exact native-versus-routed matching, repeated sync, changed featured models, effort preservation and the v2 subagent roster. Regression tests compare healthy and degraded refreshes, including multiple account selectors and a nonzero ordering index. Update both English and French ordering guides.

Split from #3568 following review. This PR contains only catalog behavior and docs; message recovery stays in #3568. No personal model list, credentials, provider endpoint changes or application-bundle modifications are included.

## Review follow-ups

- Recompute retained routed spawn ranks from the current featured roster during degraded discovery, including promotion, demotion, reordering, raw/encoded IDs and multiple accounts. Healthy and degraded results agree and repeated merges remain stable.
- When a native baseline is absent, use saved natural priority before display priority so repeated syncs cannot accumulate display offsets.
- Ignore whitespace-only ordering entries. A fixed `SUPPORTED_NATIVE_OPENAI_SLUGS` gate was not adopted: the documented opt-in accepts bare catalog IDs, including newly discovered native models not yet in the pinned roster. Requiring the pinned set would narrow that contract.
- Mark the provider-effort entries as examples and link to OpenCode Go's current model roster instead of duplicating a changing full roster here.

## Verification

Refreshed onto `dev` at `71edeec88`. Resolved the append-only provider-docs conflict by retaining both the newly merged agent-message section and this PR's reasoning-effort section. Runtime patches replayed unchanged. Typecheck, privacy scan, docs build and the full test suite were rerun on the refreshed branch. Earlier focused and live checks are recorded separately below.

- TypeScript check including `catalog-full-picker-order.test.ts` with the root compiler settings — passed; corrected the warning-policy fixture value and fixture types.
- Focused catalog suites — 425 passed, 0 failed before the final full-suite run.
- `bun run test` — full runner passed: 19,548 passed, 15 skipped, 0 failed.
- `bun run build` in `docs-site` — passed; 425 pages.
- `git diff --check origin/dev` — passed.
- Bundled Codex app-server checks of the equivalent local 2.42.0 patch confirmed native-first ordering, readable Go labels, exact configured efforts and the five-model Go v2 roster after startup sync and management refresh.

Live checks exercised the local release patch; this port to `dev` has the automated coverage above. Effort preservation follows the configured ladder for the canonical `opencode-go` provider; it does not discover or validate provider capabilities at runtime.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - `modelPickerOrder` can now arrange the complete model selector, including native and featured models, with listed models shown first.
  - Exact catalog and account-qualified identifiers are supported, while display ordering remains separate from subagent selection priorities.
  - OpenCode Go models now preserve their configured reasoning-effort options without unrelated fallback tiers.

- **Documentation**
  - Updated English and French guides and provider references with ordering behavior, identifier matching, migration guidance, and Go reasoning-effort details.

- **Tests**
  - Added coverage for full picker ordering, degraded discovery, and exact reasoning-effort preservation.


lidge-jun:
## 리뷰 · 우선순위 66 / 80

이 PR은 #3568에서 **카탈로그만** 떼어 낸 조각입니다. 지금 `dev` HEAD(`24cc558d5`, package 2.43.0)의 `src/codex/catalog/sync.ts`는 카탈로그 sync 때 reasoning ladder에 합성 `max`/`ultra`를 붙이는 경로가 있고, `modelPickerOrder`는 routed-only display로 남아 있어 bare native id는 무시됩니다. OpenCode Go 행은 upstream이 모델마다 다른 enum을 쓰는데, 합성 tier가 붙으면 서브에이전트 override 화면이 거짓 선택지를 보여 줍니다. 또 네이티브를 피커 맨 위에 두고 싶어도 `modelPickerOrder`만으로는 전체 순서를 못 바꿉니다.

고치는 축은 두 갈래입니다. (1) `deriveEntry`에서 `preserveExactReasoning = preserveExact || provider === "opencode-go"`로 Go는 설정한 ladder만 쓰고, merge 경로에서도 `opencode-go/` slug는 합성 `max`를 다시 붙이지 않습니다. 다른 프로바이더는 예전처럼 virtual tier를 유지합니다. (2) 새 `applyFullModelPickerOrder`는 order 안에 `/` 없는 bare native id가 하나라도 있으면 피커 전체 `priority`를 재정렬하고, 원래 순위는 `opencodex_spawn_priority`에 남겨 spawn 후보는 건드리지 않습니다. featured가 바뀌면 보존 행에서 spawn 필드를 지우고 다시 계산합니다. EN/FR ordering 가이드와 providers 참고 문서에 마이그레이션 노트가 있고, unset/empty/routed-only는 예전 동작을 유지한다고 명시합니다.

현재 `dev`와의 관계는 분명합니다. #3568 리뷰에서 권했던 “catalog·picker를 recovery와 분리”를 그대로 따랐고, base는 `dev`입니다. 새 테스트 `catalog-go-exact-efforts`·`catalog-full-picker-order`가 exact matching·반복 sync·featured 변경·routed-only 유지를 고정합니다. 작성자 주장대로 typecheck·privacy:scan·full suite 18027 pass·docs-site build가 있다고 하니, hosted CI만 초록이면 merge 후보입니다. types/config 스플릿 무효화·close-don't-rebase 대상은 아닙니다. JSDoc만 `modelPickerOrder` 의미를 바꿉니다.

```src/codex/catalog/sync.ts``` `deriveEntry` - `preserveExactReasoning`이 Go provider에 exact ladder를 겁니다. 합성 tier가 서브에이전트 화면을 속이지 않게 하려는 의도와 맞습니다.

```src/codex/catalog/sync.ts``` merge 경로 `opencode-go/` prefix 가드 - derive는 `provider === "opencode-go"`이고 merge는 slug prefix입니다. 보통은 같지만, slug만 다른 이상 행이 있으면 한쪽만 exact가 됩니다. 같은 조건식으로 맞추면 더 안전합니다.

```src/codex/catalog/sync.ts``` `applyFullModelPickerOrder` - bare native 판별이 `!slug.includes("/")`입니다. account-qualified native만 있는 리스트는 complete-picker로 안 들어갑니다. 문서 약속과 같지만, 실수 설정이 조용히 routed-only로 남을 수 있습니다.

```src/types/config.ts``` `modelPickerOrder` JSDoc - 새 계약(bare native면 전체 피커)로 고쳐졌습니다. 예전 “routed-only, native는 subagentModels로” 불변식은 폐기됩니다. 마이그레이션 노트는 가이드에 있으니 릴리즈 노트에도 한 줄이 있으면 좋습니다.

경로 docs EN/FR model-ordering - 마이그레이션·exact id 매칭·spawn 후보 불변을 잘 설명합니다. FR 앵커가 EN 해시와 어긋나지 않는지만 확인하면 됩니다.

경로 #3568 - 같은 catalog 변경이 아직 남아 있으면 이 PR 랜딩 뒤 #3568에서 catalog hunk를 빼야 충돌/중복이 없습니다.

메인테이너의 판단이 필요한 지점
- `modelPickerOrder`에 bare native id가 있으면 전체 피커를 재정렬하는 새 계약을 정식으로 받아들일지
- #3568에 남은 catalog 조각을 이 PR 머지 후 바로 정리할지
- Go exact ladder를 provider 필드와 slug prefix 중 하나로 통일할지

너의 추천
hosted CI가 초록이면 **머지하세요**. #3568보다 먼저 올려도 됩니다. 랜딩 직후 #3568에서 catalog/picker hunk를 제거하고 recovery·Go message normalize만 남기라고 작성자 @voiys에게 한 줄 요청하면 됩니다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
The OpenCode Go exact-effort half is directionally sound, but the complete-picker transition is not safe on this head.

Blocking issue: `applyFullModelPickerOrder` returns immediately when the new order is empty or routed-only. A row retained from a degraded provider can already carry a full-order display `priority` plus `opencodex_spawn_priority` from the previous sync. Changing, for example, `["gpt-5.6-sol", "provider/model"]` to `["provider/model"]` or `[]` therefore leaves that retained row at the old complete-picker rank instead of restoring its natural priority and applying the current mode. The existing test only calls the routed-only mode on a fresh row, so it cannot catch this state transition.

Please restore retained rows from `SPAWN_PRIORITY_FIELD` and clear the marker before the complete-order no-op/routed-only path, then apply the current routed-only ordering rules. Add degraded-provider merge regressions for full -> routed-only and full -> empty transitions. Please also document the exact upstream OpenCode Go model IDs beside the friendly model names; that documentation point is non-blocking but makes the configuration examples actionable.

Once the transition tests pass on the exact head and hosted CI is green, I can re-review the focused fix.

Ingwannu (changes requested):
The original degraded-retention blocker is fixed on `7e3402310`; the new transition tests pass locally. Two remaining compatibility edges still need a small follow-up before approval:

1. Normalize the order before both activation sites. `buildCatalogEntriesFromObservedState` filters empty strings, but `applyFullModelPickerOrder` and the retained-row branch test the raw array. A hand-edited `["", "provider/model"]` therefore activates complete ordering even though the builder discarded the empty entry.
2. Preserve the existing raw/encoded slug compatibility. Routed-only ordering accepts both the public encoded slug and the raw provider plus slash-containing model id through `pickerOrderPriority(slug, altSlug)`. Once a bare native id enables full ordering, the final pass looks up only `entry.slug`, overwrites the builder priority, and loses the configured position for upstream model IDs containing `/`. Resolve exact matches first, then the established slug-equivalence relation, and add a native-first regression with a slash-containing native model id.

Please also use the valid `warningPolicy: "suppress"` test value instead of the out-of-contract `"silent"`; Bun executes it, but root typecheck does not typecheck test files, so the typo currently passes silently.

Validation on this head: focused catalog suites 414/414 and root typecheck pass. The remaining request is about the uncovered transition inputs, not the prior fix.

voiys:
@Ingwannu sry, missed that transition case. fixed it and added tests for both removing the order and switching back to routed-only during a provider outage. full suite passes on the new commit. thanks for catching it 🙏

Ingwannu (approved):
Approved on exact head `339371165ea05142c399afea1f0dcff8a496a0c1`.

The follow-up resolves all remaining requested boundaries:

- empty entries are filtered before both full-order activation paths;
- exact catalog ids retain precedence while raw/encoded routed spellings use the existing equivalence key;
- retained degraded rows restore their natural priority before routed-only or empty ordering is reapplied;
- the test fixture now uses the valid `warningPolicy: "suppress"` contract;
- regressions cover full-to-empty, full-to-routed-only, and slash-containing upstream ids.

Focused catalog validation on this exact head: 418 pass / 0 fail. Root typecheck also passes. CodeRabbit and hosted cross-platform CI must still complete green before merge.

lidge-jun:
Superseded by #3700, now merged into dev as 76356176c86aa123220c82b65321453e81897405. The latest source head 09acfba64596011c308f0d9cbac070123bb9faeb was audited as the rebased equivalent of the carried source; no author behavior is missing. The source author and Co-authored-by credit are preserved. Exact-head CI and the actual current-dev merge composition passed verification.

PR #3600, opened 2026-09-05
test(oauth): exercise the account-removal path the quorum test is named for

lidge-jun:
## Summary

Follow-up to #3530. The restored contract test `removing an account invalidates immediately, not after the TTL` never removed an account: it cleared session affinity, marked the store unread, and asserted only that the store was re-read. With two accounts still in the roster the predicate could not observe the transition the test is named for, so deleting the `removeAccount` call from the DELETE route would have left it green. The test now mirrors the route order (`removeAccount` first, then `clearAnthropicSessionAffinityForAccount`) and asserts quorum is actually `false` afterward.

Unit: `devlog/_plan/260905_open_work_closeout/` (050 E0, 051).

## Verification

- `bun test tests/routing/anthropic-quorum-cache.test.ts` — 7 pass / 0 fail. RED probe: with the new `removeAccount` line commented out the new `toBe(false)` assertion fails (6 pass / 1 fail), proving the assertion observes the roster reduction; the sibling "manual account selection invalidates immediately" test still passes with two accounts intact.
- `bun run typecheck` — exit 0.
- Test-only change; hosted CI on the final dev tip is the batch gate per maintainer instruction.

## Checklist

- [x] Targets `dev`
- [x] Test proven RED without the behavior it guards
- [x] No runtime change

## Summary by CodeRabbit

- **Tests**
  - Updated routing tests to verify that removing an account immediately invalidates the associated quorum cache.
  - Test behavior now matches the account deletion flow more closely.

### DIFF
--- a/tests/routing/anthropic-quorum-cache.test.ts
+++ b/tests/routing/anthropic-quorum-cache.test.ts
@@ -25,7 +25,7 @@ import {
   resetAnthropicRoutingForManualSelection,
   rotateAnthropicAccountOn429,
 } from "../../src/oauth/anthropic-routing";
-import { getAccountSet, markAccountNeedsReauth, saveCredential } from "../../src/oauth/store";
+import { getAccountSet, markAccountNeedsReauth, removeAccount, saveCredential } from "../../src/oauth/store";
 import { removeTreeWithRetry } from "../helpers/remove-tree";
 
 const originalHome = process.env.OPENCODEX_HOME;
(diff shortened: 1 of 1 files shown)


lidge-jun:
Maintainer admin merge (ruleset bypass recorded per MAINTAINERS.md): test-only follow-up to #3530; 7/0 locally with a RED probe; per maintainer instruction the final dev-tip CI run is the acceptance gate for this batch. wp5 E0 of the 260905 open-work closeout.

Issue #3735, opened 2026-09-06
[Responses] Codex task-input envelopes are rejected by translated adapters


yrlan-montagnier:
### Client or integration

Codex App

### Provider or upstream service

Translated Responses adapters, reproduced with Kiro

### OpenCodex version

2.44.0 / af344a28eabcee09a5e04c48ab897449792719c2

### Endpoint or capability

POST /v1/responses, Codex task coordination

### Current behaviour

Codex task coordination can add a result-shaped external input item to the destination task. The observed envelope has type function_call_output, non-empty id, name, and namespace fields, a representable output, and no call_id field. It represents input supplied by another task rather than the result of a tool call made in the destination task.

The common Responses parser currently treats this envelope as a standard tool result. The translated-adapter guard added by #3471 then correctly sees a missing tool identity and returns HTTP 400 before calling the upstream adapter. This prevents Codex from creating or resuming tasks through translated adapters.

### Expected behaviour

Only the complete external task-input envelope described above should be converted to user input before adapter translation. Standard function_call_output or custom_tool_call_output items without a non-empty call_id must continue returning HTTP 400, as deliberately implemented by #3471. Items with call_id remain tool results, and incomplete or malformed external envelopes must also remain invalid.

### Minimal redacted request or reproduction

Submit a Responses request through a translated adapter with this item in input:

~~~json
{
  "type": "function_call_output",
  "id": "fco_redacted",
  "name": "send_message_to_thread",
  "namespace": "codex_app",
  "output": "<redacted external task input>"
}
~~~

The destination task fails before the provider is called.

### Actual response or error

~~~text
HTTP 400
tool result requires a non-empty string call_id
~~~

### Upstream documentation

OpenAI Responses API reference: https://platform.openai.com/docs/api-reference/responses/create

The public Responses contract uses call_id to associate actual function tool outputs with calls. No public specification was found for Codex App's inter-task input envelope; that envelope is a concrete Codex App client requirement observed during task creation, follow-up messaging, and task resumption.

### Suggested mapping or implementation notes

Preserve the translated-adapter guard from #3471. Before producing common parser messages, recognize only a function_call_output that has no call_id property, has non-empty string id, name, and namespace fields, and has representable output content. Convert that narrowly identified external item to user input, preserving its supported content exactly and without adding an orphan-tool marker such as `[tool output for unknown call]`.

[...]

PR #4051, opened 2026-09-08
wp2: land seven contributor bug fixes (quota, catalog, streaming, storage, history)

lidge-jun:
## Summary

Second sequential integration branch of the 2.49.0 backlog closeout. Seven reviewed bug fixes from four contributors, carried as squash commits that each keep a `Co-authored-by` trailer. Stacked on `dev` after #4049.

- **#4018** (@cb8010d6) — a Spark model window with a sub-day duration was dropped entirely: `sparkWeekly` was the only selector and it excludes short windows, so a payload carrying `GPT-5.3-Codex-Spark` 5h + weekly parsed to the weekly row alone. Both rows now parse. The 5h row is filtered by the same `showCodexSparkQuota` gate as the weekly one, which defaults off, so a default-config account card is byte-identical.
- **#4008** (@cb8010d6) — `mergeAccountQuota` carried weekly, monthly, short, and reset credits forward across a partial header update but had no `else` branch for `customWindows`, so any ordinary response with rate-limit headers erased a stored Spark window. The matching carry-forward is added. Composes with #4018: without it the new 5h row would vanish on the first response.
- **#3981** (@yansigit) — a catalog or models-cache write left the app-server observation memo (a 5s local `ps`+`stat` cache) serving pre-write state. Invalidation is added at the two atomic write sites and after the sync admission gate. No provider call is involved; the cost of a dropped entry is one cold local probe.
- **#3979** (@yansigit) — the inactivity timer stayed armed after a terminal event, so the post-terminal drain could race a false `RoutedModelInactivityError`. One `clearInactivity()` after the terminal is held; the drain keeps its own bounded timeout.
- **#3964** (@ildunari) — direct `api.meta.ai` returns 400 for Muse web_search fields that the existing strip list already covers for Zen. The Meta Responses URL is added to that allowlist. Muse-via-router traffic is unchanged.
[...]

PR #3914, opened 2026-09-07
feat(sponsors): OrcaRouter placement, overview introduction and links

lidge-jun:
## Summary

OrcaRouter is the first Standard sponsor under [SPONSORS.md](https://github.com/lidge-jun/opencodex/blob/dev/SPONSORS.md) (agreement completed 2026-09-07). This PR delivers the placement side of that agreement and the mechanism the rule set promised.

**Mechanism (shared with the PackyCode branch):** a registry `sponsor: { tier, url }` field surfaces on the derived preset as `sponsor` / `sponsorUrl`. The dashboard Add provider catalog pins sponsor rows first — Main before Standard, alphabetical by label within a tier, an order no sponsor can buy — and draws a `Sponsor` chip before the auth badge. `ocx provider presets` prints the tier. Derived preset order stays registry order; routing, failover, quota, and defaults never read the field.

**OrcaRouter:** the existing `orcarouter` preset (openai-chat, `https://api.orcarouter.ai/v1`) gets the sponsor field; README gets the first Standard row (wordmark from orcarouter.ai, blurb drawn from their public product page pending the sponsor's own copy); the providers guide gets a paragraph. PKCE login arrives separately in #3908 and is untouched here.

### gui — Add provider picker (Paid tab)

OrcaRouter pinned first with the Sponsor chip in the Add provider picker

### README Sponsors section

README Sponsors section with the OrcaRouter row

### Provider overview — introduction and links

A compact, localized sponsor introduction now sits above connection details, with an explicit Sponsor label and direct provider links. The existing referral URL is preserved; identical destinations are shown once. Editable notes appear once in the wider column, while statistics and account quota stay separate. Custom endpoints do not inherit sponsor branding. Narrow tabs keep their labels on one line.

The following are **screenshot mockups rendered from the actual Providers page**, with intercepted fixture responses and synthetic usage (362 requests / 3.2M tokens), not live sponsor account measurements.

orcarouter provider overview — desktop

<details>
<summary>Korean mobile overview (390px)</summary>

orcarouter Korean mobile provider overview

</details>

## Verification

- Root `bun run typecheck`, `bun run privacy:scan` — passed.
- GUI `bun run lint:i18n`, `bun run lint`, `bun run build` — passed. Existing large-bundle advisory remains.
- Focused sponsor matching/render, note-save, catalog ordering, locale parity and French localization tests — passed. Fixed the pre-existing French Sponsor label failure without weakening its test.
[...]

PR #4058, opened 2026-09-08
wp4: fix four reported bugs with no owning pull request

### LINKED ISSUE #3807: [Bug] 2.43.0 unpaired-tool-result guard rejects Codex desktop sub-agent seed shape: every routed-model delegated thread dies instantly with 400 "tool result requires a non-empty string call_id"
### Client or integration

Codex App

### Area

Proxy and routing

### Summary

The unpaired-tool-result guard added by PR #3471 (for issue #3259) in `src/server/responses/core.ts` rejects, for all non-passthrough adapters, any request whose parsed history contains a tool-result item without a non-empty string `call_id`. The proxy answers HTTP 400 `tool result requires a non-empty string call_id`, so the request never reaches the provider.

Codex desktop seeds every agent-created (sub-agent) thread's first model request with exactly such an item: a lone `function_call_output` (name `create_thread`, output = the `<codex_delegation>` prompt) that carries `id`/`name`/`namespace`/`output` but no `call_id` field at all. The identical seed shape appears in both surviving and dead child rollouts across app-server CLI 0.152.0 and 0.153.4.

Net effect behind a local OpenCodex proxy: sub-agents on routed models (opencode-go, zai, meta, google, any translating adapter) die in under one second on their first turn. Retries, resumes, and fresh re-dispatches all reproduce it, so sub-agent delegation is 100% unusable on routed models. Sub-agents on `openai-responses` passthrough providers are unaffected because the guard is keyed on non-passthrough adapters, which makes the breakage look model-specific from the client side. The same guard also kills a long-running routed child mid-session once its history re-includes the seed item.

### Reproduction

Minimal probe, no Codex desktop needed. While any translating-adapter provider is configured, send a tool-result item with no `call_id`:

```
curl -s http://127.0.0.1:10101/v1/responses \
  -d '{"model":"<any-routed-model>","input":[{"type":"function_call_output","output":"x"}],"max_output_tokens":1}'
```

Response:

```
{"error":{"message":"tool result requires a non-empty string call_id","type":"invalid_request_error","code":"invalid_request_error"}}
```

Adding any non-empty `"call_id":"call_x"` to the same item returns HTTP 200, so the check is emptiness-only, not actual pairing.

End-to-end steps:

1. Upgrade the proxy to 2.43.0 from 2.41.0 (the guard is absent in 2.41.0; zero matching 400s on any earlier day in `usage.jsonl`).
2. In Codex desktop on a routed-model thread, dispatch a sub-agent via `create_thread` with no model override so the child inherits the routed model.
3. The child's first turn fails in under 1 second with the exact 400 above; `wait_threads` reports `status: systemError`.

Counted exposure: 90 matching 400s through 2026-09-06 13:36 ET across 8 routed models (qwen3.8-flash 68, glm-5.3 6, muse-spark-1.3-contributor 5, deepseek-v4-pro 3, glm-5.3-flash 2, gemini-3.8-flash 2, deepseek-v4-flash 2, gemini-3.7-flash 2). Control: passthrough-model child threads created in the same window do not hit the error.

### Version

2.43.0

### Operating system

Windows 11 25H2 (build 26200)

### Provider and model

opencode-go / qwen3.8-flash (also zai, meta, google translating adapters; `openai-responses` passthrough unaffected)

### Logs or error output

```
{"error":{"message":"tool result requires a non-empty string call_id","type":"invalid_request_error","code":"invalid_request_error"}}
```

`usage.jsonl` shows the first occurrence about 28 minutes after the 2.41.0 to 2.43.0 self-update on 2026-09-05, then 90 matching entries through 2026-09-06 across 8 routed models. Zero matching occurrences on any earlier day.

No response

### Redacted configuration

```json
{
  "providers": ["opencode-go", "zai", "meta", "google"],
  "adapter_family": "openai-chat translating adapters",
  "unaffected": "openai-responses passthrough provider",
  "client": "Codex desktop, app-server CLI 0.153.4"
}
```

Suggested fix. The guard's own comment notes the reverse case is repaired, not rejected (`chat/outbound.ts` synthesizes `call_{uuid}` when a tool call lacks an id). Options, in order of preference:

1. Repair instead of reject on the translating path: synthesize a placeholder `call_id` for unpaired tool results and optionally log a diagnostic, mirroring the reverse-case repair. This still fixes the actual failure mode from #3259 (a literal `undefined` reaching the wire).
2. Or scope the hard rejection to interior unpaired results (a real pairing bug mid-conversation) and tolerate or repair a leading unpaired `function_call_output`.
3. Or make it configurable (for example `rejectUnpairedToolResults: "repair" | "reject"`) with `repair` as the default.

Workaround currently in use: dispatch sub-agents with an explicit passthrough/OpenAI-hosted model override on `create_thread` until the guard is repaired or made configurable.

Happy to re-test any candidate fix against the live proxy; the curl probe above is exact.

### LINKED ISSUE #4023: [Bug][macOS][Dashboard] Stop button can unload launchd service before native Codex teardown completes
### Client or integration

OpenCodex dashboard with Codex Desktop.

### Area

Dashboard / service lifecycle / Codex config restore.

### Summary

On macOS with the launchd background service installed, pressing the dashboard Stop proxy button stopped the proxy and made the dashboard unreachable, but left Codex routed to `http://127.0.0.1:10100/v1`. Codex then could not send requests until the proxy was manually started again.

The current `POST /api/stop` control flow appears vulnerable to terminating its own process before shared teardown completes:

1. `stopServiceIfInstalledDetailed()` is called.
2. On macOS it synchronously invokes `launchctl unload` for the LaunchAgent.
3. Only after that returns does the handler await `performStopTeardown()`, which calls the native Codex restore path.

Because the API handler runs inside the launchd-managed proxy process, unloading its own job can terminate it before step 3 finishes. This matches the observed state: the listener and dashboard were gone, while `openai_base_url`, `experimental_realtime_ws_base_url`, and `model_catalog_json` remained injected in the Codex config.

This ordering is still present in tag `v2.48.0` at `src/server/management-api.ts` around lines 275-348 and `src/service.ts` around lines 2351 and 3927-3932.

Expected: dashboard Stop must restore native Codex and verify the restore before allowing launchd to terminate the proxy, or delegate the complete stop to an external process that survives unloading the service. If safe completion cannot be guaranteed, the dashboard should refuse and instruct the user to run `ocx stop`, similar to the existing Task Scheduler protection.

### Reproduction

1. On macOS, install and run OpenCodex as a launchd background service.
2. Confirm Codex routing is injected and the dashboard is available on port 10100.
3. Press Stop proxy in the dashboard and confirm.
4. Observe that the proxy and dashboard stop.
5. Inspect the effective Codex `config.toml`.
6. Observe that the loopback routing keys may remain, leaving Codex pointed to the stopped proxy.
7. Start the proxy again; Codex requests work again because the stale route becomes reachable.

### Version

Observed on 2.42.0. Source ordering verified unchanged in tag 2.48.0.

### Operating system

macOS, Apple Silicon, launchd user service.

### Provider and model

Not provider-specific.

### Logs or error output

After manually starting the proxy again, the service log reported that the previous session did not shut down cleanly and restored Codex state from the journal. There was no successful native-restore message associated with the dashboard stop.

Relevant source:
- https://github.com/lidge-jun/opencodex/blob/v2.48.0/src/server/management-api.ts#L275-L348
- https://github.com/lidge-jun/opencodex/blob/v2.48.0/src/service.ts#L2351
- https://github.com/lidge-jun/opencodex/blob/v2.48.0/src/service.ts#L3927-L3932

Related but not duplicate:
- #1798 covers Codex rewriting the config after injection and marker/hash-based restore failure.
- #3008 covers Windows dashboard update/stop lifecycle behavior.

### Redacted configuration

```toml
experimental_realtime_ws_base_url = "http://127.0.0.1:10100/v1"
model_catalog_json = "/redacted/.codex/opencodex-catalog.json"
```

### Checks

- [x] I searched existing open and closed issues.
- [x] I removed secrets, tokens, account details, credentials, and personal paths.

### LINKED ISSUE #4032: Chained clients (provider hub) drop per-model context windows the hub already serves — every routed row materializes at the 128k fallback
### Client or integration

Other

### Area

Catalog / models

### Summary

A chained client (single provider `hub` pointing at a remote hub data plane, `liveModels: true` — the provider-based chained setup, not `ocx connect`) materializes **every** hub-routed row with `context_window: 128000` — the parser's compatibility fallback — even though the hub already serves correct per-model windows on two surfaces. Effect: Codex compacts at ~121.6k effective tokens on 1M-class models (cursor GPT-5.6, zai GLM-5.3), silently discarding ~87% of usable context on chained machines.

### Reproduction

1. Hub 2.48.0, `runtimeRole: hub`, per-provider `modelContextWindows` configured. The hub's own surfaces carry the truth:
   - `GET /v1/catalog` (the shared serializer used by the dashboard and the data-plane route): all 436 entries have `context_window > 0`, e.g. `"slug":"cursor/gpt-5.6-sol","context_window":1000000`, `"slug":"zai/glm-5.3-flash","context_window":1000000`, `"slug":"alibaba-token-plan-intl/qwen3.8-max","context_window":983616`.
   - `GET /v1/models`: `"id":"cursor/gpt-5.6-sol","capabilities":{"context_length":922000,"max_output_tokens":128000,...}` — per-model metadata is present here too.
2. Client 2.48.0: single provider `hub` (`baseUrl` → hub data plane `/v1`, `liveModels: true`, `allowPrivateNetwork: true`), `ocx service install` + `ocx sync`.
3. Client catalog `~/.codex/opencodex-catalog.json`: **every** `hub/*` row = `"context_window": 128000, "max_context_window": 128000`. `hub/cursor-gpt-5.6-sol` 128000 (real 1,000,000), `hub/zai-glm-5.3-flash` 128000 (real 1,000,000), `hub/alibaba-token-plan-intl-qwen3.8-max` 128000 (real 983,616). Only rows backed by the client's own local `openai` forward provider keep real roster values (272000) — the loss is specific to routed rows.
4. Where it drops: routed-row assembly never maps the hub-served metadata into the discovered window (`applyProviderConfigHints` in `src/codex/catalog/provider-fetch.ts` sees no discovered value), so `src/codex/catalog/parsing.ts` applies its `128000` compatibility fallback at materialization.
5. Workaround we run today (effective, but per-client and manual): set `providers.hub.modelContextWindows` on each client, keyed by the **`/v1/models` id space**, plus `modelMaxInputTokens`. `configuredContextWindow()` fills the value because discovery is absent, and the gather fingerprint (`providerCatalogFingerprint` includes `ctxW`) invalidates the cache so the next `ocx sync` picks it up. Result: 227/449 rows corrected on one client, 220/441 on another; a streamed call through `hub/gpt-5.6-luna` completes.

Papercut discovered while keying that map: the two hub surfaces disagree on id shape for nested vendors — catalog slugs are single-slash (`zenmux/openai-gpt-5.6-sol`) while `/v1/models` ids are multi-slash (`zenmux/openai/gpt-5.6-sol`), and the client flattens all slashes to `-`. Any cross-referencing code (or operator) has to collapse both sides to match.

### Version

2.48.0 (hub and clients)

### Operating system

Hub: Ubuntu 24.04 (Contabo VPS); clients: macOS 26 and Ubuntu VPS

### Provider and model

Any hub-routed provider. Observed: cursor/gpt-5.6-sol (1,000,000 → 128,000), zai/glm-5.3-flash (1,000,000 → 128,000), alibaba-token-plan-intl/qwen3.8-max (983,616 → 128,000)

### Logs or error output

No error — silent metadata loss. Verified comparison:

```text
model                                        hub /v1/catalog   hub /v1/models capabilities   client catalog after sync
cursor/gpt-5.6-sol                           1,000,000         context_length 922,000          128,000
zai/glm-5.3-flash                            1,000,000         —                              128,000
alibaba-token-plan-intl/qwen3.8-max          983,616           —                              128,000
gpt-5.6-sol (hub's own openai roster row)    272,000           context_length 922,000          128,000 (routed) vs 272,000 (local native row — correct)
```

_No response_

### Redacted configuration

Client (chained setup):

```json
{
  "providers": {
    "hub": {
      "baseUrl": "http://<hub-tailnet-ip>:10100/v1",
      "allowPrivateNetwork": true,
      "authMode": "key",
      "liveModels": true
    }
  },
  "defaultProvider": "hub"
}
```

Workaround added per client (keys in `/v1/models` id space):

```json
{
  "providers": {
    "hub": {
      "modelContextWindows": {
        "cursor/gpt-5.6-sol": 1050000,
        "zenmux/openai/gpt-5.6-sol": 1050000,
        "zai/glm-5.3-flash": 1000000
      },
      "modelMaxInputTokens": { "cursor/gpt-5.6-sol": 922000 }
    }
  }
}
```

### Suggested fix

1. When assembling routed rows for a provider whose listing carries per-model metadata, map `capabilities.context_length` / `max_output_tokens` from `GET /v1/models` into the discovered window instead of dropping it — that alone fixes provider-based chained clients with zero operator config.
2. Alternatively (or additionally), have the provider-based sync path consume the same `GET /v1/catalog` payload the `connect` flow already downloads — it carries the operator-curated `modelContextWindows` truth byte-for-byte.
3. Worth preserving while fixing: bare native rows correctly keep the roster's plan-effective window today (e.g. 272000 for GPT-5.6 via a ChatGPT plan, where `pricing.overrides.min_prompt_tokens: 272000` applies); routed API-billed rows should get the API truth instead of the 128k floor.
4. Papercut: unify catalog-slug vs `/v1/models`-id shape for nested vendors (single- vs multi-slash) so the two surfaces can be cross-referenced without collapsing separators.

### LINKED ISSUE #4035: [Bug]: Codex App update invalidates the persisted codex-runtime.json pin — the dead hashed path is never re-resolved
### Client or integration

Codex App

### Area

CLI

### Summary

`~/.opencodex/codex-runtime.json` persists an absolute path to the Codex App's bundled CLI, and that path lives inside a **content-hashed** directory:

```
%LOCALAPPDATA%\OpenAI\Codex\bin\<hash>\codex.exe
```

The Codex App replaces that hashed directory on every update and deletes the previous one. The persisted `configured` pin then points at a path that no longer exists, and runtime selection falls through to `fallback`:

```
Codex runtime: codex
Codex version: unknown
Codex source: fallback
⚠️  No validated Codex runtime found (path does not exist); falling back to `codex`.
```

A dead `configured` entry is never discarded or re-resolved. Because the Codex App does not put its hashed `bin` directory on `PATH`, the `env → configured → shim → PATH → fallback` chain introduced in #359 has nothing left to find, and `codex` does not resolve as a bare command on this host.

I expected a recorded runtime path that no longer exists to be treated as absent — dropped, with resolution continuing down the chain and the file re-persisted — rather than retained while runtime selection degrades to a bare `codex` fallback.

This is not #3442. That fix covers a pin whose path is real but whose recorded version is old (the entitlement `client_version` floor). This is a pin whose path does not exist at all.

### Reproduction

Observed twice on the same host, the second time live within a single 15-minute window.

**Instance 1 — how the pin died**

1. On 2026-09-02, `codex-runtime.json` recorded `"command": "...\\OpenAI\\Codex\\bin\\b99306303521e97e\\codex.exe"`, `"source": "configured"`, `"selectedVersion": "0.151.0-alpha.7.2"`.
2. On 2026-09-04, the Codex App updated to 0.153.0. The bundled CLI moved to `...\bin\9ba750cce02d5e5c\`, and `b99306303521e97e` was removed.
3. From then on, `ocx status` and `ocx doctor` reported `source=fallback`, `version=unknown`, and the "No validated Codex runtime found (path does not exist)" warning. This survived `ocx service restart` and an upgrade from 2.42.0 to 2.48.0.

**Instance 2 — the same directory replacement observed live**

4. At 2026-09-08 ~21:26 KST, `...\OpenAI\Codex\bin\` contained `9ba750cce02d5e5c` (holding `codex.exe`, `codex-cli 0.153.0`) and `e5208f10301fe5e6` (holding `rg.exe`).
5. A Codex App update was applied at ~21:38 KST.
6. `...\bin\` then contained `8e5b6932251c2c1c` and `c60635126245daef`. `9ba750cce02d5e5c` no longer existed — roughly 15 minutes after being the live runtime directory.

Any pin written against a hashed directory is therefore invalidated by the next Codex App update.

**A stable path exists and is not used**

`%USERPROFILE%\.codex\plugins\.plugin-appserver\codex.exe` carries no hash and is updated in place. Across the update in step 5 it kept the same path and went from `codex-cli 0.153.0` to `codex-cli 0.153.4`. Searching this repository for `plugin-appserver` returns no results, so the resolver does not appear to consider this location.

**Workaround**

Repointing `command` in `codex-runtime.json` to that stable path restored `source=configured` with an attested version, and it survived the step-5 update untouched. `ocx doctor` reports no warnings afterwards.

### Analysis and suggested directions

1. **Validate `configured` at resolution time.** When the recorded `command` does not exist, treat the entry as absent: continue down `env → configured → shim → PATH → fallback` and re-persist the result, instead of retaining a dead pin and degrading to `fallback`.
2. **Add the Codex App's stable location as a discovery candidate.** `%USERPROFILE%\.codex\plugins\.plugin-appserver\codex.exe` (and the macOS/Linux equivalents) is the only Codex App path observed here that survives an update.
3. **Refresh `selectedVersion` on drift.** After the workaround, the file still records `0.153.0` while the resolved binary reports `0.153.4`. `ocx status` shows the probed value, so the recorded field goes stale. Harmless here — well above the 0.144.0 gated floor — but it is the same field that tier 2 of `resolveCodexEntitlementClientVersion` reads (#3442).

Two notes on existing recovery paths:

- `ocx doctor --fix-codex-runtime`, added in #359, is not listed in `ocx doctor --help` on 2.48.0. I did not run it.
- Even if it were available, it could not have recovered this host: with the pin dead and no Codex CLI on `PATH`, the chain has nothing to resolve to, which is exactly the observed `fallback` state.

Related: #359 (the resolver and `codex-runtime.json`), #3442 (stale-but-valid pin version), #356 (shim restore after external npm updates — the analogous problem for npm-installed CLIs).

### Version

2.48.0

### Operating system

Windows 11 Pro 10.0.26200

### Provider and model

_No response_

### Logs or error output

```shell
# Before — dead pin (2.48.0, after the 2026-09-04 Codex App update)
$ ocx status
   Codex runtime: codex
   Codex version: unknown
   Codex source: fallback
   /!\  No validated Codex runtime found (path does not exist); falling back to `codex`.
        Run ocx doctor for diagnosis and recovery.

$ ocx system codex-cli-update check --json
{
  "candidateAvailable": false,
  "versionEvidence": { "kind": "unavailable" },
  "provenance": "unknown",
  "managed": false,
  "reason": "candidate_unavailable",
  "location": null
}

# Intermediate — with the hashed dir temporarily added to PATH, selection
# succeeded, but the dead pin was still preferred over it
$ ocx doctor
  Codex runtime selection
    ok  Selected runtime: ...\OpenAI\Codex\bin\9ba750cce02d5e5c\codex.exe (0.153.0, source=path)
    !!  Preferred runtime unavailable; fell back to ...\bin\9ba750cce02d5e5c\codex.exe

# Hashed directory replacement, observed live
# 21:26 KST
9ba750cce02d5e5c/   e5208f10301fe5e6/
# 21:38 KST, after the Codex App update
8e5b6932251c2c1c/   c60635126245daef/
$ ls .../OpenAI/Codex/bin/9ba750cce02d5e5c
ls: cannot access '.../OpenAI/Codex/bin/9ba750cce02d5e5c': No such file or directory

# The stable path across that same update
# before: codex-cli 0.153.0    after: codex-cli 0.153.4
# path unchanged: %USERPROFILE%\.codex\plugins\.plugin-appserver\codex.exe

# After — pin repointed to the stable path
$ ocx status
   Codex runtime: C:\Users\[USER]\.codex\plugins\.plugin-appserver\codex.exe
   Codex version: 0.153.4
   Codex source: configured
$ ocx doctor
   (no warnings)
```

No screenshots attached. Every value above was read from `ocx status`, `ocx doctor`, `ocx system codex-cli-update check --json`, directory listings under `%LOCALAPPDATA%\OpenAI\Codex\bin\`, and `codex.exe --version`. Account identifiers, tokens, credentials and quota data are excluded.

### Redacted configuration

```json
// ~/.opencodex/codex-runtime.json — the recorded command path no longer exists
{
  "version": 1,
  "command": "C:\\Users\\[USER]\\AppData\\Local\\OpenAI\\Codex\\bin\\b99306303521e97e\\codex.exe",
  "source": "configured",
  "selectedVersion": "0.151.0-alpha.7.2",
  "updatedAt": "2026-09-02T05:57:19.180Z"
}
```

### Checks

lidge-jun:
## Summary

Fourth sequential integration branch of the 2.49.0 backlog closeout. Four bounded fixes for open bug reports that had no owning pull request, one commit each, every one carrying a regression test proven red on `dev` before the fix.

- **#4032** — a chained OpenCodex hub reports the per-model window on the same capability record that `catalogHintsFromModelsApiItem` already reads for `max_output_tokens`, but the window was never read from it. Every routed row fell through to the 128k compatibility floor in `parsing.ts` while local forward rows kept their real values. One argument appended last to the existing `positiveSafeInteger` list, so a provider that already resolves a window is unaffected. A hub serving `context_length: 922000` now materializes at 922000 instead of 128000.
- **#4035** — when a Codex App update moves the hashed runtime path, `resolveAndPersistCodexRuntime` degrades to `fallback` but skips the persist branch, so the dead `configured` pin stays on disk and is re-probed forever. The pin is now cleared when resolution degrades to `fallback` for a path-does-not-exist reason, and the failure names the persisted command. A pin rejected for any other reason, or a resolve that succeeds, is left exactly as before.
- **#4023** — the dashboard Stop button could unload the launchd job before native Codex teardown finished, leaving client config pointed at a dead proxy. `installedServiceRespawnRisk` gains a `self-unload` verdict for a proxy running as its own service (keyed on `OCX_SERVICE=1` plus the definition file), and the management route answers 409 `self_unload_service` instead of stopping. This covers Linux systemd in the same change. `ocx stop`, manually started proxies, and the receipt path are untouched.

  Review caught that the first version of this keyed on `OCX_SERVICE=1`, which does not identify the managed job: `ocx claude` and `ocx opencode` both set it on the detached proxy they spawn, so a user with the service installed but stopped would have had their dashboard Stop refused for a proxy no manager supervises. The plist and unit now also write `OCX_SERVICE_MANAGED=1` and the refusal discriminates on that, with a regression case that fails against the old discriminator.
- **#3807** — narrower than the report reads. The reported seed shape is already admitted by `externalTaskInputContent()` since `a73bb160f` (2.44.0); what still failed is that the admission test checked for the *field* (`"call_id" in item`), so a client sending `call_id: null` or `""` still got a 400. A `hasPairingKey()` predicate replaces the field check. `src/server/responses/core.ts` is byte-identical, so the #3259 unpaired-tool-result guard keeps its protection.

#3807 intentionally inverts four assertions landed by #3735 (`empty call_id` and `null call_id`, in both `responses-parser` and `responses-compaction-routing`) and replaces them with positive 200 tests. Those rows asserted that a seed with a null or empty `call_id` stays off the user path; since neither value can ever pair with a `function_call`, classifying them as paired results was wrong independently of #3259.

Closes #4032
Closes #4035
Closes #4023
Closes #3807

## Verification

- `bun x tsc --noEmit` on the stacked tree — exit 0.
- `bun test` on the five changed test files plus both layout guards — **303 pass / 0 fail / 1670 expect()**.
- Every fix proven non-vacuous by running its test before the source change: #4032 4 pass / 2 fail → 6 pass / 0 fail; #4035 36 pass / 1 fail → 37 pass / 0 fail; #4023 30 pass / 3 fail → 33 pass / 0 fail; #3807 118 pass / 3 fail → 121 pass / 0 fail.
- Consumer sets run per item: 708 pass across fourteen runtime consumers (#4035), 1983 pass across `tests/responses` and `tests/server/responses` (#3807), 96 pass across the catalog neighbours (#4032), and `grok-lifecycle` 32 pass with the landed #3008 ordering assertion intact (#4023).
- Not run locally: full `bun run test` (hosted CI covers Linux, Windows, macOS).

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Improved model compatibility by recognizing context-window limits reported by hub-style providers.
  - External task inputs with empty or null call IDs are now accepted correctly.

- **Bug Fixes**
  - Stopping a proxy running as an installed macOS or Linux service now provides safe guidance to run `ocx stop`, preventing incomplete restoration.
  - Automatically recovers from persisted runtime selections whose executable is no longer available.

- **Documentation**
  - Documented the new service-stop behavior and related API response.


lidge-jun:
## 리뷰 · 우선순위 74 / 80

지금 `dev` 끝은 `164bd2b36` (#4053 wp3) 이고, 이 PR은 그 위에 얹는 **wp4** 입니다. 요약하면 이렇게입니다. 2.49.0 백로그를 닫는 네 번째 묶음인데, 아직 주인이 없는 열린 버그 네 개(#4032, #4035, #4023, #3807)를 **커밋 하나씩** 고칩니다. 각각 고치기 전에 테스트가 빨갛게 실패하는 걸 먼저 확인하고, 고친 뒤에야 초록으로 바뀌게 만든 형태입니다. 지금 `dev`가 이미 wp2 정확성(#4051)과 wp3 소규모 기여자·피드백 안정화(#4053)까지 올린 상태라, 이 네 건은 “아직 손 안 간 실사용 버그”를 바로 닫는 다음 한 걸음입니다. types.ts/config.ts 큰 쪼개기와는 겹치지 않고, preview 배포도 아닙니다.

첫 번째(#4032)는 카탈로그입니다. 체이닝된 OpenCodex 허브가 모델마다 `context_length`를 이미 내려주는데, `src/codex/catalog/provider-fetch.ts`의 `catalogHintsFromModelsApiItem`이 그걸 안 읽어서, 라우팅된 행이 전부 `parsing.ts`의 128k 호환 바닥으로 떨어졌습니다. 로컬 forward 행은 진짜 창을 유지하고 허브 행만 128k로 뭉개지는 불균형이었습니다. 고침은 `positiveSafeInteger(...)` 목록 **맨 끝**에 `capabilityRecord?.context_length` 하나만 더하는 것입니다. 같은 함수가 이미 `capabilityRecord?.max_output_tokens`를 읽고 있어서, 허브가 쓰는 그 capability 레코드를 그대로 이어 받은 셈입니다. 이미 다른 필드로 창을 찾은 프로바이더는 순서가 끝이라 행동이 안 바뀝니다. 허브가 `922000`을 주면 이제 922000으로 남습니다.

두 번째(#4035)는 Codex App 업데이트 뒤입니다. 해시된 런타임 경로가 바뀌면 `resolveAndPersistCodexRuntime`이 `fallback`으로 떨어지는데, 예전에 쓰던 `configured` 핀은 `codex-runtime.json`에 그대로 남았습니다. `source === "fallback"`일 때는 persist 분기를 안 타기 때문입니다. 그래서 죽은 경로를 영원히 다시 프로브합니다. 이번 고침은 `PATH_MISSING_REASON`(`path does not exist`)으로 **그 핀 명령**이 거절됐을 때만 `clearPersistedCodexRuntime`으로 파일을 지웁니다. 다른 이유로 거절되거나 resolve가 성공하면 예전과 같습니다. 메모 캐시도 persist와 같이 비웁니다.

세 번째(#4023)는 대시보드 Stop입니다. 프록시가 자기 자신인 launchd/systemd 잡일 때, 매니저를 안에서 멈추면 네이티브 Codex 복원이 끝나기 전에 프로세스가 죽어 `openai_base_url` 등이 죽은 프록시를 가리킬 수 있습니다. `installedServiceRespawnRisk`에 `self-unload` 판정을 추가하고(`OCX_SERVICE=1` + plist/unit 파일 존재), `/api/stop`가 409 `self_unload_service`로 거절합니다. Windows Task Scheduler 쪽 `respawnable`/`unknown`은 그대로입니다. `ocx stop`, 수동 기동 프록시, receipt 경로는 손대지 않습니다. 가이드·management-api 문서에도 같은 코드가 적혔습니다.

네 번째(#3807)는 서브에이전트 시드입니다. 보고된 “필드 없는” 시드 모양은 이미 `a73bb160f`(2.44.0)부터 `externalTaskInputContent()`가 받습니다. 남은 구멍은 클라이언트가 `call_id: null`이나 `""`를 **명시적으로** 보낼 때입니다. 예전에는 `"call_id" in item`만 봐서 필드만 있으면 페어링된 tool result로 취급했고, 그다음 #3259 가드가 400을 냈습니다. 지금은 `hasPairingKey()`로 바꿉니다. null·빈 문자열·공백만 있는 문자열은 키가 없고, 진짜 문자열 키만 페어링으로 봅니다. 숫자·객체 같은 잘못된 타입은 그대로 거절해서 #3259 보호는 유지합니다. `src/server/responses/core.ts`는 바이트 단위로 안 건드렸다고 본문에 적혀 있습니다. 그래서 #3735가 넣었던 “empty/null call_id는 toolResult” 단언 네 줄은 의도적으로 뒤집고, 유저 경로 200 테스트로 바꿉니다. 페어링될 수 없는 값을 tool result로 분류한 게 원래 틀린 단언이었기 때문입니다.

검증 쪽도 정리가 잘 되어 있습니다. 로컬에서 바뀐 테스트 다섯 개 + 레이아웃 가드 303 pass, 항목별로 고치기 전 실패→고친 뒤 통과가 적혀 있고, 카탈로그·런타임·responses·grok-lifecycle 이웃 스위트도 돌렸다고 합니다. CI는 이 글을 쓰는 시점엔 hygiene/enforce-target/linux-systemd/macos-launchd 등은 통과했고, 본 테스트 샤드·크로스플랫폼은 아직 pending입니다. 파일 목록도 네 이슈에 맞춰 `provider-fetch.ts` / `runtime.ts` / `service.ts`+`management-api.ts` / `task-input.ts`와 대응 테스트·문서뿐입니다. types/config 쪼개기나 sponsor 지연(#3914/#3915) 같은 열린 큰 축과 충돌하지 않습니다.

`src/codex/catalog/provider-fetch.ts` (`catalogHintsFromModelsApiItem`) - `capabilityRecord?.context_length`를 `positiveSafeInteger` **맨 끝**에 넣는 선택은 맞다. 다만 허브가 `capabilities.limits` 쪽에만 창을 넣는 변형이면 여전히 128k로 떨어질 수 있다. 지금 이슈 #4032 재현 모양이 capability 레코드 top-level인지 한 번만 확인하면 충분하다.

`src/codex/runtime.ts` (`resolveAndPersistCodexRuntime`) - 핀 삭제가 `fallback` + 해당 핀의 `PATH_MISSING_REASON`으로만 묶여 있어서, “있는 바이너리인데 spawn 불가” 같은 다른 거절은 핀을 남긴다. 의도된 좁은 범위다. 운영자가 수동으로 죽은 절대경로를 다시 넣고 싶을 때만 헷갈릴 수 있다.

`src/service.ts` (`installedServiceRespawnRisk`) - darwin/linux만 `self-unload`이고 WinSW 등 다른 non-win32는 `none`이다. 현재 제품 경로상 맞지만, 나중에 WinSW를 “자기 자신 unload”로 취급해야 하면 같은 구멍이 다시 열린다.

`src/responses/task-input.ts` (`hasPairingKey`) - 공백만 있는 `call_id`는 헬퍼에서는 시드로 받지만, parser 쪽 schema(`z.string().min(1)`)가 먼저 잡으면 이 경로에 안 온다. 테스트 주석에 그 차이가 적혀 있어서 동작은 일관되지만, “헬퍼 단위”와 “HTTP envelope 단위” 기대를 읽는 사람이 헷갈릴 수 있다.

`tests/responses/responses-parser.test.ts` - #3735 단언을 뒤집는 변경이라, 예전에 empty/null call_id를 toolResult로 가정하던 외부 문서나 이슈 코멘트가 남아 있으면 같이 고쳐야 한다. 코드 자체 회귀는 새 200 테스트가 막는다.

메인테이너의 판단이 필요한 지점
- #4053 직후 tip으로 바로 올릴지, 아니면 본 테스트 샤드(test 1/4–4/4, macos, windows)가 초록 될 때까지 잠깐 기다릴지
- 머지 직후 이슈 #4032/#4035/#4023/#3807에 `Landed via #4058 at <commit>` + `landed-via-maintainer` 위생을 같은 호흡으로 할지 (본문이 이미 Closes를 걸고 있음)
- #3807이 #3735 단언을 뒤집는 점을 릴리즈 노트/변경 요약에 한 줄로 남길지

너의 추천
CI 본 테스트 샤드가 초록이면 **머지**하세요. 범위가 네 이슈에 딱 맞고, 현재 `dev`(wp3 tip) 방향과도 맞고, 로컬에서 이미 red→green을 증명했습니다. 머지 직후 네 이슈 leftover 위생 코멘트·라벨을 바로 처리하세요. CI가 한 샤드에서만 실패하면 그 샤드만 보고 고치면 되고, 전체를 다시 쪼갤 필요는 없습니다.

이 댓글은 grok-bot이 작성했습니다

Issue #3719, opened 2026-09-05
Bug: preserve Anthropic thinking replay through proxy-auth translation and clarify prompt-cache behavior


lidge-jun:
### Client or integration

Claude Code

Also reported with Claude Desktop 3P against a remote OpenCodex hub, using a connected client's proxy admission credential.

### Area

Proxy and routing

### Summary

Separate follow-up for the thinking/replay and prompt-cache request in #3646. This issue applies when **Anthropic is the intended upstream**. It does not cover Desktop date/hash alias discovery, hub registry ownership, client profile apply, or rejection of unresolved managed aliases; those belong to the remote-alias slice of #3646. Landing that slice does not establish that this replay/cache request is resolved.

The reporter asks that assistant `thinking` and `redacted_thinking` blocks survive subsequent turns to the intended Anthropic route, or that a supported untranslated path preserve the original Anthropic exchange. The guide should also explain how proxy authentication, native passthrough eligibility, and prompt-cache retention differ.

Evidence and limits:

- The current inbound translator explicitly drops assistant `thinking` and `redacted_thinking` blocks on replay (`src/claude/inbound.ts`, assistant-content conversion). The Claude Code guide documents this policy.
- Native passthrough requires an eligible unchanged Anthropic model and an actual Anthropic credential; a hub admission token by itself is not that credential. An exposed listener also requires the dedicated proxy admission header. Proxy-auth translation and native passthrough are different routes.
- **Translation does not mean prompt caching is disabled.** The Anthropic adapter still applies prompt-cache controls, including automatic caching for the native Anthropic endpoint. `cacheRetention` supports `short` (default, five minutes), `long` (one hour), and `none`.
- #3646 reports repeated cache misses and faster quota consumption after thinking responses. Neither a per-turn miss guarantee nor its causal attribution to dropped replay blocks has been reproduced in this follow-up. The report's claim that native Claude typically uses one-hour breakpoints also needs a controlled comparison. Do not infer billing or quota savings from the alias fix or from the configured retention duration alone.

Expected outcome:

- [ ] Define and implement supported Anthropic-to-Anthropic replay fidelity for proxy-auth traffic, preserving upstream-issued thinking signatures and opaque redacted-thinking blocks where the upstream/client contract requires them. Do not manufacture signatures or forward provider-specific blocks to unrelated providers.
[...]

PR #4072, opened 2026-09-09
fix(openai): drop tool-schema patterns Python re cannot compile

itismyfield:
## Summary

OpenAI-family upstreams validate a function tool's JSON Schema `pattern` by compiling it with Python's `re`. Python `re` has no support for Unicode property escapes (`\p{...}` / `\P{...}`), so any pattern using them is refused before the request is routed:

```
400 Invalid schema for function 'Artifact':
'^(?!__.*__$)[^\\p{Cc}\\p{Cf}\\p{Zl}\\p{Zp}"\\\\./[\\]]{1,200}$' is not a 'regex'.
```

The pattern is valid ECMA-262 — it comes from a client that authors its schemas in JavaScript. Claude Code 2.1.265 introduced exactly this one on the `field` parameter of its built-in `Artifact` tool. Built-in tool definitions ship on **every** request, so this takes out the whole GPT route for Claude Code users on 2.1.265+: the first message of any session fails, whether or not the tool is ever called.

Reproduced against a local proxy — identical request, only the `pattern` key differs:

| tools payload | result |
| --- | --- |
| without the `\p{...}` pattern | `http=200` |
| with the `\p{...}` pattern | `http=400 ... is not a 'regex'` |

Same string, three engines:

```
python re     -> PatternError: bad escape \p at position 14
ECMA (no flag) -> OK
ECMA (u flag)  -> OK
```

## Changes

- **`src/adapters/responses-tool-schema.ts`** — new `stripUnicodePropertyPatterns`, sitting next to `stripResponsesOnlyEncryptedMarker` and following the same explicit-stack walk (schema depth is caller-controlled; recursion would turn a deep schema into a stack overflow on the request path). It returns the input by identity when nothing was dropped, so callers keep their existing "did this change?" checks. The name-bag / literal-value key sets are now shared by both strippers and renamed accordingly — both need to tell "the keyword `x`" apart from "a property someone named `x`".
- **`src/adapters/openai-responses.ts`** — applied in `normalizeFunctionToolSchema`. This is the one tool-schema transform in the passthrough `buildRequest` that runs outside the `isCanonicalOpenAiForwardProvider` guards, so a single call covers the ChatGPT backend (`authMode: "forward"`), generic `openai-responses` gateways, Azure via `contractParent`, and the WebSocket codex transport, which re-parses the very string this path serializes.
- **`src/adapters/openai-chat.ts`** — composed into the existing sanitizer slot in `toolsToChatFormat`, since the chat-completions upstream is a separate serializer built from `parsed.context.tools`.

Not sanitized at `src/claude/inbound-content-options.ts`: that layer is pinned as a verbatim forwarder by `tests/claude-integration/claude-inbound.test.ts:697`, and fixing it there would cover Claude-Code-origin traffic only. Not at `src/responses/parser-tools.ts` either — the passthrough adapter ships `_rawBody` and never reads `parsed.context.tools`.

## Why it's safe

`pattern` is an advisory hint for the model; neither this proxy nor the upstream enforces it on the arguments a tool is actually called with. Dropping only what the destination cannot compile keeps every tool's shape — the same trade #50 made for the validation keywords Bedrock rejects.

The strip is narrow by construction:

- only patterns that actually use a property escape are dropped. A plain `^[a-z0-9_-]{1,64}$` survives, and so does a lookahead — Python `re` supports those, and Claude Code's sibling `collection` pattern is exactly that case;
- `\\p{2}` is an escaped backslash followed by a quantified `p`, which Python compiles fine. A substring scan for `\p{` would discard that working pattern, so the check tracks escaping instead;
- name bags are respected: a property, `$defs` entry, or `patternProperties` key literally named `pattern` is data, and `const`/`default`/`enum`/`examples` payloads are left whole;
- no tool is ever dropped, so `tool_choice` reconciliation is untouched.

**Not in scope:** `patternProperties` *keys* are themselves regexes and would hit the same validator, but no failure has been observed there and removing a key would change the schema's structure rather than an advisory hint. Left alone deliberately — happy to extend if a maintainer wants it covered.

## Testing

- `bun test tests/adapters/openai/openai-chat-hardening.test.ts tests/responses/openai-responses-passthrough.test.ts` -> 243 pass / 0 fail
- `bun run test:changed` -> 8818 pass / 2 skip / 0 fail across 407 files
- `bun x tsc --noEmit` -> clean
- `bun run privacy:scan` -> passed
- Mutation check: reverting the `normalizeFunctionToolSchema` call fails exactly one test — the codex-forward wire assertion — so the new coverage is not vacuous.

`bun run test` does not complete on my machine, **on this branch and on unmodified `dev` alike**: `tests/routing/routing-policy-surface-parity.test.ts` dies with `worker crashed: SIGSEGV` under the default `--parallel=4`, and the sibling workers abort in cascade (1063 aborted on stashed `dev`, 1064 here — the delta is the tests this PR adds). Diffing both logs, the only non-cascade failure is that same SIGSEGV file in both runs, so this change introduces no new failure. Run alone it passes: `bun test tests/routing/routing-policy-surface-parity.test.ts` -> 6 pass / 0 fail. Looks like a local environment issue rather than something this PR touches, but flagging it rather than claiming a green full suite I did not get.

New tests: a direct unit block beside the existing `stripResponsesOnlyEncryptedMarker` tests (drop vs. keep, `\P{...}`, escaped-backslash literal, name-bag preservation, identity return, and a 50 000-deep walk mirroring the sibling stack test), plus a wire assertion in `tests/responses/openai-responses-passthrough.test.ts` driven through `_rawBody` on the codex forward provider — that one exercises the exact path that 400s.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for tool schemas containing Unicode property escape patterns.
  * Incompatible patterns are now removed before schemas are sent through Chat and Responses integrations, while supported patterns and data payloads are preserved.
  * Deeply nested schemas are handled without stack overflow issues.

### DIFF
--- a/src/adapters/openai-chat.ts
+++ b/src/adapters/openai-chat.ts
@@ -27,7 +27,7 @@ import {
   type ResolvedFastPolicy,
 } from "../providers/fastwire";
 import { openaiChatCompletionsUrl } from "./openai-chat-url";
-import { stripResponsesOnlyEncryptedMarker } from "./responses-tool-schema";
+import { stripResponsesOnlyEncryptedMarker, stripUnicodePropertyPatterns } from "./responses-tool-schema";
 import { agentRouterDefaultHeaders, frameAgentRouterMessages } from "./agentrouter";
 import {
   isXaiSchemaTarget,
@@ -1331,7 +1331,7 @@ function toolsToChatFormat(parsed: OcxParsedRequest, provider: OcxProviderConfig
       : moonshotTarget
         ? normalizeMoonshotToolParameters(t.parameters)
         : ensureRootObjectType(t.parameters);
-    const parameters = stripResponsesOnlyEncryptedMarker(normalized);
+    const parameters = stripUnicodePropertyPatterns(stripResponsesOnlyEncryptedMarker(normalized));
 
     if (parameters === undefined) return [];
     return [{
--- a/src/adapters/openai-responses.ts
+++ b/src/adapters/openai-responses.ts
@@ -25,6 +25,7 @@ import { rewriteRoutedToolSearchForUpstream } from "../responses/tool-search-com
 import { rewriteRoutedNamespaceToolsForUpstream } from "../responses/namespace-tool-compat";
 import { openaiResponsesUrl } from "./openai-responses-url";
 import { normalizeResponsesCodeMode } from "./responses-code-mode";
+import { stripUnicodePropertyPatterns } from "./responses-tool-schema";
 import { injectXaiResponsesXSearch, normalizeXaiResponsesWebSearch } from "./xai-web-search";
 import { EMPTY_TOOL_OUTPUT_ANNOTATION, isWhitespaceOnlyTextPartArray } from "./empty-tool-output-annotation";
 import {
@@ -649,14 +650,18 @@ function mapRoutedResponsesReasoningEffort(
 
 function normalizeFunctionToolSchema(tool: unknown, xaiTarget: boolean): unknown | undefined {
   if (!isPlainObject(tool) || tool.type !== "function") return tool;
+  // Runs for every Responses destination, forward auth included: the ChatGPT backend is where
+  // the `\p{…}` rejection was observed, and it reaches this function through the same seam.
+  const compatible = stripUnicodePropertyPatterns(tool);
+  const source = isPlainObject(compatible) ? compatible : tool;
   if (xaiTarget) {
-    const parameters = normalizeXaiToolParameters(isPlainObject(tool.parameters) ? tool.parameters : {});
-    return parameters === undefined ? undefined : { ...tool, parameters };
+    const parameters = normalizeXaiToolParameters(isPlainObject(source.parameters) ? source.parameters : {});
+    return parameters === undefined ? undefined : { ...source, parameters };
   }
-  if (isPlainObject(tool.parameters) && tool.parameters.type === "object") return tool;
+  if (isPlainObject(source.parameters) && source.parameters.type === "object") return source;
   return {
-    ...tool,
-    parameters: { ...(isPlainObject(tool.parameters) ? tool.parameters : {}), type: "object" },
+    ...source,
+    parameters: { ...(isPlainObject(source.parameters) ? source.parameters : {}), type: "object" },
   };
 }
 
--- a/src/adapters/responses-tool-schema.ts
+++ b/src/adapters/responses-tool-schema.ts
@@ -98,11 +98,14 @@ function usesUnicodePropertyEscape(pattern: string): boolean {
  *   '^(?!__.*__$)[^\p{Cc}\p{Cf}\p{Zl}\p{Zp}"\\./[\]]{1,200}$' is not a 'regex'.
  *
  * A client that ships such a pattern on a built-in tool therefore loses every request, not just
- * the calls to that tool — Claude Code 2.1.265 does exactly this on its `Artifact` tool. Since
- * `pattern` is advisory for the model, and neither this proxy nor the upstream enforces it on
- * the arguments a tool is actually called with, dropping just the patterns the destination
- * cannot compile keeps the tool's shape while letting the request through. That is the same
- * trade the Kiro adapter makes for the validation keywords Bedrock rejects.
+ * the calls to that tool — Claude Code 2.1.265 does exactly this on its `Artifact` tool.
+ * Dropping only the patterns the destination cannot compile keeps the tool's shape while letting
+ * the request through, the same trade the Kiro adapter makes for the validation keywords Bedrock
+ * rejects. What is given up is bounded: an upstream that enforces `pattern` does so under strict
+ * Structured Outputs, and a pattern this function drops is one that upstream could not have
+ * compiled in the first place — it refuses the whole schema before any argument is generated. So
+ * the choice is a dropped constraint versus no request at all, not a silently weakened one that
+ * would otherwise have been enforced.
  *
  * Returns `node` itself when nothing was dropped, so callers can use identity to tell whether
  * the schema changed. Walks an explicit stack for the same reason as the stripper above.
--- a/src/adapters/responses-tool-schema.ts
+++ b/src/adapters/responses-tool-schema.ts
@@ -1,8 +1,7 @@
-// Codex multi-agent v2 stamps a Responses-only `encrypted: true` marker on
-// collaboration tool schemas (openai/codex 5f4d06ef; issue #85). It is an
-// annotation for the ChatGPT backend only, so translated provider schemas must
-// drop it without removing properties or definitions literally named `encrypted`.
-const ENCRYPTED_MARKER_NAME_BAG_KEYS = new Set([
+// Keys whose *children's names* are caller-chosen rather than schema keywords, and keys whose
+// values are literal payloads rather than schemas. Shared by both strippers below: each one has
+// to tell "the keyword `x`" apart from "a property someone named `x`".
+const SCHEMA_NAME_BAG_KEYS = new Set([
   "properties",
   "patternProperties",
   "$defs",
@@ -11,9 +10,14 @@ const ENCRYPTED_MARKER_NAME_BAG_KEYS = new Set([
   "dependentSchemas",
   "dependentRequired",
 ]);
-const ENCRYPTED_MARKER_LITERAL_VALUE_KEYS = new Set(["const", "default", "enum", "examples"]);
+const SCHEMA_LITERAL_VALUE_KEYS = new Set(["const", "default", "enum", "examples"]);
 
 /**
+ * Codex multi-agent v2 stamps a Responses-only `encrypted: true` marker on collaboration tool
+ * schemas (openai/codex 5f4d06ef; issue #85). It is an annotation for the ChatGPT backend only,
+ * so translated provider schemas must drop it without removing properties or definitions
+ * literally named `encrypted`.
+ *
  * The schema is caller-supplied, so its nesting depth is attacker-influenced. Native recursion
  * would turn a deep schema into a stack overflow that takes down the request path, so this walks
  * an explicit stack instead: depth costs heap, which is bounded and recoverable.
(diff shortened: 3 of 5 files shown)


lidge-jun:
## 리뷰 · 우선순위 74 / 80

설명

이 PR은 OpenAI 계열(ChatGPT forward, 일반 openai-responses, Azure contractParent, WebSocket codex, openai-chat)으로 나가는 function tool의 JSON Schema `pattern` 안에서, Python `re`가 컴파일하지 못하는 유니코드 프로퍼티 이스케이프(`\p{...}` / `\P{...}`)만 골라 지우는 수정이다. 지금 `dev` HEAD는 `8026405d9`(`#4067` wp7 proxy stop 거부 사유)이고, package.json은 2.49.0이다. 현재 방향은 2.49.x 백로그 마감이며, 이 변경은 열려 있는 슬라이스(`#3719` live replay, `#3379` selector rename, `#3774`/`#3781`/`#3782`)와 겹치지 않는 독립 버그픽스다.

왜 급한가. Claude Code 2.1.265가 내장 `Artifact` 도구의 `field` 파라미터에 ECMA-262용 패턴을 넣었고, 그 도구 정의는 매 요청마다 실려 나간다. OpenAI 쪽은 `pattern`을 Python `re`로 검사하는데 `\p{Cc}` 같은 문법은 지원하지 않아서, 도구를 한 번도 안 써도 GPT 경로 첫 메시지부터 400이 난다. PR 본문에 적은 재현표(동일 요청에서 pattern만 바꿨을 때 200 vs 400)와 세 엔진 비교가 그 사실을 잘 보여 준다.

고치는 위치도 맞다. 새 `stripUnicodePropertyPatterns`는 기존 `src/adapters/responses-tool-schema.ts`의 `stripResponsesOnlyEncryptedMarker` 옆에 두고, name-bag / literal-value 키 집합을 `SCHEMA_*`로 공유한다. Responses 쪽은 `normalizeFunctionToolSchema`에 넣었는데, 이 함수는 forward-provider 가드 밖에서도 돌아가서 ChatGPT backend까지 한 번에 덮는다. Chat 쪽은 `toolsToChatFormat`의 기존 sanitizer 자리에 이어서 호출한다. Claude inbound verbatim 레이어나 `parser-tools`에 안 넣는 이유도, 지금 체크아웃 기준으로 타당하다(inbound는 verbatim 테스트로 고정, passthrough는 `_rawBody`를 탄다).

안전성도 좁게 잡혀 있다. `pattern`은 모델에게 주는 힌트일 뿐이고, 실제 tool call 인자 검증에는 안 쓰인다. 프로퍼티 이스케이프가 있는 패턴만 지우고, lookahead나 일반 `^[a-z...]`는 남긴다. `\\p{2}`처럼 이스케이프된 백슬래시 뒤의 `p{...}`는 리터럴로 보고 유지한다. 속성 이름이 우연히 `pattern`인 경우와 `const`/`default`/`enum`/`examples` 안의 값은 건드리지 않는다. 도구 자체를 지우지 않으니 `tool_choice` 재조정과도 안 싸운다. 같은 종류의 타협은 이미 `src/adapters/kiro-tools.ts`가 Bedrock이 거부하는 validation 키워드를 지울 때 쓰고 있다(다만 Kiro는 `pattern` 전체를 지우고, 여기는 유니코드 프로퍼티만 지운다).

테스트는 단위(드롭/유지, `\P`, 이스케이프 백슬래시, name-bag, identity 반환, 5만 깊이 스택)와 passthrough wire(`_rawBody` codex forward)가 같이 들어 있다. 작성자가 `test:changed` 8818 pass와 mutation check(호출 되돌리면 wire 테스트 하나만 깨짐)까지 적었고, 전체 `bun run test`의 SIGSEGV는 수정 없는 `dev`에서도 같은 파일이라 이 PR 회귀로 보이지 않는다. types.ts/config.ts 분할 캠페인에 무효화될 성격도 아니다.

라인 - 이게 무슨 문제다

`src/adapters/openai-responses.ts` `normalizeFunctionToolSchema` - 지금은 `tool` 전체 객체를 `stripUnicodePropertyPatterns`에 넘긴다. 실제로는 `parameters` 안에만 `pattern`이 있으므로 동작은 맞고, 드롭이 없을 때 identity도 유지된다. 다만 의도를 더 드러내려면 `parameters`만 넘기고 다시 붙이는 편이 읽기 쉽다. 필수는 아니다.

`src/adapters/responses-tool-schema.ts` `stripUnicodePropertyPatterns` - `patternProperties`의 키 문자열 자체는 손대지 않는다. 작성자도 “구조가 바뀌고 실측 실패도 없어서 일부러 안 함”이라고 적었다. 키워드 `pattern`만 지우는 현재 범위는 안전하지만, 나중에 같은 400이 키 쪽에서 나면 후속이 필요하다.

`src/adapters/responses-tool-schema.ts` 공유 `SCHEMA_NAME_BAG_KEYS` - 기존 encrypted stripper의 상수 이름만 바꾼 것이고, `anthropic.ts` 등 다른 호출부는 export 함수를 그대로 쓰므로 깨질 면은 없다. 리뷰어가 rename diff에 놀라지 않으면 된다.

PR 본문 “#50” 언급 - 코드 주석과 실제 선례는 `src/adapters/kiro-tools.ts`의 Bedrock validation 키워드 제거다. 이슈 번호 표기는 헷갈릴 수 있으니 머지 전 본문만 Kiro 쪽으로 맞추면 충분하다.

CI - hygiene/label은 통과했고, enforce-target과 CodeRabbit은 이 리뷰 시점 기준 아직 pending이다. 머지 전에 green 확인만 하면 된다.

메인테이너의 판단이 필요한 지점

- `patternProperties` 키 regex까지 같은 규칙으로 지울지, 지금처럼 “실측 실패 나올 때까지 보류”로 둘지.
- normalize 입력을 `tool` 전체로 둘지, `parameters`만으로 좁힐지(동작 동일, 가독성 취향).
- 2.49.x 마감 열차에 바로 태울지, wp 번호 없이 독립 핫픽스로 `dev`에 먼저 넣을지. 슬라이스 충돌은 없다.

너의 추천

CI(enforce-target 포함) green 확인 후 `dev`에 머지. Claude Code 2.1.265+ 사용자가 GPT 경로를 쓰는 순간 세션 전체가 죽는 실측 버그이고, 범위가 좁고 테스트가 경로를 직접 짚는다. `patternProperties` 확장은 이 PR에 억지로 넣지 말고, 실패 리포트가 오면 후속으로 받는 편이 낫다. 중복/무효화 close 대상 아님.

이 댓글은 grok-bot이 작성했습니다

us4c0d3:
**Independent confirmation:** this PR fixes the observed Claude Code `Artifact` → OpenAI Codex Responses failure.

Verified at `87447a7d48a6e7f19d2c71c26b81f7c2ffb20151`:

- Complete captured Artifact schema on unpatched OpenCodex: **HTTP 400**. Removing only `properties.field.pattern`: **HTTP 200 + completion**.
- The same schema passed through this PR's actual adapter, then its serialized request sent to OpenAI: **HTTP 200 + `response.completed`**, with the original input preserved.
- The two focused test files below: **243 passed, 0 failed**. I did not independently verify the full suite or the reported baseline SIGSEGV.

Two scope points worth retaining in review: the strip also affects other destinations using these adapters, while my live verification covered only the OpenAI Codex backend; removing `pattern` weakens the advertised schema constraint rather than preserving its semantics.

<details>
<summary>Reproduction details and evidence limits</summary>

**Environment:** OpenCodex 2.48.0, Claude Code 2.1.266, macOS 15.3 arm64. Inbound `/v1/messages` → `openai-responses` → `https://chatgpt.com/backend-api/codex/responses`, model `gpt-6-astra`.

The real client tool declaration was captured at a loopback receiver. To expose Artifact in print mode, the child process enabled `CLAUDE_CODE_ARTIFACT=1`, `CLAUDE_CODE_ARTIFACT_DB=1`, and `CLAUDE_CODE_ARTIFACT_DB_STR_REPLACE=1`. With those diagnostic controls, even the first `hello` advertised Artifact without preceding tool search. These controls were not changes to the production proxy.

All provider probes used `tool_choice: none`: advertising the schema alone was sufficient to fail.

| Pattern variation | HTTP |
| --- | --- |
| Exact full client pattern, including a repeat | 400 |
| `\p{Cc}`, `\p{Cf}`, `\p{Zl}`, `\p{Zp}` individually | 400 each |
| Empty object, offending pattern removed, or simple ASCII pattern | 200 each |
| Negative lookahead alone | 200 |
| Original character-class escaping without Unicode properties | 200 |

Minimal failing parameters:

```json
{"type":"object","properties":{"field":{"type":"string","pattern":"\\p{Cc}"}}}
```

The complete captured schema was structurally identical across Claude ingress, `anthropicToResponsesBody`, `parseRequest`, adapter `_rawBody`, and decoded serialization in the installed runtime. This supports a regex compatibility gap, not an escaping change. Python `re` rejects the pattern locally; the black-box tests do not prove OpenAI's internal validator implementation.

Focused checks on the PR commit:

```sh
bun test tests/adapters/openai/openai-chat-hardening.test.ts tests/responses/openai-responses-passthrough.test.ts
```

All nine matching historical errors available locally were Claude Messages requests. The initial Codex App/deferred-tool hypothesis was not established. Historical full bodies were unavailable, so the original activation moment and pre-failure catalog diff remain unproven. Stateless replay still failed with a fresh synthetic conversation identity when the same schema was included; this is not an interactive failed-session recovery test.

This is distinct from #122's deferred `automation_update` / nested-root-`oneOf` issue. No production installation was patched, and no credentials, account identifiers, private prompts, or raw private logs are included here.

</details>

lidge-jun:
Maintainer fix round on top of 87447a7d (independent review: GO-WITH-FIXES, 0 blockers).

The runtime scope was sound, but the chat side had no test that would fail if it regressed. Every new chat-side assertion called stripUnicodePropertyPatterns directly, so reverting the normalization at src/adapters/openai-chat.ts:1334 would have left this PR green while the wire silently went back to shipping patterns Python cannot compile. Added a regression that drives buildRequest and asserts tools[0].function.parameters: the Artifact-like unescaped \p{Cc} pattern is gone, the compilable lookahead sibling survives byte-for-byte, and required is untouched.

Also narrowed the responses-tool-schema.ts comment. Saying pattern is advisory and unenforced is not accurate — strict Structured Outputs can enforce it. The tradeoff is still sound and the comment now says why: an upstream that enforces pattern could not have compiled the dropped pattern either, so it refuses the whole schema before generating any argument. The choice is a dropped constraint versus a lost request, not a silently weakened one.

Left alone deliberately: the duplicated walkers and the patternProperties name-bag keys, since no defect was reported against them.

Verification: remote CI approved on this exact head (34325974852 Cross-platform CI, 34325974807 React Doctor); local typecheck/tests NOT RUN (local, user restriction).

PR #4133, opened 2026-09-09
docs(security): document the private follow-up path for pending reports

lidge-jun:
## Summary

`SECURITY.md` explained how to file a report but stopped there. A reporter whose private
advisory had gone quiet had no stated follow-up route, no statement of what a public issue may
contain in that situation, and no way to tell an acknowledgment apart from completed triage —
which is the gap #4073 was opened to close.

Three additions, all wording only:

- **Public or private.** The test is whether the weakness is *already public* — the fix has
  shipped, or it is already described in a published advisory, issue, or pull request. Being
  findable in the source is explicitly called out as not disclosure, since opencodex is
  source-available and the opposite reading would route nearly every finding into the open. When
  unsure, file privately: a report can be moved to public review, not back.
- **Following up on a report you already filed.** Follow-up stays inside the private advisory
  thread. GitHub private vulnerability reporting is stated plainly as the only technical channel;
  there is no security email. If that thread itself is stalled, a public issue may carry
  **coordination only** — no reproduction, no exploit reasoning, no logs or attachments, no
  affected-version narrowing, and no advisory identifiers or links, since naming the report in
  public is itself a signal. A public maintainer reply confirms the route, not the content.
- **Response expectations.** Receipt is not triage: an acknowledgment, including a maintainer
  confirming they can reach the private queue, means the message arrived and nothing more. A
  landed public patch does not close the corresponding private report or settle disclosure for
  anything else. And there is **no response deadline** — stated explicitly so it is not read
  into the acknowledgment or into the triage-step list.

`README.md` carries the same three facts next to its existing security pointer.

**No response-time target or SLA is introduced.** Per @Ingwannu on #4073, a first-response target
needs @lidge-jun's agreement before it becomes a project commitment, so this PR states the
absence of one rather than proposing a number. No new reporting channel or email address is
invented either.

Closes #4073

## Verification

- `git diff origin/dev` — two files, documentation only. No `src/`, `gui/`, `scripts/`, workflow,
  or test file is touched, and nothing in the build, typecheck, or test path reads `SECURITY.md`
  or `README.md` prose (`rg 'SECURITY\.md' tests/ scripts/ .github/` finds only
  `.github/ISSUE_TEMPLATE/config.yml` and `.github/CODEOWNERS`, neither of which asserts on text).
- Independent audit of the diff and both full files by a separate agent, checking for an implied
  SLA, an invented channel, contradictions with the existing text, any wording that would push
[...]

PR #4142, opened 2026-09-09
fix(web-search): opt-in hosted web-search bridge for the key-auth Responses passthrough

### LINKED ISSUE #3761: [Provider compatibility] Ollama Cloud Responses passthrough skips web-search sidecar; hosted web_search is echoed as a client function_call
### Deferred; no hosted-search execution fix landed

The four-track delivery did not implement a raw Responses passthrough search/continuation bridge. Search-result insertion, mixed-tool continuity and replay semantics remain outstanding. This issue stays open.

Verified against dev `5759d9ea2f1e7281cdc01eb9628f2e0a123fb59c`. Original report by @foo1maker. The attribution record was added in [#3811](https://github.com/lidge-jun/opencodex/pull/3811).

### Client or integration

Codex App

Ollama Cloud (`https://ollama.com/v1`, key-auth `openai-responses` passthrough)

### OpenCodex version

2.43.0 (`@bitkyc08/opencodex`)

### Endpoint or capability

`/v1/responses` + hosted `{type:"web_search"}` / `webSearchSidecar`

### Current behaviour

Codex App always speaks Responses to OpenCodex (`POST /v1/responses`) and advertises the hosted tool `{type:"web_search"}`.

For a key-auth provider with `adapter: "openai-responses"` (Ollama Cloud GLM / DeepSeek, etc.) OpenCodex takes the **passthrough** branch in `src/server/responses/core.ts` and returns before `runWithWebSearch`. `planWebSearch` / `shouldResolveOpenAiWebSearchSidecar` also fail closed on `isPassthrough`:

```ts
// src/web-search/index.ts
if (!parsed._webSearch || isPassthrough) return undefined;
```

The comment assumes passthrough means ChatGPT, which already executes hosted search server-side.

Ollama Cloud does **not** execute OpenAI hosted `{type:"web_search"}` on `/v1/responses`. The model instead emits a client `function_call` named `web_search`. Two failures follow:

1. Stock `responses-undeclared-tool-guard` treats that echo as an undeclared client tool and aborts the SSE stream:
   `routed provider emitted undeclared client tool "web_search"; only request-declared tools may be called`
   Codex App then reconnects 5/5 and the worker/turn dies (`last_agent_message: null`).

2. If the guard is relaxed, Codex App still does not execute a function named `web_search` (it expects the **provider** to run hosted search). The next hop is `unsupported call: web_search`. The model can continue without live results. Dashboard `webSearchSidecar` (e.g. gpt-5.6-luna) stays idle.

Ollama *does* have a working search API, but it is a **separate client-executed REST endpoint**, not ChatGPT hosted search:

- `POST https://ollama.com/api/web_search`
- docs: https://docs.ollama.com/capabilities/web-search

Official OpenCodex backends (`openai` / `anthropic` / `xai` / `gemini` / `exa`) never call it. The sidecar loop that would intercept a synthetic `web_search` function tool never runs on this passthrough path.

Switching the provider to `openai-chat` would enter the sidecar loop, but Codex App cannot be switched off Responses, and translating GLM/DeepSeek off `/v1/responses` drops reasoning summaries and the native tool surface that this client requires.

### Expected behaviour

On key-auth Responses passthrough to a gateway that does **not** execute OpenAI hosted `web_search`:

1. Do not abort the stream for a `function_call` named `web_search` when the request catalog already declared hosted `{type:"web_search"}` (same catalog entry, different wire shape).
2. Intercept that call in the proxy (do not relay it to Codex as an unsupported client tool).
3. Execute search and inject results as `function_call_output` (and/or a hosted `web_search_call` item Codex understands).
4. Prefer Ollama's own `POST /api/web_search` when the route is Ollama Cloud (same credential, no ChatGPT sidecar spend). Luna/Exa remain valid fallbacks for translated routes, which is the current sidecar design.

Passthrough should keep meaning “preserve the Responses wire”, not “the upstream is ChatGPT and already searched”.

```json
{
  "defaultProvider": "ollama-relay",
  "webSearchSidecar": { "enabled": true },
  "providers": {
    "ollama-relay": {
      "baseUrl": "https://ollama.com/v1",
      "authMode": "key",
      "selectedModels": ["glm-5.3-flash"]
    },
    "openai": {
      "authMode": "forward"
    }
  }
}
```

```toml
# Codex App
model = "glm-5.3-flash"
```

1. `ocx start` (2.43.0 stock).
2. In Codex App, ask a routed GLM/DeepSeek turn for current web facts (pricing, docs, anything that triggers search).
3. Observe either the undeclared-tool abort or `unsupported call: web_search` with no sidecar / no `POST https://ollama.com/api/web_search`.

### Actual response or error

```text
stream disconnected before completion: routed provider emitted undeclared client tool "web_search"; only request-declared tools may be called
```

If that guard is bypassed, the Codex client hop is:

```text
unsupported call: web_search
```

Upstream GLM item shape (redacted):

```json
{
  "type": "function_call",
  "name": "web_search",
  "arguments": "{\"query\":\"site:example.com current pricing\"}"
}
```

OpenCodex logs show no `[web-search]` sidecar enter on these turns. Ollama `/api/web_search` is not requested.

### Upstream documentation

- Ollama web search REST (client-executed): https://docs.ollama.com/capabilities/web-search
- Ollama OpenAI compatibility (`/v1/responses` is non-stateful; hosted ChatGPT `web_search` is not listed as provider-executed): https://docs.ollama.com/api/openai-compatibility
- OpenAI hosted `web_search` tool (what Codex App sends): https://platform.openai.com/docs/guides/tools-web-search

Stock code already has the intercept loop for **translated** adapters (`runWithWebSearch` + `buildWebSearchTool`). The gap is that `adapter.passthrough` both:

- skips that loop (`planWebSearch(..., isPassthrough)` / early return in `core.ts`), and
- forwards hosted `{type:"web_search"}` unchanged, which Ollama echoes as a named `function_call`.

A minimal passthrough-aware path:

1. In `addNamelessClientCallTypes`, treat hosted `web_search` / `web_search_preview` as authorizing a client `function_call` named `web_search` (same catalog, echo shape). This at least stops Mode 2 / App workers from 5/5-dying.
2. On non-forward (`authMode: "key"`) passthrough SSE, intercept `function_call` `name === "web_search"`, run an executor, hide the synthetic call from Codex, inject the tool result into the next upstream hop.
3. Add an `ollama` web-search backend that POSTs `https://ollama.com/api/web_search` with the existing provider key. Keep luna for `openai-chat` / Anthropic routes.

Do **not** require operators to change Codex App to Chat Completions. The App wire is Responses; Ollama Cloud also speaks `/v1/responses`. The missing piece is executing search for that pair.

### Additional context

Searched existing issues. Closest prior art is passthrough tool-surface gaps (e.g. #1950 tool_search rewrite on Responses passthrough) and web-search sidecar bugs on **translated** openai-chat routes (#265, #1001). I did not find an open issue for Ollama Cloud hosted-`web_search` echo vs `/api/web_search`.

Related but separate: Ollama `/v1/responses` also lacks `previous_response_id` store (documented). `statelessResponses` addresses replay; it does not execute search.

### Checks

lidge-jun:
## Summary

Codex always POSTs `/v1/responses` with the hosted `{type:"web_search"}` tool. On the KEY-auth Responses passthrough OpenCodex reads that declaration as "the destination executes search itself" and relays it unchanged, which is true for the ChatGPT backend and for xAI. It is false for an OpenAI-shaped key gateway that does not run the hosted tool: Ollama Cloud GLM answers with a plain `{type:"function_call", name:"web_search"}`, nothing on either side executes it, and the undeclared-tool guard ends the turn with `undeclared_tool_call` because a hosted declaration never authorizes a client function name.

This adds an opt-in, default-OFF bridge for exactly that case: `providers.<name>.webSearchBridge = { enabled: true, backend: "ollama" }`.

When it arms, the passthrough SSE body is wrapped before the relay rewrites. The bridge intercepts the `web_search` call out of the upstream stream, runs the configured search backend itself, appends the call plus its `function_call_output` to the same raw outbound body, POSTs the next leg to the same upstream, and emits the hosted `web_search_call` frames Codex already renders (the shape `src/bridge.ts` produces for the sidecar). The stream it hands back is ordinary Responses SSE, so every existing rewrite — including the undeclared-tool guard, the payload repairs, terminal-outcome recording, and the continuation cache — still applies to the client-facing stream.

Before/after with the bridge enabled on an Ollama Cloud provider, for a turn where GLM asks to search:

| | client stream | second upstream body |
| --- | --- | --- |
| before | `response.failed` / `undeclared_tool_call` | none |
| after | `web_search_call` in_progress + completed, then the answer | `function_call` + `function_call_output` carrying the search result |

Arming is fail-closed on every axis. It requires `parsed._webSearch`, the passthrough adapter, `authMode: "key"`, `enabled: true`, a `tool_choice` that still allows search, and a streaming turn. It never arms for `authMode: "forward"` (ChatGPT already searches with the caller's own credential) or for a stored OAuth credential, and it never pre-empts a provider that executes hosted search upstream. This is a new planner, `planPassthroughWebSearchBridge`, rather than a relaxation of `isPassthrough` in `planWebSearch`: the sidecar rewrites normalized messages, while this path has to preserve the raw Responses conversation.

`web_search` is **not** added to the guard's allowed names. The call is removed from the stream because the proxy answers it; an unrelated undeclared tool still fails closed, and that is pinned by a sibling fixture.

Backend and destination policy:

- Only `ollama` has a shipped executor. `openai`, `anthropic`, `xai`, `gemini`, and `exa` are accepted by the config union and stay inert, the same explicit-only contract `webSearchSidecar` uses for backends whose executor has not landed. Nothing auto-selects a paid Luna or Exa search.
- The `ollama` backend derives `POST <origin>/api/web_search` only when the provider's `baseUrl` origin is `https://ollama.com`. Any other origin requires an explicit `endpoint`, because the bridge reuses the provider's own route key on that URL and an operator naming the endpoint is the authorization for it. The executor sends `redirect: "manual"` so a redirect cannot carry the key elsewhere, and scrubs the literal key out of every error string.

Deliberate boundaries of this first slice, kept narrow on purpose:

- Streaming turns only; a non-streaming turn stays on the existing path.
- A leg that mixes the search call with any other client tool call fails closed with an explicit `web_search_bridge_mixed_tools` error instead of dropping the client's call. Answering both needs the raw mixed-tool continuation contract the 2.47 track deferred (`devlog/_plan/260907_track2_protocol/040_hosted_search_disposition.md`), and this PR does not open that.
- Continuation legs use a direct send rather than the core recovery ladder. The first leg still goes through it, and a KEY-auth destination has no OAuth refresh to replay on a later leg.
- The client stream is renumbered (`sequence_number` and `output_index`) because events are both dropped and injected; a plain relay cannot preserve upstream numbering through that. The terminal `response.output` is rebuilt from the items the client actually received so the snapshot matches the streamed turn.
- The bridge's own SSE buffer is bounded (8 MiB) rather than charged to the translator budget.

With the flag off the change is one planner call that returns `undefined`; the relay is otherwise untouched.

Closes #3761

## Verification

Local checks were **NOT RUN** per maintainer instruction: no `bun run typecheck`, no `bun run test`, no `bun run lint:gui`, no `bun run build:gui`, no `bun install`. The exact-head remote CI on this PR is the gate.

New coverage in `tests/web-search/web-search-passthrough-bridge.test.ts`:

- Arming matrix for `planPassthroughWebSearchBridge`: armed only for an enabled ollama-backed key provider on the canonical origin; disarmed without the opt-in, for `forward` and `oauth` auth, off the passthrough, without hosted `web_search`, for non-streaming turns, for a `tool_choice` that excludes search, for every backend without an executor, and for a non-canonical origin with no operator `endpoint`. Out-of-range bounds fall back to the documented defaults.
- Runner behavior against a fake first leg, a fake `send`, and a fake executor: the `web_search` `function_call` never reaches the client, the hosted `web_search_call` added/done pair carries the expected id, status, `action`, and `sources`, the second upstream body carries the call and its `function_call_output`, both legs land in one monotonic client numbering, and the terminal snapshot matches the streamed items.
- A turn with no search is relayed untouched and never re-sends; a mixed-tool leg fails closed without sending or executing; the search budget is bounded so the turn terminates instead of looping; an executor failure is reported as the tool result rather than as a dead turn.
- End to end through `handleResponses` with a stubbed upstream: with the flag off the reported `undeclared_tool_call` abort still happens; with the flag on the client sees the hosted cell plus the answer and the search result reaches the second upstream body; an unrelated undeclared tool (`frobnicate`) still fails closed through the bridged stream.

Docs: `providers.md` gains the `webSearchBridge` row. Test layout registered in `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`.

### Independent audit round

A grok-4.6 subagent audited the diff against the source; four real defects were found and fixed in `3557ada7d`:

- A leg that mixed the search call with another client tool call streamed that tool call and *then* failed the turn, so Codex could start running it under a turn that never completes. Client-executed call events are now withheld for the leg and released only when the turn actually ends there; on the mixed-tool failure they are dropped and the opened search cell is closed as `failed` instead of leaving a spinner running.
- The hosted `web_search_call` cell was injected after the whole leg, so a search that was not the last output item was reordered after later assistant items. The cell now opens at intercept time and keeps that `output_index`; a reserved snapshot slot keeps the terminal `response.output` in the same order the client saw.
- A cancelled client could still bill a search and open a continuation. Cancelling the bridged stream latches a local abort, and the abort state is now checked before each execute and each send.
- The continuation body bypassed the outbound size ceiling the first leg was admitted under. Core now re-applies `checkOutboundBodySize` to every continuation body and the bridged turn fails rather than sending an unbounded one.

Suppression is also keyed on `item_id` as well as `output_index`, so an argument event that carries only `item_id` cannot leak the intercepted call name.

Known limitation, stated rather than fixed here: on the Darwin `auto` tee path a client disconnect cancels the client branch while the inspection branch still drains, and `upstream.abort()` runs at the end of that bounded drain. During that window the bridge can complete one search and one continuation before its abort check fires. It is bounded by `maxSearches` and the existing drain bounds; closing it means changing where the relay tees, which is outside this PR.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added an optional web-search bridge for key-authenticated Responses passthrough providers.
  - Hosted web searches can now be executed through Ollama and returned as standard hosted search results.
  - Added configurable search limits, timeouts, endpoints, and backend settings.
  - Malformed bridge settings are safely ignored without preventing configuration loading.
- **Documentation**
  - Documented configuration, defaults, streaming requirements, and supported usage constraints.


lidge-jun:
## 리뷰 · 우선순위 72 / 80

이 PR은 Codex가 항상 선언하는 hosted ``web_search`` 도구를, **키 인증 Responses 패스스루**에서 그대로 중계할 때 생기는 구멍을 막는 옵트인 브리지입니다. 지금 `dev`(HEAD ``f982af48b``, tip #4133 보안 문서)에서는 ChatGPT/`authMode: forward`나 xAI처럼 업스트림이 검색을 직접 실행하는 쪽은 중계가 맞고, Ollama Cloud GLM처럼 OpenAI 모양 KEY 게이트웨이는 ``function_call name:web_search``만 내보낸 뒤 아무도 실행하지 않아 undeclared-tool 가드가 턴을 끊습니다. 그게 #3761입니다.

이번 변경은 기본 OFF인 ``providers.<name>.webSearchBridge``를 켠 제공자에 한해, 그 호출을 스트림에서 가로채고 Ollama ``/api/web_search``로 검색을 실행한 다음 같은 업스트림에 continuation POST를 보내고, Codex에는 익숙한 hosted ``web_search_call`` 셀을 보여 줍니다. 새 파일 ``src/web-search/passthrough-bridge.ts``·``ollama-executor.ts``가 핵심이고, ``src/server/responses/core.ts`` 패스스루 SSE 분기에서 플래너가 무장할 때만 스트림을 감쌉니다. 타입·설정은 ``src/types/provider.ts`` / ``src/config.ts`` / management ``auth-cors.ts``에 들어가고, 문서와 레이아웃·테스트까지 같이 옵니다. 기본 경로를 바이트 단위로 유지하면서 Ollama Cloud 실사용 실패를 고치려는 방향이라, 현재 `dev`의 2.50.0 / pre-2.49.0 라인에서도 가치가 큽니다.

설계상 좋은 점: (1) ``enabled: true`` + ``backend: "ollama"``가 둘 다 있어야만 무장, (2) forward/oauth는 절대 안 탐, (3) 캐논 origin ``https://ollama.com``이 아니면 endpoint를 운영자가 명시해야 키가 나감, (4) redirect manual + 키 리터럴 scrub, (5) 다른 클라이언트 툴과 섞이면 fail-closed, (6) 브리지가 만든 스트림을 다시 core rewrite/undeclared-tool 가드 아래로 통과시킴. 테스트도 무장 조건·스트림·혼합 툴·undeclared 잔존을 꽤 촘촘히 잡았습니다.

라인 단위로 보면 아래가 남습니다.

`src/web-search/passthrough-bridge.ts` `planPassthroughWebSearchBridge` - 주석·문서는 "업스트림이 hosted 검색을 직접 실행하는 제공자(예: xAI)에는 무장하지 않는다"고 하지만, 코드 조건은 `authMode === "key"` + enabled + backend ollama + endpoint뿐입니다. 키 인증 xAI 행에 실수로 켜면 진짜 업스트림 검색을 프록시 Ollama 검색으로 바꿔 버릴 수 있습니다.

`docs-site/.../providers.md` `webSearchBridge?` 타입 칸 - 표기에는 `backend?: "ollama"`만 있는데, 스키마/유니온은 openai|anthropic|xai|gemini|exa도 받습니다. 본문에는 inert라고 적혀 있어 충돌은 아니지만, 타입 칸만 보면 다른 id가 거절되는 것처럼 읽힙니다.

`src/config.ts` `providerConfigSchema.webSearchBridge` - 로드 시 `.catch(undefined)`로 잘못된 블록을 조용히 지웁니다. management 쓰기는 크게 거절하니 의도는 맞지만, 파일 편집으로 오타 낸 운영자는 "켰는데 안 됨"을 설정 무효로 못 볼 수 있습니다.

`src/server/responses/core.ts` continuation `send` - 첫 다리는 recovery ladder를 타지만 이후 다리는 직접 fetch입니다. KEY-auth라 OAuth refresh가 없다는 설명은 타당하고, 다만 중간 다리의 일시 5xx/연결 리셋은 기존 transient 재시도 밖에 있습니다(의도된 1차 슬라이스인지 확인).

메인테이너의 판단이 필요한 지점
- #3761을 이 PR로 close할지, 아니면 "streaming + ollama-only + mixed-tools fail-closed" 한계를 이슈에 남기고 partial close로 둘지
- xAI 등 hosted-search 제공자를 플래너에서 명시적으로 배제할지(문서 약속을 코드로 고정), 아니면 운영자 설정 책임으로 둘지
- 설정 로드 시 invalid bridge를 silent drop 유지 vs soft-warn/diagnostic 노출
- types.ts/config.ts 분리 캠페인 관점: 이번 필드는 새 기능이라 무효화 대상이 아님. 다만 큰 패스스루 코어 터치라 다른 responses 열차와 충돌하면 리베이스보다 랜딩 순서만 맞출지

너의 추천
원격 CI가 초록이면 **merge 쪽으로 진행**하세요. 머지 전에 (1) xAI/hosted-search 제공자 배제를 플래너에 한 줄 넣거나, 문서 문장을 "운영자가 켜면 대체된다"로 고치고, (2) docs 타입 칸을 유니온과 맞추고, (3) #3761 acceptance를 이 슬라이스 범위로 짧게 갱신한 뒤 close 여부를 정하면 됩니다. 기본 OFF·fail-closed 성격상 머지 위험은 낮고, Ollama Cloud 실사용 구멍을 지금 `dev`에서 막는 이득이 큽니다.

이 댓글은 grok-bot이 작성했습니다

PR #4156, opened 2026-09-09
feat(catalog): classify discovered model pricing and filter free models

lidge-jun:
## Summary

Live `/models` rows already carried a `pricing` object, but nothing downstream kept it. `catalogHintsFromModelsApiItem` returned only window / modalities / reasoning / capabilities, so `CatalogModel` had no cost field and both Dashboard catalog lists could only substring-search. The "Free" control that already existed is **provider tier** (`freeTier` / `keyOptional`), which is exactly why OpenRouter — `freeTier` unset, paid and `:free` slugs mixed in one catalog — could not be narrowed to $0 models.

This adds `discoveredPricingStatus`, which reads the provider's own `prompt`/`completion` (or `input`/`output`) rates and classifies a row as free, paid, or unknown, and then wires that field into the catalog projection and its three consumers.

The classifier is fail-closed. Only a complete pair of non-negative numeric rates classifies at all, so a missing, one-sided, non-numeric, or negative rate stays unknown and is excluded from a free-only filter. Showing a paid model under a Free filter spends the user's money; hiding a free one costs a click. Two things that look like evidence are deliberately not used: a `:free` id suffix is an OpenRouter naming convention rather than a price — Nous ships `:free` slugs on a provider whose `freeTier` is `false` on purpose — and the operator's `modelCosts` overlay is an estimate they typed rather than something the provider published. Classification is on numeric zero, never on a unit conversion, since OpenRouter quotes USD per token while the overlays quote per 1M and zero is zero in both.

`catalogHintsFromModelsApiItem` **omits** the field when the classification is unknown rather than emitting the literal string. That follows the function's existing contract that an unknown property is absent rather than present-and-empty, which ten deep-equal assertions in `tests/providers/provider-model-discovery-contract.test.ts` and `tests/providers/featherless-provider.test.ts` already encode. Callers that need to distinguish "the provider published no prices" from "this build does not classify" call `discoveredPricingStatus` directly.

`listManagementModelRows` already spreads the `CatalogModel`, so `GET /api/models` carries the field with no change there, and `deriveEntry` builds Codex catalog entries from an explicit field list, so it is never serialized into the catalog and never affects routing or visibility.

Consumers:

[...]

PR #4158, opened 2026-09-09
fix(models): explain the discovery dependency when model sync fails

### LINKED ISSUE #4075: Improve Gemini provider setup UX: unclear model discovery dependency causes misleading “model sync failed” state
### Area

Catalog / models

I was trying to add a Gemini API provider and use gemini-3.8-flash as a custom model in OpenCodex.
The provider itself was configured correctly, including the Gemini API endpoint and API key. I also added the model manually in the model configuration.

### What prevents this today?

The setup workflow is confusing because several related settings are distributed across different pages, while the dependency between them is not clearly communicated.
In my case, the provider had “Discover models from provider” enabled. However, I was adding the model manually rather than relying on provider-side model discovery.
This resulted in a “model sync failed” / “发现失败” state, but the UI did not clearly explain:
- that model discovery was enabled;
- that this setting was relevant to the model configuration;
- why the manually added model could not be used in this state;
- that disabling “Discover models from provider” would resolve the problem.
The relevant setting is located on the provider's Settings page, while the resulting model state is presented on a separate Models page. As a result, the problem appears to be an API/provider/model failure rather than a configuration or workflow issue.
I spent a considerable amount of time troubleshooting the API configuration before eventually discovering that disabling “Discover models from provider” fixed the problem.
The main issue is therefore not that the setting exists, but that the UI does not make the relationship between these settings and the resulting error state sufficiently clear.

### What should OpenCodex do?

OpenCodex should make this workflow more discoverable and provide actionable diagnostics.
For example:
1. If “Discover models from provider” is enabled, clearly indicate that OpenCodex will attempt to synchronize the provider's model list.
2. If a user manually adds a custom model while provider model discovery is enabled, explain whether the two modes are compatible.
3. If model synchronization fails, show a meaningful error explaining the likely cause instead of only showing a generic “model sync failed” state.
4. When a model is unavailable because of a related provider setting, provide a direct link or navigation path to the relevant setting.
5. Consider grouping provider-level model discovery and model-level configuration into a more coherent workflow, or explicitly showing their relationship.
More generally, the UI should expose configuration dependencies rather than requiring users to discover them by trial and error.

A possible improvement would be an inline message such as:
Model discovery is enabled for this provider.
OpenCodex will attempt to discover models automatically. If you want to manage models manually, disable “Discover models from provider” in Provider Settings.

And if synchronization fails:
Model discovery failed.
Check the provider configuration or disable “Discover models from provider” if you want to use manually configured models.
[Open Provider Settings]

Alternatively, the Models page could display the dependency directly:
Model discovery: Enabled
Models are currently managed through provider discovery.
To use manually added models, disable model discovery in Provider Settings.

The important point is that the user should be able to understand what is wrong and where to fix it from the page where the problem is displayed.

The workaround I eventually discovered was:
1. Open the provider's Settings page.
2. Disable “Discover models from provider”.
3. Keep the Gemini model configured manually.
4. The manually configured model can then be used normally.
However, this workaround is difficult to discover because the relevant setting is on a different page and the model synchronization error does not point to it.

### Additional context

This was particularly difficult to diagnose because the provider configuration itself appeared reasonable:
- Provider: Google
- Base URL: https://generativelanguage.googleapis.com
- Authentication: API key
- Custom model: gemini-3.8-flash
The model was successfully added to the model configuration, but its status still indicated that it was unavailable / disabled, while the provider displayed a “发现失败” (discovery failed) state.

### Checks

lidge-jun:
## Summary

A provider whose live model fetch fails gets an amber "Discovery failed" badge on its group header and nothing else. `EmptyProviderHint` already carries the guidance and the link into provider settings, but it only renders when the group has **no** rows, so the reporter — who had added a Gemini model by hand — saw a failure badge next to a model that would not work, with nothing connecting the two.

The mechanism they eventually found on their own is that discovery being **on** is what holds those rows back. A newly added key provider is stamped `initialModelSelection.status = "pending"`; failed discovery is degraded, so initialization never finalizes; pending rows are then forced `disabled` and dropped from the Codex catalog. Turning "Discover models from provider" off makes the seed authoritative and releases them, which is exactly the workaround the report describes.

`DiscoveryDependencyHint` renders that explanation for a failed group that **has** rows, alongside the same settings link the empty-group hint uses. The control's name is interpolated from `pws.liveModels` rather than restated, so the sentence names the real switch in all nine locales and cannot drift from its label if that label is ever reworded. The link uses the existing `navigateHash("providers")`: `hashBelongsToPage` has no `providers/<name>` arm and rewrites `providers/workspace` to `providers`, so an invented per-provider deep link would be normalised out of the URL bar.

For the record, the English interface never says "model sync failed" — that is the reporter's paraphrase of `models.discoveryFailedBadge`.

## Deliberately out of scope

Both of these are maintainer policy rather than copy, and neither is needed to close the issue as filed:

- Flipping the registry `google` entry to `liveModels: false` would change seed behaviour for every Gemini install.
- Treating degraded discovery plus configured rows as authoritative enough to finalize `initialModelSelection` contradicts `tests/providers/initial-model-selection.test.ts:278-285`. That second one is what would make a manually added model usable *without* turning discovery off; it deserves its own issue rather than being folded in here.

## Verification

Local checks: **NOT RUN**. The maintainer set an explicit constraint for this delivery round that no local product suite runs — no `bun test`, `bun install`, `bun run typecheck`, `bun run build`, `bun run test:changed`, lint, or `privacy:scan`. Exact-head remote CI is the only gate for this PR.

New regression coverage in `gui/tests/models-discovery-failed-hint.test.tsx`, next to the existing `models-empty-provider` harness it borrows its setup from. It mounts the real page against one provider that has a row, and asserts:

- with `discovery: { status: "failed", reason: "http", httpStatus: 401 }`, the rendered group keeps the "Discovery failed" badge **and** now also names the mechanism and the exact control, with the settings link. The badge alone is the state before this change, so the first two assertions are what turn red without it.
- with `{ status: "ok" }` and with no discovery state at all, neither the badge nor the new sentence appears.
- the hint's link renders as a `link-btn` inside a `role="status"` region and carries no `providers/<name>` target.

The new key `models.discoveryFailedDependency` was added by hand to all nine locale catalogs with real translations rather than English placeholders. No layout registration was needed: the test lives under `gui/tests/`, which is outside the `tests/` tree that `scripts/test-layout/layout.json` maps.

No screenshot is attached. Producing one requires building the dashboard, which this round's no-local-build constraint forbids. The visual change is one informational line inside an existing provider group — the same icon, tone, and settings link the empty-group hint already renders, appearing under the group header when discovery has failed and the group still lists models.

## Checklist

- [x] Docs or release notes were updated when needed. This is interface copy for an existing control that is already documented; no docs page states the old behaviour.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. No auth, credential, workflow, or release-automation surface is touched. The hint renders no provider detail — the failure reason it sits beside is the existing server-owned, sanitized summary.

Closes #4075.

## Summary by CodeRabbit

* **New Features**
  * Added a clear message to provider model groups when model discovery fails repeatedly.
  * Explained how to disable live model discovery and use manually added or static models instead.
  * Added a direct link to provider settings from the guidance message.
  * Added translations for the message in multiple supported languages.

* **Bug Fixes**
  * Improved visibility of failed model discovery dependencies in the Models page.


lidge-jun:
## 리뷰 · 우선순위 64 / 80

이 PR은 이슈 #4075를 닫으려는 Lane B 두 번째 단위다. 지금 `dev`(HEAD `a7509fe00`, 패키지 2.50.0)에서는 Models 페이지 프로바이더 그룹에 라이브 모델 검색이 실패하면 헤더에만 노란 `Discovery failed` 배지가 붙고, 안내 문장(`EmptyProviderHint`)은 **행이 하나도 없을 때**만 나온다. 그래서 사용자가 Gemini 같은 키 프로바이더에 모델을 손으로 추가해 둔 상태에서는, 배지와 “안 되는 모델 행”만 보이고 둘을 잇는 설명이 없다. 실제로 막히는 이유는 discovery가 켜진 채 실패한 뒤 `initialModelSelection.status = "pending"`이 끝나지 않고, pending 행이 disabled로 강제되어 Codex 카탈로그에서 빠지기 때문이다. 프로바이더 설정의 “프로바이더에서 모델 검색”(`pws.liveModels`)을 끄면 시드가 권위가 되어 풀린다. 이 PR은 그 문장을 `DiscoveryDependencyHint`로, **행이 있는 실패 그룹**에도 같은 설정 링크와 함께 보여 준다.

코드 경로는 현재 체크아웃과 맞다. `gui/src/pages/models-provider-hints.tsx`에 힌트 컴포넌트를 추가하고, `gui/src/pages/Models.tsx`의 `renderGroup`에서 `rows.length > 0 && discoveryFailure`일 때만 렌더한다. 컨트롤 이름은 문구를 다시 쓰지 않고 `t("pws.liveModels")`를 끼워 넣어 9개 로케일이 라벨과 어긋나지 않게 했다. 링크는 기존 `navigateHash("providers")`만 쓴다. 딥링크 `providers/<name>`은 `hashBelongsToPage`에 없고 정규화로 사라지므로 의도적으로 피한 선택이다. 로케일 키 `models.discoveryFailedDependency`는 en/ko 포함 9개에 실번역이 들어 있다. 회귀 테스트 `gui/tests/models-discovery-failed-hint.test.tsx`는 실패/정상/discovery 없음과 링크 해시를 검증한다.

같은 브랜치 두 번째 커밋은 #3666 후속이다. `pricingStatus`가 discovery 힌트 → `applyProviderConfigHints` → `listManagementModelRows` → `GET /api/models` 네 홉을 살아남는지 `tests/codex-integration/catalog-free-pricing-status.test.ts`에 와이어 회귀를 넣고, Provider Workspace의 Free only 스위치를 Models 페이지와 같이 검색창 **위**로 맞췄다. 베이스는 `lane-b/1-3666`(드래프트)이라 #4156(무료 모델 필터) 스택 위에 얹힌 형태다. round-2 플랜에서 #4075는 Lane B의 기계적 UX 단위로 적혀 있고, 본문이 명시한 정책 변경(레지스트리 `google`의 `liveModels: false`, degraded discovery + 설정 행을 authoritative finalize로 취급)은 의도적으로 빼 두었다. 후자는 `tests/providers/initial-model-selection.test.ts` 278–285행의 “degraded discovery does not complete initialization” 계약과 충돌한다.

라인 수준의 문제는 치명적이지 않다. 드래프트·스택·범위 혼합·컴포넌트 계약 정도다.

gui/src/pages/models-provider-hints.tsx (DiscoveryDependencyHint) - props가 없어서 부모 게이트(`rows.length > 0 && discoveryFailure`)에만 의존한다. 지금은 맞지만, 다른 화면에서 그대로 꽂으면 “검색이 켜져 있다”는 문장이 항상 나온다.
gui/src/pages/Models.tsx (힌트 삽입부) - 행이 보이기만 하면(실제 picker에 안 나가도) 힌트가 뜬다. #4075 증상과 맞아서 의도일 가능성이 크지만, “보이는 행만”으로 좁힐지는 선택지다.
PR 제목/본문 vs 두 번째 커밋 - 제목은 #4075 카피 수정인데, #3666 와이어 핀·Free 스위치 정렬이 같이 들어 있다. 스택 후속으로는 자연스럽지만 리뷰어가 한 이슈로만 읽기 쉽다.
베이스 lane-b/1-3666 / draft - `dev`에 바로 머지할 수 없다. #4156(또는 lane-b/1)이 먼저 안정돼야 한다.
검증 - 라운드 제약으로 로컬 스위트 미실행, remote CI만. 새 GUI 테스트·와이어 테스트가 있으므로 CI 그린이 실질 게이트다.

메인테이너의 판단이 필요한 지점
- Lane B 순서대로 #4156 → 이 PR을 올릴지, 아니면 #4075 카피만 쪼개 `dev`에 단독으로 올릴지
- 본문이 빼 둔 정책 두 가지(구글 기본 liveModels 끄기, degraded+수동 행 finalize)를 후속 이슈로 열지, 당분간 카피만으로 #4075를 닫을지
- 드래프트를 CI 그린 후 바로 Ready로 올릴지, lane-b/1 머지 확인까지 기다릴지

너의 추천
#4156(lane-b/1-3666)이 `dev`에 안착한 뒤 이 드래프트를 Ready로 바꾸고, 정책 확장 없이 카피+힌트+회귀 테스트(+ #3666 와이어 핀)만으로 #4075를 닫고 머지하는 쪽이 맞다. google liveModels 기본값 변경이나 initialModelSelection finalize 완화는 별 이슈로 남겨 두자. CI가 깨지면 그 테스트부터 고치고, 범위는 더 넓히지 말 것.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
**Screenshot gate waived by the project owner.**

This is a deliberate waiver of the screenshot requirement on a real GUI change, and it is not a claim that the gate misfired. The `gui-screenshot-waived` label is described in this repository as a waiver for false-positive screenshot requirements; that is a narrower reading than the use here, so the record should say plainly what happened.

The visual states this round adds — a decode-rate figure stacked on a live request row, a Free-only catalog filter over discovered pricing, and an inactive badge driven by real quota exhaustion — do not render from a static build. Each needs a running proxy sitting in a specific upstream state, so a screenshot here would cost a live reproduction rather than a build step. The description carries a written account of the surface instead.

The decision and its reasoning are recorded in `devlog/_plan/260910_post249_round2/120_landing_gate_decision.md`.

PR #4165, opened 2026-09-09
feat(catalog): mark quota-exhausted models and combos inactive

lidge-jun:
## Summary

Served catalog rows were always stamped `visibility: "list"` with no field that could say "offered, but a request would fail right now". The quota was already known — `ProviderQuota.creditsUsd.remaining` is written on probe and read back within 30 minutes, and the combo loop already refuses exhausted targets and throws `NoAvailableComboTargetsError` — but it never reached the catalog. The catalog's only "inactive" mechanisms **remove** the row: the live visibility filter, the disabled-routed-key merge drop, and native `visibility: "hide"`. Hiding is precisely what this issue rejects.

`quotaInactiveReason()` returns `"no_credit"` when every **usable** target has positive exhaustion evidence. It reuses the runtime rules in `targetProviderIsUsable` rather than the Dashboard's `quotaStateFromReport`, which is harsher: that one treats `remaining <= 0` as exhausted without requiring `percent >= 100`, and ignores an elapsed `resetAt`. A row marked inactive on the harsher rule would contradict the router, which would still send the request.

Three inherited rules carry the correctness. A removed or disabled target drops out of the vote rather than counting as evidence, and if nothing usable is left the row is unavailable for an operator reason rather than a quota one. The canonical ChatGPT forward provider is exempt, because native account selection owns model-scoped quota and a provider-level summary cannot veto it. A stale cache is not exhaustion: `getCachedProviderQuota` returns null past its 30-minute window, and a null reading ends the vote, so an unprobed provider is never advertised as out of credit. "Every usable target" is the bar because one target that can still serve makes the row serviceable — the same conclusion the request path reaches when it hops past an exhausted target.

The field is stamped once, at the single point that has both the finished model list and the config, so routed, combo, and custom rows are treated alike. `deriveEntry` writes it as `opencodex_inactive_reason`, matching the existing `opencodex_*` extension style. `visibility` is never touched, the operator-disable filter is not involved, and `ManagementModelRow.disabled` is not reused, since that means the operator's own `disabledModels`.

## The limitation, stated plainly

**A custom catalog field cannot grey out the native picker.** Codex Desktop and app-server understand only `visibility: "list" | "hide"` and hold an in-memory roster, so they ignore an unknown field. This marks the entry for OpenCodex-aware consumers — the dashboard — and does nothing in the native picker. Using `hide` instead is exactly what the issue asks not to do.

[...]

Issue #1711, opened 2026-08-14
[Feature]: grey out zero-credit models/combos in the served model catalog and picker


ardjo-s:
### Area

Catalog / models

### Goal

Make the model picker reflect known quota exhaustion without removing entries from the catalog.

When a routed model or combo has zero remaining credit across all of its routable targets, OpenCodex should mark that catalog entry as inactive so picker consumers can grey it out and explain why it is unavailable. The entry should become active again automatically when quota resets or refills.

### Current problem

Models and combos with known zero remaining credit can still appear as normal selectable entries in the catalog-backed model picker. Users can therefore select an option that OpenCodex already knows cannot currently serve a request.

### In scope

1. **Catalog state.** Represent known all-target quota exhaustion in the served model catalog without hiding the model/combo.
2. **Picker semantics.** Expose enough stable metadata for picker consumers to render the entry as inactive, ideally with a reason such as `no_credit`.
3. **Recovery.** Remove the inactive state automatically when authoritative quota state shows that credit has reset or refilled.
4. **Honest unknowns.** Do not mark an entry exhausted merely because quota data is absent or stale; only positive exhaustion evidence should disable it.

Example shape, if compatible with the catalog contract:

```json
{
  "disabled_reason": "no_credit"
}
```

The exact field name is an implementation decision; the required product behavior is a stable visible entry that can be rendered inactive with an exhaustion reason.

### Out of scope

- **Combo workspace controls and per-target quota badges:** tracked by child issue #1702 and its implementation work.
- **Runtime candidate failover / retry-loop prevention after a terminal candidate failure:** tracked by #1572.
- **Sticky per-target routing state after a runtime 402/429:** belongs with runtime routing/failover policy, not this catalog/picker issue.

### Relationship

- Parent UX/catalog tracker: this issue.
- Child workspace issue: #1702.
- Runtime routing/failover owner: #1572.

### Example

**Before:** `combo/grok-4.5` has known zero credit across every routable target but appears active in the picker.

**After:** the entry remains visible, is marked inactive with a no-credit reason, and becomes active again after quota recovery.

### Alternatives considered

- Manually checking the dashboard before selecting a model: works but duplicates information OpenCodex already has.
- Hiding exhausted models entirely: rejected because keeping the entry visible preserves catalog stability and explains why it is unavailable.

### Additional context

- #1702 covers the combo-workspace UI side.
- #1572 covers runtime re-evaluation/failover when a selected candidate fails terminally.
[...]

PR #4166, opened 2026-09-10
feat(logs): show an estimated decode rate alongside end-to-end throughput

### LINKED ISSUE #4038: Show estimated decode tok/s alongside end-to-end throughput in Logs
> *This was generated by AI during triage.*

## Area

Dashboard

Compare routed model generation throughput without conflating first-token latency with post-first-output delivery speed. The Logs page already records TTFT and output token counts, so users should be able to inspect an estimated decode-rate metric alongside the existing end-to-end `tok/s` value.

## What prevents this today?

The current `tok/s` metric is intentionally calculated as `outputTokens / full request duration`. A request with a large TTFT therefore appears slow even when tokens arrive quickly after the first output. TTFT is shown separately in the detail dialog, but the dashboard does not derive or display the post-TTFT rate.

## What should OpenCodex do?

Keep the existing end-to-end `tok/s` metric and its filter semantics unchanged. Add an explicitly estimated decode-rate value calculated as:

`outputTokens / (durationMs - firstOutputMs)`

Use the request-relative TTFT for the top-level row and each attempt's own attempt-relative TTFT for combo rows. Return the metric as unavailable when usage, positive output tokens, TTFT, or a positive post-TTFT duration is unavailable. Label the value as estimated because completion framing, network pauses, and stream finalization can still be included after the first output.

## Example usage or interface

For a log entry with 240 output tokens, 10,000 ms total duration, and 2,000 ms TTFT:

- Existing end-to-end rate: `24 tok/s`
- Estimated decode rate: `30 tok/s`

The Logs table and request detail should show both values without changing existing speed filters.

Users can manually export the log fields and compute the value, but this is slow and error-prone, especially for combo attempts whose TTFT is relative to the attempt rather than the parent request.

## Additional context

This is a display-time derived metric only. It should not mutate persisted request logs or usage history.

## Checks

## Proposed acceptance criteria

- Keep the current end-to-end `tok/s` value and speed-filter behavior unchanged.
- Add `displayMetrics.decodeTokPerSecond` only at management-API response time; do not change `RequestLogEntry`, `usage.jsonl`, or persisted schemas.
- Use the parent request's `firstOutputMs` for the parent metric and each combo attempt's own `firstOutputMs` for the attempt metric.
- Return `ttft_missing` when TTFT was not recorded, and `invalid_duration` when the post-TTFT interval is non-positive or invalid.
- Mark every decode-rate value as estimated, including when token usage itself was reported exactly.
- Stack end-to-end and decode values in the existing Logs rate column to avoid widening the fixed-layout table; show separate labeled values in request details and combo attempt details.
- Keep this metric Logs-only; routing analytics remain unchanged.
- Cover parent/attempt separation, unavailable cases, persistence immutability, locale parity, and unchanged speed-filter semantics with tests.

lidge-jun:
## Summary

Logs reported one rate. `tokensPerSecond()` never subtracted TTFT, `tokPerSecondResult` passed the full duration, and the `MetricSource` Pick did not even include `firstOutputMs`, so a decode-rate metric could not be computed at all — even though TTFT is already recorded per request and per combo attempt. This was never a calculation bug: the end-to-end metric is documented as end-to-end. It is a missing metric, and the two are easy to confuse when only one is shown.

`decodeTokPerSecondResult` reports output tokens over the window **after** the first token, alongside the existing rate rather than replacing it.

## Why this is not a straight re-land of #4040

Contributor PR **#4040** implemented this same metric and was closed unmerged by the reporter, on the grounds that proxy TTFT and the provider's own generation window do not line up, and that a small post-TTFT remainder makes the estimate explode. The issue stayed open, so the repository has held both an acceptance criterion and a rejection of the same feature.

Two things answer that rejection, and neither is optional.

**`MIN_DECODE_WINDOW_MS`.** Below one second the window is dominated by things that are not decoding — TTFT jitter, the proxy's own buffering, and timestamp granularity. A 240-token response whose first token arrived 50 ms before the last is not a 4800 tok/s model. A window under the floor therefore yields **no value at all**, and the test that pins that case is the one that answers the objection.

**The estimate flag.** The value is always `estimated: true`, because proxy TTFT is when the first byte reached the *proxy*, not when the provider began generating. No window length makes that exact, so the payload says so rather than implying a precision it does not have.

**Scope decision taken in this PR.** The plan asked the maintainer to choose between retrying with the guard, dropping it, and re-landing the #4040 shape confined to the detail dialog. This implements the recorded recommendation — retry with the guard — because the failure mode named in the #4040 close is bounded, and a bounded failure is a smaller change than abandoning a metric the issue still wants.

## Strictly additive

`tokensPerSecond`, `tokPerSecondResult`, `filterLogs`, `RequestLogEntry` and `usage.jsonl` are untouched. The end-to-end rate keeps its meaning and its value, the speed filter still sorts on it, `src/routing/analytics.ts` is unchanged, and this stays out of `/api/request-history` even though it shares the DTO. The parent row uses the request's own TTFT and each attempt uses its own attempt-relative one, so a combo parent never borrows a child's window. On the interface side the field is optional, because a row cached by an older build has none; absent renders nothing rather than an empty slot.

## Verification

Local checks: **NOT RUN**. The maintainer set an explicit constraint for this round that no local product suite runs — no `bun test`, `bun install`, `bun run typecheck`, `bun run build`, `bun run test:changed`, lint, or `privacy:scan`. Exact-head remote CI is the only gate.

New cases in `tests/server/management-api-logs-metrics.test.ts`, next to the existing end-to-end pin, which still asserts 240 tokens over 2000 ms is 120 and is undisturbed because the assertions there read `.tokPerSecond` rather than the whole `displayMetrics` object:

- 240 tokens, 10000 ms, TTFT 2000 gives decode 30 while the end-to-end rate stays 24, and the stored log is still not mutated.
- A long clean window is still marked estimated.
- **TTFT 9950 of 10000 ms yields `decode_window_too_short` rather than 4800 tok/s**, with the end-to-end rate unaffected. This is the #4040 case.
- A missing TTFT is `ttft_missing`, which is a different fact from a bad duration and gets its own reason.
- A TTFT at or past the total duration is `invalid_duration`, checked at both the boundary and beyond it.
- Zero output tokens is `output_missing`; unsupported usage stays `usage_unsupported`.
- A combo attempt measures its own window while the parent, having no TTFT of its own, reports `ttft_missing` rather than borrowing the attempt's.

`METRIC_REASON_KEYS` is a `satisfies Record<MetricUnavailableReason, string>`, so the two new reasons are compile-checked into the interface rather than left to drift. Both, plus the metric label, were added by hand to all nine locale catalogs with real translations.

## Checklist

- [x] Docs or release notes were updated when needed. No documented behaviour changes; the existing end-to-end metric keeps its meaning and its description.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. No auth, credential, workflow, or release-automation surface is touched. The metric is derived at response time from timings already present in the log, is never persisted, and carries no request content.

Closes #4038.

## Summary by CodeRabbit

* **New Features**
  * Added estimated decode throughput to request logs, shown alongside end-to-end throughput.
  * Added explanations when decode throughput is unavailable, including missing first-token timing or an insufficient measurement window.
  * Added localized labels and messages across supported languages.
  * Decode throughput is displayed in log tables and detail views, including individual attempts.

* **Tests**
  * Added coverage for decode-rate calculations, unavailable cases, and per-attempt measurements.


lidge-jun:
## 리뷰 · 우선순위 68 / 80

설명

이 PR은 이슈 #4038을 닫습니다. Logs가 보여 주던 속도는 처음부터 끝까지의 end-to-end 처리량이고, TTFT를 빼지 않습니다. `tokensPerSecond()`도, `tokPerSecondResult`도 전체 `durationMs`로 나누며, 예전 `MetricSource` Pick에는 `firstOutputMs`조차 없어서 디코드 구간 속도를 계산할 수 없었습니다. 계산 버그가 아니라 **없는 지표**였고, 하나만 보여 주면 둘을 헷갈리기 쉽습니다.

역사도 중요합니다. 기여자 PR #4040이 같은 지표를 넣었다가, 프록시 TTFT와 프로바이더 생성 창이 어긋나고 짧은 잔여 구간에서 숫자가 폭발한다는 이유로 머지 없이 닫혔습니다. 그런데 이슈는 열린 채로 남아서, 저장소가 "원해"와 "거절"을 동시에 들고 있었습니다. 이번 브랜치는 그 거절을 두 장치로 받습니다. `MIN_DECODE_WINDOW_MS = 1000` — 1초 미만 창은 값을 내지 않고 `decode_window_too_short`로 끝냅니다(240토큰·50ms → 4800 tok/s 같은 #4040 케이스). 그리고 값이 나와도 항상 `estimated: true`입니다. 프록시 TTFT는 첫 바이트가 프록시에 도착한 시각이지 프로바이더 생성 시작이 아니기 때문입니다.

구현은 `src/server/management/shared.ts`의 `decodeTokPerSecondResult`와 DTO의 `displayMetrics.decodeTokPerSecond` 추가뿐입니다. 기존 `tokensPerSecond` / `tokPerSecondResult` / `RequestLogEntry` / `usage.jsonl` / 속도 필터 / `src/routing/analytics.ts` / `/api/request-history`는 그대로입니다. 부모는 요청 자신의 TTFT만 쓰고, 콤보 attempt는 attempt-relative TTFT를 씁니다. 부모가 자식 창을 빌리지 않습니다. GUI `Logs.tsx` 상세에 선택 필드로 붙이고, 없는 옛 캐시 행은 빈 칸이 아니라 아예 안 그립니다. 새 reason 키는 `METRIC_REASON_KEYS`의 `satisfies Record`로 컴파일 체크되고 9개 locale에 번역이 있습니다. 테스트가 e2e 유지·추정 플래그·짧은 창 거절·ttft_missing·invalid_duration·output_missing·콤보 부모/시도 분리까지 잠급니다.

지금 `dev` Lane B 남은 줄에서 #4038은 #1711 다음입니다. 테제 파일은 #4165와 거의 안 겹치지만 **아홉 i18n 파일은 #4165·#4156과 동시 충돌**합니다. 로컬 스위트 스킵, 원격 CI만 게이트, 그리고 enforce-target이 UI screenshot missing으로 DRAFT를 유지하는 점도 #4165와 같습니다.

`gui/src/pages/Logs.tsx` - 상세 패널에만 붙는지, 목록 행/속도 정렬에는 안 넣는지 PR 범위상 의도된 축소입니다. #4040 재논의에서 "상세에만" 옵션도 있었으므로, 목록에 안 보이는 것이 제품인지 확인만 하면 됩니다.

`MIN_DECODE_WINDOW_MS` = 1000 - 바닥값 자체는 제품 감각입니다. 너무 높이면 짧은 답변이 전부 unavailable이 되고, 너무 낮으면 #4040 폭발이 돌아옵니다. 테스트가 50ms 잔여를 거절하는 쪽을 고정해 두었습니다.

`decodeTokPerSecondResult` - `usage.estimated`와 무관하게 항상 `estimated: true`입니다. usage가 정확한 카운트여도 창 경계가 추정이라서입니다. UI 카피가 "est."인 것과 맞습니다.

enforce-target - Logs UI를 바꿨는데 스크린샷이 없어 draft가 풀리지 않습니다.

메인테이너의 판단이 필요한 지점
- 1초 바닥을 제품 기본으로 받을지, 상세 전용으로만 더 낮출지
- Lane B에서 #1711(#4165)보다 뒤에 둘지, i18n만 맞추고 독립 랜딩할지
- 목록/필터에는 계속 안 넣고 상세만 둘지(PR 권장: 상세만)

너의 추천
상세 화면 스크린샷을 본문에 넣고, CI가 초록이면 머지하세요. #4040이 닫힌 이유를 테스트로 정면 반박한 형태라 재오픈 논쟁 비용이 작습니다. #4165와 같이 올릴 때는 locale 충돌만 rebase로 정리하면 되고, types/config 스플릿이나 close-don't-rebase 대상은 아닙니다. 엔드투엔드 숫자는 그대로 두라는 제약을 코드가 지키고 있습니다.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
**Screenshot gate waived by the project owner.**

This is a deliberate waiver of the screenshot requirement on a real GUI change, and it is not a claim that the gate misfired. The `gui-screenshot-waived` label is described in this repository as a waiver for false-positive screenshot requirements; that is a narrower reading than the use here, so the record should say plainly what happened.

The visual states this round adds — a decode-rate figure stacked on a live request row, a Free-only catalog filter over discovered pricing, and an inactive badge driven by real quota exhaustion — do not render from a static build. Each needs a running proxy sitting in a specific upstream state, so a screenshot here would cost a live reproduction rather than a build step. The description carries a written account of the surface instead.

The decision and its reasoning are recorded in `devlog/_plan/260910_post249_round2/120_landing_gate_decision.md`.

PR #4164, opened 2026-09-09
fix(service): recover a stale launchd job with bootout before load

lidge-jun:
## Summary

After `ocx update` the background service never came back. Update stops the service, replaces the binary, then runs `ocx service repair`, which on darwin is `installLaunchd`. That function best-effort `unload`ed the plist, ran `load -w`, and threw on any stderr matching `Load failed` or `Bootstrap failed`.

`unload` is the legacy verb and it does **not** evict a job bootstrapped into the GUI domain. That is exactly the state modern launchd reports by writing `Load failed: 5: Input/output error` to stderr **and exiting 0** — so a live-but-stale job was precisely the case that could not repair itself. The thrown text carried the `launchctl bootout` recipe as a hint, and nothing ever executed it. `startLaunchd` already handles the same stderr correctly by asking whether the live job matches the current plist, but repair does not go through `startLaunchd`.

The fix evicts with `bootout` before loading, and if `load -w` still reports the job as bootstrapped, boots out once more and retries the load a single time before keeping the existing throw.

### The decision a reviewer should be able to object to

**`bootout` kills the live gui job.** That is the repair this issue asks for, and it is also exactly why the previous code only printed the command instead of running it. This is a product decision, not a bug fix that speaks for itself.

Two things bound the blast radius, and both are load-bearing rather than reassurance:

- It runs **only inside `installLaunchd`**, which is already the "put the job back" path and which has just rewritten the plist — so whatever launchd is running is stale by construction and there is nothing to preserve. `ocx service start` is untouched: `startLaunchd` still refuses to evict, because that throw exists so `start` on a healthy service never kills it.
- It fires **only after `load -w` has already failed**. A healthy job is evicted once and immediately reloaded from the plist that was just written, and is never retried.

If you disagree with taking it, the alternative is the status quo: repair keeps printing a command the operator has to run by hand, and `ocx update` keeps leaving the service down.

### Deliberately not changed

`launchctlLoadFailed` is unchanged. That regex is the 2026-08-02 silent-success guard, and the fix is to recover from the condition rather than to stop detecting it. Relatedly, the retry is scoped to *that signal* rather than to a non-zero exit, so a malformed plist surfaces its real stderr immediately instead of being retried pointlessly.

`stopLaunchd`, `statusLaunchd` and `uninstallLaunchd` keep legacy `unload`. Once install boots out before loading, changing them is not required, and each has a test pinning its exact string.

[...]

PR #4152, opened 2026-09-09
fix(service): stop the test suite from mutating a live service manager

lidge-jun:
## Summary

Running this repository's test suite can stop the proxy the developer is running. That is a bad thing to do to anyone dogfooding opencodex, and it is the one class of test damage the existing isolation does not cover.

`tests/preload.ts` sandboxes `HOME`, `OPENCODEX_HOME` and `CODEX_HOME` on every invocation, including a bare `bun test <file>`. That covers everything addressed by a path. A service manager is not addressed by a path. `systemctl --user stop opencodex-proxy.service` addresses a job by name and talks to the user manager that is already running; `launchctl bootout gui/<uid>/com.opencodex.proxy` talks to launchd the same way. Neither consults `HOME`, so a test that falls through to either one reaches the live service no matter how well the home is sandboxed.

Windows has refused this for a while. `querySchtasks` throws on every non-query call while the test-home guard is armed, after a partially-faked service test replaced a real scheduled task with a launcher inside a temporary test home — the test passed, and cleanup deleted the launcher. macOS and Linux never got the equivalent.

`sh()` and the real `runLaunchctl` runner now refuse a mutating `launchctl` or `systemctl` invocation while the guard is armed, with a message that says what to do instead. Three properties keep it from being disruptive in the other direction:

- Read-only verbs stay allowed. `launchctl list`, `launchctl print`, `systemctl --user show`, `is-active`, `is-enabled`, `status` and `show-environment` are what the diagnostics are built on, and observation cannot take a service down.
- An injected `spawnSync` stand-in is untouched. The existing `runLaunchctl` and `startLaunchd` tests exercise output parsing through injected runners and keep working unchanged.
- It is inert in production. Arming requires `OCX_TEST_HOME_GUARD=1`, which only this repository's test preload sets, so a user running `ocx service restart` is unaffected.

`sh()` is the choke point rather than each call site, so a `systemctl` or `launchctl` call added later is covered without anyone remembering to guard it — the same fail-closed shape as the home guard itself.

## Verification

| Check | Status |
| --- | --- |
| `bun run test` (full suite) | NOT RUN — deliberately skipped; running it is the behavior this PR is about |
| `bun run typecheck` | NOT RUN |
| `bun test tests/service/live-service-manager-guard.test.ts` | NOT RUN |
| CI on this head | the authoritative evidence |

Local execution was skipped on purpose for this one. The suite is what reaches a live service manager, and the machine this was written on is running opencodex.

[...]

PR #4170, opened 2026-09-10
fix(stop): name the real refusal cause instead of asserting ownership

### LINKED ISSUE #4169: [Bug]: every stop refusal is reported as a CODEX_HOME ownership mismatch, hiding respawnable_service and looping the operator
### Client or integration

`ocx` CLI on Windows (Task Scheduler service backend). Observed on an installed **2.43.0**; the code path is unchanged on **2.49.0**.

### Area

Service lifecycle

### Summary

When the proxy refuses a stop, `stopProxy` reports **every** refusal as a `CODEX_HOME`/`OPENCODEX_HOME` ownership mismatch. The management API distinguishes three different 409 refusals, but the CLI collapses them into one message that names a cause the operator can neither confirm nor act on.

In my case the homes matched exactly, and the server's actual answer was `respawnable_service`. Because the printed message said the opposite, I spent three attempts re-exporting `CODEX_HOME` before probing `POST /api/stop` directly and seeing the real code.

The advice also forms a loop for the operator:

- `POST /api/stop` → 409 `respawnable_service`: *"the stop must be run by `ocx stop`"*
- `ocx stop` → *"a service installed under a different CODEX_HOME/OPENCODEX_HOME owns it. Run the stop from that home."*

Each command points at the other, and neither names the Task Scheduler wrapper that is actually refusing.

### Reproduction

1. On Windows, install the service backend (Task Scheduler) so the proxy is respawnable.
2. Run `ocx update` (or `ocx stop`) from the **same** `CODEX_HOME` the service was installed with.
3. Observe the printed failure.

Observed:

```
🛑 Service manager stopped.
❌ Failed to stop proxy (PID <pid>).
   The running proxy refused to stop: a service installed under a different
   CODEX_HOME/OPENCODEX_HOME owns it. Run the stop from that home.
```

The server's real answer, probed directly with the admin token:

```
$ curl -s -o body -w "HTTP %{http_code}\n" -X POST http://127.0.0.1:10100/api/stop -H "Authorization: Bearer <token>"
HTTP 409

$ cat body
{"success":false,"code":"respawnable_service","message":"This proxy is managed by a Task Scheduler wrapper that can respawn it, so the stop must be run by `ocx stop`, which verifies the respawn window. Nothing was changed."}
```

Ownership was **not** mismatched. `~/.opencodex/service-state.json`:

```json
{
  "version": 2,
  "codexHome": "C:\\Users\\<user>\\.codex",
  "opencodexHome": "C:\\Users\\<user>\\.opencodex",
  "backend": "scheduler"
}
```

…and the command ran with `CODEX_HOME=C:\Users\<user>\.codex`, i.e. equal to `state.codexHome`.

### Version

2.43.0 (observed). Re-checked against 2.49.0 and dev (2.50.0): the message strings and branching are unchanged.

### Operating system

Windows 11 (Task Scheduler service backend). Codex runtime 0.153.4.

### Expected

The refusal should name the cause the server actually reported. A `respawnable_service` or `service_state_unknown` refusal should not be described as an ownership mismatch, and the recommended next command should not be the command the operator just ran.

### Where it comes from

`src/server/management-api.ts` returns three distinct refusal codes for `POST /api/stop`:

- `respawnable_service`
- `self_unload_service`
- `service_state_unknown`

`src/lib/process-control.ts` keeps the server text in `lastRefusalMessage` and uses it when present, but the fallback hardcodes the ownership wording for all three, and the surrounding comments state ownership as fact:

```ts
// The proxy refused on purpose (foreign service owns it). Forcing would strip shared
// config while that service keeps the proxy alive.
throw new ProxyOwnershipRefusedError(
  lastRefusalMessage
  ?? "The running proxy refused to stop: a service installed under a different "
    + "CODEX_HOME/OPENCODEX_HOME owns it. Run the stop from that home.",
);
```

The same assumption appears at the 409 branch of `stopProxyGracefully` ("a service installed under another home owns it and would respawn it anyway") and in the `ProxyOwnershipRefusedError` name itself.

So there are two separable defects:

1. **The fallback asserts a specific wrong cause.** Whenever `lastRefusalMessage` is absent, the operator is told about `CODEX_HOME` regardless of which of the three refusals occurred.
2. **In my run the server message was present but not shown.** The printed text was the hardcoded fallback even though `POST /api/stop` returned a non-empty `message` moments later under the same conditions. I have not isolated why `lastRefusalMessage` was null on that path, so I am reporting (1) as the actionable defect and (2) as an observation, not a diagnosis.

### What I am not claiming

- I did not reduce (2) to a standalone reproducer, and I did not verify it on 2.49.0 — only that the message strings and branching are unchanged there.
- The eventual workaround on my side was terminating the proxy process directly and letting the scheduler respawn it on the new package, which is outside what this report is about.

### Suggested direction

Carry the server's `code` alongside the message and select the fallback wording from it, so a refusal without a body still names the right family (respawnable wrapper / self-unload service / unreadable scheduler state) rather than defaulting to ownership. Renaming `ProxyOwnershipRefusedError` to something cause-neutral would keep call sites honest.

Happy to open a PR against `dev` if this direction looks right.

### Checks

yeongjunyoo:
Fixes #4169.

## What was wrong

`POST /api/stop` refuses for three distinct reasons:

| code | meaning |
| --- | --- |
| `respawnable_service` | a service manager can respawn the proxy |
| `self_unload_service` | the proxy **is** the installed service (#4023) |
| `service_state_unknown` | the scheduler state could not be read |

`stopProxy` preferred the server's own message, but when no readable message arrived it fell back to a single hardcoded sentence naming a fourth cause that the server never reported:

```ts
throw new ProxyOwnershipRefusedError(
  lastRefusalMessage
  ?? "The running proxy refused to stop: a service installed under a different "
    + "CODEX_HOME/OPENCODEX_HOME owns it. Run the stop from that home.",
);
```

#4023 already carried the message through precisely because there is more than one cause. The fallback was the half of that fix still guessing, and the existing test says so in its own comment — *"stopProxy used to report the first of those unconditionally, sending an operator whose proxy is simply the service to a CODEX_HOME that does not exist."*

The cost is not cosmetic. On Windows with the Task Scheduler backend, the two commands point at each other:

- `POST /api/stop` → 409 `respawnable_service`: *"the stop must be run by `ocx stop`"*
- `ocx stop` → *"a service installed under a different CODEX_HOME/OPENCODEX_HOME owns it. Run the stop from that home."*

I hit this with matching homes (`service-state.json.codexHome` equal to the invoking `CODEX_HOME`) and spent three attempts re-exporting `CODEX_HOME` before probing the endpoint directly and seeing `respawnable_service`. The printed advice names a home that is already correct, and neither message mentions the wrapper that is actually refusing.

## What this changes

- Capture the refusal `code` next to the message (`lastStopRefusalCode()`), so a refusal with an empty or unparseable body still knows its cause.
- Select the fallback wording from that code. The default is **cause-neutral** — naming no cause is better than naming the wrong one, because the operator acts on what it says.
- Drop "foreign" from the CLI's teardown notice. A respawning wrapper, or the service itself, is not another home's proxy.

No behavioral change when the server sends a message: that path is unchanged and still wins.

## Tests

Added to `tests/lib/process-control-graceful.test.ts`:

- the refusal code is captured alongside the message, and is `null` when the body cannot be parsed (no JSON means no code to recover — the fallback then has to be cause-neutral, not code-selected)
- a refusal **without a message** falls back by code (`respawnable_service` / `self_unload_service` / `service_state_unknown`), and a refusal with neither falls back to cause-neutral wording. **None of the four may contain `CODEX_HOME` or `OPENCODEX_HOME`** — the regression this PR is named after
- concurrent refusals each keep their own cause (see below)

`tests/providers/xai/grok-lifecycle.test.ts` asserted the old "foreign proxy" wording and was updated with it. The source-oracle assertions in that file (409 does not escalate to a forced kill; `graceful === "refused"` precedes `killProxy`) still hold and pass unchanged.

Verified locally on Windows:

- `bun test tests/lib/ tests/providers/xai/grok-lifecycle.test.ts tests/cli/uninstall.test.ts tests/service/service.test.ts` → 579 pass, 0 fail
- `bun run typecheck` → clean
- `bun run privacy:scan` → passed

## Per-call refusal, and a counterfactual for it

Review follow-up: `stopProxyGracefully` published the 409 message and code to module-scoped state that `stopProxy` read after awaiting it, so two overlapping stops could interleave and the first would throw with the second's cause. That applied to the pre-existing `lastRefusalMessage` as well, not only the code this PR adds.

`stopProxyGracefullyDetailed` now returns `{ result, refusal }` and `stopProxy` builds the error from that value. The exported `stopProxyGracefully` signature and `GracefulStopResult` are unchanged (it is a thin wrapper), and `lastStopRefusalMessage()` / `lastStopRefusalCode()` remain as observational accessors.

The first version of the overlap test did not actually guard this. It parked one call inside `fetch` and let the other finish completely before resuming — a schedule the global handoff also passes, because the parked call republishes its own globals last. **I checked that against the pre-fix commit and it passed there, so it was proving nothing.**

The test now starts both stops together. Counterfactual, run against the pre-fix implementation (506473d):

```
error: expect(received).toContain(expected)
Expected to contain: "respawn"
Received: "The running proxy refused to stop: it is the installed service itself, ..."
(fail) concurrent stops keep their own cause
```

The `respawnable_service` call throws the `self_unload_service` cause. It passes on the per-call result.

## What I did not do

- I did not rename `ProxyOwnershipRefusedError`. The name now overstates what it carries, but it is used across `src/cli/index.ts` and renaming it is a separate, wider change — happy to follow up if you want it.
- Issue #4169 also notes that in my run the server's message was present when probed directly yet the hardcoded fallback was printed. I could not isolate why `lastRefusalMessage` was null on that path, so this PR does not claim to fix it. It does make that class of failure non-misleading: whatever the reason a message goes missing, the operator is no longer sent to the wrong home.

## Review readiness checklist

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -952,7 +952,7 @@ async function handleStop() {
       if (detail) console.error(`   ${detail}`);
       if (err instanceof ProxyOwnershipRefusedError) {
         ownershipBlocked = true;
-        console.error("   Skipping shared teardown (native Codex restore, Grok config): the foreign proxy is still running.");
+        console.error("   Skipping shared teardown (native Codex restore, Grok config): the refusing proxy is still running.");
       }
     }
   } else {
@@ -979,7 +979,7 @@ async function handleStop() {
         if (detail) console.error(`   ${detail}`);
         if (err instanceof ProxyOwnershipRefusedError) {
           ownershipBlocked = true;
-          console.error("   Skipping shared teardown (native Codex restore, Grok config): the foreign proxy is still running.");
+          console.error("   Skipping shared teardown (native Codex restore, Grok config): the refusing proxy is still running.");
         }
       }
     } else if (live) {
--- a/src/lib/process-control.ts
+++ b/src/lib/process-control.ts
@@ -142,9 +142,28 @@ export class ProxyOwnershipRefusedError extends Error {}
  * attest the process exit code or completion of every drain/shutdown hook.
  */
 export async function stopProxyGracefully(pid: number, io: GracefulStopIo = {}): Promise<GracefulStopResult> {
+  return (await stopProxyGracefullyDetailed(pid, io)).result;
+}
+
+/**
+ * The refusal a single stop attempt received, carried back to that attempt's caller.
+ *
+ * Module-scoped state cannot do this job: two overlapping stops race, and the first would
+ * report the second's cause. The exported accessors stay as observational state for callers
+ * that only want the last refusal, but the error text is built from this per-call value.
+ */
+type StopRefusal = { message: string | null; code: string | null };
+
+async function stopProxyGracefullyDetailed(
+  pid: number,
+  io: GracefulStopIo = {},
+): Promise<{ result: GracefulStopResult; refusal: StopRefusal }> {
+  const refusal: StopRefusal = { message: null, code: null };
+  const done = (result: GracefulStopResult): { result: GracefulStopResult; refusal: StopRefusal } =>
+    ({ result, refusal });
   const readRuntime = io.readRuntime ?? readRuntimePort;
   const runtime = io.runtimeEndpoint ?? readRuntime(pid);
-  if (!runtime?.port) return false;
+  if (!runtime?.port) return done(false);
   const env = io.env ?? process.env;
   const headers: Record<string, string> = {};
   const token = configuredAdminToken(env.OPENCODEX_HOME?.trim() || undefined, env as NodeJS.ProcessEnv);
@@ -173,7 +192,7 @@ export async function stopProxyGracefully(pid: number, io: GracefulStopIo = {}):
     // here would run the daemon's cleanup and strip shared config out from under the
     // still-running service. Report the refusal instead of forcing.
     if (res.status === 409) {
-      const refusal = await res.json()
+      const parsed = await res.json()
         .then(body => {
           const record = body as { message?: unknown; code?: unknown } | null;
           const message = record?.message;
@@ -184,11 +203,13 @@ export async function stopProxyGracefully(pid: number, io: GracefulStopIo = {}):
           };
         })
         .catch(() => ({ message: null, code: null }));
-      lastRefusalMessage = refusal.message;
-      lastRefusalCode = refusal.code;
-      return "refused";
+      refusal.message = parsed.message;
+      refusal.code = parsed.code;
+      lastRefusalMessage = parsed.message;
+      lastRefusalCode = parsed.code;
+      return done("refused");
     }
-    if (!res.ok) return false;
+    if (!res.ok) return done(false);
     const body: unknown = await res.json().catch(() => null);
     const expectedTeardown = io.deferSharedTeardownNonce ? "deferred" : "performed";
     sharedTeardownConfirmed = body !== null
@@ -197,14 +218,14 @@ export async function stopProxyGracefully(pid: number, io: GracefulStopIo = {}):
       && "success" in body && body.success === true
       && "sharedTeardown" in body && body.sharedTeardown === expectedTeardown;
   } catch {
-    return false;
+    return done(false);
   }
   const waitExit = io.waitExit ?? waitForExit;
   // Honor the server's own drain window: /api/stop answers 200 first, then drains for
   // config.shutdownTimeoutMs. Waiting less than that hard-kills mid-drain.
   const exitTimeoutMs = io.exitTimeoutMs ?? drainDeadlineMs();
-  if (!waitExit(pid, exitTimeoutMs)) return false;
-  return sharedTeardownConfirmed ? true : "teardown-unconfirmed";
+  if (!waitExit(pid, exitTimeoutMs)) return done(false);
+  return done(sharedTeardownConfirmed ? true : "teardown-unconfirmed");
 }
 
 function drainDeadlineMs(): number {
@@ -219,13 +240,14 @@ function drainDeadlineMs(): number {
 export async function stopProxy(pid: number, io: GracefulStopIo = {}): Promise<boolean> {
   if (!isProcessAlive(pid)) return false;
   const runtime = io.runtimeEndpoint ?? readRuntimePort(pid);
-  const graceful = await stopProxyGracefully(pid, io);
+  const { result: graceful, refusal } = await stopProxyGracefullyDetailed(pid, io);
   if (graceful === "refused") {
     // The proxy refused on purpose. Forcing would strip shared config while whatever owns
     // the process keeps it alive. The server's own message is preferred; the fallback is
-    // selected from its code so an empty body still names the right cause.
+    // selected from its code so an empty body still names the right cause. Both come from
+    // THIS attempt, so an overlapping stop cannot lend it the wrong reason.
     throw new ProxyOwnershipRefusedError(
-      lastRefusalMessage ?? refusalFallbackMessage(lastRefusalCode),
+      refusal.message ?? refusalFallbackMessage(refusal.code),
     );
   }
   if (graceful === "teardown-unconfirmed") {
--- a/src/lib/process-control.ts
+++ b/src/lib/process-control.ts
@@ -80,9 +80,47 @@ export type GracefulStopResult = boolean | "refused" | "teardown-unconfirmed";
  */
 let lastRefusalMessage: string | null = null;
 
+/**
+ * The server's machine-readable reason for the most recent 409, captured alongside the
+ * message so a refusal that arrives without a body still names the right cause. Without it
+ * the fallback has to guess, and guessing "ownership" sent operators to re-check
+ * CODEX_HOME for a refusal the scheduler wrapper had issued (#4169).
+ */
+let lastRefusalCode: string | null = null;
+
 /** The server's explanation for the most recent 409, or `null` when it sent none. */
 export function lastStopRefusalMessage(): string | null {
   return lastRefusalMessage;
+}
+
+/** The server's `code` for the most recent 409, or `null` when it sent none. */
+export function lastStopRefusalCode(): string | null {
+  return lastRefusalCode;
+}
+
+/**
+ * Wording for a refusal whose body carried no message. Each branch mirrors a refusal the
+ * management API can return from `POST /api/stop`; the default stays cause-neutral because
+ * naming the wrong cause is worse than naming none — it costs the operator the time they
+ * spend acting on it.
+ */
+function refusalFallbackMessage(code: string | null): string {
+  switch (code) {
+    case "respawnable_service":
+      return "The running proxy refused to stop: a service manager that can respawn it owns "
+        + "the process. Run `ocx stop`, which verifies the respawn window.";
+    case "self_unload_service":
+      return "The running proxy refused to stop: it is the installed service itself, so "
+        + "stopping the manager from inside it would end the process before native Codex is "
+        + "restored. Run `ocx stop`.";
+    case "service_state_unknown":
+      return "The running proxy refused to stop: the service manager state could not be read, "
+        + "so it cannot tell whether a wrapper would respawn it. Run `ocx service status` to "
+        + "see the query error.";
+    default:
+      return "The running proxy refused to stop and sent no reason. Run `ocx service status` "
+        + "to inspect the service state.";
+  }
 }
 
 /**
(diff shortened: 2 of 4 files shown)


lidge-jun:
## 리뷰 · 우선순위 61 / 80

이 PR은 Windows에서 `ocx stop` / `ocx update`가 프록시를 끌 때 나오는 거절 메시지를 고칩니다. 지금 `dev`의 `src/server/management-api.ts` `POST /api/stop`은 409를 세 가지로 나눕니다. `respawnable_service`(리스폰 가능한 스케줄러 래퍼), `self_unload_service`(프록시 자신이 설치된 서비스), `service_state_unknown`(스케줄러 상태를 못 읽음). 그런데 클라이언트의 `src/lib/process-control.ts` `stopProxy`는 서버 메시지가 비면, 네 번째 원인인 “다른 CODEX_HOME/OPENCODEX_HOME이 소유한다” 문장 하나로 전부 덮어씁니다. #4023이 서버 메시지를 그대로 옮기게 한 뒤에도, 메시지가 비는 반쪽은 그대로 추측이었습니다.

실제 피해는 말장난이 아닙니다. #4169처럼 집이 이미 맞는데도, API는 `respawnable_service`인데 CLI는 소유권 불일치를 말합니다. 그러면 운영자는 CODEX_HOME을 다시 맞추고, `/api/stop`은 “`ocx stop`으로 하라”고 하고, `ocx stop`은 “다른 집에서 하라”고 해서 서로 가리키는 루프가 됩니다. 최근 `dev`에는 #4152 서비스 매니저 가드와 #4164 launchctl bootout 같은 Lane A 서비스 작업이 이미 올라와 있어서, 정지 경로의 메시지를 정확히 하는 일은 그 레인과 바로 맞닿아 있습니다. 이 패치는 거절 `code`를 메시지 옆에 저장하고(`lastStopRefusalCode`), 본문이 비었을 때만 code별 대체 문장을 고르며, 기본값은 원인을 단정하지 않습니다. CLI 안내의 “foreign proxy”도 “refusing proxy”로 바꿉니다. 서버가 메시지를 보낸 정상 경로는 그대로입니다.

테스트는 `tests/lib/process-control-graceful.test.ts`에 code 캡처와 bodyless fallback(네 경우 모두 CODEX_HOME/OPENCODEX_HOME 금지)을 넣고, `tests/providers/xai/grok-lifecycle.test.ts`의 문구 오라클만 맞춰 두었습니다. 범위가 `process-control` + CLI 문구 + 테스트로 좁고, types/config 분할 캠페인과도 무관합니다.

라인 문제:

process-control.ts · refusalFallbackMessage(respawnable_service) - 대체 문장이 또 “Run `ocx stop`”을 말합니다. 신고 경로는 이미 `ocx stop` 안인데 메시지가 비어 fallback이 뜬 경우라, 같은 명령을 다시 가리키면 루프가 남습니다. 서버 본문이 있을 때도 같은 조언이지만, CLI 전용 fallback은 “이미 ocx stop 중이면 service status / deferral receipt를 보라”처럼 한 단계 더 구체적이면 좋습니다.

management-api.ts 약 331–335 · isServiceOwnershipError 409 - 여기 응답은 message만 있고 code가 없습니다. 메시지가 또 빠지면 새 default(“sent no reason”)로 가는데, 이 경우만은 진짜 소유권 거절이라 code를 하나 붙이는 follow-up이 맞습니다(예: foreign_home 또는 ownership). failed 서비스 정지 409(약 343행)도 code가 없습니다.

process-control.ts · lastRefusalCode/Message - 새 stop 시도 시작 때 null로 비우지 않습니다. 이번 throw 경로에서는 409 직후 값이 쓰이니 당장 깨지진 않지만, 내보낸 `lastStopRefusalCode()`를 다른 호출자가 읽으면 이전 거절이 남을 수 있습니다. stopProxyGracefully 진입 시 둘 다 초기화하는 편이 안전합니다.

#4169 본문 · lastRefusalMessage가 null이 된 근본 원인 - 작성자도 직접 probe에선 message가 있었는데 CLI에선 fallback이 찍힌 이유를 못 잡았다고 적었습니다. 이번 PR은 “틀리게 단정하지 않기”까지이고, 왜 parse/전달이 비었는지는 열려 있습니다. 재현되면 fetch/json/admin 토큰/호스트 경로를 따로 이슈로 남기는 게 좋습니다.

ProxyOwnershipRefusedError 이름 - 작성자 말대로 이제 소유권만이 아닙니다. 이번 범위 밖으로 두는 선택은 이해되지만, 메인테이너가 이름 변경 follow-up을 받을지 정하면 됩니다.

메인테이너의 판단이 필요한 지점
- draft 체크리스트가 0/4입니다. 로컬 검증(bun test/typecheck/privacy)은 본문에 있으니, 박스만 채우면 review-ready로 올릴지 지금 merge 검토할지.
- ownership 409에 code를 이 PR에 같이 넣을지, 후속으로 뺄지.
- respawnable fallback 문장을 CLI 맥락용으로 더 고칠지, 서버 메시지와 대칭을 유지할지.
- ProxyOwnershipRefusedError 이름 변경을 받을지(별 PR 권장).

너의 추천
체크리스트를 채운 뒤 merge 후보로 두면 됩니다. 메시지 오안내를 끊는 좁은 수정이고 테스트가 회귀를 잠급니다. merge 전에 respawnable fallback의 “다시 ocx stop” 문장만 한 줄 다듬거나, 최소한 메인테이너 코멘트로 “이미 stop 경로면 status를 보라”를 남기세요. 소유권 409 code와 Error 이름 변경은 후속 PR로 분리하는 편이 안전합니다. types/config 분할과 무관하니 close-don't-rebase 대상이 아닙니다.

이 댓글은 grok-bot이 작성했습니다

yeongjunyoo:
Thanks — the module-scoped state point is right, and it applied to the pre-existing `lastRefusalMessage` too, not just the code I added.

Pushed b7d744f:

- `stopProxyGracefullyDetailed` returns `{ result, refusal }` per call; `stopProxy` builds the error from that value, so an overlapping stop can no longer lend it the wrong cause.
- `stopProxyGracefully` and `GracefulStopResult` keep their signatures — it is now a thin wrapper — and `lastStopRefusalMessage`/`lastStopRefusalCode` remain as observational accessors.
- Added the interleaved test you asked for: one stop parks inside `fetch` until a second has resolved and written its own code, then each is asserted to keep its own cause.

One thing worth flagging: the 409 source-oracle in `tests/providers/xai/grok-lifecycle.test.ts` asserted the literal `return "refused"`, which a wrapped return breaks. I loosened it to match `return done("refused")` as well, since the invariant it guards is that the 409 branch yields "refused" before the `!res.ok` fallthrough — ordering, not spelling. Happy to restructure instead if you would rather keep that assertion literal.

Local: `bun test tests/lib/ tests/providers/xai/grok-lifecycle.test.ts tests/cli/uninstall.test.ts tests/service/service.test.ts` → 579 pass / 0 fail; `bun run typecheck` clean; `privacy:scan` passed.

Ingwannu (commented):
The per-call refusal result is the right direction: stopProxy now reads its own result rather than the observational module globals, and 409 still returns before wait/kill escalation. Code-specific and neutral fallback wording no longer invents another home as the cause. The source-oracle's wrapped-return allowance retains the refusal-before-fallthrough ordering check.

One test limitation remains: the overlap case parks the first call inside fetch, lets the second stop finish completely, then resumes the first. A module-global implementation can also pass that schedule because the first response writes its own globals after the second is done. Please add a counterfactual run showing the test fails with the old global-result handoff, arranging overlap between a refusal being published and that call's stopProxy continuation consuming it. Keep server-message precedence and unknown/unparseable-body controls.

Also update the description to include the per-call change and avoid saying an unparseable/bodyless response retains its code—the parser cannot recover a code from no JSON. This does not yet prove the original missing-message incident or the Windows stop loop resolved. Keep Draft pending exact-head full CI and Jun's final wording review; no local stop/service operation was executed.

yeongjunyoo:
You were right that the overlap test proved nothing, and I confirmed it rather than assuming: I checked out the pre-fix `process-control.ts` (506473d) and **the old test passed there**. Parking one call inside `fetch` and letting the other finish first is a schedule the global handoff survives, because the parked call republishes its own globals last.

Pushed 4d72ef0 with a schedule that does discriminate — both stops started together. Counterfactual against 506473d:

```
error: expect(received).toContain(expected)
Expected to contain: "respawn"
Received: "The running proxy refused to stop: it is the installed service itself, ..."
(fail) concurrent stops keep their own cause
```

The `respawnable_service` call throws the `self_unload_service` cause, which is exactly the publish/consume overlap you described. It passes on the per-call result. Looped 20× since the interleaving is scheduler-dependent (the pre-fix code fails on the first iteration). Server-message precedence and the unknown/unparseable-body controls are untouched.

Also fixed the wording you flagged: a response with no readable body has no code either, so the description and the test name no longer suggest a code survives an unparseable body — that case falls back to cause-neutral text, not to a code.

The PR description now covers the per-call change and the counterfactual. Leaving it in Draft for exact-head CI and Jun's wording review as you asked.

On the two open questions from #4169: I did **not** add `code` to the ownership 409 in `management-api.ts`, and did **not** rename `ProxyOwnershipRefusedError` — both look like maintainer calls on scope, and I would rather not widen a fix PR into them unasked. Happy to do either here or as a follow-up.

Local: `bun test tests/lib/ tests/providers/xai/grok-lifecycle.test.ts` → 334 pass / 0 fail; `typecheck` clean.

Ingwannu (commented):
The 4d72ef010363b80cd78f65148a5228d6797a3117 follow-up addresses my earlier test objection: it now starts both stops together, checks each resulting cause, and records a pre-fix failure instead of relying on a schedule the global handoff also passed. Server-message precedence and unknown-body controls remain intact. I inspected the delta; I did not independently execute the reported 20-iteration counterfactual.

The description still has one stale sentence under “What this changes”: “an empty or unparseable body still knows its cause.” Your newer test section correctly says the opposite. Please remove that sentence's contradiction before final review. Keeping the original missing-message/root-cause investigation separate is appropriate. Do not add unrelated scheduler behavior or rename the shared error type just to widen this fix. Draft/readiness and exact-head CI are still pending.

yeongjunyoo:
Field note, since @Ingwannu asked whether the Windows stop loop is resolved — it is not, and I hit a worse form of it today: a routine `ocx service install` was rolled back by this same refusal and left the machine on `service: absent` + `routing=native` with a live proxy. Details and logs in #4169 (comment).

To be explicit about scope: **this PR does not fix that.** It only stops the refusal from naming a cause the server never reported. Whether `install` should tolerate a stop refusal the way `uninstall` already does — the two are asymmetric today — is a separate maintainer call, and I did not widen the PR into it.

No code change here; the branch is unchanged since 4d72ef0.

lidge-jun:
Closing as landed: naming the real refusal cause instead of asserting ownership is on dev via commit 1ada8f5ff on dev (1ada8f5ff), verified as an ancestor of origin/dev at d865aacf93.

This one needed no carry at all. Your change is already on dev with a Co-authored-by trailer naming you and a Closes #4169 link, and dev is a superset of the source head: it adds refusalNextStep and removes "Run ocx stop" from the fallback text, which is what actually breaks the loop #4169 described.

Your authorship is preserved by a Co-authored-by trailer in the landed commit itself, not just in the pull request body, so it counts on your contributor graph.

If you think something from this branch did not make it to dev, say so and I will reopen.

Issue #4236, opened 2026-09-10
macOS: `ocx service repair` unconditionally evicts the running LaunchAgent, accepts a no-op `launchctl load` as success, and leaves the hub down with no log line


lidge-jun:
**Severity:** high — a single `ocx service repair` can take a hub fully offline (public proxy + management ingress + loopback client listener) with no error shown, no rollback, and no automatic recovery. Recovery needs a `launchctl` command the CLI never prints.

**Environment**
- opencodex `2.51.0` (source checkout run via bundled Bun 1.4.2)
- macOS 26 / Darwin 27.0.0, arm64
- `runtimeRole: "hub"`, `hostname: "100.76.170.81"` (Tailscale-only public bind), `port: 10100`
- `hub.managementIngress: { enabled: true, port: 10102 }`
- `unauthenticatedLoopbackListener: { enabled: true, port: 10104 }`
- LaunchAgent `com.opencodex.proxy` at `~/Library/LaunchAgents/com.opencodex.proxy.plist`, `KeepAlive: true`

All line references are to the source checkout at commit `6d3ad12e3` (main, after PR #4196).

---

## Summary

Four defects, all on the macOS launchd path. (1) is the outage. (2) can send an operator into (1) on a false premise. (3) and (4) turn permanently-fatal config states into silent `KeepAlive` crash loops / false warnings.

1. **`installLaunchd()` evicts the live job with a domain-explicit `bootout`, re-registers with the domain-implicit legacy `launchctl load -w`, and accepts exit-0-with-empty-stderr as success** — then writes install state. No pre-check that the job is already healthy, no settle delay, no `bootstrap` fallback, no plist backup, no rollback. macOS is the only backend with "evict first, verify never"; the Windows branch of the same function has a full preserve/restore protocol.
2. **`diagnoseService()` derives "loaded" from `sh("launchctl list | grep <label> || true")`** — session-relative domain, and `|| true` plus a `catch` swallow every exit code, so every failure mode collapses into "not loaded". That single bit also drives `enabled`, `running`, `viable` and therefore `isServiceViable()`, which the update path uses to decide whether to start a competing proxy.
3. **An occupied `unauthenticatedLoopbackListener.port` / `hub.managementIngress.port` is a fatal startup crash that is then misdiagnosed as a *public* port conflict**, so the retry waits on the wrong port and the job `KeepAlive`-loops with `stdio: "ignore"` upstream.
[...]

PR #4240, opened 2026-09-10
fix(client): refuse a hub catalog the local Codex CLI cannot parse

lidge-jun:
## Summary

A remote client reported `connected`, catalog present and freshly synced, token owned, rotation clean — and its installed Codex CLI still exited before making a single request:

```
failed to parse model_catalog_json <CODEX_HOME>/opencodex-catalog.json as JSON:
unknown variant `max`, expected one of `none`, `minimal`, `low`, `medium`, `high`, `xhigh`
```

The connection state answered a different question from the one the operator was asking. It proved the hub was reachable and the credential worked; nothing on that path proved the **selected local runtime** could consume what was written. `validateRemoteCatalog` checks JSON shape only, and connect writes the hub's bytes verbatim, so the effort clamp that already exists for local catalog sync never saw the downloaded file. That is why `ocx status` could report an active effort clamp for the older runtime while the materialized client catalog still failed to parse — the two paths never met.

The client path now establishes compatibility before it materializes anything, and **fails closed**, which is the decision recorded for this issue:

- `catalogEffortCompatibility` reports which reasoning levels the observed local ladder would reject, **without mutating the catalog**. The clamp beside it is mutate-and-continue, which is right when this process owns the file it is about to write and wrong for a catalog that has to keep agreeing with hub truth.
- Both hub-download writes — `connectClient` and `syncConnectedClient` — are gated on it. The download is refused *before* the write, so the previous known-good catalog is still on disk, no connection is committed, and no success is printed. Refusing before the write is stronger than writing and restoring: there is no window in which an unparseable catalog exists.
- The two restore paths stay ungated. Refusing to restore a catalog this machine already accepted would strand the client with no catalog at all.

The refusal names the level, what was preserved, and both ways out — upgrade the Codex CLI, or point `CODEX_CLI_PATH` at a build that supports it and run `ocx sync`. It never suggests editing the hub catalog and it does not touch running Codex processes; the issue rules both out.

**One decision the issue left open.** An unobservable local ladder is reported as *unverified* rather than *incompatible*, and does not block. A client machine may legitimately have no Codex CLI to observe, and refusing on absent evidence would break a working configuration. I read *"preserve the prior known-good catalog if compatibility cannot be established"* as the incompatible case — the alternative to the compatible-projection branch offered in the same sentence — not as the inconclusive one. Flagging it because the other reading is defensible and would be a one-line change.

## Verification

[...]

PR #4249, opened 2026-09-11
fix(service): macOS repair no longer evicts a healthy launchd job, and verifies the one it loads

lidge-jun:
> **Stack** (hub single-port, #4236): 1 #4249 → 2 #4250 → 3 #4251 → 4 #4252 → 5 docs (next). Each PR targets the previous branch; retarget to `dev` as the one below lands. Local suite deliberately not run (operator instruction); hosted CI on the pushed head is the proof.

## Summary

`ocx service repair` on macOS IS `installLaunchd`, and it was the only backend that evicted first and verified never. It booted the live job out of `gui/$uid`, re-registered with the domain-**implicit** legacy `launchctl load -w`, and accepted exit-0-with-empty-stderr as proof. On a hub that takes the public proxy, the management ingress and the unauthenticated loopback listener down in one command, with no error, no rollback, and nothing appended to the service log — because the job never started again. Recovery needed exactly one command the CLI never printed: `launchctl bootstrap gui/$(id -u) ~/Library/LaunchAgents/com.opencodex.proxy.plist`.

The second half is what sends an operator there on a false premise. `diagnoseService()` derived "loaded" from `sh("launchctl list | grep <label> || true")`: the **caller's** bootstrap domain rather than `gui/$uid`, every exit code swallowed twice (`|| true` plus a `catch`), the label matched unanchored anywhere on a line, and no "unknown" state at all. A serving gui-domain job queried from a non-Aqua session read as `installed, not loaded` with `re-run 'ocx service repair'` attached — the command that causes the outage. That one bit also fed `enabled`, `running`, `viable` and therefore `isServiceViable()`, which `src/update/index.ts` and `src/update/job.ts` use to decide whether to start a competing proxy on the service's own port.

**`installLaunchd` now has a protocol instead of a sequence.**

- It asks launchd what it is running **before touching anything**, through the tri-state `probeLaunchdLoadState` — never the two-state `launchdJobMatchesPlist`, which reports `loaded: false` for *every* non-zero `launchctl print` (EPERM from a non-Aqua ssh/cron context, an unspawnable launchctl, an undocumented status). That distinction is the whole point: with the two-state answer a healthy serving hub failed the pre-check (evict), failed the verification the same way (evict again, roll back) and was finally told "IS NOT RUNNING" while it was up. An `unknown` probe is not evidence, so it refuses: nothing is written, no verb is run, and the error says the job may be RUNNING and names `launchctl print gui/$uid/<label>` and `launchctl print user/$uid/<label>`.
- A `loaded-current` job, an unchanged data-token file and an identical plist mean there is nothing to repair: it re-asserts 0600, refreshes install state, logs `service is already loaded from the current plist; nothing to do.` and returns without a single launchctl call. A repair of a healthy service must not be an outage. **"Identical" has to tolerate `PATH`**: `buildPlist` bakes `process.env.PATH`, and repair runs from whatever has a shell — a tray helper, `ocx update`'s child, ssh — so a whole-file comparison missed almost every time it mattered, and the healthy hub was evicted *and* had its PATH narrowed to the caller's. When PATH is the ONLY difference and the live job runs the exec line this install baked, the previous definition's PATH is put back (`reusePreviousPlistPathVariable`), the files then compare equal on their own terms, and the PATH the service already runs with survives. Anything else differing is a real rewrite, PATH included — that residual is deliberate, since PATH must stay updatable and the eviction was going to happen anyway.
- The previous bytes are kept in memory and at `<plist>.prev` before the overwrite. Not `.prev.plist` — launchd globs `~/Library/LaunchAgents/*.plist` at login, so that spelling would be a second registration of the same Label fighting the real one for the port.
- The reload is `bootstrap gui/$uid <plist>`, the verb that pairs with the bootout target, after a bounded 5 × 200 ms settle while `launchctl print` still answers 0. `bootout` is asynchronous, which is why the old back-to-back retry raced the same exiting job twice and added nothing. A `bootout` that exited 3 is not settled — nothing is exiting to wait for.
- **Every domain the probe can answer for is evicted, not just `gui/`.** The probe asks `gui/$uid` *and* `user/$uid`; the mutating verbs asked `gui/$uid` alone. Against a `user/`-domain registration of our Label that made `ocx service stop` a silent no-op (exit 3 in a domain that never held it), made the install cleanup lay new assets over a live manager, and left `installLaunchd` bootstrapping a **second** registration into `gui/` — two `KeepAlive` jobs fighting for one port. `launchdEvictionTargets()` is now the shared list for `installLaunchd`, `stopLaunchd` and the install-cleanup `stop`; `bootout` against a label a domain does not hold exits 3 and changes nothing, so both are addressed unconditionally rather than enumerated first.
- Success is the probe answering `loaded-current`, never a stderr regex, and `writeServiceInstallState` runs only after it does. `<plist>.prev` is deleted at that point instead of surviving until the next `uninstall`. It compares the command **this install baked**, not `expectedLaunchdCommand(installedServiceListenPort())`: a fresh install has no state yet, and a lost state file makes that helper fall back to the Bun + CLI pair, which would call a correctly loaded launcher job stale (#3464) and turn every first install into a rollback.
- One retry, routed by the failure. Measured on this host (macOS 26 / Darwin 27.0.0, arm64) with a throwaway `com.opencodex.test-probe` label, `Bootstrap failed: 5: Input/output error` has **two** causes: something is still bootstrapped under the label, or the label sits in the domain's disabled list. The `-w` in `load -w` was silently doing the second job, so dropping it without `enable` would have made a disabled job permanently unrepairable. The retry tries `kickstart -k` (live job) then `enable` + bootstrap (disabled list) — and `enable` runs only there, so an ordinary repair does not quietly undo a deliberate `launchctl disable`. Exit 0 with a disagreeing probe is the silent no-op and earns one more eviction. Anything else (malformed plist, EPERM) throws immediately so the real stderr reaches the operator undelayed. An `unknown` probe is never retried: a retry is another eviction.
- **`kickstart -k` is trusted only for bytes already on disk.** launchd restarts the definition it has **cached**; it does not re-read the plist. When only `EnvironmentVariables` changed the exec line is unchanged, so the verification would agree, install state would be written and repair would report success while launchd kept the OLD environment. New bytes (a fresh install included) therefore skip `kickstart` and go to `enable` + evict/bootstrap, the only way to hand launchd a new definition.
- On terminal failure it restores the previous bytes, tries to bootstrap them back, and throws an error that says what the probe actually found — `not-loaded`: evicted and not running (or restored and re-bootstrapped); `loaded-stale`: still loaded, from a different command than the plist just written, because telling that operator "nothing is listening" sends them to fix the wrong thing. Both name `launchctl bootstrap gui/$uid <plist>` as the remedy and point at `launchctl print` and `launchctl print-disabled`. When the probe answers `unknown` after the attempts it neither evicts again nor rolls back (a rollback is another eviction): an accepted bootstrap warns and records state, a refused one throws with the real stderr and an explicit "this says nothing about whether it is running".
- `stableLauncherEntry()` prefers the launcher the install already recorded while it is still an absolute executable file. A repair run from a context without `ocx` on `PATH` — a tray helper, `ocx update`'s child, ssh — used to rewrite a working launcher-form plist into the version-pinned Bun + CLI pair and then evict the healthy job to load it. **This reaches Linux too**: `installSystemd` resolves the same function, so a PATH-less repair now keeps the `ExecStart` the unit already has. The consequence there is milder — systemd reloads and restarts, it never evicts into nothing — but the silent rewrite was identical, so the behaviour is shared deliberately rather than branched, and `tests/service/service.test.ts` covers the systemd side directly.

**`ocx service restart` now restarts, instead of inheriting the no-op.** `restart` mapped to the repair path, so the fix above turned it into a lie on macOS: a restart of a healthy service reloaded nothing, and the operator documentation had to tell people to run `launchctl kickstart -k gui/$(id -u)/com.opencodex.proxy` by hand. `installLaunchd` now returns `LaunchdInstallOutcome { reloaded }` — `false` only on the no-op path, which is the only path where launchd was never asked for anything — and `repairService` takes the verb. On darwin, `restart` plus `reloaded: false` runs `kickstart -k gui/<uid>/<label>`: it restarts what the domain already holds, so there is no eviction window, and it cannot publish bytes, which is irrelevant when there are none to publish. Success is the same tri-state probe, against the exec line an install would bake; it logs one line (`service restarted (launchctl kickstart -k gui/<uid>/com.opencodex.proxy).`), warns on `unknown` because that is not evidence, and throws on absence or a refused kickstart — after which the repair branch still runs its `reportServiceServing` health wait, now reporting `restarted`. **`repair` keeps the no-op**: a repair of a healthy service must not be an outage, and only the verb that promises a new process costs one. `normalizeServiceSubcommand` therefore stops folding `restart` into `repair` (a bare `ocx service` still selects `repair` — it is an idempotent "make it current", not a request to bounce a healthy hub). **Windows and Linux are unchanged and were checked rather than assumed**: the scheduler repair already stops then starts the task, WinSW repair restarts the service, and `installSystemd` ends in an unconditional `systemctl --user restart`, so neither platform has a no-op to compensate for and neither reads the verb. One related message moved with it: `src/cli/version-skew.ts` advised `ocx service repair` for an old running proxy, which a byte-identical definition now makes a no-op, so it advises `ocx service restart`.

**The "loaded" bit is now a four-state verdict.** New `probeLaunchdLoadState()` asks `launchctl print` in **both** `gui/<uid>` and `user/<uid>` the way `inspectLaunchd` does, keeps the 112/113 distinction, and returns `loaded-current` / `loaded-stale` / `not-loaded` / `unknown`. `deriveLaunchdServiceDiagnostic()` is pure, so all four are testable without a live launchd.

`unknown` is the one that used to do damage, and it is handled in both directions. Its summary says the state could not be verified and names **no** repair command. And it keeps `viable` **true**, because `isServiceViable() === false` is precisely what makes the update fallback treat a successful repair as a dead supervisor and start a competing proxy; a probe that could not be run is not evidence against the service. `startable` is likewise untouched, so the tray still hands the start to `ocx service start`, which no-ops on an already-loaded job. `src/cli/status.ts` still appends "registered but NOT serving … re-run 'ocx service repair'" only when `installed && !live`, which stays honest: `unknown` above a live proxy prints the unverified summary under `✅ Proxy: running` and recommends nothing. `loaded-stale` keeps the viability the grep era gave it, so the update path behaves exactly as before, and only gains a summary that finally says the live job came from an older plist — the one case where `repair` is right.

Around it: the install-cleanup twin uses the same probe and `bootout` in both domains (legacy `unload` cannot evict a gui-domain job, as the file's own comment says), treats 0/3/112/113 per domain as benign and throws on anything else, and still fails closed on `unknown`, because installing new assets over a manager nobody could query is the unsafe direction. `stopLaunchd`/`uninstallLaunchd` use `bootout` in both domains too, with `unload` kept only for `status === null`, i.e. launchctl could not be spawned at all. And `serviceCommand`'s repair branch wraps `repairService()` so `reportServiceServing("repaired")` runs even on a throw — which matters more now that darwin repair can roll back, since the operator needs the "did anything come back?" answer. The exit code stays non-zero either way.

**Two pre-existing test-safety holes, both of which were hitting the maintainer's own machine.** `os.homedir()` reads the password database rather than `$HOME`, so the suite's HOME sandbox never moved `~/Library/LaunchAgents` and the existing `withLaunchAgentHome()` helper was inert: every `installLaunchd` case rewrote the live `com.opencodex.proxy.plist` with a definition whose token file, log path and homes pointed into a temp directory. launchd keeps its own parsed copy, so nothing broke until the job next restarted. `assertNotRealLaunchAgentsUnderTest` plus an injectable `plistPath` make that refusal mechanical. Separately, `serviceStatePaths()` now drops its legacy default-home entry under an armed test process — that entry is the real `~/.opencodex/service-state.json`, and a sandboxed run was observed replacing the live record's `codexHome` and `opencodexHome` with `/var/folders/...` paths. That filter asks the guard's own `isProtectedHomeUnderTest()`, so one canonicalization decides it (a local `resolve()` calls `/var/folders/...` and `/private/var/folders/...` different directories on macOS), and `writeServiceInstallState` goes through `serviceStateWritePaths()`, which **throws** when the filter leaves nothing: with `OPENCODEX_HOME` unset under an armed guard both candidates resolve to the real home, the list went empty, and the writer silently wrote nowhere while reporting success. Reads stay quiet, because an empty read list really does mean "no install state".

Issue defects 3 and 4 (an occupied secondary listener port misdiagnosed as a public-port conflict; `ocx status` comparing client fences against the public port only) are left to the next PR in this stack, which owns the loopback listener. `startLaunchd` deliberately keeps `load -w`: switching it would change `ocx service start` semantics, since `-w` clears the disabled list and `bootstrap` does not, and it already cross-checks with `launchdJobMatchesPlist` before throwing.

This PR targets `dev` and is the base of a four-PR hub single-port stack.

## Verification

- `tests/service/launchd-repair.test.ts`, **54 cases**, registered in both `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`. Every case injects `plistPath` into its own fixture directory and a scripted tri-state `probe`, so none of them can reach the real LaunchAgents path or a live launchctl — and one case asserts the guard refuses that path. `OPENCODEX_HOME` is pinned per case and restored in `afterEach`, so this file no longer leaks its temp home into the next file in the same Bun worker.
- Coverage: the healthy-and-identical repair makes **zero** launchctl calls; a plist differing **only** in the baked PATH is also a no-op and keeps the installed PATH on disk; an identical plist whose live job runs an older command still reloads and deletes `<plist>.prev`; the reload is domain-explicit `bootstrap` preceded by a `bootout` of **both** domains with no `load`/`unload` anywhere; the settle loop waits while `print` answers 0, stops at 5 × 200 ms per evicted domain, and is skipped for a `bootout` that exited 3; exit 0 with a disagreeing probe is a failure whatever stderr says; exit 5 tries `kickstart -k` for unchanged bytes and **refuses to trust it** for an env-only change (two bootstraps, no kickstart); a disabled job takes `enable` + bootstrap in a pinned twelve-verb sequence while an ordinary repair never runs `enable`; a malformed plist is not retried; terminal failure restores the previous bytes and names the manual remedy and `print-disabled`; a `loaded-stale` outcome is reported as loaded rather than down; a fresh install invents no rollback; an `unknown` pre-check changes nothing at all — no verb, no file, no `.prev` — and names both `print` commands; an `unknown` after an accepted bootstrap neither retries nor rolls back; an `unknown` after a refused one throws without "IS NOT RUNNING". Plus `reusePreviousPlistPathVariable` (PATH-only, anything-else, identical, a PATH containing `$&`, a definition with no PATH entry), `launchdEvictionTargets`, the probe tri-state from 0/112/113 and a spawn failure including "113 in `gui/` is not absence, ask `user/` too", and all four diagnostic states with `unknown` producing neither "installed, not loaded" nor a repair recommendation and not disproving `viable`. Source-oracle cases cover what cannot be driven without a live launchd or a whole CLI process: the repair branch's try/catch ordering, the install-cleanup ops (both domains, benign statuses), `installLaunchd` never reaching for `launchdJobMatchesPlist` while `startLaunchd` deliberately still does, the PATH pre-check, the state-path filter plus its fail-loud write path, stop/uninstall, and the launcher resolver shared with `installSystemd`.
- The five obsolete `installLaunchd` cases in `tests/service/service.test.ts` asserted the `load` verb and the stderr success condition, so they are replaced rather than extended; what they proved about bounded retries and about not retrying an unrelated failure is preserved in the new file. `stableLauncherEntry`'s two discovery cases now pass `state: null`, one new case covers the recorded-launcher preference (and the refusal of a relative value), and another covers the systemd half of it.
- The restart verb adds 8 cases to that file: restart of a healthy loaded-current job runs exactly `kickstart -k` on the gui domain and **no** `bootout`/`bootstrap`; repair of the same state runs zero launchctl calls and never reaches the restarter; restart of a not-loaded job takes the ordinary evict/bootstrap path with no kickstart; `installLaunchd` returns `{ reloaded: false }` / `{ reloaded: true }` for the two paths; `restartLaunchdJob` prints the one line naming the command it ran, throws when the job is gone afterwards, and only warns on `unknown`; and a source-oracle case pins the darwin wiring (restart verb only, defaulting to the real restarter) together with the unconditional `systemctl --user restart` that makes Linux need no equivalent. Every one of them injects the restart seam — the default would kickstart the maintainer's live hub, and `kickstart` is not on the live-service-manager guard's read-only list, so even a default call from an armed test process fails closed. `tests/service/service.test.ts` covers the verb surviving `normalizeServiceSubcommand`, `planServiceCommand` and `selectServiceSubcommand`, plus the shared dispatch branch.
- Commands run, from a worktree with `node_modules` symlinked:
  - `bun run typecheck` → clean.
  - `bun test tests/service/launchd-repair.test.ts` → **54 pass, 0 fail**, 195 expect().
  - `bun test tests/service/service.test.ts` → **205 pass, 0 fail**, 674 expect().
  - `bun test tests/cli/cli-version-skew.test.ts` → **29 pass, 0 fail**; `bun test tests/cli/cli-help.test.ts` → **17 pass, 0 fail** (both touched by the verb's wording).
  - `bun test tests/service/launchd-repair.test.ts tests/service/service.test.ts tests/test-layout.test.ts tests/test-layout-tooling.test.ts` → **268 pass, 0 fail**, 1388 expect().
  - `bun test tests/service tests/update tests/cli/uninstall.test.ts` → **765 pass, 9 fail**.
  - `bun run privacy:scan` → passed.
- **Those 9 failures are pre-existing**: 8 `winsw` cases plus `xAI API-key runtime injects priority while OAuth does not`. They are cross-file `OPENCODEX_HOME` pollution inside one domain-wide `bun test` invocation — `bun test tests/service/winsw.test.ts` alone → 25 pass — and the same 9 failed on `dev` before this branch existed. Not touched here.
- **The full suite was not run, by operator instruction** (focused files plus typecheck only). Hosted CI on the pushed head is the proof.
- The launchctl exit codes this change depends on were measured on the affected host rather than assumed, with a throwaway `com.opencodex.test-probe` label in a temp plist. The live `com.opencodex.proxy` job was never booted out, bootstrapped, kickstarted or enabled — in either round: `launchctl print gui/501/com.opencodex.proxy` returned 0 before and after every probe, and the probe was booted out and its files deleted. `launchctl enable`/`disable` write a per-uid override database with no removal verb, so `print-disabled` retains an inert `"com.opencodex.test-probe" => enabled` record for a label that no longer exists. The full table is in the devlog.
- The two real files the pre-existing test holes damage were verified and restored again at the end of this round: plist 1985 bytes, 0600, `plutil -lint` OK, command and `EnvironmentVariables` byte-identical to the running job's own `launchctl print`; `service-state.json` 320 bytes with the real homes; `service-api-token` untouched; `/healthz` on 10100 → 200; no `*.prev` left behind. The damage happened **twice** during this round and neither time from this branch's tests: every `installLaunchd` write here is refused by `assertNotRealLaunchAgentsUnderTest`, and the only other writer of that path (`uninstallLaunchd`) is guarded identically. Concurrent sessions in other worktrees ran the *unguarded* suite and wrote `/var/folders/.../opencodex-test-*` homes into both files; the clobbered state record names its writer in `cliPath` (once `.claude/worktrees/agent-ae68eb45f20e1f3c6`, once a sibling agent's scratchpad worktree). launchd keeps its cached copy, so nothing broke until the next restart would have — which is both the case for landing these guards and the reminder that a guard only protects the branch carrying it. Worth re-checking the two live files once any concurrent session finishes.
- `devlog/_plan/260911_hub_single_port/010_launchd_repair.md` records what shipped, the review round and each fix, the measured launchctl semantics, the decisions, and these commands with their counts.

## Checklist

Refs #4236

## Summary by CodeRabbit

- **New Features**
  - Added distinct `service repair` and `service restart` behaviors.
  - `repair` avoids interrupting healthy services when no changes are needed.
  - `restart` explicitly restarts services, including unchanged macOS jobs.

- **Bug Fixes**
  - Improved service installation, recovery, rollback, stopping, and removal reliability.
  - Improved handling of stale, missing, disabled, and unverifiable service states.
  - Improved service verification across supported platforms.

- **Diagnostics**
  - Service status now reports clearer loaded, running, viable, and startable states.

### DIFF
--- a/devlog/_plan/260911_hub_single_port/010_launchd_repair.md
+++ b/devlog/_plan/260911_hub_single_port/010_launchd_repair.md
@@ -1,0 +1,226 @@
+# PR1 — macOS launchd repair/status (issue #4236 defects 1 & 2)
+
+Branch `codex/260911-l4-launchd-repair`, based on `dev` (`babb76449`). First of the four-PR
+hub single-port stack; the others target this branch's head in turn.
+
+Scope: `src/service.ts` macOS path only, plus the two shared test-safety guards the work
+uncovered. Defects 3 and 4 from the issue (secondary-port misdiagnosis, `ocx status` fence
+comparison) are deliberately left to PR2, which owns the loopback listener.
+
+## What shipped
+
+### A — `installLaunchd()` (repair must not be an outage)
+
+1. **No-op pre-check.** The plist is rendered BEFORE anything is written. If the rendered
+   bytes equal the on-disk bytes, the data-token file is unchanged, and
+   `launchdJobMatchesPlist()` reports `{loaded:true, matchesPlist:true}`, the function
+   re-asserts 0600 on the plist, refreshes install state, logs
+   `service is already loaded from the current plist; nothing to do.` and returns. launchd is
+   not touched at all. This is the headline fix: a repair of a healthy hub used to evict it
+   unconditionally.
+2. **Backup + rollback.** The previous plist bytes are held in memory and copied to
+   `<plist>.prev` before the overwrite. On terminal failure the bytes go back, a
+   bootout/settle/bootstrap tries to re-register them, and the thrown error states whether
+   that worked.
+3. **`bootstrap gui/$uid <plist>` replaces `load -w`.** `bootout` was already
+   domain-explicit; `load` acts on the CALLER's bootstrap domain, so from ssh/cron/another
+   bootstrap context the old pair deleted the gui-domain job and registered nothing.
+4. **Bounded settle between bootout and bootstrap** — up to 5 × 200 ms while
+   `launchctl print gui/$uid/<label>` still answers 0, the launchd twin of the Windows
+   `SCHEDULER_SETTLE_DELAYS_MS` idea. `bootout` is asynchronous, so the old back-to-back
+   retry raced the same exiting job twice and added nothing.
+5. **Success is `launchctl print` agreeing, never stderr.** `launchdJobMatchesPlist` is
+   consulted against the command this install actually baked, and `writeServiceInstallState`
+   runs only after it agrees. Stderr regexes are advisory routing signals now.
+6. **One retry, routed by the failure.** Exit 5 / `Bootstrap failed` → `kickstart -k`, then
+   `enable` + a second bootout/bootstrap (see the launchctl findings below). Exit 0 with a
+   disagreeing `print` → one more bootout/bootstrap, because that is the silent no-op. Any
+   other failure (malformed plist, EPERM) throws immediately so the real stderr reaches the
+   operator undelayed — the property the previous code had and kept.
+7. **Error text names the outage and the remedy**: the job was evicted from `gui/<uid>` and
+   is not running, `launchctl bootstrap gui/<uid> <plist>` recovers it, plus
+   `launchctl print` and `launchctl print-disabled` to inspect.
+8. **`stableLauncherEntry()` prefers the recorded launcher** when it is still an absolute
+   executable file, falling back to the PATH walk otherwise (defect 1g). A repair from a
+   context without `ocx` on PATH no longer rewrites a working launcher-form plist into the
+   version-pinned Bun + CLI pair.
+
+### B — `statusLaunchd()` / `diagnoseService()` darwin branch
+
+`launchctl list | grep <label> || true` is gone. New `probeLaunchdLoadState()` asks
+`launchctl print` in BOTH `gui/<uid>` and `user/<uid>` (the way `inspectLaunchd` does), keeps
+the 112/113 distinction, and returns a four-state verdict:
+
+| state | `running` | `viable` | summary |
+|---|---|---|---|
+| `loaded-current` | true | `!stale` | `installed and loaded (launchd; …)` |
+| `loaded-stale` | true | `!stale` | `installed and loaded from an OLDER plist (launchd; …)` |
+| `not-loaded` | false | false | `installed, not loaded (launchd; …)` |
+| `unknown` | false | `!stale` | `installed; launchd state could not be verified — <reason> (launchd; …)` |
+
+`deriveLaunchdServiceDiagnostic()` is a pure function so all four are testable without a live
+launchd. `platformServiceInstallCleanupOps` (the install-cleanup twin) now uses the same probe
+and `bootout` instead of `launchctl list` + legacy `unload`, and keeps failing closed on
+`unknown` — installing new assets over a manager we could not query is the unsafe direction.
+
+### C — `serviceCommand` repair branch
+
+`repairService()` is wrapped, so `reportServiceServing("repaired")` runs even when repair
+throws. The failure is printed and `process.exitCode` is 1 either way. This matters precisely
+because darwin repair can now roll back: the operator needs the "did anything come back?"
+answer, and a throw used to escape to the top level and skip it.
+
+### D — `stopLaunchd` / `uninstallLaunchd`
+
+`bootout gui/<uid>/<label>` first; legacy `unload` survives only for `status === null`, i.e.
+launchctl could not be spawned at all. Exit 3 ("No such process") is the not-loaded case, not
+a failure. `uninstallLaunchd` routes through `stopLaunchd` and also removes `<plist>.prev`.
+
+### Two test-safety guards this work uncovered
+
+Both were pre-existing, and both were hitting this machine.
+
+- **`assertNotRealLaunchAgentsUnderTest`** (`src/lib/test-home-guard.ts`) plus an injectable
+  `plistPath` on `installLaunchd`. `os.homedir()` reads the password database, not `$HOME`, so
+  the suite's HOME sandbox does not move `~/Library/LaunchAgents`. The existing
+  `withLaunchAgentHome()` helper in `service.test.ts` was therefore inert, and every
+  `installLaunchd` case rewrote the developer's live `com.opencodex.proxy.plist` with a
+  definition whose token file, log path and homes pointed into a temp sandbox. launchd holds
+  its own parsed copy, so nothing broke until the job next restarted.
+- **`serviceStatePaths()` drops the legacy default-home entry under an armed test process.**
+  That entry exists so an install made before `OPENCODEX_HOME` was set can still be found, but
+  it is the real `~/.opencodex/service-state.json`, so a sandboxed test still wrote the live
+  record. Observed directly: one run replaced this host's `codexHome`/`opencodexHome` with
+  `/var/folders/...` paths.
+
+Both real files on this machine were restored from the live job's own
+`launchctl print` output and re-verified (`plutil -lint` OK, command identical to the running
+job, 0600, `/healthz` 200). Nothing was left in `~/Library/LaunchAgents` by this work.
+
+## launchctl semantics, measured on this host
+
+macOS 26 / Darwin 27.0.0, arm64, uid 501. Measured with a throwaway
+`com.opencodex.test-probe` label in a temp plist running `/bin/sleep 600`, never against the
+live `com.opencodex.proxy` job (`launchctl print gui/501/com.opencodex.proxy` returned 0
+before and after every probe). The probe was booted out and its files deleted.
+
+```
+launchctl print gui/501/<label>            → 113 absent, 0 loaded
+launchctl print user/501/<label>            → 113 (the shipped agent lives in gui/ only)
+launchctl print gui/99999/<label>           → 112 (no such domain)
+launchctl bootstrap gui/501 <plist>         → 0 first time
+launchctl bootstrap gui/501 <plist> (again) → 5  "Bootstrap failed: 5: Input/output error"
+launchctl load -w <plist> while bootstrapped→ 0  AND "Load failed: 5: Input/output error"   ← the silent no-op
+launchctl kickstart -k gui/501/<label>      → 0 when loaded, 113 when absent
+launchctl bootout gui/501/<label>           → 0 when loaded, 3 "Boot-out failed: 3: No such process"
+```
+
+Second probe, the finding that changed the design:
+
+```
+launchctl disable gui/501/<label>           → 0
+launchctl bootstrap gui/501 <plist>         → 5  "Bootstrap failed: 5: Input/output error"   ← same code, different cause
+launchctl kickstart -k gui/501/<label>      → 113
+launchctl load -w <plist>                   → 0, job loaded, and print-disabled flips to "enabled"
+launchctl enable gui/501/<label>            → 0
+launchctl bootstrap gui/501 <plist>         → 0, job loaded
+```
+
+So **exit 5 is ambiguous**: either something is still bootstrapped under the label, or the
+label sits in the domain's disabled list. The `-w` in the legacy `load -w` was doing the
+second job silently, and dropping it without `enable` would have made a disabled job
+permanently unrepairable. The retry therefore tries `kickstart -k` (live job) and then
+`enable` + bootstrap (disabled list), in that order. `enable` runs ONLY on that retry, so an
+ordinary repair does not quietly undo a deliberate `launchctl disable`.
+
+Residue worth noting: `launchctl enable`/`disable` write a per-uid override database, so
+`launchctl print-disabled gui/501` now carries an inert `"com.opencodex.test-probe" => enabled`
+entry for a label that no longer exists. There is no launchctl verb to remove an override
+record; it references nothing and is harmless.
+
+## Decisions
+
+- **`unknown` keeps `viable` true.** `isServiceViable() === false` is what makes
+  `src/update/index.ts` (after a service refresh exits 0) and `src/update/job.ts:1229` treat a
+  healthy supervisor as dead and start a COMPETING proxy on the service's own port. A probe
+  that could not be run is not evidence against the service, so `unknown` reports honestly in
+  the summary — no "not loaded", no `ocx service repair` — and does not disprove viability.
+  `startable` is likewise untouched, so the tray still hands the start to `ocx service start`,
+  which no-ops on an already-loaded job. `src/cli/status.ts:216`-`222` still appends
+  "registered but NOT serving … re-run 'ocx service repair'" only when `installed && !live`,
+  which stays honest: with `unknown` AND a live proxy it prints the unverified summary under
+  `✅ Proxy: running` and recommends nothing.
+- **`loaded-stale` keeps the viability the grep era gave it** (loaded ⇒ viable). Only the
+  summary is upgraded, so the update fallback behaves exactly as before while the operator
+  finally learns the live job came from an older plist — the one case where `repair` is right.
+- **Verification compares the command THIS install baked**, via a new shared
+  `launchdServiceCommand()` that `buildPlist` also uses, not
+  `expectedLaunchdCommand(installedServiceListenPort())`. A fresh install has no install state
+  yet, and a lost state file makes `expectedLaunchdCommand` fall back to the Bun + CLI pair —
+  which would call a correctly loaded launcher job stale (#3464) and turn every first install
+  into a rollback.
+- **`<plist>.prev`, not `<plist>.prev.plist`.** launchd globs
+  `~/Library/LaunchAgents/*.plist` at login, so a backup ending in `.plist` would be a second
+  registration of the same Label fighting the real one for the port.
+- **`installLaunchd` stays synchronous** (`ServiceOps.install` and
+  `RepairServiceDeps.repairLaunchd` are `() => void`), so the settle loop uses
+  `Bun.sleepSync` behind an injectable `sleepSync` seam rather than making the whole
+  call chain async.
+- **`startLaunchd` still uses `load -w`, deliberately.** Switching it would change
+  `ocx service start` semantics (the `-w` clears the disabled list, `bootstrap` does not), and
+  it already cross-checks with `launchdJobMatchesPlist` before throwing. Out of scope here.
+- **`repairService()` itself still does not consult `diag.running`/`diag.viable`** (issue
+  defect 1c). The no-op now lives inside `installLaunchd`, which is the only function that
+  knows whether the rendered plist differs — a check in `repairService` would have to re-render
+  it to be correct.
+
+## Tests
+
+New `tests/service/launchd-repair.test.ts` (30 cases), registered in
+`scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`. The five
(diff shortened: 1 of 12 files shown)


lidge-jun:
## 리뷰 · 우선순위 74 / 80

이 PR은 지금 `dev`(HEAD `babb76449`, #4240 L4 client-catalog 직후) 위에서 macOS 허브가 `ocx service repair` 한 번에 통째로 꺼지는 문제를 고칩니다. 이슈 #4236의 결함 1·2만 다룹니다. 허브 단일 포트 스택의 맨 아래 칸입니다. 위에 #4250 → #4251 → #4252가 이 브랜치를 베이스로 쌓여 있고, 결함 3·4(부포트 오진·`ocx status` fence)는 다음 PR 몫으로 남겨 둔 상태입니다.

지금 `dev`의 `src/service.ts` 안 `installLaunchd`는 먼저 `bootout`으로 살아 있는 잡을 쫓아내고, 도메인을 명시하지 않는 옛 `launchctl load -w`로 다시 올리고, stderr가 비면 성공으로 칩니다. 검증이 없습니다. 그래서 공개 프록시·관리 ingress·루프백 리스너가 한 명령에 같이 내려가고, 롤백도 없고, CLI가 복구 명령도 안 알려 줍니다. 같은 파일의 `diagnoseService` darwin 쪽은 `launchctl list | grep`에 `|| true`를 붙여 “로드됨” 한 비트만 뽑습니다. 호출자 세션 도메인만 보고, 실패도 전부 “안 올라감”으로 뭉갭니다. 그 비트가 `enabled`/`running`/`viable`로 흘러 `isServiceViable()`까지 가고, `src/update/index.ts`와 `src/update/job.ts`는 그걸 보고 서비스 포트에 경쟁 프록시를 띄울지 결정합니다. 건강한 허브를 “죽었다”고 오진하면 repair 추천 → 실제 장애, 혹은 업데이트 경로의 이중 프록시로 이어집니다.

이 PR이 넣는 프로토콜은 대략 이렇게 읽힙니다. 새 `probeLaunchdLoadState`가 `gui/$uid`와 `user/$uid` 둘 다에 `launchctl print`를 묻고 `loaded-current` / `loaded-stale` / `not-loaded` / `unknown` 네 값을 돌려줍니다. `installLaunchd`는 파일을 쓰기 전에 이 프로브를 보고, `unknown`이면 손대지 않고 거부합니다. 이미 `loaded-current`이고 plist·토큰이 같으면 launchctl을 한 번도 안 부르고 “nothing to do”로 끝냅니다. PATH만 다른 경우는 `reusePreviousPlistPathVariable`로 설치본 PATH를 되살려 통째 비교가 맞게 합니다. 진짜로 고쳐야 할 때만 `<plist>.prev` 백업 뒤 양쪽 도메인 `bootout` → settle → `bootstrap gui/$uid`, 성공은 프로브가 `loaded-current`일 때만, 실패 시 롤백과 “지금은 내려간 건지 / 다른 커맨드로 올라와 있는지”를 구분한 에러를 냅니다. `kickstart -k`는 디스크에 이미 있는 바이트에만 믿고, 새 바이트면 enable+재퇴거로 갑니다. `deriveLaunchdServiceDiagnostic`은 순수 함수라 `unknown`일 때 `viable`을 참으로 남겨 업데이트 폴백이 경쟁 프록시를 안 띄우게 합니다. `stableLauncherEntry`가 기록된 런처를 우선해서, PATH 없는 tray/`ocx update` 자식에서도 런처형 plist를 Bun+CLI 쌍으로 바꾸지 않습니다. 이 함수는 `installSystemd`도 쓰므로 Linux 쪽 silent rewrite도 같이 막습니다. 테스트가 산 개발자 LaunchAgent·`service-state.json`을 덮어쓰던 구멍은 `assertNotRealLaunchAgentsUnderTest`와 `serviceStateWritePaths`로 막았습니다. `tests/service/launchd-repair.test.ts`가 프로토콜·PATH 재사용·양쪽 도메인·unknown 거부를 직접 커버합니다.

라인 1193 - `probeLaunchdLoadState`가 `printedText.includes(expected)`로 current/stale을 가릅니다. 기대 커맨드 문자열이 print 출력 어딘가에만 있으면 current로 칩니다. 지금은 커맨드가 길고 특이해서 실무 위험은 낮지만, 부분 문자열이라 이론상 오탐 여지가 있습니다.
라인 1213-1214 - settle이 최대 5×200ms(약 1초)입니다. 부하 큰 Mac에서 `bootout`이 더 늦으면 Bootstrap failed: 5 재시도 경로로 넘어갑니다. 재시도가 있어서 치명적이진 않지만, CI/실기에서 settle 부족 신호가 나오면 횟수만 늘리는 게 안전합니다.
라인 2668-2750 - 퇴거는 `gui`+`user` 둘 다인데 재등록은 항상 `bootstrap gui/$uid`만 합니다. headless( gui 도메인 112 )에서는 의도적으로 not-loaded/실패로 떨어지게 설계했지만, user 도메인만 쓰는 특수 환경은 이 PR 범위 밖입니다.
`startLaunchd` / `launchdJobMatchesPlist` - start 경로는 의도적으로 옛 2상태 헬퍼를 유지합니다. repair/status만 4상태로 바꿨습니다. 문서와 소스 오라클에 명시돼 있어 버그는 아니지만, 나중에 start도 같은 프로브로 맞출지 한 번은 결정이 필요합니다.
로컬 스위트 - PR 본문에 “로컬 스위트 고의로 안 돌림, hosted CI가 증거”라고 적혀 있습니다. 지금 메인 test/macos shard는 아직 pending이고, 전용 `macos-launchd` 잡은 통과로 보입니다. merge 전에 전체 shard 녹색을 확인하는 게 맞습니다.

메인테이너의 판단이 필요한 지점
- 이 PR을 #4236 스택의 베이스로 먼저 `dev`에 넣을지, CI 전부 녹색 후에만 넣을지
- `startLaunchd`를 이번 트레인에서 4상태 프로브로 올릴지, 아니면 “start는 그대로 / repair·status만”으로 이번 범위를 고정할지
- settle 1초 상한을 실기 허브에서 충분한지, 아니면 보수적으로 늘릴지
- headless/ssh-only Mac에서 repair가 “거부하고 안내만” 하는 현재 동작이 운영 정책과 맞는지

너의 추천
CI 메인 test·macos shard가 녹색이면 #4249를 스택 1번으로 먼저 merge하세요. #4236 결함 1·2의 실제 장애를 끊는 베이스이고, 위에 얹힌 #4250/#4251/#4252는 이 브랜치를 기다리므로 여기가 막히면 전체가 멈춥니다. merge 후 이슈 #4236은 결함 3·4가 남을 테니 바로 close하지 말고, PR2 쪽으로 범위를 남기세요. 로컬 풀스위트는 운영자 지시대로 스킵해도 되나, merge 직전 hosted CI 전체와 `macos-launchd` 재확인은 필수입니다.

이 댓글은 grok-bot이 작성했습니다

PR #4282, opened 2026-09-11
provider: follow the DeepSeek V4.1 transition through the registry

lidge-jun:
## Summary

DeepSeek moved the whole V4 name set on 2026-09-10, and opencodex had both retired ids hand-written into thirteen provider presets. This follows the transition through the registry, the adapters and the docs.

**What upstream changed** ([V4.1-Flash announcement](https://api-docs.deepseek.com/news/news260910/), [API docs](https://api-docs.deepseek.com/)):

- V4.1-Flash shipped on 2026-09-10. The first-party API id is `deepseek-flash`.
- `deepseek-v4-flash` and `deepseek-v4-flash-vision-exp` retire as models; the names survive as compatibility aliases routed to V4.1-Flash.
- `deepseek-v4-pro` retires from 2026-09-14 04:00 UTC, and its requests route to V4.1-Flash until a V4.1 Pro exists.

**1. The thinking set now derives three ways.** One shared list could not express this transition, because the spelling depends on who serves the route: the first-party API answers to `deepseek-flash`, while the Zen gateway exposes `deepseek-v4.1-flash` (#4253, and the Command Code rows in this PR). A single list either starved the new id of its ladder and `reasoning_content` replay — the `deepseek` preset derives five model maps from that constant — or leaked the gateway spelling into the first-party preset. `DEEPSEEK_NATIVE_THINKING_MODELS` and `DEEPSEEK_GATEWAY_THINKING_MODELS` split it, and a parity test pins both directions including the negatives.

**2. `deepseek-v4-pro` is gone, by two different mechanisms.** Static rosters drop the row. Live-discovery providers cannot be handled that way: deleting their row removes the context window, the effort ladder and the text-only hint while the model keeps arriving from `/models`, so the retired route would stay visible with its capabilities broken — a regression, not a removal. Those eight slugs go into `ROUTED_MODEL_COMPATIBILITY_EXCLUSIONS`, which is what actually takes a slug out of the routed catalog. `volcengine-agent-plan` moves its default onto the roster entry that survives.

**3. Command Code effort ladders** for `deepseek/deepseek-v4.1-flash` and `Qwen/Qwen3.8-Flash`, carried from #4258 with the original commit and a `Co-authored-by` trailer. That PR was reviewed and its CI was green; updating its branch onto current `dev` to clear an unrelated version-line gate reset the contributor readiness checklist, and ticking an author attestation on someone's behalf is not something a maintainer should do. Credit stays with @gitgarmin.

[...]

PR #4281, opened 2026-09-11
fix(cursor): capture the conversation checkpoint a suspended tool turn actually sends

lidge-jun:
## Summary

Closes #4245. A Cursor turn that emits a client tool full-replayed every time, reporting `cached_tokens: 0`, while the same account cache-hits under `cursor-agent` directly.

The cause is not the one the issue was first triaged with. Instrumented against a live Cursor account, the refusal is over-determined: `checkpoint-commit-refused` reports `capturedBytes: 0` for the native router, native composer **and** external Claude alike, so the `isCursorExternalWireModel` gate one line above it is never reached. Relaxing that gate — the obvious-looking fix — changes nothing.

What actually happens: a client-tool turn suspends before `turnEnded`, and upstream sends that turn's `conversationCheckpointUpdate` right after `toolCallStarted`. `CLIENT_TOOL_FINALIZE_GRACE_MS` is **50 ms**, so `cancelCursorRun()` kills the stream first and the bytes never arrive.

Measured on macOS against a real account, with the tool catalog held constant on the wire so only the local grace varies (`parallel_tool_calls` is read locally and never protobuf-encoded):

| Local grace | Post-`toolCallStarted` checkpoint | `capturedBytes` |
|---|---|---|
| 50 ms | no | 0 |
| 1500 ms | yes | 3036 |

Reproduced in both directions within one session.

## What this changes

A drained client-tool turn gets **one** bounded extension when a checkpoint is wanted and none has arrived, and fires early the moment one does.

Three details are load-bearing rather than incidental:

- The extension runs **before** `finalizeAfterDrain`. That call reaches `finalizeTurnEvents`, which sets `state.terminated`, and then returns `[]` for a terminated state — draining first and re-arming would make the retry return early at the length check and leave the stream uncancelled. The new predicate mirrors its two guards instead of calling it.
- The early fire is deferred one tick, so the checkpoint frame finishes being mapped and pushed before the terminal events go out.
- A turn that never sends a checkpoint pays the window once and still dies at a known deadline. Visible in the suite: the no-checkpoint case finalizes at 1816 ms, the checkpoint case at 256 ms.

[...]

PR #4284, opened 2026-09-11
fix(codex): let an operator account selection outrank the pool cursor

lidge-jun:
## Summary

Fixes the reported defect: the pool moves the active Codex account to B, the operator then selects A through the dashboard or `ocx account use`, and the runtime keeps serving B.

The cause was not the write path the operator takes. GUI and CLI both issue the same `PUT /api/codex-auth/active`, and that handler already clears the runtime cursor. The cause is that nothing stopped the next automatic pick from immediately overwriting it again: `rememberActiveCodexAccount` is reached from the round-robin commit, the fill-first commit, the failover promote and priority preemption, and none of them consulted the operator's choice. The pin was only ever a priority-tier ceiling, not a selection input.

This adds an unspent one-shot preference, mirroring the `manualPreference` shape Anthropic already has and Codex lacked.

Three things make it safe, and each was forced by a measured failure rather than guessed:

- **It is spent by success, not by time.** `recordCodexUpstreamOutcome` consumes it on a successful dispatch for that scope. Wiring the guard without a consume site first makes the preference permanent and freezes the automatic cursor: that shape fails 15 of 69 rotation tests.
- **The failover promote is exempt.** It only runs because the account in use just failed, so it is never an automatic pick competing with the operator. Guarding it traps routing on a cooled account and fails 6 rotation tests, all of them 429 promotion cases.
- **Reads are untouched.** `getEffectiveActiveCodexAccountId` keeps its existing contract. Overlaying the preference there breaks the callers that now treat it as the live shared cursor, including health evidence and the management snapshots.

The user-facing rule, stated plainly: an account past its switch threshold is temporarily spent, not wrong. The pool serves the request from another account, and the operator's selection stays pointed where the operator put it, so the window rolling over returns routing to it without a second manual pick.

### What the last audit round changed

The first version of this PR shipped three tests that **all passed with the production change reverted**. They re-asserted what `resetCodexRoutingForManualSelection` and the deliberately exempt failover promote already did, so the guard, the preference map and the consume site had no coverage at all. Driving them red first surfaced three real defects, now fixed here:

- **Deletion never revoked the preference.** Pause and exclusion route through `reconcileCodexActiveAfterExclusion`; delete does not. It reaches routing through `clearCodexUpstreamHealthForAccount`, which cleared two health maps and left the preference behind. Once the named account is gone nothing can succeed on it, so the one-shot could never be spent and every later automatic write stayed suppressed until process restart. The generation sweep in `reconcileCodexRoutingHealth` had the same hole.
- **The model-detour promote was unguarded**, twelve lines above the preemption site that already yields. Both are automatic picks competing with the operator; only the failover promote earns the exemption.
- **The independent-scope preference entries were dead state** — every write site the guard protects is already skipped for independent scopes, so those keys were seeded and consumed but never read. Removed.

Each replacement case is red against the variant that removes the piece it covers:

| Case | Red against |
|---|---|
| an over-threshold operator account is served around, not replaced | parent branch: reads `b`, expected `a` |
| deleting the preferred account releases the hold | pre-fix head `63217d161`: reads `undefined`, expected `b` |
| a successful dispatch spends the one-shot so the pool may move again | guard without consume: 15 of 69 rotation tests fail |

Design and the full audit trail: `devlog/_plan/260911_account_pool_unification/010_phase1_manual_selection.md`. Six audit rounds and three measured implementation passes are recorded there, including the two failing shapes above and the empty-coverage round.

## Verification

- `bun x tsc --noEmit` - pass
- `bun test tests/codex-integration/codex-pool-rotation.test.ts` - 74 pass, 0 fail
- Red-control: production file reverted to the parent branch, new tests kept - 1 fail (the guard case), proving the coverage is not vacuous
- Red-control: production file reverted to `63217d161` - 1 fail (the delete case)
- `bun run privacy:scan` - passed

## Checklist

### DIFF
--- a/devlog/_plan/260911_account_pool_unification/010_phase1_manual_selection.md
+++ b/devlog/_plan/260911_account_pool_unification/010_phase1_manual_selection.md
@@ -203,6 +203,36 @@ the design may assume.
    accept and document that `GET /api/codex-auth/active` is manual-sticky until the
    preference is consumed.
 
+## Measured: the consume call site is the whole design, not a detail
+
+A first implementation pass built the preference map, the seeding inside
+`resetCodexRoutingForManualSelection`, the `rememberActiveCodexAccount` guard, the
+`getEffectiveActiveCodexAccountId` overlay and the exclusion revoke, and left the
+consume call site unwired. It typechecked, and then
+`tests/codex-integration/codex-pool-rotation.test.ts` went from green to **15 failures
+out of 69**, including "fill-first picks the same sequence with no stored order as
+before the feature".
+
+That is the correct result, and it is worth recording rather than repeating. Without a
+consume site the one-shot is permanent: the first operator selection freezes the
+automatic cursor forever, because `rememberActiveCodexAccount` stays a no-op and no
+code path ever clears the entry. Every rotation-strategy test that expects the pool to
+keep moving after a manual selection fails, and they are right to.
+
+So the implementation order matters. Build the consume path FIRST, not last:
+
+1. Find the point that already records a successful upstream outcome for the resolved
+   account and call `consumeCodexManualPreference(poolKey)` there, for a non-quota
+   success only. This is the Codex analogue of `commitAnthropicSelectionRouting`
+   (`anthropic-routing.ts` :799-800), which Codex has no direct equivalent of.
+2. Only then add the `rememberActiveCodexAccount` guard, so the suite never passes
+   through a state where the cursor can freeze.
+3. Gate honouring on the account being selectable, per blocker 3 above, so a cooled
+   preferred account is skipped for that dispatch without being spent.
+
+The pass was reverted rather than pushed. The branch `codex/manual-selection-wins`
+carries this document and no source change.
+
 The generic OAuth kind gets no preference in this layer; that arrives with the
 kernel in phase 2. No management or GUI change.
 
--- a/devlog/_plan/260911_account_pool_unification/010_phase1_manual_selection.md
+++ b/devlog/_plan/260911_account_pool_unification/010_phase1_manual_selection.md
@@ -233,6 +233,40 @@ So the implementation order matters. Build the consume path FIRST, not last:
 The pass was reverted rather than pushed. The branch `codex/manual-selection-wins`
 carries this document and no source change.
 
+## Measured again: guarding the writer is right, but the 429 path needs an exemption
+
+A second pass followed the order above. The consume site went in first, at the
+`outcomeClass === "success"` branch of `recordCodexUpstreamOutcome` (:2356), keyed by
+`codexPoolKeyForScope(quotaScope)` which that function already computes. Seeding and the
+exclusion revoke followed. At each of those two steps
+`tests/codex-integration/codex-pool-rotation.test.ts` stayed **69 pass, 0 fail**, which
+confirms the ordering advice above is correct.
+
+Adding the `rememberActiveCodexAccount` guard then produced **6 failures out of 69**, down
+from 15, and every single one is a 429 promotion case:
+
+- fill-first 429 advances to next stable account, not lowest usage
+- RR 429 promotes via ring, not lowest usage
+- 429 retry reuse promoteAccountId avoids a second RR ring advance
+- fill-first transient failover advances stable order, not lowest usage
+- scoped reset 429s retain strategy while excluding only the affected native quota
+- fill-first preserves its pre-feature fallback when every ordered tier is drained
+
+That is exactly the hazard blocker 3 named, and it is sharper than the blocker stated it.
+Gating the guard on `isCodexAccountSelectable(preferred)` is NOT sufficient: at the moment
+`promoteActiveCodexAccount` runs, the preferred account can still read as selectable
+because the 429 cooldown is recorded on a different path, so the guard holds and the
+promotion cannot land.
+
+The conclusion for the next pass: guarding the writer closes all four call sites at once,
+which is still the right shape, but the failover promote needs an explicit exemption. It
+only ever runs because the account in use just failed, so it is never an automatic pick
+competing with the operator. Either pass an explicit "this is a failover promote" flag
+through `rememberActiveCodexAccount`, or leave the writer unguarded and guard the two
+strategy commit sites plus preemption instead, accepting three guards rather than one.
+
+Reverted again rather than pushed. The measurement is the deliverable.
+
 The generic OAuth kind gets no preference in this layer; that arrives with the
 kernel in phase 2. No management or GUI change.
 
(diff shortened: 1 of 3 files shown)


lidge-jun:
## 리뷰 · 우선순위 74 / 80

이 PR은 Codex 계정 풀이 이미 활성 계정을 B로 옮긴 뒤에, 운영자가 대시보드나 `ocx account use`로 A를 골라도 런타임이 계속 B를 쓰는 버그를 고칩니다. CURRENT `dev` HEAD는 `c1ce2560e`(패키지 `2.52.0`, tip #4282 DeepSeek V4.1 레지스트리 전환)이고, 이 결함의 무대인 `src/codex/routing.ts`는 그 tip에도 그대로 있습니다. 베이스는 `dev`가 아니라 `codex/pool-kernel`(열린 #4279)이고, 헤드는 `codex/manual-selection-wins`입니다. 스택은 대략 #4275(로드맵 문서) → #4277(키 풀) → #4279(커널 이동) → 이 PR 순서입니다.

원인을 짧게 말하면 이렇습니다. GUI와 CLI는 둘 다 같은 `PUT /api/codex-auth/active`를 타고, 그 핸들러는 이미 `resetCodexRoutingForManualSelection`으로 런타임 커서를 지웁니다. 문제는 그 다음입니다. 라운드로빈 커밋, 필스트 커밋, 페일오버 promote, 우선순위 preemption이 모두 `rememberActiveCodexAccount`로 커서를 다시 씁니다. 예전의 핀은 우선순위 천장일 뿐이라, 운영자 선택을 입력값으로 보지 않았습니다. 그래서 풀이 한 번 옮기면 운영자가 A를 골라도 다음 자동 픽이 바로 커서를 덮어썼습니다.

고치는 모양은 Anthropic 쪽에 이미 있는 `manualPreference`를 Codex에 맞춘 일회성(one-shot) 선호입니다. `resetCodexRoutingForManualSelection`이 공유 풀 키와 독립 쿼터 스코프에 같은 계정 id를 심고, `pickUnboundStrategyAccount`의 remember 두 곳과 preemption remember만 그 선호가 다른 계정을 가리킬 때 막습니다. 성공한 디스패치에서만 `recordCodexUpstreamOutcome`의 success 분기가 `consumeManualPreference`로 소모합니다. 페일오버 promote는 의도적으로 가드 밖입니다. 쿨다운 계정에 가두면 429 승격 테스트 6개가 깨진다는 측정이 문서에 남아 있습니다. `getEffectiveActiveCodexAccountId`는 건드리지 않습니다. 여기 선호를 얹으면 헬스·관리 스냅샷이 라이브 커서로 읽는 계약이 깨집니다.

문서 쪽도 같이 옵니다. `devlog/_plan/260911_account_pool_unification/010_phase1_manual_selection.md`에 구현 진입 감사, consume를 먼저 안 넣으면 69개 중 15개가 깨지는 측정, promote 가드가 429에서 6개를 깨는 두 번째 측정이 적혀 있습니다. 테스트는 `tests/codex-integration/codex-pool-rotation.test.ts`에 세 개를 더합니다. 풀이 옮긴 뒤 수동 픽이 이기고, 성공이 one-shot을 쓰며, 선호 계정 429는 여전히 승격합니다. 본문 검증은 tsc, 풀 로테이션 72통과, 라우팅·auth-api 509통과, privacy:scan 통과입니다. CI는 이 리뷰 시점에 아직 돌아가는 중이라 merge 상태는 UNSTABLE입니다.

지금 `dev`가 최적화하는 축(#4282 레지스트리 정리, #4281 커서 체크포인트, 그 아래 풀·쿼터 열차)과 겹칩니다. 계정 풀 통일 유닛(#4275)의 phase 1 결함 수정이라 방향은 맞고, 사용자에게 바로 보이는 운영자 선택 무시 버그라 점수도 높습니다. 다만 이 PR 자체는 `dev`에 바로 머지할 수 없고, 아래 스택이 먼저 열려야 합니다.

src/codex/routing.ts / 베이스 브랜치 - 베이스가 `dev`가 아니라 `codex/pool-kernel`(#4279)이라, #4275→#4277→#4279가 안 닫히면 이 픽스도 `dev` tip에 못 올라갑니다. 실제 코드 변경은 `routing.ts`와 테스트·phase1 문서뿐이라, 급하면 `dev`로 체리픽·리타겟을 검토할 여지가 있습니다.
라인 pickUnboundStrategyAccount remember 가드 - 가드는 `rememberActiveCodexAccount`만 막고, 함수가 돌려주는 `picked` 자체는 막지 않습니다. RR/fill-first 경로에서는 strategyPick이 있으면 그 값이 곧 응답 계정이 됩니다. 이번 테스트가 통과하는 이유는 reset이 링을 시드해서 다음 픽이 고른 계정이 되기 때문이지, 선호가 resolve 결과를 강제해서가 아닙니다. sticky가 넘어가거나 동시 unbound가 있으면 공유 커서는 A인데 요청은 B로 나가는 틈이 남을 수 있습니다.
경로 applyQuotaAutoSwitch / setActiveCodexAccount - 쿼터 전략의 persist 경로는 선호 맵을 보지 않고 `config.activeCodexAccountId`를 바꿉니다. 보고된 결함은 RR 쪽이지만, 쿼터 모드에서 운영자 선택 직후 자동전환이 선호를 우회할 수 있는지 한 번 더 확인할 필요가 있습니다.
경로 isEffectiveCodexAccountPinned - phase1 문서 감사 6번(핀 판정이 persisted+runtime만 볼지, GET active가 수동 sticky로 남을지)이 코드/주석으로 닫히지 않았습니다. 동작은 바뀔 수 있으니 문서에 결정만 적어도 됩니다.
경로 clearCodexUpstreamHealthForAccount - 전체 clear는 `manualPreference.clear()`를 하지만, 계정 단위 clear는 선호를 지우지 않습니다. 계정이 풀에 남아 있는 한 보통은 괜찮고, exclusion/pause는 `forgetManualPreference`로 처리합니다. 의도인지 한 줄만 확인하면 됩니다.

메인테이너의 판단이 필요한 지점
- 스택 유지(#4279 뒤) vs 사용자 보이는 버그라 `dev`에 체리픽·리타겟해서 먼저 넣을지
- strategyPick이 선호와 다를 때 return 값까지 막을지, 지금처럼 remember만 막을지(시드+커서 가드에 맡길지)
- 쿼터 `setActiveCodexAccount`에도 같은 선호 규칙을 얹을지, RR/fill-first 범위로 끝낼지
- `isEffectiveCodexAccountPinned` / GET active의 수동 sticky 해석을 문서에 어떻게 남길지

너의 추천
CI(크로스플랫폼 test/gates) 초록을 보고, 스택이 그대로면 #4279 머지 직후 이 PR을 이어서 머지하세요. 운영자 선택 무시가 지금 현장 이슈면 `routing.ts`+테스트+phase1 문서만 `dev`로 체리픽하는 쪽을 먼저 검토하세요. 라벨은 바꾸지 않았습니다.

이 댓글은 grok-bot이 작성했습니다

PR #4289, opened 2026-09-11
feat(oauth): let the generic pool consume its strategy behind pool.kernel

lidge-jun:
## Summary

Makes a generic OAuth provider's stored pool `strategy` actually select an account, behind a new opt-in `pool.kernel` flag. Until now `strategy` and `autoSwitchThreshold` were persisted, reported through the management contract, settable from the CLI — and read by nothing. The DTO said so with `inert: true`.

Two paths had to branch, not one. Branching only the pre-dispatch preference would leave the setting inert in practice the moment anything actually failed, which is the case an operator picks a strategy for.

| Strategy | initial preference | 429 rotation |
|---|---|---|
| flag off, or absent / `quota` | unchanged | unchanged |
| `round-robin` | peek the shared ring | advance the ring |
| `fill-first` | hold the active account under its threshold | take the next eligible account, never the cooled one |

Three things are worth calling out, because each was forced by the code rather than chosen:

- **Both quota guards are skipped for the two new strategies.** `hasHeadroomEvidence` refuses every provider with no quota data at all — which is exactly where round-robin is the point — and the healthy-active early return fires before `autoSwitchThreshold` can ever be read, so fill-first could never reach its own test. They still guard the quota answer, which is unchanged.
- **The round-robin commit is at admission, not at proposal.** `peekRoundRobinAccount` never creates the pool state and `notePoolRotationSuccess` returns immediately without it, so a peek-only path leaves the ring with nothing to advance and round-robin proposes the same account forever. `noteGenericPoolSelection` takes the live pick at commit, the same shape `commitAnthropicSelectionRouting` already uses. Its early return is the safety story for the core path: it is reached on every generic first dispatch, so anything but round-robin leaves the cursor untouched.
- **Fill-first walks the sorted full roster.** The account store holds accounts in login order, so two operators who added the same accounts in a different sequence would otherwise rotate differently, and walking the eligible subset changes the wrap order whenever an ineligible id sits between two eligible ones — the bug the shared kernel already carries a `stableAll` argument to avoid.

Also fixed here, the generic half of the defect #4284 closed for Codex: a manual account selection cleared the presence count but never seeded the rotation cursor, so an operator's pick lost the very next dispatch to sticky rotation.

`stickyLimit` joins the generic contract (the CLI verb already routed it; only the server refused it). `quotaWindow` is still refused.

[...]

PR #4292, opened 2026-09-11
feat(providers): use the warm API key on the first attempt, and rank by quota

lidge-jun:
## Summary

#4277 shipped `selectProactiveApiKey` and `forgetApiKeyRotationCursor` and deliberately stopped there: both were unit-tested and called from **nowhere** in production. This connects them, and nothing else.

An API key pool could only react. Rotation needed a 429 first, so a request arriving while the committed key was already cooling was spent earning a refusal the runtime could already predict. That state is not exotic — it is what an operator has after the pool rotated and a restart, a manual edit or a config reload pointed `apiKey` back at the spent key.

**Where the call goes is the whole correctness argument.** `route.provider` is final for a key-auth request at the transport pin in `core.ts`, and all four first-send consumers — the image bridge, web search, `runTurn` and the generic HTTP path — read that same object. One assignment placed ahead of the pin therefore serves every one of them. It has to be ahead of the pin rather than merely before the send, because `adapterProvider` is copied immediately after it, the adapter binds from that copy, and the HTTP path bakes its request later. The HTTP and `runTurn` paths could re-read a stale selection through `refreshDispatchAdapter`; the image bridge and web search call `providerFetch(route.provider)` directly and have no second chance.

Native chat is a separate entry path — `chat-completions.ts` routes there directly and never through the Responses core — so it gets its own call rather than inheriting one.

`forgetApiKeyRotationCursor` joins `clearKeyCooldowns` at the five routes that already reset key state. A cursor that predates an operator's choice would hand the next proactive pick straight back to whichever key the pool had reached.

**No new import edge on the core path.** `core.ts` already imports `hasKeyPoolFailover` from the same module, which matters because it is one of the three files that must never reach `src/lab`.

Scope is deliberately narrow. The picker itself, the reactive 429/401 rotation and the strategy semantics are untouched. It returns null unless a strategy is configured **and** the committed key is cooling, so an install that never set `apiKeyPoolStrategy` evaluates one predicate and stops — and never performs the persisted config write.

Two first-send paths are **not** covered here and are registered as their own work-phase rather than ridden along untested: native compact for `openai-apikey` never enters `core.ts`, and the keyed `/v1/images` path reads `candidates.keyed.apiKey` directly instead of a provider object. Each needs its own dispatch harness.

Design and audit trail: `devlog/_plan/260911_account_pool_unification/040_phase4_key_pool_strategy.md`. The plan audit returned FAIL — the first draft said "wire the call" without assigning the return, which would have been a no-op that still wrote config — and passed on re-audit.

## Verification

- `bun x tsc --noEmit` — pass
- `bun test tests/server/server-key-failover-e2e.test.ts tests/adapters/key-failover.test.ts tests/lab/core-lab-boundary.test.ts tests/providers/provider-api-keys.test.ts` — 76 pass, 0 fail
- Red control: with both call sites removed, the new dispatch test fails with `Bearer synthetic-first` instead of `Bearer synthetic-second`. The companion no-op test correctly still passes, since it pins the unchanged path.
- The Lab boundary suite is included because this touches `src/server/responses/core.ts`.

## Checklist

---

## Second layer in this PR: quota-aware key selection

`apiKeyPoolStrategy` gains a third value, `quota`, so an API key pool can finally do what every other pool here already does — prefer the credential with the most room left.

**The obstacle was timing, not ranking.** Per-key quota already exists for a long list of providers, but it is reachable only through an `async` reader that probes the network on a cache miss. The selector is synchronous and sits on the first-attempt path, where it must not await anything. So `quota-key-accounts.ts` grows one cache-only reader that never probes, never awaits, and never schedules a read.

**A cache hit is not automatically evidence.** The cache keeps a *last-good* quota attached for up to thirty minutes after a probe starts failing. Returning it would rank on a half-hour-old number — and rank it above a key with no row at all. `cachedApiKeyQuota` returns null whenever the row is marked unavailable. Last-good is a display value, not a selection input.

Ranking matches `headroomOf` on the OAuth side so the two pools cannot disagree about what "more room" means, and mixed evidence uses the same three buckets as `rankAccountsByHeadroom`: measured-with-headroom, then unmeasured, then measured-and-spent. An unmeasured key is not assumed spent, and not assumed fresh either. A provider that reports every key at the same percent ties across the board and falls through to the roster order — exactly today's behaviour.

The new branch is an `else if` placed after round-robin and **before** the `eligible[0]` default, because that default *is* fill-first; replacing it would have silently retargeted a shipped strategy.

`apiKeyPoolStrategy` also had no docs-site row at all. It has one now, covering all three values.

### Added verification

- `bun test tests/adapters/key-failover.test.ts tests/server/server-key-failover-e2e.test.ts tests/providers/provider-quota.test.ts tests/lab/core-lab-boundary.test.ts` — 222 pass, 0 fail
- Red control: with the `quota` branch removed, 3 of the 4 new quota cases fail. The fourth is the no-evidence fallback and is supposed to stay green.

---

## Third layer: pin the legacy contracts, then collapse the duplicate validator

Before any of the surfaces below moved, the three pool endpoints that already shipped got golden tests that pin their exact request and response shapes — `/api/codex-auth/auto-switch`, `/api/codex-auth/pool-strategy`, and `/api/oauth/accounts/pool`. They are not there for coverage. They are the only thing that can tell a consolidation apart from a silent behaviour change, and a later layer in this same PR retires all three from the GUI.

With the contracts pinned, the strategy/sticky validator that existed twice — once per namespace, already drifting on the sticky upper bound — became one function. The collapse is safe precisely because the goldens fail if either namespace starts answering differently.

## Fourth layer: a bound thread stays on its account until that account is spent

Behind `pool.cacheAffinity`, an account that already holds a thread's prompt cache outranks a higher-quota account. Rotating off it throws away the cache and pays full prefill on the next turn, which is the opposite of what a quota-ranked pool is trying to save.

The ordering is affinity **above quota** but **below exhaustion**: a spent or paused account loses regardless of affinity, so this can never pin a thread to a credential that cannot serve it. That boundary is the whole design, and the audit resized the change surface twice before it was right — the first draft treated the usage threshold as the exhaustion bar, which would have let affinity hold a thread on an account the pool had already decided to leave.

## Fifth layer: the two first-send paths #4277 could not reach

The first layer of this PR deliberately left two paths uncovered, and they are covered here rather than assumed:

- **Native compact** for `openai-apikey` never enters `core.ts`, so it needs its own pick.
- **The keyed `/v1/images` path** reads `candidates.keyed.apiKey` — a *snapshot*, not the live provider object. Assigning to `route.provider` there would have type-checked, passed a naive test, and shipped a no-op. The pick has to be placed where the snapshot is taken.

That second one is the same trap the audit caught in the first layer, in a third disguise: a call whose return value nothing reads.

## Sixth layer: one pool-settings contract for all three kinds

`GET | PUT | PATCH /api/pool/settings` replaces the three namespace-specific endpoints. Every kind — Codex OAuth, generic OAuth, API keys — answers the same shape, and `provider` selects which one.

The response also gains `enabledEffective`, which fixes a real defect rather than adding a field: a pool could be *inherited* on by a parent setting while its own `enabled` was absent, and every caller that read `enabled` directly concluded it was off. The new field is computed, and it stays **off the legacy DTOs** — the goldens from the third layer are what proved that, and the audit refused an earlier draft that added it to the shipped shapes.

## Seventh layer: one GUI client for every pool kind

`gui/src/pool-settings.ts` is now the single client for the unified route, and it owns the request mapping that the old per-namespace callers each re-derived: `threshold` → `autoSwitchThreshold`, `provider` always sent, Codex as `"openai"`. `codex-auto-switch.ts` and `AnthropicAccountPoolSettings.tsx` both delegate to it, and no file under `gui/src` references any of the three retired endpoints anymore. `/api/codex-auth/active` stays, deliberately — it is a mixed pin-plus-pool read, not a pool settings write.

The mapping is the load-bearing part. The unified route ignores an unknown `threshold` field and answers 200 without writing, so a client that forgot to rename it would look healthy and change nothing. Reverting that one rename fails six auto-switch tests, which is how it is known to be load-bearing rather than decorative.

`putCodexPoolStrategy` moved out of `account-pool-strategy.ts` and in beside the client, because keeping it where it was created an import cycle the bundler flagged as `INEFFECTIVE_DYNAMIC_IMPORT`. `account-pool-strategy.ts` is a pure value module again.

### Rendered GUI proof

Codex Set → Multi-auth, rendered from an isolated `OPENCODEX_HOME` with the Codex pool strategy set to round-robin and a sticky count of 3. Both controls are served and written through the new client:

Codex Set rotation strategy and sticky assignment count, served by the unified pool settings client

## Added verification for the layers above

- `bun test tests/server/account-pool-management-api.test.ts tests/cli/cli-account-pool-verbs.test.ts tests/oauth tests/lab/core-lab-boundary.test.ts` — pass after merging current `dev`
- `cd gui && bun test tests` — 1954 pass, 0 fail
- `bun run lint:gui` — clean
- `bun run build:gui` — clean
- `bun x tsc --noEmit` — pass
- Acceptance grep: no file under `gui/src` matches `/api/codex-auth/auto-switch`, `/api/codex-auth/pool-strategy`, or `/api/oauth/accounts/pool`
- Red controls: reverting the `threshold` → `autoSwitchThreshold` rename fails 6 auto-switch tests; removing the cache-affinity comparator fails the affinity ordering cases; removing the images-path pick fails the keyed image first-send test

Work-phase records for each layer live under `devlog/_plan/260911_account_pool_unification/`.

## Summary by CodeRabbit

- **New Features**
  - Added unified pool-settings management for Codex, Anthropic, and generic OAuth providers.
  - Added proactive API-key failover for chat, responses, compact requests, and image generation.
  - Added quota-aware API-key selection and the `quota` strategy.
  - Added optional Codex cache affinity to keep bound tasks on an account until exhaustion.
  - Added CLI and GUI support for unified pool settings and auto-switch controls.

- **Documentation**
  - Updated management API, provider configuration, and CLI references.
  - Documented the unified endpoint and legacy endpoint compatibility.


lidge-jun:
## 리뷰 · 우선순위 72 / 80

이 PR은 이미 `dev`에 들어간 #4277의 `selectProactiveApiKey` / `forgetApiKeyRotationCursor`를 **실제로 호출하는 자리**에 꽂습니다. #4277은 피커와 커서 지우기 함수만 만들고, 프로덕션에서 부르는 곳이 없었습니다. 그래서 API 키 풀은 429가 난 뒤에야 돌았고, 이미 식은(쿨다운) 키가 `apiKey`로 다시 잡혀 있는 상태(재시작·수동 편집·설정 리로드 뒤)에서는 첫 요청을 거절당하는 데 써 버렸습니다. 이 PR은 그 첫 시도를 **아직 따뜻한 키**로 보내게 만듭니다.

지금 `dev`(HEAD `29d632ff2`)는 계정 풀 통일의 앞 단계가 이미 올라와 있습니다. #4279 OAuth 풀 커널, #4284 수동 계정 선택이 풀 커서보다 우선, 그리고 #4277 키 풀 전략 피커입니다. 이 PR은 그중에서 **키 쪽 phase 4 배선(wp4b)** 입니다. 설계 기록은 `devlog/_plan/260911_account_pool_unification/040_phase4_key_pool_strategy.md`에 있고, 감사에서 한 번 FAIL(반환값을 안 넣고 호출만 하면 no-op인데 config는 씀) 난 뒤 고친 흔적도 본문에 적혀 있습니다.

호출 위치 설명이 핵심입니다. Responses 경로에서는 `src/server/responses/core.ts`에서 OAuth 선호 계정 고르기 바로 다음, `resolveProviderTransport` 핀 **앞**에 `selectProactiveApiKey` 결과를 `route.provider`에 넣습니다. 이미지 브리지·웹 검색·`runTurn`·일반 HTTP가 같은 객체를 읽기 때문에, 핀 앞에 한 번만 넣으면 네 길이 같이 따라갑니다. 네이티브 채팅은 Responses를 안 타므로 `src/server/chat-native.ts`에서 `activeProvider`를 묶기 직전에 같은 호출을 한 번 더 합니다. `core.ts`는 원래 `hasKeyPoolFailover`를 같은 모듈에서 가져오므로 Lab 경계 import가 새로 생기지 않습니다.

커서 쪽은 수동으로 키를 고치거나 지울 때 `clearKeyCooldowns(name)` 옆에 `forgetApiKeyRotationCursor(name)`을 붙입니다. 예전 라운드로빈 커서가 남으면, 운영자가 방금 고른 키를 다음 선제 선택이 다시 밀어낼 수 있기 때문입니다. `oauth-account-routes.ts` 세 곳과 `provider-routes.ts` 이름 있는 한 곳까지는 연결됐습니다. 전략이 없고 커밋된 키가 식지 않았으면 피커는 `null`이라 일반 설치는 조건 한 번만 보고 끝납니다. types/config 분할 캠페인과 겹치지 않는 좁은 PR입니다.

베이스는 `dev`가 아니라 `codex/generic-pool-kernel`(열린 #4289)입니다. merge-base는 지금 `dev` tip과 같고, 배선 커밋 네 개는 #4289 위에 쌓여 있습니다. 코드 자체(#4277 피커 배선)는 OAuth generic kernel과 직접 의존이 약해 보여서, `dev`로 다시 겨냥해 병렬로 넣을지 #4289 뒤에 태울지가 운영 판단입니다.

tests/server/server-key-failover-e2e.test.ts - 새 e2e 두 개는 `describe`가 닫힌 **밖**에 있습니다. 모듈 단위 `beforeEach`/`afterEach`와 `upstream` 덕분에 실행은 되지만, 파일만 보면 describe 밖 orphan처럼 보여 나중에 고칠 때 헷갈립니다.

tests/server/server-key-failover-e2e.test.ts / `/v1/chat/completions` - 어댑터가 `openai-chat`이라 `isNativeChatRouteEligible` → `chat-native.ts` 경로만 탑니다. `core.ts` 핀 앞 대입은 이 테스트로 증명되지 않습니다. 본문의 “양쪽 call site를 지우면 red” 주장은 chat-native만 지워도 깨질 수 있고, Responses/이미지/웹검색 경로는 여전히 빈칸입니다.

src/server/management/provider-routes.ts:931 - 플랜이 적어 둔 “다섯 경로” 중 `clearKeyCooldowns()`(이름 없이 전부 비우기)에는 `forgetApiKeyRotationCursor`가 없습니다. 이름 단위 delete만 있는 API라 전역 짝이 애매하지만, 쿨다운만 지우고 커서는 남는 비대칭이 생깁니다.

PR 본문 / 미포함 경로 - 본문이 밝힌 대로 `openai-apikey` native compact와 keyed `/v1/images`는 이번 범위 밖입니다. 의도된 후속이지만, 키 풀을 쓰는 그 경로들은 첫 시도 선제가 아직 없습니다.

메인테이너의 판단이 필요한 지점
- 베이스를 #4289(`codex/generic-pool-kernel`)에 둔 채 스택으로 갈지, 배선만 `dev`로 retarget/cherry-pick 해서 #4289와 병렬로 넣을지.
- `core.ts` 선제 선택을 Responses(또는 이미지/웹검색) e2e로 한 줄 더 증명할지, chat-native 커버만으로 머지할지.
- `provider-routes.ts` 전역 `clearKeyCooldowns()`에 커서 전체 초기화(헬퍼 추가)를 같은 PR에 넣을지, 후속으로 미룰지.
- Cross-platform CI가 아직 도는 중(UNSTABLE). 초록 확인 후 넣을지.

너의 추천
CI가 초록이면 넣는 쪽이 맞습니다. 가능하면 베이스를 `dev`로 맞추거나 #4289 머지 직후 바로 이어서 넣고, `core.ts` 경로 smoke 테스트 한 개는 follow-up으로라도 남기세요. types/config 분할과 무관하니 닫지 마세요. 전역 cooldown clear(:931) 커서 비대칭은 짧게 이슈/후속 커밋으로 기록하면 충분합니다.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
All six findings were checked against the code. Five were real and are fixed; one had the right class and the wrong mechanism, and that distinction changed what the regression test could honestly claim.

**Rebuild the routed provider after selecting a key (P1) — fixed, mechanism corrected.**
The class is real and this module already documented it: `rotateKeyAfterFailure`'s docblock says its return "carries none of the registry backfills `routedProviderConfig` merges in at request time" and that "request paths must not assign it to an active route wholesale". `selectProactiveApiKey` returns exactly such a snapshot and all four first-send sites were assigning it wholesale.

The named crash is not reachable, though: `adapter` and `baseUrl` are required on a stored provider row, so a built-in provider cannot be persisted without them and `resolveAdapter()` cannot see `undefined`. The backfill that actually goes missing is the credential — a stored `\${VAR}` or keychain reference is resolved in `routedProviderConfig` and nowhere in the adapter, so a wholesale assignment sends the literal reference as the bearer token.

Fixed by reusing the seam the 429 path already uses rather than patching each call site: `selectProactiveApiKeyTransport` wraps the picker in `applyRotatedTransport`, which reapplies registry metadata and retains only explicit runtime transport state (`fetch`, generated OpenCode session affinity). All four sites — Responses core, native chat, native compact, keyed images — now call it.

**Exercise the Responses-core selection path (P1) — fixed.**
Correct, the file only sent /v1/chat/completions. There is now a /v1/responses case whose red control works: remove the pick from core.ts and the upstream sees the cooled key.

It does not claim to prove the snapshot-versus-rebuild distinction, because on that path it cannot: `refreshDispatchAdapter` re-derives the transport from config before dispatch, so the Responses core self-heals a wholesale assignment. That contract is pinned as a unit in `tests/adapters/key-failover.test.ts`, where the red control does fail — returning the snapshot hands back `\${OCX_KEYFAILOVER_WARM}` instead of the resolved key.

**Expire cached quota before ranking keys (P2) — fixed.**
Confirmed: `cachedApiKeyQuota` rejected `unavailable` and empty rows but checked neither `entry.ts` nor `quota.updatedAt`, while the async `readEntry` it shadows applies both. Nothing on the selection path probes or sweeps, so a successful row could outlive `ACCOUNT_QUOTA_TTL_MS` and keep a roomy ten-minute-old measurement ranked above a key with no evidence until an unrelated write swept it. It now applies `readEntry`'s freshness predicate inverted, and still never probes. Red control returns the stale roomier key.

**Reset every rotation cursor after batch edits (P2) — fixed, with one narrowing.**
Confirmed that the batch `PUT /api/providers` called `clearKeyCooldowns()` with no name and no cursor reset at all, and that there was no all-provider helper. `forgetApiKeyRotationCursor` now takes an optional name, mirroring `clearKeyCooldowns`, and the PUT calls it with none.

The narrowing: this endpoint rejects `apiKeyPool` as editor write authority, so the reorder-and-change-active scenario is not reachable through it. What is reachable is the asymmetry itself — the PUT clears every cooldown and left a cursor behind, so the next pick resumed after the pre-edit position rather than the roster head. The regression test pins that, and its red control returns `sk-alpha-three` instead of the first eligible key.

**Synchronize the owned server architecture docs (P1) — fixed.**
`structure/transports/responses.md` now has a "Pre-dispatch API-key pool pick" section, with the placement argument and the Transport-versus-snapshot rule; `inbound-compat.md`, `images.md` and `transports/inventory.md` carry the pointers for the paths that repeat the call. `bun scripts/structure-ssot.ts` passes.

**Document the cache-affinity configuration switch (P1) — fixed.**
`pool.cacheAffinity` now has a row in `docs-site/src/content/docs/reference/configuration/providers.md` with its default and the threshold-versus-exhaustion boundary, and the unqualified `autoSwitchThreshold` / `accountPoolStrategy` claims about a bound task moving are corrected. The locales carrying those rows (ko, ja, zh-cn, zh-tw, fr, ru, tr) were updated to match; there is no `de` locale for this page.

lidge-jun:
Maintainer integration record (AGENTS.md branch policy / MAINTAINERS.md `dev` self-integration).

- Exact head under CI: `c67e596a7`
- Every required check passed at that head: `gates`, `test 1/4`-`4/4`, `macos 1/2`-`2/2`, `npm-global` on all three OSes, `keyring` on all three OSes, `docker smoke`, `storage policy`, `api usage`, `hygiene`, `enforce-target`, `label`, `react-doctor`, `changes`. `macos control` and the Windows shard matrix were skipped by their own conditions.
- `enforce-target` green with the rendered GUI screenshot embedded in the description.
- All eight review findings — six from Codex, two from CodeRabbit — were verified against the code, fixed, and their threads resolved. Three were caught only because CI or a red control disproved the first attempt: the headless-parity sweep had no row for the unified route, the surface map drifted against the PR merge commit after #4306, and the first regression test for the rebuild contract passed without the fix and had to be replaced with one that fails.
- No outstanding maintainer objection.
- Security review surface considered: this touches API-key selection and therefore credential handling. Nothing is logged, serialized, or newly exposed — the pre-dispatch pick moves an already-persisted credential onto the route through the same seam the 429 path uses, the unresolvable-credential case now fails closed rather than falling back to a stale key, and `privacy:scan` is green in `gates`. No OAuth flow, workflow permission, release script, or dependency install is modified.

Integrating through this pull request with a merge commit.

PR #4453, opened 2026-09-13
feat(devin): request prompt caching and give the session cache an identity

lidge-jun:
## Summary

We were reusing the Devin session id specifically so the server-side prompt cache would hit, and then never asking for a cache entry. The request encoder emitted fields 1, 2, 3, 7, 8, 10, 15, 16, 20 and 21 and omitted **field 13, `PromptCacheOptions`**, so the server created nothing to reuse and re-read the whole prefix every turn.

Send it, unconditionally. The bytes are `6a 02 08 01` — tag `(13<<3)|2`, length 2, inner field 1 varint 1 — verified against both implementations that already do this.

The comparison is the point: neither reference gets the whole thing.

| | session reuse | asks for a cache entry |
|---|---|---|
| native Devin CLI | yes | yes |
| CLIProxyAPIPlus | **no**, new session per request | yes |
| opencodex before | yes | **no** |
| opencodex after | yes | yes |

CLIProxyAPIPlus asks for a cache it then cannot hit, because it mints a fresh session and cascade on every request. We kept the session and never asked. Both halves now hold.

### Cache identity

The session cache was keyed by the raw credential, which puts the token into a `Map` key that any heap dump or debugger walks, and leaves the other per-account caches without a shared name for an account. It is now keyed by a sha256 identity.

That identity is what makes invalidation safe. `clearSessionIds` was a global `Map.clear()`, and this proxy serves several accounts from one process — a per-provider logout calling it would have stripped the session and cascade of accounts that were mid-turn. It was exported with **zero callers** precisely because there was no safe place to call it. `invalidateSessionIdentity` drops one account.

No epoch is introduced. A turn in flight received its `SessionIds` object at request start and never re-reads the map, so deleting an entry cannot disturb it.

## Verification

- Hosted CI on this exact head is the merge proof. Local product tests, typecheck, build and install were **NOT RUN** in the authoring session, by explicit instruction.
- New coverage in `tests/providers/devin-prompt-cache.test.ts`: the exact 4 bytes are present, they are present with no tools and with `tools: []`, exactly one cache-options field is emitted, the identity never contains the credential and is stable and account/host separating, a crafted credential cannot forge the host boundary, and the module surface no longer exports the global clear.
- Registered in `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`.
- Reviewed independently before implementation. It walked both encoders byte for byte and confirmed `6a 02 08 01` matches `devinEncodeSubMessage(13, devinEncodeCacheOptions())`; confirmed field order is not a wire correctness condition and that placing 13 between 10 and 15 is caution; confirmed the unconditional send matches Plus, which appends it outside its tools gate; confirmed 16 hex gives ~1.8e-15 collision at the bounded cache size and that nothing parses the map key; and confirmed `clearSessionIds` is not re-exported by the `cloud-direct` barrel, so removing it cannot break the build.
- The reviewer argued the epoch specified in the plan was dead complexity and I accepted that, so it is not in this change.

## Checklist

Maintainer integration under `MAINTAINERS.md`: `dev` only, with exact-head CI evidence recorded before merge.

## Summary by CodeRabbit

- **New Features**
  - Enabled ephemeral prompt caching for Devin conversations, improving continuity across turns.
  - Added secure, account-specific cache identity handling to keep sessions isolated across hosts and credentials.
  - Added targeted session invalidation so signing out one account no longer affects unrelated sessions.

- **Bug Fixes**
  - Corrected prompt-cache request handling for conversations with or without tools.
  - Improved resilience when invalidating unknown session identities.

### DIFF
--- a/scripts/test-layout/layout.json
+++ b/scripts/test-layout/layout.json
@@ -646,6 +646,7 @@
     "devin-cli-authmode-migration.test.ts": "providers",
     "devin-cli-login.test.ts": "providers",
     "devin-hardening.test.ts": "providers",
+    "devin-prompt-cache.test.ts": "providers",
     "devin-stream-deadline.test.ts": "providers",
     "digitalocean-scaleway-provider.test.ts": "providers",
     "docs-429-failover-claims.test.ts": "ci-workflows",
--- a/src/adapters/devin/cloud-direct/chat.ts
+++ b/src/adapters/devin/cloud-direct/chat.ts
@@ -77,16 +77,34 @@ export const cloudStreamHeadersMsForTests = cloudStreamHeadersMs;
 const MAX_FRAME_LEN = 16 * 1024 * 1024;
 
 /**
- * Per-(apiKey, host) session/cascade ID cache. Cloud uses these for
- * server-side context caching across turns of the same conversation; if we
- * mint a fresh sessionId on every call (which we used to), every turn looks
- * like a brand-new session and the prompt-cache hit ratio is zero.
+ * PromptCacheOptions.type = EPHEMERAL. Marks the system prefix as a cache entry
+ * the server may reuse on the next turn of the same session.
+ */
+const PROMPT_CACHE_EPHEMERAL = 1;
+
+/**
+ * Per-identity session/cascade ID cache. Cloud uses these for server-side
+ * context caching across turns of the same conversation; if we mint a fresh
+ * sessionId on every call (which we used to), every turn looks like a
+ * brand-new session and the prompt-cache hit ratio is zero.
  * Single-process scope is enough: opencode lives in one runtime for a TUI
  * session, and CLI one-shots don't benefit from caching anyway.
  */
 interface SessionIds {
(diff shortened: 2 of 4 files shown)


lidge-jun:
**Maintainer integration into `dev`** under [`MAINTAINERS.md`](../blob/dev/MAINTAINERS.md).

- Exact head merged: `8831c86824fb02c65c230578072b4b440390e0f4`.
- Hosted CI on that exact head: **25 successful checks, 0 failed, 0 cancelled**, covering the 4 Linux test shards plus the macOS and Windows suites, gates, keyring, npm-global, docker smoke, storage policy and api usage.
- Skipped jobs are not counted as passing.
- Local product suite, typecheck, build and install were **NOT RUN** in the authoring session by explicit instruction; hosted exact-head CI is the sole merge proof.
- Integrated without a second maintainer approval, which the policy permits for `dev` only. This is maintainer integration, not a self-approval.

PR #4477, opened 2026-09-13
fix(providers): let field-masked writes reach canonical OpenAI past stored overlays

lidge-jun:
## Summary

Carry of #4447 (Veritas-7, `ed9655286`) onto `dev`. The source pull request targets `main` and is left alone.

Lane S is held for maintainer security review. It must not land as a side effect of another lane's tip merge.

### Overlay precedence, before and after

**Before.** `providerManagementConfigError` guarded canonical OpenAI with `sameCanonicalProviderSeed`: an exact key-set comparison between the submitted provider and the built-in registry seed (`adapter`, `baseUrl`, `authMode`, `codexAccountMode`, and the other keys `providerConfigSeed` actually emits). That strictness is right for full-object writes (`POST /api/providers`), where an overlay riding on a canonical seed must still be rejected.

Merge-based writes — `PATCH /api/providers`, the provider editor PUT, and reload — validate a candidate that was already merged onto the persisted row. Once an operator overlay lands on disk, every later candidate carries that extra key and fails the exact-key check.

The overlay that actually bricks the row in production is `selectedModels`, written by `PUT /api/selected-models` from the Models page. After that write, **every** field-masked write to built-in OpenAI returned 400 `provider openai must equal the canonical built-in provider seed`: `modelContextWindows`, `disabled`, `defaultModel`, and so on. The overlays themselves were admitted by their own write boundaries and cannot widen what the forward proxy claims. Seed keys (`baseUrl`, `adapter`, `authMode`) were already rejected by value mismatch; the bug was that *unrelated* overlays made those later writes unreachable.

**After.** Merge-based paths pass `{ allowOperatorOverlays: true }`. `matchesCanonicalProviderSeed` still requires every key the seed **defines** to be present and byte-equal; keys the seed never defines are ignored. `POST` keeps the exact-key comparison.

### What a field-masked write can now reach

On a canonical OpenAI row that already carries `selectedModels` (or `disabled`, or another non-seed overlay):

- `PATCH { "modelContextWindows": { "gpt-6-astra": 872000 } }` now returns 200 and persists the window. The stored `selectedModels` survive.
- Seed-key tampering still returns 400. `PATCH { "baseUrl": "https://attacker.example.com/v1" }` does not persist.

### What it still must not reach

`allowPrivateNetwork` is the explicit destination-policy opt-in that makes `providerDestinationResolvedError` skip DNS classification (loopback, RFC1918, metadata). The canonical ChatGPT forward seed does not define that key, so overlay-tolerant comparison would otherwise persist it on `openai` and later destination probes would honor it.

Canonical `openai` now rejects the field outright (`provider openai must not include allowPrivateNetwork`), including when the row already carries `selectedModels`. CodeRabbit's proposed fix on #4447 was to `delete` the field from the *comparison candidate*; that would have allowed persistence. This carry rejects it instead. Disabled-row recovery still strips a stale flag on the re-enable path and still ignores it for the DNS probe; that path does not go through overlay-tolerant seed comparison.

`autoReviewModel` / `autoReviewModelOverrides` and runtime fields stay rejected.

## Verification

- `bun test tests/server/management-provider-validation.test.ts` — 127 pass, including:
  - canonical OpenAI with `selectedModels` can still PATCH `modelContextWindows`
  - the same row still rejects `baseUrl` tampering
  - the same row still rejects `allowPrivateNetwork` and does not persist it
- `bun run typecheck` — clean
- `bun run structure:check` — clean
- Local full product suite, build and install: **NOT RUN**. Hosted exact-head CI on this tip is the suite proof.

## Checklist

This PR is the security-review hold for the CORS / management-route change. Do not merge until a maintainer security review lands, or record it deferred with that reason.

Co-authored-by: Veritas-7 <234569343+Veritas-7@users.noreply.github.com>

## Summary by CodeRabbit

- **Bug Fixes**
  - Provider editor updates and reloads now preserve saved model selections and other operator settings.
  - Valid provider setting changes are no longer rejected when saved overlay fields are present.
  - Unauthorized changes to provider connection details—including destination URLs, network access settings, and headers—continue to be rejected.

- **Documentation**
  - Clarified how field-masked updates handle provider defaults and operator-managed settings.


lidge-jun:
## 리뷰 · 우선순위 74 / 80

설명
이 PR은 Veritas-7의 #4447(`main` 타깃 WRONG BRANCH)을 `dev`로 캐리한 것이다. 원본 PR은 건드리지 않는다. 본문이 말한 Lane S(보안 리뷰 레인)도 다른 레인 tip 머지의 부작용으로 넣지 말라는 제약이 있다.

문제: canonical OpenAI 행에 `selectedModels` 같은 운영자 오버레이가 디스크에 있으면, `sameCanonicalProviderSeed`의 exact-key 비교가 “시드와 키가 다르다”고 보고 PATCH(예: modelContextWindows)를 전부 거절했다. 필드 마스크 쓰기는 저장된 오버레이를 후보에 실을 수밖에 없는데, 시드 검사까지 그 키를 transport 변조로 본 것이다.

수정은 `src/server/auth-cors.ts`에 `matchesCanonicalProviderSeed`(시드가 정의한 키만 일치, 시드 밖 키는 무시)를 두고, `providerManagementConfigError(..., { allowOperatorOverlays: true })`일 때 그걸 쓴다. POST(전체 객체)는 예전 exact-key를 유지해 위조 오버레이가 transport 시드에 못 타게 한다. canonical openai의 `allowPrivateNetwork`는 오버레이 관용 비교에서도 명시 거절한다(DNS 분류 우회 방지). `provider-routes.ts`의 editor/reload/PATCH/replay 경로만 `allowOperatorOverlays: true`를 켠다. 테스트는 selectedModels가 있어도 PATCH가 되고, transport 변조는 여전히 거절되는지 잠근다.

우선순위 74는 대시보드에서 canonical OpenAI를 사실상 못 고치던 운영 막힘을 풀고, POST vs PATCH 경계를 보안 관점에서 나눴기 때문이다. tip `b4dec3469` / `2.53.0`, types/config 분할 무관. Lane S 별도 리뷰 요구는 메인테이너 판단으로 남긴다.

`src/server/auth-cors.ts` `matchesCanonicalProviderSeed` / `allowOperatorOverlays` / `allowPrivateNetwork` 거절 - 오버레이 관용과 forward 안전장치의 핵심.

`src/server/management/provider-routes.ts` editor·reload·PATCH·replay - 관용 플래그 켜는 호출부. POST는 그대로 엄격.

`tests/server/management-provider-validation.test.ts` - selectedModels + PATCH 성공, transport tamper 실패.

심볼 원본 #4447(main) - WRONG BRANCH는 그대로 두고 이 캐리만 `dev`에 넣는다. Lane S 메모를 무시하고 다른 레인에 섞지 말 것.

메인테이너의 판단이 필요한 지점
- Lane S 보안 사인오프를 이 PR에 요구할지, 테스트로 충분한지.
- #4447(main)을 닫거나 retarget할지, 아니면 문서만 “superseded by #4477”로 남길지.

너의 추천
관리 API 테스트 초록이면 `dev`에 머지하세요. #4447(main)에는 superseded 코멘트만 남기고 브랜치 타깃은 억지로 바꾸지 마세요. Lane S가 별도 게이트면 머지 전 체크리스트에만 표시하세요.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
## Maintainer security review

This is the security review this lane was held for. It was done against the diff and the surrounding code rather than the description, and it changed the outcome, so the reasoning is recorded here.

**Verdict: approved with one fix, now applied in c39098ba3d.**

### What the review established

The overlay-tolerant comparison is reached only when `name === "openai"`; other seeded providers never enter that branch, so the suspected cross-provider asymmetry does not exist. They could already accept `allowPrivateNetwork` through PATCH, and this change opens nothing new for them.

The live `providerConfigSeed` for openai defines exactly four keys — `adapter`, `authMode`, `baseUrl`, `codexAccountMode`. That is a much smaller seed than the description implies, so "keys the seed never defines are ignored" means nearly every config key. Destination and auth class stay byte-pinned, and `applyProviderPatchFields` is an explicit per-field allowlist, so a request cannot introduce an arbitrary new key — the extra keys in a candidate come from the persisted row. `apiKey` is separately rejected at the PATCH boundary.

The `allowPrivateNetwork` deny is correct and necessary: it is patchable at `provider-routes.ts:409`, it disables destination DNS classification, and under overlay tolerance it would have persisted on the ChatGPT forward row and been honored by the later destination probe. Denying it also costs nothing, because the exact-key comparison already rejected it.

### The gap that had to be fixed first

`headers` is in the same class and was not covered. Canonical OpenAI has no registry `staticHeaders`, the PATCH field mask writes `headers` with a shallow merge at `provider-routes.ts:623`, and the forward adapter applies `provider.headers` at `openai-responses.ts:2327` and `:2365` **before** the incoming forward headers, so a persisted value wins whenever the caller omits that header.

Concretely, before c39098ba3d a dashboard-session `PATCH {"headers":{"chatgpt-account-id":"..."}}` would persist on the canonical row and ride every subsequent request to `https://chatgpt.com/backend-api/codex/responses` that did not carry the header itself. POST still refused it through the strict comparison; PATCH, the provider editor and reload did not. Existing tests covered POST headers and PATCH `baseUrl`/`allowPrivateNetwork`, not PATCH headers.

c39098ba3d denies `headers` on canonical openai the same way, and adds the missing PATCH regression. The test was driven red before it was accepted: removing the guard fails exactly that case and nothing else in the file (127 pass / 1 fail without it, 128 pass / 0 fail with it). `bun run typecheck`, `bun run structure:check` and `bun run privacy:scan` are clean. No local full suite was run; hosted exact-head CI on the new tip is the suite proof.

### Recorded, not blocking

The overlay tolerance is a denylist, and a denylist rots. `PROVIDER_CONFIG_FIELD_POLICY` is exhaustive and forces a new provider field to be classified, but it does not force an overlay decision, so a future `editor` field that touches a trust boundary would become silently reachable on the canonical row. `codexToolMode` is the current example: editor-writable, no openai deny, and it feeds catalog tool exposure. The durable fix is an explicit allowlist of known-safe overlays plus a guard test that fails when a new editor field is neither allowlisted nor denied. That is a follow-up on this repository, not a defect in this carry.

Original work by @Veritas-7 in #4447; the trailer is in the landed commit.

PR #4476, opened 2026-09-13
fix(cli): bind hub status credentials to connection snapshots

lidge-jun:
## Summary

Carries #4382 by @luvs01.

`ocx status` on a connected client read the connection snapshot and the current `service-api-token` independently. `collectRemoteHubStatus` awaits a dynamic import before it reads credentials, so a reconnect or a key rotation inside that window could send the new token to the snapshot's hub, or send the snapshot's hub a token that no longer belongs to it.

The collector now rereads the persisted connection and passes a token only when that connection still matches the snapshot's `serverUrl`, `apiKeyId` and `connectedAt`, and the token fingerprint matches that connection's `tokenFingerprint`. Otherwise it skips the live request and falls back to the snapshot owner's cached hub state, or reports `unavailable`.

Beyond the source branch, a withheld token now names its own cause. `resolveHubState` reported every null token as "this client has no usable data-plane token". That is false for a client that reconnected and holds a perfectly good token for a different hub, and it sends the operator to re-enroll a credential that is not the problem. The caller supplies the cause through the new `withheldTokenReason`, so a changed connection, a missing token file and a fingerprint mismatch read differently:

```
before: State unavailable - this client has no usable data-plane token
after:  State unavailable - the saved connection no longer matches the one this status reports
```

That folds the maintainer review finding on #4382 ("reason 문자열을 한 번 읽어 보면 좋습니다"), and it is the same misdiagnosis class as #4169 in the stop path.

This is the bottom link of lane L in `devlog/_plan/260913_contributor_carry_train/`. Its head commit carries `[skip ci]`; the lane tip runs the matrix for the whole lane.

## Verification

- `bun test tests/cli/cli-status-hub-state.test.ts` — 20 pass
- `bun test tests/clients/client-hub-state.test.ts tests/server/v1-hub-state.test.ts` — 35 pass
- `bun run typecheck` — clean
- `bun run structure:check` — passed
- `bun run privacy:scan` — passed
- `cd docs-site && bun install --frozen-lockfile && bun run build` — 441 pages built. This is the one actionable CodeRabbit finding on the source pull request, and `docs-site/AGENTS.md` requires it for `docs-site/**` changes.

Local full suite was not run, per the lane's instruction. Hosted CI on the lane tip is the suite proof.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults.

[...]

PR #4481, opened 2026-09-13
feat(codex): pull an authenticated remote catalog into local Codex state

### LINKED ISSUE #3729: feat(codex): pull an authenticated remote catalog into local Codex state
## Area

Catalog / models

I need to keep a Codex CLI or Desktop client synchronized with the complete model catalog generated by a remotely hosted OpenCodex instance.

OpenCodex already exposes the persisted catalog through the authenticated `GET|HEAD /v1/catalog` data-plane endpoint. The missing workflow is safely materializing that snapshot into the remote client's local `CODEX_HOME`, rebuilding `models_cache.json`, and reporting when a running Codex app-server or Desktop process still holds the previous catalog in memory.

This would make centrally hosted OpenCodex deployments usable without maintaining deployment-specific web roots, cron jobs, PowerShell downloaders, or ad hoc cache-rebuild scripts.

## What prevents this today?

`model_catalog_json` accepts a local filesystem path, not a URL. `ocx sync` builds a catalog from locally configured providers, while `ocx export --client` generates client-specific configuration; neither consumes a complete catalog from another OpenCodex server.

Operators currently need to implement the entire client-side lifecycle themselves:

1. authenticate to `/v1/catalog`;
2. download into a same-directory temporary file;
3. validate JSON structure and model slug uniqueness;
4. preserve the last-known-good catalog on network, HTTP, or validation failure;
5. atomically replace the local catalog;
6. rebuild `$CODEX_HOME/models_cache.json`;
7. detect stale Codex app-server/Desktop processes;
8. schedule the workflow separately on Windows, macOS, and Linux.

This duplicates security-sensitive filesystem and credential handling that OpenCodex already implements elsewhere and makes reliable Desktop synchronization unnecessarily difficult for non-technical client users.

## What should OpenCodex do?

Provide an explicit remote-catalog pull workflow that consumes the existing `/v1/catalog` contract and installs the result as a local Codex catalog.

Observable requirements:

- accept an HTTPS OpenCodex catalog URL;
- read authentication from an environment reference or existing secure credential mechanism, never a plaintext CLI argument;
- support conditional requests through `ETag` / `If-None-Match`, treating `304` as a no-op;
- reject URL-embedded credentials and avoid forwarding authorization across origins;
- enforce response-size and timeout bounds;
- validate the complete catalog before the first filesystem mutation, including a non-empty `models` array and unique, valid slugs;
- write through the existing catalog serialization and atomic replacement primitives;
- preserve the last-known-good catalog and cache on every download or validation failure;
- update `models_cache.json` in the same coordinated operation;
- preserve mtimes and avoid process handling when the pulled catalog is unchanged;
- report stale Codex processes after a real update;
- keep `--restart-codex` and Windows `--restart-desktop-app` explicit and opt-in, because either can interrupt active work;
- never mutate server configuration, provider credentials, or the source catalog.

The first implementation can stay intentionally narrow: one-shot pull plus cache update. A managed cross-platform scheduled integration can be discussed separately after the command contract is stable.

## Example usage or interface

```bash
export OPENCODEX_CATALOG_AUTH_TOKEN='...'

ocx catalog pull \
  https://proxy.example.com/v1/catalog \
  --auth-env OPENCODEX_CATALOG_AUTH_TOKEN \
  --restart-codex
```

Machine-readable automation:

```bash
ocx catalog pull \
  https://proxy.example.com/v1/catalog \
  --auth-env OPENCODEX_CATALOG_AUTH_TOKEN \
  --json
```

Expected no-op result after an unchanged conditional request:

```json
{
  "status": "unchanged",
  "catalogWritten": false,
  "cacheSynced": false,
  "codexRestarted": false
}
```

The exact command and environment-variable names are proposed for discussion; the important contract is authenticated remote-to-local materialization with coordinated catalog/cache writes and opt-in process restart.

- **Serve a deployment-specific `/dl` directory:** works, but duplicates the existing authenticated `/v1/catalog` endpoint and makes nginx/web-root permissions part of catalog correctness.
- **Cron plus shell/PowerShell scripts:** currently workable, but each operator must reimplement validation, locking, atomic replacement, cache generation, error handling, and secret hygiene.
- **Point `model_catalog_json` at an HTTPS URL:** incompatible with Codex's local-path contract and startup behavior.
- **Extend `ocx export --client`:** misleading because that command exports client-specific configuration rather than the complete Codex catalog.
- **Server-side watcher or publisher daemon:** does not solve client-side `models_cache.json` or in-memory Desktop staleness and adds another service lifecycle.
- **Automatically restart Codex Desktop:** rejected as a default because it can terminate active conversations; restart must remain explicit.

## Additional context

Related existing work solves adjacent, but different, layers:

- #709 / #772 introduced the management catalog endpoint.
- #809 and the current `/v1/catalog` implementation address least-privilege remote catalog delivery.
- #241 / #1244 cover routed-model catalog preservation in Codex App.
- #2292 / #2382 establish that full Desktop restart is separate, disruptive, and opt-in.
- #3023 / #3054 harden model entitlement freshness for exported/client-visible catalogs.

A production deployment currently uses a validated atomic catalog publisher plus a Windows downloader/cache rebuilder. That proves the workflow is useful, but the proposal deliberately avoids upstreaming deployment-specific paths, nginx configuration, cron syntax, or public unauthenticated `/dl` behavior.

Security boundaries for implementation review:

- no token in argv, output, errors, redirects, or persisted catalog metadata;
- HTTPS by default;
- no cross-origin credential forwarding;
- bounded body acquisition;
- validation before mutation;
- catalog/cache writes only under the existing per-`CODEX_HOME` serialization permit;
- no implicit process termination.

## Checks

lidge-jun:
## Summary

Carries #4413 by @rrmlima. Closes #3729.

Adds `ocx catalog pull <https-url> [--auth-env <NAME>] [--json] [--restart-codex]`, which installs a complete catalog served by another OpenCodex instance's `/v1/catalog` endpoint and then synchronizes `models_cache.json`. `ocx sync` builds a catalog from locally configured providers and `ocx sync-cache` rebuilds the cache from the catalog already on disk; neither consumes a finished catalog from another server, so operators were reimplementing authentication, atomic replacement, cache rebuild and process handling per deployment.

Acquisition is fail-closed before any local write. HTTPS is required except on loopback, URL-embedded credentials, queries, fragments and redirects are refused, the body is bounded by size and inactivity, and slugs and input modalities are validated. The token is read only by environment-variable name, never from argv. Catalog and cache writes share the Codex catalog write lock and the atomic writer.

### Review findings folded in

A cache rebuild that failed **after** the catalog write left a new catalog paired with a stale `models_cache.json`, while the caller was told `catalogWritten: false`. The write permit rolls back SQLite; `replaceActiveCodexCatalog` is an atomic file write that nothing else undid. The pull now restores the previous catalog bytes — or removes the file when the home had none — before throwing `write_failed`. Both cases have a regression test, and I confirmed both fail without the restore.

`--restart-codex` reported success when only some app-servers stopped. `afterCatalogWriteHandleAppServers` returns `failed` and `surviving` without throwing, so `stopped.length > 0` produced `ok: true` and exit 0 while a stale app-server was still serving the previous catalog from memory. A restart now counts only when nothing failed, nothing survived, and every listed process stopped:

```
before: {"ok":true,"status":"updated","codexRestarted":true}   exit 0   # one survivor ignored
after:  {"ok":false,"status":"updated","catalogWritten":true,
         "cacheSynced":true,"codexRestarted":false,"code":"restart_incomplete"}   exit 1
```

The envelope keeps `catalogWritten` and `cacheSynced` true there, because those writes did land; reporting the pull as having written nothing would be a second false report.

The unused `statSync` import is removed.

Documentation now covers the full `--json` envelope, every failure code with its exit status, the host-root `/v1/catalog` path contract, and the two deliberate Phase 1 omissions: no `ETag`/`If-None-Match` conditional request and no Windows `--restart-desktop-app`. Identical bytes remain a complete no-op, so a home whose catalog is correct but whose cache is broken is repaired by `ocx sync-cache` rather than by this command — that limit is now written down instead of implied. The seven localized lifecycle pages carry the command.

This is the tip of lane L in `devlog/_plan/260913_contributor_carry_train/`, stacked on #4476. Its hosted run is the suite proof for both links.

## Verification

- `bun test tests/codex-integration/catalog-remote-pull.test.ts` — 30 pass, including the two new rollback regressions
- Counterfactual: with the restore removed, both new rollback tests fail; with it, all 30 pass
- `bun test tests/ci-workflows/skill-ocx.test.ts tests/test-layout.test.ts tests/test-layout-tooling.test.ts` — 33 pass
- `bun test tests/cli/cli-help.test.ts tests/cli/cli-dispatch.test.ts tests/cli/cli-registry.test.ts` — 74 pass
- `bun run typecheck`, `bun run structure:check`, `bun run privacy:scan`, `bun run skill:surface` + `skill:surface:check` — all clean
- `cd docs-site && bun run build` — 441 pages, required by `docs-site/AGENTS.md`

The layout entries in `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json` came over with the carry and are verified by the two layout guards above. Local full suite was not run, per the lane's instruction.

## Checklist


lidge-jun:
## 리뷰 · 우선순위 69 / 80

설명
이 PR은 @rrmlima의 #4413을 tip(`b4dec3469`, `2.53.0`)으로 캐리하고 #3729를 닫는다. 운영자가 다른 OpenCodex의 완성 카탈로그를 받아 쓰려면 인증·원자적 교체·캐시 재빌드·프로세스 처리를 배포마다 다시 만들고 있었다. `ocx sync`는 로컬 provider로 카탈로그를 만들고, `ocx sync-cache`는 디스크 카탈로그로 캐시만 다시 짠다. 원격 `/v1/catalog`를 소비하는 공식 명령이 없었다.

추가되는 명령은 `ocx catalog pull <https-url> [--auth-env <NAME>] [--json] [--restart-codex]`다. 구현은 `src/codex/catalog/remote.ts`와 CLI wiring(`catalog.ts`/`dispatch`/`help`/`registry`). HTTPS만(루프백 HTTP 예외), URL 내장 자격·쿼리·fragment·리다이렉트·과대 응답·잘못된 카탈로그는 로컬 쓰기 전에 거절. 인증은 `--auth-env`로 환경변수 이름만 받고 argv로 비밀을 받지 않는다. 카탈로그·캐시는 공유 Codex catalog lock 아래 쓰고, 실패 시 last-good 유지, 동일 바이트면 mtime 보존 no-op. `--restart-codex`는 실제 쓰기 뒤에만. docs-site lifecycle 다국어와 `tests/codex-integration/catalog-remote-pull.test.ts`가 계약을 잠근다.

우선순위 69는 멀티 인스턴스 운영 통증을 직접 줄이고, 보안 경계(비밀을 argv에 안 둠·쓰기 전 검증)가 분명하기 때문이다. types/config 분할과 무관하다.

`src/codex/catalog/remote.ts` `pullRemoteCatalog` - 원격 취득·검증·원자 교체·에러 분류.

`src/cli/catalog.ts` / dispatch·help·registry - `catalog pull` UX와 exit 코드(`usage`/`auth_env_missing` 등).

`tests/codex-integration/catalog-remote-pull.test.ts` - HTTPS/거절/auth-env/원자성 회귀.

심볼 #4413 / #3729 - 캐리 머지 시 둘 다 닫히는 관계. ETag·Desktop 재시작은 범위 밖(본문도 명시).

메인테이너의 판단이 필요한 지점
- `/v1/catalog`가 호스트 루트여야 한다는 URL 제약을 프록시 path prefix 환경에서 완화할지.
- #4413 leftover와 #3729 동시 종료 코멘트 형식.

너의 추천
보안·lock 테스트 초록이면 머지하세요. #4413 Landed via #4481, #3729 closes 확인하세요.

이 댓글은 grok-bot이 작성했습니다

PR #4497, opened 2026-09-13
fix(chat): accept Pi and Anthropic-shaped image parts on the chat wire

lidge-jun:
## Summary

- Aside (Pi agent core, openai-completions API) returns read_file screenshots as tool-result parts shaped {type:"image", data, mimeType}. The chat inbound translator (src/chat/inbound.ts) only recognized image_url parts and silently dropped these, so vision models answered that the image data never arrived.
- Live-reproduced against the running proxy (2.53.0 dev dogfood): claude-opus-5 with a Pi-style tool image reports the image missing; the identical request in image_url form is described correctly. Anthropic-messages, chat user-message, and Responses inbound image paths were probed and are unaffected.
- Fix: normalize Pi/MCP-style {type:"image", data, mimeType} and Anthropic source-object ({type:"base64", media_type, data} / {type:"url", url}) parts to the data-URI form the Responses pipeline already handles. This restores screenshot reading for every Pi-adapter client, any model.
- Regression tests added to the chatCompletionsToResponsesBody image-parts table (5 new cases: Pi base64+mimeType, mediaType alias, pre-formed data URI, Anthropic base64 source, Anthropic URL source).

## Verification

- Local suite/build/typecheck: NOT RUN (maintainer rule for this task; hosted exact-head CI is the gate).
- Live probes against the running proxy (user's own instance, tiny 200px screenshot): bug reproduced pre-fix (model: "the image data didn't come through"); standard shapes confirmed working pre-existing.
- Translator-level regression cases added under tests/responses/chat-completions-endpoint.test.ts.

## Checklist

### DIFF
--- a/src/chat/inbound.ts
+++ b/src/chat/inbound.ts
@@ -46,10 +46,35 @@ function contentToText(content: unknown): string {
 }
 
 function imageUrlFromPart(part: Rec): string | null {
-  if (part.type !== "image_url") return null;
-  const imageUrl = part.image_url;
-  if (typeof imageUrl === "string" && imageUrl.length > 0) return imageUrl;
-  if (isRec(imageUrl) && typeof imageUrl.url === "string" && imageUrl.url.length > 0) return imageUrl.url;
+  if (part.type === "image_url") {
+    const imageUrl = part.image_url;
+    if (typeof imageUrl === "string" && imageUrl.length > 0) return imageUrl;
+    if (isRec(imageUrl) && typeof imageUrl.url === "string" && imageUrl.url.length > 0) return imageUrl.url;
+    return null;
+  }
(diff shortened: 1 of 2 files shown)


lidge-jun:
Maintainer integration record (MAINTAINERS.md dev exception): integrating as repository owner without a second maintainer approval. No outstanding maintainer change requests. Exact-head verification: all required checks green on head 9b9675b202 (ci aggregator, gates, hygiene, enforce-target, docker smoke, api usage, CodeRabbit completed; windows/macos-control skipped by changes-gate design). Local suite NOT RUN per task rule; hosted exact-head CI is the gate. User-reported blocker (Aside/Pi screenshots blind on every model) — root cause live-reproduced pre-fix; fix verified at translator level by the added regression cases; end-to-end re-verification against the live proxy follows the merge.

PR #4510, opened 2026-09-13
Make the Codex desktop-app restart cross-platform and fold it into --restart-codex

lidge-jun:
## Summary

`ocx sync --restart-codex` stopped having any observable effect. The matcher was never
the problem: the Codex app-server it signals is a **child of the desktop app** (measured:
pid 16733 under 15901 on macOS, 3285204 under 3284901 on Linux), so the app respawns it
while the renderer keeps the model roster it built at launch. Only restarting the shell
that owns the picker fixes it — which is exactly what the Windows-only
`--restart-desktop-app` already did, and what macOS and Linux had no way to do at all.

This makes that capability a cross-platform shared surface and folds it into
`--restart-codex`.

**Before:** `--restart-codex` sent SIGTERM to app-servers only. A full desktop restart
needed `--restart-desktop-app`, which was Windows-only and never implied.

**After:** `--restart-codex` restarts the app-servers **and** fully quits and relaunches
the Codex desktop app on macOS, Linux and Windows. `--restart-desktop-app` is a
deprecated alias that says so. `--restart-app-server-only` is new and carries the old
narrow behaviour; when flags conflict the **narrow** scope wins, because losing live
conversations is unrecoverable and a stale picker is not. The three flags now mean the
same thing in `sync`, `sync-cache` and `catalog pull`, and `ocx system codex-restart`
reaches the same module.

`src/codex/desktop-app-restart.ts` becomes a platform-independent ladder over three
adapters behind `DesktopAppAdapter`. The fail-closed probing, PID-reuse re-verification,
root selection and ancestry handling are identical everywhere; only identity, discovery,
membership, the two stop primitives and relaunch differ.

### Why #2292's decision is reversed

#2292 deliberately kept the flags separate: quitting the app ends live conversations,
which is a larger consent than restarting a background helper. That reasoning was sound.
It is superseded by an explicit maintainer instruction that `--restart-codex` must mean
the app is fully stopped and started again. Nothing is lost — the narrow scope moved to a
flag that names it, which is a better arrangement than having it be the unnamed default.

### Three things the measurements changed

Root selection now requires the process to be the app **shell**, not merely a member
whose parent sits outside the tree. macOS crashpad handlers run at ppid 1 and stale ones
outlive the instance that spawned them; on a live machine the old rule selected four of
them alongside the real shell, and a survivor would have blocked every relaunch.

The Linux relaunch environment is read from a **child**, because the root zeroes its own
environ block after startup — measured as 1902 NUL bytes — and only children still carry
`XDG_RUNTIME_DIR`. Reading the root would have produced an app that starts and never
reaches the compositor.

The ancestry walk distinguishes a **dead parent** (a clean end of chain, and the normal
state of an orphaned helper on Windows) from a hop it could not read or a bound it hit,
which both fail closed. Getting these backwards breaks the feature in opposite
directions.

### Restarting the app you are running inside

The self-ancestry guard is right to refuse a direct restart, but on a developer machine
it fires in the normal case: the shell is `zsh → bundled app-server → ChatGPT → launchd`,
so anything run from a Codex terminal is inside the tree. A detached helper now outlives
the caller, waits for it to exit, re-enumerates, and restarts from outside the tree. It
re-runs the ancestry check itself and passes `allowHandoff: false`, so recursion is
structurally impossible. A singleton lock is transferred to it rather than contended for,
which is what avoids a helper waiting on a lock its own parent holds.

`relaunch_failed` is a new reason: a failed relaunch previously reported
`targets_survived` with an empty surviving list, sending operators to look for processes
that had all exited.

## Verification

**Platform topology measured live** on macOS (local + `macmini-cf`), Linux (`lidge`,
Ubuntu 24.04 deb) and Windows (`mini`, MSIX). Recorded in
`devlog/_plan/260913_cross_platform_desktop_app_restart/001_platform_topology.md`.

**Focused tests:** 48 pass / 0 fail across
`tests/clients/desktop-app-restart.test.ts` (19, all pre-existing Windows cases passing
unchanged, which is the evidence the move preserved that platform),
`desktop-app-restart-posix.test.ts` (15) and `desktop-restart-handoff.test.ts` (14).

**Local product suite, build and typecheck: NOT RUN**, by maintainer constraint for this
work. `zod` is also absent from the working tree, so `src/config`-importing tests cannot
run locally at all; that predates this branch. **Hosted CI at this exact head is the
gate.**

**Ten independent audit rounds** across five work phases found eighteen blockers, all
folded. Two would have terminated the caller's own session (Linux ancestry failing open;
a lock that was not actually exclusive), and one turned the hidden helper command into a
file-deletion oracle.

**Live three-host restart proof is still outstanding** and will be added to this PR
before merge.

### Known limitations, reviewed and accepted

- On Windows the app-server exclusion is a no-op: the probe enumerates `ChatGPT.exe`
  while Windows app-servers run as `codex.exe` / `codex-code-mode-host`. The restart is
  correct; the cost is one extra interrupted turn. Closing it means widening the query
  that decides what may be killed, which needs its own verification. Documented at the
  decision point in `src/cli/restart-scope.ts`.
- `ocx system codex-restart` refuses instead of handing off when the proxy itself runs
  inside the Codex app, because a proxy never exits and the handoff waits for the caller
  to do so. An honest refusal beats a promise the architecture cannot keep.

## Checklist

- [x] Targets `dev`
- [x] Focused regression tests added near the existing tests for the subsystem
- [x] User-facing behaviour change documented (English plus `fr`, `ja`, `ko`, `ru`, `tr`,
      `zh-cn`, `zh-tw`; the `catalog pull` desktop-restart exclusion sentence removed
      everywhere it appeared)
- [x] Generated surfaces regenerated, not hand-edited (`bun run skill:surface`)
- [x] No request bodies, API keys or account identifiers logged; the restart response and
      the handoff log carry pid counts and a closed vocabulary, never command lines or OS
      error text
- [ ] Local CI green — NOT RUN by constraint; hosted CI at the exact head is the gate

Plan, measurements and audit history:
`devlog/_plan/260913_cross_platform_desktop_app_restart/`.

## Summary by CodeRabbit

- **New Features**
  - `--restart-codex` now fully quits and relaunches Codex Desktop on macOS, Linux, and Windows.
  - Added `--restart-app-server-only` to restart app servers while keeping the desktop app open.
  - Added desktop restart support to `ocx catalog pull` and `ocx system codex-restart`.
  - Restarts initiated from within Codex can complete through a detached handoff.
  - `--restart-desktop-app` is now a deprecated alias for `--restart-codex`.

- **Documentation**
  - Updated CLI references, guides, and localized documentation with the new restart behavior and options.


lidge-jun:
## Live three-host verification

Proof is a **root-process identity change**, not an exit code and not the command's own
success message — twice during this work the command reported a restart it had not
performed, so the message is not evidence.

All three runs were made at `c8c1c1a9cc`. Only `tests/` and `devlog/` changed between
that commit and the current head, so the runtime proven below is byte-identical to the
one CI is green on.

| platform | host | root pid before → after | corroboration |
|---|---|---|---|
| Linux | `lidge` (Ubuntu 24.04, deb) | `3425189` → `3431091` | reparented to init, so the relaunch came from our detached `setsid` spawn and not a human; a live `--type=gpu-process --ozone-platform=x11` child confirms it reached the compositor |
| macOS | `macmini-cf` (over ssh, outside the app tree) | `74114` → `99260` | bundled app-server respawned as a child of the new root; singleton lock released |
| Windows | `mini` (MSIX, interactive session) | `29132` → `23136` | `{"attempted":true,"stopped":[29132],"surviving":[],"relaunch":"started"}` matching the observed tree |

The Linux corroboration is the one that matters most. The root process **zeroes its own
environment block** after startup — measured as 1902 NUL bytes — so the graphical session
variables have to be read from a child that inherited them before the scrub. Had that been
wrong, the pid would still have changed and the command would still have said it
succeeded, while the user got no visible app. The GPU process is what distinguishes those
two outcomes.

### Two defects this verification found that review did not

Ten audit rounds over five work phases found eighteen blockers by reading code. These two
were invisible to all of it and surfaced within seconds of running on a real Windows host.

**A false success.** The ladder returned
`{"stopped":[27788],"surviving":[],"relaunch":"started"}` while the app kept its original
pid *and* start time throughout. A stop was claimed on pid-liveness alone, and a re-probe
that merely *failed* was read as "already exited". Reporting a restart that did not happen
is worse than the stale picker this PR exists to fix.

**Then its mirror image.** Requiring the enumeration to confirm exposed the opposite bug:
`Win32_Process` **lags after a kill**, so the single confirming query still listed a
process that was already dead. The ladder declared `targets_survived`, skipped the
relaunch, and left the host with the app killed and never restarted. I restored it
immediately.

Both came from treating one weak reading as proof. Confirmation now polls until the
platform's own process list agrees, with a final look after the deadline, and an
unrunnable probe keeps the loop going rather than deciding either way. The kill and
relaunch primitives were correct on Windows the whole time; only the confirmation step was
wrong, in both directions. Two regression tests pin this and were driven red against the
unfixed ladder before being fixed.

### What CI caught that local runs could not

The macOS cases pointed at `/Applications/ChatGPT.app`, and discovery resolves the bundle
through `realpathSync`, which touches the real filesystem and cannot be intercepted by the
exec seam. They passed here because this machine has Codex installed. They now build a
real bundle under a temp directory. The privacy scan also caught a second user's home path
in two devlog files.

Both were invisible locally — the first because of what this machine happens to have
installed, the second because the scan was never run here. That is the argument for the
hosted gate, made concrete.

### Gate status

- Hosted CI at the exact head `156e10776b`: **24 pass, 0 fail**, 2 skipping.
- Local product suite, build and typecheck: **NOT RUN** by maintainer constraint. `zod` is
  also absent from the working tree, so `src/config`-importing tests cannot run here at all.
- Focused desktop-restart files locally: 50 pass / 0 fail.

### Known limitations, reviewed and accepted

- On Windows the app-server exclusion is a no-op: the probe enumerates `ChatGPT.exe` while
  Windows app-servers run as `codex.exe` / `codex-code-mode-host`. The restart is correct;
  the cost is one extra interrupted turn. Widening that query changes what may be killed,
  so it needs its own verification. Documented at the decision point in
  `src/cli/restart-scope.ts`.
- `ocx system codex-restart` refuses rather than handing off when the proxy itself runs
  inside the Codex app, because a proxy never exits and the handoff waits for the caller to
  do so.
- The local-macOS handoff path (restarting the app you are running inside) is implemented
  and unit-tested but not yet proven on a live host, because proving it ends the session
  that issues it.

lidge-jun:
## 리뷰 · 우선순위 75 / 80

이 PR은 Codex 데스크톱 앱 재시작을 Windows 전용에서 macOS/Linux까지 공통 표면으로 올리고, 그 동작을 `--restart-codex`에 접습니다. 지금 `dev`의 `src/codex/desktop-app-restart.ts` 주석은 분명히 Windows only이고, `src/cli/dispatch.ts`/`registry.ts`는 `--restart-desktop-app`을 Windows opt-in으로 두며 `--restart-codex`는 app-server SIGTERM만 약속합니다. #2292 당시 판단은 "앱을 끄면 대화가 끊기니 동의가 다르다"였습니다.

측정 결과는 그 약속이 실제 UX와 어긋난다는 쪽입니다. app-server는 데스크톱 셸의 자식이라, 자식만 죽이면 셸이 다시 띄우고 렌더러는 시작 때 만든 모델 목록을 유지합니다. 피커를 고치려면 셸 자체를 재시작해야 합니다. PR은 `DesktopAppAdapter` 뒤로 darwin/linux/windows를 두고, fail-closed 프로브·PID 재검증·루트 선택·ancestry 가드는 공통으로 가져가며 플랫폼별 identity/stop/relaunch만 바꿉니다. `--restart-codex`는 app-server + 데스크톱 full restart, `--restart-desktop-app`은 deprecated alias, `--restart-app-server-only`가 옛 좁은 동작입니다. 플래그가 충돌하면 좁은 쪽이 이깁니다.

테스트가 `tests/clients/desktop-app-restart.test.ts`, `desktop-app-restart-posix.test.ts`, `desktop-restart-handoff.test.ts`로 나뉘어 48 pass를 주장하고, Windows 기존 케이스가 깨지지 않았다는 점을 증거로 씁니다. 문서도 다국어 guides/cli에 같이 갱신합니다. 변경량이 크지만("god-file"이 아니라 desktop-app 모듈로 쪼갠 형태) 방향은 최근 유지보수 흐름과 맞습니다.

리스크는 동의 모델입니다. 예전에 `--restart-codex`만 쓰던 사람은 이제 대화가 끊길 수 있습니다. 본문은 메인테이너 지시로 #2292 결정을 뒤집었다고 명시합니다. 그 전제가 맞다면 계약 변경은 문서·help·doctor 문구까지 같이 가는 이 PR 형태가 맞습니다.

경로/심볼 - `src/cli/restart-scope.ts` (신규) - 세 플래그 우선순위를 한곳에서 해석하는 점이 좋다. sync/sync-cache/catalog pull/system codex-restart가 정말 같은 모듈을 쓰는지 tip diff에서 최종 확인.
경로/심볼 - `src/codex/desktop-app-restart.ts` + `desktop-app/{darwin,linux,windows,handoff,lock}.ts` - 루트를 "멤버"가 아니라 "셸"로 고르는 변경은 crashpad 오탐을 줄이려는 측정 결과라 설득력 있다. self-ancestry/handoff 실패 시 relaunch 금지 계약이 테스트에 고정돼 있는지 보라.
라인 - CLI help/registry 문구 - "--restart-codex는 app-server만"이던 문장이 남아 있으면 사용자 사고가 난다. 이 PR이 모든 표면을 고쳤는지 체크리스트로 보면 된다.
경로/심볼 - 로컬 product suite NOT RUN / zod 부재 - 본문이 정직하다. hosted CI tip이 게이트인 조건을 유지하자.

메인테이너의 판단이 필요한 지점
- #2292 동의 모델 역전을 릴노트에 얼마나 크게 알릴지
- 기본 `--restart-codex`가 full restart인 최종 UX가 맞는지(지시대로라면 맞음)
- Linux/macOS에서 패키지 발견 실패 시 수동 안내 문구 수위

너의 추천
방향은 merge. tip CI 초록과 help/doctor/registry 문구 일관성만 확인하고 넣자. 릴노트에 "`--restart-codex` now fully restarts the desktop app; use `--restart-app-server-only` for the old scope"를 명시하라.

이 댓글은 grok-bot이 작성했습니다

PR #4554, opened 2026-09-13
test(cursor): pin the spare-budget refund's budget and ordering

lidge-jun:
## Summary

Regression coverage for the invocation-argument refund that landed in #4543. No behaviour change: the
only `src` edit is a comment, and it corrects a claim rather than adding one.

- **A just-over-cap argument must come back byte-exact.** ~2,117 bytes against the 2,048-byte cap, so
  the cap clips about 70 bytes. The existing fixture is 4,600 bytes, where thousands of spare bytes
  surround the decision and an off-by-one in the cost arithmetic or in the newline-anchored search
  cannot show.
- **No result may be evicted to pay for a wider invocation line**, and under the fixture's uniform
  per-round costs the restored set must be the newest contiguous suffix. The second half is a
  direction check: flipping the walk to oldest-first makes it a prefix and turns the test red. The
  comment states that contiguity is *not* guaranteed under mixed argument sizes, because the pass
  skips an unaffordable line with `continue` rather than `break` — so a cheaper older line
  may legitimately be filled in after a dearer newer one was passed over.
- **A checkpoint-covered call must keep its argument tail in the replayed suffix.** Drop
  `knownCallsOffset` from the pass's `callBefore` bound and only this case notices, because
  that term is identically zero on the full-replay path.
- **A multi-byte argument must survive the round trip intact**, with U+FFFD asserted absent so a
  failure names itself instead of only reporting unequal strings.

### The `outputElided` skip is load bearing, and finding that out took two tries

This is the part worth a reviewer's attention. A sweep of single-result fixtures concluded the guard
was dead code: elision appeared to always cut the invocation line too, leaving nothing for the pass to
widen, so the earlier clipped-line lookup would always decline the root first. An adversarial
counter-read found the configuration that sweep could not reach, and the conclusion was wrong.

Truncation on its own genuinely cannot pay for a restoration — it undershoots its own budget by about
28 bytes. Initiator recovery can. With a ~519.7 KiB system prompt the history budget is roughly
4.6 KiB, the equal-share pass cuts two trailing results to ~2.3 KiB each — far enough to lose
`output:` but not the clipped invocation line — and recovery then drops the older elided sibling
so the user turn fits. Those freed bytes become spare, and the surviving elided root holds a clipped
line the pass could now afford. Removing the `outputElided` term puts the full 3,000-byte
argument into a root that shows the model no output at all.

That window is only about 24 bytes wide, so it moves whenever an envelope header changes length:
[...]

PR #4555, opened 2026-09-13
fix(web-search): assess the bridge search endpoint as a destination (#4519)

### LINKED ISSUE #4519: [Bug]: webSearchBridge.endpoint skips destination policy, so a metadata-address endpoint receives the serving provider API key
## Summary

The web-search bridge's `endpoint` option accepts any parseable `http` or `https` URL and does not run the destination policy that provider `baseUrl` values go through. `resolveOllamaWebSearchEndpoint` and the config validation in `src/config.ts` check the URL shape only — no `providerDestinationConfigError`, so no loopback, RFC1918 or cloud-metadata classification.

The consequence is narrow but real: an operator `endpoint` of `http://169.254.169.254/` receives the serving provider's API key, because the ollama backend is the one backend that spends the serving provider's own credential rather than a sidecar credential.

## Scope

This is ollama-only and predates #4515. The other backends armed by #4515 ignore `webSearchBridge.endpoint` entirely — their destinations are executor-owned (the ChatGPT forward pin, the Anthropic sidecar `baseUrl` which does pass config-time destination policy, the `api.x.ai` origin, the registry CCA URL, and the hardcoded `api.exa.ai`). #4515 did not widen this; it was found while reviewing that change and is filed separately so it is not attributed to it.

It also requires operator configuration rather than anything a client can reach. `webSearchBridge.enabled` must be true, `backend` must be `ollama`, and the `endpoint` must be written into config. That makes it a defense-in-depth gap rather than a remote defect — but the same argument was available for every provider `baseUrl`, and those are classified anyway.

## Suggested fix

Run the same destination classification on `webSearchBridge.endpoint` that `providerBaseUrlConfigError` and `providerDestinationConfigError` run on a provider `baseUrl`, at config-validation time so a bad value is refused at the write boundary rather than at send time. If a local Ollama endpoint on loopback is a legitimate configuration, that is the case to decide deliberately — either allow it explicitly the way `allowPrivateNetwork` is an explicit opt-in elsewhere, or require the operator to say so.

## Verification

Not fixed here; this is a report. Found during the security review of #4515 at `5b707d3a5c`, reading `src/web-search/passthrough-bridge.ts` around `resolveOllamaWebSearchEndpoint` and `src/config.ts` around the `webSearchBridge` schema.

lidge-jun:
Closes #4519.

## Summary

`providers.<name>.webSearchBridge.endpoint` names the URL that receives that provider's **own API key as a `Bearer` token** when the `ollama` bridge backend executes a search (`src/web-search/ollama-executor.ts` does the `fetch`). Two checks stood in front of it and neither was a destination assessment:

- `providerWebSearchBridgeConfigError` (`src/config.ts`) did `new URL` plus an `http:`/`https:` protocol test.
- `resolveOllamaWebSearchEndpoint` (`src/web-search/passthrough-bridge.ts`) returned the configured value whenever `originOf` parsed it.

Provider `baseUrl` has been assessed by `providerDestinationConfigError` for a long time. The endpoint was not assessed at all, so `endpoint: "http://169.254.169.254/latest/meta-data"` was accepted and the provider key was sent to it.

Both boundaries now call the existing `providerDestinationConfigError` from `src/lib/destination-policy.ts`. No new classifier, and no DNS added at this boundary.

**What the assessment does here**, stated exactly, because the limits matter more than the headline:

- Metadata destinations are refused unconditionally and are **not** waived by `allowPrivateNetwork`. The set is exact strings, not CIDRs: hosts `instance-data.ec2.internal`, `metadata.azure.internal`, `metadata.google.internal`; IPv4 `169.254.169.254`, `169.254.170.2`, `100.100.100.200`; IPv6 `fd00:ec2::254`; plus `::ffff:`-mapped and well-known-NAT64 wrappers of those IPv4s. Decimal and hex spellings of the IMDS address canonicalize before classification, so they are closed too.
- Loopback, localhost and private space are refused **unless** the provider sets `allowPrivateNetwork`, or its name is a registry entry that is local by definition (`ollama`, `vllm`, `lm-studio`, `litellm`). That is what keeps a self-hosted Ollama on `127.0.0.1` — or on a LAN address — working, and it is the same waiver `baseUrl` already honors.
- Public and hostname destinations pass.

**Why two call sites, and which one is load-bearing.** They cover disjoint entry paths, and the runtime one carries the weight:

- The management write path (`POST`/`PUT /api/providers`, local provider reload) runs `providerWebSearchBridgeConfigError` through `providerManagementConfigError`.
- A hand-edited `config.json`, `ocx config set` and `ocx config import` **do not**. Load-time validation is `configSchema.safeParse`, where the field is `.catch(undefined)` and the superRefine assesses `baseUrl` only. A well-shaped block with a metadata endpoint survives into running config.

So config-time alone would leave the primary operator path open. What closes it is that `resolveOllamaWebSearchEndpoint` is **the only reader of this field anywhere in `src/`**, and it now refuses: a value that survives file load can no longer be spent. Config-time validation is where an operator is told *why*; the plan-time check is the authorization boundary. The runtime refusal discards the message and fails closed silently, which is deliberate.

I deliberately did **not** add the assessment to the config-file schema. `.catch(undefined)` would swallow it into a silent disarm, and raising it through the superRefine would turn a bad opt-in block into a hard config-load failure for a field whose existing comment says it must not trip invalid-config recovery.

I chose to widen `providerWebSearchBridgeConfigError` to take the provider name and the provider rather than inlining the check at its single caller, so all `webSearchBridge` validation stays in one function. The caller already had both values in scope; they are the same pair it passes for `baseUrl`.

**Operator signal.** A refused endpoint disarms the bridge and returns no error, because disarming is what keeps the key unspent. That silence was a real regression: a provider keyed under a CUSTOM name (for example `my-ollama`) pointing at a loopback endpoint used to arm, and now does not, since only the registry ids are local by default. The config file never reaches the config-time validator, so nothing else would tell the operator. `resolveOllamaWebSearchEndpoint` therefore emits one warning per provider and endpoint naming the remedy. The destination URL is deliberately omitted and the provider key is passed through `redactSecretString`, because a provider key is caller-controlled and can be token-shaped.

### Residuals a reviewer should weigh

1. **A hostname endpoint that resolves to metadata or private space still passes.** The synchronous classifier is literal-only by design. This residual is **strictly larger than `baseUrl`'s**: `baseUrl` additionally gets the async `providerDestinationResolvedError` at management write, which the endpoint does not. Closing that gap at management write would still leave the hand-edited-file path uncovered, because the plan-time boundary is synchronous. Out of scope here by instruction; happy to follow up.
2. **Generic link-local (`169.254.1.1`) and `0.0.0.0` pass this classifier.** Pre-existing `providerDestinationConfigError` behavior, identical for `baseUrl`.
3. **The executor still trusts `plan.endpoint`.** `runOllamaWebSearch` fetches whatever string it is handed and the executor factories do not re-check. That is not a bypass today because the planner is the only production builder of `plan.endpoint`, but it is a one-caller invariant rather than a fetch-site guarantee. Redirects are already `redirect: "manual"`, so the key is not followed off-endpoint.
4. **If the serving provider is named `anthropic`, `google-antigravity` or `orcarouter-oauth`**, `providerDestinationConfigError` also applies its OAuth-https rule to the endpoint, and the rewritten message then mentions OAuth credentials. It only ever *adds* a restriction, and the bridge arms for `authMode: "key"` providers only, so the edge is narrow and safe.
5. The config-time check runs regardless of `backend`, though `endpoint` is only read for `backend: "ollama"`. Stricter than the spend path, and intentional since `backend` can be flipped later.

## Verification

- **The local product test suite, `bun run typecheck`, the build and `bun install` were NOT RUN.** This worktree has no `node_modules` and running them was excluded by the delivery policy for this change. No local result is cited as evidence.
- The only proof is hosted Cross-platform CI at the exact head SHA `e8b36b0e202025780e84759542a78cb1488b2333`. Run id: `34781031241`.
- In place of local execution: three read-only recon passes (endpoint entry paths, exact classifier semantics, full call-site and test impact), one adversarial audit of the plan, and one adversarial counter-read of the finished patch. The audit rejected an earlier framing of the config-file decision and replaced it with the single-reader invariant above. The counter-read re-derived the compile surface, every caller, the classifier expectations behind each new test, and the send path, and returned no blockers.
- New coverage in `tests/web-search/web-search-passthrough-bridge.test.ts`. Endpoint policy at plan time: metadata disarms; metadata still disarms under `allowPrivateNetwork`; the Aliyun address stays refused under the opt-in; a private endpoint disarms without the opt-in and arms with it; `providerName: "ollama"` arms both a loopback and a LAN endpoint via the registry default; the same loopback endpoint is refused under a name with no default; a public endpoint arms; and a metadata-lookalike hostname arms, pinning residual 1 as an asserted fact rather than an assumption. Config time, which had no test at all before: the metadata refusal names `webSearchBridge.endpoint` rather than `baseUrl` and persists under the opt-in; a private endpoint errors without the opt-in and passes with it; public and absent endpoints pass; and the shape check still runs before the destination check.
- `structure/runtime.md` (owner of `src/web-search/` per `structure/INDEX.md`) records the rule and the residual. No new test file, so `layout.json` and `tests/fixtures/test-layout-expected.json` are unchanged.

## Checklist

This change is security-sensitive and requests independent review under `MAINTAINERS.md`. It touches no credential handling or OAuth flow: it reads a URL and a boolean and decides whether the bridge may arm. The default direction is refusal, and the one place existing behavior is preserved rather than tightened is the local-by-default registry waiver, which is required for self-hosted Ollama and matches `baseUrl`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Web search bridge endpoints are now validated consistently with other provider destinations.
  - Metadata endpoints are rejected, while loopback and private-network endpoints require explicit provider permission unless locally trusted.
  - Invalid or unsafe endpoints are declined during use and reported during configuration.
  - Refused endpoints now generate a deduplicated warning with guidance for reauthorization.

- **Documentation**
  - Updated runtime guidance to describe endpoint validation rules and the authorization boundary for hosted web-search passthrough.


lidge-jun:
## 리뷰 · 우선순위 74 / 80

설명
이 PR은 열려 있는 보안 구멍 #4519를 막습니다. 지금 `dev`(HEAD `56c956715`, 직전에 #4554 spare-budget 테스트가 들어옴)의 스냅샷에도 “`webSearchBridge.endpoint`가 destination policy를 건너뛴다”가 그대로 남아 있습니다. 문제는 간단합니다. `providers.<name>.webSearchBridge.endpoint`는 ollama 브릿지 백엔드가 검색을 돌릴 때 그 프로바이더의 API 키를 Bearer로 붙여서 보내는 URL입니다. 그런데 지금까지는 `providerWebSearchBridgeConfigError`가 URL 모양과 http/https만 보고, `resolveOllamaWebSearchEndpoint`도 `originOf`로 파싱만 되면 그대로 통과시켰습니다. 프로바이더 `baseUrl`은 오래전부터 `src/lib/destination-policy.ts`의 `providerDestinationConfigError`로 메타데이터/사설망을 거르는데, endpoint만 그 검사를 안 받아서 `http://169.254.169.254/...` 같은 값이 설정되면 키가 클라우드 메타데이터로 나갈 수 있었습니다.

고치는 방식은 새 분류기를 만들지 않고 기존 `providerDestinationConfigError`를 두 군데에 붙이는 것입니다. (1) 관리 API 쓰기 경로의 `providerWebSearchBridgeConfigError`(`src/config.ts`, 호출은 `src/server/auth-cors.ts`의 `providerManagementConfigError`) — 여기서는 운영자에게 “왜 안 되는지” 메시지를 줍니다. (2) 계획 시점의 `resolveOllamaWebSearchEndpoint`(`src/web-search/passthrough-bridge.ts`) — 손수 고친 `config.json`, `ocx config set`, `ocx config import`는 스키마만 타고 관리 검증을 안 거치므로, 이 함수가 트리에서 endpoint를 읽는 유일한 곳이고 여기서 거절하는 것이 실제 권한 경계입니다. 거절은 일부러 조용히 `undefined`를 돌려 브릿지를 무장 해제합니다. 메타데이터는 `allowPrivateNetwork`로도 열리지 않고, 루프백/사설망은 옵트인 또는 레지스트리 로컬 기본값(`ollama` 등)이 있을 때만 통과합니다. `planPassthroughWebSearchBridge`에 `providerName`을 필수로 넣고 `src/server/responses/core.ts`에서 `route.providerName`을 넘기도록 바꿨으며, 테스트와 `structure/runtime.md`에 규칙·잔여 위험을 적어 두었습니다. Cross-platform CI가 헤드 SHA에서 돌아가고 있고, 일부 잡(테스트 샤드, gates, hygiene 등)은 아직 pending입니다. MAINTAINERS.md 기준으로 보안 민감 변경이라 독립 리뷰가 필요합니다.

라인 `src/config.ts` `providerWebSearchBridgeConfigError` - 에러 문구를 `.replace(/^baseUrl/, "webSearchBridge.endpoint")`로 고칩니다. 지금 sync 메시지는 모두 `baseUrl`로 시작해서 동작하지만, 나중에 destination-policy 문구가 바뀌면 필드 이름이 안 바뀌거나 반만 바뀔 수 있습니다. 메시지 조립을 정책 쪽에 넘기거나 prefix를 인자로 받는 편이 더 단단합니다.
라인 `src/web-search/passthrough-bridge.ts` `resolveOllamaWebSearchEndpoint` - 계획 시점 거절이 조용해서, 손수 설정한 메타데이터/사설 endpoint는 브릿지만 꺼지고 운영자 로그/에러가 없을 수 있습니다. 보안상 맞지만, 디버깅 힌트(내부 로그 한 줄)를 남길지 판단이 필요합니다.
경로 `runOllamaWebSearch` / `plan.endpoint` - PR이 밝힌 대로 executor는 받은 endpoint 문자열을 다시 검사하지 않습니다. 지금은 planner가 유일한 생산자이고 redirect는 manual이라 우회는 아니지만, fetch 직전에 한 번 더 검사하면 한 호출자 불변식에 덜 의존합니다.
경로 DNS residual - sync 분류만 써서 호스트명이 나중에 메타데이터/사설 IP로 풀리면 통과합니다. `baseUrl`은 관리 쓰기에 async `providerDestinationResolvedError`가 있는데 endpoint는 없습니다. PR이 범위를 의도적으로 좁힌 것이므로 후속 이슈로 열지 여부만 정하면 됩니다.
경로 CI - 헤드에서 api usage/changes/docker smoke 등은 통과했지만 test 샤드·gates·hygiene·npm-global 일부가 pending입니다. 머지 전에 초록을 확인해야 합니다.

메인테이너의 판단이 필요한 지점
- 보안 민감 PR로 독립 리뷰어를 한 명 더 붙일지(본인이 연 PR이기도 함).
- executor fetch 직전 재검사와 관리 쓰기 async DNS를 이번 머지에 넣을지, #4519 후속으로 쪼갤지.
- 계획 시점 조용한 거절을 그대로 둘지, 내부 관측 가능하게 할지.
- CI 전체 초록을 머지 게이트로 할지(로컬 스위트는 의도적으로 안 돌림).

너의 추천
CI가 전부 초록이 되면 독립 리뷰 한 번 받은 뒤 `dev`에 머지하세요. #4519를 닫는 정직한 최소 패치이고, 스냅샷에 남아 있던 endpoint 정책 공백을 바로 없앱니다. DNS parity와 executor 재검사는 머지 후 follow-up 이슈로 남겨도 됩니다. residual을 문서·테스트로 고정한 점은 유지하세요.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
Adversarial security review of this PR, posted so the record exists before any merge decision. **I am not approving it and I will not self-integrate it.** MAINTAINERS.md line 69 requires explicit security review for credential-handling and security-boundary changes, and this endpoint receives the serving provider's API key as a Bearer token. The `dev` self-integration exception covers the missing second approval; it does not cover this. @Ingwannu, this one is yours.

**Verdict: fail, one blocker.**

A provider keyed under a non-registry name — `my-ollama`, `local-llm` — with `endpoint: http://127.0.0.1:11434/api/web_search` was armed before this change and now disarms with no operator-visible signal at all. The config file load at `src/config.ts`:684 uses `.catch(undefined)` and runs no error function, so the block loads cleanly, and the plan-time refusal in `resolveOllamaWebSearchEndpoint` discards its message by design and returns `undefined`. Web search simply stops. Registry ids carrying `allowPrivateNetworkByDefault` (`ollama`, `vllm`, `lm-studio`, `litellm`) and an explicit `allowPrivateNetwork: true` are unaffected, so the break lands exactly on the custom-name case, which is the common one. The fix is one operator-visible warning naming the provider and the remedy.

**What held up under attack.** Every literal bypass is closed by WHATWG canonicalization plus `normalizeHostname`: mixed-case hosts, `::1`, `::ffff:127.0.0.1`, hex-mapped forms, decimal `2852039166`, octal `0177.0.0.1`, `127.1`, trailing dots, userinfo, and NAT64 `64:ff9b::` embedding. Redirects are safe because the executor sets `redirect: "manual"`. Entry-path enumeration found no route that reaches the executor unvalidated: management writes all funnel through `providerManagementConfigError`, and although the file-load path is not validated, `resolveOllamaWebSearchEndpoint` is the only reader of the field in the tree, so an unvalidated value cannot be spent. Error messages carry classification detail and the opt-in hint but never the URL or the credential.

**Declared residual, not a defect to fix here.** A public hostname that resolves into private or metadata space passes a literal-only check, and unlike `baseUrl` the endpoint gets no asynchronous `providerDestinationResolvedError` at management write. The plan-time boundary has to stay synchronous, so this matches the project's already-recorded rebinding residual — but it should be stated in the description rather than left for a reviewer to find.

The lane is fixing the blocker now. Once it pushes, the exact-head CI run id will be added here.

lidge-jun:
Blocker fixed and re-verified. Updating the record.

Head is now `e8b36b0e2`. Cross-platform CI run [34781031241](https://github.com/lidge-jun/opencodex/actions/runs/34781031241) completed **success** at that SHA. Local product suite, typecheck, build and install NOT RUN.

The fix does what the review asked and nothing more: a refused endpoint now emits one warning per provider-and-endpoint pair naming the remedy — set `allowPrivateNetwork: true`, or key the provider under a registry id — while the refusal itself still returns `undefined`, because disarming is what keeps the key unspent. The URL is deliberately omitted from the message and the provider name goes through `redactSecretString`, since a provider key is caller-controlled and can be token-shaped. The dedupe set is bounded at 64 and cleared rather than grown, and there is an explicit test seam because the dedupe is process-wide.

The comment above `resolveOllamaWebSearchEndpoint` was also corrected: it previously claimed config-time is where an operator is told why, which is not true for a hand-edited file that never reaches `providerWebSearchBridgeConfigError`.

**Still not merging this.** It remains a credential-destination change under MAINTAINERS.md line 69, and the `dev` self-integration exception covers a missing second approval, not the security review. @Ingwannu — the adversarial review above plus this fix is the full package; the one declared residual is that a public hostname resolving into private or metadata space passes a literal-only check, matching the project's existing rebinding residual on `baseUrl`.

lidge-jun:
Merging this now on the explicit instruction of the project owner, who told me to use admin merge on my own judgment. Recording exactly what that does and does not change.

**What was waived, stated plainly.** MAINTAINERS.md line 69 requires explicit security review for credential-handling and security-boundary changes, and this endpoint receives the serving provider's API key as a Bearer token. No second maintainer reviewed it; @Ingwannu was requested and had not responded. The owner's instruction is the authority for proceeding, not a finding that the requirement did not apply. This comment exists so that fact is in the record rather than implied by an `--admin` flag.

**What was actually verified.** Cross-platform CI run [34781031241](https://github.com/lidge-jun/opencodex/actions/runs/34781031241) completed **success** at `e8b36b0e202025780e84759542a78cb1488b2333`, which is the head being merged. Local product suite, typecheck, build and install were NOT RUN — in a lane worktree without `node_modules` they could not have run — so hosted CI at that SHA is the only evidence, and it is evidence for exactly this commit.

An adversarial security review was performed against this diff and returned **fail** on a real finding, which was fixed and re-verified before this merge. The finding was a silent regression: a provider keyed under a custom name such as `my-ollama` pointing at a loopback endpoint used to arm the bridge, and the new destination policy refused it with no operator-visible signal at all, because the config load path runs no error function and the plan-time refusal returns `undefined` by design. Web search would simply have stopped. The fix emits one warning per provider-and-endpoint pair naming the remedy, with the URL omitted and the provider name redacted because a provider key is caller-controlled and can be token-shaped.

The same review checked the attack surface and found the literal bypasses closed by WHATWG canonicalization plus `normalizeHostname`: mixed-case hosts, `::1`, `::ffff:127.0.0.1`, hex-mapped forms, decimal and octal IPv4, `127.1`, trailing dots, userinfo, and NAT64 embedding. Redirects are safe because the executor sets `redirect: "manual"`. Entry-path enumeration found no route reaching the executor unvalidated: management writes funnel through `providerManagementConfigError`, and although the file-load path is not validated, `resolveOllamaWebSearchEndpoint` is the only reader of the field in the tree, so an unvalidated value cannot be spent.

**Declared residual, carried into `dev` knowingly.** A public hostname that resolves into private or metadata space passes a literal-only check, and unlike `baseUrl` this endpoint gets no asynchronous `providerDestinationResolvedError` at management write. The plan-time boundary has to stay synchronous, so this matches the project's already-recorded rebinding residual rather than introducing a new class of gap.

PR #4575, opened 2026-09-14
fix(providers): close the Devin rekey host window and diagnose Antigravity quota blocks

### LINKED ISSUE #3781: [Bug]: Antigravity quota refresh failure — investigate missing canonical Fake-IP handling
### Canonical quota transport landed; field acceptance remains open

[#3799](https://github.com/lidge-jun/opencodex/pull/3799) (`f7bc920051`) supplies fixed-destination canonical proof and aligns Antigravity quota fallback transport. Authenticated TUN/Fake-IP observation and safe user-facing failure categorization remain outstanding; this is not a claim that the reporter's full scenario was verified.

Verified against dev `5759d9ea2f1e7281cdc01eb9628f2e0a123fb59c`. Original report by @jaychou0642-create. The attribution record was added in [#3811](https://github.com/lidge-jun/opencodex/pull/3811).

## Problem

The Google Antigravity Accounts page can display “额度刷新失败” (Quota refresh failed) and “速率限制数据暂时不可用” (Rate-limit data temporarily unavailable) after clicking Refresh quota.

## Code-level investigation

In the 2.43.0 implementation, `fetchAntigravityUsageQuota()` in `src/providers/quota.ts` calls `providerOutboundPost()` for the fixed Google `retrieveUserQuotaSummary` and `fetchAvailableModels` URLs. Its outbound dependencies default to an empty object.

In `src/lib/provider-outbound.ts`, the transparent Fake-IP exception requires an `isCanonicalUrl` proof when there is no explicit outbound proxy. Model discovery supplies such a proof, but the quota path appears not to.

This suggests a compatibility gap worth testing: quota probes may reject Fake-IP DNS answers even where model discovery supports transparent proxying. This is a hypothesis from source inspection, not a confirmed end-to-end root cause.

## Expected behavior

- Test quota probing with Fake-IP DNS and no explicit HTTP proxy.
- If needed, provide a narrowly scoped canonical-destination proof for the fixed Google quota URLs.
- Preserve TLS verification, redirect restrictions, and rejection of unrelated private or metadata destinations.
- Surface a safe error category for failed quota probes instead of only the generic unavailable message.

No authenticated upstream response or verified workaround is included in this report.

### LINKED ISSUE #4503: Devin provider merge: detached credential rekey leaves a host-selection window (and a test-coverage note)
### Client or integration
Direct HTTP/API client

### Area

### Summary
`runDevinProviderMergeStartupMigration` (src/providers/devin-provider-merge-migration.ts:202, commit f7d9dbad) writes the merged `providers["devin"]` config synchronously, then fires `void rekeyProviderCredentials("devin-cli","devin")` detached. `getCredential` is a literal slot key, so until the rekey lands — or permanently for that process if it fails (retried next boot) — `resolveDevinApiServer` reads an empty `devin` slot and falls back to the default US host. EU/FedStart tenants send their key to the wrong host in that window; GUI login status is similarly stale. The request-time TOKEN path is unaffected (resolveDevinToken uses provider.apiKey / forwarded Authorization / env, never the OAuth slot), so this is a host-selection and status-window defect, not a credential-loss defect. Collision handling refuses by design with a logged warning.

Bundled test-coverage note from the same audit: PR #4497's Pi-shape image-part regression cases cover the tool path by composition; a direct `role:"tool"` envelope fixture would catch a tool-branch-only regression.

### Reproduction
1. Configure a devin-cli OAuth credential for an EU/FedStart tenant on 2.52.0.
2. Upgrade to dev @ eb81eaaf8d and start the proxy.
3. During the window between the config save and the detached rekey (or after a rekey failure), requests resolve the default US host instead of the tenant host.

Found by the 2.53.0 release regression audit (lane13, maintainer-downgraded from the lane's P1 claim after verifying the token path). Synthesis disposition: follow-up issue — self-healing, logged, retried on boot; not promotion blocking.

### Version
dev @ eb81eaaf8d (post-2.52.0, pre-2.53.0)

### Operating system
macOS (any; codepath is platform-independent)

### Provider and model
devin (OAuth, EU/FedStart tenants)

### Logs or error output
[devin-provider-merge] credential rekey failed and will retry on the next start: ...

N/A — evidence is source-level, cited inline.

### Redacted configuration
providers.devin (merged from devin-cli)

### Checks

lidge-jun:
## Summary

Two provider account-lifecycle edges, both reported against the account pool.

**#4503 — Devin merge left a host-selection window.** `runDevinProviderMergeStartupMigration` writes the merged `providers["devin"]` config row synchronously, then fires `rekeyProviderCredentials("devin-cli","devin")` detached, because `startServer` is synchronous and `mutateStore` is async-only. `resolveDevinApiServer` read `getCredential(providerId)` as a literal slot key, so in the window between the config save and the rekey landing the row already said `devin` while the credential was still under `devin-cli`. `getCredential("devin")` returned nothing and the tenant host fell through to the configured `baseUrl` or the US default `https://server.codeium.com` — an EU or FedStart account then sent `GetUserJwt`, api_key included, to a host it is not provisioned on. A failed or collision-refused rekey made that permanent for the process, not just transient.

The literal slot is still consulted first, and that ordering is load-bearing: an unmigrated row still named `devin-cli` owns its own slot, so normalizing the id up front would read the wrong one. Only when the literal slot yields no usable `apiBaseUrl` are the slots linked by `DEPRECATED_OAUTH_PROVIDER_ALIASES` consulted, in both directions, and every candidate still passes `validateDevinApiBaseUrl` — an alias slot is not trusted more than the literal one. Ordering after that is unchanged: configured `baseUrl`, then the US default.

This introduces an import cycle (`src/oauth/devin.ts` -> `./index` for the alias map, while `./index` imports `loginDevin`/`refreshDevinToken` from `./devin`). It is safe in both entry orders because the map is read inside a function at call time, never during module evaluation, and the two symbols `index.ts` needs from `devin.ts` are hoisted `export async function` declarations that are initialized at instantiation. Deriving the pair from the map rather than hard-coding a second `"devin-cli"` literal keeps the map the single source of truth.

**#3781 — Antigravity quota refresh failure.** The canonical-destination proof the issue asks for landed earlier as `isCanonicalAntigravityQuotaUrl`, so both fixed Google accounting URLs already get the transparent Fake-IP exception. What was still missing is the issue's last expectation: a safe error category instead of a generic unavailable. `probeAntigravityUsageQuota` discarded the summary attempt's failure in a bare `catch {}`, so when the summary was rejected by outbound policy (`destination_blocked`) or DNS (`dns_failed`) — exactly the Clash/Surge/Mihomo TUN symptom in the report — and the models fallback then also failed, the operator saw only the fallback's vaguer diagnosis and was pointed at Google instead of at their own network. The summary's classified failure is now retained and outranks a non-policy fallback diagnosis. A successful fallback still clears it completely, the summary's immediate `redirect_blocked`/`access_denied` returns are unchanged, and the `legacy` channel keeps throwing the exact same error object so `fetchAntigravityUsageQuota`'s identity contract holds.

Diagnoses stay inside the closed `QUOTA_FAILURE_CODES` union; no upstream text, URL, credential or routing policy enters one.

## Verification

**Local suite runs were NOT RUN.** No `bun run test`, no bare `bun test`, no `bun run typecheck`, no `bun install`, no `bun run build:gui` — `node_modules` is absent in this worktree by design for this lane. Proof for this unit is hosted CI at the exact final head of this branch and nothing else; the run id and conclusion are reported with the head SHA.

In place of local execution, every new call site was checked against the real declaration by reading it: `createDevinAdapter(provider, { providerId })`, `ProviderAdapter.runTurn(parsed, IncomingMeta, emit)`, `IncomingMeta`'s `{ headers, translatorBudget }` requirement, `OcxContext.tools` being optional, `saveCredential(provider, cred)`, and `validateDevinApiBaseUrl` accepting the EU, FedStart and staging hosts used as fixtures.

Regression coverage added next to the existing tests for each subsystem:

- `tests/providers/devin-login.test.ts` — the migration window resolves the tenant host from the un-rekeyed alias slot; that tenant beats a leftover configured `baseUrl`; the literal slot still wins when both slots hold a credential; an alias slot carrying a non-Devin origin is not trusted.
- `tests/providers/devin-adapter.test.ts` — the same guarantee at the host the adapter actually dispatches to, driving a real `runTurn` with a fetch stub that records every outbound URL.
- `tests/providers/provider-account-quota.test.ts` — the Fake-IP probing the issue explicitly asked to test: with no proxy environment, a `198.18.x.x` answer is admitted for both canonical quota URLs with TLS verification intact; the exception stays narrow against a lookalike host, a different path, an appended query string and a different provider name; loopback, RFC1918, link-local and cloud-metadata answers are still rejected and surface as `destination_blocked`.
- `tests/responses/chat-completions-endpoint.test.ts` — the coverage note bundled in #4503: a direct `role:"tool"` envelope carrying Pi- and Anthropic-shaped image parts, which the PR #4497 cases only reached by composition through the `user` branch.

Planning notes: `devlog/_plan/260914_l5_provider_account_edges/`.

## Checklist

Closes #4503
Closes #3781

## Summary by CodeRabbit

- **Bug Fixes**
  - Devin requests now use the correct tenant API host, including during provider-account migration, while preserving configured and default fallbacks.
  - Antigravity quota errors now retain and prioritize actionable network diagnostics, such as blocked destinations or DNS failures.
  - Antigravity quota checks now support approved fake-IP responses while continuing to reject unsafe destinations.
  - Tool results containing non-OpenAI image formats are now correctly converted and preserved in requests.


lidge-jun:
## 리뷰 · 우선순위 73 / 80

지금 `dev`(HEAD `d45a7e749`, 방금 #4298 bridge zero-copy가 올라온 상태)에서 Devin은 `devin-cli` → `devin` 병합을 씁니다. 설정 행은 동기로 `providers["devin"]`에 저장되는데, credential 슬롯 rekey는 `startServer`가 동기라서 떨어져 나갑니다. 그 사이(그리고 rekey가 실패하거나 목적지 슬롯 충돌로 거절되면 그 프로세스 내내) `resolveDevinApiServer`가 리터럴 `"devin"` 슬롯만 보면 EU/FedStart 테넌트의 `apiBaseUrl`을 못 읽고 US 기본 호스트 `https://server.codeium.com`으로 키를 보냅니다. 그게 열린 #4503입니다. Antigravity 쪽은 canonical Fake-IP 예외(`isCanonicalAntigravityQuotaUrl`)는 이미 들어와 있는데, `probeAntigravityUsageQuota`가 summary 실패를 맨몸 catch로 버려서 Clash/Surge/Mihomo TUN 같은 `destination_blocked`/`dns_failed`가 fallback의 더 흐린 진단에 가려집니다. 그게 열린 #3781의 남은 구멍입니다.

이 PR은 두 구멍을 한 레인으로 막습니다. `src/oauth/devin.ts`의 `resolveDevinApiServer`는 리터럴 슬롯을 먼저 보고, 쓸 만한 `apiBaseUrl`이 없을 때만 `DEPRECATED_OAUTH_PROVIDER_ALIASES`로 양방향 연결된 슬롯을 같은 allowlist(`validateDevinApiBaseUrl`)로 읽습니다. 리터럴을 먼저 두는 순서가 맞습니다. 아직 이름이 `devin-cli`인 행은 자기 슬롯을 가져야 하니까요. Antigravity는 summary 실패를 `summaryFailure`로 남기고, fallback도 unavailable일 때 네트워크 정책 진단을 우선하는 `antigravityUnavailableFailure`를 둡니다. fallback이 성공하면 첫 실패는 지워지고, summary의 즉시 `redirect_blocked`/`access_denied` 반환은 그대로입니다. 테스트는 마이그레이션 창·어댑터 실제 dispatch URL·Fake-IP 좁은 예외·tool 경로 이미지 파트를 각각 덮습니다. 계정 수명주기/아웃바운드 진단이라 지금 `dev` 방향과 바로 맞고, 타입스플릿 무효화 대상도 아닙니다.

`src/oauth/devin.ts` `resolveDevinApiServer` - 리터럴 슬롯 우선 + alias 양방향 조회 + 동일 allowlist 순서가 맞습니다. 다만 `./index`에서 `DEPRECATED_OAUTH_PROVIDER_ALIASES`를 가져오면서 import cycle이 생깁니다. 호출 시점에만 맵을 읽는다고 본문에 적혀 있고 지금 구조상 안전해 보이지만, 모듈 평가 순서가 바뀌면 깨질 수 있으니 cycle을 남길지 별도 상수 모듈로 뺄지 한 번만 더 보면 좋습니다.

`src/providers/quota.ts` `antigravityUnavailableFailure` / `probeAntigravityUsageQuota` - summary의 `destination_blocked`·`dns_failed`만 fallback보다 올리는 조건이 이슈 증상과 맞습니다. 진단이 `QUOTA_FAILURE_CODES` 안에만 머무르는 것도 좋습니다. 성공한 models fallback이 summary 실패를 완전히 지우는지도 테스트로 고정돼 있으면 더 안심입니다.

`devlog/_plan/260914_l5_provider_account_edges/010_wp1_account_lifecycle_edges.md` - 게이트가 `empty_catch`로 막았습니다. 실제 코드의 bare `catch {}`가 아니라, 계획 문서에 적힌 `catch {}` 문자열을 hygiene이 경로로 잡은 겁니다. 라벨 `intake: hygiene-blocked`와 draft 게이트가 여기 걸려 있습니다. 문서 표현만 바꿔도 풀릴 가능성이 큽니다.

`tests/providers/devin-adapter.test.ts` / `devin-login.test.ts` / `provider-account-quota.test.ts` - 회귀 범위는 이슈와 잘 맞습니다. 로컬 스위트는 안 돌렸다고 명시했고, 이 시각 호스팅 테스트 샤드는 아직 pending입니다. exact-head 초록을 머지 조건으로 두면 됩니다.

메인테이너의 판단이 필요한 지점
- plan 마크다운의 `catch {}` 문구 때문에 hygiene이 막힌 상태를 문서 수정으로 풀지, 스캐너 예외를 둘지.
- `oauth/devin.ts` ↔ `oauth/index` import cycle을 이번 레인에서 끊을지, 호출 시점 읽기만으로 받아들일지.
- #4503과 #3781을 한 PR로 같이 닫을지(지금은 L5 레인 한 커밋으로 묶여 있음).

너의 추천
hygiene의 plan-문서 `empty_catch`만 먼저 지우고, exact-head CI가 초록이면 `dev`에 머지하세요. #4503·#3781은 이 PR로 닫는 게 맞습니다. 테넌트 호스트로 키가 새는 창이라 우선순위가 높습니다.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:

Exact-head evidence at 7d223ee7b: 25 successes, 2 skips, no failures.

This branch touches src/oauth/devin.ts, which is inside the AGENTS.md security boundary, so it got an explicit security review rather than an ordinary one. That review failed the first version, for a specific reason worth recording: the alias-tenant fallback keyed off the literal credential having no usable apiBaseUrl, which also covers a credential that exists and merely carries a missing or off-allowlist host. Because the rekey refuses an occupied destination slot, both slots can hold credentials belonging to two different accounts, and in that state a request kept account A's bearer while taking account B's EU or FedStart host.

The fallback now runs only when the literal slot holds no credential at all, which is precisely the unmigrated window the fix was written for: the config row already says devin while the credential still sits under devin-cli. A credential that exists with an unusable host falls through to the configured base URL and then the US default, exactly as before. A regression test pins the case the original tests were missing — both slots occupied, literal host unusable, alias holding a different tenant.

One correction on top of that: getCredential returns null rather than undefined for an empty slot, so the first version of the narrowed guard disabled the fallback outright and the un-rekeyed case fell through to the configured base URL. The guard accepts either absent form now.

The quota change in src/providers/quota.ts retains the summary probe's classified failure so it outranks a non-policy fallback diagnosis, which is the remaining ask on the Antigravity refresh report.

No token, key, or account identifier is newly logged, serialized, or placed in an error message, and missing credentials still fail closed.

PR #4580, opened 2026-09-14
fix(codex): hold a bound thread's account for its prompt cache (#4546)

lidge-jun:
## Summary

Pooled Codex routing moved a **live** conversation whenever its account crossed `autoSwitchThreshold`, and moved it to whichever account was strictly cooler with no floor under the destination. Provider prompt caches are account-isolated, so every hop re-sent the entire prefix. Once a pool sat in the 80-99% band the coolest account was still over the threshold, so the thread was handed on again the next turn, and the turn after that. The same predicate short-circuits the 60-second re-score interval, so in that band a bound thread was re-scored on every request. #4546 reports the result: roughly 1.9B total tokens and 323M uncached tokens across 15,607 requests in about thirteen hours, with a 7k-token turn arriving upstream as a 150k-token turn.

`pool.cacheAffinity` (#4292) already raised the eviction bar, but it was opt-in and off, so the install that gets hurt is exactly the install that never heard of it. And turning it on did not close the hole: a transient failure streak deleted the binding through a different path that never consulted the flag.

Three changes, each independently revertible:

- **`pool.cacheAffinity` defaults ON.** A bound thread keeps its account until that account genuinely cannot serve: unusable, paused, plan-excluded, credential-invalid, generation-stale, TTL-expired, quota-refused, or known at 100%. An explicit `false` restores capacity-first routing byte-for-byte.
- **A move needs somewhere worth moving to.** A bound thread may only move to a destination with genuine quota headroom *and* strictly lower usage, under **either** setting of the flag. `CODEX_UNKNOWN_USAGE_SCORE` is 101, so an unmeasured account can never be strictly cooler than a known over-threshold score and is excluded without a special case. When the whole pool is hot, nobody moves - there is nothing to win by trading a warm prefix for an equally hot account.
- **A transient streak is a detour, not an eviction.** A 5xx streak routes the current request to a *remembered* alternate while the thread keeps its home, and the thread returns as soon as its own account serves again. Bounded at 10 minutes, after which the binding is released normally, because an unbounded hold would keep a thread detouring past an outage that is not coming back. This is independent of `cacheAffinity`: attributing a 5xx is not a quota preference, and #4269 already showed how badly failure classification misfires when it reads the wrong signal. The model-detour lane gets the same hold, otherwise a model-scoped request drops its pin on three 503s and falls back to a home account that may not be entitled to the model.

[...]

PR #4581, opened 2026-09-14
fix(codex): keep a bound thread's prompt cache and tell configured routing apart from adopted routing

### LINKED ISSUE #4546: [Bug]: Account pool routing destroys prompt cache and triggers 10x-50x token burn death-spiral above 80% usage threshold
# [Bug]: Account pool routing destroys prompt cache and triggers 10x–50x token burn death-spiral above 80% usage threshold

**Repository**: `lidge-jun/opencodex` (Package: `@bitkyc08/opencodex`)  
**Labels**: `bug`, `critical`, `routing`, `performance`, `cost-anomaly`

---

## 1. Summary

When multiple accounts are registered in OpenCodex, the current routing algorithm in `src/codex/routing.ts` creates two major compounding failure modes under heavy autonomous / agentic workloads:

1. **Subagent Workload Cascading**: When a large volume of one-off, single-turn subagent tasks (e.g. IDE background checks or repo inspections, ~150k tokens each with 0% cache) hits the proxy, OpenCodex routes them sequentially to fresh accounts as each hits the `autoSwitchThreshold = 80%` (default: `config.autoSwitchThreshold ?? 80`). This drains every account in the pool sequentially.
2. **The 80% Multi-Turn Ping-Pong Death Spiral**: Once accounts reach `autoSwitchThreshold`, `reevaluateAffinityQuota()` bypasses the 60-second session affinity lock on **every turn** and calls `pickLowerUsageAccount()`. Active, long-running multi-turn conversations (which previously enjoyed a **95%+ cache hit rate**) are suddenly bounced between Account 1 <-> Account 2 <-> Account 3 on **literally every turn**. Because OpenAI Codex prompt caches (GPT-6 Astra / GPT-5.6 Luna / Sol) are strictly tenant/account-isolated, this reduces the multi-turn cache hit rate to **0%**, turning a 7k-token turn into a 150k-token raw billing catastrophe.

**Real-world telemetry from our audit: Over 1.90 Billion tokens (1,902,323,250 tokens across 15,607 requests) were burned in ~13 hours across 5 Pro accounts.**

---

## 2. Environment & Version

* **OpenCodex Version**: `@bitkyc08/opencodex` (latest)
* **Upstream Provider**: OpenAI Codex (`chatgpt.com/backend-api/codex` / responses API)
* **Models**: `gpt-6-astra`, `gpt-5.6-luna`, `gpt-5.6-sol`, `gpt-5.6-terra`
* **Account Configuration**: Multi-account pool (5 Pro subscriptions)
* **Client**: Orca / Codex CLI connected to `http://127.0.0.1:10100`

---

### 3.1 Phase 1: How Account 2 & 3 Reached 80% (Subagent Workload Cascading)
Our detailed investigation of `usage.jsonl` revealed:
* Account 2 (`orca-d***0`) processed 741 total conversations.
* **732 of these were one-off single-turn subagent calls** dispatched during autonomous workflows.
* Each single-turn call submitted ~148k tokens of repository context with **0% to 4% cache hit rate**, consuming **104,188,859 uncached raw tokens** in just 2 hours (16:00–18:00 KST).
* In contrast, the 9 long-running multi-turn conversations transferred to Account 2 only consumed **14.8M uncached tokens** (maintaining a 95.6% cache hit rate after turn 1).
* **OpenCodex Flaw**: Rather than throttling or warning that the workload was an un-cacheable runaway subagent storm, OpenCodex blindly cascaded the storm: Account 1 reached 80% -> poured into Account 2 (104M uncached) -> Account 2 reached 80% -> poured into Account 3 (108M uncached) -> Account 3 reached 80%.

### 3.2 Phase 2: The 80% Multi-Turn Ping-Pong Death Spiral (`src/codex/routing.ts`)
```typescript
// Conceptual bug pattern in routing logic (src/codex/routing.ts: reevaluateAffinityQuota)
const threshold = config.autoSwitchThreshold ?? 80;

if (usage >= threshold) {
    // BUG: Discards the 60s session affinity re-eval interval entirely!
    // ("// Over-threshold pins re-eval immediately so a depleted primary does not keep serving for up to 60s")
    // Instead of sticking with the account until a hard 429 is reached,
    // it selects whichever account has 1% lower usage on every single request.
    selectedAccount = pickLowerUsageAccount(config, entry.accountId, usage, now);
}
```
* Once accounts 1, 2, and 3 were all >= 80%, **the 60-second affinity lock was permanently bypassed on every request**.
* The main multi-turn sessions (e.g. `b36bfd009...` with 2,788 turns) were subjected to extreme ping-pong:
  * Turn N (Account 1): 150k context
  * Turn N+1 (Account 2): 150k context (0% cache on Account 2!)
  * Turn N+2 (Account 3): 150k context (0% cache on Account 3!)
  * Turn N+3 (Account 1): 150k context (Cache evicted or mismatched!)
* The multi-turn sessions, which should have cost ~7k tokens/turn, suddenly cost **150k raw tokens/turn**, multiplying token burn by **20x–25x** until every account returned HTTP 429.

---

## 4. Production Telemetry & Empirical Evidence

Data collected from full log audits of `usage.jsonl` (15,607 requests) and local Orca transaction databases (`logs_2.sqlite`):

### 4.1 Telemetry by Account (Masked)

| Account Hash | Account Identifier | Quota Reset Window | Requests | Tokens Burned | Uncached Tokens | Primary Workload Pattern |
| :--- | :--- | :---: | :---: | :---: | :---: | :--- |
| `p***e` | `orca-3***7` (`h***d@gmail.com`) | Sep 12 18:00 KST | 8,696 | 956,521,557 | 64,870,299 | Primary session host (93.2% cache) until 80% threshold |
| `p***f` | `orca-d***0` (`s***t@agentryx-ai.com`) | Sep 13 15:48 KST | 3,317 | 462,569,476 | 119,039,343 | Hit by 732 subagent calls (104M uncached) -> 80% |
| `p***a` | `orca-b***e` (`w***g@gmail.com`) | Sep 13 18:20 KST | 2,644 | 382,378,090 | 112,431,200 | Hit by 773 subagent calls (108M uncached) -> 80% |
| `p***d` | `orca-6***d` (`r***2@gmail.com`) | Sep 12 17:14 KST | 650 | 91,859,644 | 18,240,110 | Swept into 4-way ping-pong death spiral |
| `p***7` | `orca-f***e` (`m***y@gmail.com`) | Sep 13 09:09 KST | 4 | 0 | 0 | Failed immediately with HTTP 429 (pre-exhausted) |
| *Unassigned* | *Direct / Pre-routing* | - | 296 | 9,004,482 | 8,912,450 | Initial direct requests / zero-attempt drops |
| **Total** | **5 Accounts + Direct** | - | **15,607** | **1,902,333,249** | **323,493,402** | **Entire pool drained within ~13 hours** |

### 4.2 Cache Hit Rate Dynamics
* **Single-Account Stable Sessions**: ~95.6% – 96.0% cache hit rate (~7,800 uncached tokens/turn).
* **Subagent One-Off Requests**: ~4.0% cache hit rate (~142,000 uncached tokens/turn).
* **Cross-Account Ping-Pong (Above 80%)**: 0.0% handover cache hit rate (~150,000 uncached tokens/turn).

---

### 1. Mandatory Session Sticky Routing
* Active multi-turn sessions **MUST NEVER switch accounts mid-conversation**, regardless of whether the account is at 80%, 90%, or 95%.
* An account should only be abandoned for a session if OpenAI explicitly returns `HTTP 429: The usage limit has been reached`.

### 2. Subagent / One-Off Task Circuit Breaker
* OpenCodex should recognize high-volume, uncached single-turn bursts and avoid cascading them to fresh accounts automatically without user consent.
* Allow users to designate a specific "worker/agent account" rather than letting subagent bursts wipe out the entire pool.

### 3. Sequential Account Draining Instead of Load-Balancing
* For LLM caching architectures, load-balancing across accounts above 80% is an anti-pattern.
* Implement **Sequential Drain**: Dedicate 100% of traffic to Account 1 until 429 is reached, then switch to Account 2.

### LINKED ISSUE #4550: [Bug]: Existing Codex CLI thread bypasses pool over direct WebSocket while status reports opencodex-local (2.53.0)
### Client or integration

Codex CLI

### Area

Proxy and routing

### Summary

An existing Codex CLI thread repeatedly hits the native ChatGPT usage limit while OpenCodex 2.53.0 reports `routing=opencodex-local` and its two-account pool has a working secondary account. Thread-correlated Codex diagnostics establish that the failing requests connect directly to `wss://chatgpt.com/backend-api/codex/responses`, bypassing the pool.

This initially looked like quota/affinity failover, but the failing thread's actual destination changes the diagnosis. Please investigate the active-client routing/configuration lifecycle and whether status can distinguish configured routing from the route retained by a running client.

### Reproduction

Observed sequence; not yet a deterministic clean-start reproduction:

1. Use Codex CLI 0.154.0 with ChatGPT login and an existing long-running thread using `gpt-6-astra`.
2. Have OpenCodex 2.53.0 running locally with the canonical OpenAI provider in pool mode, two Plus accounts, quota strategy, and auto-switch threshold 95.
3. The primary account reaches its quota while the secondary has substantial quota remaining. Manual secondary selection successfully serves other proxied requests.
4. Run `ocx status`: live proxy health succeeds and routing is reported as `opencodex-local`.
5. Continue the existing CLI thread. It repeatedly returns `usage_limit_exceeded` instead of reaching the pool.
6. Correlate the thread's Codex SQLite diagnostics and session completion events: the Responses WebSocket destination is the direct ChatGPT endpoint. There is no corresponding proxy usage record for these turns.

Expected: the client uses the configured proxy route, or the integration clearly identifies that an existing client must be restarted/resumed to adopt it. A configured-route status should not be mistaken for proof that an active client is routed.

Actual: repeated direct upstream requests exhaust the native credential while a healthy local pool is available.

### Version

OpenCodex 2.53.0, Codex CLI 0.154.0. The latest upstream release checked at reporting time is v2.53.0.

### Operating system

Linux / WSL2

### Provider and model

OpenAI ChatGPT login, `gpt-6-astra`, low reasoning effort.

### Logs or error output

Sanitized extraction, with thread IDs, request IDs, account labels, headers, and credentials omitted:

```text
2026-09-13 19:07:50 UTC
  target: codex_api::endpoint::responses_websocket
  destination: wss://chatgpt.com/backend-api/codex/responses
2026-09-13 19:07:51 UTC
  same thread task_complete.error.codex_error_info: usage_limit_exceeded

2026-09-13 19:08:41 UTC
  target: codex_api::endpoint::responses_websocket
  destination: wss://chatgpt.com/backend-api/codex/responses
2026-09-13 19:08:43 UTC
  same thread task_complete.error.codex_error_info: usage_limit_exceeded

CLI: You've hit your usage limit.

ocx status, checked during investigation:
  Proxy: running
  Health: http://127.0.0.1:10100/healthz ok (live)
  routing=opencodex-local, service=installed-but-unhealthy, shim=absent
```

The service warning is real: the systemd unit is inactive, but a separate live 2.53.0 proxy process answers health checks and serves other traffic. Restarting the proxy previously did not resolve the affected CLI thread.

The affected CLI process was associated with the exact session via its open rollout file. Its environment has no `OPENAI_BASE_URL` or `CODEX_HOME` override, and its command line has no `base_url/model_provider` override. No raw diagnostics are attached because they can contain sensitive headers.

### Redacted configuration

Current root keys, verified by parsing the TOML:

```toml
experimental_realtime_ws_base_url = "http://127.0.0.1:10100/v1"
```

No explicit root `model_provider`. Relevant OpenCodex settings:

```json
{
  "autoSwitchThreshold": 95,
  "multiAgentMode": "v1",
  "providers": {
    "openai": {
      "authMode": "forward",
      "codexAccountMode": "pool"
    }
  }
}
```

`accountPoolStrategy` is omitted and defaults to quota in 2.53.0. `pool.cacheAffinity` is omitted. A secondary account is manually pinned; its identifier is intentionally omitted.

### Scope and remaining uncertainty

- Direct routing of the failed turns is confirmed. This report does not establish a defect in pool priority selection or quota failover.
- The current config is confirmed, but its exact contents when the affected process started are not established. Retained pre-injection configuration is a possible explanation. A fresh-process reproduction remains necessary before claiming Codex 0.154.0 ignores `openai_base_url`.
- Other proxied requests and manual secondary success do not prove this thread was proxied.
- Earlier 429 usage records belong to other requests and are deliberately excluded as evidence for this thread.
- The observed direct upstream error's wire representation is not established by the task-completion error alone.
- Related closed issue: #2411 concerned an absent override and missing routing status. Here the override exists and status explicitly says `opencodex-local`, while the active client's actual destination is direct.

### Checks

- [x] I searched existing issues and documentation/source relevant to routing.

lidge-jun:
## Summary

Two account-pool defects where the proxy's behaviour and its reporting disagreed with what the operator believed.

**#4546 — pool rotation destroyed the prompt cache of a bound conversation.** Under the `quota` strategy a live thread binding could move to any *strictly cooler* eligible account once its account crossed `autoSwitchThreshold`. "Strictly cooler" had no floor, so once every account sat in the 80-100% band the coolest was still over the threshold and a long-running conversation was handed from account to account on consecutive turns. Codex prompt caches are account-isolated, so each hop restarted from a cold prefix: the reporter measured 7k-token turns becoming 150k-token turns, and 1.9 billion tokens across 15,607 requests in about 13 hours on five accounts.

A live binding may now move only to an account that has genuine quota headroom **and** is strictly cooler than the bound account. Headroom alone is not enough, because `hasCodexQuotaHeadroom` deliberately answers true for unknown usage — correct for an unbound pick, a guess when a warm prefix is at stake — and `CODEX_UNKNOWN_USAGE_SCORE` is 101, so the second bar excludes an unobserved destination without a special case. Before: with a=95, b=90, c=97 the thread moved every turn. After: it stays, and movement is bounded by the number of accounts rather than the number of turns.

This is the same bar `reset-first` already applied through `hasCodexQuotaHeadroom`; `quota` was the asymmetric outlier. It narrows a *preference* and never a refusal — a 429/402 with no success since, a failover streak, pause, cooldown, a lost generation and an unusable account all still release the binding before the rule is consulted. Unbound selection is untouched, and `pool.cacheAffinity` keeps its meaning and its default as the stronger sticky-until-exhaustion opt-in.

**#4550 — `ocx status` presented configured routing as adopted routing.** It printed `routing=opencodex-local` read from config on disk, so an operator read it as proof their running Codex CLI was pooled. A client that started before the route was injected keeps dialing `wss://chatgpt.com/backend-api/codex/responses` directly, which is exactly what the report shows: repeated `usage_limit_exceeded` while a healthy secondary account sat in the pool, and no proxy usage record for those turns.

Status now carries adoption evidence. A running Codex client whose start time precedes the recorded injection cannot have read the injected route, so it is named as `pending-client-restart` with its pid and the summary says to restart it. `adopted` is deliberately an inference about *opportunity*, not an observation of live traffic, and the module comment names every false-`adopted` source this evidence does not cover — a client the matcher misses, a resumed thread holding an already-open direct WebSocket, `OPENAI_BASE_URL` overrides, same-second starts, an empty match set, a different `CODEX_HOME`, and surfaces such as `codex-code-mode-host` and VS Code extension hosts. Anything unverifiable is `unknown`: a failed process walk stays distinct from an empty one, because an empty walk means "no clients running" and would read as `adopted`.

The Codex CLI matching rules were private and reachable only through a count, which can answer "is Codex busy" but cannot name a stale pid, so they are now one exported predicate with two callers rather than a second copy. The new `routingAdoption` field is optional and the routing detail string keeps its old shape as a prefix, so the GUI and server consumers of `StartupHealth` are untouched and a stale client never changes `status`, `protection`, `rebootSafe` or `recommendedCommand`.

`structure/providers/openai-tiers.md` records the new rebind invariant, as `structure/manifest.json` ownership of `src/codex/` requires. Planning notes are in `devlog/_plan/260914_l2_pool_routing_cache/`.

## Verification

**Local suite runs were NOT RUN for this unit, by explicit instruction.** No `bun run test`, no bare `bun test`, no `bun run typecheck`, no `bun install`, no `bun run build:gui` — not locally, and not by any delegated agent. The proof for this pull request is hosted CI at the exact final head SHA and nothing else; a green run at an earlier head is not evidence for a later one.

- Hosted CI run: Cross-platform CI [34806827740](https://github.com/lidge-jun/opencodex/actions/runs/34806827740) at head `4b22b25f61c254cd29a823ed125ece5233b9289a` — conclusion **success**, all 24 checks green across Linux, macOS and Windows. That run is the only proof for this head.
- Regression coverage added instead of local execution, per AGENTS.md for `src/` behaviour changes:
  - `tests/codex-integration/codex-pool-rotation.test.ts` — four cases beside the existing cache-affinity tests. The first is the #4546 death spiral and fails without the fix: with a=95, b=90, c=97 the pre-fix rule handed the thread to whichever account was one point cooler on every resolve. The scores are deliberately unequal, since equal scores would not move even before the fix and the case would pass for the wrong reason. The other three keep the change a narrowing rather than a pin — a thread still moves onto an account with genuine headroom, a 429 still releases a binding the preference rule would have kept, and a fully spent account still yields to a sibling with headroom. Preview is asserted beside every resolve because both call sites share one helper.
  - `tests/service/autostart-health.test.ts` — adoption derivation through its pure and injectable seams: a pre-injection client is `pending-client-restart` with its pid, a post-injection client is `adopted`, a same-second start is not stale, enumeration failure and unreadable start times are `unknown`, a stale client outranks an unreadable one, non-`opencodex-local` routing is `not-applicable`, the detail string stays a prefix, and a stale client leaves the restart-safety classification unchanged.
- Reviewed by two independent reviewer rounds before and after implementation. Round 1 returned FAIL on three blocking findings: the planned Codex-client process set could not see a CLI process at all, `adopted` was not sound as written, and the write scope omitted the `structure/` doc that owns `src/codex/`. All three were fixed. Round 2 returned NEAR-PASS and verified the committed `pickCacheSafeQuotaReplacement` against every call site, finding no case that returns the wrong account and no weakened release path.
- Existing bound-rebind assertions were audited by hand rather than executed: every one of them moves to a destination with real headroom (b at 5%), so the narrowing keeps them green.

## Checklist

Docs: the `autoSwitchThreshold`, `accountPoolStrategy`, pool-routing prose and strategy table now state the headroom bar, in the English reference and in all seven translated locales. No credential, token or request content is read or logged; the adoption check reads process ids and start times plus the journal's own timestamp, and it never mutates the journal — it parses `JOURNAL_PATH` directly rather than calling `readJournal`, which can delete a corrupt file.

Closes #4546
Closes #4550


lidge-jun:
## 리뷰 · 우선순위 64 / 80

지금 `dev`(HEAD `c66709f31`)에서 #4546의 핵심은 그대로입니다. `quota` 전략이 바인딩된 스레드를 `autoSwitchThreshold` 넘김 + `pickLowerUsageAccount`(조금만 더 시원하면 OK)로 옮기고, `pool.cacheAffinity`는 여전히 `!== true`라 기본 OFF입니다. 같은 시각에 열린 #4580은 그 사고를 cache-first 기본 + transient hold + headroom floor로 막으려 하고, 이 PR(`codex/260914-l2-pool-routing-cache`)은 L2 레인으로 **문턱 eviction은 유지**한 채 목적지 headroom만 좁히는 쪽입니다. #4580 계획 문서가 이 레인의 `010_cache_safe_rebind.md`를 흡수한다고 써 둔 그 분석입니다.

이 PR은 두 덩어리입니다. (1) `src/codex/routing.ts`에 `pickCacheSafeQuotaReplacement`을 넣고 `previewReusableAffinityAccount` / `reevaluateAffinityQuota`가 `hasCodexQuotaHeadroom` + 엄격히 더 낮은 usage인 곳으로만 옮기게 합니다. 풀이 전부 80–99%면 아무도 안 옮겨서 #4546 ping-pong의 **절반**(목적지 하한)은 막힙니다. 하지만 `mayRebindAffinityForQuota`는 여전히 `cacheAffinity !== true`면 문턱만으로 재바인딩을 허용하고, 기본값을 ON으로 뒤집지도 않으며, transient streak가 바인딩을 지우는 구멍도 이 레인에 없습니다. 그래서 「플래그를 모르는 설치」는 85%→5%로 워밍된 prefix를 버릴 수 있습니다 — #4580이 1순위로 막는 바로 그 케이스입니다. (2) 새 파일 `src/codex/routing-adoption.ts` + `autostart-health` / `native-profile-processes` / `app-server-processes` 연동은 #4550용입니다. 주입 시각 이후 시작된 클라이언트만 `adopted`로 추론하고, 같은 초·열거 실패·미매칭 표면은 `unknown` / `pending-client-restart`로 남깁니다. 「adopted = 라이브 트래픽 관측」이 아니라고 주석에 못 박은 점이 좋습니다. types/config 스플릿에 무효화되는 PR은 아닙니다.

라인 - `pickCacheSafeQuotaReplacement`과 #4580의 `boundThreadReplacement`이 `src/codex/routing.ts` 같은 자리를 고칩니다. 둘 다 머지하면 충돌·이중 정책이 납니다. #4580이 랜딩하면 이 파일의 cache-rebind 절반은 superseded입니다.
라인 - #4546을 Closes로 걸기엔 이 레인만으로는 부족합니다. 기본 `cacheAffinity` OFF + transient eviction 미처리라서, 제목에 나온 「keep a bound thread's prompt cache」는 headroom floor만으로는 끝까지 못 지킵니다. 이슈 closes는 #4580에 맡기는 편이 맞습니다.
경로/심볼 - `routing-adoption.ts`의 `adopted`는 빈 매칭 집합을 vacuously adopted로 둡니다(주석에 명시). 운영 UI가 그걸 「전부 OK」로 보여 주면 #4550 오해가 다시 납니다 — 표시 문구에 observedClients=0 / unknown 가능성을 같이 실어야 합니다.
경로/심볼 - 두 관심사(풀 재바인딩 비용 + 라우팅 채택 보고)가 한 PR에 묶여 있습니다. cache 절반이 #4580에 흡수되면 adoption 절반(#4550)만 남은 채 브랜치가 무거워질 수 있습니다.

메인테이너의 판단이 필요한 지점
- #4546 정본을 #4580으로 두고 이 PR의 `routing.ts` 절반은 close(superseded)할지
- `routing-adoption` / autostart-health 절반(#4550)을 새 PR로 쪼개 살릴지, 이 브랜치에서 cache 커밋만 빼고 재기반할지
- 이 PR만 먼저 얇게 넣어 headroom만 막고 default flip은 #4580으로 이어갈지 (동시 진행 시 `routing.ts` 충돌 비용)

너의 추천
#4546은 #4580을 우선 머지(CI·스크린샷 게이트 정리 후). 이 PR(#4581)은 cache-rebind 커밋을 superseded로 정리하고, `routing-adoption.ts` + health/process 연동만 `dev` 최신 위에 새 브랜치/PR로 분리해 #4550을 다루는 쪽을 추천합니다. 지금 상태에서 둘 다 머지하지 말 것. rebase로 #4580과 싸우기보다 close-don't-rebase가 유지보수 비용이 덜합니다.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:

Exact-head evidence at 4b22b25f6: 30 successes, 2 skips. The two cancelled entries are label and enforce-target runs that later runs at the same SHA superseded with successes.

Reviewed independently before merge, and the review is worth recording because it is narrower than the pull request title suggests.

src/codex/routing.ts routes live quota rebinds through a cache-safe replacement check at both the affinity re-evaluation and the reusable-account preview, so a bound thread can move only to an account with real quota headroom that is also strictly cooler. That is what stops the all-hot ping-pong that was discarding the account-isolated prompt-cache prefix, and the new cases pin stay-put, a single move onto real headroom, release on 429, and exhaustion with a sibling. The spiral is genuinely prevented rather than made less likely. What is not done: a below-threshold sibling still takes the thread once, so the cache is lost once before it goes sticky, and unbound subagent cascading is untouched.

The pooled-status half is honest reporting, not transport interception. A pre-injection matched CLI is now named as pending-client-restart with its pid, adopted stays an inference, and the routing string is byte-identical for the adopted and unknown cases. So the printed claim is accurate when a stale CLI is found, and it is still not proof about live traffic when the matcher misses a process or a resumed WebSocket stays direct.

Because of those residuals I am leaving both issues open with the landed scope recorded on them rather than closing them against this merge.

PR #4585, opened 2026-09-14
fix(codex): resolve the Codex App runtime on Windows and stop a stale pin from overriding it

### LINKED ISSUE #4204: [Bug][Windows] Stale persisted CLI 0.135.0 strips max/ultra while Codex Desktop runs 0.153.4
## Summary

On Windows, OpenCodex 2.50.0 can strip `max` and `ultra` from the shared Codex catalog using a stale standalone CLI, while the running Codex Desktop app uses a newer runtime that supports both. Starting/syncing OpenCodex can therefore remove options that the actual Desktop runtime accepts.

The installed sources for `runtime.ts`, `catalog/bundled.ts`, and `catalog/effort.ts` were verified byte-for-byte against the corresponding Git blobs at upstream HEAD `2d4d7a22381a2e497c2442902104619e25f937c7` during this investigation.

## Client / Area / Environment

- Client: Codex App; area: Catalog / models, Windows runtime discovery.
- Windows 11 Pro, build 26200, x64.
- OpenCodex: `@bitkyc08/opencodex` 2.50.0.
- Installed MSIX: `OpenAI.Codex` 26.903.9818.0; app.asar package version: 26.903.71938.
- Persisted OpenCodex runtime: `%LOCALAPPDATA%\Programs\OpenAI\Codex\bin\codex.exe`, `codex-cli 0.135.0`, source `configured`.
- Running Desktop child runtime, verified through the process table: `%LOCALAPPDATA%\OpenAI\Codex\bin\<hash>\codex.exe`, `codex-cli 0.153.4`.
- Desktop parent executable: `%ProgramFiles%\WindowsApps\OpenAI.Codex_<version>_x64__<publisher>\app\ChatGPT.exe`.

## Reproduction And Fresh Proof

The existing `codex-runtime-clamp.json` records `removedEfforts: ["max", "ultra"]`, runtime version `0.135.0`, at `2026-09-10T12:01:09.525Z`. Affected rows include `gpt-6-astra`, the GPT-5.6 family, and routed Anthropic models.

A non-mutating probe imported the installed OpenCodex functions `supportedCodexReasoningEffortsFromObservedCatalog` and `clampCatalogModelsToObservedCodexSupport`. For each real executable it ran `--version` and `debug models --bundled`, then applied the derived support set to the same in-memory GPT-6 Astra row. It did not run live sync or rewrite the user's catalog.

| Runtime | Derived support | Removed | Surviving |
| --- | --- | --- | --- |
| persisted CLI 0.135.0 | low, medium, high, xhigh | max, ultra | low, medium, high, xhigh |
| active Desktop CLI 0.153.4 | low, medium, high, xhigh, max, ultra | none | low, medium, high, xhigh, max, ultra |

Reproduction shape, with `legacyCli` and `desktopCli` set to the two real executable paths:

```ts
import { execFileSync } from "node:child_process";
import {
  supportedCodexReasoningEffortsFromObservedCatalog,
  clampCatalogModelsToObservedCodexSupport,
} from "./src/codex/catalog/effort";

for (const exe of [legacyCli, desktopCli]) {
  const bundled = JSON.parse(execFileSync(exe, ["debug", "models", "--bundled"], { encoding: "utf8" }));
  const rows = [{
    slug: "gpt-6-astra",
    default_reasoning_level: "medium",
    supported_reasoning_levels: ["low", "medium", "high", "xhigh", "max", "ultra"].map(effort => ({ effort })),
  }];
  console.log(clampCatalogModelsToObservedCodexSupport(
    rows, supportedCodexReasoningEffortsFromObservedCatalog(bundled),
  ));
}
```

The user reports disappearing picker options after starting OpenCodex. The full UI/startup sequence was not replayed during this investigation, to avoid restarting an active Desktop session. At inspection time the on-disk catalog already contained `max`/`ultra` for Astra and Sol (last write `2026-09-10T12:55:35Z`), while the earlier clamp diagnostic remained. Thus that diagnostic is historical evidence, not a claim that the current file still lacks the options. The controlled probe reproduces the destructive projection itself.

## OpenCodex Source Trace

1. [runtime.ts:576](https://github.com/lidge-jun/opencodex/blob/2d4d7a22381a2e497c2442902104619e25f937c7/src/codex/runtime.ts#L576): resolution orders environment, persisted configuration, shim, PATH, fallback. A still-runnable persisted CLI wins even when another runtime is newer. This resolver does not reconcile its choice with the running Desktop child. `resolveAndPersistCodexRuntime` also persists automatically discovered selections, so `configured` alone does not establish that the user intentionally pinned an old runtime.
2. [catalog/bundled.ts:241](https://github.com/lidge-jun/opencodex/blob/2d4d7a22381a2e497c2442902104619e25f937c7/src/codex/catalog/bundled.ts#L241): bundled-catalog loading probes the single selected runtime and defaults `discoverAlternatives` to false.
3. [catalog/effort.ts:313](https://github.com/lidge-jun/opencodex/blob/2d4d7a22381a2e497c2442902104619e25f937c7/src/codex/catalog/effort.ts#L313): the effort vocabulary is derived from bundled bare native models.
4. [catalog/effort.ts:348](https://github.com/lidge-jun/opencodex/blob/2d4d7a22381a2e497c2442902104619e25f937c7/src/codex/catalog/effort.ts#L348): unsupported rungs are removed from rows; the list-level helper applies this across models.
5. [catalog/sync.ts:1945](https://github.com/lidge-jun/opencodex/blob/2d4d7a22381a2e497c2442902104619e25f937c7/src/codex/catalog/sync.ts#L1945): sync invokes the clamp. [cli/index.ts:488](https://github.com/lidge-jun/opencodex/blob/2d4d7a22381a2e497c2442902104619e25f937c7/src/cli/index.ts#L488) calls the startup sync gate.

## Why This Appears On Windows

The inspected Windows Desktop app and OpenCodex do not share executable discovery. Desktop runs a relocated, hash-addressed runtime under `%LOCALAPPDATA%\OpenAI\Codex\bin`, while OpenCodex retains a runnable older executable under `%LOCALAPPDATA%\Programs\OpenAI\Codex\bin`.

The installed Desktop `app.asar` was read directly, without modifying it. In `.vite/build/main-CMBCj4XL.js:1`, runtime resolution contains Windows-specific bundled-path and WindowsApps relocation handling (`Zr`, `Qr`, `$r`; `relocateWindowsApps`). This differs from OpenCodex's persisted/PATH resolver. The process table and both `--version` probes confirm that these paths actually differ on the affected machine.

This is a Windows packaging/runtime-discovery mismatch, not evidence that Windows cannot support these efforts. The clamp itself is platform-independent, and the same stale-runtime arrangement could affect other OSes. No macOS machine was tested, so strict Windows exclusivity is not established.

## Separate Desktop Display Conditions

The installed Desktop renderer also filters efforts independently of the catalog:

- `webview/assets/app-initial-f094ef01c64d.js:1`, `M2i`: intersects model effort rows with `enabledReasoningEfforts`, and removes Ultra when `includeUltraReasoningEffort` is false.
- The same bundle defines `enabled-reasoning-efforts`; its default list omits `max`. Ultra additionally has a feature gate in the model query (`1186680773`) and a `show-ultra-in-model-picker-slider` preference defaulting to false.
- `webview/assets/agent-settings-bfb770313a62.js:1`: the available-efforts settings depend on `hasModelSupportingMaxReasoningEffort` / `hasModelSupportingUltraReasoningEffort`. A clamped model list can therefore remove the settings choices themselves.

These are source-observed conditions, not a claim about the affected user's current feature-gate values. They explain why restoring catalog bytes alone is insufficient to guarantee visible controls. No Windows-only max/ultra exclusion was established in these inspected filter functions.

## Expected Behavior / Suggested Fix

For Desktop catalog synchronization, detect and report the consuming Desktop runtime and compare it with the selected OpenCodex CLI before destructively reducing the shared catalog. Distinguish explicit runtime pins from automatically persisted discovery. Account for Windows MSIX relocation and process identity; do not simply probe a protected WindowsApps path or silently replace an intentional pin.

Keep old-CLI compatibility handling where it is actually needed. Do not unconditionally enable arbitrary efforts: a newer Desktop and an older CLI may share the catalog, so runtime-specific projections or an explicit mismatch/recovery flow may be necessary. Diagnostics should separate runtime mismatch, catalog clamp, stale Desktop model cache, and renderer visibility preferences/gates.

Regression coverage should include a runnable legacy configured CLI beside a newer MSIX Desktop child, and verify that Desktop synchronization does not silently use the legacy ladder.

## Related / Checks

- Related closed issue #297 covers the general bundled-ladder clamp; this report adds a reproduced Windows stale-runtime selection mismatch on 2.50.0.
- Related closed issue #2292 covers stale Windows Desktop model caches. Cache staleness alone cannot explain the recorded clamp or the controlled probe, but remains a separate possible cause of current UI state.
- Existing issues and official configuration documentation were checked. Public documentation alone does not resolve this installed-build runtime mismatch.
- No credentials, account data, user-specific absolute paths, or full proprietary app sources are included.

### LINKED ISSUE #4458: [Bug][Windows] Prompt probe misses Codex App runtime and base prompt source
## Client or integration

OpenCodex dashboard

## Area

Platform (Windows / macOS / Linux)

## Summary

On Windows, the Codex prompt probe can report `Codex program not found` even when Codex is installed by the Codex App. The installed executable is under the stable product root `%LOCALAPPDATA%\OpenAI\Codex\bin\<changing-version>\codex.exe`, while the prompt probe only checked fixed POSIX paths. The probe also did not expose the selected model's published base prompt in its response.

## Reproduction

1. Install Codex on Windows through the Codex App.
2. Start the OpenCodex service in the normal Windows service flow.
3. Call the prompt-text probe through the existing Codex prompt management endpoint.
4. Observe `Codex program not found`, although the Codex executable exists below `%LOCALAPPDATA%\OpenAI\Codex\bin`.
5. The response has no selected-model base prompt source or representation metadata.

Expected:

- The probe should resolve the installed Codex executable using the shared runtime-discovery rules, including the Windows installation root and the inherited Windows PATH.
- The prompt response should identify the selected model and source file, return the complete published base prompt, and distinguish expanded text from an unexpanded template.
- Probe failures should distinguish program not found, unsupported command, execution failure, and invalid output.

Actual:

- The probe reports repeated `path does not exist` candidates and stops before invoking the installed Windows Codex executable.
- The base prompt is absent from the probe response even when the selected model catalog contains `base_instructions`.

## Version

- OpenCodex: local production reproduction on 2.52.0; candidate verification is based on current `dev` at `dc33113a9e6f5597ef9b2eecd692d7dc4b28dbd3` (package version 2.53.0).
- Codex CLI: `codex-cli 0.154.0-alpha.6.2`.

## Operating system

Windows (local reproduction; exact version omitted for privacy)

## Provider and model

Not provider-specific.

## Logs or error output

- Codex is installed below `%LOCALAPPDATA%\OpenAI\Codex\bin\<changing-version>\codex.exe`.
- The selected model catalog contains a `base_instructions` entry for the active model.

No screenshot is included because this change does not modify the GUI.

## Redacted configuration

- No account tokens, user names, or machine-specific absolute paths are included here.

## Checks

The candidate fix was verified locally without a paid model request. Focused backend tests, type checking, privacy scanning, and diff whitespace checks passed. The broad changed-test run timed out at the local 240-second runner limit with 8,692 passed and 172 failures across unrelated provider/network, account-pool, management/API, and lab suites; it is not represented as upstream CI evidence. The candidate's prompt-probe and runtime tests passed in that run. This report is limited to the backend prompt-probe and runtime path; it does not propose GUI changes.

lidge-jun:
## Summary

On Windows the Codex prompt probe reported `Codex program not found` on machines where Codex was plainly installed, and a stale persisted CLI kept overriding a newer runtime that was sitting right there.

**The prompt probe could not see the Codex App (issue 4458).** `prompt-text-probe.ts` carried its own resolver that checked four hardcoded POSIX paths. The Codex App installs `codex.exe` under `%LOCALAPPDATA%\OpenAI\Codex\bin\<version>\`, which that list can never match, so the probe walked its candidates, rejected every one with `path does not exist`, and gave up before reaching the installed executable.

The probe now asks the shared runtime resolver, and the resolver learned two things it needed:

- A new `installed` source enumerates the Windows Codex App bin root newest-directory-first, with a name tie-break so the order is deterministic when two directories share an mtime. POSIX keeps the same four paths the probe used to hardcode, so nothing that resolved before stops resolving. It ranks after PATH, so PATH stays authoritative.
- `probeVersion: false` selects a spawnable candidate without running `codex --version`. A request-path probe needs a command it can spawn, not a version, and ~1s of blocking exec per candidate is what made it report an absent candidate instead of a deferred inspection. That deferred selection gets its own memo and never publishes into runtime authority, because `peekCodexRuntimeProcessCache` feeds convergence and the bundled catalog, which would read a null version as "unknown version".

The probe also spawns through `codexExecInvocation` so a resolved Windows `.cmd` launches through `cmd.exe`, and reports a stable `failure.kind` (`program-not-found` / `command-unsupported` / `execution-failed` / `output-invalid`) instead of only a prose sentence. An ENOENT from the version probe is now classified program-not-found rather than a generic `--version` failure, so a missing program stops looking like a broken one. Process stderr is read only to classify and never returned, and the reported runtime path goes through `displayCodexRuntimePath` because this response is served over the management API and a Codex App path contains the account name.

**A stale auto-discovered runtime kept winning (issue 4204).** `codex-runtime.json` recorded command, source and version, but not how the record got there. `resolveAndPersistCodexRuntime` writes every automatically discovered selection, so a `configured` entry proved nothing about operator intent — and resolution stuck to it even when a newer runtime was present. On the reporting machine a still-runnable `codex-cli 0.135.0` kept beating the 0.153.4 the Codex App was actually running, and the catalog derived its reasoning ladder from the older binary.

- `PersistedCodexRuntimeState` gains `origin: "pinned" | "discovered"`. `resolveAndPersistCodexRuntime` writes `discovered`; a direct `persistCodexRuntime` call, which is how `doctor --fix` selects, stays `pinned`. A record with no origin reads as discovered, because auto-discovery wrote every one of them and reading it as a pin would leave the bug unfixed on exactly the installs that have it.
- An unpinned record hands over only to a strictly newer valid candidate. A pin is never touched, equal versions stick, and an unknown version on either side is not evidence of an upgrade. The handover is reported as `supersededDiscovered`, kept separate from `replacedConfigured` so "superseded" never reads as "gone".
- The no-discovery fast path still probes the bounded Codex App roots when the persisted record is unpinned. That path is what the catalog's bundled loader uses, so without it the comparison could never run where the bug actually bites. PATH-wide discovery stays off, and a pinned record skips even the bounded scan.

No config-schema field was added; `codex-runtime.json` is runtime-owned state.

### Carried contributor work

This carries PR #4461 by @S0RYUASUKA, with a `Co-authored-by` trailer on the first branch commit. Carried as designed: the `installed` runtime source and its newest-mtime ordering, the deferred `probeVersion` selection, the isolated non-authoritative memo, the ENOENT classification, and the resolver swap in the prompt probe. Reimplemented against current `dev`: the deferred memo is keyed off the same `resolveCacheKey` the rest of the file uses rather than a parallel key, and the new `readdirSync`/`statSync` dependencies are injected and registered in that key's injection guard, which the original patch did not do — without it an injected test listing would poison the process memo for every later test in the file.

Not carried, and deliberately out of scope this round: the base-prompt-source reader (`BasePromptText`, catalog and `model_instructions_file` reading), the management route shape, and the Windows service PATH wrapper. Those are separate surfaces owned elsewhere, so this PR does not claim the "base prompt source" half of issue 4458.

### What issue 4458 still leaves open

The probe now resolves and spawns the Codex App runtime, which is the `Codex program not found` half. The report also asks for the selected model's published base prompt in the probe response; that is not in this PR. Tracked as follow-up work on the prompt-source surface.

## Verification

- `bun run test` / `bun test` / `bun run typecheck` / `bun install` — **NOT RUN.** This branch was developed under an explicit no-local-suite constraint. Hosted CI at the exact final head is the only proof claimed here.
- `git diff --check` — clean.
- Scope check against the merge base: no change to `src/config.ts`, `src/types/config.ts`, `src/types/provider.ts`, `src/server/`, `src/service/`, `src/web-search/`, or `gui/`.
- Focused regression tests added beside the existing subsystem tests. `tests/codex-integration/codex-runtime.test.ts`: Windows App discovery from an injected listing, deterministic equal-mtime ordering, deferred selection with no exec and no published authority, ENOENT classification, PATH outranking an installed candidate, the POSIX locations still resolving, the 0.135.0-versus-0.153.4 handover with and without a pin, equal and unknown versions sticking, the origin each writer records, parse acceptance of a missing origin and rejection of a junk one, and the bounded App-root scan on the no-discovery path. `tests/codex-integration/codex-prompt-text-probe.test.ts`: a resolver-found runtime is spawned rather than reported missing, a resolver that finds nothing yields `program-not-found` with zero spawn attempts, unparseable output yields `output-invalid`, an unknown subcommand yields `command-unsupported`, and no failure detail echoes process stderr.
- No GUI files changed, so no screenshot applies.

Closes #4204
Closes #4458

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added automatic Codex runtime discovery for Windows installations and common Unix locations.
  * Newer automatically discovered installations can replace older discovered versions, while manually selected runtimes remain unchanged.
  * Added deferred runtime version checks to reduce startup delays.

* **Bug Fixes**
  * Improved prompt probing with clearer failure reporting for unavailable, unsupported, failed, or invalid commands.
  * Runtime paths and sensitive error details are now redacted from probe results.


lidge-jun:
## 리뷰 · 우선순위 72 / 80

지금 `dev`(HEAD `8c7f01451`, #4580 cache-affinity 기본 ON + transient hold 직후)에서 Codex 런타임 해석은 `src/codex/runtime.ts`의 `resolveCodexRuntime` / `resolveAndPersistCodexRuntime`이 맡고, 프롬프트 프로브는 `src/codex/prompt-text-probe.ts`에 **따로 박아 둔 POSIX 네 경로**만 봅니다. Windows Codex App은 `%LOCALAPPDATA%\OpenAI\Codex\bin\<version>\codex.exe`에 깔리는데, 그 목록에는 절대 안 들어가서 프로브가 `path does not exist`만 반복하다 `Codex program not found`로 끝납니다. 그게 열린 이슈 #4458의 「프로그램 못 찾음」절반입니다. 같은 축에서 `codex-runtime.json`은 command/source/version만 남기고 **어떻게 그 기록이 생겼는지**는 안 남깁니다. `resolveAndPersistCodexRuntime`이 자동 발견을 계속 쓰는데도, 한번 잡힌 옛 `codex-cli 0.135.0`이 그대로 「configured」처럼 붙어 있으면 옆에 있는 Codex App 0.153.4보다 이깁니다. 카탈로그 reasoning ladder도 그 옛 바이너리에서 나옵니다 — 그게 #4204입니다. 비용 가드(#4580/#4581)와는 다른 축이고, 지금 `dev`만으로는 Windows App 설치를 프로브·카탈로그 둘 다 제대로 못 봅니다.

이 PR(`codex/260914-l6-codex-runtime-windows`, 헤드 `6817fa467`)는 두 구멍을 같이 막습니다. 첫째, `runtime.ts`에 `installed` 소스를 넣고 Windows App bin 루트를 mtime 최신 우선(이름 tie-break)으로 열거하고, POSIX는 예전 프로브가 쓰던 네 경로를 그대로 둡니다. 순위는 PATH 다음이라 PATH가 권위입니다. `probeVersion: false`면 `--version`을 안 돌리고 spawn 가능한 후보만 고르며, 그 결과는 **별도 deferred 메모**에만 넣고 `peekCodexRuntimeProcessCache`/번들 카탈로그 권위로는 안 올립니다(null version = unknown으로 읽히는 구멍). 프로브는 공유 리졸버를 묻고 `codexExecInvocation`으로 Windows `.cmd`를 `cmd.exe` 경유 spawn하며, 실패는 `program-not-found` / `command-unsupported` / `execution-failed` / `output-invalid`로 나눕니다. stderr는 분류에만 쓰고 응답에 안 실으며, 경로는 `displayCodexRuntimePath`로 마스킹합니다. 둘째, `PersistedCodexRuntimeState`에 `origin: "pinned" | "discovered"`를 둡니다. `resolveAndPersistCodexRuntime`은 `discovered`, `doctor --fix`의 직접 `persistCodexRuntime`은 기본 `pinned`. origin 없는 옛 파일은 **discovered로 읽습니다**(핀으로 읽으면 #4204가 그 설치에서 그대로 남음). 언핀만 엄격히 더 새 valid 후보로 넘기고, 그 핸드오버는 `supersededDiscovered`로 `replacedConfigured`와 분리합니다. `discoverAlternatives: false` 패스트 패스도 언핀이면 유계 App 루트만 추가로 프로브합니다(카탈로그 bundled loader가 쓰는 경로). `src/config.ts` / `src/types/config.ts`는 안 건드렸고, `codex-runtime.json`은 런타임 소유 상태입니다. 기여자 PR #4461(@S0RYUASUKA)의 installed/deferred/ENOENT/프로브 리졸버 스왑을 가져왔고 Co-authored-by가 첫 커밋에 있습니다. 본문이 명시한 대로 base-prompt-source·management route·Windows service PATH wrapper는 이번 범위 밖입니다. types/config 스플릿에 통째로 무효화되는 PR이 아닙니다.

경로/심볼 - PR 본문과 GitHub `Closes #4458`이 이슈 전체를 닫습니다. 그런데 #4458 제목·본문은 「Codex App 런타임 해석」과 「선택 모델 base prompt source」**둘**을 요구하고, 이 PR 본문도 base prompt 절반은 아직 열려 있다고 적습니다. 그 절반은 열린 #4461(`fix(codex): read base prompt source on Windows`) 쪽입니다. 이대로 머지하면 #4458이 반만 고치고 닫힙니다. `Closes #4458`을 빼거나, 이슈를 「프로그램 못 찾음」만 닫고 base prompt는 #4461/#4458을 분리·남기는 쪽이 맞습니다. #4204 closes는 이 PR 범위와 맞습니다.

경로/심볼 - #4461이 아직 OPEN이고 `src/codex/runtime.ts` / `src/codex/prompt-text-probe.ts` / 같은 테스트 파일을 겹쳐 만집니다. 이 PR이 랜딩하면 #4461의 runtime/probe 「프로그램 찾기」절반은 superseded입니다. base-prompt·route·taskxml 절반만 남기거나, 이 헤드 위에 리베이스한 뒤 남은 절반만 새 커밋으로 좁혀야 합니다. 그대로 같이 머지하면 충돌·이중 수정입니다.

라인 - `src/codex/runtime.ts`의 `resolveAndPersistCodexRuntime` `selectionUnchanged`는 command/source/selectedVersion만 비교하고 `origin`은 안 봅니다. 선택이 그대로인 옛 파일(origin 없음)은 디스크에 `discovered`를 다시 안 씁니다. 읽기 쪽은 missing=discovered라 #4204 수정은 유지되지만, 운영자가 파일을 열어 보면 origin 필드가 영원히 없을 수 있습니다. 선택이 같고 origin만 비었을 때 한 번 stamp하는 편이 관측에 낫습니다(필수는 아님).

경로/심볼 - Verification이 로컬 `bun test` / typecheck / install을 의도적으로 안 돌렸다고 명시합니다. 이 시각 기준 hygiene·changes·enforce-target·api usage·keyring macos 등은 초록이고, test 샤드·windows·gates·docker·npm-global·macos는 pending입니다. 이 변경의 핵심 재현면이 Windows App 루트 enumeration + `.cmd` spawn이라, **Windows 샤드/키링까지 초록인 이 헤드 SHA**가 머지 증거가 되어야 합니다. 제목의 「stale pin」은 실제로는 stale **discovered**(자동 기록)를 말하는 것이고, true pin(`origin: "pinned"`)은 건드리지 않습니다 — 릴리즈 노트/이슈 코멘트에서 그 구분을 한 줄 박아 두면 #4204 독자가 덜 헷갈립니다.

메인테이너의 판단이 필요한 지점
- #4458을 이 PR만으로 닫을지, 아니면 program-not-found만 닫고 base prompt는 #4461(또는 후속 이슈)에 남길지
- #4461을 이 랜딩 직후 `Landed via #4585`로 runtime/probe 절반 superseded 처리할지, base-prompt만 남긴 새 PR로 갈지
- origin 없는 기존 `codex-runtime.json`에 `discovered`를 한 번 stamp할지(관측), 아니면 missing=discovered 읽기만으로 충분하다고 둘지
- Windows CI가 아직 pending인 상태에서 머지할지, Windows 샤드 초록까지 기다릴지

너의 추천
호스트 CI가 이 헤드에서 test/windows/gates까지 전부 초록이면 squash merge into `dev`. 머지 전에 `Closes #4458`을 조정해 base prompt 절반을 열린 채로 두세요(#4204는 closes 유지). 머지 직후 #4461에 `Landed via #4585 at <commit>` + runtime/probe superseded 정리(base-prompt·route만 남기면 됨). CI가 Windows enumeration/spawn 쪽에서만 빨개지면 `src/codex/runtime.ts` / `src/codex/prompt-text-probe.ts` / `tests/codex-integration/codex-runtime*.ts`만 고치고 같은 브랜치에 push — types/config 스플릿이나 다른 레인으로 범위 넓히지 말 것. 우선순위 높음: #4204/#4458(프로그램 절반) Windows 실화이고 지금 `dev`에 아직 없음. types/config 스플릿 때문에 close-don't-rebase 할 대상 아님.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
Maintainer self-integration on `dev` under the MAINTAINERS.md policy, recorded with exact-head CI evidence.

- Head merged: `e584fb270221840cf4144b8bd044d337be3ea7da`
- Cross-platform CI run [34811180612](https://github.com/lidge-jun/opencodex/actions/runs/34811180612) — **success** at that exact SHA. All suite shards (test 1-4/4), macOS 1/2 and 2/2, gates, storage policy, api usage, docker smoke, keyring on all three platforms, and npm-global on all three platforms passed. The Windows shards and macOS control are conditionally skipped by the workflow.
- Enforce PR target branch, PR hygiene, PR Labeler, and React Doctor all green at the same SHA.
- No local suite was run for this branch; hosted CI at this head is the only verification claimed.

The previous head `6817fa467d` failed one test, `PATH still outranks an installed candidate when both are valid`, on macOS 1/2 and test 1/4. The fixture used a Windows-style `C:\on-path` PATH entry, which splits on `:` on the POSIX runners, so no PATH candidate matched and the installed runtime won instead. `e584fb2702` moves that fixture to a colon-free entry; nothing in `src/` changed between the two heads.

No outstanding maintainer objection. Security review is tracked separately and is not claimed here.

lidge-jun:

Exact-head evidence at e584fb270: 29 successes, 2 skips, no failures. An earlier head was red on a new test that wrote a Windows PATH into a Linux shard, where the path split on the wrong separator and the App binary won; that is fixed at this head.

Reviewed independently before merge. src/codex/runtime.ts adds an installed-runtime source, a deferred memo that never publishes into the process cache, and a pinned-versus-discovered origin on the runtime record, so an unpinned stale 0.135.0 entry yields to the 0.153.4 that is actually present instead of stripping max and ultra from the ladder. src/codex/prompt-text-probe.ts moves onto that resolver with a redacted display path, stable failure kinds, and classified stderr that is never returned verbatim.

The lane respected the round's schema constraint: no field was added to src/config.ts or src/types/config.ts, which another lane owned this round.

Scope of the claim, corrected on the way in. The description says it closes both assigned issues. It closes the stale-CLI one. The Windows prompt-probe issue is only half fixed — program-not-found and deferred spawn are handled, the published base-prompt source is not — so that issue stays open with the landed half recorded on it rather than being closed against this merge. Since pull requests here target dev, GitHub does not auto-close either one, so the correction costs nothing.

Thanks to @S0RYUASUKA, whose PR #4461 is carried in with a Co-authored-by trailer on a branch commit.

PR #4637, opened 2026-09-14
feat(responses): put combo hops and adapter inner retries on the shared send budget (#4546)

lidge-jun:
## Summary

Combo fan-out and the adapter inner retries now draw from the per-request send budget. A combo child already inherited the request's send counter, but nothing read it as a limit across targets, so target 1 could spend the budget and target 2 still drew a fresh full policy allowance — which is why a three-target combo measured twelve model sends.

Sharing the counter alone was not enough either. With one request-wide base allowance of three, a first target that 5xx-streaks leaves `remainingBaseSends` at zero, and the second target's initial send would be handed `attempts: 0` — a refusal, not a hop. A combo therefore needs a per-target policy rather than the single-target account-move profile, whose `maxTargetTransitions: 1` would refuse the second hop of a three-target combo outright.

`handleComboResponses` derives its policy from the combo definition. Declared transitions are `targets - 1`, and the total is capped at the first target's full ladder plus one send per further declared target plus the one shared final-recovery reserve, so a three-target combo is bounded at six model sends and a one-target combo reduces to the guarded profile exactly. Each target runs on a scope whose base allowance is `used + 3`, clamped so one send is held back for every target still declared after it: an early target cannot eat the send the last target is entitled to. All scopes share one `used` counter through an accessor onto the parent, because the budget factory reads that field back off its own object and a copied number would run a ladder against a stale total.

Kiro and Cursor could not be counted before this because both report one send per adapter call while retrying internally — Kiro reaches roughly eighteen sends per call, about thirty-six with the text fallback. `AdapterFetchContext` gains an optional `onPhysicalSend` that Kiro raises for every physical send across its throttle loop, endpoint fallback and reset ladder. Cursor is a `runTurn` adapter and never sees that context, so the budget reaches its transport through an optional `IncomingMeta.sendBudget`, and `runCursorTurnWithRetry` admits each attempt before a transport is built. Every new field is optional and unlimited when absent, so a context-free or meta-free adapter unit test behaves exactly as before.

Stacked on #4634.

## Verification

**Not run, by explicit instruction:** the local suite, `bun run typecheck`, `bun install`, and any build. The verification posture for this unit is hosted CI at the exact final head SHA and nothing else; this push used `--no-verify`.

[...]

PR #4592, opened 2026-09-14
feat(codex): record why a live binding was kept, moved, or released (#4546)

lidge-jun:
## Summary

Completes the last unmet requirement of the #4546 cost-guard work: a live-binding move is now recorded as an event with its cause, instead of something an operator has to infer from account labels across log lines. That inference is how #4546 had to be diagnosed in the first place.

`logCtx.affinity` already existed as a type and was already persisted by `appendUsageEntry`, but nothing ever assigned it, and routing had no reason to report. Three pieces:

- `resolveCodexAccountForThreadDetailed` now returns a `CodexAffinityDecision` alongside the account: what happened (`reused`, `held`, `detour`, `rebound`, `new_bind`, `cleared`) and why (`healthy`, `quota_headroom`, `quota_refusal`, `transient`, `transient_hold_expired`, `unusable`, `generation`, `expired`, `model_lane`). The release reason is captured at the point the binding is dropped and carried into the selection that replaces it, so the request that pays for a cold prefix is the one that says what it paid for.
- The pool auth context carries the decision to the request path.
- `logCtx.affinity` and a new `logCtx.affinityReason` are assigned where the provider log label is already derived from the auth context, and both are persisted on the usage entry.

Both lanes report: the ordinary binding and the model detour. `held` is the case worth having a name for — the thread was served by its own account while something transient was wrong with it, which is different from an ordinary healthy reuse and different again from a detour.

Follow-on to #4580, #4588 and #4589. Plan: `devlog/_plan/260914_cost_guard_stabilization/030_move_reason_evidence.md`.

## Verification

- **No local suite, typecheck, install or build was run**, by explicit instruction. Hosted CI at the exact final head SHA is the only proof.
- One regression in `tests/codex-integration/codex-pool-rotation.test.ts` walks a single thread through four decisions and asserts each one: first placement is `new_bind`, a healthy second request is `reused`, a 5xx streak is `detour` onto another account, and a 429 is `rebound` with reason `quota_refusal`.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults.

## Summary by CodeRabbit

- **Improvements**
  - Request logs now capture account-routing outcomes, including reused, replaced, deferred, or newly assigned accounts.
  - Logs include the reason for each routing decision, such as quota limits, transient issues, or healthy account availability.
[...]

PR #4638, opened 2026-09-14
feat(usage): report sends, spend and cache provenance per logical request (#4546)

lidge-jun:
## Summary

An operator can now read what one logical request actually cost. `sendCount` already counted physical sends per attempt, which answers the wrong question: a turn that failed over and fanned out across combo targets is one user turn, and nothing summed it.

A request log row now carries `logicalRequestId` and a `spend` record aggregating every attempt under that turn, combo children included, split into `settled` — sends whose attempt reached a terminal status — and `unresolved`, an attempt abandoned in flight or a budget charge no attempt row accounted for. Unresolved spend is never folded into settled, because an unexplained send is exactly the quantity this record exists to expose. `reserved` and `policyVersion` report the request execution budget's final state, and `moveReasons` keeps every pool-binding move that discarded a warmed cache prefix rather than only the last. `/api/usage` totals these as `sends`, `settledSends`, `unresolvedSends` and `spendRequests`.

Cache detail is now qualified by provenance rather than read as measurement. Strict-client normalization emits zero-default token-detail objects on every bridged wire, so a `cached_tokens: 0` recovered from a parsed wire proves nothing. `cacheProvenance` keeps `observed`, `synthesized` and `unknown` distinct from capture through persistence to the summary, only observed input tokens reach the `cacheHitRate` denominator, and the summary counts the three provenances separately instead of averaging an unknown as a zero. Rows written before the field existed are reconstructed from their own shape so historical readings do not move.

Two defects are fixed alongside. Affinity was set on the row and omitted from the field-by-field disk projection in `addRequestLog`, so every move was lost at the next restart — the same whitelist trap #4592 hit one layer up. And a no-account selection handed its cause to whatever later resolve happened to succeed, which for a pool that stays exhausted is never; reasons are now held per thread and model lane, consumed by that lane alone, and attached to the failing request's own row.

Stacked on #4637.

## Verification

**Not run, by explicit instruction:** the local suite, `bun run typecheck`, `bun install`, and any build. The verification posture for this unit is hosted CI at the exact final head SHA and nothing else; this push used `--no-verify`.

New coverage: `tests/server/request-spend-instrumentation.test.ts` and `tests/usage/usage-spend-cache-provenance.test.ts`, both registered in `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`. One pinned assertion in `tests/usage/usage-summary.test.ts` is updated with its reason: it divided a single measured row's cache reads by 256 rows that never measured cache.

Known open: the producer for `logCtx.executionBudget` lives at the ingress in `src/server/responses/core.ts`, outside this lane's write set, so until it is wired the budget-derived fields stay absent while `spend.sends` remains live from attempts alone. No UI column was added and `gui/` is untouched, so the Logs attempts table still has nowhere to render this.

## Checklist


lidge-jun:
## 리뷰 · 우선순위 64 / 80

이 PR은 스택 문서의 **wpg(spend instrumentation)** 레이어입니다. base가 `codex/4546-wpb-combo-adapter-retries`(#4637)라서, 지금 `dev`(627274b8f)만 보면 아직 없는 콤보·어댑터 전송 숫자를 집계 쪽에 올리는 작업입니다. #4546이 남긴 「논리 요청당 전송·지출·캐시 출처를 운영자가 읽을 수 없다」를 닫으려는 마지막 관측 층이고, wpc~wpb가 만든 숫자를 읽기 좋게 묶는 자리입니다.

지금 `dev`의 `sendCount`는 시도(attempt) 단위입니다. 콤보 페일오버·어댑터 안쪽 재시도가 한 사용자 턴에 묶여도 합산 행이 없었습니다. 이 PR은 요청 로그에 `logicalRequestId`와 `spend`(settled/unresolved/reserved/policyVersion)를 두고, unresolved를 settled에 섞지 않으며, 캐시 provenance와 moveReasons를 같이 남깁니다. `src/usage/log.ts`·`summary.ts`·management 공유 스키마·구조 문서·단위 테스트가 한 세트입니다.

가치는 큽니다. 다만 wpg는 「앞 레이어가 만든 것을 보고한다」가 설계 전제라서, wpb가 CI에서 `[3,3,3]`으로 아직 틀리면 여기 집계도 틀린 숫자를 예쁘게 보여 줄 뿐입니다. 호스티드 CI도 test·macos·enforce-target이 빨간불이고, base가 `dev`가 아니라 스택 tip이라 enforce-target 정책과도 겹칩니다. 로컬 스위트는 돌리지 않은 스택 공통 자세입니다.

라인 src/usage/log.ts (PersistedRequestSpend / unsettled 분해) - settled+unresolved≤sends 불변식을 파서에서 거르는 방향은 맞습니다. 다만 Cursor처럼 cap은 걸려도 `sendCount`가 1로 남는 경로(#4637 known open)가 있으면 settled가 과소 보고됩니다. wpg 문서/테스트에 「어댑터 미관측 전송은 unresolved가 아니라 그냥 누락」인지 한 줄로 밝혀 두는 편이 좋습니다.

경로 structure/gui-and-management-api.md - API 스키마를 문서에 맞춘 것은 좋습니다. GUI가 아직 이 필드를 안 그리면 「운영자가 읽는다」는 요약과 실제 화면 사이에 간격이 남습니다. 이번 범위 밖이면 후속 이슈 번호만 박아 주세요.

경로 CI enforce-target - base가 `dev`/`preview`가 아니라 wpb 브랜치라 게이트가 실패할 수 있습니다. 스택 PR 특성상 예상 가능한데, 랜딩 전에 base를 순차적으로 `dev`로 옮기는 기차만 지키면 됩니다.

메인테이너의 판단이 필요한 지점
- wpb(#4637) 녹색 없이 wpg를 먼저 머지할지(관측만 선행) 말지(스택 문서 순서)
- Cursor 미관측 전송을 unresolved로 칠지, wpb 후속으로 미룰지
- types/config 분할과 무관. 중복 이슈 아님

너의 추천
#4637이 `[3,2,1]`·CI 녹색을 증명한 뒤에 이어서 머지하세요. 지금은 리뷰·설계만 인정하고 랜딩은 보류가 맞습니다. 머지 시 `dev`에 올린 뒤 epic #4546 남은 관측 항목을 이 PR SHA로 체크하면 됩니다.

이 댓글은 grok-bot이 작성했습니다

Vivamisu:
Follow-up question on preserving cache evidence before wire normalization (observed on v2.58.0).

Thank you for introducing cache provenance here. There still appears to be an observability gap on the parsed, bridged Chat Completions response path: a zero-default `cached_tokens` recovered after normalization cannot distinguish an upstream explicitly reported zero from an absent upstream cache counter. The PR description already acknowledges that a parsed-wire zero proves nothing.

In local v2.58.0 request records, successful HTTP 200 requests intermittently have `cachedInputTokens: 0` with `cacheProvenance: "synthesized"`, interspersed with positive cache counts marked `observed`. This is not evidence of cache eviction, an upstream caching failure, or incorrect billing. Original upstream usage was not retained, so those historical zeros cannot be classified retrospectively.

Could a follow-up preserve cache-counter presence and its original numeric value at the upstream usage parsing boundary, before compatibility defaults are inserted, and carry that evidence through to persisted request/attempt records?

Desired semantics (field names are up to the project):
- Explicit upstream `cached_tokens: 0`: observed zero.
- Missing upstream cache counter: unknown, not measured zero.
- Positive upstream cache counter: observed value.
- Compatibility output may still emit zero where strict clients require it, without overwriting the original provenance used for logs and summaries.

For per-request presentation, an unobserved value should be shown as unknown / unavailable rather than an apparent measured 0%. Only usage field presence/numeric counters need to be retained; no prompts, response text, credentials, or raw request bodies are needed.

Would you prefer tracking this as a follow-up to #4638 or a separate issue? If an existing path already preserves this distinction for bridged Chat Completions, a pointer would be appreciated.

PR #4639, opened 2026-09-14
fix(codex): stop retrying a doomed pool credential refresh, and say it was local (#4546)

lidge-jun:
## Summary

A Codex pool account whose forced credential refresh fails now backs off, loses selection preference, and says on the record that the refusal was local. Before this, one unhealthy account could pin the whole pool at 503 indefinitely while healthy siblings sat idle.

**Reproduced live.** Every request routed to one pool account returned `503` with code `server_is_overloaded` — five of five sequential probes, two different models, ~600ms each. The stored record was healthy: present, no `deletedAt`, refresh token present, grant fingerprint present, expiry a week out. Not one line reached the service log. Switching the active account made it stop on the very next request.

The refusal was this proxy's own `poolCredentialRefreshIncompleteResponse`, and three things made it permanent. `isTerminalPoolRefreshFailure` admitted only `TokenRefreshError` with reason `revoked` or `expired`, so a missing record, a missing grant fingerprint, a token-endpoint 5xx and a generation CAS loss all became a **retryable** 503 whose body asks the client to retry — a loop that sustains the condition it waits out. Selection kept returning to the account because it still read `healthy`. And nothing distinguished this locally synthesized refusal from a genuine upstream overload.

`src/codex/pool-refresh-backoff.ts` adds a per-account cooldown. Consecutive non-terminal failures open a bounded growing window (2s through 60s); inside it no new forced refresh starts and `isCodexPoolRefreshCooling` removes the account from every selection path, while the binding is kept and the account is **not** quarantined — a token-endpoint 5xx must not retire a healthy account (#2887). A failure inside an open window does not grow it, so a burst of concurrent requests cannot race one account to the ceiling. The first successful refresh clears everything.

`CodexCredentialUnavailableError` replaces the bare `Error` for a missing record or fingerprint and is classified terminal, so the operator is told to sign in again instead of being asked to retry something that can never succeed. `isTerminalPoolRefreshFailure` now delegates to one definition.

The 503 wire contract is deliberately unchanged — its status and exact wording are what make the client apply retry-after backoff, and the word "reauthentication" would flip the classification. What changes is the record: the entry carries `terminalSource: "synthetic"` and a distinct `localTerminalReason`, because the dashboard renders this sentence under a field named "Upstream reason", which is what sent an experienced maintainer to the provider's status page for a fault that never left his own process.

Stacked on #4638.

## Verification

[...]

PR #4640, opened 2026-09-14
feat(codex): give V2 threads real lineage and place a child on its parent's serving account (#4546)

lidge-jun:
## Summary

Codex V2 threads now carry a real parent/child/grandchild relation, and a newly spawned child starts on the account that is actually serving its parent instead of cold.

`codexPoolAffinityKey` preferred `x-codex-parent-thread-id`, so every child of one parent bound under the raw parent id: one shared entry, unrelated to the root's own `app:HMAC(session, thread)` binding. No child could hold a binding of its own, a grandchild keyed on something nobody had ever bound, and a newly spawned child started with a cold prefix while its parent was being served warm somewhere else.

A request carrying `thread-id` now keys as `app:HMAC(session-id ?? parent, thread-id)`. The set of requests that bind at all is deliberately unchanged — a bare `thread-id` with no session and no parent still has no family anchor and stays unbound — and only the value moves, for requests naming a parent.

`src/codex/lineage.ts` records, per authenticated scope, each thread's conversation key, its immediate parent and its transitive root, so a grandchild resolves to its root rather than to its parent. It is bounded in both dimensions, by idle TTL plus an LRU cap on records per scope and on scopes, and is never keyed across authentication scopes.

Placement, and only placement, changes: a thread that has never bound starts on the account **currently** serving its parent — a live transient detour included, rather than the parent's stale home — then on a compatible sibling's, then on ordinary cold placement. The child holds an ordinary binding of its own, and the hint does not move the shared active-account cursor. The asymmetry is pinned: a parent bound to A but served by B places a new child on B, and when the parent later moves to C, the child already progressing on B is untouched while the next new child starts on C.

Four review findings are fixed on top of that. A parent-only turn now resolves through the existing scoped parent record instead of re-deriving `HMAC(parent, parent)`, which equals the root key only when `session-id` happens to equal `thread-id`. A lineage miss reads the legacy raw-parent key once, adopts its binding and retires the entry, so a conversation live across an in-process code swap is not silently cold-rebound — the exact defect this unit exists to prevent. Preview derives lineage from the same transformed auth headers and eligibility predicate `resolveCodexAuthContext` uses, so it cannot follow a family binding that final auth deliberately refuses to create. And "currently serving" now inspects the compatible model-detour entry before the ordinary binding.

Stacked on #4639.

## Verification

**Not run, by explicit instruction:** the local suite, `bun run typecheck`, `bun install`, and any build. The only proof for this unit is hosted CI at the exact final head SHA; this push used `--no-verify`.

New coverage in `tests/codex-integration/codex-lineage-placement.test.ts`: twelve tests covering the unchanged unbound set, transitive root resolution inside one auth scope, bounds in both dimensions, a child starting on the parent's account under its own key, following the account actually serving the parent, the parent-move asymmetry, sibling placement, cold fallback, a parent-only turn continuing the parent's conversation with or without a session id, adoption of a binding left under the old raw-parent key, following the parent's model detour, and preview reading the family only for a request that may own Pool state. `tests/codex-integration/codex-auth-context.test.ts` has its one raw-value assertion rewritten to the new contract.

Known open: the lineage-backed worker/interactive answer is exported but not consumed — admission still classifies from headers alone. The cost-attribution root lookup is exported and nothing spends against it yet.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Codex threads now maintain independent account affinity instead of sharing a parent thread’s lane.
  * New conversation lineage tracking helps newly created child threads start on the account serving their parent or a compatible sibling.
  * Parent-only requests continue using the parent conversation lane.
  * Routing previews and final credential selection now use consistent thread lineage information.
  * Lineage state is bounded and expires when inactive.

* **Documentation**
  * Updated OpenAI tier documentation to describe thread affinity and lineage-based placement.


lidge-jun:
## 리뷰 · 우선순위 69 / 80

이 PR은 스택 문서의 **wpd(V2 lineage placement)** 입니다. 문서 표에서는 wpf 다음·wpa 앞(4번)이었는데, 실제 브랜치는 `codex/4546-wph-pool-refresh-join` 위에 쌓여 순서가 표와 어긋납니다. 지금 `dev`(627274b8f)에는 아직 V2 parent/child 바인딩 키 붕괴가 남아 있고, 그게 프롬프트 캐시를 깨는 #4546 갈래 중 하나입니다.

문제는 plain합니다. `codexPoolAffinityKey`가 `x-codex-parent-thread-id`를 선호하면서, 같은 부모의 자식들이 raw parent id 하나로 묶이고, 루트의 `app:HMAC(session, thread)` 바인딩과 무관해졌습니다. 손자는 아무도 바인딩한 적 없는 키로 가고, 새 자식은 부모가 따뜻한 계정에 있는데도 차가운 prefix로 시작합니다.

`src/codex/lineage.ts`가 호출자 스코프(HMAC)·TTL·엔트리/스코프/형제 상한을 둔 프로세스 로컬 혈통을 만들고, 자식 첫 배치를 부모 serving account에 올리며, 레거시 raw-parent 키를 새 키로 한 번 이주합니다. 재시작 비내구·예산 미구현·워커 분류는 노출만 하고 미배선이라는 주석이 정직합니다. auth-context·routing·request-log·core·구조 문서·통합 테스트가 같이 옵니다.

라인 src/codex/lineage.ts (프로세스 로컬·HMAC 키) - 재시작 후 혈통이 사라지는 것은 의도입니다. 다만 롤링 배포 중 두 프로세스가 같은 대화를 나눠 보면 배치 힌트가 갈라집니다. 단일 인스턴스 전제인지 운영 노트에 남기세요.

라인 스택 순서 - 문서는 wpd가 wpa/wpb와 독립(계약 모듈)이라고 했지만, 실제 base는 wph입니다. 독립이 맞다면 `dev`에 가깝게 리베이스해 표 순서(wpd→wpa→…)를 복구할 여지가 있습니다. 의존이 생겼다면 문서 표를 갱신해야 합니다.

경로 CI - test 1/4 실패·다수 pending. 앞 스택 빨간불과 겹칩니다. wpd 고유 실패인지 분리해 보세요.

메인테이너의 판단이 필요한 지점
- 표를 코드 스택에 맞출지, 브랜치를 표에 맞게 다시 쌓을지
- 워커 분류(`codexLineageWorkflowLane`) 배선을 이 PR에 넣을지 후속으로 둘지
- types/config 분할 무관

너의 추천
내용 우선순위는 높습니다. 머지 전에 (1) 문서 순서 vs 실제 base를 맞추고, (2) 호스티드 CI 최종 SHA 녹색을 확인하세요. 독립이 유지되면 wpd를 더 아래 스택에서 빼 `dev`에 가깝게 올리는 편이 리뷰·되돌리기에 유리합니다.

이 댓글은 grok-bot이 작성했습니다

PR #4660, opened 2026-09-14
fix(providers): Baseten routed rows must not advertise `text.verbosity`

### LINKED ISSUE #4630: [Provider compatibility] Baseten catalog advertises unsupported Responses text.verbosity
### Client or integration

Codex App

Baseten Model APIs (`baseten`)

### OpenCodex version

Reproduced with `@bitkyc08/opencodex` 2.42.0. The Baseten registry entry still lacks a verbosity opt-out on current `main` / package version 2.55.0.

### Endpoint or capability

Inbound `/v1/responses` conversion to Baseten `/v1/chat/completions`; `text.verbosity` capability metadata.

### Current behaviour

The generated Codex catalog advertises verbosity support for `baseten/deepseek-ai-DeepSeek-V4.1-Flash`:

```json
{
  "slug": "baseten/deepseek-ai-DeepSeek-V4.1-Flash",
  "support_verbosity": true,
  "default_verbosity": "low"
}
```

Codex consequently includes `text.verbosity` in its Responses request. OpenCodex rejects the request while converting it to the Baseten Chat Completions route, before any model output is produced.

The selected `high` value is the reasoning effort and is unrelated to the rejected verbosity field.

### Expected behaviour

Baseten documents its Model APIs as OpenAI **Chat Completions** compatible. The routed catalog should therefore not advertise the Responses-only `text.verbosity` capability unless OpenCodex can translate it. The request should omit that field and proceed normally.

```shell
# OpenCodeX 2.42.0; Baseten provider configured with a valid redacted API key.
ocx sync

# In Codex App:
# 1. Select baseten/deepseek-ai-DeepSeek-V4.1-Flash.
# 2. Select reasoning effort "high".
# 3. Start a turn with: Reply with OK.

# Relevant request shape produced by Codex from the generated catalog:
POST /v1/responses
{
  "model": "baseten/deepseek-ai-DeepSeek-V4.1-Flash",
  "input": "Reply with OK.",
  "reasoning": { "effort": "high" },
  "text": { "verbosity": "low" }
}
```

### Actual response or error

```shell
HTTP 400

{
  "error": {
    "code": "400",
    "message": "Validation: Failed to convert responses request: `text.verbosity` is not supported",
    "type": "Bad Request"
  }
}
```

### Upstream documentation

Baseten Model APIs overview: https://docs.baseten.co/inference/model-apis/overview

Baseten Chat Completions reference: https://docs.baseten.co/reference/inference-api/chat-completions

These pages specify OpenAI Chat Completions compatibility. I found no public Baseten specification for the OpenAI Responses API or its `text.verbosity` field.

Declare the capability on the Baseten provider registry entry:

```ts
{
  id: "baseten",
  // ...
  supportsVerbosity: false,
}
```

The generated entry should then contain `"support_verbosity": false` and omit `default_verbosity`. A focused catalog regression test for a live-discovered Baseten model would cover the provider-wide fallback.

Environment: macOS 26.6.2, `codex-cli 0.154.0`. Current `main` version 2.55.0 still has no `supportsVerbosity` declaration in the Baseten registry entry. Existing issues were searched for Baseten, `text.verbosity`, and this model ID; no duplicate was found.

### Checks

abhisheksharma2411:
Closes #4630.

## Summary

- Baseten documents its Model APIs as **Chat Completions** compatible. `text.verbosity` is an OpenAI **Responses** parameter, so there is nothing on that wire for it to become — but the Baseten registry entry carried no opt-out, so every routed row serialized `support_verbosity: true` with `default_verbosity: "low"`, and Codex seeds its picker from exactly that.
- Adds `supportsVerbosity: false` to the `baseten` entry in `src/providers/registry/entries-extended.ts` (the leaf, post-#4655 — not the old monolithic `registry.ts`).
- Adds two regression cases to `tests/codex-integration/catalog-verbosity-default.test.ts`, alongside the existing xAI and Kiro ones.

**Provider-wide, not per-model**, matching how xAI (`entries-core.ts:250`) and Ollama (`entries-extended.ts:831`) already spell their opt-outs, and for the reason Ollama's comment gives: Baseten is `liveModels: true`, so a slug discovered tomorrow supports `text.verbosity` no more than this one. Pinning it per-model would leave the next discovered id advertising the control again.

### Reproduced before fixing

Wrote the assertion first and watched it go red on unmodified `dev`:

```
84 |     expect(baseten?.support_verbosity).toBe(false);
                                            ^
Expected: false
Received: true
```

So the catalog really does advertise the control today — that isn't inferred from the registry entry, it's the serialized row.

### The live-discovery case, from review

@coderabbitai and @lidge-jun both pointed out that the original test seeded a static `models` list with `liveModels: false`, which meant the "tomorrow's slug" reasoning above was a *comment* rather than something the suite checked — and live discovery is the authoritative path for this provider.

Added a second case that seeds **no** models and lets a stubbed `/models` response supply the slug. Mutation-checked: removing `supportsVerbosity: false` from the registry now fails **both** cases, so the provider-wide claim is pinned on the path that actually matters.

### One correction to the issue, which does not change the fix

The report attributes the 400 to OpenCodex: *"OpenCodex rejects the request while converting it to the Baseten Chat Completions route."* I don't think that's where it comes from.

- The string `Failed to convert responses request` appears nowhere in this repository.
- Neither `src/chat/` nor `src/responses/` mentions `verbosity` at all — the only stripper is in `src/adapters/openai-responses.ts:428`, which is the Responses adapter, not the Chat one.

So the message reads like Baseten's own Responses→Chat shim rejecting it upstream, not our converter. Worth pinning down if you're tracking who 400s, but it doesn't move the fix either way: advertising a capability the routed wire cannot carry is wrong whichever side refuses it, and this makes the catalog honest at the source.

### Scope I left alone

31 other `adapter: "openai-chat"` providers also have no `supportsVerbosity` opt-out (openrouter, deepseek, moonshot, commandcode, nvidia, …). That may well be the same class of bug, but "Chat adapter" isn't proof on its own — several of those front gateways that do accept a Responses-shaped body — and a blanket sweep on one reproduced report would be guessing at 31 upstream contracts. @lidge-jun agreed this belongs in a separate issue; happy to open it.

## Verification

Rebased onto current `dev` (`aa91958e3`), so this is 0 commits behind.

```
bun test tests/codex-integration/catalog-verbosity-default.test.ts     6 pass, 0 fail   (red before the fix)
  + codex-catalog.test.ts + tests/providers/baseten-provider.test.ts   352 pass, 0 fail
bun run structure:check                                                structure/ SSOT checks passed
bun run privacy:scan                                                   Privacy scan passed
bun run typecheck                                                      2 errors — PRE-EXISTING
```

The typecheck failures are `src/server/responses/fetch-helpers.ts(195,7)` and `(208,7)`, both `'timeout' does not exist in type 'RequestInit'`. I stashed this branch and re-ran on unmodified `dev`: **the same 2 errors, same file, same lines.** Not introduced here, and not touched by this change.

## Checklist

- [x] Docs or release notes were updated when needed — the registry comment carries the reasoning; no user-facing doc claims Baseten verbosity support.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults — none; this is one capability flag in a provider preset.

## Review readiness checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Baseten models no longer inherit Codex text verbosity settings when accessed through Chat Completions, including routed and live-discovered models.

* **Tests**
  * Added coverage confirming verbosity opt-out behavior for both configured and dynamically discovered Baseten models.


lidge-jun:
## 리뷰 · 우선순위 58 / 80

이 PR은 이슈 #4630을 닫는 수정이다. Baseten Model APIs(`baseten`)로 라우트된 모델이 Codex 카탈로그에서 `support_verbosity: true` / `default_verbosity: "low"`처럼 보이면, Codex App이 Responses 요청에 `text: { verbosity: "low" }`를 넣고, 그 다음 Chat Completions 쪽 경로에서 ``text.verbosity` is not supported``로 첫 턴이 모델 출력 전에 400으로 막힌다. 사용자가 고른 reasoning effort와는 무관하다. Baseten 공식 문서는 Chat Completions 호환이지 Responses/`text.verbosity` 스펙이 아니다.

현재 `dev` HEAD는 `09067c586`(2.56.0, #4655 godfile round3 직후)이다. baseten 엔트리는 이미 모놀리식 `src/providers/registry.ts`가 아니라 리프 `src/providers/registry/entries-extended.ts`에 있다. 이 PR은 그 리프에 `supportsVerbosity: false` 한 줄을 넣고, `tests/codex-integration/catalog-verbosity-default.test.ts`에 Baseten GREEN 케이스를 xAI/Kiro 옆에 추가한다. types.ts/config.ts 분할·close-don't-rebase 대상이 아니고, #4655 이후 올바른 파일을 고쳤다.

왜 이게 맞는가. `src/providers/derive.ts`의 `applyVerbosityDefaults`는 레지스트리 `supportsVerbosity`가 `undefined`이면 early return이라, 명시하지 않으면 Codex 템플릿 기본(지원처럼 보임)을 물려받는다. xAI(`entries-core.ts`)와 Ollama(`entries-extended.ts`)는 이미 provider-wide로 `supportsVerbosity: false`를 박아 두고, 주석도 “live discovery로 나중에 오는 id도 같이 opt-out”이라고 설명한다. Baseten만 그 패턴이 비어 있었고, baseten은 `liveModels: true`라서 per-model 핀만 하면 내일 발견되는 슬러그가 다시 광고한다. PR 본문이 그 이유를 정확히 적었다.

400이 우리 converter인지 Baseten 쪽 Responses→Chat shim인지는 PR이 바로잡은 대로 저장소에 `Failed to convert responses request` 문자열이 없고, Chat 경로에도 verbosity stripper가 없다. 광고 자체를 끄면 Codex가 필드를 안 보내므로, 어느 쪽이 거절하든 체감 버그는 같은 한 줄로 막힌다. 작성자가 red→green으로 재현했고(미수정 `dev`에서 `support_verbosity`가 true), 카탈로그·baseten·structure/privacy 테스트 통과를 보고했다. typecheck 2건은 `fetch-helpers.ts` timeout으로 이 PR과 무관한 선재 오류다.

라인 수준에서 보면 본문 변경은 깨끗하다. 테스트는 정적 `models` + `liveModels: false`로 라우트 행을 검증한다. 이것으로도 “레지스트리 opt-out이 직렬화 행에 반영된다”는 핵심 회귀는 잡힌다. 다만 CodeRabbit이 지적한 대로, 실제 baseten 경로는 live discovery가 권위라서 stub로 `liveModels: true` + discovery 응답 한 건을 넣으면 “내일 슬러그” 주석과 테스트가 한 줄로 맞는다. 필수는 아니고 있으면 더 좋다. 다른 `openai-chat` 31개 제공자 일괄 감사는 이 PR 범위 밖이 맞고, 별도 이슈로 추적하는 편이 안전하다. PR은 아직 draft(체크리스트 0/4)이고 MERGEABLE이며 하이진/enforce-target 등은 녹색이다.

경로/src/providers/registry/entries-extended.ts baseten 엔트리 - `supportsVerbosity: false`와 live-discovery 주석이 xAI/Ollama 패턴과 같고, #4655 이후 올바른 리프를 고쳤다. 문제 없음.
경로/tests/codex-integration/catalog-verbosity-default.test.ts Baseten GREEN - `support_verbosity === false`와 `default_verbosity` 부재를 직렬화 철자로 검사한다. 좋음.
심볼/gatherRoutedModels liveModels:false - 정적 시드만 검증한다. discovery 권위 경로를 한 번 더 밟으면 회귀 범위가 이슈 제안과 완전히 일치한다.
경로/src/providers/derive.ts applyVerbosityDefaults - 이 PR이 건드리지 않지만, 레지스트리에 false를 넣는 순간 provider config에 opt-out이 심어지는 기존 경로다. 추가 코드 불필요.
심볼/Closes #4630 - 이슈 제안(`supportsVerbosity: false` provider-wide + 카탈로그 회귀)과 구현이 일치한다. 중복 PR로 보이지 않는다.

메인테이너의 판단이 필요한 지점
- draft 체크리스트(0/4)를 작성자가 채운 뒤 `review-ready`로 올릴지, 내용이 이미 충분하니 메인테이너가 바로 머지할지
- 테스트에 live discovery stub를 한 케이스 더 넣을지(있으면 좋음, 없어도 머지 가능)
- 나머지 `openai-chat` 제공자 verbosity 감사는 후속 이슈로 열지, 지금은 Baseten만 닫을지
- 400 메시지의 실제 발신처(우리 vs 업스트림)를 이슈 코멘트로만 남길지, 변환기 strip 완화책을 병행할지(카탈로그 수정만으로도 #4630은 닫힘)

너의 추천
체크리스트만 채우게 한 뒤, 이 상태로 `dev`에 머지하는 것을 추천한다. 필요하면 liveModels discovery stub 한 줄을 후속 커밋으로 받아도 되고, 없어도 #4630 본문은 해결된다. 다른 openai-chat 제공자 감사는 별도 이슈로 열고 이 PR에는 넣지 말 것. 라벨은 건드리지 않았다.

이 댓글은 grok-bot이 작성했습니다

abhisheksharma2411:
@lidge-jun — took the live-discovery suggestion; checklist filled 4/4.

**The stub test was worth doing, and not just for tidiness.** You and @coderabbitai both noted the original case seeded a static `models` list with `liveModels: false`, so the "tomorrow's slug" reasoning in the registry comment was prose rather than something the suite checked — on a provider where discovery is the authoritative path. Added a second case that seeds **no** models and lets a stubbed `/models` response supply the slug.

Mutation-checked rather than just watched green: removing `supportsVerbosity: false` from the registry now fails **both** cases, so the provider-wide claim is pinned on the path that actually matters, not only on the seeded one.

Rebased onto `aa91958e3` first so "pushed to the latest dev commit" is literally true rather than 6 behind. 352 pass / 0 fail across `catalog-verbosity-default`, `codex-catalog` and `baseten-provider` after the rebase.

On your other two open questions:

- **openai-chat verbosity audit** — agreed it's a separate issue. I'll open it with the 31 providers listed, unless you'd rather file it yourself so the triage priority is yours to set.
- **the 400's real sender** — I'd leave it as the issue comment for now. The string `Failed to convert responses request` is in neither `src/chat/` nor `src/responses/`, and neither mentions `verbosity` at all, so I'm fairly confident it's Baseten's own shim. Relaxing our converter would be a fix for a problem we don't appear to have.

Ingwannu (approved):
Approved on exact head `29996e108d5421d55334d35d23d20535459f9f18`. The provider-wide opt-out matches the Baseten Model APIs Chat Completions wire: routed rows must not advertise the Responses-only `text.verbosity` control. The static and genuinely live-discovered cases both prove the default is removed for present and future slugs, the review thread is resolved, and exact-head cross-platform CI is green. The cancelled duplicate target-gate run is superseded by a successful target-gate run at the same SHA. No Go counterpart is applicable to this registry/catalog-only TypeScript provider correction.

Issue #4024, opened 2026-09-08
[Feature]: OpenRouter — automatic key rotation & model failover when the free-tier quota is exhausted


nordz0r:
### Area

Provider adapters

### What are you trying to accomplish?

Use OpenRouter free-tier models (`openrouter/minimax/minimax-m3:free`, `openrouter/openrouter/free`, etc.) behind a combo/failover target without manual intervention when a key's weekly free quota is exhausted (OpenRouter replies with 429 `rate_limit_error`, `Weekly/Monthly Limit Exhausted... will reset at <date>`).

### What prevents this today?

OCX already has two related mechanisms:

1. **`apiKeyPool` multi-key failover** (`src/providers/key-failover.ts`): on a 429, `rotateKeyOn429()` moves to the next key in the pool and puts the exhausted key into cooldown (default 60 s, cap 10 min, `Retry-After` respected).
2. **Combo failover across targets**: a combo can fail over to another provider/model.

The gap for the OpenRouter free-tier case specifically:

- The free-quota reset horizon is **weekly** (`Your limit will reset at 2026-09-09 03:30:06`), but `MAX_COOLDOWN_MS` caps a key's cooldown at **10 minutes**. After 10 minutes the exhausted key becomes active again, and roughly every other request goes to a dead key, failing with 429 and only then rotating — the wasted attempt repeats all week until the quota actually resets.
- The OpenRouter error body carries the exact reset timestamp (`will reset at <date>`), but the cooldown logic appears to only honor the `Retry-After` HTTP header; the body-parsed reset date is not used to extend the cooldown.
- There is no per-key quota-state notion (e.g., `exhausted_until: <ts>`), so an exhausted key cannot be parked for the rest of the window the upstream itself declares.

### What should OpenCodex do?

1. Parse the OpenRouter 429 JSON body for `rate_limit_error` and extract the reset timestamp from the message; park the key in cooldown until that timestamp (bounded, e.g. up to 7 days, configurable).
2. Optionally: a generic `quotaExhaustedCooldownMs` / `quotaExhaustedUntilReset: true` knob under `retryOn429` (or a sibling policy) so other providers that report reset windows in the error body get the same behavior.
3. With a single key in the pool, the same detection should feed combo-level failover: mark the target as quota-exhausted so the combo routes to the next target for the remainder of the window instead of burning one failed attempt per request.

### Example usage or interface

```json
{
  "providers": {
    "openrouter": {
      "authMode": "key",
[...]

PR #4733, opened 2026-09-15
feat(providers): park a key until the reset instant the upstream declared

abhisheksharma2411:
Closes #4024.

## Summary

A free-tier quota exhaustion **is** dated — OpenRouter replies `Weekly/Monthly Limit Exhausted ... will reset at <date>`. But it puts that date in the **body**, and sends no `Retry-After`. `parseRetryAfterMs` only reads the header, so the key falls back to `DEFAULT_COOLDOWN_MS`, returns after a minute, takes another 429, rotates again — every minute, for the rest of the quota window. @Ingwannu confirmed the same reading of `key-failover.ts` on `dev@29bb221c3`.

- `parseQuotaResetAt(body)` — pulls `reset[s] at <date>` from a bounded 4 KB prefix.
- `readQuotaResetAt(response)` — reads it off a **cloned** response, so the caller can still `body.cancel()` to release the socket, which `adapter-dispatch.ts` does on the next line.
- When present, it outranks both the header and the default. It is the only one of the three that knows when the quota actually comes back.

## The cap, and why it is not the existing one

`MAX_COOLDOWN_MS` stays 10 minutes. That shortness is correct for an **undated** 429 — it's a guess, and a guess shouldn't park a working key for long. A dated one isn't a guess, so it gets its own ceiling: `MAX_QUOTA_COOLDOWN_MS = 32 days`.

**I had that wrong at first and the test caught it.** I wrote 8 days, reasoning about the weekly case in the issue. But the wording being parsed is `Weekly/**Monthly** Limit Exhausted`, and an 8-day cap silently clamps every monthly reset ~23 days early — straight back into the loop this exists to remove. 31 days plus a day of slack for month length and zone.

Bounded at all because the date is upstream-controlled input: `reset at 2999-01-01` must not park a working key past any horizon an operator would think to look at.

## A runtime divergence worth knowing about

The parser pins a bare `YYYY-MM-DD hh:mm:ss` to UTC explicitly. That looks redundant on this runtime, and I want to be straight about why it isn't:

```
Node  Date.parse("2026-09-09 03:30:06")  →  2026-09-09T10:30:06Z   (local, PDT box)
Bun   Date.parse("2026-09-09 03:30:06")  →  2026-09-09T03:30:06Z   (UTC)
```

ECMA-262 says a date-time form with no offset is **local**; Node follows it, Bun currently doesn't. So on Bun the normalisation is a no-op **today** — and if Bun ever conforms, an un-normalised parse would shift every park-until by the operator's offset, with the early direction resuming the 429 tax.

**The honest consequence: no Bun test can observe that branch being removed.** I mutated it away and the suite stayed green. The explicit-zone case (`+05:30`) is the part the suite *can* pin, and it does. Flagging rather than claiming coverage I don't have.

## Verification

```
openrouter-quota-reset-cooldown-4024.test.ts                         10 pass, 0 fail
every other suite importing key-failover (5 files)                  105 pass, 0 fail
bun run structure:check                                             passed
bun run privacy:scan                                                passed
bun run typecheck                                                   2 errors — PRE-EXISTING
```

Typecheck failures are `src/server/responses/fetch-helpers.ts(195,7)` and `(208,7)`, `'timeout' does not exist in type 'RequestInit'` — same two as on untouched `dev`.

Mutation-tested:

| Mutation | Result |
|---|---|
| ignore the body date entirely (today's behaviour) | **6 fail** |
| remove the 32-day cap | cap test fails |
| scan the whole body instead of 4 KB | prefix test fails |
| drop the UTC normalisation | **survives — see above** |
| restored | 10 pass |

## Scope

Wired into `adapter-dispatch.ts`, the main rotation path. The other `rotateProviderTransportOn429` call sites (`chat-native`, `adapter-continuation`, `encrypted-payload`, `compact`) pass nothing and keep exactly today's behaviour — the option is optional, so this is additive. Happy to thread the rest in this PR or a follow-up; I stopped at one path rather than touch five hot sites in a change that needs live-traffic confirmation anyway.

The issue's second half — reset-aware **combo target** parking, as opposed to key parking — is not here. That's a different mechanism and @Ingwannu's note about rotation persisting the replacement key suggests it wants its own discussion.

## Review readiness checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved provider failover when quota limits are reported in a response body.
  - Quota reset times are now recognized and used to schedule key cooldowns, including UTC, date-only, and time-zone-aware timestamps.
  - Reset schedules take precedence over generic retry timing, with safeguards for invalid, expired, or excessively distant dates.
  - Response bodies remain available after reset-time detection.
  - Existing behavior remains unchanged when reset details are unavailable or for authentication failures.


lidge-jun:
## 리뷰 · 우선순위 73 / 80

이 PR(작성자 abhisheksharma2411, draft, base `dev`, MERGEABLE)은 OpenRouter 무료 티어 할당량이 바닥났을 때 생기는 **키 쿨다운이 너무 짧은 문제**를 고친다. 지금 `dev` tip은 `45cfb04e9757a5a257ab6290d9f24d2ea0bc7573`(package 2.57.0, #4702 UA 보존 직후)이다. 현재 `src/providers/key-failover.ts`의 `rotateKeyAfterFailure`는 429일 때 `parseRetryAfterMs(Retry-After 헤더)`만 보고, 없으면 `DEFAULT_COOLDOWN_MS`(60초)로 키를 잠깐만 재운다. `MAX_COOLDOWN_MS` 상한도 10분이다. 그런데 OpenRouter 무료 할당량 고갈은 본문에 `Weekly/Monthly Limit Exhausted ... will reset at <날짜>`라고 **언제 돌아오는지**를 적어 주고, `Retry-After` 헤더는 안 보낸다. 그래서 키가 1분~10분마다 다시 살아나고, 또 429를 맞고, 또 돌아가기를 할당량이 풀릴 때까지 반복한다. 이슈 #4024가 정확히 그 구멍을 적었고, 이 PR은 그중 **키 주차(park-until)** 절반을 닫는다.

고치는 방법은 세 조각이다. (1) `parseQuotaResetAt(body)`가 본문 앞 4KB에서 `reset[s]? at <날짜>`를 정규식으로 뽑고, 오프셋 없는 시각은 UTC로 고정한 뒤 epoch ms를 돌려준다. 이미 지난 시각·이상 먼 미래(2999년 등)는 거절하거나 `MAX_QUOTA_COOLDOWN_MS = 32일`로 자른다. (2) `readQuotaResetAt(response)`는 응답을 `clone()`한 뒤 읽어, 호출측이 원본 바디를 `cancel()`해 소켓을 비울 수 있게 한다. (3) `rotateKeyAfterFailure` / `rotateKeyOn429` / `rotateProviderTransportOn429` 옵션에 `quotaResetAt`을 추가해, 이 값이 있으면 헤더·기본값보다 **우선**해서 `cooldownUntil = now + (quotaResetAt - now)`로 키를 재운다. 배선은 메인 경로 `src/server/responses/adapter-dispatch.ts`의 multi-key 429 루프에만 넣었다. 작성자가 밝힌 대로 `chat-native` / `adapter-continuation` / `encrypted-payload` / `compact` / `sidecar-execution` 등 다른 `rotateProviderTransportOn429` 호출처는 옵션을 안 넘겨서 **오늘과 동일한 짧은 쿨다운**을 유지한다.

32일 상한 선택은 이슈 본문이 제안한 ‘대략 7일·설정 가능’보다 길다. 이유는 파싱 문구가 `Weekly/**Monthly** Limit Exhausted`라서, 8일 상한이면 월간 리셋이 약 23일 일찍 잘려 **다시 같은 429 루프**로 돌아가기 때문이다. 작성자가 처음 8일로 썼다가 캡 테스트에 걸려 32일(31일+여유)로 고친 과정이 PR 본문에 솔직히 적혀 있다. 날짜가 업스트림이 통제하는 입력이라 상한 자체는 필수다. Bun vs Node의 `Date.parse` 로컬/UTC 차이도 본문에 측정값과 함께 적어 두었고, ‘Bun 테스트로는 UTC 정규화 분기를 제거해도 초록이 유지된다’는 한계까지 스스로 표시했다. 명시 타임존(`+05:30`) 케이스는 스위트가 고정한다.

테스트 `tests/providers/openrouter-quota-reset-cooldown-4024.test.ts`(+89)는 OpenRouter 문구·명시 존·`resets at`·날짜만·파싱 실패·과거 시각·월간 창·적대적 미래·4KB 밖 무시·clone 후 bodyUsed=false·바디 없는/이미 소비된 응답을 덮는다. 뮤테이션 표(바디 날짜 무시 → 6 fail, 32일 캡 제거 → fail, 전체 바디 스캔 → fail)도 설득력 있다. types.ts/config.ts 분할 캠페인과 겹치지 않고, godfile facade 경로도 건드리지 않는다. #4024의 둘째 절반(콤보 타깃을 할당량 고갈로 표시해 다음 타깃으로 넘기기)과 설정 노브(`quotaExhaustedUntilReset`)는 의도적으로 빠져 있다. 체크리스트 네 칸은 모두 비어 있고 draft라 Ready는 아니다. 점수 73은 ‘실제 비용/루프 버그를 닫는 좁고 잘 설명된 패치’라서 70대 중반이지만, 배선이 adapter-dispatch 한 곳뿐·`readQuotaResetAt`의 실제 읽기 상한·프로세스 로컬 Map에 32일 주차·draft 미완이라 76 이상으로 올리지 않았다.

라인 66 - `readQuotaResetAt`이 `response.clone().text()`로 **전체 본문**을 메모리에 올린 뒤에야 `parseQuotaResetAt`의 4KB slice가 돌아간다. 주석·테스트는 ‘bounded prefix / 거대한 바디가 회전 경로를 막지 못하게’라고 말하지만, 네트워크·메모리 읽기 자체는 아직 상한이 없다. 4KB만 읽고 끊는 reader(또는 Content-Length/스트림 상한)가 주석과 맞다.
라인 85 - 정규식 `reset[s]?\s+at\s+...`는 OpenRouter 전용 마커(`rate_limit_error` / Limit Exhausted) 없이 광범위하다. 다른 프로바이더 429 본문에 비슷한 영어가 있으면 의도치 않게 긴 park가 걸릴 수 있다. OpenRouter 문구에 더 가깝게 좁히거나 provider 힌트를 받는 편이 안전하다.
라인 505-506 - `quotaResetAt`이 있으면 `Retry-After`를 완전히 무시한다. OpenRouter(헤더 없음)에는 맞지만, 헤더와 본문 날짜가 둘 다 있는 업스트림에서는 더 짧은 헤더를 버리고 한 달까지 키를 재울 수 있다. ‘본문이 이기고, 다만 헤더보다 길 때만’ 같은 규칙을 메인테이너가 원하는지 확인이 필요하다.
경로 chat-native.ts / adapter-continuation.ts / sidecar-execution.ts / compact.ts / encrypted-payload.ts - 같은 `rotateProviderTransportOn429`를 쓰지만 `quotaResetAt`을 안 넘긴다. Responses adapter-dispatch로만 들어오는 OpenRouter 트래픽은 고쳐지고, native chat·continuation·사이드카 경로는 #4024 루프가 남는다. 후속 PR로 묶을지, 이번 PR에서 같이 넣을지 결정이 필요하다.
경로 keyCooldowns(Map) - 쿨다운은 여전히 프로세스 메모리다. 32일 park는 재시작·다중 인스턴스에서 바로 사라진다. 기존 설계이지만 상한이 커진 만큼 운영 기대치(‘한 주 동안 안 돌아온다’)와 어긋날 수 있다. #4024/#3376 쪽 durable quota 이야기와 연결할지 메모만 남긴다.
PR 본문 checklist / draft - CI·dev tip·Codex/CodeRabbit·Ready 네 칸이 비어 있다. 로컬 스위트 주장은 강하지만 Ready 전 hosted CI 초록과 tip 동기화가 필요하다.

메인테이너의 판단이 필요한 지점
- #4024를 이 PR(키 park만)로 부분 닫고 콤보 타깃 park는 후속으로 둘지, 이슈를 열어 둔 채 키 절반만 머지할지.
- adapter-dispatch 외 나머지 429 회전 호출처를 같은 PR에 넣을지, 라이브 트래픽 확인 후 follow-up으로 둘지(작성자 제안).
- 32일 하드캡 vs 이슈의 ‘~7일·설정 가능’ — 월간 창을 위해 32일을 받아들일지, 설정 노브를 요구할지.
- `readQuotaResetAt`의 전체-본문 clone 읽기를 머지 전 반드시 고칠지, OpenRouter 짧은 JSON 전제하에 후속으로 둘지.

너의 추천
draft 유지. `readQuotaResetAt`을 진짜 4KB(또는 작은 상한)만 읽도록 고치고, OpenRouter 문구에 더 가깝게 정규식을 좁힌 뒤, hosted CI가 tip `45cfb04e9` 기준 초록인지 확인한다. 키 park만 먼저 머지할 거면 #4024 본문에 ‘키 절반 landed / 콤보 절반은 open’을 남기고, chat-native·continuation·sidecar 배선은 바로 이은 좁은 follow-up으로 잡는다. types/config 분할에 무효화되지 않으니 close-don't-rebase 대상은 아니다. 지금 상태로 Ready·머지하지 말 것.

이 댓글은 grok-bot이 작성했습니다

abhisheksharma2411:
You were right to hold this, and the `readQuotaResetAt` finding was the important one — fixing it turned up something worse than what you flagged.

**You were right that the read was unbounded.** `parseQuotaResetAt` sliced 4KB, but `readQuotaResetAt` called `.text()` first, so the whole body was already in memory. My PR body claimed "a bounded 4 KB prefix"; the *parse* was bounded, the *read* wasn't. That's a wrong claim in my own description and I've corrected it.

**Fixing it exposed a hang.** My first fix pulled a bounded prefix from `response.clone()`. That still times out — `clone()` tees, and the caller leaves the original branch undrained while this runs, so the tee stalls once its buffer fills. Reproduced against a **finite 5 MB** stream: 5s timeout, no result. So the clone approach could hang the rotation path outright, which is strictly worse than reading a large body.

It no longer clones. It pulls a bounded prefix from the original and returns a `Response` that replays those bytes ahead of the remainder:

```ts
readQuotaResetAt(response) -> { at, response }
```

`adapter-dispatch` rebinds `upstreamResponse` to it. Consuming the original outright wasn't an option either — on the `!rotated` path the loop breaks out still holding that response, so it has to stay readable.

The new test counts bytes **pulled**, not bytes parsed. That distinction is the whole bug: it passes against a 5 MB body having read under 64 KB, and the old `.text()` version would have read all 5 MB while still "passing" a parse-side assertion.

On your other four points:

- **Other call sites** (`chat-native`, `adapter-continuation`, `sidecar-execution`, `compact`, `encrypted-payload`) — I'd still prefer a follow-up, and more so now. The reader is no longer a passive peek: it rebinds the caller's response. Threading that through five hot paths deserves its own review rather than riding along here.
- **32-day cap vs the issue's ~7 days/configurable** — happy either way. I chose 32 because the wording parsed is `Weekly/**Monthly** Limit Exhausted`, and an 8-day cap silently clamps a monthly reset ~23 days early, back into the loop this removes. A config knob is strictly better if you want one; say the word.
- **`keyCooldowns` is process memory** — agreed, and worth recording on #4024 rather than solving here. A 32-day park that evaporates on restart or doesn't exist on a second instance is a weaker promise than the cap implies. That's the durable-quota conversation in #3376, not this PR.
- **Partial close of #4024** — your call. If you take the key half, I'd rather the issue stay open with a note than have it auto-close; the combo-target half is a different mechanism and @Ingwannu's point about rotation persisting the replacement key suggests it needs its own design pass.

Leaving it draft as you recommended. Verification at this head: 13 pass / 0 fail on the feature suite, 105 pass / 0 fail across the five suites importing `key-failover`, `structure:check` green, typecheck 2 pre-existing (`fetch-helpers.ts` 195/208, same on untouched `dev`).

I have **not** narrowed the regex toward OpenRouter's exact wording yet — that was your other suggestion and I'd like to check it against a real 429 body first rather than guess at the phrasing. If @Ingwannu or anyone has one to hand, that'd settle it.

abhisheksharma2411:
Rebased onto `dev` (`b3035fe`, 0 behind), out of draft. All three CodeRabbit findings are addressed — two of them were correct and are fixed here.

**1. Bounded 429-body read** — fixed earlier (`cc765da`). The read is now bounded, not just the parse, and it no longer clones: `response.clone()` tees, and with the original branch undrained the tee stalls once its buffer fills. It peeks a 4 KB prefix and hands back a Response replaying the pulled bytes ahead of the remainder, so the cancel downstream still releases the socket.

**2. Calendar-invalid dates — correct, and worse than "minor" on Bun.** I checked before fixing:

```
Date.parse("2026-02-30T00:00:00Z")  ->  2026-03-02T00:00:00.000Z
Date.parse("2026-02-29T00:00:00Z")  ->  2026-03-01T00:00:00.000Z
Date.parse("2026-04-31T00:00:00Z")  ->  2026-05-01T00:00:00.000Z
Date.parse("2026-13-01T00:00:00Z")  ->  NaN
```

Only the month is rejected. So a malformed body parked the key past the instant it actually named — bounded by the 32-day cap, but still wrong.

Validated on the **date text**, not by round-tripping the parsed instant: a value carrying an explicit offset (`…T23:00+05:30`) legitimately lands on a different UTC day than the one written, so a round-trip check would reject good input. Century rule included — 2000 is a leap year, 2100 is not. Four mutants, all killed: drop the guard, `leap = true`, drop the century rule, `day <= 31`.

**3. No dispatch-path coverage — correct, and the gap was real.** Added to `tests/server/server-key-failover-e2e.test.ts`: a two-key pool where the first attempt returns a 429 carrying **both** `retry-after: 30` and a body date six hours out, then asserts the failed key's cooldown via `getKeyCooldownUntil` is the body's instant rather than the header's 30s.

I mutation-tested it against exactly the two regressions you described, and both are caught:

| mutation | result |
|---|---|
| drop `quotaResetAt` at the `adapter-dispatch.ts` call site | **1 fail** |
| make `Retry-After` win over the body instant in `key-failover.ts` | **1 fail** |

**On "all CI tests are green" — worth stating plainly rather than just ticking.** Targeted suites are `1453 pass / 0 fail` (51 files, `--isolate`). The *full* local suite is `25387 pass / 28 fail`, but that baseline is **identical on `dev`** — I ran the failing suites on both and got `543 pass / 18 fail` either way. They're macOS-local (e.g. `remote workspace Linux command sandbox` expects bubblewrap), so they're green on the ubuntu runners. This branch introduces no new failures. `bun run typecheck` also fails identically on `dev` at `claude-messages.ts:611` and `responses/fetch-helpers.ts:208`, untouched here.

Worth knowing separately: the 100-failure count you'd see running these suites **without** `--isolate` is cross-test pollution, not real. Same 100 on `dev`.

abhisheksharma2411:
Thanks for the rebase — left it alone rather than force-pushing over it. Answering the open points, several of which are already done since that review was written against tip `45cfb04e9`.

**Bounded read — done.** `readQuotaResetAt` no longer reads the whole body and no longer clones. `response.clone()` was the wrong tool twice over: it tees, and with the original branch undrained the tee stalls once its buffer fills — I reproduced a 5s hang on a finite 5 MB body. It now peeks a 4 KB prefix and hands back a Response replaying the pulled bytes ahead of the remainder, so the downstream cancel still releases the socket.

**Checklist / draft — done.** All four boxes ticked and out of draft, each literally true at the time: targeted suites `1453 pass / 0 fail` under `--isolate`, and rebased to the tip before ticking rather than attesting while behind.

Worth recording, because it invalidates test numbers in earlier comments on this PR: the full local suite shows `25387 pass / 28 fail`, but that baseline is **identical on `dev`** — the failing suites give `543 pass / 18 fail` on both, and they're macOS-local (`remote workspace Linux command sandbox` wants bubblewrap). Separately, the ~100 failures you'd see running provider suites **without** `--isolate` are cross-test pollution, same count on `dev`. CI uses `--isolate`; I hadn't been, and every test number I reported for this repo before that was measured wrong.

**Cooldowns are process memory — you're right, and it matters more now.** A 32-day park in `keyCooldowns` evaporates on restart or in a second instance, and the longer the cap the wider that gap between stated and actual behaviour. Pre-#4024 the worst case was a 10-minute cooldown, where losing it costs one wasted request; at 32 days the operator expectation ("this key is parked for the month") is simply not what the process guarantees. I don't think that blocks this PR — it's strictly better than re-hammering an exhausted key every 10 minutes — but it should be said out loud rather than discovered, and it's the natural bridge to the durable-quota thread on #4024/#3376. Happy to open a follow-up issue stating the durability limit so it's tracked rather than implied.

**32 days vs the issue's "~7 days, configurable":** 32 is deliberate — it's the smallest cap that survives a monthly reset window, and an 8-day cap silently clamped a real monthly reset ~23 days early (its own test caught that). No objection to a config knob, but I'd rather the default not truncate a window the upstream actually declared.

**Scope calls are yours,** and I'd only note a preference on the last: splitting #4024 into "keys landed / combo targets open" seems right to me, since the key path is verifiable from the 429 body alone while the combo-target path needs the routing decision.

**Regex narrowing:** happy to tighten toward OpenRouter's exact phrasing if you want, though note it currently also accepts the `resets at` spelling and a bare date — both observed shapes. Say the word and I'll narrow it.

lidge-jun (commented):
Reviewed against current head `a137d4c33ae267693fc4be7d3a8b72f0763924b9`. The old response-clone finding is closed. The current implementation reads the original stream and reconstructs a response with the original status, status text, and headers (`src/providers/key-failover.ts:63-107`), replays already-pulled bytes, and streams the remainder lazily instead of buffering the whole body.

Four issues remain; the last two are merge blockers.

1. The 4 KiB read ceiling is not actually enforced. `src/providers/key-failover.ts:74-80` appends the entire chunk and checks the limit only before the next read. A single multi-megabyte chunk is therefore retained and decoded in full. The test at `tests/providers/openrouter-quota-reset-cooldown-4024.test.ts:125-146` uses 64 KiB chunks and only asserts that less than 1.25 MiB was pulled, so it passes while exceeding the documented 4,096-byte ceiling by at least 16x. Slice the boundary chunk at the remaining byte budget and replay its overflow before the unread remainder, then assert an exact `<= 4_096` pull/retention contract with a chunk larger than the cap.

2. `reader.read()` at `src/providers/key-failover.ts:75` has no deadline and receives no abort signal. A body that stalls before reaching the ceiling blocks the rotation path indefinitely. Thread the request abort signal into this helper and bound the peek with a short deadline; cancellation or timeout must terminate the read and preserve a usable response/error path rather than leaving the failover loop pending.

3. **MERGE BLOCKER:** the blanket catch at `src/providers/key-failover.ts:108-110` returns the original response while the reader created at line 69 still owns the stream lock. The caller therefore does not receive a usable response on that path. More seriously, if a fetch-backed read rejects because the client aborted, that catch converts cancellation into an ordinary “no body date” result; `src/server/responses/adapter-dispatch.ts:676-689` then continues into rotation, records cooldown, and persists a replacement key for a request the client cancelled. The reader is neither cancelled nor released. Distinguish abort from parse/read failure, propagate cancellation to the dispatch path, and use `finally` to cancel or release the reader so an error cannot return a locked body or mutate key state after client cancellation.

4. **MERGE BLOCKER:** `parseQuotaResetAt` is a broad regex over arbitrary response prose (`src/providers/key-failover.ts:150-176`), and `adapter-dispatch` invokes it for every 429 from every provider with a multi-key pool (`src/server/responses/adapter-dispatch.ts:667-685`; pool eligibility is only the generic 2+ key check at `src/providers/key-failover.ts:245-252`). Nothing confirms OpenRouter or its `rate_limit_error` quota-exhaustion envelope. An unrelated provider's generic 429 containing “resets at <date>” can therefore override a valid `Retry-After` and park a working key for up to 32 days. Gate body-date parsing on the canonical OpenRouter provider/capability and a validated OpenRouter quota error shape, leaving `Retry-After` authoritative for unrelated upstream prose.

Please also remove `Closes #4024` and keep #4024 open. This head adds reset-aware key parking to one Responses dispatch path for a multi-key pool. It does not implement single-key-to-combo failover, combo-target exhaustion, the other OpenRouter surfaces, configurable policy, or parking that survives restart. The cooldown map is process-local (`src/providers/key-failover.ts:200-205`), and pools with fewer than two keys return before recording any cooldown (`src/providers/key-failover.ts:523-526`). Closing the broader issue would hide those still-open requirements.

lidge-jun:
Landed via #4980 at e1717d631693b4d260ed3de84f19dded654e1fb9

lidge-jun:
Landed on `dev` as #4980, squashed as `e1717d631693b4d260ed3de84f19dded654e1fb9`, carrying your work with a `Co-authored-by` trailer so the commit is attributed to you in the contributor graph rather than only in prose.

Closing this one because the work is on `dev`, not because the contribution was unwanted — the OpenRouter quota reset cooldown hardening is yours and it is shipping. A review this round found a defect that had to be fixed before it could land, and rather than leave the branch waiting indefinitely on a round trip, the fix was made on a carry branch. The carry PR describes exactly what was changed relative to your branch and why, so the difference is reviewable rather than silent.

If you disagree with any part of the change made on top of your work, say so on #4980 and it can be revisited. Thanks for the contribution.

lidge-jun:
Landed via #4980 at e1717d631693b4d260ed3de84f19dded654e1fb9

lidge-jun:
Landed via #4980 at e1717d631693b4d260ed3de84f19dded654e1fb9

PR #4747, opened 2026-09-16
fix(tests): refuse removals that reach a protected home (#4705)

lidge-jun:
## Summary

The destructive-home guard recognised exactly one shape: a single-line `rmSync(getConfigDir())`, or a `const` bound from it on the same line. A multiline call, a `let` alias, a helper wrapper, a namespaced `fs.rmSync` and the sibling resolvers `getConfigPath()` and `usageLogPath()` all passed it, so a test could reintroduce the 2026-09-15 incident while CI stayed green. The specific quota test that caused that incident is correctly fixed; this is about the guard.

Three parts now hold it shut, because each covers what the others cannot see.

**An unconditional removal refusal** in `src/lib/test-home-guard.ts`. It is deliberately not gated on `OCX_TEST_HOME_GUARD`: Bun resolves `bunfig.toml`, and with it the preload, from the current working directory, so the run that caused the incident armed nothing and left `OPENCODEX_HOME` unset. It refuses a protected tree, any path inside one, and any ancestor of one, judged on both the canonical and the lexical form — the canonical form defeats a symlink alias, the lexical one still names the target when the protected path is itself a link. `removeTreeWithRetry` calls it before the injected remover runs, so the refusal lands before any filesystem call.

**A shared ownership fixture** in `tests/helpers/temp-home.ts`. `createTempHome()` pins both `OPENCODEX_HOME` and `CODEX_HOME` at a directory it created and returns a handle that owns it, so `removeOwnedTree()` can refuse a path nobody handed out. Provenance becomes a checkable property instead of four steps every fixture re-implements, and the step that goes missing is the pin.

**A widened source scan** in `tests/helpers/home-destruction-scan.ts`, replacing the inline line matcher. It blanks comments and string bodies while keeping template interpolations visible, propagates taint to a fixpoint through bindings and helper returns, and derives its resolver set from `src/` rather than a hand-written list, so a resolver added tomorrow is covered the day it lands. Two tiers: removing the home root is refused outright, and removing a path inside the home is allowed only in a file that created the home it pinned — which is what the seventeen existing fixtures already do, and what the incident's test did not.

The guard's own tests feed the scanner twenty-one adversarial sources, so an empty offender list is evidence rather than a broken regex.

Closes #4705

## Verification

[...]

PR #4720, opened 2026-09-15
chore(lanes): land the Codex account and catalog lane (nine PRs)

lidge-jun:
## Summary

Lane tip for the Codex account and catalog group of the 32-PR bug landing train. Cumulative chain: merges `origin/dev` at `45cfb04e9757`, then each member as a merge commit, so every member's own commits, authorship and `Co-authored-by` trailers survive rather than being rewritten.

Members, bottom to top:

| PR | Author | Change |
| --- | --- | --- |
| #4199 | @leonclab | Custom models are hidden when their provider is disabled |
| #4615 | @luvs01 | Loopback HTTP routed through Bun proxies is rejected for the catalog |
| #4576 | @luvs01 | Provider references survive a late history migration |
| #4569 | @luvs01 | Refresh-lock acquisition and release are scoped to file identity |
| #4144 | @remorser58 | Rate-limited warmup no longer advises reauthentication |
| #4571 | @luvs01 | The reset-credit consume response is bounded like every other read |
| #4178 | @luvs01 | A deferred Windows CLI inspection is reported instead of an absent candidate |
| #4564 | @luvs01 | Desktop restart membership and POSIX-only cases are repaired on Windows |
| #4610 | @luvs01 | The full Codex desktop restart scope is described |

#4144 sits below #4571 because both change `tests/codex-integration/codex-auth-api.test.ts` and #4144 also shares `docs-site/src/content/docs/guides/codex-integration.md` with #4576. Ordering does not avoid those conflicts — nothing does, because almost every pull request in this area appends to the same `structure/` owner docs — it only makes each conflict a single resolution at a known point.

Three merges conflicted and were resolved under two explicit rules: **add/add**, where the base block is empty and both sides added distinct content at one anchor, so keeping both is the only resolution that does not drop a contributor's work; and **extension**, where both sides appended to the same single base line, so the union is that line carrying both suffixes. A resolver that refuses anything else did the work, and it did refuse once before the extension case was recognised. Resolutions: #4569 across five `structure/` docs, #4571 on one test import, #4178 as two add/adds plus one extension.

Why a lane: hosted CI is roughly ten to fifteen minutes per pull request and the chain is cumulative, so this tip's run executes the content of all nine members. The maintainer authorised the tip-only model for this batch. Design and limits: `devlog/_plan/260915_bug_pr_landing_train/030_wp3_lane_stack.md`.

Merging with `--merge` rather than `--squash` is deliberate: a squash discards the ancestry and GitHub then marks each member closed instead of merged.

[...]

PR #4754, opened 2026-09-16
[WRONG BRANCH] fix(tests): walk the Lab activation chain instead of its first hop (#4704)

lidge-jun:
## Summary

The synchrony guard followed the direct callees of `startServer`. That catches `activateLab` becoming async, but `activateLab` calls `installLabAutomationRuntime` and `startAutomationIfEnabled` without awaiting them (`src/lib/lab-activation.ts:176,200`), so making either one async with an await before its registration call leaves `activateLab` parsing as perfectly synchronous. Every assertion stayed green while `startServer` returned before Lab was registered — the one ordering the window exists to protect, and the reason AGENTS.md pins it.

`walkActivationChain` now follows the chain from `activateLab` to a bounded depth, failing on any node declared async or carrying a body-level await. At the current tip it reaches 61 nodes with zero failures.

The depth-one comment warned that a deeper walk would treat dynamic dispatch as startup callees and produce false positives. What avoids that is what this walk refuses to follow:

- **Nested functions** are already skipped by `collectBodyLevelCalls`, so timer callbacks, shutdown hooks, promise continuations and the deferred route executor are not activation edges. They run later by construction.
- **Receiver calls are not followed.** A method that turns async cannot suspend its caller unless the caller awaits it, and that await is already reported on the caller's own body. Following `Map.get` or `record.staticDetach.push` would add noise and catch nothing.
- **Names imported from outside this repository are classified automatically.** Hand-listing every `join` and `readFileSync` in the chain would bury the six names that actually need human judgement, which are listed with a reason each and asserted to still be reachable.

`inspectActivationDeclaration` extends inspection to const arrows. The function-only inspector reported `activationKey` as missing, and a walk that read *missing* as *fine* would skip every const-arrow node in the chain. That needed its own return-type skipper: the existing `skipReturnType` treats a top-level `=>` as part of a function-*type* annotation and keeps scanning for a body brace that a concise arrow never has.

The walk is bounded three ways: a 200-node ceiling that fails rather than running away, a visited set, and an eight-hop limit on re-export resolution.

Closes #4704

## Verification

Local test suites, individual test files, `bun run typecheck`, `bun run test`, `bun run test:changed`, the dashboard build and dependency installation were **NOT RUN** — the maintainer forbids local suite execution in this lane, since a local test run previously deleted real `~/.opencodex` data.

What was done instead:

- **The walker was driven against the real sources read-only** and every assertion this PR adds was evaluated ahead of time: 16/16 match. That includes the exact failure strings, the classified-name equality, and the `body: " a"` value asserted for the arrow inspector.
- **All four mutation cases were exercised** through the injected source loader against the real `src/lib/lab-activation.ts` and `src/lab/automation/orchestrator.ts`: a nested callee declared async, an await added inside that callee, a suspension three hops down in `startLabAutomationScheduler`, and the arrow forms. Each is reported; on the same mutated source the depth-one scan still returns `{ found: true, async: false, awaitLines: [] }` for `activateLab`, which is the defect being fixed.
- **No false positives:** the unmutated chain produces zero failures across all 61 nodes.
- **Transpile check** of the changed file. No test was executed.
- Hosted CI for this layer runs on the lane tip, which contains this commit; this layer's own commit carries `[skip ci]` under the maintainer-approved DEV-STACK-08 tip-only policy.

## Checklist


lidge-jun:
## 리뷰 · 우선순위 78 / 80

이 PR은 Lab 동기 활성화 가드가 `startServer`의 **한 홉 직접 호출**만 보던 구멍을 막습니다. 지금 `dev`와 부모 #4747(rg1) 위에서도 `tests/lab/core-lab-boundary.test.ts`는 `activateLab`가 async가 되는지만 잡습니다. 그런데 실제 `src/lib/lab-activation.ts`에서 `activateLab`는 `installLabAutomationRuntime`와 `startAutomationIfEnabled`를 await 없이 호출합니다(대략 176·200행). 그 안쪽이 async로 바뀌고 registration 전에 await가 생겨도 `activateLab` 시그니처는 동기처럼 보여 기존 단언이 전부 초록인데, `startServer`는 Lab 등록 전에 돌아갈 수 있습니다. AGENTS.md가 지키려는 바로 그 순서입니다. 이 PR은 `walkActivationChain`으로 유계 깊이까지 따라가며, async 선언이거나 body-level await가 있는 노드를 실패시킵니다. const 화살표도 `inspectActivationDeclaration`로 보고, nested/timer/shutdown/receiver 호출은 따라가지 않아 옛 depth-one 주석이 걱정한 오탐 기계를 피합니다. 파일은 테스트 하나뿐입니다.

베이스는 `dev`가 아니라 `codex/rg1-destructive-home-guard`(#4747)입니다. ratchet/lab 스택 tip(rg2)이라 **혼자 `dev`에 올리면 안 됩니다**. 부모 #4747이 먼저 랜딩된 뒤 tip을 올려야 하고, #4720 이후 `dev` HEAD `b3035fe`와의 관계는 부모가 머지될 때 함께 정리하면 됩니다. 이슈 #4704를 닫는다고 본문에 명시돼 있습니다.

`tests/lab/core-lab-boundary.test.ts` walkActivationChain - 유계 깊이·외부 import 자동 분류·수신자 호출 비추적. 현재 tip에서 61노드 0실패라고 본문에 적힘. 로컬에서 재현은 부모 체크아웃이 필요.
inspectActivationDeclaration - 예전 function-only 검사기는 const 화살표를 missing으로 보고 walk가 건너뜀. return-type skipper도 화살표 본문 없는 경우를 맞춤. 회귀 포인트.
뮤테이션 4종 - nested async, callee 내부 await, 3홉 아래 startLabAutomationScheduler 중단, 화살표 형태. depth-one이 같은 뮤테이션에 초록인 것도 assert해서 “구멍의 증명”을 남김.
스택 - #4747(rg1, base dev) → **#4754(rg2 tip)**. 이후 rg3/rg4(#4766 등)와 이어질 수 있음. tip만 squash 머지 금지.

메인테이너의 판단이 필요한 지점
- #4747 → #4754 순으로 랜딩할지(권장), 아니면 rg 레인 전체를 한 landing PR로 묶을지.
- walk 깊이와 “외부 import 자동 분류” 화이트리스트를 문서/AGENTS에 한 줄로 고정할지.
- #4704를 tip 머지와 동시에 닫을지.

너의 추천
**WAIT-FOR-STACK**: 부모 #4747을 먼저 머지한 뒤 #4754를 머지하세요. 스테일 베이스로 혼자 `dev`에 올리지 마세요. 머지 후 #4704 클로즈. 내용은 KEEP급.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
Cascading downward. The lab synchronous-activation guard now walks the activation chain recursively instead of stopping one hop after startServer, with boundaries that keep it from reporting nested callbacks and receiver calls that cannot suspend the caller.

Evidence at the verified tip baaf691fb406ae7ee319805cbe848d6d99909b64 (tree `3728219b8b5c1758a768108e4b69b648fa0828f6`), from dispatch run 35053542913:
- test 1-4/4, macos 1-2/2 and macos control all completed with conclusion success, confirmed through the check-runs API rather than the check rollup. windows 1/6 through 6/6 all succeeded, as did gates, storage policy, api usage, docker smoke, keyring and npm-global on three platforms, and the aggregate `ci` check.
- A dispatch run is the evidence here rather than the pull-request run, and that is not incidental. This lane's tip layer changes only documentation, so the `changes` path filter skips every heavy job on a `pull_request` event while the aggregate check still reports green. Because the layers below carry `[skip ci]`, the pull-request signal alone would have covered nothing at all. The dispatch runs against the tip tree, which contains all four layers.
- The dispatch ran at commit c086fda3a78bd684eaf20a85d34f6046b7fdfed4, whose tree is byte-identical to this tip's tree. The lane re-absorbed `dev` from the bottom layer upward afterwards to keep each pull request's diff to its own layer (5 / 1 / 3 / 9 files); only the merge order changed, so the run remains valid evidence for this tree.
- windows 5/6 failed once with `EPERM ... rm ocx-management-auth-*` at `tests/helpers/remove-tree.ts:38` and passed on re-run. That is pre-existing Windows behaviour, not a regression from this lane: the guard added here runs once before the retry loop, line 38 is the existing rethrow-after-retries path, the raised error is rmSync's EPERM rather than a guard refusal, the identical code passed in run 35050601423, and `src/config/paths.ts:57` already documents `hardenConfigDir()` leaving `icacls.exe` holding the directory. The retry budget being too small under six concurrent Windows shards is filed separately as #4789.
- `git merge-tree --write-tree origin/dev <tip>` reports a clean merge.
- Ancestry verified so each layer closes as MERGED: rg1, rg2 and rg3 are all ancestors of this tip.

Chained-child stacks merge top-down, so this lands in the parent branch and cascades to `dev`.

Maintainer integration decision under MAINTAINERS.md / AGENTS.md: a maintainer with maintain or admin access may integrate into `dev` without a second maintainer approval, recording the decision and exact-head CI evidence.

PR #4749, opened 2026-09-16
[WRONG BRANCH] fix(service): decode schtasks output with the Windows text decoder (#4691)

lidge-jun:
## Summary

On a zh-CN host (ACP/OEMCP 936) with a CJK account name, `ocx service repair` and the dashboard repair/install buttons failed against a registration OpenCodex had created itself:

```
Service repair failed: Task Scheduler registration is not a recognized legacy OpenCodex definition; it was preserved for manual review.
```

Redirected `schtasks /query /xml` output follows the **console output code page of the spawning process tree**, not the XML document encoding. In any 936 context — including the no-console background service on a zh-CN host — those bytes are GBK. `decodeSchtasksOutput` probed UTF-16 and then fell back to a plain UTF-8 decode, so the CJK account name inside `<SessionStateChangeTrigger><UserId>` became U+FFFD. The correctly resolved expected identity `[SID, MACHINE\<name>]` then never matched the trigger scope, `windowsTaskRegistrationHealthy` returned false, and repair aborted at its recognition gate. The same mojibake rolled back fresh installs at post-create verification.

The fix is entirely in byte decoding, before any XML is parsed. `decodeSchtasksOutput` now delegates to `decodeWindowsTextBytes` (`src/lib/windows-text.ts`), the decoder this project already built for this class: UTF-16, then **strict** UTF-8, then the locale's legacy code page. It already fixed the sibling `whoami`/PowerShell decode in `src/lib/windows-user-principal.ts` (#2914, and #722 for CP949); this call site was the last one still ending in a lossy UTF-8 decode. The decoder selects the code page from the process locale rather than a console handle, which is why it also works in the no-console service context.

**Task ownership is deliberately untouched.** `windowsTaskTriggerScopeAcceptable` still requires an exact identity match, and the added tests assert that a different account and the mojibake spelling are both still rejected. Forgiving a replacement character there would let two different non-ASCII accounts collapse to the same value, which is a worse failure than the refusal it replaces.

Delegating also removes a latent UTF-16BE edge: the old local copy allocated `buffer.length - 2` bytes for an odd-length payload and left a trailing uninitialized byte, while the shared decoder rounds the payload down.

Stacked on #4746, so this PR targets `codex/win1-update-teardown`. Retarget to `dev` once the parent lands.

Closes #4691

## Verification

[...]

PR #4746, opened 2026-09-16
fix(update): keep a history-preflight refusal from aborting the update (#4718)

lidge-jun:
## Summary

On Windows, "ocx update" stopped the service, entered the pending shared teardown path, printed `Native restore refused: history_paginated_requires_native_writer`, and then aborted with `could not stop the running proxy`. The service was down, no listener was left, and the old package was still installed. Installing the same target by hand succeeded, which is what the reporter did.

The refusal itself is correct and is unchanged. The Codex history preflight runs **before** the config half of the restore, so it returns an envelope whose `config`, `catalog` and `history` artifacts are all `skipped` — nothing was attempted. `restoreSharedClientStateAfterStop` recognised only two shapes, a later `history` failure and everything else, so the refusal fell through to "everything else", `ocx stop` exited 1, and `decidePostStopUpdate` read 1 as a proxy that would not die. The reported lane is `bin/ocx.mjs`; the Bun updater imports the same decision module and had the same defect.

The obligation genuinely **is** outstanding here: config and catalog were never restored, so the client still points at the proxy that just stopped. Reusing the existing history-only outcome would have discharged the receipt and lost that, so this adds a third outcome rather than widening the second.

- `CodexNativeRestoreResult.historyPreflightRefusal` carries the refusal as a structured reason. The artifact states cannot carry it on their own: an ownership refusal and a desired-state skip produce the same three `skipped` values, and matching the human-readable message would put a safety decision on prose.
- `ocx stop` keeps the receipt, says so on stderr, and exits **80**. Eighty is deliberately not 79: seventy-nine means the teardown ran and only history metadata is pending, and a caller reading it discharges the obligation.
- Eighty is emitted only when `pendingTeardownsAreExactly` confirms the obligations left in the home are exactly the ones this run chose to keep. A quarantined receipt or a concurrent stop's claim falls back to exit 1, which is the pre-existing behaviour, so the fallback loses nothing that used to work.
- `decidePostStopUpdate` lets 80 past the teardown gate and nothing else. Surviving runtime records, a live proxy and an unreadable liveness probe abort exactly as before, because a history refusal is evidence about history and says nothing about whether the proxy is gone.
- Both updater lanes report the deferral as its own outcome instead of reusing the manifest warning, which would imply config and catalog came back.

History protection is not weakened anywhere: `preflightCodexHistoryInjection` is untouched, it still refuses before any config, catalog, manifest, SQLite or rollout mutation, and the backup manifest is still retained for review.

[...]

PR #4759, opened 2026-09-16
fix(service): stage elevated Task Scheduler XML instead of inlining it (#4692)

lidge-jun:
## Summary

When `ocx service repair` re-registered the Task Scheduler task through the **elevated** fallback, the spawn failed before UAC ever appeared:

```
WindowsElevationError: ENAMETOOLONG: name too long, uv_spawn
    at startPowerShellCommand (src/lib/windows-elevation.ts:560)
    at runWindowsElevatedScheduledTaskRegistration (src/lib/windows-elevation.ts:704)
    at registerFreshWindowsSchedulerTask (src/service/windows-ops.ts:313)
```

`runWindowsElevatedScheduledTaskRegistration` embedded the new task XML **and** the expected-existing snapshot as base64(utf16le) inside an inner PowerShell script, which was then base64(utf16le)-encoded again into `-EncodedCommand`. Two base64 layers over UTF-16 cost roughly 14.2 command-line characters per XML character, and a replacement carries two payloads, so a ~2 KB definition put the outer command past the Windows limit. On a host where Task Scheduler exports the trigger scope as an account name (#4691) `identityUpgradeNeeded` stays true, so the re-register path runs on **every** repair and repair could never exit 0.

Both payloads are now staged to files, and the command carries two paths and two 64-character digests. Its length no longer depends on the size of the XML at all.

A file an administrator process will read is itself a privilege-escalation surface, so three properties hold together and none of them is sufficient alone:

- **Access.** The staging directory is created fresh by `mkdtemp` and ACL-hardened through the existing `hardenSecretDir`/`hardenSecretPath` **before** anything is written into it, so the payload is private from the moment it exists.
- **No redirection.** Each artifact is inspected with `lstat` and rejected unless it is what it claims to be. Exclusive `wx` creation inside a directory that did not exist a moment ago is the atomic step — there is no replace path to race — and the explicit check keeps that guarantee from resting on a reading of `O_EXCL` semantics.
- **Tamper evidence.** The digest covers the exact bytes written, and the elevated script reads the file once, hashes what it read, and refuses before decoding. An ACL cannot cover this: a process running as the same user has the same SID and can rewrite the file, so the digest is what makes a swap during the UAC prompt fail closed instead of registering a different definition.

Cleanup runs on every exit — success, UAC cancellation, a synchronous spawn failure, a failed digest check, and a partial staging failure — and a cleanup error is aggregated with the registration error rather than replacing it.

The original "immutable bytes, never a caller-writable pathname" goal is kept by different means rather than abandoned, and **the replacement precondition is untouched**: the elevated process still re-queries the live registration and compares it to the verified predecessor before passing `-Force`. All three existing safety checks remain — capture-and-validate before elevation, re-check the predecessor immediately before UAC, and re-query and compare after consent.

Payloads are UTF-16LE with no BOM and are decoded straight into `Register-ScheduledTask`, so what is hashed is exactly what is registered, with no trimming step the two sides could disagree about.

This is the tip of a three-layer stack (#4746 -> #4749 -> this), so it targets `codex/win2-schtasks-gbk`. Retarget to `dev` once the parents land.

Closes #4692

## Verification

**No local test suite, single test file, typecheck, build or install was run — the repository owner prohibits it for this lane.** Local verification is explicitly NOT RUN. The Windows job in this PR's hosted CI run is the only platform evidence for all three layers in the stack; the two parent layers carry `[skip ci]` by design, so this run gates the whole lane.

Static verification performed:

- Traced the failing path: `repairService` -> `reregisterWindowsSchedulerTask` -> `registerFreshWindowsSchedulerTask` -> structured `create`/`access-denied` -> `runWindowsElevatedScheduledTaskRegistration` -> `startPowerShellCommand`, and confirmed the access-denied classification and the UAC-cancellation mapping are correct and unchanged; only the transport was wrong.
- Derived the growth analytically before changing anything: `inner = C + B(N) + r*B(M)` and `outer = 4*ceil(2*inner/3)`, giving about `14.22 * N` for a replacement with equal-sized payloads, so a 2 KB definition contributes roughly 29,000 characters against the 32,767 limit. That is the arithmetic the size-independence test now pins.
- Confirmed the two remaining production callers both hold an already-validated string and both route through the new stage/cleanup wrapper, and that the injected `deps.elevate` seam keeps its string signature so existing capture-before-elevation tests still exercise the same invariant.
- Verified the elevated helper hashes and decodes the **same** byte array from a single `ReadAllBytes`, since hashing a path and reopening it would reintroduce the swap window.

Regression tests added or updated (they run in CI, not locally):

- `tests/windows/windows-elevation-spawn.test.ts` asserts the elevated script carries the staged paths and digests, performs `ReadAllBytes` -> `ComputeHash` -> digest comparison **before** decoding, contains neither XML's base64 nor any `FromBase64String`, and that the outer command length is identical across payloads and stays under a conservative 8 KB bound for both the create and the replace shapes. It also pins that a replacement without a captured predecessor still throws.
- `tests/service/service.test.ts` pins the staging contract: the directory is hardened before either payload is written, the digest equals a SHA-256 computed independently over the bytes actually on disk, the bytes are UTF-16LE with no BOM, cleanup is idempotent and removes everything, a redirected path is refused without hardening it, and a failure on the *second* payload still removes the first.

Not provable without a real Windows host, and not claimed: the exact quoting libuv produces before `CreateProcessW` and the precise length at which this Bun version reports `ENAMETOOLONG`; PowerShell 5.1 behaviour for `SHA256`/`Unicode.GetString` in this launcher chain; split-token UAC access to the current-user-hardened file on every account type; NTFS ACL and reparse behaviour on a real volume; and antivirus interference with cleanup timing.

One behavioural limitation worth calling out for review: `hardenSecretPath` grants the current user's SID and strips inheritance, so the staged files are readable by a split-token elevation of the same user (the normal case) but not by an elevation performed with a *different* administrator's credentials. The previous inline form had no such dependency. Widening the ACL to SYSTEM/Administrators would be a deliberate change to a security-sensitive module and is deliberately not bundled here.

## Checklist

- [x] Docs or release notes were updated when needed. No user-facing documentation describes the elevated registration transport; the rationale lives in the doc comments on `stageElevatedSchedulerRegistration` and `runWindowsElevatedScheduledTaskRegistration`.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. **This change adds a file that an elevated process reads, so it warrants explicit security review per MAINTAINERS.md.** The staged payloads contain a task definition, not a credential; the directory is hardened before first write; paths are lstat-checked against redirection; the digest makes same-account tampering fail closed; cleanup covers every exit path; and the post-UAC compare-before-`-Force` precondition is unchanged. No credential, token, OAuth, workflow or release-automation surface is touched, and nothing new is logged.

## Prior-art check

No existing pull request, open or closed, implements this fix. Searched the repository's pull requests by issue number and by implementation signature, and inspected the adjacent antecedents (#3040, #4313, #2918, #3067) — each addresses a different problem and no code is carried from any of them. No `Co-authored-by` trailer is therefore owed. Recording the check here so the question does not have to be reopened.

## Follow-up

The ACL limitation described above is tracked as #4779: whether the staged payloads should also grant SYSTEM and Administrators is a change to a security-sensitive module and is deliberately a separate decision. This PR makes the failure diagnosable rather than silent — the elevated read failure has its own protocol exit code and the parent reports the cause and the remedy instead of a bare number, so the limitation cannot present as an unexplained "access denied".

## CI evidence (lane tip)

Hosted CI at the exact tip head `8a1b01011dbf68cd5eef07c568281590981fcaea`:

- **Cross-platform CI, `workflow_dispatch` lane=all — success**: [run 35054231781](https://github.com/lidge-jun/opencodex/actions/runs/35054231781). All 26 jobs succeeded, with no non-success conclusion. `test 1/4`–`4/4` and `windows 1/6`–`6/6` are all `completed success`.
- The Windows suite is the reason for the dispatch: `platform-windows` is gated on `github.event_name == 'workflow_dispatch'`, so it never runs on a `pull_request` event. Since all three layers of this lane are Windows-only, a PR-event run alone would carry no platform evidence for them.
- `enforce-target`, `PR hygiene` and `PR Labeler` are all green at the same head.

An earlier `pull_request` run at the previous head `6a2b148f5a` was also green ([run 35051066482](https://github.com/lidge-jun/opencodex/actions/runs/35051066482)); the dispatch run above is a strict superset of its job set.

One earlier dispatch surfaced a `windows 2/6` failure whose nine assertions were all in `tests/clients/desktop-app-restart-posix.test.ts`. That defect is not from this lane: the identical file and the identical nine line numbers appear on the dev-only dispatch [34795291889](https://github.com/lidge-jun/opencodex/actions/runs/34795291889) from 2026-09-14. It is fixed on `dev` by `f79c147309`, which landed after this lane's original base, so `dev` was merged up the chain to absorb it. Verified by blob: this tree carried `src/codex/desktop-app/windows.ts` at `863a20057a` before and `c73e067e5b` after.


lidge-jun:
## 리뷰 · 우선순위 77 / 80

이 PR은 Windows에서 `ocx service repair`가 elevated 폴백으로 Task Scheduler 작업을 다시 등록할 때, UAC가 뜨기도 전에 `ENAMETOOLONG: name too long, uv_spawn`으로 죽던 버그(#4692)를 고칩니다. 지금 `dev` HEAD(`3070d64d8`, 패키지 2.57.0)의 `runWindowsElevatedScheduledTaskRegistration`(`src/lib/windows-elevation.ts`)는 새 작업 XML과(교체 시) 기존 스냅샷 XML을 둘 다 base64(utf16le)로 안쪽 PowerShell에 넣고, 그 스크립트를 다시 base64(utf16le)로 `-EncodedCommand`에 실습니다. UTF-16 위에 base64가 두 겹이면 대략 XML 글자당 명령줄 14자 가까이 불어나고, 교체 경로는 페이로드가 두 개라 ~2KB 정의만으로도 Windows 한계(32767)를 넘기기 쉽습니다.

더 아픈 점은 #4691과 겹칩니다. Task Scheduler가 트리거 범위를 계정 이름으로 내보내는 호스트에서는 `identityUpgradeNeeded`가 수리마다 true라 재등록 경로가 매번 타고, 원래 작업이 관리자 소유면 비상승 `schtasks /create /f`는 Access denied → 결국 elevated만 남습니다. 그래서 수리(exit 0)가 영원히 안 됩니다. 이 PR은 페이로드를 명령줄에 넣지 않고, `mkdtemp`로 만든 스테이징 디렉터리에 UTF-16LE(BOM 없음) 파일로 쓴 뒤 경로+64자 SHA-256만 명령에 실어 명령줄 길이가 XML 크기와 무관해지게 만듭니다.

보안 모델도 같이 바뀝니다. 예전 목표는 “불변 바이트, 호출자가 쓰는 pathname은 안 넘김”이었고, 지금은 그 목표를 다른 수단으로 지킵니다. (1) 디렉터리를 먼저 `hardenSecretDir`/`hardenSecretPath`로 ACL 잠근 뒤 쓰고, (2) `wx` 생성 + `lstat`로 리다이렉트/심링크를 거절하고, (3) elevated 쪽이 `ReadAllBytes` 한 번으로 읽은 바이트를 해시·검증한 뒤에만 디코드합니다. 같은 사용자 SID면 ACL만으로는 UAC 대기 중 파일 교체를 막지 못하니, digest가 fail-closed의 핵심입니다. 교체 전제조건(라이브 등록 재조회 → 캡처된 predecessor와 비교 → 그다음에야 `-Force`)은 그대로입니다.

코드 경로는 `stageElevatedSchedulerRegistration` + `runStagedElevatedSchedulerRegistration`(`src/service/windows-ops.ts`)이 스테이징·정리·에러 합치기를 맡고, `registerFreshWindowsSchedulerTask` / `restoreWindowsSchedulerTaskIfAbsent`의 elevate seam이 그 래퍼로 바뀝니다. `service.ts`는 `stageElevatedSchedulerRegistration`을 re-export합니다. 테스트는 `windows-elevation-spawn.test.ts`에서 인라인 base64/`FromBase64String` 부재·digest-before-decode·명령 길이 불변(<8KB)·predecessor 없는 replace 거절을 고정하고, `service.test.ts`에서 harden-before-write·디스크 바이트 SHA-256·리다이렉트 거절·두 번째 페이로드 실패 시 첫 파일 정리를 고정합니다. 로컬 스위트/타입체크는 이 레인 금지이고, 부모 레이어가 `[skip ci]`인 스택이라 이 PR의 Windows 호스티드 CI가 세 층 전체의 플랫폼 증거가 됩니다.

현재 `dev` 대비 방향은 맞고, types/config 스플릿에 무효화되는 모놀리스 편집도 아닙니다. 다만 베이스가 `dev`가 아니라 `codex/win2-schtasks-gbk`(스택 #4746 → #4749 → 여기)이고, `mergeable_state: unstable`이며 Cross-platform/Service lifecycle Windows job이 아직 도는 중입니다. PR이 밝힌 ACL 한계(현재 사용자 SID만 열어 두면, 다른 관리자 자격으로 elevate하면 스테이징 파일을 못 읽을 수 있음)도 의도적 스코프 밖이지만 메인테이너가 한 번 짚을 만합니다.

라인 ~670+ (windows-elevation.ts `READ_STAGED_TASK_XML`) - 한 번 읽은 바이트로 해시→비교→디코드 순서가 맞고, 경로를 다시 여는 창을 닫았습니다. 경로·digest는 `psSingleQuote`로 들어가므로 일반적인 `mkdtemp` 경로에서는 안전해 보입니다.
경로 stageElevatedSchedulerRegistration (windows-ops.ts) - digest는 디스크를 다시 읽지 않고 메모리 `bytes`에 대해 계산합니다. 기본 `writePayload`는 그 바이트를 그대로 쓰므로 일치하고, 테스트도 온디스크 해시로 고정합니다. 주입 seam이 다른 바이트를 쓰면 elevated에서 fail-closed됩니다.
경로 hardenSecretPath ACL - 현재 사용자 SID + inheritance strip이라 split-token 같은 사용자 elevate는 되지만, 다른 관리자 자격 elevate는 스테이징을 못 읽을 수 있습니다. 예전 인라인 형태에는 없던 의존성입니다.
경로 base `codex/win2-schtasks-gbk` - `dev` 직접 머지 대상이 아닙니다. #4746·#4749가 먼저 안착해야 리타겟/머지 순서가 열립니다.
경로 Windows CI / Service lifecycle - 이 팁이 스택 전체의 플랫폼 게이트입니다. 그린 전에는 머지 증거가 부족합니다.

메인테이너의 판단이 필요한 지점
- #4746 → #4749 → 이 PR을 스택 순서 머지할지, 부모만 `dev`에 올린 뒤 이 브랜치를 `dev`로 리타겟할지
- 다른 관리자 자격 elevate를 지원하려면 SYSTEM/Administrators ACL 확장을 별도 보안 PR로 열지, 지금처럼 같은 사용자 split-token만 지원할지
- MAINTAINERS.md가 요구하는 elevated-read 파일 보안 리뷰를 이 댓글/코드리뷰로 충분하다고 볼지, 별도 체크리스트 사인오프를 남길지

너의 추천
열어 두고 부모 #4746·#4749와 이 PR의 호스티드 Windows CI 그린을 기다린 뒤 스택 순서 머지(또는 부모 안착 후 `dev` 리타겟)하세요. 인라인 base64로 되돌리거나 교체 precondition을 느슨하게 만들지 마세요. Closes #4692는 머지 시 함께 닫히면 됩니다. types/config 스플릿 close-dont-rebase 대상 아닙니다.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
Landing the Windows service and update recovery lane, top layer first. Elevated Task Scheduler re-registration no longer passes the XML on the command line, so the payload size stops deciding whether the spawn exceeds the limit. The staging directory is ACL-hardened before anything is written to it, and the elevated script verifies the bytes it read by digest rather than reopening a path.

Evidence at the verified tip 8a1b01011dbf68cd5eef07c568281590981fcaea (tree `5fb70432366550aab0177be5255b5097379d7935`), from dispatch run 35054231781:
- Every check-run at this commit concluded success. There is no failing, cancelled or pending check outside the always-skipped matrix placeholders.
- test 1-4/4 and macos 1-2/2 all completed with conclusion success, confirmed through the check-runs API rather than the check rollup.
- windows 1/6 through 6/6 all completed with conclusion success. That evidence is required rather than incidental here: every change in this lane is Windows-specific, and the `platform-windows` job is `workflow_dispatch`-only, so a `pull_request` run never exercises it.
- The current head carries no `pull_request` run because the three branches were pushed together and the base and head of the tip moved almost simultaneously, so no synchronize event fired. The job sets were compared directly: the dispatch run covers 26 jobs against the pull-request run's 21, and no real job present in a pull-request run is absent from the dispatch. The only difference is a skipped matrix placeholder. The dispatch is a superset, at the exact head.
- The lane absorbed `dev` at 5e3029e6fd from the bottom layer upward, so each pull request keeps its own layer diff (13 / 2 / 4 files). Before that absorption, windows 2/6 failed on nine desktop-restart assertions that were live in `dev` and repaired by #4564; the same shard passes here.
- `git merge-tree --write-tree origin/dev <tip>` reports a clean merge.
- Ancestry verified so each layer closes as MERGED: win1-update-teardown and win2-schtasks-gbk are both ancestors of this tip.

Chained-child stacks merge top-down, so this lands in the parent branch and cascades to `dev`. CI evidence transfers by tree identity at each step.

Maintainer integration decision under MAINTAINERS.md / AGENTS.md: a maintainer with maintain or admin access may integrate into `dev` without a second maintainer approval, recording the decision and exact-head CI evidence.

PR #4560, opened 2026-09-13
feat(gui): redesign provider accounts workspace with adaptive grid, collapsible rails, and quota controls

agentHits:
## Summary

Comprehensive redesign of the **Provider Accounts workspace** (specifically optimizing Google Antigravity and multi-account setups) to dramatically enhance usability, information hierarchy, and responsiveness across all desktop and mobile viewports, plus integrated **Cockpit Tools** token import:

### Motivation & Project Roadmap
- **Community & Personal Need**: I initially designed this for my own multi-account daily workflow, but decided to contribute it upstream so the entire OpenCodeX community can enjoy a much more intuitive, dense, and beautiful interface.
- **Part 1 of 2 (UI Foundation & Import)**: This PR represents **Part 1** — building a solid, clean, and responsive visual foundation for provider accounts, quota windows, live filtering, and seamless account onboarding (including Cockpit Tools).
- **Part 2 (Upcoming Account Pool Rotation)**: Once this visual foundation is in place, Part 2 will implement automatic multi-account pool rotation for Google Antigravity (automatically switching to the next ready account in the pool when 5-hour or weekly quota limits are reached on the active account). My proposal was shared over two weeks ago without an upstream implementation, so I took the initiative to build it step by step starting with this PR.
- **Maintainer Freedom**: Everything has been thoroughly tested locally across all supported resolutions and locales with zero CI/hygiene regressions. If maintainers wish to adapt, refactor, or tweak any part of the implementation to match project architecture preferences, please feel free to do so! If merged, great for all users; if not accepted, no hard feelings at all — I will happily maintain this enhanced UX on my fork on top of upstream releases.

### Key Changes
1. **Responsive Two-Column Grid**: Accounts cards adapt dynamically via Container Queries (`repeat(auto-fit, minmax(min(100%, 460px), 1fr))`), collapsing smoothly to single-column on narrower viewports or when sidebars are expanded.
2. **Independent Dual Collapsible Sidebars**:
   - **Provider Rail (58px)**: Toggle button in search row to collapse providers into vertical icon buttons with status dots and tooltips.
   - **Global Sidebar (68px)**: Collapse navigation items into icon-only mode with centered action buttons.
   - Both states independently persist across page reloads in `localStorage`.
3. **Unified Dropdown Filter with Live Counters**: Consolidated separate filter buttons into a single clean dropdown with real-time counts for active quotas (Gemini/Claude), all accounts, and exhausted states.
[...]

PR #4824, opened 2026-09-16
fix(combos): allow single-target combo with waitForCooldownMs to retry after cooling

agentHits:
## Summary

When a combo is configured with a single target and has `waitForCooldownMs > 0`, a retryable upstream error (such as HTTP 503 or 429) previously caused an immediate failure. The target was placed into cooldown and recorded in `pick.attempted`, after which `pickWithWait({ exclude: pick.attempted })` immediately returned `null` because the only target was excluded. Consequently, `waitForCooldownMs` was never honored during the active request for single-target combos.

This change allows a single-target combo with `waitForCooldownMs > 0` to wait for its cooldown to expire and retry the target within the request execution budget. It also explicitly bounds the retry via `comboTargetsDispatched <= 1` alongside the dispatch budget scope, and documents loop termination mechanics.

## Verification

- Added regression tests in `tests/server/server-combo-failover-e2e.test.ts`:
  - `single-target combo with waitForCooldownMs waits and retries on failure` (verifies cooldown wait and successful retry)
  - `single-target combo with waitForCooldownMs stops after one retry when upstream fails continuously` (verifies loop terminates after exactly 1 retry when upstream fails persistently)
  - `single-target combo with unset waitForCooldownMs fails immediately on 503` (verifies fast-fail when wait budget is 0)
- Ran focused test suite:
  - `bun test tests/server/server-combo-failover-e2e.test.ts` (174 passed, 0 failed)
  - `bun test tests/codex-integration/combos.test.ts` (89 passed, 0 failed)
- Ran typecheck & hygiene:
  - `bun run typecheck` (zero errors)
  - `bun run structure:check` (SSOT checks passed)
  - `bun run privacy:scan` (privacy scan passed)

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Single-target combo requests can retry once after the configured cooldown when eligible, improving recovery from temporary service failures.
  - Requests that continue failing stop after the retry limit and return the failure response.
  - Requests without cooldown waiting continue to fail immediately after the initial attempt.

- **Documentation**
  - Clarified that configured cooldown waiting can hold a single-target request and retry the same target.

- **Tests**
  - Added coverage for successful cooldown retries, repeated failures, and disabled retry waiting.


agentHits:
Hi @lidge-jun and @Ingwannu 👋

Just wanted to provide additional context on why this fix is crucial for smooth failover and eliminating unexpected stalls/crashes:

### The Problem
When using single-target combos (or dedicated routing targets with `waitForCooldownMs > 0`), any momentary upstream burst error (HTTP 503 or brief 429) previously caused an instant failure. The single target was put into cooldown and added to `attempted`, so `pickTarget()` returned `null`. The request terminated immediately, completely ignoring the configured `waitForCooldownMs` wait budget. In Codex CLI and IDEs, this manifests as an abrupt session drop or stall, forcing the user to manually retry.

### The Fix & The Bigger Picture
This minimal 6-line patch allows single-target combos to honor `waitForCooldownMs`, gracefully waiting for the brief cooldown within the active request budget before retrying. The request recovers seamlessly with 200 OK without user intervention.

This fix directly complements our broader multi-account pool and rotation work (PR #4560, PR #4676, and #3375). I already have the complete, seamless account pool rotation system running locally and fully verified with zero-downtime failover. Landing this clean bugfix ensures that temporary single-target cooldowns recover smoothly without stalls while the broader pool PRs are moving through review.

All 4/4 readiness checks and local tests pass green. Thank you! 🚀

lidge-jun:
## 리뷰 · 우선순위 71 / 80

이 PR은 콤보 타겟이 **하나**이고 `waitForCooldownMs > 0`일 때, 재시도 가능한 업스트림 실패(503/429 등) 뒤에도 cooldown 대기를 안 하고 바로 실패하던 구멍을 막는다. 지금 `dev` HEAD `6ed8986c64afb2e37b1055e1afeb7e12d32ca8e4`의 `src/server/responses/core-combo.ts` 실패 분기(대략 748–756행)는 `advanceComboAfterFailure`가 다음 타겟을 못 고르면 `pickWithWait({ exclude: pick.attempted, ... })`를 부른다. 단일 타겟이면 그 타겟이 이미 `attempted`에 들어가 있어서 exclude 때문에 즉시 `null`이 되고, 설정해 둔 `waitForCooldownMs`는 이번 요청 안에서 한 번도 쓰이지 않는다. 다중 타겟 페일오버는 다른 타겟으로 넘어가니 드러나지 않고, 단일 타겟+대기 설정만 조용히 깨진다.

패치는 그 `pickWithWait(exclude: attempted)`가 빈손일 때, 단일 타겟이고 `waitForCooldownMs > 0`이며 abort되지 않았으면 exclude 없이 `pickWithWait`를 한 번 더 부른다. 그러면 cooldown이 풀릴 때까지 기다렸다가 같은 타겟을 다시 고를 수 있다. `tests/server/server-combo-failover-e2e.test.ts`에 회귀 두 개가 있다. waitForCooldownMs가 있으면 503 뒤 200으로 회복·hits 2, 없으면 즉시 503·hits 1. 동작과 비동작을 같이 고정한 점이 좋다. types/config 분리와 무관하고, 프로덕션 변경은 핵심 파일 한 곳·여섯 줄이라 범위도 작다.

한 가지 확인할 점은, 두 번째 `pickWithWait`가 exclude를 빼므로 "attempted라서 영구 제외"가 아니라 "cooldown만 보고 다시 고르기"가 된다는 의도인데, 바깥 루프·실행 예산이 재시도를 몇 번까지 허용하는지는 기존 `pickComboTargetWithWait` / request budget에 의존한다. 테스트는 hits===2로 한 번 회복만 본다. abortSignal 가드가 있는 것은 맞다. 기여자 PR(`agentHits`)이고 `dev` 방향(콤보 페일오버·쿨다운)과 잘 맞는다.

`src/server/responses/core-combo.ts` 753 근처(PR 추가 블록) - 단일 타겟+waitForCooldownMs일 때만 exclude 없는 재pick. 다중 타겟 경로와 abort 경로는 그대로 둔다.
같은 파일 751–755 - 기존 `exclude: pick.attempted` 호출이 단일 타겟에서 wait를 무력화하던 직접 원인이다. 패치가 그 바로 아래를 보완한다.
`tests/server/server-combo-failover-e2e.test.ts` 새 케이스 두 개 - 긍정/부정 회귀가 명확하다. cooldown 50ms·wait 500ms는 e2e에 적당한 짧은 값이다.
재시도 횟수 상한 - 이 패치 자체에 "몇 번까지"는 없고 기존 루프/예산에 맡긴다. 긴 장애에서 요청이 wait 예산만큼 늘어날 수 있다.

메인테이너의 판단이 필요한 지점
- 단일 타겟이 cooldown 대기 후 다시 실패하면 루프가 예산을 소진할 때까지 반복하는 현행 동작이 원하는지, 명시적 max-retry가 필요한지
- `combo.targets.length === 1` 조건 대신 "exclude 때문에 후보가 비었고 waitForCooldownMs>0"처럼 다중 타겟 전멸+대기 경우까지 일반화할지

너의 추천
CI 초록이면 합쳐라. 버그 설명이 코드와 일치하고 회귀 테스트가 양쪽을 고정한다. 머지 후 긴 장애에서 단일 타겟 요청이 wait 예산만큼 늘어나는지만 한 번 보면 충분하다. types/config 닫기 대상 아니다.

이 댓글은 grok-bot이 작성했습니다

abhisheksharma2411 (commented):
Correct diagnosis, and the two tests are a good pair — the second one pinning that an unset `waitForCooldownMs` still fails immediately is what makes the first safe to land.

The thing I wanted to check is that this stays bounded. `exclude: pick.attempted` is what normally terminates the `while (pick)` ladder: every target accumulates into `attempted`, so a multi-target combo eventually picks `null`. The new branch deliberately drops `exclude`, which removes exactly that guarantee for the single-target case — so if the target never recovers, what stops the loop?

Measured it rather than guessed, with your own harness and an upstream that 503s forever:

```
PROBE_RESULT hits=2 status=503 elapsedMs=110
```

One retry, then out — `174 pass / 0 fail` on the suite. The bound is the hop budget at the top of the loop: `comboSendScope?.reserveDispatch({ sendClass: "combo-failover", ... })` refuses the second failover and `break`s, returning the last real upstream answer rather than spinning. So the ladder is terminated by budget rather than by target exhaustion, and the behaviour is right.

**Worth one line in the code saying that**, because it is not locally obvious: the comment above the new branch explains *why* the retry is wanted but not *what* stops it, and the `exclude` that a reader would expect to be the stopper is conspicuously absent. Someone widening this later — say to two-target combos, or lifting it out of the `!pick` branch — would be removing a bound without realising the budget is the only thing left.

The one case I could not exercise: `comboSendScope` is `undefined` when `options.sendBudget` is not a `RequestExecutionBudget` (L196), and then `hopDecision` is `undefined` and the `break` never fires. I could not construct that through the e2e harness, so treat it as a question rather than a finding — is that path reachable for a combo request, and if so does anything else bound the retry there?

Nice, small fix otherwise.

agentHits:
Thanks @abhisheksharma2411 and @lidge-jun for the thoughtful review and measurements! 🙌

Pushed an update addressing all points:

### 1. Documenting loop termination & local bound (`comboTargetsDispatched <= 1`)
Added a detailed comment in `src/server/responses/core-combo.ts` explaining both why `exclude` is omitted for single-target combos and what stops the ladder:
- **Primary dispatch bound:** `comboSendScope?.reserveDispatch` refuses a second failover hop via `comboExecutionBudgetPolicy(1)` (`maxAlternateTargetSends: 1` for a single declared target).
- **Local bound (`comboTargetsDispatched <= 1`):** Answers the question regarding what happens if `options.sendBudget` is not a `RequestExecutionBudget` (e.g., in direct unit invocations where `comboSendScope` is undefined). The `comboTargetsDispatched <= 1` guard ensures that the cooldown retry is executed at most once per request regardless of whether a budget scope is present, and avoids sleeping an extra cooldown period before breaking.

### 2. Regression test for persistent failure
Added `single-target combo with waitForCooldownMs stops after one retry when upstream fails continuously` to `tests/server/server-combo-failover-e2e.test.ts`, asserting that continuous 503s stop after exactly one retry (`hits === 2`) and return 503 without spinning.

### 3. Multi-target combos
Keeping this scoped to `combo.targets.length === 1` is deliberate: multi-target combos already have distinct alternate targets for failover, and sequential cooldown re-attempts across multiple exhausted targets would compound request latency and exceed transition budgets.

All checks and tests pass green (174/174). Ready for review! 🚀

lidge-jun:
## Review — termination holds, and the three bounds are genuinely independent

I checked this against one question: do the wait, the cancellation and the retry count all
terminate? They do, and the comment in the diff is right that the guard is doubled — but the two
bounds it names are worth stating precisely, because one of them is not where the comment says it
is.

### Retry count

The new branch requires `comboTargetsDispatched <= 1`, and the dispatch it enables runs
`comboTargetsDispatched += 1` at `core-combo.ts:463`, making it `2`. The branch cannot fire a
second time for the same request. That bound is local and does not depend on a budget existing.

The budget bound is real but arrives one layer up rather than through `comboTargetSendBudget`.
`comboExecutionBudgetPolicy(1)` yields `maxAlternateTargetSends: Math.max(1, 0) = 1`
(`core-combo.ts:107-108`), and the retry is the first non-`initial` dispatch, so
`reserveDispatch` admits it and would refuse the next one at
`src/lib/request-execution-budget.ts:244`. Worth noting explicitly: this means a single-target
combo now spends its one alternate-target slot on *itself*. That is defensible for a combo with
no alternate to move to, but it is a change in what that slot means.

The per-target clamp is unaffected. On the retry,
`combo.targets.length - 1 - comboTargetsDispatched` evaluates to `-1`, and
`comboTargetSendBudget` clamps it with `Math.max(0, targetsDeclaredAfterThisOne)`, so nothing is
held back and nothing goes negative.

### Wait

`pickComboTargetWithWait` computes `delay = earliest.expiry - now` and returns `null` outright
when `delay > options.waitForCooldownMs` (`src/combos/resolve.ts:399-401`). The sleep is therefore
never longer than the configured ceiling, whatever `Retry-After` the provider sent, and
`src/combos/types.ts:158-162` validates that ceiling to `0..600000`.

### Cancellation

The local `pickWithWait` wrapper forwards `options.abortSignal`, and
`pickComboTargetWithWait` returns `null` both when the sleep rejects and when the signal is
already aborted on wake. The caller then answers `clientCancelledResponse()`. The explicit
`!options.abortSignal?.aborted` in the new condition is redundant with that but harmless, and it
does save an unnecessary wait.

### Default behaviour

`COMBO_DEFAULT_WAIT_FOR_COOLDOWN_MS` is `0` (`src/combos/types.ts:6`), so a combo that never
configured a wait is untouched. The third test pins exactly that, which is the test I would have
asked for if it were missing.

### No replay of committed work

The branch sits after `comboFailureDecision` already returned `hop`, on a non-2xx failure path. A
streamed attempt that committed output returns before reaching it. So this does not introduce a
path that re-sends a turn whose output the client already saw.

### Two things I would still change

**Say what this does to the user-visible contract.** A single-target combo that previously failed
immediately now holds the client for up to `waitForCooldownMs` before answering. The reference
entry at `docs-site/src/content/docs/reference/configuration/routing.md:92` describes
`waitForCooldownMs` in terms of "the earliest eligible cooling target", which remains literally
true, but it was written when only a multi-target combo could wait. One sentence there would stop
an operator from reading a latency change as a hang.

**The second `pickWithWait` call is identical to the first except for the missing `exclude`,**
which is the entire point and is easy to lose in a later edit. Naming it — a local
`retryAfterCooldown` or an inline comment on the call rather than only above the `if` — would make
the difference survive a refactor.

Neither is a blocker. I found no unbounded wait, no unbounded retry, and no cancellation gap.

agentHits:
Thanks for the thorough termination analysis — both suggestions are applied in `62e2e0916`:

1. The `waitForCooldownMs` reference entry now states that a single-target combo with a nonzero wait holds the request up to the ceiling and retries the same target instead of failing immediately.
2. The second selector is now a named `retryAfterCooldown` local (same selector minus `exclude`), with the comment moved onto the helper so the deliberate difference survives refactors.

Verified on the new head: `bun run typecheck` clean, `server-combo-failover-e2e` 174/174 green.

abhisheksharma2411 (commented):
Re-checked on `873ac121` — both applied, and the second guard closes the one thing I couldn't verify.

```
PROBE hits=2 status=503 ms=56        (175 pass / 0 fail)
```

`comboTargetsDispatched <= 1` is the answer to the open question in my review. I'd flagged that `comboSendScope` is `undefined` when `options.sendBudget` isn't a `RequestExecutionBudget`, which would make `reserveDispatch` a no-op and leave the retry with no bound — and said I couldn't construct that path through the e2e harness. A local counter that doesn't depend on the budget existing at all removes the question rather than answering it, which is better; the comment naming both guards and saying explicitly "even if sendBudget scope is absent" means the next reader doesn't have to rediscover why there are two.

Naming `retryAfterCooldown` and putting the comment on the helper also does what I was actually after: the missing `exclude` now reads as a decision rather than an omission, so a future refactor that "tidies up" the duplication has to confront it.

Agreed on keeping it to `targets.length === 1`. A multi-target combo has real alternates and sequential cooldown waits across several exhausted targets would stack latency inside one request — the narrow scope is the feature, not a limitation.

Nothing further from me.

PR #4835, opened 2026-09-16
test: fix the causes behind two disabled test families and re-enable them

lidge-jun:
## Summary

Two families of tests were disabled to make red go away. Both causes are now fixed and every case runs again.

**Storage worker teardown was quarantined off Linux and macOS.** `tests/storage/storage-worker-teardown-isolate.test.ts` skipped its four Worker-spawning cases everywhere except win32. The reason given was real: Bun 1.3.14 segfaulted at `0xFFFFFFFFFFFFFFF8` mid-file with a *balanced* `workers_spawned === workers_terminated` count — exit 133 on macOS Silicon (run 30691129351) and exit 132 on ubuntu GHA (run 30700011812) — which no JavaScript teardown of ours can prevent. Bun 1.4.0, the version this repository pins, contains the upstream fix: worker threads are parent-owned and joined before the parent VM is destroyed, `bun:sqlite` and other native resources are torn down before JSC, and a termination gate keeps native callbacks out of a stopping worker (oven-sh/bun#37075, oven-sh/bun#38299). oven-sh/bun#38519 reproduces this exact class and records 3/3 crashes on 1.3.14 against 3 x 400 clean terminate cycles on 1.4.0. The skip is deleted rather than re-scoped, and the churn count is a single `8` on every platform: the one-cycle macOS and two-cycle Linux caps were crash avoidance, and a one-cycle "repeated spawn/reset" case does not test what its name claims. The meta-test that pinned those per-platform caps goes with them. The OS-join settle in `src/storage/worker-lifecycle.ts` is unchanged, because Bun's `close` event is still not a thread-exit proof.

**A composed acceptance case declared itself platform-independent and skipped on Windows.** Run 32344670867 shows why it was skipped: `CLI watchdog: ocx restore --json` at 45197 ms, on a shard where neighbouring cases took 54-106 s. There is no envelope, no SQLite error and no failed assertion in that log — the child was still waiting out production's own busy budget (5 s per attempt, two attempts, 500 ms apart) inside a real CLI process. That wait is not the assertion, so it is shortened instead of budgeted for. In-process history tests already shrink it with `setHistoryDbBusyTimeoutForTests`; a child process could not be reached that way, and neither could the history Worker, which is a separate module realm that starts from the provider's default. The run message now carries the parent realm's busy timeout exactly as it already carries the homes — validated, and refused when malformed — and the Worker adopts it before its first `state_5.sqlite` open. Production sends the same codex-rs-matching 5 s the Worker would have used on its own, so the happy path is unchanged. The test gives that one child a `--preload` that applies the knob only when `OCX_TEST_HISTORY_BUSY_TIMEOUT_MS` is present on its environment; no production module reads that variable. The lock, the two-attempt retry, the exit code, the exact JSON envelope, the byte-identity check, the release and the convergence assertion are all untouched.

**Two fail-open environment skips became hard preconditions in CI.** `model-metadata-sync` skipped its entire drift gate when `scripts/model-metadata.source.json` was absent, so the only check that the committed `src/generated/model-metadata.ts` still matches its source could vanish silently. That snapshot is tracked here, so its absence is a broken checkout and is now asserted. `server-startup-reconcile-resilience` probed `Bun.serve` and skipped four startup cases whenever the probe failed. That is correct in a sandboxed agent environment that denies `Bun.serve` outright and wrong in hosted CI, where a runner that cannot bind loopback is a broken runner and four assertions disappeared without a trace. The probe now suppresses cases only outside CI, and a CI-only guard case asserts the bind capability so an unbindable runner names itself instead of quietly shrinking the suite.

No skip was converted into a widened timeout, a retry, or a weaker assertion, and no budget was raised anywhere in this PR.

## Verification

**No local suite, focused test, typecheck, build, or install was run for this change.** Hosted GitHub Actions is the only verification instrument used here, at head `6d417ce6ead7428a14bac32b2aae5ef5ad3c5cfc`.

**Run 35141447585** (`pull_request`, head `6d417ce6ea`) — conclusion **success**. `test 1/4`, `test 2/4`, `test 3/4`, `test 4/4`, `macos 1/2`, `macos 2/2`, `gates`, `api usage`, `storage policy`, `docker smoke`, all three `keyring` jobs and all three `npm-global` jobs passed. The Windows test leg is `workflow_dispatch`-only, so it is skipped on this event.

**Run 35141987375** (`workflow_dispatch`, `lane=all`, same head `6d417ce6ea`) — all six Windows shards **success**, plus the Linux and macOS legs and `gates`.

The point of the change is that these cases now execute, so here is the direct log evidence rather than an aggregate conclusion:

- The four formerly quarantined worker cases, on the platforms that had them disabled — Linux `test 4/4`: `drain joins a fire-and-forget terminate before the isolate boundary` 268.58 ms, `repeated spawn/reset cycles leave no live workers` 2606.23 ms, `async beforeEach-style join between cycles leaves no live workers` 1688.74 ms, `terminateStorageWorker is joinable and idempotent across callers` 263.92 ms. macOS `macos 2/2`: 331.99 ms, 2291.47 ms, 1750.76 ms, 277.83 ms. No exit 132/133, no balanced-count panic, at eight churn cycles on both.
- The restore-busy case on Windows, which is the one that was skipped there: `windows 6/6`, `(pass) WP13 composed toggle acceptance > Restore truth: JSON distinguishes a busy history restore from native artifact recovery [11121.48ms]` against the 45 s Windows watchdog. The same case runs in 2524.41 ms on Linux and 3744.88 ms on macOS, where it previously spent about 10.5 s waiting.
- The two former fail-open skips, now executing as preconditions: `(pass) generated model metadata stays in sync with its source > regenerating reproduces the committed file byte for byte`, and `(pass) hosted CI can bind a loopback listener for the startup cases` followed by the four startup cases, on both Linux and Windows.

One unrelated failure appeared and did not reproduce. Windows shard 5/6 attempt 1 failed a case this PR does not touch: `tests/codex-integration/codex-plugins-doctor.test.ts:370`, `status --json includes a codexPlugins block and never writes CODEX_HOME`, which timed out at exactly its own hand-rolled 20 000 ms budget while its `spawnSync` child was still starting (`Received: null`). That same case runs in 2052.10 ms on the same shard on `dev` (run 35118018849) and 1404.71 ms on attempt 2 here, and the shard's totals match `dev` (4105 tests in 1569 s versus 3875 in 1531 s), so nothing in this patch slowed that leg — the only delta this PR makes to that shard is the removal of one instantaneous meta-test. It is the documented Windows child-start contention class: that test's spawn is intrinsic to its assertion but it uses a hand-rolled 20 s instead of `SPAWN_BUDGET_MS` (90 s on Windows), which is a separate lane's fix, not a change to sneak in here.

Static review covered the call chain from the test preload through `runCodexHistoryJob` to the Worker's first `openStateDb`; the argv composition for the added `--preload` against `withOwnedServiceHomePreload`, so each preload stays a separate argv element and a checkout path with spaces is still safe on Windows; the child's environment whitelist, which is why the new variable has to be passed explicitly; the file-size ratchet (no baseline caps cover the touched files, and `src/codex/history-provider.ts` stays under the 2000-line threshold); `structure/manifest.json`, which binds no invariant to the deleted per-platform cap meta-test; and `scripts/test-layout/layout.json` plus `tests/fixtures/test-layout-expected.json`, which need no entry because no test file was added or moved.

The dispatch run's overall conclusion reads `cancelled` for one reason unrelated to any assertion: `macos control`, the unsharded serial lane, hit its 30-minute wall while still executing tests. It recorded 19 453 passing cases and **zero failures** before being cut off, including all four re-enabled worker cases, the restore-busy case at 5203.83 ms, the metadata drift gate and the new bind guard. The same lane was cancelled the same way on `dev`'s most recent dispatch (run 35118018849, at the Bun 1.4.0 pin commit `2b19983bfd`), and it completed in 18-19 minutes on the three dispatches before that, so this is the truncation class `ci.yml` already documents for the sharded legs rather than anything this PR introduces. This change adds roughly three seconds to that lane: eight churn cycles instead of one on macOS, against a case that already ran there.

## Checklist

- [x] Docs or release notes were updated when needed. No `structure/` doc claims the busy-timeout realm boundary or the storage churn caps, and no user-facing behaviour changed, so `docs-site/` needs nothing.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. The new `busyTimeoutMs` field is validated as a finite non-negative number before adoption, the test-only environment variable is read by a test helper rather than any production module, and no default, authorization surface, or credential path changed.


lidge-jun:
## 리뷰 · 우선순위 66 / 80

이 PR은 지금 `dev`(HEAD `e2304ce0e`, tip #4825 subprocess kill-grace) 위에서, 예전에 빨간불을 가리려고 꺼 두었던 테스트 두 가족을 다시 켜는 작업이다. 제목의 “두 가족”은 (1) `tests/storage/storage-worker-teardown-isolate.test.ts`의 Worker 스폰/종료 격리 테스트들, (2) `tests/codex-integration/codex-composed-acceptance.test.ts`의 Windows에서 꺼져 있던 restore-busy JSON 계약 테스트다. 여기에 더해, 환경이 안 맞으면 그냥 조용히 건너뛰던 CI 전제 두 곳(`model-metadata-sync`, `server-startup-reconcile-resilience`)을 “없으면 실패”로 단단히 만드는 커밋도 같이 들어 있다. 사용자 기능 추가는 없고, 검증이 실제로 돌아가는지를 되살리는 쪽이다.

첫 번째 가족은 스토리지 Bun Worker 격리 teardown이다. 예전에 Bun 1.3.14가 Linux/macOS에서 `workers_spawned === workers_terminated`인데도 중간 segfault(exit 132/133)를 내서, Worker를 스폰하는 네 케이스를 win32만 남기고 건너뛰고 있었다. 이 저장소는 이미 `package.json`에 Bun `1.4.0`을 핀해 두었고(#4821), 업스트림이 worker 수명·native teardown을 고쳤다는 전제 아래 스킵을 지운다. 플랫폼별 churn 캡(darwin 1 / linux 2 / win32 8)과 그걸 고정하던 메타 테스트도 함께 없애고, 모든 OS에서 `WORKER_CHURN_CYCLES = 8` 하나로 맞춘다. `src/storage/worker-lifecycle.ts`의 OS-join settle은 손대지 않았다 — Bun `close`가 스레드 종료 증거가 아니라는 기존 판단은 유지한다.

두 번째 가족은 composed acceptance의 restore-busy 계약이다. 원래 “플랫폼 무관 계약”이라고 적어 놓고도 win32에서 `test.skipIf`로 꺼 두었는데, 실제 원인은 계약 실패가 아니라 production busy 예산(시도당 5초 × 2 + 재시도 간격 ≈ 10.5초)을 진짜 CLI 자식 안에서 다 기다리느라 45초 watchdog에 걸린 것이었다(run 32344670867). 이 PR은 타임아웃을 늘리거나 단언을 약하게 만들지 않는다. 대신 in-process 테스트가 쓰던 `setHistoryDbBusyTimeoutForTests`와 같은 짧아진 busy timeout을, 자식 프로세스와 history Worker Realm 경계 너머로 넘겨 준다.

그 통로가 production 쪽 작은 변경이다. `history-job.ts`의 run 메시지에 `busyTimeoutMs: currentHistoryDbBusyTimeoutMs()`를 실어 보내고, `history-worker.ts`는 메시지 검증 뒤 첫 `openStateDb` 전에 `adoptHistoryDbBusyTimeout`으로 받아들인다. 테스트는 새 헬퍼 `tests/helpers/history-busy-timeout-preload.ts`를 `--preload`로만 붙이고, 환경변수 `OCX_TEST_HISTORY_BUSY_TIMEOUT_MS=250`이 있을 때만 자식을 짧게 만든다. production 모듈은 그 변수를 읽지 않는다. 행복한 경로는 양쪽 모두 기본 5초라서 행동이 같다. `codex-history-worker.test.ts`에는 메시지 검증·malformed 거절·adopt 거절 테스트가 추가됐다.

세 번째 묶음은 fail-open 스킵을 CI 전제로 바꾸는 일이다. `model-metadata-sync`는 `scripts/model-metadata.source.json`이 없으면 통째로 스킵하던 것을, 파일이 저장소 소유이므로 없으면 `expect`로 깨지게 바꿨다. `server-startup-reconcile-resilience`는 `Bun.serve` 루프백 바인드 실패 시 네 케이스를 건너뛰던 것을, CI 밖(샌드박스)에서만 스킵하고 CI에서는 돌리며, CI 전용 “루프백 바인드 가능” 가드 케이스도 추가했다. PR 본문 주장대로 스킵을 넓은 timeout/재시도/약한 단언으로 바꾸지 않았고, 예산도 올리지 않았다. 방향은 현재 `dev`의 테스트·런타임 신뢰성 레인(#4821 Bun pin, #4825 subprocess grace)과 잘 맞는다. 다만 호스트 CI가 아직 pending이라, 실제로 Linux/macOS isolate churn 8과 Windows restore-busy가 초록인지가 이 PR의 진짜 합격선이다.

라인 68 (`tests/codex-integration/codex-composed-acceptance.test.ts`) - 상수 `HISTORY_BUSY_TIMEOUT_ENV`만 쓰려고 사이드이펙트가 있는 preload 파일을 import한다. 부모 테스트 프로세스에 `OCX_TEST_HISTORY_BUSY_TIMEOUT_MS`가 실수로 잡혀 있으면, 이 파일이 로드되는 순간 in-process history busy timeout까지 짧아질 수 있다. 상수만 담은 작은 모듈로 나누거나, 문자열 리터럴을 쓰는 편이 더 안전하다.
라인 90-92 (`src/codex/history-provider.ts` `adoptHistoryDbBusyTimeout`) - 잘못된 값은 조용히 return한다. 메시지 게이트(`isHistoryWorkerRunMessage`)가 막아 주긴 하지만, 나중에 검증을 우회하는 호출이 생기면 기본 5초가 남는지 실패인지 로그로 안 보인다. 테스트 전용 setter와 역할이 갈리니, 주석에 “게이트 통과 전제”를 한 줄 더 박아 두면 좋다.
경로 `tests/storage/storage-worker-teardown-isolate.test.ts` / `WORKER_CHURN_CYCLES = 8` - Bun 1.4.0 핀을 전제로 Linux/macOS에서 예전에 segfault 나던 부하를 다시 켠다. 핀이 맞으면 맞고, CI가 빨개지면 원인 재확인이 필요하다. 스킵을 되살리지 말고, 런타임/정리를 다시 보는 쪽이 PR 취지에 맞다.
경로 CI checks (작성 시점 pending) - 본문이 “hosted CI evidence will be added”라고 했고, 지금 gates/test shards가 아직 안 끝났다. 이 PR의 핵심 주장이 CI 초록이므로, 머지 판단은 체크 완료 후로 두는 것이 맞다.

메인테이너의 판단이 필요한 지점
- preload 상수 import의 사이드이펙트 분리(상수 전용 파일 vs 현재 구조)를 이 PR에서 고칠지, 후속으로 둘지
- Linux/macOS isolate churn 8이 CI에서 한 번이라도 흔들리면, 캡을 다시 줄일지(비추천) 아니면 Bun/teardown 쪽을 다시 볼지
- production run 메시지에 `busyTimeoutMs`를 항상 실어 보내는 작은 표면 확장을 테스트 전용 seam으로만 남길지, 지금처럼 “Realm 경계에 homes와 같이 명시”로 둘지(현재 설계가 더 일관됨)

너의 추천
호스트 CI(특히 storage isolate 관련 shard + Windows composed restore)가 초록인지 확인한 뒤 머지하는 것을 추천한다. 상수 import 사이드이펙트는 작은 follow-up으로 분리해도 되고, 지금 구조로 가도 부모 env에 테스트 전용 변수가 없는 한 실해는 낮다. types/config 분할이나 중복 PR로 무효화될 내용은 없다. CI 빨강이면 스킵 부활이 아니라 실패 로그부터 본다.

이 댓글은 grok-bot이 작성했습니다

PR #4834, opened 2026-09-16
test(budget): restore the 45s spawn signal and stop hardening a disposable port

lidge-jun:
## Summary

PR #4830 raised `SPAWN_BUDGET_MS` on win32 from 45s to 90s to fix one slow test case. The blast radius was much wider than the fix: 31 test files read that constant and nine hand it straight to `setDefaultTimeout`, so the edit reached 339 Windows cases — 315 went 45s → 90s, 22 that multiply it went 90s → 180s, and one derivation chain in `codex-sync-api.test.ts` reached 265s. Several of the affected files never spawn a process at all; the detectors that now report twice as late are exactly the contention ones (`codex-write-lock`, the cross-process history-lock exclusions, the sync-cache `models_cache` case, the Windows shim and main-account startup subprocess cases).

The 265s case matters beyond slow reporting. The Windows shards have a 30-minute job timeout (`.github/workflows/ci.yml` L772) and run `bun test --isolate --timeout 60000 tests --shard=N/6` (L846), so an explicit per-test timeout overrides the 60s default and the job timeout becomes the only outer bound. Run `35129771091` measured Windows 5/6 at 26m39s and 6/6 at 25m13s. One hang in that 265s case puts its shard within roughly half a minute of an opaque job cancellation instead of a readable Bun timeout.

The measurement that justified #4830 was also contaminated by the test's own instrumentation. The child published its port with `atomicWriteFile`, the production **secret** writer, which on Windows runs `hardenSecretPath(..., required: true)` twice (`src/config/atomic-write.ts`), each able to spawn PowerShell for SID resolution and several `icacls` passes budgeted at 30s apiece. That ACL ceremony ran inside the window the parent measures as "time to reach a port" — performed on a disposable, non-secret port number.

What this changes:

- `SPAWN_BUDGET_MS` returns to `45_000` on every platform and every call site.
- A new `COLD_SPAWN_BUDGET_MS` (90s on win32, 45s elsewhere) is consumed **exactly once**: the readiness wait of the first proxy child spawned in `native-profile-startup.test.ts`. Every later child in that file and every other importer is back to 45s.
- The child publishes its port and settled markers through a plain temp-file `renameSync` in the same directory. That preserves the only contract the parent has (#1061: never observe the file between create and write) and removes the PowerShell/`icacls` ceremony from the measured window.
[...]

PR #4836, opened 2026-09-16
perf(windows): stop hardening every secret temp twice

lidge-jun:
## Summary

Every ACL-hardened write on Windows ran the icacls mutation sequence twice against the same file, and this is a production path: a cold `ocx start` publishes `ocx.pid` and `runtime-port.json` through `atomicWriteFile` after the listener binds.

`atomicWriteFile` hardens the temp while it is still empty — so it is never readable by another principal between creation and the first byte — and hardens it again before the rename. Both calls are `required: true`. The content written in between moves the success memo's freshness component in `src/lib/windows-secret-acl.ts`, which is `ctimeNs`, and libuv reports `st_ctim` on Windows from the NTFS ChangeTime. So the second call missed the memo and reapplied the ACL the file already had: `/grant:r`, `/inheritance:r`, `/remove:g`, plus up to three `/findsid` probes when `/remove:g` reports non-zero.

After the content write the writer now re-asserts that the path still resolves to the object its descriptor holds, then re-attributes the memo to that same object through the new `reattributeHardenedSecretPath`. The pre-rename harden resolves through the memo instead of repeating the mutation. Windows also stops taking the `chmod` on that path: there it sets the read-only attribute rather than the DACL, so it is not the secret boundary, and its ChangeTime bump is what retired the memo a line later. POSIX keeps the `chmod` and is unchanged.

**Concrete cost.** Per hardened write on Windows, icacls invocations go from 6 to 3 (from 12 to 6 when `/remove:g` reports non-zero and the three `/findsid` probes run). The two writes a cold `ocx start` performs after binding go from 4 mutation sequences to 2. PowerShell spawns are unchanged at one per process.

### Which of the four candidate directions this takes, and why not the others

**Taken: direction 4, collapsing the double harden** — but by making the second call a memo hit, not by deleting it. The call stays, stays `required: true`, and performs the full sequence whenever attribution cannot be proven. The pre-write harden is kept: the window it closes is real.

**Direction 1, memoize the account SID per process, is already implemented.** `cachedIdentity` in `src/lib/windows-user-principal.ts` is process-wide and is consulted before `timeoutMs` on both the sync and async paths, and the async path additionally single-flights. A successful lookup costs one PowerShell spawn per process, not one per harden. No change was needed.

**Direction 2, a native token lookup instead of PowerShell, is not shipped.** It is reachable — the repository already uses `bun:ffi` for `GetSystemDirectoryW` — but it cannot be validated on a non-Windows workstation, and its failure mode is not a fallback. `runIcacls` grants the returned principal Full Control and then removes inheritance, so a wrong SID either hands the secret to another account or strands the file with no usable ACE, which is exactly the hazard the no-name-fallback rule in `windows-user-principal.ts` exists for. It saves one spawn per process; that is not worth shipping blind.

**Direction 3, skipping a path whose ACL is already compliant, already exists and is deliberately off.** `existingAclAlreadyCompliant` landed in `dd3ff4231b` (#2963) behind `OPENCODEX_ACL_VERIFY_EXISTING=1`, and `devlog/_fin/260830_release_readiness_train/050_wp5_issue_1298_acl_proof.md` records why the default is off: a parser false-positive could expose credentials. Flipping that default is a maintainer decision about parser trust, not a cost fix, so this PR does not touch it. Operators who have it on already get the double harden collapsed to one read.

### Security accounting

**Worst case if the comparison is wrong.** The memo is refused, not wrongly satisfied, in every direction the code can be wrong about: no memo, an unreadable stat, or a different `dev:ino` all retire the entry and the pre-rename harden runs the full sequence. A refusal costs exactly the work this PR removes. Nothing relaxes `required: true`; a hardening failure still throws and the write still fails closed, with the temp scrubbed and unlinked.

**Why a secret cannot become readable by another principal.** The ACL is applied to the empty temp before any content exists, and it is applied to the same object that is renamed into place — that is what the second descriptor/path identity assertion proves, and it fails the write closed if it cannot. Writing data does not alter a DACL.

The honest cost is stated in the code: `PathObservation` documents that freshness also moves when permissions change, and re-attribution cannot tell a data write from a DACL change, so a DACL change landing inside the caller's own content write is absorbed rather than forcing a re-harden. Two things bound that. First, detection is relocated, not removed — every permission change *after* the re-attribution still moves ctime, still misses the memo, and is still hardened in full; there is a regression test for exactly that. Second, consider who can be inside the absorbed window at all. Once the first harden has run, the DACL is an explicit owner-only ACE with inheritance removed, so no other principal can open the file for `WRITE_DAC`. The one principal who could still rewrite it is one holding a handle opened *before* the harden, and Windows preserves the access granted to an open handle — that principal can rewrite the DACL after any later harden and after the rename, on the same object. A second mutation pass never bounded that capability; it stripped an ACE the holder could immediately re-add. Declining to repeat it therefore removes no guarantee anyone had.

### Files changed

- `src/lib/windows-secret-acl.ts` — memo stores the `PathObservation` instead of a joined `dev:ino:ctimeNs` string, so object and freshness stay separately askable; new `reattributeHardenedSecretPath`.
- `src/config/atomic-write.ts` — post-write identity assertion and re-attribution in both temp writers; no `chmod` on the Windows branch of both default harden closures; a module-local `setWindowsHardeningForTests` seam whose untouched default is the previous `process.platform === "win32"` gate.
- `tests/windows/windows-acl-start-cost.test.ts` — new, with `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json` entries.
- `structure/runtime.md`, `structure/config.md` — the owning docs for `src/config/atomic-write.ts`.

The new tests live in their own file rather than in `tests/windows/windows-secret-acl.test.ts` because that file is on the file-size ratchet baseline at 2310 lines and growing it fails `file-size-ratchet`.

## Verification

**No local test suite, focused test, typecheck, build, or install was run for this change.** Verification is static reasoning plus hosted GitHub Actions CI. All evidence below is at head SHA `f4b10c4aa5496edb08e44f935481f4177c66e8b4`.

**Pull-request aggregate — run [35142055301](https://github.com/lidge-jun/opencodex/actions/runs/35142055301) (attempt 2): `ci` = success.** Every job green: changes, select windows runner, test 1/4–4/4, storage policy, api usage, gates, macos 1/2, macos 2/2, keyring macos/ubuntu/windows, docker smoke, npm-global macos/ubuntu/windows. `macos control` and `platform-windows` are skipped by design — `platform-windows` is gated on `workflow_dispatch`, so a pull request never runs Windows. Attempt 1 was cancelled when `macos 2/2` stalled inside `tests/codex-integration/`; attempt 2 ran the same commit and passed.

**Windows — dispatched twice at the same SHA because the leg is gated off pull requests.**

- [35148850553](https://github.com/lidge-jun/opencodex/actions/runs/35148850553): windows 1/6, 2/6, 4/6, 5/6, 6/6 success; **windows 3/6 failure** — two cases in `tests/clients/client-connect.test.ts` (`rotate` and `commit-lost` in "Desktop copy coherence across client lifecycle") hit the 15 s child-process deadline in `runDesktopLifecycleScenario`.
- [35152226272](https://github.com/lidge-jun/opencodex/actions/runs/35152226272): windows 1/6, **3/6**, 4/6, 5/6, 6/6 success; **windows 2/6 failure** — a different test in a different file (`tests/server/server-management-auth.test.ts`, "the data-plane token does not authorize the restart route") failed on `EPERM: operation not permitted, rm` while removing its temp tree in `removeTreeWithRetry`.

The blame moved between runs while the change did not, which is the signature `tests/helpers/test-budget.ts` already describes for this leg ("a run of windows-latest failures kept blaming a different test each time. The blame moved; the cause did not"). Neither failing test touches this change: one is bounded by `INTERNAL_DEADLINE_MS` on a spawned `bun test` child, the other is the documented Windows temp-removal `EPERM`.

**This change's own tests passed on Windows in both dispatches** (`tests/windows/windows-acl-start-cost.test.ts`, shard 1/6, jobs 104971744013 and 104983103630), each case in 1–2 ms.

**Contention forensics for the windows 3/6 failure.** Comparing every test present in both that shard and an all-green baseline (`1504caaa83`, run 35134620067) at >200 ms, the ratio distribution is median 0.90, p75 1.36, p90 2.36, max 10.03. The slowest ratios are unrelated to ACL hardening — `ocx init piped stdin` 10.03×, Remote Workspace hub WebSocket 9.46×, Codex Log Guard reclaim 5.28×, Codex autostart shim 5.98×. The two failing cases sit at 4.68× and 5.01×, inside that tail rather than above it, and the shard's wall time was 1398 s against the baseline's 1130 s. On the two all-green baselines the same four cases ran 3171/3265 ms, 3048/2836 ms, 3278/2891 ms and 6662/3059 ms against the same 15 s deadline — a 2.3–5× margin, not a marginal one.

Static checks performed instead of a local run:

- Traced every new test through the real production code by hand for call counts, including the `OPENCODEX_ACL_VERIFY_EXISTING` gate, the `/findsid` branch that only runs when `/remove:g` fails, `assertPrivateTempDescriptor`'s POSIX `0o600` check, and `recordOwnedConfigPath` ignoring destinations outside the config root.
- Confirmed the source oracle in `tests/windows/windows-secret-acl.test.ts` still finds exactly two `openSync(path, "wx", 0o600)` call sites in `atomic-write.ts`.
- Confirmed no production caller sets the new test seam and that `src/config.ts` does not re-export it.

Ablation, reasoned statically for each behavioural test: reverting the re-attribution makes the two `atomicWriteFile` cases observe six icacls verbs instead of three; reverting the freshness half of the memo comparison makes "freshness moving again after re-attribution still forces a full harden" observe three instead of six. Neither case waits on a clock. Freshness is modelled through the `setStatForTests` seam precisely so these assert this module's attribution rule rather than a volume's timestamp resolution, which on a coarse-resolution filesystem could let two adjacent operations share a tick.

## Unresolved risk — two Windows CI flake classes observed at this head

Both were observed at `f4b10c4aa5` and neither is attributed to this change; the contention forensics are in Verification above. Recorded here so the next lane starts from evidence rather than from zero. The host is separately dispatching `lane=all` on `dev` (`5d98281eea`), which does not contain this change, to confirm both classes reproduce without it.

### Class A — `client-connect` "Desktop copy coherence" child-probe timeout

**Where.** Run [35148850553](https://github.com/lidge-jun/opencodex/actions/runs/35148850553), job `104971744046` (`windows 3/6`).

```
2026-09-16T20:55:12.7872812Z ClientStateProbeError: Client state probe timed out (status=null, signal=SIGKILL, category=child_failed)
      at runDesktopLifecycleScenario (D:\a\opencodex\opencodex\tests\clients\client-connect.test.ts:1196:160)
2026-09-16T20:55:12.7906195Z (fail) Desktop copy coherence across client lifecycle > rotate settles Desktop before reporting committed [15285.63ms]
2026-09-16T20:55:54.2856989Z (fail) Desktop copy coherence across client lifecycle > commit-lost settles Desktop before reporting committed [15329.32ms]
```

Passing siblings in the same describe on that run: `recover-both` 14838.82 ms, `recover-current` 11328.70 ms, `normal failed candidate` 9689.79 ms, `recover-backup` 8500.21 ms, `same-old` 7052.39 ms.

**Failing operation.** `tests/clients/client-connect.test.ts:1187` — `spawnSync(process.execPath, ["test", "--preload", …, fixturePath], { timeout: INTERNAL_DEADLINE_MS, killSignal: "SIGKILL" })`. The child was killed at 15 s, so it wrote no `{"fixtureResult":…}` marker, so the category fell through to `child_failed` at `:1195`. The signal is the diagnosis being lost, not the cause.

**Hypothesis — a misclassified budget, plus two deadlines racing.** `INTERNAL_DEADLINE_MS` is 15 000 ms and `tests/helpers/test-budget.ts` defines it as "a deadline *inside* a test, for an await that would otherwise hang forever", adding "keep these at least a few times under the surrounding budget". It is used here for a real child `bun test` process. The constant for that category is `SPAWN_BUDGET_MS` — 90 s on Windows, and its comment says why: "the leg runs four Bun pools on one runner, and the first child spawned in a file pays a cold start the later ones do not". The same file already uses `SPAWN_BUDGET_MS` for its other spawn-based cases at `:549` and `:584`; this family is the inconsistency. Worse, the identical 15 000 ms also bounds the generated fixture's own `bun test` case at `:1180`, so inner and outer deadlines race and whichever fires first decides whether the failure is diagnosable.

**Why it is contention, not a regression.** On the all-green baselines the same four cases ran 3171.42 / 3048.52 / 3278.82 / 6662.30 ms (run 35139132889 @ `44de45dfdc`, job `104938976212`) and 3265.52 / 2836.47 / 2891.07 / 3059.59 ms (run 35134620067 @ `1504caaa83`, job `104923799425`) — a 2.3–5× margin against the same 15 s bound. On the failing shard, unrelated spawn-bound tests ran up to 10.03× slower while the median across 693 shared tests was 0.90.

**Concrete budget-category correction.** Order the three bounds strictly, so the innermost always reports first and every failure is categorized instead of a blind kill:

| bound | location | from | to |
|---|---|---|---|
| fixture's own `bun test` case | `client-connect.test.ts:1180` | `INTERNAL_DEADLINE_MS` | `SPAWN_BUDGET_MS` |
| parent `spawnSync` timeout | `client-connect.test.ts:1187` | `INTERNAL_DEADLINE_MS` | `SPAWN_BUDGET_MS + INTERNAL_DEADLINE_MS` |
| each `test.each` case budget | the describe at `:1202` | implicit (CI `--timeout 60000`) | `SPAWN_BUDGET_MS + 2 * INTERNAL_DEADLINE_MS` |

This invents no tuned number; it applies the existing spawn constant and spaces the layers by the existing internal deadline. It meets both halves of the standard in `test-budget.ts`: the spawned child **is** the scenario, and the assertions are on the fixture's returned state, so reverting "Desktop settles before commit" still fails regardless of budget. It was deliberately not applied in this PR — it is outside this lane's write scope and sets flake policy beyond it.

### Class B — `EPERM` temp-tree removal, already named in the source

**Where.** Run [35152226272](https://github.com/lidge-jun/opencodex/actions/runs/35152226272), job `104983103569` (`windows 2/6`).

```
2026-09-16T21:40:06.9370293Z [opencodex] ACL hardening timed out (ETIMEDOUT) — transient icacls stall; … (probe failed) — continuing without NTFS ACL harden
2026-09-16T21:40:07.4743094Z [opencodex] Config mutation coordination directory ACL hardening did not complete; continuing without it. ACL hardening timed out (ETIMEDOUT) …
2026-09-16T21:40:23.6961626Z EPERM: operation not permitted, rm 'C:\Users\RUNNER~1\AppData\Local\Temp\opencodex-test-vwyZZv\tmp\ocx-management-auth-kgvXCT'  (errno -4048)
      at removeTestTempTree (scripts\test-temp.ts:195:7)
      at removeTreeWithRetry (tests\helpers\remove-tree.ts:27:3)
      at <anonymous> (tests\server\server-management-auth.test.ts:234:17)
2026-09-16T21:40:23.7018576Z (fail) codex app-server restart routes ride the management gate > the data-plane token does not authorize the restart route [15367.90ms]
```

**Failing operation.** The `afterEach` at `server-management-auth.test.ts:234` (`if (testHome) removeTreeWithRetry(testHome)`). The 15 367 ms case duration is almost entirely `REMOVE_RETRY_BUDGET_MS` (15 000 ms, `scripts/test-temp.ts:33`) being spent on a 50 ms→250 ms doubling ladder before `removeTestTempTree` rethrows. The failing case name is incidental: it is simply the test whose teardown ran.

**Hypothesis — a live `icacls.exe` still holding the directory, and this class is not independent of Class A's runner load or of this PR's subject.** `src/config/paths.ts:31` starts a fire-and-forget optional flight, `hardenSecretDirAsync(dir, { required: false })`, tracked in `configDirHardeningFlights`. The doc on `flushConfigDirHardening` (`src/config/paths.ts:61`) states the consequence directly: "The flight spawns `icacls.exe`, which holds the directory open until it exits. Windows file locking is mandatory, so anything that removes or renames that directory after a 'clean' shutdown — a test fixture teardown, an uninstaller, a home move — gets EPERM/EBUSY unless the process that started the child also waits for it." Only `server.stop` drains it, at `src/server/index.ts:799`.

`src/lib/bounded-subprocess.ts:20-40` records the same failure with the same fixture prefix and three previous fixes: "#4789 blamed the removal retry budget…; #4796 gave it a 15s exponential schedule; a later change awaited the hardening flight from the test hook. Windows shard 1/6 failed identically through all three… run 35108652486 burned the full 15s budget and still threw `EPERM … rm ocx-management-auth-fDchUb`, with two `ACL hardening timed out (ETIMEDOUT) - transient icacls stall` lines logged beside it."

The occurrence above is the same signature — `ocx-management-auth-*`, full budget burned, icacls stall lines beside it — and it is the **fourth** attempt's aftermath: `e2304ce0e3` ("wait for a killed subprocess to die before abandoning it", #4825) is the tip of `dev` this branch is based on, so that fix is present and the class still recurs. The reason it can is stated in that same comment: "abandonment is still the fallback". `waitForSubprocessExit` kills at the deadline and abandons after `SUBPROCESS_KILL_GRACE_MS` (2 000 ms), and a child that has already blown a 30 s harden budget is precisely the one that will not die inside 2 s. An abandoned-but-live `icacls.exe` keeps the handle, and no drain can recover it.

Two consequences worth naming for whoever takes this: the remaining exposure is proportional to **how many icacls children are started**, which is what this PR reduces per write; and a fix that bounds the removal or the drain cannot close it, because the holder outlives both. The lever is either not starting the child, or not letting an abandoned child keep a handle on a tree someone is about to remove.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Windows secret-file protection during writes by applying ACL hardening consistently before and after content updates.
  - Prevented redundant hardening operations when the file remains associated with the same underlying object.
  - Added safeguards to detect file replacement or identity changes before completing a write.
  - Failed Windows hardening operations now stop the write without leaving a destination file.

- **Documentation**
  - Updated runtime and configuration documentation to describe Windows ACL handling and write verification.


lidge-jun:
## 리뷰 · 우선순위 73 / 80

지금 `dev` 끝은 #4825(e2304ce0e, `waitForSubprocessExit` 킬 그레이스) 이고, 패키지는 여전히 2.58.0 입니다. 이 PR(#4836, 헤드 f4b10c4aa)은 Windows에서 시크릿 temp를 `atomicWriteFile` 로 쓸 때마다 icacls 변이 시퀀스가 두 번 돌아가던 생산 경로 비용을 한 번으로 줄입니다. 손대는 핵심은 `src/lib/windows-secret-acl.ts` 와 `src/config/atomic-write.ts` 이고, 새 테스트 `tests/windows/windows-acl-start-cost.test.ts` 와 structure 문서(`structure/config.md`, `structure/runtime.md`)가 같이 갑니다. types.ts/config.ts 분할 캠페인과 무관해서 닫을 이유가 없고, 같은 주제의 열린 중복 PR도 안 보입니다(#4834 는 테스트 예산/픽스처 쪽이지 이 ACL 메모 재귀속이 아닙니다).

문제는 간단합니다. `atomicWriteFile` 은 temp가 비어 있을 때 한 번 굳히고(다른 주체가 첫 바이트 전에 읽지 못하게), rename 직전에 또 한 번 굳힙니다. 둘 다 `required: true` 입니다. 그런데 그 사이에 내용을 쓰면 Windows에서 libuv가 보고하는 `ctimeNs`(NTFS ChangeTime)가 움직여서, 성공 메모의 freshness가 어긋나고 두 번째 호출이 메모 히트가 아니라 전체 `/grant:r` + `/inheritance:r` + `/remove:g`(그리고 `/remove:g` 가 실패하면 `/findsid` 최대 세 번)를 다시 돌립니다. 이미 있는 ACL을 다시 바르는 셈입니다. 차가운 `ocx start` 가 리스너 bind 뒤 `ocx.pid` 와 `runtime-port.json` 을 이 writer로 쓰니, 시작 경로에서 변이 시퀀스가 4번 → 2번으로 줄어드는 체감이 있습니다.

선택한 방향은 '두 번째 harden 호출을 지우지 않고 메모 히트로 만들기'입니다. 내용 쓰기 뒤 디스크립터/경로 동일성을 `assertPrivateTempDescriptor` 로 다시 확인하고, 디스크립터를 닫은 뒤 `reattributeHardenedSecretPath` 로 같은 객체에 freshness만 다시 붙입니다. 객체가 바뀌었거나 stat이 안 되면 메모를 지워서 다음 harden이 풀 시퀀스로 돌아가게 합니다. 메모 값 타입도 문자열 `dev:ino:ctimeNs` 에서 `PathObservation` 자체로 바꿔 object/freshness를 따로 물을 수 있게 했습니다. Windows 분기에서는 `chmod` 도 빼는데, 거기서는 읽기 전용 속성만 건드리고 DACL이 아니며 ChangeTime을 올려 메모를 바로 깨뜨리기 때문입니다. POSIX의 mode 경계는 그대로입니다.

보안 회계도 본문에 길게 적혀 있고, 읽기 기준은 맞습니다. 비교가 틀리면 메모를 '잘못 만족'시키는 쪽이 아니라 거부하고 풀 harden으로 돌아가며, `required: true` 는 그대로라 실패하면 쓰기 자체가 닫힙니다. 흡수되는 창(내용 쓰기 중 DACL 변경을 freshness 이동과 구분 못 함)에 대해서도, harden harden 뒤에는 다른 주체가 `WRITE_DAC` 로 열 수 없고, harden 핸들을 가진 쪽은 두 번째 변이 패스 뒤에도 같은 객체를 다시 쓸 수 있다는 이유로 '반복이 주던 보장이 원래 없었다'고 설명합니다. 재귀속 뒤 freshness가 또 움직이면 풀 harden이 다시 돈다는 회귀 테스트도 있습니다. 네 후보 중 SID 프로세스 메모는 이미 있고, native 토큰 조회는 검증 불가·실패 모드가 fallback이 아니라 보류, `OPENCODEX_ACL_VERIFY_EXISTING` 기본 ON은 파서 신뢰 문제라 이번 비용 수정 범위 밖으로 둔 판단도 일관됩니다.

테스트는 `windows-secret-acl.test.ts` 파일 크기 래칫(2310줄) 때문에 새 파일로 분리했고, layout/expected에도 넣었습니다. `setWindowsHardeningForTests` 는 `windowsSecretAclApplies()` 를 재사용하지 않는 로컬 시임이라, 다른 스위트가 `setPlatformForTests("win32")` 만 켜도 설정 쓰기에 icacls가 붙지 않게 했습니다. `src/config.ts` 재수출에도 안 넣었습니다. 본문이 로컬 테스트/타입체크/빌드를 안 돌렸다고 명시했고, 호스티드 CI는 이 웨이크 시점에도 여러 잡이 pending입니다. 정적 추적·ablation 논리는 꼼꼼하지만, 머지 버튼은 초록을 보고 누르는 게 맞습니다. 직접 lidge-jun perf 랜드이고 머지 트레인 잔여 원본 정리도 필요 없습니다.

라인 atomic-write.ts · carryHardenAcrossContentWrite - 재귀속을 디스크립터 `closeSync` 뒤(핸들 밖)에서 돌립니다. 주석대로 Windows가 열린 핸들 동안 새 ChangeTime을 path 조회에 안 올릴 수 있어서, 닫기 전에 메모를 갱신하면 다음 독자가 보는 시각과 어긋나 메모가 다시 빗나갈 수 있습니다. 의도는 맞지만, 이 순서가 깨지면(예: close 전에 재귀속) 비용 절감이 조용히 사라지므로 순서 계약을 테스트/주석으로 계속 지켜야 합니다.
라인 windows-secret-acl.ts · reattributeHardenedSecretPath - 파일 메모(`hardenedPaths`)만 다루고 디렉터리 메모(`hardenedDirectories`)는 건드리지 않습니다. 이번 호출자는 temp 파일이라 맞고, 나중에 디렉터리 쪽에도 같은 패턴을 붙일 때 대칭 API가 없다는 점만 알아두면 됩니다.
경로 atomic-write.ts · setWindowsHardeningForTests - 모듈 전역 오버라이드입니다. 새 테스트 `afterEach` 에서 `null` 로 되돌리므로 이 PR 안에서는 정리되어 있습니다. 다른 스위트가 이 시임을 쓰기 시작하면 `afterEach` 누락이 교차 오염을 만드니, 시임이 leaf에만 있고 facade 재수출이 없는 현재 계약은 유지하는 편이 좋습니다.
경로 CI / Verification - 본문이 로컬 실행 없음을 솔직히 적었고, 이 웨이크에서 `gh pr checks` 는 changes/hygiene/windows runner 선택 등이 pending, resolve-pr만 pass였습니다. 실패-닫힘(`EICACLS`)·한 시퀀스 고정·객체 교체 시 메모 폐기 케이스가 있으니, Windows/테스트 샤드 초록이 확인되면 머지 근거가 됩니다.

메인테이너의 판단이 필요한 지점
- 내용 쓰기 창에 흡수되는 DACL 변경을 '반복 harden이 원래 막지 못하던 능력'으로 두고 이번 라운드에서 받아들일지, 아니면 파서 기반 `OPENCODEX_ACL_VERIFY_EXISTING` 기본값 논의와 같이 별도 보안 트레인으로 남길지.
- 차가운 start에서 pid/port 두 번의 절감만으로 충분한지, 아니면 같은 writer를 쓰는 다른 핫 경로(인증/설정 저장)까지 벤치 숫자를 남긴 뒤 머지할지.
- #4834(픽스처가 `atomicWriteFile` ACL을 피하도록 rename으로 바꾼 테스트 예산 PR)와 머지 순서를 나란히 둘지. 서로 충돌하지는 않지만, Windows 시작/스폰 측정 창에서 ACL 비용이 동시에 빠지면 관측이 겹칩니다.

너의 추천
호스티드 CI(특히 Windows·테스트 샤드·hygiene/gates)가 초록이면 `dev` 에 머지하세요. 두 번째 harden 호출을 지우지 않고 메모 히트로 바꿔 비용을 줄이면서 `required: true` 실패-닫힘을 유지한, 범위가 선명한 생산 경로 perf입니다. types/config 분할과 무관하고 중복도 아니며, 추가 커밋 없이 머지 트레인이면 됩니다. 머지 후 남은 원본 PR 정리도 필요 없습니다(직접 lidge-jun 랜드).

이 댓글은 grok-bot이 작성했습니다

PR #4837, opened 2026-09-16
ci: fail on the first timeout or Bun crash instead of retrying to green

lidge-jun:
## Summary

CI converted real failures into green in four places. This removes all of them and makes the aggregate gate say what it means.

`scripts/ci/run-bun-test-batches.sh` classified a batch process timeout separately from a crash, re-ran every file of the failed batch one per process, retried each singleton once more, and returned SUCCESS when that sweep passed. Singleton isolation removes precisely the conditions that produce this class of failure — batch concurrency, shared process state, resource pressure — so the sweep was always going to pass and always going to report nothing. It hid batch-only deadlocks, order dependence, state leakage and contention. Run `35087572377`, job `104766021341` logged `panic(main thread): Segmentation fault at address 0x10` and `oh no: Bun has crashed.` in batches 11, 15, 18, 19, 22 and 26, and the run reported green.

The macOS shard, the macOS control and the Windows shard each carried `for attempt in 1 2` and reported success if attempt 2 passed. A Bun runtime crash is process death a user would have seen; a second execution that happens not to die does not un-kill the first.

A process timeout or a runtime crash now fails its job on its first occurrence. The one-file-per-process sweep survives, renamed `attribute_batch_file_by_file`, and runs only after the shard has already failed — so a human still learns whether any single file reproduces alone, and nothing it prints can change the disposition. The shared classifier from `scripts/ci/bun-crash-signatures.sh` still runs in every lane; its job is now to choose the error message, never the outcome.

The aggregate `ci` gate accepted `skipped` from any job unconditionally, so it could not distinguish "this event did not request the job" from "this event requested it and it never started". It now derives the requested set from the event and the `changes` outputs, mirroring each job's own `if:`, and requires `success` from every requested job and `skipped` from every job the event did not request. On a `workflow_dispatch` with `lane=all` it additionally reads the run's own job list and requires six concrete successful `windows N/6` results, because a matrix rollup reports `success` for five successes and one skipped leg — the shape of run `35112645195`, where Windows job `104850075282` reported `skipped` with zero steps while aggregate job `104855808104` concluded success.

`tests/ci-workflows/ci-bun-crash-classifier.test.ts` asserted shell SOURCE TEXT with `includes`/regex and never executed anything, and one of its cases — "a timeout may still recover" — pinned the mask in place. The disposition contract moves to a new test that runs the real scripts:

[...]

PR #4844, opened 2026-09-16
diagnostic: probe C, trivial file on PR #4837 base (not for merge)

lidge-jun:
## Summary

**Diagnostic probe. Not for merge; will be closed once it has answered its question.**

Same trivial one-assertion file as #4839, at the same sorted position, but based on `e2304ce0e3` — the base of PR #4837 — instead of current `dev`.

#4839 (trivial file on current dev) and #4840's companion probe (PR #4837's real test file on current dev) both left `macos 1/2` green at 663 files. PR #4837 itself wedges that shard twice: 149 files, then 17 minutes of silence, then the 20-minute wall. The remaining difference is the base. `dev` has since landed #4835, which fixed Worker-teardown causes and re-enabled `storage-worker-teardown-isolate` — the same isolate/Worker-boundary family as this wedge.

If `macos 1/2` wedges here, the defect belongs to the base and not to anything in #4837.

## Verification

No local suite, typecheck, or build was run. The hosted `macos 1/2` job on this head is the experiment.

## Checklist


lidge-jun:
Diagnostic probe complete. Result recorded in the forensics table on #4837; closing.

PR #4879, opened 2026-09-17
docs(devlog): plan the L6 contract and quality fix lane

lidge-jun:
## Summary

- Adds the planning unit for lane L6 under `devlog/_plan/260917_l6_contract_quality_fixes/`. Six narrow contract defects that share no code path, each shipping as its own pull request against `dev`: the Anthropic clean-stop phase regression (#4855), Z.AI model discovery (#4822), model capacity on the `/v1/models` top level (#4802), the Alibaba Token Plan catalog refresh (#4788), custom-model context validation (#4863), and the Claude `message_start` usage contract (#4857).
- Records the constraints each unit operates under, so the implementation pull requests do not have to relitigate them. Two of these are repository gates checked against the tree rather than assumed:
  - `gui/src/pages/Models.tsx` is recorded in `tests/fixtures/file-size-baseline.json` at 2792 lines and measures 2792, so the custom-model validation change would fail `tests/ci-workflows/file-size-ratchet.test.ts` at its submitted net +6. The plan resolves this rather than deferring it.
  - `enforce-target` requires a screenshot for any pull request mentioning the dashboard UI, which is why that unit stops at "branch pushed, pull request open" and escalates instead of working around the gate.
- Records how the three units that build on existing contributor work will attribute it. Each of those implementation branches gets a `Co-authored-by` trailer in a branch commit, per the "Landing another author's work" rule in `AGENTS.md`. Those trailers live on the implementation branches; this pull request contains no contributor code, only Markdown.

Documentation only. Nothing in the build, typecheck, or test path reads from `devlog/`.

## Verification

- No local test, typecheck, install, build, or `ocx` invocation was run for this lane. Verification is static source reading plus hosted CI at the exact head.
- Every source claim in these documents was read at `f1dfda8e48`: `src/bridge/sse.ts:1174`, `src/bridge/response-json.ts:530`, `src/responses/truncated-stop-reason.ts`, the `zai` row in `src/providers/registry/entries-extended.ts`, `providerModelsUrl` and `extractProviderModelItems` in `src/providers/model-discovery.ts`, `messageSnapshot`/`ensureStarted`/`finish` in `src/claude/outbound.ts`, and `scripts/file-size-ratchet.ts` against `tests/fixtures/file-size-baseline.json`.
[...]

PR #4889, opened 2026-09-17
fix(models): mirror model capacity onto the /v1/models top level

lidge-jun:
Continues the work in #4802 by @Yum-wu on current `dev`, with the long-tier case pinned. The `Co-authored-by` trailer is in the branch commit so it survives the squash. #4802 stays open.

## Summary

A model row declared its capacity only inside the nested `capabilities` object that Cursor validates. Clients that read the flat top level — the common shape for external discovery — saw a model with no declared context window or output limit, so a row that Cursor displayed correctly looked capacity-less everywhere else.

`context_window`, `context_length` and `max_output_tokens` are now mirrored onto the row's top level from the same values that feed the nested object. A value that does not pass `positiveInt` omits the key rather than emitting `0` or `NaN`, which would be worse for a client than an absent field.

The submitted change covered the flat input and the empty input but never pinned the long-tier row, which left the actual policy unrecorded. When a model has both `contextWindow` and `longContextWindow`, the mirrored value follows `effectiveContextLength` and therefore advertises the long window. This adds the assertion for that behaviour as implemented, so the choice is visible in the suite instead of implied by it.

Whether the long window or the base window is the right thing to advertise to a client that does not know about Max Mode is a product decision, and it is left to the maintainer. This records the current answer rather than changing it.

## Verification

- No local verification was run for this lane: no `bun test` in any form, no `bun run test`, `test:changed`, `typecheck`, `bun x tsc`, `bun install`, `build:gui`, or `ocx` invocation. A past local suite run deleted real `~/.opencodex` data, so this lane verifies by static source reading and treats hosted CI at the exact head as the evidence of record.
- `positiveInt` was read directly and rejects non-finite values, zero, negatives and non-safe integers, so an invalid input omits the top-level key.
- Nested and top-level values are produced from the same `effectiveContextLength` and `maxOutputTokens` locals, so the two shapes cannot drift for a given row.
- Added coverage in the existing `tests/providers/cursor/cursor-local-models-schema.test.ts` — no new test file and no `layout.json` entry. It pins omission for absent, zero, negative, `NaN` and non-safe-integer inputs; pins a `272000`/`922000` long-tier row advertising `922000` in both shapes; and extends the existing `/v1/models` fetch case, which previously asserted only nested values, to assert the top-level fields on both a flat row and a long-tier row. That last gap was the CodeRabbit finding on the original.
- The module header records that the top-level keys exist for external client discovery and that a new nested capacity value must gain its mirror in the same change, so the next contributor does not edit one shape alone.
- `structure/catalog.md` is updated in this commit because it owns this surface. No changed file is in `tests/fixtures/file-size-baseline.json`.

## What the long tier advertises, and what this pull request does not decide

When a row has a long tier, both shapes now report the same effective long window, and the base window stays visible as `pricing.overrides[].min_prompt_tokens`, which is the long-context threshold Cursor reads. That optimistic number is not introduced here: `capabilities.context_length` on `dev` already resolves to the long window through the same `effectiveContextLength`, and the threshold override already ships beside it. This change mirrors that value onto the top level so the two shapes agree.

Lowering only the top level to the base window was considered and rejected. It would make one row state two different capacities depending on where a client looks, which is worse than a single optimistic number a client can reconcile against the threshold override. Whether the base window is the right thing to advertise at all is a separate product decision, and making it would have to move the nested value, the top-level mirror and the threshold together in its own change. That is out of scope here.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup. Additive top-level keys and the assertions that pin them.
- [x] Docs or release notes were updated when needed. `structure/catalog.md` records the contract.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. The added fields are capacity integers already present in the response under a different key; no auth, credential, or account identifier is exposed and nothing new is logged.

Co-authored-by: Yum-wu <118118663+Yum-wu@users.noreply.github.com>


lidge-jun:
## 리뷰 · 우선순위 76 / 80

이 PR은 L6 계약·품질 레인의 U4를 코드로 옮긴다. 플랜 #4879의 `devlog/_plan/260917_l6_contract_quality_fixes/030_carried_catalog_and_gui_prs.md`가 적은 대로, 기여자 @Yum-wu의 #4802를 현재 `dev`에 다시 올려 담는다. Co-authored-by는 브랜치 커밋에 들어 있어 스쿼시 뒤에도 남는다. 원본 #4802는 열린 채로 둔다. 현재 `dev` HEAD는 `a1fe84b31`(패키지 2.58.0, tip #4880 L4 플랜; 아래 #4879 L6·#4882 L7). 브랜치 `codex/l6-models-top-level-capacity` merge-base는 `f1dfda8e48`라 tip보다 docs 3커밋 뒤다. types.ts/config.ts 분할과는 무관하다.

문제는 발견 표면이 둘인데 값이 한쪽에만 있다는 것이다. Cursor는 중첩 `capabilities.context_length` / `capabilities.max_output_tokens`를 읽는다. 그런데 pi-ai·DSH·LibreChat처럼 흔한 외부 클라이언트는 `/v1/models` 행의 맨 위 `context_window`·`context_length`·`max_output_tokens`만 본다. Cursor 화면에서는 용량이 보이는데, 같은 행을 flat으로 읽으면 창·출력 한도가 없는 모델처럼 보인다. 그 간극이 이 픽스의 전부다.

고치는 곳은 `src/server/models-capabilities.ts`의 `modelCapabilityFields` 한 함수다. 이미 쓰던 `positiveInt`로 통과한 값만 중첩 capabilities에 넣었는데, 같은 유효 값을 탑레벨에도 거울처럼 붙인다. 0·음수·NaN·unsafe 정수는 키 자체를 빼서, 가짜 0이 용량처럼 보이는 일을 막는다. long-tier가 있으면 `effectiveContextLength` = long window를 중첩·탑레벨 양쪽에 쓰고, `pricing.overrides[].min_prompt_tokens`는 예전처럼 base window에 둔다. 캐리 쪽에서 long-tier 단언을 테스트에 새로 박아, 「긴 창을 광고한다」는 제품 선택이 스위트에 기록되게 했다. 그 선택이 맞는지까지는 이 유닛이 바꾸지 않는다.

변경 범위는 작다. 구현 +16 근처, `tests/providers/cursor/cursor-local-models-schema.test.ts`에 flat·empty·long-tier·raw list 단언, `structure/catalog.md`에 한 문단. 호스트 CI는 changes/gates/hygiene/docker smoke/keyring 등이 이미 초록이다. mergeStatus는 BLOCKED(리뷰·남은 게이트)지만 코드 증거는 충분해 보인다. tip 리베이스만 하면 exact-head도 맞는다.

라인 `src/server/models-capabilities.ts` ~171 - `effectiveContextLength`로 long/base를 한곳으로 모은 뒤 capabilities와 탑레벨이 같은 숫자를 본다.
라인 같은 파일 ~191-197 - `context_window`/`context_length`/`max_output_tokens` 탑레벨 미러. 유효하지 않으면 키 생략.
테스트 long-tier - 272k/922k 입력이 탑레벨·capabilities 모두 922k인지 고정. 플랜이 지적한 「정책이 테스트에 안 보임」 구멍을 메운다.
`structure/catalog.md` - raw `/v1/models`이 중첩과 탑레벨을 같이 광고한다고 적는다.

메인테이너의 판단이 필요한 지점
- long window를 탑레벨에 광고하는 지금 선택이 제품으로 맞는지. 바꾸려면 이 PR이 아니라 후속 정책 PR이 맞다.
- tip(`a1fe84b31`) 리베이스 후 exact-head CI를 한 번 더 볼지.
- 원본 #4802는 열린 채 유지(캐리 정책). 랜딩 후 landed-via 코멘트·라벨로 정리할지.

너의 추천
KEEP — tip에 맞춘 뒤 CI 초록이면 머지. #4802는 캐리 원본으로 남기고, 머지 후 `Landed via #4889` 정리. types/config 분할에 무효화되지 않으므로 닫지 말고 랜딩한다.

이 댓글은 grok-bot이 작성했습니다

Issue #4893, opened 2026-09-17
Provider transientRetryOn5xx never reaches the Responses passthrough lane


lidge-jun:
### Client or integration

Direct HTTP/API client

### Area

Proxy and routing

### Summary

A provider's `transientRetryOn5xx` policy never reaches the Responses passthrough lane. Configuring it on a key-auth provider whose adapter is `openai-responses` has no effect on `/v1/responses`, because that lane never reads the policy and instead uses a fixed constant.

I expected an operator who sets `transientRetryOn5xx: { attempts: 5 }` on such a provider to change how many upstream sends that provider gets on a transient 5xx. Today the number is always `TRANSIENT_RETRY_MAX_ATTEMPTS`.

The evidence is three source facts that compose:

1. `transientRetryPolicyFor` has exactly four call sites in `src/`: `src/server/chat-native.ts:301`, `src/server/responses/adapter-continuation.ts:207` and `:411`, and `src/server/responses/adapter-dispatch.ts:305`.
2. `createResponsesPassthroughAdapter` declares `passthrough: true as const` (`src/adapters/openai-responses/passthrough.ts:190-193`), and `src/server/responses/core.ts:116` returns into `executePassthroughResponse` on exactly that flag — before `createAdapterContinuations` or the adapter dispatch path is constructed. So three of those four call sites are unreachable for this adapter.
3. `src/server/responses/passthrough-dispatch.ts` does not import `transientRetryPolicyFor` at all. Its initial send is wrapped at `:760` with `attempts: remainingTransientSendBudget(TRANSIENT_RETRY_MAX_ATTEMPTS)`, and three further call sites in the same file do the same for the recovery legs. `TRANSIENT_RETRY_MAX_ATTEMPTS` is `3` (`src/lib/upstream-retry.ts:106`).

So the lane is not missing transient-5xx replay — it has an unconditional one that no provider setting can tune, in either direction.

### Reproduction

This is a configuration-surface defect, and it is visible statically as well as at runtime.

Statically, from a checkout at the `dev` tip:

```
rg -n "transientRetryPolicyFor" src/          # four call sites, none in passthrough-dispatch.ts
rg -n "passthrough" src/adapters/openai-responses/passthrough.ts | head
sed -n '110,125p' src/server/responses/core.ts # the early return on adapter.passthrough
sed -n '755,790p' src/server/responses/passthrough-dispatch.ts # the fixed constant
```

At runtime:

[...]

PR #4925, opened 2026-09-17
fix(responses): honor the provider transient-5xx policy on the passthrough lane

lidge-jun:
## Summary

A provider's `transientRetryOn5xx` never reached the Responses passthrough lane. An operator who
set it on a key-auth provider whose adapter is `openai-responses` changed nothing in either
direction — `attempts: 1` still sent three times and `attempts: 10` also sent three — while the
same provider on `openai-chat` was tuned normally. That asymmetry is the defect.

Two things had to change together, which is why widening the adapter gate alone does not fix it:

- `transientRetryPolicyFor` rejected every adapter but `openai-chat`.
- The lane never called it. `createResponsesPassthroughAdapter` sets `passthrough: true` and
  `core.ts` returns into `executePassthroughResponse` on that flag, before the three call sites
  that read the policy are constructed. `passthrough-dispatch.ts` passed the constant
  `TRANSIENT_RETRY_MAX_ATTEMPTS` at four dispatch sites — the initial send, the OAuth-401 replay,
  the same-target 429 replay and the validated rebuild — and asked `sendBudgetExhausted()` at that
  same constant twice more.

With the gate widened and nothing else, the reproduction is unchanged at three sends. This PR
carries the gate edit from #4800 with a `Co-authored-by` trailer and wires the lane to it, so the
ladder now resolves from the provider row at every leg including recovery.

### Budget accounting

`remainingTransientSendBudget(cap)` resolves to `RequestExecutionBudget.remainingBaseSends(cap)`,
which is `min(cap, baseSendAllowance - spent)`, and `core.ts` gives every Responses request the
guarded profile whose `baseSendAllowance` is 3.

The shape of that function is the one real trap here. `cap` bounds what **remains**, not what the
request may spend in total. That is the right reading for the fixed constant — every leg may ask
for up to three and the request-wide allowance bounds the total — but
`transientRetryOn5xx.attempts` is documented as the total sends for one request including the
first. Passing the configured value straight through would silently turn it into a per-leg
ceiling, so a provider configured at one send could still reach upstream again on a recovery leg.
The first revision of this PR did exactly that and its own regression caught it; the second commit
is the fix.

`transientSendCapFor(configured, sendsUsed)` therefore reduces the configured total by what the
request has already sent, and the result is intersected with the base allowance. An absent policy
returns the constant unchanged, so a provider that configures nothing is unaffected at every call
site.

That intersection is the deliberate settlement the issue asked for, and it answers whether a
provider can now widen a request-wide bound: it cannot. Configuring below the allowance narrows
the request exactly, so `attempts: 1` sends once and no recovery leg may dispatch — the direction
the reporter demonstrated as broken. Configuring above it does not raise the bound that exists to
stop per-request amplification (#4546). This limit is now stated in the reference rather than left
implicit, because leaving it implicit would reproduce the original complaint one threshold higher.
Raising `baseSendAllowance` per provider is a policy decision about the guarded profile, not a
wiring fix, and is out of scope here.

Every leg had to move together. `sendBudgetExhausted()` asking at the constant while the sends
dispatch at a configured value would tell a provider with headroom it was spent, and would let one
configured below the constant pass the check and then be refused at the send. It now takes the cap
as a parameter, defaulted so every other caller is unchanged.

### The non-replayable boundary

`isNonReplayableResponse` is checked inside `fetchWithTransientRetry` and at each recovery
branch, and is not a function of `attempts`, so a higher configured value cannot become a way to
obtain the resend that marker forbids. The existing case pinning a marked 504 returning after one
send under `attempts: 3` still holds.

### Scope

`authMode` stays fail-closed, so the ChatGPT account pool (`authMode: "forward"`) still receives
`null` from the policy and keeps exactly the ladder it has always had. Only key-auth providers on
the two named adapters can tune anything.

Closes #4893.

### Relationship to #4800

#4800 is correct as far as it goes and is not superseded in intent. Its change is the gate edit
in `src/providers/key-failover.ts`, carried here verbatim in behavior with
`Co-authored-by: Yum-wu`, because the gate and the lane wiring are not separable: merging the gate
alone leaves the reported behavior unchanged and would read as a fix that is not one. Leaving that
PR's disposition to the maintainers.

### Still open

The report also mentions a key-auth `openai-responses` provider terminating on its first send with
an "interrupted with no error" experience. This change does not explain that and does not claim
to: the lane already retried three times, so a single send means something else ended the loop.
The four candidates the issue lists remain the right ones and each is decidable from one captured
response.

## Verification

Local suites were not run for this change, by explicit maintainer instruction; correctness is
argued from source and proven by hosted CI at this head.

- `tests/responses/responses-passthrough-transient-policy.test.ts` (new). A source oracle over
  `passthrough-dispatch.ts` asserts that every `fetchWithTransientRetry` site takes its attempts
  from the resolver, that the recovery allowance does too, that no dispatch or exhaustion check is
  left on the constant, and that no `sendBudgetExhausted()` call omits the cap. Matching is
  whitespace-independent so reformatting cannot silently retire an assertion. Behavioural cases
  drive `transientSendCapFor` and `createResponsesSendBudget` directly: an absent policy resolves
  to the constant at every send count, a configured total is measured against what the request
  already sent, a configured 1 ends a ladder the constant would have continued, and a configured
  10 does not lift the request-wide allowance.
- `tests/providers/upstream-transient-retry.test.ts` updated so key-auth `openai-responses`
  qualifies, other adapters stay rejected, and OAuth/forward/local stay rejected for both admitted
  adapters.
- Registered in `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`.
- `docs-site` reference updated in English and all seven translated locales, which previously
  documented the defect as intended behaviour ("never reads this option").
- Hosted `Cross-platform CI` at this exact head is the gate.

## Checklist

Co-authored-by: Yum-wu <118118663+Yum-wu@users.noreply.github.com>

## Summary by CodeRabbit

* **New Features**
  * Key-authenticated `openai-responses` providers now support `transientRetryOn5xx`.
  * Configured retry limits apply consistently across passthrough recovery attempts, subject to the request-wide send limit.
  * Forward-authenticated providers retain their default retry behavior.

* **Documentation**
  * Updated provider configuration guidance across supported languages.

* **Tests**
  * Added regression coverage for configurable passthrough retry behavior and request budget limits.


lidge-jun:
## 리뷰 · 우선순위 77 / 80

이 PR은 이슈 #4893을 고칩니다. 지금 `dev`(`43cd1ade1`, package 2.59.0)에서는 키 인증 `openai-responses` 프로바이더에 `transientRetryOn5xx`를 넣어도 `/v1/responses` 패스스루 레인이 그걸 전혀 읽지 않습니다. 레인 안에서는 고정 상수 `TRANSIENT_RETRY_MAX_ATTEMPTS`(=3)만 쓰이기 때문입니다. 같은 프로바이더를 `openai-chat`으로 두면 설정이 먹히는데, 패스스루만 안 먹는 비대칭이 바로 버그입니다.

원인은 한 군데가 아닙니다. `transientRetryPolicyFor`가 예전에는 `openai-chat`만 통과시켰고, 동시에 `createResponsesPassthroughAdapter`가 `passthrough: true`를 켜면 `core.ts`가 `executePassthroughResponse`로 바로 빠져서 `adapter-dispatch` / `adapter-continuation`에 있는 정책 읽기 세 곳이 패스스루에는 닿지 않습니다. 그 결과 `passthrough-dispatch.ts`는 초기 전송·OAuth-401 재전송·동일 타깃 429 재전송·validated rebuild 네 곳에서 상수만 넘기고, `sendBudgetExhausted()`도 같은 상수로만 물어봤습니다. #4800은 게이트만 `openai-responses`까지 넓히는데, 레인 배선이 없으면 재현은 여전히 “항상 세 번”입니다. 이 PR은 #4800의 게이트 수정을 `Co-authored-by: Yum-wu`로 그대로 가져오고, 레인까지 한 번에 연결합니다.

배선 방식도 예산 규칙을 깨지 않게 잡혀 있습니다. `transientSendAttempts()`가 `route.provider`를 호출마다 다시 읽고(복구 루프에서 프로바이더가 바뀔 수 있음), 그 값을 `remainingTransientSendBudget` / `recoverySendAllowance` / `sendBudgetExhausted(cap)`에 같이 넣습니다. 설정값은 사다리(cap)이지 새 허용량이 아니고, `RequestExecutionBudget.remainingBaseSends`와 교차하므로 `attempts: 1`은 한 번만 보내고 `attempts: 10`은 요청 전체 `baseSendAllowance`(가드 프로필 기준 3)을 올리지 못합니다. 이건 #4546 증폭 방어를 유지하면서 #4893의 “줄이기” 방향을 고치는 타당한 타협입니다. `authMode`는 여전히 fail-closed라 ChatGPT `forward` 풀은 `null`을 받아 예전 사다리를 유지하고, `isNonReplayableResponse`도 attempts와 무관해서 표시된 504를 더 보내게 만들지 않습니다.

검증은 소스 오라클(`responses-passthrough-transient-policy.test.ts`)로 `fetchWithTransientRetry` 네 곳 이상이 정책 리졸버/`allowance.attempts`를 쓰는지, 상수 잔존이 없는지, 소진 검사가 같은 cap인지 잡고, 예산 쪽은 `createResponsesSendBudget`으로 `attempts: 1` vs 상수 3의 차이를 직접 확인합니다. `upstream-transient-retry.test.ts`는 게이트 자격과 oauth/forward/local 거부를 두 어댑터 모두에 대해 고정합니다. 영문 reference는 교차(intersection)와 recovery legs까지 고쳤고, 레이아웃에 새 테스트가 등록됐습니다. 로컬 스위트는 메인테이너 지시대로 안 돌렸고 hosted CI가 게이트입니다(작성 시점 pending). types/config 분할 캠페인과 충돌하지 않는 좁은 버그픽스 열차 항목입니다.

docs-site/src/content/docs/ko/reference/configuration/providers.md (및 fr/ja/ru/tr/zh-cn/zh-tw) - 어댑터 목록·forward 제외는 맞췄지만, 영문에만 있는 “패스스루에서 request-wide send allowance와 교차한다 / recovery legs까지 적용된다” 문장이 번역본에 없음. 운영자가 `attempts: 10`을 올렸는데도 3에서 멈추는 이유를 영문만 설명함
#4800 - 게이트만 있는 선행 PR이 열려 있음. 이 PR이 게이트+레인 배선까지 가져왔으므로 머지 후 #4800은 superseded/close가 자연스러움(본문도 disposition을 메인테이너에 맡김)
이슈 #4893의 “첫 전송에서 interrupted with no error” 증상 - 본문이 의도적으로 범위 밖으로 둠. 패스스루는 원래도 세 번 재시도하므로 그 증상은 다른 원인 후보 네 개를 따로 잡아야 함

메인테이너의 판단이 필요한 지점
- #4800을 이 PR 머지와 함께 `landed-via`/`superseded`로 닫을지, 아니면 기여자 크레딧만 남기고 별도 코멘트로 정리할지
- 번역 reference에 영문과 같은 allowance-교차 설명을 이번 PR에 더 넣을지, 후속 docs PR로 미룰지
- `baseSendAllowance`를 프로바이더별로 올리는 정책은 이번 범위 밖인데, 그걸 따로 로드맵에 올릴지(본문은 out of scope로 명시)

너의 추천
hosted Cross-platform CI가 이 헤드(`e21b78beb`)에서 초록이면 `dev`에 머지. 머지 직후 #4800은 superseded로 닫고 #4893은 Closes로 같이 닫히게 두면 됨. 번역본 allowance-교차 문장은 머지를 막지 말고, 원하면 짧은 후속 docs로 맞추면 됨. “interrupted with no error”는 이 PR과 분리해서 캡처된 응답 하나로 네 후보를 가리는 후속 이슈로 남겨 두기.

이 댓글은 grok-bot이 작성했습니다

PR #4927, opened 2026-09-17
fix(combos): hop a response_format capability refusal to the next target

lidge-jun:
## Summary

A combo opens a new conversation, the shadow title call carries `response_format`, and the first
target's gateway refuses it with HTTP 400. The chain stops instead of trying the target behind
it, so the user gets `Provider error 400` from a combo that had a working alternative.

`comboFailureDecision` reaches
`["origin_rejected", "context_length_exceeded", "invalid_request_error"].includes(error.code)`
and returns `stop`. `isRequestLocalTargetIncompatibility` runs first and could have returned
`hop`, but it refuses at its own first guard: the gateway's `invalid_parameter_error` is not in
the generic code set, and none of its three accepted shapes — `Unsupported parameter: user`, an
`unsupported_value` on `reasoning.effort`, and a model-scoped image-input rejection — describes a
`response_format` refusal.

There is a second blocker the issue body does not name, and it explains the reported text. The
gateway reports the refusal inside a single SSE frame, so `normalizeUpstreamErrorText` cannot
parse `data: {...}` as JSON: `classificationText` keeps the raw frame and `upstreamCode` arrives
`undefined`. Widening the code set alone would still fail on the unparsed frame, which is why
this was re-derived rather than patched from the issue text.

### Why neither obvious option was taken

Hopping on every 400 would replay a genuinely malformed request against every remaining target.
Dropping `response_format` would silently change the output contract the caller asked for, on a
path whose entire purpose is a structured result.

The verdict is narrowed to a capability claim instead: the message must name `response_format`
**and** state that it is unavailable or unsupported. An invalid-schema complaint names the field
and claims nothing about capability, so it stays terminal.

### The envelope

It keeps the discipline the existing predicate uses — HTTP 400, intact provider JSON,
`type: "invalid_request_error"`, a three-envelope depth budget, a 16,384-character bound — and
adds exactly two things:

- Its code set is the shared generic one plus `invalid_parameter_error`, held in its own set so
  the `user` and image branches are not widened by a code they were never reasoned about.
- One `data:` prefix is unwrapped, and only on a single-line body. That unwraps one frame rather
  than parsing a stream, so a multi-event body is left alone and still fails closed.

`param` may be absent or explicitly null, as the reported gateway sends; a param naming another
field contradicts the message and fails closed.

### What the next target receives

The same request, `response_format` included. A target that can honour the contract honours it,
and one that cannot is skipped in turn. Traversal stays finite because combo excludes each
attempted target and policy tries each candidate once. The verdict records no cooldown, because a
capability gap says the target is healthy and the request did not fit it. Cancellation, structured
origin and cyber-policy refusals, and the non-replayable post-send codes are all tested before it
and remain authoritative.

Closes #4903.

### Relationship to #4817

#4817 forwards a zero-output SSE bare error event to the next target only when
`comboFailureDecision` already says `hop`. This issue is the opposite half: the decision said
`stop`, so that path could never carry it. The two are complementary and neither closes the other.

## Verification

Local suites were not run for this change, by explicit maintainer instruction; correctness is
argued from source and proven by hosted CI at this head.

- `tests/routing/router-combo-failover-classification.test.ts` gains a `response_format`
  capability block next to the existing request-local cases. It pins the hop and the absent
  cooldown across all four envelopes the report produces, including the bare frame and the
  wrapper-plus-frame form; keeps a malformed `response_format` and an unavailability claim about
  another field terminal; and holds the envelope closed against reflected prose, a truncated body,
  a non-400 status, an unrecognized code, a wrong `type`, an oversized message, and a multi-event
  stream body. A final case asserts that origin, non-replayable and cyber-policy codes still
  outrank the new verdict.
- `structure/runtime.md` records the fourth envelope, since it enumerates the accepted three.
- No new test file, so the layout registry is unchanged.
- Hosted `Cross-platform CI` at this exact head is the gate.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of requests using `response_format` when a provider reports that the capability is unavailable or unsupported.
  * Requests now automatically try the next compatible combo target instead of stopping prematurely.
  * Providers are no longer temporarily excluded for this capability-specific refusal.
  * Added safeguards so unrelated, malformed, or ambiguous errors continue to be handled as terminal failures.


lidge-jun:
## 리뷰 · 우선순위 77 / 80

이 PR은 콤보가 새 대화를 열 때 그림자 제목 생성 호출에 `response_format`이 붙고, 첫 타깃 게이트웨이(보고된 Alibaba)가 HTTP 400으로 거절하면 체인 전체가 멈추던 버그를 고칩니다. 증상은 이슈 #4903에 나온 그대로입니다. 사용자는 `Provider error 400: data: {...}`만 보고, 뒤에 있는 deepseek 같은 대안 타깃은 한 번도 시도되지 않았습니다.

현재 `dev`(HEAD `43cd1ade1`, 패키지 2.59.0)의 `comboFailureDecision`은 먼저 `isRequestLocalTargetIncompatibility`로 hop 후보를 보고, 그다음 `origin_rejected` / `context_length_exceeded` / `invalid_request_error`면 stop 합니다. 게이트웨이는 `type: "invalid_request_error"`와 `code: "invalid_parameter_error"`를 돌려주는데, 기존 request-local 세 가지 모양(Unsupported parameter: user, reasoning.effort unsupported_value, 이미지 입력 거절)에는 `response_format` 거절이 없습니다. 게다가 거절이 한 줄 SSE `data:` 프레임 안에만 있어서 `normalizeUpstreamErrorText`가 JSON으로 못 풀고, `upstreamCode`가 undefined로 옵니다. 코드 집합만 넓혀도 프레임이 안 풀리면 여전히 stop입니다. 이 PR은 그 두 막힘을 같이 풉니다.

설계 선택이 중요합니다. 모든 400을 hop하면 진짜로 잘못된 요청이 남은 타깃마다 다시 나가고, `response_format`을 빼면 호출자가 원한 구조화 출력 계약이 조용히 바뀝니다. 그래서 hop 조건은 메시지가 `response_format`을 말하고, 동시에 unavailable/unsupported 계열로 “이 타깃이 그 능력을 못 준다”고 말할 때만입니다. “Invalid schema for response_format”처럼 필드만 가리키고 능력은 말하지 않으면 stop으로 남깁니다. 다음 타깃에는 `response_format`이 그대로 전달되고, 쿨다운은 없습니다(능력 공백이지 타깃이 아픈 게 아님). #4817(제로 출력 SSE bare error를 이미 hop인 결정에 실어 보냄)과 반대쪽 반쪽이라 서로 닫지 않고 보완합니다.

변경은 `src/combos/failover.ts`에 `RESPONSE_FORMAT_REFUSAL_CODES` / `namesResponseFormatIncapability` / `isResponseFormatCapabilityRefusal`을 추가하고, `comboFailureDecision`에서 일반 `invalid_request_error` stop 바로 앞에 hop을 넣으며, cooldown scope에도 none으로 연결합니다. 테스트는 보고된 네 봉투(raw / Provider error 래핑 / data: / 둘 다), 잘못된 스키마·다른 필드·봉투 경계·하드 거절 우선순위를 고정합니다. `structure/runtime.md`와 plan 노트도 같이 갱신합니다. Closes #4903. 로컬 스위트는 메인테이너 지시로 생략했고, hosted Cross-platform CI가 exact head 게이트입니다.

라인 structure/runtime.md (response_format 단락) - 새 글을 “a fourth envelope”라고 부르는데, 바로 아래 context-window overflow 단락도 여전히 “the fourth request-local verdict”입니다. 번호가 둘 다 네 번째라 읽는 사람이 헷갈립니다. response_format을 “fourth envelope among the 400 incompatibilities”처럼 세거나, overflow를 fifth로 고치는 쪽이 맞습니다.

라인 src/combos/failover.ts (isResponseFormatCapabilityRefusal, data: unwrap) - 한 줄 바디에서만 `data:`를 한 번 벗깁니다. 보고된 게이트웨이에는 맞고 멀티 이벤트는 fail-closed라 안전하지만, 앞에 공백/`data: `변형·CRLF만 다른 게이트웨이가 오면 다시 stop으로 떨어질 수 있습니다. 지금은 범위가 맞고, 새 재현이 오면 그때 넓히면 됩니다.

경로 tests/routing/router-combo-failover-classification.test.ts - hop/stop/쿨다운/봉투 경계·하드 코드 우선순위가 잘 고정돼 있습니다. 로컬 미실행은 지시대로 이해합니다. merge 전에는 이 exact head의 Cross-platform CI가 녹색이어야 합니다(지금 체크는 아직 pending/queued).

메인테이너의 판단이 필요한 지점
- structure/runtime.md의 “fourth” 번호 충돌을 이 PR에서 고칠지, 후속 문서 PR로 넘길지
- #4903를 이 PR만으로 닫을지(본문 Closes #4903), 아니면 다른 콤보 400 모양까지 더 볼지
- CI가 이 head에서 통과한 뒤에만 merge할지(트레인 규칙: exact SHA, cancelled ≠ pass)

너의 추천
exact-head Cross-platform CI가 녹색이면 merge 하세요. 문서의 “fourth/fourth” 번호만 한 줄 정리하면 더 깔끔하고, 기능 범위(#4903 + #4817 보완)는 이미 merge 가능한 수준입니다. types.ts/config.ts 분할에 무효화되지 않습니다.

이 댓글은 grok-bot이 작성했습니다

Issue #4874, opened 2026-09-17
[Provider compatibility] Cursor: incomplete tool-call streams poison conversation reuse


MerryEcho:
### Client or integration

Codex App

### Provider or upstream service

cursor

### OpenCodex version

2.57.0

### Endpoint or capability

/v1/responses Cursor protobuf tool calls

### Current behaviour

On Codex Desktop through OpenCodex, Cursor models (including `cursor/grok-4.6`) talk Connect protobuf, not Responses. When a Cursor stream ends with a client tool still open, OpenCodex fail-closes:

`Cursor stream ended with incomplete tool call(s)`

That error is streamed. It is not thrown, so the overflow/invalid_argument remint ladder never runs. `_cursorConversationId` is assigned at the start of `runTurn` and persisted by `continuationStateForResponse`. The next turn reuses that id (or the deterministic thread hash). Cursor's side of the conversation is still waiting for `mcpResult`, so later turns in the same thread interrupt more often. Neighboring threads that never interrupted stay clean.

Native Composer replay also flushes unpaired `toolCallStep` history without a result, which looks like an interrupted attempt. External wire models such as grok-4.6 skip native `mcpToolCall` replay, so history repair alone cannot recover them; conversation remint is the path.

This is not a Cursor 429 / Ultra quota issue. The incomplete-tool event is a truncated stream, and `CursorStreamTruncatedError` is defined but never thrown.

### Expected behaviour

Keep fail-closed on the current turn: do not emit a partial `tool_call_start`, and do not retry that send. After the incomplete-tool error is streamed, eligible non-isolated turns should remint the Cursor conversation id (same helper as overflow remint), persist the thread override for store:false clients, and invalidate the inherited checkpoint so the next turn does not resume a conversation left waiting for `mcpResult`. Isolated helper/compaction turns must not remint or donate that recovery to the parent.

Native Composer unpaired `toolCallStep` history should carry an honest `[missing tool_result for this tool_use in history]` placeholder (`isError: true`), matching Google #2199. External grok-4.6 stays on conversation remint.

### Minimal redacted request or reproduction

Codex App -> local OpenCodex proxy -> `cursor/grok-4.6` (or another Cursor model). Run a tool-using thread until the proxy reports `Cursor stream ended with incomplete tool call(s)`, then continue the same thread. Later turns in that thread interrupt more often than a neighboring thread that never had a truncated tool stream.

### Actual response or error

```text
[...]

PR #4929, opened 2026-09-17
fix(cursor): stop grok-4.6 tool-result echo from poisoning later turns

lidge-jun:
## Summary

Grok-4.6 through Cursor writes a real sentence and then pastes the replayed `[Tool Result]`
envelope. `CursorToolResultEchoSniffer` only watches the opening bytes of a turn, so a mid-message
echo is never seen: it reaches Codex, is stored as assistant text, and the next turn replays it,
which primes the model to echo again. That is the snowball reported in #4874 and still observed
after the incomplete-tool remint in #4875.

Four changes, each scoped to one part of that loop:

- **Assistant root replay drops echoed envelopes** (`stripAssistantEchoedToolEnvelope`), so the
  transcript stops feeding itself. This applies only to assistant-role history; the envelopes the
  adapter itself builds for `toolResult` messages are untouched.
- **A mid-stream echo remints the conversation for the NEXT turn**, on its own bounded allowance.
- **The retained thread remint override beats a stale stored `_cursorConversationId`**, so a second
  Responses chain in one Codex thread stops ping-ponging the rotated-away conversation.
- **`Write` joins the unavailable neighboring-agent tool list**, which Grok kept probing.

### Three scopes kept separate

| | |
|---|---|
| Retry the current send | **Not done.** The echo already reached the client; resending would be an uncertain replay. |
| Prevent a new conversation from being poisoned | **Done.** Replay strip plus next-turn remint. |
| Recover an already-poisoned conversation | **Not attempted.** An affected thread still needs a new task. |

### Why the echo remint gets its own budget

`dev` gained a bounded incomplete-tool remint allowance during #4875's review, precisely to stop
unbounded conversation rotation. A mid-stream echo cannot be quarantined, so left uncapped a model
that echoes every turn would remint on every turn — the same failure that allowance was added to
prevent. It is a **separate** counter rather than a shared one because echoing is cheap and
repeatable while an incomplete client-tool stream is rare and structural; on one counter the cheap
failure would spend the allowance the other recovery depends on. When the incomplete-tool arm has
already reminted in the same turn, the echo arm does not rotate again.

The two bounded budgets now share one internal `createCursorRemintBudget` helper instead of a
second hand-copied map. The extracted code preserves the existing record/prune/LRU behaviour
exactly; the incomplete-tool budget's caps, TTL, and exported function names are unchanged.

### Why the replay strip stops at a blank line

The envelope has no terminator we can recognise — it is a marker line plus arbitrary result text —
and the observed copies are not byte-exact (live probing caught a whitespace-spliced call id), so
matching against the replayed envelope is not available either. Truncating from the marker to the
end of the message was the alternative, and it discards a genuine answer whenever the model resumes
after the echo. The strip therefore starts at a whole-line marker and ends at the next blank line.

The tradeoff is stated in the code and in `structure/providers/cursor.md`: an envelope whose pasted
body contains its own blank line leaves a remainder in replay. That is the safer direction to be
wrong in, because conversation remint is the primary defence and this filter only stops the
transcript from re-priming itself. Inline prose such as "the string [Tool Result] appeared"
survives, since only whole-line markers count.

## Relationship to #4900

Carries #4900 onto current `dev` rather than rebasing it. That branch holds a **pre-squash copy of
#4875** plus one increment (`5702c8e4`); #4875 landed on `dev` as `ee2883316095174824cfd2b02f569cbee7cd8ae0`
with review hardening the copy predates. All three conflicting files — `src/adapters/cursor.ts`,
`tests/providers/cursor/cursor-adapter.test.ts`, and `structure/providers/cursor.md` — conflict for
that reason alone, so `dev` is authoritative for every shared hunk and only the increment is
reapplied here. `structure/providers/cursor.md` is untouched by that increment, so its entire
conflict was stale-copy residue.

The increment is reapplied with two changes rather than verbatim: the remint is routed through the
bounded budget described above (the original condition was written against the unbounded version
and would have bypassed both the cap and the `contextUsageStoreCheckpoints` isolation), and the
replay strip is bounded at the blank line instead of truncating to the end of the message.

The conversation-id priority reorder is carried unchanged. It was reviewed and is correct:
`rememberCursorThreadConversation` is written only by the remint path in `cursor.ts`, so the thread
store holds remint results and a disagreeing stored id is the pre-remint value.

Refs #4874. Does not close it: an already-poisoned conversation is still unrecovered.

## Verification

Static only. This lane does not run the local suite, typecheck, build, install, or `ocx`; a past
local run deleted real user data under `~/.opencodex`. Evidence is source reasoning plus hosted CI
on this head.

Traced rather than executed:

- The strip is reached only from `assistantRootText`, so the `[Tool Result]` roots that
  `cursor-blob.test.ts`, `cursor-tool-result-invocation.test.ts`, and
  `cursor-repetition-breaker.test.ts` assert on are built in the `toolResult` branch and are
  unaffected.
- The new adapter test uses the `{ createTransport: factory as never }` shape its passing
  neighbours in the same file already use, and `toolResultBody` supplies the `toolResult` message
  the mid-stream observer needs to arm on an external wire model.
- Every helper the new tests import is exported: `stripAssistantEchoedToolEnvelope`,
  `CURSOR_ENVELOPE_ECHO_REMINT_MAX`, `cursorEnvelopeEchoRemintScopeKey`,
  `recordCursorEnvelopeEchoRemint`, the two `clear*ForTests` helpers,
  `lookupCursorThreadConversation`, and `rememberCursorThreadConversation`.
- No new test file, so `scripts/test-layout/layout.json` and
  `tests/fixtures/test-layout-expected.json` need no entry; both touched test files are already
  registered. Neither touched source file carries a `tests/fixtures/file-size-baseline.json` cap.

New coverage, in already-registered files:

- `cursor-envelope-echo-retry.test.ts` — the strip keeps leading commentary, **keeps a real answer
  written after the echo**, ignores an inline marker mention, and empties a prefix-only envelope; a
  mid-stream echo rotates the conversation and the next turn uses the new id while the echo still
  reaches the client; the echo allowance is bounded at its cap and leaves the incomplete-tool
  allowance untouched.
- `cursor-request-builder.test.ts` — the thread remint override beats a stale stored id, and
  isolated helpers ignore it.

Not verified: no live Cursor traffic was run, so this does not claim grok-4.6 stops echoing. It
claims the echo no longer survives into the next turn's replay and no longer reuses the same
conversation, with rotation bounded.

## Checklist

- [x] Docs or release notes were updated when needed. `structure/providers/cursor.md` gains a
      "Mid-stream envelope echo" section covering the remint, its separate bounded allowance, the
      replay strip and its stated tradeoff, and the conversation-id precedence.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. No auth,
      credential, or network-boundary change; diagnostics keep logging only a truncated
      conversation hash and the wire model, never content.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Cursor responses when tool-result envelopes are echoed mid-stream.
  - Preserves surrounding assistant prose while removing duplicated tool-result content.
  - Automatically refreshes the conversation for eligible follow-up turns, with bounded recovery to prevent repeated retries.
  - Ensures refreshed conversations are reused consistently while isolated helper requests remain separate.

- **Improvements**
  - Updated tool guidance to recognize `Write` and avoid suggesting unavailable Cursor-native tools.

- **Documentation**
  - Added Cursor provider documentation covering echoed tool envelopes, response cleanup, and conversation recovery.


lidge-jun:
## 리뷰 · 우선순위 77 / 80

이 PR은 Cursor 쪽에서 grok-4.6이 진짜 문장을 쓴 뒤 `[Tool Result]` 봉투를 그대로 붙여 넣는 문제를 막습니다. 지금 `dev`의 `CursorToolResultEchoSniffer`(접두 sniffer)는 턴 앞부분만 봅니다. 그래서 중간에 오는 에코는 못 잡고, 그 텍스트가 Codex에 들어가 assistant 기록으로 남고, 다음 턴 리플레이가 그걸 다시 먹여서 모델이 또 에코하게 됩니다. #4874에서 말한 눈덩이이고, #4875 incomplete-tool remint만으로는 안 잡히는 구멍입니다. 이슈 #4874 자체는 이미 닫혀 있지만, 본문이 말하는 mid-stream 에코 루프는 아직 `dev`에 남아 있습니다.

고치는 축은 네 개입니다. (1) `stripAssistantEchoedToolEnvelope`가 assistant 루트 리플레이에서만 에코 봉투를 지웁니다. `toolResult` 메시지가 만드는 정상 봉투는 건드리지 않습니다. (2) mid-stream 관찰자가 찾으면 이번 턴은 그대로 두고, 다음 턴 conversation id만 돌립니다. (3) `resolveCursorConversationId`에서 thread remint override가 오래된 `_cursorConversationId`보다 앞섭니다. 한 Codex 스레드 안 두 번째 Responses 체인이 방금 버린 대화를 다시 붙잡는 일을 막습니다. (4) 이웃 에이전트 도구 목록에 `Write`를 넣어, grok이 계속 찔러 보던 경로를 안내에서 끊습니다.

예산 설계가 핵심입니다. #4875가 incomplete-tool remint에 상한을 둔 이유와 같습니다. 에코는 싸고 매 턴 반복될 수 있어서, 같은 카운터를 쓰면 incomplete-tool 회복 예산을 다 써 버립니다. 그래서 `createCursorRemintBudget` 헬퍼 하나로 두 예산을 만들고, 에코 쪽은 `CURSOR_ENVELOPE_ECHO_REMINT_MAX = 3`으로 따로 둡니다. incomplete-tool이 같은 턴에 이미 remint했으면 에코 팔은 한 번 더 돌리지 않습니다. 스트립은 마커 줄부터 다음 빈 줄까지입니다. 끝까지 자르면 에코 뒤에 쓴 진짜 답이 다음 리플레이에서 사라지므로, 빈 줄에서 멈춥니다. 봉투 본문 안에 빈 줄이 있으면 나머지가 남을 수 있다는 tradeoff는 `structure/providers/cursor.md`와 코드 주석에 명시되어 있습니다.

베이스는 지금 tip `43cd1ade`(`#4623` offload notes 정리, package 2.59.0)입니다. #4900을 rebase하지 않고, #4875가 `dev`에 들어간 `ee288331` 이후 상태 위에 increment만 다시 얹었습니다. #4900은 지금 `dirty`(충돌)라서 이 PR이 그 자리의 깨끗한 운반선입니다. 검증은 static + hosted CI만입니다. 로컬 suite/typecheck/build/`ocx`는 돌리지 않았고, 본문대로 과거 로컬 실행이 `~/.opencodex` 사용자 데이터를 지운 적이 있어 의도적으로 피했습니다. 테스트는 이미 등록된 `cursor-envelope-echo-retry.test.ts`와 `cursor-request-builder.test.ts`에 스트립·remint·예산 독립·stale id 우선순위를 추가했습니다.

라인 stripAssistantEchoedToolEnvelope - 빈 줄에서 끊기 때문에, 에코 본문 안에 빈 줄이 있으면 그 뒤 조각이 리플레이에 남을 수 있습니다. remint가 본방어라서 방향은 맞지만, 이미 독이 든 스레드는 여전히 새 태스크가 필요합니다.

경로 resolveCursorConversationId - remint override가 stored `_cursorConversationId`보다 앞서는 재배치는 #4900에서 이미 리뷰된 축입니다. `rememberCursorThreadConversation`이 remint 경로에서만 쓰인다는 전제가 깨지면 우선순위가 다시 흔들립니다.

경로 CURSOR_ENVELOPE_ECHO_REMINT_MAX - 상한 3에 도달하면 대화를 그대로 두고 `midstream-envelope-echo-remint-exhausted`만 남깁니다. 계속 에코하는 모델은 그 시점부터 다시 눈덩이 위험이 있습니다.

경로 tool-guidance Write - 에코 루프와 직접 관련은 약하지만, grok이 Write를 찌르는 실측이 있으면 같은 PR에 넣는 게 맞습니다. 범위가 살짝 넓어 보이면 후속으로 빼도 됩니다.

경로 #4900 - 이 PR이 합류하면 dirty인 #4900은 `landed-via-maintainer`로 닫는 게 train 규칙과 맞습니다. #4874는 이미 closed라 Refs만으로 충분합니다.

메인테이너의 판단이 필요한 지점
- 이미 독이 든 대화를 복구하지 않는 제품 선택이 그대로 맞는지 (본문: 새 태스크 필요)
- 에코 remint 상한 3이 실측 grok-4.6 빈도 대비 충분한지, exhausted 이후 UX를 더 강하게 끊을지
- #4900을 이 PR merge 직후 자동 close할지, 수동으로 landed 코멘트만 남길지
- live Cursor 트래픽 없이 hosted CI만으로 합류할지 (본문이 주장하는 범위는 “다음 턴 리플레이·conversation 재사용 차단”까지)

너의 추천
exact-head CI(특히 cursor adapter/envelope 테스트가 들어 있는 test shard)가 초록이면 merge 후보입니다. #4900은 합류 후 `Landed via #4929 at <commit>` + `landed-via-maintainer`로 닫으세요. #4874는 이미 닫혀 있으니 다시 열지 마세요. live grok 재현은 merge 후 한 스레드만 smoke해도 충분합니다. CI가 막히면 이 변경과 macOS/Windows flake를 먼저 가른 뒤 합류하세요.

이 댓글은 grok-bot이 작성했습니다

PR #4947, opened 2026-09-17
fix(responses): enforce configured total sends during recovery

lidge-jun:
## Summary

Restore the configured total-send invariant found by the 2.59.0 merged-tree audit. A provider configured with attempts=1 authorizes one physical upstream send total. A later rebuild previously passed zero remaining attempts to the shared recovery helper, which could still grant a final-reserve send.

Configured exhaustion now stops rebuild work before response consumption or request mutation and returns the original upstream response. The recovery helper preserves its default reserve behavior for unconfigured providers. No retry count, timeout, delay or shared request cap is increased.

## Verification

- Local verification was NOT RUN because this lane forbids local tests, suites, typecheck, builds, installs and proxy execution. Hosted CI is the executable verification.
- Static source review, independent Sol review and git diff --check completed.
- Separate composed-budget regressions prove configured attempts=1 cannot draw reserve and the unconfigured default still allocates/refunds its fourth-send reserve.
- An end-to-end Console Go recovery case checks one physical send and the original 400 response body. Existing recovery regressions remain intact.
- First hosted run 35279024423 passed the physical-send and response assertions but found that the new no-recovery assertion expected an absent field. The existing request-attempt initializer creates an empty array. Head 50ecf0627347343a6eaa0c9f55aee25a6a906b13 now asserts that initialized empty list and has fresh CI run 35279717343; no production behavior or send-count assertion was weakened.
- On that corrected head, Linux 1/4–4/4, gates and the new send-cap cases passed. macOS 2/2 job 105398860800 failed only the separately reproduced combo connect-cancellation teardown hook at 30050.44 ms (12810 pass, 12 skip, 1 fail), matching the failure on the status-fixture-only PR #4948. The release host assigned that hook to another lane. This PR remains draft rather than claiming a fully green exact head; no test rerun or timeout change was used.
- Exact base: a0f611d4aceb9476d44268e43722273b7b211846. The release host will verify the new combined dev SHA after integration.

## Checklist

- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. Configured limits only narrow dispatch and cannot spend an extra recovery send.


lidge-jun:
Merging on the macOS exception now recorded in #4956, with the Windows evidence this change needed.

At its exact head all nine Windows shards are green in the dispatched `lane=all` run, along with the full Linux suite, `gates`, and the cross-platform smokes. The outstanding macOS legs are the long-standing defect in #4956, which was shown this evening to affect Windows as well and which has now been traced to spawned Bun child processes that are assigned a PID, write nothing to either stream, and never exit. A dedicated lane owns that investigation.

Holding this fix for that defect would delay a correctness fix for a problem it does not share and cannot influence. The release candidate remains held on #4956; this merge is not a promotion and makes no claim about macOS at this SHA.

PR #4948, opened 2026-09-17
test(cli): measure cold status setup before timed projections

lidge-jun:
## Summary

Separate the cold full-CLI status path from the version-projection assertions. Windows run 35277467396 killed the first status child after 15,587 ms; the next row took 13,074 ms and subsequent two-child rows took 1,518–1,741 ms. The first row was measuring cold setup plus behavior while later rows measured the warmed path. A later passing run does not resolve that defect.

One Windows beforeAll fixture now executes the same real status and health path before the matrix. Setup must succeed and is bounded by the existing 45-second spawn budget; its child has 25 seconds, leaving the existing 15-second Windows tree-removal bound plus five seconds for reap. All sixteen original JSON/human children retain their 15-second deadlines and result assertions. Cold setup and individual child timings are logged for hosted evidence. No retry, global budget increase or platform skip is added.

The observed cold-path dependence does not isolate Bun transpilation, Defender, filesystem caches or external diagnostic probes. This change makes that setup cost explicit and measured without claiming an unobserved subsystem cause.

## Verification

- Local verification was NOT RUN because this lane forbids local suites, focused tests, typecheck, builds, installs and proxy execution. Hosted CI is the executable verification.
- Original failure: run 35277467396, Windows 7/9 job 105391449887, cli-status-json.test.ts:77.
- Static review retained every original output assertion and both formats for all eight rows; child reap, server shutdown and temporary-home cleanup remain mandatory.
- Independent review required enough teardown margin; the setup child allowance was narrowed from 40 to 25 seconds before publication.
- Exact base: a0f611d4aceb9476d44268e43722273b7b211846. The release host will re-freeze and execute the combined tree after integration.
- Exact head cca06b1693fc4f7d4711bb920ea5f5417655edf5 has all nine Windows shards and all four Linux shards successful in full dispatch 35279223062, with gates green. Windows job 105397453194 recorded setup at 1755 ms and all sixteen measured children at 583–933 ms.
- The release owner explicitly authorized integration on that targeted evidence while macOS combo-failover teardown is investigated separately. PR run 35279139497 has one macOS shard-2 failure in the unrelated connect-cancellation hook (job 105396939652; 12807 pass, 12 skip, 1 fail). This is an attributed integration exception, not a claim that the full macOS workflow is green; remaining macOS jobs and the final merged-tree run are separate evidence.

## Checklist

- [x] Docs or release notes were updated when needed. The fixture documents where the real cold start is measured; the audit record tracks the hosted evidence.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. Only synthetic fixture values and elapsed times are logged; isolated paths and teardown are retained.


lidge-jun:
Merging this on a recorded exception, so the reasoning is in the record rather than in a thread.

The evidence this change needs is Windows, and it has it. At exact head `cca06b1693fc4f7d4711bb920ea5f5417655edf5`, dispatch run 35279223062 has all nine Windows shards green, Linux 1/4 through 4/4 and `gates` are green in both that run and the PR-event run 35279139497, and the sixteen timed projection assertions pass with the setup child at 1755ms and every timed child between 583ms and 933ms. That is the fixture doing exactly what it was changed to do: the cold cost moved into an explicit setup outside the measured window, and every assertion now measures the same warm path against the same unchanged 15s bound.

The one red is `macos 2/2`, and it is not this change. It is `server combo failover 030 activation matrix > connect cancellation wins with 499, no backup, warning, or cooldown` timing out in a `beforeEach`/`afterEach` hook at 30066ms, in a file this branch does not touch. A dedicated lane owns that investigation together with two longer silent hangs on sharded macOS legs at the same tip. Holding a Windows fix hostage to an unrelated macOS teardown defect would delay the release without making anything safer, and attributing that failure to this PR would be false.

Recording the boundary explicitly: this PR is merged with Windows and Linux evidence at its exact head, with a known unrelated macOS failure outstanding and owned elsewhere. It is not a claim that macOS is green at this SHA, and the release candidate is still held for exactly that reason.

Issue #4905, opened 2026-09-17
[Bug]: macOS control is cancelled near its 30-minute limit in full dispatch CI


luvs01:
### Client or integration

Other

GitHub Actions — the `macos-control` job (`macos control`) in `.github/workflows/ci.yml`, not an installed client or a provider request.

### Area

Platform (Windows / macOS / Linux)

### Summary

The unsharded macOS control job does not produce a completed suite result in the reviewed full `workflow_dispatch` run: it is cancelled approximately 30 minutes after starting while test output is still being produced. The workflow configures `timeout-minutes: 30` for this job. In the reviewed run, every other job succeeded, but the final `ci` aggregate failed.

This is a concrete CI reliability/evidence problem worth investigating. It is **not** a claim that every dispatch must fail, that normal pull-request runs execute this job, or that increasing the timeout alone is a proven fix. The record does not establish the suite's healthy full-run duration or rule out slow tests, process-lifecycle problems, or runner contention.

Expected behavior: the deliberately unsharded control should finish with a usable pass/fail result within an evidence-based, bounded budget. Genuine assertion failures and runtime crashes must remain failures; a green sharded run is not proof that the whole-pool control would pass.

### Reproduction

1. Inspect full dispatch [run 35217283983](https://github.com/luvs01/opencodex/actions/runs/35217283983), from `luvs01/opencodex` at commit `aa67f23ac9bceeef5728ba23ffaa71e3b84dec6b` on `agent/reauth-cancellation-upstream-20260914`. Its recorded event is `workflow_dispatch`.
2. Open [`macos control`, job 105188784487](https://github.com/luvs01/opencodex/actions/runs/35217283983/job/105188784487). Checkout, Bun setup, dependency installation, and GUI build succeeded. The Test step was cancelled and the subsequent CLI help smoke step was skipped.
3. The control invokes the entire test tree without a shard argument:

   ```sh
   bun test --isolate --timeout 60000 tests
   ```

4. Runner startup was `2026-09-17T11:55:44.8544970Z`; cancellation was `2026-09-17T12:26:01.9485140Z` (about 30 minutes 17 seconds). Test-file output continues immediately before cancellation, without a completed top-level suite result.
[...]

PR #4949, opened 2026-09-17
fix(oauth): drop legacy credential backup on destructive mutations

luvs01:
## Summary

- `auth.json.pre-multiauth` is a one-time downgrade-recovery copy of the whole legacy credential store, written on the first new-shape persist. Destructive mutations went through the same path: `removeCredential`/`removeAccount` left an existing backup behind, and a still-legacy store even created the backup during the logout itself — so after the user removed an account, a file holding the very refresh tokens they destroyed remained on disk.
- `mutateStore` gains a `removeLegacyBackup` option: it skips the one-time create and unlinks the backup after the store is persisted. `removeCredential` and `removeAccount` opt in.
- Removal is best-effort (`ENOENT` ignored, other failures warn) because it runs after persist — a failed unlink must not report a failed logout for an account that is already gone. A stale uninstall-manifest entry is harmless: `removeOwnedConfigState` skips missing paths, so this composes with the owned-cleanup registration from #4572.
- Automatic rollback paths (`replaceProviderAccountSet`, provider rekey) do not opt in — they are not user-initiated credential destruction.

## Verification

- Head: `8640e2039c4d28a52e5320d92d9a543a5e4575db` (tree `cc1951344b636766b638741b00ca592885f0c7f0`), based on dev `a0f611d4aceb9476d44268e43722273b7b211846`.
- `bun test tests/oauth/oauth-store-multi.test.ts` — 44 pass / 0 fail, including new regressions covering destructive-migration, pre-existing-backup, and stale-backup-on-migrated-store cases.
- `bun run structure:check` — passed (`structure/overview.md` updated to describe the removal).
- `bun scripts/file-size-ratchet.ts` — passed.

## Remaining gates

- Fork CI dispatched via `gh workflow run ci.yml`: https://github.com/luvs01/opencodex/actions/runs/35279061859. The macOS control job is expected to cancel at the 30-minute mark; that is the known upstream limitation tracked in #4905, not a signal from this change. Occasional shard flakes are likewise known upstream noise.
- This touches `src/oauth/`, which the intake gate may classify as a restricted surface requiring maintainer sponsorship.
- No `gui/` files are touched, so the UI-screenshot gate does not apply.

## Checklist

## Review readiness checklist


lidge-jun:
Reimplemented in #5598 on top of your three commits (authorship kept), with Co-authored-by on the repairs. Provider deletion is now covered too. Only the removed provider leaves the downgrade backup, so other providers keep recovery. The backup is never written through a symlink, and one an earlier install left unregistered stays unclaimed. Thank you, @luvs01. Closing in favor of #5598.

Issue #4944, opened 2026-09-17
opencode-go live rows silently materialize at the 128k context fallback: provider publishes no windows, generated registry is a generation behind, and augment never backfills existing live rows


tizerluo:
### Client or integration

Codex CLI

### Area

Catalog / models

### Summary

Every live-discovered `opencode-go` model row materializes at the 128k context-window fallback, because (a) the OpenCode Go `/v1/models` endpoint publishes no context-window field at all, (b) the generated model-metadata registry (`src/generated/model-metadata.ts`) is a generation behind the live roster, and (c) `augmentRoutedModelsWithMetadata` only **appends** registry rows that are absent from the live list — it never backfills a context window onto a row the live list already returned (`if (seen.has(key)) continue;` in `src/codex/catalog/routed-gather.ts`). The net effect is silent: a 1M-class model shows a plausible-looking 128k window, chained clients compact at ~121k, and nothing reports data loss.

Concretely, on 2.58.0 with OpenCode Go enabled, these live models exist at `https://opencode.ai/zen/go/v1/models` (response rows contain only `id`/`object`/`created`/`owned_by`) but are absent from the generated registry, so they fall back to 128000: `qwen3.8-flash`, `deepseek-flash` (the current name; legacy `deepseek-v4-flash` / `deepseek-v4-flash-vision-exp` still accepted by DeepSeek), `hy4-preview`, `omen-alpha`, `union-alpha`, `muse-spark-1.3-contributor`, plus `longcat-2.0`, `gpt-5.6-luna`, `qwen3.8-max`. Authoritative windows are published on models.dev (`limit.context`), which the generator (`scripts/generate-model-metadata.ts`) already resembles: qwen3.8-flash 1,000,000 / deepseek-v4(.1)-flash 1,000,000 / hy4-preview 1,024,000 / omen-alpha 500,000 / union-alpha 262,144 / muse-spark-1.3-contributor 1,048,576.

Expected: registry rows should backfill `contextWindow` onto live rows whose provider-published window is absent (the `seen` check should merge metadata into the existing row instead of skipping it), and the generated registry should catch up with the current live roster. Either fix alone removes the silent 128k fallback for most rows.

### Reproduction

1. `ocx provider add opencode-go` with a valid key, enable e.g. `qwen3.8-flash` (a 1M model), restart the proxy.
[...]

Issue #82, opened 2026-07-10
Opencode-Go hy3-preview


Wibias:
It seems that opencode-go doesnt have hy3-preview, but its still selectable, giving this error:

Provider error 400: {"error":{"message":"Error from provider (Console Go): Model hy3-preview is not supported on the lite model list. Use GET /inference/go/openai/v1/models to list available models.","type":"invalid_request_error","param":"model","code":"model_not_supported"}}

lidge-jun:
Thanks — this is already fixed on current main. `hy3-preview` was removed from the Zen Go lite model list upstream, but opencodex's opencode-go catalog augmentation (which appends known models missing from live `/models` discovery) was still carrying it from the generated metadata bundle. Commit 8d69372d refreshed that bundle and dropped `opencode-go/hy3-preview`, so it is no longer selectable as of v2.7.4.

I've also added a negative regression test in 8942c7c3 so the augmentation path can never resurrect it (tests/codex-catalog.test.ts). If you're on an older version, `ocx update` / reinstall to v2.7.4+ clears it.

lidge-jun:
Correction to my earlier reply: after rechecking the live endpoint, the metadata-only removal in v2.7.4 was not sufficient. `https://opencode.ai/zen/go/v1/models` still advertises `hy3-preview`, so successful live discovery could add it back even though Console Go rejects it as outside the lite model list.

The complete fix is now on `dev` in [4ccc6da6](https://github.com/lidge-jun/opencodex/commit/4ccc6da687dc50ba3782cd6cef6f23d89918856e). opencodex now filters the exact `opencode-go/hy3-preview` pair after live/static/metadata aggregation and also removes a stale copy when an on-disk catalog is preserved during an empty refresh. Other OpenCode Go live models (including newly discovered ids) and `hy3-preview` under unrelated providers remain unaffected.

Regression coverage exercises both paths: a live `/models` response that contains HY3, and a pre-existing on-disk HY3 row during empty-refresh preservation. Verification: affected catalog tests 68/68, TypeScript typecheck, and full suite 1,943/1,943 passed.

This complete fix is not in v2.7.4; it will be included in the next release. Closing as fixed on `dev`.

Issue #4956, opened 2026-09-17
[Bug]: spawned Bun child processes stop producing output and never exit, on both macOS and Windows CI legs


lidge-jun:
### Client or integration

CI only (GitHub Actions hosted macOS runners); not a client-facing path.

### Area

CI and release

### Summary

Sharded macOS test legs stop producing output mid-suite and are killed at their job budget with zero reported failures. Across four occurrences the signature is identical: a test logs PASS, then the job emits nothing at all for 16 to 18 minutes until the runner cancels it, and Bun and tee are orphan-killed. This has been present since at least the 2.58.0 release round and was recorded there as capacity cancellation, which the evidence below does not support. A related but distinct failure mode appears where a hook timeout exists to stop it: the same shard reports `a beforeEach/afterEach hook timed out for this test` at just over 30s instead of going silent.

Two consequences. Releases have been cut without complete macOS evidence while believing macOS reds were infrastructure. And because the silent form produces a cancellation rather than a failure, it is indistinguishable at a glance from the genuine capacity cancellations that also occur, so it has survived several rounds of triage.

`macos control` hitting its 30-minute budget is a separate problem with a different cause and is tracked in #4905. Do not merge the two.

### Reproduction

Occurs on hosted macOS runners during the sharded platform-macos legs. It is not reproducible locally and no local verification was attempted for this report.

Silent form, four occurrences:

1. Run 35277719805, job 105392401547, macos 1/2, at dev a0f611d4ac. Last output 21:52:22, a PASS in `tests/codex-integration/codex-inject-write-lock.test.ts` case `a held lock makes real injection report busy and write nothing`. Silence for 16m15s until cancellation at 22:08:37.
2. PR #4945, macos 2/2. Last PASS `full ordering ignores empty entries and accepts raw upstream ids with slashes` at 21:53:21.617. Silence for 17m33s until cancellation at 22:10:54.
3. Run 35247708168 attempt 2, job 105303640713, macos 2, during the 2.58.0 round. 2514 pass / 0 fail, then 17m32.9s of silence after a catalog picker test until the 20-minute cancellation.
4. Run 35254182109 attempt 1, job 105313855869, macos 2. 2512 pass / 0 fail, then 17m19.1s of silence after `full ordering ignores empty entries and accepts raw upstream ids with slashes` — the same locus as occurrence 2.

Hook-timeout form, two occurrences on unrelated branches:

[...]

PR #4930, opened 2026-09-17
fix: pin Union Alpha wire and namespace Go sessions

lidge-jun:
## Summary

- Hard-pin OpenCode Go's exact `union-alpha` model id to the Anthropic wire, based on the evidence in #4847, while leaving sibling Chat and Responses models unchanged.
- Namespace generated `x-opencode-session` values by the final wire protocol with a versioned hash separator. Sessions remain stable for the same lane and protocol, and operator-supplied headers still win unchanged.
- Protect destination recognition from wire-selection order: each request retains the original routed provider for the registry destination check, settles the final wire separately, and passes both values to the transport helper. The helper therefore hashes the settled adapter without asking the adapter-sensitive registry matcher to recognize the already-flipped provider.
- The direct Messages reproduction still returned 503 `Endpoint is unavailable` with a session header. This change does not establish successful Union Alpha inference; it fixes protocol selection and session namespacing only.

Refs #4847

## Verification

- `git diff --check`
- Static call-graph inspection of `src/server/responses/core-normalize.ts`, `src/server/chat-completions.ts`, `src/providers/opencode-go-transport.ts`, and `src/server/adapter-resolve.ts`.
- Independent SHA-256 vector calculation for Chat and Responses session namespaces, then comparison with the fixed expectations in `tests/providers/opencode-go-session-header.test.ts`.
- Confirmed every `resolveOpenCodeGoTransport` call supplies the original routed destination separately from the settled provider.
- Local Bun tests, typecheck, builds, dependency installation, and the `ocx` binary were not run because this lane explicitly forbids local verification. Hosted CI remains the executable verification.

## Checklist

- [x] Scope stays focused and avoids unrelated cleanup.
- [x] Docs or release notes were updated when needed.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults.

## Summary by CodeRabbit

- **New Features**
  - Added protocol-specific session affinity for OpenCode Go, keeping Chat, Responses, and Anthropic sessions distinct.
  - Ensured `union-alpha` consistently uses the Anthropic-compatible wire protocol.
  - Preserved OpenCode Go destination recognition when providers use adapter overrides.
  - Maintained operator-supplied session header precedence.

- **Bug Fixes**
  - Fixed routing and session handling for per-model wire protocol overrides.
  - Improved compatibility for `union-alpha` requests across supported entry points.

- **Documentation**
  - Updated transport documentation to describe protocol-specific session behavior and routing.

[...]

PR #4962, opened 2026-09-17
fix(catalog): publish opencode-go context windows so live rows stop falling back to 128k

lidge-jun:
## Summary

Every live-discovered `opencode-go` row for a model the generated registry did not know
materialized at the 128k routed fallback, so a 1M-class model advertised 128k and chained
clients compacted at roughly 121k with nothing reporting the loss. Enabling
`opencode-go/qwen3.8-flash` and reading the routed catalog produced
`"context_window": 128000`; it now produces `1000000`.

This refreshes the vendored metadata snapshot for the whole provider. **The `opencode-go`
bundle is now complete as far as models.dev publishes**: every id the authoritative source
lists for this provider is either present in the snapshot or deliberately excluded for a
reason recorded below. There is no remaining tail of un-copied numbers.

**The operative cause is (b), the generated registry being a generation behind the live
roster.** The decisive step is `applyCatalogMetadata`
(`src/codex/catalog/parsing.ts:828-836`), which writes `context_window`,
`max_context_window` and `auto_compact_token_limit` straight onto the serialized entry from
the generated table, keyed by provider and model id, without consulting the `CatalogModel`
at all. It runs for every routed row on both derive paths
(`src/codex/catalog/derive-entry.ts:171` and `:217`), and its only escape is its own
`if (!meta) return;` guard at line 830. A model id absent from the generated table therefore
gets no window written, and `ensureStrictCatalogFields`
(`src/codex/catalog/parsing.ts:589`) stamps 128000 — the only place a routed row is assigned
128000 in the catalog path.

**(c) `augmentRoutedModelsWithMetadata` is ruled out as the cause of the reported symptom.**
The skip at `src/codex/catalog/routed-gather.ts:869` (`if (seen.has(key)) continue;`) is real
and does exactly what the issue describes, but repairing it would have fixed none of the
reported ids. Of the ids named, only `deepseek-v4-flash` was present in the generated
`opencode-go` table (at 1000000), and for it `applyCatalogMetadata` already wrote the correct
window — `applyCatalogModelMetadata` (`src/codex/catalog/effort.ts:116-132`) does not clobber
it, because `resolvedContext` is undefined when the row carries no live window and no cap. For
every other id, `listModelMetadata("opencode-go")` returned no row, so a backfill loop would
have had nothing to backfill. The skip is left untouched here. It remains a latent defect for
consumers that read `CatalogModel.contextWindow` rather than the serialized entry, and it
deserves its own issue rather than being folded into a data refresh.

**(a) is a true upstream fact and the necessary trigger, but not a repository defect.** The
provider genuinely publishes no context field. The codebase already carries two independent
compensations for exactly that situation and both work: the provider registry preset
`modelContextWindows` (`src/providers/registry/entries-core.ts:672`, seeded into the gather at
`src/codex/catalog/gather-capture.ts:347` through `applyRegistryCapabilitySeedFill`, applied in
`src/codex/catalog/model-hints.ts:123-133`), and the generated table. Only the second one was
stale.

**The issue's model list is partially wrong on one id.**
`muse-spark-1.3-contributor` is already declared at `1_048_576` in
`src/providers/registry/entries-core.ts:684`, and
`tests/providers/opencode-go-muse-context.test.ts` pins that declaration end to end. It
therefore already resolved to 1M on `dev` through the registry seed and was never part of this
defect. It is included in the snapshot refresh anyway so both compensations agree on the same
value rather than one silently covering for the other.

### What changed

The fix is data, not merge logic. `scripts/model-metadata.source.json` gains eleven
`opencode-go` records and `src/generated/model-metadata.ts` is regenerated to match. Every
window is `limit.context` as published by models.dev, which the snapshot mirrors:

| id | context window | max tokens |
| --- | --- | --- |
| `qwen3.8-flash` | 1,000,000 | 131,072 |
| `qwen3.8-max` | 1,000,000 | 131,072 |
| `hy4-preview` | 1,024,000 | 64,000 |
| `omen-alpha` | 500,000 | 128,000 |
| `union-alpha` | 262,144 | 131,072 |
| `longcat-2.0` | 1,000,000 | 131,072 |
| `gpt-5.6-luna` | 1,050,000 | 128,000 |
| `muse-spark-1.3-contributor` | 1,048,576 | 131,072 |
| `glm-5.3-flash` | 1,000,000 | 131,072 |
| `muse-spark-1.2-contributor` | 1,048,576 | 131,072 |
| `ox-alpha-free` | 1,000,000 | 131,072 |

`hy4-preview` is a different id from the pinned-out `opencode-go/hy3-preview`, which stays
excluded by the generator's `EXCLUDED_MODELS` guard and is still asserted absent.

### Two ids models.dev publishes that are deliberately not added

`deepseek-v4.1-flash` and `deepseek-v4-flash-vision-exp` are the only `opencode-go` ids
models.dev lists that this snapshot does not carry, and the omission is a decision rather than
an oversight. models.dev publishes 1,000,000 for both. This repository already declares
`1_048_576` for each on this provider in `src/providers/registry/entries-core.ts`, citing
DeepSeek's official Codex setup script and describing the 1,000,000 figure as a rounded
approximation.

Copying the lower number in would fix nothing and could regress. `applyCatalogModelMetadata`
runs after `applyCatalogMetadata`, so the registry hint already wins wherever it applies —
these two ids were never broken. But on any row that receives no provider hints, the generated
table is what answers, and a 1,000,000 row would quietly advertise 48,576 fewer tokens than the
value the repository deliberately chose.

`tests/codex-integration/catalog-opencode-go-context-window.test.ts` encodes this as an
invariant rather than as an absence: a future refresh may add these ids, but only carrying a
window at least as large as the registry already promises. A blind copy from models.dev now
fails a test instead of silently lowering a user's context.

### Two deliberate limits on the record shape

- **No `cost` block.** The generated bundle cost is consulted before the expected-price
  overlay in `resolveMatchedPriceExact` (`src/usage/cost.ts`), so carrying models.dev prices
  in would have introduced a second source of truth for ids this repository already prices
  deliberately (`src/usage/expected-prices.ts:316` prices `opencode-go/qwen3.8-flash` as a
  verified-derived vendor estimate, and models.dev disagrees on the input rate). Omitting cost
  leaves `bundled?.cost` undefined, so every pricing path resolves byte-identically to today.
  One consequence worth flagging for a follow-up: the comment above test 17h in
  `tests/usage/usage-cost.test.ts` says these ids have "no jawcode bundle row", which is now
  narrower than the truth — they have a row with no cost. The assertions are unaffected.
- **No `thinking` block.** Its `minLevel`/`maxLevel` vocabulary does not map cleanly onto the
  models.dev `reasoning_options` ladders for these ids, the generator never reads the field,
  and the effort ladders these models actually expose come from the provider registry. Writing
  a guess into a vendored data file would have been worse than omitting it. `api` and `baseUrl`
  are included and follow this repository's own registry: `openai-responses` for the ids
  `entries-core.ts` declares in `modelWireDefaults` (`gpt-5.6-luna`,
  `muse-spark-1.3-contributor`, `muse-spark-1.2-contributor`), and the preset's `openai-chat`
  default at `https://opencode.ai/zen/go/v1` for the rest.

### How the generated file was produced

`src/generated/model-metadata.ts` was **not** hand-edited freehand. Its banner and
`tests/codex-integration/model-metadata-sync.test.ts` both forbid drift from the snapshot, so
the regenerated `opencode-go` line was produced by replaying the generator's exact emit rules —
`allowedProviders`, the `EXCLUDED_MODELS` guard, `localeCompare` row ordering, `compactRow`
trailing-undefined trimming, and `JSON.stringify` formatting.

The replay was validated against the committed file before each of the two passes, and both
times it reproduced 14 of the 16 provider lines byte-identically, including `opencode-go`. The
only two misses are `minimax` and `openrouter`, the bundles where `localeCompare` diverges from
code-point order; both are untouched by this change. The second validation is the stronger one:
by then the committed `opencode-go` line already contained the first pass's eight compacted
five-element rows, so reproducing it byte-for-byte exercised the trailing-undefined trimming as
well as the ordering. Row placement under `localeCompare` was checked again after the second
pass. Each commit changes exactly one line of the generated file.

### Structure docs

No `structure/` update is required, and this was checked rather than assumed.
`structure/INDEX.md:109` maps `src/generated/` to `structure/runtime.md`, which already states
that `src/generated/` is committed build output not edited by hand — still true, and this
change does not alter it. `bun run structure:check` fails on a doc naming a path the tree no
longer has, an invariant whose test is gone, or a new `src/` area nobody claimed. This change
removes no path, removes no invariant test, and adds no `src/` area, so it trips none of the
three. No `src/codex/` source changed either, so `structure/catalog.md` is untouched.

## Verification

Local verification was not run: this lane forbids running the local suite, typecheck, build,
install, or the `ocx` binary. Hosted CI is the executable verification for this change.

The correctness argument is static, over the source:

- A routed `CatalogModel` with no `contextWindow` and no `contextCap` reaches
  `applyCatalogMetadata`, which now finds a row and writes `context_window`,
  `max_context_window` and `auto_compact_token_limit = floor(window * 0.9)`.
- `applyCatalogModelMetadata` runs next and cannot clobber it: `resolvedContext` is undefined
  with no live window and no cap, and its `else if` branch only touches
  `auto_compact_token_limit`, and only when `model.maxInputTokens` is set.
- `ensureStrictCatalogFields` then leaves the window alone, because `context_window` is a
  positive number and `max_context_window` equals it rather than exceeding it.
- `ensureAutoCompactTokenLimit` only fills when the field is absent, so the value written above
  survives.

Hosted jobs and tests that exercise this change:

- `tests/codex-integration/model-metadata-sync.test.ts` — the decisive one. It reruns the real
  generator into a temp path via `MODEL_METADATA_OUT` and byte-compares against the committed
  file, so it fails if the hand-produced output differs from the generator's by a single byte.
- `tests/codex-integration/catalog-opencode-go-context-window.test.ts` — new regression added
  here; asserts the generated registry answers for each of the eleven ids, that a discovered row
  with no published window serializes with the registry window rather than 128000, that the
  derived catalog entry carries it end to end, and that the generated table never undercuts the
  registry's declared DeepSeek windows.
- `tests/codex-integration/codex-catalog.test.ts` — existing `opencode-go` augmentation cases,
  including the issue #82 assertion that `opencode-go/hy3-preview` stays absent, the
  high-risk-model window table, and the DeepSeek V4.1 live-row case that resolves through the
  registry hint.
- `tests/usage/usage-cost.test.ts` — the `opencode-go` pricing rows (test 17h), which must stay
  green given the deliberate absence of a `cost` block.
- `tests/test-layout.test.ts` and `tests/test-layout-tooling.test.ts` — the new test file is
  registered in both `scripts/test-layout/layout.json` and
  `tests/fixtures/test-layout-expected.json`.
- `tests/ci-workflows/file-size-ratchet.test.ts` — the new test file is 118 lines, well under
  the 2000-line threshold and absent from the baseline. The regression was deliberately put in
  a new file rather than appended to `codex-catalog.test.ts`, which sits exactly at its
  7985-line baseline cap and can never be raised.
- `tests/ci-workflows/structure-ssot.test.ts` (`bun run structure:check`), `bun run typecheck`,
  and `bun run privacy:scan` on Linux, Windows and macOS.

## Checklist

- [x] Docs or release notes were updated when needed. No user-facing behavior contract changed:
      this restores the windows the catalog was always meant to advertise, so `docs-site/` needs
      no edit, and `structure/` ownership was verified above.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. The
      change touches a vendored capability snapshot and its generated output only; no
      credential, auth, workflow, or release path is involved, and the added records contain no
      secrets.

Closes #4944


lidge-jun:
## 리뷰 · 우선순위 75 / 80

설명

이 PR은 OpenCode Go를 켠 뒤 카탈로그에 올라온 실제 모델 줄이, 진짜로는 100만 토큰짜리인데도 조용히 12만 8천(128000)으로 찍히는 버그를 고칩니다. 지금 `dev` HEAD는 `61ee64747`이고, 카탈로그 쪽 동작은 이미 이렇게 정해져 있습니다. 라이브 `/v1/models` 응답에는 창 크기 필드가 없고, 직렬화 직전에 `applyCatalogMetadata`(`src/codex/catalog/parsing.ts` 828줄 근처)가 생성된 표(`src/generated/model-metadata.ts`)에서 provider+model id로 창을 찾아 씁니다. 표에 없으면 바로 return 하고, 그다음 `ensureStrictCatalogFields`(`parsing.ts` 589줄)가 라우티드 기본값 128000을 찍습니다. 그래서 클라이언트가 대략 12만 1천에서 압축을 돌리는데, 어디에서도 “창을 잃었다”고 알려 주지 않습니다.

고치는 방법은 병합 로직이 아니라 데이터입니다. `scripts/model-metadata.source.json`의 `opencode-go`에 models.dev가 공개한 여덟 개 id를 넣고, 생성 규칙을 그대로 따라 `src/generated/model-metadata.ts` 한 줄을 다시 만들었습니다. 넣는 창은 models.dev `limit.context`입니다. `qwen3.8-flash`/`qwen3.8-max`/`longcat-2.0`은 1000000, `hy4-preview`는 1024000, `omen-alpha`는 500000, `union-alpha`는 262144, `gpt-5.6-luna`는 1050000, `muse-spark-1.3-contributor`는 1048576입니다. `cost` 블록은 일부러 빼서, 이미 `dev`에 올라온 `#4898` EXPECTED_PRICE_OVERLAYS(`src/usage/expected-prices.ts` 315줄 근처)와 가격이 두  Soft가 되지 않게 했습니다. `thinking`도 넣지 않았습니다. 노력 사다리는 레지스트리 쪽이 맡고 있고, 생성기는 그 필드를 읽지 않습니다.

이슈 `#4944`가 말한 (c) 보강 루프 스킵은 이번에 손대지 않았습니다. `augmentRoutedModelsWithMetadata`의 `if (seen.has(key)) continue;`(`src/codex/catalog/routed-gather.ts` 869줄)는 실제로 있지만, 보고된 id 대부분은 생성 표 자체에 없어서 스킵을 고쳐도 창이 채워지지 않습니다. `muse-spark-1.3-contributor`는 이미 `entries-core.ts`의 `modelContextWindows`에 1048576이 있어서, 시드 경로로는 `dev`에서도 1M이 나옵니다. 그래도 생성 표에 같이 넣어 두 보상이 같은 숫자를 말하게 한 선택은 맞습니다. `deepseek-flash`라는 id는 models.dev에 `opencode-go` 행이 없어서 만들지 않았고, DeepSeek 표면은 `deepseek-v4-flash`(이미 표에 있음)와 `deepseek-v4.1-flash`(레지스트리 시드)로 다룹니다. 회귀 테스트는 새 파일 `tests/codex-integration/catalog-opencode-go-context-window.test.ts`에 두었습니다. `codex-catalog.test.ts`는 7985줄 상한이라 붙이면 안 됩니다. 레이아웃 두 파일에도 등록했고, `model-metadata-sync`가 생성물 바이트를 검사합니다. 로컬 스위트는 돌리지 않았고 CI가 검증 창입니다.

라인 문제

`src/codex/catalog/routed-gather.ts` 869 - `seen.has(key)`면 그대로 건너뛰어서, 직렬화 엔트리가 아니라 `CatalogModel.contextWindow`만 읽는 소비자는 여전히 빈 창을 봅니다. 이번 데이터 고침으로는 그 경로가 안 고쳐집니다.

`tests/usage/usage-cost.test.ts` 261 근처 - 주석이 “OpenCode Go serves five ids with no jawcode bundle row”라고 하는데, 이제 표에는 행이 있고 cost만 없습니다. 테스트 결과는 그대로여도 설명이 한 세대 뒤입니다.

이슈 본문의 `deepseek-flash` id - models.dev에 없어서 행을 안 만든 판단은 맞지만, 게이트웨이가 그 이름으로만 올리면 그 id는 계속 128000으로 떨어집니다. 별도 확인이나 이슈로 남겨야 합니다.

`scripts/model-metadata.source.json`의 `union-alpha` api 필드 - 스냅샷에는 `openai-completions`로 적혀 있는데, `dev`의 `#4930`은 union-alpha를 Anthropic 와이어로 고정했습니다. 생성 Row에는 api가 안 실리므로 이번 창 고침에는 영향이 없지만, 스냅샷과 레지스트리가 서로 다른 말을 합니다.

메인테이너의 판단이 필요한 지점

- `routed-gather`의 append-only 스킵을 후속 이슈로 열지, 이번 PR 범위 밖으로 두고 Closes `#4944`만 할지.
- 이슈 댓글에 나온 poineer/pioneer 대규모 128k 패턴을 같은 세대의 후속으로 받을지, 이번과 분리할지.
- 게이트웨이에 `deepseek-flash` 같은 별칭이 실제로 사는지 확인한 뒤, 없으면 이슈 문구만 정리할지.
- CI(`model-metadata-sync` 포함)가 초록이 되기 전에 머지할지, 초록 확인 후 머지할지.

너의 추천

CI(특히 `tests/codex-integration/model-metadata-sync.test.ts`와 새 `catalog-opencode-go-context-window.test.ts`)가 초록이면 머지하세요. 코드 경로를 바꾸지 않는 데이터 고침이고, 원인 분석과 범위 제한이 `dev`의 `applyCatalogMetadata` → `ensureStrictCatalogFields` 흐름과 맞습니다. 머지 전에 usage-cost 17h 주석만 “행은 있고 cost만 없음”으로 고치거나 후속 한 줄로 잡아 두세요. `routed-gather` 스킵과 pioneer 확장은 이 PR에 섞지 말고 새 이슈로 분리하는 편이 낫습니다. Closes `#4944`는 유지해도 됩니다.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:

PR #4963, opened 2026-09-17
feat(doctor): warn when the Codex default model is not exposed by the proxy

lidge-jun:
## Summary

Closes #4646.

The issue makes three asks. One of them is a real defect; the other two describe
behavior that is working as designed, and this PR says so explicitly rather than
implementing them.

**Ask 3 (no warning when Codex's pinned default model is not exposed) was the real
defect, and is fixed.** Codex starts every session on the root `model` pin in its own
`config.toml`, and nothing compared that pin against the models this install exposes.
Neither surface checked it: `/api/startup-health` returns `readStartupHealth` verbatim
and its whole input set is restart-survivability facts, and none of the sections in
`runDoctor` read a model id. When the pin and the exposed set disagree, every turn fails
and no diagnostic points at the pin.

`ocx doctor` now prints a **Codex default model exposure** section immediately after
**Codex config compatibility**, since both read `config.toml`. Three design points:

- The exposed set is **read, not recomputed**. The live assembly needs an entitlements
  snapshot, a provider gather and account-selector expansion; duplicating it in the CLI is
  exactly the drift `formatStartupRoutingDetail` and `computeVersionSkew` were extracted
  to prevent. So doctor reads `GET /v1/models` from the proxy it has already resolved via
  `findLiveProxy`, and falls back to the on-disk catalog's `visibility: "list"` slugs.
- It reports **three states**: exposed, not exposed, and undeterminable. It never claims
  "not exposed" when it could not read the set — the install least able to answer (proxy
  down, catalog never synced) is exactly the one a fabricated failure would mislead. Both
  surfaces are consulted before any negative verdict, so an encoding or staleness
  difference between them cannot be reported as a broken pin.
- It stays a **warning**: no `recordDoctorFailure()`, no `process.exitCode` write. FAIL is
  reserved for an unusable surface; a quota-exhausted or disabled pin is degraded, not
  unusable, and must not break a legitimately green pipeline.

`StartupHealth` is deliberately not extended. `deriveStartupHealth` is a pure function over
machine-protection inputs, the GUI narrows its status to a three-colour dot, and its result
is cached by a subprocess probe whose cache key knows nothing about `disabledModels` — a
model fact cached on service-diagnostic freshness would go stale on the next quota change.
This PR takes the smaller change: doctor only, no new settings field.

**Ask 1 (sync skips rewriting an existing catalog because the file exists) is false, and
implementing it literally would be harmful.** There is no existence test in the write
decision. Sync regenerates the whole catalog from live state and byte-compares against
disk: `src/codex/catalog/retained-sync.ts` builds the content and returns
`catalogWritten: false` only when the freshly computed bytes equal the bytes on disk —
strictly stronger than the slug-set or hash comparison the issue asks for. The skip is
load-bearing, not an oversight: a no-op rewrite bumps mtime, the app-server staleness
classifier (#857) compares that mtime against each running Codex's start time, and since
#1407 a stale verdict silences opencodex's model guidance for that Codex's entire
lifetime. It landed on 2026-08-11 in `c7eec01ca4` / `642805c11e`, a month before the
report, and is already pinned by `tests/codex-integration/codex-catalog-sync-hardening.test.ts`.

Implementing the requested drift trigger would re-break both issues permanently.
`visibleNativeSlugs` filters disabled slugs out of `/v1/models` while the catalog
intentionally retains them as `visibility: "hide"` rows, so the two sets are unequal **by
design** and can never converge. A rewrite triggered on that difference would rewrite the
catalog on every single sync, forever.

**Ask 2 (disabled native slugs persist as `visibility: "hide"`) is true, and is the
documented, deliberate contract.** `desktopAllowlistSuppressedNativeSlugs` returns an empty
set when no native-alias combo is configured, and the visibility assignment keeps the row.
Retention preserves real upstream metadata so re-enabling restores it instead of
synthesizing a guess, and several code paths depend on that recovery property, so the
emission default is unchanged. Most of the contract was already documented; what was
missing was the operator-facing consequence, now added to the English
`codex-app-models.md` guide and `structure/catalog.md`: a disabled-but-hidden native slug
can still be shown and picked in Codex Desktop, selecting it is **not** refused for being
disabled (`disabledModels` is a catalog control and `src/router.ts` never consults it), and
a `nativeAlias` combo is the lever that omits the row unconditionally.

**What could not be verified statically.** The reporter's captured body shows
`catalogWritten: false` but not the `refreshOutcome` / `skippedReason` field that would
distinguish outcomes, so it is unknown whether their runs reached the byte-compare at all
or landed on an earlier discriminated refusal: `desired_disabled` re-read under the write
lock, a `null` from `revalidateRetainedCatalogSync`, a stale entitlement snapshot, or a
`readRetainedCatalogSync` preflight refusal. All of those also report
`catalogWritten: false`. The field that separates them is `refreshOutcome`:
`"committed"` with `catalogWritten: false` is the byte-identical skip described above, while
`"refused"` means the run never reached the comparison. If the reporter can supply the full
response body, that distinguishes them.

### Files

- `src/codex/catalog/parsing.ts` — `readConfiguredDefaultModel()`, beside the existing
  `readConfiguredAutoReviewModel()` precedent, same shape and same swallow-and-return-null
  error policy.
- `src/cli/doctor.ts` — `collectDefaultModelExposure()` above `runDoctor` per this file's
  helper convention, the new section, and the hint entry.
- `structure/runtime.md`, `structure/catalog.md` — source-ownership obligation for
  `src/cli/` and `src/codex/`.
- `docs-site/src/content/docs/guides/codex-app-models.md` — the ask-2 operator consequence.
- `tests/codex-integration/doctor.test.ts`, `tests/codex-integration/native-model-toggle.test.ts`
  — regressions in existing, already-mapped files (no test-layout bookkeeping).

## Verification

Local verification was not run: this lane forbids running the local suite, typecheck,
build, install, or the `ocx` binary. Hosted CI is the executable verification for this
change.

Hosted CI jobs and tests that exercise it:

- `bun run test` on Linux, Windows and macOS — specifically
  `tests/codex-integration/doctor.test.ts` (six new collect-style cases: no pin, exposed via
  proxy, not exposed with the surfaces named, undeterminable, proxy-401 falling back to the
  catalog, and a retained `hide` row not counting as exposure) and
  `tests/codex-integration/native-model-toggle.test.ts` (the ask-2 contract: hidden-but-retained
  without an alias, omitted with one).
- `tests/codex-integration/codex-catalog-sync-hardening.test.ts` — the pre-existing #857/#1407
  byte-compare regression, unchanged by this PR and the reason ask 1 is not implemented.
- `bun run typecheck` (strict) on all three platforms.
- `bun run structure:check` via `tests/ci-workflows/structure-ssot.test.ts` — the
  `structure/` doc-map and ownership gate for the `src/cli/` and `src/codex/` edits.
- `tests/ci-workflows/file-size-ratchet.test.ts` — no touched file is in the baseline and
  none approaches the 2000-line threshold.
- `bun run privacy:scan` — the new code logs a model id and no credential; it sends no
  Authorization header and reads no token.
- `tests/cli/cli-dispatch.test.ts` and `tests/cli/cli-json-contract.test.ts` — the doctor
  dispatch and exit-code contract, which this change deliberately does not alter.

Static correctness argument, since none of the above ran locally:

- `readConfiguredDefaultModel` uses `readRootTomlString(toml, "model")`, whose regex anchors
  `^\s*model\s*=`, so `model_provider = "..."` cannot match it, and which stops at the first
  table header, so a `[profiles.*]` override is not read as the root pin.
- `defaultCatalogModels()` returns `RawEntry[]` (`Record<string, unknown>`), which is exactly
  the injected row type, so the catalog path needs no cast.
- Every new doctor test injects all four dependencies, so the cases touch neither the real
  `CODEX_HOME` nor the network. The existing `runDoctor` tests write no root `model`, so the
  new section returns `not_configured` before any fetch is attempted.
- `src/cli/doctor.ts` is dynamically imported only by the `doctor` dispatch branch, so the
  new import adds no cost to other commands.

Not done, deliberately: translated locales of `codex-app-models.md` (ko, ja, zh-cn, zh-tw,
fr, ru, tr) are untouched. The change adds a new English subsection rather than altering an
existing claim, so no translation now contradicts the English source; they are simply less
complete, which is the normal state between translation passes.

## Checklist

On the third box: the new code performs one unauthenticated loopback `GET /v1/models`
against the proxy doctor already resolved. It sends no `Authorization`, `x-api-key` or
`x-opencodex-api-key` header and reads no token file, so it adds no credential surface. On a
non-loopback bind `/v1/models` requires data-plane admission, which doctor deliberately does
not hold; that 401 is handled as "this surface did not answer" and the on-disk catalog
answers instead. The request is bounded by an 8s `AbortSignal.timeout`, and every failure
mode degrades to a verdict rather than throwing out of the diagnostic. No `.github/` file is
touched.

## Summary by CodeRabbit

* **New Features**
  * Added an `ocx doctor` check that reports whether Codex’s configured default model is exposed, unavailable, or cannot be determined, with guidance when action is needed.
* **Documentation**
  * Documented behavior for disabled native models: they are hidden from model listings and dashboards while remaining routable in supported cases.
  * Clarified that models with native aliases are removed entirely.
  * Added runtime documentation for read-only diagnostics and model exposure checks.


lidge-jun:
## 리뷰 · 우선순위 74 / 80

이 PR은 이슈 #4646의 세 가지 요청을 나눠서 다룹니다. 지금 `dev` HEAD `61ee64747` (팁 #4948, 패키지 2.59.0) 기준으로 보면, 진짜 구멍은 세 번째입니다. Codex는 자기 `config.toml` 루트의 `model` 핀으로 세션을 시작하는데, opencodex 쪽은 그 핀이 지금 설치가 노출하는 모델 집합에 들어 있는지 아무도 안 봅니다. `/api/startup-health`의 `deriveStartupHealth`는 재시작·보호 입력만 보고, `runDoctor`의 기존 섹션도 모델 id를 읽지 않습니다. 그래서 쿼터 소진이나 `disabledModels`로 핀이 빠진 뒤에도 진단이 조용하고, 매 턴만 깨집니다. 이 PR은 그 구멍을 `ocx doctor`의 **Codex default model exposure** 섹션으로 막습니다. `src/codex/catalog/parsing.ts`에 `readConfiguredAutoReviewModel`과 같은 모양의 `readConfiguredDefaultModel`을 두고, `src/cli/doctor.ts`에 `collectDefaultModelExposure`를 추가합니다. 노출 집합은 CLI에서 다시 조립하지 않고, 이미 `findLiveProxy`로 찾은 프록시의 `GET /v1/models`를 읽고, 그게 안 되면 디스크 카탈로그의 `visibility: "list"` 슬러그만 봅니다. 판정은 노출됨 / 미노출 / 판정 불가 세 가지이고, 미노출만 `!!`와 힌트에 넣으며 `recordDoctorFailure()`나 exit code는 건드리지 않습니다. `StartupHealth`는 일부러 안 늘립니다. 그 결과는 서비스 진단 신선도로 캐시되고 `disabledModels`를 키에 안 넣기 때문에, 모델 사실을 넣으면 쿼터 변화에 바로 낡습니다.

첫 번째 요청(기존 카탈로그 파일이 있으면 sync가 안 다시 쓴다)은 사실과 다릅니다. `src/codex/catalog/retained-sync.ts`는 존재 여부가 아니라 새로 만든 바이트와 디스크 바이트를 비교하고, 같을 때만 `catalogWritten: false`를 돌려줍니다. 이 no-op은 #857/#1407 이후 mtime 기반 카탈로그 신선도 판정의 핵심이라, 이슈가 말한 “슬러그 집합 드리프트면 무조건 다시 쓰기”를 그대로 넣으면 `visibleNativeSlugs`와 `visibility: "hide"` 보유 설계 때문에 매 sync마다 영원히 다시 쓰게 됩니다. 두 번째 요청(비활성 네이티브가 `hide`로 남는 것)은 맞고, 의도된 계약입니다. `nativeAlias`가 없을 때 `desktopAllowlistSuppressedNativeSlugs`는 빈 집합이고 행은 남습니다. 이 PR은 동작을 바꾸지 않고, 영어 `codex-app-models.md`와 `structure/catalog.md`에 Desktop이 `hide`를 무시할 수 있고 `disabledModels`는 카탈로그 제어일 뿐 `src/router.ts` 입장 거부가 아니라는 운영자 결과를 적습니다. 테스트는 `tests/codex-integration/doctor.test.ts`에 주입형 여섯 케이스, `native-model-toggle.test.ts`에 hide 보유 vs alias 생략 계약을 고정합니다. `structure/runtime.md` ownership 표도 doctor 쪽을 갱신합니다.

라인 약 1041 근처 `collectDefaultModelExposure` (doctor.ts) - 프록시 `/v1/models`와 디스크 카탈로그 중 하나라도 핀을 포함하면 `exposed`입니다. 라이브 프록시는 이미 빠졌는데 아직 sync 전 카탈로그에 `visibility: "list"`로 남아 있으면, #4646이 겪은 “핀은 죽었는데 진단은 조용”과 비슷한 거짓 OK가 납니다. 주석은 인코딩·신선도 차이로 가짜 미노출을 막자고 했지만, 쿼터로 `/v1/models`만 먼저 빠진 경우엔 경고가 늦어질 수 있습니다.
라인 약 `fetchExposedModelIds` - 비루프백 바인드에서 doctor는 데이터플레인 키 없이 401을 받고 `proxyIds === null`로 카탈로그만 봅니다. 그때 카탈로그가 아직 `list`로 남겨 둔 핀이면, 원격 바인드 설치에서 라이브 미노출을 못 보고합니다.
경로 `src/cli/doctor.ts` 크기 - 현재 `dev`에서 1437줄이고 이 PR로 약 1619줄입니다. 2000줄 래칫 아래라 당장 실패는 아니지만, collect 헬퍼가 계속 붙는 파일이라 다음 섹션부터는 분리 후보입니다.
경로 번역 가이드 - 영어 `codex-app-models.md`만 늘렸고 ko/ja 등 번역본은 안 건드렸습니다. 기존 문장을 고친 게 아니라 새 절이라 당장 모순은 없지만, 번역 패스 전까지 로케일 문서가 덜 완전합니다.
경로 CI - 작성 시점 Cross-platform CI·PR hygiene·CodeRabbit이 아직 QUEUED/PENDING이고 `mergeStateStatus`는 REVIEW_REQUIRED로 BLOCKED입니다. 로컬 스위트는 레인 규칙상 안 돌렸다고 본문에 적혀 있으니, hosted green이 곧 실행 검증입니다.

메인테이너의 판단이 필요한 지점
- 프록시와 카탈로그를 OR로 맞춰 “둘 다 미노출일 때만 경고”를 유지할지, 라이브 `/v1/models`가 성공 응답이면 그 집합만으로 미노출을 확정할지.
- 비루프백(401) 설치에서 카탈로그 fallback만으로도 충분한지, 아니면 doctor가 읽기 전용으로 쓸 수 있는 관리면 경로를 따로 둘지.
- `StartupHealth`/GUI에 같은 사실을 넣지 않은 선택이 이번에도 맞는지(캐시 키·신선도 이유), 아니면 다음 PR에서 별도 non-cached 필드로 갈지.
- #4646을 이 PR만으로 Closes 할지, ask 1 오해에 대한 문서/이슈 코멘트를 남기고 닫을지.

너의 추천
hosted CI(특히 `tests/codex-integration/doctor.test.ts`와 `native-model-toggle.test.ts`)가 초록이면 머지 쪽으로 가도 됩니다. Closes #4646 의도도 본문 논증과 맞습니다. 머지 전에 OR-매치 거짓 OK만 한 줄로 인정할지(문서/주석), 아니면 라이브 성공 시 프록시 단독 확정으로 바꿀지 짧게 정하면 됩니다. types/config 분할에 무효화될 PR이 아니고 중복 랜딩도 아닙니다.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:

Issue #5004, opened 2026-09-18
[Bug] ocx start creates duplicate instances and hops ports on Windows instead of reusing or exiting


kge-lab:
### Client or integration

Codex CLI

### Area

CLI

### Summary

When running `ocx start` in Windows PowerShell while a proxy instance is already active (or the port is still bound), opencodex does not refuse to start or reuse the existing proxy. Instead, it issues a port busy warning, hops to an arbitrary high port, spawns a duplicate background instance, and rewires the Codex runtime config to target the secondary instance while the previous instance remains running.

### Reproduction

1. In Windows PowerShell, run `ocx start`.
2. Notice proxy starts (e.g., PID 17712, port 58285).
3. Without stopping the first process, run `ocx start` again in PowerShell.
4. Observe output warning `⚠️ Port 58285 is busy; starting opencodex on 62254.` and creating a duplicate orphan instance.

### Version

2.58.0

### Operating system

Windows 11 (x64, PowerShell)

### Logs or error output

```shell
PS C:\windows\system32> ocx start
⚠️  Proxy already running (PID 17712, port 58285). Use 'ocx stop' first.
PS C:\windows\system32> ocx start
⚠️  Port 58285 is busy; starting opencodex on 62254.
[model-rename-migration] renamed "google-antigravity/gemini-3.6-flash" to "gemini-3.7-flash" in the saved config: Google retires the previous Antigravity Flash generation from Cloud Code Assist when its successor ships, so the saved id no longer resolves to a live model.
🚀 opencodex proxy running on http://localhost:62254
   POST /v1/responses → provider translation
   POST /v1/chat/completions → OpenAI-compatible clients
   GET  /healthz      → health check
   GET  /api/*        → management API
   GET  /             → GUI dashboard
[opencodex] Codex runtime: C:\Users\[USER]\AppData\Roaming\npm\codex.opencodex-real.cmd (version=0.154.0, source=configured)
   + 30 models appended to Codex catalog (C:\Users\[USER]\.codex\opencodex-catalog.json)
Pointed Codex's built-in openai provider at the opencodex proxy (openai_base_url + realtime sideband override).
```

### Redacted configuration

```json

```

### Checks

- [x] I searched existing issues and documentation.
- [x] I removed secrets, tokens, account details, request credentials, and personal data.

lidge-jun:
Reproduced from the code path — thank you for the exact transcript, it is what made this diagnosable without a Windows machine.

[...]

PR #5007, opened 2026-09-18
fix(codex): guard the first read in the history injection preflight

lidge-jun:
## Summary

- The history injection preflight guarded the wrong operation, so the #4943 fix did not hold on macOS. `sqlite3_open_v2` never reads page 1, so a store whose header says WAL is not inspected until the first prepare. On macOS that is where the absent `-shm` is raised — one frame above `openCodexStateForPreflight`, after the classification and the immutable fallback have already returned. The error reached the catch-all, so a user with a cleanly-closed WAL store still got `history_injection_preflight_unavailable` and `ocx sync` still had no way forward.
- The failure is not a classification problem. macOS raises exactly the condition the narrowing was written for: `SQLiteError`, `code: "SQLITE_CANTOPEN"`, `errno: 14`, `unable to open database file`, with neither sidecar on disk. `isStateDbCantOpenError` matches it. It simply never saw it, because the read happened in the caller.
- The first read now happens inside the guarded attempt: `openCodexStateForPreflight` opens `{ readonly: true }` and reads `sqlite_master` before returning the connection.
- Every existing guarantee is intact. `{ readonly: true }` stays the primary path, so a live writer's WAL is still joined and a thread another process just migrated to paginated history still refuses. The immutable fallback is still admitted only when neither `-wal` nor `-shm` is on disk. Any other failure still keeps the original error and the refusal that follows from it.
- The test seam now names which step to fail (`"open"` or `"first-read"`) instead of only the constructor; existing zero-argument hooks are unaffected.
- Worth stating plainly: #4957 landed with a full green Linux suite and nine green Windows shards, and the gap was in its one case with no injected failure — the case that is deliberately platform-dependent. Bun bundles SQLite 3.53.2 on Linux and that build materializes both sidecars on the same first read, so the primary path just succeeds there. macOS uses the system `libsqlite3` (3.51.0 on `macos-26-arm64`) and refuses. Coverage looked complete because the other five cases inject the failure and therefore assert the narrowing decision rather than the platform.

Closes #4943 on macOS; Linux and Windows behavior is unchanged.

## Verification

- **Local verification was not run, because this lane forbids it.** No local suite, focused test, typecheck, build, install, or `ocx` invocation was executed at any point. Everything below is hosted CI at head `a052e02f797b4ee4038bde71557bc48a485a4b9a`.
[...]

PR #5015, opened 2026-09-18
fix(cli): never hop off a busy port into a duplicate proxy

lidge-jun:
## Summary

On Windows 11, two identical bare `ocx start` invocations against the same home produced different outcomes. The first refused correctly ("Proxy already running (PID 17712, port 58285). Use 'ocx stop' first."); the second printed "Port 58285 is busy; starting opencodex on 62254", bound a second listener, took over this home's pid and runtime-port records, and re-pointed the Codex runtime config at the duplicate while the original kept serving. The reporter was left with two proxies and an editor talking to the wrong one.

The hop is the defect this closes. `chooseListenPort` walked away from a busy preferred port without ever asking who held it: `findAvailablePort` simply took the next free port, and `allowEphemeralFallback` is `!hardPin`, so any start without an explicit `--port` could hop. Every way the owner probe can answer "nobody" - a stale record, a probe that lost a race, a loopback family split - therefore degraded into "start a duplicate and rewire Codex".

**What changed**

- `src/server/proxy-liveness.ts` gains `probePortOwner`: an identity-checked answer to "who holds this exact port", independent of the pid file and the runtime-port record, asked on both loopback families via `loopbackProbeHosts`. It also gains `START_OWNERSHIP_LIVENESS` (1500ms x 3), the budget the stop path already uses for the mirror-image decision.
- `src/cli/dispatch.ts` gains `decideBusyPreferredPort`, a pure decision next to `decideStartWithLiveOwner`. An opencodex holder is refused with the same message the owner path prints, or exits 0 under `OCX_SERVICE=1` so the service wrapper's retry loop still terminates. A holder that did not identify as opencodex is reported as unidentified rather than called foreign, because `proxyIdentityAt` returns the same `null` for a foreign server, an unreachable one, and one that lost a race - so the start stops and names `--port` instead of silently taking an arbitrary port.
- `src/cli/index.ts` wires the probe and the decision into `chooseListenPort` ahead of the hop message, and gives the pre-bind owner probe the same budget. That second change matters on its own: a negative answer there is acted on twice over, because `findProxyOwnerBeforeJournalRecovery` also deletes this home's pid record and reconciles the journal on the way past.

**What is deliberately unchanged**

An explicit `--port` still never hops - it waits for the pin through `reclaimListenPort` and fails rather than moving. A configured `port: 0` is a request for an OS-assigned port, not a collision, and still allocates one. `shouldPersistSelectedPort` and the sibling rules are untouched. The service-wrapper path keeps its exact stay-out-of-the-way semantics on both the owner check and the new busy-port check. A legitimate first start on a machine whose default port is taken is still possible with `ocx start --port <port>`; what is no longer possible is a duplicate of an existing opencodex.

**On the second defect, stated honestly**

I could not prove why the owner probe returned `null` on the reporter's second run, and I do not have a Windows runner to reproduce it. What the source shows: the refusal path of the first run writes and deletes nothing before `process.exit(1)`, so it did not strip the second run's evidence; the probe was a single 750ms attempt with no retry on transport failure, and `proxyIdentityAt` discards the error, so a timeout is indistinguishable from an empty port; and `startServer` canonicalizes a literal `localhost` bind to 127.0.0.1 precisely because Windows resolves that name IPv6-first, while `probeHostname` hands the literal name back and leaves the family to the resolver. The retry budget and the two-family probe address the two candidates that are visible in the code. The first fix is what makes the trigger stop mattering: whatever the probe answers, the outcome is now a refusal rather than a duplicate instance.

Two adjacent paths are out of scope and unchanged: the connected-client listener in `src/client/runtime.ts` has its own fallback (reached only after the owner check has already refused a live proxy), and a home configured with `port: 0` has no configured port to collide on.

Docs: `README.md`, the English lifecycle/quickstart/installation pages, and the same passages in all seven translated locales now describe the refusal instead of the hop. `structure/runtime.md` records the contract and `structure/overview.md` adds INV-START-01, bound to `tests/cli/cli-dispatch.test.ts`.

Closes #5004

## Verification

- **No local verification was run. This lane forbids it**: no `bun test` in any form, no `bun run typecheck`, no build, no install, and no `ocx` invocation. Correctness was established by reading the source and by static review of the diff; the hosted CI run below is the evidence.
- Dispatched `ci.yml` with `lane=all` on this branch so the nine Windows shards run, since `pull_request` does not schedule them. Windows is the reported platform.
- Regression coverage added, all in existing test files (no test-layout changes):
  - `tests/cli/cli-dispatch.test.ts` - the `decideBusyPreferredPort` matrix at runtime, using the reported shape (preferred 58285 held, ephemeral 62254 free): an opencodex holder refuses, an unidentified holder refuses, `OCX_SERVICE=1` stays out, only the exact `1` sentinel counts as service context, and `port: 0` / hard pin / no-hop still proceed. Plus source oracles that `chooseListenPort` probes before it decides and decides before it can print the hop message, that both refusals exit 1 and the stay-out exits 0, and that the pre-bind owner probe spends the same budget before it deletes the pid record.
  - `tests/server/proxy-liveness.test.ts` - `loopbackProbeHosts` for every bind spelling, `probePortOwner` finding an owner that answers only on the other loopback family, surviving two lost probes before succeeding, and returning `null` for both a foreign body and a refused connection.
  - `tests/cli/cli-ready.test.ts` - the existing `handleStart` oracle matched "Proxy already running ... exit(1)" anywhere in the file. `chooseListenPort` now carries an earlier stay-out/refusal pair, so the match would have silently moved to the new code and stopped asserting anything about the owner branch. It is now anchored at `decideStartWithLiveOwner({`.
- Each new test fails without the change: the decision function, `probePortOwner`, `loopbackProbeHosts`, and `START_OWNERSHIP_LIVENESS` do not exist on `dev`, and the source oracles assert call sites the old `chooseListenPort` does not have.
- Expected noise on this run: macOS legs are unreliable for the reason tracked in #4956, and there is a separate macOS regression in the WAL preflight being fixed in #5007. Both are distinguishable - timeouts, or an 8ms assertion in a file this PR does not touch.

## Checklist

Security note: no auth, credential, token, workflow, release-automation, or dependency-installation surface is touched. The new probe is the existing loopback `/healthz` identity request through `directLocalHttpFetch`, which bypasses proxy environment variables; nothing new is logged beyond the pid and port the refusal already printed.

## Summary by CodeRabbit

* **Bug Fixes**
  * `ocx start` now identifies the process using a busy preferred port and stops instead of silently switching ports.
  * Prevents a second proxy from running alongside an existing one.
  * Service-managed starts safely stay out when a proxy is already active.
  * Use `--port` to select another port or `port: 0` to request one from the operating system.

* **Documentation**
  * Updated lifecycle, quickstart, installation, and runtime-port guidance across supported languages.


lidge-jun:
Merging with one Windows shard red that is not this change.

`windows 2/9` failed `connected-client runtime probe scope > observes only the selected runtime and leaves full diagnostics available` at 15339.29ms, out of 308 passing in that shard. The assertion is `expect(result.status).toBe(0)` receiving `null`, at `tests/cli/cli-connect-readiness.test.ts:281`. A `null` status from `spawnSync` means the child was killed by the `timeout: INTERNAL_DEADLINE_MS` bound rather than exiting, so a spawned child produced nothing for the full 15 seconds. That is the pattern in #4956, and that file imports nothing this branch touches.

The other eight Windows shards, the full Linux suite, `gates`, `enforce-target` and the docs build are green at this exact head. Windows is the reported platform for #5004, which is why the nine-shard suite was dispatched rather than relying on the pull-request event.

Worth recording separately, and I have added it to #4956: this is the second test file to show the cold-spawn exposure that #4948 fixed in `tests/cli/cli-status-json.test.ts`. Same shape — a `spawnSync` child measured against a bound sized for a warm one — in a file that was never touched. The fix there moved the cold cost into an explicit setup outside the measured window, and the same remedy is available here.

PR #4977, opened 2026-09-17
fix(transport): decide fresh-connection opt-out at the dispatch boundary

lidge-jun:
Carries #4804 by @Yum-wu, rebased onto current `dev` with the review finding fixed. Original commit authorship is preserved.

## Summary

Adds `OCX_FRESH_CONNECTION_HOSTS`, a comma-separated list of hostnames whose outbound sends bypass Bun's keep-alive pool: the request goes out with `Connection: close` and `keepalive: false`. Exact hosts and their subdomains match case-insensitively, a leading dot is trimmed, and an unparseable target falls back to default behavior rather than throwing. Unset or empty means nothing changes, so every existing deployment is untouched.

This exists because pooled connection reuse against certain upstreams strands a request on a half-dead socket, and the only reliable local remedy is to stop reusing the connection for that host.

**The defect this carry fixes.** Freshness was computed inside `httpFetch`, against the URL handed to it. A `dispatchOverride` can select or rebuild the destination after that point — OAuth revalidation paths do exactly this — so the value measured was not the value used, and it was wrong in both directions: a host that only became the target after the override never matched, and a host that stopped being the target still matched. The decision now happens in the executor, which is the last place that sees the URL that actually goes on the wire. Both directions are pinned as tests.

**One thing deliberately reverted from the original branch.** The original also threaded the `beforeDispatch` headers into the dispatch init, so that a hook could not overwrite `Connection` afterwards. Moving the decision into the executor makes that unnecessary — the executor runs after the hook, so the policy wins regardless — and threading it would have quietly turned `beforeDispatch` from an observer into a mutator for every existing caller. All four current implementations (the Codex reserve dispatch guard and three selection-currency guards) only read the headers and refuse the send by throwing, so the change would have been inert today and a trap later. The hook keeps the contract it has on `dev`, and a test pins that the policy still wins against a hook that sets `Connection: keep-alive`.

**What this does not guarantee.** It is a per-host escape hatch, not a fix for the upstream behavior that makes it necessary. `keepalive: false` is passed through to Bun and its effect is whatever Bun does with it; the `Connection: close` header is the part with defined meaning. Matching is by hostname only — port and path are not considered — and the variable is read per send, so changing it mid-process takes effect on the next request.

**Also fixed on the carry:** the new test file had no entry in `scripts/test-layout/layout.json` or `tests/fixtures/test-layout-expected.json`, which both layout guards require.

## Verification

[...]

PR #5022, opened 2026-09-18
fix(transport): apply the fresh-connection policy to a selected provider transport

lidge-jun:
## Summary

`OCX_FRESH_CONNECTION_HOSTS` lets an operator name upstream hosts that must never reuse a pooled
connection, which is the mitigation for an upstream that keeps a socket open after it has stopped
serving it. The policy was enforced inside the executor `providerFetch` builds and handed to a
`dispatchOverride` as its third argument. Neither production override uses that executor when a
provider-scoped transport exists: both re-read `route.provider.fetch` at the send boundary, because
credential reselection can install a different transport after the wrapper was constructed, and
then call that implementation directly.

The result was that naming the affected host had no effect on exactly the providers that need a
custom transport. Concretely, an xAI OAuth turn resolves through `resolveProviderTransport`, which
installs a provider-scoped fetch and rewrites the destination to the Grok CLI host; that send went
out with keep-alive intact even with the host configured.

`sendWithConnectionPolicy` now applies the policy around whichever fetch performs the physical
send, so it follows the selection rather than the construction. Both overrides route through it:
`oauthDispatch` in `request-transport.ts`, and the native Chat key-revalidation override in
`chat-native.ts`, which carried the same bypass and is not mentioned in the issue.

Provider transport behavior is preserved rather than replaced: the selected implementation is still
the one that sends, so xAI's pinned `x-grok-req-id` and stable compatibility headers, pacing
(which runs outside the override), the auth-snapshot currentness check, and response observation
are all unchanged. The helper is idempotent, so an override that hands the send back to the
supplied executor passes through it twice and both passes derive the same headers from the same
wire URL.

One behavior change beyond the reported bug is worth review: the native Chat provider-transport
branch previously did not force `redirect: "manual"`, while the executor branch beside it did. It
now does, matching the credential-redirect contract that `tests/lib/credential-redirect-guard.test.ts`
holds every other credential-bearing transport to. Following a redirect there would resend the
`Authorization` header to the redirect target.

Docs: `structure/transports/responses.md` described the executor as the final boundary, which is
what the code contradicted; it now names the selected physical fetch and records why a regression
for this policy cannot be written as a cooperative override. `OCX_FRESH_CONNECTION_HOSTS` had no
public documentation at all, so the server configuration reference now documents the
comma-separated syntax, case-insensitive exact and subdomain matching, leading-dot handling, the
two resulting fetch options, and the latency cost.

Closes #4992

## Verification

Local verification was not run: this lane forbids any local suite, focused test, typecheck, build,
install, or `ocx` invocation, so hosted CI is the executable verification for this change. What was
done instead is static reasoning against the tip the branch is based on, plus a regression designed
to fail on the unfixed code.

- `tests/responses/fresh-connection-optout.test.ts` gains an end-to-end case that enters through
  `handleResponses` with an xAI OAuth credential, so it reaches the real `oauthDispatch` selection
  instead of a hand-written override. It asserts the physical send carries `Connection: close` and
  `keepalive: false`, that the destination is the rewritten provider host, and that
  `x-grok-req-id` is still present — the last assertion is what proves the provider's own fetch
  still ran rather than being replaced by the generic executor. Every existing case in that file
  passes on the unfixed code, which is why the new one had to go through the server.
- The four existing `providerFetch` cases, `tests/lib/credential-redirect-guard.test.ts`, and
  `tests/server/server-xai-header-parity.test.ts` cover the neighbourhood this change touches.
- Ratchet: no file here is at or near its cap. `fresh-connection-optout.test.ts` goes from 248 to
  340 lines with no baseline entry, against a 2000-line threshold; it is already registered in both
  `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`, so no layout
  change is needed.
- Merge-union check: the change touches the fresh-connection policy and two override call sites. No
  open pull request or commit that landed on `dev` while this was written touches
  `fetch-helpers.ts`, `oauthDispatch`, or the native Chat override, and the change adds no count,
  roster, locale catalog, or exhaustive map that a concurrent branch could disagree with.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for `OCX_FRESH_CONNECTION_HOSTS`, allowing specified upstream hosts and subdomains to use fresh connections instead of pooled connections.
  - Matching is case-insensitive, supports comma-separated hostnames, and ignores leading dots.
  - Fresh-connection settings now apply consistently across provider transports, including native chat requests.

- **Documentation**
  - Documented configuration behavior, host matching rules, and connection handling across request dispatch paths.


lidge-jun:
## 리뷰 · 우선순위 73 / 80

이 PR은 #4992를 고칩니다. 지금 `dev`(11bc4f708, package 2.59.0)에서는 #4977로 들어간 `OCX_FRESH_CONNECTION_HOSTS` 정책이 `providerFetch`가 만든 executor `dispatch` 안에만 있습니다. 그런데 production의 `oauthDispatch`(`src/server/responses/request-transport.ts`)와 native Chat 키 재검증 override(`src/server/chat-native.ts`)는 보낼 때 `route.provider.fetch`(또는 `activeProvider.fetch`)를 다시 읽고, 그게 있으면 executor 대신 그 구현을 바로 호출합니다. 그래서 xAI처럼 `resolveProviderTransport`가 provider-scoped fetch를 심고 목적지를 Grok CLI 호스트로 바꾸는 경로에서는, 운영자가 env에 그 호스트를 넣어도 keep-alive 재사용이 그대로일 수 있습니다. tip 코드로 보면 이 구멍은 아직 열려 있습니다.

고치는 방식은 정책을 “만든 래퍼”가 아니라 “실제로 보내는 fetch” 둘레로 옮기는 것입니다. `src/server/responses/fetch-helpers.ts`에 `sendWithConnectionPolicy`를 두고, `wantsFreshConnection`이 참이면 `Connection: close`와 `keepalive: false`를 넣은 뒤 `redirect: "manual"`과 함께 선택된 `physicalFetch`를 호출합니다. `providerFetch`의 executor는 이 헬퍼를 쓰고, `oauthDispatch`와 chat-native override도 선택된 fetch를 같은 헬퍼로 감쌉니다. 제공자 전용 헤더·패이싱·auth snapshot 관찰은 유지한다고 PR이 설명합니다. 헬퍼는 멱등이라, override가 다시 executor로 넘겨도 같은 wire URL로 같은 헤더가 두 번 적용됩니다.

한 가지 tip과 달라지는 행동은 chat-native provider-transport 분기입니다. 예전에는 이 분기가 `redirect: "manual"`을 강제하지 않았고, 옆 executor 분기는 강제했습니다. 지금은 헬퍼가 항상 manual redirect를 넣어서, credential을 실은 요청이 리다이렉트 대상에 `Authorization`을 다시 보내지 않게 맞춥니다. `tests/lib/credential-redirect-guard.test.ts` 계약과 맞추는 쪽이라 방향은 맞고, 의도된 보안 정렬인지 한 번만 확인하면 됩니다.

회귀는 `tests/responses/fresh-connection-optout.test.ts`에 `handleResponses` + xAI OAuth 경로로 들어가는 e2e를 추가했습니다. 기존 케이스는 override가 executor를 호출해서 고쳐지기 전에도 통과했고, 이번 케이스는 물리 send에 `Connection: close` / `keepalive: false`, 목적지가 `cli-chat-proxy.grok.com`, 그리고 `x-grok-req-id`가 남아 제공자 fetch가 교체되지 않았음을 같이 봅니다. docs는 `structure/transports/responses.md`의 “final executor boundary” 서술을 고치고, docs-site `reference/configuration/server.md`에 env 문법·매칭·지연 비용을 처음 공개 문서로 넣었습니다. types/config 스플릿과 무관하고, Preview deploy는 계획에 없습니다. mergeable은 지금 blocked라 CI 초록을 기다리는 상태입니다.

src/server/responses/fetch-helpers.ts (`sendWithConnectionPolicy`) - tip에는 없고 PR이 정책을 선택 fetch 둘레로 옮기는 본체입니다.
src/server/responses/request-transport.ts (`oauthDispatch`) - tip 약 L410은 `fetchImpl`을 직접 호출합니다. PR은 `sendWithConnectionPolicy(fetchImpl, …)`로 바꿉니다. #4992의 핵심 구멍.
src/server/chat-native.ts (키 재검증 override) - tip은 provider fetch를 직접 호출하고 redirect manual도 없습니다. PR이 같은 헬퍼로 감싸며 redirect 계약을 맞춥니다.
tests/responses/fresh-connection-optout.test.ts (OAuth e2e) - tip 스위트만으로는 이 구멍을 못 잡았습니다. 호스트 문자열 `cli-chat-proxy.grok.com`에 묶여 있어 xAI 목적지 변경 시 깨질 수 있습니다.
docs-site/.../server.md · structure/transports/responses.md - tip은 공개 env 문서가 거의 없고 structure는 executor가 최종 경계라고 적혀 tip oauthDispatch와 어긋납니다. PR이 둘 다 맞춥니다.

메인테이너의 판단이 필요한 지점

- chat-native에 `redirect: "manual"`을 새로 강제하는 것을 이번 버그 수정과 같은 PR에 둘지, 별도 보안 패치로 쪼갤지.
- chat-native 우회 경로에도 oauth e2e와 대칭인 회귀가 필요한지(지금 새 테스트는 Responses/oauthDispatch만 탑니다).
- CI가 초록이 되기 전에 merge할지, 아니면 호스티드 검증을 기다릴지(로컬 스위트는 레인 규칙으로 돌리지 않았다고 함).

너의 추천

CI가 초록이면 merge하세요. tip 기준 #4992 구멍을 정확히 막고, #4977 완화가 provider-scoped transport에도 닿게 합니다. Closes #4992와 문서·e2e가 한 묶음이라 범위도 맞습니다. chat-native redirect 정렬과 테스트 호스트 하드코딩만 인지하고 가면 됩니다. 라벨은 바꾸지 마세요.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
### CI on this branch cannot produce a test signal, for a reason that predates it

This branch is based on `11bc4f708c`, which was the `dev` tip when it was cut, and that commit
fails `bun run build:gui`:

```
src/i18n/native-main-copy.ts(11,26): error TS7053: Element implicitly has an 'any' type because
expression of type 'LabLocale' can't be used to index type '{ en: ... }'.
  Property 'vi' does not exist on type '{ en: ... }'.
```

It is a union defect between two individually correct merges. `e80e571f63` (#4984) added `vi` to
`LabLocale`, which `Locale` aliases; `11bc4f708c` (#4781) then added
`gui/src/i18n/native-main-translations.ts` as a deliberately "closed, nine-locale namespace" and
indexes it with that ten-member `Locale`. Each was green on its own. The `dev` push run for
`11bc4f708c` is red for exactly this
([35311351776](https://github.com/lidge-jun/opencodex/actions/runs/35311351776)).

Every test leg builds the GUI first, because the suite serves `gui/dist` and reads it back, so
"Test in fresh-process batches" is skipped rather than run. The red checks here are therefore not
evidence about this change in either direction.

#5020 fixed it and is merged, so `dev` at `88249ed750` builds again. I did not rebase this branch,
because integration is the maintainer's call in this lane. To get a real signal I cherry-picked this
commit onto the fixed `dev` on a throwaway branch and dispatched the full lane there; results are
linked in a follow-up comment. Updating this branch onto current `dev` is all this needs.

lidge-jun:
Your own regression is red at this head, which is the useful kind of red: `the OAuth dispatch boundary > a provider-scoped transport selected at dispatch receives the fresh-connection policy` fails at 44.22ms in `test 1/4`, with 309 other cases in that shard passing.

A 44ms assertion is not the #4956 timeout pattern and not inherited from `dev` — `dev` is green on that shard. The test is asserting the thing the change exists to deliver and not getting it, so the fix does not yet reach the path the test drives.

Worth checking against what landed while you were working: #5015 added identity-checked port-owner probing in `src/server/proxy-liveness.ts` and #4977 moved the fresh-connection opt-out decision to the dispatch boundary. If your change assumes the pre-#4977 shape of that decision, the selected transport may be receiving a policy resolved one frame earlier than you expect.

Do not relax the assertion to match current behaviour. If the assertion is wrong, say why and what the contract actually is; if it is right, the dispatch path needs to carry the policy through to the provider-scoped transport.

lidge-jun:
### Hosted CI at `2afa7b56f2`

[Run 35315333182](https://github.com/lidge-jun/opencodex/actions/runs/35315333182), dispatched with
`lane=all` so the nine Windows shards run: 18 jobs green including `gates` and all four Linux test
shards. `test 1/4` is the shard that carries the new end-to-end case, so the regression is
confirmed against the real `oauthDispatch` selection rather than only by reading.

One red leg, and it is not this change: `windows 5/9` fails
`unavailable management authority rejects log cursors before parsing` after 15.8s, in a log full of
`icacls` ACL-hardening `ETIMEDOUT` stalls ("transient icacls stall … budget exhausted"). That test
is a management/log-cursor case; this PR touches the physical-send boundary in
`fetch-helpers.ts`, `oauthDispatch` and the native Chat override, none of which it reaches. I am
reporting it rather than retrying it, since a timeout under a stalled Windows filesystem is a
platform condition and not something to paper over.

The earlier red checks on this PR came from the base commit's GUI build break, which #5020 fixed;
that is covered in the previous comment.

lidge-jun:
Merging. Your own regression is now green — `the OAuth dispatch boundary > a provider-scoped transport selected at dispatch receives the fresh-connection policy` passes at this head, which is the assertion this change exists to satisfy.

The one red is `windows 5/9`: `unavailable management authority rejects log cursors before parsing` at 15839.90ms, out of 121 cases in that shard. That is a timeout past the 15-second internal deadline in a file this branch does not touch, and it is the pattern tracked in #4956 — the cold-spawn class specifically, where a child measured against a bound sized for warm children pays a cold module graph and batch composition decides which file pays it.

Evidence at this exact head otherwise: all four Linux shards, `gates`, `storage policy`, `api usage`, and Windows shards 1, 2, 3, 4 and 9 green, with 6, 7 and 8 still running.

I dispatched the nine-shard Windows suite rather than relying on the pull-request event, because this change is in the transport dispatch path and Windows is where the fresh-connection behaviour is most likely to differ.

Issue #5019, opened 2026-09-18
Pool routing never selects the main login automatically when the request carries that login's own bearer


FredAmartey:
### Client or integration

Codex App

### Area

Authentication and account pool

### Summary

In Pool mode with the Codex app signed in as the main login and one stored account added through `ocx account login openai`, automatic placement never picks the main login. Every unpinned request goes to the stored account, whatever the usage numbers, the strategy or `codexAccountPriorities` say. Main only serves when it is pinned, or when no stored account can be selected.

The cause is in `resolveCodexAuthContext` (`src/codex/auth-context.ts`): `nativeMainReadsForbidden` is set whenever `requestScopedMainCredential` is true, and selection receives `isMainAccountTokenLive: () => preserveRequestOwnedMainPin`. That callback is true only when main is the effective pin with headroom, so for an unpinned request `codexAccountUnusableReason("__main__")` returns `main_credential_unavailable`, `getEligiblePoolAccounts` never lists main, and the strategy only compares the stored accounts. Main is reached afterwards through `resolveCallerOwnedMainContext`, but only as the fallback when nothing else is selectable.

The fence itself makes sense: a request that carries its own main bearer must not read or reconcile the stored `auth.json`. But that same bearer is exactly what would be forwarded if main were selected, so excluding main from the comparison discards the usage, priority and reset ordering the operator configured. With one stored sibling the pool degrades to "stored account until it cannot serve, then main". #3996 (fixed) made the caller-owned fallback reachable in more cases; this is about main taking part in the ordering before that fallback.

Expected: for a request carrying a forwardable main bearer, main takes part in selection as a candidate served from that bearer, ordered by the same usage, priority and strategy as the stored accounts, without touching the stored main credential.

### Reproduction

1. Canonical `openai` provider in Pool mode (`codexAccountMode: "pool"`, `authMode: "forward"`). Codex app signed in as the main login. Add one stored account: `ocx account login openai --device`.
2. Make the numbers favour main: main at lower weekly usage than the stored account (here 46% against 75%). Clear any pin: `PUT /api/codex-auth/active` with `{"accountId": null}`; `GET /api/codex-auth/active` then shows `activeCodexAccountId` and `pinnedAccountId` both `null`.
[...]

PR #4935, opened 2026-09-17
fix(codex): key pool affinity on the conversation cohort, not the thread

lidge-jun:
## Summary

Pool affinity keyed every Codex V2 thread as itself, while upstream keys its prompt cache on the
whole conversation tree. Two requests could therefore carry an identical `prompt_cache_key` and be
served by different pool accounts. The split-off member's key asserts a warm prefix that is
deterministically cold on its account, so the prompt is replayed in full. Nothing fails and no log
line reports it; only tokens burn.

This makes the binding unit the cohort the client already declares.

### The upstream contract, re-derived at the current head

Verified in the pinned clone at `/Users/jun/Developer/codex/121_openai-codex` (`095da4b7e`):

- `codex-rs/core/src/client.rs` — `prompt_cache_key()` returns `responses_metadata.session_id`,
  or `{source}:{parent_thread_id}` for an internal session. Never the thread's own id.
- `codex-rs/core/src/agent/control.rs` — `session_id` "is equal to the root thread's ID", and that
  one `AgentControl` is "shared with every sub-agent spawned from that root".
- `codex-rs/core/tests/suite/prompt_cache_key.rs` — asserts `differentThreadIds: true` while root
  and child both send `promptCacheKey == expected_session_id`.

The clone predates openai/codex#44862 ("Preserve parent cache affinity for ephemeral forks"), which
merged on **2026-09-11** and makes an ephemeral fork inherit its parent's session id for exactly
this reason. The issue body is not stale; upstream has moved further toward cohort keying since it
was written.

### What changed

A request carrying `session-id` resolves to `app:HMAC(session, session)`, and every root, child and
grandchild of that tree lands on it. Without a session the cohort is read from the parent's lineage
record, falling back to `app:HMAC(parent, parent)` when this scope has not seen that parent — the
same key that parent derives for itself, so a chain of parent-only turns converges instead of
splitting at every depth. Which requests bind at all is unchanged.

### This is not a revert of #4546 wp8

Stated in the code, in `structure/`, and here, because a reader who concludes otherwise will flip
it straight back.

wp8 fixed a genuine incoherence: a child bound under the RAW parent id, an identity unrelated to
the root's own `app:HMAC(session, thread)` binding, so siblings shared an entry the root was not on
and a grandchild keying on its own parent landed on a key nobody had ever bound. A cohort key
cannot produce that, because the root's own binding **is** the cohort key — one identity for the
tree instead of two competing ones. That property is pinned as a regression rather than left as an
[...]

PR #4921, opened 2026-09-17
fix(codex): decide flagship model availability by roster and refusal evidence

lidge-jun:
## Summary

A Codex pool holding Plus and Free accounts keeps sending `gpt-5.6-sol` and `gpt-6-astra` to a
Free account and taking the upstream unsupported-model 400, after a quota refresh and a catalog
sync, with `attempts: 1` and no alternate attempt.

The ordering rules #4797 added are present at the tip and are correct. They simply have no
evidence to act on. `withoutModelDeniedAccounts` and `preferModelEntitledAccount` both read
`cachedDeniedCodexAccountIdsForModel`, which is cache-only by contract, and the roster cache it
reads expires five minutes after a catalog sync fills it (`MODEL_ROSTER_TTL_MS`). Nothing on the
flagship request path refills it: `resolveCodexModelEntitlements` is awaited only for
`ACCOUNT_GATED_NATIVE_OPENAI_MODELS`, which has held `gpt-daybreak-blue-latest` alone since the
2026-09-04 owner decision. So for most requests the denial set is `undefined`, both ordering rules
are the identity function, and the pool selects on quota alone — which is the Free account.

The refusal itself was the missing evidence. A 400 whose body is exactly `The '<model>' model is
not supported when using Codex with a ChatGPT account.` is authenticated, account-specific and
model-specific. It was spent on one retry and then discarded, so the next request repeated the
same selection and took the same 400.

### What changed

- `src/codex/observed-model-denials.ts` (new) records a confirmed refusal per (account, model) in
  a bounded store with a six-hour retention.
- `cachedDeniedCodexAccountIdsForModel` unions that evidence with the roster-derived denials.
- The passthrough lane records the refusal when it detects one, and clears the pair when the
  account successfully serves that model. `retryCodexPoolOnAlternateAccount` records it for the
  alternate account too, so a second denied account is not chosen next time either.
- Detection reads the model upstream actually named instead of rebuilding the sentence from
  `route.modelId`. `applyCodexAccountGatedWireNormalization` rewrites Daybreak to `gpt-5.6-sol`
  before dispatch, so upstream names Sol while the route still says Daybreak; the comparison never
  matched, which silently disabled both the alternate-account retry and the eight-rung
  same-account ladder that exists for exactly that model.

### What this deliberately is not

It stays evidence for ordering, not a gate. Consumers treat an observed refusal exactly like a
roster denial, so `withoutModelDeniedAccounts` still restores denied members when filtering would
[...]

Issue #4850, opened 2026-09-16
Caller-owned thread_spawn preview still reads physical main auth.json


lidge-jun:
### Client or integration

Codex App

### Area

Authentication and account pool

### Summary

A `thread_spawn` request that authenticates with the caller's own forwardable Codex bearer still opens the operator's physical native-main `auth.json` during request preview.

The native-main read fence already prevents quota priming, entitlement discovery, denial-cache credential validation, reconciliation, and final-auth selection from reading physical main for this credential domain. Pool eligibility remains outside that fence:

1. `prepareResponsesRequest` calls `previewCodexAccountForRequest`.
2. Priority preemption reaches `getEligiblePoolAccounts`.
3. Main eligibility calls `isMainAccountCredentialUsable`, which opens and parses `auth.json`.
4. `applySubagentModelFallback` invokes the preview callback and repeats the same read.

Final authentication does not select or send the physical main credential: it uses synthetic liveness and either selects a stored pool account or preserves the caller-owned credential. The preview can nevertheless score physical-main liveness, cached quota, and plan state and use that result when deciding whether to rewrite the subagent model.

Expected: a request authenticated with its own credential performs no physical-main credential reads or scoring.

Actual: request preview reads `auth.json` twice through pool eligibility.

Severity is bounded. There is no token disclosure, refresh, upstream send, or quota consumption through the observed path, and the credential-substitution hypothesis was checked and falsified. The observable consequence is that operator-main state can influence subagent model selection for a request that owns its own credential.

ADR-0086 records "read the physical main token for identity" as an explicitly rejected alternative, and #4813 states that caller-owned preview "could read the physical main token and score main differently from final authentication" — so this path being outside the fence appears to be an omission rather than a decision.

### Reproduction

1. Configure canonical OpenAI forwarding in Pool mode with:
   - a physical native-main `auth.json`;
   - one usable stored pool account;
   - a subagent fallback chain.
2. Send a `/v1/responses` `thread_spawn` request carrying a forwardable caller Codex bearer.
3. Instrument `fs.readFileSync` calls whose path ends in `auth.json`.
4. Observe these two stacks before final dispatch:

   - `prepareResponsesRequest -> previewCodexAccountForRequest -> pickPriorityPreemption -> getEligiblePoolAccounts -> isCodexAccountUsable -> isMainAccountCredentialUsable`
[...]

PR #5024, opened 2026-09-18
fix(codex): let a request-owned main bearer take part in pool ordering

lidge-jun:
## Summary

In Pool mode with the Codex app signed in as the main login and stored accounts added alongside it,
automatic placement never picked main. Every unpinned request went to a stored account whatever the
usage numbers, the strategy or `codexAccountPriorities` said, and main served only when pinned or
when no stored account could be selected at all.

The cause is one predicate. `resolveCodexAuthContext` hands selection
`isMainAccountTokenLive: () => preserveRequestOwnedMainPin` whenever the request carries main's own
forwardable bearer. That value answers a different question — is main the effective manual pin with
headroom — so for an unpinned request it is false, `codexAccountUnusableReason` returns
`main_credential_unavailable`, `getEligiblePoolAccounts` never lists main, and the strategy compares
only the stored accounts. Main was still reachable afterwards through `resolveCallerOwnedMainContext`,
but only as the fallback when nothing else was selectable, so a pool with one stored sibling degraded
to "stored account until it cannot serve, then main".

The fence that predicate sits behind is correct and is kept: a request carrying its own main bearer
must not read or reconcile the stored `auth.json`. But that same bearer is what would be forwarded if
main were selected, so it *is* main's live credential for this request. `requestOwnedMainCredentialIsLive`
now answers from the credential the request actually carries, and main is compared against the stored
accounts on the operator's own ordering. When selection names main, the request returns
`resolveCallerOwnedMainContext()` immediately — before any physical claim, token refresh or
materialization — so nothing reads, reconciles, primes or cools the stored profile, and the
request-owned credential still owns no affinity, entitlement or health state.

Deliberately unchanged:

- **Account-gated models.** Main stays out of `modelEligibleAccountIds`, because its roster is
  discovered from the stored credential this request may not read, so candidacy alone cannot
  manufacture a grant. That set currently holds one non-flagship slug, so it does not affect the
  models in the report.
- **Retained recovery and a draining profile.** Both still make main ineligible; they fence the
  identity rather than the credential. The new expression is monotone against the old one — an
  effective pin answers exactly as before — so the only behavior added is the unpinned case.
- **`nativeMainSelectionOnly`.** Still derived from the drain alone.

The predicate is shared rather than copied, for the reason `poolStateEligible` is: preview exists to
predict the resolution, and a preview that scores main differently hands subagent fallback a different
account than the one that serves. Final authentication and both preview builders in
`request-prepare.ts` (initial and post-decryption recovery) call the one exported expression, which is
the divergence class #4850 closed.

### The behavior change a reviewer should look at first

Because main is now scored, main's own recorded model failure reaches subagent fallback, where it was
previously invisible. `responses-preview-main-read-fence.test.ts` encoded the old answer in
"ownership alone leaves selection-only off", whose assertion was `finalAuth` = `pool-a` with the
physical file deleted. That case is replaced by the two directions of the new contract rather than
relaxed, and both keep the deleted `auth.json`, so any path that still needed the physical credential
takes an ENOENT instead of a quiet fallback.

This touches credential selection, so it needs the explicit security review `MAINTAINERS.md` requires
for auth paths. I am not merging it.

Closes #5019

## Verification

Local verification was not run: this lane forbids any local suite, focused test, typecheck, build,
install, or `ocx` invocation, so hosted CI is the executable verification for this change. What was
done instead is a static trace of the full chain — `resolveCodexAuthContext` →
`nativeMainReadsForbidden` → `isMainAccountTokenLive` → `codexAccountUnusableReason` →
`getEligiblePoolAccounts` → strategy → `resolveCallerOwnedMainContext` — against the base commit,
plus a check that PRs #4921, #4676 and #4935 did not already alter it (they did not: #4676 contains no
file in this path, and #4935's cohort keying never applies to a request-owned credential because
`poolStateEligible` excludes it).

- `tests/responses/responses-preview-main-read-fence.test.ts` now covers both directions: main is
  selected and served from the caller's bearer with the model untouched and the denial-cache
  validator never running, and a failure recorded against main reaches subagent fallback. The
  discriminator in the first case is that the recorded failure belongs to `pool-a`, so leaving the
  model alone is only possible if preview scored main.
- The pin case, the "preview that owns no credential still probes physical main liveness" case, and
  the ownership-validation case in the same file bound the change from the other side: an ordinary
  request that brought no credential still probes, and the pin path answers as before.
- Ratchet: the fence test goes from 542 to 587 lines with no baseline entry against a 2000-line
  threshold. `tests/codex-integration/codex-auth-context.test.ts` has five lines of cap headroom and
  was deliberately not touched.
- Merge-union check: this adds no count, roster, locale catalog or exhaustive map. It exports one new
  expression and changes three call sites of one predicate, all within the files above.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Requests using the main account’s forwarded credential are now eligible for normal account selection without requiring a manual pin.
  - Selected main-account requests use the caller’s credential instead of unnecessarily falling back to stored accounts.
  - Recorded main-account failures now participate in fallback routing, including configured alternative models.
  - Preview and final authentication now apply consistent account eligibility and routing behavior.

- **Documentation**
  - Updated provider guidance to describe main-account selection, exclusions, and fallback behavior.


lidge-jun:
## 리뷰 · 우선순위 76 / 80

이 PR은 지금 `dev` tip(`88249ed75`, #5020 Vietnamese native-main locale, package 2.59.0) 위에서 이슈 #5019를 고친다. Pool 모드에서 Codex 앱이 main으로 로그인돼 있고 저장 계정이 하나 더 있을 때, 핀을 안 건 요청은 사용량·전략·`codexAccountPriorities`가 뭐라고 하든 항상 저장 계정만 탄다. main은 핀했을 때나 저장 계정을 하나도 고를 수 없을 때만 나온다. 원인은 한 줄의 예측이다. `resolveCodexAuthContext`(`src/codex/auth-context.ts`)가 요청이 main 자신의 forwardable bearer를 들고 있으면 `isMainAccountTokenLive: () => preserveRequestOwnedMainPin`을 selection에 넘긴다. 그 값은 “지금 main이 유효한 수동 핀이냐”에 대한 답이라, 언핀 요청에서는 false가 되고 `codexAccountUnusableReason`이 `main_credential_unavailable`을 돌려 `getEligiblePoolAccounts`에 main이 안 들어가며, 전략은 저장 계정끼리만 비교한다. 그 뒤 `resolveCallerOwnedMainContext`로는 여전히 도달하지만, “다른 후보가 없을 때” 폴백뿐이라 저장 형제 하나 있는 풀이 “저장 계정이 못 줄 때까지 저장, 그다음 main”으로 퇴화한다.

고침의 핵심은 공유 식 `requestOwnedMainCredentialIsLive`다. 요청이 들고 온 forwardable bearer가 곧 main의 live credential이므로, 저장된 `auth.json`을 읽지 않고도 “live인가?”에 yes라고 답한다. 유효 수동 핀은 예전처럼 yes다. retained recovery나 프로필 drain처럼 신원을 막는 펜스는 그대로 no다. 그래서 식은 예전 답에 대해 단조(monotone)이고, 새로 열리는 동작은 언핀 + request-owned bearer 경우뿐이다. selection이 main을 고르면 물리 claim·토큰 refresh·materialization 전에 `resolveCallerOwnedMainContext()`로 바로 돌려, 저장 프로필을 읽거나 맞추거나 식히지 않는다. 같은 식을 final auth와 `request-prepare.ts`의 초기/복구 preview 두 곳이 같이 쓰므로, preview가 main을 다르게 점수 매겨 subagent fallback에 다른 계정을 넘기는 #4850 계열 분기를 다시 막는다.

의도적으로 안 바꾼 것도 분명하다. account-gated 모델의 `modelEligibleAccountIds`에서는 main이 계속 빠진다. 로스터는 저장 credential에서 나오는데 이 요청은 그걸 읽으면 안 되므로, 후보만으로 grant를 만들 수 없다. `nativeMainSelectionOnly`는 여전히 drain만으로  유도된다. 리뷰어가 먼저 볼 동작 변화는 이것이다. main이 이제 점수에 들어가므로, main에 기록된 모델 실패가 subagent fallback에 보이게 된다. 예전엔 main을 아예 안 점수 매겨서 그 실패가 안 보였다. 테스트 `responses-preview-main-read-fence.test.ts`는 예전 “ownership alone leaves selection-only off” 케이스를 양쪽 계약으로 갈아끼웠다. 물리 `auth.json`은 계속 지운 채로, (1) main이 선택되고 caller bearer로 200이 나가며 denial-cache validator가 돌지 않는지, (2) main 실패가 fallback에 도달하는지를 잠근다. 첫 케이스의 판별자는 기록된 실패가 `pool-a` 소속이라, preview가 pool-a를 골랐다면 모델을 XAI fallback으로 바꿨을 텐데 모델이 그대로면 preview가 main을 골랐다는 뜻이다. types/config 대분할과 무관하고, 중복 open PR도 이 키워드로는 #5024(#5019)만 보인다. 작성자 본문대로 auth 경로라 `MAINTAINERS.md` 보안 리뷰가 필요하고 본인이 머지하지 않는다고 명시했다.

라인 `auth-context.ts` `requestOwnedMainCredentialIsLive` - 공유 식이 pin OR (requestScoped AND NOT blocked AND NOT draining)이다. 핀 경로는 예전과 같고, 언핀에서만 main이 후보가 된다. 식 자체는 짧고 주석이 #4850/#5019 이유를 잘 적었다.
라인 `resolveCodexAuthContext` selection 직후 early return - `selected === MAIN` 이고 request-owned이면 `releaseTransientProbeGrant()` 후 caller-owned context로 바로 나간다. 물리 claim 전에 끊는 점이 펜스와 맞다. transient probe grant를 먼저 푸는지도 맞춰 두었다.
라인 `request-prepare.ts` preview / recovery - final과 같은 식을 두 번 다시 계산한다. 복구 경로에서 route가 바뀐 뒤에도 같은 식을 쓰는 점이 좋다. 하드코딩 true/false로 갈라지지 않게 한 것이 #4850 교훈과 맞다.
라인 테스트 ownership→ordinary candidate - upstreamAuth가 pool-access-token도 physical-main-token도 아니어야 하고, denial-cache main read stack이 비어야 한다. “조용한 폴백으로 저장 토큰을 쓰는” 회귀를 직접 막는다. status 200 + kind main이 함께 있어야 계약이 성립한다.
라인 테스트 recorded main failure → fallback - main이 이제 서브에이전트 실패 캐시에 보이므로, 운영자가 “예전엔 main 실패를 무시하고 저장 계정만 돌렸다”는 체감을 잃는다. 버그 수정의 필연적 부작용이지만 릴리즈 노트/이슈 코멘트에 한 줄 적어 두는 편이 안전하다.
경로 검증 메모 - 레인 정책상 로컬 suite/typecheck를 돌리지 않았고 hosted CI가 실행 검증이다. 정적 추적(resolveCodexAuthContext → fence → liveness → unusableReason → eligible → strategy → caller-owned)과 #4921/#4676/#4935가 이 경로를 안 건드렸다는 확인은 본문에 있다. 지금 checks는 대부분 pending.

메인테이너의 판단이 필요한 지점
- MAINTAINERS.md가 요구하는 auth/credential selection 보안 리뷰를 누가 sign-off할지(요청 리뷰어 Ingwannu 포함).
- main 실패가 subagent fallback에 보이는 동작 변화를 릴리즈 노트에 명시할지, #5019 닫힘 코멘트만으로 충분할지.
- account-gated 모델에서 main을 계속 빼는 현재 경계를 유지할지(본문: 로스터를 저장 credential에서만 알 수 있음).
- CI(test 샤드·gates·hygiene) 초록 + 보안 리뷰 완료 후에만 머지할지.

너의 추천
방향은 맞다. #5019의 재현(언핀 + main bearer + 저장 형제 하나 → 항상 저장만)을 펜스를 깨지 않고 고치는 최소 축이고, preview/final 공유 식까지 맞춰 #4850 회귀도 막았다. 다만 credential selection이라 CI 초록만으로 바로 누르지 말고, 명시 보안 리뷰 한 바퀴 후 머지하는 편이 맞다. 머지 시 #5019를 Closes로 닫으면 된다. types.ts/config.ts 대분할에 무효화되지 않으므로 close-don't-rebase 대상이 아니다.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
### CI on this branch cannot produce a test signal, for a reason that predates it

This branch is based on `11bc4f708c`, which was the `dev` tip when it was cut, and that commit
fails `bun run build:gui`:

```
src/i18n/native-main-copy.ts(11,26): error TS7053: Element implicitly has an 'any' type because
expression of type 'LabLocale' can't be used to index type '{ en: ... }'.
  Property 'vi' does not exist on type '{ en: ... }'.
```

It is a union defect between two individually correct merges. `e80e571f63` (#4984) added `vi` to
`LabLocale`, which `Locale` aliases; `11bc4f708c` (#4781) then added
`gui/src/i18n/native-main-translations.ts` as a deliberately "closed, nine-locale namespace" and
indexes it with that ten-member `Locale`. Each was green on its own. The `dev` push run for
`11bc4f708c` is red for exactly this
([35311351776](https://github.com/lidge-jun/opencodex/actions/runs/35311351776)).

Every test leg builds the GUI first, because the suite serves `gui/dist` and reads it back, so
"Test in fresh-process batches" is skipped rather than run. The red checks here are therefore not
evidence about this change in either direction.

#5020 fixed it and is merged, so `dev` at `88249ed750` builds again. I did not rebase this branch,
because integration is the maintainer's call in this lane. To get a real signal I cherry-picked this
commit onto the fixed `dev` on a throwaway branch and dispatched the full lane there; results are
linked in a follow-up comment. Updating this branch onto current `dev` is all this needs.

lidge-jun:
### What the hosted suite found, and what changed because of it

The first version of this change was wrong in a way static reading had missed, and the suite caught
it: three cases in `tests/codex-integration/codex-auth-context.test.ts` failed —
"a fresh request can reuse caller main after the selected Pool account enters cooldown",
"cooldown caller-main fallback never resurrects the cooled subscription", and
"cooldown caller-main fallback follows the stable user id, not the recorded email".

The mechanism is worth stating plainly, because it is the interesting part of this PR. The cooldown
fallback asks `callerIsCooledPoolAccount` of the **selected** account, and that was sufficient only
while a request-owned bearer could not make main a candidate: the cooled account *was* the
selection, so the question and the refusal sat in the same place. Making main an ordinary candidate
moved the cooled account out of the selection and took the refusal with it, so a caller whose own
credential is a cooled Pool subscription was served as main — resurrecting the cooldown it was
inside. That is the one behaviour in this area that must not change, and it is what the second
commit restores.

Eligibility now asks the same question of every cooled sibling, so the caller that owns one keeps
the cooled account as its selection and the existing cooldown machinery produces the same refusal it
produced before, including failing closed on an unreadable caller identity. A distinct member of the
same workspace account is still served, which is what those tests separate.

Preview passes the new term as an explicit `false`. Cooldown identity is not modelled in preview and
never was — `callerIsCooledPoolAccount` has no other caller in the tree — because it decides a
refusal rather than which account serves, so a preview that scores main while the resolution refuses
still hands subagent fallback the right account. It is passed explicitly rather than defaulted so
the asymmetry is visible at both sites.

A typecheck error on the same commit (`codexQuotaScopeForModel` returns `CodexQuotaScope |
undefined`) is fixed in the third commit.

Hosted CI is re-dispatched with `lane=all` at the current head so the nine Windows shards run,
since the `pull_request` event does not schedule them.

lidge-jun:
Three of your own regressions are red at this head, all fast assertions rather than timeouts, so this is the fix not reaching the path the tests drive rather than inherited noise:

- `Codex auth context > a fresh request can reuse caller main after the selected Pool account enters cooldown` (5.13ms)
- `> cooldown caller-main fallback never resurrects the cooled subscription` (5.57ms)
- `> cooldown caller-main fallback follows the stable user id, not the recorded email` (4.82ms)

293 other cases in that shard pass, and `dev` is green on it.

The second and third are the interesting pair: one says the fallback must not resurrect a cooled subscription and the other says it must follow the stable user id rather than the recorded email. If a single change can satisfy the first case but not those two, the selection is probably keying on the wrong identity, which is the same class of defect #4593 fixed for stored Direct account identity. Worth checking whether the bearer you admit and the account you record are resolved through the same key.

Do not relax any of the three to match current behaviour. If one of them states the contract wrongly, say which and why.

lidge-jun:
## Not ready to merge: the exclusion was load-bearing for more than ordering

Hosted CI at `686b18063d` ([run 35315335211](https://github.com/lidge-jun/opencodex/actions/runs/35315335211)):
`gates`, `test 1/4`, `test 2/4`, `test 3/4` and `windows 1/9` pass; `test 4/4` fails on the same
three cooldown cases as before, and the reason is not the one the second commit addressed.

The decisive assertion is `codex-auth-context.test.ts:1351`:

```
expect(cfg.activeCodexAccountId).toBe("pool-a")
Received: "__main__"
```

That is not a tie-break or an ordering complaint. **Selecting main for a request-owned bearer wrote
main into the shared Pool cursor**, which is precisely the thing a request-owned credential must
never do: it exists for one request and owns no durable account state. `poolStateEligible` keeps
such a request out of affinity and lineage, but the strategy cursor is written by the selector
itself, and until now main could not be selected at all — so nothing had to fence that write.

The other two failures say the same thing from the other side: the cooled-subscription refusal is
keyed to the **selected** account, so it stopped protecting anything once main became selectable,
and the eligibility term added in the second commit did not restore it. I did not isolate why —
the likeliest cause is that it asks for the cooldown under `codexQuotaScopeForModel(modelId)` while
the fixture records and reads it under `"shared"` — and I am not going to keep guessing at it
through 25-minute CI rounds.

**So the shape of this fix is wrong, and that is the useful finding.** Excluding a request-owned
bearer from Pool candidacy was doing three jobs, not one: it kept main out of the comparison, out
of the shared cursor, and out of a cooldown-identity check that only ever examined the selection.
A single liveness predicate can only do the first.

What it needs instead is a candidacy that can name main without writing Pool state. The lever
already exists: `previewCodexAccountForRequest` is the non-persisting selector, used today to move
a request off the shared cursor without persisting the move. Ask it whether main wins the ordering
for a request-owned bearer; if it does, serve `resolveCallerOwnedMainContext()`; if it does not,
leave main ineligible for the real selection, so the cursor, the cooldown identity check, the probe
lease and everything else keep behaving exactly as they do today. That confines the change to the
one decision the issue is actually about and writes nothing.

I am leaving this open rather than forcing it green, because making those three tests pass by
relaxing them would delete the invariant they exist for. The first commit's diagnosis of #5019
stands and is confirmed at the tip; the remedy in it does not.

lidge-jun:
Updated read at this head, and one of these is more serious than the others.

**An existing invariant is broken.** `server local API auth > OpenAI option auth matrix keeps direct, pool, and API credentials independent` fails at 491.53ms in `test 1/4`, with 264 other cases in that shard passing. That case is not yours and it is not a timeout. It exists to assert that the three credential classes stay separate, and this change lets a request-owned main bearer take part in pool selection — which is precisely the boundary it guards.

Please treat that as the primary finding rather than as collateral. The question it forces is whether "a caller's main bearer may be selected for a pool request" can be true at all without direct and pool credentials becoming reachable from each other. If it can, the mechanism needs to be explicit about what keeps them independent; if it cannot, the feature needs a different shape. This sits on the credential boundary that `AGENTS.md` puts behind explicit security review, so a change here has to state its argument rather than demonstrate it only by tests passing.

**Your own three regressions are still red**, now on Windows as well as Linux: `a fresh request can reuse caller main after the selected Pool account enters cooldown`, `cooldown caller-main fallback never resurrects the cooled subscription`, and `> follows the stable user id, not the recorded email`. All sub-100ms assertions, so not #4956.

The pairing is suggestive: the two that fail are the ones constraining *which* identity the fallback follows and *what it must not resurrect*. If a single change satisfies the first case but not those two, the selection is probably keying on the wrong identity — the same class #4593 fixed for stored Direct account identity. Worth checking whether the bearer you admit and the account you record resolve through the same key.

Do not relax any of the four. If the independence assertion is the one that is wrong, that is a much larger claim and needs to be argued explicitly.

Ingwannu (changes requested):
I verified the selection change at exact head `686b18063d`. The report is real and this direction is close, but the new caller-main winner loses pool recovery.

When the request-owned main bearer wins ordering, `resolveCodexAuthContext` returns `kind: "main"` with no account id. Downstream alternate-account recovery is gated by `usesCodexForwardPoolAuth`, which accepts only `pool` and `main-pool`. A pre-stream 429/402 or eligible transient failure from that caller bearer therefore returns directly even when a stored sibling is available. Preserve enough request-scoped selection origin to retry with `__main__` excluded, while still refusing durable health/affinity ownership for the caller credential.

Also fix the current strict-type mismatch around an absent quota scope and strengthen the regression to compare the exact caller bearer sent upstream, not only negative non-equality assertions. This is an authentication boundary, so keep it unmerged until those paths, focused recovery tests, exact-head CI, and explicit security review are green.

lidge-jun:
Refreshed this branch onto the current `dev` (`f02f3613be`) and pushed the merge as `3af080fb4c`. The prior run was 32 commits behind; the refresh did not change the outcome, so this is not the stale-base class.

Three shards fail, and every failure is the same sentence: something that used to resolve to a pool account now resolves to `__main__`.

- `tests/codex-integration/codex-auth-context.test.ts:1351` expects `pool-a` and receives `__main__`; `:1399` and `:1446` expect a rejection and get a resolved promise.
- `tests/server/server-auth.test.ts`, `expectFileMainUnchanged` at `:1869` via `:1885`, expects `undefined` and receives `__main__`.
- `tests/codex-integration/main-account-hard-lock-auth.test.ts:384` expects a resolve and gets a rejection.

That is the title of this PR happening. Letting a request-owned main bearer take part in pool ordering means it can win an ordering that several existing contracts assume it can never enter, including the hard-lock path, which is a security-relevant one rather than a fixture detail. The failures are therefore the interesting output of the change, not an obstacle in front of it: each one names a place where the intended behavior has to be stated deliberately or excluded.

Leaving this open rather than merging. `dev` stays green at `f02f3613be`.

No local suite was run.

PR #5025, opened 2026-09-18
fix(catalog): keep live pins when the supplied baseline omits the slug

luvs01:
## Summary

Follow-up for a CodeRabbit fix that missed the #4941 merge window: the review thread was resolved against commit `89fcc5de6`, but the PR merged at the earlier head `84014036d`, so the fix never reached `dev`. This replays exactly that change onto current `dev`.

When `applyMultiAgentMode` merges a supplied `nativeDefaults` baseline, a native row whose slug is absent from the baseline fell back to the bundled `UPSTREAM_NATIVE_ENTRIES` snapshot — so a bundled "v2" pin overwrote a live "v1" pin the baseline never mentioned. The fallback now only applies when no baseline was supplied at all (`options.nativeDefaults === undefined`); an explicit baseline is authoritative for exactly the slugs it contains, and omitted slugs keep their live pins.

Regression test: `mode default keeps a live pin the supplied baseline does not mention` covers a baseline that lists only `gpt-5.5`, asserting the live `gpt-5.6-sol` "v1" pin survives.

## Verification

- `bun test tests/codex-integration/codex-v2-gate.test.ts` — 131 pass / 1 pre-existing flaky timeout (`mode-only switches translate the limit...`, passes in isolation at 4.75s vs its 5s bound; unrelated management-API surface).

## Review readiness checklist


lidge-jun:

PR #5026, opened 2026-09-18
fix(service): bind service start to the installed environment

luvs01:
## Summary

Follow-up for a CodeRabbit fix that missed the #4951 merge window: the review thread was resolved against commit `7ce6475e9`, but the PR merged at the earlier head `90471ea92`, so the fix never reached `dev`. This replays exactly that change onto current `dev`.

`ocx service start` now runs the same `assertServiceEnvironmentMatchesInstall()` guard `stop` already runs. Without it, starting the service from a shell whose `sqlite_home`/`CODEX_SQLITE_HOME`/`CODEX_HOME` differs from the recorded install silently binds the recorded database while the shell resolves another — splitting native Codex history between two databases. `status` is intentionally left unguarded (read-only diagnostics should still work to inspect the mismatch).

## Verification

- `bun test tests/service/service.test.ts` — 209 pass / 3 skip / 0 fail.

## Review readiness checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Service startup now verifies that the current environment matches the environment used during installation.
  - Prevents the service from accessing a different database when relevant home or storage settings have changed, avoiding split or missing native Codex history.

- **Tests**
  - Added coverage to confirm environment validation occurs before the service starts.


lidge-jun:
Red on `file-size ratchet: repository` — this change grows a file that is already at its cap.

The cap cannot be raised: `updateBaseline` uses `Math.min`, so a baseline only ever moves downward, and that is deliberate. 39 of the 51 tracked files sit at exactly their cap right now, so this is not bad luck; it is the normal state of this repository.

The remedy is a move, not a number. Put the new case in a sibling file, byte for byte, and register it in **both** `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`. `d3ca5522db` is the original precedent and #5011 and #5018 are recent ones.

One caution from tonight: a moved case is not automatically the same case. One moved out of `codex-v2-gate.test.ts` failed in isolation because its final assertion was reading state earlier cases had warmed rather than the contract. If yours changes colour after the move, suspect the case before the move.

The job log names the offending path; `AGENTS.md` now documents this class as well.

Ingwannu (commented):
I traced the two-line runtime change at exact head `7ce5a6322d`. Reusing `assertServiceEnvironmentMatchesInstall()` before `ops.start()` is the correct narrow follow-up: `start` launches the recorded service environment, so allowing the caller's SQLite/Codex home to disagree would split the operator's history view. Keeping read-only `status` available is also the right exception.

I found no additional source-level correctness blocker in this diff, but I am not approving a red head: hosted `test 2/4` is currently failed and macOS is still running. Please inspect the completed failure log, fix or demonstrate a baseline-only failure, and clean the accidental indentation drift in the existing source-invariant assertions before re-requesting approval.

lidge-jun:
Pushed a follow-up commit to this branch (`e71d6819be`) so it can go green.

`test 2/4` was the file-size ratchet, not the service change: `tests/service/service.test.ts` is at its committed cap of 4106 in `tests/fixtures/file-size-baseline.json` and this branch took it to 4111. Caps only move down, so the remedy is a move rather than a raise. The start-case ordering assertion now lives in `tests/service/service-start-environment.test.ts`, which also pins that `assertServiceEnvironmentMatchesInstall` is imported — a source oracle that only checked for the call could be satisfied by a local stub of the same name.

The same commit restores three lines this branch had reindented from four spaces to three. They are the `repair`/`restart` assertions and had nothing to do with the change; it looked like editor damage rather than intent.

The production change itself is unmodified, and it is the right one: `stop` already refused a moved `CODEX_SQLITE_HOME`/`CODEX_HOME` and `start` did not, which is how native Codex history could end up split across two databases with neither side reporting anything wrong.

Issue #4857, opened 2026-09-17
[Bug][claude-surface] message_start reports usage.input_tokens: 0 — third-party clients reading the first frame under-report context usage (Paseo)


ygc3817922006-sketch:
### Client or integration

Claude Code

### Area

Streaming

### Summary

The Anthropic Messages stream emitted by opencodex reports `usage: { input_tokens: 0, output_tokens: 0 }` in the `message_start` frame. The real input size (including cache read/creation) only arrives in the terminal `message_delta.usage`.

Real Anthropic populates `message_start.message.usage.input_tokens` with the actual prompt size at the start of the turn. Clients that follow the documented contract and read usage from the first frame — e.g. Paseo's context-window meter (`packages/server/src/server/agent/providers/claude/agent.ts` in getpaseo/paseo: `readStreamRequestInputTokens` reads `message_start`, `readStreamRequestOutputTokens` reads `message_delta`, and the sum drives `contextWindowUsedTokens`) — therefore see only the output tokens of the previous turn and show a nearly-empty context ring (e.g. "a few hundred tokens used" of 1M) while `/context` in Claude Code correctly shows ~97k.

Claude Code itself does not rely on the first frame, so compaction and cost accounting are unaffected — this is a display-contract gap for third-party clients only.

### Reproduction

1. Start opencodex (`ocx start`) with any provider behind the Claude surface (reproduced with an `openai-chat` adapter provider; version 2.56.0, macOS, bun runtime).
2. Send a small streaming request directly through the proxy:

```bash
curl -sN http://127.0.0.1:10100/v1/messages \
  -H "Authorization: Bearer <token>" \
  -H "anthropic-version: 2023-06-01" \
  -H "Content-Type: application/json" \
  -d '{"model":"<provider>/<model>","max_tokens":16,"stream":true,"messages":[{"role":"user","content":"reply with one word"}]}'
```

Observed:

```
event: message_start
data: {"type":"message_start","message":{...,"usage":{"input_tokens":0,"output_tokens":0}}}
...
event: message_delta
data: {"type":"message_delta","delta":{...},"usage":{"input_tokens":17,"output_tokens":147,"cache_read_input_tokens":0,"cache_creation_input_tokens":0}}
```

Expected (real Anthropic): `message_start.message.usage.input_tokens` already carries the full prompt size for this turn (and `cache_read_input_tokens` / `cache_creation_input_tokens` when caching applies).

[...]

PR #5027, opened 2026-09-18
fix(quota): drop the whole poisoned window tuple on invalid policy percent

luvs01:
## Summary

Follow-up for a CodeRabbit fix that missed the #4955 merge window: the review thread was resolved against commit `ab08fa131`, but the PR merged at the earlier head `32398e3c6`, so the fix never reached `dev`. This replays exactly that change onto current `dev`, rebased onto the landed `withoutInvalidPolicyPercentHeaders` helper.

The landed helper removed only the invalid percent header before re-parsing for the policy merge — but a surviving `window-minutes`/`reset-at` still minted a percent-less short tuple, which reads as an unknown short window and suppresses a fresh weekly block. An invalid policy percent now drops its whole window tuple (percent + reset-at + window-minutes), so a poisoned short window can no longer mask the valid weekly reading.

Regression test: `an invalid short header still merges the fresh weekly policy reading` proves a seeded expired short + fresh weekly 99 still lands `blocked` when the short percent header is corrupt.

## Verification

- `bun test tests/codex-integration/main-quota-evidence-validation.test.ts` — 145 pass / 0 fail.

## Review readiness checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved quota handling when one usage percentage is invalid, while preserving valid quota windows.
  - Ensured invalid quota data does not prevent fresh weekly limits from being applied.
  - Account availability now reflects current upstream quota information more accurately, including transitioning to a blocked state when limits are reached.


lidge-jun:
45 failures at this head, all in `declared short-window producer evidence`, and they look like your own change rather than inherited noise — they are sub-millisecond assertions, not the timeout pattern in #4956, and `dev` is green on that shard.

The shape is consistent across the sample: `header primary invalid <value> stays unknown or retains short99 until valid zero`, for invalid inputs including `-1` and `-0.01`. That contract says an invalid header must leave the prior state alone — stay unknown, or retain the existing short-window value — and only a valid zero may clear it.

Dropping the whole poisoned window tuple is a bigger hammer than that contract allows if it also discards a previously valid short-window value when the new evidence is merely unparseable. Invalid and zero are different: one is absence of information, the other is information. That distinction is the same one #4857 is open about on a different surface, and it is worth getting right here.

Do not relax these to match the new behaviour. If you believe the contract itself is wrong — that a poisoned tuple should discard a prior good value — say so explicitly and argue it, because that is a deliberate change to a documented invariant rather than a test fix.

Ingwannu (commented):
The tuple-level correction at exact head `011f104727` is directionally correct. Removing only an invalid percent leaves reset/window headers capable of recreating a percent-less short tuple, so deleting the whole corresponding tuple before the policy reparse is the necessary boundary; valid sibling windows remain available for the hard-lock merge.

I found no new source-level blocker in the focused diff, but this head is not approvable while hosted `test 3/4` and `macos 1/2` are failed and other checks are still pending. Please read the final logs and either fix the exact-head failures or provide a baseline comparison for each before requesting approval.

luvs01:
You're right — the contract is atomic at the observation level, and my change was the wrong granularity. Re-reading the failing cases: when any percent header is invalid the whole header observation is untrusted, so a valid zero on a sibling slot must NOT apply either, and the prior state (unknown or a retained short-window value) survives until a valid zero clears it. Dropping the poisoned window tuple let exactly that sibling zero through, which is the violation the 45 failures pin down.

I don't have a case for changing the contract itself. A response with a corrupt field is suspect wholesale — treating the surviving fields as clean evidence is the stronger assumption, not the safer one. The reverted branch now restores `quota.ts` and the evidence-validation test to `dev` (`32350392e`), so nothing of mine remains to land. Closing this PR; the CodeRabbit finding it carried was premised on the same misread of the semantics.

Ingwannu:
Rechecked after the 45 focused failures and the revert. My earlier review at `011f104727` was wrong: I treated invalidity as tuple-local, but this contract is atomic at the observation level. If any percent field is corrupt, accepting a sibling zero would let a partially untrusted response clear previously valid state. Invalid evidence must preserve the prior observation; only a fully valid zero may clear it.

Closing with no diff is the correct disposition. Thanks for reverting rather than weakening the tests to match the patch.

Issue #4179, opened 2026-09-10
Support self-contained Docker Compose builds from a remote Git context


Skyline-23:
### Area

Installation or packaging

### What are you trying to accomplish?

Deploy and update OpenCodex through a deployment dashboard's Raw Compose configuration, using the official repository as a remote Git build context and `pull_policy: build`.

An operator should be able to press Deploy without maintaining a custom Dockerfile, embedding `dockerfile_inline`, creating a deployment fork, or manually preparing a source checkout on the server. Docker/BuildKit should handle the build from the selected official Git ref.

### What prevents this today?

The official source-build support exists, but its build context must be prepared outside Docker first:

- The [Dockerfile](https://github.com/lidge-jun/opencodex/blob/2f3f736299dca38861f8fb9c4326a4b4d7c664bc/Dockerfile#L9-L11) verifies a pre-existing compatibility manifest before building.
- It also [copies `src/generated/compatibility-version.json` from the build context](https://github.com/lidge-jun/opencodex/blob/2f3f736299dca38861f8fb9c4326a4b4d7c664bc/Dockerfile#L45-L47).
- That generated file is intentionally [untracked](https://github.com/lidge-jun/opencodex/blob/2f3f736299dca38861f8fb9c4326a4b4d7c664bc/.gitignore#L7-L9), and the documented workflow requires running `bun scripts/generate-compatibility-version.ts` on the host before `docker compose build`.

Consequently, a clean remote Git context does not contain an input required by the Dockerfile. The operator needs host-side preparation or a custom build recipe even though the repository already provides an official Dockerfile. `pull_policy: build` does not supply that missing preparation step.

This limitation was checked against v2.49.0 / main commit `2f3f736299dca38861f8fb9c4326a4b4d7c664bc`.

### What should OpenCodex do?

Make the official Docker build self-contained, so it can prepare the required compatibility manifest as part of the build from a clean supported Git context.

Expected outcomes:

- A clean remote Git-context build succeeds without installing Bun on the host or running a separate preparation command.
- Operators can use the official Dockerfile through Raw Compose without maintaining an additional file, inline Dockerfile, or fork.
- Redeploying with `pull_policy: build` builds the selected Git ref and retains the existing persistent-state contract.
- Compatibility identity remains derived from the correct source snapshot, and existing source-integrity checks remain enforced. Removing validation or substituting a placeholder manifest would not satisfy the request.
- The documentation includes a supported remote-context Compose example and any required BuildKit options.

[...]

PR #5021, opened 2026-09-18
fix(responses): let a provider declare the hosted tools it rejects

lidge-jun:
## Summary

An OpenAI-compatible Responses gateway does not necessarily accept every capability OpenAI accepts, and until now it had no way to say so. The only mechanism was `UNSUPPORTED_HOSTED_TOOLS` in `src/responses/hosted-tool-policy.ts`, a table of `(model, baseUrl)` predicates, so supporting a narrower destination meant shipping a proxy release that named its endpoint.

The reported destination accepts plain Responses requests and `function` tools but rejects hosted `web_search` with HTTP 400 `unsupported_request`. Codex forwards its hosted declaration with every request, so even `Reply exactly with OK` failed before the model answered:

```
ERROR: { "error": { "message": "The request uses an invalid or unsupported capability.",
                    "type": "invalid_request_error", "code": "unsupported_request" } }
```

The only workaround was `web_search="disabled"`, which turns hosted web search off for every provider including the ones that support it.

This adds `unsupportedHostedTools` to the provider config. `stripUnsupportedHostedTools` now consults it alongside the built-in table and removes denied declarations from `tools`, from client-loaded `additional_tools`, and from `tool_choice` before serialization. The reporter's own suggested configuration now works as written, on a provider the registry has never heard of:

```json
{ "providers": { "agent-space": {
    "adapter": "openai-responses",
    "supportsResponsesCustomTools": false,
    "unsupportedHostedTools": ["web_search", "web_search_preview"]
} } }
```

**The declaration is additive, not a replacement.** The built-in table still covers destinations that reject a tool regardless of configuration, so an operator who never heard of the field stays protected, and a declaration can only deny more — it can never re-enable a known-broken pairing such as `grok-4.6` on OpenCode Go.

**Two deliberate design choices**, both called out because they are the kind of thing that is easier to review than to rediscover:

- *Spelling variants of one capability are aliased.* Declaring `web_search` also denies `web_search_preview`. The rest of the proxy already folds that pair into a single tool — the parser maps both to one name, Chat ingress accepts them together, and the canonical-field strip lists them in one `toolTypes` set. Honouring only the spelling the operator happened to write would reproduce the original 400 while the config claimed to have prevented it.
[...]

PR #5030, opened 2026-09-18
fix(docker): build the compatibility manifest inside the image

lidge-jun:
## Summary

The official Docker build is now self-contained: it builds from a clean Git context, including a remote one, with no host Bun and no preparation command. `pull_policy: build` against `context: https://github.com/lidge-jun/opencodex.git#main` works as the issue described.

A build-only `manifest` stage produces the canonical compatibility manifest and accepts exactly two inputs, in order:

1. **A manifest the host already generated.** The pre-existing workflow still wins, and it is verified rather than silently replaced, so a prepared checkout keeps building byte for byte as before.
2. **A clean Git context.** The generator's canonical file list comes from `git ls-files`, so the stage produces the same artifact from the selected Git snapshot.

A context with neither fails the build with a message naming both inputs. It never falls back to a placeholder — that would defeat the identity the runtime check exists to prove. Both verification passes are preserved: the read-only context is still verified before any `COPY` can dereference a source symlink, and the copied runtime files are still verified inside the image.

**On admitting Git metadata to the context.** `git ls-files` reads the index and never opens an object or a ref, so `.dockerignore` admits only `.git/index` and `.git/HEAD`. That matters: allowlisting `.git` wholesale would put this repository's 1.3 GB object store into every local `docker compose build` context, where the two files together are about 1 MB. Verified against this repository:

- `.git/index` 1,129,638 bytes, `.git/HEAD` 20 bytes.
- With only those two files plus empty `objects/` and `refs/` directories, `git ls-files -- src package.json bun.lock scripts/model-metadata.source.json` returns the full 1,273-entry inventory.

Docker cannot carry empty directories in a context, so the manifest stage copies the two files into a scratch `GIT_DIR` it owns and creates `objects/` and `refs/` there. That also keeps the read-only bind mount pristine and sidesteps the dubious-ownership refusal a context-owned `.git` would trigger for a root build process.

Git itself is installed only in the `manifest` stage and never reaches the `build` or `runtime` layers; the pinned Bun image does not ship it. No `COPY` anywhere in the Dockerfile names `.git`, and the test now asserts that by scanning every `COPY` line.

Supporting changes: `generateCompatibilityVersionManifest` and `verifyCompatibilitySnapshot` take an optional root and output/manifest path so the stage can write outside the read-only mount — both keep their existing no-argument behaviour for `prepare:package` and the in-image check. `compose.yaml` gains `pull_policy: build` and the `BUILDKIT_CONTEXT_KEEP_GIT_DIR` build argument, which a remote Git context needs and a local clone ignores. `scripts/ci/docker-smoke.ts` now removes any developer-left manifest before building, so the self-contained path is the one CI exercises, and it still restores the host artifact afterwards.

`tests/service/container-bootstrap.test.ts` encoded the old contract — it asserted that no `.dockerignore` line re-includes `.git`, and that the runtime stage copies the host manifest directly. Both are now false by design, so the test was rewritten to the new contract rather than relaxed: it pins the four-line `.git` block verbatim, rejects four broader variants that would reintroduce the object store, requires both manifest inputs and the failing third case, and keeps every existing runtime assertion. Its negative verifier cases (missing, stale, hash-mismatched, extra-source, symlink) are untouched.

## Verification

- **No local verification was run.** This lane is forbidden from running any local suite, focused test, typecheck, build, install, Docker, or the `ocx` binary, because an earlier local run deleted a real `~/.opencodex` directory. No image was built here.
- The one empirical claim above was measured with read-only Git commands outside Docker: a scratch directory containing only `.git/index`, `.git/HEAD` and empty `objects/` and `refs/` produced the complete `git ls-files` inventory. That is the assumption the whole design rests on, so it was checked rather than assumed.
- The `docker-smoke` job on `ubuntu-latest` is the real verifier and it runs on this PR, since a Dockerfile change trips the `ci` path filter. It now builds from a clean context with no host manifest, so a green run is direct evidence for the issue's primary request.

Risks only hosted CI can settle, stated plainly rather than worked around:

- Whether Compose passes `BUILDKIT_CONTEXT_KEEP_GIT_DIR` through to the BuildKit frontend on the Compose version CI runs. `docker-smoke` builds from a local checkout, which does not exercise that argument; a genuinely remote-URL context case would need its own CI step.
- Whether `apt-get install git` succeeds on the pinned digest for both platforms of the image index.
- Building from a Git **worktree** hits the failure branch, because `.git` is a file there rather than a directory. The error message names both supported inputs, so the fix is to generate the manifest on the host as before.

## Checklist

- [x] Docs or release notes were updated when needed. README and all eight Remote Hub locale pages drop the host Bun prerequisite and gain the remote-context example; the English pages also state the narrow Git-metadata contract. `structure/ops/docs-and-release.md` records the build-only manifest stage and the two verification passes.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. The context admits no `.git/config`, no refs and no objects, so no remote URL or embedded credential from a developer's Git configuration enters the build. Git is confined to a build-only stage, the runtime stage is unchanged, and no verification was removed or weakened.

Closes #4179

## Summary by CodeRabbit

- **New Features**
  - Docker builds now generate and verify the compatibility manifest automatically.
  - Remote Git build contexts are supported through Docker Compose and BuildKit configuration.
  - Existing pre-generated manifests remain supported when they pass validation.

- **Bug Fixes**
  - Builds now fail clearly when required manifest or Git metadata is unavailable.
  - Git metadata is excluded from image layers, improving container isolation.

- **Documentation**
  - Remote Hub deployment guides were updated across supported languages with the simplified build workflow.


lidge-jun:
## 리뷰 · 우선순위 76 / 80

지금 `dev` 끝은 `baae9057b` (#5023 catalog contextWindow 백필)이고, 그 아래 #5021 unsupportedHostedTools · #5020 vi locale · #4781 native-main이 깔려 있다. 이 PR은 그 제품 레인과 겹치지 않는 패키징 구멍이다. 이슈 #4179가 말한 증상은 그대로다. 공식 Dockerfile은 호스트에서 만든 `src/generated/compatibility-version.json`을 검증·복사하는데, 그 파일은 의도적으로 untracked라서 `context: https://github.com/lidge-jun/opencodex.git#main` 같은 깨끗한 원격 Git 컨텍스트에는 없다. Dokploy Raw Compose에서 `pull_policy: build`만 누르면 Deploy가 깨지고, 운영자는 Bun을 서버에 깔거나 포크·inline Dockerfile을 유지해야 했다. 지금 `dev`의 Dockerfile도 같은 계약이다(호스트 generate 주석 + 런타임 `COPY` of host manifest).

이 PR은 build-only `manifest` 스테이지를 앞에 두고 입력을 두 개만 받는다. (1) 호스트가 이미 만든 매니페스트가 있으면 예전처럼 검증만 하고 그대로 쓴다(조용히 갈아끼우지 않음). (2) 없으면 `.git/index` + `.git/HEAD`만 있는 깨끗한 Git 컨텍스트에서 `generate-compatibility-version.ts`로 만들고, 읽기 전용 bind 마운트 밖 `/manifest/...`에 쓴 뒤 `verify-compatibility.ts`로 다시 검사한다. 둘 다 없으면 `exit 1`이고 placeholder는 없다. 런타임 identity 검사를 지키는 방향이 맞다. Git 바이너리는 핀된 Bun 이미지에 없어서 `apt-get`으로 `manifest` 스테이지에만 깔리고, build/runtime 레이어에는 안 간다. `COPY` 줄에 `.git`이 없다는 것도 테스트가 스캔한다.

`.dockerignore`가 `.git` 전체를 넣지 않고 index/HEAD만 허용하는 설계가 핵심이다. `git ls-files`는 인덱스를 읽고 object/ref는 안 연다. 이 리포 기준으로 객체 저장소는 약 1.3GB인데 두 파일은 약 1MB다. 빈 `objects/`·`refs/`는 Docker 컨텍스트가 못 실어 오니 스테이지가 scratch `GIT_DIR`에 복사한 뒤 만든다. 읽기 전용 마운트를 더럽히지 않고, 컨텍스트 소유 `.git`이 root 빌드에서 거는 dubious-ownership도 피한다. `generateCompatibilityVersionManifest` / `verifyCompatibilitySnapshot`에 optional root·출력/매니페스트 경로를 붙였고, 인자 없는 예전 동작(`prepare:package`, 이미지 안 검사)은 유지된다. `compose.yaml`에 `pull_policy: build`와 `BUILDKIT_CONTEXT_KEEP_GIT_DIR=1`이 들어간다(원격 컨텍스트용, 로컬 클론엔 무해). `scripts/ci/docker-smoke.ts`는 빌드 전에 호스트 매니페스트를 지워서 self-contained 경로를 CI가 실제로 밟게 하고, 끝나면 복구한다. `tests/service/container-bootstrap.test.ts`는 예전 계약(`.git` 재포함 금지 · 런타임이 호스트 매니페스트를 직접 COPY)을 느슨하게 푼 게 아니라 네 줄 `.git` 블록 고정 · 더 넓은 네 변형 거부 · 두 입력+실패 분기 · Git이 이미지 레이어에 안 들어가는지로 다시 잠갔다. README와 remote-hub 8개 locale, `structure/ops/docs-and-release.md`도 맞춰 두었다. 로컬 스위트/도커는 안 돌렸다고 했고(이전 `~/.opencodex` 삭제 사고), hosted `docker-smoke`가 검증 경로다.

라인 - 이게 무슨 문제다

Dockerfile (host manifest 분기, `[ -e "$context_manifest" ] || [ -L ... ]`) - 호스트에 남아 있는 예전 `compatibility-version.json`이 있으면 Git 경로보다 항상 이긴다. `docker-smoke`는 지우고 빌드하니 CI는 self-contained를 본다. 다만 로컬 `docker compose build`나 대시보드가 컨텍스트에 stale 파일을 실어 주면, 지금 체크아웃과 다른 identity가 조용히 들어갈 수 있다. 의도(준비된 체크아웃 byte-for-byte)면 OK. “둘 다 있으면 Git을 우선”은 별 정책이다.

Dockerfile / `.dockerignore` (Git worktree) - worktree는 `.git`이 디렉터리가 아니라 파일이라 index/HEAD 분기에 못 들어가고 실패 메시지로 떨어진다. PR 본문에 적혀 있고 호스트 generate로 돌아가면 된다. 기여자 worktree 빌드가 흔한 팀이면 docs에 한 줄 더 있으면 덜 헷갈린다.

`scripts/ci/docker-smoke.ts` - 로컬 체크아웃 빌드라 `BUILDKIT_CONTEXT_KEEP_GIT_DIR`·진짜 `context: https://...git#ref`는 안 밟는다. PR도 위험을 솔직히 적었다. #4179의 본요청(원격 URL Compose)은 초록 smoke만으로는 간접 증거다. Compose 버전이 그 ARG를 BuildKit frontend에 넘기는지는 별 확인이 남는다.

Dockerfile `apt-get install ... git` - 패키지 버전이 핀되지 않았다. 베이스 digest는 고정돼 있어도 apt 인덱스가 바뀌면 manifest 스테이지의 git 바이너리 해시가 빌드마다 달라질 수 있다(런타임 레이어엔 안 남음). 재현성보다 성공률이 목표면 OK. 핀이 필요하면 후속.

`scripts/generate-compatibility-version.ts` CLI (`process.argv[2]`, `process.argv[3]`) - 무인자 동작은 유지된다. 실수로 한 인자만 주면 그게 `repoRoot`로 해석된다. Docker RUN과 `prepare:package` 경로는 괜찮고, 문서화만 맞으면 된다.

메인테이너의 판단이 필요한 지점
- #4179를 이 PR만으로 Closes 할지, 아니면 “원격 URL Compose 전용 CI 스텝이 초록일 때”로 이슈를 열어둘지.
- 호스트 매니페스트와 Git index가 둘 다 있을 때 호스트 우선을 유지할지, 재현성을 위해 Git 우선으로 바꿀지.
- worktree 실패를 docs에 명시할지, 지원 범위 밖으로 둘지.
- apt의 `git`을 버전 핀할지(런타임에 안 남으니 보통 불필요).

너의 추천
`docker-smoke` · hygiene · 나머지 CI가 초록이면 머지하고 #4179를 Closes로 닫아라. 원격 URL 사례는 머지 후 한 번 Dokploy/수동으로 `context: https://github.com/lidge-jun/opencodex.git#dev`를 확인하거나, 여유 있으면 smoke에 remote-context 잡 하나를 후속으로 두면 #4179 본요청이 문서가 아니라 파이프라인으로 잠긴다. 코드 방향(매니페스트를 이미지 안에서 만들고 Git은 build-only, 검증 두 번 유지)은 `dev` 계약과 잘 맞다.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
Two reds at this head and they are worth separating, because one of them is direct evidence about the thing this PR changes.

**`docker smoke` is the important one.** It fails at "verify shipped config and seed loopback-only fixture: invalid seeded config evidence". That job builds the image, so on this branch it is building from the new clean-Git-context path rather than from a host-generated manifest — which is exactly the case the issue asked for and exactly what a green run here would have proven. A red tells you the manifest stage is producing something the runtime verification does not accept.

Worth checking first, given how the stage is built: does the scratch `GIT_DIR` assembled from `.git/index` and `.git/HEAD` yield the same inventory inside the image as it did in your out-of-Docker check? You verified 1,273 entries outside Docker. If `git ls-files` sees fewer or none inside the build, the manifest is well-formed but empty or partial, and "invalid seeded config evidence" is the downstream symptom rather than the cause.

**`test 1/4` is the second red** and needs its own look rather than being folded into the first.

On the `.dockerignore` narrowing: measuring the wholesale `.git` allowlist against this repository's 1.3 GB object store, and then cutting it to `.git/index` plus `.git/HEAD` because `git ls-files` reads the index and never opens an object or a ref, is the right instinct and the right evidence. Please keep that reasoning in the final description even if the stage changes shape.

Rewriting `tests/service/container-bootstrap.test.ts` was also correct rather than convenient: it asserted that no dockerignore line re-includes `.git` and that runtime copies the host manifest, and both are false by design now. Pinning the four-line block verbatim and rejecting four broader variants keeps it a real guard rather than a weakened one.

lidge-jun:
Merging. `docker smoke` is green at this head, and on this branch that job builds the image from the clean Git context rather than from a host-generated manifest — so it is direct evidence for what the issue asked for rather than a general health check.

Two corrections in this round were the right calls and I want them recorded.

Removing `pull_policy: build` from the shipped `compose.yaml` was not just a way past the red. It forced a rebuild on every `docker compose up` and `run`, which is an unrequested behaviour change for everyone, and our compose already builds from its `build:` section when the image is absent. Keeping it in the documented remote-context example is where the issue actually wanted it.

Narrowing the `.dockerignore` allowlist from `.git` wholesale to `.git/index` plus `.git/HEAD` is the detail that makes this safe to ship. A wholesale allowlist would have put this repository's 1.3 GB object store into every local build context; measuring that rather than assuming it, and then establishing that `git ls-files` reads the index and never opens an object or a ref, is what justified the cut.

And resyncing all seven translated READMEs after the English Docker section changed was work the readme-parity gate correctly demanded — 39 structural tokens each, with the manifest `sourceSha256` verified independently rather than taken on trust.

Issue #4207, opened 2026-09-10
Connected catalog reports success while local Codex CLI rejects unsupported reasoning levels


lidge-jun:
### Client or integration

Codex CLI

### Area

Catalog / models

### Summary

A remote client reports `connected`, has a present/synced catalog, and can authenticate to the hub, but its installed Codex CLI exits before making a request because the downloaded catalog contains the unsupported reasoning level `max`.

The connection completion state does not distinguish “hub and credential work” from “the selected local Codex runtime can consume this catalog.” This is especially confusing because `ocx status` recognizes the older runtime and reports an active effort clamp while the materialized client catalog still fails to parse.

### Reproduction

1. Use Windows with Codex CLI 0.135.0 and OpenCodex client 2.50.0.
2. Connect to a hub whose exported model catalog includes `max`/`ultra` reasoning levels. The observed hub is dev 2.51.0.
3. Confirm `ocx connect status --json` reports `connected`, catalog present, token owned, and clean rotation.
4. Ensure `OPENCODEX_API_AUTH_TOKEN` is present in the child process; the audit explicitly loaded the owned client credential without printing it.
5. Launch `codex exec --ephemeral --skip-git-repo-check --color never "Reply with OK."` in a scratch directory.

Observed: exit 1 before inference with the catalog parse error below. Independently authenticated catalog retrieval and routed responses through the hub work. This is not a missing-token or upstream-model failure.

### Version

OpenCodex client 2.50.0; hub 2.51.0 / `6d3ad12e3fbfd8423f1515a467db4d3349732886`; selected Codex CLI 0.135.0.

### Operating system

Windows 11 25H2 client; macOS 27.0 hub.

### Provider and model

Catalog compatibility, before any provider request.

### Logs or error output

```text
failed to parse model_catalog_json path <CODEX_HOME>/opencodex-catalog.json as JSON:
unknown variant `max`, expected one of `none`, `minimal`, `low`, `medium`, `high`, `xhigh`

Fresh probe: exitCode=1, tokenPresent=true, catalogCompatibilityError=true
```

### Screenshots and supporting files

- [`src/client/hub-client.ts:145`](https://github.com/lidge-jun/opencodex/blob/6d3ad12e3fbfd8423f1515a467db4d3349732886/src/client/hub-client.ts#L145) validates container/model/slug shape; it does not validate entries against the consuming local Codex version.
[...]

PR #5029, opened 2026-09-18
feat(claude): opt-in ENABLE_TOOL_SEARCH injection for launched Claude Code sessions

lidge-jun:
## Summary

`ocx claude` now supports `claudeCode.toolSearch`, injected as `ENABLE_TOOL_SEARCH` into the launched Claude Code environment. It is **opt-in and off by default**, which is a deliberate departure from the issue's request — the evidence below is why.

**The reported problem is real and the named variable is the right lever.** Claude Code 2.1.273 disables MCP tool deferral on a non-first-party `ANTHROPIC_BASE_URL`, keyed on the host rather than the model, so it hits every `ocx claude` session. Its own diagnostic states both the gate and the precondition for lifting it:

> `[ToolSearch:optimistic] disabled: ANTHROPIC_BASE_URL=... is not a first-party Anthropic host. Set ENABLE_TOOL_SEARCH=true (or auto / auto:N) if your proxy forwards tool_reference blocks.`

and its settings documentation:

> "Enable it only if your endpoint forwards and accepts the request shape it will receive; when it does not, requests fail with HTTP 400."

**We do not forward those blocks on a translated route, so injecting it unconditionally would not do what it appears to do.** Deferral is a server-side context optimisation, not a smaller request. Claude Code's own deferred-tool predicate requires `input_schema` to still be present alongside `defer_loading: true`, so every schema stays on the wire and it is Anthropic's API that keeps them out of the model's context and answers the `tool_search` server tool with `tool_reference` blocks. On the translated path:

- `src/claude/inbound-content-options.ts` `toolsToResponses` converts any tool carrying `input_schema` into a full Responses function tool and ignores `defer_loading` entirely; the `tool_search_tool_*` server tool falls through to the drop branch.
- `src/claude/compatibility.ts` lists `tool_search`, `tool_reference` and `deferred_tools` in the table whose `true` entries mean *unsupported* — `compatible` is computed as `!FEATURE_CODES.some(code => detected.has(code) && FEATURES[code])`. The issue reads that table as declaring support; it declares the opposite.

So on a routed non-Anthropic model the provider would still receive all 292 schemas, while Claude Code would believe its context was nearly empty and stop compacting — trading a session that compacts often for one that overruns the provider window. An operator who has set `claudeCode.compatibility: "enforce"` gets a hard 400 on every request instead. On the **native Anthropic passthrough** route the body reaches Anthropic untouched and the feature works exactly as reported.

[...]

PR #5038, opened 2026-09-18
test(ci): warm a spawned child module graph before it is timed

lidge-jun:
## Summary

A test that spawns a real Bun child and bounds it with `INTERNAL_DEADLINE_MS` measures two things at once: the behaviour it asserts, and whatever that child had to load before it could run. The second is not a property of the code under test, it is a property of what else ran on the machine first. That is why this reads as flake and is not: whether a file passes depends on its shard's batch composition.

The decay is in the log, and it is per module graph rather than per file. Run 35305115672, Windows 2/9, job 105475642050, `tests/cli/cli-connect-readiness.test.ts`:

```text
(pass) #4207 connected-client readiness > first-time connect escapes ...        2893.60ms
(pass) #4207 connected-client readiness > an installed catalog ...               571.90ms
(fail) connected-client runtime probe scope > observes only the selected ...   15339.29ms
(pass) connected-client runtime probe scope > a rejected preferred runtime ...  2162.84ms
```

Two cold starts in one file. The first describe's first child paid 2.9s against a 0.57s warm baseline. The second describe's first child was killed at 15.3s against a 2.1s warm baseline, because the `observed` ladder additionally loads `src/codex/runtime` and probes generated runtime shims. `spawnSync` returns a null status when its own timeout kills the child, which is the `Expected: 0 / Received: null` that failed the shard while 308 other cases passed. That the two deltas differ by an order of magnitude in one process is the evidence that the cost tracks the graph: a fixed per-spawn overhead would have moved both rows equally. #4948 saw the same shape in `cli-status-json` (15587, 13074, then 1518-1741ms).

The remedy is a shared helper, `tests/helpers/cold-spawn-warmup.ts`, that pays a graph's cold load once per process in a `beforeAll` outside the measured window. No bound under test changes.

**How the warm-up stays honest.** A warm-up that names its modules by hand stops working the first time an import moves, silently, because nothing fails when it warms the wrong thing. So nothing is named by hand. `moduleGraphSpecifiers` reads the child's own source at run time and asks Bun's transpiler which modules it loads, the same `scanImports` pass already used by the import-boundary oracles in `responses-fetch-helpers-boundary` and `api-key-selection-capture`. Type-only imports are erased first, so the scan reports what the child loads rather than what it mentions. Where a child script is built from a template, the import prologue is hoisted to a single constant that the template interpolates and the warm-up scans, so there is one definition rather than a copy. Where the cold cost includes work an import cannot reach, the warm-up replays the file's own child runner with a larger deadline instead, which is what `cli-connect-readiness` and `cli-status-json` do: the warmed path and the measured path are the same call.

**What was rejected, and why.** Raising `INTERNAL_DEADLINE_MS` is refused by `tests/helpers/test-budget.ts`, which records that moving `SPAWN_BUDGET_MS` from 45s to 90s halved the reporting speed of 339 Windows cases to fix one, and that one derivation chain reached 265s. `watchdogMs()` would raise the bound per call site to 45s on Windows CI; `codex-retained-root-serialization` already uses it at one site, which is why that site has never failed this way, but it widens the window a wedged child hides in and still leaves a cold start inside a measured assertion. A runner-level preload in `tests/preload.ts` would need no per-file edit and is rejected on three counts: it would run in all four workers of every shard including those with no spawning test; that file already documents an incident where a spawn added to it timed out, threw out of the preload, and left the real-home guard disarmed for the whole worker; and it cannot know which graph to warm, since the failure above happened in the second graph of a file whose first graph was already warm.

**Coverage.** `tests/ci-workflows/cold-spawn-warmup.test.ts` scans every test file that hands `INTERNAL_DEADLINE_MS` to a child-process timeout and requires each to carry a disposition, so the next one is classified when it lands rather than after it fails on a Windows shard. Six are warmed. `codex-shim.test.ts` is recorded unwarmed with its reason: its Windows children are a cmd.exe or PowerShell driver tree, so the cold cost is shell and process startup rather than a repository module graph and an import scan has nothing to warm, and the file also sits exactly on its 2388-line ratchet cap, which only moves downward.

Files warmed beyond the scan's reach, where the deadline bounds a readiness marker rather than a spawn timeout: `cli-status-json`, `client-connect`, `oauth-refresh-lock-multiprocess`, `codex-history-lock`, `codex-history-worker`, `codex-write-lock`, `codex-retained-root-serialization`, `native-profile-manager`. `native-profile-startup` already handles its cold child explicitly through `COLD_SPAWN_BUDGET_MS`; `native-profile-crash-boundaries`, `codex-prompt-text-probe` and `codex-composed-acceptance` bound an await against an already-running child and are not exposed.

This fixes the cold-start class only. It does not close #4956, which also covers spawned children that produce nothing and never exit, and macOS shards that go silent for sixteen to eighteen minutes. A cold start is slow and finishes; a hang never does, and nothing here would help one.

## Verification

- Local verification was NOT RUN. This lane forbids local suites, focused tests, typecheck, builds, installs and any proxy execution, because a past local run deleted a live `~/.opencodex`. Hosted CI is the executable verification.
- Exact base `baae9057b6` (`dev`). Exact head **`486a2c806f48446ecef259f17c0170be4f63f4e1`**, full `lane=all` dispatch **35320336886**. The `pull_request` event does not run the nine Windows shards, and Windows is where both confirmed instances appeared.
- **The measured rows moved to their warm baseline.** Same file, same platform, before and after:

| `cli-connect-readiness` first row | before (35305115672, win 2/9) | after (35320336886, win 3/9) |
| --- | --- | --- |
| `#4207 connected-client readiness` | 2893.60ms | 1204.02ms |
| `connected-client runtime probe scope` | **15339.29ms, killed** | **2313.04ms, pass** |

  The observed-ladder row is now indistinguishable from its three siblings (2342.90, 2253.01, 2556.50ms), and the two warm-ups that absorbed the cold start are logged outside the window at 2061ms and 2308ms. Run 35328066105 reproduces the flat shape on the same shard: warm-ups at 1575ms and 1753ms, then 1800.77, 1716.50, 1580.64 and 2089.49ms. `cli-models` shows the same: warm-up 2776ms, then every case 376-690ms with no first-row penalty.
- **The scan warmed what it claimed to.** Every warm-up across all nine Windows shards reported a complete load, including all 65 runtime imports of `src/cli/index.ts`: `cli-index/models loaded=65/65`, `main-account-policy-startup-child loaded=12/12`, `client-lifecycle-fixture loaded=9/9`, `client-connect/transaction-eval loaded=6/6`, `bounded-auth-read-child 2/2`, `native-profile-lock-child 2/2`, `codex/quota-provenance-eval 2/2`, and the single-module graphs 1/1. No module failed to import in isolation, so the per-specifier fail-soft path was never taken.
- **A cold start caught in the act.** `oauth-store/eval` warmed in 9691ms on Windows 2/9. Measured inside that file's 15-second readiness deadline it would have consumed two thirds of the bound; it is now paid in a hook with a 45-second budget.
- Run 35318878762 at the previous head `fb9f02885d` failed `test 2/4` and `windows 8/9` identically and for one reason: the scan prepends module context, which moved `src/cli/index.ts`'s shebang off line 1. Fixed by stripping it, covered by a regression case, and both shards are green at the current head.
- **The one remaining red is the other class in #4956, not this change.** `windows 5/9` failed on `main quota policy at native admission > ... (retained-unknown-binding)` at 32514.25ms, SIGTERM, empty stderr. The same case with the same warm-ups in place ran 3724.31ms and passed in run 35318878762 (job 105516876880), where that shard was green; its seventeen siblings in the failing run took 2.8-9.6s. A child that produced nothing for thirty seconds and had to be killed is the hang, not a slow start. A second full `lane=all` dispatch at the identical head, run 35328066105, has all nine Windows shards and all four Linux shards green, with that case back at 3620.05ms. Three runs, one occurrence.
- Everything else at this head is green: all four Linux shards, eight of nine Windows shards, both macOS shards, and every gate and auxiliary job. `macos control` is the separate 30-minute lane tracked in #4905.
- The honesty of the scan is covered by unit tests rather than asserted in prose: that it follows the source when an import moves, that it sees `require` and dynamic import as well as static import, that erased types are not warmed, that builtins are skipped, that a specifier resolves against the directory the child resolves it against, that a shebang does not stop it, and that a real child entry resolves to real repository modules. Failure policy is covered too: a scan that finds no repository module throws rather than silently warming nothing, one warm-up runs per graph per process, and a failed warm-up is not retried.
- Static review confirmed no `INTERNAL_DEADLINE_MS` value, spawn timeout, assertion or test budget was changed, and no test was deleted. The 25s warm-up child deadline is the value #4948 derived by hand, now derived from the hook budget it reserves teardown and reap out of, and pinned by a test.
- `tests/ci-workflows/cold-spawn-warmup.test.ts` is registered in both `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`. No file crosses the 2000-line ratchet threshold and the one capped file is untouched.

## Checklist

- [x] Docs or release notes were updated when needed. The helper carries the rationale and the rejected alternatives; no user-facing behaviour changes, so `docs-site/` is unaffected.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. Tests only. The warm-up child inherits the preload's sandboxed `HOME`/`OPENCODEX_HOME`/`CODEX_HOME` and the armed real-home guard exactly as every other spawned child in the suite does; it logs module paths and elapsed times, never credentials or request bodies. One reordering in `client-connect`'s Desktop fixture was split into two prologues specifically so every module still loads after the synthetic Windows principal and icacls stubs are installed, preserving the original order.


lidge-jun:
## 리뷰 · 우선순위 74 / 80

설명

이 PR은 Windows CI에서 반복되던 `INTERNAL_DEADLINE_MS` 자식 타임아웃 flake의 원인 중 하나인 콜드 모듈 그래프 로드를 측정 구간 밖으로 빼는 테스트 인프라 변경입니다. 지금 `dev`(HEAD `7105998b7`, tip #5025 catalog live-pin 보존)에서는 `tests/helpers/test-budget.ts`가 `SPAWN_BUDGET_MS=45s`, Windows 전용 `COLD_SPAWN_BUDGET_MS=90s`, 내부 마감 `INTERNAL_DEADLINE_MS=15s`를 고정하고, 예산 올리기로 flake를 덮는 길을 명시적으로 거부합니다. 인변 `coldSpawnBudget` / `windowsDeadChild4902` / `ciFailFast4837`과도 같은 방향입니다. 이번 변경은 예산을 올리지 않고, 공유 헬퍼 `tests/helpers/cold-spawn-warmup.ts`로 그래프당 한 번만 미리 로드합니다.

본문이 든 증거는 설득력 있습니다. 같은 파일의 두 describe가 서로 다른 그래프를 쓰므로 첫 자식이 2.9s vs 15.3s처럼 다른 콜드 비용을 내고, 워밍한 뒤 형제는 0.5–2s대로 떨어집니다. 고정 오버헤드라면 두 행이 같이 밀렸을 것입니다. #4948이 `cli-status-json`에서 손수 한 처방을 공유화한 형태이고, 거부한 대안(예산 상향, `watchdogMs` 확대, `tests/preload.ts` 전역 preload) 이유도 `test-budget.ts` 역사·preload 사고·두 번째 그래프 미커버와 맞습니다.

정직성 설계가 핵심입니다. `moduleGraphSpecifiers`는 Bun `scanImports`로 자식 소스에서 실제 로드를 읽고, 타입-only·builtin은 빼고, 스캔이 비면 조용히 통과하지 않고 실패합니다. 템플릿 자식은 import prologue를 한곳에 올려 스캔과 실행이 같은 정의를 씁니다. 스캔이 못 보는 손자 런타임 비용은 `warmColdSpawn`으로 같은 runner를 더 큰 deadline으로 한 번 재현합니다. 커버리지 가드 `tests/ci-workflows/cold-spawn-warmup.test.ts`는 `timeout:…INTERNAL_DEADLINE_MS` 패턴 파일마다 disposition을 강제하고, `codex-shim.test.ts`만 shell/프로세스 기동 비용 + 2388줄 래칫 때문에 unwarmed로 남깁니다. `layout.json` / `test-layout-expected.json` 등록도 있습니다. 제품 코드·예산 상수·단언 값은 건드리지 않았습니다.

한계도 본문이 솔직합니다. #4956은 macOS/Windows에서 자식이 출력을 멈추고 안 죽는 행 클래스까지 포함하며, 콜드 스타트만 고칩니다. `pull_request` 이벤트는 Windows 9샤드를 안 돌리므로, 증거가 난 레인에서의 확인은 merge 후 `lane=all` 또는 workflow_dispatch에 남습니다. 베이스는 `baae9057b6`이고 현재 tip `7105998b7`(#5029→#5028→#5034→#5022→#5025)보다 뒤입니다. 테스트만이라 충돌 확률은 낮지만, 머지 전에 tip 재기지가 필요합니다. types/config 분할 캠페인에 무효화되는 제품 PR이 아닙니다.

tests/helpers/cold-spawn-warmup.ts warmModuleGraph - 격리 import에 실패한 specifier는 warn만 하고 파일은 통과시킵니다. 의도된 열화지만, CI 로그에서 그 줄이 묻히면 다시 콜드 측정으로 돌아갈 수 있습니다.
tests/helpers/cold-spawn-warmup.ts warmColdSpawn 메모 - 그래프 실패 Promise를 재시도하지 않고 공유합니다. 올바르지만, 한 파일의 setup 실패가 같은 키를 쓰는 다음 파일까지 한꺼번에 빨갛게 만듭니다 (history-lock/history-worker 공유 키).
tests/cli/cli-status-json.test.ts beforeAll - Windows 전용 게이트를 제거해 POSIX에서도 warm-up이 돕니다. 메커니즘 훈련에는 좋지만, 드물게 flake 나는 레인에 항상 자식 하나를 더 탑니다.
tests/ci-workflows/cold-spawn-warmup.test.ts DISPOSITIONS - 가드는 spawn timeout 패턴 파일만 봅니다. readiness 마커·다른 마감 표기 파일은 본문에서 손으로 워밍했다고 설명하지만, 패턴 밖 새 파일이 생기면 가드가 자동으로 못 잡습니다.
검증/CI - 로컬 스위트 금지가 명시되어 있고, Windows 샤드는 PR 이벤트에 없습니다. run 35318878762의 시간 비교·`[cold-spawn-warmup]` 로그를 merge 게이트로 봐야 합니다. 코드만으로 초록을 단정할 수 없습니다.

메인테이너의 판단이 필요한 지점

- tip `7105998b7`로 재기지한 뒤 merge할지, 지금 head에서 squash할지.
- #4956을 콜드 스타트만 부분 참조로 두고 hang 클래스를 별 이슈/후속으로 남길지, 아니면 이 PR 머지 후 이슈 본문을 쪼갤지.
- POSIX 전 레인 warm-up 상시 비용을 받아들일지, 다시 Windows 게이트를 둘지.
- `codex-shim.test.ts` unwarmed 상태를 래칫 여유 생길 때까지 유지할지, 별 PR로 shell-cold 처방을 넣을지.
- Windows `lane=all`에서 예전 콜드 행이 warm baseline으로 내려가는지 로그로 확인할 게이트를 필수화할지.

너의 추천

현재 tip으로 rebase한 다음, Windows를 포함한 hosted 검증에서 콜드 행 시간이 워밍 베이스라인으로 떨어지고 예산·단언이 그대로인지 확인되면 merge하세요. 방향이 `test-budget.ts` 불변과 맞고, #4948 처방을 공유·가드·실패 정책까지 한 번에 정리한 유지보수형 PR입니다. #4956 전체를 닫지 말고 Refs만 유지하세요. types/config 분할 때문에 닫을 대상이 아닙니다.

이 댓글은 grok-bot이 작성했습니다

lidge-jun:
One Windows red that lands inside your own scope, so I am holding rather than merging on the exception.

`windows 5/9` (job 105521462945): `main quota policy at native admission > fresh startup restores durable main policy only after owned recovery (retained-unknown-binding)` failed at **32514.25ms**, with 264 other cases in that shard passing.

That file is in your diff — `tests/codex-integration/main-quota-provenance.test.ts` at +18/-3 — so it received the warm-up and still timed out. Two readings, and I cannot tell them apart from here:

Either the warm-up does not cover the graph this particular child loads. Your own analysis established that the cold cost is per module graph rather than per file, and that one file can pay it twice — so a file with a warmed first child can still have a second describe whose child loads something else cold. If that is what this is, the remedy is the same one you already built, applied to the second graph.

Or this case is not the cold-spawn class at all. 32.5s is well past `INTERNAL_DEADLINE_MS`, so whatever bounds it is a different and larger budget, and a genuinely wedged child would look exactly like this.

Worth settling before this lands, because the claim the PR makes is that the class is handled. A file in the diff still failing on the symptom is the one result that would undercut it. If it turns out to be the second reading, say so and the PR stands on its own terms with that case named as out of scope.

Everything else is green at this head: all four Linux shards, `gates`, and eight of nine Windows shards.

Separately: the per-module-graph observation, with two cold starts visible in one file at 2893.60ms and 15339.29ms against 571.90ms and 2162.84ms warm baselines, is a sharper statement of the problem than the one I gave you. Keep it in the final description.

Ingwannu (changes requested):
The cold-graph diagnosis is useful, but exact head `486a2c806f` still has two test-oracle gaps:

1. The disposition guard accepts any source containing `helpers/cold-spawn-warmup`. An unused import, comment, or string literal therefore passes after the real `beforeAll` warm-up call is removed, silently restoring the original flake. Verify an actual `warmColdSpawn`/`warmModuleGraph` invocation with a structured scan or an equivalent syntax-aware assertion.
2. The module-graph scanner must include re-export edges (`export { ... } from`, `export * from`, and namespace re-exports). Otherwise a child can cold-load modules through a barrel that the warm-up never traverses, and the helper will report a misleading complete graph.

Please add regressions proving both failure modes and re-request review only after the dispatched Windows timing evidence and exact-head CI are green. This PR is intentionally about the measurement oracle, so that oracle cannot be string-presence based.

lidge-jun:
Merging. This closes the question I held it on, and the answer is that the red is not this PR's class.

`windows 5/9` failed on `main quota policy at native admission > fresh startup restores durable main policy only after owned recovery (retained-unknown-binding)` at 32514.25ms in run 35320336886. At the identical head `486a2c806f`, run 35328066105 has all nine Windows shards green, including 5/9, plus all four Linux shards and both macOS shards. The same case ran 3724.31ms in 35318878762 and 3620.05ms in 35328066105, and its seventeen siblings in the failing run took 2.8-9.6s.

That shape — a child that produces nothing for thirty seconds and has to be killed, once in three runs, with the warm-up already applied to its graph — is #4956, not a cold start. A cold start is a bounded delay at the front of a measured window; this is the absence of output. #4956 stays open and this PR says so.

On the remedy itself: deriving the module graph from the child's own source at run time, through the same `scanImports` pass the import-boundary oracles already use, is the right call. A hand-written module list is the thing that decays silently, and the evidence that the derivation is honest is that every warm-up across nine Windows shards reported a complete load — `cli-index/models` at 65/65, `main-account-policy-startup-child` 12/12, `client-lifecycle-fixture` 9/9 — so the per-specifier fail-soft path was never reached.

The measurement that convinced me the class is per module graph rather than per file is `cli-connect-readiness`: 2893.60ms and 15339.29ms in one process, an order of magnitude apart, because the second describe loads `src/codex/runtime` and probes generated shims. A fixed per-spawn overhead cannot produce that.

Declining the preload was also right, and for the reason given third: `tests/preload.ts` already records an incident where a spawn added to it timed out, threw out of the preload, and left the real-home guard disarmed for the whole worker. Arming that guard is worth more than any warm-up.

`codex-shim.test.ts` recorded unwarmed with its reason, rather than skipped, is the right disposition — its Windows children are a cmd.exe/PowerShell driver tree, so an import scan has nothing to warm — and `tests/ci-workflows/cold-spawn-warmup.test.ts` forcing a disposition on every file that hands `INTERNAL_DEADLINE_MS` to a child timeout is what keeps the class from reappearing quietly.

PR #5067, opened 2026-09-18
fix(providers): patient same-target 429 fallback for OpenCode Go

agentHits:
## Summary

- Patient same-target 429 fallback for the OpenCode Go destination: single-key pools cannot fail over, so a burst 429 now retries with six 10s-paced attempts instead of surfacing immediately.
- Retry-After is still honored and capped; an explicit `retryOn429` (including `enabled: false`) always overrides the fallback.

## Verification

- `bun x tsc --noEmit` clean.
- `bun test ./tests/providers/rate-limit-retry.test.ts` — 17 pass.

## Checklist

## Review readiness checklist

## Summary by CodeRabbit

- **New Features**
  - OpenCode Go requests now retry rate-limited responses on the same destination, using up to six attempts with 10-second intervals and honoring server-provided retry timing.
  - Retries remain within the request’s shared send budget.

- **Bug Fixes**
  - Prevented OpenCode Go requests from failing immediately on temporary rate limits when no explicit retry policy is configured.

- **Documentation**
  - Documented same-destination rate-limit retry behavior and configuration overrides.


abhisheksharma2411 (commented):
Good fix for a real problem — a single-key pool has nothing to fail over *to*, so surfacing a burst 429 immediately is the one case where same-target patience is clearly right. Two things I checked first, both fine:

**The fail-closed invariant survives the refactor.** Moving the `authMode` gate inside the explicit-policy branch looked at first like it left the new fallback unguarded — that would have let an OAuth row on the Go endpoint replay on the same token, which the original comment says must never happen. Reading the branch rather than the diff, the fallback carries its own copy:

```ts
// No explicit knob: patient fallback for the OpenCode Go destination only.
if (provider.authMode !== undefined && provider.authMode !== "key") return null;
if (!isOpenCodeGoDestination(provider)) return null;
```

Worth keeping those two lines adjacent and commented as one unit, since the duplication is now load-bearing and a future tidy-up that collapses the branches could drop one copy.

**Endpoint identity rather than adapter identity is the right call**, and the comment explaining why (muse-spark rides `openai-responses` while the preset declares `openai-chat`) is the kind of note that survives a refactor.

**One thing worth changing: the repo already has this comparison, and it's stricter in a way that matters.** `src/providers/quota/vendor-probes-key.ts` has a whole family — `isCanonicalOpenRouterBaseUrl`, `isCanonicalDeepSeekBaseUrl`, `isCanonicalClineBaseUrl`, `isCanonicalA6apiBaseUrl`, `isCanonicalOllamaCloudBaseUrl` — all built on one `normalizedBaseUrl` (L918) that parses the URL, lowercases the origin, and rejects userinfo/query/fragment. `isOpenCodeGoDestination` string-compares instead:

```
URL                                          this PR   normalizedBaseUrl
"https://opencode.ai/zen/go/v1"              true      true
"https://opencode.ai/zen/go/v1/"             true      true
"https://OpenCode.ai/zen/go/v1"              false     true      <-
"https://opencode.AI/zen/go/v1"              false     true      <-
"https://opencode.ai/zen/go/v1?x=1"          false     false
"https://user:pass@opencode.ai/zen/go/v1"    false     false
"https://opencode.ai.evil.net/zen/go/v1"     false     false
```

The divergence is **host case**, and the direction is safe — a mixed-case row silently doesn't get the patient retry, so it fails closed rather than open, and the look-alike host is correctly refused by both. But it means an operator who typed `OpenCode.ai` gets today's fail-fast behaviour with no indication why, which is a support ticket that looks like "the fix doesn't work." Reusing `normalizedBaseUrl` (or a small exported twin of it) also keeps the destination-identity question answered in one place rather than two.

Minor, and I'd land it either way. The `destinationAliases` spread is a nice touch — it means a reseller alias picks this up without a second code change.

One question rather than a finding: six attempts at 10s is up to a ~60s hold. The comment says effective replays are bounded by the shared per-request send budget — is that budget always present on this path? I asked the same thing on #4824 and the answer there was a second, budget-independent bound, which turned out to be worth having.

agentHits:
Confirming your budget question with code: yes, on the main dispatch path the second bound is there, one layer down.

In `src/server/responses/adapter-dispatch.ts` the same-target 429 budget (`rateLimitRetries` vs `policy.attempts`) lives outside the recovery loop and is shared with the terminal-guard continuation, so the main loop plus one continuation can never exceed `attempts` same-key replays per request. Underneath that, every send also draws from the shared per-request `adapterDispatchBudget` (`remainingTransientSendBudget`), and exhaustion surfaces as 429 through `SendBudgetExhaustedError` rather than buying another send. So the 6x10s policy caps the wait, and the request budget caps the sends — a long hold ends in a budgeted 429, never in an unbounded replay.

Issue #5066, opened 2026-09-18
[Bug / Config Migration] Deprecated Google Antigravity Flash model aliases trigger continuous migration logs on startup


kge-lab:
### Client or integration

Codex CLI

### Area

CLI

### Summary

On starting the proxy (`ocx start`), multiple deprecation warnings/migrations are logged for `google-antigravity/gemini-3.x-flash` aliases, remapping them to `gemini-3.7-flash`. While the migration executes successfully and the proxy boots, repetitive warnings on every startup suggest these migrated references may not be persisting cleanly in user config or default templates, or require canonical pruning.

### Reproduction

1. Start the proxy with previously cached/configured Google Antigravity settings:
   `ocx start`
2. Observe the multiple `[model-rename-migration]` log messages indicating renames to `gemini-3.7-flash`.
3. Stop the proxy cleanly:
   `ocx stop`
4. Start the proxy again:
   `ocx start`
5. Observe that the exact same batch of `[model-rename-migration]` messages is logged again instead of being suppressed or persisted.

### Version

Latest release (@bitkyc08/opencodex)

### Operating system

Windows 11 (x64 / PowerShell)

### Provider and model

google-antigravity (gemini-3.7-flash)

### Logs or error output

```shell
PS C:\windows\system32> ocx start
[model-rename-migration] renamed "google-antigravity/gemini-3.6-flash" to "gemini-3.7-flash" in the saved config: Google retires the previous Antigravity Flash generation from Cloud Code Assist when its successor ships, so the saved id no longer resolves to a live model.
[model-rename-migration] renamed "google-antigravity/gemini-3.6-flash-low" to "gemini-3.7-flash" in the saved config: Google retires the previous Antigravity Flash generation from Cloud Code Assist when its successor ships, so the saved id no longer resolves to a live model.
[model-rename-migration] renamed "google-antigravity/gemini-3.6-flash-medium" to "gemini-3.7-flash" in the saved config: Google retires the previous Antigravity Flash generation from Cloud Code Assist when its successor ships, so the saved id no longer resolves to a live model.
[model-rename-migration] renamed "google-antigravity/gemini-3.6-flash-high" to "gemini-3.7-flash" in the saved config: Google retires the previous Antigravity Flash generation from Cloud Code Assist when its successor ships, so the saved id no longer resolves to a live model.
[...]

Issue #1610, opened 2026-08-13
[Bug] Registry model renames never reach existing key-auth provider configs (qwen3.8-max-preview persists)


lidge-jun:
### Client or integration

OpenCodex dashboard

### Area

Catalog / models

### Summary

When a registry preset renames or retires a model id, existing key-auth provider configs are never updated. The stale id stays in `config.json` forever, so the dashboard keeps listing a model the vendor is retiring and the real replacement never appears.

Concrete case: `d40367c0c` renamed `qwen3.8-max-preview` to `qwen3.8-max` across both Alibaba Token Plan providers. My `alibaba-token-plan-intl` entry still carried the preview id in **six** places today, and `alibaba-token-plan-intl/qwen3.8-max` was simply absent from the picker.

The cause is the guard in `reconcileOAuthProviders` (`src/oauth/index.ts:917`):

```ts
if (!def || prov.authMode !== "oauth") continue;
```

`OAUTH_RECONCILE_FIELDS` already contains exactly the fields that go stale — `models`, `modelContextWindows`, `modelInputModalities`, `modelReasoningEfforts`, `modelReasoningEffortMap` — and there is already a heal step for a `defaultModel` that no longer exists in the refreshed list. None of it runs for `authKind: "key"` presets, because reconciliation is keyed on `OAUTH_PROVIDERS` and an `oauth` auth mode rather than on "this provider came from a registry preset".

So the repair path exists and is well tested; it is simply unreachable for every key-auth preset (Alibaba Token Plan, Moonshot, Volcengine, and the rest).

### Reproduction

1. Configure `alibaba-token-plan-intl` on a build from before `d40367c0c` (registry seeds `qwen3.8-max-preview`).
2. Upgrade to any build after `d40367c0c` (registry seeds `qwen3.8-max`).
3. Restart the service and open the dashboard model list for `alibaba-token-plan-intl`.

Observed: the list shows `alibaba-token-plan-intl/qwen3.8-max-preview`. `alibaba-token-plan-intl/qwen3.8-max` is not offered.

Expected: the retired id is replaced by the supported one, carrying the registry's current context window, modalities, and reasoning ladder.

Persisted state on my machine before the manual fix:

```
models[0]                              qwen3.8-max-preview
modelContextWindows                    qwen3.8-max-preview: 983616
modelInputModalities                   qwen3.8-max-preview: [text, image]
modelReasoningEfforts                  qwen3.8-max-preview: [low, high, xhigh]
[...]

PR #5077, opened 2026-09-18
fix(providers): stop the Antigravity rename migration re-announcing itself on every boot

lidge-jun:
## Summary

Starting the proxy with a `google-antigravity` provider printed the same nine
`[model-rename-migration] renamed "google-antigravity/gemini-3.x-flash..."` lines on every
single boot (#5066). The migration was not failing to persist; it was losing to the second half
of its own boot.

`startServer` runs `runModelRenameStartupMigration` and then `reconcileOAuthProviders`.
`ANTIGRAVITY_MODEL_CONTEXT_WINDOWS` derives a context window for every compatibility alias, so
the Antigravity OAuth preset carries a `modelContextWindows` key for all nine retired Flash ids
beside the live ones. The migration deleted those keys and wrote the config;
`applyOAuthPresetCatalog` then copied the preset's record straight back over them and wrote the
config again; the next boot found them and said so. That is also why the report shows nine
messages from a user who never selected nine models: the ids came from the registry, not from
anything they had chosen.

The fix is one guard in the projection, the mirror of the `registrySeedsTarget` check already in
that file. A field is skipped only when both of these hold:

1. the registry's own seed (`providerConfigSeed`, the row both the OAuth preset and
   `enrichProviderFromRegistry` contribute) still publishes the retired id in that field, so the
   entry arrives in the config on its own; and
2. the saved field already carries the supported id, so the rename would only delete a key,
   with no value left to carry forward.

The second condition is what keeps it narrow. A row that saved
`modelContextWindows: { "gemini-3.6-flash": 1048576 }` and nothing else still gets that value
moved onto `gemini-3.7-flash`, because dropping it there would blank the record instead of
moving it. The retired registry entries themselves are deliberate and stay put: a request that
still names `gemini-3.6-flash` routes to 3.7 and needs a window under the id it asked for.

Everything the user owns is unchanged. `models`, `selectedModels`, `retainModels`,
`defaultModel` and `disabledModels` are still renamed, and a user who has not started the proxy
since the rename shipped still gets the repair and still sees it reported -- once.

Deliberately not done: no new store, no unconditional silencing of the log, no change to which
aliases are retired, and no change to `ANTIGRAVITY_MODEL_CONTEXT_WINDOWS` itself (those windows
are pinned by `tests/adapters/google/gemini-37-flash-migration.test.ts` for a reason).

Closes #5066

## Verification

No local suite was run: this checkout is under a no-local-execution rule after an earlier local
run destroyed the operator's real `~/.opencodex`. Correctness rests on static tracing plus hosted
CI at the exact head.

- Traced the loop end to end in source: `src/server/index.ts` (migration, then reconcile),
  `src/providers/model-rename-migration.ts`, `src/oauth/index.ts` (`applyOAuthPresetCatalog`,
  `OAUTH_RECONCILE_FIELDS`), `src/providers/antigravity-models.ts`, `src/providers/derive.ts`.
  `modelContextWindows` is the only reconciled field the preset publishes keyed by a retired id,
  which is exactly the nine messages reported.
- Confirmed `enrichProviderFromRegistry` is all-or-nothing for `modelContextWindows` and that
  `loadConfig` does not enrich at all, so the writer between the two boots is reconciliation.
- Confirmed the guard is inert for the two Alibaba renames: `qwen3.8-max-preview` appears nowhere
  in `src/` outside the rename table, so #1610 coverage is untouched.
- Added three tests in `tests/providers/model-rename-migration.test.ts` (basename already
  registered in `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`,
  so no layout change is needed):
  - a two-boot fixture against a temp `OPENCODEX_HOME`: save a synthetic Antigravity config, run
    migrate-then-reconcile, assert reconciliation put the retired keys on disk, then assert the
    second boot renames nothing and prints nothing. Red on `dev` at that last assertion.
  - the intended asymmetry: a registry-published id stays in metadata while a stale
    `selectedModels` entry is still repaired and still reported once.
  - the general invariant, table-driven over every shipped rename: a provider row built by
    `providerConfigSeed` must never be something the migration wants to rewrite. Red on `dev` for
    all nine Antigravity renames.
- The first push of this branch had a broader guard; `test 4/4` caught that it also skipped a row
  saving only the retired key, which is the case that must still be migrated. That is the second
  condition above, and `gemini-37-flash-migration.test.ts:211` is the regression test for it.
- File-size ratchet: neither changed file is tracked in `tests/fixtures/file-size-baseline.json`,
  and the test file is 418 lines against the 2000-line threshold for untracked files.
- Branch is cut from `origin/dev` at 6044a9ab38.

## Checklist

- [x] Docs or release notes were updated when needed. (No user-facing behavior change beyond a log
      line that stops repeating; nothing in `docs-site/` describes it.)
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. (No auth,
      credential or workflow surface is touched; the change narrows which config fields a startup
      migration rewrites.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed repeated model metadata migrations after provider reconciliation.
  * Retired model identifiers are now repaired when user-selected, while supported registry-provided metadata remains intact.
  * Prevented migrations from repeatedly reporting changes across application restarts.


abhisheksharma2411 (commented):
The diagnosis is the valuable part here and it's exactly right — this was never a persistence failure, it was two boot phases disagreeing. `runModelRenameStartupMigration` deletes the keys, `reconcileOAuthProviders` → `applyOAuthPresetCatalog` copies the preset's record straight back, and the next boot finds them again. A migration that loses to the second half of its own boot is a genuinely hard thing to see from the symptom, and "nine messages from a user who never selected nine models" is the tell that makes it obvious in hindsight.

Traced the root cause to confirm the premise rather than take it on trust, and it holds:

```ts
export const ANTIGRAVITY_MODEL_CONTEXT_WINDOWS = {
  "gemini-3.8-flash": …, "gemini-3.7-flash": …, "gemini-3.1-pro": …,
  ...ANTIGRAVITY_WIRE_MODEL_CONTEXT_WINDOWS,
  ...Object.fromEntries(Object.entries(ANTIGRAVITY_MODEL_ALIASES).map(…)),   // <- every alias
};

export const ANTIGRAVITY_MODEL_ALIASES = {
  ...ANTIGRAVITY_VISIBLE_MODEL_ALIASES,
  ...ANTIGRAVITY_COMPATIBILITY_MODEL_ALIASES,   // <- includes the retired ids
};
```

So the seed publishes a window for every **compatibility** alias, retired ones included, which is precisely why the delete can never converge.

**Worth noting the sibling map does not do this.** `ANTIGRAVITY_MODEL_INPUT_MODALITIES`, twenty lines below, lists live ids explicitly and derives nothing. Two maps describing the same model set, one curated and one derived-over-all-aliases — that asymmetry *is* the root cause, and it'll produce the next instance of this bug in whatever field gets added next.

I'd still land this rather than chase that, and I think the PR chose correctly: narrowing `ANTIGRAVITY_MODEL_CONTEXT_WINDOWS` to live ids would strip the window from requests that still use a retired alias, which is a behaviour change on the request path to fix a logging and config-churn problem. Suppressing at the migration is the smaller blast radius. But it's worth a line in the code or the issue saying the derivation is the upstream cause, so the next person doesn't re-derive the whole diagnosis.

**On `isRegistryResidue` itself:** requiring *both* conditions — registry still publishes the retired id, and the saved field already carries the supported id — is the right pair. It also self-heals: the day the registry stops seeding a retired alias, `publishes` goes false, the migration resumes, and the key finally gets cleaned up without anyone revisiting this function. Worth saying that in the doc comment, because "skip the rename" reads like a permanent exemption and it isn't.

140 lines of test for 55 lines of fix is the right ratio for a boot-path migration.

Issue #2894, opened 2026-08-29
[Feature] SOCKS5 proxy support for outbound provider calls - and fail fast on unsupported proxy schemes


nordz0r:
﻿### Area

Proxy and routing

### What are you trying to accomplish?

Two related needs when reaching model upstreams from restricted networks:

1. Route **all** provider traffic through an outbound HTTP or SOCKS5 proxy, including SOCKS5 endpoints exposed by local tunnel clients (sing-box, v2ray/Xray, `ssh -D`, Tor), which is the most common listen mode for those tools.
2. Route **individual providers** through **different** proxies — and others directly. Example (real): Google Antigravity must exit through a US proxy, xAI through a residential/EU proxy, and zai/GLM should go direct from the local network, because a foreign exit adds latency and triggers regional risk controls on a Chinese upstream. One global `proxy` for everything cannot express this — exactly the split Omniroute has: a common default proxy plus a per-provider override with a "direct / default / custom" choice.

### What prevents this today?

1. `config.proxy` supports **HTTP(S) proxy URLs only**. Bun's fetch (the bundled runtime) does not implement SOCKS — `socks5://` is not negotiated (upstream oven-sh/bun#16812 open), and worse, Bun silently treats the unknown scheme as HTTP instead of failing (oven-sh/bun#11343). The config layer accepts any string without scheme validation, so `proxy: "socks5://127.0.0.1:1080"` is stored, mirrored into `HTTP_PROXY/HTTPS_PROXY` by `applyProxyEnv()`, and only produces confusing network-level errors at request time. A user with only a SOCKS5 endpoint must spin up an extra HTTP listener (privoxy/glider/sing-box mixed port) as a workaround.
2. `proxy` exists **only at the top level** of `OcxConfig` (`src/types/config.ts:506`). `OcxProviderConfig` (`src/types/provider.ts:138`) has no proxy field of any kind, so per-provider egress is impossible: either everything goes through the single proxy (regional upstreams suffer), or nothing does (restricted upstreams are unreachable).

### What should OpenCodex do?

**A. Per-provider proxy with a global default (Omniroute-style two-level model):**

- Global `proxy`/`noProxy` keep working exactly as today and become the **default** for every provider.
- Each provider gains `proxy` + `noProxy` with three effective states:
  - unset → inherit the global default;
  - explicit value → use this provider's own proxy (fully overriding the global for that provider);
  - `null` / `""` → force **direct** egress, exempting the provider from the global proxy.
[...]

PR #5087, opened 2026-09-18
fix(security): keep DNS pinning unless the proxy applies to the request

luvs01:
﻿## Motivation

The outbound transport decision still treats "some proxy variable exists" (`outboundProxyConfigured`) as proof that this request will ride a proxy. A scheme-mismatched variable (`HTTP_PROXY` for an `https:` target), a `NO_PROXY` match, or a non-SOCKS `ALL_PROXY` therefore still admits benchmark/fake-IP DNS answers, downgrades the validated-address transport to an unpinned fetch, and raises a spurious `NO_PROXY` error for private providers — reopening the DNS-rebinding window the pinned transport exists to close.

## Description

- Derive a single `proxyApplies` (scheme-matched `effectiveProxyFor` result that `NO_PROXY` does not exempt) and use it for benchmark-address admission, the DNS-failure proxy fallback, the post-resolution transport choice, and the private-network `NO_PROXY` error.
- Requests that will not actually ride a proxy now keep the pinned transport, matching the documented "only when this exact host will use the configured outbound proxy" contract.

## Tests

- `bun test tests/providers/provider-outbound.test.ts` — 24 pass; new regressions assert scheme-mismatched variables and `NO_PROXY` matches retain the pinned transport, do not admit benchmark addresses, and no longer demand `NO_PROXY` for private providers.
- `bun test tests/providers/provider-outbound-private-network.test.ts tests/server/proxy-env.test.ts` — 44 pass.

## Review readiness checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected outbound request routing when proxy settings do not apply to the destination.
  - Requests now bypass proxies when the proxy scheme does not match the URL or the destination is covered by `NO_PROXY`.
  - Private provider connections with inapplicable proxy settings now use the appropriate direct route.
  - Improved handling of proxy-related address restrictions and fallback behavior.
  - Plain HTTP requests on POSIX systems now honor compatible `ALL_PROXY` settings.
  - Invalid or unsupported proxy values are now ignored.


lidge-jun:
This PR had never actually run the expensive CI legs — like several others opened the same day, it was waiting on workflow approval, so the checks that looked clean were only the lightweight gates. I approved the run; here is the first real result.

`test 4/4` and `macos 1/2` fail on a pre-existing case, and it is a genuine contract conflict rather than a defect in the patch:

```text
DestinationDnsResolutionError: provider URL hostname all-proxy-only.invalid could not be resolved
  at resolvePublicAddresses (src/lib/destination-policy.ts:495)
  at async providerOutboundRequest (src/lib/provider-outbound.ts:159)
  at tests/providers/provider-outbound.test.ts
```

That hostname exists in the suite specifically to pin the opposite of what this change decides. The case asserts that with only `ALL_PROXY` set, a host that does not resolve locally is still reached, by falling through to `configuredOutboundFetch` when DNS fails — the `if (!proxyConfigured) throw error;` branch. Replacing `proxyConfigured` with `proxyApplies` removes that fall-through for an `http://` or `https://` URL whose only proxy variable is `ALL_PROXY`, because `effectiveProxyFor` is scheme-matched and returns null there. Your comment states the intent directly: "a non-SOCKS ALL_PROXY must not downgrade pinning".

So the question is not whether the patch works. It is which contract wins:

- An operator who sets only `ALL_PROXY` expects everything to route through it, including names only the proxy can resolve. That is what the existing case encodes, and it is a real deployment shape.
- DNS pinning should not be downgraded by a proxy variable that would not actually carry the request. That is what this patch argues, and it is also right as a general principle.

Both cannot hold for a non-SOCKS `ALL_PROXY` with a proxy-only hostname. Deciding that is the work here, and it is a security-boundary decision — DNS pinning falls under the explicit security-review requirement in `AGENTS.md` — so it should be argued rather than settled by whichever assertion is edited.

If you conclude the existing case is wrong, say why in the PR and change it deliberately, keeping the fall-through for the cases where it is still correct. If you conclude `ALL_PROXY` should keep admitting proxy-only DNS answers, then `proxyApplies` needs a third state rather than a boolean: a proxy that carries the request versus a proxy that can resolve for it are different questions, and the current expression answers only the first.

The four new cases you added are good and read correctly; the scheme-mismatch ones in particular pin something that was previously only implied. Leaving this open rather than merging. `dev` is unaffected.

luvs01:
The `test 4/4` / `macos 1/2` failure (`proxy mode reaches one real proxy`, `all-proxy-only.invalid`) was a real regression from this PR's gating change, now fixed at 479c3870:

- Root cause: `proxyApplies` is derived from `effectiveProxyFor`, which never consulted a non-SOCKS `ALL_PROXY`. The e2e fixture proves Bun's native fetch does honour `ALL_PROXY=http://…` for plain `http:` targets — the request reaches the proxy — so a request that will ride a proxy was classified as unpinned, and the `DestinationDnsResolutionError` was rethrown instead of falling back to the proxy transport.
- Fix: `effectiveProxyFor` now returns a non-SOCKS `ALL_PROXY` for `http:` targets on non-Windows platforms. On Windows the native fetch does not consult `ALL_PROXY` (verified locally: no proxy connection is attempted), so the pinned-transport classification stays correct there. For `https:` targets the socks5 wrapper remains the only `ALL_PROXY` route, unchanged.
- Covered by a new `effectiveProxyFor` case asserting the platform-dependent result; the e2e proxy-mode test passes locally (51 tests across provider-outbound and socks5-fetch), `tsc --noEmit` clean.

luvs01:
Author follow-up on the contract question — argued rather than settled by assertion edits.

The current head keeps the existing case exactly where it is true and narrows only the shape where the old contract was unfalsifiable:

- `proxyApplies` now counts `ALL_PROXY` only where the configured fetch will actually carry the request: SOCKS5 via the wrapper on every platform, and a usable `http(s)://` `ALL_PROXY` for `http:` targets on POSIX — which is the `all-proxy-only.invalid` shape, so that e2e case still passes there unmodified and still exercises "the proxy resolves the name".
- For `https:` + non-SOCKS `ALL_PROXY`, Bun never routes the request through that variable, so the previous fall-through was a retry through the same direct fetch — it could not succeed, and treating the variable as "a proxy can resolve this" also admitted proxy-only/fake-IP DNS answers for a transport that cannot use them. That shape now throws, deliberately.
- Malformed or non-http(s) values (`http://`, `not a url`, `ftp:`, `socks5:` in a scheme-matched variable) no longer count as an applicable proxy either.

So no third state: the boolean stays, but it now answers "will carry this request" precisely — which is what both the fake-IP admission gate and the DNS-failure deferral need. The existing case was not wrong; it was under-specified about which ALL_PROXY shapes can carry, and it is preserved exactly where they can.

One corollary to flag rather than silently edit: on Windows, Bun native fetch does not consult `ALL_PROXY` at all, so the `all-proxy-only.invalid` shape throws there by design (`effectiveProxyFor` returns null). If a Windows leg ever runs this e2e fixture, the 200 assertion needs a platform guard. Commits 308a31d..e04dc71 on the current head implement the above; the earlier CI failure ran on the pre-clause head.

Ingwannu (changes requested):
Requesting changes on exact head `375b785b7904294ae0a0998ff8aeaccc1ba0916d`.

The scheme/NO_PROXY classification is now much closer to the right contract, but the security comment's “snapshot once before the DNS await” guarantee is not actually preserved on the DNS-failure path. `effectiveProxy` is captured before `resolvePublicAddresses()`, yet the catch branch calls:

```ts
configuredOutboundFetch(url, { ...init, method, redirect: "manual" })
```

without the captured proxy. That function re-reads the live proxy environment. If the environment changes while DNS resolution is pending, admission can be decided for one proxy/NO_PROXY state and the request can leave through a different proxy or direct path. The same issue exists because `effectiveProxyFor(parsed)` and `noProxyMatches(parsed)` read `process.env` separately rather than one immutable snapshot.

Capture the relevant upper/lower-case proxy and NO_PROXY values once, derive both `effectiveProxy` and `proxyApplies` from that snapshot, and pass the captured proxy explicitly to every proxy transport branch, including DNS-failure fallback. The explicit path must work for both the SOCKS wrapper and Bun's HTTP(S) proxy option. Add a deferred-DNS regression that mutates proxy/NO_PROXY variables between admission and dispatch and proves the actual transport remains the one that was validated.

This is a DNS-pinning/credential-destination boundary, so keep it unmerged until that race and exact-head Cross-platform CI are green.

lidge-jun:
Superseded by #5264, merged to `dev` as `8e1fdea1c02ac1754f56b773d0c84c8470953301`.

Your request-scoped decision is the shape that landed: pinning and transport are now chosen by whether a proxy actually applies to the request rather than by whether one is configured, with scheme mismatch, `NO_PROXY` and DNS failure each held by a regression. The branch commit carries `Co-authored-by: luvs01`, so the contribution is attributed to you.

Closing as superseded rather than stale. The per-provider egress work that builds on this decision is tracked separately in #2894.

Issue #5122, opened 2026-09-19
[Feature]: Complete cross-layer send accounting for pre-output rate-limit replay


lidge-jun:
### Area

Proxy and routing

### What are you trying to accomplish?

Finish the explicitly deferred integration between adapter-local pre-output replay and OpenCodex request-wide send accounting, provider pacing and recovery telemetry.

### What prevents this today?

Public PR #5041 explicitly lists physical inner replay integration with the shared send budget, provider fetch wrapper and physical-attempt accounting as remaining work. At the pinned source, [src/server/responses/run-turn-execution.ts:162](https://github.com/lidge-jun/opencodex/blob/7864869c31c41cca9830d93540238f17df8faafb/src/server/responses/run-turn-execution.ts#L162) passes the budget/fetch seam, while [src/adapters/devin.ts:570](https://github.com/lidge-jun/opencodex/blob/7864869c31c41cca9830d93540238f17df8faafb/src/adapters/devin.ts#L570) does not thread it into the reset-retry wrapper.

### What should OpenCodex do?

Admit every actual inference send through the existing shared permit/observer path and count it once. Keep the zero-yielded-output replay boundary, cancellation and server-stated wait semantics unchanged. Reuse current spend reservation/refund and usage settlement semantics rather than adding another local counter.

### Example usage or interface

With a fake reset response followed by another attempt, the request-level used-send count and recorded physical attempts match actual fake-upstream sends. A refused next permit emits the existing withheld-recovery reason and makes no further inference request.

### Alternatives or workarounds

Removing safe replay discards the shipped recovery behavior. Incrementing budget.used directly misses reservation/observer/refund semantics. An adapter-only cap would preserve separate authorities.

### Additional context

Source snapshot: `7864869c31c41cca9830d93540238f17df8faafb` after fast-forwarding `dev` on 2026-09-19. This is source-grounded analysis, not a runtime test result.

This tracks a remaining item already disclosed in [#5041](https://github.com/lidge-jun/opencodex/pull/5041), not a new broad retry system. #5044/#5056 handle attribution when recovery is withheld and do not complete this adapter integration. Long-wait keep-alives and a cross-transition scheduled-wait allowance remain separate work.

### Implementation path

[...]

PR #5170, opened 2026-09-19
test(devin): assert which error the refused send raises and what the caller sees

lidge-jun:
## Summary

Two test-only follow-ups to #5152. Nothing in production changes, and the merged branch is untouched.

The adapter-direct case for a refused initial send caught the refusal with a bare `.catch(() => {})` and discarded it. That passed for any error at all: if the adapter had thrown an ordinary upstream failure instead of the budget refusal, the case would still have been green while the turn reached the operator as a provider problem rather than a budget decision. The case now captures what escaped and asserts `SendBudgetExhaustedError`, alongside its existing assertions that no inference request was made and no permit was spent.

The outer case reported the HTTP status only inside its failure message. It is now asserted. The refusal reaches the client as an error code on a buffered **failed** response, so the outer status is **200**, not 429. That is worth pinning precisely because 429 is the obvious guess and this path does not use it; a future change that turned the refusal into a transport-level status would now be caught here rather than discovered downstream.

Relates to #5122.

## Verification

- **Local checks: NOT RUN.** This lane is prohibited from running local suites, focused tests, typecheck, builds, installs or any live `ocx` command. Hosted CI on this PR is the executable evidence.
- Static screen: both changed files parse cleanly under an AST error-node check, and neither is near a file-size cap.
- No assertion was removed or relaxed; both changes are additive checks on existing cases.

## Checklist

## Summary by CodeRabbit

- **Tests**
  - Strengthened coverage for send-budget exhaustion scenarios.
  - Verified that budget refusals return the expected error type.
  - Confirmed zero-budget requests return a successful HTTP response containing the appropriate buffered error.
  - Ensured no upstream requests, observations, or budget usage occur after refusal.


lidge-jun:
Maintainer integration into `dev` under MAINTAINERS.md, by the authenticated project owner. This is an integration decision, not a self-approval.

Reviewed head: `825d4ed9c72233bc862668fec6010862bcddadea`. The two-file diff pins the actual typed send-budget refusal and the existing HTTP 200 buffered-failure response, with zero physical sends retained. Independent static review passed; no open review threads or maintainer objections were found. The merge with current dev `19b42c5b08f56a309a1b54f2ceb2c188bcfd71cf` is clean.

Exact-head applicable hosted CI and aggregate passed: https://github.com/lidge-jun/opencodex/actions/runs/35443894116 . Target enforcement passed at the same head in run 35443894094; a separate canceled duplicate is not used as proof. Event-inapplicable jobs remain skipped. Local tests, typecheck, build and runtime execution were not run, per the task constraint.

Issue #5115, opened 2026-09-19
[Feature]: Derive model selector decode hints from all classified registry identity maps


lidge-jun:
### Area

Catalog / models

### What are you trying to accomplish?

Allow an encoded provider/model selector to round-trip when its native model ID is declared only in an authoritative policy map, without falsely publishing that model as available.

### What prevents this today?

[src/router.ts:107](https://github.com/lidge-jun/opencodex/blob/7864869c31c41cca9830d93540238f17df8faafb/src/router.ts#L107) gathers known native IDs from a hand-written subset of model-keyed registry maps. [src/providers/registry/types.ts:169](https://github.com/lidge-jun/opencodex/blob/7864869c31c41cca9830d93540238f17df8faafb/src/providers/registry/types.ts#L169) declares additional identity-bearing maps. The type system permits an ID to exist only in an omitted map. No specific current production failure is claimed.

### What should OpenCodex do?

Classify every registry field whose keys are native model IDs and derive selector decode hints from that classification. Keep decode hints separate from catalog availability and retain ambiguous-selector rejection.

### Example usage or interface

A synthetic registry entry contains `vendor/model-a` only in its wire-default map. Its client-safe encoded selector decodes to that native ID. The hint alone creates no new visible catalog row.

### Alternatives or workarounds

Requiring every policy key in the static models list would interfere with live discovery. Retaining a partial hand-maintained list leaves future fields easy to omit.

### Additional context

Source snapshot: `7864869c31c41cca9830d93540238f17df8faafb` after fast-forwarding `dev` on 2026-09-19. This is source-grounded analysis, not a runtime test result.

Adjacent closed #4730 concerns catalog slug collisions, not completeness of registry decode hints. Treat this as a bounded robustness proposal, not a live incident.

### Implementation path

1. Add `src/providers/registry/model-ids.ts` with a typed classification/helper.
2. Replace the manual map list in `knownModelIdsForProvider` while retaining config IDs, transport identity checks and stale discovered hints.
3. Classify maps carefully: not every object key is a model ID. Make new model-keyed fields require an explicit classification decision.
4. Extend `tests/codex-integration/slug-codec.test.ts` and registry parity coverage using a model present only in each formerly omitted map. Update `structure/catalog.md` and `structure/providers-and-adapters.md`.

### Acceptance criteria and verification

- Each classified map independently supplies decode hints.
- Unknown/ambiguous encoded IDs retain current fail-closed or pass-through behavior as applicable.
[...]

PR #5177, opened 2026-09-19
feat(registry): derive selector decode hints from a classified registry

lidge-jun:
## Summary

An encoded provider/model selector is decoded against the native model ids a provider is known to accept. That set was a hand-written list of eight registry maps in `knownModelIdsForProvider`, so an id declared only in an unlisted identity-bearing map could not be decoded, and every new model-keyed registry field had to be remembered by whoever added it.

`src/providers/registry/model-ids.ts` now classifies **every** field of `ProviderRegistryEntry` by what its keys mean, and the router asks the registry instead of listing maps itself. Fifteen direct maps contribute keys, `keyAuthServiceTier` contributes its nested `modelSupportsServiceTier` keys, and the remaining seventy-five fields are classified as carrying no identity.

Classifying the non-model fields is the point. The `satisfies` clause makes a new registry field a **type error** until somebody decides what its keys mean, rather than letting it default to invisible. That is the difference between this and a longer hand-written list.

### Two exhaustiveness checks, deliberately different

- The `satisfies Record<keyof ProviderRegistryEntry, RegistryFieldModelIdRole>` clause proves the classification covers the interface, at compile time.
- A parity case proves the fields the shipped entries actually carry are all classified, at runtime.

A new interface field fails the first. A shipped entry carrying something unclassified fails the second. Neither substitutes for the other.

### What a hint is not

Decode hints are identity evidence only. They publish no catalog row, grant no availability or entitlement, and confer no transport authority: registry transport identity is still checked before hints are read, and an ambiguous selector is still rejected rather than guessed. `virtualModels` contributes its selectable keys only, never the `wireModelId` transport target, which is not something a client may name.

Closes #5115.

## Verification

- **Local checks: NOT RUN.** This lane is prohibited from running local suites, focused tests, typecheck, builds, installs or any live `ocx` command. Exact-head hosted CI on this PR is the executable evidence.
- Static screen: every changed file parses cleanly under an AST error-node check.
- Route-level coverage in `tests/codex-integration/slug-codec.test.ts`: all fifteen classified direct maps and the nested key-auth map each independently seed a decode, using distinct synthetic slash-bearing ids so the encoded selector round-trip is meaningful. Negatives cover unknown pass-through, ambiguous-selector rejection, transport mismatch inheriting nothing, and no catalog row from a hint alone. The helper contract is asserted too: the returned array is frozen and rejects a push, and the registry entry is byte-identical afterwards, which matters because the registry is process-wide shared state.
- Registry parity in `tests/providers/provider-registry-parity.test.ts`: runtime classification coverage over the shipped registry, per-map independent collection, non-model records (`staticHeaders`, `reasoningEffortMap`, the repair object) contributing nothing, stable first-seen-wins ordering in registry-field order, and a shipped entry whose ids are recovered from its classified maps even though the helper never reads its `models` list.
- No file-size cap moves, and no layout inventory changes because no test file was added.

## Checklist

## Summary by CodeRabbit

* **Enhancements**
  * Selector decoding now recognizes model IDs defined across the provider registry, including nested service-tier mappings and virtual model identities.
  * Model IDs are deduplicated consistently while preserving registry order.
  * Unknown and ambiguous selectors continue to be handled safely, without treating decode hints as availability or entitlement.

* **Documentation**
  * Clarified how registry model IDs support selector decoding and documented the associated behavior.


lidge-jun:
Maintainer integration into dev under MAINTAINERS.md by the authenticated project owner, not a self-approval.

Reviewed head `a654ee8e4469f2c431b378a4ec42c68765e74a90` passed all applicable hosted checks and aggregate in https://github.com/lidge-jun/opencodex/actions/runs/35449601489 . macOS logs explicitly execute all classified direct-map public decodes, nested key-auth map decoding, immutability, unknown pass-through and ambiguity rejection. Independent source, test and typed-iteration reviews passed; no unresolved threads or maintainer objections remain.

Current dev `40670044146adc714f06d8b2e29f9114f8393df5` has no changed-file intersection with this six-file change since its base. Integration tree `eb3742407d206a7a4d427c601004c5b495448c02` preserves all six files byte-for-byte and passed independent union review. The earlier policy-first merge order was a coordination preference: both source audit and the policy implementation owner confirm no semantic dependency. This PR can land standalone; the later policy consumer rebase must preserve both router imports and this helper and obtain fresh CI.

Native-stack membership is empty. Actor, dev base and exact head were revalidated. No local suites, typecheck, build or runtime execution were performed. Issue #5115 closes only after actual dev landing and ancestry verification.

Issue #5159, opened 2026-09-19
fix(volcengine): align Coding Plan preset with native Responses API


juzijia:
### Client or integration

Direct HTTP client (validated through OpenCodex's Responses endpoint; the original incompatibility was observed from a Pi/Codex-compatible client path).

### Provider or upstream service

Volcengine Ark Coding Plan (`volcengine-coding-plan`)

### OpenCodex version

2.59.0

### Endpoint or capability

Volcengine Coding Plan native Responses endpoint and tool-call continuation:

- Upstream base: `https://ark.cn-beijing.volces.com/api/coding/v3`
- Expected Responses resource: `/responses`
- Full upstream URL: `https://ark.cn-beijing.volces.com/api/coding/v3/responses`
- Capabilities involved: Responses wire selection, streaming tool calls, function-call continuation, reasoning-item replay

### Current behaviour

Volcengine's current Coding Plan Codex documentation specifies the Responses wire protocol, but OpenCodex 2.59.0 still presets `volcengine-coding-plan` to `openai-chat`.

Current built-in preset:

```text
provider: volcengine-coding-plan
baseUrl: https://ark.cn-beijing.volces.com/api/coding/v3
adapter: openai-chat
```

On the existing Chat translation path, GLM-5.3 has produced real streamed tool-call failures such as:

```text
upstream_server_error: upstream stream produced malformed tool call arguments
```

To isolate the wire protocol, I added a separate non-default test provider without modifying the production provider:

```text
adapter: openai-responses
baseUrl: https://ark.cn-beijing.volces.com/api/coding/v3
responsesPath: /responses
```

Native Responses then succeeded for all three Coding Plan models used in this environment:

- `deepseek-v4-flash`: HTTP 200, exact smoke response, `sendCount=1`, no recovery.
- `glm-5.3`: HTTP 200 function call with valid JSON arguments; continuation HTTP 200; no malformed tool-call error.
- `glm-5.3-flash`: HTTP 200 function call and continuation; valid JSON arguments; no malformed tool-call error.

A second compatibility issue was also reproduced on the native Responses path: replaying the full prior `reasoning` output item together with `function_call_output` caused Ark to return `400 InvalidParameter`. Replaying only the required `function_call` plus `function_call_output` succeeded with HTTP 200.

### Expected behaviour

The built-in `volcengine-coding-plan` preset should follow Volcengine's current Codex documentation and use native Responses by default.

Expected provider behaviour:

```text
adapter: openai-responses
[...]

Issue #5104, opened 2026-09-19
[Provider compatibility] Proxy-originated model discovery sends no User-Agent; WAF-fronted gateways reject it with HTTP 403


mdwsk88:
### Client or integration

Codex App

### Provider or upstream service

AI2API - an internal corporate gateway speaking the OpenAI Responses API (provider config: `adapter: "openai-responses"`, `authMode: "key"`). The endpoint is internal-only; there is no public hostname or public specification for the gateway. Its WAF rejects any request that carries no `User-Agent` header with HTTP 403 and an empty body.

### OpenCodex version

2.59.0 (npm `@bitkyc08/opencodex`). Source-checked against the installed package: `src/oauth/index.ts` `buildModelsRequest()` sets only `Authorization: Bearer ...` for this adapter shape (plus registry static headers where applicable), and `src/lib/provider-outbound.ts` executes it through `pinnedHttpGet()` in `src/lib/pinned-http.ts`, a Node-style `node:https` request that adds no default `User-Agent`.

### Endpoint or capability

Model discovery (`GET {baseUrl}/models`).

### Current behaviour

After adding the provider, no models ever appear: the Models page shows no rows for it and the Codex catalog gains no entries. On 2.59.0 a newly registered key-auth provider starts with `initialModelSelection.status: "pending"`, and a pending provider has all of its model rows hidden, so the visible symptom is "provider added but invisible" with no pointer to the discovery failure in the UI.

`service.log` repeats on every discovery pass:

```
[opencodex] Provider model discovery for "AI2API" failed with HTTP 403 [urlClass=provider-models, fallback=configured].
```

Root cause as observed: the gateway's WAF requires a `User-Agent` on every request. The discovery fetch is proxy-originated, so there is no client request to inherit a UA from, and the pinned Node-style transport sends none. Controlled comparison from the same machine, same credential, and same resolved address:

- `curl` (sends `User-Agent: curl/x`) - HTTP 200 with the full model list
- Bun `fetch` (sends `User-Agent: Bun/x`) - HTTP 200
- Node-style `https.request` with only `Authorization: Bearer ...` (what discovery sends today) - HTTP 403, empty body
- The same Node-style request plus any `User-Agent` value - HTTP 200

[...]

PR #5188, opened 2026-09-19
feat(registry): pin live-verified Alibaba Token Plan models to the Responses wire

mdwsk88:
## Summary

- Alibaba Token Plan (Beijing) served every inbound through the openai-chat translation layer even though the gateway now documents a native OpenAI-compatible Responses API on the same /compatible-mode/v1 base and ships an official Codex integration guide on wire_api = "responses" (#5097). The chat bridge works (88-96% prefix-cache hit on real traffic) but cannot trigger the Token Plan Harness tools, which Alibaba documents as Responses-only, and adds a translation hop with known edge cases for custom tools and reasoning envelopes.
- This PR pins the three live-verified models (qwen3.8-flash, qwen3.7-plus, glm-5.3) to openai-responses via the registry modelWireDefaults mechanism, scoped to Responses inbound only - the same shape as the DeepSeek deepseek-v4-flash precedent and the xAI entries: Codex clients ride the native wire with zero translation hops, while chat and anthropic inbound keep the provider-wide chat wire and its measured prefix-cache behavior. modelAdapters keeps winning in both directions (per-model opt-in for the rest of the family, and the opt-out from these defaults).
- Live evidence (from #5097, end-to-end on 2.59.0 against the Beijing gateway): qwen3.8-flash and qwen3.7-plus through the proxy with per-model modelAdapters - HTTP 200, a proper custom_tool_call with a correct patch, response.custom_tool_call_input.delta/.done SSE shapes, response.completed with usage, and the usage log recording adapter openai-responses; glm-5.3 direct upstream through a full agent loop (tool-call turn plus a continuation replaying custom_tool_call and custom_tool_call_output). The response marker x_billing_type from Alibaba ("response_api") confirms the Responses path was served and billed.
- Update (rebased onto dev after #5198, 260922): the delta #5198 held the pin back for is now closed with its own live evidence on the Beijing gateway, and the pin lands WITH the entry-level `preserveResponsesReasoningContent: true` — the pairing the #5198 comment and the opt-in test guard require:
  - Plaintext reasoning replay: a two-turn conversation whose second request replays the reasoning item WITH its plaintext content array (plus the assistant message and a follow-up) is accepted (HTTP 200) and the model continues correctly from it (391 -> 782), both for the gateway's own item shape (content: null, summary carried) and for a synthesized non-empty content array - the exact shape `preserveResponsesReasoningContent` keeps. Without the flag the serializer blanks that array, so the flag is what carries Chat-side parity onto the pinned wire.
[...]

Issue #5097, opened 2026-09-19
[Provider] alibaba-token-plan: native OpenAI Responses wire is officially supported upstream — request validated opt-in / default flip


mdwsk88:
### Client or integration

Codex App (verified end-to-end through the OpenCodex proxy) / Codex CLI

### Provider or upstream service

Alibaba Cloud Model Studio — Token Plan (Beijing) `alibaba-token-plan`, Token Plan (Intl) `alibaba-token-plan-intl`

### OpenCodex version

2.59.0 (npm `@bitkyc08/opencodex`); also confirmed on upstream/dev @ 7817cd57f (2026-09-19)

### Endpoint or capability

`POST {base}/compatible-mode/v1/responses` on the Token Plan gateway (`https://token-plan.cn-beijing.maas.aliyuncs.com/compatible-mode/v1`) — native OpenAI Responses wire, custom tools, streaming.

### Current behaviour

Both Token Plan registry entries pin `adapter: "openai-chat"` (`src/providers/registry/entries-extended.ts:792-795` and the intl sibling at :835), so the proxy translates every inbound Responses request to Chat Completions toward Alibaba. It works — I observed 88-96% prefix-cache hit on real traffic — but:

- Token Plan's built-in "Harness" tools (web search, code interpreter, web extractor, file_search, MCP-in-Responses) are documented by Alibaba as **Responses-only** and cannot be triggered over the chat wire.
- The chat bridge adds a translation layer for Responses-specific shapes (custom tools, reasoning envelopes) with known edge cases.

### Expected behaviour

Alibaba now officially documents an OpenAI-compatible **Responses API** on the same `/compatible-mode/v1` base, and publishes an official **Codex integration guide** with `wire_api = "responses"` against the same Token Plan base URL. OpenCodex should offer a first-class Responses wire for the Token Plan entries — opt-in first (registry `modelWireDefaults` or a documented `modelAdapters` preset), then a default flip after field validation.

### Minimal redacted request or reproduction

Provider config (redacted) that opts one model in today:

```json
"alibaba-token-plan": {
  "adapter": "openai-chat",
  "baseUrl": "https://token-plan.cn-beijing.maas.aliyuncs.com/compatible-mode/v1",
  "apiKey": "sk-sp-***",
  "modelAdapters": { "qwen3.8-flash": "openai-responses" }
}
```

Through the proxy:

```bash
[...]

PR #5198, opened 2026-09-19
Provider discovery and model surface: outbound User-Agent, Volcengine Responses, Token Plan opt-in, Command Code ladders

lidge-jun:
## Summary

Provider discovery and model surface, one branch targeting `dev`. Each item had an open contributor PR; three are carried with attribution and one is implemented here. Two of the four proposals changed a default for existing users, and both were narrowed.

| Commit | Issue | Source |
|---|---|---|
| `549d968d41` fill a default User-Agent on proxy-originated provider outbound | #5104 | #5186 (@mdwsk88) |
| `ac967a0782` point the Volcengine Coding Plan preset at the native Responses API | #5159 | #5173 (@juzijia) |
| `fdd9b8e9ec` document and lock the Alibaba Token Plan Responses opt-in | #5097 | #5188 (@mdwsk88) |
| `f005bd56db` let an operator ladder outrank the shipped effort table at the wire | #5096 | — |
| `3be1eccfeb` finish wiring the Volcengine replay-drop flag through registry and routing | #5159 | review response |
| `9e4e605c8c` make the operator ladder override an explicit declaration | #5096 | review response |
| `913d1b68b0` record the Command Code ladder authority in behavior identity | #5096 | review response |
| `fe142d5d17` recount the provider preset totals from the registry and pin them | — | documentation drift |

Closes #5104
Closes #5159

#5097 and #5096 are each addressed in part and deliberately not closed; see below.

### User-Agent on proxy-originated outbound (#5104)

Model discovery, connection tests, the Ollama show probe and the Antigravity quota probe are assembled by the proxy, so there is no client request to inherit a User-Agent from and the pinned Node-style transport sends none. WAF-fronted gateways answer with 403, which surfaces as "provider added but no models". The wrapper now fills `User-Agent: opencodex` when the caller named none, case-insensitively across all three `HeadersInit` shapes.

Two changes on top of #5186. The regression suite stubbed `pinnedGet`/`pinnedPost`, so it could not see the caller-owned-executor branch, which leaves the wrapper without touching the pinned transport — a fill applied only on the pinned path would have left that branch UA-less and still 403 behind the same WAF; there is now a case for it. And the claim that a caller keeps its header-name *spelling* is dropped from the comment, the structure doc and a test name: the pinned and SOCKS transports both rebuild the set through `new Headers()`, which lowercases every name, so the value is what survives.

### Volcengine Coding Plan on native Responses (#5159)

The preset moves to `openai-responses` with `responsesPath: /responses`, declares `supportsServiceTier: false`, and keeps the retired Chat destination as an alias. Validated Ark continuations reject the reasoning item the previous turn returned, so the entry sets a new provider-scoped `dropResponsesReasoningItems` flag; it is lossy, so it is documented as such and can be set to `false`.

**The startup config migration from #5173 is not carried.** It would have rewritten every stored canonical Chat row to Responses on the next boot. It borrowed the shape of the Z.AI wire migration while inverting the property that makes that one safe: `zai-responses-migration.ts` gates on `providerMatchesRegistryTransport` and therefore only rewrites rows the router already canonicalizes at request time, which is why its own comment calls it behavior-preserving by construction. A Volcengine Chat row is not canonicalized — `volcengine-coding-plan` is a `preserveCustomDestination` key entry, so the adapter mismatch makes `routedProviderConfig` return the stored row untouched — and a version marker introduced now cannot tell the old default apart from a deliberate pre-upgrade Chat choice. The preset default therefore applies to new rows only, existing rows keep their wire, a regression case pins that, and the docs say how to switch by hand.

Review then found the flag reached the adapter but not three places that must know about it: the exhaustive registry field classification (a compile error), the compatibility behavior record, and the resolved static policy. It also missed the one routing path this narrowing makes reachable — a row saved on Chat that opts a single model into Responses through `modelAdapters` leaves `routedProviderConfig` by the early return, so the flag has to be filled there from the destination matcher as well. All four are fixed, with the wire body asserted rather than the flag alone.

### Alibaba Token Plan Responses opt-in (#5097)

The issue asked for a validated opt-in or a default flip. This lands the opt-in and declines the flip. #5188 proposed pinning `qwen3.8-flash`, `qwen3.7-plus` and `glm-5.3` through `modelWireDefaults`, which would move every existing Codex user of those models onto a different upstream with no config change, and one delta is unresolved: the entry's `preserveReasoningContentModels` is read by the **Chat** adapter, while the Responses serializer reads `preserveResponsesReasoningContent`, which this entry does not set. Pinned models would replay with blanked reasoning content — strictly less state than they carry today. Z.AI and DeepSeek set both flags together and their entry comments say why. Blanking is the fail-safe direction, so leaving the default alone costs nobody a working setup, while setting the Responses flag on faith could 400 a continuation on a gateway nobody has verified it against.

The new suite holds both halves: the three models resolve to `openai-responses` once opted in and the request actually reaches `/responses` rather than being flipped back by the `handleResponses` replay, the wire default stays Chat on every inbound, and a guard fails if a Responses wire default is ever declared for this entry without `preserveResponsesReasoningContent` beside it. That guard is the precondition for the flip the issue asks for, so #5097 stays open.

### Command Code effort ladders (#5096)

The adapter resolved its wire effort as `commandCodeReasoningEfforts() ?? configuredReasoningEfforts()`, so a model **with** a row in the shipped table ignored `providers.command-code.modelReasoningEfforts` outright while a model **without** one honoured it. The catalog never agreed with that split — it advertises the picker from `configuredReasoningEfforts` — so an operator who widened a pinned row saw the wider ladder offered in Codex and then watched the adapter strip the rung on the way out, with no error to explain it.

Configuration can now win, but only when the provider declares `modelReasoningEffortsAuthoritative`. The first attempt inferred provenance by comparing the configured row against the shipped table, and review rejected it: `providerConfigSeed` copies the whole table into every materialized preset, and enrichment and routing both keep a persisted row over the current seed, so a row written by an older release keeps its old value and would start looking like an operator edit the moment the shipped table is corrected — at which point the stale row would outrank the correction *and* disable the rejection repair. Since the follow-up this issue asks for is exactly a table correction, that misfire was not hypothetical. A declared flag cannot be produced by seeding, so its presence is evidence and a value comparison is not.

With the flag set the ladder resolves through the same function the catalog uses, so the picker and the wire agree, and a rung the upstream refuses is returned as that error instead of being replayed without the effort. Without it, a configured row — seeded, stale or hand-written — changes nothing. Because the flag changes both the bytes sent and the recovery taken, it is part of the compatibility behavior record and the resolved static policy; two routes that disagree about it are not the same subject.

One alias asymmetry the override made reachable is fixed too: the `xhigh` branch aliases to `max` only when the ladder does not advertise `xhigh`, but the `ultra` branch aliased whenever `max` existed, so an authoritative ladder offering `ultra` would have advertised it in the picker and quietly sent `max`. No shipped row offers `ultra`, so the built-in table is unaffected.

**The issue's stated root cause is disproven.** #5096 attributes the clobber to `applyDirectReasoningEffortContracts()`, which returns immediately unless `entry.directReasoningEffortModels` is populated; neither the `command-code` nor the `commandcode` entry declares it, so that function is a no-op for this provider and editing it would have fixed nothing. The real shadowing is the `??` at the adapter. Configuration is not erased or reverted on restart — it is ignored at request construction for known rows.

**The data half is not adopted.** The issue also reports seven shipped rows narrower than the live API accepts and 38 live models with no row. The table's own provenance rules require per-row evidence; the measurements are a third party's and cannot be reproduced without a GOAT-plan key; the ids double as the router's known-ids decode source through `knownModelIdsForProvider`, so a mis-cased id has routing consequences; and at least one proposed widening contradicts an alias the file documents from the model profile (`xhigh -> max` on `deepseek/deepseek-v4-flash`). With this change an operator can apply the measured ladders from config today, and the reporter offered to open the full 46-model table as its own PR, which is where that provenance belongs. #5096 therefore stays open for that half.

### Preset-count drift

Seventeen pages restate how many built-in presets ship, and sixteen had drifted: the English provider guide said 95 total with 79 key-based, while the seven translated guides, all eight quickstarts including the English one, and `structure/ops/docs-and-release.md` still said 94 and 78. Nothing caught it, because both numbers read as plausible and no check compared either to the registry.

The registry says 95. Every `authKind` declaration in the two entry files that compose `PROVIDER_REGISTRY` is a string literal, and they group as 79 key, 12 oauth, 3 local, 1 forward — which is what the English guide already claimed. The other sixteen places now agree, and a new `tests/ci-workflows/docs-provider-preset-counts.test.ts` derives both numbers from `PROVIDER_REGISTRY` and asserts them per page, so the next preset fails every locale at once instead of drifting. Each page is anchored by a locale-specific phrase rather than by its number, so rewording a sentence fails the check and asks to be re-anchored.

## Verification

**No local verification was run, deliberately.** No test suite, no individual test file, no `bun run typecheck`, no build, no install, and no `ocx` invocation. A previous local run on this machine destroyed real `~/.opencodex` state, so this work is restricted to static reasoning and source oracles, with hosted CI on the exact head as the only execution evidence.

- Every carried change was read against its consumer in source rather than trusted from the source PR: `providerModelWireDefault` and `resolveWireProtocolOverride` for the wire defaults, `providerMatchesRegistryTransport` and `routedProviderConfig` for the Volcengine narrowing, `sanitizeReasoningInputContent` and `preserveResponsesReasoningContent` for the reasoning delta, `configuredReasoningEfforts` and `providerConfigSeed` for the Command Code precedence.
- Three rounds of adversarial static review by parallel reviewers — on each source PR before carrying, and twice on this branch's own diff. They produced both narrowings, the added executor-branch test, the registry classification that would not have compiled, the routing backfill, the replacement of the provenance heuristic, and the behavior-identity fix.
- Exhaustive unions this branch extends are registered: `dropResponsesReasoningItems` in the `satisfies Record<keyof ProviderRegistryEntry, ...>` classification, both new config fields in the `satisfies Record<keyof OcxProviderConfig, ...>` field policy and the persisted schema, `reasoning.effortsAuthoritative` in the closed behavior key set, and all three added test files in `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`.
- File-size ratchet: no file this branch touches carries a baseline cap.
- Rebased onto `c81d43053b`, which repaired the three failures this branch previously inherited from a red `dev` — the file-size ratchet on two files and three `Reserve actual catalog finalization lifecycle` cases, none of them touched here.
- Execution evidence is the exact-head CI run on this PR.

## Checklist


lidge-jun:
## 리뷰 · 우선순위 68 / 80

이 풀리퀘스트는 제공자 쪽 문제 네 개를 `dev` 한 브랜치에 넣는다. 아직 초안이다. 원격 검사는 대부분 통과했다. macOS 키링과 macOS 테스트 2/2는 이 글을 쓸 때 아직 돌고 있었다.

첫째, 프록시가 스스로 보내는 요청에 보낸 사람 칸을 채운다. 그 칸 이름이 User-Agent다. 모델 목록, 연결 시험, Ollama 확인은 사용자 요청을 그대로 넘기는 길이 아니다. 브라우저가 적어 둔 칸을 물려받을 곳이 없고, 바로 붙는 전송도 기본값을 안 넣는다. 앞단 방화벽은 칸이 빈 요청을 403으로 거절한다. 화면에는 제공자를 넣었는데 모델이 없는 것처럼 보인다. 이슈는 #5104다. 호출자가 이미 칸을 적었으면 그 값을 둔다. 없으면 `opencodex`를 넣는다. 모델을 부르는 추론 요청은 이 길을 안 탄다.

둘째, Volcengine 코딩 플랜의 새 기본선을 채팅 완성에서 Responses로 바꾼다. 주소는 같다. 직전 답의 생각 조각을 다음 요청에 넣으면 Ark가 400을 준다. 그래서 그 조각만 빼고 보낸다. 이미 저장해 둔 채팅 줄은 옮기지 않는다. 채팅 줄에서 모델 하나만 Responses로 고르면, 빼기 칸은 그 요청에도 따라간다. 이슈는 #5159다.

셋째, 알리바바 토큰 플랜은 모델 세 개를 Responses로 고를 수 있다는 것만 문서와 테스트로 잠근다. 세 개는 `qwen3.8-flash`, `qwen3.7-plus`, `glm-5.3`이다. 기본은 채팅 그대로다. 세 모델을 기본으로 바꾸면, 설정을 안 건드린 사람이 다른 길로 가고, 생각 글이 비어서 나간다. 그 칸을 확인하기 전에는 기본을 안 뒤집는다. 이슈 #5097은 열린 채로 둔다.

넷째, Command Code는 설정에 적은 노력 단계가, 표에 이미 있는 모델에서는 무시됐다. 목록에는 넓은 단계가 보이는데, 보낼 때는 좁은 표가 이겼다. 이제는 `modelReasoningEffortsAuthoritative`를 켜야 설정이 이긴다. 표와 값이 다른지만 보고 사람 수정으로 치면, 옛 버전이 저장해 둔 줄이 다음 수정표를 이긴다. 그래서 스위치를 따로 둔다. 표 자체를 46개 모델로 넓히는 절반은 안 넣었다. 이슈 #5096은 열린 채로 둔다.

문서에 적힌 프리셋 개수도 레지스트리에서 다시 세어 95와 79로 맞추고, 다음부터 어긋나면 테스트가 실패하게 했다.

라인 - `src/adapters/command-code.ts`의 `operatorChoseCommandCodeLadder`. 스위치는 제공자 전체다. `src/router.ts`의 `mergeStringArrayRecord`는 요청 때 레지스트리 표를 먼저 깔고, 사용자가 적은 줄만 그 위에 덮는다. 한 모델만 넓히려고 스위치를 켜도, 손대지 않은 모델은 표에 줄이 있으므로 같이 켜진다. 그 모델들은 업스트림이 단계를 거절해도, 단계를 빼고 한 번 더 보내지 않고 400을 그대로 돌려준다. 프리셋을 만들 때 `providerConfigSeed`가 표를 통째로 복사하는 것과 같다.

라인 - `docs-site/src/content/docs/ko/reference/configuration/providers.md`. 영어 설정 표에는 `dropResponsesReasoningItems`와 `modelReasoningEffortsAuthoritative`가 있다. 한국어 표에는 그 두 줄이 없다. 같은 번역 표도 같다. 한국어로 읽는 사람은 이 페이지에서 스위치를 못 찾는다.

메인테이너의 판단이 필요한 지점

이미 설치된 코딩 플랜을 채팅에 남겨 둘지. #5173은 다음 부팅에 Responses로 옮기려 했다. 이 브랜치는 그 이전을 빼 버렸다. 채팅으로 저장해 둔 줄은 요청 때 레지스트리 기본으로 안 바뀐다. 부팅 때 바꾸면 지금 쓰는 길이 바뀐다. 옛 기본과, 일부러 고른 채팅을, 지금 넣어 둔 숫자로는 구분할 수도 없다. 옮기지 않는 쪽이 맞다. 이미 400을 보는 사람은 설정을 직접 Responses로 바꿔야 한다.

알리바바 세 모델을 기본으로 고정하지 않은 것도 같다. #5188의 고정은 여기서 거절됐다. #5097은 닫지 않는 게 맞다.

Command Code 스위치가 표에 있는 모델 전부에서 400 재시도를 끄는 것을 그대로 둘지. 문서에 그 한 줄을 적을지, 나중에 모델 한 줄만 지정하게 바꿀지.

너의 추천

방향은 유지하면 된다. User-Agent, 코딩 플랜의 새 기본, 이전을 안 하는 이유, 알리바바를 기본으로 안 뒤집는 이유는 코드와 테스트가 맞다. 머지 전에 스위치 설명에 한 줄만 보태라. 이름이 `command-code`인 제공자는 표를 통째로 합치므로, 스위치를 켜면 손대지 않은 모델도 400을 그대로 받는다고 적으면 된다. 모델마다 스위치를 새로 만들 필요는 없다. #5186, #5173, #5188은 이 풀리퀘스트가 대신하므로 닫으면 된다. #5173의 부팅 이전은 가져오지 않는 게 맞다. `types.ts`와 `config.ts` 분할로 이 글을 무효로 볼 이유는 없다. 새 칸은 이미 나뉜 `src/types/provider.ts`와 `src/config/schema/leaf-validators.ts`에 들어가 있다. 초안을 풀고, 남은 macOS 검사가 끝나면 머지하면 된다.

이 댓글은 grok-bot이 작성했습니다

Issue #5181, opened 2026-09-19
readiness: Codex re-auth failure takes healthy multi-provider proxy out of service


nordz0r:
### Client or integration

Direct HTTP/API client

### Area

Service lifecycle

### Summary

When one or more Codex OAuth pool accounts require re-authentication, OpenCodex reports `/readyz` as HTTP 503 even though the proxy process is running and other configured providers/routes remain usable. In a single-replica Kubernetes deployment, the readiness failure removes the only Service endpoint and makes the entire proxy inaccessible.

Expected: Codex account-pool degradation should be reported as a component/provider failure. The proxy should remain ready when its HTTP server is healthy and at least one route/provider can serve traffic; native Codex traffic can fail or be marked degraded independently.

### Reproduction

1. Run OpenCodex in a container/Kubernetes pod with multiple providers and a Codex account pool.
2. Leave one or more Codex accounts expired so they require re-authentication.
3. Start the proxy and query `/healthz` and `/readyz`.
4. Observe `/healthz` and the proxy listener become available, while `/readyz` returns 503 and Kubernetes marks the pod unready.
5. With a single replica, the Service has no ready endpoint even though non-Codex provider routes are configured.

### Version

v2.59.0 (`nordz0r/opencodex` image)

### Operating system

Debian 12 / Kubernetes (single-node k3s)

### Provider and model

Codex account pool; other configured providers include Anthropic, Google Antigravity, OpenAI, and xAI routes.

### Logs or error output

```shell
🚀 opencodex proxy running on http://localhost:10100
GET /healthz      → health check
GET /api/*        → management API
[opencodex] catalog sync: ... Codex accounts ... need reauthentication

GET /readyz -> 503
Kubernetes: Startup probe failed: HTTP probe failed with statuscode: 503
Pod: Running, 0/1 Ready, no process crash
```

After changing the Kubernetes probe to `/healthz`, the pod becomes `1/1 Ready`; authenticated catalog and non-Codex inference checks succeed while the Codex pool remains partially degraded. This indicates the process is viable and the account-pool state is too broad a readiness gate.

### Screenshots and supporting files

Not applicable.

### Redacted configuration

```json
{
  "deployment": "single replica",
  "readinessProbe": "/readyz",
  "livenessProbe": "/healthz",
  "codexPool": "multiple OAuth accounts, some requiring re-authentication",
  "otherRoutes": "configured non-Codex providers"
}
```

### Checks

- [x] I searched existing issues and documentation.
[...]

Issue #4952, opened 2026-09-17
[Bug]: Model denial evidence survives same-ID reauthentication without a cached roster


lidge-jun:
### Client or integration

Codex CLI

### Area

Authentication and account pool

### Summary

An authenticated flagship model refusal is remembered for six hours under account ID and model ID, without credential generation. Reauthenticating the same internal account can leave the previous credential's denial active, avoiding an account that now has access. Expected: evidence from the old credential stops influencing the replacement, and late old replies cannot create or clear replacement evidence.

The release owner explicitly deferred this to the next round's first item. Compared with the earlier repeated-400 behavior, the current routing is an improvement; empty-filter restoration prevents a hard outage. A generation-aware change and its late-response races should be reviewed without release pressure.

### Reproduction

Static, constructible trace at a0f611d4aceb9476d44268e43722273b7b211846; not a claimed live reproduction:

1. With no roster cached, a pool account receives an authenticated unsupported-model refusal for gpt-5.6-sol. tests/codex-integration/codex-model-denial-evidence.test.ts:52 covers recording without roster evidence.
2. src/codex/observed-model-denials.ts:51-80 stores account/model plus a six-hour expiry, without credential generation.
3. Reauthenticate the same pool account. src/codex/auth-api/login-flow.ts:192 and :374 reuse the internal ID and write the replacement. src/codex/account-store.ts:160 increments generation; model-entitlements.ts:565 resolves pool:generation:upstream-account identity.
4. Existing-account reauth keeps pickerVisibilityChanged false (login-flow.ts:350), so catalog convergence returns without refresh (:158).
5. The only production account-wide forget call is model-entitlements.ts:884, inside a condition requiring an OLD cached roster with changed identity (:877-884). With no prior roster it cannot run.
6. cachedDeniedCodexAccountIdsForModel unions the unversioned refusal (:1300). A later positive roster removes the account from the returned denied set (:1292) but does not delete the six-hour record. After the five-minute roster expires, the old denial can reappear.

Fix direction: bind evidence to the credential identity captured from the dispatched auth context; validate against current identity after the exclusion-read fence. Make both denial writes and successful-response clears generation-aware. Required race: an in-flight generation-G refusal or success arrives after G+1 is saved and must neither overwrite nor clear G+1 evidence. Keep cache-only routing, no extra upstream fetch, current TTL/entry bound, positive-roster precedence and empty-filter restoration.

### Version

[...]

PR #5092, opened 2026-09-18
fix(codex): scope model denial evidence to the credential generation

abhisheksharma2411:
Closes #4952, following the fix direction in the issue.

## The bug

A refusal is remembered for six hours under `(accountId, modelId)` with **no credential generation**. Re-authenticating the same pool account keeps the internal id and increments the generation — and can swap the subscription underneath it — so the old credential's refusal keeps steering routing away from an account that now has access.

The account-wide forget that would have covered this is conditional:

```
model-entitlements.ts  — forgetObservedCodexModelDenialsForAccount(accountId)
                         guarded by a condition requiring an OLD cached roster
                         with changed identity
```

With no roster cached it cannot run — and that is the usual state on the flagship path, because nothing there refills the five-minute roster cache. So the one cleanup that existed is unavailable in exactly the situation the issue describes.

## The fix

Entries carry the generation they were observed under, and three fences use it:

| | behaviour |
|---|---|
| **read** | evidence whose generation is no longer live is **deleted**, not merely skipped — the account has reauthenticated, so the row can never become relevant again |
| **write** | a record from an older generation cannot overwrite newer evidence |
| **clear** | a clear from an older generation cannot delete newer evidence; an **equal** generation still clears |

The read fence is what makes this self-healing: it does not depend on the conditional forget firing.

Both races the issue names are covered. A generation-G refusal arriving after G+1 is saved neither resurrects nor rewrites G+1's entry; a generation-G success arriving after G+1 was refused cannot re-admit an account the current credential has just been refused by. Equal-generation clears still work, because that is the ordinary "this account just served this model" case.

Evidence that cannot name a generation is **dropped** rather than attributed to whatever is current. Main-account contexts carry no pool generation — and already carry no account id — so nothing changes for them.

## Constraints kept

Cache-only routing (no added upstream fetch), the six-hour TTL, the 512-entry bound, positive-roster precedence, and empty-filter restoration are all untouched. This remains an ordering preference, not an eligibility filter.

The liveness predicate is **injected** rather than imported, so `observed-model-denials` stays a leaf module — a unit test of the map shouldn't have to stand up the credential file. Production wires `isCodexAccountGenerationLive`, which already answers exactly this question for the account store.

## Verification

[...]

PR #5199, opened 2026-09-19
fix: stop one account failure from taking the whole proxy out of service

lidge-jun:
## Summary

Three layers of the same question: how far should one account's failure be allowed to spread?

| Commit | Issue | Change |
| --- | --- | --- |
| `ec96bbbde0` | #4212 follow-up (carries #5191) | `poolAccountDto` reports the reauthentication cause the probe observed instead of inferring it from overlapping booleans |
| `8a44021ad9` | #5181 | A local catalog-sync warning no longer puts `/readyz` into a terminal failed state |
| `fa9143bbcb` | #4952 (carries #5092) | Model denial evidence is scoped to the credential generation |
| `1e9b42d0f2` | #4952 | Fixes three defects in that carry: discarded `main-pool` evidence, a stale write that could evict valid evidence, and a read fence that deleted evidence it could not prove was superseded |

Closes #5181
Closes #4952

### #5181 — readiness

`/readyz` went terminal-failed on any nonempty `warning` from the post-startup Codex sync. A warning there describes artifacts written into the **local Codex home**: no catalog source so Codex keeps its native catalog, combos omitted from the catalog, a conversation-history relabel left to Codex's own writer, or a caught catalog-refresh exception after which config injection still runs and still reports its own `ok`. The sync continues past every one of them.

The reporter's single-replica Kubernetes deployment lost its only Service endpoint while every non-Codex route stayed healthy. Switching the probe to `/healthz` restored service, which is their own evidence that the process was viable.

`ok` and `warning` are separate fields because they answer separate questions. `ok` is the sync's verdict on the essential work — the write admission and the config injection. The gate now follows `ok` alone.

**The narrowing is only to `warning`.** A throw, a `null` outcome and `ok !== true` stay terminal, so the service-home write refusal, the external-provider injection failure, the config/integrity preflight refusal and the final injection failure all still fail readiness. Readiness also still stays *pending* until the sync **settles**, which is the race the gate was built for. This is the boundary the Claude Code roster reconciliation already has — it delays the ready transition without being allowed to fail it — applied to the catalog sync.

**The brief for this work was wrong about the mechanism, and that is recorded here rather than papered over.** The "Codex accounts … need reauthentication" line in the report comes from `warnGatedNativeSuppressedOnce` (`src/codex/catalog/gated-native-warn.ts`), a bare `console.warn` that never enters the structured result, and per-account pool refresh failures are caught and returned as `null` by `model-entitlements.ts` rather than propagating. So "a pool account needs reauth" is **not** by itself a readiness failure. The account-level fault that does reach the gate is a native **main** credential refresh failing outside that per-account catch: it rejects the entitlement gather, surfaces as the catch-all `catalog sync skipped` warning, and permanently un-readies the process. Same blast radius, different route, and closed here.

Also worth recording for whoever reads this next: `scripts/ci/docker-smoke.ts` runs in `runtimeRole: "hub"` and pre-writes a synthetic catalog, so it exercises the skip path and never the standalone-container path this issue is about.

### #4952 — denial evidence

`fa9143bbcb` is @abhisheksharma2411's PR #5092, carried unchanged. `1e9b42d0f2` fixes three defects that exact-head CI was green on:

1. **`main-pool` evidence was discarded.** Both call sites pass `kind === "pool" ? generation : undefined`, and the carried code dropped any refusal that could not name a generation. A `main-pool` context — the stored main login in rotation — has a real `accountId` and no *pool* credential generation, because its credential lives in `auth.json`. For that account #4906 reverted: the pool would re-send the model the login had just refused, on every request. A refusal with no generation is now account-scoped, which is the rule `core-codex-account.ts` already uses for the quota writer.
2. **The stale-write fence could evict valid evidence.** It only rejected a late refusal when the same key already held newer evidence. With no entry for its own key the stale row was inserted, and at the 512-entry bound that insert evicts the oldest valid row, which no read fence can restore. The write now rejects a non-live generation before touching the map.
3. **The read fence deleted rows it could not prove superseded.** It validated liveness for every non-expired entry before checking the model or the caller's exclusion set, and deleted anything not-live. `loadCodexAccountRecordStore` catches a read failure and returns `{}`, so an unreadable store deleted valid evidence permanently. The reader now matches the model and applies the exclusion set first, and *skips* a non-live row instead of deleting it.

Each liveness check also reloaded and reparsed the whole credential store per row on the request path; the seam is now a factory that opens once per lookup over one `loadCodexAccountRecordSnapshot`, and is not opened at all when no matching row carries a generation.

Cache-only routing, the six-hour TTL, the 512-entry bound, positive-roster precedence and empty-filter restoration are all unchanged. A live-generation refusal still denies.

### Not in this PR

**#5019 was dropped deliberately.** It is already implemented by open PR #5024 (`codex/lane-t-request-owned-main`) from the same maintainer, touching the same `resolveCodexAuthContext` path. A second implementation here would be a competing change to the credential fence, not a contribution. Separately, review of that path found the change is not acceptable on static evidence alone: relaxing the main-liveness predicate inside the *stateful* resolver mutates `activeCodexAccountId`, cooldown identity, and Direct/Pool/API independence before the selected-main branch can divert, so it needs a read-only preview seam, exact-head hosted tests and the security review `MAINTAINERS.md` requires for credential changes. That belongs in #5024's lane.

### File-size ratchet

`tests/codex-integration/codex-auth-api.test.ts` is capped at 6549 and sits at 6522. #5191 added 38 lines inline, which would have taken it to 6560 and broken the ratchet for this branch and every branch cut from `dev` afterwards. The two cases are registered from `tests/helpers/pool-reauth-cause.ts` through the same `registerXCases` seam the file already uses twice, leaving it at 6525 with the assertions byte-identical. `tests/server/server-live.test.ts` is at its 2253 cap exactly and did not grow.

## Verification

**No local execution.** This lane is under an absolute no-local-run constraint: no test suite, no individual test, no `typecheck`, no build, no install, no `ocx`. A previous local run in this repository destroyed real `~/.opencodex` data. Nothing here was executed locally, and no runtime evidence is claimed from this machine.

Verification was static, plus adversarial review, plus hosted CI at the exact head:

- **Source-oracle tracing** for every claim above: the four `warning` assignment sites in `src/codex/sync.ts`, the per-account `catch` in `model-entitlements.ts` that swallows pool refresh failures against the native-main branch that does not, the `main-pool` shape in `src/codex/auth-context.ts`, and `readCodexAccountRecord` being exactly `loadCodexAccountRecordStore()[id]`.
- **Four parallel adversarial reviews** (`gpt-5.6-sol`, high effort), each asked whether the change weakens a production contract and whether a real outage still fails. They refuted the brief's stated mechanism for #5181, found all three #5092 defects fixed in `1e9b42d0f2`, and produced the #5019 deferral.
- **Line counts checked against `tests/fixtures/file-size-baseline.json`** for every touched file, since caps only move downward.
- **Hosted CI at this exact head is the only execution evidence.**

Contract documentation was updated with the behaviour change: the `src/server/readiness.ts` header, the driving comment in `src/codex/desired-state.ts`, and the *Startup readiness* section of `structure/catalog.md`.

Tests changed or added: the warning case in `tests/server/proxy-liveness.test.ts` now expects ready across all four real warning producers, with new cases proving `ok=false` and a missing `ok` still fail; `tests/server/server-live.test.ts` gains an `ok=false` HTTP case so 503-after-settle stays covered; `tests/codex-integration/codex-desired-state.test.ts` pins the boundary at the caller; and `codex-model-denial-evidence.test.ts` gains cases for account-scoped `main-pool` evidence, the capacity-eviction race, the exclusion fence, and one store open per lookup.

### Rebased onto `c81d43053b`

An earlier head of this branch failed `file-size ratchet: repository` on two files it never touched — `src/codex/history-provider.ts` (2009 lines, no cap) and `tests/server/server-combo-failover-e2e.test.ts` (4192 against a 4166 cap) — plus three macOS Reserve cases. All of those were inherited from `dev` and are fixed on `dev` by #5201, which split the rollout reader into `src/codex/history-rollout-read.ts` (1849 lines now) and moved the combo cases into `tests/helpers/combo-tool-routing-cases.ts` (4153 lines now).

This branch is rebased onto that tip. The rebase was clean with no conflicts, commit authorship is preserved for both carried commits, and none of the five files #5201 touched overlaps the seventeen this PR touches. Every file this PR does touch is within its cap at the rebased head, with `codex-auth-api.test.ts` at 6525 of 6549 and `server-live.test.ts` at exactly its 2253 cap.

## Checklist

Co-authored-by: luvs01 <27862058+luvs01@users.noreply.github.com>
Co-authored-by: Abhishek Sharma <abhicse24@gmail.com>

## Summary by CodeRabbit

- **Bug Fixes**
  - Servers now remain ready when startup synchronization succeeds with recoverable warnings.
  - Reauthentication status now distinguishes unauthorized access from expired or failed credentials.
  - Model-access denial records are tied to the correct credential generation, preventing stale credentials from affecting replacement accounts.
  - Account exclusions are respected when evaluating observed model denials.

- **Tests**
  - Added coverage for startup readiness warnings, reauthentication causes, and credential-generation-specific denial behavior.

- **Documentation**
  - Clarified startup readiness behavior and the meaning of recoverable synchronization warnings.


lidge-jun:
## 리뷰 · 우선순위 64 / 80

이 PR은 계정 하나가 고장 나도 프록시 전체가 같이 멈추지 않게 막아요. 이미 `dev`에 머지됐고, 이 커밋의 검사는 통과했어요.

켜질 때 Codex 목록 맞추기가 경고만 남기면, 예전 `/readyz`는 실패로 굳었어요. 서버가 하나뿐인 쿠버네티스에서는 그 실패가 서비스 주소를 지워요. 이제는 맞추기 결과의 `ok`가 참이면 준비된 것으로 봐요. 경고는 내 컴퓨터의 Codex 폴더에 적힌 파일이 조금 나쁘다는 뜻이에요. Codex가 아닌 다른 길로 가는 요청은 그대로 받아요. 맞추기가 중간에 멈추거나, 결과가 없거나, `ok`가 참이 아니면 예전처럼 실패예요.

모델이 거절된 기록에는 로그인 세대 번호를 붙여요. 같은 계정을 다시 로그인하면 계정 이름은 그대로이고 세대만 올라가요. 죽은 세대의 거절은 다음 요청에서 건너뛰고, 늦게 온 옛 거절은 저장하지 않아요. 메인 로그인이 계정 회전에 끼는 `main-pool`은 풀 세대 번호가 없어요. 그 거절은 버리지 않고 계정 단위로 남겨요. 버리면 방금 거절한 모델을 요청마다 다시 보내게 돼요.

계정 목록은 재로그인이 필요한 이유를 겹친 예/아니오로 짐작하지 않아요. 조사 함수가 본 이유를 그대로 적어요. 사용량 조회가 거절하면 `quota_unauthorized`, 토큰 갱신이 죽으면 `refresh_failed`예요.

src/codex/auth-api/account-list.ts:113 - 강제 새로고침 직후에는 이유가 맞아요. 그런데 그 조사가 계정에 재로그인 표시도 켜요. 사용량 기록이 5분(`POOL_CACHE_TTL`) 안이면 `fetchPoolAccountQuota`(426줄)는 이유 없는 옛 사용량을 돌려줘요. 목록은 표시만 보고 113줄에서 이유를 `refresh_failed`로 바꿔요. 새로고침 없이 목록만 다시 보면, 사용량 거절이 토큰 갱신 실패로 보여요. 새 테스트는 `?refresh=1`이랑 갱신 POST만 봐서 이 길은 안 타요.

메인테이너의 판단이 필요한 지점

`src/codex/sync.ts`는 목록 갱신 중 어떤 예외든 경고 문장으로 남기고, 설정 넣기가 성공하면 `ok`는 참이에요. 준비 검사는 그 문장을 읽지 않아요. 테스트에 적힌 네 문장만 통과시키는 게 아니에요. 메인 로그인 갱신 실패처럼 이번 이슈에서 통과시키려던 경고와, 나중에 추가될 다른 경고가 같은 규칙으로 준비 상태가 돼요.

`main-pool` 거절은 세대가 없어서, 메인 로그인을 다시 받아도 최대 6시간은 그 모델을 피해요. 세대 검사를 빼 둔 쪽은 거절 직후 재전송을 막아요. 다시 로그인하면 옛 거절을 잊으라는 #4952는 이 계정에는 아직 없어요.

#5191은 닫으면 안 돼요. 두 번째 커밋 `8ba026a591`이 이 머지에 없어요. 짧은 토큰 갱신 실패는 재로그인으로 올리지 않는 것, 그리고 저장해 둔 검증 오류 `http_status:401`/`403`을 `unauthorized`/`forbidden`으로 보여주는 것이에요.

너의 추천

5분 캐시로 목록을 다시 읽을 때도, 조사 때 본 이유를 유지하세요. 재로그인 표시만 켜져 있고 이번 결과에 이유가 없으면 `refresh_failed`로 덮지 마세요. #5092는 이 PR에 실려 들어왔고, 그 안의 빠진 세 가지는 뒤 커밋이 고쳤으니 닫았어요. #5191은 두 번째 커밋을 따로 살리세요.

이 댓글은 grok-bot이 작성했습니다

abhisheksharma2411:
All three defects in the #5092 carry are real and the fixes are right. Taking them one at a time, because two of them are mistakes I'd make again without writing down why.

**1. `main-pool` — I checked the wrong half of the union.** I looked at the `main` variant, saw `accountId: null`, concluded "main never records evidence anyway", and moved on. I never enumerated the *third* shape. `main-pool` has a real `accountId` and no pool generation because its credential lives in `auth.json`, so my `typeof generation !== "number"` early-return silently dropped exactly the refusals #4906 exists to remember — and reverted it for that account on every request. Scoping a generation-less refusal to the account, matching what `core-codex-account.ts` already does for the quota writer, is the right answer and it was already in the codebase to copy.

**2. The stale-write fence.** I fenced on "does this key already hold newer evidence", which is only half the question. With no entry for its own key the stale row goes in, and at the 512-entry bound that insert evicts the oldest *valid* row — which no read fence can put back. Rejecting a non-live generation before touching the map is correct; my version protected the row and left the container unprotected.

**3. The read fence deleting rows it could not prove superseded — this is the one I got most wrong, and I argued for it in the PR body.** I chose delete-over-skip deliberately and described it as making the fix "self-healing". It isn't self-healing, it's destructive, and I gated it on a predicate that cannot tell "proven superseded" from "couldn't tell":

```rust
// account-store.ts
} catch {
  backupInvalidConfig(path);
  return {};      // <- every isCodexAccountGenerationLive() now returns false
}
```

One unreadable or corrupt `accounts.json` and my reader would have deleted **every** account's denial evidence, permanently, on the next lookup. Matching the model and the exclusion set first and then *skipping* a non-live row is right on both counts: it does less work and it makes the failure recoverable.

The general form, which is what I'll carry forward: **a destructive action must not be conditional on a predicate that fails closed to "false" when it simply couldn't determine the answer.** Skip is recoverable, delete is not, and the asymmetry should decide the default.

Also fair: reloading and reparsing the whole credential store per row on the request path was a straightforward performance defect I didn't consider at all — the seam I introduced invited it, and a factory opening once per lookup is the obvious shape in hindsight.

Verified `Co-authored-by: Abhishek Sharma` is on `09698b9b3a`, so the attribution came through. Thanks for carrying it rather than bouncing it back, and for writing the three defects up in enough detail that they're learnable rather than just corrected.

No disagreement with any of it.

Issue #5180, opened 2026-09-19
Command Code 429 rate_limit_exceeded exhausts retries on long muse-spark turns (v2.59.0)


0disoft:
### Client or integration
Codex App

### Area
Provider adapter

### Summary
Long Codex App turns via `commandcode/meta-muse-spark-1.3-contributor` (reasoning high) repeatedly die with `exceeded retry limit, last status: 429 Too Many Requests`. The failing request reports `errorCode=rate_limit_exceeded`, `closeReason=non_stream`, `upstreamError=Provider error 429: Upstream model provider is temporarily unavailable. Please try again in a moment.` I expected a transient provider throttle to be retried with cooldown / Retry-After awareness instead of killing the whole workflow, forcing manual resume.

Related but distinct from #532 (Kiro `USER_REQUEST_RATE_EXCEEDED` with independent per-request retry budgets). This is Command Code + muse-spark, single explicit-provider route, no concurrent-account pool involved.

### Reproduction
1. Start opencodex proxy/dashboard v2.59.0 on Windows, Codex App connected via `localhost:10100`.
2. Select `commandcode/meta-muse-spark-1.3-contributor`, reasoning_effort=high, explicit-provider routing.
3. Run a long multi-tool turn (blanche Go queue work: read files, run commands, edit/verify). The turn makes several sequential model + tool calls.
4. Observe intermittent failure: `exceeded retry limit, last status: 429 Too Many Requests`.
5. Open dashboard log detail for the failed request (example below). Retry manually after cooldown and it often succeeds.

Observed on 2026-09-19 ~23:08 KST (Asia/Seoul). Same pattern recurred across separate threads on the same night (including a thread doing only GitHub fetch calls), so it is not specific to one conversation.

### Version
v2.59.0 (dashboard header shows `opencodex v2.59.0`)

### Operating system
Windows 11 x64

### Provider and model
commandcode / meta/muse-spark-1.3-contributor (requested as `commandcode/meta-muse-spark-1.3-contributor`, effectiveEffort=high)

### Logs or error output
```text
time: 2026-09-19 23:08:32 +09:00
request: ocx-c28be5ec8a5880f6adbc2cbec4779f52
conversation: 9751b9b507b784dda09d5d817898616e
model: meta/muse-spark-1.3-contributor
provider: commandcode (Command Code - API)
reasoning: high (reasoning_effort=high)
status: 429
durationMs: 2351
errorCode: rate_limit_exceeded
[...]

PR #4989, opened 2026-09-17
fix(responses): gate post-header reset recovery on SSE protocol state

lidge-jun:
## Summary

Carries #3389 by @Yum-wu and replaces its byte-consumption gate with protocol evidence.

The original implementation retried when a downstream body reader had consumed zero bytes. Bun can discard already-buffered SSE chunks when a socket resets, so that condition could resend a turn after provider-executed tools or stored output had already committed.

This change uses the existing Responses preflight and shared SSE inspector through a deferred body wrapper. `handleResponses` returns HTTP response headers before the first SSE event; inspection starts when downstream body consumption begins. One replacement is eligible only after `response.created` was parsed and before any output item, delta, unknown, opaque, malformed, done, or terminal payload. A reset before `response.created` and every reset after protocol commitment retain the original failed-stream behavior. Upstream WebSocket responses bypass this wrapper so their quota, stage, and bounded eager-relay identity remain attached to the original response. Native Chat remains unchanged because it does not expose equivalent positive protocol evidence at the dispatch gate.

The accepted replacement retains the original contribution's sound boundaries: one remaining physical send, the current credential binding, HTTP-only dispatch, rejection of non-success or incompatible responses, cancellation cleanup, redacted diagnostics, and propagation of the original stream error when recovery fails.

This intentionally narrows and reverses the decision recorded in `devlog/_fin/260703_sse-midstream-reset-tail/00_plan.md`. That rule still applies whenever positive protocol evidence is absent or output has committed. The current contract and rationale are recorded in `structure/transports/streaming-health.md` and ADR-3389.

Refs #3384. This does not use a closing reference because that issue was closed as not planned after failing issue-template enforcement.

## Verification

Local verification was not run because this lane forbids running any local suite, typecheck, build, or install; hosted CI is the executable verification for this change.

Static verification performed:

- `git diff --check` passed at exact head.
- Audited `createSseInspector`, `preflightComboStreamResponse`, the deferred body wrapper, and the WS response identity markers to confirm protocol inspection begins on body consumption without stripping transport metadata.
- Existing hosted regressions cover progressive DeepSeek HTTP delivery, Codex WS bounded eager relay selection, WS quota attribution, stale credential fencing, and native WS terminal reporting.
- Added focused coverage for created-only recovery, tool/output commitment, reset-before-created refusal, malformed and done payload refusal, replacement validation, send counting, cancellation gates, and HTTP-only dispatch.
[...]

Issue #4191, opened 2026-09-10
[Bug]: no SSE fallback after an established Codex WebSocket dies mid-turn (prelude-timeout half resolved)


mehmeemow:
### Client or integration

Codex App (also reproduced from Codex CLI when resuming the same problematic thread).

### Area

Streaming

### Summary

A long-running Codex conversation became effectively unusable **only while traffic was routed through OpenCodex**. The same account, model, machine, repository and thread work normally again as soon as the OpenCodex proxy is disabled/bypassed.

This does not look like a generic network outage because the failure is both **thread-specific** and **proxy-path-specific**.

With OpenCodex enabled, the problematic parent thread repeatedly failed with either:

```text
stream disconnected before completion:
Upstream stream terminated unexpectedly:
codex websocket closed before a Responses terminal event
(close 1006 Connection ended)
```

or:

```text
stream disconnected before completion:
Upstream stream terminated unexpectedly:
codex websocket response prelude timed out
```

After OpenCodex was disabled/bypassed, the **same existing thread resumed and worked normally immediately**, without fork, rollback, or compaction recovery.

Expected behavior: OpenCodex should transparently proxy the same Codex Responses request, or safely fall back to HTTP/SSE when the upstream WebSocket path cannot complete.

This appears related to one or more WebSocket relay edge cases such as large replay payloads, loss of upstream close code/reason, the response-prelude timeout, or lack of safe HTTP/SSE fallback after an already-open WebSocket terminates.

Related issues:

- #2471 — large Codex WS upstream request / 16 MiB boundary, missing SSE fallback, close code/reason loss.
- #4083 — fixed response-prelude timeout can truncate legitimate slow-start Responses requests.
- #3976 — request to make the response-prelude timeout configurable.

This report adds a real-world A/B observation: **the exact same previously failing thread becomes healthy immediately when OpenCodex is bypassed**.

### Reproduction

1. Run Codex Desktop on Windows with Codex traffic routed through OpenCodex to the normal ChatGPT/Codex upstream.
2. Open/resume an existing long-running conversation.
3. Observe repeated `stream disconnected before completion` failures for many hours.
4. Switch the host network to a mobile hotspot. Observe the same failure.
5. On the same machine, start Codex CLI and create a fresh conversation. The fresh conversation replies normally (`Hello!`).
6. In Codex CLI, resume the **same problematic existing conversation**. The failure reproduces.
7. While the parent conversation is failing, previously spawned sub-agents can continue executing normally.
[...]

PR #5266, opened 2026-09-20
One stage, cause and resend vocabulary for a failed request

lidge-jun:
## Summary

Roadmap items 7 (retry stage table) and 14 (one event model) want the same substrate, so this lands it as one module rather than two definitions that typecheck separately and contradict each other in the merge.

`src/lib/request-failure-model.ts` states three things once:

- **Stage** — how far a failed exchange got, ordered by what the downstream caller observed: `pre-header`, `headers-only`, `protocol-prelude`, `semantic-output`, `side-effect`, `terminal`. A turn that settled carrying no output did not reach `terminal`; it stalled at `protocol-prelude`, because the caller saw no answer.
- **Cause** — one closed dictionary. `rate-limit`, `quota-exhausted`, `policy-refusal` and `ciphertext-refusal` are four separate members because their remedies are four different actions: wait, change account, change the prompt, drop stale encrypted state.
- **Resend permission** — derived from three small per-member facts, not written out as a stage-by-cause matrix. A matrix that size has to be re-derived by hand whenever a member is added, and the cell nobody revisited is how two individually correct branches merge into a wrong table.

Once the caller has observed output or an externally visible effect, no cause permits an automatic resend. A cause whose upstream execution state is unknown is not made replayable by having budget left. `refused-ambiguous` forbids an *automatic* resend; it does not forbid a bounded recovery a maintainer explicitly opted into, which is why funding follows the disposition rather than the permission — the reset replay, the empty-completion rebuild and the transient 5xx ladder all really send and all still name the allowance they draw on.

The module adds no record store. The durable shapes stay in `src/usage/log.ts` and projections read them structurally.

**Two restatements removed, both live defects.** `AttemptRecoveryKind` was written twice, as a union and as the read-back whitelist `normalizedAttempt` filters against; a member added only to the union compiles, is written to disk, and is dropped on the next read, so the row loses the field that says why it recovered. And `recoveryClass()` ended in `default: return "other"`, so a recovery kind added later compiled cleanly and vanished into a bucket an operator cannot act on. Both are now derived and total.

**A relayed ciphertext rejection now recovers.** An OpenAI-compatible gateway puts the real payload inside its own error message, so the single-shot sanitized rebuild never matched behind one and a turn carrying a stale reasoning blob failed outright. Exactly one identity is admitted through the wrapper: an embedded `invalid_request_error` carrying `invalid_encrypted_content`. The generic classifier is deliberately not re-run on the embedded payload, and a gateway envelope is now decided *only* by that payload — the pre-existing anchored-wording checks scan the whole message, and a gateway quotes the upstream's message inside its own, so a relayed caller mismatch would otherwise have satisfied the #4469 identity and gained a resend the strict check exists to withhold.

**Operator-visible metric changes.** `quota`, `policy` and `ciphertext` are new `recovery` label values. Two series change meaning, both corrections: `opaque-blob-rejection` moves from `payload` to `ciphertext` (the payload was fine, the stale encrypted state was not), and `console-go-upload-retry` moves from `payload` to `transient` (the gateway accepts the byte-identical body seconds later, so nothing about the payload was wrong). Label cardinality is unchanged in kind: every value still comes from a frozen roster.

Relates to #4191 and #5180. Both are addressed in part only and are deliberately not closed: the WebSocket failure projection and the rate-limit/quota separation land here, while the SSE fallback and the shared cooldown are transport and routing changes that are not in this branch. Carries work from #5245, narrowed; the original pull request is left open for the coordinator.

Dispositions for everything considered and not landed — #4942, #4989, #2366, #3748, #3983, #5063, and a dashboard roster defect found while mapping the substrate — are recorded in `devlog/_plan/260920_meaning_preservation_batch/020_lane_c.md` with reasons.

## Verification

Static source review plus hosted CI at this exact head. Per the batch execution constraints for this lane, **no local suite, individual test, typecheck, build, install or live run was performed**, and none may be read as passing.

Checked statically on this branch:

- every assertion in the two new test files was re-derived by hand from the declared tables; the only recovery kinds whose Prometheus class changes are the two named above.
- every `satisfies Record<Union, ...>` added here is total over its roster, and every value it produces is a declared member of the target vocabulary.
- `ATTEMPT_RECOVERY_KIND_ROSTER` is declared before both its derived type and its runtime `Set`, so the derivation introduces no temporal-dead-zone hazard.
- `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json` agree key for key, and both new test files sit in the domain they are registered to.
- no file touched here carries a `tests/fixtures/file-size-baseline.json` cap. The new ciphertext cases went into a sibling file rather than into `responses-opaque-blob-recovery.test.ts`, which sits 148 lines under the new-file threshold; `src/server/index.ts` has one line of headroom and is untouched.
- two restated counts that the diff would have invalidated were found and **derived instead of bumped**: the Prometheus sample count in `management-metrics-export.test.ts` (previously the literal `453`, now computed from the closed rosters and additionally asserted to be unchanged by 64 requests carrying distinct identities), and an "eight coarse classes" sentence in `structure/gui-and-management-api.md`.

The regressions are written so a successful request cannot satisfy them. The stage-by-cause cases loop over the declared rosters, so they have no cell to miss and cannot go stale when a member is added; the ciphertext cases are mostly negative, because recognising the wrapper is the easy half and admitting only the coded identity through it is the half a broad implementation gets wrong.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Recovery metrics now distinguish quota, policy, and ciphertext failures from other recovery outcomes.
  - Request failures are classified by observed stage, failure cause, and whether retrying is permitted.
  - Relayed ciphertext-rejection responses from supported gateways are recognized for opaque-blob recovery.

- **Documentation**
  - Management API documentation now lists recovery metric-label values and their meanings.
  - Request metrics documentation now references the maintained recovery-label vocabulary.


lidge-jun:
## 리뷰 · 우선순위 66 / 80

이 PR은 실패한 요청을 한 가지 말로 정리하는 기반을 넣습니다. 실패가 “어디까지 갔는지(stage)”, “왜 실패했는지(cause)”, “다시 보내도 되는지(resend)”를 `src/lib/request-failure-model.ts` 한곳에서 정합니다. 예전에는 같은 뜻을 여러 곳에 따로 써서, 한쪽만 고치면 타입이 통과해도 실제로는 틀린 표가 생기는 문제가 있었습니다. 이번엔 표 전체를 손으로 다시 쓰지 않고, stage의 관찰 정도와 cause의 증거·성향에서 재전송 허가를 계산합니다.

같이 고친 실사용 결함도 있습니다. OpenAI 호환 게이트웨이(LiteLLM 등)가 암호문 거절을 자기 에러 메시지 안에 감싸면, 예전엔 복구가 한 번도 안 됐습니다. 이제는 그 껍질 안에서 `invalid_request_error` + `invalid_encrypted_content`만 인정하고, 껍질 전체를 일반 문장 검사에 넘기지 않습니다. 그래서 게이트웨이가 인용한 다른 문구 때문에 잘못된 재전송이 열리는 길을 막습니다. 메트릭에서는 `quota` / `policy` / `ciphertext` 라벨이 생기고, `opaque-blob-rejection`은 payload→ciphertext, `console-go-upload-retry`는 payload→transient로 의미가 바뀝니다(종류 수는 그대로). Codex WebSocket 실패도 같은 어휘로 투영합니다. base는 `dev`이고, #4191·#5180은 일부만 다루며 닫지 않습니다. #5245 작업은 좁혀 가져왔고 원본 PR은 조율용으로 열어 둡니다. 작성자 말대로 로컬 스위트는 돌리지 않았고, 호스트 CI에 맡깁니다.

라인 - `src/server/responses/codex-ws-wire.ts` `projectCodexWsFailure` / `after-response-started`: `response.created`만 중계돼도 `relayedEvents > 0`이라 `semantic-output`로 찍힙니다. 같은 모듈 주석과 공유 어휘는 `response.created`를 `protocol-prelude`(제어 이벤트)로 둡니다. 재전송 동작은 WS가 이미 send 이후 재전송을 막아서 안 바뀌지만, HTTP와 맞춰 비교하려는 로그·메트릭 단계는 틀어집니다. 출력 이벤트를 따로 세는 편이 어휘와 맞습니다.

라인 - `tests/lib/failure-stage-model.test.ts` (대략 64–66행 주석): “terminal + rate-limit은 커밋이 없어서 다시내도 된다”고 적혀 있습니다. 모델에서 `terminal`은 `answer-delivered`라 항상 `refused-committed`입니다. 예시는 `headers-only` 같은 “아무것도 안 본” stage로 바꿔야 주석이 표와 같습니다.

라인 - 관련 PR `#5245` (`[WRONG BRANCH] …`): 이 브랜치가 좁혀 가져간 원본인데 아직 open입니다. 머지 후 landed-via로 닫을지, 조율자가 남길지 한 줄만 정해 두면 중복·무효 PR이 덜 쌓입니다.

메인테이너의 판단이 필요한 지점

WS에서 `response.created`만 온 실패를 이 PR에서 `protocol-prelude`로 고칠지, #4191 남은 일로 미룰지. 메트릭 라벨 의미 변경(특히 opaque-blob / console-go)을 운영·대시보드에 어떻게 알릴지. #5245를 이 PR 머지와 함께 닫을지.

너의 추천

어휘 한곳으로 모은 방향과 LiteLLM 감싸진 암호문 복구는 타당합니다. base `dev`도 맞습니다. 가능하면 WS stage 오분류를 이 PR에서 짧게 고치거나, 안 고치면 #4191에 “control-only vs output-bearing”으로 명시하세요. 테스트 주석 모순은 싼 수정이라 같이 고치는 편이 낫습니다. #5245는 머지 후 자신 있으면 landed-via로 닫으세요. CI가 전부 초록인지 확인한 뒤 머지하세요. preview deploy 이야기는 생략합니다.

이 댓글은 grok-bot이 작성했습니다

Issue #5213, opened 2026-09-19
`developer` messages are folded into the system prompt on every Chat host except api.openai.com


Android-Login-Analysis:
### Client or integration

OpenAI-compatible SDK

### Provider or upstream service

Any `openai-chat` provider whose base URL host is not `api.openai.com`, for example openrouter

### OpenCodex version

2.59.0 / commit `134c92a01b120162f00c7275189cc47858720379`

### Endpoint or capability

/v1/chat/completions, message role translation

### Current behaviour

A `developer` role message reaches the upstream as `developer` only when the provider base URL host is exactly `api.openai.com`. On every other OpenAI-compatible Chat endpoint the message text is appended to the system prompt and the message itself is not emitted, so the upstream never sees the `developer` role and the instruction loses its position in the conversation.

The host test is at `src/adapters/openai-chat/wire.ts:44`:

```ts
export function isNativeOpenAIChatTarget(provider: OcxProviderConfig): boolean {
  try {
    return new URL(provider.baseUrl).hostname === "api.openai.com";
  } catch {
    return false;
  }
}
```

`src/adapters/openai-chat/messages.ts:123` and `:131` hoist the text:

```ts
const nativeOpenAI = isNativeOpenAIChatTarget(provider);
...
const developerSystemParts = nativeOpenAI || chronologicalSystem
  ? []
  : context.messages
    .map(developerSystemText)
    .filter((part): part is string => part !== undefined && part.length > 0);
const systemParts = [
  ...(context.systemPrompt ?? []),
  ...developerSystemParts,
  ...(toolCatalogNudge ? [toolCatalogNudge] : []),
];
```

`src/adapters/openai-chat/messages.ts:156` then skips the original message, and the branch that
does emit rewrites the role:

```ts
if (msg.role === "developer" && !hasImages) {
  if (!nativeOpenAI && !chronologicalSystem) break;
  ...
  chatMsg = { role: nativeOpenAI ? "developer" : "system", content: text };
}
```

There is one narrow exception, `chronologicalSystem`, which applies only when the model is `deepseek-v4.1-flash` and the registry destination is `opencode-go`.

Two things are lost for every other host. The role itself, which the Chat Completions specification defines and which carries a different precedence than `system`. And the position, since a mid-conversation instruction is moved out of the timeline into the prompt preamble.

### Expected behaviour

[...]

Issue #5212, opened 2026-09-19
inline document bytes do not survive the inbound parse


Android-Login-Analysis:
### Client or integration

Codex CLI

### Provider or upstream service

openai-compatible, anthropic

### OpenCodex version

2.59.0 / commit `134c92a01b120162f00c7275189cc47858720379`

### Endpoint or capability

message content part translation

### Current behaviour

Two inbound parsers lose an attached document before any adapter runs, so no adapter can carry it even to a target that has a representation for it.

| # | inbound | caller sent | what reaches the adapter |
|---|---|---|---|
| 1 | `/v1/chat/completions` | `file` part with inline `file_data` | nothing, the part is dropped |
| 2 | `/v1/messages` | `document` block with base64 bytes | a text marker `[document: title]` |

Case 1. `src/chat/inbound.ts:75` walks the content parts and recognises three shapes:

```ts
if (typeof raw === "string") { ... continue; }
if (!isRec(raw)) continue;
if ((raw.type === "text" || raw.type === "input_text" || raw.type === "output_text") && typeof raw.text === "string") {
  blocks.push({ type: "input_text", text: raw.text });
  continue;
}
const imageUrl = imageUrlFromPart(raw);
if (imageUrl) { ... continue; }
const videoUrl = videoUrlFromPart(raw);
if (videoUrl) blocks.push({ type: "input_video", video_url: videoUrl });
```

A `{"type": "file", "file": {"filename": ..., "file_data": ...}}` part matches none of these. The loop reaches its end with nothing pushed and no else branch, so the part is dropped in silence:

```
grep -c "file_data" src/chat/inbound.ts
0
```

Case 2. `src/claude/inbound.ts:210` replaces the block with a marker:

```ts
case "document":
  // No Responses equivalent for raw document blocks; surface the title so the
  ...
  pending.push({ type: "input_text", text: `[document${typeof raw.title === "string" ? `: ${raw.title}` : ""}]` });
```

The same substitution appears at `src/claude/inbound.ts:57` for the other block position.

The marker is a real improvement over dropping the block, which is what #939 fixed, and this report does not ask for that to be reverted. The remaining gap is narrower. For a target that can hold the bytes, the marker is now the only thing sent, and the caller cannot distinguish
"the model read the document" from "the model was told a document existed".

Because both losses happen at the inbound, they apply to every routed target, not only to the ones that cannot represent a document.

### Expected behaviour

[...]

Issue #5211, opened 2026-09-19
tool choice policy is dropped on the Chat Completions path


Android-Login-Analysis:
### Client or integration

Codex CLI

### Provider or upstream service

openai-compatible

### OpenCodex version

2.59.0 / commit `134c92a01b120162f00c7275189cc47858720379`

### Endpoint or capability

tool calls

### Current behaviour

Two ways a caller restricts tool use on `/v1/chat/completions` do not reach the upstream.

| # | caller sent | forwarded | effect |
|---|---|---|---|
| 1 | `tool_choice` with `type: "allowed_tools"` | no tool choice, full tool list | the model may call a tool the caller excluded |
| 2 | `parallel_tool_calls: false`, provider metadata unset | no `parallel_tool_calls` | the upstream may emit concurrent tool calls |

Case 1. `src/chat/inbound.ts:204` handles the string forms and the `function` object form and has no branch for `allowed_tools`:

```ts
function toolChoiceToResponses(choice: unknown, body: Rec): void {
  if (choice === undefined || choice === null) return;
  if (choice === "auto" || choice === "none" || choice === "required") {
    body.tool_choice = choice;
    return;
  }
  if (!isRec(choice)) return;
  if (choice.type === "function") {
    ...
    body.tool_choice = { type: "function", name };
    return;
  }
  if (isRec(choice.function) && typeof choice.function.name === "string") {
    body.tool_choice = { type: "function", name: choice.function.name };
  }
}
```

An object with `type: "allowed_tools"` is a record, is not `type: "function"`, and has no `function` member, so it reaches the end of the function and `body.tool_choice` is never assigned. The identifier does not occur in the file:

```
grep -c "allowed_tools" src/chat/inbound.ts
0
```

Case 2 is a narrow branch, and it is **not** the case covered by #1650. That issue concerned a provider configured with `parallelToolCalls: false`, and #1662 resolved it by adding `pinParallelToolCallsFalse`. This is the branch where the provider sets no value at all and the
constraint comes from the caller's own request. `src/adapters/openai-chat.ts:251`:

```ts
if (tools) {
  if (provider.parallelToolCalls === false) {
    // NIM documents the Boolean defaulting to false and kimi rejects true; pin the
    // wire bit so Codex cannot opt in via request.options. ...
    if (provider.baseUrl === "https://integrate.api.nvidia.com/v1"
        || provider.pinParallelToolCallsFalse === true) {
      body.parallel_tool_calls = false;
    }
[...]

Issue #5210, opened 2026-09-19
tool declaration fields are dropped on the outbound adapters


Android-Login-Analysis:
### Client or integration

Claude Code

### Provider or upstream service

anthropic, google

### OpenCodex version

2.59.0 / commit `134c92a01b120162f00c7275189cc47858720379`

### Endpoint or capability

tool calls

### Current behaviour

Three fields a caller sets on a tool declaration do not reach the upstream. All three are lost at the outbound adapter, after the value has already been parsed and carried internally.

| # | route | field | forwarded |
|---|---|---|---|
| 1 | Messages to Messages | `tools[*].strict` | tool without `strict` |
| 2 | Messages to Messages | `tools[*].allowed_callers` | tool without `allowed_callers` |
| 3 | Messages to Gemini `generateContent` | `tools[*].strict` | no `toolConfig` mode expressing it |

Cases 1 and 2 are the same function. `src/adapters/anthropic.ts:850` rebuilds every tool from exactly three fields:

```ts
const converted = tools.map(t => ({
  name: toolNames.toWire(namespacedToolName(t.namespace, t.name)),
  description: t.description,
  input_schema: normalizeAnthropicInputSchema(t.parameters),
}));
```

For case 1 the value is present and deliberately kept. `src/claude/inbound-content-options.ts:37` reads it with a comment stating the intent:

```ts
// Anthropic opts into strict tool use explicitly, while Responses reads an
// omitted strict as permission to normalize the schema into strict mode. That
// turns an optional input_schema parameter into a required one and breaks the
// call, so carry the source intent instead of the destination default. A
// non-boolean value is not a valid Anthropic opt-in and must not become one.
strict: typeof raw.strict === "boolean" ? raw.strict : false,
```

`src/types/tools.ts:5` holds it, and the OpenAI Chat adapter forwards it at
`src/adapters/openai-chat/tool-schema.ts:438`:

```ts
...(t.strict !== undefined ? { strict: t.strict } : {}),
```

So the field survives the parse, reaches the adapter, is forwarded to one target, and is dropped for Anthropic, which is the target that defines it.

For case 2 there is no carrier at all. Across the whole tree the identifier occurs once:

```
grep -rn "allowed_callers\|allowedCallers" src/
src/claude/compatibility.ts:104:      if (nonDirectCaller(tool.allowed_callers)) codes.add("caller_mode");
```

[...]

PR #5271, opened 2026-09-20
fix: preserve caller constraints on the request path (#5211, #5210, #5212, #5213)

lidge-jun:
## Summary

Four defects where a request arrived carrying an explicit constraint and the proxy returned a
normal HTTP success after dropping it. A test that only asserts a successful tool call or a 200
cannot see any of them, so every regression here reads the serialized outbound request body.

**#5211 — tool choice policy on the Chat Completions path.** A `tool_choice` of type
`allowed_tools` is a record, is not `type: "function"`, and carries no `function` member, so it
fell past every branch of the Chat inbound translator and `body.tool_choice` was never assigned:
the upstream got the full catalogue and no choice at all. Chat nests the subset under
`allowed_tools` and names each entry under a member keyed by its own type, while the Responses
shape `mapToolChoice` reads carries `mode` and `tools` on the choice itself with a flat `name`.
Both levels are now flattened. An entry that cannot be named, or whose selector kind is neither
function, custom, nor a hosted type, is refused rather than skipped — dropping one widens the
subset the field was sent to narrow.

`parallel_tool_calls` had three provider states and two branches, in two places. The unset state
is the default for every provider that never configured the knob, and it dropped the caller's own
explicit `false` on the translated path and, from a second copy of the same branch, on the native
Chat passthrough. The decision now lives in one module both builders read. An explicit `true`
still omits the key, matching the configured opt-out, so strict hosts see no new knob. The NVIDIA
and `pinParallelToolCallsFalse` pins are unchanged.

**#5210 — tool declaration fields on the outbound adapters.** `strict` was kept deliberately by
the Messages inbound and already forwarded by the OpenAI Chat adapter, and dropped by Anthropic —
the target that defines it. It is now emitted when it is explicitly `true`; an unstated `strict`
stays absent, because the inbound records it as `false` and a `false` on the wire cannot be told
apart from silence. `allowed_callers` had no carrier at all, and the one occurrence of the
identifier in the tree raised a diagnostic that only becomes a refusal when the operator has set
`claudeCode.compatibility`. It now rides `OcxTool.allowedCallers` from the Messages inbound,
through the Responses tool schema — where an undeclared key is stripped, which is why it never
reached `buildTools` — to the Anthropic wire. Gemini's `functionCallingConfig.mode: "VALIDATED"`
was plumbed to the wire compiler but only reachable by matching a model name; a caller-declared
strict tool now selects it in place of the absent-choice default, and `NONE`, `ANY` and a
forced-name choice are never overwritten.

**#5212 — inline document bytes.** Both inbound parsers reduced an attachment to its name before
any adapter ran, so no adapter could forward one even to a target with a representation for it.
`OcxContentPart` gains a document member carrying the media type and the base64 payload; Anthropic
emits the document block, OpenAI Chat the file part, Gemini `inline_data`. Widening that union is
the hazard, so the part also carries the marker every text-only consumer already falls back to,
which keeps a wire with no document representation byte-identical to before. Six consumers needed
more than the fallback: `ollama-native` and the Cursor tool-result decoder would have read a
nonexistent `imageUrl`, and the Kiro, Devin, Cursor and coding-agent text serializers would have
produced an empty turn.

**#5213 — developer message position, then role.** Two commits, because they are two acceptance
conditions. A developer message kept its slot only when the base URL host was exactly
`api.openai.com`; everywhere else its text was appended to the system prompt and the message
skipped, so an instruction written to apply from the second turn onward arrived ahead of the
first. #4161 had already made the Claude inbound mint chronological developer items for exactly
this reason, so the two halves of a Claude Code route were working against each other. Placement
is now uniform. The role is separate: `developer` is part of the Chat Completions role set and is
forwarded as sent, and a destination that genuinely rejects it sets `foldDeveloperRoleToSystem`,
which converts the role in place and never moves the message.

**Refusals are default-deny.** `allowed_callers` and inline document bytes are constraints the
normalized request can carry but a wire may not be able to express. Refusing them per adapter
would leave every adapter that never learned about the carrier rebuilding without it and
answering normally, which is the defect this batch exists to remove. Both are allowlists in
`src/adapters/declaration-carrier.ts`, enforced at the single guard every registered adapter
passes through: `allowed_callers` reaches the `anthropic` wire, document bytes reach
`anthropic`, `openai-chat` and `google`, and the Responses passthrough stays exempt because it
forwards the original body. Adding an `AdapterWire` member makes the omission visible in those
lists rather than at a customer's upstream. The refusal names no tool, because the name is
caller-controlled and would put client metadata into an error body.

Native passthrough is unaffected on every route.

Carries #5237 by @Yum-wu for the #5213 position fix, with a `Co-authored-by` trailer on the
branch commit.

### Known gap

A document in a **tool result** keeps the #939 marker. The Responses tool-output vocabulary has no
file block and every adapter's tool-result path flattens to text, so carrying bytes there is a
separate change rather than a half-done one. The Chat inbound also still folds a `developer`
message into the system prompt on its own ingress; that is the inbound half of the same class and
is tracked separately as #4148.

## Verification

- Static source review of every changed file, plus repeated adversarial source review of each
  commit and of the whole branch. That review is what found the native-passthrough half of
  #5211, the six content-part consumers that would have dropped or mislabelled a document, the
  role-aware narrowing of the untranslated-media refusal, a `;notbase64,` data URL the scanner
  accepted and the decoder rejected, the missing `toolRestrictsCallers` barrel export, and the
  developer-document role demotion.
- Union-defect classes checked by hand before pushing, per `AGENTS.md`:
  `src/adapters/openai-chat.ts` was the only capped file in the touch set and is **811 lines
  against its 822-line cap** — the `parallel_tool_calls` decision moved to a sibling module
  rather than raising a number; `PROVIDER_CONFIG_FIELD_POLICY` is
  `satisfies Record<keyof OcxProviderConfig, ...>` and learned the new key; every new test file
  is registered in both `scripts/test-layout/layout.json` and
  `tests/fixtures/test-layout-expected.json`; the provider reference tables gained a row rather
  than a restated count; and the attachment marker is derived from `inlineDocumentMarker` in
  every assertion instead of being written out again.
- Rebased onto `dev` `8e1fdea1c0` and pushed at that base, so hosted CI runs at the exact head.
- Docs updated in all eight locales: `guides/claude-code.md` for the placement and role contract,
  `reference/configuration/providers.md` for `foldDeveloperRoleToSystem`, and `guides/pi.md` for
  the inline-document exception to the attachment refusal. Owning `structure/` documents updated
  for the adapter, registry and inbound-compat contracts.
- **NOT RUN:** `bun run test`, individual `bun test` files, `bun run typecheck`, `bun run build`,
  `bun install`, and live `ocx` execution. This lane verifies by static review plus exact-head
  hosted CI only; the CI run on this head is the test evidence.

Closes #5211
Closes #5210
Closes #5212
Closes #5213

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Preserved inline document bytes across supported Anthropic, OpenAI Chat, and Gemini translations.
  * Added support for `allowed_tools`, tool caller restrictions, strict tool declarations, and validated Gemini tool mode.
  * Preserved developer-message ordering across Chat routes, with optional `foldDeveloperRoleToSystem` configuration.
  * Improved handling of parallel tool-call preferences.

* **Bug Fixes**
  * Added clear errors for unsupported media, caller restrictions, empty tool selections, and unnameable tools.
  * Preserved document text across additional adapter and tool-result paths.

* **Documentation**
  * Updated provider configuration, attachment compatibility, and timeline-reminder guidance across supported languages.


lidge-jun:
## 리뷰 · 우선순위 61 / 80

이 PR은 “요청에 분명히 적힌 제약인데, 프록시가 그걸 버리고도 그냥 200으로 성공한 것처럼 보이는” 버그 네 개를 한꺼번에 고칩니다. base는 `dev`이고, 이슈 #5211(도구 선택/병렬 호출), #5210(도구 선언의 `strict`·`allowed_callers`), #5212(첨부 문서 바이트), #5213(developer 메시지 위치·역할)을 닫습니다. 핵심 방향은 맞습니다. 제약을 조용히 빼지 말고, 못 실는 와이어에서는 명시적으로 거절하고(`declaration-carrier.ts`의 default-deny allowlist), 테스트는 “도구 호출이 됐다/200이다”가 아니라 **밖으로 나간 요청 본문**을 읽게 잡았습니다. `parallel_tool_calls` 결정을 한 모듈로 모은 것, 문서 파트에 텍스트 마커를 같이 실어 옛 소비자 동작을 유지한 것, 알려진 구멍(툴 결과 안의 문서, Chat inbound의 developer 접기 #4148)을 PR 본문에 적어 둔 것도 좋습니다. #5237(Yum-wu)의 위치 수정을 이 브랜치가 흡수했다고 명시돼 있습니다. 로컬 `bun test`/`typecheck`는 안 돌렸고 hosted CI에 맡긴 상태라, 지금 시점에도 test shard가 아직 pending입니다.

라인 - `src/adapters/openai-chat/messages.ts` `foldDeveloperRoleToSystem` — 예전엔 `api.openai.com`이 아닌 Chat 호스트에서 developer를 system으로 접었는데, 지금은 기본이 `developer` 그대로입니다. 옵션 기본값이 `false`이고 내장 프리셋에 켜 둔 곳도 없어 보입니다. developer를 거절하는 OpenAI 호환 게이트웨이에서는 업그레이드 후 upstream 400이 날 수 있습니다. Codex 리뷰 P1과 같은 지점입니다.
라인 - `src/adapters/openai-chat/messages.ts` developer+image — 이미지가 있으면 여전히 `role: "user"`로 내려갑니다. 위치는 지키지만 지시 우선순위(역할)는 다시 바뀝니다. 문서만 있을 때는 role을 지키는 것과 기준이 어긋납니다.
라인 - `src/chat/inbound.ts` `allowedToolsChoiceToResponses` — `mode`가 `"required"`가 아니면 전부 `"auto"`로 떨어집니다. `"require"` 같은 오타도 조용히 선택이 아니라 선택 가능으로 바뀌고 200이 나갑니다. 이 PR이 고치려는 “조용한 완화” 유형과 같습니다.
라인 - `src/chat/inbound.ts` `allowed_tools` + `toolsToResponses` — selector는 `custom`·일부 hosted 종류를 받는데, 선언 쪽 투영은 function/web_search 위주로 보입니다. 선택만 남고 목록이 비면 도구 없는 완료가 정상처럼 나갈 수 있습니다(Codex P1).
라인 - `src/responses/input-media.ts` vs `parser-content.ts` — 스캐너는 Anthropic형 `document.source.base64`를 “바이트 있음”으로 통과시킬 수 있는데, Responses 파서는 `input_file.file_data` 위주입니다. 거절도 안 하고 바이트도 안 실리는 구멍이 생길 수 있습니다.
라인 - PR #5237 `fix/preserve-developer-role-position` — #5213 위치 수정만 다루는 열린 PR입니다. 이 배치가 그 내용을 흡수했으니 중복으로 닫을 후보입니다.

메인테이너의 판단이 필요한 지점

`foldDeveloperRoleToSystem`의 기본값을 “표준 Chat 역할 그대로”로 둘지, “옛 호환(system으로 접기)을 기본으로 두고 지원되는 대상만 앞으로 보내기”로 둘지입니다. 위치 보존(#5213 핵심)과 upstream 400 위험 중 무엇을 기본으로 잡을지 선택해야 합니다. 내장/자주 쓰는 openai-chat 프리셋에 플래그를 미리 박을지도 같이 정하면 좋습니다. 또 `allowed_tools`에서 이 ingress가 실을 수 없는 selector 종류를 거절할지, 선언 투영을 넓힐지도 정책 결정입니다.

너의 추천

머지 후보로 두고, 먼저 (1) developer 역할 기본값/프리셋 호환 정책을 정한 뒤 그에 맞게 고치고, (2) `allowed_tools.mode` 오타·매칭 실패·미투영 selector는 거절로 막고, (3) #5237은 landed-via/#5271로 닫고, (4) hosted test shard가 초록인지 확인한 뒤 합치는 쪽을 권합니다. 툴 결과 문서·Chat inbound developer 접기(#4148)는 이 PR 범위 밖 후속으로 두어도 됩니다.

이 댓글은 grok-bot이 작성했습니다

PR #5267, opened 2026-09-20
fix(codex): stop a stopped proxy from locking users out of Codex sign-in

lidge-jun:
## Summary

A Windows 11 user on 2.59.0 was locked out of **Codex sign-in** after setting up opencodex (#5261). Their screenshots show why: the default loopback integration does not add a provider, it points Codex's own built-in `openai` provider at the proxy by writing the codex-rs root key `openai_base_url = "http://127.0.0.1:10100/v1"` into `~/.codex/config.toml`. That file is on disk and survives a reboot, so once the proxy process was gone Codex had one endpoint, it answered nothing, and there was no fallback. The screen said nothing about opencodex.

Recovery already existed and already worked with the proxy down — `ocx restore` needs no proxy, no management API and no network. Nothing told the user it existed. The injected config named no command, `ocx status` on a dead proxy offered only ways to restart the proxy, and no troubleshooting page covered the state. The reporter's assistant read the config, found three unexplained lines, and hand-deleted them along with the catalog file — which is the one repair that makes things worse, because a `model_catalog_json` pointing at a file that no longer exists stops Codex loading its config at all.

This makes the failure detectable and the recovery discoverable. Excluding sign-in from the proxy path is not expressible: `openai_base_url` is a single key for a single built-in provider, and when the proxy is down no scoping of it helps.

- **The injected routing names its own undo.** Markers become `# Auto-injected by opencodex (undo: ocx restore)`. Every ownership predicate matches the marker as a substring rather than by equality, so configs written by earlier builds are still recognized, stripped and restored unchanged; an in-place rewrite refreshes the line, so an existing install gains the hint on its next `ocx start`. Scope is routing only — prompt layers keep the bare marker, because `ocx restore` is not what undoes those.
- **`ocx status` offers the other half of the choice.** With the proxy down over routing opencodex owns, the report now says sign-in fails too and names the command that does not require the proxy to come back. Routing opencodex does not own is excluded, since `ocx restore` would not remove someone else's local gateway.
- **A troubleshooting page** for the state, with both ways out, the manual edit for someone without the CLI, and an explicit warning against deleting the catalog pointer on its own.

Before / after, same file, proxy stopped:

```diff
-# Auto-injected by opencodex
+# Auto-injected by opencodex (undo: ocx restore)
 openai_base_url = "http://127.0.0.1:10100/v1"
```

### Not closing #5261

This removes the lockout's dead end, not every path into it. Left open deliberately, each being a separate change with its own risk and none of them what traps the user on its own:

[...]

Issue #5261, opened 2026-09-20
用Codex配置完opencodex后无法调用模型，重启后无法登录


Wh1tZz:
### Client or integration

Codex App

### Area

CLI

### Summary

我用codex帮我配置了opencodex，添加号池一直失败，之后我在codex内就无法调用模型了，起初我以为是codex的问题，重启后发现无法正常登录了，后来经过审查分析后，发现opencodex 之前把 Codex 的所有请求都劫持到它自己的本地代理（端口 10100），导致卡死，所有的附图均已上传

<img width="510" height="101" alt="Image" src="https://github.com/user-attachments/assets/a05f2ab8-0797-4a96-8e8a-e00165605628" />

### Reproduction

无法正常配置号池，添加帐号也无法正常显示

### Version

2.59.0

### Operating system

win11

### Logs or error output

```shell

```

### Screenshots and supporting files

<img width="936" height="450" alt="Image" src="https://github.com/user-attachments/assets/a48de1a1-c5c3-43de-8add-519e5829ed88" />

<img width="1003" height="859" alt="Image" src="https://github.com/user-attachments/assets/957bca2d-b914-40a3-b668-61bf0b39018b" />

### Redacted configuration

```json

```

### Checks

- [x] I searched existing issues and documentation.
- [x] I removed secrets, tokens, account details, request credentials, and personal data.

lidge-jun:
Thanks for the detailed report and especially for the screenshots — they show the mechanism directly, so this is confirmed rather than suspected.

### What is happening

On the default loopback setup, opencodex does not give Codex a separate provider. It points Codex's own built-in `openai` provider at the proxy, by writing a root override into `%USERPROFILE%\.codex\config.toml` — the three lines in your second screenshot:

```toml
model_catalog_json = "...\opencodex-catalog.json"
# Auto-injected by opencodex
openai_base_url = "http://127.0.0.1:10100/v1"
# Auto-injected by opencodex
experimental_realtime_ws_base_url = "http://127.0.0.1:10100/v1"
```

Those lines are on disk, so they survive a reboot, and Codex has no second endpoint to fall back to. Once nothing was listening on 10100, every Codex request failed — including the sign-in bootstrap, which is why you got a retry screen instead of a model error. Your reading of it was right.

Two details worth adding, because they explain why it persisted:

[...]

PR #5288, opened 2026-09-20
fix(codex): close the silent paths into a Codex lockout

lidge-jun:
## Summary

Follow-up to #5267, which removed the dead end a user locked out of Codex hits. This takes the four paths INTO that state that #5267 listed as not closed. None of them locks anyone out alone; the reported incident is what happens when every signal is missing at once — a proxy that stopped, an autostart that failed silently, a reboot that restarted nothing, a catalog pointer naming a deleted file, and an account flow that looked like it was working. Each step had an explanation and the user could reach none of them.

**The shim hid autostart failure, and on PowerShell could stop Codex launching at all.** It ran `ocx ensure` with both streams discarded and `|| true`, then launched Codex regardless, so a proxy that failed to come up was invisible. It now checks the exit status and prints one line on stderr naming `ocx doctor` and `ocx restore`. Ensure's own streams stay discarded on purpose: it prints progress and warnings on exit-zero runs too, and a wrapper that leaked those would put noise in front of every ordinary launch, which is how a diagnostic gets ignored.

PowerShell had the opposite defect in the same place. A throwing `ensure` escaped a `try/finally` with no `catch`, so the real launcher was never reached — the autostart helper producing the exact lockout it exists to prevent, with a test asserting that propagation as correct. It now catches, reports, and hands over, with Codex's exit status still authoritative.

**A catalog pointer whose file is gone is now reported.** Correcting my own note on #5267: injection does already repair this — the chooser refuses a missing owned path and the caller strips the stale line, with end-to-end coverage in place. The real gap is narrower and worse. That repair only reaches someone who runs opencodex again, and the whole difficulty of the state is that Codex is what stopped working, so nothing prompts them to. A `model_catalog_json` naming a missing file does not degrade Codex, it stops Codex loading its configuration at all: the same blank wall as dead routing, from a different cause. `ocx status` now names the file and offers both outcomes, regenerate or remove.

**Setup now says that routing outlives the proxy.** Applying the integration does not install a background service, so "routing written, nothing listening" is an ordinary state after a restart rather than a corruption, and setup ended on a success line without mentioning it. It now reuses the existing restart-health model, so this cannot drift from what status and doctor report about the same install.

A Windows BootTrigger is deliberately not the fix here. The scheduled task runs as the interactive user, so before logon there is no session for it to run in; the trigger would read like a fix and change nothing. Genuine pre-logon start means a different principal and a different service backend.

**A browser that never opened no longer looks like success.** The URL launcher swallowed its own failure and returned nothing, so the Codex login route answered identically whether a browser opened, failed, or was skipped; the CLI printed the URL and polled while nothing had happened. Launch failure is now reported and still never fatal — the URL is worth opening by hand and the flow stays live. The recovery line names the fixed callback port 1455, because ChatGPT supplies the redirect URI so the flow cannot move to a free port, which is the part a user cannot work out alone.

### Not closing #5261

Three residuals, each a separate change with its own risk:

- Windows shims carry no revision marker and are excluded from obsolete-shim refresh, so an existing Windows wrapper keeps the old template — including the PowerShell path that could stop Codex launching — until it is reinstalled with `ocx codex-shim uninstall` then `ocx codex-shim install`. Unix shims regenerate on their own: the Unix revision marker moved to 3. Giving Windows a revision marker and a refresh path means changing a transactional install path that this lane cannot exercise, so it is named rather than attempted.
- Generic OAuth provider login and key login still discard the launch result. Both already print the URL before launching, and both are immediately followed by a readline prompt, so reporting there needs the message sequenced against the prompt rather than raced with it. The Codex account-pool path this issue reported is fixed; these are adjacent.
- Genuine pre-logon autostart on Windows needs a different principal and service backend, as above.
- The dashboard account roster keeps its last-good rows after a failed refresh, which is why a newly added account can be absent while older ones still show. That is a user-interface change, and this repository requires a screenshot for one; this lane's verification is static review plus hosted CI with no build, so it cannot produce it. Left for a lane that can.

## Verification

- Shim behaviour is covered by running the generated script against a stand-in `ensure`, not by asserting on template text: the property under test is what the shell does with an exit status. A failing ensure produces exactly one diagnostic and Codex still runs with its own exit status; a succeeding ensure produces nothing on either stream; an ensure whose launcher does not exist at all is treated as a failed start rather than dying with it.
- That file is a sibling of `codex-shim.test.ts` because that file sits close to its line cap and the cap only moves down. Registered in both `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`, as is the new oauth test.
- The PowerShell case that asserted a throwing ensure propagates now asserts the opposite, since that propagation was the lockout. Both phases surface Codex's own error, and the shim diagnostic appears only for the ensure phase.
- Catalog detection is tested against real temp files: an owned catalog that is missing is reported, one that exists is not, and a catalog the user named is left alone whether or not it exists — claiming it would put opencodex's advice in front of a problem that is not ours to explain.
- The setup warning is tested through the health model: it speaks when the install is restart-unsafe, stays silent for native routing and for a viable service, and still speaks when only a CLI-only shim is installed, which is the reported shape.
- Eight existing mocks of the URL launcher returned void, which the awaiting caller would have read as a failed launch. All eight now resolve a result, line-neutral in ratcheted files.
- The new launcher test deliberately does not cover the started case: the command is fixed per platform, so proving it would mean opening a real browser on whatever machine runs the suite. The spawn-failure case is Linux-only for the same reason.
- Test constants that were private literal copies of source constants now come from the source module. That class broke two lanes in this batch.
- File-size ratchet checked against `tests/fixtures/file-size-baseline.json`: no tracked file is over its cap.
- NOT RUN, by lane constraint: local suite, individual tests, typecheck, build, install, `ocx` execution, service start or restart, and any change to credentials or configuration on this machine. The incident being fixed is a configuration change that locked a user out, so verification is static source review plus exact-head hosted CI on this branch.
- No user-interface change, so no screenshot.
- An adversarial review of the branch found a blocker before CI did: the PowerShell shim test's Codex-failure fixture also failed ensure, so that phase no longer isolated a Codex failure and both PowerShell variants would have failed on Windows. Its ensure now succeeds, which is what the phase always meant. The same pass caught the launcher reporting a launch on the spawn event alone, a setup warning that claimed nothing restarts the proxy when a healthy shim does, and a local-state line missing its `(local)` tag while connected to a hub. All are fixed here.
- Hosted CI also failed once on the guard that every test bounding a spawned child with the internal deadline must declare whether it warms that child's module graph. The new shim test is recorded as unwarmed: its children are throwaway shell scripts, so the cold cost is shell startup rather than a repository module graph. That comparison was reproduced statically before pushing the fix.

## Checklist

### Security review

This touches the login path, so stating it explicitly. No credential, token, account identifier or authorization URL is logged, serialized or added to any output by this change. The new login-response field is a three-value launch outcome carrying no user data. The recovery line is fixed text. A failed browser launch stays non-fatal and returns the same status code as before, so nothing becomes a login failure that was not one; making it fatal would turn an observable inconvenience into a lockout, which is the defect class this lane exists to remove. The launcher still never throws and never rejects, preserving the property that a headless host cannot kill a login flow. The shim diagnostic is fixed text on stderr and interpolates nothing; ensure's own output remains discarded, so nothing it prints can reach a terminal through the wrapper.

## Summary by CodeRabbit

- **New Features**
  - Login now reports when the browser could not be opened and provides instructions for opening the URL manually or using device login.
  - Status and setup diagnostics identify missing model catalog files and explain how to regenerate or remove stale references.
  - Startup checks warn when routing may survive a restart without the proxy restarting.
  - The GUI command now warns when no browser can be opened.

- **Bug Fixes**
  - Codex continues launching when autostart checks fail while displaying a clear diagnostic.
  - URL-launch failures are reported without interrupting active login flows.


lidge-jun:
## 리뷰 · 우선순위 59 / 80

이 PR은 #5267 다음에 남은 “조용히 Codex가 막히는 길” 네 개를 같이 닫습니다. 베이스는 `dev`입니다. (1) Codex 자동시작 심이 `ocx ensure` 실패를 삼키던 문제를 고칩니다. 유닉스는 종료 코드를 보고 한 줄만 stderr에 찍고, PowerShell은 예외를 잡아 Codex는 그대로 띄웁니다. 예전에는 PowerShell에서 ensure가 던지면 Codex 자체가 안 떠서, 막아주려던 심이 오히려 잠금을 만들었습니다. (2) `model_catalog_json`이 opencodex 카탈로그를 가리키는데 파일이 없으면 `ocx status`가 그 경로를 말하고 `ocx start` / `ocx restore`를 안내합니다. (3) 셋업이 끝날 때, 라우팅은 재부팅 뒤에도 남는데 프록시는 안 살아날 수 있다는 경고를 기존 restart-health 모델로 찍습니다. Windows BootTrigger는 일부러 안 넣었습니다. (4) 브라우저가 안 열렸는데도 로그인처럼 보이던 침묵을 끊습니다. `openUrl`이 결과를 돌려주고, Codex 로그인 응답에 `browserLaunch`가 실립니다. 실패해도 로그인은 죽지 않고, 콜백 포트 1455와 `--device` 안내만 추가합니다. 유닉스 심 revision은 3으로 올려 재생성되고, Windows 심은 재설치 전까지 옛 템플릿이 남는 잔여는 본문에 명시돼 있습니다. 테스트·layout·cold-spawn 기록도 같이 들어왔고, tip은 adversarial 리뷰에서 나온 PowerShell fixture·런처 settle·경고 문구·`(local)` 태그까지 고친 상태입니다. 리뷰 시점 CI는 gates/hygiene/windows-schtasks 등은 통과, test/macos 일부는 pending입니다.

라인 - `src/codex/shim-templates.ts` / `src/codex/shim.ts`: 유닉스만 revision 3으로 갱신됩니다. Windows 심은 revision·obsolete refresh가 없어서, 이미 깔린 PowerShell 래퍼는 `ocx codex-shim uninstall` 후 `install` 전까지 “ensure 예외 → Codex 미실행” 옛 동작이 남습니다. 이 레인이 고친 잠금 중 가장 센 구멍이 Windows에 그대로 남을 수 있습니다.

라인 - `src/cli/status.ts` `missingCodexCatalogLines`: 안내가 `ocx start`만 말합니다. 프록시가 이미 떠 있으면 `ocx start`는 live-owner 때문에 거절하고 카탈로그 동기까지 못 갑니다. 그 장면에서는 `ocx sync`(또는 동등한 재생성 경로)가 맞고, 지금 문장만 따르면 파일이 안 생깁니다.

라인 - `src/cli/init.ts` setup 끝 경고: `collectStartupHealth`가 `at-risk`이면 “이번에 라우팅을 썼다”처럼 말합니다. 사용자가 이번 셋업에서 Codex inject를 `n`으로 거절했어도, 예전에 남아 있던 at-risk 설정이면 같은 경고가 나옵니다. 이번 셋업이 쓴 라우팅이 아닐 수 있습니다.

라인 - `src/lib/open-url.ts`: 400ms settle 뒤 `started`로 굳히면, 그 다음에야 nonzero로 죽는 런처는 성공으로 남습니다. `exit`의 `code === null`(시그널 종료)도 `started`로 칩니다. 또 Codex 로그인 start 경로는 `await openUrl`이라 응답이 최대 약 0.4초 늦어집니다.

라인 - `src/oauth/login-cli.ts` / `oauth-account-routes.ts`: 제네릭 OAuth·키 로그인은 여전히 `void openUrl`이라, 브라우저 실패가 다시 침묵입니다. 본문이 남긴 잔여이고, Codex account-pool 경로만 고친 범위와 맞습니다.

라인 - 카탈로그 안내는 `ocx status` 사람 출력에만 붙어 있습니다. `ocx doctor`와 status JSON에는 같은 신호가 없습니다. 심 문구는 doctor를 가리키는데, 카탈로그 구멍은 doctor를 봐도 안 보일 수 있습니다.

메인테이너의 판단이 필요한 지점

Windows 심 revision/refresh를 이 PR에 넣을지, 아니면 본문대로 후속 레인으로 둘지. 카탈로그 안내에 `ocx sync`를 지금 tip에 넣을지. setup 경고를 “이번 inject가 실제로 썼을 때만”으로 좁힐지. 제네릭 OAuth 침묵·doctor/JSON 노출·GUI last-good roster는 #5261 잔여로 OPEN 유지할지. structure 문서 갱신을 이 레인 블로커로 볼지.

너의 추천

방향이 맞고, tip의 adversarial 수정까지 반영된 상태라 `dev` 머지 후보로 보입니다. 머지 전에 값싼 두 가지는 tip에 넣는 편이 좋습니다. (1) 카탈로그 안내에 프록시가 이미 떠 있을 때 쓸 `ocx sync`를 명시하기. (2) setup 경고를 “이번 셋업이 inject에 성공했을 때”로 가드하기. Windows 심 재설치/revision은 별 PR이 맞고, 머지 노트나 #5261에 “기존 Windows 심은 재설치 필요”를 한 줄 박아 두면 사고 재발을 줄입니다. 제네릭 OAuth·doctor 노출·GUI roster는 후속으로 두고 #5261은 OPEN 유지. test/macos pending만 초록 확인 후 머지하면 됩니다.

이 댓글은 grok-bot이 작성했습니다

PR #5327, opened 2026-09-20
fix(desktop,cli): make the build report its own state

lidge-jun:
## Summary

Three defects found by building and running the desktop app locally for the first time. Each one produced a state that its own output did not report.

**The release profile could not compile the app.** `cargo build --release` stopped at `ctor` with `can't find crate for ctor_proc_macro`, while the dev profile compiled the identical graph. `[profile.release] strip = "symbols"` is applied by cargo to build scripts and proc macros as well as to the crate under build, and a proc macro is a host dylib rustc loads by symbol, so stripping it leaves a file rustc cannot read. The error names the macro and never mentions the profile that removed its symbols. `[profile.release.build-override] strip = false` restores it.

**A stale dashboard bundle was invisible.** The dashboard is served from `gui/dist`, so a checkout that moves forward without `bun run build:gui` keeps serving the previous bundle. Nothing fails — the proxy answers, the page loads, and every feature added since the last build is simply absent, which reads as the feature being broken rather than unbuilt. A five-day-old bundle hid the entire menu-bar and widget section of the Usage page that way. `ocx status` now compares the newest mtime under `gui/src` with the served bundle and names the rebuild beside the dashboard URL. It reports and never rebuilds: a proxy compiling a frontend at startup would trade silent staleness for a slow, surprising start.

**A local build ended on a failure after succeeding.** `tauri build` always writes the updater archive and then refuses to finish without `TAURI_SIGNING_PRIVATE_KEY`, with both bundles already on disk. That exit code is right for a release and misleading on a workstation, and a wrapper cannot tell it apart from a real failure. `bun run build:local` turns the artifact off for that invocation rather than leaving the key required and unmet, so nothing is skipped unsigned.

## Verification

- `cargo build --release -p ctor` fails before the profile change and passes after, on rustc 1.95.0.
- `bun test tests/server/gui-bundle-freshness.test.ts` — 4 pass, covering stale, fresh, unknown-either-side, and a `node_modules` tree that must not make sources look newer than they are.
- Live check: touching `gui/src/main.tsx` made `ocx status` print the rebuild line, and `bun run build:gui` removed it.
- `bun run build:local` produced `OpenCodex.app` and `OpenCodex_2.61.0_aarch64.dmg` and exited 0. Selecting bundle targets alone was not enough — `createUpdaterArtifacts` is a config flag, so `--bundles app,dmg` still produced the updater archive and still failed; the override has to reach the config.
- The committed `tauri.conf.json` keeps `createUpdaterArtifacts: true`, so the release path still refuses to publish an unsigned updater artifact.
- The new test file is registered in both `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added a local desktop build command that creates the app and DMG without signing keys or updater artifacts.
  - Added status diagnostics that alert you when the dashboard bundle is out of date and provide the rebuild command.

- **Bug Fixes**
  - Improved release build reliability by preventing build tooling from failing because of symbol stripping.

- **Documentation**
  - Documented local unsigned build behavior and desktop build stabilization guidance.


Ingwannu (changes requested):
Requesting changes on exact head `7eee5f788aee8ee4d19170cb018025d8f06f1114`.

Two deterministic blockers are visible before the remaining jobs finish:

1. `tests/test-layout-tooling.test.ts` rejects the new registration as `gui-bundle-freshness.test.ts: seed gui != server`. The test is currently registered as `server` in both layout files. Move it to the owner/seed the layout tool derives (or rename/scope it consistently) and refresh the expected fixture. The separate unresolved StepFun test is from the moving base, but this GUI seed mismatch belongs to this PR.
2. `newestModifiedMs` claims a 20,000-entry bound but increments `seen` only for regular files. A tree containing arbitrarily many directories (including empty ones) remains unbounded on the synchronous `ocx status` path. Count every visited directory entry against the budget. On budget exhaustion, return an explicit unknown result rather than a partial timestamp: a partial bundle scan can falsely report stale, while a partial source scan can falsely report fresh, so the current “conservative” claim does not hold for both sides.

The exact-head `macos widget + bundle` job is also red and must be diagnosed once its log is available. Please fix these boundaries and rerun the full replacement-head CI before requesting approval.

Ingwannu:
The completed `dev` CI log identifies the macOS bundle failure: `bunx tauri build --ci --bundles app` successfully builds and bundles `OpenCodex.app`, then still creates `OpenCodex.app.tar.gz (updater)` and exits because `TAURI_SIGNING_PRIVATE_KEY` is absent. That is exactly the condition `build:local` was added to avoid, but this PR does not change the `macos widget + bundle` workflow, so its exact-head check will keep running the old command and stay red.

Please wire the unsigned CI bundle job to the same `createUpdaterArtifacts: false` override (either through the reviewed local-build entrypoint or an equivalent explicit workflow config) while leaving release signing unchanged. Because this touches `.github/workflows/ci.yml`, it needs the repository security/workflow review required by `MAINTAINERS.md`. This is in addition to the layout and bounded-tree issues in my change request.

lidge-jun:
## 리뷰 · 우선순위 56 / 80

이 PR은 데스크톱 앱을 처음 로컬에서 빌드·실행할 때 드러난 세 가지 거짓말을 고칩니다. 첫째, 릴리즈 프로필의 `strip = "symbols"`가 빌드 스크립트·프로시저 매크로까지 벗겨서 `ctor_proc_macro`를 못 찾게 만듭니다. 그래서 `build-override`로 심볼을 남깁니다. 둘째, 대시보드는 `gui/dist` 산출물인데 소스가 새로워도 예전 번들을 그대로 줍니다. `ocx status`가 소스와 번들 시각을 비교해 재빌드 안내를 찍습니다. 셋째, `tauri build`는 업데이터 아카이브까지 만든 뒤 서명 키 없이는 실패합니다. `bun run build:local`은 그 한 번만 업데이터 산출을 끄고 앱·dmg만 만듭니다. base는 `dev`이고 방향은 맞습니다. 다만 지금 head에서는 CI가 이미 빨간데, 그 실패 중 두 개는 이 PR이 직접 넣은 파일에서 납니다.

`scripts/test-layout/layout.json`, `tests/fixtures/test-layout-expected.json` - 새 테스트 `gui-bundle-freshness.test.ts`를 `server`로 올렸습니다. 레이아웃 시드는 `^gui-` 이름을 `gui`로 보므로 `seed gui != server`로 깨집니다. 파일 이름을 서버 쪽으로 바꾸거나, 시드/명시 등록을 같은 도메인으로 맞추세요. 테스트 본문은 `src/server`를 보니 `server`가 맞고, 이름만 걸립니다.

`desktop/bun.lock` - 새로 들어온 락이 `lockfileVersion: 2`입니다. `macos widget + bundle` 작업이 `bun install --frozen-lockfile`에서 `Unknown lockfile version`으로 바로 죽습니다. CI가 쓰는 bun이 이 버전을 못 읽습니다. 루트와 같은 bun으로 다시 만들거나, 그 락이 정말 필요한지부터 확인하세요. 데스크톱 패키지에는 `@tauri-apps/cli`만 있고, 예전엔 락 파일이 없었습니다.

`src/server/gui-freshness.ts` `newestModifiedMs` - 주석은 2만 개 한도를 말하지만 `seen`은 일반 파일만 셉니다. 빈 디렉터리가 아주 많으면 `ocx status`가 끝까지 동기 순회합니다. 게다가 한도에 걸리면 지금까지 본 시각을 그대로 씁니다. 번들 쪽만 잘리면 거짓 오래된 경고가 나고, 소스 쪽만 잘리면 경고가 사라질 수 있습니다. 방문한 항목마다 세고, 한도에 걸리면 “모름”으로 두세요. 모름은 stale이 아닙니다.

`desktop/scripts/build-local.ts` - 번들 목록이 `app,dmg`로 고정입니다. 리눅스·윈도우 워크스테이션에서는 로컬 경로가 아닙니다. 또 tauri가 0을 내고도 `OpenCodex.app`이 없으면 성공 메시지만 찍고 0으로 끝납니다. 기대한 산출물이 없으면 실패로 끝내세요.

`src/cli/index.ts` / `gui-freshness` - 비교 대상이 `gui/src`뿐입니다. `gui/public`, `gui/index.html`, `gui/vite.config.ts`만 바뀌어도 경고가 안 납니다. 이번 범위에서는 알아 두고, 다음에 입력 범위를 넓히거나 빌드 fingerprint를 남기면 됩니다.

메인테이너의 판단이 필요한 지점

`desktop/bun.lock`을 저장소에 둘지 정해야 합니다. 넣으면 CI bun 버전과 맞춰야 하고, 빼면 예전처럼 `bunx`/`package.json`만으로도 됩니다. `build:local`을 macOS 전용으로 문서에 못 박을지, 플랫폼별 번들 목록으로 갈지도 정해 주세요. types/config 분할이나 미리보기 배포와는 무관합니다. 같은 결함의 열린 중복 PR은 보이지 않습니다.

너의 추천

레이아웃 시드 충돌과 `bun.lock` CI 깨짐을 먼저 고치세요. `newestModifiedMs`는 항목마다 세고, 한도 초과 시 null을 돌려 “모름”으로 두세요. `build:local`은 기대한 앱·dmg가 없으면 실패하고, README에 macOS 전용임을 밝히세요. 그다음 head를 갈아끼운 뒤 전체 CI가 초록인지 확인한 다음에 승인 요청하세요. 신선도 범위 확대와 `ocx status` 통합 테스트는 막아 둔 뒤에 이어서 해도 됩니다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu:
@lidge-jun I reproduced the shared CI blocker on #5330: the unsigned macOS job successfully builds `OpenCodex.app`, then Tauri creates the updater archive and fails because the job intentionally has no `TAURI_SIGNING_PRIVATE_KEY`.

I prepared and locally validated the minimal workflow-only fix, but GitHub rejected my push because the current Ingwannu OAuth token lacks the `workflow` scope. Please apply this exact change to the `Build unsigned desktop app` step:

```yaml
# This validation job intentionally has no release signing key. Disable updater
# artifacts for this invocation only; the committed release config still creates
# and signs them, while the unsigned .app remains fully built and inspected below.
run: bunx tauri build --ci --bundles app --config '{"bundle":{"createUpdaterArtifacts":false}}'
```

I also added a focused `tests/ci-workflows/ci-workflows.test.ts` regression that parses the widget step, pins that exact command, and separately asserts `desktop/src-tauri/tauri.conf.json` still has `bundle.createUpdaterArtifacts === true`, so release signing cannot inherit this exception. With temporary `HOME`, `CODEX_HOME`, and `OPENCODEX_HOME`, that file passed **140/140**. No triggers, permissions, secrets, action pins, or release workflow are changed.

This should be included in the replacement head before approval; it is the narrow fix for the CI failure and requires the normal workflow/security review.

Issue #5321, opened 2026-09-20
[Bug]: 2.60.0 hard-refuses the authless provider-table transition on a home with paginated openai threads and offers no supported way forward


RobinBially:
### Client or integration

Codex App

### Area

Service lifecycle

### Summary

After upgrading to 2.60.0, enabling the Codex integration in its authless provider-table form is hard-refused on a Codex home with paginated history. `ocx sync` (and the dashboard sync) abort with `Codex config injection refused: history_paginated_openai_requires_native_writer. ...`, nothing is written, and the integration stays disabled. Before 2.60.0 the same transition completed: the history relabel stood down (existing `opencodex`-tagged threads keep their provider) and the routing/catalog half was still written. Expected: a supported way to complete the transition on such a home, or a named remediation command in the refusal message instead of "do not retry".

### Reproduction

1. Use a Codex home whose history Codex CLI 0.155.1 has migrated to paginated history: the `threads` table in `~/.codex/state_5.sqlite` has the `history_mode` column and the rows read `paginated`, including at least one row with `model_provider = 'openai'`.
2. Set `codexDesktopAuthless: true` in `~/.opencodex/config.json` on a loopback hostname, so the injection uses the provider-table form (`providerTableMode = true`).
3. On opencodex 2.60.0, run `ocx sync` or trigger the dashboard sync.
4. Observe the refusal; no artifacts are written and the integration stays disabled. `ocx restore` also refuses on paginated history (related: #4812).

The only ways out found on this home: delete the affected sessions (`codex delete --force <session-id>`, after backing up rollouts and the state DB), which makes the transition complete normally, or downgrade to 2.59.0. Both are only discoverable by reading the preflight code; naming one in the refusal message or the codex-integration docs would help.

### Version

2.60.0

### Operating system

macOS 27.0 (build 26A428), arm64

### Logs or error output

```shell
Codex config injection refused: history_paginated_openai_requires_native_writer.
Existing provider definitions and conversation files were preserved.
Paginated history requires native-writer coordination; do not run legacy recovery or retry this transition blindly.
```

### Redacted configuration

```json
{ "codexDesktopAuthless": true, "port": 10100 }
```

### Checks

- [x] I searched existing issues and documentation.
- [x] I removed secrets, tokens, account details, request credentials, and personal data.

RobinBially:
Update from the reporter: there is a working path on such a home, in case it is useful for the docs or the refusal message.

[...]

Issue #4812, opened 2026-09-16
ocx restore / stop / uninstall refuse to restore config.toml on paginated Codex history, leaving the CLI pointed at a dead proxy port


practical-tools-lab:
### Client or integration

Codex CLI

### Area

Service lifecycle

### Summary

When any Codex thread in `~/.codex/state_*.sqlite` is tagged `model_provider='opencodex'` on a paginated-history store — the schema every current Codex build uses — `ocx restore`, `ocx stop`, and `ocx uninstall` all refuse to touch `~/.codex/config.toml`. The proxy can then be stopped and uninstalled while `config.toml` still contains `model_provider = "opencodex"` + `[model_providers.opencodex] base_url = "http://127.0.0.1:10100/v1"` (or the root `openai_base_url` override). Every subsequent `codex` invocation fails with a connection error on the dead port until the user hand-edits TOML.

There is no `--force`, no fallback strip, and no escape hatch in the CLI. The refusal message ("Native writer coordination is required") does not tell the user how to recover.

**Expected instead:**

One of:

- a documented `--force` / "strip routing anyway" mode for restore/uninstall, accepting that `opencodex`-tagged threads will not resume;
- or, on refusal, a degraded-but-working end state: e.g. remove the root `model_provider`/`openai_base_url` while keeping the `[model_providers.opencodex]` table on disk (so tagged threads still resolve by id and native requests fall back to the built-in `openai` provider);
- or at minimum, an actionable error telling the user the exact `config.toml` lines to delete.

Notably, the *apply* path already implements the middle option: when the relabel unit stands down, injection deliberately keeps an existing `opencodex` table so tagged threads keep resolving (`src/codex/inject.ts:497-502`). The remove path has no equivalent.

### Reproduction

1. Loopback install, `codexDesktopAuthless: true`, run `ocx sync` → `config.toml` gets `model_provider = "opencodex"` + the provider table.
2. Start one Codex Desktop/CLI conversation → the thread row is written with `model_provider='opencodex'` and `history_mode='paginated'` (current Codex schema).
3. `ocx uninstall` (or `ocx stop`, or `ocx restore`).
4. Observe: the Codex-config step reports `Codex configuration preserved: history_paginated_requires_native_writer`; uninstall finishes with a failed step; `config.toml` still routes to `127.0.0.1:10100`.
5. `codex` / the desktop app → connection refused until manual TOML edit.

### Version

2.56.0

### Operating system

macOS 26 (arm64)

[...]

PR #5331, opened 2026-09-20
fix(codex): complete the provider-table transition on a paginated openai home

lidge-jun:
## Summary

Enabling the Codex integration in its authless provider-table form was hard-refused on any home whose history Codex had already migrated to paginated form, if any conversation was still tagged `openai`. `ocx sync` and the dashboard sync aborted with `history_paginated_openai_requires_native_writer`, nothing was written — no config, no reference profile, no `model_catalog_json` — and the integration stayed disabled. The message ended with "do not retry" and named no way forward; the reporter's home had 173 affected conversations, and the only exits discoverable by reading the preflight source were deleting them or downgrading to 2.59.0. Before 2.60.0 the same transition completed, with the history relabel standing down while the routing and catalog half was still written.

The danger the guard names is real. A provider-table transition removes the root `openai_base_url`, and a paginated row cannot be relabeled, so after the transition that conversation would resume against Codex's built-in OpenAI endpoint rather than this proxy. Refusing to relabel is correct; refusing the whole transition is what trapped the user.

The injector already builds the state that resolves this. `keepRootOverrideAlongsideTable` keeps the marker-owned root `openai_base_url` beside the provider table for the client-compaction form, for exactly this reason: Codex builds its provider map as `merge_configured_model_providers(built_in_model_providers(openai_base_url), model_providers)`, so the override lands on the built-in `openai` entry regardless of which id is the default. `src/codex/inject/paginated-openai-compat.ts` now selects that same retention from the preflight verdict rather than from the routing form. When the reason is the paginated-openai code and the target can own a root key, the override is retained, the reason is downgraded to the stand-down constant so the relabel unit never starts, and the transition completes. The paginated row is neither read nor written, and it keeps resolving to this proxy. The retained value is journaled as OpenCodex's own, so restore can still take it out.

Before, on a loopback authless home with one paginated `openai` thread:

```
Codex config injection refused: history_paginated_openai_requires_native_writer.
Existing provider definitions and conversation files were preserved.
Paginated history requires native-writer coordination; do not run legacy recovery or retry this transition blindly.
```

After, the same home gets its config, profile and catalog, with `[model_providers.opencodex]` and the marker-owned root override side by side, and the thread row still reads `model_provider = 'openai'`, `history_mode = 'paginated'`.

Two cases do not reach that state, and both are honest rather than silent:

- An admission-token form cannot use the root key at all, because Codex's built-in `openai` entry carries no `x-opencodex-api-key` header. It keeps the refusal, and the message now names `unauthenticatedLoopbackListener` and `syncResumeHistory` as the two settings that resolve it.
- A root `openai_base_url` the user owns is left alone. The conversation follows the destination they chose, which is the guarantee the injector already makes everywhere else about a line it does not own, and the journal correctly records that line as not ours.

**The recovery side (#4812).** That issue is already closed on `dev`, and the new refusal code cannot reach restore: it is set only under `providerTableMode`, and restore preflights with `providerTableMode = false`, whose row predicate is `model_provider = 'opencodex'`. Two gaps were still open on that side and are fixed here. `ocx restore --remove-codex-provider-table` worked but appeared in no usage or help text, so the escape hatch was reachable only by reading `dispatch.ts`. And the public guide in all eight locales still said restore and Codex config removal refuse on `history_paginated_requires_native_writer` and that a paginated home cannot be uninstalled through the product — neither has been true since the degraded restore landed.

Other hard-refusal reasons on the recovery path still have no named repair command: a missing state database with pending manifest entries, and a backup manifest that is unreadable, foreign, or schema-invalid. Those are a different failure family and are deliberately left open rather than folded in here.

**Where the logic lives.** `src/codex/inject.ts` was at 984 of its 987-line file-size-ratchet cap, so the decision could not be inlined; the injector pays one import and one net line and now sits at exactly 987. The refusal code became an exported constant in `src/codex/history-provider.ts` because the same literal in two files is how the stand-down pair drifted the first time.

## Verification

Static review plus exact-head hosted CI. Per this lane's constraints, the following were **NOT RUN** locally: `bun test`, any individual test file, `bun run typecheck`, any build, any install, live `ocx` execution, service restart, and any credential or configuration change.

Mechanical gates checked by hand before pushing:

- File-size ratchet. `src/codex/inject.ts` is at 987 of its recorded cap of 987. The end-to-end suite `tests/codex-integration/codex-inject-integration.test.ts` is at 1959 of the 2000-line threshold, which is why the resolver's own cases went to a new file rather than growing it. No cap was raised.
- Test layout. `tests/codex-integration/history-paginated-openai-compat.test.ts` is registered in both `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`, which `tests/test-layout-tooling.test.ts` requires to be byte-equal.
- Union exhaustiveness. History refusal reasons are plain strings with no closed union or roster, and `RestoreHistoryDisposition`, `RemoveCodexConfigHistoryDisposition` and `RetainedCodexProviderTable.reason` are untouched. No count or member list is restated anywhere.
- No generated file needs regeneration. No CLI capability was added, so `skills/ocx/references/01_management_surface.md` is unchanged; the registry gains only a `details` entry, which the skill gate does not read.
- `structure/config.md` (554 lines) and `structure/codex-home.md` (354) stay under the 600-line budget, and every backticked path they name exists.

Regression coverage added:

- `tests/codex-integration/history-paginated-openai-compat.test.ts` — the resolver: the retained line is written before the first table where Codex reads root keys, CRLF endings survive, a user-owned line is left untouched and not claimed, the admission-token refusal names two remedies that are asserted to exist as keys in `src/types/config.ts`, every other reason passes through with the text it always had, and a source-oracle check that the refusal code is defined once and read from that definition.
- `tests/codex-integration/codex-inject-integration.test.ts` — the end-to-end case, rewritten from "refuses" to the completed transition, asserting the config carries both the table and the marker-owned override, the rollout bytes and thread row are unchanged, and a following `ocx restore` takes the retained override back out. That last assertion is what keeps this from trading #5321 for a new #4812.
- `tests/cli/cli-restore-back.test.ts` — the removal flag is present in both help surfaces, read out of `dispatch.ts` rather than restated so a rename cannot leave the help silently wrong.

No GUI file is touched, so the screenshot gate does not apply.

## Checklist

Closes #5321

Partially addresses #4812 (already closed): the discoverability and documentation gaps left by its fix. The remaining recovery refusals listed above are not resolved here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Provider-table integration now succeeds for existing paginated OpenAI conversations while preserving their history and routing.
  * Restore, stop, and uninstall no longer fail solely because history is paginated; retained provider configuration keeps affected conversations accessible.
  * Admission-token routing remains refused when required authorization cannot be supported.

* **New Features**
  * Added `ocx restore --remove-codex-provider-table` to remove retained provider configuration.

* **Documentation**
  * Updated integration guidance and CLI help across supported locales.


Ingwannu (changes requested):
Requesting changes on exact head `034ae6058eaf2fdf4da9ec9815d47820872b9632`.

The hosted CI exposes one PR-specific blocker in addition to the unrelated StepFun layout failure. The rewritten end-to-end test fails at `codex-inject-integration.test.ts:812`: its input uses the existing `# >>> opencodex managed openai_base_url >>>` block, but the compatibility path does not normalize that ownership form to `OCX_ROUTING_MARKER_LINE`.

Concretely, `stripInjectedOpenaiBaseUrl` leaves that block in place, then `setRootOpenaiBaseUrlForTarget` sees the root URL without the current marker directly above it and reports `keptUserBaseUrl = true`. `applyPaginatedOpenaiCompat` therefore returns `retainedRootOverride = false`; `keepRootOverrideAlongsideTable` remains false and the journal does not claim the root URL. The transition happens to remain routed because the old URL bytes survive, but the PR’s stated guarantee that the retained override is owned and removed by restore is not established. CI stops before the later restore assertion can test it.

Please make the ownership contract explicit and make the end-to-end case pass: if that block is a supported legacy OpenCodex marker, migrate it to the current marker and journal the exact retained value; if it was only an invalid synthetic fixture, replace it with the real currently supported marker and add/document the intended legacy boundary separately. In either case, keep the inject -> untouched paginated row -> restore-removes-owned-root assertion executable. The `tests/test-layout.test.ts` failure for `providers/stepfun-provider.test.ts` is the current `dev` regression tracked by #5330, but this marker failure belongs to this head.

lidge-jun:
## 리뷰 · 우선순위 61 / 80

이 PR은 Codex를 프로바이더 테이블 방식으로 켤 때, 이미 페이지 나눔(paginated)된 `openai` 대화가 있으면 전부 막히던 문제를 고칩니다. 예전에는 `history_paginated_openai_requires_native_writer`로 설정·카탈로그·프로필까지 하나도 안 쓰고 통합이 꺼진 채로 남았어요. 지금은 루트 `openai_base_url`을 테이블 옆에 남겨 두어서, 대화 줄을 다시 쓰지 않고도 그 대화가 프록시를 계속 보게 합니다. admission-token 형식은 예전처럼 거절하되, 이제 `unauthenticatedLoopbackListener`와 `syncResumeHistory`를 이름으로 알려 줍니다. 사용자가 직접 쓴 `openai_base_url`은 건드리지 않습니다. `ocx restore --remove-codex-provider-table`은 도움말·레지스트리에 드러나고, 8개 언어 가이드와 structure 문서도 복원 이야기를 맞춥니다. 베이스는 `dev`입니다.

라인 - `src/codex/inject/paginated-openai-compat.ts` → `src/codex/inject.ts` — 사용자가 가진 `openai_base_url`이면 `retainedRootOverride`만 false로 두고 `keptUserBaseUrl`은 inject 쪽에 안 넘깁니다. authless 테이블 경로에서는 원래 `keptUserBaseUrl`이 false라서, 성공 문구가 “모든 모델이 opencodex 프록시로 간다” 쪽으로 갑니다. 실제로는 기존 `openai` 스레드가 사용자 URL을 따릅니다. 저널은 null이라 디스크 소유권은 맞지만, 화면 메시지·경고 분기는 틀립니다.
라인 - `src/codex/inject.ts` `observeHistoryRefusalOrThrow` — 처음 preflight가 null이었다가, 쓰는 도중에 Codex가 `openai` 줄을 paginated로 바꾸면 `HISTORY_PAGINATED_OPENAI_NEEDS_ROOT_OVERRIDE`가 나와요. 이 함수는 stand-down만 통과시키고 이 코드는 throw합니다. 그러면 전환이 되돌아가고 #5321 잠금이 다시 납니다. 일반 paginated stand-down은 중간에 다시 봐도 되는데, 이번 전용 코드는 경계에서 다시 `applyPaginatedOpenaiCompat`을 안 탑니다.
라인 - `src/codex/inject.ts` — 파일 크기 상한 987줄에 딱 맞춰 두었습니다. 다음 한 줄도 못 넣으니, 위 전파/재적용을 고칠 때 또 모듈로 빼야 합니다. 설계상 이해는 되지만 유지비가 큽니다.
라인 - `scripts/test-layout/layout.json` / StepFun — #5313 테스트 등록은 이 버그 수정과 무관합니다. CI 가드를 풀려는 짧은 우회로 보이며, 따로 커밋되어 있어 괜찮지만 리뷰 범위가 살짝 넓어집니다.

메인테이너의 판단이 필요한 지점

사용자 소유 URL + paginated `openai` 홈에서, 성공 메시지를 client-compaction과 같은 “당신 URL을 그대로 둠” 문구로 맞출지, 지금처럼 일반 성공 문구를 쓸지.
쓰기 중 마이그레이션 레이스까지 이 PR에서 고칠지, 드문 경우로 follow-up에 둘지.
남은 복원 거절(깨진 매니페스트·없는 state DB)을 언제 이름 있는 수리 명령으로 열지 — 본문에서 의도적으로 남겨 둔 부분입니다.

너의 추천

메인 경로(#5321 리포터 형태: loopback authless + 관리 마커 URL)는 테스트로 잘 고정되어 있어서 방향은 맞습니다. 합치기 전에 `keptUserBaseUrl`(또는 같은 뜻의 플래그)을 compat 결과에서 inject로 넘겨 메시지·경고를 맞추세요. 가능하면 write 경계에서 `HISTORY_PAGINATED_OPENAI_NEEDS_ROOT_OVERRIDE`가 새로 보이면 compat를 한 번 더 적용하고 후보를 다시 만든 뒤 쓰세요. admission-token 거절은 지금처럼 유지하세요. CI가 초록이면 그 두 가지 반영 후 `dev`에 넣는 걸 추천합니다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu:
I re-checked the new CodeRabbit write-boundary finding against the actual transaction flow. I do **not** consider it a merge blocker for this fix.

If the first preflight returns `null` and history becomes paginated before either write-boundary observation, `observeHistoryRefusalOrThrow` sees `history_paginated_openai_requires_native_writer` and throws. `applyNativeArtifacts` catches that and restores the config/profile preimages; no misrouted committed state is accepted. A retry then sees the specialized reason at the initial preflight and takes the new compatibility path. If migration occurs between the pre-write observation and the writes, the post-write observation throws and the same compensation runs. Rebuilding the candidate and witness inside that boundary would be a larger coordination change, not required to keep this PR safe.

The replacement fixture now uses the real `OCX_ROUTING_MARKER_LINE`, which resolves my earlier exact-head CI failure and makes the inject -> retained owned root -> restore assertion meaningful. I will clear my change request only after the replacement-head tests pass and the shared macOS CI repair (#5338) lands/re-runs this head.

PR #5300, opened 2026-09-20
Make the logs page, the ledger and the exporter agree

lidge-jun:
## Summary

Bundle 14's completion condition is that the UI, the durable log and Prometheus agree on logical and physical counts and on terminal classification. They did not, and the disagreement was concrete.

Three surfaces classified a terminal three different ways. The durable row carried `terminalStatus` and `closeReason`; the exporter kept a private `classifyResult`; the dashboard read the numeric HTTP status and nothing else. A turn cut short by `max_output_tokens` is durably `status: 200, terminalStatus: "incomplete"`, which the exporter reported as `incomplete` and the dashboard rendered as a green 200. For the same request the metric said incident and the operator saw success.

The dashboard also showed no send count at all, so an attempt that sent three times appeared as one row with nothing to say otherwise. And its recovery-kind union had drifted to nine of the durable thirteen, so `key-401`, `oauth-account-429`, `opaque-blob-rejection` and `reasoning-effort-downgrade` each reached the operator as "Unknown recovery reason" — four real causes rendered as the absence of one.

One classifier now lives in `src/usage/request-outcome.ts`. The exporter imports it, including its result label set, so those four strings are stated once. The dashboard calls it too. The data was never missing: `requestLogDto` already spreads the whole durable entry, so the page only had to declare the fields and stop reinventing the precedence.

Two things adversarial review caught before this was pushed, both worth naming because each looked correct:

- A **type-only** import still pulls the imported file's whole import graph into the dashboard's TypeScript project, which sets `erasableSyntaxOnly`. Importing the recovery roster from `src/usage/log.ts` dragged `node:fs`, `node:crypto` and the config barrel into the browser build, where a parameter property in `src/config/atomic-write.ts` does not compile. The browser-visible names now live in `src/usage/telemetry-contract.ts`, which has no imports and must keep none; `src/usage/log.ts` re-exports them so every existing importer keeps its path.
- An earlier draft reported `max(sends, reserved)` as the send total, on the reasoning that a budget charge with no attempt row behind it is still a send that left. True, and it made the dashboard say four where the exporter, summing the same attempts the recorder summed, said three. Two defensible formulas are still two answers. Every surface now reads the recorded totals and recomputes nothing.

Dispositions for the rest of the bundle are in `devlog/_plan/260920_meaning_preservation_batch/030_lane_c2.md`, each with the reason that is true against current `dev` rather than the one written earlier.

[...]

Issue #5292, opened 2026-09-20
Logs page recovery-kind union is restated, so four durable kinds have no label


lidge-jun:
### Client or integration

Dashboard

### Area

GUI

### Summary

`gui/src/pages/Logs.tsx` declares its own `AttemptRecoveryKind` union with nine members, while the durable recovery roster has thirteen. Four kinds therefore reach the Logs page with no localized label: `key-401`, `oauth-account-429`, `opaque-blob-rejection` and `reasoning-effort-downgrade`.

This was found while mapping the shared stage and cause substrate that landed as #5266 (`043aa435ff8f86095f55cbe08f74d45b9858da59`). It is the restated-union class `AGENTS.md` warns about: because the GUI union is hand-maintained rather than derived from the roster it describes, a member added on the durable side compiles cleanly on both branches and silently stops being displayed.

The fix is to derive the GUI union from the durable roster instead of restating it, so that adding a recovery kind is a typecheck failure in the GUI until it has a label. It needs new strings across the locale catalogs and a screenshot of the changed dialog, which is why it is filed rather than folded into that branch.

### Reproduction

Produce an attempt whose recovery kind is one of `key-401`, `oauth-account-429`, `opaque-blob-rejection` or `reasoning-effort-downgrade`, then open the Logs page and inspect the attempt row. The recovery kind has no localized label, unlike the nine kinds the page declares.

### Version

2.61.0-dev (`dev` at `043aa435ff8f86095f55cbe08f74d45b9858da59`)

### Operating system

Any

### Provider and model

Any

### Logs or error output

```shell

```

### Screenshots and supporting files

None.

### Redacted configuration

```json

```

### Checks

- [x] I searched existing issues and documentation.
- [x] I removed secrets, tokens, account details, request credentials, and personal data.

Ingwannu:
Confirmed against current `dev` (`600075d23e90f58636a9c12ed4431ea5fad30b7a`). `src/usage/log.ts` defines the durable 13-member `AttemptRecoveryKind`, while `gui/src/pages/Logs.tsx` independently restates only nine members. The four named values therefore pass through the management payload but miss the localized lookup and fall back to raw IDs.

[...]

PR #5332, opened 2026-09-20
fix: make a failed browser launch and a failed account refresh visible (#5261)

lidge-jun:
## Summary

Two remainders #5288 left open on #5261, plus a correction to the plan's #5292 row.

**`ocx login <provider>` now says when no browser opened.** Both generic CLI logins called `void openUrl(...)`, so on a host with no launcher each printed a URL, announced that it was opening a browser, and then asked a question that assumes it opened — indistinguishable from a login that is working. The Codex account login already reported this; these two did not.

Reporting it is an ordering problem rather than a message problem. `OAuthController.onAuth` returns `void` and all thirteen provider call sites invoke it as `ctrl.onAuth?.(...)` and continue, so the launcher's answer arrives after the flow has moved on — on a callback-server provider, after `#waitForCallback` has already called `onManualCodeInput`. A warning written at that moment lands on the line the user is typing on. Making `onAuth` awaitable would change the controller contract and every call site, which is far more than this defect warrants. So the launch reports itself when it settles, and the two things that could collide with it wait on that report: the manual-code prompt awaits it before asking, and the key login awaits it before it constructs a reader at all. A polling provider that never prompts is still told before the login claims success.

`BROWSER_LAUNCH_FAILED_HINT` keeps its ChatGPT-specific second line about the fixed callback port and `--device`, and now derives its first line from the shared notice, so the sentence has one home across all three logins.

**A failed account refresh no longer reads as a current roster.** After a successful load, a failed account read left the Codex pool reporting `ready` with the rows it had before. Keeping those rows is deliberate — blanking a populated pool because one 30s poll missed is its own defect — but nothing distinguished a list the server had just confirmed from one that predated a failure. The reported shape: add an account, the read that would bring it over fails, and the dashboard shows the older accounts with the new one simply absent.

The controller now carries `refreshFailed` beside `loadState`, for the same reason `refreshing` already lives there: `loadState` answers what the surface can draw, and a warm failure does not change that answer. It clears where the rows are painted, inside the `/accounts` success branch, rather than at the settle — `/active` runs on a much longer budget, and clearing at the settle would label the rows that just replaced the stale ones as the stale ones for that whole window. A cold failure still replaces the surface with its existing error, and the banner renders only when rows survived.

**#5292 needs no change.** The plan's table describes `gui/src/pages/Logs.tsx` as restating the recovery-kind union with nine of thirteen members. That stopped being true two hours before the plan was written: `555f0cacdf` (#5300) replaced the copy with the durable roster, closed the label map with `satisfies Record<AttemptRecoveryKind, string>`, and added the four missing labels to all ten catalogs. `tests/usage/request-outcome-agreement.test.ts` holds both halves. The row is stale, not open.

## Verification

Static review plus hosted CI at this exact head.

**NOT RUN** — no local suite, no individual test file, no `bun run typecheck`, no build, no install, no `ocx` execution, no service restart, and no credential or configuration change. This lane is not permitted to run them; CI on this head is the evidence.

Checked by reading, because these are what a merge breaks rather than what a branch's own CI catches:

- No file this changes appears in `tests/fixtures/file-size-baseline.json`; the ten i18n catalogs are in its `exempt` list.
- `tests/oauth/oauth-login-cli-browser-launch.test.ts` is registered in both `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`. `tests/test-layout.test.ts` walks the repository-root `tests/` only, so the new `gui/tests/` file needs no registration.
- `CodexAccountLoadState` gained no member. `CodexAccountPoolController` gained one, and the source-oracle roster in `gui/tests/codex-account-pool-controller.test.ts` names it.
- The one new i18n key is in all ten catalogs, which `gui/tests/locale-parity.test.ts` and `gui/tests/claude-desktop-locale.test.ts` both require.
- `CodexAccountPoolLoadStates` stopped restating the load-state union and derives it from the controller.

New coverage, with the non-regression cases labelled as such in the files:

- `tests/oauth/oauth-login-cli-browser-launch.test.ts` — records the warning and the prompt as an ordered event log, so what is asserted is that the failure reaches the user *before* the question, not merely that it reaches them. Covers the callback-server shape, the polling shape that never prompts, a launcher that throws, the key-login prompt, and the report helper's own ordering. The launcher, the terminal and the provider are injected; nothing spawns a browser or attaches to stdin.
- `gui/tests/codex-account-pool-stale-refresh.test.tsx` — drives the controller through a warm failure and back, including the case where `/accounts` has landed and `/active` has not, and asserts the rendered banner against a mounted DOM: present with surviving rows, absent after a successful refresh, and never in place of the cold error.

### The GUI screenshot gate

`enforce-target` requires a screenshot for a PR that mentions `gui`. Producing one needs `bun run build:gui` and a running proxy, and this lane is forbidden to build, install or run `ocx`, so there is no screenshot to attach.

What can be checked instead is in the diff. The rendered markup is asserted against a mounted DOM in `gui/tests/codex-account-pool-stale-refresh.test.tsx` rather than described: the banner appears only when rows survived, carries the catalog string, and yields to the cold error when both conditions hold. The visual change is one element reusing the existing `.pwi-auth-state` block with a `--stale` modifier that parallels `--error` line for line, built from the `--amber`/`--amber-soft` pair the theme already defines with `light-dark()`. It renders above the account cards, in the slot the cold error already occupies, so it displaces nothing.

## Checklist

- [x] Targets `dev`
- [x] Focused regression tests accompany the `src/` behaviour changes, and assert the ordering and staleness contracts rather than that a call succeeded
- [x] User-facing strings added to all ten locale catalogs
- [x] File-size ratchet and union-exhaustiveness checked against this head
- [x] No credential, token or request-body logging introduced
- [ ] Screenshot for the GUI change — cannot be produced under this lane's constraints; see above

Closes #5261

#5292 is already closed by #5300 and is not touched here.

## Summary by CodeRabbit

- **New Features**
  - Added clearer warnings when the browser cannot open during OAuth or API-key login, while allowing manual login to continue.
  - Manual login prompts now wait until browser-launch status is confirmed.

- **Bug Fixes**
  - Account lists remain visible when a background refresh fails, with a stale-data warning and retry option.
  - Cold-load failures continue to display the appropriate error state.

- **Localization**
  - Added stale account refresh messaging across supported languages.


Ingwannu (changes requested):
Requesting changes on exact head `7e55da746ca644e6e2fc9f136e376139f779dd47`.

The warm-empty boundary still hides the refresh failure. A successful `/accounts` response with `accounts: []` is a valid last-good roster (`hasLoadedRef.current === true`). If the user then adds their first account and the next `/accounts` read fails, the controller correctly leaves `loadState` as `ready` and sets `refreshFailed = true`, but `CodexAccountPoolLoadStates` renders the stale notice only when `accountsCount > 0`. The surface therefore continues to present the old empty state as current, and the newly added account is silently absent — the same defect this PR is meant to expose.

Please show the warm-refresh failure whenever a prior load succeeded, including an empty prior roster, while keeping the cold `loadState === "error"` precedence. Add a controller/render regression for: successful empty load -> server gains first account -> refresh fails -> stale notice and retry are visible. The GUI screenshot requirement (or an explicit maintainer waiver) and exact-head required CI also need to be satisfied before merge.

lidge-jun:
## 리뷰 · 우선순위 61 / 80

이 PR은 #5261에 남았던 두 가지를 고칩니다. 첫째, `ocx login`이 브라우저를 못 열어도 “열었어요”처럼 보이고, 그 다음에 붙여넣기 질문을 던지던 문제입니다. 이제는 열기 실패 경고가 질문보다 먼저 나옵니다. 둘째, 계정 목록을 다시 읽다 실패해도 예전 목록이 `ready`로만 보이던 문제입니다. 목록은 그대로 두고, 위에 노란 안내와 다시시도 버튼을 붙입니다. 문장 한 줄은 `BROWSER_LAUNCH_FAILED_NOTICE`에 모았고, Codex 계정 로그인의 둘째 줄만 그 뒤에 붙입니다. #5292는 #5300이 이미 끝냈다고 계획 문서에만 적어 두었습니다. base는 `dev`이고 draft입니다. 테스트는 경고→질문 순서와 배너 표시를 순서·DOM으로 잡습니다.

`src/oauth/login-cli.ts` `onAuth` - `kiro` 같은 흐름은 URL 없이 `onAuth({ url: "" })`만 부릅니다. 지금은 그 빈 주소도 `openUrl`에 넘겨서 “브라우저를 못 열었다” 경고가 납니다. URL이 비어 있으면 추적·경고를 건너뛰고, 안내 문장만 보여 주세요.

`src/oauth/login-cli.ts` `handleOAuthLogin` - `runLogin`이 `onAuth` 뒤에 바로 실패하면, 맨 아래 `await browser.settled()`까지 못 갑니다. 로그인 에러가 경고보다 먼저 나갈 수 있습니다. `withPrompt`를 `try/finally`로 감싸고, `finally`에서 `settled()`를 기다리세요.

`src/oauth/login-cli.ts` `withPrompt` - readline을 `runLogin` 시작 전에 엽니다. 파이프로 코드가 먼저 들어오면, 질문이 걸리기 전에 그 줄을 먹어 버릴 수 있습니다. 질문이 실제로 나갈 때 인터페이스를 만들고, 만들었을 때만 닫으세요.

메인테이너의 판단이 필요한 지점

GUI 스크린샷 게이트는 이 레인에서 빌드·실행이 금지라 비어 있습니다. DOM 테스트로 배너를 잡은 것이면 게이트를 넘길지, 스크린샷을 따로 받을지 정해 주세요. #5292를 이 PR에서 “이미 닫힘”으로만 정리하는 것도 맞는지 한 번만 확인해 주세요. types/config 분할이나 미리보기 배포와는 무관합니다. 같은 결함의 열린 중복 PR은 보이지 않습니다.

너의 추천

빈 URL `onAuth` 가드와 `settled()`의 `finally` 이동을 먼저 넣고, 그 두 경로 테스트를 추가하세요. readline 지연 생성은 같이 고치면 좋습니다. CI가 이 head에서 초록인지 확인한 뒤, 스크린샷 게이트만 메인테이너 규칙에 맞춰 처리하고 승인 요청하세요.

이 댓글은 grok-bot이 작성했습니다

PR #5268, opened 2026-09-20
fix(codex,usage): bind account-scoped evidence to credential generation and add privacy-bounded cache diagnostics

lidge-jun:
## Summary

Bundles 8 and 9 of the post-2.60.0 consolidation batch (devlog _plan/260920_meaning_preservation_batch, phase 2) are the same defect at two layers: an observation attributed to an account or credential generation must not survive its replacement, and a cache-warm binding must not be retired by a hint. This lane carries the five existing contributor pull requests and closes the gaps none of them covered alone.

**Account/credential-generation evidence (bundle 8)**

- Carries #5214 (entitlement refreshes fenced behind native-main admission), #5145 (learned reasoning-effort refusals scoped to the credential identity), and #5229 (Cursor live wire-spelling and Max-Mode evidence scoped by destination+credential).
- The Cursor and Devin live *roster caches* were still scoped by provider name alone: a credential switch could read the previous account's fresh or stale plan roster, and one credential's discovery failure suppressed the next credential's first fetch while offering it the previous account's stale list. The roster cache entry is now bound to an irreversible credential fingerprint (the Qoder precedent), the stale fallback is credential-scoped, and a credential with no roster of its own fetches through another credential's cooldown. Quota/rate-limit health stays account-scoped by design: it describes the subscription, not the token generation, and the 401/403 quarantine is already generation-fenced.
- A cancelled data-plane request's late entitlement refresh no longer commits: /v1/models now passes its own signal into admitted entitlement resolution, and the native-main token refresh re-checks the signal after the upstream grant resolves and before the auth.json write. False/unknown/absent stay distinct on the entitlement path (confirmed bit separate from the roster set, tri-state public state, pinned by existing tests).

**Cache affinity and diagnostics (bundle 9)**

- Carries #5209 (shared cache-affinity bindings retire on genuine 100% exhaustion, not on the proactive auto-switch threshold) and #4793 (per-model cache metrics on the Usage page; this lane owns that surface).
[...]

PR #5333, opened 2026-09-20
fix(gui): report the measured cache hit rate and make the usage models table readable

lidge-jun:
## Summary

The dashboard Usage models table showed `gpt-5.6-sol` with 2.8B cache hits and a hit rate of `—`. `gpt-6-astra`, `k3[1m]`, `gemini-3.8-flash`, `gpt-5.6-luna` and `grok-4.6` read the same way, and the table around it was squeezed hard enough that token totals folded onto a second line.

**The hit rate was not missing, it was withheld.** `calculateCacheHitRate` in `src/usage/summary.ts` averages cache reads over `cacheObservedInputTokens` — the input tokens whose cache detail was actually reported — and returns `null` only when nothing was observed. That denominator is the #4546 contract in `structure/gui-and-management-api.md`, and a provider that reports reads and never reports writes is observed and has a rate. The GUI then required that denominator to cover the row's **entire** input before it would render the number:

```tsx
model.cacheObservedInputTokens >= model.inputTokens ? model.cacheHitRate : null
```

One request in the row with no cache detail — a locally answered turn, an unreported usage record, a row written by an older proxy — drops the denominator below `inputTokens` and blanks the column. For a busy model that is every row, which is how six models with billions of measured hits all rendered an em dash. The gate arrived with the cache columns in #5268; it was never the summary's rule.

The cell now renders whatever the summary supplied, because the summary already refused to supply a number it could not justify. Coverage became a note instead of a gate: `usage.cacheHitRate.partial` names the measured and total input tokens on a partially observed row, `usage.cacheHitRate.unmeasured` explains the em dash on a row where nothing reported cache detail. That row — no basis at all — is the only one that still shows `—`. The note is carried as both a `title` and an `sr-only` span, since a `td` is not focusable and a `title` never reaches a keyboard or a touch screen. **No server change:** the denominator, the provenance split and the `null` are correct as they stand, and the structure doc that owns the contract stays accurate.

Three layout defects in the same table:

**Column order.** `Model, Provider, Share, Tokens, API list-price`, then `Requests, Measured, Input tokens, Output tokens, Cache hits, Cache writes, Hit rate`. Identity, then the three figures a reader compares models on, then the evidence behind them. Share and price were previously buried behind five cache columns.

**Sideways scroll and pinned identity columns.** `.tbl` is `width: 100%`, so twelve columns divided the shell between them until eight-digit token totals wrapped. The models table is `width: max-content; min-width: 100%` now and the shell scrolls — `.tbl-wrap` was already `overflow-x: auto`, so nothing else had to move. Model and provider are `position: sticky` at fixed widths so a row stays identifiable while its numbers scroll; both offsets are one `var(--space-3)` step negative, the same trick the sticky header plays with `top`, so a stuck cell repaints the scrollport padding it slides over, and a value too long for its column keeps its full text in the cell's `title`. Under 720px the pinning stands down.

Every selector is doubled as `.tbl.usage-models-tbl`, which is load-bearing: `styles-usage-workspace.css` is `@import`ed from the top of `styles.css`, so the whole of `styles.css` cascades after it, and a single class ties `.tbl { width: 100% }` on specificity and loses on source order. The rules already in that file buy the same margin with a `.usw-section` prefix. `gui/src/styles.css` sits exactly at its 2958-line file-size cap, so none of this could go there and none of it did.

**The exclusion caption.** `(56 requests excluded)` shared a line with the amount and folded mid-phrase; it is a block now, so the amount is the first line and the caption is the second. Its leading space stays in the markup — a block box drops leading white space when it lays out, so the rendered cell is unchanged and anything reading the cell as one string sees exactly what it saw before.

## Screenshots

Rendered from the real `UsageModelsTable` against a fixture report, captured headless at the commit below. The same three files are committed under `devlog/_plan/260920_round2_followups/r6-usage-table/`.

**Column order, hit rate, and the two-line price caption.** `gpt-5.6-sol` reports 2.8B cache hits and 0 cache writes and now shows **95%** where it showed an em dash. `glm-5.3-flash` reported no cache detail at all and is the one row that keeps the dash. Every amount sits on its own line with `(56 requests excluded)` beneath it, and no number wraps.

Usage models table at desktop width

**Scrolled sideways with model and provider pinned.** The numeric columns slide under an opaque pair; the hovered row stays opaque too, which is the defect the last commit fixes — `--hover` is a 3% overlay, so using it as the whole background of a pinned cell let the scrolled columns read through it. No CSS source assertion can see that.

Models table scrolled right with pinned identity columns

**Below the 720px breakpoint.** The pinning stands down (`position: static`, model column back to its natural width) and the table still scrolls sideways rather than compressing. Verified in the page as `matchMedia("(max-width: 720px)").matches === true` at `innerWidth: 675`.

Models table below the 720px breakpoint

## Verification

**This is a GUI change, and it cannot satisfy the screenshot gate.** Builds are not permitted in this lane, so no dashboard was rendered to photograph. The evidence offered instead is the column order and cell layout written out above, the regression assertions below, and exact-head hosted CI.

- `gui/tests/usage-custom-range.test.tsx` — the fixture already contained `partial-cache-model`, a row with `cacheObservedInputTokens: 500` against `inputTokens: 1000` and `cacheHitRate: 0.9`, and asserted that it rendered `—`. It now asserts `90%` with both the `title` and the `sr-only` note, the fully observed row asserts neither, and the row that reported no cache detail asserts the unmeasured sentence. The header sequence asserts the new twelve-column order. Both note assertions derive their expected text from the English catalog rather than restating it.
- `gui/tests/usage-layout.test.ts` — new source-oracle case binding the `max-content` sizing, the pinned columns and their offset arithmetic, and the block caption, including a negative assertion on the single-class selector. Dropping a rule fails a test instead of quietly restoring the squeeze.
- Static review of the cascade, the specificity of every new rule against `styles.css`, the rendered cell array of all three fixture rows, and locale key parity across all ten catalogs. An adversarial second pass reproduced the `.tbl` cascade defect independently before it was pushed; it is fixed here.
- Structure ownership reviewed: `structure/overview.md`, `structure/gui-and-management-api.md` and `structure/design-methodology.md` map to `gui/`. The cache-provenance contract they describe is unchanged by a presentation fix, so none needed an edit.
- **NOT RUN:** `bun run test`, `bun test` on any single file, `bun run typecheck`, `bun run lint:gui`, `bun run build:gui`, `bun install`, any `ocx` execution. Hosted CI at this exact head is the only execution evidence. GUI lint, typecheck and the `gui` test suite all run in the `gates` job of `Cross-platform CI`, which a branch push does not trigger and this pull request does.

File-size ratchet: no changed file has a cap in `tests/fixtures/file-size-baseline.json` — the ten locale catalogs are `exempt`, and `gui/src/pages/Usage.tsx`, `gui/src/styles-usage-workspace.css` and the two test files have no entry. `gui/src/styles.css` is untouched at its cap.

## Checklist

- [x] Targets `dev`
- [x] Behaviour change carries a focused regression test near the existing tests for that subsystem
- [x] New strings added to all ten locale catalogs with identical placeholder spellings
- [x] No server-side change; the usage-summary contract and its structure doc are unaffected
- [x] `gui/src/styles.css` unchanged at its exact file-size cap
- [ ] Screenshot of the UI change — **cannot be produced in this lane**, see Verification
- [ ] `bun run typecheck` and `bun run test` locally — **not run in this lane**, see Verification

## Summary by CodeRabbit

* **Bug Fixes**
  * Usage model tables now display cache hit rates when partial cache details are available.
  * Rows without cache details clearly indicate that a hit rate cannot be calculated.

* **Improvements**
  * Added explanatory tooltips and accessible notes for cache-rate coverage.
  * Reordered columns to prioritize model identity and comparison figures.
  * Enabled horizontal scrolling and pinned Model and Provider columns on wider screens.
  * Improved excluded-request caption formatting.

* **Localization**
  * Added cache hit-rate coverage messages across supported languages.


Ingwannu (changes requested):
Requesting the remaining merge evidence on exact head `71b1b36d9d0a52ae58962980bfa7158edbee539b`.

The cache-rate logic is sound: the server summary already computes over `cacheObservedInputTokens` and returns `null` when no basis exists, so the GUI should present that measured result with a coverage caveat instead of imposing a second full-coverage gate. The column order and localized accessible notes also match that contract; I found no functional code blocker in those paths.

The required GUI evidence is still missing, though. This PR materially changes twelve-column layout, horizontal scrolling, two sticky columns, truncation, hover backgrounds, and the <=720px fallback. Source-oracle CSS assertions cannot show clipping, sticky overlap, z-index/background seams, touch-width behavior, or light/dark rendering. Please attach rendered screenshots (desktop scrolled state plus the narrow fallback, ideally light and dark) or have the owner explicitly apply the repository’s screenshot waiver with a recorded reason. Required exact-head CI must also finish green before approval.

lidge-jun:
## 리뷰 · 우선순위 34 / 80

이 PR은 Usage 모델 표에서 캐시 히트율이 잘못 가려지던 버그를 고치고, 같은 표를 읽기 쉽게 만든다. 서버의 `calculateCacheHitRate`는 이미 `cacheObservedInputTokens`(캐시 상세가 실제로 보고된 입력 토큰)만으로 평균을 내고, 관측이 전혀 없을 때만 `null`을 준다. 그런데 GUI는 `#5268` 이후 `cacheObservedInputTokens >= inputTokens`일 때만 숫자를 보여 줬다. 행 안에 캐시 상세가 없는 요청이 하나라도 있으면 분모가 전체 입력보다 작아져 칸이 `—`가 되고, 바쁜 모델은 거의 전부 그렇게 보였다. 이번 변경은 그 두 번째 게이트를 빼고 요약이 준 값을 그대로 그린다. 부분 관측이면 `usage.cacheHitRate.partial`로 측정/전체 토큰을 설명하고, 관측이 전혀 없으면 `unmeasured`로 `—`를 설명한다. 설명은 `title`과 `sr-only`에 같이 넣었다. 더불어 열 순서를 “모델·프로바이더·점유·토큰·정가 → 요청·측정·입출력·캐시”로 바꾸고, 표는 가로 스크롤 + 앞 두 열 sticky로 두며, 제외 요청 안내 문구는 금액 아래 한 줄(`usage-cost-note`)로 떨어뜨린다. i18n 10개 카탈로그에 문자열이 들어갔고, 회귀 테스트와 CSS source-oracle도 붙었다. base는 `dev`다. 서버/구조 문서 계약은 건드리지 않았다.

라인 - `gui/src/pages/Usage.tsx` (`cacheHitRateTitle` / 히트율 셀): 부분 관측 행은 화면에는 `90%`처럼 숫자만 보인다. 설명은 마우스 `title`과 `.sr-only`에만 있다. `td`는 포커스가 안 되므로, 호버하지 않는 시력 사용자·터치 사용자는 “행 전체 입력에 대한 히트율”로 오해할 수 있다. Codex P2와 같은 지점이다.
라인 - `gui/src/styles-usage-workspace.css` (`nth-child(1)`/`nth-child(2)` sticky): sticky 대상이 열 순서에 묶여 있다. 열을 다시 바꾸면 CSS와 테스트 인덱스를 같이 고쳐야 한다. 주석·source-oracle로 막아 두긴 했지만, 열 순서 변경 시 깨지기 쉬운 계약이다.
라인 - `enforce-target` / 게이트 댓글: exact head에서 UI 스크린샷 없음으로 quality gate가 DRAFT다. 본문도 이 레인에서는 빌드·스크린샷이 불가하다고 적었다. `gates` 등 GUI 테스트는 통과한 상태다.

메인테이너의 판단이 필요한 지점

부분 관측 히트율을 “숫자 + 숨은 설명”으로 둘지, 표 안에도 짧은 각주/배지를 보이게 할지. 지금 설계는 접근성(스크린리더)은 챙겼지만, 시력·터치 쪽의 “보이는 주의”는 약하다.

스크린샷 게이트를 면제할지, 아니면 다른 환경에서 한 장만 올리고 머지할지. Owner도 exact head 증거 보강을 요청한 상태다.

너의 추천

히트율 게이트 제거와 열/스크롤 수정 방향은 맞고, 서버 계약을 GUI가 다시 막던 실수도 잘 짚었다. 테스트도 partial→`90%`와 unmeasured→`—`를 직접 고정한다. 머지 전에 (1) 부분 관측 주의문을 눈에 보이게 할지 한 줄 정하고, (2) 스크린샷 면제 또는 첨부만 처리하면 된다. types.ts/config.ts 분할·중복 tip 닫기나 프리뷰 배포 이야기는 이 PR과 무관하다.

이 댓글은 grok-bot이 작성했습니다

PR #5338, opened 2026-09-20
ci(desktop): keep unsigned verification out of updater signing

lidge-jun:
## Summary

The macOS `macos widget + bundle` verification job successfully builds the app, widget extension, and sidecar, then Tauri also tries to create an updater archive and refuses because the job intentionally has no `TAURI_SIGNING_PRIVATE_KEY`.

This PR passes `createUpdaterArtifacts: false` only to that unsigned CI invocation. The committed Tauri release configuration still enables updater artifacts, and the release workflow still requires the signing secrets before publication.

## Scope

The final PR diff against current `dev` is exactly one file: `.github/workflows/ci.yml`.

Unchanged:

- workflow triggers and permissions
- action pins and checkout credential behavior
- repository secrets and key material
- `desktop/src-tauri/tauri.conf.json`
- `.github/workflows/release.yml`
- release eligibility and updater publication checks
- post-build assertions for the app, embedded WidgetKit appex, executable sidecar, and code signatures

The temporary dangling-role test repair was removed after #5344 landed the focused fix directly on `dev`. This branch now includes current `dev` (`ab90bea6b1e53ba5c8f8c1eda538e88ec70682f3`) and contains no overlapping test diff.

## Validation

- `git diff --check origin/dev...HEAD` — clean
- final file list — `.github/workflows/ci.yml` only
- exact-head hosted CI is the required execution evidence, especially `macos widget + bundle`

## Security and release review

The command uses a repository-authored literal JSON override, not event-controlled input. It does not add release credentials to the verification job or relax release signing. Explicit workflow/security review has been completed; merge remains gated on replacement exact-head hosted CI.

## Integration note

No Go counterpart exists for this GitHub Actions-only change. The remote repository currently has no `dev2-go` branch despite stale transition wording in `MAINTAINERS.md`.

## Checklist

- [x] Targets `dev`; `main` is untouched.
- [x] Final diff is workflow-only and minimal.
- [x] Explicit workflow/security review completed.
- [ ] Replacement exact-head hosted CI is fully green.

PR #5342, opened 2026-09-20
Rework #4942 and #4989 into one ambiguous-resend gate with one grant per request

lidge-jun:
## Summary

A native Responses send can fail with the caller having observed nothing in two different places: the connection dies before any response head, or it dies after the head while the SSE body has carried only control events. #4942 handled the first and #4989 handled the second, and each bought its own replacement send. A request that hit both would send a possibly-executed inference twice more, which is the exact duplicate the refusal exists to prevent. This reworks the two into one change.

`src/lib/request-resend-gate.ts` is the single gate. It derives stage, cause, permission and send class from the table #5266 landed in `request-failure-model.ts` and adds only the operator override that table already names for `refused-ambiguous`. Three bounds apply at once: the provider opts in with `providers.<name>.retryOnReset`, the request has to be one `reset-replay.ts` can judge self-contained (`store: false`, complete input, no server-side continuation state, client-executed tools only), and the whole logical request holds **one** replacement grant no matter how many stages ask. The grant lives on the request's execution budget beside the physical-send ledger, so a combo child that derives its own scope draws on the same counter instead of holding a second.

Concretely, before and after for a provider with `retryOnReset: {}` whose connection resets before the head and whose replacement stream then dies after `response.created`:

- before this PR (the two branches merged): two replacement sends, each charged by a different counter
- after: one replacement, charged once, and the second stage answers `ambiguous-allowance-spent` and settles as the existing 429 `upstream_reset_replay_refused`

Four things changed relative to the original branches, and they are worth a reviewer's attention:

- `replayResets: number` became a `claimAmbiguousResend()` callback on `ResetRetryOptions`. A count spread into every dispatch leg is a count every leg holds, and the rotation, refresh and same-target 429 legs all carry the same turn.
- The post-header gate no longer requires `response.created`. The preflight reports the stage it observed — `headers-only` before any parsed event, `protocol-prelude` after `response.created`, `semantic-output` once anything else arrives including a payload the inspector could not parse — and the table gives `headers-only` the same commitment, so refusing it was refusing a row the table permits.
- The 401 replay leg is routed through `fetchWithTransientRetry`. It previously rejected straight into `transportFailureResponse`, which answers 502, and 502 is the status the Codex client re-sends the whole turn on.
- `retryOnReset.attempts` became `retryOnReset.replacements` (1..2, default 1). The semantics genuinely changed: it is no longer a per-leg total send count but the number of duplicate inferences one logical request may risk. The old key is rejected rather than silently ignored.

The deferred SSE preflight only wraps a body when the provider opted in, so a proxy that configures nothing buffers nothing and its time to first byte is unchanged.

Carries #4942 (Fred Amartey) and #4989. Both remain open; neither is superseded by merging this.

**#4191 and #5180 are not closed by this PR.** Both are recorded with evidence in `devlog/_plan/260920_round2_followups/050_r4_remainders.md`:

- #4191: the gate deliberately excludes WebSocket bodies, so the durable threading is untouched. The investigation did find that `classifyCodexWsFailure` reports `after-response-started` — projected as `semantic-output` — as soon as `relayedEvents > 0`, and the exchange counts `response.created` as a relayed event. The failure model puts `response.created` in `protocol-prelude`. Fixing it needs a counter, and `CodexWsStageRecord` derives from `CodexWsFailureStage` by `Omit`, so a naive addition lands in a persisted record whose read-back whitelist would reject every existing row. It belongs with the lane that threads `failureStage`/`failureCause` into the record.
- #5180: a received 429 is `headers-only` with cause `rate-limit`, which the table already answers `permitted` and funds from the transient class — it needs no grant and no override. The reported symptom is that `rateLimitRetryPolicyFor` returns null for every provider but the OpenCode Go destination, and that `keyCooldowns` cannot be written by a single-key provider at all.

## Verification

Static review plus exact-head hosted CI. Local suites, individual test files, `bun run typecheck`, builds, installs and live `ocx` execution were **NOT RUN** — this lane verifies through hosted CI only. `bun run structure:check` and `bun run privacy:scan` were also **NOT RUN** for the same reason; the structure edits were made against `structure/AGENTS.md` and `structure/manifest.json` by hand, and no new source area or manifest entry was needed because both new modules sit under the already-mapped `src/lib/` and `src/server/`.

Checks done by reading:

- File-size ratchet: no touched file is in `tests/fixtures/file-size-baseline.json` except `src/config.ts`, which sits at its cap of 460 and is modified in place on one existing export line, so it stays at 460. `passthrough-dispatch.ts` grows to 1762 against the 2000 threshold for untracked files.
- Union-exhaustive classes: `RESEND_REFUSALS` is a frozen roster with the type derived from it, and a test asserts every declared member is reachable and every produced refusal is declared, so the roster cannot carry a dead name or miss a live one.
- No source constant or count is restated. The gate tests iterate `REQUEST_FAILURE_STAGES` × `REQUEST_FAILURE_CAUSES` and `ATTEMPT_RECOVERY_KIND_ROSTER` and compare against `resendPermission`, `resendSendClass` and `causeForRecoveryKind` rather than against written-out expectations.
- New test files are registered in both `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`.
- No GUI file is touched, so the screenshot gate does not apply.

Regression coverage added: `tests/lib/request-resend-gate.test.ts` (the gate over the full cross product, and the one-grant-per-request property that is the whole point of the rework), `tests/lib/upstream-retry.test.ts` (the claim is asked once, never when there is no send left, never for a replay-safe operation, and a non-reset failure after a replacement settles as the refusal rather than a rejection), `tests/lib/execution-budget-permits.test.ts` (parent and derived child share one grant), `tests/responses/responses-reset-replay.test.ts` (the body judgment and the lazy allowance), `tests/routing/combo-stream-preflight.test.ts` (the observed stage at each boundary) and `tests/server/management-provider-reset-replay.test.ts` (the write-boundary validation, including that `retryOn429` keeps its own field name now that both validators share one formatter).

## Checklist

The refusal body is synthesized and never carries the upstream exception, which can hold credentials or request data. The management write boundary redacts secret-shaped unknown field names and provider names, and a test pins that. The replacement send re-checks credential currency through `selectionIsCurrent` before dispatch and re-runs the Codex reserve guard, so it cannot go out on a credential the request has since moved off. `retryOnReset` is off unless the operator writes it, and it degrades to absent on a malformed hand edit rather than sending the operator through invalid-config recovery.

## Summary by CodeRabbit

* **New Features**
  * Added the optional `retryOnReset` setting for native Responses providers.
  * Eligible self-contained requests can receive limited replacement sends after pre-response disconnects or control-only SSE interruptions.
  * Replacements are capped per logical request and are not attempted after output or tool calls are observed.

* **Documentation**
  * Added configuration guidance for `retryOnReset` across supported languages.
  * Documented replacement eligibility, limits, and billing considerations.

* **Bug Fixes**
  * Improved validation and handling of reset-replay configuration.
  * Improved SSE failure classification to distinguish unobserved failures from committed output.


lidge-jun:
## 리뷰 · 우선순위 74 / 80

이 PR은 “연결이 끊겼는데, 클라이언트가 아직 답을 못 본 경우”에만 요청을 한 번 더 보낼지 정하는 문을 하나로 모아요. 예전에는 머리(헤더) 앞에서 끊긴 경우(#4942)와 머리 뒤에서 SSE만 조금 받은 경우(#4989)가 각각 따로 다시 보냈어요. 같은 요청이 두 번 다 걸리면 추론이 두 번 더 나갈 수 있었어요. 지금은 `request-resend-gate.ts` 한곳에서 보고, 요청 예산에 있는 교체 허락을 딱 한 번만 써요. 공급자가 `retryOnReset`을 켠 때만 동작하고, 몸통이 다시 보내도 같은 추론만 하는 요청인지(`store: false`, 서버 이어가기 없음, 클라이언트 도구만)도 같이 봐요. base는 `dev`예요. #4191·#5180은 이 PR이 안 닫아요.

방향과 게이트·예산·테스트 짜임은 좋아요. 그런데 아래 한 줄이 그대로면, 켠 공급자에서도 교체가 사실상 안 나가요. 로컬 스위트는 안 돌렸다고 적혀 있고, 이 테스트가 CI에서 먼저 깨질 거예요.

`src/server/responses/reset-replay.ts` `ambiguousResendAllowanceFor` — 두 번째 인자를 요청 몸통으로 보고 `selfContainedResponsesBody(inboundBody)`를 바로 불러요. 그런데 `passthrough-dispatch.ts`는 `requestIsSelfContained`(함수)를 넘기고, `tests/responses/responses-reset-replay.test.ts`도 `() => true` / `() => false`를 넘겨요. 함수는 몸통 객체가 아니라서 `selfContainedResponsesBody`는 항상 `false`예요. 그러면 게이트는 늘 `ambiguous-request-not-replayable`로 거절해요. 테스트는 `selfContained === true`와 “게으른 판정”(만들 때는 안 부르고, 게이트가 볼 때 부름)을 기대해요. 시그니처·구현을 콜백(+ getter)으로 맞추거나, 호출부를 몸통으로 바꿔야 해요. 지금 타입은 `unknown`이라 컴파일은 통과해요.

`src/server/responses/passthrough-dispatch.ts` / `combo-stream-preflight.ts` — `replayReadErrors` 결과가 `accepted`에서 `read-error`로 바뀌어요. 콤보·opaque 경로는 `failed`만 특별 취급하고 나머지는 응답을 그대로 써서, 당장 깨지진 않아 보여요. 다만 `kind === "accepted"`만 보는 코드가 생기면 조용히 빠질 수 있어요.

메인테이너의 판단이 필요한 지점

이 PR이 합쳐지면 #4942·#4989는 같은 일을 두 갈래로 하던 옛 PR이에요. 본문은 “둘 다 열린 채, 이 PR이 대체한다고 쓰지 말자”고 해요. 합친 뒤 둘을 닫을지, 남은 조각만 이슈로 남길지 정해 주세요. `retryOnReset.attempts`는 거절되고 `replacements`만 받아요. 옛 키를 쓴 설정이 있으면 로드 시 기능이 꺼진 것처럼 보여요.

너의 추천

`ambiguousResendAllowanceFor`를 테스트·호출부와 같은 계약으로 고치세요. 두 번째 인자는 `() => boolean`이고, `selfContained`는 그 콜백을 읽는 getter여야 해요. 그게 통과하기 전에는 합치지 마세요. 합친 뒤에는 #4942·#4989를 닫거나, 닫지 않을 이유를 PR/이슈에 한 줄로 남기세요. types.ts·config.ts는 재수출만이라 이 PR에서는 그대로 둬도 돼요.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
Requesting changes on exact head `862502add9411111dc335ee90cfd4aa370aebab4`.

`src/server/responses/reset-replay.ts:93-103` treats the second argument to `ambiguousResendAllowanceFor` as a parsed Responses body and eagerly calls `selfContainedResponsesBody(inboundBody)`. However, `src/server/responses/passthrough-dispatch.ts:722-733` and the new tests pass a lazy `() => boolean` callback. Because the parameter is typed as `unknown`, TypeScript accepts this, but the function object is not a request body, so `selfContainedResponsesBody` always returns false. The opt-in is therefore functionally disabled and every otherwise eligible retry is refused as `ambiguous-request-not-replayable`.

Please make this contract consistently callback-based and preserve the intended laziness, for example by accepting `requestIsSelfContained: () => boolean` and exposing `selfContained` through a getter that invokes it only when the gate reads it. Then rerun replacement exact-head CI. The current head already has failing test shards, so it must not merge as-is.

Ingwannu (commented):
Re-reviewed replacement head `4c94b3384cebc1ef365b1860ae6e14cb533633cb`. The requested blocker is fixed correctly: `ambiguousResendAllowanceFor` now accepts `() => boolean`, exposes `selfContained` through a getter, and therefore preserves both the caller memoization and no-opt-in laziness while making the old function-as-body mistake a type error. The follow-up cancellation change also avoids issuing a guaranteed-rejected cancel against a body locked by the in-flight preflight; the settled selected body remains the cleanup owner.

I am leaving the prior requested-changes state in place until replacement exact-head CI completes. This branch also inherits the current `dev` failures now being repaired in #5338, so after #5338 lands it must be refreshed and rerun before approval or merge. No additional code blocker found in the follow-up diff.

PR #5345, opened 2026-09-20
fix(desktop,release): sign the WidgetKit extension with the Developer ID

lidge-jun:
## Summary

The macOS app would have installed with no widget, and nothing in the build would have said so.

`macOS.files` in `desktop/src-tauri/tauri.conf.json` puts `PlugIns/OpenCodexWidget.appex` into the bundle. The Tauri bundler copies it and never signs it: `copy_custom_files_to_bundle` in tauri-bundler 2.5.0 writes the file and does not add it to `sign_paths`, which only ever holds `Contents/MacOS`, `Contents/Frameworks` and the `.app` itself, and there is no `--deep` anywhere in that path. Whatever signature `build-widget.sh` leaves is therefore the signature that ships.

It left an ad-hoc one. The script's signing branch keys on `MACOS_SIGN_IDENTITY`, and the release workflow set that variable only on `Build desktop bundles` — the step *after* the widget is built. `Build WidgetKit extension` carried no `env:` block at all, so the script always took its `codesign --force --sign -` fallback, with `--timestamp=none` and no hardened runtime. macOS does not register an extension signed that way, and notarization rejects any Mach-O in a bundle that lacks the hardened runtime.

**This has not hurt anyone yet.** No release has shipped a macOS app — v2.58, v2.59 and v2.60 all carry zero desktop assets — and the `APPLE_*` secrets were added to the repository hours after the last release ran, so the signed branch of this script has never executed. The next release is where it would have landed.

### The fix

The script resolves its signing identity **before** the Swift build. A run that holds Developer ID material and somehow has no identity now fails in a second instead of after a universal build, and never leaves a half-built unsigned appex behind — the previous ordering did exactly that when the refusal fired. `WIDGET_SIGN_REQUIRED=1` makes refusal the behaviour whenever the workflow holds a certificate; the ad-hoc branch stays for local builds and now carries the hardened runtime too, so the two paths differ only in who signed.

Signing walks **every Mach-O the bundle actually contains, chosen by magic bytes rather than by name.** Today that set is one file. A suffix filter is the thing that fails silently when that stops being true: a helper tool or an embedded dylib carries no extension to match, stays unsigned, and the submission comes back `The binary is not signed with a valid Developer ID certificate` while the containing bundle looks perfectly signed. Every signature carries `--options runtime`, and the script then re-reads its own result and fails if the runtime flag is missing.

That self-check captures codesign output instead of piping it into a matcher. Under `set -o pipefail`, a matcher that exits on its first hit kills codesign with SIGPIPE, and the check fails on exactly the signatures it was written to accept — which it did on the first draft of this change.

The workflow imports the certificate into a temporary keychain before the widget is built, because codesign resolves an identity through the keychain search list and Tauri does not build its own keychain until the bundling step. Tauri reads the current search list and re-adds itself to it, so the two keychains do not collide, and a cleanup step deletes ours on any outcome.

## Verification

Run on macOS with Xcode 27.0 and a real Developer ID Application identity in the keychain.

- **Signed path:** `flags=0x10000(runtime)`, `Authority=Developer ID Application`, `TeamIdentifier` set, secure timestamp present, `com.apple.security.app-sandbox` preserved, `codesign --verify --deep --strict` clean. `CFBundleShortVersionString` and `CFBundleVersion` both resolve to the Tauri version.
- **Ad-hoc path, no identity:** `flags=0x10002(adhoc,runtime)`. The local build gains the hardened runtime it was missing.
- **`WIDGET_SIGN_REQUIRED=1` with no identity:** refuses in under a second, before the build.
- **Bundle simulation.** An `.app` holding the signed appex under `Contents/PlugIns`, signed the way Tauri signs it — inner executables first, then the bundle, no `--deep` — keeps the nested Developer ID signature, runtime flag, team identifier and entitlements intact, and `codesign --verify --deep --strict` reports `--validated:...OpenCodexWidget.appex` with the app satisfying its designated requirement. The same simulation over an appex left unsigned fails outer signing with `In subcomponent: .../OpenCodexWidget.appex`, which is what the old refusal path would have produced.
- **`tests/ci-workflows/release-desktop-scripts.test.ts`** — 11 pass locally. New cases bind the workflow wiring, the import-before-build ordering, the cleanup on any outcome, the magic-byte sweep, the hardened runtime and the script's own check of it, and the refusal.
- **Not run:** end-to-end notarization of a complete OpenCodex `.app`. That needs a full `tauri build` with the release credentials, and the release workflow is where it belongs. Exact-head hosted CI covers the rest.

## Checklist

- [x] Targets `dev`
- [x] Behaviour change carries a focused regression test beside the existing desktop release script tests
- [x] Local verification on real Developer ID material, both signed and ad-hoc paths, plus the refusal
- [x] No secret is echoed; the decoded `.p12` is removed immediately after import and the keychain is deleted with `if: always()`
- [ ] **Security review required.** This touches release automation, credential handling and GitHub Actions workflow permissions, which `AGENTS.md` and `MAINTAINERS.md` place under explicit security review. Please read the keychain import step directly rather than approving on the summary.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved macOS widget signing for release builds using configured Developer ID certificates.
  - Release builds now reject unsigned or ad-hoc-signed widgets when proper signing is required.
  - Widget binaries are signed with hardened runtime protections and verified before packaging.
  - Temporary signing credentials are securely cleaned up after the build.

- **Development**
  - Local macOS builds continue to support ad-hoc signing when release signing is not configured.


lidge-jun:
## 리뷰 · 우선순위 67 / 80

이 PR은 macOS 데스크톱 앱에 들어가는 WidgetKit 확장(appex)이 잘못된 서명으로 나갈 뻔한 구멍을 막습니다. Tauri는 `PlugIns/OpenCodexWidget.appex`를 번들에 복사만 하고 다시 서명하지 않습니다. 그래서 `build-widget.sh`가 남긴 서명이 그대로 출고됩니다. 그런데 릴리스 워크플로는 위젯을 만드는 단계에 서명 환경이 없었고, 그다음 단계에만 `MACOS_SIGN_IDENTITY`가 있었습니다. 스크립트는 항상 임시(ad-hoc) 서명으로 갔고, hardened runtime도 없었습니다. 그러면 macOS가 확장을 등록하지 않고, 공증도 거절합니다. 아직 macOS 앱이 릴리스로 나가지 않아서 피해는 없지만, 다음 릴리스에서 터질 문제였습니다. 고친 내용은 세 갈래입니다. 스크립트는 빌드 전에 서명 이름을 정하고, 인증서가 있는 릴리스에서는 임시 서명을 거절합니다. 번들 안의 Mach-O를 이름 말고 파일 종류로 찾아 안쪽부터 서명하고, 마지막에 번들에 entitlements를 붙인 뒤 runtime 플래그가 있는지 스스로 확인합니다. 워크플로는 위젯 빌드 전에 임시 키체인에 인증서를 넣고, 끝난 뒤(실패해도) 지웁니다. 베이스는 `dev`이고, 워크플로·스크립트 계약을 잠그는 테스트가 같이 옵니다. types.ts/config.ts 분할·중복 tip과는 무관합니다.

라인 - `desktop/scripts/build-widget.sh`의 `find` 순회: 지금은 Mach-O가 하나라 문제 없습니다. 나중에 헬퍼·dylib가 늘면 `find` 기본 순서는 “안쪽 → 바깥”을 보장하지 않습니다. 그때는 깊이 순으로 정렬하거나 명시적 목록이 필요합니다.
라인 - 같은 파일의 `flags=…runtime` 검사: `codesign --display` 문장에 기대 문자열이 있는지만 봅니다. 지금은 맞고, display 형식이 바뀌면 통과/실패가 어긋날 수 있습니다.
라인 - `.github/workflows/release.yml` 키체인 정리: `security delete-keychain … || true`라서 삭제 실패가 조용히 넘어갑니다. GitHub 러너는 보통 일회성이라 실무 위험은 작지만, 실패를 로그에라도 남길지는 선택입니다.
라인 - 전체 OpenCodex `.app` 공증(notarization) e2e는 이 PR에서 돌리지 않았습니다. appex 단위·시뮬레이션·스크립트 테스트는 있고, “Tauri가 서명한 완성 앱 + 공증”은 다음 릴리스 파이프라인이 첫 실전입니다.

메인테이너의 판단이 필요한 지점

작성자가 체크리스트에 적어 둔 대로, 이 PR은 릴리스 자동화·인증서·키체인·Actions 권한을 건드립니다. 요약만 보고 승인하지 말고 `Import the Apple signing certificate for the widget` 스텝을 사람이 직접 읽어 주세요. 임시 키체인·p12 즉시 삭제·`always()` 정리는 방향이 맞아 보이지만, 보안 리뷰 게이트를 통과시킬지는 메인테이너 결정입니다.

완성 `.app` 공증을 이 PR에 억지로 넣을지, 아니면 다음 서명 릴리스 워크플로에서 확인할지. 본문은 후자를 택했습니다.

너의 추천

방향이 분명하고, “위젯만 임시 서명으로 조용히 나가는” 실패 모드를 막는 고칩니다. 워크플로 배선·거절·Mach-O 스윕·runtime 자가 검사를 잠근 테스트도 핵심을 잘 덮습니다. 보안 리뷰만 통과하면 `dev`에 머지해도 됩니다. 머지 후 첫 서명 macOS 릴리스에서 appex가 Developer ID·runtime·팀 ID를 유지하는지 한 번만 확인하면 됩니다. 닫을 중복·무효 tip은 없습니다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
Requesting changes after explicit release/workflow security review of exact head `e9ded746e0b1e606adbb9684444d6d81d831c1ae`.

The signing direction and the fail-closed widget contract are valuable, and the diff does not broaden triggers, permissions, action pins, or release eligibility. Three credential-handling blockers remain in `.github/workflows/release.yml` / `desktop/scripts/build-widget.sh`:

1. The macOS job decodes the certificate with `base64 --decode`. The BSD `/usr/bin/base64` on GitHub macOS runners uses `-D`; the long GNU option is not portable and can fail before the signing identity is imported. Use the existing repository-compatible macOS decode form.
2. Cleanup is registered too late. `WIDGET_SIGNING_KEYCHAIN` is appended to `GITHUB_ENV` only after decode, keychain creation, import, partition-list setup, and search-list mutation all succeed. Any earlier failure leaves the temporary keychain and/or decoded `.p12` outside the `always()` cleanup path until runner teardown. Publish both cleanup paths before the first fallible operation, have the cleanup step remove the certificate as well as the keychain, and cover failure-before-import in the workflow contract test.
3. On a runtime-flag verification failure, `build-widget.sh` prints the complete `codesign --display --verbose=4` output. That output includes Developer ID authority and `TeamIdentifier`, exposing an Apple account identifier in public Actions logs. Keep the raw description in memory for the assertion, but emit only a generic or narrowly filtered diagnostic on failure.

Please fix all three on a replacement head and rerun exact-head hosted CI. The inside-out signing logic, hardened-runtime requirement, ad-hoc local fallback, and release refusal otherwise look sound.

PR #5347, opened 2026-09-20
Derive the four telemetry pull requests from the landed recorder

lidge-jun:
## Summary

The four telemetry pull requests lane C deferred "as implemented", rebuilt as consumers of the
recorder that #5266 and #5300 landed. Reimplements #2366, #3748 and #3983; carries the size-limit
half of #5063. No new store: the durable shapes stay `PersistedUsageAttempt` and
`PersistedUsageEntry`, and every projection reads them.

**Why a failed request is now attributable (#2366).** `failureStage` and `failureCause` ride the
attempt that ended a request and the logical row, both closed roster members. The PR's own
`FailureSide` and seven-member `FailureStage` are not here — two attribution vocabularies for one
question is the class that blocked 2.60.0 — and its widening of `transportPhase` and
`terminalSource` to arbitrary strings is not here either. `terminalStatus`, which was a plain
`string`, joins the closed validators beside it, because it is now a grouping-key slot assembled
from an upstream frame. The derivation reads only closed values; `errorCode` and `upstreamError` are
excluded because both carry upstream text, so a classification keyed on them is a different answer
per provider and per locale. That exclusion is what lets the pair be a Prometheus label and a
fingerprint component with no masking pass. The resend verdict they imply is **never stored** —
`/api/logs` computes it at read time, so a row written by an older build cannot assert a permission
the current tables refuse.

**Why recurring failures group without a second ledger (#3748).** `src/telemetry/` and its SQLite
store are not built. Failed rows are folded during the existing cooperative ledger scan into a
versioned fingerprint over a fixed-arity tuple of closed members, inheriting that scanner's 1 MiB
row ceiling, chunk size, checkpoint identity and boundary digest. The free-text signature and its
regex masking are replaced by construction rather than by a better regex: an expression can only
assert it removed what it matched, while a tuple whose every slot comes from a frozen list has
nothing to remove. Read with `GET /api/usage?failures=1`, opt-in because it costs a scan.

**Why stream diagnostics need no second emission path (#3983).** `emitDebugLine` writes the ring
**and** stderr, and a service manager redirects stderr to a file, so the PR's per-event lines would
give an installed service a durable per-event history beside the ledger; its per-payload HMAC used a
process-global key, making every repeated prompt fragment and tool name correlatable for the process
lifetime. Five bounded counts on the attempt answer the same questions and cannot carry content.
Adapter events are counted at the adapter-parse seam and relayed frames after a successful enqueue —
counting both at the reader would make them equal by construction and erase the loss signal. The
debug ring now formats one line per finalized attempt from those counts.

**Why retention can delete rows safely (#5063).** `usageLedgerMaxBytes` is unset by default and
unset means unlimited. The defect this closes: #5063 captured a size, copied a suffix and renamed
over whatever was there, so a row appended in between was silently dropped. Now the compaction runs
inside the synchronous append call stack, a second server on the same home cannot append at all
because the existing ledger-owner lease refuses it, and `validateBeforeRename` re-opens the target
immediately before the rename and refuses unless identity, size and revision metadata are
byte-for-byte what was copied. Rows are copied and never parsed, which is what keeps a field a newer
build wrote intact through a compaction. A compaction also discards the Logs ring, the retained
aggregates and the request-history index — otherwise `/api/logs` keeps serving rows the ledger no
longer has.

Before/after, concretely: a turn whose upstream stream died mid-flight was a green row with a bare
`upstream_server_error`; it is now `transport-ambiguous` at `protocol-prelude`, reported identically
by the dashboard, the durable row and `opencodex_request_failures_total`, with a resend verdict that
names which refusal it is.

Closes #2366
Closes #3983

#3748's grouping is built and readable, but its operator surface is one query parameter rather than
a page; the coordinator decides whether that closes it.

**#5063 is deliberately not closed.** Its Usage-page control is not here. This lane may not build or
run the GUI, so it cannot produce the screenshot that gate requires, and shipping an unverifiable
control is worse than shipping the policy it would set. Remaining scope: the dashboard control, its
management route, and the ten catalog strings. The limit is settable in `config.json` today and the
configuration reference documents it.

## Verification

Static source review plus exact-head hosted CI, and four high-effort adversarial reviews — one per
carried pull request before implementation, then three over the finished branch covering typecheck
hazards, repository gates, and runtime correctness and privacy. Their findings are in the branch:
transport evidence now outranks a synthetic status, 402 maps to `quota-exhausted`,
`transport-unsent` is no longer the fall-through (it permits a resend, so it is reachable only from
a site that can prove a pre-connect failure), the streamed writer fsyncs the parent directory,
buffered responses no longer read as total relay loss, two tests read rosters instead of restating
them, and the invariant is split so each of INV-RESEND-01 and INV-ATTRIBUTION-01 binds exactly one
test.

Checked statically on this branch:

- no touched file is at or over its file-size ratchet cap; `src/server/index.ts` sits at 884 against
  893, and the ten locale catalogs are exempt;
- `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json` agree key for key,
  and each new test's regex seed resolves to the domain it is registered to;
- `src/usage/telemetry-contract.ts` still has no imports, `src/usage/request-outcome.ts` still reaches
  nothing but it, and `gui/src/pages/Logs.tsx` still never names `src/usage/log`;
- all 29 new label keys exist in all ten catalogs and those edits are purely additive;
- no test or document restates a source constant.

**NOT RUN on this branch, by instruction, and not to be recorded as passing:** `bun run test`, any
individual `bun test` file, `bun run typecheck`, `bun run build:gui`, `bun run lint:gui`,
`bun install`, `bun run structure:check`, `bun run privacy:scan`, and any live `ocx` execution.

### The GUI screenshot gate

`missing_ui_screenshot` fires, and it is a fair gate here: this adds UI, not only strings. What
follows is the evidence for a judgement, not an argument for a waiver.

**What is new on screen.** Three rows in the Logs detail dialog's Basic section, and one changed
value in the attempt table:

| Where | Before | After |
| --- | --- | --- |
| Detail → Basic | Time, Outcome, Upstream sends, Request, … | adds **Failure cause** (with the stage it reached, inline in parentheses) and **Resend**, both only when the row failed |
| Detail → Attempts, "Reason" column | the raw wire `errorCode`, e.g. `upstream_server_error` | the localized cause, e.g. "Upstream fault"; `errorCode` remains the next fallback |

**Why the layout risk is low, stated precisely.** Both new rows are label/value pairs inside the
existing `.log-detail-grid`, which is a two-column `max-content minmax(0, 1fr)` grid that already
renders Time, Outcome and Upstream sends the same way. No CSS is added or changed, no new class
appears, no column is added to any table, and nothing becomes conditional that was not already —
the adjacent Upstream sends row uses the identical `{cond && (<>…</>)}` pattern. The only genuinely
new visual is a longer value string in the second grid column, which that column already wraps via
`minmax(0, 1fr)`.

**What a screenshot would still add.** Text length. The longest English value is "Not resent: the
same request would fail again", and German and Vietnamese are longer still; at the 7rem label
column the narrow breakpoint uses, that wraps. Wrapping is the expected behaviour of this grid, but
"expected" is a source claim, and only a capture shows whether it reads well next to the row above
it.

**What source checking already settles.** The ten catalog edits are +29 lines and 0 removed in each
file, every one of the 29 keys exists in all ten catalogs, and three `satisfies` clauses make a
missing label a typecheck failure rather than the silent fallback that previously rendered four real
recovery kinds as "Unknown recovery reason".

### Two limits this branch does not close

Only the attempt that ends a request, plus the one sealed by a key-account rotation, carry
attribution; six intermediate finalizers in `policy-fallback.ts` and `core-combo.ts` still reach the
ledger unattributed. Each has different evidence in scope, and a branch verified by static review
alone should not add six classification call sites at once. The logical row is attributed in every
case, which is what the projection and the exporter read.

A ciphertext or reasoning-parameter recovery that succeeded, followed by an unrelated 400 on the same
attempt, still reads as that recovery's cause. The rule is narrowed to the last recorded kind on the
matching status; the proper fix belongs in the recovery path.

Both are recorded in `devlog/_plan/260920_round2_followups/050_lane_r5.md`, along with a pre-existing
defect found while reviewing the atomic writer: its scrub fallback opens with `"wx"` and so always
fails on an existing temp. That one predates this branch and sits on a security-adjacent path, so it
is left for its own change rather than fixed inside a telemetry branch.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Request logs now show failure causes, processing stages, and whether a request can be safely resent.
  * Usage reporting can group failed requests by recurring failure patterns through the optional failures view.
  * Added optional usage-history size limits that retain the newest complete records while preserving data integrity.

* **Localization**
  * Added translated failure details and resend-status labels across supported languages.

* **Documentation**
  * Documented usage-history size limits, retention behavior, minimum values, and current configuration limitations.


lidge-jun:
## 리뷰 · 우선순위 56 / 80

이 PR은 예전에 따로 열려 있던 텔레메트리 네 갈래(#2366·#3748·#3983·#5063의 크기 제한 절반)를, 이미 들어온 사용량 recorder 위에서만 다시 만든 것입니다. 새 SQLite·새 타임라인 저장소는 없습니다. 실패한 요청은 닫힌 목록의 `failureStage`/`failureCause`로만 원인을 붙이고, 같은 실패끼리 묶는 지문은 그 닫힌 칸들로만 만듭니다. 다시 보내도 되는지(`resendPermission`)는 디스크에 안 넣고 `/api/logs`가 읽을 때 계산합니다. 스트림 진단은 내용이 아닌 시도별 카운트 다섯 개로 답하고, `usageLedgerMaxBytes`가 켜져 있으면 덧붙이는 그 안에서 바이트 복사·이름 바꾸기 직전 검사로 줄입니다. 베이스는 `dev`입니다. GUI 스크린샷 게이트와 #5063 Usage 페이지 조작은 이 PR이 일부러 안 닫습니다. types/config의 `usageLedgerMaxBytes`는 추가만입니다.

라인 - `tests/usage/request-outcome-agreement.test.ts`의 `import type { AttemptRecoveryKind` 검사: `gui/src/pages/Logs.tsx`는 이제 여러 줄 `import type {` / `AttemptRecoveryKind,` 형태입니다. 한 줄 문자열과 안 맞아서 이 테스트는 지금 헤드에서 깨집니다. 본문도 로컬 테스트를 안 돌렸다고 적혀 있습니다. CI `test`가 돌면 여기서 막힐 가능성이 큽니다.
라인 - `src/server/request-log.ts` 최종 귀속: `outputObserved`는 요청 전체의 `logCtx.firstOutputMs`를 보는데, `sideEffectObserved`는 마지막 `activeAttempt.deliverySummary`만 봅니다. 앞 시도에서 도구 호출을 내보낸 뒤 넘긴 요청이면 부작용이 빠진 채 단계가 낮아지고, 다시 보내기 거절(`refused-committed`)이 약해질 수 있습니다. 출력 쪽 주석이 말한 “안전한 쪽으로 틀어라”와 같은 기준이 부작용에는 안 적용된 상태입니다.
라인 - `src/server/usage-ledger-retention.ts`의 `discardRetainedUsageAggregate()`: 주석은 유지 집계를 비운다고 하지만, 기존 함수는 가장 오래된 항목 하나만 떼는 메모리 압박용 API입니다. `discardRetainedFailureProjection()`은 통째로 null인데 집계 쪽만 다릅니다. 보통은 다음 읽기에서 다시 만들지만, 주석·동작이 어긋납니다.
라인 - `src/usage/failure-projection-cache.ts`: 줄이는 동안 돌고 있던 스캔이 끝난 뒤 옛 그룹을 다시 `retained`에 올릴 수 있습니다. 세대/에포크 가드가 없습니다. 또한 이 캐시는 앱 소유 메모리 명단에 없어, 지문이 많으면 압박 시 안 비워질 수 있습니다.
라인 - 본문이 이미 적은 한계도 그대로입니다. 중간 finalizer 여섯 곳은 시도에 원인이 안 붙고, 암호문 복구 성공 뒤 같은 시도의 다른 400은 복구 원인으로 남을 수 있습니다. 이 브랜치 범위 밖이라고 적어 둔 점은 맞습니다.

메인테이너의 판단이 필요한 지점

이 PR이 머지되면 초안 #2366·#3983은 Closes로 닫히고, #3748은 “조회 파라미터만 있고 전용 페이지는 없다”고 조율자 판단에 맡깁니다. #3748을 이걸로 닫을지, 운영자 UI를 남길지 정해 주세요. #5063은 Usage 페이지·관리 라우트·카탈로그가 없어 일부러 안 닫습니다. 동의합니다.

드래프트 #5305(`usage.jsonl` 고정 64MiB·단일 아카이브 회전)는 이 PR의 선택적 `usageLedgerMaxBytes`+개정 안전 축소와 같은 파일을 두고 싸웁니다. #5305는 대체·닫기 후보로 보입니다.

`gui/`와 로케일 열이 바뀌어 `missing_ui_screenshot`이 뜹니다. `gui-screenshot-waived` 라벨이나 메인테이너 면제 댓글이 문서상 해법입니다. 보이는 변화는 실패 상세의 원인·단계·재전송 판정 세 줄입니다.

너의 추천

한 줄 import 문자열 검사를 여러 줄 import에도 맞게 고친 뒤에만 머지하세요. 가능하면 최종 귀속의 `sideEffectObserved`도 요청 범위로 맞추세요(앞 시도 합산 또는 요청 단위 플래그). 집계 discard·프로젝션 in-flight는 주석과 맞추거나 짧게 고쳐 두면 좋습니다. 방향·프라이버시(닫힌 어휘, 원인 키에 upstream 문구 배제, 바이트 복사 축소)는 좋고, types/config 분할·중복 tip으로 이 PR을 닫을 이유는 없습니다. 머지 후 #5305는 닫거나 대체 이유를 한 줄로 남기세요.

이 댓글은 grok-bot이 작성했습니다

PR #5377, opened 2026-09-20
fix(transports): answer an ambiguous resend the same way on every HTTP surface

lidge-jun:
## Summary

- A pre-header connection loss is answered with `upstream_reset_replay_refused`: the turn may already have executed, so this proxy refuses to send it again. Two of the three HTTP surfaces said that; the translated Chat wrapper in `src/server/chat-completions.ts` did not. It preserved only the cyber-policy code and `model_not_found`, assigned the upstream code only when `classifyError` had produced none, and then attached the default `Retry-After: 2`. A client on `/v1/chat/completions` against a Responses-wire provider was therefore told the provider had throttled it, and told when to send the turn again — the exact duplicate the refusal exists to prevent.
- The verdict is now a property of the response the surfaces share instead of something each one re-derives. `retainReplayRefusal` and `carryReplayRefusal` (`src/lib/upstream-retry.ts`) restate the in-process marker at every re-wrap, including the deferred-logging wrapper in `src/server/relay.ts` that dropped it silently. All four formatters — native Chat, translated Chat, the Responses passthrough formatter and the shared bridge formatter — read the marker, or the code a body kept through an intermediate formatter, and never the status. 429 is precisely what a refusal and a real rate limit have in common, which is why reading the status is what produced this defect.
- The translated surface also restates the refusal when it arrives as a failed Responses envelope rather than a non-2xx, and an unreadable refusal body on the passthrough path now keeps the code instead of degrading to a generic `upstream_error`.

### The status decision, and why the status alone is not the fix

The status stays **429**. Codex builds its retry policy from `retry_429: false, retry_5xx: true` over four attempts, so a 5xx here multiplies the duplicate send; 429 is where that client stops.

That is not sufficient for everyone else. The Stainless-generated clients — `openai` and `anthropic`, Python and Node — decide from a status table that retries 408, 409, 429 and every 5xx, and they compute their own backoff when no wait is named. Removing `Retry-After` therefore does not stop a resend by itself; it only removes the schedule. Those clients do read one explicit signal before that table, so every surface now also emits **`x-should-retry: false`** alongside the absent `Retry-After`. Keeping 429 and adding the header covers both client families without inventing a status that describes neither failure.

Scope is deliberately narrow: the two WebSocket post-send verdicts (`upstream_no_response`, `upstream_closed_before_response`) are genuine upstream observations and keep their existing 502/504 contract.

## Verification

- **Local checks: NOT RUN.** No local suite, focused test, typecheck, build, install or proxy run was performed for this branch, by instruction. The evidence is static source review plus hosted CI on this exact head.
- Acceptance is a count rather than a shape. `tests/server/replay-refusal-parity.test.ts` starts the proxy on a real socket, counts physical upstream sends at the boundary, and drives all three surfaces with a client that implements the published SDK retry rule (an explicit `x-should-retry` first, then the 408/409/429/5xx table). Each logical request must produce exactly one upstream send and one client attempt, with status `REPLAY_REFUSED_STATUS`, code `upstream_reset_replay_refused`, no `Retry-After` and the suppression header — all derived from the source constants rather than restated as literals.
- The header name and its two values are deliberate literals inside the client double: that function stands in for the third party and has to keep believing what those clients believe even if our own constant changed.
- A rate-limit control in the same file drives the same client against an ordinary upstream 429 and asserts that it does resend. Without it, "one send" would be a property of the double rather than of the answer.
- The new file is registered in `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`; no file at its size cap grew.
- Static review covered the existing coverage for these paths (`tests/lib/upstream-retry.test.ts`, `tests/server/retry-after-429.test.ts`, `tests/providers/upstream-transient-retry.test.ts`, `tests/responses/responses-send-budget-counts.test.ts`, `tests/responses/responses-account-label.test.ts`, `tests/codex-integration/reserve-dispatch.test.ts`, `tests/usage/request-log.test.ts`): each remaining assertion holds, since ordinary and cyber-policy formatting is untouched and only a marked or allowlisted refusal enters the new branches.

## Checklist

- [x] Docs or release notes were updated when needed. `structure/transports/responses.md` records the property and the client contract; `docs-site` states the wire behaviour users see.
- [x] Security-sensitive changes were reviewed for secrets, auth, and unsafe defaults. No credential, auth or workflow path is touched; the refusal body is unchanged and still never carries the raw exception.

## Summary by CodeRabbit

- **Bug Fixes**
  - Replay refusals now return a dedicated response that prevents supported OpenAI and Anthropic SDKs from automatically resending requests.
  - Retry instructions are consistently applied across native Chat, translated Chat, and Responses APIs.
  - Ordinary rate-limit responses retain their existing retry behavior.
  - Replay-refusal details now remain intact when responses are reformatted or logged.

- **Documentation**
  - Clarified replay-refusal behavior and retry handling across supported API surfaces.

- **Tests**
  - Added coverage verifying consistent no-retry behavior across all three API surfaces.


lidge-jun:
## 리뷰 · 우선순위 74 / 80

이 PR은 “업스트림이 헤더도 받기 전에 끊겼는데, 그 턴이 이미 돌아갔을 수도 있다”는 상황에서 프록시가 다시 보내지 않기로 한 답(`upstream_reset_replay_refused`, HTTP 429)이 HTTP 표면마다 달라지던 구멍을 막습니다. 예전에 native Chat과 Responses 쪽은 거절을 제대로 말했는데, Responses 선을 거쳐 Chat으로 번역하는 `src/server/chat-completions.ts`만 달랐습니다. 그 래퍼는 사이버 정책 코드와 `model_not_found`만 지키고, `classifyError`가 이미 레이트리밋 코드를 붙인 뒤에는 업스트림 거절 코드를 살리지 못했고, 기본 `Retry-After: 2`까지 붙여 클라이언트가 “잠깐 쉬었다가 같은 턴을 다시 보내라”고 읽게 만들었습니다. 지금은 거절 판정을 상태코드에서 다시 추측하지 않고, 응답에 붙는 표시(`retainReplayRefusal` / `carryReplayRefusal`)와 본문 코드로 네 포맷터(native Chat, translated Chat, Responses passthrough, bridge formatter)가 같이 읽습니다. `relay`의 지연 로그 래퍼가 표시를 떨어뜨리던 길도 `carryReplayRefusal`로 이었습니다. `Retry-After`만 빼는 것으로는 Stainless 계열(`openai`/`anthropic`)이 429를 자기 표로 다시 보내므로, 거절마다 `x-should-retry: false`를 같이 냅니다. 수락 증거는 응답 모양이 아니라 `tests/server/replay-refusal-parity.test.ts`가 소켓 위에서 업스트림 물리 전송 횟수를 세고, SDK처럼 재시도하는 클라이언트로 세 표면이 각각 “보낸다 1번 / 클라이언트 시도 1번”인지 보는 것입니다. 진짜 레이트리밋 대조 케이스도 같은 클라이언트가 다시 보내는 것을 보여 “한 번만”이 더블 탓이 아님을 가립니다. base는 `dev`이고 types/config 분할·프리뷰 배포와는 무관합니다.

src/server/chat-completions.ts 번역 Chat 경로 - 예전에 거절을 레이트리밋+Retry-After로 바꾸던 구멍은 이번 분기와 헤더로 막혔습니다. 다만 이 헤더를 모르는 raw `fetch`·구형 SDK·자체 재시도 표는 여전히 429를 다시 보낼 수 있습니다. PR이 고른 계약(Codex는 429에서 멈추고, Stainless는 `x-should-retry`를 먼저 본다) 밖의 클라이언트는 이 패치만으로 중복 전송이 안 사라진다고 보는 게 맞습니다.

tests/server/replay-refusal-parity.test.ts - 수락 단위를 “업스트림 전송 횟수”로 잡은 방향은 버그 성격과 잘 맞습니다. 로컬 스위트는 안 돌렸고(작성자 명시), 이 리뷰 시점 hosted CI의 test shard·macos 일부는 아직 pending입니다. 머지 전에 그 초록을 확인하는 게 안전합니다.

메인테이너의 판단이 필요한 지점

거절을 계속 429로 두고 `x-should-retry: false`를 추가한 선택이 Codex(5xx면 네 번 더 보냄)와 Stainless 양쪽을 같이 잡으려는 타협입니다. `x-should-retry`를 무시하는 클라이언트를 이번 범위 밖으로 둘지, 아니면 별도 상태/문서 경고가 더 필요한지는 제품 판단입니다. WebSocket 쪽 `upstream_no_response` / `upstream_closed_before_response`의 502/504는 의도적으로 안 건드렸고, 그 분리는 타당해 보입니다.

너의 추천

핵심 수리(표시를 래핑마다 이어 주기 + 번역 Chat이 거절 코드를 살림 + `x-should-retry: false`)와 전송 횟수 패리티 테스트 방향은 맞습니다. CI test shard가 초록이면 머지해도 됩니다. types/config 중복 PR을 닫을 대상은 아닙니다.

이 댓글은 grok-bot이 작성했습니다

PR #5378, opened 2026-09-20
fix(runtime): one resend budget across composed recovery legs

lidge-jun:
## Summary

The ambiguous-resend gate already shared one counter across every leg of a logical request. Its ceiling was not shared. `claimAmbiguousResend(limit)` compared the request's spent count against whatever number the asking leg presented, and a leg reads that number from `route.provider`, which credential rotation, OAuth refresh, transport resolution and each combo target reassign inside one request. A request that had already spent the single replacement a strict provider row granted bought a second one the moment a more permissive row asked, so how many duplicate inferences one turn could produce depended on the order the rows happened to ask in. The request now keeps the smallest ceiling any leg has presented.

A second counter lived in the derivation bridge. `deriveRequestExecutionBudget` shares the parent's ledger when the parent came from the factory; when it did not, the bridge kept a local claim count, so two scopes derived from one hand-built parent, or a scope beside the parent it came from, each replaced an unknown-state send once. Pending external bookings and the durable-spend observer genuinely cannot cross that boundary because they are private to the factory, but the grant can: `claimAmbiguousResend` is public on the parent. The bridge now asks it, and fails closed when the parent does not implement it.

A zero or nonsense ceiling still refuses on its own and leaves the request's ceiling alone. That is a caller which cannot state a grant rather than an operator narrowing this request, and a leg with no policy is refused as `ambiguous-no-policy` before it ever reaches the claim.

### What the new coverage observes

`tests/lib/ambiguous-resend-composition.test.ts` walks one logical request through the composition the existing tests cannot see: first send, a reset before any head, the authorised replacement, a disconnect after the head carrying only control events, a combo candidate's derived scope, a credential-refresh replay, the 429 leg, and finally the single final-recovery reserve moving to a row the operator tuned higher.

It counts two numbers apart. One is every send that physically left. The other is the subset that carried a turn the origin may already have run, taken from the gate's own verdict rather than inferred. Four sends is within budget whether one of them replaces a possibly-executed turn or three of them do, which is why a single send count cannot hold this property. The chain ends at the request's send ceiling with the grant spent exactly once, and the two counts are reconciled against the request's own ledger so a leg that books a send it never made shows up here rather than as a ceiling that quietly stopped firing.

Values are derived from `resetReplayPolicyFor` and the budget policy rather than written as literals, so the test states the relationship instead of today's numbers. A source scan over `src/` also asserts that every `refetchAfterProtocolSafeReset` call site passes `authorize`, since that option is optional on the helper and an omitted one is a post-header replacement no gate ever weighed.

### One existing assertion changes

`tests/lib/execution-budget-permits.test.ts` asserted that raising the ceiling released exactly the difference. That is the widening described above, written down as intended behaviour, and it now expects a refusal. The nonsense-ceiling case beside it holds as written.

### Not closed by this pull request

Issue #4191 is untouched. It is about an established WebSocket dying mid-turn with no SSE fallback and the response-prelude timeout, none of which this change reaches; a resend budget bounds how often a turn may be replaced, not whether a dead relay can fall back to another transport.

Issue #5180 is untouched. It reports a provider 429 exhausting retries on long turns and asks for cooldown and `Retry-After` awareness. The 429 leg appears here only as one participant in the composition, and its retry and cooldown policy is unchanged.

Two further resends of an ambiguous failure bypass this gate by design and were left alone as out of scope: the destination-scoped `replaySafe` initial leg, which is already pinned to exactly one leg by an existing source test, and the adapter-owned reset ladders, which assert replay safety inside the adapter rather than through the stage table. Both remain bounded by the request's physical-send budget.

## Verification

local checks: NOT RUN. No test suite, focused test, typecheck, build, install, proxy or service was run for this branch; the lane is hosted-CI-only. Verification was static: the composition was hand-simulated send by send against `fetchWithResetRetry`, `refetchAfterProtocolSafeReset`, `reserveDispatch` and the `used` setter, including the interaction between an externally counted reservation and its later report, and the whole tree was searched for other holders of the old counter and for assertions depending on the previous ceiling behaviour.

Hosted CI on the exact head is the acceptance evidence.

The new test file is registered in both `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`, and no touched file is tracked by the file-size ratchet.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Enforced a single ambiguous-resend allowance across all recovery, refresh, rotation, transport, and combination steps within one request.
  - The strictest resend limit presented during a request now applies throughout; later, more permissive limits cannot grant additional replacements.
  - Prevented independent or manually created budget views from issuing duplicate resend allowances.
- **Documentation**
  - Clarified that resend limits apply per request rather than per recovery step.
- **Tests**
  - Added comprehensive coverage for combined recovery scenarios and shared resend limits.


lidge-jun:
## 리뷰 · 우선순위 71 / 80

이 PR은 한 번의 논리 요청 안에서 “모호한 실패(ambiguous failure) 때 한 번 더 보내도 되나?”라는 허가를 다리마다 따로 세던 구멍을 막습니다. 예전에는 허가를 쓴 횟수(카운터)는 요청 하나에서 같이 썼는데, 상한(ceiling)은 지금 물어본 다리가 들고 온 숫자를 그대로 봤습니다. 그 숫자는 `route.provider`에서 오고, 자격 증명 회전·OAuth 갱신·전송 방식 결정·콤보 대상 전환이 요청 도중에 그 행을 바꿉니다. 그래서 엄한 행이 먼저 “1번만”이라고 한 뒤 허가를 썼는데, 나중에 느슨한 행이 “2번까지”라고 물으면 같은 턴을 한 번 더 중복 추론할 수 있었습니다. 지금은 요청이 지금까지 본 상한 중 가장 작은 값을 붙잡고, 나중에 더 큰 숫자가 와도 허가를 더 주지 않습니다. 또 손으로 만든(브리지) 부모에서 파생한 스코프가 자기만의 카운터를 갖던 길도, 부모의 `claimAmbiguousResend`를 물어보고 없으면 거절하도록 바꿨습니다. 새 `ambiguous-resend-composition.test.ts`가 리셋·헤더 이후 끊김·콤보 파생·자격 증명 갱신·429·최종 복구 예비분까지 한 요청으로 이어서, 물리 전송 횟수와 중복 가능 전송 횟수를 따로 셉니다. base는 `dev`이고 types/config 분할·프리뷰 배포와는 무관합니다.

structure/overview.md INV-RESEND-02 - “Enforced by”가 여전히 `tests/lib/ambiguous-resend-gate.test.ts`만 가리킵니다. 그 파일은 게이트를 한 질문씩 묻지만, 이번에 고친 “요청 단위 최소 상한”과 “브리지가 같은 허가를 쓰는지”는 `ambiguous-resend-composition.test.ts` / `execution-budget-permits.test.ts`에 있습니다. 불변식 문서의 묶음이 옛 테스트만 가리키면, 그 테스트만 초록이어도 천장 회귀를 놓칩니다. Enforced by에 새 합성 테스트를 같이 적거나, 천장 케이스를 이름 붙인 게이트 테스트로 옮기세요.

src/lib/request-execution-budget.ts `createAmbiguousResendGrant` - 상한은 “지금까지 물어본 다리”의 최솟값입니다. 느슨한 행이 먼저 와서 허가를 다 쓴 뒤에 엄한 행이 오면, 이미 나간 중복은 되돌리지 않습니다. 고친 버그(엄함→느슨함으로 상한이 열리는 순서)와는 반대 방향입니다. 운영에서 failover가 더 엄한 행으로만 가는 경우가 있으면, “요청 전체의 진짜 최솟값”과는 어긋날 수 있습니다.

메인테이너의 판단이 필요한 지점

로컬 검증은 안 돌렸고 hosted CI만 증거로 두었습니다. 이 리뷰 시점엔 테스트·gates 다수가 아직 pending입니다. “느슨함 먼저 → 엄함 나중” 잔여 순서를 이번 범위에서 받아들일지, 아니면 요청 시작 행의 상한만 고정하는 식으로 더 좁힐지도 제품 판단입니다. Issue #4191·#5180은 의도적으로 안 닫는다고 했고, 그 범위 분리는 타당해 보입니다.

너의 추천

천장 동작의 Enforced by를 새 합성 테스트와 맞춘 뒤, CI가 초록이면 머지해도 됩니다. 핵심 수리(요청 단위 최소 천장 + 브리지 fail-closed)와 합성 테스트 방향은 맞습니다. 반대 순서 잔여는 문서/후속으로 남겨도 되고, failover가 더 엄한 행으로 자주 가면 그때 이어서 좁히면 됩니다.

이 댓글은 grok-bot이 작성했습니다

PR #4161, opened 2026-09-09
fix(claude): keep mid-conversation system messages in the timeline

lidge-jun:
## Summary

Claude Code sends `role: "system"` entries inside `messages`, and every one of them was folded into `instructions`. `parseRequest` pushes `data.instructions` onto the system prompt before anything else, so each injected reminder rewrote the **head** of the prompt — which is exactly the span an upstream KV cache matches on. Every mid-conversation reminder therefore invalidated the cache prefix. When the client sends no `metadata.user_id`, the Desktop `prompt_cache_key` fallback hashes the same system parts, so the cache key rotated along with the prefix and the request could not even land on the same routing cohort.

Each non-empty in-messages system message now becomes a chronological input item:

```json
{ "type": "message", "role": "developer", "content": [{ "type": "input_text", "text": "..." }] }
```

Top-level Anthropic `system` keeps flowing through `systemToInstructions` into `body.instructions` unchanged. `instructions` now has exactly one source and stops moving mid-conversation.

**Why `developer` and not `system`.** The schema admits a `system` role in `input`, but ChatGPT Codex rejects it live, `parseRequest` re-hoists it onto the system prompt, and the canonical Responses forward folds text-only system items back into `instructions` — so it would land back where it started. `developer` is first-class in the schema, survives `parseRequest` as an ordinary chronological message, and preserves timeline order.

**The original fold was deliberate**, not an oversight: `cee918ce3` introduced it so the native ChatGPT backend would not 400 on a system input item. That constraint is real and is still respected here — it rules out `system`, not `developer`. What turned out to be stale is the accompanying 2026-07-11 note that folding is "the only shape that works on every route".

### A scope decision a reviewer may want to object to

**All** in-messages system messages become developer items, not only the ones after the first user turn.

The narrower alternative — hoist a leading system message, keep only mid-conversation ones in the timeline — would have left the old contract test untouched, but it does not close the issue. A client that injects a fresh leading system message on every turn still mutates `instructions` every turn, and that is the reported failure. Taking the narrow option would have meant shipping something that looks like a fix and still rotates the prefix for the reporter.

The cost is that privileged system text now arrives as developer text. On Anthropic and Google outbound, developer items are presented as chronological `user` messages, so there is real semantic drift from "this is a system instruction" to "this is part of the conversation". That is prefix-stable, which is the property this change is buying.

`systemParts` deliberately keeps its **array** shape inside the cache-key hash. Joining it to a string would rotate every existing cohort key for a reason unrelated to this bug.

[...]

Issue #4148, opened 2026-09-09
[Bug]: Mid-conversation role: system messages in Claude inbound are aggregated into top-level instructions, destroying prompt cache prefix


richardfeiliu-a11y:
### Client or integration

Claude Code

### Area

Proxy and routing

### Summary

In `src/claude/inbound.ts:334-348`, OpenCodex translates incoming Claude Messages by iterating through `raw.messages`. When it encounters `msg.role === "system"`, it appends the message text to `systemParts`, which is then joined and assigned to `body.instructions`:

```typescript
// src/claude/inbound.ts:334-348
else if (msg.role === "system") {
  const text = systemMessageText(msg.content);
  if (text.length > 0) systemParts.push(text);
}
...
if (systemParts.length > 0) body.instructions = systemParts.join("\n\n");
```

When clients like Claude Desktop or Claude Code inject dynamic system messages during a multi-turn conversation (e.g., style reminders, date changes, or subagent notices):
- Turn 1: `instructions` = `[topLevelSystem + reminder_1]`
- Turn 2: `instructions` = `[topLevelSystem + reminder_1 + reminder_2]`

Because `instructions` is placed at the very head of the LLM prompt sequence, mutating `instructions` on every turn completely invalidates the prompt cache prefix (KV cache) for all subsequent user/assistant/tool message history.

### Real-world Impact & Production Evidence

In real-world long-running Claude Desktop sessions (e.g. 100+ turns), this defect causes prompt cache reuse to freeze at the initial system prompt boundary:
- **Observed historical degradation:** `cachedInputTokens` remained strictly frozen at **24,576 tokens** across dozens of turns while total input expanded from 60,000 to 137,000 tokens. The cache hit rate collapsed from **40.7% down to 19.4%**, resulting in >110,000 un-cached tokens billed at full price on every single turn.
- **Controlled mitigation proof:** By suppressing dynamic mid-conversation reminders on the client side (`outputStyle: default` and `totalTokensReminder: off`), `instructions` length growth across turns became exactly **0**. In the exact same Claude Desktop 3P harness over 10 turns (70k-92k tokens), `cachedInputTokens` successfully grew with the conversation (51k -> 70k -> 79k), achieving **96.24% ~ 99.93% cache hit rate**.

### Reproduction

1. Start OpenCodex on port 10100.
[...]

PR #5381, opened 2026-09-20
fix(chat): keep a mid-conversation instruction in its slot on the translated Chat inbound

lidge-jun:
## Summary

- A Chat Completions request that carries an instruction after the conversation has started lost where it was written. `src/chat/inbound.ts` routed every `system` and `developer` message into `systemParts` and joined them into `body.instructions`, so `U1 -> A1 -> D2 -> U2` reached the router as instructions plus a three-message input. The position was gone before any adapter ran, and the outbound adapter that has kept that slot since #4161 had nothing left to preserve.
- This is not a rare internal path. An ordinary `openai-chat` route sends straight to the Chat wire and forwards the caller's `messages` untouched, while a combo, a policy route, a synthetic effort row or a preprocessing route enters the Responses pipeline and translates. The same transcript therefore reached a destination in two different orders depending on whether a routing feature was on.
- Past the first turn the message now becomes the input item the rest of the pipeline already reads: `{ type: "message", role: "developer", content: [input_text] }`, the representation `src/claude/inbound.ts` has minted for the same shape since #4148. A leading block still folds into `body.instructions`, so the ordinary prompt head and its prefix-cache behaviour are unchanged.
- The role is `developer` rather than `system` on purpose: the native ChatGPT backend refuses a `system` item inside `input`, and canonical forwarding folds a message-shaped `system` item back onto `instructions` (`src/adapters/openai-responses/canonical-forward.ts`), which would undo the placement one hop later.
- An instruction that arrives between a tool call and its result waits for the batch to drain instead of splitting the pair. That guard is not decorative: `src/adapters/kiro/payload.ts` refuses an interrupted pair, and the Anthropic and Google mappers synthesize a missing result when a call and its output stop being adjacent. The held text rejoins the timeline when the batch closes or at the next user or assistant turn, so its order relative to the conversation never changes.
- Scope is the inbound only. `src/adapters/openai-chat/passthrough.ts` and the role question belong to the neighbouring change in the same unit, and `structure/providers/chat-compat.md` and the `docs-site` provider reference are left to the documentation lane that reconciles them.

## Verification

- local checks: NOT RUN. This branch was produced under a no-local-execution constraint, so no test, typecheck, build, install or proxy run happened locally. Verification here is static source reading plus hosted CI on this exact head.
- New coverage in `tests/responses/chat-inbound-developer-position.test.ts`. The translator cases hold the split this change introduces: a leading block still becomes `instructions`, a later one becomes a chronological item, a mixed transcript puts each in a different place, an instruction inside a tool batch lands after the batch rather than between a call and its result, an abandoned batch still releases before the next turn, and an empty instruction produces no item at all.
- The expected item for a mid-conversation `system` message is derived from `src/claude/inbound.ts` by translating the same transcript through that inbound and comparing the items, rather than restated as a literal, so the two inbounds cannot drift apart silently.
- The cross-path cases send one transcript three ways against one recording upstream and read the final wire body: the native Chat route, a combo route that forces translation, and the Responses endpoint carrying the same input. All three must equal the transcript the caller sent. A case that hands an already-translated object to the adapter cannot see this defect, because the position was gone before the adapter ran.
- The new file is a sibling of the existing Chat inbound coverage rather than an addition to it, and is registered in `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`. No file this branch touches carries a `tests/fixtures/file-size-baseline.json` cap.
- Static review of the downstream adapters confirmed that a text-only chronological `developer` message is preserved in place by the OpenAI Chat mapper, mapped in place by the Anthropic, Google, Cursor, Kiro, Devin, Ollama and command-code mappers, and passed through untouched by the native Responses passthrough. The one behaviour worth naming is that the CodeBuddy and Qoder projection hoists any developer message into its appended system prompt, which is that adapter's existing contract and applies equally to the items `src/claude/inbound.ts` already produces.

## Checklist

## Lane and integration notes

- Owned files for this lane are `src/chat/inbound.ts`, the Chat path in `src/responses/parser.ts`, and this lane's tests (`devlog/_plan/260921_cross_path_contract/010_lane_boundaries.md`). `src/responses/parser.ts` needed nothing: it already carries a `developer` input item as a chronological conversation message, which is why reusing that representation is the whole fix.
- One file outside every lane row is touched and is called out here as that document requires: `structure/data-planes/inbound-compat.md`. It describes the Chat inbound and said nothing about which of the two homes an instruction reaches on the translated bridge. `structure/providers/chat-compat.md` and the `docs-site` reference are untouched; they belong to the documentation lane, and the paragraph added here links the contract that lane owns rather than restating it.
- The branch is rebased onto `dev` at `a6e8df4ec1`, so it carries the native developer-role change and the developer-role policy documents that landed before it. With `foldDeveloperRoleToSystem: false`, `applyExplicitChatDeveloperRole` returns the caller's array unchanged and the translated adapter emits `developer`, which is why the cross-path comparison holds on the integrated tree rather than only on the branch it was written against.
- The two registration entries are appended at the tail of `scripts/test-layout/layout.json` and `tests/fixtures/test-layout-expected.json`. The rebase resolution is exactly the `dev` map plus this one entry, verified by comparing the parsed maps rather than the diff.

## Hosted CI

- Cross-platform CI is green on this exact head: every job success or skipped, aggregate `ci: success`, including all four test shards, `gates`, `structure gate`, `storage policy`, `api usage`, `docker smoke`, the three `npm-global` smokes, the three `keyring` jobs, both macOS shards and `desktop shell`. React Doctor is green on the same head.

## Summary by CodeRabbit

- **Bug Fixes**
  - Chat-to-Responses translation now preserves the chronological placement of mid-conversation developer instructions.
  - Leading system and developer messages continue to appear as request instructions.
  - Instructions interleaved with tool calls are deferred until the tool exchange completes, preserving valid tool/result ordering.
  - Behavior is consistent across native Chat, failover, and Responses routes.

- **Documentation**
  - Updated compatibility guidance to describe instruction placement and tool-call handling.


lidge-jun:
## 리뷰 · 우선순위 69 / 80

대화가 이미 시작된 뒤에 오는 system/developer 지시가, Chat → Responses 번역 길에서 맨 앞 `instructions`로 올라가 버리던 버그를 고칩니다. 예전에는 `U1 → A1 → D2 → U2`가 “맨 앞 지시 + 세 줄 대화”로 바뀌어서, 지시가 어디에 있었는지가 어댑터에 닿기 전에 사라졌습니다. 같은 대화라도 일반 openai-chat 경로는 순서를 그대로 보내고, 콤보·정책·전처리처럼 번역을 타는 경로만 순서가 달라질 수 있었습니다. 이제는 맨 앞 덩어리만 예전처럼 `instructions`에 두고, 그 이후 지시는 Claude inbound(#4148)와 같은 `{ type: "message", role: "developer", ... }` 한 줄로 타임라인에 남깁니다. 역할을 `system`이 아니라 `developer`로 둔 이유도 맞습니다. ChatGPT 백엔드는 input 안의 system을 거절하고, 정규 전달은 system을 다시 instructions로 접어 위치를 다시 깨뜨리기 때문입니다. 도구 호출과 결과 사이에 끼는 지시는 짝이 끝날 때까지 잠시 붙잡아 두었다가, 배치가 닫히거나 다음 user/assistant 턴 앞에서 다시 붙입니다. Kiro가 끊긴 짝을 거절하고 Anthropic/Google이 빠진 결과를 만들어 넣는 이유를 본문에 적어 둔 것도 맞습니다. 범위는 inbound와 테스트·호환 문서뿐이고, base는 `dev`입니다. types.ts/config.ts 분할이나 같은 결함의 다른 열린 PR은 보이지 않습니다. 본문에 로컬 테스트는 안 돌렸다고 적혀 있고, 이 head에서 호스트 테스트 샤드도 아직 전부 끝나지 않았습니다.

라인 - `src/chat/inbound.ts` 418행 근처: “대화가 시작됐는지”를 `input.length === 0`으로만 봅니다. 빈 user/assistant는 input에 안 들어가서, `[user:"", developer:"D2", user:"U2"]`처럼 빈 턴 뒤의 지시는 또 `instructions`로 올라갑니다. CodeRabbit이 지적한 것과 같습니다. user/assistant/function/tool를 본 순간 `conversationStarted` 같은 깃발을 따로 켜고, 그걸로 leading을 가르면 됩니다. 빈 턴 + 뒤따르는 지시 회귀 테스트도 하나 있으면 됩니다.

라인 - `src/chat/inbound.ts` `function` 분기: legacy function 결과는 input에 안 넣고, 대화 시작 깃발도 안 켭니다. 흔하지는 않지만, function 줄만 지나간 뒤의 developer도 leading으로 오인할 수 있습니다. 위 깃발을 function/tool에도 같이 켜는 편이 안전합니다.

라인 - 도구 배치 안에서 보류한 지시가 여러 개면 `\n\n`로 이어 하나의 developer 줄이 됩니다. 순서는 지키지만 “지시 몇 개” 개수는 줄어듭니다. 지금 테스트는 한 줄만 봅니다.

라인 - 검증: 새 `tests/responses/chat-inbound-developer-position.test.ts`는 leading/나중 지시 분리, Claude 아이템 동형, 도구 배치 보류·포기 배치 해제, Chat/combo/Responses 세 경로 와이어 일치를 잘 잡습니다. 다만 빈 턴 케이스는 없고, 작성자가 로컬을 안 돌렸으며 이 SHA의 테스트 샤드는 리뷰 시점에도 아직 pending이었습니다. 머지 전에 tip 초록을 보세요.

메인테이너의 판단이 필요한 지점

빈 턴 뒤 지시를 이번 PR에서 깃발로 막을지, 아니면 “빈 content는 대화로 안 친다”를 계약으로 두고 후속으로 둘지. CodeBuddy/Qoder가 developer를 시스템 프롬프트로 올리는 기존 동작은 본문대로 이번 범위 밖으로 둘지.

너의 추천

방향과 테스트 뼈대는 좋습니다. 머지 전에 `conversationStarted` 깃발과 빈 턴 회귀 테스트만 넣고, 호스트 CI tip이 초록인지 확인하세요. 그 한 칸만 닫히면 넣어도 됩니다. 닫을 중복 PR은 없습니다.

이 댓글은 grok-bot이 작성했습니다

PR #5384, opened 2026-09-21
Desktop shell: one exit path, one startup surface and a real tray probe

lidge-jun:
## Summary

Launching the desktop app, closing its window and quitting it were three ways of reaching the same
code, and one of them was destructive. There was no `ExitRequested` handler, so the platform's quit
gesture fell through to `RunEvent::Exit`, which called `CommandChild::kill()` on the proxy the app
had started. That is a SIGKILL on Unix: the in-flight requests, the client-configuration restore and
the state-file clearing that `ocx stop` performs were all cut off by a keystroke the user reads as
"hide". This implements D7, D2 and D6 from `devlog/_plan/260921_app_runtime_ownership/`, with
resolutions R1 and R2.

**What ends the app.** `desktop/src-tauri/src/exit.rs` intercepts the exit request and decides what
it meant. With a usable tray, closing the window and the quit gesture hide. Only the tray's Quit asks
to end, and an installed update asks for a coordinated restart (R2) rather than restarting straight
into the kill. Both hold the exit, drain the app-owned runtime through the management stop, and let
the process end or come back only afterwards. Nothing kills the child. A runtime this app did not
start is never stopped.

macOS needed more than the handler, and this is the part that looks finished while being broken:
Tauri installs a default menu whose Quit is a predefined item wired to Cocoa's `terminate:`, and the
pinned tao implements only `applicationWillTerminate`, never the cancellable
`applicationShouldTerminate`. That Cmd+Q therefore reaches `RunEvent::Exit` without ever raising
`ExitRequested`, so `prevent_exit` never sees it. `menu.rs` rebuilds the default menu with an
ordinary item on the same accelerator, keeping the clipboard items the failure diagnostic needs the
user to have.

**What proves a runtime stopped.** Only the child reporting its own exit through the spawn event
stream, or the endpoint refusing a connection. A timeout, an unauthorized reply or a body that will
not parse are not proof — reading any error as proof is how a stop that never happened gets reported
as a completed drain.

**The startup surface.** Everything used to run inside `setup()` before a window existed, and the
spawn event stream was destructured into `_events` and dropped, so a sidecar that exited immediately
presented exactly like a slow one. The window is created before anything else now — a manual launch
shows it immediately, a login launch after the tray verdict, since R1 shows it after all when there
turns out to be nowhere to hide. `startup.rs` runs
the whole sequence inside it as named states — registering, resolving, probing, attaching or
starting, waiting — under one 30-second deadline, with every probe bounded by the time remaining
rather than by the HTTP client's own four-second timeout. Registering comes first, because a login
launch starts hidden and a tray installed only after a successful start would leave a failed start
with no window and no icon. The failure state carries a retry, the child's exit code and a copyable
diagnostic naming the state, the endpoint, the configuration home and the runtime's last output. A
retry waits on a child that has not exited rather than racing it for the port. Registering happens
once per process, so a retry cannot build a second tray icon with its own refresh loop. The exit
coordinator reserves the spawn rather than holding its lock across process creation — holding it
would put spawning in front of the main thread's exit handler — and a quit arriving in between is
deferred until the child is owned and then drains it.

**What counts as a tray.** `tray_availability.rs` asks the session bus whether
`org.kde.StatusNotifierWatcher` reports a host registered. Construction success is not the question —
the pinned Linux backend creates an AppIndicator and returns `Ok` with nothing attached — and neither
is the watcher merely existing, since a watcher with no host accepts registrations and draws nothing.
Linux assumes no tray until the probe answers, and the verdict is published only once an icon
actually exists, so a tray that fails to build is a session without one rather than a claimed one.
Where there is no host, no icon is claimed, the window is shown whatever the launch origin (R1), and
closing it quits through the same drain. Tray menu handles are copied out from under the menu mutex
before any setter is called, because those setters dispatch to the main thread and the tray is built
on the main thread holding that mutex.

**One trade worth objecting to if you disagree.** A drain that has not completed within
`DRAIN_DEADLINE` is reported and the exit still proceeds, which can leave the runtime running.
Refusing to quit when the user asked is the worse answer on a window that is showing the dashboard
and has nowhere to explain itself, and a standing runtime is recoverable with `ocx stop` while a
half-restored client configuration is not. `structure/desktop-shell.md` states this plainly rather
than claiming the runtime is always confirmed gone.

**The app's half of the ownership claim.** Rebased onto the landed lane C. The shared service
install state now records who owns the running proxy, and the claim names the owning *installation*,
so the app holds an id of its own to compare against: `identity.rs` mints one into the app's config
directory, once and exclusively, because two launches racing to mint would answer to two ids and the
second would find a claim that is not its own and ask again for consent already given. An id kept
only in the shared record would be whoever wrote it last, which is why D3 accepted two records.

`ownership.rs` mirrors the claim, the three answers a read can give and `ownershipGrantedTo`, all of
which `src/service/state.ts` defines. It does not read the record — resolving one means reading
every state path and failing closed on an unreadable one, a corrupt anchor and paths that disagree,
and a weaker second implementation of a question core already answers is the mistake that gave
`discovery.rs` its own port guess. The types are the CLI's answer as it will arrive on the wire,
field for field.

**Lane seams.** Two, both named in the source. Resolution still calls `discovery::current()`, but it
runs inside the `resolving` state with a diagnostic and a retry around it, so lane A's resolve verb
replaces one call site. `ownership::resolve` is empty for the same reason, and returns *unavailable*
rather than `Recorded::None`: the shell has not been told nobody owns the runtime, it has not asked,
so no takeover is attempted and nothing is recorded. The takeover stop that ends somebody else's
*managed* runtime is D4 and is not in this diff; the management stop here only ever targets a child
this process spawned.

## What the external re-audit changed

Two independent reviews read this branch at a fixed SHA. Both said the direction was right and
neither called it shippable, on a distinction worth keeping: *better than before* and *safe in the
failure path* are different verdicts. Seven findings were in this lane's scope and all are fixed.

**The update did not actually drain first.** Removing the direct kill put the in-app update on a
coordinated path only if the coordination is reached, and on Windows it was not: the pinned
updater's install hands off to the installer process and ends this one with `process::exit(0)`, so
the restart asked for after `download_and_install()` never ran, and the package was replaced under a
runtime still serving out of those files. The order is now download and signature-check, confirm who
owns the running runtime, drain it and confirm the child is gone, then install. A drain that did not
complete refuses the install and leaves the update pending.

**A failed drain was recorded as a drain.** Both outcomes ran the same completion path. For a quit
that is a defensible trade; for a restart it is not the same judgement, because the new app comes
back attached to the old runtime while the user believes they upgraded. `DrainFailed` and
`OwnershipUnknown` are states of their own now, a quit proceeds from either, and a coordinated
restart refuses both.

**Ownership survived an attach.** It was a bool set at spawn, so a child that dies and a service that
takes the port back left the connection pointing at somebody else's runtime with the flag still set,
and Stop or Quit would send an owner's stop there. Durable consent and current process ownership are
separate now: consent stays in the recorded claim, and ownership is re-established each time from
the pid the endpoint reports. An answer that cannot be read leaves the app owning nothing.

**Tray host and registered icon were one fact.** They are three, and only the last one counts: the
verdict is published after an icon exists, and a construction failure, a main-thread delivery
failure and a lost result all read as no tray, which shows the window. The tray's Stop also ran its
own drain beside the coordinator, so Stop twice, Stop then Quit and Stop during an update were
separate executions over one child. They share one phase now, and a quit landing during a stop is
deferred and run afterwards.

**The startup budget was counted from the wrong point.** Registration is inside the overall deadline,
and the budget for finding an existing runtime now starts when probing starts. Counted from process
start, a slow tray or session-bus registration spent it and then presented as nothing listening,
which starts a second proxy beside the one already there.

**The Windows app origin is allowed.** The pinned Tauri serves the app from `tauri.localhost` there
because wry needs an http origin, so the window's first navigation to its own page fell through to
the external-browser branch. That exact host with no port: not localhost generally, and no remote
IPC widening, since the capability file still declares no `remote` entry.

**The management client has its own network policy.** It refuses redirects, because the pinned
reqwest does not treat this custom credential header as sensitive and it would carry across a hop,
and it refuses system proxies. It will not send the token until it has confirmed from the
unauthenticated health body that the instance answering is the one the shell bound to, and a request
is bound to that pid, that port and the generation it was authorised under. The updater's download
client is untouched and keeps its own policy.

One finding is **not** this lane's and is not addressed here: the installed-artifact gate running
destructive cleanup in a `finally` block after its preflight refuses. That is lane F.

## The `macos 1/2` flake at `4ced72931f`, and why it is not this change

At head `4ced72931f` the requested job `macos 1/2` failed on a single case,
`loop propagates parent abort into a hanging iteration`, which reported
`this test timed out after 1000ms` at 2003ms.

It is not reachable from this branch. The diff at that SHA touches no `src/` file at all — only
`desktop/`, `structure/`, `tests/clients/` and the two test-layout maps. The abort and exit work in
this lane is Rust in the desktop shell, a separate binary the Bun suite never loads, and the four
new test files are source oracles that read files and assert on strings without importing anything
from `src/`. The failing case exercises `runWithWebSearch` in `src/` through a local adapter.

The failure shape points the same way. `hangUntilAbort` in that file has no timer — it settles only
on abort — so nothing in the path can produce two seconds, and the loop's own log lines on that run
report the cancellation completing in 9ms. A case whose work took 9ms and whose wall clock was
2003ms was not scheduled, which is the runner-starvation shape this repository has seen on
`macos 1/2` before.

Re-running that one job at the same SHA settles it: attempt 2 of `macos 1/2` at `4ced72931f` passed with identical code, so runner load was the only variable that changed.

No timeout was widened and no test was removed or skipped.

## Rebase onto the landed lanes

Rebased onto `origin/dev` at `34ddb4d5fd`, which brought in lane C (`#5386`) and lane E (`#5388`). The
PR had been conflicting, and a conflicting PR gets no merge ref, so GitHub was not dispatching the
matrix at all — only the PR gates ran. That is why the last full run was at `4ced72931f`.

The one conflict was in `desktop/src-tauri/src/proxy.rs`, where lane E added `no_proxy()` to the same
builder this lane added `redirect(Policy::none())` and the instance check to. Both survive: the
builder turns off system proxy resolution and redirect following, and the resolution keeps E's
reasoning for the first and this lane's for the second rather than replacing one with the other.
E's `desktop-proxy-direct-transport` oracle, which strips comments and requires `.no_proxy()` inside
the builder region, passes against the merged file.

## D5 and D4, wired to the landed lane A

Rebased onto `origin/dev` at `bcb2b92e6f`. The shell now drives the two CLI surfaces instead of
answering either question itself.

**D5 — resolution.** `desktop/src-tauri/src/resolve.rs` runs `ocx resolve --json` and reads one
`ocx-resolve/1` document for the configuration home, the effective port and liveness. The file it
replaces read `runtime-port.json`, fell back to 10100 and started there, so a user with a configured
`config.port` was started on a port they had not chosen — and the probe budgets that decision needs
were sitting unused one layer down.

Liveness keeps its three answers and the third one is the point. `live` attaches as a guest;
`absent-proven` means every recorded and configured endpoint was definitively dead, and **only that
authorises starting a runtime**. Everything else is unknown — a non-zero exit, a timeout, output that
will not parse, a schema this shell does not know, a missing binary — and unknown fails the state
with a diagnostic and a retry rather than being read as absence.

Two things a `live` verdict does not settle on its own, both found by review. Core's liveness
predicate accepts a connected client's listener on purpose so duplicate-start avoidance can see it,
and its own comment says a caller needing the management plane must discriminate on the role rather
than narrow the predicate — this shell needs it, so a client listener is live and *unusable*. A
runtime bound somewhere `127.0.0.1` cannot reach is the same kind of answer. Neither is an absence,
so neither authorises a start. The sequence also stopped reporting Ready against an instance it
could not identify: `bind` returns its answer now instead of swallowing it, and both call sites fail
the state on `None`, since the management token is only ever sent to a bound instance.

**D4 — stopping.** `desktop/src-tauri/src/runtime_stop.rs` runs `ocx stop --json` and reads the
`ocx-stop/1` summary. The shell was ending the runtime with a management call from inside the process
it was ending, which cannot own its own teardown: launchd and systemd can terminate the request
handler during self-unload, and the Windows respawn window can only be verified after the process
exits.

A stop counts only when five facts hold together. The process exited 0 **and** the document says so
through both `ok` and `exitCode`, so 1, 79 and 80 are refusals however the rest reads — taking the
summary's word for its own exit status is taking a claim as its own evidence. `runtimeDown` has to be
true, because a service that failed while the proxy happened to stop is exactly the case that may
respawn it. And the document has to agree with itself: only a `stopped` outcome beside a `stopped` or
`stopped-orphan` proxy, or `not-running` beside `not-running`. An outcome or proxy state this shell
does not know fails to parse, which is the same answer as a stop that did not happen.

### The four earlier re-audit items, re-confirmed after the rewiring

Checked against the rewired source rather than assumed:

- the update still downloads and signature-checks, then confirms ownership and drains, then
  installs, and refuses the install on anything but `RestartReadiness::Ready`;
- a failed drain still becomes `DrainFailed` or `OwnershipUnknown`, which a quit tolerates and a
  coordinated restart refuses — the restart branch fires only on `DrainVerdict::Drained`;
- `attach` still clears confirmed ownership, so pointing at a different runtime cannot carry the last
  one's ownership into an owner's stop;
- the tray verdict is still published only after an icon exists, and all three install failure paths
  read as no tray.

## Merging with #5399

Rebased onto `origin/dev` at `3b1fdd8d8b`, which brought in #5399 — the Windows shell loading its
own origin and hiding the console — on three files this lane owns. Both intents are in the tree;
neither side was dropped.

The console attribute in `main.rs` carries through untouched. The Stop-settle intent in `tray.rs`
carries through as behaviour rather than as code: #5399 polled `/healthz` up to ten times before
deciding the proxy was gone and printed a warning when it was not, and that block is now
`exit::request_stop`, which confirms the stop through the bundled CLI's own summary and reports a
stuck one through the logger rather than to a console that #5399 just hid.

The app origin is one function instead of the two the auto-merge left side by side. It keeps
#5399's contract — the custom scheme everywhere, the `http` spelling WebView2 needs, `https` refused
because that is not the scheme the pinned Tauri serves the app over, and no platform gate — and adds
this lane's tightening: no port, because a port means something else is answering rather than the
app. #5399's test asserted through `navigation_allowed`, which now takes an `AppHandle` and cannot be
constructed in a unit test, so its cases moved onto the helper directly; its loopback-endpoint case
is covered by the source oracle instead.

### The four hold conditions, re-confirmed after this rebase

Read out of the rebased source, not carried forward as an assumption:

- the update downloads and signature-checks, then confirms ownership and drains, then installs, and
  refuses the install on anything but `RestartReadiness::Ready`;
- a failed drain becomes `DrainFailed` or `OwnershipUnknown`, which a quit tolerates and a coordinated
  restart refuses — the restart branch fires only on `DrainVerdict::Drained`;
- `attach` clears confirmed ownership, so pointing at a different runtime cannot carry the last one's
  ownership into an owner's stop;
- the tray verdict is published only after an icon exists, and all three install failure paths read
  as no tray.

## The macOS-only compile failure at `96de2a4d91`

Every requested job at that head was green except `macos widget + bundle`, which failed at the
`Build unsigned desktop app` step and took the aggregate `ci` down with it:

    error[E0004]: non-exhaustive patterns: `&ProxyError::Foreign` not covered
      --> src/widget.rs:93:15

This lane added `ProxyError::Foreign` to `proxy.rs` — the listener answered, but not as the instance
this client is bound to — and `widget.rs` matches that enum exhaustively. `widget.rs` sits behind
`#[cfg(target_os = "macos")]`, so `desktop shell` on Linux compiles the crate without it and stayed
green through fmt, clippy `-D warnings` and `cargo test`. `macos widget + bundle` is the only job that
compiles the file, and it had been queued or cancelled on every earlier head, so this was the first
run that reached the compiler at all.

The fix adds an explicit arm rather than a catch-all. A runtime this app did not
start is a different event from a fault, and the widget is the one surface that shows it with no
context around it: `degraded` would claim the proxy is misbehaving and `unreachable` would claim
nothing is there, so either would tell a user whose own npm or CLI runtime holds the port that a
working setup is broken. It gets its own `foreign` state titled `External runtime`, and the widget
draws it in the neutral secondary colour: `tone` in `app/Sources/OpenCodexWidget/Views.swift` falls
through to `.secondary` for a state string it does not know. Nothing downstream needs teaching,
because `WidgetSnapshot.state` is a free-form `String` on the Swift side and the closed `ProxyState`
enum is not on this path. Keeping the match exhaustive means the next variant added to
`ProxyError` is a compile error here again rather than a silent mislabel.

The wording matches what the startup surface already says for the same condition — "the runtime
answered but did not identify itself, so this app did not attach" in `startup.rs`.

Two things were checked before pushing, because both have bitten this branch: `widget.rs` carries no
size-ratchet cap and no file outside the crate pins either variant list, and the crate's own
`#[cfg(test)]` mapping test was extended in the same commit, since a stale assertion there fails
clippy `--all-targets` and skips the `cargo test` step behind it. `cargo fmt` reformatted nothing.

The widget commit touches one file, `desktop/src-tauri/src/widget.rs`, 29 insertions and 1
deletion, so the four hold conditions below are unchanged by it; they
were re-read out of the source at this head regardless: `updater.rs` refuses the install on
anything but `RestartReadiness::Ready`, `exit.rs` grants a restart only on
`DrainVerdict::Drained`, `State::attach` stores `false` into `confirmed` before it swaps the
client, and `from_host_registered` answers `Unavailable` for both `Some(false)` and `None`.

## A failure that existed only in the merge

Fixing the compile error surfaced a second one of a different kind. At `0b9d276432`, `test 1/4`
failed on this lane's own oracle,
`desktop install identity > the owner values are the ones the record accepts`, which had passed at
every earlier head.

Nothing in that push touched `src/`. #5400 landed on `dev` in the meantime and moved the runtime
validation of an ownership claim out of `src/service/state.ts` into the new
`src/service/install-state-contract.mjs`, renaming the receiver from `ownership` to `value`. Each
branch was correct at its own head: #5400 moved a check nobody else was reading, and this lane read
a check nobody else was moving. Only the merge has both.

The oracle now reads both halves of core's answer instead of that one literal — the runtime
rejection a record on disk actually meets, in the contract module, and the exported `ServiceOwner`
type every caller is compiled against, in `state.ts`. That is the stronger assertion anyway: a
parse that accepted a third owner and a type that forbade it would disagree exactly where a
takeover happens, which is the case this file exists to catch.

The branch is rebased onto `dev` at `bd4822bcea`, which also brings in #5387 and #5401, and the
rebase was clean. Of the four `src/` files this lane's oracles read — `src/cli/resolve.ts`,
`src/cli/stop-report.ts`, `src/server/index/serve-options.ts` and `src/service/state.ts` — only
the last is touched by anything in that range. Every literal the oracles assert was re-extracted
from the rebased tree and confirmed present, with the extraction itself asserted non-empty so a
silently failing scan cannot read as a pass.

## Verification

**Local checks: NOT RUN.** The suite, `test:changed`, typecheck, builds, cargo test/build/clippy,
dependency installs and any live runtime are not run for this lane. Hosted CI at the exact head is
the evidence, read per job and distinguishing what the event requested from what it skipped.

Earlier heads are worked examples of why that is the judge rather than a formality. On `dc85c9f`, `desktop shell` failed at the clippy step, and the `cargo test` step after it was therefore **skipped**, so that run is not evidence about the Rust tests at all. On `f63c3ed`, `test 1/4` failed on a scoped assertion that still looked for a guard `claim_drain` had replaced with a match. Both are fixed, and per-step results are read rather than the job conclusion.

- Every claim about the pinned dependencies was read out of the vendored crate sources rather than
  assumed: that `tao` registers only `applicationWillTerminate`, that Tauri installs its default
  macOS menu when none is set, that `prevent_exit` is ignored for `RESTART_EXIT_CODE`, that
  `ExitRequested` carries `None` for a user gesture, and that Tauri checks the ACL for any invoke
  from a non-local origin, so `withGlobalTauri` does not give the loopback dashboard a command.
- `dbus` is already compiled for every Linux build of this shell: `tao` enables its own `dbus`
  feature by default, so naming it adds no package and no system library.
- Independent reviewers audited the diff under the same no-local-execution rule; their blockers are
  folded in rather than argued with. The macOS menu hole, the any-error-means-stopped drain, the
  unbounded probes, the quit-racing-startup spawn, the watcher-versus-host question, the pre-window
  work in `setup()`, the tray verdict published before its icon, the retry that would install a
  second tray, and the menu mutex held across a main-thread setter all come from those passes.
- Every assertion literal and every ordering the four test files rely on was checked against the
  current source by reading it.

Regression tests are written and registered but **not executed here**. Each is a source oracle,
because CI builds this shell against a zero-byte sidecar and has no graphical session to press Cmd+Q
in. What each would reject:

- `tests/clients/desktop-exit-ownership.test.ts` (INV-DESKTOP-01) — goes red on a tray Quit that
  calls `app.exit` directly, on any `.kill()` in any Rust file under the shell (enumerated from
  disk, so a new module cannot opt out), on the predefined macOS Quit returning, on a drain that
  reads any endpoint error as a stopped runtime, on a spawn that ignores an exit already in flight,
  and on a close handler that hides without consulting the tray verdict.
- `tests/clients/desktop-startup-surface.test.ts` — goes red if `setup()` resolves, registers or
  starts anything again, if the window is built after the sequence begins, if the spawn events go
  back to `_events`, if a probe uses the unbounded `is_alive()`, if the page stops deriving its
  state list from the shell, if it reaches for a platform dialog, or if either entry point can fail
  without reporting into the page.
- `tests/clients/desktop-tray-availability.test.ts` (INV-DESKTOP-02) — goes red on a return to
  `NameHasOwner`, on an unanswerable probe being read as available, on Linux assuming a tray before
  the probe answers, and on a tray installed without checking the verdict.
- `tests/clients/desktop-start-at-login-default.test.ts` keeps the existing write-ordering contract
  and follows it to its new home in `startup.rs`; it gains one case for the one-time login-item
  rewrite, which claims its marker only after the rewrite succeeds.

- `tests/clients/desktop-install-identity.test.ts` reads both halves of the ownership claim in one
  file — the shell's and `src/service/state.ts`'s — so a change to the owner values, the wire field
  names, the three resolution kinds or the comparison rule breaks there rather than leaving the two
  to disagree somewhere only a real takeover would reveal. It also pins what the comparison does not
  look at: the generation moves on every grant, and comparing it would make a consent the app
  already holds look foreign.

All four new files are registered in both `scripts/test-layout/layout.json` and
`tests/fixtures/test-layout-expected.json`. No file at its size-ratchet cap gains a line.

## Checklist

Scope is `desktop/`, the four source-oracle tests, their layout registration, and the two structure
documents. No credential, token or workflow path is touched. The one security-adjacent change is
`withGlobalTauri`, which injects the JS API object; it grants nothing, because the capability file
declares no `remote` entry and Tauri rejects any invoke from a non-local origin.

Co-authored-by: devin-ai-integration[bot] <158243242+devin-ai-integration[bot]@users.noreply.github.com>


lidge-jun:
## 리뷰 · 우선순위 64 / 80

바탕화면 앱에서 창 닫기, Cmd+Q 같은 종료 제스처, 트레이 Quit이 예전에는 거의 같은 길로 갔어요. 그중 하나는 세게 죽이는 길이었어요. `ExitRequested`를 안 잡아서 플랫폼 종료가 바로 `RunEvent::Exit`로 갔고, 앱이 띄운 런타임에 `kill()`이 갔어요. 유닉스에서는 SIGKILL이라서 `ocx stop`이 하던 정리(요청 마무리, 설정 복구, 상태 파일 청소)가 끊겼어요.

이 PR은 그걸 나눕니다. `exit.rs`가 종료 요청을 잡고, 트레이가 있으면 창 닫기·종료 제스처는 숨김, 트레이 Quit만 진짜 종료예요. 업데이트 재시작은 같은 정리(drain) 뒤에 다시 켜요. 정리 중에는 자식을 `kill`하지 않아요. 이 앱이 띄운 런타임만 멈춰요. macOS는 기본 메뉴 Quit이 Cocoa `terminate:`로 바로 가서 `ExitRequested`가 안 뜨는 구멍이 있어서, `menu.rs`가 같은 단축키의 일반 메뉴로 바꿉니다.

시작도 바뀌어요. 예전에는 창 없이 `setup()` 안에서 돌고, 사이드카 이벤트 스트림을 버려서 바로 죽은 자식과 느린 시작이 같아 보였어요. 지금은 창을 먼저 만들고 `startup.rs`가 등록→해석→탐색→붙기/시작→대기 상태를 30초 한도로 보여 줘요. 리눅스는 세션 버스에서 StatusNotifier 호스트가 있는지 보고, 트레이가 없으면 창을 보여 주고 닫으면 종료해요. 베이스는 `dev`예요. 로컬 스위트는 PR에 안 돌렸고, 소스 읽기 테스트와 CI를 증거로 적었어요.

라인 - `desktop/src-tauri/src/startup.rs` `register` (대략 454–468행) - 프로브가 `Available`인데 `tray::install`이 실패하면 로그만 남기고 `ExitCoordinator`의 `hides_to_tray`는 그대로예요. 로그인 자동 시작이면 창도 안 켜진 채 아이콘도 없을 수 있어요. 수동 실행이면 창은 보이지만 닫기·Cmd+Q가 숨김으로 가서, 다시 열 트레이가 없으면 프로세스가 남아요.

라인 - `desktop/src-tauri/src/menu.rs` + `exit.rs` - 앱 메뉴 Quit/`Cmd+Q`는 고쳤어요. 도크의 Quit이나 다른 `terminate:` 경로는 여전히 `ExitRequested` 없이 프로세스가 끝날 수 있어요. 예전 `RunEvent::Exit`의 `kill()`은 뺐으니 런타임은 SIGKILL은 안 당하지만, drain도 안 타고 남을 수 있어요.

라인 - `desktop/src-tauri/src/tray_availability.rs` - 질문은 `IsStatusNotifierHostRegistered`예요. 문서대로 리눅스 백엔드는 AppIndicator를 만들어요. 호스트 등록과 아이콘이 실제로 보이는 조건이 어긋나는 데스크톱이 있으면, “있다”고 숨긴 뒤 아이콘이 안 나올 수 있어요.

라인 - `desktop/src-tauri/src/sidecar.rs` `DRAIN_DEADLINE` / `exit.rs` `start_drain` - 마감이 지나도 종료는 진행해요. PR·`structure/desktop-shell.md`에 적힌 선택이에요. 런타임이 남으면 `ocx stop`으로 거둘 수 있다고 봤어요.

메인테이너의 판단이 필요한 지점

트레이 설치 실패 때 `hides_to_tray`를 바로 `false`로 되돌릴지 정해 주세요. 실패 후에는 창을 보여 주고, 닫기를 종료(drain)로 두는 편이 안전해 보여요.

macOS에서 도크 Quit까지 drain에 넣어야 하면, 메뉴 교체만으로는 부족해요. 그 경로를 막을 수 있는지, 아니면 문서에 “도크 Quit은 정리를 건너뛸 수 있다”고 쓸지 정해 주세요.

drain 마감 후 그냥 나가는 선택은 문서와 맞춰 두었어요. 런타임을 반드시 확인한 뒤에만 끝내야 한다면 지금 동작은 안 맞아요.

너의 추천

`tray::install` 실패(또는 메인 스레드 설치가 끝나지 않음)면 `set_tray(Unavailable)`로 되돌리고, 숨긴 창이 있으면 보여 줘라. 도크/`terminate:` 경로는 가능하면 막거나, 못 막으면 셸 문서에 한계를 한 줄로 적어라. StatusNotifier와 AppIndicator가 어긋난 제보는 이슈로 남겨 두고, 지금은 호스트 프로브+설치 실패 롤백이 우선이다. drain 마감 후 종료는 문서대로 유지해도 된다. CI의 desktop/소스 오라클이 초록인지 보고 머지해라.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
Requesting changes on exact head `64cd7a4e5844f38be86f90b9dd2786c7a631dc42`.

The tray-install rollback from the first review is fixed: the coordinator now publishes `Unavailable` unless an icon was actually installed. Three blockers remain:

1. `desktop/src-tauri/src/exit.rs:301-318` restarts after any drain outcome. If an update restart gets `StillRunning` or `Refused`, the old owned runtime remains alive, the new app can attach as a guest, and the old runtime may later disappear underneath it. Keep the documented proceed-on-failure tradeoff for an explicit user quit if desired, but a coordinated restart must abort/defer until the owned runtime is confirmed stopped.
2. `desktop/src-tauri/src/tray.rs:202` refreshes immediately while startup has not attached or spawned a proxy, so it always returns early. `finish()` only calls `set_owned`; the title and widget stay empty/stale until the 60-second timer. Add an explicit refresh once the runtime reaches Ready and pin it in the startup contract test.
3. The PR and `structure/desktop-shell.md` claim that the platform quit gesture shares the coordinated path, but only the rebuilt application-menu Cmd+Q is intercepted. Dock Quit and other Cocoa `terminate:` paths still bypass `ExitRequested` with the pinned tao delegate. Either close that path or state the limitation precisely; the current one-exit-path claim is false.

Replacement exact-head CI is also still running. Please address these before approval.

Ingwannu (changes requested):
Re-reviewed current head `4ced72931f574e45b15fa3175d34dec131bdb78b` after the runtime-ownership integration. The new ownership identity work is relevant and directionally sound, but it does not resolve the three blockers from the prior review:

1. `desktop/src-tauri/src/exit.rs` still calls `app.restart()` after every drain outcome. A coordinated update restart must not launch a successor after `StillRunning` or `Refused`; that can attach the new shell as a guest to the old runtime and later lose it. Keep the documented proceed-on-failure tradeoff only for an explicit user quit if desired.
2. `tray::install` still performs its refresh before the proxy is attached, and `startup::finish` only calls `set_owned`. There is no post-Ready `refresh_now`, so title/widget state may remain empty for the 60-second timer.
3. The macOS Dock/Cocoa `terminate:` limitation remains documented rather than closed. That is acceptable only if the PR/title/current contract stop claiming one complete platform exit path; otherwise this remains a functional gap.

The current branch also has to resolve the still-open startup-phase reporting finding: a failed operational phase must not appear in both `completed` and `failed_in`.

Please address those on a replacement head and rerun exact-head desktop/hosted CI. The ownership integration alone does not make this mergeable.

PR #5386, opened 2026-09-21
Make background-service runtime ownership durable, and stop every path that would undo it

lidge-jun:
## Summary

A packaged desktop app can take the background proxy over from an npm install, and today that
takeover does not survive anything. Ownership is a boolean the desktop shell recomputes at every
launch from whether it happened to spawn a child, and the shared service install state has no field
for it, so "ask once, then own permanently" cannot be expressed and no test can see it go wrong.
This puts the claim in the one record both sides share and makes every writer carry it.

`src/service/state.ts` gains an owner (`cli` or `desktop`), an opaque `installId` naming the owning
installation, and a `consentGeneration`. An absent claim keeps meaning what it means on every
installation that predates the field: the npm install owns the runtime.

`swapServiceInstallState` is the only writer. It holds an `O_EXCL` lock beside the anchor record for
the whole read-modify-write, re-reads the anchor immediately before committing, and compares the
committed bytes afterwards; `revision` is the compare-and-swap token. The lock excludes cooperating
writers, because the revision check alone is not atomic — two processes can both pass it, both
commit and both verify their own bytes, after which the second silently drops the first and reports
success. The check stays for a writer that does not take the lock, such as an older `ocx`.

The durability fix is in `writeServiceInstallState`. Every install, repair, update and stop path in
`orchestration.ts`, `launchd.ts`, `systemd.ts`, `windows-ops.ts`, `windows-scheduler.ts` and
`repair.ts` reaches the record through it, and it used to rebuild the whole thing and replace the
file — so a repair started by a tray helper or by `ocx update` erased a takeover the user had
consented to, silently. It now rebuilds only the install provenance and carries the claim across.

Reads are fail-closed. `resolveServiceOwnership` reads every state path and answers `none`, `owned`
or `unknown`; absence is the only thing that means no claim, so an unreadable path, a corrupt anchor
record, or paths naming different owners all refuse rather than reading as CLI-owned.
`consentGenerationCeiling` survives a release, so grant / release / grant cannot reuse a number an
app-local record may still hold.

The verbs that activate the npm registration then refuse a foreign or unknown owner:

- `ocx service repair` and `ocx service restart` stop before they assert, write, stop or start
  anything. The message names the claim, states that the registration was left untouched, and
  points at the way back.
- `ocx service start` reports the same refusal. The Windows tray starts the service automatically,
  which would otherwise put a second proxy beside the one the app is running. It reports rather
  than throws, because the tray drives it through a caller that does not catch.
- `stop` and `uninstall` stay ungated; they deactivate.
- `ocx update` vetoes both the pre-update stop and the post-update service refresh. Under a desktop
  owner the running server is the app's bundled sidecar, which the update cannot bring back, and
  the refresh would re-enable the npm launcher the takeover superseded.

All three update lanes apply that veto, from one plain-ESM rule in
`src/update/runtime-ownership.mjs`: the Bun path in `src/update/index.ts`, the npm and pnpm path in
`bin/ocx.mjs`, and the dashboard worker in `src/update/job.ts`. The launcher is in scope because it
is the lane most npm users take, and its post-repair fallback would have started a second proxy.
The dashboard worker is in scope because it defaults to restarting: after the package updater
correctly left the runtime alone, the sidecar's unchanged pid failed its restart evidence, so it
reclaimed the port, ran the repair that now refuses, and fell through to a direct npm start. Both
package updaters re-read the claim before any direct-start fallback, because ownership is sampled
before an install that can take minutes and the refusal is indistinguishable from a generic failure
at that layer.

`ocx service install` is the one verb that takes the runtime back, and it releases the marker only
after the registration succeeded — releasing first meant a cancelled UAC prompt or a failed cleanup
left the retained registration looking CLI-owned. The registration itself is never deleted.

One pre-existing defect had to be fixed here: `parseServiceInstallState` rejected `cliPath: null`,
which `cliEntry()` returns for a standalone binary, so every standalone install wrote a record its
own parser refused. That read back as "no install state" before, and as "nobody owns the runtime"
once ownership moved into the record.

The contract the app half needs is published in `structure/runtime.md`.
`ownershipGrantedTo(ownership, owner, installId)` is the comparison an installation applies to its
own locally stored install id: true means it already holds consent, false against a recorded claim
means a different installation owns the runtime and consent must be asked again, and a null claim
means the CLI install still owns it. The app-local record is not in this PR.

Accepted limit: the veto reads the recorded claim, not the live process. An app removed without
releasing leaves a stale claim, and proving which runtime is answering needs the identity the
bundled CLI's resolve contract will carry. Until then every refusal names `ocx service install` as
the way to clear it.

## Verification

**Local checks were NOT RUN for this head, by instruction.** No `bun run test`, no `bun test`, no
`bun run typecheck`, no build, no dependency install, no `ocx` execution, and no service or
credential mutation. Hosted CI at the exact head SHA is the adjudicator for this branch.

What was done instead, by reading the source:

- Traced every caller of the changed exports (`writeServiceInstallState`,
  `parseServiceInstallState`, `repairService`, and the three update lanes) and confirmed each
  reaches the record only through the preserving writer.
- Confirmed by import graph that nothing in this diff is reachable from `startServer`, the Codex
  catalog, or the Codex entitlement path: `src/codex/model-entitlements.ts` and `src/codex/catalog*`
  have no edge into `src/service/` or `src/update/`, and the only `src/server/index.ts` import from
  the service barrel is `currentServiceHomes` / `serviceStatePathsForOpenCodexHome`, neither of
  which changed behaviour. No new module-load side effect was introduced.
- Checked the contract documents against the code: `structure/runtime.md` names only shapes and
  functions this tree has.

Regression tests are included but not executed. What each would catch, and the change that would
make it fail:

| test file | case | fails if |
| --- | --- | --- |
| `service-ownership-state` | ownership survives a repair | `writeServiceInstallState` stops merging the claim onto the rebuilt provenance |
| `service-ownership-state` | the swap detects a late writer | the pre-commit revision re-read is removed |
| `service-ownership-state` | the anchor lock blocks a held write | the lock is opened without `O_EXCL` |
| `service-ownership-state` | unreadable and corrupt records resolve to `unknown` | either fail-closed branch of `resolveServiceOwnership` is removed |
| `service-ownership-state` | the generation cannot be reused | the ceiling is dropped on release |
| `service-ownership-state` | a standalone record parses | `cliPath: null` is rejected again |
| `service-ownership-handover` | repair and restart refuse a foreign or unknown owner | the gate in `repairService` is removed, or moved after any seam |
| `service-ownership-handover` | install releases only after success | the release moves back before `installServiceSafely` |
| `update-desktop-owner` | the veto and its notices | `planUpdateRuntimeHandling` stops distinguishing a foreign or unknown owner |
| `update-desktop-owner` | all three lanes gate, and two re-read | any lane drops its gate or its pre-fallback re-read |

The ownership fixture restores every path in `serviceStatePaths()`. A write lands on the legacy
`~/.opencodex/service-state.json` entry as well as the pinned home, and under the suite that second
path is a sandbox shared by the files in one worker — without the restore, a claim recorded here is
visible to `tests/service/service.test.ts` and to the dashboard worker's tests, where the repair
gate and the restart veto then fire on state those files never wrote.

Three new test files are registered in both `scripts/test-layout/layout.json` and
`tests/fixtures/test-layout-expected.json`. `src/update/job.ts` sits one line under the 2000-line
ratchet threshold, so the dashboard worker's gate is extracted to
`src/update/restart-ownership.ts` rather than inlined; no cap was moved.

Two existing source-oracle assertions were adjusted to the new shape while keeping what they
encoded. `tests/service/service-start-environment.test.ts` now proves the environment guard still
precedes `ops.start()` with nothing that reaches the service manager in between, instead of
requiring the two statements to be adjacent. `tests/update/update-stop-first.test.ts` reads the stop
gate with the ownership veto in front of the four signals that still reach it unchanged.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added runtime ownership tracking for background services.
  - Updates now respect desktop-owned or unknown runtimes and avoid stopping, refreshing, or restarting them.
  - Service start and repair actions refuse to proceed when ownership cannot be verified.
  - Added clearer guidance when an ownership conflict prevents an action.

- **Bug Fixes**
  - Preserved ownership information across service installation, repair, and update operations.
  - Improved handling of unreadable or conflicting service state to fail safely.


lidge-jun:
## 리뷰 · 우선순위 61 / 80

이 PR은 “데스크톱 앱이 백그라운드 프록시를 넘겨받은 뒤에도, 그 소유권이 재시작·repair·update에 안 지워지게” 만드는 변경이다. 예전에는 소유권이 앱이 켜질 때마다 다시 세는 일시적 값에 가까웠고, `service-state.json`을 고치는 경로(repair, update, stop 등)가 파일을 통째로 새로 쓰면서 동의를 조용히 지울 수 있었다. 지금은 공유 상태 파일에 `owner` / `installId` / `consentGeneration`을 넣고, 쓰기는 `swapServiceInstallState` 한곳으로 모으며, provenance만 다시 쌓을 때 소유권은 유지한다. 읽기는 실패하면 CLI 소유로 취급하지 않고 `unknown`으로 막는다. `repair`/`restart`/`start`와 `ocx update` 세 갈래(Bun, npm/pnpm 런처, 대시보드 worker)는 남의 소유이거나 판독 불가면 npm 등록을 다시 켜거나 사이드카를 죽이지 않는다. `install`만 성공 뒤에 마커를 풀어 CLI가 다시 가져간다. standalone의 `cliPath: null` 파싱도 같이 고쳤다. base는 `dev`이고, 앱 로컬 consent 저장은 이 PR 밖이다. 로컬 테스트는 PR 설명대로 이 head에서 안 돌렸고, CI가 판정한다.

라인 - `src/service/state.ts` `writeServiceInstallState` — `resolveServiceOwnership()`을 락/스왑 밖에서 한 번 찍고, `preservedConsent`가 `kind === "owned"`일 때 그 스냅샷을 `current`보다 앞세운다. legacy 미러에만 있는 claim을 앵커로 옮기려는 의도는 맞지만, 그 사이에 다른 프로세스가 `releaseServiceOwner`/`recordServiceOwner`를 하면 방금 푼 claim이 다시 살아나거나, 더 새 claim을 덮어쓸 수 있다. 멀티패스 merge는 락 안에서 evidence를 다시 보는 편이 안전하다.
라인 - `bin/ocx.mjs` `readOwnership` vs `src/update/index.ts` `resolvedRuntimeOwnership` — Bun 경로는 `resolveServiceOwnership`(모든 state path, fail-closed)을 쓰고, npm/pnpm 런처는 `configDir()/service-state.json` 한 파일만 본다. `OPENCODEX_HOME`과 legacy `~/.opencodex`가 갈라지거나, 레코드 전체가 깨졌는데 `ownership` 객체만 멀쩡한 경우 두 레인의 거부/진행이 달라질 수 있다. 공유 룰(`planUpdateRuntimeHandling`)은 같아도 입력 판독이 같지 않다.
라인 - `src/service/state.ts` `withServiceStateLock` — 락 파일이 30초보다 오래되면 age만 보고 `unlink`한다. 아주 느린 쓰기/멈춘 홀더와 겹치면 두 writer가 동시에 들어갈 수 있다. 협력 writer + revision 재시도로 많이 막히지만, “락이 원자성을 닫는다”는 설명보다는 완화된 배제에 가깝다.
라인 - 수락된 한계(문서에도 있음) — veto는 살아 있는 프로세스가 아니라 기록된 claim만 본다. 앱을 지우며 release를 안 하면 `ocx update`/`repair`가 막힌 채로 남고, 복구는 `ocx service install`뿐이다. 의도된 trade-off이지만 UX/지원 부담은 남는다.
라인 - 검증 — 이 head에서 로컬 `bun test`/typecheck는 안 돌렸다고 명시돼 있다. 회귀 테스트 파일·레이아웃 등록은 있어 보이지만, merge 판단은 호스티드 CI(특히 service ownership / update-desktop-owner / 크로스 OS 서비스 잡) 결과에 묶는 게 맞다.

메인테이너의 판단이 필요한 지점

락 밖 ownership 스냅샷을 legacy 미러 승격용으로 남길지, 락 안 재조회로 바꿀지. npm 런처도 멀티패스/`unknown`을 TS와 같은 기준으로 맞출지(아니면 “런처는 단일 path”를 계약으로 고정할지). stale claim UX를 이 PR에서 더 깎을지, resolve 계약 후속으로 미룰지.

너의 추천

방향은 맞고(소유권을 공유 기록에 두고 활성화 경로를 fail-closed로 막는 것), 테스트·문서 밀도도 높다. merge 전에 (1) `writeServiceInstallState`의 ownership merge를 락 안으로 옮기거나, 최소한 그 TOCTOU를 테스트/코멘트로 고정하고 (2) `bin/ocx.mjs` 판독을 Bun 레인과 입력 계약까지 맞춘 뒤, CI 초록을 보고 합치는 쪽을 권한다. types/config 중복 무효 PR 이슈는 이 diff와 무관하다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
Requesting changes on exact head `b9bd4137d58fc1ede1ec88f572e8cc12b937a652`. Hosted CI is green, but the ownership boundary still has merge-blocking races and divergent readers:

1. `writeServiceInstallState` resolves cross-path ownership before `swapServiceInstallState` acquires the anchor lock, then `preservedConsent` prefers that stale resolution over the locked `current` record. A concurrent release/newer desktop claim can be resurrected or overwritten without tripping the revision check. Resolve/merge ownership under the same lock, or compare generations without allowing a pre-lock value to replace a newer locked claim.
2. `bin/ocx.mjs` reads one `service-state.json` and treats any parseable object without `ownership` as known-unowned. The authoritative Bun path validates the complete state, inspects every configured/legacy path, and refuses invalid, unreadable, or conflicting evidence. Move that resolution contract to a shared plain-ESM implementation or consume an authoritative machine-readable command, with regressions for an invalid ownership-free anchor, unreadable secondary path, and conflicting claims.
3. `commitServiceStateFile` writes directly to the anchor. An interrupted/truncated write converts a valid claim into `unknown` and blocks every lifecycle/update path. Use a same-directory temporary file plus atomic rename while preserving mode/ACL hardening.
4. The stale-lock eviction is age-only while the lock mtime is never refreshed; a live slow writer can be unlinked after 30 seconds and overlap with a second writer. Add holder identity/heartbeat or refuse age-only eviction. `git diff --check` also reports an extra blank line at EOF in `tests/service/service-ownership-handover.test.ts`.

The new user-facing ownership refusals and explicit `ocx service install` takeover path also need durable docs-site documentation, not console text only. Keep this blocked until these are resolved on a new head and exact-head CI is green.

PR #5387, opened 2026-09-21
fix(gui): replace the platform dialogs the desktop webview cannot draw

lidge-jun:
## Summary

Inside the desktop app, the dashboard's consent gates and result reports do nothing. wry implements three `WKUIDelegate` methods — the file open panel, the media capture permission request, and window creation for a navigation action — and WKWebView draws no JavaScript dialog when its UI delegate does not implement the matching panel method. So `confirm()` returns false without drawing anything, `alert()` draws nothing, and `prompt()` cannot collect input.

Thirteen consent gates and seven result reports were inoperative there. Among them the sidebar's red power orb, so stopping the proxy from the app did nothing at all, and the refresh orb beside it, whose handler opens with `if (!confirm(...)) return;` and therefore took that early return on every click. Removing a provider key, an account, a routing profile or a custom model, hiding a model, switching provider account mode, uninstalling the tray helper, and the memory drain-and-restart all failed the same way: the user clicks, is silently declined, and sees nothing. Alias editing used `window.prompt()`, which cannot collect input at all, so those edits could not be made in the app. The same dashboard works in a browser, which is why this reads as "the app is broken" rather than "the dashboard is broken".

`confirm`, `alert` and `prompt` now leave `gui/src` entirely.

- `confirmAction` and `requestTextValue` in `gui/src/action-dialogs.ts` open a real `<dialog>`, built the way `gui/src/admin-token-dialog.ts` already builds one and using the same modal classes from `styles.css`. They keep the call-site shape — ask, await an answer, act on it — so a refusal still issues no request, and they carry what the existing in-page dialogs get right: cancel without issuing the request, focus return to the trigger, Escape and the backdrop as refusals rather than unanswered closes, input validation, and an accessible result.
- The seven `alert()` reports move to the existing `ToastNotice`. `useCodexRestart` now requires its consumer to supply that sink, so a forgotten report fails to compile instead of vanishing silently; both consumers report through the shell, which outlives a page that a 30-second restart can outlive.
- `requestTextValue` mirrors the alias contract both credential routes already enforce — trimmed, at most 80 characters, no control characters — so an over-long or pasted value is reported beside the field instead of returning an opaque 400.
- `ProxyStopOutcome` and `CodexRestartOutcome` become discriminated unions, so a failure always carries its reason and the reporting surface never has to invent a status it never saw.
- `gui/src/pages/Models.tsx` sits on its recorded line cap, so alias editing moved to a sibling module rather than growing it.

Two decisions a reviewer may want. Implementing the macOS delegate was rejected because WebView2 enables script dialogs by default and WebKitGTK shows them through its default handler, so repairing one platform would leave the product's consent UI platform-dependent. And `ProviderModels` now claims its single flight *before* awaiting consent: the old gate was synchronous and nothing could interleave with it, whereas a browser's modal inertness is the platform's courtesy rather than this component's invariant.

### The source guard

`tests/gui/platform-dialog-guard.test.ts` is what stops them coming back. It matches the global **call form**, not the identifier, because the identifier is legitimate in four places this repository actually uses: a `confirm()` method on a session object, the `promptForAdminToken` helper, a `confirm` prop, and an executable sample string containing the word. A lexical ban would reject all four.

The scanner masks comments, string literals, template text and regular expressions before matching, and tracks the previous *token* rather than the previous character. That is not tidiness: a JSX closing tag `</div>` read as a regular expression, or an apostrophe in JSX body text read as a string, masks the rest of that line — which in dense JSX is exactly where a handler calling a dialog would sit. Both were real blind spots in an earlier revision of this scanner. It also reads `confirm?.(x)` and `(confirm)(x)`.

### Screenshot

The two surfaces that replaced them: the consent gate behind the sidebar's stop orb, and the text entry that replaced `window.prompt()` for alias editing. Captured against a freshly started proxy with an empty config home, so it carries no account data.

In-page consent dialog and text-entry dialog replacing confirm() and prompt()

## Verification

**Local checks: NOT RUN.** This lane was switched mid-flight to a no-local-execution contract. Nothing has been run since: no suite, no typecheck, no build, no install. That includes everything in the last two commits — the navigation dismissal, the Escape move, the scanner fixes, and every test added or edited alongside them were written and reviewed by reading, never executed. Hosted CI at the exact head is the only evidence of correctness, and a skipped or cancelled job is neither a pass nor a failure.

Checks that ran *before* that instruction arrived, recorded here because they are real results and not claimed as current:

- `cd gui && bun test tests` — 2258 pass, 0 fail. The same suite at the base commit `a6e8df4` in the same worktree gave 2247 pass, 0 fail, so the 11 new tests are the whole delta and nothing regressed.
- `cd gui && bun x tsc -b --noEmit`, `bun run lint`, `bun run lint:i18n` — clean.
- `bun run typecheck`, `bun run privacy:scan`, `bun run structure:check` — clean.
- `bun test tests/gui/platform-dialog-guard.test.ts` and `bun test tests/test-layout.test.ts tests/test-layout-tooling.test.ts` — pass.
- The repository-wide `bun run test` is **NOT RUN**. It cannot be judged from this working tree anyway: the checkout sits inside the Codex home, so `test-home-guard` refuses every test-owned temp cleanup and hundreds of unrelated suites fail on that alone.

Static verification used instead, by reading the sources:

- Every converted call site keeps its refusal path. `confirmAction` resolves `false` for Cancel, Escape and the backdrop, so each `if (!consented) return;` returns before any `fetch`, exactly as the synchronous `confirm()` did. `requestTextValue` resolves `null` on dismissal, so the alias editors keep their `if (entered === null) return;` guard and a dismissal is never read as a request to clear the alias.
- `ProviderSettings` captures `e.currentTarget` before awaiting, then restores the select's value on refusal — required because React re-renders that controlled `<select>` only when `accountMode` changes, and a refusal leaves it unchanged.
- The guard's own coverage is argued from the scanner source rather than from a run: it masks comments, strings, template text and regular expressions, tracks the previous token, and matches the bare, `window.`-qualified, optional-call and parenthesized call forms.

What must go red, and why, if this change were reverted:

- `tests/gui/platform-dialog-guard.test.ts` → `no dashboard source calls confirm, alert or prompt`. Reverting restores 24 global calls in `gui/src`; the test collects every one as `file:line form` and asserts the offender list is empty, so it fails listing all of them. The same test also pins the scanner itself against synthetic sources, so it fails if the matcher is weakened rather than the sources changed.
- `gui/tests/action-dialogs.test.ts` → all ten cases. They import `confirmAction`/`requestTextValue` directly; without the module the file does not resolve. Each test also installs a throwing trap for `confirm`, `alert` and `prompt`, and `afterEach` asserts none was reached, so a reverted call site fails on the trap rather than on a missing assertion.
- `gui/tests/memory-observability-card.test.tsx` → `Drain & restart posts /api/system/restart only after the in-page dialog is accepted` asserts `restartPosts === 0` and a mounted dialog after the click; with `window.confirm` restored the trap throws instead. `dismissing the restart dialog sends no request` covers the refusal half, so "withholds the request" cannot be satisfied by a component that never restarts at all.
- `gui/tests/codex-stale-banner-dom.test.tsx` → `a dismissed consent dialog sends no request and does not refresh`, and `a timeout is localized…` now reads the recorded report instead of a stubbed `alert`, so it fails if `useCodexRestart` stops reporting through the injected sink.
- `gui/tests/provider-model-management.test.tsx` → `single flight blocks a second row until the first write and all reconciliation reads finish` asserts exactly one dialog after two clicks, so it fails if the flight is claimed after consent instead of before.
- `gui/tests/app-stop.test.ts` → the handler oracle now requires `await confirmAction(` and `report(outcome.message, "err")` and asserts the handler contains no `alert(`.

None of these assertions is vacuous: each names a value the reverted code cannot produce, and the two source oracles fail on content that only exists after the change.

## Checklist

## New Features
- Added in-app confirmation and text-entry dialogs across dashboard actions.
- Added toast notifications for success, warning, and error feedback, including restart results.
- Added provider and model alias editing with validation for aliases up to 80 characters and no control characters.
- Added localized proxy-stop status and validation messages.

## Bug Fixes
- Replaced browser-native dialogs with desktop-compatible in-app dialogs.
- Destructive actions now require explicit confirmation and safely handle dismissal.
- Improved proxy-stop reporting when shutdown is rejected or cannot be confirmed.

Two further rounds of static review are folded in, and their fixes are unrun like the rest:

- A dialog is mounted on `<body>`, so it outlives the React subtree that opened it. Navigating with Back/Forward left it on screen, and accepting it afterwards resumed a closed-over handler against a page the user had already left — account removal or device revocation from a surface they cannot see. Navigation is now a refusal, and both listeners come off when the dialog settles.
- Escape moved to the document. With `showModal` the dialog reports its own `cancel` event and the document listener never fires; without it the element is merely `open`, so a dialog-level listener missed Escape as soon as focus sat elsewhere. That fallback is not a modality boundary and now says so, instead of implying a focus trap it does not have.
- The guard did not report `window.confirm?.(x)`, `(window.confirm)(x)`, `((confirm))(x)`, a call at a statement start after automatic semicolon insertion, `export default prompt(x)`, a postfix `++` making the next slash a division, or computed access through ``globalThis["alert"]``. Each was a real hole rather than a stylistic gap; all are now matched and pinned by cases in the guard test.
- The provider-model dismissal cases were vacuous: they answered through a helper that returns silently when no dialog opens, so they would have passed if Delete and Hide became no-ops — the very defect this lane exists for. They now drive the dialog and assert it appeared. The dialog test file also stops installing the `HTMLElement` global, so it proves the focus check stays realm-independent instead of hiding a reintroduction.

Added with the same red-if-reverted argument, and likewise never executed:

- `gui/tests/action-dialogs.test.ts` → `navigating away is a refusal, and leaves no dialog behind` dispatches `popstate` while a dialog is open and asserts the promise resolves `false` with nothing left mounted; without the window listeners the promise never settles. `a hash change is the same refusal` covers the other event, `Escape is answered at the document when the dialog is not modal` fails if the listener returns to the element (this DOM has no `showModal`, so it always takes that branch), and `a settled dialog stops listening for navigation` fails if `finish` stops removing them.
- `gui/tests/provider-model-management.test.tsx` → both dismissal cases now assert `actionDialogOpen(...)` before dismissing, so a control that silently does nothing fails instead of passing.

### Known limitation: the consent surface differs before takeover

When the app attaches to an existing npm-installed runtime, the dashboard inside the app window is served by **that** runtime, not by this build. These changes live in the new GUI, so until the app owns the runtime the window still shows a dashboard whose consent gates are `confirm()` and whose reports are `alert()` — and inside the app those still draw nothing. This PR cannot reach that case; it is a property of which backend is serving the page. It belongs in the shipping decision for the takeover work rather than here.

### Re-audit follow-ups folded in, also unrun

- The stop client mapped every transport failure to `accepted`. A timeout, a request that never arrived, a user abort and a connection dropped after the server accepted are different events, and collapsing them turned *the user pressed the button* into *the server acted*. `ProxyStopOutcome` is now `accepted` / `rejected` / `unknown`, and an unknown is settled by re-**reading** the instance, never by re-sending the mutation, which would repeat a stop that did arrive against whatever now holds the port. A refused follow-up connection is the evidence the runtime went away; a proxy that still answers is unknown with that reason; a client disconnect ends no process so there is nothing to re-read; and a 2xx whose body cannot be parsed is unknown, because `success: false` rides in the body that was lost.
- A consent could outlive its subject. The dialog is asynchronous, so the surface can unmount or the backend base can change while the question is on screen, and the request then went to the closure's captured target. Consent now carries a cancellation lifetime: a changed base or an unmount withdraws the open dialog, and the confirmed path re-checks its captured base before sending.
- The guard's template-literal handling is now pinned both ways: a backtick literal's text is prose (the dashboard ships an executable sample that spells out a `prompt` call on purpose), while a `${...}` hole is code and must be reported.

Still open, and not attempted here: the same consent-lifetime binding is applied to the restart path the re-audit named, not to every dialog call site. The other sites capture a target that a route change can invalidate, and the generic `AbortSignal` hook now exists for them.


lidge-jun:
## 리뷰 · 우선순위 52 / 80

이 PR은 데스크톱 앱 안 대시보드에서 `confirm` / `alert` / `prompt`가 안 그려지던 문제를 고칩니다. wry의 WKWebView는 자바스크립트 대화창용 UI 대리자를 안 넣어서, `confirm()`은 그림 없이 false만 주고, `alert()`은 아무 것도 안 그리고, `prompt()`는 입력을 못 받습니다. 그래서 사이드바 정지·새로고침, 키·계정·프로필·커스텀 모델 삭제, 별칭 편집, 메모리 재시작 같은 확인·결과 표시가 앱에서는 조용히 실패하고 브라우저에서는 정상처럼 보였습니다. 지금은 `gui/src/action-dialogs.ts`의 페이지 안 `<dialog>`와 기존 Toast로 바꾸고, 거절하면 요청을 안 보내게 흐름을 유지합니다. ProviderModels는 확인을 기다리기 전에 한 번에 하나만 돌게 잠그고, 페이지를 떠나면(해시·뒤로가기) 열린 대화창은 거절로 닫습니다. 호출 형태로 `confirm`/`alert`/`prompt`가 다시 들어오는지도 막는 가드 테스트가 있습니다. base는 `dev`이고 types/config 분할·프리뷰 배포와는 무관합니다.

라인 - tests/gui/platform-dialog-guard.test.ts:112 — 기대값은 `(confirm)(` 인데, 스캐너는 지금 `confirm)(` 을 냅니다. 패턴을 고친 뒤 옛 기대값을 안 맞춰서, hosted CI의 `test 1/4`가 이 한 줄에서 빨갛습니다. 가드가 호출을 못 찾는 게 아니라, 문자열 모양이 서로 안 맞습니다.

라인 - gui/tests/action-dialogs.test.ts (Escape is answered at the document when the dialog is not modal) — 이 테스트는 `showModal`이 없다고 가정하고 document에 Escape를 보냅니다. 그런데 happy-dom에는 `showModal`이 있어서 실제로는 모달 분기로 가고, document 키 리스너는 안 붙습니다. 그래서 프로미스가 안 끝나고 5초 타임아웃으로 `gates`가 빨갛습니다. 프로덕션 브라우저는 `showModal`을 쓰므로 실사용 경로는 다른 Escape/cancel 테스트가 이미 덮습니다. 깨진 건 “없으면” 분기 테스트입니다.

메인테이너의 판단이 필요한 지점

작성자 말대로 마지막 커밋들(내비게이션 닫기·Escape·가드 구멍)은 로컬에서 안 돌렸고, hosted CI가 유일한 증거입니다. 그 CI가 위 두 실패로 지금 헤드에서 빨갛습니다. `showModal` 없는 DOM 분기를 계속 테스트하려면 그 테스트에서 `showModal`을 잠깐 빼면 되고, “배송 표면은 전부 showModal”이면 그 케이스를 기대로 맞추거나 빼도 됩니다. macOS 대리자만 고치지 않고 페이지 안 대화창으로 통일한 선택은 타당해 보입니다.

너의 추천

방향(페이지 안 확인·토스트·거절 시 요청 없음·가드·단일 flight)은 맞습니다. 머지 전에 가드 기대 문자열을 스캐너 출력과 맞추고, Escape 비모달 테스트를 happy-dom 실제 분기에 맞게 고치세요. 그 두 줄이 초록이면 머지해도 됩니다.

이 댓글은 grok-bot이 작성했습니다

Ingwannu (changes requested):
Requesting changes on exact head `06f19f9f63271d22c33aec5b80ed626e0d82efa0`.

The two hosted-test failures called out in the maintainer review were corrected, and the navigation/Escape cleanup direction is sound. The current head still fails the required React Doctor job. Reproducing React Doctor 0.9.11 in an isolated temporary HOME confirms PR-touched blockers:

- `ProviderSettings.tsx:422` launches `confirmAction(...).then(...)` from a JSX handler with no rejection path.
- `startup-sections.tsx:201` does the same for tray uninstall.

A rejected dialog promise becomes an uncaught event-handler rejection that no React error boundary handles. Convert both handlers to `async` with `try/catch`, or add a deliberate `.catch` that reports through the existing toast/reporting surface. The new test helper also exports `actionDialogText` and `fillActionDialog` without any import; remove the unused exports or use them in the intended tests so the changed-file scan is clean.

Please rerun replacement React Doctor and the hosted GUI checks after the fix. Do not merge while the required job is red.

Ingwannu (changes requested):
Re-reviewed replacement head `f73219fc680352c24438b5a7255a292fab7d37f1`.

The prior floating-`.then` syntax is gone, but the rejection boundary is not actually fixed: both named async functions are still invoked with `void` from JSX and have no `try/catch`. A rejection from `confirmAction` or the awaited action still becomes an unhandled event-handler promise. Please catch and report it through the existing notice/toast surface; moving the same promise behind a function name only silences one lint shape.

Additional exact-head blockers:

1. `Models` now requires `reportRestart`, while multiple component-test renderers still mount `<Models apiBase=... />` without it (including `gui/tests/compatibility-lab.test.tsx`, picker/order, custom-context, discovery, empty-provider, price/status/workspace tests). Update the shared harnesses/renderers with a spy or no-op; do not weaken the production prop.
2. `tests/helpers/platform-dialog-scan.ts` does not recognize valid optional computed access such as `window?.["confirm"](...)`: its regex accepts a bare optional `?` before `[` instead of the complete `?.` token. Add every global receiver as a regression.
3. The bare-call backward scan still skips nested calls such as `run(confirm(message))` because it crosses `(` and then rejects the preceding identifier. Track an opening parenthesis as call-position evidence while preserving declaration/member exclusions; cover both `if (confirm(...))` and the nested form.
4. `actionDialogText` and `fillActionDialog` remain exported with no consumer, so the changed-file dead-export finding remains.

The screenshot requirement is satisfied. Please replace the head, rerun React Doctor, GUI tests, typecheck, and exact-head hosted CI before another review.

PR #5400, opened 2026-09-21
Close the review findings on runtime ownership: one install-state contract, locked resolution, atomic replace

lidge-jun:
## Summary

Follow-up to #5386, which landed on `dev` as `2fb2dfb947`. Codex and CodeRabbit raised six
findings against that head; the PR was merged before they could be folded in, so they are
addressed here. Each was verified against the merged code before being fixed.

**The launcher and the authoritative reader were not the same reader.** `bin/ocx.mjs` kept its own
ownership parser, and it was weaker in two ways that decide authorization: it inspected only the
anchor state path, and it returned "known unowned" whenever the `ownership` field was simply absent
— including from a record that fails the install-state contract outright, such as `{"version": 2}`
or one with an unsupported version. The Bun updater resolves that same record as `unknown` and
refuses; the npm and pnpm lane permitted the stop and the service refresh. That is an authorization
gap, not a cosmetic divergence, and my own comment there claimed the two used "the same rule".

The record shape, the state-path list and the resolution rule now live in
`src/service/install-state-contract.mjs`, which `src/service/state.ts` and `bin/ocx.mjs` both
import. The launcher's parser is deleted rather than kept aligned by a test, because being aligned
is exactly what it was not.

**Ownership was resolved before the lock.** `writeServiceInstallState` called
`resolveServiceOwnership()` and then entered `swapServiceInstallState`. A takeover recorded in
between reaches `current`, passes the revision check untouched, and is overwritten by the older
claim — a lost update the compare-and-swap cannot see, because the stale value never came from the
base record. It now resolves inside the swap, while the lock is held, and where the anchor and the
cross-path resolution still disagree the higher `consentGeneration` wins, with an equal generation
keeping the anchor as the fail-safe tie.

**An interrupted commit bricked the service verbs.** The in-place write truncates first, so a kill
or a failed write left the anchor empty or half-serialized. Before #5386 that read back as "no
install state"; after it, the fail-closed reader reports `unknown`, which blocks `start`, `repair`,
`restart` and every update until the operator runs a takeover install. Each file is now published by
writing a sibling temporary file and renaming it, hardened before the rename rather than after.
Windows can refuse the replace transiently while a scanner holds the destination, so it retries
briefly and then falls back to the in-place write — a narrow torn-write window is a better failure
than an install that cannot record what it just registered.

**The lock could evict a live holder and delete its replacement.** The stale threshold was thirty
seconds while a Windows commit runs `hardenSecretPath` per path, whose own documented worst case is
far beyond that; and release unlinked the pathname without checking it still owned it. The lock file
now carries a token naming its holder, eviction re-reads it before unlinking, release removes only
its own instance, and the threshold exceeds the longest legitimate critical section rather than the
typical one.

**Ownership decisions were not read at the point of action.** Both package updaters formed their
plan, then ran the Windows tray handoff — which spawns children — before the stop. They now re-read
the claim immediately before the stop, in addition to the existing re-read before the direct-start
fallback.

**The refusals had no documentation.** `ocx service start`, repair, restart and update gained
user-visible refusals with `ocx service install` as the recovery, and nothing in `docs-site/`
described them. The CLI lifecycle reference gains a Runtime ownership section covering the record,
which subcommands refuse and which stay open, the refusal text, and the recovery command.

## What is deliberately not fixed here

CodeRabbit asks for each ownership check to sit inside an action-scoped ownership boundary that is
held across the child process. Re-reading narrows the window between a decision and its action; it
does not remove it. Closing it needs a lease that is not the state lock — the reviewer notes, and I
agree, that holding the state lock across `ocx stop` or a service refresh would deadlock against the
child's own write. That lease is a design change larger than this follow-up and belongs with the
app-side handover contract, so it is recorded as a residual in `structure/runtime.md` rather than
claimed as done.

Translated `docs-site/` locales are not updated. They already run several hundred lines behind the
English source, and nothing in them now contradicts it.

## Verification

**Local checks were NOT RUN, by instruction.** No `bun run test`, no `bun test`, no
`bun run typecheck`, no build, no dependency install, no `ocx` execution, and no service or
credential mutation. Hosted CI at the exact head SHA is the adjudicator.

What was done instead, by reading the source:

- Checked each finding against the merged code before fixing it, and confirmed the one it names.
  The launcher case reproduces by inspection: `parsed.ownership === undefined` returned
  `ownershipUnknown: false` regardless of whether the rest of the record satisfied the contract.
- Traced every caller of the moved functions. `parseServiceInstallState`, `parseServiceOwnership`,
  `inspectServiceStateEvidence`, `resolveServiceOwnership` and `serviceStatePathsForOpenCodexHome`
  keep their signatures and semantics; the path-list dedupe is the same `resolve()` comparison,
  lower-cased on Windows, that `normalizePathForCompare` applied.
- Confirmed `parseRecordedOwnership` has no remaining reference in `src/`, `bin/` or `tests/`.
- Checked the declaration file against the assertion sites so each cast is a legal narrowing rather
  than a same-shape coincidence.
- Confirmed no new test file was added, so both test-layout maps are unchanged, and that every
  touched file stays under the 2000-line ratchet threshold.

Regression tests are included but not executed. What each would catch, and the change that would
make it fail:

| test file | case | fails if |
| --- | --- | --- |
| `update-desktop-owner` | an ownership-free record that fails the contract is `unknown` | the launcher-style "absent field means unowned" shortcut returns |
| `update-desktop-owner` | a legacy-path claim, a cross-path conflict, an unreadable path | the resolution stops spanning every path |
| `update-desktop-owner` | the launcher reads only through the contract | a parser or a `JSON.parse` of the record reappears in `bin/ocx.mjs` |
| `service-ownership-state` | the resolution is read inside the swap | `resolveServiceOwnership()` moves back outside the lock |
| `service-ownership-state` | a higher generation wins | `preservedConsent` prefers the resolution unconditionally again |
| `service-ownership-state` | staging and rename | `commitServiceStateFile` returns to an in-place write, or hardens after the rename |
| `service-ownership-state` | release removes only its own lock | `releaseHeldLock` unlinks by pathname without checking the token |

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved runtime ownership detection across CLI and desktop installations, including legacy installation locations.
  * Ownership conflicts and unreadable service state are now handled safely instead of being treated as unowned.

* **Bug Fixes**
  * Updates now recheck runtime ownership immediately before taking action, reducing the risk of stopping or starting the wrong proxy.
  * Service-state updates are more resilient against interrupted writes and concurrent operations.

* **Documentation**
  * Added guidance for runtime ownership behavior across service commands, updates, and recovery actions.


lidge-jun:
## 리뷰 · 우선순위 73 / 80

이 PR은 `#5386`이 `dev`에 들어간 뒤 남은 리뷰 지적 여섯 가지를 닫습니다. 핵심은 “누가 백그라운드 프록시를 소유하는지”를 읽는 규칙이 런타임마다 달랐던 구멍을 막는 일입니다. 예전 npm/pnpm 런처(`bin/ocx.mjs`)는 앵커 파일만 보고, `ownership` 칸이 없으면 “아무도 안 가졌다”고 봤습니다. 그래서 계약에 맞지 않는 깨진 기록도 스톱·서비스 갱신이 허용됐고, Bun 쪽은 같은 기록을 `unknown`으로 거절했습니다. 이제는 기록 모양·경로 목록·해석 규칙이 `src/service/install-state-contract.mjs` 한곳에 있고, `state.ts`와 런처가 같이 씁니다. 소유권은 락을 잡은 뒤 스왑 안에서 다시 읽고, 세대가 높으면 이기고 같으면 앵커를 지킵니다. 상태 파일은 임시 파일에 쓴 뒤 이름 바꾸기로 올립니다. 락 파일에는 토큰을 넣고, 오래된 락 기준도 30초에서 5분으로 늘렸습니다. 업데이트는 윈도우 트레이 작업 뒤에 스톱 직전 소유권을 한 번 더 읽습니다. 거절 문구와 복구(`ocx service install`)는 영문 CLI 문서에 적혔습니다. 액션 전체를 감싸는 ownership lease는 데드락 때문에 일부러 안 했고, `structure/runtime.md`에 잔여로 남겼습니다. base는 `dev`입니다. 같은 주제의 열린 중복 PR은 없습니다.

라인 - `src/service/state.ts` `commitServiceStateFile`: 이름 바꾸기가 다섯 번 실패하면 예외 종류를 가리지 않고 예전처럼 제자리 쓰기로 돌아갑니다. 스캐너가 잠깐 잠근 경우뿐 아니라, 영구 권한 오류·디스크 오류도 같은 경로로 갑니다. 그러면 이 PR이 막으려던 “중간에 잘린 파일 → fail-closed `unknown` → start/repair/update 전부 막힘” 창이 다시 열립니다.

라인 - `src/service/state.ts` `withServiceStateLock`: 락 토큰 쓰기는 `best-effort`입니다. 쓰기가 실패하면 파일은 비어 있고, release는 “내 토큰이 아니다”로 보고 언링크를 건너뜁니다. 그 락은 최대 약 5분(stale)까지 남을 수 있습니다. 토큰으로 막으려던 “남의 락을 지우는” 사고는 줄이지만, 실패 시에는 락 누수로 바뀝니다.

라인 - `tests/service/service-ownership-state.test.ts` / `tests/update/update-desktop-owner.test.ts`: 동작뿐 아니라 소스 문자열(`renameSync`, `resolveServiceOwnership()` 위치)을 검사합니다. 리팩터만 해도 깨집니다. 의도된 핀이지만, 동작 테스트와 역할이 겹칩니다.

메인테이너의 판단이 필요한 지점

윈도우에서 rename 실패 시 제자리 쓰기 fallback을 유지할지 정해 주세요. 유지한다면 “어떤 errno만 재시도·fallback할지”를 좁히는 편이 안전합니다. 락 토큰 쓰기 실패는 지금처럼 조용히 넘어갈지, 쓰기를 필수로 할지(실패 시 락 획득 실패로 처리)도 한 번만 정해 주세요. lease 잔여는 이 PR 범위 밖으로 두는 작성자 판단에 동의합니다.

너의 추천

머지 방향은 맞습니다. 런처·Bun이 같은 계약을 쓰게 한 것과, 락 안 해석·원자적 교체·스톱 직전 재읽기는 유지하세요. 가능하면 rename fallback은 일시 오류 코드만 허용하고, 토큰 쓰기 실패는 락 획득 실패로 다루세요. 소스 문자열 테스트는 동작 핀이 생긴 뒤 줄여도 됩니다. 닫을 무효·중복 PR은 없고, 프리뷰 배포와는 무관합니다.

이 댓글은 grok-bot이 작성했습니다

PR #5445, opened 2026-09-21
fix(desktop): ship the startup surface as one page the policy can name

lidge-jun:
## Summary

On Linux the desktop bootstrap window ran none of its JavaScript. The page rendered, the phase
checklist stayed empty, the headline stayed on the markup default, and no terminal state was ever
reached — so the one surface whose job is to report a failed start could not report anything.
Closes #5416.

The cause is not the asset and not the MIME type. The webview is not served the policy in
`tauri.conf.json`: the shell appends its own hashes and nonces to `script-src` before serving the
page. Once a hash or a nonce appears in that directive, any inline allowance is inert, so every
script has to be named individually. The shell's own nonce injector matches `script[src^='http']`
only, and this page loaded its script by relative path, so it was never named. Where the asset
origin does not satisfy `'self'` — Linux, measured — the script is refused and the page is a
static picture.

The page and its script now ship as one file. An inline script carrying the nonce token is named
by the same mechanism that names the shell's own scripts, so the surface runs with the policy
intact rather than by weakening it. `desktop/ui/main.js` is gone; its contents are unchanged
inside `index.html`, and the comment above it records why the file may not come back.

## Verification

- **Local checks: NOT RUN.** No repository test, typecheck, build or install was run; verification
  is static review plus hosted CI at the exact head.
- Measured on a real GNOME/X11 session, building and installing the Debian package from this
  branch each time:
  - the asset itself was never the problem: a probe that fetched it from the page saw
    `status=200`, `content-type: text/javascript`, 5941 bytes;
  - with the configured policy removed the page ran and the checklist rendered, which is what
    first placed the cause in the policy rather than the asset;
  - naming the asset-protocol scheme and host in `script-src` changed nothing, and adding
    `'unsafe-inline'` changed nothing either — the signal that a hash or nonce was already present
    and making inline allowances inert;
  - an inline script carrying the nonce token **did** run, while the external script carrying the
    same token still did not, which is what makes inlining the fix rather than a workaround;
  - with this branch installed, the surface renders all six phase rows with the policy unchanged.
- Read out of the pinned shell dependency rather than inferred: the injector's selector is
  `script[src^='http']`, and the nonce replacement scans the asset for the token, substitutes a
  fresh value, and adds `'nonce-…'` plus `'self'` to the directive.
- The source oracle in `tests/clients/desktop-startup-surface.test.ts` now reads the single page
  and asserts both that no second script file is referenced and that the inline script carries the
[...]

Issue #5416, opened 2026-09-21
Linux: the desktop bootstrap page runs no JavaScript, so the startup surface can never report


lidge-jun:
### What happened

On Linux the desktop shell's bootstrap window is a static picture. The page renders, but none of
its JavaScript executes, so the startup surface never reports a phase, never reaches a terminal
state, and never explains itself. The window sits on the markup's default headline — "Starting
OpenCodex…" — with an empty checklist for as long as the app runs.

This is the surface whose entire purpose is to report when startup goes wrong, so on Linux a user
whose runtime fails to start has no information at all.

### Steps to reproduce

1. Build the Debian package from the desktop shell and install it on a GNOME session.
2. Launch the installed app.
3. Observe the window: headline unchanged, no phase rows, no terminal state.

### What I expected

The phase checklist from the shell's startup sequence, ending in a ready or a failed state with a
copyable diagnostic, as the surface does on the other platforms.

### Evidence

The asset itself is fine. An inline probe that fetches the script sees the embedded response:

    status=200 ct=text/javascript len=5941

So it is neither a missing asset nor a wrong MIME type. Narrowing further:

- With the configured `csp` removed from the shell configuration, the page comes fully alive: the
  checklist renders, "Registering the tray and the login item" completes, and "Resolving the
  configuration home and port" becomes the active row.
- With the configured `csp` present, **no** script runs — not the page's external script, and not
  an inline probe added ahead of it.
- Widening `script-src` to name the asset-protocol scheme and host changed nothing.
- Adding `'unsafe-inline'` to `script-src` also changed nothing, which is the signal that matters:
  `'unsafe-inline'` is ignored when a nonce or hash is present in the same directive. The policy the
  webview actually enforces is therefore not the policy in the configuration file — something is
  rewriting `script-src`, and the rewritten value admits neither the page's own script nor an
  inline one.

### Why the other surfaces look fine

The dashboard is loaded from the proxy's loopback origin and carries that origin's own headers, so
the shell's asset-protocol policy does not apply to it. Only the embedded bootstrap page is
affected, which is why the product appears to work right up until the moment it needs to explain
itself.

### Scope

The fix has to keep a real policy on the embedded page rather than removing it, so it is a decision
about how the bootstrap script is admitted — a nonce or hash the shell injects, or serving the page
so that its own origin is expressible — not a widening of the existing sources. Whatever the
mechanism, it needs a regression that fails when the surface cannot run its own script, because
[...]

PR #5451, opened 2026-09-21
fix(desktop): make the startup surface unable to wait forever

lidge-jun:
## Summary

On Linux the bootstrap window drew its six phase rows and then never changed again. The headline
stayed on the markup's default and every row stayed pending; captures at 26s and 46s were
pixel-identical. The page's own handshake deadline is 5s and it never fired, so all three invokes
answered — `apply` had been handed a falsy progress and returned at its first line, and no
`startup-phase` event ever arrived. Reported in #5416's follow-up; the machine had a user-account
npm runtime holding the port with the app not yet attached.

Three separate holes made that screen reachable, and each is closed here.

**`startup_snapshot` could answer `None`.** `app.try_state::<Startup>()` returns an `Option` and
the command mapped it straight through. The page returns early on a falsy progress, so the one
case it cannot render — a shell with no startup state — arrived as silence rather than as a
diagnostic. The command now always answers with a state, and a missing startup state is reported
as a failure with a copyable diagnostic instead of as nothing at all.

**The sequence could return before it had published anything.** `run` looked up `AppState` before
its first `report`, so that early return said nothing. The first report now happens before any
lookup that can fail.

**Nothing enforced the 30-second ceiling.** Several returns inside `run` publish no terminal state
— the missing `AppState`, and the spawn declined because an exit is already in flight — and a step
that outlives the deadline publishes none either. The surface then kept whatever it was last told
for as long as the process lived. A guard bound to the run now reports a terminal state when the
run does not. It is idempotent and generation-scoped, so it cannot overwrite a result the run
reported and one left over from an earlier run cannot fail the retry that replaced it. It waits a
short grace past the ceiling so the run's own failure, which names the endpoint, the home and how
the child ended, is the diagnostic on screen rather than the guard's thinner one.

`Phase::NotStarted` is new and deliberately absent from the phase list the page draws its
checklist from. The state used to be seeded with `registering`, so a shell that never began
rendered exactly like one that had just begun, on the surface whose whole job is to tell those
apart. A checklist row for it would be a step that never completes; it renders as a headline with
every row still pending, which is the honest picture.

### What this does not change

The page still reads the snapshot once on load and then follows the event stream. If the event
stream itself delivers nothing, the surface now freezes on a state it was actually told rather
than on its own markup — an improvement and a diagnostic, but not a cure. Making the page re-read
the snapshot until the run is terminal would close that too, and it is left out here because it
edits `desktop/ui/index.html`, which #5445 is currently rewriting. It is worth doing on top of
that PR.

## Verification

**Local checks: NOT RUN.** The suite, `test:changed`, typecheck, builds, `cargo` in any form,
dependency installs and any live runtime were not run for this change. Hosted CI at the exact head
is the evidence, read per step rather than per job conclusion.

- Every claim about the failure was read out of the source rather than inferred from the symptom.
  The 5s page deadline is what rules out a hung command: `HANDSHAKE_DEADLINE_MS` bounds all three
  invokes, and a rejection routes to `reportPageFailure`, which rewrites the headline. A headline
  still on its default at 46s therefore means the invokes answered and the value was falsy.
- Each literal the new source oracles assert was extracted from the rebased tree and confirmed
  present, with the extraction asserted non-empty first so a silently failing scan cannot read as
  a pass. The three pre-existing ordering oracles in this file — registering before resolving
  before starting, the deadline line, and the diagnostic slice between `pub fn diagnostic(` and
  `fn report(` — were re-checked against the moved `report` call and still hold.
- Line widths were checked so the formatter has nothing to do: the only lines over 100 columns in
  the two Rust files are pre-existing prose comments.

Regression coverage, written and registered but **not executed here**:

- `tests/clients/desktop-startup-surface.test.ts` gains four cases: the snapshot command answers
  with a state rather than an `Option`; `not-started` exists and is not in `PHASES`; the run
  publishes before the lookup that can fail; and the run is followed by a settle, with a deadline
  guard past the ceiling that is both idempotent and generation-scoped.
- `startup.rs` gains four unit tests in its own `#[cfg(test)]` module: `NotStarted` is absent from
  the checklist and unresolvable from a published id while every checklist phase round-trips; a
  fresh sequence reports not-started and is unsettled; the no-state snapshot is a retryable
  failure carrying a reason; and only a terminal state settles a run.

`structure/desktop-shell.md` records the terminal-state guarantee and the not-started state, since
`desktop/src-tauri/src` is an area that document owns.

### A neighbouring oracle that was passing vacuously

Pushing the fix turned `test 4/4` red on a case in `desktop-cli-contracts` that had nothing to do
with the change in behaviour. It sliced the sequence at the literal
`async fn run(app: &AppHandle)`; giving `run` its start instant changed that string, `indexOf`
answered -1, and `slice(-1)` returns the last character rather than failing. Every index the case
computes then became -1, and its four ordering assertions were comparing -1 against -1. It failed
loudly only because the first of them is `toBeGreaterThan(-1)`; the other three would have passed
against an empty string.

The second commit anchors both run oracles on the function name, which is what they are actually
about, and asserts the anchor was found before slicing. Every `indexOf` and `toContain` literal
in all seven `desktop-*` oracles was then re-extracted and checked against the shell sources —
335 anchors, no misses beyond the three files that read `app/Package.swift` and their own
fixtures, which are known and unrelated.

### Hosted CI at this head

Green, read per step. `desktop shell` ran all three of its checks rather than skipping past a
failure: `Check Rust formatting`, `Run Rust clippy` and `Run Rust tests`, which is where the four
new unit tests executed. All four `test` shards, `macos 1/2`, `macos 2/2`,
`macos widget + bundle`, `structure gate`, `gates`, `storage policy`, `api usage`,
`docker smoke`, the three keyring jobs, the three service smokes and the aggregate `ci` are all
success.

## Checklist

Scope is the startup sequence, its command surface, the structure document that owns it and the
matching tests. No credential, token, workflow or network path is touched, and the new diagnostic
text carries the version, the OS, the phase and the reason — nothing about the user or the host.

## Summary by CodeRabbit

- **Bug Fixes**
  - Desktop startup now always displays a clear status, including when startup state is unavailable.
  - Startup failures and runs exceeding the time limit now resolve to a terminal status instead of remaining incomplete.
  - Startup progress now distinguishes between not started and actively registering.
  - Early startup failures provide more useful diagnostics.

- **Tests**
  - Expanded coverage for startup states, fallback messaging, timeout handling, and progress sequencing.

- **Documentation**
  - Updated desktop shell documentation to reflect startup status and timeout behavior.


lidge-jun:
## 리뷰 · 우선순위 52 / 80

리눅스에서 데스크톱 앱을 켜면, 시작 화면이 여섯 줄을 그려 놓고 그대로 멈췄습니다. 제목은 처음 글 그대로였고, 줄은 전부 대기였습니다. 26초 화면과 46초 화면이 같았습니다. 페이지는 5초 안에 답이 없으면 실패로 바꾸는데, 그 실패도 안 떴습니다. 명령은 답을 했습니다. 답이 비어 있어서 화면이 그 답을 버렸을 뿐입니다.

빈 답이 나온 길은 세 가지입니다. 시작 상태를 물어보면 `없음`이 올 수 있었습니다. 시작 함수는 말하기 전에 앱 상태를 먼저 찾아서, 거기서 끝나면 아무 말도 안 했습니다. 30초가 지나도 끝을 강제하는 장치가 없어서, 중간에 그냥 돌아가거나 한 단계가 오래 걸리면 화면은 마지막 말을 앱이 꺼질 때까지 유지했습니다.

이제는 시작 상태를 물으면 항상 상태 하나가 옵니다. 상태가 없으면 "시작 상태를 못 찾았다"는 실패 글입니다. 시작 함수는 찾기 전에 "등록 중"을 먼저 알립니다. 30초에 2초를 더 기다려도 끝이 없으면, 그 실행만 실패로 닫습니다. 이미 성공이나 실패를 말한 실행은 덮지 않습니다. 다시 시도로 바뀐 다음 실행은 예전 타이머가 실패시키지 않습니다. "아직 시작 안 함"은 체크 목록 줄이 아닙니다. 예전에는 시작도 안 한 화면이 "등록 중"과 똑같이 보였습니다.

베이스는 `dev`입니다. 이 커밋에서 desktop shell과 테스트 잡은 통과했습니다.

라인 - `desktop/src-tauri/src/startup.rs`의 `unavailable` - 실패를 만들면 다시 시도가 켜집니다. `Progress::new`가 실패면 `can_retry`를 참으로 둡니다. 이 함수는 시작 상태가 앱에 없을 때만 쓰입니다. 다시 시도는 `begin`을 부르는데, `begin`은 그 상태가 없으면 바로 돌아갑니다. 버튼은 눌리지만 아무 일도 없습니다. 테스트 `the_snapshot_never_answers_with_nothing`은 버튼이 켜진 상태를 정답으로 박아 둡니다.

라인 - 같은 파일 `begin`의 `running` - 32초 타이머가 실패를 말해도 `running`은 `run`이 끝날 때까지 참입니다. 화면은 다시 시도를 켭니다. 누르면 `begin`이 이미 돌고 있다고 보고 새 실행을 만들지 않습니다. 페이지는 그 클릭에서 버튼을 끄고, 새 소식이 올 때까지 다시 켜지 않습니다. 막힌 단계가 영영 안 끝나면, 화면은 더 이상 기다리지 않지만 다시 시도는 한 번 누르면 다시 안 됩니다.

라인 - 같은 파일 `settle` - 이미 끝났는지 본 뒤, 발표하기 전에 `run`이 성공을 쓸 수 있습니다. 그 다음 `emit`이 성공을 실패로 덮습니다. 확인과 발표가 한 잠금 안에 있지 않습니다. 2초 여유는 이 틈을 줄일 뿐입니다. 타이머와 `run`이 끝난 뒤의 `settle`이 둘 다 "아직 안 끝남"을 통과하면, 더 자세한 실패 글이 짧은 글로 바뀌기도 합니다.

메인테이너의 판단이 필요한 지점

시작 상태가 없을 때 다시 시도 버튼을 끌지, 그때도 다시 시도가 실제로 돌게 할지 정해 주세요. 32초 타이머가 실패를 말할 때 `running`을 내려 다시 시도가 되게 할지, 막힌 `run`이 끝날 때까지 버튼을 꺼 둘지도요. 페이지가 끝날 때까지 스냅샷을 다시 읽는 일은 #5445가 `desktop/ui`를 고치고 있어서 이번에 빼 두었습니다. 그 후속을 #5445 다음에 할지도 정해 주세요.

너의 추천

방향은 맞습니다. 합치기 전에 `unavailable`의 다시 시도를 끄고, 타이머가 실패를 말할 때 `running`을 내리세요. `settle`의 끝 확인과 실패 발표는 같은 잠금 안에서 처리하면 됩니다. 베이스는 `dev`입니다. types.ts/config.ts 분할은 없고, 닫을 중복 PR도 없습니다.

이 댓글은 grok-bot이 작성했습니다

PR #5452, opened 2026-09-21
feat(desktop): open a usage popup from the tray icon

lidge-jun:
## Summary

Carries #5436 onto the current desktop shell. A left click on the tray icon now opens a small
anchored window showing today and 30-day usage, the chart, the model rows and per-account limits,
instead of opening the whole dashboard. Reading the current numbers is the reason to look at a
tray icon at all, and the dashboard is still one menu item away.

The original branch could not be merged: the shell renamed the tray menu handle, stopped passing a
proxy into `install` (the tray is built before a runtime is resolved, so every use reads the
current client from the app), moved tray installation and the Linux window decision into the
startup sequence, and routed Quit and Stop through the exit coordinator. The feature is therefore
re-expressed on that shape rather than merged into it, and three of the review findings on the
original are resolved by construction:

- **The tray title keeps its behaviour.** The original rewrote the title with empty usage whenever
  a settings read failed, so a good number became "Unavailable" the moment one request did not
  answer. This tree already returns early on a failed read and leaves the last title standing, and
  that is kept. The title also keeps its existing period; the popup answers the detailed question,
  so adding it does not silently change what the title means. `usage_summary` is restored for that
  reason — the original deleted it.
- **Linux does not open the dashboard unconditionally.** The original forced the window open on
  every Linux start. The startup sequence already probes for a real tray host and shows the window
  only when there is none, which is the same intent stated more precisely.
- **The two usage columns are independent.** `showToday` hid the 30-day figure along with today's.
  The setting names one column and there is no setting that asks for neither, so each column now
  answers for itself.

Two further changes, both small and both about not lying to the user:

- The popup ignores focus loss for 400 ms after being shown. Closing on focus loss is what makes
  it feel like a menu; the cost is that a platform which hands focus back to the shell right after
  the click closes it in the same gesture that opened it, and the user sees a flash and no window.
- The navigation gate reads its own constants. `is_tray_url`, `is_close_url` and `is_dashboard_url`
  restated the query and fragment that `TRAY_PATH`, `CLOSE_PATH` and `DASHBOARD_PATH` already
  define, so the constants were dead outside tests and a matcher could have kept answering yes
  after the page it names moved. These three decide what the popup may navigate to, so a stale yes
  is the failure that matters.

The token abbreviation fix comes along unchanged: ten million tokens rendered as `1M` because the
trailing zero was trimmed from an integer result.

## The popup now looks like the widget it sits next to

It was a flat `#202022` rectangle beside a WidgetKit widget that uses the system material,
rounded numerals and secondary-tone labels. macOS applies the active HUD window material at a
12-point radius and Windows applies Acrylic, both through Tauri's own effects builder, so no new
dependency is involved.

Linux stays opaque, and that asymmetry is the part that needed care: blur there belongs to the
compositor, and a transparent stylesheet on an opaque window paints a hole instead of a panel. One
`cfg` constant, `VIBRANT_SURFACE`, decides both the native builder and the `data-tray-vibrancy`
hook the page selects on, so the two cannot disagree. `gui-tray-vibrancy-surface.test.ts` reads
`popup.rs` and `tray.css` together and fails if the flag and the selector drift apart — nothing in
either toolchain connects them, and a silent rename would either lose the glass or punch the hole.

Transparent windows on macOS require the `macos-private-api` feature and `app.macOSPrivateApi`.
That forecloses Mac App Store submission; this shell ships as a Developer ID DMG, so the door it
closes is one this channel does not use.

The stylesheet also folds in the duplicate rules that had accumulated at the end of the file,
overriding blocks a few lines above them.

## React Doctor and rustfmt

Two gates were red and neither failure was visible where a reviewer would look.

`desktop shell` failed on `cargo fmt --check`, which gates clippy and the Rust tests, so neither
had run since the popup landed on this branch. Fixed; `cargo fmt --check` and `cargo check` both
exit 0 locally.

`react-doctor` reported nine blocking findings, and the action is configured with `comment: false`,
`review-comments: false` and `commit-status: false`, so they existed only inside the run's job
summary. Reproduced with the repository's own pinned scan and fixed at the root: the barrel import,
five array lookups inside loops hoisted to Sets, the pure helper moved to module scope, and the
quota rows keyed by an identity `quotaWindows()` now assigns. That last one also closes a bug the
index key was hiding — a provider may report two custom windows under one label, and the row key
has to stay unique anyway or React reconciles two different windows onto the same row. The
post-await state writes keep their single `active()` guard, which checks both disposal and the
`AbortSignal`; the rule cannot see the property through a helper, so that one finding carries a
scoped suppression stating the evidence.

## Verification

- **Repository suites, typecheck and build on the development machine: NOT RUN**, per the standing
  rule; hosted CI at the exact head is the authority.
- `cargo fmt --check` and `cargo check` on the desktop crate: exit 0. The pinned React Doctor
  scan and `bun run lint` in `gui`: clean.
- Compiled and packaged on a Linux verification machine from this branch: `tauri build --bundles
  deb` finishes with **zero warnings and zero errors**, which matters because the desktop-shell job
  treats warnings as errors. Two were found and fixed that way: the popup imported a module this
  tree does not have, and the two path constants were dead outside tests.
- Not verified: the popup itself on a real tray. The Linux session used for packaging has no
  StatusNotifier host, which is exactly why the menu item exists there, so neither the icon click
  nor the menu item can be exercised on it. macOS rendering and Escape-to-hide were verified by the
  original author; Windows click-then-focus behaviour remains unverified, which is what the grace
  period above is for.

## Checklist

Co-authored-by: JayYun98 <JayYun98@users.noreply.github.com>

## Summary by CodeRabbit

- **New Features**
  - Added a tray-anchored usage popup with daily usage, costs, charts, model details, and provider quota information.
  - Tray interactions open the popup when available, with fallback to the main dashboard; Linux also provides a “Show Usage” menu option.
  - Added automatic refresh, retry controls, localized labels, and native translucent styling on supported platforms.
- **Bug Fixes**
  - Preserved trailing zeros in integer abbreviations while continuing to trim unnecessary zeros from fractional values.
- **Documentation**
  - Added guides covering tray usage, popup behavior, metrics, quotas, settings, and the compact usage view.

## Screenshot

Built and run on macOS from this branch. The window is the popup's page, rendered by the bundled
GUI against a runtime started with an isolated configuration home, so the numbers are that
instance's own traffic rather than anything from the machine's real accounts.

Tray usage popup: Today and 30d columns, a model row, the provider section with its accounts, a refresh line and a dashboard link

Both usage columns are visible together, which is the fix for the setting that used to hide the
30-day figure along with today's.

## What the macOS build turned up

Two things worth recording, neither of them introduced by this change:

- **The desktop bundle can ship a stale CLI.** `prepare-sidecar` builds the standalone binary only
  when the file is missing, so a bundle built today shipped yesterday's `ocx`. The app reported it
  precisely — "the bundled CLI's resolve contract has not landed" — which is the startup surface
  doing its job, but the packaging step should not be able to produce that mismatch silently.
- **A freshly compiled standalone binary is killed on macOS until it is re-signed.** The compiler's
  linker-signed ad-hoc signature is refused; `codesign --force -s -` on the binary fixes it. Before
  that the app showed "the bundled CLI could not resolve the runtime (exit no exit code)", which is
  exactly what a SIGKILL looks like from the parent's side.


lidge-jun:
## 리뷰 · 우선순위 40 / 80

이 PR은 트레이 아이콘을 왼쪽 클릭하면 대시보드 전체가 아니라, 아이콘 옆에 붙는 작은 사용량 창이 열리게 합니다. 오늘과 30일 합계, 차트, 모델 목록, 계정별 한도가 그 창에 나옵니다. 대시보드는 메뉴와 창 안 링크로 그대로 갈 수 있습니다. #5436을 지금 데스크톱 껍데기 위에 다시 올린 것이고, base는 `dev`입니다. 트레이 제목은 예전처럼 두고, 숫자 줄임에서 천만 토큰이 `1M`으로 잘리던 것도 고쳤습니다. 창이 막 열린 뒤 0.4초 동안은 포커스를 잃어도 닫히지 않게 해서, 클릭하자마자 사라지는 깜빡임을 막으려 합니다.

라인 - `desktop/src-tauri/src/popup.rs` (`mark_shown`, `within_focus_grace`): 처음 보여줄 때 시계가 그 순간에 시작해서 경과 시간이 0이 됩니다. 0은 “한 번도 안 열림”으로 취급해서, 첫 클릭에는 0.4초 유예가 없습니다. 유예가 막으려던 깜빡임이 첫 열기에서 그대로 날 수 있습니다. 0이 아닌 시각을 저장하세요.
라인 - 같은 파일 `Focused(false)`: 0.4초 안에 온 포커스 상실은 버리고 끝입니다. 그 뒤에 다시 확인하지 않아서, 창이 포커스 없이 맨 위에 남고 바깥 클릭으로 안 닫힐 수 있습니다. 유예가 끝난 뒤에도 포커스가 없으면 그때 닫으면 됩니다.
라인 - `gui/src/pages/tray-data.ts` (`quotaWindows`): 대시보드 `QuotaBars`는 `creditsUsd`를 구독 크레딧 줄로 보여 주는데, 이 창은 퍼센트 창과 `customWindows`만 봅니다. 크레딧만 있는 API 키 계정은 한도가 있는데도 “할당량 없음”으로 나옵니다.
라인 - `docs-site/src/content/docs/guides/desktop-app.md`: 트레이 메뉴에 **Refresh now**가 없고, **Show Usage**는 리눅스에만 있습니다. 제목은 `/api/usage?range=7d`의 지표 하나를 보여 주지, 오늘 요청·토큰·비용을 같이 보여 주지 않습니다. **Today**를 끄면 팝업의 오늘 칸만 숨고 제목은 안 바뀝니다. “리눅스는 시작할 때 대시보드가 열린다”도, 트레이가 없을 때만 여는 지금 동작보다 넓습니다.
라인 - `desktop/src-tauri/src/popup.rs`: `cargo fmt --check`가 여기서 실패합니다. desktop shell CI는 포맷에서 멈춰서 clippy와 Rust 테스트는 안 돌았습니다. 긴 `matches` 줄과 빈 줄 두 개입니다.
라인 - `desktop/src-tauri/src/tray.rs` 리눅스 `show-usage`: 위치를 `(0, 0)`으로 고정해서, 트레이와 멀리 떨어진 화면 구석에 창이 뜰 수 있습니다.

메인테이너의 판단이 필요한 지점

#5436은 아직 열려 있습니다. 이 PR이 그 기능을 현재 셸에 다시 쓴 것이면 #5436은 닫는 쪽이 맞습니다. 리눅스 메뉴로 열 때 창을 구석에 둘지, 커서나 트레이 쪽으로 둘지도 정해 주세요. 가이드를 지금 코드에 맞출지, 없는 메뉴를 실제로 넣을지도요. UI 스크린샷이 없어서 enforce-target이 실패하고 draft로 남아 있습니다. 작성자도 실제 트레이 클릭은 확인하지 못했다고 적었습니다.

너의 추천

방향은 맞습니다. 합치기 전에 첫 열기 유예(0을 “안 열림”으로 보지 말 것), 유예 중 포커스 상실을 나중에 다시 확인하는 것, `creditsUsd` 줄, 가이드를 지금 메뉴와 제목에 맞게 고치고, `cargo fmt`를 통과시키면 됩니다. #5436은 이 PR로 대체된 중복이니 닫으세요. types.ts/config.ts 분할은 이 PR과 관계없습니다.

이 댓글은 grok-bot이 작성했습니다

